PR #2193, opened 2025-11-10
[storage/adb/any] use authenticated journal throughout adb::any

roberto-bayardo:
- Make adb::any variants use Authenticated journal, allowing removal of more shared code.

- Change authenticated journal root() to be non-mut for API compatibility.

- Make shared floor raising logic independent of the MMR, moving it up one level in the hierarchy, and allowing it to be used by the unauthenticated store as well.

- Made the adb::any variants + current adb variants use the authenticated journal's hasher instead of their own.

- changed authenticated journal's "op_count" to "size" since "operation" is a database level concept.

### DIFF
--- a/storage/src/adb/any/fixed/sync.rs
+++ b/storage/src/adb/any/fixed/sync.rs
@@ -9,7 +9,7 @@ use crate::{
     mmr::{Location, Position, StandardHasher},
     translator::Translator,
 };
-use commonware_codec::{CodecFixed, Encode as _};
+use commonware_codec::CodecFixed;
 use commonware_cryptography::Hasher;
 use commonware_runtime::{
     buffer::Append, telemetry::metrics::status::GaugeExt, Blob, Clock, Metrics, Storage,
@@ -57,7 +57,7 @@ where
     async fn from_sync_result(
         context: Self::Context,
         db_config: Self::Config,
-        mut log: Self::Journal,
+        log: Self::Journal,
         pinned_nodes: Option<Vec<Self::Digest>>,
         range: Range<Location>,
         apply_batch_size: usize,
@@ -82,8 +82,10 @@ where
         )
         .await?;
 
-        let mut hasher = StandardHasher::<H>::new();
-        let mmr = authenticated::Journal::align(mmr, &mut log, &mut hasher).await?;
+        let hasher = StandardHasher::<H>::new();
+        let log =
+            authenticated::Journal::from_components(mmr, log, hasher, apply_batch_size as u64)
+                .await?;
 
         // Build the snapshot from the log.
         let mut snapshot =
@@ -91,15 +93,8 @@ where
         let active_keys =
             build_snapshot_from_log(range.start, &log, &mut snapshot, |_, _| {}).await?;
 
-        // Create an authenticated journal from the MMR and log.
-        let authenticated_log = authenticated::Journal {
-            mmr,
-            journal: log,
-            hasher,
-        };
-
         let mut db = Any {
-            log: authenticated_log,
+            log,
             active_keys,
             inactivity_floor_loc: range.start,
             snapshot,
--- a/storage/src/journal/authenticated.rs
+++ b/storage/src/journal/authenticated.rs
@@ -86,13 +86,29 @@ where
     O: Encode,
     H: Hasher,
 {
-    /// Align `mmr` to be consistent with `journal`.
-    /// Any elements in `mmr` that aren't in `journal` are popped, and any elements in `journal`
-    /// that aren't in `mmr` are added to `mmr`.
-    pub(crate) async fn align(
+    /// Create a new [Journal] from the given components after aligning the MMR with the journal.
+    pub async fn from_components(
+        mmr: Mmr<E, H>,
+        journal: C,
+        mut hasher: StandardHasher<H>,
+        apply_batch_size: u64,
+    ) -> Result<Self, Error> {
+        let mmr = Self::align(mmr, &journal, &mut hasher, apply_batch_size).await?;
+        Ok(Self {
+            mmr,
+            journal,
+            hasher,
+        })
+    }
+
+    /// Align `mmr` to be consistent with `journal`. Any elements in `mmr` that aren't in `journal` are popped, and any
+    /// elements in `journal` that aren't in `mmr` are added to `mmr`. Operations are added to `mmr` in batches of size
+    /// `apply_batch_size` to avoid memory bloat.
+    async fn align(
         mut mmr: Mmr<E, H>,
         journal: &C,
         hasher: &mut StandardHasher<H>,
+        apply_batch_size: u64,
     ) -> Result<Mmr<E, H>, Error> {
         // Pop any MMR elements that are ahead of the journal.
         // Note mmr_size is the size of the MMR in leaves, not positions.
@@ -114,10 +130,16 @@ where
             );
 
             let mut mmr = mmr.into_dirty();
+            let mut batch_size = 0;
             while mmr_size < journal_size {
                 let op = journal.read(*mmr_size).await?;
                 mmr.add_batched(hasher, &op.encode()).await?;
                 mmr_size += 1;
+                batch_size += 1;
+                if batch_size >= apply_batch_size {
+                    mmr = mmr.merkleize(hasher).into_dirty();
+                    batch_size = 0;
+                }
             }
             let mut mmr = mmr.merkleize(hasher);
             mmr.sync().await?;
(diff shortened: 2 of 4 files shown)


storage/src/adb/any/fixed/ordered.rs
```diff
@@ -663,57 +652,36 @@ impl<
...
+    /// Simulate an unclean shutdown by consuming the db. If sync_log is true, the log will be
+    /// synced before consuming.
     #[cfg(any(test, feature = "fuzzing"))]
-    pub async fn simulate_failure(
-        mut self,
-        sync_log: bool,
-        sync_mmr: bool,
-        write_limit: usize,
-    ) -> Result<(), Error> {
+    pub async fn simulate_failure(mut self, sync_log: bool) -> Result<(), Error> {
```
danlaine: Do we need this method anymore? If `sync_log` is true, this is equivalent to `sync`, in which case it's not really a failure mode. If it's `false`, then we could simply `drop` `self` instead of calling this.

storage/src/adb/any/fixed/unordered.rs
```diff
@@ -290,57 +279,36 @@ impl<E: Storage + Clock + Metrics, K: Array, V: CodecFixed<Cfg = ()>, H: Hasher,
...
+    /// Simulate an unclean shutdown by consuming the db. If sync_log is true, the log will be
+    /// synced before consuming.
     #[cfg(any(test, feature = "fuzzing"))]
-    pub async fn simulate_failure(
-        mut self,
-        sync_log: bool,
-        sync_mmr: bool,
-        write_limit: usize,
-    ) -> Result<(), Error> {
+    pub async fn simulate_failure(mut self, sync_log: bool) -> Result<(), Error> {
```
danlaine, line 311: Same comment as above about `simulate_failure`

storage/src/adb/any/variable/mod.rs
```diff
@@ -64,19 +68,13 @@ pub struct Config<T: Translator, C> {
     pub buffer_pool: PoolRef,
 }
 
+type AuthenticatedLog<E, K, V, H> =
+    authenticated::Journal<E, Journal<E, Operation<K, V>>, Operation<K, V>, H>;
 /// A key-value ADB based on an MMR over its log of operations, supporting authentication of any
```
danlaine, line 75:
```suggestion

/// A key-value ADB based on an MMR over its log of operations, supporting authentication of any
```

storage/src/adb/any/fixed/ordered.rs
roberto-bayardo: Changed to commit() instead of sync.  In this case it's simulating a failure but one that should already be tested by the authenticated log I suppose, so maybe it's still worth removing.

storage/src/adb/any/variable/mod.rs
roberto-bayardo: fixed

storage/src/adb/any/fixed/sync.rs
```diff
@@ -456,7 +461,7 @@ mod tests {
 
             // Capture the database state before closing
             let final_synced_op_count = got_db.op_count();
-            let final_synced_log_size = got_db.log.size();
+            let final_synced_log_size = got_db.log.journal.size();
```
danlaine: Nit: this could be `got_db.log.journal.size()`? Should be the same thing (and also the same as `got_db.op_count()` so either way is fine. Same for a few other places.

storage/src/adb/any/fixed/ordered.rs
```diff
@@ -40,6 +35,9 @@ enum UpdateLocResult<K: Array + Ord, V: CodecFixed<Cfg = ()>> {
     NotExists(KeyData<K, V>),
 }
 
+type AuthenticatedLog<E, K, V, H> =
```
danlaine: Nit: we could share this definition with `unordered.rs` if we declare it in `fixed/mod.rs`:

```rs
pub(super) type AuthenticatedLog<E, O, H> = authenticated::Journal<E, Journal<E, O>, O, H>;
```

storage/src/adb/current/ordered.rs
```diff
@@ -299,13 +296,12 @@ impl<
 
         // Apply the commit operation with the new inactivity floor.
         let loc = self.any.inactivity_floor_loc;
-        let mut shared = self.any.as_shared();
-        shared.apply_op(Operation::CommitFloor(loc)).await?;
+        self.any.log.append(Operation::CommitFloor(loc)).await?;
```
danlaine: Nit: this returns the same value we assign to `self.last_commit_loc` so we could do:

```rs
self.last_commit_loc = Some(self.any.log.append(Operation::CommitFloor(loc)).await?);
```

or similar. I'm fine with it as is, too.

storage/src/adb/current/unordered.rs
```diff
@@ -265,22 +264,21 @@ impl<
...
                 .raise_floor_with_bitmap(&mut self.status, loc)
                 .await?;
         }
         self.any.steps = 0;
 
         // Apply the commit operation with the new inactivity floor.
         let loc = self.any.inactivity_floor_loc;
-        let mut shared = self.any.as_shared();
-        shared.apply_op(Operation::CommitFloor(loc)).await?;
+        self.any.log.append(Operation::CommitFloor(loc)).await?;
```
danlaine: Same comment as above about this returning `last_commit_loc`

storage/src/journal/authenticated.rs
```diff
@@ -262,8 +255,12 @@ where
     }
 
     /// Return the root of the MMR.
-    pub fn root(&mut self) -> H::Digest {
-        self.mmr.root(&mut self.hasher)
+    pub fn root(&self, hasher: &mut StandardHasher<H>) -> H::Digest {
+        self.mmr.root(hasher)
+    }
+
+    pub fn size(&self) -> Location {
```
danlaine, line 291:
```suggestion
    /// Return the number of operations in the journal.
    pub fn size(&self) -> Location {
```

storage/src/journal/authenticated.rs
```diff
@@ -400,23 +397,121 @@ where
...
+    O: Encode,
+    H: Hasher,
+{
+    type Item = O;
+
+    async fn append(&mut self, item: Self::Item) -> Result<u64, JournalError> {
+        let res = self.append(item).await.map_err(|error| match error {
+            Error::Journal(inner) => inner,
+            Error::Mmr(inner) => {
+                JournalError::Corruption(format!("MMR error during append: {}", inner))
```
danlaine: Perhaps not in this PR, but I wonder whether we should change the return type of `Contiguous` to allow for more natural-feeling values here. We use "corruption" elsewhere to mean that persisted data is borked, which isn't exactly what's happening here, but the "right" `JournalError` to return here isn't super apparent.

storage/src/journal/authenticated.rs
```diff
@@ -215,20 +213,20 @@ where
     /// - Returns [Error::Journal] with [crate::journal::Error::ItemPruned] if `start_loc` has been pruned.
     pub async fn historical_proof(
```
danlaine, line 242: Nit: the comment still references renamed parameter `op_count`

storage/src/journal/authenticated.rs
```diff
@@ -653,15 +748,15 @@ mod tests {
 
             // Don't sync - these are uncommitted
             // After alignment, they should be discarded
-            let op_count_before = journal.op_count();
-            assert_eq!(op_count_before, Location::new_unchecked(20));
+            let op_count_before = journal.size();
```
danlaine: nit
```suggestion
            let size_before = journal.size();
```

storage/src/adb/any/fixed/sync.rs
roberto-bayardo: done

storage/src/adb/any/fixed/ordered.rs
roberto-bayardo: done

storage/src/adb/current/ordered.rs
roberto-bayardo: done

storage/src/adb/current/unordered.rs
roberto-bayardo: done

storage/src/journal/authenticated.rs
roberto-bayardo: fixed

storage/src/journal/authenticated.rs

storage/src/journal/authenticated.rs
roberto-bayardo: Changed to 

`/// Returns the number of items in the journal.`

storage/src/journal/authenticated.rs
roberto-bayardo: Changed to return an "Mmr" error using anyhow::Error

PR #2196, opened 2025-11-11
[coding] Align CodecCfg to just take a usize parameter

### LINKED ISSUE #2190: [coding] Add a uniform static codec config
It's a lot easier for applications to have:

- a single type for the config needed to decode [Shard](https://docs.rs/commonware-coding/latest/commonware_coding/trait.Scheme.html#associatedtype.Shard) and [ReShard](https://docs.rs/commonware-coding/latest/commonware_coding/trait.Scheme.html#associatedtype.ReShard),
- and a type which can be determined statically without changing as reconfiguration happens.

Guaranteeing this will likely take the form of an upper bound on size.

cronokirby:
Closes #2190.

### DIFF
--- a/coding/src/reed_solomon/mod.rs
+++ b/coding/src/reed_solomon/mod.rs
@@ -80,13 +80,10 @@ impl<H: Hasher> Write for Chunk<H> {
 
 impl<H: Hasher> Read for Chunk<H> {
     /// The maximum size of the shard.
-    type Cfg = crate::Cfg;
+    type Cfg = crate::CodecConfig;
 
-    fn read_cfg(
-        reader: &mut impl Buf,
-        &(max_bytes, _): &Self::Cfg,
-    ) -> Result<Self, commonware_codec::Error> {
-        let shard = Vec::<u8>::read_range(reader, ..=max_bytes)?;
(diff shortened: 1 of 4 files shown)


coding/src/reed_solomon/mod.rs
```diff
@@ -80,13 +80,10 @@ impl<H: Hasher> Write for Chunk<H> {
 
 impl<H: Hasher> Read for Chunk<H> {
     /// The maximum size of the shard.
-    type Cfg = crate::Cfg;
+    type Cfg = crate::CodecConfig;
```
clabby, line 83: nit - can we just have this be a type alias for `usize` and document that it's an upper bound, rather than having it be a struct?

If you prefer this for readability, not opposed, though.

cronokirby: I think a field name is a lot better documentation than a comment, and opens the way up to having more fields if that ends up being better.

If you use `coding::CodecConfig { maximum_shard_size: ... }` then you don't even need to look up the docs, right?

clabby (approved):

clabby:
Rebased on #2197 to satisfy the new merge group requirement.

Issue #2075, opened 2025-10-31
panic in simplex/actors/voter/actor.rs#L174


dnkolegov-ar:
This is another issue like https://github.com/commonwarexyz/monorepo/issues/2070 while trying to implement the Twins test.

In [add_recovered_proposal](https://github.com/commonwarexyz/monorepo/blob/20ed6d2d261c43c17722427372b881029116417b/consensus/src/simplex/actors/voter/actor.rs#L174), the node panics if `proposal != previous`.

```
fn add_recovered_proposal(&mut self, proposal: Proposal<D>) {
        if self.proposal.is_none() {
            debug!(?proposal, "setting verified proposal");
            self.proposal = Some(proposal);
        } else if let Some(previous) = &self.proposal {
            assert_eq!(proposal, *previous);
        }
    }
```

From my perspective, it is a DoS vulnerability, allowing an equivocating leader to crash a correct node:
1. The target node first receives a valid notarize vote with the proposal A itself from the malicious leader, which sets `self.proposal = Some(A)`.
2. Later, it receives a valid notarization for proposal B for the same view (produced because the leader equivocated).
3. `add_recovered_proposal(B)` sees `self.proposal = Some(A)` and hits assert_eq!(B, A), panicking.

This is also reproduced in the same setting in the `all_online` test as the basis for twins.

patrick-ogrady:
Confirmed this is not handled right (IIRC I did prior to the buffer refactor but clearly what is in main is wrong)

PR #2192, opened 2025-11-10
[consensus::simplex] Refactor Simplex Recovery

BrendanChou:
Fixes: #2149 
Fixes: #1224

Despite 2f+1 honest nodes being online, there exists a certain configuration of certificates that prevents progress from being made (no proposal ever has `f+1` support to drive recovery).

## TODO
- [x] send certificates to `simplex::resolver` on restart
- [x] `resolver` now blocks peers that send invalid responses
- [x] `resolver` no longer starves requests when peers are saturated
- [x] missing certificates during proposal no longer prevent further proposal attempts
- [x] introduce `StaticManager` util

Issue #2177, opened 2025-11-08
[consensus/simplex] Add `Activity::ConflictingCertificate` Activity


patrick-ogrady:
In some cases, we'll detect equivocation with `ConflictingNotarize`. However, a smart adversarial leader will send us only 1 notarize vote and they way we'll detect equivocation is by receiving a certificate for a proposal different than the leader's notarize. We can return this notarize or finalize and certificate as `ConflictingCertificate` (could be either notarization or finalization).

PR #2107, opened 2025-11-03
[consensus/simplex] Add Equivocation Testing

andresilva:
Fixes #2075

PR #2199, opened 2025-11-11
[consensus/simplex] `voter` Refactor

patrick-ogrady:
Replaces: #2194 
Replaces: #2107
Fixes: #2075 

This is a substantial refactor of our existing logic. While that always introduces the possibility of bugs, I feel this is a much better foundation to build on.

## TODO
- [x] Restore logs: https://github.com/commonwarexyz/monorepo/pull/2057
- [x] Cleanup/review tests (many are auto-generated)
- [x] Review helpers from Round 
- [x] Add test for notarize after nullify
- [ ] only operate over certificates in `voter`: https://github.com/commonwarexyz/monorepo/issues/2228
- [ ] remove f+1 fault fetch (future work): https://github.com/commonwarexyz/monorepo/pull/2192
- [ ] Add equivocation events: https://github.com/commonwarexyz/monorepo/issues/2177

### DIFF
--- a/consensus/src/simplex/actors/voter/actor.rs
+++ b/consensus/src/simplex/actors/voter/actor.rs
@@ -896,7 +896,7 @@ impl<
                     };
 
                     // Block if the epoch is not the current epoch
-                if msg.epoch() != self.state.epoch() {
+                    if msg.epoch() != self.state.epoch() {
                         warn!(?sender, "blocking peer for epoch mismatch");
                         self.blocker.block(sender).await;
                         continue;
--- a/consensus/src/simplex/actors/voter/round.rs
+++ b/consensus/src/simplex/actors/voter/round.rs
@@ -110,7 +110,7 @@ impl<S: Scheme, D: Digest> Round<S, D> {
     #[allow(clippy::type_complexity)]
     /// Returns the leader key and proposal when the view is ready for verification.
     pub fn should_verify(&self) -> Option<(Leader<S::PublicKey>, Proposal<D>)> {
-        let leader = self.leader.clone()?;
+        let leader = self.leader.as_ref()?;
         if self.is_signer(leader.idx) {
             return None;
         }
@@ -118,7 +118,7 @@ impl<S: Scheme, D: Digest> Round<S, D> {
             return None;
         }
         let proposal = self.proposal.proposal().cloned()?;
-        Some((leader, proposal))
+        Some((leader.clone(), proposal))
     }
 
     /// Marks that verification is in-flight; returns `false` to avoid duplicate requests.
@@ -207,8 +207,8 @@ impl<S: Scheme, D: Digest> Round<S, D> {
     }
 
     /// Returns how much time elapsed since the round started, if the clock monotonicity holds.
-    pub fn elapsed_since_start(&self, now: SystemTime) -> Option<Duration> {
-        now.duration_since(self.start).ok()
+    pub fn elapsed_since_start(&self, now: SystemTime) -> Duration {
+        now.duration_since(self.start).unwrap_or_default()
     }
 
     /// Completes the local proposal flow after the automaton returns a payload.
--- a/consensus/src/simplex/actors/voter/slot.rs
+++ b/consensus/src/simplex/actors/voter/slot.rs
@@ -163,8 +163,7 @@ where
                     self.requested_verify = true;
                 } else {
                     // If this isn't a certificate, we keep the proposal as-is
-                    retained = existing.clone();
-                    dropped = proposal.clone();
+                    (retained, dropped) = (dropped, retained);
                 }
                 self.status = Status::Replaced;
                 Change::Replaced { dropped, retained }
--- a/consensus/src/simplex/actors/voter/state.rs
+++ b/consensus/src/simplex/actors/voter/state.rs
@@ -23,6 +23,7 @@ use prometheus_client::metrics::{counter::Counter, gauge::Gauge, histogram::Hist
 use rand::{CryptoRng, Rng};
 use std::{
     collections::BTreeMap,
+    mem::replace,
     sync::{atomic::AtomicI64, Arc},
     time::{Duration, SystemTime},
 };
@@ -515,7 +516,7 @@ impl<E: Clock + Rng + CryptoRng + Metrics, S: Scheme, D: Digest> State<E, S, D> 
         let now = self.context.current();
         self.views
             .get(&view)
-            .and_then(|round| round.elapsed_since_start(now))
+            .map(|round| round.elapsed_since_start(now))
     }
 
     /// Immediately expires `view`, forcing its timeouts to trigger on the next tick.
@@ -602,23 +603,17 @@ impl<E: Clock + Rng + CryptoRng + Metrics, S: Scheme, D: Digest> State<E, S, D> 
     /// Drops any views that fall below the activity horizon and returns them for logging.
     pub fn prune(&mut self) -> Vec<View> {
         let min = self.min_active();
-        let mut removed = Vec::new();
-        while let Some(view) = self.views.keys().next().copied() {
-            if view >= min {
-                break;
-            }
-            self.views.remove(&view);
-            removed.push(view);
-        }
+        let kept = self.views.split_off(&min);
+        let removed = replace(&mut self.views, kept).into_keys().collect();
 
         // Update metrics
         let _ = self.tracked_views.try_set(self.views.len());
         removed
     }
 
-    /// Returns the certified payload for a view, checking certificates first,
-    /// then falling back to checking if we have enough votes to certify.
-    fn certified_payload(&self, view: View) -> Option<&D> {
+    /// Returns the quorum payload for a view, checking certificates first,
+    /// then falling back to checking if we have quorum votes.
+    fn quorum_payload(&self, view: View) -> Option<&D> {
         // Ensure proposal exists
         let round = self.views.get(&view)?;
         let payload = &round.proposal()?.payload;
@@ -656,7 +651,7 @@ impl<E: Clock + Rng + CryptoRng + Metrics, S: Scheme, D: Digest> State<E, S, D> 
             if cursor == GENESIS_VIEW {
                 return Ok((GENESIS_VIEW, self.genesis.unwrap()));
             }
-            if let Some(parent) = self.certified_payload(cursor) {
+            if let Some(parent) = self.quorum_payload(cursor) {
                 return Ok((cursor, *parent));
             }
             if self.is_nullified(cursor) {
@@ -693,7 +688,7 @@ impl<E: Clock + Rng + CryptoRng + Metrics, S: Scheme, D: Digest> State<E, S, D> 
                 if cursor == GENESIS_VIEW {
                     return Some(self.genesis.unwrap());
                 }
-                let payload = self.certified_payload(cursor).copied()?;
+                let payload = self.quorum_payload(cursor).copied()?;
                 return Some(payload);
             }
             if cursor == GENESIS_VIEW {
@@ -721,7 +716,7 @@ impl<E: Clock + Rng + CryptoRng + Metrics, S: Scheme, D: Digest> State<E, S, D> 
             notarizations: Vec::new(),
             nullifications: Vec::new(),
         };
-        if parent != GENESIS_VIEW && self.certified_payload(parent).is_none() {
+        if parent != GENESIS_VIEW && self.quorum_payload(parent).is_none() {
             missing.notarizations.push(parent);
         }
         missing.nullifications = ((parent + 1)..view)


consensus/src/simplex/types.rs
```diff
@@ -107,6 +113,87 @@ impl<T: Attributable> AttributableMap<T> {
...
 }
 
+/// Tracks notarize/nullify/finalize votes for a view.
+pub struct VoteTracker<S: Scheme, D: Digest> {
+    notarizes: AttributableMap<Notarize<S, D>>,
+    nullifies: AttributableMap<Nullify<S>>,
+    finalizes: AttributableMap<Finalize<S, D>>,
+}
+
+impl<S: Scheme, D: Digest> VoteTracker<S, D> {
```
patrick-ogrady, line 133: add docs

consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -483,10 +161,17 @@ impl<
         // Initialize store
         let (mailbox_sender, mailbox_receiver) = mpsc::channel(cfg.mailbox_size);
         let mailbox = Mailbox::new(mailbox_sender);
+        let core = SimplexCore::new(SimplexCoreConfig {
```
patrick-ogrady: nit: improve naming

consensus/src/simplex/actors/voter/mod.rs
```diff
@@ -1089,4 +1092,437 @@ mod tests {
...
+    #[test_traced]
+    fn test_certificate_overrides_existing_proposal() {
+        certificate_overrides_existing_proposal(bls12381_threshold::<MinPk, _>);
+        certificate_overrides_existing_proposal(bls12381_threshold::<MinSig, _>);
+        certificate_overrides_existing_proposal(bls12381_multisig::<MinPk, _>);
+        certificate_overrides_existing_proposal(bls12381_multisig::<MinSig, _>);
+        certificate_overrides_existing_proposal(ed25519);
+    }
+
+    /// Test that our proposal is dropped when it conflicts with a peer's notarize vote.
```
patrick-ogrady, line 1314: Handling this behavior is required for twins test to work properly.

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -0,0 +1,1573 @@
...
+            slot.update(&proposal_b, true),
+            ProposalChange::Replaced { .. }
+        ));
+        assert!(matches!(
+            slot.update(&proposal_b, true),
+            ProposalChange::Skipped
+        ));
+        assert_eq!(slot.status(), ProposalStatus::Replaced);
+    }
+}
```
patrick-ogrady, line 1521: Add an equivocation replay test

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -0,0 +1,2170 @@
...
+            .filter(|candidate| !self.is_nullified(*candidate))
+            .collect();
+        if missing.is_empty() {
+            return None;
+        }
+        Some(missing)
+    }
+}
+
+#[cfg(test)]
```
patrick-ogrady, line 732: TODO: review test usefulness

consensus/src/simplex/types.rs
```diff
@@ -107,6 +113,114 @@ impl<T: Attributable> AttributableMap<T> {
...
+    /// Per-signer nullify votes keyed by validator index.
+    nullifies: AttributableMap<Nullify<S>>,
+    /// Per-signer finalize votes keyed by validator index.
+    ///
+    /// Finalize votes include the proposal digest so the entire certificate can be
+    /// reconstructed once the quorum threshold is hit.
+    finalizes: AttributableMap<Finalize<S, D>>,
+}
+
+impl<S: Scheme, D: Digest> VoteTracker<S, D> {
```
BrendanChou, line 133: I'm not 100% convinced this is makes the code any cleaner than just a struct with 3 public `AttributableMap` fields, and a single `new(participants: usize)` function.

Seems fine to just call `votes.notarizes.insert(vote)` rather than `votes.insert_notarize(vote)` for example

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -0,0 +1,2331 @@
...
+/// Missing certificate data required for safely replaying proposal ancestry.
+#[derive(Debug, Clone, Default, PartialEq, Eq)]
+pub struct MissingCertificates {
+    /// Parent view referenced by the proposal.
+    pub parent: View,
+    /// All parent views whose notarizations we still need.
+    pub notarizations: Vec<View>,
+    /// All intermediate views whose nullifications we still need.
+    pub nullifications: Vec<View>,
+}
```
BrendanChou: Consider whether `notarizations` should not just be an `Option<View>` since it can always only just be the `parent` view. Also consider whether this type is necessary at all or whether the `missing_certificates` function can simply return a tuple of `Vec` that can be passed straight to the resolver if needed.

Also consider whether this should just be a type defined in `resolver/ingress.rs`

consensus/src/simplex/state.rs
```diff
@@ -0,0 +1,2331 @@
...
+            return false;
+        }
+        self.status = ProposalStatus::Verified;
+        true
+    }
+
+    #[cfg(test)]
+    fn has_requested_verify(&self) -> bool {
+        self.requested_verify
+    }
```
BrendanChou: Nit: is this function needed? Is the `requested_verify ` field not visible within the testing in this file?

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -0,0 +1,2440 @@
...
+            nullification: None,
+            broadcast_nullify: false,
+            broadcast_nullification: false,
+            finalization: None,
+            broadcast_finalize: false,
+            broadcast_finalization: false,
+        }
+    }
+
+    fn can_begin_propose(&self) -> Result<Leader<S::PublicKey>, ProposalError<S::PublicKey>> {
```
BrendanChou: The error result is simply thrown away when this function is called. Is it necessary to have it return the specific error? Is this just for testing?

consensus/src/simplex/state.rs
```diff
@@ -0,0 +1,2440 @@
...
+            self.activity_timeout,
+            self.last_finalized,
+            self.view,
+            pending,
+            allow_future,
+        )
+    }
+
+    pub fn is_me(&self, idx: u32) -> bool {
+        self.scheme.me().map(|me| me == idx).unwrap_or(false)
```
BrendanChou: Idiom: you use `.map(...).unwrap_or(false)` a lot of places, consider using `is_some_and(...)`

consensus/src/simplex/actors/voter/state.rs
BrendanChou, line 1: Should this file live in `actors/voter/`? Is it only used there?

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -0,0 +1,2440 @@
...
+            }
+            Err(_) => return VerifyStatus::NotReady,
+        };
+        let round = match self.views.get_mut(&view) {
+            Some(round) => round,
+            None => return VerifyStatus::NotReady,
+        };
+        if round.reserve_verify().is_err() {
+            return VerifyStatus::NotReady;
+        }
```
BrendanChou: you've called `let round = match self.views.get_mut(&view)` for the second time in this function. Move the `reserve_verify()` call to the top of this function when you grabbed the round the first time

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -0,0 +1,2440 @@
...
+    MissingAncestor(View),
+    NotReady,
+}
+
+/// Status of preparing a peer proposal for verification.
+#[derive(Debug, Clone)]
+pub enum VerifyStatus<P: PublicKey, D: Digest> {
+    Ready(VerifyReady<P, D>),
+    NotReady,
+}
```
BrendanChou: Based on where/how this is used, I think you should remove this type and just use `Option<VerifyReady>` where this type was previously used. And in fact I'm not sure that even keeping `VerifyReady` as a standalone type is much better than just using a tuple `(context, proposal)`

consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -528,560 +198,285 @@ impl<
...
+            }
+            Some(Err(HandleError::NotPending)) => {
+                debug!(?round, reason = "not pending", "dropping verified proposal");
+                false
+            }
+            None => {
+                // View is no longer relevant, drop the verified proposal
+                false
+            }
         }
```
BrendanChou: The ergonomics could be much better here if you:
- Have `verified` return a `Result<>` rather than a `Option<Result<>>`
- Make `HandleError` a true `thiserror` type with debug

That way you can just use `?err` in the debug rather than having this have so many different cases

consensus/src/simplex/state.rs
patrick-ogrady: Removed 👍

consensus/src/simplex/state.rs
patrick-ogrady: 👍

consensus/src/simplex/actors/voter/state.rs
patrick-ogrady: Could also be used for better logging I suppose

consensus/src/simplex/actors/voter/state.rs
patrick-ogrady: very good idea 🤦

consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -528,560 +201,282 @@ impl<
...
-        Some((
-            Self::is_me(&self.scheme, round.leader?),
-            elapsed.as_secs_f64(),
-        ))
+        let elapsed = self
+            .state
+            .elapsed_since_start(view, self.context.current())?;
+        let leader = self.state.leader_index(view)?;
+        Some((self.state.is_me(leader), elapsed.as_secs_f64()))
     }
```
BrendanChou: nit: just return `Option<f64>`. The option is not used anyway if the bool is false. So just return the elapsed value if you know the leader AND the leader is me

consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -528,560 +201,282 @@ impl<
...
-        let advance_deadline = self.context.current() + self.notarization_timeout;
-        let round = self.round_mut(view);
-        round.leader_deadline = Some(leader_deadline);
-        round.advance_deadline = Some(advance_deadline);
-        round.set_leader(seed);
-        self.view = view;
-
-        // Update metrics
-        let _ = self.current_view.try_set(view);
     }
```
BrendanChou: It occurs to me that the handling of the timeouts is a little strange (code smell) here and elsewhere in this file. Right now, the state handles the leader and notarization timeouts, but the voter actor handles the nullify retry timeout.

I think it would be cleaner if either the voter or the state were responsible wholly for the timeouts. To me it seems the timeout logic could live entirely in the voter and things would be simpler. You don't need timers per-round since you're only focusing on timing out a single round at a time anyway.

consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -528,560 +201,282 @@ impl<
...
+        if let Some(journal) = self.journal.as_mut() {
+            journal.sync(view).await.expect("unable to sync journal");
+        }
+    }
+
+    /// Emit `msg` to every peer while recording the associated outbound metric.
+    async fn broadcast_all<T: Sender>(
+        &mut self,
+        sender: &mut WrappedSender<T, Voter<S, D>>,
+        metric: &'static metrics::Outbound,
```
BrendanChou: In my opinion you should remove this as a parameter and use a `match` in this function (based on the `msg` type) to determine the `metric`. This makes things easier at the call site and prevents the caller from accidentally using the wrong metric

consensus/src/simplex/actors/voter/round.rs
```diff
@@ -0,0 +1,858 @@
...
+        }
+        self.clear_deadlines();
+        self.nullification = Some(nullification);
+        true
+    }
+
+    pub fn add_verified_finalization(
+        &mut self,
+        finalization: Finalization<S, D>,
+    ) -> (bool, Option<S::PublicKey>) {
```
BrendanChou, line 409: Remove the `bool`, it's just equal to `option.is_some()`

In fact, I might recommend `Result<(), S::PublicKey>` instead since this way, the linter will warn you that you're throwing away the equivocation information if you call `add_verified_finalization(...)` without checking the result (without, for example, at least using `let _ =` to show that you don't care about the result)

consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -1221,311 +594,360 @@ impl<
...
+            }
+            None => {
+                timer.cancel();
+                None
+            }
         }
-        if !round.broadcast_notarize {
-            // Ensure we notarize before we finalize
-            return None;
+    }
```
BrendanChou: minor suggestion here and in the other `construct_*` methods

```suggestion
    fn construct_finalization(&mut self, view: u64, force: bool) -> Option<Finalization<S, D>> {
        let mut timer = self.recover_latency.timer();
        let candidate = self.state.finalization_candidate(view, force) {
        if let Some((true, _)) = candidate {
            timer.observe();
        } else {
            timer.cancel();
        }
        candidate.map(|(_, finalization)| finalization)
    }
```

consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -1221,311 +594,360 @@ impl<
...
+            warn!(?equivocator, "blocking equivocator");
+            self.blocker.block(equivocator).await;
         }
-        if !round.broadcast_notarization {
-            // Ensure we broadcast notarization before we finalize
-            return None;
+    }
+
+    /// Attempt to retransmit whichever certificate allowed us to enter `view + 1`.
+    async fn rebroadcast_entry_certificates<Sr: Sender>(
```
BrendanChou: consider having this function:
- take in the current view rather than the previous view
- determine whether view is the GENESIS_VIEW or not

...rather than having the caller determine that information

consensus/src/simplex/actors/voter/round.rs
patrick-ogrady: The bool indicates the finalization was set (the equivocator may or may not be none)

consensus/src/simplex/actors/voter/actor.rs
patrick-ogrady: I think you're right.

consensus/src/simplex/actors/voter/actor.rs
patrick-ogrady: Agree it is a bit weird to be straddling both...working on it.

Tough balance is making it so that `state.rs` doesn't just become `voter` lol

consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -528,560 +200,221 @@ impl<
...
-            if round.broadcast_notarization {
-                return Action::Skip;
-            }
+        if self.state.has_broadcast_notarization(view) {
+            return Action::Skip;
         }
 
         // Verify notarization
-        if !notarization.verify(&mut self.context, &self.scheme, &self.namespace) {
+        if !notarization.verify(&mut self.context, self.state.scheme(), &self.namespace) {
```
patrick-ogrady: Its odd that certificate construction is in `state` but verify is here (also odd that creating votes is also here).

It seems we could move `notarization/nullification/finalization` into `state` and just return an action. This, however, would require the state to know context (as this requires rng).

patrick-ogrady: Once `batcher` provides certs, much of this complexity (other than signing) can be avoided.

consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -194,19 +165,131 @@ impl<
...
+                "pruned view"
+            );
+        }
+        if let Some(journal) = self.journal.as_mut() {
+            journal
+                .prune(self.state.min_active())
+                .await
+                .expect("unable to prune journal");
+        }
+        let _ = self.tracked_views.try_set(self.state.tracked_views());
```
BrendanChou: should the `tracked_views` metric should just live in state like `recover_latency` does?

consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -194,19 +165,131 @@ impl<
...
+    }
+
+    async fn block_equivocator(&mut self, equivocator: Option<S::PublicKey>) {
+        if let Some(equivocator) = equivocator {
+            warn!(?equivocator, "blocking equivocator");
+            self.blocker.block(equivocator).await;
+        }
+    }
+
+    /// Attempt to retransmit whichever certificate allowed us to enter `view + 1`.
```
BrendanChou: nit: `current_view` rather than `view + 1` (not technically incorrect, but `current_view` is the parameter of the function)

consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -194,19 +165,131 @@ impl<
...
+    async fn rebroadcast_entry_certificates<Sr: Sender>(
+        &mut self,
+        recovered_sender: &mut WrappedSender<Sr, Voter<S, D>>,
+        current_view: View,
+    ) -> bool {
+        // Offset view
+        let Some(view) = current_view.checked_sub(1) else {
+            return false;
+        };
```
BrendanChou: I think you're now missing a check that `view` is not `GENESIS_VIEW`

consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -239,25 +322,23 @@ impl<
...
         // this view so slow peers do not stall our next round.
-        let past_view = current_view - 1;
         if was_retry
-            && past_view > 0
             && !self
-                .rebroadcast_entry_certificates(recovered_sender, past_view)
+                .rebroadcast_entry_certificates(recovered_sender, current_view)
                 .await
         {
             warn!(
```
BrendanChou: I think it would simplify the code flow to simply do this `warn` inside of `rebroadcast_entry_certificates`. To keep the behavior the same, don't print the warning unless you get to the bottom `false` of the function. That is, don't print it if the current view is <= 1

consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -1275,10 +1052,12 @@ impl<
...
                     debug!(view, ?leader, "skipping leader timeout due to inactivity");
                     self.skipped_views.inc();
-                    let now = self.context.current();
-                    self.state.set_leader_deadline(current_view, now, Some(now));
+                    self.state.expire_round(current_view);
                 }
             }
+
+            // Update the current view
+            let _ = self.current_view.try_set(current_view);
```
BrendanChou: also wonder if state should just manage this metric

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -118,156 +155,276 @@ impl<S: Scheme, D: Digest> State<S, D> {
...
+    pub fn next_timeout_deadline(&mut self) -> SystemTime {
+        let now = self.context.current();
+        let nullify_retry = self.nullify_retry;
+        let round = self.create_round(self.view);
+        round.next_timeout_deadline(now, nullify_retry)
     }
 
-    pub fn handle_timeout(&mut self, view: View, now: SystemTime) -> bool {
-        self.create_round(view, now).handle_timeout()
+    pub fn handle_timeout(&mut self, view: View) -> (bool, Option<Nullify<S>>) {
```
BrendanChou: would be helpful to have comments on these functions to tell what it means if the nullify is Some or None. I'm guessing it's only none if `self.scheme` is not a signer?

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -0,0 +1,1444 @@
...
+        }
+        // Walk backwards from the previous view until we reach the parent, ensuring
+        // every skipped view is nullified and the parent is notarized.
+        let mut cursor = current_view - 1;
+        loop {
+            if cursor == proposal.parent {
+                if cursor == GENESIS_VIEW {
+                    return Some(self.genesis.unwrap());
+                }
+                let payload = self.notarized_payload(cursor).copied()?;
```
andresilva:
```suggestion
                let payload = self
                    .notarized_payload(cursor)
                    .or_else(|| self.finalized_payload(cursor))
                    .copied()?;
```


If a finalization certificate reaches us before the notarization we progress to the next view but can't recover the parent digest, we'll stall the next view until the notarization is fetched.

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -0,0 +1,1444 @@
...
+            return None;
+        }
+        let proposal = round.proposal()?;
+        let parent = proposal.parent;
+        let mut missing = MissingCertificates {
+            parent,
+            notarizations: Vec::new(),
+            nullifications: Vec::new(),
+        };
+        if parent != GENESIS_VIEW && self.notarized_payload(parent).is_none() {
```
andresilva:
```suggestion
        if parent != GENESIS_VIEW
            && self.notarized_payload(parent).is_none()
            && self.finalized_payload(parent).is_none()
        {
```

If we have a finalization that should be enough?

consensus/src/simplex/actors/voter/actor.rs
patrick-ogrady: I think you're right

patrick-ogrady: moving some metrics into `state` was a late-stage change...haven't properly updated actor.

consensus/src/simplex/actors/voter/actor.rs
patrick-ogrady: yeah, I think you're right

consensus/src/simplex/actors/voter/state.rs
patrick-ogrady: Yeah, I think that's right

consensus/src/simplex/actors/voter/state.rs
patrick-ogrady: ooo, nice

patrick-ogrady: I feel there should just be a `payload()` for this?

consensus/src/simplex/actors/voter/actor.rs
patrick-ogrady: just removed `rebroadcast_entry_certificates` and placed this logic into `handle_timeout`

consensus/src/simplex/actors/voter/slot.rs
```diff
@@ -0,0 +1,359 @@
...
+                if recovered {
+                    // If we receive a certificate for a conflicting proposal, we replace the
+                    // local proposal (may just be from a vote)
+                    self.proposal = Some(retained.clone());
+                    self.requested_build = true;
+                    self.requested_verify = true;
+                } else {
+                    // If this isn't a certificate, we keep the proposal as-is
+                    retained = existing.clone();
+                    dropped = proposal.clone();
```
BrendanChou: [Note]

Are you able to do something like `(retained, dropped) = (dropped, retained);` instead of cloning both again?

consensus/src/simplex/actors/voter/round.rs
```diff
@@ -0,0 +1,845 @@
...
+
+    /// Clears the outstanding parent gap watermark once the data arrives.
+    pub fn clear_parent_missing(&mut self) {
+        self.proposal.clear_parent_missing();
+    }
+
+    #[allow(clippy::type_complexity)]
+    /// Returns the leader key and proposal when the view is ready for verification.
+    pub fn should_verify(&self) -> Option<(Leader<S::PublicKey>, Proposal<D>)> {
+        let leader = self.leader.clone()?;
```
BrendanChou: [Note]

tiny optimization: I think do `let leader = self.leader?;` and then only clone at the end of the function: `Some((leader.clone(), proposal))` as long as this doesn't introduce any borrowing issues

consensus/src/simplex/actors/voter/round.rs
```diff
@@ -0,0 +1,845 @@
...
+    }
+
+    /// Returns how many finalize votes we currently track.
+    pub fn len_finalizes(&self) -> usize {
+        self.votes.len_finalizes()
+    }
+
+    /// Returns how much time elapsed since the round started, if the clock monotonicity holds.
+    pub fn elapsed_since_start(&self, now: SystemTime) -> Option<Duration> {
+        now.duration_since(self.start).ok()
```
BrendanChou: [Note]

kinda feels like `unwrap_or_default` is fine to remove the possibility of `None`. It doesn't end up making the code much cleaner since the `state.elapsed_since_start()` function that calls this still has to return an `Option` (if the round doesn't exist in state) but I think we "should" record the duration as zero if the clock monotonicity doesn't hold

consensus/src/simplex/actors/voter/round.rs
```diff
@@ -0,0 +1,845 @@
...
+    pub fn set_nullify_retry(&mut self, when: Option<SystemTime>) {
+        self.nullify_retry = when;
+    }
+
+    /// Handles a timeout event, marking that we've broadcast a nullify vote.
+    ///
+    /// Returns `true` if this is a retry (we've already broadcast nullify before),
+    /// `false` if this is the first timeout for this round.
+    pub fn handle_timeout(&mut self) -> bool {
+        let retry = replace(&mut self.broadcast_nullify, true);
```
BrendanChou, line 268: [Meme]

fancy_pooh.jpeg

consensus/src/simplex/actors/voter/round.rs
```diff
@@ -0,0 +1,845 @@
...
+                    ?retained,
+                    "finalize conflicts with certificate proposal (equivocation detected)"
+                );
+                return equivocator;
+            }
+            ProposalChange::Skipped => return None,
+        }
+        self.votes.insert_finalize(finalize);
+        None
+    }
```
BrendanChou, line 367: [Clarification/Major]

I'm confused why a single `finalize` or `notarize` vote would cause you to call `self.proposal.update()`? How can you know that such a vote has the correct proposal?

In my mind `self.proposal.update()` should only be called by `add_verified_notarization()` or `add_verified_finalization()` which both do so via `add_recovered_proposal()`

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -0,0 +1,1526 @@
...
+        let mut timer = self.recover_latency.timer();
+        let notarization = self
+            .views
+            .get_mut(&view)
+            .and_then(|round| round.notarizable());
+        if notarization.is_some() {
+            timer.observe();
+        } else {
+            timer.cancel();
+        }
```
BrendanChou, line 378: [Note]

The timer will automatically be observed when dropped as long as it is not canceled. So it's sufficient to simply call `cancel` if `notarization.is_none()`.

This gives me an idea for a possible `cancel_if` function we could add: `timer.cancel_if(notarization.is_none());` but maybe not

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -0,0 +1,1526 @@
...
+    pub fn prune(&mut self) -> Vec<View> {
+        let min = self.min_active();
+        let mut removed = Vec::new();
+        while let Some(view) = self.views.keys().next().copied() {
+            if view >= min {
+                break;
+            }
+            self.views.remove(&view);
+            removed.push(view);
+        }
```
BrendanChou: [Note]

consider using `BTreeMap::split_off`

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -0,0 +1,1526 @@
...
+
+        // Update metrics
+        let _ = self.tracked_views.try_set(self.views.len());
+        removed
+    }
+
+    /// Returns the certified payload for a view, checking certificates first,
+    /// then falling back to checking if we have enough votes to certify.
+    fn certified_payload(&self, view: View) -> Option<&D> {
```
BrendanChou: [Minor]

"certify" is the term I would like to use in the future (when calling out to the Automaton after getting a notarization). I would prefer not "pre-overloading" the term at this point

consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -1895,7 +898,7 @@ impl<
                     };
 
                     // Block if the epoch is not the current epoch
-                    if msg.epoch() != self.epoch {
```
BrendanChou, line 1898: [Minor]

accidentally undid this indent, not sure why linting doesn't fix

consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -1906,24 +909,34 @@ impl<
                     // We opt to not filter by `interesting()` here because each message type has a different
                     // configuration for handling `future` messages.
                     view = msg.view();
-                    let action = match msg {
+                    let action;
+                    match msg {
```
BrendanChou, line 911: [Note]

nit: if you want you can still use `let action = match msg {` if you simply do below:

```
Voter::Notarize(_) | Voter::Nullify(_) | Voter::Finalize(_) => Action::Block,
```

Though I guess it may slightly change the `"blocking peer"` message to be generic instead of specifically calling it out for an invalid message type

consensus/src/simplex/actors/voter/round.rs
patrick-ogrady: shaving off microseconds

consensus/src/simplex/actors/voter/round.rs
patrick-ogrady: Ah, that's what the boolean arg is for.

If true (certificate), it is considered "definitive". If false, it can be used as evidence a leader equivocated (we must've seen a leader proposal to get a vote from batcher).

This will be much cleaner after #2228

consensus/src/simplex/actors/voter/state.rs
patrick-ogrady: ah, forgot about drop semantics. I prefer the explicitness (otherwise you need to ponder where the drop might be)

consensus/src/simplex/actors/voter/state.rs
patrick-ogrady: will ponder

consensus/src/simplex/actors/voter/actor.rs
patrick-ogrady: hmm

consensus/src/simplex/actors/voter/slot.rs
patrick-ogrady: noice

consensus/src/simplex/actors/voter/round.rs
patrick-ogrady: makes sense

consensus/src/simplex/actors/voter/actor.rs
patrick-ogrady: I think our select messes this up

consensus/src/simplex/actors/voter/actor.rs
patrick-ogrady: Looks ugly because we have to handle action in the certificates.

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -602,23 +603,17 @@ impl<E: Clock + Rng + CryptoRng + Metrics, S: Scheme, D: Digest> State<E, S, D>
...
-        let mut removed = Vec::new();
-        while let Some(view) = self.views.keys().next().copied() {
-            if view >= min {
-                break;
-            }
-            self.views.remove(&view);
-            removed.push(view);
-        }
+        let kept = self.views.split_off(&min);
+        let removed = replace(&mut self.views, kept).into_keys().collect();
```
BrendanChou, line 607: fancy_pooh.gif

PR #2200, opened 2025-11-11
 fix: Fetcher::retain doc to reflect predicate-based API

MozirDmitriy:
The previous comment incorrectly stated that retain keeps “keys greater than the given key,” but the method accepts an arbitrary predicate.


clabby:

PR #2144, opened 2025-11-05
[consensus/marshal] Enshrine Finalizer Wait

patrick-ogrady:
Fixes: #2135

Instead of asking the application to block on `Update::Block` and not on `Update::Tip`, we enshrine this "acknowledged delivery" behavior with a `oneshot::Sender`.

PR #2202, opened 2025-11-11
[consensus/marshal] specific reporters type to send to multiple consumers

SuperFluffy:
This patch provides a marshal-specific `Reporters` type to allow sending the `Update<B>` enum to more than one consumer.

https://github.com/commonwarexyz/monorepo/pull/2144 enshrined finalizer await by including a oneshot channel in the `marshal::Update<B>` enum, which is a good change. However, it broke the `consensus::Reporter for Reporters<R1, R2>` impl by requiring that the `Activity: Clone`. https://github.com/commonwarexyz/monorepo/blob/fe0dcf4bab6ab5a944d43f3675a6346cbed07862/consensus/src/reporter.rs#L31-L36.

This was likely not caught because no commonware example uses `consensus::Reporters` in the marshal agent.

### DIFF
--- a/consensus/src/marshal/mod.rs
+++ b/consensus/src/marshal/mod.rs
@@ -65,7 +65,7 @@ pub mod resolver;
 use crate::{simplex::signing_scheme::Scheme, types::Epoch, Block, Reporter};
 use futures::{
     channel::oneshot,
-    future::{self, join, Either},
+    future::{join, try_join},
 };
 use std::sync::Arc;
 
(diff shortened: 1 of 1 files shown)


clabby (commented):
Looks good. Mind adding a short test for 2 & 3 reporters (the latter nesting the new struct: `Reporters<A, Reporter, Reporters<A, Reporter, Reporter>>`)?

SuperFluffy:
> Looks good. Mind adding a short test for 2 & 3 reporters (the latter nesting the new struct: `Reporters<A, Reporter, Reporters<A, Reporter, Reporter>>`)?

Added some very simple tests checking receipt of acks and blocks.

consensus/src/marshal/actor.rs
```diff
@@ -754,11 +768,18 @@ impl<
 
         let (height, commitment) = (block.height(), block.commitment());
         let (ack_tx, ack_rx) = oneshot::channel();
-        application.report(Update::Block(block, ack_tx)).await;
-        self.pending_ack.replace(PendingAck {
+        let application = application.clone();
```
clabby: note - We already have the assumption that `application` is a cheap clone (either a mailbox or a shallow struct for `Marshaled`.) This should be low-cost.

consensus/src/marshal/mod.rs
```diff
@@ -62,10 +62,78 @@ pub mod ingress;
...
+    channel::oneshot,
+    future::{self, join, Either},
+};
 use std::sync::Arc;
 
+/// A marshal-specific reporter to report finalizer activity to more than one reporter.
+///
+/// Use this type if to send marshal [`Update`] to more than one consumer.
+///
+/// The reason this exists in addition [`crate::reporter::Reporters`], is that
```
patrick-ogrady, line 76: I considered making a new type for this to avoid a custom reporter but it becomes very difficult to actually know when it is safe to "listen" on this type (as one actor may finish before the "sender" is cloned).

It seems we need to be explicit in the number of recipients to ensure we properly wait for everything.

consensus/src/marshal/mod.rs
```diff
@@ -1135,4 +1206,167 @@ mod tests {
...
+    }
+    impl Reporter for Mailbox {
+        type Activity = Update<B>;
+
+        async fn report(&mut self, activity: Self::Activity) {
+            self.inner.send(activity).await.unwrap()
+        }
+    }
+
+    #[test_async]
```
patrick-ogrady, line 1223: forgot about this lol...fml

consensus/src/marshal/actor.rs
```diff
@@ -754,11 +768,18 @@ impl<
...
+        pending_ack.replace(PendingAck {
             height,
             commitment,
-            receiver: ack_rx,
+            receiver: async move {
+                application
+                    .clone()
```
patrick-ogrady: This second clone is unnecessary (pushed a fix)

consensus/src/marshal/actor.rs
```diff
@@ -754,11 +768,17 @@ impl<
...
-        application.report(Update::Block(block, ack_tx)).await;
-        self.pending_ack.replace(PendingAck {
+        pending_ack.replace(PendingAck {
             height,
             commitment,
-            receiver: ack_rx,
+            receiver: {
+                let mut application = application.clone();
+                async move {
+                    application.report(Update::Block(block, ack_tx)).await;
```
patrick-ogrady, line 777: I don't like that we call report in async here but not for tip. It implies a very specific function of report() that to me indicates of a deeper issue.

Issue #1632, opened 2025-09-16
[storage/index] support index sharding


roberto-bayardo:
The storage index currently puts everything in a single shard, so the bytes/key is at best the number of bytes used by the translator (~8-9 for large DBs) plus the number of bytes required by the value (typically 8 for a location index). We could support sharding the index using the first "n" (where n is ~1-3) bytes of the key to remove n bytes from the average bytes/key.

roberto-bayardo:
Sharded index is now implemented by the above PRs.  Leaving this issue open until we create a concrete qmdb instance leveraging it.

PR #2206, opened 2025-11-12
[storage/index] ordered variant of partitioned index

roberto-bayardo:
Implementation of the ordered variant of a partitioned index, with tests + comparison to other variants by adding it to the index insertion benchmark.

context: https://github.com/commonwarexyz/monorepo/issues/1632

### DIFF
--- a/storage/src/index/mod.rs
+++ b/storage/src/index/mod.rs
@@ -1325,7 +1325,7 @@ mod tests {
         let runner = deterministic::Runner::default();
         runner.start(|context| async move {
             {
-                let mut index = new_partitioned_unordered(context);
+                let mut index = new_partitioned_unordered(context.clone());
                 run_index_remove_middle_then_next(&mut index);
             }
             {


storage/src/index/partitioned/ordered.rs
```diff
@@ -0,0 +1,417 @@
...
+            }
+        }
+
+        for partition in self.partitions[..partition_index].iter().rev() {
+            let mut iter = partition.last_translated_key().peekable();
+            if iter.peek().is_some() {
+                return Box::new(iter) as Box<dyn Iterator<Item = &'a Self::Value> + 'a>;
+            }
+        }
+
```
danlaine, line 173: Do we want to have wrap-around logic? For example, if we have partitions 0, 1 and 2, and each has 1 element, then what should happen if we try to get the `prev_translated_key` of the (only) key in 1?

storage/src/index/partitioned/ordered.rs
```diff
@@ -0,0 +1,417 @@
...
+        for partition in self.partitions.iter().skip(partition_index + 1) {
+            let mut iter = partition.first_translated_key().peekable();
+            if iter.peek().is_some() {
+                return Box::new(iter) as Box<dyn Iterator<Item = &'a Self::Value> + 'a>;
+            }
+        }
+
```
danlaine, line 198: Same question as above about wrap-around logic

storage/src/index/partitioned/mod.rs
```diff
@@ -14,4 +14,120 @@
...
+pub mod ordered;
 pub mod unordered;
+
+// Because the prefix length has a max of 3, we can safely use a 4-byte int for the index type
+// used by prefix conversion.
+const INDEX_INT_SIZE: usize = 4;
+
+/// Get the partition index for the given key, along with the prefix-stripped key for probing
+/// the referenced partition. The returned index value is in the range `[0, 2^(P*8) - 1]`.
+fn partition_index_and_sub_key<const P: usize>(key: &[u8]) -> (usize, &[u8]) {
```
danlaine, line 26: This is sort of implementing a `Translator`, right? I wonder whether we should implement a `Translator` to hold this logic.

storage/src/index/partitioned/ordered.rs
roberto-bayardo: The wrap around is being handled at the app layer. If we do it here I think we'd need to return an indicator of whether a wrap-around took place.

storage/src/index/partitioned/mod.rs
roberto-bayardo: Yes basically. But since it's so specialized I thought it was best to just implement it this way and avoid some unecessary recomputation that would happen if we were to invoke 2 translators per lookup.

storage/src/index/mod.rs
```diff
@@ -1393,8 +1555,14 @@ mod tests {
...
         let runner = deterministic::Runner::default();
         runner.start(|context| async move {
-            let mut index = new_partitioned(context);
-            run_index_cursor_insert_before_next(&mut index);
+            {
+                let mut index = new_partitioned_unordered(context.clone());
+                run_index_cursor_insert_before_next(&mut index);
+            }
+            {
+                let mut index = new_partitioned_ordered(context.clone());
```
patrick-ogrady: We don't need clone on the context for the second test here?

storage/src/index/partitioned/ordered.rs
```diff
@@ -0,0 +1,417 @@
...
+        Self {
+            partitions,
+            _phantom: PhantomData,
+        }
+    }
+
+    /// Get the partition for the given key, along with the prefix-stripped key for probing it.
+    fn get_partition<'a>(&self, key: &'a [u8]) -> (&I, &'a [u8]) {
+        let (i, sub_key) = super::partition_index_and_sub_key::<P>(key);
```
patrick-ogrady: nit: I think we can just import this function (and not use `super`)?

storage/src/index/partitioned/unordered.rs
```diff
@@ -35,33 +31,17 @@ impl<T: Translator, I: Unordered<T>, const P: usize> Index<T, I, P> {
...
+        let (i, sub_key) = super::partition_index_and_sub_key::<P>(key);
 
         (&self.partitions[i], sub_key)
     }
 
     /// Get the mutable partition for the given key, along with the prefix-stripped key for probing
     /// it.
     fn get_partition_mut<'a>(&mut self, key: &'a [u8]) -> (&mut I, &'a [u8]) {
-        let (i, sub_key) = Self::partition_index_and_sub_key(key);
+        let (i, sub_key) = super::partition_index_and_sub_key::<P>(key);
```
patrick-ogrady: see question about just importing this function at top of file.

.github/workflows/slow.yml
```diff
@@ -158,7 +158,7 @@ jobs:
     - name: Run miri tests for storage::index
       run: |
         cd storage
-        just miri index:: --partition hash:${{matrix.partition}}/8
+        just miri index::tests --partition hash:${{matrix.partition}}/8
```
patrick-ogrady, line 161: this results in basically just running the `mod` tests?

storage/src/index/mod.rs
roberto-bayardo: changed

storage/src/index/partitioned/ordered.rs
roberto-bayardo: done

storage/src/index/partitioned/unordered.rs
roberto-bayardo: done

.github/workflows/slow.yml
roberto-bayardo: Yeah, that's where the Cursor tests are (the ones involving "unsafe") that are invoked across all variants.

PR #2209, opened 2025-11-12
[Storage] Move authenticated bitmap to `storage::AuthenticatedBitMap`

danlaine:
It doesn't feel right for this data structure to be in `mmr`. It also can't go in the utils crate because it'd make the utils crate depend on the storage crate. I think this is a reasonable way for storage crate users to import this data structure.

### DIFF
--- a/storage/fuzz/fuzz_targets/mmr_bitmap.rs
+++ b/storage/fuzz/fuzz_targets/mmr_bitmap.rs
@@ -3,7 +3,7 @@
 use arbitrary::Arbitrary;
 use commonware_cryptography::{sha256, Sha256};
 use commonware_runtime::{deterministic, Runner};
-use commonware_storage::AuthenticatedBitmap as BitMap;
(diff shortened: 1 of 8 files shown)


storage/src/bitmap/authenticated.rs
```diff
@@ -38,7 +38,7 @@ use tracing::{debug, error, warn};
 ///
 /// Even though we use u64 identifiers for bits, on 32-bit machines, the maximum addressable bit is
 /// limited to (u32::MAX * N * 8).
-pub struct BitMap<D: Digest, const N: usize> {
+pub struct Authenticated<D: Digest, const N: usize> {
```
roberto-bayardo: Why not just keep it named Bitmap? ("authenticated" is already in the path).

danlaine: Changed back to `BitMap`

PR #1881, opened 2025-10-13
DKG Refactor

cronokirby:
Closes #1861
Closes #1840 
Closes #1996 

The crux of this change is in `dkg.rs` which provides an API which handles the cryptographic details for everything except providing authenticated channels, which are just a baseline used everywhere, through p2p. The resulting API is more abstract, but should guide people towards correct implementations much more easily.

The massive size of this PR (sorry peeps 😅) comes from refactoring code to take advantage of this. In particular, I rewrote the DKG-related portion of the reshare example, in order to:

- remove cryptographic operations now performed inside of `dkg.rs`,
- modify the abstractions to better represent the modelling being `dealer`, `player` (`observer`), having removed the notion of `arbiter`,
- cleanup the actor logic in general.

I also took the liberty of rewriting the tests there to try and avoid duplicated logic (this made my own refactoring easier).

For review I think it would be good to focus on us being happy with `dkg.rs`, and then looking at the rest.

PR #2214, opened 2025-11-12
[cryptography] derive Clone for DKG objects

SuperFluffy:
There seems to be no downside to making `dkg::{Arbiter, Player, Dealer}: Clone`. I need this for my use case to do "speculative" finalization of the DKG ceremony: I need to write the DKG ceremony on the last block of an epoch, but since DKG acts on finalized blocks but proposals on notarized block, there is no guarantee that the outcome of the DKG process is ready in time.

Cloning the DKG objects allows me to finalize on propose and verify and - in case of mismatch - keep processing finalized blocks.

This is likely a stopgap solution until a more efficient method is implemented.

### DIFF
--- a/cryptography/src/bls12381/dkg/arbiter.rs
+++ b/cryptography/src/bls12381/dkg/arbiter.rs
@@ -65,6 +65,7 @@ pub struct Output<V: Variant> {
 }
 
 /// Gather commitments, acknowledgements, and reveals from all dealers.
+#[derive(Clone)]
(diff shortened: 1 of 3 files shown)


SuperFluffy:
~~Making this a draft again. This design is not very nice. Feel free to close it.~~

Wrong PR to make this comment.

patrick-ogrady:
This is a minimal workaround until #1881 is in (which should make doing something like this much simpler) 👍

PR #2216, opened 2025-11-13
[utils] Introduce `Acknowledgement`

patrick-ogrady:
Replaces: #2202 
Replaces: #2212

Instead of adding a custom `Reporter` impl, I think the most maintainable approach is to treat `Acknowledgement` as a first-party action.

### DIFF
--- a/utils/src/acknowledgement.rs
+++ b/utils/src/acknowledgement.rs
@@ -24,6 +24,7 @@ pub struct Canceled;
 pub trait Acknowledgement: Clone + Send + Sync + Debug + 'static {
     /// Future resolved once the acknowledgement is handled.
     type Waiter: Future<Output = Result<(), Self::Error>> + Send + Sync + Unpin + 'static;
+
     /// Error produced if the acknowledgement is not handled.
     type Error: Debug + Send + Sync + 'static;
 


consensus/src/marshal/actor.rs
```diff
@@ -129,7 +132,7 @@ pub struct Actor<
     // Last height processed by the application
     last_processed_height: u64,
     // Pending application acknowledgement, if any
-    pending_ack: OptionFuture<PendingAck<B>>,
+    pending_ack: OptionFuture<PendingAck<B, A>>,
```
patrick-ogrady, line 135: I originally considered migrating to `channels::tracked` (so we could enqueue multiple blocks at once) but opted to take a simpler approach at first to unblock.

consensus/src/marshal/mod.rs
```diff
@@ -79,17 +79,18 @@ pub trait SchemeProvider: Clone + Send + Sync + 'static {
 ///
 /// Finalized tips are reported as soon as known, whether or not we hold all blocks up to that height.
 /// Finalized blocks are reported to the application in monotonically increasing order (no gaps permitted).
-#[derive(Debug)]
-pub enum Update<B: Block> {
+#[derive(Clone, Debug)]
```
patrick-ogrady, line 82: This will allow usage in the existing reporter helpers (and permits doing so without blocking on report)

utils/src/acknowledgement.rs
```diff
@@ -0,0 +1,158 @@
...
+    /// Fulfill the acknowledgement.
+    fn acknowledge(self);
+}
+
+/// Default acknowledgement implementation that requires a minimum number of acknowledgements.
+///
+/// If any acknowledgement is not handled (before the minimum number of acknowledgements is received), the acknowledgement will be cancelled.
+pub struct Min<const N: usize = 1> {
+    state: Arc<AckState<N>>,
+    acknowledged: AtomicBool,
```
patrick-ogrady: I don't think this needs to be an atomic?

consensus/src/marshal/mod.rs
```diff
@@ -79,17 +79,18 @@ pub trait SchemeProvider: Clone + Send + Sync + 'static {
...
     Tip(u64, B::Commitment),
-    /// A new finalized block and a channel to acknowledge the update.
+    /// A new finalized block and an acknowledgement handle for the application to signal once processed.
     ///
-    /// To ensure all blocks are delivered at least once, marshal waits to mark
-    /// a block as delivered until the application explicitly acknowledges the update.
+    /// To ensure all blocks are delivered at least once, marshal waits to mark a block as delivered
+    /// until the application explicitly acknowledges the update. The acknowledgement is clonable,
+    /// allowing the application to share it across async tasks if needed.
     /// If the sender is dropped before acknowledgement, marshal will exit (assuming
```
patrick-ogrady: Fix description

utils/src/acknowledgement.rs
```diff
@@ -0,0 +1,160 @@
...
+
+    fn acknowledge(mut self) {
+        self.state.acknowledge();
+        self.acknowledged = true;
+    }
+}
+
+/// State for the [Min] acknowledgement.
+struct AckState<const N: usize> {
+    sender: Mutex<Option<oneshot::Sender<()>>>,
```
clabby: Let's use a `Mutex` that can yield here (https://docs.rs/futures/latest/futures/lock/struct.Mutex.html).

Tests pass because the deterministic runtime doesn't have a concurrent scheduler, but this will cause issues on the tokio runtime if `acknowledge` is called within a future that's currently in the same thread as the runtime's scheduler.

`AckState::acknowledge` & `Acknowledgement::acknowledge` should also be colored async to facilitate this.

utils/src/acknowledgement.rs
```diff
@@ -0,0 +1,160 @@
...
+
+    /// Acknowledge the completion of a task.
+    fn acknowledge(&self) {
+        let mut current = self.remaining.load(Ordering::Acquire);
+        while current != 0 {
+            match self.remaining.compare_exchange(
+                current,
+                current - 1,
+                Ordering::AcqRel,
+                Ordering::Acquire,
```
clabby: `Ordering`s LGTM

patrick-ogrady: Just updated this to avoid the loop lol 😓

utils/src/acknowledgement.rs
patrick-ogrady: oneshot sender isn't async, so I didn't feel it was necessary?

utils/src/acknowledgement.rs
patrick-ogrady: (ordering was maintained, so resolving)

utils/src/acknowledgement.rs
patrick-ogrady: >the thread can get pre-empted while waiting but not the "task"

patrick-ogrady: We, unfortunately, need to call `cancel()` in `drop` ... so, adding async locking here would be pretty gnar.

I also believe we really shouldn't see contention on the mutex (it should only be called once unless `cancel()` -> but that should be rare).

patrick-ogrady: We could drop the `Mutex` and just use `AtomicPtr`:

```
diff --git a/utils/src/acknowledgement.rs b/utils/src/acknowledgement.rs
index 508dbcfbb..693f0be8a 100644
--- a/utils/src/acknowledgement.rs
+++ b/utils/src/acknowledgement.rs
@@ -5,8 +5,8 @@ use futures::channel::oneshot;
 use std::{
     future::Future,
     sync::{
-        atomic::{AtomicUsize, Ordering},
-        Arc, Mutex,
+        atomic::{AtomicPtr, AtomicUsize, Ordering},
+        Arc,
     },
 };
 
@@ -84,7 +84,7 @@ impl<const N: usize> Acknowledgement for Exact<N> {
 
 /// State for the [Exact] acknowledgement.
 struct AckState<const N: usize> {
-    sender: Mutex<Option<oneshot::Sender<()>>>,
+    sender: AtomicPtr<oneshot::Sender<()>>,
     remaining: AtomicUsize,
 }
 
@@ -92,11 +92,23 @@ impl<const N: usize> AckState<N> {
     /// Create a new acknowledgement state.
     fn new(sender: oneshot::Sender<()>) -> Self {
         Self {
-            sender: Mutex::new(Some(sender)),
+            sender: AtomicPtr::new(Box::into_raw(Box::new(sender))),
             remaining: AtomicUsize::new(N),
         }
     }
 
+    /// Take ownership of the sender exactly once, if it hasn't been taken yet.
+    fn take_sender(&self) -> Option<oneshot::Sender<()>> {
+        let ptr = self.sender.swap(core::ptr::null_mut(), Ordering::AcqRel);
+        if ptr.is_null() {
+            None
+        } else {
+            // SAFETY: We exclusively own the pointer after swapping it to null.
+            let boxed = unsafe { Box::from_raw(ptr) };
+            Some(*boxed)
+        }
+    }
+
     /// Acknowledge the completion of a task.
     fn acknowledge(&self) {
         // If not the last acknowledgement, do nothing.
@@ -117,14 +129,23 @@ impl<const N: usize> AckState<N> {
         }
 
         // Send the acknowledgement to the waiter.
-        if let Some(tx) = self.sender.lock().unwrap().take() {
+        if let Some(tx) = self.take_sender() {
             let _ = tx.send(());
         }
     }
 
     /// Cancel the acknowledgement.
     fn cancel(&self) {
-        if let Some(tx) = self.sender.lock().unwrap().take() {
+        if let Some(tx) = self.take_sender() {
+            drop(tx);
+        }
+    }
+}
+
+impl<const N: usize> Drop for AckState<N> {
+    fn drop(&mut self) {
+        // Ensure we don't leak if no one took the sender.
+        if let Some(tx) = self.take_sender() {
             drop(tx);
         }
     }

```

patrick-ogrady: I think we could avoid cancel contention if we used a sentinel value in the AtomicUsize for cancel.

clabby: https://github.com/commonwarexyz/monorepo/pull/2217

clabby (approved):

utils/src/acknowledgement.rs
```diff
@@ -0,0 +1,213 @@
...
+#[derive(Debug, thiserror::Error)]
+#[error("acknowledgement was cancelled")]
+pub struct Canceled;
+
+/// A mechanism for acknowledging the completion of a task.
+pub trait Acknowledgement: Clone + Send + Sync + Debug + 'static {
+    /// Future resolved once the acknowledgement is handled.
+    type Waiter: Future<Output = Result<(), Self::Error>> + Send + Sync + Unpin + 'static;
+    /// Error produced if the acknowledgement is not handled.
+    type Error: Debug + Send + Sync + 'static;
```
SuperFluffy, line 29: Does it need to be `Sync`?

utils/src/acknowledgement.rs
```diff
@@ -0,0 +1,213 @@
...
+    fn handle() -> (Self, Self::Waiter);
+
+    /// Fulfill the acknowledgement.
+    fn acknowledge(self);
+}
+
+/// [Acknowledgement] that returns once exactly `N` acknowledgements are received.
+///
+/// If any acknowledgement is not handled, the acknowledgement will be cancelled.
+pub struct Exact<const N: usize = 1> {
```
SuperFluffy: If we don't enshrine in the type but simply increment on clone (like an arc/rc), this would be more ergonomic.

EDIT: I also believe this will have issues with optional reporters in `Reporters`.

utils/src/acknowledgement.rs
```diff
@@ -0,0 +1,213 @@
...
+    fn acknowledge(&self) {
+        // If not the last acknowledgement, do nothing.
+        match self
+            .remaining
+            .fetch_update(Ordering::AcqRel, Ordering::Acquire, |remaining| {
+                if remaining == 0 {
+                    None
+                } else {
+                    Some(remaining - 1)
+                }
```
SuperFluffy:
```suggestion
                remaining.checked_sub(1)
```

utils/src/acknowledgement.rs
patrick-ogrady: There is a race condition with incrementing on clone unfortunately (commented on one of your PRs).

If we don't know how many listeners there could be, all items cloned so far could acknowledge and `marshal` could move forward (although some clones haven't been made yet).

patrick-ogrady:
> EDIT: I also think this might run into troubles with the Reporters forking two one or more optional reporters.

Correct. You need to know exactly how many participants you want to acknowledge delivery a prioi (see race condition point above).

We could alternatively accept some sort of arg into `handle()` for listeners but the reporter logic (and what may or may not be listening) is opaque as you know.

SuperFluffy:
> Correct. You need to know exactly how many participants you want to acknowledge delivery a prioi (see race condition point above).
> We could alternatively accept some sort of arg into handle() for listeners but the reporter logic (and what may or may not be listening) is opaque as you know.

That's right, thanks for pointing that out.

I have pushed a possible solution to https://github.com/commonwarexyz/monorepo/pull/2218:

Instead of requiring `Activity: Clone`, we can require `Activity: Splittable`, which is just:
```rust
trait Splittable: Send {
    fn split(self) -> (Self, Self);
}
```

Modulo me overlooking some other race conditions, I think this might be an elegant way of a) hiding the cardinality of how many drops are expected from the consumer (some mailbox should not need to know how many other mailboxes need to ack the update), b) it solves the problem with the `Reporters` splitter and `Option<R>`, and c) it solves the race condition because cloning is banned and every consumer gets their "own" increment.

consensus/src/marshal/mod.rs
SuperFluffy: Propose a new `Splittable` trait in https://github.com/commonwarexyz/monorepo/pull/2218 instead of `Clone`.

patrick-ogrady:
> > Correct. You need to know exactly how many participants you want to acknowledge delivery a prioi (see race condition point above).
> > We could alternatively accept some sort of arg into handle() for listeners but the reporter logic (and what may or may not be listening) is opaque as you know.
> 
> That's right, thanks for pointing that out.
> 
> I have pushed a possible solution to #2218:
> 
> Instead of requiring `Activity: Clone`, we can require `Activity: Splittable`, which is just:
> 
> ```rust
> trait Splittable: Send {
>     fn split(self) -> (Self, Self);
> }
> ```
> 
> Modulo me overlooking some other race conditions, I think this might be an elegant way of a) hiding the cardinality of how many drops are expected from the consumer (some mailbox should not need to know how many other mailboxes need to ack the update), b) it solves the problem with the `Reporters` splitter and `Option<R>`, and c) it solves the race condition because cloning is banned and every consumer gets their "own" increment.

Figured out a way to avoid the race condition (and we can just use `Clone` 🚀 )

utils/src/acknowledgement.rs
```diff
@@ -0,0 +1,210 @@
...
+            .finish()
+    }
+}
+
+impl Clone for Exact {
+    fn clone(&self) -> Self {
+        // Because acknowledge consumes self, we know that there is no way for there
+        // to remain 0 references before the last acknowledgement has been cloned (i.e.
+        // the acknowledgement won't resolve while we are still creating new clones).
+        self.state.increment();
```
clabby, line 59: 👍

clabby (approved):
Neat :)

PR #2218, opened 2025-11-13
[utils] introduce Splittable

SuperFluffy:
Introduces a `Splittable` trait to remove the const generic on the `Exact` acknowledgment and implements the `Reporters` splitter in terms of it.

Demonstrates how the `Reporters` splitter with optional branches now passes tests where before it would have failed.

### DIFF
--- a/utils/src/acknowledgement.rs
+++ b/utils/src/acknowledgement.rs
@@ -38,7 +38,7 @@ pub trait Acknowledgement: Splittable + Send + Sync + Debug + 'static {
     fn acknowledge(self);
 }
 
-/// [Acknowledgement] that returns once as many acknowledgments returned as `Exact` was split.
+/// [`Acknowledgement`] that returns once as many acknowledgments are received as `Exact` was split.
 ///
 /// If any acknowledgement is not handled, the acknowledgement will be cancelled.
 ///


consensus/src/reporter.rs
```diff
@@ -86,3 +86,37 @@ impl<A, R1, R2> From<(R1, R2)> for Reporters<A, R1, R2> {
...
+                activity.acknowledge();
+            }
+        }
+
+        let mut split =
+            Reporters::<Exact<2>, Reporter, Reporter>::from((Some(Reporter), None::<Reporter>));
+
+        let (ack, waiter) = Exact::<2>::handle();
+        split.report(ack).await;
+        assert!(!waiter.now_or_never().unwrap().is_ok());
```
SuperFluffy: This will panic.

SuperFluffy: Introduced `Splittable` to address this problem.

examples/reshare/src/dkg/ingress.rs
```diff
@@ -48,6 +47,20 @@ where
     sender: mpsc::Sender<Message<H, C, V, A>>,
 }
 
+impl<H, C, V, A> Clone for Mailbox<H, C, V, A>
```
SuperFluffy, line 50: Necessary because `A: !Clone`, and so the derive complains.

consensus/src/simplex/types.rs
```diff
@@ -1748,6 +1748,12 @@ pub enum Activity<S: Scheme, D: Digest> {
     NullifyFinalize(NullifyFinalize<S, D>),
 }
 
+impl<S: Scheme, D: Digest> Splittable for Activity<S, D> {
```
patrick-ogrady, line 1751: Seems we could derive Splittable for anything that implements Clone?

consensus/src/reporter.rs
```diff
@@ -20,17 +21,30 @@ impl<A: Send, R: Reporter<Activity = A>> Reporter for Option<R> {
...
 
     _phantom: PhantomData<A>,
 }
 
+impl<A, R1, R2> Clone for Reporters<A, R1, R2>
+where
+    R1: Clone,
+    R2: Clone,
+{
+    fn clone(&self) -> Self {
```
patrick-ogrady, line 36: Suspect this complains for the same reason you mentioned below around A not being clone?

patrick-ogrady (commented):
Interesting approach...would be VERY nice to not have to specify an `N`.

consensus/src/simplex/types.rs
SuperFluffy: 99% sure we can't because Rust does not have specialization and you run into conflicting trait impls:

If `T: Clone` and also `T: Splittable`, then which impl do you choose? The blanket impl or the specific impl? Rust blocks you from that even if `T` does not yet impl `Clone`.

consensus/src/reporter.rs
SuperFluffy: Yeah. Derive macros essentially do something like `Clone for MyType<A> where A: Clone`, but `A: !Clone` almost always.

Another one of those things that'd be nice to have: https://github.com/rust-lang/lang-team/issues/152

consensus/src/simplex/types.rs
SuperFluffy: https://std-dev-guide.rust-lang.org/policy/specialization.html

patrick-ogrady: Ah, I was thinking keep `Splittable` but just implement `Splittable` for anything that implements `Clone` in `utils` (no specialization should be required)...will try it myself (admit could be off lol)

SuperFluffy: Ah, yeah. Probably. I impl'd it for the simplex `Activity`, because that came to mind. Maybe there are other objects that you frequently "report" on and that might be consumed by more than one reporter?

Bikeshed: technically, isn't the marshal actor the reporter and the consumer the reportee? ^.^

PR #2226, opened 2025-11-13
[consensus/simplex] Restore Comments and Logs

patrick-ogrady:


consensus/src/simplex/actors/voter/state.rs
```diff
@@ -196,54 +203,63 @@ impl<E: Clock + Rng + CryptoRng + Metrics, S: Scheme, D: Digest> State<E, S, D>
...
+            return (retry, nullify, Some(Voter::Notarization(notarization)));
         }
         if let Some(nullification) = self.construct_nullification(entry_view, true) {
-            return (nullify, Some(Voter::Nullification(nullification)));
+            return (retry, nullify, Some(Voter::Nullification(nullification)));
         }
 
         // If we couldn't find any entry certificates, return the nullify
-        (nullify, None)
+        (retry, nullify, None)
```
patrick-ogrady, line 253: If a certificate for entry doesn't exist, we still need to indicate it is a replay (to avoid re-writing nullify)

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -196,54 +203,63 @@ impl<E: Clock + Rng + CryptoRng + Metrics, S: Scheme, D: Digest> State<E, S, D>
...
+            return (retry, nullify, Some(Voter::Finalization(finalization)));
         }
         if let Some(notarization) = self.construct_notarization(entry_view, true) {
-            return (nullify, Some(Voter::Notarization(notarization)));
+            return (retry, nullify, Some(Voter::Notarization(notarization)));
         }
         if let Some(nullification) = self.construct_nullification(entry_view, true) {
-            return (nullify, Some(Voter::Nullification(nullification)));
+            return (retry, nullify, Some(Voter::Nullification(nullification)));
         }
```
patrick-ogrady, line 250: TODO: add a log here

consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -527,6 +535,7 @@ impl<
             .await;
     }
 
+    /// Emits any votes or certificates that became available for `view`.
```
patrick-ogrady, line 543: TODO: add note we do this because our response is scoped to a view (so we don't need to check if any view can do something)

consensus/src/simplex/actors/voter/actor.rs
patrick-ogrady, line 1: TODO: don't log round and proposal (round is included in proposal)

consensus/src/simplex/actors/voter/state.rs
patrick-ogrady, line 1: TODO: note that signing can only fail if we aren't a participant, so we don't need to worry about waiting to update state (otherwise we will attempt each view).

patrick-ogrady, line 1: Add note there is a strong assumption we only receive votes for the leader payload.

consensus/src/simplex/actors/voter/slot.rs
patrick-ogrady, line 1: Rename `record_proposal` and `update` (confusing what each does?)

patrick-ogrady, line 1: Handle bug in update that doesn't overwrite proposal on equivocation.

PR #2227, opened 2025-11-14
[coding] Rename Error -> ZodaError

cronokirby:
Not having it be exported like this was a mistake.


coding/src/lib.rs
```diff
@@ -25,7 +25,7 @@ mod no_coding;
 pub use no_coding::{NoCoding, NoCodingError};
```
patrick-ogrady: I think this should be `Error as NoCodingError` (a change in no_coding)?

PR #2235, opened 2025-11-14
[storage/adb] Introduce the OperationLog abstraction

roberto-bayardo:
- Introduce the OperationLog which pairs a snapshot with an authenticated log, and change all keyed adb variants to use it.
- Clean up & unify rewind operations and behavior, e.g. be consistent about never guaranteeing durability until next sync/commit.
- Some initial code unification in adb::current (though too much still remains duplicated, to be addressed later).

### DIFF
--- a/storage/src/journal/authenticated.rs
+++ b/storage/src/journal/authenticated.rs
@@ -401,7 +401,10 @@ where
         journal.rewind_to(rewind_predicate).await?;
 
         // Align the MMR and journal.
-        let mmr = Self::align(mmr, &journal, &mut hasher, APPLY_BATCH_SIZE).await?;
+        let mut mmr = Self::align(mmr, &journal, &mut hasher, APPLY_BATCH_SIZE).await?;
+
+        // Sync the journal and MMR to disk to avoid having to repeat any recovery that may have
+        // been performed on next startup.
         journal.sync().await?;
         mmr.sync().await?;
 


storage/src/adb/any/mod.rs
```diff
@@ -4,3 +4,415 @@
...
+        &mut self,
+        status: &mut AuthenticatedBitMap<D, N>,
+    ) -> Result<Location, Error> {
+        if self.is_empty() {
+            self.inactivity_floor_loc = self.op_count();
+            debug!(tip = ?self.inactivity_floor_loc, "db is empty, raising floor to tip");
+        } else {
+            let loc = self.inactivity_floor_loc;
+            self.inactivity_floor_loc = self
+                .as_floor_helper()
```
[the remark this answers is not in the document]
roberto-bayardo: fixed + regression test added.

storage/src/adb/any/mod.rs
```diff
@@ -4,3 +4,418 @@
...
+            translator: PhantomData,
+        })
+    }
+
+    /// Returns an [OperationLog] initialized directly from the given components. The log is
+    /// replayed from `start_loc` to build the snapshot, and that value is used as the inactivity
+    /// floor. The last commit location is set to None and it is the responsibility of the caller to
+    /// ensure it is set correctly.
+    pub async fn from_components(
+        start_loc: Location,
```
danlaine:
```suggestion
        inactivity_floor_loc: Location,
```

storage/src/adb/any/mod.rs
```diff
@@ -4,3 +4,418 @@
...
+    }
+
+    /// Returns the number of operations that have been appended to the log, including those that
+    /// are not yet committed, and any operations that have been pruned.
+    pub(super) fn op_count(&self) -> Location {
+        self.log.size()
+    }
+
+    /// Whether the snapshot currently has no active keys.
+    pub fn is_empty(&self) -> bool {
```
danlaine:
```suggestion
    pub(super) fn is_empty(&self) -> bool {
```

storage/src/adb/store/mod.rs
```diff
@@ -268,7 +271,7 @@ where
         .await?;
 
         // Rewind log to remove uncommitted operations.
-        let log_size = rewind_uncommitted(&mut log).await?;
+        let log_size = log.rewind_to(|op| op.is_commit()).await?;
```
danlaine, line 274: `rewind_uncommitted` synced the log if it removed anything but `rewind_to` doesn't, so I think we need to sync here

storage/src/journal/authenticated.rs
```diff
@@ -419,7 +398,7 @@ where
             variable::Journal::init(context.with_label("journal"), journal_cfg).await?;
 
         // Rewind to last matching operation.
-        rewind(&mut journal, rewind_predicate).await?;
+        journal.rewind_to(rewind_predicate).await?;
```
danlaine, line 401: `rewind` used to `sync` the journal but `rewind_to` doesn't so I think we need to do that here

storage/src/journal/authenticated.rs
```diff
@@ -374,22 +345,29 @@ where
 {
     /// Create a new [Journal] for fixed-length operations.
     ///
-    /// The journal will be rewound to the last operation that matches the `rewind_predicate` on initialization.
+    /// The journal will be rewound to the last operation that matches the `rewind_to` on
```
danlaine:
```suggestion
    /// The journal will be rewound to the last operation that matches the `rewind_predicate` on
```

storage/src/journal/authenticated.rs
```diff
@@ -419,7 +398,7 @@ where
             variable::Journal::init(context.with_label("journal"), journal_cfg).await?;
 
         // Rewind to last matching operation.
-        rewind(&mut journal, rewind_predicate).await?;
+        journal.rewind_to(rewind_predicate).await?;
 
         // Align the MMR and journal.
         let mmr = Self::align(mmr, &journal, &mut hasher, APPLY_BATCH_SIZE).await?;
```
danlaine: `Self::align` no longer syncs the MMR so I think we need to do that here. Alternatively, we could sync the MMR inside `Self::align`. I'd imagine we always want to sync the MMR after we align it to the journal?

danlaine: If we don't sync the MMR inside `align` then I think we may also want to change `from_components` (which calls `align`) to sync the MMR.

danlaine (commented):
Still working through this but leaving the comments I've got so far

storage/src/journal/authenticated.rs
roberto-bayardo: I don't think the sync is needed so long as the recovery always treats the log as source of truth.  Maybe it's a nice to have though?

storage/src/adb/store/mod.rs
roberto-bayardo: These syncs appear obsolete which is why I dropped them now that a committed log is all you need to ensure durability.

storage/src/journal/authenticated.rs
roberto-bayardo: fixed

storage/src/adb/any/mod.rs
roberto-bayardo: fixed

storage/src/adb/any/mod.rs
roberto-bayardo: changed

storage/src/journal/authenticated.rs
roberto-bayardo: I've changed all the rewind operations to consistently have a "no sync" contract.

storage/src/journal/authenticated.rs
roberto-bayardo: Alright I've re-instated syncing on new/init for any structure that potentially performs recovery.

danlaine (approved):

PR #2240, opened 2025-11-14
[ci] Upload all coverage reports at the same time

clabby:
## Overview

Uploads all partitions' coverage reports simultaneously rather than as each partition completes to avoid partially-complete status updates from codecov.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -51,15 +51,15 @@ jobs:
         tool: cargo-llvm-cov@0.6.19,cargo-nextest@0.9.104
     - name: Generate coverage report
       run: cargo llvm-cov nextest --workspace --lcov --output-path lcov-${{ matrix.partition }}.info --partition hash:${{ matrix.partition }}/8 --run-ignored all
-    - name: Upload coverage artifact
+    - name: Store coverage report
       uses: actions/upload-artifact@330a01c490aca151604b8cf639adc76d48f6c5d4 # v5.0.0
       with:
         name: coverage-lcov-${{ matrix.partition }}
         path: lcov-${{ matrix.partition }}.info
         retention-days: 1
 
-  UploadCoverage:
-    name: Upload Combined Coverage
+  Upload:
+    name: Upload
     runs-on: ubuntu-latest
     needs: Coverage
     steps:
@@ -70,7 +70,7 @@ jobs:
       with:
         pattern: coverage-lcov-*
         merge-multiple: true
-    - name: Upload coverage to Codecov
+    - name: Upload to Codecov
       uses: codecov/codecov-action@5a1091511ad55cbe89839c7260b706298ca349f7 # v5.5.1
       with:
         files: lcov-*.info


.github/workflows/coverage.yml
```diff
@@ -50,11 +50,28 @@ jobs:
...
+    - name: Download coverage reports
+      uses: actions/download-artifact@018cc2cf5baa6db3ef3c5f8a56943fffe632ef53 # v6.0.0
+      with:
+        pattern: coverage-lcov-*
+        merge-multiple: true
     - name: Upload coverage to Codecov
       uses: codecov/codecov-action@5a1091511ad55cbe89839c7260b706298ca349f7 # v5.5.1
       with:
-        files: lcov.info
+        files: lcov-1.info,lcov-2.info,lcov-3.info,lcov-4.info,lcov-5.info,lcov-6.info,lcov-7.info,lcov-8.info
```
clabby: This is unfortunate, but no glob pattern support. https://github.com/codecov/codecov-action?tab=readme-ov-file#arguments

clabby: Sike, just not documented 👀 https://github.com/codecov/codecov-action/issues/657

.github/workflows/coverage.yml
```diff
@@ -50,11 +50,28 @@ jobs:
       with:
         tool: cargo-llvm-cov@0.6.19,cargo-nextest@0.9.104
     - name: Generate coverage report
-      run: cargo llvm-cov nextest --workspace --lcov --output-path lcov.info --partition hash:${{ matrix.partition }}/8 --run-ignored all
+      run: cargo llvm-cov nextest --workspace --lcov --output-path lcov-${{ matrix.partition }}.info --partition hash:${{ matrix.partition }}/8 --run-ignored all
+    - name: Upload coverage artifact
+      uses: actions/upload-artifact@330a01c490aca151604b8cf639adc76d48f6c5d4 # v5.0.0
+      with:
+        name: coverage-lcov-${{ matrix.partition }}
```
patrick-ogrady, line 57: do we need to give this a unique name by commit?

.github/workflows/coverage.yml
```diff
@@ -50,11 +50,28 @@ jobs:
...
+        path: lcov-${{ matrix.partition }}.info
+
+  UploadCoverage:
+    name: Upload Combined Coverage
+    runs-on: ubuntu-latest
+    needs: Coverage
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@08eba0b27e820071cde6df949e0beb9ba4906955 # v4
+    - name: Download coverage reports
```
patrick-ogrady, line 68: should we delete after this?

.github/workflows/coverage.yml
clabby: They're already isolated ([src](https://docs.github.com/en/enterprise-server%403.12/actions/writing-workflows/choosing-what-your-workflow-does/storing-and-sharing-data-from-a-workflow?utm_source=chatgpt.com#about-workflow-artifacts)):

> Downloading files: You can only download artifacts that were uploaded during the same workflow run. When you download a file, you can reference it by name.

.github/workflows/coverage.yml
clabby: Shouldn't need to. There's no official "delete artifact" action from GH, and github will automatically clean them up after the retention period.

Went ahead and specified the retention period as 1d (minimum), in-case our repo's settings are higher for other artifacts.

patrick-ogrady:
Will approve once coverage finishes.

patrick-ogrady (dismissed):

patrick-ogrady (approved):

Issue #2223, opened 2025-11-13
[storage/adb/ordered] Batch-optimize ordered updates


roberto-bayardo:
Each update in the ordered ADB currently does a historical read to determine the required key ordering updates.  These could be batched and done during commit() for better performance.

roberto-bayardo:
Our adb benchmarks don't currently demonstrate a huge difference between ordered/unordered ADB variants. This is not to say this feature is unimportant -- it more likely exposes the inadequacy of the current benchmarks. As part of this issue we should probably find a simple enough benchmark that clearly reveals the performance concern so we can quantify the impact of any solution.

bochunbai:
Difference happen mostly for larger batch writes.

Changing [NUM_ELEMENTS](https://github.com/commonwarexyz/monorepo/blob/main/storage/src/adb/benches/fixed/generate.rs#L17) from `1_000` to `1_000_000` produces difference between ordered and unordered:

```
adb::fixed_generate/variant=any::fixed::unordered elements=1000000 operations=10000
                        time:   [2.9061 s 2.9239 s 2.9435 s]
```

```
adb::fixed_generate/variant=any::fixed::ordered elements=1000000 operations=10000
                        time:   [9.2933 s 9.3689 s 9.4399 s]
```

With default configuration, they both about 500ms-600ms

roberto-bayardo:
Thanks. Likely a lot of cache misses on those reads. The buffer pool size in those benchmarks is intentionally set rather small to test impact of cache misses, so making a note to increase the buffer pool size when we start testing batching impacts.

roberto-bayardo:
Done in https://github.com/commonwarexyz/monorepo/pull/2472

roberto-bayardo:
> Changing [NUM_ELEMENTS](https://github.com/commonwarexyz/monorepo/blob/main/storage/src/adb/benches/fixed/generate.rs#L17) from `1_000` to `1_000_000` produces difference between ordered and unordered:

The batch update algo just merged brings this performance inline. The performance improvement results from the batch algorithm only inserting one update per new key when the previous key is also a new key, whereas the non-batch algorithm will create the key and then update the previous key unconditionally. I don't know if this will be that common in practice though since I'd imagine in a fully populated DB most new keys will have existing preceeding keys in the DB?

PR #2242, opened 2025-11-14
[WiP] [storage/adb/store] add a batch update API to db trait + batcher implementation

roberto-bayardo:
- Adds a batch-update call to the Db API, with a default implementation that doesn't actually batch.
- Adds a batcher structure that implements the Db trait that creates and execute batches from standard unbatched API calls. Also adds a delete_unchecked operation that doesn't return any value and avoids any calls to the underlying Db for greater efficiency.
- Extends db generation benchmarks to compare against a batched version of each variant.

Initial work towards https://github.com/commonwarexyz/monorepo/issues/2223

---

> [!NOTE]
> Adds a `batch_update` API to `Db` and a `Batcher` wrapper that transparently batches updates/deletes with tests.
> 
> - **Storage/ADB Store**:
>   - **Db trait**: Add `batch_update` with default (non-batching) implementation.
>   - **New batching layer**: Introduce `batcher::Batcher<E,K,V,T,D>` that caches `update`/`delete` calls and applies them in bulk on `commit`, `sync`, or `apply_updates`.
>     - Supports `delete_unchecked` and overlay-aware `get`; warns on close/destroy with pending updates.
>   - **Exports**: Expose `pub mod batcher` in `store::mod`.
>   - **Tests**: Add unit tests validating batched vs. unbatched equivalence and batch interactions.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c55f0431a1bf7b4209fbaa6501105244c46c4176. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/storage/src/adb/store/batcher.rs
+++ b/storage/src/adb/store/batcher.rs
@@ -1,7 +1,6 @@
-//! The [Batcher] implements the [ trait to provide a transparent batching layer on top of any
-//! other [Db] implementation. Calls to the batcher's update and delete methods are cached and
-//! applied to the wrapped database in batch upon calling [Batcher::commit] or
-//! [Batcher::apply_updates].
+//! The [Batcher] implements the [Batchable] trait to provide a transparent batching layer on top of
+//! a [Db]. Calls to the batcher's update and delete methods are cached and applied to the wrapped
+//! database in batch upon calling [Batcher::commit] or [Batcher::apply_updates].
 
 use super::{Batchable, Db, Error};
 use crate::translator::Translator;
@@ -52,7 +51,8 @@ impl<
         &self.db
     }
 
-    /// Take the underlying database from the batcher. Any cached updates will be applied before returning.
+    /// Take the underlying database from the batcher. Any cached updates will be applied before
+    /// returning.
     pub async fn take(mut self) -> Result<D, Error> {
         self.apply_updates().await?;
 
@@ -64,8 +64,8 @@ impl<
     ///
     /// # Errors
     ///
-    /// - Propagates any underlying database errors. Because these errors are generally unrecoverable,
-    ///   no effort is made to preserve batcher state should one occur.
+    /// - Propagates any underlying database errors. Because these errors are generally
+    ///   unrecoverable, no effort is made to preserve batcher state should one occur.
     pub async fn apply_updates(&mut self) -> Result<(), Error> {
         let updates = std::mem::take(&mut self.updates);
         let updates_iter =
--- a/storage/src/adb/store/mod.rs
+++ b/storage/src/adb/store/mod.rs
@@ -130,6 +130,7 @@ pub struct Config<T: Translator, C> {
     pub buffer_pool: PoolRef,
 }
 
+/// Implemented by databases that support batchable operations and batching layers that wrap them.
 pub trait Batchable<K: Array, V: Codec> {
     /// Get the value of `key` in the db, or None if it has no value.
     fn get(&self, key: &K) -> impl Future<Output = Result<Option<V>, Error>>;


storage/src/adb/store/batcher.rs
```diff
@@ -0,0 +1,406 @@
...
+                });
+        let deletes_iter =
+            updates
+                .iter()
+                .filter_map(|(key, update_or_delete)| match update_or_delete {
+                    UpdateOrDelete::Update(_) => None,
+                    UpdateOrDelete::Delete => Some(key.clone()),
+                });
+        self.db.batch_update(updates_iter, deletes_iter).await
+    }
```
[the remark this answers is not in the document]
roberto-bayardo: This is fine, if batch_update fails then recovery is not possible anyway.

storage/src/adb/store/batcher.rs
```diff
@@ -0,0 +1,406 @@
...
+        if let Some(update_or_delete) = self.updates.get_mut(&key) {
+            match update_or_delete {
+                UpdateOrDelete::Update(_) => {
+                    *update_or_delete = UpdateOrDelete::Delete;
+                }
+                UpdateOrDelete::Delete => {}
+            }
+        }
+        self.updates.insert(key, UpdateOrDelete::Delete);
+    }
```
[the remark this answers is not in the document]
roberto-bayardo: you are right, fixed!

storage/src/adb/benches/variable/generate.rs
```diff
@@ -21,57 +25,83 @@ fn bench_variable_generate(c: &mut Criterion) {
...
+        let new_db = gen_random_kv(batched_db, elements, operations, commit_frequency).await;
+        db = new_db.take().await?;
+    } else {
+        db = gen_random_kv(db, elements, operations, commit_frequency).await;
+    };
+    let elapsed = start.elapsed();
+
+    db.destroy().await.unwrap(); // don't time destroy
+
+    Ok(elapsed)
```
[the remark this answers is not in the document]
roberto-bayardo: These benchmarks aren't meant to test equivalent operations, but variable benchmark performs the sync/prune too just within the generation function instead.

storage/src/adb/store/batcher.rs
```diff
@@ -0,0 +1,452 @@
...
+    async fn get(&self, key: &K) -> Result<Option<V>, Error> {
+        if let Some(update_or_delete) = self.updates.get(key) {
+            return match update_or_delete {
+                UpdateOrDelete::Update(value) => Ok(Some(value.clone())),
+                UpdateOrDelete::Delete => Ok(None),
+            };
+        }
+
+        self.db.get(key).await
+    }
```
danlaine, line 104: Nit: I think this is slightly cleaner but fine either way.
```suggestion
    async fn get(&self, key: &K) -> Result<Option<V>, Error> {
        match self.updates.get(key) {
            Some(UpdateOrDelete::Update(value)) => Ok(Some(value.clone())),
            Some(UpdateOrDelete::Delete) => Ok(None),
            None => self.db.get(key).await,
        }
    }
```

storage/src/adb/store/batcher.rs
```diff
@@ -0,0 +1,452 @@
...
+
+impl<
+        E: Storage + Clock + Metrics,
+        K: Array,
+        V: Codec + Clone,
+        T: Translator,
+        D: Db<E, K, V, T>,
+    > Db<E, K, V, T> for Batcher<E, K, V, T, D>
+{
+    fn op_count(&self) -> Location {
```
danlaine: I worry that users might mistakenly expect `op_count` to include updates/deletions which are in `updates` but haven't yet been flushed to the underlying database.

Do we need the `Batcher` to implement `Db`?

An alternative approach would be to have a new trait `BatchableDb: Db` (or whatever name we like) which has a method `fn start_batch(&mut self) -> Batch<'_>` where `Batch` has some of `Db`'s methods (e.g. `update`, `delete`) but not, for example, `op_count`, `close`, `sync`.

Something similar to how `commonware-utils::bitmap::historical` works, in other words.

storage/src/adb/store/mod.rs
```diff
@@ -175,6 +177,39 @@ pub trait Db<E: RStorage + Clock + Metrics, K: Array, V: Codec, T: Translator> {
...
+    ///
+    /// The default implementation simply calls the unbatched update/delete methods for each
+    /// updated/deleted key in the order they appear in the input.
+    fn batch_update<'a>(
+        &'a mut self,
+        updates: impl Iterator<Item = (K, V)> + 'a,
+        deletes: impl Iterator<Item = K> + 'a,
+    ) -> impl Future<Output = Result<(), Error>> + 'a
+    where
+        V: 'a,
```
danlaine: I think we can simplify these lifetimes a bit and let the compiler figure it out.

```suggestion
    fn batch_update(
        &mut self,
        updates: impl Iterator<Item = (K, V)>,
        deletes: impl Iterator<Item = K>,
    ) -> impl Future<Output = Result<(), Error>>
```

danlaine: We could also remove `deletes` and change `updates` to be

```rs
updates: impl Iterator<Item = (K, Option<V>)>
```

(where `None` means the key has been deleted) to remove ambiguity about the order in which the operations are applied.

I think this would allow use to use `into_iter` rather than `iter` in `apply_updates` which would allow us to remove the key and value `.clone()`s when we're turning the `updates` HashMap into an iterator.

storage/src/adb/store/batcher.rs
roberto-bayardo: Yeah this is a bug. it's much nicer if it does fully implement Db (e.g. see benchmarks).  Let me figure out a good way to fix this.

storage/src/adb/store/mod.rs
roberto-bayardo: True, however we want to offer implementations the opportunity to re-order for max-performance through a bit of "query optimization", so ambiguity is good & intentional here. My suspicion is most implementations will want them separated out in determining an optimal plan, but let me think a bit more about it.

storage/src/adb/store/batcher.rs
roberto-bayardo: yes much nicer, changed.

storage/src/adb/store/mod.rs
roberto-bayardo: For now I've removed the (AI generated!) lifetime cruft.

storage/src/adb/store/batcher.rs
```diff
@@ -0,0 +1,449 @@
...
+    /// Apply the cached updates then sync the underlying db.
+    async fn sync(&mut self) -> Result<(), Error> {
+        self.apply_updates().await?;
+
+        self.db.sync().await
+    }
+
+    async fn prune(&mut self, prune_loc: Location) -> Result<(), Error> {
+        self.db.prune(prune_loc).await
+    }
```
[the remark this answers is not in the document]
roberto-bayardo: prune() only prunes inactive operations/bits, so this is fine.

storage/src/adb/store/batcher.rs
roberto-bayardo: I have changed the batcher's op_count apply updates before returning op_count.  This isn't exactly ideal but the alternative would be potential non-monotonicity in op_count if cached updates are later deleted, which might be weird.

Let's revisit whether to adopt a first-class Batch abstraction? Seems like it might be a much bigger change which I'm not quite ready to pursue.

roberto-bayardo: Bah, unfortunately this will require making op_count async.

roberto-bayardo:
Putting this back in draft while we sort out how the API should look.

storage/src/adb/store/batcher.rs
roberto-bayardo: OK, I've added a "Batchable" trait that is implemented by every Db separating out the batcher-specific methods. Not sure I like what I have though. We can discuss more out-of-band.

roberto-bayardo:
closing in favor of 2283

Issue #2244, opened 2025-11-14
[cryptography] Avoid collisions between signing messages with and without namespace


cronokirby:
If you look at the way we handle `sign_message`, there's a collision between:

- including a namespace, say `N`,
- including no namespace, but starting a message with `len(N) || N`.

We should have a better story here, e.g. by mandating namespaces (using the empty namespace if users really want to provide none), or by adding an extra byte to indicate the presence or absence of a namespace, or by using the transcript API.

cronokirby:
This is now fixed, by requiring a namespace always.

PR #2245, opened 2025-11-14
Add a few comments explaining TLE divergences from paper

### LINKED ISSUE #1528: TLE Issues/observations
https://github.com/commonwarexyz/monorepo/blob/1f7f174bccd4ea848d6b4ddfae621493a55db5f6/cryptography/src/bls12381/tle.rs#L163-L174
1. do we also not want to add an `h3` to `combined`? This is probably intentional due to the hash to field construction but we were wondering if it were an oversight.
2. It may be clearer to have message be a `Block` to match the paper's definition of `h3` being a map from `{0,1}^l  x {0,1}^l -> Zp*`
3. Should we also ensure the hash to field's output is nonzero just to conform to the spec? Obviously this is extremely unlikely to happen and it is not strictly necessary.

https://github.com/commonwarexyz/monorepo/blob/1f7f174bccd4ea848d6b4ddfae621493a55db5f6/cryptography/src/bls12381/tle.rs#L276-L279 here we compute `e(r*P_pub,Qid)` but the paper computes `e(P_pub,Qid)^r`. By bilinearity this should be equivalent and probably more efficient due to doing scalar multiplication in `G1` rather than in `Fp_12` , but I would prefer a comment explaining this optimization.

https://github.com/commonwarexyz/monorepo/blob/1f7f174bccd4ea848d6b4ddfae621493a55db5f6/cryptography/src/bls12381/tle.rs#L255-L262 there is a potential hash collision here between messages that have no namespace and a namespace. For any message `m` and namespace `n`, we have that`q_id(None, len(n)+n+m) = q_id(n,m)`. The following test shows this by showing how the empty namespace collides with a message starting with 0 and no namespace.

```rust
    #[test]
    fn test_hash_collision() {
        let empty_namespace: &[u8] = &[];
        let target1 = (None,[0,1,2,3].as_slice());
        let target2 = (Some(empty_namespace),[1,2,3].as_slice());
        let compute_q_id = |target| {
            let q_id = match target {
            (None, target) => hash_message::<MinPk>(MinPk::MESSAGE, target),
            (Some(namespace), target) => hash_message_namespace::<MinPk>(MinPk::MESSAGE, namespace, target),
        };
        q_id
        };
        let q_id1 = compute_q_id(target1);
        let q_id2 = compute_q_id(target2);
        assert_eq!(q_id1,q_id2); 
    }
```
I personally think any message without a namespace should just get serialized as though it has an empty namespace.

cronokirby:
Closes #1528.

The point about namespace collision remains valid, but I moved that to https://github.com/commonwarexyz/monorepo/issues/2244, as the issue affects more than just TLE.

### DIFF
--- a/cryptography/src/bls12381/tle.rs
+++ b/cryptography/src/bls12381/tle.rs
@@ -22,7 +22,7 @@
 //! 2. Deriving encryption randomness r = H3(sigma || message)
 //! 3. Computing the ciphertext components:
 //!    - U = r * G (commitment in G1)
-//!    - V = sigma ⊕ H2(e(r * P_pub, Q_id)) (masked random value)
+//!    - V = sigma ⊕ H2(e(P_pub, Q_id)^r) (masked random value)
 //!    - W = M ⊕ H4(sigma) (masked message)
 //!
 //! Where Q_id = H1(target) maps the target to a point in G2.
@@ -114,7 +114,7 @@ impl From<Digest> for Block {
 pub struct Ciphertext<V: Variant> {
     /// First group element U = r * Public::one().
     pub u: V::Public,
-    /// Encrypted random value V = sigma XOR H2(e(r * P_pub, Q_id)).
+    /// Encrypted random value V = sigma XOR H2(e(P_pub, Q_id)^r).
     pub v: Block,
     /// Encrypted message W = M XOR H4(sigma).
     pub w: Block,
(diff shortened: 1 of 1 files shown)


cryptography/src/bls12381/tle.rs
```diff
@@ -273,7 +276,8 @@ pub fn encrypt<R: CryptoRngCore, V: Variant>(
     let mut u = V::Public::one();
     u.mul(&r);
 
-    // Compute e(r * P_pub, Q_id)
+    // Compute e(P_pub, Q_id)^r = e(r * P_pub, Q_id)
+    // (The latter expression is more efficient to compute).
```
patrick-ogrady: nit: reformulate the `(...)` into a sentence and put after a `\n` like the above change.

cryptography/src/bls12381/tle.rs
patrick-ogrady, line 1: should we add a `Divergence: ...` to make it clear why we are commenting on these?

cronokirby: That would make sense if the mapping with the spec were a lot tighter. For example, if each portion of the code mapped to a portion of the specification. But, we're already not doing things as tightly, so I think this doesn't add much utility.

Issue #2222, opened 2025-11-13
[storage/(a)db] Support "create" operation


roberto-bayardo:
Instead of just update/upsert, support a "create" operation that would error out if they key exists.

roberto-bayardo:
re-opening...the optimized create was removed accidentally when we removed non-batch database updates.

PR #2248, opened 2025-11-15
[storage/adb/store] Add create op to Db trait + other cleanup

roberto-bayardo:
- Add a create operation to the Db trait + implementations for all affected stores. 
- Remove duplicate definitions of Db trait method implementations.
- Make upsert part of Db trait with a default implementation, allowing removal of the existing multiple definitions.
- Remove oldest_retained_loc from the public keyed db apis since this is a contiguous journal concept, not an adb one. for adbs, inactivity_floor_loc should be used instead as a lower-bound on location.

Resolves https://github.com/commonwarexyz/monorepo/issues/2222

---

> [!NOTE]
> Adds `Db::create` with boolean result, makes `upsert` a default trait method, unifies Db impls across ADBs (and changes `delete` to return bool), removes `oldest_retained_loc` from public APIs, and updates examples, fuzzers, tests, and benches accordingly.
> 
> - **API**:
>   - Add `Db::create(key, value) -> bool` and make `Db::upsert` a default method.
>   - Change `Db::delete` to return `bool` indicating whether a key was deleted.
>   - Remove `oldest_retained_loc` from public ADB APIs (use `log.oldest_retained_pos/loc` internally).
> - **Core Storage**:
>   - Introduce helpers `update_key` and `create_key`; replace prior `update_loc` usages.
>   - Consolidate `Db` trait implementations across `any::fixed` (ordered/unordered), `any::variable`, `current` (ordered/unordered), and `store`; move commit/sync/prune/close/destroy into `Db` impls where applicable.
>   - Add ordered-any `create_with_callback`; extend internal update path to support create-only.
> - **Examples/Benchmarks/Fuzz**:
>   - Update sync client and example DB to use `adb::store::Db` methods (e.g., `Db::commit`).
>   - Adjust fuzz targets to new `create`/`delete -> bool` semantics and remove `OldestRetainedLoc` op; reference `inactivity_floor_loc`/`log.oldest_retained_*` as needed.
>   - Benchmarks import `Db` trait and use unified APIs.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0d66d7d0a9d743aa35814024120597f21f60fd4e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/storage/src/adb/any/fixed/ordered.rs
+++ b/storage/src/adb/any/fixed/ordered.rs
@@ -460,8 +460,8 @@ impl<
                 value,
                 next_key: key,
             });
-            self.log.append(op).await?;
             callback(None);
+            self.log.append(op).await?;
             self.active_keys += 1;
             return Ok(true);
         }


storage/src/adb/current/ordered.rs
```diff
@@ -772,35 +674,102 @@ impl<
...
+            .delete_with_callback(key, |append, loc| {
+                if let Some(loc) = loc {
+                    self.status.set_bit(*loc, false);
+                }
+                self.status.push(append);
+                r = true;
+            })
+            .await?;
+
+        Ok(r)
```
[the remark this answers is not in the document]
roberto-bayardo: the callback won't be invoked if nothing is deleted 👍

storage/src/adb/any/fixed/ordered.rs
```diff
@@ -695,43 +653,78 @@ impl<E: Storage + Clock + Metrics, K: Array, V: CodecFixed<Cfg = ()>, H: Hasher,
...
+
+    async fn create(&mut self, key: K, value: V) -> Result<bool, Error> {
+        self.create_with_callback(key, value, |_| {}).await
     }
 
-    async fn delete(&mut self, key: K) -> Result<(), Error> {
-        self.delete(key).await
+    async fn delete(&mut self, key: K) -> Result<bool, Error> {
+        let mut r = false;
+        self.delete_with_callback(key, |_, _| r = true).await?;
```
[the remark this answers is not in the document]
roberto-bayardo: the callback won't be invoked if nothing is deleted 👍

storage/src/adb/any/fixed/ordered.rs
```diff
@@ -461,11 +444,52 @@ impl<
...
+                });
+                self.log.append(value_update_op).await?;
+                self.log.append(next_key_update_op).await?;
+            }
+        };
+
+        // Creating a new key involves inactivating a previous span, requiring we increment `steps`.
+        self.steps += 1;
+
+        Ok(true)
```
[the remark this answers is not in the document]
roberto-bayardo: no, update_loc will invoke the callback twice.

storage/src/adb/any/fixed/ordered.rs
```diff
@@ -426,7 +409,7 @@ impl<
             self.active_keys += 1;
             return Ok(());
         }
-        let res = self.update_loc(&key, next_loc, callback).await?;
+        let res = self.update_loc(&key, false, next_loc, callback).await?;
```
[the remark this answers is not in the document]
roberto-bayardo: no, update_loc will invoke the callback twice.

examples/sync/src/databases/any.rs
```diff
@@ -83,27 +84,27 @@ where
                     database.delete(key).await?;
                 }
                 Operation::CommitFloor(_) => {
-                    database.commit().await?;
+                    <Self as Db<_, _, _, _>>::commit(database).await?;
```
clabby:
```suggestion
                    Db::commit(database).await?;
```

examples/sync/src/databases/any.rs
```diff
@@ -83,27 +84,27 @@ where
...
                 }
             }
         }
         Ok(())
     }
 
     async fn commit(&mut self) -> Result<(), commonware_storage::adb::Error> {
-        self.commit().await
+        <Self as Db<_, _, _, _>>::commit(self).await
```
clabby:
```suggestion
        Db::commit(self).await
```

examples/sync/src/databases/any.rs
```diff
@@ -83,27 +84,27 @@ where
...
     }
 
     fn root(&self, hasher: &mut Standard<commonware_cryptography::Sha256>) -> Key {
         self.root(hasher)
     }
 
     fn op_count(&self) -> Location {
-        self.op_count()
+        <Self as Db<_, _, _, _>>::op_count(self)
```
clabby:
```suggestion
        Db::op_count(self)
```

examples/sync/src/databases/any.rs
```diff
@@ -83,27 +84,27 @@ where
...
     }
 
     fn lower_bound(&self) -> Location {
-        self.inactivity_floor_loc()
+        <Self as Db<_, _, _, _>>::inactivity_floor_loc(self)
```
clabby:
```suggestion
        Db::inactivity_floor_loc(self)
```

storage/src/adb/any/fixed/unordered.rs
```diff
@@ -321,43 +232,88 @@ impl<E: Storage + Clock + Metrics, K: Array, V: CodecFixed<Cfg = ()>, H: Hasher,
...
+
+        Ok(true)
+    }
+
+    async fn delete(&mut self, key: K) -> Result<bool, Error> {
+        if self.delete_return_loc(key).await?.is_none() {
+            Ok(false)
+        } else {
+            Ok(true)
+        }
```
clabby:
```suggestion
        self.delete_return_loc(key).await.map(|loc| loc.is_some())
```

storage/src/adb/current/ordered.rs
```diff
@@ -772,35 +674,102 @@ impl<
...
+                if let Some(loc) = loc {
+                    self.status.set_bit(*loc, false);
+                }
+            })
+            .await?
+        {
+            return Ok(false);
+        }
+
+        Ok(true)
```
clabby:
```suggestion
        self.any
            .create_with_callback(key, value, |loc| {
                self.status.push(true);
                if let Some(loc) = loc {
                    self.status.set_bit(*loc, false);
                }
            })
            .await
```

storage/src/adb/store/mod.rs
```diff
@@ -143,10 +143,37 @@ pub trait Db<E: RStorage + Clock + Metrics, K: Array, V: Codec, T: Translator> {
...
+    where
+        V: Default,
+    {
+        async {
+            let mut value = self.get(&key).await?.unwrap_or_default();
+            update(&mut value);
+
+            self.update(key, value).await
+        }
+    }
```
clabby, line 166: Nice 😄

storage/src/adb/current/ordered.rs
roberto-bayardo: done

storage/src/adb/any/fixed/unordered.rs
roberto-bayardo: done

examples/sync/src/databases/any.rs
roberto-bayardo: done (+ others)

storage/src/adb/any/fixed/ordered.rs
```diff
@@ -461,11 +444,53 @@ impl<
...
+            // We're inserting the very first key. For this special case, the next-key value is the
+            // same as the key.
+            self.snapshot.insert(&key, next_loc);
+            let op = Operation::Update(KeyData {
+                key: key.clone(),
+                value,
+                next_key: key,
+            });
+            self.log.append(op).await?;
+            callback(None);
```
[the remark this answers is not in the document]
clabby: This could be true if appending to the log fails; `callback` would update the bitmap w/ a dangling bit in `update_with_callback`.

clabby (approved):

storage/src/adb/any/fixed/ordered.rs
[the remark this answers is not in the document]
roberto-bayardo: The inconsistent order is a good callout, but in general we aren't trying to maintain consistency in the event of Storage errors (most applications would just shutdown in that case). 

Moved the callback call before the append to make it match the implementation in update_with_callback.

PR #2253, opened 2025-11-16
[workspace] Use `nextest` filters instead of `#[ignore]`

clabby:
## Overview

Moves over to using `nextest` filters instead of `#[ignore]`. This pattern is a good bit more flexible, enabling us to create multiple test groups, [test group priorities](https://nexte.st/docs/configuration/test-priorities/), [dedicating more threads to specific test](https://nexte.st/docs/configuration/threads-required/), [and more](https://nexte.st/docs/configuration/per-test-overrides/). The `#[ignore]` attribute should generally be reserved for tests that are flaky.

---

> [!NOTE]
> Switches from #[ignore]-based heavy tests to nextest profile filtering, updates CI to run heavy and all tests via profiles, and renames tests across crates to `test_heavy_*`.
> 
> - **Testing Configuration**:
>   - Add `.config/nextest.toml` with profiles: `default` (excludes `test_heavy_*`), `heavy` (only `test_heavy_*`), and `all` (runs everything with priorities).
> - **CI**:
>   - Coverage workflow: use `cargo llvm-cov nextest --profile all` (replacing `--run-ignored`).
>   - Slow workflow: run heavy tests via `--profile heavy` (replacing `-- --ignored`).
> - **Codebase-wide test updates**:
>   - Rename previously `#[ignore]` tests to `test_heavy_*` and drop `#[ignore]` across multiple crates: `consensus`, `examples/reshare`, `p2p/authenticated/{discovery,lookup}`, `runtime/network/{audited,deterministic,iouring,metered,tokio}`, `storage/{archive,cache,freezer,metadata,ordinal}`.
>   - No functional logic changes beyond test naming/attributes to align with nextest filters.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6404cc2ad7a98ff5aa62922aa5f85b2600dae1b3. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,11 +1,1 @@
-[test-groups]
-heavy = { max-threads = 1 }
-
-[profile.default]
-default-filter = 'not test(/test_heavy_.*/)'
-
-[profile.heavy]
-default-filter = 'test(/test_heavy_.*/)'
-
-[profile.all]
-default-filter = 'test(/.*/)'
+../nextest.toml
--- a/.github/workflows/slow.yml
+++ b/.github/workflows/slow.yml
@@ -56,8 +56,8 @@ jobs:
       uses: taiki-e/install-action@0c5db7f7f897c03b771660e91d065338615679f4 # v2.60.0
       with:
         tool: just@1.43.0,cargo-nextest@0.9.104
-    - name: Run heavy tests
-      run: just test ${{ matrix.flags }} --partition hash:${{matrix.partition}}/8 --verbose --profile heavy
+    - name: Run slow tests
+      run: just test ${{ matrix.flags }} --partition hash:${{matrix.partition}}/8 --verbose --profile slow
 
   Tests-Gate:
     name: "Slow-Tests"
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -118,8 +118,11 @@ just miri <module>::
 ### Extended Checks (before PR)
 
 ```bash
-# Long-running tests
-just test --workspace -- --ignored
+# Long-running tests only
+just test --workspace --profile slow
+
+# All tests
+just test --workspace --profile all
 
 # Fuzz testing (60 seconds per target)
 just fuzz <primitive>/fuzz
--- a/macros/src/lib.rs
+++ b/macros/src/lib.rs
@@ -5,6 +5,7 @@
     html_favicon_url = "https://commonware.xyz/favicon.ico"
 )]
 
+use crate::nextest::configured_test_groups;
 use proc_macro::TokenStream;
 use proc_macro2::Span;
 use proc_macro_crate::{crate_name, FoundCrate};
@@ -142,8 +143,8 @@ pub fn test_traced(attr: TokenStream, item: TokenStream) -> TokenStream {
 
 /// Prefix a test name with a nextest filter group.
 ///
-/// This renames `test_some_behavior` into `test_<group>_some_behavior`, making
-/// it easy to filter tests by group prefixes in nextest.
+/// This renames `test_some_behavior` into `test_some_behavior_<group>_`, making
+/// it easy to filter tests by group postfixes in nextest.
 #[proc_macro_attribute]
 pub fn test_group(attr: TokenStream, item: TokenStream) -> TokenStream {
     if attr.is_empty() {
@@ -162,24 +163,20 @@ pub fn test_group(attr: TokenStream, item: TokenStream) -> TokenStream {
         Ok(group) => group,
         Err(err) => return err.to_compile_error().into(),
     };
-    if let Err(err) = nextest::ensure_group_known(&group, group_literal.span()) {
+    let groups = match configured_test_groups() {
+        Ok(groups) => groups,
+        Err(_) => {
+            // Don't fail the compilation if the file isn't found; just return the original input.
+            return TokenStream::from(quote!(#input));
+        }
+    };
+
+    if let Err(err) = nextest::ensure_group_known(groups, &group, group_literal.span()) {
         return err.to_compile_error().into();
     }
 
     let original_name = input.sig.ident.to_string();
-    let suffix = original_name
-        .strip_prefix("test_")
-        .filter(|suffix| !suffix.is_empty())
-        .unwrap_or(original_name.as_str());
-
-    let new_ident = if suffix.is_empty() {
-        Ident::new(&format!("test_{}", group), input.sig.ident.span())
-    } else {
-        Ident::new(
-            &format!("test_{}_{}", group, suffix),
-            input.sig.ident.span(),
-        )
-    };
+    let new_ident = Ident::new(&format!("{original_name}_{group}_"), input.sig.ident.span());
 
     input.sig.ident = new_ident;
 
--- a/nextest.toml
+++ b/nextest.toml
@@ -1,0 +1,11 @@
+[test-groups]
+slow = { max-threads = "num-cpus" }
+
+[profile.default]
+default-filter = 'not test(/.*_slow_$/)'
+
+[profile.slow]
+default-filter = 'test(/.*_slow_$/)'
+
+[profile.all]
+default-filter = 'test(/.*/)'
--- a/runtime/src/network/iouring.rs
+++ b/runtime/src/network/iouring.rs
@@ -370,8 +370,8 @@ mod tests {
         .await;
     }
 
+    #[test_group("slow")]
     #[tokio::test]
-    #[test_group("heavy")]
     async fn test_stress_trait() {
         tests::stress_test_network_trait(|| {
             Network::start(
--- a/consensus/src/ordered_broadcast/mod.rs
+++ b/consensus/src/ordered_broadcast/mod.rs
@@ -552,7 +552,7 @@ mod tests {
         });
     }
 
-    #[test_group("heavy")]
+    #[test_group("slow")]
     #[test_traced]
     fn test_network_partition() {
         network_partition::<MinPk>();
@@ -623,7 +623,7 @@ mod tests {
         slow_and_lossy_links::<MinSig>(0);
     }
 
-    #[test_group("heavy")]
+    #[test_group("slow")]
     #[test_traced]
     fn test_determinism() {
         // We use slow and lossy links as the deterministic test
@@ -1020,13 +1020,13 @@ mod tests {
         })
     }
 
-    #[test_group("heavy")]
+    #[test_group("slow")]
     #[test_traced]
     fn test_1k_min_pk() {
         run_1k::<MinPk>();
     }
 
-    #[test_group("heavy")]
+    #[test_group("slow")]
     #[test_traced]
     fn test_1k_min_sig() {
         run_1k::<MinSig>();
--- a/consensus/src/simplex/mod.rs
+++ b/consensus/src/simplex/mod.rs
@@ -1007,7 +1007,7 @@ mod tests {
         }
     }
 
-    #[test_group("heavy")]
+    #[test_group("slow")]
     #[test_traced]
     fn test_unclean_shutdown() {
         unclean_shutdown(bls12381_threshold::<MinPk, _>);
@@ -2103,7 +2103,7 @@ mod tests {
         });
     }
 
-    #[test_group("heavy")]
+    #[test_group("slow")]
     #[test_traced]
     fn test_partition() {
         partition(bls12381_threshold::<MinPk, _>);
@@ -2269,7 +2269,7 @@ mod tests {
         slow_and_lossy_links(0, ed25519);
     }
 
-    #[test_group("heavy")]
+    #[test_group("slow")]
     #[test_traced]
     fn test_determinism() {
         // We use slow and lossy links as the deterministic test
@@ -2485,7 +2485,7 @@ mod tests {
         });
     }
 
-    #[test_group("heavy")]
+    #[test_group("slow")]
     #[test_traced]
     fn test_conflicter() {
         for seed in 0..5 {
@@ -2651,7 +2651,7 @@ mod tests {
         });
     }
 
-    #[test_group("heavy")]
+    #[test_group("slow")]
     #[test_traced]
     fn test_invalid() {
         for seed in 0..5 {
@@ -2818,7 +2818,7 @@ mod tests {
         });
     }
 
-    #[test_group("heavy")]
+    #[test_group("slow")]
     #[test_traced]
     fn test_impersonator() {
         for seed in 0..5 {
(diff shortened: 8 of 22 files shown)


.config/nextest.toml
```diff
@@ -0,0 +1,12 @@
+[profile.default]
+default-filter = 'not test(/test_heavy_.*/)'
+
+[profile.heavy]
+default-filter = 'test(/test_heavy_.*/)'
+
+[profile.all]
+default-filter = 'test(/.*/)'
+overrides = [
+    { filter = 'not test(/test_heavy_*/)', priority = -100 },
```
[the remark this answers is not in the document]
BrendanChou: Is this not a legit comment?

BrendanChou (dismissed):

.config/nextest.toml
```diff
@@ -6,7 +9,3 @@ default-filter = 'test(/test_heavy_.*/)'
 
 [profile.all]
 default-filter = 'test(/.*/)'
-overrides = [
```
patrick-ogrady: much prefer explicitly naming `test_groups` over test names ❤️

.config/nextest.toml
```diff
@@ -0,0 +1,11 @@
+[test-groups]
+heavy = { max-threads = 1 }
+
+[profile.default]
```
patrick-ogrady: do we need any of these other profiles?

patrick-ogrady: I suspect the macro you wrote relies on them?

clabby: Yep; The way that we specify which tests to run is still by regex. We're somewhat mis-using test groups here as the nextest authors intended, see https://nexte.st/docs/configuration/test-groups.

`default` is non-heavy tests, `heavy` is just heavy tests, `all` is everything.

Unfortunately there's no matching groups -> regexes 😞

macros/src/lib.rs
```diff
@@ -138,6 +140,52 @@ pub fn test_traced(attr: TokenStream, item: TokenStream) -> TokenStream {
     TokenStream::from(expanded)
 }
 
+/// Prefix a test name with a nextest filter group.
+///
+/// This renames `test_some_behavior` into `test_<group>_some_behavior`, making
```
patrick-ogrady: I don't think `test_` is always present ... is there a way to produce this filter using the expanded test macro instead 🤔

.config/nextest.toml
patrick-ogrady: If so, I think we should do something more "exotically named" as you may accidentally use `test_heavy`?

macros/src/lib.rs
clabby: We don't expect it to always be present 😄 

It'll attempt to strip `test_` (if it is there), and the new test name will be `test_<group>_<original.strip_prefix("test_")>`

.config/nextest.toml
clabby: Sure - any suggestions? (fwiw, there weren't any instances of `test_heavy` in the existing codebase.)

patrick-ogrady: `000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f_commonware_test_`?

patrick-ogrady: (I want something impossible to do by accident lol)

.github/workflows/slow.yml
```diff
@@ -56,8 +56,8 @@ jobs:
       uses: taiki-e/install-action@0c5db7f7f897c03b771660e91d065338615679f4 # v2.60.0
       with:
         tool: just@1.43.0,cargo-nextest@0.9.104
-    - name: Run ignored tests
-      run: just test ${{ matrix.flags }} --partition hash:${{matrix.partition}}/8 --verbose -- --ignored
+    - name: Run heavy tests
+      run: just test ${{ matrix.flags }} --partition hash:${{matrix.partition}}/8 --verbose --profile heavy
```
patrick-ogrady: We need to update the `AGENTS.md` file with `--profile`

.config/nextest.toml
patrick-ogrady, line 1: Got this feedback (not sure if relevant):
```
  - [P1] Procedural macro now depends on workspace-only file — /Users/patrickogrady/code/monorepo/macros/src/nextest.rs:111-135
    The new #[test_group] macro in commonware-macros reads .config/nextest.toml by walking the ancestors of CARGO_MANIFEST_DIR (lines 111‑135) and emits a compile error if the file cannot be found. This works inside this monorepo because the file lives at the workspace root, but every publishable crate (e.g. commonware-consensus, commonware-storage, etc.)
  now uses #[test_group] inside its tests while the packaged crate on crates.io does not contain .config/nextest.toml. Any downstream user who downloads one of these crates and runs cargo test will immediately receive unable to locate .config/nextest.toml … from the proc macro, making the published crate’s test suite uncompilable. The macro needs to fall back
  when the config file is absent (or the config needs to be shipped per crate) so published crates remain self-contained.
  ```

Comment (patrick-ogrady) [reply .config/nextest.toml:1]: I'd be down to just ignore the groups macro if no toml is present (if this is a fair reading)?

File: .config/nextest.toml [hunk 6]
```diff
@@ -0,0 +1,11 @@
...
+default-filter = 'not test(/test_heavy_.*/)'
+
+[profile.heavy]
```
patrick-ogrady: let's call this `slow` instead of `heavy`?

.config/nextest.toml
clabby: Makes sense. I can't remember if it'd get mad for the `cfg(test)` build when publishing, but to be on the safe side, if it can't find the file, it'll now just not modify the test name 👍

.config/nextest.toml
```diff
@@ -0,0 +1,11 @@
+[test-groups]
+slow = { max-threads = 1 }
+
+[profile.default]
+default-filter = 'not test(/test_slow_.*/)'
```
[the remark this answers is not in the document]
clabby: Updated these. Originally why I chose `heavy`; `test_slow_...` seems like it's much more plausible to be written normally.

AGENTS.md
```diff
@@ -118,8 +118,11 @@ just miri <module>::
 ### Extended Checks (before PR)
 
 ```bash
-# Long-running tests
-just test --workspace -- --ignored
+# Long-running tests only
+just test --workspace --profile slow
```
patrick-ogrady, line 122: ❤️

.github/workflows/slow.yml
```diff
@@ -56,8 +56,8 @@ jobs:
       uses: taiki-e/install-action@0c5db7f7f897c03b771660e91d065338615679f4 # v2.60.0
       with:
         tool: just@1.43.0,cargo-nextest@0.9.104
-    - name: Run ignored tests
-      run: just test ${{ matrix.flags }} --partition hash:${{matrix.partition}}/8 --verbose -- --ignored
+    - name: Run heavy tests
```
patrick-ogrady: nit: `Run slow tests`

.config/nextest.toml
```diff
@@ -0,0 +1,11 @@
...
+
+[profile.slow]
+default-filter = 'test(/test_slow_.*/)'
```
patrick-ogrady: As discussed, `__group_slow_.*`

consensus/src/aggregation/mod.rs
```diff
@@ -807,7 +807,7 @@ mod tests {
         unclean_shutdown_with_unsigned_index::<MinSig>();
     }
 
-    fn slow_and_lossy_links<V: Variant>(seed: u64) -> String {
+    fn lossy_and_slow_links<V: Variant>(seed: u64) -> String {
```
patrick-ogrady: Can revert these test names changes

macros/src/lib.rs
```diff
@@ -138,6 +141,60 @@ pub fn test_traced(attr: TokenStream, item: TokenStream) -> TokenStream {
...
```
patrick-ogrady: Make sure to change this comment when you go to `__`

macros/src/lib.rs
```diff
@@ -138,6 +141,60 @@ pub fn test_traced(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+    let group_literal = parse_macro_input!(attr as LitStr);
+
+    let group = match nextest::sanitize_group_literal(&group_literal) {
+        Ok(group) => group,
+        Err(err) => return err.to_compile_error().into(),
+    };
+    let groups = match configured_test_groups() {
+        Ok(groups) => groups,
+        Err(_) => {
+            // Don't fail the compilation if the file isn't found; just return the original input.
```
patrick-ogrady, line 169: ❤️

macros/src/lib.rs
```diff
@@ -138,6 +141,60 @@ pub fn test_traced(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+            return TokenStream::from(quote!(#input));
+        }
+    };
+
+    if let Err(err) = nextest::ensure_group_known(groups, &group, group_literal.span()) {
+        return err.to_compile_error().into();
+    }
+
+    let original_name = input.sig.ident.to_string();
+    let suffix = original_name
```
patrick-ogrady: As discussed, don't do anything fancy with test names

.config/nextest.toml
```diff
@@ -0,0 +1,11 @@
+[test-groups]
+slow = { max-threads = 1 }
```
patrick-ogrady: Do we want this to be 1? It seems we want to use as many cores as possible on whatever runner is processing a partition?

clabby: The deterministic runtime should be single-threaded anyways, and nextest runs a single process per test. Wonder if it'd actually speed anything up?

patrick-ogrady: Ah interesting. Yes, it is single-threaded.

I need to look at what that config does.

.config/nextest.toml
```diff
@@ -0,0 +1,11 @@
...
+
+[profile.default]
+default-filter = 'not test(/__slow_.*/)'
+
+[profile.slow]
+default-filter = 'test(/__slow_.*/)'
```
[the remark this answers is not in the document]
patrick-ogrady: This is handled by the `test_group` macro!

runtime/src/network/iouring.rs
```diff
@@ -370,8 +371,8 @@ mod tests {
     }
 
     #[tokio::test]
-    #[ignore]
-    async fn stress_test_trait() {
+    #[test_group("slow")]
```
patrick-ogrady, line 373: nit: put before tokio?

.config/nextest.toml
```diff
@@ -0,0 +1,11 @@
+[test-groups]
+slow = { max-threads = "num-cpus" }
```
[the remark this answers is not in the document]
clabby: We're somewhat misusing groups here; This is known.

patrick-ogrady (dismissed):

patrick-ogrady (approved):

macros/src/lib.rs
```diff
@@ -138,6 +141,51 @@ pub fn test_traced(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+        Ok(group) => group,
+        Err(err) => return err.to_compile_error().into(),
+    };
+    let groups = match configured_test_groups() {
+        Ok(groups) => groups,
+        Err(_) => {
+            // Don't fail the compilation if the file isn't found; just return the original input.
+            return TokenStream::from(quote!(#input));
+        }
+    };
```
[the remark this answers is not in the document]
clabby: This is intentional.

macros/src/lib.rs
```diff
@@ -142,7 +143,7 @@ pub fn test_traced(attr: TokenStream, item: TokenStream) -> TokenStream {
 
 /// Prefix a test name with a nextest filter group.
 ///
-/// This renames `test_some_behavior` into `test_<group>_some_behavior`, making
+/// This renames `test_some_behavior` into `__<group>_test_some_behavior`, making
```
patrick-ogrady: nit: this is wrong now

macros/src/lib.rs
```diff
@@ -162,24 +163,23 @@ pub fn test_group(attr: TokenStream, item: TokenStream) -> TokenStream {
...
-    let new_ident = if suffix.is_empty() {
-        Ident::new(&format!("test_{}", group), input.sig.ident.span())
-    } else {
-        Ident::new(
-            &format!("test_{}_{}", group, suffix),
-            input.sig.ident.span(),
-        )
-    };
+    let new_ident = Ident::new(
+        &format!("{original_name}_group_{group}_"),
```
patrick-ogrady: let's actually do `_{group}_` (I think shorter extra is better)

macros/src/lib.rs
clabby: Updated -- mb!

macros/src/lib.rs
clabby: This does fail the lint check.

e: Updated to `_{group}_`.

nextest.toml
```diff
@@ -0,0 +1,11 @@
+[test-groups]
+slow = { max-threads = "num-cpus" }
+
+[profile.default]
+default-filter = 'not test(/.*_slow_/)'
+
+[profile.slow]
+default-filter = 'test(/.*_slow_/)'
```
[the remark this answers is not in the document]
clabby: already fixed but good spot 🙇

macros/src/lib.rs
```diff
@@ -138,6 +141,51 @@ pub fn test_traced(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+        }
+    };
+
+    if let Err(err) = nextest::ensure_group_known(groups, &group, group_literal.span()) {
+        return err.to_compile_error().into();
+    }
+
+    let original_name = input.sig.ident.to_string();
+    let new_ident = Ident::new(
+        &format!("{original_name}_group_{group}_"),
```
patrick-ogrady: I think this still includes `_group_`?

clabby: 🤦 updated

macros/src/lib.rs
```diff
@@ -138,6 +141,48 @@ pub fn test_traced(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+            return TokenStream::from(quote!(#input));
+        }
+    };
+
+    if let Err(err) = nextest::ensure_group_known(groups, &group, group_literal.span()) {
+        return err.to_compile_error().into();
+    }
+
+    let original_name = input.sig.ident.to_string();
+    let new_ident = Ident::new(&format!("{original_name}_{group}_"), input.sig.ident.span());
```
patrick-ogrady, line 179: As discussed, putting `_{group}_` at the end allows for "expected" syntax when running via cargo test (if we prepended with the name it would make prohibit running via `module::test`)

patrick-ogrady: (recording for LLMs lol)

patrick-ogrady (approved):

PR #2257, opened 2025-11-17
[p2p/simulated] remove redundant sort/dedup in rebalance

MozirDmitriy:
The initial pass over `self.all_flows` iterates a `BTreeMap` in key order and pushes each completed flow ID at most once, so the completed vector produced there is already sorted and unique, making the first `sort()` and `dedup()` unnecessary. Later, the allocation phase may append additional completions (e.g., unlimited-rate assignments), so a final `sort()` preserves deterministic ordering; however, duplicates cannot occur by construction because Phase B operates only on flows that still had remaining work, while Phase A already excluded them, thus `dedup()` is redundant. This change preserves behavior and determinism while removing unnecessary work.

### DIFF
--- a/p2p/src/simulated/transmitter.rs
+++ b/p2p/src/simulated/transmitter.rs
@@ -373,9 +373,9 @@ impl<P: PublicKey> State<P> {
     /// Recomputes bandwidth allocations and collects any flows that finished in the interval.
     fn rebalance(&mut self, now: SystemTime) -> Vec<Completion<P>> {
         let mut completed = Vec::new();
-
+        let mut active: Vec<Flow<P>> = Vec::new();
         for (&flow_id, meta) in self.all_flows.iter_mut() {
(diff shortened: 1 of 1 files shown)


patrick-ogrady:
Reopening PR to bump tests

patrick-ogrady:
I think this looks good...need to think about it a tad but afaict this is a nice change.

p2p/src/simulated/transmitter.rs
```diff
@@ -389,25 +389,15 @@ impl<P: PublicKey> State<P> {
...
-
-        completed.sort();
-        completed.dedup();
-
-        let mut active: Vec<Flow<P>> = Vec::new();
-        for (&flow_id, meta) in self.all_flows.iter() {
-            if meta.remaining.is_zero() {
-                continue;
-            }
-            active.push(Flow {
```
patrick-ogrady, line 403: Was able to avoid this loop 👍

Issue #1540, opened 2025-09-03
[runtime] `create_pool` Only Works in `tokio`


patrick-ogrady:
Because `deterministic` is single-threaded, it stalls when `create_pool` is used:

https://github.com/commonwarexyz/monorepo/blob/df8437d35da1d9e9f583578d4750d8fa6d1d0120/runtime/src/utils/mod.rs#L59-L84

We should consider adding this to the runtime trait for each runtime to implement in a way that "works" (although I'm not exactly sure what we should do for runtime atm).

This blocks: #903

patrick-ogrady:
The right fix here is to probably implement a rayon-compatible thread pool that only actually blocks when there is work to do (somehow).

patrick-ogrady:
A different fix for deterministic is probably to have a virtual process here that allows us to "reschedule" when it yields: 

https://github.com/commonwarexyz/monorepo/blob/b617062d6fdf7910c0ca74898729f560296513cc/runtime/src/deterministic.rs#L885-L928

Issue #903, opened 2025-05-03
[crypography/bls12381] Migrate to using Rayon Pools


patrick-ogrady:
Migrate to using shared rayon pools instead of providing a concurrency argument. When making this work, we should create a pool trait (rather than taking the pool itself) so that we can call context.spawn_blocking at the same time (rayon's default `pool.install` blocks on the calling thread).

patrick-ogrady:
Closed by #2621

PR #2260, opened 2025-11-17
[coding] Reed-Solomon optimizations

clabby:
## Overview

Adds a few low-hanging optimizations to the `reed_solomon` coding scheme:
1. When preparing data for encoding, the data is copied as a slice rather than byte-by-byte. (godbolt: [byte-by-byte](https://godbolt.org/z/dqMf8eWfT) vs. [slice-copy](https://godbolt.org/z/YnrnEvsTx))
2. Reduces unnecessary copies in `decode`
3. Allows for (optional) parallel hashing of shards when constructing the leaves of the BMT. Hashing is by far the hottest part of `reed_solomon`'s `encode/decode` path, comprising of over 90% of the samples in a profile:
Screenshot 2025-11-17 at 1 42 39 PM

### DIFF
--- a/coding/src/lib.rs
+++ b/coding/src/lib.rs
@@ -86,13 +86,15 @@ pub struct CodecConfig {
 /// use commonware_coding::{Config, ReedSolomon, Scheme as _};
 /// use commonware_cryptography::Sha256;
 ///
+/// const CONCURRENCY: usize = 1;
+///
 /// type RS = ReedSolomon<Sha256>;
 ///
 /// let config = Config { minimum_shards: 2, extra_shards: 1 };
 /// let data = b"Hello!";
 /// // Turn the data into shards, and a commitment to those shards.
 /// let (commitment, shards) =
-///      RS::encode(&config, data.as_slice()).unwrap();
+///      RS::encode(&config, data.as_slice(), CONCURRENCY).unwrap();
 ///
 /// // Each person produces reshards, their own checked shard, and checking data
 /// // to check other peoples reshards.
@@ -113,11 +115,11 @@ pub struct CodecConfig {
 ///   checked_shards.push(RS::check(&config, &commitment, &checking_data, i as u16, reshard).unwrap())
 /// }
 ///
-/// let data2 = RS::decode(&config, &commitment, checking_data, &checked_shards[..2]).unwrap();
+/// let data2 = RS::decode(&config, &commitment, checking_data, &checked_shards[..2], CONCURRENCY).unwrap();
 /// assert_eq!(&data[..], &data2[..]);
 ///
 /// // Decoding works with different shards, with a guarantee to get the same result.
-/// let data3 = RS::decode(&config, &commitment, checking_data, &checked_shards[1..]).unwrap();
+/// let data3 = RS::decode(&config, &commitment, checking_data, &checked_shards[1..], CONCURRENCY).unwrap();
 /// assert_eq!(&data[..], &data3[..]);
 /// ```
 pub trait Scheme: Debug + Clone + Send + Sync + 'static {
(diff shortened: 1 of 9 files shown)


coding/src/lib.rs
```diff
@@ -146,6 +148,7 @@ pub trait Scheme: Debug + Clone + Send + Sync + 'static {
     /// should equal `config.minimum_shards + config.extra_shards`.
     #[allow(clippy::type_complexity)]
     fn encode(
+        concurrency: usize,
```
patrick-ogrady: in other functions with concurrency, we tend to set it as the last arg rather than the first.

coding/src/lib.rs
```diff
@@ -275,7 +280,14 @@ mod test {
             others.push(my_checked_shard);
             others
         };
-        let decoded = S::decode(&config, &commitment, checking_data, &checked_shards).unwrap();
+        let decoded = S::decode(
+            CONCURRENCY,
```
patrick-ogrady: Linking #903 (and blocker #1540)

coding/src/reed_solomon/mod.rs
```diff
@@ -197,12 +206,24 @@ fn encode_inner<H: Hasher>(total: u16, min: u16, data: Vec<u8>) -> Result<Encodi
...
-            hasher.update(shard);
-            hasher.finalize()
+    if concurrency > 1 {
+        let pool = ThreadPoolBuilder::new()
+            .num_threads(concurrency)
+            .build()
+            .expect("unable to build thread pool");
+        let shard_hashes = pool.install(|| {
+            shards
+                .par_iter()
```
patrick-ogrady, line 216: nit: is there a way to set a base var per worker (could reuse the hasher)?

coding/src/reed_solomon/mod.rs
```diff
@@ -131,40 +135,45 @@ fn prepare_data(data: Vec<u8>, k: usize, m: usize) -> Vec<Vec<u8>> {
...
+    data.take(data_len).copied().collect()
 }
 
 /// Type alias for the internal encoding result.
 type Encoding<H> = (bmt::Tree<H>, Vec<Vec<u8>>);
 
 /// Inner logic for [encode()]
-fn encode_inner<H: Hasher>(total: u16, min: u16, data: Vec<u8>) -> Result<Encoding<H>, Error> {
+fn encode_inner<H: Hasher>(
+    concurrency: usize,
```
patrick-ogrady: nit: `concurrency` last?

coding/src/reed_solomon/mod.rs
```diff
@@ -327,25 +355,38 @@ fn decode<H: Hasher>(
...
+            .build()
+            .expect("unable to build thread pool");
+        let shard_hashes = pool.install(|| {
+            shards
+                .par_iter()
+                .map_init(H::new, |hasher, shard| {
+                    hasher.update(shard);
+                    hasher.finalize()
+                })
+                .collect::<Vec<_>>()
```
patrick-ogrady, line 377: do we need to `collect()` here? can we just populate the builder on the iterator?

coding/src/reed_solomon/mod.rs
```diff
@@ -197,12 +206,27 @@ fn encode_inner<H: Hasher>(total: u16, min: u16, data: Vec<u8>) -> Result<Encodi
...
```
cronokirby, line 221: Could we do a for each to add them all in the builder instead, or a fold?

patrick-ogrady: had the same thought

cronokirby (commented):
Only nits are related to materializing shard hashes in memory unnecessarily.

coding/src/reed_solomon/mod.rs
clabby: We do; There's no way to convert a parallel iterator into a serial iterator without heap allocation (that I know of.) Could maybe do it with channels, but that'd probably be worse?

PR #2262, opened 2025-11-17
[consensus/simplex] Resolver Nits

patrick-ogrady:


resolver/src/p2p/fetcher.rs
```diff
@@ -197,11 +197,9 @@ impl<E: Clock + GClock + Rng + Metrics, P: PublicKey, Key: Span, NetS: Sender<Pu
...
         }
 
-        // If there is something, ensure the waiter is respected first
-        if let Some(waiter) = self.waiter {
-            return Some(waiter);
-        }
-        self.peek_pending().map(|(deadline, _)| deadline)
+        // Return the greater of the waiter and the next pending deadline
+        let pending_deadline = self.peek_pending().map(|(deadline, _)| deadline);
+        pending_deadline.max(self.waiter)
```
BrendanChou, line 202: `Option::max` so fancy

p2p/src/utils/mod.rs
```diff
@@ -33,11 +33,12 @@ impl<P: PublicKey> Manager for StaticManager<P> {
     type PublicKey = P;
     type Peers = Ordered<P>;
 
-    async fn update(&mut self, _: u64, peers: Ordered<P>) {
-        self.peers = peers;
+    async fn update(&mut self, _: u64, _: Ordered<P>) {
+        unreachable!("updates are not supported");
```
BrendanChou, line 37: nit: this is technically reachable if someone uses this incorrectly right? or is this type not public outside the crate? If reachable, then probably `panic!` is more appropriate

patrick-ogrady: Yeah, you're right

consensus/src/simplex/actors/resolver/state.rs
```diff
@@ -77,7 +77,10 @@ impl<S: Scheme, D: Digest> State<S, D> {
                 }
 
                 // Set last finalized
-                if self.floor.as_ref().is_none_or(|floor| view > floor.view()) {
+                if self.floor.as_ref().is_none_or(|floor| {
+                    (matches!(floor, Voter::Notarization(_)) && view == floor.view())
+                        || view > floor.view()
+                }) {
                     self.floor = Some(Voter::Finalization(finalization));
                 }
```
BrendanChou, line 85:
```suggestion
// Overwrite lesser-view floors, or equal-view floors if they are only notarizations
if match self.floor.as_ref() {
    None => true,
    Some(floor) => {
        match view.cmp(&floor.view()) {
            Ordering::Greater => true,
            Ordering::Equal => matches!(floor, Voter::Notarization(_)),
            Ordering::Less => false,
        }
    }
} {
    self.floor = Some(Voter::Finalization(finalization));
}
```

PR #2265, opened 2025-11-18
[consensus/marshal] don't track outstanding resolver requests

andresilva:
These were not being used consistently, e.g. notarized requests weren't accounted for at all, and [here](https://github.com/commonwarexyz/monorepo/blob/main/consensus/src/marshal/actor.rs#L471) this block request wasn't checked for in-flight. The resolver already keeps track of pending requests so this seems unnecessary to me.

### DIFF
--- a/consensus/src/marshal/actor.rs
+++ b/consensus/src/marshal/actor.rs
@@ -1007,10 +1007,9 @@ impl<
             // exist. If not, we rely on the recursive digest fetch above.
             let gap_end = std::cmp::max(cursor.height(), gap_start);
             debug!(gap_start, gap_end, "requesting any finalized blocks");
-            let mut requests = Vec::new();
-            for height in gap_start..gap_end {
-                requests.push(Request::<B>::Finalized { height });
-            }
+            let requests = (gap_start..gap_end)
+                .map(|height| Request::<B>::Finalized { height })
+                .collect();
             resolver.fetch_all(requests).await;
         }
     }
--- a/resolver/src/lib.rs
+++ b/resolver/src/lib.rs
@@ -38,10 +38,10 @@ pub trait Resolver: Clone + Send + 'static {
     /// Type used to uniquely identify data.
     type Key: Span;
 
-    /// Initiate a fetch request for a batch of keys.
+    /// Initiate a fetch request for a single key.
     fn fetch(&mut self, key: Self::Key) -> impl Future<Output = ()> + Send;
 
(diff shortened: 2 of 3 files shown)


consensus/src/marshal/actor.rs
```diff
@@ -1017,11 +1007,11 @@ impl<
             // exist. If not, we rely on the recursive digest fetch above.
             let gap_end = std::cmp::max(cursor.height(), gap_start);
             debug!(gap_start, gap_end, "requesting any finalized blocks");
+            let mut requests = Vec::new();
             for height in gap_start..gap_end {
-                if self.waiting_finalized.insert(height) {
-                    resolver.fetch(Request::<B>::Finalized { height }).await;
-                }
+                requests.push(Request::<B>::Finalized { height });
             }
```
clabby:
```suggestion
            let requests = (gap_start..gap_end).map(|height| Request::<B>::Finalized { height }).collect()
```

resolver/src/lib.rs
```diff
@@ -38,9 +38,12 @@ pub trait Resolver: Clone + Send + 'static {
     /// Type used to uniquely identify data.
     type Key: Span;
 
-    /// Initiate a fetch request.
+    /// Initiate a fetch request for a batch of keys.
```
andresilva:
```suggestion
    /// Initiate a fetch request for the given key.
```

resolver/src/lib.rs
```diff
@@ -38,9 +38,12 @@ pub trait Resolver: Clone + Send + 'static {
...
     fn fetch(&mut self, key: Self::Key) -> impl Future<Output = ()> + Send;
 
+    /// Initiate a fetch request for all keys.
```
andresilva:
```suggestion
    /// Initiate a fetch request for a batch of keys.
```

resolver/src/p2p/ingress.rs
```diff
@@ -41,11 +41,21 @@ impl<K: Span> Resolver for Mailbox<K> {
...
     }
 
+    /// Send a fetch request to the peer actor for all keys.
+    ///
+    /// Panics if the send fails.
+    async fn fetch_all(&mut self, keys: Vec<Self::Key>) {
+        self.sender
+            .send(Message::Fetch { keys })
+            .await
+            .expect("Failed to send fetch_batch");
```
andresilva:
```suggestion
            .expect("Failed to send fetch_all");
```

PR #2268, opened 2025-11-18
[consensus/application/marshaled] Validate block ancestry

andresilva:
The `Marshaled` verifier could panic when processing byzantine blocks with invalid ancestry. A malicious proposer could broadcast a block with non-contiguous heights (e.g., claiming to build on height 21 while being at height 35). When the verifier fetched both blocks and passed them to `AncestorStream::new`, it would trigger an assertion failure.

### DIFF
--- a/consensus/src/marshal/mod.rs
+++ b/consensus/src/marshal/mod.rs
@@ -1140,12 +1140,8 @@ mod tests {
         })
     }
 
-    /// Test that Marshaled verifier validates block ancestry before verification.
-    ///
-    /// This test ensures that byzantine blocks with invalid ancestry (non-contiguous heights
-    /// or mismatched parent commitments) are rejected early.
     #[test_traced("WARN")]
-    fn test_marshaled_app_rejects_invalid_ancestry() {
+    fn test_marshaled_rejects_invalid_ancestry() {
         #[derive(Clone)]
         struct MockVerifyingApp {
             genesis: B,
@@ -1175,6 +1171,7 @@ mod tests {
                 _context: (deterministic::Context, Self::Context),
                 _ancestry: AncestorStream<Self::SigningScheme, Self::Block>,
             ) -> bool {
+                // Ancestry verification occurs entirely in `Marshaled`.
                 true
             }
         }
--- a/examples/reshare/src/application/core.rs
+++ b/examples/reshare/src/application/core.rs
@@ -100,9 +100,9 @@ where
         _: (E, Self::Context),
         _: AncestorStream<Self::SigningScheme, Self::Block>,
     ) -> bool {
-        // We will wrap this application with `Marshaled`, which handles
-        // ancestry verification (parent commitment and height contiguity),
-        // hence there is nothing to verify here.
+        // We wrap this application with `Marshaled`, which handles ancestry
+        // verification (parent commitment and height contiguity), hence there is
+        // nothing to verify here.
         true
     }
 }


consensus/src/marshal/mod.rs
```diff
@@ -1138,4 +1139,119 @@ mod tests {
             });
         })
     }
+
+    /// Test that Marshaled verifier validates block ancestry before verification.
+    ///
+    /// This test ensures that byzantine blocks with invalid ancestry (non-contiguous heights
+    /// or mismatched parent commitments) are rejected early.
+    #[test_traced("WARN")]
+    fn test_marshaled_app_rejects_invalid_ancestry() {
```
andresilva: I wrote this beforehand to test the panic. Can rewrite it or move it elsewhere.

consensus/src/application/marshaled.rs
```diff
@@ -347,6 +347,28 @@ where
                     return;
                 }
 
+                // Validate that the block's parent commitment matches what consensus expects.
+                if block.parent() != parent.commitment() {
```
patrick-ogrady, line 367: This is checked in the application: https://github.com/commonwarexyz/monorepo/blob/8bde089e174ba0ab729f8f7b2ffdc9eda500b0fa/examples/reshare/src/application/core.rs#L103-L109

patrick-ogrady: (at least in our usage)

clabby: The issue is that `AncestorStream::new` currently asserts that the blocks are contiguous: https://github.com/commonwarexyz/monorepo/blob/HEAD/consensus/src/marshal/ingress/mailbox.rs#L340-L347

andresilva: Yes, but we crash on creating the `AncestorStream`, so before that verification.

patrick-ogrady: I wonder if it would make sense to put this in the AncestorStream itself (I think you may need to do this check for all blocks read)?

I agree it is a bit deceptive it isn't done for you.

patrick-ogrady: nvm, the parent should already have a notarization

patrick-ogrady: Ah, missed the assertion check

andresilva:
Updated marshaled docs and removed validation from examples/reshare.

clabby (approved):

consensus/src/application/marshaled.rs
```diff
@@ -52,11 +52,19 @@ use rand::Rng;
...
 ///
-/// This wrapper intercepts consensus operations to enforce epoch boundaries. It prevents
-/// blocks from being produced outside their valid epoch and handles the special case of
-/// re-proposing boundary blocks during epoch transitions.
+/// This wrapper intercepts consensus operations to enforce epoch boundaries and validate
+/// block ancestry. It prevents blocks from being produced outside their valid epoch,
+/// handles the special case of re-proposing boundary blocks during epoch transitions,
+/// and ensures all blocks have valid parent linkage and contiguous heights.
+///
+/// # Ancestry Validation
```
patrick-ogrady, line 62: Does it make sense to note only verifying the parent is sufficient because it must be notarized (here or inline)?

examples/reshare/src/application/core.rs
```diff
@@ -98,14 +98,8 @@ where
...
+        _: AncestorStream<Self::SigningScheme, Self::Block>,
     ) -> bool {
-        let Some(block) = ancestry.next().await else {
-            return false;
-        };
-        let Some(parent) = ancestry.next().await else {
-            return false;
-        };
-        block.height() == parent.height() + 1 && block.parent() == parent.digest()
+        true
```
patrick-ogrady, line 106: nit: add a comment that ancestry verification enforced elsewhere (this is an example)

consensus/src/marshal/mod.rs
```diff
@@ -1138,4 +1139,119 @@ mod tests {
...
+            // 2. Fetch malicious_block (height 35) from marshal based on digest
+            // 3. Validate height is contiguous
+            // 4. Return false
+            let verify = marshaled
+                .verify(byzantine_context, malicious_commitment)
+                .await;
+
+            let result = verify.await.unwrap();
+            assert!(
+                !result,
```
patrick-ogrady: nit: `!verify.await.unwrap()`?

consensus/src/marshal/mod.rs
patrick-ogrady: This is fine here 👍

patrick-ogrady (approved):

PR #2269, opened 2025-11-18
[consensus/simplex] cap floor to the given view

andresilva:


patrick-ogrady:
We now try to align validators around the highest notarization/finalization rather than ensuring they have any particular historical view (as all that is required to vote on a new proposal is some parent notarization/finalization and nullifications from then to the current view).

If a new certificate is processed (such that view 4 is no longer the latest notarization/finalization) and the leader's proposal built on view 4 is nullified, it is most likely to push consensus forward if they help to get the new certificate distributed through the network (not view 4) as it means less nullifications to distribute between said view and the current view.

TL;DR what I did was intentional and I think it is optimal (although the case where it actually provides a benefit is very very rare -> by the time nullification is observed it is unlikely you got a view higher than what you built on unless you're behind).

andresilva:
There was also something else that I missed, which is that the resolver floor is only moved with notarization/finalization.

Issue #1706, opened 2025-09-24
[marshal] Allow starting from state-sync


BrendanChou:
Right now `marshal` requires full-sync from genesis. Instead, allow for `marshal` to start from a "floor" other than genesis (height `0`).

The typical use-case is that a node comes online, state-syncs to a recent block, and then `marshal` should full-sync from that block height to get to the tip (rather than full-syncing from genesis).

Remember that this means that `marshal` is no longer guaranteed to have every block since genesis

patrick-ogrady:
The `init` function we add could just accept a finalization from some height (although that may be overkill because the person instantiating marshal should be "trusted").

SuperFluffy:
> The init function we add could just accept a finalization from some height (although that may be overkill because the person instantiating marshal should be "trusted").

The difference between stuffing things into `init` vs putting it into `config` is not always clear with the various actors. I usually prefer to see things in `config`, because that feels more natural.

Issue #2048, opened 2025-10-29
[consensus] Don't want for `Finalizer` to backfill more gaps


clabby:
## Overview

Marshal currently waits for the `Finalizer` to inform it that it has a gap before performing a backfill:

https://github.com/commonwarexyz/monorepo/blob/922d05a3afe4d7451eb68c94b3f992df8f59f260/consensus/src/marshal/actor.rs#L488-L494
https://github.com/commonwarexyz/monorepo/blob/922d05a3afe4d7451eb68c94b3f992df8f59f260/consensus/src/marshal/finalizer.rs#L108-L111

This is sub-optimal, since marshal already has all of the information to discover the gap(s) that it needs to repair. We could improve sync speed here by:
1. Directly consulting the `finalized_blocks` archive's `Ordinal`, which can give us an iterator over _all_ of the gaps in the archive.
2. Not relying on the `Finalizer` to drive this process, and instead drive from another thread that isn't concerned with the execution.

patrick-ogrady:
**GREAT** idea 🤦 . Archive's `Ordinal` should provide everything we need to do this already.

BrendanChou:
I agree with this. Thanks for making the issue.

Ultimately I think we can deprecate the finalizer actor. The sending of the finalized blocks to the reporter could be done by the engine itself, and awaiting response/acknowledgement can be done by the select in the main loop

Issue #1651, opened 2025-09-18
[marshal] Support External Block Store


patrick-ogrady:
Instead of mandating that `marshal` persists all blocks for the application, provide an interface that can be used to query "immutable" block data from the application by digest/height.

For execution environments tightly coupled to block storage, this can make integration much more efficient (avoids duplicate storage of blocks).

PR #2271, opened 2025-11-18
[consensus/marshal] Sync height floor

clabby:
## Overview

Adds the ability for marshal to start at a pre-defined block height rather than always forcing syncing from genesis.

closes https://github.com/commonwarexyz/monorepo/issues/1706

### DIFF
--- a/consensus/src/marshal/actor.rs
+++ b/consensus/src/marshal/actor.rs
@@ -47,8 +47,7 @@ use rand::{CryptoRng, Rng};
 use std::{
     collections::{btree_map::Entry, BTreeMap},
     future::Future,
-    num::NonZeroU64,
-    ops::Range,
+    num::NonZeroUsize,
     time::Instant,
 };
 use tracing::{debug, error, info, warn};
@@ -120,7 +119,7 @@ pub struct Actor<
     // Minimum number of views to retain temporary data after the application processes a block
     view_retention_timeout: ViewDelta,
     // Maximum number of blocks to repair at once
-    max_repair: NonZeroU64,
+    max_repair: NonZeroUsize,
     // Codec configuration for block type
     block_codec_config: B::Cfg,
 
@@ -535,24 +534,16 @@ impl<
                         Message::SetFloor { height } => {
                             if let Some(stored_height) = self.application_metadata.get(&LATEST_KEY) {
                                 if *stored_height >= height {
-                                    warn!(height, "sync floor not updated, lower than existing");
+                                    warn!(height, existing = stored_height, "sync floor not updated, lower than existing");
                                     continue;
                                 }
                             }
 
-                            if let Err(e) = self
-                                .application_metadata
-                                .put_sync(LATEST_KEY.clone(), height)
-                                .await
-                            {
-                                error!(?e, height, "failed to update sync floor");
+                            // Update the processed height
+                            if let Err(err) = self.set_processed_height(height, &mut resolver).await {
+                                error!(?err, height, "failed to update sync floor");
                                 return;
                             }
-                            self.last_processed_height = height;
-                            let _ = self.processed_height.try_set(self.last_processed_height);
-
-                            // Cancel any existing requests below the new sync floor.
-                            resolver.retain(Request::<B>::Finalized { height }.predicate()).await;
 
                             // Drop the pending acknowledgement, if one exists. We must do this to prevent
                             // an in-process block from being processed that is below the new sync floor
@@ -795,12 +786,10 @@ impl<
         commitment: B::Commitment,
         resolver: &mut impl Resolver<Key = Request<B>>,
     ) -> Result<(), metadata::Error> {
-        let _ = self.processed_height.try_set(height);
-        self.last_processed_height = height;
-        self.application_metadata
-            .put_sync(LATEST_KEY.clone(), height)
-            .await?;
+        // Update the processed height
+        self.set_processed_height(height, resolver).await?;
 
+        // Cancel any useless requests
         resolver.cancel(Request::<B>::Block(commitment)).await;
 
         if let Some(finalization) = self.get_finalization_by_height(height).await {
@@ -979,30 +968,26 @@ impl<
         resolver: &mut impl Resolver<Key = Request<B>>,
         application: &mut impl Reporter<Activity = Update<B, A>>,
     ) {
-        let gaps = self.identify_gaps();
-
-        // Cancel any outstanding requests by height that sit outside of identified gaps.
-        // These requests are no longer needed, and clog up the resolver.
-        let predicate = {
-            let gaps = gaps.clone();
-            move |height: &u64| gaps.iter().any(|range| range.contains(height))
-        };
-        resolver
-            .retain(move |key| match key {
-                Request::Finalized { height } => predicate(height),
-                _ => true,
-            })
-            .await;
-
-        for Range { start, end } in gaps {
-            // Attempt to repair the gap backwards from the end of the gap, using
-            // blocks from our local storage.
-            let Some(mut cursor) = self.get_finalized_block(end).await else {
-                panic!("gapped block missing that should exist: {end}");
+        let start = self.last_processed_height.saturating_add(1);
+        'cache_repair: loop {
+            let (gap_start, Some(gap_end)) = self.finalized_blocks.next_gap(start) else {
+                // No gaps detected
+                return;
             };
 
+            // Attempt to repair the gap backwards from the end of the gap, using
+            // blocks from our local storage.
+            let Some(mut cursor) = self.get_finalized_block(gap_end).await else {
+                panic!("gapped block missing that should exist: {gap_end}");
+            };
+
+            // Compute the lower bound of the recursive repair. `gap_start` is `Some`
+            // if `start` is not in a gap. We add one to it to ensure we don't
+            // re-persist it to the database in the repair loop below.
+            let gap_start = gap_start.map(|s| s.saturating_add(1)).unwrap_or(start);
+
             // Iterate backwards, repairing blocks as we go.
-            while cursor.height() > start {
+            while cursor.height() > gap_start {
                 let commitment = cursor.parent();
                 if let Some(block) = self.find_block(buffer, commitment).await {
                     let finalization = self.cache.get_finalization_for(commitment).await;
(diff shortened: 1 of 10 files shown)


consensus/src/marshal/actor.rs
```diff
@@ -534,6 +571,37 @@ impl<
...
+                                return;
+                            }
+
+                            // Cancel any existing requests below the new sync floor.
+                            resolver.retain(
+                                Request::<B>::Finalized { height: height.saturating_sub(1) }.predicate()
+                            ).await;
+
+                            // Update the last processed height, if it is below the new sync floor.
+                            if self.last_processed_height < height {
```
clabby: Note that this update can cause the application to receive a "skipped" block, i.e. it breaks the contiguous stream assumption.

We should assume that the application would trigger any sync height floor updates and therefore know to handle it, but curious if we should add a new `Update` variant that informs the app of the sync height floor being changed?

consensus/src/marshal/ingress/mailbox.rs
```diff
@@ -109,6 +109,14 @@ pub(crate) enum Message<S: Scheme, B: Block> {
         /// The verified block.
         block: B,
     },
+    /// A request to set the sync floor, which is the minimum height to sync from.
+    ///
+    /// This sets the sync floor only if the provided height is higher than the
+    /// previously recorded floor.
```
BrendanChou: Maybe make this even more clear what "floor" means (and maybe even that 0 is the default floor?). Also maybe put a similar comment on `fn set_sync_floor()`

utils/src/acknowledgement.rs
```diff
@@ -207,4 +208,18 @@ mod tests {
         drop(ack2);
         assert!(waiter.now_or_never().unwrap().is_err());
     }
+
+    #[test]
+    fn dropping_waiter_does_not_interfere_with_acknowledgement() {
```
BrendanChou, line 213: is this related? 🤠

consensus/src/marshal/actor.rs
```diff
@@ -534,6 +532,33 @@ impl<
...
+                                .await
+                            {
+                                error!(?e, height, "failed to update sync floor");
+                                return;
+                            }
+                            self.last_processed_height = height;
+                            let _ = self.processed_height.try_set(self.last_processed_height);
+
+                            // Cancel any existing requests below the new sync floor.
+                            resolver.retain(Request::<B>::Finalized { height }.predicate()).await;
```
BrendanChou: makes sense, how come we dont call `resolver.retain(Request::<B>::Finalized { height }.predicate()).await;` anywhere else in this file? Is that a bug?

BrendanChou: Ideally we could reuse similar code to `self.handle_block_processed(height)` to share logic there, but I guess firstly we don't have the block commitment, and secondly we know we might be skipping blocks and therefore have to do extra stuff like setting the `pending_ack` back to `None`

consensus/src/marshal/mod.rs
```diff
@@ -40,16 +40,17 @@
...
+//! sync from a specific height rather than from genesis.
 //!
 //! ## Limitations and Future Work
 //!
 //! - Only works with [crate::simplex] rather than general consensus.
 //! - Assumes at-most one notarization per view, incompatible with some consensus protocols.
-//! - No state sync supported. Will attempt to sync every block in the history of the chain.
-//! - Stores the entire history of the chain, which requires indefinite amounts of disk space.
+//! - Does not prune blocks that exist in the immutable archive below the sync floor. Although these
+//!   blocks are no longer needed, the structure does not allow for it.
```
BrendanChou: retain mention that "this requires indefinite amounts of disk space" due to the no-pruning

consensus/src/marshal/actor.rs
```diff
@@ -534,6 +532,33 @@ impl<
                                 }
                             }
                         }
+                        Message::SetSyncFloor { height } => {
+                            if let Some(m_height) = self.application_metadata.get(&LATEST_KEY) {
+                                if *m_height >= height {
+                                    debug!(height, "sync floor not updated, lower than existing");
```
patrick-ogrady: nit: let's make this a `warn`

utils/src/acknowledgement.rs
clabby: It is! We now drop the `pending_ack` to prevent an in-progress block that's being processed from updating the processed height below the new floor.

Just wanted to add coverage that shows the application (and any other receivers of the ack handle) could still call `acknowledge` if marshal has dropped the waiter without issue.

consensus/src/marshal/ingress/mailbox.rs
```diff
@@ -282,6 +290,18 @@ impl<S: Scheme, B: Block> Mailbox<S, B> {
             error!("failed to send verified message to actor: receiver dropped");
         }
     }
+
+    /// A request to set the sync floor (conditionally advances if higher).
+    pub async fn set_sync_floor(&mut self, height: u64) {
```
patrick-ogrady: nit: `set_floor`?

I feel like including `sync` implies there are multiple floors?

consensus/src/marshal/actor.rs
```diff
@@ -262,7 +259,8 @@ impl<
         )
         .await
         .expect("failed to initialize application metadata");
-        let last_processed_height = *application_metadata.get(&LATEST_KEY).unwrap_or(&0);
+
```
patrick-ogrady: nit: remove `\n`

consensus/src/marshal/actor.rs
```diff
@@ -534,6 +532,33 @@ impl<
                                 }
                             }
                         }
+                        Message::SetSyncFloor { height } => {
+                            if let Some(m_height) = self.application_metadata.get(&LATEST_KEY) {
```
patrick-ogrady: nit: `Some(stored_height)`

consensus/src/marshal/actor.rs
clabby: It is intentional; After #2048, we retain in-progress finalized requests upon block delivery due to filling multiple gaps at once.

consensus/src/marshal/actor.rs
```diff
@@ -1020,20 +1045,34 @@ impl<
     fn identify_gaps(&self) -> Vec<(u64, u64)> {
         let mut remaining = self.max_repair.get();
         let mut gaps = Vec::new();
-        let mut previous_end = FIRST_HEIGHT_IN_ARCHIVE.saturating_sub(1);
+
+        let last_processed = self
```
patrick-ogrady: `self.last_processed_height`?

consensus/src/marshal/mod.rs
```diff
@@ -498,6 +496,166 @@ mod tests {
...
+                    .await;
+
+                // Finalize block by all validators except for the first.
+                let fin = make_finalization(proposal, &schemes, QUORUM);
+                for actor in actors.iter_mut().skip(1) {
+                    actor.report(Activity::Finalization(fin.clone())).await;
+                }
+            }
+
+            // Check that all applications (except for the first) received all blocks.
```
patrick-ogrady, line 588: You can now wait to create `actor` until after this point but 🤷

consensus/src/marshal/mod.rs
```diff
@@ -40,16 +40,20 @@
...
+//! all blocks that were in the archive prior to updating the starting height will be
+//! retained.
 //!
 //! ## Limitations and Future Work
 //!
 //! - Only works with [crate::simplex] rather than general consensus.
 //! - Assumes at-most one notarization per view, incompatible with some consensus protocols.
-//! - No state sync supported. Will attempt to sync every block in the history of the chain.
-//! - Stores the entire history of the chain, which requires indefinite amounts of disk space.
+//! - Does not prune blocks that exist in the immutable archive below the sync floor. Although these
```
patrick-ogrady, line 54: The future work here is to just revert back to prunable archive if you plan on pruning (the interface should be the same)?

Maybe we do this as part of the finalized block refactor (making this pluggable allows the user to pick)?

patrick-ogrady: A nice motivation for making it pluggable at the same time..

consensus/src/marshal/ingress/mailbox.rs
```diff
@@ -109,10 +109,16 @@ pub(crate) enum Message<S: Scheme, B: Block> {
...
+    ///
+    /// The sync floor is the latest block that the application has processed. Marshal
+    /// will not attempt to sync blocks below this height nor deliver blocks below
+    /// this height to the application.
     ///
     /// This sets the sync floor only if the provided height is higher than the
     /// previously recorded floor.
+    ///
+    /// The default sync floor is height 0.
     SetSyncFloor {
```
BrendanChou: consistency: SetFloor (?)

consensus/src/marshal/mod.rs
patrick-ogrady: Tagging here: #1651

patrick-ogrady: Adding to staged 👍

consensus/src/marshal/mod.rs
clabby: Yeah, we could - done.

consensus/src/marshal/mod.rs
clabby: That'd be nice, but the caveat to using prunable is a memory blow-up if the intervals between pruning are large, yeah?

utils/src/acknowledgement.rs
patrick-ogrady: Nice

consensus/src/marshal/mod.rs
patrick-ogrady: Yeah, but its not that bad. I think its like ~16B per key (and it then becomes a single disk lookup)

patrick-ogrady: This was further improved by Roberto's recent changes with the partitioning.

consensus/src/marshal/ingress/mailbox.rs
```diff
@@ -109,6 +109,20 @@ pub(crate) enum Message<S: Scheme, B: Block> {
...
+    ///
+    /// The sync floor is the latest block that the application has processed. Marshal
+    /// will not attempt to sync blocks below this height nor deliver blocks below
+    /// this height to the application.
+    ///
+    /// This sets the sync floor only if the provided height is higher than the
+    /// previously recorded floor.
+    ///
+    /// The default sync floor is height 0.
+    SetSyncFloor {
```
patrick-ogrady: Should match the function name (`SyncFloor`)?

consensus/src/marshal/actor.rs
```diff
@@ -1020,20 +1044,29 @@ impl<
     fn identify_gaps(&self) -> Vec<(u64, u64)> {
```
BrendanChou: wait a minute we can just use this! https://github.com/commonwarexyz/monorepo/blob/40088062eff4cfcc32efbe1bb4b305c6d5f9e18f/storage/src/ordinal/storage.rs#L323

consensus/src/marshal/actor.rs
```diff
@@ -1017,23 +1039,35 @@ impl<
     /// Identifies one or more of the earliest gaps in the finalized blocks archive. The gaps
     /// returned are half-open ranges, where `start` is inclusive and `end` is exclusive. The total
     /// number of missing heights covered by the returned gaps is bounded by `self.max_repair`.
```
BrendanChou: Edit comment to note that the last gap returned is a sub-gap which is the latest part of the actual gap (?)

consensus/src/marshal/actor.rs
```diff
@@ -1017,23 +1039,35 @@ impl<
...
                 let take = gap_size.min(remaining);
-                let gap_start = range_start.saturating_sub(take);
-                gaps.push((gap_start, range_start));
-                remaining -= take;
+                if take > 0 {
+                    let gap_start = clamped_start.saturating_sub(take);
+                    gaps.push(Range {
+                        start: gap_start,
+                        end: clamped_start,
+                    });
```
BrendanChou: `gaps.push((gap_start..clamped_start));`

consensus/src/marshal/actor.rs
BrendanChou:
```suggestion
        // Ignore ranges below the last processed height
        let mut start = self.last_processed_height.saturating_add(1);

        // The remaining number of items to repair
        let mut rem = self.max_repair.get();

        // All gaps must be before a range, so we iterate over the ranges and collect the gaps.
        // Remember that (s, e) are the start and end of the next full range, but we want to collect the gaps.
        let gaps = Vec::new();
        for ((s, e)) in self.finalized_blocks.ranges() {
            // Break if finished
            if rem == 0 {
                break;
            }

            // Skip if range is below where we start
            if s <= start {
                continue;
            } 
            
            // The end of the gap is equal to the beginning of the filled range
            let end = s;
            assert!(end > start);

            // Bound the size of the gap by the remaining amount
            let len = (end - start).min(rem);
            rem -= len;

            // Prefer the elements at the end of the range
            gaps.push((end - len)..end);

            // The start of the next gap is equal to the index after the end of the filled range
            start = start.max(e.saturating_add(1));
        });
        gaps
```

BrendanChou: You can also do this in a loop instead of a `filter_map` as you do today, just the logic is slightly different and you have to take care. My main issue with the existing code is that there are a few unnecessary variables and calculations that obscure whats going on. Instead of listing them I just provide the example above

consensus/src/marshal/actor.rs
```diff
@@ -1017,28 +1039,39 @@ impl<
...
 
-            previous_end = range_end;
+            // The end of the gap is equal to the beginning of the filled range
+            let end = s;
+            assert!(end > start);
+            // Bound the size of the gap by the remaining amount
+            let len = (end - start).min(rem);
+            rem -= len;
+            // Prefer the elements at the end of the range
+            gaps.push((end - len)..end);
```
BrendanChou:
```suggestion
            // Found a filled range that starts above the start of our gap
            if s > start {
                // The end of the gap is equal to the beginning of the filled range
                let end = s;
                assert!(end > start);
                // Bound the size of the gap by the remaining amount
                let len = (end - start).min(rem);
                rem -= len;
                // Prefer the elements at the end of the range
                gaps.push((end - len)..end);
            }
```

consensus/src/marshal/actor.rs
```diff
@@ -1017,28 +1039,39 @@ impl<
...
     /// returned are half-open ranges, where `start` is inclusive and `end` is exclusive. The total
     /// number of missing heights covered by the returned gaps is bounded by `self.max_repair`.
-    fn identify_gaps(&self) -> Vec<(u64, u64)> {
-        let mut remaining = self.max_repair.get();
+    ///
+    /// The last gap may in the returned list be partial, i.e., it may cover fewer heights
+    /// than the full gap size, if the remaining number of heights to repair exceeds the
+    /// `max_repair` limit.
+    fn identify_gaps(&self) -> Vec<Range<u64>> {
+    // Ignore ranges below the last processed height
```
BrendanChou: indentation. Also maybe note that `start` may not initially be the start of a gap (until we update it at the end of the loop)?

consensus/src/marshal/actor.rs
```diff
@@ -534,6 +531,33 @@ impl<
                                 }
                             }
                         }
+                        Message::SetFloor { height } => {
+                            if let Some(stored_height) = self.application_metadata.get(&LATEST_KEY) {
+                                if *stored_height >= height {
+                                    warn!(height, "sync floor not updated, lower than existing");
```
BrendanChou: maybe log the `stored_height` here. nit: `existing` is perhaps better?

consensus/src/marshal/actor.rs
```diff
@@ -534,6 +531,33 @@ impl<
...
+                                .await
+                            {
+                                error!(?e, height, "failed to update sync floor");
+                                return;
+                            }
+                            self.last_processed_height = height;
+                            let _ = self.processed_height.try_set(self.last_processed_height);
+
+                            // Cancel any existing requests below the new sync floor.
+                            resolver.retain(Request::<B>::Finalized { height }.predicate()).await;
```
BrendanChou: `self::set_last_processed_height()` might be a good function that sets the value, updates the metadata, and retains the correct resolver requests

PR #2273, opened 2025-11-19
fix(sync-example): config not getting logged

samlaf:
info! was being called before the tracing context was initialized.


patrick-ogrady:
Nice find!

PR #2277, opened 2025-11-19
[deployer] avoid full HashSet clone in destroy job scheduling

sashass1315:
Replace iteration over a cloned HashSet with iter().cloned() in destroy.rs to eliminate unnecessary allocation and re-hashing. The loop body requires an owned String for the async move closure, which iter().cloned() provides without duplicating the entire set. This change aligns with how other EC2 modules iterate region sets (e.g., create.rs and update.rs), keeps all_regions intact for the second pass, and has no behavioral impact beyond reduced overhead.


clabby:

Issue #2264, opened 2025-11-18
[consensus/simplex] Avoid Certificate "Boomerang"


patrick-ogrady:
When backfilling a certificate from the `resolver`, the `voter` will immediately send the certificate back to the `resolver` (during `*_broadcast`).

While not harmful, we should avoid this unnecessary action.

PR #2278, opened 2025-11-19
[consensus/simplex] Remove Votes from `Voter`

patrick-ogrady:
Fixes: #2264


consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -589,13 +662,30 @@ impl<
             };
             timer.observe();
 
+            // Send messages to voter
```
patrick-ogrady: nit: remove duplicate

PR #2280, opened 2025-11-19
[consensus/marshal] Optimize finalization fetching

### LINKED ISSUE #2279: [consensus/marshal] Optimize gap repair
## Overview

Currently, marshal's backfill logic attempts to fill out finalizations within the back-half of the gaps it detects. Instead, it should still attempt the cache check (parent walkback,) but it should try to fill finalizations closest to the beginning of the gap.

Image

clabby:
## Overview

Fetches finalizations towards the beginning of the gap rather than the end. We'll still attempt to consult the cache towards the end of the next gap (in-case we can repair part of it,) but favor fetching finalizations towards the tip of the application.

closes #2279

### DIFF
--- a/consensus/src/marshal/actor.rs
+++ b/consensus/src/marshal/actor.rs
@@ -47,8 +47,7 @@ use rand::{CryptoRng, Rng};
 use std::{
     collections::{btree_map::Entry, BTreeMap},
     future::Future,
-    num::NonZeroU64,
-    ops::Range,
+    num::NonZeroUsize,
     time::Instant,
 };
 use tracing::{debug, error, info, warn};
@@ -120,7 +119,7 @@ pub struct Actor<
     // Minimum number of views to retain temporary data after the application processes a block
     view_retention_timeout: ViewDelta,
     // Maximum number of blocks to repair at once
-    max_repair: NonZeroU64,
+    max_repair: NonZeroUsize,
     // Codec configuration for block type
     block_codec_config: B::Cfg,
 
@@ -801,7 +800,11 @@ impl<
             .put_sync(LATEST_KEY.clone(), height)
             .await?;
 
+        // Cancel any useless requests
         resolver.cancel(Request::<B>::Block(commitment)).await;
+        resolver
+            .retain(Request::<B>::Finalized { height }.predicate())
+            .await;
 
         if let Some(finalization) = self.get_finalization_by_height(height).await {
             // Trail the previous processed finalized block by the timeout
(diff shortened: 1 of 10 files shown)


storage/src/archive/mod.rs
```diff
@@ -86,6 +86,9 @@ pub trait Archive {
     /// This is useful for driving backfill operations over the archive.
     fn next_gap(&self, index: u64) -> (Option<u64>, Option<u64>);
 
+    /// Get up to the next `max` missing items after `start`.
```
BrendanChou: update this comment, should be inclusive (?)

consensus/src/marshal/actor.rs
```diff
@@ -979,97 +978,53 @@ impl<
...
+            .finalized_blocks
+            .missing_items(next_needed, self.max_repair.get());
+
+        // If we haven't fully repaired the gap, then also request any possible
+        // finalizations for the blocks in the remaining gap. This may help
+        // shrink the size of the gap if finalizations for the requests heights
+        // exist. If not, we rely on the recursive digest fetch above.
+        let requests = missing_items
+            .into_iter()
+            .filter(|h| *h < cursor.height())
```
BrendanChou: You don't want to filter by this since you do possibly want to "spill over" into the next gap (past the cursor)

consensus/src/marshal/actor.rs
```diff
@@ -979,97 +978,53 @@ impl<
...
-        gaps
+
+        let missing_items = self
+            .finalized_blocks
+            .missing_items(next_needed, self.max_repair.get());
+
+        // If we haven't fully repaired the gap, then also request any possible
+        // finalizations for the blocks in the remaining gap. This may help
+        // shrink the size of the gap if finalizations for the requests heights
+        // exist. If not, we rely on the recursive digest fetch above.
```
BrendanChou: The implication of this comment is not quite right anymore, since we will request over multiple gaps. It's not conditional on having fully repaired the gap or not

consensus/src/marshal/actor.rs
```diff
@@ -979,97 +978,53 @@ impl<
...
+        let requests = missing_items
+            .into_iter()
+            .filter(|h| *h < cursor.height())
+            .map(|height| Request::<B>::Finalized { height })
+            .collect();
+        resolver.fetch_all(requests).await;
```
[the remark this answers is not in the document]
clabby: This isn't true. `missing_items` can span multiple gaps.

consensus/src/marshal/actor.rs
```diff
@@ -979,97 +978,53 @@ impl<
...
-        // The remaining number of items to repair
-        let mut rem = self.max_repair.get();
-        // All gaps must be before a range, so we iterate over the ranges and collect the gaps.
-        // Remember that (s, e) are the start and end of the next full range, but we want to collect the gaps.
-        let mut gaps = Vec::new();
-        for (s, _) in self.finalized_blocks.ranges() {
-            // Break if finished
-            if rem == 0 {
+        // Iterate backwards, repairing blocks as we go.
+        while cursor.height() > next_needed {
```
BrendanChou: If we end up repairing the entire gap from locally-sourced blocks, I think we should loop and call `self.finalized_blocks.next_gap()` again to find the NEXT possible gap that we can iterate over

consensus/src/marshal/mod.rs
```diff
@@ -447,7 +447,7 @@ mod tests {
                         .iter_mut()
                         .enumerate()
                     {
-                        if (do_finalize && i <= QUORUM as usize)
+                        if (do_finalize && i < QUORUM as usize)
```
clabby, line 450: This was actually a bug in the test; `QUORUM` is `3` & `NUM_VALIDATORS` is `4`, so we'd show all validators the finalization when `do_finalize` was `true` 🫣

PR #2220, opened 2025-11-13
[consensus] generalize `Scheme` to all consensus engines

andresilva:
Resolves https://github.com/commonwarexyz/monorepo/issues/1994.

### TODO
- [x] Remove `Supervisor` and `ThresholdSupervisor`
  - [x] Add `SequencersProvider` for ordered-broadcast
- [x] Cleanup macro
- [x] Generate all existing signing schemes for all engines
- [x] Tests
  - [x] Add "raw" signing scheme tests 
  - [x] Fix existing tests
    - [x] simplex
    - [x] marshal
    - [x] aggregation (make generic over scheme)
    - [x] ordered_broadcast (make generic over scheme)
- [x] Docs
- [x] Migrate certificate to `cryptography`: https://github.com/commonwarexyz/monorepo/pull/2509

PR #2287, opened 2025-11-20
[examples/reshare] Send outdated finalization directly to marshal

### LINKED ISSUE #2032: [examples/reshare] Send Outdated Finalization Directly to `Marshal`
In reshare, we assume that sending a finalization back to consensus on the `recovered` channel will ultimately lead to `marshal` receiving said finalization (but won't teardown consensus until we've finalized all blocks in the epoch).

We should consider instead manually inserting epoch boundary finalizations into `marshal` and preemptively shutting down consensus (all it will do until we shut it down after processing all blocks in the epoch is spam nullify messages).

andresilva:
This PR implements an orchestrator protocol that enables validators to request and share epoch boundary finalizations across committee rotations. When a validator receives a message from a future epoch, it requests the boundary finalization from that peer via the orchestrator channel. The peer responds with the finalization, which is validated and injected directly into marshal for processing. Consensus engine are now now shut down immediately after finalizing the boundary block.

To support verifying these finalizations after committee rotations (when validators may no longer have the full signing scheme available), the PR adds a `certificate_verifier()` method to the `SchemeProvider` trait. This provides a lightweight verifier for validating certificates without requiring active signing capabilities, enabling validators to verify old epoch finalizations even after they've transitioned to a new committee. It's expected that only providers using BLS threshold scheme will use this method, the default implementation falls back to using the regular scheme.

Closes #2032.


consensus/src/marshal/mod.rs
```diff
@@ -73,6 +73,20 @@ pub trait SchemeProvider: Clone + Send + Sync + 'static {
...
+    ///
+    /// This method allows implementations to provide a lightweight verifier
+    /// that can validate certificates without requiring full signing capabilities.
+    /// This is useful for verifying epoch-boundary finalizations after committee
+    /// rotations when the full signing scheme is no longer available.
+    ///
+    /// The default implementation delegates to `scheme`, preserving backward
+    /// compatibility for implementations that do not distinguish between
+    /// full signing schemes and certificate-only verifiers.
+    fn certificate_verifier(&self, epoch: Epoch) -> Option<Arc<Self::Scheme>> {
```
patrick-ogrady: I wonder if we should suggest this is "epoch-less" (with no fallback)?

andresilva: I considered that as well, I guess it makes the interface more "honest" in its intention. We'd still need to fall back to the epoch scheme in marshal where this is used (for non-threshold schemes).

examples/reshare/src/validator.rs
```diff
@@ -964,27 +1017,27 @@ mod test {
             let signer = signers[0].clone();
             let share = shares[0].clone();
             let public_key = signer.public_key();
-            let engine =
-                engine::Engine::<_, _, _, _, Sha256, MinSig, ThresholdScheme<MinSig>>::new(
```
andresilva, line 972: It's not obvious to me whether we were intentionally hard coding `ThresholdScheme` rather than using `S`.

examples/reshare/src/validator.rs
```diff
@@ -1123,27 +1187,27 @@ mod test {
                 }
 
                 let public_key = signer.public_key();
-                let engine =
-                    engine::Engine::<_, _, _, _, Sha256, MinSig, ThresholdScheme<MinSig>>::new(
```
andresilva, line 1131: Similar to above, this test wasn't generic over `S` at all.

examples/reshare/src/validator.rs
```diff
@@ -1379,27 +1455,27 @@ mod test {
...
```
andresilva, line 1387: Same, also not generic over  `S`.

examples/reshare/src/validator.rs
```diff
@@ -1633,27 +1720,27 @@ mod test {
                     let public_key = signer.public_key();
                     public_keys.insert(public_key.clone());
 
-                    let engine =
-                        engine::Engine::<_, _, _, _, Sha256, MinSig, ThresholdScheme<MinSig>>::new(
```
andresilva, line 1641: Same.

consensus/src/marshal/mod.rs
```diff
@@ -77,6 +77,23 @@ pub trait SchemeProvider: Clone + Send + Sync + 'static {
...
+    ///
+    /// For example, a BLS threshold scheme that maintains a constant public key across
+    /// epochs can verify certificates from any epoch using that public key, even after
+    /// the committee has rotated and the underlying secret shares have changed.
+    ///
+    /// The default implementation returns `None`. Callers should fall back to
+    /// `scheme(epoch)` for epoch-specific verification.
+    fn certificate_verifier(&self) -> Option<Arc<Self::Scheme>> {
+        None
+    }
```
clabby, line 99: We may be able to expose just one function here? It looks like all references to `scheme` are in fallback clauses to `certificate_verifier` calls; Rather than potentially messing that up, we could handle the fallback internally?

andresilva:
Can't comment directly on cursor bot's comment, but it's wrong. We shut down the engine as soon as we transition, regardless of whether it's a dkg or not.

examples/reshare/src/orchestrator/actor.rs
```diff
@@ -246,45 +267,111 @@ where
...
-                            false,
-                        )
-                        .await.is_err() {
-                            warn!("failed to forward finalization, shutting down orchestrator");
-                            break;
+                            if epoch > our_epoch {
+                                debug!(%epoch, %our_epoch, ?from, "received orchestrator request for future epoch, blocking peer");
+                                self.oracle.block(from).await;
+                                continue;
+                            }
```
clabby: non-actionable note; this is safe due to the request dispatch logic only asking peers that are broadcasting themselves as ahead, initially thought this could happen:

```
A [epoch 1] -- broadcasts --> request for boundary finalization for epoch 1
B [epoch 3] -- sees --> broadcast from A, processes it and responds
C [epoch 0] -- sees --> broadcast from A, blocks A 
```

clabby (approved):

consensus/src/marshal/mod.rs
andresilva: Currently we only use `SchemeProvider` in marshal and we only care about verification, so there's no functional difference between a full scheme and a certificate verifier. But in https://github.com/commonwarexyz/monorepo/pull/2220 I moved this interface to `consensus` so that it can be used in `aggregation` and `ordered_broadcast` (replacing the `ThresholdSupervisor`), where we do make a distinction between capabilities (i.e. signing vs verifying).

consensus/src/marshal/actor.rs
```diff
@@ -737,6 +738,12 @@ impl<
         }
     }
 
+    fn get_scheme_certificate_verifier(&self, epoch: Epoch) -> Option<Arc<S>> {
```
patrick-ogrady, line 707: nit: add comment

consensus/src/marshal/mod.rs
```diff
@@ -77,6 +77,23 @@ pub trait SchemeProvider: Clone + Send + Sync + 'static {
...
     fn scheme(&self, epoch: Epoch) -> Option<Arc<Self::Scheme>>;
+
+    /// Return a certificate verifier that can validate certificates independent of epoch.
+    ///
+    /// This method allows implementations to provide a verifier that can validate
+    /// certificates from any epoch without requiring epoch-specific signing state. This
+    /// is useful for schemes that can verify certificates using long-lived cryptographic
+    /// material that remains valid across epoch boundaries.
+    ///
+    /// For example, a BLS threshold scheme that maintains a constant public key across
```
patrick-ogrady: nit: link to scheme?

examples/reshare/src/orchestrator/actor.rs
```diff
@@ -214,6 +216,16 @@ where
                         continue;
                     }
 
+                    if let Entry::Occupied(entry) = self.finalization_requests.entry(our_epoch) {
```
patrick-ogrady: Add a comment that we don't need to worry about tracking folks because we should hear something again soon?

examples/reshare/src/orchestrator/actor.rs
```diff
@@ -222,19 +234,28 @@ where
...
+                        "received backup message from future epoch, requesting boundary finalization"
                     );
 
-                    // Send the request to the orchestrator. This operation is best-effort.
+                    // Send a request to the peer's orchestrator to get the finalization for our latest epoch.
+                    let request =
+                        wire::Message::<S, H::Digest>::Request(our_epoch);
+
+                    // Track this request.
+                    self.finalization_requests.insert(our_epoch, (from.clone(), Instant::now()));
```
patrick-ogrady: Use `SystemTime`?

examples/reshare/src/orchestrator/actor.rs
```diff
@@ -246,45 +267,111 @@ where
...
-                        .send(
-                            epoch.get(),
-                            Recipients::One(from),
-                            message.encode().freeze(),
-                            false,
-                        )
-                        .await.is_err() {
-                            warn!("failed to forward finalization, shutting down orchestrator");
-                            break;
+                            if epoch > our_epoch {
```
patrick-ogrady: nit: add comment why this can't happen?

examples/reshare/src/orchestrator/actor.rs
```diff
@@ -246,45 +267,111 @@ where
...
+                                continue;
+                            }
+
+                            debug!(
+                                %epoch,
+                                ?from,
+                                "verified requested finalization certificate, injecting finalization into marshal"
+                            );
+
+                            self.marshal.finalization(finalization).await;
```
patrick-ogrady: I suppose in the threshold case, we could really handle the latest finalization?

patrick-ogrady: We only need this for non-threshold where there is no certificate verifier?

patrick-ogrady: (although this could lead to not having finalizations at the boundary conditions)

examples/reshare/src/orchestrator/actor.rs
```diff
@@ -246,45 +267,111 @@ where
...
+                                .await
+                                .is_err()
+                            {
+                                warn!("failed to send orchestrator response, shutting down orchestrator");
+                                break;
+                            }
                         }
+                        wire::Message::Response(epoch, finalization) => {
+                            // Check if we actually requested this finalization from this peer
+                            match self.finalization_requests.entry(epoch) {
```
patrick-ogrady: If we receive a late reply, we'll block the peer. I think that's too strict.

I suggest:
```
                            match self.finalization_requests.entry(epoch) {
                                Entry::Occupied(entry) if entry.get().0 == from => {
                                    entry.remove();
                                }
                                Entry::Occupied(entry) => {
                                    debug!(
                                        %epoch,
                                        expected = ?entry.get().0,
                                        ?from,
                                        "received finalization from unexpected peer, ignoring"
                                    );
                                    continue;
                                }
                                Entry::Vacant => {
                                    debug!(
                                        %epoch,
                                        ?from,
                                        "received finalization with no pending request, ignoring"
                                    );
                                    continue;
                                }
                            }
```

examples/reshare/src/orchestrator/actor.rs
patrick-ogrady: Given we aren't permitting marshal to fully jump ahead, I wonder if we should slim this PR to avoid certificate verifier (I guess we should retain it because that's actually optimal if you don't need to support different schemes)?

examples/reshare/src/orchestrator/actor.rs
```diff
@@ -214,6 +216,16 @@ where
                         continue;
                     }
 
+                    if let Entry::Occupied(entry) = self.finalization_requests.entry(our_epoch) {
+                        if entry.get().1.elapsed() > FINALIZATION_REQUEST_TIMEOUT {
```
patrick-ogrady: I suppose the old approach led to us spamming a bunch of requests at the same time 🤔 . Feels a bit janky to have this request manager floating between the lines here...seems nice to have a utility where we register who reports they have an epoch and we just shuffle between?

patrick-ogrady (commented):
Starting to feel the tension between doing what's optimal for threshold vs non-threshold (where in threshold we could just give marshal the most recent finalization we are now just sending it the end of an epoch).

Still mulling over some ideas here (although I think only 1 thing is "wrong": blocking too aggressively).

examples/reshare/src/orchestrator/actor.rs
andresilva: `Instant` seems a better fit for this case since unlike `SystemTime` it's monotonically nondecreasing, that's why `elapsed()` on `SystemTime` returns a `Result` (unlike in `Instant`).

examples/reshare/src/orchestrator/actor.rs
```diff
@@ -305,19 +391,19 @@ where
...
+                            &mut recovered_mux,
+                            &mut resolver_mux,
+                        )
+                        .await;
                         engines.insert(transition.epoch, engine);
 
                         info!(epoch = %transition.epoch, "entered epoch");
                     }
-                    Message::Exit(epoch) => {
+                    ingress::Message::Exit(epoch) => {
```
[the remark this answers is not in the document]
andresilva: Good catch.

examples/reshare/src/orchestrator/finalization_tracker.rs
```diff
@@ -0,0 +1,155 @@
...
+        }
+
+        // Ignore duplicates
+        if self.known_peers.contains(&peer) {
+            return false;
+        }
+
+        // If no pending request, caller should send to this peer
+        if self.pending.is_none() {
+            return true;
```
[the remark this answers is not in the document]
andresilva: When `try_request` returns `true` then we will issue a request, in which case there is no point in adding the peer to `known_peers`.

examples/reshare/src/orchestrator/finalization_tracker.rs
```diff
@@ -0,0 +1,345 @@
...
+
+/// Maximum number of peers to track per epoch.
+const MAX_PEERS: usize = 10;
+
+/// Tracks finalization requests and known peers for a single epoch.
+/// Emits timeout events via a stream when requests need to be retried.
+///
+/// Only tracks one epoch at a time since we only ever request a finalization for our
+/// latest known epoch.
+pub struct FinalizationTracker<E, P> {
```
patrick-ogrady: This is dangerously close to the implementation of `resolver` (just that we limit who is a possible recipient for a given request).

I wonder if we should just add this functionality to `resolver` 🤔

patrick-ogrady: (that being said, this is definitely the mechanism I had in mind!)

patrick-ogrady: Actually, what we could do is use the resolver but just wait for at least f+1 messages in some future epoch before enqueueing it 🤔

examples/reshare/src/orchestrator/actor.rs
```diff
@@ -219,71 +222,154 @@ where
...
-                    Ok(epoch) => epoch,
+
+                // Decode the orchestrator wire message
+                let message = match wire::Message::<S, H::Digest>::read_staged(
+                    &mut bytes.as_ref(),
+                    &self.scheme_provider,
+                ) {
+                    Ok(Some(msg)) => msg,
+                    Ok(None) => {
+                        debug!(?from, "no scheme available to decode response");
```
patrick-ogrady: This is an odd case I want to think more about 👀

examples/reshare/src/orchestrator/actor.rs
```diff
@@ -219,71 +222,154 @@ where
...
                 );
 
-                // Send the request to the orchestrator. This operation is best-effort.
-                if orchestrator_sender.send(
-                    Recipients::One(from),
-                    our_epoch.encode().freeze(),
-                    true
-                ).await.is_err() {
+                // Send a request to the peer's orchestrator to get the finalization for our latest epoch.
+                let request = wire::Message::<S, H::Digest>::Request(our_epoch);
```
patrick-ogrady: nit: add a comment that we our only requesting our boundary rather than the "best"

We should link an issue here.

examples/reshare/src/orchestrator/finalization_tracker.rs
patrick-ogrady: Given the fix I recently made to the simplex resolver, I'm leaning towards not using this f+1 trick again lol.

examples/reshare/src/application/scheme.rs
```diff
@@ -54,6 +56,24 @@ impl<S: Scheme, C: Signer> SchemeProvider<S, C> {
...
+    /// and do not need to distinguish between epoch-independent and epoch-specific
+    /// verifiers.
+    ///
+    /// Returns `None` if neither an epoch-independent verifier nor an epoch-specific
+    /// scheme is available.
+    pub fn get_certificate_verifier(&self, epoch: Epoch) -> Option<Arc<S>> {
+        self.certificate_verifier
+            .clone()
+            .or_else(|| self.scheme(epoch))
+    }
```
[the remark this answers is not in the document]
andresilva: Certificates are verified against the stable identity which doesn't change with reshare.

examples/reshare/src/orchestrator/actor.rs
```diff
@@ -223,71 +267,94 @@ where
...
-                        message.encode().freeze(),
-                        false,
-                    )
-                    .await.is_err() {
-                        warn!("failed to forward finalization, shutting down orchestrator");
-                        break;
+                        // Inject finalization into marshal and clear fetch state
+                        self.marshal.finalization(finalization).await;
+                        fetching_epoch = None;
+                        finalization_resolver.clear().await;
```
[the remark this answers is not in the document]
andresilva: Epoch length is defined in blocks, not views. This is still a relevant point though, because otherwise peers can send us valid finalizations for the epoch that are not for the boundary block.

PR #2290, opened 2025-11-20
[consensus::simplex] Explicitly avoid nullify after finalize

### LINKED ISSUE #2284: [consensus::simplex] More explicitly prevent constructing a `nullify` vote if already constructed a `finalize` vote
While not currently a bug (due to assumptions and code flows outside of round), it is currently a footgun that `round` doesn't check `broadcast_finalize` before it allows construction of a `nullify`. More explicitly check this.

BrendanChou:
Renamed function to more closely match `Round::construct_notarize()` and `Round::construct_finalize()`.
Fixes #2284

### DIFF
--- a/consensus/src/simplex/actors/voter/state.rs
+++ b/consensus/src/simplex/actors/voter/state.rs
@@ -214,19 +214,17 @@ impl<E: Clock + Rng + CryptoRng + Metrics, S: Scheme, D: Digest> State<E, S, D> 
 
         // Try to construct entry certificates for the previous view
         // Prefer the strongest proof available so lagging replicas can re-enter quickly.
-        if let Some(finalization) = self.finalization(entry_view).cloned() {
-            return (retry, nullify, Some(Voter::Finalization(finalization)));
-        }
-        if let Some(notarization) = self.notarization(entry_view).cloned() {
-            return (retry, nullify, Some(Voter::Notarization(notarization)));
-        }
-        if let Some(nullification) = self.nullification(entry_view).cloned() {
-            return (retry, nullify, Some(Voter::Nullification(nullification)));
-        }
-
-        // If we couldn't find any entry certificates, return the nullify
-        warn!(%entry_view, "entry certificate not found during timeout");
-        (retry, nullify, None)
+        let cert = if let Some(finalization) = self.finalization(entry_view).cloned() {
+            Some(Voter::Finalization(finalization))
+        } else if let Some(notarization) = self.notarization(entry_view).cloned() {
+            Some(Voter::Notarization(notarization))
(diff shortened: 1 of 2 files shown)


consensus/src/simplex/actors/voter/state.rs
```diff
@@ -201,7 +201,9 @@ impl<E: Clock + Rng + CryptoRng + Metrics, S: Scheme, D: Digest> State<E, S, D>
     /// (if this is a retry timeout and we can construct one).
     pub fn handle_timeout(&mut self) -> (bool, Option<Nullify<S>>, Option<Voter<S, D>>) {
         let view = self.view;
-        let retry = self.create_round(view).handle_timeout();
+        let Some(retry) = self.create_round(view).construct_nullify() else {
```
patrick-ogrady, line 204: nit: we could comment that this shouldn't really ever happen and that we handle for completeness (as this would mean we are trying to timeout a view that isn't latest)?

patrick-ogrady: We could also still try to broadcast a cert in this case but given it shouldn't happen, am ok to just return.

patrick-ogrady: On that note, we "could" expect() on this but maybe that is too strict 🤔

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -201,7 +201,9 @@ impl<E: Clock + Rng + CryptoRng + Metrics, S: Scheme, D: Digest> State<E, S, D>
     /// (if this is a retry timeout and we can construct one).
     pub fn handle_timeout(&mut self) -> (bool, Option<Nullify<S>>, Option<Voter<S, D>>) {
```
patrick-ogrady, line 202: Wonder if `construct_nullify` makes sense here too 🤔

patrick-ogrady: I suppose you kept as `handle_timeout` because it also returns retry + cert?

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -212,19 +214,17 @@ impl<E: Clock + Rng + CryptoRng + Metrics, S: Scheme, D: Digest> State<E, S, D>
 
         // Try to construct entry certificates for the previous view
         // Prefer the strongest proof available so lagging replicas can re-enter quickly.
-        if let Some(finalization) = self.finalization(entry_view).cloned() {
```
patrick-ogrady, line 215: ❤️

patrick-ogrady (approved):

PR #2208, opened 2025-11-12
[consensus/marshal] Erasure coding

clabby:
## Overview

Revisited version of #1680

PR #2293, opened 2025-11-20
[consensus/marshal] Finalized {Blocks/Certificates} DB abstraction

### LINKED ISSUE #1651: [marshal] Support External Block Store

clabby:
## Overview

Adds an abstraction for the finalized blocks + certificates DBs in marshal. Default implementations are provided for `immutable::Archive` and `prunable::Archive`, though this opens the door to using external databases as well (i.e. deduplicating the store when using an EL that already stores blocks persistently.)

closes #1651

### DIFF
--- a/consensus/src/marshal/actor.rs
+++ b/consensus/src/marshal/actor.rs
@@ -38,7 +38,7 @@ use commonware_utils::{
     acknowledgement::Exact,
     futures::{AbortablePool, Aborter, OptionFuture},
     sequence::U64,
-    Acknowledgement,
+    Acknowledgement, BoxedError,
 };
 use futures::{
     channel::{mpsc, oneshot},
@@ -50,7 +50,6 @@ use prometheus_client::metrics::gauge::Gauge;
 use rand::{CryptoRng, Rng};
 use std::{
     collections::{btree_map::Entry, BTreeMap},
-    error::Error,
     future::Future,
     num::NonZeroUsize,
 };
@@ -498,7 +497,7 @@ where
                                 // Prune the finalized blocks archive
                                 async {
                                     self.finalized_blocks.prune(height).await.map_err(Box::new)?;
-                                    Ok::<_, Box<dyn Error + Send + Sync>>(())
+                                    Ok::<_, BoxedError>(())
                                 },
                                 // Prune the finalization certificate archive
                                 async {
@@ -506,7 +505,7 @@ where
                                         .prune(height)
                                         .await
                                         .map_err(Box::new)?;
-                                    Ok::<_, Box<dyn Error + Send + Sync>>(())
+                                    Ok::<_, BoxedError>(())
                                 }
                             ) {
                                 error!(?err, height, "failed to prune finalized archives");
@@ -856,7 +855,7 @@ where
             // Update the finalized blocks archive
             async {
                 self.finalized_blocks.put(block).await.map_err(Box::new)?;
-                Ok::<_, Box<dyn Error + Send + Sync>>(())
+                Ok::<_, BoxedError>(())
             },
             // Update the finalizations archive (if provided)
             async {
@@ -866,7 +865,7 @@ where
                         .await
                         .map_err(Box::new)?;
                 }
-                Ok::<_, Box<dyn Error + Send + Sync>>(())
+                Ok::<_, BoxedError>(())
             }
         ) {
             panic!("failed to finalize: {e}");
--- a/utils/src/lib.rs
+++ b/utils/src/lib.rs
@@ -10,7 +10,7 @@
 extern crate alloc;
 
 #[cfg(not(feature = "std"))]
-use alloc::{string::String, vec::Vec};
+use alloc::{boxed::Box, string::String, vec::Vec};
 use bytes::{BufMut, BytesMut};
 use commonware_codec::{EncodeSize, Write};
 use core::{
@@ -51,6 +51,9 @@ mod stable_buf;
 pub use stable_buf::StableBuf;
 #[cfg(feature = "std")]
 pub mod concurrency;
+
+/// Alias for boxed errors that are `Send` and `Sync`.
+pub type BoxedError = Box<dyn core::error::Error + Send + Sync>;
 
 /// Converts bytes to a hexadecimal string.
 pub fn hex(bytes: &[u8]) -> String {


consensus/src/marshal/store.rs
```diff
@@ -0,0 +1,321 @@
...
+};
+use std::{error::Error, future::Future};
+
+/// Durable store for [Finalizations](Finalization) keyed by height and commitment.
+pub trait FinalizationStore: Send + Sync + 'static {
+    /// The type of commitment used by the application's [Blocks](Block).
+    type BlockCommitment: Digest;
+
+    /// The type of commitment used by consensus.
+    type ConsensusCommitment: Digest;
```
clabby: Note to reviewer - it may not be obvious why we make this delineation, but this is for forwards-compatibility with the erasure coded variant of marshal where these two are not the same. The archive of finalizations in that variant is keyed by block digests, but consensus agrees upon a commitment that is the concatenation of the block digest, erasure coding commitment, and erasure coding configuration.

consensus/src/marshal/actor.rs
```diff
@@ -891,15 +862,19 @@ impl<
         // In parallel, update the finalized blocks and finalizations archives
         if let Err(e) = try_join!(
             // Update the finalized blocks archive
-            self.finalized_blocks.put_sync(height, commitment, block),
+            async {
```
patrick-ogrady, line 856: There has to be a cleaner way to write this...

consensus/src/marshal/config.rs
```diff
@@ -36,25 +36,6 @@ where
...
-    pub freezer_table_resize_chunk_size: u32,
-
-    /// The target size of the freezer journal.
-    pub freezer_journal_target_size: u64,
-
-    /// The compression level to use for the freezer journal.
-    pub freezer_journal_compression: Option<u8>,
-
     /// The buffer pool to use for the freezer journal.
     pub freezer_journal_buffer_pool: PoolRef,
```
patrick-ogrady: I think this should just be `buffer_pool` (if used at all)?

consensus/src/marshal/store.rs
```diff
@@ -0,0 +1,321 @@
...
+use commonware_storage::{
+    archive::{self, immutable, prunable, Archive, Identifier},
+    translator::Translator,
+};
+use std::{error::Error, future::Future};
+
+/// Durable store for [Finalizations](Finalization) keyed by height and commitment.
+pub trait FinalizationStore: Send + Sync + 'static {
+    /// The type of commitment used by the application's [Blocks](Block).
+    type BlockCommitment: Digest;
```
patrick-ogrady: The `Commitment` here should be the same for consensus/application. The digest, however, may differ?

consensus/src/marshal/actor.rs
patrick-ogrady: I know you are just trying to capture the errors but maybe we can use `anyhow::Error` (https://github.com/commonwarexyz/monorepo/blob/d0be5fc7c1d73536abd58795eef93b174f0de3d0/storage/src/journal/mod.rs#L39)?

consensus/src/marshal/actor.rs
```diff
@@ -95,13 +99,20 @@ struct BlockSubscription<B: Block> {
...
-pub struct Actor<
+pub struct Actor<E, B, P, S, FC, FB, A = Exact>
+where
     E: Rng + CryptoRng + Spawner + Metrics + Clock + GClock + Storage,
     B: Block,
     P: SchemeProvider<Scheme = S>,
     S: Scheme,
-    A: Acknowledgement = Exact,
-> {
+    FC: FinalizationStore<
```
patrick-ogrady: dumb idea: why not just `Certificates` and `Blocks`? I suppose it could be clearer if we include finalized in the trait name but this could shield the user from thinking too hard about things 🤔 

If you want to keep finalize in the name, what you have is pretty close...

patrick-ogrady (commented):
Very close!

consensus/src/marshal/actor.rs
clabby: `anyhow::Error` is just a type-erased `dyn Error` under the hood; I'd prefer not to add another dependency to `commonware-consensus` just for this unless you feel strongly

consensus/src/marshal/config.rs
clabby: Renamed! It's still used by the ephemeral `Cache` in marshal

consensus/src/marshal/store.rs
clabby: After coding, `Block` is no longer `Committable`; See #2208.

clabby: Left a comment above on this @ https://github.com/commonwarexyz/monorepo/pull/2293#discussion_r2550423532

consensus/src/marshal/actor.rs
clabby: Not super opinionated! Updated to just `Certificates` and `Blocks` for the trait names.

consensus/src/marshal/store.rs
clabby: (I also find our interchangeable use of "commitment" and "digest" a bit confusing, #2208 does a bit to clean this up)

clabby: Renamed this associated type to `BlockDigest`, hopefully that's a bit more clear 🙏

consensus/src/marshal/actor.rs
```diff
@@ -549,6 +492,26 @@ impl<
...
+                                    self.finalized_blocks.prune(height).await.map_err(Box::new)?;
+                                    Ok::<_, Box<dyn Error + Send + Sync>>(())
+                                },
+                                // Prune the finalization certificate archive
+                                async {
+                                    self.finalizations_by_height
+                                        .prune(height)
+                                        .await
+                                        .map_err(Box::new)?;
+                                    Ok::<_, Box<dyn Error + Send + Sync>>(())
```
patrick-ogrady: Can we create an alias for `Box<dyn Error + Send + Sync>` in `utils` and use it to replace `anyhow`?

consensus/src/marshal/store.rs
clabby: Reverted this for now. #2208 will make the change there.

utils/src/lib.rs
```diff
@@ -52,6 +52,9 @@ pub use stable_buf::StableBuf;
 #[cfg(feature = "std")]
 pub mod concurrency;
 
+/// Alias for boxed errors that are `Send` and `Sync`.
+pub type BoxedError = Box<dyn core::error::Error + Send + Sync>;
```
patrick-ogrady, line 56: ❤️

consensus/src/marshal/actor.rs
clabby: Added the `BoxedError` alias.

PR #2295, opened 2025-11-20
[Storage] Allow deferred merkleization in ADBs

danlaine:
* Compute MMR root digest during `merkleize()`, the dirty -> clean MMR state transition
* Store the MMR root digest in the `Clean` state
* Add deferred merkleization API to ADB implementations except `current`::*`, meaning the ADB can be manipulated without calculating the root digest between each operation
* Change `Mmr::update_leaf` parameter from `Position` to `Location`
* Split `Contiguous` implementation into `Contiguous`, `MutableContiguous`, `PersistedContiguous`
* Remove `ThreadPool` from in-memory MMR.

### DIFF
--- a/storage/src/adb/current/unordered.rs
+++ b/storage/src/adb/current/unordered.rs
@@ -116,7 +116,7 @@ impl<
         // Ensure consistency between the bitmap and the db.
         let height = Self::grafting_height();
         let inactivity_floor_loc =
-            AnyLog::<E, K, V, H, T, Clean<DigestOf<H>>>::recover_inactivity_floor(&log).await?;
+            AnyLog::<_, _, _, _, T, Clean<DigestOf<H>>>::recover_inactivity_floor(&log).await?;
         if status.len() < inactivity_floor_loc {
             // Prepend the missing (inactive) bits needed to align the bitmap, which can only be
             // pruned to a chunk boundary.
--- a/storage/src/journal/contiguous/mod.rs
+++ b/storage/src/journal/contiguous/mod.rs
@@ -165,7 +165,7 @@ pub trait MutableContiguous: Contiguous {
     }
 }
 
-pub trait PersistedContiguous: MutableContiguous {
+pub trait PersistableContiguous: MutableContiguous {
     /// Durably persist the journal but does not write all data, potentially leaving recovery
     /// required on startup.
     ///
--- a/storage/src/mmr/mem.rs
+++ b/storage/src/mmr/mem.rs
@@ -39,6 +39,7 @@ mod private {
 
 /// Trait for valid MMR state types.
 pub trait State<D: Digest>: private::Sealed + Sized {
+    /// Add the given leaf digest to the MMR, returning its position.
     fn add_leaf_digest<H: Hasher<D>>(mmr: &mut Mmr<D, Self>, hasher: &mut H, digest: D)
         -> Position;
 }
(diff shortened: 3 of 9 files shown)


storage/src/adb/current/unordered.rs
```diff
@@ -101,17 +104,19 @@ impl<
...
+            &mut hasher,
         )
         .await?;
 
         // Ensure consistency between the bitmap and the db.
-        let mut hasher = StandardHasher::<H>::new();
         let height = Self::grafting_height();
-        let inactivity_floor_loc = AnyLog::<E, K, V, H, T>::recover_inactivity_floor(&log).await?;
+        let inactivity_floor_loc =
+            AnyLog::<E, K, V, H, T, Clean<DigestOf<H>>>::recover_inactivity_floor(&log).await?;
```
roberto-bayardo: I think type inference will let you do just  `Clean<_>` if you want to shorten this (& other uses) a bit.

storage/src/mmr/mem.rs
```diff
@@ -9,36 +9,53 @@ use crate::mmr::{
...
+
 /// Sealed trait for MMR state types.
 mod private {
     pub trait Sealed {}
 }
 
 /// Trait for valid MMR state types.
-pub trait State: private::Sealed + Default {}
+pub trait State<D: Digest>: private::Sealed + Sized {
+    fn add_leaf_digest<H: Hasher<D>>(mmr: &mut Mmr<D, Self>, hasher: &mut H, digest: D)
```
roberto-bayardo, line 43: Document this here? It's a little confusing why this is the only function needed in the trait.

storage/src/mmr/mem.rs
```diff
@@ -524,38 +455,76 @@ impl<D: Digest> Mmr<D, Clean> {
     /// mutating the original, and the thread pool if any is not cloned.
     ///
     /// Runtime is Log_2(n) in the number of elements even if the original MMR is never pruned.
-    pub fn clone_pruned(&self) -> Self {
+    #[cfg(test)]
+    pub(super) fn clone_pruned(&self, hasher: &mut impl Hasher<D>) -> Self {
```
roberto-bayardo: If this isn't used anymore (and it appears it isn't) we should just get rid of it.

storage/src/journal/contiguous/mod.rs
```diff
@@ -152,15 +163,9 @@ pub trait Contiguous {
...
 
-    /// Read the item at the given position.
-    ///
-    /// # Errors
-    ///
-    /// - Returns [Error::ItemPruned] if the item at `position` has been pruned.
-    /// - Returns [Error::ItemOutOfRange] if the item at `position` does not exist.
-    fn read(&self, position: u64) -> impl std::future::Future<Output = Result<Self::Item, Error>>;
-
+pub trait PersistedContiguous: MutableContiguous {
```
roberto-bayardo: Since contiguous is the module name, would it be better just to use "Persisted" and "Mutable" for these traits?  

[Edit] I guess these extend the Contiguous trait, so maybe making the specialization clear in this way could be helpful.

roberto-bayardo: ..and should be it "Persistable" instead of "Persisted"?

storage/fuzz/fuzz_targets/mmr_operations.rs
```diff
@@ -217,9 +217,11 @@ fn fuzz(input: FuzzInput) {
                         }
 
                         let size_before = mmr.size();
-                        let root_before = mmr.root(&mut hasher);
+                        let root_before = *mmr.root();
 
-                        mmr.update_leaf(&mut hasher, pos, limited_data).unwrap();
+                        let leaf_loc =
+                            Location::try_from(pos).expect("leaf position should map to location");
+                        mmr.update_leaf(&mut hasher, leaf_loc, limited_data).unwrap();
```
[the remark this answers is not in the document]
danlaine: Bug bot is mistaken. The code is correct as is. The argument passes to `ReferenceMmr::update_leaf` is an index into `leaf_data`. This fuzz test could probably use some cleanup. I'm a bit dubious about comparing results to `ReferenceMmr`.

storage/src/mmr/mem.rs
danlaine: removed

storage/src/mmr/mem.rs
danlaine: Added a comment

storage/src/adb/current/unordered.rs
danlaine: done

storage/src/journal/contiguous/mod.rs
danlaine: Changed to `Persistable`. I kind of like including `Contiguous` as a suffix to express that it's also `Contiguous`. Maybe `Ext` would be the more rusty way to express this?

roberto-bayardo (approved):

PR #2296, opened 2025-11-21
[consensus/simplex] Fix Boomerang

patrick-ogrady:
Fixes: #2264


consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -36,6 +37,13 @@ use rand::{CryptoRng, Rng};
 use std::num::NonZeroUsize;
 use tracing::{debug, info, trace, warn};
 
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
```
patrick-ogrady: Remove this

PR #2301, opened 2025-11-21
[consensus/simplex] broadcast first then report to the application

andresilva:
In #2287 we are now shutting down the simplex engine immediately after the last block in the epoch is finalized. 

With the current ordering of broadcast/report it's possible that we shut down the engine before it has actually broadcast a certificate. This PR changes the ordering of all reports to the application to happen only after the certificate has been broadcast.

In #2287, this shaved a couple of seconds from the tests, indicating that less validators spend time going through the fallback request-response protocol to fetch epoch boundary finalizations. This won't necessarily translate to real world deployments, but this change seems trivial enough to be harmless?

I guess the only downside is that this can add a small delay to the app being informed, but not sure entirely sure it's relevant.

### DIFF
--- a/consensus/src/simplex/actors/voter/mod.rs
+++ b/consensus/src/simplex/actors/voter/mod.rs
@@ -848,16 +848,24 @@ mod tests {
             }
             assert_eq!(finalized_view, Some(view));
 
+            // Wait for a finalization to be recorded
+            loop {
+                {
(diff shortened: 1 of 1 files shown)


patrick-ogrady:
I actually am a big fan of this...nothing should get ahead of consensus's work.

That being said, I suspect the improvement from this change is negligible (other than in the simulated env).

PR #2303, opened 2025-11-21
[storage/adb] unify fixed & variable operation adb variants

roberto-bayardo:
Restructures the adb::any module such that the top-level submodules are only "ordered" and "unordered".  Within each of those, the code for variable vs fixed variants is now largely unified other than a very thin shim.  (I will add the ordered::variable implementation in a followup to keep the diff here from growing.)

### DIFF
--- a/storage/src/adb/any/ordered/fixed.rs
+++ b/storage/src/adb/any/ordered/fixed.rs
@@ -72,6 +72,12 @@ impl<E: Storage + Clock + Metrics, K: Array, V: CodecFixed<Cfg = ()>, H: Hasher,
         Self::init_with_callback(context, cfg, None, |_, _| {}).await
     }
 
+    /// Initialize the DB, invoking `callback` for each operation processed during recovery.
+    ///
+    /// If `known_inactivity_floor` is provided and is less than the log's actual inactivity floor,
+    /// `callback` is invoked with `(false, None)` for each location in the gap. Then, as the snapshot
+    /// is built from the log, `callback` is invoked for each operation with its activity status and
+    /// previous location (if any).
     pub(crate) async fn init_with_callback(
         context: E,
         cfg: Config<T>,
--- a/storage/src/adb/any/unordered/fixed.rs
+++ b/storage/src/adb/any/unordered/fixed.rs
@@ -50,6 +50,12 @@ impl<E: Storage + Clock + Metrics, K: Array, V: CodecFixed<Cfg = ()>, H: Hasher,
         Self::init_with_callback(context, cfg, None, |_, _| {}).await
     }
 
+    /// Initialize the DB, invoking `callback` for each operation processed during recovery.
+    ///
+    /// If `known_inactivity_floor` is provided and is less than the log's actual inactivity floor,
+    /// `callback` is invoked with `(false, None)` for each location in the gap. Then, as the snapshot
+    /// is built from the log, `callback` is invoked for each operation with its activity status and
+    /// previous location (if any).
     pub(crate) async fn init_with_callback(
         context: E,
         cfg: Config<T>,


storage/src/adb/any/ordered/mod.rs
```diff
@@ -0,0 +1,843 @@
...
+    fn new_update(key: Self::Key, value: Self::Value, next_key: Self::Key) -> Self;
+
+    /// Return a new delete operation variant.
+    fn new_delete(key: Self::Key) -> Self;
+
+    /// Return a new commit-floor operation variant.
+    fn new_commit_floor(inactivity_floor_loc: Location) -> Self;
+
+    /// Return this operation's key data, or None if this operation variant doesn't have any.
+    fn to_key_data(&self) -> Option<&KeyData<Self::Key, Self::Value>>;
```
clabby: Small nit - let's name this function `as_key_data`, in-line with the [API guidelines](https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv) for borrowed->borrowed conversions.

storage/src/adb/any/ordered/mod.rs
```diff
@@ -0,0 +1,843 @@
...
+            inactivity_floor_loc,
+            snapshot,
+            last_commit,
+            steps: 0,
+            active_keys,
+            translator: PhantomData,
+        })
+    }
+
+    async fn get_update_op(
```
clabby, line 143:
```suggestion
    /// Get an update operation's [KeyData] at a given location.
    ///
    /// ## Panics
    ///
    /// Panics if an operation was found at the given location, but doesn't
    /// have key data.
    async fn get_update_op(
```

storage/src/adb/any/unordered/mod.rs
```diff
@@ -0,0 +1,799 @@
...
+            steps: 0,
+            active_keys,
+            translator: PhantomData,
+        })
+    }
+
+    /// Whether the snapshot currently has no active keys.
+    pub fn is_empty(&self) -> bool {
+        self.active_keys == 0
+    }
```
clabby, line 123: This is implemented both on the `IndexedLog` type itself and as a part of the `AnyDb` trait impl on `IndexedLog`; deduplicate?

storage/src/adb/any/ordered/mod.rs
```diff
@@ -0,0 +1,843 @@
...
+            .into_key_data()
+            .expect("update operation must have key data");
+
+        Ok(Some((data.value, data.next_key)))
+    }
+
+    /// Whether the snapshot currently has no active keys.
+    pub fn is_empty(&self) -> bool {
+        self.active_keys == 0
+    }
```
clabby: This is implemented both on the`IndexedLog` type itself and as a part of the `AnyDb` trait impl on `IndexedLog`; deduplicate?

storage/src/adb/any/mod.rs
```diff
@@ -2,406 +2,158 @@
...
+    pub translator: T,
+
+    /// An optional thread pool to use for parallelizing batch operations.
+    pub thread_pool: Option<ThreadPool>,
+
+    /// The buffer pool to use for caching data.
+    pub buffer_pool: PoolRef,
+}
+
+/// Configuration for an `Any` authenticated db.
```
clabby:
```suggestion
/// Configuration for an `Any` authenticated db with variable-sized values.
```

storage/src/adb/any/mod.rs
```diff
@@ -2,406 +2,158 @@
...
-    /// Returns a proof of inclusion of all operations in the range starting at (and including)
-    /// location `start_loc`, and ending at the first of either:
-    /// - the last operation performed, or
-    /// - the operation `max_ops` from the start.
-    ///
-    /// Also returns a vector of operations corresponding to this range.
-    pub(super) async fn historical_proof(
+        start_loc: Location,
+        max_ops: NonZeroU64,
+    ) -> impl Future<Output = Result<(Proof<H::Digest>, Vec<O>), Error>>;
```
clabby: Let's add `+ Send` here; Very useful when we need to schedule the future in a `try_join` or something.

storage/src/adb/any/mod.rs
```diff
@@ -2,406 +2,158 @@
...
         historical_size: Location,
         start_loc: Location,
         max_ops: NonZeroU64,
-    ) -> Result<(Proof<H::Digest>, Vec<O>), Error> {
-        self.log
-            .historical_proof(historical_size, start_loc, max_ops)
-            .await
-            .map_err(Into::into)
-    }
+    ) -> impl Future<Output = Result<(Proof<H::Digest>, Vec<O>), Error>>;
```
clabby: Let's add `+ Send` here; Very useful when we need to schedule the future in a `try_join` or something.

storage/src/adb/any/ordered/mod.rs
clabby: Seems right to document the underlying panic's possibility, since the location could hold another type of operation iiuc?

storage/src/adb/any/mod.rs
clabby: To match up with:
https://github.com/commonwarexyz/monorepo/blob/448b7765e82981f131c397778613c23ceadb7f74/storage/src/adb/any/mod.rs#L48-L50

danlaine:
Should we consolidate the variable-length value `Operation`

```rs
pub enum Operation<K: Array, V: Codec> {
    // Operations for immutable stores.
    Set(K, V),
    Commit(Option<V>),
    // Operations for mutable stores.
    Delete(K),
    Update(K, V),
    CommitFloor(Option<V>, Location),
}
```

and the fixed-length value `Operation`

```rs
pub enum Operation<K: Array, V: CodecFixed> {
    /// Indicates the key no longer has a value.
    Delete(K),

    /// Indicates the key now has the wrapped value.
    Update(K, V),

    /// Indicates all prior operations are no longer subject to rollback, and the floor on inactive
    /// operations has been raised to the wrapped value.
    CommitFloor(Location),
}
```

`adb::Immutable` is the only place `Set` is used -- perhaps `Immutable` should have its own `Operation` implementation? Then we could have one shared implementation for operations used in `adb::any`:

```rs
pub enum Operation<K: Array, V: Codec> {
    Delete(K),
    Update(K, V),
    CommitFloor(Option<V>, Location),
}
```

storage/src/adb/any/ordered/mod.rs
roberto-bayardo: Thanks, agree `to_key_data` is wrong. But would just `key_data` be in line with guidelines to be consistent with the `Keyed` operation trait which provides `key` and `value` reference-returning accessors? If that's not idiomatic, I guess I should change those to `as_key` and `as_value`.

storage/src/adb/any/unordered/mod.rs
roberto-bayardo: thanks, yes, I neglected to remove these when I added the trait. Fixed.

storage/src/adb/any/ordered/mod.rs
roberto-bayardo: oops, correct, not sure why I dropped that comment.

storage/src/adb/any/ordered/mod.rs
roberto-bayardo: done

storage/src/adb/any/mod.rs
roberto-bayardo: This seems to propagate Send requirements more widely than I'd like (for purposes of this PR). Leave for when it's needed?

storage/src/adb/any/mod.rs
roberto-bayardo: done

storage/src/adb/any/unordered/sync.rs
```diff
@@ -90,16 +89,13 @@ where
                 .await?;
         // Build the snapshot from the log.
         let snapshot = Index::init(context.with_label("snapshot"), db_config.translator.clone());
-        let log = OperationLog::from_components(range.start, log, snapshot).await?;
-
-        let mut db = Any { log };
-        db.sync().await?;
+        let db = IndexedLog::from_components(range.start, log, snapshot).await?;
```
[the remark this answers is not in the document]
roberto-bayardo: Thanks, I added the sync() call to the authenticated log from_components() implementation, where it now best belongs.

danlaine:
Looks like `storage/src/adb/any/fixed/mod.rs` isn't included in the build tree and can be removed

storage/src/adb/any/mod.rs
```diff
@@ -2,472 +2,169 @@
...
-type AuthenticatedLog<E, C, O, H, S = Clean<DigestOf<H>>> = authenticated::Journal<E, C, O, H, S>;
+pub mod ordered;
+pub mod unordered;
 
-/// Type alias for the floor helper state wrapper used by the [OperationLog].
-type FloorHelperState<'a, E, C, O, I, H, T, S> =
-    FloorHelper<'a, T, I, AuthenticatedLog<E, C, O, H, S>, O>;
+/// Trait for an authenticated database (ADB) that provides succinct proofs of _any_ value ever
+/// associated with a key.
+pub trait AnyDb<O: Keyed, H: Hasher>: Db<O::Key, O::Value> {
```
danlaine:
```suggestion
pub trait AnyDb<O: Keyed, D: Digest>: Db<O::Key, O::Value> {
```

roberto-bayardo:
> Looks like `storage/src/adb/any/fixed/mod.rs` isn't included in the build tree and can be removed

guess the rebase brought that back in. removed

storage/src/adb/any/mod.rs
```diff
@@ -2,472 +2,169 @@
...
+
+    /// The size of the write buffer to use for each blob in the MMR journal.
+    pub mmr_write_buffer: NonZeroUsize,
+
+    /// The name of the [Storage] partition used for the MMR's metadata.
+    pub mmr_metadata_partition: String,
+
+    /// The name of the [Storage] partition used to persist the log of operations.
+    pub log_partition: String,
+    /// The size of the write buffer to use for each blob in the log journal.
```
danlaine, line 111: nit
```suggestion

    /// The size of the write buffer to use for each blob in the log journal.
```

danlaine: Also, do we want to align naming between `FixedConfig` and `VariableConfig` now that they're adjacent? e.g. `log_items_per_section` vs `log_items_per_blob`. I'd be OK deferring this to keep the PR smaller.

storage/src/adb/any/mod.rs
```diff
@@ -2,472 +2,169 @@
...
+    /// An optional thread pool to use for parallelizing batch operations.
+    pub thread_pool: Option<ThreadPool>,
+
+    /// The buffer pool to use for caching data.
+    pub buffer_pool: PoolRef,
 }
 
-impl<E, C, O, I, H, T> OperationLog<E, C, O, I, H, T, Clean<H::Digest>>
-where
+pub(super) type AuthenticatedLog<E, O, H, S> = authenticated::Journal<E, Journal<E, O>, O, H, S>;
```
danlaine:
```suggestion
type AuthenticatedLog<E, O, H, S> = authenticated::Journal<E, Journal<E, O>, O, H, S>;
```

storage/src/adb/any/ordered/fixed.rs
```diff
@@ -7,671 +7,102 @@
...
-
-        // If we get here, then `key` must precede the first key in the snapshot, in which case we
-        // have to cycle around to the very last key.
-        let iter = self.log.snapshot.last_translated_key();
-        let span = Self::find_span(&self.log.log, iter, key)
-            .await?
-            .expect("a span that includes any given key should always exist if db is non-empty");
-
-        Ok(Some(span))
+impl<K: Array, V: CodecFixed<Cfg = ()>> OperationTrait for Operation<K, V> {
```
danlaine, line 29: Should we put this impl in the same place as `Operation` in `adb/operation/fixed/ordered.rs`?

storage/src/adb/any/unordered/fixed.rs
```diff
@@ -3,46 +3,46 @@
...
 use commonware_utils::Array;
-use std::num::NonZeroU64;
 
-type Contiguous<E, K, V> = Journal<E, Operation<K, V>>;
+pub type Operation<K, V> = FixedOperation<K, V>;
 
-/// Type alias for the operation log of this [Any] database variant.
-pub(crate) type AnyLog<E, K, V, H, T, S> =
-    OperationLog<E, Contiguous<E, K, V>, Operation<K, V>, Index<T, Location>, H, T, S>;
+impl<K: Array, V: CodecFixed<Cfg = ()>> OperationTrait for Operation<K, V> {
```
danlaine, line 25: Similar question to above: should this impl go next to the definition of `Operation`?

storage/src/adb/any/unordered/fixed.rs
```diff
@@ -59,14 +59,12 @@ impl<
...
         value: V,
     ) -> Result<Option<Location>, Error> {
-        self.log
-            .update_key_with_op(Operation::Update(key, value))
-            .await
+        self.update_key_with_op(Operation::Update(key, value)).await
     }
 
     /// Deletes `key` from the db, returning the old location of the key if any.
     pub(crate) async fn delete_return_loc(&mut self, key: K) -> Result<Option<Location>, Error> {
```
danlaine: nit: this only has 1 usage. We could remove this helper and just call `delete_key(Operation::Delete(key))` there? Same for `update_return_loc`.

storage/src/adb/any/mod.rs
roberto-bayardo: changed

storage/src/adb/any/mod.rs
roberto-bayardo: done & done

storage/src/adb/any/mod.rs
roberto-bayardo: done

storage/src/adb/any/ordered/fixed.rs
roberto-bayardo: I was hesitant to further clutter up the operations/ module without knowing if these specializaions were going to be more generally useful.  (OperationTrait from which this derives is also in any/ not operation/ )

storage/src/adb/any/unordered/fixed.rs
roberto-bayardo: done

storage/src/adb/any/ordered/mod.rs
```diff
@@ -0,0 +1,874 @@
...
+use commonware_runtime::{Clock, Metrics, Storage};
+use core::{marker::PhantomData, num::NonZeroU64};
+use tracing::debug;
+
+pub mod fixed;
+
+type AuthenticatedLog<E, C, O, H, S = Clean<DigestOf<H>>> = authenticated::Journal<E, C, O, H, S>;
+
+/// Type alias for the floor helper state wrapper used by the [IndexedLog].
+pub(crate) type FloorHelperState<'a, E, C, O, I, H, T> =
```
danlaine: Or we could remove this? We only use it in one place

```suggestion
type FloorHelperState<'a, E, C, O, I, H, T> =
```

storage/src/adb/any/ordered/mod.rs
```diff
@@ -0,0 +1,874 @@
...
+    FloorHelper<'a, T, I, AuthenticatedLog<E, C, O, H>, O>;
+
+/// Type alias for a location and its associated key data.
+pub type LocatedKey<K, V> = (Location, KeyData<K, V>);
```
danlaine:
```suggestion
type LocatedKey<K, V> = (Location, KeyData<K, V>);
```

storage/src/adb/any/unordered/variable.rs
```diff
@@ -135,14 +83,16 @@ impl<
...
 {
-    /// Returns a [Any] adb initialized from `cfg`. Any uncommitted log operations will be
+    /// Returns an [Any] adb initialized from `cfg`. Any uncommitted log operations will be
     /// discarded and the state of the db will be as of the last committed operation.
     pub async fn init(
         context: E,
-        cfg: Config<T, <Operation<K, V> as Read>::Cfg>,
+        cfg: VariableConfig<T, <Operation<K, V> as Read>::Cfg>,
     ) -> Result<Self, Error> {
+        let translator = cfg.translator.clone();
```
danlaine: Nit: I don't think we need to clone() here, we can just pass as `cfg.translator` into `init_from_log`

storage/src/adb/any/unordered/fixed.rs
```diff
@@ -3,211 +3,71 @@
...
-    async fn close(self) -> Result<(), Error> {
-        self.log.close().await
-    }
-
-    async fn destroy(self) -> Result<(), Error> {
-        self.log.destroy().await
+        Self::init_with_callback(context, cfg, None, |_, _| {}).await
+    }
+
+    pub(crate) async fn init_with_callback(
```
danlaine, line 59:
```suggestion
    /// Initialize the DB, invoking `callback` for each operation processed during recovery.
    ///
    /// If `known_inactivity_floor` is provided and is less than the log's actual inactivity floor,
    /// `callback` is invoked with `(false, None)` for each location in the gap. Then, as the snapshot
    /// is built from the log, `callback` is invoked for each operation with its activity status and
    /// previous location (if any).
    pub(crate) async fn init_with_callback(
```

danlaine: (same for ordered)

storage/src/adb/any/ordered/mod.rs
roberto-bayardo: Indeed. Removed this one & the one in unordered too.

storage/src/adb/any/ordered/mod.rs
roberto-bayardo: done

storage/src/adb/any/unordered/fixed.rs
roberto-bayardo: done

storage/src/adb/any/unordered/variable.rs
roberto-bayardo: done

danlaine (approved):

PR #2307, opened 2025-11-23
fix(coding): scheme doctest was using wrong shards

samlaf:
It was popping from the back of checking_data_w_shard, but skipping from the front of reshards, which ended up with duplicate shards in checked_shards.

This was never caught because decode was only ever called with subsets of checked_shards, which never contained the duplicate shard.

### DIFF
--- a/coding/src/lib.rs
+++ b/coding/src/lib.rs
@@ -98,7 +98,7 @@ pub struct CodecConfig {
 ///
 /// // Each person produces reshards, their own checked shard, and checking data
 /// // to check other peoples reshards.
-/// let (mut checking_data_w_shard, reshards): (Vec<_>, Vec<_>) = shards
+/// let (mut checking_data_w_shard, mut reshards): (Vec<_>, Vec<_>) = shards
 ///         .into_iter()
 ///         .enumerate()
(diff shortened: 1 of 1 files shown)


coding/src/lib.rs
```diff
@@ -108,19 +108,26 @@ pub struct CodecConfig {
...
 /// let data2 = RS::decode(&config, &commitment, checking_data, &checked_shards[..2], CONCURRENCY).unwrap();
 /// assert_eq!(&data[..], &data2[..]);
 ///
 /// // Decoding works with different shards, with a guarantee to get the same result.
 /// let data3 = RS::decode(&config, &commitment, checking_data, &checked_shards[1..], CONCURRENCY).unwrap();
 /// assert_eq!(&data[..], &data3[..]);
+///
+/// // Of course, we can also recover from all shards; this will also check that there are no duplicates.
+/// let data4 = RS::decode(&config, &commitment, checking_data, &checked_shards, CONCURRENCY).unwrap();
+/// assert_eq!(&data[..], &data4[..]);
```
samlaf, line 130: this assert failed in the previous version

clabby:

PR #2312, opened 2025-11-23
[utils] Move `OrderedExt` to `utils`

### LINKED ISSUE #2294: [utils] Move `OrderedExt` to `utils::OrderedQuorum`
Move the implementation of `OrderedExt` (https://github.com/commonwarexyz/monorepo/blob/b82ca8410d190a9f3a4b8ca03babe7a661d44f8c/consensus/src/simplex/types.rs#L314-L345) in `consensus` to `utils` so it can be used in `cryptography` (for `bls12381::dkg`).

This is particularly reasonable because `quorum` and `max_faults` (and `Ordered`) already live in `utils`...

clabby:
## Overview

Moves `commonware-consensus`' `OrderedExt` trait to `commonware_utils::set::OrderedQuorum` and uses this trait in `bls12381::dkg`.

closes #2294

### DIFF
--- a/utils/src/set.rs
+++ b/utils/src/set.rs
@@ -164,7 +164,10 @@ impl<T: Ord> From<Ordered<T>> for Vec<T> {
 }
 
 /// Extension trait for `Ordered` participant sets providing quorum and index utilities.
-pub trait OrderedQuorum<T> {
+pub trait OrderedQuorum {
+    /// The type of items in this set.
+    type Item: Ord;
+
     /// Returns the quorum value (2f+1) for this participant set.
     ///
     /// ## Panics
@@ -180,17 +183,19 @@ pub trait OrderedQuorum<T> {
     fn max_faults(&self) -> u32;
 
     /// Returns the participant key at the given index.
-    fn key(&self, index: u32) -> Option<&T>;
+    fn key(&self, index: u32) -> Option<&Self::Item>;
 
     /// Returns the index for the given participant key, if present.
     ///
     /// ## Panics
     ///
     /// Panics if the participant index exceeds `u32::MAX`.
-    fn index(&self, key: &T) -> Option<u32>;
+    fn index(&self, key: &Self::Item) -> Option<u32>;
 }
 
-impl<T: Ord> OrderedQuorum<T> for Ordered<T> {
+impl<T: Ord> OrderedQuorum for Ordered<T> {
+    type Item = T;
+
     fn quorum(&self) -> u32 {
         crate::quorum(u32::try_from(self.len()).expect("too many participants"))
     }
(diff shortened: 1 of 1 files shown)


utils/src/set.rs
```diff
@@ -163,6 +163,55 @@ impl<T: Ord> From<Ordered<T>> for Vec<T> {
     }
 }
 
+/// Extension trait for `Ordered` participant sets providing quorum and index utilities.
+pub trait OrderedQuorum<T> {
```
patrick-ogrady: Was debating with @cronokirby whether or not we should just implement this on `Ordered`? It definitely seems like "extra stuff" but given its only implemented by one struct I was torn...

clabby: Not super opinionated; either works. my 2c, am a fan of extension traits when it comes to functionality that isn't used out of the box. Nice to namespace functionality that doesn't always apply to a given context. For example, `OrderedQuorum` doesn't make sense in `p2p::authenticated::discovery`'s ordered peer sets:

https://github.com/commonwarexyz/monorepo/blob/641d61e39f2c030b336beacab7e468bfa47add0d/p2p/src/authenticated/discovery/actors/tracker/ingress.rs#L255-L264

otoh, @cronokirby mentioned he isn't a fan of them due to the extended functionality not being straight-forward to find. When you do need it, you need to know it's there, which the LSP isn't always helpful with.

cryptography/src/bls12381/dkg/player.rs
```diff
@@ -53,11 +53,11 @@ impl<P: PublicKey, V: Variant> Player<P, V> {
         recipients: Ordered<P>,
         concurrency: usize,
     ) -> Self {
-        let me_idx = recipients.position(&me).expect("player not in recipients") as u32;
+        let me_idx = recipients.index(&me).expect("player not in recipients");
```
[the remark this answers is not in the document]
clabby: Good shout, but think this is fine as is.

utils/src/set.rs
```diff
@@ -163,6 +163,55 @@ impl<T: Ord> From<Ordered<T>> for Vec<T> {
...
+        crate::max_faults(u32::try_from(self.len()).expect("too many participants"))
+    }
+
+    fn key(&self, index: u32) -> Option<&T> {
+        self.get(index as usize)
+    }
+
+    fn index(&self, key: &T) -> Option<u32> {
+        self.position(key)
+            .and_then(|index| u32::try_from(index).ok())
```
patrick-ogrady: nit: I think we should expect on the conversion

patrick-ogrady (changes requested):

utils/src/set.rs
patrick-ogrady: I think the way you have it seems fine (it was what I first did in my attempt actually).

We can continue to debate but undoubtedly better than what was there.

utils/src/set.rs
```diff
@@ -163,6 +163,52 @@ impl<T: Ord> From<Ordered<T>> for Vec<T> {
...
+        crate::max_faults(u32::try_from(self.len()).expect("too many participants"))
+    }
+
+    fn key(&self, index: u32) -> Option<&T> {
+        self.get(index as usize)
+    }
+
+    fn index(&self, key: &T) -> Option<u32> {
+        self.position(key)
+            .map(|position| u32::try_from(position).expect("participant index overflow"))
```
patrick-ogrady: nit: "too many participants"?

patrick-ogrady (approved):

Issue #2266, opened 2025-11-18
[codec] Allow using `()` as a storage key


cronokirby:
Sometimes it's convenient to use a keyed storage, but only with one possible key. For example, you might use metadata storage, with partitions for different actors, and have a situation where one actor stores its metadata under a single key.

In these cases, using `()` would be more elegant than using some other type with a special value, or using `FixedBytes<0>` to accomplish the same thing in a bit more unclear way.

PR #2313, opened 2025-11-23
[utils] Add `Unit` marker type

clabby:
## Overview

Adds a new `Unit` marker type that implements `Span` and `Array`. This is useful, for example, for areas in the codebase where we have a `Metadata` store that only ever accepts one possible key.

closes https://github.com/commonwarexyz/monorepo/issues/2266

### DIFF
--- a/consensus/src/types.rs
+++ b/consensus/src/types.rs
@@ -649,7 +649,7 @@ mod tests {
     #[test]
     fn test_round_display() {
         let r = Round::new(Epoch::new(5), View::new(100));
-        assert_eq!(format!("{r:?}"), "(5, 100)");
+        assert_eq!(format!("{r}"), "(5, 100)");
     }
 
     #[test]


consensus/src/marshal/actor.rs
```diff
@@ -201,7 +198,7 @@ where
         )
         .await
         .expect("failed to initialize application metadata");
-        let last_processed_height = application_metadata.get(&LATEST_KEY).copied().unwrap_or(0);
+        let last_processed_height = application_metadata.get(&Unit).copied().unwrap_or(0);
```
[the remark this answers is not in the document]
clabby: This is correct, though not a bug. The https://github.com/commonwarexyz/monorepo/labels/breaking label was added for this.

consensus/src/marshal/actor.rs
```diff
@@ -55,9 +55,6 @@ use std::{
 };
 use tracing::{debug, error, info, warn};
 
-/// The key used to store the last processed height in the metadata store.
-const LATEST_KEY: U64 = U64::new(0xFF);
```
patrick-ogrady, line 59: I actually prefer this approach because it allows us to add more keys here seamlessly.

patrick-ogrady (changes requested):

consensus/src/marshal/actor.rs
patrick-ogrady: The proposed technique will make it much more difficult to add more items to metadata (something imho makes sense to keep open).

patrick-ogrady: I'm ok to add a UNIT type but prefer not to do these migrations.

clabby: Rolled back the migrations. Makes sense for cases where we anticipate needing more keys, though do we anticipate needing more keys here? The metadata stores in DKG / `Cache` seem to have a single purpose, `marshal`'s is a bit more iffy.

patrick-ogrady: Famous last words 😉

utils/src/sequence/unit.rs
```diff
@@ -0,0 +1,85 @@
...
+/// An `Array` implementation for the unit type `()`.
+#[derive(Clone, Copy, Default, Eq, PartialEq, Ord, PartialOrd, Hash)]
+pub struct Unit;
+
+impl Write for Unit {
+    fn write(&self, _: &mut impl BufMut) {}
+}
+
+impl FixedSize for Unit {
+    const SIZE: usize = 0;
```
patrick-ogrady, line 18: I think this is correct...just wondering if there is some weird DoS this opens where we can process "infinite" Units 🤔 

I think its fine because we should parameterize whatever struct they come in with some max cardinality.

patrick-ogrady (approved):

consensus/src/types.rs
```diff
@@ -649,7 +649,7 @@ mod tests {
     #[test]
     fn test_round_display() {
         let r = Round::new(Epoch::new(5), View::new(100));
-        assert_eq!(format!("{}", r), "(5, 100)");
+        assert_eq!(format!("{r:?}"), "(5, 100)");
```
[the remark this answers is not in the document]
patrick-ogrady: Whoops

patrick-ogrady (approved):

PR #2315, opened 2025-11-24
[ci] Improve Caching

patrick-ogrady:


.github/actions/setup/action.yml
```diff
@@ -7,3 +12,5 @@ runs:
       with:
         cache-on-failure: true
         save-if: ${{ github.ref == 'refs/heads/main' }}
+        key: ${{ inputs.cache-key }}
+        cache-workspace-crates: false
```
patrick-ogrady: We will almost always be rebuilding workspace crates in CI, so we shouldn't cache them.

clabby: It will cache the incremental build, so it is smart enough to recompile if it does need to change. Though this will save a bit of space.

patrick-ogrady:
> This will work to differentiate the cache of each partition, but it will contribute towards blowing up the cache even more since each partition stores its own separate cache (there's a lot of duplicate intermediate artifacts in each.)
> 
> We're already over our usage limit as well; Two things we could do:
> 
> * Pay for more cache and just do this.
> * Be smarter about sharing a cache between all jobs with the same compilation settings using a cache factory like `rust-libp2p` does. We'd need:
>   
>   * Windows (debug)
>     
>     * w/ `--features commonware-runtime/external`
>   * MacOS (debug)
>     
>     * w/ `--features commonware-runtime/external`
>   * Linux
>     
>     * `debug` w/ `commonware-runtime/external`
>     * `debug` w/ `commonware-runtime/iouring-storage`
>     * `debug` w/ `commonware-runtime/iouring-network`
>     * `debug` w/ `no-default-features`
>     * `bench`
>     * `bench` w/ `commonware-runtime/iouring-storage`
> 
> (Not worth caching WASM/no_std builds IMO; those are speedy.)
> 
> Defer to you - certainly a lot easier just to pay for more cache and hold a cache per job.

I'm very OK paying. I thought 10GB was a hard cap?

.github/actions/setup/action.yml
patrick-ogrady: Will remove this

patrick-ogrady:
Thank you based GitHub 🙇 

https://github.blog/changelog/2025-11-20-github-actions-cache-size-can-now-exceed-10-gb-per-repository/

.github/actions/setup/action.yml
```diff
@@ -1,9 +1,15 @@
...
+    default: ""
 runs:
   using: composite
   steps:
     - name: Rust Cache
       uses: Swatinem/rust-cache@f13886b937689c021905a6b90929199931d60db1 # v2.8.1
       with:
-        cache-on-failure: true
         save-if: ${{ github.ref == 'refs/heads/main' }}
+        key: ${{ replace(replace(inputs.cache-key, ' ', '-'), '/', '-') }}
```
patrick-ogrady: This should fix the spaces and slashes.

.github/actions/setup/action.yml
```diff
@@ -1,9 +1,15 @@
...
+    default: ""
 runs:
   using: composite
   steps:
     - name: Rust Cache
       uses: Swatinem/rust-cache@f13886b937689c021905a6b90929199931d60db1 # v2.8.1
       with:
-        cache-on-failure: true
         save-if: ${{ github.ref == 'refs/heads/main' }}
+        key: ${{ replace(replace(inputs.additional-cache-key, ' ', '-'), '/', '-') }}
```
patrick-ogrady: Should prevent any odd characters in key

PR #1960, opened 2025-10-22
[consensus] Add `Automaton::certify` check to be carried out after notarization

BrendanChou:
Fixes #1767 

Changes:
- Add `certify` function to the `Automaton` trait
  - A default implementation will return a `oneshot::Receiver<bool>` with `true` sent to it
- Modify the Simplex voter actor
  - Call `automaton.certify(...)` once any proposal is notarized
  - Do not move forward views on a notarized-but-not-certified (NbnC) proposal
  - Do not build on top of a NbnC proposal
  - Do-not-finalize a NbnC proposal
  - Do not allow future proposals to build on top of a NbnC proposal
  - Write (and read on startup) `certify()` results to disk
  - Immediately nullify a NbnC proposal (by timing-out that view)

TODO: cleanup and "matrixify" tests by testing with both always-certify and sometimes-certify tests

PR #2317, opened 2025-11-24
[Temp] certify branch

BrendanChou:
Will eventually push these changes into #1960. Putting in a different PR for now while things stabilize

### DIFF
--- a/consensus/src/simplex/actors/voter/actor.rs
+++ b/consensus/src/simplex/actors/voter/actor.rs
@@ -357,12 +357,21 @@ impl<
         view: View,
         success: bool,
     ) {
-        if self.state.certified(view, success) {
-            // Persist certification result for recovery
-            let msg = Voter::Certification(Rnd::new(self.state.epoch(), view), success);
-            self.append_journal(view, msg.clone()).await;
-            self.sync_journal(view).await;
-            resolver.updated(msg).await;
+        let Some(notarization) = self.state.certified(view, success) else {
+            return;
+        };
+
+        // Persist certification result for recovery
+        let msg = Voter::Certification(Rnd::new(self.state.epoch(), view), success);
+        self.append_journal(view, msg.clone()).await;
+        self.sync_journal(view).await;
+        resolver.updated(msg).await;
+
+        // Inform listeners of successful certification
+        if success {
+            self.reporter
+                .report(Activity::Certification(notarization))
+                .await;
         }
     }
 
@@ -382,7 +391,7 @@ impl<
     async fn handle_nullification(
         &mut self,
         nullification: Nullification<S>,
-    ) -> Option<Voter<S, D>> {
+    ) -> Option<impl Iterator<Item = Voter<S, D>>> {
         let view = nullification.view();
         let msg = Voter::Nullification(nullification.clone());
 
@@ -392,11 +401,11 @@ impl<
         }
         self.append_journal(view, msg).await;
 
-        // If we were the leader, we should emit the "best" certificate that we have.
-        // The best certificate is the highest certificate for a certified view.
+        // If we were the leader, we should emit a certifiable chain of ancestor certificates such
+        // that other peers are able to verify my proposal in the future.
         let leader = self.state.leader_index(view)?;
         if self.state.is_me(leader) {
-            return self.state.get_best_certificate();
+            return self.state.get_certifiable_ancestry(view);
         }
         None
     }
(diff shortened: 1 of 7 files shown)


consensus/src/lib.rs
```diff
@@ -99,6 +99,26 @@ cfg_if::cfg_if! {
...
+            /// Applications that employ erasure-coding can override this to delay or prevent
+            /// finalization until they have reconstructed and validated the full block (e.g., after
+            /// receiving enough shards).
+            fn certify(
+                &mut self,
+                _context: Self::Context,
+                _payload: Self::Digest,
+            ) -> impl Future<Output = oneshot::Receiver<bool>> + Send {
+                #[allow(clippy::async_yields_async)]
+                async move {
```
patrick-ogrady, line 116: wen specialization 🙃

consensus/src/simplex/types.rs
```diff
@@ -809,6 +812,8 @@ pub enum Voter<S: Scheme, D: Digest> {
     Finalize(Finalize<S, D>),
     /// A recovered certificate for a finalization (scheme-specific)
     Finalization(Finalization<S, D>),
+    /// A local record indicating the proposal for this round was certified by the automaton
+    Certification(Round, bool),
```
patrick-ogrady, line 782: I think it makes sense to make this a `Voter` but we may want to think about the enum comment we currently have (as for what this is) 🤔

consensus/src/simplex/actors/resolver/state.rs
```diff
@@ -119,28 +101,20 @@ impl<S: Scheme, D: Digest> State<S, D> {
...
-            if !self.nullifications.contains_key(&cursor) {
-                requests.push(cursor.into());
-                debug!(%cursor, "requested missing nullification");
-            }
-
-            // Increment cursor
-            cursor = cursor.next();
-        }
+        let start = self.floor_view().next();
+        let requests = View::range(start, self.current_view)
```
patrick-ogrady, line 120: I don't think we should make this fetch change in this PR. The existing logic was intentional (to avoid enqueueing all items per call to fetch).

consensus/src/simplex/actors/resolver/state.rs
```diff
@@ -10,78 +10,52 @@ use commonware_cryptography::Digest;
...
-                // Remove from pending and cancel request
-                resolver.cancel(view.into()).await;
             }
             Voter::Notarization(notarization) => {
-                // Update current view
                 let view = notarization.view();
-                if view > self.current_view {
-                    self.current_view = view;
+                if self.encounter_view(view) {
+                    self.notarizations.insert(view, notarization);
```
patrick-ogrady, line 57: Confirming your removal of the floor update here was intentional?

consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -26,16 +26,53 @@ use commonware_runtime::{
...
+
+impl<'a, D: Digest, P: PublicKey, R> Future for Waiter<'a, D, P, R> {
+    type Output = (Context<D, P>, Result<R, oneshot::Canceled>);
+
+    fn poll(self: Pin<&mut Self>, cx: &mut task::Context<'_>) -> Poll<Self::Output> {
+        let Waiter(slot) = self.get_mut();
+        match slot.as_mut() {
+            None => Poll::Pending,
+            Some(Request(ctx, ref mut receiver)) => match Pin::new(receiver).poll(cx) {
+                Poll::Ready(res) => Poll::Ready((ctx.clone(), res)),
```
patrick-ogrady: I guess we have to clone() again here because we only get a mutable reference to cx?

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -575,6 +605,73 @@ impl<E: Clock + Rng + CryptoRng + Metrics, S: Scheme, D: Digest> State<E, S, D>
...
+
+        // Log the result and exit early if certification failed since we should not move to the
+        // next view until a nullification is formed.
+        if is_success {
+            // Enter next view. The leader should already be set.
+            self.enter_view(view.next());
+            assert!(self
+                .views
+                .get(&view.next())
+                .is_some_and(|r| r.leader().is_some())); // TODO: remove this assert
```
patrick-ogrady: remove assert?

consensus/src/simplex/actors/resolver/state.rs
patrick-ogrady: I guess we may be able to put a "take" after the filter to limit (so we still benefit from new styling)

consensus/src/simplex/actors/resolver/state.rs
BrendanChou: Yes, the floor must be only `finalization`s now as `notarization`s may be discarded completely since they may be not-usable if the block does not certify

consensus/src/simplex/actors/voter/actor.rs
BrendanChou: Lemme see if it can be done without a clone

consensus/src/simplex/actors/resolver/state.rs
BrendanChou: So the reason that we can't have a limit is that:
- There is not guaranteed to be a nullification at every queried view (the resolver doesn't understand the status of any `certification`)
- The floor is only `finalizations` now, not `notarizations`

So there may be an unbounded number of views with a notarization but no nullification. Since such notarizations don't currently raise the floor (unless we start passing "certified=true" messages to the resolver), such views may take up all the possible limit, causing a deadlock (no resolving of future nullifications)

consensus/src/simplex/actors/voter/actor.rs
BrendanChou: Avoids a clone in https://github.com/commonwarexyz/monorepo/pull/2317/commits/de7c3dd25765513a220564b6a90c20da0ead6079

Issue #1116, opened 2025-06-17
Investigate if the Twins framework can be used for BFT-consensus testing


dnkolegov-ar:
[The Twins methodology](https://arxiv.org/abs/2004.10617) tests Byzantine behaviour by simply instantiating “twin” copies of correct replicas and using network partitioning, giving a lightweight but powerful fault model. However, its assumptions and APIs may not align with the current architecture. 

The reference implementations:
- [era-consensus](https://github.com/matter-labs/era-consensus/tree/5543639fb389896bc36090d7b6bff66d6236a83b/node/components/bft/src/testonly/twins)
- [relab](https://github.com/relab/hotstuff/tree/master/twins)

patrick-ogrady:
As discussed, there are a few layers of defense against both connecting to duplicate peers and handling messages from them. I think the cleanest way to implement this would be to "share" a simulated p2p channel between multiple logical consensus instances. We'd need to create a wrapper around "receive" to duplicate messages that the peer is supposed to receive but wouldn't be too hard at all.

dnkolegov-ar:
This simplest implementation is in https://github.com/commonwarexyz/monorepo/pull/1522

Issue #2060, opened 2025-10-30
[p2p/simulated] Lift simulated network backend into `commonware-runtime`


clabby:
## Overview

In tests, `p2p::simulated` is used to have granular control over participants and the links between them. This is a useful tool, but it's starting to show its warts:
1. Applications cannot share their p2p network initialization code between tests / the production entrypoint.
2. Peer set management differs between`p2p::simulated` and `p2p::{discovery/lookup}`.
3. Tests don't cover the p2p modules used in the production application, just everything in-front of the `Sender` / `Receiver` interfaces.

We should consider doing away with `p2p::simulated` in favor of an extension to the deterministic runtime's network. This would allow apps to run in the deterministic or tokio runtimes without any differences in the entrypoint, give more coverage over the production p2p modules, and prevent more ugly workarounds like the ones added in the solution to #1935 (#1984).

patrick-ogrady:
We'd still permit the "underling links" to be configured between IPs (which would be assigned to public keys). This is a necessary ingredient to run partition/bad links tests.

patrick-ogrady:
We didn't opt for this originally because it felt wasteful to test p2p while testing other layers (i.e. if p2p adheres to an interface and consensus works with that interface, it should just work).

The challenge, it turns out, with this approach is that the "mocks" for lower layers are actually quite sophisticated to build/ensure they maintain the same invariants as what is used in prod. We should consider not building sophisticated mocks and bundling all complex simulation/mock function into `runtime`.

We'll see how it comes out...putting all "mocks" in `runtime` should also help with #1398

PR #2319, opened 2025-11-24
[p2p/simulated] Support Split Channels

patrick-ogrady:
Related: #1116

Add `with_split` support to both `Sender` and `Receiver` to permit a single identity to be backed by two logical applications (with the ability to send/receive messages dynamically based on some condition).

### DIFF
--- a/p2p/src/simulated/network.rs
+++ b/p2p/src/simulated/network.rs
@@ -36,6 +36,7 @@ type Task<P> = (Channel, P, Recipients<P>, Bytes, oneshot::Sender<Vec<P>>);
 
 /// Target for a message in a split receiver.
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[must_use]
 pub enum SplitTarget {
     Primary,
     Secondary,
@@ -44,6 +45,7 @@ pub enum SplitTarget {
 
 /// Origin of a message in a split sender.
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[must_use]
 pub enum SplitOrigin {
     Primary,
     Secondary,


p2p/src/simulated/network.rs
```diff
@@ -967,6 +1020,184 @@ mod tests {
         });
     }
 
+    #[test]
```
patrick-ogrady, line 1112: nit: can reduce this to a single test

p2p/src/simulated/network.rs
```diff
@@ -632,6 +647,29 @@ impl<P: PublicKey> Sender<P> {
             processor,
         )
     }
+
+    /// Split this sender into two logical replicas that can route per message.
+    ///
+    /// The shared `router` receives the replica identifier, the requested recipients, and the
+    /// message payload, and returns the actual recipients to send to.
+    pub fn split_with<F>(self, router: F) -> (RoutedSender<P>, RoutedSender<P>)
```
patrick-ogrady: This approach allows us to [1] avoid throwing off bandwidth tracking and [2] avoid piping through the "split sender" all the way through the networking core.

p2p/src/simulated/network.rs
patrick-ogrady, line 1: It may have been possible to implement this over the p2p Sender/Receiver traits rather than in `simulated` but don't want to encourage usage in production.

This may be something we need to reconsider for #2060

p2p/src/simulated/network.rs
```diff
@@ -34,6 +34,43 @@ use tracing::{debug, error, trace, warn};
...
+}
+
+/// Origin of a message in a split sender.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum SplitOrigin {
+    Primary,
+    Secondary,
+}
+
+/// A function that forwards messages from [SplitOrigin] to [Recipients].
```
patrick-ogrady, line 54: Opted to support both split forwarding/routing because we will want to map each twin to some unique set of participants (each peer should only get a message from one twin).

patrick-ogrady:
While this PR "does the job", I still don't love the style of it. Open to any suggestions.

p2p/src/simulated/network.rs
```diff
@@ -677,6 +765,42 @@ impl<P: PublicKey> crate::Receiver for Receiver<P> {
...
+                if matches!(direction, SplitTarget::Primary | SplitTarget::Both) {
+                    if let Err(err) = primary_tx.send(message.clone()).await {
+                        error!(?err, "failed to send message to primary");
+                    }
+                }
+                if matches!(direction, SplitTarget::Secondary | SplitTarget::Both) {
+                    if let Err(err) = secondary_tx.send(message).await {
+                        error!(?err, "failed to send message to secondary");
+                    }
+                }
```
clabby: nit: can avoid the clone when only sending to the primary end by using a (slightly more readable IMO) match:
```suggestion
                match direction {
                    SplitTarget::Primary => {
                        if let Err(err) = primary_tx.send(message).await {
                            error!(?err, "failed to send message to primary");
                        }
                    }
                    SplitTarget::Secondary => {
                        if let Err(err) = secondary_tx.send(message).await {
                            error!(?err, "failed to send message to secondary");
                        }
                    }
                    SplitTarget::Both => {
                        if let Err(err) = primary_tx.send(message.clone()).await {
                            error!(?err, "failed to send message to primary");
                        }
                        if let Err(err) = secondary_tx.send(message).await {
                            error!(?err, "failed to send message to secondary");
                        }
                    }
              }
```

p2p/src/simulated/network.rs
```diff
@@ -34,6 +34,43 @@ use tracing::{debug, error, trace, warn};
...
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum SplitTarget {
+    Primary,
+    Secondary,
+    Both,
+}
+
+/// Origin of a message in a split sender.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum SplitOrigin {
```
clabby, line 49: nit

```suggestion
#[must_use]
pub enum SplitOrigin {
```

p2p/src/simulated/network.rs
```diff
@@ -34,6 +34,43 @@ use tracing::{debug, error, trace, warn};
 /// Task type representing a message to be sent within the network.
 type Task<P> = (Channel, P, Recipients<P>, Bytes, oneshot::Sender<Vec<P>>);
 
+/// Target for a message in a split receiver.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum SplitTarget {
```
clabby, line 40: nit
```suggestion
#[must_use]
pub enum SplitTarget {
```

p2p/src/simulated/network.rs
clabby: If we do end up implementing this on `Sender` / `Receiver` traits rather than gating it to simulated, strongly would prefer this functionality is gated behind `#[cfg(test)]`.

p2p/src/simulated/network.rs
```diff
@@ -677,6 +765,52 @@ impl<P: PublicKey> crate::Receiver for Receiver<P> {
...
+    /// Split this [Receiver] into a [SplitTarget::Primary] and [SplitTarget::Secondary] receiver.
+    pub fn split_with<E: Spawner, R: SplitRouter<P>>(
+        mut self,
+        context: E,
+        router: R,
+    ) -> (Self, Self) {
+        let (mut primary_tx, primary_rx) = mpsc::unbounded();
+        let (mut secondary_tx, secondary_rx) = mpsc::unbounded();
+        context.spawn(move |_| async move {
+            while let Some(message) = self.receiver.next().await {
```
clabby, line 780: Actually, this seems like an issue; The splitter task will never exit if we drop both the secondary and primary receiver?

patrick-ogrady: I suppose we'd assume the underlying receiver would drop but hm.

patrick-ogrady: I'll add a boolean to exit if both are dropped.

p2p/src/simulated/network.rs
```diff
@@ -677,6 +765,52 @@ impl<P: PublicKey> crate::Receiver for Receiver<P> {
...
+                        if let Err(err) = primary_tx.send(message.clone()).await {
+                            error!(?err, "failed to send message to primary");
+                        }
+                        if let Err(err) = secondary_tx.send(message).await {
+                            error!(?err, "failed to send message to secondary");
+                        }
+                    }
+                }
+            }
+        });
```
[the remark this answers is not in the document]
clabby: good cursor bot 🫡

p2p/src/simulated/network.rs
patrick-ogrady: *can use `.is_closed()`

p2p/src/simulated/network.rs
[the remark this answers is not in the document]
patrick-ogrady: wow

clabby (dismissed):
LGTM. Not attached to the `#[must_use]` clauses on the directions, but they are nice to make sure those are checked.

p2p/src/simulated/network.rs
patrick-ogrady: Yeah, makes sense.

p2p/src/simulated/network.rs
```diff
@@ -677,6 +765,58 @@ impl<P: PublicKey> crate::Receiver for Receiver<P> {
...
+                        }
+                    }
+                    SplitTarget::Both => {
+                        if let Err(err) = primary_tx.send(message.clone()).await {
+                            error!(?err, "failed to send message to primary");
+                        }
+                        if let Err(err) = secondary_tx.send(message).await {
+                            error!(?err, "failed to send message to secondary");
+                        }
+                    }
```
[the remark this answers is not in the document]
patrick-ogrady: This would be the correct behavior.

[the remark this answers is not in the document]
clabby: This is fine IMO.

PR #2323, opened 2025-11-25
[macros] avoid String clone on HashSet insert in load_nextest_groups

radik878:
Removed unnecessary allocation by avoiding normalized.clone() when inserting into a HashSet during test group collection. Since HashSet::insert takes ownership, the previous code cloned the string to still use it in the duplicate error path. We now first check contains(&normalized) to format the error while retaining ownership and then insert(normalized) on the non-duplicate path. This preserves exact behavior and error messages, reduces allocations on the hot path, and trades it for an additional hash lookup which is negligible given the small expected number of test groups

### DIFF
--- a/macros/src/nextest.rs
+++ b/macros/src/nextest.rs
@@ -89,13 +89,14 @@ fn load_nextest_groups() -> Result<NextestGroups, String> {
                 msg
             )
         })?;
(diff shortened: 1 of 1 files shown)


clabby:

PR #2324, opened 2025-11-25
[Storage] Remove `T: Translator` from `Index` trait

danlaine:
Allows for removing a bunch of `T`s.

Also consolidates some generic usage in `storage/src/adb/mod.rs` helper methods.


clabby (approved):
Nice

PR #2328, opened 2025-11-25
[consensus/marshal] Optionally verify using epoch-less certificate verifier

andresilva:
Extracted from #2287.

### DIFF
--- a/consensus/src/marshal/actor.rs
+++ b/consensus/src/marshal/actor.rs
@@ -701,8 +701,8 @@ where
 
     /// Returns a scheme suitable for verifying certificates at the given epoch.
     ///
-    /// Prefers an epoch-less certificate verifier if available, otherwise falls back
-    /// to the epoch-specific scheme.
+    /// Prefers a certificate verifier if available, otherwise falls back
+    /// to the scheme for the given epoch.
     fn get_scheme_certificate_verifier(&self, epoch: Epoch) -> Option<Arc<S>> {
         self.scheme_provider
             .certificate_verifier()
--- a/consensus/src/marshal/mod.rs
+++ b/consensus/src/marshal/mod.rs
@@ -82,17 +82,13 @@ pub trait SchemeProvider: Clone + Send + Sync + 'static {
     /// Return a certificate verifier that can validate certificates independent of epoch.
     ///
     /// This method allows implementations to provide a verifier that can validate
-    /// certificates from any epoch without requiring epoch-specific signing state. This
-    /// is useful for schemes that can verify certificates using long-lived cryptographic
-    /// material that remains valid across epoch boundaries.
-    ///
-    /// For example, [`bls12381_threshold::Scheme`](crate::simplex::signing_scheme::bls12381_threshold::Scheme)
-    /// maintains a constant public key across epochs and can verify certificates from any
-    /// epoch using that public key, even after the committee has rotated and the underlying
-    /// secret shares have changed.
+    /// certificates from any epoch (without epoch-specific state). For example,
+    /// [`bls12381_threshold::Scheme`](crate::simplex::signing_scheme::bls12381_threshold::Scheme)
+    /// maintains a static public key across epochs that can be used to verify certificates from any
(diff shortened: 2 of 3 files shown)


consensus/src/marshal/cache.rs
```diff
@@ -175,7 +175,8 @@ impl<
     async fn init_epoch(&mut self, epoch: Epoch) {
         let scheme = self
             .scheme_provider
-            .scheme(epoch)
+            .certificate_verifier()
+            .or_else(|| self.scheme_provider.scheme(epoch))
```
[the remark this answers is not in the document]
clabby: This is incorrect; `SchemeProvider::certificate_verifier` and `SchemeProvider::scheme` are bound to return the same `Scheme` impl and will have the same `certificate_codec_config` type.

For the sake of the archives, though, we may want to move over to `certificate_codec_config_unbounded`?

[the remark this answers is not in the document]
andresilva: Indeed, didn't notice that earlier. We can remove the entire `scheme_provider` from cache I think. Will update.

consensus/src/marshal/mod.rs
```diff
@@ -78,6 +78,24 @@ pub trait SchemeProvider: Clone + Send + Sync + 'static {
...
+    ///
+    /// For example, [`bls12381_threshold::Scheme`](crate::simplex::signing_scheme::bls12381_threshold::Scheme)
+    /// maintains a constant public key across epochs and can verify certificates from any
+    /// epoch using that public key, even after the committee has rotated and the underlying
+    /// secret shares have changed.
+    ///
+    /// The default implementation returns `None`. Callers should fall back to
+    /// `scheme(epoch)` for epoch-specific verification.
+    fn certificate_verifier(&self) -> Option<Arc<Self::Scheme>> {
+        None
```
patrick-ogrady, line 93: Should we just call `scheme` by default here? Then we wouldn't need to handle the `.unwrap_or_else` cases above?

patrick-ogrady: Admittedly, this would require accepting `epoch` which may be more confusing (if it isn't meant to require one).

patrick-ogrady: Just seems like we should be able to handle this boilerplate:
```
        let scheme = self
            .scheme_provider
            .certificate_verifier()
            .or_else(|| self.scheme_provider.scheme(epoch))
```

clabby: https://github.com/commonwarexyz/monorepo/pull/2287#discussion_r2550198061 <- conversation from the other PR

andresilva: The original API of `certificate_verifier` in #2287 did take an epoch as parameter precisely to implement the fallback in the trait. It made the API a bit less "honest", since the point of the certificate verifier is to be used with schemes that won't make a distinction wrt epoch. I can revert to that if you feel that's the way to go.

andresilva: Given that we now only need to handle the fallback in one place (i.e. marshal actor), I think it may be fine to keep as is.

consensus/src/marshal/actor.rs
```diff
@@ -597,7 +597,7 @@ where
                                 },
                                 Request::Finalized { height } => {
                                     let epoch = utils::epoch(self.epoch_length, height);
-                                    let Some(scheme) = self.scheme_provider.scheme(epoch) else {
+                                    let Some(scheme) = self.get_scheme_certificate_verifier(epoch) else {
```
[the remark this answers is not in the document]
andresilva: This is wrong, `certificate_codec_config_unbounded` is always more permissive (or as strict) as `certificate_codec_config`. `certificate_codec_config_unbounded` should never be used on untrusted input (coming from network).

consensus/src/marshal/mod.rs
patrick-ogrady: I think how you have it is fine/clear 👍

patrick-ogrady (approved):

PR #2088, opened 2025-11-02
Fuzz dkg

dnkolegov-ar:
This PR adds a fuzz test for DKG operations.

PR #2331, opened 2025-11-26
[cryptography] add fuzz tests for DKG

cronokirby:
This mainly consists of implementing arbitrary for the plan, and then conditionally exposing it using a feature flag, so that the fuzz target is a very simple wrapper, and we don't have to expose implementation details of the protocol.

The fuzz harness tests:
- various configurations of round, with increasing and decreasing (potentially) dealers and players
- that tampering with the inputs to signatures causes them to fail
- having players omit acks to particular dealers
- having dealers send bad shares to particular players
- having dealers put an incorrect reveal value in their log
- having a dealer use a different share, rather than their previous share
- having a dealer use a polynomial of the wrong degree.

The core of the harness logic is that given such a round configuration, some dealers are known to be bad, statically, and we assert that they don't get selected by the DKG. Furthermore, we may also realize that there are too many bad dealers for the DKG to succeed, in which case we assert that it fails. In all other cases, we expect an observer and all players to agree that the DKG is successful, and on what the output is, and that we can perform threshold signing with the result.

### DIFF
--- a/cryptography/src/bls12381/dkg.rs
+++ b/cryptography/src/bls12381/dkg.rs
@@ -1843,7 +1843,9 @@ mod test_plan {
                     // Apply BadShare perturbations
                     for (player, priv_msg) in &mut priv_msgs {
                         if let Some(i_player) = players.index(player) {
-                            if round.bad_shares.contains(&(i_dealer, i_player)) {
+                            // Convert position to key index
+                            let player_key_idx = round.players[i_player as usize];
+                            if round.bad_shares.contains(&(i_dealer, player_key_idx)) {
                                 priv_msg.share = Scalar::from_rand(&mut rng);
                             }
                         }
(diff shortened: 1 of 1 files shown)


dnkolegov-ar:
This is the PR I created several weeks ago. Feel free to merge into this branch or close it https://github.com/commonwarexyz/monorepo/pull/2088

cryptography/src/bls12381/dkg.rs
```diff
@@ -1361,31 +1360,26 @@ pub fn deal_anonymous<V: Variant>(
     (output.public().clone(), shares.values().to_vec())
 }
 
-#[cfg(test)]
-mod test {
+#[cfg(any(feature = "fuzz", test))]
```
patrick-ogrady, line 1433: When you discussed this pattern this morning, I was afraid it required feature flags inline. I think (at a high-level) what you have here is totally fine.

PR #2334, opened 2025-11-26
[Storage] Add dirty/clean type state to `bitmap::authenticated`, `adb::current`

danlaine:
* Expose `Dirty`/`Clean` state in `bitmap::authenticated`, `adb::current::{ordered, unordered}`, and for each type, implement applicable traits, and move methods to applicable `impl`s (for Dirty, Clean, either.) For each type, the root is now cached in the clean state.
* Update `fn merkleize(self) -> Self::Clean` to `async fn merkleize(self) -> Result<Self::Clean, Error>` for `DirtyStore` trait to accommodate current databases, whose `merkleize` is `async` and may error.
* Add `AnyExt` wrapper for `Any` databases which exposes a "traditional" CRUD key-value store without exposing merkleization status. It handles dirty/clean state transitions internally.
* Remove `adb::Error::UncommittedOperations` 🥳 
* Update fuzz tests to accommodate API changes.
* Benchmark trait bound nits

### DIFF
--- a/storage/src/adb/any/ext.rs
+++ b/storage/src/adb/any/ext.rs
@@ -7,7 +7,7 @@
 use super::{CleanAny, DirtyAny};
 use crate::{
     adb::{
-        store::{Batchable, DirtyStore, LogStore, LogStorePrunable},
+        store::{Batchable, CleanStore, DirtyStore, LogStore, LogStorePrunable},
         Error,
     },
     mmr::Location,
@@ -181,6 +181,16 @@ where
 impl<A> Batchable for AnyExt<A>
 where
     A: CleanAny,
-    <A as LogStore>::Value: Clone,
+    <A as CleanStore>::Dirty: Batchable<Key = A::Key, Value = A::Value>,
 {
+    async fn write_batch(
+        &mut self,
+        iter: impl Iterator<Item = (Self::Key, Option<Self::Value>)>,
+    ) -> Result<(), Error> {
+        self.ensure_dirty();
+        match self.inner.as_mut().expect("wrapper should never be empty") {
+            State::Dirty(dirty) => dirty.write_batch(iter).await,
+            _ => unreachable!("ensure_dirty guarantees Dirty state"),
+        }
+    }
 }


storage/fuzz/fuzz_targets/mmr_bitmap.rs
```diff
@@ -179,23 +214,30 @@ fn fuzz(input: FuzzInput) {
...
-                    BitMap::<_, CHUNK_SIZE>::restore_pruned(
+                    let bitmap = CleanAuthenticatedBitMap::<_, CHUNK_SIZE>::restore_pruned(
                         context.clone(),
                         "fuzz_mmr_bitmap_test_partition",
                         None,
                         &mut hasher,
                     )
                     .await
                     .unwrap();
+                    Bitmap::Clean(bitmap)
```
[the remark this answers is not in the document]
danlaine: This case previously called `restore_pruned` without actually assigning it to anything so this wasn't an issue before. Updated the code to update the bit count and pruned_bits after restore_pruned.

storage/src/adb/any/ext.rs
```diff
@@ -0,0 +1,186 @@
...
+
+    /// Merkleize if in dirty state, ensuring we're in clean state.
+    async fn ensure_clean(&mut self) -> Result<(), Error> {
+        let state = self.inner.take().expect("wrapper should never be empty");
+        self.inner = Some(match state {
+            State::Clean(clean) => State::Clean(clean),
+            State::Dirty(dirty) => State::Clean(dirty.merkleize().await?),
+        });
+        Ok(())
+    }
```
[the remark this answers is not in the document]
danlaine: We have implicit invariants in a lot of places that an error during a mutable storage operation should be considered fatal. So I think this is OK since we do the same elsewhere?

storage/src/adb/current/ordered.rs
```diff
@@ -507,75 +510,39 @@ impl<
...
-                    self.status.set_bit(*loc, false);
-                }
-                self.status.push(append);
-                r = true;
-            })
-            .await?;
+        self.any.apply_commit_op(commit_op).await?;
 
-        Ok(r)
+        Ok(status)
```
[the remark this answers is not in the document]
danlaine: Same as [here](https://github.com/commonwarexyz/monorepo/pull/2334#discussion_r2590675321) about not handling state cleanup on error during mutable method.

storage/src/adb/any/ext.rs
```diff
@@ -0,0 +1,186 @@
...
+        Error,
+    },
+    mmr::Location,
+    store::{Store as StoreTrait, StoreDeletable, StoreMut, StorePersistable},
+};
+
+/// An extension wrapper for [CleanAny] databases that provides a traditional mutable key-value
+/// store interface by internally handling Clean/Dirty state transitions.
+pub struct AnyExt<A: CleanAny> {
+    // Invariant: always Some
```
roberto-bayardo, line 20: Interesting pattern. I guess we need the option to allow for taking the contents?

storage/src/adb/any/ext.rs
```diff
@@ -0,0 +1,186 @@
...
+        // Merkleize before prune
+        self.ensure_clean().await?;
+        match self.inner.as_mut().expect("wrapper should never be empty") {
+            State::Clean(clean) => clean.prune(prune_loc).await,
+            _ => unreachable!("ensure_clean guarantees Clean state"),
+        }
+    }
+}
+
+impl<A> Batchable for AnyExt<A>
```
roberto-bayardo, line 181: I guess this would override the optimized implementations if any provided by the wrapped db?

roberto-bayardo: I would expect you'd see a performance regression for any::unordered::* in the generation benchmarks if so.

storage/src/adb/current/ordered.rs
```diff
@@ -92,6 +96,134 @@ pub enum ExclusionProofInfo<K: Array, V: Value, const N: usize> {
     /// completely empty (having no operations at all against which to prove).
     DbEmpty,
 }
+impl<
```
roberto-bayardo, line 100: add a newline above?

storage/src/bitmap/authenticated.rs
```diff
@@ -201,178 +404,36 @@ impl<D: Digest, const N: usize> BitMap<D, N> {
...
+    ///
+    /// The chunk containing `bit` and all subsequent chunks are retained. All chunks
+    /// before it are pruned from the bitmap and the underlying MMR.
+    ///
+    /// If `bit` equals the bitmap length, this prunes all complete chunks while retaining
+    /// the empty trailing chunk, preparing the bitmap for appending new data.
+    ///
+    /// # Warning
+    ///
+    /// - Returns [Error::DirtyState] if there are unmerkleized updates.
```
roberto-bayardo: I think this warning is obsolete?

roberto-bayardo: If so I think you can get rid of the Error::DirtyState error variant entirely?

storage/src/adb/any/ext.rs
danlaine: Yup. Without `Option` we'd have to `std::mem::take` or `std::mem::replace` the underlying database which isn't really feasible.

storage/src/adb/current/mod.rs
```diff
@@ -206,10 +194,10 @@ async fn range_proof<
 /// Performs merkleization of a grafted bitmap.
 async fn merkleize_grafted_bitmap<H, const N: usize>(
     hasher: &mut StandardHasher<H>,
-    status: &mut BitMap<H::Digest, N>,
+    status: DirtyBitMap<H::Digest, N>,
     mmr: &impl crate::mmr::storage::Storage<H::Digest>,
     grafting_height: u32,
-) -> Result<(), Error>
+) -> Result<CleanBitMap<H::Digest, N>, Error>
```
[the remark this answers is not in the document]
danlaine: I think this is OK as it is.

storage/src/adb/current/ordered.rs
danlaine: done

storage/src/bitmap/authenticated.rs
danlaine: Removed all references to DirtyState 🥳

storage/src/adb/any/ext.rs
danlaine: Good catch -- updated to call the underlying `write_batch` to use the specialized implementation (if any)

storage/src/adb/current/mod.rs
```diff
@@ -206,10 +194,10 @@ async fn range_proof<
 /// Performs merkleization of a grafted bitmap.
```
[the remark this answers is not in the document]
danlaine: updated comment

roberto-bayardo (approved):

PR #2335, opened 2025-11-26
[consensus/simplex] Gracefully shutdown batcher and resolver

clabby:
## Overview

Gracefully shuts down the `batcher` and `resolver` actors on `context.stopped()` + adds a test for graceful shutdown support.

### DIFF
--- a/consensus/src/simplex/actors/voter/actor.rs
+++ b/consensus/src/simplex/actors/voter/actor.rs
@@ -752,6 +752,8 @@ impl<
             let view;
             select! {
                 _ = &mut shutdown => {
+                    debug!("context shutdown, stopping voter");
+
                     // Close journal
                     self.journal
                         .take()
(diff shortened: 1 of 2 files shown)


cronokirby:
Thoughts on debug vs info for the tracing?

clabby:
> Should we add the shutdown to `voter` too?

`voter` already had a shutdown handler:

https://github.com/commonwarexyz/monorepo/blob/f7946ab0dff33db0c210b65291fb3282eca345e4/consensus/src/simplex/actors/voter/actor.rs#L754-L763

clabby:
> Thoughts on debug vs info for the tracing?

I'd be open to `info`, though kept it as `debug` to align with the existing log levels on shutdown information. If we do change it, IMO should swap to `info` for shutdowns across the board.

patrick-ogrady (approved):
Rationale:

1) Each actor should be responsible for coordinating its own shutdown (be explicit rather than assuming channels we communicate over will close and lead us to close).
2) By using our select!, we can safely assert that sub-actors don't finish before graceful shutdown is called. If a handle is aborted, we will drop tasks rather than complete them (if the root task is dropped first, no asserts will be hit -> don't love this behavior but doesn't seem wrong)
3) During graceful shutdown, we should log debug (as there is no action to take/no unexpected behavior).

PR #2338, opened 2025-11-26
[Storage] Factor `adb::store::Db` trait into multiple traits; relax contiguous journal bounds

danlaine:
Factoring out `adb::store::Db` is required for removal of clean -> clean state transitions in the MMR because dirty/clean databases will have different methods defined on them. For example, an ADB whose underlying MMR is dirty (root digest not calculated) can't be committed because the MMR node digests can't be written until they're computed. Open to suggestions on what the cleanest way to bundle these methods is. I suspect we'll end up tweaking this further but this is a step in the right direction.

This PR also relaxes the type bounds on contiguous journals so that non-mutable (non-persistable) journals can be used for operations that don't require mutability (persistence).

### DIFF
--- a/cryptography/src/bls12381/primitives/group.rs
+++ b/cryptography/src/bls12381/primitives/group.rs
@@ -14,11 +14,11 @@ use super::variant::Variant;
 #[cfg(not(feature = "std"))]
 use alloc::{vec, vec::Vec};
 use blst::{
-    blst_bendian_from_scalar, blst_expand_message_xmd, blst_fp12, blst_fr, blst_fr_add,
-    blst_fr_from_scalar, blst_fr_from_uint64, blst_fr_inverse, blst_fr_mul, blst_fr_sub,
-    blst_hash_to_g1, blst_hash_to_g2, blst_keygen, blst_p1, blst_p1_add_or_double, blst_p1_affine,
-    blst_p1_compress, blst_p1_from_affine, blst_p1_in_g1, blst_p1_is_inf, blst_p1_mult,
-    blst_p1_to_affine, blst_p1_uncompress, blst_p1s_mult_pippenger,
+    blst_bendian_from_fp12, blst_bendian_from_scalar, blst_expand_message_xmd, blst_fp12, blst_fr,
+    blst_fr_add, blst_fr_from_scalar, blst_fr_from_uint64, blst_fr_inverse, blst_fr_mul,
+    blst_fr_sub, blst_hash_to_g1, blst_hash_to_g2, blst_keygen, blst_p1, blst_p1_add_or_double,
+    blst_p1_affine, blst_p1_compress, blst_p1_from_affine, blst_p1_in_g1, blst_p1_is_inf,
+    blst_p1_mult, blst_p1_to_affine, blst_p1_uncompress, blst_p1s_mult_pippenger,
     blst_p1s_mult_pippenger_scratch_sizeof, blst_p2, blst_p2_add_or_double, blst_p2_affine,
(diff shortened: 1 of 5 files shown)


storage/src/adb/store/mod.rs
```diff
@@ -176,20 +171,34 @@ pub trait Db<K: Array, V: Codec> {
...
+    /// pruned and those that are not yet committed.
+    fn op_count(&self) -> Location;
+
+    /// Return the inactivity floor location. This is the location before which all operations are
+    /// known to be inactive. Operations before this point can be safely pruned.
+    fn inactivity_floor_loc(&self) -> Location;
+}
+
+/// A key-value store backed by a log that can be committed and persisted.
+pub trait PersistedKeyValueStore<K: Array, V: Codec>: KeyValueStore<K, V> {
```
roberto-bayardo: thought I left this suggestion before but I don't see it, so apologies if this is duplicate:

Should this be Persistable instead of Persisted?  (It's not persisted if it's dirty...)

storage/src/adb/store/mod.rs
```diff
@@ -176,20 +171,34 @@ pub trait Db<K: Array, V: Codec> {
...
+    /// Get the metadata associated with the last commit, or None if no commit has been made.
+    fn get_metadata(&self) -> impl Future<Output = Result<Option<V>, Error>>;
 
     /// Commit any pending operations to the database, ensuring their durability upon return from
-    /// this function. Also raises the inactivity floor according to the schedule. Returns the
-    /// `(start_loc, end_loc]` location range of committed operations. The end of the returned range
-    /// includes the commit operation itself, and hence will always be equal to `op_count`.
+    /// this function. Returns the `(start_loc, end_loc]` location range of committed operations.
+    /// The end of the returned range includes the commit operation itself, and hence will always be
```
roberto-bayardo, line 201: is floor raising no longer part of the contract or did you think it was just too much detail here?

storage/src/adb/store/mod.rs
```diff
@@ -176,20 +171,34 @@ pub trait Db<K: Array, V: Codec> {
...
+    /// equal to `Log::op_count`.
```
roberto-bayardo: `Self::op_count` ?

roberto-bayardo: oh hmm I guess this isn't a Log type.  Feels odd to reference a method that doesn't exist on this type?

storage/src/adb/store/mod.rs
roberto-bayardo: Since this is always backed by a log, should it just specialize the Log trait?

danlaine: Renamed `PersistableKeyValueStore`

EDIT: These methods are now defined on `Db`

danlaine: Fair point. Based on this and https://github.com/commonwarexyz/monorepo/pull/2338#discussion_r2578165359, I think just move the trait methods from `PersistedKeyValueStore` to `Db`

storage/src/adb/store/mod.rs
danlaine: Re-added the line about raising the inactivity floor

storage/src/adb/store/mod.rs
danlaine: This is now defined on `Db` which is `Log`

roberto-bayardo (approved):

roberto-bayardo (approved):

danlaine:
Replaced by #2360 and #2370

PR #2341, opened 2025-11-26
[storage/adb] add metadata to fixed operation commit

roberto-bayardo:
Operation types that allow for variable-length values support a convenient "metadata" to be provided with a commit.  This PR adds that capability to fixed-length operation types, and then adds `get_metdata` to the Db trait.

The main downside is that for fixed-length value databases with keys less than 9 bytes in length, this adds an extra `9.saturating_sub(len(Key))` bytes per operation. (For example, if your key is a U64, this requires 1 byte wasted per non-commit operation.)

This PR also changes get_metadata behavior to return only the metadata instead of a (metadata, location of last commit) pair. Instead, the `commit` operation of the Db trait is changed to return the start_range of the committed location to facilitate range proof generation and syncing of the new operations.

### DIFF
--- a/storage/src/adb/any/ordered/mod.rs
+++ b/storage/src/adb/any/ordered/mod.rs
@@ -19,7 +19,7 @@ use crate::{
 };
 use commonware_cryptography::{Digest, DigestOf, Hasher};
 use commonware_runtime::{Clock, Metrics, Storage};
-use core::num::NonZeroU64;
+use core::{num::NonZeroU64, ops::Range};
 use tracing::debug;
 
 pub mod fixed;
@@ -822,10 +822,7 @@ impl<
         Ok(r)
     }
 
-    async fn commit(
-        &mut self,
-        metadata: Option<Value<C::Item>>,
-    ) -> Result<(Location, Location), Error> {
+    async fn commit(&mut self, metadata: Option<Value<C::Item>>) -> Result<Range<Location>, Error> {
         let start_loc = if let Some(last_commit) = self.last_commit {
             last_commit + 1
         } else {
@@ -838,7 +835,7 @@ impl<
         self.apply_commit_op(C::Item::new_commit_floor(metadata, inactivity_floor_loc))
             .await?;
 
-        Ok((start_loc, self.op_count()))
+        Ok(start_loc..self.op_count())
     }
 
     async fn sync(&mut self) -> Result<(), Error> {
--- a/storage/src/adb/current/ordered.rs
+++ b/storage/src/adb/current/ordered.rs
@@ -21,6 +21,7 @@ use commonware_codec::{CodecFixed, FixedSize};
 use commonware_cryptography::{DigestOf, Hasher};
 use commonware_runtime::{Clock, Metrics, Storage as RStorage};
 use commonware_utils::Array;
+use core::ops::Range;
 use std::num::NonZeroU64;
 
 /// A key-value ADB based on an MMR over its log of operations, supporting key exclusion proofs and
@@ -570,7 +571,7 @@ impl<
         Ok(r)
     }
 
-    async fn commit(&mut self, metadata: Option<V>) -> Result<(Location, Location), Error> {
+    async fn commit(&mut self, metadata: Option<V>) -> Result<Range<Location>, Error> {
         let start_loc = if let Some(last_commit) = self.any.last_commit {
             last_commit + 1
         } else {
@@ -587,7 +588,7 @@ impl<
         // Prune bits that are no longer needed because they precede the inactivity floor.
         self.status.prune_to_bit(*self.any.inactivity_floor_loc())?;
 
-        Ok((start_loc, self.op_count()))
+        Ok(start_loc..self.op_count())
     }
 
     async fn sync(&mut self) -> Result<(), Error> {
--- a/storage/src/adb/current/unordered.rs
+++ b/storage/src/adb/current/unordered.rs
@@ -22,6 +22,7 @@ use commonware_codec::{CodecFixed, FixedSize};
 use commonware_cryptography::{DigestOf, Hasher};
 use commonware_runtime::{Clock, Metrics, Storage as RStorage};
 use commonware_utils::Array;
+use core::ops::Range;
 use std::num::NonZeroU64;
 
 /// A key-value ADB based on an MMR over its log of operations, supporting authentication of whether
@@ -391,7 +392,7 @@ impl<
         Ok(true)
     }
 
-    async fn commit(&mut self, metadata: Option<V>) -> Result<(Location, Location), Error> {
+    async fn commit(&mut self, metadata: Option<V>) -> Result<Range<Location>, Error> {
         let start_loc = if let Some(last_commit) = self.any.last_commit {
             last_commit + 1
         } else {
@@ -409,7 +410,7 @@ impl<
         // Prune bits that are no longer needed because they precede the inactivity floor.
         self.status.prune_to_bit(*self.any.inactivity_floor_loc())?;
 
-        Ok((start_loc, self.op_count()))
+        Ok(start_loc..self.op_count())
     }
 
     async fn sync(&mut self) -> Result<(), Error> {
@@ -522,8 +523,8 @@ pub mod test {
             assert!(db.create(k1, v1).await.unwrap());
             assert_eq!(db.get(&k1).await.unwrap().unwrap(), v1);
             let range = db.commit(None).await.unwrap();
-            assert_eq!(range.0, 0);
-            assert_eq!(range.1, 3);
+            assert_eq!(range.start, 0);
+            assert_eq!(range.end, 3);
             assert!(db.get_metadata().await.unwrap().is_none());
             assert_eq!(db.op_count(), 3); // 1 update, 1 commit, 1 move.
             let root1 = db.root(&mut hasher).await.unwrap();
@@ -541,8 +542,8 @@ pub mod test {
             assert!(db.delete(k1).await.unwrap());
             let metadata = Sha256::hash(&1u64.to_be_bytes());
             let range = db.commit(Some(metadata)).await.unwrap();
-            assert_eq!(range.0, 3);
-            assert_eq!(range.1, 5);
+            assert_eq!(range.start, 3);
+            assert_eq!(range.end, 5);
 
             assert_eq!(db.op_count(), 5); // 1 update, 2 commits, 1 move, 1 delete.
             assert_eq!(db.get_metadata().await.unwrap().unwrap(), metadata);
--- a/storage/src/adb/store/mod.rs
+++ b/storage/src/adb/store/mod.rs
@@ -100,7 +100,7 @@ use crate::{
 use commonware_codec::{Codec, Read};
 use commonware_runtime::{buffer::PoolRef, Clock, Metrics, Storage as RStorage};
 use commonware_utils::Array;
-use core::future::Future;
+use core::{future::Future, ops::Range};
 use std::num::{NonZeroU64, NonZeroUsize};
 use tracing::debug;
 
@@ -196,7 +196,7 @@ pub trait Db<K: Array, V: Codec> {
     fn commit(
         &mut self,
         metadata: Option<V>,
-    ) -> impl Future<Output = Result<(Location, Location), Error>>;
+    ) -> impl Future<Output = Result<Range<Location>, Error>>;
 
     /// Sync all database state to disk. While this isn't necessary to ensure durability of
     /// committed operations, periodic invocation may reduce memory usage and the time required to
@@ -430,7 +430,7 @@ where
         Ok(true)
     }
 
-    async fn commit(&mut self, metadata: Option<V>) -> Result<(Location, Location), Error> {
+    async fn commit(&mut self, metadata: Option<V>) -> Result<Range<Location>, Error> {
         let start_loc = if let Some(last_commit) = self.last_commit {
             last_commit + 1
         } else {
@@ -463,7 +463,7 @@ where
         // Commit the log to ensure this commit is durable.
         self.log.commit().await?;
 
-        Ok((start_loc, self.op_count()))
+        Ok(start_loc..self.op_count())
     }
 
     async fn sync(&mut self) -> Result<(), Error> {
@@ -560,8 +560,8 @@ mod test {
             // Test calling commit on an empty db which should make it (durably) non-empty.
             let metadata = vec![1, 2, 3];
             let range = db.commit(Some(metadata.clone())).await.unwrap();
-            assert_eq!(range.0, 0);
-            assert_eq!(range.1, 1);
+            assert_eq!(range.start, 0);
+            assert_eq!(range.end, 1);
             assert_eq!(db.op_count(), 1);
             assert!(matches!(db.prune(db.inactivity_floor_loc()).await, Ok(())));
             assert_eq!(db.get_metadata().await.unwrap(), Some(metadata.clone()));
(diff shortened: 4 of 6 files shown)


storage/src/adb/operation/fixed/ordered.rs
```diff
@@ -102,35 +110,41 @@ impl<K: Array + Ord, V: CodecFixed<Cfg = ()>> Keyed for Operation<K, V> {
...
 }
 
 impl<K: Array + Ord, V: CodecFixed> Committable for Operation<K, V> {
     fn is_commit(&self) -> bool {
-        matches!(self, Self::CommitFloor(_))
+        matches!(self, Self::CommitFloor(_, _))
+    }
+}
+
+impl<K: Array, V: CodecFixed<Cfg = ()>> Ordered for Operation<K, V> {
```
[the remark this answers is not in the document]
roberto-bayardo: Array implies Ord.

storage/src/adb/operation/fixed/ordered.rs
```diff
@@ -200,7 +210,17 @@ impl<K: Array + Ord, V: CodecFixed> Display for Operation<K, V> {
                     hex(&data.value.encode())
                 )
             }
-            Self::CommitFloor(loc) => write!(f, "[commit with inactivity floor: {loc}]"),
+            Self::CommitFloor(metadata, loc) => {
+                if let Some(value) = metadata {
+                    write!(
+                        f,
+                        "[commit {} with inactivity floor: {loc}]",
+                        hex(&value.encode())
```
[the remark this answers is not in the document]
roberto-bayardo: I think this is clear as is.

storage/src/adb/current/ordered.rs
```diff
@@ -422,10 +427,10 @@ impl<
 
                 (info.loc, info.chunk, element)
             }
-            ExclusionProofInfo::Commit((loc, chunk)) => {
+            ExclusionProofInfo::Commit((loc, metadata, chunk)) => {
```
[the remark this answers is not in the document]
roberto-bayardo: I don't see anything getting shadowed here.

examples/sync/src/bin/server.rs
danlaine: We can actually avoid cloning the operations altogether by recording the length before passing to `add_oeprations`. I should've done this originally 😅 

```suggestion
    DB: Syncable,
    E: Storage + Clock + Metrics + RngCore,
{
    let mut last_time = state.last_operation_time.write().await;
    let now = context.current();

    if now.duration_since(*last_time).unwrap_or(Duration::ZERO) >= config.op_interval {
        *last_time = now;

        // Generate new operations
        let new_operations =
            DB::create_test_operations(config.ops_per_interval, context.next_u64());
        let new_operations_len = new_operations.len();

        // Add operations to database and get the new root
        let root = {
            let mut database = state.database.write().await;
            if let Err(err) = DB::add_operations(&mut *database, new_operations).await {
                error!(?err, "failed to add operations to database");
            }
            DB::root(&*database)
        };

        state.ops_counter.inc_by(new_operations_len as u64);
        let root_hex = root
            .as_ref()
            .iter()
            .map(|b| format!("{b:02x}"))
            .collect::<String>();
        info!(
            new_operations_len,
```

examples/sync/src/bin/server.rs
```diff
@@ -289,6 +290,7 @@ async fn handle_client<DB, E>(
 ) -> Result<(), Box<dyn std::error::Error>>
 where
     DB: Syncable + Send + Sync + 'static,
+    DB::Operation: Decode + Read<Cfg = ()> + Clone + Send,
```
danlaine:
```suggestion
    DB::Operation:Read<Cfg = ()> + Send,
```

examples/sync/src/bin/server.rs
```diff
@@ -396,6 +398,7 @@ async fn run_helper<DB, E>(
...
```
danlaine:
```suggestion
    DB::Operation:Read<Cfg = ()> + Send,
```

examples/sync/src/databases/mod.rs
```diff
@@ -56,7 +55,10 @@ pub trait Syncable {
     ) -> impl Future<Output = Result<(), adb::Error>>;
 
     /// Commit pending operations to the database.
-    fn commit(&mut self) -> impl Future<Output = Result<(), adb::Error>>;
+    fn commit(
```
danlaine: We never pass a non-None metadata in this example so we can leave this function signature as is and just pass `None` when we call `commit` on the underlying implementation.

storage/fuzz/fuzz_targets/adb_any_fixed_sync.rs
```diff
@@ -168,7 +168,7 @@ fn fuzz(input: FuzzInput) {
                 }
 
                 Operation::Commit => {
-                    db.commit().await.expect("Commit should not fail");
+                    db.commit(None).await.expect("Commit should not fail");
```
danlaine: Should we add a fuzz operation to check that the most recently written metadata is correct? We could have the metadata be a counter that tracks the number of commits, or something.

storage/src/adb/any/ordered/mod.rs
```diff
@@ -32,22 +32,15 @@ type AuthenticatedLog<E, C, H, S = Clean<DigestOf<H>>> = authenticated::Journal<
...
     /// Return a new commit-floor operation variant.
-    fn new_commit_floor(inactivity_floor_loc: Location) -> Self;
-
-    /// Return this operation's key data, or None if this operation variant doesn't have any.
-    fn key_data(&self) -> Option<&KeyData<Self::Key, Self::Value>>;
-
-    /// Convert this operation into its key data, or None if this operation variant doesn't have
-    /// any.
-    fn into_key_data(self) -> Option<KeyData<Self::Key, Self::Value>>;
+    fn new_commit_floor(metadata: Option<Value<Self>>, inactivity_floor_loc: Location) -> Self;
```
danlaine: This is super nit-picky, but should we keep this aligned with using `Self::Value` like in `fn new_update`?

I realize I just added `type Value<T> = <T as Keyed>::Value;` a few days ago but I did that so that I could do `Value<C::Item>` instead of `<C::Item as Keyed>::Value` in a bunch of places below where `C::Item` is an `Operation`.

Or we could change `new_update` to use `Value<Self>`.

```suggestion
    fn new_commit_floor(metadata: Option<Self::Value>, inactivity_floor_loc: Location) -> Self;
```

storage/src/adb/operation/fixed/ordered.rs
```diff
@@ -22,73 +22,81 @@ pub enum Operation<K: Array + Ord, V: CodecFixed> {
...
-
-    /// Asserts that the size of `Self` is greater than the minimum operation size.
-    #[inline(always)]
-    const fn assert_valid_size() {
-        assert!(
-            Self::SIZE >= Self::_MIN_OPERATION_LEN,
-            "array size too small for commit op"
-        );
+impl<K: Array, V: CodecFixed> Operation<K, V> {
+    // Commit op has a context byte, an option indicator, a metadata value, and a u64 location.
```
danlaine, line 29: Nit: I think this is the more commonly used word? I think it should be clear what this is either way.

```suggestion
    // Commit op has a discriminant, an option indicator, a metadata value, and a u64 location.
```

storage/src/adb/operation/fixed/unordered.rs
```diff
@@ -21,42 +21,41 @@ pub enum Operation<K: Array, V: CodecFixed> {
...
-
-    /// Asserts that the size of `Self` is greater than the minimum operation size.
-    #[inline(always)]
-    const fn assert_valid_size() {
-        assert!(
-            Self::SIZE >= Self::_MIN_OPERATION_LEN,
-            "array size too small for commit op"
-        );
+impl<K: Array + Ord, V: CodecFixed> Operation<K, V> {
+    // Commit op has a context byte, an option indicator, a metadata value, and a u64 location.
```
danlaine, line 28: Same nit about context byte vs discriminant

storage/src/adb/operation/fixed/unordered.rs
```diff
@@ -21,42 +21,41 @@ pub enum Operation<K: Array, V: CodecFixed> {
...
+    const COMMIT_OP_SIZE: usize = 1 + 1 + V::SIZE + u64::SIZE;
+
+    // Update op has a context byte, a key, and a value.
+    const UPDATE_OP_SIZE: usize = 1 + K::SIZE + V::SIZE;
+
+    const DELETE_OP_SIZE: usize = 1 + K::SIZE;
```
danlaine, line 35:
```suggestion

    // Delete op has a discriminant and a key.
    const DELETE_OP_SIZE: usize = 1 + K::SIZE;
```

storage/src/adb/operation/mod.rs
```diff
@@ -335,18 +410,27 @@ mod tests {
...
+        let key_data = {
+            let key = key.clone();
+            let value = value.clone();
+            let next_key = key.clone();
+            KeyData {
+                key,
+                value,
+                next_key,
+            }
+        };
```
danlaine:
```suggestion
        let key_data = KeyData {
            key: key.clone(),
            value: value.clone(),
            next_key: key.clone(),
        };
```

danlaine (commented):
Would you mind adding tests that ensure `get_metadata` doesn't return stale metedata? i.e. Do a commit with `Some` metadata, then another with `None` metadata or different metadata, and ensure that the newest metadata is returned and not the old one.

roberto-bayardo:
> Would you mind adding tests that ensure `get_metadata` doesn't return stale metedata? i.e. Do a commit with `Some` metadata, then another with `None` metadata or different metadata, and ensure that the newest metadata is returned and not the old one.

Added this to the "multiple_commits" tests in adb::any::(un)ordered::fixed

storage/src/adb/operation/fixed/ordered.rs
roberto-bayardo: isn't context-byte more explicit? Personally I would not know what a "discriminant" means in this context, so I asked the AI what would be better:

> Both terms are accurate here. "Discriminant" might be clearer to Rust developers, while "context byte" emphasizes that it's providing context for deserialization. You could use either term, or even say "discriminant byte" or "variant tag" to be more explicit.

storage/src/adb/operation/mod.rs
roberto-bayardo: done

storage/src/adb/operation/fixed/unordered.rs
roberto-bayardo: done (though stuck with "context byte")

storage/src/adb/any/ordered/mod.rs
roberto-bayardo: Not that nit-picky, consistency is good. I (or the AI?) should have stuck with Self::Value here. Changed.

examples/sync/src/databases/mod.rs
roberto-bayardo: done

examples/sync/src/bin/server.rs
roberto-bayardo: done

examples/sync/src/bin/server.rs
roberto-bayardo: done

examples/sync/src/bin/server.rs
roberto-bayardo: done

storage/fuzz/fuzz_targets/adb_any_fixed_sync.rs
roberto-bayardo: done, though I can't run fuzz tests locally for some reason (getting some verbose linker error?). Let's see if it passes on the CI...

storage/fuzz/fuzz_targets/adb_any_fixed_sync.rs
```diff
@@ -212,7 +233,7 @@ fn fuzz(input: FuzzInput) {
 
                     db = Any::<_, Key, Value, Sha256, TwoCap>::init(
                         context.clone(),
-                        test_config("src"),
```
roberto-bayardo, line 215: This seemed to be a bug in the fuzz test?

storage/src/adb/store/mod.rs
```diff
@@ -184,11 +187,13 @@ pub trait Db<K: Array, V: Codec> {
     fn delete(&mut self, key: K) -> impl Future<Output = Result<bool, Error>>;
 
     /// Commit any pending operations to the database, ensuring their durability upon return from
-    /// this function. Also raises the inactivity floor according to the schedule.
+    /// this function. Also raises the inactivity floor according to the schedule. Returns the
+    /// starting location of the committed range of operations. (The end of the range will be
+    /// `op_count`.)
```
roberto-bayardo: We might consider returning a pair of values representing the range of committed operations, where the second value is always op_count().

storage/src/adb/operation/fixed/unordered.rs
```diff
@@ -21,42 +21,42 @@ pub enum Operation<K: Array, V: CodecFixed> {
...
+    } else {
+        b
     }
 }
 
-impl<K: Array, V: CodecFixed> CodecFixedSize for Operation<K, V> {
-    const SIZE: usize = u8::SIZE + K::SIZE + V::SIZE;
+impl<K: Array + Ord, V: CodecFixed> CodecFixedSize for Operation<K, V> {
+    // Make sure operation array is large enough to hold the maximum of all ops.
+    const SIZE: usize = max(Self::UPDATE_OP_SIZE, Self::COMMIT_OP_SIZE);
```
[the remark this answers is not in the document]
roberto-bayardo: Array trait already includes Ord.

storage/src/adb/operation/fixed/ordered.rs
danlaine: I was going from [here](https://doc.rust-lang.org/reference/items/enumerations.html#custom-discriminant-values-for-fieldless-enumerations) but I think it's apparent what the byte is with either name :+1:

storage/src/adb/current/unordered.rs
```diff
@@ -141,8 +141,8 @@ impl<
     }
 
     /// Commit pending operations to the adb::any, ensuring their durability upon return from this
-    /// function.
-    async fn commit_ops(&mut self) -> Result<(), Error> {
+    /// function. Returns the start location of the committed range.
```
danlaine: We return `()` here?

storage/fuzz/fuzz_targets/adb_any_fixed_sync.rs
```diff
@@ -181,9 +199,12 @@ fn fuzz(input: FuzzInput) {
                     if db.op_count() == 0 {
                         continue;
                     }
-                    db.commit()
+                    let mut commit_id = [0u8; 32];
+                    commit_id[..8].copy_from_slice(&input.commit_counter.to_be_bytes());
+                    input.commit_counter += 1;
+                    db.commit(Some(FixedBytes::new(commit_id)))
```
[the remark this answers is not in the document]
danlaine: SyncFull and Commit both increment before committing

storage/src/adb/store/mod.rs
roberto-bayardo: this is now done.

storage/src/adb/current/unordered.rs
roberto-bayardo: fixed

storage/src/adb/store/mod.rs
```diff
@@ -184,11 +187,16 @@ pub trait Db<K: Array, V: Codec> {
...
+    /// `(start_loc, end_loc]` location range of committed operations. The end of the returned range
+    /// includes the commit operation itself, and hence will always be equal to `op_count`.
     ///
     /// Failures after commit (but before `sync` or `close`) may still require reprocessing to
     /// recover the database on restart.
-    fn commit(&mut self) -> impl Future<Output = Result<(), Error>>;
+    fn commit(
+        &mut self,
+        metadata: Option<V>,
+    ) -> impl Future<Output = Result<(Location, Location), Error>>;
```
danlaine: I think it might be a bit more Rusty to return `Range<Location>` here. That's also the type we use in the ADB state sync code to represent the range of live operations to be fetched/applied to the log.

roberto-bayardo: right, of course!  changed.

PR #2346, opened 2025-11-28
Fix fuzz tests for utils lib

dnkolegov-ar:
This PR fixes fuzz tests for some utils' functions


utils/fuzz/fuzz_targets/lib_functions.rs
patrick-ogrady, line 1: Were the tests removed failing for some reason?

dnkolegov-ar: No, just not all functions/types were covered

Issue #1996, opened 2025-10-24
Potential panic at bls12381/dkg/arbiter.rs:L211


dnkolegov-ar:
Arbiter's `disqualify` lets the caller [add](https://github.com/commonwarexyz/monorepo/blob/8c07397a10630e0f6cda8aa8abd4047b6d682ffc/cryptography/src/bls12381/dkg/arbiter.rs#L106) arbitrary public keys to the disqualification set without checking membership, so `finalization` [panics](https://github.com/commonwarexyz/monorepo/blob/8c07397a10630e0f6cda8aa8abd4047b6d682ffc/cryptography/src/bls12381/dkg/arbiter.rs#L211) when it unwraps the dealer index for a spoofed participant, letting an attacker crash the DKG round.

cronokirby:
Yeah, we should make sure that unknown public keys are ignored completely here.

dnkolegov-ar:
PoC

```
//! Orchestrator of the DKG/Resharing procedure.
//!
//! # Deployment Options
//!
//! ## Recommended: All Participants Run the Arbiter
//!
//! Each participant should run its own instance of the arbiter over a replicated log (deterministic
//! order of events across all dealers) of commitments, acknowledgements, and reveals.
//! All correct participants, when given the same log, will arrive at the same result (will recover
//! the same group polynomial and a share that can generate partial signatures over it). Using a
//! replicated log allows us to provide both reliable broadcast (all honest dealers see all messages from
//! all other honest dealers) and to enforce a "timeout" (using log index).
//!
//! _For an example of this approach, refer to <https://docs.rs/commonware-reshare>._
//!
//! ## Alternative: Trusted Process
//!
//! It is possible to run the arbiter as a standalone process that dealers
//! must trust to track commitments, acknowledgements, and reveals and then notify
//! all parties which commitments and shares to use to generate the group public key and shares.
//!
//! _For an example of this approach, refer to <https://docs.rs/commonware-vrf>._
//!
//! # Disqualification on Attributable Faults
//!
//! Submitting duplicate and/or unnecessary information (i.e. a dealer submitting the same commitment twice
//! or submitting an acknowledgement for a disqualified dealer) will throw an error but not disqualify the
//! dealer. It may not be possible for dealers to know the latest state of the arbiter when submitting
//! information and penalizing them for this is not helpful.
//!
//! Submitting invalid information (invalid commitment) qualifies as an attributable fault that disqualifies a
[...]

PR #2347, opened 2025-11-28
[cryptography/bls12381] Ignore Unknown Disqualifications

patrick-ogrady:
Closes: #1996

### DIFF
--- a/cryptography/src/bls12381/dkg/mod.rs
+++ b/cryptography/src/bls12381/dkg/mod.rs
@@ -162,7 +162,7 @@ pub mod player;
 pub use player::Player;
 pub mod types;
 
-#[derive(thiserror::Error, Debug)]
+#[derive(thiserror::Error, Debug, PartialEq, Eq)]
 pub enum Error {
     #[error("unexpected polynomial")]
     UnexpectedPolynomial,
@@ -734,7 +734,7 @@ mod tests {
         let unknown = PrivateKey::from_seed(n as u64).public_key();
 
         // Disqualifying an unknown public key should be ignored
-        assert!(matches!(arb.disqualify(unknown), Err(Error::DealerInvalid)));
+        assert_eq!(arb.disqualify(unknown).unwrap_err(), Error::DealerInvalid);
 
         // Add commitments from all dealers
         for con in &contributors {


cryptography/src/bls12381/dkg/arbiter.rs
```diff
@@ -102,9 +102,19 @@ impl<P: PublicKey, V: Variant> Arbiter<P, V> {
...
-        self.disqualified.insert(participant);
+    /// Disqualify a dealer from the DKG for external reason (i.e. sending invalid messages).
+    ///
+    /// If the participant is not a dealer, this is a no-op.
+    ///
+    /// # Warning
+    ///
+    /// If the [Arbiter] is being run by all participants, all participants must disqualify the
+    /// same public keys (or else will derive different group polynomials).
+    pub fn disqualify(&mut self, dealer: P) {
```
patrick-ogrady: We could return a result here (with InvalidDealer)?

clabby: Would be nice to signal misuse, though I imagine users would also treat that error as a no-op. It would give them the chance to log, though.

patrick-ogrady: Another good reason for an error here is that you can fail to process whatever payload is being handled if this returns an error (like a block).

As a reminder, in any case, this is only an issue if you explicitly call `disqualify` (we already handle invalid dealers in other processing).

examples/reshare/src/dkg/manager.rs
```diff
@@ -501,6 +506,15 @@ where
             return;
         }
 
+        // Verify the dealer is part of this round.
+        //
+        // This rule does not prevent a dealer from submitting a dealing of another dealer. If this
+        // is desired, it could be enforced by checking the proposer of the block matches the dealer.
+        if self.dealers.index(&outcome.dealer).is_none() {
```
patrick-ogrady, line 513: Would test this in integration but we don't have a harness setup yet for byzantine in this example.

examples/reshare/src/application/core.rs
```diff
@@ -101,8 +101,14 @@ where
...
-        // verification (parent commitment and height contiguity), hence there is
-        // nothing to verify here.
+        // verification (parent commitment and height contiguity).
+        //
+        // You could opt to verify the deal_outcome in the block here (both that it is valid
+        // and that the dealer is the proposer) but we opt to only process deal data after the
+        // block has been finalized to keep verification as fast as possible. The downside
+        // of this approach is that invalid data can be included in the canonical chain (which
+        // makes certificates over finalized blocks less useful because the verifier must still
+        // check the block contents).
```
clabby, line 111: ❤️

cryptography/src/bls12381/dkg/mod.rs
```diff
@@ -715,6 +715,43 @@ mod tests {
...
+
+        // Create arbiter
+        let mut arb =
+            Arbiter::<_, MinSig>::new(None, contributors.clone(), contributors.clone(), 1);
+
+        // Create a unknown public key
+        let unknown = PrivateKey::from_seed(n as u64).public_key();
+
+        // Disqualifying an unknown public key should be ignored
+        assert!(matches!(arb.disqualify(unknown), Err(Error::DealerInvalid)));
```
clabby: Not necessary, but `unwrap_err` would be nice here.

clabby (dismissed):
Nice. I suspect this API will be used less by consumers following Lucas' changes, but definitely makes this footgun harder to hit.

cryptography/src/bls12381/dkg/mod.rs
patrick-ogrady: Will swap

clabby (approved):

examples/vrf/src/handlers/arbiter.rs
```diff
@@ -93,7 +93,7 @@ impl<E: Clock + Spawner, C: PublicKey> Arbiter<E, C> {
                         let msg = match wire::Dkg::decode_cfg(msg, &self.contributors.len()) {
                             Ok(msg) => msg,
                             Err(_) => {
-                                arbiter.disqualify(peer);
+                                arbiter.disqualify(peer).expect("failed to disqualify peer");
```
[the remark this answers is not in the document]
patrick-ogrady: This misconfiguration would be a bug in `VRF`.

PR #2305, opened 2025-11-22
[consensus/simplex] Avoid `Resolver` Boomerang

patrick-ogrady:
Fixes: #2264

PR #2348, opened 2025-11-30
feat: avoid Resolver Boomerang 

0xKarl98:
This pr aims to resolve the issue  #2264 , summerized below : 

-  Track view+certificate type for backfilled proofs and suppress sending the same proof back to the resolver  , while still notifying it for locally assembled certificates. 

-  Adds a test to ensure backfilled finalizations no longer bounce to the resolver , which have already be done in this PR https://github.com/commonwarexyz/monorepo/pull/2305 , i.e. assert!(resolver_receiver.next().now_or_never().is_none());

-  Pass the test in step 2 , which noted as not passing in main 
cc @patrick-ogrady


patrick-ogrady:
This was addressed in #2351

PR #2350, opened 2025-11-30
[workspace] Pedantic lint configuration

clabby:
## Overview

Configures a few useful lints for the workspace.

### DIFF
--- a/cryptography/src/ed25519/scheme.rs
+++ b/cryptography/src/ed25519/scheme.rs
@@ -4,7 +4,7 @@ cfg_if::cfg_if! {
         use crate::BatchVerifier;
         use std::borrow::{Cow, ToOwned};
     } else {
-        use alloc::borrow::ToOwned;
+        use alloc::borrow::{Cow, ToOwned};
     }
 }
 use bytes::{Buf, BufMut};


utils/src/bitmap/historical/bitmap.rs
```diff
@@ -436,13 +436,8 @@ impl<const N: usize> BitMap<N> {
             // Use or_insert_with so we don't overwrite chunks already captured
             // by capture_modified_chunks (which runs first and takes precedence).
             changes.entry(chunk_idx).or_insert_with(|| {
-                if let Some(old_chunk) = self.get_chunk(chunk_idx) {
-                    // Chunk existed before: store its old data
```
roberto-bayardo, line 440: I think it's worth preserving this comment (the change makes what's going on even less explicit).

clabby: Will re-add! Existing changes were auto-fixed; Need to comb over the diff this afternoon.

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -215,6 +215,7 @@ impl<E: Clock + Rng + CryptoRng + Metrics, S: Scheme, D: Digest> State<E, S, D>
 
         // Try to construct entry certificates for the previous view
         // Prefer the strongest proof available so lagging replicas can re-enter quickly.
+        #[allow(clippy::option_if_let_else)]
```
clabby, line 218: IMO, in this specific case, `if let else` is actually more readable.

cronokirby:
One concern about the lint for "can be a const fn" is that because we provide libraries that people can use, we have to be mindful that *removing const* is a breaking change. So, I'd be cautious of having this be a blocking lint. Perhaps it would be better for this to be some kind of advisory that you get on a PR? Something like "you have these functions which could potentially be const, consider making them so", and in PRs we take care of actually reasoning about whether we want that.

clabby:
> One concern about the lint for "can be a const fn" is that because we provide libraries that people can use, we have to be mindful that _removing const_ is a breaking change. So, I'd be cautious of having this be a blocking lint. Perhaps it would be better for this to be some kind of advisory that you get on a PR? Something like "you have these functions which could potentially be const, consider making them so", and in PRs we take care of actually reasoning about whether we want that.

Yeah - that's a valid concern. I think in general, we should be making functions `const` when we can. If we anticipate something being extended, I think we can just `#[allow(clippy::missing_const_for_fn)]`. Should be similar to an OOB check, but forces us to be explicit.

consensus/src/simplex/signing_scheme/bls12381_multisig.rs
```diff
@@ -649,7 +649,7 @@ mod tests {
             &certificate,
         ));
 
-        let mut corrupted = certificate.clone();
+        let mut corrupted = certificate;
```
patrick-ogrady, line 652: I think this needs to be clone for this test? (it isn't copy afaict)

consensus/src/simplex/signing_scheme/bls12381_threshold.rs
```diff
@@ -1032,7 +1032,7 @@ mod tests {
             &certificate,
         ));
 
-        let mut corrupted = certificate.clone();
+        let mut corrupted = certificate;
```
patrick-ogrady, line 1035: see previous comment about clone here?

consensus/src/simplex/signing_scheme/ed25519.rs
```diff
@@ -638,7 +638,7 @@ mod tests {
             &certificate,
         ));
 
-        let mut corrupted = certificate.clone();
+        let mut corrupted = certificate;
```
patrick-ogrady, line 636: see previous comment about clone?

cryptography/src/bls12381/primitives/ops.rs
```diff
@@ -1333,7 +1333,7 @@ mod tests {
         // Failure with signatures from different public_keys
         let signer2 = &shares[1];
         let partial2 = partial_sign_message::<V>(signer2, messages[0].0, messages[0].1);
-        let mut partials_mixed_public_keys = partials.clone();
+        let mut partials_mixed_public_keys = partials;
```
patrick-ogrady, line 1340: need .clone()?

p2p/src/authenticated/lookup/actors/tracker/directory.rs
```diff
@@ -68,7 +68,7 @@ impl<E: Spawner + Rng + Clock + GClock + RuntimeMetrics, C: PublicKey> Directory
         let rate_limiter = RateLimiter::hashmap_with_clock(cfg.rate_limit, context.clone());
 
         // TODO(#1833): Metrics should use the post-start context
```
patrick-ogrady: nit: I think we can remove this comment

consensus/src/simplex/signing_scheme/bls12381_multisig.rs
clabby: It is not copy, but `certificate` can be moved here - `certificate` is never used again in the fn body.

consensus/src/simplex/signing_scheme/bls12381_threshold.rs
clabby: Same https://github.com/commonwarexyz/monorepo/pull/2350#discussion_r2583300958

consensus/src/simplex/signing_scheme/ed25519.rs
clabby: Same https://github.com/commonwarexyz/monorepo/pull/2350#discussion_r2583300958

cryptography/src/bls12381/primitives/ops.rs
clabby: `partials` is not used for the rest of the test, it can be moved.

consensus/src/simplex/signing_scheme/bls12381_multisig.rs
patrick-ogrady: My b...the GitHub diff viewer made it look like these were in the same test.

patrick-ogrady (approved):

Issue #2228, opened 2025-11-14
[consensus/simplex] Certificate-Only Voter


patrick-ogrady:
It would greatly simplify the code if the `voter` was only responsible for certificates (and `batcher` passed recovered certificates to the voter).

The "cleanest" way to do this would involve discarding votes that never enter a certificate (instead of retaining them in a journal) but that might be controversial. Given we only persist what makes it through the batcher anyways, this isn't actually that big of a change...

PR #2351, opened 2025-12-01
[consensus/simplex] Remove Votes from Voter

patrick-ogrady:
Related: #2228 
Fixes: #2264 
Future Work: #2381 

Outside of a significant improvement in code quality, this PR also reduce the number of buffered signature verifications per view by > ~33% (2 Notarize/1 Finalize -> 0-1 Notarize/0-1 Finalize).

### DIFF
--- a/consensus/src/simplex/actors/voter/actor.rs
+++ b/consensus/src/simplex/actors/voter/actor.rs
@@ -898,11 +898,11 @@ impl<
 
             // Attempt to send any new view messages
             //
-            // It is possible that the batcher may drop any votes we construct during this call
-            // if it has not yet been updated to the message's view (and it sees the vote as not interesting).
-            // This is acceptable because we update the batcher after this call if we enter a new view,
-            // and any votes for old views that are dropped have no impact on liveness (we may, however, miss
-            // building a finalization for an old view where we otherwise could have contributed).
+            // The batcher may drop votes we construct here if it has not yet been updated to the
+            // message's view. This only happens when we skip ahead multiple views, which always
+            // coincides with entering a new view (triggering a batcher update below before we send
+            // any votes for the new current view). This has no impact on liveness, however, we may miss
+            // building a finalization for an old view where we otherwise could have contributed.
             self.notify(
                 &mut batcher,
                 &mut resolver,


consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -556,18 +534,18 @@ impl<
         mut self,
         batcher: batcher::Mailbox<S, D>,
         resolver: resolver::Mailbox<S, D>,
+        batcher_output_receiver: mpsc::Receiver<batcher::BatcherOutput<S, D>>,
```
patrick-ogrady: use our existing mailbox?

consensus/src/simplex/actors/REFACTOR_PLAN.md
patrick-ogrady, line 1: remove this

consensus/src/simplex/metrics.rs
```diff
@@ -68,38 +68,17 @@ impl Inbound {
         }
     }
 
-    pub fn notarization(peer: &impl Array) -> Self {
```
patrick-ogrady, line 71: re-add this?

consensus/src/simplex/actors/batcher/ingress.rs
```diff
@@ -9,17 +12,37 @@ use futures::{
...
         finalized: View,
 
         active: oneshot::Sender<bool>,
     },
+    /// Our constructed vote (needed for quorum).
     Constructed(Voter<S, D>),
 }
 
+/// Messages sent from batcher to voter.
+pub enum BatcherOutput<S: Scheme, D: Digest> {
```
patrick-ogrady: Just use the voter mailbox.

consensus/src/simplex/actors/batcher/ingress.rs
```diff
@@ -9,14 +9,19 @@ use futures::{
 };
 use tracing::error;
 
+/// Messages sent from voter to batcher.
```
patrick-ogrady: cleanup these comments

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -28,6 +29,17 @@ use rand::{CryptoRng, Rng};
 use std::{collections::BTreeMap, sync::Arc};
 use tracing::{debug, trace, warn};
 
+/// Result of adding a vote to a round.
+enum AddResult<D: Digest> {
```
patrick-ogrady: nit: call this `action`

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -117,15 +190,28 @@ impl<
...
+                        };
+
                         self.reporter
                             .report(Activity::Notarize(notarize.clone()))
                             .await;
                         self.votes.insert_notarize(notarize.clone());
                         self.verifier.add(Voter::Notarize(notarize), false);
-                        true
+
+                        match proposal {
```
patrick-ogrady: nit: just return this in the proposal if/else?

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -207,58 +303,89 @@ impl<
...
+                        AddResult::Skipped
                     }
                     None => {
                         self.reporter
                             .report(Activity::Finalize(finalize.clone()))
                             .await;
                         self.votes.insert_finalize(finalize.clone());
                         self.verifier.add(Voter::Finalize(finalize), false);
-                        true
+                        AddResult::Added
```
patrick-ogrady: It is possible this could also be the leader proposal?

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -410,24 +612,36 @@ impl<
 
     pub fn start(
         mut self,
-        consensus: voter::Mailbox<S, D>,
-        receiver: impl Receiver<PublicKey = P>,
+        voter_mailbox: voter::Mailbox<S, D>,
+        pending_receiver: impl Receiver<PublicKey = P>,
+        recovered_receiver: impl Receiver<PublicKey = P>,
```
patrick-ogrady: nit: unify into a single channel (for votes/certs)?

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -543,43 +761,189 @@ impl<
...
+                        finalized,
+                        current,
+                        view,
+                        true, // allow future
+                    ) {
+                        continue;
+                    }
+
+                    // Get or create the round
+                    let round = work
```
patrick-ogrady: Wait to create round until we verify

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -594,25 +958,52 @@ impl<
...
+            }
+
+            // Try to construct and forward certificates
+            if let Some(notarization) = round.try_construct_notarization(&self.scheme) {
+                debug!(%view, "constructed notarization, forwarding to voter");
+                voter_mailbox
+                    .verified(Voter::Notarization(notarization))
+                    .await;
+            }
+
```
patrick-ogrady: nit: remove `\n`

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -28,6 +29,17 @@ use rand::{CryptoRng, Rng};
...
+enum AddResult<D: Digest> {
+    /// Vote was not added (duplicate, invalid, or certificate already exists).
+    Skipped,
+    /// Vote was added and needs verification.
+    Added,
+    /// Vote was added and this is the leader's first notarize vote (forward proposal immediately).
+    LeaderProposal(Proposal<D>),
+}
+
+/// Per-view state for vote accumulation and certificate tracking.
```
patrick-ogrady: add comment about how we don't care/think about equivocation here (where leader proposal differs from cert -> this is handled by voter)

consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -292,13 +282,10 @@ impl<
         }
     }
 
-    /// Records a locally verified nullify vote and ensures the round exists.
+    /// Persists our nullify vote to the journal for crash recovery.
     async fn handle_nullify(&mut self, nullify: Nullify<S>) {
```
patrick-ogrady, line 318: nit: remove single-line function

consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -322,14 +309,10 @@ impl<
         self.state.emit_floor(view)
     }
 
-    /// Persistently records a notarize vote we verified ourselves.
+    /// Persists our notarize vote to the journal for crash recovery.
     async fn handle_notarize(&mut self, notarize: Notarize<S, D>) {
```
patrick-ogrady, line 345: nit: remove single-line function

consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -343,14 +326,10 @@ impl<
         self.block_equivocator(equivocator).await;
     }
 
-    /// Records a finalize vote emitted by the verifier pipeline.
+    /// Persists our finalize vote to the journal for crash recovery.
     async fn handle_finalize(&mut self, finalize: Finalize<S, D>) {
-        self.append_journal(finalize.view(), Voter::Finalize(finalize.clone()))
+        self.append_journal(finalize.view(), Voter::Finalize(finalize))
```
patrick-ogrady: nit: remove single-line function

consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -558,18 +537,11 @@ impl<
         resolver: resolver::Mailbox<S, D>,
         pending_sender: impl Sender<PublicKey = P>,
         recovered_sender: impl Sender<PublicKey = P>,
```
patrick-ogrady: nit: `certificate_sender` (if keep separate)

consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -839,122 +806,48 @@ impl<
...
                             }
                         }
-                        Voter::Notarize(_) | Voter::Nullify(_) | Voter::Finalize(_) => {
-                            warn!(?sender, "blocking peer for invalid message type");
-                            self.blocker.block(sender).await;
-                            continue;
-                        }
-                    };
-                    match action {
-                        Action::Process => {}
```
patrick-ogrady, line 946: Is `Action` still used?

consensus/src/simplex/actors/voter/ingress.rs
```diff
@@ -1,21 +1,38 @@
...
 use commonware_cryptography::Digest;
 use futures::{channel::mpsc, SinkExt};
 use tracing::error;
 
+/// Messages sent to the voter from resolver and batcher.
+pub enum Message<S: Scheme, D: Digest> {
+    /// Certificate from resolver or batcher.
+    Voter(Voter<S, D>),
+    /// Leader's proposal from batcher.
+    Proposal(Proposal<D>),
```
patrick-ogrady, line 12: nit: put `Proposal` first

consensus/src/simplex/actors/voter/round.rs
```diff
@@ -138,23 +126,8 @@ impl<S: Scheme, D: Digest> Round<S, D> {
...
-    ///
-    /// When multiple conflicting proposals slip into the same round we cannot
-    /// safely keep any partial certificates that were built on top of the old
-    /// payload. Clearing the vote trackers forces the round to rebuild honest
-    /// quorums from scratch, while bubbling the equivocator's public key up so
-    /// higher layers can quarantine or slash them.
-    pub fn record_equivocation_and_clear(&mut self) -> Option<S::PublicKey> {
-        self.clear_votes();
+    /// Returns the leader's public key when equivocation is detected.
+    fn record_equivocation(&self) -> Option<S::PublicKey> {
```
patrick-ogrady: nit: remove this one-line function

consensus/src/simplex/actors/voter/round.rs
```diff
@@ -426,83 +339,40 @@ impl<S: Scheme, D: Digest> Round<S, D> {
         (true, equivocator)
     }
 
-    /// Constructs a notarization certificate if we have enough votes.
+    /// Returns a notarization certificate for broadcast if we have one and haven't broadcast it yet.
     pub fn notarizable(&mut self) -> Option<Notarization<S, D>> {
-        // Ensure we haven't already broadcast a notarization certificate.
```
patrick-ogrady, line 435: nit: retain comments

consensus/src/simplex/actors/voter/round.rs
```diff
@@ -635,22 +504,10 @@ mod tests {
         assert!(accepted);
         assert!(equivocator.is_some());
         assert_eq!(equivocator.unwrap(), participants[2]);
-
-        // Conflict clears votes
-        assert_eq!(round.votes.len_notarizes(), 0);
-
-        // Skip new attempts
-        assert!(round.construct_notarize().is_none());
```
patrick-ogrady, line 647: retain tests?

consensus/src/simplex/actors/voter/ingress.rs
```diff
@@ -1,21 +1,53 @@
...
 impl<S: Scheme, D: Digest> Mailbox<S, D> {
-    pub fn new(sender: mpsc::Sender<Voter<S, D>>) -> Self {
+    pub fn new(sender: mpsc::Sender<Message<S, D>>) -> Self {
         Self { sender }
     }
 
+    /// Send a verified certificate from the batcher.
+    ///
+    /// TODO: change name of this function
     pub async fn verified(&mut self, voter: Voter<S, D>) {
```
patrick-ogrady: nit: change name of function

consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -702,6 +695,7 @@ impl<
         batcher
             .update(observed_view, leader, self.state.last_finalized())
             .await;
+        // TODO: add comment why we don't send any constructed messages here (we expect the view we start in to have been longer over or timeout immediately)
```
patrick-ogrady: nit: add comment

consensus/src/simplex/actors/voter/mod.rs
```diff
@@ -1333,16 +936,33 @@ mod tests {
                 }
                 context.sleep(Duration::from_millis(10)).await;
             }
+
+            // Verify that voter broadcasts a finalize vote for proposal B (not A)
+            //
+            // TODO: how is this happening?
```
patrick-ogrady: TODO: how does this pass?

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -410,28 +144,38 @@ impl<
...
+        mut voter: voter::Mailbox<S, D>,
+        vote_receiver: impl Receiver<PublicKey = P>,
+        certificate_receiver: impl Receiver<PublicKey = P>,
     ) {
-        // Wrap channel
-        let mut receiver: WrappedReceiver<_, Voter<S, D>> =
-            WrappedReceiver::new(self.scheme.certificate_codec_config(), receiver);
+        // Wrap channels
+        //
+        // vote_receiver: receives votes from network
```
patrick-ogrady: nit: remove this

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -543,43 +431,53 @@ impl<
...
-                }
+                },
             }
 
             // Look for a ready verifier (prioritizing the current view)
             let mut timer = self.verify_latency.timer();
-            let mut selected = None;
-            if let Some(verifier) = work.get_mut(&current) {
-                if verifier.ready_notarizes() {
+            #[allow(clippy::type_complexity)]
```
patrick-ogrady: nit: remove this

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -410,28 +144,38 @@ impl<
...
+        // certificate_receiver: receives certificates from network
+        let mut vote_receiver: WrappedReceiver<_, Voter<S, D>> =
+            WrappedReceiver::new(self.scheme.certificate_codec_config(), vote_receiver);
+        let mut certificate_receiver: WrappedReceiver<_, Voter<S, D>> =
+            WrappedReceiver::new(self.scheme.certificate_codec_config(), certificate_receiver);
 
         // Initialize view data structures
         let mut current = View::zero();
         let mut finalized = View::zero();
+        let mut leader = 0u32;
```
patrick-ogrady: nit: remove this?

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -450,11 +190,18 @@ impl<
...
                             current = new_current;
                             finalized = new_finalized;
-                            work.entry(current)
+                            let proposal = work
+                                .entry(current)
                                 .or_insert_with(|| self.new_round(initialized))
                                 .set_leader(leader);
                             initialized = true;
 
+                            // If we already had the leader's vote, forward the proposal
```
patrick-ogrady: We need to handle this case explicitly now because the verifier doesn't just forward a vote for us.

patrick-ogrady: Actually, we should wait to do this until after verification.

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -410,28 +144,34 @@ impl<
...
+        mut voter: voter::Mailbox<S, D>,
+        vote_receiver: impl Receiver<PublicKey = P>,
+        certificate_receiver: impl Receiver<PublicKey = P>,
     ) {
-        // Wrap channel
-        let mut receiver: WrappedReceiver<_, Voter<S, D>> =
-            WrappedReceiver::new(self.scheme.certificate_codec_config(), receiver);
+        // Wrap channels
+        let mut vote_receiver: WrappedReceiver<_, Voter<S, D>> =
+            WrappedReceiver::new(self.scheme.certificate_codec_config(), vote_receiver);
```
patrick-ogrady: This is a bit confusing because we don't actually accept certificates over this channel...

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -594,35 +477,63 @@ impl<
...
-            // Drop any rounds that are no longer interesting
-            loop {
-                let Some((view, _)) = work.first_key_value() else {
-                    break;
-                };
-                let view = *view;
-                if interesting(self.activity_timeout, finalized, current, view, false) {
-                    break;
+            // Get the round for this view to construct certificates
+            let Some(round) = work.get_mut(&view) else {
```
patrick-ogrady: can force get

consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -839,122 +844,56 @@ impl<
...
+                                Certificate::Finalization(finalization) => {
+                                    trace!(%view, from_resolver, "received finalization");
+                                    self.handle_finalization(finalization).await;
+                                    if from_resolver {
+                                        resolved = Resolved::Finalization;
+                                    }
+                                }
                             }
                         }
-                        Voter::Notarize(_) | Voter::Nullify(_) | Voter::Finalize(_) => {
```
patrick-ogrady, line 939: No longer required ❤️

consensus/src/simplex/actors/batcher/verifier.rs
patrick-ogrady, line 1: This is copy and pasted from `types`

consensus/src/simplex/actors/voter/round.rs
```diff
@@ -426,83 +333,40 @@ impl<S: Scheme, D: Digest> Round<S, D> {
         (true, equivocator)
     }
 
-    /// Constructs a notarization certificate if we have enough votes.
-    pub fn notarizable(&mut self) -> Option<Notarization<S, D>> {
-        // Ensure we haven't already broadcast a notarization certificate.
+    /// Returns a notarization certificate for broadcast if we have one and haven't broadcast it yet.
+    pub fn broadcast_notarization(&mut self) -> Option<Notarization<S, D>> {
```
patrick-ogrady, line 341: cleaned up these names

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -1305,15 +1129,8 @@ mod tests {
...
-            let nullify = Nullify::sign::<Sha256Digest>(
-                &schemes[1],
-                &namespace,
-                Rnd::new(Epoch::new(1), view),
-            )
-            .unwrap();
-            state.add_verified_nullify(nullify);
 
-            // Attempt to notarize
+            // Attempt to notarize after timeout - should fail (nullify was broadcast)
```
patrick-ogrady: Is this test correct?

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -1023,31 +858,25 @@ mod tests {
...
+            let notarization_votes: Vec<_> = schemes
+                .iter()
+                .map(|scheme| Notarize::sign(scheme, &namespace, parent_proposal.clone()).unwrap())
+                .collect();
+            let notarization =
+                Notarization::from_notarizes(&verifier, notarization_votes.iter()).unwrap();
+            state.add_notarization(notarization);
 
-            // Get parent
+            // Get parent - should succeed now
```
patrick-ogrady: nit: remove `-`

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -500,18 +241,155 @@ impl<
...
+                        view,
+                        true, // allow future
+                    ) {
+                        continue;
+                    }
+
+                    match message {
+                        Certificate::Notarization(notarization) => {
+                            // Update metrics
+                            self.inbound_messages
```
patrick-ogrady: Should we increment this sooner?

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -500,18 +241,155 @@ impl<
...
+                                .or_insert_with(|| self.new_round(initialized))
+                                .set_finalization(finalization.clone());
+                            voter
+                                .recovered(Certificate::Finalization(finalization))
+                                .await;
+                        }
+                    }
+                },
+                // Handle votes from the network
+                message = vote_receiver.recv() => {
```
patrick-ogrady, line 357: We need to increment metrics here.

consensus/src/simplex/actors/batcher/round.rs
```diff
@@ -0,0 +1,415 @@
...
+
+    /// Returns true if we already have a finalization certificate for this view.
+    pub fn has_finalization(&self) -> bool {
+        self.finalization.is_some()
+    }
+
+    /// Stores a notarization certificate.
+    pub fn set_notarization(&mut self, notarization: Notarization<S, D>) {
+        self.notarization = Some(notarization);
+    }
```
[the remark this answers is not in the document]
patrick-ogrady: This is a misunderstanding of `batcher`'s role: it is **only** used at tip to help create new certificates (not at all for "catchup").

We could opt to set the proposal here but would then need to consider vote<>certificate equivocation in the batcher (which is nice to contain just to the `voter`).

consensus/src/simplex/actors/batcher/round.rs
```diff
@@ -0,0 +1,412 @@
...
+                        true
+                    }
+                }
+            }
+        }
+    }
+
+    /// Adds a vote that we constructed ourselves to the verifier.
+    ///
+    /// Returns true if the vote was added (may be needed for quorum).
```
patrick-ogrady: nit: fix this return (we always add)

patrick-ogrady: I found it simpler to only avoid updating verified votes/reconstruction later (rather than trying to prevent addition).

consensus/src/simplex/actors/batcher/round.rs
```diff
@@ -0,0 +1,412 @@
...
+                    .await;
+
+                // Our own votes are already verified
+                self.pending_votes.insert_finalize(finalize.clone());
+                self.verified_votes.insert_finalize(finalize.clone());
+            }
+        }
+        self.verifier.add(message, true);
+        true
+    }
```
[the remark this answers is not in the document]
patrick-ogrady: The comment here should be updated.

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -594,35 +486,65 @@ impl<
...
-                let Some((view, _)) = work.first_key_value() else {
-                    break;
-                };
-                let view = *view;
-                if interesting(self.activity_timeout, finalized, current, view, false) {
-                    break;
+            // Store verified votes for certificate construction
+            let round = work.get_mut(&view).expect("round must exist");
+            for valid in voters {
+                if let Some(proposal) = round.add_verified(valid) {
```
patrick-ogrady: Question: we set proposal on verifier when receiving the first proposer message (before we've verified it)? Is that a problem?

Given we only accept 1, that should be ok but interesting (we could then just send proposal during message processing rather than waiting for it here)?

patrick-ogrady: We could then remove the "force notarize" path and just do one verification!

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -594,35 +486,65 @@ impl<
...
-            loop {
-                let Some((view, _)) = work.first_key_value() else {
-                    break;
-                };
-                let view = *view;
-                if interesting(self.activity_timeout, finalized, current, view, false) {
-                    break;
+            // Store verified votes for certificate construction
+            let round = work.get_mut(&view).expect("round must exist");
+            for valid in voters {
```
patrick-ogrady, line 473: TODO: add a test where the batcher sends a proposal after we get a certificate (may already be tested -> should show "replaced")

consensus/src/simplex/actors/batcher/verifier.rs
```diff
@@ -0,0 +1,1295 @@
...
+                if let Some(ref leader_proposal) = self.leader_proposal {
+                    // If leader proposal is set and the message is not for it, drop it
+                    if leader_proposal != &notarize.proposal {
+                        return;
+                    }
+                } else if let Some(leader) = self.leader {
+                    // If leader is set but leader proposal is not, set it
+                    if leader == notarize.signer() {
+                        // Set the leader proposal
+                        self.set_leader_proposal(notarize.proposal.clone());
```
patrick-ogrady, line 125: We could opt to recover this proposal from certificates (rather than just `notarize`) but IMHO we should not try to work too hard for a lazy leader.

The original motivation for this "only notarize" is that the proposer's notarize is treated as their "PROPOSE" message (setting proposal on any other message would be equivalent to locking the proposal on "PREPARE" or "PRECOMMIT")

consensus/src/simplex/actors/batcher/verifier.rs
```diff
@@ -0,0 +1,1295 @@
...
+    }
+
+    /// Clears any pending messages that are not for the leader's proposal and forces
+    /// the notarizes to be verified.
+    ///
+    /// We force verification because we need to know the leader's proposal
+    /// to begin verifying it.
+    fn set_leader_proposal(&mut self, proposal: Proposal<D>) {
+        // Drop all notarizes/finalizes that aren't for the leader proposal
+        self.notarizes.retain(|n| n.proposal == proposal);
```
patrick-ogrady, line 83: We could have attempted to track all proposals (and then verify as soon as 1 has quorum) rather than treating the leader's notarize as "PROPOSE".

consensus/src/simplex/actors/voter/actor.rs
BrendanChou: Not resolved (unless you meant to keep this). Same for the other `handle_{vote}` functions

consensus/src/simplex/actors/voter/mod.rs
```diff
@@ -1465,6 +1460,32 @@ mod tests {
...
+            context.sleep(Duration::from_millis(100)).await;
+
+            // Add a notarization certificate for conflicting proposal
+            let (_, conflicting_notarization) =
+                build_notarization(&schemes, &namespace, &conflicting_proposal, quorum);
+            mailbox
+                .recovered(Certificate::Notarization(conflicting_notarization.clone()))
+                .await;
+
+            // Wait for a finalize vote to be broadcast (we drop our own conflicting proposal rather than marking as replaced)
```
patrick-ogrady, line 1475: This is a little odd given we would otherwise refrain from voting (if we detected this on any other participant) 🤔

patrick-ogrady: (This is no change to existing logic...just making sure it is codified)

consensus/src/simplex/actors/batcher/round.rs
```diff
@@ -0,0 +1,418 @@
...
+                // Our own votes are already verified
+                assert!(
+                    self.pending_votes.insert_finalize(finalize.clone()),
+                    "duplicate finalize"
+                );
+                self.verified_votes.insert_finalize(finalize.clone());
+            }
+        }
+        self.verifier.add(message, true);
+    }
```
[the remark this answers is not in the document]
patrick-ogrady: Wow! Nice catch!

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -450,40 +187,36 @@ impl<
...
+                                            break;
+                                        }
+
+                                        // If the leader is active, we can stop
+                                        if round.is_active(leader) {
+                                            active = true;
+                                            break;
+                                        };
+                                    }
+                                    active
```
BrendanChou: I feel like theres probably a more canonical way to do this whole loop, like:

```
let active = work.range(min_view..).iter().rev().any(|r| r.is_active(leader))
```

And maybe you don't even need the `rev` since I don't think it saves on expected iterations either way

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -500,18 +233,158 @@ impl<
...
+                            self.inbound_messages
+                                .get_or_create(&Inbound::nullification(&sender))
+                                .inc();
+                        }
+                        Certificate::Finalization(_) => {
+                            self.inbound_messages
+                                .get_or_create(&Inbound::finalization(&sender))
+                                .inc();
+                        }
+                    }
```
BrendanChou:
```suggestion
                    // Update metrics
                    let label = match &message {
                        Certificate::Notarization(_) => Inbound::notarization(&sender)),
                        Certificate::Nullification(_) => Inbound::nullification(&sender)),
                        Certificate::Finalization(_) => Inbound::finalization(&sender)),
                    };
                    self.inbound_messages.get_or_create(label).inc();
```

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -500,18 +233,158 @@ impl<
...
+
+                    // If the epoch is not the current epoch, block
+                    if message.epoch() != self.epoch {
+                        warn!(?sender, "blocking peer for epoch mismatch");
+                        self.blocker.block(sender).await;
+                        continue;
+                    }
+
+
```
BrendanChou: nit: two blank lines

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -524,6 +397,25 @@ impl<
...
+                            self.inbound_messages
+                                .get_or_create(&Inbound::nullify(&sender))
+                                .inc();
+                        }
+                        Vote::Finalize(_) => {
+                            self.inbound_messages
+                                .get_or_create(&Inbound::finalize(&sender))
+                                .inc();
+                        }
+                    }
```
BrendanChou: Same suggested style cleanup here as above

consensus/src/simplex/actors/batcher/verifier.rs
```diff
@@ -0,0 +1,1214 @@
...
+        }
+    }
+
+    /// Sets the leader for the current consensus view.
+    ///
+    /// If a notarize vote from the leader has already been received, this will
+    /// also set the leader's proposal, filtering out any pending votes for other
+    /// proposals.
+    pub fn set_leader(&mut self, leader: u32) {
+        assert!(self.leader.is_none());
```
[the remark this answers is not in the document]
patrick-ogrady: We should never call `update` multiple times in a single view.

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -594,35 +494,63 @@ impl<
...
+            if let Some(nullification) = round.try_construct_nullification(&self.scheme) {
+                recover_timer.observe();
+                debug!(%view, "constructed nullification, forwarding to voter");
+                voter
+                    .recovered(Certificate::Nullification(nullification))
+                    .await;
+            } else {
+                recover_timer.cancel();
+            }
+            let mut recover_timer = self.recover_latency.timer();
```
BrendanChou: nit: this `recover_timer` 3x (that also requires an `else` statement attached) almost seems a bit of a code-smell. Wondering if it would be possible to use some sort of generic function even to do this...

```
fn time_option<T>(metric, predicate: func() => Option<T>) {
    let timer = metric.timer();
    let result = predicate();
    match result {
        None => timer.cancel();
        Some(_) => timer.observe();
    }
    result
}
```

Another way is to add a method on the `Timer` struct `observe_if(bool)` which will either observe or cancel.
`recover_time.observe_if(result.is_some());`

patrick-ogrady: hmmm interesting idea

consensus/src/simplex/actors/voter/actor.rs
patrick-ogrady: Opted to leave as is.

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -543,47 +395,56 @@ impl<
...
+                        .await {
+                            self.added.inc();
+                        }
+                },
+            }
+
+            // Forward leader's proposal to voter (if we're not the leader and haven't already)
+            if let Some(round) = work.get_mut(&current) {
+                if let Some(me) = self.scheme.me() {
+                    if let Some(proposal) = round.get_proposal_to_forward(me) {
```
patrick-ogrady: nit: `.forward_proposal()`

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -543,47 +395,56 @@ impl<
...
+                    .map(|(view, round)| (*view, round));
+                if let Some((view, round)) = potential {
                     let (voters, failed) =
-                        verifier.verify_finalizes(&mut self.context, &self.namespace);
+                        round.verify_finalizes(&mut self.context, &self.namespace);
                     selected = Some((view, voters, failed));
                 }
             }
             let Some((view, voters, failed)) = selected else {
+                timer.cancel();
```
patrick-ogrady, line 446: If not `cancel`, we `observe` (dropping would lead to erroneous latency measurements)

consensus/src/simplex/actors/batcher/verifier.rs
```diff
@@ -0,0 +1,1214 @@
...
+};
+use commonware_cryptography::Digest;
+use rand::{CryptoRng, Rng};
+
+/// `Verifier` is a utility for tracking and batch verifying consensus messages.
+///
+/// In consensus, verifying multiple signatures at the same time can be much more efficient
+/// than verifying them one by one. This struct collects messages from participants in consensus
+/// and signals they are ready to be verified when certain conditions are met (e.g., enough messages
+/// to potentially reach a quorum, or when a leader's message is received).
```
patrick-ogrady: nit: remove "or when ..."

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -192,31 +192,16 @@ impl<
...
-                                    }
-                                    active
-                                };
+                                // Not enough views have passed to judge activity
+                                current < View::new(self.skip_timeout.get())
+                                // We don't have enough history to judge activity
+                                || (work.len() as u64) < self.skip_timeout.get()
+                                // Leader was active in at least one recent round
+                                || work
+                                    .range(current.saturating_sub(self.skip_timeout)..)
```
BrendanChou: nit: feels like we should use `checked_sub` first given that we do `current < self.skip_timeout` above, meaning that `saturating_sub` actually should never saturate. (It can still return 0 in a non-saturating way if `current == self.skip_timeout`)

Eh, tried to come up with something and it's not more straightforward than the way you're doing it
```
let is_active = 
  (work.len() as u64) < self.skip_timeout.get()
  || current.saturating_sub(self.skip_timeout).map_or(
    true,
    |floor| work.range(floor..).any(|(_, round)| round.is_active(leader)
  );
```

consensus/src/simplex/actors/voter/mod.rs
```diff
@@ -878,14 +779,14 @@ mod tests {
...
-    fn replay_duplicate_votes<S, F>(mut fixture: F)
+    fn certificate_conflicts_proposal<S, F>(mut fixture: F)
     where
         S: Scheme<PublicKey = ed25519::PublicKey>,
         F: FnMut(&mut deterministic::Context, u32) -> Fixture<S>,
     {
         let n = 5;
         let quorum = quorum(n);
-        let namespace = b"finalization_without_notarization".to_vec();
+        let namespace = b"certificate_override_test".to_vec();
```
patrick-ogrady: nit: certificate_conflicts_proposal

consensus/src/simplex/actors/voter/mod.rs
```diff
@@ -1923,4 +1853,170 @@ mod tests {
         finalization_from_resolver(bls12381_multisig::<MinSig, _>);
         finalization_from_resolver(ed25519);
     }
+
+    /// Test that certificates received from the resolver are not sent back to it.
+    ///
+    /// This is a regression test for the "boomerang" bug where:
```
patrick-ogrady: nit: remove

consensus/src/simplex/actors/voter/mod.rs
patrick-ogrady: I'm leaning towards just leaving this be for now ...

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -192,31 +192,16 @@ impl<
...
-                                            active = true;
-                                            break;
-                                        };
-                                    }
-                                    active
-                                };
+                                // Not enough views have passed to judge activity
+                                current < View::new(self.skip_timeout.get())
+                                // We don't have enough history to judge activity
+                                || (work.len() as u64) < self.skip_timeout.get()
```
BrendanChou: The logic here actually seems a bit inconsistent...

You're checking that you have at least `skip_timeout` total views in `work` but you might not actually iterate over that many views (lets say you have recent gaps in the range above `current - skip_timeout`

patrick-ogrady: This was an issue in the original code as well. Working on a fix.

patrick-ogrady: Pushed.

consensus/src/simplex/actors/batcher/actor.rs
patrick-ogrady: Just going to leave for now.

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -543,47 +398,56 @@ impl<
...
+                            self.added.inc();
+                        }
+                },
+            }
+
+            // Forward leader's proposal to voter (if we're not the leader and haven't already)
+            if let Some(round) = work.get_mut(&current) {
+                if let Some(me) = self.scheme.me() {
+                    if let Some(proposal) = round.forward_proposal(me) {
+                        voter.proposal(proposal).await;
```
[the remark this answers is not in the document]
patrick-ogrady: This is intentional (and that comment is out-of-date).

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -196,12 +196,15 @@ impl<
...
-                                    .range(current.saturating_sub(self.skip_timeout)..)
-                                    .any(|(_, round)| round.is_active(leader));
+                                // Check leader activity in recent window
+                                || {
+                                    // Not enough recent history to judge (assume active)
+                                    let min_view = current.saturating_sub(self.skip_timeout);
+                                    let recent: Vec<_> = work.range(min_view..).collect();
+                                    (recent.len() as u64) < self.skip_timeout.get()
+                                    // Leader was active in at least one recent round
+                                    || recent.iter().any(|(_, round)| round.is_active(leader))
```
BrendanChou: Probably fine if `skip_timeout` is small, which I guess it should be, but I think ideally you just iterate once without collecting, and just increment a counter as you go, making sure that the counter at the end is equal-to-or-above skip-timeout. However, since this is such a strict requirement, this is also just equal to checking that the key is exactly equal to the previous key plus one (i.e. that there are no gaps). This strikes me as too strict.

BrendanChou (dismissed):

consensus/src/simplex/actors/batcher/actor.rs
patrick-ogrady: Too delineate this conversation, what is in main is in main doesn't make a ton of sense/is wrong (was meant to check what this function now does -> there are skip_timeout views and not active in any).

As for whether or not this is too strict, that's a fair point. Does a view simply existing somehow now make it safe for us to consider skipping (may only have 1 vote)?

andresilva: `round.is_active` should probably also check for any finalize vote, it shouldn't make a difference in pratice, since finalize implies notarize, but network could do weird things? Also I wonder whether `skip_timeout` should imply activity in the last N **consecutive** views. Something like this:

```rust
let mut inactive = 0;
let mut is_active = true;
let min_view = current.saturating_sub(self.skip_timeout);
let mut expected = current.next();
for (&view, round) in work.range(min_view..=current).rev() {
    // enforce contiguity
    if view.next() != expected {
        break;
    }
    expected = view;

    // empty view doesn't count against the leader
    if !round.any_activity() {
        break;
    }
    // leader showed activity
    if round.is_active(leader) {
        break;
    }

    inactive += 1;
    if inactive >= self.skip_timeout.get() {
        is_active = false;
        break;
    }
}
```

```rust
// round.rs
pub fn any_activity(&self) -> bool {
    self.pending_votes.len_notarizes() > 0
        || self.pending_votes.len_nullifies() > 0
        || self.pending_votes.len_finalizes() > 0
}

pub fn is_active(&self, leader: u32) -> bool {
    self.pending_votes.has_notarize(leader)
        || self.pending_votes.has_nullify(leader)
        || self.pending_votes.has_finalize(leader)
}
```

andresilva (dismissed):

consensus/src/simplex/actors/batcher/actor.rs
patrick-ogrady: By the round "existing", I assume:
```
pub fn any_activity(&self) -> bool {
    self.pending_votes.len_notarizes() > 0
        || self.pending_votes.len_nullifies() > 0
        || self.pending_votes.len_finalizes() > 0
}
```

(That isn't strictly true I suppose but feel it is close enough)

> round.is_active should probably also check for any finalize vote, it shouldn't make a difference in pratice, since finalize implies notarize, but network could do weird things? 

Added in my branch.

Still pondering how to handle gaps in skip timeout 🤔

patrick-ogrady: I actually think what I had may be "not too bad" (ensures we don't activate this until we've seen a handful of views)

patrick-ogrady: Resolved: #2375

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -316,7 +42,7 @@ pub struct Actor<
     reporter: R,
 
     activity_timeout: ViewDelta,
-    skip_timeout: ViewDelta,
+    skip_timeout: usize,
```
andresilva: I guess this is done to avoid the `.get()` later on, but instead we introduce a `usize::from()`, not sure that's worth it.

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -450,40 +188,19 @@ impl<
...
-                                }
-                            }
+                            // Check if the leader has been active recently
+                            let is_active =
+                                // Ensure we have enough data to judge activity (none of this
+                                // data may be in the last skip_timeout views if we jumped ahead
+                                // to a new view)
+                                work.len() < self.skip_timeout
+                                // Leader active in at least one recent round
+                                || work.iter().rev().take(self.skip_timeout).any(|(_, round)| round.is_active(leader));
```
andresilva: This is probably good enough. Not sure whether enforcing contiguous views is worth it. This seems to strike a good enough balance at getting the heuristic to work with all the data we have available locally (for example, we can use future rounds information to judge whether the leader is active, unlike my proposal which cut off at current).

consensus/src/simplex/actors/batcher/actor.rs
andresilva: Or is this because semantically the timeout isn't strictly a number of views (because there might be gaps)?

patrick-ogrady: I made this change because we were casting to usize during each `Update` message (and only using it there). Figured it was good to be pedantic and try to cast once (panicking if it exceeded usize rather than truncating).

patrick-ogrady: Reverted

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -543,47 +394,56 @@ impl<
...
+                            self.added.inc();
+                        }
+                },
+            }
+
+            // Forward leader's proposal to voter (if we're not the leader and haven't already)
+            if let Some(round) = work.get_mut(&current) {
+                if let Some(me) = self.scheme.me() {
+                    if let Some(proposal) = round.forward_proposal(me) {
+                        voter.proposal(proposal).await;
```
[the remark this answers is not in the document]
patrick-ogrady: This is ok (as resolved previously)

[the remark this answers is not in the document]
patrick-ogrady: I don't actually think this is a bug (although its so nuanced it makes me want to add a comment for sure)...

Basically, if we receive a cert (in a view we aren't yet in) we may construct a vote (let's say a notarize after seeing notarization) and send to the batcher. We'll then send a batcher update to increase the view by 1 (above cert).

If the view of the cert didn't exist, we may handle the message (if its a big jump over our last view we may ignore the message because its not interesting) and then the update will add the leader to view + 1 (the new current). We'll never set the leader for the old view (fine/expected) but we won't send a message for view + 1 until the next iteration of the loop.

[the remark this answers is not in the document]
patrick-ogrady: Added a clarification that this is ok.

andresilva (dismissed):

consensus/src/simplex/actors/batcher/actor.rs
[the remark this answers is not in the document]
patrick-ogrady: As noted multiple times previously, this is intended.

PR #2353, opened 2025-12-01
[cryptography] DKG Refactor

cronokirby:
This creates a new `dkg2` module, implementing a refactored interface, which should be much easier to audit and use. The crux of the new interface is handling all cryptographic operations, including signing, inside of the module. We also get rid of the "auditor". Instead, the log passed to finalize players can also be passed to an "observe" function, to compute the public output. This log must be agreed upon by all parties.

### DIFF
--- a/utils/src/set.rs
+++ b/utils/src/set.rs
@@ -274,7 +274,7 @@ impl<K, V> OrderedAssociated<K, V> {
         self.position(key).and_then(|index| self.values.get(index))
     }
 
-    /// Returns a mutable referene to the associated value for `key`, if it exists.
+    /// Returns a mutable reference to the associated value for `key`, if it exists.
     pub fn get_value_mut(&mut self, key: &K) -> Option<&mut V>
     where
         K: Ord,


cryptography/src/bls12381/dkg2.rs
```diff
@@ -0,0 +1,2352 @@
...
+//!     round_info,
+//!     dealer_logs,
+//!     1 // Increase this for parallelism.
+//! )?;
+//! println!("DKG completed with threshold {}", observer_output.quorum());
+//! # Ok(())
+//! # }
+//! ```
+//!
+//! For a complete production example with resharing, see [commonware-reshare](https://docs.rs/commonware-reshare).
```
patrick-ogrady: nit: remove `production` (would recommend storing key material more securely in prod)

cryptography/src/bls12381/dkg2.rs
```diff
@@ -0,0 +1,2352 @@
...
+//! # Usage
+//!
+//! ## Core Types
+//!
+//! * [`Info`]: Configuration for a DKG/Reshare round, containing the dealers, players, and optional previous output
+//! * [`Output`]: The public result of a successful DKG round, containing the public polynomial and player list
+//! * [`Share`]: A player's private share of the distributed key (from `primitives::group`)
+//! * [`Dealer`]: State machine for a dealer participating in the protocol
+//! * [`Player`]: State machine for a player receiving shares
+//! * [`SignedDealerLog`]: A dealer's signed transcript of their interactions with players
```
patrick-ogrady, line 35: While the name of this type is a bit more verbose than I'd typically prefer, I think the specificity of including <Dealer/Player><Obj> makes it less error-prone 👍

cryptography/src/bls12381/dkg2.rs
```diff
@@ -0,0 +1,2352 @@
...
+//! - A [`DealerPubMsg`] containing the polynomial commitment to broadcast
+//! - A vector of `(player_id, DealerPrivMsg)` pairs to send privately
+//!
+//! The [`DealerPubMsg`] contains a public polynomial commitment of degree `2f` where `f = max_faults(n)`.
+//! Each [`DealerPrivMsg`] contains a scalar evaluation of the dealer's private polynomial at the player's index.
+//!
+//! ### Step 3: Player Verification
+//!
+//! Each player creates a [`Player`] instance via [`Player::new`], then for each dealer message:
+//! - Call [`Player::dealer_message`] with the [`DealerPubMsg`] and [`DealerPrivMsg`]
```
patrick-ogrady, line 67: nit: should we add a `trusted_dealer_message` that skips verification (for replay on restart)?

cryptography/src/bls12381/dkg2.rs
```diff
@@ -0,0 +1,2352 @@
...
+//!
+//! With collected [`SignedDealerLog`]s:
+//! - Call [`SignedDealerLog::check`] to verify and extract [`DealerLog`]s
+//! - Players call [`Player::finalize`] with all logs to compute their [`Share`] and [`Output`]
+//! - Observers call [`observe`] with all logs to compute just the [`Output`]
+//!
+//! The [`Output`] contains:
+//! - The final public polynomial (sum of dealer polynomials for DKG, interpolation for reshare)
+//! - The list of players who received shares
+//! - A digest for verification
```
patrick-ogrady: nit: can you elaborate on this (i.e. two ceremonies run identically will have the same digest but if any difference, it will differ)?

cryptography/src/bls12381/dkg2.rs
```diff
@@ -0,0 +1,2352 @@
...
+//!
+//! ## Utility Functions
+//!
+//! * [`deal`]: Generate shares non-interactively for testing (returns [`Output`] and shares)
```
patrick-ogrady: I think we could improve the docs here?

cryptography/src/bls12381/dkg2.rs
```diff
@@ -0,0 +1,2352 @@
...
+//!
+//! This crate does not provide an integrated mechanism for tracking complaints from players (of malicious dealers). However, it is
+//! possible to implement your own mechanism and to manually disqualify dealers from a given round in the arbiter. This decision was made
+//! because the mechanism for communicating commitments/shares/acknowledgements is highly dependent on the context in which this
+//! construction is used.
+//!
+//! In practice:
+//! - [`Player::dealer_message`] returns `None` for invalid messages (implicit complaint)
+//! - [`Dealer::receive_player_ack`] validates acknowledgements
+//! - External arbiters can exclude misbehaving dealers before calling [`observe`] or [`Player::finalize`]
```
patrick-ogrady: If we are reading over a log, all "should" see the same invalid messages (and exclude)? I suppose the rational to not bake this in was just to provide flexibility to the caller?

cryptography/src/bls12381/dkg2.rs
```diff
@@ -0,0 +1,2352 @@
...
+    }
+
+    /// Return the quorum, i.e. the number of players needed to reconstruct the key.
+    pub fn quorum(&self) -> u32 {
+        self.players.quorum()
+    }
+
+    /// Get the public polynomial associated with this output.
+    ///
+    /// This is useful to verify partial signatures, with [crate::bls12381::primitives::ops::partial_verify_message].
```
patrick-ogrady: nit: "for verifying"

cryptography/src/bls12381/dkg2.rs
```diff
@@ -0,0 +1,2352 @@
...
+///
+/// This is used to bind signatures to the current round, and to provide the
+/// information that dealers, players, and observers need to perform their actions.
+#[derive(Debug, Clone)]
+pub struct Info<V: Variant, P: PublicKey> {
+    round: u64,
+    previous: Option<Output<V, P>>,
+    dealers: Ordered<P>,
+    players: Ordered<P>,
+    /// Never written when encoded, always computed from the previous fields.
```
patrick-ogrady: nit: prefer a `\n` above this line

cryptography/src/bls12381/dkg2.rs
```diff
@@ -0,0 +1,2352 @@
...
+    }
+}
+
+impl<V: Variant, P: PublicKey> Info<V, P> {
+    /// Figure out what the dealer share should be.
+    ///
+    /// If there's no previous round, we need a random value, hence `rng`.
+    ///
+    /// However, if there is a previous round, we expect a share, hence `Result`.
+    fn generate_dealer_share_if_necessary(
```
patrick-ogrady: nit: just `generate_dealer_share`

cryptography/src/bls12381/dkg2.rs
```diff
@@ -0,0 +1,2352 @@
...
+    /// `players` should be the list of public keys for the players.
+    pub fn new(
+        round: u64,
+        previous: Option<Output<V, P>>,
+        dealers: Ordered<P>,
+        players: Ordered<P>,
+    ) -> Result<Self, Error> {
+        let participant_range = 1..u32::MAX as usize;
+        if !participant_range.contains(&dealers.len()) {
+            return Err(Error::InsufficientDealers(dealers.len()));
```
patrick-ogrady: nit: this can also error because too many dealers? Maybe should just be `Error::InvalidDealers`?

cryptography/src/bls12381/dkg2.rs
```diff
@@ -0,0 +1,2352 @@
...
+                .iter()
+                .find(|d| previous.players.position(d).is_none())
+            {
+                return Err(Error::UnknownDealer(format!("{unknown:?}")));
+            }
+            if dealers.len() < previous.quorum() as usize {
+                return Err(Error::InsufficientDealers(dealers.len()));
+            }
+        }
+        let summary = Transcript::new(NAMESPACE)
```
patrick-ogrady, line 575: ❤️

cryptography/src/bls12381/dkg2.rs
```diff
@@ -0,0 +1,2352 @@
...
+use rayon::{
+    iter::{IntoParallelIterator, ParallelIterator as _},
+    ThreadPoolBuilder,
+};
+use std::collections::BTreeMap;
+use thiserror::Error;
+
+const NAMESPACE: &[u8] = b"commonware-bls12381-dkg";
+const NAMESPACE_DEALER: &[u8] = b"dealer";
+const NAMESPACE_SIGNED_LOG: &[u8] = b"signed-log";
```
patrick-ogrady: nit: `_DEALER_SIGNED_LOG`?

cryptography/src/bls12381/dkg2.rs
```diff
@@ -0,0 +1,2352 @@
...
+use core::num::NonZeroU32;
+use rand_core::CryptoRngCore;
+use rayon::{
+    iter::{IntoParallelIterator, ParallelIterator as _},
+    ThreadPoolBuilder,
+};
+use std::collections::BTreeMap;
+use thiserror::Error;
+
+const NAMESPACE: &[u8] = b"commonware-bls12381-dkg";
```
patrick-ogrady: our typical format for `NAMESPACE` is `_COMMONWARE_...`

cryptography/src/bls12381/dkg2.rs
```diff
@@ -0,0 +1,2352 @@
...
+
+impl<V: Variant, P: PublicKey> Info<V, P> {
+    /// Create a new [`Info`].
+    ///
+    /// `round` should be a counter, always incrementing, even for failed DKGs.
+    /// `previous` should be the result of the previous successful DKG.
+    /// `dealers` should be the list of public keys for the dealers. This MUST
+    /// be a subset of the previous round's players.
+    /// `players` should be the list of public keys for the players.
+    pub fn new(
```
patrick-ogrady, line 550: We should allow the caller to provide their own `namespace` (like we do in consensus) and then concatenate with a `NAMESPACE` const.

This prevents DKGs from different networks from interfering (shouldn't happen on reshare unless same shares but could happen on initial DKG)

cryptography/src/bls12381/dkg2.rs
```diff
@@ -0,0 +1,2352 @@
...
+
+impl Read for DealerPrivMsg {
+    type Cfg = ();
+
+    fn read_cfg(
+        buf: &mut impl bytes::Buf,
+        cfg: &Self::Cfg,
+    ) -> Result<Self, commonware_codec::Error> {
+        Ok(Self {
+            share: Read::read_cfg(buf, cfg)?,
```
patrick-ogrady: nit: call with `read` (like we do for `P::Signature` below)

cryptography/src/bls12381/dkg2.rs
```diff
@@ -0,0 +1,2352 @@
...
+
+    fn read_cfg(
+        buf: &mut impl bytes::Buf,
+        &max_players: &Self::Cfg,
+    ) -> Result<Self, commonware_codec::Error> {
+        let tag = u8::read(buf)?;
+        match tag {
+            0 => Ok(Self::Ok(Read::read_cfg(
+                buf,
+                &(RangeCfg::from(0..=max_players.get() as usize), (), ()),
```
patrick-ogrady, line 826: does this need to be min 1?

cryptography/src/bls12381/dkg2.rs
```diff
@@ -0,0 +1,2352 @@
...
+        mut rng: impl CryptoRngCore,
+        round_info: Info<V, S::PublicKey>,
+        me: S,
+        share: Option<Share>,
+    ) -> Result<(Self, DealerPubMsg<V>, Vec<(S::PublicKey, DealerPrivMsg)>), Error> {
+        // Check that this dealer is defined in the round.
+        round_info.dealer_index(&me.public_key())?;
+        let share =
+            round_info.generate_dealer_share_if_necessary(&mut rng, share.map(|x| x.private))?;
+        let my_poly = new_with_constant(round_info.degree(), &mut rng, share.clone());
```
patrick-ogrady: nit: put rng first

cryptography/src/bls12381/dkg2.rs
```diff
@@ -0,0 +1,2352 @@
...
+    pub fn finalize(self) -> SignedDealerLog<V, S> {
+        let reveals = self
+            .results
+            .values()
+            .iter()
+            .filter(|x| x.is_reveal())
+            .count() as u32;
+        // Omit results if there are too many reveals.
+        let results = if reveals > self.round_info.max_reveals() {
+            LogResults::TooManyReveals
```
patrick-ogrady: nit: `DealerResult` rather than `LogResults`?

cryptography/src/bls12381/dkg2.rs
```diff
@@ -0,0 +1,2352 @@
...
+#[allow(clippy::type_complexity)]
+fn select<V: Variant, P: PublicKey>(
+    round_info: &Info<V, P>,
+    logs: BTreeMap<P, DealerLog<V, P>>,
+) -> Result<Vec<(P, DealerLog<V, P>)>, Error> {
+    let required_commitments = round_info.required_commitments() as usize;
+    let transcript = transcript_for_round(round_info);
+    let out = logs
+        .into_iter()
+        .filter_map(|(dealer, log)| {
```
patrick-ogrady, line 1129: For performance, I performed `checks` prior to `select` (makes the end-of-epoch operation longer).

Was this change intentional/coming at it from a different lens I'm not considering?

cryptography/src/bls12381/dkg2.rs
```diff
@@ -0,0 +1,2352 @@
...
+                        .collect::<BTreeSet<_>>();
+                    for &i_dealer in &round.dealers {
+                        if round.bad(previous_successful_round.is_some(), i_dealer) {
+                            assert!(!good_pks.contains(&keys[i_dealer as usize].public_key()));
+                        }
+                    }
+                }
+                // Run observer
+                let observe_result = observe(round_info.clone(), dealer_logs.clone(), 1);
+
```
patrick-ogrady: nit: prefer removing `\n`

cryptography/src/bls12381/dkg2.rs
```diff
@@ -0,0 +1,2352 @@
...
+                if round.expect_failure(previous_successful_round) {
+                    assert!(
+                        observe_result.is_err(),
+                        "Round {} should have failed but succeeded",
+                        i_round
+                    );
+                    continue;
+                }
+
```
patrick-ogrady: nit: prefer removing `\n`

cryptography/src/bls12381/dkg2.rs
```diff
@@ -0,0 +1,2352 @@
...
+#[cfg(test)]
+mod test {
+    use super::{test_plan::*, *};
+    use crate::{bls12381::primitives::variant::MinPk, ed25519, PrivateKeyExt};
+    use anyhow::anyhow;
+    use core::num::NonZeroI32;
+    use rand::SeedableRng;
+    use rand_chacha::ChaCha8Rng;
+
+    // ========================================================================
```
patrick-ogrady: nit: remove this comment block

cryptography/src/bls12381/dkg2.rs
```diff
@@ -0,0 +1,2352 @@
...
+                plan.validate()
+                    .map_err(|_| arbitrary::Error::IncorrectFormat)?;
+                Ok(plan)
+            }
+        }
+    }
+}
+
+#[cfg(feature = "fuzz")]
+pub use test_plan::Plan as FuzzPlan;
```
patrick-ogrady, line 2146: nit: can we just use `Plan`?

cryptography/fuzz/fuzz_targets/bls12381_dkg.rs
patrick-ogrady, line 1: Very cool that we can use our tests for fuzzing too 👍

cryptography/src/bls12381/dkg2.rs
cronokirby: I don't think the performance of re-verifying is onerous enough to warrant this, especially given the expected duration of a DKG.

Additionally, this introduces a way of "mis-using" the API by calling the wrong function, which I don't like. Also, re-verifying seems like a smart idea, in order to prevent bugs related to confusion around which messages to replay, etc.

cryptography/src/bls12381/dkg2.rs
cronokirby: I think `Plan` isn't clear enough that it's not linked to testing. I don't like the idea of having to expose all of the test_plan module either.

Open to names besides `FuzzPlan` though.

cryptography/src/bls12381/dkg2.rs
cronokirby: Added some commentary about this.

cryptography/src/bls12381/dkg2.rs
cronokirby: I don't like just `generate_dealer_share` because it sounds like it always generates a new share, when in fact this method only generates one if there's no existing share.

Some other suggestions:
- `unwrap_or_random_share`,
- `calculate_share`.

cryptography/src/bls12381/dkg2.rs
cronokirby: Strictly speaking, no, but the idea is that this gives us one value to pass around as the codec config for the module, which is "maximum number of players across all rounds", which, as a value, does have to be >= 1.

There are a few different structs where 0 would be fine, but then you need to call .get() as you decode certain structs, and this seems worse than just having one config value for everything, or ()

cryptography/src/bls12381/dkg2.rs
cronokirby: It makes the responsibilities of what you need to come to consensus on simpler. Everybody just needs to agree on which dealer posted what, i.e. a single `BTreeMap<Dealer, DealerLog>`, and then `observe` or `Player::finalize` can be called.

You could make both of those APIs work in a kind of "streaming" mode, but that would make it more complicated.

If we were to do that, we would have an `Observer` struct, which would have one method to add a single dealer, log pair, and then a finalize method to produce an output. `Player` would need a corresponding method, and its finalize method would no longer take those arguments.

If you do that, you also now need to worry about managing the state of this object, which might mean doing the checks more than once if there's crashes.

My two cents are that it would be better to keep the API simpler, and optimize things via parallelism, or doing batched algorithms for the checks. The worry with the DKG to me is less that it wouldn't be fast enough, and more so that you implement it wrong, so a simpler API seems better.

cryptography/src/bls12381/dkg2.rs
```diff
@@ -298,9 +304,9 @@ use rayon::{
 use std::collections::BTreeMap;
 use thiserror::Error;
 
-const NAMESPACE: &[u8] = b"commonware-bls12381-dkg";
+const NAMESPACE: &[u8] = b"_COMMONWARE_BLS12381-DKG";
```
patrick-ogrady: prefer: `_COMMONWARE_BLS12381_DKG`

cryptography/src/bls12381/dkg2.rs
```diff
@@ -298,9 +304,9 @@ use rayon::{
...
 const NAMESPACE_DEALER: &[u8] = b"dealer";
```
patrick-ogrady: in other places we do `union_unique(namespace, `_DEALER`):

cryptography/src/bls12381/dkg2.rs
```diff
@@ -298,9 +304,9 @@ use rayon::{
...
-const NAMESPACE_SIGNED_LOG: &[u8] = b"signed-log";
+const NAMESPACE_DEALER_SIGNED_LOG: &[u8] = b"dealer-signed-log";
```
patrick-ogrady, line 310: Should there be a namespace for `_ACK`?

cryptography/src/bls12381/dkg2.rs
```diff
@@ -298,9 +304,9 @@ use rayon::{
...
```
patrick-ogrady: `_DEALER_LOG`?

cryptography/src/bls12381/dkg2.rs
```diff
@@ -535,20 +542,23 @@ impl<V: Variant, P: PublicKey> Info<V, P> {
...
         dealers: Ordered<P>,
         players: Ordered<P>,
     ) -> Result<Self, Error> {
         let participant_range = 1..u32::MAX as usize;
         if !participant_range.contains(&dealers.len()) {
-            return Err(Error::InsufficientDealers(dealers.len()));
+            return Err(Error::NumDealers(dealers.len()));
         }
         if !participant_range.contains(&players.len()) {
             return Err(Error::InsufficientPlayers(players.len()));
```
patrick-ogrady: Same issue with `NumPlayers`

cryptography/src/bls12381/dkg2.rs
```diff
@@ -602,6 +614,7 @@ impl<V: Variant, P: PublicKey> Write for Info<V, P> {
         self.previous.write(buf);
         self.dealers.write(buf);
         self.players.write(buf);
+        self.summary.write(buf);
```
patrick-ogrady: In the comments above, it says `summary` is never written?

cryptography/src/bls12381/dkg2.rs
patrick-ogrady: It seems you are just using the base namespace for this in transcript to prevent signing conflicts (which seems fine).

cryptography/src/bls12381/dkg2.rs
patrick-ogrady: Unless the set of signers is particularly large, I agree it should be < ~1s of extra time. I'm fine leaving as-is for now and we can revisit if we want to speed things up.

cryptography/src/bls12381/dkg2.rs
patrick-ogrady: I prefer `unwrap_or_random_share` to what we have now (seems the most rusty)

cryptography/src/bls12381/dkg2.rs
patrick-ogrady: I'm ok with "returning to fundamentals" here but suspect we'll need to re-optimize this.

patrick-ogrady (commented):
~namespace nits for consistency:
* https://github.com/commonwarexyz/monorepo/blob/c448f2af8a0c41bc02adccd7727371aceb4fd7a7/examples/reshare/src/dkg/manager.rs#L41-L45
* https://github.com/commonwarexyz/monorepo/blob/c448f2af8a0c41bc02adccd7727371aceb4fd7a7/examples/reshare/src/dkg/manager.rs#L307

cryptography/src/bls12381/dkg2.rs
cronokirby: The purpose of these two namespaces are to distinguish between the things that we might be signing, versus an application namespace. Because of this, I think using the lower case convention would be better.

I think right now the names are a bit confusing though, and we should probably have these instead be named `SIG_ACK` and `SIG_LOG` instead or something.

cryptography/src/bls12381/dkg2.rs
cronokirby: Right, I had to change this to support adding namespaces.

Either we add the namespace to the struct, or we add the hash to the encoded bytes. It also seems fine to "trust" the hash you read out, because in the situations you serialize an Info, it should be you that writes it and reads it (e.g. from storage), you shouldn't trust an Info received from some other peer blindly.

cryptography/src/bls12381/dkg2.rs
```diff
@@ -0,0 +1,2359 @@
...
+use core::num::NonZeroU32;
+use rand_core::CryptoRngCore;
+use rayon::{
+    iter::{IntoParallelIterator, ParallelIterator as _},
+    ThreadPoolBuilder,
+};
+use std::collections::BTreeMap;
+use thiserror::Error;
+
+const NAMESPACE: &[u8] = b"_COMMONWARE_BLS12381-DKG";
```
patrick-ogrady: prefer: `_COMMONWARE_BLS12381_DKG`

cryptography/src/bls12381/dkg2.rs
```diff
@@ -0,0 +1,2359 @@
...
+    log: &DealerLog<V, P>,
+) -> Transcript {
+    let mut out = transcript_for_round(round_info).fork(SIG_LOG);
+    out.commit(log.encode());
+    out
+}
+
+pub struct Dealer<V: Variant, S: Signer> {
+    me: S,
+    round_info: Info<V, S::PublicKey>,
```
patrick-ogrady: nit: `info:`

cryptography/src/bls12381/dkg2.rs
```diff
@@ -0,0 +1,2359 @@
...
+    ///
+    /// The private message MUST be sent encrypted (or, in some other way, privately)
+    /// to the target player. Similarly, that player MUST be convinced that this dealer
+    /// sent it that message, without any possibility of impersonation. A simple way
+    /// to provide both guarantees is through an authenticated channel, e.g. via
+    /// [crate::handshake], or [commonware-p2p](https://docs.rs/commonware-p2p/latest/commonware_p2p/).
+    #[allow(clippy::type_complexity)]
+    pub fn start(
+        mut rng: impl CryptoRngCore,
+        round_info: Info<V, S::PublicKey>,
```
patrick-ogrady: nit: `info: `

cryptography/src/bls12381/dkg2.rs
```diff
@@ -0,0 +1,2359 @@
...
+            pub_msg: self.pub_msg,
+            results,
+        };
+        SignedDealerLog::sign(&self.me, &self.round_info, log)
+    }
+}
+
+#[allow(clippy::type_complexity)]
+fn select<V: Variant, P: PublicKey>(
+    round_info: &Info<V, P>,
```
patrick-ogrady: nit: `info: `

cryptography/src/bls12381/dkg2.rs
```diff
@@ -0,0 +1,2359 @@
...
+}
+
+struct ObserveInner<V: Variant, P: PublicKey> {
+    output: Output<V, P>,
+    weights: Option<BTreeMap<u32, Weight>>,
+}
+
+impl<V: Variant, P: PublicKey> ObserveInner<V, P> {
+    fn reckon(
+        round_info: Info<V, P>,
```
patrick-ogrady: nit: `info: `

cryptography/src/bls12381/dkg2.rs
```diff
@@ -0,0 +1,2359 @@
...
+/// Observe the result of a DKG, using the public results.
+///
+/// The log mapping dealers to their log is the shared piece of information
+/// that the participants (players, observers) of the DKG must all agree on.
+///
+/// From this log, we can (potentially, as the DKG can fail) compute the public output.
+///
+/// This will only ever return [`Error::DkgFailed`].
+pub fn observe<V: Variant, P: PublicKey>(
+    round_info: Info<V, P>,
```
patrick-ogrady: nit: `info: `

cryptography/src/bls12381/dkg2.rs
```diff
@@ -0,0 +1,2359 @@
...
+
+/// Represents a player in the DKG / reshare process.
+///
+/// The player is attempting to get a share of the key.
+///
+/// They need not have participated in prior rounds.
+pub struct Player<V: Variant, S: Signer> {
+    me: S,
+    me_pub: S::PublicKey,
+    round_info: Info<V, S::PublicKey>,
```
patrick-ogrady: nit: `info`

cryptography/src/bls12381/dkg2.rs
patrick-ogrady: dumb question: when would we ever serialize `Info`? It seems like whenever we sign over it, we just use `summary` anyways?

I looked at the reshare follow-up and didn't see where either (maybe I'm glossing right over it).

cryptography/src/bls12381/dkg2.rs
```diff
@@ -0,0 +1,2359 @@
...
+    pub_msg: &DealerPubMsg<V>,
+) -> Transcript {
+    let mut out = transcript.fork(SIG_ACK);
+    out.commit(dealer.encode());
+    out.commit(pub_msg.encode());
+    out
+}
+
+fn transcript_for_log<V: Variant, P: PublicKey>(
+    round_info: &Info<V, P>,
```
patrick-ogrady: nit: `info`

cryptography/src/bls12381/dkg2.rs
```diff
@@ -0,0 +1,2359 @@
...
+    ) -> Result<Self, commonware_codec::Error> {
+        Ok(Self {
+            dealer: ReadExt::read(buf)?,
+            log: Read::read_cfg(buf, cfg)?,
+            sig: ReadExt::read(buf)?,
+        })
+    }
+}
+
+fn transcript_for_round<V: Variant, P: PublicKey>(round_info: &Info<V, P>) -> Transcript {
```
patrick-ogrady: nit: `info`

cryptography/src/bls12381/dkg2.rs
```diff
@@ -0,0 +1,2359 @@
...
+    /// Check this log for a particular round.
+    ///
+    /// This will produce the public key of the dealer that signed this log,
+    /// and the underlying log that they signed.
+    ///
+    /// This will return [`Option::None`] if the check fails.
+    #[allow(clippy::type_complexity)]
+    pub fn check(
+        self,
+        round_info: &Info<V, S::PublicKey>,
```
patrick-ogrady: nit: `info`

Issue #2049, opened 2025-10-29
[p2p/utils] Remove `wrap`


patrick-ogrady:
Instead of accepting a `Codec` message in `p2p`, we accept `Bytes` and provide a wrapper utility (called `wrap`) to turn this into a `Codec` stream: https://github.com/commonwarexyz/monorepo/blob/main/p2p/src/utils/codec.rs

Given that `Bytes` is already `Codec` (https://github.com/commonwarexyz/monorepo/blob/main/codec/src/types/bytes.rs), we should update `p2p` to just accept codec (still supports generic byte-sending but doesn't require a wrapper on the happy path).

PR #1796, opened 2025-10-03
[examples] Reshare over replicated log

clabby:
## Overview

Adds an example for resharing over a replicated log, rather than using a trusted arbiter (a. la. [`commonware-vrf`](https://docs.rs/commonware-vrf)).

makes progress on #1708 - to also reconfigure after performing a reshare, we'll need to have the epocher example in.

PR #2354, opened 2025-12-01
[cryptography] Propagate DKG Refactor

cronokirby:
Builds on #2353, and should be merged after.

This updates the codebase to use the new DKG interface.

The reshare example is probably the biggest change, where a lot of the code ends up looking different now that the DKG module itself does more of the work.

### DIFF
--- a/examples/reshare/src/validator.rs
+++ b/examples/reshare/src/validator.rs
@@ -366,12 +366,14 @@ mod test {
         }
     }
 
+    #[derive(Clone)]
     struct Crash {
         frequency: Duration,
         downtime: Duration,
         count: usize,
     }
 
+    #[derive(Clone)]
     struct Plan {
         seed: u64,
         total: u32,
@@ -382,7 +384,7 @@ mod test {
     }
 
     impl Plan {
-        async fn run_inner<S>(self, mut ctx: deterministic::Context) -> anyhow::Result<()>
+        async fn run_inner<S>(self, mut ctx: deterministic::Context) -> anyhow::Result<String>
         where
             S: Scheme<PublicKey = PublicKey>,
             SchemeProvider<S, PrivateKey>:
@@ -494,7 +496,7 @@ mod test {
 
                         if status.values().filter(|x| **x >= epoch).count() >= self.total as usize {
                             if successes >= self.target {
-                                return Ok(());
+                                return Ok(ctx.auditor().state());
                             } else {
                                 current_epoch = current_epoch.next();
                             }
@@ -542,13 +544,45 @@ mod test {
             }
         }
 
-        fn run<S>(self) -> anyhow::Result<()>
+        fn run<S>(self) -> anyhow::Result<String>
         where
             S: Scheme<PublicKey = PublicKey>,
             SchemeProvider<S, PrivateKey>:
                 EpochSchemeProvider<Variant = MinSig, PublicKey = PublicKey, Scheme = S>,
         {
             Runner::seeded(self.seed).start(|ctx| self.run_inner(ctx))
+        }
+    }
+
+    #[test_group("slow")]
+    #[test_traced("INFO")]
+    fn plan_deterministic() {
+        let plan = Plan {
+            seed: 0,
+            total: 4,
+            per_round: vec![4],
+            link: Link {
+                latency: Duration::from_millis(0),
+                jitter: Duration::from_millis(0),
+                success_rate: 1.0,
+            },
+            target: 1,
+            crash: None,
+        };
+        for seed in 0..3 {
+            let res0 = Plan {
+                seed,
+                ..plan.clone()
+            }
+            .run::<EdScheme>()
+            .unwrap_or_else(|e| panic!("failure: {e}"));
+            let res1 = Plan {
+                seed,
+                ..plan.clone()
+            }
+            .run::<EdScheme>()
+            .unwrap_or_else(|e| panic!("failure: {e}"));
+            assert_eq!(res0, res1);
         }
     }
 


Cargo.toml
```diff
@@ -20,7 +20,6 @@ members = [
     "examples/flood",
     "examples/log",
     "examples/sync",
-    "examples/vrf",
```
patrick-ogrady, line 23: IIRC there is a reference on the main README.md and in some blog posts?

consensus/src/aggregation/mocks/reporter.rs
```diff
@@ -63,7 +60,7 @@ impl<V: Variant, D: Digest> Reporter<V, D> {
     ) -> (Self, Mailbox<V, D>) {
         let (sender, receiver) = mpsc::channel(1024);
         let identity = *poly::public::<V>(&polynomial);
```
patrick-ogrady, line 62: Should we similarly define `public()` on `polynomial` like we did for `evaluate_all`?

consensus/src/ordered_broadcast/ack_manager.rs
```diff
@@ -254,20 +256,31 @@ mod tests {
     /// Different payloads for the same chunk produce distinct thresholds.
     fn chunk_different_payloads<V: Variant>() {
         let num_validators = 6;
-        let quorum = 3;
```
patrick-ogrady, line 257: lol...just totally broken

consensus/src/ordered_broadcast/ack_manager.rs
```diff
@@ -603,17 +616,12 @@ mod tests {
...
 
-        let submissions = [
-            (0, &chunk1),
-            (1, &chunk2),
-            (2, &chunk1),
-            (3, &chunk2),
-            (4, &chunk1),
-            (5, &chunk2),
-        ];
+        let submissions = (0..2 * quorum)
```
patrick-ogrady, line 619: :fancy:

cryptography/src/bls12381/benches/dkg_recovery.rs
```diff
@@ -1,64 +1,100 @@
...
-                    let contributors = (0..n)
-                        .map(|i| ed25519::PrivateKey::from_seed(i as u64).public_key())
-                        .collect::<Ordered<_>>();
+        for &concurrency in CONCURRENCY {
+            c.bench_function(
+                &format!("{}/n={} t={} conc={}", module_path!(), n, t, concurrency),
+                |b| {
+                    b.iter_batched(
+                        || {
+                            let private_keys = (0..n)
```
patrick-ogrady: As noted previously, the entire block in `iter_batched` is shared between all values of `CONCURRENCY`. We could save a lot of setup time (at least we did before) by computing outside the concurrency loop to reuse.

cryptography/src/bls12381/benches/dkg_recovery.rs
```diff
@@ -1,64 +1,100 @@
...
 
-/// Concurrency isn't used in DKG recovery, so we set it to 1.
-const CONCURRENCY: usize = 1;
-
 // Configure contributors based on context
 #[cfg(not(full_bench))]
 const CONTRIBUTORS: &[u32] = &[5, 10, 20, 50];
 #[cfg(full_bench)]
 const CONTRIBUTORS: &[u32] = &[5, 10, 20, 50, 100, 250, 500];
+const CONCURRENCY: &[usize] = &[1, 4, 8];
```
patrick-ogrady: nit: set `CONCURRENCY` to 1 for not(full_bench)

cryptography/src/bls12381/benches/dkg_reshare_recovery.rs
```diff
@@ -12,97 +13,89 @@ use std::{collections::BTreeMap, hint::black_box};
...
-            outputs.push(
-                player
-                    .finalize(commitments.clone(), BTreeMap::new())
-                    .unwrap(),
-            );
-        }
-
-        for concurrency in [1, 8] {
+        let t = quorum(n);
+        for &concurrency in CONCURRENCY {
```
patrick-ogrady: See note in `dkg_recovery` about reusing the iter_batched block.

cryptography/src/bls12381/benches/evaluate_point.rs
```diff
@@ -11,8 +14,7 @@ fn benchmark_evaluate_point(c: &mut Criterion) {
             b.iter_batched(
                 || {
                     let mut rng = StdRng::seed_from_u64(0);
-                    let (polynomial, _) =
```
patrick-ogrady, line 14: nice!

cryptography/src/bls12381/dkg.rs
```diff
@@ -727,7 +727,7 @@ impl<P: PublicKey> AckOrReveal<P> {
 impl<P: PublicKey> std::fmt::Debug for AckOrReveal<P> {
     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
         match self {
-            Self::Ack(x) => write!(f, "Ack({x:?})"),
+            Self::Ack(x) => write!(f, "Ack({:?})", x),
```
patrick-ogrady: put `x` in write

cryptography/src/bls12381/dkg.rs
```diff
@@ -1564,12 +1564,20 @@ mod test_plan {
             // Check dealer/player ranges
             for &d in &self.dealers {
                 if d >= num_participants {
-                    return Err(anyhow!("dealer {d} out of range [1, {num_participants}]"));
+                    return Err(anyhow!(
+                        "dealer {} out of range [1, {}]",
+                        d,
```
patrick-ogrady: put values inline

cryptography/src/bls12381/dkg.rs
```diff
@@ -1912,7 +1920,8 @@ mod test_plan {
                 if round.expect_failure(previous_successful_round) {
                     assert!(
                         observe_result.is_err(),
-                        "Round {i_round} should have failed but succeeded"
+                        "Round {} should have failed but succeeded",
```
patrick-ogrady: put values inline (warnings for me locally)

cryptography/src/bls12381/primitives/poly.rs
```diff
@@ -78,21 +76,11 @@ impl<C: Element> EncodeSize for Eval<C> {
...
-    // Reference: https://github.com/celo-org/celo-threshold-bls-rs/blob/a714310be76620e10e8797d6637df64011926430/crates/threshold-bls/src/poly.rs#L46-L52
-    new_from(degree, &mut OsRng)
-}
-
 // Returns a new scalar polynomial of the given degree where each coefficient is
 // sampled at random from the provided RNG.
 ///
 /// In the context of secret sharing, the threshold is the degree + 1.
-pub fn new_from<R: CryptoRngCore>(degree: u32, rng: &mut R) -> Poly<Scalar> {
+pub fn new_from<R: CryptoRngCore>(rng: &mut R, degree: u32) -> Poly<Scalar> {
```
patrick-ogrady, line 83: :pray:

examples/reshare/src/main.rs
```diff
@@ -1,22 +1,73 @@
...
 mod application;
 mod dkg;
 mod engine;
+mod namespace;
 mod orchestrator;
+mod self_channel;
 mod setup;
 mod validator;
 
+struct SaveFileOnUpdate {
```
patrick-ogrady, line 40: nit: can you add some comments explaining what this is for (and why we don't use something like Metadata)?

examples/reshare/src/main.rs
```diff
@@ -1,22 +1,73 @@
...
+impl UpdateCallBack<MinSig, PublicKey> for SaveFileOnUpdate {
+    fn on_update(
+        &mut self,
+        update: Update<MinSig, PublicKey>,
+    ) -> Pin<Box<dyn Future<Output = PostUpdate> + Send>> {
+        let config_path = self.path.clone();
+        Box::pin(async move {
+            match update {
+                Update::Failure { epoch } => {
+                    info!(epoch = %epoch, "dkg failed ; retrying");
```
patrick-ogrady: use a `warn!`?

examples/reshare/src/main.rs
```diff
@@ -1,22 +1,73 @@
...
+
+impl UpdateCallBack<MinSig, PublicKey> for SaveFileOnUpdate {
+    fn on_update(
+        &mut self,
+        update: Update<MinSig, PublicKey>,
+    ) -> Pin<Box<dyn Future<Output = PostUpdate> + Send>> {
+        let config_path = self.path.clone();
+        Box::pin(async move {
+            match update {
+                Update::Failure { epoch } => {
```
patrick-ogrady, line 58: nit: `dkg failed; retrying`
i

examples/reshare/src/main.rs
```diff
@@ -1,22 +1,73 @@
...
+                    info!(epoch = %epoch, "dkg failed ; retrying");
+                    PostUpdate::Continue
+                }
+                Update::Success {
+                    output,
+                    share,
+                    epoch,
+                } => {
+                    info!(epoch = %epoch, "dkg succeeded ; saving file");
```
patrick-ogrady: nit: `dkg succeeded; stopping`

examples/reshare/src/main.rs
```diff
@@ -115,9 +166,14 @@ fn main() {
...
             Subcommands::Setup(args) => setup::run(args),
-            Subcommands::Dkg(args) => validator::run::<EdScheme>(context, args).await,
+            Subcommands::Dkg(args) => {
+                let config_path = args.config_path.clone();
+                validator::run::<EdScheme>(context, args, SaveFileOnUpdate::boxed(config_path))
+                    .await;
+            }
             Subcommands::Validator(args) => {
-                validator::run::<ThresholdScheme<MinSig>>(context, args).await
+                validator::run::<ThresholdScheme<MinSig>>(context, args, Box::new(ContinueOnUpdate))
```
patrick-ogrady: nit: `ContinueOnUpdate::boxed()`? (for consistency)

examples/reshare/src/engine.rs
```diff
@@ -47,33 +51,27 @@ const MAX_REPAIR: NonZero<usize> = NZUsize!(50);
 
 pub struct Config<C, P, B, V>
 where
-    C: Signer,
     P: Manager<PublicKey = C::PublicKey, Peers = Ordered<C::PublicKey>>,
+    C: PrivateKey,
```
patrick-ogrady: nit: use `Signer`

examples/reshare/src/engine.rs
```diff
@@ -47,33 +51,27 @@ const MAX_REPAIR: NonZero<usize> = NZUsize!(50);
...
-
     pub partition_prefix: String,
     pub freezer_table_initial_size: u32,
 }
 
 pub struct Engine<E, C, P, B, H, V, S>
 where
     E: Spawner + Metrics + Rng + CryptoRng + Clock + GClock + Storage + Network,
-    C: Signer,
+    C: PrivateKey,
```
patrick-ogrady: nit: `Signer`

examples/reshare/src/engine.rs
```diff
@@ -113,7 +111,7 @@ where
 impl<E, C, P, B, H, V, S> Engine<E, C, P, B, H, V, S>
 where
     E: Spawner + Metrics + Rng + CryptoRng + Clock + GClock + Storage + Network,
-    C: Signer,
+    C: PrivateKey,
```
patrick-ogrady: nit: `Signer`

examples/reshare/src/engine.rs
```diff
@@ -318,6 +312,7 @@ where
             mpsc::Receiver<handler::Message<Block<H, C, V>>>,
             commonware_resolver::p2p::Mailbox<handler::Request<Block<H, C, V>>>,
         ),
+        update_cb: Box<dyn UpdateCallBack<V, C::PublicKey>>,
```
patrick-ogrady: does it make sense to alias `UpdateCallBack` as `Box<dyn UpdateCallBack<V, C::PublicKey>>`?

examples/reshare/src/engine.rs
```diff
@@ -318,6 +312,7 @@ where
             mpsc::Receiver<handler::Message<Block<H, C, V>>>,
             commonware_resolver::p2p::Mailbox<handler::Request<Block<H, C, V>>>,
         ),
```
patrick-ogrady, line 314: nit: `callback:` rather than `update_cb:`

examples/reshare/src/dkg/types.rs
```diff
@@ -1,238 +0,0 @@
-//! Types for the DKG/reshare protocol.
```
patrick-ogrady, line 1: Good riddance.

examples/reshare/src/orchestrator/actor.rs
```diff
@@ -38,7 +38,7 @@ pub struct Config<B, V, C, H, A, S>
 where
     B: Blocker<PublicKey = C::PublicKey>,
     V: Variant,
-    C: Signer,
+    C: PrivateKey,
```
patrick-ogrady: nit: `Signer`

examples/reshare/src/orchestrator/actor.rs
```diff
@@ -63,7 +63,7 @@ where
     E: Spawner + Metrics + Rng + CryptoRng + Clock + GClock + Storage + Network,
     B: Blocker<PublicKey = C::PublicKey>,
     V: Variant,
-    C: Signer,
+    C: PrivateKey,
```
patrick-ogrady: nit: `Signer`

examples/reshare/src/orchestrator/actor.rs
```diff
@@ -90,7 +90,7 @@ where
...
```
patrick-ogrady: nit: `Signer`

examples/reshare/src/self_channel.rs
```diff
@@ -0,0 +1,121 @@
...
+/// The possible errors that `self_channel` will produce.
+#[derive(Debug, Error)]
+pub enum Error {
+    #[error("network error: {0}")]
+    NetworkError(#[from] anyhow::Error),
+    #[error("the receiver has closed")]
+    ReceiverClosed,
+}
+
+/// A [Sender] allowing messages sent to yourself.
```
patrick-ogrady, line 20: I don't like this pattern (I think sending messages to yourself like this is error-prone/confusing and I excluded on purpose in `p2p`). If we want to allow this (would require a discussion), we should change `p2p` to support this rather than trying to workaround it here.

examples/reshare/src/setup.rs
```diff
@@ -68,20 +74,53 @@ impl ParticipantConfig {
...
+    pub fn max_participants_per_round(&self) -> u32 {
+        self.num_participants_per_round
+            .iter()
+            .copied()
+            .max()
+            .unwrap()
+    }
+
+    pub fn num_participants_in_round(&self, round: u64) -> u32 {
+        self.num_participants_per_round
```
patrick-ogrady, line 102: Woah, this is cool.

examples/reshare/src/setup.rs
```diff
@@ -68,20 +74,53 @@ impl ParticipantConfig {
...
+            [(round % self.num_participants_per_round.len() as u64) as usize]
+    }
+
+    /// Pick the dealers for a particular round.
+    ///
+    /// The first round will use the first `num_participants_per_epoch` players
+    /// as the dealers.
+    ///
+    /// Subsequent rounds will choose the same number, pseudo-randomly,
```
patrick-ogrady: nit: revise this comment...they can't chose the "same number" (because rounds may be of different sizes/depends on whether last round was successful?)

examples/reshare/src/setup.rs
```diff
@@ -133,42 +172,44 @@ fn generate_identities(
...
-) -> (Option<Public<MinSig>>, Vec<(PrivateKey, Option<Share>)>) {
+) -> (
+    Option<Output<MinSig, PublicKey>>,
+    Vec<(PrivateKey, Option<Share>)>,
+) {
+    // Generate p2p private keys
+    let peer_signers = (0..num_peers)
+        .map(|_| PrivateKey::from_rng(&mut OsRng))
+        .collect::<Vec<_>>();
     // Generate consensus key
```
patrick-ogrady, line 181: nit: add `\n` before

examples/reshare/src/setup.rs
```diff
@@ -274,7 +302,7 @@ mod serde_hex {
     }
 }
 
-mod serde_hex_vec {
+mod serde_hex_ordered {
```
patrick-ogrady, line 301: :heart:

examples/reshare/src/validator.rs
```diff
@@ -156,1928 +155,710 @@ where
...
-                break;
-            }
-            runs += 1;
+    #[test_group("slow")]
+    #[test_traced("INFO")]
+    fn four_epoch_threshold_scheme_changing_size() {
+        if let Err(e) = (Plan {
+            seed: 0,
+            total: 8,
+            per_round: vec![3, 4, 5],
```
patrick-ogrady, line 700: I suppose this implicitly tests `[5, 3, 4]` too?

examples/reshare/src/dkg/actor/player/state.rs
```diff
@@ -0,0 +1,132 @@
...
+    /// `round` identifies the round, with each state given a different round.
+    /// `max_read_size` is a parameter governing read sizes. This should correspond
+    /// with the number of players we might acknowledge.
+    pub async fn load(ctx: E, partition: String, round: u64, max_read_size: NonZeroU32) -> Self {
+        let mut storage = Metadata::<E, U64, PlayerData<V, P>>::init(
+            ctx,
+            metadata::Config {
+                partition,
+                codec_config: (
+                    RangeCfg::new(0..=max_read_size.get() as usize),
```
patrick-ogrady: For our disk, we could set max_read_size to be u32::MAX?

patrick-ogrady: And avoid the function arg?

examples/reshare/src/dkg/actor/player.rs
```diff
@@ -0,0 +1,226 @@
...
+
+    pub fn start(mut self) -> Handle<()> {
+        spawn_cell!(self.ctx, self.run().await)
+    }
+
+    async fn run(mut self) {
+        let mut stopped = self.ctx.stopped().fuse();
+        // Exiting this loop is stopping the actor.
+        let finalize = loop {
+            select_biased! {
```
patrick-ogrady: As noted previously, please use `select` (or the new `select_loop` here).

examples/reshare/src/dkg/actor/player.rs
```diff
@@ -0,0 +1,226 @@
...
+                // Context has stopped, so terminate the actor.
+                _ = stopped => break None,
+                // Some dealer sent us a message, or the network is dead.
+                msg = self.from_dealers.recv().fuse() => {
```
patrick-ogrady: You should be able to remove this `fuse()` call once you use our select

examples/reshare/src/dkg/actor/player.rs
```diff
@@ -0,0 +1,226 @@
...
+                }
+                res = self.inbox.next() => {
+                    let Some(msg) = res else {
+                        break None;
+                    };
+                    match msg {
+                        Message::Transmit => if self.transmit().await.is_err() {
+                            break None;
+                        },
+                        Message::Finalize { logs, cb_in } => break Some((logs,cb_in)),
```
patrick-ogrady: nit: please write out `cb_in`

examples/reshare/src/dkg/actor/player.rs
```diff
@@ -0,0 +1,226 @@
...
+                )
+                .await?;
+        }
+        Ok(())
+    }
+
+    fn finalize(
+        self,
+        logs: BTreeMap<C::PublicKey, DealerLog<V, C::PublicKey>>,
+        cb_in: oneshot::Sender<PlayerOutput<V, C::PublicKey>>,
```
patrick-ogrady: nit: write out `cb_in`

examples/reshare/src/dkg/actor/player.rs
```diff
@@ -0,0 +1,226 @@
...
+    ) {
+        let _ = cb_in.send(self.player.finalize(logs, 1));
```
patrick-ogrady: nit: leave a comment why we drop this

examples/reshare/src/dkg/actor/observer.rs
```diff
@@ -0,0 +1,93 @@
...
+
+impl<E, V, S> Observer<E, V, S>
+where
+    E: Clock + Storage + Metrics,
+    V: Variant,
+    S: PrivateKey,
+{
+    /// Load the storage from disk.
+    pub async fn load(ctx: E, partition: String, round: u64, max_read_size: NonZeroU32) -> Self {
+        let mut storage = Metadata::<E, U64, Data<V, S::PublicKey>>::init(
```
patrick-ogrady: If this requires keeping multiple instances of metadata open, we need to avoid that (it isn't safe).

I think you could just have an observer instance that accepts round arguments for the other functions?

examples/reshare/src/dkg/actor/dealer/state.rs
```diff
@@ -0,0 +1,160 @@
...
+    E: Clock + Storage + Metrics,
+    P: PublicKey,
+{
+    /// Load the state from storage.
+    ///
+    /// `round` identifies the round, with each state given a different round.
+    /// `max_read_size` is a parameter governing read sizes. This should correspond
+    /// with the number of players we might acknowledge.
+    pub async fn load(ctx: E, partition: String, round: u64, max_read_size: usize) -> Self {
+        let storage = Metadata::<E, U64, DealerData<P>>::init(
```
patrick-ogrady: See earlier comment about not creating multiple instances of metadata.

patrick-ogrady: Would be better to accept metadata as an arg (and/or wrap it with Arc<Mutex<>>)

examples/reshare/src/dkg/actor/dealer.rs
```diff
@@ -0,0 +1,208 @@
...
+            round_info.round(),
+            max_read_size.get() as usize,
+        )
+        .await;
+        let mut ctx = ContextCell::new(ctx);
+        let (outbox, inbox) = mpsc::channel(1);
+        let mailbox = Mailbox(outbox);
+
+        let (dealer, pub_msg, priv_msgs) = Dealer::start(
+            Transcript::resume(state.seed(ctx.as_mut()).await).noise(b"dealer rng"),
```
patrick-ogrady: nit: prefer label with no spaces

examples/reshare/src/dkg/actor.rs
```diff
@@ -1,114 +1,145 @@
...
+    pub peer_config: PeerConfig<C::PublicKey>,
 }
 
 pub struct Actor<E, P, H, C, V>
 where
     E: Spawner + Metrics + CryptoRngCore + Clock + GClock + Storage,
     P: Manager<PublicKey = C::PublicKey, Peers = Ordered<C::PublicKey>>,
     H: Hasher,
-    C: Signer,
+    C: PrivateKey,
```
patrick-ogrady: nit: Signer

examples/reshare/src/dkg/actor.rs
```diff
@@ -1,114 +1,145 @@
...
+    partition_prefix: String,
 }
 
 impl<E, P, H, C, V> Actor<E, P, H, C, V>
 where
     E: Spawner + Metrics + CryptoRngCore + Clock + GClock + Storage,
     P: Manager<PublicKey = C::PublicKey, Peers = Ordered<C::PublicKey>>,
     H: Hasher,
-    C: Signer,
+    C: PrivateKey,
```
patrick-ogrady: nit: Signer

examples/reshare/src/dkg/actor.rs
```diff
@@ -139,550 +169,341 @@ where
...
-        active_participants: Vec<C::PublicKey>,
-        inactive_participants: Vec<C::PublicKey>,
+        output: Option<Output<V, C::PublicKey>>,
+        share: Option<Share>,
         orchestrator: impl Reporter<Activity = orchestrator::Message<V, C::PublicKey>>,
         dkg_chan: (
             impl Sender<PublicKey = C::PublicKey>,
             impl Receiver<PublicKey = C::PublicKey>,
         ),
+        update_cb: Box<dyn UpdateCallBack<V, C::PublicKey>>,
```
patrick-ogrady: nit `callback`

examples/reshare/src/dkg/actor.rs
```diff
@@ -139,550 +169,341 @@ where
...
             impl Sender<PublicKey = C::PublicKey>,
             impl Receiver<PublicKey = C::PublicKey>,
         ),
+        mut update_cb: Box<dyn UpdateCallBack<V, C::PublicKey>>,
     ) {
-        let is_dkg = initial_public.is_none();
+        let max_read_size = NZU32!(self.peer_config.max_participants_per_round());
+        let is_dkg = output.is_none();
+
+        if self.epoch_metadata.get(&FixedBytes::new([])).is_none() {
```
patrick-ogrady: nit: use a `U64` key for forward compatibility (in case we add more keys)

examples/reshare/src/dkg/actor.rs
```diff
@@ -139,550 +169,341 @@ where
...
-        active_participants: Vec<C::PublicKey>,
-        inactive_participants: Vec<C::PublicKey>,
+        output: Option<Output<V, C::PublicKey>>,
+        share: Option<Share>,
         mut orchestrator: impl Reporter<Activity = orchestrator::Message<V, C::PublicKey>>,
         (sender, receiver): (
             impl Sender<PublicKey = C::PublicKey>,
             impl Receiver<PublicKey = C::PublicKey>,
         ),
+        mut update_cb: Box<dyn UpdateCallBack<V, C::PublicKey>>,
```
patrick-ogrady, line 128: nit: `callback`

examples/reshare/src/dkg/actor.rs
patrick-ogrady: Can make a const :thubs:

examples/reshare/src/dkg/actor.rs
```diff
@@ -139,550 +169,341 @@ where
...
+                    },
+                )
+                .await
+                .expect("should be able to update state");
+        }
 
-        // Start a muxer for the physical channel used by DKG/reshare
+        // Start a muxer for the physical channel used by DKG/reshare.
+        // Make sure to use a channel allowing sending messages to ourselves.
+        let (sender, receiver) = self_channel(self.signer.public_key(), 0, sender, receiver);
```
patrick-ogrady, line 153: See comment about not using this pattern.

examples/reshare/src/dkg/actor.rs
```diff
@@ -139,550 +169,341 @@ where
...
-                            next_players,
-                            &mut dkg_mux,
-                            self.rate_limit,
-                            &mut self.round_metadata,
-                        )
-                        .await;
+                        // If we did not block on processing the block, marshal could continue processing finalized blocks and start
+                        // at a future block after restart (leaving the application in an unrecoverable state where we are beyond the last epoch height
+                        // and not willing to enter the next epoch).
+                        response.acknowledge();
```
patrick-ogrady, line 353: This means that we may be required to re-process a block before this hits. Confirming this works (and that we have restart tests).

examples/reshare/src/dkg/actor.rs
```diff
@@ -139,550 +169,341 @@ where
...
-                    .cloned()
-                    .collect::<Vec<_>>();
-                candidates.extend(additions);
-                candidates
-            }
-            Ordering::Greater => {
-                candidates.truncate(target);
-                candidates
+            let logs = observer.logs().clone();
+            let (success, next_round, next_output, next_share) = if let Some(mb_player) = mb_player
```
patrick-ogrady, line 359: nit: `Some(player) = mb_player`?

examples/reshare/src/dkg/actor.rs
```diff
@@ -139,550 +169,341 @@ where
...
+                                    info!("dkg dealer exited");
+                                    break 'actor;
+                                }
+                            }
+                        }
 
-                            break;
+                        // At the midpoint of the epoch, construct the deal outcome for inclusion.
+                        if relative_height == BLOCKS_PER_EPOCH / 2 {
+                            if let Some(mb_dealer) = mb_dealer.take() {
```
patrick-ogrady, line 336: nit `Some(dealer) = `

examples/reshare/src/dkg/actor.rs
```diff
@@ -139,550 +169,341 @@ where
...
-                candidates
-            }
-            Ordering::Greater => {
-                candidates.truncate(target);
-                candidates
+            let logs = observer.logs().clone();
+            let (success, next_round, next_output, next_share) = if let Some(mb_player) = mb_player
+            {
+                let Ok(res) = mb_player.finalize(logs).await else {
+                    info!("dkg player exited");
```
patrick-ogrady, line 362: nit: would we want to panic here? Or is this just the case if the dkg fails?

patrick-ogrady: Or is this "shutdown"?

examples/reshare/src/dkg/actor.rs
```diff
@@ -139,550 +169,341 @@ where
...
+
+            // Persist the next epoch information
+            let epoch_state = EpochState {
+                epoch: next_epoch,
+                round: next_round,
+                output: next_output.clone(),
+                share: next_share.clone(),
+            };
+            self.epoch_metadata
+                .put_sync(FixedBytes::new([]), epoch_state)
```
patrick-ogrady: Is there a reason we don't keep multiple epochs of state? IMHO we should store by epoch or round number and then on startup just get the latest in epoch?

We can then delete our old share once the new epoch finalizes (to ensure we don't accidentally delete the share).

patrick-ogrady (changes requested):
Getting close!

consensus/src/ordered_broadcast/ack_manager.rs
```diff
@@ -254,20 +256,31 @@ mod tests {
...
 
         let chunk1 = Chunk::new(sequencer.clone(), height, Sha256::hash(b"payload1"));
         let chunk2 = Chunk::new(sequencer, height, Sha256::hash(b"payload2"));
 
-        let threshold1 =
-            helpers::add_acks_for_indices(&mut acks, &shares, chunk1, epoch, quorum, &[0, 1, 2]);
-        let threshold2 =
-            helpers::add_acks_for_indices(&mut acks, &shares, chunk2, epoch, quorum, &[3, 4, 5]);
+        let threshold1 = helpers::add_acks_for_indices(
+            &mut AckManager::<PublicKey, V, <Sha256 as Hasher>::Digest>::new(),
```
patrick-ogrady, line 269: We need to share the AckManager in both instances here.

consensus/src/aggregation/mocks/reporter.rs
cronokirby: In the math refactor I've been working on, I've replaced this function with a method on Poly, so that could be nice, but I don't think we need to change that here.

consensus/src/ordered_broadcast/ack_manager.rs
cronokirby: One issue with the test is that previously, the quorum was set to 3, so you could exercise this behavior of having two non-overlapping quorums with the same manager. Now, the issue is that since the quorum is set correctly, they will necessarily overlap, and a shared ack manager will (rightfully) complain that it's already seen signatures from a particular signer.

What's this test trying to exercise?

cryptography/src/bls12381/benches/dkg_recovery.rs
cronokirby: I refactored both of these to move the logic outside, and to share the logic.

examples/reshare/src/engine.rs
cronokirby: I don't think so, don't like hiding `Box<dyn>` behind an alias unless the type is really cumbersome.

it would maybe look like `DynUpdateCallBack<V, C::PublicKey>` but writing out `Box<dyn>` instead of `Dyn` is only an extra 5 characters, and is much clearer imo.

examples/reshare/src/self_channel.rs
cronokirby: It's definitely a hack, I'll give you that.

However, it is used with a very limited scope: a channel passed just to the dkg player and dealer.

This code does have to exist in some form somewhere. We could inline its behavior into the dealer and player, so that they both know that messages intended for a specific public key actually need to avoid hitting the p2p layer, and get routed to the other actor running adjacently. If you do this, you end up duplicating the logic in both the dealer and the player. This kind of naturally pushes you into having some kind of abstraction like this.

Maybe a cleaner way to represent what's going on is to be able to have some kind of "virtual" identity beyond just the node itself (kind of like how mux let's you do). Then, conceptually, you would have one identity for the dealer on this node, and one identity for the player on this node, and then the dealer would be sending messages to the player, preserving the property that an identity can't send messages to itself, only to other identities.

examples/reshare/src/validator.rs
cronokirby: exactly!

examples/reshare/src/dkg/actor/player/state.rs
cronokirby: It also allows you to pre-allocate a good size, so I think we want this.

examples/reshare/src/dkg/actor/player.rs
cronokirby: `select_loop!` can't be used because you break with a value.

When you try and replace the loop with `select!`, you run into the issue that that macro doesn't (seem to?) allow the bodies of each branch to borrow self mutable again.

examples/reshare/src/dkg/actor.rs
cronokirby: I think changing the name is more confusing than using the same name.

examples/reshare/src/dkg/actor.rs
cronokirby: This would only happen if the DKG Player actor shuts down before it's able to respond with the answer.

In general the approach I've taken for all the actors is to not treat "contagious shutdowns" as panics, but to just gracefully die as well.

Happy to change what the pattern is if this isn't idiomatic.

cryptography/src/bls12381/benches/dkg.rs
```diff
@@ -0,0 +1,142 @@
...
+}
+
+fn benchmark_dkg(c: &mut Criterion, previous: bool) {
+    let mut rng = StdRng::seed_from_u64(0);
+    for &n in CONTRIBUTORS {
+        let t = quorum(n);
+        let bench = Bench::new(&mut rng, previous, n);
+        for &concurrency in CONCURRENCY {
+            c.bench_function(
+                &format!("{}/n={} t={} conc={}", module_path!(), n, t, concurrency),
```
patrick-ogrady: If you take this route (makes sense), we need to add `previous` to this `format` line.

patrick-ogrady: We need to use `previous` here if you take this dedup route (which is nice). Otherwise, both tests will share the same module_path?

cryptography/src/bls12381/dkg.rs
```diff
@@ -1578,7 +1578,7 @@ mod test_plan {
                 // Every dealer must have been a player in the previous round
                 for &d in &self.dealers {
                     if !prev_players.contains(&d) {
-                        return Err(anyhow!("dealer {d} was not a player in previous round"));
+                        return Err(anyhow!("dealer {} was not a player in previous round", d));
```
patrick-ogrady: nit: put `d` in `{}`

cryptography/src/bls12381/dkg.rs
```diff
@@ -1957,7 +1957,7 @@ mod test_plan {
                 }
 
                 // Generate and verify threshold signature
-                let test_message = format!("test message round {i_round}").into_bytes();
+                let test_message = format!("test message round {}", i_round).into_bytes();
```
patrick-ogrady: nit: put `i_round` in `{}`

docs/blogs/commonware-the-anti-framework.html
```diff
@@ -132,7 +132,7 @@ <h2>Commonware: Now Including Consensus</h2>
...
-        <p>You can see what it looks like to build with consensus::simplex in <a href="https://docs.rs/commonware-log">examples::log</a>, p2p::authenticated in <a href="https://docs.rs/commonware-chat">examples::chat</a>, and cryptography::bls12381 in <a href="https://docs.rs/commonware-vrf">examples::vrf</a>.</p>
+        <p>You can see what it looks like to build with consensus::simplex in <a href="https://docs.rs/commonware-log">examples::log</a>, p2p::authenticated in <a href="https://docs.rs/commonware-chat">examples::chat</a>, and cryptography::bls12381 in <a href="https://github.com/commonwarexyz/monorepo/tree/v0.0.63/examples/vrf">examples::vrf</a>.</p>
```
patrick-ogrady, line 135: Nice!

docs/index.html
```diff
@@ -116,7 +116,7 @@ <h3><a href="https://docs.rs/commonware-reshare">reshare</a></h3>
         <p>Reshare a threshold secret over an epoched log.</p>
         <h3><a href="https://docs.rs/commonware-sync">sync</a></h3>
         <p>Synchronize state between a server and client.</p>
-        <h3><a href="https://docs.rs/commonware-vrf">vrf</a></h3>
+        <h3><a href="https://github.com/commonwarexyz/monorepo/tree/v0.0.63/examples/vrf">vrf</a></h3>
```
patrick-ogrady: Let's remove this from here (ok to keep links like you did in blogs but we don't need on homepage)

examples/reshare/src/setup.rs
```diff
@@ -203,30 +245,17 @@ fn generate_configs(
...
                 .map(|(signer, _)| signer.public_key())
-                .skip(args.num_participants_per_epoch as usize)
-                .take((args.num_peers - args.num_participants_per_epoch) as usize)
                 .collect(),
         }
     } else {
         PeerConfig {
-            num_participants_per_epoch: args.num_participants_per_epoch,
-            active: identities
+            num_participants_per_round: vec![args.num_participants_per_epoch],
```
patrick-ogrady, line 252: Nice approach to permit the same code to work for tests.

examples/reshare/src/dkg/state.rs
```diff
@@ -0,0 +1,376 @@
...
+    /// Returns all finalized dealer logs for the given epoch.
+    pub async fn logs(&self, epoch: Epoch) -> BTreeMap<P, DealerLog<V, P>> {
+        self.inner
+            .read()
+            .await
+            .dkg_msgs(epoch)
+            .await
+            .filter_map(|x| {
+                futures::future::ready(match x {
+                    DkgEvent::Log(x0, x1) => Some((x0, x1)),
```
patrick-ogrady, line 310: I don't think this is an issue but there can be duplicates here. Do we care if they are the same or not?

examples/reshare/src/dkg/state.rs
```diff
@@ -0,0 +1,376 @@
...
+use std::{
+    collections::BTreeMap,
+    num::{NonZeroU32, NonZeroUsize},
+    sync::Arc,
+};
+
+const PAGE_SIZE: NonZeroUsize = NZUsize!(1 << 12);
+const POOL_CAPACITY: NonZeroUsize = NZUsize!(1 << 20);
+const WRITE_BUFFER: NonZeroUsize = NZUsize!(1 << 12);
+const READ_BUFFER: NonZeroUsize = NZUsize!(1 << 12);
```
patrick-ogrady: Would suggest increasing the READ_BUFFER given how many scans you'll be doing?

examples/reshare/src/dkg/state.rs
```diff
@@ -0,0 +1,376 @@
...
+                },
+            )
+            .await?,
+            dkg_msgs: SVJournal::init(
+                cell.with_label("dkg-msgs"),
+                SVConfig {
+                    partition: format!("{partition_prefix}_dkg_msgs"),
+                    compression: None,
+                    codec_config: max_read_size,
+                    buffer_pool: buffer_pool.clone(),
```
patrick-ogrady: nit: unnecessary clone?

examples/reshare/src/dkg/egress.rs
```diff
@@ -0,0 +1,62 @@
...
+    dkg::Output,
+    primitives::{group::Share, variant::Variant},
+};
+use std::{future::Future, pin::Pin};
+
+/// An update from the DKG Actor.
+pub enum Update<V: Variant, P> {
+    /// DKG at this epoch has failed.
+    Failure {
+        #[allow(dead_code)]
```
patrick-ogrady: If there is dead_code, we should remove it?

examples/reshare/src/dkg/actor/dealer.rs
```diff
@@ -0,0 +1,204 @@
...
+        }
+        debug!("dealer shutting down");
+    }
+
+    async fn ack(&mut self, replay: bool, player: C::PublicKey, ack: PlayerAck<C::PublicKey>) {
+        if !self.unsent_priv_msgs.contains_key(&player) {
+            return;
+        }
+        if let Err(e) = self.dealer.receive_player_ack(player.clone(), ack.clone()) {
+            info!("bad player ack: {}", e);
```
patrick-ogrady: nit: warn? can we also log `player`?

You should write as `info!(?player, ?e, "bad player ack");`

examples/reshare/src/dkg/actor/dealer.rs
```diff
@@ -0,0 +1,204 @@
...
+        share: Option<Share>,
+    ) -> (Self, Mailbox<V, C>) {
+        let (outbox, inbox) = mpsc::channel(1);
+        let mailbox = Mailbox(outbox);
+
+        let epoch = Epoch::new(round_info.round());
+        let replay = state.player_acks(epoch).await;
+
+        let (dealer, pub_msg, priv_msgs) = Dealer::start(
+            Transcript::resume(rng_seed).noise(b"dealer rng"),
```
patrick-ogrady: nit: `dealer-rng`

examples/reshare/src/dkg/actor.rs
patrick-ogrady: I think leaving as graceful is fine...we should just make `debug` if it is meant to handle this?

examples/reshare/src/dkg/actor.rs
```diff
@@ -139,550 +97,337 @@ where
...
+                // won't need to finalize further blocks after the DKG completes).
+                orchestrator
+                    .report(orchestrator::Message::Exit(epoch))
+                    .await;
+                // Keep running until killed to keep the orchestrator mailbox alive, allowing
+                // peers that may have gone offline to catch up.
+                //
+                // The initial DKG process will never be exited automatically, assuming coordination
+                // between participants is manual.
+                info!("DKG told to stop post update, now waiting...");
```
patrick-ogrady: nit: `"DKG complete...waiting for shutdown`

patrick-ogrady (commented):
I think the only remaining points of contention are the "self channel" and "select_biased!"

cryptography/src/bls12381/benches/dkg.rs
cronokirby: Oh yeah, I overlooked this

cronokirby: Oh yeah, I overlooked this

cronokirby: Oh yeah, I overlooked this

examples/reshare/src/dkg/state.rs
cronokirby: Honest dealers should never post a log twice. For dishonest dealers, all that matters is that people agree on which log to use if multiple get posted. In this case, people will deterministically pick the most recent log. If the picked the least recent log this would also work.

cronokirby: Honest dealers should never post a log twice. For dishonest dealers, all that matters is that people agree on which log to use if multiple get posted. In this case, people will deterministically pick the most recent log. If the picked the least recent log this would also work.

cronokirby: Honest dealers should never post a log twice. For dishonest dealers, all that matters is that people agree on which log to use if multiple get posted. In this case, people will deterministically pick the most recent log. If the picked the least recent log this would also work.

cronokirby: Honest dealers should never post a log twice. For dishonest dealers, all that matters is that people agree on which log to use if multiple get posted. In this case, people will deterministically pick the most recent log. If the picked the least recent log this would also work.

examples/reshare/src/dkg/actor/dealer.rs
```diff
@@ -0,0 +1,206 @@
...
+                        finalize = Some(cb_in);
+                        break;
+                    }
+                }
+            }
+        }
+        if let Some(cb_in) = finalize {
+            self.finalize(cb_in);
+        }
+        debug!("dealer shutting down");
```
clabby: Let's add some more information here
```suggestion
        debug!(epoch = self.epoch, "dealer shutting down");
```

examples/reshare/src/dkg/actor/dealer.rs
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    async fn ack(&mut self, replay: bool, player: C::PublicKey, ack: PlayerAck<C::PublicKey>) {
+        if !self.unsent_priv_msgs.contains_key(&player) {
+            return;
+        }
+        if let Err(e) = self.dealer.receive_player_ack(player.clone(), ack.clone()) {
+            warn!(?player, ?e, "bad player ack");
```
clabby:
```suggestion
            warn!(epoch = self.epoch, ?player, ?e, "bad player ack");
```

examples/reshare/src/dkg/actor/player.rs
```diff
@@ -0,0 +1,233 @@
...
+                let Ok((dealer, mut msg_bytes)) = msg else {
+                    break;
+                };
+                let Ok((pub_msg, priv_msg)) = <(DealerPubMsg<V>, DealerPrivMsg) as Read>::read_cfg(
+                    &mut msg_bytes,
+                    &(self.max_read_size, ()),
+                ) else {
+                    // If we can't read the message, ignore it.
+                    continue;
+                };
```
clabby, line 165: We should block the dealer's public key if they send us a malformed message.

examples/reshare/src/dkg/actor/dealer.rs
```diff
@@ -0,0 +1,206 @@
...
+                break;
+            },
+            msg = self.from_players.recv() => {
+                let Ok((player, mut msg_bytes)) = msg else {
+                    // The network is dead, so terminate.
+                    break;
+                };
+                let Ok(ack) = PlayerAck::<C::PublicKey>::read(&mut msg_bytes) else {
+                    continue;
+                };
```
clabby, line 149: We should block the player if they send us a malformed acknowledgement

examples/reshare/src/dkg/actor/player.rs
```diff
@@ -0,0 +1,233 @@
...
+                        finalize = Some((logs, output));
+                        break;
+                    }
+                }
+            }
+        }
+        if let Some((logs, cb_in)) = finalize {
+            self.finalize(logs, cb_in);
+        }
+        debug!("player shutting down");
```
clabby:
```suggestion
        debug!(epoch = self.epoch, "player shutting down");
```

examples/reshare/src/validator.rs
clabby, line 1: The test refactor here is _amazing_ - nice work.

AFAICT, we didn't lose any coverage except for determinism. Can you add a test that runs two seeded `Plan`s  and asserts that `context.auditor().state()` is equal?

examples/reshare/src/dkg/state.rs
```diff
@@ -0,0 +1,376 @@
...
+            Self::Log(x0, x1) => x0.encode_size() + x1.encode_size(),
+        }
+    }
+}
+
+impl<V: Variant, P: PublicKey> Write for DkgEvent<V, P> {
+    fn write(&self, buf: &mut impl bytes::BufMut) {
+        match self {
+            Self::Dealer(x0, x1, x2) => {
+                0u8.write(buf);
```
clabby, line 98: Not necessary, but since these are used in `Write` + `Read`, maybe add some `const`s?

examples/reshare/src/namespace.rs
clabby, line 1: nit - this can just live in `validator.rs`.

examples/reshare/src/dkg/actor.rs
clabby, line 1: I can't comment on `core.rs`, but perhaps we should include a block validity condition on the `SignedDealerLog` having both:
1. A valid signature
2. Correct codec
?

examples/reshare/src/dkg/actor.rs
```diff
@@ -139,550 +97,337 @@ where
...
+                output,
+                share,
+            };
+            state.append_dkg_state(initial_state).await;
+        };
 
-        // Start a muxer for the physical channel used by DKG/reshare
+        // Start a muxer for the physical channel used by DKG/reshare.
+        // Make sure to use a channel allowing sending messages to ourselves.
+        let (sender, receiver) = self_channel(self.signer.public_key(), 0, sender, receiver);
```
clabby, line 153: Buffer size needs to be > 0

examples/reshare/src/dkg/actor/dealer.rs
```diff
@@ -0,0 +1,206 @@
...
+        }
+        self.unsent_priv_msgs.remove(&player);
+        if !replay {
+            self.state.append_player_ack(self.epoch, player, ack).await;
+        }
+    }
+
+    async fn transmit(&mut self) -> Result<(), S::Error> {
+        for (target, priv_msg) in &self.unsent_priv_msgs {
+            self.to_players
```
clabby, line 193: We need to rate limit this channel

examples/reshare/src/dkg/actor/player.rs
```diff
@@ -0,0 +1,233 @@
...
+                self.state
+                    .append_dealer_msg(self.epoch, dealer, pub_msg, priv_msg)
+                    .await;
+            }
+        };
+    }
+
+    async fn transmit(&mut self) -> Result<(), S::Error> {
+        for (dealer, ack) in &self.acks {
+            self.to_dealers
```
clabby, line 213: We need to rate limit this channel

examples/reshare/src/namespace.rs
cronokirby: It's also used in `dkg/actor`, and I don't want to have that depend on `validator` directly.

examples/reshare/src/dkg/state.rs
cronokirby: This is the common pattern across the codebase, afaict.

It is possible to make a mistake here, but that's also the case if you introduce consts.

Instead of having to remember that

```
Dealer => 0
...
```

you remember that


```
Dealer => TAG_DEALER
...
```

Certainly, the association is a bit clearer, but a mistake is nonetheless possible.

Given that it's used in exactly two places, side by side, I don't think it's too harmful to just have it inline.

clabby: Fair enough - not very opinionated on this one.

examples/reshare/src/dkg/actor.rs
cronokirby: This is not the buffer size for the network, but for the `futures::channel::mpsc` which is used to send and receive messages from yourself; i.e. between the dealer and the player actors in the same runtime. The Mux also then clones the sender, so both the dealer and player can send 1 message to each other without blocking.

The back pressure after sending more than 1 message is intentional. In any case, I expect the other actor running in the same runtime to be able to pick up the message in a timely fashion. If this *isn't* possible, this is indicative of a bug in the way the actors are wired up.

My thinking is that you want to discover such bugs early, and hiding them by increasing the buffer size artificially only makes them flakier.

For buffers with the outside world, that's a different issue, but for internal buffers between actors you're running in the same runtime, I think we want things to backpressure immediately.

examples/reshare/src/dkg/actor/player.rs
cronokirby: I agree on the case where you receive something, you parse it, and it's ostensibly really only sent by someone misbehaving. But here, this could also indicate some kind of bug in the program, where you're not handling the fact that you could receive some other message, so I'm less keen on blocking.

examples/reshare/src/dkg/actor/dealer.rs
cronokirby: What's the thinking behind rate limiting it here, versus just relying on the rate limiting inherited from validator?

examples/reshare/src/dkg/actor/player.rs
clabby: Here we know that we've reserved this sub-channel for this message type; A bug in the program that sent a different type over the channel would be caught with an explicit block.

I agree that it'd be nice to express this in the type of the sender/receiver (#2049), though I think we should enforce this.

examples/reshare/src/validator.rs
cronokirby: So, it turns out that things aren't deterministic.

I think this might be because the runtime seed determines the RNG you use for the DKG itself?

examples/reshare/src/dkg/actor/dealer.rs
clabby: The rate limiting on our own sends is intended to ensure that other peers don't rate limit us. The rate limit we establish on the validator is per-peer for incoming messages, but we need to ensure we don't send too many outbound messages as to hit this limit. Patrick pointed this out on my original implementation [here](https://github.com/commonwarexyz/monorepo/pull/1796#discussion_r2440831182).

examples/reshare/src/validator.rs
clabby: Just tried it out - the test passes for me on a two-epoch run w/ threshold:

```diff
diff --git a/examples/reshare/src/validator.rs b/examples/reshare/src/validator.rs
index 6204e6de7c..fb82a1b519 100644
--- a/examples/reshare/src/validator.rs
+++ b/examples/reshare/src/validator.rs
@@ -382,7 +382,7 @@
     }
 
     impl Plan {
-        async fn run_inner<S>(self, mut ctx: deterministic::Context) -> anyhow::Result<()>
+        async fn run_inner<S>(self, mut ctx: deterministic::Context) -> anyhow::Result<String>
         where
             S: Scheme<PublicKey = PublicKey>,
             SchemeProvider<S, PrivateKey>:
@@ -494,7 +494,7 @@
 
                         if status.values().filter(|x| **x >= epoch).count() >= self.total as usize {
                             if successes >= self.target {
-                                return Ok(());
+                                return Ok(ctx.auditor().state());
                             } else {
                                 current_epoch = current_epoch.next();
                             }
@@ -542,7 +542,7 @@
             }
         }
 
-        fn run<S>(self) -> anyhow::Result<()>
+        fn run<S>(self) -> anyhow::Result<String>
         where
             S: Scheme<PublicKey = PublicKey>,
             SchemeProvider<S, PrivateKey>:
@@ -596,6 +596,42 @@
 
     #[test_group("slow")]
     #[test_traced("INFO")]
+    fn two_epoch_threshold_scheme_success_deterministic() {
+        for seed in 0..5 {
+            let a = (Plan {
+                seed,
+                total: 4,
+                per_round: vec![4],
+                link: Link {
+                    latency: Duration::from_millis(0),
+                    jitter: Duration::from_millis(0),
+                    success_rate: 1.0,
+                },
+                target: 2,
+                crash: None,
+            }
+            .run::<ThresholdScheme<MinSig>>())
+            .unwrap();
+            let b = (Plan {
+                seed,
+                total: 4,
+                per_round: vec![4],
+                link: Link {
+                    latency: Duration::from_millis(0),
+                    jitter: Duration::from_millis(0),
+                    success_rate: 1.0,
+                },
+                target: 2,
+                crash: None,
+            }
+            .run::<ThresholdScheme<MinSig>>())
+            .unwrap();
+            assert_eq!(a, b);
+        }
+    }
+
+    #[test_group("slow")]
+    #[test_traced("INFO")]
     fn four_epoch_ed_scheme_all_participants() {
         if let Err(e) = (Plan {
             seed: 0,
```

cronokirby: Oh, I see, yeah I was testing that two runs with different seeds have the same state, but that seems wrong.

cryptography/src/bls12381/benches/dkg.rs
```diff
@@ -0,0 +1,154 @@
...
+                        BatchSize::SmallInput,
+                    );
+                },
+            );
+        }
+    }
+}
+
+fn benchmark_dkg_recovery(c: &mut Criterion) {
+    benchmark_dkg(c, false);
```
patrick-ogrady, line 143: nit: just put `suffix` here

examples/reshare/src/setup.rs
```diff
@@ -68,20 +74,50 @@ impl ParticipantConfig {
...
 #[derive(Debug, Clone, Serialize, Deserialize)]
-pub struct PeerConfig {
-    /// The number of participants per epoch.
-    pub num_participants_per_epoch: u32,
+pub struct PeerConfig<P: commonware_cryptography::PublicKey = PublicKey> {
+    /// The number of participants per round.
+    ///
+    /// This is a vec which cycles through different numbers in each round.
+    ///
+    /// This MUST be non-empty.
```
patrick-ogrady, line 82: Another case where @andresilva's non-empty vec could be useful...

examples/reshare/src/self_channel.rs
patrick-ogrady, line 1: Its a bigger lift than I thought to push this into existing `p2p` implementations (and is better implemented as a wrapper in `p2p`, if we go this route).

examples/reshare/src/setup.rs
```diff
@@ -68,20 +74,50 @@ impl ParticipantConfig {
...
+
+    /// Pick the dealers for a particular round.
+    ///
+    /// The first round will use the first [`Self::num_participants_in_round`] players
+    /// as the dealers.
+    ///
+    /// Subsequent rounds will their corresponding [`Self::num_participants_in_round`], and
+    /// so on.
+    pub fn dealers(&self, round: u64) -> Set<P> {
+        let p_iter = self.participants.iter().cloned();
```
patrick-ogrady, line 114: I think this should already be deduped?

examples/reshare/src/setup.rs
```diff
@@ -203,32 +242,17 @@ fn generate_configs(
...
-                .filter(|(_, share)| share.is_some())
-                .map(|(signer, _)| signer.public_key())
-                .collect(),
-            inactive: identities
-                .iter()
-                .filter(|(_, share)| share.is_none())
-                .map(|(signer, _)| signer.public_key())
-                .collect(),
+            num_participants_per_round: vec![args.num_participants_per_epoch],
+            participants: Set::from_iter_dedup(
```
patrick-ogrady, line 253: Should error here rather than allowing duplicates?

patrick-ogrady (commented):
thinking about self_channel

examples/reshare/src/dkg/actor/player.rs
```diff
@@ -0,0 +1,235 @@
...
+    /// `from_dealers` lets us receive messages from the dealers.
+    /// `round_info` is the configuration for this round of the DKG.
+    /// `me` is the private key identifying this player.
+    pub async fn init(
+        ctx: E,
+        state: State<E, V, C::PublicKey>,
+        to_dealers: S,
+        from_dealers: R,
+        round_info: Info<V, C::PublicKey>,
+        max_read_size: NonZeroU32,
```
patrick-ogrady, line 110: We should infer this from `round_info`

examples/reshare/src/dkg/state.rs
```diff
@@ -0,0 +1,376 @@
...
+struct Inner<E: Storage + Metrics, V: Variant, P: PublicKey> {
+    dkg_states: CVJournal<ContextCell<E>, DkgState<V, P>>,
+    dkg_msgs: SVJournal<ContextCell<E>, DkgEvent<V, P>>,
+}
+
+impl<E: Storage + Metrics, V: Variant, P: PublicKey> Inner<E, V, P> {
+    async fn init(
+        context: E,
+        partition_prefix: &str,
+        max_read_size: NonZeroU32,
```
patrick-ogrady, line 147: I think it would make sense to keep this as `max_participants_per_epoch`?

examples/reshare/src/dkg/state.rs
```diff
@@ -0,0 +1,376 @@
...
+    fn clone(&self) -> Self {
+        Self {
+            inner: self.inner.clone(),
+        }
+    }
+}
+
+impl<E: Storage + Metrics, V: Variant, P: PublicKey> State<E, V, P> {
+    /// Initialize storage, creating partitions if needed.
+    pub async fn init(context: E, partition_prefix: &str, max_read_size: NonZeroU32) -> Self {
```
patrick-ogrady, line 258: see earlier comment about `max_participants_per_epoch`

examples/reshare/src/setup.rs
```diff
@@ -322,7 +346,7 @@ mod serde_hex_vec {
 
                     out.push(T::decode(&mut bytes.as_ref()).map_err(serde::de::Error::custom)?);
                 }
-                Ok(out)
+                Ok(Set::from_iter_dedup(out))
```
patrick-ogrady, line 349: See earlier comment about erroring on incorrect configs rather than handling them?

patrick-ogrady (commented):
Very close now. Working on a better supported self channel now.

examples/reshare/src/dkg/actor/player.rs
cronokirby: Maybe, but that's a bit of an anti-pattern in general because now this configuration represents the largest size across epochs (because of storage). In this specific actor you could use a value derived from a round, but being able to derive such a value from a round tempts the developer into having round-specific codec configs in other places that create bugs when the committee size changes.

examples/reshare/src/dkg/actor.rs
```diff
@@ -140,551 +97,338 @@ where
...
+            let round_info = Info::new(
+                namespace::APPLICATION,
+                epoch.get(),
+                dkg_state.output.clone(),
+                dealers,
+                players,
+            )
+            .expect("round info configuration should be correct");
+            let (to_players, from_dealers) = dkg_mux
+                .register(0)
```
patrick-ogrady, line 224: We should be setting this based on `epoch`? The code in `main` properly separates these channels by epoch (whereas this may involve you getting messages from a peer from an old epoch).

examples/reshare/src/dkg/actor.rs
```diff
@@ -140,551 +97,338 @@ where
...
+                Some(mb)
+            } else {
+                None
+            };
+            let mut mb_player = if am_player {
+                let (player, mb) = player::Actor::init(
+                    self.context.with_label("player"),
+                    state.clone(),
+                    to_dealers,
+                    from_dealers,
```
patrick-ogrady, line 252: `dealer` puts `(to_players, from_players)` as a single arg.

examples/reshare/src/dkg/state.rs
patrick-ogrady: > Honest dealers should never post a log twice

In the current code they may, we only check if their commitment is finalized before posting (if post twice in a row, before parent is finalized, could happen).

examples/reshare/src/dkg/state.rs
```diff
@@ -0,0 +1,376 @@
...
+        }
+    }
+}
+
+struct Inner<E: Storage + Metrics, V: Variant, P: PublicKey> {
+    dkg_states: CVJournal<ContextCell<E>, DkgState<V, P>>,
+    dkg_msgs: SVJournal<ContextCell<E>, DkgEvent<V, P>>,
+}
+
+impl<E: Storage + Metrics, V: Variant, P: PublicKey> Inner<E, V, P> {
```
patrick-ogrady, line 143: While this approach is cool (avoids having to maintain a map and log like we do in consensus), I wonder how much slower it is.

examples/reshare/src/dkg/actor.rs
```diff
@@ -140,551 +97,338 @@ where
...
         (sender, receiver): (
             impl Sender<PublicKey = C::PublicKey>,
             impl Receiver<PublicKey = C::PublicKey>,
         ),
+        mut update_cb: Box<dyn UpdateCallBack<V, C::PublicKey>>,
     ) {
-        let is_dkg = initial_public.is_none();
+        let max_read_size = NZU32!(self.peer_config.max_participants_per_round());
+        let is_dkg = output.is_none();
+        // **This stores persist private key material to disk. In a production
```
patrick-ogrady, line 132: Add `\n` before.

examples/reshare/src/dkg/actor.rs
patrick-ogrady: Commenting to resurrect this convo (I'm ok as-is)

examples/reshare/src/dkg/actor/dealer.rs
patrick-ogrady: Resurrecting conversation about rate limiting (to prevent stalling all message handling on peer).

examples/reshare/src/dkg/actor/player.rs
patrick-ogrady: My thinking is just to use a tighter bound if we have one. I don't think that confuses too much.

examples/reshare/src/dkg/actor/dealer.rs
patrick-ogrady: Yep, thanks for calling out.

examples/reshare/src/dkg/actor.rs
```diff
@@ -140,551 +97,338 @@ where
...
+            }
 
-            // Exit last consensus instance to avoid useless work while we wait for shutdown (we
-            // won't need to finalize further blocks after the DKG completes).
-            orchestrator
-                .report(orchestrator::Message::Exit(current_epoch))
+            info!(
+                success,
+                ?epoch,
+                "finalized epoch's reshare; instructing reconfiguration after reshare.",
```
patrick-ogrady, line 394: nit: improve this log message

examples/reshare/src/validator.rs
```diff
@@ -156,1932 +155,744 @@ where
 mod test {
```
patrick-ogrady, line 155: TODO: rename channels to `votes`/`certificates`/`resolver`

examples/reshare/src/validator.rs
```diff
@@ -156,1932 +155,744 @@ where
...
+        fn reckon(mut rng: impl CryptoRngCore, total: u32, per_round: &[u32]) -> Self {
+            let mut participants = (0..total)
+                .map(|i| {
+                    let sk = PrivateKey::from_seed(i as u64);
+                    (sk.public_key(), (sk, None::<Share>))
+                })
+                .collect::<BTreeMap<_, _>>();
+            let peer_config = PeerConfig {
+                num_participants_per_round: per_round.to_vec(),
+                participants: Set::from_iter_dedup(participants.keys().cloned()),
```
patrick-ogrady, line 242: Error on duplicate here.

examples/reshare/src/validator.rs
```diff
@@ -156,1932 +155,744 @@ where
...
-                            inactive_participants: Vec::default(),
-                            num_participants_per_epoch: validators.len() as u32,
-                            dkg_rate_limit: Quota::per_second(NZU32!(128)),
-                            orchestrator_rate_limit: Quota::per_second(NZU32!(1)),
-                            partition_prefix: format!("validator_{idx}"),
-                            freezer_table_initial_size: 1024, // 1mb
-                        },
-                    )
-                    .await;
+            debug!("creating team with {} participants", self.total);
```
patrick-ogrady, line 408: (participants = self.total, "creating team")

examples/reshare/src/validator.rs
```diff
@@ -156,1932 +155,744 @@ where
...
+                        let Some(update) = update else {
+                            return Err(anyhow!("update channel closed unexpectedly"));
+                        };
+                        let (epoch, output) = match update.update {
+                            Update::Failure { epoch } => {
+                                info!(epoch = ?epoch, pk = ?update.pk, "DKG failure");
+                                (epoch, None)
+                            }
+                            Update::Success { epoch, output, .. } => {
+                                info!(epoch = ?epoch, pk = ?update.pk, ?output, "DKG success");
```
patrick-ogrady, line 454: nit: consistent `\n`?

examples/reshare/src/validator.rs
```diff
@@ -156,1932 +155,744 @@ where
...
+                latency: Duration::from_millis(0),
+                jitter: Duration::from_millis(0),
+                success_rate: 1.0,
+            },
+            target: 4,
+            crash: None,
+        }
+        .run::<ThresholdScheme<MinSig>>())
+        {
+            panic!("failure: {e}");
```
patrick-ogrady, line 669: nit: just call run.unwrap?

examples/reshare/src/dkg/actor.rs
patrick-ogrady: A fix here is to instead have a single channel and route messages to a dealer or player actor based on type (in a single logical loop) rather than trying to create two different sub-actors.

examples/reshare/src/dkg/actor.rs
```diff
@@ -140,551 +97,338 @@ where
...
-                .try_collect::<Set<_>>()
-                .expect("participants are unique"),
-            &mut dkg_mux,
-            self.rate_limit,
-            &mut self.round_metadata,
-        )
-        .await;
+            let self_pk = self.signer.public_key();
+            // If we've already submitted a log, there's no point acting as the dealer.
+            let am_dealer = dealers.position(&self_pk).is_some()
```
patrick-ogrady, line 198: I "think" this only works because dealer doesn't actually recover a share (whereas if we were a player that had already acked and we used observe, we wouldn't get our share.)

examples/reshare/src/dkg/actor/player.rs
patrick-ogrady: To do this, we need to fix the sub-channel indexing first (as we may receive a message from an old epoch as currently written).

examples/reshare/src/dkg/actor/player.rs
```diff
@@ -0,0 +1,235 @@
...
+                    continue;
+                };
+                self.dealer_message(false, dealer, pub_msg, priv_msg).await;
+            },
+            res = self.inbox.next() => {
+                let Some(msg) = res else {
+                    break;
+                };
+                match msg {
+                    Message::Transmit => if self.transmit().await.is_err() {
```
patrick-ogrady, line 173: Why do we wait to send back to dealers until transmit is called? If we send right away, they can stop sending their commitment to us.

The current logic has us send back all our acks each block.

examples/reshare/src/dkg/actor/player.rs
```diff
@@ -0,0 +1,235 @@
...
+            if !replay {
+                self.state
+                    .append_dealer_msg(self.epoch, dealer, pub_msg, priv_msg)
+                    .await;
+            }
+        };
+    }
+
+    async fn transmit(&mut self) -> Result<(), S::Error> {
+        for (dealer, ack) in &self.acks {
```
patrick-ogrady, line 212: In dealer, we prevent re-sending by tracking what's been sent. We can't do as much here because we don't get an ack from the dealer.

Instead, we should just rely on the dealer to keep pinging us if there is something to sign and respond (rather than sending each block during the first part of the epoch).

examples/reshare/src/dkg/actor/player.rs
```diff
@@ -0,0 +1,235 @@
...
+    }
+
+    async fn dealer_message(
+        &mut self,
+        replay: bool,
+        dealer: C::PublicKey,
+        pub_msg: DealerPubMsg<V>,
+        priv_msg: DealerPrivMsg,
+    ) {
+        if let Some(ack) =
```
patrick-ogrady, line 198: If we previously received a dealer message, we need to ensure whatever we sign is consistent?

examples/reshare/src/dkg/actor.rs
```diff
@@ -140,551 +97,338 @@ where
...
-                                    success,
-                                ) => (next_dealers, Some(public), Some(share), success),
-                                (next_dealers, RoundResult::Polynomial(public), success) => {
-                                    (next_dealers, Some(public), None, success)
-                                }
-                                (next_dealers, RoundResult::None, success) => {
-                                    (next_dealers, None, None, success)
+                        if let Some(log) = block.log {
+                            if let Some((dealer, log)) = log.check(&round_info) {
+                                // If we see our dealing outcome in a finalized block,
```
patrick-ogrady, line 307: This doesn't prevent us from ever posting twice (can post before previous is finalized) but does help. Let's improve the comment.

examples/reshare/src/dkg/actor/player.rs
patrick-ogrady, line 212: Unlike in dealer, we re-send acks here forever because we never receive an ack from the dealer that our message was received.

The old code just responded to dealers to prevent this issue...

PR #2355, opened 2025-12-01
[Storage] Implement separate `Operation` type for `adb::immutable`

danlaine:
Also moves `keyless` operation underneath `variable` sub-module


storage/src/adb/operation/variable/immutable.rs
```diff
@@ -0,0 +1,289 @@
+//! Operations for immutable authenticated databases.
```
roberto-bayardo, line 1: I think it makes sense to move immutable.rs and keyless.rs into the variable/ sub-module.

danlaine: done

PR #2356, opened 2025-12-01
[storage/adb/operation] move fixed operation tests into fixed submodule

roberto-bayardo:


storage/src/adb/operation/fixed/mod.rs
```diff
@@ -18,3 +18,385 @@ fn ensure_zeros(buf: &mut impl Buf, size: usize) -> Result<(), CodecError> {
...
     Ok(())
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::{
+        adb::operation::{
+            fixed::{ordered::Operation as FixedOrdered, unordered::Operation as FixedUnordered},
+            keyless::Operation as Keyless,
```
danlaine: This is actually a variable-length value operation, isn't it? https://github.com/commonwarexyz/monorepo/pull/2355 moves `keyless` to the variable sub-module.

roberto-bayardo: Oh shoot didn't see these in there (just blindly copied).  I'll fix this up once your change is in and I rebase on it.

roberto-bayardo: fixed

Issue #2361, opened 2025-12-02
[cryptography] Require namespaces


clabby:
## Overview

In the `Verifier` and `Signer` traits in `cryptography`, `namespace` is currently optional:

https://github.com/commonwarexyz/monorepo/blob/5da33d2ebe54ce8ac1960f0b8b131d6fbbd216ce/cryptography/src/lib.rs#L87-L94

https://github.com/commonwarexyz/monorepo/blob/5da33d2ebe54ce8ac1960f0b8b131d6fbbd216ce/cryptography/src/lib.rs#L49-L59

We should change this to require namespaces always in the traits, while preserving the ability for the tests to run over vectors that do not use namespaces in each individual scheme.

PR #2364, opened 2025-12-02
[cryptography]! Require namespace in `Signer` + `Verifier` + `BatchVerifier`

clabby:
## Overview

Requires a namespace in `Signer` + `Verifier` + `BatchVerifier`.

closes https://github.com/commonwarexyz/monorepo/issues/2361

### DIFF
--- a/cryptography/src/secp256r1/scheme.rs
+++ b/cryptography/src/secp256r1/scheme.rs
@@ -468,7 +468,7 @@ mod tests {
     #[test]
     fn test_codec_signature() {
         let private_key = create_private_key();
-        let original = private_key.sign(b"test-namespace", "Hello World".as_bytes());
+        let original = private_key.sign(NAMESPACE, "Hello World".as_bytes());
 
         let encoded = original.encode();
         assert_eq!(encoded.len(), SIGNATURE_LENGTH);
--- a/cryptography/src/transcript.rs
+++ b/cryptography/src/transcript.rs
@@ -265,11 +265,15 @@ impl Transcript {
     /// - signing the operations that have been performed on the transcript,
     /// - or, equivalently, signing randomness or a summary extracted from the transcript.
     pub fn sign<S: Signer>(&self, s: &S) -> <S as Signer>::Signature {
+        // Note: We pass an empty namespace here, since the namespace may be included
+        // within the transcript summary already via `Self::new`.
         s.sign(b"", self.summarize().hash.as_bytes())
     }
 
     /// Verify a signature produced by [Transcript::sign].
     pub fn verify<V: Verifier>(&self, v: &V, sig: &<V as Verifier>::Signature) -> bool {
+        // Note: We pass an empty namespace here, since the namespace may be included
+        // within the transcript summary already via `Self::new`.
         v.verify(b"", self.summarize().hash.as_bytes(), sig)
     }
 }


cryptography/src/transcript.rs
```diff
@@ -265,12 +265,12 @@ impl Transcript {
     /// - signing the operations that have been performed on the transcript,
     /// - or, equivalently, signing randomness or a summary extracted from the transcript.
     pub fn sign<S: Signer>(&self, s: &S) -> <S as Signer>::Signature {
-        s.sign(None, self.summarize().hash.as_bytes())
+        s.sign(b"", self.summarize().hash.as_bytes())
     }
 
     /// Verify a signature produced by [Transcript::sign].
     pub fn verify<V: Verifier>(&self, v: &V, sig: &<V as Verifier>::Signature) -> bool {
```
patrick-ogrady, line 274: I think we should accept namespace as an arg here?

p2p/src/authenticated/discovery/actors/tracker/record.rs
```diff
@@ -313,7 +313,7 @@ mod tests {
         S: PrivateKeyExt,
     {
         let signer = S::from_seed(signer_seed);
-        let signature = signer.sign(None, &(socket, timestamp).encode());
+        let signature = signer.sign(b"", &(socket, timestamp).encode());
```
patrick-ogrady: nit: add a const for namespace in tests?

p2p/src/authenticated/discovery/types.rs
```diff
@@ -364,7 +364,7 @@ mod tests {
             socket: SocketAddr::from(([127, 0, 0, 1], 8080)),
             timestamp: 1234567890,
             public_key: c.public_key(),
-            signature: c.sign(None, &[1, 2, 3, 4, 5]),
+            signature: c.sign(b"", &[1, 2, 3, 4, 5]),
```
patrick-ogrady: add a const for namespace in tests?

cryptography/src/secp256r1/scheme.rs
```diff
@@ -452,7 +465,7 @@ mod tests {
     #[test]
     fn test_codec_signature() {
         let private_key = create_private_key();
-        let original = private_key.sign(None, "Hello World".as_bytes());
+        let original = private_key.sign_inner(None, "Hello World".as_bytes());
```
patrick-ogrady: for anything other than test vectors, I think we should use `sign`?

cryptography/src/bls12381/scheme.rs
patrick-ogrady, line 1: Should we also require namespace in `primitives/ops.rs`? I suppose it is more low-level and you should know what you are doing but 🤷

clabby: I figure it's nice to keep the namespace optional there, since we use that mod against all of the test vectors. Also nice to keep the core impl flexible for other usecases; cw's BLS impl is used by a few other teams nowadays.

cryptography/src/transcript.rs
```diff
@@ -265,12 +265,12 @@ impl Transcript {
...
-        v.verify(None, self.summarize().hash.as_bytes(), sig)
+        v.verify(b"", self.summarize().hash.as_bytes(), sig)
```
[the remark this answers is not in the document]
clabby: The transcript accepts a namespace upon construction.

cryptography/src/transcript.rs
clabby: I was confused about this at first too, `Transcript` accepts a namespace upon construction. It's already a part of the message.

https://github.com/commonwarexyz/monorepo/blob/5da33d2ebe54ce8ac1960f0b8b131d6fbbd216ce/cryptography/src/transcript.rs#L137-L151

cryptography/src/secp256r1/scheme.rs
```diff
@@ -465,7 +468,7 @@ mod tests {
     #[test]
     fn test_codec_signature() {
         let private_key = create_private_key();
-        let original = private_key.sign_inner(None, "Hello World".as_bytes());
+        let original = private_key.sign(b"test-namespace", "Hello World".as_bytes());
```
patrick-ogrady: nit: use `NAMESPACE`?

cryptography/src/transcript.rs
patrick-ogrady: Can you leave a comment inline indicating why this is ok (for both sing/verify)?

patrick-ogrady: Other dumb question (can be addressed in a different PR): would it be better to just use `transcript` for all of the message formatting (rather than `union_unique`)?

IIUC many of the crypto schemes don't want you to pre-hash the message (use a hash-to-curve function) but I could be being too pedantic here (cc @cronokirby)?

cronokirby: The only impact of an extra layer of hashing is a performance cost. The issues come from *not* hashing something that should be hashed, allowing carefully chosen messages to impact something deeper in the algorithm.

cronokirby: With a bit of elbow grease on our own side, we could also use transcript for the hash to curve directly. The hash to curve method really just needs a source of randomness, and so rather producing a summary with `Transcript`, and then passing to that the hard-baked XOF which hashes that summary to produce random bytes for the hash to curve method, we could use Transcript's XOF itself, to avoid a layer of hashing.

But, the overhead of an extra bit of hashing really shouldn't be that much.

Perhaps what this looks like is adding a method to the signing trait which signs a transcript directly. The default impl would sign the summary, but a more optimized impl could take advantage of such optimizations.

cryptography/src/transcript.rs
```diff
@@ -265,12 +265,16 @@ impl Transcript {
...
+        // Note: We pass an empty namespace here, since the namespace may be included
+        // within the transcript summary already via `Trasncript::new`.
+        s.sign(b"", self.summarize().hash.as_bytes())
     }
 
     /// Verify a signature produced by [Transcript::sign].
     pub fn verify<V: Verifier>(&self, v: &V, sig: &<V as Verifier>::Signature) -> bool {
-        v.verify(None, self.summarize().hash.as_bytes(), sig)
+        // Note: We pass an empty namespace here, since the namespace may be included
+        // within the transcript summary already via `Trasncript::new`.
```
cronokirby:
```suggestion
        // within the transcript summary already via `Self::new`.
```

cryptography/src/transcript.rs
patrick-ogrady: We'll need to make a call here before we consider this "stable". Can you add to your backlog to consider?

For now, don't think this should block anything.

patrick-ogrady (approved):

PR #2366, opened 2025-12-02
[consensus/simplex] Remove Force Notarize

patrick-ogrady:


consensus/src/simplex/actors/batcher/round.rs
```diff
@@ -57,15 +56,8 @@ impl<
         R: Reporter<Activity = Activity<S, D>>,
     > Round<P, S, B, D, R>
 {
-    pub fn new(participants: Ordered<P>, scheme: S, blocker: B, reporter: R, batch: bool) -> Self {
-        // Configure quorum params
-        let quorum = if batch {
```
patrick-ogrady, line 62: Now that we know participants during init, we don't need to do this `initialized` check.

consensus/src/simplex/actors/batcher/verifier.rs
```diff
@@ -30,8 +30,6 @@ pub struct Verifier<S: Scheme, D: Digest> {
 
     /// Pending notarize votes waiting to be verified.
     notarizes: Vec<Notarize<S, D>>,
-    /// Forces notarize verification as soon as possible (set when the leader proposal is known).
-    notarizes_force: bool,
```
patrick-ogrady, line 34: Given we already sent a proposal before verification (a leader could be wrong but as long as we had any other vote for the proposal would send), we might as well be smarter about it.

Issue #2365, opened 2025-12-02
[consensus/marshal] Add Test with Finalizations at Different Views for Same Height


patrick-ogrady:
At the end of an epoch (due to asynchrony), it is possible that there could exist multiple finalizations (in different views) for the same block. We should ensure that `marshal` has a test case explicitly handling this (where different validators may be storing finalizations from different views).

One "sanity check" is to make sure this doesn't cause `archive` to panic (for writing different certificates at the same key). This "should" be ok would be good to memorialize with a test.

PR #2369, opened 2025-12-02
[consensus/marshal] Duplicate finalization test

clabby:
## Overview

Adds a test that demonstrates marshal's ability to handle duplicate finalizations for the same block at different views.

Also updates the documentation of `Certificates` and `Blocks` to clarify that implementations must ignore overwrites of
finalizations or blocks at the same height (It should be fine in practice iiuc, but there's no need - seems practical to document that the behavior should match that of `{immutable/prunable}::Archive`.)

closes https://github.com/commonwarexyz/monorepo/issues/2365

### DIFF
--- a/consensus/src/marshal/mod.rs
+++ b/consensus/src/marshal/mod.rs
@@ -29,6 +29,10 @@
 //! to a potential block in the chain. The actor will only finalize a block (and its ancestors)
 //! if it has a corresponding finalization from consensus.
 //!
+//! _It is possible that there may exist multiple finalizations for the same block in different views. Marshal
+//! only concerns itself with verifying a valid finalization exists for a block, not that a specific finalization
+//! exists. This means different Marshals may have different finalizations for the same block persisted to disk._
+//!
 //! ## Backfill
 //!
 //! The actor provides a backfill mechanism for missing blocks. If the actor notices a gap in its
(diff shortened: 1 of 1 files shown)


consensus/src/marshal/store.rs
```diff
@@ -25,8 +25,9 @@ pub trait Certificates: Send + Sync + 'static {
 
     /// Store a finalization certificate, keyed by height and commitment.
     ///
-    /// Implementations must durably sync the write before returning; successful completion
-    /// implies that the certificate is persisted.
+    /// Implementations must:
+    /// - Durably sync the write before returning; successful completion implies that the certificate is persisted.
+    /// - Disallow overwriting an existing finalization at the same height or commitment.
```
patrick-ogrady: I think we may want to use a different term than "disallow" here. Archive will just ignore duplicate writes (but won't error) -> AFAICT marshal doesn't do any checking of its own to prevent calling Archive?

clabby: Updated the comment. Correct that marshal doesn't do any checking of its own; That happens here:

https://github.com/commonwarexyz/monorepo/blob/119e5f6e7541d1addddcc13ef08a7e17506cda33/storage/src/archive/immutable/storage.rs#L235-L238

& here:
https://github.com/commonwarexyz/monorepo/blob/119e5f6e7541d1addddcc13ef08a7e17506cda33/storage/src/archive/prunable/storage.rs#L301-L304

clabby: (That said, I don't think it'd actually be harmful to allow overwrites here in another implementation of `Certificates`. All that'd happen is you'd have a later finalization for the same view, which would still represent the same chain.)

patrick-ogrady: I think that's correct.

patrick-ogrady: (Actually, that "must" be correct because "ignore" implies, at least in distributed systems, that someone else could've seen the other thing first)

consensus/src/marshal/store.rs
```diff
@@ -25,8 +25,9 @@ pub trait Certificates: Send + Sync + 'static {
 
     /// Store a finalization certificate, keyed by height and commitment.
     ///
-    /// Implementations must durably sync the write before returning; successful completion
-    /// implies that the certificate is persisted.
+    /// Implementations must:
+    /// - Durably sync the write before returning; successful completion implies that the certificate is persisted.
+    /// - Ignore overwrites for an existing finalization at the same height or commitment.
```
patrick-ogrady: nit: prefer `Handle overwrites for an existing finalization or block at the same height or commitment.`?

PR #2372, opened 2025-12-03
fix spelling: "extendible" → "extendable"

maradini77:


storage/src/freezer/mod.rs
```diff
@@ -10,7 +10,7 @@
 //!
 //! # Format
 //!
-//! The [Freezer] uses a two-level architecture: an extendible hash table (written in a single [commonware_runtime::Blob])
+//! The [Freezer] uses a two-level architecture: an extendable hash table (written in a single [commonware_runtime::Blob])
```
patrick-ogrady, line 13: This is the correct naming: https://www.youtube.com/watch?v=r4GkXtH1la8

PR #2375, opened 2025-12-03
[consensus/simplex] Refine Skip Policy

patrick-ogrady:
Related: #2351

### DIFF
--- a/consensus/src/simplex/actors/batcher/actor.rs
+++ b/consensus/src/simplex/actors/batcher/actor.rs
@@ -42,7 +42,7 @@ pub struct Actor<
     reporter: R,
 
     activity_timeout: ViewDelta,
-    skip_timeout: ViewDelta,
+    skip_timeout: usize,
     epoch: Epoch,
     namespace: Vec<u8>,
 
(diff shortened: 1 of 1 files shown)


consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -194,16 +194,15 @@ impl<
...
+                                work.len() < self.skip_timeout.get() as usize
+                                // Leader active in at least one recent round
                                 || {
-                                    // Not enough recent history to judge (assume active)
                                     let min_view = current.saturating_sub(self.skip_timeout);
                                     let recent: Vec<_> = work.range(min_view..).collect();
-                                    (recent.len() as u64) < self.skip_timeout.get()
-                                    // Leader was active in at least one recent round
-                                    || recent.iter().any(|(_, round)| round.is_active(leader))
+                                    recent.iter().any(|(_, round)| round.is_active(leader))
```
BrendanChou: I think just do `work.iter().rev().take(self.skip_timeout).any(|(_, round)| round.is_active(leader))`

Don't enforce a `min_view` related to `skip_timeout` as the minimum key in `work` is already `current - self.activity_timeout`. So since `activity_timeout >= skip_timeout`, this allows for up to `n` gaps where `n = activity_timeout - skip_timeout`.

patrick-ogrady: Basically, this becomes "not active in the last `skip_timeout` views we know about" rather than the "the last `skip_timeout` views"....feels reasonable.

PR #2378, opened 2025-12-03
[p2p/lookup] Track all IPs for each peer

### LINKED ISSUE #2352: [p2p/lookup] Listen on All IPs for a Peer
When a peer updates its IP, we will only accept connections from the latest IP rather than all registered across tracked peer sets: https://github.com/commonwarexyz/monorepo/blob/936dc58f7d5414201eb95d41c97cfce2d3b246c0/p2p/src/authenticated/lookup/actors/tracker/directory.rs#L232-L241

We should consider allowing a peer to dial us from any of their tracked IPs (can help with complex end-of-epoch procedures where a peer waits to update their IP until they've exited the epoch ... and generally seems more "correct").

This is not an issue in `discovery` because we don't filter by IP.

RonTuretzky:
## Summary

- Update the tracker to maintain a set of all IP addresses a peer has been associated with
- Allow incoming connections from any IP a peer has used while in an active peer set
- Add comprehensive unit and integration tests

This addresses the issue where a peer updating its IP would only allow connections from the latest IP, rather than all IPs registered across tracked peer sets. This helps with complex end-of-epoch procedures where a peer may wait to update their IP until they've exited the epoch.

## Changes

- Add `tracked_ips: HashSet<IpAddr>` field to `Record` struct
- Update `Record::update()` to accumulate IPs in the set and return `bool` indicating if a new IP was added
- Clear tracked IPs when a peer is blocked via `Record::block()`
- Add `Record::ips()` method to iterate over tracked IPs with optional private IP filtering
- Change `Directory::registered()` to return all tracked IPs instead of just the current socket's IP
- Add info logging when a new IP is added for a peer

## Test plan

- [x] Added unit tests for `Record` IP tracking behavior (`test_tracked_ips_*`)
- [x] Added integration tests for `Directory` (`test_registered_returns_all_tracked_ips`, `test_registered_clears_ips_on_peer_removal`, `test_registered_clears_ips_on_block`)
- [ ] Run `just test -p commonware-p2p`
- [ ] Run `just lint`

Closes #2352

### DIFF
--- a/p2p/src/authenticated/lookup/actors/tracker/directory.rs
+++ b/p2p/src/authenticated/lookup/actors/tracker/directory.rs
@@ -129,7 +129,7 @@ impl<E: Spawner + Rng + Clock + GClock + RuntimeMetrics, C: PublicKey> Directory
             let record = match self.peers.entry(peer.clone()) {
                 Entry::Occupied(entry) => {
                     let entry = entry.into_mut();
-                    let new_ip = entry.update(*addr);
+                    let _new_ip = entry.update(*addr);
                     entry
                 }
                 Entry::Vacant(entry) => {


p2p/src/authenticated/lookup/actors/tracker/directory.rs
```diff
@@ -229,13 +232,16 @@ impl<E: Spawner + Rng + Clock + GClock + RuntimeMetrics, C: PublicKey> Directory
...
+    /// Returns all tracked IPs for each allowed peer, enabling connections
+    /// from any IP a peer has used while in an active peer set.
     pub fn registered(&self) -> HashSet<IpAddr> {
         // Using `.allowed()` here excludes any peers that are still connected but no longer
         // part of a peer set (and will be dropped shortly).
         self.peers
             .values()
             .filter(|r| r.allowed(self.allow_private_ips))
-            .filter_map(|r| r.socket().map(|s| s.ip()))
+            .flat_map(|r| r.ips(self.allow_private_ips))
```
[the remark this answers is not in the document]
RonTuretzky: [901c263](https://github.com/commonwarexyz/monorepo/pull/2378/commits/901c2638725e99cb0c212c5098b3e87ea323e3f0)

p2p/src/authenticated/lookup/actors/tracker/directory.rs
```diff
@@ -130,7 +130,10 @@ impl<E: Spawner + Rng + Clock + GClock + RuntimeMetrics, C: PublicKey> Directory
             let record = match self.peers.entry(peer.clone()) {
                 Entry::Occupied(entry) => {
                     let entry = entry.into_mut();
-                    entry.update(*addr);
+                    let new_ip = entry.update(*addr);
+                    if new_ip {
+                        info!(?peer, ip = ?addr.ip(), "added new IP for peer");
```
patrick-ogrady: We don't need to log here 👍

p2p/src/authenticated/lookup/actors/tracker/record.rs
```diff
@@ -45,51 +48,61 @@ pub struct Record {
 
     /// If `true`, the record should persist even if the peer is not part of any peer sets.
     persistent: bool,
+
+    /// All IP addresses this peer has been associated with.
+    /// Used for accepting incoming connections from any known IP.
+    tracked_ips: HashSet<IpAddr>,
```
patrick-ogrady: need to use BTreeSet for determinsim

p2p/src/authenticated/lookup/actors/tracker/directory.rs
RonTuretzky: addressed with latest changes

p2p/src/authenticated/lookup/actors/tracker/record.rs
RonTuretzky: addressed with latest changes

p2p/src/authenticated/lookup/actors/tracker/directory.rs
```diff
@@ -129,7 +129,7 @@ impl<E: Spawner + Rng + Clock + GClock + RuntimeMetrics, C: PublicKey> Directory
             let record = match self.peers.entry(peer.clone()) {
                 Entry::Occupied(entry) => {
                     let entry = entry.into_mut();
-                    entry.update(*addr);
+                    let new_ip = entry.update(*addr);
```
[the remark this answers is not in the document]
RonTuretzky: addressed with a prefix

Issue #2337, opened 2025-11-26
[macros] Add a variant of `select_loop!` + `select!` with a shutdown handler


clabby:
## Overview

Throughout the codebase, we use `select!` + `select_loop!` in actors, which should all handle their own shutdown signals. Until #2133 lands, it would be nice to _require_ handling graceful shutdown, rather than leaving the option open to exclude a shutdown handler.

We should add new variants of these macros that take in a `Signal` future and require adding a branch to handle its resolution.

clabby:
Note that #2133 forces the shutdown signal into every actor's control loop and allows for (optional) auxiliary shutdown handling logic. That said, the migration to that (unfinished) API may take a while - this'd be a great intermediate band-aid.

PR #2383, opened 2025-12-03
[macros] Require shutdown handler in `select_loop!`

clabby:
## Overview

Changes `select_loop!` to require handling a shutdown signal received by `Spawner::stopped`.

closes https://github.com/commonwarexyz/monorepo/issues/2337

### DIFF
--- a/examples/sync/src/bin/server.rs
+++ b/examples/sync/src/bin/server.rs
@@ -296,7 +296,7 @@ where
         mpsc::channel::<wire::Message<DB::Operation, Key>>(RESPONSE_BUFFER_SIZE);
     select_loop! {
         context,
-        on_shutdown => {
+        on_stopped => {
             debug!("context shutdown, closing client connection");
         },
         incoming = recv_frame(&mut stream, MAX_MESSAGE_SIZE) => {
@@ -422,7 +422,7 @@ where
     let mut next_op_time = context.current() + config.op_interval;
     select_loop! {
         context,
-        on_shutdown => {
+        on_stopped => {
             debug!("context shutdown, stopping server");
         },
         _ = context.sleep_until(next_op_time) => {
--- a/macros/src/lib.rs
+++ b/macros/src/lib.rs
@@ -435,7 +435,7 @@ pub fn select(input: TokenStream) -> TokenStream {
 
 /// Input for [select_loop].
 ///
-/// Parses: `context, on_shutdown => { block }, { branches... }`
+/// Parses: `context, on_stopped => { block }, { branches... }`
 struct SelectLoopInput {
     context: Expr,
     shutdown_block: Block,
@@ -448,12 +448,12 @@ impl Parse for SelectLoopInput {
         let context: Expr = input.parse()?;
         input.parse::<Token![,]>()?;
 
-        // Parse `on_shutdown =>`
-        let on_shutdown_ident: Ident = input.parse()?;
-        if on_shutdown_ident != "on_shutdown" {
+        // Parse `on_stopped =>`
+        let on_stopped_ident: Ident = input.parse()?;
+        if on_stopped_ident != "on_stopped" {
             return Err(Error::new(
-                on_shutdown_ident.span(),
-                "expected `on_shutdown` keyword",
+                on_stopped_ident.span(),
+                "expected `on_stopped` keyword",
             ));
         }
         input.parse::<Token![=>]>()?;
@@ -510,7 +510,7 @@ impl Parse for SelectLoopInput {
 /// ```rust,ignore
 /// select_loop! {
 ///     context,
-///     on_shutdown => { cleanup },
+///     on_stopped => { cleanup },
 ///     pattern = future => block,
 ///     // ...
 /// }
@@ -527,7 +527,7 @@ impl Parse for SelectLoopInput {
 /// async fn run(context: impl commonware_runtime::Spawner) {
 ///     select_loop! {
 ///         context,
-///         on_shutdown => {
+///         on_stopped => {
 ///             println!("shutting down");
 ///             drop(shutdown);
 ///         },
--- a/macros/tests/select.rs
+++ b/macros/tests/select.rs
@@ -36,6 +36,10 @@ mod tests {
         });
     }
 
+    /// A mock signaler that never resolves.
+    ///
+    /// Used in place of a proper `Signal` from `commonware_runtime` to avoid cyclical
+    /// dependencies.
     struct MockSignaler;
 
     impl MockSignaler {
@@ -44,6 +48,10 @@ mod tests {
         }
     }
 
+    /// A mock signaler that resolves immediately.
+    ///
+    /// Used in place of a proper `Signal` from `commonware_runtime` to avoid cyclical
+    /// dependencies.
     struct MockSignalerResolves;
 
     impl MockSignalerResolves {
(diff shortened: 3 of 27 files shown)


macros/src/lib.rs
```diff
@@ -433,68 +433,144 @@ pub fn select(input: TokenStream) -> TokenStream {
...
 /// use commonware_macros::select_loop;
-/// use futures::executor::block_on;
-/// use futures_timer::Delay;
 ///
-/// async fn task() -> usize {
-///     42
-/// }
-//
-/// block_on(async move {
+/// async fn run(context: impl commonware_runtime::Spawner) {
```
patrick-ogrady, line 527: This will fail to publish I think (runtime released after)?

macros/tests/select.rs
```diff
@@ -34,89 +34,4 @@ mod tests {
             assert_eq!(completed, vec![3, 2, 1]);
         });
     }
-
-    #[test]
-    fn test_select_loop_basic() {
```
patrick-ogrady, line 39: Suggest implementing a mock for stopped here (should avoid commonware-runtime because of publish order)?

macros/src/lib.rs
clabby: This doc comment is ```rust,ignore```; it won't run 👍

patrick-ogrady: Ah, gotcha

examples/sync/src/bin/server.rs
```diff
@@ -295,6 +295,10 @@ where
     let (response_sender, mut response_receiver) =
         mpsc::channel::<wire::Message<DB::Operation, Key>>(RESPONSE_BUFFER_SIZE);
     select_loop! {
+        context,
+        on_shutdown => {
```
patrick-ogrady: Technically, we should call this on_stopped I think (signal is on "stopped")?

examples/sync/src/bin/server.rs
```diff
@@ -336,14 +340,16 @@ where
...
                 }
             } else {
                 // Channel closed
-                break Ok(());
+                return Ok(());
             }
         }
     }
+
+    Ok(())
```
patrick-ogrady, line 352: It seems like we only return `Ok(())`?

macros/src/lib.rs
```diff
@@ -433,68 +433,144 @@ pub fn select(input: TokenStream) -> TokenStream {
...
+}
+
+impl Parse for SelectLoopInput {
+    fn parse(input: ParseStream<'_>) -> Result<Self> {
+        // Parse context expression
+        let context: Expr = input.parse()?;
+        input.parse::<Token![,]>()?;
+
+        // Parse `on_shutdown =>`
+        let on_shutdown_ident: Ident = input.parse()?;
```
patrick-ogrady: I suppose we could alternatively assume an application has no "shutdown" logic if not populated (I do like enforcing you provide something tho...)

macros/src/lib.rs
patrick-ogrady: Dumb question: does `ignore` at least enforce proper syntax? i.e. if the code here changes, will the example get out of date?

macros/src/lib.rs
```diff
@@ -433,68 +433,144 @@ pub fn select(input: TokenStream) -> TokenStream {
...
-            commonware_macros::select! {
-                #input
+        {
+            let mut shutdown = #context.stopped();
+            loop {
+                commonware_macros::select! {
+                    _ = &mut shutdown => {
+                        #shutdown_block
+
+                        // Break the loop after handling shutdown. Some implementations
```
patrick-ogrady, line 567: :heart:

macros/tests/select.rs
```diff
@@ -35,6 +36,22 @@ mod tests {
         });
     }
 
+    struct MockSignaler;
```
patrick-ogrady, line 43: Should we leave a comment pointing to runtime's signaler? I suppose that's probably overkill.

p2p/src/authenticated/discovery/actors/listener.rs
```diff
@@ -158,12 +158,11 @@ impl<E: Spawner + Clock + ReasonablyRealtime + Network + Rng + CryptoRng + Metri
...
 
         // Loop over incoming connections as fast as our rate limiter allows
-        let mut shutdown = self.context.stopped();
         let mut accepted = 0;
         select_loop! {
-            _ = &mut shutdown => {
+            self.context,
+            on_shutdown => {
                 debug!("context shutdown, stopping listener");
-                break;
```
patrick-ogrady, line 166: I suppose it is "obvious" you don't need to break here anymore (although doing so won't cause any issues)?

p2p/src/authenticated/lookup/actors/listener.rs
```diff
@@ -180,11 +180,10 @@ impl<E: Spawner + Clock + ReasonablyRealtime + Network + Rng + CryptoRng + Metri
...
-        let mut shutdown = self.context.stopped();
         select_loop! {
-            _ = &mut shutdown => {
+            self.context,
+            on_shutdown => {
                 debug!("context shutdown, stopping listener");
-                break;
             },
             update = self.mailbox.next() => {
                 let Some(registered_ips) = update else {
```
patrick-ogrady, line 189: May be cool to auto-handle this sort of unwrapping for certain channels (exiting if closing). Won't apply to all things (like accept below) but would suspect generally useful.

runtime/src/utils/signal.rs
```diff
@@ -62,13 +62,15 @@ use std::{
...
 ///     context.with_label("waiter").spawn(|context| async move {
 ///         // Wait for signal or sleep
-///         select_loop! {
-///             sig = &mut signal => {
-///                 println!("Received signal: {}", sig.unwrap());
-///                 break;
-///             },
-///             _ = context.sleep(Duration::from_secs(1)) => {},
-///         };
+///         loop {
```
patrick-ogrady, line 65: Somewhat confusing we don't just use runtime here (as it incorporates signaler directly). I suppose this, however, is the best way to demonstrate?

macros/src/lib.rs
clabby: It does not unfortunately, since it doesn't build the code block.

examples/sync/src/bin/server.rs
clabby: The loop always returns `()` due to the implicit `break` that is rendered in the macro. We need to add an `Ok(())` after the fact to return after the loop has been exited.

macros/src/lib.rs
clabby: We should fairly heavy-handidly require that an application handles graceful shutdown if they use our actor helpers, IMO

macros/tests/select.rs
clabby: Did anyways, can't hurt.

p2p/src/authenticated/discovery/actors/listener.rs
clabby: Yeah, doing so won't cause any issues, but it is handled implicitly. The reason why we do it in the macro is because it's actually a bug that will panic if you don't exit the loop right away, since the shared future ends up being polled twice.

p2p/src/authenticated/lookup/actors/listener.rs
clabby: Agree. I think that'd be a nice follow-up as we expand this.

runtime/src/utils/signal.rs
clabby: Yeah I wanted to avoid using the runtime context to showcase the primitive itself. We could use `context.stopped()`, but that's not the only usecase for `Signal` 🤷

PR #2384, opened 2025-12-03
[Storage] avoid extra copy in create_proof_store

GarmashAlex:
Replace encode().to_vec() with encode() when preparing elements for ProofStore::new. ProofStore::new only needs AsRef<[u8]>, and Encode::encode() returns BytesMut which qualifies. This eliminates unnecessary allocations and copies per operation and aligns with how verify_proof and verify_proof_and_extract_digests already pass elements.


patrick-ogrady (approved):
Nice!

PR #2385, opened 2025-12-04
[storage/adb/any/unordered] optimized batch update for adb::any::unordered

roberto-bayardo:
This is the first specialized batch-update implementation, limited to the adb::any::unordered (fixed + variable) variants.  The implementation gathers a set of locations to look up concurrently via `try_join`, then processes those together to determine the precise operations to append to the log.

Speedup on the generation benchmarks is ~10-40% depending on # of operations in a batch (larger batches result in greater speedup).

Fixed non-determinism issue with batching, and changed all batching tests to verify determinism across 2 runs.

Related work towards: https://github.com/commonwarexyz/monorepo/issues/2223

Closes: https://github.com/commonwarexyz/monorepo/issues/2390

### DIFF
--- a/storage/src/adb/mod.rs
+++ b/storage/src/adb/mod.rs
@@ -268,6 +268,10 @@ fn update_known_loc<I: Index<Value = Location>>(
 
 /// For the given `key` which is known to exist in the snapshot with location `old_loc`, delete
 /// it from the snapshot.
+///
+/// # Panics
+///
+/// Panics if `key` is not found in the snapshot or if `old_loc` is not found in the cursor.
 fn delete_known_loc<I: Index<Value = Location>>(snapshot: &mut I, key: &[u8], old_loc: Location) {
     let mut cursor = snapshot.get_mut(key).expect("key should be known to exist");
     assert!(
--- a/storage/src/adb/store/mod.rs
+++ b/storage/src/adb/store/mod.rs
@@ -1119,13 +1119,8 @@ mod test {
 
     #[test_traced("DEBUG")]
     fn test_batch() {
-        // Run the batch tests twice and check for determinism.
-        let state1 = batch_tests::test_batch(|ctx| async move {
-            create_test_store(ctx.with_label("batch")).await
-        });
-        let state2 = batch_tests::test_batch(|ctx| async move {
-            create_test_store(ctx.with_label("batch")).await
-        });
-        assert_eq!(state1, state2);
+        batch_tests::test_batch(
+            |ctx| async move { create_test_store(ctx.with_label("batch")).await },
+        );
     }
 }
--- a/storage/src/adb/any/ordered/fixed.rs
+++ b/storage/src/adb/any/ordered/fixed.rs
@@ -995,9 +995,6 @@ mod test {
 
     #[test_traced("DEBUG")]
     fn test_batch() {
-        // Run the batch tests twice and check for determinism.
-        let state1 = batch_tests::test_batch(|ctx| async move { create_test_db(ctx).await });
-        let state2 = batch_tests::test_batch(|ctx| async move { create_test_db(ctx).await });
-        assert_eq!(state1, state2);
+        batch_tests::test_batch(|ctx| async move { create_test_db(ctx).await });
     }
 }
--- a/storage/src/adb/any/unordered/fixed.rs
+++ b/storage/src/adb/any/unordered/fixed.rs
@@ -794,9 +794,6 @@ pub(super) mod test {
 
     #[test_traced("DEBUG")]
     fn test_batch() {
-        // Run the batch tests twice and check for determinism.
-        let state1 = batch_tests::test_batch(|ctx| async move { create_test_db(ctx).await });
-        let state2 = batch_tests::test_batch(|ctx| async move { create_test_db(ctx).await });
-        assert_eq!(state1, state2);
+        batch_tests::test_batch(|ctx| async move { create_test_db(ctx).await });
     }
 }
--- a/storage/src/adb/any/unordered/variable.rs
+++ b/storage/src/adb/any/unordered/variable.rs
@@ -613,17 +613,10 @@ pub(super) mod test {
 
     #[test_traced("DEBUG")]
     fn test_batch() {
-        // Run the batch tests twice and check for determinism.
-        let state1 = batch_tests::test_batch(|mut ctx| async move {
+        batch_tests::test_batch(|mut ctx| async move {
             let seed = ctx.next_u64();
             let cfg = db_config(&format!("batch_{seed}"));
             AnyTest::init(ctx, cfg).await.unwrap()
         });
-        let state2 = batch_tests::test_batch(|mut ctx| async move {
-            let seed = ctx.next_u64();
-            let cfg = db_config(&format!("batch_{seed}"));
-            AnyTest::init(ctx, cfg).await.unwrap()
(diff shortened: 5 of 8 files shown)


storage/src/adb/any/unordered/mod.rs
```diff
@@ -118,7 +123,7 @@ impl<
     /// Returns the active operation for `key` with its location, or None if the key is not active.
     pub(crate) async fn get_key_op_loc(
         &self,
-        key: &<C::Item as Keyed>::Key,
+        key: &Key<C::Item>,
```
patrick-ogrady, line 127: ❤️

storage/src/adb/any/unordered/mod.rs
```diff
@@ -566,6 +585,62 @@ impl<
...
+                    .append(C::Item::new_update(key.clone(), value))
+                    .await?;
+            } else {
+                self.delete_known_loc(key, old_loc);
+                self.log.append(C::Item::new_delete(key.clone())).await?;
+            }
+        }
+
+        // Process the creates.
+        for (key, value) in updates {
```
patrick-ogrady, line 781: If we are iterating over `updates`, we should be using `BTreeMap`?

storage/src/adb/any/unordered/mod.rs
```diff
@@ -566,6 +585,62 @@ impl<
...
+        let mut locations = Vec::new();
+        for (key, value) in iter {
+            let iter = self.snapshot.get(&key);
+            locations.extend(iter.copied());
+            updates.insert(key, value);
+        }
+
+        // Concurrently look up all possible matching locations.
+        locations.sort();
+        let futures = locations.iter().map(|loc| self.log.read(*loc));
```
patrick-ogrady, line 756: nit: dedup `loc`?

storage/src/adb/any/unordered/mod.rs
roberto-bayardo: Since the order doesn't matter, I don't think it would help?  Or are you thinking to make it more unsafe friendly?

storage/src/adb/any/unordered/mod.rs
roberto-bayardo: done

storage/src/adb/any/unordered/mod.rs
```diff
@@ -566,6 +556,63 @@ impl<
...
+        locations.sort();
+        locations.dedup();
+        let futures = locations.iter().map(|loc| self.log.read(*loc));
+        let results = try_join_all(futures).await?;
+
+        // Process the deletes & updates of existing keys, which must appear in the results.
+        for (op, old_loc) in (results.into_iter()).zip(locations) {
+            let key = op.key().expect("updates should have a key");
+            let Some(update) = updates.remove(key) else {
+                continue; // translated key collision
```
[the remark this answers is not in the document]
roberto-bayardo: that's too much detail.

storage/src/adb/any/unordered/mod.rs
```diff
@@ -566,6 +556,63 @@ impl<
...
+            } else {
+                delete_known_loc(&mut self.snapshot, key, old_loc);
+                self.log.append(C::Item::new_delete(key.clone())).await?;
+            }
+        }
+
+        // Process the creates.
+        for (key, value) in updates {
+            let Some(value) = value else {
+                continue; // attempt to delete a non-existent key
```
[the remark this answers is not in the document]
roberto-bayardo: deleting potentially non-existent keys is going to be expected from delete_unchecked usage.

storage/src/adb/any/unordered/mod.rs
```diff
@@ -566,6 +556,63 @@ impl<
...
+        &mut self,
+        iter: impl Iterator<Item = (Key<C::Item>, Option<Value<C::Item>>)>,
+    ) -> Result<(), Error> {
+        let mut updates = HashMap::new();
+        let mut locations = Vec::new();
+        for (key, value) in iter {
+            let iter = self.snapshot.get(&key);
+            locations.extend(iter.copied());
+            updates.insert(key, value);
+        }
```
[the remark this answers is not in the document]
roberto-bayardo: changed to use size_hint.

storage/src/adb/any/unordered/mod.rs
patrick-ogrady: We call `log.append()` in a different order with HashMap (which should break the runtime's determinism check if used in any chain).

We should add a test to ensure determinism is upheld for this operation here as well?

storage/src/adb/any/unordered/mod.rs
```diff
@@ -566,6 +556,63 @@ impl<
...
+        for (key, value) in updates {
+            let Some(value) = value else {
+                continue; // attempt to delete a non-existent key
+            };
+            self.snapshot.insert(&key, self.op_count());
+            self.log.append(C::Item::new_update(key, value)).await?;
+        }
+
+        Ok(())
+    }
```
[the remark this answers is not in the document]
roberto-bayardo: good bot. fixed.

storage/src/adb/any/unordered/mod.rs
patrick-ogrady: (I suppose this would also lead to a root mismatch on different nodes)

danlaine (dismissed):

storage/src/adb/any/unordered/mod.rs
roberto-bayardo: Changed to BTreeMap

storage/src/adb/store/mod.rs
```diff
@@ -1110,14 +1119,13 @@ mod test {
...
-        let executor = deterministic::Runner::default();
-        executor.start(|context| async move {
-            batch_tests::run_batch_tests(|| {
-                let ctx = context.clone();
-                async move { create_test_store(ctx.with_label("batch")).await }
-            })
-            .await
-            .unwrap();
+        // Run the batch tests twice and check for determinism.
+        let state1 = batch_tests::test_batch(|ctx| async move {
```
danlaine: Should we update `test_batch` to factor out the logic for running the test twice and comparing the states of both executions?

storage/src/adb/mod.rs
```diff
@@ -242,6 +246,37 @@ where
...
+    assert!(
+        cursor.find(|&loc| *loc == old_loc),
+        "prev_key with given old_loc should have been found"
+    );
+    cursor.update(new_loc);
+}
+
+/// For the given `key` which is known to exist in the snapshot with location `old_loc`, delete
+/// it from the snapshot.
+fn delete_known_loc<I: Index<Value = Location>>(snapshot: &mut I, key: &[u8], old_loc: Location) {
```
danlaine, line 275: For consistency with the new comments above
```suggestion
///
/// # Panics
///
/// Panics if `key` is not found in the snapshot or if `old_loc` is not found in the cursor.
fn delete_known_loc<I: Index<Value = Location>>(snapshot: &mut I, key: &[u8], old_loc: Location) {
```

storage/src/adb/store/mod.rs
roberto-bayardo: done

storage/src/adb/mod.rs
roberto-bayardo: done

PR #2388, opened 2025-12-04
Add support for separate ingress/egress sockets and DNS resolution

### LINKED ISSUE #1578: [p2p] Support Dialing Peers with DNS Entries
It would be nice to be able to dial peers identified by DNS entries rather than strictly IPs.

I'm not convinced it would be a good idea to gossip DNS entries in discovery (just the IP you connected on) but may ease the operational burden of running clusters of nodes in smaller/lookup-based networks.

### LINKED ISSUE #1742: [p2p] In `lookup/discovery` allow for different IPs for dialing and listening on a per-peer basis
Peers may have infra where they dial from a different IP than they listen on (from where they accept connections). We currently assume they are the same, but ideally we can instead make the type a tuple (pair) or an enum that is capable of describing a setup where the two IPs are different. This allows peers to not have their dial handshakes dropped by listeners if they aren't dialing from the same IP as their listener IP

### LINKED ISSUE #2033: [p2p] Unblock DNS Resolution
Some operators would prefer to advertise their dialable IP via DNS (rather than providing a specific IP). We should consider adding first-class support for domains in peer registry (and gossip) or paving a road to incorporate some sort of periodic refresh (external to p2p) that calls "register" every few minutes to refresh dialable IPs.

commonware-llm:
This commit introduces the ability for peers to register either a single
socket address or separate addresses for ingress and egress connections.
Additionally, DNS entries are now supported for ingress addresses.

Key changes:
- Add Resolver trait to runtime for DNS resolution
- Implement DNS resolver for deterministic runtime with configurable mappings
- Implement DNS resolver for tokio runtime using built-in DNS functionality
- Add Socket enum (Direct/Dns) and Address enum (Single/Split) to p2p
- Update Bootstrapper type to use Address instead of SocketAddr
- Update Directory::init to handle the new Address type

The new Address type allows:
- Single(SocketAddr): Same address for both ingress and egress (backward compatible)
- Split { ingress: Socket, egress: SocketAddr }: Different addresses for each direction

For DNS resolution:
- Deterministic runtime: Uses configurable dns_mappings in Config
- Tokio runtime: Uses tokio::net::lookup_host for real DNS resolution

Closes #2033, closes #1578, closes #1742

### DIFF
--- a/p2p/src/authenticated/discovery/actors/tracker/actor.rs
+++ b/p2p/src/authenticated/discovery/actors/tracker/actor.rs
@@ -509,8 +509,10 @@ mod tests {
         executor.start(|context| async move {
             let (_boot_signer, boot_pk) = new_signer_and_pk(99);
             let boot_addr = SocketAddr::new(Ipv4Addr::LOCALHOST.into(), 9999);
-            let cfg_with_boot =
-                default_test_config(PrivateKey::from_seed(0), vec![(boot_pk.clone(), boot_addr)]);
+            let cfg_with_boot = default_test_config(
+                PrivateKey::from_seed(0),
+                vec![(boot_pk.clone(), boot_addr.into())],
+            );
             let TestHarness {
                 mailbox: mut new_mailbox,
                 ..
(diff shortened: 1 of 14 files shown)


clabby:

PR #2391, opened 2025-12-04
[resolver/p2p] support targeted peer selection

andresilva:
This PR adds the ability to restrict resolver fetch requests to specific target peers, enabling callers to direct requests to peers they believe have the data rather than trying peers at random.

When consensus or other components learn that specific peers have certain data (through protocol messages, advertisements, or protocol-specific knowledge), they can now pass that information to the resolver. This can significantly reduce fetch latency by avoiding round-trips to peers that don't have the data.

The new `fetch_targeted` and `fetch_all_targeted` methods restrict fetches to specified peers with no automatic fallback. Targets persist through transient failures since the peer might be slow or receive the data later. Calling `fetch_targeted` on an in-progress fetch accumulates targets, while calling `fetch` clears them to allow fallback to any peer. Targets are automatically cleared on success, cancel, or when a peer is blocked for sending invalid data.

### DIFF
--- a/resolver/src/p2p/fetcher.rs
+++ b/resolver/src/p2p/fetcher.rs
@@ -129,7 +129,7 @@ impl<E: Clock + GClock + Rng + Metrics, P: PublicKey, Key: Span, NetS: Sender<Pu
 
         // Try pending keys until one succeeds or all participants are rate-limited
         let mut min_wait: Option<Duration> = None;
-        let mut selected: Option<(Key, P, ID)> = None;
+        let mut selected = None;
         for (key, (_, retry)) in self.pending.iter() {
             // Try to find a peer for the key
             let (result, is_targeted) = match self.targets.get(key) {


p2p/src/utils/requester/requester.rs
```diff
@@ -118,9 +118,43 @@ impl<E: Clock + GClock + Rng + Metrics, P: PublicKey> Requester<E, P> {
...
+    /// Only participants for which `filter` returns `true` will be considered.
+    /// This is useful when the caller knows which specific peers have the data.
+    ///
+    /// Returns `Ok((participant, id))` if a request can be made.
+    /// Returns `Err(duration)` if no participant is available:
+    /// - `Err(Duration::MAX)`: No eligible participants exist (none pass the filter, all blocked,
+    ///   excluded, or none registered)
+    /// - `Err(wait_time)`: Matching participants exist but are rate-limited, retry after
+    ///   `wait_time`
+    pub fn request_filtered(
```
andresilva, line 150: I opted to make these two separate methods rather than having `request` take `Option<impl Fn(&P) -> bool>` since when this is `None` the compiler can't infer the type, which would make the API very annoying to use.

andresilva:
~~I wrote an integration test for this but it's a bit useless since externally we cannot observe whether hints are actually used, and eventually we fall back to querying all peers anyway.~~ I'll add some metrics for hints which should make the integrations tests useful.

resolver/src/p2p/fetcher.rs
```diff
@@ -90,18 +105,45 @@ impl<E: Clock + GClock + Rng + Metrics, P: PublicKey, Key: Span, NetS: Sender<Pu
...
+    /// On send failure, the key is retried (and if the peer was hinted, it is removed).
     pub async fn fetch(&mut self, sender: &mut WrappedSender<NetS, wire::Message<Key>>) {
         // Reset waiter
         self.waiter = None;
 
-        // Get peer to send request to
-        let (_, retry) = self.peek_pending().unwrap();
-        let (peer, id) = match self.requester.request(retry) {
+        // Peek at the pending key to check for hints
+        let (key, (_, retry)) = self.pending.peek().unwrap();
```
clabby: With this no longer using `peek_pending`, think we can remove that fn; It's only used in `get_pending_deadline` now.

resolver/src/p2p/fetcher.rs
```diff
@@ -90,18 +105,45 @@ impl<E: Clock + GClock + Rng + Metrics, P: PublicKey, Key: Span, NetS: Sender<Pu
...
+
+        // Get peer to send request to, using hints if available
+        let result = match self.hints.get(key) {
+            Some(hints) => {
+                match self
+                    .requester
+                    .request_filtered(*retry, |p| hints.contains(p))
+                {
+                    Ok(selection) => Ok(selection),
+                    Err(wait) if wait == Duration::MAX => {
```
clabby: This is unrelated to your changes, but not a huge fan of how we use `Duration::MAX` as a magic number to specify a different error variant. Would be nice to update this API to return `Result<_, Option<Duration>>`

resolver/src/p2p/ingress.rs
```diff
@@ -88,3 +92,28 @@ impl<K: Span> Resolver for Mailbox<K> {
...
+    /// Multiple hints can be registered for the same key. Hints can be added
+    /// before or after calling `fetch()` - new hints will be used on retry.
+    ///
+    /// Hints are automatically cleared when:
+    /// - The fetch succeeds (data received)
+    /// - The fetch is canceled
+    /// - All hinted peers fail (empty hints trigger fallback to all peers)
+    ///
+    /// Panics if the send fails.
+    pub async fn hint(&mut self, key: K, peer: P) {
```
clabby: Should this be a separate operation than enqueueing the fetch? This seems error prone in the case where someone mistakenly hints following a fetch and that fetch is already resolved, for example - the hint would then sit in memory forever.

Instead maybe fetch should take an optional hint of peers that might have the data along with the list of keys, so that we can prevent that misuse?

clabby: i.e.

```rust
mailbox.fetch(vec![key_1]).await;
// ... fetch completes successfully, clear_hints(key_1) called ...
mailbox.hint(key_1, peer_a).await;  // Hint added AFTER completion
// This hint persists until next fetch/cancel/retain for key_1 (if that ever happens!)
```

clabby: (or, we could add a TTL / ring buffer for hints, though I like combining the hint w/ the fetch a bit more.)

andresilva: Good point. I implemented it this way as I was targeting the use case in #2287, i.e. we trigger a fetch when we see some peer in a future epoch, but we also keep adding other peers as hints as we see them. In #2287 I just called `resolver.clear()` once we got the finalization, which clears all the hints. But I agree that the API might be prone to misuse. I think being able to add hints after the fetch has started is useful, so maybe the API can be as follows:
- add `fetch_hinted` that takes a vec of peers as hints
- `hint` only adds the hint if there is any on-going request for that key, otherwise the hint is ignored

resolver/src/p2p/fetcher.rs
andresilva: I pondered about that as well, will wait for input from @patrick-ogrady before making that change.

resolver/src/p2p/ingress.rs
clabby: This suggestion sgtm 👍

resolver/src/p2p/fetcher.rs
andresilva: Created a proper error type for this.

resolver/src/p2p/ingress.rs
```diff
@@ -88,3 +101,70 @@ impl<K: Span> Resolver for Mailbox<K> {
...
+    /// or send failure), only that peer is removed from hints. As hints deplete through
+    /// failures, the resolver naturally falls back to trying all peers.
+    ///
+    /// Hints are automatically cleared when:
+    /// - The fetch succeeds (data received)
+    /// - The fetch is canceled
+    /// - All hinted peers fail (empty hints trigger fallback to all peers)
+    ///
+    /// Panics if the send fails.
+    pub async fn fetch_hinted(&mut self, key: K, hints: Vec<P>) {
```
andresilva: Wondering whether these methods should be added to the `Resolver` trait or not.

resolver/src/p2p/fetcher.rs
```diff
@@ -288,6 +405,11 @@ impl<E: Clock + GClock + Rng + Metrics, P: PublicKey, Key: Span, NetS: Sender<Pu
         self.requester.len_blocked()
     }
 
+    /// Returns the total number of hints (sum of all hints across all keys).
+    pub fn len_hints(&self) -> usize {
+        self.hints.values().map(|v| v.len()).sum()
```
andresilva: Considered adding a running counter but seemed like premature optimization.

resolver/src/p2p/ingress.rs
clabby: IMO, no; This seems fairly specific to the p2p backend. The exoware backend won't need this concept.

clabby (dismissed):
New API looks great 🙏

andresilva:
bugbot run

resolver/src/p2p/fetcher.rs
```diff
@@ -67,6 +75,13 @@ pub struct Fetcher<
     /// Whether requests are sent with priority over other network messages
     priority_requests: bool,
 
+    /// Per-key hint peers indicating which peers likely have data for each key.
+    /// Hinted peers are tried first, waiting for them if rate-limited. If all hinted
+    /// peers are unavailable (disconnected or blocked), hints are cleared and any peer
+    /// is tried. On failure, only the failing peer is removed; on success, all hints
+    /// for that key are cleared.
+    hints: HashMap<Key, HashSet<P>>,
```
patrick-ogrady: We need to replace `HashSet` with some ordered struct (otherwise we lose determinism).

patrick-ogrady: Well, let me see if we actually iterate over these.

p2p/src/utils/requester/mod.rs
```diff
@@ -7,4 +7,4 @@ mod requester;
 
 pub use config::Config;
 pub(crate) use metrics::{Metrics, PeerLabel};
-pub use requester::{Request, Requester, ID};
+pub use requester::{Request, RequestError, Requester, ID};
```
patrick-ogrady: We can just call this `Error`.

resolver/src/p2p/fetcher.rs
andresilva: We don't iterate over the set but we can use `BTreeSet<P>` since `P` is `PublicKey` so it implements `Ord`. Should be fine to keep as is though, we only use this to pass as a filter so only `contains` is being called.

resolver/src/p2p/fetcher.rs
```diff
@@ -264,10 +339,52 @@ impl<E: Clock + GClock + Rng + Metrics, P: PublicKey, Key: Span, NetS: Sender<Pu
...
+    ///
+    /// Also removes the peer from all hint sets.
     pub fn block(&mut self, peer: P) {
+        // Remove peer from all hint sets
+        for hints in self.hints.values_mut() {
+            hints.remove(&peer);
+        }
+
+        // Clean up empty hint sets
+        self.hints.retain(|_, v| !v.is_empty());
```
andresilva: This will iterate over all the hints, I assumed this will stay reasonably small so refrained from optimizing it. Otherwise we could maintain a reverse index `HashMap<P, HashSet<Key>>` to directly find which keys have a peer hinted.

resolver/src/p2p/fetcher.rs
```diff
@@ -264,10 +321,52 @@ impl<E: Clock + GClock + Rng + Metrics, P: PublicKey, Key: Span, NetS: Sender<Pu
...
+        self.targets.insert(key, peers.into_iter().collect());
+    }
+
+    /// Clear targeting for a key.
+    ///
+    /// After this call, the fetch will try any available peer instead of being
+    /// restricted to targets.
+    pub fn untarget(&mut self, key: &Key) {
+        self.targets.remove(key);
+    }
```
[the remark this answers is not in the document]
andresilva: 🏆

p2p/src/utils/requester/requester.rs
```diff
@@ -166,7 +212,12 @@ impl<E: Clock + GClock + Rng + Metrics, P: PublicKey> Requester<E, P> {
 
         // Increment failed metric if no participants are available
         self.metrics.created.inc(Status::Failure);
-        Err(next)
+
+        if next == Duration::MAX {
+            Err(Error::NoEligibleParticipants)
```
patrick-ogrady: We should change `next` to `Option<>`...it is possible `limit.wait_time_from` could return Duration::MAX as well.

resolver/src/p2p/engine.rs
```diff
@@ -404,11 +439,16 @@ impl<
...
         // Get the key associated with the response, if any
-        let Some(key) = self.fetcher.pop_by_id(id, &peer, false) else {
+        let Some((key, targeted)) = self.fetcher.pop_by_id(id, &peer, false) else {
             // It's possible that the key does not exist if the request was canceled
             return;
         };
 
+        // Target peer didn't have the data
+        if targeted {
+            self.metrics.target_miss.inc();
```
patrick-ogrady: now that we don't fallback, I think we can just use our existing metrics?

resolver/src/p2p/fetcher.rs
```diff
@@ -264,10 +321,50 @@ impl<E: Clock + GClock + Rng + Metrics, P: PublicKey, Key: Span, NetS: Sender<Pu
     }
 
     /// Blocks a peer from being used to fetch data.
+    ///
+    /// Also removes the peer from all target sets.
     pub fn block(&mut self, peer: P) {
+        // Remove peer from all target sets (keeping empty entries)
+        for targets in self.targets.values_mut() {
```
patrick-ogrady, line 345: I think we "should" just skip this but probably ok to do anyways?

resolver/src/p2p/fetcher.rs
```diff
@@ -264,10 +321,50 @@ impl<E: Clock + GClock + Rng + Metrics, P: PublicKey, Key: Span, NetS: Sender<Pu
...
 
+    /// Add or replace target peers for fetching a key.
+    ///
+    /// If `replace` is true, atomically replaces the target set.
+    /// If `replace` is false, adds to the existing target set.
+    ///
+    /// If the resulting target set is empty, the fetch will wait until targets are added.
+    ///
+    /// Clears the waiter to allow immediate retry if the fetch was blocked waiting for targets.
+    pub fn add_targets(&mut self, key: Key, peers: impl IntoIterator<Item = P>, replace: bool) {
```
patrick-ogrady: For the initial pass, I'm ok not offering this atomic replace. Once you add targets you can't "undo" (unless a peer is blocked).

resolver/src/p2p/ingress.rs
```diff
@@ -88,3 +114,111 @@ impl<K: Span> Resolver for Mailbox<K> {
...
+    ///
+    /// Atomically replaces the target set. Use this when you want to change
+    /// which peers are tried without adding to the existing set.
+    ///
+    /// If `targets` is empty, the fetch will wait until targets are added via
+    /// [`target`](Self::target). To clear targeting and try any peer, use
+    /// [`untarget`](Self::untarget) instead.
+    ///
+    /// Panics if the send fails.
+    pub async fn retarget(&mut self, key: K, targets: Vec<P>) {
```
patrick-ogrady: As noted elsewhere, let's remove this for the initial pass

resolver/src/p2p/ingress.rs
```diff
@@ -88,3 +114,111 @@ impl<K: Span> Resolver for Mailbox<K> {
...
+        self.sender
+            .send(Message::Target {
+                key,
+                targets: Some((targets, true)),
+            })
+            .await
+            .expect("Failed to send retarget");
+    }
+
+    /// Clear targeting for an **ongoing** fetch.
```
patrick-ogrady: Let's just do this implicitly if you fetch (with no targets)

resolver/src/p2p/ingress.rs
```diff
@@ -88,3 +114,111 @@ impl<K: Span> Resolver for Mailbox<K> {
...
+    ///
+    /// Multiple calls add to the existing target set. Use [`retarget`](Self::retarget)
+    /// to replace targets, or [`untarget`](Self::untarget) to clear targeting.
+    ///
+    /// Targets are automatically cleared when the fetch succeeds or is canceled.
+    /// When a peer is blocked (sent invalid data), only that peer is removed
+    /// from the target set.
+    ///
+    /// Panics if the send fails.
+    pub async fn target(&mut self, key: K, targets: Vec<P>) {
```
patrick-ogrady: Let's do this implicitly if you call `fetch_targeted`

patrick-ogrady (changes requested):
I think the best approach is to thin down the ingress interface for resolver. With the current approach, we need to track outstanding requests on the caller side to properly manipulate the targeting. I think we should avoid this duplicated work and "just handle it".

resolver/src/p2p/fetcher.rs
andresilva: This will eventually get cleared as the requests resolve, so I guess it's OK to just leave blocked peers there?

resolver/src/p2p/engine.rs
andresilva: TBH these were initially mostly useful for the integration tests, so that we can externally observe whether the targets (then hints) were being used (initially hints). I'll see if we can do the same assertions in the integration tests without these.

resolver/src/p2p/engine.rs
```diff
@@ -228,20 +228,26 @@ impl<
...
-                        Message::Fetch { keys } => {
-                            for key in keys {
+                        Message::Fetch(requests) => {
+                            for FetchRequest { key, targets } in requests {
                                 trace!(?key, "mailbox: fetch");
 
                                 // Check if the fetch is already in progress
-                                if self.fetch_timers.contains_key(&key) {
-                                    trace!(?key, "duplicate fetch");
-                                    self.metrics.fetch.inc(Status::Dropped);
```
patrick-ogrady, line 238: I agree we shouldn't inc drop if we get a duplicate request

resolver/src/p2p/ingress.rs
```diff
@@ -88,3 +110,64 @@ impl<K: Span> Resolver for Mailbox<K> {
...
+    /// Send fetch requests for multiple keys, each with their own targets.
+    ///
+    /// See [`fetch_targeted`](Self::fetch_targeted) for details on target behavior.
+    ///
+    /// # Panics
+    ///
+    /// Panics if any `targets` is empty or if the send fails.
+    pub async fn fetch_all_targeted(&mut self, requests: Vec<(K, Vec<P>)>) {
+        for (_, targets) in &requests {
+            assert!(
```
patrick-ogrady: Can we move this `assert` into the `map()` to avoid duplicate looping?

resolver/src/p2p/fetcher.rs
```diff
@@ -90,22 +109,43 @@ impl<E: Clock + GClock + Rng + Metrics, P: PublicKey, Key: Span, NetS: Sender<Pu
...
-        let (peer, id) = match self.requester.request(retry) {
+        // Peek at the pending key to check for targets
+        let (key, (_, retry)) = self.pending.peek().unwrap();
+
+        // Get peer to send request to, using targets if available
+        let result = match self.targets.get(key) {
+            Some(targets) if targets.is_empty() => Err(Error::NoEligibleParticipants),
+            Some(targets) => self
+                .requester
+                .request_filtered(*retry, |p| targets.contains(p)),
```
patrick-ogrady: Interesting question (I don't think it should block merge): if the next request is "targeted" and that peer is rate limited, it is possible that we'll then encounter head-of-line blocking on that request (although other peers could be used to fulfill other requests)

resolver/src/p2p/fetcher.rs
```diff
@@ -90,22 +109,43 @@ impl<E: Clock + GClock + Rng + Metrics, P: PublicKey, Key: Span, NetS: Sender<Pu
...
-            Err(next) => {
-                let waiter = self.context.current().saturating_add(next);
-                self.waiter = Some(waiter);
+            Err(Error::RateLimited(wait)) => {
+                self.waiter = Some(self.context.current().saturating_add(wait));
+                return;
+            }
+            Err(Error::NoEligibleParticipants) => {
+                // No peers available at all - set waiter to far future to avoid busy loop
+                self.waiter = Some(self.context.current().saturating_add(Duration::MAX));
```
patrick-ogrady: A request to a target that gets blocked or is removed could also block forever 🤔

patrick-ogrady (commented):
Looks good module the Head-of-Line blocking 🤔

resolver/src/p2p/fetcher.rs
patrick-ogrady: Resolved: #2410

PR #2393, opened 2025-12-04
[consensus/marshal] Cache proposed blocks

clabby:
Future Work: https://github.com/commonwarexyz/monorepo/issues/2394

## Overview

Changes marshal to cache proposed blocks on `broadcast`.

### DIFF
--- a/consensus/src/marshal/actor.rs
+++ b/consensus/src/marshal/actor.rs
@@ -355,7 +355,7 @@ where
                             };
                             let _ = response.send(info);
                         }
-                        Message::Broadcast { round, block } => {
+                        Message::Proposed { round, block } => {
                             self.cache_verified(round, block.commitment(), block.clone()).await;
                             let _peers = buffer.broadcast(Recipients::All, block).await;
                         }
--- a/consensus/src/marshal/ingress/mailbox.rs
+++ b/consensus/src/marshal/ingress/mailbox.rs
@@ -97,8 +97,8 @@ pub(crate) enum Message<S: Scheme, B: Block> {
         /// A channel to send the retrieved block.
         response: oneshot::Sender<B>,
     },
-    /// A request to broadcast a block to all peers.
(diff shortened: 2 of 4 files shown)


consensus/src/marshal/actor.rs
```diff
@@ -355,8 +355,9 @@ where
                             };
                             let _ = response.send(info);
                         }
-                        Message::Broadcast { block } => {
-                            let _peers = buffer.broadcast(Recipients::All, block).await;
+                        Message::Broadcast { round, block } => {
+                            let _peers = buffer.broadcast(Recipients::All, block.clone()).await;
+                            self.cache_verified(round, block.commitment(), block).await;
```
patrick-ogrady: We should store before broadcast (if we want to ensure we have a block before people vote for it).

consensus/src/marshal/actor.rs
```diff
@@ -355,8 +355,9 @@ where
                             };
                             let _ = response.send(info);
                         }
-                        Message::Broadcast { block } => {
-                            let _peers = buffer.broadcast(Recipients::All, block).await;
+                        Message::Broadcast { round, block } => {
```
patrick-ogrady: nit: I think we should change this to `Proposed`? It is a better description of how marshal things of this (in that we store + broadcast.

consensus/src/marshal/ingress/mailbox.rs
```diff
@@ -97,8 +97,10 @@ pub(crate) enum Message<S: Scheme, B: Block> {
         /// A channel to send the retrieved block.
         response: oneshot::Sender<B>,
     },
-    /// A request to broadcast a block to all peers.
-    Broadcast {
+    /// A request to broadcast a proposed block to all peers.
+    Propose {
```
patrick-ogrady: nit: `proposed` like we do for `verified`

Issue #1767, opened 2025-10-01
[consensus::simplex] Allow for a second, post-notarization-certificate, `verify(...)` hook to the `Automaton`, called `finalizable(...)`


BrendanChou:
Currently, the consensus engine will query the `Automaton` about the validity of a block when it initially receives a block proposal in a view, using the `Automaton.verify(commitment)` function. Essentially, it is asking the `Automaton` whether it can safely notarize the commitment. In "normal" consensus, this ensures that the application can access the block and that it's well-formatted.

However, in block-encoding constructions, rather than giving each validator the entire block directly, proposer ask validators to notarize a commitment with a block "shard". The validators "echo" the block shard with their attached notarization to all other validators. Since each validator cannot verify the entire block at this point, the `verify(...)` function can only verify that the single shard is received and well-formatted.

Once the consensus engine receives a notarization certificate, each validator should have enough shards to reconstruct the block. At this point, it should once again hook into the `Automaton` using a new `finalizable(...)` function. Now that the validator has the full block, it can properly validate the block. Technically there are three possibilities for the block given the shards seen:
- Reconstructable and Valid
- Reconstructable and Invalid
- Not Reconstructable

The determination of each of these cases will be deterministic and consistent across all honest validators, even if they see different subsets of shards (proof left as an exercise to the reader). For either of the last two cases above, the `Automaton.finalizable(commitment)` function should return `false`. This allows the consensus engine to treat the notarization purely as a nullification. This means that the validator should move to the next view, not vote to finalize the block, and not vote to build on that block.

Open questions:
- How to easily "communicate" or show that an invalid notarization should be treated as a nullification?
- How can this be adapted to Minimmit?

patrick-ogrady:
This is very similar to the [DispersedSimplex](https://eprint.iacr.org/2023/1916) construction with the slight tweak that observing a "complaint" locally permits you to enter the next view before seeing a "nullification" form (i.e. a "virtual nullification").

BrendanChou:
Looks like I actually made this overly complicated. Originally I suggested treating a `notarization` for an invalid block as a `nullification`. Since each honest validator would come to the same conclusion, this saves a round of communication instead of actually having to send a `nullify` message and forming a true `nullification` certificate.

[...]

PR #2396, opened 2025-12-04
[consensus::simplex] Add `leader` and `parent_payload` to `Proposal`

### LINKED ISSUE #2380: [consensus::simplex] Add `leader` and `parent_payload` to `Proposal`
This way, a certificate (notarization or finalization) self-contain all the necessary fields to populate the `Context`. It's slightly more data per-vote and per-certificate (roughly 1 megabit of bandwidth per block for 1000 validators, or 1ms of latency when on a Gigabit connection).

Breaking: the wire and storage format of votes in Simplex

BrendanChou:
Fixes #2380 by making it such that a certificate (notarization or finalization) contains all necessary fields to populate the `Context`. This will make the future issue https://github.com/commonwarexyz/monorepo/issues/1767 much cleaner to implement.

I added one small optimization that checks the existence of a notarization/finalization at the parent view before checking the chain of intermediate nullifications (which may be unbounded in size

### DIFF
--- a/consensus/fuzz/src/disrupter.rs
+++ b/consensus/fuzz/src/disrupter.rs
@@ -306,35 +306,40 @@ where
         mutation: Mutation,
     ) -> Proposal<Sha256Digest> {
         match mutation {
-            Mutation::Payload => Proposal::new(
-                Round::new(original.epoch(), original.view()),
-                original.parent,
-                self.payload(),
-            ),
-            Mutation::View => Proposal::new(
-                Round::new(original.epoch(), View::new(self.random_view(self.view))),
-                original.parent,
-                original.payload,
-            ),
-            Mutation::Parent => Proposal::new(
-                Round::new(original.epoch(), original.view()),
-                View::new(self.parent()),
-                original.payload,
-            ),
-            Mutation::All => Proposal::new(
-                Round::new(original.epoch(), View::new(self.random_view(self.view))),
-                View::new(self.parent()),
-                self.payload(),
-            ),
+            Mutation::Payload => Proposal {
+                round: Round::new(original.epoch(), original.view()),
+                leader: original.leader,
+                parent: original.parent,
+                payload: self.payload(),
+            },
+            Mutation::View => Proposal {
+                round: Round::new(original.epoch(), View::new(self.random_view(self.view))),
+                leader: original.leader,
+                parent: original.parent,
+                payload: original.payload,
+            },
+            Mutation::Parent => Proposal {
+                round: Round::new(original.epoch(), original.view()),
+                leader: original.leader,
+                parent: (View::new(self.parent()), self.payload()),
+                payload: original.payload,
+            },
+            Mutation::All => Proposal {
+                round: Round::new(original.epoch(), View::new(self.random_view(self.view))),
+                leader: original.leader,
+                parent: (View::new(self.parent()), self.payload()),
+                payload: self.payload(),
+            },
         }
     }
 
     async fn send_random(&mut self, sender: &mut impl Sender) {
-        let proposal = Proposal::new(
-            Round::new(Epoch::new(EPOCH), View::new(self.random_view(self.view))),
-            View::new(self.parent()),
-            self.payload(),
-        );
+        let proposal = Proposal {
+            round: Round::new(Epoch::new(EPOCH), View::new(self.random_view(self.view))),
+            leader: 0,
+            parent: (View::new(self.parent()), self.payload()),
+            payload: self.payload(),
+        };
 
         if self.participants.index(&self.validator).is_none() {
             let bytes = self.bytes();


consensus/src/simplex/types.rs
```diff
@@ -635,27 +635,22 @@ impl<S: Scheme, D: Digest> From<Certificate<S, D>> for Artifact<S, D> {
 pub struct Proposal<D: Digest> {
     /// The round in which this proposal is made
     pub round: Round,
+    /// The index of the leader proposing this block
+    pub leader: u32,
     /// The view of the parent proposal that this one builds upon
     pub parent: View,
```
patrick-ogrady: nit: (View, D)?

patrick-ogrady:
For reviewers, can you leave a comment why this makes implementing coding support easier?

BrendanChou:
I could also add the following, but it may not be entirely necessary since we are simply treating invalid values similarly to a no-proposal. So even though the behavior is byzantine, we aren't really expending any additional resources here.
- Immediately reject proposals with incorrect `parent_payload` or `leader`
- Block proposers with incorrect `parent_payload` or `leader`

clabby:
bugbot run

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -478,30 +484,35 @@ impl<E: Clock + Rng + CryptoRng + Metrics, S: Scheme, D: Digest> State<E, S, D>
...
 
-        // May return `None` if the parent view is not yet notarized or finalized.
-        self.quorum_payload(parent).copied()
+        // Ensure we have a certificate at the parent view.
+        if Some(&proposal.parent.1) != self.quorum_payload(parent) {
+            return false;
+        };
+
+        // Check that there are nullifications for all views between the parent and the proposal view.
+        View::range(parent.next(), view).all(|v| self.is_nullified(v))
```
andresilva:
```suggestion
        // Validate the declared parent by walking backward through certified parents and nullifications.        
        match self.find_parent(view) {
            Ok(parent) => parent == proposal.parent,
            Err(_) => false,
        }
```

?

andresilva (dismissed):

andresilva:
bugbot run

consensus/src/marshal/mod.rs
```diff
@@ -523,7 +523,11 @@ mod tests {
                 // Notarize block by the validator that broadcasted it
                 let proposal = Proposal {
                     round,
-                    parent: View::new(height.checked_sub(1).unwrap()),
+                    leader: 0,
+                    parent: (
+                        View::new(height.checked_sub(1).unwrap()),
+                        Sha256Digest::from([0u8; 32]),
```
patrick-ogrady: Shouldn't this be `height-1` block?

consensus/src/marshal/mod.rs
```diff
@@ -667,7 +671,11 @@ mod tests {
...
```
patrick-ogrady: height - 1 block?

consensus/src/marshal/mod.rs
patrick-ogrady: This will fail when we remove this from `block` (which we should now do I think?): https://github.com/commonwarexyz/monorepo/blob/ad77e2400313dfbcb389fe6f06afb5d07097fb9d/consensus/src/lib.rs#L45-L46

patrick-ogrady: It would only complicate things (as there are now 2 ways to get parent?)

consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -797,11 +797,14 @@ impl<
...
-                        context.round,
-                        context.parent.0,
-                        proposed,
-                    );
+                    let leader = self.state.leader_index(context.round.view())
+                        .expect("leader must be set when proposing");
+                    let proposal = Proposal {
+                        round: context.round,
+                        leader,
+                        parent: context.parent,
```
patrick-ogrady, line 805: ❤️

consensus/src/simplex/actors/voter/mod.rs
```diff
@@ -1604,11 +1631,12 @@ mod tests {
...
             let view = View::new(2);
-            let proposal = Proposal::new(
-                Round::new(Epoch::new(333), view),
-                view.previous().unwrap(),
-                Sha256::hash(b"finalize_without_notarization"),
-            );
+            let proposal = Proposal {
+                round: Round::new(Epoch::new(333), view),
+                leader: 0,
+                parent: (view.previous().unwrap(), Sha256Digest::from([0u8; 32])),
```
patrick-ogrady, line 1637: should make this accurate?

consensus/src/simplex/types.rs
```diff
@@ -635,26 +635,18 @@ impl<S: Scheme, D: Digest> From<Certificate<S, D>> for Artifact<S, D> {
 pub struct Proposal<D: Digest> {
     /// The round in which this proposal is made
     pub round: Round,
-    /// The view of the parent proposal that this one builds upon
-    pub parent: View,
+    /// The index of the leader proposing this block
+    pub leader: u32,
+    /// The view and payload of the parent proposal that this one builds upon
+    pub parent: (View, D),
```
patrick-ogrady, line 641: I originally opted to avoid this approach because it now means there are really 2 hash chains (one maintained by consensus and probably one in the application).

This is no longer the case, however, in commonware-optimized execution environments that would no longer need to include the parent digest in the block (as it is maintained here) 🤔

patrick-ogrady: At that time, I felt having a single linkage (maintained in application land) was preferred.

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -478,30 +484,35 @@ impl<E: Clock + Rng + CryptoRng + Metrics, S: Scheme, D: Digest> State<E, S, D>
...
+    /// intermediate views are nullified.
+    fn has_valid_proposal_context(&self, proposal: &Proposal<D>) -> bool {
         // Sanity check that the parent view is less than the proposal view.
-        let (view, parent) = (proposal.view(), proposal.parent);
-        if view <= parent {
-            return None;
+        let (view, parent) = (proposal.view(), proposal.parent.0);
+
+        // Invalid parent view - parent must be less than the proposal view.
+        if parent >= view {
```
patrick-ogrady: I prefer `view <= parent` here

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -478,30 +484,35 @@ impl<E: Clock + Rng + CryptoRng + Metrics, S: Scheme, D: Digest> State<E, S, D>
...
-            return None;
+        // Validate leader index matches the expected leader for this view.
+        if Some(proposal.leader) != self.leader_index(view) {
+            return false;
         }
 
-        // May return `None` if the parent view is not yet notarized or finalized.
-        self.quorum_payload(parent).copied()
+        // Ensure we have a certificate at the parent view.
+        if Some(&proposal.parent.1) != self.quorum_payload(parent) {
```
patrick-ogrady: ❤️

consensus/src/marshal/mod.rs
patrick-ogrady: Ah shoot, this isn't possible actually.

patrick-ogrady: We may only have the block (no certificate). This change would require us **always** having the certificate....

consensus/src/simplex/actors/voter/state.rs
BrendanChou: No because `find_parent` will find the highest possible parent, however if that parent also has a nullification at the same view, then the proposed parent may also be valid.

That is to say, there may be multiple valid parents, so we should validate the proposed parent "directly" rather than trying to find the only valid parent and ensuring the proposal matches it.

consensus/src/simplex/actors/voter/mod.rs
BrendanChou: Sorry, which part is not accurate?

consensus/src/simplex/types.rs
```diff
@@ -635,26 +635,18 @@ impl<S: Scheme, D: Digest> From<Certificate<S, D>> for Artifact<S, D> {
 pub struct Proposal<D: Digest> {
     /// The round in which this proposal is made
     pub round: Round,
-    /// The view of the parent proposal that this one builds upon
-    pub parent: View,
+    /// The index of the leader proposing this block
+    pub leader: u32,
```
patrick-ogrady, line 639: I think the leader certificate could be dramatically more useful if this is a public key (at least as it pertains to downstream usage 🤔)

patrick-ogrady:
Ultimately, the best "pitch" I came to for this is to make `simplex` internally consistent (where no block data is required to run all necessary mechanisms).

consensus/src/simplex/types.rs
BrendanChou: That's fine but is more data which is why I avoided it at first

BrendanChou:
Yes that's the main idea

BrendanChou:
Closing in favor of #2565

Issue #2377, opened 2025-12-03
Add shim between proptest and fuzzing


cronokirby:
In `dkg2`, we've created a nice pattern where we define a test plan for various scenarios, use that to define unit tests, and then implement `arbitrary::Arbitrary` to allow the fuzzer to try different cases.

In other places, we use `proptest`, as a convenient kind of fuzzer with nice shrinking, for in crate tests.

It would be nice to avoid having to define strategies twice, for `proptest` and for `arbitrary`. We could instead define a "shim" which allows a test plan that implements `proptest::Arbitrary`, and has a way to be executed, to be exercised by a fuzzer.

This shouldn't be too hard, because the `proptest` crate has the ability to have randomness provided to it from some external source, so we could use the fuzzer to guide the proptest runner for that structure.

PR #2398, opened 2025-12-05
Create math crate

### LINKED ISSUE #2225: Idea: Factor out polynomial operations into a "math" crate
We should create an abstraction over polynomials that can work both for the BLS use-case, and for things like ZODA, to avoid duplicating optimizations, especially as we implement more complicated things like NTTs.

The motivation was looking at #1836, and realizing that a lot of the logic we would need to do that is present in the ZODA impl, just tied to a specific field. Our existing BLS modules in the cryptography crate have a bit of an abstraction for working with polynomials with different coefficients, but this is just the minimum needed to support both scalar values and group values. A proper abstraction would result in less code to maintain, and easier code to reason about, since it's not tied in to the details of a specific curve as much.

Starting to develop a proper library of abstractions would also make it easy to re-use the code we have now over different curves (which I imagine would be useful). For example, the DKG code could be written over any group, allowing it to be used, with, e.g. Ed25519 keys as well.

### LINKED ISSUE #2367: [cryptography] Create abstraction over "polynomial" of a threshold signature
Right now, we don't have a concept of a "dealing" or a "sharing" of the key. Instead, we have a public polynomial, provided directly. This is fine if you assume a particular configuration, which is that players are u32 indices, and they map to points by adding 1, and then converting to a scalar.

This poses a problem for implementing #1836, where now you need a different mechanism to map indices to points.

In general, exposing this things directly introduces a lot of coupling between the inner details of how threshold signatures work, and the actual API that you need. Having an abstraction representing the public reset of some secret sharing scheme would be nice, since it isolates these details, making changes easier, and provides a nice place for information to aid in preserving backwards compatibility. For example, we could have a version tag which would allow adopting a new set of interpolation points without a breaking change, by allowing the tag to determine the interpolation method.

cronokirby:
Closes #2225
Closes #2367 

This adds a new crate, `commonware-math`, containing:
- a trait hierarchy for algebraic types, with support for overloading operators,
- polynomial operations,
- the goldilocks field
  - and some matrix types that use this field, for coding, but should be generalized, once generic NTT support is added,

This PR also propagates the trait hierarchy throughout the cryptography crate, and replaces the BLS-specific polynomial type with the type added in this crate.

The cryptography crate was also refactored accordingly, with:
- `Point` and `Element` being removed as traits, in favor of the new traits introduced in math,
- `Digest` now inherits from `Random` (from math) instead of having its own randomness method,
- `PrivateKeyExt` is removed, with `Signer` now requiring `Random`, and the testing method to generate a private key from a seed moving into `Signer`.

### DIFF
--- a/cryptography/conformance.toml
+++ b/cryptography/conformance.toml
@@ -12,7 +12,7 @@ hash = "0081c641fd7d19194dcb85e69527d4e4fb6f722f10029ed754813e97da825e89"
 
 ["commonware_cryptography::bls12381::dkg::test::conformance::CodecConformance<DealerLog<MinPk, ed25519::PublicKey>>"]
 n_cases = 65536
-hash = "109cb246909f98ab6137e243518730cd5dff22f769ac8e631854f1392e7f5501"
+hash = "c4a3575f319b605d8479c569f04cdf73224f6067e5f6f6bab7ac2485245c0382"
 
 ["commonware_cryptography::bls12381::dkg::test::conformance::CodecConformance<DealerPrivMsg>"]
 n_cases = 65536
@@ -20,7 +20,7 @@ hash = "d112d774a4c9b012b53e44dc2a3487a0e2f87cd7b14fb8433ec7cea4997f440a"
 
 ["commonware_cryptography::bls12381::dkg::test::conformance::CodecConformance<DealerPubMsg<MinPk>>"]
 n_cases = 65536
-hash = "2366a9f077c7e9e2e6df5e32976658126c9f277a9a6e28ece9b28b82c3d2bf02"
+hash = "3ec46a7e3d5627d15285c6b51e669348bfe988829c7ea493a4bacf5d98d632f7"
 
 ["commonware_cryptography::bls12381::dkg::test::conformance::CodecConformance<DealerResult<ed25519::PublicKey>>"]
 n_cases = 65536
@@ -28,7 +28,7 @@ hash = "06cd5f1dca291e1812e01a82474926520794c8844813d64882c9a1cd6106e771"
 
 ["commonware_cryptography::bls12381::dkg::test::conformance::CodecConformance<Output<MinPk, ed25519::PublicKey>>"]
 n_cases = 65536
-hash = "aec33e38ce0e3e2be9e5343ee3d416519738d9283b49ce6a70325af926ce0326"
+hash = "aa59e152a5402fe91de7af791ee90d0301ac788904b3ee2232c97e049ceacb80"
 
 ["commonware_cryptography::bls12381::dkg::test::conformance::CodecConformance<PlayerAck<ed25519::PublicKey>>"]
 n_cases = 65536
@@ -36,7 +36,7 @@ hash = "0cbb8c8644dffe0be78d1307fb5034dd804921d66799c1aaec3d335b44e9616c"
 
 ["commonware_cryptography::bls12381::dkg::test::conformance::CodecConformance<SignedDealerLog<MinPk, ed25519::PrivateKey>>"]
 n_cases = 65536
-hash = "701ef97491885b06d40c051cb924dc4bfd975cc7f25315f8576525689bb849bc"
+hash = "b4e91bface5ddaa1fac0287e9495e51152b809350a7a42ffb042d43526180146"
 
 ["commonware_cryptography::bls12381::primitives::group::tests::conformance::CodecConformance<G1>"]
 n_cases = 65536


cryptography/fuzz/fuzz_targets/bls12381_primitive_operations.rs
```diff
@@ -226,7 +215,7 @@ enum FuzzOperation {
 
 impl<'a> Arbitrary<'a> for FuzzOperation {
     fn arbitrary(u: &mut Unstructured<'a>) -> Result<Self, arbitrary::Error> {
-        let choice = u.int_in_range(0..=47)?;
+        let choice = u.int_in_range(0..=44)?;
 
         match choice {
```
patrick-ogrady, line 212: At some point, I hope there is a better way to do this match.

cryptography/src/bls12381/dkg.rs
```diff
@@ -582,10 +542,6 @@ impl<V: Variant, P: PublicKey> Info<V, P> {
         dealers: Set<P>,
         players: Set<P>,
     ) -> Result<Self, Error> {
-        if let Some(previous) = previous.as_ref() {
```
patrick-ogrady: :heart:

cryptography/src/bls12381/dkg.rs
```diff
@@ -1178,16 +1128,16 @@ impl<V: Variant, S: Signer> Dealer<V, S> {
...
             .iter()
-            .enumerate()
-            .map(|(i, pk)| {
+            .map(|pk| {
                 (
                     pk.clone(),
                     DealerPrivMsg {
-                        share: my_poly.evaluate(i as u32).value,
+                        share: my_poly
+                            .eval_msm(&info.player_scalar(pk).expect("player should exist")),
```
patrick-ogrady, line 1140: dumb question...is there a different type of `eval` we end up doing (and if so, would we ever want to expose it here?)

cryptography/src/bls12381/dkg.rs
```diff
@@ -1305,44 +1256,37 @@ fn select<V: Variant, P: PublicKey>(
...
-        selected: Vec<(P, DealerLog<V, P>)>,
+        selected: Map<P, DealerLog<V, P>>,
         concurrency: usize,
     ) -> Result<Self, Error> {
         let (public, weights) = if let Some(previous) = info.previous.as_ref() {
-            let (indices, commitments) = selected
+            let weights = previous
+                .public()
+                .mode()
+                .subset_interpolator(previous.players(), selected.keys())
```
patrick-ogrady, line 1272: lol...a dramatic win here

cryptography/src/bls12381/dkg.rs
```diff
@@ -1512,25 +1448,26 @@ pub fn deal<V: Variant, P: Clone + Ord>(
     if players.is_empty() {
         return Err(Error::NumPlayers(0));
     }
-    let n = players.len() as u32;
-    let t = quorum(n);
-    let private = poly::new_from(&mut rng, t - 1);
+    let n = NZU32!(players.len() as u32);
+    let t = quorum(n.get());
```
patrick-ogrady: I think you can call quorum on `Set`?

cryptography/src/bls12381/dkg.rs
```diff
@@ -1512,25 +1448,26 @@ pub fn deal<V: Variant, P: Clone + Ord>(
...
+                private: eval,
             };
             (p.clone(), share)
         })
         .try_collect()
         .expect("players are unique");
     let output = Output {
         summary: Summary::random(&mut rng),
-        public: Sharing::new(mode, NZU32!(n), Poly::commit(private)),
+        public: Sharing::new(mode, n, Poly::commit(private)),
```
patrick-ogrady, line 1470: This is an example where calling `mode(n)` would be nicer :shrug:

cryptography/src/bls12381/primitives/group.rs
```diff
@@ -242,12 +213,7 @@ pub type Private = Scalar;
 pub const PRIVATE_KEY_LENGTH: usize = SCALAR_LENGTH;
 
 impl Scalar {
-    /// Generates a random scalar using the provided RNG.
-    pub fn from_rand<R: CryptoRngCore>(rng: &mut R) -> Self {
-        // Generate a random 64 byte buffer
-        let mut ikm = [0u8; 64];
-        rng.fill_bytes(&mut ikm);
-
+    fn from_64_bytes(mut ikm: [u8; 64]) -> Self {
```
patrick-ogrady: nit: `from_bytes`?

cryptography/src/bls12381/primitives/group.rs
```diff
@@ -958,45 +1098,122 @@ impl Point for G2 {
...
 #[cfg(test)]
 mod tests {
     use super::*;
     use commonware_codec::{DecodeExt, Encode};
+    use commonware_math::algebra::test_suites;
+    use proptest::prelude::*;
     use rand::prelude::*;
     use std::collections::{BTreeSet, HashMap};
 
+    impl Arbitrary for Scalar {
```
patrick-ogrady, line 1134: nit: should we define out of test for conformance? or is this done elsewhere?

cryptography/src/bls12381/primitives/ops.rs
```diff
@@ -2348,15 +2163,15 @@ mod tests {
...
         }
 
         // Batch verification
         assert!(MinPk::batch_verify(&mut OsRng, &publics, &hms, &signatures).is_ok());
 
         // Fail batch verification with a manipulated signature
-        signatures[0].add(&<MinPk as Variant>::Signature::one());
+        signatures[0] += &<MinPk as Variant>::Signature::generator();
         assert!(MinPk::batch_verify(&mut OsRng, &publics, &hms, &signatures).is_err());
     }
```
patrick-ogrady, line 2177: use `mode` in `threshold_derive_missing_partials`

math/src/poly.rs
```diff
@@ -0,0 +1,615 @@
...
+
+#[cfg(feature = "arbitrary")]
+mod fuzz {
+    use super::*;
+    use arbitrary::Arbitrary;
+
+    impl<'a, F: Arbitrary<'a>> Arbitrary<'a> for Poly<F> {
+        fn arbitrary(u: &mut arbitrary::Unstructured<'a>) -> arbitrary::Result<Self> {
+            let size = u.arbitrary_len::<F>()?.min(1);
+            let coeffs = u
```
patrick-ogrady, line 500: I think this should be .max()?

cryptography/src/bls12381/dkg.rs
```diff
@@ -1305,44 +1256,37 @@ fn select<V: Variant, P: PublicKey>(
...
             (public, Some(weights))
         } else {
             let mut public = Poly::zero();
-            for (_, log) in selected.iter() {
-                public.add(&log.pub_msg.commitment);
+            for log in selected.values() {
+                public += &log.pub_msg.commitment;
             }
             (public, None)
         };
```
patrick-ogrady, line 1292: Can't comment on it directly but `info.mode` should be used in `Output`?

cryptography/src/bls12381/scheme.rs
```diff
@@ -166,10 +168,11 @@ impl PrivateKeyExt for PrivateKey {
 #[cfg(feature = "arbitrary")]
 impl arbitrary::Arbitrary<'_> for PrivateKey {
     fn arbitrary(u: &mut arbitrary::Unstructured<'_>) -> arbitrary::Result<Self> {
+        use commonware_math::algebra::Random;
```
patrick-ogrady: nit: necessary?

cryptography/src/bls12381/scheme.rs
```diff
@@ -293,10 +296,11 @@ impl Display for PublicKey {
 impl arbitrary::Arbitrary<'_> for PublicKey {
     fn arbitrary(u: &mut arbitrary::Unstructured<'_>) -> arbitrary::Result<Self> {
         use crate::Signer;
+        use commonware_math::algebra::Random;
```
patrick-ogrady: nit: necessary?

cryptography/src/bls12381/scheme.rs
```diff
@@ -395,10 +399,11 @@ impl Display for Signature {
 impl arbitrary::Arbitrary<'_> for Signature {
     fn arbitrary(u: &mut arbitrary::Unstructured<'_>) -> arbitrary::Result<Self> {
         use crate::Signer;
+        use commonware_math::algebra::Random;
```
patrick-ogrady: nit: necessary?

cryptography/src/ed25519/scheme.rs
```diff
@@ -362,10 +365,11 @@ impl Display for Signature {
 impl arbitrary::Arbitrary<'_> for Signature {
     fn arbitrary(u: &mut arbitrary::Unstructured<'_>) -> arbitrary::Result<Self> {
         use crate::Signer;
+        use commonware_math::algebra::Random;
```
patrick-ogrady, line 367: see necessary?

cryptography/src/lib.rs
```diff
@@ -57,28 +58,21 @@ pub trait Signer: Send + Sync + Clone + 'static {
...
     fn sign(&self, namespace: &[u8], msg: &[u8]) -> Self::Signature;
-}
-
-/// A [Signer] that can be serialized/deserialized.
-pub trait PrivateKey: Signer + Sized + ReadExt + Encode + PartialEq + Array {}
 
-/// A [PrivateKey] that can be generated from a seed or RNG.
-pub trait PrivateKeyExt: PrivateKey {
-    /// Create a [PrivateKey] from a seed.
+    /// Create a [Signer] from a seed.
```
patrick-ogrady, line 62: Admittedly, I liked this as `Ext` because I didn't want to enforce all signers could be created from a seed...maybe that it is too pedantic...

examples/bridge/src/types/outbound.rs
```diff
@@ -82,11 +82,12 @@ mod tests {
     };
     use commonware_cryptography::{
         bls12381::primitives::{
-            group::{self, Element},
+            group::{self},
```
patrick-ogrady: nit: `group`

patrick-ogrady (commented):
Nice work!

cryptography/src/lib.rs
cronokirby: This doesn't quite enforce it so much as it provides a convenience method for doing so. Arguably, we could remove this entirely because there's so many other places where the API requires an RNG, and the test creates one.

cryptography/src/bls12381/dkg.rs
cronokirby: `eval_msm` always returns the same result as `eval` (there's a test for this!), it should just be faster for a polynomial of group points. Whereas, for a polynomial of scalars, you want to avoid the cost of materializing the powers of the input in memory, so you use `eval`

cryptography/fuzz/fuzz_targets/bls12381_primitive_operations.rs
cronokirby: Yeah we could just `derive(Arbitrary)` on the enum

cronokirby: (the fuzz testing could be cleaned up substantially, especially with #2377 )

cryptography/src/bls12381/primitives/group.rs
cronokirby: This is `proptest::Arbitrary`, not `arbitrary::Arbitrary`, which is, admittedly, confusing. Avoiding having to define both of these is what #2377 should help with.

cryptography/src/bls12381/dkg.rs
cronokirby: Yep that was a mistake

cryptography/src/bls12381/dkg.rs
patrick-ogrady: Ahhh, duh. Makes sense.

cryptography/src/lib.rs
patrick-ogrady: I think a better place for this could actually be in `math` (then we don't have to maintain an insecure random thin here)

cronokirby:
Seems like the conformance tests are failing just because we changed the arbitrary impl for `Poly`. I think the current implementation is more idiomatic than doing `u.arbitrary<u32>() % 10` to generate the size, so opting to regenerate.

cronokirby:
The only actual breaking format change is to `Output`, the rest should be false positives.

cryptography/src/bls12381/primitives/group.rs
```diff
@@ -319,6 +293,14 @@ impl Scalar {
...
     /// Creates a new scalar from the provided index (a scalar offset by 1).
+    ///
+    /// This is important, because when doing polynomial secret sharing,
+    /// `f(0)` is the secret, and `f(x_0), f(x_1), ...` are the shares, for
+    /// particular points `x_i`.
+    ///
+    /// One simple method of choosing the points is this method, which
+    /// maps `0` to `1`, `1` to `2`, etc. ensuring that none of the points
+    /// collide with `0`, which would reveal the secret.
     pub fn from_index(i: u32) -> Self {
```
patrick-ogrady: I guess my only real feedback at this point is whether or not this should be part of `Mode` (more or less replacing our call to `from_index` to be `from_u64` with this 1 offset and comment.)

patrick-ogrady (commented):
Very close now.

cryptography/src/bls12381/primitives/group.rs
cronokirby: I think removing this makes sense, yeah.

cryptography/src/lib.rs
cronokirby: Yeah, this would definitely fit on `Random`, but really I think the compositional approach of just making the user create a seeded RNG themselves if that's what they want is better.

In fact, most of the time, this method is called immediately below:

```
|context| {
  ...
```

so it would not be a big lift at all to just use the context instead.

But, that would be a big diff. Maybe let's create an issue for that?

math/Cargo.toml
```diff
@@ -0,0 +1,40 @@
+[package]
+name = "commonware-math"
+edition.workspace = true
+publish = true
+version.workspace = true
+license.workspace = true
+description = "Mathematical utilities and operations."
```
patrick-ogrady: nit: we use a "verb" and some succicnt action for these:

Image

math/Cargo.toml
patrick-ogrady, line 1: Create a `README` with the following content:

```
# commonware-cryptography

[![Crates.io](https://img.shields.io/crates/v/commonware-cryptography.svg)](https://crates.io/crates/commonware-cryptography)
[![Docs.rs](https://docs.rs/commonware-cryptography/badge.svg)](https://docs.rs/commonware-cryptography)

Generate keys, sign arbitrary messages, and deterministically verify untrusted signatures.

## Status 

`commonware-cryptography` is **ALPHA** software and is not yet recommended for production use. Developers should expect breaking changes and occasional instability.
```

patrick-ogrady, line 1: Add to `publish.yml` after `codec` and `utils`: https://github.com/commonwarexyz/monorepo/blob/5da17273b64d77f24708b67ad20f45cdbce67da3/.github/workflows/publish.yml#L32

patrick-ogrady (commented):
publish nits

cronokirby:
Pushed doc changes

cryptography/src/bls12381/dkg.rs
```diff
@@ -328,62 +329,17 @@ pub enum Error {
...
-    }
-}
-
 /// The output of a successful DKG.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct Output<V: Variant, P> {
     summary: Summary,
     players: Set<P>,
-    public: Public<V>,
+    public: Sharing<V>,
```
patrick-ogrady, line 337: Making `Mode` now makes sense to me (as something you'd persist with the sharing to look at some historical output)

cronokirby:
> +        let mut neg_rhs = rhs.0;
> +        let ptr = &raw mut self.0;
> +        // SAFETY: blst_p1_cneg and blst_p1_add_or_double are safe with valid pointers.
> +        // blst_p1_add_or_double supports in-place (ret==a). Raw pointer avoids aliased refs.
> +        unsafe {
> +            blst_p1_cneg(&mut neg_rhs, true);
> +            blst_p1_add_or_double(ptr, ptr, &neg_rhs);
> +        }
>      }

The LLMs were smart to catch this ; there was indeed an unnecessary copy, but by being a bit smarter without how we arrange methods, we can eliminate it without creating a new unsafe block 😎

patrick-ogrady (approved):

Issue #2034, opened 2025-10-28
[p2p/lookup] Add Support for Persistent Peers


patrick-ogrady:
Peers can only be added in `lookup` if they are registered. Some operators want to add other nodes as well (that receive certificate broadcasts from consensus, for example). We should offer a config option to include these sorts of peers instead of asking operators to modify the peer set for a given index (which is error-prone, as a synchronized list there is necessary for proper function in `discovery`).

patrick-ogrady:
We should consider tuning the resolver to only use trusted-peers if the node isn't part of any tracked set (as no other peers will accept their connection let alone handle a request).

patrick-ogrady:
It may be interesting to create a "listener" p2p primitive that works with both lookup/discovery where the p2p participant adds some list of peer identities (and their IPs) that are considered when broadcasting "all" and a standalone client that doesn't care about the actual peer set and is optimized to just pull from a list of registered peers).

patrick-ogrady:
It may also be nice to create a "multicast" dialect that allows anyone to connect and just passes on "all" messages (not sure how to make this safe but could be coupled with trusted peers to create a standalone peer binary).

patrick-ogrady:
Instead of having "persistent peers" be listed by public key (lot of configuration work), it may be better to just open a separate port that if accessible anyone can join.

0x00101010:
@patrick-ogrady opened a draft PR to start conversation for alignment, let me know what you think, details in the PR: https://github.com/commonwarexyz/monorepo/pull/2399

PR #2030, opened 2025-10-28
[consensus/simplex] WebSocket reporter

andresilva:
This PR adds a WebSocket reporter for the simplex consensus engine that streams real-time consensus activities as binary over the network. The reporter uses a broadcast channel with two components: a `WebSocketReporter` that implements the `Reporter` trait and can be passed to the consensus engine, and a separate consumable `WebSocketServer` actor that handles the actual WebSocket connections. When clients connect, they first receive a context message containing participant information followed by a continuous stream of `Activity` messages. The implementation requires a tokio runtime and uses tokio-tungstenite for the WebSocket server.

Related #1921.

PR #2399, opened 2025-12-05
[consensus::simplex] Add observers for listening to consensus certificates

0x00101010:
Resolves https://github.com/commonwarexyz/monorepo/issues/2034

Initial draft PR opened for conversation:

Current implementation:
1. Adding a new observer actor that handles external connections (unauthenticated)
2. Observer sits at simplex layer instead of p2p layer, since we're not a participant in the p2p network
4. broadcast certificates in obserer
5. broadcasts replay certificates

Things that need alignment on:
1. Seems to me that just open a websocket port for observers to listen might be the better solution for this case (if people just want to observe the network consensus progress), Though current implementation takes in a `Listener` type currently, ideally we take in just a `WebSocketStream`, let me know if there might be other scenarios that require lower level Listener
2. imo we should by default not broadcast replay messages (tons of waste basically), but in case that listener wants historical state (for backfill or whatever), it could be helpful. Do we want to make it configurable or just simply not broadcast them?

Remaining work:
1. update tests & examples (reshare / bridge, etc) with new setup
2. tests
3. example client or websocket client (depending on if we use Listener or WebSocketStream)
6. docs

### DIFF
--- a/consensus/src/simplex/actors/observer/mod.rs
+++ b/consensus/src/simplex/actors/observer/mod.rs
@@ -20,7 +20,7 @@ impl Default for Config {
     fn default() -> Self {
         Self {
             listen_addr: "0.0.0.0:8001".parse().unwrap(),
-            max_observers: 1000,
+            max_observers: 100,
         }
     }
 }
--- a/consensus/src/simplex/actors/resolver/actor.rs
+++ b/consensus/src/simplex/actors/resolver/actor.rs
@@ -266,16 +266,11 @@ impl<
         }
     }
 
-    async fn broadcast_to_observers(observer: &mut Option<observer::Mailbox<S, D>>, voter: Voter<S, D>) {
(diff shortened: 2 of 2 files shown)


consensus/src/simplex/engine.rs
```diff
@@ -89,6 +96,14 @@ impl<
             },
         );
 
+        let observer_config = cfg.listen_addr.map(|listen_addr| observer::Config {
```
patrick-ogrady, line 99: This should require no changes to `simplex` and can be defined entirely over the `Reporter` stream 👍

patrick-ogrady: See #2030 for an example of this

patrick-ogrady: I also wrote up an issue discussing what e2e impl we'd want to build with this: #2466

PR #2401, opened 2025-12-05
[workspace] just fix

andresilva:
Add command for fixing clippy lints and command to fix all issues (clippy, format and toml).

### DIFF
--- a/justfile
+++ b/justfile
@@ -3,7 +3,7 @@ set positional-arguments := true
 env_nightly_version := env("NIGHTLY_VERSION", "nightly")
 nightly_version := if env_nightly_version != "" { "+" + env_nightly_version } else { "" }
 
-alias f := fix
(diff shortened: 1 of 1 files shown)


justfile
```diff
@@ -3,7 +3,7 @@ set positional-arguments := true
 env_nightly_version := env("NIGHTLY_VERSION", "nightly")
 nightly_version := if env_nightly_version != "" { "+" + env_nightly_version } else { "" }
 
-alias f := fix-fmt
+alias f := fix
```
andresilva: Can revert this if too breaking.

clabby: hmm - maybe lets do 2 separate aliases (one for fix fmt, one for full fix?)

I use this alias quite a bit and would prefer that it doesn't run clippy/toml fmt every time.

PR #2402, opened 2025-12-05
[AGENTS.md] Add comment about storage failures

danlaine:


AGENTS.md
```diff
@@ -470,6 +470,13 @@ fn test_storage_conformance() {
 - **Test Pruning**: Verify old data can be safely removed
 - **Test Concurrent Access**: Multiple readers/writers on same storage
 
+### Handling Storage Failures
```
patrick-ogrady: nit: would just add another bullet:

```
- **Failures Are Fatal**: Errors returned by mutable methods (e.g. `put`,
`delete`, `sync`) are treated as unrecoverable. The database may be in an inconsistent state after
such an error. Callers must not use a database after a mutable method returns an error. Reviews
need not comment the database being in an inconsistent state after such an error.

patrick-ogrady (changes requested):

AGENTS.md
danlaine: Done

patrick-ogrady (approved):

PR #2403, opened 2025-12-05
[macros] Fuse futures inline

clabby:
## Overview

Fuses the futures for `commonware_macros::select!` in-line to prevent holding mutable references across select branches.


cronokirby:
This fixes the issues I had using select in #2354!

patrick-ogrady (approved):
Major skill issue on my part 😓

PR #2405, opened 2025-12-05
[Storage] Replace `IndexedLog` generic `I: Index` with `T: Translator`

danlaine:
For the unordered/ordered implementations of `IndexedLog`, we always use the same index implementation `I`.

This PR reflects that by explicitly naming the concrete index type in `IndexedLog`.


roberto-bayardo:
I plan on instantiating some variants with a PartitionedIndex, so I believe we may need to keep this generic.

PR #2407, opened 2025-12-05
fix(stream/crypto): bind handshake transcript to app namespace

phrwlk:
Bind the handshake transcript to an application-specific namespace to enforce cross-application domain separation. Previously, stream::Config::namespace was ignored while handshake used a fixed namespace, which contradicted stream docs and weakened replay isolation. This plumbs the namespace from stream::dial/listen into handshake::Context and seeds the transcript with 
union(b"commonware/handshake", app_namespace). Updated handshake tests and benches to pass the new argument. Breaking change: handshake::Context::new signature adds app_namespace.

### DIFF
--- a/cryptography/src/handshake.rs
+++ b/cryptography/src/handshake.rs
@@ -474,7 +474,7 @@ mod test {
 
     #[cfg(feature = "arbitrary")]
     mod conformance {
-        use super::{Ack, Syn, SynAck};
+        use super::*;
         use commonware_codec::conformance::CodecConformance;
 
         commonware_conformance::conformance_tests! {


cryptography/src/bls12381/dkg2.rs
```diff
@@ -2123,7 +2123,7 @@ mod test_plan {
                     rounds.push(round);
                     Ok(ControlFlow::Continue(()))
                 })?;
-                let plan = Plan {
+                let plan = Self {
```
cronokirby, line 2221: Let's move this out of the diff

cryptography/src/handshake.rs
```diff
@@ -184,21 +185,25 @@ pub struct Context<S, P> {
...
     ) -> Self {
         Self {
             current_time: current_time_ms,
             ok_timestamps,
             my_identity,
             peer_identity,
+            app_namespace,
         }
     }
 }
```
cronokirby: Great idea!

The change I'd like to see is to:

- take in an `&Transcript` as the first argument,
- store `Transcript` inside of `Context`,
- initialize this by calling `fork` on the first argument, using the file namespace as the string.

Then stream would pass in `&Transcript::new(config.namespace)`

cronokirby:
- Breaks the API to create a handshake,
- Streams will not be compatible cross version.

patrick-ogrady:
>Streams will not be compatible cross version.

Would make `breaking-format` imo

cronokirby:
> > Streams will not be compatible cross version.
> 
> Would make `breaking-format` imo

Yeah I was hesitating about that

phrwlk:
Hope I didn't miss anything

cryptography/src/handshake.rs
```diff
@@ -367,22 +374,24 @@ mod test {
 
         let (d_state, msg1) = dial_start(
```
patrick-ogrady, line 402: Can we add a test where dial/listen configured with different transcripts don't connect?

patrick-ogrady (dismissed):

PR #2408, opened 2025-12-05
[workspace] Remove `commonware-deployer`

clabby:
## Overview

Removes `commonware-deployer` from the workspace - to be moved to the https://github.com/exowarexyz organization.


clabby:
to revisit soon

PR #2409, opened 2025-12-05
[Storage] Rename `adb` to `qmdb`

danlaine:
Keeps the text of the blog posts the same but updates links from `adb` to `qmdb`.


docs/blogs/qmdb-any.html
```diff
@@ -11,10 +11,10 @@
     <meta name="author" content="Roberto Bayardo, Patrick O'Grady">
     <meta name="keywords" content="commonware, open source, common goods, software, internet, ownership, trust, blockchain, decentralization, crypto">
 
-    <meta property="og:url" content="https://commonware.xyz/blogs/adb-any.html" />
+    <meta property="og:url" content="https://commonware.xyz/blogs/qmdb-any.html" />
```
patrick-ogrady: We should keep the same.

patrick-ogrady: Otherwise it'll break links on X, etc.

patrick-ogrady: Was just suggesting we change the docs.rs and github links basically 👍

storage/src/qmdb/mod.rs
```diff
@@ -1,4 +1,4 @@
-//! A collection of authenticated databases (ADB).
+//! A collection of authenticated databases.
```
patrick-ogrady: nit: `A collection of authenticated databases inspired by QMDB (Quick Merkle Database).`

patrick-ogrady: Add an `Acknowledgements` section to the end (like we do in cryptography) and add a link to https://arxiv.org/abs/2501.05262

docs/blogs/qmdb-any.html
danlaine: Fixed

storage/src/qmdb/mod.rs
danlaine: Done

patrick-ogrady (approved):

PR #2411, opened 2025-12-05
[Temp] Certify Branch 3rd try

BrendanChou:


consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -314,7 +355,42 @@ impl<
...
+        self.append_journal(view, artifact.clone()).await;
+        self.sync_journal(view).await;
+
+        // Inform listeners of successful certification
+        if success {
+            self.reporter
+                .report(Activity::Certification(notarization))
+                .await;
+        }
+    }
```
patrick-ogrady, line 391: As discussed, we should also update the `resolver` here (and remove update during normal `notarization`)

BrendanChou:
Closing in favor of #2565

PR #2412, opened 2025-12-05
[storage/adb/ordered] implement optimized batch update for any::ordered

roberto-bayardo:
Implements an optimized batch updater for ordered any databases. This does not yet affect the current-variant dbs.

Includes a new fuzz test that exercises random batch updates, verifying both the end map state & resulting key spans are as expected.

It seems that the benefits from this come mostly when a batch creates a lot of new keys that do not have other existing keys between them in the key space, since non-batched updates would result in 2 update operations per create (one for the new key and one for updating the previous) whereas the batching algorithm determines that only 1 update per create is needed when the previous key is also a newly created key.

Works towards: https://github.com/commonwarexyz/monorepo/issues/2223

### DIFF
--- a/storage/src/qmdb/any/ordered/mod.rs
+++ b/storage/src/qmdb/any/ordered/mod.rs
@@ -1137,14 +1137,14 @@ where
         let mut possible_previous = BTreeMap::new();
 
         // We divide keys in the batch into three disjoint sets:
-        // - `deleted`
-        // - `created`
-        // - `updated`
-        // For deleted keys only, we immediately update the log and snapshot during this process.
+        //   - `deleted`
+        //   - `created`
+        //   - `updated`
+        //
+        // Populate the the deleted and updated sets, and for deleted keys only, immediately update
+        // the log and snapshot.
         let mut deleted = Vec::new();
         let mut updated = BTreeMap::new();
-        let mut created = BTreeMap::new();
-
         for (op, old_loc) in (results.into_iter()).zip(locations) {
             let key = op.key().expect("updates should have a key").clone();
             let key_data = op.into_key_data().expect("updates should have key data");
@@ -1176,6 +1176,7 @@ where
         }
 
         // Any key remaining in `mutations` must be a new key so move it to the created map.
+        let mut created = BTreeMap::new();
         for (key, value) in mutations {
             let Some(value) = value else {
                 continue; // can happen from attempt to delete a non-existent key


storage/src/qmdb/any/ordered/mod.rs
```diff
@@ -1078,13 +1081,220 @@ impl<
...
+            let next_key = find_next_key(&key, &possible_next);
+            update_known_loc(&mut self.snapshot, &key, loc, new_loc);
+
+            let op = C::Item::new_update(key.clone(), value.clone(), next_key.clone());
+            self.log.append(op).await?;
+            self.steps += 1;
+            already_updated.insert(key);
+        }
+
+        // Create each new key, and update its previous key if it hasn't already been updated.
```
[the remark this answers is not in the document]
roberto-bayardo: "previous key" is analagous to "next key"

storage/src/qmdb/any/ordered/mod.rs
```diff
@@ -30,7 +35,7 @@ type Value<T> = <T as Keyed>::Value;
 type AuthenticatedLog<E, C, H, S = Clean<DigestOf<H>>> = authenticated::Journal<E, C, H, S>;
 
 /// Type alias for a location and its associated key data.
-type LocatedKey<K, V> = (Location, KeyData<K, V>);
+type LocatedKey<K, V> = Option<(Location, KeyData<K, V>)>;
```
danlaine, line 38: ❤️

storage/src/qmdb/any/ordered/mod.rs
```diff
@@ -1078,13 +1081,231 @@ impl<
...
+        .clone()
+}
+
+/// Returns the previous key to `key` within `possible_previous`. The result will "cycle around"
+/// to the last key if `key` is the first key.
+///
+/// # Panics
+///
+/// Panics if `possible_previous` is empty.
+fn find_prev_key<'a, K: Ord, V>(key: &K, possible_previous: &'a BTreeMap<K, V>) -> (&'a K, &'a V) {
```
danlaine, line 1093: Should we have this return an owned `K` and `V` to match `find_next_key`? Both callers of this method clone the returned values.

storage/src/qmdb/any/ordered/mod.rs
```diff
@@ -1078,13 +1081,231 @@ impl<
...
+            possible_next.insert(key);
+        }
+
+        // Complete the `possible_previous` and `possible_next` sets by including entries from the
+        // previous _translated_ key for any created or deleted key.
+        let mut locations = Vec::new();
+        for key in deleted.iter().chain(created.keys()) {
+            let iter = self.snapshot.prev_translated_key(key);
+            locations.extend(iter.copied());
+        }
```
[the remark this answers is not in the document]
roberto-bayardo: the "find_prev_key" helper function used by `write_batch` handles the fallback.

storage/src/qmdb/any/ordered/mod.rs
```diff
@@ -1078,13 +1081,231 @@ impl<
...
+            let key = op.key().expect("updates should have a key").clone();
+            let key_data = op.into_key_data().expect("updates should have key data");
+            possible_next.insert(key_data.next_key);
+            possible_previous.insert(key, (key_data.value, old_loc));
+        }
+
+        // Remove deleted keys from the possible_* sets.
+        for key in deleted.iter() {
+            possible_previous.remove(key);
+            possible_next.remove(key);
```
danlaine, line 1215: In the loop above we're adding the `deleted` elements into `possible_*` only to remove them here

storage/src/qmdb/any/ordered/mod.rs
```diff
@@ -1078,13 +1081,231 @@ impl<
...
+        let mut updated = BTreeMap::new();
+        let mut created = BTreeMap::new();
+
+        for (op, old_loc) in (results.into_iter()).zip(locations) {
+            let key = op.key().expect("updates should have a key").clone();
+            let key_data = op.into_key_data().expect("updates should have key data");
+
+            let Some(update) = mutations.remove(&key) else {
+                // Due to translated key collisions, we may look up keys that aren't in the
+                // mutations set. Any colliding key could potentially be a previous key of some new
```
danlaine: Could it also be a next key for some new or deleted key?

storage/src/qmdb/any/ordered/mod.rs
```diff
@@ -1078,13 +1081,231 @@ impl<
...
+                // or deleted key.
+                possible_previous.insert(key, (key_data.value, old_loc));
+                continue;
+            };
+
+            if let Some(value) = update {
+                // This is an update of an existing key.
+                updated.insert(key.clone(), (value, old_loc));
+                possible_next.insert(key_data.next_key.clone());
```
danlaine: Could this also be a previous key for some key?

roberto-bayardo: If it is, it will already get inserted by the other logic that inserts possible-previous keys for all newly created & deleted keys.

roberto-bayardo:
> I don't totally understand the `possible_*` logic but I might just need to stare at it a bit longer

Yes this is the hard part of this change beacause it's quite nuanced. We need to make sure for every newly created key and every deleted key, the correct "next key" in the keyspace is in possible_next, and the correct previous key for it is in possible_previous.  *and* that there are no inactive keys within it that may prevent looking up the correct one.

[edit] actually that's not quite accurate, I also put the next_key values for updated keys in there, since we need access to them when we re-write the key in the log with its new value. (We could propagate those specific next_keys through in a different way, but this is simpler.). I updated the comments to better reflect this.

storage/src/qmdb/any/ordered/mod.rs
roberto-bayardo: restult of find_next_key is always used, whereas the result of find_prev_key might be discarded (so a clone would be wasteful).  I've restructured the code a bit to make this more clear.

storage/src/qmdb/any/ordered/mod.rs
roberto-bayardo: The loop above is inserting the "previous translated keys" to each deleted element, not the deleted elements themselves.

storage/src/qmdb/any/ordered/mod.rs
roberto-bayardo: If it is, the other logic for populating possible_next *should* include it (but I will double check).

storage/src/qmdb/any/ordered/mod.rs
```diff
@@ -1078,13 +1081,230 @@ impl<
...
+        for (op, old_loc) in (results.into_iter()).zip(locations) {
+            let key = op.key().expect("updates should have a key").clone();
+            let key_data = op.into_key_data().expect("updates should have key data");
+
+            let Some(update) = mutations.remove(&key) else {
+                // Due to translated key collisions, we may look up keys that aren't in the
+                // mutations set. Any colliding key could potentially be a previous key of some new
+                // or deleted key.
+                possible_previous.insert(key, (key_data.value, old_loc));
+                continue;
```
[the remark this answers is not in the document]
roberto-bayardo: Good point. More critically though, newly created keys are not being put into possible_previous, which seems problematic. Looking into this.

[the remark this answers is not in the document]
roberto-bayardo: Actually newly created keys don't need to be in possible_previous since any newly created_key will be inserted with the proper next_key value regardless. And any previous key that needs to be updated because of some created key will be in the set.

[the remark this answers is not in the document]
roberto-bayardo: Need to add a test case for "key created between two colliding keys" before resolving this.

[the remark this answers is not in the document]
roberto-bayardo: unit test added that captures this scenario & confirms it's fixed.

roberto-bayardo:
Updated the fuzz test to verify all spans, which catches the edge case identified by bug bot, and also detected another tricky one (where we need to cycle-back to the last key when trying to find previous keys of some updated key). Both are fixed now and fuzz test seems happy so far.  Added regression tests for both these edge cases.

storage/src/qmdb/any/ordered/mod.rs
```diff
@@ -1373,4 +1592,82 @@ mod test {
...
+            assert_eq!(db.get(&mid_key).await.unwrap().unwrap(), val);
+
+            let span1 = db.get_span(&preceeding_key).await.unwrap().unwrap();
+            assert_eq!(span1.1.next_key, mid_key);
+            let span2 = db.get_span(&mid_key).await.unwrap().unwrap();
+            assert_eq!(span2.1.next_key, preceeding_key);
+
+            db.destroy().await.unwrap();
+        });
+    }
```
patrick-ogrady, line 1653: Claude wants to add this (although it should be hit by fuzzing):

```
    /// Builds a db with three keys A < B < C, then batch-deletes B. Verifies that A's next_key is
    /// correctly updated to C (skipping the deleted B).
    #[test_traced("WARN")]
    fn test_ordered_any_batch_delete_middle_key() {
        let executor = Runner::default();
            let mut db = open_fixed_db(context.clone()).await;

            let key_a = FixedBytes::from([0x11u8; 4]);
            let key_b = FixedBytes::from([0x22u8; 4]);
            let key_c = FixedBytes::from([0x33u8; 4]);
            let val = Sha256::fill(1u8);

            // Create three keys in order: A -> B -> C -> A (circular)
            db.create(key_a.clone(), val).await.unwrap();
            db.create(key_b.clone(), val).await.unwrap();
            db.create(key_c.clone(), val).await.unwrap();

            // Verify initial spans
            let span_a = db.get_span(&key_a).await.unwrap().unwrap();
            assert_eq!(span_a.1.next_key, key_b);
            let span_b = db.get_span(&key_b).await.unwrap().unwrap();
            assert_eq!(span_b.1.next_key, key_c);
            let span_c = db.get_span(&key_c).await.unwrap().unwrap();
            assert_eq!(span_c.1.next_key, key_a);

            // Batch-delete the middle key B
            let mut batch = db.start_batch();
            batch.delete(key_b.clone()).await.unwrap();
            db.write_batch(batch.into_iter()).await.unwrap();

            // Verify B is deleted
            assert!(db.get(&key_b).await.unwrap().is_none());

            // Verify A's next_key is now C (not B)
            let span_a = db.get_span(&key_a).await.unwrap().unwrap();
            assert_eq!(span_a.1.next_key, key_c, "A's next_key should skip deleted B");

            // Verify C's next_key is still A
            let span_c = db.get_span(&key_c).await.unwrap().unwrap();
            assert_eq!(span_c.1.next_key, key_a);

            db.destroy().await.unwrap();
        });
    }
    ```

File: storage/src/qmdb/any/ordered/mod.rs [hunk 10]
```diff
@@ -1078,13 +1081,229 @@ impl<
     }
 }
 
+/// Returns the next key to `key` within `possible_next`. The result will "cycle around" to the
+/// first key if `key` is the last key.
+///
+/// # Panics
+///
+/// Panics if `possible_next` is empty.
+fn find_next_key<K: Ord + Clone>(key: &K, possible_next: &BTreeSet<K>) -> K {
```
patrick-ogrady, line 1074: nit: for consistency, return `&'a K` like `find_prev_key`?

storage/src/qmdb/any/ordered/mod.rs
```diff
@@ -1078,13 +1081,229 @@ impl<
...
+
+        // We divide keys in the batch into three disjoint sets:
+        // - `deleted`
+        // - `created`
+        // - `updated`
+        // For deleted keys only, we immediately update the log and snapshot during this process.
+        let mut deleted = Vec::new();
+        let mut updated = BTreeMap::new();
+        let mut created = BTreeMap::new();
+
```
patrick-ogrady: nit: remove `\n`

storage/src/qmdb/any/ordered/mod.rs
```diff
@@ -1078,13 +1081,229 @@ impl<
...
+        for (op, old_loc) in (results.into_iter()).zip(locations) {
+            let key = op.key().expect("updates should have a key").clone();
```
patrick-ogrady, line 1149: I suppose we clone here because we can't call into_key_data() if not? If that isn't an issue, I think we should try to use a reference until cloning in the `update ={}` block below.

storage/src/qmdb/any/ordered/mod.rs
```diff
@@ -1078,13 +1081,229 @@ impl<
...
+}
+
 impl<E, C, I, H> Batchable for IndexedLog<E, C, I, H>
 where
     E: Storage + Clock + Metrics,
     C: MutableContiguous<Item: Operation>,
     I: Index<Value = Location>,
     H: Hasher,
 {
+    async fn write_batch(
```
patrick-ogrady, line 1113: Greatly appreciate the comments here.

storage/fuzz/fuzz_targets/qmdb_ordered_batching.rs
```diff
@@ -0,0 +1,197 @@
...
+
+#[derive(Arbitrary, Debug)]
+struct FuzzInput {
+    operations: Vec<QmdbOperation>,
+}
+
+const PAGE_SIZE: usize = 555;
+const PAGE_CACHE_SIZE: usize = 100;
+
+fn fuzz(data: FuzzInput) {
```
patrick-ogrady, line 43: :pray:

storage/src/qmdb/any/ordered/mod.rs
```diff
@@ -1373,4 +1592,82 @@ mod test {
...
+
+            let span1 = db.get_span(&preceeding_key).await.unwrap().unwrap();
+            assert_eq!(span1.1.next_key, mid_key);
+            let span2 = db.get_span(&mid_key).await.unwrap().unwrap();
+            assert_eq!(span2.1.next_key, preceeding_key);
+
+            db.destroy().await.unwrap();
+        });
+    }
 }
```
patrick-ogrady, line 1757: Can you create a determinism test on batches similar to how we have in `archive` (with all the sorting/dedup seems a useful sanity check: https://github.com/commonwarexyz/monorepo/blob/930deb3a98c937bf8ad0845d7ee2732ab46613a0/storage/src/archive/mod.rs#L691-L707)

storage/src/qmdb/any/ordered/mod.rs
```diff
@@ -1373,4 +1592,82 @@ mod test {
...
+            assert_eq!(db.get(&preceeding_key).await.unwrap().unwrap(), val);
+            assert_eq!(db.get(&mid_key).await.unwrap().unwrap(), val);
+
+            let span1 = db.get_span(&preceeding_key).await.unwrap().unwrap();
+            assert_eq!(span1.1.next_key, mid_key);
+            let span2 = db.get_span(&mid_key).await.unwrap().unwrap();
+            assert_eq!(span2.1.next_key, preceeding_key);
+
+            db.destroy().await.unwrap();
+        });
```
patrick-ogrady, line 1652: The LLMs also recommend:

```
   /// Exercises multiple mutations of the same key within a single batch to ensure the final
    /// mutation wins and bookkeeping is consistent.
    #[test_traced("WARN")]
    fn test_ordered_any_batch_double_touch_same_key() {
        let executor = Runner::default();
            let key_a = FixedBytes::from([0x01u8; 4]);
            let key_b = FixedBytes::from([0x02u8; 4]);
            let val_a1 = Sha256::fill(1u8);
            let val_a2 = Sha256::fill(2u8);
            let val_b = Sha256::fill(3u8);

            // Update then delete the same key in a single batch.
            let mut db = open_variable_db(context.clone()).await;
            db.update(key_a.clone(), val_a1).await.unwrap();
            db.update(key_b.clone(), val_b).await.unwrap();
            let base_op_count = db.op_count();

            let mut batch = db.start_batch();
            batch.update(key_a.clone(), val_a2).await.unwrap();
            batch.delete(key_a.clone()).await.unwrap();
            db.write_batch(batch.into_iter()).await.unwrap();

            assert!(db.get(&key_a).await.unwrap().is_none());
            assert_eq!(db.get(&key_b).await.unwrap(), Some(val_b));
            // Delete of an existing key should add exactly two ops (delete + span update).
            assert_eq!(db.op_count(), base_op_count + 2);
            let span_b = db.get_span(&key_b).await.unwrap().unwrap();
            assert_eq!(span_b.1.next_key, key_b);
            db.destroy().await.unwrap();

            // Delete then update the same key in a batch should leave the key active with the
            // updated value.
            let mut db = open_variable_db(context.clone()).await;
            db.update(key_a.clone(), val_a1).await.unwrap();
            db.update(key_b.clone(), val_b).await.unwrap();
            let base_op_count = db.op_count();

            let mut batch = db.start_batch();
            batch.delete(key_a.clone()).await.unwrap();
            batch.update(key_a.clone(), val_a2).await.unwrap();
            db.write_batch(batch.into_iter()).await.unwrap();

            assert_eq!(db.get(&key_a).await.unwrap(), Some(val_a2));
            assert_eq!(db.get(&key_b).await.unwrap(), Some(val_b));
            // Updating an existing key in a batch should add one op.
            assert_eq!(db.op_count(), base_op_count + 1);
            let span_a = db.get_span(&key_a).await.unwrap().unwrap();
            assert_eq!(span_a.1.next_key, key_b);
            let span_b = db.get_span(&key_b).await.unwrap().unwrap();
            assert_eq!(span_b.1.next_key, key_a);
            db.destroy().await.unwrap();
        });
    }

    /// Batch create/delete cases where the deleted key is the previous key of a newly created key,
    /// and vice-versa.
    #[test_traced("WARN")]
    fn test_ordered_any_batch_create_delete_prev_links() {
        let executor = Runner::default();
            let key1 = FixedBytes::from([0x10u8, 0x00, 0x00, 0x00]);
            let key2 = FixedBytes::from([0x20u8, 0x00, 0x00, 0x00]);
            let key3 = FixedBytes::from([0x30u8, 0x00, 0x00, 0x00]);
            let val1 = Sha256::fill(1u8);
            let val2 = Sha256::fill(2u8);
            let val3 = Sha256::fill(3u8);

            // Delete the previous key of a newly created key.
            let mut db = open_variable_db(context.clone()).await;
            db.update(key1.clone(), val1).await.unwrap();
            db.update(key3.clone(), val3).await.unwrap();

            let mut batch = db.start_batch();
            batch.delete(key1.clone()).await.unwrap();
            batch.create(key2.clone(), val2).await.unwrap();
            db.write_batch(batch.into_iter()).await.unwrap();

            assert!(db.get(&key1).await.unwrap().is_none());
            assert_eq!(db.get(&key2).await.unwrap(), Some(val2));
            assert_eq!(db.get(&key3).await.unwrap(), Some(val3));
            let span2 = db.get_span(&key2).await.unwrap().unwrap();
            assert_eq!(span2.1.next_key, key3);
            let span3 = db.get_span(&key3).await.unwrap().unwrap();
            assert_eq!(span3.1.next_key, key2);
            db.destroy().await.unwrap();

            // Create a key that becomes the previous key of a concurrently deleted key.
            let mut db = open_variable_db(context.clone()).await;
            db.update(key1.clone(), val1).await.unwrap();
            db.update(key3.clone(), val3).await.unwrap();

            let mut batch = db.start_batch();
            batch.create(key2.clone(), val2).await.unwrap();
            batch.delete(key3.clone()).await.unwrap();
            db.write_batch(batch.into_iter()).await.unwrap();

            assert_eq!(db.get(&key1).await.unwrap(), Some(val1));
            assert_eq!(db.get(&key2).await.unwrap(), Some(val2));
            assert!(db.get(&key3).await.unwrap().is_none());
            let span1 = db.get_span(&key1).await.unwrap().unwrap();
            assert_eq!(span1.1.next_key, key2);
            let span2 = db.get_span(&key2).await.unwrap().unwrap();
            assert_eq!(span2.1.next_key, key1);
            db.destroy().await.unwrap();
        });
    }
    ```

storage/src/qmdb/any/ordered/mod.rs
roberto-bayardo: One of the new tests I added to the generic test_batch already tests batch-deleting (and batch-creating) keys between other keys, though without verifying the spans, since it's also applied to the non-ordered case. Given the spans are another tricky part of all this, I suppose it makes sense to add this too.

storage/src/qmdb/any/ordered/mod.rs
roberto-bayardo: All the test cases run within test_batch check for determinism already 👍 
https://github.com/commonwarexyz/monorepo/blob/8c02cafcf4349a8e270c4bb5ee3b8cb2723f4753/storage/src/qmdb/store/batch.rs#L195

storage/src/qmdb/any/ordered/mod.rs
roberto-bayardo: done

storage/src/qmdb/any/ordered/mod.rs
roberto-bayardo: Added the second one. First one is redundant given the tests in batch.rs.

storage/src/qmdb/any/ordered/mod.rs
roberto-bayardo: The next_key is always consumed by the caller, so I thought it was better to just clone it before returning it and avoid the lifetime ugliness.

storage/src/qmdb/any/ordered/mod.rs
roberto-bayardo: Right, into_key_data consumes op and invalidates any references.  There is no "into key & key_data" or similar that might allow us to avoid this (though we could add it...)

storage/src/qmdb/any/ordered/mod.rs
roberto-bayardo: Added a comment below this to clarify the loop is populating only the first two of those sets.

danlaine (approved):

PR #2414, opened 2025-12-05
[p2p] Optionally self-limit outbound message rate

clabby:
## Overview

Handles outbound rate limiting at the `Sender` level in `discovery` and `lookup`, making per-component rate limiters redundant. Outbound rate limiting can be toggled off in the configuration if that behavior is not desired.

### DIFF
--- a/p2p/src/authenticated/discovery/channels.rs
+++ b/p2p/src/authenticated/discovery/channels.rs
@@ -106,13 +106,12 @@ where
             self.peer_subscription.as_mut().unwrap()
         };
 
-        let rate_limiter = self.rate_limiter.lock().await;
-
         // Attempt to update known peers if there's a new update, but do not
         // wait for one.
         //
         // When the subscription is first created, it is guaranteed to have
         // the initial list of peers ready immediately.
+        let rate_limiter = self.rate_limiter.lock().await;
         if let Some(peers) = subscription.next().now_or_never().flatten() {
             self.known_peers = peers;
 
@@ -132,7 +131,6 @@ where
             .into_iter()
             .filter(|peer| rate_limiter.check_key(peer).is_ok())
             .collect();
-
         drop(rate_limiter);
 
         // If no recipients are allowed, short-circuit and signal that no peers could
--- a/p2p/src/authenticated/discovery/mod.rs
+++ b/p2p/src/authenticated/discovery/mod.rs
@@ -755,8 +755,8 @@ mod tests {
                     break;
                 }
 
-                // Ensure we don't rate limit outbound sends; Skip past
-                // the rate limit and try again.
+                // Ensure we don't rate limit outbound sends while
+                // waiting for peers to connect
                 context.sleep(Duration::from_mins(1)).await
             }
 
--- a/p2p/src/authenticated/lookup/channels.rs
+++ b/p2p/src/authenticated/lookup/channels.rs
@@ -106,13 +106,12 @@ where
             self.peer_subscription.as_mut().unwrap()
         };
 
-        let rate_limiter = self.rate_limiter.lock().await;
-
         // Attempt to update known peers if there's a new update, but do not
         // wait for one.
         //
         // When the subscription is first created, it is guaranteed to have
         // the initial list of peers ready immediately.
+        let rate_limiter = self.rate_limiter.lock().await;
         if let Some(peers) = subscription.next().now_or_never().flatten() {
             self.known_peers = peers;
 
@@ -132,7 +131,6 @@ where
             .into_iter()
             .filter(|peer| rate_limiter.check_key(peer).is_ok())
             .collect();
-
         drop(rate_limiter);
 
         // If no recipients are allowed, short-circuit and signal that no peers could
--- a/p2p/src/authenticated/lookup/mod.rs
+++ b/p2p/src/authenticated/lookup/mod.rs
@@ -683,8 +683,8 @@ mod tests {
                     break;
                 }
 
-                // Ensure we don't rate limit outbound sends; Skip past
-                // the rate limit and try again.
+                // Ensure we don't rate limit outbound sends while
+                // waiting for peers to connect
                 context.sleep(Duration::from_mins(1)).await
             }
 


p2p/src/utils/requester/requester.rs
```diff
@@ -143,12 +131,6 @@ impl<E: Clock + GClock + Rng + Metrics, P: PublicKey> Requester<E, P> {
                 continue;
             }
 
-            // Check if rate limit is exceeded (and update rate limiter if not)
-            if let Err(limit) = self.rate_limiter.check_key(participant) {
```
patrick-ogrady, line 193: I think this change would mean the first request would always get routed to the "best" peer (no shuffling). This means we'd get bottlenecked by their rate limit on all new requests.

clabby: Updated to retain the old requester logic and just remove the explicit rate limits in the reshare example. That rate limit in the requester now just serves as a way to spill over requests into lower-priority peers if a bunch of requests are sent at once if the new `rate_limit_outbound` flag is set, but felt better to retain the old behavior rather than exposing the rate limited status of the peers' outbound channel in discovery/lookup.

p2p/src/authenticated/discovery/actors/peer/actor.rs
```diff
@@ -159,11 +211,13 @@ impl<E: Spawner + Clock + ReasonablyRealtime + Rng + CryptoRng + Metrics, C: Pub
                     },
                     msg_high = self.high.next() => {
                         let msg = Self::validate_outbound_msg(msg_high, &rate_limits)?;
+                        Self::apply_outbound_rate_limit(&context, &outbound_rate_limits, &self.outbound_rate_limited, &peer, msg.channel).await;
```
patrick-ogrady: What I had in mind here was instead some sort of feedback to indicate we should backoff. Sleeping here is pretty suboptimal (as it blocks all messages sent across all channels when we exceed our limit -> whereas when we have access to the governor upstream we can avoid getting in this situation and keep sending on other channels).

patrick-ogrady: For example, look at the orchestrator fetch boundary finalization case. If we got a handful of messages from the peer, we may accidentally clog up our connection to them (whereas before we'd just skip if we were already at the RL).

clabby: I've moved things around a good bit - rather than the peer actors handling rate limiting behind the scenes, each `Sender` in `lookup` / `discovery` now holds onto a `RateLimiter`. When a `Sender::send` call is made and any of the recipients are rate limited, they are excluded from the send (and that is reflected in the returned vector of peers.)

The rate limiting logic in `requester` is retained because it serves a dual purpose (spilling over requests into lower-priority peers) + the `Sender` is decoupled.

p2p/src/authenticated/discovery/mod.rs
```diff
@@ -689,9 +690,11 @@ mod tests {
         });
     }
 
-    #[test_traced]
-    #[should_panic(expected = "no messages should be rate limited")]
-    fn test_rate_limiting() {
+    #[rstest]
+    #[should_panic]
+    #[case::outbound_disabled_triggers_inbound(false)]
+    #[case::outbound_enabled_prevents_inbound(true)]
```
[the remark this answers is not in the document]
clabby: This isn't the case; it only renders on the case it annotates

p2p/src/authenticated/lookup/channels.rs
```diff
@@ -64,14 +83,57 @@ impl<P: PublicKey> crate::Sender for Sender<P> {
             return Err(Error::MessageTooLarge(message_len));
         }
 
-        // Wait for messenger to let us know who we sent to
+        // Get the concrete list of peers to send to
+        let peers: Vec<Self::PublicKey> = match recipients {
+            Recipients::One(peer) => vec![peer],
+            Recipients::Some(peers) => peers,
+            Recipients::All => self.messenger.connected().await,
```
patrick-ogrady: I think requesting connected() on each send would be pretty large overhead 🤔 .

The right way to do this imho is to have senders listen on `subscribe()` and just mark all possible (don't worry about who is actually online).

p2p/src/authenticated/lookup/channels.rs
```diff
@@ -64,14 +83,57 @@ impl<P: PublicKey> crate::Sender for Sender<P> {
             return Err(Error::MessageTooLarge(message_len));
         }
 
-        // Wait for messenger to let us know who we sent to
+        // Get the concrete list of peers to send to
+        let peers: Vec<Self::PublicKey> = match recipients {
```
patrick-ogrady, line 123: Is there a way to check if rate_limited before using (I suppose we could just tailor our logic elsewhere to eagerly attempt send rather than checking then send)?

p2p/src/authenticated/lookup/channels.rs
clabby: This has been updated to use #2459

p2p/src/authenticated/lookup/channels.rs
clabby: Discussed offline; opting to not fork `governor` to add this ability nor add a `bypass_limit` flag to the `Sender::send` API.

examples/reshare/src/dkg/actor.rs
```diff
@@ -101,12 +93,11 @@ pub struct Config<C: Signer, P> {
     pub mailbox_size: usize,
     pub partition_prefix: String,
     pub peer_config: PeerConfig<C::PublicKey>,
-    pub rate_limit: Quota,
```
patrick-ogrady, line 104: 😍

examples/reshare/src/orchestrator/actor.rs
```diff
@@ -220,9 +214,6 @@ where
                 // If we're not in the committee of the latest epoch we know about and we observe another
                 // participant that is ahead of us, send a message on the orchestrator channel to prompt
                 // them to send us the finalization of the epoch boundary block for our latest known epoch.
-                if rate_limiter.check_key(&from).is_err() {
```
patrick-ogrady, line 223: I suppose we would technically be doing more work here now by checking if we need the finalization if the peer is hammering us with messages...

p2p/src/authenticated/discovery/actors/router/actor.rs
```diff
@@ -117,10 +120,12 @@ impl<E: Spawner + Metrics, P: PublicKey> Actor<E, P> {
                         debug!(?peer, "peer ready");
                         self.connections.insert(peer, relay);
                         let _ = channels.send(routing.clone());
+                        self.notify_subscribers().await;
```
patrick-ogrady, line 123: Ah...I think we only need to send this in the existing subscription flow (when a new peer set is registered).

I think sending for actively online is unnecessarily granular.

p2p/src/authenticated/discovery/channels.rs
```diff
@@ -36,6 +69,12 @@ impl<P: PublicKey> crate::Sender for Sender<P> {
     /// If a recipient is offline at the time a message is sent, the message will be dropped.
     /// It is up to the application to handle retries (if necessary).
     ///
+    /// # Rate Limiting
+    ///
+    /// If outbound rate limiting is enabled, recipients that exceed their rate limit
```
patrick-ogrady: I think we should always enable. There is no reason to ever send above what you are permitted (only can be misconfigured).

p2p/src/authenticated/discovery/channels.rs
```diff
@@ -60,14 +100,75 @@ impl<P: PublicKey> crate::Sender for Sender<P> {
...
+        // Filter peers by rate limit, consuming rate tokens only for allowed peers
+        let allowed_peers = if let Some(ref rate_limiter) = self.rate_limiter {
+            let rate_limiter = rate_limiter.lock().await;
+            let filtered = peers
+                .into_iter()
+                .filter(|peer| rate_limiter.check_key(peer).is_ok())
+                .collect::<Vec<_>>();
+
+            // Clean up limiter state
+            rate_limiter.shrink_to_fit();
```
patrick-ogrady: We only need to call this when a subscription update is processed

p2p/src/authenticated/discovery/config.rs
```diff
@@ -117,6 +117,11 @@ pub struct Config<C: Signer> {
 
     /// Quota for peers messages a peer can send us.
     pub allowed_peers_rate: Quota,
+
+    /// Whether or not to rate limit outgoing sends using the same [Quota]
+    /// as inbound sends. This should generally be enabled to prevent
+    /// other peers from rate limiting this instance.
+    pub rate_limit_outbound: bool,
```
patrick-ogrady: See comment about always doing so.

p2p/src/authenticated/discovery/network.rs
```diff
@@ -110,16 +110,21 @@ impl<E: Spawner + Clock + ReasonablyRealtime + Rng + CryptoRng + RNetwork + Metr
...
     ///   without impacting the ability to process messages on other channels.
+    #[allow(clippy::type_complexity)]
     pub fn register(
         &mut self,
         channel: Channel,
         rate: Quota,
         backlog: usize,
     ) -> (
-        channels::Sender<C::PublicKey>,
+        channels::Sender<C::PublicKey, ContextCell<E>>,
```
patrick-ogrady: I think we can just use `E` here (ContextCell implements E)

p2p/src/authenticated/lookup/actors/router/actor.rs
```diff
@@ -116,10 +119,12 @@ impl<E: Spawner + Metrics, P: PublicKey> Actor<E, P> {
                         debug!(?peer, "peer ready");
                         self.connections.insert(peer, relay);
                         let _ = channels.send(routing.clone());
+                        self.notify_subscribers().await;
                     }
                     Message::Release { peer } => {
                         debug!(?peer, "peer released");
                         self.connections.remove(&peer);
+                        self.notify_subscribers().await;
```
patrick-ogrady, line 127: See comment about using `tracker` for this...

p2p/src/authenticated/lookup/actors/tracker/directory.rs
```diff
@@ -48,8 +45,7 @@ pub struct Directory<E: Rng + Clock + GClock + RuntimeMetrics, C: PublicKey> {
     sets: BTreeMap<u64, Set<C>>,
 
     /// Rate limiter for connection attempts.
-    #[allow(clippy::type_complexity)]
-    rate_limiter: RateLimiter<C, HashMapStateStore<C>, E, NoOpMiddleware<E::Instant>>,
```
patrick-ogrady, line 52: So long overdue...

examples/reshare/src/orchestrator/actor.rs
patrick-ogrady: I think we should add an issue to track a `check()` (that doesn't consume a token) to the sender interface.

I don't think this pattern will be an anomaly.

patrick-ogrady: Actually, changed my mind. I think we should offer a `pre_send` that checks_key/returns something we can call send on (which bypasses RL).

clabby: Added an issue for consideration / discussion, since this'd be a fairly invasive change: #2468

p2p/src/authenticated/discovery/actors/router/actor.rs
clabby: Discussed offline, the traffic here should be minimal enough to not warrant plumbing a subscription from the tracker -> the sender.

p2p/src/authenticated/lookup/actors/router/actor.rs
clabby: https://github.com/commonwarexyz/monorepo/pull/2414#discussion_r2612389713

p2p/src/authenticated/discovery/network.rs
clabby: We unfortunately can't coerce into `Sender<_, E>` without `.take()`, but that should be safe - updated.

(There's some ongoing work in upstream rust to allow for coercing `Deref` types into lower-rank types during monomorphization, which'll be super neat if/when it lands.)

p2p/src/authenticated/lookup/network.rs
```diff
@@ -104,16 +104,21 @@ impl<E: Spawner + Clock + ReasonablyRealtime + Rng + CryptoRng + RNetwork + Metr
...
         backlog: usize,
     ) -> (
-        channels::Sender<C::PublicKey>,
+        channels::Sender<C::PublicKey, E>,
         channels::Receiver<C::PublicKey>,
     ) {
-        self.channels.register(channel, rate, backlog)
+        let clock = self
+            .context
+            .with_label(&format!("channel_{channel}"))
```
patrick-ogrady, line 119: Hm...I think I may have messed up the Cell impl to return Self on this.

p2p/src/authenticated/discovery/channels.rs
```diff
@@ -60,14 +97,70 @@ impl<P: PublicKey> crate::Sender for Sender<P> {
...
+            Recipients::Some(peers) => peers,
+            Recipients::All => self.known_peers.clone(),
+        };
+
+        // Filter peers by rate limit, consuming rate tokens only for allowed peers
+        let allowed_peers: Vec<_> = peers
+            .into_iter()
+            .filter(|peer| rate_limiter.check_key(peer).is_ok())
+            .collect();
+
```
patrick-ogrady: `drop(rate_limiter)`?

I think the guard is held until end of scope?

p2p/fuzz/src/lib.rs
```diff
@@ -296,7 +299,7 @@ impl NetworkScheme for Discovery {
 pub struct Lookup;
 
 impl NetworkScheme for Lookup {
-    type Sender = lookup::Sender<ed25519::PublicKey>;
+    type Sender = lookup::Sender<ed25519::PublicKey, Context>;
```
patrick-ogrady: nit: deterministic::Context

p2p/fuzz/src/lib.rs
```diff
@@ -205,7 +208,7 @@ pub trait NetworkScheme: Send + 'static {
 pub struct Discovery;
 
 impl NetworkScheme for Discovery {
-    type Sender = discovery::Sender<ed25519::PublicKey>;
+    type Sender = discovery::Sender<ed25519::PublicKey, Context>;
```
patrick-ogrady: nit: deterministic::Context

patrick-ogrady (commented):
One small issue, otherwise LGTM!

p2p/src/authenticated/discovery/channels.rs
clabby: good looks - yep, we should drop this.

p2p/src/authenticated/lookup/network.rs
clabby: ~~Will put up an issue~~

Actually, returning something different would require breaking `Metrics`' interface. `.take()` is always safe following a clone, so think it's fine 🤷

p2p/src/authenticated/discovery/channels.rs
```diff
@@ -60,14 +97,72 @@ impl<P: PublicKey> crate::Sender for Sender<P> {
...
+        // Attempt to update known peers if there's a new update, but do not
+        // wait for one.
+        //
+        // When the subscription is first created, it is guaranteed to have
+        // the initial list of peers ready immediately.
+        if let Some(peers) = subscription.next().now_or_never().flatten() {
+            self.known_peers = peers;
+
+            // Clean up limiter state
+            rate_limiter.shrink_to_fit();
```
patrick-ogrady, line 119: I created this issue: https://github.com/commonwarexyz/monorepo/issues/2473

patrick-ogrady:
Marking this as `breaking-api` because the send semantics are slightly different.

PR #2415, opened 2025-12-06
[p2p] Consistent `subscribe()`

patrick-ogrady:
`p2p::authenticated::{discovery, lookup}` returned "self" in subscription messages even if not tracked in any set. This PR aligns behavior with `p2p::simulated` and adds a regression test to all implementations to protect from reintroduction.

### DIFF
--- a/p2p/src/simulated/mod.rs
+++ b/p2p/src/simulated/mod.rs
@@ -2720,17 +2720,49 @@ mod tests {
             let mut manager = oracle.manager();
             let mut subscription = manager.subscribe().await;
 
-            // Register a peer set that includes self
+            // Register a peer set that does NOT include self
+            manager
+                .update(1, vec![other_pk.clone()].try_into().unwrap())
+                .await;
+
+            // Receive subscription notification
+            let (id, new, all) = subscription.next().await.unwrap();
+            assert_eq!(id, 1);
+            assert_eq!(new.len(), 1);
+            assert_eq!(all.len(), 1);
+
+            // Self should NOT be in the new set
+            assert!(
+                new.position(&self_pk).is_none(),
+                "new set should not include self"
+            );
+            assert!(
+                new.position(&other_pk).is_some(),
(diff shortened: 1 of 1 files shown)


p2p/src/simulated/mod.rs
```diff
@@ -2575,15 +2574,15 @@ mod tests {
...
-            let mut manager = oracle.manager();
-            let mut subscription = manager.subscribe().await;
-
             // Create peers
             let pk1 = PrivateKey::from_seed(1).public_key();
             let pk2 = PrivateKey::from_seed(2).public_key();
             let pk3 = PrivateKey::from_seed(3).public_key();
 
+            // Subscribe to peer set updates
+            let mut manager = oracle.manager(pk1.clone());
```
patrick-ogrady: its odd that `manager(pk).update` updates for all peers?

p2p/src/simulated/mod.rs
```diff
@@ -2694,4 +2693,52 @@ mod tests {
             assert_eq!(id3, 2);
         });
     }
+
+    #[test]
+    fn test_subscription_includes_self() {
```
patrick-ogrady: This fails on main (we don't include `self` in all)

p2p/src/authenticated/lookup/actors/tracker/directory.rs
```diff
@@ -195,11 +195,12 @@ impl<E: Spawner + Rng + Clock + GClock + RuntimeMetrics, C: PublicKey> Directory
 
     // ---------- Getters ----------
 
-    /// Returns all tracked peers.
+    /// Returns all peers that are part of at least one peer set.
     pub fn tracked(&self) -> Set<C> {
```
patrick-ogrady, line 199: This function is **only** used in subscribe, so don't worry about breaking something else.

It is a question for a future PR if we need to include "me" in peers if not part of an active set.

PR #2418, opened 2025-12-07
fix(utils): correct PrioritySet peek/pop documentation

aso20455:
Fixed misleading documentation for peek() and pop() methods in PrioritySet.
The docs claimed they return "highest priority" item, but they actually returnthe item with the lowest priority value (first in ascending order). This contradicted the struct's own documentation about "priority-ascending order" and could cause integration bugs.


clabby:

PR #2421, opened 2025-12-08
[resolver/p2p] Fix state leaks in retain and clear handlers

andresilva:
Retain and clear handlers only cleaned fetcher state but left `fetch_timers` entries behind. This caused future fetches for the same key to be silently rejected as duplicates, and consumers were never notified that their fetches were cancelled.

### DIFF
--- a/resolver/src/p2p/engine.rs
+++ b/resolver/src/p2p/engine.rs
@@ -249,7 +249,6 @@ impl<
                                     trace!(?key, "updated targets for existing fetch");
                                 }
                             }
-                            assert_eq!(self.fetcher.len(), self.fetch_timers.len());
                         }
                         Message::Cancel { key } => {
                             trace!(?key, "mailbox: cancel");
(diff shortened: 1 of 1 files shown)


resolver/src/p2p/engine.rs
```diff
@@ -261,27 +261,48 @@ impl<
...
-                            self.fetcher.retain(predicate);
-                            let after = self.fetcher.len();
-                            if before == after {
+
+                            // Remove from fetcher
+                            self.fetcher.retain(&predicate);
+
+                            // Clean up timers and notify consumer
+                            let before = self.fetch_timers.len();
+                            let removed = self.fetch_timers.extract_if(|k, _| !predicate(k)).collect::<Vec<_>>();
```
patrick-ogrady, line 270: I agree it is easier to collect all canceled keys from fetch_timers rather than via fetcher.

Issue #2420, opened 2025-12-08
Potential panic at threshold_signature_recover_multiple: bls12381_threshold.rs:435


dnkolegov-ar:
This is just a weakness found during the fuzzing. 

`threshold_signature_recover_multiple` [assumes](monorepo/cryptography/src/bls12381/primitives/ops.rs:485) at least one set of partial signatures and unconditionally `swap_remove(0)`, calling it with an empty many_evals vector immediately panics.

Despite the fact that before calling the function, there is a [check](monorepo/consensus/src/simplex/signing_scheme/bls12381_threshold.rs:435) , a more defensive approach would be to check if the 0-element exists explicitly.

Test:
```
#![no_main]

use arbitrary::{Arbitrary, Unstructured};
use commonware_codec::ReadExt;
use commonware_cryptography::bls12381::primitives::{
    group::{Element, G2, G2_ELEMENT_BYTE_LENGTH},
    ops::threshold_signature_recover_multiple,
    poly::Eval,
    variant::MinPk,
};
use libfuzzer_sys::fuzz_target;

const MAX_GROUPS: usize = 8;
const MAX_PARTIALS_PER_GROUP: usize = 16;
const MAX_THRESHOLD: u32 = 32;
const MAX_CONCURRENCY: usize = 8;

#[derive(Arbitrary, Debug)]
struct PartialSignatureInput {
    index: u32,
    bytes: [u8; G2_ELEMENT_BYTE_LENGTH],
    use_zero: bool,
}

#[derive(Arbitrary, Debug)]
struct SignatureGroupInput {
    entries: Vec<PartialSignatureInput>,
}

#[derive(Arbitrary, Debug)]
struct FuzzInput {
    threshold: u32,
    concurrency: usize,
    groups: Vec<SignatureGroupInput>,
}

fn eval_from_input(input: PartialSignatureInput) -> Eval<G2> {
    let PartialSignatureInput {
        index,
        bytes,
        use_zero,
    } = input;

    let mut cursor = bytes.as_slice();
    let value = G2::read(&mut cursor).unwrap_or_else(|_| {
        if use_zero {
            G2::zero()
        } else {
            G2::one()
        }
    });

    Eval { index, value }
}

fn fuzz(input: FuzzInput) {
    let threshold = input.threshold.clamp(1, MAX_THRESHOLD);
    let concurrency = input.concurrency.clamp(1, MAX_CONCURRENCY);

[...]

PR #2422, opened 2025-12-08
[utils] add NonEmptyVec wrapper

andresilva:
Related (fixed in #2398 but this came out of discussions around that refactor): #2420 

A vec wrapper that is guaranteed to always have at least one element.

### DIFF
--- a/utils/src/vec.rs
+++ b/utils/src/vec.rs
@@ -7,15 +7,9 @@ use bytes::{Buf, BufMut};
 use commonware_codec::{EncodeSize, RangeCfg, Read, Write};
 use core::{
     num::NonZeroUsize,
-    ops::{Deref, DerefMut, Index, IndexMut},
-    slice::SliceIndex,
+    ops::{Deref, DerefMut},
 };
 use thiserror::Error;
-
-#[cfg(not(feature = "std"))]
-type VecIntoIter<T> = alloc::vec::IntoIter<T>;
-#[cfg(feature = "std")]
-type VecIntoIter<T> = std::vec::IntoIter<T>;
 
 /// Errors that can occur when creating a [`NonEmptyVec`].
 #[derive(Error, Debug, PartialEq, Eq)]
@@ -236,20 +230,6 @@ impl<T> AsRef<Vec<T>> for NonEmptyVec<T> {
     }
 }
 
-impl<T, I: SliceIndex<[T]>> Index<I> for NonEmptyVec<T> {
-    type Output = I::Output;
-
-    fn index(&self, index: I) -> &Self::Output {
-        &self.0[index]
-    }
-}
-
-impl<T, I: SliceIndex<[T]>> IndexMut<I> for NonEmptyVec<T> {
-    fn index_mut(&mut self, index: I) -> &mut Self::Output {
-        &mut self.0[index]
-    }
-}
-
 impl<T> From<NonEmptyVec<T>> for Vec<T> {
     fn from(vec: NonEmptyVec<T>) -> Self {
         vec.0
@@ -311,7 +291,7 @@ impl<T> TryFromIterator<T> for NonEmptyVec<T> {
 
 impl<T> IntoIterator for NonEmptyVec<T> {
     type Item = T;
-    type IntoIter = VecIntoIter<T>;
+    type IntoIter = <Vec<T> as IntoIterator>::IntoIter;
 
     fn into_iter(self) -> Self::IntoIter {
         self.0.into_iter()
@@ -414,7 +394,7 @@ impl<'a, T: arbitrary::Arbitrary<'a>> arbitrary::Arbitrary<'a> for NonEmptyVec<T
 /// let v = non_empty_vec![42; @n];
 /// assert_eq!(v.len().get(), 2);
 ///
-/// // Vec form: wrap an existing Vec (caller must ensure non-empty)
+/// // Vec form: wrap an existing Vec (panics if empty)
 /// let vec = vec![1, 2, 3];
 /// let v = non_empty_vec![@vec];
 /// assert_eq!(v.len().get(), 3);


cryptography/src/bls12381/primitives/ops.rs
```diff
@@ -473,7 +473,7 @@ where
 /// each set of partial signatures has the same indices.
 pub fn threshold_signature_recover_multiple<'a, V, I>(
     threshold: u32,
-    mut many_evals: Vec<I>,
+    many_evals: NonEmptyVec<I>,
```
cronokirby: For this case specifically, I think the issue is not that the evals should be non-empty, but instead that the function should return an empty vec if you pass in an empty vec. In other words, if you try and recover zero signatures, you should successfully recover zero signatures.

utils/src/vec.rs
```diff
@@ -0,0 +1,646 @@
...
+
+impl<T> Deref for NonEmptyVec<T> {
+    type Target = [T];
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl<T> DerefMut for NonEmptyVec<T> {
```
cronokirby, line 215: It's very confusing to me that `&mut v` would be mutating the first element, but `&v` would be a slice of the entire vector.

utils/src/vec.rs
```diff
@@ -0,0 +1,646 @@
...
+#[derive(Error, Debug, PartialEq, Eq)]
+pub enum Error {
+    /// The collection was empty.
+    #[error("cannot create NonEmptyVec from empty collection")]
+    Empty,
+}
+
+/// A vector that is guaranteed to contain at least one element.
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct NonEmptyVec<T>(Vec<T>);
```
cronokirby, line 24: I think we could simplify the maintenance burden of a type like this by avoiding trying to have all the methods of Vec.

We should instead have just the method to try and create it from a Vector. I also expect that formost use cases, being able to iterator over the values is the only API you need.

I would try and see if we could get away with not implementing `push`, etc. but if we do, a more scalable approach would be to implement some kind of `mutate_tail` function which gets mutable access to a vec of all elements past the first. (This breaks being able to do `AsRef<[T]>`, so maybe not worth it). Otherwise, you could have a method which lets you do anything to the vec, but then panics if the resulting len is 0. Basically a "you can run a closure, but it better be correct" method.

examples/vrf/src/handlers/wire.rs
```diff
@@ -287,7 +288,7 @@ mod tests {
         let mut public = <MinSig as Variant>::Public::one();
         let scalar = group::Scalar::from_rand(&mut thread_rng());
         public.mul(&scalar);
-        poly::Public::<MinSig>::from(vec![public; T])
+        poly::Public::<MinSig>::from(NonEmptyVec::try_from(vec![public; T]).unwrap())
```
patrick-ogrady: Initial comment: use `non_empty_vec!` everywhere we parse + unwrap immediately?

cryptography/fuzz/fuzz_targets/bls12381_threshold_operations.rs
```diff
@@ -490,7 +491,7 @@ fn fuzz(op: FuzzOperation) {
                     .collect();
                 let _ = threshold_signature_recover_multiple::<MinPk, _>(
                     threshold,
-                    groups_refs,
+                    NonEmptyVec::try_from(groups_refs).unwrap(),
```
patrick-ogrady: ex: `non_empty_vec!`?

utils/src/vec.rs
andresilva: Currently it implements the minimal API for it to be useful without having to convert it back to a vector, in particular it implements some vec methods that have different types due to the invariant, i.e. `first` and `last` no longer return `Option` (if you look at the changes in the bmt code this removes a lot of  unwraps), `len` returns `NonZeroUsize`, `resize` requires `NonZeroUsize` and `remove` returns an `Option` (if you try to remove the last element). I think this is useful because you can lean on the non-empty invariant at compile-time and still have the type be generally usable without having to implement the entire vec API.

utils/src/vec.rs
andresilva: Not sure what you mean, this returns `&mut [T]`.

cryptography/fuzz/fuzz_targets/bls12381_threshold_operations.rs
andresilva: This particular case can't use the macro since the vec is being computed at runtime, but I updated all the other cases.

cryptography/src/bls12381/primitives/ops.rs
andresilva: Updated to early exit with empty vec of signatures.

utils/src/vec.rs
andresilva: I added a `mutate` method for checked operations on the underlying vec.

utils/src/vec.rs
```diff
@@ -0,0 +1,769 @@
...
+/// Creates a [`NonEmptyVec`] containing the given elements.
+///
+/// # Forms
+///
+/// | Syntax | Count type | Guarantee |
+/// |--------|------------|-----------|
+/// | `non_empty_vec![a, b, c]` | - | Compile-time (at least one element required) |
+/// | `non_empty_vec![elem; N]` | const `usize` | Compile-time (N must be const and > 0) |
+/// | `non_empty_vec![elem; NZUsize!(N)]` | [`NZUsize!`] | Runtime (panics if N == 0) |
+/// | `non_empty_vec![elem; @n]` | [`NonZeroUsize`] | Type-safe (n is already non-zero) |
```
andresilva, line 367: I know this will be a point of contention 🙈 An alternative is to create a distinct macro, e.g. `non_empty_vec_nz!`.

utils/src/vec.rs
```diff
@@ -0,0 +1,769 @@
...
+    /// # Panics
+    ///
+    /// Panics if `index >= len`.
+    pub fn remove(&mut self, index: usize) -> Option<T> {
+        if self.0.len() > 1 {
+            Some(self.0.remove(index))
+        } else {
+            None
+        }
+    }
```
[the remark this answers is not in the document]
andresilva: Good bot 🥇

utils/src/vec.rs
patrick-ogrady: yoinks

math/src/poly.rs
```diff
@@ -501,7 +508,9 @@ mod fuzz {
                 .arbitrary_iter::<F>()?
                 .take(size)
                 .collect::<Result<Vec<_>, _>>()?;
-            Ok(Self { coeffs })
+            Ok(Self {
+                coeffs: NonEmptyVec::from_vec_unchecked(coeffs),
+            })
         }
```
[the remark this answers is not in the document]
andresilva: Updated to return `arbitrary::Error::NotEnoughData`.

utils/src/vec.rs
patrick-ogrady: I extended this 😓

math/src/poly.rs
```diff
@@ -497,10 +492,10 @@ mod fuzz {
     impl<'a, F: Arbitrary<'a>> Arbitrary<'a> for Poly<F> {
         fn arbitrary(u: &mut arbitrary::Unstructured<'a>) -> arbitrary::Result<Self> {
             let size = u.arbitrary_len::<F>()?.max(1);
-            let coeffs = u
+            let coeffs = non_empty_vec![@u
                 .arbitrary_iter::<F>()?
```
patrick-ogrady: I messed this up (I believe this can create an empty iter)

patrick-ogrady (changes requested):
I introduced a regression (I think)

patrick-ogrady:
This is not `breaking-format` (we just updated the approach we use for generating arbitrary). Going to manually merge.

math/src/poly.rs
```diff
@@ -496,11 +491,12 @@ mod fuzz {
...
-            let coeffs = u
-                .arbitrary_iter::<F>()?
-                .take(size)
-                .collect::<Result<Vec<_>, _>>()?;
+            let first: F = u.arbitrary()?;
+            let rest: Vec<F> = u.arbitrary()?;
+            let coeffs = iter::once(first)
+                .chain(rest)
+                .try_collect()
+                .expect("has at least one element");
```
cronokirby: We should probably have a method to create a NonEmptyVec infallibly from a head and a tail.

Perhaps, instead, we want to have an Arbitrary impl for NonEmptyVec directly? Then this would be a derive.

utils/src/vec.rs
```diff
@@ -0,0 +1,848 @@
...
+    pub fn new(first: T) -> Self {
+        Self(vec![first])
+    }
+
+    /// Creates a [`NonEmptyVec`] from a [`Vec`] without checking if it's empty.
+    ///
+    /// # Panics
+    ///
+    /// Methods on the returned `NonEmptyVec` will panic if the vector is empty.
+    pub const fn from_unchecked(vec: Vec<T>) -> Self {
```
cronokirby: I feel like it would be better to just add an assertion here?

math/src/poly.rs
andresilva: > We should probably have a method to create a NonEmptyVec infallibly from a head and a tail.

That's `NonEmptyVec::new(head)`, you can just push the tail afterwards. Creating an arbitrary for `NonEmptyVec` makes sense to me.

patrick-ogrady: >Perhaps, instead, we want to have an Arbitrary impl for NonEmptyVec directly? Then this would be a derive.

ooooo good call.

utils/src/vec.rs
```diff
@@ -0,0 +1,848 @@
...
+        self.0
+    }
+
+    /// Returns an iterator over the elements.
+    pub fn iter(&self) -> core::slice::Iter<'_, T> {
+        self.0.iter()
+    }
+
+    /// Returns a mutable iterator over the elements.
+    pub fn iter_mut(&mut self) -> core::slice::IterMut<'_, T> {
```
cronokirby: What's the motivation behind returning the concrete type over an impl Iterator?

utils/src/vec.rs
patrick-ogrady: I think I can remove this and assert in the conversion of the macro.

patrick-ogrady: But yeah, I agree should assert here.

utils/src/vec.rs
```diff
@@ -0,0 +1,860 @@
...
+        let rest: Vec<T> = u.arbitrary()?;
+        let mut vec = Vec::with_capacity(1 + rest.len());
+        vec.push(first);
+        vec.extend(rest);
+        Ok(Self(vec))
+    }
+}
+
+/// Creates a [`NonEmptyVec`] containing the given elements.
+///
```
cronokirby, line 359:
```suggestion
        let mut vec: Vec<T> = u.arbitrary()?;
        vec.insert(0, first);
```

Same linear copy, but cleaner imo.

(Also possible, but would require another conformance regen, to do this more smartly by swapping first and vec[0], then pushing that to the end. Note that you don't want to generate the last element after the vec, because that might lead to unintuitive shrinking, but that could also work)

cronokirby (commented):
Suggested cleanup on the arbitrary impl :)

utils/src/vec.rs
patrick-ogrady: I prefer the look of what is in the PR (and suspect it'll be more memory efficient on average because we set capacity up front) 👍

PR #2427, opened 2025-12-08
[consensus/simplex] Fix fuzzer

dnkolegov-ar:
This PR mainly fixes certificate [handling](https://github.com/commonwarexyz/monorepo/pull/2427/files#diff-99619a1f72561dfc9b5c180c4649e5b96b99ae15d64094ba73c933b411217746R268) that did not work before.

### DIFF
--- a/consensus/fuzz/src/disrupter.rs
+++ b/consensus/fuzz/src/disrupter.rs
@@ -37,7 +37,7 @@ pub struct Disrupter<E: Clock + Spawner + Rng + CryptoRng, S: Scheme> {
     participants: Set<PublicKey>,
     namespace: Vec<u8>,
     fuzz_input: FuzzInput,
-    view: u64,
+    last_vote: u64,
     last_finalized: u64,
     last_nullified: u64,
     last_notarized: u64,
@@ -56,7 +56,7 @@ where
         fuzz_input: FuzzInput,
     ) -> Self {
         Self {
-            view: 0,
+            last_vote: 0,
             last_finalized: 0,
             last_nullified: 0,
             last_notarized: 0,
@@ -211,11 +211,11 @@ where
                 .await;
         }
 
-        let Ok(last_vote) = Vote::<S, Sha256Digest>::read(&mut msg.as_slice()) else {
+        let Ok(vote) = Vote::<S, Sha256Digest>::read(&mut msg.as_slice()) else {
             return;
         };
-        self.view = last_vote.view().get();
-        match last_vote {
+        self.last_vote = vote.view().get();
+        match vote {
             Vote::Notarize(notarize) => {
                 if self.fuzz_input.random_bool() {
                     let mutated = self.mutate_bytes(&msg);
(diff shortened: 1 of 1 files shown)


consensus/fuzz/src/disrupter.rs
```diff
@@ -258,38 +265,20 @@ where
                 .await;
         }
 
-        let cfg = Default::default();
+        let cfg = self.scheme.certificate_codec_config();
```
patrick-ogrady, line 268: Nice catch. Its annoying that calling default there works 🤔

consensus/fuzz/src/disrupter.rs
```diff
@@ -258,38 +265,20 @@ where
...
         let Ok(cert) = Certificate::<S, Sha256Digest>::read_cfg(&mut msg.as_slice(), &cfg) else {
             return;
```
patrick-ogrady, line 270: I suppose we could add a check that anything fails here was mutated?

consensus/fuzz/src/lib.rs
```diff
@@ -32,7 +32,8 @@ pub const EPOCH: u64 = 333;
 
 const PAGE_SIZE: NonZeroUsize = NZUsize!(1024);
 const PAGE_CACHE_SIZE: NonZeroUsize = NZUsize!(10);
-const REQUIRED_CONTAINERS: u64 = 50;
+const MIN_REQUIRED_CONTAINERS: u64 = 10;
+const MAX_REQUIRED_CONTAINERS: u64 = 40;
```
patrick-ogrady: nit: might as well keep at 50

consensus/fuzz/src/disrupter.rs
```diff
@@ -356,7 +347,8 @@ where
                 }
             }
             Message::Nullify => {
-                let round = Round::new(Epoch::new(EPOCH), View::new(self.view));
+                let view = self.random_view(self.view);
+                let round = Round::new(Epoch::new(EPOCH), View::new(view));
```
patrick-ogrady: nit: put random view call inline?

consensus/fuzz/src/disrupter.rs
```diff
@@ -206,14 +214,14 @@ where
         let Ok(vote) = Vote::<S, Sha256Digest>::read(&mut msg.as_slice()) else {
             return;
         };
+        self.view = vote.view().get();
```
patrick-ogrady: This approach could lead view to go backwards (do we want to just allow monotonic increase)?

dnkolegov-ar: I think it is fine for fuzzer. Getting view from certificates is more deterministic

consensus/fuzz/src/lib.rs
dnkolegov-ar: fixed

consensus/fuzz/src/disrupter.rs
dnkolegov-ar: fixed

consensus/fuzz/src/disrupter.rs
patrick-ogrady: Can you change this to `last_vote` if you want to keep logic as-is?

Issue #2258, opened 2025-11-17
[runtime + CI] Error on Breaking Changes


patrick-ogrady:
We should cache message/storage formats in a `txt` file on main and fail a new CI job if any format changes (unless a PR includes a `breaking` label). This won't capture subtle behavior changes (which should be largely ok).

Future Work: https://github.com/commonwarexyz/monorepo/issues/2496
Future Work: https://github.com/commonwarexyz/monorepo/issues/2497

clabby:
Following #2430, we'll break out the existing conformance testing utilities into its own `commonware-conformance` crate for more general use. Beyond just breaking changes to `Codec` impls, a more generic trait should be able to be applied to frame transmission order in p2p, storage structure, etc.

---

As for the codec flavor of conformance testing, the list of crates to cover:
- [x] `commonware-codec`
- [x] `commonware-consensus`
- [x] `commonware-cryptography`
- [ ] `commonware-p2p`
- [ ] `commonware-storage`
- [ ] `commonware-coding`
- [ ] `commonware-resolver`
- [ ] `commonware-collector`

PR #2430, opened 2025-12-08
[codec] Codec conformance tests

clabby:
## Overview

Adds tests that verify `Codec` implementations maintain backward compatibility by comparing encoded output against known-good hash values stored in TOML files. This prevents accidental breaking changes to wire formats.

### Running Conformance Tests

```bash
# Run all codec conformance tests
just test-codec-fixtures

# Regenerate fixtures (use only for INTENTIONAL format changes)
just regenerate-codec-fixtures
```

**WARNING**: Running `just regenerate-codec-fixtures` is effectively a manual approval of a breaking codec change. Only use this when you have intentionally changed the wire format and have verified that the change is correct. This will update the hash values in `codec_conformance.toml` files throughout the repository.

### Adding Conformance Tests for New Types

When creating a new type that implements `Write` (the codec trait), add conformance tests:

#### Step 1: Add `Arbitrary` Implementation

Add an `arbitrary::Arbitrary` impl gated by the `arbitrary` feature flag. Place this near the other trait impls for the type:

```rust
#[cfg(feature = "arbitrary")]
impl arbitrary::Arbitrary<'_> for MyType {
        // ... construct your type using the unstructured data ...
        Ok(my_instance)
    }
}
```

#### Step 2: Add Conformance Test Module

Inside the `#[cfg(test)] mod tests` block, add a `conformance` submodule gated by the `arbitrary` feature:

```rust
#[cfg(test)]
mod tests {
    use super::*;

    // ... other tests ...

    #[cfg(feature = "arbitrary")]
    mod conformance {
        use super::*;

        commonware_codec::conformance_tests! {
            MyType => 1024,
        }
    }
}
```

The number (1024) is the number of test cases to generate and hash together.

#### Step 3: Run Tests to Generate Fixtures

Run `just test-codec-fixtures` to generate the initial hash values. The test framework will automatically add new types to the appropriate `codec_conformance.toml` file.

### How It Works

1. Tests generate `n` deterministic values using `arbitrary`
2. Each value is encoded and all encodings are hashed together
3. The hash is compared against the stored value in `codec_conformance.toml` in each crate root that employs codec conformance testing.
4. Hash mismatches cause test failures (format changed)
5. Missing types are automatically added to the TOML file

## TODO

- [x] Should generated `arbitrary::Arbitrary` types have to be _valid codec_, or _valid instances_ of types?
    - a: for now, just valid codec.
- [x] Use `arbitrary` features over `fuzz` features
- [x] Add `zepter` to CI to enforce feature propagation
- [ ] Conformance tests
    - [x] `commonware-codec`
    - [x] `commonware-consensus`
    - [x] `commonware-cryptography`
    - [ ] `commonware-p2p`
    - [ ] `commonware-storage`
    - [ ] `commonware-coding`
    - [ ] `commonware-resolver`
    - [ ] `commonware-collector`

related https://github.com/commonwarexyz/monorepo/issues/2258

### DIFF
--- a/consensus/codec_conformance.toml
+++ b/consensus/codec_conformance.toml
@@ -52,7 +52,7 @@ hash = "9d2ee4f3c6bf86521c09f8858683de48d6310a987643ee379c143480fc7e9a34"
 
 ["commonware_consensus::simplex::signing_scheme::bls12381_multisig::tests::conformance::Certificate<MinSig>"]
 n_cases = 65536
-hash = "d1ae100f7f4bf0cdc8be5d1007df5810c47bda1fd084ac4e0ceb6978a10467aa"
+hash = "300d04d83f33581f5bce771224f35a4f41a50887e6f83648a43013d8e8292815"
 
 ["commonware_consensus::simplex::signing_scheme::bls12381_threshold::tests::conformance::Seed<MinSig>"]
 n_cases = 65536
@@ -64,11 +64,11 @@ hash = "030caacb9ab46d08601977418808dc5ff729bddb9f6bb2ea685920539b69ecfb"
 
 ["commonware_consensus::simplex::signing_scheme::ed25519::tests::conformance::Certificate"]
 n_cases = 65536
-hash = "5e989bb870ca2ae8185bfc68f03a604f292d60328944b6593c5ceeb956ed0230"
+hash = "eab4ff22bb7ed99fd73107c8b067e39bd25813b65614df95b4b5c46a312557ad"
 
 ["commonware_consensus::simplex::signing_scheme::utils::tests::conformance::Signers"]
 n_cases = 65536
-hash = "3d503d2014be640dbd6a48f0c91d3233f6be81731c73f1ddb5ecd2b93665128d"
+hash = "41ef7a306dd2032d7d1b73fe2799cff6f2e8fb1edcd02a0b9798f18f900a19fa"
 
 ["commonware_consensus::simplex::types::tests::conformance::Activity<Scheme, Sha256Digest>"]
 n_cases = 65536
--- a/consensus/src/simplex/signing_scheme/utils.rs
+++ b/consensus/src/simplex/signing_scheme/utils.rs
@@ -86,7 +86,7 @@ impl Read for Signers {
 #[cfg(feature = "arbitrary")]
 impl arbitrary::Arbitrary<'_> for Signers {
     fn arbitrary(u: &mut arbitrary::Unstructured<'_>) -> arbitrary::Result<Self> {
-        let participants = u.arbitrary_len::<u8>()? % 255;
+        let participants = u.arbitrary_len::<u8>()? % 10;
         let signer_count = u.arbitrary_len::<u8>()?.min(participants);
         let signers = (0..signer_count as u32).collect::<Vec<_>>();
         Ok(Self::from(participants, signers))


codec/src/types/bytes.rs
```diff
@@ -81,4 +81,50 @@ mod tests {
...
+        }
+
+        impl Arbitrary<'_> for Bytes {
+            fn arbitrary(u: &mut arbitrary::Unstructured<'_>) -> arbitrary::Result<Self> {
+                let len = u.arbitrary::<u8>()?;
+                let bytes: Vec<u8> = u
+                    .arbitrary_iter()?
+                    .take(len as usize)
+                    .collect::<Result<Vec<_>, _>>()
+                    .unwrap();
```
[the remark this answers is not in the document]
clabby: This is fine; Internal interface.

.github/workflows/fast.yml
```diff
@@ -282,3 +282,17 @@ jobs:
         run: zepter format features
       - name: Lint feature propagation
         run: zepter run check
+
+  codec-conformance:
```
patrick-ogrady: nit: "conformance"

.github/workflows/fast.yml
```diff
@@ -282,3 +282,17 @@ jobs:
...
+    name: Codec Conformance
```
patrick-ogrady: nit: "Breaking Changes"

consensus/codec_conformance.toml
```diff
@@ -0,0 +1,155 @@
...
+
+["commonware_consensus::ordered_broadcast::types::tests::conformance::Ack<PublicKey, MinPk, Sha256Digest>"]
+n_cases = 1024
+hash = "3707adcc6b87e219fa52fe45f91666ad5f4f5802172ea08c8d0b49441b8962a7"
+
+["commonware_consensus::ordered_broadcast::types::tests::conformance::Activity<PublicKey, MinPk, Sha256Digest>"]
+n_cases = 1024
+hash = "0406b0fa175b2fefc7de9d43a008dc3b02eed8d58fc36a4dc672fc1d67b7a8a0"
+
+["commonware_consensus::ordered_broadcast::types::tests::conformance::Chunk<PublicKey, Sha256Digest>"]
```
patrick-ogrady, line 33: Should we place outside of `tests`?

patrick-ogrady: Given we don't actually check conformance in unit tests, that feels appropriate.

justfile
```diff
@@ -84,3 +84,11 @@ check-features:
 # Fix feature propagation and formatting
 fix-features:
     zepter && zepter format features --fix
+
+# Test codec conformance
+test-codec-fixtures:
+    just test --features arbitrary --profile codec_conformance
```
patrick-ogrady, line 90: nit: Maybe we just call `conformance`? It feels we could implement outside of `codec`?

codec/src/types/btree_map.rs
```diff
@@ -260,4 +260,13 @@ mod tests {
 
         assert_eq!(map1.encode(), map2.encode());
     }
+
+    #[cfg(feature = "arbitrary")]
+    mod conformance {
+        use super::*;
+
+        crate::conformance_tests! {
+            BTreeMap<u32, u32> => 1024,
```
patrick-ogrady: Let's make this a const (65536) and allow it to be optionally overridden.

patrick-ogrady (changes requested):
Very cool

patrick-ogrady:
I think we should change `closes` on the linked issue to `related` (given there are still more types to go)

codec/src/conformance.rs
```diff
@@ -0,0 +1,452 @@
...
+
+/// Compute the conformance hash for a type.
+///
+/// Generates `n_cases` arbitrary values, encodes each one, and hashes
+/// all the encoded bytes together using SHA-256.
+pub fn compute_conformance_hash<T>(n_cases: usize) -> String
+where
+    T: Encode + for<'a> Arbitrary<'a>,
+{
+    let mut hasher = Sha256::new();
```
cronokirby, line 160: We should use the `Transcript` API instead.

clabby: Using the `Transcript` API would require `commonware-codec` to depend on `commonware-cryptography`, unfortunately.

cronokirby: Right, that's fair enough.

It does make me think though, that given this section of `codec` is gated behind a feature flag anyways, if it might not just be simpler to have this be a separate crate.

clabby:
> Also, why add an additional layer of feature gating for arbitrary, versus just having arbitrary be a dev dependency, used for the test? If you're just adding the conformance tests to the crate, that's all you'd need, right?

We need to inherit the arbitrary implementations across crates; i.e. `commonware-consensus` relies on `commonware-cryptography`'s arbitrary impls.

clabby:
> Instead of having the `cfg` hack, what if instead we had the behavior be that it will generate the file if non exists (I think it works this way now)? Then the guidance is to simply remove the file yourself if you want to regenerate it. We avoid needing the whole cfg hackery then.

Either way would work. I don't think the `cfg` is hacky; They're used for build configs 😄

cronokirby:
Another point: I'm not sure if this tests exactly what we want.

For one, we don't actually care if the arbitrary instance generates different values for a given seed. That's not a breaking change in and of itself.

Also, let's say you have an enum, and you add a variant. With how the harness works, this is considered a breaking change, because you're going to see a different hash. But, if you can read old values, that's not really a breaking change. If you want to have protocol evolution, you need the ability to add new variants to messages, which is fine as long as you can read old messages.

Perhaps a better strategy would be to save encoded values of the structure under test, then to verify that you can decode old values, and that when you encode a decoded old value you get the same result. It's fine to generate _new_ values you couldn't before, as long as you can handle old ones.

That definitely blows up the size of the test files though.

If you did that you'd kind of want a means not of generating a random structure, but of specifically generating a random instance of each relevant variant.

cronokirby:
> > Also, why add an additional layer of feature gating for arbitrary, versus just having arbitrary be a dev dependency, used for the test? If you're just adding the conformance tests to the crate, that's all you'd need, right?
> 
> We need to inherit the arbitrary implementations across crates; i.e. `commonware-consensus` relies on `commonware-cryptography`'s arbitrary impls.

mhmm, good point

clabby:
> Another point: I'm not sure if this tests exactly what we want.
> 
> For one, we don't actually care if the arbitrary instance generates different values for a given seed. That's not a breaking change in and of itself.
> 
> Also, let's say you have an enum, and you add a variant. With how the harness works, this is considered a breaking change, because you're going to see a different hash. But, if you can read old values, that's not really a breaking change. If you want to have protocol evolution, you need the ability to add new variants to messages, which is fine as long as you can read old messages.
> 
> Perhaps a better strategy would be to save encoded values of the structure under test, then to verify that you can decode old values, and that when you encode a decoded old value you get the same result. It's fine to generate _new_ values you couldn't before, as long as you can handle old ones.
> 
> That definitely blows up the size of the test files though.
> 
> If you did that you'd kind of want a means not of generating a random structure, but of specifically generating a random instance of each relevant variant.

It's imperfect for some of the reasons you've mentioned here, but it does catch _changes_. The changes may not be breaking (for example, in the case of adding a new variant,) but we may as well opt-into new values that the `Write` impl can produce.

I'd prefer not to save encoded values of the struct for tons of cases; The test files get very very large.

patrick-ogrady:
> > Another point: I'm not sure if this tests exactly what we want.
> > For one, we don't actually care if the arbitrary instance generates different values for a given seed. That's not a breaking change in and of itself.
> > Also, let's say you have an enum, and you add a variant. With how the harness works, this is considered a breaking change, because you're going to see a different hash. But, if you can read old values, that's not really a breaking change. If you want to have protocol evolution, you need the ability to add new variants to messages, which is fine as long as you can read old messages.
> > Perhaps a better strategy would be to save encoded values of the structure under test, then to verify that you can decode old values, and that when you encode a decoded old value you get the same result. It's fine to generate _new_ values you couldn't before, as long as you can handle old ones.
> > That definitely blows up the size of the test files though.
> > If you did that you'd kind of want a means not of generating a random structure, but of specifically generating a random instance of each relevant variant.
> 
> It's imperfect for some of the reasons you've mentioned here, but it does catch _changes_. The changes may not be breaking (for example, in the case of adding a new variant,) but we may as well opt-into new values that the `Write` impl can produce.
> 
> I'd prefer not to save encoded values of the struct for tons of cases; The test files get very very large.

fwiw I'd like to be forced to regenerate if we add a new variant

consensus/codec_conformance.toml
clabby: This is mostly used as a unique key. It's not the actual path of the type.

justfile
clabby: The fixtures for now are just related to codec; If we add another purpose in the future let's revisit 👍

patrick-ogrady (approved):

.github/workflows/fast.yml
```diff
@@ -283,8 +283,8 @@ jobs:
       - name: Lint feature propagation
         run: zepter run check
 
-  codec-conformance:
-    name: Codec Conformance
+  conformance:
```
patrick-ogrady, line 286: I think we should make the just file consistent with this change.

codec/src/conformance.rs
clabby: My reasoning for placing it in codec is so that we can also run the conformance tests on the types in `codec`, since they underly all of our other more complex codec impls. If we split out just the helpers, we'd have a cyclical dependency between `codec` and the test crate.

This'd work if we also split out all the primitive implementations of codec and just left `commonware-codec` as the traits, but that may be a bit invasive.

patrick-ogrady: See: https://github.com/commonwarexyz/monorepo/issues/2258#issuecomment-3639513300

patrick-ogrady (approved):

Issue #1836, opened 2025-10-08
[cryptography/bls12381/dkg] Investigate Quasi-Linear Optimization


patrick-ogrady:
A few years ago, some folks suggested an interesting optimization to the `recover` step (one of our most time-intensive operations in threshold). We should investigate it and see if it is worth applying:
* https://alinush.github.io/threshold-bls
* https://github.com/alinush/libpolycrypto/

PR #2432, opened 2025-12-09
Create abstraction around Sharing

### LINKED ISSUE #2367: [cryptography] Create abstraction over "polynomial" of a threshold signature

cronokirby:
Closes #2367.

Instead of directly using a `primitives::poly::Public` you now use a `primitives::Sharing`, which will allow for supporting other interpolation schemes without breaking changes. Furthermore, `dkg2` is updated to take an extra mode argument, which is currently enforced to just be the default mode. There are no others now, but we can now introduce them without a breaking API change.

### DIFF
--- a/consensus/src/simplex/signing_scheme/bls12381_threshold.rs
+++ b/consensus/src/simplex/signing_scheme/bls12381_threshold.rs
@@ -36,7 +36,7 @@ use commonware_cryptography::{
     },
     Digest, PublicKey,
 };
-use commonware_utils::ordered::{Quorum, Set};
+use commonware_utils::ordered::Set;
 use rand::{CryptoRng, Rng};
 use std::{
     collections::{BTreeSet, HashMap},
@@ -84,9 +84,9 @@ impl<P: PublicKey, V: Variant> Scheme<P, V> {
     /// * `share` - local threshold share for signing
     pub fn signer(participants: Set<P>, polynomial: Sharing<V>, share: Share) -> Option<Self> {
         assert_eq!(
-            polynomial.required(),
-            participants.quorum(),
-            "polynomial threshold must equal quorum"
+            polynomial.total().get() as usize,
+            participants.len(),
+            "polynomial total must equal participant len"
         );
         polynomial.precompute_partial_publics();
         let partial_public = polynomial
@@ -113,9 +113,9 @@ impl<P: PublicKey, V: Variant> Scheme<P, V> {
     /// * `polynomial` - public polynomial for threshold verification
     pub fn verifier(participants: Set<P>, polynomial: Sharing<V>) -> Self {
         assert_eq!(
-            polynomial.required(),
-            participants.quorum(),
-            "polynomial threshold must equal quorum"
+            polynomial.total().get() as usize,
+            participants.len(),
+            "polynomial total must equal participant len"
         );
         polynomial.precompute_partial_publics();
 


cryptography/src/bls12381/primitives/sharing.rs
```diff
@@ -0,0 +1,240 @@
...
+            o => Err(commonware_codec::Error::InvalidEnum(o)),
+        }
+    }
+}
+
+/// Represents the public output of a polynomial secret sharing.
+///
+/// This does not contain any secret information.
+#[derive(Clone, Debug)]
+pub struct Sharing<V: Variant> {
```
patrick-ogrady, line 76: How would you feel about `Assignment` or `Dealing`? I find `Sharing` confusing because it is a verb.

cryptography/src/bls12381/primitives/sharing.rs
```diff
@@ -0,0 +1,240 @@
...
+#[cfg(feature = "std")]
+use std::sync::{Arc, OnceLock};
+
+/// Configures how participants are assigned shares of a secret.
+///
+/// More specifically, this configures how evaluation points of a polynomial
+/// are assigned to participant identities.
+#[derive(Copy, Clone, Default, PartialEq, Eq, Debug)]
+#[repr(u8)]
+pub enum SharingMode {
```
patrick-ogrady: nit: just call `Mode`

cryptography/src/bls12381/primitives/sharing.rs
```diff
@@ -0,0 +1,240 @@
...
+
+    fn read_cfg(
+        buf: &mut impl bytes::Buf,
+        cfg: &Self::Cfg,
+    ) -> Result<Self, commonware_codec::Error> {
+        let mode = ReadExt::read(buf)?;
+        // We bound total to the config, in order to prevent doing arbitrary
+        // computation if we precompute public keys.
+        let total = {
+            let out: u32 = ReadExt::read(buf)?;
```
patrick-ogrady, line 195: nit: can we read as `NonZeroU32`? :thinking:

cryptography/src/bls12381/primitives/sharing.rs
```diff
@@ -0,0 +1,240 @@
...
+
+/// Configures how participants are assigned shares of a secret.
+///
+/// More specifically, this configures how evaluation points of a polynomial
+/// are assigned to participant identities.
+#[derive(Copy, Clone, Default, PartialEq, Eq, Debug)]
+#[repr(u8)]
+pub enum SharingMode {
+    #[default]
+    NonZeroCounter = 0,
```
patrick-ogrady, line 25: Add a TODO for #1836

cryptography/src/bls12381/primitives/sharing.rs
```diff
@@ -0,0 +1,240 @@
...
+impl<V: Variant> PartialEq for Sharing<V> {
+    fn eq(&self, other: &Self) -> bool {
+        self.mode == other.mode && self.total == other.total && self.poly == other.poly
+    }
+}
+
+impl<V: Variant> Eq for Sharing<V> {}
+
+impl<V: Variant> Sharing<V> {
+    #[allow(dead_code)]
```
patrick-ogrady: Why does this need to be dead?

cryptography/src/bls12381/primitives/sharing.rs
```diff
@@ -0,0 +1,240 @@
...
+    pub(crate) fn new(mode: SharingMode, total: NonZeroU32, poly: Public<V>) -> Self {
+        Self {
+            mode,
+            total,
+            poly: Arc::new(poly),
+            #[cfg(feature = "std")]
+            evals: Arc::new(vec![OnceLock::new(); total.get() as usize]),
```
patrick-ogrady, line 99: :fancy:

cryptography/src/bls12381/primitives/sharing.rs
```diff
@@ -0,0 +1,240 @@
...
+
+    /// Call this to pre-compute the results of [`Self::partial_public`].
+    ///
+    /// This should be used if you expect to access many of the partial public
+    /// keys, e.g. if verifying several public signatures.
+    ///
+    /// The first time this method is called can be expensive, but subsequent
+    /// calls are idempotent, and cheap.
+    #[cfg(feature = "std")]
+    pub fn precompute_partial_publics(&self) {
```
patrick-ogrady, line 134: nit: I wonder if we could get away with `precompute` here?

patrick-ogrady: Or even, `prepare`?

patrick-ogrady: I suppose `prepare` sounds a bit more required

examples/bridge/src/bin/validator.rs
```diff
@@ -118,13 +114,12 @@ fn main() {
...
     // Configure threshold
-    let threshold = validators.quorum();
     let identity = matches
         .get_one::<String>("identity")
         .expect("Please provide identity");
     let identity = from_hex(identity).expect("Identity not well-formed");
-    let identity: Public<MinSig> =
-        Poly::decode_cfg(identity.as_ref(), &RangeCfg::exact(NZU32!(threshold)))
+    let identity: Sharing<MinSig> =
+        Sharing::decode_cfg(identity.as_ref(), &NZU32!(validators.len() as u32))
```
patrick-ogrady, line 122: I wonder if we can add to `Quorum` in `ordered` to return `NonZeroU32` for len or at least len as `u32`? Maybe that's overkill...

cryptography/src/bls12381/primitives/sharing.rs
cronokirby: I'm not particular attached to `Sharing`.

One point against `Dealing` is that one might infer that this contains shares.

cronokirby: Some ideas:

- Roster
- Docket
- Caucus
- Muster
- Conclave
- Assembly
- Scaffold

cryptography/src/bls12381/primitives/sharing.rs
cronokirby: `Mode` works great inside of just the crate, but I'm imagining someone making calls like:

```rust
dkg_stuff(Mode::Foo, ...
```

in their own code, where it seems like:

```rust
dkg_stuff(SharingMode::Foo, ...
```

would be clearer, if they also have a `Sharing` import.

cryptography/src/bls12381/primitives/sharing.rs
cronokirby: I think tying the name to the method that it makes faster is clearest, and this doesn't end up called in too many places.

cryptography/src/bls12381/primitives/sharing.rs
cronokirby: Unfortunately we don't have impls for that yet ; we should create an issue.

cryptography/src/bls12381/primitives/sharing.rs
patrick-ogrady: In other parts of the codebase, we do exactly this (`Mode`) and require users to rename if they want it (otherwise it studders with the import path `sharing::SharingMode`)

cryptography/src/bls12381/primitives/sharing.rs
```diff
@@ -0,0 +1,240 @@
...
+use std::sync::{Arc, OnceLock};
+
+/// Configures how participants are assigned shares of a secret.
+///
+/// More specifically, this configures how evaluation points of a polynomial
+/// are assigned to participant identities.
+#[derive(Copy, Clone, Default, PartialEq, Eq, Debug)]
+#[repr(u8)]
+pub enum SharingMode {
+    // TODO: Add a mode for sub O(N^2) interpolation
```
patrick-ogrady: nit: `TODO (#1836): Add a mode for sub O(N^2) interpolation`

cryptography/src/bls12381/primitives/sharing.rs
```diff
@@ -0,0 +1,240 @@
...
+/// Represents the public output of a polynomial secret sharing.
+///
+/// This does not contain any secret information.
+#[derive(Clone, Debug)]
+pub struct Sharing<V: Variant> {
+    mode: SharingMode,
+    total: NonZeroU32,
+    poly: Arc<Public<V>>,
+    #[cfg(feature = "std")]
+    evals: Arc<Vec<OnceLock<V::Public>>>,
```
patrick-ogrady, line 81: We involve `Arc` here just to avoid losing evals computed on clones, right?

cryptography/src/bls12381/primitives/sharing.rs
patrick-ogrady: Given we have the `partial_public` function below, I'm ok with this.

cryptography/src/bls12381/primitives/sharing.rs
```diff
@@ -0,0 +1,240 @@
...
+impl<V: Variant> PartialEq for Sharing<V> {
+    fn eq(&self, other: &Self) -> bool {
+        self.mode == other.mode && self.total == other.total && self.poly == other.poly
+    }
+}
+
+impl<V: Variant> Eq for Sharing<V> {}
+
+impl<V: Variant> Sharing<V> {
+    pub(crate) fn new(mode: SharingMode, total: NonZeroU32, poly: Public<V>) -> Self {
```
patrick-ogrady: Dumb question: why accept a `SharingMode` rather than just making `Sharing` generic over `Mode`?

It would complicate the `read_cfg` implementation (would have to know mode before reading) but figure that isn't a blocker (maybe I'm being short-sighted).

cryptography/src/bls12381/primitives/sharing.rs
patrick-ogrady: I'd prefer `Assignment` over your suggestions (I think they get a little too creative).

patrick-ogrady: I guess the rationale of not using `polynomial` is that it is more than that...it is a `polynomial` and rules of how to evaluate?

Maybe we call it an `Evaluation`?

patrick-ogrady (commented):
left some more comments

cryptography/src/bls12381/primitives/sharing.rs
cronokirby: The idea is that you'll be able to upgrade to a new `Mode` with no code changes, other than making it so that on the next DKG, you pass in the mode you want to use. Without this, you would need to like, have the DKG be parameterized over the mode of the previous iteration, and the mode of the current one, and then have a version of the codebase that you run that exists just for the transition period between the two, to run a dkg migrating from mode A to mode B. With a runtime parameter, you'll just use the old mode in the sharing to do everything, and if you get a new sharing with a faster mode, you'll do that.

In other words, the idea behind "mode" is to support upgrading without requiring code changes, making this generic goes against that.

cryptography/src/bls12381/primitives/sharing.rs
cronokirby: Yes, the idea is to allow `Sharing` to be cheaply cloneable.

cryptography/src/bls12381/primitives/sharing.rs
cronokirby: It isn't really an evaluation of anything either...

While it does contain a polynomial, that doesn't get at what its purpose is. The point of the struct is that it's what lets you verify partial signatures, and turn partial signatures into a group signature.

`Assignment` indicates that to some extent, but it seems to not be quite there...

cryptography/src/bls12381/primitives/poly.rs
```diff
@@ -349,7 +356,7 @@ impl<C: Element> Poly<C> {
     ///
     /// This function assumes that each evaluation has a unique index. If there are duplicate indices, the function may
     /// fail with an error when attempting to compute the inverse of zero.
-    pub fn recover<'a, I>(t: u32, evals: I) -> Result<C, Error>
+    pub fn recover<'a, I>(mode: Mode, n: NonZeroU32, t: u32, evals: I) -> Result<C, Error>
```
patrick-ogrady, line 387: I suppose its too strict but I guess you could require `Sharing` here (should provide `mode`, `n`, and `t`)?

patrick-ogrady: I understand why you avoided though given you don't actually need to call anything on the underlying polynomial :thinking:

cryptography/src/bls12381/primitives/sharing.rs
patrick-ogrady: That explains the use of `Sharing`. Maybe that isn't so bad then.

cryptography/src/bls12381/primitives/poly.rs
```diff
@@ -359,7 +366,7 @@ impl<C: Element> Poly<C> {
 
         // Compute weights
         let indices = evals.iter().map(|e| e.index).collect::<Vec<_>>();
-        let weights = compute_weights(indices)?;
+        let weights = compute_weights(mode, n, indices)?;
```
patrick-ogrady, line 397: I guess the underlying assumption here that whatever `Sharing` is used, we use the same "weighted recovery" mechanism to get back to a valid signature?

cryptography/src/bls12381/primitives/poly.rs
```diff
@@ -140,12 +143,16 @@ where
...
+    mode: Mode,
+    total: NonZeroU32,
+    indices: Vec<u32>,
+) -> Result<BTreeMap<u32, Weight>, Error> {
     // Compute weights for all provided evaluation indices
     let mut weights = BTreeMap::new();
     for i in &indices {
         // Convert i_eval.index to x-coordinate (x = index + 1)
-        let xi = Scalar::from_index(*i);
+        let xi = mode.scalar(total, *i).ok_or(Error::InvalidIndex)?;
```
patrick-ogrady, line 183: Call me crazy, but I'd prefer to call `sharing.scalar` here without involving `mode` or `total`

patrick-ogrady (commented):
Did a quick scan...biggest question is whether we should use `Sharing` even more?

cryptography/src/bls12381/primitives/poly.rs
cronokirby: The reason I held back in the `poly` module is that `recover` and related functions get moved to `math` (in the form of the `Interpolator` API, and the plan is in #2398 to then add a method on sharing to create the interpolator, basically realizing this.

patrick-ogrady: Ah, got it.

cryptography/src/bls12381/primitives/group.rs
```diff
@@ -319,6 +319,14 @@ impl Scalar {
     }
 
     /// Creates a new scalar from the provided index (a scalar offset by 1).
+    ///
+    /// This is important, because when doing polynomial secret sharing,
```
patrick-ogrady, line 323: ❤️

cryptography/src/bls12381/primitives/poly.rs
```diff
@@ -167,12 +170,16 @@ where
...
```
patrick-ogrady, line 184: Should `xj` also use `mode`?

patrick-ogrady: Can we write a regression test for this if so?

cryptography/src/bls12381/primitives/sharing.rs
patrick-ogrady: Bumping: `// TODO (#1836): Add a mode for sub O(N^2) interpolation`

cryptography/src/bls12381/primitives/sharing.rs
```diff
@@ -0,0 +1,240 @@
...
+/// are assigned to participant identities.
+#[derive(Copy, Clone, Default, PartialEq, Eq, Debug)]
+#[repr(u8)]
+pub enum Mode {
+    // TODO: Add a mode for sub O(N^2) interpolation
+    #[default]
+    NonZeroCounter = 0,
+}
+
+impl Mode {
```
patrick-ogrady, line 28: I think where I've landed on Mode is that its ok as an enum because there are no "types" associated with it (unlike Variant or a Scheme...it just drives behavior over the same types).

That being said, I do really like the specificity of binding to a generic (and suspect you would know it for the epoch you are in because you'd need to verify whatever is coming in over the wire) :thinking:

cryptography/src/bls12381/primitives/ops.rs
```diff
@@ -1768,43 +1738,35 @@ mod tests {
             .map(|s| partial_sign_message::<MinSig>(s, namespace, msg))
             .collect();
```
patrick-ogrady: nit: remove `\n`

cryptography/src/bls12381/primitives/ops.rs
```diff
@@ -1768,43 +1738,35 @@ mod tests {
...
 
+        sharing.precompute_partial_publics();
```
patrick-ogrady, line 1740: nit: add `\n`

cryptography/src/bls12381/primitives/mod.rs
```diff
@@ -66,3 +67,5 @@ pub enum Error {
     #[error("evaluation index is invalid")]
     InvalidIndex,
 }
+
+pub use sharing::{Mode, Sharing};
```
patrick-ogrady: nit: for the `Mode` change to look right (imo), we should just `pub mod sharing`?

cryptography/src/bls12381/dkg.rs
```diff
@@ -428,7 +430,7 @@ impl<V: Variant, P: PublicKey> Read for Output<V, P> {
         Ok(Self {
             summary: ReadExt::read(buf)?,
             players: Read::read_cfg(buf, &(RangeCfg::new(1..=max_players.get() as usize), ()))?,
-            public: Read::read_cfg(buf, &RangeCfg::from(NZU32!(1)..=max_players))?,
+            public: Read::read_cfg(buf, &max_players)?,
```
patrick-ogrady, line 433: Without generics, I think we need to bind `Mode` here?

patrick-ogrady: Or is the idea you always check mode before doing something after parse?

cryptography/src/bls12381/primitives/sharing.rs
patrick-ogrady: I suppose this does, however, serve as an interesting workaround for "trait specialization" if more changes end up being made internally than just using a different scalar.

cryptography/src/bls12381/primitives/sharing.rs
```diff
@@ -0,0 +1,240 @@
...
+impl<V: Variant> PartialEq for Sharing<V> {
+    fn eq(&self, other: &Self) -> bool {
+        self.mode == other.mode && self.total == other.total && self.poly == other.poly
+    }
+}
+
+impl<V: Variant> Eq for Sharing<V> {}
+
+impl<V: Variant> Sharing<V> {
+    pub(crate) fn new(mode: Mode, total: NonZeroU32, poly: Public<V>) -> Self {
```
patrick-ogrady, line 93: What if we just defined `total` on `Mode` (if the variant required it)?

cryptography/src/bls12381/primitives/poly.rs
patrick-ogrady, line 1: Should `evaluate` also accept `mode`? It also includes `Scalar::from_index`?

cryptography/src/bls12381/primitives/ops.rs
patrick-ogrady, line 1: `threshold_derive_missing_partials` hardcodes `from_index`?

cryptography/src/bls12381/primitives/poly.rs
patrick-ogrady, line 1: `tests::evaluate` hardcodes `Scalar::from_index`

cryptography/src/bls12381/primitives/sharing.rs
cronokirby: Another way to think about it is that the mode is the same kind of data as the shares or the polynomial. The polynomial in combination with the evaluation points determines the shares. Using the wrong mode is like using the wrong polynomial, but checking that you're using the right polynomial at compile time isn't what we want to do.

cryptography/src/bls12381/dkg.rs
cronokirby: You never check mode. The idea is that the sharing itself has enough information to create partial signatures, and recover them. The way new modes get introduced is that you'll run the DKG with the new mode passed at runtime, and then the sharing you get out of that will use that mode. But, the previous sharing will use the old mode, and really, the only point you care about the mode is when you're calling a method that creates a sharing, like deal, or dkg, etc. Otherwise, you just need to pass around the sharing.

cryptography/src/bls12381/primitives/poly.rs
cronokirby: `evaluate` should really accept `Scalar`, but that's coming in #2398.

cryptography/src/bls12381/primitives/ops.rs
cronokirby: I think that's fine because it also hard-codes a specific mode ; when the test is updated to use different modes it can be updated for correctness as well.

cryptography/src/bls12381/primitives/sharing.rs
cronokirby: I don't like that, because then you would then do something like:

```
dkg(Mode::NTT(players.len()), players, ...
```

when really the mode itself isn't tied to a specific length, but rather is able to function for any length, which gets determined by the participant set you end up actually sharing with.

Also, it's not that a mode may _contain_ a total: it's that a mode is allowed to _use_ a total, which is not the same.

`Mode::NTT(100)` vs `Mode::NTT(200)` wouldn't really make sense, they're both the same mode.

consensus/src/simplex/signing_scheme/bls12381_threshold.rs
```diff
@@ -84,34 +82,25 @@ impl<P: PublicKey, V: Variant> Scheme<P, V> {
...
+        if partial_public == share.public::<V>() {
+            Some(Self::Signer {
+                participants,
+                polynomial,
+                share,
+            })
+        } else {
+            None
+        }
     }
```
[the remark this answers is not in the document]
cronokirby: Hmmm, we should probably check that the totals match instead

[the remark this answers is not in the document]
cronokirby: Pushed a commit

cronokirby:
Closing in favor of #2398.

PR #2434, opened 2025-12-09
deployer: remove unused VPC and subnet CIDR maps

Bashmunta:
Removed the vpc_cidrs and subnet_cidrs HashMaps from the EC2 create path because they were only written to and never read. The effective VPC CIDR per region is already stored in RegionResources.vpc_cidr and used for peering routes and security group rules, so these extra maps were redundant.


patrick-ogrady:
> Hm. You are correct.
> 
> Now, I'm wondering if our lack of their use is a bug (or if this was a remnant of an earlier approach I abandoned) 🤔

I've concluded this is from before the use of `RegionResources` and this is safe to remove.

PR #2435, opened 2025-12-09
docs: fix incorrect code examples in AGENTS.md

aso20455:
Fixed two incorrect code examples in AGENTS.md:

1. Stateful Recovery Testing example used non-existent `context.recover()` method and wrong `start()` return signature. Replaced with correct `start_and_recover()` API that returns `(Output, Checkpoint)`.

2. Dynamic Network Conditions example used non-existent `oracle.update_link()`. Replaced with actual pattern: `remove_link()` followed by `add_link()`, as `add_link` returns `LinkExists` error if link already exists.


clabby:

Issue #2005, opened 2025-10-26
[runtime] Escape or Error on `/` in Partition/Blob Names


patrick-ogrady:
We don't handle `/` and may inadvertently create nested directories if `/` is in the partition or blob name.

PR #2437, opened 2025-12-09
[runtime] Validate partition names

clabby:
## Overview

Validates partition names to contain alphanumeric characters, dashes (`-`), and underscores (`_`) only.

closes https://github.com/commonwarexyz/monorepo/issues/2005

### DIFF
--- a/runtime/src/storage/mod.rs
+++ b/runtime/src/storage/mod.rs
@@ -423,7 +423,13 @@ pub(crate) mod tests {
         }
 
         // Invalid partition names should return PartitionNameInvalid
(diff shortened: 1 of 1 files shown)


storage/src/journal/contiguous/variable.rs
```diff
@@ -1608,7 +1608,7 @@ mod tests {
         let executor = deterministic::Runner::default();
         executor.start(|context| async move {
             let cfg = Config {
-                partition: "init_a  t_size_boundary".to_string(),
```
patrick-ogrady, line 1611: 🤦

PR #2438, opened 2025-12-09
[ci] Clear Disk Space

clabby:
## Overview

Adds a re-usable action that clears disk space on linux, macos, and windows.

### DIFF
--- a/.github/workflows/slow.yml
+++ b/.github/workflows/slow.yml
@@ -50,6 +50,8 @@ jobs:
     steps:
     - name: Checkout repository
       uses: actions/checkout@08eba0b27e820071cde6df949e0beb9ba4906955 # v4
+    - name: Clear disk space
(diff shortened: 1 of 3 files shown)


.github/workflows/slow.yml
```diff
@@ -50,6 +50,8 @@ jobs:
     steps:
     - name: Checkout repository
       uses: actions/checkout@08eba0b27e820071cde6df949e0beb9ba4906955 # v4
+    - name: Clear disk space
```
patrick-ogrady, line 53: Does it make sense to invoke this in setup?

clabby: We could, though it takes a while. Figure it's best to do on an as-needed basis.

patrick-ogrady (approved):

patrick-ogrady:
Noice

PR #2443, opened 2025-12-10
[utils] NZU* macros validate literals at compile time

andresilva:
Updates the `NZUsize!`, `NZU8!`, `NZU16!`, `NZU32!`, and `NZU64!` macros to validate at compile time when passed literal values. Previously, all validation happened at runtime, now literals are wrapped in a `const { }` block, so `NZU32!(0)` fails to compile rather than panicking at runtime.


patrick-ogrady:
Huge improvement

PR #2444, opened 2025-12-10
[consensus/simplex] make signing scheme signer construction explicit

andresilva:
Refactors the signing scheme API to make signer construction explicit by renaming `new()` to `signer()` and returning `Option<Self>` instead of `Self`.

Previously, calling `new()` with a private key that didn't match any participant would silently create a verifier instance instead of a signer. This could mask configuration errors where a node believes it's participating in consensus but is actually only observing. In scenarios where falling back to a verifier is acceptable users can now handle this explicitly.


cronokirby (approved):
LGTM

patrick-ogrady (approved):
nice

PR #2445, opened 2025-12-10
[runtime/utils/buffer/pool] handle page number overflow

### LINKED ISSUE #1878: overflow in runtime/src/utils/buffer/pool.rs:251
attempt to add with overflow in [pool.rs#L251 ](https://github.com/commonwarexyz/monorepo/blob/d0964ace3eea67c123b456901e04da867c0dc9e0/runtime/src/utils/buffer/pool.rs#L251)
```
#![no_main]

use arbitrary::Arbitrary;
use commonware_runtime::{buffer::PoolRef, deterministic, Runner};
use std::num::NonZeroUsize;

#[derive(Arbitrary, Debug)]
struct FuzzInput {
    seed: u64,
    page_size: u16,
    pool_capacity: u16,
    blob_id: u64,
    offset: u64,
    data: Vec<u8>,
}

fuzz_target!(|input: FuzzInput| {
    if input.page_size == 0 || input.pool_capacity == 0 {
        return;
    }
    
    let page_size = input.page_size as usize;
    let pool_capacity = input.pool_capacity as usize;
    
    if page_size > 1024 * 1024 || pool_capacity > 10000 {
        return;
    }
    
    let executor = deterministic::Runner::seeded(input.seed);
    let _ = executor.start(|_context| async move {
        let pool_ref = PoolRef::new(
            NonZeroUsize::new(page_size).unwrap_or(NonZeroUsize::new(1024).unwrap()),
            NonZeroUsize::new(pool_capacity).unwrap_or(NonZeroUsize::new(10).unwrap())
        );
        
        // The vulnerability is triggered when offset is not page-aligned
        // We need to avoid the panic in the fuzz target itself
        let (_, offset_in_page) = (input.offset / page_size as u64, (input.offset % page_size as u64) as usize);
        if offset_in_page != 0 {
            // Skip non-page-aligned offsets to prevent the fuzz target from panicking
            // In a real scenario, this would be the vulnerability we're testing for
            return;
        }
        
        let data_len = input.data.len().min(1024 * 1024);
        let data = &input.data[..data_len];
        
        let _ = pool_ref.cache(input.blob_id, data, input.offset).await;
    });
});
```

andresilva:
Fix #1878.

### DIFF
--- a/runtime/src/utils/buffer/pool.rs
+++ b/runtime/src/utils/buffer/pool.rs
@@ -495,6 +495,7 @@ mod tests {
             // Use the minimum page size to force the page index to reach u64::MAX and trigger the
             // overflow guard.
             let pool_ref = PoolRef::new(NZUsize!(1), NZUsize!(2));
+
             // Caching across the maximum page should stop before overflow and report the remainder.
             let remaining = pool_ref.cache(0, &[1, 2], u64::MAX).await;
             assert_eq!(remaining, 1);


runtime/src/utils/buffer/pool.rs
```diff
@@ -248,7 +253,10 @@ impl PoolRef {
             while buf.len() >= self.page_size {
                 buffer_pool.cache(self.page_size, blob_id, &buf[..self.page_size], page_num);
                 buf = &buf[self.page_size..];
-                page_num += 1;
+                page_num = match page_num.checked_add(1) {
```
andresilva, line 258: I guess an alternative would be to forbid 1-byte pages, but the branch predictor will very likely make this check zero-cost anyway.

patrick-ogrady (approved):

PR #2447, opened 2025-12-10
[storage/metadata] support variable-length keys

### LINKED ISSUE #1339: [storage/metadata] Migrate Keys to `sequence::Span`
Related: #1337

andresilva:
This PR relaxes the key type constraint on Metadata from `Array` to `Span`, allowing for variable-size keys.
Made some small API changes around key prefixes, since `Span` doesn't provide `AsRef<[u8]>` like `Array`:

- `keys(Option<&[u8]>)` becomes `keys()`, returning an iterator over all keys. Callers that need prefix filtering can use `.filter()` on the iterator with their key type's methods.
- `remove_prefix(&[u8])` becomes `retain(FnMut(&K, &V) -> bool)`, giving callers full control over the removal predicate.

The marshal cache in consensus now uses `u8` directly instead of `FixedBytes<1>` for its single metadata key.

Closes #1339

### DIFF
--- a/storage/src/metadata/storage.rs
+++ b/storage/src/metadata/storage.rs
@@ -312,10 +312,12 @@ impl<E: Clock + Storage + Metrics, K: Span, V: Codec> Metadata<E, K, V> {
 
     /// Retain only the keys that satisfy the predicate.
     pub fn retain(&mut self, mut f: impl FnMut(&K, &V) -> bool) {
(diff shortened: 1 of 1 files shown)


patrick-ogrady (approved):

patrick-ogrady:
I mistakingly applied a "breaking-format" label here, however, `u8` and `FixedBytes<1>` actually encode the same way.

PR #2450, opened 2025-12-10
[consensus/simplex] immediately nullify proposals that fail verification

### LINKED ISSUE #1768: [consensus/simplex] Emit `nullify` if `verify()` fails
If proposal verify explicitly fails, we should consider immediately voting `nullify` (and opt to never verify again).

Insipration: #1767

andresilva:
When a proposal fails verification, the voter now emits a nullify vote immediately rather than waiting for the timeout to expire.

Closes #1768.

### DIFF
--- a/consensus/src/simplex/actors/voter/mod.rs
+++ b/consensus/src/simplex/actors/voter/mod.rs
@@ -2181,9 +2181,9 @@ mod tests {
 
             // Register network channels for the validator
             let (vote_sender, _vote_receiver) =
-                oracle.control(me.clone()).register(0).await.unwrap();
+                oracle.control(me.clone()).register(0, TEST_QUOTA).await.unwrap();
             let (certificate_sender, _certificate_receiver) =
-                oracle.control(me.clone()).register(1).await.unwrap();
+                oracle.control(me.clone()).register(1, TEST_QUOTA).await.unwrap();
 
             // Start the actor
             voter.start(


consensus/src/simplex/actors/voter/mod.rs
```diff
@@ -2019,4 +2020,211 @@ mod tests {
         no_resolver_boomerang(bls12381_multisig::<MinSig, _>);
         no_resolver_boomerang(ed25519);
     }
+
+    /// Tests that when proposal verification fails, the voter emits a nullify vote
+    /// immediately rather than waiting for the timeout.
+    fn verification_failure_emits_nullify_immediately<S, F>(mut fixture: F)
```
andresilva, line 2100: I checked that this test fails (as expected) if we comment out the `handle_timeout` on failed verification in the voter actor.

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -406,11 +406,11 @@ impl<E: Clock + Rng + CryptoRng + Metrics, S: Scheme, D: Digest> State<E, S, D>
...
     }
 
-    /// Marks proposal verification as complete when the peer payload validates.
-    pub fn verified(&mut self, view: View) -> bool {
+    /// Marks proposal verification as complete.
+    pub fn verified(&mut self, view: View, valid: bool) -> bool {
         self.views
             .get_mut(&view)
-            .map(|round| round.verified())
+            .map(|round| round.verified(valid))
```
patrick-ogrady: We should never request verification twice (and just not calling `verified` should be enough "I think")?

consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -817,24 +817,25 @@ impl<
...
+                    // Mark verification complete
                     view = context.view();
-                    if !self.state.verified(view) {
+                    if !self.state.verified(view, valid) {
                         continue;
                     }
+                    if !valid && view == self.state.current_view() {
+                        // Verification failed for current view proposal, treat as immediate timeout
+                        debug!(round = ?context.round, "proposal failed verification");
+                        self.handle_timeout(&mut batcher, &mut vote_sender, &mut certificate_sender)
```
patrick-ogrady, line 834: I'd instead just do:

```
if valid {
  self.state.verified()
} else {
  self.handle_timeout()
}
```

consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -815,25 +815,34 @@ impl<
                     // Clear verify waiter
                     let context = pending_verify_context.take().unwrap();
                     pending_verify = None;
+                    view = context.view();
+
+                    if view != self.state.current_view() {
```
andresilva: This shouldn't happen but IMO it's better to be safe, since `handle_timeout` will nullify the current view.

consensus/src/simplex/actors/voter/state.rs
andresilva: I updated the PR to do this, although I think the previous version had some advantages by strictly enforcing some invariants, e.g. that we never voted to notarize or finalize a proposal that failed verification, and that once a verification status becomes "failed verification" it could never change.

consensus/src/simplex/mocks/application.rs
```diff
@@ -176,6 +178,10 @@ impl<E: Clock + RngCore + Spawner, H: Hasher, P: PublicKey> Application<E, H, P>
         )
     }
 
+    pub const fn set_fail_verification(&mut self, fail: bool) {
```
patrick-ogrady, line 181: I thought about proposing we pass this into the config, but its nice to be able to set this dynamically!

Issue #1833, opened 2025-10-08
[runtime] Refactor Context API


patrick-ogrady:
While working on ~#1817~ #1839 (via #1692), I became quite frustrated with the error-prone design of context.

It tries to both care deeply about tracking exact usage (by avoiding cloning with spawn_ref and supporting with_label) but at the same time still permits interacting with contexts that have been cloned at different levels of the call stack (often to workaround Rust's borrow checker in an async context). This is further complicated by `spawn` providing a child context (that is then interacted with) rather than passing through the context used to kickoff the task.

We should consider redesigning `context` to enforce a more strict hierarchy (which IMHO will unlock more interesting usage).

clabby:
Lots can be learned from moro + prior art on structured concurrency/parallelism imo:
- std's scoped tasks API https://doc.rust-lang.org/stable/std/thread/fn.scope.html
- moro: https://docs.rs/moro/latest/moro/macro.async_scope.html (Niko's attempt at expressing it well)
- without boats' words on the scoped task trilemma: https://without.boats/blog/the-scoped-task-trilemma/
- tree-structured concurrency: https://blog.yoshuawuyts.com/tree-structured-concurrency/
- Evan Schwartz' thoughts: https://emschwartz.me/async-rust-can-be-a-pleasure-to-work-with-without-send-sync-static/

While we probably won't land on the "Borrowing + Concurrency" end of the triangle that srrrse describes in the scoped task trilemma, the APIs that've been explored around this area seem really close to what we want.

patrick-ogrady:
My favorite is definitely Evan Schwartz's thoughts.

Another good one: https://blog.yoshuawuyts.com/tasks-are-the-wrong-abstraction/

patrick-ogrady:
#1839 cleaned things up (dropped `_ref` and `_blocking`) but did so by allowing the context to be further modified (just as `with_label` changes the tracked name). I'm generally not a fan of modifying contexts (and like that is usually just a "getter") but other approaches are pretty verbose (like supplying the supervision/execution mode during each call to spawn).

patrick-ogrady:
I suppose we could have a graceful shutdown signal on context itself for partial graceful shutdown (but figure the global shutdown signal we have is sufficient for actual needs).

PR #2452, opened 2025-12-10
fix: register metrics with post-start context

crStiv:
Resolves TODO(#1833) by moving metrics registration from Engine::new() to Engine::run().

Metrics are now registered using the post-start context, which ensures they are properly initialized after the engine is spawned. This change separates metrics creation from registration, allowing the metrics to use the correct runtime context.


clabby:

PR #2459, opened 2025-12-11
[utils] `channels::ring`

clabby:
## Overview

Adds a channel to `utils` that drops the oldest value upon reaching capacity rather than applying backpressure.

### DIFF
--- a/utils/src/channels/ring.rs
+++ b/utils/src/channels/ring.rs
@@ -9,10 +9,10 @@
 //! ```
 //! use futures::executor::block_on;
 //! use futures::{SinkExt, StreamExt};
-//! use commonware_utils::channels::ring;
+//! use commonware_utils::{NZUsize, channels::ring};
 //!
 //! block_on(async {
-//!     let (mut sender, mut receiver) = ring::channel::<u32>(2);
+//!     let (mut sender, mut receiver) = ring::channel::<u32>(NZUsize!(2));
 //!
 //!     // Fill the channel
 //!     sender.send(1).await.unwrap();
@@ -38,11 +38,9 @@ use std::{
 use thiserror::Error;
 
 /// Error returned when sending to a channel whose receiver has been dropped.
-///
-/// Contains the item that failed to send, allowing recovery of the value.
 #[derive(Debug, Error)]
-#[error("receiver dropped")]
-pub struct SendError<T>(pub T);
+#[error("channel closed")]
+pub struct ChannelClosed;
 
 struct Shared<T: Send + Sync> {
     buffer: VecDeque<T>,
@@ -66,7 +64,7 @@ pub struct Sender<T: Send + Sync> {
 impl<T: Send + Sync> Sender<T> {
     /// Returns whether the receiver has been dropped.
     ///
-    /// If this returns `true`, subsequent sends will fail with [`SendError`].
+    /// If this returns `true`, subsequent sends will fail with [`ChannelClosed`].
     pub fn is_closed(&self) -> bool {
         let shared = self.shared.lock().unwrap();
         shared.receiver_dropped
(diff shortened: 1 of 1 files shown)


clabby:
bugbot run

utils/src/channels/ring.rs
```diff
@@ -0,0 +1,412 @@
...
+
+/// Creates a new ring channel with the specified capacity.
+///
+/// Returns a ([`Sender`], [`Receiver`]) pair. The sender can be cloned to create
+/// multiple producers.
+///
+/// # Panics
+///
+/// Panics if `capacity` is zero.
+pub fn channel<T>(capacity: usize) -> (Sender<T>, Receiver<T>) {
```
patrick-ogrady: does this need to be `T: Send + Sync`?

utils/src/channels/ring.rs
```diff
@@ -0,0 +1,412 @@
...
+}
+
+/// Creates a new ring channel with the specified capacity.
+///
+/// Returns a ([`Sender`], [`Receiver`]) pair. The sender can be cloned to create
+/// multiple producers.
+///
+/// # Panics
+///
+/// Panics if `capacity` is zero.
```
patrick-ogrady: Use `NZUsize`? (I suppose you did this for compatibility with how other channels are typically initialized?)

utils/src/channels/ring.rs
```diff
@@ -0,0 +1,412 @@
...
+pub struct Sender<T> {
+    shared: Arc<Mutex<Shared<T>>>,
+}
+
+impl<T> Sender<T> {
+    /// Returns whether the receiver has been dropped.
+    ///
+    /// If this returns `true`, subsequent sends will fail with [`SendError`].
+    pub fn is_closed(&self) -> bool {
+        let shared = self.shared.lock().unwrap();
```
patrick-ogrady, line 69: nit: could use a `RWLock`?

utils/src/channels/ring.rs
```diff
@@ -0,0 +1,412 @@
...
+
+        Self {
+            shared: self.shared.clone(),
+        }
+    }
+}
+
+impl<T> Drop for Sender<T> {
+    fn drop(&mut self) {
+        let Ok(mut shared) = self.shared.lock() else {
```
patrick-ogrady, line 88: Is the idea that the lock is more likely to be poisoned in `drop` (and `unwrap` is too aggressive)?

utils/src/channels/ring.rs
```diff
@@ -0,0 +1,412 @@
...
+            w.wake();
+        }
+    }
+}
+
+impl<T> Sink<T> for Sender<T> {
+    type Error = SendError<T>;
+
+    fn poll_ready(self: Pin<&mut Self>, _cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        // This channel never applies backpressure - it drops the oldest item instead.
```
patrick-ogrady: nit: emdash strikes again ❗

utils/src/channels/ring.rs
```diff
@@ -0,0 +1,412 @@
...
+    task::{Context, Poll, Waker},
+};
+use thiserror::Error;
+
+/// Error returned when sending to a channel whose receiver has been dropped.
+///
+/// Contains the item that failed to send, allowing recovery of the value.
+#[derive(Debug, Error)]
+#[error("receiver dropped")]
+pub struct SendError<T>(pub T);
```
patrick-ogrady: :fancy:

utils/src/channels/ring.rs
clabby: i hate to say but a human wrote this (non-em) dash 😂

utils/src/channels/ring.rs
clabby: Yeah, this is mainly to conform with other channels. Though unlike `futures::mpsc`, a `0` capacity isn't valid, so maybe that is the right way to go...

I'll change it up.

utils/src/channels/ring.rs
clabby: It's not worth using an `RwLock` because there should never be frequent reads on the sender side, it's an mpsc-style channel. The extra overhead is only worth it when we expect many concurrent readers.

utils/src/channels/ring.rs
clabby: Correct. If, for example, the mutex was poisoned due to a panic in another thread that held the lock, we want to explicitly avoid a double panic. https://doc.rust-lang.org/std/ops/trait.Drop.html#panics

utils/src/channels/ring.rs
patrick-ogrady: That's what a bot would say

utils/src/channels/ring.rs
patrick-ogrady: Nice

patrick-ogrady (approved):
neato

PR #2460, opened 2025-12-11
fix(cryptography): prevent BloomFilter arbitrary from generating zero hashers

reallesee:
Fixed a bug in BloomFilter's Arbitrary implementation where hashers could be zero.
When hashers=0, the indices() method returns an empty iterator, which causes contains() to always return true regardless of whether the item was actually inserted. This completely breaks the bloom filter semantics (100% false positive rate).

The public constructor already requires NonZeroU8, so arbitrary should respect the same invariant. Added .max(1) to ensure at least one hasher is always generated.

### DIFF
--- a/cryptography/conformance.toml
+++ b/cryptography/conformance.toml
@@ -4,7 +4,7 @@ hash = "c0501d4a691d1fccec7c5906e8608228569d24164150edd215838593e3b77512"
 
 ["commonware_cryptography::bloomfilter::tests::conformance::CodecConformance<BloomFilter>"]
 n_cases = 65536
-hash = "a75d6312366816126114abdc9d7fbd246891cab210eeae6a02781625aa1ad6a4"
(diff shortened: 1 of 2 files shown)


danlaine:
I think it might make more sense to make the field `hashers: NonZeroU8`?

reallesee:
The public API already requires NonZeroU8, so the invariant is enforced at construction. This just fixes arbitrary to respect the same constraint. Changing the internal field would be a larger refactor for no practical benefit.

patrick-ogrady (dismissed):

patrick-ogrady:
With this change, you'll need to call `regenerate-conformance` (in the `justfile`).

reallesee:
> With this change, you'll need to call `regenerate-conformance` (in the `justfile`).

Did it, thanks

patrick-ogrady (approved):

reallesee:
@patrick-ogrady CI needs breaking-format label, can you add it?

andresilva:
#2729 includes this as well (didn't see this PR before).

clabby:

Issue #2424, opened 2025-12-08
[consensus/marshal] Explore Support for "Lossy" Applications


patrick-ogrady:
When an application restarts, it may find it has not persisted the block data it acknowledged (to optimize performance, applications may only periodically "sync" processed state to disk): https://github.com/commonwarexyz/monorepo/blob/24a318a7213c919096da3686bbf643201841f848/consensus/src/marshal/mod.rs#L110-L118

While possible to fetch necessary blocks during propose and verify via the [AncestorStream](https://github.com/commonwarexyz/monorepo/blob/24a318a7213c919096da3686bbf643201841f848/consensus/src/lib.rs#L147), the application has no way to know what the finalized floor is during this walk back (which may change how state is synced to disk).

Like how applications can fetch the last finalized block (which will eventually be the last processed block), they could be equipped to fetch the last processed block (doing the minimum work necessary to align with `marshal` -> `marshal` will stream finalized blocks above last processed already, so we don't need to fetch twice).

We should either explicitly discourage this "lossy" application behavior (and double-down on acknowledgement being a real acknowledgement) or offer a path for applications to fetch what their processed state should be...

PR #2461, opened 2025-12-11
[consensus/marshal] Support for "Lossy" Applications

stevencartavia:
Return marshal's processed height from `Actor::init()` to support "lossy" application recovery scenarios. 
towards #2424

### DIFF
--- a/consensus/src/marshal/mod.rs
+++ b/consensus/src/marshal/mod.rs
@@ -431,7 +431,7 @@ mod tests {
                 .update(0, participants.clone().try_into().unwrap())
                 .await;
             for (i, validator) in participants.iter().enumerate() {
-                let (application, actor, _) = setup_validator(
+                let (application, actor, _processed_height) = setup_validator(
                     context.with_label(&format!("validator-{i}")),
                     &mut oracle,
                     validator.clone(),
@@ -577,7 +577,7 @@ mod tests {
                 .update(0, participants.clone().try_into().unwrap())
                 .await;
             for (i, validator) in participants.iter().enumerate().skip(1) {
-                let (application, actor, _) = setup_validator(
+                let (application, actor, _processed_height) = setup_validator(
                     context.with_label(&format!("validator-{i}")),
                     &mut oracle,
                     validator.clone(),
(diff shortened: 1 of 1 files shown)


consensus/src/marshal/mod.rs
```diff
@@ -1715,6 +1715,89 @@ mod tests {
...
+            // Verify application received both blocks in order
+            assert_eq!(application.blocks().len(), 2);
+            assert_eq!(application.tip(), Some((2, commitment2)));
+
+            // Test setting floor to same value (should be idempotent)
+            actor.set_floor(2).await;
+            assert_eq!(actor.get_processed_height().await, Some(2));
+
+            // Setting floor to lower value should be ignored (safety feature)
+            actor.set_floor(1).await;
```
patrick-ogrady: Forgot we added that "ignore" check...great test.

stevencartavia:
This approach works well during operation for monitoring/debugging maybe. However, I switched to an init-only approach and it's a bit cleaner. Lmk.

patrick-ogrady:
> This approach works well during operation for monitoring/debugging maybe. However, I switched to an init-only approach and it's a bit cleaner. Lmk.

I think this should already be covered by metrics sufficiently. I'm moreso thinking of "operational" mechanisms that would change their behavior based on this value (that aren't already a subscriber to events) 🤔

stevencartavia:
> > This approach works well during operation for monitoring/debugging maybe. However, I switched to an init-only approach and it's a bit cleaner. Lmk.
> 
> I think this should already be covered by metrics sufficiently. I'm moreso thinking of "operational" mechanisms that would change their behavior based on this value (that aren't already a subscriber to events) 🤔

So this gap detection should be handled through existing metrics queries instead of a new API? Or is there a different approach you prefer for this restart scenario?

patrick-ogrady:
> > > This approach works well during operation for monitoring/debugging maybe. However, I switched to an init-only approach and it's a bit cleaner. Lmk.
> > 
> > 
> > I think this should already be covered by metrics sufficiently. I'm moreso thinking of "operational" mechanisms that would change their behavior based on this value (that aren't already a subscriber to events) 🤔
> 
> So this gap detection should be handled through existing metrics queries instead of a new API? Or is there a different approach you prefer for this restart scenario?

I was just commenting that any monitoring/debugging can occur via metrics. Not advocating for any other changes here.

patrick-ogrady (approved):

PR #2462, opened 2025-12-11
[consensus] clarify validation helpers in ordered_broadcast and aggregation

Forostovec:
Changed the documentation for the validate_node, validate_ack, and validate_chunk helpers so that their comments match the actual signatures and behavior. The functions now explicitly document that they either return Ok(()) (or Option<Chunk> for validate_node) on success or an error on failure, instead of claiming to return full tuples or threshold signatures.


clabby:

Issue #2376, opened 2025-12-03
[runtime] Allow controlling bytes of the RNG in tests directly


cronokirby:
It could be useful to control the exact bytes output by the RNG, especially for fuzzing, where you might want to try and explore "adjacent" randomness for tests, in order to exercise behavior more effectively.

This is a common approach in various prop testing / fuzzing libraries, where an RNG pulling from a fixed pool of bytes is one option.

A simple way to integrate this would be some kind of `with_rng_prefix` method on the deterministic Runtime, which makes it so that the RNG will first emit the bytes provided to that method, and then proceed with the seeded RNG, once it runs out.

Another approach would be to return all 0s after you run out, but this seems worse for most of our tests.

patrick-ogrady:
May be better to just take an `Rng` as a `seed` (rather than an integer). This allows for providing custom sources of randomness without having to bring this all into the runtime.

TL;DR replace `with_seed` with `with_rng` (I suppose may be best to just support both and have `with_seed` call `with_rng`).

evonide:
Thanks, I'll take a stab at it.

PR #2470, opened 2025-12-11
Provide `with_rng` to runtime

evonide:
Draft for: https://github.com/commonwarexyz/monorepo/issues/2376.

Adds `Runner::with_rng()` to allow injecting a custom RNG implementation into the deterministic runtime. This enables fuzzing scenarios where the fuzzer controls randomness by providing its own `RngCore` implementation instead of the default seeded `StdRng`.

The change is backwards compatible since existing seed-based usage continues to work unchanged. When `with_rng` is not called, the runtime falls back to `StdRng::seed_from_u64(cfg.seed)`. The custom RNG is ignored when recovering from a `Checkpoint` since the checkpoint already contains the RNG state.

However, we currently add this on `Runner` rather than `Config` because `Config` derives `Clone` and maintains `const fn` setters while `Box<dyn RngCore + Send>` is not cloneable.

@patrick-ogrady Please advise if this is acceptable given we have `with_seed` on the `Config` and this does feel somewhat clunky.

### DIFF
--- a/runtime/src/deterministic.rs
+++ b/runtime/src/deterministic.rs
@@ -391,7 +391,11 @@ impl From<Config> for Runner {
 
 impl From<(Config, Box<dyn RngCore + Send>)> for Runner {
     fn from((cfg, rng): (Config, Box<dyn RngCore + Send>)) -> Self {
-        Self::new(cfg).with_rng(rng)
+        cfg.assert();
+        Self {
+            state: State::Config(cfg),
+            rng: Some(rng),
+        }
     }
 }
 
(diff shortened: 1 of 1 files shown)


evonide:
We still have the `with_seed` logic and seeds in the runtime which we should change. However, we can't call `with_rng` which is in the `Runner` from `Config` it seems.

runtime/src/deterministic.rs
```diff
@@ -403,9 +415,18 @@ impl Runner {
...
         Self {
             state: State::Config(cfg),
+            rng: None,
         }
     }
 
+    /// Override the source of randomness for the runtime.
+    ///
+    /// Ignored if this [Runner] was created from a [Checkpoint].
+    pub fn with_rng(mut self, rng: impl RngCore + Send + 'static) -> Self {
```
patrick-ogrady, line 425: I think we should instead define this on `Config` (replacing `seed` with `rng` and implementing `seed` as  `StdRng::seed_from_u64(cfg.seed)`?

evonide: Agreed that this would be better, however, this is currently part of `Runner` rather than `Config` because `Config` derives `Clone` while `Box<dyn RngCore + Send>` is not cloneable.

patrick-ogrady: IMHO we should do whatever it takes to introduce RNG here.

evonide: Sorry I meant to agree and ask a question here. What are options to overcome the mentioned limitation to add it to `Config`? We could drop `Clone` from `Config` but then would have to update all Config consumers and many tests. Maybe there's another way than to use `Box<dyn RngCore + Send>` here?

patrick-ogrady:
Replaced with #2704

PR #2477, opened 2025-12-12
zoda: Improve Zoda error specificity by splitting ambiguous variants

syrupipy:
Currently when a Zoda reshard verification fails, we get a generic `invalid reshard` error. When debugging, it's impossible to tell if the issue is dimension mismatch, inclusion proof failure, or checksum error - you have to add debug prints to figure it out.

This splits the generic errors into specific variants:
- `InvalidShard` → `ShardCommitmentMismatch` / `ShardChecksumDimensionsMismatch` 
- `InvalidReShard` → `ReshardDimensionsMismatch` / `ReshardInclusionProofFailed` / `ReshardChecksumMismatch`

Now errors include the actual context:
```
reshard dimensions mismatch: rows=10, cols=20, expected rows=15, cols=20
reshard checksum mismatch at shuffled index 42
```

Makes debugging encoding issues much faster.

### DIFF
--- a/coding/src/zoda.rs
+++ b/coding/src/zoda.rs
@@ -477,12 +477,17 @@ impl<H: Hasher> CheckingData<H> {
         transcript.commit(root.encode());
         let expected_commitment = transcript.summarize();
         if *commitment != expected_commitment {
-            return Err(Error::InvalidShard);
+            return Err(Error::ShardCommitmentMismatch);
         }
         let transcript = Transcript::resume(expected_commitment);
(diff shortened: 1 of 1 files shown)


cronokirby:
This is intentionally limited because all of these errors come from someone else doing something incorrectly, and those are errors you have to deal with just by nature of being in an adversarial environment. The other errors have details, because they represent mistakes the programmer using the API is making (like not checking they have enough shards, providing the wrong shards, etc.)

PR #2481, opened 2025-12-12
[storage/qmdb/current] Improve current DB proving API

roberto-bayardo:
This PR gets rid of the "ProofInfo" objects, and instead creates specialized Proof types for current db inclusion & exclusion proofs containing all necessary data other than that which is not already part of the proof statement.

### DIFF
--- a/storage/fuzz/fuzz_targets/current_ordered_operations.rs
+++ b/storage/fuzz/fuzz_targets/current_ordered_operations.rs
@@ -225,7 +225,7 @@ fn fuzz(data: FuzzInput) {
                             assert!(
                                 Current::<deterministic::Context, Key, Value, Sha256, TwoCap, 32>::verify_range_proof(
                                     hasher.inner(),
-                                    &proof,
+                                    proof,
                                     start_loc,
                                     &ops,
                                     &chunks,
@@ -258,7 +258,7 @@ fn fuzz(data: FuzzInput) {
 
                         let _ = Current::<deterministic::Context, Key, Value, Sha256, TwoCap, 32>::verify_range_proof(
                             hasher.inner(),
-                            &proof,
+                            proof,
                             start_loc,
                             &res.1,
                             chunks,
--- a/storage/fuzz/fuzz_targets/current_unordered_operations.rs
+++ b/storage/fuzz/fuzz_targets/current_unordered_operations.rs
@@ -203,7 +203,7 @@ fn fuzz(data: FuzzInput) {
                             assert!(
                                 Current::<deterministic::Context, Key, Value, Sha256, TwoCap, 32>::verify_range_proof(
                                     hasher.inner(),
-                                    &proof,
+                                    proof,
                                     start_loc,
                                     &ops,
                                     &chunks,
@@ -236,7 +236,7 @@ fn fuzz(data: FuzzInput) {
 
                         let _ = Current::<deterministic::Context, Key, Value, Sha256, TwoCap, 32>::verify_range_proof(
                             hasher.inner(),
-                            &proof,
+                            proof,
                             start_loc,
                             &res.1,
                             chunks,
--- a/storage/src/qmdb/current/mod.rs
+++ b/storage/src/qmdb/current/mod.rs
@@ -96,6 +96,180 @@ pub struct OperationProof<D: Digest, const N: usize> {
     pub proof: Proof<D>,
 }
 
+impl<D: Digest, const N: usize> OperationProof<D, N> {
+    /// Return an inclusion proof that incorporates activity status for the operation designated by
+    /// `loc`.
+    pub async fn new<H: CHasher, S: Storage<H::Digest>>(
+        hasher: &mut H,
+        status: &CleanBitMap<H::Digest, N>,
+        grafting_height: u32,
+        grafted_mmr: &S,
+        loc: Location,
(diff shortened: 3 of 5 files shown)


storage/fuzz/fuzz_targets/current_ordered_operations.rs
```diff
@@ -280,26 +278,20 @@ fn fuzz(data: FuzzInput) {
...
+                    match db.key_value_proof(hasher.inner(), k.clone()).await {
+                        Ok(proof) => {
+                            let value = value.expect("key should exist if proof was generated");
                             let verification_result = Current::<deterministic::Context, Key, Value, Sha256, TwoCap, 32>::verify_key_value_proof(
                                 hasher.inner(),
-                                &proof,
-                                info.clone(),
+                                k,
+                                value.clone(),
+                                proof,
```
danlaine, line 288: We clone k so I think we can get the value inside the `Ok` case

```suggestion
let current_root = db.root();

                    match db.key_value_proof(hasher.inner(), k.clone()).await {
                        Ok(proof) => {
                            let value = db.get(&k).await.unwrap().unwrap();
                            let verification_result = Current::<deterministic::Context, Key, Value, Sha256, TwoCap, 32>::verify_key_value_proof(
                                hasher.inner(),
                                k,
                                value,
                                proof,
```

storage/fuzz/fuzz_targets/current_unordered_operations.rs
```diff
@@ -258,29 +258,20 @@ fn fuzz(data: FuzzInput) {
 
                     let current_root = db.root();
 
-                    match db.key_value_proof(hasher.inner(), k).await {
-                        Ok((proof, info)) => {
+                    // Get the current value first (before proof generation consumes k)
+                    let value = db.get(&k).await.expect("get should not fail");
```
danlaine: Same as above about being able to get `value` inside the `Ok` case

storage/src/qmdb/current/ordered.rs
```diff
@@ -376,21 +344,15 @@ impl<
...
-                value: key_data.value,
-                next_key: key_data.next_key,
-                loc,
-                chunk,
-            },
-        ))
+        Ok(KeyValueProof {
+            proof: kvp,
+            next_key: data.next_key,
+        })
```
danlaine: nit
```suggestion
        Ok(KeyValueProof {
            proof: OperationProof { loc, chunk, proof },
            next_key: data.next_key,
        })
```

storage/fuzz/fuzz_targets/current_ordered_operations.rs
roberto-bayardo: done

storage/fuzz/fuzz_targets/current_unordered_operations.rs
roberto-bayardo: done

storage/src/qmdb/current/ordered.rs
roberto-bayardo: done

storage/src/qmdb/current/mod.rs
```diff
@@ -82,6 +83,193 @@ impl<T: Translator> Config<T> {
...
+    pub async fn new<H: CHasher, S: Storage<H::Digest>>(
+        hasher: &mut H,
+        status: &CleanBitMap<H::Digest, N>,
+        grafting_height: u32,
+        grafted_mmr: &S,
+        loc: Location,
+    ) -> Result<OperationProof<H::Digest, N>, Error> {
+        let grafted_mmr = GraftingStorage::<'_, H, _, _>::new(status, grafted_mmr, grafting_height);
+
+        // loc is valid so it won't overflow from + 1
```
danlaine, line 111: We haven't checked that `loc` is valid here

storage/src/qmdb/current/mod.rs
```diff
@@ -82,6 +83,193 @@ impl<T: Translator> Config<T> {
...
+            proof.digests.push(hasher.finalize());
+        }
+
+        Ok(OperationProof { loc, chunk, proof })
+    }
+
+    /// Verify that the proof proves that `operation` is active in the database with the given
+    /// `root`.
+    pub fn verify<H: CHasher<Digest = D>, O: Codec>(
+        mut self,
```
danlaine, line 128: It'd be nice if we could make this `&self` by removing the "if there is a partial chunk then pop off the last digest" logic. Perhaps we could add an additional field `Option<Digest>` to `OperationProof` that contains the partial chunk's digest, if there is one?

Alternatively we could change `self.proof.digests.pop()` to `self.proof.digests.last()` and change

```rs
let mmr_root = match self
            .proof
            .reconstruct_root(&mut verifier, &[element], self.loc)
```

to something like

```rs
let mmr_root = match self
            .proof
            .take(self.proof.len() - 1)
            .reconstruct_root(&mut verifier, &[element], self.loc)
```
to avoid actually modifying `self`

roberto-bayardo: I like the idea of adding an Option<Digest>.  Let me do that.

roberto-bayardo:
These changes have been moved into #2507

storage/src/qmdb/current/mod.rs
roberto-bayardo: this is done in the replacement PR.

Issue #2487, opened 2025-12-12
[broadcast] Panic on Duration


patrick-ogrady:
```
thread '<unnamed>' (3660) panicked at /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/governor-0.10.2/src/nanos.rs:37:18:
Duration is longer than 584 years: TryFromIntError(())
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
==3660== ERROR: libFuzzer: deadly signal
```

Likely a function of #2483

patrick-ogrady:
Claude's summary:

```
  1. Fuzz test calls context.sleep(Duration::from_millis(u64::MAX))
  2. Deterministic runtime advances its internal clock by ~584 million years
  3. Rate limiter's check_key() is called (line 575) when trying to send a message
  4. Governor calls clock.now() which returns the runtime's SystemTime (now far in the future)
  5. Governor converts the duration to nanoseconds - this is where it panics

  The conversion happens inside governor because it stores timestamps as nanoseconds (u64). The maximum u64 in nanoseconds is ~584 years (not million years):

  u64::MAX nanoseconds ≈ 18.4 × 10^18 ns ≈ 584 years

  But u64::MAX milliseconds is ~584 million years - vastly exceeding governor's representable range.
```

PR #2488, opened 2025-12-12
[broadcast/fuzz] Fix Unbounded Sleep

patrick-ogrady:
Closes: #2487

### DIFF
--- a/broadcast/fuzz/fuzz_targets/broadcast_engine_operations.rs
+++ b/broadcast/fuzz/fuzz_targets/broadcast_engine_operations.rs
@@ -26,7 +26,7 @@ const TEST_QUOTA: Quota = Quota::per_second(NonZeroU32::MAX);
 ///
 /// Capped to avoid overflow in governor rate limiter which uses nanoseconds internally
 /// and can only represent durations up to ~584 years.
-const MAX_SLEEP_DURATION: u64 = 1000;
+const MAX_SLEEP_DURATION_MS: u64 = 1000;
 
 #[derive(Clone, Debug, Arbitrary)]
 pub enum RecipientPattern {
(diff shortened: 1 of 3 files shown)


broadcast/fuzz/fuzz_targets/broadcast_engine_operations.rs
```diff
@@ -22,6 +22,12 @@ use std::{collections::BTreeMap, num::NonZeroU32, time::Duration};
 /// Default rate limit set high enough to not interfere with normal operation
 const TEST_QUOTA: Quota = Quota::per_second(NonZeroU32::MAX);
 
+/// Maximum sleep duration in milliseconds for fuzz tests.
+///
+/// Capped to avoid overflow in governor rate limiter which uses nanoseconds internally
+/// and can only represent durations up to ~584 years.
+const MAX_SLEEP_DURATION: u64 = 1000;
```
cronokirby: Nit:
```suggestion
const MAX_SLEEP_DURATION_MS: u64 = 1000;
```

broadcast/fuzz/fuzz_targets/broadcast_engine_operations.rs
```diff
@@ -98,6 +104,34 @@ enum BroadcastAction {
...
+                digest: u.arbitrary()?,
+            }),
+            2 => Ok(BroadcastAction::Get {
+                peer_index: u.arbitrary()?,
+                sender: u.arbitrary()?,
+                commitment: u.arbitrary()?,
+                digest: u.arbitrary()?,
+            }),
+            _ => Ok(BroadcastAction::Sleep {
+                duration_ms: u.int_in_range(0..=MAX_SLEEP_DURATION)?,
```
cronokirby: Maybe we should create a wrapper struct called `SleepDuration`, implement Arbitrary for that using this, and then avoid having to create all this code?

cronokirby (commented):
I think it might be clearer to suffix the message types.

We could also avoid having to generate the enum manually in Arbitrary if we added a small wrapper type.

broadcast/fuzz/fuzz_targets/broadcast_engine_operations.rs
patrick-ogrady: I think we should create a follow-up issue for this (prefer to just unblock CI for now)

PR #2490, opened 2025-12-12
Fix clippy lints for rust 1.92.0

cronokirby:
When upgrading to the latest stable rust version, `just lint` fails.

This fixes the new complaints.

### DIFF
--- a/codec/src/types/hash_set.rs
+++ b/codec/src/types/hash_set.rs
@@ -306,7 +306,7 @@ mod tests {
         2u8.write(&mut expected2); // Item 2
         5u8.write(&mut expected2); // Item 5
         assert_eq!(set2.encode(), expected2.freeze());
-        assert_eq!(set2.encode_size(), 1 + 3 * u8::SIZE);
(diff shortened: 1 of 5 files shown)


codec/src/types/hash_set.rs
```diff
@@ -306,7 +306,7 @@ mod tests {
         2u8.write(&mut expected2); // Item 2
         5u8.write(&mut expected2); // Item 5
         assert_eq!(set2.encode(), expected2.freeze());
-        assert_eq!(set2.encode_size(), 1 + 3 * u8::SIZE);
+        assert_eq!(set2.encode_size(), 1 + 3);
```
clabby, line 309: Not thrilled that this fires a warning now :/

Maybe we `#[allow(clippy::identity_op)]` here or just disable that lint globally?

clabby: discussed offline; keeping given it's in a test.

clabby (approved):

PR #2492, opened 2025-12-13
[runtime] Add multi-headed deterministic runtime support

### LINKED ISSUE #2417: [runtime/deterministic] Add Support for Multi-Headed Runtime
I think it would be much more natural to write tests for the deterministic runtime if we allowed for creating "sub-instances" that were seen by applications as an independent instance (but still driven by a shared task queue for determinism).

This would allow for giving each instance its own IP (right now the same IP is shared by all: #2344), the same storage partition is shared by all (need to explicitly manage partitioning when handing to applications), and metrics are lumped together (need to ensure each instance has namespaced tasks).

If we do this right, it should allow for unblocking #2061 by allowing us to define link latency between IPs rather than peer identities. This would also unblock instance-specific storage performance #156.

TL;DR have an option to instantiate a "deterministic runtime manager" that can return "deterministic runtimes".

commonware-llm:
Add a Manager and Instance API for creating multiple isolated runtime
instances that share a single deterministic executor. Each instance has:

- Its own IP address for network operations
- Its own storage namespace (partitions are prefixed)
- Its own metrics namespace (metrics are prefixed)

This enables simulating multiple distributed system nodes in tests where
each node needs network isolation while sharing the same simulated time
and task scheduler.

Closes #2417


runtime/src/network/deterministic.rs
```diff
@@ -64,30 +67,53 @@ type Dialable = mpsc::UnboundedSender<(
...
         }
     }
 }
 
+impl Network {
+    /// Create a new network with a specific IP address.
+    ///
+    /// This shares the ephemeral port counter and listeners with other networks
+    /// created from the same base, but uses a different IP for its own connections.
+    pub fn with_ip(&self, ip: Ipv4Addr) -> Self {
```
patrick-ogrady, line 96: Require an IP (no defaults)

runtime/src/deterministic.rs
```diff
@@ -1346,6 +1346,205 @@ impl crate::Storage for Context {
...
+        /// The returned context has:
+        /// - Network operations using this instance's IP
+        /// - A metrics prefix based on the instance namespace
+        /// - Storage operations should use the namespace prefix manually or use
+        ///   the provided helper methods on Instance.
+        pub fn context(&self) -> Context {
+            // Return a context with the instance namespace as its label
+            // Create a proper child tree for supervision
+            let (child, _) = Tree::child(&self.context.tree);
+            let name = format!("i{}", self.namespace);
```
patrick-ogrady, line 1520: Calling `encode()` on this spawned context won't lead to isolated metrics?

patrick-ogrady: Maybe we wan't joined (just a guarantee not to conflict)?

It would be cool to have both...

p2p/src/simulated/ingress.rs
```diff
@@ -156,30 +141,22 @@ impl<P: PublicKey> Oracle<P> {
     /// setting will be used.
     ///
     /// Link can be called before a peer is registered or bandwidth is specified.
-    pub async fn add_link(&mut self, sender: P, receiver: P, config: Link) -> Result<(), Error> {
+    pub async fn add_link(&mut self, sender: P, receiver: P, link: Link) -> Result<(), Error> {
```
patrick-ogrady: At this point, it is probably better to just remove all link management from p2p::simulated and use runtime for it?

clabby:

PR #2489, opened 2025-12-12
[conformance] Complete codec conformance coverage

clabby:
## Overview

Adds more codec conformance tests for `coding`, `storage`, `p2p`, `utils`, `math`, and `resolver`.

Coverage for codec conformance within our library crates is complete with this PR:

```sh
rg "write for" -i -g '!fuzz' -g '!mocks.rs' -g '!examples' | wc -l
# -> 154
```

closes https://github.com/commonwarexyz/monorepo/issues/2258

PR #2493, opened 2025-12-13
[conformance] Human-readable test names

### LINKED ISSUE #2491: [conformance] Human-readable test names
## Overview

The `conformance` crate currently uses a recursive declarative macro to generate several test cases for types implementing the `Conformance` trait. Because some of these types have characters that aren't valid in an identifier, we can't easily use `paste!` to create test names from the type name (i.e. `test_vec<u8>`.) The workaround is just appending an `x` to the test name on each iteration.

We should figure out a method to make the test names legible so it's easier to spot what is failing without looking at the error message.

clabby:
## Overview

Adds a new `commonware-conformance-macros` crate that contains procedural macros for `commonware-conformance`. 

Using a procedural macro for test generation cleans up our existing macro definition with a more readable alternative, and it opens the door for us to add `derive` macros etc. going forwards. For now, the procedural macro for generating tests operates exactly the same as the recursive declarative macro, but with the advantage of being able to emit human-readable test names.

closes #2491


cryptography/conformance.toml
```diff
@@ -54,14 +54,6 @@ hash = "d112d774a4c9b012b53e44dc2a3487a0e2f87cd7b14fb8433ec7cea4997f440a"
...
 hash = "169550a1597994aad7ba634091b89633424be84f6753fe95542d2f3fd4108b9f"
 
-["commonware_cryptography::bls12381::primitives::poly::tests::arbitrary_tests::CodecConformance<Eval<Scalar>>"]
-n_cases = 65536
-hash = "169550a1597994aad7ba634091b89633424be84f6753fe95542d2f3fd4108b9f"
-
-["commonware_cryptography::bls12381::primitives::poly::tests::arbitrary_tests::CodecConformance<Poly<Scalar>>"]
-n_cases = 65536
-hash = "6b9cc50d4e055367a4f830285f084d030d2221414d5448d48e9db58251f8df7c"
-
```
clabby, line 64: These have moved over to `math`, which does not yet have codec conformance coverage. That is added in #2489.

clabby:
> and it opens the door for us to add derive macros etc. going forwards.

Example of a macro for deriving the tests: https://github.com/commonwarexyz/monorepo/compare/cl/conformance-macros...cl/derive-conformance-test?expand=1

Not sure if I'm a huge fan of this over just declaring the tests at the bottom of the file, but an argument could be made that it's nice to be able to see at a glance that the type is covered.

PR #2499, opened 2025-12-13
fix mmr doc

roynalnaruto:
Nitpick fix in MMR module docs


roynalnaruto:
Fixed in https://github.com/commonwarexyz/monorepo/pull/2691 hence closing

PR #2505, opened 2025-12-14
fix: move unused quorum import to test_plan module

dizer-ti:
The `quorum` function was imported at the top level of dkg.rs but only used inside the `test_plan` submodule. This caused an "unused import" warning during compilation. Moved the import to where it's actually needed.


patrick-ogrady:
Please merge in main and the unrelated error should go away.

Seems this happens because `cargo clippy --all-targets` enables all features (so the arbitrary is triggered).

PR #2507, opened 2025-12-14
[storage/qmdb] Improve current-db proving API, enforce "commit always exists" invariant in all qmdbs.

roberto-bayardo:
- gets rid of the "ProofInfo" objects, and instead creates specialized Proof types for current db inclusion & exclusion proofs containing all necessary data other than that which is not already part of the proof statement.

- makes all qmdb variants enforce the invariant that there is always one active commit by ensuring there is always a commit op at location 0.

### DIFF
--- a/storage/src/qmdb/current/mod.rs
+++ b/storage/src/qmdb/current/mod.rs
@@ -100,10 +100,10 @@ impl<D: Digest> RangeProof<D> {
         hasher: &mut H,
         status: &CleanBitMap<H::Digest, N>,
         grafting_height: u32,
-        grafted_mmr: &S,
+        mmr: &S,
         range: Range<Location>,
     ) -> Result<RangeProof<H::Digest>, Error> {
-        let grafted_mmr = GraftingStorage::<'_, H, _, _>::new(status, grafted_mmr, grafting_height);
+        let grafted_mmr = GraftingStorage::<'_, H, _, _>::new(status, mmr, grafting_height);
         let proof = verification::range_proof(&grafted_mmr, range).await?;
 
         let (last_chunk, next_bit) = status.last_chunk();
@@ -238,13 +238,12 @@ impl<D: Digest, const N: usize> OperationProof<D, N> {
         hasher: &mut H,
         status: &CleanBitMap<D, N>,
         grafting_height: u32,
-        grafted_mmr: &S,
+        mmr: &S,
         loc: Location,
     ) -> Result<Self, Error> {
         // Since `loc` is assumed to be in-bounds, `loc + 1` won't overflow.
         let range_proof =
-            RangeProof::<D>::new(hasher, status, grafting_height, grafted_mmr, loc..loc + 1)
-                .await?;
+            RangeProof::<D>::new(hasher, status, grafting_height, mmr, loc..loc + 1).await?;
         let chunk = *status.get_chunk_containing(*loc);
 
         Ok(Self {
@@ -341,10 +340,8 @@ async fn range_proof<E: RStorage + Clock + Metrics, H: CHasher, C: Contiguous, c
     let end_loc = core::cmp::min(max_loc, leaves);
 
     // Generate the proof from the grafted MMR.
-    let grafted_mmr = GraftingStorage::<'_, H, _, _>::new(status, mmr, height);
     let proof =
-        RangeProof::<H::Digest>::new(hasher, status, height, &grafted_mmr, start_loc..end_loc)
-            .await?;
+        RangeProof::<H::Digest>::new(hasher, status, height, mmr, start_loc..end_loc).await?;
 
     // Collect the operations necessary to verify the proof.
     let mut ops = Vec::with_capacity((*end_loc - *start_loc) as usize);


storage/src/qmdb/any/unordered/fixed.rs
```diff
@@ -45,7 +46,14 @@ impl<E: Storage + Clock + Metrics, K: Array, V: FixedValue, H: Hasher, T: Transl
         callback: impl FnMut(bool, Option<Location>),
     ) -> Result<Self, Error> {
         let translator = cfg.translator.clone();
-        let log = init_fixed_authenticated_log(context.clone(), cfg).await?;
+        let mut log = init_fixed_authenticated_log(context.clone(), cfg).await?;
+        if log.size() == 0 {
+            warn!("Authenticated log is empty, initializing new db");
```
danlaine, line 51: nit: should this be `info!`? `warn!` seems to suggest something potentially bad happened, but this is expected.

storage/src/qmdb/keyless/mod.rs
```diff
@@ -105,18 +106,15 @@ impl<E: Storage + Clock + Metrics, V: VariableValue, H: Hasher, S: State<DigestO
     }
 
     /// Returns the location of the last commit, if any.
```
danlaine:
```suggestion
    /// Returns the location of the last commit.
```

storage/src/qmdb/keyless/mod.rs
```diff
@@ -1085,13 +1086,18 @@ mod test {
             db.commit(None).await.unwrap();
 
             // Test getting valid locations - should succeed
-            assert_eq!(db.get(Location::new_unchecked(0)).await.unwrap().unwrap(), v1);
-            assert_eq!(db.get(Location::new_unchecked(1)).await.unwrap().unwrap(), v2);
+            assert_eq!(db.get(Location::new_unchecked(1)).await.unwrap().unwrap(), v1);
+            assert_eq!(db.get(Location::new_unchecked(2)).await.unwrap().unwrap(), v2);
+
+            // Test getting out of bounds location
```
danlaine, line 1091:
```suggestion
```

storage/src/qmdb/keyless/mod.rs
```diff
@@ -215,17 +218,20 @@ impl<E: Storage + Clock + Metrics, V: VariableValue, H: Hasher> Keyless<E, V, H,
     }
 
     /// Commit any pending operations to the database, ensuring their durability upon return from
-    /// this function. Caller can associate an arbitrary `metadata` value with the commit.
+    /// this function. Caller can associate an arbitrary `metadata` value with the commit. Returns
+    /// the `(start_loc, end_loc]` location range of committed operations. The end of the returned
+    /// range includes the commit operation itself, and hence will always be equal to `op_count`.
+    /// the range of locations that were committed.
```
danlaine:
```suggestion
```

storage/src/qmdb/any/unordered/mod.rs
```diff
@@ -373,26 +359,26 @@ where
...
             active_keys,
         })
     }
 
     /// Returns an [IndexedLog] initialized directly from the given components. The log is
     /// replayed from `inactivity_floor_loc` to build the snapshot, and that value is used as the
-    /// inactivity floor. The last commit location is set to None and it is the responsibility of the
-    /// caller to ensure it is set correctly.
+    /// inactivity floor.
     async fn from_components(
```
danlaine, line 377: It seems like we should be able to figure out the last commit location during `build_snapshot_from_log`, right? Perhaps we should have been doing that to begin with.

storage/src/qmdb/any/unordered/sync.rs
```diff
@@ -97,7 +97,8 @@ where
         .await?;
         // Build the snapshot from the log.
         let snapshot = Index::new(context.with_label("snapshot"), db_config.translator.clone());
-        let db = Self::from_components(range.start, log, snapshot).await?;
+        let db =
+            Self::from_components(range.start, log, snapshot, Location::new_unchecked(0)).await?;
```
danlaine: We're setting the last commit location to 0 here, which may not be correct. This isn't a new bug, exactly -- in the existing code, we leave the last commit location as None. Figuring the last commit location [inside `from_components`](https://github.com/commonwarexyz/monorepo/pull/2507/changes#r2620577195) should fix this.

storage/src/qmdb/any/unordered/fixed.rs
roberto-bayardo: My thinking was in the typical case you're using an existing DB, and initializing a new DB would be unusual, often meaning you provided a bad config?  It's a mistake I make often anyway!

storage/src/qmdb/any/unordered/mod.rs
roberto-bayardo: I wasn't sure what was going on here with last commit location TBH. I thought maybe you might sync a range that doesn't contain a commit?  Or should that never happen?

storage/src/qmdb/any/unordered/sync.rs
roberto-bayardo: I was concerned about this too.. see my comment re: might we provide a range to from_components that contains no commit?

storage/src/qmdb/keyless/mod.rs
roberto-bayardo: done

storage/src/qmdb/keyless/mod.rs
roberto-bayardo: fixed

storage/src/qmdb/keyless/mod.rs
roberto-bayardo: done

storage/src/qmdb/any/unordered/fixed.rs
danlaine: Fair enough, I'm OK keeping this

storage/src/qmdb/any/unordered/mod.rs
danlaine: I think the range populated by state sync should always be an `(inactivity floor, commit)` pair because that provides the active set of key-value pairs

roberto-bayardo: ok, changed it to use the last entry as the last commit, but this doesn't seem to be the case in a couple of the tests. debugging now.

roberto-bayardo: tests fixed as discussed.

storage/src/qmdb/current/mod.rs
```diff
@@ -150,7 +342,9 @@ async fn range_proof<E: RStorage + Clock + Metrics, H: CHasher, C: Contiguous, c
 
     // Generate the proof from the grafted MMR.
     let grafted_mmr = GraftingStorage::<'_, H, _, _>::new(status, mmr, height);
-    let mut proof = verification::range_proof(&grafted_mmr, start_loc..end_loc).await?;
+    let proof =
+        RangeProof::<H::Digest>::new(hasher, status, height, &grafted_mmr, start_loc..end_loc)
+            .await?;
```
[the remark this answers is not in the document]
roberto-bayardo: nice catch, removed the redundant mapping.

PR #2508, opened 2025-12-14
[resolver] Absorb `p2p::requester`

clabby:
## Overview

Absorbs `p2p`'s `requester` util into `commonware-resolver`. After this, `resolver::p2p` no longer maintains its own auxiliary rate-limiting, relying on the underlying outbound rate limiting in `p2p`.

### DIFF
--- a/resolver/src/p2p/fetcher.rs
+++ b/resolver/src/p2p/fetcher.rs
@@ -275,14 +275,14 @@ where
         let mut earliest_rate_limit: Option<SystemTime> = None;
         let mut found_eligible_peers = false;
         for (key, retry) in pending_keys {
-            let peers = self.get_eligible_peers(&key, retry);
-
             // Skip keys with no eligible peers
+            let peers = self.get_eligible_peers(&key, retry);
             if peers.is_empty() {
                 self.requests_created.inc(Status::Dropped);
                 continue;
             }
 
+            // Mark that an eligible peer was found
             self.requests_created.inc(Status::Success);
             found_eligible_peers = true;
 


resolver/src/p2p/engine.rs
```diff
@@ -113,13 +113,19 @@ impl<
     pub fn new(context: E, cfg: Config<P, D, B, Key, Con, Pro>) -> (Self, Mailbox<Key, P>) {
         let (sender, receiver) = mpsc::channel(cfg.mailbox_size);
 
+        use super::fetcher::Config as FetcherConfig;
```
patrick-ogrady: import at top of file

resolver/src/p2p/fetcher.rs
```diff
@@ -94,133 +148,205 @@ impl<E: Clock + GClock + Rng + Metrics, P: PublicKey, Key: Span, NetS: Sender<Pu
...
-                    // should skip it (may eventually become fetchable if the peer set changes).
                 }
             }
-        }
 
-        // Send request if we found a key to fetch
-        if let Some((key, peer, id)) = selected {
-            self.pending.remove(&key);
-            return self.send_request(sender, key, peer, id).await;
+            // All peers exhausted for this key, try next key
```
patrick-ogrady: The purpose of the previous loop was to stop early if all peers were rate-limited (determined by an untargeted key not being sent to anyone). It would be great to retain this (so we don't need to loop over all keys for each request if no one is able).

resolver/src/p2p/fetcher.rs
```diff
@@ -94,133 +148,205 @@ impl<E: Clock + GClock + Rng + Metrics, P: PublicKey, Key: Span, NetS: Sender<Pu
...
-                self.requester.fail(req);
-                self.add_retry(key);
-            }
-            // If the message was sent to someone, add the request to the map
-            Ok(()) => {
-                self.active.insert(id, key);
-            }
-        }
+        // No keys could be fetched, set waiter to retry later
+        self.waiter = Some(self.context.current().saturating_add(self.retry_timeout));
```
patrick-ogrady: Ah, I forgot about this nuance. We want to set the next time to the "moment" the first eligible peer is able to take a request. This approach could dramatically reduce the number of requests we send.

patrick-ogrady (commented):
some nits based on required changes

resolver/src/p2p/fetcher.rs
patrick-ogrady: I think the only real way we could do this is to have a different return value for rate-limited peers (that specifies when they can be sent to again 🤔 )

resolver/src/p2p/fetcher.rs
clabby: Added this

p2p/src/simulated/network.rs
```diff
@@ -958,22 +930,21 @@ impl<P: PublicKey, E: Clock, F: SplitForwarder<P>> std::fmt::Debug for SplitSend
...
-        let recipients = (self.forwarder)(self.replica, &recipients, &message);
-        let Some(recipients) = recipients else {
+        recipients: Recipients<Self::PublicKey>,
+    ) -> Result<Self::Checked<'_>, SystemTime> {
+        // Route recipients via forwarder
+        let routed_recipients = (self.forwarder)(self.replica, &recipients, &Bytes::new());
+        let Some(routed_recipients) = routed_recipients else {
             // If the forwarder returns None, drop the message
-            return Ok(Vec::new());
+            return Err(SystemTime::now());
```
clabby: This is a bit weird, but should produce the same effect. We could add a new checked sender here to return `Ok(Vec::new())` instead though if need be..

p2p/src/utils/mux.rs
```diff
@@ -232,19 +232,18 @@ pub struct SubSender<S: Sender> {
...
         &mut self,
-        recipients: Recipients<S::PublicKey>,
-        payload: Bytes,
-        priority: bool,
-    ) -> Result<Vec<S::PublicKey>, S::Error> {
+        recipients: Recipients<Self::PublicKey>,
+    ) -> Result<Self::Checked<'_>, SystemTime> {
         self.inner
-            .send(self.subchannel, recipients, payload, priority)
+            .check(recipients)
```
patrick-ogrady, line 244: I think we were not properly RL'ing on subchannel before (if we ran two instances of simplex on subchannels, we'd double the usage).

p2p/src/simulated/network.rs
clabby: nvm; the twins test actually uses the decoded vote to direct the message.

p2p/src/simulated/network.rs
```diff
@@ -958,22 +930,61 @@ impl<P: PublicKey, E: Clock, F: SplitForwarder<P>> std::fmt::Debug for SplitSend
...
+        // Do a preliminary rate limit check with the original recipients.
+        // The actual routing may differ based on message content.
+        let _ = self.inner.check(recipients.clone()).await?;
+        Ok(SplitCheckedSender {
+            inner: &mut self.inner,
+            replica: self.replica,
+            forwarder: self.forwarder.clone(),
+            recipients,
+        })
+    }
```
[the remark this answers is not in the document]
clabby: This is true, but there's not a wonderful way around it with the current pattern where we filter after the fact. The twins test basically has an unlimted quota, so shouldn't be a problem in practice.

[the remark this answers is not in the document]
clabby: (We could remove the first check, at the cost of never receiving feedback for the latest time that all recipients are available if all are limited.)

examples/log/src/main.rs
```diff
@@ -222,7 +222,6 @@ fn main() {
             activity_timeout: ViewDelta::new(10),
             skip_timeout: ViewDelta::new(5),
             fetch_concurrent: 32,
```
patrick-ogrady, line 224: is there a way to expose a `hint()` from the limited sender? That would allow us to provide a dynamic concurrency recommendation (`online * RL`)?

patrick-ogrady: This would be a great followup issue 👁️

p2p/src/utils/mux.rs
```diff
@@ -304,12 +303,65 @@ impl<S: Sender> GlobalSender<S> {
...
+
+    async fn check(
+        &mut self,
+        recipients: Recipients<Self::PublicKey>,
+    ) -> Result<Self::Checked<'_>, SystemTime> {
+        self.inner
+            .check(recipients)
+            .await
+            .map(|checked| CheckedGlobalSender {
+                subchannel: 0,
```
patrick-ogrady: I almost think we should use Option<> here and assert if with_subchannel isn't called ?

p2p/src/lib.rs
```diff
@@ -37,14 +37,14 @@ pub enum Recipients<P: PublicKey> {
     One(P),
 }
 
-/// Interface for sending messages to a set of recipients.
-pub trait Sender: Debug + Clone + Send + Sync + 'static {
-    /// Error that can occur when sending a message.
-    type Error: Debug + StdError + Send + Sync;
-
+/// Interface for sending messages to a set of recipients without rate-limiting restrictions.
+pub trait UnlimitedSender: Clone + Send + Sync + 'static {
```
patrick-ogrady, line 41: I think `Unbounded` is more common in rust but understand it plays with "rate limiting" and `limitedsender`

resolver/src/p2p/fetcher.rs
```diff
@@ -75,146 +126,247 @@ pub struct Fetcher<E: Clock + Rng + Metrics, P: PublicKey, Key: Span, NetS: Send
...
+
+            // Track if this is an untargeted key (tries all eligible peers)
+            let is_untargeted = !self.targets.contains_key(&key);
+            // Track if all peers were rate-limited (sender.check errors)
+            let mut all_rate_limited = true;
+
+            // Try each peer in order until one succeeds
+            for peer in peers {
+                // Generate request ID
+                let id = self.next_id();
```
patrick-ogrady: We should wait to generate until we know we're sending.

resolver/src/p2p/fetcher.rs
```diff
@@ -75,146 +126,247 @@ pub struct Fetcher<E: Clock + Rng + Metrics, P: PublicKey, Key: Span, NetS: Send
...
+            // If this untargeted key had all peers rate-limited, no other key will
+            // succeed either (rate limits are per-peer), so stop early
+            if is_untargeted && all_rate_limited {
+                break;
             }
         }
+
+        // No keys could be fetched, set waiter to retry as soon as possible
+        self.waiter = earliest_next_try
+            .or_else(|| Some(self.context.current().saturating_add(self.retry_timeout)))
```
patrick-ogrady: use the time returned from check!

resolver/src/p2p/fetcher.rs
```diff
@@ -75,146 +126,247 @@ pub struct Fetcher<E: Clock + Rng + Metrics, P: PublicKey, Key: Span, NetS: Send
...
-                    // If a request with targets fails to find a peer, other keys may still be fetchable
-                }
-                Err(Error::NoEligibleParticipants) => {
-                    // If a request with no valid targets exists (blocked or targets no longer allowed), we
-                    // should skip it (may eventually become fetchable if the peer set changes).
-                }
-            }
-        }
+        // Collect pending keys to iterate (we need to mutate self during iteration)
+        let pending_keys: Vec<(Key, bool)> = self
```
patrick-ogrady, line 268: We shouldn't have a TON of items here but just a good reminder not to 👀

p2p/src/simulated/network.rs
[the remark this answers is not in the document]
patrick-ogrady: hmmm...this is a fun one. need to think about what's preferred here.

resolver/src/p2p/fetcher.rs
clabby: This is the time returned from `check` (the minimum among all that were rate limited,) with a fallback to the retry timeout if no peers were rate limited (implying just regular failures.) The existing code has a similar fallback: https://github.com/commonwarexyz/monorepo/blob/66c2f637029774652e976b10e81449d0c17a863f/resolver/src/p2p/fetcher.rs#L166-L171

clabby: Rather than setting a short backoff, went back to the (effectively) infinite backoff. Using `retry_timeout` works if no peers were rate-limited and all requests fail, but we shouldn't need to re-attempt.

examples/log/src/main.rs
clabby: https://github.com/commonwarexyz/monorepo/issues/2549

resolver/src/p2p/fetcher.rs
patrick-ogrady: Ah, duh. I'm an idiot. my b.

patrick-ogrady: On the max duration point, this should really only be hit if there are no peers to request.

p2p/src/simulated/network.rs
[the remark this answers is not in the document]
patrick-ogrady: Opted to address a bit differently (based on what this actually does in practice): https://github.com/commonwarexyz/monorepo/pull/2550

Won't say it is perfect, but I think it is "better" (the back-pressure in "check" is a critical function IMHO)

resolver/src/p2p/fetcher.rs
```diff
@@ -1370,21 +1670,22 @@ mod tests {
     fn test_skips_keys_with_rate_limited_targets() {
```
patrick-ogrady, line 1709: Restored this test 👍

resolver/src/p2p/fetcher.rs
```diff
@@ -1031,22 +1331,22 @@ mod tests {
     fn test_waiter_cleared_on_target_modification() {
```
patrick-ogrady, line 1324: restored this test 👍

resolver/src/p2p/fetcher.rs
```diff
@@ -984,21 +1284,21 @@ mod tests {
     fn test_waiter_after_empty() {
```
patrick-ogrady, line 1279: restored this test 👍

resolver/src/p2p/fetcher.rs
```diff
@@ -75,146 +129,251 @@ pub struct Fetcher<E: Clock + Rng + Metrics, P: PublicKey, Key: Span, NetS: Send
...
-                self.active.insert(id, key);
+
+            // If this untargeted key had all peers rate-limited, no other key will
+            // succeed either (rate limits are per-peer), so stop early
+            if is_untargeted && all_rate_limited {
+                break;
             }
         }
+
+        // No keys could be fetched, set waiter to retry as soon as possible.
```
patrick-ogrady: If no peers are rate limited but no sends are successful, this logic may lead to it getting stuck?

patrick-ogrady: Resolved.

patrick-ogrady (approved):

PR #2518, opened 2025-12-15
[runtime] Vectorized sink

### LINKED ISSUE #784: [runtime/stream] Support the use of `OwnedWriteHalf::write_vectored`
Right now, `send_frame` makes two syscalls to write a frame, one to write the size of the frame, and one to write the actual data. I suspect this issue may be exacerbated further when using TCP_NODELAY.

Instead, we could concatenate the data together and `send` once in order to make a single syscall, but this would require copying the memory, which is especially undesirable for large messages.

Instead, it's probably best to simply support tokio's `write_vectored()` function which takes in a slice of slices and will send them all without having to do copying (besides the references)

clabby:
## Overview

Adjusts the interface of `runtime::Sink` to accept a `Buf`, allowing implementations to use vectorized write primitives (e.g. `tokio`'s `write_vectorized`, or `iouring`'s `writev` op.)

closes #784

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -445,7 +445,10 @@ pub trait Sink: Sync + Send + 'static {
     ///
     /// Implementations restrict the maximum number of buffers that can be
     /// written at once to `16`.
-    fn send(&mut self, bufs: impl Buf + Send) -> impl Future<Output = Result<(), Error>> + Send;
+    fn send(
+        &mut self,
+        bufs: impl Buf + Send + 'static,
+    ) -> impl Future<Output = Result<(), Error>> + Send;
 }
 
 /// Interface that any runtime must implement to receive
@@ -2604,7 +2607,7 @@ mod tests {
                     let request = format!(
                         "GET /metrics HTTP/1.1\r\nHost: {address}\r\nConnection: close\r\n\r\n"
                     );
-                    sink.send(request.as_bytes()).await.unwrap();
+                    sink.send(bytes::Bytes::from(request)).await.unwrap();
 
                     // Read and verify the HTTP status line
                     let status_line = read_line(&mut stream).await.unwrap();
--- a/stream/src/utils/codec.rs
+++ b/stream/src/utils/codec.rs
@@ -6,7 +6,7 @@ use commonware_runtime::{Sink, Stream};
 /// Returns an error if the message is too large or the stream is closed.
 pub async fn send_frame<S: Sink>(
     sink: &mut S,
-    buf: impl Buf + Send,
+    buf: impl Buf + Send + 'static,
     max_message_size: usize,
 ) -> Result<(), Error> {
     // Validate frame size
--- a/runtime/src/mocks.rs
+++ b/runtime/src/mocks.rs
@@ -50,7 +50,7 @@ pub struct Sink {
 }
 
 impl SinkTrait for Sink {
-    async fn send(&mut self, mut buf: impl Buf + Send) -> Result<(), Error> {
+    async fn send(&mut self, mut buf: impl Buf + Send + 'static) -> Result<(), Error> {
         let (os_send, data) = {
             let mut channel = self.channel.lock().unwrap();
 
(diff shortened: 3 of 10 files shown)


runtime/src/lib.rs
```diff
@@ -436,11 +436,15 @@ pub trait Listener: Sync + Send + 'static {
...
-    ) -> impl Future<Output = Result<(), Error>> + Send;
+    /// Send messages to the sink using vectored I/O.
+    ///
+    /// All buffers are written in order as if they were concatenated
+    /// into a single contiguous message. The implementation guarantees
+    /// that either all bytes are written or an error is returned.
+    ///
+    /// Implementations restrict the maximum number of buffers that can be
+    /// written at once to `16`.
+    fn send(&mut self, bufs: &[&[u8]]) -> impl Future<Output = Result<(), Error>> + Send;
```
patrick-ogrady: We need to support `StableBuf` for iouring safety

patrick-ogrady: (unless we were doing something very dumb?)

clabby: We do need to keep the buffer alive until the op finishes; still drafting the initial changes, PR is up to run tests on Linux.

clabby: For now (to retain cancel safety,) I've just gone ahead and allocated a stable buffer underneath the `Sink::send` impl for `io_uring`.

I'm not 100% sure if I love this, but the alternative would be to have `Sink::sink`'s signature be:

```rs
/// Interface that any runtime must implement to send
/// messages over a network connection.
pub trait Sink: Sync + Send + 'static {
    /// Send messages to the sink using vectored I/O.
    ///
    /// All buffers are written in order as if they were concatenated
    /// into a single contiguous message. The implementation guarantees
    /// that either all bytes are written or an error is returned.
    ///
    /// Implementations restrict the maximum number of buffers that can be
    /// written at once to `16`.
    fn send(&mut self, bufs: Vec<StableBuf>) -> impl Future<Output = Result<(), Error>> + Send;
}
```

which forces taking ownership over the buffers, even though that's only required for `io_uring` (given that `tokio`'s `write_vectored` function operates using the synchronous `writev`/`pwritev` syscalls.)

I'll think about this a bit. The current changes yield a big benefit for the base `tokio` network, but we still force a copy of the whole sent buffer(s) w/ `io_uring`.

We could potentially go all the way up the stack and require `Sender::send` + `send_frame` to take ownership over the data:

https://github.com/commonwarexyz/monorepo/blob/2be92d6e4ef04ac83ada7d82ec08b9abb80989d2/stream/src/lib.rs#L288-L307

patrick-ogrady: As discussed, I think taking StableBuf all the way at the top works (if it means we avoid all these copies)!

runtime/src/iouring/mod.rs
```diff
@@ -205,14 +210,22 @@ pub struct Op {
...
     /// completes, preventing write-after-free issues.
     pub buffer: Option<StableBuf>,
+    /// A boxed `Buf` to keep alive for the duration of the operation and return
+    /// to the caller. This is useful for vectored I/O where the original buffer
+    /// must remain valid and the caller needs it back to call `advance()`.
+    pub buf: Option<Box<dyn Buf + Send>>,
+    /// Additional data to keep alive for the duration of the operation but not
+    /// returned to the caller. Useful for things like iovec arrays that the kernel
+    /// references during the operation.
+    pub keepalive: Option<Box<dyn Send>>,
```
clabby, line 228: kept the original `StableBuf` in here given that storage still uses it; If we proceeded, we'd likely want to do a full swap of `StableBuf` -> `Bytes`/`Buf` throughout the storage and network primitives.

stream/src/utils/codec.rs
```diff
@@ -1,26 +1,25 @@
...
     }
 
-    // Prefix `buf` with its length and send it
-    let mut prefixed_buf = BytesMut::with_capacity(4 + buf.len());
     let len: u32 = n.try_into().map_err(|_| Error::SendTooLarge(n))?;
-    prefixed_buf.put_u32(len);
-    prefixed_buf.extend_from_slice(buf);
-    sink.send(prefixed_buf).await.map_err(Error::SendFailed)
+    let len_bytes = len.to_be_bytes();
+    sink.send(Bytes::from_owner(len_bytes).chain(buf))
```
patrick-ogrady, line 20: Ah, this is the benefit of non-contiguous bytes here?

clabby: Yep. `Buf` operates over chunks, so you can have a reader that spans non-contiguous memory. Really useful in cases like here where we want to not re-allocate a large slab. There's a few other areas of the codebase that this pattern would be nice to use in as well.

In general, anywhere you would use `Vec<u8>` / `Bytes` as an input argument, make it `impl Buf`, and always return `Bytes` > `Vec<u8>`.

clabby:
#2558

PR #2525, opened 2025-12-16
[runtime] Support DNS Resolution

patrick-ogrady:
Related: #2033

### DIFF
--- a/runtime/src/deterministic.rs
+++ b/runtime/src/deterministic.rs
@@ -1401,7 +1401,9 @@ mod tests {
     use crate::FutureExt;
     #[cfg(feature = "external")]
     use crate::Spawner;
-    use crate::{deterministic, reschedule, utils::run_tasks, Blob, Metrics, Runner as _, Storage};
+    use crate::{
+        deterministic, reschedule, utils::run_tasks, Blob, Metrics, Resolver, Runner as _, Storage,
+    };
     use commonware_macros::test_traced;
     #[cfg(not(feature = "external"))]
     use futures::future::pending;
@@ -1586,8 +1588,6 @@ mod tests {
 
     #[test]
     fn test_recover_dns_mappings_persist() {
-        use crate::Resolver;
-
         // Initialize the first runtime
         let executor = deterministic::Runner::default();
         let host = "example.com";


runtime/src/deterministic.rs
```diff
@@ -925,6 +928,27 @@ impl Context {
     pub fn auditor(&self) -> Arc<Auditor> {
         self.executor().auditor.clone()
     }
+
+    /// Register a DNS mapping for a hostname.
+    ///
+    /// If `addrs` is `None`, the mapping is removed.
+    /// If `addrs` is `Some`, the mapping is added or updated.
+    pub fn register(&self, host: impl Into<String>, addrs: Option<Vec<IpAddr>>) {
```
patrick-ogrady: Figured it was worth making this dynamic to unblock testing of dynamic DNS in `p2p`.

runtime/src/lib.rs
```diff
@@ -414,6 +416,17 @@ pub trait Network: Clone + Send + Sync + 'static {
     ) -> impl Future<Output = Result<(SinkOf<Self>, StreamOf<Self>), Error>> + Send;
 }
 
+/// Interface for DNS resolution.
+pub trait Resolver: Clone + Send + Sync + 'static {
```
patrick-ogrady, line 420: The alternative here would be to define a DNS trait in `p2p` but given this uses the host's DNS configuration, I felt it was correct.

runtime/src/deterministic.rs
```diff
@@ -925,6 +928,27 @@ impl Context {
...
+        let executor = self.executor();
+        let host = host.into();
+        executor.auditor.event(b"register", |hasher| {
+            hasher.update(host.as_bytes());
```
patrick-ogrady, line 944: add addrs here

andresilva (dismissed):
I guess the reason to use hickory rather than libc getaddrinfo is to provide an async interface? Just worried that hickory might not exactly mimic what libc does (e.g. https://github.com/hickory-dns/hickory-dns/issues/2251 https://github.com/hickory-dns/hickory-dns/issues/250 https://github.com/hickory-dns/hickory-dns/issues/251). But maybe let's roll with this and revert to using libc if we encounter real-world issues.

patrick-ogrady:
> I guess the reason to use hickory rather than libc getaddrinfo is to provide an async interface? Just worried that hickory might not exactly mimic what libc does (e.g. [hickory-dns/hickory-dns#2251](https://github.com/hickory-dns/hickory-dns/issues/2251) [hickory-dns/hickory-dns#250](https://github.com/hickory-dns/hickory-dns/issues/250) [hickory-dns/hickory-dns#251](https://github.com/hickory-dns/hickory-dns/issues/251)). But maybe let's roll with this and revert to using libc if we encounter real-world issues.

Yeah, exactly. Its async 👍

patrick-ogrady:
> > I guess the reason to use hickory rather than libc getaddrinfo is to provide an async interface? Just worried that hickory might not exactly mimic what libc does (e.g. [hickory-dns/hickory-dns#2251](https://github.com/hickory-dns/hickory-dns/issues/2251) [hickory-dns/hickory-dns#250](https://github.com/hickory-dns/hickory-dns/issues/250) [hickory-dns/hickory-dns#251](https://github.com/hickory-dns/hickory-dns/issues/251)). But maybe let's roll with this and revert to using libc if we encounter real-world issues.
> 
> Yeah, exactly. Its async 👍

Opted to just use OS for now.

runtime/src/deterministic.rs
```diff
@@ -925,6 +928,44 @@ impl Context {
     pub fn auditor(&self) -> Arc<Auditor> {
         self.executor().auditor.clone()
     }
+
+    /// Register a DNS mapping for a hostname.
+    ///
+    /// If `addrs` is `None`, the mapping is removed.
+    /// If `addrs` is `Some`, the mapping is added or updated.
+    pub fn register(&self, host: impl Into<String>, addrs: Option<Vec<IpAddr>>) {
```
clabby: nit: `register` is fairly generic, let's rename to `register_dns` so it's a bit more self-descriptive.

andresilva (dismissed):

runtime/src/lib.rs
```diff
@@ -2625,6 +2639,39 @@ mod tests {
...
+            let result = context.resolve("example.com").await;
+            assert!(matches!(result, Err(Error::ResolveFailed(_))));
+        });
+    }
+
+    #[test]
+    fn test_tokio_resolver() {
+        let executor = tokio::Runner::default();
+        executor.start(|context| async move {
+            let addrs = context.resolve("commonware.xyz").await.unwrap();
```
clabby: Maybe let's use `localhost`? This test will fail when we're developing offline.

clabby (dismissed):

codec/src/types/net.rs
```diff
@@ -107,30 +106,58 @@ impl Write for SocketAddr {
...
+        self.ip().encode_size() + u16::SIZE
+    }
+}
+
+impl Read for SocketAddr {
+    type Cfg = ();
+
+    #[inline]
+    fn read_cfg(buf: &mut impl Buf, _: &()) -> Result<Self, Error> {
+        let ip = IpAddr::read(buf)?;
```
patrick-ogrady, line 155: Much cleaner!

runtime/src/deterministic.rs
```diff
@@ -1536,6 +1584,38 @@ mod tests {
         });
     }
 
+    #[test]
+    fn test_recover_dns_mappings_persist() {
+        use crate::Resolver;
```
patrick-ogrady: move this

codec/src/types/net.rs
patrick-ogrady: (as confirmed by `conformance`, doesn't actually change the wire format either!)

patrick-ogrady: (Opted to retain direct SocketAddrV4/SocketAddrV6 encoding for those that want it)

PR #2526, opened 2025-12-16
[storage/qmdb] add capability to stream active key/value pairs in key order

roberto-bayardo:
Adds a `stream_range` capability to ordered databases that allows iterating over all active key/value pairs in key-order from a designated starting point.

This initial implementation should be fine for streaming small ranges of keys, but for larger ranges it could be made much more efficient, which we are tracking here: https://github.com/commonwarexyz/monorepo/issues/2527

Extends the Ordered index trait to return whether a previous/next "cycled".

### DIFF
--- a/storage/src/index/mod.rs
+++ b/storage/src/index/mod.rs
@@ -182,6 +182,14 @@ pub trait Ordered: Unordered {
     // translated key if `key` is greater than or equal to the last translated key. The returned
     // boolean indicates whether the result is from cycling. Returns None if there are no keys in
     // the index.
+    ///
+    /// For example, if the translator is looking only at the first byte of a key, and the index
+    /// contains values for translated keys 0b, 1c, and 2d, then `get_next([0b, 01, 02, ...])` would
+    /// return the values associated with 1c, `get_next([2a, 01, 02, ...])` would return the values
+    /// associated with 2d, and `get_next([2d])` would "cycle around" to the values associated with
+    /// 0b, returning true for the bool. Because values associated with the same translated key can
+    /// appear in any order, keys with the same first byte in this example would need to be ordered
+    /// by the caller if a full ordering over the untranslated keyspace is desired.
     fn next_translated_key<'a>(&'a self, key: &[u8]) -> Option<(Self::Iterator<'a>, bool)>
     where
         Self::Value: 'a;
--- a/storage/src/index/ordered.rs
+++ b/storage/src/index/ordered.rs
@@ -167,16 +167,6 @@ impl<T: Translator, V: Eq> Ordered for Index<T, V> {
         self.last_translated_key().map(|res| (res, true))
     }
 
-    /// Get the values associated with the translated key that lexicographically follows the result
-    /// of translating `key`.
-    ///
-    /// For example, if the translator is looking only at the first byte of a key, and the index
-    /// contains values for translated keys 0b, 1c, and 2d, then `get_next([0b, 01, 02, ...])` would
-    /// return the values associated with 1c, `get_next([2a, 01, 02, ...])` would return the values
-    /// associated with 2d, and `get_next([2d])` would return `None`. Because values associated with
-    /// the same translated key can appear in any order, keys with the same first byte in this
-    /// example would need to be ordered by the caller if a full ordering over the untranslated
-    /// keyspace is desired.
     fn next_translated_key<'a>(&'a self, key: &[u8]) -> Option<(Self::Iterator<'a>, bool)>
     where
         Self::Value: 'a,
@@ -189,8 +179,6 @@ impl<T: Translator, V: Eq> Ordered for Index<T, V> {
         self.first_translated_key().map(|res| (res, true))
     }
 
-    /// Get the values associated with the lexicographically first translated key, or None if the
-    /// index is empty.
     fn first_translated_key<'a>(&'a self) -> Option<Self::Iterator<'a>>
     where
         Self::Value: 'a,
@@ -200,8 +188,6 @@ impl<T: Translator, V: Eq> Ordered for Index<T, V> {
             .map(|(_, record)| ImmutableCursor::new(record))
     }
 
-    /// Get the values associated with the lexicographically last translated key, or None if the
-    /// index is empty.
     fn last_translated_key<'a>(&'a self) -> Option<Self::Iterator<'a>>
     where
         Self::Value: 'a,


storage/src/qmdb/any/ordered/mod.rs
```diff
@@ -320,6 +323,77 @@ where
...
+
+                if let Some(iter) = self.snapshot.next_translated_key(&driver_key) {
+                    // TODO(https://github.com/commonwarexyz/monorepo/issues/2527): concurrently
+                    // fetch a much larger batch of "pending" keys.
+                    match self.fetch_all_updates(iter, Some(&driver_key)).await {
+                        Ok(None) => None,
+                        Ok(Some(mut items)) => {
+                            let item = items
+                                .pop()
+                                .unwrap_or_else(|| unreachable!("items should not be empty"));
```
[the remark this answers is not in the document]
roberto-bayardo: I think the brief message is fine.

storage/src/qmdb/any/ordered/mod.rs
```diff
@@ -320,6 +323,77 @@ where
...
+            .fetch_all_updates(start_iter, None)
+            .await?
+            .expect("wrap-around shouldn't happen");
+        init_pending.retain(|x| x.key >= *start);
+
+        Ok(stream::unfold(
+            (start.clone(), init_pending),
+            move |(driver_key, mut pending): (K, Vec<KeyData<K, V::Value>>)| async move {
+                if !pending.is_empty() {
+                    let item = pending.pop().expect("pending is not empty");
```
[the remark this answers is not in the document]
roberto-bayardo: It's already plenty clear.

storage/src/qmdb/any/ordered/mod.rs
```diff
@@ -320,6 +323,76 @@ where
...
+            .await?
+            .expect("wrap-around shouldn't happen");
+        init_pending.retain(|x| x.key >= start);
+
+        Ok(stream::unfold(
+            (start, init_pending),
+            move |(driver_key, mut pending): (K, Vec<KeyData<K, V::Value>>)| async move {
+                if !pending.is_empty() {
+                    let item = pending.pop().expect("pending is not empty");
+                    return Some((Ok((item.key, item.value)), (driver_key, pending)));
```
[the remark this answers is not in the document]
roberto-bayardo: fixed from the above change.

storage/src/qmdb/any/ordered/mod.rs
```diff
@@ -1873,4 +1947,122 @@ mod test {
...
+            // Start key is in the DB.
+            {
+                let mut stream = db.stream_range(key1.clone()).await.unwrap().boxed_local();
+                assert_eq!(stream.next().await.unwrap().unwrap().0, key1);
+                assert!(stream.next().await.is_none());
+            }
+
+            // Start key collides & precedes the only key in the db.
+            {
+                let start = FixedBytes::from([0x00u8, 0x00, 0x00, 0x01]);
```
danlaine: `start` and `key1` don't collide (`0x00, 0x00` and `0x10, 0x00`)

storage/src/qmdb/any/ordered/mod.rs
```diff
@@ -1873,4 +1947,122 @@ mod test {
...
+                let mut stream = db.stream_range(start).await.unwrap().boxed_local();
+                assert_eq!(stream.next().await.unwrap().unwrap().0, key1);
+                assert!(stream.next().await.is_none());
+            }
+
+            // Start key collides & follows the only key in the dsb.
+            {
+                let start = FixedBytes::from([0x20u8, 0x00, 0x00, 0xFF]);
```
danlaine: `start` and `key1` don't collide

storage/src/qmdb/any/ordered/mod.rs
```diff
@@ -1873,4 +1947,122 @@ mod test {
...
+
+            // Start key collides & precedes the only key in the db.
+            {
+                let start = FixedBytes::from([0x00u8, 0x00, 0x00, 0x01]);
+                let mut stream = db.stream_range(start).await.unwrap().boxed_local();
+                assert_eq!(stream.next().await.unwrap().unwrap().0, key1);
+                assert!(stream.next().await.is_none());
+            }
+
+            // Start key collides & follows the only key in the dsb.
```
danlaine:
```suggestion
            // Start key collides & follows the only key in the db.
```

storage/src/qmdb/any/ordered/mod.rs
```diff
@@ -1873,4 +1947,122 @@ mod test {
...
+            {
+                let start = FixedBytes::from([0x20u8, 0x00, 0x00, 0xFF]);
+                let mut stream = db.stream_range(start).await.unwrap().boxed_local();
+                assert!(stream.next().await.is_none());
+            }
+
+            // Start key precedes the key in the DB without colliding.
+            {
+                let start = FixedBytes::from([0x00u8, 0x00, 0x00, 0x01]);
```
danlaine, line 1492: This test case exists above, too

storage/src/qmdb/any/ordered/mod.rs
```diff
@@ -1873,4 +1947,122 @@ mod test {
...
+        executor.start(|context| async move {
+            let mut db = open_fixed_db(context.clone()).await;
+
+            let key1 = FixedBytes::from([0x10u8, 0x00, 0x00, 0x05]);
+            let val = Sha256::fill(1u8);
+
+            // Test the single-bucket case.
+            db.create(key1.clone(), val).await.unwrap();
+            db.commit(None).await.unwrap();
+
```
danlaine, line 1467: LLM found a unit test that fails if you put it here:

```rs
            // BUG DEMONSTRATION: Start key collides (same translated key) and precedes key1.
            // With TwoCap translator, keys with same first 2 bytes share a bucket.
            // key1 = [0x10, 0x00, 0x00, 0x05], start = [0x10, 0x00, 0x00, 0x00]
            // Both have translated key 0x1000, but start < key1 lexicographically.
            // Expected: stream returns [key1] once.
            // Actual (BUG): stream returns [key1, key1] because driver_key isn't updated
            // when draining init_pending, causing the wrap-around check to fail.
            {
                let start = FixedBytes::from([0x10u8, 0x00, 0x00, 0x00]); // Same bucket as key1
                let mut stream = db.stream_range(start).await.unwrap().boxed_local();
                assert_eq!(stream.next().await.unwrap().unwrap().0, key1);
                // This assertion will FAIL - a second key1 is incorrectly returned
                assert!(stream.next().await.is_none(), "BUG: key1 returned twice!");
            }
```

roberto-bayardo: yep the test case (now that I've fixed it) is picking this up.  This seems to be a downside of pushing the wrap-around logic down. Wondering if I should change the API to return a bool indicating if the wrap took place.

roberto-bayardo: Given how error prone cycling behavior can be, I opted to make it explicit in the return result, and changed the implementation to use that rather than trying to detect it.

storage/src/qmdb/any/ordered/mod.rs
roberto-bayardo: only because the above one wasn't as intended I think, so should be fixed.

storage/src/qmdb/any/ordered/mod.rs
roberto-bayardo: fixed

storage/src/qmdb/any/ordered/mod.rs
roberto-bayardo: fixed

storage/src/qmdb/any/ordered/mod.rs
roberto-bayardo: fixed, and this now catches the bug the LLM pointed out.

roberto-bayardo:
pretty sure this will need to rebase once your change merges ahead of it!

storage/src/qmdb/any/ordered/mod.rs
```diff
@@ -191,6 +194,63 @@ where
...
+                // TODO(https://github.com/commonwarexyz/monorepo/issues/2527): concurrently
+                // fetch a much larger batch of "pending" keys.
+                match self.fetch_all_updates(iter).await {
+                    Ok(mut pending) => {
+                        let item = pending.pop().expect("pending is not empty");
+                        let key = item.key.clone();
+                        Some((Ok((item.key, item.value)), (key, pending)))
+                    }
+                    Err(e) => Some((Err(e), (driver_key, pending))),
```
patrick-ogrady, line 234: I think we want to jump to the end here/breakout (otherwise we can return the same error in a loop forever)?

patrick-ogrady (commented):
one small bug

storage/src/qmdb/any/ordered/mod.rs
patrick-ogrady: Disregard. This is consistent with how we handle this elsewhere (we expect consumer to stop on err)

roberto-bayardo: The caller will typically abort after the first error. If they don't then they are kind of asking for it :)

PR #2529, opened 2025-12-16
[consensus/simplex] make leader election pluggable

### LINKED ISSUE #2467: [consensus/simplex] Make `signing_scheme::bls12381_threshold` Leader Election Generic
Instead of requiring random leader election if a "seed" exists (via the embedded VRF), applications should be allowed to provide their own leader election function when using `bls12381_threshold`: https://github.com/commonwarexyz/monorepo/blob/ecc04c8cf679b64cc2f201c763c962cfd8089ce4/consensus/src/simplex/actors/voter/round.rs#L135-L147

This makes the changes in #2380 more interesting because someone that only cares about certificates (#2466) doesn't need to compute the leader somehow.

andresilva:
This PR introduces a pluggable `Elector` trait that enables applications to customize how leaders are selected for each consensus round. Previously, leader selection in simplex was tightly coupled to the signing scheme through the `SeededScheme` trait. By decoupling leader election from the signing scheme, applications can now use whatever leader election logic they see fit and implement custom strategies.

The `Elector` trait is simple:

```rust
pub trait Elector<S: Scheme>: Send + 'static {
    fn elect(&self, round: Round, certificate: Option<&S::Certificate>) -> u32;
}
```

Participants are provided at construction time rather than at each `elect()` call. This enables electors to precompute data structures (like shuffled permutations) once and reuse them. The certificate parameter is `None` only for view 1, for subsequent views the certificate from the previous view is provided.

Two built-in electors are provided:
- `RoundRobin`: Deterministic rotation through participants based on `(epoch + view) % n`. Supports an optional seed for shuffling the rotation order via `shuffled(participants, seed)`.
- `Random`: Uses randomness derived from BLS threshold signatures for unpredictable leader selection. Falls back to round-robin for view 1 when no certificate is available. Only works with bls12381_threshold scheme.

Applications can implement custom electors for strategies like stake-weighted selection or hybrid approaches that combine deterministic and random selection. 

The `Elector` trait is threaded through the engine from the configuration. The `SeededScheme` trait has been removed entirely since seed extraction is now the elector's responsibility.

Closes #2467.

### DIFF
--- a/consensus/src/simplex/actors/voter/mod.rs
+++ b/consensus/src/simplex/actors/voter/mod.rs
@@ -5,7 +5,7 @@ mod slot;
 mod state;
 
 use crate::{
-    simplex::{elector::Elector, types::Activity},
+    simplex::{elector::Config as Elector, types::Activity},
     types::{Epoch, ViewDelta},
     Automaton, Relay, Reporter,
 };
@@ -53,7 +53,7 @@ mod tests {
     use crate::{
         simplex::{
             actors::{batcher, resolver},
-            elector::{Elector, Random, RoundRobin},
+            elector::{Config as ElectorConfig, Elector, Random, RoundRobin, RoundRobinElector},
             mocks,
             scheme::{bls12381_multisig, bls12381_threshold, ed25519, Scheme},
             types::{Certificate, Finalization, Finalize, Notarization, Notarize, Proposal, Vote},
@@ -129,7 +129,7 @@ mod tests {
     where
         S: Scheme<Sha256Digest, PublicKey = PublicKey>,
         F: FnMut(&mut deterministic::Context, u32) -> Fixture<S>,
-        L: Elector<S>,
+        L: ElectorConfig<S>,
     {
         let n = 5;
         let quorum = quorum(n);
@@ -364,7 +364,7 @@ mod tests {
     where
         S: Scheme<Sha256Digest, PublicKey = PublicKey>,
         F: FnMut(&mut deterministic::Context, u32) -> Fixture<S>,
-        L: Elector<S>,
+        L: ElectorConfig<S>,
     {
         let n = 5;
         let quorum = quorum(n);
@@ -643,7 +643,7 @@ mod tests {
     where
         S: Scheme<Sha256Digest, PublicKey = PublicKey>,
         F: FnMut(&mut deterministic::Context, u32) -> Fixture<S>,
-        L: Elector<S>,
+        L: ElectorConfig<S>,
     {
         let n = 5;
         let quorum = quorum(n);
@@ -825,7 +825,7 @@ mod tests {
     where
         S: Scheme<Sha256Digest, PublicKey = PublicKey>,
         F: FnMut(&mut deterministic::Context, u32) -> Fixture<S>,
-        L: Elector<S>,
+        L: ElectorConfig<S>,
     {
         let n = 5;
         let quorum = quorum(n);
@@ -1020,7 +1020,7 @@ mod tests {
     where
         S: Scheme<Sha256Digest, PublicKey = PublicKey>,
         F: FnMut(&mut deterministic::Context, u32) -> Fixture<S>,
-        L: Elector<S>,
+        L: ElectorConfig<S>,
     {
         let n = 5;
         let quorum = quorum(n);
@@ -1198,7 +1198,7 @@ mod tests {
     where
         S: Scheme<Sha256Digest, PublicKey = PublicKey>,
         F: FnMut(&mut deterministic::Context, u32) -> Fixture<S>,
-        L: Elector<S>,
+        L: ElectorConfig<S>,
     {
         let n = 5;
         let quorum = quorum(n);
@@ -1403,9 +1403,10 @@ mod tests {
 
             // Figure out who the leader will be for view 2
             let view2_round = Round::new(epoch, View::new(2));
-            let mut elector = <RoundRobin>::default();
-            Elector::<S>::initialize(&mut elector, schemes[0].participants());
-            let leader_idx = Elector::<S>::elect(&elector, view2_round, None);
+            let elector_config = RoundRobin::<Sha256>::default();
+            let temp_elector: RoundRobinElector<S> =
+                elector_config.clone().build(schemes[0].participants());
+            let leader_idx = temp_elector.elect(view2_round, None);
             let leader = participants[leader_idx as usize].clone();
 
             // Create a voter with the leader's identity
@@ -1429,7 +1430,7 @@ mod tests {
                 namespace: namespace.clone(),
                 participants: participants.clone().try_into().unwrap(),
                 scheme: leader_scheme.clone(),
-                elector: elector.clone(),
+                elector: elector_config.clone(),
             };
             let reporter =
                 mocks::reporter::Reporter::new(context.with_label("reporter"), reporter_cfg);
@@ -1437,7 +1438,7 @@ mod tests {
             // Initialize voter actor
             let voter_cfg = Config {
                 scheme: leader_scheme.clone(),
-                elector,
+                elector: elector_config,
                 blocker: oracle.control(leader.clone()),
                 automaton: application.clone(),
                 relay: application.clone(),
@@ -1592,7 +1593,7 @@ mod tests {
     where
         S: Scheme<Sha256Digest, PublicKey = PublicKey>,
         F: FnMut(&mut deterministic::Context, u32) -> Fixture<S>,
-        L: Elector<S>,
+        L: ElectorConfig<S>,
     {
         let n = 5;
         let quorum = quorum(n);
@@ -1820,7 +1821,7 @@ mod tests {
     where
         S: Scheme<Sha256Digest, PublicKey = PublicKey>,
         F: FnMut(&mut deterministic::Context, u32) -> Fixture<S>,
-        L: Elector<S>,
+        L: ElectorConfig<S>,
     {
         // This is a regression test as the resolver didn't use to send
         // finalizations to the voter
@@ -1983,7 +1984,7 @@ mod tests {
     where
         S: Scheme<Sha256Digest, PublicKey = PublicKey>,
         F: FnMut(&mut deterministic::Context, u32) -> Fixture<S>,
-        L: Elector<S>,
+        L: ElectorConfig<S>,
     {
         let n = 5;
         let quorum = quorum(n);
@@ -2155,7 +2156,7 @@ mod tests {
     where
         S: Scheme<Sha256Digest, PublicKey = PublicKey>,
         F: FnMut(&mut deterministic::Context, u32) -> Fixture<S>,
-        L: Elector<S>,
+        L: ElectorConfig<S>,
     {
         let n = 5;
         let quorum = quorum(n);
--- a/consensus/src/simplex/actors/voter/state.rs
+++ b/consensus/src/simplex/actors/voter/state.rs
@@ -1,7 +1,7 @@
 use super::round::Round;
 use crate::{
     simplex::{
-        elector::Elector,
+        elector::{Config as ElectorConfig, Elector},
         interesting, min_active,
         scheme::Scheme,
         types::{
@@ -28,7 +28,7 @@ use tracing::{debug, warn};
 const GENESIS_VIEW: View = View::zero();
 
 /// Configuration for initializing [`State`].
-pub struct Config<S: certificate::Scheme, L: Elector<S>> {
+pub struct Config<S: certificate::Scheme, L: ElectorConfig<S>> {
     pub scheme: S,
     pub elector: L,
     pub namespace: Vec<u8>,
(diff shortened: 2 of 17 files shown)


consensus/src/simplex/actors/voter/round.rs
```diff
@@ -132,18 +129,11 @@ impl<S: SeededScheme, D: Digest> Round<S, D> {
...
-            seed,
-        );
-        debug!(round=?self.round, ?leader, ?leader_idx, "leader elected");
-        self.leader = Some(Leader {
-            idx: leader_idx,
-            key: leader,
-        });
+    /// Sets the leader for this round using the pre-computed leader index.
+    pub fn set_leader(&mut self, leader: u32) {
+        let key = self.scheme.participants()[leader as usize].clone();
```
patrick-ogrady: nit: `.position()`? or `.index()` ... I forgot

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -130,16 +136,34 @@ impl<E: Clock + Rng + CryptoRng + Metrics, S: Scheme<D>, D: Digest> State<E, S,
...
+    /// If `seed` is `None`, this **must** be the first view after genesis (view 1).
+    /// For all subsequent views, a seed derived from the previous view's certificate
+    /// must be provided.
+    fn enter_view(&mut self, view: View, seed: Option<L::Seed>) -> bool {
         if view <= self.view {
             return false;
         }
+
+        // Seed is required for all views after view 1
+        assert!(seed.is_some() || view == GENESIS_VIEW.next());
```
patrick-ogrady: Not sure I love this (I think just `.elect(participants, round, seed)` is fine.

consensus/src/simplex/scheme/bls12381_threshold.rs
```diff
@@ -696,14 +696,6 @@ impl<P: PublicKey, V: Variant + Send + Sync> certificate::Scheme for Scheme<P, V
     fn certificate_codec_config_unbounded() -> <Self::Certificate as Read>::Cfg {}
 }
 
-impl<P: PublicKey, V: Variant + Send + Sync> SeededScheme for Scheme<P, V> {
```
patrick-ogrady, line 699: We can't switch to just using the macro because we "overload" the message, right?

consensus/src/simplex/config.rs
```diff
@@ -29,6 +33,14 @@ pub struct Config<
     /// the wrong validator.
     pub scheme: S,
 
+    /// Leader election strategy.
+    ///
+    /// Determines how leaders are selected for each view. Built-in options include
+    /// [`RoundRobin`](super::elector::RoundRobin) for deterministic rotation and
+    /// [`ThresholdRandomness`](super::elector::ThresholdRandomness) for unpredictable
```
patrick-ogrady: nit: `Random`

patrick-ogrady (commented):
Looks very reasonable

consensus/src/simplex/actors/voter/state.rs
patrick-ogrady: We can infer `first` in there?

andresilva: Do you mean just remove this assert? Or do you mean change `elect` to take `Option<Seed>`? I split the interface into `first` / `elect` because otherwise you always have to deal with the `None` case in `elect`, e.g. in the `Random` variant we'd need to fallback to round-robin if no seed is provided (which should only happen for the first view since there's no certificate for genesis). I guess we can collapse `first` / `elect` if you are OK with panicking on the implementation if no seed is provided for view > 1.

consensus/src/simplex/scheme/bls12381_threshold.rs
andresilva: Yeah, for simplex our threshold implementation does two signatures one for the "vote" another for the "seed". I guess you could use the generic threshold scheme with `RoundRobin` elector if you never care about having a randomness beacon.

consensus/src/simplex/elector.rs
```diff
@@ -0,0 +1,354 @@
+//! Leader election strategies for simplex consensus.
+//!
+//! This module provides the [`Elector`] trait for customizing how leaders are selected
+//! for each consensus round, along with built-in implementations.
+//!
+//! # Built-in Electors
+//!
+//! - [`RoundRobin`]: Simple deterministic rotation through participants based on view number.
```
patrick-ogrady: Would be nice for this to take some randomness on new (just once) to compute a permutation of leader order (just so it isn't the same neighbors in a stable peer set each epoch).

consensus/src/simplex/actors/voter/mod.rs
```diff
@@ -1540,17 +1602,20 @@ mod tests {
         });
     }
 
-    #[test]
+    #[test_traced]
     fn test_drop_our_proposal_on_conflict() {
+        drop_our_proposal_on_conflict(bls12381_threshold::fixture::<MinPk, _>);
```
andresilva, line 1585: Now we can run this test with BLS threshold using round-robin elector 🧠

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -204,23 +219,15 @@ impl<E: Clock + Rng + CryptoRng + Metrics, S: Scheme<D>, D: Digest> State<E, S,
...
     ) -> (bool, Option<S::PublicKey>) {
         let view = notarization.view();
-        let seed = self
-            .scheme
-            .seed(notarization.round(), &notarization.certificate);
-        let added = self.create_round(view).add_notarization(notarization);
-        self.enter_view(view.next(), seed);
-        added
+        self.enter_view(view.next(), Some(&notarization.certificate));
+        self.create_round(view).add_notarization(notarization)
```
andresilva, line 228: The order of these operations is switched to avoid having to clone the certificate, AFAIK there are no issues with switching these since `enter_view` will operate on the next view's round, but can revert and just clone it.

consensus/src/simplex/config.rs
```diff
@@ -106,12 +118,13 @@ pub struct Config<
...
         B: Blocker<PublicKey = S::PublicKey>,
         D: Digest,
         A: Automaton<Context = Context<D, S::PublicKey>>,
         R: Relay,
         F: Reporter<Activity = Activity<S, D>>,
-    > Config<S, B, D, A, R, F>
+    > Config<S, L, B, D, A, R, F>
 {
     /// Assert enforces that all configuration values are valid.
     pub fn assert(&self) {
```
andresilva, line 130: Guess one assertion that could make sense here is that `scheme.participants() == elector.participants()`, but this would require adding a `participants()` to the `Elector` trait, which would require instances to store the participants (which they don't do right now). Let me know if you think it's worth it.

consensus/src/simplex/actors/voter/state.rs
patrick-ogrady: I think this is fine.

consensus/src/simplex/elector.rs
```diff
@@ -0,0 +1,386 @@
...
+    /// The seed determines the permutation of leader indices, so different seeds
+    /// produce different rotation orders.
+    pub fn shuffled(participants: &Set<S::PublicKey>, seed: u64) -> Self {
+        assert!(!participants.is_empty(), "no participants");
+
+        let mut permutation = (0..participants.len() as u32).collect::<Vec<_>>();
+        permutation.sort_by_key(|&index| {
+            let mut hasher = Blake3::new();
+            hasher.update(&seed.to_le_bytes());
+            hasher.update(&index.to_le_bytes());
```
patrick-ogrady: nice...this is clever.

consensus/src/simplex/elector.rs
```diff
@@ -0,0 +1,386 @@
...
+    /// Creates a new round-robin elector with shuffled order based on seed.
+    ///
+    /// The seed determines the permutation of leader indices, so different seeds
+    /// produce different rotation orders.
+    pub fn shuffled(participants: &Set<S::PublicKey>, seed: u64) -> Self {
+        assert!(!participants.is_empty(), "no participants");
+
+        let mut permutation = (0..participants.len() as u32).collect::<Vec<_>>();
+        permutation.sort_by_key(|&index| {
+            let mut hasher = Blake3::new();
```
patrick-ogrady: We can make shuffled generic over Hasher (instead of enshrining Blake3).

I also think we can just use update/update/finalize (don't need to create in each op).

I also think we can just call ".encode()" on seed/index.

consensus/src/simplex/elector.rs
```diff
@@ -0,0 +1,386 @@
...
+    permutation: Vec<u32>,
+    _phantom: PhantomData<S>,
+}
+
+impl<S: Scheme> RoundRobin<S> {
+    /// Creates a new round-robin elector with shuffled order based on seed.
+    ///
+    /// The seed determines the permutation of leader indices, so different seeds
+    /// produce different rotation orders.
+    pub fn shuffled(participants: &Set<S::PublicKey>, seed: u64) -> Self {
```
patrick-ogrady: I think we should accept `&[u8]` here for seed (we may take a hash).

consensus/src/simplex/config.rs
```diff
@@ -29,6 +33,14 @@ pub struct Config<
     /// the wrong validator.
     pub scheme: S,
 
+    /// Leader election strategy.
+    ///
+    /// Determines how leaders are selected for each view. Built-in options include
+    /// [`RoundRobin`](super::elector::RoundRobin) for deterministic rotation and
+    /// [`Random`](super::elector::Random) for unpredictable selection using BLS
+    /// threshold signatures.
+    pub elector: L,
```
patrick-ogrady, line 42: Ah, we need to provide it because we may seed it externally...

consensus/src/simplex/config.rs
patrick-ogrady: Good question...maybe we refactor elector to have a separate `initialize` method that is called within consensus (rather than externally) with participants? 

So you can seed it with whatever, but we'll handle init 🤔

consensus/src/simplex/config.rs
andresilva: The trait only guarantees a `new` constructor (which is fine to abstract over in tests), but it's easy to imagine other implementations might have their own constructor.

consensus/src/simplex/config.rs
andresilva: That makes sense!

consensus/src/simplex/elector.rs
andresilva: ~~We should recreate the hasher on every iteration because the order of `sort_by_key` is unspecified.~~ nevermind

consensus/src/simplex/elector.rs
```diff
@@ -0,0 +1,438 @@
...
+}
+
+/// Simple round-robin leader election.
+///
+/// Rotates through participants based on `(epoch + view) % num_participants`.
+/// The rotation order can be shuffled at construction using a seed.
+///
+/// Works with any signing scheme.
+#[derive(Clone, Debug)]
+pub struct RoundRobin<S: Scheme, H: Hasher = Blake3> {
```
andresilva: The default hasher is done for convenience, otherwise users that are not interested in shuffling will still have to provide this type parameter for no reason.

patrick-ogrady: reasonable

consensus/src/simplex/elector.rs
```diff
@@ -47,18 +53,27 @@ use std::marker::PhantomData;
...
+pub trait Elector<S: Scheme>: Clone + Default + Send + 'static {
+    /// Initializes the elector with the participant set.
+    ///
+    /// Called by consensus before any [`elect`](Elector::elect) calls.
     ///
     /// # Panics
     ///
     /// Implementations should panic if `participants` is empty.
-    fn new(participants: &Set<S::PublicKey>) -> Self;
+    fn initialize(&mut self, participants: &Set<S::PublicKey>);
```
patrick-ogrady: Arguably this should transform `Elector` such that we can't accept the previously initialized struct (like in `coding`).

consensus/src/simplex/elector.rs
```diff
@@ -71,45 +86,55 @@ pub trait Elector<S: Scheme>: Clone + Send + 'static {
 ///
 /// Works with any signing scheme.
 #[derive(Clone, Debug)]
-pub struct RoundRobin<S: Scheme> {
+pub struct RoundRobin<S: Scheme, H: Hasher = Blake3> {
```
patrick-ogrady: let's use Hasher = Sha256 for now (that's what we do elsewhere)

consensus/src/simplex/elector.rs
patrick-ogrady: Completed in #2545

consensus/src/simplex/config.rs
patrick-ogrady: Completed on #2545

PR #2531, opened 2025-12-16
[storage/qmdb] make qmdb associated value type require Clone

roberto-bayardo:
Changes the qmdb associated value type to require the Clone trait, which doesn't seem to have any significant downsides. 

Also changes ExclusionProof verification to accept the proof by reference, which is now possible thanks to being able to clone the value within it.

### DIFF
--- a/storage/src/qmdb/current/ordered.rs
+++ b/storage/src/qmdb/current/ordered.rs
@@ -156,7 +156,7 @@ impl<
     pub fn verify_exclusion_proof(
         hasher: &mut H,
         key: &K,
-        proof: ExclusionProof<K, V, H::Digest, N>,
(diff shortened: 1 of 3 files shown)


storage/src/qmdb/current/ordered.rs
```diff
@@ -171,14 +171,17 @@ impl<
                     return false;
                 }
 
-                (op_proof, Operation::Update(data))
+                (op_proof, Operation::Update(data.clone()))
```
patrick-ogrady, line 174: I suppose we should wait to find out how to clone `data` and `metadata` until we prove it is a bottleneck?

patrick-ogrady (approved):
Seems reasonable!

PR #2533, opened 2025-12-16
[p2p] Introduce `LimitedSender` & `CheckedSender`

clabby:
## Overview

Introduces the `LimitedSender` and `CheckedSender` abstractions, allowing for an optional check-and-send usage of p2p sender handles. `Sender` is retained, just using the check-and-send pattern under the hood.

### DIFF
--- a/p2p/src/simulated/network.rs
+++ b/p2p/src/simulated/network.rs
@@ -7,7 +7,7 @@ use super::{
     Error,
 };
 use crate::{
-    utils::limited::{LimitedSender, Peers},
+    utils::limited::{Connected, LimitedSender},
     Channel, CheckedSender as _, Message, Recipients,
 };
 use bytes::Bytes;
@@ -379,7 +379,7 @@ impl<E: RNetwork + Spawner + Rng + Clock + Metrics, P: PublicKey> Network<E, P> 
                 }
                 self.subscribers.push(sender);
             }
-            ingress::Message::SubscribePeers { response } => {
+            ingress::Message::SubscribeConnected { response } => {
                 // Create a ring channel for the subscriber
                 let (mut sender, receiver) = ring::channel(NZUsize!(1));
 
@@ -716,15 +716,15 @@ impl<E: RNetwork + Spawner + Rng + Clock + Metrics, P: PublicKey> Network<E, P> 
     }
 }
 
-/// Provides peer subscriptions for the simulated network.
+/// Provides online peers from the simulated network.
 ///
-/// Implements [`crate::utils::limited::Peers`] to provide peer list updates
+/// Implements [`crate::utils::limited::Connected`] to provide peer list updates
 /// to [`crate::utils::limited::LimitedSender`].
-pub struct PeerSource<P: PublicKey, E: Clock> {
+pub struct ConnectedPeerProvider<P: PublicKey, E: Clock> {
     sender: mpsc::UnboundedSender<ingress::Message<P, E>>,
 }
 
-impl<P: PublicKey, E: Clock> Clone for PeerSource<P, E> {
+impl<P: PublicKey, E: Clock> Clone for ConnectedPeerProvider<P, E> {
     fn clone(&self) -> Self {
         Self {
             sender: self.sender.clone(),
@@ -732,20 +732,20 @@ impl<P: PublicKey, E: Clock> Clone for PeerSource<P, E> {
     }
 }
 
-impl<P: PublicKey, E: Clock> PeerSource<P, E> {
+impl<P: PublicKey, E: Clock> ConnectedPeerProvider<P, E> {
     const fn new(sender: mpsc::UnboundedSender<ingress::Message<P, E>>) -> Self {
         Self { sender }
     }
 }
 
-impl<P: PublicKey, E: Clock> Peers for PeerSource<P, E> {
+impl<P: PublicKey, E: Clock> Connected for ConnectedPeerProvider<P, E> {
     type PublicKey = P;
 
     async fn subscribe(&mut self) -> ring::Receiver<Vec<Self::PublicKey>> {
         let (response_tx, response_rx) = oneshot::channel();
         let _ = self
             .sender
-            .unbounded_send(ingress::Message::SubscribePeers {
+            .unbounded_send(ingress::Message::SubscribeConnected {
                 response: response_tx,
             });
         // If the network is closed, return an empty receiver
@@ -808,7 +808,7 @@ impl<P: PublicKey> crate::Sender for UnlimitedSender<P> {
 /// Also implements [crate::LimitedSender] to support rate-limit checking
 /// before sending messages.
 pub struct Sender<P: PublicKey, E: Clock> {
-    limited_sender: LimitedSender<E, UnlimitedSender<P>, PeerSource<P, E>>,
+    limited_sender: LimitedSender<E, UnlimitedSender<P>, ConnectedPeerProvider<P, E>>,
 }
 
 impl<P: PublicKey, E: Clock> Clone for Sender<P, E> {
@@ -873,7 +873,7 @@ impl<P: PublicKey, E: Clock> Sender<P, E> {
             high,
             low,
         };
-        let peer_source = PeerSource::new(oracle_sender);
+        let peer_source = ConnectedPeerProvider::new(oracle_sender);
         let limited_sender = LimitedSender::new(unlimited_sender, quota, clock, peer_source);
 
         (Self { limited_sender }, processor)
--- a/p2p/src/utils/limited.rs
+++ b/p2p/src/utils/limited.rs
@@ -12,13 +12,15 @@ use std::{cmp, fmt, sync::Arc, time::SystemTime};
 ///
 /// Implementations must be clonable so that each clone of [`LimitedSender`]
 /// can establish its own peer subscription.
-pub trait Peers: Clone + Send + Sync + 'static {
+pub trait Connected: Clone + Send + Sync + 'static {
     type PublicKey: PublicKey;
 
     /// Subscribe to peer updates.
     ///
-    /// Returns a receiver that yields the current set of known peers
-    /// whenever it changes.
+    /// Returns a receiver that yields the current set of known peers whenever it changes.
+    ///
+    /// It is assumed that when a new subscription is created, the current set of known peers
+    /// is sent immediately.
     fn subscribe(&mut self) -> impl Future<Output = ring::Receiver<Vec<Self::PublicKey>>> + Send;
 }
 
@@ -27,7 +29,7 @@ pub struct LimitedSender<E, S, P>
 where
     E: Clock,
     S: Sender,
-    P: Peers<PublicKey = S::PublicKey>,
+    P: Connected<PublicKey = S::PublicKey>,
 {
     sender: S,
     rate_limit: Arc<Mutex<KeyedRateLimiter<S::PublicKey, E>>>,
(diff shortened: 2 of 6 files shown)


p2p/src/lib.rs
```diff
@@ -54,6 +75,69 @@ pub trait Sender: Clone + Debug + Send + Sync + 'static {
...
 }
 
+/// Interface for constructing a [`CheckedSender`] from a set of [`Recipients`],
+/// filtering out any that are currently rate-limited.
+pub trait LimitedSender: Clone + Send + Sync + 'static {
+    /// Public key type used to identify recipients.
+    type PublicKey: PublicKey;
+
+    /// The type of [`GClock`] used for rate-limiting.
+    type Clock: GClock + Send + 'static;
```
clabby: Note to reviewer - kept the `GClock` associated type to remove the need to propagate `GClock<Instant = SystemTime>` throughout the codebase. All of our backing implementations are `SystemTime` anyways - we should be able to just add that bound at the callsites where we need it (i.e. the resolver.)

p2p/src/simulated/network.rs
```diff
@@ -762,15 +814,124 @@ impl<P: PublicKey> crate::Sender for Sender<P> {
...
+                    }
+                }
+            }
+        };
+
+        Ok(CheckedSender {
+            recipients,
+            sender: self,
+        })
+    }
```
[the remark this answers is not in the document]
clabby: Ugh, good call. It may be best to just go back to how we originally had simulated with managing the RL on the sender, now that we've already got to propagate the clock in the type bounds.

p2p/src/utils/limited.rs
```diff
@@ -0,0 +1,587 @@
...
+
+        let rate_limit = self.rate_limit.lock().await;
+
+        // Update known peers from subscription if available (non-blocking)
+        if let Some(ref mut subscription) = self.peer_subscription {
+            if let Some(peers) = subscription.next().now_or_never().flatten() {
+                self.known_peers = peers;
+
+                rate_limit.retain_recent();
+                rate_limit.shrink_to_fit();
```
patrick-ogrady: Just call `retain_recent()` (we should keep the capacity as-is)

p2p/src/lib.rs
```diff
@@ -54,6 +75,69 @@ pub trait Sender: Clone + Debug + Send + Sync + 'static {
...
+    /// # Returns
+    ///
+    /// A vector of recipients that the message was sent to, or an error if the
+    /// message could not be sent (e.g., too large).
+    ///
+    /// Note: a successful send does not guarantee that the recipient will
+    /// receive the message.
+    fn send(
+        self,
+        message: impl Buf + Send,
```
patrick-ogrady: Let's keep as `Bytes` for consistency...we can change this to `impl Buf` in one fell swoop later.

p2p/src/simulated/network.rs
```diff
@@ -571,15 +587,19 @@ impl<E: RNetwork + Spawner + Rng + Clock + GClock + Metrics, P: PublicKey> Netwo
...
             if let Some(limiter) = self.rate_limiters.get(&(origin.clone(), channel)) {
-                if limiter.check_key(&recipient).is_err() {
-                    trace!(
-                        ?origin,
-                        ?recipient,
-                        reason = "rate limited",
-                        "dropping message"
-                    );
-                    continue;
+                if let Some(guard) = limiter.try_lock() {
```
patrick-ogrady: The lock will also only be held over a synchronous piece of code in `sender`? I think we can just acquire.

p2p/src/simulated/network.rs
```diff
@@ -571,15 +587,19 @@ impl<E: RNetwork + Spawner + Rng + Clock + GClock + Metrics, P: PublicKey> Netwo
             };
 
             // Check rate limit for this (sender, channel) pair
+            // Note: We use try_lock here since we're in a synchronous context.
+            // If the lock is held (rare), we skip the check and let the message through.
             if let Some(limiter) = self.rate_limiters.get(&(origin.clone(), channel)) {
```
patrick-ogrady: Now that this happens in checked_sender, do we need to do it again here? Am I missing something?

p2p/src/authenticated/discovery/channels.rs
```diff
@@ -60,104 +98,16 @@ where
...
-    }
-}
-
-impl<P: PublicKey, C: GClock> Debug for Sender<P, C> {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        f.debug_struct("Sender")
-            .field("channel", &self.channel)
-            .field("max_size", &self.max_size)
-            .finish()
     }
```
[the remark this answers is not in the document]
clabby: This is generally fine IMO, though we could move the check.

p2p/src/utils/limited.rs
```diff
@@ -0,0 +1,584 @@
...
+                    Recipients::Some(allowed)
+                }
+            }
+        };
+
+        Ok(CheckedSender {
+            recipients,
+            sender: &mut self.sender,
+        })
+    }
```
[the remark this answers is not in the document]
clabby: https://github.com/commonwarexyz/monorepo/pull/2533#discussion_r2625356046

p2p/src/simulated/network.rs
clabby: This code was removed when moving the RL back to the sender, but would agree.

p2p/src/lib.rs
```diff
@@ -38,14 +39,34 @@ pub enum Recipients<P: PublicKey> {
...
-    /// Send a message to a set of recipients.
+    /// Sends a message to a set of recipients.
+    ///
+    /// # Offline Recipients
+    ///
+    /// If a recipient is offline at the time a message is sent, the message
+    /// will be dropped. It is up to the application to handle retries (if
+    /// necessary).
+    ///
+    /// # Rate Limiting
```
patrick-ogrady, line 56: As discussed offline, I can't imagine a p2p implementation (even if not fully connected) that doesn't have some notion of rate limiting. Am ok keeping this a top-level comment.

p2p/src/lib.rs
patrick-ogrady: I'm very tempted to just return SystemTime still (and swallow the complexity)...we should have context everywhere already?

patrick-ogrady (commented):
Everything but simulated/network and utils/limited

p2p/src/utils/limited.rs
```diff
@@ -0,0 +1,584 @@
...
+use commonware_runtime::{Clock, KeyedRateLimiter, Quota};
+use commonware_utils::channels::ring;
+use futures::{lock::Mutex, Future, FutureExt, StreamExt};
+use std::{cmp, fmt, sync::Arc, time::SystemTime};
+
+/// Provides peer subscriptions for resolving [`Recipients::All`].
+///
+/// Implementations must be clonable so that each clone of [`LimitedSender`]
+/// can establish its own peer subscription.
+pub trait Peers: Clone + Send + Sync + 'static {
```
patrick-ogrady: To avoid confusion with peer set subscription, I wonder if we should call `Online`?

patrick-ogrady: Or `Connected`?

p2p/src/utils/limited.rs
```diff
@@ -0,0 +1,584 @@
...
+impl<E, S, P> fmt::Debug for LimitedSender<E, S, P>
+where
+    E: Clock,
+    S: Sender,
+    P: Peers<PublicKey = S::PublicKey>,
+{
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("LimitedSender")
+            .field("known_peers", &self.known_peers.len())
+            .finish_non_exhaustive()
```
patrick-ogrady, line 67: :fancy:

p2p/src/utils/limited.rs
```diff
@@ -0,0 +1,584 @@
...
+    /// Returns a [`CheckedSender`] with only the recipients that are not
+    /// currently rate-limited. If _all_ recipients are rate-limited, returns
+    /// the earliest instant at which all recipients will be available.
+    pub async fn check(
+        &mut self,
+        recipients: Recipients<S::PublicKey>,
+    ) -> Result<CheckedSender<'_, S>, SystemTime> {
+        // Lazily establish peer subscription on first use
+        if self.peer_subscription.is_none() {
+            self.peer_subscription = Some(self.peers.subscribe().await);
```
patrick-ogrady, line 102: This assumes we hear about known peers right away (if there are no peer set changes, we may never know). Wonder if that makes sense to document or just return with subscribe?

Maybe compare to what we do with manager subscribe?

p2p/src/utils/limited.rs
```diff
@@ -0,0 +1,584 @@
...
+/// time among those that don't.
+pub(crate) fn filter_rate_limited<'a, K, C>(
+    peers: impl Iterator<Item = &'a K>,
+    rate_limit: &KeyedRateLimiter<K, C>,
+) -> (Vec<K>, Option<SystemTime>)
+where
+    K: PublicKey,
+    C: Clock,
+{
+    peers.fold(
```
patrick-ogrady, line 162: nice

p2p/src/simulated/network.rs
```diff
@@ -311,16 +321,18 @@ impl<E: RNetwork + Spawner + Rng + Clock + GClock + Metrics, P: PublicKey> Netwo
                 result,
             } => {
                 // If peer does not exist, then create it.
-                self.ensure_peer_exists(&public_key).await;
+                let (_, is_new) = self.ensure_peer_exists(&public_key).await;
 
-                // Create rate limiter for this (sender, channel) pair
+                // When not using peer sets, broadcast updated peer list to subscribers
+                if is_new && self.peer_sets.is_empty() {
```
patrick-ogrady, line 327: Tricky but I think this is right.

p2p/src/simulated/network.rs
```diff
@@ -364,14 +379,33 @@ impl<E: RNetwork + Spawner + Rng + Clock + GClock + Metrics, P: PublicKey> Netwo
                 }
                 self.subscribers.push(sender);
             }
+            ingress::Message::SubscribePeers { response } => {
```
patrick-ogrady: Maybe we call this `Listen` or `Online` or something...I don't think this is different enough from `Subscribe`?

p2p/src/simulated/network.rs
```diff
@@ -657,23 +716,126 @@ impl<E: RNetwork + Spawner + Rng + Clock + GClock + Metrics, P: PublicKey> Netwo
...
 }
 
-/// Implementation of a [crate::Sender] for the simulated network.
-#[derive(Clone, Debug)]
-pub struct Sender<P: PublicKey> {
+/// Provides peer subscriptions for the simulated network.
+///
+/// Implements [`crate::utils::limited::Peers`] to provide peer list updates
+/// to [`crate::utils::limited::LimitedSender`].
+pub struct PeerSource<P: PublicKey, E: Clock> {
```
patrick-ogrady: I don't think the naming is good here but the function looks correct

p2p/src/utils/limited.rs
clabby: I like `Connected`

p2p/src/utils/limited.rs
clabby: It does assume this (and lookup + discovery + simulated do send immediately, so when you get the channel back it already has 1 update in the buffer.)

Added some documentation for this.

p2p/src/simulated/network.rs
clabby: Updated to `ConnectedPeerProvider`.

p2p/src/simulated/network.rs
clabby: Updated to `SubscribeConnected`.

PR #2534, opened 2025-12-16
[p2p] Add Support for Asymmetric Ingress/Egress + DNS Ingress

patrick-ogrady:
Closes: #1742 
Closes: #2033 
Closes: #1578
Closes: https://github.com/commonwarexyz/monorepo/issues/2548

## TODO
- [x] unify address filtering (check private/valid all over the place) -> should just check in read_cfg?
  - [x] at the least, don't need to check hostname inline? I suppose that means we also need to reject if private ips?
- [x] ensure we track peers with invalid addresses, we just shouldn't dial them (if invalid ingress) or accept connections from them (if invalid egress)
- [x] Enforce egress == connection source: https://github.com/commonwarexyz/monorepo/pull/2547

### DIFF
--- a/p2p/src/authenticated/lookup/actors/tracker/directory.rs
+++ b/p2p/src/authenticated/lookup/actors/tracker/directory.rs
@@ -1,7 +1,7 @@
 use super::{metrics::Metrics, record::Record, Metadata, Reservation};
 use crate::{
     authenticated::lookup::{actors::tracker::ingress::Releaser, metrics},
-    types::Address as PeerAddress,
+    types::Address,
     Ingress,
 };
 use commonware_cryptography::PublicKey;
@@ -116,7 +116,7 @@ impl<E: Spawner + Rng + Clock + RuntimeMetrics, C: PublicKey> Directory<E, C> {
     }
 
     /// Stores a new peer set.
-    pub fn add_set(&mut self, index: u64, peers: Map<C, PeerAddress>) -> Option<Vec<C>> {
+    pub fn add_set(&mut self, index: u64, peers: Map<C, Address>) -> Option<Vec<C>> {
         // Check if peer set already exists
         if self.sets.contains_key(&index) {
             warn!(index, "peer set already exists");
@@ -317,7 +317,7 @@ impl<E: Spawner + Rng + Clock + RuntimeMetrics, C: PublicKey> Directory<E, C> {
 mod tests {
     use crate::{
         authenticated::{lookup::actors::tracker::directory::Directory, mailbox::UnboundedMailbox},
-        types::Address as PeerAddress,
+        types::Address,
         Ingress,
     };
     use commonware_cryptography::{ed25519, Signer};
@@ -325,8 +325,8 @@ mod tests {
     use commonware_utils::{hostname, NZU32};
     use std::net::{IpAddr, Ipv4Addr, SocketAddr};
 
-    fn addr(socket: SocketAddr) -> PeerAddress {
-        PeerAddress::Symmetric(socket)
+    fn addr(socket: SocketAddr) -> Address {
+        Address::Symmetric(socket)
     }
 
     #[test]
@@ -539,7 +539,7 @@ mod tests {
         let pk_1 = ed25519::PrivateKey::from_seed(1).public_key();
         let ingress_socket = SocketAddr::new(IpAddr::V4(Ipv4Addr::new(10, 0, 0, 1)), 8080);
         let egress_socket = SocketAddr::new(IpAddr::V4(Ipv4Addr::new(192, 168, 1, 1)), 9090);
-        let asymmetric_addr = PeerAddress::Asymmetric {
+        let asymmetric_addr = Address::Asymmetric {
             ingress: Ingress::Socket(ingress_socket),
             egress: egress_socket,
         };
@@ -547,7 +547,7 @@ mod tests {
         // Create another peer with DNS-based ingress
         let pk_2 = ed25519::PrivateKey::from_seed(2).public_key();
         let egress_socket_2 = SocketAddr::new(IpAddr::V4(Ipv4Addr::new(192, 168, 1, 2)), 9090);
-        let dns_addr = PeerAddress::Asymmetric {
+        let dns_addr = Address::Asymmetric {
             ingress: Ingress::Dns {
                 host: hostname!("node.example.com"),
                 port: 8080,
@@ -636,11 +636,11 @@ mod tests {
         // Create peers with different address types
         let pk_socket = ed25519::PrivateKey::from_seed(1).public_key();
         let socket_addr = SocketAddr::new(IpAddr::V4(Ipv4Addr::new(10, 0, 0, 1)), 8080);
-        let socket_peer_addr = PeerAddress::Symmetric(socket_addr);
+        let socket_peer_addr = Address::Symmetric(socket_addr);
 
         let pk_dns = ed25519::PrivateKey::from_seed(2).public_key();
         let egress_socket = SocketAddr::new(IpAddr::V4(Ipv4Addr::new(192, 168, 1, 1)), 9090);
-        let dns_peer_addr = PeerAddress::Asymmetric {
+        let dns_peer_addr = Address::Asymmetric {
             ingress: Ingress::Dns {
                 host: hostname!("node.example.com"),
                 port: 8080,
@@ -700,11 +700,11 @@ mod tests {
         // Create peer with public egress IP
         let pk_public = ed25519::PrivateKey::from_seed(1).public_key();
         let public_addr =
-            PeerAddress::Symmetric(SocketAddr::new(IpAddr::V4(Ipv4Addr::new(8, 8, 8, 8)), 8080));
+            Address::Symmetric(SocketAddr::new(IpAddr::V4(Ipv4Addr::new(8, 8, 8, 8)), 8080));
 
         // Create peer with private egress IP
         let pk_private = ed25519::PrivateKey::from_seed(2).public_key();
-        let private_addr = PeerAddress::Symmetric(SocketAddr::new(
+        let private_addr = Address::Symmetric(SocketAddr::new(
             IpAddr::V4(Ipv4Addr::new(10, 0, 0, 1)),
             8080,
         ));
--- a/p2p/src/authenticated/lookup/actors/tracker/ingress.rs
+++ b/p2p/src/authenticated/lookup/actors/tracker/ingress.rs
@@ -5,7 +5,7 @@ use crate::{
         mailbox::UnboundedMailbox,
         Mailbox,
     },
-    types::Address as PeerAddress,
+    types::Address,
     Ingress,
 };
 use commonware_cryptography::PublicKey;
@@ -18,10 +18,7 @@ use std::net::IpAddr;
 pub enum Message<C: PublicKey> {
     // ---------- Used by oracle ----------
     /// Register a peer set at a given index.
-    Register {
-        index: u64,
-        peers: Map<C, PeerAddress>,
-    },
+    Register { index: u64, peers: Map<C, Address> },
 
     // ---------- Used by peer set provider ----------
     /// Fetch the peer set at a given index.
@@ -190,7 +187,7 @@ impl<C: PublicKey> Oracle<C> {
 
 impl<C: PublicKey> crate::Manager for Oracle<C> {
     type PublicKey = C;
-    type Peers = Map<C, PeerAddress>;
+    type Peers = Map<C, Address>;
 
     /// Register a set of authorized peers at a given index.
     ///
--- a/p2p/src/authenticated/lookup/actors/tracker/record.rs
+++ b/p2p/src/authenticated/lookup/actors/tracker/record.rs
@@ -1,4 +1,4 @@
-use crate::types::{Address as PeerAddress, Ingress};
+use crate::types::{self, Ingress};
 use std::net::IpAddr;
 
 /// Represents information known about a peer's address.
@@ -8,7 +8,7 @@ pub enum Address {
     Myself,
 
     /// Address is provided when peer is registered.
-    Known(PeerAddress),
+    Known(types::Address),
 
     /// Peer is blocked.
     /// We don't care to track its information.
@@ -51,7 +51,7 @@ impl Record {
     // ---------- Constructors ----------
 
     /// Create a new record with a known address.
-    pub const fn known(addr: PeerAddress) -> Self {
+    pub const fn known(addr: types::Address) -> Self {
         Self {
             address: Address::Known(addr),
             status: Status::Inert,
@@ -73,7 +73,7 @@ impl Record {
     // ---------- Setters ----------
 
     /// Update the record with a new address.
-    pub fn update(&mut self, addr: PeerAddress) {
+    pub fn update(&mut self, addr: types::Address) {
         if matches!(self.address, Address::Myself | Address::Blocked) {
             return;
         }
@@ -235,8 +235,8 @@ mod tests {
         SocketAddr::from(([54, 12, 1, 9], 8080))
     }
 
-    fn test_address() -> PeerAddress {
-        PeerAddress::Symmetric(test_socket())
+    fn test_address() -> types::Address {
+        types::Address::Symmetric(test_socket())
     }
 
     #[test]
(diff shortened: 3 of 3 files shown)


p2p/fuzz/src/lib.rs
```diff
@@ -323,11 +323,11 @@ impl NetworkScheme for Lookup {
 
         // For lookup, we must provide both public keys AND addresses
         // (unlike discovery which finds addresses through the protocol)
-        let peer_list: Vec<_> = peer
+        let peer_list: Vec<(_, commonware_p2p::Address)> = peer
```
patrick-ogrady: just use `Address`

p2p/src/authenticated/discovery/types.rs
```diff
@@ -213,7 +215,7 @@ impl<C: PublicKey> Info<C> {
     pub fn verify(&self, namespace: &[u8]) -> bool {
         self.public_key.verify(
             namespace,
-            &(self.socket, self.timestamp).encode(),
+            &(self.ingress.clone(), self.timestamp).encode(),
```
patrick-ogrady, line 211: nit: make `ingress` copy

p2p/src/authenticated/discovery/types.rs
```diff
@@ -359,10 +363,12 @@ impl<C: PublicKey> InfoVerifier<C> {
         // for selecting a random subset of peers when there are too many) and allow
         // for duplicates (no need to create an additional set to check this)
         for info in infos {
-            // Check if IP is allowed
+            // Check if IP is allowed (only applicable for Socket addresses, not DNS)
             #[allow(unstable_name_collisions)]
-            if !self.allow_private_ips && !info.socket.ip().is_global() {
-                return Err(Error::PrivateIPsNotAllowed(info.socket.ip()));
+            if let Some(ip) = info.ingress.ip() {
```
patrick-ogrady: We can still check IP after DNS is resolved?

p2p/src/authenticated/lookup/actors/dialer.rs
```diff
@@ -100,6 +106,25 @@ impl<E: Spawner + Clock + GClock + Network + Rng + CryptoRng + Metrics, C: Signe
             let config = self.stream_cfg.clone();
             let mut supervisor = supervisor.clone();
             move |context| async move {
+                // Resolve address if needed
+                let address: SocketAddr = match ingress {
```
patrick-ogrady: Make this a helper!

p2p/src/types.rs
```diff
@@ -0,0 +1,397 @@
...
+    }
+}
+
+impl From<SocketAddr> for Ingress {
+    fn from(addr: SocketAddr) -> Self {
+        Self::Socket(addr)
+    }
+}
+
+/// Full address specification for lookup network (needs IP filtering).
```
patrick-ogrady: nit: generalize to any network (not just for lookup)

p2p/src/types.rs
```diff
@@ -0,0 +1,397 @@
...
+        assert_eq!(address, Address::Symmetric(addr));
+    }
+
+    #[cfg(feature = "arbitrary")]
+    mod conformance {
+        use super::*;
+        use commonware_codec::conformance::CodecConformance;
+
+        commonware_conformance::conformance_tests! {
+            CodecConformance<Ingress> => 1024,
```
patrick-ogrady: remove specific case count

p2p/src/types.rs
```diff
@@ -0,0 +1,397 @@
...
+    /// Returns the port number for this ingress address.
+    pub const fn port(&self) -> u16 {
+        match self {
+            Self::Socket(addr) => addr.port(),
+            Self::Dns { port, .. } => *port,
+        }
+    }
+
+    /// Returns the IP address if this is a Socket variant.
+    pub const fn ip(&self) -> Option<IpAddr> {
```
patrick-ogrady, line 39: is this necessary?

p2p/src/authenticated/discovery/types.rs
patrick-ogrady: We could just do this check once in dialer?

We should also assert egress satisifies?

codec/src/lib.rs
```diff
@@ -212,6 +212,9 @@ pub use codec::*;
 pub use config::RangeCfg;
 pub use error::Error;
 pub use extensions::*;
+// Re-export network types
+#[cfg(feature = "std")]
+pub use types::net::{Hostname, MAX_HOSTNAME_LABEL_LEN, MAX_HOSTNAME_LEN};
```
patrick-ogrady: Do we need to export this? I don't think so...

p2p/src/authenticated/discovery/actors/tracker/actor.rs
```diff
@@ -340,7 +342,8 @@ mod tests {
         make_sig_invalid: bool,
     ) -> Info<PublicKey> {
         let peer_info_pk = target_pk_override.unwrap_or_else(|| signer.public_key());
-        let mut signature = signer.sign(ip_namespace, &(socket, timestamp).encode());
+        let ingress = crate::Ingress::Socket(socket);
```
patrick-ogrady: nit: `.into()`

p2p/src/authenticated/discovery/actors/tracker/actor.rs
```diff
@@ -1051,7 +1060,7 @@ mod tests {
                 Some(peer::Message::Peers(infos)) => {
                     assert_eq!(infos.len(), 1, "Expected 1 Info (for peer1)");
                     assert_eq!(infos[0].public_key, peer1_pk);
-                    assert_eq!(infos[0].socket, peer1_addr);
+                    assert_eq!(infos[0].ingress, crate::Ingress::Socket(peer1_addr));
```
patrick-ogrady: nit: `.into()`

p2p/src/authenticated/discovery/actors/tracker/mod.rs
```diff
@@ -23,9 +23,10 @@ pub use reservation::Reservation;
 pub struct Config<C: Signer> {
     pub crypto: C,
     pub namespace: Vec<u8>,
-    pub address: SocketAddr,
+    pub address: Ingress,
```
patrick-ogrady, line 26: Confirm, this isn't necessarily what we listen on (may want to advertise different address than what we listen on).

p2p/src/authenticated/discovery/actors/tracker/record.rs
```diff
@@ -216,15 +216,31 @@ impl<C: PublicKey> Record<C> {
...
+        }
+        let ingress = match &self.address {
+            Address::Bootstrapper(ingress) => ingress,
+            Address::Discovered(info, _) => &info.ingress,
+            _ => return false,
+        };
+        if !ingress.is_valid(allow_dns) {
+            return false;
+        }
+        // For Socket addresses, check if private IPs are allowed
```
patrick-ogrady: We can just keep these checks in dialer?

p2p/src/authenticated/discovery/types.rs
```diff
@@ -359,12 +343,6 @@ impl<C: PublicKey> InfoVerifier<C> {
         // for selecting a random subset of peers when there are too many) and allow
         // for duplicates (no need to create an additional set to check this)
         for info in infos {
-            // Check if IP is allowed
```
patrick-ogrady, line 362: This would've actually led to us continually dropping signed IPs for some peer if two peers had misaligned configs...we should instead accept peer records and store (to ensure we don't ask for again).

p2p/src/authenticated/lookup/actors/tracker/directory.rs
```diff
@@ -120,22 +131,30 @@ impl<E: Spawner + Rng + Clock + RuntimeMetrics, C: PublicKey> Directory<E, C> {
             }
         }
 
-        // Create and store new peer set
-        for (peer, addr) in &peers {
+        // Create and store new peer set (all peers are tracked regardless of address validity)
+        let mut set_peers = Vec::new();
```
patrick-ogrady: We don't need this vec?

p2p/src/authenticated/lookup/actors/tracker/directory.rs
```diff
@@ -228,14 +248,14 @@ impl<E: Spawner + Rng + Clock + RuntimeMetrics, C: PublicKey> Directory<E, C> {
...
     }
 
-    /// Return all registered IP addresses.
+    /// Return all registered IP addresses (egress IPs for filtering).
     pub fn registered(&self) -> HashSet<IpAddr> {
         // Using `.allowed()` here excludes any peers that are still connected but no longer
         // part of a peer set (and will be dropped shortly).
         self.peers
             .values()
             .filter(|r| r.allowed(self.allow_private_ips))
```
patrick-ogrady: Ensure this checks egress is not private IP

p2p/src/authenticated/lookup/actors/tracker/mod.rs
```diff
@@ -24,5 +24,7 @@ pub struct Config<C: Signer> {
     pub tracked_peer_sets: usize,
     pub allowed_connection_rate_per_peer: Quota,
     pub allow_private_ips: bool,
+    /// Whether DNS-based ingress addresses are allowed.
```
patrick-ogrady: nit: remove

patrick-ogrady (commented):
pretty close now

utils/src/hostname.rs
```diff
@@ -0,0 +1,414 @@
...
+
+    fn try_from(value: &str) -> Result<Self, Self::Error> {
+        Self::new(value)
+    }
+}
+
+impl CodecWrite for Hostname {
+    #[inline]
+    fn write(&self, buf: &mut impl BufMut) {
+        self.0.as_bytes().to_vec().write(buf);
```
clabby:
```suggestion
        self.0.as_bytes().write(buf);
```

utils/src/hostname.rs
```diff
@@ -0,0 +1,414 @@
...
+    }
+}
+
+impl EncodeSize for Hostname {
+    #[inline]
+    fn encode_size(&self) -> usize {
+        self.0.as_bytes().to_vec().encode_size()
```
clabby:
```suggestion
        self.0.as_bytes().encode_size()
```

p2p/src/types.rs
```diff
@@ -0,0 +1,473 @@
...
+        match self {
+            Self::Socket(addr) => allow_private_ips || IpAddrExt::is_global(&addr.ip()),
+            Self::Dns { .. } => allow_dns,
+        }
+    }
+
+    /// Resolve this ingress address to a socket address.
+    ///
+    /// For `Socket` variants, returns the address directly.
+    /// For `Dns` variants, performs DNS resolution and returns the first resolved IP.
```
clabby: Should we return an iterator over the ips here? Folks may want to try each until one connects. Some of the more complicate infra setups of our customers might have round robin DNS setups.

examples/reshare/src/validator.rs
```diff
@@ -70,7 +70,12 @@ pub async fn run<S>(
         &p2p_namespace,
         SocketAddr::new(IpAddr::V4(Ipv4Addr::LOCALHOST), config.port),
         SocketAddr::new(IpAddr::V4(Ipv4Addr::LOCALHOST), config.port),
-        config.bootstrappers.clone().into_iter().collect::<Vec<_>>(),
+        config
+            .bootstrappers
+            .clone()
```
BrendanChou: nit: this creates an intermediary map that's cloned. Instead of calling `bootstrappers.clone()`, you could do `.map(|(k, v)| (k.clone(), v.clone().into()))` to avoid the intermediary map. Depending on whether `into()` works on references, you might not have to `v.clone()` at all. 

Although since the performance doesn't matter for this one-time function, the current code may be more readable

p2p/fuzz/src/lib.rs
```diff
@@ -326,7 +326,7 @@ impl NetworkScheme for Lookup {
             .topo
             .peers
             .iter()
-            .map(|p| (p.public_key.clone(), p.address))
+            .map(|p| (p.public_key.clone(), Address::from(p.address)))
```
BrendanChou: consistency: I think you use `.into()` most other places rather than `Address::from`

p2p/src/types.rs
```diff
@@ -0,0 +1,473 @@
...
+            INGRESS_SOCKET_PREFIX => {
+                let addr = SocketAddr::read(buf)?;
+                Ok(Self::Socket(addr))
+            }
+            INGRESS_DNS_PREFIX => {
+                let host = Hostname::read(buf)?;
+                let port = u16::read(buf)?;
+                Ok(Self::Dns { host, port })
+            }
+            _ => Err(CodecError::Invalid("Ingress", "Invalid prefix")),
```
BrendanChou: I think we typically use `CodecError::InvalidEnum`

...arguably it should come with a `&'static str` to give more context but it only takes in the u8 value for now

p2p/src/types.rs
```diff
@@ -0,0 +1,473 @@
...
+    }
+
+    /// Returns whether this ingress address is allowed given the configuration.
+    ///
+    /// - `Socket` addresses must have a global IP (or `allow_private_ips` must be true).
+    /// - `Dns` addresses are allowed only if `allow_dns` is `true`.
+    ///
+    /// Note: For `Dns` addresses, private IP checks are performed after resolution in
+    /// [`resolve_filtered`](Self::resolve_filtered).
+    pub fn is_valid(&self, allow_private_ips: bool, allow_dns: bool) -> bool {
```
BrendanChou, line 53: nit: I kind of think `address.is_valid(bool1, bool2)` is kind of opaque what the booleans mean and it's easy to mess up the order. I wonder if there's a more graceful way to do this

p2p/src/types.rs
```diff
@@ -0,0 +1,473 @@
...
+
+const INGRESS_SOCKET_PREFIX: u8 = 0;
+const INGRESS_DNS_PREFIX: u8 = 1;
+
+const ADDRESS_SYMMETRIC_PREFIX: u8 = 0;
+const ADDRESS_ASYMMETRIC_PREFIX: u8 = 1;
+
+/// What we dial to connect to a peer.
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
+pub enum Ingress {
```
BrendanChou, line 17: The naming confused me because it seems "backwards"? Like isn't dialing considered a form of egress?

utils/src/hostname.rs
```diff
@@ -0,0 +1,414 @@
...
+    }
+}
+
+impl CodecRead for Hostname {
+    type Cfg = ();
+
+    #[inline]
+    fn read_cfg(buf: &mut impl Buf, _: &()) -> Result<Self, CodecError> {
+        let bytes = Vec::<u8>::read_cfg(buf, &(RangeCfg::new(..=MAX_HOSTNAME_LEN), ()))?;
```
BrendanChou, line 161: If you follow Clabby's advice and write as a Bytes then you should probably also read as a Bytes and not a Vec although I think it's the same encoding

p2p/src/types.rs
patrick-ogrady: I think that's fair.

p2p/src/types.rs
patrick-ogrady: yeah, I think the answer would be some sort of config struct or something but seemed like overkill.

utils/src/hostname.rs
patrick-ogrady: `.as_bytes()` converts to `&[u8]`?

p2p/src/types.rs
patrick-ogrady: I think keeping this straight (either way), is annoying/tough. I prefer to track both ingress/egress as the value actually reported by the peer.

That means they dial us from egress and we dial them at ingress.

p2p/src/authenticated/lookup/actors/tracker/record.rs
```diff
@@ -8,7 +8,7 @@ pub enum Address {
     Myself,
 
     /// Address is provided when peer is registered.
-    Known(SocketAddr),
+    Known(PeerAddress),
```
patrick-ogrady: nit: Address

p2p/src/authenticated/lookup/actors/tracker/ingress.rs
```diff
@@ -182,7 +190,7 @@ impl<C: PublicKey> Oracle<C> {
 
 impl<C: PublicKey> crate::Manager for Oracle<C> {
     type PublicKey = C;
-    type Peers = Map<C, SocketAddr>;
+    type Peers = Map<C, PeerAddress>;
```
patrick-ogrady: nit: address

PR #2535, opened 2025-12-17
[Storage] Consolidate `qmdb::any` `IndexedLog` implementations

danlaine:
* Export databases as `qmdb::any::ordered::fixed::Db`, `qmdb::any::unordered::fixed::Db`,`qmdb::any::ordered::variable::Db`,`qmdb::any::unordered::variable::Db`.
* Export operations as `qmdb::any::unordered::fixed::Operation`, `qmdb::any::unordered::variable::Operation`,`qmdb::any::ordered::fixed::Operation`,`qmdb::any::ordered::variable::Operation`
* Similar for updates
* Add `qmdb::any::db::Db` which contains factored out `IndexedLog` implementations from ordered and unordered implementations.
* Change `qmdb::any`'s `oldest_retained_loc()` to return `Location` instead of `Option<Location>`

### DIFF
--- a/storage/src/qmdb/any/db.rs
+++ b/storage/src/qmdb/any/db.rs
@@ -32,7 +32,7 @@ use tracing::debug;
 pub(crate) type AuthenticatedLog<E, C, H, S = Clean<DigestOf<H>>> =
     authenticated::Journal<E, C, H, S>;
 
-/// An "Any" QMDB implementation generic over ordered/ordered keys and variable/fixed values.
+/// An "Any" QMDB implementation generic over ordered/unordered keys and variable/fixed values.
 /// Consider using one of the following specialized variants instead, which may be more ergonomic:
 /// - [crate::qmdb::any::ordered::fixed::Db]
 /// - [crate::qmdb::any::ordered::variable::Db]


storage/src/qmdb/any/db.rs
```diff
@@ -0,0 +1,502 @@
...
+        log: AuthenticatedLog<E, C, H>,
+        known_inactivity_floor: Option<Location>,
+        mut callback: F,
+    ) -> Result<Self, Error>
+    where
+        F: FnMut(bool, Option<Location>),
+    {
+        // If the last-known inactivity floor is behind the current floor, then invoke the callback
+        // appropriately to report the inactive bits.
+        let inactivity_floor_loc = Self::recover_inactivity_floor(&log).await?;
```
roberto-bayardo: Do you want to just inline this function here?  I think it's the only remaining callsite.  (I notice it also computes the last_commit_loc, so you can avoid the recomputation on L186).

storage/src/qmdb/any/db.rs
```diff
@@ -0,0 +1,502 @@
...
+        if let Some(mut known_inactivity_floor) = known_inactivity_floor {
+            while known_inactivity_floor < inactivity_floor_loc {
+                callback(false, None);
+                known_inactivity_floor += 1;
+            }
```
roberto-bayardo:
```suggestion
            (known_inactivity_floor..inactivity_floor_loc).for_each(|_| callback(false, None));
```
might be a more fun way to do this 😃

storage/src/qmdb/any/db.rs
```diff
@@ -0,0 +1,502 @@
...
+pub(crate) type AuthenticatedLog<E, C, H, S = Clean<DigestOf<H>>> =
+    authenticated::Journal<E, C, H, S>;
+
+/// An "Any" QMDB implementation generic over ordered/ordered keys and variable/fixed values.
+/// Consider using one of the following specialized variants instead, which may be more ergonomic:
+/// - [crate::qmdb::any::ordered::Fixed]
+/// - [crate::qmdb::any::ordered::Variable]
+/// - [crate::qmdb::any::unordered::Fixed]
+/// - [crate::qmdb::any::unordered::Variable]
+pub struct Db<
```
roberto-bayardo, line 41: I'm not sure I'm fond of calling this structure _Db_ but could be convinced. I suppose since it isn't public it's not that important.

roberto-bayardo: So the module is pub(crate) but the struct is pub -- should they both be pub(crate)?  Surprised this doesn't trigger a visibility warning.

storage/src/qmdb/any/db.rs
```diff
@@ -0,0 +1,502 @@
...
+    }
+
+    /// Whether the snapshot currently has no active keys.
+    pub const fn is_empty(&self) -> bool {
+        self.active_keys == 0
+    }
+
+    /// Returns the location of the oldest operation that remains retrievable.
+    pub fn oldest_retained_loc(&self) -> Option<Location> {
+        self.log.oldest_retained_loc()
```
roberto-bayardo: The return value need not be an option (there should always be one retained commit).
(I know this is my code, just noticed it now!)

storage/src/qmdb/any/db.rs
danlaine: done

storage/src/qmdb/any/db.rs
danlaine: done

storage/src/qmdb/any/db.rs
danlaine: removed option

storage/src/qmdb/any/db.rs
danlaine: > I'm not sure I'm fond of calling this structure Db but could be convinced. I suppose since it isn't public it's not that important.

I thought `Db` might be easier to grok than `IndexedLog` or `Any`. In defense of that name, it is the type that holds, well, the database state. The various implementations of `qmdb::any` are all this under the hood.

> So the module is pub(crate) but the struct is pub -- should they both be pub(crate)? Surprised this doesn't trigger a visibility warning.

The upshot of having `pub(crate) mod db` and `pub struct Db` is that outside the storage crate, you use the database as `commonware_storage::qmdb::any::ordered::Fixed`, `commonware_storage::qmdb::any::unordered::Variable`, etc. `commonware_storage::qmdb::any::db` is not visible. `Db` needs to be `pub` (not `pub(crate)`) since the aliases `Fixed`, `Variable` are `pub`.

storage/src/qmdb/any/ordered/mod.rs
```diff
@@ -256,7 +176,7 @@ where
                 op.is_update(),
                 "location does not reference update operation. loc={loc}"
             );
-            if op.key().expect("update operation must have key") == key {
+            if *op.key().expect("update operation must have key") == *key {
```
roberto-bayardo: Are the derefs needed here?

danlaine: nope! Removed

storage/src/qmdb/any/db.rs
roberto-bayardo: As we discussed OOB, let's call all the top-level structs "Db" and allow the module names to indicate the specific variant.

storage/src/qmdb/benches/variable/mod.rs
```diff
@@ -5,7 +5,7 @@ use commonware_runtime::{buffer::PoolRef, create_pool, tokio::Context, ThreadPoo
 use commonware_storage::{
     qmdb::{
         any::{
-            ordered::variable::Any as OAny, unordered::variable::Any as UAny,
+            ordered::variable::{Db as OVariable, Db as UVariable},
```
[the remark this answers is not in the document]
danlaine: Good bot. Fixed.

PR #2537, opened 2025-12-17
[conformance + cryptography] Capture Handshake

### LINKED ISSUE #2497: [conformance + cryptography] Capture Handshake
While we ensure conformance of handshake types, it would be great to capture the entire flow of a handshake (to ensure the mechanism itself doesn't change)!

stevencartavia:
should close #2497

### DIFF
--- a/cryptography/src/handshake.rs
+++ b/cryptography/src/handshake.rs
@@ -497,6 +497,11 @@ mod test {
                         listener_key.public_key(),
                     ),
                 );
+
+                // Capture transcript of all messages exchanged
+                let mut transcript = Vec::new();
+                transcript.extend(syn.encode());
+
                 let (l_state, syn_ack) = listen_start(
                     &mut rng,
                     Context::new(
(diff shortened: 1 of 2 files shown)


cryptography/src/handshake.rs
```diff
@@ -476,11 +476,59 @@ mod test {
...
+                let (_l_send, mut l_recv) = listen_end(l_state, ack).expect("handshake failed");
+
+                // Test end-to-end encryption
+                let msg = seed.to_le_bytes();
+                let ciphertext = d_send.send(&msg).expect("send failed");
+                let decrypted = l_recv.recv(&ciphertext).expect("recv failed");
+                assert_eq!(msg, decrypted.as_slice());
+
+                // Return deterministic representation of the complete flow
+                ciphertext
```
patrick-ogrady: We want to return a view of the "channel state" or "transcript" (the messages sent back and forth to establish the handshake).

IMHO the right way to do this is to just add all messages sent back and forth to a hasher (syn -> syn_ack -> etc).

Just returning the ciphertext here won't enforce that is stable.

patrick-ogrady:
@clabby ended up taking this on in #2542

Issue #2513, opened 2025-12-15
[p2p] `Quota` from `governor` is not re-exported


fedemagnani:
Currently, downstream crates must depend on `governor` directly in order to use `Quota`. This can easily lead to version mismatches between the governor version used by `commonware_p2p` and the one pulled in by the downstream crate, resulting in errors such as `mismatched types
you can use cargo tree to explore your dependency tree`.

This problem would be fixed by simply re-exporting the `Quota` struct from the `governor` crate.

fedemagnani:
Closed by #2540

PR #2540, opened 2025-12-17
[runtime] Expand `Clock` Trait

patrick-ogrady:


runtime/src/lib.rs
```diff
@@ -277,7 +291,14 @@ pub type RateLimiter<K, C> = governor::RateLimiter<
 ///
 /// It is necessary to mock time to provide deterministic execution
 /// of arbitrary tasks.
-pub trait Clock: Clone + Send + Sync + 'static {
+pub trait Clock:
```
patrick-ogrady, line 294: Alternatively, we could create a `RateLimitedClock` or `GovernedClock` but I think keeping them closely related is more than OK for now (given we almost always use RL with the clock).

examples/estimator/src/main.rs
```diff
@@ -290,9 +289,7 @@ fn run_single_simulation(
 }
 
 /// Core simulation logic that runs the network simulation
-async fn run_simulation_logic<
-    C: Spawner + Clock + Clone + Metrics + RNetwork + RngCore + GClock,
->(
+async fn run_simulation_logic<C: Spawner + Clock + Clone + Metrics + RNetwork + RngCore>(
```
clabby: nit: drop `Clone`, implied by `Clock`.

runtime/src/lib.rs
```diff
@@ -259,13 +259,27 @@ pub trait Metrics: Clone + Send + Sync + 'static {
     fn encode(&self) -> String;
 }
 
+/// Re-export of [governor::Quota] for rate limiting configuration.
+pub use governor::Quota;
```
clabby, line 263: Let's re-export all of governor. This'd also address #2513 / #2514

clabby: @patrick-ogrady doesn't want to do this yet.

PR #2542, opened 2025-12-17
[cryptography] Handshake conformance

clabby:
## Overview

Adds a conformance test for the handshake mechanism, committing to a log of the initial exchange, raw messages, and ciphertext.

closes https://github.com/commonwarexyz/monorepo/issues/2497

### DIFF
--- a/cryptography/src/handshake/conformance.rs
+++ b/cryptography/src/handshake/conformance.rs
@@ -12,7 +12,7 @@ use commonware_math::algebra::Random;
 use rand::{Rng, SeedableRng};
 use rand_chacha::ChaCha8Rng;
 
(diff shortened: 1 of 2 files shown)


cryptography/src/handshake/conformance.rs
```diff
@@ -0,0 +1,91 @@
...
+    transcript::Transcript,
+    Signer,
+};
+use commonware_codec::Encode;
+use commonware_conformance::{conformance_tests, Conformance};
+use commonware_math::algebra::Random;
+use rand::{Rng, SeedableRng};
+use rand_chacha::ChaCha8Rng;
+
+const NAMESPACE: &[u8] = b"_COMMONWARE_HANDSHAKE_CONFORMANCE_TESTS_";
```
patrick-ogrady: nit: you can drop the trailing `_` (we assume suffixes provide `_XXX`)

PR #2546, opened 2025-12-17
[storage/qmdb/current] restructure current db in preparation for variable variants

roberto-bayardo:
- Move current-specific proof types into a `proof` module.
- Rename `Current` structs to `Db` (as was done with the Any variants)
- Created fixed submodules in ordered & unordered for the existing fixed current variants (like we do with the Any variants).

This work is in preparation for upcoming variable variants of current dbs that will appear in new [un]ordered/variable modules.

### DIFF
--- a/storage/src/qmdb/current/ordered/fixed.rs
+++ b/storage/src/qmdb/current/ordered/fixed.rs
@@ -1,5 +1,11 @@
-//! An _ordered_ variant of a [crate::qmdb::current] authenticated database that maintains the
-//! lexicographic-next active key of each active key, allowing for exclusion proofs.
+//! An _ordered_ variant of a [crate::qmdb::current] authenticated database optimized for fixed-size
+//! values
+//!
+//! This variant maintains the lexicographic-next active key for each active key, enabling exclusion
+//! proofs (proving a key is currently inactive). Use [super::super::unordered::fixed] if exclusion
+//! proofs are not needed.
+//!
+//! See [Db] for the main database type and [ExclusionProof] for proving key inactivity.
 
 use crate::{
     bitmap::{CleanBitMap, DirtyBitMap},
--- a/storage/src/qmdb/current/ordered/mod.rs
+++ b/storage/src/qmdb/current/ordered/mod.rs
@@ -1,5 +1,11 @@
-//! _Ordered_ variants of a [crate::qmdb::current] authenticated database. These variants maintain
-//! the lexicographic-next active key of each active key, allowing for exclusion proofs.
+//! _Ordered_ variants of a [crate::qmdb::current] authenticated database.
+//!
+//! These variants maintain the lexicographic-next active key for each active key, enabling
+//! exclusion proofs via [ExclusionProof]. This adds overhead compared to [super::unordered]
+//! variants.
+//!
+//! Variants:
+//! - [fixed]: Variant optimized for values of fixed size.
 
 use crate::qmdb::{
     any::{ordered::fixed::Update, FixedValue},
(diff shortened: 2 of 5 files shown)


storage/src/qmdb/current/proof.rs
```diff
@@ -0,0 +1,276 @@
+use crate::{
```
andresilva, line 7:
```suggestion
//! Proof types for [`crate::qmdb::current`] authenticated databases.
//!
//! This module provides:
//! - [`RangeProof`]: Proves a range of operations exist in the database.
//! - [`OperationProof`]: Proves a specific operation is active, combining an MMR proof with
//!   the bitmap chunk containing the operation's activity status.
use crate::{
```

storage/src/qmdb/current/ordered/fixed.rs
```diff
@@ -10,10 +10,15 @@ use crate::{
     },
```
andresilva, line 16: github UI doesn't allow me to target the correct line, but here's a suggestion for module docs:

```
//! An _ordered_, fixed-size value variant of a [`crate::qmdb::current`] authenticated database.
//!
//! This variant maintains the lexicographic-next active key for each active key, enabling
//! exclusion proofs (proving a key does *not* exist). Use [`super::super::unordered::fixed`]
//! if exclusion proofs are not needed.
//!
//! See [`Db`] for the main database type and [`ExclusionProof`](super::ExclusionProof) for
//! proving key non-existence.
```

storage/src/qmdb/current/unordered/fixed.rs
```diff
@@ -1,7 +1,3 @@
-//! A [crate::qmdb::current] authenticated database that does not maintain an ordering over active
-//! keys, and hence does not support exclusion proofs. Use the [super::ordered] variant if exclusion
-//! proofs are required.
-
 use crate::{
```
andresilva, line 9:
```suggestion
//! An _unordered_, fixed-size value variant of a [`crate::qmdb::current`] authenticated database.
//!
//! This variant does not maintain key ordering, so it cannot generate exclusion proofs.
//! Use [`super::super::ordered::fixed`] if exclusion proofs are required.
//!
//! See [`Db`] for the main database type.
use crate::{
```

storage/src/qmdb/current/ordered/mod.rs
```diff
@@ -0,0 +1,26 @@
+//! _Ordered_ variants of a [crate::qmdb::current] authenticated database. These variants maintain
+//! the lexicographic-next active key of each active key, allowing for exclusion proofs.
```
andresilva:
```suggestion
//! _Ordered_ variants of a [`crate::qmdb::current`] authenticated database.
//!
//! These variants maintain the lexicographic-next active key for each active key, enabling
//! exclusion proofs via [`ExclusionProof`]. This adds overhead compared to
//! [`super::unordered`] variants.
//!
//! Variants:
//! - [`fixed`]: Fixed-size value variant.
```

storage/src/qmdb/current/unordered/mod.rs
```diff
@@ -0,0 +1,5 @@
+//! [crate::qmdb::current] variants that do not maintain an ordering over active keys, and hence do
+//! not support exclusion proofs. Use the [super::ordered] variants if exclusion proofs are
+//! required.
```
andresilva:
```suggestion
//! _Unordered_ variants of a [`crate::qmdb::current`] authenticated database.
//!
//! These variants do not maintain key ordering, so they cannot generate exclusion proofs.
//! Use [`super::ordered`] variants if exclusion proofs are required.
//!
//! Variants:
//! - [`fixed`]: Fixed-size value variant.
```

storage/src/qmdb/current/ordered/mod.rs
```diff
@@ -0,0 +1,26 @@
...
+use crate::qmdb::{
+    any::{ordered::fixed::Update, FixedValue},
+    current::proof::OperationProof,
+};
+use commonware_cryptography::Digest;
+use commonware_utils::Array;
+
+pub mod fixed;
+
+/// Proof information for verifying a key is not currently active in the database.
```
andresilva:
```suggestion
/// Proof that a key does not exist in the database.
///
/// When the database has active keys, exclusion is proven by showing the key falls within
/// a span between two adjacent active keys. When the database is empty, exclusion is proven
/// via the most recent commit operation.
///
/// Verify using [`Db::verify_exclusion_proof`](fixed::Db::verify_exclusion_proof).
```

storage/src/qmdb/current/ordered/mod.rs
```diff
@@ -0,0 +1,26 @@
...
+#[derive(Clone, Eq, PartialEq, Debug)]
+pub enum ExclusionProof<K: Array, V: FixedValue, D: Digest, const N: usize> {
+    /// For the KeyValue variant, we're proving that a span over the keyspace exists in the
+    /// database, allowing one to prove any key falling within that span (but not at the beginning)
+    /// is excluded.
```
andresilva:
```suggestion
    /// Proves that a span over the keyspace exists in the database, allowing any key
    /// falling strictly within that span (but not at the beginning) to be proven excluded. 
    /// The span's start key cannot be excluded since it is active.
```

storage/src/qmdb/current/ordered/mod.rs
```diff
@@ -0,0 +1,26 @@
...
+    KeyValue(OperationProof<D, N>, Update<K, V>),
+
+    /// For the Commit variant, we're proving that there exists a Commit operation in the database
+    /// that establishes an inactivity floor equal to its own location. This implies there are no
+    /// active keys, and therefore any key can be proven excluded against it. The wrapped values
+    /// consist of the location of the commit operation and its digest.
```
andresilva:
```suggestion
    /// Proves that a commit operation exists with an inactivity floor equal to its own
    /// location, implying no keys are active. Any key can be proven excluded against it.
    /// Contains the location of the commit operation and its digest.
```

roberto-bayardo:
@andresilva Thanks! Accepted all your suggestions with a bit of tweaking. PTAL?

Issue #928, opened 2025-05-13
[p2p] Consider new `Payload` type `Greeting(PeerInfo<C>)`


BrendanChou:
On connection, peers will send their own `PeerInfo<C>` as a `Payload::Peers(...)` of length 1.

We can consider sending it as its own message so that we can:
- Make sure we receive it once and only once
- Validate the information slightly differently than we would a `Peers` message
- Don't consume the rate-limiter for the `Peers` messages (since we will shortly receive a second `Peers` message, if the rate-limiter is set too low, we could rate-limit the second, legitimate message)

PR #2553, opened 2025-12-18
[p2p/discovery] Implement `Greeting`

patrick-ogrady:
Closes: #928

### DIFF
--- a/p2p/src/authenticated/discovery/config.rs
+++ b/p2p/src/authenticated/discovery/config.rs
@@ -113,7 +113,7 @@ pub struct Config<C: Signer> {
     /// and should be sent more often than the read_timeout.
     ///
     /// This also determines the rate limit for incoming BitVec and Peers messages
-    /// (one per this frequency).
+    /// (one per half this frequency to account for jitter).
     pub gossip_bit_vec_frequency: Duration,
 
     /// Maximum number of peers we will send or consider valid when receiving in a single message.
--- a/p2p/src/authenticated/lookup/config.rs
+++ b/p2p/src/authenticated/lookup/config.rs
@@ -75,7 +75,8 @@ pub struct Config<C: Signer> {
 
     /// Frequency at which we send ping messages to peers.
     ///
-    /// This also determines the rate limit for incoming ping messages (one per this frequency).
+    /// This also determines the rate limit for incoming ping messages (one per half this
+    /// frequency to account for jitter).
     pub ping_frequency: Duration,
 
     /// Average frequency at which we make a single dial attempt across all peers.


p2p/src/authenticated/discovery/actors/peer/actor.rs
```diff
@@ -200,14 +203,41 @@ impl<E: Spawner + Clock + Rng + CryptoRng + Metrics, C: PublicKey> Actor<E, C> {
...
+                    // Ensure we start with a greeting message and then never receive another
+                    if let types::Payload::Greeting(info) = msg {
+                        if greeting_received {
+                            debug!(?peer, "received duplicate greeting");
+                            return Err(Error::DuplicateGreeting);
+                        }
+                        greeting_received = true;
+
+                        // Verify the greeting is from the expected peer
+                        if info.public_key != peer {
```
patrick-ogrady, line 232: This check is new (we weren't doing this on main over the first peers message received (but probably should've)

p2p/src/authenticated/discovery/types.rs
```diff
@@ -58,6 +60,13 @@ pub struct PayloadConfig {
 /// Payload is the only allowed message format that can be sent between peers.
 #[derive(Clone, Debug)]
 pub enum Payload<C: PublicKey> {
+    /// A greeting message containing the peer's own information.
+    ///
+    /// This must be the first message sent after connection establishment.
+    /// The connection will be terminated if this message is not received first
+    /// or if it is received more than once.
+    Greeting(Info<C>),
```
andresilva, line 71: I guess this would be a good place to include a network protocol version?

patrick-ogrady: Yeah, I think so.

Working on first making sure we actually send the greeting first on the sender side 😉

p2p/src/authenticated/discovery/actors/tracker/actor.rs
```diff
@@ -196,9 +196,9 @@ impl<E: Spawner + Rng + Clock + RuntimeMetrics, C: Signer> Actor<E, C> {
                 // Mark the record as connected
                 self.directory.connect(&public_key, dialer);
 
-                // Proactively send our own info to the peer
+                // Send greeting with our own info
                 let info = self.directory.info(&self.crypto.public_key()).unwrap();
-                let _ = peer.peers(vec![info]).await;
+                let _ = peer.greeting(info).await;
```
patrick-ogrady: I don't think this is strong enough to guarantee we send before any other message

p2p/src/authenticated/discovery/actors/spawner/actor.rs
```diff
@@ -153,12 +162,9 @@ impl<E: Spawner + Clock + Rng + CryptoRng + Metrics, O: Sink, I: Stream, C: Publ
                                 // Register peer with the router
                                 let channels = router.ready(peer.clone(), messenger).await;
 
-                                // Register peer with tracker
-                                tracker.connect(peer.clone(), is_dialer, peer_mailbox).await;
```
patrick-ogrady, line 157: This approach had a race where the router may send us a message first...we need to explicitly wait for the tracker to get back to us before we can start normal operation.

p2p/src/authenticated/discovery/actors/spawner/actor.rs
```diff
@@ -132,9 +132,18 @@ impl<E: Spawner + Clock + Rng + CryptoRng + Metrics, O: Sink, I: Stream, C: Publ
                             let is_dialer = matches!(reservation.metadata(), Metadata::Dialer(..));
                             let info_verifier = self.info_verifier.clone();
                             move |context| async move {
+                                // Get greeting from tracker (returns None if not eligible)
+                                let Some(greeting) = tracker.connect(peer.clone(), is_dialer).await
```
patrick-ogrady, line 130: This new approach avoids spinning up a peer if we connect at the same time via some other means

p2p/src/authenticated/discovery/actors/tracker/ingress.rs
```diff
@@ -1,9 +1,8 @@
 use super::Reservation;
+// Re-export for use in peer actor
+pub use crate::authenticated::discovery::actors::peer;
```
patrick-ogrady: remove this

p2p/src/authenticated/discovery/actors/peer/actor.rs
```diff
@@ -281,3 +319,411 @@ impl<E: Spawner + Clock + Rng + CryptoRng + Metrics, C: PublicKey> Actor<E, C> {
...
+    }
+
+    fn signed_peer_info(
+        signer: &mut PrivateKey,
+        socket: SocketAddr,
+        timestamp: u64,
+    ) -> types::Info<PublicKey> {
+        let ingress: crate::Ingress = socket.into();
+        let ip_namespace = commonware_utils::union(NAMESPACE, b"_IP");
+        let signature = signer.sign(&ip_namespace, &(ingress.clone(), timestamp).encode());
```
patrick-ogrady: We should have a helper for this?

p2p/src/authenticated/discovery/actors/spawner/actor.rs
```diff
@@ -132,9 +132,18 @@ impl<E: Spawner + Clock + Rng + CryptoRng + Metrics, O: Sink, I: Stream, C: Publ
                             let is_dialer = matches!(reservation.metadata(), Metadata::Dialer(..));
                             let info_verifier = self.info_verifier.clone();
                             move |context| async move {
+                                // Get greeting from tracker (returns None if not eligible)
```
patrick-ogrady, line 129: We don't resign info, so we should be able to cache this and avoid frequent lookups (although, we do need to know if the peer should be shutdown).

p2p/src/authenticated/discovery/actors/tracker/ingress.rs
```diff
@@ -39,18 +39,19 @@ pub enum Message<C: PublicKey> {
...
         /// The public key of the peer.
         public_key: C,
 
         /// `true` if we are the dialer, `false` if we are the listener.
         dialer: bool,
 
-        /// The mailbox of the peer actor.
-        peer: Mailbox<peer::Message<C>>,
+        /// One-shot channel to return the greeting info (or None if peer is not eligible).
+        responder: oneshot::Sender<Option<types::Info<C>>>,
```
BrendanChou: Thinking... should it just be type `oneshot::Sender<types::Info<C>>` and simply drop the oneshot instead of sending `None`? Or is that less clear

p2p/src/authenticated/discovery/types.rs
```diff
@@ -30,12 +30,14 @@ pub enum Error {
...
+const GREETING_PREFIX: u8 = 0;
 /// Prefix byte used to identify a [Payload] with variant BitVec.
-const BIT_VEC_PREFIX: u8 = 0;
+const BIT_VEC_PREFIX: u8 = 1;
 /// Prefix byte used to identify a [Payload] with variant Peers.
-const PEERS_PREFIX: u8 = 1;
+const PEERS_PREFIX: u8 = 2;
 /// Prefix byte used to identify a [Payload] with variant Data.
-const DATA_PREFIX: u8 = 2;
+const DATA_PREFIX: u8 = 3;
```
BrendanChou: nit: If we are making this breaking change anyway, it feels like maybe `DATA_PREFIX` should be `0` since we will always need this type in any p2p setup (discovery, etc.).

Then the other "special" p2p types can be 1...N. That way if we add more later, we don't have 3 special types, then the data type at 3, then more special types. It could also make debugging easier between p2p types since data type is consistently 0 for any commonware-p2p system

p2p/src/authenticated/discovery/actors/tracker/ingress.rs
patrick-ogrady: Ooooo, I like that.

p2p/src/authenticated/discovery/types.rs
patrick-ogrady: Ah, that's a good idea.

p2p/src/authenticated/discovery/actors/spawner/actor.rs
patrick-ogrady: I actually prefer leaving this as-is in case we start refreshing this.

p2p/src/authenticated/discovery/types.rs
patrick-ogrady: In lookup, this would look like adding a version to `PING`.

p2p/src/authenticated/discovery/actors/tracker/ingress.rs
```diff
@@ -39,18 +39,20 @@ pub enum Message<C: PublicKey> {
...
         /// The public key of the peer.
         public_key: C,
 
         /// `true` if we are the dialer, `false` if we are the listener.
         dialer: bool,
 
-        /// The mailbox of the peer actor.
-        peer: Mailbox<peer::Message<C>>,
+        /// One-shot channel to return the greeting info. Dropped if peer is not eligible.
+        responder: oneshot::Sender<types::Info<C>>,
```
patrick-ogrady, line 55: We should add our peer set knowledge to this to expedite discovery?

BrendanChou (approved):

p2p/src/authenticated/discovery/metrics.rs
```diff
@@ -18,6 +18,7 @@ impl Peer {
 
 #[derive(Clone, Copy, Debug, Hash, PartialEq, Eq)]
 pub enum MessageType {
+    Greeting,
```
clabby, line 22: for consistency, let's have the order of these variants match the order of `Payload`

p2p/src/authenticated/discovery/metrics.rs
```diff
@@ -27,6 +28,7 @@ pub enum MessageType {
 impl EncodeLabelValue for MessageType {
     fn encode(&self, encoder: &mut LabelValueEncoder<'_>) -> Result<(), std::fmt::Error> {
         match self {
+            Self::Greeting => encoder.write_str("greeting"),
```
clabby, line 32: same, match order of `Payload`?

p2p/src/authenticated/discovery/types.rs
```diff
@@ -58,31 +60,47 @@ pub struct PayloadConfig {
 /// Payload is the only allowed message format that can be sent between peers.
 #[derive(Clone, Debug)]
 pub enum Payload<C: PublicKey> {
+    /// Arbitrary data sent between peers.
```
clabby, line 63: Add docs that this can only be sent after the greeting has been exchanged?

patrick-ogrady: We already note this on discovered, I think it'd be redundant to write again.

p2p/src/authenticated/discovery/config.rs
```diff
@@ -111,19 +111,16 @@ pub struct Config<C: Signer> {
     ///
     /// If there is no other network activity, this message is used as a ping
     /// and should be sent more often than the read_timeout.
+    ///
+    /// This also determines the rate limit for incoming BitVec and Peers messages
+    /// (one per this frequency).
```
clabby:
```suggestion
    /// (one per half this frequency).
```

p2p/src/authenticated/lookup/config.rs
```diff
@@ -74,11 +74,10 @@ pub struct Config<C: Signer> {
     pub allowed_handshake_rate_per_subnet: Quota,
 
     /// Frequency at which we send ping messages to peers.
+    ///
+    /// This also determines the rate limit for incoming ping messages (one per this frequency).
```
clabby:
```suggestion
    /// This also determines the rate limit for incoming ping messages (one per half this frequency).
```

clabby: https://github.com/commonwarexyz/monorepo/pull/2553/changes#diff-c6474b08d3cfc2ac9245a525f3d93c93f6e0b47f16a70ba6340760936fd5857bR103

p2p/src/authenticated/discovery/config.rs
clabby: https://github.com/commonwarexyz/monorepo/pull/2553/changes#diff-fe4f2af0d16ac70ed16379a10df2228e65c7a4f5a3f6097df443bff6f70efc5bR184

p2p/src/authenticated/discovery/actors/tracker/ingress.rs
patrick-ogrady: We already send a BitVec, so this is unnecessary (more maintainable not to combo messages)

PR #2555, opened 2025-12-18
[stream] encode length as varint

### LINKED ISSUE #753: [stream] Consider framing using a varint rather than a fixed number of bytes
This would save 1-3 bytes on all network messages that are 2MB or fewer (so basically all of them). No bandwidth savings for messages 2-256 MB. Bandwidth loss of one byte for messages larger than 256 MB

Relevant functions: `send_frame/recv_frame` in the `commonware_stream` package

Sort of annoying to implement since you have to read one byte at a time because of how `send/recv` work. Consider modifying the traits that the methods are a part of.

andresilva:
This PR implements varint-based framing for `send_frame`/`recv_frame` in commonware_stream, replacing the fixed 4-byte length prefix with a variable-length encoding. This saves 1-3 bytes on most network messages (those 2MB or smaller), which covers most real-world usage.

As part of this change, `max_message_size` was changed from `usize` to `u32` throughout the codebase.

Closes #753.


stream/src/lib.rs
```diff
@@ -310,15 +312,15 @@ impl<O: Sink> Sender<O> {
...
+    max_message_size: u32,
 }
 
 impl<I: Stream> Receiver<I> {
     /// Receives and decrypts a message from the peer.
     pub async fn recv(&mut self) -> Result<Bytes, Error> {
         let c = recv_frame(
             &mut self.stream,
-            self.max_message_size + CIPHERTEXT_OVERHEAD,
+            self.max_message_size + CIPHERTEXT_OVERHEAD as u32,
```
patrick-ogrady: change CIPHERTEXT_OVERHEAD to `u32`?

stream/src/utils/codec.rs
```diff
@@ -1,40 +1,51 @@
...
 pub async fn send_frame<S: Sink>(
     sink: &mut S,
     buf: &[u8],
-    max_message_size: usize,
+    max_message_size: u32,
 ) -> Result<(), Error> {
     // Validate frame size
     let n = buf.len();
-    if n > max_message_size {
+    if n > max_message_size as usize {
```
patrick-ogrady, line 19: Could do a `u32::try_into()` before this check to avoid so much casting

stream/src/utils/codec.rs
```diff
@@ -1,40 +1,51 @@
...
+            Ok(Some(len)) => break len as usize,
+            Ok(None) => continue,
+            Err(_) => return Err(Error::InvalidVarint),
+        }
+    };
 
     // Validate frame size
-    let len = u32::from_be_bytes(len_buf.as_ref()[..4].try_into().unwrap()) as usize;
-    if len > max_message_size {
+    if len > max_message_size as usize {
```
patrick-ogrady, line 48: May be out of the loop here, but why doesn't Decoder return `u32`?

patrick-ogrady (commented):
small nits

patrick-ogrady:
Will initially be a little less efficient but should be same ~speed with buffered networking ops

stream/src/utils/codec.rs
andresilva: I cast it to `usize` above.

stream/src/utils/codec.rs
andresilva: These casts should never fail though (unless you run this code in a 16-bit computer).

stream/src/lib.rs
andresilva: Started this but it proliferated a bit more.

PR #2558, opened 2025-12-18
[runtime/p2p] Use `Buf` / `BufMut` in networking APIs

### LINKED ISSUE #784: [runtime/stream] Support the use of `OwnedWriteHalf::write_vectored`

clabby:
## Overview

Alters the core networking APIs to use `Buf` / `BufMut` rather than `StableBuf`. This unlocks passing non-contiguous buffers + vectorized read/write operations.

closes #784

### DIFF
--- a/p2p/src/utils/mux.rs
+++ b/p2p/src/utils/mux.rs
@@ -359,7 +359,7 @@ impl<'a, S: Sender> CheckedSender for CheckedGlobalSender<'a, S> {
     ) -> Result<Vec<Self::PublicKey>, Self::Error> {
         let subchannel = UInt(self.subchannel.expect("subchannel not set"));
         self.inner
-            .send(subchannel.encode().freeze().chain(message), priority)
+            .send(subchannel.encode().chain(message), priority)
             .await
     }
 }
--- a/stream/src/lib.rs
+++ b/stream/src/lib.rs
@@ -197,13 +197,13 @@ pub async fn dial<R: CryptoRngCore + Clock, S: Signer, I: Stream, O: Sink>(
                 peer,
             ),
         );
-        send_frame(&mut sink, syn.encode().freeze(), config.max_message_size).await?;
+        send_frame(&mut sink, syn.encode(), config.max_message_size).await?;
 
         let syn_ack_bytes = recv_frame(&mut stream, config.max_message_size).await?;
         let syn_ack = SynAck::<S::Signature>::decode(syn_ack_bytes)?;
 
         let (ack, send, recv) = dial_end(state, syn_ack)?;
-        send_frame(&mut sink, ack.encode().freeze(), config.max_message_size).await?;
+        send_frame(&mut sink, ack.encode(), config.max_message_size).await?;
 
         Ok((
             Sender {
@@ -264,12 +264,7 @@ pub async fn listen<
             ),
             msg1,
         )?;
-        send_frame(
-            &mut sink,
-            syn_ack.encode().freeze(),
-            config.max_message_size,
-        )
-        .await?;
+        send_frame(&mut sink, syn_ack.encode(), config.max_message_size).await?;
 
         let ack_bytes = recv_frame(&mut stream, config.max_message_size).await?;
         let ack = Ack::decode(ack_bytes)?;
--- a/stream/src/utils/codec.rs
+++ b/stream/src/utils/codec.rs
@@ -21,7 +21,7 @@ pub async fn send_frame<S: Sink>(
 
     // Prefix `buf` with its varint-encoded length and send it
     let len = UInt(n as u32);
-    let data = len.encode().freeze().chain(buf);
+    let data = len.encode().chain(buf);
     sink.send(data).await.map_err(Error::SendFailed)
 }
 
--- a/codec/src/codec.rs
+++ b/codec/src/codec.rs
@@ -1,7 +1,7 @@
 //! Core traits for encoding and decoding.
 
 use crate::error::Error;
-use bytes::{Buf, BufMut, BytesMut};
+use bytes::{Buf, BufMut, Bytes, BytesMut};
 
 /// Trait for types with a known, fixed encoded size.
 ///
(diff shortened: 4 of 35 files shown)


p2p/src/authenticated/lookup/actors/router/ingress.rs
```diff
@@ -74,15 +74,15 @@ impl<P: PublicKey> Messenger<P> {
...
+        mut message: impl Buf + Send,
         priority: bool,
     ) -> Vec<P> {
         let (sender, receiver) = oneshot::channel();
         self.sender
             .send(Message::Content {
                 recipients,
                 channel,
-                message,
+                message: message.copy_to_bytes(message.remaining()),
```
clabby, line 85: This is an unfortunate limitation... thinking about ways around it.

clabby: I think that this is going to be too complex for the performance benefits. We already get a good benefit from the existing changes preventing copies closer to the runtime / stream.

clabby: To do this I think we'd need to sort out some way to completely bypass the router ^^

p2p/src/utils/mux.rs
```diff
@@ -354,14 +354,13 @@ impl<'a, S: Sender> CheckedSender for CheckedGlobalSender<'a, S> {
...
+        message: impl Buf + Send,
         priority: bool,
     ) -> Result<Vec<Self::PublicKey>, Self::Error> {
         let subchannel = UInt(self.subchannel.expect("subchannel not set"));
-        let mut buf = BytesMut::with_capacity(subchannel.encode_size() + message.len());
-        subchannel.write(&mut buf);
-        buf.put_slice(&message);
-        self.inner.send(buf.freeze(), priority).await
+        self.inner
+            .send(subchannel.encode().chain(message), priority)
```
andresilva, line 362: 🚀

runtime/src/network/tokio.rs
```diff
@@ -38,19 +38,30 @@ pub struct Stream {
...
+            let len = buf.remaining_mut();
+            while read < len {
+                let n = self
+                    .stream
+                    .read_buf(&mut buf)
+                    .await
+                    .map_err(|_| Error::RecvFailed)?;
+
+                if n == 0 {
+                    return Err(Error::Closed);
```
andresilva: We should update deterministic and io-uring to match this.

p2p/src/authenticated/lookup/actors/router/ingress.rs
andresilva: With the current code this is only an issue for `mux` which uses chain and will require a memcopy here, everything else that gets here will just be a cheap refcount increase since it's already `Bytes` AFAIU.

stream/src/utils/codec.rs
```diff
@@ -1,31 +1,28 @@
...
         return Err(Error::SendTooLarge(n));
     }
 
     // Prefix `buf` with its varint-encoded length and send it
     let len = UInt(n as u32);
-    let mut prefixed_buf = BytesMut::with_capacity(len.encode_size() + buf.len());
-    len.write(&mut prefixed_buf);
-    prefixed_buf.extend_from_slice(buf);
-    sink.send(prefixed_buf).await.map_err(Error::SendFailed)
+    let data = len.encode().freeze().chain(buf);
```
andresilva: 🚀

(this one will actually get vectored)

p2p/src/authenticated/lookup/actors/router/ingress.rs
andresilva: Also when this reaches stream it needs to be contiguous memory anyway for chacha20poly1305 😢.

stream/src/lib.rs
```diff
@@ -305,7 +310,7 @@ impl<O: Sink> Sender<O> {
         let c = self.cipher.send(msg)?;
         send_frame(
```
andresilva, line 309: Can't comment on the lines above but would it make sense to take `msg: impl Buf` here?

p2p/src/authenticated/lookup/actors/router/ingress.rs
andresilva: I guess we could have a fast path for a single recipient where copying isn't needed, but probably not worth it?

runtime/src/network/tokio.rs
clabby: Great spot.. I think for consistency, we should just return `RecvFailed` here.

stream/src/lib.rs
clabby: https://github.com/commonwarexyz/monorepo/pull/2558/changes/bba6c259f64e418c93294e1d57126d829cffcdd2

runtime/src/network/tokio.rs
clabby: https://github.com/commonwarexyz/monorepo/pull/2558/changes/b4094d92bfca95834409c16336baca7e02ee8854

p2p/src/authenticated/lookup/actors/router/ingress.rs
clabby: > I guess we could have a fast path for a single recipient where copying isn't needed, but probably not worth it?

Was thinking the same.. We need better benchmarks for the networking stack to know for sure, though. Leaning towards not adding complexity here unless we have #s that show it's worth it.

runtime/src/lib.rs
```diff
@@ -490,10 +488,7 @@ pub trait Stream: Sync + Send + 'static {
     /// # Warning
     ///
     /// If the stream returns an error, partially read data may be discarded.
-    fn recv(
-        &mut self,
-        buf: impl Into<StableBuf> + Send,
-    ) -> impl Future<Output = Result<StableBuf, Error>> + Send;
+    fn recv(&mut self, buf: impl BufMut + Send) -> impl Future<Output = Result<(), Error>> + Send;
```
patrick-ogrady, line 491: Should we add a trait alias for `BufMut + Send`

clabby: I don't think so; It's only used 7 times in the whole codebase & is a bit more clear to have inline.

andresilva (changes requested):
Maybe we could have:
- `Encode::encode_mut() -> BytesMut`
- `Encode::encode() -> Bytes { self.encode_mut().freeze() }` 

This would remove all of the `encode().freeze()` noise and make sure we actually freeze as much as we can (I think we are missing a lot of places?). We probably wouldn't use `encode_mut()` anywhere right now (but not sure about this, should be trivial anyway).

PR #2559, opened 2025-12-18
[cryptography/certificate] make namespace a part of the scheme

### LINKED ISSUE #2521: [consensus] Avoid Dynamic Computation of Namespace Helpers
In #2220, we dynamically generate namespaces like:

```
impl<D: Digest> Subject for &Item<D> {
    fn namespace_and_message(&self, namespace: &[u8]) -> (Bytes, Bytes) {
        (ack_namespace(namespace).into(), self.encode().freeze())
    }
}
```

We should instead construct these namespaces once and clone them (via `Bytes::clone()` rather than re-generating each time).

andresilva:
This PR refactors the certificate scheme API to store the namespace in the scheme at construction time rather than passing it on every sign/verify call. This avoids redundant namespace derivation and simplifies the signing API. 

The previous API required passing namespace bytes to every `sign()`, `verify_attestation()`, and `verify_attestations()` call. For protocols like simplex with multiple message types (notarize, nullify, finalize, seed), this meant recomputing derived namespace bytes on every operation.

The new design introduces a `Namespace` trait for types that can derive themselves from a base namespace. The `Subject` trait now has an associated `Namespace` type, and the scheme computes this namespace once at construction. Sign and verify methods no longer need the namespace parameter since it's already stored in the scheme.

Closes #2521.

### DIFF
--- a/consensus/src/ordered_broadcast/types.rs
+++ b/consensus/src/ordered_broadcast/types.rs
@@ -645,14 +645,14 @@ impl<P: PublicKey, S: Scheme, D: Digest> Node<P, S, D> {
                 .ok_or(Error::ParentMissing)?,
             parent.digest,
         );
+        let parent_scheme = provider
+            .scoped(parent.epoch)
+            .ok_or(Error::UnknownScheme(parent.epoch))?;
         let ctx = AckSubject {
             chunk: &parent_chunk,
             epoch: parent.epoch,
         };
-        let scheme = provider
-            .scoped(parent.epoch)
-            .ok_or(Error::UnknownScheme(parent.epoch))?;
-        if !scheme.verify_certificate::<R, D>(rng, ctx, &parent.certificate) {
+        if !parent_scheme.verify_certificate::<R, D>(rng, ctx, &parent.certificate) {
             return Err(Error::InvalidCertificate);
         }
         Ok(Some(parent_chunk))


consensus/src/aggregation/types.rs
```diff
@@ -90,6 +90,19 @@ fn ack_namespace(namespace: &[u8]) -> Vec<u8> {
     union(namespace, ACK_SUFFIX)
 }
 
+/// Namespace type for aggregation acknowledgments.
+///
+/// This type encapsulates the pre-computed namespace bytes used for signing and
+/// verifying acks.
+#[derive(Clone, Debug)]
+pub struct Namespace(Vec<u8>);
```
patrick-ogrady, line 98: Considered suggesting we enshrine this type in `cryptography` but think that is out of scope/separate PR-worthy)

consensus/src/simplex/scheme/mod.rs
```diff
@@ -35,9 +35,56 @@ pub mod secp256r1;
 #[cfg(not(target_arch = "wasm32"))]
 pub mod reporter;
 
+/// Pre-computed namespaces for simplex voting subjects.
+///
+/// This struct holds the pre-computed namespace bytes for each vote type.
+#[derive(Clone, Debug)]
+pub struct Namespace {
```
patrick-ogrady, line 42: Ah, now I see why this isn't enshrined.

consensus/src/ordered_broadcast/ack_manager.rs
```diff
@@ -612,13 +612,13 @@ mod tests {
     fn interleaved_payloads<S, F>(fixture: F)
     where
         S: Scheme<PublicKey, Sha256Digest>,
-        F: FnOnce(&mut StdRng, u32) -> Fixture<S>,
+        F: FnOnce(&[u8], &mut StdRng, u32) -> Fixture<S>,
```
patrick-ogrady: nit: put `rng` before namespace?

consensus/src/ordered_broadcast/mod.rs
```diff
@@ -382,7 +383,7 @@ mod tests {
     fn unclean_shutdown<S, F>(fixture: F)
     where
         S: Scheme<PublicKey, Sha256Digest>,
-        F: Fn(&mut deterministic::Context, u32) -> Fixture<S> + Clone,
+        F: Fn(&[u8], &mut deterministic::Context, u32) -> Fixture<S> + Clone,
```
patrick-ogrady: put `context` first?

consensus/src/ordered_broadcast/types.rs
```diff
@@ -482,17 +500,11 @@ impl<P: PublicKey, S: Scheme, D: Digest> Node<P, S, D> {
...
             chunk: &parent_chunk,
             epoch: parent.epoch,
         };
-        if !parent_scheme.verify_certificate::<R, D>(
-            rng,
-            &ack_namespace,
-            ack_ctx,
-            &parent.certificate,
-        ) {
+        if !parent_scheme.verify_certificate::<R, D>(rng, ack_ctx, &parent.certificate) {
```
patrick-ogrady: rng before context here (explains order 🤔 )

examples/bridge/src/bin/validator.rs
patrick-ogrady, line 225: We should be able to pass 2 verifiers in here instead of this.

consensus/src/ordered_broadcast/types.rs
```diff
@@ -581,6 +589,74 @@ impl<P: PublicKey, S: Scheme, D: Digest> Node<P, S, D> {
...
+            return Ok(None);
+        };
+
+        // Verify parent (if present)
+        let parent_chunk = Chunk::new(
+            self.chunk.sequencer.clone(),
+            self.chunk
+                .height
+                .checked_sub(1)
+                .expect("non-genesis nodes should have height > 0"),
```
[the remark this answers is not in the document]
patrick-ogrady: fixing

PR #2565, opened 2025-12-18
[consensus] Add `certify()` without `Context`

### LINKED ISSUE #1767: [consensus::simplex] Allow for a second, post-notarization-certificate, `verify(...)` hook to the `Automaton`, called `finalizable(...)`

BrendanChou:
Fixes #1767 

- Resolver only uses finalizations as the "floor" (rather than notarizations)
- `emit_floor` replaced by `parent_certificate` which is more efficient

### DIFF
--- a/consensus/src/simplex/actors/resolver/state.rs
+++ b/consensus/src/simplex/actors/resolver/state.rs
@@ -86,7 +86,7 @@ impl<S: Scheme, D: Digest> State<S, D> {
             }
             Certificate::Finalization(finalization) => {
                 let view = finalization.view();
-                if self.encounter_view(view) {
+                if self.encounter_view(view) || self.can_upgrade_floor(view) {
                     self.floor = Some(Certificate::Finalization(finalization));
                     self.prune(resolver).await;
                 }
@@ -164,6 +164,17 @@ impl<S: Scheme, D: Digest> State<S, D> {
             .as_ref()
             .map(|floor| floor.view())
             .unwrap_or(View::zero())
+    }
+
+    /// Returns true if the floor can be upgraded at the given view.
+    ///
+    /// A finalization can upgrade a notarization at the same view since
+    /// finalization is a stronger proof than notarization.
+    fn can_upgrade_floor(&self, view: View) -> bool {
+        matches!(
+            self.floor.as_ref(),
+            Some(Certificate::Notarization(n)) if n.view() == view
+        )
     }
 
     /// Inform the [Resolver] of any missing nullifications.
@@ -547,5 +558,46 @@ mod tests {
         assert!(!state.satisfied_by.contains_key(&View::new(5)));
         // View 5 should not be marked as failed
         assert!(!state.is_failed(View::new(5)));
+    }
+
+    #[test_async]
+    async fn finalization_upgrades_certified_notarization_at_same_view() {
+        let (schemes, verifier) = ed25519_fixture();
+        let mut state: State<TestScheme, Sha256Digest> = State::new(10);
+        let mut resolver = MockResolver::default();
+
+        // Create and certify a notarization at view 5
+        let notarization_v5 = build_notarization(&schemes, &verifier, View::new(5));
+        state
+            .handle(
+                Certificate::Notarization(notarization_v5.clone()),
+                None,
+                &mut resolver,
+            )
+            .await;
+        state
+            .handle_certified(View::new(5), true, &mut resolver)
+            .await;
+
+        // Floor should be the notarization at view 5
+        assert!(
+            matches!(state.floor.as_ref(), Some(Certificate::Notarization(n)) if n == &notarization_v5)
+        );
+        assert_eq!(state.floor_view(), View::new(5));
+
+        // A finalization at the same view should upgrade the floor
+        let finalization_v5 = build_finalization(&schemes, &verifier, View::new(5));
+        state
+            .handle(
+                Certificate::Finalization(finalization_v5.clone()),
+                None,
+                &mut resolver,
+            )
+            .await;
+
+        // Floor should now be the finalization (stronger proof)
+        assert!(
+            matches!(state.floor.as_ref(), Some(Certificate::Finalization(f)) if f == &finalization_v5)
+        );
     }
 }
--- a/consensus/src/simplex/actors/voter/actor.rs
+++ b/consensus/src/simplex/actors/voter/actor.rs
@@ -762,9 +762,11 @@ impl<
                             continue;
                         };
                         resolver.certified(round.view(), success).await;
-                        self.reporter
-                            .report(Activity::Certification(notarization))
-                            .await;
+                        if success {
+                            self.reporter
+                                .report(Activity::Certification(notarization))
+                                .await;
+                        }
                     }
                     Artifact::Nullify(nullify) => {
                         self.handle_nullify(nullify.clone()).await;
@@ -969,9 +971,11 @@ impl<
                                 continue;
                             };
                             resolver.certified(view, certified).await;
-                            self.reporter
-                                .report(Activity::Certification(notarization))
-                                .await;
+                            if certified {
+                                self.reporter
+                                    .report(Activity::Certification(notarization))
+                                    .await;
+                            }
                         }
                         Err(err) => {
                             // The application did not explicitly respond whether certification succeeded.


BrendanChou:
bugbot run

consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -25,14 +25,21 @@ use commonware_runtime::{
     buffer::PoolRef, spawn_cell, Clock, ContextCell, Handle, Metrics, Spawner, Storage,
 };
 use commonware_storage::journal::segmented::variable::{Config as JConfig, Journal};
+use commonware_utils::futures::AbortablePool;
+use core::{future::Future, panic};
```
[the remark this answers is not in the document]
BrendanChou: panic is used. no issue

consensus/src/simplex/mocks/relay.rs
```diff
@@ -40,7 +50,12 @@ impl<D: Digest, P: PublicKey> Relay<D, P> {
         receiver
     }
 
-    pub async fn broadcast(&self, sender: &P, payload: (D, Bytes)) {
+    /// Broadcasts a payload to all registered recipients.
+    pub async fn broadcast(&self, sender: &P, (payload, data): (D, Bytes)) {
```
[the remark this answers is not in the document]
BrendanChou: no change here, was already sender

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -508,30 +565,24 @@ impl<E: Clock + Rng + CryptoRng + Metrics, S: Scheme<D>, L: ElectorConfig<S>, D:
             return None;
         }
 
-        // May return `None` if the parent view is not yet notarized or finalized.
-        self.quorum_payload(parent).copied()
+        // May return `None` if the parent view is not yet either:
+        // - notarized and certified
+        // - finalized
```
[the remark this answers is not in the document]
BrendanChou: Seems clear to me

consensus/src/simplex/actors/resolver/state.rs
```diff
@@ -97,16 +75,24 @@ impl<S: Scheme, D: Digest> State<S, D> {
         // If view is <= floor, return the floor
         if let Some(floor) = &self.floor {
             if view <= floor.view() {
-                return Some(floor.clone());
+                return Some(Certificate::Finalization(floor.clone()));
```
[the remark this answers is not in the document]
BrendanChou: Doesn't make sense to me

consensus/src/simplex/mocks/application.rs
```diff
@@ -110,6 +126,18 @@ const GENESIS_BYTES: &[u8] = b"genesis";
...
 type Latency = (f64, f64);
 
+/// Predicate to determine whether a payload should be certified.
+/// Returning true means certify, false means reject.
+pub enum CertifyFn<D: Digest> {
+    /// Always certify.
+    Always,
+    /// Certify sometimes, but not always. The behavior is to certify pseudorandomly
+    /// (but deterministically) 82% of the time, depending on the last byte of the payload.
+    Sometimes,
```
patrick-ogrady, line 138: nit: add an argument on this to configure likelihood?

consensus/src/simplex/scheme/reporter.rs
```diff
@@ -101,7 +101,8 @@ impl<
                 }
                 Activity::Notarization(_)
                 | Activity::Nullification(_)
-                | Activity::Finalization(_) => {
+                | Activity::Finalization(_)
+                | Activity::Certification(_) => {
```
patrick-ogrady: Certification before finalization?

consensus/src/simplex/types.rs
```diff
@@ -488,6 +488,8 @@ pub enum Artifact<S: Scheme, D: Digest> {
     Finalize(Finalize<S, D>),
     /// A recovered certificate for a finalization.
     Finalization(Finalization<S, D>),
+    /// A notarization was locally certified.
```
patrick-ogrady: Certification before finalization?

consensus/src/lib.rs
```diff
@@ -99,6 +99,25 @@ cfg_if::cfg_if! {
...
+            /// finalization until they have reconstructed and validated the full block (e.g., after
+            /// receiving enough shards).
+            fn certify(
+                &mut self,
+                _payload: Self::Digest,
+            ) -> impl Future<Output = oneshot::Receiver<bool>> + Send {
+                #[allow(clippy::async_yields_async)]
+                async move {
+                    let (sender, receiver) = oneshot::channel();
+                    let _ = sender.send(true);
```
patrick-ogrady, line 125: Its a bit odd (maybe) that consumers of this default certify function will receive activity for it?

consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -339,8 +394,13 @@ impl<
         }
         self.append_journal(view, artifact).await;
 
-        // If we were the proposer, we should emit the notarization that we built our proposal on
-        self.state.emit_floor(view)
+        // If we were the leader, we should emit a certifiable chain of ancestor certificates such
```
patrick-ogrady: nit: improve wording

consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -813,29 +907,48 @@ impl<
...
+                        continue;
+                    };
+
+                    // Handle response to our certification request.
+                    view = round.view();
+                    match certified {
+                        Ok(certified) => {
+                            self.handle_certification(view, certified).await;
+                        }
+                        Err(err) => {
```
patrick-ogrady, line 980: nit: not sure if we should error here?

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -197,20 +205,21 @@ impl<E: Clock + Rng + CryptoRng + Metrics, S: Scheme<D>, L: ElectorConfig<S>, D:
...
+        // Get the certificate for the previous view. Prefer finalizations since they are the
+        // strongest proof available. Prefer nullifications over notarizations since they do not
+        // require certification to move to the next view.
         #[allow(clippy::option_if_let_else)]
         let cert = if let Some(finalization) = self.finalization(entry_view).cloned() {
             Some(Certificate::Finalization(finalization))
-        } else if let Some(notarization) = self.notarization(entry_view).cloned() {
-            Some(Certificate::Notarization(notarization))
         } else if let Some(nullification) = self.nullification(entry_view).cloned() {
             Some(Certificate::Nullification(nullification))
```
patrick-ogrady, line 221: nit: explain we prefer nullifications because they `overwrite` a notarization of uncertified data

consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -681,6 +764,9 @@ impl<
                             .report(Activity::Finalization(finalization))
                             .await;
                     }
+                    Artifact::Certification(round, success) => {
+                        self.handle_certification(round.view(), success).await;
+                    }
                 }
             }
         }
```
patrick-ogrady, line 799: Do we trigger re-certify on restart?

consensus/src/simplex/actors/resolver/state.rs
```diff
@@ -1,32 +1,34 @@
...
 use commonware_cryptography::{certificate::Scheme, Digest};
 use commonware_resolver::Resolver;
 use commonware_utils::sequence::U64;
 use std::collections::BTreeMap;
-use tracing::debug;
 
 /// Tracks all known certificates from the last
-/// notarized or finalized view to the current view.
+/// finalized view to the current view.
 pub struct State<S: Scheme, D: Digest> {
```
patrick-ogrady, line 13: Make sure we update mod.rs accordingly

consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -339,8 +394,13 @@ impl<
...
         self.append_journal(view, artifact).await;
 
-        // If we were the proposer, we should emit the notarization that we built our proposal on
-        self.state.emit_floor(view)
+        // If we were the leader, we should emit a certifiable chain of ancestor certificates such
+        // that other peers are able to verify my proposal in the future.
+        let leader = self.state.leader_index(view)?;
+        if self.state.is_me(leader) {
```
patrick-ogrady: can make more succinct response

consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -713,42 +799,51 @@ impl<
         let mut shutdown = self.context.stopped();
 
         // Process messages
-        let mut pending_set = None;
-        let mut pending_propose_context = None;
-        let mut pending_propose = None;
-        let mut pending_verify_context = None;
-        let mut pending_verify = None;
+        let mut pending_propose: Option<Request<Context<D, S::PublicKey>, D>> = None;
```
patrick-ogrady, line 829: very nice cleanup here 😓

consensus/src/simplex/actors/voter/actor.rs
patrick-ogrady: This can happen on shutdown and we don't want to lock-in on a failed certification?

I think it may be nicer to just break here (I suppose this could happen though if we finalize something higher)?

patrick-ogrady (commented):
initial comments

consensus/src/simplex/actors/resolver/state.rs
```diff
@@ -118,29 +98,28 @@ impl<S: Scheme, D: Digest> State<S, D> {
     /// Inform the [Resolver] of any missing nullifications.
     async fn fetch(&mut self, resolver: &mut impl Resolver<Key = U64>) {
         // We must either receive a nullification or a notarization (at the view or higher),
-        // so we don't need to worry about getting stuck. All requests will be resolved.
-        let mut cursor = self.floor_view().next();
-        let mut requests = Vec::new();
```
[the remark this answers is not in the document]
patrick-ogrady: This is totally fine (not an issue). We don't want to re-request just because the finalization floor went up?

consensus/src/lib.rs
patrick-ogrady: I suppose the best we could do here is to provide a config for "certify events" but that may be a bit annoying to keep straight (especially if it is expected we don't write certifications in that case)

consensus/src/simplex/actors/voter/actor.rs
patrick-ogrady: This is now well-handled after #2575

consensus/src/simplex/actors/resolver/state.rs
patrick-ogrady: Handled in #2575

consensus/src/simplex/mocks/application.rs
patrick-ogrady: This is good enough...

consensus/src/simplex/mocks/relay.rs
```diff
@@ -54,12 +69,31 @@ impl<D: Digest, P: PublicKey> Relay<D, P> {
...
-                if let Err(e) = listener.send((payload.0, payload.1.clone())).await {
+                if let Err(e) = listener.send((payload, data.clone())).await {
                     error!(?e, ?recipient, "failed to send message to recipient");
                 }
             }
         }
     }
+
+    /// Requests that a payload is sent to a public key.
+    pub async fn request(&self, payload: D, public_key: P) {
```
patrick-ogrady: nit: public_key first

consensus/src/simplex/mocks/application.rs
```diff
@@ -110,6 +126,18 @@ const GENESIS_BYTES: &[u8] = b"genesis";
 
 type Latency = (f64, f64);
 
+/// Predicate to determine whether a payload should be certified.
+/// Returning true means certify, false means reject.
+pub enum CertifyFn<D: Digest> {
```
patrick-ogrady: nit: `Certifier`

consensus/src/simplex/actors/voter/round.rs
```diff
@@ -207,6 +249,15 @@ impl<S: Scheme, D: Digest> Round<S, D> {
         }
     }
 
+    /// Marks proposal certification as complete.
+    pub fn certified(&mut self, is_success: bool) {
+        assert!(
+            self.certified.is_none_or(|v| v == is_success),
+            "certification should not conflict"
+        );
+        self.certified = Some(is_success);
```
andresilva, line 258: Can clear `certify_handle`?

consensus/src/simplex/actors/resolver/state.rs
```diff
@@ -36,54 +38,26 @@ impl<S: Scheme, D: Digest> State<S, D> {
...
-
-                // Set last notarized
-                if self.floor.as_ref().is_none_or(|floor| view > floor.view()) {
-                    self.floor = Some(Certificate::Notarization(notarization));
-                }
-
-                // Prune old nullifications
-                self.prune(resolver).await;
+                // Update current view (notarizations don't set the floor)
+                self.encounter_view(notarization.view());
```
andresilva: In the common case where a view is certified but not yet finalized (or finalization lags), the fetches for nullifications never resolve and remain outstanding until the view gets finalized. Should we stop requesting nullifications for views that have already been notarized **and** certified?

patrick-ogrady: You are right, not all requests will resolve anymore 👁️

patrick-ogrady: The issue with this is that unfetchable nullifications can block fetching of useful data

patrick-ogrady: I think the summary here is that we expect it to be very rare that there isn't a finalization or a nullification at some view (i.e. just a certified notarization).

To improve performance in this case, it is possible that we could stop fetching on some view if we have a local certification but then we run the risk of not collecting necessary nullifications to skip some view.

Ultimately, we need some "default" strategy that causes everyone to align.

patrick-ogrady:
TODO: add a success guard around reporter certification

PR #2567, opened 2025-12-18
[storage] make Persistable trait a top-level Storage construct

roberto-bayardo:
Persistable is a trait that is useful more generally than for K/V stores and journals. This PR pulls it up a level to replace both `PersistentContiguous` (persistent journal) and `StorePersistent` (k/v stores), allowing it to be implemented by any Storage structure.


roberto-bayardo:
Closing this, it makes more sense in the context of https://github.com/commonwarexyz/monorepo/pull/2580

PR #2571, opened 2025-12-19
[workspace] `retain_recent` on rate-limiters

### LINKED ISSUE #2473: Review usage of `governor::shrink_to_fit()`
I think we should actually be calling `retain_recent()` rather than `shrink_to_fit()` (which just ensures the data structure capacity tightly fits its size but doesn't actually remove any keys).

clabby:
## Overview

Adjusts the usage of `shrink_to_fit` and `retain_recent` on the keyed rate limiters in the codebase.

closes #2473

### DIFF
--- a/p2p/src/authenticated/discovery/actors/listener.rs
+++ b/p2p/src/authenticated/discovery/actors/listener.rs
@@ -191,12 +191,8 @@ impl<E: Spawner + Clock + Network + Rng + CryptoRng + Metrics, C: Signer> Actor<
 
                 // Cleanup the rate limiters periodically
                 if accepted > CLEANUP_INTERVAL {
-                    // Clean up old entries and shrink the capacity of the rate limiters
-                    // to free up memory.
                     ip_rate_limiter.retain_recent();
-                    ip_rate_limiter.shrink_to_fit();
                     subnet_rate_limiter.retain_recent();
-                    subnet_rate_limiter.shrink_to_fit();
                     accepted = 0;
                 }
                 accepted += 1;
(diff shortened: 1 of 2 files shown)


p2p/src/authenticated/lookup/actors/listener.rs
```diff
@@ -219,7 +219,11 @@ impl<E: Spawner + Clock + Network + Rng + CryptoRng + Metrics, C: Signer> Actor<
 
                 // Cleanup the rate limiters periodically
                 if accepted > CLEANUP_INTERVAL {
+                    // Clean up old entries and shrink the capacity of the rate limiters
+                    // to free up memory.
```
clabby: These cleanups happen relatively infrequently and the RLs grow a lot larger than the others. It probably makes sense to clean up the capacity here and in discovery's listener actor due to a potential large volume of incoming requests being able to indefinitely expand memory otherwise..

p2p/src/authenticated/discovery/actors/listener.rs
```diff
@@ -191,7 +191,11 @@ impl<E: Spawner + Clock + Network + Rng + CryptoRng + Metrics, C: Signer> Actor<
 
                 // Cleanup the rate limiters periodically
                 if accepted > CLEANUP_INTERVAL {
+                    // Clean up old entries and shrink the capacity of the rate limiters
+                    // to free up memory.
+                    ip_rate_limiter.retain_recent();
```
patrick-ogrady, line 194: We already bound the growth here with IP rate limiting, so this can't actually grow unbounded?

clabby: That limit is keyed though, yeah? i.e. subsequent distributed requests can populate the limiter over a range of keys. (Though ideally, this is handled via system ingress RL, outside of the application.)

patrick-ogrady: yeah, there can definitely be runs of different "worseness" based on the spread of keys but still bounded.

patrick-ogrady: I think its ok to keep around just-in-case but yeah, I'd be totally ok to just retain_recent here (under the assumption that if it has grown to some size, it'll probably stay that size).

patrick-ogrady (approved):

BrendanChou:
bugbot run

PR #2572, opened 2025-12-19
[runtime] Disallow invalid metric names

clabby:
## Overview

Disallows invalid metric names in `Metrics::with_label`, using a regex that's slightly more strict than [Prometheus' recommendations](https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels) (`[a-zA-Z0-9_:*`, but with an ASCII-letter as the first character.)


patrick-ogrady:
🙏

patrick-ogrady (approved):
While I originally pondered overwriting what was provided in labels, it feels more correct to assert (given it'll bork metrics).

PR #2573, opened 2025-12-19
[runtime] Remove `AsRef` and `AsMut` impls for `Cell<T>`

clabby:
## Overview

Removes the `AsRef<C>` and `AsMut<C>` impls from `Cell<T>` in favor of functions that document explicitly that they will panic under the condition that the context is `Missing`.

https://doc.rust-lang.org/stable/std/convert/trait.AsRef.html#relation-to-borrow / https://doc.rust-lang.org/stable/std/convert/trait.AsMut.html both state that implementations "must not fail."

### DIFF
--- a/runtime/src/utils/cell.rs
+++ b/runtime/src/utils/cell.rs
@@ -63,26 +63,36 @@ impl<C> Cell<C> {
         }
     }
 
-    /// Consume the slot, returning the context and panicking if it is missing.
-    pub fn into(self) -> C {
+    /// Returns a reference to the context.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the context is missing.
+    pub fn as_present(&self) -> &C {
         match self {
             Self::Present(context) => context,
             Self::Missing => panic!("{}", MISSING_CONTEXT),
         }
     }
-}
 
-impl<C> AsRef<C> for Cell<C> {
-    fn as_ref(&self) -> &C {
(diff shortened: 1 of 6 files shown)


runtime/src/utils/cell.rs
```diff
@@ -63,26 +63,36 @@ impl<C> Cell<C> {
...
     }
 
-    /// Consume the slot, returning the context and panicking if it is missing.
-    pub fn into(self) -> C {
+    /// Returns a reference to the context.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the context is missing.
+    pub fn as_present(&self) -> &C {
```
patrick-ogrady, line 71: Is `as_present*` common lingo for this (just curious)?

clabby: `as_present` isn't, just using `as_<variant_name>`. Generally borrow -> borrow conversions are `as_` though: https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv

patrick-ogrady: yep, got it.

patrick-ogrady: tbh, forgot the inner enum used present lol

patrick-ogrady (approved):
nice

Issue #2425, opened 2025-12-08
[examples/reshare] Orchestrator cannot validate epoch boundary finalizations


andresilva:
The reshare orchestrator fetches epoch boundary finalizations from peers to trigger epoch transitions. When it receives a finalization, it can verify the certificate signature and confirm the epoch matches, but it cannot verify that the finalization is for the boundary block (the last block in the epoch).

This is because finalization certificates contain the view, not the height (views and heights differ when nullifications occur, a block at height 100 might be finalized in view 105). A malicious peer could send a valid finalization for a non-boundary block in the same epoch, and the orchestrator has no way to detect this.

Marshal solves this by always fetching `(Finalization, Block)` pairs, checking `block.height()` and that the finalization payload digest matches the block.

### Potential solutions

1. Include block height in the certificate - Add height to the signed portion of finalization certificates. Recipients can then verify height without needing the block. This is a breaking change to the wire format but makes certificates self-describing, while adding a bit of data overhead that is redundant in some cases (e.g. if you already have the block).

2. Expose marshal's height-based fetch - Marshal already solves this by fetching `(Finalization, Block)` pairs via `Request::Finalized { height }` and validating `block.height()`. The orchestrator could trigger marshal to perform this fetch on its behalf and retrieve the validated finalization once complete. This would also remove the need for the orchestrator to inject finalizations directly into marshal (which might be error-prone since they are expected to have been previously verified)

Related #2032 + #2380

andresilva:
One possibility is to pair the certificate with the block header, since the header contains the height. This approach is "perfect" in the sense that there's no redundant data baked into the certificate itself. However, it does constrain the block format somewhat by assuming the existence of a header (though in the worst case the header could just be the entire block). There's also a potential downside: even though the certificate itself wouldn't carry redundant data, we might actually end up transmitting more bytes overall. For example, to prove height we'd need to include the entire header rather than just the u64 we actually care about.

That said, consensus itself doesn't care about the header or any redundant data, it's purely there for external consumers of the certificates, which is an argument for keeping the certificate clean and not baking in data that only matters outside the consensus protocol.

There doesn't seem to be an obviously right choice here. We're balancing certificate usability against potential data duplication, and the best answer likely depends on the most common use cases.

andresilva:
Closed in #2582.

PR #2582, opened 2025-12-19
[consensus/marshal, examples/reshare] trigger fetch finalization

### LINKED ISSUE #2032: [examples/reshare] Send Outdated Finalization Directly to `Marshal`

andresilva:
Closes #2032 #2425. 

Replaces #2287.

### DIFF
--- a/examples/reshare/src/orchestrator/actor.rs
+++ b/examples/reshare/src/orchestrator/actor.rs
@@ -2,7 +2,7 @@
 
 use crate::{
     application::{Block, EpochProvider, Provider},
-    orchestrator::{Mailbox, Message},
+    orchestrator::{ingress::Message, Mailbox},
     BLOCKS_PER_EPOCH,
 };
 use commonware_consensus::{
--- a/examples/reshare/src/dkg/actor.rs
+++ b/examples/reshare/src/dkg/actor.rs
@@ -10,10 +10,7 @@ use crate::{
 };
 use bytes::{Buf, BufMut};
 use commonware_codec::{Encode, EncodeSize, Error as CodecError, Read, ReadExt, Write};
-use commonware_consensus::{
-    types::{Epoch, EpochPhase, Epocher, FixedEpocher},
-    Reporter,
-};
+use commonware_consensus::types::{Epoch, EpochPhase, Epocher, FixedEpocher};
 use commonware_cryptography::{
     bls12381::{
         dkg::{observe, DealerPrivMsg, DealerPubMsg, Info, Output, PlayerAck},
@@ -167,7 +164,7 @@ where
         mut self,
         output: Option<Output<V, C::PublicKey>>,
         share: Option<Share>,
-        orchestrator: impl Reporter<Activity = orchestrator::Message<V, C::PublicKey>>,
+        orchestrator: orchestrator::Mailbox<V, C::PublicKey>,
         dkg: (
             impl Sender<PublicKey = C::PublicKey>,
             impl Receiver<PublicKey = C::PublicKey>,
@@ -187,7 +184,7 @@ where
         mut self,
         output: Option<Output<V, C::PublicKey>>,
         share: Option<Share>,
-        mut orchestrator: impl Reporter<Activity = orchestrator::Message<V, C::PublicKey>>,
+        mut orchestrator: orchestrator::Mailbox<V, C::PublicKey>,
         (sender, receiver): (
             impl Sender<PublicKey = C::PublicKey>,
             impl Receiver<PublicKey = C::PublicKey>,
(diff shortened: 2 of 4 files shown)


consensus/src/marshal/ingress/mailbox.rs
```diff
@@ -219,6 +229,38 @@ impl<S: Scheme, B: Block> Mailbox<S, B> {
...
+    /// Fetches a [Finalization] by height, triggering a network fetch if not locally available.
+    ///
+    /// Unlike [get_finalization](Self::get_finalization) which only checks local storage,
+    /// this method will request the finalization from the network via the resolver if it
+    /// is not available locally. The returned finalization has been validated by marshal
+    /// to ensure the block height matches the requested height and the cryptographic
+    /// signature is valid.
+    ///
+    /// Returns `None` if the fetch fails or times out (caller should retry).
+    pub async fn fetch_finalization(
```
patrick-ogrady: I think we could just call this `enqueue_fetch` (and not wait for the result). You can already do that on subscribe, so doing here as well seems redundant?

patrick-ogrady: I think that should greatly minimize the diff

patrick-ogrady: or `register_simplification`

examples/reshare/src/orchestrator/actor.rs
```diff
@@ -346,6 +292,21 @@ where
                     }
                 }
             },
+            result = &mut pending_boundary_fetch => {
```
patrick-ogrady: seems we only use the result for logging ... trying to think of a case where listening on this is useful (rather than just enqueueing).

examples/reshare/src/orchestrator/actor.rs
```diff
@@ -223,81 +206,17 @@ where
...
-                        epoch.get(),
-                        Recipients::One(from),
-                        message.encode().freeze(),
-                        false,
-                    )
-                    .await.is_err() {
-                        warn!("failed to forward finalization, shutting down orchestrator");
-                        break;
-                    }
+                self.marshal.ensure_finalization(boundary_height).await;
```
patrick-ogrady: Very dumb question...we don't actually know if the finalization exists at this height because someone could just send us a vote in an arbitrary epoch?

consensus/src/marshal/ingress/mailbox.rs
```diff
@@ -84,6 +85,22 @@ pub(crate) enum Message<S: Scheme, B: Block> {
...
+    ///
+    /// Unlike [Self::GetFinalization] which only checks local storage, this triggers
+    /// a network fetch if the finalization is not available locally. This is
+    /// fire-and-forget, the finalization will be stored in marshal and delivered
+    /// via the normal finalization flow when available.
+    ///
+    /// Targets are required because this is typically called when a peer claims to
+    /// be ahead. By targeting only that peer, we can block them if they don't
+    /// actually have the finalization.
+    EnsureFinalization {
```
patrick-ogrady: I think the name of this is misleading...I think we could call it `UntrustedFinalization` or `Hint` or something?

consensus/src/marshal/actor.rs
```diff
@@ -430,6 +436,21 @@ where
...
+                            }
+
+                            // Skip if finalization is already available locally
+                            if self.get_finalization_by_height(height).await.is_some() {
+                                continue;
+                            }
+
+                            // Trigger a targeted fetch via the resolver
+                            let request = Request::<B>::Finalized { height };
+                            resolver.fetch_targeted(request, targets).await;
```
patrick-ogrady, line 452: If this request is already outstanding to "all", I don't think we want to override with targets.

Can we add a test to resolver where we permit targets to grow but not go from "all" to "targets"?

patrick-ogrady: (just ignore in that case)

resolver/src/lib.rs
```diff
@@ -38,12 +39,45 @@ pub trait Resolver: Clone + Send + 'static {
     /// Type used to uniquely identify data.
     type Key: Span;
 
+    /// Type used to identify peers for targeted fetches.
+    type PublicKey: PublicKey;
```
patrick-ogrady, line 43: My thinking here is that we'll just ignore stuff like this if we back the resolver with some API (still useful to know who told us) 👍

Issue #2429, opened 2025-12-08
[consensus/simplex + marshal] Add Single Node Test


patrick-ogrady:
It should be possible for a single node network to run correctly (pass all existing tests). For completeness, we should run this case.

At a minimum, this is a useful test to ensure a rollup/sequencer can expand to a multi-sequencer setup seamlessly (with ~no code changes).

PR #2583, opened 2025-12-19
[reshare] Single-validator, single-epoch test

clabby:
Closes: #2429 

## Overview

Adds a test for a single validator in `reshare` and fixes the batcher to eagerly construct certificates if `quorum == 1`.

### DIFF
--- a/consensus/src/simplex/actors/batcher/actor.rs
+++ b/consensus/src/simplex/actors/batcher/actor.rs
@@ -169,7 +169,7 @@ impl<
         let mut shutdown = self.context.stopped();
         loop {
             // Track which view was modified (if any) for certificate construction
-            let updated_view: Option<View>;
+            let updated_view;
 
             // Handle next message
             select! {
@@ -204,7 +204,7 @@ impl<
                             active.send(is_active).unwrap();
 
                             // Setting leader may enable batch verification
-                            updated_view = Some(current);
+                            updated_view = current;
                         }
                         Some(Message::Constructed(message)) => {
                             // If the view isn't interesting, we can skip
@@ -225,7 +225,7 @@ impl<
                                 .add_constructed(message)
                                 .await;
                             self.added.inc();
-                            updated_view = Some(view);
+                            updated_view = view;
                         }
                         None => {
                             break;
@@ -409,9 +409,13 @@ impl<
                         .await {
                             self.added.inc();
                         }
-                    updated_view = Some(view);
+                    updated_view = view;
                 },
             }
+            assert!(
+                updated_view != View::zero(),
+                "updated view must be greater than zero"
+            );
 
             // Forward leader's proposal to voter (if we're not the leader and haven't already)
             if let Some(round) = work.get_mut(&current) {
@@ -423,10 +427,7 @@ impl<
             }
 
             // Process the updated view (if any)
-            let Some(view) = updated_view else {
-                continue;
-            };
-            let Some(round) = work.get_mut(&view) else {
+            let Some(round) = work.get_mut(&updated_view) else {
                 continue;
             };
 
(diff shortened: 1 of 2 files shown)


consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -439,65 +439,71 @@ impl<
                     selected = Some((view, voters, failed));
                 }
             }
-            let Some((view, voters, failed)) = selected else {
+
+            // Process batch verification results (if any)
+            if let Some((view, voters, failed)) = selected {
+                timer.observe();
+
+                let batch = voters.len() + failed.len();
```
patrick-ogrady, line 451: retain comment here

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -439,65 +439,71 @@ impl<
...
-            }
-            if let Some(finalization) = self
-                .recover_latency
-                .time_some(|| round.try_construct_finalization(&self.scheme))
-            {
-                debug!(%view, "constructed finalization, forwarding to voter");
-                voter
-                    .recovered(Certificate::Finalization(finalization))
-                    .await;
+            // Try to construct and forward certificates for the current view.
```
patrick-ogrady: nit:
```
            // Try to construct and forward certificates for the current view.
            // This runs regardless of whether batch verification happened above,
            // because the last vote processed may have come from us.
```

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -471,18 +475,25 @@ impl<
...
+                None
+            };
+
+            // Try to construct and forward certificates.
+            // We always try for the current view (because the last vote processed
+            // may have come from us), and also for the batch-verified view if different.
+            for view in [Some(current), verified_view]
+                .into_iter()
+                .flatten()
+                .collect::<BTreeSet<_>>()
```
patrick-ogrady: feels like overkill to use a BTreeSet here

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -439,65 +442,80 @@ impl<
...
-                    .recovered(Certificate::Nullification(nullification))
-                    .await;
-            }
-            if let Some(finalization) = self
-                .recover_latency
-                .time_some(|| round.try_construct_finalization(&self.scheme))
+            // Try to construct and forward certificates.
+            // We always try for the current view (because the last vote processed
+            // may have come from us), and also for the batch-verified view if different.
+            for view in [Some(current), verified_view]
```
patrick-ogrady: we should dedup these two (they may be the same)?

I suppose we'll just fallthrough so maybe a comment is all that is required.

clabby: We use the `BTreeSet` + flatten to deduplicate

clabby: Updated to just use a `match`

consensus/src/simplex/actors/batcher/actor.rs
clabby: https://github.com/commonwarexyz/monorepo/pull/2583#discussion_r2636170861

consensus/src/simplex/actors/batcher/actor.rs
patrick-ogrady: dumb question: should we be doing this walkback regardless of whether selected is none?

I think we could probably simplify this code a good bit by using the "view" approach taken by `votor` (where we just act on some view that was updated ... because that's the only time a condition may have changed)

patrick-ogrady: I don't think we'll delay processing of finalization here because we only process one vote at a time (and thus a notarize can't prevent finalize)

examples/reshare/src/validator.rs
```diff
@@ -863,6 +863,46 @@ mod test {
...
+        Plan {
+            seed: 0,
+            total: 1,
+            per_round: vec![1],
+            link: Link {
+                latency: Duration::from_millis(10),
+                jitter: Duration::from_millis(1),
+                success_rate: 1.0,
+            },
+            mode: Mode::Reshare(1),
```
patrick-ogrady: can you make this 2 (just so we know it handles the epoch transition right)?

patrick-ogrady: (would need to change test name)

PR #2584, opened 2025-12-19
[consensus/simplex] Simplify batcher loop

clabby:
## Overview

Simplifies the batcher actor's loop by using a similar pattern to the voter, where we act on a single view per iteration (and only when the view was updated). ctx: https://github.com/commonwarexyz/monorepo/pull/2583#discussion_r2636179103

### DIFF
--- a/consensus/src/simplex/actors/batcher/actor.rs
+++ b/consensus/src/simplex/actors/batcher/actor.rs
@@ -168,6 +168,9 @@ impl<
         let mut work = BTreeMap::new();
         let mut shutdown = self.context.stopped();
         loop {
+            // Track which view was modified (if any) for certificate construction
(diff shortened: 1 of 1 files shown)


consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -349,6 +356,9 @@ impl<
                                 .await;
                         }
                     }
+
+                    // Certificates are already forwarded to voter, no need for construction
+                    updated_view = None;
```
patrick-ogrady: I think in this case you could just continue?

patrick-ogrady: And then you wouldn't need an option?

PR #2585, opened 2025-12-19
[examples/reshare] provide certificate verifier

andresilva:


examples/reshare/src/application/scheme.rs
```diff
@@ -61,6 +67,10 @@ impl<S: Scheme, C: Signer> certificate::Provider for Provider<S, C> {
         let schemes = self.schemes.lock().unwrap();
         schemes.get(&epoch).cloned()
     }
+
+    fn all(&self) -> Option<Arc<S>> {
```
patrick-ogrady, line 70: Whoops...should've implemented this.

examples/reshare/src/application/scheme.rs
patrick-ogrady, line 65: Wonder if we should instead change this trait to be `Option<&S>`?

I suppose the benefit is that we can take it through async with Arc?

examples/reshare/src/application/scheme.rs
```diff
@@ -23,13 +26,16 @@ pub type EdScheme = simplex::scheme::ed25519::Scheme;
 #[derive(Clone)]
 pub struct Provider<S: Scheme, C: Signer> {
     schemes: Arc<Mutex<HashMap<Epoch, Arc<S>>>>,
+    /// An epoch-independent certificate verifier.
```
patrick-ogrady: tbh: I think we can remove this comment.

examples/reshare/src/engine.rs
```diff
@@ -218,7 +218,13 @@ where
...
         info!(elapsed = ?start.elapsed(), "restored finalized blocks archive");
 
-        let provider = Provider::new(config.signer.clone());
+        // Create the certificate verifier from the initial output if available.
+        // This allows epoch-independent certificate verification.
+        let certificate_verifier = config
+            .output
+            .as_ref()
+            .and_then(<Provider<S, C> as EpochProvider>::certificate_verifier);
+        let provider = Provider::new(config.signer.clone(), certificate_verifier);
```
patrick-ogrady, line 227: nit: put `\n` after `proivder`

patrick-ogrady (commented):
just nits

examples/reshare/src/application/scheme.rs
andresilva: Through a lock as well, although I guess in this case it's less likely that someone will need to put a lock on this since it should never change?

Issue #786, opened 2025-04-21
[runtume/stream] Consider using `tokio::io::BufReader` to wrap `tokio::net::TcpStream`


BrendanChou:
`tokio::io::BufReader` wraps an underlying asynchronous reader (like a `tokio::net::TcpStream` which implements `tokio::io::AsyncRead`). Note: do not use `std::io::BufReader` which is not intended for async operation.

When you try to read from the `BufReader` and its internal buffer doesn't contain enough data to satisfy the read, it performs a potentially larger read call on the underlying source to fill up its internal buffer (the size of this buffer is configurable, often defaulting to something like 8KB).

Subsequent read calls to the BufReader are served directly from this in-memory buffer without making further asynchronous calls or system calls to the underlying source, as long as the requested data is available in the buffer. Once the buffer is depleted, the next read triggers another larger read from the underlying source. This may reduce the overhead from requesting many small syscall requests to fetch data from the network.

Since `stream` performs at least two read operations (thefirst is reading the frame-length prefix), BufReader can satisfy these more efficiently. In the case that we modify the length prefix to be a varint (#753), this would be even more efficient.

patrick-ogrady:
The challenge here is that we can't necessarily wait for some buffer to fill to process messages (it is almost like we need an "async filler" that will optimistically fill as data comes in). If we did wait, it is very possible that an "efficient" application could see very poor performance (only filling the buffer after enough timeout retries had elapsed).

PR #2593, opened 2025-12-20
[runtime/network/tokio] add read buffering

commonware-llm:
Related: #786 

Add an internal read buffer to the tokio Stream implementation to reduce
syscall overhead. Multiple small recv() calls can now be satisfied from
the buffer without additional network operations.

Key changes:
- Add configurable read_buffer_size to Config (default 64KB)
- Buffer smaller reads to batch syscalls
- For large reads (>= buffer size), bypass buffer and read directly

### DIFF
--- a/runtime/src/network/tokio.rs
+++ b/runtime/src/network/tokio.rs
@@ -74,17 +74,26 @@ impl Stream {
         // Compact first to maximize space for reading
         self.compact();
 
-        // Read at least min_bytes more, up to buffer capacity
+        // Use a single deadline for the entire operation to prevent slow-drip attacks
+        let deadline = tokio::time::Instant::now() + self.read_timeout;
         let target = self.end + min_bytes;
+
+        // Read at least min_bytes more, up to buffer capacity
         while self.end < target {
+            // Compute the remaining time and check if we've timed out
+            let remaining_time = deadline.saturating_duration_since(tokio::time::Instant::now());
+            if remaining_time.is_zero() {
+                return Err(Error::Timeout);
+            }
+
+            // Read up to the remaining time
             let bytes_read = timeout(
-                self.read_timeout,
+                remaining_time,
                 self.stream.read(&mut self.buffer[self.end..]),
             )
             .await
             .map_err(|_| Error::Timeout)?
             .map_err(|_| Error::RecvFailed)?;
-
             if bytes_read == 0 {
                 return Err(Error::RecvFailed); // EOF
             }


runtime/src/network/tokio.rs
```diff
@@ -73,16 +171,19 @@ impl crate::Listener for Listener {
         }
 
         // Return the sink and stream
-        let (stream, sink) = stream.into_split();
+        let (read_half, write_half) = stream.into_split();
```
patrick-ogrady: nit: stream/sink

runtime/src/network/tokio.rs
```diff
@@ -152,6 +270,7 @@ impl Default for Config {
             tcp_nodelay: None,
             read_timeout: Duration::from_secs(60),
             write_timeout: Duration::from_secs(30),
+            read_buffer_size: DEFAULT_READ_BUFFER_SIZE,
```
patrick-ogrady: nit: just define inline

runtime/src/network/tokio.rs
```diff
@@ -204,25 +323,31 @@ impl crate::Network for Network {
         }
 
         // Return the sink and stream
-        let (stream, sink) = stream.into_split();
+        let (read_half, write_half) = stream.into_split();
```
patrick-ogrady: nit: stream/sink

runtime/src/network/tokio.rs
```diff
@@ -29,23 +29,121 @@ impl crate::Sink for Sink {
...
+    /// Returns the total number of bytes read, or an error.
+    async fn fill_buffer(&mut self, min_bytes: usize) -> Result<usize, Error> {
+        if min_bytes == 0 {
+            return Ok(0);
+        }
+
+        // Compact if we need space
+        self.compact();
+
+        let mut total_read = 0;
```
patrick-ogrady: nit: move to just above loop

runtime/src/network/tokio.rs
```diff
@@ -29,23 +29,121 @@ impl crate::Sink for Sink {
 }
 
 /// Implementation of [crate::Stream] for the [tokio] runtime.
+///
+/// This stream uses an internal read buffer to reduce syscall overhead.
```
patrick-ogrady: nit: `# Buffering`

runtime/src/network/tokio.rs
```diff
@@ -29,23 +29,120 @@ impl crate::Sink for Sink {
...
+    /// Returns the total number of bytes read, or an error.
+    async fn fill_buffer(&mut self, min_bytes: usize) -> Result<usize, Error> {
+        // Compact first to maximize space for reading
+        self.compact();
+
+        // Read at least min_bytes more, up to buffer capacity
+        let target = self.end + min_bytes;
+        while self.end < target {
+            let bytes_read = timeout(
+                self.read_timeout,
```
patrick-ogrady: We want to use the same read_timeout to cover the entire fill_buffer/read operation?

This approach would allow some peer to slowly drip bytes to us?

runtime/src/network/tokio.rs
```diff
@@ -29,23 +29,129 @@ impl crate::Sink for Sink {
...
-        if buf.is_empty() {
+        let needed = buf.len();
+        if needed == 0 {
             return Ok(buf);
         }
+        let mut filled = 0;
 
-        // Time out if we take too long to read
-        timeout(self.read_timeout, self.stream.read_exact(buf.as_mut()))
+        // First, drain any buffered data
```
patrick-ogrady, line 127: Tokio has some built-ins but AFAICT nothing quite right

runtime/src/network/tokio.rs
```diff
@@ -29,23 +29,129 @@ impl crate::Sink for Sink {
...
+        self.compact();
+
+        // Use a single deadline for the entire operation to prevent slow-drip attacks
+        let deadline = tokio::time::Instant::now() + self.read_timeout;
+        let target = self.end + min_bytes;
+
+        // Read at least min_bytes more, up to buffer capacity
+        while self.end < target {
+            // Compute the remaining time and check if we've timed out
+            let remaining_time = deadline.saturating_duration_since(tokio::time::Instant::now());
```
patrick-ogrady, line 84: This is the only real ~gotcha on this PR! We don't want to let an adversary force us to sit in this loop indefinitely by feeding us 1 byte at a time.

runtime/src/network/tokio.rs
```diff
@@ -29,23 +29,129 @@ impl crate::Sink for Sink {
...
+    async fn fill_buffer(&mut self, min_bytes: usize) -> Result<usize, Error> {
+        // Compact first to maximize space for reading
+        self.compact();
+
+        // Use a single deadline for the entire operation to prevent slow-drip attacks
+        let deadline = tokio::time::Instant::now() + self.read_timeout;
+        let target = self.end + min_bytes;
+
+        // Read at least min_bytes more, up to buffer capacity
+        while self.end < target {
```
roberto-bayardo, line 82: What's preventing me from trying to fill_buffer with min_bytes >> self.buffer.len() ?  Because it looks like you'll infinite loop in that event (since `self.end` will always be less than `target`)?

roberto-bayardo: (If this should never happen, perhaps document the invariant with an assertion check on the min_bytes input)

runtime/src/network/tokio.rs
```diff
@@ -29,23 +29,129 @@ impl crate::Sink for Sink {
...
+/// # Buffering
+///
+/// This stream uses a read buffer to reduce syscall overhead.
+/// Multiple small reads can be satisfied from the buffer without
+/// additional network operations.
 pub struct Stream {
     read_timeout: Duration,
     stream: OwnedReadHalf,
+    /// Internal buffer for batching reads.
+    buffer: Vec<u8>,
```
clabby, line 42: We should use `BytesMut` here along with [`OwnedReadHalf::read_buf`](https://docs.rs/tokio/latest/tokio/io/trait.AsyncReadExt.html#method.read_buf), to align with #2558. Would also get rid of the need to manually track `start/end`.

runtime/src/network/tokio.rs
```diff
@@ -29,23 +29,129 @@ impl crate::Sink for Sink {
...
+    /// Moves any remaining data to the front of the buffer.
+    #[inline]
+    fn compact(&mut self) {
+        if self.start == 0 {
+            return;
+        }
+
+        let remaining = self.end - self.start;
+        if remaining > 0 {
+            self.buffer.copy_within(self.start..self.end, 0);
```
patrick-ogrady, line 65: I'm not sure we ever need to actually do this. Whenever we need to fill the buffer, there is nothing of interest left (should I think be good to just start at 0 in buffered read).

runtime/src/network/tokio.rs
patrick-ogrady: That's a case we check before starting the buffered read.

runtime/src/network/tokio.rs
clabby: https://github.com/commonwarexyz/monorepo/pull/2595 may be a better soln

runtime/src/network/tokio.rs
roberto-bayardo: Yeah I see where min_bytes is computed on L138 but feels like you should still assert it's not too big within this function, in case another use cases pops up and someone gets it wrong...

patrick-ogrady:
Replaced by #2595

PR #2595, opened 2025-12-20
[runtime/network/tokio] Use `BufReader` for read buffering

clabby:
Closes: #786 

## Overview

Alternative for #2593, but using `tokio`'s `BufReader` rather than implementing our own read buffering.

### DIFF
--- a/runtime/src/network/tokio.rs
+++ b/runtime/src/network/tokio.rs
@@ -244,7 +244,6 @@ mod tests {
     };
     use commonware_macros::test_group;
     use std::time::{Duration, Instant};
-    use tokio::sync::oneshot;
 
     #[tokio::test]
     async fn test_trait() {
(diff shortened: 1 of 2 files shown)


runtime/src/network/tokio.rs
```diff
@@ -248,4 +269,85 @@ mod tests {
...
+        sink.send(vec![1u8, 2, 3, 4, 5]).await.unwrap();
+
+        // Wait for the reader to complete
+        let (result, elapsed) = reader.await.unwrap();
+        assert!(matches!(result, Err(crate::Error::Timeout)));
+
+        // Verify the timeout occurred around the expected time
+        assert!(elapsed >= read_timeout);
+        // Allow some margin for timing variance
+        assert!(elapsed < read_timeout + Duration::from_millis(10));
```
patrick-ogrady: Let's make the margin read_timeout * 2

runtime/src/network/tokio.rs
```diff
@@ -204,25 +222,28 @@ impl crate::Network for Network {
         }
 
         // Return the sink and stream
-        let (stream, sink) = stream.into_split();
+        let (read_half, sink) = stream.into_split();
```
patrick-ogrady: Nit: keep it "stream"

PR #2600, opened 2025-12-22
fix(broadcast): use correct metric for dropped subscribe waiters

eeemmmmmm:
The cleanup_waiters function was incrementing the `get` metric for dropped waiters, but waiters are only created by `subscribe` requests (handle_get responds immediately without creating waiters). Fixed to use `subscribe` metric instead.


clabby:

Issue #2597, opened 2025-12-20
[runtime/network/iouring] Add Support for Buffered Reading


patrick-ogrady:
Perform a similar optimization as #786 for iouring.

PR #2601, opened 2025-12-22
[runtime/network/iouring] buffered reading

andresilva:
Closes https://github.com/commonwarexyz/monorepo/issues/2597.

### DIFF
--- a/runtime/src/network/iouring.rs
+++ b/runtime/src/network/iouring.rs
@@ -423,10 +423,10 @@ impl crate::Stream for Stream {
 
             let remaining = buf_len - bytes_received;
 
-            // Buffer is empty, read from socket
-            // For large reads or unbuffered mode, read directly into caller's buffer
-            let buffer_capacity = self.buffer.capacity();
-            if buffer_capacity == 0 || remaining >= buffer_capacity {
+            // Skip internal buffer if disabled, or if the read is large enough
+            // to fill the buffer and immediately drain it
+            let buffer_len = self.buffer.len();
+            if buffer_len == 0 || remaining >= buffer_len {
                 let (returned_buf, result) = self.submit_recv(buf, bytes_received, remaining).await;
                 buf = returned_buf;
                 bytes_received += result?;


runtime/src/network/iouring.rs
```diff
@@ -282,69 +298,144 @@ impl crate::Sink for Sink {
...
 /// Implementation of [crate::Stream] for an io-uring [Network].
+///
+/// Uses an internal buffer to reduce syscall overhead. Multiple small reads
+/// can be satisfied from the buffer without additional network operations.
 pub struct Stream {
     fd: Arc<OwnedFd>,
     /// Used to submit recv operations to the io_uring event loop.
     submitter: mpsc::Sender<iouring::Op>,
+    /// Internal read buffer.
+    buffer: Vec<u8>,
```
andresilva, line 309: `BytesMut` wasn't used because its methods for consuming data (`split_to()`, `advance()`) reduce the effective capacity by moving internal pointers rather than resetting them. When we read data from the buffer and advance past it, the capacity shrinks permanently. Calling `clear()` resets the length but not the start pointer, so the usable capacity keeps decreasing.

runtime/src/network/iouring.rs
```diff
@@ -282,69 +298,144 @@ impl crate::Sink for Sink {
...
-            let (result, got_buf) = rx.await.map_err(|_| crate::Error::RecvFailed)?;
-            buf = got_buf.unwrap();
             if should_retry(result) {
                 continue;
             }
 
-            // Non-positive result indicates an error or EOF.
             if result <= 0 {
-                return Err(crate::Error::RecvFailed);
+                let err = if result == -libc::ETIMEDOUT {
```
patrick-ogrady, line 377: nice ❤️

runtime/src/network/iouring.rs
```diff
@@ -282,69 +298,144 @@ impl crate::Sink for Sink {
...
+                self.buffer_pos += to_copy;
+                bytes_received += to_copy;
+                continue;
+            }
+
+            let remaining = buf_len - bytes_received;
+
+            // Buffer is empty, read from socket
+            // For large reads or unbuffered mode, read directly into caller's buffer
+            let buffer_capacity = self.buffer.capacity();
```
patrick-ogrady: Is there a particular reason to use `.capacity()` here instead of `.len()` (like we do in `fill_buffer`).

It shouldn't matter for correctness but feels more correct to align?

runtime/src/network/iouring.rs
```diff
@@ -282,69 +298,144 @@ impl crate::Sink for Sink {
...
+            if buffer_capacity == 0 || remaining >= buffer_capacity {
```
patrick-ogrady: nit: add a small comment indicating this is just handling the empty buffer case?

runtime/src/network/iouring.rs
```diff
@@ -39,13 +39,31 @@ use std::{
...
     /// If Some, explicitly sets TCP_NODELAY on the socket.
     /// Otherwise uses system default.
     pub tcp_nodelay: Option<bool>,
     /// Configuration for the iouring instance.
     pub iouring_config: iouring::Config,
+    /// Size of the read buffer for batching network reads.
+    ///
+    /// A larger buffer reduces syscall overhead by reading more data per call,
+    /// but uses more memory per connection. Defaults to 64 KB.
+    pub read_buffer_size: usize,
```
patrick-ogrady, line 56: I suppose we don't have a test ensuring the tokio receiver can have a read_buffer_size of 0? Can you add a test for it here and in tokio?

runtime/src/network/iouring.rs
andresilva: It's also handling the case where we are doing a read larger than the rear buffer, in which case we read directly into the caller buffer, to avoid filling the read buffer only to drain it after.

runtime/src/network/iouring.rs
andresilva: I added a test here for this (test_unbuffered_mode), will replicate for tokio.

andresilva:
@patrick-ogrady Addressed!

PR #2602, opened 2025-12-22
[Storage] Implement state sync for `qmdb::any::unordered::variable`

danlaine:
* Moves `qmdb/any/unordered/{fixed, variable}.rs` to `storage/src/qmdb/any/unordered/{fixed, variable}/mod.rs`
* Moves `qmdb/any/unordered/sync.rs` to `storage/src/qmdb/any/unordered/fixed/sync.rs`
* Implement new `storage/src/qmdb/any/unordered/variable/sync.rs`
* Factor out shared tests into `qmdb/any/unordered/sync_tests.rs`

### DIFF
--- a/storage/src/qmdb/any/unordered/sync_tests.rs
+++ b/storage/src/qmdb/any/unordered/sync_tests.rs
@@ -5,7 +5,7 @@
 
 use crate::{
     journal::contiguous::Contiguous,
-    mmr::Location,
+    mmr::{Location, Position},
     qmdb::{
         self,
         any::CleanAny,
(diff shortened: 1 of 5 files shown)


storage/src/qmdb/any/unordered/sync_tests.rs
```diff
@@ -0,0 +1,1237 @@
...
+
+/// Type alias for the config type of a harness.
+pub(crate) type ConfigOf<H> = <DbOf<H> as qmdb::sync::Database>::Config;
+
+/// Type alias for the journal type of a harness.
+pub(crate) type JournalOf<H> = <DbOf<H> as qmdb::sync::Database>::Journal;
+
+/// Trait for cleanup operations in tests.
+pub(crate) trait Closeable {
+    fn close(self) -> impl std::future::Future<Output = Result<(), qmdb::Error>> + Send;
```
roberto-bayardo, line 44: As we were discussing yesterday, do we need close()?

roberto-bayardo:
lgtm, is it worth adding this to examples/ ?

storage/src/qmdb/any/unordered/sync_tests.rs
danlaine: I'll make a separate PR to remove `close`

roberto-bayardo (approved):

danlaine:
> lgtm, is it worth adding this to examples/ ?

We already support running the sync example with `qmdb::any::unordered::fixed`. I'm not sure adding `qmdb::any::unordered::variable` would add much to the example because the code would be pretty much identical to the existing implementation, with `variable` substituted for `fixed`.

Issue #1841, opened 2025-10-08
[workspace] Upgrade to 2024 edition


clabby:
## Overview

The 2024 edition of Rust was released [along with `rustc` `1.85.0` earlier this year](https://blog.rust-lang.org/2025/02/20/Rust-1.85.0/). A few nice-to-have language features are gated behind the new edition such as:
- if-let chains
- async closures
- temporary if-let scope
- Improved taildrop semantics

It would be nice to update the repository to use this edition when time permits.

PR #2604, opened 2025-12-23
[consensus/marshaled] Deferred verification

clabby:
## Overview

Adds support for deferred verification in `Marshaled`.

### Before

```mermaid
sequenceDiagram
     participant Validator
     participant DA as Data Availability
     participant Verify as Verification Routine
     participant Network
     Validator->>DA: Wait for block
     DA-->>Validator: Block available
     Validator->>Verify: Run verification
     Verify-->>Validator: Verification result
     Validator->>Network: Cast notarize vote
```

### After

```mermaid
sequenceDiagram
     participant Validator
     participant DA as Data Availability
     participant Verify as Verification Routine
     participant Network
     participant Certify
     Validator->>DA: Wait for block
     DA-->>Validator: Block available
     Validator->>Network: Optimistic notarize vote
     par Vote in flight
         Network-->>Validator: Vote delivered
     and Verify block
         Validator->>Verify: Run verification
         Verify-->>Validator: Verification result
     end
     Validator->>Certify: Certify block is correct
     Certify-->>Validator: Ready to finalize
     Validator->>Network: Cast finalize vote
```

### DIFF
--- a/consensus/src/application/marshaled.rs
+++ b/consensus/src/application/marshaled.rs
@@ -157,8 +157,9 @@ where
     /// Verifies a proposed block's application-level validity.
     ///
     /// This method validates that:
-    /// 1. The block's height is exactly one greater than the parent's height
-    /// 2. The underlying application's verification logic passes
+    /// 1. The block's parent commitment matches the expected parent
+    /// 2. The block's height is exactly one greater than the parent's height
+    /// 3. The underlying application's verification logic passes
     ///
     /// Verification is spawned in a background task and returns a receiver that will contain
     /// the verification result. Valid blocks are reported to the marshal as verified.
@@ -208,10 +209,17 @@ where
                     }
                 };
 
-                // Validate that heights are contiguous.
-                //
-                // When called via Automaton::verify, parent commitment was already validated
-                // against the context's parent.
+                // Validate parent commitment and height contiguity.
+                if block.parent() != parent.commitment() || parent.commitment() != parent_commitment
+                {
+                    debug!(
+                        block_parent = %block.parent(),
+                        expected_parent = %parent.commitment(),
+                        "block parent commitment does not match expected parent"
+                    );
+                    tx.send_lossy(false);
+                    return;
+                }
                 if parent.height().next() != block.height() {
                     debug!(
                         parent_height = %parent.height(),
@@ -499,7 +507,7 @@ where
                         return;
                     }
 
-                    // Valid re-proposal. Create a completed verification task for certify().
+                    // Valid re-proposal. Create a completed verification task for `certify`
                     let round = context.round;
                     marshal.verified(round, block).await;
 
@@ -512,17 +520,6 @@ where
                         .insert((round, commitment), task_rx);
 
                     tx.send_lossy(true);
-                    return;
-                }
-
-                // If the block is not a re-proposal, its parent must match the context's parent.
-                if block.parent() != context.parent.1 {
-                    debug!(
-                        block_parent = %block.parent(),
-                        expected_parent = %context.parent.1,
-                        "block parent commitment does not match expected parent"
-                    );
-                    tx.send_lossy(false);
                     return;
                 }
 
@@ -627,8 +624,9 @@ where
 
                 // Re-proposal detection for certify path: we don't have the consensus context,
                 // only the block's embedded context from original proposal. Infer re-proposal from:
-                // 1. Block is at epoch boundary
-                // 2. Certification round is later than block's original round
+                // 1. Block is at epoch boundary (only boundary blocks can be re-proposed)
+                // 2. Certification round's view > embedded context's view (re-proposals retain their
+                //    original embedded context, so a later view indicates the block was re-proposed)
                 // 3. Same epoch (re-proposals don't cross epoch boundaries)
                 let embedded_context = block.context();
                 let is_reproposal =
@@ -636,6 +634,8 @@ where
                         && round.view() > embedded_context.round.view()
                         && round.epoch() == embedded_context.round.epoch();
                 if is_reproposal {
+                    // NOTE: It is possible that, during crash recovery, we call `marshal.verified`
+                    // twice for the same block. That function is idempotent, so this is safe.
                     marshaled.marshal.verified(round, block).await;
                     tx.send_lossy(true);
                     return;
--- a/consensus/src/marshal/mod.rs
+++ b/consensus/src/marshal/mod.rs
@@ -2469,7 +2469,7 @@ mod tests {
     /// A re-proposal occurs when the parent commitment equals the block being verified,
     /// meaning the same block is being proposed again in a new view.
     #[test_traced("INFO")]
-    fn test_reproposal_validation() {
+    fn test_marshaled_reproposal_validation() {
         #[derive(Clone)]
         struct MockVerifyingApp {
             genesis: B,


consensus/src/application/marshaled.rs
```diff
@@ -87,42 +104,255 @@ where
...
+                // Helper to clean up verification context for this round
+                let cleanup = || async {
+                    let mut contexts_guard = verification_contexts.lock().await;
+                    if let Some(map) = contexts_guard.get_mut(&digest) {
+                        map.remove(&context.round);
+
+                        // Retain context maps for digests that relate to future rounds only.
+                        contexts_guard.retain(|_, map| {
+                            map.iter().any(|(m_round, _)| m_round > &context.round)
+                        });
```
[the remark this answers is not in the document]
clabby: This is the correct behavior. The voter should never attempt to go back and certify previous rounds, if I understand correctly.

consensus/src/application/marshaled.rs
```diff
@@ -2,7 +2,7 @@
 //!
 //! # Overview
 //!
-//! [Marshaled] is an adapter that wraps any [VerifyingApplication] implementation to handle
+//! [`Marshaled`] is an adapter that wraps any [`VerifyingApplication`] implementation to handle
```
roberto-bayardo, line 5: I've been omitting the backticks within square brackets. Is there a reason we should be including them?

clabby: It's part of the stdlib style guide / [`RFC-1574`](https://github.com/rust-lang/rfcs/blob/master/text/1574-more-api-documentation-conventions.md#link-all-the-things).

Not the end of the world to not include them, though, mostly a preference.

roberto-bayardo: Asking since I prefer consistency.  If this is the "preferred" style I am happy to change my ways.

clabby: I prefer following the standard, unless the standard fails the Patrick lint 😄

consensus/src/lib.rs
```diff
@@ -114,14 +115,17 @@ cfg_if::cfg_if! {
         pub trait CertifiableAutomaton: Automaton {
             /// Determine whether a verified payload is safe to commit.
             ///
-            /// If context is required during certify, it must be included in the
-            /// data associated with the payload.
+            /// The round identifies which consensus round is being certified. This is necessary
```
patrick-ogrady: As noted elsewhere, this is really only true if the payload doesn't incorporate round (which would be highly recommended).

In cases where we just "bolt on" consensus, however, this is not guaranteed. Or at the end of an epoch (where we re-propose a notarized block).

examples/reshare/src/engine.rs
```diff
@@ -278,12 +278,14 @@ where
...
         .await;
 
-        let application = Marshaled::new(
+        let application = Marshaled::init(
             context.with_label("application"),
             Application::new(dkg_mailbox.clone()),
             marshal_mailbox.clone(),
             FixedEpocher::new(BLOCKS_PER_EPOCH),
-        );
+            config.partition_prefix.clone(),
```
patrick-ogrady: I think we'd want to do `{}_marshaled` (like we do below for orchestrator)?

consensus/src/marshal/mod.rs
```diff
@@ -2280,6 +2304,161 @@ mod tests {
         })
     }
 
+    /// Regression test for cleanup bug in #2604.
```
patrick-ogrady: We should remove PR specifics from this imo...just talk about the case in plain english

consensus/src/application/marshaled.rs
patrick-ogrady: We should be following the standard (almost added a lint for this but it changed almost every file...probably worth adding over the weekend or something)

consensus/src/application/marshaled.rs
```diff
@@ -13,18 +13,27 @@
 //! Instead, it re-proposes the boundary block to avoid producing blocks that would be pruned
 //! by the epoch transition.
 //!
+//! # Asynchronous Verification
+//!
+//! Using [`CertifiableAutomaton`], it is possible to perform block verification asynchronously
```
patrick-ogrady: nit: prefer Deferred Verification (it is already async in that it doesn't block consensus) 

```
# Deferred Verification

//! By default, [`Marshaled`] performs deferred verification, optimistically voting in favor
//! of the proposed block but waiting to finalize until the verification is complete. In
//! [XXX] verification is started and in [`CertifiableAutomaton::certify`] we await
//! the result (before voting to finalize).

Comment (patrick-ogrady) [reply consensus/src/application/marshaled.rs]: (also just want to confirm this is what we do? it is what we should be doing)

File: consensus/src/application/marshaled.rs [hunk 7]
```diff
@@ -87,42 +104,229 @@ where
...
+    async fn store_verification_context(
+        contexts_lock: &Arc<Mutex<VerificationContexts<E, B, S>>>,
+        context: <Self as Automaton>::Context,
+        digest: B::Commitment,
+    ) {
+        contexts_lock
+            .lock()
+            .await
+            .upsert_sync(digest, |map| {
+                map.insert(context.round, context);
```
patrick-ogrady: We should just be able to store by round with value being digest + context?

(no need for inner map)

consensus/src/application/marshaled.rs
```diff
@@ -289,148 +508,22 @@ where
...
-                    }
-                };
-
-                if application_valid {
-                    marshal.verified(context.round, block).await;
-                }
-                tx.send_lossy(application_valid);
-            });
+        let (tx, rx) = oneshot::channel();
+        let _ = tx.send(true);
```
patrick-ogrady: can we use FailableExt here?

consensus/src/application/marshaled.rs
```diff
@@ -441,12 +534,34 @@ where
...
+        // We clone rather than remove so the context remains available for crash recovery.
+        // Contexts are cleaned up when finalization advances past them (see report()).
+        let contexts_guard = self.verification_contexts.lock().await;
+        let context = contexts_guard
+            .get(&payload)
+            .and_then(|map| map.get(&round).cloned());
+        drop(contexts_guard);
+
+        if let Some(context) = context {
+            self.verify(context, payload).await
```
patrick-ogrady: Ah, I think we should be doing this long before?

consensus/src/application/marshaled.rs
clabby: `Metadata` requires a key that implements `Span`, which `Round` does not

consensus/src/lib.rs
clabby: I made some updates here to clarify.

consensus/src/marshal/actor.rs
```diff
@@ -920,8 +928,10 @@ where
         }
 
         // Update metrics and send tip update to application
-        if height > self.tip {
-            application.report(Update::Tip(height, commitment)).await;
+        if let Some(round) = round.filter(|_| height > self.tip) {
```
BrendanChou, line 931: this is fine but I think `if let Some(round) = round && height > self.tip {` might be more readable? `filter` is nice and functional if there is a chain but this seems slightly more ergonomic since many types (e.g. Vec) have a `.filter()` function

consensus/src/application/marshaled.rs
```diff
@@ -87,42 +107,231 @@ where
     marshal: marshal::Mailbox<S, B>,
     epocher: ES,
     last_built: Arc<Mutex<Option<(Round, B)>>>,
+    verification_contexts: Arc<Mutex<VerificationContexts<E, B, S>>>,
+    verification_tasks: Arc<Mutex<HashMap<Round, oneshot::Receiver<bool>>>>,
```
BrendanChou: We rarely use `Arc<Mutex<T>>` in our actors so this stood out to me. Probably a good reason we use it here but just want to make sure there's not a better way to avoid locks

BrendanChou: Perhaps `utils::futures::Pool` is the right thing to use instead of `Arc<Mutex<HashMap<Round, oneshot::Receiver<bool>>>>` but not sure

consensus/src/application/marshaled.rs
```diff
@@ -441,12 +545,43 @@ where
...
+            let contexts_guard = self.verification_contexts.lock().await;
+            let context = contexts_guard
+                .get(&payload)
+                .and_then(|map| map.get(&round).cloned());
+            drop(contexts_guard);
+
+            if let Some(context) = context {
+                self.verify(context, payload).await
+            } else {
+                // Verify is always called before certify for a given proposal, so if we
```
BrendanChou: It's been a while for me but is this true? If we come back online and see a notarization (that we hadn't contributed to by notarizing), then we don't attempt to certify it?

consensus/src/application/marshaled.rs
```diff
@@ -441,12 +545,43 @@ where
...
+                // don't have a verification context here, it means this proposal was never
+                // verified. Return a receiver that never resolves to signal to consensus
+                // that we should time out the view and vote to nullify.
+                let (_tx, rx) = oneshot::channel();
+                rx
+            }
+        }
```
BrendanChou, line 579: I'm slightly confused about the three paths here:
```
if let Some(task) = task {
  // path 1... this was being verified, right?
  task
} else {
  ...
  if let Some(context) = context {
    // path 2 ... does this mean you somehow have the context but you weren't verifying already?
    self.verify(context, payload).await
  } else {
    // path 3 ... you don't have the context and you weren't verifying already
    let (_tx, rx) = oneshot::channel();
  }
}
```

Mostly I'm confused how you get to path 2

consensus/src/application/marshaled.rs
BrendanChou: Nah actually I think `futures::Pool` is only for when you're waiting on a group of tasks to finish and you want to await them as a group (if any of them finish, keep giving to me the result).

For your code, you want to pull out the exact task. We could consider adding something to `utils` if you think something similar will be useful in the future

consensus/src/application/marshaled.rs
clabby: You could reach path #2 if the node crashes between verification and certification. The verification tasks are in-memory, but the contexts are durable.

consensus/src/marshal/actor.rs
clabby: I agree with you, but unfortunately we don't have let-chains in the edition of Rust that the monorepo uses.

consensus/src/application/marshaled.rs
clabby: Good call, not necessarily true. Will update for clarity.

consensus/src/application/marshaled.rs
clabby: The primary reason we use `Arc<Mutex<T>>` here is to allow for marshaled to be cloned (with shared state), as well as to allow accessing this state in the tasks we spin off (i.e. `verify`.) There should be low (if any) lock contention, since only one instance of `Marshaled` will actually be driven by consensus.

consensus/src/marshal/actor.rs
clabby: relevant: #1841

consensus/src/application/marshaled.rs
```diff
@@ -441,12 +545,44 @@ where
...
+            if let Some(context) = context {
+                self.verify(context, payload).await
+            } else {
+                // Verify is always called before certify for a given proposal (if we are
+                // online and don't see the notarization certificate first), so if we don't
+                // have a verification context here, it means this proposal was never verified
+                // by us. Return a receiver that never resolves to signal to consensus that
+                // we should time out the view and vote to nullify.
+                let (_tx, rx) = oneshot::channel();
+                rx
```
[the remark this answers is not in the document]
clabby: You led me astray before 🙇 

Test for this: https://github.com/commonwarexyz/monorepo/blob/84ca844f363b2c9a388a349e39fb3b44443c1667/consensus/src/simplex/actors/voter/mod.rs#L3266-L3266

BrendanChou (approved):

patrick-ogrady:
I was pondering this PR some more and I'm now wondering if it is safe (as context isn't extracted from the block).

If a notarization is formed with f Byzantine and f+1 honest, the other f honest may not have seen anything during verify (and thus have no context). If no nullification is produced and no finalization is produced (the f Byzantine can just withhold their votes), I think the f honest will just get stuck?

That would mean all subsequent rounds would have only f+1 honest available (and the chain could deadlock).

consensus/src/marshal/actor.rs
```diff
@@ -920,8 +928,10 @@ where
...
+            application
+                .report(Update::Tip(height, commitment, round))
+                .await;
```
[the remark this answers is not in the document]
clabby: Yes, but we should have finalizations the vast majority of the time.

patrick-ogrady:
I thought a good bit about our conversation on “blind voting” (issue vote before we have data).

While it is an interesting idea (and something we may want to explore in the future under different trust assumptions), I think it is the wrong approach for this initial pass. The node will attempt to fetch data for any notarization it learns about (indefinitely). If notarization no longer implies availability, it would be trivial for an adversary to force us to burn our request quota on fetching impossible to receive data. I've tried to adhere to a principle that any fetch request we make to the network must be resolvable by at least f+1 honest nodes to bound the worst-case where we fill our requester with a bunch of junk requests (and starve good ones).

If you make the tweak to wait before vote in your PR, should be good to merge!

consensus/src/lib.rs
```diff
@@ -114,14 +133,20 @@ cfg_if::cfg_if! {
...
             ///
-            /// Applications that employ erasure coding can override this method to delay or prevent
-            /// finalization until they have reconstructed and validated the full block (e.g. after
-            /// receiving enough shards).
+            /// This is particularly useful for applications that employ erasure coding, which
+            /// can override this method to delay or prevent finalization until they have
+            /// reconstructed and validated the full block (e.g., after receiving enough shards).
             fn certify(
                 &mut self,
+                _round: Round,
```
patrick-ogrady, line 144: Now that we can extract context from the block, do we need to provide round here?

I think the block is the better place because of the re-proposal logic at the end of an epoch (may get weird that this doesn't match what is in the block).

consensus/src/marshal/mocks/block.rs
```diff
@@ -1,104 +1,136 @@
...
-pub struct Block<D: Digest> {
+pub struct Block<D: Digest, C> {
     /// The parent block's digest.
     pub parent: D,
 
     /// The height of the block in the blockchain.
     pub height: Height,
 
+    /// The consensus context that was used when this block was proposed.
+    pub context: C,
```
patrick-ogrady: imo we should put this first

consensus/src/marshal/mod.rs
```diff
@@ -148,12 +151,38 @@ mod tests {
...
     type K = PublicKey;
+    type Ctx = crate::simplex::types::Context<D, K>;
+    type B = Block<D, Ctx>;
     type V = MinPk;
     type S = bls12381_threshold_vrf::Scheme<K, V>;
     type P = ConstantProvider<S, Epoch>;
 
+    /// Default leader key for tests.
+    fn default_leader() -> K {
+        use commonware_cryptography::{ed25519::PrivateKey, Signer as _};
```
patrick-ogrady: don't do inline imports

consensus/src/lib.rs
clabby: We don't, I don't think (need to validate for the coding PR, but I think same thing applies.)

clabby: Actually, it's still nice to have the round. See the `Marshaled` impl of `CertifiableAutomaton` - we want to fetch the verification task by `(round, payload)` in the happy-path, and by having the round, we don't need to fetch the block from marshal.

For that reason I think we should keep it. Likely a minimal performance benefit, but easier to reason about (and supplying `round` to `certify` doesn't involve that much plumbing.)

consensus/src/application/marshaled.rs
```diff
@@ -278,151 +472,85 @@ where
...
-                    return;
-                }
-
-                // Blocks are invalid if they are not within the current epoch and they aren't
-                // a re-proposal of the boundary block.
-                let Some(block_bounds) = epocher.containing(block.height()) else {
+                // Check if this is a re-proposal (the block commitment matches the parent from
+                // the context). Re-proposals at epoch boundaries retain their original embedded
+                // context, so we skip the context check for them.
+                let is_reproposal = context.parent.1 == digest;
```
patrick-ogrady: Because this is in verify, we still need to check the height is the last in the epoch?

Otherwise, I think some proposer could trigger this ahead of time?

patrick-ogrady: Probably a good opportunity for a regression test if one doesn't already exist.

consensus/src/application/marshaled.rs
```diff
@@ -114,10 +154,164 @@ where
...
+                        return;
+                    }
+                };
+
+                // You can only re-propose the same block if it's the last height in the epoch.
+                if parent.commitment() == block.commitment() {
+                    let last_in_epoch = epocher
+                        .last(context.epoch())
+                        .expect("current epoch should exist");
+                    let is_valid = block.height() == last_in_epoch;
```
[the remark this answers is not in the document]
clabby: This was an already-existing bug, nice. (Though I don't think it'd actually trigger, since the consensus engine shouldn't be advancing the epoch without being re-initialized.)

consensus/src/marshal/actor.rs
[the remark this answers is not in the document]
clabby: This is fine.

consensus/src/application/marshaled.rs
clabby: Spoke about this offline, we need to keep it. Test added 👍

consensus/src/application/marshaled.rs
```diff
@@ -13,17 +13,27 @@
 //! Instead, it re-proposes the boundary block to avoid producing blocks that would be pruned
 //! by the epoch transition.
 //!
+//! # Deferred Verification
+//!
+//! [`Marshaled`] uses deferred verification. Before casting a notarize vote, it waits
+//! for the block to become available (DA) and verifies that the block's embedded context
```
patrick-ogrady: (see next)

consensus/src/application/marshaled.rs
```diff
@@ -32,29 +42,39 @@
 //! - Genesis blocks are handled specially: epoch 0 returns the application's genesis block,
 //!   while subsequent epochs use the last block of the previous epoch as genesis
 //! - Blocks are automatically verified to be within the current epoch
+//!
+//! # Future Work
+//!
+//! - It is possible to not wait for data availability during notarization by caching the consensus
```
patrick-ogrady: nit: rewrite for clarity

consensus/src/application/marshaled.rs
patrick-ogrady: "for the block data to become available (DA)

consensus/src/application/marshaled.rs
```diff
@@ -73,20 +93,35 @@ use tracing::{debug, warn};
...
+/// # Verification Context Recovery
+///
+/// With deferred verification, validators wait for DA and verify the context before voting.
+/// If a validator crashes after voting but before certification, they lose their in-memory
+/// verification task. When recovering, validators use the block-embedded context.
+///
+/// Blocks implement [`CertifiableBlock`] which embeds the consensus context. This embedded
+/// context is trustworthy because the notarizing quorum (which contains at least f+1 honest
+/// validators) verified that the block's context matched the consensus context before voting.
+/// This allows validators who crashed or never participated in notarization to use the
```
patrick-ogrady: nit: This allows validators that crashed

consensus/src/application/marshaled.rs
```diff
@@ -73,20 +93,35 @@ use tracing::{debug, warn};
 /// This means the entire ancestry chain back to genesis is transitively validated.
 ///
 /// Applications do not need to re-implement these checks in their own verification logic.
+///
+/// # Verification Context Recovery
```
patrick-ogrady: nit: Context Recovery (may need for certification too)

consensus/src/application/marshaled.rs
```diff
@@ -73,20 +93,35 @@ use tracing::{debug, warn};
...
+/// With deferred verification, validators wait for DA and verify the context before voting.
+/// If a validator crashes after voting but before certification, they lose their in-memory
+/// verification task. When recovering, validators use the block-embedded context.
+///
+/// Blocks implement [`CertifiableBlock`] which embeds the consensus context. This embedded
+/// context is trustworthy because the notarizing quorum (which contains at least f+1 honest
+/// validators) verified that the block's context matched the consensus context before voting.
+/// This allows validators who crashed or never participated in notarization to use the
+/// block-embedded context for finalization. If a Byzantine proposer had embedded a malicious
+/// context, the f+1 honest validators would have rejected the mismatch during notarization,
```
patrick-ogrady: nit: this is already sufficiently documented, drop the last setence

consensus/src/application/marshaled.rs
```diff
@@ -114,10 +154,169 @@ where
...
+    #[inline]
+    async fn verify(
+        &mut self,
+        context: <Self as Automaton>::Context,
+        block: B,
+    ) -> oneshot::Receiver<bool> {
+        let mut marshal = self.marshal.clone();
+        let mut application = self.application.clone();
+        let epocher = self.epocher.clone();
+
```
patrick-ogrady: nit: remove \n

consensus/src/application/marshaled.rs
```diff
@@ -114,10 +154,169 @@ where
...
+        let (mut tx, rx) = oneshot::channel();
+        self.context
+            .with_label("verify")
```
patrick-ogrady: nit: given we pass this into the application, we should probably use `.with_attribute()` to ensure that any metrics applied aren't duplicates?

May be overkill rn

consensus/src/marshal/mocks/block.rs
```diff
@@ -66,39 +87,50 @@ impl<D: Digest> Read for Block<D> {
...
 }
 
-impl<D: Digest> crate::Block for Block<D> {
+impl<D: Digest, C: Codec<Cfg = ()> + Clone + Send + Sync + 'static> crate::Block for Block<D, C> {
     fn parent(&self) -> Self::Commitment {
         self.parent
     }
 }
+
+impl<D: Digest, C: Codec<Cfg = ()> + Clone + Send + Sync + 'static> crate::CertifiableBlock
```
patrick-ogrady, line 128: is `'static` necessary here?

consensus/src/simplex/actors/voter/mod.rs
```diff
@@ -3294,4 +3294,212 @@ mod tests {
...
+    /// stuck validators escape via nullification.
+    ///
+    /// This test extends the previous scenario to show that:
+    /// 1. A stuck validator (view 3) cannot be rescued by notarizations from future views
+    /// 2. The only escape route is a finalization certificate (which requires Byzantine cooperation)
+    ///
+    /// Once the f+1 honest validators certify view 3 and advance to view 4,
+    /// they can only vote to nullify view 4 (their current view) without equivocating.
+    /// The `handle_timeout` function only votes to nullify `self.view` (current view).
+    fn future_notarization_does_not_rescue_stuck_validator<S, F>(mut fixture: F)
```
patrick-ogrady: nit: only_finalization_rescues_validator

examples/reshare/src/application/types.rs
```diff
@@ -22,6 +24,9 @@ where
     /// The current height.
     pub height: Height,
 
+    /// The consensus context when this block was proposed.
+    pub context: Context<H::Digest, C::PublicKey>,
```
patrick-ogrady: nit: like above, put this first

consensus/src/marshal/mod.rs
```diff
@@ -81,8 +84,8 @@ use commonware_utils::{acknowledgement::Exact, Acknowledgement};
 /// Finalized blocks are reported to the application in monotonically increasing order (no gaps permitted).
 #[derive(Clone, Debug)]
 pub enum Update<B: Block, A: Acknowledgement = Exact> {
-    /// A new finalized tip.
-    Tip(Height, B::Commitment),
+    /// A new finalized tip and the finalization round.
+    Tip(Height, B::Commitment, Round),
```
patrick-ogrady: nit: put `Round` first (consensus data, then application data)

consensus/src/lib.rs
```diff
@@ -49,6 +49,24 @@ pub trait Block: Heightable + Codec + Digestible + Committable + Send + Sync + '
     fn parent(&self) -> Self::Commitment;
 }
 
+/// CertifiableBlock extends [Block] with consensus context information.
+///
+/// This trait is required for blocks used with deferred verification in Simplex consensus.
```
patrick-ogrady: nit: optimistic verification

patrick-ogrady: nit: remove talk of `Simplex`...anything at this level (in this file) should be generic?

consensus/src/lib.rs
```diff
@@ -49,6 +49,24 @@ pub trait Block: Heightable + Codec + Digestible + Committable + Send + Sync + '
...
+/// It allows the verification context to be derived directly from the block when a validator
+/// needs to participate in certification but never verified the block locally.
+///
+/// This is important for liveness: when Byzantine validators withhold their finalization votes,
+/// honest validators who weren't in the original notarizing quorum may need to help complete
+/// finalization. Since the block was notarized by 2f+1 validators (at least f+1 honest), the
+/// context embedded in the block can be trusted.
+pub trait CertifiableBlock: Block {
```
patrick-ogrady, line 57: nit: could call this ContextableBlock too I suppose 🤔

consensus/src/application/marshaled.rs
```diff
@@ -278,151 +477,107 @@ where
...
-                        .expect("current epoch should exist");
-                    if block.height() == last_in_epoch {
-                        marshal.verified(context.round, block).await;
-                        tx.send_lossy(true);
-                    } else {
+                // Check if this is a re-proposal (the block commitment matches the parent from
+                // the context). Re-proposals are only valid at epoch boundaries.
+                let is_reproposal = context.parent.1 == digest;
+                if is_reproposal {
+                    // Use the block's epoch (not context's epoch) since the block being
```
patrick-ogrady: It feels like we have these checks in two places? Can we somehow dedup?

consensus/src/application/marshaled.rs
```diff
@@ -438,18 +593,77 @@ where
...
+    B: CertifiableBlock<Context = <A as Application<E>>::Context>,
     ES: Epocher,
 {
-    // Uses default certify implementation which always returns true
+    async fn certify(&mut self, round: Round, payload: Self::Digest) -> oneshot::Receiver<bool> {
+        // Attempt to retrieve the existing verification task for this (round, payload).
+        let mut tasks_guard = self.verification_tasks.lock().await;
+        let task = tasks_guard.remove(&(round, payload));
+        drop(tasks_guard);
+
```
patrick-ogrady: nit: remove `\n`

consensus/src/application/marshaled.rs
```diff
@@ -114,10 +154,169 @@ where
...
+                        debug!(
+                            reason = "consensus dropped receiver",
+                            "skipping verification"
+                        );
+                        return;
+                    }
+                };
+
+                // You can only re-propose the same block if it's the last height in its epoch.
+                if parent.commitment() == block.commitment() {
```
patrick-ogrady: I think we should centralize all these validity checks in the outer verify? We don't need to redo them here (may be called by certify but at that point f+1 honest must've already checked?)

consensus/src/application/marshaled.rs
```diff
@@ -438,18 +593,77 @@ where
...
+                        debug!(
+                            reason = "consensus dropped receiver",
+                            "skipping certification"
+                        );
+                        return;
+                    }
+                };
+
+                let context = block.context();
+                let verify_rx = marshaled.verify(context, block).await;
```
patrick-ogrady: note: this performs a bunch of the end-of-epoch checks and doesn't need to.

consensus/src/lib.rs
patrick-ogrady: I recently asked you to call this `Deferred Verification` but think we should call it `Optimistic Verification` or `Optimistic Voting` instead?

patrick-ogrady (commented):
very close now...just nits (but looks correct)

consensus/src/application/marshaled.rs
patrick-ogrady: upside of doing this now is that tracing will be glorious

consensus/src/application/marshaled.rs
patrick-ogrady: (I took this note for myself, so disregard)

consensus/src/marshal/mocks/block.rs
patrick-ogrady: (or any of these others you added)

consensus/src/application/marshaled.rs
clabby: glad to do now

consensus/src/marshal/mocks/block.rs
clabby: It is, unfortunately, since `Committable` is `'static`. We really need to do an audit of all of those shared traits.

consensus/src/application/marshaled.rs
clabby: Yeah, we can remove them. I'll leave a comment noting that a notarization couldn't have formed to certify to begin with.

consensus/src/lib.rs
clabby: I'm ambivalent - either works.

consensus/src/application/marshaled.rs
clabby: Took a stab at re-wording.. lmk.

PR #2608, opened 2025-12-23
[Storage] Remove `close` methods

danlaine:
Removes `close` methods in favor of `drop`ping for several database types. (Note some types still have a close `method` -- those will be removed in ~~subsequent PRs~~ #2613 .)

Using `drop` is more Rusty. All uses of `close` are replaced by either `sync` (which `close` called) or simply `drop`ping the value.

Also makes the `qmdb::any` state `sync` engine persist the underlying journal before returning the synced database to the caller.

### DIFF
--- a/storage/fuzz/fuzz_targets/fixed_journal_operations.rs
+++ b/storage/fuzz/fuzz_targets/fixed_journal_operations.rs
@@ -135,6 +135,9 @@ fn fuzz(input: FuzzInput) {
                 JournalOperation::Restart => {
                     drop(journal);
                     journal = Journal::init(context.clone(), cfg.clone()).await.unwrap();
+                    // Reset tracking variables to match recovered state
+                    journal_size = journal.size();
+                    oldest_retained_pos = journal.oldest_retained_pos().unwrap_or(0);
                 }
 
                 JournalOperation::Destroy => {
--- a/storage/fuzz/fuzz_targets/mmr_journaled.rs
+++ b/storage/fuzz/fuzz_targets/mmr_journaled.rs
@@ -439,6 +439,8 @@ fn fuzz(input: FuzzInput) {
                     )
                     .await
                     .unwrap();
+                    // Reset tracking variables to match recovered state
+                    leaves.clear();
                     historical_sizes.clear();
                     MmrState::Clean(new_mmr)
                 }


storage/src/journal/contiguous/fixed.rs
```diff
@@ -605,18 +605,6 @@ impl<E: Storage + Metrics, A: CodecFixed<Cfg = ()>> Journal<E, A> {
         Ok(())
     }
 
-    /// Syncs and closes all open sections.
-    pub async fn close(self) -> Result<(), Error> {
-        for (i, blob) in self.blobs.into_iter() {
```
danlaine, line 610: All non-tail blobs should be unmodified since their last write so I think this loop was not required to persist all journal data

storage/src/mmr/journaled.rs
```diff
@@ -1236,13 +1230,13 @@ mod tests {
             assert_eq!(mmr.size(), 498);
             assert_eq!(mmr.root(), root);
 
-            // Make sure closing it and re-opening it persists the recovered state.
-            mmr.close().await.unwrap();
+            // Make sure dropping it and re-opening it persists the recovered state.
+            drop(mmr);
```
danlaine, line 1234: No sync needed here, we haven't updated it

storage/src/mmr/journaled.rs
```diff
@@ -1236,13 +1230,13 @@ mod tests {
...
             let mmr = Mmr::init(context.clone(), &mut hasher, test_config())
                 .await
                 .unwrap();
             assert_eq!(mmr.size(), 498);
-            mmr.close().await.unwrap();
+            drop(mmr);
```
danlaine, line 1239: No sync needed here, we haven't updated it

storage/src/qmdb/any/ordered/fixed.rs
```diff
@@ -546,9 +547,9 @@ mod test {
             }
             db.commit(None).await.unwrap();
             let root = db.root();
-            db.close().await.unwrap();
 
             // Simulate a failed commit and test that the log replay doesn't leave behind old data.
+            drop(db);
```
danlaine, line 552: No sync needed here, last operation was commit

storage/src/qmdb/store/mod.rs
```diff
@@ -1093,8 +1087,8 @@ mod test {
             let db = create_test_store(context.with_label("store")).await;
             assert_eq!(db.op_count(), op_count);
 
-            // Close and reopen the store to ensure the final commit is preserved.
-            db.close().await.unwrap();
+            // Reopen the store to ensure the final commit is preserved.
+            drop(db);
```
danlaine, line 1098: db unmodified, no need to sync

storage/src/qmdb/benches/variable/init.rs
```diff
@@ -40,21 +40,18 @@ fn bench_variable_init(c: &mut Criterion) {
                             let mut db =
                                 gen_random_kv(db, elements, operations, COMMIT_FREQUENCY).await;
                             db.prune(db.inactivity_floor_loc()).await.unwrap();
-                            db.close().await.unwrap();
```
danlaine, line 43: I didn't think the call to close is a meaningful part of the benchmark so I removed it

storage/src/qmdb/benches/fixed/init.rs
```diff
@@ -129,51 +132,44 @@ fn bench_fixed_init(c: &mut Criterion) {
                                             .await
                                             .unwrap();
                                         assert_ne!(db.op_count(), 0);
-                                        db.close().await.unwrap();
```
danlaine, line 132: I didn't think the call to close is a meaningful part of the benchmark so I removed it

storage/src/qmdb/benches/fixed/init.rs
```diff
@@ -48,23 +48,20 @@ fn bench_fixed_init(c: &mut Criterion) {
                                 gen_random_kv(db, elements, operations, Some(COMMIT_FREQUENCY))
                                     .await;
                             db.prune(db.inactivity_floor_loc()).await.unwrap();
-                            db.close().await.unwrap();
```
danlaine, line 51: I didn't think the call to close is a meaningful part of the benchmark so I removed it

storage/src/journal/authenticated.rs
```diff
@@ -1267,9 +1252,10 @@ mod tests {
                 Location::new_unchecked(20),
                 "commit should be at location 20"
             );
-            journal.close().await.unwrap();
+            journal.sync().await.unwrap();
```
danlaine, line 1258: This is `test_close_with_pending_operations`...do we still want this test now that `close` is no more? I'll leave it for now.

storage/fuzz/fuzz_targets/fixed_journal_operations.rs
```diff
@@ -132,11 +131,6 @@ fn fuzz(input: FuzzInput) {
                     }
                 }
 
-                JournalOperation::Close => {
-                    journal.close().await.unwrap();
```
roberto-bayardo, line 136: Instead of removing this, we could drop(journal) then return.

storage/fuzz/fuzz_targets/mmr_journaled.rs
roberto-bayardo, line 438: Since this is using the same config as the previous one, should the previous one be dropped first? (Not sure if some kind of resource conflict could arise.)

storage/src/journal/contiguous/fixed.rs
roberto-bayardo: agree, in fact there's a longstanding open PR from andre to make these fully immutable!

storage/src/journal/authenticated.rs
roberto-bayardo: Probably good to keep, maybe just rename the test to test_sync instead of test_close?

storage/fuzz/fuzz_targets/mmr_journaled.rs
danlaine: good catch, fixed

storage/fuzz/fuzz_targets/fixed_journal_operations.rs
danlaine: If we just drop the journal and return, I'm not sure that adds additional test coverage. I could see a case for dropping and re-creating the journal, though.

storage/src/journal/authenticated.rs
danlaine: Updated the name

storage/fuzz/fuzz_targets/fixed_journal_operations.rs
danlaine: I added `Restart` operation type to do this ^

storage/fuzz/fuzz_targets/fixed_journal_operations.rs
```diff
@@ -132,9 +132,9 @@ fn fuzz(input: FuzzInput) {
...
-                    return;
+                JournalOperation::Restart => {
+                    drop(journal);
+                    journal = Journal::init(context.clone(), cfg.clone()).await.unwrap();
```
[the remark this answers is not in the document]
danlaine: Updated restart case to update tracking variables

storage/fuzz/fuzz_targets/mmr_journaled.rs
```diff
@@ -430,17 +430,8 @@ fn fuzz(input: FuzzInput) {
...
-                    // Close the existing MMR
-                    match mmr {
-                        MmrState::Clean(m) => {
-                            m.close().await.unwrap();
-                        }
-                        MmrState::Dirty(m) => {
-                            m.merkleize(&mut hasher).close().await.unwrap();
-                        }
-                    }
-
```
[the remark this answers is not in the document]
danlaine: Updated to clear leaves on re-init

PR #2613, opened 2025-12-24
[Storage] Remove `close` methods

danlaine:
Follow-up to #2608 to remove more `close` methods from `commonware-storage` types.

Replaces `close` with `sync` and/or `drop` as needed. (In some places, we don't need to `sync` because the store is unmodified. In some places, we don't need to `drop` because the value is dropped anyway.)

Leaves the `close` method on `Freezer` because that has logic other than just syncing its internal stores.


storage/src/freezer/storage.rs
```diff
@@ -1003,8 +1003,6 @@ impl<E: Storage + Metrics + Clock, K: Array, V: Codec> Freezer<E, K, V> {
         // Sync any pending updates before closing
         let checkpoint = self.sync().await?;
 
-        self.journal.close().await?;
```
danlaine, line 1009: `self.sync()` above syncs all journal sections

storage/src/freezer/storage.rs
```diff
@@ -1003,8 +1003,6 @@ impl<E: Storage + Metrics + Clock, K: Array, V: Codec> Freezer<E, K, V> {
...
-        self.table.sync().await?;
```
danlaine, line 1010: `self.sync()` above syncs `self.table`

storage/src/metadata/benches/restart.rs
```diff
@@ -33,9 +33,8 @@ fn bench_restart(c: &mut Criterion) {
                 for _ in 0..iters {
                     let start = Instant::now();
                     // This is the benchmarked operation
-                    let metadata = init(ctx.clone()).await;
+                    let _metadata = init(ctx.clone()).await;
                     total += start.elapsed();
-                    metadata.close().await.unwrap();
```
danlaine, line 38: We don't modify `metadata` so syncing here would be a no-op

storage/src/journal/segmented/variable.rs
```diff
@@ -1491,7 +1472,7 @@ mod tests {
                 }
                 assert!(items.is_empty());
             }
-            journal.close().await.expect("Failed to close journal");
+            drop(journal);
```
roberto-bayardo, line 1475: should we sync here first?

roberto-bayardo: oh looks like you never changed it so should be ok.

roberto-bayardo (approved):

PR #2618, opened 2025-12-25
fix: clarify TipManager invariants

Snezhkko:
Previously the TipManager comment claimed that stored tips always had parent certificates and had already been acked by the local validator, which is not enforced by TipManager itself and can be violated (e.g. during verification or in tests). This change updates the comment to document the actual invariants TipManager maintains (monotonic heights per sequencer and payload consistency at a given height) and makes it explicit that certificates and acknowledgements are tracked by other components such as the Engine and AckManager. No behavior changes were made.


clabby:

PR #2619, opened 2025-12-25
[storage/qmdb/current] validate the number of chunks in range_proof input

roberto-bayardo:
- DIsallows extra chunks in range_proof verification since it could lead to potential malleability issues.
- Adds regression tests for the issue.

### DIFF
--- a/storage/src/qmdb/current/proof.rs
+++ b/storage/src/qmdb/current/proof.rs
@@ -108,8 +108,8 @@ impl<D: Digest> RangeProof<D> {
 
         // Gather the chunks necessary to verify the proof.
         let chunk_bits = CleanBitMap::<H::Digest, N>::CHUNK_SIZE_BITS;
-        let start = *start_loc / chunk_bits; // chunk that contains the very first bit.
-        let end = (*end_loc - 1) / chunk_bits; // chunk that contains the very last bit.
+        let start = *start_loc / chunk_bits; // chunk that contains the first bit
+        let end = (*end_loc - 1) / chunk_bits; // chunk that contains the last bit
         let mut chunks = Vec::with_capacity((end - start + 1) as usize);
         for i in start..=end {
             let bit_offset = i * chunk_bits;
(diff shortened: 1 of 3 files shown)


storage/src/qmdb/current/proof.rs
```diff
@@ -144,11 +144,22 @@ impl<D: Digest> RangeProof<D> {
...
+                ?op_count, "verification failed, invalid range"
             );
             return false;
         }
 
+        // Validate the number of input chunks.
+        let chunk_bits = CleanBitMap::<H::Digest, N>::CHUNK_SIZE_BITS;
+        let start = *start_loc / chunk_bits; // chunk that contains first bit
+        let end = (*end_loc - 1) / chunk_bits; // chunk that contains the last bit
+        let expected = end - start + 1;
```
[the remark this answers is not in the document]
roberto-bayardo: thanks bot, fixed + regression test added.

storage/src/qmdb/current/proof.rs
```diff
@@ -144,11 +148,22 @@ impl<D: Digest> RangeProof<D> {
...
-                ?op_count, "proof verification failed, invalid range"
+                ?op_count, "verification failed, invalid range"
             );
             return false;
         }
 
+        // Validate the number of input chunks.
+        let chunk_bits = CleanBitMap::<H::Digest, N>::CHUNK_SIZE_BITS;
+        let start = *start_loc / chunk_bits; // chunk that contains first bit
+        let end = (*end_loc.saturating_sub(1)) / chunk_bits; // chunk that contains the last bit
```
clabby, line 159: `end_loc` can never be `0` here, since this path is gated by the `if ops.is_empty() || chunks.is_empty()` check on L138. 

```suggestion
        let end = (*end_loc - 1) / chunk_bits; // chunk that contains the last bit
```

clabby (approved):
Nice. Let's add the https://github.com/commonwarexyz/monorepo/labels/bug label to help track for the next release's changelog?

storage/src/qmdb/current/proof.rs
roberto-bayardo: Yes, my thinking was it doesn't hurt to be extra safe here though since this is code that executes on untrusted input. (And indeed my first attempt at this didn't include the earlier empty checks, which resulted in a panic! vulnerability.)

PR #2621, opened 2025-12-26
[parallel] Introduce data parallelism abstraction

### LINKED ISSUE #1540: [runtime] `create_pool` Only Works in `tokio`

### LINKED ISSUE #2187: Idea: abstract parallelism
In a few places, we use [rayon](https://docs.rs/rayon/latest/rayon/) to parallelize code. The current pattern is to have a `concurrency: usize` parameter, controlling the number of threads to use, with 1 disabling parallelization entirely. We also need to gate that section of the code to support no_std, if the crate wants to do that.

It could be a bit nicer to have some kind of trait representing the "parallelism effect", and then allow code to depend on and use that trait, with a single code path, regardless of whether or not it uses parallelism. Depending on the trait would also work without needing to feature gate based on no_std compatibility.

The crate could provide a rayon-based implementation, as well as a plain implementation with no parallelism. Right now, the operations we would need to support are simple enough that a "parallel fold" is likely all we'd need the trait to supply. One risk in this approach is that the trait might become very complicated if we use parallelism in a more complicated way. I expect that for most of our use cases, things will be expressible as a map and fold, or, at most, a scan.

These implementations could then be baked into the runtimes, so that the tokio and deterministic runtimes would implement this trait, allowing code that has a runtime already to conveniently use that runtime as the "parallelizer" for, e.g. cryptography code it calls.

clabby:
## Overview

Inspired by #2187, introduces `commonware-parallel`, which offers a thin abstraction over a `Strategy` that can perform a `fold`or `join` operation either sequentially (with `Iterator`) or in parallel (with `rayon`'s `ParallelIterator`.) These strategies are objects, rather than marker traits, so that types can be given a `Strategy` to be used in any operations attached to it. The `Parallel` strategy owns a `commonware_runtime::ThreadPool` (`Arc<rayon::ThreadPool>`).

In addition, fixes `commonware-runtime`'s `create_pool` function to work with the `deterministic` runtime by using `rayon`'s [`ThreadPoolBuilder::use_current_thread`](https://docs.rs/rayon/latest/rayon/struct.ThreadPoolBuilder.html#method.use_current_thread) option. When set, this enables the current thread to steal queued work when yielded to.

**TODO:**
- [x] `coding`
- [x] `math`
- [x] `cryptography`
- [x] Expose `Strategy` across the board? There's quite a few places where the concurrency argument was not exposed, and this PR tries to honor that. It may be beneficial, though, to allow passing a `Strategy` through every function / type that can make use of it? -> Exposed it in the areas I think make sense.

closes #1540
closes #2187

### DIFF
--- a/cryptography/src/bls12381/primitives/ops/aggregate.rs
+++ b/cryptography/src/bls12381/primitives/ops/aggregate.rs
@@ -201,8 +201,7 @@ mod tests {
     };
     use blst::BLST_ERROR;
     use commonware_codec::Encode;
-    use commonware_parallel::{Rayon, Sequential};
-    use commonware_utils::{test_rng, union_unique, NZUsize};
+    use commonware_utils::{test_rng, union_unique};
 
     fn blst_aggregate_verify_same_message<'a, V, I>(
         public: I,


math/src/algebra.rs
```diff
@@ -200,7 +201,7 @@ pub trait Space<R>:
     /// that they have the same length.
     ///
     /// For empty slices, the result should be [`Additive::zero`];
-    fn msm(points: &[Self], scalars: &[R], _concurrency: usize) -> Self {
+    fn msm<S: ParStrategy>(points: &[Self], scalars: &[R], _strategy: &S) -> Self {
```
patrick-ogrady: nit: can we just use `Strategy` in places like this? I tend to only rename if there is an import conflict?

math/src/algebra.rs
```diff
@@ -627,7 +631,8 @@ mod test {
 
         #[test]
         fn test_msm_2(a: [F; 2], b: [F; 2]) {
-            assert_eq!(F::msm(&a, &b, 1), a[0] * b[0] + a[1] * b[1]);
+            use commonware_parallel::Sequential;
```
patrick-ogrady: nit: we should avoid inline definitions like this?

math/src/algebra.rs
clabby: There is an import conflict, w/ `proptest`'s `Strategy`.

consensus/src/simplex/scheme/bls12381_threshold.rs
```diff
@@ -493,23 +535,27 @@ impl<P: PublicKey, V: Variant> certificate::Scheme for Scheme<P, V> {
...
+            ),
+        ];
+        match &self.pool {
+            Some(pool) => batch::verify_same_signer::<_, V, _, _>(
+                rng,
+                &evaluated,
+                entries,
+                &Parallel::new(pool.clone()),
+            ),
+            None => batch::verify_same_signer::<_, V, _, _>(rng, &evaluated, entries, &Sequential),
```
clabby: Still deliberating on this... Having `Scheme` just hold a `ThreadPool` is nice because we don't have to add another generic to it for `Strategy`, but sort of defeats the purpose of "write the code once."

clabby: Thinking we should go with making `Scheme` generic over `Strategy`.

parallel/src/lib.rs
```diff
@@ -0,0 +1,970 @@
...
+    ///
+    /// let (sum, product) = strategy.join(
+    ///     || (1..=10).sum::<i32>(),
+    ///     || (1..=10).product::<i32>(),
+    /// );
+    ///
+    /// assert_eq!(sum, 55);
+    /// assert_eq!(product, 3628800);
+    /// ```
+    fn join<L, LO, R, RO>(&self, left: L, right: R) -> (LO, RO)
```
cronokirby: Are there places, already, that need this granular level of parallelism?

I think exposing this could be fine, but if so, we should use it to base other methods off of, to avoid burdening the trait too much.

parallel/src/lib.rs
```diff
@@ -0,0 +1,970 @@
...
+        use alloc::vec::Vec;
+    }
+}
+
+/// A strategy for executing fold and join operations.
+///
+/// This trait abstracts over sequential and parallel execution, allowing algorithms
+/// to be written generically and then executed with different strategies depending
+/// on the use case (e.g., sequential for testing/debugging, parallel for production).
+pub trait Strategy: Clone + Send + Sync + fmt::Debug + 'static {
```
cronokirby, line 79: I think this would make more sense as a concrete type, with several methods of initialization, deciding at runtime how to execute these things.

parallel/src/lib.rs
```diff
@@ -0,0 +1,970 @@
...
+                LO: Send,
+                RO: Send,
+            {
+                self.thread_pool.install(|| rayon::join(left, right))
+            }
+        }
+    }
+}
+
+#[cfg(test)]
```
cronokirby, line 441: One idea to simplify the test suite is to enforce that:

- parallel fold_init does the same thing as sequential fold_init,
- sequential "other method" does the same thing as the expression in terms of fold_init

I think you could probably streamline the test suite a lot with that.

I think we could do that in a follow up PR though, I think the existing tests provide enough coverage.

parallel/src/lib.rs
```diff
@@ -0,0 +1,970 @@
...
+///
+/// // Slices also implement it
+/// let slice: &[i32] = &[1, 2, 3];
+/// let _iter = slice.into_iter();
+///
+/// // Ranges work too
+/// let range = 0..100;
+/// let _iter = range.into_iter();
+/// ```
+pub trait IntoStrategyIterator: IntoIterator {
```
cronokirby: I think having a cfg_if defined body of the trait is really hacky.

Also, having rayon leak into the public interface of the module seems less than ideal.

cronokirby: With a concrete struct strategy, one approach would be to have this trait represent the concept of being able to call fold on itself (or whatever the ur-method ends up being) using the strategy as an argument. Then, for feature gating, we would just create a nice impl which hooks into a concrete method on strategy which has parallel iterator as a bound. This would avoid having to know about Rayon to implement this for types that don't need parallelism.

cronokirby (changes requested):
The main RC is to move from a trait with open ended impls, to a concrete struct which dynamically knows how to parallelize things.

I think we should also eschew providing something as low level as join, in favor of focusing on "fold_init" as the source of the difference in behavior.

parallel/src/lib.rs
clabby: Right now, no. The primary reason I added this was for future flexibility in areas where we may want to use parallel divide-and-conquer, for example in the bisection-style verification for BLS.

Glad to remove this for now given that no one uses it, if you'd prefer.

parallel/src/lib.rs
clabby: With having `rayon` leak into the public interface, what are the issues you foresee there? `IntoStrategyIterator` intentionally bridges regular iterators with `rayon`'s parallel iterators (when in an `std` context.) Hiding `rayon` would allow us to use other backends for data parallelism, but I couldn't think of another that we may want to use.

> This would avoid having to know about Rayon to implement this for types that don't need parallelism.

The way I was thinking about it is that you would only use `Strategy` for situations where you want to allow for parallelism via `rayon` specifically. If you don't need parallelism, I was assuming you likely wouldn't use `Strategy` for the type at all. (This was also the intention behind the `IntoStrategyIterator` bridge requiring that `T: IntoParallelIterator` outside of `no_std` contexts.)

Is "use `Strategy` when you may or may not want to parallelize `T`'s operations with `rayon`" too narrow?

cronokirby: Re point 2: you need to use Strategy if you want consumers of your code to be able to parallelize your code. It's correct that there aren't situations where you'd use Strategy if you _never_ want parallelism. However, there are many situations where you need to use Strategy, while also wanting your code to be run Sequentially in, e.g. WASM, or other constrained environments. In that case, you need a public API that's easy to satisfy regardless of what capabilities you have, since that kind of code will have to assume the least possible. You want to avoid the virality induced by having a public API required just to call a function on strategy be feature flagged. Instead, the feature flagging should really only affect what means you have to _construct_ a strategy. Once you have one, it should be usable even if no_std environments.

Re point 1: the main issue is that it makes using the crate more annoying and require reading more documentation needlessly. The user shouldn't have to understand both this crate and read up docs on rayon on how to construct things or what types implement the trait we're linking them to, it should just be a complete interface. It would also be nice to have the option to use other backends, especially if the limitation there comes from exposing implementation details, rather than a higher level conceptual blocker.

clabby: > However, there are many situations where you need to use Strategy, while also wanting your code to be run Sequentially in, e.g. WASM, or other constrained environments. In that case, you need a public API that's easy to satisfy regardless of what capabilities you have, since that kind of code will have to assume the least possible. You want to avoid the virality induced by having a public API required just to call a function on strategy be feature flagged. Instead, the feature flagging should really only affect what means you have to construct a strategy. Once you have one, it should be usable even if no_std environments.

This makes sense. I think the existing API is relatively friendly to deal with here personally, but def open to disagreement. Users should never manually implement `IntoStrategyIterator` for themselves, their code would look something like:

```rs
struct Thing([u8; 16]);

impl IntoIterator for Thing { /* ... */ }

#[cfg(feature = "std")]
impl IntoParallelIterator for Thing { /* ... */ }
```

```toml
[dependencies]
rayon = { workspace = true, optional = true }

[features]
std = ["dep:rayon"]
```

where feature flagging would be necessary anyways due to including `rayon` in the `std` feature. `Thing` here would automatically implement `commonware_parallel::IntoStrategyIterator`, whether `std` was enabled or not.

> the main issue is that it makes using the crate more annoying and require reading more documentation needlessly. The user shouldn't have to understand both this crate and read up docs on rayon on how to construct things or what types implement the trait we're linking them to, it should just be a complete interface.

I mainly didn't want to prematurely abstract here, didn't know of any other backends we may want to use. Are there any that we're thinking of adding?

cronokirby: I agree that whatever trait bounds we require for a collection, they should be automatically implemented on any type a user will realistically provide, and it can be more complex to implement it manually.

It's less that we would consider using something other than rayon, but more so that we don't want to commit to a public API based on rayon, because then you're stuck with its way of modelling things forever.

Even more importantly, you want to avoid a situation in which to understand and use this crate, you also need to read a bunch of rayon docs first to figure out how to construct the things you need to pass in here, or what types this crate is asking for, etc. In other words, we want to abstract over rayon, not provide a slightly different interface in addition to it.

parallel/src/lib.rs
clabby: Removed this for now.

parallel/src/lib.rs
clabby: Updated these while removing `IntoStrategyIterator` 👍

coding/src/benches/bench.rs
```diff
@@ -18,6 +21,7 @@ pub(crate) fn benchmark_encode_generic<S: Scheme>(name: &str, c: &mut Criterion)
                     minimum_shards: min as u16,
                     extra_shards: (chunks - min) as u16,
                 };
+                let pool = Arc::new(ThreadPoolBuilder::new().num_threads(conc).build().unwrap());
```
patrick-ogrady: Can we create a helper for this in parallel?

coding/src/benches/bench.rs
```diff
@@ -28,7 +32,17 @@ pub(crate) fn benchmark_encode_generic<S: Scheme>(name: &str, c: &mut Criterion)
...
-                            |data| S::encode(&config, data.as_slice(), conc),
+                            |data| {
+                                if conc > 1 {
+                                    S::encode(
+                                        &config,
+                                        data.as_slice(),
+                                        &Parallel::new(pool.clone()),
+                                    )
+                                } else {
+                                    S::encode(&config, data.as_slice(), &Sequential)
```
patrick-ogrady: A shame that there isn't a single variable we could use to house `Sequential/Parallel` to avoid rewriting the function.

patrick-ogrady: I suppose we could have an enum of sorts that we just provide `conc` to but maybe that defeats the purpose.

examples/bridge/src/bin/validator.rs
```diff
@@ -217,11 +218,20 @@ fn main() {
         );
 
         // Initialize application
+        let n_threads = std::thread::available_parallelism().unwrap().get();
```
patrick-ogrady: There are other threads active here, I don't think we should take all (for example the ones running tokio?)

patrick-ogrady: Let's just do 2

examples/reshare/src/validator.rs
```diff
@@ -116,7 +121,10 @@ pub async fn run<S, L>(
     };
     let marshal = marshal_resolver::init(&context, resolver_cfg, marshal);
 
-    let engine = engine::Engine::<_, _, _, _, Sha256, MinSig, S, L>::new(
+    let n_threads = std::thread::available_parallelism().unwrap().get();
```
patrick-ogrady: another case where we shouldn't take all

parallel/README.md
```diff
@@ -0,0 +1,10 @@
+# commonware-parallel
+
+[![Crates.io](https://img.shields.io/crates/v/commonware-parallel.svg)](https://crates.io/crates/commonware-parallel)
+[![Docs.rs](https://docs.rs/commonware-parallel/badge.svg)](https://docs.rs/commonware-parallel)
+
+Abstract data parallelism over iterators and tasks.
```
patrick-ogrady: nit: We should make the first word a verb

(maybe `Parallelize...`)

runtime/src/utils/mod.rs
```diff
@@ -295,6 +298,23 @@ mod tests {
         });
     }
 
+    #[test_traced]
+    fn test_create_pool_deterministic() {
```
patrick-ogrady: ❤️

parallel/src/lib.rs
```diff
@@ -0,0 +1,537 @@
...
+        /// let data: Vec<i64> = (0..1000).collect();
+        /// let sum = strategy.fold(&data, || 0i64, |acc, &n| acc + n, |a, b| a + b);
+        /// assert_eq!(sum, 499500);
+        /// ```
+        #[derive(Debug, Clone)]
+        pub struct Parallel {
+            thread_pool: Arc<ThreadPool>,
+        }
+
+        impl Parallel {
```
cronokirby: It would be useful to have a constructor here which can construct a thread pool itself, e.g. in tests.

parallel/src/lib.rs
```diff
@@ -0,0 +1,537 @@
...
+        use alloc::vec::Vec;
+    }
+}
+
+/// A strategy for executing fold and join operations.
+///
+/// This trait abstracts over sequential and parallel execution, allowing algorithms
+/// to be written generically and then executed with different strategies depending
+/// on the use case (e.g., sequential for testing/debugging, parallel for production).
+pub trait Strategy: Clone + Send + Sync + fmt::Debug + 'static {
```
cronokirby, line 79: This is a bit of a bikeshed, but in most places, you'd expect people to see `&impl Strategy` as an argument, and then `&Parallel` or `&Sequential`. I think `&impl Strategy` might, itself, be a bit confusing, since you need to see the import from parallel to understand what it is. `Parallel` also seems like a bad name for something rayon specific.

Perhaps:

- trait `Parallel`
- impl `ParallelNone`
- impl `ParallelRayon`

would be better?

parallel/src/lib.rs
```diff
@@ -0,0 +1,537 @@
...
+            pub const fn new(thread_pool: Arc<ThreadPool>) -> Self {
+                Self { thread_pool }
+            }
+        }
+
+        impl From<Arc<ThreadPool>> for Parallel {
+            fn from(thread_pool: Arc<ThreadPool>) -> Self {
+                Self::new(thread_pool)
+            }
+        }
```
cronokirby: I think `&Parallel::new()` would be a lot clearer, and I think we want to encourage a pattern of storing a generic or a dyn impl of the trait, rather than storing a thread pool and creating this, so I think we should remove.

math/src/algebra.rs
cronokirby: `_: &impl Parallel` would be nice here, eh?

coding/src/benches/bench.rs
cronokirby: We should use `Box<dyn Strategy>`

clabby: This is the discussion around using a concrete type (or `Box<dyn Strategy>`) vs. the trait.

parallel/src/lib.rs
clabby: Makes sense. Removed the `From` impl.

coding/src/benches/bench.rs
clabby: From what we spoke about on slack - `Strategy` (now `Parallel`) cannot be (usefully) dyn-compatible because each generic method requires `Self: Sized` and therefore cannot be included in the trait object's vtable.

runtime/src/utils/mod.rs
```diff
@@ -89,22 +88,25 @@ pub type ThreadPool = Arc<RThreadPool>;
...
-            context
-                .with_label("rayon_thread")
-                .dedicated()
-                .spawn(move |_| async move { thread.run() });
-            Ok(())
-        })
-        .build()?;
+    commonware_parallel::create_pool_with(concurrency, |mut builder| {
+        if rayon::current_thread_index().is_none() {
+            builder = builder.use_current_thread()
```
patrick-ogrady: Confirming we only do this in deterministic?

clabby: All runtimes. There's no way to "detect" whether we're using `deterministic`, with the detached `create_pool` method.

parallel/src/lib.rs
```diff
@@ -0,0 +1,548 @@
...
+
+        /// Creates a [`ThreadPool`] with custom configuration.
+        pub fn create_pool_with(
+            num_threads: NonZeroUsize,
+            build: impl FnOnce(ThreadPoolBuilder) -> Result<RThreadPool, ThreadPoolBuildError>,
+        ) -> Result<ThreadPool, ThreadPoolBuildError> {
+            let builder = ThreadPoolBuilder::new().num_threads(num_threads.get());
+            build(builder).map(Arc::new)
+        }
+    }
```
cronokirby: Why not have these be associated methods on `ParallelRayon`, returning `Self`?

clabby: Either works. These are separate right now to be used with `commonware_runtime::create_pool`.

clabby: Updated these.

runtime/src/deterministic.rs
```diff
@@ -1064,22 +1064,23 @@ impl crate::Spawner for Context {
...
-                })
-                .build()
-        })
+        if rayon::current_thread_index().is_none() {
+            builder = builder.use_current_thread()
+        }
+
+        builder
+            .spawn_handler(move |thread| {
+                // Tasks spawned in a thread pool are expected to run longer than any single
```
patrick-ogrady: nit: remove this comment?

unlike tokio, these all happen on the same thread?

Now that this is part of deterministic, we may be able to do something much dumber here.

cryptography/fuzz/fuzz_targets/bls12381_aggregate_operations.rs
```diff
@@ -164,14 +166,15 @@ fn fuzz(op: FuzzOperation) {
...
             concurrency,
         } => {
-            if !messages.is_empty() && concurrency > 0 {
+            if !messages.is_empty() {
                 let messages_refs: Vec<(&[u8], &[u8])> = messages
                     .iter()
                     .map(|(ns, msg)| (ns.as_slice(), msg.as_slice()))
                     .collect();
 
+                let strategy = Rayon::new(NonZeroUsize::new(concurrency).unwrap()).unwrap();
```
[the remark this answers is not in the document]
clabby: See the `Arbitrary` impl. `concurrency: u.int_in_range(1..=8)?`.

PR #2622, opened 2025-12-26
[storage/index] allow Translator capping at sizes that are not equal to some int type

roberto-bayardo:
The current "*Cap" translators only support sizes equivalent to some int type.  This PR allows capping at arbitrary sizes from [1-8].  Thus we can create an effective "9 byte cap" using a 2-byte partitioned index with a 7-byte translator to match the parameters used in the QMDB whitepaper.

### DIFF
--- a/storage/src/translator.rs
+++ b/storage/src/translator.rs
@@ -1,6 +1,9 @@
 //! Primitive implementations of [Translator].
 
-use std::hash::{BuildHasher, Hash, Hasher};
+use std::{
+    hash::{BuildHasher, Hash, Hasher},
+    marker::PhantomData,
+};
 
 /// Translate keys into a new representation (often a smaller one).
 ///
@@ -134,7 +137,6 @@ pub struct UnhashedArray<const N: usize> {
 impl<const N: usize> Hash for UnhashedArray<N> {
     #[inline]
     fn hash<H: Hasher>(&self, state: &mut H) {
-        const { assert!(N <= 8 && N > 0) };
         state.write(&self.inner);
     }
 }
@@ -148,13 +150,16 @@ impl<const N: usize> PartialEq<[u8; N]> for UnhashedArray<N> {
 /// Translators for keys that are not the length of a standard integer.
 #[derive(Clone, Copy)]
 pub struct Cap<const N: usize> {
-    _marker: UnhashedArray<N>,
+    _phantom: PhantomData<[u8; N]>,
 }
 
 impl<const N: usize> Cap<N> {
     pub const fn new() -> Self {
+        const {
+            assert!(N <= 8 && N > 0, "Cap must be between 1 and 8");
+        };
         Self {
-            _marker: UnhashedArray { inner: [0; N] },
+            _phantom: PhantomData,
         }
     }
 }
@@ -170,7 +175,9 @@ impl<const N: usize> Translator for Cap<N> {
 
     #[inline]
     fn transform(&self, key: &[u8]) -> Self::Key {
-        const { assert!(N <= 8 && N > 0) };
+        const {
+            assert!(N <= 8 && N > 0, "Cap must be between 1 and 8");
+        };
         UnhashedArray {
             inner: cap::<N>(key),
         }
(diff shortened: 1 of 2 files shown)


roberto-bayardo:
Should I also retire the macro-derived `One/Two/Four/EightCap` types in favor of always using `Cap<N>`?

roberto-bayardo:
> Should I also retire the macro-derived `One/Two/Four/EightCap` types in favor of always using `Cap<N>`?

There's a slight performance penalty (~10%) using Cap::<4> over FourCap with ordered index. I suspect the default comparison operation for 4-byte arrays is just slower than the u32 key type used by FourCap :shrug:

storage/src/translator.rs
```diff
@@ -120,6 +123,69 @@ define_cap_translator!(TwoCap, 2, u16);
...
+impl<const N: usize> PartialEq<[u8; N]> for UnhashedArray<N> {
+    fn eq(&self, other: &[u8; N]) -> bool {
+        &self.inner == other
+    }
+}
+
+/// Translators for keys that are not the length of a standard integer.
+#[derive(Clone, Copy)]
+pub struct Cap<const N: usize> {
+    _marker: UnhashedArray<N>,
```
clabby: nit: use `core::marker::PhantomData` so that `size_of::<Cap<N>> == 0`.

storage/src/translator.rs
```diff
@@ -120,6 +123,69 @@ define_cap_translator!(TwoCap, 2, u16);
...
+/// prefix.
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Debug)]
+pub struct UnhashedArray<const N: usize> {
+    pub inner: [u8; N],
+}
+
+impl<const N: usize> Hash for UnhashedArray<N> {
+    #[inline]
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        const { assert!(N <= 8 && N > 0) };
```
clabby: nit: Add an error message here so that folks get a nice compile error if they misuse the type.

storage/src/translator.rs
```diff
@@ -120,6 +123,69 @@ define_cap_translator!(TwoCap, 2, u16);
...
+        Self::new()
+    }
+}
+
+impl<const N: usize> Translator for Cap<N> {
+    type Key = UnhashedArray<N>;
+
+    #[inline]
+    fn transform(&self, key: &[u8]) -> Self::Key {
+        const { assert!(N <= 8 && N > 0) };
```
clabby: nit: Add an error message here so that folks get a nice compile error if they misuse the type.

storage/src/translator.rs
```diff
@@ -120,6 +123,69 @@ define_cap_translator!(TwoCap, 2, u16);
...
+
+/// Translators for keys that are not the length of a standard integer.
+#[derive(Clone, Copy)]
+pub struct Cap<const N: usize> {
+    _marker: UnhashedArray<N>,
+}
+
+impl<const N: usize> Cap<N> {
+    pub const fn new() -> Self {
+        Self {
```
clabby, line 161: Add an `const { assert!(N <= 8 && N > 0) }` block here so that the error presents itself on initialization?

clabby (commented):
Nice. Couple small nits, agree that it'd be nice to move away from `One/Two/Four/Eight` cap in favor of just `Cap<N>`. Maybe just add some type aliases?

clabby:
> The reason it's slower is probably the difference between calling `UintIdentity::write_u32` w/ `FourCap` vs. `UintIdentity::write` w/ `Cap<N>`.

It's only slower on ordered indices though (no performance difference at all with unordered) which is what leads me to suspect comparison (not equality) ops. Isn't a u32 comparison on little-endian architecture different than Array[4] (byte-by-byte)?

storage/src/translator.rs
roberto-bayardo: added.

storage/src/translator.rs
roberto-bayardo: done

storage/src/translator.rs
roberto-bayardo: I've removed this assertion, since it's actually OK to use UnhashedArray with a larger N.  The assertion should be part of Cap (as you also suggested) so moved it there.

storage/src/translator.rs
roberto-bayardo: done

Issue #1919, opened 2025-10-18
[workspace] prefer CryptoRngCore


andresilva:
We can replace trait bounds like `Rng + CryptoRng` with just `CryptoRngCore`. Currently we mix both usages and we should prefer the latter.

clabby:
`Rng + CryptoRng` is a more featured variant, since `CryptoRngCore` just implies `CryptoRng + RngCore`. Not everywhere needs the features of `Rng`, but we may still need a mix.

Issue #1353, opened 2025-08-04
[consensus/simplex] Compressed Nullifications


patrick-ogrady:
While at the [Science and Engineering of Consensus 2025](https://tselab.stanford.edu/workshop-sbc25/) workshop, during [Victor Shoup's](https://www.shoup.net/) presentation about [Kudzu](https://arxiv.org/abs/2505.08771) someone asked about the unbounded growth of "memory" required to store nullifications between productive views of Simplex (and made the claim this is much worse than the amount of "memory" required to safely run PBFT-style consensus).

Following his presentation, Victor and I chatted about this a bit and an interesting idea emerged that you could create a constant-size certificate that represents any number of consecutive nullifications (if the underlying Simplex construction uses threshold signatures). In short, you could construct the following:

```txt
start: View
end: View
proof: Signature
```

To verify said certificate, you would just run [aggregate_verify_multiple_messages](https://github.com/commonwarexyz/monorepo/blob/65ee8a7532a698b2a45bcfdc509f42c4fd49dfac/cryptography/src/bls12381/primitives/ops.rs#L618-L679) where the messages are all the views from `start` to `end`.

I consider this a much more powerful mechanism than the one discussed in #1078 (especially if you did a "notarization" + "nullification certificate" in each block), if we continue discussing such practical considerations.

andresilva:
I started some work on this: https://github.com/andresilva/monorepo/compare/main...andresilva:monorepo:andre/threshold-simplex-compressed-nullifications.

Added `NullificationRange` type that compresses consecutive nullified views into a single proof with one aggregated BLS signature.

Added `Nullifications` store that automatically compacts consecutive nullifications and merges adjacent ranges for optimal storage:
- Compaction - Singles at views 5,6,7 automatically become range [5-7]
- Merging - Adding 9 to [5-8] and single 10 creates [5-10]
- Querying - `get(view)` returns the appropriate proof, `gaps()` finds missing nullifications
- Pruning - Removes old nullifications while preserving partial ranges
- Achieves optimal compaction by exhaustively checking for merge opportunities with adjacent entries whenever singles or ranges are added

I think the nullification store can later be abstracted (and moved elsewhere) for reuse in Minimmit. It's possible that the store might be a bit overengineered, in particular because we are not expecting long periods of nullifications, but the entire point of the issue is to prevent edge cases that could lead to DoS. Let me know what you think.

[...]

PR #2624, opened 2025-12-28
[cryptography/bls12381] Refactor Batch Verification

patrick-ogrady:
Closes: #1919 

This PR fails the conformance label check because we added new types.

### DIFF
--- a/cryptography/src/bls12381/certificate/multisig/mod.rs
+++ b/cryptography/src/bls12381/certificate/multisig/mod.rs
@@ -541,7 +541,6 @@ mod tests {
     use commonware_codec::{Decode, Encode};
     use commonware_math::algebra::{CryptoGroup, Random};
     use commonware_utils::{ordered::BiMap, quorum, test_rng, TryCollect};
-    use rand::{rngs::StdRng, SeedableRng};
 
     const NAMESPACE: &[u8] = b"test-bls12381-multisig";
     const MESSAGE: &[u8] = b"test message";
@@ -562,19 +561,17 @@ mod tests {
     impl_certificate_bls12381_multisig!(TestSubject<'a>);
 
     fn setup_signers<V: Variant>(
+        rng: &mut impl CryptoRngCore,
         n: u32,
-        seed: u64,
     ) -> (
         Vec<Scheme<ed25519::PublicKey, V>>,
         Scheme<ed25519::PublicKey, V>,
     ) {
-        let mut rng = StdRng::seed_from_u64(seed);
-
         // Generate identity keys (ed25519) and consensus keys (BLS)
         let identity_keys: Vec<_> = (0..n)
-            .map(|_| Ed25519PrivateKey::random(&mut rng))
+            .map(|_| Ed25519PrivateKey::random(&mut *rng))
             .collect();
-        let consensus_keys: Vec<Private> = (0..n).map(|_| Private::random(&mut rng)).collect();
+        let consensus_keys: Vec<Private> = (0..n).map(|_| Private::random(&mut *rng)).collect();
 
         // Build BiMap of identity public keys -> consensus public keys
         let participants: BiMap<ed25519::PublicKey, V::Public> = identity_keys
@@ -611,14 +608,15 @@ mod tests {
     }
 
     fn test_sign_vote_roundtrip<V: Variant>() {
-        let (schemes, _) = setup_signers::<V>(4, 42);
+        let mut rng = test_rng();
+        let (schemes, _) = setup_signers::<V>(&mut rng, 4);
         let scheme = &schemes[0];
 
         let attestation = scheme
             .sign::<Sha256Digest>(NAMESPACE, TestSubject { message: MESSAGE })
             .unwrap();
         assert!(scheme.verify_attestation::<_, Sha256Digest>(
-            &mut test_rng(),
+            &mut rng,
             NAMESPACE,
             TestSubject { message: MESSAGE },
             &attestation
@@ -632,7 +630,8 @@ mod tests {
     }
 
     fn test_verifier_cannot_sign<V: Variant>() {
-        let (_, verifier) = setup_signers::<V>(4, 43);
+        let mut rng = test_rng();
+        let (_, verifier) = setup_signers::<V>(&mut rng, 4);
         assert!(verifier
             .sign::<Sha256Digest>(NAMESPACE, TestSubject { message: MESSAGE })
             .is_none());
@@ -645,7 +644,8 @@ mod tests {
     }
 
     fn test_verify_attestations_filters_invalid<V: Variant>() {
-        let (schemes, _) = setup_signers::<V>(5, 44);
+        let mut rng = test_rng();
+        let (schemes, _) = setup_signers::<V>(&mut rng, 5);
         let quorum = quorum(schemes.len() as u32) as usize;
 
         let attestations: Vec<_> = schemes
@@ -657,7 +657,6 @@ mod tests {
             })
             .collect();
 
-        let mut rng = StdRng::seed_from_u64(45);
         let result = schemes[0].verify_attestations::<_, Sha256Digest, _>(
             &mut rng,
             NAMESPACE,
@@ -699,7 +698,8 @@ mod tests {
     }
 
     fn test_assemble_certificate<V: Variant>() {
-        let (schemes, _) = setup_signers::<V>(4, 46);
+        let mut rng = test_rng();
+        let (schemes, _) = setup_signers::<V>(&mut rng, 4);
         let quorum = quorum(schemes.len() as u32) as usize;
 
         let attestations: Vec<_> = schemes
@@ -722,26 +722,31 @@ mod tests {
     }
 
     fn test_assemble_certificate_sorts_signers<V: Variant>() {
-        let (schemes, _) = setup_signers::<V>(4, 47);
+        let mut rng = test_rng();
+        let (schemes, _) = setup_signers::<V>(&mut rng, 4);
 
-        // Create votes in non-sorted order (indices 2, 0, 1)
+        // Get indices and sort them to create attestations in guaranteed reverse order
+        let mut indexed: Vec<_> = (0..3).map(|i| (schemes[i].me().unwrap(), i)).collect();
+        indexed.sort_by_key(|(idx, _)| *idx);
+
+        // Create attestations in reverse sorted order (guaranteed non-sorted)
         let attestations = vec![
-            schemes[2]
+            schemes[indexed[2].1]
                 .sign::<Sha256Digest>(NAMESPACE, TestSubject { message: MESSAGE })
                 .unwrap(),
-            schemes[0]
+            schemes[indexed[1].1]
                 .sign::<Sha256Digest>(NAMESPACE, TestSubject { message: MESSAGE })
                 .unwrap(),
-            schemes[1]
+            schemes[indexed[0].1]
                 .sign::<Sha256Digest>(NAMESPACE, TestSubject { message: MESSAGE })
                 .unwrap(),
         ];
 
         let certificate = schemes[0].assemble(attestations).unwrap();
-        assert_eq!(
-            certificate.signers.iter().collect::<Vec<_>>(),
-            vec![0, 1, 2]
-        );
+
+        // Verify signers are sorted by signer index
+        let expected: Vec<_> = indexed.iter().map(|(idx, _)| *idx).collect();
+        assert_eq!(certificate.signers.iter().collect::<Vec<_>>(), expected);
     }
 
     #[test]
@@ -751,7 +756,8 @@ mod tests {
     }
 
     fn test_verify_certificate<V: Variant>() {
-        let (schemes, verifier) = setup_signers::<V>(4, 48);
+        let mut rng = test_rng();
+        let (schemes, verifier) = setup_signers::<V>(&mut rng, 4);
         let quorum = quorum(schemes.len() as u32) as usize;
 
         let attestations: Vec<_> = schemes
@@ -765,7 +771,6 @@ mod tests {
 
         let certificate = schemes[0].assemble(attestations).unwrap();
 
-        let mut rng = StdRng::seed_from_u64(49);
         assert!(verifier.verify_certificate::<_, Sha256Digest>(
             &mut rng,
             NAMESPACE,
@@ -782,7 +787,7 @@ mod tests {
 
     fn test_verify_certificate_detects_corruption<V: Variant>() {
         let mut rng = test_rng();
-        let (schemes, verifier) = setup_signers::<V>(4, 50);
+        let (schemes, verifier) = setup_signers::<V>(&mut rng, 4);
         let quorum = quorum(schemes.len() as u32) as usize;
 
         let attestations: Vec<_> = schemes
@@ -822,7 +827,8 @@ mod tests {
     }
 
     fn test_certificate_codec_roundtrip<V: Variant>() {
-        let (schemes, _) = setup_signers::<V>(4, 51);
+        let mut rng = test_rng();
+        let (schemes, _) = setup_signers::<V>(&mut rng, 4);
         let quorum = quorum(schemes.len() as u32) as usize;
 
         let attestations: Vec<_> = schemes
@@ -848,7 +854,8 @@ mod tests {
     }
 
     fn test_certificate_rejects_sub_quorum<V: Variant>() {
-        let (schemes, _) = setup_signers::<V>(4, 52);
+        let mut rng = test_rng();
+        let (schemes, _) = setup_signers::<V>(&mut rng, 4);
         let sub_quorum = 2; // Less than quorum (3)
 
         let attestations: Vec<_> = schemes
@@ -870,7 +877,8 @@ mod tests {
     }
 
     fn test_certificate_rejects_invalid_signer<V: Variant>() {
-        let (schemes, _) = setup_signers::<V>(4, 53);
+        let mut rng = test_rng();
+        let (schemes, _) = setup_signers::<V>(&mut rng, 4);
         let quorum = quorum(schemes.len() as u32) as usize;
 
         let mut attestations: Vec<_> = schemes
@@ -895,7 +903,8 @@ mod tests {
     }
 
     fn test_verify_certificate_rejects_sub_quorum<V: Variant>() {
-        let (schemes, verifier) = setup_signers::<V>(4, 54);
+        let mut rng = test_rng();
+        let (schemes, verifier) = setup_signers::<V>(&mut rng, 4);
         let participants_len = schemes.len();
 
         let attestations: Vec<_> = schemes
@@ -915,7 +924,7 @@ mod tests {
         certificate.signers = Signers::from(participants_len, signers);
 
         assert!(!verifier.verify_certificate::<_, Sha256Digest>(
-            &mut test_rng(),
+            &mut rng,
             NAMESPACE,
             TestSubject { message: MESSAGE },
             &certificate
@@ -929,7 +938,8 @@ mod tests {
     }
 
     fn test_verify_certificate_rejects_signers_size_mismatch<V: Variant>() {
-        let (schemes, verifier) = setup_signers::<V>(4, 55);
+        let mut rng = test_rng();
+        let (schemes, verifier) = setup_signers::<V>(&mut rng, 4);
         let participants_len = schemes.len();
 
         let attestations: Vec<_> = schemes
@@ -948,7 +958,7 @@ mod tests {
         certificate.signers = Signers::from(participants_len + 1, signers);
 
         assert!(!verifier.verify_certificate::<_, Sha256Digest>(
-            &mut test_rng(),
+            &mut rng,
             NAMESPACE,
             TestSubject { message: MESSAGE },
             &certificate
@@ -962,7 +972,8 @@ mod tests {
     }
 
     fn test_verify_certificates_batch<V: Variant>() {
-        let (schemes, verifier) = setup_signers::<V>(4, 56);
+        let mut rng = test_rng();
+        let (schemes, verifier) = setup_signers::<V>(&mut rng, 4);
         let quorum = quorum(schemes.len() as u32) as usize;
 
         let messages = [b"msg1".as_slice(), b"msg2".as_slice(), b"msg3".as_slice()];
@@ -985,7 +996,6 @@ mod tests {
             .zip(&certificates)
             .map(|(msg, cert)| (TestSubject { message: msg }, cert));
 
-        let mut rng = StdRng::seed_from_u64(57);
         assert!(verifier.verify_certificates::<_, Sha256Digest, _>(&mut rng, NAMESPACE, certs_iter));
     }
 
@@ -996,7 +1006,8 @@ mod tests {
     }
 
     fn test_verify_certificates_batch_detects_failure<V: Variant>() {
-        let (schemes, verifier) = setup_signers::<V>(4, 58);
+        let mut rng = test_rng();
+        let (schemes, verifier) = setup_signers::<V>(&mut rng, 4);
         let quorum = quorum(schemes.len() as u32) as usize;
 
         let messages = [b"msg1".as_slice(), b"msg2".as_slice()];
@@ -1022,7 +1033,6 @@ mod tests {
             .zip(&certificates)
             .map(|(msg, cert)| (TestSubject { message: msg }, cert));
 
-        let mut rng = StdRng::seed_from_u64(59);
         assert!(
             !verifier.verify_certificates::<_, Sha256Digest, _>(&mut rng, NAMESPACE, certs_iter)
         );
@@ -1035,7 +1045,8 @@ mod tests {
     }
 
     fn test_assemble_certificate_rejects_duplicate_signers<V: Variant>() {
-        let (schemes, _) = setup_signers::<V>(4, 60);
+        let mut rng = test_rng();
+        let (schemes, _) = setup_signers::<V>(&mut rng, 4);
 
         let mut attestations: Vec<_> = schemes
             .iter()
@@ -1066,7 +1077,8 @@ mod tests {
     }
 
     fn test_scheme_clone_and_verifier<V: Variant>() {
-        let (schemes, verifier) = setup_signers::<V>(4, 60);
+        let mut rng = test_rng();
+        let (schemes, verifier) = setup_signers::<V>(&mut rng, 4);
 
         // Clone a signer
         let signer = schemes[0].clone();
@@ -1093,7 +1105,8 @@ mod tests {
     }
 
     fn test_certificate_decode_validation<V: Variant>() {
-        let (schemes, _) = setup_signers::<V>(4, 61);
+        let mut rng = test_rng();
+        let (schemes, _) = setup_signers::<V>(&mut rng, 4);
         let participants_len = schemes.len();
 
         let attestations: Vec<_> = schemes
@@ -1137,7 +1150,8 @@ mod tests {
     }
 
     fn test_verify_certificate_rejects_unknown_signer<V: Variant>() {
-        let (schemes, verifier) = setup_signers::<V>(4, 62);
+        let mut rng = test_rng();
+        let (schemes, verifier) = setup_signers::<V>(&mut rng, 4);
         let participants_len = schemes.len();
 
         let attestations: Vec<_> = schemes
@@ -1157,7 +1171,7 @@ mod tests {
         certificate.signers = Signers::from(participants_len + 1, signers);
 
         assert!(!verifier.verify_certificate::<_, Sha256Digest>(
-            &mut test_rng(),
+            &mut rng,
             NAMESPACE,
             TestSubject { message: MESSAGE },
             &certificate,
@@ -1171,8 +1185,8 @@ mod tests {
     }
 
     fn test_verify_attestations_rejects_malleability<V: Variant>() {
-        let mut rng = StdRng::seed_from_u64(12345);
-        let (schemes, _) = setup_signers::<V>(4, 63);
+        let mut rng = test_rng();
+        let (schemes, _) = setup_signers::<V>(&mut rng, 4);
 
         let attestation1 = schemes[0]
             .sign::<Sha256Digest>(NAMESPACE, TestSubject { message: MESSAGE })
--- a/cryptography/src/bls12381/certificate/threshold/mod.rs
+++ b/cryptography/src/bls12381/certificate/threshold/mod.rs
@@ -579,7 +579,6 @@ mod tests {
     use commonware_codec::{DecodeExt, Encode};
     use commonware_math::algebra::{Additive, Random};
     use commonware_utils::{ordered::Set, quorum, test_rng, TryCollect, NZU32};
-    use rand::{rngs::StdRng, SeedableRng};
 
     const NAMESPACE: &[u8] = b"test-bls12381-threshold";
     const MESSAGE: &[u8] = b"test message";
@@ -601,18 +600,16 @@ mod tests {
 
     #[allow(clippy::type_complexity)]
     fn setup_signers<V: Variant>(
+        rng: &mut impl CryptoRngCore,
         n: u32,
-        seed: u64,
     ) -> (
         Vec<Scheme<ed25519::PublicKey, V>>,
         Scheme<ed25519::PublicKey, V>,
         Sharing<V>,
     ) {
-        let mut rng = StdRng::seed_from_u64(seed);
-
         // Generate identity keys (ed25519)
         let identity_keys: Vec<_> = (0..n)
-            .map(|_| Ed25519PrivateKey::random(&mut rng))
+            .map(|_| Ed25519PrivateKey::random(&mut *rng))
             .collect();
         let participants: Set<ed25519::PublicKey> = identity_keys
             .iter()
@@ -622,7 +619,7 @@ mod tests {
 
         // Generate threshold polynomial and shares using DKG
         let (polynomial, shares) =
-            dkg::deal_anonymous::<V>(&mut rng, Default::default(), NZU32!(n));
+            dkg::deal_anonymous::<V>(&mut *rng, Default::default(), NZU32!(n));
 
         let signers = shares
             .into_iter()
@@ -635,14 +632,15 @@ mod tests {
     }
 
     fn test_sign_vote_roundtrip<V: Variant>() {
-        let (schemes, _, _) = setup_signers::<V>(4, 42);
+        let mut rng = test_rng();
+        let (schemes, _, _) = setup_signers::<V>(&mut rng, 4);
         let scheme = &schemes[0];
 
         let attestation = scheme
             .sign::<Sha256Digest>(NAMESPACE, TestSubject { message: MESSAGE })
             .unwrap();
         assert!(scheme.verify_attestation::<_, Sha256Digest>(
-            &mut test_rng(),
+            &mut rng,
             NAMESPACE,
             TestSubject { message: MESSAGE },
             &attestation
@@ -656,7 +654,8 @@ mod tests {
     }
 
     fn test_verifier_cannot_sign<V: Variant>() {
-        let (_, verifier, _) = setup_signers::<V>(4, 43);
+        let mut rng = test_rng();
+        let (_, verifier, _) = setup_signers::<V>(&mut rng, 4);
         assert!(verifier
             .sign::<Sha256Digest>(NAMESPACE, TestSubject { message: MESSAGE })
             .is_none());
@@ -669,7 +668,8 @@ mod tests {
     }
 
     fn test_verify_attestations_filters_invalid<V: Variant>() {
-        let (schemes, _, _) = setup_signers::<V>(5, 44);
+        let mut rng = test_rng();
+        let (schemes, _, _) = setup_signers::<V>(&mut rng, 5);
         let quorum = quorum(schemes.len() as u32) as usize;
 
         let attestations: Vec<_> = schemes
@@ -681,7 +681,6 @@ mod tests {
             })
             .collect();
 
-        let mut rng = StdRng::seed_from_u64(45);
         let result = schemes[0].verify_attestations::<_, Sha256Digest, _>(
             &mut rng,
             NAMESPACE,
@@ -723,7 +722,8 @@ mod tests {
     }
 
     fn test_assemble_certificate<V: Variant>() {
-        let (schemes, verifier, _) = setup_signers::<V>(4, 46);
+        let mut rng = test_rng();
+        let (schemes, verifier, _) = setup_signers::<V>(&mut rng, 4);
         let quorum = quorum(schemes.len() as u32) as usize;
 
         let attestations: Vec<_> = schemes
@@ -739,7 +739,7 @@ mod tests {
 
         // Verify the assembled certificate
         assert!(verifier.verify_certificate::<_, Sha256Digest>(
-            &mut test_rng(),
+            &mut rng,
             NAMESPACE,
             TestSubject { message: MESSAGE },
             &certificate
@@ -753,7 +753,8 @@ mod tests {
     }
 
     fn test_verify_certificate<V: Variant>() {
-        let (schemes, verifier, _) = setup_signers::<V>(4, 48);
+        let mut rng = test_rng();
+        let (schemes, verifier, _) = setup_signers::<V>(&mut rng, 4);
         let quorum = quorum(schemes.len() as u32) as usize;
 
         let attestations: Vec<_> = schemes
@@ -768,7 +769,7 @@ mod tests {
         let certificate = schemes[0].assemble(attestations).unwrap();
 
         assert!(verifier.verify_certificate::<_, Sha256Digest>(
-            &mut test_rng(),
+            &mut rng,
             NAMESPACE,
             TestSubject { message: MESSAGE },
             &certificate
@@ -783,7 +784,7 @@ mod tests {
 
     fn test_verify_certificate_detects_corruption<V: Variant>() {
         let mut rng = test_rng();
-        let (schemes, verifier, _) = setup_signers::<V>(4, 50);
+        let (schemes, verifier, _) = setup_signers::<V>(&mut rng, 4);
         let quorum = quorum(schemes.len() as u32) as usize;
 
         let attestations: Vec<_> = schemes
@@ -822,7 +823,8 @@ mod tests {
     }
 
     fn test_certificate_codec_roundtrip<V: Variant>() {
-        let (schemes, _, _) = setup_signers::<V>(4, 51);
+        let mut rng = test_rng();
+        let (schemes, _, _) = setup_signers::<V>(&mut rng, 4);
         let quorum = quorum(schemes.len() as u32) as usize;
 
         let attestations: Vec<_> = schemes
@@ -847,7 +849,8 @@ mod tests {
     }
 
     fn test_certificate_rejects_sub_quorum<V: Variant>() {
-        let (schemes, _, _) = setup_signers::<V>(4, 52);
+        let mut rng = test_rng();
+        let (schemes, _, _) = setup_signers::<V>(&mut rng, 4);
         let sub_quorum = 2; // Less than quorum (3)
 
         let attestations: Vec<_> = schemes
@@ -869,7 +872,8 @@ mod tests {
     }
 
     fn test_verify_certificates_batch<V: Variant>() {
-        let (schemes, verifier, _) = setup_signers::<V>(4, 56);
+        let mut rng = test_rng();
+        let (schemes, verifier, _) = setup_signers::<V>(&mut rng, 4);
         let quorum = quorum(schemes.len() as u32) as usize;
 
         let messages = [b"msg1".as_slice(), b"msg2".as_slice(), b"msg3".as_slice()];
@@ -892,7 +896,6 @@ mod tests {
             .zip(&certificates)
             .map(|(msg, cert)| (TestSubject { message: msg }, cert));
 
-        let mut rng = StdRng::seed_from_u64(57);
         assert!(verifier.verify_certificates::<_, Sha256Digest, _>(&mut rng, NAMESPACE, certs_iter));
     }
 
@@ -903,7 +906,8 @@ mod tests {
     }
 
     fn test_verify_certificates_batch_detects_failure<V: Variant>() {
-        let (schemes, verifier, _) = setup_signers::<V>(4, 58);
+        let mut rng = test_rng();
+        let (schemes, verifier, _) = setup_signers::<V>(&mut rng, 4);
         let quorum = quorum(schemes.len() as u32) as usize;
 
         let messages = [b"msg1".as_slice(), b"msg2".as_slice()];
@@ -929,7 +933,6 @@ mod tests {
             .zip(&certificates)
             .map(|(msg, cert)| (TestSubject { message: msg }, cert));
 
-        let mut rng = StdRng::seed_from_u64(59);
         assert!(
             !verifier.verify_certificates::<_, Sha256Digest, _>(&mut rng, NAMESPACE, certs_iter)
         );
@@ -942,7 +945,8 @@ mod tests {
     }
 
     fn test_certificate_verifier<V: Variant>() {
-        let (schemes, _, polynomial) = setup_signers::<V>(4, 60);
+        let mut rng = test_rng();
+        let (schemes, _, polynomial) = setup_signers::<V>(&mut rng, 4);
         let quorum = quorum(schemes.len() as u32) as usize;
 
         let attestations: Vec<_> = schemes
@@ -962,7 +966,7 @@ mod tests {
 
         // Should be able to verify certificates
         assert!(cert_verifier.verify_certificate::<_, Sha256Digest>(
-            &mut test_rng(),
+            &mut rng,
             NAMESPACE,
             TestSubject { message: MESSAGE },
             &certificate
@@ -997,13 +1001,14 @@ mod tests {
     }
 
     fn test_verifier_accepts_votes<V: Variant>() {
-        let (schemes, verifier, _) = setup_signers::<V>(4, 62);
+        let mut rng = test_rng();
+        let (schemes, verifier, _) = setup_signers::<V>(&mut rng, 4);
 
         let vote = schemes[1]
             .sign::<Sha256Digest>(NAMESPACE, TestSubject { message: MESSAGE })
             .unwrap();
         assert!(verifier.verify_attestation::<_, Sha256Digest>(
-            &mut test_rng(),
+            &mut rng,
             NAMESPACE,
             TestSubject { message: MESSAGE },
             &vote
@@ -1017,7 +1022,8 @@ mod tests {
     }
 
     fn test_scheme_clone_and_verifier<V: Variant>() {
-        let (schemes, verifier, _) = setup_signers::<V>(4, 63);
+        let mut rng = test_rng();
+        let (schemes, verifier, _) = setup_signers::<V>(&mut rng, 4);
 
         // Clone a signer
         let signer = schemes[0].clone();
@@ -1044,7 +1050,8 @@ mod tests {
     }
 
     fn certificate_verifier_panics_on_vote<V: Variant>() {
-        let (schemes, _, _) = setup_signers::<V>(4, 37);
+        let mut rng = test_rng();
+        let (schemes, _, _) = setup_signers::<V>(&mut rng, 4);
         let certificate_verifier =
             Scheme::<ed25519::PublicKey, V>::certificate_verifier(*schemes[0].identity());
 
@@ -1054,7 +1061,7 @@ mod tests {
 
         // CertificateVerifier should panic when trying to verify a vote
         certificate_verifier.verify_attestation::<_, Sha256Digest>(
-            &mut test_rng(),
+            &mut rng,
             NAMESPACE,
             TestSubject { message: MESSAGE },
             &vote,
@@ -1115,7 +1122,7 @@ mod tests {
     }
 
     fn signer_polynomial_threshold_must_equal_quorum<V: Variant>() {
-        let mut rng = StdRng::seed_from_u64(7);
+        let mut rng = test_rng();
         let participants = make_participants(&mut rng, 5);
         // Create a polynomial with threshold 4, but quorum of 5 participants is 4
         // so this should succeed. Let's use threshold 2 to make it fail.
@@ -1138,7 +1145,7 @@ mod tests {
     }
 
     fn verifier_polynomial_threshold_must_equal_quorum<V: Variant>() {
-        let mut rng = StdRng::seed_from_u64(7);
+        let mut rng = test_rng();
         let participants = make_participants(&mut rng, 5);
         // Create a polynomial with threshold 2, but quorum of 5 participants is 4
         // quorum(5) = 4, but polynomial.required() = 2, so this should panic
@@ -1159,7 +1166,8 @@ mod tests {
     }
 
     fn certificate_decode_rejects_length_mismatch<V: Variant>() {
-        let (schemes, _, _) = setup_signers::<V>(4, 65);
+        let mut rng = test_rng();
+        let (schemes, _, _) = setup_signers::<V>(&mut rng, 4);
         let quorum = quorum(schemes.len() as u32) as usize;
 
         let attestations: Vec<_> = schemes
@@ -1184,7 +1192,8 @@ mod tests {
     }
 
     fn sign_vote_partial_matches_share<V: Variant>() {
-        let (schemes, _, _) = setup_signers::<V>(4, 66);
+        let mut rng = test_rng();
+        let (schemes, _, _) = setup_signers::<V>(&mut rng, 4);
         let scheme = &schemes[0];
 
         let signature = scheme
--- a/cryptography/src/ed25519/certificate/mod.rs
+++ b/cryptography/src/ed25519/certificate/mod.rs
@@ -656,7 +656,6 @@ mod tests {
     use commonware_codec::{Decode, Encode};
     use commonware_math::algebra::Random;
     use commonware_utils::{ordered::Set, quorum, test_rng, TryCollect};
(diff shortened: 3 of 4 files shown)


consensus/src/aggregation/engine.rs
```diff
@@ -665,7 +665,7 @@ impl<
         }
 
         // Validate signature
-        if !ack.verify(&*scheme, &self.namespace) {
+        if !ack.verify(&mut rand::thread_rng(), &*scheme, &self.namespace) {
```
patrick-ogrady: use context

consensus/src/ordered_broadcast/engine.rs
```diff
@@ -946,7 +946,7 @@ impl<
         }
 
         // Validate the vote signature
-        if !ack.verify(&self.namespace, scheme.as_ref()) {
+        if !ack.verify(&mut rand::thread_rng(), &self.namespace, scheme.as_ref()) {
```
patrick-ogrady: use context

consensus/src/ordered_broadcast/types.rs
```diff
@@ -1611,10 +1612,10 @@ mod tests {
         let epoch = Epoch::new(5);
 
         let ack = Ack::sign(NAMESPACE, &fixture.schemes[0], chunk, epoch).expect("Should sign ack");
-        assert!(ack.verify(NAMESPACE, &fixture.verifier));
+        assert!(ack.verify(&mut rand::thread_rng(), NAMESPACE, &fixture.verifier));
```
patrick-ogrady: use seeded randomness?

consensus/src/aggregation/types.rs
```diff
@@ -486,7 +487,7 @@ mod tests {
         // Verify the restored ack
         assert_eq!(restored_ack.item, item);
         assert_eq!(restored_ack.epoch, Epoch::new(1));
-        assert!(restored_ack.verify(&schemes[0], NAMESPACE));
+        assert!(restored_ack.verify(&mut rand::thread_rng(), &schemes[0], NAMESPACE));
```
patrick-ogrady: use seeded randomness

consensus/src/ordered_broadcast/types.rs
```diff
@@ -1900,7 +1901,7 @@ mod tests {
         .expect("Should sign ack");
 
         // Verification should succeed
-        assert!(ack.verify(NAMESPACE, &fixture.verifier));
+        assert!(ack.verify(&mut rand::thread_rng(), NAMESPACE, &fixture.verifier));
```
patrick-ogrady: use seeded randomness

consensus/src/simplex/scheme/bls12381_threshold.rs
```diff
@@ -499,7 +506,8 @@ impl<P: PublicKey, V: Variant + Send + Sync> certificate::Scheme for Scheme<P, V
 
         let polynomial = self.polynomial();
         let (vote_namespace, vote_message) = vote_namespace_and_message(namespace, &subject);
-        if let Err(errs) = partial_verify_multiple_public_keys::<V, _>(
+        if let Err(errs) = partial_verify_multiple_public_keys::<_, V, _>(
+            rng,
```
patrick-ogrady: not sure we need the rng here?

consensus/src/simplex/scheme/bls12381_threshold.rs
```diff
@@ -849,7 +866,8 @@ mod tests {
                 },
             )
             .unwrap();
-        assert!(scheme.verify_attestation(
+        assert!(scheme.verify_attestation::<_, Sha256Digest>(
+            &mut rand::thread_rng(),
```
patrick-ogrady: use seeded randomness

consensus/src/simplex/types.rs
```diff
@@ -2808,7 +2823,7 @@ mod tests {
         let decoded = ConflictingNotarize::<S, Sha256>::decode(encoded).unwrap();
 
         assert_eq!(conflicting, decoded);
-        assert!(decoded.verify(&fixture.schemes[0], NAMESPACE));
+        assert!(decoded.verify(&mut rand::thread_rng(), &fixture.schemes[0], NAMESPACE));
```
patrick-ogrady: use seeded randomness

cryptography/fuzz/fuzz_targets/common/mod.rs
```diff
@@ -1,5 +1,6 @@
-use crate::Message;
 use arbitrary::Unstructured;
+
+pub type Message = (Option<Vec<u8>>, Vec<u8>);
```
patrick-ogrady: define later

cryptography/src/bls12381/benches/aggregate_verify_multiple_messages.rs
```diff
@@ -5,36 +5,46 @@ use rand::{thread_rng, Rng};
...
         }
-        let msgs = msgs
-            .iter()
-            .map(|msg| (Some(&namespace[..]), msg.as_ref()))
-            .collect::<Vec<_>>();
         for concurrency in [1, 8].into_iter() {
             c.bench_function(
-                &format!("{}/conc={} msgs={}", module_path!(), concurrency, n,),
+                &format!(
+                    "{}::aggregate_verify_multiple_messages/conc={} msgs={}",
```
patrick-ogrady: remove ::...

cryptography/src/bls12381/benches/verify_multiple_messages.rs
```diff
@@ -0,0 +1,57 @@
...
+        let mut msgs: Vec<[u8; 32]> = Vec::with_capacity(n);
+        for _ in 0..n {
+            let mut msg = [0u8; 32];
+            thread_rng().fill(&mut msg);
+            msgs.push(msg);
+        }
+        for concurrency in [1, 8].into_iter() {
+            c.bench_function(
+                &format!(
+                    "{}::verify_multiple_messages/conc={} msgs={}",
```
patrick-ogrady: remove ::...

cryptography/src/bls12381/certificate/multisig/mod.rs
```diff
@@ -606,7 +611,8 @@ mod tests {
         let attestation = scheme
             .sign::<Sha256Digest>(NAMESPACE, TestSubject { message: MESSAGE })
             .unwrap();
-        assert!(scheme.verify_attestation::<Sha256Digest>(
+        assert!(scheme.verify_attestation::<_, Sha256Digest>(
+            &mut rand::thread_rng(),
```
patrick-ogrady: use seeded randomness

cryptography/src/bls12381/certificate/threshold/mod.rs
```diff
@@ -642,7 +641,8 @@ mod tests {
         let attestation = scheme
             .sign::<Sha256Digest>(NAMESPACE, TestSubject { message: MESSAGE })
             .unwrap();
-        assert!(scheme.verify_attestation::<Sha256Digest>(
+        assert!(scheme.verify_attestation::<_, Sha256Digest>(
+            &mut rand::thread_rng(),
```
patrick-ogrady: seeded randomness

cryptography/src/ed25519/certificate/mod.rs
```diff
@@ -688,7 +693,8 @@ mod tests {
         let attestation = scheme
             .sign::<Sha256Digest>(NAMESPACE, TestSubject { message: MESSAGE })
             .unwrap();
-        assert!(scheme.verify_attestation::<Sha256Digest>(
+        assert!(scheme.verify_attestation::<_, Sha256Digest>(
+            &mut rand::thread_rng(),
```
patrick-ogrady: seeded randomness

cryptography/src/bls12381/primitives/ops.rs
```diff
@@ -202,84 +202,71 @@ pub fn partial_verify_message<V: Variant>(
...
 }
 
-/// Aggregates multiple partial signatures into a single signature.
-///
-/// # Warning
-///
-/// This function assumes a group check was already performed on each `signature` and
-/// that each `signature` is unique. If any of these assumptions are violated, an attacker can
-/// exploit this function to verify an incorrect aggregate signature.
-pub fn partial_aggregate_signatures<'a, V, I>(partials: I) -> Option<(u32, V::Signature)>
```
patrick-ogrady, line 212: nit: restore this

cryptography/src/bls12381/primitives/ops.rs
```diff
@@ -543,34 +540,169 @@ where
...
-/// hashed message and public key). If the public key itself is an aggregate of multiple public keys, an attacker can exploit
-/// this optimization to cause this function to return that an aggregate signature is valid when it really isn't.
+/// This function assumes a group check was already performed on `public` and `signature`.
+/// It is not safe to provide an aggregate public key or to provide duplicate messages.
+#[cfg(feature = "std")]
 pub fn aggregate_verify_multiple_messages<'a, V, I>(
     public: &V::Public,
     messages: I,
     signature: &V::Signature,
-    #[cfg_attr(not(feature = "std"), allow(unused_variables))] concurrency: usize,
```
patrick-ogrady, line 563: Just use this approach

cryptography/src/bls12381/primitives/ops.rs
```diff
@@ -985,161 +1145,125 @@ mod tests {
         aggregate_verify_wrong_public_key_count::<MinSig>();
     }
 
-    fn blst_aggregate_verify_multiple_messages<'a, V, I>(
```
patrick-ogrady, line 988: retain blst test

cryptography/src/bls12381/primitives/ops.rs
```diff
@@ -985,161 +1145,125 @@ mod tests {
...
-                    e => Err(e),
-                }
-            }
-            _ => panic!("Unsupported Variant"),
-        }
-    }
-
-    fn aggregate_verify_multiple_messages_correct<V: Variant>() {
-        // Generate signatures
+    fn aggregate_verify_malleability_vulnerability<V: Variant>() {
```
patrick-ogrady: aggregate_verify_fail_on_malleability

patrick-ogrady (commented):
initial review

cryptography/src/bls12381/benches/aggregate_verify_messages.rs
```diff
@@ -5,36 +5,41 @@ use rand::{thread_rng, Rng};
 fn benchmark_aggregate_verify_multiple_messages(c: &mut Criterion) {
     let namespace = b"namespace";
     for n in [2, 10, 100, 1000, 10000].into_iter() {
-        let mut msgs = Vec::with_capacity(n);
+        let mut msgs: Vec<[u8; 32]> = Vec::with_capacity(n);
```
patrick-ogrady: Do we need to specify?

cryptography/src/bls12381/certificate/multisig/mod.rs
```diff
@@ -935,7 +941,7 @@ mod tests {
         certificate.signers = Signers::from(participants_len + 1, signers);
 
         assert!(!verifier.verify_certificate::<_, Sha256Digest>(
-            &mut thread_rng(),
+            &mut StdRng::seed_from_u64(0),
```
patrick-ogrady: just use thread_rng like before

cryptography/src/bls12381/certificate/threshold/mod.rs
```diff
@@ -642,7 +641,8 @@ mod tests {
         let attestation = scheme
             .sign::<Sha256Digest>(NAMESPACE, TestSubject { message: MESSAGE })
             .unwrap();
-        assert!(scheme.verify_attestation::<Sha256Digest>(
+        assert!(scheme.verify_attestation::<_, Sha256Digest>(
+            &mut StdRng::seed_from_u64(0),
```
patrick-ogrady: just use thread_rng like before

cryptography/src/ed25519/certificate/mod.rs
```diff
@@ -843,7 +849,7 @@ mod tests {
 
         // Valid certificate passes
         assert!(verifier.verify_certificate::<_, Sha256Digest>(
-            &mut thread_rng(),
+            &mut StdRng::seed_from_u64(0),
```
patrick-ogrady: just use thread_rng like before

cryptography/src/bls12381/primitives/ops.rs
```diff
@@ -1033,113 +1108,148 @@ mod tests {
...
     #[test]
-    fn test_aggregate_verify_multiple_messages() {
+    fn test_aggregate_verify_multiple_messages_correct() {
         aggregate_verify_multiple_messages_correct::<MinPk>();
         aggregate_verify_multiple_messages_correct::<MinSig>();
     }
 
-    fn aggregate_verify_wrong_messages<V: Variant>() {
-        // Generate signatures
+    fn aggregate_verify_fail_on_malleability<V: Variant>() {
```
patrick-ogrady: nit: verify_multiple_messages_...

consensus/src/simplex/scheme/bls12381_threshold.rs
```diff
@@ -667,24 +685,21 @@ impl<P: PublicKey, V: Variant + Send + Sync> certificate::Scheme for Scheme<P, V
...
-            identity,
-            &messages
-                .iter()
-                .map(|(namespace, message)| (namespace.as_deref(), message.as_ref()))
-                .collect::<Vec<_>>(),
-            &signature,
-            1,
-        )
-        .is_ok()
+        let entries_refs: Vec<_> = entries
```
patrick-ogrady, line 693: nit: add a comment that we care about the correctness of individual signatures so we don't use aggregate verification

consensus/src/ordered_broadcast/types.rs
```diff
@@ -1611,10 +1612,11 @@ mod tests {
         let epoch = Epoch::new(5);
 
         let ack = Ack::sign(NAMESPACE, &fixture.schemes[0], chunk, epoch).expect("Should sign ack");
-        assert!(ack.verify(NAMESPACE, &fixture.verifier));
+        let mut rng = StdRng::seed_from_u64(0);
```
patrick-ogrady: can just use thread_rng like other tests?

cryptography/src/bls12381/primitives/ops/aggregate.rs
```diff
@@ -0,0 +1,439 @@
...
+/// # Warning
+///
+/// This function assumes the caller has performed a group check and collected a proof-of-possession
+/// for all provided `public`. This function assumes a group check was already performed on the
+/// `signature`. It is not safe to provide duplicate public keys.
+pub fn aggregate_verify_multiple_public_keys<'a, V, I>(
+    public: I,
+    namespace: Option<&[u8]>,
+    message: &[u8],
+    signature: &V::Signature,
```
patrick-ogrady: Add a type for aggregate signature?

cryptography/src/bls12381/primitives/ops/aggregate.rs
```diff
@@ -0,0 +1,439 @@
...
+/// that were aggregated from different messages. An attacker can redistribute
+/// signature components between messages while keeping the aggregate unchanged.
+/// Use [`batch::verify_multiple_messages`](super::batch::verify_multiple_messages) instead when signatures are provided individually.
+///
+/// This function assumes a group check was already performed on `public` and `signature`.
+/// It is not safe to provide an aggregate public key or to provide duplicate messages.
+pub fn aggregate_verify_multiple_messages<'a, V, I>(
+    public: &V::Public,
+    messages: I,
+    signature: &V::Signature,
```
patrick-ogrady: nit: use type?

cryptography/src/bls12381/primitives/ops/core.rs
patrick-ogrady, line 1: Move these implementations into `mod.rs`

cryptography/src/bls12381/primitives/ops/mod.rs
```diff
@@ -0,0 +1,41 @@
...
+//! the domain separation tag is `BLS_POP_BLS12381G2_XMD:SHA-256_SSWU_RO_POP_`. For signatures over other messages, the
+//! domain separation tag is `BLS_SIG_BLS12381G2_XMD:SHA-256_SSWU_RO_POP_`. You can read more about DSTs [here](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-bls-signature-05#section-4.2).
+//!
+//! # Batch vs Aggregate Verification
+//!
+//! Use [`batch`] when you need to ensure each individual signature is valid. Use [`aggregate`]
+//! when you only need to verify that the aggregate is valid (more efficient, but an attacker
+//! could redistribute signature components between signers while keeping the aggregate unchanged).
+//! Batch verification uses random scalar weights internally to prevent this attack.
+
```
patrick-ogrady, line 20: remove all this re-exporting (and remove `aggregate_`, `partial_`, `threshold_signature_` from start of functions...just use paths)

cryptography/src/bls12381/certificate/multisig/mod.rs
```diff
@@ -502,7 +493,7 @@ mod macros {
                     certificates: I,
                 ) -> bool
                 where
-                    R: rand::Rng + rand::CryptoRng,
+                    R: rand_core::CryptoRngCore,
```
patrick-ogrady, line 496: This change to use CryptoRngCore is not necessary and we could revert

cryptography/src/bls12381/certificate/multisig/mod.rs
```diff
@@ -606,7 +597,8 @@ mod tests {
         let attestation = scheme
             .sign::<Sha256Digest>(NAMESPACE, TestSubject { message: MESSAGE })
             .unwrap();
-        assert!(scheme.verify_attestation::<Sha256Digest>(
+        assert!(scheme.verify_attestation::<_, Sha256Digest>(
+            &mut thread_rng(),
```
patrick-ogrady: use thread_rng (altho not deterministic) because other tests do.

cryptography/src/bls12381/primitives/ops/aggregate.rs
```diff
@@ -0,0 +1,546 @@
...
+        );
+
+        let valid_entries = vec![(None, msg1, sig1), (None, msg2, sig2)];
+        verify_multiple_messages::<_, V, _>(&mut thread_rng(), &public, &valid_entries, 1)
+            .expect("secure function should accept valid signatures");
+    }
+
+    #[test]
+    fn test_aggregate_verify_fail_on_malleability() {
+        aggregate_verify_fail_on_malleability::<MinPk>();
```
patrick-ogrady: This should probably live in batch?

cryptography/src/bls12381/primitives/ops/mod.rs
```diff
@@ -373,4 +373,59 @@ mod tests {
...
+            "secure function should reject forged signatures"
+        );
+
+        let valid_entries = vec![(None, msg1, sig1), (None, msg2, sig2)];
+        batch::verify_multiple_messages::<_, V, _>(&mut rng, &public, &valid_entries, 1)
+            .expect("secure function should accept valid signatures");
+    }
+
+    #[test]
+    fn test_aggregate_verify_fail_on_malleability() {
```
patrick-ogrady: nit: improve naming

cryptography/src/bls12381/primitives/ops/aggregate.rs
```diff
@@ -0,0 +1,490 @@
...
+
+/// Combines multiple public keys into an aggregate public key.
+///
+/// # Warning
+///
+/// This function assumes a group check was already performed on all `public_keys`,
+/// that each `public_key` is unique, and that the caller has a Proof-of-Possession (PoP)
+/// for each `public_key`. If any of these assumptions are violated, an attacker can
+/// exploit this function to verify an incorrect aggregate signature.
+pub fn combine_public_keys<'a, V, I>(public_keys: I) -> PublicKey<V>
```
cronokirby, line 192: What if we required passing in the requisite proofs of possession here?

cryptography/src/bls12381/primitives/ops/aggregate.rs
```diff
@@ -0,0 +1,490 @@
...
+pub fn combine_signatures<'a, V, I>(signatures: I) -> Signature<V>
+where
+    V: Variant,
+    I: IntoIterator<Item = &'a V::Signature>,
+    V::Signature: 'a,
+{
+    let mut s = V::Signature::zero();
+    for sig in signatures {
+        s += sig;
+    }
```
cronokirby, line 221: It wouldn't be that expensive to check the uniqueness, and panic.

cryptography/src/bls12381/primitives/ops/aggregate.rs
cronokirby: Similarly, we could also easily check the uniqueness here, and panic.

cryptography/src/bls12381/primitives/ops/aggregate.rs
```diff
@@ -0,0 +1,490 @@
...
+pub struct Signature<V: Variant>(V::Signature);
+
+impl<V: Variant> Signature<V> {
+    /// Returns the inner signature value.
+    pub const fn inner(&self) -> &V::Signature {
+        &self.0
+    }
+
+    /// Creates a zero aggregate signature.
+    pub fn zero() -> Self {
```
cronokirby: Does this need to be pub?

cryptography/src/bls12381/primitives/ops/aggregate.rs
```diff
@@ -0,0 +1,490 @@
...
+/// An aggregated signature from multiple individual signatures.
+///
+/// This type is returned by [`combine_signatures`] and ensures that
+/// aggregated signatures are not confused with individual signatures.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct Signature<V: Variant>(V::Signature);
+
+impl<V: Variant> Signature<V> {
+    /// Returns the inner signature value.
+    pub const fn inner(&self) -> &V::Signature {
```
cronokirby: Does this need to be pub?

cryptography/src/bls12381/primitives/ops/aggregate.rs
```diff
@@ -0,0 +1,490 @@
...
+/// An aggregated public key from multiple individual public keys.
+///
+/// This type is returned by [`combine_public_keys`] and ensures that
+/// aggregated public keys are not confused with individual public keys.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct PublicKey<V: Variant>(V::Public);
+
+impl<V: Variant> PublicKey<V> {
+    /// Returns the inner public key value.
+    pub const fn inner(&self) -> &V::Public {
```
cronokirby: Does this need to be pub?

cryptography/src/bls12381/primitives/ops/aggregate.rs
```diff
@@ -0,0 +1,490 @@
...
+/// This function assumes a group check was already performed on each `signature` and
+/// that each `signature` is unique. If any of these assumptions are violated, an attacker can
+/// exploit this function to verify an incorrect aggregate signature.
+pub fn combine_signatures<'a, V, I>(signatures: I) -> Signature<V>
+where
+    V: Variant,
+    I: IntoIterator<Item = &'a V::Signature>,
+    V::Signature: 'a,
+{
+    let mut s = V::Signature::zero();
```
cronokirby: Should we use the `zero` method on `Signature`? If not, maybe we should just remove the method entirely.

cryptography/src/bls12381/primitives/ops/aggregate.rs
```diff
@@ -0,0 +1,490 @@
...
+    //
+    // We can take advantage of the bilinearity property of pairings to aggregate public keys
+    // that have all signed the same message (as long as all public keys are unique).
+    let agg_public = combine_public_keys::<V, _>(public);
+
+    // Compute the hash of the message
+    let hm = namespace.map_or_else(
+        || hash_message::<V>(V::MESSAGE, message),
+        |ns| hash_message_namespace::<V>(V::MESSAGE, ns, message),
+    );
```
cronokirby: We should probably create a utility for this at this point.

cryptography/src/bls12381/primitives/ops/aggregate.rs
```diff
@@ -0,0 +1,490 @@
...
+/// # Warning
+///
+/// This function is vulnerable to signature malleability when used with signatures
+/// that were aggregated from different messages. An attacker can redistribute
+/// signature components between messages while keeping the aggregate unchanged.
+/// Use [`batch::verify_multiple_messages`](super::batch::verify_multiple_messages) instead when signatures are provided individually.
+///
+/// This function assumes a group check was already performed on `public` and `signature`.
+/// It is not safe to provide an aggregate public key or to provide duplicate messages.
+pub fn verify_multiple_messages<'a, V, I>(
```
cronokirby: We should remove this method completely. Signature aggregation should only be used in the case where you want to aggregate signatures from separate entities, not as "batched verification with tricky caveats"

cryptography/src/bls12381/primitives/ops/aggregate.rs
```diff
@@ -0,0 +1,490 @@
...
+                    sum
+                })
+        })
+    };
+
+    V::verify(public, &hm_sum, signature.inner())
+}
+
+/// Computes the sum over the hash of each message.
+fn compute_hm_sum<'a, V, I>(messages: I) -> V::Signature
```
cronokirby: I think it would be more appropriate to create a utility for hashing messages, and just do away with this method.

cryptography/src/bls12381/primitives/ops/aggregate.rs
```diff
@@ -0,0 +1,490 @@
...
+}
+
+/// Verifies the aggregate signature over a single message from multiple public keys.
+///
+/// # Warning
+///
+/// This function assumes the caller has performed a group check and collected a proof-of-possession
+/// for all provided `public`. This function assumes a group check was already performed on the
+/// `signature`. It is not safe to provide duplicate public keys.
+pub fn verify_multiple_public_keys<'a, V, I>(
```
cronokirby:
```suggestion
pub fn verify<'a, V, I>(
```

Given that this is namespaced, we can use a much simpler name.

cronokirby: A more advanced idea:

Have the `PublicKey` type in this module (an aggregate public key) implement the `PublicKey` trait, with an associated type of `Signature` (an aggregate signature)

cryptography/src/bls12381/primitives/ops/batch.rs
```diff
@@ -0,0 +1,262 @@
...
+/// Verifies multiple signatures over multiple unique messages from a single public key,
+/// ensuring each individual signature is valid.
+///
+/// Each entry is a tuple of (namespace, message, signature).
+///
+/// # Warning
+///
+/// This function assumes a group check was already performed on `public` and each `signature`.
+/// It is not safe to provide an aggregate public key or to provide duplicate messages.
+pub fn verify_multiple_messages<'a, R, V, I>(
```
cronokirby: It actually **is** safe to provide duplicate messages, because of the random combinations, and we should update the documentation to reflect that, along with an explanation demonstrating that we considered this.

cryptography/src/bls12381/primitives/ops/batch.rs
```diff
@@ -0,0 +1,262 @@
...
+    while let Some((start, end)) = stack.pop() {
+        let slice = &entries[start..end];
+        if slice.is_empty() {
+            continue;
+        }
+
+        // Generate random scalars for each signature in this slice
+        let scalars: Vec<Scalar> = (0..slice.len())
+            .map(|_| Scalar::random(&mut *rng))
+            .collect();
```
cronokirby: We can do this once outside of the loop, and re-use the weights at each iteration.

cronokirby: In fact, what you want is to avoid doing the scalar multiplications by the weights more than necessary.

cryptography/src/bls12381/primitives/ops/threshold.rs
```diff
@@ -0,0 +1,1247 @@
...
+        .map(|idx| pending[idx].1)
+        .collect()
+}
+
+/// Attempts to verify multiple [PartialSignature]s over the same message as a single
+/// aggregate signature (or returns any invalid signature found).
+///
+/// This function applies random scalar weights to prevent signature malleability attacks
+/// where an attacker could redistribute signature components while keeping the aggregate
+/// unchanged.
```
cronokirby: I would the language here and above to focus less on a particular edge case, and what exactly randomness is used for, to instead say something like "randomness is needed to ensure that the batch verification process returns the same result as checking each signature individually" which captures this case, and many more, while also not burdening the consumer with cognitive load unnecessarily.

cryptography/src/bls12381/primitives/ops/threshold.rs
```diff
@@ -0,0 +1,1247 @@
...
+}
+
+/// Aggregates multiple partial signatures into a single signature.
+///
+/// # Warning
+///
+/// This function assumes a group check was already performed on each `signature` and
+/// that each `signature` is unique. If any of these assumptions are violated, an attacker can
+/// exploit this function to verify an incorrect aggregate signature.
+pub fn aggregate_signatures<'a, V, I>(partials: I) -> Option<(u32, V::Signature)>
```
cronokirby: Is this method still usable, given that `aggregate` now has a wrapper type?

cronokirby (dismissed):
I think splitting the operations up like this makes sense long term, but I think we can go a bit further in simplifying things.

cryptography/src/bls12381/primitives/ops/threshold.rs
patrick-ogrady: We should remove this...this is a remnant of the days when we wanted to combine partial signatures to reward participation (before we realized they could be derived).

cryptography/src/bls12381/primitives/ops/threshold.rs
patrick-ogrady: Yeah, this is some LLM junk it felt the liberty to insert.

cryptography/src/bls12381/primitives/ops/aggregate.rs
patrick-ogrady: This function was added to support #1353 (compressed nullifications). While we have been delayed in implementing it, I do think it is indicative of there being usefulness to this method and wish to keep it for now.

I'll make sure to update the documentation.

cryptography/src/bls12381/primitives/ops/aggregate.rs
patrick-ogrady: Prefer leaving this to a follow-up, just to get the rest of this (now large) refactor across.

fwiw I prefer not requiring PoPs here because it may be annoying to fetch those from onchain just to call this (I suppose we could cache heavily but we won't have the time to actually verify them).

cryptography/src/bls12381/primitives/ops/aggregate.rs
patrick-ogrady: See comment about punting this to a future PR.

cryptography/src/bls12381/primitives/ops/aggregate.rs
patrick-ogrady: We should actually make this (and the type) private. The recommended flow is to use a bitvec to represent the signers and for the verifier to provide all public keys (to combine during the verify call).

I suppose we could keep to allow for caching this combination (could pre-compute all combos) but that seems like future work (at best).

consensus/src/ordered_broadcast/types.rs
patrick-ogrady: reuse rng

consensus/src/ordered_broadcast/types.rs
```diff
@@ -1605,16 +1597,16 @@ mod tests {
...
     {
-        let fixture = setup(4, fixture);
+        let fixture = fixture(&mut test_rng(), 4);
         let public_key = sample_scheme(0).public_key();
         let chunk = Chunk::new(public_key, 42, sample_digest(1));
         let epoch = Epoch::new(5);
 
         let ack = Ack::sign(NAMESPACE, &fixture.schemes[0], chunk, epoch).expect("Should sign ack");
-        assert!(ack.verify(NAMESPACE, &fixture.verifier));
+        assert!(ack.verify(&mut test_rng(), NAMESPACE, &fixture.verifier));
```
patrick-ogrady: reuse rng

consensus/src/ordered_broadcast/types.rs
```diff
@@ -1946,10 +1938,10 @@ mod tests {
                 .expect("Should sign ack");
 
         // Verification should succeed with correct verifier
-        assert!(ack.verify(NAMESPACE, &fixture.verifier));
+        assert!(ack.verify(&mut test_rng(), NAMESPACE, &fixture.verifier));
```
patrick-ogrady: reuse RNG?

consensus/src/simplex/scheme/bls12381_threshold.rs
```diff
@@ -418,14 +414,20 @@ impl<P: PublicKey, V: Variant + Send + Sync> certificate::Scheme for Scheme<P, V
         let share = self.share()?;
 
         let (vote_namespace, vote_message) = vote_namespace_and_message(namespace, &subject);
-        let vote_signature =
-            partial_sign_message::<V>(share, Some(vote_namespace.as_ref()), vote_message.as_ref())
-                .value;
+        let vote_signature = threshold::sign_message::<V>(
+            share,
+            Some(vote_namespace.as_ref()),
```
patrick-ogrady: require namespace?

consensus/src/simplex/scheme/bls12381_threshold.rs
```diff
@@ -438,45 +440,53 @@ impl<P: PublicKey, V: Variant + Send + Sync> certificate::Scheme for Scheme<P, V
...
-
-        aggregate_verify_multiple_messages::<V, _>(
+        batch::verify_messages::<_, V, _>(
+            rng,
             &evaluated,
             &[
-                (Some(vote_namespace.as_ref()), vote_message.as_ref()),
-                (Some(seed_namespace.as_ref()), seed_message.as_ref()),
+                (
+                    Some(vote_namespace.as_ref()),
```
patrick-ogrady: require namespace?

consensus/src/simplex/types.rs
```diff
@@ -3075,9 +3077,9 @@ mod tests {
         let finalize = Finalize::sign(&fixture.schemes[0], NAMESPACE, proposal).unwrap();
         let conflict = NullifyFinalize::new(nullify, finalize);
 
-        assert!(conflict.verify(&fixture.schemes[0], NAMESPACE));
-        assert!(!conflict.verify(&fixture.schemes[0], b"wrong_namespace"));
-        assert!(!conflict.verify(&wrong_fixture.verifier, NAMESPACE));
+        assert!(conflict.verify(&mut test_rng(), &fixture.schemes[0], NAMESPACE));
```
patrick-ogrady: reuse rng?

cryptography/fuzz/fuzz_targets/bls12381_primitive_operations.rs
```diff
@@ -159,8 +159,8 @@ enum FuzzOperation {
         signature: G1,
     },
 
-    // Partial signature operations - simplified
-    PartialSignMessage {
+    // Sign message with share
+    SignMessage {
```
patrick-ogrady: ThresholdSignMessage

cryptography/fuzz/fuzz_targets/bls12381_primitive_operations.rs
```diff
@@ -176,18 +176,16 @@ enum FuzzOperation {
         use_g1: bool,
     },
 
-    // Simple aggregate operations
-    AggregatePublicKeysG1 {
-        keys: Vec<G1>,
+    // Verify messages (individual signatures)
+    VerifyMessagesMinPk {
```
patrick-ogrady: BatchVerifyMessages*

cryptography/fuzz/fuzz_targets/bls12381_primitive_operations.rs
```diff
@@ -176,18 +176,16 @@ enum FuzzOperation {
...
     },
-    AggregatePublicKeysG2 {
-        keys: Vec<G2>,
-    },
-    AggregateSignaturesG1 {
-        sigs: Vec<G1>,
-    },
-    AggregateSignaturesG2 {
-        sigs: Vec<G2>,
+    VerifyMessagesMinSig {
```
patrick-ogrady: BatchVerifyMessages*

cryptography/fuzz/fuzz_targets/bls12381_primitive_operations.rs
patrick-ogrady: move to threshold_operations

cryptography/fuzz/fuzz_targets/bls12381_primitive_operations.rs
patrick-ogrady: move to bls12381_batch_operations

cryptography/src/bls12381/benches/combine_public_keys.rs
patrick-ogrady, line 1: remove

cryptography/src/bls12381/benches/combine_signatures.rs
patrick-ogrady, line 1: combine_signatures

cryptography/src/bls12381/benches/scheme_verify_messages.rs
```diff
@@ -4,7 +4,7 @@ use criterion::{criterion_group, BatchSize, Criterion};
 use rand::{thread_rng, Rng};
 use std::hint::black_box;
 
-fn benchmark_batch_verify_multiple_messages(c: &mut Criterion) {
+fn benchmark_scheme_verify_messages(c: &mut Criterion) {
```
patrick-ogrady: scheme_batch_verify_multiple_messages

cryptography/src/bls12381/benches/scheme_verify_public_keys.rs
```diff
@@ -4,7 +4,7 @@ use criterion::{criterion_group, BatchSize, Criterion};
 use rand::{thread_rng, Rng};
 use std::hint::black_box;
 
-fn benchmark_batch_verify_multiple_public_keys(c: &mut Criterion) {
+fn benchmark_scheme_verify_public_keys(c: &mut Criterion) {
```
patrick-ogrady: scheme_batch_verify_multiple_public_keys

cryptography/src/bls12381/benches/threshold_batch_verify_public_keys_precomputed.rs
```diff
@@ -61,11 +61,16 @@ fn benchmark_partial_verify_multiple_public_keys_precomputed(c: &mut Criterion)
...
-                                black_box(primitives::ops::partial_verify_multiple_public_keys::<
+                                black_box(primitives::ops::threshold::batch_verify_public_keys::<
+                                    _,
                                     MinSig,
                                     _,
                                 >(
-                                    &polynomial, Some(namespace), msg, &signatures
+                                    &mut rng,
+                                    &polynomial,
+                                    Some(namespace),
```
patrick-ogrady: mandatory namespace?

cryptography/src/bls12381/benches/threshold_batch_verify_public_keys_precomputed.rs
```diff
@@ -11,7 +11,7 @@ use criterion::{criterion_group, BatchSize, Criterion};
 use rand::{rngs::StdRng, seq::SliceRandom, SeedableRng};
 use std::hint::black_box;
 
-fn benchmark_partial_verify_multiple_public_keys_precomputed(c: &mut Criterion) {
+fn benchmark_threshold_batch_verify_public_keys_precomputed(c: &mut Criterion) {
```
patrick-ogrady: remove _precomputed?

cryptography/src/bls12381/certificate/multisig/mod.rs
```diff
@@ -785,17 +777,17 @@ mod tests {
...
             &certificate
         ));
 
         // Corrupted certificate fails
         let mut corrupted = certificate;
-        corrupted.signature = V::Signature::zero();
+        corrupted.signature = aggregate::Signature::zero();
         assert!(!verifier.verify_certificate::<_, Sha256Digest>(
-            &mut thread_rng(),
+            &mut test_rng(),
```
patrick-ogrady: reuse rng?

cryptography/src/bls12381/primitives/ops/aggregate.rs
```diff
@@ -0,0 +1,463 @@
...
+use commonware_math::algebra::Additive;
+#[cfg(feature = "std")]
+use rayon::{prelude::*, ThreadPoolBuilder};
+
+/// An aggregated public key from multiple individual public keys.
+///
+/// This type is returned by [`combine_public_keys`] and ensures that
+/// aggregated public keys are not confused with individual public keys.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct PublicKey<V: Variant>(V::Public);
```
patrick-ogrady, line 28: Make private (+ remove encoding)

cryptography/src/bls12381/primitives/ops/batch.rs
```diff
@@ -0,0 +1,268 @@
...
+        );
+
+        // Batch verification accepts valid signatures
+        let valid_entries = vec![(None, msg1, sig1), (None, msg2, sig2)];
+        verify_messages::<_, V, _>(&mut rng, &public, &valid_entries, 1)
+            .expect("batch verification should accept valid signatures");
+    }
+
+    #[test]
+    fn test_resists_signature_redistribution() {
```
patrick-ogrady: test_resists_signature_malleability

cryptography/src/bls12381/primitives/ops/threshold.rs
```diff
@@ -0,0 +1,1205 @@
...
+        ops::verify_message::<V>(sharing.public(), None, b"payload2", &sig_2).unwrap();
+    }
+
+    #[test]
+    fn test_recover_multiple() {
+        recover_multiple_test::<MinPk>();
+        recover_multiple_test::<MinSig>();
+    }
+
+    fn partial_aggregate_signature_correct<V: Variant>() {
```
patrick-ogrady: nit: recover

cryptography/src/bls12381/primitives/ops/threshold.rs
```diff
@@ -0,0 +1,1205 @@
...
+        ops::verify_message::<V>(sharing.public(), namespace, msg, &threshold_sig).unwrap();
+    }
+
+    #[test]
+    fn test_partial_aggregate_signature_correct() {
+        partial_aggregate_signature_correct::<MinPk>();
+        partial_aggregate_signature_correct::<MinSig>();
+    }
+
+    fn partial_aggregate_signature_bad_namespace<V: Variant>() {
```
patrick-ogrady: nit: recover

cryptography/src/bls12381/primitives/ops/threshold.rs
```diff
@@ -0,0 +1,1205 @@
...
+        ));
+    }
+
+    #[test]
+    fn test_partial_aggregate_signature_insufficient() {
+        partial_aggregate_signature_insufficient::<MinPk>();
+        partial_aggregate_signature_insufficient::<MinSig>();
+    }
+
+    fn partial_aggregate_signature_bad_share<V: Variant>() {
```
patrick-ogrady: nit: recover

cryptography/src/bls12381/primitives/ops/threshold.rs
```diff
@@ -0,0 +1,1205 @@
...
+        }
+    }
+
+    #[test]
+    fn test_threshold_derive_missing_partials() {
+        threshold_derive_missing_partials::<MinPk>();
+        threshold_derive_missing_partials::<MinSig>();
+    }
+
+    fn batch_verify_public_keys_fail_on_malleability<V: Variant>() {
```
patrick-ogrady: batch_verify_resists_malleability

cryptography/src/bls12381/primitives/ops/threshold.rs
```diff
@@ -0,0 +1,1205 @@
...
+        .expect("secure function should accept valid partial signatures");
+    }
+
+    #[test]
+    fn test_batch_verify_public_keys_fail_on_malleability() {
+        batch_verify_public_keys_fail_on_malleability::<MinPk>();
+        batch_verify_public_keys_fail_on_malleability::<MinSig>();
+    }
+
+    fn batch_verify_messages_fail_on_malleability<V: Variant>() {
```
patrick-ogrady: rejects_malleability

cryptography/conformance.toml
```diff
@@ -58,6 +58,14 @@ hash = "d112d774a4c9b012b53e44dc2a3487a0e2f87cd7b14fb8433ec7cea4997f440a"
 n_cases = 65536
 hash = "169550a1597994aad7ba634091b89633424be84f6753fe95542d2f3fd4108b9f"
 
+["commonware_cryptography::bls12381::primitives::ops::aggregate::tests::conformance::CodecConformance<PublicKey<MinSig>>"]
```
patrick-ogrady, line 65: remove

cryptography/src/bls12381/primitives/ops/aggregate.rs
patrick-ogrady: Alternatively, we could accept this aggregate::PublicKey type in the verify function to give flexibility to how it is constructed (that would be more consistent with how we accept an aggregate signature.

cryptography/src/bls12381/primitives/ops/aggregate.rs
patrick-ogrady: I think this something we can come back to.

cryptography/src/bls12381/primitives/ops/aggregate.rs
patrick-ogrady: Cleaned this up a bit. Not in love with it but its better.

cryptography/src/bls12381/primitives/variant.rs
```diff
@@ -157,14 +157,9 @@ impl Variant for MinPk {
...
-        // Generate random non-zero scalars.
+        // Generate random scalars.
         let scalars: Vec<Scalar> = (0..publics.len())
-            .map(|_| loop {
-                let scalar = Scalar::random(&mut *rng);
-                if scalar != Scalar::zero() {
-                    return scalar;
-                }
-            })
+            .map(|_| Scalar::random(&mut *rng))
```
andresilva, line 162: I think we need this to be non-zero?

cryptography/src/bls12381/primitives/variant.rs
```diff
@@ -308,14 +303,9 @@ impl Variant for MinSig {
...
```
andresilva, line 308: Same as above?

cryptography/src/bls12381/primitives/ops/batch.rs
```diff
@@ -0,0 +1,272 @@
...
+{
+    if entries.is_empty() {
+        return Vec::new();
+    }
+
+    let hm = hash_with_namespace::<V>(V::MESSAGE, namespace, message);
+
+    // Generate random scalars once for all entries
+    let scalars: Vec<Scalar> = (0..entries.len())
+        .map(|_| Scalar::random(&mut *rng))
```
andresilva, line 61: Needs to be non-zero?

cryptography/src/bls12381/primitives/ops/batch.rs
```diff
@@ -0,0 +1,272 @@
...
+{
+    let entries: Vec<_> = entries.into_iter().collect();
+
+    if entries.is_empty() {
+        return Ok(());
+    }
+
+    // Generate random scalars for each message/signature pair
+    let scalars: Vec<Scalar> = (0..entries.len())
+        .map(|_| Scalar::random(&mut *rng))
```
andresilva, line 136: Ensure non-zero?

cryptography/src/bls12381/primitives/variant.rs
andresilva, line 1: Unrelated to this PR but maybe we can add codec conformance test to `PartialSignature`?

cryptography/src/bls12381/primitives/group.rs
andresilva, line 1: Does blst guarantee non-zero from `blst_keygen`? If so then my comments are moot, but please add a comment on this in `Scalar::from_bytes`.

patrick-ogrady: It does guarantee non-zero. Added comments.

cryptography/src/bls12381/primitives/group.rs
```diff
@@ -214,7 +214,11 @@ pub const PRIVATE_KEY_LENGTH: usize = SCALAR_LENGTH;
 
 impl Scalar {
     fn from_bytes(mut ikm: [u8; 64]) -> Self {
-        // Generate a scalar from the randomly populated buffer
+        // Generate a scalar from the randomly populated buffer using IETF BLS KeyGen.
+        //
+        // Per IETF draft-irtf-cfrg-bls-signature-05 Section 2.3, KeyGen is guaranteed
```
patrick-ogrady: nit: don't invoke IETF here, just say it is non-zero

cryptography/src/bls12381/primitives/ops/batch.rs
```diff
@@ -0,0 +1,272 @@
+//! Batch verification for BLS12-381 signatures.
+//!
+//! This module provides batch verification functions that ensure each individual
+//! signature is valid (not just the aggregate). Use [`aggregate`](super::aggregate) instead
+//! if you only need to verify aggregate validity (more efficient).
+//!
+//! # How It Works
+//!
+//! These functions apply random scalar weights to ensure that the batch verification
```
andresilva: Could we expand with a one‑paragraph intuition on soundness? Something like "Write each signature as `sig_i = sig_i_valid + e_i`, batch verification cancels the valid parts and accepts only if `sum(r_i * e_i) == 0`." That makes it explicit that batch verification is probabilistic (false accept ~1/2^255 per batch) and relies on unpredictable weights. It helps explain why randomness is required and why the source of randomness needs to be secure (i.e. attacker can't predict the random weights).

cryptography/src/bls12381/primitives/ops/batch.rs
```diff
@@ -0,0 +1,276 @@
...
+#[cfg(not(feature = "std"))]
+use alloc::{vec, vec::Vec};
+use commonware_math::algebra::{Additive, Random, Space};
+use rand_core::CryptoRngCore;
+
+/// Verifies multiple signatures over the same message from different public keys,
+/// ensuring each individual signature is valid.
+///
+/// Returns the indices of any invalid signatures found. Uses bisection internally
+/// to efficiently identify which signatures are invalid.
```
cronokirby:
```suggestion
/// Returns the indices of any invalid signatures found.
```

The fact that bisection is the strategy used doesn't matter.

cryptography/src/bls12381/primitives/ops/aggregate.rs
```diff
@@ -0,0 +1,551 @@
+//! Aggregation operations for BLS12-381 signatures.
+//!
+//! This module provides functions for aggregating public keys and signatures,
+//! as well as verifying aggregate signatures.
+//!
+//! # Security Considerations
+//!
+//! Aggregate operations ensure the aggregate is valid, but not that the individual elements are valid.
+//! Use [`batch`](super::batch) when you need to ensure each individual signature is valid.
```
cronokirby, line 9: I think we should just eliminate this module completely. `certificate` provides the same functionality in a much better and more robust way, and the level of abstraction here is needlessly low.

cryptography/src/bls12381/primitives/ops/batch.rs
```diff
@@ -0,0 +1,276 @@
...
+//!
+//! # How It Works
+//!
+//! These functions apply random scalar weights to each signature before internally performing
+//! [`aggregate`](super::aggregate) verification. Without weights, an attacker could forge invalid
+//! signatures that cancel out when aggregated (e.g., one signature "too high" and another "too low"
+//! by the same amount). With random weights `r_i`, the errors must satisfy `sum(r_i * err_i) = 0`,
+//! which requires predicting the weights before they're generated (probability ~1/2^255 per invalid
+//! signature). Note, the weights must be unpredictable to the attacker for this to work (i.e. they
+//! must be generated securely).
```
cronokirby, line 15: This is useful for readers of the source code, but not for readers of the documentation.

cryptography/src/bls12381/primitives/ops/batch.rs
```diff
@@ -0,0 +1,276 @@
+//! Batch verification for BLS12-381 signatures.
```
cronokirby, line 1: I think the naming combination `verify_{public_keys,messages}` is not ideal: they emphasize the term which is _not_ salient. The limitation of each method is more salient than the what needs to be passed in (which, in the first case, is not "public keys", but rather "public keys paired with signatures".

I would suggest: `verify_same_message`, `verify_same_signer`.

cryptography/src/bls12381/primitives/ops/batch.rs
```diff
@@ -0,0 +1,276 @@
...
+
+/// Verifies multiple signatures over multiple messages from a single public key,
+/// ensuring each individual signature is valid.
+///
+/// Each entry is a tuple of (namespace, message, signature).
+///
+/// # Warning
+///
+/// This function assumes a group check was already performed on `public` and each `signature`.
+/// It is not safe to provide an aggregate public key. Duplicate messages are safe because
```
cronokirby:
```suggestion
/// Duplicate messages are safe to provide because
```

Not only is it not safe, it also doesn't type check.

cryptography/src/bls12381/primitives/ops/aggregate.rs
cronokirby: In any case, I think the current API provides too many moving parts, which reflect the operations that happen, but not the semantics you want.

Conceptually, what we're trying to do is to _compress_ a list of signatures into a single artifact. We could just as well do this by simply defining this to be an actual list.

I think a better API would be:

```
AggregateSignature::new(&[(PublicKey, Signature]) -> Self;

AggregateSignature::verify(&self, &[PublicKey]) -> bool;
```

the surface is much simpler, and provides what you need for certificates, rather than allowing odd operations.

(You also don't need combine_messages at all, for which the batch API is more appropriate).

cryptography/src/bls12381/primitives/ops/batch.rs
patrick-ogrady: There are many places in our public documentation where we take the time to go into detail. I think its worth doing here.

cryptography/src/bls12381/primitives/ops/batch.rs
patrick-ogrady: Fair enough

cryptography/src/bls12381/primitives/ops/aggregate.rs
```diff
@@ -0,0 +1,551 @@
...
+/// # Precomputed Combined Message
+///
+/// Instead of requiring all messages that participated in the aggregate signature (and generating
+/// the combined message on-demand), this function accepts a precomputed combined message to allow
+/// the caller to cache previous constructions and/or perform parallel combination.
+///
+/// # Warning
+///
+/// This function assumes a group check was already performed on `public` and `signature`.
+pub fn verify_messages<V: Variant>(
```
cronokirby: I think we don't want to keep this method around, because it fails to guarantee that each individual signature is valid. (Unlike the other one, where the proofs of possession guarantee that)

cryptography/src/bls12381/primitives/ops/aggregate.rs
patrick-ogrady: Future work: https://github.com/commonwarexyz/monorepo/issues/2676

cryptography/src/bls12381/primitives/ops/aggregate.rs
cronokirby: > Prefer leaving this to a follow-up, just to get the rest of this (now large) refactor across.
> 
> fwiw I prefer not requiring PoPs here because it may be annoying to fetch those from onchain just to call this (I suppose we could cache heavily but we won't have the time to actually verify them).

My thinking is that you would create a type specifically for public keys whose proof of possession you've verified, and then would store that and retrieve that from disk etc. etc.

cronokirby: > Prefer leaving this to a follow-up, just to get the rest of this (now large) refactor across.
> 
> fwiw I prefer not requiring PoPs here because it may be annoying to fetch those from onchain just to call this (I suppose we could cache heavily but we won't have the time to actually verify them).

My thinking is that you would create a type specifically for public keys whose proof of possession you've verified, and then would store that and retrieve that from disk etc. etc.

cryptography/src/bls12381/primitives/ops/batch.rs
patrick-ogrady: I think its worth including (because it can still be constructed by abuse of group ops).

patrick-ogrady: but will clean up wording

cryptography/src/bls12381/primitives/ops/aggregate.rs
patrick-ogrady: Ah, that's nice. I guess I've "assumed" you've done that already but using a type is great.

patrick-ogrady: I'll add to #2676

cryptography/src/bls12381/primitives/ops/batch.rs
patrick-ogrady: Opted to actually remove this, its confusing (given all the new types).

cryptography/src/bls12381/primitives/ops/aggregate.rs
patrick-ogrady: Added to #2676

PR #2633, opened 2025-12-29
feat(storage,consensus): add index_for_key to Archive trait

doocho:
Add index_for_key method to Archive trait and get_height to Blocks trait for retrieving block height from commitment without deserializing the full block.
**Note** : Optimization applies to prunable archive only. Immutable archive falls back to full block fetch.


clabby:

PR #2423, opened 2025-12-08
[cryptography] never emit secrets when formatting

SuperFluffy:
Replaces all secrets (that I know of) by `"XXXXXXX"` in their `std::fmt::Display` and `std::fmt::Debug` impls.

It's too easy to accidentally leak these in logs otherwise.

If somebody really wants to leak the secrets, they can still manually encode the raw on-disk representation. But at least then this becomes explicit.

PR #2640, opened 2025-12-29
[cryptography] Add Secret<T> wrapper for protected key material

andresilva:
This PR introduces a `Secret<T>` wrapper type that provides secure handling of sensitive cryptographic material.

The wrapper prevents accidental leakage of secrets, `Debug` and `Display` implementations always print [REDACTED] rather than the actual value. When the secret is dropped, the memory is explicitly zeroized to prevent data from lingering in freed memory.

All comparisons use constant-time algorithms from the `subtle` crate to prevent timing side-channel attacks, examining every byte rather than short-circuiting on the first difference.

To read a secret, callers must use the `expose()` method with a closure. Currently the wrapper will just call the closure with the inner secret value and not do any special handling, but this API allows easily auditing all places where secrets are used (i.e. just grep for `expose`), and in the future this API can be augmented with OS-specific memory hardening.

All private key types in the cryptography crate now store their sensitive material in `Secret<T>` wrappers.

Replaces #2423.

### DIFF
--- a/cryptography/src/bls12381/primitives/group.rs
+++ b/cryptography/src/bls12381/primitives/group.rs
@@ -216,7 +216,6 @@ pub type Private = Scalar;
 pub const PRIVATE_KEY_LENGTH: usize = SCALAR_LENGTH;
 
 impl Scalar {
-
     /// Maps arbitrary bytes to a scalar using RFC9380 hash-to-field.
     pub fn map(dst: DST, msg: &[u8]) -> Self {
         // The BLS12-381 scalar field has a modulus of approximately 255 bits.


cronokirby:
This is a really good idea!

I think we should lean on some of the existing ecosystem crates like "subtle", because the compiler is really good at detecting your constant time comparison and messing it up, so it's better to use a nice dependency which can use the most current tricks to fool it.

cryptography/src/secret.rs
```diff
@@ -0,0 +1,528 @@
...
+            let ptr = ptr as *mut T;
+
+            // SAFETY: ptr is valid and properly aligned (mmap returns page-aligned memory)
+            unsafe { core::ptr::write(ptr, value) };
+
+            // SAFETY: ptr points to valid mmap'd memory of size `size`
+            // In soft-mlock mode (tests/benchmarks), we continue even if mlock fails.
+            // The memory will still be protected via mprotect, just not pinned in RAM.
+            if unsafe { libc::mlock(ptr as *const libc::c_void, size) } != 0 {
+                #[cfg(not(any(test, feature = "soft-mlock")))]
```
andresilva: Each `Secret<T>` allocates a full page (4KB minimum) via `mmap` and calls `mlock` to pin it in RAM. Systems have a per-process limit on locked memory (typically 8MB, check with `ulimit -l`). In tests and benchmarks with many participants, we quickly exceed this limit. For example, with 50 DKG contributors:
- Each dealer sends a `DealerPrivMsg` to every player: 50 * 50 = 2500 `Secret<Scalar>` instances
- Plus shares: 50 `Secret<Private>` instances
- Total: ~2,550 secrets * 4KB = ~10MB > 8MB limit

The `soft-mlock` feature allows `mlock` to fail gracefully, memory is still protected via `mprotect`, just not pinned in RAM. This is acceptable for tests/benchmarks.

cryptography/src/secret.rs
```diff
@@ -0,0 +1,576 @@
...
+        }
+
+        /// Exposes the secret value for read-only access within a closure.
+        ///
+        /// Memory is re-protected immediately after the closure returns, even if
+        /// the closure panics.
+        #[inline]
+        pub fn expose<R>(&self, f: impl FnOnce(&T) -> R) -> R {
+            // Scope guard that re-protects memory on drop (including panic unwinding)
+            struct ReprotectGuard {
```
patrick-ogrady: Seems we could define outside the function (just for consistency) 🤷

cryptography/Cargo.toml
```diff
@@ -41,6 +42,9 @@ zeroize = { workspace = true, features = ["zeroize_derive"] }
 version = "0.2.15"
 features = ["js"]
 
+[target.'cfg(unix)'.dependencies]
+libc = "0.2"
```
patrick-ogrady: can we do a full version specification?

andresilva: I guess this should also go at the root `Cargo.toml`.

andresilva:
bugbot run

andresilva:
bugbot run

cryptography/src/secret.rs
patrick-ogrady: Could we avoid this just by more aggressively dropping keys? Or is there a test case that reaches 1k in memory (maybe there is)?

patrick-ogrady: I suppose the hard fail here is preferred over a WARN (as it may be "too late" then?)

cryptography/src/secret.rs
```diff
@@ -0,0 +1,721 @@
...
+        let s1 = Secret::new([1u8, 2, 3, 4]);
+        let s2 = Secret::new([1u8, 2, 3, 4]);
+        let s3 = Secret::new([5u8, 6, 7, 8]);
+        assert_eq!(s1, s2);
+        assert_ne!(s1, s3);
+    }
+
+    #[test]
+    fn test_ordering() {
+        use core::cmp::Ordering;
```
patrick-ogrady: import at top of mod?

cryptography/src/secret.rs
```diff
@@ -0,0 +1,721 @@
...
+        assert!(s1 >= s3);
+
+        assert_eq!(s1.partial_cmp(&s2), Some(core::cmp::Ordering::Less));
+        assert_eq!(s2.partial_cmp(&s1), Some(core::cmp::Ordering::Greater));
+        assert_eq!(s1.partial_cmp(&s3), Some(core::cmp::Ordering::Equal));
+    }
+
+    #[test]
+    fn test_scalar_equality() {
+        use crate::bls12381::primitives::group::Scalar;
```
patrick-ogrady: import at top of mod?

cryptography/src/secret.rs
```diff
@@ -0,0 +1,721 @@
...
+
+        // Same scalar should be equal
+        assert_eq!(s1, s2);
+        // Different scalars should (very likely) be different
+        assert_ne!(s1, s3);
+    }
+
+    #[test]
+    fn test_scalar_ordering() {
+        use crate::bls12381::primitives::group::Scalar;
```
patrick-ogrady: top of mod?

cryptography/src/secret.rs
andresilva: Here's the list of failing tests:

```
Summary [ 475.665s] 84 tests run: 70 passed (29 slow), 14 failed, 2637 skipped
        FAIL [   0.495s] commonware-reshare::bin/commonware-reshare validator::test::dkg_deterministic_slow_
        FAIL [   0.337s] commonware-reshare::bin/commonware-reshare validator::test::dkg_four_epochs_lossy_slow_
        FAIL [   0.511s] commonware-reshare::bin/commonware-reshare validator::test::dkg_four_epochs_slow_
        FAIL [   0.636s] commonware-reshare::bin/commonware-reshare validator::test::dkg_four_epochs_with_crashes_slow_
        FAIL [   0.499s] commonware-reshare::bin/commonware-reshare validator::test::dkg_single_epoch_slow_
        FAIL [   0.498s] commonware-reshare::bin/commonware-reshare validator::test::dkg_with_delay_slow_
        FAIL [   0.532s] commonware-reshare::bin/commonware-reshare validator::test::dkg_with_forced_failure_slow_
        FAIL [   0.523s] commonware-reshare::bin/commonware-reshare validator::test::dkg_with_many_forced_failures_slow_
        FAIL [   4.593s] commonware-reshare::bin/commonware-reshare validator::test::reshare_four_epochs_lossy_slow_
        FAIL [   7.265s] commonware-reshare::bin/commonware-reshare validator::test::reshare_with_delay_slow_
        FAIL [ 104.046s] commonware-consensus simplex::tests::test_determinism_slow_
        FAIL [  14.857s] commonware-reshare::bin/commonware-reshare validator::test::reshare_with_delay_subset_slow_
        FAIL [  41.321s] commonware-reshare::bin/commonware-reshare validator::test::reshare_four_epochs_changing_size_slow_
        FAIL [  32.000s] commonware-reshare::bin/commonware-reshare validator::test::reshare_with_many_forced_failures_slow_
```

I didn't look at why `simplex::tests::test_determinism_slow_` fails (can probably be fixed), but everything else is related to DKG where a lot of `DealerPrivMsg`s are created, each involving a `Secret<Scalar>`. On top of this all the cryptography benchmarks fail, e.g. [here](https://github.com/commonwarexyz/monorepo/blob/main/cryptography/src/ed25519/benches/signature_generation.rs#L19-L24) criterion will start by creating ~2000 keys.

andresilva: We could warn instead of hard failing, but for production environments that seems a bit counter-productive, it's very easy for the WARN to just be ignored, and it would defeat the purpose of this feature. If we're going to keep the feature flag it should probably be renamed to `unsafe-mlock` or something like that.

patrick-ogrady (commented):
It feels odd to me to not wrap the stream secret in this (given we may send shares over it) but suspect we'd rapidly exceed the allocation limit?

I suppose that quickly turns into a rabbit hole where we need to think about all encoded bytes on the way into/out of network?

patrick-ogrady:
Would it be possible for someone to spam you with shares in a DKG to cause exhaustion of the mlock mechanism?

andresilva:
@patrick-ogrady ~~Good point, and yes currently that would be possible!~~ We wrap the share in `Secret` at deserialization time before any validation. I suppose we could wrap the share in `Secret` only after validation has been done on it? Another idea that I thought about but didn't pursue was storing multiple secrets per page. This reduces the security level since you need to expose all secrets from that page whenever you manipulate one secret, but maybe could make sense for secrets in the same context, e.g. all DKG shares. This would complicate the API considerably though and not make it as transparent as it is now (i.e. you just do `Secret::new(my_secret_thing)`.

Edit: looking at reshare more closely this isn't as bad as I thought: you deserialize the message, immediately call `Player::dealer_message` which fails, and the message gets dropped (freeing the mlock slot). So I don't think an attacker can cause unbound exhaustion.

patrick-ogrady:
> Edit: looking at reshare more closely this isn't as bad as I thought: you deserialize the message, immediately call Player::dealer_message which fails, and the message gets dropped (freeing the mlock slot). So I don't think an attacker can cause unbound exhaustion.

IIRC we allow the peer to send many messages.

cryptography/src/secret.rs
```diff
@@ -0,0 +1,320 @@
...
+/// Zeroize memory at the given pointer using volatile writes.
+///
+/// # Safety
+///
+/// `ptr` must point to allocated, writable memory of at least `size_of::<T>()` bytes.
+#[inline]
+unsafe fn zeroize_ptr<T>(ptr: *mut T) {
+    let slice = core::slice::from_raw_parts_mut(ptr as *mut u8, core::mem::size_of::<T>());
+    slice.zeroize();
+}
```
cronokirby, line 33: WIth the concrete types this gets used with now, this is fine, but in general it might be better to require things inside of `Secret` to implement `Zeroize`. If we do that, we avoid the need to maintain some of this tricky code by deferring to https://docs.rs/zeroize/latest/zeroize/struct.Zeroizing.html as a wrapper inside of Secret.

cronokirby (changes requested):
- I think we should just not implement `Ord` or `PartialOrd` for now. I don't think we need this, and it seems difficult to use the resulting result correctly wrt constant-time correctness. I think expose provides a sufficient amount of leway for someone who really knows what they're doing to get around this limitation, while also notifying them enough that dragons are located in proximity.
- I think we can probably do away with all of the unsafe code in this module by leaning on the "zeroize" crate more. I think it's fine to require a bound of Zeroize for `Secret`, which will also be correct in case you end up with more complex types. One example which would already break the assumption here, is if you want to have a `Poly<Scalar>` inside of `Secret`. Internally, this struct contains a Vec, and so zeroizing the bytes of the struct is not sufficient, you also need to chase a pointer and zeroize that. So, requiring a `Zeroize` bound is both more correct, and lets us avoid having more unsafe code to worry about.

cryptography/src/secret.rs
andresilva: `p256::ecdsa::SigningKey` doesn't implement `Zeroize` unfortunately. I guess we can just require `ZeroizeOnDrop` which it does implement. This didn't work before with the linux-specific implementation where we just stored a pointer, but should be fine now I think. Need to test it.

andresilva: - ed25519_consensus::SigningKey
- x25519_dalek::SharedSecret
- x25519_dalek::EphemeralSecret

Don't implement `ZeroizeOnDrop` :disappointed:. I'm a bit torn on this, because the scope of this PR was significantly reduced, and while we could in theory support types that use the heap with what we have here, I'm not sure this is something that we want to enshrine. The current code assumes that the type being wrapped is self-contained, and as such the way we zeroize the memory works. Even if we had `ZeroizeOnDrop` for the wrapped type, we could still end up leaking unzeroed memory if that type implements `Copy`, whereas the current code is careful to avoid that.

andresilva:
I removed `Ord` implementation for `Secret`. I also updated the docs to make it a bit more explicit that this is only meant to be used for flat data types. I guess an alternative would be to have a marker trait `unsafe trait SecretSafe {}` and `Secret<T: SecretSafe> {}`, this would at least force users to implement an unsafe trait for their own types and hopefully read the docs? Although it also means that they might need to create wrapper types if they're trying to use `Secret` with a type they don't own.

cryptography/src/secret.rs
```diff
@@ -0,0 +1,204 @@
...
+
+impl<T> ZeroizeOnDrop for Secret<T> {}
+
+impl<T: Clone> Clone for Secret<T> {
+    fn clone(&self) -> Self {
+        self.expose(|v| Self::new(v.clone()))
+    }
+}
+
+impl<const N: usize> PartialEq for Secret<[u8; N]> {
```
andresilva: Ideally we'd have:

`impl<T: ConstantTimeEq> PartialEq for Secret<T> { ... }`

But subtle doesn't implement `ConstantTimeEq` for arrays (https://github.com/dalek-cryptography/subtle/pull/114), so this is the least worse option for our codebase. An alternative would be to use https://crates.io/crates/ctutils instead of subtle (which seems unmaintained at this point), it seems to implement `ConstantTimeEq` for arrays and slices.

cryptography/src/secp256r1/common.rs
```diff
@@ -64,70 +62,29 @@ impl Read for PrivateKeyInner {
...
     fn read_cfg(buf: &mut impl Buf, _: &()) -> Result<Self, CodecError> {
         let raw = <[u8; PRIVATE_KEY_LENGTH]>::read(buf)?;
-        let result = SigningKey::from_slice(&raw);
+        let key = SigningKey::from_slice(&raw);
         #[cfg(feature = "std")]
-        let key = result.map_err(|e| CodecError::Wrapped(CURVE_NAME, e.into()))?;
+        let key = key.map_err(|e| CodecError::Wrapped(CURVE_NAME, e.into()))?;
         #[cfg(not(feature = "std"))]
-        let key = result
-            .map_err(|e| CodecError::Wrapped(CURVE_NAME, alloc::format!("{:?}", e).into()))?;
```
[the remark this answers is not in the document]
andresilva: I intend to do a more in-depth audit on such cases in a follow-up PR (there are other cases that I already identified).

cryptography/src/secret.rs
andresilva: I pushed a commit to replace subtle with ctutils. Can revert if you prefer subtle.

patrick-ogrady: good with either

cryptography/src/secret.rs
```diff
@@ -0,0 +1,204 @@
...
+//! - Access to the inner value requires an explicit `expose()` call
+//! - Comparisons use constant-time operations to prevent timing attacks
+//!
+//! # Type Constraints
+//!
+//! **Important**: `Secret<T>` is designed for flat data types without pointers
+//! (e.g. `[u8; N]`). It does NOT provide full protection for types with
+//! indirection. Types like `Vec<T>`, `String`, or `Box<T>` will only have their
+//! metadata (pointer, length, capacity) zeroized, the referenced data remains
+//! intact. Do not use `Secret` with types that contain pointers.
```
patrick-ogrady, line 15: Is this `do not` because it isn't effective or because it'll cause some stability issue?

andresilva: The zeroization isn't effective, the heap data won't be zeroized because we don't follow any pointers, but it works and the memory will be freed (since we call drop).

cryptography/src/handshake/cipher.rs
```diff
@@ -77,17 +73,15 @@ impl RecvCipher {
         rng.fill_bytes(&mut key[..]);
```
patrick-ogrady: Should we do?

```
  ⎿  Added 2 lines, removed 1 line
      72      pub fn new(mut rng: impl CryptoRngCore) -> Self {
      73          let mut key = [0u8; 32];
      74          rng.fill_bytes(&mut key[..]);
      75 +        let key = Secret::new(key);
      76          Self {
      77              nonce: CounterNonce::new(),
      77 -            inner: Secret::new(ChaCha20Poly1305::new(&key.into())),
      78 +            inner: key.expose(|k| Secret::new(ChaCha20Poly1305::new(k.into()))),
      79          }
      80      }
      81

```

patrick-ogrady: Even better:

```
--- a/cryptography/src/handshake/cipher.rs
+++ b/cryptography/src/handshake/cipher.rs
@@ -44,11 +44,11 @@ pub struct SendCipher {
 impl SendCipher {
     /// Creates a new sending cipher with a random key.
     pub fn new(mut rng: impl CryptoRngCore) -> Self {
-        let mut key = [0u8; 32];
-        rng.fill_bytes(&mut key[..]);
         Self {
             nonce: CounterNonce::new(),
-            inner: Secret::new(ChaCha20Poly1305::new(&key.into())),
+            inner: Secret::new(ChaCha20Poly1305::new(
+                &ChaCha20Poly1305::generate_key(&mut rng),
+            )),
         }
     }

@@ -69,11 +69,11 @@ pub struct RecvCipher {
 impl RecvCipher {
     /// Creates a new receiving cipher with a random key.
     pub fn new(mut rng: impl CryptoRngCore) -> Self {
-        let mut key = [0u8; 32];
-        rng.fill_bytes(&mut key[..]);
         Self {
             nonce: CounterNonce::new(),
-            inner: Secret::new(ChaCha20Poly1305::new(&key.into())),
+            inner: Secret::new(ChaCha20Poly1305::new(
+                &ChaCha20Poly1305::generate_key(&mut rng),
+            )),
         }
     }
     ```
     ?

patrick-ogrady (approved):

PR #2653, opened 2025-12-30
[p2p/lookup] Fix `allow_unregistered_handshakes` (now `bypass_ip_check`)

patrick-ogrady:
In a previous PR (#2534), we broke `allow_unregistered_handshakes` (when ensuring a peer is dialing from their IP rather than any allowed IP).

This PR fixes that regression and adds regression tests.

### DIFF
--- a/p2p/src/authenticated/lookup/actors/listener.rs
+++ b/p2p/src/authenticated/lookup/actors/listener.rs
@@ -34,7 +34,7 @@ pub struct Config<C: Signer> {
     pub address: SocketAddr,
     pub stream_cfg: StreamConfig<C>,
     pub allow_private_ips: bool,
-    pub allow_unknown_ips: bool,
+    pub bypass_ip_check: bool,
     pub max_concurrent_handshakes: NonZeroU32,
     pub allowed_handshake_rate_per_ip: Quota,
     pub allowed_handshake_rate_per_subnet: Quota,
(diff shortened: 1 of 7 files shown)


p2p/src/authenticated/lookup/actors/tracker/record.rs
```diff
@@ -480,9 +483,68 @@ mod tests {
...
         assert!(
-            !record_blocked.acceptable(egress_ip),
+            !record_blocked.acceptable(egress_ip, false),
+            "Not acceptable when blocked"
+        );
+    }
+
+    #[test]
+    fn test_acceptable_allow_unregistered() {
+        use std::net::IpAddr;
```
patrick-ogrady: nit: remove from inline

p2p/src/authenticated/lookup/config.rs
```diff
@@ -30,9 +30,9 @@ pub struct Config<C: Signer> {
     /// is selected from the results (shuffled for each dial attempt).
     pub allow_dns: bool,
 
-    /// Whether or not to attempt handshakes on incoming connections
-    /// from unregistered IP addresses.
-    pub attempt_unregistered_handshakes: bool,
+    /// Whether or not to skip IP verification for incoming connections
+    /// (allows known peers to connect from unknown IPs).
+    pub allow_unknown_ips: bool,
```
patrick-ogrady: We could offer the option for peers to register with a `None` egress but that seems like it would encourage the wrong behavior (we should attempt to use a known egress to avoid handshakes with random peers).

p2p/src/authenticated/lookup/actors/listener.rs
```diff
@@ -262,7 +262,8 @@ impl<E: Spawner + Clock + Network + Rng + CryptoRng + Metrics, C: Signer> Actor<
                     let supervisor = supervisor.clone();
                     move |context| async move {
                         Self::handshake(
-                            context.into_present(), address, stream_cfg, sink, stream, tracker, supervisor,
+                            context.into_present(), address, stream_cfg, sink, stream,
+                            tracker, supervisor,
```
andresilva, line 266: The joy of rustfmt not working inside macros 🫠

andresilva (approved):

BrendanChou (approved):

PR #2656, opened 2025-12-31
[cryptography] [storage] Replace `Hasher::EMPTY` with `Digest::EMPTY`

### LINKED ISSUE #2318: [Storage] Remove use of dummy hashes in MMR
As of #2295 we use the `Hasher` in `DirtyMmr::add_leaf_digest` only to create an `empty()` hash. We should remove this and implement `Default` for `Digest` or something.

```rs
    pub(super) fn add_leaf_digest<H: Hasher<D>>(&mut self, _hasher: &mut H, digest: D) -> Position {
        // Compute the new parent nodes, if any.
        let nodes_needing_parents = nodes_needing_parents(self.peak_iterator())
            .into_iter()
            .rev();
        let leaf_pos = self.size();
        self.nodes.push_back(digest);

        let mut height = 1;
        for _ in nodes_needing_parents {
            let new_node_pos = self.size();
            self.nodes
                .push_back(<H::Inner as commonware_cryptography::Hasher>::empty());
            self.state.dirty_nodes.insert((new_node_pos, height));
            height += 1;
        }

        leaf_pos
    }
```

danlaine:
Resolves #2318.

Removes `Hasher::EMPTY` which is the digest of an empty input. Adds `Digest::EMPTY` which is an all-zero digest.

`Hasher::EMPTY`'s only usage was in `storage/src/mmr/mem.rs` where it was used as a dummy value.

This change allows us to remove a `_hasher` parameter which was only used to create (`EMPTY`) dummy values.


cryptography/fuzz/fuzz_targets/blake3_hasher.rs
```diff
@@ -88,9 +88,6 @@ fn fuzz_diff_hash(data: &[u8]) {
 }
 
 fn fuzz_digest_operations(data: &[u8]) {
-    let empty_digest = OurBlake3::EMPTY;
```
danlaine, line 91: I opted to remove this rather than replace it because I'm not sure that asserting the length of a constant buys us anything

examples/reshare/src/application/types.rs
```diff
@@ -132,5 +132,5 @@ where
     C: Signer,
     V: Variant,
 {
-    Block::new(H::EMPTY, 0, None)
+    Block::new(<<H as Hasher>::Digest as Digest>::EMPTY, 0, None)
```
danlaine, line 135: `Block` could be re-written to be parameterized on `D: Digest` instead of `H: Hasher` but I opted to make the minimal diff change here

PR #2660, opened 2026-01-01
[cryptography/certificate] declare if a scheme benefits from batch verification

andresilva:


andresilva:
Code coverage will increase in a follow-up PR 😬.

consensus/src/simplex/actors/batcher/verifier.rs
```diff
@@ -1078,15 +1112,23 @@ mod tests {
...
+            let is_last = i == pending_schemes.len() - 1;
             verifier.add(
                 Vote::Finalize(create_finalize(scheme, round, View::new(0), 1)),
                 false,
             );
+            if is_last {
+                assert!(verifier.ready_finalizes());
+            } else if S::is_batchable() {
+                // Batchable schemes wait for quorum
+                assert!(!verifier.ready_finalizes());
```
patrick-ogrady, line 1130: Can we add a test that non-batchable doesn't trigger verification when we have > quorum (but already have a certificate)?

AFAICT the code is correct and doesn't do that but seems like an easy regression we may introduce by accident.

andresilva: We already have tests for this `ready_*_quorum_already_met_by_verified`, they are generic over scheme, so once I add a non-batchable scheme (in a follow-up) those tests should assert this property holds.

consensus/src/simplex/actors/batcher/verifier.rs
```diff
@@ -5,12 +5,13 @@ use crate::simplex::{
...
 
-/// `Verifier` is a utility for tracking and batch verifying consensus messages.
+/// `Verifier` is a utility for tracking and verifying consensus messages.
 ///
-/// In consensus, verifying multiple signatures at the same time can be much more efficient
-/// than verifying them one by one. This struct collects messages from participants in consensus
-/// and signals they are ready to be verified (when there exist enough messages to potentially reach
-/// a quorum).
+/// For schemes where [`Scheme::is_batchable()`](commonware_cryptography::certificate::Scheme::is_batchable)
+/// returns `true` (such as ed25519 and BLS12-381), this struct collects messages and defers verification
```
patrick-ogrady: nit: can we link to these schemes?

consensus/src/simplex/actors/batcher/verifier.rs
```diff
@@ -5,12 +5,13 @@ use crate::simplex::{
...
+/// until enough messages exist to potentially reach a quorum, enabling efficient batch verification. For
+/// schemes where `is_batchable()` returns `false` (such as SECP256R1), signatures are verified eagerly as
```
patrick-ogrady: nit: can we link to this scheme?

cryptography/src/ed25519/certificate/mod.rs
```diff
@@ -379,6 +379,12 @@ impl Generic {
         true
     }
 
+    pub const fn is_batchable() -> bool {
+        // Batch verification requires the `std` feature because it depends on
+        // `ed25519_consensus::batch::Verifier`.
+        cfg!(feature = "std")
```
patrick-ogrady, line 385: :fancy:

cryptography/src/certificate.rs
```diff
@@ -263,6 +263,16 @@ pub trait Scheme: Clone + Debug + Send + Sync + 'static {
     /// return `true`.
     fn is_attributable() -> bool;
 
+    /// Returns whether this scheme benefits from batch verification.
+    ///
+    /// Schemes that benefit from batch verification (like [`ed25519`], [`bls12381_multisig`]
+    /// and [`bls12381_threshold`]) should return `true`, allowing callers to optimize by
+    /// deferring verification until multiple signatures are available.
+    ///
+    /// Schemes that don't benefit from batch verification (like `secp256r1`) should
```
patrick-ogrady, line 272: nit: link to this

consensus/src/simplex/actors/batcher/verifier.rs
patrick-ogrady: I think we can link to the raw mod, not the cert (but if you want to do that I can see why you're waiting)

andresilva: Yeah I wanted to link to the certificate mods, I can update these, and will link to secp256r1 in the follow-up.

cryptography/src/certificate.rs
andresilva: Will update in the follow-up to link to the certificate mod.

consensus/src/simplex/actors/batcher/verifier.rs
andresilva: Updated.

PR #2661, opened 2026-01-01
[cryptography/secp256r1] Add batch verification

commonware-llm:
## Summary

Add batch verification for secp256r1 recoverable signatures using algebraic batching with random coefficients.

### Features
- Add `Batch` struct implementing `BatchVerifier` trait for recoverable signatures
- Use algebraic batch verification: `sum(z_i * u1_i) * G + sum(z_i * u2_i * P_i) - sum(z_i * R_i) = O`
- Random coefficients (z_i) prevent cancelling forgery attacks
- Explicit rejection of malleable (high-s) signatures
- Montgomery batch inversion optimization (1 inversion + 3(n-1) multiplications instead of n inversions)

### Why Only Recoverable Signatures?

Standard ECDSA cannot benefit from algebraic batch verification because determining R requires computing `u1*G + u2*P` per signature - which IS the main verification cost. Recoverable signatures include the recovery ID needed to reconstruct R directly via point decompression.

## Performance

Benchmark comparing batch verification vs individual verification (Apple Silicon):

| Batch Size | Individual | Batch | Speedup | Per-sig (Individual) | Per-sig (Batch) |
|------------|------------|-------|---------|----------------------|-----------------|
| 1 | 406 µs | 306 µs | **1.33x** | 406 µs | 306 µs |
| 2 | 808 µs | 502 µs | **1.61x** | 404 µs | 251 µs |
| 4 | 1,649 µs | 926 µs | **1.78x** | 412 µs | 232 µs |
| 8 | 3,219 µs | 1,704 µs | **1.89x** | 402 µs | 213 µs |
| 16 | 6,432 µs | 3,073 µs | **2.09x** | 402 µs | 192 µs |
| 32 | 12,928 µs | 6,063 µs | **2.13x** | 404 µs | 190 µs |
| 64 | 26,178 µs | 12,065 µs | **2.17x** | 409 µs | 189 µs |
| 128 | 51,977 µs | 23,930 µs | **2.17x** | 406 µs | 187 µs |

### Key Results
- **Asymptotic speedup: ~2.17x** for large batches
- **Per-signature cost drops from ~406 µs to ~187 µs** (54% reduction)
- Even single signatures benefit (1.33x) because batch verification uses algebraic verification instead of public key recovery


patrick-ogrady:
This isn't really worth the squeeze/complexity (the only place where batch verification makes sense is recoverable signatures where we already have the associated public keys).

patrick-ogrady:
> This isn't really worth the squeeze/complexity (the only place where batch verification makes sense is recoverable signatures where we already have the associated public keys).

It may be worth it if we want to just use recoverable signatures always (but seems to be slower than #2684)

clabby:

PR #2663, opened 2026-01-01
[consensus/marshal] Improve `set_floor` documentation

clabby:
## Overview

Improves the documentation for the `set_floor` function. The old documentation was slightly incorrect, since marshal starts syncing + delivering finalized blocks starting at `floor + 1`:

> The sync floor is the latest block that the application has processed. Marshal will not attempt to sync blocks below this height nor deliver blocks below this height to the application.


patrick-ogrady (changes requested):
Proposed some nits in #2664 (to better clarify this isn't to be used for pruning)

Issue #2607, opened 2025-12-23
[storage] Consider page-level CRCs instead of record level


roberto-bayardo:
Page level CRCs would reduce space / CPU overhead in our storage systems, but would introduce some additional complexity. It's worth exploring whether it's worth moving towards.

The main benefit of page-level CRCs over record level is it's trivial to discard them after verification after loading a page into the buffer pool.  This would lead to both increased cache utilization since we don't store them in RAM, and better CPU utilization since we verify only on page-read.

Issue #1219, opened 2025-07-07
[storage/journal/fixed.rs] Immutable blob wrapper


roberto-bayardo:
(Re)consider having an Immutable blob wrapper for historical blobs so we can avoid the overhead of the Write buffer for each, even though they are never written to.

PR #2667, opened 2026-01-02
[storage & runtime] Switch to page-level checksums instead of record-level checksums

roberto-bayardo:
This PR extends the buffer pool and its associated blob wrappers to perform integrity verification at a page level, allowing removal of the "application-level" integrity checking from storage structures such as journal which were explicitly CRCing every item.

Note that a good deal of complexity (e.g. dual CRCs per page) pertains to being able to rewrite the final (partial) page without affecting durability of any previously commited version.

Resolves: https://github.com/commonwarexyz/monorepo/issues/2607
Resolves: https://github.com/commonwarexyz/monorepo/issues/1219
Resolves: https://github.com/commonwarexyz/monorepo/issues/2728

### DIFF
--- a/storage/conformance.toml
+++ b/storage/conformance.toml
@@ -1,10 +1,10 @@
 ["commonware_storage::archive::conformance::ArchiveImmutable"]
 n_cases = 128
-hash = "eb7860fd0cd068f243570ed34a8f8a7125deadc4600579237c4fe3e7fa73fa37"
+hash = "6acfa1bc0c17920b5c0e0437af106e09ee57dcd37459091402192f2c146afdb5"
 
 ["commonware_storage::archive::conformance::ArchivePrunable"]
 n_cases = 128
-hash = "76a9d4914d02db2eb777d0ba12f2532379dc48687aa34f5c0d2146b0f05b7f6a"
+hash = "cb063a05c6a75902893f790e9802b4906be506f2f7e5d10b46dff90a92e40819"
 
 ["commonware_storage::archive::immutable::storage::conformance::CodecConformance<Record>"]
 n_cases = 65536
@@ -60,11 +60,11 @@ hash = "e077ce8c6d9a79c87cf9b48866c65f387ffbec9d8e8c65dd40c46b4296cfc050"
 
 ["commonware_storage::journal::conformance::SegmentedGlob"]
 n_cases = 512
-hash = "c35d6e8bfab7f0a31ce305ca79022eec37d59b2f2beedecd68afaa499d9c4e68"
+hash = "adb1efeef12c203c05879ce4d1d03ef443c767737a6c6b57433189100eec9197"
 
 ["commonware_storage::journal::conformance::SegmentedOversized"]
 n_cases = 512
-hash = "bc4d61617b625d38b0e194e4c456fbd287d609843afa14eceb8b723534cd16e1"
+hash = "b815138329a06cbe235cf547ed62774165bd2108e68c65bf15ae152bedf84b3a"
 
 ["commonware_storage::journal::conformance::SegmentedVariable"]
 n_cases = 512


runtime/src/utils/buffer/pool_crc/append.rs
```diff
@@ -0,0 +1,988 @@
...
+//! # Immutability
+//!
+//! The wrapper can be created in (or converted to) an immutable state, which will prevent any
+//! modifications while still supporting cached reads. This can be used to reduce its memory
+//! footprint and/or to prevent unintended modifications.
+//!
+//! # Page-level structure
+//!
+//! Data is appended to the underlying blob in _pages_, with the page size dictated by the managing
+//! buffer pool. A full _page_ consists of `page_size - 4` bytes of _logical_ data, followed by a
```
patrick-ogrady: I think we need to store `[(crc,len), (crc,len)]` at the end of the page for this to work?

If we don't, an unclean shutdown during write could lead us to unwind the entire blob (eventhough earlier offsets returned from `sync`?)

roberto-bayardo: Not sure I understand? Since CRCs appear at fixed positions (end of each page, except the very last page), if a previous CRC was written and survived a sync() it should be "permanent" (unless the user explicitly resizes the blob to an earlier point).

storage/src/qmdb/any/unordered/fixed/sync.rs
```diff
@@ -209,6 +209,7 @@ pub(crate) async fn init_journal_at_size<E: Storage + Metrics, A: CodecFixed<Cfg
     // Calculate the tail blob index and number of items in the tail
     let tail_index = size / cfg.items_per_blob;
     let tail_items = size % cfg.items_per_blob;
+    // TODO: this doesn't handle page-level CRCs
```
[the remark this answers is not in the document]
roberto-bayardo: yep, as the TODO indicates it's being worked on.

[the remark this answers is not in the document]
roberto-bayardo: Also, the code was fine.  in this new model even "dummy data" from resizing will pass checksum checks since resize is checksum-aware.  The test just needed tweaking.

storage/src/journal/segmented/variable.rs
```diff
@@ -388,11 +357,9 @@ impl<E: Storage + Metrics, V: Codec> Journal<E, V> {
         // Replay all blobs in order and stream items as they are read (to avoid occupying too much
         // memory with buffered data)
         Ok(stream::iter(blobs).flat_map(
-            move |(section, blob, max_offset, blob_size, codec_config, compressed)| {
-                // Created buffered reader
-                let mut reader = Read::new(blob, blob_size, buffer);
+            move |(section, mut blob, max_offset, codec_config, compressed)| {
                 if section == start_section && offset != 0 {
-                    if let Err(err) = reader.seek_to(offset as u64 * ITEM_ALIGNMENT) {
```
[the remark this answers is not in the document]
roberto-bayardo: users would generally abort on any error, but I fixed this anyway.

runtime/src/utils/buffer/pool/read.rs
```diff
@@ -0,0 +1,390 @@
...
+        if remaining > CRC_USIZE {
+            if !verify_crc(&self.buffer[offset..]) {
+                debug!(
+                    page = self.blob_page + (offset / self.page_size) as u64,
+                    "CRC mismatch"
+                );
+                return Err(Error::InvalidChecksum);
+            }
+            self.buffer.truncate(offset + remaining - CRC_USIZE);
+        }
```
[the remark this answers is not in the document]
roberto-bayardo: There is a floor on the capacity which is 2x the page size, so this should never happen.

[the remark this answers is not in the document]
roberto-bayardo: oops, you're right, this is in the Read class.  Fixed by forcing capacity to be a multiple of page size.

runtime/src/utils/buffer/pool/append.rs
```diff
@@ -0,0 +1,1004 @@
...
+
+        let (partial_page_bytes, new_blob_size) =
+            Self::align_blob(&blob, blob_size, page_size).await?;
+        if blob_size != new_blob_size {
+            warn!(
+                blob_size,
+                new_blob_size, "immutable blob has invalid trailing data"
+            );
+            return Err(Error::InvalidChecksum);
+        }
```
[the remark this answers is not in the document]
roberto-bayardo: nice catch, fixed.

runtime/src/utils/buffer/pool/read.rs
```diff
@@ -0,0 +1,395 @@
...
+            error!(
+                page = self.blob_page + (offset / self.page_size) as u64,
+                "CRC mismatch"
+            );
+            return Err(Error::InvalidChecksum);
+        }
+        self.buffer.truncate(offset + remaining - CRC_USIZE);
+
+        // We should have buffered at least one byte.
+        assert!(self.buffer_position < self.buffer.len());
```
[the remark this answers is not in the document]
roberto-bayardo: Thanks, I amended the check on L145 to return BlobInsufficientLength if a fill_buffer attempt is tried when all data has been read.

storage/src/journal/contiguous/fixed.rs
```diff
@@ -220,33 +221,24 @@ impl<E: Storage + Metrics, A: CodecFixed<Cfg = ()>> Journal<E, A> {
...
-        let blobs = try_join_all(blobs.into_iter().map(|(index, (blob, size))| {
-            let pool = cfg.buffer_pool.clone();
-            async move {
-                let blob = Append::new(blob, size, cfg.write_buffer, pool).await?;
-                Ok::<_, Error>((index, (blob, size)))
-            }
-        }))
-        .await?;
-        let tail = Append::new(tail, tail_size, cfg.write_buffer, cfg.buffer_pool.clone()).await?;
-        let size = tail_index * cfg.items_per_blob.get() + (tail_size / Self::CHUNK_SIZE_U64);
```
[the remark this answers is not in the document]
roberto-bayardo: This is fine, if a blob is truncated, that is a data corruption issue that will still be detected (by trying to read beyond the end of the blob).

runtime/src/utils/buffer/pool_crc/append.rs
roberto-bayardo: As discussed in meet, I see what you mean now.  We could fail while overwriting a partial page CRC, causing a rollback of a previously committed operation. Investigating fixes now.

roberto-bayardo:
Open question is do we also want journals that do manage their own CRCs (e.g. what if we're writing really large blocks to a journal. A page-level journal might be less efficient.)

roberto-bayardo:
Converting to draft while we work out the stable checksum solution.

runtime/src/utils/buffer/pool_crc/append.rs
roberto-bayardo: double-CRC fix is now implemented.

storage/src/journal/contiguous/fixed.rs
```diff
@@ -267,51 +270,6 @@ impl<E: Storage + Metrics, A: CodecFixed<Cfg = ()>> Journal<E, A> {
...
-        }
-
-        // Truncate any records with failing checksums. This can happen if the file system allocated
-        // extra space for a blob but there was a crash before any data was written to that space.
-        while tail_size > 0 {
-            let offset = tail_size - Self::CHUNK_SIZE_U64;
-            let read = tail.read_at(vec![0u8; Self::CHUNK_SIZE], offset).await?;
-            match Self::verify_integrity(read.as_ref()) {
-                Ok(_) => break, // Valid item found, we can stop truncating.
-                Err(Error::ChecksumMismatch(_, _)) => {
```
danlaine, line 297: `Error::ChecksumMismatch` is now unused

runtime/src/utils/buffer/pool/mod.rs
```diff
@@ -0,0 +1,328 @@
...
+//!
+//! Blob data is stored in _pages_, with the page size dictated by the managing buffer pool. A
+//! _page_ consists of up to `page_size-12` bytes of _logical_ data, and the last `page_size-12`
+//! bytes storing a _CRC record_ containing:
+//!
+//! ```text
+//! | len1 (2 bytes) |  crc1 (4 bytes) | len2 (2 bytes) | crc2 (4 bytes) |
+//! ```
+//!
+//! The checksum in the record preceeded by the larger `len` value is the one that should be treated
```
danlaine:
```suggestion
//! The checksum in the record preceded by the larger `len` value is the one that should be treated
```

runtime/src/utils/buffer/pool/mod.rs
```diff
@@ -0,0 +1,328 @@
...
+    page.truncate(len as usize);
+
+    Ok(page)
+}
+
+/// Describes a CRC record stored at the end of a page.
+///
+/// The CRC accompanied by the larger length is the one that should be treated as authoritative for
+/// the page. Two checksums are stored so that partial pages can be written without overwriting a
+/// valid checksum for a committed previously commited partial page.
```
danlaine:
```suggestion
/// valid checksum for a previously committed partial page.
```

runtime/src/utils/buffer/pool/mod.rs
```diff
@@ -0,0 +1,328 @@
...
+//! ```
+//!
+//! The checksum in the record preceeded by the larger `len` value is the one that should be treated
+//! as authoritative for the page. The two `len` values will never be equal (implying at least one
+//! is non-zero). Two checksums are stored so that partial pages can be written without overwriting
+//! a valid checksum for a previously commited partial page.
+//!
+//! A _full_ page is one whose crc stores a len equal to the logical page size. Otherwise the page
+//! is called _partial_. All pages in a blob are full except for the very last page, which can be
+//! full or partial. A checksum over a partial page is computed only over the first [0-len) bytes in
```
danlaine:
```suggestion
//! full or partial. A checksum over a partial page is computed only over the first [0, len) bytes in
```

runtime/src/utils/buffer/pool/mod.rs
```diff
@@ -0,0 +1,328 @@
...
+    const fn new(len: u16, crc: u32) -> Self {
+        Self {
+            len1: len,
+            crc1: crc,
+            len2: 0,
+            crc2: 0,
+        }
+    }
+
+    fn from_bytes(buf: [u8; 12]) -> Self {
```
danlaine:
```suggestion
    fn from_bytes(buf: [u8; CRC_RECORD_SIZE]) -> Self {
```

runtime/src/utils/buffer/pool/mod.rs
```diff
@@ -0,0 +1,328 @@
...
+            crc2,
+        }
+    }
+
+    /// Return the CRC record for the page if it is valid. The provided slice is assumed to be
+    /// exactly the size of a physical page.
+    fn validate_page(buf: &[u8]) -> Option<Self> {
+        let page_size = buf.len() as u64;
+        let crc_start_idx = (page_size - CRC_RECORD_SIZE) as usize;
+        let crc_bytes: [u8; 12] = buf[crc_start_idx..].try_into().unwrap();
```
danlaine:
```suggestion
        let crc_bytes: [u8; CRC_RECORD_SIZE] = buf[crc_start_idx..].try_into().unwrap();
```

runtime/src/utils/buffer/tip.rs
```diff
@@ -16,15 +14,19 @@ pub(super) struct Buffer {
 
     /// The maximum size of the buffer.
     pub(super) capacity: usize,
+
+    /// Whether this buffer should allow new data.
+    pub(super) immutable: bool,
```
danlaine, line 20: Do we want to use the type-state trick instead of having a boolean to differentiate mutable/immutable `Buffer`s?

runtime/src/utils/buffer/pool/page_cache.rs
```diff
@@ -83,63 +93,59 @@ pub struct PoolRef {
...
-    /// Returns a new [PoolRef] with the given `page_size` and `capacity`.
-    pub fn new(page_size: NonZeroUsize, capacity: NonZeroUsize) -> Self {
+    /// Returns a new [PoolRef] that will buffer up to `capacity` pages with the
+    /// given `page_size`.
+    pub fn new(page_size: u16, capacity: NonZeroUsize) -> Self {
+        let page_size = page_size as u64;
+
         Self {
-            page_size: page_size.get(),
+            page_size,
```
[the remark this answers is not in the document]
danlaine: For what it's worth, keeping the `NonZeroUSize` usage that was removed would make this diff a lot smaller, too.

runtime/src/utils/buffer/pool/mod.rs
```diff
@@ -0,0 +1,475 @@
...
+struct CrcRecord {
+    len1: u16,
+    crc1: u32,
+    len2: u16,
+    crc2: u32,
+}
+
+impl CrcRecord {
+    /// Create a new CRC record with the given length and CRC.
+    /// The new CRC is stored in the second slot (len2/crc2), with the first slot zeroed.
```
danlaine:
```suggestion
    /// The new CRC is stored in the first slot (len1/crc1), with the second slot zeroed.
```

runtime/src/utils/buffer/pool/append.rs
```diff
@@ -0,0 +1,1724 @@
...
+    pub async fn to_mutable(&self) {
+        let mut buffer = self.buffer.write().await;
+        if !buffer.immutable {
+            return;
+        }
+        buffer.immutable = false;
+
+        // Enlarge the write buffer to the capacity that was specified when the blob was created.
+        let capacity = buffer.capacity;
+        buffer.data.reserve(capacity);
```
andresilva: This will add capacity on top of whatever the current allocation is?

```rust
let capacity = buffer.capacity;
let current_len = buffer.data.len();
if buffer.data.capacity() < capacity {
    buffer.data.reserve(capacity - current_len);
}
```

runtime/fuzz/fuzz_targets/blob_integrity.rs
```diff
@@ -0,0 +1,222 @@
...
+            .expect("cannot append data");
+        append.sync().await.expect("cannot sync");
+        drop(append);
+
+        // Step 2: Corrupt a single bit in the blob.
+        // Calculate physical size: full pages + partial page (if any).
+        let full_pages = logical_size / page_size;
+        let partial_bytes = logical_size % page_size;
+        let physical_size = if partial_bytes > 0 {
+            full_pages * physical_page_size + partial_bytes + CRC_SIZE
```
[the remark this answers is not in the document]
roberto-bayardo: You are correct! You are smarter than Claude. Fixed.

runtime/src/utils/buffer/pool/read.rs
```diff
@@ -0,0 +1,417 @@
...
+                        page = self.blob_page + (read_offset / self.page_size) as u64,
+                        "CRC mismatch"
+                    );
+                    return Err(Error::InvalidChecksum);
+                };
+                // For non-last pages, the validated length must equal logical_page_size.
+                // A partial length here indicates corruption (e.g., fallback to old partial CRC).
+                let (len, _) = record.get_crc();
+                let len = len as usize;
+                let is_last_page = read_offset + self.page_size >= self.buffer.len();
```
andresilva: This treats the last page as last in buffer, not last in blob. I think a corrupted mid‑blob page can land at the end of a read chunk and be accepted as partial, truncating data silently.

```suggestion
                let is_last_page = start_offset + read_offset + self.page_size >= self.physical_blob_size;
```

runtime/src/utils/buffer/pool/append.rs
```diff
@@ -0,0 +1,1724 @@
...
+            .read(
+                &blob_guard.blob,
+                self.id,
+                &mut buf[cached..cached + uncached_len],
+                uncached_offset,
+            )
+            .await
+    }
+
+    /// Returns the protected region info for a partial page, if any.
```
danlaine, line 578: Can we elaborate on what a protected region is?

runtime/src/utils/buffer/pool/append.rs
```diff
@@ -0,0 +1,1724 @@
...
+    ///
+    /// # Returns
+    ///
+    /// `None` if there's no existing partial page.
+    ///
+    /// `Some((prefix_len, protected_slot))` where:
+    /// - `prefix_len`: bytes `[0..prefix_len]` were already written and can be skipped
+    /// - `protected_slot`: 0 or 1, indicating which CRC slot must not be overwritten
```
danlaine: Perhaps it's overkill but should we define an enum Slot that can take one of two values (Zero, One) and return that instead of u8? Or a boolean.

runtime/src/utils/buffer/pool/read.rs
andresilva:
```rust
  #[test_traced("DEBUG")]
  fn test_read_rejects_partial_crc_on_non_last_page() {
      executor.start(|context: deterministic::Context| async move {
          let (blob, blob_size) = context.open("test_partition", b"test_blob").await.unwrap();

          let pool_ref = PoolRef::new(PAGE_SIZE as u16, NZUsize!(BUFFER_SIZE));
          let append = Append::new(blob.clone(), blob_size, BUFFER_SIZE, pool_ref)
              .await
              .unwrap();

          // Two full pages.
          let data: Vec<u8> = (0u8..=255).cycle().take(PAGE_SIZE * 2).collect();
          append.append(&data).await.unwrap();
          append.sync().await.unwrap();

          // Corrupt page 0 to claim a shorter (partial) length with a valid CRC.
          let short_len = PAGE_SIZE / 2;
          let crc = crc32fast::hash(&data[..short_len]);
          let record = super::CrcRecord::new(short_len as u16, crc);
          let crc_offset = PAGE_SIZE as u64; // CRC record starts after logical page bytes
          blob.write_at(record.to_bytes().to_vec(), crc_offset).await.unwrap();
          blob.sync().await.unwrap();

          // Capacity of one page => bug reproduces if last-page check is buffer-based.
          let mut reader = append.as_blob_reader(NZUsize!(PAGE_SIZE)).await.unwrap();
          let mut buf = vec![0u8; PAGE_SIZE];
          let result = reader.read_exact(&mut buf, PAGE_SIZE).await;

          assert!(matches!(result, Err(Error::InvalidChecksum)));
      });
  }
```

runtime/src/utils/buffer/pool/page_cache.rs
```diff
@@ -231,10 +239,11 @@ impl PoolRef {
         let fetch_result = fetch_future.await;
         if !is_first_fetcher {
```
andresilva, line 253: In this case we don't do the size validation that the first fetcher does below.

runtime/src/utils/buffer/pool/read.rs
```diff
@@ -0,0 +1,417 @@
...
+    /// Returns the number of bytes actually read into the buffer, which will be [0, max_len).
+    pub async fn read_up_to(
+        &mut self,
+        buf: impl Into<StableBuf> + Send,
+    ) -> Result<(StableBuf, usize), Error> {
+        let mut buf = buf.into();
+        let current_pos = self.position();
+        let blob_size = self.blob_size();
+        let available = (blob_size.saturating_sub(current_pos) as usize).min(buf.len());
+        if available == 0 {
```
andresilva, line 98: On `Append` we return `Error::BlobInsufficientLength` in this case.

runtime/src/utils/buffer/pool/append.rs
```diff
@@ -0,0 +1,1724 @@
...
+        // not durably persisted.
+        {
+            let buf_guard = self.buffer.write().await;
+            if !buf_guard.immutable {
+                self.flush_internal(buf_guard, true).await?;
+            }
+        }
+
+        let physical_page_size = logical_page_size + CRC_RECORD_SIZE;
+        let blob_guard = self.blob_state.read().await;
```
andresilva, line 735: Could end up with stale logical/physical sizes if a concurrent writer appends data during the gap between the lock above being released and this one being acquired?

storage/src/journal/contiguous/fixed.rs
```diff
@@ -204,14 +197,27 @@ impl<E: Storage + Metrics, A: CodecFixed<Cfg = ()>> Journal<E, A> {
...
-        // Trim invalid items from the tail blob.
-        tail_size = Self::trim_tail(&tail, tail_size, tail_index).await?;
-        if tail_size > full_size {
-            return Err(Error::InvalidBlobSize(tail_index, tail_size));
-        }
+        // Non-tail blobs can be immutable.
+        let mut blobs = try_join_all(blobs.into_iter().map(|(index, (blob, size))| {
+            let pool = cfg.buffer_pool.clone();
+            async move {
+                let blob = Append::new_immutable(blob, size, cfg.write_buffer.get(), pool).await?;
```
andresilva, line 228: Maybe validate that these are full‑sized (`logical size == full_size`)? A historical blob truncated on a page boundary could pass CRC validation with fewer items than expected, causing position miscalculation for all subsequent reads.

storage/src/journal/contiguous/fixed.rs
```diff
@@ -220,33 +226,30 @@ impl<E: Storage + Metrics, A: CodecFixed<Cfg = ()>> Journal<E, A> {
...
-        }))
-        .await?;
-        let tail = Append::new(tail, tail_size, cfg.write_buffer, cfg.buffer_pool.clone()).await?;
-        let size = tail_index * cfg.items_per_blob.get() + (tail_size / Self::CHUNK_SIZE_U64);
         let pruning_boundary = if blobs.is_empty() {
             tail_index * cfg.items_per_blob.get()
         } else {
             blobs[0].0 * cfg.items_per_blob.get()
         };
+        let size = tail_index * cfg.items_per_blob.get() + (tail_size / Self::CHUNK_SIZE_U64);
```
andresilva, line 269: Init computes size from `tail_size / CHUNK_SIZE` but leaves any leftover bytes in the tail blob. If the tail blob is not item‑aligned (e.g. page‑level truncation), later replay will error and append will hit the alignment assert. Maybe truncate the tail blob to the last full item or fail in init when `tail_size % CHUNK_SIZE != 0`?

runtime/src/utils/buffer/tip.rs
roberto-bayardo: Yes, but given the timeline I though this would be more expedient. Added a TODO to change it though: https://github.com/commonwarexyz/monorepo/issues/2731

storage/src/journal/contiguous/fixed.rs
roberto-bayardo: removed it from mod.rs

runtime/src/utils/buffer/pool/append.rs
```diff
@@ -0,0 +1,1724 @@
...
+        // Enlarge the write buffer to the capacity that was specified when the blob was created.
+        let capacity = buffer.capacity;
+        buffer.data.reserve(capacity);
+    }
+
+    /// Read and return the last valid page number, along with its logical bytes for buffering if
+    /// the last page is a partial page. Returns 0 if there is no last valid page.
+    ///
+    /// The `logical_page_size` parameter is the size of the logical data portion of each page
+    /// (not including the CrcRecord).
```
danlaine: Nit: clarifies the return values
```suggestion
    /// Scan backwards from the end of the blob to find the last page with a valid CRC.
    ///
    /// The `logical_page_size` parameter is the size of the logical data portion of each page
    /// (not including the CrcRecord).
    ///
    /// # Returns
    ///
    /// A tuple of `(partial_page, page_count, invalid_data_found)`:
    ///
    /// - `partial_page`: If the last valid page contains fewer than `logical_page_size` bytes,
    ///   returns `Some((data, crc_record))` so the caller can buffer it for further appends.
    ///   Returns `None` if the last valid page is full or if no valid pages exist.
    ///
    /// - `page_count`: The number of physical page slots up to and including the last valid page.
    ///
    /// - `invalid_data_found`: `true` if any bytes beyond the last valid page were found,
    ///   indicating the blob should be truncated.
```

storage/src/journal/contiguous/fixed.rs
roberto-bayardo: Good point, done.

However I think this is a failure mode that we don't handle properly more generally.  e.g. what if this happens to the tail blob? We could silently lose committed data.  Something to think through a bit more.

runtime/src/utils/buffer/pool/append.rs
```diff
@@ -0,0 +1,1724 @@
...
+    async fn read_last_valid_page(
+        blob: &B,
+        blob_size: u64,
+        logical_page_size: u64,
+    ) -> Result<(Option<(Vec<u8>, CrcRecord)>, u64, bool), Error> {
+        let physical_page_size = logical_page_size + CRC_RECORD_SIZE;
+        let partial_bytes = blob_size % physical_page_size;
+        let mut last_page_end = blob_size - partial_bytes;
+
+        // If the last page is not full, it can't have a valid CRC record and must be invalid.
```
danlaine: Might be worth disambiguating that we mean _physical_ page here -- people might get confused and think that partial _logical_ pages are invalid

runtime/src/utils/buffer/pool/append.rs
```diff
@@ -0,0 +1,1724 @@
...
+
+        // Ensure buffer capacity is maintained.
+        let capacity = buf_guard.capacity;
+        let current_len = buf_guard.data.len();
+        if buf_guard.data.capacity() < capacity {
+            buf_guard.data.reserve(capacity - current_len);
+        }
+
+        Ok(())
+    }
```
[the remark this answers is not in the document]
roberto-bayardo: I think this is fine? I don't see where Blob suggests resizing is durable without a sync() call.

runtime/src/utils/buffer/pool/append.rs
roberto-bayardo: My thinking was if you have a concurrent writer, your lengths will be going stale anyway the moment the lock is released. So as long as you capture the sizes from while it was in a consistent state, this is fine.

runtime/src/utils/buffer/pool/page_cache.rs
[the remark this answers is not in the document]
roberto-bayardo: I wanted to emphasize the range had to be within a u16 though, otherwise there might be a high chance of user error. We could of course panic or error if the value is too large but seemed better to leverage the compiler?

runtime/src/utils/buffer/pool/mod.rs
roberto-bayardo: Looks like I concurrently resolved this when rewording the docs here.

runtime/src/utils/buffer/pool/mod.rs
roberto-bayardo: fixed

runtime/src/utils/buffer/pool/mod.rs
roberto-bayardo: fixed

runtime/src/utils/buffer/pool/mod.rs
roberto-bayardo: fixed

runtime/src/utils/buffer/pool/mod.rs
roberto-bayardo: fixed

runtime/src/utils/buffer/pool/page_cache.rs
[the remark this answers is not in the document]
danlaine: Could we use `NonZeroU16`?

runtime/src/utils/buffer/pool/append.rs
```diff
@@ -0,0 +1,1724 @@
...
+
+        let logical_page_size = self.pool_ref.page_size();
+        let physical_page_size = logical_page_size + CRC_RECORD_SIZE;
+
+        // Flush any buffered data first to ensure we have a consistent state on disk.
+        self.sync().await?;
+
+        // Acquire both locks to prevent concurrent operations.
+        let mut buf_guard = self.buffer.write().await;
+        let mut blob_guard = self.blob_state.write().await;
```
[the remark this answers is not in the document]
roberto-bayardo: I think if you're writing concurrently while resizing, all bets are off.  You might get an OK from the write and then the resize right after would "silently" undo it even if we were to hold the locks for the duration of the resize.

runtime/src/utils/buffer/pool/read.rs
```diff
@@ -0,0 +1,417 @@
...
+        self.logical_blob_size
+    }
+
+    /// Returns the current logical position in the blob.
+    pub const fn position(&self) -> u64 {
+        let logical_page_size = (self.page_size - CRC_RECORD_USIZE) as u64;
+        self.blob_page * logical_page_size + self.buffer_position as u64
+    }
+
+    /// Reads up to `max_len` bytes from the current position, but only as many as are available.
```
danlaine: I think this comment is stale -- there is no `max_len`

runtime/src/utils/buffer/pool/mod.rs
```diff
@@ -0,0 +1,475 @@
+//! Blob wrappers for reading and writing data with integrity guarantees, plus a buffer pool that
+//! manages read caching over the data.
+//!
+//! # Page-oriented structure
+//!
+//! Blob data is stored in _pages_, with the page size dictated by the managing buffer pool. A
+//! _page_ consists of up to `page_size-12` bytes of _logical_ data, and the last `page_size-12`
```
danlaine:
```suggestion
//! _page_ consists of up to `page_size-12` bytes of _logical_ data, and the last 12
```

runtime/src/utils/buffer/pool/append.rs
roberto-bayardo: changed to enum

runtime/src/utils/buffer/pool/append.rs
roberto-bayardo: done

runtime/src/utils/buffer/pool/mod.rs
roberto-bayardo: fixed

runtime/src/utils/buffer/pool/page_cache.rs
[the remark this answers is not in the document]
roberto-bayardo: 🤦 I somehow concluded that type didn't exist. Let me make (or make claude make) that change.

runtime/src/utils/buffer/pool/page_cache.rs
roberto-bayardo: good catch. I have moved the validation into the future. It will only return full pages.

runtime/src/utils/buffer/pool/read.rs
roberto-bayardo: Oof, the AI got that one by me :/ Thanks, fixed & your test case added.

runtime/src/utils/buffer/pool/append.rs
roberto-bayardo: Went ahead and removed all data.reserve calls, since we must go *over* capacity before triggering a flush anyway, and who knows how much over we'll go. Probably better just to let the vector do its thing. It will get shrunken as soon as it goes immutable anyway.

runtime/src/utils/buffer/pool/read.rs
roberto-bayardo: technically both are not-quite-correct. Changed to return (buf,0) when the provided buffer is size 0, and return BlobInsufficientLength when there is no more blob data.

storage/src/journal/contiguous/fixed.rs
roberto-bayardo: sigh, this *was* being handled in trim_tail which I unceremoniously removed completely. I have inserted the relevant rewind code directly and added a test case (There was a test case for this, but I also naively removed it too... it was not working as intended due to the checksum recovery flagging it as an integrity failure before it got to that point.)

runtime/fuzz/fuzz_targets/buffer.rs
```diff
@@ -320,9 +328,13 @@ fn fuzz(input: FuzzInput) {
...
 
-                FuzzOperation::AppendCloneBlob => {
+                FuzzOperation::AppendAsReader { buffer_size } => {
                     if let Some(ref append) = append_buffer {
-                        let _ = append.clone_blob().await;
+                        let buffer_size = NZUsize!((buffer_size as usize).clamp(1, MAX_SIZE));
+                        let _ = append
+                            .as_blob_reader(buffer_size)
+                            .await
+                            .expect("Failed to create blob reader");
```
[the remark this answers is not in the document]
roberto-bayardo: Hmm, I think we do want to panic. I see no reason why as_blob_reader should fail if the append wrapper was created successfully.  I certainly don't see it ever failing. We can revisit should it ever trigger.

runtime/src/utils/buffer/pool/append.rs
roberto-bayardo: nice, accepted with a few tweaks. also change "logical_page_size" to just "page_size" since "page_size" should always imply logical page size.

runtime/src/utils/buffer/pool/append.rs
roberto-bayardo: yes good point. clarified.

runtime/src/utils/buffer/pool/read.rs
roberto-bayardo: right, changed to `buf.len()`.

runtime/src/utils/buffer/pool/append.rs
```diff
@@ -0,0 +1,1725 @@
...
+    pub async fn to_mutable(&self) {
+        let mut buffer = self.buffer.write().await;
+        if !buffer.immutable {
+            return;
+        }
+        buffer.immutable = false;
+
+        // Enlarge the write buffer to the capacity that was specified when the blob was created.
+        let capacity = buffer.capacity;
+        buffer.data.reserve(capacity);
```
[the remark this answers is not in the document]
roberto-bayardo: removed all capacity reservations for the data buffer, it's not useful anyway.

runtime/src/utils/buffer/pool/append.rs
```diff
@@ -0,0 +1,1725 @@
...
+        let current_size = self.size().await;
+        if size > current_size {
+            return Err(Error::BlobInsufficientLength);
+        }
+
+        let logical_page_size = self.pool_ref.page_size();
+        let physical_page_size = logical_page_size + CRC_RECORD_SIZE;
+
+        // Flush any buffered data first to ensure we have a consistent state on disk.
+        self.sync().await?;
```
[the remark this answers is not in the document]
roberto-bayardo: good catch bot, fixed

runtime/src/utils/buffer/pool/append.rs
```diff
@@ -0,0 +1,1726 @@
...
+            // We need full_pages + 1 physical pages to hold the partial data.
+            // The partial page will be padded to full physical page size.
+            (full_pages + 1) * physical_page_size
+        } else {
+            // No partial page needed.
+            full_pages * physical_page_size
+        };
+
+        // Resize the underlying blob.
+        blob_guard.blob.resize(new_physical_size).await?;
```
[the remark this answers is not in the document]
roberto-bayardo: fixed

runtime/src/utils/buffer/pool/mod.rs
roberto-bayardo: fixed

runtime/src/utils/buffer/pool/append.rs
```diff
@@ -0,0 +1,1798 @@
...
+    /// If there is unwritten data in the buffer, it will be flushed and synced before returning.
+    pub async fn to_immutable(&self) -> Result<(), Error> {
+        // Flush any buffered data. When flush_internal returns, the write_at has completed
+        // and data is in the OS buffer.
+        let mut buf_guard = self.buffer.write().await;
+        if buf_guard.immutable {
+            return Ok(());
+        }
+        buf_guard.immutable = true;
+        self.flush_internal(buf_guard, true).await?;
```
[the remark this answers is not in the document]
roberto-bayardo: this is fine, if the flush fails, we have bigger problems. typically the program will abort.

runtime/src/utils/buffer/pool/append.rs
```diff
@@ -0,0 +1,1798 @@
...
+            total_pages_in_buffer.saturating_sub(1)
+        } else {
+            total_pages_in_buffer
+        };
+
+        // Identify protected regions based on the OLD partial page state
+        let protected_regions = Self::identify_protected_regions(old_partial_page_state.as_ref());
+
+        blob_state.current_page += full_pages_written as u64;
+        blob_state.partial_page_state = partial_page_state;
```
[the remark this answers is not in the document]
roberto-bayardo: in general we don't/can't preserve consistency in the event of write failures. These typically will abort the program.

runtime/src/utils/buffer/pool/append.rs
```diff
@@ -0,0 +1,1814 @@
...
+
+        blob_guard.current_page = full_pages;
+        buf_guard.offset = full_pages * logical_page_size;
+
+        if partial_bytes > 0 {
+            // There's a partial page. Read its data from disk.
+            let page_start = full_pages * physical_page_size;
+            let buf = vec![0u8; partial_bytes as usize];
+            let partial_data: Vec<u8> = blob_guard.blob.read_at(buf, page_start).await?.into();
+            buf_guard.data = partial_data;
```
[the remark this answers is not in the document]
roberto-bayardo: thanks, fixed & regression test added.

runtime/src/utils/buffer/pool/append.rs
```diff
@@ -0,0 +1,1814 @@
...
+        let logical_page_size = self.pool_ref.page_size();
+        let physical_page_size = logical_page_size + CRC_RECORD_SIZE;
+
+        // Flush any buffered data first to ensure we have a consistent state on disk.
+        self.sync().await?;
+
+        // Acquire both locks to prevent concurrent operations.
+        let mut buf_guard = self.buffer.write().await;
+        if buf_guard.immutable {
+            return Ok(());
```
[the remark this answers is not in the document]
roberto-bayardo: fixed + regression test added.

runtime/src/utils/buffer/pool/mod.rs
danlaine: Maybe you didn't push this change?

roberto-bayardo: how odd. re-fixed.

runtime/fuzz/fuzz_targets/buffer.rs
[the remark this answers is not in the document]
roberto-bayardo: this is fine, as_blob_reader shouldn't fail.

runtime/src/utils/buffer/pool/mod.rs
```diff
@@ -0,0 +1,476 @@
...
+    page.truncate(len as usize);
+
+    Ok(page)
+}
+
+/// Describes a CRC record stored at the end of a page.
+///
+/// The CRC accompanied by the larger length is the one that should be treated as authoritative for
+/// the page. Two checksums are stored so that partial pages can be written without overwriting a
+/// valid checksum for a committed previously committed partial page.
```
danlaine:
```suggestion
/// valid checksum for a previously committed partial page.
```

runtime/src/utils/buffer/pool/page_cache.rs
```diff
@@ -1,25 +1,29 @@
+//! A buffer pool for caching _locical_ pages of [Blob] data in memory. The buffer pool is unaware
```
danlaine:
```suggestion
//! A buffer pool for caching _logical_ pages of [Blob] data in memory. The buffer pool is unaware
```

runtime/src/utils/buffer/pool/read.rs
```diff
@@ -0,0 +1,464 @@
...
+        if buf.is_empty() {
+            return Ok((buf, 0));
+        }
+        let current_pos = self.position();
+        let blob_size = self.blob_size();
+        let available = (blob_size.saturating_sub(current_pos) as usize).min(buf.len());
+        if available == 0 {
+            return Err(Error::BlobInsufficientLength);
+        }
+        let available = available.min(buf.len());
```
danlaine: We already `min` with `buf.len()` above.

runtime/fuzz/fuzz_targets/buffer.rs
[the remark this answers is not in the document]
roberto-bayardo: this is still fine.

runtime/src/utils/buffer/pool/read.rs
roberto-bayardo: removed.

runtime/src/utils/buffer/pool/page_cache.rs
roberto-bayardo: fixed

runtime/src/utils/buffer/pool/mod.rs
roberto-bayardo: fixed

runtime/src/utils/buffer/pool/append.rs
```diff
@@ -0,0 +1,1937 @@
...
+            return Err(Error::ImmutableBlob);
+        }
+
+        if !buffer.append(buf) {
+            return Ok(());
+        }
+
+        // Buffer is over capacity, so we need to write data to the blob.
+        self.flush_internal(buffer, false).await
+    }
```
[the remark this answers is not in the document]
roberto-bayardo: Removed that from the function doc. Allocating a buffer large enough to trigger overflow here would cause problems well before this.

runtime/src/utils/buffer/pool/append.rs
```diff
@@ -0,0 +1,1936 @@
...
+    /// If there is unwritten data in the buffer, it will be flushed and synced before returning.
+    pub async fn to_immutable(&self) -> Result<(), Error> {
+        // Flush any buffered data. When flush_internal returns, the write_at has completed
+        // and data is in the OS buffer.
+        let mut buf_guard = self.buffer.write().await;
+        if buf_guard.immutable {
+            return Ok(());
+        }
+        buf_guard.immutable = true;
+        self.flush_internal(buf_guard, true).await?;
```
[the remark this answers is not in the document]
roberto-bayardo: This is fine, io errors aren't recoverable anyway.

runtime/fuzz/fuzz_targets/buffer.rs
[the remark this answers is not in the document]
roberto-bayardo: please stop flagging this, this is fine.

runtime/src/utils/buffer/pool/append.rs
```diff
@@ -0,0 +1,1936 @@
...
+            // The partial page will be padded to full physical page size.
+            (full_pages + 1) * physical_page_size
+        } else {
+            // No partial page needed.
+            full_pages * physical_page_size
+        };
+
+        // Resize the underlying blob.
+        blob_guard.blob.resize(new_physical_size).await?;
+        blob_guard.partial_page_state = None;
```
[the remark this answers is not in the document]
roberto-bayardo: I think this is fine. If you rewind over a checksum, then you've basically said it's no longer guaranteed durable.

runtime/src/lib.rs
patrick-ogrady, line 1: It was my mistake not to break this error up from the beginning...we should probably create an issue to do so?

runtime/src/utils/buffer/tip.rs
```diff
@@ -16,15 +14,20 @@ pub(super) struct Buffer {
 
     /// The maximum size of the buffer.
     pub(super) capacity: usize,
+
+    /// Whether this buffer should allow new data.
```
patrick-ogrady, line 18: nit: format is typically `TODO (#2731): description of issue`

You should also make it `//`

storage/src/archive/benches/utils.rs
```diff
@@ -20,8 +20,8 @@ const ITEMS_PER_SECTION: u64 = 16_384;
 /// Number of bytes to buffer when replaying a [commonware_runtime::Blob].
 const REPLAY_BUFFER: usize = 1024 * 1024; // 1MB
 
-/// Page size for the index buffer pool.
-const PAGE_SIZE: NonZeroUsize = NZUsize!(4_096);
+/// Use a "prod sized" page size to test the performance of the journal.
```
patrick-ogrady: nit: I don't know if we need to claim `prod sized` (maybe you'd want smaller/bigger?

I think past comment is sufficient?

patrick-ogrady (commented):
review on periphery

storage/src/journal/segmented/fixed.rs
```diff
@@ -100,13 +93,9 @@ impl<E: Storage + Metrics, A: CodecFixed<Cfg = ()>> Journal<E, A> {
...
-        // Pre-allocate exact size and write directly to avoid copying
-        let mut buf: Vec<u8> = Vec::with_capacity(Self::CHUNK_SIZE);
-        item.write(&mut buf);
-        let checksum = crc32fast::hash(&buf);
-        buf.put_u32(checksum);
-
-        blob.append(buf).await?;
+        // Encode the item
+        let buf = item.encode_mut();
+        blob.append(&buf).await?;
```
patrick-ogrady, line 98: Should we instead "hand over" these bytes rather than a ref?

storage/src/journal/segmented/fixed.rs
```diff
@@ -154,18 +143,12 @@ impl<E: Storage + Metrics, A: CodecFixed<Cfg = ()>> Journal<E, A> {
...
-    fn verify_integrity(buf: &[u8]) -> Result<A, Error> {
-        let stored_checksum =
-            u32::from_be_bytes(buf[A::SIZE..].try_into().expect("checksum is 4 bytes"));
-        let checksum = crc32fast::hash(&buf[..A::SIZE]);
-        if checksum != stored_checksum {
-            return Err(Error::ChecksumMismatch(stored_checksum, checksum));
-        }
-        A::decode(&buf[..A::SIZE]).map_err(Error::Codec)
+    /// Decode an item from the buffer.
+    fn decode_item(buf: &[u8]) -> Result<A, Error> {
```
patrick-ogrady: nit: may be better to just call decode directly?

storage/src/journal/segmented/glob.rs
patrick-ogrady, line 1: glob doesn't use the underlying journal...there should be no modifications here?

storage/src/journal/segmented/oversized.rs
```diff
@@ -2392,9 +2424,20 @@ mod tests {
     fn test_recovery_entry_with_overflow_offset() {
         // Tests that an entry with offset near u64::MAX that would overflow
         // when added to size is detected as invalid during recovery.
+        // This tests semantic validation, not just integrity - the page CRC is valid
```
patrick-ogrady: combine summary here?

storage/src/journal/segmented/variable.rs
```diff
@@ -164,43 +164,38 @@ impl<E: Storage + Metrics, V: Codec> Journal<E, V> {
...
-        let mut hasher = crc32fast::Hasher::new();
-        let varint_buf = blob.read_at(vec![0; MIN_ITEM_SIZE], offset).await?;
-        let mut varint = varint_buf.as_ref();
-        let size = UInt::<u32>::read(&mut varint).map_err(Error::Codec)?.0 as usize;
-        let varint_len = MIN_ITEM_SIZE - varint.remaining();
-        hasher.update(&varint_buf.as_ref()[..varint_len]);
-        let offset = offset
+        // Read varint header (max 5 bytes for u32)
+        let buf = vec![0u8; MAX_VARINT_SIZE];
+        let (buf, available) = blob.read_up_to(buf, offset).await?;
```
patrick-ogrady, line 220: ah, nice

storage/src/journal/segmented/variable.rs
```diff
@@ -219,52 +214,44 @@ impl<E: Storage + Metrics, V: Codec> Journal<E, V> {
...
+        // Calculate the correct next offset
+        let next_offset = offset
+            .checked_add(varint_len as u64)
+            .ok_or(Error::OffsetOverflow)?
+            .checked_add(size as u64)
+            .ok_or(Error::OffsetOverflow)?;
+
+        // Get item bytes - either from buffer directly or by reading more
+        let extra_bytes = available - varint_len;
+        let item_data: Cow<'_, [u8]> = if extra_bytes >= size {
```
patrick-ogrady: We should be able to create a helper out of this varint read into next? I've seen some very similar code in a few places?

patrick-ogrady (commented):
more review

runtime/fuzz/fuzz_targets/blob_integrity.rs
```diff
@@ -0,0 +1,222 @@
...
+    /// Number of bytes to read.
+    len: u16,
+    /// Whether to use the Read wrapper (true) or Append.read_at (false).
+    use_reader: bool,
+}
+
+fn fuzz(input: FuzzInput) {
+    let executor = deterministic::Runner::default();
+    executor.start(|context| async move {
+        // Use dynamic page size from input (1-255).
```
patrick-ogrady: You can clamp this during generation by manually implementing arbitrary?

runtime/fuzz/fuzz_targets/blob_integrity.rs
```diff
@@ -0,0 +1,222 @@
...
+        let page_size = input.page_size.max(1) as u64;
+        let physical_page_size = page_size + CRC_SIZE;
+        // Pool capacity is number of pages to cache (1-10).
+        let pool_capacity = (input.pool_capacity % 10 + 1) as usize;
+        let pool_ref = PoolRef::new(NZU16!(page_size as u16), NZUsize!(pool_capacity));
+
+        // Determine how many bytes to write (1 to 10 pages worth).
+        let max_bytes = 10 * page_size as usize;
+        let logical_size = (input.num_bytes as usize).clamp(1, max_bytes) as u64;
```
patrick-ogrady: Clamp during generation?

storage/src/archive/benches/utils.rs
roberto-bayardo: yup reverted

runtime/src/utils/buffer/tip.rs
roberto-bayardo: done

runtime/fuzz/fuzz_targets/blob_integrity.rs
```diff
@@ -0,0 +1,222 @@
...
+            let end_page = (offset + len as u64 - 1) / page_size;
+            let read_touches_corrupted_page =
+                start_page <= corrupted_page && corrupted_page <= end_page;
+
+            if read_op.use_reader {
+                // Use as_blob_reader.
+                // Note: The Read wrapper buffers multiple pages at once, so corruption on ANY
+                // page in the buffer can cause a read to fail - not just the page being accessed.
+                // We can only verify that successful reads return correct data.
+                let reader_result = append.as_blob_reader(NZUsize!(256)).await;
```
patrick-ogrady: nit: make this number a const?

runtime/src/utils/buffer/pool/mod.rs
```diff
@@ -0,0 +1,476 @@
...
+        if computed_crc != crc {
+            debug!("Invalid fallback CRC: doesn't match page contents.");
+            return false;
+        }
+
+        true
+    }
+
+    /// Returns the CRC record in its storage representation.
+    fn to_bytes(&self) -> [u8; CRC_RECORD_SIZE as usize] {
```
patrick-ogrady: nit: use `Codec`?

runtime/src/utils/buffer/pool/mod.rs
```diff
@@ -0,0 +1,476 @@
...
+        let crc_start = physical_page_size - CRC_RECORD_USIZE;
+        page[crc_start..].copy_from_slice(&record.to_bytes());
+
+        // Should fail validation since primary is invalid and no fallback exists
+        let validated = CrcRecord::validate_page(&page);
+        assert!(
+            validated.is_none(),
+            "Should fail when primary is invalid and fallback has len=0"
+        );
+    }
```
patrick-ogrady, line 494: nit: add `conformance`

storage/src/journal/contiguous/fixed.rs
```diff
@@ -172,21 +172,16 @@ impl<E: Storage + Metrics, A: CodecFixed<Cfg = ()>> Journal<E, A> {
...
         if !blobs.is_empty() {
             let mut it = blobs.keys().rev();
             let mut prev_index = *it.next().unwrap();
             for index in it {
-                let (_, size) = blobs.get(index).unwrap();
                 if *index != prev_index - 1 {
                     return Err(Error::MissingBlob(prev_index - 1));
                 }
                 prev_index = *index;
-                if *size != full_size {
```
patrick-ogrady, line 186: Why remove this? I think we still want to check this?

patrick-ogrady: Ah, you check blow. Disregard

storage/src/journal/segmented/glob.rs
roberto-bayardo: ai gone wild. reverted after restoring ChecksumMismatch error.

runtime/src/utils/buffer/pool/append.rs
```diff
@@ -0,0 +1,1936 @@
...
+    }
+
+    /// Convert this blob to the mutable state if it's not already in it.
+    pub async fn to_mutable(&self) {
+        let mut buffer = self.buffer.write().await;
+        if !buffer.immutable {
+            return;
+        }
+        buffer.immutable = false;
+    }
```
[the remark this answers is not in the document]
roberto-bayardo: this is fine.

runtime/src/utils/buffer/pool/mod.rs
roberto-bayardo: don't understand?

runtime/src/utils/buffer/pool/append.rs
```diff
@@ -0,0 +1,1936 @@
...
+        }
+
+        // Sync the OS buffer to disk to ensure new_immutable on restart will succeed even in the
+        // event of a crash.
+        let blob_state = self.blob_state.read().await;
+        blob_state.blob.sync().await
+    }
+
+    /// Convert this blob to the mutable state if it's not already in it.
+    pub async fn to_mutable(&self) {
```
patrick-ogrady, line 248: When do we need to restore `to_mutable`? Could we not just have `mutable` -> `immutable`?

runtime/src/utils/buffer/pool/append.rs
```diff
@@ -0,0 +1,1936 @@
...
+        }
+
+        // Buffer is over capacity, so we need to write data to the blob.
+        self.flush_internal(buffer, false).await
+    }
+
+    /// Flush all full pages from the buffer to disk, resetting the buffer to contain only the bytes
+    /// in any final partial page. If `write_partial_page` is true, the partial page will be written
+    /// to the blob as well along with a CRC record.
+    async fn flush_internal(
```
patrick-ogrady, line 344: I think one future optimization here is to allow a flush of multiple contiguous pages to do a single write_at?

runtime/src/utils/buffer/pool/append.rs
```diff
@@ -0,0 +1,1936 @@
...
+    async fn sync(&self) -> Result<(), Error> {
+        // Flush any buffered data, including any partial page. When flush_internal returns, the
+        // write_at has completed and data is in the OS buffer.
+        let buf_guard = self.buffer.write().await;
+        if buf_guard.immutable {
+            return Ok(());
+        }
+        self.flush_internal(buf_guard, true).await?;
+
+        // Sync the OS buffer to disk. We need the blob read lock here since sync() requires access
```
patrick-ogrady: nit: "OS buffer" is probably the wrong language?

patrick-ogrady (commented):
another batch

runtime/src/utils/buffer/pool/mod.rs
```diff
@@ -0,0 +1,476 @@
...
+    Ok(page)
+}
+
+/// Describes a CRC record stored at the end of a page.
+///
+/// The CRC accompanied by the larger length is the one that should be treated as authoritative for
+/// the page. Two checksums are stored so that partial pages can be written without overwriting a
+/// valid checksum for a previously committed partial page.
+#[derive(Clone)]
+struct CrcRecord {
```
patrick-ogrady: nit: `Checksum`

runtime/src/utils/buffer/pool/mod.rs
```diff
@@ -0,0 +1,476 @@
...
+            (self.len1, self.crc1)
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    const CRC_RECORD_USIZE: usize = 12;
```
patrick-ogrady: duplicate?

runtime/src/utils/buffer/pool/append.rs
patrick-ogrady: We can do this later.

runtime/src/utils/buffer/pool/append.rs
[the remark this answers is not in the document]
patrick-ogrady: A write flush failure would be considered fatal.

runtime/src/utils/buffer/pool/append.rs
```diff
@@ -0,0 +1,1936 @@
...
+            // There's a partial page. Read its data from disk with CRC validation.
+            let page_data =
+                super::get_page_from_blob(&blob_guard.blob, full_pages, logical_page_size).await?;
+
+            // Ensure the validated data covers what we need.
+            if (page_data.len() as u64) < partial_bytes {
+                return Err(Error::InvalidChecksum);
+            }
+
+            buf_guard.data = page_data.as_ref()[..partial_bytes as usize].to_vec();
```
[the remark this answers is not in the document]
patrick-ogrady: This is not relevant. A failure here would be fatal.

patrick-ogrady (approved):
I think there is still a lot to optimize here (and am not sure `runtime` is the right place for this fancy checksummed buffer), but the written disk format (page + checksum record) is correct and we can work with this.

PR #2669, opened 2026-01-02
[cryptography] add secp256r1 certificate scheme

andresilva:
This PR adds a secp256r1 certificate scheme to the cryptography crate, following the same pattern as the existing ed25519 and bls12381 multisig schemes.

The scheme is integrated into all consensus protocols and all test suites were updated. Unlike ed25519 and bls12381, secp256r1 does not benefit from batch verification, so the batcher verifies signatures eagerly as they arrive rather than batching them.

### DIFF
--- a/consensus/src/simplex/mod.rs
+++ b/consensus/src/simplex/mod.rs
@@ -337,6 +337,91 @@ mod tests {
     const PAGE_CACHE_SIZE: NonZeroUsize = NZUsize!(10);
     const TEST_QUOTA: Quota = Quota::per_second(NonZeroU32::MAX);
 
+    #[test]
+    fn test_interesting() {
+        let activity_timeout = ViewDelta::new(10);
+
+        // Genesis view is never interesting
+        assert!(!interesting(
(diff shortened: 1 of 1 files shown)


andresilva:
This doesn't actually break format, it touches a conformance.toml file to add a new entry.

consensus/src/simplex/actors/batcher/verifier.rs
patrick-ogrady, line 1: yuge ❤️

consensus/src/simplex/mod.rs
```diff
@@ -262,6 +270,10 @@ pub(crate) fn interesting(
     pending: View,
     allow_future: bool,
 ) -> bool {
+    // If the view is genesis, skip it, genesis doesn't have votes
+    if pending.is_zero() {
```
patrick-ogrady, line 274: Can you add a test for this?

patrick-ogrady:
Going to make the conformance label check optional to allow for cases like this.

PR #2671, opened 2026-01-02
[macros] add support for on_start/on_end to select_loop!

andresilva:
This PR extends the `select_loop!` macro with optional `on_start` and `on_end` blocks, enabling cleaner patterns for code that needs to run at the beginning or end of each loop iteration.

The `on_start` block executes before the select, allowing variable initialization that's visible to all arms. The `on_end` block runs after any non-shutdown arm completes but is skipped when shutdown triggers, which is useful for cleanup or post-processing logic. Both blocks share lexical scope with the select arms.

This allows migrating some more places to use the `select_loop!` macro.

### DIFF
--- a/macros/src/lib.rs
+++ b/macros/src/lib.rs
@@ -12,7 +12,7 @@ use proc_macro_crate::{crate_name, FoundCrate};
 use quote::{format_ident, quote, ToTokens};
 use syn::{
     parse::{Parse, ParseStream, Result},
-    parse_macro_input, Block, Error, Expr, Ident, ItemFn, LitStr, Pat, Token,
+    parse_macro_input, Error, Expr, Ident, ItemFn, LitStr, Pat, Token,
 };
 
 mod nextest;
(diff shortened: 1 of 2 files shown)


andresilva:
For the places where I updated the code to use `select_loop!` this is much easier to review by using difftastic locally (although this will ignore whitespace changes).

resolver/src/p2p/engine.rs
```diff
@@ -154,211 +154,202 @@ impl<
...
+                    self.fetcher.add_retry(key);
+                }
+            },
+            // Handle pending deadline
+            _ = deadline_pending => {
+                self.fetcher.fetch(&mut sender).await;
+            },
+            // Handle mailbox messages
+            msg = self.mailbox.next() => {
+                let Some(msg) = msg else {
```
patrick-ogrady, line 221: Next feature for this actor sprint would be to automatically handle this IMO

patrick-ogrady: https://github.com/commonwarexyz/monorepo/issues/2930

PR #2674, opened 2026-01-02
[consensus] Remove Relay

### LINKED ISSUE #2394: [consensus/simplex] Review `Relay::broadcast` Usage
If the `Automaton` proposes before the timeout, we invoke `Relay::broadcast` before issuing our own vote: https://github.com/commonwarexyz/monorepo/blob/29dcbae80524fe979c5b3e0427b82822d199de56/consensus/src/simplex/actors/voter/actor.rs#L799-L812

In https://github.com/commonwarexyz/monorepo/pull/2393, we persist the block in `marshal` at this point to ensure that if our node dies between this broadcast and notarization we have the block we built.

While "correct", this is very very nuanced (and error-prone). Why should `broadcast` be expected to persist anything? If this order is changed in `simplex` (as the average person would not assume broadcast == "store this"), we could see a concerning regression.

We should review whether or not this is the right approach (or if we should just assume that a block should be stored before returning from propose like we do for verify: https://github.com/commonwarexyz/monorepo/blob/29dcbae80524fe979c5b3e0427b82822d199de56/consensus/src/application/marshaled.rs#L410-L412

patrick-ogrady:
Closes: #2394


consensus/src/simplex/mocks/application.rs
patrick-ogrady: Remove this

consensus/src/simplex/mocks/application.rs
```diff
@@ -254,7 +240,10 @@ impl<E: Clock + RngCore + Spawner, H: Hasher, P: PublicKey> Application<E, H, P>
         self.verified.insert(digest);
 
         // Store pending payload
-        self.pending.insert(digest, payload.into());
+        self.pending.insert(digest, payload.clone().into());
```
patrick-ogrady: We no longer need pending?

consensus/src/ordered_broadcast/engine.rs
patrick-ogrady, line 778: Should we assume re-broadcast is handled by the application (for getting the payload around after epoch change)?

patrick-ogrady: Like for `certify`, we could add a `RetryableAutomaton` with `repropose` but that seems a little jank.

consensus/src/ordered_broadcast/mod.rs
```diff
@@ -1063,4 +1063,114 @@ mod tests {
     fn test_1k_ed25519() {
         run_1k(ed25519::fixture);
     }
+
+    fn requires_repropose<S, F>(fixture: F)
```
patrick-ogrady, line 1067: This is a useful test regardless of what we do here.

clabby:

Issue #2673, opened 2026-01-02
[storage & runtime] Avoid holding write lock on append buffer during flush.


roberto-bayardo:
The Append blob wrapper currently holds a write-lock on its internal buffer while flushing previous contents to the underlying blob, preventing concurrent reads and buffered writes. We should instead hold the lock only for the duration required to prepare the buffer contents for writing (similar to how Append wrapper's sync() works). This way while writes to the underlying blob are happening, further appending of new data to the buffer along with reads of buffered data can take place concurrently.

PR #2679, opened 2026-01-03
[runtime/buffer] avoid write-locking the Append buffer during blob writes

roberto-bayardo:
The append write buffer is currently write-locked while we write data to underlying blob, which prevents any concurrent reads or buffered writes. This PR introduces a separate IO lock for the underlying blob so that buffer reads and appends can be performed concurrently with blob writing.

Impact on existing benchmarks are neutral, however we don't have any benchmarks that perform concurrent reads & writes, which is where this change would shine.

Resolves: https://github.com/commonwarexyz/monorepo/issues/2673

### DIFF
--- a/runtime/src/utils/buffer/append.rs
+++ b/runtime/src/utils/buffer/append.rs
@@ -280,8 +280,8 @@ impl<B: Blob> Blob for Append<B> {
         // Acquire buffer lock first.
         // NOTE: We MUST acquire the buffer lock before the blob lock to avoid deadlocks with
         // `append`, which acquires buffer then blob (via `flush_internal`).
-        let mut guard = self.buffer.write().await;
-        let (buffer, blob_size) = &mut *guard;
+        let mut buf_guard = self.buffer.write().await;
+        let (buffer, blob_size) = &mut *buf_guard;
 
         let flush_data = self.prepare_flush_data(buffer, *blob_size).await;
 


runtime/src/utils/buffer/append.rs
```diff
@@ -164,13 +212,40 @@ impl<B: Blob> Blob for Append<B> {
...
+        let uncached_offset = offset + cached as u64;
+        let uncached_len = remaining - cached;
         self.pool_ref
-            .read(&self.blob, self.id, &mut buf.as_mut()[..remaining], offset)
+            .read(
+                &*blob_guard,
+                self.id,
+                &mut buf.as_mut()[cached..cached + uncached_len],
+                uncached_offset,
+            )
```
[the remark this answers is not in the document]
roberto-bayardo: This is fine -- if the user is resizing the blob, then concurrent readers that are trying to access data trimmed by any resize should expect read failures.

runtime/src/utils/buffer/append.rs
```diff
@@ -121,22 +170,20 @@ impl<B: Blob> Append<B> {
...
         if new_data_start > 0 {
             buf.drain(0..new_data_start);
         }
-        let new_data_len = buf.len() as u64;
-        self.blob.write_at(buf, *blob_size).await?;
-        *blob_size += new_data_len;
 
-        Ok(())
+        // Return the data to write, and the offset where to write it within the blob.
+        Some((buf, blob_size))
```
danlaine: Do we need to return `blob_size`? It's unchanged by this function, so the caller already has it.

runtime/src/utils/buffer/append.rs
```diff
@@ -164,13 +212,40 @@ impl<B: Blob> Blob for Append<B> {
...
+
+        // Release buffer lock before potential I/O.
+        drop(guard);
+
         if remaining == 0 {
             return Ok(buf);
         }
 
-        // If there are bytes remaining to be read, use the buffer pool to get them.
+        // Fast path: try to read *only from pool cache without acquiring blob lock. This allows
```
danlaine:
```suggestion
        // Fast path: try to read *only* from pool cache without acquiring blob lock. This allows
```

runtime/src/utils/buffer/append.rs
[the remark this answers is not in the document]
danlaine: Would you mind adding a comment to this effect on `resize`?

runtime/src/utils/buffer/append.rs
```diff
@@ -199,17 +278,30 @@ impl<B: Blob> Blob for Append<B> {
...
+        // NOTE: We MUST acquire the buffer lock before the blob lock to avoid deadlocks with
+        // `append`, which acquires buffer then blob (via `flush_internal`).
+        let mut guard = self.buffer.write().await;
+        let (buffer, blob_size) = &mut *guard;
 
-        // Flush any buffered bytes to the underlying blob. (Note that a fancier implementation
-        // might avoid flushing those bytes that are backed up over by the next step, if any.)
-        self.flush(buffer, blob_size).await?;
+        // Acquire blob write lock to prevent concurrent reads throughout the resize.
+        let blob_guard = self.blob.write().await;
```
danlaine, line 289: I think it would be safe to acquire this after `self.prepare_flush_data` like in `flush_internal`.

runtime/src/utils/buffer/append.rs
```diff
@@ -97,16 +108,54 @@ impl<B: Blob> Append<B> {
...
     /// the buffer pool.
-    async fn flush(&self, buffer: &mut Buffer, blob_size: &mut u64) -> Result<(), Error> {
-        // Take the buffered data, if any.
-        let Some((mut buf, offset)) = buffer.take() else {
+    ///
+    /// This method acquires the blob write lock before releasing the buffer lock, ensuring readers
+    /// that need blob access will wait for the write to complete.
+    async fn flush_internal(
+        &self,
+        mut guard: crate::RwLockWriteGuard<'_, (Buffer, u64)>,
```
danlaine: nit to make this parallel with `blob_guard` below and disambiguate the guards
```suggestion
        mut buf_guard: crate::RwLockWriteGuard<'_, (Buffer, u64)>,
```

runtime/src/utils/buffer/append.rs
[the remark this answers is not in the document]
roberto-bayardo: done

runtime/src/utils/buffer/append.rs
roberto-bayardo: fixed

runtime/src/utils/buffer/append.rs
roberto-bayardo: nope, AI slop. removed.

runtime/src/utils/buffer/append.rs
roberto-bayardo: done

runtime/src/utils/buffer/append.rs
roberto-bayardo: yup, done

runtime/src/utils/buffer/append.rs
```diff
@@ -184,32 +254,55 @@ impl<B: Blob> Blob for Append<B> {
...
         // within this function since it would be inaccessible anyway. Instead we ensure it is
         // always updated should the blob grow back to the point where we have new data for the same
         // page, if any old data hasn't expired naturally by then.
 
-        // Acquire a write lock on the buffer.
-        let (buffer, blob_size) = &mut *self.buffer.write().await;
+        // Acquire buffer lock first.
+        // NOTE: We MUST acquire the buffer lock before the blob lock to avoid deadlocks with
+        // `append`, which acquires buffer then blob (via `flush_internal`).
+        let mut guard = self.buffer.write().await;
```
danlaine:
```suggestion
        let mut buf_guard = self.buffer.write().await;
```

danlaine (approved):

runtime/src/utils/buffer/append.rs
roberto-bayardo: done

PR #2570, opened 2025-12-19
[p2p] Limit Block Duration

patrick-ogrady:
Closes: #2556 
Closes: #2557 

## TODO
- [ ] Add metric for block vs unregistered attempts

## Summary

This PR adds time-based block expiry for p2p peers. When a peer is blocked, the block automatically expires after a configurable duration (default: 1 hour for production, 1 minute for tests). This allows nodes that were blocked due to temporary issues (misconfiguration, one-off bugs) to reconnect without requiring restarts of all other nodes.

### Discovery Module
- Add `block_duration` configuration field
- `Address::Blocked` now stores `SystemTime` (block expiry time) instead of being a unit variant
- Record methods now take `now: SystemTime` parameter to check block expiry:
  - `blocked(now)`, `eligible(now)`, `want(now, ...)`, `dialable(now, ...)`, `acceptable(now)`, `reserve(now)`, `update(now, ...)`
- `Acceptable` enum (`Yes`, `Blocked`, `Unknown`) for detailed rejection logging
- Listener logs distinguish blocked peers from unknown peers

### Lookup Module
- Add `block_duration` configuration field
- `Address::Blocked` now stores `SystemTime` (block expiry time)
- Record methods updated similarly to discovery module
- `ListenerFilter` struct carries blocked IPs with expiry times to listener
- Listener checks blocked IPs before accepting connections

## Behavior

1. When a peer is blocked, the block expires after `block_duration`
2. After expiry, the peer becomes eligible again (can accept incoming connections)
3. Blocked peers are not dialable even after expiry (address info is lost on block)
4. Once a new peer set is registered containing the peer, their address is restored

PR #2687, opened 2026-01-04
[p2p] Add Blocked Peers to Metrics

patrick-ogrady:
Related: #2570

Change the blocked metric from a simple Gauge (count) to a Family<Peer, Gauge> that reports which specific peers are blocked and when each block expires. This enables debugging by showing exactly who is blocked and for how long.

- When blocked: set gauge to the expiry timestamp (epoch millis)
- When unblocked: remove the metric entry via Family.remove()
- Only currently blocked peers appear in metrics
- Remaining block time can be computed via PromQL: `blocked - time() * 1000`

### DIFF
--- a/p2p/src/authenticated/lookup/actors/tracker/directory.rs
+++ b/p2p/src/authenticated/lookup/actors/tracker/directory.rs
@@ -973,6 +973,12 @@ mod tests {
             // Verify peer is blocked
             assert_eq!(directory.blocked(), 1, "Should have one blocked peer");
 
+            // Get first expiry time
+            let first_expiry = directory
+                .blocked
+                .get(&pk_1)
+                .expect("peer should be blocked");
+
             // unblock_expired should return false before expiry
             assert!(
                 !directory.unblock_expired(),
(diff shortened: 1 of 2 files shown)


patrick-ogrady:
Would be cool if we could store with the expiry time?

p2p/src/authenticated/lookup/actors/tracker/directory.rs
```diff
@@ -1276,42 +1347,75 @@ mod tests {
...
-                directory.metrics.blocked.get(),
-                2,
-                "Blocking same nonexistent peer thrice should not increment metric"
+            assert!(
+                directory
+                    .metrics
+                    .blocked
+                    .get(&metrics::Peer::new(&unknown_pk))
+                    .is_some(),
+                "Blocking same nonexistent peer thrice should not change metric"
```
danlaine, line 1438: "thrice" 🧐

danlaine (approved):
Do we want to add a test that checks the expiry time in the `blocked`? e.g. Block a peer, get the expiry time, unblock them, re-block them, get the new expiry time and assert that it's greater than the old one.

patrick-ogrady:
> Do we want to add a test that checks the expiry time in the `blocked`? e.g. Block a peer, get the expiry time, unblock them, re-block them, get the new expiry time and assert that it's greater than the old one.

IIRC we already have tests that check this but if not, definitely.

p2p/src/authenticated/lookup/actors/tracker/directory.rs
patrick-ogrady: claudio can be quite eccentric

PR #2690, opened 2026-01-05
[consensus] clarify simplex engine journal behavior

lisenokdonbassenok:
Previously the Engine::start docs claimed that it “rebuilds the state of the engine from provided Journal”, but the engine does not accept a journal parameter and never touches storage directly. Persistence and replay are implemented inside the voter::Actor, which opens and replays its own journal on startup.


clabby:

PR #2693, opened 2026-01-05
[consensus/simplex] Cleanup Notarize/Notarization Check Prior to Finalize

### LINKED ISSUE #2692: [consensus/simplex] Document "notarize before finalize" requirement as intentional spec deviation
The Simplex implementation requires that a node has broadcast both its notarize vote AND the notarization certificate before it can broadcast a finalize vote.
From `consensus/src/simplex/actors/voter/round.rs`:
```
// If we haven't broadcast our notarize vote and notarization certificate, return None.if !self.broadcast_notarize || !self.broadcast_notarization {
    return None;
}
```

This is an intentional deviation from the Simplex specification which only requires:
> "If have not broadcast nullify(v), broadcast finalize(c,v)"

The spec does not require sending notarize(c,v) or notarization(c,v) before finalize(c,v). Other implementations (Ava Labs, Shoup) also do not have this requirement.

**Context**
This was identified in #1102 (observations #7 and #9). The current behavior is likely intentional for good reasons (e.g., consistency, observability, or safety), but these reasons are not documented.

**Proposed Action**
Document in the module-level docs (consensus/src/simplex/mod.rs) under "Deviations from Simplex Consensus" that:
1. This requirement exists
2. Why it was added (rationale)
3. Any implications for liveness or performance

BrendanChou:
Fixes #2692 

Remove some historical vote construction/broadcast checks that aren't functional (and just add confusion).

### DIFF
--- a/consensus/src/simplex/actors/voter/round.rs
+++ b/consensus/src/simplex/actors/voter/round.rs
@@ -1032,4 +1032,48 @@ mod tests {
         // This should not panic - the result is simply ignored.
         round.certified(true);
     }
+
+    #[test]
+    fn construct_finalize_requires_notarization() {
+        let mut rng = test_rng();
+        let namespace = b"ns";
+        let Fixture {
+            schemes, verifier, ..
+        } = ed25519::fixture(&mut rng, namespace, 4);
+        let local_scheme = schemes[0].clone();
+
+        let now = SystemTime::UNIX_EPOCH;
+        let round_info = Rnd::new(Epoch::new(1), View::new(1));
+        let proposal = Proposal::new(round_info, View::new(0), Sha256Digest::from([1u8; 32]));
+
+        let mut round = Round::new(local_scheme, round_info, now);
+        round.set_leader(Participant::new(0));
+        assert!(round.set_proposal(proposal.clone()));
+        assert!(round.verified());
+
+        // Construct notarize succeeds
+        assert!(round.construct_notarize().is_some());
+
+        // Certify the proposal before notarization. This should never happen in
+        // practice (we only call certify after notarization) but ensures the
+        // notarization check is functional.
+        round.certified(true);
+
+        // Construct finalize fails without notarization (even though certified)
+        assert!(round.construct_finalize().is_none());
+
+        // Add notarization
+        let notarization_votes: Vec<_> = schemes
+            .iter()
+            .map(|scheme| Notarize::sign(scheme, proposal.clone()).unwrap())
+            .collect();
+        let notarization =
+            Notarization::from_notarizes(&verifier, notarization_votes.iter(), &Sequential)
+                .unwrap();
+        let (added, _) = round.add_notarization(notarization);
+        assert!(added);
+
+        // Now construct finalize succeeds
+        assert!(round.construct_finalize().is_some());
+    }
 }


dnkolegov-ar (approved):

dnkolegov-ar:
from https://github.com/commonwarexyz/monorepo/issues/2692

> Document in the module-level docs (consensus/src/simplex/mod.rs) under "Deviations from Simplex Consensus" that:
> This requirement exists
> Why it was added (rationale)
> Any implications for liveness or performance

I do not see in this PR why this deviation was added, and what the implications are.

consensus/src/simplex/mod.rs
```diff
@@ -81,6 +81,8 @@
 //!   to ensure messages are reliably delivered is with a heavyweight reliable broadcast protocol).
 //! * Immediately broadcast `nullify(v)` if the leader's proposal fails verification (rather than waiting for a
 //!   timeout to fire).
+//! * Require broadcasting `notarize(c,v)` and `notarization(c,v)` before broadcasting `finalize(c,v)` to ensure
```
patrick-ogrady: IIRC we don't vote to finalize until we notarize? Or am I misrememebering?

BrendanChou: This is documenting that we don't vote to finalize until we notarize, which is not part of the original simplex spec (which allows you to finalize if you see a notarization even if you did not see the original proposal)

consensus/src/simplex/mod.rs
```diff
@@ -81,6 +81,12 @@
 //!   to ensure messages are reliably delivered is with a heavyweight reliable broadcast protocol).
 //! * Immediately broadcast `nullify(v)` if the leader's proposal fails verification (rather than waiting for a
 //!   timeout to fire).
+//! * Require broadcasting `notarize(c,v)` and `notarization(c,v)` before broadcasting `finalize(c,v)`.
+//!   This ensures that all participants who vote to `finalize` have verified `c`. Without this requirement,
+//!   a slow but correct node that receives `2f+1` notarize votes before verifying `c` could skip directly
```
patrick-ogrady: We consider a proposal verified if we get a certificate for it:
* https://github.com/commonwarexyz/monorepo/blob/cb63ed089b48ba37a6a09e344dffa2eb5d1ff0ab/consensus/src/simplex/actors/voter/round.rs#L377
* https://github.com/commonwarexyz/monorepo/blob/cb63ed089b48ba37a6a09e344dffa2eb5d1ff0ab/consensus/src/simplex/actors/voter/slot.rs#L126-L128

So, this explanation is not correct as written. We make sure to vote to ensure people know we are online IIRC.

patrick-ogrady: The other rationale I had for doing this was to "help" with finalization ASAP but if we want to "prefer" availability to finalize, I think that is a very reasonable consideration.

patrick-ogrady: Maybe we should tweak this to `verified or certificate` + `have block`?

patrick-ogrady: I guess I somewhat think of this as defensive rather than something we need to do (now that the verification is implicit from certificate). I suspect we could remove this check without issue (and we'd still broadcast notarize/notarization if we could).

consensus/src/simplex/actors/voter/round.rs
```diff
@@ -482,11 +483,6 @@ impl<S: Scheme, D: Digest> Round<S, D> {
         // Even if we've already seen a finalization, we still broadcast our finalize vote
         // in case it is useful (in the worst case it lets others observe we are alive).
 
-        // If we haven't broadcast our notarize vote and notarization certificate, return None.
-        if !self.broadcast_notarize || !self.broadcast_notarization {
```
patrick-ogrady, line 486: This was used (prior to a number of refactors) as a way to check that the proposal was verified (i.e. that we notarized) and that there existed a notarization (which we would've broadcast). It wasn't meant to imply anything stronger.

I've cleaned up the code a bit to reflect this.

consensus/src/simplex/actors/voter/round.rs
```diff
@@ -496,6 +492,9 @@ impl<S: Scheme, D: Digest> Round<S, D> {
             return None;
         }
 
+        // If there doesn't exist a notarization certificate, return None.
+        self.notarization.as_ref()?;
```
patrick-ogrady, line 496: While this should be implied by `certified` (we will only start certify if we have a notarization), I think writing it out is clearer.

patrick-ogrady (approved):

BrendanChou:
LGTM

Issue #2404, opened 2025-12-05
[storage] Add manifest / header for storage formats (magic + versioning)


andresilva:
Our storage backends persist raw blobs with per-record CRCs but no magic number or format identifier. That means opening the wrong partition or blob can look "valid" (or entirely "invalid" / corrupt) and runtime config changes (items per blob, compression, alignment, checksum scheme) aren't recorded on disk. As we approach a stable release, we need explicit on-disk identity, versioning, and guards so we can detect misuse early, reject incompatible formats, and support migrations without silent corruption.

### Ideas

- Add a fixed-size header at the start of every blob/file. Header fields: magic bytes, `db_kind` (e.g., journal/fixed, journal/segmented, etc.), on-disk format version, header CRC, payload checksum type, and essential format params (e.g. `items_per_blob`, compression flag/level, alignment, record size, expected key/value sizes, table size)
- Enforce strict validation on open: verify magic, kind, version, header CRC, and that stored params match runtime expectations, fail fast otherwise
- Think about migration paths per store (e.g. replay old blobs and rewrite with headers, or copy-and-swap to a fresh partition). Include a tool or helper to rewrite existing data, and document the steps for downstreams to upgrade cleanly, or just automatically migrate when opening an old version

andresilva:
Also if we decide to support format migrations, we should define what is our support policy, e.g. it's not reasonable to expect to be able to automatically migrate from any previous version to the latest version (that would inflict a huge maintenance burden on us). I think it's reasonable to support only point to point format migrations. If tooling is provided to migrate from point-to-point versions then downstream users can just run the migrations sequentially. We just need to ensure that it's still easy / possible to build the old migrations tools (this might seem trivial but sometimes isn't due to tooling / dependency rot).

patrick-ogrady:
Could just put a single byte version at start of file? May be able to avoid application changes if we wrap with a VersionedBlob or make part of runtime (that is a hot take)?

PR #2694, opened 2026-01-05
[runtime/storage] Versioned Blobs

danlaine:
Closes: #2404

* Adds an 8 byte header to the start of each `Blob`. The header is composed of 4 magic bytes, a u16 header version and a u16 application version. The application in this instance is whoever is using the `Blob`.
* On `Blob` open, check for existing header and validate it. If the blob doesn't exist, we write a header.
* `Storage::open` is now a default method which uses `Header::DEFAULT_APPLICATION_VERSION` (`0`) as the only acceptable version.
* All usages of `Blob`s therefore now use application version `0` (since they are opened with `Storage::open_versioned`).
* Files with length in `[1, 7]` are treated as corrupt (it should have at least the header) and overwritten on next open.

One thing I don't like about this PR is that all Blob implementations re-implement the logic for adding the header length to offset calculations. Is it worth making a wrapper type to handle that? Ehh I dunno.

### DIFF
--- a/runtime/src/storage/tokio/mod.rs
+++ b/runtime/src/storage/tokio/mod.rs
@@ -148,7 +148,7 @@ impl crate::Storage for Storage {
                 .map_err(|e| Error::BlobResizeFailed(partition.into(), hex(name), e))?;
             file.write_all(&header.encode())
                 .await
-                .map_err(|e| Error::BlobSyncFailed(partition.into(), hex(name), e))?;
+                .map_err(|_| Error::WriteFailed)?;
             file.sync_all()
                 .await
                 .map_err(|e| Error::BlobSyncFailed(partition.into(), hex(name), e))?;


runtime/src/storage/iouring.rs
```diff
@@ -431,4 +469,101 @@ mod tests {
         run_storage_tests(storage).await;
         let _ = std::fs::remove_dir_all(storage_directory);
     }
+
+    #[tokio::test]
+    async fn test_blob_header_handling() {
+        use crate::{Blob, Header, Storage as _};
```
patrick-ogrady: import at top of mod

runtime/src/lib.rs
```diff
@@ -554,8 +571,18 @@ pub trait Storage: Clone + Send + Sync + 'static {
...
+/// # Header
+///
+/// All blobs have a 32-byte [`Header`] at the start. The header is read on open
+/// (for existing blobs) or written (for new blobs). All I/O operations use logical
+/// offsets that start after the header; the header offset is handled internally.
+/// The header is currently unused and reserved for future use.
 #[allow(clippy::len_without_is_empty)]
 pub trait Blob: Clone + Send + Sync + 'static {
+    /// Returns the blob's header.
+    fn header(&self) -> Header;
```
patrick-ogrady: I wonder if we should define a callback in `storage::open` that provides the header (if exists) or asks for a new one?

runtime/src/storage/iouring.rs
danlaine: done

runtime/src/lib.rs
danlaine: Something like this?

```rs
storage.open("partition", name, |state| match state {
    HeaderState::New => Header::new(MAGIC, VERSION, MY_DB_KIND),
    HeaderState::Existing(h) => { validate(h)?; h },
    HeaderState::Corrupted => Header::default(), // or error
})
```

(Or with `Option` instead of enum  `HeaderState` if we don't care to distinguish "new" from "corrupt" (i.e. had less than a header written)).

I think that would be reasonable.

Should we put this in a follow-up PR, though? My main goal for this one was the breaking storage format change.

danlaine: On the other hand, perhaps the logic for setting the header should be internal to the storage implementation? What are you picturing the user-provided callback doing?

runtime/src/lib.rs
```diff
@@ -540,6 +540,23 @@ pub trait Storage: Clone + Send + Sync + 'static {
...
+/// This space is reserved for future use.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+pub struct Header {
+    /// Reserved bytes for future use.
+    pub bytes: [u8; Self::SIZE],
+}
+
+impl Header {
+    /// Size of the header in bytes.
+    pub const SIZE: usize = 32;
```
danlaine: I could be convinced that `64` is reasonable, if we want to give our future selves more breathing room. The overhead is negligible for reasonably large (say, > 1 MiB) blobs.

runtime/src/lib.rs
patrick-ogrady: > On the other hand, perhaps the logic for setting the header should be internal to the storage implementation? What are you picturing the user-provided callback doing?

I'm picturing the application that wants to open blobs setting this header or learning about it when opening.

I don't think it would be set by the runtime itself?

runtime/src/lib.rs
patrick-ogrady: I don't think this header will actually be the "application header" -> moreso just a good way to do magic bytes/version. So, I'd actually be willing to go down rather than up (maybe 8 bytes for magic bytes + u32 version)?

patrick-ogrady: Actually, I think 32 is reasonable to leave more room for the application to do what it wants with it.

danlaine: I'm not picturing this as an application header. Even so, I think adding all the fields in #2404 might require > 32 bytes.

> Header fields: magic bytes, db_kind (e.g., journal/fixed, journal/segmented, etc.), on-disk format version, header CRC, payload checksum type, and essential format params (e.g. items_per_blob, compression flag/level, alignment, record size, expected key/value sizes, table size)

runtime/src/utils/buffer/read.rs
```diff
@@ -11,10 +11,13 @@ use std::num::NonZeroUsize;
 /// use commonware_utils::NZUsize;
 /// use commonware_runtime::{Runner, buffer::Read, Blob, Error, Storage, deterministic};
 ///
+/// const BLOB_VERSION: u16 = 0;
+///
 /// let executor = deterministic::Runner::default();
 /// executor.start(|context| async move {
 ///     // Open a blob and add some data (e.g., a journal file)
-///     let (blob, size) = context.open("my_partition", b"my_data").await.expect("unable to open blob");
+///     let (blob, size, blob_version) = context.open("my_partition", b"my_data", BLOB_VERSION..=BLOB_VERSION).await.expect("unable to open blob");
```
patrick-ogrady: Ah, this is why I suggested `open` and `open_versioned`

danlaine: I'm not sold on the idea of having a default application version. I think it might be better to have the user explicitly provide the version rather than implicitly versioning the `Blob` for them.

storage/src/journal/segmented/variable.rs
```diff
@@ -1895,8 +1895,13 @@ mod tests {
...
 
-        async fn open(&self, _partition: &str, _name: &[u8]) -> Result<(MockBlob, u64), RError> {
-            Ok((MockBlob {}, self.len))
+        async fn open_versioned(
+            &self,
+            _partition: &str,
+            _name: &[u8],
+            _versions: std::ops::RangeInclusive<u16>,
+        ) -> Result<(MockBlob, u64, u16), RError> {
+            Ok((MockBlob {}, self.len, Header::DEFAULT_APPLICATION_VERSION))
```
danlaine: Technically this method should error if `_versions` doesn't contain `Header::DEFAULT_APPLICATION_VERSION` but I think it's fine that it doesn't?

runtime/src/storage/audited.rs
```diff
@@ -18,22 +18,31 @@ impl<S: crate::Storage> Storage<S> {
...
-    async fn open(&self, partition: &str, name: &[u8]) -> Result<(Self::Blob, u64), Error> {
+    async fn open_versioned(
+        &self,
+        partition: &str,
+        name: &[u8],
+        versions: std::ops::RangeInclusive<u16>,
+    ) -> Result<(Self::Blob, u64, u16), Error> {
         self.auditor.event(b"open", |hasher| {
             hasher.update(partition.as_bytes());
             hasher.update(name);
```
patrick-ogrady, line 29: nit: add versions to the auditor?

runtime/src/lib.rs
```diff
@@ -564,6 +609,90 @@ pub trait Storage: Clone + Send + Sync + 'static {
...
 }
 
+/// Fixed-size header at the start of each [Blob].
+///
+/// On-disk layout (8 bytes, big-endian):
+/// - Bytes 0-3: [Header::MAGIC]
+/// - Bytes 4-5: Header Version (u16)
+/// - Bytes 6-7: Application Version (u16)
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct Header {
```
patrick-ogrady: add to conformance?

runtime/src/storage/iouring.rs
```diff
@@ -124,27 +130,47 @@ impl crate::Storage for Storage {
...
-
-        // Only sync if we created a new file
-        if newly_created {
-            // Sync the blob to ensure it is durably created
-            blob.sync().await?;
+        let raw_len = file.metadata().map_err(|_| Error::ReadFailed)?.len();
+
+        // Handle header: new/corrupted blobs get a fresh header written,
+        // existing blobs have their header read.
+        let (app_version, logical_len) = if raw_len < Header::SIZE_U64 {
```
patrick-ogrady: I wonder if we can dedup with tokio better (as this doesn't actually use iouring at all?)

storage/src/journal/segmented/variable.rs
patrick-ogrady: I think we can add an assert.

storage/Cargo.toml
```diff
@@ -65,6 +65,7 @@ std = [
 arbitrary = [
 	"commonware-codec/arbitrary",
 	"commonware-cryptography/arbitrary",
+	"commonware-runtime?/arbitrary",
```
patrick-ogrady, line 68: what is this ?

runtime/src/lib.rs
```diff
@@ -564,6 +610,133 @@ pub trait Storage: Clone + Send + Sync + 'static {
...
```
patrick-ogrady: Should live in storage?

PR #2696, opened 2026-01-05
[runtime] Release lock before notifying in Blocker

syrupipy:
Found that we were notifying while holding the lock in Blocker, which defeats the point since the woken thread can't grab it anyway. Quick fix to drop the lock before the notify.

### DIFF
--- a/runtime/src/utils/mod.rs
+++ b/runtime/src/utils/mod.rs
@@ -216,10 +216,11 @@ impl Blocker {
 
 impl ArcWake for Blocker {
     fn wake_by_ref(arc_self: &Arc<Self>) {
+        // Mark as signaled (and release lock before notifying).
         {
             let mut signaled = arc_self.state.lock().unwrap();
             *signaled = true;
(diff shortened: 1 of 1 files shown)


runtime/src/utils/mod.rs
```diff
@@ -216,8 +216,10 @@ impl Blocker {
 
 impl ArcWake for Blocker {
     fn wake_by_ref(arc_self: &Arc<Self>) {
-        let mut signaled = arc_self.state.lock().unwrap();
-        *signaled = true;
+        {
+            let mut signaled = arc_self.state.lock().unwrap();
+            *signaled = true;
+        } // Release lock before notifying to avoid unnecessary contention
```
patrick-ogrady: Move this comment above the braces and update to:
```
// Mark as signaled (and release lock before notifying).
```

Nice find!

syrupipy: Sure!

Issue #2152, opened 2025-11-06
[consensus] create newtypes for Height and Participant


andresilva:
Similar to what was done in https://github.com/commonwarexyz/monorepo/pull/2121, we should create newtypes for `Height` and `Participant`. `Height` should be used for both block height as well as to replace the current `Index` in `aggregation` and `ordered_broadcast`.

PR #2700, opened 2026-01-05
[consensus] add Height type

andresilva:
Related #2152.

### DIFF
--- a/consensus/src/marshal/ingress/mailbox.rs
+++ b/consensus/src/marshal/ingress/mailbox.rs
@@ -425,14 +425,14 @@ impl<S: Scheme, B: Block> Stream for AncestorStream<S, B> {
 
     fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
         // Because marshal cannot currently yield the genesis block, we stop at height 1.
-        let end_bound = Height::new(1);
+        const END_BOUND: Height = Height::new(1);
 
         let mut this = self.project();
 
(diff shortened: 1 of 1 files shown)


consensus/src/aggregation/config.rs
```diff
@@ -57,8 +57,8 @@ pub struct Config<
     /// The number of chunks to process concurrently.
     pub window: NonZeroU64,
```
andresilva, line 58: Should we add `NonZeroHeightDelta`?

patrick-ogrady: For symmetry with View, I think that makes sense?

andresilva: We don't have `NonZeroViewDelta` 🙈

patrick-ogrady: ❗

consensus/src/marshal/ingress/mailbox.rs
```diff
@@ -425,14 +425,14 @@ impl<S: Scheme, B: Block> Stream for AncestorStream<S, B> {
 
     fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
         // Because marshal cannot currently yield the genesis block, we stop at height 1.
-        const END_BOUND: u64 = 1;
+        let end_bound = Height::new(1);
```
patrick-ogrady: Could we keep this `const`? `new` should be `const`?

patrick-ogrady (dismissed):

patrick-ogrady (approved):

PR #2702, opened 2026-01-05
[Storage] [Runtime] Add magic bytes to blob header

danlaine:
Adds magic bytes to the header at the beginning of each blob. This allows for detection of invalid (non-Commonware blob) files on open.

Partially resolves https://github.com/commonwarexyz/monorepo/issues/2404


runtime/src/lib.rs
```diff
@@ -542,19 +544,47 @@ pub trait Storage: Clone + Send + Sync + 'static {
...
     /// Size of the header in bytes.
     pub const SIZE: usize = 32;
 
     /// Size of the header as u64 for offset calculations.
     pub const SIZE_U64: u64 = Self::SIZE as u64;
+
+    pub const MAGIC_LENGTH: usize = 4;
+
+    /// Magic bytes identifying a valid commonware blob.
+    pub const MAGIC: [u8; Self::MAGIC_LENGTH] = *b"CWIC"; // CommonWare is CWIC
```
patrick-ogrady: nit: `commonware`

danlaine: I was going for capitalization to emphasize what the bacronym is but yeah, wouldn't want to confuse people. Also we can pick something less cute than this lol.

danlaine: Updated to `Commonware`

runtime/src/storage/iouring.rs
```diff
@@ -155,7 +155,9 @@ impl crate::Storage for Storage {
             let mut header_bytes = [0u8; Header::SIZE];
             file.read_exact(&mut header_bytes)
                 .map_err(|_| Error::ReadFailed)?;
-            (Header(header_bytes), raw_len - Header::SIZE_U64)
+            let header = Header(header_bytes);
+            header.validate_magic()?;
+            (header, raw_len - Header::SIZE_U64)
```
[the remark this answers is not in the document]
danlaine: Migrations are a non-issue for now

PR #2697, opened 2026-01-05
[consensus/simplex] Add Timeout then Certify Regression Test

patrick-ogrady:

PR #2703, opened 2026-01-05
[consensus/simplex] Remove Certify Retry + Cancel Outstanding Certifications

patrick-ogrady:
Related: #2697

Like `propose`/`verify`, we should call the application once with `certify` (and it should hang on unless canceled).

Once there exists a finalization, we should cancel all outstanding certification requests that are no longer necessary.

### DIFF
--- a/consensus/src/simplex/actors/voter/round.rs
+++ b/consensus/src/simplex/actors/voter/round.rs
@@ -30,14 +30,6 @@ enum CertifyState {
     Aborted,
 }
 
-/// Result of attempting to certify a round's proposal.
-pub enum CertifyResult<D: Digest> {
-    /// Ready for certification with the proposal.
-    Ready(Proposal<D>),
-    /// Cannot certify (no notarization, already certified/aborted, or equivocation).
-    Skip,
-}
-
 /// Per-[Rnd] state machine.
 pub struct Round<S: Scheme, D: Digest> {
     start: SystemTime,
@@ -137,14 +129,12 @@ impl<S: Scheme, D: Digest> Round<S, D> {
     }
 
     /// Attempt to certify this round's proposal.
-    pub fn try_certify(&mut self) -> CertifyResult<D> {
-        let Some(notarization) = self.notarization.as_ref() else {
-            return CertifyResult::Skip;
-        };
+    pub fn try_certify(&mut self) -> Option<Proposal<D>> {
+        let notarization = self.notarization.as_ref()?;
         match self.certify {
             CertifyState::Ready => {}
             CertifyState::Outstanding(_) | CertifyState::Certified(_) | CertifyState::Aborted => {
-                return CertifyResult::Skip;
+                return None;
             }
         }
 
@@ -154,12 +144,13 @@ impl<S: Scheme, D: Digest> Round<S, D> {
         let proposal = self
             .proposal
             .proposal()
+            .cloned()
             .expect("proposal must be set if notarization is set");
         assert_eq!(
-            proposal, &notarization.proposal,
+            &proposal, &notarization.proposal,
             "slot proposal must match notarization proposal"
         );
-        CertifyResult::Ready(proposal.clone())
+        Some(proposal)
     }
 
     /// Sets the handle for the certification request.
@@ -169,13 +160,10 @@ impl<S: Scheme, D: Digest> Round<S, D> {
 
     /// Aborts the in-flight certification request.
     pub fn abort_certify(&mut self) {
-        match &self.certify {
-            CertifyState::Certified(_) | CertifyState::Aborted => {}
-            CertifyState::Ready => self.certify = CertifyState::Aborted,
-            CertifyState::Outstanding(_) => {
-                drop(replace(&mut self.certify, CertifyState::Aborted));
-            }
+        if matches!(self.certify, CertifyState::Certified(_)) {
+            return;
         }
+        self.certify = CertifyState::Aborted;
     }
 
     /// Returns the elected leader (if any) for this round.
@@ -836,7 +824,7 @@ mod tests {
         assert!(round.verified());
 
         // No notarization yet - should skip
-        assert!(matches!(round.try_certify(), CertifyResult::Skip));
+        assert!(round.try_certify().is_none());
     }
 
     #[test]
@@ -868,7 +856,7 @@ mod tests {
         assert!(added);
 
         // First try_certify should succeed
-        assert!(matches!(round.try_certify(), CertifyResult::Ready(_)));
+        assert!(round.try_certify().is_some());
 
         // Set a certify handle then mark as certified
         let mut pool = AbortablePool::<()>::default();
@@ -877,7 +865,7 @@ mod tests {
         round.certified(true);
 
         // Second try_certify should skip - already certified
-        assert!(matches!(round.try_certify(), CertifyResult::Skip));
+        assert!(round.try_certify().is_none());
     }
 
     #[test]
@@ -909,7 +897,7 @@ mod tests {
         assert!(added);
 
         // First try_certify should succeed
-        assert!(matches!(round.try_certify(), CertifyResult::Ready(_)));
+        assert!(round.try_certify().is_some());
 
         // Set a certify handle (simulating in-flight certification)
         let mut pool = AbortablePool::<()>::default();
@@ -917,7 +905,7 @@ mod tests {
         round.set_certify_handle(handle);
 
         // Second try_certify should skip - handle exists
-        assert!(matches!(round.try_certify(), CertifyResult::Skip));
+        assert!(round.try_certify().is_none());
     }
 
     #[test]
@@ -954,13 +942,13 @@ mod tests {
         round.set_certify_handle(handle);
 
         // try_certify blocked by handle
-        assert!(matches!(round.try_certify(), CertifyResult::Skip));
+        assert!(round.try_certify().is_none());
 
         // Abort transitions to Aborted state
         round.abort_certify();
 
         // try_certify still blocked after abort (no re-certification allowed)
-        assert!(matches!(round.try_certify(), CertifyResult::Skip));
+        assert!(round.try_certify().is_none());
     }
 
     #[test]
@@ -992,7 +980,7 @@ mod tests {
 
         // Has notarization and proposal came from certificate
         // try_certify returns the proposal from the certificate
-        assert!(matches!(round.try_certify(), CertifyResult::Ready(_)));
+        assert!(round.try_certify().is_some());
     }
 
     #[test]
--- a/consensus/src/simplex/actors/voter/state.rs
+++ b/consensus/src/simplex/actors/voter/state.rs
@@ -1,4 +1,4 @@
-use super::round::{CertifyResult, Round};
+use super::round::Round;
 use crate::{
     simplex::{
         elector::{Config as ElectorConfig, Elector},
@@ -471,11 +471,7 @@ impl<E: Clock + CryptoRngCore + Metrics, S: Scheme<D>, L: ElectorConfig<S>, D: D
                 if view <= self.last_finalized {
                     return None;
                 }
-                let round = self.views.get_mut(&view)?;
-                match round.try_certify() {
-                    CertifyResult::Ready(proposal) => Some(proposal),
-                    CertifyResult::Skip => None,
-                }
+                self.views.get_mut(&view)?.try_certify()
             })
             .collect()
     }


andresilva (dismissed):

consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -972,12 +972,7 @@ impl<
                             }
                         }
                         Err(err) => {
-                            // The application did not explicitly respond whether certification succeeded.
-                            // Retry the certification request (we should never assume failure here because
-                            // we persist certification results to the journal).
```
BrendanChou, line 977: Keep part of this comment?

```
// The application did not explicitly respond whether certification succeeded.
// Treat this similarly to the application stalling (never responding).
// Do not assume failure here because we persist certification results to the journal.
```

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -254,6 +270,17 @@ impl<E: Clock + CryptoRngCore + Metrics, S: Scheme<D>, L: ElectorConfig<S>, D: D
...
+
+            // Prune certification candidates at or below finalized view
+            self.certification_candidates.retain(|v| *v > view);
+
+            // Abort outstanding certifications at or below finalized view
+            let keep = self.outstanding_certifications.split_off(&view.next());
+            for v in replace(&mut self.outstanding_certifications, keep) {
+                if let Some(round) = self.views.get_mut(&v) {
+                    round.abort_certify();
+                }
```
BrendanChou, line 275:
```
self.views.get_mut(&v).inspect_mut(|round| round.abort_certify());
```

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -117,6 +122,12 @@ impl<E: Clock + CryptoRngCore + Metrics, S: Scheme<D>, L: ElectorConfig<S>, D: D
         self.last_finalized
     }
 
+    /// Returns true if the view is eligible for certification (not yet finalized).
+    #[inline]
+    fn is_certify_candidate(&self, view: View) -> bool {
+        view > self.last_finalized
+    }
```
BrendanChou: nit: I would probably just remove this function, I think having it makes things less readable since the logic is simple (it's not clear what it does based on the title of the function so you have to go reference the function to understand what you're trying to do)

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -254,6 +270,17 @@ impl<E: Clock + CryptoRngCore + Metrics, S: Scheme<D>, L: ElectorConfig<S>, D: D
...
+            // Prune certification candidates at or below finalized view
+            self.certification_candidates.retain(|v| *v > view);
+
+            // Abort outstanding certifications at or below finalized view
+            let keep = self.outstanding_certifications.split_off(&view.next());
+            for v in replace(&mut self.outstanding_certifications, keep) {
+                if let Some(round) = self.views.get_mut(&v) {
+                    round.abort_certify();
+                }
+            }
```
BrendanChou, line 276: Is this pruning logic better suited for the `prune` function?

consensus/src/simplex/actors/voter/round.rs
```diff
@@ -129,23 +119,32 @@ impl<S: Scheme, D: Digest> Round<S, D> {
     }
 
     /// Attempt to certify this round's proposal.
-    pub fn should_certify(&mut self) -> CertifyResult<Proposal<D>> {
+    ///
+    /// Returns `(Some(proposal), _)` if ready for certification.
```
BrendanChou: `false` instead of `_`

consensus/src/simplex/actors/voter/round.rs
```diff
@@ -129,23 +119,32 @@ impl<S: Scheme, D: Digest> Round<S, D> {
...
+    /// Returns `(None, true)` if pending (has notarization but no proposal yet).
```
BrendanChou: Wait how can there be a notarization but no proposal? Shouldn't the proposal be know if there is a notarization?

consensus/src/simplex/actors/voter/round.rs
```diff
@@ -782,4 +776,185 @@ mod tests {
...
+        let handle = pool.push(futures::future::pending());
+        round.set_certify_handle(handle);
+
+        // Second try_certify should fail - handle exists, not pending
+        let (proposal, pending) = round.try_certify();
+        assert!(proposal.is_none());
+        assert!(!pending);
+    }
+
+    #[test]
```
patrick-ogrady, line 911: We should be more defensive here and fail (shouldn't allow multiple times)

consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -972,6 +933,9 @@ impl<
                             }
                         }
                         Err(err) => {
+                            // Unlike propose/verify (where failing to act will lead to a timeout
+                            // and subsequent nullification), failing to certify can lead to a halt
+                            // because we'll never exit the view without a notarization + certification.
                             debug!(?err, ?round, "failed to certify proposal");
```
BrendanChou, line 942: should this be increased to a higher log level if this could lead to a halt?

I think it really only should be dropped by the application if it's shutting down

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -433,20 +460,37 @@ impl<E: Clock + CryptoRngCore + Metrics, S: Scheme<D>, L: ElectorConfig<S>, D: D
...
+                continue;
+            };
+            let (proposal, pending) = round.try_certify();
+            if let Some(proposal) = proposal {
+                ready.push(proposal);
+            } else if pending {
+                self.certification_candidates.insert(view);
+            }
+        }
+        ready
```
BrendanChou:
```
    let candidates = std::mem::take(&mut self.certification_candidates);
    candidates.into_iter().filter_map(|view| {
        if !self.is_certify_candidate(view) { return None; }
        
        let round = self.views.get_mut(&view)?;
        let (proposal, pending) = round.try_certify();

        if pending {
            self.certification_candidates.insert(view);
        }

        proposal
    }).collect()
```

consensus/src/simplex/actors/voter/round.rs
BrendanChou: This is why I thought the enum was slightly more clear despite the type proliferation

patrick-ogrady: I think you're right

consensus/src/simplex/actors/voter/actor.rs
patrick-ogrady: I think this could lead to shutdown being a bit noisy (feels odd to warn there because a correct application may see it unnecessarily and wonder if something is wrong).

I prefer keeping as-is and adding warn later if this becomes an issue.

consensus/src/simplex/actors/voter/round.rs
patrick-ogrady: We do that if there is equivocation (I think)?

patrick-ogrady: We could instead take the position we should still try to help the certificate even after equivocation...maybe we need to 🤔

consensus/src/simplex/actors/voter/state.rs
patrick-ogrady: `prune()` only happens at the activity timeout (rather than right when finalized).

consensus/src/simplex/actors/voter/state.rs
patrick-ogrady: Added this to make it more readable but will revert

consensus/src/simplex/actors/voter/round.rs
patrick-ogrady: I think we may need to (certify here actually)

consensus/src/simplex/actors/voter/state.rs
patrick-ogrady: not yet in stable rust 😢

consensus/src/simplex/actors/voter/round.rs
patrick-ogrady: Not the bug I thought but there was a bug: https://github.com/commonwarexyz/monorepo/pull/2703/changes/ad7e68450fd20882b5925644e165913b30b8114b

consensus/src/simplex/actors/voter/round.rs
```diff
@@ -138,21 +138,16 @@ impl<S: Scheme, D: Digest> Round<S, D> {
...
                 return CertifyResult::Skip;
             }
         }
-        // Proposal comes from the notarization certificate. If unavailable, it's
-        // due to equivocation which is permanent - skip rather than retry.
-        self.proposal
-            .proposal()
-            .cloned()
-            .map_or(CertifyResult::Skip, CertifyResult::Ready)
+        CertifyResult::Ready(notarization.proposal.clone())
```
patrick-ogrady: nit: add a comment explaining why

BrendanChou: put this inside the match statement?

consensus/src/simplex/actors/voter/round.rs
patrick-ogrady: Disregard. Not a bug.

consensus/src/simplex/actors/voter/round.rs
patrick-ogrady: No longer applies

consensus/src/simplex/actors/voter/round.rs
```diff
@@ -251,12 +284,14 @@ impl<S: Scheme, D: Digest> Round<S, D> {
...
-            "certification should not conflict"
-        );
-        self.certified = Some(is_success);
-        self.certify_handle = None;
+        match &self.certify {
+            CertifyState::Certified(v) => {
+                assert_eq!(*v, is_success, "certification should not conflict");
+                return;
+            }
+            CertifyState::Ready | CertifyState::Outstanding(_) | CertifyState::Aborted => {}
```
patrick-ogrady, line 280: We opt to track a certification result if we have it!

consensus/src/simplex/actors/voter/round.rs
```diff
@@ -11,23 +11,33 @@ use std::{
...
+    Aborted,
+}
+
+/// Result of attempting to certify a round's proposal.
+pub enum CertifyResult<D: Digest> {
+    /// Ready for certification with the proposal.
+    Ready(Proposal<D>),
+    /// Cannot certify (no notarization, already certified/aborted, or equivocation).
+    Skip,
+}
```
BrendanChou: In this case I would probably just use `Option<Proposal<D>>` to reduce the number of types

consensus/src/simplex/actors/voter/round.rs
```diff
@@ -129,26 +137,45 @@ impl<S: Scheme, D: Digest> Round<S, D> {
...
-    /// Clears the handle for the certification request.
-    pub fn unset_certify_handle(&mut self) {
-        self.certify_handle = None;
+    /// Aborts the in-flight certification request.
+    pub fn abort_certify(&mut self) {
+        match &self.certify {
+            CertifyState::Certified(_) | CertifyState::Aborted => {}
+            CertifyState::Ready => self.certify = CertifyState::Aborted,
+            CertifyState::Outstanding(_) => {
+                drop(replace(&mut self.certify, CertifyState::Aborted));
```
BrendanChou: why is explicit drop necessary?
to simplify, maybe just return early if `::Certified(_)`, otherwise set the value to `::Aborted`?

consensus/src/simplex/actors/voter/round.rs
```diff
@@ -129,26 +137,45 @@ impl<S: Scheme, D: Digest> Round<S, D> {
...
+            .proposal
             .proposal()
-            .cloned()
-            .map_or(CertifyResult::Pending, CertifyResult::Ready)
+            .expect("proposal must be set if notarization is set");
+        assert_eq!(
+            proposal, &notarization.proposal,
+            "slot proposal must match notarization proposal"
+        );
+        CertifyResult::Ready(proposal.clone())
```
BrendanChou: Simpler to just use `notarization.proposal.clone()` rather than doing these asserts/expects?

BrendanChou (dismissed):

consensus/src/simplex/actors/voter/round.rs
patrick-ogrady: This is a remnant of when explicitly dropping, will fix.

consensus/src/simplex/actors/voter/round.rs
patrick-ogrady: I added the assert just for our own sanity...but agree I can clone right out of self.propose to clean up.

consensus/src/simplex/actors/voter/round.rs
patrick-ogrady: Yeah, now simpler.

consensus/src/simplex/actors/voter/round.rs
```diff
@@ -129,26 +137,45 @@ impl<S: Scheme, D: Digest> Round<S, D> {
...
+        let proposal = self
+            .proposal
             .proposal()
-            .cloned()
-            .map_or(CertifyResult::Pending, CertifyResult::Ready)
+            .expect("proposal must be set if notarization is set");
+        assert_eq!(
+            proposal, &notarization.proposal,
+            "slot proposal must match notarization proposal"
+        );
```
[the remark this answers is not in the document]
patrick-ogrady: * Only one non-recovered proposal can enter the slot (from self via built() or from leader via set_proposal())
* Any conflicting proposal arrives via certificate (recovered=true)
* The certificate path already replaces the slot's proposal with the certificate's proposal

BrendanChou (approved):

PR #2704, opened 2026-01-05
Store RNG in deterministic runtime Config

### LINKED ISSUE #2376: [runtime] Allow controlling bytes of the RNG in tests directly

cronokirby:
- Modifies deterministic runtime Config to store an RNG instead of a seed.
- Makes the Config _not_ implement Clone in order to avoid confusion about shared randomness.

Closes #2376.

I think dynamic dispatch is the right choice here.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1447,6 +1447,7 @@ dependencies = [
  "pin-project",
  "prometheus-client",
  "rand 0.8.5",
+ "rand_core 0.6.4",
  "rayon",
  "sha2 0.10.9",
  "sysinfo",
--- a/runtime/Cargo.toml
+++ b/runtime/Cargo.toml
@@ -28,6 +28,7 @@ opentelemetry.workspace = true
 pin-project = { workspace = true, optional = true }
 prometheus-client.workspace = true
 rand.workspace = true
+rand_core.workspace = true
 rayon.workspace = true
 sha2.workspace = true
 sysinfo.workspace = true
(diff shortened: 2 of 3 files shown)


cronokirby:
This is similar to #2470, but opts to instead move this to the config level.

p2p/src/authenticated/lookup/mod.rs
```diff
@@ -72,7 +72,7 @@
 //!
 //! // Configure context
 //! let runtime_cfg = deterministic::Config::default();
-//! let runner = deterministic::Runner::new(runtime_cfg.clone());
```
patrick-ogrady, line 75: oof

evonide:
@cronokirby thanks for implementing this one! I overestimated the impact of removing `Clone` on Config it seems

cronokirby:
> @cronokirby thanks for implementing this one! I overestimated the impact of removing `Clone` on Config it seems

Yeah I was surprised as well

patrick-ogrady:
Got booted from the mq for what looks like a mac runner issue. Retrying.

Issue #1921, opened 2025-10-19
[consensus] Build `pvtop`-like Tool


patrick-ogrady:
To monitor/debug consensus progress, we should build a [pvtop](https://github.com/blockpane/pvtop)/[tmtop](https://github.com/QuokkaStake/tmtop)-like tool to visualize progress.

Eventhough we don't have attribution in the threshold setting, we should still be able to indicate who the leader is and what messages we've received.

patrick-ogrady:
The simplest way to integrate this would be to plugin to the reporter stream (wouldn't require any consensus changes whatsoever).

patrick-ogrady:
May be a good idea to replace the TUI for `examples/log` with this.

patrick-ogrady:
#2708 is good enough for how you'd track via prometheus...leaving open to build a TUI version

patrick-ogrady:
To close this, we should move WebSocket Reporter into an existing example to observe.

This could just be another binary you could run in examples/reshare.

patrick-ogrady:
Or for `alto`, if preferred?

patrick-ogrady:
With tracing now, this should be trivial.

PR #2708, opened 2026-01-06
[consensus/simplex] Add Voting Metrics

patrick-ogrady:
Related: #1921 (production-use adaption)

Add a family gauge metric to track the view of the last vote received from each participant in the batcher actor. This enables monitoring peer activity and computing instantaneous lag (current_view - last_vote).

- Add Peer label struct to metrics.rs
- Initialize all participants to view 0 at startup
- Update metric when valid votes are received
- Add test verifying metric tracking including late votes after quorum

### DIFF
--- a/consensus/src/simplex/actors/batcher/actor.rs
+++ b/consensus/src/simplex/actors/batcher/actor.rs
@@ -399,12 +399,6 @@ impl<
                         continue;
                     }
 
-                    // Update per-peer latest vote metric (only if higher than current)
-                    let _ = self
-                        .latest_vote
-                        .get_or_create(&Peer::new(&sender))
-                        .try_set_max(message.view().get());
-
                     // If the view isn't interesting, we can skip
                     let view = message.view();
                     if !interesting(
@@ -418,12 +412,19 @@ impl<
                     }
 
                     // Add the vote to the verifier
+                    let peer = Peer::new(&sender);
                     if work
                         .entry(view)
                         .or_insert_with(|| self.new_round())
                         .add_network(sender, message)
                         .await {
                             self.added.inc();
+
+                            // Update per-peer latest vote metric (only if higher than current)
+                            let _ = self
+                                .latest_vote
+                                .get_or_create(&peer)
+                                .try_set_max(view.get());
                         }
                     updated_view = view;
                 },


consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -384,6 +400,12 @@ impl<
                         continue;
                     }
 
+                    // Update per-peer last vote metric
```
patrick-ogrady: TODO: update with latest

patrick-ogrady:
While honest voters should never vote for views lower than latest, we want this metric to be robust to such behavior.

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -79,6 +85,16 @@ impl<
             "number of inbound messages",
             inbound_messages.clone(),
         );
+        let latest_vote = Family::<Peer, Gauge>::default();
+        context.register(
+            "latest_vote",
+            "view of latest vote received per peer",
```
patrick-ogrady, line 91: This may be annoying to work with in practice in something like `reshare` because we namespace metrics by epoch (so you'd have to constantly be updating your dashboard I think?)

But that should be considered in a follow-up.

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -384,6 +400,12 @@ impl<
                         continue;
                     }
 
+                    // Update per-peer latest vote metric (only if higher than current)
+                    let _ = self
```
patrick-ogrady: Let's actually move beneath interesting?

patrick-ogrady: Prevents someone from sending us a view way in advance?

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -79,6 +85,16 @@ impl<
             "number of inbound messages",
             inbound_messages.clone(),
         );
+        let latest_vote = Family::<Peer, Gauge>::default();
+        context.register(
+            "latest_vote",
+            "view of latest vote received per peer",
+            latest_vote.clone(),
+        );
+        let participants = cfg.scheme.participants().clone();
```
andresilva: Do we actually need to clone here?

runtime/src/telemetry/metrics/status.rs
```diff
@@ -87,6 +88,10 @@ impl Drop for CounterGuard {
 pub trait GaugeExt {
     /// Sets the [`Gauge`] using a value convertible to `i64`, if conversion is not lossy, returning the previous value if successful.
     fn try_set<T: TryInto<i64>>(&self, val: T) -> Result<i64, T::Error>;
+
+    /// Atomically sets the [`Gauge`] to the maximum of the current value and the provided value.
+    /// Returns the previous value.
+    fn set_max<T: TryInto<i64> + Copy>(&self, val: T) -> Result<i64, T::Error>;
```
andresilva: `try_set_max`?

andresilva (approved):

consensus/src/simplex/actors/batcher/actor.rs
andresilva: It's still presumably their view though?

patrick-ogrady: Yeah, you're right. Will leave.

consensus/src/simplex/actors/batcher/actor.rs
patrick-ogrady: Moving the location so clearer to see why.

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -418,12 +412,19 @@ impl<
...
                         .or_insert_with(|| self.new_round())
                         .add_network(sender, message)
                         .await {
                             self.added.inc();
+
+                            // Update per-peer latest vote metric (only if higher than current)
+                            let _ = self
+                                .latest_vote
+                                .get_or_create(&peer)
+                                .try_set_max(view.get());
```
patrick-ogrady, line 427: Opted to move into post-network validation to avoid duplicating its checks (to ensure a peer, for example is in participants -> thank you based bugbot).

patrick-ogrady: This shouldn't happen with our current p2p but we should be defensive.

PR #2711, opened 2026-01-06
refactor(codec): deduplicate read_ordered_map function

reallesee:
Moved identical read_ordered_map function from btree_map.rs and hash_map.rs to types/mod.rs


codec/src/types/btree_map.rs
```diff
@@ -5,63 +5,17 @@
 
 extern crate alloc;
 
+use super::read_ordered_map;
```
danlaine: @reallesee Actually, one nit: would you mind importing through `crate::types` instead of `super`? Same in `hash_map.rs`

reallesee: Done!

danlaine:
@reallesee Running `cargo +nightly fmt --all -- ` should fix the linter errors

PR #2712, opened 2026-01-06
[consensus/ordered-broadcast] Remove `epoch` from signed over data

### LINKED ISSUE #969: [consensus] Ordered Broadcast: Remove `epoch` from signed-over information
While `epoch` should be included in the network message (to help peers know what epoch the signature applies to), we don't actually need to sign over this information. This creates "issues" down the line since we need to know the epoch number for every threshold signature (which participants generally don't care about) to verify the threshold signature

BrendanChou:
Removes `epoch` from the signed data in ordered broadcast ack/certificate signatures. This allows certificates to be verified with just the chunk and scheme, without needing to know which epoch produced them.

`epoch` remains in the wire format for routing but is no longer part of the signed message. Replay resistance is maintained by domain-separated namespaces, the chunk binding `(sequencer, height, payload)`, and protocol-level sender/epoch validation.

Existing sign/verify tests are sufficient since verification has always been determined by the scheme (validator set + keys), not the epoch. The epoch was just a lookup key to find the right scheme. Removing it from the signed data doesn't change any verification path that tests exercise.

Fixes #969

### DIFF
--- a/consensus/src/ordered_broadcast/types.rs
+++ b/consensus/src/ordered_broadcast/types.rs
@@ -369,13 +369,11 @@ impl Namespace for AckNamespace {
 /// This is used as the context type for `Scheme` implementations for validators.
 /// It contains the chunk being acknowledged.
 ///
-/// Note: epoch is NOT included here because the p2p channel is already muxed by epoch,
-/// so including it in the signature is redundant. Not signing over epoch allows
-/// certificates to be verified without knowing the epoch.
-///
-/// Replay resistance still holds for the signed statement: signatures are domain
-/// separated with the ack namespace and bind the chunk `(sequencer, height, payload)`.
-/// The protocol also validates sender identity and epoch routing before accepting an ack.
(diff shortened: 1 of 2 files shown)


consensus/src/ordered_broadcast/types.rs
```diff
@@ -367,13 +367,15 @@ impl Namespace for AckNamespace {
 /// Context for signing/verifying validator acknowledgments.
 ///
 /// This is used as the context type for `Scheme` implementations for validators.
-/// It contains the chunk being acknowledged and the epoch of the validator set.
+/// It contains the chunk being acknowledged.
+///
+/// Note: epoch is NOT included here because the p2p channel is already muxed by epoch,
+/// so including it in the signature is redundant. Not signing over epoch allows
+/// certificates to be verified without knowing the epoch.
```
andresilva: Is it worth mentioning that the statements being signed aren't prone to replayability?

clabby:

PR #2713, opened 2026-01-06
[consensus/simplex] Cancel Outstanding Certifications

patrick-ogrady:


consensus/src/simplex/actors/voter/state.rs
```diff
@@ -447,6 +463,19 @@ impl<E: Clock + CryptoRngCore + Metrics, S: Scheme<D>, L: ElectorConfig<S>, D: D
...
+
+    /// Adds a view to the set of certification candidates if it's above last finalized.
+    pub fn add_certification_candidate(&mut self, view: View) {
+        if view > self.last_finalized {
+            self.certification_candidates.insert(view);
+        }
+    }
+
+    /// Takes all certification candidates, returning them and clearing the set.
+    pub fn take_certification_candidates(&mut self) -> BTreeSet<View> {
```
patrick-ogrady: nit: return an iterator here?

consensus/src/simplex/actors/voter/mod.rs
```diff
@@ -2642,4 +2642,202 @@ mod tests {
...
+                    MailboxMessage::Certificate(_) => continue,
+                    MailboxMessage::Certified { .. } => {
+                        panic!("unexpected Certified message before finalization processed")
+                    }
+                }
+            }
+
+            // Check that no Certified message was sent within a timeout
+            let certified_received = select! {
+                msg = resolver_receiver.next() => {
```
patrick-ogrady, line 2812: We should not check resolver_received but instead should just check the channel is cancelled in the automaton?

PR #2714, opened 2026-01-06
[cryptography] Remove aggregate::verify_same_message

cronokirby:
This method is, unfortunately, not secure as is, so just removing this until we can create a better alternative.


cronokirby:
(the wrong one, woops)

PR #2715, opened 2026-01-06
[cryptography] Remove aggregate verify_same_signer

cronokirby:
Related: #2624 

Remove the insecure verify_same_signer method while waiting for a better alternative.

### DIFF
--- a/cryptography/src/bls12381/primitives/ops/aggregate.rs
+++ b/cryptography/src/bls12381/primitives/ops/aggregate.rs
@@ -17,8 +17,6 @@ use alloc::vec::Vec;
 use bytes::{Buf, BufMut};
 use commonware_codec::{Error as CodecError, FixedSize, Read, ReadExt, Write};
 use commonware_math::algebra::Additive;
-#[cfg(feature = "std")]
-use rayon::{prelude::*, ThreadPoolBuilder};
 
 /// An aggregated public key from multiple individual public keys.
 ///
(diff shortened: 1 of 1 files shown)


cryptography/src/bls12381/primitives/ops/aggregate.rs
```diff
@@ -305,25 +300,6 @@ pub fn verify_same_message<V: Variant>(
...
-/// Instead of requiring all messages that participated in the aggregate signature (and generating
-/// the combined message on-demand), this function accepts a precomputed combined message to allow
-/// the caller to cache previous constructions and/or perform parallel combination.
-///
-/// # Warning
-///
-/// This function assumes a group check was already performed on `public` and `signature`.
-pub fn verify_same_signer<V: Variant>(
-    public: &V::Public,
-    message: &Message<V>,
```
patrick-ogrady, line 321: I think we can remove the `Message` type?

cryptography/src/bls12381/primitives/ops/aggregate.rs
```diff
@@ -463,88 +327,13 @@ mod tests {
...
-    }
-
     #[cfg(feature = "arbitrary")]
     mod conformance {
         use super::*;
         use commonware_codec::conformance::CodecConformance;
 
         commonware_conformance::conformance_tests! {
             CodecConformance<PublicKey<MinSig>>,
-            CodecConformance<Message<MinSig>>,
```
patrick-ogrady, line 547: I think you need to manually remove from `conformance`?

cc @clabby does the conformance check that there are no dangling refs?

clabby: It does not remove them (nor check.) #2717

patrick-ogrady: Will just allow as-is (and we can cleanup in 2717)

PR #2726, opened 2026-01-07
[p2p] Add Restart Tests + Introduce `FallibleExt` to Cleanup Channels

patrick-ogrady:
Future Work: https://github.com/commonwarexyz/monorepo/issues/2735

## Summary

- Add restart tests for discovery and lookup modules (deterministic + tokio)
- Introduce `FallibleExt` and `AsyncFallibleExt` traits for channel error handling
- Have `Mailbox`/`UnboundedMailbox` implement these traits
- Refactor simulated network to use `UnboundedMailbox`
- Gate `count_running_tasks` behind `#[cfg(any(test, feature = "test-utils"))]`

### DIFF
--- a/p2p/src/authenticated/discovery/mod.rs
+++ b/p2p/src/authenticated/discovery/mod.rs
@@ -1635,113 +1635,43 @@ mod tests {
         }
     }
 
-    async fn run_peer_restart_test(
-        context: impl Spawner + Clock + CryptoRngCore + RNetwork + Resolver + Metrics,
-        base_port: u16,
-        n: usize,
-    ) {
-        // Create peers
-        let peers: Vec<_> = (0..n)
-            .map(|i| ed25519::PrivateKey::from_seed(i as u64))
-            .collect();
-        let addresses: Vec<_> = peers.iter().map(|p| p.public_key()).collect();
+    #[test_traced]
+    fn test_many_peer_restart_with_new_address() {
+        let base_port = 7500;
+        let n = 5;
 
-        // Track senders/receivers/handles across restarts
-        let mut senders: Vec<Option<channels::Sender<_, _>>> = (0..n).map(|_| None).collect();
-        let mut receivers: Vec<Option<channels::Receiver<_>>> = (0..n).map(|_| None).collect();
-        let mut handles: Vec<Option<Handle<()>>> = (0..n).map(|_| None).collect();
+        let executor = deterministic::Runner::default();
+        executor.start(|context| async move {
+            // Create peers
+            let peers: Vec<_> = (0..n)
+                .map(|i| ed25519::PrivateKey::from_seed(i as u64))
+                .collect();
+            let addresses: Vec<_> = peers.iter().map(|p| p.public_key()).collect();
 
-        // Track port allocations (updated on restart)
-        let mut ports: Vec<u16> = (0..n).map(|i| base_port + i as u16).collect();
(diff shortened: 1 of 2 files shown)


patrick-ogrady:
We may be able to greatly reduce size of this PR by just taking our existing test and running for both runtimes.

p2p/src/authenticated/discovery/actors/tracker/ingress.rs
patrick-ogrady, line 1: Because we rely on `context` for shutdown (not channels closing), this simplifies the handling in each actor quite a bit (alternative would be to return send errors everywhere).

runtime/src/utils/mod.rs
```diff
@@ -194,6 +194,61 @@ impl ArcWake for Blocker {
...
+///     // Abort the task
+///     handle.abort();
+///     let _ = handle.await;
+///     context.sleep(Duration::from_millis(10)).await;
+///
+///     // Verify task is stopped
+///     let count = count_running_tasks(&context, "worker");
+///     assert_eq!(count, 0, "worker task should be stopped");
+/// });
+/// ```
```
clabby, line 236: `#[cfg(test)]`?

utils/src/channels/fallible.rs
clabby, line 1: This is nice 👍

p2p/src/authenticated/mailbox.rs
```diff
@@ -30,6 +34,36 @@ impl<T> Mailbox<T> {
...
+    }
+
+    async fn request_or_default<R, F>(&mut self, make_msg: F) -> R
+    where
+        R: Default + Send,
+        F: FnOnce(oneshot::Sender<R>) -> T + Send,
+    {
+        self.0.request_or_default(make_msg).await
+    }
+}
```
clabby: Should we expose these directly on `Mailbox` / `UnboundedMailbox`? We expose wrapper functions for all message types, this is a bit leaky.

p2p/src/authenticated/mailbox.rs
clabby: Same, doesn't feel like this should be exposed.

patrick-ogrady: Removed this

runtime/src/utils/mod.rs
patrick-ogrady: Add test-utils feature

p2p/src/authenticated/mailbox.rs
patrick-ogrady: Unless I'm missing something, I prefer to implement the trait like we do (and have people call it)?

I think we could just not impl the trait and have people use `self.0.xxxx` everywhere but feel like what we have is ok?

Maybe a skill issue on my part (didn't like defining them concretely but maybe just a style thing...)

p2p/src/authenticated/mailbox.rs
```diff
@@ -18,21 +18,9 @@ impl<T> Clone for Mailbox<T> {
...
-    /// Sends a message to the corresponding receiver.
-    pub async fn send(&mut self, message: T) -> Result<(), mpsc::SendError> {
-        self.0.send(message).await
-    }
-
-    /// Returns true if the mailbox is closed.
-    pub fn is_closed(&self) -> bool {
-        self.0.is_closed()
-    }
-}
```
clabby, line 31: 🫶

PR #2733, opened 2026-01-07
[storage/bmt] Add multi-proof support

### LINKED ISSUE #2221: [storage/bmt] Support Multi-Proofs
Like [mmr](https://github.com/commonwarexyz/monorepo/blob/8a1f6040d067f83700c26242fa1f5e75040ce23f/storage/src/adb/verify.rs#L124-L153), support multi-proofs in `bmt`.

This would allow for a slightly more efficient approach than #2215

### LINKED ISSUE #2506: [coding/zoda] Fix Encoding Logic
After changing the namespace to `_COMMONWARE_CODING_ZODA`, `reshard_roundtrip_handles_field_packing` failed with:
```
    thread 'zoda::tests::reshard_roundtrip_handles_field_packing' (11092) panicked at coding/src/zoda.rs:843:10:
    called `Result::unwrap()` on an `Err` value: InvalidLength(11)
```

After some poking around, it became clear that the bounds used for decoding were too tight (limiting the MMR proof digests to just the number of leaves): https://github.com/commonwarexyz/monorepo/blob/a2029be73d99c0cb17a0586fca1dd1102f1c074e/coding/src/zoda.rs#L441

The test previously passed because the old namespace led to a different proof size (9) which didn't have any intermediate nodes (whereas the one that failed in `main` did): https://github.com/commonwarexyz/monorepo/pull/2501

We should add more comprehensive tests for encoding/decoding that better explore the encoding space to ensure this doesn't happen again.

_The LLMs seem to think the checksum matrix decoding rules are also too strict but we can assess that when writing tests._

clabby:
## Overview

Adds multi-proof support to `storage::bmt`.

This PR also breaks the API of the MMR `Proof`'s `Read` impl - the `Read::Cfg` is now the max number of _items_, rather than the max number of _digests_ in the proof. The upper bound is inferred.

closes #2221 
closes #2506

### `zoda::encode`

mmr -> bmt, large change in perf.

```
zoda::encode/msg_len=1048576 chunks=50 conc=1
                        time:   [34.751 ms 35.043 ms 35.257 ms]
                        change: [−14.404% −13.989% −13.572%] (p = 0.00 < 0.05)
                        Performance has improved.

zoda::encode/msg_len=1048576 chunks=50 conc=4
                        time:   [21.227 ms 21.269 ms 21.304 ms]
                        change: [−30.137% −29.321% −28.654%] (p = 0.00 < 0.05)
                        Performance has improved.

zoda::encode/msg_len=1048576 chunks=50 conc=8
                        time:   [19.397 ms 19.447 ms 19.499 ms]
                        change: [−31.636% −31.395% −31.165%] (p = 0.00 < 0.05)
                        Performance has improved.
```

### `reed_solomon::encode`

Still using `bmt`, now forced to use the multi-proof path. No significant change in perf.

```
reed_solomon::encode/msg_len=1048576 chunks=50 conc=1
                        time:   [6.2472 ms 6.2612 ms 6.2816 ms]
                        change: [+0.0229% +0.5795% +1.0971%] (p = 0.07 > 0.05)
                        No change in performance detected.

reed_solomon::encode/msg_len=1048576 chunks=50 conc=4
                        time:   [2.0799 ms 2.0854 ms 2.0896 ms]
                        change: [+0.6271% +1.0928% +1.5377%] (p = 0.00 < 0.05)
                        Change within noise threshold.

reed_solomon::encode/msg_len=1048576 chunks=50 conc=8
                        time:   [1.4275 ms 1.4370 ms 1.4451 ms]
                        change: [+0.1988% +0.9165% +1.6285%] (p = 0.03 < 0.05)
                        Change within noise threshold.
```

### DIFF
--- a/storage/fuzz/fuzz_targets/proofs_malleability.rs
+++ b/storage/fuzz/fuzz_targets/proofs_malleability.rs
@@ -182,11 +182,7 @@ fn fuzz(input: FuzzInput) {
 
                     for mutation in &input.mutations {
                         let mut mutated_proof = original_proof.clone();
-                        mutate_proof_bytes(
-                            &mut mutated_proof,
-                            mutation,
-                            &commonware_storage::bmt::MAX_LEVELS,
-                        );
+                        mutate_proof_bytes(&mut mutated_proof, mutation, &1);
 
                         if mutated_proof != original_proof {
                             let is_valid = mutated_proof


storage/src/bmt/mod.rs
```diff
@@ -590,6 +629,308 @@ impl<H: Hasher> EncodeSize for RangeProof<H> {
     }
 }
 
+/// A Merkle proof for multiple non-contiguous leaves in a Binary Merkle Tree.
+///
+/// This proof type is more space-efficient than generating individual proofs
+/// for each leaf because sibling nodes that are shared between multiple paths
+/// are deduplicated.
+#[derive(Clone, Debug, Eq)]
+pub struct MultiProof<H: Hasher> {
```
patrick-ogrady: For the MMR, we reuse the same proof type for both single/multi-proof. 

The change required for that would be to enshrine `leaf_count` in `Proof`. While it adds a little more data,  we may be able to get around much of the overhead by using a varint?

I'm generally a fan of consistency between proof types but 🤷

clabby: Unified the proof types, though kept `RangeProof`.

storage/src/bmt/mod.rs
```diff
@@ -590,6 +510,249 @@ impl<H: Hasher> EncodeSize for RangeProof<H> {
...
+        self.leaf_count.write(writer);
+        self.siblings.write(writer);
+    }
+}
+
+impl<H: Hasher> Read for Proof<H> {
+    type Cfg = ();
+
+    fn read_cfg(reader: &mut impl Buf, _: &Self::Cfg) -> Result<Self, commonware_codec::Error> {
+        let leaf_count = u32::read(reader)?;
```
patrick-ogrady, line 543: We still need to bound this? It could be massive!

storage/src/bmt/mod.rs
```diff
@@ -590,6 +510,249 @@ impl<H: Hasher> EncodeSize for RangeProof<H> {
...
+    Ok(sibling_positions)
+}
+
+impl<H: Hasher> Proof<H> {
+    /// Verifies that the given `elements` at their respective positions are included
+    /// in a Binary Merkle Tree with `root`.
+    ///
+    /// Elements can be provided in any order; positions are sorted internally.
+    /// Duplicate positions will cause verification to fail.
+    pub fn verify(
```
patrick-ogrady: Wonder if we should have `verify` and `multi_verify`?

I suspect `verify` could be much cheaper than this?

storage/src/bmt/mod.rs
clabby: The vec is bounded, though we infer the range from the size of the buffer (or the max siblings, whichever is smaller.) This should still be safe, though we can switch over to a manually specified upper bound like MMR's `Proof` if you'd prefer.

storage/src/bmt/mod.rs
clabby: Went with mirroring the MMR:

https://github.com/commonwarexyz/monorepo/blob/5a999ff36e13f6a4d13bd0ba8953f05771354f10/storage/src/mmr/proof.rs#L127-L140

Glad to add if you'd like!

storage/src/bmt/mod.rs
patrick-ogrady: Ahhhh, I missed reader.remaining(). Nice.

patrick-ogrady: I think we could probably use this trick elsewhere...

storage/src/bmt/mod.rs
patrick-ogrady: Let's add it (and align with naming in `mmr`) I think?

although I kinda think that naming sucks lol

storage/src/bmt/mod.rs
patrick-ogrady: For consistency, for now let's require a u32 range?

storage/src/bmt/mod.rs
```diff
@@ -550,18 +550,12 @@ impl<H: Hasher> Write for Proof<H> {
 }
 
 impl<H: Hasher> Read for Proof<H> {
-    type Cfg = ();
+    /// The maximum number of siblings in the proof.
+    type Cfg = usize;
```
patrick-ogrady, line 537: I was thinking we could just do by `leaf`? It may be easier to follow?

Not sure what MMR does but suspect that'd be preferred there too.

storage/src/mmr/proof.rs
```diff
@@ -23,6 +23,13 @@ use core::ops::Range;
 #[cfg(feature = "std")]
 use tracing::debug;
 
+/// The maximum number of digests in a proof per element being proven.
+///
+/// This accounts for both the path from a leaf to its peak (max 62 nodes) and
+/// peak digests for mountains not containing the element (max 63 peaks).
+/// The value 128 provides a safe upper bound with room for future growth.
+pub const MAX_PROOF_DIGESTS_PER_ELEMENT: usize = 128;
```
patrick-ogrady: Can you add a test for something that does produce something right at the limit?

clabby: https://github.com/commonwarexyz/monorepo/pull/2733/changes/c4a137786d916054c8a5be955e2a6809c2cfaeb6

storage/src/mmr/proof.rs
```diff
@@ -1644,6 +1644,86 @@ mod tests {
...
+        //   - Path siblings from leaf to peak: h digests
+        //   - Other peaks (not containing the element): (62 - 1) = 61 digests
+        //   - Total: h + 61 digests
+        //
+        // Worst case: element in tallest tree (h = 61)
+        //   - Path siblings: 61
+        //   - Other peaks: 61
+        //   - Total: 61 + 61 = 122 digests
+        //
+        // This is comfortably below MAX_PROOF_DIGESTS_PER_ELEMENT = 128.
```
patrick-ogrady: Should we make this a tight bound?

patrick-ogrady: It would be great to also add some test that `const` really is the largest...maybe something like this? idk.

```
diff --git a/storage/src/mmr/proof.rs b/storage/src/mmr/proof.rs
index a4e73fca9..12ebde47e 100644
--- a/storage/src/mmr/proof.rs
+++ b/storage/src/mmr/proof.rs
@@ -25,10 +25,13 @@ use tracing::debug;
 
 /// The maximum number of digests in a proof per element being proven.
 ///
-/// This accounts for both the path from a leaf to its peak (max 62 nodes) and
-/// peak digests for mountains not containing the element (max 63 peaks).
-/// The value 128 provides a safe upper bound with room for future growth.
-pub const MAX_PROOF_DIGESTS_PER_ELEMENT: usize = 128;
+/// The worst case occurs in an MMR with maximum peaks (N = 2^62 - 1 leaves),
+/// which has 62 perfect binary trees with heights 61, 60, ..., 1, 0.
+/// For an element in the tallest tree:
+/// - Path siblings from leaf to peak: 61 digests
+/// - Other peaks (not containing the element): 61 digests
+/// - Total: 122 digests
+pub const MAX_PROOF_DIGESTS_PER_ELEMENT: usize = 122;
 
 /// Errors that can occur when reconstructing a digest from a proof due to invalid input.
 #[derive(Error, Debug)]
@@ -1647,80 +1650,135 @@ mod tests {
     #[test]
     fn test_max_proof_digests_per_element_sufficient() {
         // Verify that MAX_PROOF_DIGESTS_PER_ELEMENT (128) is sufficient for any single-element
-        // proof in the largest valid MMR.
+        // proof in any valid MMR.
         //
-        // MMR sizes follow: mmr_size(N) = 2*N - popcount(N) where N = leaf count.
-        // The number of peaks equals popcount(N).
+        // We test both candidate "largest" configurations and prove which is the worst case:
         //
-        // To maximize peaks, we want N with maximum popcount. N = 2^62 - 1 has 62 one-bits:
-        //   N = 0x3FFFFFFFFFFFFFFF = 2^0 + 2^1 + ... + 2^61
+        // 1. Single perfect tree (maximum MMR size):
+        //    - N = 2^62 leaves (a power of 2)
+        //    - MMR size = 2*N - popcount(N) = 2^63 - 1
+        //    - 1 peak of height 62
+        //    - Worst case: 62 path siblings + 0 other peaks = 62 digests
         //
-        // This gives us 62 perfect binary trees with leaf counts 2^0, 2^1, ..., 2^61
-        // and corresponding heights 0, 1, ..., 61.
+        // 2. Maximum peaks (maximum number of binary trees):
+        //    - N = 2^62 - 1 leaves (all bits set below bit 62)
+        //    - MMR size = 2*N - popcount(N) = 2^63 - 64
+        //    - 62 peaks with heights 61, 60, ..., 1, 0
+        //    - Worst case: 61 path siblings + 61 other peaks = 122 digests
         //
-        // mmr_size(2^62 - 1) = 2*(2^62 - 1) - 62 = 2^63 - 2 - 62 = 2^63 - 64
-        //
-        // For a single-element proof in a tree of height h:
-        //   - Path siblings from leaf to peak: h digests
-        //   - Other peaks (not containing the element): (62 - 1) = 61 digests
-        //   - Total: h + 61 digests
-        //
-        // Worst case: element in tallest tree (h = 61)
-        //   - Path siblings: 61
-        //   - Other peaks: 61
-        //   - Total: 61 + 61 = 122 digests
-        //
-        // This is comfortably below MAX_PROOF_DIGESTS_PER_ELEMENT = 128.
+        // Configuration 2 produces the worst case (122 > 62), which defines
+        // MAX_PROOF_DIGESTS_PER_ELEMENT.
+
+        // Helper to compute MMR size from leaf count: mmr_size(N) = 2*N - popcount(N)
+        let mmr_size_from_leaves = |n: u64| -> Position {
+            Position::new(2 * n - n.count_ones() as u64)
+        };
 
-        const NUM_PEAKS: usize = 62;
-        const MAX_TREE_HEIGHT: usize = 61;
-        const EXPECTED_WORST_CASE: usize = MAX_TREE_HEIGHT + (NUM_PEAKS - 1);
+        // Configuration 1: Single perfect tree with 2^62 leaves
+        let single_tree_leaves = 1u64 << 62;
+        let single_tree_size = mmr_size_from_leaves(single_tree_leaves);
+        assert!(single_tree_size.is_mmr_size());
+        assert_eq!(single_tree_leaves.count_ones(), 1);
 
-        let many_peaks_size = Position::new((1u64 << 63) - 64);
-        assert!(
-            many_peaks_size.is_mmr_size(),
-            "Size {many_peaks_size} should be a valid MMR size",
+        let single_tree_peaks: Vec<_> = PeakIterator::new(single_tree_size).collect();
+        assert_eq!(single_tree_peaks.len(), 1, "Single tree should have 1 peak");
+        let single_tree_height = single_tree_peaks[0].1;
+        assert_eq!(single_tree_height, 62, "Single tree peak should have height 62");
+
+        let single_tree_worst = nodes_required_for_range_proof(
+            single_tree_size,
+            Location::new_unchecked(0)..Location::new_unchecked(1),
+        )
+        .expect("should compute positions for single tree")
+        .len();
+        assert_eq!(
+            single_tree_worst,
+            single_tree_height as usize,
+            "Single tree worst case should equal tree height",
         );
 
-        let peak_count = PeakIterator::new(many_peaks_size).count();
-        assert_eq!(peak_count, NUM_PEAKS);
+        // Configuration 2: Maximum peaks with 2^62 - 1 leaves
+        let max_peaks_leaves = (1u64 << 62) - 1;
+        let max_peaks_size = mmr_size_from_leaves(max_peaks_leaves);
+        assert!(max_peaks_size.is_mmr_size());
 
-        // Verify the peak heights are 61, 60, ..., 1, 0 (from left to right)
-        let peaks: Vec<_> = PeakIterator::new(many_peaks_size).collect();
-        for (i, &(_pos, height)) in peaks.iter().enumerate() {
-            let expected_height = (NUM_PEAKS - 1 - i) as u32;
-            assert_eq!(
-                height, expected_height,
-                "Peak {i} should have height {expected_height}, got {height}",
-            );
+        let num_peaks = max_peaks_leaves.count_ones() as usize;
+        assert_eq!(num_peaks, 62, "Max peaks should have 62 peaks");
+
+        let max_peaks_peaks: Vec<_> = PeakIterator::new(max_peaks_size).collect();
+        assert_eq!(max_peaks_peaks.len(), num_peaks);
+
+        // Verify peak heights are 61, 60, ..., 1, 0 (decreasing)
+        for (i, &(_pos, height)) in max_peaks_peaks.iter().enumerate() {
+            let expected_height = (num_peaks - 1 - i) as u32;
+            assert_eq!(height, expected_height);
         }
 
-        // Test location 0 (leftmost leaf, in tallest tree of height 61)
-        // Expected: 61 path siblings + 61 other peaks = 122 digests
-        let loc = Location::new_unchecked(0);
-        let positions = nodes_required_for_range_proof(many_peaks_size, loc..loc + 1)
-            .expect("should compute positions for location 0");
+        let max_tree_height = max_peaks_peaks[0].1 as usize;
+        assert_eq!(max_tree_height, 61);
+
+        // Worst case is element in tallest tree (location 0)
+        let max_peaks_worst = nodes_required_for_range_proof(
+            max_peaks_size,
+            Location::new_unchecked(0)..Location::new_unchecked(1),
+        )
+        .expect("should compute positions for max peaks")
+        .len();
 
+        let expected_worst = max_tree_height + (num_peaks - 1);
         assert_eq!(
-            positions.len(),
-            EXPECTED_WORST_CASE,
-            "Location 0 proof should require exactly {EXPECTED_WORST_CASE} digests (61 path + 61 peaks)",
+            max_peaks_worst, expected_worst,
+            "Max peaks worst case should be {max_tree_height} path + {} other peaks",
+            num_peaks - 1,
         );
 
-        // Test the rightmost leaf (in smallest tree of height 0, which is itself a peak)
-        // Expected: 0 path siblings + 61 other peaks = 61 digests
-        let last_leaf_loc = (1u64 << 62) - 2; // Last leaf location
-        let positions = nodes_required_for_range_proof(
-            many_peaks_size,
+        // Prove max_peaks is the worst case
+        assert!(
+            max_peaks_worst > single_tree_worst,
+            "Max peaks ({max_peaks_worst}) should be worse than single tree ({single_tree_worst})",
+        );
+
+        // Verify the bound is sufficient
+        assert!(
+            max_peaks_worst <= MAX_PROOF_DIGESTS_PER_ELEMENT,
+            "Worst case ({max_peaks_worst}) should not exceed MAX_PROOF_DIGESTS_PER_ELEMENT ({MAX_PROOF_DIGESTS_PER_ELEMENT})",
+        );
+
+        // Test best case in max peaks: rightmost leaf (height 0 tree, no path siblings)
+        let last_leaf_loc = max_peaks_leaves - 1;
+        let best_case = nodes_required_for_range_proof(
+            max_peaks_size,
             Location::new_unchecked(last_leaf_loc)..Location::new_unchecked(last_leaf_loc + 1),
         )
-        .expect("should compute positions for last leaf");
-
-        let expected_last_leaf = NUM_PEAKS - 1;
+        .expect("should compute positions for last leaf")
+        .len();
         assert_eq!(
-            positions.len(),
-            expected_last_leaf,
-            "Last leaf proof should require exactly {expected_last_leaf} digests (0 path + 61 peaks)",
+            best_case,
+            num_peaks - 1,
+            "Best case should be 0 path + {} other peaks",
+            num_peaks - 1,
+        );
+    }
+
+    #[test]
+    fn test_proof_decode_rejects_excess_digests() {
+        // Verify that decoding rejects proofs with more digests than allowed by max_items.
+        let excess_digests = MAX_PROOF_DIGESTS_PER_ELEMENT + 1;
+
+        let proof = Proof {
+            size: Position::new(1), // Valid MMR size (single leaf)
+            digests: vec![test_digest(0); excess_digests],
+        };
+
+        let encoded = proof.encode();
+
+        // With max_items = 1, the maximum allowed digests is MAX_PROOF_DIGESTS_PER_ELEMENT
+        let result = Proof::<Digest>::decode_cfg(encoded, &1usize);
+        assert!(
+            result.is_err(),
+            "Decode should reject proof with {} digests when max_items=1 allows only {}",
+            excess_digests,
+            MAX_PROOF_DIGESTS_PER_ELEMENT,
         );
     }
 ```

File: storage/src/bmt/mod.rs [hunk 7]
```diff
@@ -590,10 +509,321 @@ impl<H: Hasher> EncodeSize for RangeProof<H> {
...
+    fn write(&self, writer: &mut impl BufMut) {
+        self.leaf_count.write(writer);
+        self.siblings.write(writer);
+    }
+}
+
+impl<H: Hasher> Read for Proof<H> {
+    /// The maximum number of items being proven.
+    ///
+    /// The upper bound on sibling hashes is derived as `max_items * MAX_LEVELS`.
```
patrick-ogrady, line 536: I think we're using `NonZeroRange` now? forgot

storage/src/mmr/proof.rs
```diff
@@ -23,6 +23,13 @@ use core::ops::Range;
 #[cfg(feature = "std")]
 use tracing::debug;
 
+/// The maximum number of digests in a proof per element being proven.
+///
+/// This accounts for both the path from a leaf to its peak (max 62 nodes) and
+/// peak digests for mountains not containing the element (max 63 peaks).
```
danlaine: I think the max height is `61` and the max number of peaks is `62` (like in the test)

```rs
        const NUM_PEAKS: usize = 62;
        const MAX_TREE_HEIGHT: usize = 61;
```

storage/src/bmt/mod.rs
```diff
@@ -590,10 +509,321 @@ impl<H: Hasher> EncodeSize for RangeProof<H> {
     }
 }
 
+/// A Merkle proof for multiple non-contiguous leaves in a Binary Merkle Tree.
+///
+/// This proof type is more space-efficient than generating individual proofs
+/// for each leaf because sibling nodes that are shared between multiple paths
+/// are deduplicated.
+#[derive(Clone, Debug, Eq)]
+pub struct Proof<H: Hasher> {
```
danlaine: Should we parameterize `Proof` on `D: Digest`? That's what `mmr::Proof` does.

storage/src/bmt/mod.rs
```diff
@@ -590,10 +509,321 @@ impl<H: Hasher> EncodeSize for RangeProof<H> {
...
+    /// Elements can be provided in any order; positions are sorted internally.
+    /// Duplicate positions will cause verification to fail.
+    pub fn verify_multi_inclusion(
+        &self,
+        hasher: &mut H,
+        elements: &[(H::Digest, u32)],
+        root: &H::Digest,
+    ) -> Result<(), Error> {
+        // Handle empty case
+        if elements.is_empty() {
```
danlaine, line 711: When `elements.is_empty()` we're not really verifying `that the given elements at their respective positions are included in a Binary Merkle Tree with root` since there aren't any elements. We're actually verifying that the proof is empty (i.e. we return `Ok` iff `self.leaf_count == 0 && self.siblings.is_empty() && root == [empty root]`).

Should we always return `Err(Error::NoLeaves)` when `elements.is_empty()` instead?

On the other hand, this is similar to how `mmr::Proof` behaves so there's an argument to be made we should keep this as is for now and change this behavior in both later.

storage/src/mmr/proof.rs
clabby: Whoops - thanks, originally had this wrong and didn't update docs.

storage/src/bmt/mod.rs
clabby: Sure

storage/src/bmt/mod.rs
clabby: Yeah; I figured we wanted to reject creating proofs against an empty tree re: your first point. I lean towards unifying the API between the bmt and mmr either way.

So long as we document the ability to create + verify a "proof of empty tree," I think it's fine to have that behavior? We can update the bmt to allow for it.

danlaine: > I lean towards unifying the API between the bmt and mmr either way.

Agreed. I'd say let's keep this as is for now, and we can change both (bmt/mmr) implementations later if we decide we prefer to error on `elements.is_empty()`

storage/src/mmr/proof.rs
clabby: Made it a tight bound

coding/src/reed_solomon.rs
```diff
@@ -47,13 +47,13 @@ pub struct Chunk<H: Hasher> {
     /// The index of [Chunk] in the original data.
     index: u16,
 
-    /// The proof of the shard in the [bmt] at the given index.
-    proof: bmt::Proof<H>,
+    /// The multi-proof of the shard in the [bmt] at the given index.
+    proof: bmt::Proof<H::Digest>,
```
danlaine, line 51: `Chunk` could also be parameterized on `D: Digest` and `verify` updated to take in a `&mut Hasher` like the proof `verify` methods. If we decide to do that, I think it's fine to defer to a future PR to keep this diff smaller.

storage/src/bmt/mod.rs
```diff
@@ -569,39 +479,341 @@ impl<H: Hasher> RangeProof<H> {
...
+        Ok(Self {
+            leaf_count: u.arbitrary()?,
+            siblings: u.arbitrary()?,
+        })
+    }
+}
+
+/// Returns the number of levels in a tree with `leaf_count` leaves.
+/// A tree with 1 leaf has 1 level, a tree with 2 leaves has 2 levels, etc.
+const fn levels_in_tree(leaf_count: u32) -> usize {
```
danlaine, line 574: It seems we always call this with `leaf_count` > 0 -- should we make the `0` case `unreachable!` instead of returning `1`? Not a big deal either way but I was a bit surprised that `levels_in_tree(0) == 1`.

storage/src/bmt/mod.rs
```diff
@@ -569,39 +479,341 @@ impl<H: Hasher> RangeProof<H> {
...
+            Err(Error::InvalidProof(computed.to_string(), root.to_string()))
+        }
+    }
+
+    /// Verifies that the given `elements` at their respective positions are included
+    /// in a Binary Merkle Tree with `root`.
+    ///
+    /// Elements can be provided in any order; positions are sorted internally.
+    /// Duplicate positions will cause verification to fail.
+    pub fn verify_multi_inclusion<H: Hasher<Digest = D>>(
```
danlaine, line 704: LLM suggested the change below, which appears to improve performance. Here are the results of running the `prove_multi` benchmark on the code in this PR after running it with the code in the diff below:

```
$ cargo bench -p commonware-storage --bench bmt -- prove_multi
   Compiling commonware-storage v0.0.64 (/Users/dan/Documents/monorepo/storage)
    Finished `bench` profile [optimized] target(s) in 3.85s
     Running src/bmt/benches/bench.rs (target/release/deps/bmt-52793cb6efec007d)
Gnuplot not found, using plotters backend
Benchmarking bmt::prove_multi/n=250 items=100
Benchmarking bmt::prove_multi/n=250 items=100: Warming up for 3.0000 s
Benchmarking bmt::prove_multi/n=250 items=100: Collecting 10 samples in estimated 5.0050 s (46k iterations)
Benchmarking bmt::prove_multi/n=250 items=100: Analyzing
bmt::prove_multi/n=250 items=100
                        time:   [93.256 µs 97.058 µs 100.45 µs]
                        change: [+51.205% +57.457% +63.985%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 2 outliers among 10 measurements (20.00%)
  2 (20.00%) low mild

Benchmarking bmt::prove_multi/n=1000 items=100
Benchmarking bmt::prove_multi/n=1000 items=100: Warming up for 3.0000 s
Benchmarking bmt::prove_multi/n=1000 items=100: Collecting 10 samples in estimated 5.0102 s (26k iterations)
Benchmarking bmt::prove_multi/n=1000 items=100: Analyzing
bmt::prove_multi/n=1000 items=100
                        time:   [166.41 µs 173.06 µs 179.81 µs]
                        change: [+63.616% +71.280% +78.760%] (p = 0.00 < 0.05)
                        Performance has regressed.

Benchmarking bmt::prove_multi/n=5000 items=100
Benchmarking bmt::prove_multi/n=5000 items=100: Warming up for 3.0000 s
Benchmarking bmt::prove_multi/n=5000 items=100: Collecting 10 samples in estimated 5.0127 s (16k iterations)
Benchmarking bmt::prove_multi/n=5000 items=100: Analyzing
bmt::prove_multi/n=5000 items=100
                        time:   [257.58 µs 267.55 µs 279.38 µs]
                        change: [+58.226% +66.222% +73.871%] (p = 0.00 < 0.05)
                        Performance has regressed.

Benchmarking bmt::prove_multi/n=10000 items=100
Benchmarking bmt::prove_multi/n=10000 items=100: Warming up for 3.0000 s
Benchmarking bmt::prove_multi/n=10000 items=100: Collecting 10 samples in estimated 5.0195 s (14k iterations)
Benchmarking bmt::prove_multi/n=10000 items=100: Analyzing
bmt::prove_multi/n=10000 items=100
                        time:   [298.95 µs 307.40 µs 317.10 µs]
                        change: [+59.382% +66.848% +74.621%] (p = 0.00 < 0.05)
                        Performance has regressed.

Benchmarking bmt::prove_multi/n=25000 items=100
Benchmarking bmt::prove_multi/n=25000 items=100: Warming up for 3.0000 s
Benchmarking bmt::prove_multi/n=25000 items=100: Collecting 10 samples in estimated 5.0040 s (12k iterations)
Benchmarking bmt::prove_multi/n=25000 items=100: Analyzing
bmt::prove_multi/n=25000 items=100
                        time:   [350.42 µs 360.85 µs 372.71 µs]
                        change: [+63.728% +72.344% +81.366%] (p = 0.00 < 0.05)
                        Performance has regressed.

Benchmarking bmt::prove_multi/n=50000 items=100
Benchmarking bmt::prove_multi/n=50000 items=100: Warming up for 3.0000 s
Benchmarking bmt::prove_multi/n=50000 items=100: Collecting 10 samples in estimated 5.0105 s (10k iterations)
Benchmarking bmt::prove_multi/n=50000 items=100: Analyzing
bmt::prove_multi/n=50000 items=100
                        time:   [382.94 µs 400.11 µs 418.75 µs]
                        change: [+62.117% +69.815% +77.613%] (p = 0.00 < 0.05)
                        Performance has regressed.

Benchmarking bmt::prove_multi/n=100000 items=100
Benchmarking bmt::prove_multi/n=100000 items=100: Warming up for 3.0000 s
Benchmarking bmt::prove_multi/n=100000 items=100: Collecting 10 samples in estimated 5.0098 s (9570 iterations)
Benchmarking bmt::prove_multi/n=100000 items=100: Analyzing
bmt::prove_multi/n=100000 items=100
                        time:   [426.25 µs 445.54 µs 461.60 µs]
                        change: [+66.048% +72.380% +78.885%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 2 outliers among 10 measurements (20.00%)
  1 (10.00%) low severe
  1 (10.00%) low mild
```

Here is the diff. I'm still working through to make sure I understand it but the unit tests pass with this change:

```
diff --git a/storage/src/bmt/mod.rs b/storage/src/bmt/mod.rs
index 947ad075b..724b39a35 100644
--- a/storage/src/bmt/mod.rs
+++ b/storage/src/bmt/mod.rs
@@ -40,7 +40,7 @@
 //! assert!(proof.verify_element_inclusion(&mut hasher, &digests[1], 1, &root).is_ok());
 //! ```
 
-use alloc::collections::{btree_map::BTreeMap, btree_set::BTreeSet};
+use alloc::collections::btree_set::BTreeSet;
 use bytes::{Buf, BufMut};
 use commonware_codec::{EncodeSize, Read, ReadExt, ReadRangeExt, Write};
 use commonware_cryptography::{Digest, Hasher};
@@ -726,75 +726,92 @@ impl<D: Digest> Proof<D> {
             return Err(Error::NoLeaves);
         }
 
-        // Get required sibling positions (validates positions and checks for duplicates)
-        let sibling_positions = siblings_required_for_multi_proof(
-            self.leaf_count,
-            elements.iter().map(|(_, pos)| *pos),
-        )?;
-
-        // Verify we have the exact number of siblings needed
-        if sibling_positions.len() != self.siblings.len() {
-            return Err(Error::UnalignedProof);
-        }
-
-        // Group siblings by level for efficient per-level access.
-        let levels_count = levels_in_tree(self.leaf_count);
-        let mut siblings_by_level: Vec<BTreeMap<usize, H::Digest>> =
-            (0..levels_count).map(|_| BTreeMap::new()).collect();
-        for (&(level, index), digest) in sibling_positions.iter().zip(self.siblings.iter()) {
-            siblings_by_level[level].insert(index, *digest);
-        }
-
-        // Compute position-hashed leaf digests to build initial level.
-        // Position bounds and duplicates were already validated by siblings_required_for_multi_proof.
-        let mut current: BTreeMap<usize, H::Digest> = BTreeMap::new();
+        // 1. Initialize current level nodes (sorted by position)
+        let mut current: Vec<(u32, D)> = Vec::with_capacity(elements.len());
         for (leaf, position) in elements {
+            if *position >= self.leaf_count {
+                return Err(Error::InvalidPosition(*position));
+            }
             hasher.update(&position.to_be_bytes());
             hasher.update(leaf);
-            current.insert(*position as usize, hasher.finalize());
+            current.push((*position, hasher.finalize()));
         }
 
-        // Compute tree level by level.
-        // IMPORTANT: We only check parents of nodes we have, not all possible parents.
-        // This ensures O(elements * levels) complexity instead of O(leaf_count * levels),
-        // preventing DoS attacks with maliciously large leaf_count values.
-        let mut level_size = self.leaf_count as usize;
-        for siblings in siblings_by_level.iter().take(levels_count - 1) {
-            // Add siblings for this level
-            current.extend(siblings.iter().map(|(&k, &v)| (k, v)));
-
-            // Compute unique parent indices from current level nodes
-            let candidate_parents: BTreeSet<usize> = current.keys().map(|idx| idx / 2).collect();
-
-            // Compute parent digests
-            let mut next: BTreeMap<usize, H::Digest> = BTreeMap::new();
-            for parent_idx in candidate_parents {
-                let left_child_idx = parent_idx * 2;
-                let right_child_idx = parent_idx * 2 + 1;
-
-                let left_digest = current.get(&left_child_idx);
-                let right_digest = if right_child_idx < level_size {
-                    current.get(&right_child_idx)
+        // Sort by position to enable linear processing
+        current.sort_unstable_by(|(p1, _), (p2, _)| p1.cmp(p2));
+
+        // Check for duplicates
+        for i in 1..current.len() {
+            if current[i].0 == current[i - 1].0 {
+                return Err(Error::DuplicatePosition(current[i].0));
+            }
+        }
+
+        // 2. Iterate up the tree
+        let levels = levels_in_tree(self.leaf_count);
+        let mut level_size = self.leaf_count;
+        let mut sibling_iter = self.siblings.iter();
+        let mut next_level = Vec::with_capacity(current.len() / 2 + 1);
+
+        for _ in 0..levels - 1 {
+            let mut i = 0;
+            while i < current.len() {
+                let (pos, digest) = current[i];
+                let parent_pos = pos / 2;
+
+                // Determine if we have the left or right child
+                let (left, right) = if pos % 2 == 0 {
+                    // We are the LEFT child
+                    let left = digest;
+
+                    // Check if we have the right child in our current set
+                    let right = if i + 1 < current.len() && current[i + 1].0 == pos + 1 {
+                        i += 2; // Consume both
+                        current[i - 1].1
+                    } else {
+                        i += 1; // Consume left
+
+                        // If no right child exists in tree, duplicate left
+                        if pos + 1 >= level_size {
+                            left
+                        } else {
+                            // Otherwise, must consume a sibling
+                            *sibling_iter.next().ok_or(Error::UnalignedProof)?
+                        }
+                    };
+                    (left, right)
                 } else {
-                    // Odd-sized level: duplicate left child
-                    left_digest
+                    // We are the RIGHT child
+                    // This implies the LEFT child was missing from 'current', so it must be a sibling.
+                    i += 1;
+                    let right = digest;
+                    let left = *sibling_iter.next().ok_or(Error::UnalignedProof)?;
+                    (left, right)
                 };
 
-                if let (Some(left), Some(right)) = (left_digest, right_digest) {
-                    hasher.update(left);
-                    hasher.update(right);
-                    next.insert(parent_idx, hasher.finalize());
-                }
+                // Hash parent
+                hasher.update(&left);
+                hasher.update(&right);
+                next_level.push((parent_pos, hasher.finalize()));
             }
 
-            current = next;
+            // Prepare for next level
+            current.clear();
+            current.append(&mut next_level);
             level_size = level_size.div_ceil(2);
         }
 
-        // Get the computed root (should be at index 0)
-        let computed = current.get(&0).ok_or(Error::UnalignedProof)?;
+        // 3. Verify root
+        if sibling_iter.next().is_some() {
+            return Err(Error::UnalignedProof);
+        }
+
+        if current.len() != 1 {
+            return Err(Error::UnalignedProof);
+        }
 
-        if computed == root {
+        let computed = current[0].1;
+        if computed == *root {
             Ok(())
         } else {
             Err(Error::InvalidProof(computed.to_string(), root.to_string()))
```

storage/src/bmt/mod.rs
clabby: Rationale for `levels_in_tree(0) == 1` is that a tree with 0 leaves does still have a root. I think we should probably keep all `leaf_count`s valid inputs for that reason

coding/src/reed_solomon.rs
clabby: Let's do that in a follow-up - https://github.com/commonwarexyz/monorepo/issues/2753

PR #2734, opened 2026-01-07
[utils/consensus] marshal buffer implements new `Subscribable` trait

BrendanChou:


broadcast/src/buffered/ingress.rs
```diff
@@ -47,75 +46,34 @@ impl<P: PublicKey, M: Committable + Digestible + Codec> Mailbox<P, M> {
...
-                commitment,
-                digest,
+            .send(Message::Get {
+                commitment: key,
                 responder,
             })
-            .await;
+            .await
+            .expect("mailbox closed");
+        receiver.await.expect("mailbox closed")
```
patrick-ogrady, line 64: Please use FallibleExt!

utils/src/subscribable.rs
```diff
@@ -0,0 +1,29 @@
+//! A trait for subscribing to values by key.
+
+use crate::Span;
+use futures::channel::oneshot;
+use std::future::Future;
```
clabby, line 5: This is fine, but we can expose this in `no_std` with `core::future::Future`. `oneshot` should also be supported by `futures` w/ `std` disabled.

BrendanChou:
Closing as obsolete: `marshal/actor.rs` was deleted in the marshal restructure for erasure coding (#3164), and the subscription semantics this enabled landed differently via the subscriber/resolver design in #3796 (`subscribe`/`subscribe_prepared` in `buffered/ingress.rs`). Reviving this would be a from-scratch redesign rather than a rebase.

Issue #2735, opened 2026-01-08
[utils] Migrate `consensus` to `FallibleExt`


patrick-ogrady:
We should migrate `consensus` to `FallibleExt`.

PR #2738, opened 2026-01-08
[consensus/broadcast/collector/resovler] Adopt `FallibleExt` + Shutdown Tests

patrick-ogrady:
Closes: #2735

### DIFF
--- a/consensus/src/aggregation/mod.rs
+++ b/consensus/src/aggregation/mod.rs
@@ -103,7 +103,9 @@ mod tests {
         deterministic::{self, Context},
         Clock, Metrics, Quota, Runner, Spawner,
     };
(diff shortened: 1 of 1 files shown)


p2p/src/simulated/mod.rs
```diff
@@ -523,13 +523,11 @@ mod tests {
                 Err(Error::NetworkClosed)
             ));
 
-            // Send on original
-            assert!(matches!(
-                my_sender
-                    .send(Recipients::One(other_pk.clone()), msg.clone(), false)
-                    .await,
-                Err(Error::NetworkClosed)
```
patrick-ogrady, line 531: This was (another) subtle difference with `p2p::authenticated`

PR #2756, opened 2026-01-08
[consensus/simplex] Fix `Seedable` impl for `bls12381_threshold` scheme

clabby:
## Overview

Fixes the `Seedable` implementation for `bls12381_threshold::Scheme` in simplex. API blunder from https://github.com/commonwarexyz/monorepo/pull/2621


consensus/src/simplex/scheme/bls12381_threshold.rs
```diff
@@ -75,7 +75,7 @@ enum Role<P: PublicKey, V: Variant> {
 /// The scheme is generic over a [`Strategy`] which determines whether cryptographic
 /// operations such as signature recovery and batch verification run sequentially or in parallel.
 #[derive(Clone, Debug)]
-pub struct Scheme<P: PublicKey, V: Variant, S: Strategy = Sequential> {
+pub struct Scheme<P: PublicKey, V: Variant, S: Strategy> {
```
clabby, line 78: Opted to remove the default strategy since it caused the silent API blunder.

PR #2764, opened 2026-01-09
Small scalar (and other?) optimizations for BLS

cronokirby:


cryptography/src/bls12381/primitives/group.rs
```diff
@@ -496,6 +500,26 @@ impl Random for Scalar {
     }
 }
 
+const SMALL_SCALAR_BITS: usize = SCALAR_BITS / 2;
+const SMALL_SCALAR_BYTES: usize = (SMALL_SCALAR_BITS + 7) / 8;
```
patrick-ogrady, line 504: nit: should we just hardcode?

PR #2767, opened 2026-01-09
[resolver] Remove empty target sets in block()

syrupipy:
Fixes memory leak where block() removed peers from target sets but left empty HashSet entries around. Empty sets have different semantics than missing entries (blocks fallback permanently vs allows any peer), so this also fixes stuck fetches when all targets get blocked.

Changed to use retain() which removes the peer and cleans up empty entries in one pass.

### DIFF
--- a/resolver/src/p2p/fetcher.rs
+++ b/resolver/src/p2p/fetcher.rs
@@ -501,12 +501,14 @@ where
 
     /// Blocks a peer from being used to fetch data.
     ///
-    /// Also removes the peer from all target sets.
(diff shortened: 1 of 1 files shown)


andresilva (changes requested):
The existing code is intentional, if you remove the empty set you implicitly relax a targeted fetch to an untargeted fetch (i.e. try all peers). Going from targeted to untargeted needs to be done explicitly (see the docs in the `Resolver` trait). It's fine that the empty set lingers there, either new targets get added and eventually the fetch gets resolved, or eventually the user cancels the fetch (after some timeout) which will clear this.

PR #2768, opened 2026-01-09
[deployer] Add x86_64 architecture support

commonware-llm:
Add support for deploying on x86_64 instances in addition to ARM64:
- Add Architecture enum (arm64/x86_64) with default to arm64
- Add architecture field to InstanceConfig and MonitoringConfig
- Update find_latest_ami() to select AMI based on architecture
- Rename assert_arm64_support() to assert_architecture_support()
- Update all download URLs to be architecture-aware
- Generate binary.service per-instance for correct jemalloc path


patrick-ogrady:
Redoing this PR following s3 changes.

PR #2769, opened 2026-01-09
[Storage] Fix `journal::contiguous::Variable::init_at_size`

danlaine:
**Problem Summary**

When `contiguous::variable::Journal::init_at_size(size)` is called:

* It initializes an empty data journal
* It initializes the offsets journal at size using `init_journal_at_size` (which writes dummy values)
Result: data has 0 items, offsets has `size` items

On recovery via `init()`:

* `align_journals` counts items in data journal: 0
* Sees offsets has `size` items
* `offsets_size > data_size` triggers rewind of offsets to 0
Bug:  We rewind data which is actually valid because we misinterpret data written by `init_at_size` as being caused by a crash

**Fix**

Write 1 byte dummy values to data journal so that `init()` crash recovery sees that the data and offsets journals are aligned after `init_at_size` and don't need to be rewound.

### DIFF
--- a/storage/src/journal/segmented/variable.rs
+++ b/storage/src/journal/segmented/variable.rs
@@ -54,15 +54,16 @@
 //! # Dummy Items
 //!
 //! The journal supports "dummy" items -- size-0 entries consisting of only a `varint(0)` header
-//! (a single `0x00` byte) with no data payload -- for crate internal use. These are used by the
-//! [crate::journal::contiguous::variable::Journal] to efficiently initialize a journal at a
-//! whose earliest section _logically_ contains more items than it _physically_ does.
+//! (a single `0x00` byte) with no data payload -- for crate-internal use.
 //!
-//! For example, a section that begins with 2 dummy elements followed by 3 "regular" elements
-//! would report (via `scan_section_offsets`) having 5 items, but the section would only contain
-//! 3 "regular" elements.
+//! Dummies enable [crate::journal::contiguous::variable::Journal] to efficiently initialize at
+//! a non-zero logical position. For example, to create a journal starting at logical position
+//! 1,000,007, instead of writing 1,000,007 entries, only the tail section needs entries. If
+//! `items_per_section` is 1000, position 1,000,007 falls in section 1000 at offset 7. So we
+//! write 7 dummy items to the tail section, making the next append position 1,000,007.
 //!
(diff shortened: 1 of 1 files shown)


storage/src/journal/contiguous/fixed.rs
```diff
@@ -287,6 +288,138 @@ impl<E: Storage + Metrics, A: CodecFixedShared> Journal<E, A> {
...
+    /// - Creates only the tail blob at the index that would contain the item at `size`
+    /// - Sets the tail blob size to represent the "leftover" items within that blob
+    /// - No blobs are created for pruned indices
+    ///
+    /// For example, if `items_per_blob = 10` and `size = 25`:
+    /// - Tail blob index would be 25 / 10 = 2 (third blob, 0-indexed)
+    /// - Tail blob size would be (25 % 10) * CHUNK_SIZE = 5 * CHUNK_SIZE
+    /// - No blobs are created for indices 0 and 1 (the pruned range)
+    /// - Reading from positions 0-19 will return `ItemPruned`
+    pub(crate) async fn init_at_size(context: E, cfg: Config, size: u64) -> Result<Self, Error> {
```
danlaine, line 309: Moved from `storage/src/qmdb/any/unordered/fixed/sync.rs`

storage/src/journal/contiguous/fixed.rs
```diff
@@ -287,6 +288,138 @@ impl<E: Storage + Metrics, A: CodecFixedShared> Journal<E, A> {
...
+    ///
+    /// - Fresh (no data): returns journal initialized at `range.start`
+    /// - Stale (all data before `range.start`): destroys and recreates at `range.start`
+    /// - Overlap within range: prunes to `range.start`
+    /// - Unexpected data beyond `range.end`: returns error
+    ///
+    /// # Returns
+    ///
+    /// A journal ready for sync operations with size in the given range.
+    pub(crate) async fn init_sync(
```
danlaine, line 383: Moved from `storage/src/qmdb/any/unordered/fixed/sync.rs` (and renamed from `init_journal`).

patrick-ogrady:
Other option...I think we could use a "sentinel" for the offset in the fixed journal (if an item is in fixed with a U64::MAX offset, we can conclude it was "filled")?

danlaine:
Going to take another crack at solving this by adding metadata rather than using dummy/sentinel values.

PR #2773, opened 2026-01-09
[storage/journal] replay stuff

roberto-bayardo:
…iding the vector

### DIFF
--- a/storage/src/journal/segmented/variable.rs
+++ b/storage/src/journal/segmented/variable.rs
@@ -88,9 +88,12 @@ use commonware_runtime::{
     Blob, Metrics, Storage,
 };
 use futures::stream::{self, Stream, StreamExt};
-use std::{borrow::Cow, io::Cursor, num::NonZeroUsize};
+use std::num::NonZeroUsize;
 use tracing::{trace, warn};
 use zstd::{bulk::compress, decode_all};
+
+/// Maximum size of a varint for u32 (also the minimum useful read size for parsing item headers).
+const MAX_VARINT_SIZE: usize = 5;
 
 /// Configuration for `Journal` storage.
 #[derive(Clone)]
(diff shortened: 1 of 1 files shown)


storage/src/journal/segmented/variable.rs
```diff
@@ -229,27 +226,31 @@ impl<E: Storage + Metrics, V: CodecShared> Journal<E, V> {
...
-                mut buffer,
-                filled,
+                prefix,
+                size: item_size,
                 read_offset,
             } => {
-                blob.read_into(&mut buffer[filled..], read_offset).await?;
-                Cow::Owned(buffer)
+                let mut buffer = vec![0u8; item_size];
+                buffer[..prefix.len()].copy_from_slice(prefix);
```
patrick-ogrady: I think we can create a Buf here and chain together to avoid this copy

storage/src/journal/segmented/variable.rs
```diff
@@ -229,27 +226,31 @@ impl<E: Storage + Metrics, V: CodecShared> Journal<E, V> {
         blob: &Append<E::Blob>,
         offset: u64,
     ) -> Result<(u64, u32, V), Error> {
-        // Read varint header (max 5 bytes for u32)
-        let buf = vec![0u8; MAX_VARINT_SIZE];
+        // Read varint header (max 5 bytes for u32). But allocate 64 bytes which will save one
+        // vector allocation for small enough items.
+        let buf = vec![0u8; 64];
```
patrick-ogrady: nit: this should be a const

storage/src/journal/segmented/variable.rs
```diff
@@ -355,28 +360,39 @@ impl<E: Storage + Metrics, V: CodecShared> Journal<E, V> {
...
                                     },
                                 )) => {
                                     // Item::Incomplete means the item spans buffer boundary: we have
-                                    // the varint + partial data, but not the complete item. By
-                                    // definition, this consumed the entire buffer (varint_len +
-                                    // filled = buf.len()), so advance past all of it.
+                                    // the varint + partial data, but not the complete item.
+                                    // Reuse buffer, growing if needed. Copy prefix BEFORE advancing
+                                    // (which invalidates the prefix slice).
+                                    reuse_buf.resize(item_size, 0);
```
patrick-ogrady: like above, I think we can "chain" these?

Issue #2749, opened 2026-01-08
[storage/journal] Use `segmented::fixed` in `contiguous::fixed`


patrick-ogrady:
To deduplicate code, we should probably use `segmented::fixed` to power `contiguous::fixed`?

PR #2774, opened 2026-01-09
[storage/journal] implement contiguous::fixed journal using segmented::fixed

roberto-bayardo:
Gets rid of some of the code duplication across them.

Resolves: https://github.com/commonwarexyz/monorepo/issues/2749

### DIFF
--- a/storage/src/journal/contiguous/fixed.rs
+++ b/storage/src/journal/contiguous/fixed.rs
@@ -103,7 +103,7 @@ pub struct Config {
 /// and
 /// [rocksdb](https://github.com/facebook/rocksdb/blob/0c533e61bc6d89fdf1295e8e0bcee4edb3aef401/include/rocksdb/options.h#L441-L445),
 /// the first invalid data read will be considered the new end of the journal (and the
-/// underlying [Blob] will be truncated to the last valid item). Repair is performed
+/// underlying blob will be truncated to the last valid item). Repair is performed
 /// by the underlying [SegmentedJournal] during init.
 pub struct Journal<E: Storage + Metrics, A: CodecFixedShared> {
     inner: SegmentedJournal<E, A>,


storage/src/journal/contiguous/fixed.rs
```diff
@@ -449,85 +345,78 @@ impl<E: Storage + Metrics, A: CodecFixedShared> Journal<E, A> {
...
+                                        items_per_blob - start_pos_in_section
+                                    } else {
+                                        items_per_blob
+                                    };
+
+                                    let got_items = if prev_section == start_section {
+                                        *last_pos - start_pos_in_section + 1
+                                    } else {
+                                        *last_pos + 1
+                                    };
```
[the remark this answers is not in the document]
roberto-bayardo: The calculation is correct. If we used just *last_pos + 1, we'd incorrectly count positions we'd never read.

storage/src/journal/segmented/fixed.rs
```diff
@@ -146,73 +146,118 @@ impl<E: Storage + Metrics, A: CodecFixedShared> Journal<E, A> {
...
+                                        )),
                                     }
                                 }
+                                Err(RError::BlobInsufficientLength) => {
+                                    Some((
+                                        Err(Error::Corruption(format!(
+                                            "trailing bytes in section {section} at position {position}"
+                                        ))),
+                                        (section, buf, blob, reader, blob_size, valid_size, blob_size),
+                                    ))
```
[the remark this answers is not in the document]
roberto-bayardo: I don't think that's correct, but since we're returning error which will abort the stream it doesn't matter so much anyway.

storage/src/journal/contiguous/fixed.rs
```diff
@@ -413,32 +314,27 @@ impl<E: Storage + Metrics, A: CodecFixedShared> Journal<E, A> {
...
-            if offset >= self.tail.size().await {
-                return Err(Error::ItemOutOfRange(pos));
+        self.inner.get(section, pos_in_section).await.map_err(|e| {
+            // Map segmented errors to contiguous errors
+            match e {
+                Error::SectionOutOfRange(_)
+                | Error::AlreadyPrunedToSection(_)
+                | Error::ItemOutOfRange(_) => {
+                    unreachable!("range should have already been checked")
+                }
```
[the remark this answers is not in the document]
roberto-bayardo: getting rid of the gap detection is fine, but probably we should return error corrupted instead of panicking.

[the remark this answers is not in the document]
roberto-bayardo: Changed to return Error::Corruption.

storage/src/journal/contiguous/fixed.rs
```diff
@@ -413,185 +313,108 @@ impl<E: Storage + Metrics, A: CodecFixedShared> Journal<E, A> {
...
+        // Check all non-tail sections in range are complete before starting the stream.
+        let oldest = self.inner.oldest_section().unwrap_or(start_section);
+        if let Some(newest) = self.inner.newest_section() {
+            for section in start_section.max(oldest)..newest {
+                let len = self.inner.section_len(section).await?;
+                let expected = if section == start_section {
+                    items_per_blob - start_pos_in_section
+                } else {
+                    items_per_blob
+                };
```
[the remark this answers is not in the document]
roberto-bayardo: good bot, fixed.

storage/src/journal/segmented/fixed.rs
```diff
@@ -146,73 +146,83 @@ impl<E: Storage + Metrics, A: CodecFixedShared> Journal<E, A> {
...
-            let reader = blob.as_blob_reader(buffer).await?;
-            blob_info.push((section, blob.clone(), reader, blob_size));
+            let mut reader = blob.as_blob_reader(buffer).await?;
+            // For the first section, seek to the start position
+            let initial_offset = if is_first {
+                reader.seek_to(start_byte_offset)?;
+                start_byte_offset
+            } else {
+                0
+            };
```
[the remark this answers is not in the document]
roberto-bayardo: we should return ItemPruned error in this case I think.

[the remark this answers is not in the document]
roberto-bayardo: reverted to old behavior

storage/src/journal/segmented/fixed.rs
```diff
@@ -423,7 +472,7 @@ mod tests {
 
             let items = {
                 let stream = journal
-                    .replay(0, NZUsize!(1024))
+                    .replay(0, 0, NZUsize!(1024))
```
patrick-ogrady, line 474: Should we add some test with non-zero start?

storage/src/journal/segmented/fixed.rs
```diff
@@ -319,6 +334,40 @@ impl<E: Storage + Metrics, A: CodecFixedShared> Journal<E, A> {
...
+
+        Ok(())
+    }
+
+    /// Ensure a section exists, creating an empty blob if needed.
+    ///
+    /// This is used to maintain the invariant that at least one blob always exists
+    /// (the "tail" blob), which allows reconstructing journal size on reopen.
+    pub async fn ensure_section_exists(&mut self, section: u64) -> Result<(), Error> {
+        self.manager.get_or_create(section).await?;
```
patrick-ogrady, line 367: I think we should just call `self.manager.get_or_create()`? It does the same thing?

storage/src/journal/segmented/fixed.rs
```diff
@@ -319,6 +334,40 @@ impl<E: Storage + Metrics, A: CodecFixedShared> Journal<E, A> {
...
+
+    /// Initialize a section with a specific number of zero-filled items.
+    ///
+    /// This creates the section's blob and fills it with `item_count` items worth of zeros.
+    /// The data is written through the Append wrapper which handles checksums properly.
+    ///
+    /// # Arguments
+    /// * `section` - The section number to initialize
+    /// * `item_count` - Number of zero-filled items to write
+    pub async fn init_section_at_size(
```
patrick-ogrady: We should add a test for this?

storage/src/journal/segmented/fixed.rs
```diff
@@ -187,76 +187,91 @@ impl<E: Storage + Metrics, A: CodecFixedShared> Journal<E, A> {
...
-            blob_info.push((section, replay));
+            let blob_size = blob.size().await;
+            let mut replay = blob.replay(buffer).await?;
+            // For the first section, seek to the start position
+            let initial_position = if section == start_section {
+                let start = start_position * Self::CHUNK_SIZE_U64;
+                if start > blob_size {
+                    return Err(Error::ItemOutOfRange(start_position));
+                }
+                replay.seek_to(start as usize).await?;
```
patrick-ogrady: I think `seek_to` should take `u64`?

patrick-ogrady: If I didn't follow this advice elsewhere, please lmk

storage/src/journal/segmented/fixed.rs
```diff
@@ -187,76 +187,91 @@ impl<E: Storage + Metrics, A: CodecFixedShared> Journal<E, A> {
...
-            let replay = blob.replay(buffer).await?;
-            blob_info.push((section, replay));
+            let blob_size = blob.size().await;
+            let mut replay = blob.replay(buffer).await?;
+            // For the first section, seek to the start position
+            let initial_position = if section == start_section {
+                let start = start_position * Self::CHUNK_SIZE_U64;
+                if start > blob_size {
+                    return Err(Error::ItemOutOfRange(start_position));
+                }
```
patrick-ogrady, line 202: I'm assuming this byte-filling is already an established pattern?

storage/src/journal/contiguous/fixed.rs
```diff
@@ -426,226 +313,103 @@ impl<E: Storage + Metrics, A: CodecFixedShared> Journal<E, A> {
...
+        // Calculate the tail section.
+        let tail_section = self.size / self.items_per_blob;
+
+        // Cap to tail section. The tail section is guaranteed to exist by our invariant.
+        let min_section = std::cmp::min(target_section, tail_section);
+
+        let pruned = self.inner.prune(min_section).await?;
         if pruned {
-            self.pruning_boundary = new_oldest_blob * self.cfg.items_per_blob.get();
+            self.pruning_boundary = min_section * self.items_per_blob;
```
patrick-ogrady: Do we need to maintain the pruning_boundary here? I "think" it is already tracked by `inner` (can call oldest_section())?

storage/src/journal/contiguous/fixed.rs
```diff
@@ -106,205 +86,155 @@ pub struct Config {
     pub write_buffer: NonZeroUsize,
 }
 
-/// A contiguous journal with fixed-size entries.
+/// Implementation of `Journal` storage.
 ///
-/// Items are stored sequentially across multiple blobs, with automatic blob rotation
-/// when a blob reaches `items_per_blob` items.
-///
-/// # Repair
```
patrick-ogrady, line 114: imho we should retain this repair comment?

runtime/src/utils/buffer/pool/read.rs
```diff
@@ -315,6 +315,31 @@ impl<B: Blob> Replay<B> {
...
+
+    /// Seeks to `offset` in the blob, returning `Err(BlobInsufficientLength)` if `offset` exceeds
+    /// the blob size.
+    pub async fn seek_to(&mut self, offset: usize) -> Result<(), Error> {
+        if offset as u64 > self.reader.blob_size() {
+            return Err(Error::BlobInsufficientLength);
+        }
+
+        // Clear existing buffer state
+        self.buffer.buffers.clear();
```
patrick-ogrady: I think this would be better handled with a dedicated `.reset()` or `.clear()` method instead of introspecting?

storage/src/journal/contiguous/fixed.rs
roberto-bayardo: it's not quite accurate though. If there's invalid internal segment we return Corruption error rather than try to repair it.  The only "repair" is to rewind the tail blob if it's incomplete.

storage/src/journal/segmented/fixed.rs
roberto-bayardo: yes, it's previous behavior for state syncing that starts in the middle of the blob.

storage/src/journal/contiguous/fixed.rs
roberto-bayardo: no we don't, changed. Technically I guess we don't really need to maintain size() either, should I remove that caching as well?

storage/src/journal/contiguous/fixed.rs
roberto-bayardo: Oh hmm I guess it is accurate since the segmented journal does this on init now (not replay).

roberto-bayardo: OK clarified in both segmented and contiguous that that blobs are trimmed of incomplete items on init to handle the case where an underlying write fails at a page boundary and the last item in the page gets truncated.

storage/src/journal/segmented/fixed.rs
roberto-bayardo: my bad, changed.

storage/src/journal/segmented/fixed.rs
roberto-bayardo: this is called from contiguous::fixed journal, are you ok with this introspection?

runtime/src/utils/buffer/pool/read.rs
roberto-bayardo: sorry misunderstood the first request, fixed now.

storage/src/journal/segmented/fixed.rs
roberto-bayardo: Done. (There was one in contiguous which exercises this, but probably good to exercise directly as well.)

storage/src/journal/segmented/fixed.rs
roberto-bayardo: added

storage/src/journal/contiguous/fixed.rs
```diff
@@ -106,205 +91,203 @@ pub struct Config {
...
-                .map_err(Error::Runtime)?;
-            let index = match name.try_into() {
-                Ok(index) => u64::from_be_bytes(index),
-                Err(nm) => return Err(Error::InvalidBlobName(hex(&nm))),
-            };
-            debug!(blob = index, size, "loaded blob");
-            blobs.insert(index, (blob, size));
-        }
 
-        // Check that there are no gaps in the historical blobs.
```
roberto-bayardo, line 188: Removed this integrity checking during init() in favor of integrity checking during replay. This means you only check integrity of data that matters (old corrupted blobs may just get quietly pruned..) and you also aren't subject to silent replay failures should corruption happen after init.

runtime/src/utils/buffer/pool/read.rs
```diff
@@ -315,6 +323,25 @@ impl<B: Blob> Replay<B> {
...
+        }
+
+        self.buffer.clear();
+        self.exhausted = false;
+
+        let page_size = self.reader.logical_page_size as u64;
+        self.reader.blob_page = offset / page_size;
+
+        let remainder = (offset % page_size) as usize;
+        assert!(self.ensure(remainder).await?, "bounds already checked");
```
patrick-ogrady: Do we need to call `ensure`? I think the application should be calling this before it reads?

Not to mention, I think this is ensuring all the bytes up to the start offset (which seems wasteful)?

storage/src/journal/contiguous/fixed.rs
```diff
@@ -106,205 +91,203 @@ pub struct Config {
...
+        let pos_in_section = position % self.items_per_blob;
+        (section, pos_in_section)
     }
 
     /// Sync any pending updates to disk.
     pub async fn sync(&mut self) -> Result<(), Error> {
-        self.synced.inc();
-        debug!(blob = self.tail_index, "syncing blob");
-        self.tail.sync().await.map_err(Error::Runtime)
+        self.inner.sync_all().await
```
patrick-ogrady: Do we need to sync_all or just the tip?

runtime/src/utils/buffer/pool/read.rs
roberto-bayardo: Without it, test_replay_seek_to fails, but let me look if there's a better way.

storage/src/journal/segmented/fixed.rs
```diff
@@ -58,18 +58,13 @@ pub struct Config {
...
 ///
-/// Like
-/// [sqlite](https://github.com/sqlite/sqlite/blob/8658a8df59f00ec8fcfea336a2a6a4b5ef79d2ee/src/wal.c#L1504-L1505)
-/// and
-/// [rocksdb](https://github.com/facebook/rocksdb/blob/0c533e61bc6d89fdf1295e8e0bcee4edb3aef401/include/rocksdb/options.h#L441-L445),
-/// the first invalid data read will be considered the new end of the journal (and the
-/// underlying [Blob] will be truncated to the last valid item). Repair occurs during
-/// init by checking each blob's size.
+/// If a write fails at a page boundary, there's a chance the journal will contain a "partial" item
+/// at the end of the last appended blob. All blobs are checked on init and any incomplete items are
```
patrick-ogrady: This "incomplete" items handles things that span multiple pages correctly?

storage/src/journal/contiguous/fixed.rs
patrick-ogrady: The same is also true of `variable` I suppose (may speed things up a good bit 👀 )

runtime/src/utils/buffer/pool/read.rs
roberto-bayardo: (Since we set blob_page though I don't think it's buffering anything other than the page containing the requested seek offset)

roberto-bayardo: ok fixed, though had to introduce a `first_fill_skip` to handle the first buffer fill after seeking properly.  Extended the test case to exercise this appropriately.

storage/src/journal/segmented/fixed.rs
roberto-bayardo: It should, since the truncated size logic is simply:

`let valid_size = size - (size % Self::CHUNK_SIZE_U64);`

... which is page size agnostic.  I could add a test though?

storage/src/journal/segmented/fixed.rs
```diff
@@ -319,6 +329,40 @@ impl<E: Storage + Metrics, A: CodecFixedShared> Journal<E, A> {
...
+
+    /// Initialize a section with a specific number of zero-filled items.
+    ///
+    /// This creates the section's blob and fills it with `item_count` items worth of zeros.
+    /// The data is written through the Append wrapper which handles checksums properly.
+    ///
+    /// # Arguments
+    /// * `section` - The section number to initialize
+    /// * `item_count` - Number of zero-filled items to write
+    pub async fn init_section_at_size(
```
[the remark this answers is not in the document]
roberto-bayardo: done

storage/src/journal/segmented/fixed.rs
```diff
@@ -319,6 +329,40 @@ impl<E: Storage + Metrics, A: CodecFixedShared> Journal<E, A> {
...
+        blob.resize(target_size).await?;
+
+        Ok(())
+    }
+
+    /// Ensure a section exists, creating an empty blob if needed.
+    ///
+    /// This is used to maintain the invariant that at least one blob always exists
+    /// (the "tail" blob), which allows reconstructing journal size on reopen.
+    pub async fn ensure_section_exists(&mut self, section: u64) -> Result<(), Error> {
```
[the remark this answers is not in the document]
roberto-bayardo: done

runtime/src/utils/buffer/pool/read.rs
patrick-ogrady: Pushed a small cleanup to this

storage/src/journal/contiguous/fixed.rs
patrick-ogrady: This is similar to variable (and probably is ok).

patrick-ogrady (approved):

patrick-ogrady (approved):

PR #2765, opened 2026-01-09
[cryptography/bls12381] Reduce Batch Verify Latency

patrick-ogrady:
Replaces: #2764
Replaces: #2759

PR #2759, opened 2026-01-08
[cryptography/bls12381] Improve Batch Verify Performance

patrick-ogrady:

PR #2777, opened 2026-01-09
[cryptography] Use parallelism and small scalars for batch verification

cronokirby:
Integrate the ideas of #2765, #2764, #2759

### DIFF
--- a/cryptography/src/bls12381/dkg.rs
+++ b/cryptography/src/bls12381/dkg.rs
@@ -303,7 +303,7 @@ use commonware_parallel::{Sequential, Strategy};
 use commonware_utils::Bft3f1;
 use commonware_utils::{
     ordered::{Map, Quorum, Set},
-    Bft3f1, Faults, Participant, TryCollect, NZU32,
+    Faults, Participant, TryCollect, NZU32,
 };
 use core::num::NonZeroU32;
 use rand_core::CryptoRngCore;
--- a/cryptography/src/bls12381/benches/dkg.rs
+++ b/cryptography/src/bls12381/benches/dkg.rs
@@ -147,9 +147,9 @@ fn benchmark_dkg(c: &mut Criterion, reshare: bool) {
                         || bench.pre_finalize(),
                         |(player, logs)| {
                             if concurrency > 1 {
-                                black_box(player.finalize::<_, Bft3f1>(logs, &strategy).unwrap());
+                                black_box(player.finalize::<Bft3f1>(logs, &strategy).unwrap());
                             } else {
-                                black_box(player.finalize::<_, Bft3f1>(logs, &Sequential).unwrap());
+                                black_box(player.finalize::<Bft3f1>(logs, &Sequential).unwrap());
                             }
                         },
                         BatchSize::SmallInput,


cronokirby:
[bench.txt](https://github.com/user-attachments/files/24538686/bench.txt)
Some local benchmarks against main.

consensus/src/simplex/scheme/bls12381_threshold.rs
```diff
@@ -559,32 +559,48 @@ impl<P: PublicKey, V: Variant> certificate::Scheme for Scheme<P, V> {
...
+                    strategy,
+                ) {
+                    Ok(()) => BTreeSet::new(),
+                    Err(errs) => errs.into_iter().map(|p| p.index).collect(),
+                }
+            },
+        );
+
+        // Merge invalid sets
+        let invalid: BTreeSet<_> = vote_invalid.union(&seed_invalid).copied().collect();
```
patrick-ogrady, line 603: nit: remove `\n`

cryptography/src/bls12381/benches/pairing_verify.rs
```diff
@@ -0,0 +1,32 @@
...
+            || {
+                let mut rng = StdRng::seed_from_u64(0);
+                let (private, public) = primitives::ops::keypair::<_, MinSig>(&mut rng);
+                let sig = primitives::ops::sign_message::<MinSig>(&private, b"ns", b"msg");
+                let hm =
+                    primitives::ops::hash_with_namespace::<MinSig>(MinSig::MESSAGE, b"ns", b"msg");
+                (public, hm, sig)
+            },
+            |(public, hm, sig)| {
+                black_box(MinSig::verify(&public, &hm, &sig)).unwrap();
```
patrick-ogrady: We already benchmark this in `signature_verification.rs` (albeit via `scheme`). Not sure if we want to rename it at all to make it clearer/less dup?

cryptography/src/bls12381/benches/threshold_batch_verify_same_message.rs
```diff
@@ -18,62 +19,93 @@ fn benchmark_threshold_batch_verify_same_message(c: &mut Criterion) {
...
-                                .collect::<Vec<_>>();
-                            (rng, output.public().clone(), signatures)
-                        },
-                        |(mut rng, polynomial, mut signatures)| {
-                            // Shuffle faults
-                            if invalid > 0 {
-                                signatures.shuffle(&mut rng);
-                            }
+            for concurrency in [1, 8] {
+                let strategy = Rayon::new(NonZeroUsize::new(concurrency).unwrap()).unwrap();
```
patrick-ogrady: I think this is better defined inline if concurrency > 1?

cryptography/src/bls12381/benches/threshold_batch_verify_same_message_precomputed.rs
```diff
@@ -18,64 +19,95 @@ fn benchmark_threshold_batch_verify_same_message_precomputed(c: &mut Criterion)
...
-                                .collect::<Vec<_>>();
-                            (rng, polynomial, signatures)
-                        },
-                        |(mut rng, polynomial, mut signatures): (_, _, Vec<_>)| {
-                            // Shuffle faults
-                            if invalid > 0 {
-                                signatures.shuffle(&mut rng);
-                            }
+            for concurrency in [1, 8] {
+                let strategy = Rayon::new(NonZeroUsize::new(concurrency).unwrap()).unwrap();
```
patrick-ogrady: better defined inline if concurrency > 1?

cryptography/src/bls12381/dkg.rs
```diff
@@ -1315,10 +1315,10 @@ struct ObserveInner<V: Variant, P: PublicKey> {
 }
 
 impl<V: Variant, P: PublicKey> ObserveInner<V, P> {
-    fn reckon<S: ParStrategy>(
+    fn reckon(
         info: Info<V, P>,
         selected: Map<P, DealerLog<V, P>>,
-        strategy: &S,
+        strategy: &impl ParStrategy,
```
patrick-ogrady: nit: just `Strategy`?

cryptography/src/bls12381/primitives/ops/batch.rs
```diff
@@ -13,16 +13,120 @@
...
+        let mut values = BTreeMap::new();
+        let len = leaves.len();
+        if len == 0 {
+            return Self { len, values };
+        }
+
+        // Use an explicit stack to build bottom-up with halving intervals.
+        // Phase 0 = first visit (push children), Phase 1 = second visit (compute value)
+        let mut stack: Vec<(usize, usize, u8)> = vec![(0, len, 0)];
+
```
patrick-ogrady, line 25: nit: remove `\n`

cryptography/src/bls12381/primitives/ops/batch.rs
```diff
@@ -13,16 +13,120 @@
...
+fn bisect_par<V: Variant>(
+    entries: &[(V::Public, V::Signature)],
+    hm: &V::Signature,
+    par: &impl Strategy,
+) -> Vec<usize> {
+    if entries.is_empty() {
+        return Vec::new();
+    }
+    let par_hint = par.parallelism_hint();
```
patrick-ogrady: I think this pre-setup may lead to lower performance in the case bisection is one "section" (I though my "batched verify" was the right balance here)

cryptography/src/bls12381/primitives/ops/batch.rs
```diff
@@ -46,6 +151,7 @@ pub fn verify_same_message<R, V>(
     namespace: &[u8],
     message: &[u8],
     entries: &[(V::Public, V::Signature)],
+    par: &impl Strategy,
```
patrick-ogrady, line 204: nit: `strategy:`

cryptography/src/bls12381/primitives/ops/batch.rs
```diff
@@ -13,16 +13,120 @@
...
+        }
+        (0..self.len).filter(|x| !good.contains(x)).collect()
+    }
+}
+
+fn bisect<V: Variant>(entries: &[(V::Public, V::Signature)], hm: &V::Signature) -> Vec<usize> {
+    SumTree::<V>::build(entries).verify(hm)
+}
+
+fn bisect_par<V: Variant>(
```
patrick-ogrady: nit: `bisect` and just call `SumTree::<V>::build(entries).verify(hm)` inline?

cryptography/src/bls12381/primitives/variant.rs
```diff
@@ -145,12 +119,12 @@ impl Variant for MinPk {
...
     ///
     /// Source: <https://ethresear.ch/t/security-of-bls-batch-verification/10748>
-    fn batch_verify<R: CryptoRngCore, S: Strategy>(
+    fn batch_verify<R: CryptoRngCore>(
         rng: &mut R,
         publics: &[Self::Public],
         hms: &[Self::Signature],
         signatures: &[Self::Signature],
-        strategy: &S,
+        par: &impl Strategy,
```
patrick-ogrady, line 127: nit: `strategy:`
nit: if we already are using <>, should we just keep it there?

cryptography/src/bls12381/primitives/variant.rs
```diff
@@ -235,32 +189,7 @@ impl Variant for MinSig {
...
-        // Aggregate `e(pk,hm)`
-        pairing.raw_aggregate(&p, &q);
-
-        // Finalize the pairing accumulation and verify the result
-        //
-        // If `finalverify()` returns `true`, it means `e(pk,hm) * e(-G2::one(),sig) == 1`. This
-        // is equivalent to `e(pk,hm) == e(G2::one(),sig)`.
-        pairing.commit();
-        if !pairing.finalverify(None) {
+        if !G1::multi_pairing_check(&[*hm], &[*public], signature, &-G2::generator()) {
```
patrick-ogrady, line 192: `&-G2::generator()`

Would using a const be better than negating on the fly?

math/src/poly.rs
```diff
@@ -126,7 +126,7 @@ impl<K> Poly<K> {
     /// This method uses more scratch space, and requires cloning values of
     /// type `R` more, but should be better if [`Space::msm`] has a better algorithm
     /// for `K`.
-    pub fn eval_msm<R: Ring, S: ParStrategy>(&self, r: &R, strategy: &S) -> K
+    pub fn eval_msm<R: Ring>(&self, r: &R, strategy: &impl ParStrategy) -> K
```
patrick-ogrady: nit: `Strategy`

cryptography/src/bls12381/primitives/group.rs
```diff
@@ -98,12 +105,61 @@ impl arbitrary::Arbitrary<'_> for Scalar {
...
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct SmallScalar {
+    /// Stored as blst_scalar with only lower 128 bits populated.
+    inner: blst_scalar,
+}
+
+impl SmallScalar {
+    /// Generates a random 128-bit scalar.
+    pub fn random(mut rng: impl CryptoRngCore) -> Self {
+        let mut bytes = [0u8; 32]; // blst_scalar is 32 bytes
```
patrick-ogrady: nit: put this comment above `bytes`?

cryptography/src/bls12381/primitives/group.rs
```diff
@@ -604,6 +660,96 @@ impl G1 {
...
+    }
+
+    /// Checks that `sum_i (p1[i] ⊙ p2[i]) + t1 ⊙ t2 == 0`.
+    ///
+    /// `p1` and `p2` MUST have the same length.
+    #[must_use]
+    pub(crate) fn multi_pairing_check(p1: &[Self], p2: &[G2], t1: &Self, t2: &G2) -> bool {
+        assert_eq!(p1.len(), p2.len());
+
+        let mut pairing = Pairing::new(false, &[]);
```
patrick-ogrady, line 689: nit: add a `comment

cryptography/src/bls12381/primitives/group.rs
```diff
@@ -604,6 +660,96 @@ impl G1 {
...
+        let p1_affine = Self::batch_to_affine(p1);
+        let p2_affine = G2::batch_to_affine(p2);
+        for (p1, p2) in iter::once((&t1.as_blst_p1_affine(), &t2.as_blst_p2_affine()))
+            .chain(p1_affine.iter().zip(p2_affine.iter()))
+        {
+            pairing.raw_aggregate(p2, p1);
+        }
```
patrick-ogrady, line 696: nit: add a `\n`

cryptography/src/bls12381/primitives/group.rs
```diff
@@ -888,6 +1012,83 @@ impl G2 {
...
+
+        out
+    }
+
+    /// Checks that `sum_i (p1[i] ⊙ p2[i]) + t1 ⊙ t2 == 0`.
+    ///
+    /// `p1` and `p2` MUST have the same length.
+    #[must_use]
+    pub(crate) fn multi_pairing_check(p1: &[Self], p2: &[G1], t1: &Self, t2: &G1) -> bool {
+        G1::multi_pairing_check(p2, p1, t2, t1)
```
patrick-ogrady, line 1040: nice

cryptography/src/bls12381/primitives/variant.rs
cronokirby: The cost of doing a pairing dwarfs the cost of negating a group element (one scalar addition, basically).

To make a const you'd need to add an additional constant just for this value, which seems unnecessary.

math/src/poly.rs
cronokirby: There's a naming conflict later with proptest's `Strategy`. We could rename that one instead.

cryptography/src/bls12381/primitives/variant.rs
cronokirby: I prefer the rule of consistently using `&impl Strategy` (which should always be possible) rather than having it depend on whether or not there happen to be generics. We could modify things to use `&mut impl CryptoRngCore` as well though.

cryptography/src/bls12381/primitives/ops/batch.rs
cronokirby: I think `bisect` serves as a nice logical abstraction over the particular choice of algorithm, and it makes it clear that the way parallelization happens is by calling the algorithm over chunks, rather than calling a specific implementation over chunks. This is important because the correctness comes from just using the algorithm abstractly, not the particular implementation.

In other words, the whole `SumTree` stuff could be inlined into `bisect`.

cryptography/src/bls12381/primitives/ops/batch.rs
cronokirby: There might be some data structure overhead (this could be optimized), but you compute no fewer additions than you would need to check the top level pairing.

cryptography/src/bls12381/benches/pairing_verify.rs
cronokirby: I'd be in favor of removing it if it overlaps

consensus/src/simplex/scheme/bls12381_threshold.rs
cronokirby: I think this one actually looks better with the extra line, given how heavy the previous expression is

cryptography/src/bls12381/benches/threshold_batch_verify_same_message.rs
cronokirby: They have different types, so you'd have to pull out the relevant functions and make them generic...

cronokirby: I think if we're concerned about the overhead of this, we should just make Rayon only actually create a thread pool for > 1 threads.

cryptography/src/bls12381/primitives/ops/batch.rs
```diff
@@ -57,52 +162,32 @@ where
...
-        .zip(&scalars)
-        .map(|((pk, _), s)| *pk * s)
-        .collect();
-    let weighted_sigs: Vec<V::Signature> = entries
-        .iter()
-        .zip(&scalars)
-        .map(|((_, sig), s)| *sig * s)
-        .collect();
+    // Generate 128-bit random scalars (sufficient for batch verification security)
+    let scalars: Vec<SmallScalar> = (0..len).map(|_| SmallScalar::random(&mut *rng)).collect();
```
patrick-ogrady: I think we can just call `entries.len()` inline?

cryptography/src/bls12381/benches/pairing_verify.rs
patrick-ogrady, line 1: Let's remove this.

cryptography/src/bls12381/benches/threshold_batch_verify_same_message.rs
patrick-ogrady: We already hardcode Sequential in the conc 1 case, so I think we're already past that?

cryptography/src/bls12381/benches/pairing_verify.rs
patrick-ogrady: Yeah, let's do it.

cryptography/src/bls12381/primitives/variant.rs
patrick-ogrady: I think we already have a const for this (or did)?

cryptography/src/bls12381/primitives/ops/batch.rs
patrick-ogrady: If you are locked in on having a standalone `bisect` and `bisect_par` function, I think it is better to keep SumTree outside the function definition.

I just don't think you'd ever call `bisect` directly and thus we can take the win on having `bisect` be the thing called in other functions.

cryptography/src/bls12381/primitives/ops/batch.rs
patrick-ogrady: I'm referring more to pre-allocating cores to a specific subset of the range rather than retaining their capacity by level.

cryptography/src/bls12381/benches/threshold_batch_verify_same_message_precomputed.rs
cronokirby: So, I think the reason we don't want to define this inline is to not benchmark the overhead of creating the thread pool, so having this outside makes sense.

cryptography/src/bls12381/benches/threshold_batch_verify_same_message.rs
cronokirby: see below comment about pool creation overhead

cronokirby:
False positive on the conformance, just changing the generation pattern

patrick-ogrady (approved):

PR #2782, opened 2026-01-10
[codec] Switch to LE

### LINKED ISSUE #2781: [codec] Switch Default Encoding to Little-Endian
Modern processors are natively little-endian. By switching to Little-Endian encoding, we enable the compiler to emit direct load/store instructions rather than byte-swapping operations when reading data off the wire (~30-50% faster).

This also unblocks the creation of a zero-copy decoder that maps bytes directly into a struct in-memory (can't do this if types are represented as big-endian on a little-endian system).

patrick-ogrady:
Closes: #2781


clabby:

Issue #2639, opened 2025-12-29
State root mismatch before and after commit when using QMDB


NikhilSharmaWe:
## Description

When using the **QMDB** crate inside a monorepo, I am observing a **state root mismatch** between:

- the state root obtained after calling `merkleize` (before `commit`), and  
- the state root obtained after calling `commit`.

This happens even when no additional state changes are made between these two steps.

## Expected Behavior

- `merkleize` computes the state root based on the in-memory state after execution.
- `commit` should only persist the already-computed state to the database.

Given this, the state root **should remain the same before and after `commit`**, since no new state transitions are introduced during persistence.

## Actual Behavior

- Calling `commit` introduces a new commit operation that changes the resulting state root.
- As a result, the state root after `merkleize` does not match the state root after `commit`.

## Question / Clarification Needed

Is this the intended design for QMDB?

Specifically:
- Should `commit` affect the state root?
- Or should the state root be fully determined by execution + `merkleize`, with `commit` only persisting data to storage?

This distinction is important for verification logic, where the state root is expected to remain stable between execution and persistence.

roberto-bayardo:
Thanks for the report. `commit` in our QMDB implementations is a state-changing operation, as it:
- triggers "floor raising" (aka compaction) which repositions older operations to eliminate gaps and
- writes an explicit "commit" operation to the log which contains the new `inactivity_floor` location.

Since the state root in QMDB is the state of the log (and not strictly the state of the active key/values) both these steps will affect it.

I will clarify this in the documentation before closing this issue.

PR #2786, opened 2026-01-12
[storage/qmdb] remove commit capability from Merkleized state

roberto-bayardo:
We were allowing committing of a qmdb when in the Merkleized state, but this isn't quite correct, since commit doesn't just ensure durability, but it also changes the DB root state. It should therefore be allowed only in the Mutable state.
This PR removes this capability, which was unlikely to be used anyway.   It also clearly documents that commit is always a root-state changing operation.

Typically you commit and only THEN merkleize, in which case there would be no reason to immediately commit again. You'd want to go back into the mutable state first.

Resolves: https://github.com/commonwarexyz/monorepo/issues/2639

### DIFF
--- a/storage/src/qmdb/current/unordered/fixed.rs
+++ b/storage/src/qmdb/current/unordered/fixed.rs
@@ -966,18 +966,14 @@ pub mod test {
             }
             assert!(db.status.get_bit(*db.op_count() - 1));
 
-            // Test that we can do a non-durable root.
+            // Test that we can get a non-durable root.
             let mut db = db.into_mutable();
             db.update(k1, v1).await.unwrap();
             let db = db.into_merkleized().await.unwrap();
             assert_ne!(db.root(), root3);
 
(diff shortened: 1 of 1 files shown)


patrick-ogrady:
Should we attach #2639?

patrick-ogrady:
Was this support added to allow for merklization before commit (to get intermediate results)? Or am I missing something?

roberto-bayardo:
> Should we attach #2639?

done (and appropriately updated the docs as I suggested there)

roberto-bayardo:
> Was this support added to allow for merklization before commit (to get intermediate results)? Or am I missing something?

Yes you can still merkleize before commit, but it wasn't added for that reason.  It's just one of the four possible states when you decouple merklization from durablility.  The main goal was to allow committing without merkleizing, which I expect to be the usual use case. Everything else was just icing (and I guess I put on too much icing)

patrick-ogrady (approved):

PR #2790, opened 2026-01-12
[utils] update test for array intersection and difference

ikwuoz:
This PR also removes the needless call to `min()` as `limit` is already restrained by prior `if` checks.


patrick-ogrady (approved):
nice

PR #2792, opened 2026-01-12
remove unnesscary allocs

malik672:
remove unnecessary allocation

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -559,11 +559,7 @@ pub trait Storage: Clone + Send + Sync + 'static {
     ) -> impl Future<Output = Result<(Self::Blob, u64), Error>> + Send {
         async move {
             let (blob, size, _) = self
-                .open_versioned(
-                    partition,
-                    name,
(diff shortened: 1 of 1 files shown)


patrick-ogrady:
Once you fix `fmt`, should be g2g (although you refer to it as a train wreck 👀 )!

malik672:
> Once you fix `fmt`, should be g2g (although you refer to it as a train wreck 👀 )!

😂 fixed

runtime/src/lib.rs
```diff
@@ -557,15 +557,9 @@ pub trait Storage: Clone + Send + Sync + 'static {
         partition: &str,
         name: &[u8],
     ) -> impl Future<Output = Result<(Self::Blob, u64), Error>> + Send {
-        let partition = partition.to_string();
-        let name = name.to_vec();
```
clabby, line 561: This was indeed a train wreck lmao. Nice to see you again 🫡

malik672: Same 🫡

PR #2794, opened 2026-01-12
[utils] Use `num-traits` for `FaultModel`

clabby:
## Overview

Adjusts `FaultModel` to use `num_traits::ToPrimitive` in the `quorum` and `max_faults` functions, so that either can accept any integer type.

The semantics of the API are slightly adjusted. The argument must both fit in a `u32` and be `> 0`.

### DIFF
--- a/cryptography/src/bls12381/dkg.rs
+++ b/cryptography/src/bls12381/dkg.rs
@@ -301,7 +301,7 @@ use commonware_math::{
 use commonware_parallel::{Sequential, Strategy as ParStrategy};
 use commonware_utils::{
     ordered::{Map, Quorum, Set},
-    Faults, Bft3f1, Participant, TryCollect, NZU32,
+    Bft3f1, Faults, Participant, TryCollect, NZU32,
 };
 use core::num::NonZeroU32;
 use rand_core::CryptoRngCore;


cryptography/src/bls12381/dkg.rs
```diff
@@ -441,7 +441,7 @@ where
         )
         .map_err(|_| arbitrary::Error::IncorrectFormat)?;
 
-        let max_revealed = commonware_utils::Bft3f1::max_faults(total as u32) as usize;
+        let max_revealed = commonware_utils::Bft3f1::max_faults(total) as usize;
```
patrick-ogrady: Just import `Bft3f1`?

utils/src/faults.rs
```diff
@@ -90,8 +106,12 @@ pub trait FaultModel {
...
 
 impl FaultModel for Bft3f1 {
-    fn max_faults(n: u32) -> u32 {
-        n.saturating_sub(1) / 3
+    fn max_faults(n: impl ToPrimitive) -> u32 {
+        let n = n
+            .to_u32()
+            .expect("n must be a non-negative integer that fits in u32");
+        assert!(n > 0, "n must not be zero");
+        (n - 1) / 3
```
patrick-ogrady, line 114: retain saturating?

utils/src/faults.rs
```diff
@@ -131,8 +151,12 @@ impl FaultModel for Bft3f1 {
...
 
 impl FaultModel for Bft5f1 {
-    fn max_faults(n: u32) -> u32 {
-        n.saturating_sub(1) / 5
+    fn max_faults(n: impl ToPrimitive) -> u32 {
+        let n = n
+            .to_u32()
+            .expect("n must be a non-negative integer that fits in u32");
+        assert!(n > 0, "n must not be zero");
+        (n - 1) / 5
```
patrick-ogrady, line 159: retain saturating

utils/src/faults.rs
clabby: See non-zero assertion above, shouldn't be necessary?

PR #2795, opened 2026-01-12
[ci] Reduce Disk Usage During `fuzz`

patrick-ogrady:


justfile
```diff
@@ -67,6 +67,7 @@ fuzz fuzz_dir max_time='60' max_mem='4000':
     #!/usr/bin/env bash
     for target in $(cargo {{nightly_version}} fuzz list --fuzz-dir {{fuzz_dir}}); do
         cargo {{nightly_version}} fuzz run $target --fuzz-dir {{fuzz_dir}} -- -max_total_time={{max_time}} -rss_limit_mb={{max_mem}}
+        rm -f {{fuzz_dir}}/target/*/release/$target
```
patrick-ogrady, line 70: This avoid deleting the dependency cache between runs in the same crate

patrick-ogrady:
(before merging, confirming this clears out right file)

cronokirby:
What does this file contain?

patrick-ogrady:
> What does this file contain?

A new binary is compiled for each fuzz target, so I suppose the answer to your question is "the completed fuzz test".

patrick-ogrady:
Storage is hitting this because I think it has the most (~30 or so)

PR #2799, opened 2026-01-12
[cryptography/bls12381] Final Nits

patrick-ogrady:


cryptography/src/bls12381/primitives/group.rs
```diff
@@ -78,23 +78,8 @@ pub struct Scalar(blst_fr);
 #[cfg(feature = "arbitrary")]
 impl arbitrary::Arbitrary<'_> for Scalar {
     fn arbitrary(u: &mut arbitrary::Unstructured<'_>) -> arbitrary::Result<Self> {
```
patrick-ogrady, line 80: This ensures conformance actually checks behavior was unmodified (using the "prod" code)

patrick-ogrady:
This indicates a conformance break but that isn't accurate (we just changed how we generate scalar in arbitrary).

cryptography/src/bls12381/primitives/ops/batch.rs
```diff
@@ -18,111 +18,157 @@ use super::{
...
     }
+
+    // Single chunk: skip root verification (caller already verified aggregate is invalid).
     let par_hint = strategy.parallelism_hint();
     let chunk_size = entries.len().div_ceil(par_hint);
+    if entries.len() <= chunk_size {
+        let mut out = SegmentTree::<V>::build(entries).verify(hm, true);
+        out.sort_unstable();
+        return out;
+    }
```
cronokirby, line 154: Why bother with handling this case separately, the general algorithm should work?

cronokirby: If the concern is that calling "fold" with a small input is inefficient, that should be addressed at the level of the parallelism API. One of the goals of that API is that it can absorb concerns like that, to avoid proliferation of little workarounds like this.

cryptography/src/bls12381/benches/batch_verify_same_signer.rs
```diff
@@ -32,6 +31,7 @@ fn benchmark_batch_verify_same_signer(c: &mut Criterion) {
                         },
                         |(public, entries)| {
                             if concurrency > 1 {
+                                let strategy = Rayon::new(NZUsize!(concurrency)).unwrap();
```
cronokirby: I think this measures the time it takes to create the thread pool at each iteration, which seems unrealistic, because in practice an application will amortize the cost over several invocations.

cryptography/src/bls12381/primitives/ops/batch.rs
patrick-ogrady: It was just some nuance about handling the root node being valid or not (not for fold overhead)

cronokirby: I think whether or not the root node is valid should depend on how `bisect` is called, not the relation between the number of entries and the chunk size? e.g. this should be a param to `bisect`, and in practice would be called with `true`, because we only bisect after knowing that the batch is bad overall.

PR #2800, opened 2026-01-12
[storage] Make QMDB `Send` friendly

roberto-bayardo:
Builds on https://github.com/commonwarexyz/monorepo/pull/2801 to fix compiler Send errors.

PR #2801, opened 2026-01-13
Keep qmdb Any futures Send

### LINKED ISSUE #2787: RPITIT async QMDB APIs yield !Send futures (rust-lang/rust#100013)
Summary

  - Authenticated QMDB async APIs (e.g. any::unordered::variable::Db::get) return RPITIT futures that borrow &self.
  - This triggers rust-lang/rust#100013 (“lifetime bound not satisfied”) even inside commonware-storage.
  - It also blocks using these APIs in Send-bound traits like commonware_consensus::Application/Reporter.

  Minimal repro

```rust
  use commonware_cryptography::{sha256::Digest, Sha256};
  use commonware_runtime::deterministic;
  use commonware_storage::{
      kv::Gettable as _,
      qmdb::any::unordered::variable::Db,
      translator::TwoCap,
  };

  type AnyDb = Db<deterministic::Context, Digest, Vec<u8>, Sha256, TwoCap>;

  // This fails to compile due to rust-lang/rust#100013.
  fn needs_send(db: &AnyDb, key: Digest) -> impl std::future::Future<Output = ()> + Send {
      async move {
          let _ = db.get(&key).await;
      }
  }

  fn main() {}
```

  Observed error

  error: lifetime bound not satisfied
    = note: this is a known limitation that will be removed in the future (see issue #100013)

  Why this matters

  - Consensus traits require impl Future + Send. As soon as authenticated QMDB APIs are used (e.g. preview_root in Application::propose/verify), compilation fails.
  - Callers are forced to add a Send boundary (actor, blocking executor) around QMDB usage.

  Possible direction

  - Provide a Send-safe wrapper in QMDB (actor/request queue) that exposes async methods returning Send futures.
  - Or expose non-RPITIT async APIs that do not capture &self.

diegomrsantos:
Closes #2787
  
  Issue #2787 reported that the futures returned by the Any helper methods were not `Send`. The fix now copies each translated-key location list before performing the awaited log work, so the async body captures only owned data and the future retains the Send property. A new deterministic regression test (`ordered_any_futures_are_send`) invokes every affected API (`get`, `get_all`, `get_with_loc`, `get_span`, `write_batch`, `update`, `create`, `delete`) and asserts the future each one returns is still `Send`.

### DIFF
--- a/storage/src/qmdb/any/ordered/mod.rs
+++ b/storage/src/qmdb/any/ordered/mod.rs
@@ -78,13 +78,13 @@ where
         }
     }
 
-    /// Finds and returns the location and KeyData for the lexicographically-last key produced by
+    /// Finds and returns the location and Update for the lexicographically-last key produced by
     /// the provided locations, skipping over locations that are beyond the log's range.
     async fn last_key_in_iter(
         &self,
         locs: impl IntoIterator<Item = Location>,
-    ) -> Result<LocatedKey<K, V::Value>, Error> {
-        let mut last_key: LocatedKey<K, V::Value> = None;
+    ) -> Result<LocatedKey<K, V>, Error> {
+        let mut last_key: LocatedKey<K, V> = None;
         for loc in locs {
             if loc >= self.op_count() {
                 // Don't try to look up operations that don't yet exist in the log. This can happen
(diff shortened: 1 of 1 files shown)


roberto-bayardo:
Thank you, I have rebased #2800 on this commit, will take it from there!

PR #2804, opened 2026-01-13
[storage] Make Storage Traits `Send` (via reference refactor and `use<>`)

patrick-ogrady:
Related: #2800
Related: #2801
Closes: https://github.com/commonwarexyz/monorepo/issues/2787

### DIFF
--- a/storage/src/kv/batch.rs
+++ b/storage/src/kv/batch.rs
@@ -49,7 +49,7 @@ where
     }
 }
 
-impl<'b, K, V, D> Gettable for Batch<'b, K, V, D>
+impl<'a, K, V, D> Gettable for Batch<'a, K, V, D>
 where
     K: Array,
     V: CodecShared + Clone,
@@ -70,7 +70,7 @@ where
     }
 }
 
-impl<'b, K, V, D> Updatable for Batch<'b, K, V, D>
+impl<'a, K, V, D> Updatable for Batch<'a, K, V, D>
 where
     K: Array,
     V: CodecShared + Clone,
@@ -84,7 +84,7 @@ where
     }
 }
 
-impl<'b, K, V, D> Deletable for Batch<'b, K, V, D>
+impl<'a, K, V, D> Deletable for Batch<'a, K, V, D>
 where
     K: Array,
     V: CodecShared + Clone,


storage/src/qmdb/any/unordered/variable/mod.rs
```diff
@@ -432,61 +434,45 @@ pub(super) mod test {
...
-    /// Helper to assert a future is Send at compile time.
     fn assert_send<T: Send>(_: T) {}
 
-    /// Test that futures returned by Durable Db methods are Send.
-    #[allow(dead_code)]
-    fn test_durable_futures_are_send(db: &mut AnyTest) {
+    /// Regression test for issue #2787: verifies that QMDB async methods can be used
+    /// in functions that return `impl Future + Send`.
+    #[allow(clippy::manual_async_fn)]
+    fn _issue_2787_repro(
```
patrick-ogrady: nit: better represent this?

storage/src/kv/batch.rs
```diff
@@ -49,7 +49,7 @@ where
     }
 }
 
-impl<'a, K, V, D> Gettable for Batch<'a, K, V, D>
+impl<'b, K, V, D> Gettable for Batch<'b, K, V, D>
```
andresilva: Any reason for this rename (and the one below)?

patrick-ogrady:
bugbot run

storage/src/kv/batch.rs
patrick-ogrady:
```
  2. 'a → 'b Lifetime Rename - Avoiding Shadowing

  The rename is necessary to avoid lifetime parameter shadowing.

  The struct is defined as:
  pub struct Batch<'a, K, V, D> { db: &'a D, ... }

  The trait method introduces its own 'a:
  fn get<'a>(&'a self, key: &'a Self::Key)
      -> impl Future<...> + Send + use<'a, Self>;

  If the impl used 'a:
  impl<'a, K, V, D> Gettable for Batch<'a, K, V, D> {
      fn get<'a>(...) // This 'a shadows the impl's 'a!
  }

  The method's 'a would shadow the struct's 'a, causing confusion about which lifetime use<'a, Self> refers to.

  By using 'b for the struct lifetime:
  impl<'b, K, V, D> Gettable for Batch<'b, K, V, D> {
      fn get<'a>(&'a self, ...) -> ... + use<'a, Self>
      //    ^^ method lifetime    ^^ captures method's 'a and Self (which includes 'b)
  }

  There's no ambiguity - 'a is the method's borrow lifetime, 'b is the struct's database reference lifetime, and Self = Batch<'b, K, V, D>.
  ```

patrick-ogrady: i.e. nuance of `use<>`

andresilva: I think the LLM is being stupid here 😬 there's no reference to `'a` in this context, this compiles just fine if you rename `'b` back to `'a`.

patrick-ogrady: Yeah, you rite.

patrick-ogrady:
I think we may be able to make these bounds more minimal and/or remove a bunch of the inherents (just use trait directly), however, this works and I'd prefer to minimize after-the-fact.

PR #2807, opened 2026-01-13
[storage/journal] Unify Fixed Journals Nits

patrick-ogrady:
Related: #2774

### DIFF
--- a/storage/src/journal/contiguous/fixed.rs
+++ b/storage/src/journal/contiguous/fixed.rs
@@ -44,11 +44,11 @@
 //!
 //! # State Sync
 //!
-//! `Journal::init_sync` allows for initializing a journal for use in state sync. When opened in
-//! this mode, we attempt to populate the journal within the given range with persisted data. If the
-//! journal is empty, we create a fresh journal at the specified position. If the journal is not
-//! empty, we prune the journal to the specified lower bound and rewind to the specified upper
-//! bound.
+//! `Journal::init_sync` initializes a journal for state sync, handling existing data appropriately:
+//! - If no data exists, creates a journal at the sync range start
+//! - If data exists within range, prunes to the lower bound
+//! - If data exceeds the range, returns an error
+//! - If data is stale (before range), destroys and recreates
 //!
 //! # Replay
 //!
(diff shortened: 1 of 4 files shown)


storage/src/qmdb/any/unordered/fixed/sync.rs
```diff
@@ -402,10 +314,13 @@ mod tests {
             // Initialize journal with sync boundaries when no existing data exists
             let lower_bound = 10;
             let upper_bound = 26;
-            let mut sync_journal =
-                init_journal(context.clone(), cfg.clone(), lower_bound..upper_bound)
-                    .await
-                    .expect("Failed to initialize journal with sync boundaries");
+            let mut sync_journal = fixed::Journal::<_, Digest>::init_sync(
```
patrick-ogrady, line 317: We should probably move these tests to journal?

storage/src/journal/contiguous/fixed.rs
```diff
@@ -794,36 +894,55 @@ mod tests {
...
                 .await
                 .expect("Failed to open blob");
             blob.resize(size - 1).await.expect("Failed to corrupt blob");
             blob.sync().await.expect("Failed to sync blob");
 
-            // The segmented journal will trim the incomplete blob on init, resulting in the blob
-            // missing one item. This should be detected as corruption during replay.
-            let journal = Journal::<_, Digest>::init(context.clone(), cfg.clone())
+            // Init should fail due to corrupted historical blob (not full)
+            match Journal::<_, Digest>::init(context.clone(), cfg.clone()).await {
```
roberto-bayardo, line 909: This means we're not testing the replay corruption detection.

storage/src/journal/contiguous/fixed.rs
```diff
@@ -190,25 +195,120 @@ impl<E: Storage + Metrics, A: CodecFixedShared> Journal<E, A> {
...
+                    return Err(Error::MissingBlob(expected));
+                }
+                // Verify non-tail section is full (has exactly items_per_blob items)
+                let len = inner.section_len(prev).await?;
+                if len != items_per_blob {
+                    return Err(Error::InvalidBlobSize(prev, len * Self::CHUNK_SIZE_U64));
+                }
+                prev = section;
+            }
+        }
```
roberto-bayardo, line 310: I got rid of this because I added corruption checking to replay() which detects any of these cases.  Do we want both?

I thought putting it in replay was strictly better since you typically replay the stuff you care about on startup, so if an earlier blob is corrupted, doesn't really matter. And, if there's corruption to some relevant blob *after* this point and you call replay again,  this would not have previously been detected.

roberto-bayardo:
merged into #2774

roberto-bayardo:
Merged this into my PR, but removed the integrity checking from init() for reasons outlined in my comment above.

PR #2810, opened 2026-01-13
[runtime/network/iouring] support vectored writes

### LINKED ISSUE #2705: [runtime] io_uring network vectored writes (writev)
## Summary

The io_uring network implementation currently copies buffers to contiguous memory before writing. We should use `writev` to support vectored writes and avoid this copy.

## Current Behavior

When sending with io_uring, we copy the `impl Buf` to a contiguous `StableBuf`:

```rust
// runtime/src/network/iouring.rs
async fn send(&mut self, mut msg: impl Buf + Send) -> Result<(), crate::Error> {
    // For now, collect the message into a stable buffer. In the future,
    // we could optimize this by using writev w/ a keepalive for the iovecs.
    let mut msg: StableBuf = {
        let buf = msg.copy_to_bytes(msg.remaining());
        BytesMut::from(buf).into()
    };
    // ...
}
```

## Proposed Change

Use io_uring's `writev` operation to write multiple buffer chunks in a single syscall without concatenation. This requires:

1. Iterating over `Buf::chunk()` to get contiguous segments
2. Building iovec array for the segments
3. Keeping the source buffers alive until the write completes

## Related

- #784 - vectored writes for tokio (addressed)
- #1434 - use `Buf` trait instead of `Bytes` (enables this optimization)
- #2554 - `impl Buf` for storage blob writes

andresilva:
Closes #2705.

### DIFF
--- a/runtime/src/iouring/mod.rs
+++ b/runtime/src/iouring/mod.rs
@@ -77,7 +77,7 @@ use std::{collections::HashMap, sync::Arc, time::Duration};
 /// Returns the maximum number of iovecs per sendmsg call.
 /// Queries the kernel via sysconf(_SC_IOV_MAX), falling back to 1024 (Linux default).
 #[cfg(feature = "iouring-network")]
-fn iov_max() -> usize {
+pub fn iov_max() -> usize {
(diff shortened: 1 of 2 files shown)


andresilva:
Replaced by #3273.

PR #2814, opened 2026-01-13
[runtime] use Buf / BufMut in Blob APIs

### LINKED ISSUE #2554: [storage] Take Advantage of `impl Buf` on Runtime Blob
We can avoid a lot of intermediate copying once #784 is in.

andresilva:
Closes #2554.


andresilva:
Replaced by #2881.

PR #2817, opened 2026-01-13
[consensus/application] Async Verify Nits

patrick-ogrady:
Related: #2604

### DIFF
--- a/consensus/src/marshal/mod.rs
+++ b/consensus/src/marshal/mod.rs
@@ -1651,7 +1651,9 @@ mod tests {
                 .verify(byzantine_context, malicious_commitment)
                 .await
                 .await;
-            let verify = marshaled.certify(byzantine_round, malicious_commitment).await;
+            let verify = marshaled
+                .certify(byzantine_round, malicious_commitment)
+                .await;
 
             assert!(
                 !verify.await.unwrap(),
@@ -1698,7 +1700,9 @@ mod tests {
                 .verify(byzantine_context, malicious_commitment)
                 .await
                 .await;
-            let verify = marshaled.certify(byzantine_round, malicious_commitment).await;
+            let verify = marshaled
+                .certify(byzantine_round, malicious_commitment)
+                .await;
 
             assert!(
                 !verify.await.unwrap(),


consensus/src/application/marshaled.rs
```diff
@@ -196,24 +197,6 @@ where
                 let tx_closed = tx.closed();
                 pin_mut!(tx_closed);
 
-                // Helper to clean up verification context for this round
```
patrick-ogrady, line 199: Ah, I now understand what this was doing (if we attempt to build on a notarization it assumes its been certified)

patrick-ogrady: Is it safe to remove all certifications less than this (if there are forks tho)?

consensus/src/marshal/actor.rs
```diff
@@ -918,9 +926,13 @@ where
             panic!("failed to finalize: {e}");
         }
 
-        // Update metrics and send tip update to application
+        // Update metrics and send tip update to application.
+        // The tip must have a direct finalization (see get_latest() comment).
```
patrick-ogrady: nit: remove this line

consensus/src/marshal/actor.rs
```diff
@@ -918,9 +926,13 @@ where
...
         if height > self.tip {
```
patrick-ogrady: a safer route may be to only enter if `round` is not none?

consensus/src/application/marshaled.rs
```diff
@@ -539,7 +537,14 @@ where
         context: Context<Self::Digest, S::PublicKey>,
         digest: Self::Digest,
     ) -> oneshot::Receiver<bool> {
-        Self::store_verification_context(&self.verification_contexts, context, digest).await;
+        let finalized_round = self.last_finalized_round.lock().await.take();
```
patrick-ogrady: Why not just pass the Arc<Mutex<>> in?

consensus/src/application/marshaled.rs
patrick-ogrady: Yes, it's safe. Once a block at round R is finalized, all blocks on competing forks at rounds <= R are dead and will never be certified. We only need verification contexts for blocks that might still be certified, and those are all at rounds > the finalized round.

consensus/src/application/marshaled.rs
```diff
@@ -106,6 +106,7 @@ where
     last_built: Arc<Mutex<Option<(Round, B)>>>,
     verification_contexts: Arc<Mutex<VerificationContexts<E, B, S>>>,
     certification_txs: Arc<Mutex<Vec<oneshot::Sender<bool>>>>,
```
patrick-ogrady: I don't think we actually need to store these (just dropping if we don't have the certification should be sufficient)?

consensus/src/application/marshaled.rs
```diff
@@ -560,35 +534,26 @@ where
...
-            drop(txs_guard);
-
-            // If we don't have a verification context, we cannot verify the block.
-            // In this event, we return a receiver that will never resolve to signal
-            // to consensus that we should time out the view and vote to nullify.
+            // Verify is always called before certify for a given proposal, so if we
+            // don't have a verification context here, it means this proposal was never
+            // verified. Return a receiver that never resolves to signal to consensus
+            // that we should time out the view and vote to nullify.
+            let (_tx, rx) = oneshot::channel();
```
clabby, line 556: This won't hang like the previous code did, since we drop `tx` right away. https://play.rust-lang.org/?version=stable&mode=release&edition=2024&gist=e3cd9e2cf8a70334c9ee27740fa699ba

You'll end up entering this branch in the voter:
https://github.com/commonwarexyz/monorepo/blob/3375949367fd2fc64be4a9a635efc5e3be709b0e/consensus/src/simplex/actors/voter/actor.rs#L935-L943

rather than hitting the timeout:
https://github.com/commonwarexyz/monorepo/blob/3375949367fd2fc64be4a9a635efc5e3be709b0e/consensus/src/simplex/actors/voter/actor.rs#L851-L855

clabby: (Is this intentional? If so, docs here should be updated.)

patrick-ogrady: Yes, intentional

consensus/src/application/marshaled.rs
```diff
@@ -560,35 +534,26 @@ where
...
-    async fn certify(&mut self, payload: Self::Digest) -> oneshot::Receiver<bool> {
-        // Look up context by block digest and pop the oldest (by round) context from the map.
-        // Using pop_first ensures each concurrent certification for the same digest gets a
-        // unique context, which is critical when a block is re-proposed across multiple views.
-        //
-        // This works because the voter certifies views in ascending order: when the same block
-        // appears in multiple views (e.g., re-proposals at epoch boundaries), certify is called
-        // for the lower view first.
-        let mut contexts_guard = self.verification_contexts.lock().await;
+    async fn certify(&mut self, round: Round, payload: Self::Digest) -> oneshot::Receiver<bool> {
```
patrick-ogrady, line 537: Do we need both of these? I feel like `round` ~should be sufficient?

patrick-ogrady: Ah, if you don't get verify you'll still want payload ...

consensus/src/marshal/mod.rs
```diff
@@ -2031,6 +2041,161 @@ mod tests {
         })
     }
 
+    /// Regression test for cleanup bug in #2604.
```
patrick-ogrady, line 2044: nit: remove PR specific statements, just include details of edge case

consensus/src/application/marshaled.rs
patrick-ogrady: We should add a comment explaining as much in the consensus trait

patrick-ogrady: If we are taking the view that blocks may not contain any consensus data, this is a must.

If we instead assume the digest changes because it contains `Round`, then this is a no-op.

PR #2820, opened 2026-01-14
[consensus/simplex] Test for `nullify` on cancelled `certify` response channel

clabby:
## Overview

Adds a test to the voter actor that ensures a `nullify` vote is cast, even if a `certify` call returns a channel that yields `Err(Canceled)` .

### DIFF
--- a/consensus/src/simplex/actors/voter/mod.rs
+++ b/consensus/src/simplex/actors/voter/mod.rs
@@ -3264,10 +3264,22 @@ mod tests {
 
     #[test_collect_traces]
     fn test_cancelled_certification_does_not_hang(traces: TraceStorage) {
-        cancelled_certification_does_not_hang(bls12381_threshold::fixture::<MinPk, _>, traces.clone());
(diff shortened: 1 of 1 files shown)


consensus/src/simplex/actors/voter/mod.rs
```diff
@@ -3138,4 +3141,191 @@ mod tests {
...
+                    _ = context.sleep(Duration::from_secs(5)) => {
+                        panic!(
+                            "voter should emit nullify for view {target_view} despite cancelled certification",
+                        );
+                    },
+                }
+            }
+
+            // Verify the "failed to certify proposal" log was emitted with the correct round
+            let expected_round = format!("Round {{ epoch: Epoch(333), view: View({target_view}) }}");
```
patrick-ogrady, line 3245: :fancy:

patrick-ogrady: I generally think relying on logs in tests can be flaky but this is cool (and a very critical part of the code where we'd want to know if we changed logs)

Issue #2823, opened 2026-01-14
QMDB cursor insert panics on translator collisions


diegomrsantos:
QMDB snapshot update/create panics when a translated-key collision occurs and the key is not found. The cursor is used for insert without being advanced, triggering `must call Cursor::next()`

PR #2824, opened 2026-01-14
fix(storage): handle qmdb cursor insert after collisions

diegomrsantos:
fixes https://github.com/commonwarexyz/monorepo/issues/2823

## Summary
- call `cursor.next()` before insert when adding a new key after translated-key collisions
- add regression test for collision updates in QMDB unordered fixed

## Testing
- cargo test -p commonware-storage test_update_handles_translator_collisions

### DIFF
--- a/storage/src/qmdb/mod.rs
+++ b/storage/src/qmdb/mod.rs
@@ -233,6 +233,7 @@ where
     }
 
     // The key wasn't in the snapshot, so add it to the cursor.
+    cursor.next();
     cursor.insert(new_loc);
 
     Ok(None)
@@ -262,6 +263,7 @@ where
     }
 
     // The key doesn't exist, so add it to the cursor.
+    cursor.next();
     cursor.insert(new_loc);
 
     Ok(true)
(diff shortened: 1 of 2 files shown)


storage/src/qmdb/mod.rs
```diff
@@ -233,6 +233,7 @@ where
     }
 
     // The key wasn't in the snapshot, so add it to the cursor.
+    cursor.next();
```
roberto-bayardo, line 236: Interesting. Calling next() here should not be necessary. What happens without it?

roberto-bayardo: I commented  out these calls and your test still seems to pass?

diegomrsantos: You're right. I was trying to reproduce a stack overflow issue that I was getting with qmdb, and then I got "internal error: entered unreachable code: must call Cursor::next()". I tried to create a smaller test to reproduce only the panic, but I see it doesn't work. I'll try to reproduce the panic again. Sorry for the confusion.

roberto-bayardo: no problem, glad you're exercising the code.  closing this one out.

diegomrsantos: I was running this in a problematic code in my branch https://github.com/commonwarexyz/monorepo/pull/2495/changes#diff-74753da578bf4633361a97dde454eaad726d3623ab9b69016177b707324c7f7f

PR #2826, opened 2026-01-14
[cryptography] 2701, document RecvCipher::recv caveats

### LINKED ISSUE #2701: [cryptography/handshake] Nonce desync DoS in RecvCipher::recv
### Summary

`RecvCipher::recv` consumes the next nonce before attempting AEAD decryption. On any decryption failure, the receiver's counter is advanced while the sender remains on the previous value, permanently desynchronizing the stream (or at minimum making a single bad frame a reliable session-kill).

### Location

- `cryptography/src/handshake/cipher.rs`
  - `CounterNonce::inc` increments before returning bytes
  - `RecvCipher::recv` calls `self.nonce.inc()` inside `ChaCha20Poly1305::decrypt(...)`
- Used by `stream/src/lib.rs::Receiver::recv`

### Details

Today, `CounterNonce::inc` works as:

```rust
pub const fn inc(&mut self) -> Result<[u8; 128 / 8], Error> {
    if self.inner >= 1 << (8 * NONCE_SIZE_BYTES) {
        return Err(Error::MessageLimitReached);
    }
    let out = self.inner.to_le_bytes();
    self.inner += 1;
    Ok(out)
}
```

`RecvCipher::recv` then does:

```rust
pub fn recv(&mut self, encrypted_data: &[u8]) -> Result<Vec<u8>, Error> {
    self.inner
        .decrypt(
            (&self.nonce.inc()?[..NONCE_SIZE_BYTES]).into(),
            encrypted_data,
        )
        .map_err(|_| Error::DecryptionFailed)
}
```

If an attacker injects a single invalid ciphertext on a live stream:

1. `RecvCipher::recv` calls `self.nonce.inc()`, advancing the local counter.
2. AEAD authentication fails and returns `Error::DecryptionFailed`.
3. The sender's `SendCipher` is still at the old counter value; on the next frame it will encrypt with nonce `N+1`, while the receiver will decrypt with `N+2`.

Unless the caller treats *any* decrypt failure as fatal and tears down the connection immediately, the cipher state is now permanently desynchronized and all future frames will fail. Even when callers do tear down on error, this still gives a trivial "kill the connection with one bad frame" lever to any malicious or on-path peer.

### Impact

- **Reliable DoS / channel kill**: A malicious peer or man-in-the-middle can break a stream with a single bogus frame.
- **Permanent desync**: If callers keep using the same `RecvCipher` after a `DecryptionFailed`, the receive direction is wedged for the lifetime of the session.
- This affects the record layer used by `stream::Receiver`, so it can take down higher-level protocols that rely on the stream abstraction.

### Suggested fix

Avoid advancing the counter before we know the ciphertext is valid:

- Split `CounterNonce` into `current()` and `advance()` (or equivalent), so callers can:
  1. derive a nonce from the current counter value,
  2. attempt `decrypt(nonce, ciphertext)`,
  3. **only on success** call `advance()`.

Conceptually:

```rust
impl CounterNonce {
    pub fn current(&self) -> [u8; 128/8] { self.inner.to_le_bytes() }
    pub fn advance(&mut self) -> Result<(), Error> { /* overflow check + self.inner += 1 */ }
}

pub fn recv(&mut self, encrypted_data: &[u8]) -> Result<Vec<u8>, Error> {
    let out = self.nonce.current();
    let nonce = (&out[..NONCE_SIZE_BYTES]).into();

    let pt = self.inner
        .decrypt(nonce, encrypted_data)
        .map_err(|_| Error::DecryptionFailed)?;

    self.nonce.advance()?;
    Ok(pt)
}
```

Even if the application chooses to close the connection on any decrypt failure, this change prevents accidental state desynchronization before that decision is made.

### Notes

- `SendCipher::send` is already safe in the sense that advancing the send nonce on failure cannot cause future valid frames to be misinterpreted; it just caps the number of usable messages.
- The critical requirement is that the *receive* counter must only advance for ciphertexts that actually authenticated and decrypted successfully.

cronokirby:
Closes #2701.

Documents the behavior of `RecvCipher::recv` after a ciphertext fails to authenticate.

### DIFF
--- a/cryptography/src/handshake/cipher.rs
+++ b/cryptography/src/handshake/cipher.rs
@@ -80,15 +80,15 @@ impl RecvCipher {
     ///
     /// # Errors
     ///
-    /// This will return an error in the following situations:
+    /// This function will return an error in the following situations:
     ///
     /// - Too many messages have been received with this cipher.
     /// - The ciphertext was corrupted in some way.
     ///
     /// In *both* cases, the `RecvCipher` will no longer be able to return
     /// valid ciphertexts, and will always return an error on subsequent calls
-    /// to [`Self::recv`]. Terminating, and potentially reëstablishing,
-    /// the connection is a simple way to handle this case.
+    /// to [`Self::recv`]. Terminating (and optionally reestablishing) the connection
+    /// is a simple (and safe) way to handle this scenario.
     pub fn recv(&mut self, encrypted_data: &[u8]) -> Result<Vec<u8>, Error> {
         let nonce = self.nonce.inc()?;
         self.inner


cryptography/src/handshake/cipher.rs
```diff
@@ -77,6 +77,18 @@ impl RecvCipher {
...
+    /// # Errors
+    ///
+    /// This will return an error in the following situations:
+    ///
+    /// - Too many messages have been received with this cipher.
+    /// - The ciphertext was corrupted in some way.
+    ///
+    /// In *both* cases, the `RecvCipher` will no longer be able to return
+    /// valid ciphertexts, and will always return an error on subsequent calls
+    /// to [`Self::recv`]. Terminating, and potentially reëstablishing,
```
patrick-ogrady: that's a new way to spell re-establishing 👀

patrick-ogrady (approved):

PR #2827, opened 2026-01-14
[consensus/marshal] Add pruning API

### LINKED ISSUE #2825: [consensus/marshal] Provide API to prune finalized stores
## Overview

We should provide an API to prune the finalized blocks and certificates store in marshal. This API should only allow pruning these stores below the floor (latest processed height.)

clabby:
## Overview

Adds a new `prune` function to marshal's mailbox, allowing pruning of the finalized blocks / certificates store below the latest processed height without affecting the sync floor.

closes #2825

### DIFF
--- a/consensus/src/marshal/mod.rs
+++ b/consensus/src/marshal/mod.rs
@@ -766,109 +766,124 @@ mod tests {
             } = bls12381_threshold::fixture::<V, _>(&mut context, NAMESPACE, NUM_VALIDATORS);
 
             let validator = participants[0].clone();
-            let provider = ConstantProvider::new(schemes[0].clone());
-
-            // Create config
-            let config = Config {
-                provider,
-                epocher: FixedEpocher::new(BLOCKS_PER_EPOCH),
-                mailbox_size: 100,
-                view_retention_timeout: ViewDelta::new(10),
-                max_repair: NZUsize!(10),
-                block_codec_config: (),
-                partition_prefix: format!("prune-test-{}", validator.clone()),
-                prunable_items_per_section: NZU64!(10),
-                replay_buffer: NZUsize!(1024),
-                key_write_buffer: NZUsize!(1024),
-                value_write_buffer: NZUsize!(1024),
-                buffer_pool: PoolRef::new(PAGE_SIZE, PAGE_CACHE_SIZE),
-                strategy: Sequential,
-            };
-
-            // Create the resolver
+            let partition_prefix = format!("prune-test-{}", validator.clone());
+            let buffer_pool = PoolRef::new(PAGE_SIZE, PAGE_CACHE_SIZE);
             let control = oracle.control(validator.clone());
-            let backfill = control.register(1, TEST_QUOTA).await.unwrap();
-            let resolver_cfg = resolver::Config {
-                public_key: validator.clone(),
-                manager: oracle.manager(),
-                blocker: control.clone(),
-                mailbox_size: config.mailbox_size,
-                initial: Duration::from_secs(1),
-                timeout: Duration::from_secs(2),
-                fetch_retry_timeout: Duration::from_millis(100),
-                priority_requests: false,
-                priority_responses: false,
+
+            // Closure to initialize marshal with prunable archives
+            let init_marshal = || {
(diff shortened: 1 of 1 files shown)


consensus/src/marshal/actor.rs
```diff
@@ -525,21 +525,20 @@ where
...
-                                }
-                            ) {
+                            if let Err(err) = self.prune_finalized_archives(height).await {
+                                error!(?err, %height, "failed to prune finalized archives");
+                                return;
+                            }
+                        }
+                        Message::Prune { height } => {
+                            // Only allow pruning at or below the current floor
+                            if height > self.last_processed_height {
```
patrick-ogrady, line 537: Should we use `last_processed_height` in set_floor above?

why `>` here but `>=` above?

clabby: `>=` above since there are other side-effects - If you call it with the current floor, you don't want to clear the pending ack (even though pruning there is valid,) since you're not actually updating anything.

`>` here since pruning all items below the current floor is valid.

consensus/src/marshal/ingress/mailbox.rs
```diff
@@ -126,17 +126,26 @@ pub(crate) enum Message<S: Scheme, B: Block> {
...
         /// The candidate floor height.
         height: Height,
     },
+    /// Prunes finalized blocks and certificates below the given height.
+    ///
+    /// Unlike [Message::SetFloor], this does not affect the sync starting point.
+    /// The height must be at or below the current floor (last processed height),
+    /// otherwise the prune request is ignored.
+    Prune {
+        /// The minimum height to keep (blocks below this are pruned).
```
patrick-ogrady, line 146: Can you add this nuance to the public `fn`?

consensus/src/marshal/mod.rs
```diff
@@ -749,6 +749,270 @@ mod tests {
...
+                },
+            )
+            .await
+            .expect("failed to initialize finalizations by height archive");
+
+            // Initialize finalized blocks using PRUNABLE archive
+            let finalized_blocks = prunable::Archive::init(
+                context.with_label("finalized_blocks"),
+                prunable::Config {
+                    translator: DigestTranslator,
```
patrick-ogrady: I think we can just use our standard Two/Four/EightCap here? Should reduce diff.

consensus/src/marshal/ingress/mailbox.rs
clabby: Should already be there:

https://github.com/commonwarexyz/monorepo/blob/c6fa30214ecaacf8eee4ead0871e8d7c8905016c/consensus/src/marshal/ingress/mailbox.rs#L305

consensus/src/marshal/mod.rs
clabby: https://github.com/commonwarexyz/monorepo/pull/2827/changes/c6fa30214ecaacf8eee4ead0871e8d7c8905016c

consensus/src/marshal/mod.rs
```diff
@@ -749,6 +752,225 @@ mod tests {
...
+                context.clone(),
+                finalizations_by_height,
+                finalized_blocks,
+                config,
+            )
+            .await;
+            let application = Application::<B>::default();
+
+            // Start the application
+            actor.start(application.clone(), buffer, resolver);
```
patrick-ogrady: Would it be overkill to add a restart at the end of this test (just to ensure we don't introduce some sort of regression where memory makes us think the data has been pruned correctly, when it hasn't -> either deleting too much or too little)?

clabby: Added

consensus/src/marshal/mod.rs
```diff
@@ -968,6 +983,32 @@ mod tests {
                 mailbox.get_finalization(Height::new(20)).await.is_some(),
                 "finalization 20 should still exist"
             );
+
+            // Restart to verify pruning persisted to storage (not just in-memory)
+            drop(mailbox);
+            let (mut mailbox, _application) = init_marshal().await;
```
patrick-ogrady, line 989: nice approach

PR #2833, opened 2026-01-15
Defer deserializing signatures

### LINKED ISSUE #2732: [cryptography] Consider adding BLS operations which can accept invalid points
Checking that a point is valid involves a scalar multiplication. In some situations, you can reduce the total amount of work you would need to do if you checked each point individually, and then did some operation with the points. For example, when batch verifying signatures, you could accept invalid points, and then combine the scalar multiplication for the subgroup check with the weight you use for batch verification.

cronokirby:
Closes #2732.

This introduces a new type in codec, `Lazy`, which doesn't immediately deserialize its value, instead storing the bytes. This allows, among other use cases, deferring the deserialization of signatures until they actually need to be verified. Batch verification can do the deserialization in parallel.

### DIFF
--- a/consensus/src/simplex/scheme/bls12381_threshold/vrf.rs
+++ b/consensus/src/simplex/scheme/bls12381_threshold/vrf.rs
@@ -638,7 +638,7 @@ impl<P: PublicKey, V: Variant> certificate::Scheme for Scheme<P, V> {
     {
         let namespace = self.namespace();
         let (partials, failures) =
-            strategy.map_filter_collect_vec(attestations.into_iter(), |attestation| {
+            strategy.map_partition_collect_vec(attestations.into_iter(), |attestation| {
                 let index = attestation.signer;
                 let value = attestation.signature.get().map(|sig| {
                     (
@@ -726,7 +726,7 @@ impl<P: PublicKey, V: Variant> certificate::Scheme for Scheme<P, V> {
         M: Faults,
     {
         let (partials, failures) =
-            strategy.map_filter_collect_vec(attestations.into_iter(), |attestation| {
+            strategy.map_partition_collect_vec(attestations.into_iter(), |attestation| {
                 let index = attestation.signer;
                 let value = attestation.signature.get().map(|sig| {
                     (
--- a/cryptography/src/bls12381/certificate/multisig/mod.rs
+++ b/cryptography/src/bls12381/certificate/multisig/mod.rs
@@ -160,7 +160,7 @@ impl<P: PublicKey, V: Variant, N: Namespace> Generic<P, V, N> {
         T: Strategy,
     {
         let (filtered, failures) =
-            strategy.map_filter_collect_vec(attestations.into_iter(), |attestation| {
+            strategy.map_partition_collect_vec(attestations.into_iter(), |attestation| {
                 let signer = attestation.signer;
                 let value = self
                     .participants
--- a/cryptography/src/bls12381/certificate/threshold/mod.rs
+++ b/cryptography/src/bls12381/certificate/threshold/mod.rs
@@ -268,7 +268,7 @@ impl<P: PublicKey, V: Variant, N: Namespace> Generic<P, V, N> {
         T: Strategy,
     {
         let (partials, failures) =
-            strategy.map_filter_collect_vec(attestations.into_iter(), |attestation| {
+            strategy.map_partition_collect_vec(attestations.into_iter(), |attestation| {
                 let index = attestation.signer;
                 let partial = attestation
                     .signature
@@ -313,7 +313,7 @@ impl<P: PublicKey, V: Variant, N: Namespace> Generic<P, V, N> {
         M: Faults,
     {
         let (partials, failures) =
-            strategy.map_filter_collect_vec(attestations.into_iter(), |attestation| {
+            strategy.map_partition_collect_vec(attestations.into_iter(), |attestation| {
                 let index = attestation.signer;
                 let value = attestation
                     .signature
--- a/parallel/src/lib.rs
+++ b/parallel/src/lib.rs
@@ -16,7 +16,7 @@
 //! - [`map_collect_vec`](Strategy::map_collect_vec): Maps elements and collects into a `Vec`
 //! - [`map_init_collect_vec`](Strategy::map_init_collect_vec): Like `map_collect_vec` with
 //!   per-partition initialization
-//! - [`map_filter_collect_vec`](Strategy::map_filter_collect_vec): Maps elements, collecting
+//! - [`map_partition_collect_vec`](Strategy::map_partition_collect_vec): Maps elements, collecting
 //!   successful results and tracking indices of filtered elements
 //!
 //! Two implementations are provided:
@@ -309,7 +309,7 @@ pub trait Strategy: Clone + Send + Sync + fmt::Debug + 'static {
     /// let strategy = Sequential;
     /// let data = vec![1, 2, 3, 4, 5];
     ///
-    /// let (evens, odd_values): (Vec<i32>, Vec<i32>) = strategy.map_filter_collect_vec(
+    /// let (evens, odd_values): (Vec<i32>, Vec<i32>) = strategy.map_partition_collect_vec(
     ///     data.iter(),
     ///     |&x| (x, if x % 2 == 0 { Some(x * 10) } else { None }),
     /// );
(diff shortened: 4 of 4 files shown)


codec/src/types/lazy.rs
```diff
@@ -0,0 +1,100 @@
...
+
+#[derive(Clone)]
+pub struct Lazy<T: Read> {
+    bytes: Bytes,
+    cfg: Option<T::Cfg>,
+    value: OnceLock<Option<T>>,
+}
+
+impl<T: Read> Lazy<T> {
+    pub fn new(buf: &mut impl Buf, cfg: T::Cfg) -> Self {
```
patrick-ogrady: As discussed, may unblock things if we accept a fixed codec here only.

codec/src/types/lazy.rs
```diff
@@ -0,0 +1,284 @@
+//! This module exports the [`Lazy`] type.
```
patrick-ogrady, line 1: nit: add \n

codec/src/types/lazy.rs
```diff
@@ -0,0 +1,284 @@
...
+            .expect("Lazy should have a value if proto is None")
+            .write(buf);
+    }
+}
+
+impl<T: Read + FixedSize> Read for Lazy<T> {
+    type Cfg = T::Cfg;
+
+    fn read_cfg(buf: &mut impl Buf, cfg: &Self::Cfg) -> Result<Self, crate::Error> {
+        Ok(Self::deferred(&mut buf.take(T::SIZE), cfg.clone()))
```
patrick-ogrady, line 186: I think we could skip deferred here:
```
 let bytes = buf.copy_to_bytes(T::SIZE);
 Ok(Self {
     proto: Some(Proto { bytes, cfg: cfg.clone() }),
     value: Default::default(),
 })
 ```
 
 I don't think it should be any more overhead to take but 🤷

**Review by patrick-ogrady — COMMENTED** [review]
very minimal nits

File: codec/src/types/lazy.rs [hunk 4]
```diff
@@ -0,0 +1,284 @@
...
+/// of `T` as well. These methods will force deserialization of the value.
+#[derive(Clone)]
+pub struct Lazy<T: Read> {
+    /// This should only be `None` if `value` is initialized.
+    proto: Option<Proto<T>>,
+    value: OnceLock<Option<T>>,
+}
+
+#[derive(Clone)]
+struct Proto<T: Read> {
```
patrick-ogrady: nit: `Pending`

codec/src/types/lazy.rs
cronokirby: > ⏺ The key differences:
> 
>   1. Stack frame: 32 → 112 bytes (+80 bytes)
> 
>   2. The Take wrapper adds min calculation:
>   ```
>   ; via_deferred:
>   ldr x1, [x0, #16]           ; x1 = remaining (direct)
> 
>   ; via_read:                                                                                                                                   ldr x8, [x0, #16]           ; load remaining
>   mov w9, #32                 ; load SIZE constant
>   cmp x8, #32                 ; compare
>   csel x1, x8, x9, lo         ; x1 = min(remaining, 32)
>   mov x8, sp                  ; result goes to stack
>   ```
> 
>   3. Extra struct copying (7 memory ops):
>   ```
>   ldp q1, q0, [sp]            ; load from stack
>   stur q0, [x20, #24]         ; copy to return slot
>   ldp q2, q0, [sp, #32]
>   stur q0, [x20, #56]
>   ldr q0, [sp, #64]
>   stur q0, [x20, #72]
>   stur q1, [x20, #8]
>   stur q2, [x20, #40]
>   str xzr, [x20]
>   ```
> 
>   4. Lost tail call: br x3 → blr x8 + ret
> 
>   ---
>   Realistic overhead roughly 3-5 nanoseconds per call.

The struct has non-trivial details, so I think the very minimal overhead here is not worth adding an extra construction point

consensus/src/simplex/scheme/bls12381_threshold/vrf.rs
```diff
@@ -599,17 +607,25 @@ impl<P: PublicKey, V: Variant> certificate::Scheme for Scheme<P, V> {
             },
         );
         // Merge invalid sets
-        let invalid: BTreeSet<_> = vote_invalid.union(&seed_invalid).copied().collect();
+        let mut invalid: BTreeSet<_> = vote_invalid.union(&seed_invalid).copied().collect();
 
-        let verified = vote_partials
+        for (signer, vote, seed) in partials.iter() {
```
patrick-ogrady: nit: add a comment explaining what we are doing here

codec/src/types/lazy.rs
```diff
@@ -0,0 +1,306 @@
...
+///
+/// [`Lazy`] can be serialized and deserialized, implementing [`Read`], [`Write`],
+/// and [`EncodeSize`], based on the underlying implementation of `T`.
+///
+/// Furthermore, we implement [`Eq`], [`Ord`], [`Hash`] based on the implementation
+/// of `T` as well. These methods will force deserialization of the value.
+#[derive(Clone)]
+pub struct Lazy<T: Read> {
+    /// This should only be `None` if `value` is initialized.
+    proto: Option<Pending<T>>,
```
patrick-ogrady: We can support `no_std` by just making `Lazy` a no-op I think (just go right into T)

codec/src/types/lazy.rs
```diff
@@ -66,7 +67,10 @@ use std::sync::OnceLock;
 pub struct Lazy<T: Read> {
     /// This should only be `None` if `value` is initialized.
     proto: Option<Pending<T>>,
```
patrick-ogrady: should we put this into the `std` block?

should we change the field name to `pending:`?

codec/src/types/lazy.rs
```diff
@@ -0,0 +1,307 @@
...
+    fn cmp(&self, other: &Self) -> core::cmp::Ordering {
+        self.get().cmp(&other.get())
+    }
+}
+
+impl<T: Read + Hash> Hash for Lazy<T> {
+    fn hash<H: core::hash::Hasher>(&self, state: &mut H) {
+        self.get().hash(state);
+    }
+}
```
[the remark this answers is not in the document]
cronokirby: I think this behavior is fine. Anything "invalid" is treated the same.

codec/src/types/lazy.rs
```diff
@@ -0,0 +1,307 @@
...
+        #[cfg(feature = "std")]
+        if let Some(pending) = &self.pending {
+            // Write raw bytes without length prefix (Bytes::write adds a length prefix)
+            buf.put_slice(&pending.bytes);
+            return;
+        }
+        self.get()
+            .expect("Lazy should have a value if pending is None")
+            .write(buf);
+    }
```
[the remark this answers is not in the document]
cronokirby: stale, addressed

consensus/src/simplex/scheme/bls12381_threshold/vrf.rs
```diff
@@ -512,16 +512,20 @@ impl<P: PublicKey, V: Variant> certificate::Scheme for Scheme<P, V> {
         let vote_message = subject.message();
         let seed_message = seed_message_from_subject(&subject);
 
+        let Some(signature) = attestation.signature.get() else {
```
patrick-ogrady, line 607: This is the ~only real load remaining on tokio in `alto`

patrick-ogrady:
After some benchmarking, realized that certificate signatures aren't getting the same treatment (and we get as many certificates as we do votes!)

Minimal spike to show (not a fix): https://github.com/commonwarexyz/monorepo/pull/2854 (**works**)

cronokirby:
> After some benchmarking, realized that certificate signatures aren't getting the same treatment (and we get as many certificates as we do votes!)
> 
> Minimal spike to show (not a fix): #2854 (**works**)

Merged into this PR

consensus/src/simplex/scheme/bls12381_threshold/vrf.rs
```diff
@@ -350,6 +352,77 @@ where
     }
 }
 
+/// Lazy certificate for BLS12-381 threshold VRF signatures.
+///
+/// This is the same as [`Signature`] but with lazy decoding of the BLS points
+/// to defer expensive point decompression from the network receive path.
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub struct Certificate<V: Variant> {
```
patrick-ogrady, line 357: It would be great if we could share this with `Signature` above? I did this originally as a lazy workaround (to do that though, I think we'd need to "nested Lazy" if you just wrap the Signature at the attestation level.

cronokirby: Perhaps, but this is also, semantically, a different thing, even if it happens to share the same field structure, so it seems fine to keep it as a separate struct? That's more amenable to change if we want to add more information to the certificate or use a specific type for it being a recovered signature, etc.

patrick-ogrady: Do we have the same issue with bls12381 multisig certificates (for lazy signatures)?

Need to check but just confirming we address.

patrick-ogrady: Basically, I think this should also be Lazy? https://github.com/commonwarexyz/monorepo/blob/c79fa1669c90bb24ec13356b61646c32e16cda93/cryptography/src/bls12381/certificate/multisig/mod.rs#L336

patrick-ogrady (dismissed):

PR #2834, opened 2026-01-15
[storage/qmdb/benches] split init from generation benchmarks

roberto-bayardo:
Split out the init and generate benchmarks into separate binaries.  The init benches have a huge setup cost, which gets performed even if we wish to only run the generation benchmarks.

Also moves out the pool setup from the timed part of the variable::init benchmark, which is how it's done in the fixed::init benchmark.

### DIFF
--- a/storage/src/qmdb/benches/generate_bench.rs
+++ b/storage/src/qmdb/benches/generate_bench.rs
@@ -6,12 +6,12 @@ use criterion::criterion_main;
 
 #[allow(dead_code)]
 mod fixed;
-mod keyless_generate;
+mod keyless;
 #[allow(dead_code)]
 mod variable;
 
 criterion_main!(
     fixed::generate::benches,
-    keyless_generate::benches,
+    keyless::generate::benches,
     variable::generate::benches,
 );
--- a/storage/src/qmdb/benches/keyless/generate.rs
+++ b/storage/src/qmdb/benches/keyless/generate.rs
@@ -1,3 +1,5 @@
+//! Benchmark the generation of a large randomly generated keyless database.
+
 use commonware_cryptography::Sha256;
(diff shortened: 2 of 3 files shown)


storage/src/qmdb/benches/variable/init.rs
```diff
@@ -81,19 +84,22 @@ fn bench_variable_init(c: &mut Criterion) {
                     |b| {
                         b.to_async(&runner).iter_custom(|iters| async move {
                             let ctx = context::get::<commonware_runtime::tokio::Context>();
+                            let pool = ctx.clone().create_pool(THREADS).unwrap();
+                            let any_cfg = any_cfg(pool);
```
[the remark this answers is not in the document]
roberto-bayardo: this is fine, the expensive buffers inside are wrapped in an Arc for fast cloning.

storage/Cargo.toml
```diff
@@ -75,9 +75,15 @@ bench = false
...
 [[bench]]
-name = "qmdb"
+name = "qmdb_generate"
 harness = false
-path = "src/qmdb/benches/bench.rs"
+path = "src/qmdb/benches/generate_bench.rs"
+required-features = ["test-traits"]
+
+[[bench]]
+name = "qmdb_init"
```
patrick-ogrady, line 84: is `init` the same as replay elsewhere?

roberto-bayardo: replay is a journal-level benchmark.  init builds a big qmdb, then closes it, then times re-opening it.  It does use replay but it does a lot more than just that.

patrick-ogrady (approved):
I'd like to see better naming here/`generate` broken apart but this is definitely an improvement.

PR #2846, opened 2026-01-16
[storage/journal] Fix Init at Size Recovery

patrick-ogrady:
Replaces: #2769

PR #2848, opened 2026-01-16
[storage/journal] Fix Init at Size Recovery

patrick-ogrady:
Replaces: #2769
Replaces: #2846

This fixes the issue called out in #2769, however, it makes me wonder if we should be using metadata for this (and just offsetting the index in the first section by the start offset)?


clabby:

PR #2856, opened 2026-01-17
fix(utils): correct priority_set peek/pop documentation

strmfos:
Fix misleading docs for peek() and pop() - they return lowest priority item, not highest.


clabby:

Issue #2857, opened 2026-01-17
[storage] finer grained locking for improved concurrency


roberto-bayardo:
Our storage structures use coarse-grained locking, which, for example, prevents any concurrent reads during I/O heavy operations such as `sync`.  Finer grained could improve concurrency & hence performance.

PR #2865, opened 2026-01-18
[storage/journal/contiguous] allow concurrent reads during Persistable::sync()

roberto-bayardo:
This PR allows concurrent reads to a `Persistable` storage structure during a `sync` operation.  Previously sync was &mut, locking out concurrent readers. Implementations now leverage interior mutability to avoid the &mut requirement.

Unfortunately the change requires updating a lot of call sites since many getters are now async in order to acquire the interior read locks.

Only `freezer` remains to be updated with this capability, tracked here: https://github.com/commonwarexyz/monorepo/issues/2910

Changes all "replay" functions to have the same argument order (write buffer size before `start_pos`). [breaking api]

Towards: https://github.com/commonwarexyz/monorepo/issues/2857

### DIFF
--- a/examples/sync/src/databases/any.rs
+++ b/examples/sync/src/databases/any.rs
@@ -19,7 +19,7 @@ use commonware_storage::{
     },
 };
 use commonware_utils::{NZUsize, NZU16, NZU64};
-use std::{future::Future, num::NonZeroU64};
+use std::num::NonZeroU64;
 use tracing::error;
 
 /// Database type alias for the Clean state.
@@ -112,25 +112,25 @@ where
         panic!("operations should end with a commit");
     }
 
-    fn root(&self) -> Key {
-        self.root()
+    async fn root(&self) -> Key {
+        self.root().await
     }
 
-    fn size(&self) -> Location {
-        LogStore::bounds(self).end
+    async fn size(&self) -> Location {
+        LogStore::bounds(self).await.end
     }
 
-    fn inactivity_floor(&self) -> Location {
+    async fn inactivity_floor(&self) -> Location {
         self.inactivity_floor_loc()
     }
 
-    fn historical_proof(
+    async fn historical_proof(
         &self,
         op_count: Location,
         start_loc: Location,
         max_ops: NonZeroU64,
-    ) -> impl Future<Output = Result<(Proof<Key>, Vec<Self::Operation>), qmdb::Error>> + Send {
-        self.historical_proof(op_count, start_loc, max_ops)
+    ) -> Result<(Proof<Key>, Vec<Self::Operation>), qmdb::Error> {
+        self.historical_proof(op_count, start_loc, max_ops).await
     }
 
     fn name() -> &'static str {
--- a/examples/sync/src/databases/immutable.rs
+++ b/examples/sync/src/databases/immutable.rs
@@ -13,7 +13,7 @@ use commonware_storage::{
     },
 };
 use commonware_utils::{NZUsize, NZU16, NZU64};
-use std::{future::Future, num::NonZeroU64};
+use std::num::NonZeroU64;
 use tracing::error;
 
 /// Database type alias for the clean (merkleized, durable) state.
@@ -113,27 +113,27 @@ where
         unreachable!("operations must end with a commit");
     }
 
-    fn root(&self) -> Key {
-        self.root()
+    async fn root(&self) -> Key {
+        self.root().await
     }
 
-    fn size(&self) -> Location {
-        LogStore::bounds(self).end
+    async fn size(&self) -> Location {
+        LogStore::bounds(self).await.end
     }
 
-    fn inactivity_floor(&self) -> Location {
+    async fn inactivity_floor(&self) -> Location {
         // For Immutable databases, all retained operations are active,
         // so the inactivity floor equals the pruning boundary.
-        LogStore::bounds(self).start
+        LogStore::bounds(self).await.start
     }
 
-    fn historical_proof(
+    async fn historical_proof(
         &self,
         op_count: Location,
         start_loc: Location,
         max_ops: NonZeroU64,
-    ) -> impl Future<Output = Result<(Proof<Key>, Vec<Self::Operation>), qmdb::Error>> + Send {
-        self.historical_proof(op_count, start_loc, max_ops)
+    ) -> Result<(Proof<Key>, Vec<Self::Operation>), qmdb::Error> {
+        self.historical_proof(op_count, start_loc, max_ops).await
     }
 
     fn name() -> &'static str {
--- a/storage/src/freezer/storage.rs
+++ b/storage/src/freezer/storage.rs
@@ -3,7 +3,7 @@ use crate::{
     journal::segmented::oversized::{
         Config as OversizedConfig, Oversized, Record as OversizedRecord,
     },
-    kv, Persistable,
+    kv,
 };
 use commonware_codec::{CodecShared, Encode, FixedSize, Read, ReadExt, Write as CodecWrite};
 use commonware_cryptography::{crc32, Crc32, Hasher};
@@ -1166,25 +1166,6 @@ impl<E: Storage + Metrics + Clock, K: Array, V: CodecShared> kv::Gettable for Fr
 impl<E: Storage + Metrics + Clock, K: Array, V: CodecShared> kv::Updatable for Freezer<E, K, V> {
     async fn update(&mut self, key: Self::Key, value: Self::Value) -> Result<(), Self::Error> {
         self.put(key, value).await?;
-        Ok(())
-    }
-}
-
-impl<E: Storage + Metrics + Clock, K: Array, V: CodecShared> Persistable for Freezer<E, K, V> {
-    type Error = Error;
-
-    async fn commit(&mut self) -> Result<(), Self::Error> {
-        self.sync().await?;
-        Ok(())
-    }
-
-    async fn sync(&mut self) -> Result<(), Self::Error> {
-        self.sync().await?;
-        Ok(())
-    }
-
-    async fn destroy(self) -> Result<(), Self::Error> {
-        self.destroy().await?;
         Ok(())
     }
 }
--- a/storage/src/journal/contiguous/fixed.rs
+++ b/storage/src/journal/contiguous/fixed.rs
@@ -185,6 +185,62 @@ impl<E: Clock + Storage + Metrics, A: CodecFixedShared> super::ContiguousReader
                 other => other,
             })
     }
+
+    async fn replay(
+        &self,
+        buffer: NonZeroUsize,
+        start_pos: u64,
+    ) -> Result<impl Stream<Item = Result<(u64, A), Error>> + Send, Error> {
+        let items_per_blob = self.items_per_blob;
+        let pruning_boundary = self.guard.pruning_boundary;
+
+        // Validate bounds.
+        if start_pos > self.guard.size {
+            return Err(Error::ItemOutOfRange(start_pos));
+        }
+        if start_pos < pruning_boundary {
+            return Err(Error::ItemPruned(start_pos));
+        }
+
+        let start_section = start_pos / items_per_blob;
+        let section_start = start_section * items_per_blob;
+
+        // Calculate start position within the section.
+        let first_in_section = pruning_boundary.max(section_start);
+        let start_pos_in_section = start_pos - first_in_section;
+
+        // Check all middle sections (not oldest, not tail) in range are complete.
+        let journal = &self.guard.journal;
+        if let (Some(oldest), Some(newest)) =
+            (journal.oldest_section(), journal.newest_section())
+        {
+            let first_to_check = start_section.max(oldest + 1);
+            for section in first_to_check..newest {
+                let len = journal.section_len(section).await?;
+                if len < items_per_blob {
+                    return Err(Error::Corruption(format!(
+                        "section {section} incomplete: expected {items_per_blob} items, got {len}"
+                    )));
+                }
+            }
+        }
+
+        let inner_stream = journal
+            .replay(start_section, start_pos_in_section, buffer)
+            .await?;
+
+        // Transform (section, pos_in_section, item) to (global_pos, item).
+        let stream = inner_stream.map(move |result| {
+            result.map(|(section, pos_in_section, item)| {
+                let section_start = section * items_per_blob;
+                let first_in_section = pruning_boundary.max(section_start);
+                let global_pos = first_in_section + pos_in_section;
+                (global_pos, item)
+            })
+        });
+
+        Ok(stream)
+    }
 }
 
 impl<E: Clock + Storage + Metrics, A: CodecFixedShared> Journal<E, A> {
@@ -721,82 +777,6 @@ impl<E: Clock + Storage + Metrics, A: CodecFixedShared> Journal<E, A> {
             })
     }
 
-    /// Returns an ordered stream of all items in the journal with position >= `start_pos`.
-    ///
-    /// # Errors
-    ///
-    /// - [Error::ItemOutOfRange] if `start_pos > size`
-    /// - [Error::ItemPruned] if `start_pos < pruning_boundary`
-    /// - [Error::Corruption] if a middle section is incomplete
-    pub async fn replay(
-        &self,
-        buffer: NonZeroUsize,
-        start_pos: u64,
-    ) -> Result<impl Stream<Item = Result<(u64, A), Error>> + Send, Error> {
-        let items_per_blob = self.items_per_blob;
-
-        // Validate inputs and get inner stream.
-        let (inner_stream, pruning_boundary) = {
-            let inner = self.inner.read().await;
-
-            // Validate bounds.
-            if start_pos > inner.size {
-                return Err(Error::ItemOutOfRange(start_pos));
-            }
-
-            // Make sure we aren't trying to read pruned items.
-            if start_pos < inner.pruning_boundary {
-                return Err(Error::ItemPruned(start_pos));
-            }
-
-            let start_section = start_pos / self.items_per_blob;
-            let section_start = start_section * self.items_per_blob;
-
-            // Calculate start position within the section
-            let first_in_section = inner.pruning_boundary.max(section_start);
-            let start_pos_in_section = start_pos - first_in_section;
-
-            let items_per_blob = self.items_per_blob;
-
-            // Check all middle sections (not oldest, not tail) in range are complete.
-            // The oldest section may be partial due to mid-section pruning boundary.
-            // The tail section may be partial because it's not been fully filled yet.
-            let journal = &inner.journal;
-            if let (Some(oldest), Some(newest)) =
-                (journal.oldest_section(), journal.newest_section())
-            {
-                // Start from max(start_section, oldest+1) to skip oldest which may be partial
-                let first_to_check = start_section.max(oldest + 1);
-                for section in first_to_check..newest {
-                    let len = journal.section_len(section).await?;
-                    if len < items_per_blob {
-                        return Err(Error::Corruption(format!(
-                        "section {section} incomplete: expected {items_per_blob} items, got {len}"
-                    )));
-                    }
-                }
-            }
-
-            let inner_stream = journal
-                .replay(start_section, start_pos_in_section, buffer)
-                .await?;
-            // Return inner stream.
-            (inner_stream, inner.pruning_boundary)
-        };
-
-        // Transform (section, pos_in_section, item) to (global_pos, item).
-        let stream = inner_stream.map(move |result| {
-            result.map(|(section, pos_in_section, item)| {
-                let section_start = section * items_per_blob;
-                let first_in_section = pruning_boundary.max(section_start);
-                let global_pos = first_in_section + pos_in_section;
-                (global_pos, item)
-            })
-        });
-
-        Ok(stream)
-    }
-
     /// Allow the journal to prune items older than `min_item_pos`. The journal may not prune all
     /// such items in order to preserve blob boundaries, but the amount of such items will always be
     /// less than the configured number of items per blob. Returns true if any items were pruned.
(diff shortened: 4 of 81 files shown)


storage/src/journal/contiguous/variable.rs
```diff
@@ -500,39 +540,53 @@ impl<E: Storage + Metrics, V: CodecShared> Journal<E, V> {
...
-        // Transform the stream to include position information
-        let transformed = data_stream.enumerate().map(move |(idx, result)| {
-            result.map(|(_section, _offset, _size, item)| {
-                // Calculate position: start_pos + items read
-                let pos = start_pos + idx as u64;
-                (pos, item)
-            })
+            // Check if position has been pruned (can happen if prune runs between iterations)
+            if pos < inner.oldest_retained_pos {
+                return Some((Err(Error::ItemPruned(pos)), pos + 1));
```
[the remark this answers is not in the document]
roberto-bayardo: This should be fixed when I rebase in https://github.com/commonwarexyz/monorepo/pull/2872

storage/src/qmdb/store/mod.rs
```diff
@@ -99,8 +99,13 @@ pub trait MerkleizedStore: LogStore {
         start_loc: Location,
         max_ops: NonZeroU64,
     ) -> impl Future<Output = Result<(Proof<Self::Digest>, Vec<Self::Operation>), Error>> + Send
+    where
+        Self: Sync,
```
danlaine:
```suggestion
```

I think we can remove this? It compiles without these lines.

roberto-bayardo: yes removed

storage/src/journal/contiguous/fixed.rs
```diff
@@ -409,43 +473,93 @@ impl<E: Storage + Metrics, A: CodecFixedShared> Journal<E, A> {
...
+                .segmented
+                .oldest_section()
+                .map(|s| s * self.items_per_blob)
+                .unwrap_or(0);
+
+            // Effective start position is max of requested start and pruning boundary
+            start_pos.max(pruning_boundary)
+        };
+
+        // Check if effective start would be at or beyond the end
```
danlaine: Should we keep the `self.inner.read()` lock here and on lines 506 - 509 instead of releasing it and then re-acquiring it?

storage/src/mmr/journaled.rs
```diff
@@ -486,29 +513,48 @@ impl<E: RStorage + Clock + Metrics, D: Digest> CleanMmr<E, D> {
...
+            }
+
+            // Now that the pinned node set has been recomputed, it's safe to prune the mem_mmr
+            // and reinstate them.
+            inner.mem_mmr.prune_all();
+            inner.mem_mmr.add_pinned_nodes(pinned_nodes);
+
+            nodes_to_append
+        };
+
```
danlaine, line 570: At this point `nodes_to_append` aren't in `inner.mem_mmr` or `self.journal`, and we don't have `self.inner` locked, so it's possible a call to `get_node` could occur and see that these nodes are missing even though they shouldn't be.

roberto-bayardo: Right, I shouldn't have moved the mem_mmr pruning up here. Let me move it back.

storage/src/journal/contiguous/fixed.rs
```diff
@@ -294,32 +307,52 @@ impl<E: Storage + Metrics, A: CodecFixedShared> Journal<E, A> {
...
+            let mut inner = self.inner.write().await;
+            let position = inner.size;
+            let (section, _pos_in_section) = self.position_to_section(position);
+
+            inner.segmented.append(section, item).await?;
+            inner.size += 1;
+
+            let section_filled = inner.size.is_multiple_of(self.items_per_blob);
+            (position, section, section_filled)
+        }; // Write lock released here
```
danlaine: To be clear, should we specify that it's `self.inner`'s write lock that's dropped here

storage/src/journal/contiguous/fixed.rs
```diff
@@ -294,32 +307,52 @@ impl<E: Storage + Metrics, A: CodecFixedShared> Journal<E, A> {
...
+
```
danlaine: At this point, we don't have `self.inner` locked but we've already incremented `inner.size` so this execution is possible:

1. Thread A enters `append`, increments `inner.size` resulting in `section_filled`
2. Thread A releases `self.inner.write()` here
3. Thread B enters `sync`, sees `tail_section` is the new (logical) tail sectIon
4. Thread B calls `inner.segmented.sync(tail_section)` for `tail_section` that hasn't been created because thread A didn't run `inner.segmented.ensure_section_exists(section + 1).await?;`

storage/src/mmr/journaled.rs
roberto-bayardo: OK, moved mem_mmr pruning back to after the journal.appends.  This allows a read lock on inner during the "gather" phase, and a write lock only during the mem_mmr pruning phase.

storage/src/journal/contiguous/fixed.rs
roberto-bayardo: sigh, good catch.  Delayed the size increment until *after* the section is guaranteed to exist.

storage/src/journal/contiguous/fixed.rs
roberto-bayardo: I removed this (AI added) comment, I think it's clear from the structure what lock is being released.

storage/src/journal/contiguous/fixed.rs
roberto-bayardo: fixed

storage/src/journal/contiguous/fixed.rs
```diff
@@ -294,32 +307,52 @@ impl<E: Storage + Metrics, A: CodecFixedShared> Journal<E, A> {
...
+        self.inner.read().await.size
     }
 
     /// Append a new item to the journal. Return the item's position in the journal, or error if the
     /// operation fails.
-    pub async fn append(&mut self, item: A) -> Result<u64, Error> {
-        let position = self.size;
-        let (section, _pos_in_section) = self.position_to_section(position);
-
-        self.inner.append(section, item).await?;
```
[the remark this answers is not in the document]
roberto-bayardo: yes Dan pointed this out and it should be fixed now.

storage/src/journal/contiguous/fixed.rs
```diff
@@ -454,65 +568,90 @@ impl<E: Storage + Metrics, A: CodecFixedShared> Journal<E, A> {
...
-        self.inner
+        inner
+            .segmented
             .init_section_at_size(tail_section, tail_items)
             .await?;
-        self.inner.sync_all().await?;
+        inner.segmented.sync_all().await?;
 
-        self.size = new_size;
+        inner.size = new_size;
```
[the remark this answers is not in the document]
roberto-bayardo: this is fine, we assume failures like that would be unrecoverable anyway.

storage/src/journal/contiguous/fixed.rs
roberto-bayardo: Actually this would be fine, since you are allowed to sync non-existent sections (the segmented journals just return Ok in this case).

roberto-bayardo: I have reverted to the original flow, with a comment added to sync() to clarify why it's safe.

storage/src/journal/contiguous/fixed.rs
```diff
@@ -409,43 +479,91 @@ impl<E: Storage + Metrics, A: CodecFixedShared> Journal<E, A> {
...
+            // Effective start position is max of requested start and pruning boundary
+            let effective_start_pos = start_pos.max(pruning_boundary);
+
+            if effective_start_pos >= inner.size {
+                return Ok(futures::future::Either::Left(futures::stream::empty()));
+            }
+
+            effective_start_pos
+        };
+
```
danlaine: We still drop and then re-acquire `self.inner.read()` -- should we combine these scopes so we don't need to re-acquire it?

storage/src/journal/contiguous/fixed.rs
```diff
@@ -409,43 +479,91 @@ impl<E: Storage + Metrics, A: CodecFixedShared> Journal<E, A> {
...
+            // Check if position has been pruned (can happen if prune runs between iterations)
+            let pruning_boundary = inner
+                .segmented
+                .oldest_section()
+                .map(|s| s * items_per_blob)
+                .unwrap_or(0);
+
+            if pos < pruning_boundary {
+                // Skip to the pruning boundary instead of returning an error
```
danlaine: This does return an error, though?

roberto-bayardo: As it should!  Updated the comment here, and clarified in the function documentation for both this and variable.rs replay that ItemPruned error will be returned if any requested item is pruned.

storage/src/journal/contiguous/fixed.rs
roberto-bayardo: Yes done.

storage/src/journal/contiguous/fixed.rs
```diff
@@ -378,74 +435,137 @@ impl<E: Storage + Metrics, A: CodecFixedShared> Journal<E, A> {
...
+    ///  - [Error::ItemOutOfRange] if `start_pos` exceeds the journal size.
+    ///
     /// # Integrity
     ///
     /// If any corrupted data is found, or if any non-tail section has fewer items than
     /// `items_per_blob`, the stream will return an error.
     pub async fn replay(
         &self,
-        buffer: NonZeroUsize,
+        _buffer: NonZeroUsize,
```
danlaine: Should we remove `_buffer` from this concrete impl since it's no longer used? Kind of a bummer that we lose the buffering from the inner layer here but I'm sure we can optimize this later.

roberto-bayardo: Yeah I meant to revisit this. I want to restore the buffering or I imagine (though haven't tested) this will be a huge performance regression.

roberto-bayardo: I have restored replay to use inner journal replay with buffering both here and for variable.

storage/src/ordinal/storage.rs
```diff
@@ -376,25 +376,28 @@ impl<E: Storage + Metrics + Clock, V: CodecFixed<Cfg = ()>> Ordinal<E, V> {
...
 
+        // Take pending sections atomically
+        let pending = std::mem::take(&mut *self.pending.write().await);
+
         // Sync all modified blobs
-        let mut futures = Vec::with_capacity(self.pending.len());
-        for &section in &self.pending {
-            futures.push(self.blobs.get(&section).unwrap().sync());
+        let mut futures = Vec::with_capacity(pending.len());
+        for section in &pending {
```
patrick-ogrady, line 393: I suppose we already assume that a write landing while we are syncing won't be an issue (as its like a sync that partially completed?)

storage/src/qmdb/immutable/mod.rs
```diff
@@ -464,17 +469,20 @@ impl<
...
+    async fn inactivity_floor_loc(&self) -> Location {
+        self.journal.pruning_boundary().await
     }
 
     fn is_empty(&self) -> bool {
-        self.op_count() == 0
+        // An immutable store is never empty since it always has at least one commit.
+        // We consider it "empty" if there are only commits and no data operations.
+        // Since we always have at least one commit, this is a rough approximation.
+        self.last_commit_loc == 0
```
danlaine: This is implementing

```rs
/// Returns true if there are no active keys in the database.
fn is_empty(&self) -> bool;
 ````
 
 And this isn't exactly doing that. But the old code works the same way, so I think perhaps its the comment on `is_empty()` which should be updated? It also seems like `is_empty()` is only used in `test_any_db_empty` in `qmdb/any/unordered/mod.rs`so we could just remove it altogether.

File: storage/src/ordinal/storage.rs (hunk 12 above)
Comment (danlaine) [reply storage/src/ordinal/storage.rs:393]: ~~Yeah we treat failures during write operations, including `sync()`, to be fatal and in-memory state may be inconsistent after a failed write~~

Edit: misread comment above

Comment (roberto-bayardo) [reply storage/src/ordinal/storage.rs:393]: I think he may have been referring to concurrent (successful) writes?  In that case the contract of sync is only that writes which completed *before* invocation are guaranteed to be durable.

File: storage/src/qmdb/immutable/mod.rs (hunk 13 above)
Comment (roberto-bayardo) [reply storage/src/qmdb/immutable/mod.rs]: Agree this is pretty hacky as is_empty just isn't well defined for these kinds of stores where we can prune (& hence lose track) of previously set items.  Have removed is_empty from the traits.

File: storage/src/journal/contiguous/variable.rs [hunk 14]
```diff
@@ -126,22 +126,15 @@ impl<C> Config<C> {
 /// Note that we don't recover from the case where offsets.oldest_retained_pos() >
 /// data.oldest_retained_pos(). This should never occur because we always prune the data journal
 /// before the offsets journal.
-pub struct Journal<E: Clock + Storage + Metrics, V: Codec> {
+/// Inner state protected by a single lock.
```
danlaine: `pub struct journal` should have the `///` comment for visibility

danlaine (approved):

danlaine (dismissed):

storage/src/journal/contiguous/variable.rs
```diff
@@ -434,25 +471,28 @@ impl<E: Clock + Storage + Metrics, V: CodecShared> Journal<E, V> {
...
         }
     }
 
     /// Returns the location before which all items have been pruned.
-    pub fn pruning_boundary(&self) -> u64 {
-        self.oldest_retained_pos().unwrap_or(self.size)
+    pub async fn pruning_boundary(&self) -> u64 {
+        self.oldest_retained_pos()
+            .await
+            .unwrap_or(self.size().await)
```
patrick-ogrady:
```
  --- a/storage/src/journal/contiguous/variable.rs
  +++ b/storage/src/journal/contiguous/variable.rs
  @@ -489,9 +489,12 @@ impl<E: Clock + Storage + Metrics, V: CodecShared> Journal<E, V> {

       /// Returns the location before which all items have been pruned.
       pub async fn pruning_boundary(&self) -> u64 {
  -        self.oldest_retained_pos()
  -            .await
  -            .unwrap_or(self.size().await)
  +        let inner = self.inner.read().await;
  +        if inner.size == inner.oldest_retained_pos {
  +            inner.size
  +        } else {
  +            inner.oldest_retained_pos
  +        }
       }
```

(to avoid 2 locks)

roberto-bayardo: this actually just needs to return "inner.oldest_retained_pos"  (no if necessary), which makes me wonder why we never renamed this variable to "pruning_boundary" which better describes what it is (this is the name we use in fixed.rs). I have updated this implementation and also renamed this struct field to pruning_boundary.

storage/src/qmdb/store/mod.rs
```diff
@@ -100,7 +97,10 @@ pub trait MerkleizedStore: LogStore {
         max_ops: NonZeroU64,
     ) -> impl Future<Output = Result<(Proof<Self::Digest>, Vec<Self::Operation>), Error>> + Send
     {
-        self.historical_proof(self.op_count(), start_loc, max_ops)
+        async move {
```
patrick-ogrady, line 108: Given this is over self, can we just have historical_proof get the op_count? Or add a version that does?

I suspect this would involve acquiring a lock and then releasing it.

storage/src/qmdb/keyless/mod.rs
```diff
@@ -200,7 +202,7 @@ impl<E: Storage + Clock + Metrics, V: VariableValue, H: Hasher>
         start_loc: Location,
         max_ops: NonZeroU64,
     ) -> Result<(Proof<H::Digest>, Vec<Operation<V>>), Error> {
-        self.historical_proof(self.op_count(), start_loc, max_ops)
+        self.historical_proof(self.op_count().await, start_loc, max_ops)
```
patrick-ogrady: Another spot where we "wrap an await"...maybe this is as good as it gets.

storage/src/qmdb/immutable/mod.rs
```diff
@@ -199,7 +200,7 @@ impl<
         start_index: Location,
         max_ops: NonZeroU64,
     ) -> Result<(Proof<H::Digest>, Vec<Operation<K, V>>), Error> {
-        let op_count = self.op_count();
+        let op_count = self.op_count().await;
```
patrick-ogrady: We inline this in other places

examples/sync/src/databases/any.rs
```diff
@@ -111,16 +111,17 @@ where
         panic!("operations should end with a commit");
     }
 
-    fn root(&self) -> Key {
+    fn root(&self) -> impl Future<Output = Key> + Send {
```
patrick-ogrady: You should just be able to use `async` here?

examples/sync/src/databases/immutable.rs
```diff
@@ -112,15 +112,15 @@ where
         unreachable!("operations must end with a commit");
     }
 
-    fn root(&self) -> Key {
+    fn root(&self) -> impl Future<Output = Key> + Send {
```
patrick-ogrady: async here?

storage/src/freezer/storage.rs
```diff
@@ -1172,25 +1172,6 @@ impl<E: Storage + Metrics + Clock, K: Array, V: CodecShared> kv::Updatable for F
     }
 }
 
-impl<E: Storage + Metrics + Clock, K: Array, V: CodecShared> Persistable for Freezer<E, K, V> {
```
patrick-ogrady, line 1173: Assuming this is to avoid implementing some sort of ref sync?

storage/src/qmdb/current/ordered/db.rs
```diff
@@ -170,8 +170,9 @@ where
         };
         let height = Self::grafting_height();
         let mmr = &self.any.log.mmr;
-        let proof =
-            OperationProof::<H::Digest, N>::new(hasher, &self.status, height, mmr, loc).await?;
+        let status = self.status.read().await;
+
```
patrick-ogrady: remove \n?

patrick-ogrady (commented):
Everything except `fixed.rs`, `variable.rs`, `journaled.rs`

storage/src/qmdb/store/mod.rs
roberto-bayardo: Perhaps I'm misunderstanding you, but historical_proof exists to allow proofs over historical sizes, so we wouldn't want it to assume it knows the size intended by the caller.

storage/src/freezer/storage.rs
roberto-bayardo: yep

roberto-bayardo: (tracking issue to fix this is in the PR description)

examples/sync/src/databases/immutable.rs
roberto-bayardo: done

examples/sync/src/databases/any.rs
roberto-bayardo: yup done

storage/src/qmdb/immutable/mod.rs
roberto-bayardo: changed to inlined.

storage/src/qmdb/current/ordered/db.rs
roberto-bayardo: done

patrick-ogrady:
While reviewing this PR, it occurred to me that we may not be appreciating the subtleties of `append`, `prune`, and `sync` all being immutable when we frequently access getters in different calls (dropping the lock between) and then get data (which now may have been pruned beneath us).

I suspect we may either produce incorrect calculations if we have a function with `size().await -oldest().await` and/or check `request > oldest().await` and then fetch?

I asked an LLM to find some examples and it provided (3 is ~cheating but it gives a good example):

```
  1. keyless/mod.rs:99-107 - TOCTOU

  pub async fn get(&self, loc: Location) -> Result<Option<V>, Error> {
      let op_count = self.op_count().await;  // Lock released
      if loc >= op_count {
          return Err(Error::LocationOutOfBounds(loc, op_count));
      }
      let op = self.journal.read(loc).await?;  // Prune could have happened
      Ok(op.into_value())
  }
  Issue: Bounds check passes, then prune() runs, then read() gets ItemPruned.

  2. immutable/mod.rs:126-158 - Double Check Race

  pub async fn get(&self, key: &K) -> Result<Option<V>, Error> {
      let oldest = self.oldest_retained_loc().await;  // Check 1
      for &loc in iter {
          if loc < oldest { continue; }
          if let Some(v) = self.get_from_loc(key, loc).await? {  // Calls check 2
              return Ok(Some(v));
          }
      }
  }

  async fn get_from_loc(&self, key: &K, loc: Location) -> Result<Option<V>, Error> {
      if loc < self.oldest_retained_loc().await {  // Check 2 - stale by now?
          return Err(Error::OperationPruned(loc));
      }
      let Operation::Set(k, v) = self.journal.read(loc).await?  // Read
  Issue: oldest_retained_loc() is called twice (line 127 and 145). Between them, pruning could advance, causing the second check or the read to fail unexpectedly.

  3. authenticated.rs:273-302 - Bounds Check to Batch Read Race

  pub async fn historical_proof(...) -> Result<(Proof<H::Digest>, Vec<C::Item>), Error> {
      let leaves = self.size().await;           // Line 279
      if historical_leaves > leaves { ... }      // Line 280
      if start_loc >= historical_leaves { ... }  // Line 283
      let end_loc = ...;                         // Line 286

      // ... MMR proof generation (await) ...

      let futures = (*start_loc..*end_loc)
          .map(|i| self.journal.read(i))         // Line 295 - reads happen here
          .collect::<Vec<_>>();
      try_join_all(futures).await?               // Prune could have run by now
  Issue: Size checked at 279, but reads happen at 295-298 after other async work. Window for prune() to invalidate the range.

  4. qmdb/store/db.rs:199-208 - Assert Then Read

  async fn get_op(&self, loc: Location) -> Result<Operation<K, V>, Error> {
      assert!(loc < self.op_count().await);  // Panic if out of bounds
      self.log.read(*loc).await.map_err(|e| match e {
          crate::journal::Error::ItemPruned(_) => Error::OperationPruned(loc),
          e => Error::Journal(e),
      })
  }
  Issue: Assert passes, then prune() runs, then read() returns ItemPruned. This one at least handles ItemPruned gracefully, converting it to OperationPruned.

```

I think its OK if we error in such cases but can't panic (obvs). I can't find any such cases but the free form nature of these getters may be tricky to use as a result in other places.

roberto-bayardo:
Putting this in draft until I can merge this more contained PR first: https://github.com/commonwarexyz/monorepo/pull/3080

PR #2872, opened 2026-01-19
[Storage] Fix `journal::contiguous::Variable::init_at_size`

danlaine:
Replaces #2769

## Problem

The `init_at_size` method creates a journal in a "pruned" state at a given logical position, which is used for state sync. However, after a crash and restart, the journal could not correctly recover its state because:

1. **Fixed Journal**: The logical `size` and `pruning_boundary` were not persisted. After `init_at_size(N)` where `N` is mid-section (e.g., `N % items_per_blob != 0`), reopening the journal would compute the wrong size from physical state alone.

2. **Variable Journal**: The `align_journals` function assumed `oldest_retained_pos` was always section-aligned, causing assertion failures or incorrect size computation after `init_at_size` with a mid-section position.

## Solution

Added a `Metadata` to the contiguous fixed journal implementation which persists the pruning boundary when it is not section-aligned. This allows us to do away with dummy data and eliminates the scenario that exists in the old code in which `init_at_size` would cause the journal to report that it has locations that it actually doesn't.

## Backward Compatibility

Added a new partition for the metadata with suffix `-metadata` after the partition given to the variable journal. If this metadata isn't present (i.e. because the journal was already used before this PR) then it falls back to using the state of the blobs. This PR is marked `breaking-format` because the addition of the metadata partition changes the conformance test hashes, but existing journals should still be usable.

## Other Changes

`contiguous::fixed::Journal::replay` returns `Error::ItemPruned` when asked to `replay` starting from a location before the pruning boundary. This seems less error prone than silently skipping over pruned items, as the old code did.


storage/src/journal/contiguous/fixed.rs
```diff
@@ -403,10 +649,10 @@ impl<E: Storage + Metrics, A: CodecFixedShared> Journal<E, A> {
 
     /// Returns an ordered stream of all items in the journal with position >= `start_pos`.
     ///
-    /// # Integrity
```
danlaine, line 403: ~~This check is now done in `init`~~

Edit: moved the integrity check back to `replay` from `init` but updated the comment to be a bit more specific about the errors it can return.

storage/src/journal/contiguous/fixed.rs
```diff
@@ -130,25 +156,69 @@ impl<E: Storage + Metrics, A: CodecFixedShared> Journal<E, A> {
...
-        let size = match (oldest_section, newest_section) {
-            (Some(_), Some(newest)) => {
-                // Compute size from the tail (newest) section
-                let tail_len = inner.section_len(newest).await?;
-                newest * items_per_blob + tail_len
-            }
-            _ => 0,
+        // Initialize metadata store
+        let meta_cfg = MetadataConfig {
+            partition: format!("{}{}", cfg.partition, META_SUFFIX),
```
patrick-ogrady: We usually do a slightly different form for computing this partition.

storage/src/journal/contiguous/fixed.rs
```diff
@@ -130,25 +156,69 @@ impl<E: Storage + Metrics, A: CodecFixedShared> Journal<E, A> {
         let items_per_blob = cfg.items_per_blob.get();
 
         let segmented_cfg = SegmentedConfig {
-            partition: cfg.partition,
+            partition: cfg.partition.clone(),
```
patrick-ogrady: As noted, we'd usually call this `_journal` and the below `_partition` but we need to be sure we don't break existing usage: https://github.com/tempoxyz/tempo/blob/a253353341c34a693cfa3fba57880b1fd42c1762/crates/commonware-node/src/dkg/manager/actor/state.rs#L53

storage/src/journal/contiguous/fixed.rs
```diff
@@ -295,8 +540,20 @@ impl<E: Storage + Metrics, A: CodecFixedShared> Journal<E, A> {
     /// Only the tail section can have pending updates since historical sections are synced
     /// when they become full.
     pub async fn sync(&mut self) -> Result<(), Error> {
+        // Sync inner journal
         let tail_section = self.size / self.items_per_blob;
-        self.inner.sync(tail_section).await
+        self.inner.sync(tail_section).await?;
```
roberto-bayardo, line 542: Are both these sync() calls needed for crash recovery? If not we should introduce a commit() that just syncs one of them.  (Hoping to avoid a performance regression here...)

danlaine: I think we do need to sync the metadata here. Otherwise in `init()` the metadata may be stale and it won't be possible to tell whether its pruning boundary and size should be trusted.

roberto-bayardo: hrmph. I'm not sure this is a good solution then. We really should keep the commit path to a single fsync(), IMO. What's kind of odd is variable dbs will be fine since they sync only the data journal (one fsync). But the fixed dbs will require this double-sync if I understand correctly?

danlaine: I'll have a crack at updating this so that we don't persist the size (this removes some corruption detection added by this PR, but we'll no worse off than the existing code in that regard) which will allow us to only write the metadata when the pruning boundary switches from being section-aligned to not section-aligned and vice versa.

storage/src/journal/contiguous/fixed.rs
roberto-bayardo: We don't also check the oldest blob that it contains the expected # of items (based on pruning boundary)?

storage/src/journal/contiguous/fixed.rs
```diff
@@ -464,36 +688,65 @@ impl<E: Storage + Metrics, A: CodecFixedShared> Journal<E, A> {
...
 
-        // Initialize the tail section if needed
+        // Ensure tail section exists
         let tail_section = new_size / self.items_per_blob;
-        let tail_items = new_size % self.items_per_blob;
-        self.inner
-            .init_section_at_size(tail_section, tail_items)
-            .await?;
-        self.inner.sync_all().await?;
+        self.inner.ensure_section_exists(tail_section).await?;
```
roberto-bayardo, line 801: I guess the fact that we delete the blob, then recreate it here means there's no need to sync it and it's guaranteed empty on restart?

storage/src/journal/contiguous/fixed.rs
roberto-bayardo, line 750: What happens if we prune the blobs, but fail before we write the updated metadata?  Feels like we're in an inconsistent state.

roberto-bayardo: Oh I see that recover_boundary handles this (always use blob boundary if > pruning_boundary).

storage/src/journal/contiguous/fixed.rs
```diff
@@ -160,9 +242,90 @@ impl<E: Storage + Metrics, A: CodecFixedShared> Journal<E, A> {
...
+
+        let (pruning_boundary, needs_update) = match meta_pruning_boundary {
+            // Mid-section metadata: validate against blobs
+            Some(meta_pruning_boundary)
+                if !meta_pruning_boundary.is_multiple_of(items_per_blob) =>
+            {
+                let meta_oldest_section = meta_pruning_boundary / items_per_blob;
+                match inner.oldest_section() {
+                    Some(oldest_section) if meta_oldest_section < oldest_section => {
+                        info!(
```
roberto-bayardo: elsewhere I use warn! when reporting crash recovery, we should probably try to be consistent.

storage/src/journal/contiguous/fixed.rs
roberto-bayardo: Mind adding a comment here that metadata write MUST happen after blobs are pruned to avoid inconsistencies?

storage/src/journal/contiguous/fixed.rs
```diff
@@ -105,50 +123,114 @@ pub struct Config {
...
+            Err(commonware_runtime::Error::PartitionMissing(_)) => Ok(Vec::new()),
+            Err(err) => Err(Error::Runtime(err)),
+        }
+    }
+
+    /// Select the blobs partition using legacy-first compatibility rules.
+    ///
+    /// If both legacy and new blobs partitions contain data, returns corruption.
+    /// If neither contains data, defaults to the new blobs partition.
+    async fn select_blob_partition(context: &E, cfg: &Config) -> Result<String, Error> {
```
roberto-bayardo, line 164: Our very first legacy cruft!

storage/src/journal/contiguous/fixed.rs
danlaine: Yup

storage/src/journal/contiguous/fixed.rs
danlaine: done

storage/src/journal/contiguous/fixed.rs
danlaine: 😭

storage/src/journal/contiguous/fixed.rs
danlaine: [Updated](https://github.com/commonwarexyz/monorepo/pull/2872/commits/e15f6ed71bba61a5557fbdbcd6dfffb714ad4d3c) comment on `metadata` -- how's this look?

storage/src/journal/contiguous/fixed.rs
roberto-bayardo: ping ^

storage/src/journal/contiguous/fixed.rs
danlaine: Updated this PR so that if we see existing blobs in `cfg.partition`, we keep using that directory to store blobs. Otherwise, we use `{cfg.partition}_blobs`.

storage/src/journal/contiguous/fixed.rs
danlaine: Added validation of oldest section length in [this](https://github.com/commonwarexyz/monorepo/pull/2872/commits/1f0e736ae2c6f008ebb6f65b0918c116b75ea45c) commit

storage/src/journal/contiguous/fixed.rs
```diff
@@ -464,36 +730,65 @@ impl<E: Storage + Metrics, A: CodecFixedShared> Journal<E, A> {
...
+            self.metadata.put(
+                PRUNING_BOUNDARY_KEY,
+                self.pruning_boundary.to_be_bytes().to_vec(),
+            );
+            self.metadata.sync().await?;
+        } else if self.metadata.get(&PRUNING_BOUNDARY_KEY).is_some() {
+            self.metadata.remove(&PRUNING_BOUNDARY_KEY);
+            self.metadata.sync().await?;
+        }
+
```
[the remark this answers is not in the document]
danlaine: > The clear_to_size function doesn't sync the inner journal before syncing/removing metadata, violating the stated invariant (lines 139-140) that metadata must be persisted AFTER the inner journal.

Calling `sync` on the removed blobs / `self.inner` isn't necessary because blob removal / creation is durable.

> If a crash occurs after inner.clear() and ensure_section_exists() but before the metadata removal completes, recovery can fail with a corruption error when old mid-section metadata references a higher section than the newly created blobs.

This is true -- we would return `Error::Corruption` on startup. The alternative (within the current design, at least) would be to remove the metadata first, and then the blobs. This is worse. Consider this execution:

* `items_per_blob == 5`
* `init_at_size(12)` called --> create section 2
* `clear_to_size(N)` called
* Crash after we remove metadata
* On `init`, oldest section is 2 and there is no data. This is interpreted as a journal with pruning boundary 10, size 10.

I think this result is worse than the existing behavior? In either case I'm not sure it matters a ton because this operation `clear_to_size` is creating an _empty journal_. If a crash results in a "corrupted" journal, but we're about to blow away that journal, I don't think it's an issue.

storage/src/journal/segmented/fixed.rs
```diff
@@ -341,31 +341,6 @@ impl<E: Storage + Metrics, A: CodecFixedShared> Journal<E, A> {
...
-        item_count: u64,
-    ) -> Result<(), Error> {
-        // Get or create the blob for this section
-        let blob = self.manager.get_or_create(section).await?;
-
-        // Calculate the target byte size
-        let target_size = item_count * Self::CHUNK_SIZE_U64;
-
-        // Resize grows the blob by appending zeros, which handles checksums properly
-        blob.resize(target_size).await?;
```
patrick-ogrady, line 367: 🔥

storage/src/journal/contiguous/fixed.rs
```diff
@@ -105,50 +123,116 @@ pub struct Config {
...
+        }
+    }
+
+    /// Select the blobs partition using legacy-first compatibility rules.
+    ///
+    /// If both legacy and new blobs partitions contain data, returns corruption.
+    /// If neither contains data, defaults to the new blobs partition.
+    async fn select_blob_partition(context: &E, cfg: &Config) -> Result<String, Error> {
+        let legacy_partition = cfg.partition.as_str();
+        let new_partition = format!("{}{}", cfg.partition, BLOB_SUFFIX);
```
patrick-ogrady: in other parts of the code, we do `{}-suffix` I believe now?

storage/src/journal/contiguous/fixed.rs
patrick-ogrady: 🙃

storage/src/journal/contiguous/fixed.rs
```diff
@@ -105,50 +123,116 @@ pub struct Config {
...
+        let blob_partition = Self::select_blob_partition(&context, &cfg).await?;
         let segmented_cfg = SegmentedConfig {
-            partition: cfg.partition,
+            partition: blob_partition,
             buffer_pool: cfg.buffer_pool,
             write_buffer: cfg.write_buffer,
         };
 
-        let mut inner = SegmentedJournal::init(context, segmented_cfg).await?;
+        let mut inner = SegmentedJournal::init(context.clone(), segmented_cfg).await?;
```
patrick-ogrady: named context?

storage/src/journal/contiguous/fixed.rs
```diff
@@ -174,39 +379,51 @@ impl<E: Storage + Metrics, A: CodecFixedShared> Journal<E, A> {
...
+        inner.ensure_section_exists(tail_section).await?;
 
-        // Initialize the tail section with zero-filled items if needed. This uses resize internally
-        // which appropriately uses the underlying page-oriented layout with checksums.
-        inner.init_section_at_size(tail_section, tail_items).await?;
-        inner.sync_all().await?;
+        // Initialize and populate metadata
+        let meta_cfg = MetadataConfig {
+            partition: format!("{}{}", cfg.partition, META_SUFFIX),
```
patrick-ogrady: see other comment about consistent partition naming format

storage/src/journal/contiguous/fixed.rs
```diff
@@ -174,39 +379,51 @@ impl<E: Storage + Metrics, A: CodecFixedShared> Journal<E, A> {
...
         // Initialize the segmented journal (empty)
         let segmented_cfg = SegmentedConfig {
-            partition: cfg.partition,
+            partition: blob_partition,
             buffer_pool: cfg.buffer_pool,
             write_buffer: cfg.write_buffer,
         };
 
-        let mut inner = SegmentedJournal::init(context, segmented_cfg).await?;
+        let mut inner = SegmentedJournal::init(context.clone(), segmented_cfg).await?;
```
patrick-ogrady: name this context

storage/src/journal/contiguous/fixed.rs
danlaine: Removed `const` and replaced usages with `{}-blobs` or `{}-metadata`

storage/src/journal/contiguous/fixed.rs
danlaine: Added `.with_label("blobs")`

storage/src/journal/contiguous/fixed.rs
danlaine: https://github.com/commonwarexyz/monorepo/pull/2872#discussion_r2714646336

fixed

storage/src/journal/contiguous/fixed.rs
danlaine: Added `.with_label("blobs")`

storage/src/journal/contiguous/fixed.rs
```diff
@@ -160,9 +240,137 @@ impl<E: Storage + Metrics, A: CodecFixedShared> Journal<E, A> {
...
+                        // if we crash during clear_to_size/init_at_size after blobs update but
+                        // before metadata update. Fall back to blob state.
+                        warn!(
+                            meta_oldest_section,
+                            oldest_section,
+                            "crash repair: metadata ahead of blobs, computing from blobs"
+                        );
+                        (blob_boundary, true)
+                    }
+                    _ => (meta_pruning_boundary, false), // valid mid-section metadata
```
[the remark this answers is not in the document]
danlaine: Good bot. Fixed in https://github.com/commonwarexyz/monorepo/pull/2872/commits/7a4298bcc0d4511fb829eee2585c0dbd43ead91d

storage/src/journal/contiguous/fixed.rs
```diff
@@ -464,36 +754,70 @@ impl<E: Storage + Metrics, A: CodecFixedShared> Journal<E, A> {
...
-        self.inner.prune(min_section).await
+        let pruned = self.inner.prune(min_section).await?;
+
+        // After pruning, update pruning_boundary to the start of the oldest remaining section
+        if pruned {
+            let new_oldest = self
+                .inner
+                .oldest_section()
+                .expect("all sections pruned - violates tail section invariant");
+            self.pruning_boundary = new_oldest * self.items_per_blob;
```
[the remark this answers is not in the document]
patrick-ogrady: Seems we should align these? Is either necessary?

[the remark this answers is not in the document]
danlaine: `pruned == true` means the pruning boundary has advanced beyond the section that the old `self.pruning_boundary` falls in, so we don't need to `max` because it's guaranteed that the new pruning boundary > the old one. I added a `debug_assert` to make this more explicit.

storage/src/journal/contiguous/fixed.rs
```diff
@@ -762,6 +762,8 @@ impl<E: Clock + Storage + Metrics, A: CodecFixedShared> Journal<E, A> {
                 .inner
                 .oldest_section()
                 .expect("all sections pruned - violates tail section invariant");
+            // Pruning boundary only moves forward
```
patrick-ogrady, line 766: I think we should assert on this?

storage/src/journal/contiguous/fixed.rs
```diff
@@ -105,50 +119,117 @@ pub struct Config {
...
     /// Total number of items appended (not affected by pruning).
     size: u64,
+
+    /// If the journal's pruning boundary is mid-section (that is, the oldest retained item's
+    /// position is not a multiple of `items_per_blob`), then the metadata stores the pruning
+    /// boundary. Otherwise, the metadata is empty.
+    ///
+    /// When the journal is pruned, `metadata` must be persisted AFTER the inner journal is
+    /// persisted to ensure that its pruning boundary is never after the inner journal's size.
+    // TODO(#2939): Remove metadata
```
patrick-ogrady, line 136: I think the goal is to retain metadata but remove the search for legacy?

storage/src/journal/contiguous/fixed.rs
```diff
@@ -160,9 +241,147 @@ impl<E: Storage + Metrics, A: CodecFixedShared> Journal<E, A> {
...
+                            meta_oldest_section,
+                            oldest_section, "crash repair: metadata stale, computing from blobs"
+                        );
+                        (blob_boundary, true)
+                    }
+                    Some(oldest_section) if meta_oldest_section > oldest_section => {
+                        // Metadata references a section ahead of the oldest blob. This can happen
+                        // if we crash during clear_to_size/init_at_size after blobs update but
+                        // before metadata update. Fall back to blob state.
+                        warn!(
```
patrick-ogrady, line 295: This only happens if that later blob is aligned?

patrick-ogrady: Is there a case where update metadata and then fallback to a blob with offset start?

storage/src/journal/contiguous/fixed.rs
danlaine: https://github.com/commonwarexyz/monorepo/issues/2942

Made issue and linked (above `fn select_blob_partition`)

storage/src/journal/contiguous/fixed.rs
danlaine: Changed `debug_assert!` to `assert!` since we use the latter everywhere else. Although perhaps we should consider changing some of those usages to `debug_assert!` ?

storage/src/journal/contiguous/fixed.rs
patrick-ogrady, line 1: Proposed issue:

```
› >   - [P2] Use pruning boundary when oldest_retained_pos is None — /Users/patrickogrady/code/monorepo-pr-2872/storage/src/journal/contiguous/fixed.rs:624-633
      With the new semantics, oldest_retained_pos returns None whenever pruning_boundary == size (e.g., after init_at_size with a mid-section start and no data). mmr/journaled.rs currently does
      journal.oldest_retained_pos().unwrap_or(0) when validating metadata, so a fully‑pruned journal that starts at a non‑zero position is now treated as if its pruning boundary were 0. If the MMR
  metadata
      partition is missing/reset, this silently accepts stale metadata and sets pruned_to_pos to 0, which can later allow reads/prunes below the real boundary and surface missing-node/ElementPruned
  errors.
      Consider propagating the real boundary (e.g., journal.pruning_boundary() or journal.size()) when oldest_retained_pos is None.
      ```

File: storage/src/journal/contiguous/fixed.rs (hunk 16 above)
Comment (patrick-ogrady) [reply storage/src/journal/contiguous/fixed.rs:766]: If it should never happen, we should fail loudly IMO.

File: storage/src/journal/contiguous/fixed.rs
Comment (patrick-ogrady) [reply storage/src/journal/contiguous/fixed.rs:1]: The LLMs like this:
```
diff --git a/storage/src/mmr/journaled.rs b/storage/src/mmr/journaled.rs
index 35bb5ba2a..8c60ac414 100644
--- a/storage/src/mmr/journaled.rs
+++ b/storage/src/mmr/journaled.rs
@@ -282,13 +282,19 @@ impl<E: RStorage + Clock + Metrics, D: Digest> CleanMmr<E, D> {
                     .expect("metadata prune position is not 8 bytes"),
             )
         });
-        let oldest_retained_pos = journal.oldest_retained_pos().unwrap_or(0);
+        let oldest_retained_pos = journal
+            .oldest_retained_pos()
+            .unwrap_or_else(|| journal.pruning_boundary());
         if metadata_prune_pos != oldest_retained_pos {
             assert!(metadata_prune_pos >= oldest_retained_pos);
             // These positions may differ only due to blob boundary alignment, so this case isn't
             // unusual.
             journal.prune(metadata_prune_pos).await?;
-            if journal.oldest_retained_pos().unwrap_or(0) != oldest_retained_pos {
+            if journal
+                .oldest_retained_pos()
+                .unwrap_or_else(|| journal.pruning_boundary())
+                != oldest_retained_pos
+            {
                 // This should only happen in the event of some failure during the last attempt to
                 // prune the journal.
                 warn!(
```

Comment (patrick-ogrady) [reply storage/src/journal/contiguous/fixed.rs:1]: still pondering

Comment (patrick-ogrady) [reply storage/src/journal/contiguous/fixed.rs:1]: Wondering if we should remove `oldest_retained_pos()` for `pruning_boundary()`? Seems to do the same thing?

File: storage/src/journal/contiguous/fixed.rs (hunk 16 above)
Comment (danlaine) [reply storage/src/journal/contiguous/fixed.rs:766]: Sure, just pointing out there _is_ some non-zero performance cost of `assert!`ing stuff and we might eventually decide that in some cases we may decide that we're comfortable enough with correctness that we only enable the assertions in non-prod builds.

File: storage/src/journal/contiguous/fixed.rs [hunk 19]
```diff
@@ -581,6 +909,113 @@ mod tests {
...
+            new_blob.sync().await.expect("Failed to sync new blob");
+
+            let result =
+                Journal::<_, Digest>::init(context.with_label("second"), cfg.clone()).await;
+            assert!(matches!(result, Err(Error::Corruption(_))));
+        });
+    }
+
+    #[test_traced]
+    fn test_fixed_journal_init_prefers_legacy_partition() {
```
patrick-ogrady, line 961: ❤️

storage/src/journal/contiguous/fixed.rs
patrick-ogrady: More extended scenario to consider:
```
diff --git a/storage/src/journal/contiguous/fixed.rs b/storage/src/journal/contiguous/fixed.rs
index 14a25422d..d41b64b86 100644
--- a/storage/src/journal/contiguous/fixed.rs
+++ b/storage/src/journal/contiguous/fixed.rs
@@ -53,7 +53,7 @@
 //!
 //! `Journal::init_sync` initializes a journal for state sync, handling existing data appropriately:
 //! - If no data exists, creates a journal at the sync range start
-//! - If data exists within range, prunes to the lower bound
+//! - If data exists within range, prunes toward the lower bound (section-aligned)
 //! - If data exceeds the range, returns an error
 //! - If data is stale (before range), destroys and recreates
 //!
@@ -457,7 +457,9 @@ impl<E: Clock + Storage + Metrics, A: CodecFixedShared> Journal<E, A> {
     /// Handles sync scenarios based on existing journal data vs. the given sync range:
     ///
     /// 1. **No existing data**: Creates journal at `range.start` (or empty if `range.start == 0`)
-    /// 2. **Data within range**: Prunes to `range.start` and reuses existing data
+    /// 2. **Data within range**: Prunes toward `range.start` and reuses existing data.
+    ///    Since prune only removes complete sections, some items before `range.start`
+    ///    may be retained (from the section boundary to `range.start - 1`).
     /// 3. **Data exceeds range**: Returns error
     /// 4. **Stale data**: Destroys and recreates at `range.start`
     pub(crate) async fn init_sync(
diff --git a/storage/src/journal/contiguous/variable.rs b/storage/src/journal/contiguous/variable.rs
index 63823ecb4..b3c72edfd 100644
--- a/storage/src/journal/contiguous/variable.rs
+++ b/storage/src/journal/contiguous/variable.rs
@@ -274,7 +274,8 @@ impl<E: Clock + Storage + Metrics, V: CodecShared> Journal<E, V> {
     /// - Stale (all data strictly before `range.start`): destroys existing data and returns an
     ///   empty journal.
     /// - Overlap within [`range.start`, `range.end`]:
-    ///   - Prunes to `range.start`
+    ///   - Prunes toward `range.start` (section-aligned, so some items before
+    ///     `range.start` may be retained)
     /// - Unexpected data beyond `range.end`: returns [crate::qmdb::Error::UnexpectedData].
     ///
     /// # Arguments
diff --git a/storage/src/mmr/journaled.rs b/storage/src/mmr/journaled.rs
index 35bb5ba2a..7bbd8e936 100644
--- a/storage/src/mmr/journaled.rs
+++ b/storage/src/mmr/journaled.rs
@@ -282,13 +282,18 @@ impl<E: RStorage + Clock + Metrics, D: Digest> CleanMmr<E, D> {
                     .expect("metadata prune position is not 8 bytes"),
             )
         });
-        let oldest_retained_pos = journal.oldest_retained_pos().unwrap_or(0);
-        if metadata_prune_pos != oldest_retained_pos {
-            assert!(metadata_prune_pos >= oldest_retained_pos);
-            // These positions may differ only due to blob boundary alignment, so this case isn't
-            // unusual.
+        let oldest_retained_pos = journal
+            .oldest_retained_pos()
+            .unwrap_or_else(|| journal.pruning_boundary());
+        if metadata_prune_pos > oldest_retained_pos {
+            // Metadata is ahead of journal (crashed before completing journal prune).
+            // Prune the journal to match metadata.
             journal.prune(metadata_prune_pos).await?;
-            if journal.oldest_retained_pos().unwrap_or(0) != oldest_retained_pos {
+            if journal
+                .oldest_retained_pos()
+                .unwrap_or_else(|| journal.pruning_boundary())
+                != oldest_retained_pos
+            {
                 // This should only happen in the event of some failure during the last attempt to
                 // prune the journal.
                 warn!(
@@ -296,8 +301,19 @@ impl<E: RStorage + Clock + Metrics, D: Digest> CleanMmr<E, D> {
                     metadata_prune_pos, "journal pruned to match metadata"
                 );
             }
+        } else if metadata_prune_pos < oldest_retained_pos {
+            // Metadata is stale (e.g., missing/corrupted while journal has valid state).
+            // Use the journal's state as authoritative.
+            warn!(
+                metadata_prune_pos,
+                oldest_retained_pos, "metadata stale, using journal pruning boundary"
+            );
         }
 
+        // Use the more restrictive (higher) pruning boundary between metadata and journal.
+        // This handles both cases: metadata ahead (crash during prune) and metadata stale.
+        let effective_prune_pos = std::cmp::max(metadata_prune_pos, oldest_retained_pos);
+
         let last_valid_size = PeakIterator::to_nearest_size(journal_size);
         let mut orphaned_leaf: Option<D> = None;
         if last_valid_size != journal_size {
@@ -331,7 +347,7 @@ impl<E: RStorage + Clock + Metrics, D: Digest> CleanMmr<E, D> {
             },
             hasher,
         )?;
-        let prune_pos = Position::new(metadata_prune_pos);
+        let prune_pos = Position::new(effective_prune_pos);
         Self::add_extra_pinned_nodes(&mut mem_mmr, &metadata, &journal, prune_pos).await?;
 
         let mut s = Self {
@@ -366,12 +382,12 @@ impl<E: RStorage + Clock + Metrics, D: Digest> CleanMmr<E, D> {
     ///
     /// 2. **Prune and Reuse**: range.start ≤ existing_size ≤ range.end
     ///    - Sets in-memory MMR size to `existing_size`
-    ///    - Prunes the journal to `range.start`
+    ///    - Prunes the journal toward `range.start` (section-aligned)
     ///
     /// 3. **Prune and Rewind**: existing_size > range.end
     ///    - Rewinds the journal to size `range.end`
     ///    - Sets in-memory MMR size to `range.end`
-    ///    - Prunes the journal to `range.start`
+    ///    - Prunes the journal toward `range.start` (section-aligned)
     pub async fn init_sync(
         context: E,
         cfg: SyncConfig<D>,
@@ -1755,4 +1771,106 @@ mod tests {
             sync_mmr.destroy().await.unwrap();
         });
     }
+
+    // Test that MMR init() handles stale metadata (lower pruning boundary than journal).
+    // Before the fix, this would panic with an assertion failure. After the fix, it
+    // returns a MissingNode error (which is expected when metadata is corrupted and
+    // pinned nodes are lost).
+    #[test_traced("WARN")]
+    fn test_journaled_mmr_init_stale_metadata_returns_error() {
+            let mut hasher = Standard::<Sha256>::new();
+
+            // Create an MMR with some data and prune it
+            let mut mmr = Mmr::init(context.with_label("init"), &mut hasher, test_config())
+                .await
+                .unwrap();
+
+            // Add 50 elements
+            for i in 0..50 {
+                mmr.add(&mut hasher, &test_digest(i)).await.unwrap();
+            }
+            mmr.sync().await.unwrap();
+
+            // Prune to position 20 (this stores pinned nodes in metadata for position 20)
+            let prune_pos = Position::new(20);
+            mmr.prune_to_pos(prune_pos).await.unwrap();
+            drop(mmr);
+
+            // Tamper with metadata to have a stale (lower) pruning boundary
+            let meta_cfg = MConfig {
+                partition: test_config().metadata_partition,
+                codec_config: ((0..).into(), ()),
+            };
+            let mut metadata =
+                Metadata::<_, U64, Vec<u8>>::init(context.with_label("meta_tamper"), meta_cfg)
+                    .await
+                    .unwrap();
+
+            // Set pruning boundary to 0 (stale)
+            let key = U64::new(PRUNE_TO_POS_PREFIX, 0);
+            metadata.put(key, 0u64.to_be_bytes().to_vec());
+            metadata.sync().await.unwrap();
+            drop(metadata);
+
+            // Reopen the MMR - before the fix, this would panic with assertion failure
+            // After the fix, it returns MissingNode error (pinned nodes for the lower
+            // boundary don't exist since they were pruned from journal and weren't
+            // stored in metadata at the lower position)
+            let result = CleanMmr::<_, Digest>::init(
+                context.with_label("reopened"),
+                &mut hasher,
+                test_config(),
+            )
+            .await;
+
+            match result {
+                Err(Error::MissingNode(_)) => {} // expected
+                Ok(_) => panic!("expected MissingNode error, got Ok"),
+                Err(e) => panic!("expected MissingNode error, got {:?}", e),
+            }
+        });
+    }
+
+    // Test that MMR init() handles the case where metadata pruning boundary is ahead
+    // of journal (crashed before journal prune completed). This should successfully
+    // prune the journal to match metadata.
+    #[test_traced("WARN")]
+    fn test_journaled_mmr_init_metadata_ahead() {
+            let mut hasher = Standard::<Sha256>::new();
+
+            // Create an MMR with some data
+            let mut mmr = Mmr::init(context.with_label("init"), &mut hasher, test_config())
+                .await
+                .unwrap();
+
+            // Add 50 elements
+            for i in 0..50 {
+                mmr.add(&mut hasher, &test_digest(i)).await.unwrap();
+            }
+            mmr.sync().await.unwrap();
+
+            // Prune to position 30 (this stores pinned nodes and updates metadata)
+            let prune_pos = Position::new(30);
+            mmr.prune_to_pos(prune_pos).await.unwrap();
+            let expected_root = mmr.root();
+            let expected_size = mmr.size();
+            drop(mmr);
+
+            // Reopen the MMR - should recover correctly with metadata ahead of
+            // journal boundary (metadata says 30, journal is section-aligned to 28)
+            let mmr = Mmr::init(context.with_label("reopened"), &mut hasher, test_config())
+                .await
+                .unwrap();
+
+            assert_eq!(mmr.pruned_to_pos(), prune_pos);
+            assert_eq!(mmr.size(), expected_size);
+            assert_eq!(mmr.root(), expected_root);
+
+            mmr.destroy().await.unwrap();
+        });
+    }
 }

```

danlaine: Applied the above diff

storage/src/journal/contiguous/variable.rs
```diff
@@ -667,86 +667,106 @@ impl<E: Storage + Metrics, V: CodecShared> Journal<E, V> {
...
-                // data journal, but crashed before persisting the offsets journal.
-                // We can recover if offsets.size() matches data_oldest_pos (proper pruning).
+                // Offsets journal is empty (pruning_boundary >= size) but data exists.
+                // Recoverable if same section (init_at_size + crash before first sync).
+                // Corruption if different sections (e.g., offsets partition was deleted).
                 let offsets_size = offsets.size();
-                if offsets_size != data_oldest_pos {
+                if offsets_size / items_per_section != data_first_section {
                     return Err(Error::Corruption(format!(
-                        "offsets journal empty: size ({offsets_size}) != data oldest pos ({data_oldest_pos})"
```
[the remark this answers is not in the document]
danlaine: Simplified in [this](https://github.com/commonwarexyz/monorepo/pull/2872/commits/95f2dd59aec83cdc6fc07dd0282fb7977ec77404) commit

storage/src/journal/contiguous/variable.rs
```diff
@@ -705,24 +705,18 @@ impl<E: Clock + Storage + Metrics, V: CodecShared> Journal<E, V> {
             Some(_) => {
                 // Both journals are pruned to the same position.
             }
-            None if data_oldest_pos > 0 => {
-                // Offsets journal is empty (pruning_boundary >= size) but data exists.
```
patrick-ogrady: Should we retain this comment?

danlaine: I feel like the error message/warn probably suffice but added this:
```
                // Offsets journal is empty but data journal isn't.
                // It should always be in the same section as the data journal, though.
```

PR #2881, opened 2026-01-20
[runtime] Introduce `IoBuf` types and update network and storage APIs

### LINKED ISSUE #2554: [storage] Take Advantage of `impl Buf` on Runtime Blob

andresilva:
This PR overhauls buffer management across the runtime, networking, and storage layers. The changes fall into three main areas: new buffer types, updated runtime trait signatures, and optimizations enabled by these changes.

The core addition is a family of buffer types in the runtime crate: `IoBuf` and `IoBufMut` for single buffers, plus `IoBufs` and `IoBufsMut` for potentially chunked data. These are concrete types rather than traits, giving the runtime control over future specialization (e.g. buffer pooling, io_uring buffer registration, etc.).

The `Sink` and `Stream` traits now use these types directly. `Sink::send` accepts `impl Into<IoBufs>`, allowing callers to pass single buffers or vectored data. More significantly, `Stream::recv` no longer takes a caller-provided buffer. Instead, it accepts a length and returns framework-allocated `IoBufs`. This matches how network I/O actually works: data arrives asynchronously, so the runtime needs buffers ready before data arrives. With this interfaces the runtime will be able to use buffer groups and other optimizations without exposing that complexity to callers.

Storage takes the opposite approach. `Blob::read_at` now accepts caller-provided `IoBufsMut` and returns the same buffer(s) filled with data. This makes sense because storage reads are pull-based: the caller knows exactly what offset and size they want. Caller-provided buffers enable reuse patterns where the same buffer is read into repeatedly, avoiding allocation churn in hot paths like journal replay or cache population.

The `p2p` crate gains an `EncodedData` type that pre-encodes a message once before broadcasting to multiple recipients. Previously, sending to N peers meant encoding the same payload N times. Now the encoding happens once and the pre-encoded bytes are cloned (cheaply, since `IoBuf` is reference-counted) to each peer's send queue. The `Message<P>` type also switches from `Bytes` to `IoBuf` for consistency. The p2p layer relies on `IoBuf` and `IoBufMut` to guarantee contiguous memory and does not use the chunked variants.

Throughout the storage layer, the argument order for `read_at` and `write_at` changes to put offset first, matching typical "read at offset into buffer" mental models. The buffered writer and append pool implementations now read directly into caller-provided buffers rather than allocating internally and copying.

While the interfaces accept chunked types (`IoBufs`/`IoBufsMut`), this PR does not implement actual vectored I/O. Chunked inputs are coalesced before being passed to the underlying I/O operations. Vectorization support (using readv/writev or io_uring scatter-gather) is left for future work.

The `StableBuf` type in commonware-utils is removed since `IoBuf`/`IoBufMut` now serve that purpose with better integration into the I/O stack.

Related #2883.
Closes #2554.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -346,12 +346,12 @@ fn test_storage_operations() {
             .expect("Failed to open blob");
 
         // Write data at offset
-        blob.write_at(vec![1, 2, 3, 4], 0)
+        blob.write_at(0, vec![1, 2, 3, 4])
             .await
             .expect("Failed to write");
 
         // Read data from offset
-        let data = blob.read_at(vec![0u8; 4], 0)
+        let data = blob.read_at(0, vec![0u8; 4])
             .await
             .expect("Failed to read");
 
@@ -410,7 +410,7 @@ fn test_corruption_recovery() {
             .unwrap();
 
         // Corrupt checksum or truncate data
-        blob.write_at(vec![0xFF; 4], size - 4).await.unwrap();
+        blob.write_at(size - 4, vec![0xFF; 4]).await.unwrap();
         blob.sync().await.unwrap();
 
         // Re-initialize and verify recovery
@@ -471,7 +471,7 @@ fn test_storage_conformance() {
 
         // Hash blob contents to verify format
         let (blob, size) = context.open(&partition, &name).await.unwrap();
-        let buf = blob.read_at(vec![0u8; size as usize], 0).await.unwrap();
+        let buf = blob.read_at(0, vec![0u8; size as usize]).await.unwrap().coalesce();
         let digest = hash(buf.as_ref());
 
         // Compare against known hash


runtime/src/lib.rs
```diff
@@ -48,6 +46,8 @@ cfg_if::cfg_if! {
         pub mod benchmarks;
     }
 }
+pub mod iobuf;
+pub use iobuf::{IoBuf, IoBufMut, IoBufs, IoBufsMut};
```
andresilva, line 52: I wonder if it makes sense to re-export `bytes::{Buf, ButMut}` since these are central to our own types.

runtime/src/lib.rs
```diff
@@ -48,6 +46,8 @@ cfg_if::cfg_if! {
         pub mod benchmarks;
     }
 }
+pub mod iobuf;
```
patrick-ogrady, line 49: Would this be better places in utils like StableBuf was? Or are these only ever implementable by Runtime?

runtime/src/lib.rs
```diff
@@ -614,19 +614,23 @@ pub trait Sink: Sync + Send + 'static {
...
-    /// Reads exactly the number of bytes that fit in the buffer.
+    /// Receive exactly `len` bytes from the stream.
+    ///
+    /// The runtime allocates the buffer and returns it as `IoBufs`.
     ///
     /// # Warning
     ///
     /// If the stream returns an error, partially read data may be discarded.
-    fn recv(&mut self, buf: impl BufMut + Send) -> impl Future<Output = Result<(), Error>> + Send;
+    fn recv(&mut self, len: u64) -> impl Future<Output = Result<IoBufs, Error>> + Send;
```
patrick-ogrady, line 635: I guess this can lead to multiple small allocations when doing varint reads?

runtime/src/iobuf.rs
```diff
@@ -0,0 +1,1499 @@
...
+//! - [`IoBufsMut`]: Container for one or more mutable buffers
+
+use bytes::{Buf, BufMut, Bytes, BytesMut};
+use std::{collections::VecDeque, ops::RangeBounds};
+
+/// Immutable byte buffer.
+///
+/// Cloning is cheap and does not copy.
+#[derive(Clone, Debug, Default, PartialEq, Eq)]
+pub struct IoBuf {
```
clabby, line 16: I think we should implement `PartialEq<&[u8]>` & `PartialEq<[u8]>`, there's a good few places where we can remove `.as_ref()` in equalities.

runtime/src/lib.rs
clabby: I agree.

runtime/src/lib.rs
andresilva: The plan is to implement `peek` on `Stream` to also avoid the async bound on small 1-byte reads (e.g. on commonware-stream when reading the message size varint). Also since this returns `IoBufs` (immutable) we'll be able to return a slice to the underlying buffer without any extra allocation.

runtime/src/lib.rs
andresilva: Right now the types are pretty general but soon they'll become tied to buffer pool and also some io-uring specific data (e.g. an `IoBuf` might contain an `u16` which is the registration index shared with the kernel through io-uring). That's why I decided to add them to runtime directly.

p2p/src/lib.rs
```diff
@@ -74,7 +76,7 @@ pub trait UnlimitedSender: Clone + Send + Sync + 'static {
     fn send(
         &mut self,
         recipients: Recipients<Self::PublicKey>,
-        message: impl Buf + Send,
+        message: impl Into<IoBufMut> + Send,
```
andresilva, line 77: The reason we take a mutable buffer here is so that in the future we can prepend data without allocating/copying.

p2p/src/authenticated/discovery/actors/router/ingress.rs
```diff
@@ -17,16 +23,16 @@ pub enum Message<P: PublicKey> {
...
     },
     /// Notify the router that a peer is no longer available.
     Release { peer: P },
-    /// Send a message to one or more recipients.
+    /// Send pre-encoded data to one or more recipients.
     Content {
         recipients: Recipients<P>,
-        channel: Channel,
-        message: Bytes,
+        /// Pre-encoded payload ready for transmission.
```
patrick-ogrady: nit: verbo comment

p2p/src/authenticated/discovery/actors/router/ingress.rs
```diff
@@ -74,21 +80,31 @@ impl<P: PublicKey> Messenger<P> {
...
     pub async fn content(
         &mut self,
         recipients: Recipients<P>,
         channel: Channel,
-        mut message: impl Buf + Send,
+        message: IoBufMut,
         priority: bool,
     ) -> Vec<P> {
-        let message = message.copy_to_bytes(message.remaining());
+        // Build Data and encode Payload::Data once for all recipients
```
patrick-ogrady, line 91: This is a massive savings?

p2p/src/authenticated/lookup/actors/router/ingress.rs
```diff
@@ -17,16 +23,16 @@ pub enum Message<P: PublicKey> {
...
     },
     /// Notify the router that a peer is no longer available.
     Release { peer: P },
-    /// Send a message to one or more recipients.
+    /// Send pre-encoded data to one or more recipients.
     Content {
         recipients: Recipients<P>,
-        channel: Channel,
-        message: Bytes,
+        /// Pre-encoded payload ready for transmission.
```
patrick-ogrady: unneccessary

p2p/src/utils/mux.rs
```diff
@@ -34,9 +33,12 @@ pub enum Error {
 }
 
 /// Parse a muxed message into its subchannel and payload.
-pub fn parse(mut bytes: Bytes) -> Result<(Channel, Bytes), CodecError> {
-    let subchannel: Channel = UInt::read(&mut bytes)?.into();
-    Ok((subchannel, bytes))
+///
+/// Takes an immutable `IoBuf` (received from p2p layer) and returns
```
patrick-ogrady: remove this

p2p/src/lib.rs
```diff
@@ -27,7 +27,9 @@ pub use types::{Address, Ingress};
 ///
 /// This message is guaranteed to adhere to the configuration of the channel and
 /// will already be decrypted and authenticated.
-pub type Message<P> = (P, Bytes);
+///
+/// Uses `IoBuf` (immutable) since received messages are read-only.
```
patrick-ogrady: nit: remove

runtime/src/network/iouring.rs
```diff
@@ -633,7 +633,7 @@ mod tests {
         let (buf1, buf2) = reader.await.unwrap();
 
         // Verify we got the right data
-        assert_eq!(buf1.as_slice(), &[1u8, 2, 3, 4, 5]);
-        assert_eq!(buf2.as_slice(), &[6u8, 7, 8, 9, 10]);
+        assert_eq!(buf1.chunk(), &[1u8, 2, 3, 4, 5]);
```
patrick-ogrady: is chunk guaranteed to give us this data? I thought it was only the next contiguous?

runtime/src/storage/audited.rs
```diff
@@ -79,26 +78,31 @@ pub struct Blob<B: crate::Blob> {
...
+        &self,
+        offset: u64,
+        buf: impl Into<IoBufsMut> + Send,
+    ) -> Result<IoBufsMut, Error> {
         let buf = buf.into();
+        let len = buf.len() as u64;
         self.auditor.event(b"read_at", |hasher| {
             hasher.update(self.partition.as_bytes());
             hasher.update(&self.name);
-            hasher.update(buf.as_ref());
```
patrick-ogrady, line 87: fair point lol

runtime/src/storage/mod.rs
```diff
@@ -728,4 +726,68 @@ pub(crate) mod tests {
             "Expected BlobVersionMismatch error"
         );
     }
+
+    /// Test that read_at returns the same buffer that was passed in (contract verification).
+    async fn test_read_at_returns_same_buffer<S>(storage: &S)
```
patrick-ogrady, line 756: not sure this is required...

patrick-ogrady: I suppose if we want to maintain this write into chunks thing, then its nice?

runtime/src/utils/buffer/pool/append.rs
```diff
@@ -301,8 +303,8 @@ impl<B: Blob> Append<B> {
                     let len = len as u64;
                     if len != page_size {
                         // The page is partial (logical data doesn't fill the page).
-                        let buf: Vec<u8> = buf.into();
-                        let logical_bytes = buf[..(len as usize)].to_vec();
+                        // TODO: avoid the copy here by returning IoBuf instead of Vec<u8>
```
patrick-ogrady: nit: should add a TODO

runtime/src/utils/buffer/pool/append.rs
patrick-ogrady, line 281: fwiw...I think we could return IoBuf here

runtime/src/utils/buffer/read.rs
```diff
@@ -100,22 +99,11 @@ impl<B: Blob> Read<B> {
...
-                .await?
-                .into();
-            self.buffer.as_mut()[0..bytes_to_read].copy_from_slice(&tmp_buffer[0..bytes_to_read]);
-        } else {
-            self.buffer = self
-                .blob
-                .read_at(std::mem::take(&mut self.buffer), self.blob_position)
-                .await?;
-        }
+        // Resize buffer and pass directly to read_at (avoids copy)
```
patrick-ogrady, line 102: nice

runtime/src/utils/buffer/write.rs
```diff
@@ -91,58 +87,90 @@ impl<B: Blob> Blob for Write<B> {
...
+                let remaining = buffer.extract(single.as_mut(), offset);
+                if remaining > 0 {
+                    let blob_buf = IoBufMut::zeroed(remaining);
+                    let blob_result = self.blob.read_at(offset, blob_buf).await?;
+                    single.as_mut()[..remaining].copy_from_slice(blob_result.coalesce().as_ref());
+                }
+                Ok(IoBufsMut::Single(single))
+            }
+            IoBufsMut::Chunked(mut chunks) => {
+                // Read into a temporary contiguous buffer
```
patrick-ogrady: This is just done to minimize the number of `read_at` calls?

runtime/src/iobuf.rs
```diff
@@ -0,0 +1,1633 @@
...
+    pub const fn is_single(&self) -> bool {
+        matches!(self, Self::Single(_))
+    }
+
+    /// Freeze into immutable `IoBufs`.
+    pub fn freeze(self) -> IoBufs {
+        match self {
+            Self::Single(buf) => IoBufs::Single(buf.freeze()),
+            Self::Chunked(bufs) => {
+                let mut result = IoBufs::default();
```
patrick-ogrady: +                let frozen: VecDeque<IoBuf> = bufs.into_iter().map(|b| b.freeze()).collect();
+                if frozen.len() == 1 {
+                    IoBufs::Single(frozen.into_iter().next().unwrap())
+                } else {
+                    IoBufs::Chunked(frozen)
                 }

runtime/src/iobuf.rs
```diff
@@ -0,0 +1,1633 @@
...
+                for buf in bufs.iter_mut() {
+                    let avail = buf.remaining_mut();
+                    if remaining <= avail {
+                        buf.advance_mut(remaining);
+                        return;
+                    }
+                    buf.advance_mut(avail);
+                    remaining -= avail;
+                }
+                panic!("advance_mut overflow");
```
patrick-ogrady: -                panic!("advance_mut overflow");
+                panic!("cannot advance past end of buffer");

runtime/src/lib.rs
```diff
@@ -714,21 +720,27 @@ pub trait Storage: Clone + Send + Sync + 'static {
...
+    /// The caller provides the buffer, and the implementation fills it with data
+    /// read from the blob starting at `offset`. Returns the same buffer, filled
+    /// with data.
+    ///
+    /// # Contract
     ///
-    /// `read_at` does not return the number of bytes read because it
-    /// only returns once the entire buffer has been filled.
+    /// - The output `IoBufsMut` is the same as the input, with data filled from offset
+    /// - The total bytes read equals the total capacity of the input buffer(s)
```
patrick-ogrady: -    /// - The total bytes read equals the total capacity of the input buffer(s)
+    /// - The total bytes read equals the total initialized length of the input buffer(s)

runtime/src/iobuf.rs
```diff
@@ -0,0 +1,1633 @@
...
+        let buf1 = IoBufMut::from(b"hello".as_ref());
+        let buf2 = IoBufMut::from(b" world".as_ref());
+        let bufs = IoBufsMut::from(vec![buf1, buf2]);
+        assert!(!bufs.is_single());
+        assert_eq!(bufs.len(), 11);
+        assert_eq!(bufs.chunk(), b"hello");
+    }
+
+    #[test]
+    fn test_iobufsmut_freeze() {
```
patrick-ogrady: #[test]
-    fn test_iobufsmut_freeze() {
+    fn test_iobufsmut_freeze_single() {
         let buf = IoBufMut::from(b"hello".as_ref());
         let bufs = IoBufsMut::from(buf);
         let frozen = bufs.freeze();
@@ -1207,6 +1210,28 @@ mod tests {
         assert_eq!(frozen.chunk(), b"hello");
     }

+    #[test]
+    fn test_iobufsmut_freeze_chunked() {
+        let buf1 = IoBufMut::from(b"hello".as_ref());
+        let buf2 = IoBufMut::from(b" world".as_ref());
+        let bufs = IoBufsMut::from(vec![buf1, buf2]);
+        let frozen = bufs.freeze();
+
+        // Should be chunked with exactly 2 buffers (no spurious empty buffer)
+        assert!(!frozen.is_single());
+        match frozen {
+            IoBufs::Chunked(ref chunks) => {
+                assert_eq!(chunks.len(), 2);
+                assert_eq!(chunks[0], b"hello");
+                assert_eq!(chunks[1], b" world");
+            }
:

runtime/src/storage/mod.rs
andresilva: Yeah, I made sure the invariant was respected in all the `Blob` implementations.

runtime/src/utils/buffer/pool/append.rs
andresilva: Avoided the copy altogether.

runtime/src/utils/buffer/write.rs
andresilva: In the future this will be replaced by vectored IO but I didn't want to do all of that in this PR.

runtime/src/network/iouring.rs
andresilva: I updated all the places in tests where we were using `.chunk()` to `.coalesce()`.

runtime/src/utils/buffer/pool/append.rs
andresilva: Yes, this entire thing will eventually be migrated to use `IoBuf`/`IoBufMut`, but I didn't want to do all those changes in this PR.

runtime/src/iobuf.rs
andresilva: Nice!

andresilva: We can also filter empty ones and collapse.

patrick-ogrady:
Recommend adding these tests (based on incorrect LLM feedback):

```
diff --git a/runtime/src/iobuf.rs b/runtime/src/iobuf.rs
index 8929f8594..8e874118f 100644
--- a/runtime/src/iobuf.rs
+++ b/runtime/src/iobuf.rs
@@ -1679,6 +1679,64 @@ mod tests {
         assert_eq!(chain_bytes.as_ref(), b"hello wo");
     }
 
+    #[test]
+    fn test_iobufmut_len_equals_remaining_after_advance() {
+        let mut buf = IoBufMut::from(b"hello world");
+
+        // Before advance
+        assert_eq!(buf.len(), buf.remaining());
+        assert_eq!(buf.as_ref(), buf.chunk());
+
+        // After partial advance
+        buf.advance(6);
+        assert_eq!(buf.len(), buf.remaining());
+        assert_eq!(buf.as_ref(), buf.chunk());
+        assert_eq!(buf.len(), 5);
+        assert_eq!(buf.as_ref(), b"world");
+
+        // After advancing to end
+        buf.advance(5);
+        assert_eq!(buf.len(), buf.remaining());
+        assert_eq!(buf.as_ref(), buf.chunk());
+        assert_eq!(buf.len(), 0);
+    }
+
+    #[test]
+    fn test_iobufsmut_len_equals_remaining_after_advance() {
+        let buf1 = IoBufMut::from(b"hello");
+        let buf2 = IoBufMut::from(b" world");
+        let mut bufs = IoBufsMut::from(vec![buf1, buf2]);
+
+        // Before advance
+        assert_eq!(bufs.len(), bufs.remaining());
+        assert_eq!(bufs.len(), 11);
+
+        // After partial advance (within first buffer)
+        bufs.advance(3);
+        assert_eq!(bufs.len(), bufs.remaining());
+        assert_eq!(bufs.len(), 8);
+
+        // After advance past first buffer
+        bufs.advance(4);
+        assert_eq!(bufs.len(), bufs.remaining());
+        assert_eq!(bufs.len(), 4);
+    }
+
+    #[test]
+    fn test_iobuf_len_equals_remaining_after_advance() {
+        let mut buf = IoBuf::from(b"hello world");
+
+        // Before advance
+        assert_eq!(buf.len(), buf.remaining());
+        assert_eq!(buf.as_ref(), buf.chunk());
+
+        // After advance
+        buf.advance(6);
+        assert_eq!(buf.len(), buf.remaining());
+        assert_eq!(buf.as_ref(), buf.chunk());
+        assert_eq!(buf.len(), 5);
+    }
+
     #[cfg(feature = "arbitrary")]
     mod conformance {
         use super::IoBuf;

```

patrick-ogrady:
A few more:

```
diff --git a/runtime/src/iobuf.rs b/runtime/src/iobuf.rs
index 2cb1803f8..d529c941d 100644
--- a/runtime/src/iobuf.rs
+++ b/runtime/src/iobuf.rs
@@ -1749,6 +1749,52 @@ mod tests {
         assert_eq!(bufs.len(), 4);
     }
 
+    #[test]
+    fn test_iobufsmut_freeze_after_advance() {
+        let buf1 = IoBufMut::from(b"hello");
+        let buf2 = IoBufMut::from(b" world");
+        let mut bufs = IoBufsMut::from(vec![buf1, buf2]);
+
+        // Advance partway through first buffer
+        bufs.advance(3);
+        assert_eq!(bufs.len(), 8);
+
+        // Freeze and verify only remaining data is preserved
+        let frozen = bufs.freeze();
+        assert_eq!(frozen.len(), 8);
+        assert_eq!(frozen.coalesce(), b"lo world");
+    }
+
+    #[test]
+    fn test_iobufsmut_freeze_after_advance_to_boundary() {
+        let buf1 = IoBufMut::from(b"hello");
+        let buf2 = IoBufMut::from(b" world");
+        let mut bufs = IoBufsMut::from(vec![buf1, buf2]);
+
+        // Advance exactly to first buffer boundary
+        bufs.advance(5);
+        assert_eq!(bufs.len(), 6);
+
+        // First buffer should be fully consumed (empty after advance)
+        // freeze() filters empty buffers, so result should be Single
+        let frozen = bufs.freeze();
+        assert!(frozen.is_single());
+        assert_eq!(frozen.coalesce(), b" world");
+    }
+
+    #[test]
+    fn test_iobufsmut_coalesce_after_advance_to_boundary() {
+        let buf1 = IoBufMut::from(b"hello");
+        let buf2 = IoBufMut::from(b" world");
+        let mut bufs = IoBufsMut::from(vec![buf1, buf2]);
+
+        // Advance exactly past first buffer
+        bufs.advance(5);
+
+        // Coalesce should only include second buffer's data
+        assert_eq!(bufs.coalesce(), b" world");
+    }
+
     #[cfg(feature = "arbitrary")]
     mod conformance {
         use super::IoBuf;
```

runtime/src/iobuf.rs
```diff
@@ -0,0 +1,1807 @@
...
+    fn default() -> Self {
+        Self::Single(IoBuf::default())
+    }
+}
+
+impl IoBufs {
+    /// Number of bytes remaining to read.
+    #[inline]
+    pub fn len(&self) -> usize {
+        self.remaining()
```
patrick-ogrady, line 426: I find it a bit odd that `IoBufs` calls .remaining() for `.len()` but `IoBuf` doesn't?

Meanwhile, we compute this manually in `IoBufsMut`?

stream/src/utils/codec.rs
```diff
@@ -1,40 +1,40 @@
...
-pub async fn recv_frame<T: Stream>(stream: &mut T, max_message_size: u32) -> Result<Bytes, Error> {
+pub async fn recv_frame<T: Stream>(stream: &mut T, max_message_size: u32) -> Result<IoBufs, Error> {
     // Read and decode the varint length prefix byte-by-byte
     let mut decoder = Decoder::<u32>::new();
-    let mut buf = [0u8; 1];
     let len = loop {
-        stream.recv(&mut buf[..]).await.map_err(Error::RecvFailed)?;
-        match decoder.feed(buf[0]) {
+        let buf = stream.recv(1).await.map_err(Error::RecvFailed)?;
+        match decoder.feed(buf.chunk()[0]) {
```
patrick-ogrady: this looks weird to me (the `.chunk()[0]` is a bit tricksy) but I don't think it is wrong.

runtime/src/iobuf.rs
andresilva: We must implement `remaining` for `IoBufs` in the `Buf` trait, `Bytes` does the same (https://docs.rs/bytes/latest/src/bytes/bytes.rs.html#696-698), they just do it in the opposite direction.

patrick-ogrady: IMHO we should have all of .len() rely on remaining() (or the other way around)?

```
diff --git a/runtime/src/iobuf.rs b/runtime/src/iobuf.rs
index d529c941d..ab7337af9 100644
--- a/runtime/src/iobuf.rs
+++ b/runtime/src/iobuf.rs
@@ -28,16 +28,16 @@ impl IoBuf {
         }
     }
 
-    /// Number of bytes in the buffer.
+    /// Number of bytes remaining in the buffer.
     #[inline]
-    pub const fn len(&self) -> usize {
-        self.inner.len()
+    pub fn len(&self) -> usize {
+        self.remaining()
     }
 
     /// Whether the buffer is empty.
     #[inline]
-    pub const fn is_empty(&self) -> bool {
-        self.inner.is_empty()
+        self.remaining() == 0
     }
 
     /// Get raw pointer to the buffer data.
@@ -230,13 +230,13 @@ impl IoBufMut {
     /// Number of initialized bytes.
     #[inline]
     pub fn len(&self) -> usize {
-        self.inner.len()
+        self.remaining()
     }
 
     /// Whether the buffer is empty.
     #[inline]
     pub fn is_empty(&self) -> bool {
-        self.inner.is_empty()
+        self.remaining() == 0
     }
 
     /// Freeze into immutable `IoBuf`.
@@ -625,16 +625,15 @@ impl Default for IoBufsMut {
 
 impl IoBufsMut {
     /// Total initialized length across all buffers.
+    #[inline]
     pub fn len(&self) -> usize {
-        match self {
-            Self::Single(buf) => buf.len(),
-            Self::Chunked(bufs) => bufs.iter().map(|b| b.len()).fold(0, usize::saturating_add),
-        }
+        self.remaining()
     }
 
     /// Whether all buffers are empty.
+    #[inline]
     pub fn is_empty(&self) -> bool {
-        self.len() == 0
+        self.remaining() == 0
     }
 
     /// Whether this contains a single contiguous buffer.

```

andresilva: Good point about `IoBufsMut`, we can do the same thing for consistency, i.e. just define `len` through `remaining` (or vice-versa).

stream/src/utils/codec.rs
andresilva: I kept `chunk()` here because we only request a byte, so even if the respose were chunked it would be guaranteed to have our byte in the first chunk. We can `coalesce` though.

storage/src/metadata/storage.rs
```diff
@@ -154,8 +154,10 @@ impl<E: Clock + Storage + Metrics, K: Span, V: Codec> Metadata<E, K, V> {
         }
 
         // Read blob
-        let len = len.try_into().map_err(|_| Error::BlobTooLarge(len))?;
```
danlaine, line 157: Looks like `Error::BlobTooLarge` is now unused -- should we remove it?

stream/src/utils/codec.rs
andresilva: Replaced with `buf.get_u8()`.

storage/src/metadata/storage.rs
andresilva: Good catch. Removed.

andresilva:
@danlaine Updated `AGENTS.md` (`copilot-instructions.md` is just a symlink).

PR #2882, opened 2026-01-20
docs: fix comment mismatches in broadcast, storage, and resolver

zeroprooff:
Fixed a few doc comments that didn't match the actual code - Subscribe variant was missing some params in its description, NextStep was being called StepResult, and Engine was referred to as Actor. Just cleaning up the docs to match what the code actually does.

### DIFF
--- a/broadcast/src/buffered/ingress.rs
+++ b/broadcast/src/buffered/ingress.rs
@@ -16,7 +16,7 @@ pub enum Message<P: PublicKey, M: Committable + Digestible> {
         responder: oneshot::Sender<Vec<P>>,
     },
 
-    /// Subscribe to a message by peer (optionally), commitment, and digest (optionally).
+    /// Subscribe to a message by commitment.
     ///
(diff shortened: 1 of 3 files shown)


broadcast/src/buffered/ingress.rs
```diff
@@ -16,7 +16,7 @@ pub enum Message<P: PublicKey, M: Committable + Digestible> {
         responder: oneshot::Sender<Vec<P>>,
     },
 
-    /// Subscribe to receive a message by digest.
+    /// Subscribe to a message by peer (optionally), commitment, and digest (optionally).
```
andresilva: This seems a bit too verbose.

andresilva (dismissed):

broadcast/src/buffered/ingress.rs
zeroprooff: Fixed, thanks!

andresilva (approved):

clabby:

PR #2888, opened 2026-01-21
 add impl write

malik672:
add write impl for append 
@roberto-bayardo


clabby:

PR #2891, opened 2026-01-21
[marshaled] Remove Duplicate Verification

patrick-ogrady:


consensus/src/application/marshaled.rs
```diff
@@ -655,8 +622,18 @@ where
                     }
                 };
 
+                // Check if this is a re-proposal (the block commitment matches the parent from
+                // the context).
                 let context = block.context();
-                let verify_rx = marshaled.verify(context, block).await;
+                if context.parent.1 == block.commitment() {
```
patrick-ogrady: This is the wrong check.

PR #2893, opened 2026-01-21
consensus: optimize commitment-to-height lookup in marshal

meetrick:
## Motivation
Resolves the TODO in `consensus/src/marshal/actor.rs`:

```rust
// TODO: Instead of pulling out the entire block, determine the
// height directly from the archive by mapping the commitment to
// the index, which is the same as the height.
```
When looking up block info by commitment, the marshal actor was loading the entire block just to extract its height. Since `height == index` in the archive, we can retrieve the index directly without loading the block data.

### New API

- **`Archive::index_of(key)`**: Returns the index associated with a key without loading the full value
- **`Blocks::height_of(commitment)`**: Wrapper that converts index to Height

### Implementation
| Archive Type | Strategy |
|--------------|----------|
| **prunable** | Uses existing in-memory `keys` map for O(1) lookup |
| **immutable** | Returns `None` (falls back to loading full block) |

#### Why only prunable is optimized?
- `prunable::Archive` already maintains `keys: MultiMap<K, u64>` in memory, making optimization straightforward
- `immutable::Archive` would require additional memory (~40 bytes/block) to maintain a reverse mapping, which conflicts with its "minimal memory footprint" design goal

### New Benchmark Added
Added `storage/src/archive/benches/index_of.rs` to compare `index_of` vs `get` performance for key-to-index lookups.

Note: Only `prunable::Archive` is benchmarked because `immutable::Archive` does not support the `index_of` optimization.

```bash
cargo bench -p commonware-storage --bench archive -- "index_of"
```

### Results: `index_of` vs `get` (1,000 key lookups)
| Archive | `get` | `index_of` | Improvement |
|---------|-------|------------|-------------|
| prunable (no compression) | 97.08 ms | 0.19 ms | **514x faster** |
| prunable (compression=3) | 99.58 ms | 0.19 ms | **520x faster** |

## Testing
- Added `test_index_of_*` unit tests for both archive types (with/without compression)
- All existing marshal tests pass


storage/src/archive/immutable/storage.rs
```diff
@@ -97,6 +97,9 @@ pub struct Archive<E: Storage + Metrics + Clock, K: Array, V: CodecShared> {
     /// Ordinal for the archive.
     ordinal: Ordinal<E, Cursor>,
 
+    /// Reverse mapping from cursor to index for efficient key→index lookups.
+    cursor_to_index: HashMap<Cursor, u64>,
```
patrick-ogrady: This would, unfortunately, use a ton of extra memory 😢

storage/src/freezer/storage.rs
```diff
@@ -947,6 +947,47 @@ impl<E: Storage + Metrics + Clock, K: Array, V: CodecShared> Freezer<E, K, V> {
...
+            // Follow the chain
+            let Some(next) = key_entry.next() else {
+                break; // End of chain
+            };
+            section = next.0;
+            position = next.1;
+        }
+
+        Ok(None)
+    }
```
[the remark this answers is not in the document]
meetrick: Intentional trade-off. The methods have different return types (V vs Cursor), and extracting a shared helper would add complexity for a straightforward loop. The duplication is localized and the performance gain justifies it.

storage/src/archive/immutable/storage.rs
meetrick: @patrick-ogrady Fair point. I thought the 11x speedup justified the memory cost, but I understand immutable::Archive prioritizes minimal memory footprint.

If the memory overhead is a concern, I'm happy to revert the immutable optimization and keep it as a fallback (first commit state). prunable would still benefit from 500x improvement with no additional memory.

What's your preference?

meetrick: Removed patch for immutable. Could you review again?

consensus/src/marshal/store.rs
```diff
@@ -170,6 +170,23 @@ pub trait Blocks: Send + Sync + 'static {
...
+    ///
+    /// * `commitment`: The block commitment to look up.
+    ///
+    /// # Returns
+    ///
+    /// `Ok(Some(height))` if the commitment exists, `Ok(None)` if not found, or `Err` on read failure.
+    fn height_of(
+        &self,
+        commitment: &<Self::Block as Committable>::Commitment,
+    ) -> impl Future<Output = Result<Option<Height>, Self::Error>> + Send;
```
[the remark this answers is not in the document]
meetrick: solved it

clabby:

Issue #2895, opened 2026-01-21
storage/fuzz/fuzz_targets/proofs_malleability.rs panics:171 after refactoring


dnkolegov-ar:
After https://github.com/commonwarexyz/monorepo/pull/2853 [proofs_malleability.rs#L171](https://github.com/commonwarexyz/monorepo/blob/main/storage/fuzz/fuzz_targets/proofs_malleability.rs#L171) can trigger a panic

```
thread '<unnamed>' (7001287) panicked at storage/src/mmr/mem.rs:451:9:
range start Location(0) >= leaf count Location(0)
```

roberto-bayardo:
It is weird we panic on out of bounds range but return errors for empty range, pruned range, and locaiton overflow.  WDYT @danlaine ?

PR #2896, opened 2026-01-21
Fix issue 2895: validate range data

dnkolegov-ar:
This PR fixes https://github.com/commonwarexyz/monorepo/issues/2895


roberto-bayardo:
Thanks, I wonder if we should just change range_proof to return an error for this case given it returns errors for other bad range conditions?  cc @danlaine

patrick-ogrady:
Ah, will defer to roberto/dan here...disregard my accept

danlaine:
> Thanks, I wonder if we should just change range_proof to return an error for this case given it returns errors for other bad range conditions? cc @danlaine

Agreed, made https://github.com/commonwarexyz/monorepo/pull/2899 to resolve this issue

danlaine:
Closing in favor of #2899

Issue #2883, opened 2026-01-20
[runtime, storage, p2p] High-performance I/O: zero-copy and io_uring integration


andresilva:
This meta-issue tracks a comprehensive redesign of the runtime's I/O layer to maximize throughput and minimize latency by fully exploiting io_uring capabilities and eliminating unnecessary copies.

Goals:
- Pool-backed buffers to eliminate allocator pressure on I/O paths
- Fully exploit io_uring: registered buffers, registered fds, buffer groups, multishot recv
- Push-based network receive with data already buffered when `recv()` is called
- Zero-copy paths from kernel to userspace and between layers
- Reduce copies to the unavoidable minimum (encryption boundaries)
- Graceful degradation on non-io_uring backends

Design documents:
- https://gist.github.com/andresilva/d4c9f03bcde25fd02e3c43f4d1bf68f0#file-1-iobuf-api-md - Buffer types, pool management, io_uring integration, runtime traits
- https://gist.github.com/andresilva/d4c9f03bcde25fd02e3c43f4d1bf68f0#file-2-copy-minimization-md - Copy minimization strategies, send/receive path analysis

PR #2906, opened 2026-01-22
[runtime] Add `Stream::peek` for non-consuming buffer inspection

andresilva:
This PR adds a `peek` method to the `Stream` trait that returns a slice of buffered data without consuming it. The method is synchronous and performs no I/O, making it useful for parsing length prefixes or inspecting incoming data before committing to a read.

The primary motivation is optimizing `recv_frame` in the stream crate. Previously, the varint length prefix was decoded by reading one byte at a time. Now it first attempts to decode from the `peek` buffer, avoiding async overhead when data is already buffered. If the `peek` buffer is empty or contains an incomplete varint, it falls back to byte-by-byte reading, reusing the same decoder instance.

Related #2883.

### DIFF
--- a/codec/src/varint.rs
+++ b/codec/src/varint.rs
@@ -760,6 +760,26 @@ mod tests {
         test_single_bits::<u128>();
     }
 
+    #[test]
+    fn test_max_varint_size_constants() {
+        let mut buf = Vec::new();
+
+        write(u16::MAX, &mut buf);
+        assert_eq!(buf.len(), MAX_U16_VARINT_SIZE);
+        buf.clear();
+
+        write(u32::MAX, &mut buf);
+        assert_eq!(buf.len(), MAX_U32_VARINT_SIZE);
+        buf.clear();
+
+        write(u64::MAX, &mut buf);
+        assert_eq!(buf.len(), MAX_U64_VARINT_SIZE);
+        buf.clear();
+
+        write(u128::MAX, &mut buf);
+        assert_eq!(buf.len(), MAX_U128_VARINT_SIZE);
+    }
+
     #[cfg(feature = "arbitrary")]
     mod conformance {
         use super::*;
--- a/runtime/src/mocks.rs
+++ b/runtime/src/mocks.rs
@@ -80,8 +80,8 @@ impl SinkTrait for Sink {
                 .as_ref()
                 .is_some_and(|(requested, _)| *requested <= channel.buffer.len())
             {
-                let (requested, os_send) = channel.waiter.take().unwrap();
                 // Send up to read_buffer_size bytes (but at least requested amount)
+                let (requested, os_send) = channel.waiter.take().unwrap();
                 let send_amount = channel
                     .buffer
                     .len()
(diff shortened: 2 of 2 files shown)


codec/src/varint.rs
```diff
@@ -53,6 +53,22 @@ const DATA_BITS_MASK: u8 = 0x7F;
 /// if clear this is the final byte.
 const CONTINUATION_BIT_MASK: u8 = 0x80;
 
+/// Maximum encoded size of a u16 varint in bytes.
+pub const MAX_U16_VARINT_SIZE: usize =
+    (size_of::<u16>() * BITS_PER_BYTE).div_ceil(DATA_BITS_PER_BYTE);
```
andresilva, line 58: We could have a const function for this, the problem is that it would be possible to call it with e.g.`max_size_for_varint<Vec<u8>>` which is wrong. There's no trait that all numeric types share to restrict the `T`.

runtime/src/network/iouring.rs
```diff
@@ -636,4 +652,60 @@ mod tests {
         assert_eq!(buf1.coalesce(), &[1u8, 2, 3, 4, 5]);
         assert_eq!(buf2.coalesce(), &[6u8, 7, 8, 9, 10]);
     }
+
+    #[tokio::test]
+    async fn test_peek_with_buffered_data() {
+        use crate::{Listener as _, Network as _, Sink as _, Stream as _};
```
patrick-ogrady: import at top of mod

codec/src/varint.rs
```diff
@@ -53,6 +53,22 @@ const DATA_BITS_MASK: u8 = 0x7F;
 /// if clear this is the final byte.
 const CONTINUATION_BIT_MASK: u8 = 0x80;
 
+/// Maximum encoded size of a u16 varint in bytes.
+pub const MAX_U16_VARINT_SIZE: usize =
```
patrick-ogrady, line 57: Can we tests these consts with `u16::MAX`, etc?

runtime/src/mocks.rs
```diff
@@ -66,7 +81,12 @@ impl SinkTrait for Sink {
                 .is_some_and(|(requested, _)| *requested <= channel.buffer.len())
             {
                 let (requested, os_send) = channel.waiter.take().unwrap();
-                let data = channel.buffer.copy_to_bytes(requested);
+                // Send up to read_buffer_size bytes (but at least requested amount)
```
patrick-ogrady, line 83: nit: put comment one line up IMO

runtime/src/mocks.rs
```diff
@@ -280,4 +320,117 @@ mod tests {
...
+            context.sleep(Duration::from_millis(10)).await;
+
+            // Send more than capacity
+            sink.send(b"ABCDEFGHIJKLMNOP".as_slice()).await.unwrap();
+
+            // Recv should get its 3 bytes
+            let received = recv_handle.await.unwrap();
+            assert_eq!(received.coalesce(), b"ABC");
+        });
+    }
```
patrick-ogrady, line 465: claude thinks this will better exercise the waiter code path:

```
  #[test]
  fn test_peek_after_recv_wakeup() {
      let (mut sink, mut stream) = Channel::init_with_read_buffer_size(64);

          // Spawn recv that will block waiting
          let (tx, rx) = futures::channel::oneshot::channel();
          let recv_handle = context.clone().spawn(|_| async move {
              let data = stream.recv(3).await.unwrap();
              tx.send(stream).ok();
              data
          });

          // Let recv set up waiter
          context.sleep(Duration::from_millis(10)).await;

          // Send more than requested
          sink.send(b"ABCDEFGHIJ".as_slice()).await.unwrap();

          // Recv gets its 3 bytes
          let received = recv_handle.await.unwrap();
          assert_eq!(received.coalesce(), b"ABC");

          // Get stream back and verify peek sees remaining data
          let stream = rx.await.unwrap();
          assert_eq!(stream.peek(100), b"DEFGHIJ");
      });
  }
```

PR #2909, opened 2026-01-22
[codec] Share cache across cloned `Lazy` instances

clabby:
## Overview

Updates `Lazy` to share the cached value across cloned instances, when a `Lazy` that was constructed with `deferred` was cloned several times before the value was populated.


cronokirby:
I thought about this, and I think `Arc<Lazy<T>>` is a better fit for doing this, because `Lazy<T>` is of supposed to have the semantics of `T`, essentially. For example, this would preclude having a `get_mut` function (at least, without locking)

PR #2918, opened 2026-01-23
[consensus/simplex] Votes DB abstraction

SuperFluffy:
This patch makes the voter actor generic over the kind of storage it uses. It follows the same approach as previously employed for the marshal actor.

Related https://github.com/commonwarexyz/monorepo/pull/2293


consensus/src/simplex/store.rs
```diff
@@ -0,0 +1,176 @@
...
+use commonware_storage::journal::{self, segmented::variable::Journal};
+use commonware_utils::NZUsize;
+use futures::{Stream, TryStreamExt};
+use std::{error::Error, future::Future, num::NonZeroUsize, pin::Pin};
+
+/// Durable store for consensus [Artifacts](Artifact) keyed by view.
+///
+/// This trait abstracts the storage of votes and certificates in the voter actor,
+/// enabling different storage backends (e.g., in-memory for testing, on-disk for production).
+pub trait Votes: Send + Sync + 'static {
```
SuperFluffy, line 16: Design consideration: by requiring `&mut self` on all trait methods makes it maximally flexible. On the one hand, the current `Journal` requires `&mut` for many of its methods and we don't want to use (async) Mutex every time we access them. Also, many crates implement clients to take `&mut self` (as is the case with tonic), and make client objects cheaply cloneable.

It seems it makes sense that this trait method then uses `&mut self` on all methods.

consensus/src/simplex/store.rs
SuperFluffy, line 1: TODO: does the Votes trait have to take into account epochs? This design assumes that a new store is passed to the simplex engine on every epoch (in line with the journal implementation). But this becomes strange when passing around clients to an external database. We could always wrap them to make them per-epoch.

clabby:

PR #2921, opened 2026-01-23
[cryptography/ed25519] canonicalize public keys

### LINKED ISSUE #2919: [Low] Ed25519 public key canonicalization
For ed25519, `PublicKey::read_cfg` stores the raw input bytes unchanged, and `PublicKey` derives `Eq/Ord/Hash` on that raw encoding. `ed25519-consensus` explicitly accepts non‑canonical encodings for public keys (ZIP‑215). So the same curve point can be represented by multiple encodings, and the current codebase will treat them as distinct identities.

```
#[test]
    fn ed25519_public_key_allows_multiple_encodings_of_same_point() {
        use commonware_codec::DecodeExt;
        use commonware_utils::ordered::Set;
        use curve25519_dalek::edwards::CompressedEdwardsY;

        use crate::ed25519::PublicKey;

        // Canonical encoding of identity: y = 1, sign bit = 0.
        let canonical: [u8; 32] = [
            1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
            0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
        ];

        // Non-canonical encoding of the same y: 1 + (2^255 - 19) = 2^255 - 18.
        let non_canonical: [u8; 32] = [
            0xee, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
            0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0x7f,
        ];

        // Both encodings decompress to the same curve point.
        let p1 = CompressedEdwardsY(canonical).decompress().unwrap();
        let p2 = CompressedEdwardsY(non_canonical).decompress().unwrap();
        assert_eq!(p1, p2);

        let pk1 = PublicKey::decode(canonical.as_ref()).unwrap();
        let pk2 = PublicKey::decode(non_canonical.as_ref()).unwrap();

        // Different identities because raw bytes differ.
        assert_ne!(pk1, pk2);

        let set: Set<PublicKey> = [pk1, pk2].try_into().unwrap();
        assert_eq!(set.len(), 2);
    }
```

andresilva:
Fixes #2919.

### DIFF
--- a/cryptography/src/ed25519/scheme.rs
+++ b/cryptography/src/ed25519/scheme.rs
@@ -464,6 +464,10 @@ mod tests {
     use commonware_codec::{DecodeExt, Encode};
     use commonware_math::algebra::Random;
     use commonware_utils::test_rng;
+    use std::{
+        cmp::Ordering,
+        hash::{BuildHasher, RandomState},
+    };
 
     fn test_sign_and_verify(
         private_key: PrivateKey,
@@ -880,10 +884,6 @@ mod tests {
 
     #[test]
     fn test_non_canonical_encoding_treated_as_same_identity() {
-        use core::hash::BuildHasher;
-        use curve25519_dalek::edwards::CompressedEdwardsY;
-        use std::hash::RandomState;
-
         // Canonical encoding of identity: y = 1, sign bit = 0.
         let canonical: [u8; 32] = [
             1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
@@ -913,7 +913,7 @@ mod tests {
         assert_eq!(pk1, pk2);
 
         // Verify ordering is consistent.
-        assert_eq!(pk1.cmp(&pk2), core::cmp::Ordering::Equal);
+        assert_eq!(pk1.cmp(&pk2), Ordering::Equal);
 
         // Verify hashes are equal.
         let hash_builder = RandomState::new();


cryptography/src/ed25519/scheme.rs
```diff
@@ -401,6 +439,23 @@ impl Batch {
...
+/// point.
+///
+/// Ed25519 public keys can have multiple valid encodings for the same curve point due to
+/// non-canonical y-coordinates (ZIP-215). This function ensures a consistent canonical
+/// representation.
+///
+/// Returns `None` if the bytes do not represent a valid curve point.
+fn canonicalize_public_key(bytes: [u8; PUBLIC_KEY_LENGTH]) -> Option<[u8; PUBLIC_KEY_LENGTH]> {
+    Some(
+        CompressedEdwardsY(bytes)
```
patrick-ogrady, line 452: We may be able to get this by just using `TryFrom` and then re-encoding using ed25519-consensus?

https://docs.rs/ed25519-consensus/2.1.0/src/ed25519_consensus/verification_key.rs.html#163

patrick-ogrady: (assuming we want to go this route)

cryptography/src/ed25519/scheme.rs
```diff
@@ -131,12 +132,43 @@ impl arbitrary::Arbitrary<'_> for PrivateKey {
 }
 
 /// Ed25519 Public Key.
-#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
+///
+/// The `raw` field preserves the original encoding for serialization, while the
```
cronokirby, line 136: We should not mention private fields in public documentation. We want to adjust the wording on this paragraph a bit to talk about things at a higher level, e.g. saying that public keys use a specific algorithm for comparing equality, allowing public keys which serialize differently to compare the same.

PR #2923, opened 2026-01-23
[cryptography] Validate BLS12-381 elements in From impls

### LINKED ISSUE #2916: [Medium] From trait implementations allow constructing zero/identity/nongroup cryptographic elements without validation
From<Private> allows zero scalar (bypasses decode validation)
```
#[test]
    fn from_private_allows_zero_scalar() {
        use crate::bls12381::primitives::group::{Private, Scalar};
        use crate::bls12381::PrivateKey;

        // Decode path rejects zero scalar.
        let zero_bytes = [0u8; Scalar::SIZE];
        assert!(Scalar::decode(zero_bytes.as_ref()).is_err());

        // But From<Private> allows wrapping zero scalar.
        let zero_private = Private::new(Scalar::zero());
        let priv_key = PrivateKey::from(zero_private);
        let pub_key = priv_key.public_key();
        assert_eq!(pub_key.encode(), G1::zero().encode());
    }
```

From<Public> allows identity point elements
```
#[test]
    fn from_public_allows_identity_point() {
        use crate::bls12381::primitives::group::G1;
        use crate::bls12381::PublicKey;
        
        let id = G1::zero();

        // Decode path rejects infinity/identity.
        assert!(G1::decode(id.encode().as_ref()).is_err());
        
        let _ = PublicKey::from(id);
        assert!(PublicKey::decode(id.encode().as_ref()).is_err());
    }
```
From<Public> allows non-subgroup elements
```
#[test]
    fn public_key_from_allows_non_curve_point() {
        use blst::blst_p1;
        use commonware_codec::DecodeExt;
        use crate::bls12381::primitives::group::G1;
        use crate::bls12381::PublicKey;

        const P1_SIZE: usize = core::mem::size_of::<blst_p1>();
        let raw = [0xAAu8; P1_SIZE];

        let p = unsafe { (raw.as_ptr() as *const blst_p1).read_unaligned() };
        let g1 = G1::from_blst_p1(p);
        let pk = PublicKey::from(g1);

        let encoded = pk.encode();
        assert!(PublicKey::decode(encoded.as_ref()).is_err());
    }

### PR DESCRIPTION
**Pull request by andresilva**
Fixes #2916.

### CODE REVIEW

File: cryptography/src/bls12381/scheme.rs [hunk 1]
```diff
@@ -224,6 +224,14 @@ impl Deref for PublicKey {
 
 impl From<<MinPk as Variant>::Public> for PublicKey {
```
andresilva: Opted to just panic here to make this ergonomic to use, but can move to `TryFrom` if preferred.

cryptography/src/bls12381/scheme.rs
```diff
@@ -472,6 +488,33 @@ mod tests {
         assert!(private_key.public_key().verify(b"ns", msg, &sig));
     }
 
+    #[test]
+    #[should_panic(expected = "cannot create Private from zero scalar")]
+    fn test_private_rejects_zero_scalar() {
+        use commonware_math::algebra::Additive;
```
patrick-ogrady: move to top of mod (we may want to add to `AGENTS.md` (if not already there)

cryptography/src/bls12381/scheme.rs
patrick-ogrady: I think we should go the `TryFrom` route to adhere to rust semantics (I've done this before and since changed to Try)

cryptography/src/bls12381/scheme.rs
andresilva: https://github.com/commonwarexyz/monorepo/pull/2925

cryptography/src/bls12381/scheme.rs
```diff
@@ -224,6 +224,14 @@ impl Deref for PublicKey {
 
 impl From<<MinPk as Variant>::Public> for PublicKey {
     fn from(key: <MinPk as Variant>::Public) -> Self {
```
patrick-ogrady: Do we need this?

cryptography/src/bls12381/primitives/group.rs
```diff
@@ -1052,18 +1072,21 @@ impl Read for G1 {
...
+            return Err(Invalid("G1", "Infinity"));
         }
-        Ok(Self(ret))
+
+        // Verify that the deserialized element is in G1
+        if !g1.is_in_group() {
+            return Err(Invalid("G1", "Outside G1"));
+        }
+
+        Ok(g1)
```
cronokirby: I don't like this, because now you have to break the invariant that we never construct a point that is the identity or outside the group, for no particular reason?

cryptography/src/bls12381/primitives/group.rs
```diff
@@ -1472,18 +1507,21 @@ impl Read for G2 {
...
+            return Err(Invalid("G2", "Infinity"));
         }
-        Ok(Self(ret))
+
+        // Verify that the deserialized element is in G2
+        if !g2.is_in_group() {
+            return Err(Invalid("G2", "Outside G2"));
+        }
+
+        Ok(g2)
```
cronokirby: Same issue as G1

cryptography/src/bls12381/primitives/group.rs
```diff
@@ -414,7 +414,15 @@ pub struct Private {
...
-    pub const fn new(private: Scalar) -> Self {
+    ///
+    /// # Panics
+    ///
+    /// Panics if the scalar is zero.
+    pub fn new(private: Scalar) -> Self {
+        assert!(
+            private != Scalar::zero(),
+            "cannot create Private from zero scalar"
+        );
```
cronokirby: Do we actually need this method publicly? Or is it just used in tests?

cryptography/src/bls12381/scheme.rs
```diff
@@ -326,6 +334,14 @@ impl Deref for Signature {
...
 impl From<<MinPk as Variant>::Signature> for Signature {
     fn from(signature: <MinPk as Variant>::Signature) -> Self {
+        assert!(
+            !signature.is_identity(),
+            "cannot create Signature from identity point"
+        );
+        assert!(
+            signature.is_in_group(),
+            "cannot create Signature from point outside G2"
+        );
```
cronokirby: I don't see this as being necessary, because it isn't possible to create a signature element that's outside of the group. For identity points, we can shift the burden to making sure public keys are good, in which case signature verification can take care of this.

cryptography/src/bls12381/scheme.rs
```diff
@@ -224,6 +224,14 @@ impl Deref for PublicKey {
...
 impl From<<MinPk as Variant>::Public> for PublicKey {
     fn from(key: <MinPk as Variant>::Public) -> Self {
+        assert!(
+            !key.is_identity(),
+            "cannot create PublicKey from identity point"
+        );
+        assert!(
+            key.is_in_group(),
+            "cannot create PublicKey from point outside G1"
+        );
```
cronokirby: This check isn't necessary, the `G1` type already checks this by construction.

cronokirby (commented):
I think some of the checks are redundant; would be a good opportunity to spell out the assumptions and guarantees a bit more clearly.

cryptography/src/bls12381/primitives/group.rs
andresilva: It's used in DKG but can be made `pub(crate)`.

cryptography/src/bls12381/scheme.rs
andresilva: It's used to implement `Signer::public_key`.

cronokirby (approved):
This looks good. I also did a pass around the rest of BLS impacted code (DKG, aggregate, etc.) to make sure we weren't accidentally assuming things incorrectly.

cryptography/src/bls12381/primitives/group.rs
```diff
@@ -1037,32 +1051,26 @@ impl Read for G1 {
...
             }
             blst_p1_from_affine(&mut ret, &affine);
-
-            // Verify that deserialized element isn't infinite
-            if blst_p1_is_inf(&ret) {
-                return Err(Invalid("G1", "Infinity"));
-            }
-
-            // Verify that the deserialized element is in G1
-            if !blst_p1_in_g1(&ret) {
```
patrick-ogrady, line 1062: Wait, I think we need this?

cronokirby: We do need this actually, and we should add a test for that!

cronokirby (changes requested):
We should add back in the on curve checks, and add a test for that to make sure the regression is caught

cryptography/src/bls12381/primitives/group.rs
andresilva: AFAIU both of these checks were redundant as they were already performed by `blst_p1_uncompress` above, there is an error for `BLST_POINT_NOT_IN_GROUP`.

patrick-ogrady: I guess the errors tell us everything we need to know (we were doing this twice)?

cryptography/src/bls12381/primitives/group.rs
```diff
@@ -1037,32 +1051,26 @@ impl Read for G1 {
...
-        // SAFETY: bytes is a valid 48-byte array. blst_p1_uncompress validates encoding.
-        // Additional checks for infinity and subgroup membership prevent small subgroup attacks.
+        // SAFETY: bytes is a valid 48-byte array. blst_p1_uncompress validates encoding,
+        // curve membership, and subgroup membership.
         unsafe {
             let mut affine = blst_p1_affine::default();
             match blst_p1_uncompress(&mut affine, bytes.as_ptr()) {
                 BLST_ERROR::BLST_SUCCESS => {}
+                BLST_ERROR::BLST_PK_IS_INFINITY => {
+                    // Identity point allowed for algebraic G1 (rejected at PublicKey layer)
```
patrick-ogrady, line 1077: if it returns INFINITY as an error, does it bypass other checks later in the function that are relevant (is there anything left to check in that case)?

cryptography/src/bls12381/primitives/group.rs
```diff
@@ -1037,32 +1051,26 @@ impl Read for G1 {
...
+                    return Ok(Self::zero());
+                }
                 BLST_ERROR::BLST_BAD_ENCODING => return Err(Invalid("G1", "Bad encoding")),
                 BLST_ERROR::BLST_POINT_NOT_ON_CURVE => return Err(Invalid("G1", "Not on curve")),
```
patrick-ogrady, line 1081: Given not all these errors can be returned, I wonder if we should `|` the ones that can't be and panic?

patrick-ogrady: Maybe that is too strict.

andresilva: It's not too obvious which ones can be returned, and maybe it could change with a new version?

cronokirby (dismissed):

cryptography/src/bls12381/primitives/group.rs
```diff
@@ -414,7 +414,15 @@ pub struct Private {
 
 impl Private {
     /// Creates a new private key from a scalar.
-    pub const fn new(private: Scalar) -> Self {
+    ///
+    /// # Panics
+    ///
+    /// Panics if the scalar is zero.
+    pub(crate) fn new(private: Scalar) -> Self {
```
patrick-ogrady: We shouldn't hit this panic if a malicious dealer provides us with a 0 share, right?

patrick-ogrady (commented):
For consistency, should we then be changing `Scalar` to accept 0?

```
impl Read for Scalar {
    type Cfg = ();

    fn read_cfg(buf: &mut impl Buf, _: &()) -> Result<Self, Error> {
        let bytes = Zeroizing::new(<[u8; Self::SIZE]>::read(buf)?);
        let mut ret = blst_fr::default();
        // SAFETY: bytes is a valid 32-byte array. blst_sk_check validates non-zero and in-range.
        // We use blst_sk_check instead of blst_scalar_fr_check because it also checks non-zero
        // per IETF BLS12-381 spec (Draft 4+).
        //
        // References:
        // * https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-bls-signature-03#section-2.3
        // * https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-bls-signature-04#section-2.3
        unsafe {
            let mut scalar = blst_scalar::default();
            blst_scalar_from_bendian(&mut scalar, bytes.as_ptr());
            if !blst_sk_check(&scalar) {
                return Err(Invalid("Scalar", "Invalid"));
            }
            blst_fr_from_scalar(&mut ret, &scalar);
        }
        Ok(Self(ret))
    }
}
```

cryptography/src/bls12381/primitives/group.rs
andresilva: If it’s the 0 point you know it’s on the curve and in the subgroup.

cryptography/src/bls12381/primitives/group.rs
andresilva: We will validate that it's non-zero on decode, this is only used internally in the DKG.

cryptography/src/bls12381/dkg.rs
```diff
@@ -1947,62 +1949,64 @@ mod test_plan {
...
-                                info: info.clone(),
-                                pub_msg: pub_msg.clone(),
-                                results,
-                                transcript,
-                            };
-                            (dealer, pub_msg, priv_msgs)
-                        } else {
-                            Dealer::start::<N3f1>(&mut rng, info.clone(), sk.clone(), share)?
+                                        ))
+                                        .expect("polynomial evaluation is non-zero"),
```
andresilva: This is the only change here, the rest is just rustfmt fallout.

cryptography/src/bls12381/primitives/group.rs
```diff
@@ -456,14 +479,14 @@ impl FixedSize for Private {
 
 impl Random for Private {
     fn random(rng: impl CryptoRngCore) -> Self {
-        Self::new(Scalar::random(rng))
+        Self::new(Scalar::random(rng)).expect("random scalar is non-zero")
```
andresilva: We could update `Scalar::random` to allow zero, and only do the `from_ikm` which uses `blst_keygen` for `Private`.

cronokirby: The likelihood of this panicking is negligible, so this is fine.

cronokirby:
For the DKG, the properties around your share are guaranteed by checking the polynomial itself. It's important to rely only on public information when generating errors, so that everybody agrees on whether or not the DKG succeeded. I think what we should add, instead of the last commit, is to additionally require that a dealer's pub msg's polynomial does not have 0 as the constant element of the polynomial, and then we'll have a guarantee that whatever share you end up with at the end can be converted to a Private without issue (so we can panic).

cryptography/src/bls12381/primitives/group.rs
andresilva: It should never panic because `Scalar::random` is guaranteed to be non-zero (it uses `blst_keygen` which will just loop until a non-zero scalar is generated).

cryptography/src/bls12381/dkg.rs
```diff
@@ -1178,25 +1182,22 @@ impl<V: Variant, S: Signer> Dealer<V, S> {
...
-            &mut rng,
-            // We are extracting the private scalar from `Secret` protection because
-            // `Poly::new_with_constant` requires an owned value. The extracted scalar is
-            // scoped to this function and will be zeroized on drop (i.e. the secret is
-            // only exposed for the duration of this function).
-            share.map(|x| x.private.expose_unwrap()),
-        )?;
-        let my_poly = Poly::new_with_constant(&mut rng, info.degree::<M>(), share);
+        let share = info.unwrap_or_random_share(&mut rng, share.map(|x| x.private))?;
+        // We are extracting the private scalar from `Secret` protection because
```
patrick-ogrady, line 1223: I think this would go above the share extraction?

andresilva: The extraction is happening in the line after this comment with `share.expose_unwrap()`, it's meant as moving the scalar out of `Secret` box.

cronokirby (dismissed):

cryptography/src/bls12381/primitives/group.rs
patrick-ogrady: I'm wondering if we just have a `TryFrom` for Scalar -> Private and remove `new()`

andresilva: Updated.

cryptography/src/bls12381/primitives/group.rs
```diff
@@ -443,6 +428,22 @@ impl Private {
...
 
+impl TryFrom<Scalar> for Private {
+    type Error = ();
+
+    /// Converts a scalar to a private key.
+    ///
+    /// Returns `Err(())` if the scalar is zero.
+    fn try_from(scalar: Scalar) -> Result<Self, Self::Error> {
+        if scalar == Scalar::zero() {
+            return Err(());
```
patrick-ogrady: For consistency with the TryFrom<G1/G2>, I think we would use CodecError here?

If not that, we could add another variant to primitives (for InvalidPublicKey/InvalidPrivateKey) and use them in both spots?

cryptography/src/bls12381/primitives/group.rs
```diff
@@ -435,6 +428,22 @@ impl Private {
...
+impl TryFrom<Scalar> for Private {
+    type Error = Error;
+
+    /// Converts a scalar to a private key.
+    ///
+    /// Returns an error if the scalar is zero.
+    fn try_from(scalar: Scalar) -> Result<Self, Self::Error> {
+        if scalar == Scalar::zero() {
+            return Err(Invalid("Private", "Zero"));
+        }
```
patrick-ogrady, line 440: I think we need to `blst_sk_check` here?

patrick-ogrady: now that a scalar just has `blst_scalar_check`?

patrick-ogrady: I think we should add a regression test for a non-zero valid scalar failing here?

cryptography/src/bls12381/dkg.rs
```diff
@@ -547,6 +547,11 @@ impl<V: Variant, P: PublicKey> Info<V, P> {
         if self.degree::<M>() != pub_msg.commitment.degree_exact() {
             return false;
         }
+        // Reject dealers committing to a zero secret (identity point as constant term).
+        // This ensures all accepted shares are non-zero.
+        if *pub_msg.commitment.constant() == V::Public::zero() {
```
patrick-ogrady, line 552: If this were to happen (somehow), we'd just fail to parse the inbound share I suppose?

cryptography/src/bls12381/primitives/group.rs
```diff
@@ -456,14 +483,16 @@ impl FixedSize for Private {
...
+            .try_into()
+            .expect("random scalar is non-zero")
     }
 }
 
 #[cfg(feature = "arbitrary")]
 impl arbitrary::Arbitrary<'_> for Private {
     fn arbitrary(u: &mut arbitrary::Unstructured<'_>) -> arbitrary::Result<Self> {
-        Ok(Self::new(u.arbitrary::<Scalar>()?))
+        Ok(Self::try_from(u.arbitrary::<Scalar>()?).expect("arbitrary scalar is non-zero"))
```
patrick-ogrady, line 495: I think we need to loop on this (there are some scalars which won't pass the sk_check)?

cryptography/src/bls12381/primitives/group.rs
andresilva: I can add `blst_sk_check` but it's redundant. `Scalar` wraps `blst_fr` (which was checked with `blst_scalar_fr_check`) so it's already guaranteed to be reduced modulo the field order, the only meaningful check is non-zero. See https://github.com/supranational/blst/blob/806515251de5cfb9761ad12eb2f2f151b1ad0277/src/exports.c#L107-L113.

cryptography/src/bls12381/primitives/group.rs
andresilva: The current implementation of arbitrary/random `Scalar` guarantees non-zero (it's generated with `blst_keygen`), so the `Scalar` we wrap here should always be a valid private key.

cryptography/src/bls12381/dkg.rs
andresilva: Yes, I believe so, since the resulting share would also be zero. When the recipient tried to parse it as a `Private`, it would fail since it rejects zero scalars.

cryptography/src/bls12381/primitives/group.rs
```diff
@@ -445,8 +454,26 @@ impl Read for Private {
     type Cfg = ();
 
     fn read_cfg(buf: &mut impl Buf, _: &()) -> Result<Self, Error> {
```
cronokirby, line 456: For this method, why not use the Scalar::read impl and then try_into?

cryptography/src/bls12381/dkg.rs
```diff
@@ -687,15 +692,13 @@ where
 
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct DealerPrivMsg {
-    share: Secret<Scalar>,
+    share: Private,
```
cronokirby, line 695: Using `Private` here is actually wrong. It's perfectly fine to pick a polynomial `f` such that `f(j) = 0` for some participant `j`.

cronokirby (changes requested):
One rabbit hole I went down, is that if you think about it, it doesn't quite make sense to require `Share` to have a non-zero scalar. You don't really care if it's zero, so long as the resulting public key has enough entropy. It's "fine" for some adversary to pick a polynomial such that `f(j) = 0`, the security is guaranteed by having enough contributions.

One reason why you can't just do this is that you expect to be able to map shares to public keys. If you require public keys to be non-zero, you have an issue there.

But this made me think? Why even require public keys to be non-zero? There are other low-entropy points which are equally bad.

cronokirby (commented):
(overwriting previous blocking review)

cryptography/src/bls12381/primitives/group.rs
```diff
@@ -435,6 +428,22 @@ impl Private {
...
 }
 
+impl TryFrom<Scalar> for Private {
+    type Error = Error;
+
+    /// Converts a scalar to a private key.
+    ///
+    /// Returns an error if the scalar is zero.
+    fn try_from(scalar: Scalar) -> Result<Self, Self::Error> {
+        if scalar == Scalar::zero() {
```
patrick-ogrady, line 438: I think we should call `blst_sk_check` here?

clabby:

PR #2926, opened 2026-01-23
Don't store bytes again for ed25519 keys

cronokirby:
This also removes an unnecessary `PartialEq` requirement on the `PrivateKey` trait.

The `ed25519_consensus` library already provides convenient access to the bytes we were storing ourselves, at no cost. There's no need to store the bytes of a private key and public key again ourselves, when the struct we're using also stores them.

### DIFF
--- a/cryptography/src/ed25519/scheme.rs
+++ b/cryptography/src/ed25519/scheme.rs
@@ -232,22 +232,19 @@ impl arbitrary::Arbitrary<'_> for PublicKey {
 
 /// Ed25519 Signature.
 ///
-/// This signature is *non-malleable*: an adversary with access to many messages
-/// and signatures, verifying against an honestly generated public key, cannot
-/// find a new signature which will verify, even by tampering or modify the signatures
-/// that it has seen previously.
+/// Signatures from honestly generated keys are *non-malleable*: an adversary
+/// with access to many messages and signatures, verifying against an honestly
+/// generated public key, cannot find a new signature which will verify, even by
+/// tampering or modifying the signatures that it has seen previously.
 ///
-/// Like any signature, it's also not possible to have a signature verifying
-/// against one message also verify against another.
-///
-/// This property does not hold for maliciously generated public keys. In particular,
-/// it's possible to craft public keys (which would otherwise not be honestly generatable)
-/// for which a signature will verify against any message.
+/// Like any signature, it's also not possible to have a signature that verifies against
+/// one message also verify against another. This property does not hold for maliciously
+/// generated public keys. In particular, it's possible to craft public keys (which would
+/// otherwise not be honestly generatable) for which a signature will verify against any message.
 #[derive(Clone, Eq, PartialEq)]
 pub struct Signature {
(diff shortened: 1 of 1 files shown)


patrick-ogrady:
As discussed, let's add a test to close #2919 (different pubkeys with same decompressed encoding can't be used to verify the same signature).

cronokirby:
So, with such a check, the only feasible way to check the canonical vs non-canonical encoding is to test it for the identity point, which has both:

- `y < 19`
- a known discrete logarithm

What's interesting is that this hits on a different "issue" which is that there are points of small order which are all trivial to forge signatures for. These are points on the curve of order <= 8, which in the ed25519_consensus algorithm, ends up all having trivial signatures.

So, for all of those points, it is the case that one can have a signature which verifies even if the public key changes (although, this public key is also trivial to forge signatures for, as well).

I think this is an orthogonal issue to this PR, but maybe we should add a check to exclude those? It depends on whether or not we want to provide this guarantee even up to public keys that are _adversarially_ chosen and come with drawbacks (like trivial forgery) rather than being able to mess with the bytes of an honest public key.

cryptography/src/ed25519/scheme.rs
```diff
@@ -229,10 +231,24 @@ impl arbitrary::Arbitrary<'_> for PublicKey {
...
+/// Like any signature, it's also not possible to have a signature verifying
+/// against one message also verify against another.
+///
+/// This property does not hold for maliciously generated public keys. In particular,
+/// it's possible to craft public keys (which would otherwise not be honestly generatable)
+/// for which a signature will verify against any message.
 #[derive(Clone, Eq, PartialEq)]
 pub struct Signature {
+    // There's not really a point in storing an [`ed25519_consensus::Signature`],
+    // because that type is easily created form this byte array, just by copying
```
patrick-ogrady: nit: `from this byte array`

cryptography/src/ed25519/scheme.rs
```diff
@@ -229,10 +231,24 @@ impl arbitrary::Arbitrary<'_> for PublicKey {
...
+///
+/// This signature is *non-malleable*: an adversary with access to many messages
+/// and signatures, verifying against an honestly generated public key, cannot
+/// find a new signature which will verify, even by tampering or modify the signatures
+/// that it has seen previously.
+///
+/// Like any signature, it's also not possible to have a signature verifying
+/// against one message also verify against another.
+///
+/// This property does not hold for maliciously generated public keys. In particular,
```
patrick-ogrady: This reads a bit like a counterpoint to `This signature is *non-malleable*`.

I think you could do:

```
/// Signatures from honestly generated keys are *non-malleable*: an adversary with access to many messages
/// and signatures, verifying against an honestly generated public key, cannot
/// find a new signature which will verify, even by tampering or modify the signatures
/// that it has seen previously. Like any signature, it's also not possible to have a signature verifying
/// against one message also verify against another.

<existing>
```

cryptography/src/ed25519/scheme.rs
patrick-ogrady: (keep the bots away)

PR #2927, opened 2026-01-23
[treewide] rustfmt over select!

andresilva:
I added support for `select!` and `select_loop!` to rustfmt. This is the result 😬.

https://github.com/andresilva/rustfmt/commit/e20ec8b1841860e7793896f676dbe3e57c0ba883

Depends on #2671.

### DIFF
--- a/broadcast/src/buffered/engine.rs
+++ b/broadcast/src/buffered/engine.rs
@@ -168,15 +168,31 @@ impl<E: Clock + Spawner + Metrics, P: PublicKey, M: Committable + Digestible + C
                     break;
                 };
                 match msg {
-                    Message::Broadcast{ recipients, message, responder } => {
+                    Message::Broadcast {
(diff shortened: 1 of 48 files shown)


patrick-ogrady:
> I added support for select! and select_loop! to rustfmt. This is the result 😬.

https://github.com/andresilva/rustfmt/commit/e20ec8b1841860e7793896f676dbe3e57c0ba883

So, the idea is to just run it locally once in a while? Can we store a patch in CI and apply it to the latest rustfmt or something to catch in CI?

andresilva:
I think running this in CI would be ideal, not sure how fast that patch will rot, so might be best to just build that `rustfmt` binary once and download it in CI? Most likely in a separate format step, so that we keep running the main format step with the up-to-date `rustfmt`. WDYT? We can then update the patch and re-build the binary every once in a while.

PR #2932, opened 2026-01-24
Deduplicate `read_ordered_map` helper function

zeevick10:
Removes code duplication by extracting the `read_ordered_map` helper function into a shared location (`codec/src/types/mod.rs`), following the existing pattern used for `read_ordered_set`.

The `read_ordered_map` function was duplicated in both `hash_map.rs` and `btree_map.rs` with identical logic for validating key ordering and detecting duplicates during deserialization. This duplication creates a maintenance risk where future changes to validation rules could be applied to one implementation but forgotten in the other, leading to inconsistent behavior between `HashMap` and `BTreeMap` codec implementations.

### DIFF
--- a/codec/src/types/btree_map.rs
+++ b/codec/src/types/btree_map.rs
@@ -8,59 +8,13 @@ extern crate alloc;
 use crate::{
     codec::{EncodeSize, Read, Write},
     error::Error,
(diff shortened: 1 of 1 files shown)


patrick-ogrady:
Duplicate of #2711

PR #2935, opened 2026-01-25
fix: remove dead sum_hash in metamorph lthash fuzz target

MozirDmitriy:
The AddMultipleElementsThenSubtractSum mutation in metamorph_lthash.rs created a separate sum_hash LtHash and called sum_hash.add(&element) for each element, but this value was never used in any further computation or assertion. This made the fuzz target harder to read and performed redundant LtHash work that had no effect on the final checksum or the property being tested.


clabby:

PR #2937, opened 2026-01-26
[consensus/marshal] make store traits fully async and fallible

SuperFluffy:
Makes `Certificates` and `Blocks` traits from `consensus::marshal::store` fully async and fallible. This is necessary to implement the traits for objects that only provide async reads (and can potentially error on every interaction, as is the case with database connection pools).

The current definition still has the implementation details of the various `Archive` objects leaking through, which allows non-async, infallible access to various storage indices.

Related: #2918

### DIFF
--- a/consensus/src/marshal/actor.rs
+++ b/consensus/src/marshal/actor.rs
@@ -952,7 +952,16 @@ where
     /// should have the associated block (in the `finalized_blocks` archive) for the information
     /// returned.
     async fn get_latest(&mut self) -> Option<(Height, B::Commitment, Round)> {
-        let height = self.finalizations_by_height.last_index().await.ok()??;
+        let height = match self.finalizations_by_height.last_index().await {
+            Err(err) => {
+                warn!(
+                    error = &err as &dyn std::error::Error,
+                    "unable to read latest available finalized height",
+                );
+                return None;
+            }
+            Ok(height) => height?,
+        };
         let finalization = self
             .get_finalization_by_height(height)
             .await
@@ -994,9 +1003,18 @@ where
     ) {
         let start = self.last_processed_height.next();
         'cache_repair: loop {
-            let Ok((gap_start, Some(gap_end))) = self.finalized_blocks.next_gap(start).await else {
+            let (gap_start, gap_end) = match self.finalized_blocks.next_gap(start).await {
+                Err(err) => {
+                    warn!(
+                        error = &err as &dyn std::error::Error,
+                        "unable to query storage for missing gaps",
+                    );
+                    return;
+                }
+                Ok((gap_start, Some(gap_end))) => (gap_start, gap_end),
+
                 // No gaps detected
-                return;
+                _ => return,
             };
 
             // Attempt to repair the gap backwards from the end of the gap, using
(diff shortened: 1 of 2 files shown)


consensus/src/marshal/actor.rs
```diff
@@ -952,7 +952,7 @@ where
     /// should have the associated block (in the `finalized_blocks` archive) for the information
     /// returned.
     async fn get_latest(&mut self) -> Option<(Height, B::Commitment, Round)> {
-        let height = self.finalizations_by_height.last_index()?;
+        let height = self.finalizations_by_height.last_index().await.ok()??;
```
SuperFluffy: Silently drops potential errors - emit an event here? Or make `get_latest` fallible?

consensus/src/marshal/actor.rs
```diff
@@ -994,7 +994,7 @@ where
     ) {
         let start = self.last_processed_height.next();
         'cache_repair: loop {
-            let (gap_start, Some(gap_end)) = self.finalized_blocks.next_gap(start) else {
+            let Ok((gap_start, Some(gap_end))) = self.finalized_blocks.next_gap(start).await else {
```
SuperFluffy: Emit an error event here?

consensus/src/marshal/actor.rs
```diff
@@ -1038,15 +1038,19 @@ where
...
-            .into_iter()
-            .map(|height| Request::<B>::Finalized { height })
-            .collect::<Vec<_>>();
-        if !requests.is_empty() {
-            resolver.fetch_all(requests).await
+            .missing_items(start, self.max_repair.get())
+            .await
+        {
+            let requests = missing_items
+                .filter_map(|item| async move { item.ok() })
```
SuperFluffy: Simiarly issue to the rest - emit an error event in case the missing items cursor returns an error?

consensus/src/marshal/store.rs
```diff
@@ -146,7 +151,11 @@ pub trait Blocks: Send + Sync + 'static {
     /// A vector containing up to `max` missing heights from gaps between ranges.
     /// The vector may contain fewer than `max` items if there aren't enough gaps.
     /// If there are no more ranges after the current position, no items are returned.
-    fn missing_items(&self, start: Height, max: usize) -> Vec<Height>;
+    fn missing_items(
+        &self,
+        start: Height,
+        max: usize,
```
SuperFluffy, line 157: Another option would be to remove the `max` argument and allow the caller to `.take(max)` amounts of items, as they see fit.

clabby (commented):
Looks fine to me - thanks.

Re: Error events in the marshal actor, I think we should add those for now. If we want to follow up with bubbling up failures, we can do so in another PR.

consensus/src/marshal/store.rs
clabby: This may be hard to work with given the underlying `{prunable, immutable}::Archive` implementations don't return a stream. Could refactor those interfaces at a later point, if this API proves burdensome.

clabby (approved):

patrick-ogrady:
Pondering now. I guess the question is if we want to treat those storage errors as a fault (like we do everywhere else)?

For consistency, we'd do a Result<Option> and if Error was returned we'd exit (I suppose a DB would just swallow transient errors 🤔 ).

From a trait perspective, however, I do agree with this direction (the current traits take advantage of the archive's efficient storage of "ranges" when other DBs may not have such support).

SuperFluffy:
> Pondering now. I guess the question is if we want to treat those storage errors as a fault (like we do everywhere else)?
> 
> 
> 
> For consistency, we'd do a Result<Option> and if Error was returned we'd exit (I suppose a DB would just swallow transient errors 🤔 ).
> 
> 
> 
> From a trait perspective, however, I do agree with this direction (the current traits take advantage of the archive's efficient storage of "ranges" when other DBs may not have such support).

I guess you could push this into the implementor and require that implementations only return unrecoverable errors?

Since you can't really do error handling/recovery at the actor level when using generic storage.

clabby:

PR #2940, opened 2026-01-26
[runtime] introduce `IoBuf` pool

andresilva:
This PR introduces `BufferPool`, a pool of reusable, aligned buffers for I/O operations. The pool reduces allocation overhead in hot paths by recycling buffers across operations, which is particularly important for network and storage workloads that repeatedly allocate similarly-sized buffers.

Buffers are organized into power-of-two size classes (e.g., 4KB, 8KB, 16KB up to a configurable maximum). Each size class is backed by a lock-free freelist using `crossbeam::ArrayQueue`, enabling concurrent allocation and return without contention. When a buffer is requested, the pool rounds up to the appropriate size class and attempts to pop from that freelist, if empty, a fresh buffer is allocated. Returned buffers are pushed back to their freelist for reuse.

The implementation provides two distinct pools with different alignment requirements: a storage pool with page-aligned buffers (typically 4KB) to support direct I/O operations like `O_DIRECT` on Linux, and a network pool with cache-line-aligned buffers for efficient message passing. These pools are exposed to the runtime through the `Pooling` trait, with `BufferPools` providing a registry that holds both pools.

`PooledBufMut` can be frozen into an immutable `IoBuf` without copying by leveraging `Bytes::from_owner` to wrap the pooled buffer in a custom owner type that holds a weak reference to the pool. When all clones of the `IoBuf` (including slices) are dropped, the buffer automatically returns to the pool. This enables efficient patterns where a buffer is written once, frozen, cloned for multiple consumers, and recycled when all consumers are done. The implementation also supports a cursor that tracks the start of the readable region, enabling `advance()` to consume bytes from the front without reallocating, matching `BytesMut` semantics.

`IoBufMut` now supports both owned (`BytesMut`-backed) and pooled variants through an internal enum, with the public API remaining largely unchanged.

In this PR, only the network receive path has been updated to use the pool. Follow-up PRs will expand pool usage to other hot paths such as network send and storage I/O.

During development, I implemented a `resize` method on `PooledBufMut` with complex logic to handle growing beyond the current buffer's capacity: attempting to allocate from larger size classes in the pool and falling back to heap allocation when exhausted. However `resize` was only used in a single place in the entire codebase and I instead replaced its usage, and given the complexity and the risk of pooled buffers silently becoming heap-allocated, `resize` was removed from `IoBufMut`.

Related #2883.

### DIFF
--- a/runtime/src/deterministic.rs
+++ b/runtime/src/deterministic.rs
@@ -66,6 +66,8 @@ use crate::{Blocker, Pacer};
 use commonware_codec::Encode;
 use commonware_macros::select;
 use commonware_parallel::ThreadPool;
+#[cfg(miri)]
+use commonware_utils::NZUsize;
 use commonware_utils::{hex, time::SYSTEM_TIME_PRECISION, SystemTimeExt};
 #[cfg(feature = "external")]
 use futures::task::noop_waker;
@@ -856,12 +858,28 @@ impl Context {
         let network = MeteredNetwork::new(network, runtime_registry);
 
         // Initialize buffer pools
+        cfg_if::cfg_if! {
+            if #[cfg(miri)] {
+                // Reduce max_per_class to avoid slow atomics under miri
+                let network_config = BufferPoolConfig {
+                    max_per_class: NZUsize!(32),
+                    ..BufferPoolConfig::for_network()
+                };
+                let storage_config = BufferPoolConfig {
+                    max_per_class: NZUsize!(32),
+                    ..BufferPoolConfig::for_storage()
+                };
+            } else {
+                let network_config = BufferPoolConfig::for_network();
+                let storage_config = BufferPoolConfig::for_storage();
+            }
+        }
         let network_buffer_pool = BufferPool::new(
-            BufferPoolConfig::for_network(),
+            network_config,
             runtime_registry.sub_registry_with_prefix("network_buffer_pool"),
         );
         let storage_buffer_pool = BufferPool::new(
-            BufferPoolConfig::for_storage(),
+            storage_config,
             runtime_registry.sub_registry_with_prefix("storage_buffer_pool"),
         );
 
@@ -925,12 +943,28 @@ impl Context {
         let network = MeteredNetwork::new(network, runtime_registry);
 
         // Initialize buffer pools
+        cfg_if::cfg_if! {
+            if #[cfg(miri)] {
+                // Reduce max_per_class to avoid slow atomics under Miri
+                let network_config = BufferPoolConfig {
+                    max_per_class: NZUsize!(32),
+                    ..BufferPoolConfig::for_network()
+                };
+                let storage_config = BufferPoolConfig {
+                    max_per_class: NZUsize!(32),
+                    ..BufferPoolConfig::for_storage()
+                };
+            } else {
+                let network_config = BufferPoolConfig::for_network();
+                let storage_config = BufferPoolConfig::for_storage();
+            }
+        }
         let network_buffer_pool = BufferPool::new(
-            BufferPoolConfig::for_network(),
+            network_config,
             runtime_registry.sub_registry_with_prefix("network_buffer_pool"),
         );
         let storage_buffer_pool = BufferPool::new(
-            BufferPoolConfig::for_storage(),
+            storage_config,
             runtime_registry.sub_registry_with_prefix("storage_buffer_pool"),
         );
 
--- a/runtime/src/iobuf/pool.rs
+++ b/runtime/src/iobuf/pool.rs
@@ -1333,7 +1333,6 @@ mod tests {
     }
 
     #[test]
-    #[cfg_attr(miri, ignore)]
     fn test_multithreaded_alloc_freeze_return() {
         let page = page_size();
         let mut registry = test_registry();
@@ -1341,11 +1340,20 @@ mod tests {
 
         let mut handles = vec![];
 
+        // Reduce iterations under miri (atomics are slow)
+        cfg_if::cfg_if! {
+            if #[cfg(miri)] {
+                let iterations = 100;
+            } else {
+                let iterations = 1000;
+            }
+        }
+
         // Spawn multiple threads that allocate, freeze, clone, and drop
         for _ in 0..10 {
             let pool = pool.clone();
             let handle = thread::spawn(move || {
-                for _ in 0..1000 {
+                for _ in 0..iterations {
                     let buf = pool.try_alloc(100).unwrap();
                     let iobuf = buf.freeze();
 
@@ -1753,8 +1761,25 @@ mod tests {
         let cache_line = cache_line_size();
         let mut registry = test_registry();
 
+        // Reduce max_per_class under miri (atomics are slow)
+        cfg_if::cfg_if! {
+            if #[cfg(miri)] {
+                let storage_config = BufferPoolConfig {
+                    max_per_class: NZUsize!(32),
+                    ..BufferPoolConfig::for_storage()
+                };
+                let network_config = BufferPoolConfig {
+                    max_per_class: NZUsize!(32),
+                    ..BufferPoolConfig::for_network()
+                };
+            } else {
+                let storage_config = BufferPoolConfig::for_storage();
+                let network_config = BufferPoolConfig::for_network();
+            }
+        }
+
         // Storage preset - page aligned
-        let storage_buffer_pool = BufferPool::new(BufferPoolConfig::for_storage(), &mut registry);
+        let storage_buffer_pool = BufferPool::new(storage_config, &mut registry);
         let mut buf = storage_buffer_pool.try_alloc(100).unwrap();
         assert_eq!(
             buf.as_mut_ptr() as usize % page,
@@ -1763,7 +1788,7 @@ mod tests {
         );
 
         // Network preset - cache-line aligned
-        let network_buffer_pool = BufferPool::new(BufferPoolConfig::for_network(), &mut registry);
+        let network_buffer_pool = BufferPool::new(network_config, &mut registry);
         let mut buf = network_buffer_pool.try_alloc(100).unwrap();
         assert_eq!(
             buf.as_mut_ptr() as usize % cache_line,


runtime/src/iobuf/pool.rs
```diff
@@ -0,0 +1,1984 @@
...
+    ///
+    /// Network operations typically need multiple concurrent buffers per connection
+    /// (message, encoding, encryption) so we allow 4096 buffers per size class.
+    /// Cache-line alignment is used because network buffers don't require page
+    /// alignment for DMA, and smaller alignment reduces internal fragmentation.
+    pub const fn for_network() -> Self {
+        let cache_line = cache_line_size();
+        Self {
+            min_size: cache_line,
+            max_size: 1048576,
```
danlaine: nit: 1024 * 1024 to match `max_size: 64 * 1024,` below?

runtime/src/iobuf/pool.rs
```diff
@@ -0,0 +1,1984 @@
...
+    /// # Errors
+    ///
+    /// Returns `None` if:
+    /// - `capacity` exceeds the configured `max_size` (oversized request)
+    /// - The pool is exhausted for the required size class
+    ///
+    /// # Initialization
+    ///
+    /// The returned buffer contains **uninitialized memory**. Do not read from
+    /// it until data has been written. Use `resize()` to initialize with a
```
danlaine: I think this `resize()` reference is stale? Same for comment on `pub struct PooledBufMut`

danlaine (commented):
A few nits otherwise LGTM!

runtime/src/iobuf/pool.rs
andresilva: I think this should probably just be `64KB` like in storage, otherwise we might allocate too much memory by default. I made these changes when I was testing on `examples/flood`.

runtime/src/iobuf/pool.rs
andresilva: Yeah I removed `resize()` in one of the last commits because it was a bit complex with pooled buffers. Forgot to update some stale references.

runtime/src/iobuf/pool.rs
```diff
@@ -0,0 +1,1952 @@
...
+    pub fn with_defaults(registry: &mut Registry) -> Self {
+        Self::new(
+            BufferPoolConfig::for_network(),
+            BufferPoolConfig::for_storage(),
+            registry,
+        )
+    }
+
+    /// Returns the network buffer pool.
+    pub const fn network(&self) -> &BufferPool {
```
patrick-ogrady: I think this could be a MUCH more powerful mechanism if pools were registered with a `u8`?

runtime/src/deterministic.rs
```diff
@@ -854,6 +856,10 @@ impl Context {
         let network = AuditedNetwork::new(DeterministicNetwork::default(), auditor.clone());
         let network = MeteredNetwork::new(network, runtime_registry);
 
+        // Initialize buffer pools
+        let buffer_pools =
+            BufferPools::with_defaults(runtime_registry.sub_registry_with_prefix("buffer_pool"));
```
patrick-ogrady: nit: make `buffer_pool` a const

runtime/src/lib.rs
```diff
@@ -713,6 +716,12 @@ stability_scope!(BETA {
         /// Ensure all pending data is durably persisted.
         fn sync(&self) -> impl Future<Output = Result<(), Error>> + Send;
     }
+
+    /// Interface that any runtime must implement to provide buffer pools.
+    pub trait Pooling: Clone + Send + Sync + 'static {
```
patrick-ogrady: We can do better than this

patrick-ogrady: Should also change `RayonPoolSpawner`

runtime/src/tokio/runtime.rs
```diff
@@ -301,6 +302,11 @@ impl crate::Runner for Runner {
             }
         }
 
+        // Initialize buffer pools (before network, as network uses the pool)
+        let buffer_pools = crate::BufferPools::with_defaults(
+            runtime_registry.sub_registry_with_prefix("buffer_pool"),
```
patrick-ogrady: nit: use const

runtime/src/iobuf/pool.rs
```diff
@@ -0,0 +1,1952 @@
...
+    network: BufferPool,
+    storage: BufferPool,
+}
+
+impl BufferPools {
+    /// Creates buffer pools with the given configurations.
+    pub fn new(
+        network_config: BufferPoolConfig,
+        storage_config: BufferPoolConfig,
+        registry: &mut Registry,
```
patrick-ogrady: put metrics first?

runtime/src/iobuf/pool.rs
```diff
@@ -0,0 +1,1952 @@
...
+}
+
+/// Composite type holding buffer pools for different I/O domains.
+#[derive(Clone)]
+pub struct BufferPools {
+    network: BufferPool,
+    storage: BufferPool,
+}
+
+impl BufferPools {
```
patrick-ogrady: need wrapper for this? `get_network_pool`?

runtime/src/iobuf/pool.rs
```diff
@@ -0,0 +1,1952 @@
...
+
+impl BufferPoolInner {
+    /// Try to allocate a buffer from the given size class.
+    ///
+    /// Retries a few times on CAS contention to avoid unnecessary fallback
+    /// to heap allocations when pool capacity is available.
+    fn try_alloc(&self, class_index: usize) -> Option<AlignedBuffer> {
+        // Spurious CAS failures are possible under extreme contention, so we
+        // retry several times before giving up and falling back to heap.
+        const MAX_RETRIES: usize = 8;
```
patrick-ogrady: move out of function

runtime/src/iobuf/pool.rs
```diff
@@ -0,0 +1,1952 @@
...
+    }
+}
+
+impl BufferPool {
+    /// Creates a new buffer pool with the given configuration.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the configuration is invalid.
+    pub fn new(config: BufferPoolConfig, registry: &mut Registry) -> Self {
```
patrick-ogrady: uncharacteristic to make this `pub` without `impl Metrics` as the metrics?

runtime/src/iobuf/pool.rs
```diff
@@ -0,0 +1,1952 @@
...
+    classes: Vec<SizeClass>,
+    metrics: PoolMetrics,
+}
+
+impl BufferPoolInner {
+    /// Try to allocate a buffer from the given size class.
+    ///
+    /// Retries a few times on CAS contention to avoid unnecessary fallback
+    /// to heap allocations when pool capacity is available.
+    fn try_alloc(&self, class_index: usize) -> Option<AlignedBuffer> {
```
patrick-ogrady, line 392: seems like we are could make this a lot simpler if we just prealloc?

runtime/src/iobuf/pool.rs
```diff
@@ -0,0 +1,1952 @@
...
+            self.min_size >= self.alignment,
+            "min_size ({}) must be >= alignment ({})",
+            self.min_size,
+            self.alignment
+        );
+        assert!(
+            self.max_size >= self.min_size,
+            "max_size must be >= min_size"
+        );
+        assert!(self.max_per_class > 0, "max_per_class must be > 0");
```
patrick-ogrady: NonZero?

runtime/src/iobuf/pool.rs
```diff
@@ -0,0 +1,1952 @@
...
+    ///
+    /// Returns `None` if:
+    /// - `capacity` exceeds the configured `max_size` (oversized request)
+    /// - The pool is exhausted for the required size class
+    ///
+    /// # Initialization
+    ///
+    /// The returned buffer contains **uninitialized memory**. Do not read from
+    /// it until data has been written.
+    pub fn alloc(&self, capacity: usize) -> Option<IoBufMut> {
```
patrick-ogrady: Is this wrapping necessary?

runtime/src/iobuf/pool.rs
patrick-ogrady: We keep `Array/Span` in `codec` because we don't run this issue (and then we'd put something like this in `utils`.

patrick-ogrady:
I think `IoBuf` is interesting enough to live outside of `runtime` but will wait to advocate for that refactor until it works well within `runtime`.

runtime/src/deterministic.rs
patrick-ogrady: I see, `iouring_network`/`iouring_storage` don't do this...

runtime/src/iobuf/pool.rs
patrick-ogrady: remove `BufferPools`

runtime/src/utils/buffer/read.rs
```diff
@@ -99,9 +99,17 @@ impl<B: Blob> Read<B> {
...
-        buf.resize(bytes_to_read, 0);
+        buf.clear();
+        let buf = if buf.capacity() >= bytes_to_read {
+            // SAFETY: We just cleared the buffer and verified capacity.
+            // The bytes will be overwritten by read_at before being read.
+            unsafe { buf.set_len(bytes_to_read) };
+            buf
+        } else {
+            IoBufMut::zeroed(bytes_to_read)
+        };
```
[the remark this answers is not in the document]
andresilva: We are **writing** into uninitialized memory, which is fine.

runtime/src/iobuf/pool.rs
```diff
@@ -540,8 +541,15 @@ impl BufferPool {
...
-        self.try_alloc(capacity)
-            .unwrap_or_else(|_| IoBufMut::with_capacity(capacity))
+        self.try_alloc(capacity).unwrap_or_else(|_| {
+            // Fall back to an untracked aligned allocation.
+            // Using Weak::new() means the buffer won't be returned to the pool on drop.
+            let size = capacity
+                .next_power_of_two()
+                .max(self.inner.config.min_size.get());
+            let buffer = AlignedBuffer::new(size, self.inner.config.alignment.get());
+            IoBufMut::from_pooled(PooledBufMut::new(buffer, Weak::new()))
```
patrick-ogrady, line 541: If this is returned before other outstanding buffers, I wonder if we should add it to the pool and just drop who returns after?

patrick-ogrady: I think we can explore this in a follow-up (if ever).

andresilva: We could do this, and I guess it would optimize for replenishing the pool as fast as possible, one change we'd need to make here is to make sure we allocate to the closest size class, whereas right now in the fallback case we're allocating the exact requested capacity.

patrick-ogrady (dismissed):
I think we can still improve the interface here but this is reasonable enough to unblock continued work.

PR #2941, opened 2026-01-26
[Storage] [Runtime] Add configurable faultiness to deterministic storage

### LINKED ISSUE #2911: [Storage] [Runtime] Add deterministic, faulty storage implementation
Unit tests are driven by the deterministic runtime, whose storage implementation is kept in memory and never fails on write. A real storage implementation _can_ fail on write. We should add the ability to run the deterministic storage such that it is possible to deterministically cause it to fail on writes. The failed write may either fail completely (it does nothing) or fail after a partial write. This will allow us to better test crash scenarios in which storage writes fail. In particular, we should be able to use this in fuzz tests. In the fuzz tests, we can perform operations on the type that uses the deterministic storage, wait until it fails, and then assert that the resulting state is recoverable.

danlaine:
Resolves #2911.

* Adds `commonware-runtime::faulty::Storage` which implements `commonware-runtime::Storage`. This implementation wraps an underlying storage implementation and, based on its config and an RNG, deterministically injects failures.
* Adds `commonware-runtime::faulty::Blob`, the associated `Blob` type for `commonware-runtime::faulty::Storage`.
* Adds fuzz tests `storage/fuzz/fuzz_targets/journal_crash_recovery.rs` and `storage/fuzz/fuzz_targets/mmr_journaled_crash_recovery.rs` to fuzz crash (i.e. stop after failed/partial write) contiguous journals and journaled MMRs respectively.

### DIFF
--- a/runtime/src/deterministic.rs
+++ b/runtime/src/deterministic.rs
@@ -42,7 +42,7 @@
 //! });
 //! ```
 
-pub use crate::storage::faulty::FaultConfig;
+pub use crate::storage::faulty::Config as FaultConfig;
 use crate::{
     network::{
         audited::Network as AuditedNetwork, deterministic::Network as DeterministicNetwork,
--- a/runtime/src/storage/faulty.rs
+++ b/runtime/src/storage/faulty.rs
@@ -13,7 +13,7 @@ use std::{
 
 /// Operation types for fault injection.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
-enum FaultOp {
+enum Op {
     Open,
     Read,
     Write,
@@ -27,7 +27,7 @@ enum FaultOp {
 ///
 /// Each rate is a probability from 0.0 (never fail) to 1.0 (always fail).
 #[derive(Clone, Debug, Default)]
-pub struct FaultConfig {
+pub struct Config {
     /// Failure rate for `open_versioned` operations.
     pub open_rate: Option<f64>,
 
@@ -62,17 +62,17 @@ pub struct FaultConfig {
     pub scan_rate: Option<f64>,
 }
 
-impl FaultConfig {
+impl Config {
     /// Get the failure rate for an operation type.
-    fn rate_for(&self, op: FaultOp) -> f64 {
+    fn rate_for(&self, op: Op) -> f64 {
         match op {
-            FaultOp::Open => self.open_rate,
-            FaultOp::Read => self.read_rate,
-            FaultOp::Write => self.write_rate,
-            FaultOp::Sync => self.sync_rate,
-            FaultOp::Resize => self.resize_rate,
-            FaultOp::Remove => self.remove_rate,
-            FaultOp::Scan => self.scan_rate,
+            Op::Open => self.open_rate,
+            Op::Read => self.read_rate,
+            Op::Write => self.write_rate,
+            Op::Sync => self.sync_rate,
+            Op::Resize => self.resize_rate,
+            Op::Remove => self.remove_rate,
+            Op::Scan => self.scan_rate,
         }
         .unwrap_or(0.0)
     }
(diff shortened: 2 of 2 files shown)


runtime/src/deterministic.rs
```diff
@@ -844,10 +860,21 @@ impl Context {
             .timeout
             .map(|timeout| start_time.checked_add(timeout).expect("timeout overflowed"));
         let auditor = Arc::new(Auditor::default());
+
+        // Create shared RNG (used by both executor and storage)
+        let rng = Arc::new(Mutex::new(cfg.rng));
+
+        // Create storage fault config (default to disabled if None)
+        let storage_fault_config = Arc::new(RwLock::new(cfg.storage_faults.unwrap_or_default()));
```
patrick-ogrady: If there is a default, should we just set that in the config?

runtime/src/storage/faulty.rs
```diff
@@ -0,0 +1,698 @@
...
+use std::{
+    io::Error as IoError,
+    sync::{
+        atomic::{AtomicU64, Ordering},
+        Arc, Mutex, RwLock,
+    },
+};
+
+/// Type alias for a boxed dynamic RNG.
+pub type BoxDynRng = Box<dyn CryptoRngCore + Send + 'static>;
```
patrick-ogrady: Is this already defined somewhere?

runtime/src/deterministic.rs
danlaine: Removed `Option`

runtime/src/storage/faulty.rs
danlaine: Right you are, removed

storage/fuzz/fuzz_targets/journal_crash_recovery.rs
```diff
@@ -0,0 +1,594 @@
...
+}
+
+fn bounded_write_buffer(u: &mut Unstructured<'_>) -> Result<usize> {
+    u.int_in_range(1..=MAX_WRITE_BUF)
+}
+
+fn bounded_nonzero_rate(u: &mut Unstructured<'_>) -> Result<f64> {
+    let percent: u8 = u.int_in_range(1..=100)?;
+    Ok(f64::from(percent) / 100.0)
+}
```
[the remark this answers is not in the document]
danlaine: Eh, I think that might be overkill. We've tolerated a fair bit of repetition in fuzz tests to this point. I think keeping these (trivial) helpers in the files where they're used makes the test files somewhat more readable/self-documenting.

runtime/src/deterministic.rs
```diff
@@ -42,14 +42,15 @@
 //! });
 //! ```
 
+pub use crate::storage::faulty::FaultConfig;
```
patrick-ogrady: nit: `::Config as FaultConfig`?

runtime/src/storage/faulty.rs
```diff
@@ -0,0 +1,694 @@
...
+    io::Error as IoError,
+    sync::{
+        atomic::{AtomicU64, Ordering},
+        Arc, Mutex, RwLock,
+    },
+};
+
+/// Operation types for fault injection.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum FaultOp {
```
patrick-ogrady: nit: `Op`

runtime/src/storage/faulty.rs
```diff
@@ -0,0 +1,694 @@
...
+    Resize,
+    Remove,
+    Scan,
+}
+
+/// Configuration for deterministic storage fault injection.
+///
+/// Each rate is a probability from 0.0 (never fail) to 1.0 (always fail).
+#[derive(Clone, Debug, Default)]
+pub struct FaultConfig {
```
patrick-ogrady: nit: `Config`

runtime/src/storage/faulty.rs
```diff
@@ -0,0 +1,694 @@
...
+    /// Set the scan failure rate.
+    pub const fn scan(mut self, rate: f64) -> Self {
+        self.scan_rate = Some(rate);
+        self
+    }
+}
+
+/// Shared fault injection context.
+#[derive(Clone)]
+struct FaultCtx {
```
patrick-ogrady: nit: `Oracle`

runtime/src/storage/faulty.rs
danlaine: Done

runtime/src/storage/faulty.rs
danlaine: Done

runtime/src/storage/faulty.rs
danlaine: Done

Issue #2753, opened 2026-01-08
[coding] Parameterize `reed_solomon::Chunk` on `Digest` rather than `Hasher`


clabby:
## Overview

`reed_solomon::Chunk` is currently parameterized on `Hasher`, but following #2733, we can move it over to `Digest` to be in-line with the `mmr::Proof` and `bmt::Proof`:

https://github.com/commonwarexyz/monorepo/blob/ecd5e3c3f2de3e98c237e9dafd9b184af75f451e/coding/src/reed_solomon.rs#L41-L52

PR #2951, opened 2026-01-27
[storage] `bmt::Tree<Hasher>` -> `bmt::Tree<Digest>`

clabby:
## Overview

Updates `bmt::Tree` to be parameterized over a `Digest` rather than a `Hasher`, making the API more similar to the MMR.

closes https://github.com/commonwarexyz/monorepo/issues/2753

### DIFF
--- a/coding/src/reed_solomon.rs
+++ b/coding/src/reed_solomon.rs
@@ -1,7 +1,7 @@
 use crate::{Config, Scheme};
 use bytes::{Buf, BufMut};
 use commonware_codec::{EncodeSize, FixedSize, Read, ReadExt, ReadRangeExt, Write};
-use commonware_cryptography::Hasher;
+use commonware_cryptography::{Digest, Hasher};
 use commonware_parallel::Strategy;
 use commonware_storage::bmt::{self, Builder};
 use reed_solomon_simd::{Error as RsError, ReedSolomonDecoder, ReedSolomonEncoder};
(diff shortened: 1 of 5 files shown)


coding/src/reed_solomon.rs
```diff
@@ -80,22 +80,22 @@ impl<H: Hasher> Chunk<H> {
...
-impl<H: Hasher> Read for Chunk<H> {
+impl<D: Digest> Read for Chunk<D> {
     /// The maximum size of the shard.
     type Cfg = crate::CodecConfig;
 
     fn read_cfg(reader: &mut impl Buf, cfg: &Self::Cfg) -> Result<Self, commonware_codec::Error> {
         let shard = Vec::<u8>::read_range(reader, ..=cfg.maximum_shard_size)?;
         let index = u16::read(reader)?;
-        let proof = bmt::Proof::<H::Digest>::read_cfg(reader, &1)?;
+        let proof = bmt::Proof::<D>::read_cfg(reader, &1)?;
```
danlaine, line 98: We could elide the `<D>` but I'm fine keeping it too
```suggestion
        let proof = bmt::Proof::read_cfg(reader, &1)?;
```

coding/src/zoda.rs
```diff
@@ -676,7 +680,7 @@ impl<H: Hasher> Scheme for Zoda<H> {
         let checksum = Arc::new(data.mul(&checking_matrix));
 
         // Step 7: Produce the shards in parallel.
-        let shard_results: Vec<Result<Shard<H>, Error>> =
+        let shard_results: Vec<Result<Shard<H::Digest>, Error>> =
```
danlaine, line 683: We could elide the type but I'm fine keeping it too. (Same for `let row_hashes...` above.)
```suggestion
        let shard_results =
```

PR #2962, opened 2026-01-27
[cryptography] Use `aws-lc-rs` for ChaCha20-Poly1305 on x86_64/aarch64

clabby:
## Overview

Switch the handshake cipher from the pure-Rust chacha20poly1305 crate to aws-lc-rs on supported platforms (x86_64, aarch64). This leverages hardware-accelerated SIMD instructions for ~7x faster encrypt/decrypt.

| Message Size | `chacha20poly1305` | `aws-lc-rs` | Speedup |
|-------------:|-------------------:|------------:|--------:|
| 4 KB         | 29.7 µs            | 4.5 µs      | **6.6x** |
| 64 KB        | 466 µs             | 67.7 µs     | **6.9x** |
| 1 MB         | 7.4 ms             | 1.05 ms     | **7.1x** |

Handshake latency (key exchange + signatures) is unchanged (~170 µs). The original `chacha20poly1305` crate is retained as a fallback for other targets (e.g., wasm32). Conformance tests verify the output is identical.

### DIFF
--- a/cryptography/src/handshake/cipher.rs
+++ b/cryptography/src/handshake/cipher.rs
@@ -42,25 +42,13 @@ impl CounterNonce {
     }
 }
 
-/// ChaCha20-Poly1305 cipher backend abstraction for platform-specific implementations.
-trait Backend: Sized {
-    /// Creates a new cipher backend from a 256-bit key.
-    fn from_key(key: &[u8; KEY_SIZE_BYTES]) -> Self;
-
-    /// Encrypts plaintext using the given nonce, returning ciphertext with appended auth tag.
-    fn encrypt(&self, nonce: &[u8; NONCE_SIZE_BYTES], data: &[u8]) -> Result<Vec<u8>, Error>;
-
-    /// Decrypts ciphertext using the given nonce, verifying and stripping the auth tag.
-    fn decrypt(&self, nonce: &[u8; NONCE_SIZE_BYTES], data: &[u8]) -> Result<Vec<u8>, Error>;
-}
-
 cfg_if::cfg_if! {
     if #[cfg(any(target_arch = "x86_64", target_arch = "aarch64"))] {
         use aws_lc_rs::aead::{self, LessSafeKey, UnboundKey, CHACHA20_POLY1305};
 
         struct Cipher(LessSafeKey);
 
-        impl Backend for Cipher {
+        impl Cipher {
             fn from_key(key: &[u8; KEY_SIZE_BYTES]) -> Self {
                 let unbound_key = UnboundKey::new(&CHACHA20_POLY1305, key)
                     .expect("key size should match algorithm");
@@ -92,7 +80,7 @@ cfg_if::cfg_if! {
 
         struct Cipher(ChaCha20Poly1305);
 
-        impl Backend for Cipher {
+        impl Cipher {
             fn from_key(key: &[u8; KEY_SIZE_BYTES]) -> Self {
                 Self(ChaCha20Poly1305::new(key.into()))
             }


cryptography/src/handshake/cipher.rs
```diff
@@ -27,72 +29,159 @@ impl CounterNonce {
...
         if self.inner >= 1 << (8 * NONCE_SIZE_BYTES) {
             return Err(Error::MessageLimitReached);
         }
         let out = self.inner.to_le_bytes();
         self.inner += 1;
-        Ok(out)
+        // Extract only the lower 96 bits (12 bytes) for the nonce
+        let mut nonce = [0u8; NONCE_SIZE_BYTES];
+        let mut i = 0;
+        while i < NONCE_SIZE_BYTES {
```
patrick-ogrady: I think we can optimize this?

patrick-ogrady: (shouldn't we be able to just take out[:NONCE_SIZE_BYTES]?)

patrick-ogrady: or something of the sort

patrick-ogrady: (will handle myself)

clabby: It's const at the moment, but we can break that and use `copy_from_slice`.

clabby: (fwiw, optimizes away anyways.)

patrick-ogrady:
Wondering if its worth doing something like: https://github.com/commonwarexyz/monorepo/pull/2963

patrick-ogrady:
I actually got even better performance with this PR:

```
⏺ Comparison: Old (chacha20poly1305) vs New (aws-lc-rs)
  ┌───────────┬────────────────────────┬─────────────────┬─────────┐
  │   Size    │ Old (chacha20poly1305) │ New (aws-lc-rs) │ Speedup │
  ├───────────┼────────────────────────┼─────────────────┼─────────┤
  │ 4 KB      │ 24.3 µs                │ 4.77 µs         │ 5.1x    │
  ├───────────┼────────────────────────┼─────────────────┼─────────┤
  │ 64 KB     │ 1.25 ms                │ 75.3 µs         │ 16.6x   │
  ├───────────┼────────────────────────┼─────────────────┼─────────┤
  │ 1 MB      │ 23.9 ms                │ 1.16 ms         │ 20.6x   │
  ├───────────┼────────────────────────┼─────────────────┼─────────┤
  │ Handshake │ 291 µs                 │ 213 µs          │ 1.4x    │
  └───────────┴────────────────────────┴─────────────────┴─────────┘
  ```

**Comment by cronokirby** [comment]
Is it possible to not feature switch so much of the code? It seems undesirable to maintain twice the surface

**Comment by patrick-ogrady** [comment]
> Is it possible to not feature switch so much of the code? It seems undesirable to maintain twice the surface

A lot of similar but different types...I suppose we could try to abstract over them differently?

**Comment by cronokirby** [comment]
> > Is it possible to not feature switch so much of the code? It seems undesirable to maintain twice the surface
> 
> A lot of similar but different types...I suppose we could try to abstract over them differently?

Yeah, as a follow up we should create a shim to reduce duplication here, but this is fine for now

**Comment by clabby** [comment]
Will rebase on #2971 in a few.

File: cryptography/src/handshake/cipher.rs [hunk 2]
```diff
@@ -27,52 +28,128 @@ impl CounterNonce {
...
+
+        // Extract only the lower 96 bits (12 bytes) for the nonce
+        let mut nonce = [0u8; NONCE_SIZE_BYTES];
+        nonce.copy_from_slice(&out[..NONCE_SIZE_BYTES]);
+        Ok(nonce)
     }
 }
 
+/// ChaCha20-Poly1305 cipher backend abstraction for platform-specific implementations.
+trait Backend: Sized {
```
cronokirby: As the code is written, we don't actually need a trait, because the `Cipher` struct is instead defined conditionally, and used directly, rather than us having something that's generic over the trait.

patrick-ogrady: I liked the comments the trait afforded but admit is unnecessary machinery:

```
diff --git a/cryptography/src/handshake/cipher.rs b/cryptography/src/handshake/cipher.rs
index 1cc704136..095a5522c 100644
--- a/cryptography/src/handshake/cipher.rs
+++ b/cryptography/src/handshake/cipher.rs
@@ -42,25 +42,13 @@ impl CounterNonce {
     }
 }
 
-/// ChaCha20-Poly1305 cipher backend abstraction for platform-specific implementations.
-trait Backend: Sized {
-    /// Creates a new cipher backend from a 256-bit key.
-    fn from_key(key: &[u8; KEY_SIZE_BYTES]) -> Self;
-
-    /// Encrypts plaintext using the given nonce, returning ciphertext with appended auth tag.
-    fn encrypt(&self, nonce: &[u8; NONCE_SIZE_BYTES], data: &[u8]) -> Result<Vec<u8>, Error>;
-
-    /// Decrypts ciphertext using the given nonce, verifying and stripping the auth tag.
-    fn decrypt(&self, nonce: &[u8; NONCE_SIZE_BYTES], data: &[u8]) -> Result<Vec<u8>, Error>;
-}
-
 cfg_if::cfg_if! {
     if #[cfg(any(target_arch = "x86_64", target_arch = "aarch64"))] {
         use aws_lc_rs::aead::{self, LessSafeKey, UnboundKey, CHACHA20_POLY1305};
 
         struct Cipher(LessSafeKey);
 
-        impl Backend for Cipher {
+        impl Cipher {
             fn from_key(key: &[u8; KEY_SIZE_BYTES]) -> Self {
                 let unbound_key = UnboundKey::new(&CHACHA20_POLY1305, key)
                     .expect("key size should match algorithm");
@@ -92,7 +80,7 @@ cfg_if::cfg_if! {
 
         struct Cipher(ChaCha20Poly1305);
 
-        impl Backend for Cipher {
+        impl Cipher {
             fn from_key(key: &[u8; KEY_SIZE_BYTES]) -> Self {
                 Self(ChaCha20Poly1305::new(key.into()))
             }
```

patrick-ogrady: will make the change just to reduce the size

PR #2963, opened 2026-01-27
[stream] Interop Test

patrick-ogrady:


cryptography/src/handshake/cipher.rs
```diff
@@ -233,4 +249,45 @@ mod tests {
         let tag_only = vec![0u8; CIPHERTEXT_OVERHEAD];
         assert!(matches!(recv.recv(&tag_only), Err(Error::DecryptionFailed)));
     }
+
+    #[cfg(any(target_arch = "x86_64", target_arch = "aarch64"))]
+    mod interop {
```
patrick-ogrady, line 254: This should be implied by the unchanging conformance test output in #2962, so may be superflous? But could help ensure it remains consistent until we run conformance on more architectures?

PR #2973, opened 2026-01-28
[runtime] rename buffer::pool to buffer::paged

andresilva:
Rename the page cache module and related types/fields to better reflect their purpose:

- `buffer::pool` → `buffer::paged`
- `PoolRef` → `CacheRef`
- `buffer_pool` → `page_cache` (and similar field names)

Prepares for #2940


patrick-ogrady (changes requested):
1 naming change then should be good

consensus/src/aggregation/config.rs
```diff
@@ -9,7 +9,7 @@ use commonware_cryptography::{
 };
 use commonware_p2p::Blocker;
 use commonware_parallel::Strategy;
-use commonware_runtime::buffer::PoolRef;
+use commonware_runtime::buffer::paged::CacheRef;
```
patrick-ogrady, line 12: It annoyed me that it wasn't clear this cache was over paged (not just unchecksummed bytes -> error-prone imho).

PR #2974, opened 2026-01-28
[storage, runtime]: measure blob and archive latencies

SuperFluffy:
Provides metrics for measuring blob storage and archive performance.

For blobs this patched extends the `commonware_runtime::storage::metered` API and is probably at the lowest possible level and should thus hopefully get the most faithful picture of the performance impact that swapping out the underlying blob store should have (also, doing this at the `Journal`s level seemed infeasible).

Because we don't actually have `Context`s available at the time the metered storage is constructed I needed to provide `DeterministicClock` and `RuntimeClock` for the deterministic and tokio runtimes, respectively. This patch also changes the `Clock for tokio::Context` impl to be entirely in terms of `RuntimeClock` (since it's essentially just a marker trait this is ends up being very clean).

The other part of this patch is `commonware_storage::archive::Metered`, which measures all trait methods when delegating to its wrapped `T: Archive`. This wrapper is intended to be used by integrators when plugging their own objects into the marshal actor (for example, when swapping out the archive for a reth-provider).

Question 1: how do you prefer to test this?

Question 2: we can't really opt out of the blob metrics. I don't think they will have performance impact, but it's basically all-or-nothing. Should we make this configurable?

### DIFF
--- a/runtime/src/deterministic.rs
+++ b/runtime/src/deterministic.rs
@@ -300,7 +300,7 @@ pub struct Executor {
     metrics: Arc<Metrics>,
     auditor: Arc<Auditor>,
     rng: Mutex<BoxDynRng>,
-    time: Mutex<SystemTime>,
+    time: Arc<Mutex<SystemTime>>,
     tasks: Arc<Tasks>,
     sleeping: Mutex<BinaryHeap<Alarm>>,
     shutdown: Mutex<Stopper>,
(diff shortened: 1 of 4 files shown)


runtime/src/deterministic.rs
```diff
@@ -1391,6 +1394,33 @@ impl Pacer for Context {
     }
 }
 
+#[derive(Clone)]
+pub struct DeterministicClock(Arc<Mutex<SystemTime>>);
```
patrick-ogrady, line 1398: I would've thought the existing Clock impl on deterministic would've been sufficient?

SuperFluffy: Bit of a chicken-egg problem. We initialize `MeteredStorage` before we have a deterministic context to plug into it.

Similar problem as with the `runtime_registry` that's passed in directly and not through `Context: Metrics`.

clabby:

PR #2979, opened 2026-01-28
[storage/freezer] Add chain traversal limit to prevent infinite loops

syrupipy:
The collision chain traversal in get_key() had no upper bound - if storage gets corrupted and forms a cycle,   we'd spin forever. Added a 64k iteration limit that returns an error if hit. Normal chains are tiny (bounded by table_resize_frequency) so this only triggers on actual corruption.


patrick-ogrady:
If there was a cycle, this would be very broken. We don't need to bound search (if we did, it would mean we run the risk of not returning a value for a key we know).

PR #2981, opened 2026-01-28
[p2p] Avoid panic in Drop implementations

syrupipy:
Replace expect() with if-let in Drop implementations to avoid potential double-panic during unwinding. Affects SubReceiver and Reservation types.


patrick-ogrady:
Rust already ensures that `::drop()` will not be called twice. I don't believe these could panic during unwind?

syrupipy:
> Rust already ensures that `::drop()` will not be called twice. I don't believe these could panic during unwind?

Ah yeah, WOW you are the true RUST master

PR #2983, opened 2026-01-28
Towards unifying fuzz targets and module level tests

### LINKED ISSUE #2377: Add shim between proptest and fuzzing

cronokirby:
Closes #2377.

I implement this idea in the math crate, with the idea of expanding it to other places, e.g. cryptography, in subsequent work.

The pain felt was that there could be a lot of duplication between fuzz testing code and property testing code, and that we were not doing particularly intelligent fuzzing, nor using property testing enough.

To alleviate both of these at once, I wanted a way to easily define property tests, and then have those drive fuzz testing as well.

The implementation here has each crate create a top level "fuzzing plan" which an then be run both as a fuzz target, and as a property test.

To make it easy to run fuzzing for that plan inside of a module level test, I cooked up a little fuzzing library, which we can easily tweak as needed. For now, it works well enough. It can generate about 400k samples a second, which seems more than ample to not be bottlenecked by the actual time to run whatever function you're testing, so I'm not worried about that.

<sup><a href="https://cursor.com/dashboard?tab=bugbot">Cursor Bugbot</a> reviewed your changes and found no issues for commit <u>2d89bb9</u></sup>

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -35,11 +35,6 @@ jobs:
       continue-on-error: true
       env:
         CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
-    - name: Publish invariants
-      run: cargo publish --manifest-path invariants/Cargo.toml
-      continue-on-error: true
-      env:
-        CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
     - name: Publish parallel
       run: cargo publish --manifest-path parallel/Cargo.toml
       continue-on-error: true
@@ -52,6 +47,11 @@ jobs:
         CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
     - name: Publish utils
       run: cargo publish --manifest-path utils/Cargo.toml
+      continue-on-error: true
+      env:
+        CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
+    - name: Publish invariants
+      run: cargo publish --manifest-path invariants/Cargo.toml
       continue-on-error: true
       env:
         CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}


math/fuzz/fuzz_targets/main.rs
```diff
@@ -0,0 +1,9 @@
+#![no_main]
+
+libfuzzer_sys::fuzz_target!(|input: &[u8]| {
+    use arbitrary::{Arbitrary, Unstructured};
```
patrick-ogrady: import outside of loop?

math/src/lib.rs
```diff
@@ -19,5 +19,35 @@ commonware_macros::stability_scope!(BETA {
     pub mod poly;
 });
 
-#[cfg(test)]
-pub(crate) mod test;
+commonware_macros::stability_scope!(BETA {
+    #[cfg(any(test, feature = "fuzz"))]
+    pub(crate) mod test;
```
patrick-ogrady, line 24: Thus far, I've taken the approach that all testing infra should be `ALPHA`.

test/README.md
```diff
@@ -0,0 +1,10 @@
+# commonware-test
+
+[![Crates.io](https://img.shields.io/crates/v/commonware-test.svg)](https://crates.io/crates/commonware-test)
+[![Docs.rs](https://docs.rs/commonware-test/badge.svg)](https://docs.rs/commonware-test)
+
+Simplify testing and fuzzing of distributed systems.
+
+## Status
+
+`commonware-test` is **ALPHA** software and is not yet recommended for production use. Developers should expect breaking changes and occasional instability.
```
patrick-ogrady: Use the standard stability status line.

test/src/lib.rs
```diff
@@ -0,0 +1,8 @@
+//! Simplify testing and fuzzing of distributed systems.
```
patrick-ogrady: IIRC we also add the `Status` line here?

invariants/Cargo.toml
patrick-ogrady, line 1: Wondering if we should just place `minifuzz` in `utils`? I suppose the idea is that we'll have more testing utilities that don't make sense to put anywhere else?

Cargo.toml
```diff
@@ -18,6 +18,7 @@ members = [
     "runtime",
     "storage",
     "stream",
+    "test",
```
patrick-ogrady: I think we could call this "invariant"

patrick-ogrady: On the theme of performance

README.md
```diff
@@ -50,6 +50,7 @@ _Sometimes, we opt to maintain software that is neither a primitive nor an examp
 * [docker](./docker): Dockerfiles used for cross-compilation and CI.
 * [macros](./macros/README.md): Augment the development of primitives with procedural macros.
 * [mcp](./mcp/README.md): Interact with the Commonware Library via MCP at https://mcp.commonware.xyz.
+* [invariants](./invariants/README.md): Define and exercise invariants in your code
```
patrick-ogrady: fix alphabetical order

codec/src/lib.rs
```diff
@@ -215,10 +215,12 @@ commonware_macros::stability_scope!(BETA {
     pub use extensions::*;
 });
 
-#[cfg(feature = "arbitrary")]
-pub mod conformance;
+commonware_macros::stability_scope!(BETA {
```
patrick-ogrady: Why does this need to be exported?

coding/src/lib.rs
```diff
@@ -362,29 +373,41 @@ mod test {
...
-    proptest! {
-        #![proptest_config(ProptestConfig::with_cases(64))]
-
-        #[test]
-        fn proptest_roundtrip_reed_solomon(
-            (config, data, selected) in roundtrip_strategy()
-        ) {
-            roundtrip::<ReedSolomon<Sha256>>(&config, &data, &selected);
-        }
+    // Reed-Solomon requires extra_shards >= 1 (i.e., total > min).
```
patrick-ogrady: I think we can now remove this comment (as we use NonZero for all)?

coding/src/lib.rs
```diff
@@ -341,29 +352,41 @@ mod test {
...
+    #[test]
+    fn minifuzz_roundtrip_reed_solomon() {
+        minifuzz::Builder::default()
+            .with_seed(0)
+            .with_search_limit(64)
+            .test(|u| {
+                let (config, data, selected) = generate_case(u, 1)?;
+                roundtrip::<ReedSolomon<Sha256>>(&config, &data, &selected);
+                Ok(())
+            });
```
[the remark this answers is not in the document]
patrick-ogrady: Probably worth addressing ^

AGENTS.md
```diff
@@ -172,10 +172,11 @@ cargo llvm-cov --workspace --lcov --output-path lcov.info
 
 ## Testing Strategy
 
-- Unit tests: Core logic validation
-- Integration tests: Cross-primitive interaction
-- Fuzz tests: Input validation and edge cases
-- MIRI tests: Memory safety verification for unsafe code
+- Property: Core logic and invariant validation (use "minifuzz" from the invariants crate)
+- Unit: Testing edge cases or particular scenarios.
```
patrick-ogrady, line 176: remove `.` for consistency

invariants/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+[package]
+name = "commonware-invariants"
+edition.workspace = true
+publish = true
+version.workspace = true
+license.workspace = true
+description = "Define and exercise invariants in your code"
```
patrick-ogrady: nit: `Define and exercise invariants.`

invariants/README.md
```diff
@@ -0,0 +1,10 @@
+# commonware-invariants
+
+[![Crates.io](https://img.shields.io/crates/v/commonware-invariants.svg)](https://crates.io/crates/commonware-invariants)
+[![Docs.rs](https://docs.rs/commonware-invariants/badge.svg)](https://docs.rs/commonware-invariants)
+
+Define and exercise invariants in your code
```
patrick-ogrady: nit: `Define and exercise invariants.`

invariants/README.md
```diff
@@ -0,0 +1,10 @@
...
+
+## Status
+
+`commonware-invariants` is **ALPHA** software and is not yet recommended for production use. Developers should expect breaking changes and occasional instability.
```
patrick-ogrady, line 10: Use the new status language from other crates

invariants/src/minifuzz.rs
```diff
@@ -0,0 +1,564 @@
...
+//! is less extensive than a fuzz or property test.
+//!
+//! When using fuzzing in Rust, you have control over how random bytes are turned
+//! into inputs, so they effectively act more like property tests. This module provides
+//! a simple fuzzer, intended for unit tests. Unlike an actual fuzzer, this harness
+//! does not use coverage information, and is not going to run for as long, so it will
+//! naturally find fewer bugs. However, it can find many bugs pretty quickly,
+//! and can usually do a much better job than a unit test.
+//!
+//! In places where you'd normally write a unit test, you should instead consider
```
patrick-ogrady: nit: imho we don't want to encourage this as a replacement for specific edge case testing (as it only has probabilistic coverage) but a way to generalize an edge case?

invariants/src/minifuzz.rs
```diff
@@ -0,0 +1,564 @@
...
+//!         // same test body
+//!         Ok(())
+//!     });
+//! ```
+//!
+//! Use `with_search_limit` or `with_search_time` to control how long the fuzzer runs.
+//!
+//! # Why minifuzz?
+//!
+//! Minifuzz's goal is to both make it easier to write property tests, and to make
```
patrick-ogrady, line 32: Should we also clarify that it doesn't exercise invalid cases (like a fuzz test normally would)?

patrick-ogrady: In which case, is `fuzz` the wrong word?

invariants/src/minifuzz.rs
patrick-ogrady, line 1: Like `conformance`, I think we could call this `invariants` or just use `lib.rs`? I'm not sure what else you plan to add here?

invariants/src/minifuzz.rs
```diff
@@ -0,0 +1,564 @@
...
+    const fn new(seed: u64) -> Self {
+        Self {
+            seed: (seed >> 32) as u32,
+            thread: seed as u32,
+            size: 0,
+        }
+    }
+
+    fn try_from_hex(s: &str) -> Option<Self> {
+        let s = s.strip_prefix("0x").unwrap_or(s);
```
patrick-ogrady: We already have `hex` handling in utils

invariants/src/minifuzz.rs
```diff
@@ -0,0 +1,564 @@
...
+        let start = prefix_len + (self.rng.next_u32() as usize) % (self.buf.len() - prefix_len);
+        let len = ((self.rng.next_u32() as usize) % (self.buf.len() - start)).max(1);
+        let delta = self.rng.next_u32() as u8;
+        for b in &mut self.buf[start..start + len] {
+            *b = b.wrapping_add(delta);
+        }
+    }
+
+    fn pick_strategy(&mut self) -> u32 {
+        let mut weights = STRATEGY_WEIGHTS;
```
patrick-ogrady, line 254: do we need to assign a var here?

invariants/Cargo.toml
cronokirby: It's always tempting to add something to `utils`, but it's important to resist that temptation when there's actually a coherent domain starting to form. Having a different crate also lets us easily depend on this feature only for testing, improving build times.

coding/src/lib.rs
cronokirby: This was missed when we updated that, I'd prefer this change in a separate PR.

invariants/Cargo.toml
cronokirby: image

invariants/src/minifuzz.rs
cronokirby: Yeah, this language would be better as "you should _additionally_ consider".

invariants/src/minifuzz.rs
cronokirby: Yes, I prefer to think of our existing fuzz suites as being more like property tests, because we have control over input generation.

invariants/src/minifuzz.rs
cronokirby: In the future we might have strategies for particular types, more more advanced testing strategies, like whitebox fuzzing, symbolic checking, etc. Having a place for such contributions makes sense. If we were to collapse minifuzz down into lib.rs, it would make more sense, imo, to just have that be the crate itself, i.e. call it "commonware-minifuzz"

math/src/lib.rs
cronokirby: Yeah, this is a good point, I need to make the fuzz module inside of poly.rs ALPHA.

invariants/src/minifuzz.rs
cronokirby: Yes, in the function you use different weights depending on whether or not bytes are unused.

codec/src/lib.rs
cronokirby: It was exported before; the reason is that many crates use the utilities in this module to generate conformance tests exercising the codec implementation.

coding/src/lib.rs
cronokirby: Eh, might as well do this here.

invariants/Cargo.toml
patrick-ogrady: Ah, this is outdated. Sorry.

patrick-ogrady: Agree this is the right place.

invariants/src/minifuzz.rs
```diff
@@ -0,0 +1,603 @@
...
+                            tries >= limit || deadline.is_some_and(|d| Instant::now() >= d);
+                        if past_min && past_limit {
+                            break 'search;
+                        }
+                    }
+                }
+            }
+            branch = branch.next();
+            sampler.switch(branch);
+        }
```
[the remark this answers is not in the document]
cronokirby: Yes, this is intentional. The tradeoff is stalling rather than reporting a false positive.

PR #2988, opened 2026-01-29
[cryptography/dkg] provide getters for logs

SuperFluffy:
These would be very useful to have. I want to inspect the contents of the dealer logs I am reading from chain without having to rely on debug logs. So right now I have created shadow-types that have the same `Read` and `Write` impls so I can encode the upstream logs and decode into my custom types.

### DIFF
--- a/cryptography/src/bls12381/dkg.rs
+++ b/cryptography/src/bls12381/dkg.rs
@@ -1065,7 +1065,7 @@ impl<V: Variant, S: Signer> SignedDealerLog<V, S> {
         &self.dealer
     }
 
-    /// Returns the raw, unsigned [`DealerLog`] wrapped by [`Self`].
+    /// Returns an unsigned [`DealerLog`].
     ///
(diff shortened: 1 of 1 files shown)


cryptography/src/bls12381/dkg.rs
```diff
@@ -982,6 +982,14 @@ impl<V: Variant, P: PublicKey> Read for DealerLog<V, P> {
 }
 
 impl<V: Variant, P: PublicKey> DealerLog<V, P> {
+    pub fn commitment(&self) -> &Poly<V::Public> {
```
patrick-ogrady: Can you add comments to these new public functions?

SuperFluffy: Done. I also removed the `signature` getter because it was pretty useless.

cronokirby:
https://github.com/commonwarexyz/monorepo/pull/2992 would be a bit less intrusive

PR #2995, opened 2026-01-29
[storage::queue] Implement a durable queue

BrendanChou:
Queue provides a persistent message queue with at-least-once delivery semantics.

<sup><a href="https://cursor.com/dashboard?tab=bugbot">Cursor Bugbot</a> reviewed your changes and found no issues for commit <u>5cf4df5</u></sup>

### DIFF
--- a/storage/src/queue/storage.rs
+++ b/storage/src/queue/storage.rs
@@ -28,7 +28,7 @@ pub struct Config<C> {
     /// Codec configuration for encoding/decoding items.
     pub codec_config: C,
 
-    /// Page cache for caching data.
+    /// Page cache for buffering reads from the underlying journal.
     pub page_cache: CacheRef,
 
     /// Write buffer size for each section.
--- a/storage/src/journal/contiguous/variable.rs
+++ b/storage/src/journal/contiguous/variable.rs
@@ -73,7 +73,7 @@ pub struct Config<C> {
     /// [Codec] configuration for encoding/decoding items.
     pub codec_config: C,
 
-    /// Page cache for caching data.
+    /// Page cache for buffering reads from the underlying storage.
     pub page_cache: CacheRef,
 
     /// Write buffer size for each section.


BrendanChou:
@bugbot review

storage/src/queue/mod.rs
```diff
@@ -0,0 +1,117 @@
+//! A durable, at-least-once delivery queue backed by a journal.
```
patrick-ogrady: nit: link to which journal

storage/src/queue/mod.rs
```diff
@@ -0,0 +1,117 @@
...
+//!
+//! [Queue] provides a persistent message queue with at-least-once delivery semantics.
+//! Items are durably stored in a journal and will survive crashes. The reader must
+//! explicitly acknowledge each item after processing.
+//!
+//! On restart, the queue replays from the journal's pruning boundary. Items that were
+//! acknowledged but not yet pruned will be re-delivered. This provides simple, robust
+//! at-least-once delivery without complex metadata synchronization.
```
patrick-ogrady: nit: imho not worth leaking storage primitive decisions here (at least the "why not X" part)?

storage/src/queue/mod.rs
```diff
@@ -0,0 +1,117 @@
...
+//! - Parallel processing with multiple workers
+//! - Selective retries (one stuck item doesn't block others)
+//! - Flexible processing order
+//!
+//! Acknowledged items are tracked in-memory using an "ack floor" plus ranges of acked
+//! items above the floor. When items are acked contiguously from the floor, the floor
+//! advances. Pruning happens during [`Persistable::sync`](crate::Persistable::sync) and
+//! removes complete sections below the ack floor.
+//!
+//! Note: The ack state is not persisted. On restart, items not yet pruned will be
```
patrick-ogrady: I think you can remove this (you've more or less said the same thing a number of times).

storage/src/queue/shared.rs
```diff
@@ -0,0 +1,447 @@
...
+/// Panics if there are still outstanding references to the shared state
+/// (which shouldn't happen if both handles are passed).
+pub async fn destroy<E: Clock + Storage + Metrics, V: CodecShared>(
+    writer: QueueWriter<E, V>,
+    reader: QueueReader<E, V>,
+) -> Result<(), Error> {
+    // Drop reader first to close the channel
+    drop(reader);
+
+    // Extract the queue from the Arc<Mutex<...>>
```
patrick-ogrady: Isn't it possible there are writer clones outstanding?

storage/src/queue/shared.rs
```diff
@@ -0,0 +1,447 @@
...
+    /// Try to dequeue the next unacknowledged item without waiting.
+    ///
+    /// Returns `None` immediately if no unacked item is available.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the underlying storage operation fails.
+    pub async fn try_recv(&mut self) -> Result<Option<(u64, V)>, Error> {
+        // Drain any pending notifications (stop if channel empty or closed)
+        while self.notify.try_recv().is_ok() {}
```
patrick-ogrady: Curious if you could explain the rationale here? Is this just so we don't have unbounded growth of notify (if we don't use `recv)?

storage/src/queue/storage.rs
```diff
@@ -0,0 +1,1266 @@
...
+
+        // On restart, ack_floor is the pruning boundary (items below are deleted).
+        // acked_above is empty (in-memory state lost on restart).
+        let ack_floor = journal.pruning_boundary();
+        let acked_above = RMap::new();
+
+        debug!(ack_floor, size = journal.size(), "queue initialized");
+
+        // Set initial metric values
+        metrics.size.set(journal.size() as i64);
```
patrick-ogrady: it feels odd to store `metrics` in a separate file/struct if we just call into it here anyways?

storage/src/queue/storage.rs
```diff
@@ -0,0 +1,1266 @@
...
+        if self.acked_above.get(&position).is_some() {
+            return Ok(());
+        }
+
+        if position == self.ack_floor {
+            // Advance floor
+            self.ack_floor = position + 1;
+
+            // Consume any contiguous acked items
+            while self.acked_above.get(&self.ack_floor).is_some() {
```
patrick-ogrady: I suppose this could be made more efficient if we grouped contiguous items (what `rmap` does)...but this "should" be good enough for the happy path.

storage/src/queue/storage.rs
danlaine: How do you mean? We have `struct Metrics` in a bunch of places across the repository. Are you suggesting that we remove `Metrics` and put its fields inside `Queue`?

storage/src/queue/storage.rs
```diff
@@ -0,0 +1,1266 @@
...
+            let (p, item) = queue.dequeue().await.unwrap().unwrap();
+            assert_eq!(p, 15);
+            assert_eq!(item, vec![15]);
+
+            queue.destroy().await.unwrap();
+        });
+    }
+
+    #[test_traced]
+    fn test_multiple_sequential_prunes() {
```
danlaine: Based on the name I expected the test to call `prune` multiple times but it doesn't

storage/src/queue/storage.rs
```diff
@@ -0,0 +1,1266 @@
...
+                    let (p, item) = queue.dequeue().await.unwrap().unwrap();
+                    assert_eq!(p, i);
+                    assert_eq!(item, vec![i as u8]);
+                }
+
+                assert!(queue.dequeue().await.unwrap().is_none());
+                queue.destroy().await.unwrap();
+            }
+        });
+    }
```
danlaine, line 927: Perhaps we should assert that the pruning boundary after restart is what we expect? 

```suggestion
            let expected_pruning_boundary = {
                let mut queue = Queue::<_, Vec<u8>>::init(context.with_label("first"), cfg.clone())
                    .await
                    .unwrap();

                // Enqueue items across multiple sections (items_per_section = 10)
                for i in 0..25u8 {
                    queue.enqueue(vec![i]).await.unwrap();
                }

                // Ack items 0-14 to advance floor past section 0
                for i in 0..15 {
                    queue.ack(i).unwrap();
                }
                assert_eq!(queue.ack_floor(), 15);

                // Sync triggers pruning
                queue.sync().await.unwrap();

                // Verify pruning occurred
                let pruning_boundary = queue.journal.pruning_boundary();
                assert!(pruning_boundary > 0, "expected some pruning to occur");

                pruning_boundary
            };

            // Second session: only non-pruned items are available
            {
                let mut queue =
                    Queue::<_, Vec<u8>>::init(context.with_label("second"), cfg.clone())
                        .await
                        .unwrap();

                // ack_floor = pruning_boundary (items 0-9 were pruned)
                let pruning_boundary = queue.journal.pruning_boundary();
                assert_eq!(queue.ack_floor(), pruning_boundary);
                assert_eq!(pruning_boundary, expected_pruning_boundary);

                // Items from pruning_boundary to 24 are re-delivered
                for i in pruning_boundary..25 {
                    let (p, item) = queue.dequeue().await.unwrap().unwrap();
                    assert_eq!(p, i);
                    assert_eq!(item, vec![i as u8]);
                }

                assert!(queue.dequeue().await.unwrap().is_none());
                queue.destroy().await.unwrap();
            }
        });
    }
```

storage/fuzz/fuzz_targets/queue_operations.rs
```diff
@@ -0,0 +1,269 @@
...
+    let runner = deterministic::Runner::default();
+
+    runner.start(|context| async move {
+        let cfg = Config {
+            partition: "queue_operations_fuzz_test".to_string(),
+            items_per_section: NZU64!(5),
+            compression: None,
+            codec_config: ((0usize..).into(), ()),
+            page_cache: CacheRef::new(PAGE_SIZE, NZUsize!(PAGE_CACHE_SIZE)),
+            write_buffer: NZUsize!(1024),
```
danlaine: Should we have the fuzzer generate `items_per_section`, `PAGE_SIZE`, `PAGE_CACHE_SIZE`, and `write_buffer` size to cover more cases? We could bound them to reasonable values. See [here](https://github.com/commonwarexyz/monorepo/blob/main/storage/fuzz/fuzz_targets/journal_crash_recovery.rs#L100) for an example.

storage/src/queue/storage.rs
```diff
@@ -0,0 +1,1266 @@
...
+    /// directly advances the ack floor. Pruning happens during [Persistable::sync].
+    ///
+    /// # Arguments
+    ///
+    /// * `up_to` - The exclusive upper bound. Items at positions `[ack_floor, up_to)` are acknowledged.
+    ///
+    /// # Errors
+    ///
+    /// - Returns [Error::PositionOutOfRange] if `up_to > queue size`.
+    pub fn ack_up_to(&mut self, up_to: u64) -> Result<(), Error> {
```
danlaine, line 257: This method can cause this invariant to not hold:

```rs
    /// Position of the next item to dequeue.
    ///
    /// Invariant: `ack_floor <= read_pos <= journal.size()`
    read_pos: u64,
```

For example, if we enqueue 10 items, then dequeue 5 without acknowledging any then we have:
* `ack_floor == 0`
* `read_pos == 5`
* `size == 10`

Then if we do `ack_up_to(8)` we will have:
* `ack_floor == 8`
* `read_pos == 5`
* `size == 10`

Which violates the invariant. There's no bug in `dequeue` because it calls `is_acked`, but may still be a point of confusion.

storage/src/queue/storage.rs
```diff
@@ -0,0 +1,1266 @@
...
+                }
+
+                // Ack items 0, 1, 2 - but items_per_section=10, so no pruning
+                queue.ack(0).unwrap();
+                queue.ack(1).unwrap();
+                queue.ack(2).unwrap();
+                assert_eq!(queue.ack_floor(), 3);
+
+                queue.sync().await.unwrap();
+                drop(queue);
```
danlaine: Nit: this is a no-op because `queue` is about to go out of scope so this could be removed
```suggestion
```

storage/src/queue/mod.rs
BrendanChou: fixed in https://github.com/commonwarexyz/monorepo/pull/2995/changes/493fed6526779613fce20e3b9d225428fae80995

storage/src/queue/mod.rs
BrendanChou: fixed in https://github.com/commonwarexyz/monorepo/pull/2995/changes/493fed6526779613fce20e3b9d225428fae80995

storage/src/queue/storage.rs
BrendanChou: Renamed to `test_ack_across_sections` in https://github.com/commonwarexyz/monorepo/pull/2995/changes/493fed6526779613fce20e3b9d225428fae80995

storage/src/queue/storage.rs
BrendanChou: Fixed in https://github.com/commonwarexyz/monorepo/pull/2995/commits/60a34c067959abba786d40a5bbd6e502ae51d93a

storage/src/queue/storage.rs
BrendanChou: We have it both ways in a lot of different places in the repo, but I usually prefer it in a standalone struct since I find it way more clear what's going on when you do something like `self.metrics.size.set(...)` rather than `self.size.set(...)`. And I generally just prefer it going in a separate file but it could also just go in the same file.

storage/src/queue/mod.rs
BrendanChou: "why not x" fixed in https://github.com/commonwarexyz/monorepo/pull/2995/changes/493fed6526779613fce20e3b9d225428fae80995

the rest fixed in https://github.com/commonwarexyz/monorepo/pull/2995/changes/2fc4654787e35db93877eabc1ab7a0c726923672

storage/src/queue/shared.rs
BrendanChou: clarified in https://github.com/commonwarexyz/monorepo/pull/2995/changes/493fed6526779613fce20e3b9d225428fae80995

storage/src/queue/storage.rs
BrendanChou: It is rmap! I will fix it

BrendanChou: fixed in https://github.com/commonwarexyz/monorepo/pull/2995/commits/40704949d45168145ad0f2c290339c488ad929c0

storage/src/queue/storage.rs
BrendanChou: Fixed in https://github.com/commonwarexyz/monorepo/pull/2995/commits/493fed6526779613fce20e3b9d225428fae80995

storage/src/queue/storage.rs
BrendanChou: Fixed by editing the invariant comment in https://github.com/commonwarexyz/monorepo/pull/2995/changes/493fed6526779613fce20e3b9d225428fae80995#diff-0c7f82d3d8324da4f4c4bb571fdb5221371a2c757faf0e5ed9b220a5598de351R73

storage/src/queue/shared.rs
BrendanChou: writer clones accounted for here https://github.com/commonwarexyz/monorepo/pull/2995/commits/5cf4df5d4ea54e80749e093ed2f233a13c7fdea3

must drop them before destroy returns a non-error

BrendanChou: We could also consider just removing this function. If you need to destroy the journal, you're probably not using `shared` anyway, you may be just using the `Queue` directly... WDYT?

storage/src/queue/storage.rs
```diff
@@ -0,0 +1,1297 @@
...
+        self.read_pos = self.ack_floor;
+        debug!(
+            old_read_pos = old_pos,
+            new_read_pos = self.read_pos,
+            "reset read position"
+        );
+    }
+}
+
+impl<E: Clock + Storage + Metrics, V: CodecShared> Persistable for Queue<E, V> {
```
patrick-ogrady: Is this required? For a `journal`, I'm not sure this really buys us anything?

storage/src/queue/storage.rs
```diff
@@ -0,0 +1,1297 @@
...
+
+    /// Enqueue an item, returning its position.
+    ///
+    /// The item is durably persisted before returning. If this method returns
+    /// successfully, the item is guaranteed to survive crashes.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the underlying storage operation fails.
+    pub async fn enqueue(&mut self, item: V) -> Result<u64, Error> {
```
patrick-ogrady, line 168: How would you feel about adding an `append` and `sync` call to allow for batched addition before `sync/commit`?

storage/src/queue/storage.rs
```diff
@@ -390,37 +386,28 @@ impl<E: Clock + Storage + Metrics, V: CodecShared> Queue<E, V> {
...
     ///
-    /// Enqueued items are already durable (persisted in [Queue::enqueue]). This method
-    /// prunes items below the ack floor and commits the journal state.
-    ///
-    /// Items that were acked but not pruned (i.e., ack_floor hasn't crossed a section
-    /// boundary) will be re-delivered on restart. This is at-least-once semantics.
-    async fn commit(&mut self) -> Result<(), Self::Error> {
+    /// Faster than [Self::sync] but recovery may be required on startup if a crash
+    /// occurs before the next sync.
+    pub async fn commit(&mut self) -> Result<(), Error> {
```
patrick-ogrady: I don't think we need `commit` and `sync`? We don't do anything during `sync` that we don't do in `commit`?

BrendanChou: Yeah sorry idk why my changes got messed up. I did `flush` which just makes sure the writes complete durably (doesn't prune) without pruning, and then `commit` which does `flush` and `prune` durably

storage/src/queue/storage.rs
```diff
@@ -0,0 +1,1266 @@
...
+            return Ok(None);
+        }
+
+        let item = self.journal.read(self.read_pos).await?;
+        let pos = self.read_pos;
+        self.read_pos += 1;
+
+        debug!(position = pos, "dequeued item");
+        Ok(Some((pos, item)))
+    }
```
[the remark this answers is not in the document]
BrendanChou: Yes this would break if the API of RMap changed, but what can you do?

storage/src/queue/shared.rs
```diff
@@ -0,0 +1,509 @@
...
+        Ok(pos)
+    }
+
+    /// Flush all previously appended items to disk. After this returns
+    /// successfully, all appended items are durable.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the underlying storage operation fails.
+    pub async fn flush(&self) -> Result<(), Error> {
```
patrick-ogrady: nit: `sync` for compatibility?

storage/fuzz/fuzz_targets/queue_crash_recovery.rs
```diff
@@ -0,0 +1,460 @@
...
+    pending: Vec<u8>,
+
+    /// The ack floor at the time of last successful commit.
+    committed_ack_floor: u64,
+
+    /// Current in-memory ack floor (lost on crash).
+    current_ack_floor: u64,
+
+    /// Items per section (reserved for recovery bound calculations).
+    _items_per_section: u64,
```
patrick-ogrady: unused?

storage/src/queue/storage.rs
```diff
@@ -0,0 +1,1307 @@
...
+///
+/// # Operations
+///
+/// - [append](Self::append) / [flush](Self::flush): Write items to the journal
+///   buffer, then persist. Items are readable immediately after append (before flush),
+///   but are lost on restart if not flushed.
+/// - [enqueue](Self::enqueue): Append + flush in one step; the item is durable before return.
+/// - [dequeue](Self::dequeue) / [peek](Self::peek): Return the next unacked item in FIFO order.
+/// - [ack](Self::ack) / [ack_up_to](Self::ack_up_to): Mark items as processed (in-memory only).
+/// - [commit](Self::commit): Flush, then prune completed sections below the ack floor.
```
patrick-ogrady: why does `commit` prune?

storage/src/queue/storage.rs
```diff
@@ -0,0 +1,1307 @@
...
+    }
+
+    /// Persist appended items without pruning. Use [Self::commit] to
+    /// also prune acknowledged sections.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the underlying storage operation fails.
+    pub async fn flush(&mut self) -> Result<(), Error> {
+        self.journal.commit().await?;
```
patrick-ogrady: nit: just return commit()?

storage/src/queue/storage.rs
```diff
@@ -0,0 +1,1307 @@
...
+        debug!(position = pos, "dequeued item");
+        Ok(Some((pos, item)))
+    }
+
+    /// Like [Self::dequeue], but does not advance the read position.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the underlying storage operation fails.
+    pub async fn peek(&self) -> Result<Option<(u64, V)>, Error> {
```
patrick-ogrady: Is this used at all?

storage/src/queue/metrics.rs
```diff
@@ -0,0 +1,42 @@
...
+
+    /// Current ack floor position.
+    ///
+    /// All items below this position are considered acknowledged.
+    pub ack_floor: Gauge,
+
+    /// Number of out-of-order ack ranges currently tracked.
+    ///
+    /// High values indicate many sparse acknowledgments, which increases memory usage.
+    pub acked_above_ranges: Gauge,
```
patrick-ogrady: Why not just call this `pending` (from ack_floor to highest)?

patrick-ogrady: or even better, just `tip`?

storage/src/queue/shared.rs
```diff
@@ -0,0 +1,509 @@
...
+    }
+
+    /// Enqueue a batch of items with a single flush, returning positions
+    /// `[start, end)`. The lock is held for the full batch, so no reader can
+    /// see any item until the entire batch is durable.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if any append or the final flush fails.
+    pub async fn enqueue_bulk(&self, items: &[V]) -> Result<Range<u64>, Error>
```
patrick-ogrady: accept an iter

storage/src/queue/shared.rs
```diff
@@ -0,0 +1,509 @@
...
+    }
+
+    /// Flush all previously appended items to disk. After this returns
+    /// successfully, all appended items are durable.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the underlying storage operation fails.
+    pub async fn flush(&self) -> Result<(), Error> {
+        self.queue.lock().await.flush().await
```
patrick-ogrady: sync for consistency?

storage/src/queue/shared.rs
```diff
@@ -0,0 +1,509 @@
...
+    pub async fn ack_up_to(&self, up_to: u64) -> Result<(), Error> {
+        self.queue.lock().await.ack_up_to(up_to)
+    }
+
+    /// See [Queue::peek].
+    pub async fn peek(&self) -> Result<Option<(u64, V)>, Error> {
+        self.queue.lock().await.peek().await
+    }
+
+    /// Returns the current ack floor.
```
patrick-ogrady: see `ack_floor`?

storage/src/queue/shared.rs
```diff
@@ -0,0 +1,509 @@
...
+    pub async fn ack_floor(&self) -> u64 {
+        self.queue.lock().await.ack_floor()
+    }
+
+    /// Returns the current read position.
```
patrick-ogrady: see read_position?

storage/src/queue/shared.rs
```diff
@@ -0,0 +1,509 @@
...
+    }
+
+    /// Reset the read position to re-deliver all unacknowledged items.
+    pub async fn reset(&self) {
+        self.queue.lock().await.reset();
+    }
+
+    /// See [Queue::prune].
+    ///
+    /// See [Queue::prune] for details.
```
patrick-ogrady: nit: duplicate comment

storage/src/queue/shared.rs
```diff
@@ -0,0 +1,509 @@
...
+///         _ = shutdown => break,
+///     }
+/// }
+/// ```
+pub async fn init<E: Clock + Storage + Metrics, V: CodecShared>(
+    context: E,
+    cfg: Config<V::Cfg>,
+) -> Result<(QueueWriter<E, V>, QueueReader<E, V>), Error> {
+    let queue = Arc::new(Mutex::new(Queue::init(context, cfg).await?));
+    let (notify_tx, notify_rx) = mpsc::channel(1);
```
patrick-ogrady, line 233: this should be configurable?

patrick-ogrady: (otherwise no parallelism)?

storage/src/queue/mod.rs
```diff
@@ -0,0 +1,86 @@
...
+
+/// Errors that can occur when interacting with [Queue].
+#[derive(Debug, Error)]
+pub enum Error {
+    #[error("journal error: {0}")]
+    Journal(#[from] crate::journal::Error),
+
+    #[error("position out of range: {0} (queue size is {1})")]
+    PositionOutOfRange(u64, u64),
+}
```
patrick-ogrady, line 87: Add a conformance test

storage/src/queue/shared.rs
```diff
@@ -0,0 +1,509 @@
...
+    ///
+    /// Returns `None` immediately if no unacked item is available.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the underlying storage operation fails.
+    pub async fn try_recv(&mut self) -> Result<Option<(u64, V)>, Error> {
+        // Drain any pending notifications to keep the channel state consistent
+        // with the queue state when polling via try_recv.
+        while self.notify.try_recv().is_ok() {}
```
patrick-ogrady: Could we spin here forever with a busy writer?

storage/src/queue/shared.rs
```diff
@@ -0,0 +1,509 @@
...
+
+    /// Try to dequeue the next unacknowledged item without waiting.
+    ///
+    /// Returns `None` immediately if no unacked item is available.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the underlying storage operation fails.
+    pub async fn try_recv(&mut self) -> Result<Option<(u64, V)>, Error> {
+        // Drain any pending notifications to keep the channel state consistent
```
patrick-ogrady: This seems very error-prone to mix with `recv`? I think we should instead just take 1 item?

storage/src/queue/shared.rs
```diff
@@ -0,0 +1,509 @@
...
+    /// 2. Wait for the writer to enqueue new items if the queue is empty
+    /// 3. Return `None` if the writer is dropped (no more items will arrive)
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the underlying storage operation fails.
+    pub async fn recv(&mut self) -> Result<Option<(u64, V)>, Error> {
+        loop {
+            // Try to dequeue an item
+            if let Some(item) = self.queue.lock().await.dequeue().await? {
```
patrick-ogrady, line 138: This can cause the `self.notify` queue to grow unbounded I think?

patrick-ogrady: I suppose it will eventually be cleared in the next step if the writer is then locked?

storage/src/queue/shared.rs
```diff
@@ -0,0 +1,509 @@
...
+    ///
+    /// This method is designed for use with `select!`. It will:
+    /// 1. Return immediately if an unacked item is available
+    /// 2. Wait for the writer to enqueue new items if the queue is empty
+    /// 3. Return `None` if the writer is dropped (no more items will arrive)
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the underlying storage operation fails.
+    pub async fn recv(&mut self) -> Result<Option<(u64, V)>, Error> {
```
patrick-ogrady, line 135: Should we use the `Acknowledgement` pattern here? Then we don't need to "callback"?

storage/src/queue/shared.rs
```diff
@@ -0,0 +1,509 @@
...
+    pub async fn ack(&self, position: u64) -> Result<(), Error> {
+        self.queue.lock().await.ack(position)
+    }
+
+    /// See [Queue::ack_up_to].
+    ///
+    /// # Errors
+    ///
+    /// Returns [super::Error::PositionOutOfRange] if `up_to` is invalid.
+    pub async fn ack_up_to(&self, up_to: u64) -> Result<(), Error> {
```
patrick-ogrady, line 179: I'm not sure we need this complexity? I think `ack()` is enough?

storage/fuzz/fuzz_targets/queue_operations.rs
```diff
@@ -0,0 +1,332 @@
...
+
+fn bounded_write_buffer(u: &mut Unstructured<'_>) -> Result<usize> {
+    u.int_in_range(1..=MAX_WRITE_BUF)
+}
+
+#[derive(Arbitrary, Debug, Clone)]
+enum QueueOperation {
+    /// Enqueue a new item (append + flush).
+    Enqueue { value: u8 },
+
```
patrick-ogrady: IIRC other code doesn't include a `\n` after each item?

storage/fuzz/fuzz_targets/queue_crash_recovery.rs
```diff
@@ -0,0 +1,460 @@
...
+
+            run_operations(&mut queue, &operations, items_per_section_val).await
+        }
+    });
+
+    // Recovery phase - re-initialize queue from checkpoint
+    let runner = deterministic::Runner::from(checkpoint);
+    runner.start(|ctx| async move {
+        // Disable fault injection for recovery verification
+        *ctx.storage_fault_config().write().unwrap() = deterministic::FaultConfig::default();
```
patrick-ogrady, line 424: I'm actually pretty surprised this works 🤔

storage/src/queue/storage.rs
```diff
@@ -0,0 +1,1307 @@
...
+    ///
+    /// If set, items will be compressed before storage. Higher values provide
+    /// better compression but use more CPU.
+    pub compression: Option<u8>,
+
+    /// Codec configuration for encoding/decoding items.
+    pub codec_config: C,
+
+    /// Page cache for caching data.
+    pub page_cache: CacheRef,
```
patrick-ogrady, line 32: This is just for the indexes (glob isn't cached)

storage/src/queue/storage.rs
```diff
@@ -0,0 +1,1307 @@
...
+    /// All items at positions < ack_floor are considered acknowledged.
+    ///
+    /// On restart, this is initialized to `journal.bounds().start`.
+    ack_floor: u64,
+
+    /// Ranges of acknowledged items at positions >= ack_floor (in-memory only).
+    ///
+    /// When an item at position == ack_floor is acked, the floor advances
+    /// and any contiguous acked items are consumed. Lost on restart.
+    acked_above: RMap,
```
patrick-ogrady, line 88: nit: don't love the `_about`

patrick-ogrady (commented):
My remaining thoughts...planning to address these myself in a PR against #2995

storage/src/queue/storage.rs
BrendanChou: Good question I guess we could just delete the function, I thought it might be nice to have one function to flush all changes to disk but you could just call `flush` then `prune`. I guess in that case `flush` could also just be renamed to `commit`?

storage/src/queue/storage.rs
BrendanChou: You're right it can be deleted

storage/src/queue/metrics.rs
BrendanChou: It's a metric, so I think `pending_ranges` would be good. It's not the number of raw items acked above the floor since that takes `O(N)` to calculate on the number of ranges, it's the number of ranges tracked

storage/fuzz/fuzz_targets/queue_crash_recovery.rs
BrendanChou: Yeah can be deleted or modified to be actually used

storage/src/queue/shared.rs
BrendanChou: Yeah I think I was having an issue with lifetimes or it requiring the items to be `Sized` or something due to async iter

storage/src/queue/shared.rs
BrendanChou: It's just a place for notifying the reader that there could be new something in the queue, so it shouldn't need to be greater than `1` in size I think (i.e. if there's already a `()` in the channel, then it's fine to fail the send due to the channel being full). It's a bit like using an `AtomicBool` or other async flagging mechanism

storage/src/queue/shared.rs
BrendanChou: It's taking unlimited notifications. It takes only one item from the queue

storage/src/queue/shared.rs
BrendanChou: I believe `notify` is hardcoded to be size 1. I think it gains a size for each clone of it's sender, but otherwise cannot be unlimited size

storage/src/queue/shared.rs
BrendanChou: Ack is enough. I guess adding `ack_up_to` could be overengineering but is similar to how something like Kafka works with setting the "high water mark". Seems like a nicety but I guess we could remove it until we definitely use it ourselves

storage/src/queue/storage.rs
BrendanChou: You mean the `_above`? Just rename to `acked` would be fine yeah

storage/src/queue/shared.rs
patrick-ogrady: I can see how it would be useful in the general case...moreso wondering if we can get away with less to start.

storage/src/queue/shared.rs
patrick-ogrady: We have a struct for this in utils called `ring` (at least I think that's what it is called).

While it is just to "notify the reader", I think it'll block the writer?

storage/src/queue/storage.rs
patrick-ogrady: I guess I would expect `ack` to prune (and not tie deletion to commit)?

storage/fuzz/fuzz_targets/queue_crash_recovery.rs
```diff
@@ -0,0 +1,460 @@
...
+
+    /// Current in-memory ack floor (lost on crash).
+    current_ack_floor: u64,
+
+    /// Items per section (reserved for recovery bound calculations).
+    _items_per_section: u64,
+
+    /// Items that were appended but not yet flushed (position -> value).
+    /// These may be lost on crash. On flush, they move to committed.
+    unflushed: BTreeMap<u64, u8>,
```
patrick-ogrady: why not bool?

PR #2996, opened 2026-01-29
[math] Clarify that 2^lg_rows needs to be a usize

### LINKED ISSUE #2722: attempt to shift left with overflow in math/src/ntt.rs:846
Panicking when attempting to shift left with overflow in https://github.com/commonwarexyz/monorepo/blob/42352d5356cd51a7b7c0d00ae3ff28e2ec34dde3/math/src/ntt.rs#L846.

There is no documentation on whether this public function can panic, so it may be worth documenting that or adding input validation.

The harness source:

```
#![no_main]

use arbitrary::{Arbitrary, Unstructured};
use commonware_math::ntt::EvaluationVector;

const MAX_COLS: usize = 64;
const MAX_SAFE_LG_ROWS: usize = 16;

#[derive(Arbitrary, Debug)]
struct FuzzInput {
    lg_rows: u16,
    cols: u16,
}

fn fuzz(input: FuzzInput) {
    let lg_rows = input.lg_rows as usize;
    if lg_rows < usize::BITS as usize && lg_rows > MAX_SAFE_LG_ROWS {
        return;
    }
    let cols = (usize::from(input.cols) % (MAX_COLS + 1)).max(1);
    let _ = EvaluationVector::empty(lg_rows, cols);
}

fuzz_target!(|data: &[u8]| {
    let mut unstructured = Unstructured::new(data);
    let Ok(input) = FuzzInput::arbitrary(&mut unstructured) else {
        return;
    };
    fuzz(input);
});
```

cronokirby:
Closes #2722.

### DIFF
--- a/math/src/ntt.rs
+++ b/math/src/ntt.rs
@@ -842,7 +842,13 @@ impl EvaluationVector {
     }
 
     /// Create an empty element of this struct, with no filled rows.
+    ///
+    /// `2^lg_rows` must be a valid `usize`.
     pub fn empty(lg_rows: usize, cols: usize) -> Self {
+        assert!(
+            lg_rows < usize::BITS as usize,
+            "2^lg_rows must be a valid usize"
+        );
         let data = Matrix::zero(1 << lg_rows, cols);
         let active = BitMap::zeroes(data.rows as u64);
         Self {


math/src/ntt.rs
```diff
@@ -842,6 +842,8 @@ impl EvaluationVector {
     }
 
     /// Create an empty element of this struct, with no filled rows.
+    ///
+    /// `2^lg_rows` must be a valid `usize`.
```
patrick-ogrady, line 846: Does it make sure to add a test or assert this in some way?

cronokirby: Added an assertion to catch this directly, but this is a bit of a minor concern. You need this to match the number of rows you use for other operations, and more likely the bug you'll see is that you have the wrong value here, rather than an absurdly large one. But, I can imagine people accidentally passing in the number of rows or something like that, so the assertion will panic early.

cronokirby: Right now, without overflows on, you could have something silently off if it overflows.

PR #2954, opened 2026-01-27
[ci] Fix Warning Enforcement

patrick-ogrady:

PR #3000, opened 2026-01-30
[ci] Deny Warnings

patrick-ogrady:
Replaces: #2954 
Replaces: #2980

### DIFF
--- a/storage/src/journal/contiguous/variable.rs
+++ b/storage/src/journal/contiguous/variable.rs
@@ -597,13 +597,7 @@ impl<E: Clock + Storage + Metrics, V: CodecShared> Journal<E, V> {
     ///
     /// Unlike `destroy`, this keeps the journal alive so it can be reused.
     /// After clearing, the journal will behave as if initialized with `init_at_size(new_size)`.
-    #[cfg(not(any(
-        commonware_stability_BETA,
-        commonware_stability_GAMMA,
-        commonware_stability_DELTA,
-        commonware_stability_EPSILON,
-        commonware_stability_RESERVED
-    )))] // ALPHA
+    #[commonware_macros::stability(ALPHA)]
     pub(crate) async fn clear_to_size(&mut self, new_size: u64) -> Result<(), Error> {
         self.data.clear().await?;
 


storage/src/mmr/mem.rs
```diff
@@ -18,7 +18,8 @@ cfg_if::cfg_if! {
         use commonware_parallel::ThreadPool;
         use rayon::prelude::*;
     } else {
-        struct ThreadPool;
+        /// Placeholder for no_std builds where parallelism is unavailable.
```
patrick-ogrady, line 21: Add `TODO` for integrating commonware-parallel

patrick-ogrady: https://github.com/commonwarexyz/monorepo/issues/3001

Cargo.toml
```diff
@@ -52,6 +52,7 @@ rust-version = "1.91.1"
 unexpected_cfgs = { level = "warn", check-cfg = ['cfg(full_bench)', 'cfg(generate_conformance_tests)', 'cfg(commonware_stability_BETA)', 'cfg(commonware_stability_GAMMA)', 'cfg(commonware_stability_DELTA)', 'cfg(commonware_stability_EPSILON)', 'cfg(commonware_stability_RESERVED)'] }
 unused-must-use = "deny"
 rust-2018-idioms = "deny"
+warnings = "deny"
```
clabby, line 55: Just to be sure, this won't break publishing / building docs on `crates.io` / linting locally? We don't specify a `--cfg commonware_stability_X` in normal development, and that produces

```
warning: commonware-macros@0.0.65: stability cfg not set; set RUSTFLAGS="--cfg commonware_stability_X" (ALPHA/BETA/GAMMA/DELTA/EPSILON) to enforce stability gating
```

patrick-ogrady: For some reason, that doesn't trigger this.

clabby: Tested locally, doesn't disallow that warning (likely because it has an external source.)

patrick-ogrady: I think its just because we "print" rather than properly feeding this somewhere...checking now.

patrick-ogrady:
```
⏺ Right - the cargo:warning= from build.rs is a build script warning, not a Rust compiler warning. It shows up in the output but doesn't fail the build even with warnings = "deny".

  The [workspace.lints.rust] warnings = "deny" only affects Rust compiler warnings (unused imports, dead code, etc.), not build script messages.
```

PR #3003, opened 2026-01-30
[macros] commonware_macros::select! uses tokio::select!

andresilva:
This PR migrates `commonware_macros::select!` from `futures::select_biased!` to `tokio::select!` in biased mode.

To make this work, the macros crate is split into two: `macros/impl` (the proc-macro crate) and `macros/` (a wrapper crate that re-exports the macros and provides runtime dependencies via `__reexport`). This pattern allows the generated code to reference `tokio::select!` transitively without requiring users to add tokio as an explicit dependency. The wrapper crate is what external code imports as `commonware-macros`.

A clippy lint (`disallowed_macros`) is added to prevent direct usage of `futures::select!` and `futures::select_biased!`, ensuring all code uses our wrapper.

One notable quirk discovered during migration: `tokio::select!` imports `Poll::{Ready, Pending}` internally, which shadows any local type named `Pending`. This required using `self::Pending` in `consensus/src/aggregation/engine.rs` to disambiguate.

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -7,6 +7,16 @@ inputs:
 runs:
   using: composite
   steps:
+    # Use /mnt for Cargo builds on Linux (has ~70GB vs ~14GB on /)
+    - name: Relocate Cargo directories to /mnt on Linux
+      if: runner.os == 'Linux'
+      shell: bash
+      run: |
+        sudo mkdir -p /mnt/target /mnt/.cargo/bin
+        sudo chown -R "$(id -u):$(id -g)" /mnt/target /mnt/.cargo
+        echo "CARGO_TARGET_DIR=/mnt/target" >> "$GITHUB_ENV"
+        echo "CARGO_HOME=/mnt/.cargo" >> "$GITHUB_ENV"
+        echo "/mnt/.cargo/bin" >> "$GITHUB_PATH"
     - name: Normalize cache key
       id: cache-key
       shell: bash
@@ -22,3 +32,4 @@ runs:
         key: ${{ steps.cache-key.outputs.key }}
         cache-on-failure: false
         cache-workspace-crates: true
+        cache-directories: ${{ env.CARGO_TARGET_DIR }}
--- a/.github/workflows/fast.yml
+++ b/.github/workflows/fast.yml
@@ -275,17 +275,17 @@ jobs:
     - name: Add WASM target
       run: rustup target add wasm32-unknown-unknown
     - name: Build cryptography
-      run: cargo build --target wasm32-unknown-unknown --release --manifest-path cryptography/Cargo.toml && du -h target/wasm32-unknown-unknown/release/commonware_cryptography.wasm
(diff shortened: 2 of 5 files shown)


consensus/src/aggregation/engine.rs
```diff
@@ -321,10 +321,10 @@ impl<
                 self.pending
                     .iter_mut()
                     .for_each(|(_, pending)| match pending {
-                        Pending::Unverified(acks) => {
+                        self::Pending::Unverified(acks) => {
```
clabby, line 324: Can be removed iiuc

consensus/src/aggregation/engine.rs
```diff
@@ -321,10 +321,10 @@ impl<
...
                             acks.retain(|epoch, _| *epoch >= min_epoch);
                         }
-                        Pending::Verified(_, acks) => {
+                        self::Pending::Verified(_, acks) => {
```
clabby, line 327: Same

macros/Cargo.toml
```diff
@@ -13,23 +13,12 @@ documentation = "https://docs.rs/commonware-macros"
...
 [dependencies]
-futures.workspace = true
-proc-macro-crate.workspace = true
-proc-macro2.workspace = true
-quote.workspace = true
-syn = { workspace = true, features = ["full"] }
-toml.workspace = true
-tracing.workspace = true
-tracing-subscriber.workspace = true
+commonware-macros-impl = { path = "impl" }
```
clabby: This won't work, `commonware-macros` can't be published with a local dependency. We'll need to publish `commonware-macros-impl` (add to `publish.yaml` above `macros`) and pull in the dep from the workspace.

consensus/src/aggregation/engine.rs
clabby: nvm, tried locally

andresilva: `tokio::select!` imports `Poll::{Ready, Pending}` internally, which shadows our own `Pending`.

macros/Cargo.toml
andresilva: Yep, updated.

macros/src/lib.rs
```diff
@@ -1,914 +1,27 @@
-//! Augment the development of primitives with procedural macros.
+//! Procedural macros for async select, stability annotations, and test utilities.
```
clabby: Fitting with other crates, let's keep the first line the description?
```suggestion
//! Augment the development of primitives with procedural macros.
```

macros/impl/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+
+[dependencies]
+proc-macro-crate.workspace = true
+proc-macro2.workspace = true
+quote.workspace = true
+syn = { workspace = true, features = ["full"] }
+toml.workspace = true
+
+[dev-dependencies]
+commonware-macros = { workspace = true, features = ["std"] }
```
clabby: This shouldn't create an issue for testing, but I think this breaks publishing. https://github.com/rust-lang/cargo/issues/4242

andresilva: I guess I need move the tests and doc-tests to `macros` rather than `macros-impl`?

PR #3005, opened 2026-01-30
[runtime] Support `PROXY`

### LINKED ISSUE #3002: [runtime + p2p] Add Support for `PROXY` Protocol
When running a node behind a proxy, the IP check in `p2p::lookup` can fail because the "source ip" is considered to be the proxy's IP rather than the external party making the connection.

We should add support for re-mapping IPs to the correct source behind a config (we should not default because a malicious peer could then send a message that maps to someone else's IP).

As part of this, we should use the true "source IP" for RL on IP ranges.

patrick-ogrady:
Resolves: #3002


runtime/src/network/iouring.rs
```diff
@@ -254,16 +267,42 @@ impl crate::Listener for Listener {
...
+            self.read_buffer_size,
+            self.pool.clone(),
+        );
+
+        // Parse PROXY header if configured and connection is from trusted proxy
+        let client_addr = if let Some(ref proxy_cfg) = self.proxy {
+            if proxy_cfg.is_trusted(tcp_addr.ip()) {
+                // Keep reading until we have enough data to parse the PROXY header
+                loop {
+                    io_stream.fill_buffer().await?;
```
patrick-ogrady: move this into the proxy handler?

runtime/src/network/proxy.rs
```diff
@@ -0,0 +1,119 @@
...
+//! ```
+//!
+//! # Security
+//!
+//! Only parse PROXY headers from explicitly trusted proxy IPs. Connections from
+//! untrusted sources return the TCP-level address unchanged. Invalid PROXY headers
+//! from trusted proxies cause the connection to be rejected.
+
+use crate::Error;
+/// Re-export for convenience so users don't need to add ipnet as a dependency.
```
patrick-ogrady, line 32: remove comment

runtime/src/network/tokio.rs
```diff
@@ -516,4 +586,100 @@ mod tests {
 
         reader.await.unwrap();
     }
+
+    // PROXY protocol integration tests
```
patrick-ogrady, line 590: remove this block

macros/src/lib.rs
```diff
@@ -256,6 +258,22 @@ pub use commonware_macros_impl::stability_mod;
...
+/// stability_scope!(ALPHA {
+///     pub fn alpha_only() {}
+/// } else {
+///     pub fn non_alpha_fallback() {}
+/// });
+///
+/// // Multi-level fallback
+/// stability_scope!(ALPHA {
+///     pub fn alpha_path() {}
+/// } else BETA {
```
patrick-ogrady, line 272: else if?

runtime/src/network/iouring.rs
```diff
@@ -254,16 +267,42 @@ impl crate::Listener for Listener {
...
         let fd = Arc::new(OwnedFd::from(stream));
+        let mut io_stream = Stream::new(
+            fd.clone(),
+            self.recv_submitter.clone(),
+            self.read_buffer_size,
+            self.pool.clone(),
+        );
+
+        // Parse PROXY header if configured and connection is from trusted proxy
+        let client_addr = if let Some(ref proxy_cfg) = self.proxy {
```
patrick-ogrady, line 278: nit: remote_addr

clabby:

PR #3006, opened 2026-01-30
[Utils] Use type state pattern in historical bitmap 

danlaine:
`utils::bitmap::historical::BitMap` has a batch API where `BatchGuard` mutably borrows a bitmap.

This PR changes the historical bitmap implementation to use type state instead. The "dirty" state is equivalent to the batch state in the existing code.

This change will make the historical bitmap more in line with the MMR implementations, and will make it possible to use the historical bitmap inside `bitmap::authenticated::BitMap`, which already follows a type state pattern.

### DIFF
--- a/justfile
+++ b/justfile
@@ -23,7 +23,7 @@ pre-pr: lint test-docs test check-stability
 
 # Fixes the formatting of the workspace
 fix-fmt *args='':
-    @find "{{source_dir()}}" -not \( -path '{{source_dir()}}/target/*' -prune \) -name '*.rs' -type f -print0 | xargs -0 {{ rustfmt }} {{ nightly_version }} --edition 2021 {{ args }}
+    find . -path ./target -prune -o -name '*.rs' -type f -print0 | xargs -0 {{ rustfmt }} {{ nightly_version }} --edition 2021 {{ args }}
 
 # Fixes the formatting of the `Cargo.toml` files in the workspace
 fix-toml-fmt:


utils/src/bitmap/historical/bitmap.rs
```diff
@@ -1,11 +1,86 @@
...
+///
+/// **The dirty state de-duplicates during operations, not at commit time.**
+///
+/// Operations that cancel out are handled automatically:
+///
+/// ```text
+/// Example 1: push + pop = no-op
+///   push(true)  → appended_bits=[true], projected_len=11
+///   pop()       → appended_bits=[], projected_len=10
+///   Result: dirty state unchanged from base!
```
roberto-bayardo: nit: drop the "!" (the AI must have been excited about this!)

utils/src/bitmap/historical/bitmap.rs
```diff
@@ -393,19 +846,15 @@ impl<const N: usize> BitMap<N> {
...
-        changes: &mut BTreeMap<usize, ChunkDiff<N>>,
-    ) {
-        for &bit in batch.modified_bits.keys() {
+    fn capture_modified_chunks(&self, changes: &mut BTreeMap<usize, ChunkDiff<N>>) {
+        for &bit in self.state.modified_bits.keys() {
             let chunk_idx = Prunable::<N>::unpruned_chunk(bit);
             changes.entry(chunk_idx).or_insert_with(|| {
                 // `modified_bits` only contains bits from the base region that existed
-                // at batch creation. Since current hasn't changed yet (we're still
+                // at dirty creation. Since current hasn't changed yet (we're still
```
roberto-bayardo: I think you can just say "..only contains bits from the base region, so the chunks must exist."  

(Since the base region is now immutable, it's given that it hasn't changed.)

utils/src/bitmap/historical/bitmap.rs
```diff
@@ -451,36 +896,36 @@ impl<const N: usize> BitMap<N> {
...
         };
 
         // Capture all chunks between the new and old endpoints.
-        // Skip chunks that were already pruned before this batch started.
+        // Skip chunks that were already pruned before this started.
         for chunk_idx in new_last_chunk..=old_last_chunk {
-            if chunk_idx < batch.base_pruned_chunks {
-                // This chunk was already pruned before the batch, skip it
+            if chunk_idx < self.state.base_pruned_chunks {
+                // This chunk was already pruned before, skip it
```
roberto-bayardo: Not really new in this PR, but why would you be able to pop a bit in a chunk that's been pruned? (I would expect to get an error if I tried that..)

utils/src/bitmap/historical/bitmap.rs
danlaine: Removed

utils/src/bitmap/historical/bitmap.rs
danlaine: Updated comment

utils/src/bitmap/historical/bitmap.rs
danlaine: Explanation courtesy of my pal Claude

Screenshot 2026-01-30 at 6 19 50 PM
l

roberto-bayardo: Somethign like `let start_idx = max(base_pruned_chunks, new_last_chunk)` might be a bit more intuitive here.

roberto-bayardo: Also seems it's just the last-chunk edge case, so "skip chunks" in the comment where chunks is plural is a bit confusing (I think that's what me wonder what was going on.)

A better comment might be "// handle the case where we popped all unpruned bits" or something.

danlaine: [Updated](https://github.com/commonwarexyz/monorepo/pull/3006/commits/63bfc215d5c9f0d610e274711bc2f0bb6847043c) :+1:

PR #3007, opened 2026-01-30
[math] fix no_std

andresilva:
Fix: https://github.com/commonwarexyz/monorepo/actions/runs/21526803149/job/62032468362?pr=2993#logs

### DIFF
--- a/math/src/fields/goldilocks.rs
+++ b/math/src/fields/goldilocks.rs
@@ -428,6 +428,8 @@ impl Field for F {
 #[cfg(test)]
 mod test {
     use super::*;
+    #[cfg(not(feature = "std"))]
+    use alloc::vec::Vec;
     use crate::algebra;
     use proptest::prelude::*;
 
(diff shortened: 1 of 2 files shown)


clabby (approved):

math/src/fields/goldilocks.rs
```diff
@@ -428,6 +428,8 @@ impl Field for F {
 #[cfg(test)]
 mod test {
     use super::*;
+    #[cfg(not(feature = "std"))]
```
patrick-ogrady, line 431: When are we running a test without std?

andresilva: https://github.com/commonwarexyz/monorepo/blob/main/.github/workflows/fast.yml#L72
https://github.com/commonwarexyz/monorepo/blob/main/.github/workflows/slow.yml#L33

clabby: Yeah - the `math` crate doesn't do `#![cfg_attr(not(any(feature = "std", test)), no_std)]`, it just does `#![cfg_attr(not(feature = "std"), no_std)]`

andresilva: The only crate with `no_std` that also gates on `test` is `parallel`.

clabby: I'm a fan of allowing `std` in test, but glad to follow-up on that change.

PR #3008, opened 2026-01-30
[Storage] Use`into_merkleized()` function instead of manual re-implementation

danlaine:


storage/src/qmdb/current/db.rs
```diff
@@ -309,16 +301,8 @@ where
...
-            log: self.any.log.merkleize(),
-            inactivity_floor_loc: self.any.inactivity_floor_loc,
-            last_commit_loc: self.any.last_commit_loc,
-            snapshot: self.any.snapshot,
-            durable_state: self.any.durable_state,
-            active_keys: self.any.active_keys,
-            _update: core::marker::PhantomData,
-        };
+        // Merkleize the any db
+        let mut any = self.any.into_merkleized();
```
roberto-bayardo: You could actually just have one definition of into_merkleized here by creating an impl block that allows either durability state.  In fact we should do the same in any/db.rs.

PR #3009, opened 2026-01-30
[Storage] Add type for `bitmap::authenticated` type state

danlaine:
Adds `pub trait State` for `bitmap::authenticated`, and uses this new  `State` in `bitmap::authenticated`, replacing `dirty_chunks: HashSet<usize>`, `cached_root: Option<D>,` with `state: S`. This is a better use of type state -- field like `root` and `dirty_chunks` only exist when the `BitMap` is in the appropriate state.

Similarly, adds `pub trait State` for `qmdb::current`. This type uses two associated types `AnyState` and `BitMapState` to tie together the type state types for the authenticated bitmap and the `qmdb::any`.

Also updates some usage of "dirty" and "clean" to use "unmerkleized" and "merkleized".

### DIFF
--- a/storage/src/qmdb/current/db.rs
+++ b/storage/src/qmdb/current/db.rs
@@ -109,12 +109,6 @@ where
     D: DurabilityState,
     Operation<K, V, U>: Codec,
 {
-    /// The number of operations that have been applied to this db, including those that have been
-    /// pruned and those that are not yet committed.
-    pub fn op_count(&self) -> Location {
-        self.any.op_count()
-    }
-
     /// Return the inactivity floor location. This is the location before which all operations are
     /// known to be inactive. Operations before this point can be safely pruned.
     pub const fn inactivity_floor_loc(&self) -> Location {
@@ -124,11 +118,6 @@ where
     /// Whether the snapshot currently has no active keys.
     pub const fn is_empty(&self) -> bool {
         self.any.is_empty()
-    }
-
-    /// Returns the location of the oldest operation that remains retrievable.
-    pub fn oldest_retained_loc(&self) -> Location {
-        self.any.oldest_retained_loc()
     }
 
     /// Get the metadata associated with the last commit.
@@ -365,7 +354,7 @@ where
 
         self.any.apply_commit_op(commit_op).await?;
 
-        Ok(start_loc..self.op_count())
+        Ok(start_loc..self.any.log.bounds().end)
     }
 
     /// Commit any pending operations to the database, ensuring their durability upon return.
@@ -500,8 +489,8 @@ where
 {
     type Value = V::Value;
 
-    fn op_count(&self) -> Location {
-        self.op_count()
+    fn bounds(&self) -> std::ops::Range<Location> {
+        self.any.bounds()
     }
 
     fn inactivity_floor_loc(&self) -> Location {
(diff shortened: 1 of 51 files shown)


storage/src/qmdb/current/db.rs
```diff
@@ -151,7 +184,7 @@ where
     Operation<K, V, U>: Codec,
 {
     pub const fn root(&self) -> H::Digest {
-        self.cached_root.expect("Cached root must be set")
+        self.state.root
```
danlaine, line 176: No more `.expect`! 😄

storage/src/bitmap/authenticated.rs
```diff
@@ -471,21 +503,24 @@ impl<E: Clock + RStorage + Metrics, D: Digest, const N: usize> CleanBitMap<E, D,
...
+        UnmerkleizedAuthenticatedBitMap {
             bitmap: self.bitmap,
             authenticated_len: self.authenticated_len,
             mmr: self.mmr.into_dirty(),
-            dirty_chunks: self.dirty_chunks,
             pool: self.pool,
-            cached_root: self.cached_root,
+            state: Unmerkleized {
+                dirty_chunks: HashSet::new(),
+            },
```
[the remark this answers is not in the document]
danlaine: Fair comment, but we still use `into_dirty` in the MMR/journal/bitmap code so I'm inclined to keep this as is for now

storage/src/bitmap/authenticated.rs
```diff
@@ -63,16 +98,21 @@ pub type DirtyBitMap<E, D, const N: usize> = BitMap<E, D, N, Dirty>;
 /// # Type States
 ///
 /// The bitmap uses the type-state pattern to enforce at compile-time whether the bitmap has pending
-/// updates that must be merkleized before computing proofs. CleanBitMap represents a bitmap
-/// whose root digest has been computed and cached. DirtyBitMap represents a bitmap with pending
-/// updates. A dirty bitmap can be converted into a clean bitmap by calling [DirtyBitMap::merkleize].
+/// updates that must be merkleized before computing proofs. [MerkleizedAuthenticatedBitMap] represents a bitmap
```
roberto-bayardo: long lines?

storage/src/bitmap/authenticated.rs
```diff
@@ -49,11 +49,46 @@ pub(crate) fn partial_chunk_root<H: Hasher, const N: usize>(
...
+
+impl private::Sealed for Unmerkleized {}
+impl<D: Digest> State<D> for Unmerkleized {
+    type MmrState = Dirty;
+}
 
-/// A bitmap in the dirty state (has pending updates not yet reflected in the root).
-pub type DirtyBitMap<E, D, const N: usize> = BitMap<E, D, N, Dirty>;
+/// A merkleized bitmap whose root digest has been computed and cached.
+pub type MerkleizedAuthenticatedBitMap<E, D, const N: usize> = BitMap<E, D, N, Merkleized<D>>;
```
roberto-bayardo: Merkeleization basically implies authenticated, so I wonder if we should shorten to `MerkleizedBitMap` and `UnmerkleizedBitMap`?

danlaine: Removed `Authenticated` from these names

storage/src/bitmap/authenticated.rs
danlaine: Updated to be <100 columns

roberto-bayardo (dismissed):

PR #3010, opened 2026-01-30
[justfile] don't use cargo fmt

andresilva:
`cargo fmt --all` can't find most files now because of `stability_scope!` and `stability_mod!` macros.


clabby (dismissed):

clabby:
Double checked - Works on mac, too.

patrick-ogrady (approved):

PR #3014, opened 2026-02-01
[runtime] IoBuf Pool Review

patrick-ogrady:
- [ ] Use buffer pool for stream/p2p message construction?

### DIFF
--- a/cryptography/src/handshake/cipher.rs
+++ b/cryptography/src/handshake/cipher.rs
@@ -55,28 +55,41 @@ cfg_if::cfg_if! {
                 Self(LessSafeKey::new(unbound_key))
             }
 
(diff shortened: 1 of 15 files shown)


runtime/src/lib.rs
```diff
@@ -716,12 +719,6 @@ stability_scope!(BETA {
         /// Ensure all pending data is durably persisted.
         fn sync(&self) -> impl Future<Output = Result<(), Error>> + Send;
     }
-
-    /// Interface that any runtime must implement to provide buffer pools.
-    pub trait Pooling: Clone + Send + Sync + 'static {
-        /// Returns the buffer pools for this runtime.
-        fn buffer_pools(&self) -> &BufferPools;
```
patrick-ogrady, line 723: I now see this is useful to reuse the same buffer pool for message sending.

PR #3017, opened 2026-02-01
[storage/qmdb] require batch api for database modifications

roberto-bayardo:
Remove the ability to create/update/delete a QMDB db outside of the batch API, allowing simplification and encouraging performant usage.

### DIFF
--- a/storage/src/qmdb/any/unordered/variable.rs
+++ b/storage/src/qmdb/any/unordered/variable.rs
@@ -320,12 +320,14 @@ pub(crate) mod test {
             let db = open_db(context.with_label("open1")).await;
             let root = db.root();
             let mut db = db.into_mutable();
-
-            for i in 0u64..ELEMENTS {
-                let k = Sha256::hash(&i.to_be_bytes());
-                let v = vec![(i % 255) as u8; ((i % 13) + 7) as usize];
-                db.write_batch([(k, Some(v.clone()))]).await.unwrap();
-            }
+            db.write_batch((0..ELEMENTS).map(|i| {
+                (
+                    Sha256::hash(&i.to_be_bytes()),
+                    Some(vec![(i % 255) as u8; ((i % 13) + 7) as usize]),
+                )
+            }))
+            .await
+            .unwrap();
 
             // Simulate a failure and test that we rollback to the previous root.
             drop(db);
--- a/storage/src/qmdb/benches/fixed/generate.rs
+++ b/storage/src/qmdb/benches/fixed/generate.rs
@@ -41,8 +41,7 @@ fn bench_fixed_generate(c: &mut Criterion) {
                             let ctx = context::get::<Context>();
                             let mut total_elapsed = Duration::ZERO;
                             for _ in 0..iters {
-                                let commit_frequency =
-                                    (operations / COMMITS_PER_ITERATION) as u32;
+                                let commit_frequency = (operations / COMMITS_PER_ITERATION) as u32;
                                 let duration = match variant {
                                     Variant::AnyUnorderedFixed => {
                                         let db = get_any_unordered_fixed(ctx.clone()).await;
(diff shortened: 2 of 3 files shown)


storage/src/qmdb/current/ordered/db.rs
```diff
@@ -285,6 +285,26 @@ where
...
+        let status = &mut self.status;
+        self.any
+            .write_batch_with_callback(iter, move |append: bool, loc: Option<Location>| {
+                status.push(append);
+                if let Some(loc) = loc {
+                    status.set_bit(*loc, false);
+                }
+            })
+            .await
+    }
```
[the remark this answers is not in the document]
roberto-bayardo: fixed

storage/src/qmdb/store/db.rs
```diff
@@ -412,6 +412,25 @@ where
...
+        &mut self,
+        iter: impl Iterator<Item = (K, Option<V>)> + Send,
+    ) -> Result<(), Error> {
+        for (key, value) in iter {
+            if let Some(value) = value {
+                self.update(key, value).await?;
+            } else {
+                self.delete(key).await?;
+            }
+        }
```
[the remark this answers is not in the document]
roberto-bayardo: fixed

storage/src/qmdb/any/mod.rs
```diff
@@ -838,6 +834,8 @@ pub(crate) mod test {
         db.destroy().await.unwrap();
     }
 
+    // Consolidated test that runs test_any_db_multiple_commits_delete_replayed on all 8 variants.
```
danlaine:
```suggestion
```

storage/src/qmdb/any/unordered/mod.rs
```diff
@@ -230,6 +150,18 @@ where
...
     }
+
+    /// Writes a batch of key-value pairs to the database.
+    ///
+    /// For each item in the iterator:
+    /// - `(key, Some(value))` updates or creates the key with the given value
+    /// - `(key, None)` deletes the key
+    pub async fn write_batch(
+        &mut self,
+        iter: impl Iterator<Item = (K, Option<V::Value>)>,
```
danlaine: If we change this to `IntoIterator` then we can remove `.into_iter()` from most of the call sites

storage/src/qmdb/any/mod.rs
roberto-bayardo: removed.

storage/src/qmdb/any/unordered/mod.rs
roberto-bayardo: done. I had tried that before but had issues with the Send requirement, but Opus 4.6 figured out the magic rust trait bounds required.

storage/src/qmdb/any/unordered/variable.rs
```diff
@@ -324,7 +324,7 @@ pub(crate) mod test {
             for i in 0u64..ELEMENTS {
                 let k = Sha256::hash(&i.to_be_bytes());
                 let v = vec![(i % 255) as u8; ((i % 13) + 7) as usize];
-                db.update(k, v.clone()).await.unwrap();
+                db.write_batch([(k, Some(v.clone()))]).await.unwrap();
```
danlaine: With the new API we could do:

```rs
            db.write_batch((0..ELEMENTS).map(|i| {
                (
                    Sha256::hash(&i.to_be_bytes()),
                    Some(vec![(i % 255) as u8; ((i % 13) + 7) as usize]),
                )
            }))
            .await
            .unwrap();
```

or otherwise combine all the key-value writes into one batch as opposed to doing `write_batch` `ELEMENTS` times.

I'm also fine keeping this as is / changing this in a subsequent PR.

danlaine (approved):

storage/src/qmdb/any/unordered/variable.rs
roberto-bayardo: changed

danlaine (approved):

PR #3021, opened 2026-02-02
[broadcast] Fix incorrect metric in cleanup_waiters

Fallengirl:
Fixed a bug in `cleanup_waiters()` where dropped waiters were incorrectly incrementing `metrics.get` instead of `metrics.subscribe`.

Waiters are created exclusively by `handle_subscribe`, so their cleanup should update the `subscribe` metric, not the `get` metric. This was causing inaccurate monitoring data for both metrics.


clabby:

PR #3030, opened 2026-02-03
[deployer] Extract duplicate file upload logic to shared function

syrupipy:
- Extract ~100 lines of identical file upload code from `create.rs` and `update.rs` into `s3::upload_instance_files()`
- Both commands now call the same function for uploading binaries and configs to S3

<sup><a href="https://cursor.com/dashboard?tab=bugbot">Cursor Bugbot</a> reviewed your changes and found no issues for commit <u>4a474c2</u></sup>

### DIFF
--- a/deployer/src/aws/s3.rs
+++ b/deployer/src/aws/s3.rs
@@ -12,8 +12,16 @@ use aws_sdk_s3::{
     Client as S3Client,
 };
 use commonware_cryptography::{Hasher as _, Sha256};
-use futures::{future::try_join_all, stream::{self, StreamExt, TryStreamExt}};
-use std::{collections::{BTreeMap, BTreeSet, HashMap}, io::Read, path::Path, time::Duration};
+use futures::{
+    future::try_join_all,
+    stream::{self, StreamExt, TryStreamExt},
+};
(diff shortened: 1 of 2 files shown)


deployer/src/aws/s3.rs
```diff
@@ -441,3 +441,135 @@ pub fn is_no_such_bucket_error(error: &Error) -> bool {
...
+) -> Result<InstanceFileUrls, Error> {
+    // Collect unique binary and config paths (dedup before hashing)
+    let mut unique_binary_paths: BTreeSet<String> = BTreeSet::new();
+    let mut unique_config_paths: BTreeSet<String> = BTreeSet::new();
+    for instance in instances {
+        unique_binary_paths.insert(instance.binary.clone());
+        unique_config_paths.insert(instance.config.clone());
+    }
+
+    // Compute digests concurrently for unique files only
```
patrick-ogrady, line 484: I think we should retain the log here `computing file digests`?

I think we could add to it a count (which is just the len of `unique_paths`)?

syrupipy: Done

PR #3031, opened 2026-02-03
Improve check-stability speed

cronokirby:
This improves the performance of the check by:

1. using a wrapper script, rather than RUSTFLAGS, to avoid recompiling deps,
2. using "check" instead of "build", to skip codegen.

### DIFF
--- a/justfile
+++ b/justfile
@@ -144,9 +144,9 @@ check-stability *args='':
         ln -sf "$(pwd)/scripts/rustc_stability_wrapper.sh" "target/stability-wrappers/wrapper_${name}"
     done
     if [ -z "$level" ]; then
-        for l in 1 2 3 4; do
-            echo "Checking commonware_stability_${LEVEL_NAMES[$l]}..."
-            COMMONWARE_STABILITY_LEVEL="${LEVEL_NAMES[$l]}" RUSTC_WORKSPACE_WRAPPER="target/stability-wrappers/wrapper_${LEVEL_NAMES[$l]}" cargo check --workspace --lib {{ stability_excludes }} $extra_args || exit 1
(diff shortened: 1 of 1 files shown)


clabby (approved):

justfile
```diff
@@ -140,14 +140,19 @@ check-stability *args='':
             extra_args="$all_args"
         fi
     fi
+    # Create level-specific wrapper symlinks so Cargo sees different fingerprints
+    mkdir -p target/stability-wrappers
+    for name in BETA GAMMA DELTA EPSILON; do
```
patrick-ogrady: I prefer we have consistency of using name/level in both spots.

I'm down to remove the "LEVEL_NAMES" array and just use BETA... inline?

cronokirby: Reduced the duplication here.

justfile
```diff
@@ -140,14 +138,19 @@ check-stability *args='':
             extra_args="$all_args"
         fi
     fi
+    # Create level-specific wrapper symlinks so Cargo sees different fingerprints
+    mkdir -p target/stability-wrappers
+    for name in "${LEVEL_NAMES[@]:1}"; do
+        ln -sf "$(pwd)/scripts/rustc_stability_wrapper.sh" "target/stability-wrappers/wrapper_${name}"
+    done
     if [ -z "$level" ]; then
         for l in 1 2 3 4; do
```
patrick-ogrady: `for name in` here too?

scripts/rustc_stability_wrapper.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
```
patrick-ogrady, line 1: add \n

PR #3033, opened 2026-02-03
[stream] use buffer pool

andresilva:
This PR integrates the buffer pool into the encrypted stream layer to reduce allocation overhead during message encryption and decryption.

Previously, every call to send or receive allocated a new vector for the ciphertext or plaintext. Now, buffers are obtained from the runtime's network buffer pool, which allows for buffer reuse across messages and reduces pressure on the allocator in high-throughput scenarios.

To support this, `send_in_place` and `recv_in_place` methods were added to the cipher API. These allow encrypting and decrypting directly into caller-provided buffers, which is necessary for working with pooled allocations.

Related #2883.

<sup><a href="https://cursor.com/dashboard?tab=bugbot">Cursor Bugbot</a> reviewed your changes and found no issues for commit <u>97cc9cd</u></sup>

### DIFF
--- a/stream/src/encrypted.rs
+++ b/stream/src/encrypted.rs
@@ -350,7 +350,6 @@ impl<I: Stream> Receiver<I> {
             self.max_message_size.saturating_add(TAG_SIZE),
         )
         .await?;
-
         let ciphertext_len = encrypted.remaining();
 
         // Allocate buffer from pool for decryption.


cryptography/src/handshake/cipher.rs
```diff
@@ -4,9 +4,12 @@ use rand_core::CryptoRngCore;
 use std::vec::Vec;
 use zeroize::Zeroizing;
 
+/// Size of the ChaCha20-Poly1305 authentication tag.
+pub const TAG_SIZE: usize = 16;
+
 /// The amount of overhead in a ciphertext, compared to the plain message.
 /// ChaCha20-Poly1305 uses a 128-bit (16 byte) authentication tag.
-pub const CIPHERTEXT_OVERHEAD: usize = 16;
+pub const CIPHERTEXT_OVERHEAD: usize = TAG_SIZE;
```
patrick-ogrady: Do we need both of these consts?

stream/src/encrypted.rs
```diff
@@ -314,18 +342,39 @@ pub struct Receiver<I> {
...
-        Ok(self.cipher.recv(encrypted.as_ref())?.into())
+        .await?;
+
+        let ciphertext_len = encrypted.remaining();
+
+        // Allocate buffer from pool for decryption.
+        let mut decryption_buf = self.pool.alloc(ciphertext_len);
+
+        // SAFETY: We will initialize exactly `ciphertext_len` bytes below.
+        unsafe { decryption_buf.set_len(ciphertext_len) };
```
patrick-ogrady: second-guessing this...is there some reason we don't just have `alloc` pre-do this?

cryptography/src/handshake/cipher.rs
```diff
@@ -171,9 +195,36 @@ impl RecvCipher {
...
+    /// Returns an error if:
+    /// - `encrypted_data.len() < TAG_SIZE`
+    /// - Too many messages have been received with this cipher
+    /// - The ciphertext was corrupted or tampered with
+    ///
+    /// In the last two cases, the `RecvCipher` will no longer be able to return
+    /// valid ciphertexts, and will always return an error on subsequent calls
+    /// to [`Self::recv`]. Terminating (and optionally reestablishing) the connection
+    /// is a simple (and safe) way to handle this scenario.
+    pub fn recv_in_place(&mut self, encrypted_data: &mut [u8]) -> Result<usize, Error> {
```
patrick-ogrady, line 202: imho this should go before `recv`

cryptography/src/handshake/cipher.rs
andresilva: My thinking was that `TAG_SIZE` is more implementation-specific, whereas `CIPHERTEXT_OVERHEAD` is more high-level. With the new `*_in_place` API it didn't make sense to return the tag as `[u8; CIPHERTEXT_OVERHEAD]` though. Maybe we just keep `TAG_SIZE` and remove `CIPHERTEXT_OVERHEAD`?

stream/src/encrypted.rs
andresilva: `alloc` returns uninitialized memory, so returning with the length set to the capacity would be a footgun. The reason we're doing `set_len` here is because we know that we'll only write into that uninitialized memory (reading would be UB). I updated the code to use `put`/`put_slice` from `BufMut` though, this way we don't need to `set_len` explicitly.

runtime/src/iobuf/mod.rs
```diff
@@ -774,15 +779,53 @@ impl IoBufsMut {
...
+            Self::Single(buf) => {
+                let mut result = pool.alloc(buf.len() + extra);
+                result.put_slice(buf.as_ref());
+                result
+            }
             Self::Chunked(bufs) => {
-                let total_len: usize = bufs.iter().map(|b| b.len()).fold(0, usize::saturating_add);
-                let mut result = IoBufMut::with_capacity(total_len);
+                let total: usize = bufs.iter().map(|b| b.len()).fold(0, usize::saturating_add);
+                let mut result = pool.alloc(total + extra);
```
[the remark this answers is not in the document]
andresilva: I believe we'll have usages for this once we add buffer pool support in storage (at least while we don't have vectored reads).

stream/src/encrypted.rs
```diff
@@ -303,31 +305,48 @@ pub struct Sender<O> {
...
-            encryption_buf.freeze(),
-            self.max_message_size.saturating_add(TAG_SIZE),
-        )
-        .await?;
-        Ok(())
+        // Write varint prefix into reserved space.
+        let mut prefix = &mut frame_buf.as_mut()[..];
+        UInt(ciphertext_len as u32).write(&mut prefix);
+
+        self.sink
```
patrick-ogrady: I don't like breaking through this interface 🤔

andresilva: I initially created a `send_frame_reserved` (or whatever I called it), but `send_frame` couldn't use it, so felt it was superfluous and decided to inline it here.

patrick-ogrady (approved):
I was tempted to squeeze more juice out of these changes but concluded it was better to do in a subsequent PR)

PR #2446, opened 2025-12-10
Refactor and adapt simplex fuzzer for other testing modes

dnkolegov-ar:
This PR adds the following:
- using the same logic for unit tests and for fuzz tests
- using the same tests will be used by the deterministic runtime fuzzer (external) with task order mutations
- adds documentation
- adds SmallScope, FutureScope, AnyScope, round-based injections strategies based on [ByzzFuzz](https://gleissen.github.io/papers/byzzfuzz.pdf) paper
- Twinable mode where the second twin is a mutating actor (Disruptor)
- adding property-based tests without coverage feedback
- prioritizes `connected` partitioning
- adding new configuration (N=4,F=3,C=1) - modeling a case when a correct node is partitioned with byzantine nodes
- support for resolver types
- mutation for proposals and view
- remove panic filtering
- fixes invariant that if a view has a finalization then it must also have notarization

PR #3035, opened 2026-02-03
[consensus] Simplex Fuzzing Nits

patrick-ogrady:
Related: #2446


consensus/src/simplex/actors/batcher/round.rs
```diff
@@ -132,12 +132,15 @@ impl<
                 // Try to reserve
                 match self.pending_votes.notarize(index) {
                     Some(previous) => {
-                        if previous != &notarize {
+                        if previous.proposal != notarize.proposal {
```
patrick-ogrady, line 135: Related: #2999

PR #3036, opened 2026-02-03
[consensus::simplex] Improve `nullify` documentation

### LINKED ISSUE #2928: [Low] Nullify is sent if the proposal is not valid
According to https://github.com/commonwarexyz/monorepo/blob/374285d20a3333acf06ffd954e6eb7eb2826568e/consensus/src/simplex/mod.rs#L33 `nullify` is sent only if `t_l`, `t_a` or `t_r` are fired.

At the same time, in  https://github.com/commonwarexyz/monorepo/blob/64d38b71f9d82d00a8ccbf32ff29b5911a1b5d4e/consensus/src/simplex/actors/voter/actor.rs#L904 `handle_timeout` is called also if `Verification failed for current view proposal` which is treated as `immediate timeout`.

From my perspective, this semantics are not obvious and in essence, is optimization that is also missed in the original protocol
```
If the timer fires, and player is still in iteration h, player sends to everyone a vote for the dummy block
```

That said, the suggestion is to add this into the informal description of the protocol in https://github.com/commonwarexyz/monorepo/blob/374285d20a3333acf06ffd954e6eb7eb2826568e/consensus/src/simplex/mod.rs#L1

BrendanChou:
Fixes #2928

### DIFF
--- a/consensus/src/simplex/mod.rs
+++ b/consensus/src/simplex/mod.rs
@@ -34,6 +34,7 @@
 //! * Cancel `t_l`
 //! * If the container's parent `c_parent` is notarized at `v_parent` and we have nullifications for all views
 //!   between `v` and `v_parent`, verify `c` and broadcast `notarize(c,v)`
+//!     * If verification of `c` fails, immediately broadcast `nullify(v)`
 //!
 //! Upon receiving `2f+1` `notarize(c,v)`:
 //! * Cancel `t_a`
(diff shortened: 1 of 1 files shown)


consensus/src/simplex/mod.rs
```diff
@@ -34,6 +34,7 @@
 //! * Cancel `t_l`
 //! * If the container's parent `c_parent` is notarized at `v_parent` and we have nullifications for all views
 //!   between `v` and `v_parent`, verify `c` and broadcast `notarize(c,v)`
+//!     * If verification of `c` fails, immediately broadcast `nullify(v)`
```
BrendanChou, line 37: Do we need this clarification here or is it covered by line 81-82 below as a deviation from the default simplex consensus?

patrick-ogrady: I think of that as a "summary" of interesting points here and that what is here should "stand alone".

PR #3039, opened 2026-02-04
[runtime/iobuf] Improve Coalesce Ergonomics

patrick-ogrady:
Related: #3033

### DIFF
--- a/runtime/src/iobuf/mod.rs
+++ b/runtime/src/iobuf/mod.rs
@@ -487,8 +487,11 @@ impl From<BytesMut> for IoBufMut {
 
 impl From<Bytes> for IoBufMut {
     fn from(bytes: Bytes) -> Self {
+        // try_into_mut() avoids copying if refcount == 1 (single owner).
+        // Falls back to BytesMut::from() which copies for shared data.
+        let bytes_mut = bytes.try_into_mut().unwrap_or_else(BytesMut::from);
         Self {
-            inner: IoBufMutInner::Owned(BytesMut::from(bytes)),
+            inner: IoBufMutInner::Owned(bytes_mut),
         }
     }
 }


stream/src/encrypted.rs
```diff
@@ -302,28 +303,19 @@ pub struct Sender<O> {
...
-    pub async fn send(&mut self, buf: impl Into<IoBufs>) -> Result<(), Error> {
-        let mut bufs = buf.into();
-        let ciphertext_len = bufs.remaining() + TAG_SIZE as usize;
-
-        // Allocate buffer from pool for ciphertext (plaintext + tag).
-        let mut encryption_buf = self.pool.alloc(ciphertext_len);
-
-        // Copy plaintext into buffer.
-        encryption_buf.put(&mut bufs);
+    pub async fn send(&mut self, buf: impl Into<IoBufsMut>) -> Result<(), Error> {
```
patrick-ogrady, line 305: We aren't really benefiting from this yet but seems worthwhile to support it for when we do?

Maybe we're getting ahead of things...

patrick-ogrady: This would mean that p2p sending (where we zero-copy for all recipients), would no longer be possible. I'm not sure if it is any more copies (but may be one less).

broadcast/src/buffered/engine.rs
```diff
@@ -145,7 +152,9 @@ impl<E: Clock + Spawner + Metrics, P: PublicKey, M: Committable + Digestible + C
 
     /// Inner run loop called by `start`.
     async fn run(mut self, network: (impl Sender<PublicKey = P>, impl Receiver<PublicKey = P>)) {
-        let (mut sender, mut receiver) = wrap(self.codec_config.clone(), network.0, network.1);
+        let pool = self.context.network_buffer_pool().clone();
```
patrick-ogrady, line 155: updating `wrap` to take `pool` (and then provide `IoBufsMut` to `stream`) was the cleanest way I could think of to get pretty decent performance OOB

runtime/src/iobuf/mod.rs
```diff
@@ -487,8 +487,11 @@ impl From<BytesMut> for IoBufMut {
 
 impl From<Bytes> for IoBufMut {
     fn from(bytes: Bytes) -> Self {
+        // try_into_mut() avoids copying if refcount == 1 (single owner).
+        // Falls back to BytesMut::from() which copies for shared data.
+        let bytes_mut = bytes.try_into_mut().unwrap_or_else(BytesMut::from);
```
patrick-ogrady, line 492: Other cool trick in this PR that could be pulled out (maybe some way to incorporate pool here too)

andresilva: This only works for `Bytes` that are backed by a vec (i.e. normal `Bytes`), in our case where we use `Bytes::from_owner` this will always copy. See https://docs.rs/bytes/latest/bytes/struct.Bytes.html#method.from_owner.

runtime/src/lib.rs
```diff
@@ -583,12 +583,12 @@ stability_scope!(BETA {
...
         /// Receive exactly `len` bytes from the stream.
         ///
-        /// The runtime allocates the buffer and returns it as `IoBufs`.
+        /// The runtime allocates the buffer and returns it as `IoBufsMut`.
         ///
         /// # Warning
         ///
         /// If the stream returns an error, partially read data may be discarded.
-        fn recv(&mut self, len: u64) -> impl Future<Output = Result<IoBufs, Error>> + Send;
+        fn recv(&mut self, len: u64) -> impl Future<Output = Result<IoBufsMut, Error>> + Send;
```
andresilva, line 591: In practice this will result in copying if you're using pooled buffers, and conflicts with my future ideas for registered buffers and multishot recv in io_uring (we'll have an `IoBufMut` registered with the kernel where the kernel pushes socket data, and then we will just freeze slices and give that back to the user).

stream/src/encrypted.rs
andresilva: I think it makes more sense to keep this immutable because as you said we can keep a single (frozen) message from the p2p layer all the way down here. This is the only point in the entire stack where we absolutely need to copy (because of encryption), so just allocating a new `IoBufMut` here (that is appropriately sized for the final payload) and copying makes more sense to me. If this takes `IoBufsMut` then it means we have two copies: one in p2p (since the mut message can't be shared), and another one here for the encryption.

runtime/src/iobuf/mod.rs
```diff
@@ -774,15 +789,53 @@ impl IoBufsMut {
...
     pub fn coalesce(self) -> IoBufMut {
+        self.coalesce_with(IoBufMut::with_capacity)
+    }
+
+    /// Coalesce all buffers into a single contiguous `IoBufMut`, using the pool
+    /// for allocation if multiple buffers need to be merged.
+    ///
+    /// Zero-copy if only one buffer. Uses pool allocation if multiple buffers.
+    pub fn coalesce_with_pool(self, pool: &BufferPool) -> IoBufMut {
+        self.coalesce_with(|len| pool.alloc(len))
```
andresilva, line 821: LGTM, should add to `IoBufs` as well.

runtime/src/iobuf/mod.rs
```diff
@@ -487,8 +487,11 @@ impl From<BytesMut> for IoBufMut {
 
 impl From<Bytes> for IoBufMut {
     fn from(bytes: Bytes) -> Self {
+        // try_into_mut() avoids copying if refcount == 1 (single owner).
+        // Falls back to BytesMut::from() which copies for shared data.
+        let bytes_mut = bytes.try_into_mut().unwrap_or_else(BytesMut::from);
         Self {
-            inner: IoBufMutInner::Owned(BytesMut::from(bytes)),
+            inner: IoBufMutInner::Owned(bytes_mut),
```
andresilva, line 494: `BytesMut::from(Bytes)` already does the optimization above if possible.

PR #3041, opened 2026-02-04
[runtime] Split `Blob::read_at` into `read_at` and `read_at_buf`

andresilva:
`Blob::read_at` previously required callers to pass a pre-initialized buffer, which meant every read site had to allocate and zero a buffer via `IoBufMut::zeroed(n)` even though it would be immediately overwritten. The alternative was to explicitly `set_len` which is an unsafe call.

This PR splits the `read_at` method into two:

- `read_at_buf(offset, buf, len)` reads `len` bytes into a caller-provided buffer, which lets callers reuse pooled or pre-existing buffers without having to initialize them.
- `read_at(offset, len)` allocates a buffer internally and delegates to `read_at_buf`. This is the simple API that most call sites currently use.

The tokio, io_uring and memory backends use a `BufferPool` for allocations in `read_at`. All other implementations (`utills::buffer` wrappers) currently allocate via `IoBufMut::with_capacity`.

### DIFF
--- a/storage/src/metadata/storage.rs
+++ b/storage/src/metadata/storage.rs
@@ -153,7 +153,8 @@ impl<E: Clock + Storage + Metrics, K: Span, V: Codec> Metadata<E, K, V> {
         }
 
         // Read blob
-        let buf = blob.read_at(0, len as usize).await?.coalesce();
+        let len: usize = len.try_into().expect("blob too large for platform");
+        let buf = blob.read_at(0, len).await?.coalesce();
 
         // Verify integrity.
         //


andresilva:
Looking at the diff this change might look redundant but it is deliberate: `len` is the read contract, `capacity` is just an allocation detail. Many call sites may reuse buffers or over-allocate for amortized performance, the pool can also hand back a buffer larger than requested (size-classed). If we inferred "how much to read" from capacity, we could silently read too much, change behavior depending on allocator/pool state, and make the API nondeterministic. Explicit `len` keeps `read_at` stable and safe, while still allowing callers to choose any buffer capacity (including pooled buffers). It also avoids requiring callers to use unsafe `set_len` or zero-fill just to perform a read (which was the problem with the existing API since it just infered how much to read from `buf.len()`). An alternative would be to make `read_at` append‑style (write at current `len` up to `capacity`), but that introduces its own problems: it makes behavior depend on buffer state, complicates chunked buffers, and is unfriendly to `O_DIRECT` (append offsets will most likely be unaligned).

runtime/src/lib.rs
```diff
@@ -688,18 +690,25 @@ stability_scope!(BETA {
...
         ///
         /// - The output `IoBufsMut` is the same as the input, with data filled from offset
-        /// - The total bytes read equals the total initialized length of the input buffer(s)
+        /// - The total bytes read equals `len`
+        /// - `len` must be less than or equal to the total capacity of the input buffer(s)
         fn read_at(
             &self,
             offset: u64,
             buf: impl Into<IoBufsMut> + Send,
+            len: usize,
```
andresilva, line 716: I went with `usize` here rather than `u64`, since it's tied to the capacity of the buffer (which must be `usize`).

andresilva, line 716: Would `offset, len, buf` be preferable?

runtime/src/iobuf/mod.rs
```diff
@@ -753,6 +753,55 @@ impl IoBufsMut {
...
+                        continue;
+                    }
+                    let cap = buf.capacity();
+                    let use_len = remaining.min(cap);
+                    buf.set_len(use_len);
+                    remaining -= use_len;
+                }
+            }
+        }
+    }
```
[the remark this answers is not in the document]
andresilva: `set_len` takes `usize`, so the user can never request a length bigger than `usize::MAX`, if we have more capacity than that (i.e. we overflowed `usize` while computing the total capacity), then there's no issue, we are guaranteed to be able to initialize `set_len(usize::MAX)`.

runtime/src/lib.rs
patrick-ogrady: Yes, we should do that.

runtime/src/deterministic.rs
```diff
@@ -910,6 +896,24 @@ impl Context {
             runtime_registry.sub_registry_with_prefix("storage_buffer_pool"),
         );
 
+        // Create storage fault config (default to disabled if None)
+        let storage_fault_config = Arc::new(RwLock::new(cfg.storage_faults));
+        let storage = MeteredStorage::new(
+            AuditedStorage::new(
+                FaultyStorage::new(
+                    MemStorage::new(storage_buffer_pool.clone()),
```
patrick-ogrady, line 904: As discussed during standup, I think its a great idea to include the buffer pool here to make things faster.

runtime/src/deterministic.rs
```diff
@@ -910,6 +896,24 @@ impl Context {
...
+                    rng.clone(),
+                    storage_fault_config,
+                ),
+                auditor.clone(),
+            ),
+            runtime_registry,
+        );
+
+        // Create network
+        let network = AuditedNetwork::new(DeterministicNetwork::default(), auditor.clone());
```
patrick-ogrady, line 914: Should we add the network buffer pool here?

runtime/src/iobuf/mod.rs
```diff
@@ -287,6 +287,21 @@ impl IoBufMut {
...
     }
 
+    /// Prepares the buffer for a read operation of `len` bytes.
+    ///
+    /// # Safety
+    ///
+    /// Caller must initialize all `len` bytes before the buffer is read.
+    pub(crate) unsafe fn prepare_read(&mut self, len: usize) -> Result<(), crate::Error> {
+        if len > self.capacity() {
+            return Err(crate::Error::BufferTooSmall);
```
patrick-ogrady: nit: I think we could add (capacity/len to this error)?

runtime/src/lib.rs
```diff
@@ -686,22 +688,32 @@ stability_scope!(BETA {
...
             &self,
             offset: u64,
+            len: usize,
             buf: impl Into<IoBufsMut> + Send,
         ) -> impl Future<Output = Result<IoBufsMut, Error>> + Send;
 
+        /// Read `len` bytes at `offset`, returning a buffer(s) with exactly `len` bytes
+        /// of data read from the blob starting at `offset`.
+        ///
+        /// For reusing buffers, use [`Blob::read_at_buf`].
```
patrick-ogrady: nit: `To reuse a buffer,...`

runtime/src/storage/audited.rs
```diff
@@ -78,20 +78,30 @@ pub struct Blob<B: crate::Blob> {
...
+
+    async fn read_at_buf(
         &self,
         offset: u64,
+        len: usize,
         buf: impl Into<IoBufsMut> + Send,
     ) -> Result<IoBufsMut, Error> {
         let buf = buf.into();
-        let len = buf.len() as u64;
         self.auditor.event(b"read_at", |hasher| {
```
patrick-ogrady: nit: `read_at_buf`

runtime/src/storage/iouring.rs
```diff
@@ -246,30 +257,44 @@ impl Blob {
...
     ) -> Result<IoBufsMut, Error> {
-        let input_buf = buf.into();
+        let mut input_buf = buf.into();
+        // SAFETY: `len` bytes are filled via io_uring read loop below.
+        unsafe { input_buf.prepare_read(len)? };
 
         // For single buffers, read directly into them (zero-copy).
         // For chunked buffers, use a temporary and copy to preserve the input structure.
-        let buf_len = input_buf.len();
+        let buf_len = len;
```
patrick-ogrady: remove?

storage/src/journal/segmented/variable.rs
```diff
@@ -250,7 +250,11 @@ impl<E: Storage + Metrics, V: CodecShared> Journal<E, V> {
     ) -> Result<(u64, u32, V), Error> {
         // Read varint header (max 5 bytes for u32)
         let (buf, available) = blob
-            .read_up_to(IoBufMut::zeroed(MAX_U32_VARINT_SIZE), offset)
+            .read_up_to(
+                offset,
+                MAX_U32_VARINT_SIZE,
+                IoBufMut::with_capacity(MAX_U32_VARINT_SIZE),
```
patrick-ogrady, line 256: should we have a version that just returns an IoBuf here instead?

storage/src/metadata/storage.rs
```diff
@@ -154,10 +153,7 @@ impl<E: Clock + Storage + Metrics, K: Span, V: Codec> Metadata<E, K, V> {
         }
 
         // Read blob
-        let buf = blob
-            .read_at(0, IoBufMut::zeroed(len as usize))
-            .await?
-            .coalesce();
+        let buf = blob.read_at(0, len as usize).await?.coalesce();
```
patrick-ogrady: Should the arg for `load` be `usize`?

This feels wrong to cast here?

Maybe we should instead change runtime to u64 to prevent this type of casting (all our storage is in u64) and then panic if we can't read into usize?

runtime/src/utils/buffer/paged/append.rs
```diff
@@ -292,7 +292,7 @@ impl<B: Blob> Append<B> {
             // Read the last page and parse its CRC record.
             let page_start = last_page_end - physical_page_size;
             let buf = blob
-                .read_at(page_start, IoBufMut::zeroed(physical_page_size as usize))
+                .read_at(page_start, physical_page_size as usize)
```
patrick-ogrady, line 295: Another case where u64 is more ergonomic

runtime/src/network/iouring.rs
```diff
@@ -447,7 +447,7 @@ impl crate::Stream for Stream {
         // SAFETY: We will write exactly `len` bytes before returning
         // (loop continues until bytes_received == len). The buffer contents
         // are uninitialized but we only write to it, never read.
-        unsafe { owned_buf.set_len(len) };
+        unsafe { owned_buf.prepare_read(len)? };
```
patrick-ogrady: Should we just remove `set_len`?

andresilva: It's still useful for low-level situations where you don't know how much you will read ahead-of-time, but you want to set the length after the fact, e.g. see `fill_buffer` in iouring network.

storage/src/journal/segmented/variable.rs
andresilva: I'm not sure about it because this will change once the page cache uses `IoBuf`s.

runtime/src/deterministic.rs
andresilva: There's no use for the pool right now in the way the mock stream is written, it just keeps sent data in a `BytesMut` and then does `split_to` on reads, so we never allocate.

runtime/src/lib.rs
```diff
@@ -686,22 +688,32 @@ stability_scope!(BETA {
...
         ) -> impl Future<Output = Result<IoBufsMut, Error>> + Send;
 
+        /// Read `len` bytes at `offset`, returning a buffer(s) with exactly `len` bytes
+        /// of data read from the blob starting at `offset`.
+        ///
+        /// To reuse a buffer(s), use [`Blob::read_at_buf`].
+        fn read_at(
+            &self,
+            offset: u64,
+            len: usize,
```
patrick-ogrady, line 716: align network traits with usize

runtime/src/network/tokio.rs
```diff
@@ -45,7 +45,7 @@ impl crate::Stream for Stream {
             // SAFETY: We will write exactly `len` bytes before returning
             // (read_exact fills the entire buffer). The buffer contents
             // are uninitialized but we only write to it, never read.
-            unsafe { buf.set_len(len) };
+            unsafe { buf.prepare_read(len)? };
```
patrick-ogrady: do a scan for `set_len()` to make sure we use it right?

runtime/src/iobuf/mod.rs
```diff
@@ -287,6 +287,22 @@ impl IoBufMut {
         }
     }
 
+    /// Prepares the buffer for a read operation of `len` bytes.
+    ///
+    /// # Safety
+    ///
+    /// Caller must initialize all `len` bytes before the buffer is read.
+    pub(crate) unsafe fn prepare_read(&mut self, len: usize) -> Result<(), crate::Error> {
```
patrick-ogrady: why not just use `set_len()` for this?

runtime/src/storage/iouring.rs
```diff
@@ -246,30 +257,43 @@ impl Blob {
...
+    async fn read_at_buf(
         &self,
         offset: u64,
+        len: usize,
         buf: impl Into<IoBufsMut> + Send,
     ) -> Result<IoBufsMut, Error> {
-        let input_buf = buf.into();
+        let mut input_buf = buf.into();
+        // SAFETY: `len` bytes are filled via io_uring read loop below.
+        unsafe { input_buf.prepare_read(len)? };
```
patrick-ogrady: with set_len() we can make this explicit that you provide a big enough buf? You're going to panic anyways?

runtime/src/storage/audited.rs
```diff
@@ -78,20 +78,30 @@ pub struct Blob<B: crate::Blob> {
 }
 
 impl<B: crate::Blob> crate::Blob for Blob<B> {
-    async fn read_at(
+    async fn read_at(&self, offset: u64, len: usize) -> Result<IoBufsMut, Error> {
+        self.auditor.event(b"read_at", |hasher| {
+            hasher.update(self.partition.as_bytes());
+            hasher.update(&self.name);
+            hasher.update(&offset.to_be_bytes());
+            hasher.update(&(len as u64).to_be_bytes());
```
patrick-ogrady: just len.to_be_bytes?

patrick-ogrady (dismissed):

patrick-ogrady (approved):

PR #3046, opened 2026-02-04
[consensus] Documentation: explicitly call out the determinism requirement of certify()

### LINKED ISSUE #3042: Determinism requirement only documented in consensus/src/simplex/mod.rs
`consensus/src/simplex/mod.rs` states that
```
_The decision returned by `certify` must be deterministic and consistent across all honest participants to ensure liveness._
```
but this critical requirement is not reflected in the trait-level API docs in [monorepo/consensus/src/lib.rs](https://github.com/commonwarexyz/monorepo/blob/57ddd8c8899a2198e2a22c0449adfef94b145c8c/consensus/src/lib.rs#L114), so the developers may miss it and introduce non-deterministic certification logic.

BrendanChou:
Fixes #3042

### DIFF
--- a/consensus/src/lib.rs
+++ b/consensus/src/lib.rs
@@ -140,8 +140,7 @@ commonware_macros::stability_scope!(BETA, cfg(not(target_arch = "wasm32")) {
         /// # Determinism Requirement
         ///
         /// The decision returned by `certify` must be deterministic and consistent across
-        /// all honest participants to ensure liveness. If honest participants disagree on
-        /// certification, the network may fail to reach the quorum needed for finalization.
+        /// all honest participants to ensure liveness.
         fn certify(
             &mut self,
             _round: Round,


clabby (approved):

consensus/src/lib.rs
```diff
@@ -136,6 +136,12 @@ commonware_macros::stability_scope!(BETA, cfg(not(target_arch = "wasm32")) {
         /// This is particularly useful for applications that employ erasure coding, which
         /// can override this method to delay or prevent finalization until they have
         /// reconstructed and validated the full block (e.g., after receiving enough shards).
+        ///
+        /// # Determinism Requirement
+        ///
+        /// The decision returned by `certify` must be deterministic and consistent across
+        /// all honest participants to ensure liveness. If honest participants disagree on
+        /// certification, the network may fail to reach the quorum needed for finalization.
```
patrick-ogrady: I "think" this could also lead to a stall (if half have exited view and have remain)? So more than just a loss of finality?

clabby: This is true, though this is captured in "... to ensure liveness" ? We could be more specific, though.

patrick-ogrady: I'm going to push a small tweak.

patrick-ogrady (approved):

consensus/src/lib.rs
patrick-ogrady: As a trait, it would be difficult to enumerate all ways this could mess up consensus. I think we can just say "required for liveness" and leave it there.

PR #3050, opened 2026-02-04
[Storage] [Utils] Use absolute indices rather than relative-to-pruning indices in bitmaps

danlaine:
Updates bitmaps to consume/return absolute chunk indices as opposed to relative-to-pruning indices. I find this to be confusing -- it seems more straightforward for the bitmap's API to deal only with absolute indices and handle absolute-to-relative conversions internally.

Renames `bitmap::BitMap::chunk` to `to_chunk_index`.

### DIFF
--- a/utils/src/bitmap/mod.rs
+++ b/utils/src/bitmap/mod.rs
@@ -162,6 +162,10 @@ impl<const N: usize> BitMap<N> {
 
     /// Get a reference to a chunk by its index in the current bitmap.
     /// Note this is an index into the chunks, not a bit.
+    ///
+    /// # Warning
+    ///
+    /// Panics if the `chunk` is out of bounds.
     #[inline]
     pub(super) fn get_chunk(&self, chunk: usize) -> &[u8; N] {
         assert!(


storage/src/bitmap/authenticated.rs
```diff
@@ -116,7 +118,8 @@ pub struct BitMap<
     /// The underlying bitmap.
     bitmap: PrunableBitMap<N>,
 
-    /// The number of bitmap chunks currently included in the `mmr`.
```
roberto-bayardo, line 119: Did you mean to delete this comment?  A bit unusual to start a function doc by stating an invariant rather than a general description.

utils/src/bitmap/mod.rs
roberto-bayardo, line 164: While you're here mind documenting this panics if the index is beyond the length?

storage/src/bitmap/authenticated.rs
danlaine: I suppose I consider `Chunks in range [0, authenticated_len) are in mmr` to just be a more formal statement of `The number of bitmap chunks currently included in the mmr`

utils/src/bitmap/mod.rs
danlaine: done

PR #3052, opened 2026-02-04
stability: add unstable public to justfile

cronokirby:
Now `just check-stability` checks everything, including the unstable public items.

Also added a change to generate public docs once, speeding the script up.

AGENTS was modified in order to tell them to use the one entry point for stability.

Also updates github workflows to use the entry point, so as to not run the script multiple times.

<sup><a href="https://cursor.com/dashboard?tab=bugbot">Cursor Bugbot</a> reviewed your changes and found no issues for commit <u>e64524b</u></sup>

### DIFF
--- a/scripts/find_unstable_public.sh
+++ b/scripts/find_unstable_public.sh
@@ -15,8 +15,6 @@ STABILITY_CFG="commonware_stability_RESERVED"
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 source "$SCRIPT_DIR/stability_helpers.sh"
 
-# Items to ignore (#[macro_export] macros can't be inside stability_scope due to Rust limitations)
-IGNORE_ITEMS="NZDuration|NZU8|NZU16|NZU32|NZU64|NZUsize"
 
 if ! command -v jq &> /dev/null; then
     echo "Error: jq is required but not installed." >&2
--- a/scripts/stability_helpers.sh
+++ b/scripts/stability_helpers.sh
@@ -5,6 +5,9 @@
 # Crates to exclude: examples, internal tooling, heavy external deps, fuzz targets, proc-macro crates
 SKIP_REGEX="commonware-bridge|commonware-chat|commonware-estimator|commonware-flood|commonware-log|commonware-sync|commonware-reshare|commonware-conformance|commonware-deployer|-fuzz$|-macros$"
 
+# Public items to ignore (#[macro_export] macros can't be inside stability_scope due to Rust limitations)
+IGNORE_ITEMS="NZDuration|NZU8|NZU16|NZU32|NZU64|NZUsize"
+
 # Returns --exclude flags for cargo commands
 stability_exclude_flags() {
     cargo metadata --no-deps --format-version=1 2>/dev/null | \


.github/workflows/fast.yml
```diff
@@ -135,12 +135,9 @@ jobs:
       run: just test-docs ${{ matrix.flags }} -p ${{ matrix.package }}
 
   Stability:
-    name: "Stability (${{ matrix.level }})"
+    name: "Stability"
     runs-on: ubuntu-latest
     timeout-minutes: 30
-    strategy:
-      matrix:
-        level: [BETA, GAMMA, DELTA, EPSILON]
```
patrick-ogrady, line 143: This was an explicit decision to run these checks concurrently.

justfile
```diff
@@ -153,4 +153,6 @@ check-stability *args='':
         echo "Checking commonware_stability_${LEVEL_NAMES[$level]}..."
         COMMONWARE_STABILITY_LEVEL="${LEVEL_NAMES[$level]}" RUSTC_WORKSPACE_WRAPPER="target/stability-wrappers/wrapper_${LEVEL_NAMES[$level]}" cargo check --workspace --lib {{ stability_excludes }} $extra_args
     fi
+    echo "Checking for unmarked public items..."
+    ./scripts/find_unstable_public.sh
```
patrick-ogrady: Ah good catch, we should add this to lint (forgot about that)

AGENTS.md
```diff
@@ -144,7 +144,7 @@ cargo build --target wasm32-unknown-unknown --release -p commonware-cryptography
 just miri <module>::
 
 # 7. Stability coverage (if adding public API)
```
patrick-ogrady: This is now probably better described as `Stability coverage and consistency` (and we should remove the `if adding public API)?

justfile
```diff
@@ -41,8 +41,8 @@ clippy *args='':
 fix-clippy *args='':
     cargo clippy --all-targets --fix --allow-dirty $@
 
-# Runs all lints (fmt, clippy, docs, and features.)
-lint: check-fmt clippy check-docs check-features
+# Runs all lints (fmt, clippy, docs, features, and stability.)
```
patrick-ogrady: nit: remove `.` right before `)`?

AGENTS.md
cronokirby: Merging with section 1, since the lints now do this.

justfile
```diff
@@ -143,12 +143,20 @@ check-stability *args='':
     for name in "${LEVEL_NAMES[@]:1}"; do
         ln -sf "$(pwd)/scripts/rustc_stability_wrapper.sh" "target/stability-wrappers/wrapper_${name}"
     done
+    # Handle COVERAGE as a special level that only runs find_unstable_public.sh
+    if [ "$first_arg" = "COVERAGE" ]; then
```
patrick-ogrady: As discussed, I think it is error-prone to treat `COVERAGE` as a level. I'm wondering if there is another way to represent this (maybe via `--XXX` to ensure we don't do for each level in CI)?

cronokirby: Probably cleanest to have a separate workflow for that

scripts/find_unstable_public.sh
```diff
@@ -23,10 +24,72 @@ if ! command -v jq &> /dev/null; then
     exit 1
 fi
 
+usage() {
+    cat >&2 <<'EOF'
+Usage: ./scripts/find_unstable_public.sh [crate-name...] [-p <crate>] [--package <crate>] [--package=<crate>]
```
patrick-ogrady: I decided to support `-p` when used with `cargo` commands in `justfile`

justfile
```diff
@@ -95,8 +95,9 @@ test-conformance *args='':
 regenerate-conformance *args='':
     RUSTFLAGS="--cfg generate_conformance_tests" just test --features arbitrary --profile conformance {{ args }}
 
-# Packages to exclude from stability checks (examples, fuzz targets)
-stability_excludes := "--exclude commonware-bridge --exclude commonware-chat --exclude commonware-estimator --exclude commonware-flood --exclude commonware-log --exclude commonware-reshare --exclude commonware-sync --exclude commonware-broadcast-fuzz --exclude commonware-codec-fuzz --exclude commonware-coding-fuzz --exclude commonware-collector-fuzz --exclude commonware-consensus-fuzz --exclude commonware-cryptography-fuzz --exclude commonware-p2p-fuzz --exclude commonware-runtime-fuzz --exclude commonware-storage-fuzz --exclude commonware-stream-fuzz --exclude commonware-utils-fuzz"
```
patrick-ogrady, line 99: Decided to `dedup` these to prevent drift.

PR #3055, opened 2026-02-05
[stream] Avoid chunked `IoBufs` in encrypted sender

andresilva:
Follow-up to #3033.

Adds a `send_frame_with` which extracts the core frame logic (size validation, varint prefix construction, and send) into a helper that delegates frame assembly to the caller via a closure. This allows `Sender::send` to allocate a single contiguous buffer for the entire frame, writing the prefix directly into reserved space rather than prepending it as a separate buffer.

The existing `send_frame` now delegates to `send_frame_with` using the chunked approach.

Related #2883.

### DIFF
--- a/stream/src/encrypted.rs
+++ b/stream/src/encrypted.rs
@@ -67,7 +67,7 @@ use commonware_cryptography::{
 };
 use commonware_macros::select;
 use commonware_runtime::{
-    Buf, BufMut, BufferPool, BufferPooler, Clock, Error as RuntimeError, IoBufs, Sink, Stream,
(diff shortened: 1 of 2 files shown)


runtime/src/iobuf/mod.rs
```diff
@@ -1263,8 +1263,19 @@ mod tests {
 
     #[test]
     fn test_iobufs_coalesce_with_pool() {
+        cfg_if::cfg_if! {
```
andresilva, line 1266: Forgot to do this in #3033, it does finish (https://github.com/commonwarexyz/monorepo/actions/runs/21703906741/job/62597582220), but it takes almost 2h 😅.

patrick-ogrady (approved):
Was tempted to pass in encoded prefix but this is ~optimal.

Issue #3056, opened 2026-02-05
stream e2e fuzz test panicked at stream/fuzz/fuzz_targets/e2e.rs:244


dnkolegov-ar:
```
cd stream
cargo fuzz run e2e
```

The fuzz test panics after refatoring in https://github.com/commonwarexyz/monorepo/pull/3033

```
thread '<unnamed>' (19789343) panicked at stream/fuzz/fuzz_targets/e2e.rs:244:55:
called `Result::unwrap()` on an `Err` value: HandshakeError(DecryptionFailed)
```

PR #3057, opened 2026-02-05
Fix panic in e2e fuzz test

dnkolegov-ar:
This PR fixes https://github.com/commonwarexyz/monorepo/issues/3056

### DIFF
--- a/stream/fuzz/fuzz_targets/e2e.rs
+++ b/stream/fuzz/fuzz_targets/e2e.rs
@@ -46,7 +46,7 @@ impl<'a> arbitrary::Arbitrary<'a> for Message {
         } else {
             MAX_CIPHERTEXT_SIZE as usize
         };
-        let len = u.int_in_range(0..=max_len)?;
+        let len = u.int_in_range(0..=max_len.min(u.len()))?;
         let msg = u.bytes(len)?.to_vec();
         let out = if authenticated {
             Self::Authenticated(direction, msg)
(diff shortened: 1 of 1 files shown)


stream/fuzz/fuzz_targets/e2e.rs
```diff
@@ -239,12 +247,18 @@ fn fuzz(input: FuzzInput) {
...
+                    let frame = recv_frame(a_in, MAX_CIPHERTEXT_SIZE).await.unwrap();
+                    send_frame(a_out, frame, MAX_CIPHERTEXT_SIZE).await.unwrap();
                     let data2 = receiver.recv().await.unwrap();
                     assert_eq!(data2.coalesce(), data.as_slice(), "expected data to match");
                 }
                 Message::Unauthenticated(direction, data) => {
+                    if stream_corrupted {
+                        continue;
+                    }
+                    if data.len() > MAX_CIPHERTEXT_SIZE as usize {
```
patrick-ogrady: Do we not bound data.len() prior to this?

dnkolegov-ar: We did not, but now I am not sure we need this. After this PR, we should still be able to find that bug.

dnkolegov-ar:
Should be updated and retested after https://github.com/commonwarexyz/monorepo/pull/3063

stream/fuzz/fuzz_targets/e2e.rs
patrick-ogrady: Given the other PR doesn't change the invariants, I'd suspect some of these changes are still useful?

dnkolegov-ar: Since it doesn't change the invariant, I suggest closing this PR, it was created before the real issue was identified.

dnkolegov-ar:
> Should be updated and retested after #3063

The original fuzz test can't reproduce the bug after the fix

stream/fuzz/fuzz_targets/e2e.rs
patrick-ogrady: I think it was just by chance that we hadn't hit this though? Looking closer.

PR #3063, opened 2026-02-05
[cryptopgrahy/handshake] increase nonce before checking message len

### LINKED ISSUE #3056: stream e2e fuzz test panicked at stream/fuzz/fuzz_targets/e2e.rs:244

andresilva:
Fix #3056.

### DIFF
--- a/cryptography/src/handshake/cipher.rs
+++ b/cryptography/src/handshake/cipher.rs
@@ -343,23 +343,34 @@ mod tests {
 
     #[test]
     fn test_nonce_sync_after_truncated_recv() {
-        // Regression test for nonce desynchronization bug.
         let mut send = SendCipher::new(&mut test_rng());
         let mut recv = RecvCipher::new(&mut test_rng());
 
-        // Send first message (sender nonce: 0 -> 1)
-        let _ = send.send(b"message 1").unwrap();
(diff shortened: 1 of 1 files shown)


cryptography/src/handshake/cipher.rs
```diff
@@ -200,10 +200,10 @@ impl RecvCipher {
     /// is a simple (and safe) way to handle this scenario.
     #[inline]
     pub fn recv_in_place(&mut self, encrypted_data: &mut [u8]) -> Result<usize, Error> {
+        let nonce = self.nonce.inc()?;
```
patrick-ogrady, line 203: Can you add a regression test for this?

cronokirby:
One question is if we want to support recovery in this case, or if we want to maintain the property that "If DecryptionFailed" is ever returned, then you will never be able to successfully decrypt. (This is the property currently upheld)

PR #3070, opened 2026-02-06
[cryptography/handshake] poison recv cipher on error

andresilva:


cronokirby (approved):

cronokirby:
The inspect is a nice trick to make sure to always poison it!

patrick-ogrady:
I thought this PR was going a bit further than we needed (in that it is handling behavior after returning an error on recv). I think we should attempt to minimize complexity outside of our operating window and would prefer to close this.

cronokirby:
If we decide against this, then we should update the documentation to be a bit more clear about the current behavior, which could be described as: "If a message is corrupted, then the receiver will return an error when receiving it, but will otherwise be ready to receive the next message produced by send"

andresilva:
I don't have a strong opinion either way, but the added complexity here is just a trivial conditional and a single `.inspect_err(|_|)`. In return, we get an interface that's more resilient to misuse by turning a documented usage rule into an enforced invariant, which feels like a good trade-off.

clabby:

PR #3072, opened 2026-02-06
[Storage] Refactor `qmdb::current` grafting

danlaine:
Consolidates and simplifies the code responsible for "grafting" `qmdb::current`'s activity status bitmap onto the operations MMR.

Before this PR, grafting logic is spread across:
* `qmdb::current`
* `bitmap::authenticated`
* `mmr::grafting`

After this PR, all grafting logic is in `qmdb::current`.

This PR is not a format-breaking change. It doesn't change how `qmdb::current`'s root digest is calculated, but does change how state is internally represented.

This PR adds fields to `qmdb::current::Db`, replacing `status: bitmap::BitMap<E, H::Digest, N, S::BitMapState>` with 

```rs
    /// The bitmap over the activity status of each operation. Supports augmenting [Db] proofs in
    /// order to further prove whether a key _currently_ has a specific value.
    pub(super) status: BitMap<N>,

    /// Each leaf is hash(chunk || ops_subtree_root) for a complete bitmap chunk and
    /// the ops MMR node at the grafting height.
    /// Internal nodes are hashed using their position in the ops MMR rather than their
    /// grafted position.
    pub(super) grafted_mmr: mmr::mem::CleanMmr<H::Digest>,

    /// Persists:
    /// - The number of pruned bitmap chunks at key [PRUNED_CHUNKS_PREFIX]
    /// - The grafted MMR pinned nodes at key [NODE_PREFIX]
    pub(super) metadata: Metadata<E, U64, Vec<u8>>,

    /// Optional thread pool for parallelizing grafted leaf computation.
    pub(super) thread_pool: Option<ThreadPool>,
 ```
 
 These fields, and the logic that operates over them, are similar to those `bitmap::authenticated`. However, `qmdb::current` uses them in a meaningfully different way. Namely, unlike `bitmap::authenticated`, `qmdb::current` does not maintain an MMR representation of the the activity status bitmap. 
 
Also adds new crash recovery fuzz test for `qmdb::current` to add confidence that `commit()` durably persists data and that `qmdb::current` can gracefully recover from crashes.

### CODE REVIEW

File: storage/src/mmr/mod.rs [hunk 1]
```diff
@@ -79,7 +79,6 @@ pub(crate) mod conformance;
 
 cfg_if::cfg_if! {
     if #[cfg(feature = "std")] {
-        pub mod grafting;
```
danlaine, line 82: Moved to `current/grafting.rs`

roberto-bayardo:
Looking good, but looks like we lost the ability to merkleize the grafted-part of the tree in parallel?  Codex's suggestion for restoring:

Yes. It can be parallelized without changing root semantics.

What is currently serial.
Leaf recompute loop in [db.rs (line 724)](app://-/index.html#).
Ancestor propagation in [grafting.rs (line 91)](app://-/index.html#).
Peak fetch in compute_root is also serial now (for loop) at [db.rs (line 668)](app://-/index.html#).
What to change.
Recompute dirty grafted leaves in parallel (each chunk is independent) using thread pool tasks with per-task hasher instances (H: Default + Send + Sync).
Propagate ancestors level-by-level, but parallelize nodes within each level, then apply results before moving to next level (preserves dependencies).
Fetch root peaks with try_join_all (old code did this pattern), so peak reads are concurrent even without CPU pool.
Plumbing needed.
Current path does not expose a pool at this layer; Mmr keeps it private (pool in [journaled.rs (line 108)](app://-/index.html#)).
Easiest route is storing Option<ThreadPool> in current Db and threading it from current config ([mod.rs (line 191)](app://-/index.html#)).
Safety/determinism.
Result stays deterministic if you sort positions and merge writes deterministically.
No wire/storage format change; only compute strategy changes.

danlaine:
> Looking good, but looks like we lost the ability to merkleize the grafted-part of the tree in parallel? Codex's suggestion for restoring:
> 
> Yes. It can be parallelized without changing root semantics.
> 
> What is currently serial. Leaf recompute loop in db.rs (line 724). Ancestor propagation in grafting.rs (line 91). Peak fetch in compute_root is also serial now (for loop) at db.rs (line 668). What to change. Recompute dirty grafted leaves in parallel (each chunk is independent) using thread pool tasks with per-task hasher instances (H: Default + Send + Sync). Propagate ancestors level-by-level, but parallelize nodes within each level, then apply results before moving to next level (preserves dependencies). Fetch root peaks with try_join_all (old code did this pattern), so peak reads are concurrent even without CPU pool. Plumbing needed. Current path does not expose a pool at this layer; Mmr keeps it private (pool in journaled.rs (line 108)). Easiest route is storing Option in current Db and threading it from current config (mod.rs (line 191)). Safety/determinism. Result stays deterministic if you sort positions and merge writes deterministically. No wire/storage format change; only compute strategy changes.

Ah, yeah, you're right. I'll re-add parallelized hashing.

storage/src/qmdb/current/db.rs
```diff
@@ -85,11 +93,25 @@ pub struct Db<
...
-    pub(super) any: any::db::Db<E, C, I, H, U, S::AnyState, D>,
+    pub(super) any: any::db::Db<E, C, I, H, U, S::MmrState, D>,
+
+    /// The raw bitmap over the activity status of each operation.
+    pub(super) status: BitMap<N>,
+
+    /// Cache of grafted digests keyed by ops MMR positions. At the grafting height, entries are
+    /// `hash(chunk || ops_subtree_root)`. Above the grafting height, entries are standard MMR
+    /// internal nodes using ops-space positions in their hash pre-image.
+    pub(super) grafted_digests: BTreeMap<Position, H::Digest>,
```
roberto-bayardo: Can we use a dequeue here? (you'd index by the "peak tree position"). I'm concerned about the RAM blowup from the BTree. The LZ folks already flagged large memory usage from the previous Dequeue based impl.

storage/src/qmdb/current/db.rs
```diff
@@ -524,55 +666,215 @@ where
...
+
+    // Compute grafted leaves for each unpruned complete chunk and propagate upward. Pinned
+    // nodes (peaks of the pruned subtree) serve as siblings during propagation.
+    let complete_chunks = bitmap.complete_chunks();
+    recompute_grafted_leaves::<H, N>(
+        hasher,
+        bitmap,
+        &mut map,
+        ops_mmr,
+        pruned_chunks..complete_chunks,
```
roberto-bayardo: This is a huge range and recompute_grafted_leaves doesn't look particularly memory efficient (BTreeMaps, a couple of vectors of size corresponding to # of chunks).  Should we break this into pieces?

roberto-bayardo: Perhaps initialization can use its own optimized path. For example I wonder if a sequential scan of the ops journal is a better way to retrieve the grafted digests than to parallelize point lookups?

danlaine: Changed to use an in-memory MMR instead of BTreeMap to hold grafted nodes

storage/src/qmdb/current/db.rs
danlaine: Changed to use an in-memory MMR instead of BTreeMap to hold grafted nodes

storage/src/qmdb/current/db.rs
danlaine: Changed to use an in-memory MMR instead of BTreeMap to hold grafted nodes, which is more memory efficient. That said, we do store the grafted MMR in memory. Is that a concern for you?

storage/src/qmdb/current/db.rs
```diff
@@ -520,55 +692,221 @@ where
...
+            chunks
+                .map(|chunk_idx| {
+                    let chunk = *bitmap.get_chunk(chunk_idx);
+                    let ops_pos = grafting::chunk_idx_to_ops_pos(chunk_idx as u64, grafting_height);
+                    async move {
+                        let ops_digest = ops_mmr.get_node(ops_pos).await?.ok_or(
+                            mmr::Error::MissingGraftedLeaf(
+                                Location::try_from(ops_pos).unwrap_or_default(),
+                            ),
+                        )?;
```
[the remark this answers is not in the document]
danlaine: changed the value in this error to `Position` to remove this conversion

storage/src/qmdb/current/db.rs
```diff
@@ -343,10 +495,20 @@ where
...
         // Raise the inactivity floor by taking `self.steps` steps, plus 1 to account for the
         // previous commit becoming inactive.
-        let inactivity_floor_loc = self.any.raise_floor_with_bitmap(&mut self.status).await?;
+        let dirty_chunks = &mut self.state.dirty_chunks;
+        let inactivity_floor_loc = self
+            .any
+            .raise_floor_with_bitmap(&mut self.status, &mut |old_loc, new_loc| {
+                dirty_chunks.insert(BitMap::<N>::to_chunk_index(*old_loc));
+                dirty_chunks.insert(BitMap::<N>::to_chunk_index(*new_loc));
+            })
```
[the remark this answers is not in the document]
danlaine: Our options here are:

1. Remove `dirty_chunks.insert(BitMap::<N>::to_chunk_index(*new_loc));` and replace with a comment explaining that `new_loc` is guaranteed to be in a chunk >= `old_grafted_leaves`. It would feel a bit awkward to violate `dirty_chunks` invariant that it contains "Bitmap chunks modified since the last merkleization" by omitting in this case.
2. Change `dirty_chunks` to only contain indices of chunks that were complete at last merkleization. This requires us to move the `dirty_chunks` guard to prevent duplicate chunk indices from `into_merkleized` to `raise_floor_with_bitmap`.
3. Do nothing.

I opted for (2) but think (3) would also have been reasonable.

storage/fuzz/fuzz_targets/current_crash_recovery.rs
```diff
@@ -0,0 +1,318 @@
...
+use commonware_cryptography::{Hasher as _, Sha256};
+use commonware_runtime::{
+    buffer::paged::CacheRef,
+    deterministic::{self, Context},
+    Metrics as _, Runner,
+};
+use commonware_storage::{
+    mmr::Location,
+    qmdb::{
+        current::{unordered::fixed::Db as Current, FixedConfig},
```
roberto-bayardo: I think it would be better to use the Variable current as it's strictly more complex (and still works with fixed sized values, so you should only have to change initialization).

danlaine: Done

storage/src/qmdb/mod.rs
```diff
@@ -78,6 +78,9 @@ pub use verify::{
 /// Errors that can occur when interacting with an authenticated database.
 #[derive(Error, Debug)]
 pub enum Error {
+    #[error("data corrupted: {0}")]
+    DataCorrupted(&'static str),
```
patrick-ogrady, line 82: nit: `String`? (do we need a str/String at all)?

storage/src/qmdb/current/proof.rs
```diff
@@ -284,13 +258,7 @@ impl<D: Digest, const N: usize> OperationProof<D, N> {
...
-        self.range_proof.verify(
-            hasher,
-            grafting_height,
-            self.loc,
-            &[operation],
-            &[self.chunk],
-            root,
-        )
+        self.range_proof
+            .verify(hasher, self.loc, &[operation], &[self.chunk], root)
```
patrick-ogrady, line 262: IIRC there is a rust-y way to do single item [] now?

storage/src/qmdb/current/mod.rs
```diff
@@ -356,7 +506,9 @@ pub mod tests {
             log_journal_partition: format!("{partition_prefix}_partition_prefix"),
             log_items_per_blob: NZU64!(7),
             log_write_buffer: NZUsize!(1024),
-            bitmap_metadata_partition: format!("{partition_prefix}_bitmap_metadata_partition"),
+            grafted_mmr_metadata_partition: format!(
```
patrick-ogrady, line 509: I suppose we'll want to switch all these to being `-` separated partitions at some point (don't need to do here but probably a good issue)?

storage/src/qmdb/current/db.rs
```diff
@@ -520,55 +695,212 @@ where
...
+pub(super) async fn init_metadata<E: Storage + Clock + Metrics, D: Digest>(
+    context: E,
+    partition: &str,
+) -> Result<(Metadata<E, U64, Vec<u8>>, usize, Vec<D>), Error> {
+    let metadata_cfg = MConfig {
+        partition: partition.to_string(),
+        codec_config: ((0..).into(), ()),
+    };
+    let metadata =
+        Metadata::<_, U64, Vec<u8>>::init(context.with_label("metadata"), metadata_cfg).await?;
```
patrick-ogrady, line 872: I suppose we could use u64 as the value here but it would be harder to add new stuff.

danlaine: I was mostly just copying `mmr::journaled`, `bitmap::authenticated`, `journal::contiguous` -- I don't see any reason in particular to use `U64` instead of `u64` here

storage/src/qmdb/mod.rs
danlaine: What does `String` buy us? I could see an argument for declaring more specific DataCorruptedX values with the message in their `#[error(...)]` declaration

storage/src/qmdb/current/mod.rs
danlaine: Will make a follow up PR rn

storage/src/qmdb/current/db.rs
patrick-ogrady: Was referring to `Vec<u8>`, not `U64`.

danlaine: Ah 🤦 

The `Metadata` also has the pinned nodes, which aren't `u64`.

storage/src/qmdb/current/mod.rs
danlaine: https://github.com/commonwarexyz/monorepo/pull/3114

Issue #1110, opened 2025-06-16
[actors] Document and Refine Invariants


patrick-ogrady:
Recent work on `p2p` (https://github.com/commonwarexyz/monorepo/pull/1108 + https://github.com/commonwarexyz/monorepo/pull/1067) has uncovered a set of implicit (and important) invariants that our existing actors assume. We should take the time (for our own sanity) to properly document said invariants (and or refine them):

* Throughout the codebase, we call `unwrap()` in `Mailbox` (https://github.com/commonwarexyz/monorepo/blob/c46312f8af42edb7bbeefc1bd8c0216eaff871f3/p2p/src/authenticated/actors/tracker/ingress.rs#L111-L188) and (incorrectly) assume that anytime we can't send a message to another actor we should panic. During graceful shutdown, this can occur and cause an unexpected panic. We should either surface the disconnected error from mailbox/return a boolean indicating the drop/or just ignore the panic (may make it difficult to assess when it is time to shutdown).
* A recent change in `p2p` (https://github.com/commonwarexyz/monorepo/pull/921), made message ordering to `tracker` (from peers) critical for correctness (i.e. not panicking). We should form a strict opinion on whether other actors should ensure they always send messages that arrive at the destination in the right order (also implicitly enforcing the recipient to process incoming messages in a single loop) or make a best effort (dropping messages that don't make sense for the current actor state/returning an error).

I've historically found it "sane" to reason about these actors as independent entities that can handle invalid inputs behind well-defined interfaces (as a way to build large/complex crates without locks + better take advantage of parallelism). However, leaking the internal state machine of some actor to all other actors seems like the worst of all worlds (and very difficult to maintain). Getting on top of this seems prudent.

patrick-ogrady:
It may be worthwhile to "ban" all instances of actors waiting for responses (or to only permit in very specific cases). With "waiting", it is pretty easy to accidentally create a circular dependency of messages that eventually causes an actor to deadlock (they can't send more messages to a recipient actor and said actor can't send a message back to them).

danlaine:
Additional things to consider during this cleanup:

1. Should we adopt a `_tx`/`_rx` naming convention for senders and receivers?
2. Should we rename `Mailbox` to `Sender` to make it more clear that it wraps an `mpsc::Sender`?

patrick-ogrady:
For the different mailbox types, we could consider offering delivery tracking via: https://github.com/commonwarexyz/monorepo/pull/1396

[...]

PR #3077, opened 2026-02-08
[actor] Introduce Actor SDK

### LINKED ISSUE #2395: [runtime] Abstraction for pipelining read-only requests
## Overview

In commonware's actors, it's fairly common to serve requests in serial. This is usually fine for performance, but in cases such as database reads that may be costly, it's desirable to be able to serve multiple requests concurrently.

Take for example:

```rust
loop {
    select! {
        _ = read_request => {
            // <expensive database lookup>
        },
        // ... others ...
    }
}
```

If we sent 10 read requests, database reads are served one-by-one.

It may be worth looking for a general purpose abstraction to allow concurrently serving requests that do not mutate state for cases such as this.

clabby:
## Overview

Introduces an SDK for writing actor services with `commonware-runtime`.

related #1110


clabby:
For the core control loop service, we effectively want to multiplex over several sources, mapping to the same event message type.

These sources can be of several types:
- "Lane"s, which are known to be `mpsc::Receiver`s that contain the actor's ingress type already.
- "Source"s, which are opaque futures that are iteration-scoped. These futures resolve to a type `T`, and the user provides a function to map them to the actor's event message type.

The abstraction that we currently have is close, but it has some large problems:
- How do we express "iteration-scoped future builders"?
    - The `AsyncFn` and `AsyncFnMut` traits are stable as of rust `1.85.0`, but we currently [cannot name their return types (nightly only.)](https://doc.rust-lang.org/beta/std/ops/trait.AsyncFn.html#tymethod.async_call)
        - For context, these traits allow us to express `Fn<'a>() -> Fut<'a>`. See [`async_closures`](https://rust-lang.github.io/rfcs/3668-async-closures.html).
    - Writing generic code over the return type of an async future that borrows from the closure's inputs are not expressible in stable Rust. We _might_ be able to work around this with `Pin<Box<dyn Future<Output = T> + 'a>>`, but this method incurs cost.
- Even if we get the above to work, we run into another big problem: field projection.
    - If we had access to the nightly API and wrote future builder with GATs that looked like:
      ```rust
      /// A polled source armed for a single service-loop iteration.
      pub trait ArmedSource<I>: Send {
          /// Poll for one ingress message.
          ///
          /// - [`Poll::Ready`] with `Some(_)` yields an ingress event.
          /// - [`Poll::Ready`] with `None` indicates source exhaustion and stops the actor.
          /// - [`Poll::Pending`] indicates no event is ready.
          fn poll_next(&mut self, cx: &mut Context<'_>) -> Poll<Option<I>>;
      }

      /// A source that can be armed once per service-loop iteration.
      pub trait Source<E, A, Init, I>: Send + 'static {
          /// Armed state for one iteration.
          type Armed<'a>: ArmedSource<I> + 'a
          where
              Self: 'a,
              A: 'a,
              E: 'a,
              Init: 'a;

          /// Arm this source for one iteration.
          fn arm<'a>(
              &'a mut self,
              actor: &'a mut A,
              context: &'a E,
              init: &'a mut Init,
          ) -> Self::Armed<'a>;
      }
      ```
      we still have a big problem, because a single source captures a potential mutable borrow over all sources. This means that one source that used the `context` for a sleep future would prevent a future that borrows from the actor's fields, for example.
    - This means we need a way to perform field projection (like the Rust compiler would handle for us in a concrete context).

Exploring some ways around this.

---

An idea came to mind which is a bit less "nice", but simple to reason about and sidesteps the issue of dealing with this in a generic context (taken from the v1 but adapted to be more useful.) We can add a function to `Actor`:

```rs
fn auxiliary(
    &mut self,
    context: &mut E,
    init: &mut Self::Init
) -> impl Future<Output = Option<Self::Ingress>> + Send;
```

that function could then be implemented by end-users with a `select!` body that receives from external futures and maps them to the `Self::Ingress` type. We could then add that future to the service's select. Might need to do a bit of work to ensure polling order, but theoretically a lot more simple.

actor/src/lib.rs
```diff
@@ -0,0 +1,233 @@
...
+        /// exhaustion and stops the actor (calls [`Actor::on_shutdown`]).
+        ///
+        /// # Cancellation safety
+        ///
+        /// This future is dropped and recreated whenever a lane message wins
+        /// the race. Implementations must be **cancellation-safe**: any work
+        /// done before an internal `.await` point may be lost if the future
+        /// is cancelled at that point. Hold intermediate state in `self` or
+        /// `args` rather than in local variables across `.await` boundaries.
+        fn on_external(
```
clabby: The one downside to this, relative to the original approach I tried to take with `AsyncFn` registry in https://github.com/commonwarexyz/monorepo/pull/3077#issuecomment-3868905970 is that you cannot bias the "external" futures above the mailbox lanes.

In practice, this should not be a problem, though worth noting.

actor/src/lib.rs
```diff
@@ -0,0 +1,233 @@
...
+        ///
+        /// Returning `Err` is fatal: the service loop logs the error, drains
+        /// remaining in-flight reads, and then calls [`Actor::on_shutdown`]
+        /// before exiting.
+        fn on_read_only(
+            _context: E,
+            _snapshot: Self::Snapshot,
+            _message: <Self::Ingress as IntoIngressEnvelope>::ReadOnlyIngress,
+        ) -> impl Future<Output = Result<(), Self::Error>> + Send {
+            futures::future::pending()
```
andresilva: I wonder if `unreachable!()` is a better default. If the user doesn't declare `ask` items then this won't be reached, if it does and forgets to implement `on_read_only` then those requests will just hang forever. Probably more obvious to the user if this just panics immediately?

actor/macros/src/ingress/mod.rs
```diff
@@ -0,0 +1,803 @@
...
+    let readonly_phantom = if readonly_variants.is_empty() {
+        &phantom_variant
+    } else {
+        &empty
+    };
+    let read_write_phantom = if read_write_variants.is_empty() {
+        &phantom_variant
+    } else {
+        &empty
+    };
```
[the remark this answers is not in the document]
clabby: This is intentional, we want to throw an error if the generics are unused.

actor/src/service/driver.rs
```diff
@@ -0,0 +1,396 @@
...
+            if reads.is_full() {
+                if self.wait_for_read_capacity(&mut args, &mut reads).await {
+                    return;
+                }
+
+                self.actor
+                    .postprocess(self.context.as_present_mut(), &mut args)
+                    .await;
+                continue;
+            }
```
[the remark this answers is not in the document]
clabby: We do process a message here, since we wait for a read to finish to clear capacity.

actor/src/service/driver.rs
```diff
@@ -0,0 +1,396 @@
...
+                .await;
+
+            if reads.is_full() {
+                if self.wait_for_read_capacity(&mut args, &mut reads).await {
+                    return;
+                }
+
+                self.actor
+                    .postprocess(self.context.as_present_mut(), &mut args)
+                    .await;
```
andresilva: This is technically in-line with the documentation but maybe surprising in the sense that people might expect `postprocess` to be run every time a new message is processed. Maybe should be called out explicitly in docs?

actor/src/service/driver.rs
```diff
@@ -0,0 +1,396 @@
...
+    type Output = Option<I>;
+
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        if self.lanes.is_empty() {
+            return Poll::Pending;
+        }
+        for lane in self.lanes.iter_mut() {
+            match lane.poll_recv(cx) {
+                Poll::Ready(Some(message)) => return Poll::Ready(Some(message)),
+                Poll::Ready(None) => return Poll::Ready(None),
```
andresilva: Should probably document that any lane closing shuts down the actor (which is the most reasonable thing to do IMO).

actor/src/lib.rs
```diff
@@ -0,0 +1,233 @@
...
+        /// done before an internal `.await` point may be lost if the future
+        /// is cancelled at that point. Hold intermediate state in `self` or
+        /// `args` rather than in local variables across `.await` boundaries.
+        fn on_external(
+            &mut self,
+            _context: &mut E,
+            _args: &mut Self::Args,
+        ) -> impl Future<Output = Option<<Self::Ingress as IntoIngressEnvelope>::ReadWriteIngress>> + Send {
+            futures::future::pending()
+        }
```
andresilva: I wonder if this could instead be a `Stream` that's created on startup. It will be slightly less flexible and probably require `Box::pin` but would avoid the cancellation safety requirement.

```rust
  fn make_external(
      &mut self,
      _context: &mut E,
      args: &mut Self::Args,
  ) -> Pin<Box<dyn Stream<Item = ExternalMailboxReadWriteMessage> + Send>> {
      let rx = std::mem::take(args);
      Box::pin(rx.map(|n| ExternalMailboxReadWriteMessage::TellMsg { n }))
  }
```

Do you think something like this could work?

actor/src/service/driver.rs
```diff
@@ -0,0 +1,396 @@
...
+                    .postprocess(self.context.as_present_mut(), &mut args)
+                    .await;
+                continue;
+            }
+
+            let event = {
+                let external = self
+                    .actor
+                    .on_external(self.context.as_present_mut(), &mut args);
+                Self::recv_event(&mut self.shutdown, &mut self.lanes, pin!(external)).await
```
andresilva: Should we also poll read completions here so fatal `on_read_only` errors are observed immediately without needing a mailbox "poke"?

actor/src/service/builder.rs
```diff
@@ -0,0 +1,300 @@
...
+        self.max_inflight_reads = max_inflight_reads;
+        self
+    }
+
+    /// Add a bounded lane, transitioning to a [`MultiLaneServiceBuilder`].
+    ///
+    /// `capacity` is per-lane queue depth and must be non-zero.
+    pub fn with_lane<L>(self, lane: L, capacity: NonZeroUsize) -> MultiLaneServiceBuilder<A, L>
+    where
+        L: Copy + Ord + Send + 'static,
```
andresilva: Is `Copy` really needed?

andresilva (commented):
I wonder if it would make sense to have another ingress type for synchronous read-only requests, maybe `peek`? E.g. in the counter example you could use an `AtomicU64` for the state, in which case you can service multiple synchronous read-requests (or anything behind a `RwLock` that isn't inherently asynchronous). I'm thinking just as general purpose functionality, maybe we don't have enough usage to justify this pattern.

actor/src/lib.rs
clabby: It could work, though the reason I opted to create this on each iteration is to allow for patterns where we replace the futures (for example, in the voter actor, where we need to continuously update timeouts and such.)

If we only had to deal with streams that lived as long as the actor (i.e. `p2p::Receiver`s), then I think this would be better.

actor/src/service/builder.rs
clabby: Good catch - no!

actor/src/lib.rs
clabby: I would prefer to do this, but unfortunately the compiler won't let us with the `impl Future` return type.

You get:

```
error[E0277]: `()` is not a future
   --> actor/src/lib.rs:156:14
    |
156 |         ) -> impl Future<Output = Result<(), Self::Error>> + Send {
    |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `()` is not a future
    |
    = help: the trait `futures::Future` is not implemented for `()`
```

actor/src/service/driver.rs
clabby: Good point!

actor/src/service/driver.rs
clabby: This is documented on the `Actor` trait, `on_shutdown`, and in the `ServiceBuilder`

actor/src/service/driver.rs
clabby: Good idea

actor/README.md
```diff
@@ -0,0 +1,323 @@
...
+
+- ingress enums used by the actor implementation (read-only vs read-write)
+- wrapper message types for `tell`/`ask`/`subscribe`
+- a typed mailbox with convenience methods for `pub` messages
+
+### Message kinds
+
+- `tell`: fire-and-forget message
+- `ask`: request/response message that does not mutate state
+- `ask read_write`: request/response message that may mutate state
```
patrick-ogrady: I wonder if there is a better word here for a "mutating" ask? Maybe `put`?

actor/README.md
```diff
@@ -0,0 +1,323 @@
...
+
+`ingress!` is the Actors' entrypoint. You describe what events an actor will handle, and the macro generates:
+
+- ingress enums used by the actor implementation (read-only vs read-write)
+- wrapper message types for `tell`/`ask`/`subscribe`
+- a typed mailbox with convenience methods for `pub` messages
+
+### Message kinds
+
+- `tell`: fire-and-forget message
```
patrick-ogrady: maybe `broadcast`

actor/README.md
```diff
@@ -0,0 +1,323 @@
...
+- `ask`: request/response message that does not mutate state
```
patrick-ogrady: maybe `get`

actor/README.md
```diff
@@ -0,0 +1,323 @@
...
+
+## Declare incoming messages with `ingress!`
+
+`ingress!` is the Actors' entrypoint. You describe what events an actor will handle, and the macro generates:
+
+- ingress enums used by the actor implementation (read-only vs read-write)
+- wrapper message types for `tell`/`ask`/`subscribe`
+- a typed mailbox with convenience methods for `pub` messages
+
+### Message kinds
```
patrick-ogrady: may feel more familiar if we try to do a CRUD thing

actor/README.md
andresilva: FWIW `ask` and `tell` are pretty common names around actor systems, at least I remember using these names on akka with the same exact semantics.

clabby: Good eye, these names are from akka: https://doc.akka.io/libraries/akka-core/current/actors.html :)

patrick-ogrady: lol 😅

actor/src/lib.rs
```diff
@@ -0,0 +1,237 @@
...
+    ///    then `postprocess`. If the source
+    ///    yields `None` (lane closed or `on_external` exhaustion), skip
+    ///    directly to `on_shutdown`.
+    /// 3. `on_shutdown` once, on graceful exit (runtime stop, lane closure,
+    ///    or `on_external` returning `None`).
+    ///
+    /// Returning `Err` from `on_read_only` or `on_read_write` is fatal: the error is logged,
+    /// remaining in-flight reads are drained, and then `on_shutdown` is called
+    /// before the loop exits.
+    pub trait Actor<E>: Send + 'static {
```
clabby, line 180: There are some cases where we may actually want to process more than 1 message from a lane, without forcing the user to use `on_external` with a custom setup. We should consider adding an abstraction that lets users attempt to drain multiple messages from a lane channel.

.github/workflows/publish.yml
```diff
@@ -25,6 +25,11 @@ jobs:
       continue-on-error: true
       env:
         CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
+    - name: Publish actor-macros
```
patrick-ogrady: Is there a reason not to publish just before `actor`? I suppose we want to allow other crates to use as much as possible?

README.md
```diff
@@ -12,6 +12,7 @@
 
 _Primitives are designed for deployment in adversarial environments. If you find an exploit, please refer to our [security policy](./SECURITY.md) before disclosing it publicly (an exploit may equip a malicious party to attack users of a primitive)._
 
+* [actor](./actor/README.md): Coordinate actors with protocol-driven ingress and lane-aware control loops.
```
patrick-ogrady: Let's make this shorter? I think something like `Coordinate actors with priority ingest.` (or something like it) is the right form?

actor/README.md
```diff
@@ -0,0 +1,327 @@
+# commonware-actor
+
+[![Crates.io](https://img.shields.io/crates/v/commonware-actor.svg)](https://crates.io/crates/commonware-actor)
+[![Docs.rs](https://docs.rs/commonware-actor/badge.svg)](https://docs.rs/commonware-actor)
+
+Coordinate actors with protocol-driven ingress and lane-aware control loops.
+
+## Declare incoming messages with `ingress!`
```
patrick-ogrady: I tend to put all this in the `mod` comment instead?

actor/README.md
```diff
@@ -0,0 +1,327 @@
...
+    CounterMailbox,
+
+    pub tell Increment { amount: u64 };
+    pub ask Get -> u64;
+    pub ask read_write AddAndGet { amount: u64 } -> u64;
+    pub subscribe WaitForNext -> u64;
+}
+```
+
+The declaration above generates a mailbox API with the shape:
```
patrick-ogrady: woah

actor/README.md
```diff
@@ -0,0 +1,327 @@
...
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+enum Lane {
+    Control,
+    Data,
+}
+
+# let context = unimplemented!();
+# let actor = unimplemented!();
+let (_lanes, _service) = ServiceBuilder::new(actor)
+    .with_lane(Lane::Control, NonZeroUsize::new(32).expect("non-zero"))
```
patrick-ogrady: use `NZUsize` in example?

actor/macros/README.md
```diff
@@ -0,0 +1,9 @@
+# commonware-actor-macros
+
+[![Crates.io](https://img.shields.io/crates/v/commonware-actor-macros.svg)](https://crates.io/crates/commonware-actor-macros)
+[![Docs.rs](https://docs.rs/commonware-actor-macros/badge.svg)](https://docs.rs/commonware-actor-macros)
+
+Procedural macros for `commonware-actor`.
+
+This crate currently provides the `ingress!` macro used by
```
patrick-ogrady: stability header?

actor/macros/src/ingress/mod.rs
```diff
@@ -0,0 +1,803 @@
...
+use quote::{format_ident, quote};
+use std::collections::BTreeSet;
+use syn::{parse_macro_input, visit::Visit, GenericParam, Generics, Ident, Result, Type};
+
+mod parsing;
+use parsing::{Field, IngressInput, Item, ItemKind, MailboxKind};
+
+#[derive(Default)]
+struct GenericUsage {
+    type_params: BTreeSet<String>,
```
patrick-ogrady: at a small size, would `Vec<String>` be faster?

patrick-ogrady: I suppose this is only used in setup?

actor/src/lib.rs
```diff
@@ -0,0 +1,265 @@
+#![doc = include_str!("../README.md")]
```
patrick-ogrady: Ah, disregard my earlier README comment.

actor/src/lib.rs
```diff
@@ -0,0 +1,265 @@
...
+        fn on_read_write(
+            &mut self,
+            _context: &mut E,
+            _args: &mut Self::Args,
+            _message: <Self::Ingress as IntoIngressEnvelope>::ReadWriteIngress,
+        ) -> impl Future<Output = Result<(), Self::Error>> + Send {
+            futures::future::pending()
+        }
+
+        /// Poll external per-iteration sources and map them to ingress.
```
patrick-ogrady: dumb question: where would we want to use this?

patrick-ogrady (commented):
First pass + LLM feedback

```
  What I changed

  - actor/src/mailbox.rs:106
    Switched bounded Mailbox::tell_lossy from blocking send(...).await to non-blocking try_send(...), so full channels now return false immediately.
  - actor/src/mailbox.rs:297
    Added regression test bounded_tell_lossy_does_not_block_when_full.
  - actor/macros/src/ingress/mod.rs:135 and actor/macros/src/ingress/mod.rs:791
    Changed phantom generation logic to be based on per-enum generic usage, not just enum emptiness.
    This now covers valid cases where generics are unused in one generated enum branch, and also references const generics correctly ([(); N]).
  - actor/src/service/driver.rs:206
    Added dispatched_ingress tracking so postprocess only runs after actual ingress dispatch.
    Removed postprocess call from the read-capacity wait path (no ingress dispatched there).
  - actor/src/service/tests/mod.rs:868
    Added test_generic_mailbox_with_readonly_generic_response regression case (the macro case that previously produced E0392/E0283).
  - actor/src/service/tests/mod.rs:987
    Added test_postprocess_runs_only_after_dispatch regression case.

  Verification

  - Before fixes:
      - Macro regression reproduced with E0392/E0283 in GenericReadOnlyResponseMailbox....
      - bounded_tell_lossy_does_not_block_when_full failed (future was pending on full channel).
      - test_postprocess_runs_only_after_dispatch failed (left: 2, right: 1).
 ```
 
 Patch suggested:
 
 ```
 diff --git a/actor/macros/src/ingress/mod.rs b/actor/macros/src/ingress/mod.rs
index 45d7786c5..65dab22ef 100644
--- a/actor/macros/src/ingress/mod.rs
+++ b/actor/macros/src/ingress/mod.rs
@@ -132,6 +132,85 @@ fn collect_usage_from_fields(fields: &[Field], names: &GenericParamNames) -> Gen
     usage
 }
 
+fn collect_readonly_ingress_usage(items: &[Item], names: &GenericParamNames) -> GenericUsage {
+    let mut usage = GenericUsage::default();
+    for item in items {
+        if let ItemKind::Ask {
+            response,
+            read_write: false,
+        } = &item.kind
+        {
+            usage.merge(collect_usage_from_fields(&item.fields, names));
+            usage.merge(collect_usage_from_type(response, names));
+        }
+    }
+    usage
+}
+
+fn collect_read_write_ingress_usage(items: &[Item], names: &GenericParamNames) -> GenericUsage {
+    let mut usage = GenericUsage::default();
+    for item in items {
+        match &item.kind {
+            ItemKind::Tell => {
+                usage.merge(collect_usage_from_fields(&item.fields, names));
+            }
+            ItemKind::Ask {
+                response,
+                read_write: true,
+            }
+            | ItemKind::Subscribe { response } => {
+                usage.merge(collect_usage_from_fields(&item.fields, names));
+                usage.merge(collect_usage_from_type(response, names));
+            }
+            ItemKind::Ask {
+                read_write: false, ..
+            } => {}
+        }
+    }
+    usage
+}
+
+fn has_unused_generics(generics: &Generics, usage: &GenericUsage) -> bool {
+    generics.params.iter().any(|param| match param {
+        GenericParam::Type(param) => !usage.type_params.contains(&param.ident.to_string()),
+        GenericParam::Lifetime(param) => {
+            !usage
+                .lifetime_params
+                .contains(&param.lifetime.ident.to_string())
+        }
+        GenericParam::Const(param) => !usage.const_params.contains(&param.ident.to_string()),
+    })
+}
+
+fn phantom_variant_for_unused_generics(generics: &Generics, usage: &GenericUsage) -> TokenStream2 {
+    if generics.params.is_empty() || !has_unused_generics(generics, usage) {
+        return quote!();
+    }
+
+    let phantom_args: Vec<_> = generics
+        .params
+        .iter()
+        .map(|p| match p {
+            GenericParam::Type(tp) => {
+                let ident = &tp.ident;
+                quote!(#ident)
+            }
+            GenericParam::Lifetime(lp) => {
+                let lt = &lp.lifetime;
+                quote!(&#lt ())
+            }
+            GenericParam::Const(param) => {
+                let ident = &param.ident;
+                quote!([(); #ident])
+            }
+        })
+        .collect();
+    quote! {
+        #[doc(hidden)]
+        _Phantom(::core::marker::PhantomData<(#(#phantom_args),*)>),
+    }
+}
+
 fn wrapper_generics_tokens(
     generics: &Generics,
     usage: &GenericUsage,
@@ -709,41 +788,10 @@ pub(crate) fn expand(input: TokenStream) -> TokenStream {
         MailboxKind::Unbounded => quote!(#actor::mailbox::UnboundedMailbox<#ingress #ty_generics>),
     };
 
-    let phantom_variant = if generics.params.is_empty() {
-        quote!()
-    } else {
-        let phantom_args: Vec<_> = generics
-            .params
-            .iter()
-            .map(|p| match p {
-                GenericParam::Type(tp) => {
-                    let ident = &tp.ident;
-                    quote!(#ident)
-                }
-                GenericParam::Lifetime(lp) => {
-                    let lt = &lp.lifetime;
-                    quote!(&#lt ())
-                }
-                GenericParam::Const(_) => quote!(()),
-            })
-            .collect();
-        quote! {
-            #[doc(hidden)]
-            _Phantom(::core::marker::PhantomData<(#(#phantom_args),*)>),
-        }
-    };
-
-    let empty = quote!();
-    let readonly_phantom = if readonly_variants.is_empty() {
-        &phantom_variant
-    } else {
-        &empty
-    };
-    let read_write_phantom = if read_write_variants.is_empty() {
-        &phantom_variant
-    } else {
-        &empty
-    };
+    let readonly_usage = collect_readonly_ingress_usage(&items, &names);
+    let read_write_usage = collect_read_write_ingress_usage(&items, &names);
+    let readonly_phantom = phantom_variant_for_unused_generics(&generics, &readonly_usage);
+    let read_write_phantom = phantom_variant_for_unused_generics(&generics, &read_write_usage);
 
     quote! {
         pub enum #readonly_ingress #generics #where_clause {
diff --git a/actor/src/mailbox.rs b/actor/src/mailbox.rs
index c7d32630b..4e4643fda 100644
--- a/actor/src/mailbox.rs
+++ b/actor/src/mailbox.rs
@@ -101,12 +101,13 @@ impl<I> Mailbox<I> {
 
     /// Send a fire-and-forget message, ignoring closed-mailbox errors.
     ///
-    /// Returns `true` when ingress was delivered and `false` if the mailbox was closed.
+    /// Returns `true` when ingress was delivered and `false` if the mailbox was
+    /// closed or at capacity.
     pub async fn tell_lossy<T>(&self, msg: T) -> bool
     where
         T: Tell<I>,
     {
-        self.tx.send(msg.into_ingress()).await.is_ok()
+        self.tx.try_send(msg.into_ingress()).is_ok()
     }
 
     /// Send a request and wait for a response.
@@ -244,6 +245,7 @@ impl<I> From<mpsc::UnboundedSender<I>> for UnboundedMailbox<I> {
 mod tests {
     use super::*;
     use commonware_runtime::{deterministic, Runner};
+    use std::task::Poll;
 
     enum TestIngress {
         Tell,
@@ -291,6 +293,27 @@ mod tests {
         });
     }
 
+    #[test]
+    fn bounded_tell_lossy_does_not_block_when_full() {
+        let runner = deterministic::Runner::default();
+        runner.start(|_| async move {
+            let (tx, mut rx) = mpsc::channel::<TestIngress>(1);
+            let mailbox = Mailbox::new(tx);
+
+            assert!(mailbox.tell_lossy(TellMsg).await);
+
+            let second = mailbox.tell_lossy(TellMsg);
+            futures::pin_mut!(second);
+            assert!(matches!(futures::poll!(second.as_mut()), Poll::Ready(false)));
+
+            assert!(matches!(rx.recv().await, Some(TestIngress::Tell)));
+            assert!(matches!(
+                rx.try_recv(),
+                Err(mpsc::error::TryRecvError::Empty)
+            ));
+        });
+    }
+
     #[test]
     fn unbounded_tell_returns_closed_when_receiver_dropped() {
         let (tx, rx) = mpsc::unbounded_channel::<TestIngress>();
diff --git a/actor/src/service/driver.rs b/actor/src/service/driver.rs
index f7bcd2778..bccca816f 100644
--- a/actor/src/service/driver.rs
+++ b/actor/src/service/driver.rs
@@ -203,14 +203,12 @@ where
                 .preprocess(self.context.as_present_mut(), &mut args)
                 .await;
 
+            let mut dispatched_ingress = false;
+
             if reads.is_full() {
                 if self.wait_for_read_capacity(&mut args, &mut reads).await {
                     return;
                 }
-
-                self.actor
-                    .postprocess(self.context.as_present_mut(), &mut args)
-                    .await;
                 continue;
             }
 
@@ -239,6 +237,7 @@ where
                         return;
                     }
                     LoopEvent::Mailbox(lane, Some(message)) => {
+                        dispatched_ingress = true;
                         if self.dispatch_ingress(&mut args, &mut reads, message).await {
                             return;
                         }
@@ -247,6 +246,7 @@ where
                         }
                     }
                     LoopEvent::External(Some(message)) => {
+                        dispatched_ingress = true;
                         if self.handle_read_write(&mut args, &mut reads, message).await {
                             return;
                         }
@@ -263,9 +263,11 @@ where
                 }
             }
 
-            self.actor
-                .postprocess(self.context.as_present_mut(), &mut args)
-                .await;
+            if dispatched_ingress {
+                self.actor
+                    .postprocess(self.context.as_present_mut(), &mut args)
+                    .await;
+            }
         }
     }
 
diff --git a/actor/src/service/tests/mod.rs b/actor/src/service/tests/mod.rs
index 28d8c1e58..3491101b3 100644
--- a/actor/src/service/tests/mod.rs
+++ b/actor/src/service/tests/mod.rs
@@ -865,6 +865,145 @@ fn test_generic_mailbox_actor() {
     });
 }
 
+struct GenericReadOnlyResponseActor<T: Clone + Send + 'static> {
+    value: T,
+}
+
+ingress! {
+    GenericReadOnlyResponseMailbox<T: Clone + Send + 'static>,
+
+    pub tell GenericReadOnlyPing;
+    pub ask GenericReadOnlyGetValue -> T;
+}
+
+impl<E: Spawner, T: Clone + Send + 'static> Actor<E> for GenericReadOnlyResponseActor<T> {
+    type Mailbox = GenericReadOnlyResponseMailbox<T>;
+    type Ingress = GenericReadOnlyResponseMailboxMessage<T>;
+    type Error = std::convert::Infallible;
+    type Args = ();
+    type Snapshot = T;
+
+    fn snapshot(&self, _args: &Self::Args) -> Self::Snapshot {
+        self.value.clone()
+    }
+
+    async fn on_read_only(
+        _context: E,
+        snapshot: Self::Snapshot,
+        message: GenericReadOnlyResponseMailboxReadOnlyMessage<T>,
+    ) -> Result<(), Self::Error> {
+        match message {
+            GenericReadOnlyResponseMailboxReadOnlyMessage::GenericReadOnlyGetValue { response } => {
+                response.send_lossy(snapshot);
+                Ok(())
+            }
+        }
+    }
+
+    async fn on_read_write(
+        &mut self,
+        _context: &mut E,
+        _args: &mut Self::Args,
+        message: GenericReadOnlyResponseMailboxReadWriteMessage<T>,
+    ) -> Result<(), Self::Error> {
+        match message {
+            GenericReadOnlyResponseMailboxReadWriteMessage::GenericReadOnlyPing => Ok(()),
+            GenericReadOnlyResponseMailboxReadWriteMessage::_Phantom(_) => unreachable!(),
+        }
+    }
+}
+
+#[test]
+fn test_generic_mailbox_with_readonly_generic_response() {
+    let runner = deterministic::Runner::default();
+    runner.start(|context| async move {
+        let actor = GenericReadOnlyResponseActor {
+            value: "value".to_string(),
+        };
+        let (mailbox, service) =
+            ServiceBuilder::new(actor).build(context.with_label("generic_readonly_response"));
+        service.start();
+
+        mailbox
+            .generic_read_only_ping()
+            .await
+            .expect("ping failed");
+        let value = mailbox
+            .generic_read_only_get_value()
+            .await
+            .expect("ask failed");
+        assert_eq!(value, "value".to_string());
+    });
+}
+
+struct PostprocessIngressOnlyActor {
+    postprocess_count: Arc<AtomicUsize>,
+}
+
+ingress! {
+    PostprocessIngressOnlyMailbox,
+
+    pub ask SlowRead -> ();
+}
+
+impl<E: Spawner + Clock> Actor<E> for PostprocessIngressOnlyActor {
+    type Mailbox = PostprocessIngressOnlyMailbox;
+    type Ingress = PostprocessIngressOnlyMailboxMessage;
+    type Error = std::convert::Infallible;
+    type Args = ();
+    type Snapshot = ();
+
+    fn snapshot(&self, _args: &Self::Args) -> Self::Snapshot {}
+
+    async fn on_read_only(
+        context: E,
+        _snapshot: Self::Snapshot,
+        message: PostprocessIngressOnlyMailboxReadOnlyMessage,
+    ) -> Result<(), Self::Error> {
+        match message {
+            PostprocessIngressOnlyMailboxReadOnlyMessage::SlowRead { response } => {
+                context.sleep(Duration::from_millis(10)).await;
+                response.send_lossy(());
+                Ok(())
+            }
+        }
+    }
+
+    async fn on_read_write(
+        &mut self,
+        _context: &mut E,
+        _args: &mut Self::Args,
+        message: PostprocessIngressOnlyMailboxReadWriteMessage,
+    ) -> Result<(), Self::Error> {
+        match message {}
+    }
+
+    async fn postprocess(&mut self, _context: &mut E, _args: &mut Self::Args) {
+        self.postprocess_count.fetch_add(1, Ordering::SeqCst);
+    }
+}
+
+#[test]
+fn test_postprocess_runs_only_after_dispatch() {
+    let runner = deterministic::Runner::default();
+    runner.start(|context| async move {
+        let postprocess_count = Arc::new(AtomicUsize::new(0));
+        let actor = PostprocessIngressOnlyActor {
+            postprocess_count: postprocess_count.clone(),
+        };
+        let (mailbox, service) =
+            ServiceBuilder::new(actor).build(context.with_label("postprocess_dispatch_only"));
+        let handle = service.start();
+
+        mailbox.slow_read().await.expect("ask failed");
+        context.sleep(Duration::from_millis(1)).await;
+        assert_eq!(postprocess_count.load(Ordering::SeqCst), 1);
+
+        drop(mailbox);
+        let _ = handle.await;
+    });
+}
+
 struct SubscribeActor {
     pending: Option<commonware_utils::channel::oneshot::Sender<String>>,
 }
```

actor/src/lib.rs
clabby: Everywhere! Actors have timeout futures, need to select over network ingress, etc. that aren't the mailbox lanes.

actor/macros/src/ingress/mod.rs
clabby: Yeah, this is only in the proc macro.

.github/workflows/publish.yml
clabby: Yeah - I added it earlier so that it's published after all of its dependencies.

actor/macros/README.md
clabby: The stability header is enforced in the crate that re-exports the macros (`actor`). The types it uses are all gated behind `ALPHA` as well.

(There's a section in the readme below this with the note that's in all other crates.)

actor/src/lib.rs
```diff
@@ -0,0 +1,267 @@
...
+        ///
+        /// Returning `Err` is fatal: the service loop logs the error, drains
+        /// remaining in-flight reads, and then calls [`Actor::on_shutdown`]
+        /// before exiting.
+        fn on_read_only(
+            _context: E,
+            _snapshot: Self::Snapshot,
+            _message: <Self::Ingress as IntoIngressEnvelope>::ReadOnlyIngress,
+        ) -> impl Future<Output = Result<(), Self::Error>> + Send {
+            futures::future::pending()
```
andresilva:
```suggestion
            async {
                panic!(
                    "Actor::on_read_only must be implemented when read-only ingress is dispatched"
                )
            }
```

actor/src/lib.rs
```diff
@@ -0,0 +1,267 @@
...
+        /// Returning `Err` is fatal: the service loop logs the error, drains
+        /// remaining in-flight reads, and then calls [`Actor::on_shutdown`]
+        /// before exiting.
+        fn on_read_write(
+            &mut self,
+            _context: &mut E,
+            _args: &mut Self::Args,
+            _message: <Self::Ingress as IntoIngressEnvelope>::ReadWriteIngress,
+        ) -> impl Future<Output = Result<(), Self::Error>> + Send {
+            futures::future::pending()
```
andresilva:
```suggestion
            async {
                panic!(
                    "Actor::on_read_write must be implemented when read-write ingress is dispatched"
                )
            }
```

actor/macros/src/lib.rs
```diff
@@ -0,0 +1,112 @@
...
+/// ```ignore
+/// use commonware_actor_macros::ingress;
+///
+/// ingress! {
+///     PeerMailbox<P: Clone + Send + 'static>,
+///     pub tell Connect { peer: P };
+///     pub ask IsConnected { peer: P } -> bool;
+/// }
+/// ```
+#[proc_macro]
```
andresilva, line 216: Might be worth adding an example of what this macro expands to?

andresilva (dismissed):
This looks great! I barely reviewed the macro code (pretty sure the LLMs can do a better job than me there anyway). Did some local passes with LLMs and didn't flag anything.

README.md
```diff
@@ -12,6 +12,7 @@
 
 _Primitives are designed for deployment in adversarial environments. If you find an exploit, please refer to our [security policy](./SECURITY.md) before disclosing it publicly (an exploit may equip a malicious party to attack users of a primitive)._
 
+* [actor](./actor/README.md): Coordinate actors with priority ingress.
```
patrick-ogrady: `Coordinate actors with messages of varying priority.`

actor/README.md
```diff
@@ -0,0 +1,328 @@
...
+
+## The `Actor` Trait
+
+An [`Actor`] defines lifecycle hooks and message handlers.
+
+At minimum, you provide:
+
+- `type Mailbox`: generated mailbox type from `ingress!`
+- `type Ingress`: generated envelope enum (`<MailboxName>Message`)
+- `type Error`: fatal handler error type
```
patrick-ogrady: To confirm, this is `fatal`?

actor/README.md
```diff
@@ -0,0 +1,328 @@
...
+    Control,
+    Data,
+}
+
+# let context = unimplemented!();
+# let actor = unimplemented!();
+let (_lanes, _service) = ServiceBuilder::new(actor)
+    .with_lane(Lane::Control, NZUsize!(32))
+    .with_lane(Lane::Data, NZUsize!(256))
+    .build(context)
```
patrick-ogrady: Can you call `build_with_capacity` here? What will it do?

actor/macros/src/ingress/mod.rs
```diff
@@ -0,0 +1,849 @@
...
+            where_clause,
+            generics: &generics,
+            names: &names,
+        };
+        emit_wrapper(item, &ctx)
+    });
+    let methods = items
+        .iter()
+        .filter(|item| item.expose_on_mailbox)
+        .map(|item| emit_mailbox_method(item, &actor, mailbox_kind, false));
```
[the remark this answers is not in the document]
patrick-ogrady: Not sure if this is still relevant?

actor/macros/src/ingress/mod.rs
patrick-ogrady, line 1: LLM feedback:

```
  - [P1] Preserve postprocess when batched lane closes — /Users/patrickogrady/code/monorepo-pr-3077/actor/src/service/driver.rs:391-398
    When max_lane_batch() > 1, this branch can run after at least one message in the winning lane has already been dispatched, and it returns
    immediately after shutdown_gracefully. That bypasses the outer-loop postprocess call for the final batch. In practice, if callers drop the
    last lane sender after enqueueing work, the actor processes that work but postprocess is skipped, which can drop required end-of-batch
    actions (for example, flush/commit logic done in postprocess).
  - [P2] Avoid timeout arg collision in generated ask timeout methods — /Users/patrickogrady/code/monorepo-pr-3077/actor/macros/src/ingress/
    mod.rs:664-668
    For non-unit ask ingress, the generated *_timeout method appends a parameter named timeout after reusing all user field names as parameters.
    If a message field is also named timeout, macro expansion generates duplicate parameter names and the field initializer resolves to the
    future argument, causing compilation errors (E0415/E0308). Using an internal parameter name (for example __timeout) avoids this collision.
  - [P2] Use non-conflicting temp names in subscribe method generation — /Users/patrickogrady/code/monorepo-pr-3077/actor/macros/src/ingress/
    mod.rs:693-695
    These generated subscribe helpers bind local variables as tx/rx and then use shorthand field initializers from user arguments. If a subscribe
    field is named tx or rx, the locals shadow the user parameters, so the generated message literal uses the wrong types and fails to compile.
    Internal temporary names should be prefixed/uniquified to avoid colliding with valid field names.
    
```

```
  - [P2] Preserve non-usize const generics in phantom variant — /Users/patrickogrady/code/monorepo-pr-3077/actor/macros/src/ingress/mod.rs:200-
    203
    The generated hidden _Phantom variant encodes every const generic as [(); #ident], which only type-checks when the const parameter is usize.
    An ingress! declaration like Mailbox<const FLAG: bool> currently fails to compile as soon as a phantom variant is needed (for example, when
    the const generic is unused in one ingress branch), so valid const-generic mailbox declarations are rejected.
  - [P2] Apply ask item attributes to timeout helper methods — /Users/patrickogrady/code/monorepo-pr-3077/actor/macros/src/ingress/mod.rs:646-648
    For pub ask items, attributes are attached to the main generated method but not to the generated *_timeout method. When an ask item is
    conditionally compiled (for example with #[cfg(...)]), the wrapper/variant/main method are removed but the timeout helper is still emitted
    and references missing symbols, causing compilation failures in the disabled configuration.
  - [P2] Honor cfg-gated fields in generated wrapper assignments — /Users/patrickogrady/code/monorepo-pr-3077/actor/macros/src/ingress/
    mod.rs:311-317
    Field attributes are preserved on generated enum/struct fields, but assignment and argument generation still unconditionally references every
    declared field. With field-level cfgs (for example #[cfg(any())] hidden: u64), the field is removed from the generated structs/variants while
    field_assignments still emits self.hidden, which makes macro expansion fail to compile.
    
```

actor/src/lib.rs
patrick-ogrady: I think we may want to put a bit more elbow grease in here. We need to be able to prioritize a timeout, for example, over messages?

actor/src/service/mod.rs
```diff
@@ -0,0 +1,19 @@
...
+mod builder;
+pub use builder::{MultiLaneServiceBuilder, MultiLaneUnboundedServiceBuilder, ServiceBuilder};
+
+mod driver;
+pub use driver::ActorService;
+
+mod types;
+pub use types::{DuplicateLaneError, Lanes};
+
+const DEFAULT_MAILBOX_CAPACITY: NonZeroUsize = NZUsize!(64);
```
patrick-ogrady: Elsewhere, I think we default to ~1024?

patrick-ogrady (commented):
Definitely getting there. My biggest concern rn is being able to prioritize timeouts over message processing.

actor/README.md
clabby: Yes, the error is returned when a fatal error occurs during message handling.

actor/README.md
clabby: You cannot. The `ServiceBuilder` uses a typestate pattern that exposes only the functions that make sense with the given configuration. `ServiceBuilder::build_with_capacity` is a short-hand function that's only exposed on single-lane actors.

actor/src/service/driver.rs
```diff
@@ -0,0 +1,505 @@
...
+    /// The returned handle resolves when the actor loop exits.
+    pub fn start_with(mut self, args: A::Args) -> Handle<()> {
+        let context = self.context.take();
+        context.spawn(move |context| async move {
+            self.context.restore(context);
+            self.enter(args).await
+        })
+    }
+
+    /// read-only or read-write handlers, and exits on shutdown, lane
```
clabby: Update this comment, looks to have been truncated.

actor/macros/src/ingress/input.rs
```diff
@@ -0,0 +1,228 @@
...
+fn parse_item(input: ParseStream<'_>) -> Result<Item> {
+    let attrs = input.call(Attribute::parse_outer)?;
+    let expose_on_mailbox = if input.peek(Token![pub]) {
+        input.parse::<Token![pub]>()?;
+        true
+    } else {
+        false
+    };
+
+    let kind = if input.peek(kw::tell) {
```
clabby, line 181: `tell` items don't validate against using `response` as a field name, while `ask` and `subscribe` both reject it. A `tell` with a field named `response` will compile and work, but it's inconsistent and could confuse someone reading the generated read-write enum variants (where `response` has special meaning for ask/subscribe).

Should likely reserve the `response` field:

```rust
ItemKind::Tell => {
    // ... after parse_fields ...
    if let Some(field) = fields.iter().find(|f| f.name == "response") {
        return Err(syn::Error::new(
            field.name.span(),
            "`response` is reserved for the implicit response channel in ask/subscribe items",
        ));
    }
}

File: actor/macros/src/ingress/parsing.rs [hunk 34]
```diff
@@ -0,0 +1,234 @@
...
+    };
+
+    let name: Ident = input.parse()?;
+    let fields = parse_fields(input)?;
+    if let Some(field) = fields.iter().find(|f| f.name == "response") {
+        return Err(syn::Error::new(
+            field.name.span(),
+            "`response` is reserved for the implicit response channel in ask/subscribe items",
+        ));
+    }
```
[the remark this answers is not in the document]
clabby: This is expected for consistency.

clabby:
Closing in favor of an updated PR using the new `mailbox` pattern.

PR #3079, opened 2026-02-08
[runtime] allow configuring runtime buffer pools

andresilva:


runtime/src/deterministic.rs
```diff
@@ -217,17 +215,38 @@ pub struct Config {
     /// Configuration for deterministic storage fault injection.
     /// Defaults to no faults being injected.
     storage_faults: FaultConfig,
```
patrick-ogrady: Nit: storage_fault_cfg?

andresilva: Updated, as well as the setter and getter.

PR #3080, opened 2026-02-08
[storage/journal/contiguous] Allow concurrent reads during sync

roberto-bayardo:
Change contiguous journals to not require &mut for sync (via interior mutability) in order to allow concurrent reads during sync.

Add a "reader()" API that provides stable reads even if there is a concurrent pruner.

### DIFF
--- a/storage/src/qmdb/any/value.rs
+++ b/storage/src/qmdb/any/value.rs
@@ -35,9 +35,9 @@ impl<V: VariableValue> sealed::ValueEncoding for VariableEncoding<V> {
 }
 
 /// A fixed-size, clonable value.
-pub trait FixedValue: CodecFixedShared + Clone + 'static {}
-impl<T: CodecFixedShared + Clone + 'static> FixedValue for T {}
+pub trait FixedValue: CodecFixedShared + Clone {}
+impl<T: CodecFixedShared + Clone> FixedValue for T {}
 
 /// A variable-size, clonable value.
-pub trait VariableValue: CodecShared + Clone + 'static {}
-impl<T: CodecShared + Clone + 'static> VariableValue for T {}
+pub trait VariableValue: CodecShared + Clone {}
+impl<T: CodecShared + Clone> VariableValue for T {}
--- a/storage/src/qmdb/store/batch.rs
+++ b/storage/src/qmdb/store/batch.rs
@@ -21,11 +21,11 @@ pub mod tests {
     use rand::Rng;
     use std::collections::BTreeSet;
 
-    pub trait TestKey: Array + Copy + Send + Sync + 'static {
+    pub trait TestKey: Array + Copy + Send + Sync {
         fn from_seed(seed: u64) -> Self;
     }
 
-    pub trait TestValue: Codec + Eq + PartialEq + Debug + Send + Sync + 'static {
+    pub trait TestValue: Codec + Eq + PartialEq + Debug + Send + Sync {
         fn from_seed(seed: u64) -> Self;
     }
 


storage/src/journal/contiguous/mod.rs
```diff
@@ -15,52 +15,63 @@ pub mod variable;
...
-    /// # Errors
+    /// While the returned guard exists, operations that need the journal's
+    /// internal write lock (such as `append`, `prune`, and `rewind`) may block
+    /// until the guard is dropped. This ensures any position within
+    /// `reader.bounds()` remains readable.
+    fn reader(&self) -> impl Future<Output = impl ContiguousReader<Item = Self::Item> + '_> + Send;
+
+    /// Returns [start, end) where `start` is the index of the oldest retained operation (affected
+    /// by pruning) and `end - 1` is the index of the newest retained operation.
+    fn bounds(&self) -> impl Future<Output = Range<u64>> + Send;
```
patrick-ogrady: Are there cases where we do want to yolo use bounds/size here?

patrick-ogrady: (i.e. why don't we make `Contiguous` the reader and remove this indirection?

roberto-bayardo: I've been going back on this, but there are a few places where we need to know the size without reading. I suppose we could leave size() as public but keep bounds() as part of the reader?

roberto-bayardo: ok, I removed bounds() from Congiguous API (leaving only size). Users would need to access bounds through the reader.

patrick-ogrady: So, I suppose the tradeoff here is that it is slightly less performant (because reader prevents write?)

roberto-bayardo: Reading used to prevent writes previously as well, however now the reader lock can be held across multiple reads, so that's perhaps the only "less performant" case.  I've added a TODO to allow for concurrent appends while the reader() is held. (I don't think any QMDB code would benefit from this optimization just yet since QMDB currently blocks any concurrent updates during reads.)

storage/src/qmdb/immutable/mod.rs
```diff
@@ -463,17 +467,20 @@ impl<
...
+    // equals the pruning boundary.
+    async fn inactivity_floor_loc(&self) -> Location {
+        self.journal.bounds().await.start
     }
 
     fn is_empty(&self) -> bool {
-        self.bounds().is_empty()
+        // An immutable store starts with an initial commit at location 0.
+        // It is empty if no Set operations have been committed (last commit is still the initial).
+        *self.last_commit_loc == 0
```
danlaine: The spec for `is_empty` is:

```rs
    /// Returns true if there are no active keys in the database.
    fn is_empty(&self) -> bool;
```

The old implementation satisfies this comment but this new implementation checks something else. Same for `qmdb::keyless`.

danlaine: Or we could remove `is_empty` as discussed [here](https://github.com/commonwarexyz/monorepo/pull/2865#discussion_r2737175692)

storage/src/journal/contiguous/variable.rs
```diff
@@ -855,18 +917,61 @@ impl<E: Clock + Storage + Metrics, V: CodecShared> MutableContiguous for Journal
 impl<E: Clock + Storage + Metrics, V: CodecShared> Persistable for Journal<E, V> {
     type Error = Error;
 
-    async fn commit(&mut self) -> Result<(), Error> {
```
danlaine, line 858: I think we want to keep this so we don't use the default implementation, which calls `sync()`, right?

storage/src/journal/contiguous/fixed.rs
```diff
@@ -520,131 +697,62 @@ impl<E: Clock + Storage + Metrics, A: CodecFixedShared> Journal<E, A> {
...
-    ///  - [Error::ItemOutOfRange] if the item at position `pos` does not exist.
-    pub async fn read(&self, pos: u64) -> Result<A, Error> {
-        let bounds = self.bounds();
-        if pos >= bounds.end {
-            return Err(Error::ItemOutOfRange(pos));
-        }
-        if pos < bounds.start {
-            return Err(Error::ItemPruned(pos));
+    /// Note that this value could be older than the `min_item_pos` last passed to prune.
+    pub async fn oldest_retained_pos(&self) -> Option<u64> {
```
danlaine: #2975 removed `oldest_retained_pos` functions -- should we remove this and have users call `pruning_boundary` instead?

storage/src/qmdb/immutable/mod.rs
roberto-bayardo: Depends how you think about it. The way I think about it is the database is never empty after the first update given its immutability property. Pruning the DB doesn't technically "empty" the database, it just results in the updates becoming forgetten by the snapshot (that is, you can still prove inclusion of pruned elements against the root).

But in this world, that means inactivity floor should probably always return 0 or 1, and I didn't change that. 

This of course isn't 100% correct since one could "commit" without updating anything and therefore is_empty would return false even though it technically is still empty, but not sure it's worth perfectly handling this?

For now I've changed inactivity floor to return 0 if `is_empty` and 1 otherwise, and also updated the documentation to explain the odd behavior.

roberto-bayardo: We could get rid of is_empty but the inactivity_floor behavior would remain problematic. I suppose we could get rid of inactivity floor from the trait too.

storage/src/journal/contiguous/variable.rs
roberto-bayardo: Yes indeed. I wish I understood why this was removed (_shakes fist at Claude_)

restored.

storage/src/journal/contiguous/fixed.rs
roberto-bayardo: Yes, especially since variable.rs already removes it. done.

roberto-bayardo:
> I think these changes are correct, modulo the minor issues I've commented on. That said, I think these changes add a significant amount of complexity. In particular, it takes a meaningful amount of cognitive effort to inspect and reason about the correctness of the locking code. To a greater or lesser extent this is always true for concurrent code. Are we confident that the performance gains this PR will bring -- namely when we are reading while syncing -- is worth the added complexity? If we're concerned about performance issues related to reading while syncing, could we perhaps mitigate that in a different layer above the journal level in the storage stack?

I think a stop-the-world fsync at the DB level is going to constantly be a performance thorn in our side, so personally I'm eager to resolve it, though in the least intrusive way possible.

storage/src/qmdb/immutable/mod.rs
roberto-bayardo: Went ahead and pulled the is_empty removal into this PR.  Also looked at removing inactivity_floor but it's used pretty extensively in testing, so leaving it. Have tried to document its "unusual" behavior in immutable+keyless with a ## Warning.

roberto-bayardo:
Just added a couple more commits where I removed all `read`/`bounds` capabilities that might bypass the reader. While this introduces more boilerplate it should help prevent misuse.

storage/src/qmdb/immutable/mod.rs
```diff
@@ -114,26 +117,27 @@ impl<
...
+        self.journal.bounds().await
     }
 
     /// Get the value of `key` in the db, or None if it has no value or its corresponding operation
     /// has been pruned.
     pub async fn get(&self, key: &K) -> Result<Option<V>, Error> {
-        let oldest = self.journal.bounds().start;
+        let oldest = self.journal.bounds().await.start;
         let iter = self.snapshot.get(key);
+        let reader = self.journal.reader().await;
```
danlaine:
```suggestion
        let reader = self.journal.reader().await;
        let oldest = reader.bounds().start;
        let iter = self.snapshot.get(key);
```
Avoids grabbing, dropping, re-grabbing the lock

storage/src/qmdb/keyless/mod.rs
```diff
@@ -95,13 +98,13 @@ impl<
...
-    /// Returns [Error::LocationOutOfBounds] if `loc` >= `self.journal.bounds().end`.
+    /// Returns [Error::LocationOutOfBounds] if `loc` >= `self.journal.bounds().await.end`.
     pub async fn get(&self, loc: Location) -> Result<Option<V>, Error> {
-        let op_count = self.journal.bounds().end;
+        let op_count = self.journal.bounds().await.end;
         if loc >= op_count {
             return Err(Error::LocationOutOfBounds(loc, op_count));
         }
-        let op = self.journal.read(loc).await?;
+        let op = self.journal.reader().await.read(*loc).await?;
```
danlaine:
```suggestion
        let reader = self.journal.reader().await;
        let op_count = reader.bounds().end;
        if loc >= op_count {
            return Err(Error::LocationOutOfBounds(
                loc,
                Location::new_unchecked(op_count),
            ));
        }
        let op = reader.read(*loc).await?;
```

Avoids grabbing, dropping, re-grabbing the lock

roberto-bayardo: done

storage/src/qmdb/immutable/mod.rs
roberto-bayardo: Fixed, plus went ahead and removed ability to get bounds() without using the reader, which would have caught this.

examples/sync/src/databases/any.rs
```diff
@@ -116,11 +116,11 @@ where
         self.root()
     }
 
-    fn size(&self) -> Location {
-        LogStore::bounds(self).end
+    async fn size(&self) -> Location {
```
patrick-ogrady, line 119: should `size` and `inactivity_floor` be returned in a single atomic op?

storage/fuzz/fuzz_targets/journal_crash_recovery.rs
```diff
@@ -160,8 +161,8 @@ trait FuzzJournal: Sized {
         cfg: Self::Config,
     ) -> impl std::future::Future<Output = Result<Self, commonware_storage::journal::Error>> + Send;
 
-    fn size(&self) -> u64;
-    fn bounds(&self) -> Range<u64>;
+    fn size(&self) -> impl std::future::Future<Output = u64> + Send;
```
patrick-ogrady: nit: `Future<>`

storage/src/qmdb/immutable/sync.rs
```diff
@@ -1094,8 +1113,8 @@ mod tests {
 
             // Verify the synced database has the expected state
             assert_eq!(synced_db.root(), root);
-            assert_eq!(synced_db.bounds().end, upper_bound);
-            assert_eq!(synced_db.bounds().start, lower_bound);
+            assert_eq!(synced_db.bounds().await.end, upper_bound);
```
patrick-ogrady: reuse fetched bounds?

storage/src/qmdb/store/db.rs
```diff
@@ -298,15 +300,23 @@ where
...
-            Location::new_unchecked(log.size().checked_sub(1).expect("commit should exist"));
-        let op = log.read(*last_commit_loc).await?;
-        let inactivity_floor_loc = op.has_floor().expect("last op should be a commit");
-
+        let last_commit_loc = Location::new_unchecked(
+            log.size()
+                .await
+                .checked_sub(1)
+                .expect("commit should exist"),
+        );
```
patrick-ogrady, line 308: retain `\n`

storage/src/qmdb/mod.rs
```diff
@@ -52,7 +52,7 @@
 
 use crate::{
     index::{Cursor, Unordered as Index},
-    journal::contiguous::{Contiguous, MutableContiguous},
+    journal::contiguous::{ContiguousReader, MutableContiguous},
```
patrick-ogrady: we'd normally do `journal::contiguous::{Reader, Mutable}`?

roberto-bayardo: Right now the traits are:  `Contiguous`, `MutableContiguous`, and `ContiguousReader`.   

The first two preceeded this PR, so was sticking with always having "Contiguous" in the name.  Would you prefer I rename both MutableContiguous & ContiguousReader to "Mutable" and "Reader" in this PR, or just ContiguousReader ro "Reader"? (Or something else...)

storage/src/qmdb/immutable/sync.rs
roberto-bayardo: done

storage/src/qmdb/store/db.rs
roberto-bayardo: restored

storage/fuzz/fuzz_targets/journal_crash_recovery.rs
roberto-bayardo: done

examples/sync/src/databases/any.rs
roberto-bayardo: Interesting question.  @danlaine and I had already discussed removing inactivity floor from the LogStore API, so once we do that, it might make more sense as part of a Reader API specific to DB's where inactivity floor makes sense?  We were going to clean that up in another PR though.

roberto-bayardo: Thinking a bit more about this, yes it does feel like the QMDB types that have the concept of an inactivity floor should expose that only through a (stable) reader.  I can address this in the subsequent PR that will add non-read-blocking sync to the various DB types?  Would rather keep the bulk of changes in this PR specific to contiguous journals.

patrick-ogrady: As discussed, we can punt the QMDB-y focused stuff to a follow-up.

storage/src/qmdb/immutable/sync.rs
patrick-ogrady: Another case where we could share bounds

storage/src/queue/storage.rs
```diff
@@ -197,11 +203,11 @@ impl<E: Clock + Storage + Metrics, V: CodecShared> Queue<E, V> {
             return Ok(None);
         }
 
-        let item = self.journal.read(self.read_pos).await?;
+        let item = reader.read(self.read_pos).await?;
         let pos = self.read_pos;
         self.read_pos += 1;
         self.metrics.next.set(self.read_pos as i64);
-        debug!(position = pos, "dequeued item");
+        debug!(pos, "dequeued item");
```
patrick-ogrady: use `position = `

storage/src/queue/storage.rs
```diff
@@ -120,13 +124,14 @@ impl<E: Clock + Storage + Metrics, V: CodecShared> Queue<E, V> {
 
         // On restart, ack_floor is the pruning boundary (items below are deleted).
         // acked_above is empty (in-memory state lost on restart).
-        let ack_floor = journal.bounds().start;
+        let bounds = journal.reader().await.bounds();
+        let ack_floor = bounds.start;
```
patrick-ogrady: just do inline

storage/src/journal/contiguous/fixed.rs
```diff
@@ -53,6 +53,7 @@
 //!
 //! The `replay` method supports fast reading of all unpruned items into memory.
 
+use super::ContiguousReader;
```
patrick-ogrady: nit: `Reader`?

storage/src/journal/contiguous/fixed.rs
```diff
@@ -185,13 +311,15 @@ impl<E: Clock + Storage + Metrics, A: CodecFixedShared> Journal<E, A> {
...
-        let mut inner = SegmentedJournal::init(context.with_label("blobs"), segmented_cfg).await?;
+        let mut journal =
+            SegmentedJournal::init(context.with_label("blobs"), segmented_cfg).await?;
 
         // Initialize metadata store
         let meta_cfg = MetadataConfig {
             partition: format!("{}-metadata", cfg.partition),
             codec_config: ((0..).into(), ()),
         };
+
```
patrick-ogrady, line 322: nit: remove `\n`

storage/src/journal/contiguous/fixed.rs
```diff
@@ -454,59 +589,99 @@ impl<E: Clock + Storage + Metrics, A: CodecFixedShared> Journal<E, A> {
...
+        // hold the outer write_lock since we don't want to start writing to the new section until
+        // this completes.
+        {
+            let inner = self.inner.read().await;
+            inner.journal.sync(section).await?;
+        }
+
+        // Ensure the new tail section exists, as required to maintain the invariant. This must
+        // happen after the previous section is synced.
+        {
```
patrick-ogrady: don't love this multi-lock flow but it isn't wrong

storage/src/journal/contiguous/variable.rs
```diff
@@ -130,37 +183,83 @@ impl<C> Config<C> {
...
+            .read(position, self.items_per_section, &self.offsets)
+            .await
+    }
+
+    async fn replay(
+        &self,
+        buffer_size: NonZeroUsize,
+        start_pos: u64,
+    ) -> Result<impl Stream<Item = Result<(u64, V), Error>> + Send, Error> {
+        let items_per_section = self.items_per_section;
```
patrick-ogrady: just use self.items_per_section?

patrick-ogrady (commented):
latest thoughts (will take ... you can disregard)

storage/src/qmdb/any/unordered/mod.rs
```diff
@@ -427,16 +449,15 @@ pub(super) mod test {
             let (clean_db, _) = db.commit(None).await.unwrap();
             // Distance should equal 3 after the second commit, with inactivity_floor
             // referencing the previous commit operation.
-            assert!(clean_db.bounds().end - clean_db.inactivity_floor_loc() <= 3);
+            assert!(clean_db.bounds().await.end - clean_db.inactivity_floor_loc().await <= 3);
```
patrick-ogrady, line 452: Suspect this will all get cleaned up in a follow-up PR 👀

storage/src/journal/contiguous/mod.rs
```diff
@@ -15,56 +15,64 @@ pub mod variable;
...
+    fn reader(&self) -> impl Future<Output = impl Reader<Item = Self::Item> + '_> + Send;
+
+    /// Return the total number of items that have been appended to the journal.
     ///
-    /// - Returns [Error::ItemPruned] if the item at `position` has been pruned.
-    /// - Returns [Error::ItemOutOfRange] if the item at `position` does not exist.
-    fn read(&self, position: u64) -> impl Future<Output = Result<Self::Item, Error>> + Send;
+    /// This count is NOT affected by pruning. The next appended item will receive this
+    /// position as its value.
+    fn size(&self) -> impl Future<Output = u64> + Send;
```
patrick-ogrady, line 71: Is this the same as `reader.bounds().end`? If so, is it worth keeping here?

roberto-bayardo: Yes it is.  It's called in a lot of places, so it's a handy shortcut?

storage/src/queue/storage.rs
```diff
@@ -124,14 +124,14 @@ impl<E: Clock + Storage + Metrics, V: CodecShared> Queue<E, V> {
 
         // On restart, ack_floor is the pruning boundary (items below are deleted).
         // acked_above is empty (in-memory state lost on restart).
-        let bounds = journal.reader().await.bounds();
-        let ack_floor = bounds.start;
+        let ack_floor = journal.reader().await.bounds().start;
+        let size = journal.size().await;
```
roberto-bayardo: size can just be bounds.end

storage/src/qmdb/immutable/mod.rs
```diff
@@ -463,17 +488,25 @@ impl<
...
+    /// of the first update. This implementation assumes that if an update exists, it's at location
+    /// 1, even though technically there may be multiple commits without any updates.
+    ///
+    /// # Warning
+    ///
+    /// Note that the immutable store allows active operations to be pruned, so the inactivity floor
+    /// may precede the pruning boundary. If you wish to know the range of currently retrievable
+    /// operations, use `bounds`.
+    async fn inactivity_floor_loc(&self) -> Location {
+        if *self.last_commit_loc == 0 {
```
patrick-ogrady: Should we just leave this unmodified?

patrick-ogrady: If we want to keep, I would've expected a test or two here?

storage/src/qmdb/keyless/mod.rs
```diff
@@ -354,18 +363,26 @@ impl<
...
+    /// All updates are treated as "active" in a keyless store, so the inactivity floor is always the
+    /// location of the first update. This implementation assumes that if an update exists, it's at
+    /// location 1, even though technically there may be multiple commits without any updates.
+    ///
+    /// # Warning
+    ///
+    /// Note that the keyless store allows active operations to be pruned, so the inactivity floor
+    /// may precede the pruning boundary. If you wish to know the range of currently retrievable
+    /// operations, use `bounds`.
+    async fn inactivity_floor_loc(&self) -> Location {
```
patrick-ogrady, line 371: See earlier comment about keeping as a bounds read?

storage/src/queue/storage.rs
```diff
@@ -154,8 +158,8 @@ impl<E: Clock + Storage + Metrics, V: CodecShared> Queue<E, V> {
     /// Returns an error if the underlying storage operation fails.
     pub async fn append(&mut self, item: V) -> Result<u64, Error> {
         let pos = self.journal.append(item).await?;
-        self.metrics.tip.set(self.journal.size() as i64);
-        debug!(position = pos, "appended item");
```
patrick-ogrady, line 158: retain position=

storage/src/qmdb/any/value.rs
patrick-ogrady, line 1: Is adding `'static` to these a good idea?

roberto-bayardo: it doesn't hurt, and it makes the older compiler happy.

roberto-bayardo: (This is why I asked yesterday whether we could bump the min version...)

storage/src/qmdb/immutable/mod.rs
roberto-bayardo: Yeah this is the change the inspired our discussion around removing it entirely.  I've reverted it for now but it's going to go away entirely.

storage/src/qmdb/keyless/mod.rs
roberto-bayardo: reverted

PR #3081, opened 2026-02-09
Improve code coverage of the simplex fuzz tests

dnkolegov-ar:
This PR improves code coverage of the simplex fuzz tests by the following mechanisms:
- Adding custom election types and adding a random elector for some fuzz configurations
- It also adds an adversarial network wrapper that prioritizes messages from the byzantine nodes and filters messages from the correct replicas.
- It also instantiates only one node on the test and loads it with byzantine messages. The main goal is to be able to use fuzzing against the target node with high exec/s
- adding fuzz events for `journal`: after ending a fuzz session we take the checkpoint of the honest node and reinitiate it using the checkpoint


consensus/fuzz/src/lib.rs
```diff
@@ -158,9 +163,9 @@ impl Arbitrary<'_> for FuzzInput {
         // AnyScope mutations - 10%,
         // FutureScope mutations with round-based injections - 10%
         let fault_rounds_bound = u.int_in_range(1..=required_containers)?;
-        let max_faults = fault_rounds_bound / FAULT_INJECTION_RATIO;
-        let min_faults = MIN_NUMBER_OF_FAULTS.min(fault_rounds_bound);
-        let fault_rounds = u.int_in_range(0..=max_faults)?.max(min_faults);
+        let max_fault_rounds = fault_rounds_bound / FAULT_INJECTION_RATIO;
+        let min_fault_rounds = MIN_NUMBER_OF_FAULTS.min(fault_rounds_bound);
+        let fault_rounds = u.int_in_range(0..=max_fault_rounds)?.max(min_fault_rounds);
```
patrick-ogrady: why not just min..=max?

consensus/fuzz/src/lib.rs
```diff
@@ -384,11 +430,68 @@ fn spawn_honest_validator<P: simplex::Simplex>(
...
+    context: deterministic::Context,
+    oracle: &Oracle<Ed25519PublicKey, deterministic::Context>,
+    participants: &[Ed25519PublicKey],
+    scheme: P::Scheme,
+    validator: Ed25519PublicKey,
+    byzantine_router: crate::network::Router<Ed25519PublicKey, deterministic::Context>,
+    relay: Arc<relay::Relay<Sha256Digest, Ed25519PublicKey>>,
+    channels: NetworkChannels,
+) -> reporter::Reporter<deterministic::Context, P::Scheme, P::Elector, Sha256Digest> {
+    use crate::network::ByzantineFirstReceiver;
```
patrick-ogrady: add import to top of file

consensus/fuzz/src/lib.rs
```diff
@@ -668,32 +844,43 @@ fn run_with_twin_mutator<P: simplex::Simplex>(input: FuzzInput) {
     });
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Mode {
```
patrick-ogrady, line 887: dumb question: is there a reason we don't select this during fuzz input generation? I'm assuming it is so we can just run one strategy if we want targeted fuzzing?

consensus/fuzz/src/lib.rs
```diff
@@ -177,13 +182,18 @@ impl Arbitrary<'_> for FuzzInput {
         let remaining = u.len().min(MAX_RAW_BYTES);
         let raw_bytes = u.bytes(remaining)?.to_vec();
 
+        // Only used by `Mode::AdversarialNetwork`.
+        let honest_messages_drop_ratio =
```
patrick-ogrady, line 193: I suppose this would go away if we generated mode here

consensus/fuzz/src/lib.rs
```diff
@@ -336,6 +345,41 @@ fn spawn_honest_validator<P: simplex::Simplex>(
...
+#[allow(clippy::too_many_arguments)]
+fn spawn_honest_validator_with_network<P: simplex::Simplex>(
+    context: deterministic::Context,
+    oracle: &Oracle<Ed25519PublicKey, deterministic::Context>,
+    participants: &[Ed25519PublicKey],
+    scheme: P::Scheme,
+    validator: Ed25519PublicKey,
+    relay: Arc<relay::Relay<Sha256Digest, Ed25519PublicKey>>,
+    vote_network: (
+        impl commonware_p2p::Sender<PublicKey = Ed25519PublicKey> + 'static,
```
patrick-ogrady: is `+ 'static` required?

dnkolegov-ar: no, fixed

consensus/fuzz/src/lib.rs
dnkolegov-ar: This is a good question. I thought it was a compromise between the randomly chosen fuzz configuration and the minimal fuzz mode we want to set manually. So, now, we manually set `Standard`, `Twinnable`, and `AdversarialNetwork`, and within one of these modes, instantiate fuzz configuration randomly.

consensus/fuzz/src/id_mock.rs
```diff
@@ -0,0 +1,414 @@
...
+/// - the signature's valid flag is true.
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub struct Signature {
+    pub signer: u32,
+    pub valid: bool,
+}
+
+impl Write for Signature {
+    fn write(&self, writer: &mut impl BufMut) {
+        self.signer.write(writer);
```
patrick-ogrady, line 190: This approach doesn't tolerate message manipulation over the wire. What I had in mind was something where we'd record somewhere what a scheme had signed, so that if manipulated we could simply check (rather than reading these off-the-wire?)

dnkolegov-ar: Updated

PR #3082, opened 2026-02-09
[runtime, storage] use buffer pool and `IoBuf*` in `utils::buffer`

andresilva:
Related #2883.

### DIFF
--- a/runtime/src/utils/buffer/paged/append.rs
+++ b/runtime/src/utils/buffer/paged/append.rs
@@ -242,9 +242,7 @@ impl<B: Blob> Append<B> {
             let len = buf_guard.data.len();
             if len > 0 {
                 let mut shrunk = IoBufMut::with_capacity(len);
-                // SAFETY: We initialize all bytes immediately below.
-                unsafe { shrunk.set_len(len) };
-                shrunk.as_mut().copy_from_slice(buf_guard.data.as_ref());
+                shrunk.put_slice(buf_guard.data.as_ref());
                 buf_guard.data = shrunk;
             } else {
                 buf_guard.data = IoBufMut::default();


consensus/src/ordered_broadcast/mod.rs
```diff
@@ -1000,7 +1012,11 @@ mod tests {
                             sequencer.public_key()
                         ),
                         journal_compression: Some(3),
-                        journal_page_cache: CacheRef::new(PAGE_SIZE, PAGE_CACHE_SIZE),
+                        journal_page_cache: CacheRef::new(
+                            PAGE_SIZE,
+                            PAGE_CACHE_SIZE,
+                            context.storage_buffer_pool().clone(),
```
patrick-ogrady: let's put this first in `CacheRef`

storage/src/queue/conformance.rs
```diff
@@ -33,9 +35,11 @@ impl Conformance for QueueConformance {
     async fn commit(seed: u64) -> Vec<u8> {
         let runner = deterministic::Runner::seeded(seed);
         runner.start(|mut context| async move {
-            let mut queue = Queue::<_, Vec<u8>>::init(context.with_label("queue"), config(seed))
-                .await
-                .unwrap();
+            let pool = context.storage_buffer_pool().clone();
+            let mut queue =
+                Queue::<_, Vec<u8>>::init(context.with_label("queue"), config(seed, pool.clone()))
```
patrick-ogrady: Should we just create the pool from the provided context? Does it need to be in config?

storage/src/qmdb/immutable/mod.rs
```diff
@@ -530,17 +530,19 @@ pub(super) mod test {
...
         deterministic::{self},
-        Runner as _,
+        BufferPooler, Runner as _,
     };
     use commonware_utils::{NZUsize, NZU16, NZU64};
     use std::num::NonZeroU16;
 
     const PAGE_SIZE: NonZeroU16 = NZU16!(77);
     const PAGE_CACHE_SIZE: NonZeroUsize = NZUsize!(9);
+
```
patrick-ogrady: nit: remove \n?

storage/src/qmdb/benches/keyless/generate.rs
```diff
@@ -61,7 +64,7 @@ type KeylessMutable = Keyless<Context, Vec<u8>, Sha256, Unmerkleized, NonDurable
 /// committed after every `COMMIT_FREQUENCY` operations.
 async fn gen_random_keyless(ctx: Context, num_operations: u64) -> KeylessDb {
     let pool = ctx.clone().create_thread_pool(THREADS).unwrap();
-    let keyless_cfg = keyless_cfg(pool);
+    let keyless_cfg = keyless_cfg(pool, ctx.storage_buffer_pool().clone());
     let clean = KeylessDb::init(ctx, keyless_cfg).await.unwrap();
```
patrick-ogrady, line 65: Another case where we may be better off creating the storage buffer pool from ctx?

storage/src/qmdb/benches/fixed/mod.rs
```diff
@@ -233,7 +242,7 @@ async fn get_current_unordered_variable(ctx: Context) -> UVCurrentDb {
 /// Get an ordered variable current QMDB instance.
 async fn get_current_ordered_variable(ctx: Context) -> OVCurrentDb {
     let pool = ctx.clone().create_thread_pool(THREADS).unwrap();
-    let variable_current_cfg = variable_current_cfg(pool);
+    let variable_current_cfg = variable_current_cfg(pool, ctx.storage_buffer_pool().clone());
     OVCurrent::<_, _, _, Sha256, EightCap, CHUNK_SIZE>::init(ctx, variable_current_cfg)
```
patrick-ogrady, line 232: Another case where we should just create the storage buffer pool off ctx?

patrick-ogrady:
> Initial pass...I think we could use the provided `ctx` in many dbs to minimize diff.
> 
> LLMs also recommended the following:
> 
> ```
> diff --git a/consensus/src/aggregation/mod.rs b/consensus/src/aggregation/mod.rs
> index 020a3ab0e..f2fa39ab7 100644
> --- a/consensus/src/aggregation/mod.rs
> +++ b/consensus/src/aggregation/mod.rs
> @@ -1083,7 +1083,11 @@ mod tests {
>                          journal_replay_buffer: NZUsize!(4096),
>                          journal_heights_per_section: std::num::NonZeroU64::new(6).unwrap(),
>                          journal_compression: Some(3),
> -                        journal_page_cache: CacheRef::new(context.storage_buffer_pool().clone(), PAGE_SIZE, PAGE_CACHE_SIZE),
> +                        journal_page_cache: CacheRef::new(
> +                            context.storage_buffer_pool().clone(),
> +                            PAGE_SIZE,
> +                            PAGE_CACHE_SIZE,
> +                        ),
>                          strategy: Sequential,
>                      },
>                  );
> diff --git a/runtime/fuzz/fuzz_targets/blob_integrity.rs b/runtime/fuzz/fuzz_targets/blob_integrity.rs
> index cb5c59681..8ba3796be 100644
> --- a/runtime/fuzz/fuzz_targets/blob_integrity.rs
> +++ b/runtime/fuzz/fuzz_targets/blob_integrity.rs
> @@ -97,7 +97,8 @@ fn fuzz(input: FuzzInput) {
>          let cache_ref = CacheRef::new(
>              context.storage_buffer_pool().clone(),
>              NZU16!(page_size as u16),
> -            NZUsize!(cache_capacity));
> +            NZUsize!(cache_capacity),
> +        );
>  
>          // Compute logical size from number of pages.
>          let logical_size = input.num_pages as u64 * page_size;
> diff --git a/runtime/src/utils/buffer/read.rs b/runtime/src/utils/buffer/read.rs
> index 8dc20530d..30721d0ae 100644
> --- a/runtime/src/utils/buffer/read.rs
> +++ b/runtime/src/utils/buffer/read.rs
> @@ -114,7 +114,7 @@ impl<B: Blob> Read<B> {
>              .blob
>              .read_at_buf(self.blob_position, bytes_to_read, buf)
>              .await?;
> -        self.buffer = read_result.coalesce();
> +        self.buffer = read_result.coalesce_with_pool(&self.pool);
>          self.buffer_valid_len = bytes_to_read;
>  
>          Ok(bytes_to_read)
> diff --git a/runtime/src/utils/buffer/write.rs b/runtime/src/utils/buffer/write.rs
> index 658d06688..f29e44f59 100644
> --- a/runtime/src/utils/buffer/write.rs
> +++ b/runtime/src/utils/buffer/write.rs
> @@ -74,8 +74,7 @@ impl<B: Blob> Write<B> {
>  
>  impl<B: Blob> Blob for Write<B> {
>      async fn read_at(&self, offset: u64, len: usize) -> Result<IoBufsMut, Error> {
> -        self.read_at_buf(offset, len, IoBufMut::with_capacity(len))
> -            .await
> +        self.read_at_buf(offset, len, self.pool.alloc(len)).await
>      }
>  
>      async fn read_at_buf(
> ```

I suppose the justification for not returning a pool buffer here is that an object that holds bytes will retain this indefinitely. On the other hand, if we are decoding into a struct (almost always) then this is a big win.

runtime/fuzz/fuzz_targets/buffer.rs
```diff
@@ -130,7 +130,12 @@ fn fuzz(input: FuzzInput) {
                         }
                     }
 
-                    read_buffer = Some(Read::new(blob, blob_size.min(size), NZUsize!(buffer_size)));
+                    read_buffer = Some(Read::new(
+                        blob,
+                        blob_size.min(size),
+                        NZUsize!(buffer_size),
+                        context.storage_buffer_pool().clone(),
```
patrick-ogrady: ::from_pooler

storage/fuzz/fuzz_targets/mmr_journaled_crash_recovery.rs
```diff
@@ -294,6 +295,7 @@ fn fuzz(input: FuzzInput) {
                     page_cache_size,
                     items_per_blob,
                     write_buffer,
+                    &ctx,
```
patrick-ogrady: should be first? should be created outside this function?

storage/fuzz/fuzz_targets/qmdb_any_fixed_sync.rs
```diff
@@ -150,10 +152,13 @@ fn fuzz(mut input: FuzzInput) {
     let runner = deterministic::Runner::default();
 
     runner.start(|context| async move {
-        let mut db = FixedDb::init(context.clone(), test_config(TEST_NAME))
-            .await
-            .expect("Failed to init source db")
-            .into_mutable();
+        let mut db = FixedDb::init(
+            context.clone(),
+            test_config(TEST_NAME, context.storage_buffer_pool().clone()),
```
patrick-ogrady: use from_pooler

storage/src/freezer/storage.rs
```diff
@@ -703,6 +706,7 @@ impl<E: Storage + Metrics + Clock, K: Array, V: CodecShared> Freezer<E, K, V> {
                     config.table_resize_frequency,
                     Some(checkpoint.epoch),
                     config.table_replay_buffer,
+                    &context,
```
patrick-ogrady: provide context first?

storage/src/journal/contiguous/fixed.rs
```diff
@@ -1423,7 +1423,7 @@ mod tests {
         // Start the test within the executor
         executor.start(|context| async move {
             // Initialize the journal, allowing a max of 3 items per blob.
-            let cfg = test_cfg(NZU64!(3));
+            let cfg = test_cfg(NZU64!(3), context.storage_buffer_pool().clone());
```
patrick-ogrady: provide pooler in cfg?

storage/src/journal/segmented/fixed.rs
```diff
@@ -488,7 +490,7 @@ mod tests {
         // Test that replay with a non-zero start_position correctly skips items.
         let executor = deterministic::Runner::default();
         executor.start(|context| async move {
-            let cfg = test_cfg();
+            let cfg = test_cfg(context.storage_buffer_pool().clone());
```
patrick-ogrady: nit: pooler?

storage/src/journal/segmented/oversized.rs
```diff
@@ -532,10 +535,12 @@ mod tests {
     fn test_oversized_append_and_get() {
         let executor = deterministic::Runner::default();
         executor.start(|context| async move {
-            let mut oversized: Oversized<_, TestEntry, TestValue> =
-                Oversized::init(context.clone(), test_cfg())
-                    .await
-                    .expect("Failed to init");
+            let mut oversized: Oversized<_, TestEntry, TestValue> = Oversized::init(
+                context.clone(),
+                test_cfg(context.storage_buffer_pool().clone()),
```
patrick-ogrady: use BufferPooler?

runtime/src/utils/buffer/paged/append.rs
```diff
@@ -623,18 +608,22 @@ impl<B: Blob> Append<B> {
...
-    ) -> (Vec<u8>, Option<Checksum>) {
+    ) -> (IoBuf, Option<Checksum>) {
         let logical_page_size = self.cache_ref.page_size() as usize;
         let physical_page_size = logical_page_size + CHECKSUM_SIZE as usize;
         let pages_to_write = buffer.data.len() / logical_page_size;
-        let mut write_buffer = Vec::with_capacity(pages_to_write * physical_page_size);
+        let mut write_buffer = self
+            .cache_ref
+            .pool()
+            .alloc((pages_to_write + 1) * physical_page_size);
```
patrick-ogrady: why +1?

runtime/src/utils/buffer/paged/append.rs
```diff
@@ -666,16 +655,22 @@ impl<B: Blob> Append<B> {
...
         let crc = Crc32::checksum(partial_page);
 
         // Pad with zeros to fill up to logical_page_size.
-        write_buffer.resize(write_buffer.len() + (logical_page_size - partial_len), 0);
+        let zero_count = logical_page_size - partial_len;
+        if zero_count > 0 {
+            let previous_len = write_buffer.len();
+            // SAFETY: We immediately initialize all newly exposed bytes.
+            unsafe { write_buffer.set_len(previous_len + zero_count) };
+            write_buffer.as_mut()[previous_len..previous_len + zero_count].fill(0);
```
patrick-ogrady: fill(0) necessary?

runtime/src/utils/buffer/paged/append.rs
```diff
@@ -845,8 +842,11 @@ impl<B: Blob> Blob for Append<B> {
         // Handle growing by appending zero bytes.
         if size > current_size {
             let zeros_needed = (size - current_size) as usize;
-            let zeros = vec![0u8; zeros_needed];
-            self.append(&zeros).await?;
+            let mut zeros = self.cache_ref.pool().alloc(zeros_needed);
+            // SAFETY: We fully initialize the buffer with zeros immediately.
+            unsafe { zeros.set_len(zeros_needed) };
+            zeros.as_mut().fill(0);
```
patrick-ogrady: fill 0?

runtime/src/utils/buffer/paged/cache.rs
```diff
@@ -319,17 +343,23 @@ impl CacheRef {
...
+    /// of `capacity` pages, each of size `page_size` bytes.
+    pub fn new(pool: BufferPool, page_size: NonZeroU16, capacity: NonZeroUsize) -> Self {
         let page_size = page_size.get() as usize;
         let capacity = capacity.get();
+        let mut slots = Vec::with_capacity(capacity);
+        for _ in 0..capacity {
+            let mut slot = pool.alloc(page_size);
+            // SAFETY: cache() always overwrites the full slot via copy_from_slice
+            // before any read, so initialization here is unnecessary.
+            unsafe { slot.set_len(page_size) };
```
patrick-ogrady: restore .fill()?

runtime/src/utils/buffer/paged/append.rs
andresilva: Because of the partial page, the previous code didn't have to handle this because it used `extend_from_slice`. I updated the code to make this more obvious.

runtime/src/utils/buffer/paged/append.rs
andresilva: Yes, when flushing a partial page we still write a full physical page so we need to have the unwritten tail initialized. I updated the code to use `put_bytes` instead, which avoids the unsafe `set_len`.

runtime/src/utils/buffer/paged/append.rs
andresilva: Same as above AFAIU.

runtime/src/utils/buffer/paged/cache.rs
andresilva: Slots are always overwritten before reads so this isn't strictly necessary, but probably makes sense to initialize for safety and the cost is only paid once at construction anyway.

storage/src/freezer/storage.rs
andresilva: For consistency with all config helpers made it the second argument, first argument is always a `&str` for `test_name`/`partition_suffix`.

runtime/src/utils/buffer/paged/cache.rs
```diff
@@ -319,17 +343,23 @@ impl CacheRef {
...
+    pub fn new(pool: BufferPool, page_size: NonZeroU16, capacity: NonZeroUsize) -> Self {
         let page_size = page_size.get() as usize;
         let capacity = capacity.get();
+        let mut slots = Vec::with_capacity(capacity);
+        for _ in 0..capacity {
+            let mut slot = pool.alloc(page_size);
+            // SAFETY: We immediately initialize all bytes below.
+            unsafe { slot.set_len(page_size) };
+            slot.as_mut().fill(0);
+            slots.push(slot);
```
patrick-ogrady, line 354: This will make more sense once #3093 is complete 😉

storage/src/qmdb/any/ordered/fixed.rs
```diff
@@ -195,17 +195,20 @@ pub(crate) mod test {
...
     }
 
     /// Create a test database with unique partition names
     pub(crate) async fn create_test_db(mut context: Context) -> CleanAnyTest {
         let seed = context.next_u64();
-        CleanAnyTest::init(context, fixed_db_config::<TwoCap>(&seed.to_string()))
-            .await
-            .unwrap()
+        CleanAnyTest::init(
+            context.clone(),
```
patrick-ogrady: we should just init config before context.clone()

storage/src/queue/conformance.rs
```diff
@@ -33,9 +33,10 @@ impl Conformance for QueueConformance {
     async fn commit(seed: u64) -> Vec<u8> {
         let runner = deterministic::Runner::seeded(seed);
         runner.start(|mut context| async move {
-            let mut queue = Queue::<_, Vec<u8>>::init(context.with_label("queue"), config(seed))
-                .await
-                .unwrap();
+            let mut queue =
+                Queue::<_, Vec<u8>>::init(context.with_label("queue"), config(seed, &context))
```
patrick-ogrady, line 37: This looks a bit weird (to use the &context for cache creation rather than the labeled context), but it is "most correct" (the page cache will be shared across instances to limit total memory usage).

I suppose a compelling follow-up question is very simply: should it be shared? Should different DBs be able to configure their own CacheRefs (I suppose this is allowed with the current approach while sharing wouldn't be).

storage/src/qmdb/immutable/mod.rs
```diff
@@ -528,19 +528,18 @@ pub(super) mod test {
...
-        Runner as _,
-    };
+    use commonware_runtime::{deterministic, BufferPooler, Runner as _};
     use commonware_utils::{NZUsize, NZU16, NZU64};
     use std::num::NonZeroU16;
 
     const PAGE_SIZE: NonZeroU16 = NZU16!(77);
     const PAGE_CACHE_SIZE: NonZeroUsize = NZUsize!(9);
+
     const ITEMS_PER_SECTION: u64 = 5;
```
patrick-ogrady, line 558: \n still exists

storage/src/qmdb/current/unordered/fixed.rs
```diff
@@ -162,9 +162,12 @@ pub mod test {
         context: deterministic::Context,
         partition_prefix: String,
     ) -> CleanCurrentTest {
-        CleanCurrentTest::init(context, fixed_config::<TwoCap>(&partition_prefix))
-            .await
-            .unwrap()
+        CleanCurrentTest::init(
+            context.clone(),
```
patrick-ogrady: we can avoid this clone

patrick-ogrady (dismissed):

runtime/src/iobuf/pool.rs
```diff
@@ -587,6 +587,34 @@ impl BufferPool {
...
+    }
+
+    /// Allocates a zero-initialized buffer with readable length `len`.
+    ///
+    /// Equivalent to `alloc(len)` followed by `put_bytes(0, len)`.
+    ///
+    /// Use this for read APIs that require an initialized `&mut [u8]`.
+    /// This avoids `unsafe set_len` at callsites, at the cost of zero-filling
+    /// `len` bytes before the read.
+    pub fn alloc_zeroed(&self, len: usize) -> IoBufMut {
```
andresilva, line 612: We could use https://doc.rust-lang.org/stable/std/alloc/fn.alloc_zeroed.html on the `AlignedBuffer` which should be faster, but maybe unnecessary since it will make this more complicated.

patrick-ogrady: Seems like a good follow-up?

```
index b0f49b6b6..abb6bf537 100644
@@ -43,7 +43,7 @@ use prometheus_client::{
     registry::Registry,
 };
 use std::{
-    alloc::{alloc, dealloc, Layout},
+    alloc::{alloc, alloc_zeroed, dealloc, Layout},
     mem::ManuallyDrop,
     num::NonZeroUsize,
     ptr::NonNull,
@@ -369,6 +369,21 @@ impl AlignedBuffer {
         Self { ptr, layout }
     }
 
+    /// Allocates a new zero-initialized buffer with the given capacity and alignment.
+    ///
+    /// # Panics
+    ///
+    /// Panics if allocation fails or alignment is not a power of two.
+    fn new_zeroed(capacity: usize, alignment: usize) -> Self {
+        let layout = Layout::from_size_align(capacity, alignment).expect("invalid layout");
+
+        // SAFETY: Layout is valid (non-zero size, power-of-two alignment).
+        let ptr = unsafe { alloc_zeroed(layout) };
+        let ptr = NonNull::new(ptr).expect("allocation failed");
+
+        Self { ptr, layout }
+    }
+
     /// Returns the capacity of the buffer.
     #[inline]
     const fn capacity(&self) -> usize {
@@ -425,6 +440,28 @@ impl SizeClass {
     }
 }
 
+/// Internal allocation result for pooled class allocations.
+enum AllocatedBuffer {
+    /// Existing buffer from freelist.
+    Reused(AlignedBuffer),
+    /// Freshly created buffer for an available slot.
+    New(AlignedBuffer),
+}
+
+impl AllocatedBuffer {
+    fn into_buffer(self) -> AlignedBuffer {
+        match self {
+            Self::Reused(buffer) | Self::New(buffer) => buffer,
+        }
+    }
+}
+
+/// How to initialize a newly created pooled buffer.
+enum NewBufferInit {
+    Uninitialized,
+    Zeroed,
+}
+
 pub(crate) struct BufferPoolInner {
     config: BufferPoolConfig,
@@ -435,6 +472,21 @@ pub(crate) struct BufferPoolInner {
 impl BufferPoolInner {
     /// Try to allocate a buffer from the given size class.
     fn try_alloc(&self, class_index: usize) -> Option<AlignedBuffer> {
+        self.try_alloc_internal(class_index, NewBufferInit::Uninitialized)
+            .map(AllocatedBuffer::into_buffer)
+    }
+
+    /// Try to allocate a buffer from the given size class, creating new
+    /// buffers with zero-initialized memory when possible.
+    fn try_alloc_zeroed(&self, class_index: usize) -> Option<AllocatedBuffer> {
+        self.try_alloc_internal(class_index, NewBufferInit::Zeroed)
+    }
+
+    fn try_alloc_internal(
+        &self,
+        class_index: usize,
+        new_buffer_init: NewBufferInit,
+    ) -> Option<AllocatedBuffer> {
         let class = &self.classes[class_index];
         let label = SizeClassLabel {
             size_class: class.size as u64,
@@ -447,14 +499,20 @@ impl BufferPoolInner {
                 self.metrics.allocations_total.get_or_create(&label).inc();
                 self.metrics.allocated.get_or_create(&label).inc();
                 self.metrics.available.get_or_create(&label).dec();
-                Some(buffer)
+                Some(AllocatedBuffer::Reused(buffer))
             }
             Some(None) => {
                 // Create new buffer (we have a slot)
                 class.allocated.fetch_add(1, Ordering::Relaxed);
                 self.metrics.allocations_total.get_or_create(&label).inc();
                 self.metrics.allocated.get_or_create(&label).inc();
-                Some(AlignedBuffer::new(class.size, class.alignment))
+                let buffer = match new_buffer_init {
+                    NewBufferInit::Uninitialized => AlignedBuffer::new(class.size, class.alignment),
+                    NewBufferInit::Zeroed => {
+                        AlignedBuffer::new_zeroed(class.size, class.alignment)
+                    }
+                };
+                Some(AllocatedBuffer::New(buffer))
             }
             None => {
                 // Pool exhausted (no slots available)
@@ -559,6 +617,15 @@ impl BufferPool {
         }
     }
 
+    /// Returns the size class index for `capacity`, recording oversized metrics on failure.
+    fn class_index_or_record_oversized(&self, capacity: usize) -> Option<usize> {
+        let class_index = self.inner.config.class_index(capacity);
+        if class_index.is_none() {
+            self.inner.metrics.oversized_total.inc();
+        }
+        class_index
+    }
+
     /// Allocates a buffer with the given capacity.
     ///
     /// The returned buffer has `len() == 0` and `capacity() >= capacity`,
@@ -604,14 +671,38 @@ impl BufferPool {
 
     /// Allocates a zero-initialized buffer with readable length `len`.
     ///
-    /// Equivalent to `alloc(len)` followed by `put_bytes(0, len)`.
-    ///
     /// Use this for read APIs that require an initialized `&mut [u8]`.
-    /// This avoids `unsafe set_len` at callsites, at the cost of zero-filling
-    /// `len` bytes before the read.
+    /// This avoids `unsafe set_len` at callsites, at the cost of ensuring
+    /// bytes in `0..len` are zeroed before the read.
     pub fn alloc_zeroed(&self, len: usize) -> IoBufMut {
-        let mut buf = self.alloc(len);
-        buf.put_bytes(0, len);
+        if let Some(class_index) = self.class_index_or_record_oversized(len) {
+            if let Some(allocated) = self.inner.try_alloc_zeroed(class_index) {
+                match allocated {
+                    AllocatedBuffer::Reused(buffer) => {
+                        // Reused buffers may contain old bytes; re-zero requested readable range.
+                        let mut buf =
+                            IoBufMut::from_pooled(PooledBufMut::new(buffer, Arc::downgrade(&self.inner)));
+                        buf.put_bytes(0, len);
+                        return buf;
+                    }
+                    AllocatedBuffer::New(buffer) => {
+                        // Newly created pooled buffers are zeroed in try_alloc_zeroed().
+                        let mut buf =
+                            IoBufMut::from_pooled(PooledBufMut::new(buffer, Arc::downgrade(&self.inner)));
+                        // SAFETY: buffer was allocated with alloc_zeroed, so bytes in 0..len are initialized.
+                        unsafe { buf.set_len(len) };
+                        return buf;
+                    }
+                }
+            }
+        }
+
+        // Pool exhausted or oversized: allocate untracked zeroed memory.
+        let size = len.max(self.inner.config.min_size.get());
+        let buffer = AlignedBuffer::new_zeroed(size, self.inner.config.alignment.get());
+        let mut buf = IoBufMut::from_pooled(PooledBufMut::new(buffer, Weak::new()));
+        // SAFETY: buffer was allocated with alloc_zeroed, so bytes in 0..len are initialized.
+        unsafe { buf.set_len(len) };
         buf
     }
 
@@ -626,13 +717,9 @@ impl BufferPool {
     /// - [`PoolError::Oversized`]: `capacity` exceeds `max_size`
     /// - [`PoolError::Exhausted`]: Pool exhausted for required size class
     pub fn try_alloc(&self, capacity: usize) -> Result<IoBufMut, PoolError> {
-        let class_index = match self.inner.config.class_index(capacity) {
-            Some(idx) => idx,
-            None => {
-                self.inner.metrics.oversized_total.inc();
-                return Err(PoolError::Oversized);
-            }
-        };
+        let class_index = self
+            .class_index_or_record_oversized(capacity)
+            .ok_or(PoolError::Oversized)?;
 
         let buffer = self
             .inner
@@ -1077,6 +1164,41 @@ mod tests {
         assert!(buf.as_ref().iter().all(|&b| b == 0));
     }
 
+    #[test]
+    fn test_alloc_zeroed_fallback_uses_untracked_zeroed_buffer() {
+        let pool = BufferPool::new(test_config(page, page, 1), &mut registry);
+
+        // Exhaust pooled capacity for this class.
+        let _pooled = pool.try_alloc(100).unwrap();
+
+        let buf = pool.alloc_zeroed(100);
+        assert!(!buf.is_pooled());
+        assert_eq!(buf.len(), 100);
+        assert!(buf.as_ref().iter().all(|&b| b == 0));
+    }
+
+    #[test]
+    fn test_alloc_zeroed_reuses_dirty_pooled_buffer() {
+        let pool = BufferPool::new(test_config(page, page, 1), &mut registry);
+
+        let mut first = pool.alloc_zeroed(100);
+        assert!(first.is_pooled());
+        assert!(first.as_ref().iter().all(|&b| b == 0));
+
+        // Dirty the buffer before returning it to the pool.
+        first.as_mut().fill(0xAB);
+        drop(first);
+
+        let second = pool.alloc_zeroed(100);
+        assert!(second.is_pooled());
+        assert_eq!(second.len(), 100);
+        assert!(second.as_ref().iter().all(|&b| b == 0));
+    }
+
     #[test]
     fn test_pool_exhaustion() {
         let page = page_size();

```

andresilva: https://github.com/commonwarexyz/monorepo/issues/3104

patrick-ogrady (approved):

runtime/src/utils/buffer/paged/append.rs
```diff
@@ -242,9 +242,7 @@ impl<B: Blob> Append<B> {
             let len = buf_guard.data.len();
             if len > 0 {
                 let mut shrunk = IoBufMut::with_capacity(len);
```
patrick-ogrady, line 244: Given this is for an "immutable blob", I figured it was worth sparing the pool here?

PR #3085, opened 2026-02-10
[utils] Add BitMap::is_unset_in_range

cronokirby:
Useful for efficiently checking the absence of an entire range.


utils/src/bitmap/mod.rs
```diff
@@ -572,6 +573,82 @@ impl<const N: usize> BitMap<N> {
             other.len()
         );
     }
+
+    /// Check if all the bits in a given range are 0.
+    ///
+    /// The semantics are that this function returns true if for every
+    /// index in the range, it is not the case that [`Self::get`] returns
+    /// `Some(1)` for that index. In particular, indices that are out of bounds
```
danlaine: nit: `get` returns a `bool`
```suggestion
    /// true for that index. In particular, indices that are out of bounds
```

utils/src/bitmap/mod.rs
```diff
@@ -572,6 +573,82 @@ impl<const N: usize> BitMap<N> {
             other.len()
         );
     }
+
+    /// Check if all the bits in a given range are 0.
+    ///
+    /// The semantics are that this function returns true if every index in the
+    /// range is either out of bounds, or such that [`Self::get`] returns true.
+    /// In particular, indices that are out of bounds of this bitmap do not.
```
danlaine: nit: errant `.`
```suggestion
    /// In particular, indices that are out of bounds of this bitmap do not
```

utils/src/bitmap/mod.rs
```diff
@@ -572,6 +573,82 @@ impl<const N: usize> BitMap<N> {
             other.len()
         );
     }
+
+    /// Check if all the bits in a given range are 0.
+    ///
+    /// The semantics are that this function returns true if every index in the
+    /// range is either out of bounds, or such that [`Self::get`] returns true.
```
danlaine: Oops didn't catch this when I commented yesterday -- this should be `returns false` not `returns true`, right?

cronokirby: Yep, that's right!

utils/src/bitmap/mod.rs
```diff
@@ -576,6 +577,81 @@ impl<const N: usize> BitMap<N> {
...
     }
+
+    /// Check if all the bits in a given range are 0.
+    ///
+    /// Returns `true` if every index in the range is either out of bounds
+    /// or unset (i.e. [`Self::get`] returns `false`). Indices beyond the
+    /// length of the bitmap do not affect the result. Returns `true` if
+    /// the range is empty or entirely out of bounds.
+    pub fn is_unset(&self, range: Range<u64>) -> bool {
+        let start = range.start.min(self.len);
```
patrick-ogrady: AFAICT other methods panic in this case? Would we want to do the same for consistency?

PR #3086, opened 2026-02-10
[runtime / p2p] Add `closer` to the connection to allow for closing cleanly or force-closing

### LINKED ISSUE #1260: [runtime] Set SO_LINGER=0 on Connection Drop from Invalid IP or Failed Handshake
Use: https://docs.rs/tokio/latest/tokio/net/struct.TcpStream.html#method.set_linger

(AFAICT it defaults to `None` and leaves it to the OS default)

BrendanChou:
Fixes #1260


BrendanChou:
bugbot run

patrick-ogrady:
Thought about this for a good bit and decided there was no reason to not always enable this (we don't trust our peers...even if we have had a long-lived connection, we don't want them to consume resources for a moment longer than needed).

Replaced with  #3135

PR #3090, opened 2026-02-10
[cryptography] Refine DKG error messages

syrupipy:
Split the vague `DkgFailed` error into two specific variants:

- `InsufficientCommitments { got, required }` - when not enough valid dealer commitments are collected
- `PublicKeyMismatch` - when the reconstructed public key doesn't match the previous round during reshare

These two failure modes have very different implications: the first is typically a network/availability issue, while the second could indicate malicious behavior. Having distinct errors makes it easier to diagnose and respond appropriately in production.


cronokirby:
This is intentionally kept opaque, and I don't think that one can meaningfully respond differently in these cases.

PR #3092, opened 2026-02-10
[Storage] Implement historical proofs for `qmdb::current`

### LINKED ISSUE #2560: [storage/qmdb] Fix `CleanStore` for Current variants.
The `CleanStore` trait requires implementing `proof` and `historical_proof` methods. For current-style qmdb variants, this simply invokes those methods on the underlying anydb. But these proofs can't be compared against the current database's root, since no grafting is performed. What's needed is to return a current-db RangeProof, but this object is a different type than an any db's proof since it requires a bit of extra data.

danlaine:
Resolves #2560.

Replaces `qmdb::current`'s bitmap with a historical bitmap. Each call to `into_merkleized` creates an in-memory diff that allows for restoring the bitmap to its state at the time of merkleization. These in-memory diffs are cleaned up in `prune`. It is worth noting that `into_merkleized()`, and therefore the creation of the in-memory diffs that allow for serving historical range proof at the current size, does _not_ automatically happen when `qmdb::current` is committed.


storage/src/qmdb/current/db.rs
```diff
@@ -373,14 +539,25 @@ where
...
-        let new_grafted_leaves = status.complete_chunks();
+        let new_grafted_leaves = {
+            let len = status.len();
+            let total_chunks = len.div_ceil(BitMap::<N>::CHUNK_SIZE_BITS) as usize;
+            if len.is_multiple_of(BitMap::<N>::CHUNK_SIZE_BITS) {
+                total_chunks
+            } else {
+                total_chunks - 1
+            }
+        };
```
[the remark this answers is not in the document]
danlaine: Changed to use floor division

storage/src/qmdb/current/db.rs
```diff
@@ -223,14 +241,139 @@ where
...
+
+        // Rebuild the grafted MMR using the historical bitmap.
+        let grafted_mmr = build_grafted_mmr::<H, N>(
+            &mut StandardHasher::<H>::new(),
+            &historical_bitmap,
+            &pinned_nodes,
+            &self.any.log.mmr,
+            self.thread_pool.as_ref(),
+        )
+        .await?;
```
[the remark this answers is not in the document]
danlaine: Fixed in https://github.com/commonwarexyz/monorepo/pull/3092/changes/05b73d5741a39e094c74ddc288b05567bfee50b2

utils/src/bitmap/historical/bitmap.rs
```diff
@@ -122,12 +123,34 @@ pub struct BitMap<const N: usize, S: State = Clean> {
     state: S,
 }
 
+impl<const N: usize, S: State> BitMap<N, S> {
+    /// Return the earliest (lowest) commit number, if any commits exist.
```
roberto-bayardo, line 127: Is it possible for no commit to exist?  (QMDB ensures there's always at least one unpruned commit.)

roberto-bayardo: I guess the concept of a commit here isn't exactly 1-1 and you could use this without QMDB.

danlaine: > I guess the concept of a commit here isn't exactly 1-1 and you could use this without QMDB.

Yup, that's the idea. Unfortunately `commit` is a pretty overloaded term. It's especially confusing in this case because we commit the historical bitmap when we merkleize the `qmdb::current`, not when we _commit_ it. I would be open to changing the terminology but that might be overkill.

storage/src/qmdb/current/mod.rs
```diff
@@ -129,6 +129,56 @@
...
+//! During merkleization, bitmap chunks fully below the inactivity floor are pruned
+//! (`prune_to_bit`). All their bits are guaranteed to be 0 (inactive), so discarding
+//! them does not lose information. This advances `pruned_chunks` but does NOT advance
+//! the grafted MMR's in-memory pruning boundary.
+//!
+//! ### Metadata writes (in `sync` and `prune`)
+//!
+//! `sync_metadata()` persists two things to the metadata store:
+//! - The number of pruned bitmap chunks (`PRUNED_CHUNKS_PREFIX`).
+//! - The grafted MMR peak digests covering the pruned region (`NODE_PREFIX`).
```
roberto-bayardo, line 148: Do you mean the digests for the pinned nodes corresponding to the bitmap's pruning boundary?

storage/src/qmdb/current/mod.rs
```diff
@@ -129,6 +129,56 @@
...
+//! ### Grafted MMR pruning (in `prune`)
+//!
+//! `prune()` executes in this order:
+//!
+//! 1. **`sync_metadata()`** -- persist peaks before the ops log is pruned. If the
+//!    process crashes after this step but before step 2, the metadata is ahead of
+//!    the log, which is safe: recovery will recompute from the un-pruned log and
+//!    the metadata simply records peaks that haven't been pruned yet. The reverse
+//!    order (prune first, write metadata second) would be unsafe: a pruned log
+//!    with stale metadata would lose peak digests permanently.
```
roberto-bayardo, line 162: same concern, we call this pinned nodes elsewhere and have used "peaks" only to refer to the peaks of the current MMR's state.

utils/src/bitmap/historical/bitmap.rs
```diff
@@ -329,6 +342,34 @@ impl<const N: usize> CleanBitMap<N> {
         count - self.commits.len()
     }
 
+    /// Remove the commits whose post-commit `pruned_chunks` is below `min_pruned_chunks`.
+    ///
+    /// A commit's post-commit `pruned_chunks` equals the next commit's pre-commit
```
roberto-bayardo, line 347: I think it would help to define what you mean by "post-commit pruned chunks" in the header of this file, as you use that term a lot outside this function. My first interpretation of it was it must be chunks that follow the commit in the bitmap, leaving me confused.

danlaine:
We decided that we're going to push responsibility for serving historical range proofs to components outside `qmdb::current` rather than adding logic to `qmdb::current` to allow it to serve them.`qmdb::current` will continue to only store the active bitmap state rather than historical diffs over the bitmap.

PR #3094, opened 2026-02-11
runtime+p2p: align deterministic bind(0) semantics and preserve simulated churn scale

diegomrsantos:
## Why

`bind(0)` should mean the same thing across runtime implementations.

On `main`, deterministic networking treated `:0` as a literal socket key, while tokio/iouring treat it as "auto-assign a free local port". That mismatch makes cross-runtime behavior less predictable.

At the same time, simulated p2p tests rely on high peer churn. Using a small localhost auto-bind pool for every simulated peer risks turning churn-heavy scenarios into port-exhaustion failures.

## What

This PR does two focused things:

1. Align deterministic runtime semantics with tokio/iouring for `bind(0)`.
   - `bind(0)` now auto-assigns a free port from a deterministic pool.
   - It retries on conflicts and wraps the cursor instead of failing on the first collision.
   - Includes dedicated tests for conflict-skip and cursor-wrap behavior.

2. Preserve simulated-network scale for long-running churn.
   - Simulated peers use virtual loopback addresses in `127.0.0.0/8` (non-zero ports), rather than consuming a small localhost auto-bind pool.
   - Keeps simulated churn practical while staying deterministic.
   - Includes allocator-focused coverage and full simulated test validation.

## Why this shape

- The runtime fix establishes a clear cross-runtime contract for `bind(0)`.
- The p2p change keeps simulation behavior robust under churn, which is a core use case for deterministic testing.
- Together, this improves consistency without changing public APIs.

## Validation

- `just test -p commonware-runtime network::deterministic::tests::`
- `just test -p commonware-p2p simulated::`

### DIFF
--- a/p2p/src/simulated/network.rs
+++ b/p2p/src/simulated/network.rs
@@ -40,6 +40,12 @@ use tracing::{debug, error, trace, warn};
 /// Task type representing a message to be sent within the network.
 type Task<P> = (Channel, P, Recipients<P>, IoBuf, oneshot::Sender<Vec<P>>);
 
+/// 127.0.0.0/8 prefix used for simulated peer listener addresses.
+const LOOPBACK_PREFIX: u32 = 0x7F00_0000;
+const LOOPBACK_HOST_MASK: u32 = 0x00FF_FFFF;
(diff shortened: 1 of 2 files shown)


clabby:

Issue #3093, opened 2026-02-10
[runtime] Reduce page cache copies with `IoBuf` slots and `PooledBuf` recycling


andresilva:
The page cache currently stores mutable `IoBufMut` buffers in each slot. Every read, whether a cache hit or miss, copies data from the slot into the caller's buffer. Since `IoBuf` is reference-counted, we can instead store `IoBuf` in cache slots and hand readers a cheap clone. This eliminates all data copies on the read path: cache hits become an `Arc` bump, and cache misses read directly into a pool-allocated buffer, freeze it, store it, and return a clone.

To make this work well, we need a new `PooledBuf` type. Today, when a `PooledBufMut` is frozen, it goes through `Bytes::from_owner`, which means `Bytes::try_into_mut` always fails, there's no way to recycle the underlying allocation even when the refcount drops to 1. This matters for the cache because readers typically decode immediately and drop their reference, so by eviction time the cache almost always holds the only copy. `PooledBuf` solves this by wrapping the pool allocation in an `Arc` directly, with `try_into_mut` using `Arc::try_unwrap` to recycle when the refcount is 1. The shared core (`PooledBufInner`: aligned buffer + pool weak ref) is extracted from `PooledBufMut` so both mutable and immutable variants share the same drop logic that returns buffers to the pool. `PooledBuf` also supports zero-copy slicing (offset/len over the `Arc`'d inner) and a zero-copy `copy_to_bytes` via `Bytes::from_owner`.

On the API side, `read_shared` is added to `SectionBuffer` as the preferred read method, with a default implementation that delegates to `read_at` and freezes the result. Append overrides this to serve cached pages directly as `IoBuf` slices, multi-page reads return `IoBufs::Chunked` with one slice per page. Write overrides to return the blob portion zero-copy, copying only the tip buffer overlap (which is small and bounded by the tip buffer capacity). Storage callers switch from `blob.read_at(...).coalesce()` to `blob.read_shared(...)` and decode directly from `IoBufs`, which implements `Buf` (no coalesce needed).

Related #2883 #3082.

andresilva:
#3082 caused a significant performance regression on storage:

<img width="1488" height="1322" alt="Image" src="https://github.com/user-attachments/assets/ab57a2c7-e544-4c9a-afe9-8a1b664c3e45" />

We need to make sure that once this issue is closed we are back to (or better than) baseline before the pool introduction.

andresilva:
[...]

PR #3099, opened 2026-02-12
[runtime] pooled immutable `IoBuf` backing

andresilva:
This PR adds a new immutable pooled buffer backing for `IoBuf` via `PooledBuf`, so `IoBuf` can now be backed by either `Bytes` or pooled immutable storage. The implementation preserves content and ownership semantics while keeping zero-copy behavior for immutable operations, including slicing. Empty pooled slices detach from the underlying allocation instead of retaining it.

The core change is mutable ownership recovery. Pooled immutable buffers now support `try_into_mut` when uniquely owned, including uniquely owned sliced views, while preserving both readable content and the current readable window.

Related #3093 #2883.

### DIFF
--- a/runtime/src/iobuf/mod.rs
+++ b/runtime/src/iobuf/mod.rs
@@ -224,7 +224,20 @@ impl Buf for IoBuf {
     fn copy_to_bytes(&mut self, len: usize) -> Bytes {
         match &mut self.inner {
             IoBufInner::Bytes(b) => b.copy_to_bytes(len),
-            IoBufInner::Pooled(p) => p.copy_to_bytes(len),
+            IoBufInner::Pooled(p) => {
+                // Full non-empty drain: transfer ownership so the drained source no
+                // longer retains the pooled allocation. Keep len == 0 on the normal
+                // path to avoid creating an empty Bytes that still pins pool memory.
+                if len != 0 && len == p.remaining() {
+                    let inner = std::mem::replace(&mut self.inner, IoBufInner::Bytes(Bytes::new()));
+                    match inner {
+                        IoBufInner::Pooled(p) => Bytes::from_owner(p),
+                        IoBufInner::Bytes(_) => unreachable!(),
+                    }
+                } else {
+                    p.copy_to_bytes(len)
+                }
+            }
         }
     }
 }
--- a/runtime/src/iobuf/pool.rs
+++ b/runtime/src/iobuf/pool.rs
@@ -1502,6 +1502,94 @@ mod tests {
     }
 
     #[test]
+    fn test_copy_to_bytes_full_drain_releases_pool_from_source() {
+        let page = page_size();
+        let mut registry = test_registry();
+        let pool = BufferPool::new(test_config(page, page, 2), &mut registry);
+
+        let mut buf = pool.try_alloc(100).unwrap();
+        buf.put_slice(&[0xAB; 100]);
+        let mut iobuf = buf.freeze();
+        assert_eq!(get_allocated(&pool, page), 1);
+
+        // Full drain of remaining data.
+        let extracted = iobuf.copy_to_bytes(100);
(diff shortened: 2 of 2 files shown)


danlaine (commented):
Did a pass and didn't find any issues :+1:

patrick-ogrady (commented):
The LLMs think we should do something like the following (to mirror what `Bytes` does):

```
diff --git a/runtime/src/iobuf/mod.rs b/runtime/src/iobuf/mod.rs
index 72c7fd6f5..deaca36ee 100644
--- a/runtime/src/iobuf/mod.rs
+++ b/runtime/src/iobuf/mod.rs
@@ -224,7 +224,22 @@ impl Buf for IoBuf {
     fn copy_to_bytes(&mut self, len: usize) -> Bytes {
         match &mut self.inner {
             IoBufInner::Bytes(b) => b.copy_to_bytes(len),
-            IoBufInner::Pooled(p) => p.copy_to_bytes(len),
+            IoBufInner::Pooled(p) => {
+                if len == p.remaining() {
+                    // Full drain: transfer ownership so the source no longer
+                    // pins the pooled allocation.
+                    let inner = std::mem::replace(
+                        &mut self.inner,
+                        IoBufInner::Bytes(Bytes::new()),
+                    );
+                    match inner {
+                        IoBufInner::Pooled(p) => Bytes::from_owner(p),
+                        _ => unreachable!(),
+                    }
+                } else {
+                    p.copy_to_bytes(len)
+                }
+            }
         }
     }
 }
index 7ce451d55..5f19344f6 100644
@@ -1501,6 +1501,65 @@ mod tests {
         drop(extracted);
     }
 
+    #[test]
+    fn test_copy_to_bytes_full_drain_releases_pool() {
+        let pool = BufferPool::new(test_config(page, page, 2), &mut registry);
+
+        let mut buf = pool.try_alloc(100).unwrap();
+        buf.put_slice(&[0xAB; 100]);
+        let mut iobuf = buf.freeze();
+        assert_eq!(get_allocated(&pool, page), 1);
+
+        // Full drain: extract all remaining bytes.
+        let extracted = iobuf.copy_to_bytes(100);
+        assert_eq!(&extracted[..], &[0xAB; 100]);
+        assert_eq!(iobuf.remaining(), 0);
+
+        // The drained source must not pin the pooled allocation.
+        drop(iobuf);
+        // Only the extracted Bytes should hold the pool slot now.
+        assert_eq!(get_allocated(&pool, page), 1);
+
+        drop(extracted);
+        assert_eq!(get_allocated(&pool, page), 0);
+        assert_eq!(get_available(&pool, page), 1);
+    }
+
+    #[test]
+    fn test_copy_to_bytes_partial_then_full_drain_releases_pool() {
+        let pool = BufferPool::new(test_config(page, page, 2), &mut registry);
+
+        let mut buf = pool.try_alloc(100).unwrap();
+        buf.put_slice(&[0xCD; 100]);
+        let mut iobuf = buf.freeze();
+
+        // Partial drain first.
+        let partial = iobuf.copy_to_bytes(30);
+        assert_eq!(iobuf.remaining(), 70);
+        assert_eq!(get_allocated(&pool, page), 1);
+
+        // Full drain of remainder.
+        let rest = iobuf.copy_to_bytes(70);
+        assert_eq!(iobuf.remaining(), 0);
+
+        // Source should be detached after full drain.
+        drop(iobuf);
+
+        // Both extracted Bytes share the allocation.
+        assert_eq!(get_allocated(&pool, page), 1);
+
+        drop(partial);
+        assert_eq!(get_allocated(&pool, page), 1);
+
+        drop(rest);
+        assert_eq!(get_allocated(&pool, page), 0);
+        assert_eq!(get_available(&pool, page), 1);
+    }
+
     #[test]
     fn test_concurrent_clones_and_drops() {
         let page = page_size();

```

patrick-ogrady (approved):

Issue #2500, opened 2025-12-14
[cryptography] Review Namespace Constants


patrick-ogrady:
All namespace constants should be `_COMMONWARE_<CRATE>_<USAGE>`.

patrick-ogrady:
#2503 forces this issue re-open.

We should also review the `_DKG` usage (may be able to better break up).

PR #3108, opened 2026-02-13
[coding/zoda] Fix Namespace

patrick-ogrady:
Resolves: #2500
Resolves: #2506


cronokirby:
Maybe we should move this test up to the lib level, make it generic over the coding config, and have it run over all the schemes?

PR #3109, opened 2026-02-13
[runtime/iobuf] add fast paths for double and triple chunk `IoBufs(Mut)`

andresilva:
`IoBufs` and `IoBufsMut` are now opaque public types with private inner enums. Internally, they no longer treat every multi-chunk value the same: there are dedicated `Pair` and `Triple` fast paths for 2-and 3-chunk cases, and `Chunked` is used only for 4+ chunks. This avoids `VecDeque` overhead in the common small-multi-chunk paths.

The implementation now keeps canonical shapes as buffers are consumed or transformed. Empty/drained chunks are removed where appropriate, and representations collapse back down to smaller forms when possible.

On the API side, usage sites were migrated away from internal variant handling to a generic interface (`as_single` / `as_single_mut`), with a temporary contiguous-buffer fallback where vectored IO is not available.

`copy_to_bytes` was specialized for both `IoBufs` and `IoBufsMut` across pair/triple/chunked paths.

Related #2883.

### DIFF
--- a/runtime/src/iobuf/pool.rs
+++ b/runtime/src/iobuf/pool.rs
@@ -43,7 +43,7 @@ use prometheus_client::{
     registry::Registry,
 };
 use std::{
-    alloc::{alloc, dealloc, Layout},
+    alloc::{alloc, alloc_zeroed, dealloc, handle_alloc_error, Layout},
     mem::ManuallyDrop,
     num::NonZeroUsize,
     ops::{Bound, RangeBounds},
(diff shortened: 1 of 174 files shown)


runtime/src/iobuf/mod.rs
```diff
@@ -1003,16 +1316,25 @@ impl IoBufsMut {
...
                 .fold(0, usize::saturating_add),
         }
     }
 
-    /// Sets the length of the buffer(s) to `len`, distributing across chunks.
+    /// Sets the length of the buffer(s) to `len`, distributing across chunks
+    /// while preserving the current chunk layout.
+    ///
+    /// This is useful for APIs that must fill caller-provided buffer structure
+    /// in place (for example `Blob::read_at_buf`).
```
clabby: nit: link `Blob::read_at_buf`

runtime/src/iobuf/pool.rs
```diff
@@ -801,6 +801,17 @@ impl PooledBuf {
             Err(inner) => Err(Self { inner, offset, len }),
         }
     }
+
+    /// Converts this pooled view into `Bytes` without copying.
```
clabby: nit: link `Bytes`

runtime/src/iobuf/pool.rs
```diff
@@ -801,6 +801,17 @@ impl PooledBuf {
...
+    ///
+    /// Empty views return detached `Bytes::new()` so pooled memory is not
```
clabby: nit: link `Bytes::new`

runtime/src/lib.rs
```diff
@@ -694,16 +694,18 @@ stability_scope!(BETA {
         ///
         /// # Contract
         ///
-        /// - The output `IoBufsMut` is the same as the input, with `len` bytes filled from offset
+        /// - The returned buffers reuse caller-provided storage, with exactly `len`
+        ///   bytes filled from `offset`.
+        /// - Caller-provided chunk layout is preserved.
```
clabby, line 715: :heart:

runtime/src/utils/buffer/write.rs
```diff
@@ -110,56 +110,52 @@ impl<B: Blob> Blob for Write<B> {
...
+            let remaining = buffer.extract(buf.as_mut(), offset);
+
+            // If bytes remain, read directly from the blob. Any remaining bytes reside at the beginning
+            // of the range.
+            if remaining > 0 {
+                let blob_result = self
+                    .blob
+                    .read_at_buf(offset, remaining, self.pool.alloc(remaining))
+                    .await?;
+                buf.as_mut()[..remaining].copy_from_slice(blob_result.coalesce().as_ref());
```
clabby, line 131: Optional, but maybe add a comment as to why we can directly `copy_from_slice` here (safe, but a bit of machinery above.)

andresilva:
Did a pass over iobuf module for all missing links.

runtime/src/iobuf/mod.rs
```diff
@@ -683,66 +795,100 @@ impl IoBufs {
...
     ///
     /// Zero-copy if only one buffer. Copies if multiple buffers.
     #[inline]
     pub fn coalesce(mut self) -> IoBuf {
-        match self {
-            Self::Single(buf) => buf,
-            Self::Chunked(_) => self.copy_to_bytes(self.remaining()).into(),
+        match self.inner {
+            IoBufsInner::Single(buf) => buf,
+            _ => self.copy_to_bytes(self.remaining()).into(),
```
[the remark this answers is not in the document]
andresilva: Well it does compile 😅

PR #3111, opened 2026-02-13
[utils] specialized RNG for fuzzing

andresilva:
This PR replaces `BytesRng` with `FuzzRng` which is a deterministic RNG built to preserve mutation locality.

For each block, it reads a wrapping 8-byte window from the fuzz input at the current counter, interprets that window as a little-endian u64, XORs with the counter and a fixed domain constant, then applies a SplitMix64-style mixer.

The key property is locality: mutating one fuzz-input byte typically changes only the nearby output blocks whose windows include that byte (with wraparound), instead of globally changing the whole stream. That gives the fuzzer a smoother gradient, because small input mutations lead to mostly local randomness changes.

### DIFF
--- a/utils/src/rng.rs
+++ b/utils/src/rng.rs
@@ -42,7 +42,7 @@ const BLOCK_BYTES: usize = size_of::<u64>();
 ///
 /// block ctr = i:
 ///   word_i bytes = [b(i+0)%N, b(i+1)%N, ... b(i+7)%N]
-///   word_i       = little-endian u64 of those bytes
+///   word_i       = big-endian u64 of those bytes
 ///   out_i        = mix64(word_i ^ i ^ DOMAIN)
 /// ```
 ///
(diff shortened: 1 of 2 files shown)


andresilva:
We should ignore the conformance label check, doesn't make sense to tag this PR as `breaking-format` since `FuzzRng` is only meant to be used for testing.

utils/src/rng.rs
```diff
@@ -98,160 +171,235 @@ impl RngCore for BytesRng {
...
+        /// conformance covers variable-length inputs including non-aligned
+        /// lengths that exercise wrapping.
+        struct FuzzRngConformance;
 
-        impl Conformance for BytesRngConformance {
+        impl Conformance for FuzzRngConformance {
             async fn commit(seed: u64) -> Vec<u8> {
-                let mut rng = BytesRng::new(seed.to_be_bytes().to_vec());
+                let mut seed_rng = test_rng_seeded(seed);
+                let len = seed_rng.gen_range(1..=64);
```
patrick-ogrady, line 385: Better tests edge cases (with wrapping, etc)

PR #1834, opened 2025-10-08
[runtime] Introduce `ContextCell` + Remove `_ref`

patrick-ogrady:
Closes: #1692
Related: #1833 
Succeeds: #1817 

Add a "Builder API" on `Spawner` that permits configuring tasks with `Supervised (default)/Detached` and `Shared[blocking] (default)/Dedicated` behavior. This unblocks #1460 and paves the way for using dedicated threads on "hot loops".

PR #3113, opened 2026-02-13
[consensus/aggregation] fix unused rebroadcast metric

syrupipy:
The `rebroadcast` metric was registered to Prometheus but the field was dropped from the struct during refactoring in #1834. This caused the metric to always show 0.

Restore the field and record success/failure status in `handle_rebroadcast`.


clabby:

PR #3114, opened 2026-02-13
[Storage] Replace `_` with `-` as file name separator

danlaine:
Backward compatible; only changes test, benchmark, and example code.


patrick-ogrady:
Some of me wishes we kept everything `_` to match `context` labels but we are unfortunately past that point now.

roberto-bayardo:
super nitty & feel free to ignore: I noticed sometimes we use into() and sometimes we use to_string() for converting an str to a String. Maybe let's consistently use `into()` (just because it's shorter)?

danlaine:
> super nitty & feel free to ignore: I noticed sometimes we use into() and sometimes we use to_string() for converting an str to a String. Maybe let's consistently use `into()` (just because it's shorter)?

Will make a follow-up PR to do this in the storage crate

PR #3120, opened 2026-02-15
[coding] More Fixes

patrick-ogrady:
## Summary

  This PR hardens consensus/marshal/coding against adversarial inputs and tightens verification semantics.

  ## Motivation

  A Commitment includes block digest, coding root, context digest, and coding config.

  Shard reconstruction validated digest/root/config, but did not validate the context digest field. That allowed a forged commitment (wrong
  context hash) to map to reconstructed block data and be cached under the wrong commitment key.

  ## What Changed

  - Added missing reconstructed-context validation in shard reconstruction:
      - Check hash_context(reconstructed_block.context()) == commitment.context.
      - Return ContextMismatch and reject reconstruction on mismatch.
      - File: consensus/src/marshal/coding/shards/engine.rs
  - Hardened shard decoding against truncated input:
      - Replaced raw tag read with codec read to return EndOfBuffer instead of panicking.
      - File: consensus/src/marshal/coding/types.rs
  - Adjusted re-proposal cancellation semantics:
      - If verify receiver is dropped, do not synthesize false; resolve without explicit verdict.
      - File: consensus/src/marshal/coding/marshaled.rs

  ## Tests Added

  - test_failed_reconstruction_context_mismatch_then_recovery
      - Verifies forged context-hash commitment is rejected and engine still recovers for later valid commitment.
      - File: consensus/src/marshal/coding/shards/engine.rs
  - test_distribution_shard_decode_truncated_returns_error
      - Verifies truncated shard decode returns error and does not panic.
      - File: consensus/src/marshal/coding/types.rs
  - test_reproposal_verify_receiver_drop_does_not_synthesize_false
      - Verifies cancellation does not become an explicit negative verification result.
      - File: consensus/src/marshal/coding/mod.rs

### DIFF
--- a/consensus/src/marshal/coding/shards/engine.rs
+++ b/consensus/src/marshal/coding/shards/engine.rs
@@ -3777,7 +3777,11 @@ mod tests {
                 context.sleep(config.link.latency * 2).await;
 
                 assert!(
-                    peers[receiver_idx].mailbox.get(fake_commitment).await.is_none(),
+                    peers[receiver_idx]
+                        .mailbox
+                        .get(fake_commitment)
+                        .await
+                        .is_none(),
                     "block should not be available after ContextMismatch"
                 );
                 assert!(


clabby (approved):
Note that the check in the shard engine is redundant with the `Marshaled` validation, but I think it makes the checks a bit more legible with it being local to the shard engine.

Issue #3121, opened 2026-02-15
[marshal] Add Batched Signature Verification


patrick-ogrady:
During backfill, it can be MUCH more efficient to batch the verification of certificate signatures rather than doing 1-by-1.

Issue #3110, opened 2026-02-13
[marshal] Only Sync Finalization DBs When Processing Acknowledgement


patrick-ogrady:
When backfilling, syncing on each stored item slows down performance significantly. We should instead only sync when an application acknowledges delivery (concurrently with metadata sync).

Related: https://github.com/commonwarexyz/alto/pull/169

PR #3126, opened 2026-02-16
[marshal] Improve Sync Performance

patrick-ogrady:
Closes: #3110
Closes: #3121

This takes the `alto` follower from ~30 to 400 blocks-per-second.

### DIFF
--- a/storage/src/journal/segmented/oversized.rs
+++ b/storage/src/journal/segmented/oversized.rs
@@ -379,7 +379,8 @@ impl<E: BufferPooler + Storage + Metrics, I: Record + Send + Sync, V: CodecShare
                     .checked_add(u64::from(size))
                     .ok_or(Error::OffsetOverflow)?
             }
-            Ok(None) | Err(Error::SectionOutOfRange(_)) => 0,
+            Ok(None) => 0,
+            Err(Error::SectionOutOfRange(_)) if index_size == 0 => 0,
             Err(e) => return Err(e),
         };
 
@@ -403,7 +404,8 @@ impl<E: BufferPooler + Storage + Metrics, I: Record + Send + Sync, V: CodecShare
                     .checked_add(u64::from(size))
                     .ok_or(Error::OffsetOverflow)?
             }
-            Ok(None) | Err(Error::SectionOutOfRange(_)) => 0,
+            Ok(None) => 0,
+            Err(Error::SectionOutOfRange(_)) if index_size == 0 => 0,
             Err(e) => return Err(e),
         };
 
@@ -3017,6 +3019,42 @@ mod tests {
                 Err(Error::SectionOutOfRange(0))
             ));
             assert_eq!(oversized.value_size(0).await.unwrap(), 0);
+
+            oversized.destroy().await.expect("Failed to destroy");
+        });
+    }
+
+    #[test_traced]
+    fn test_rewind_nonzero_on_missing_section_errors() {
+        let executor = deterministic::Runner::default();
+        executor.start(|context| async move {
+            let cfg = test_cfg(&context);
+            let mut oversized: Oversized<_, TestEntry, TestValue> =
+                Oversized::init(context, cfg).await.expect("Failed to init");
+
+            let result = oversized.rewind(0, 1).await;
+            assert!(
+                matches!(result, Err(Error::SectionOutOfRange(0))),
+                "nonzero index_size on missing section must fail, got: {result:?}"
+            );
+
+            oversized.destroy().await.expect("Failed to destroy");
+        });
+    }
+
+    #[test_traced]
+    fn test_rewind_section_nonzero_on_missing_section_errors() {
+        let executor = deterministic::Runner::default();
+        executor.start(|context| async move {
+            let cfg = test_cfg(&context);
+            let mut oversized: Oversized<_, TestEntry, TestValue> =
+                Oversized::init(context, cfg).await.expect("Failed to init");
+
+            let result = oversized.rewind_section(0, 1).await;
+            assert!(
+                matches!(result, Err(Error::SectionOutOfRange(0))),
+                "nonzero index_size on missing section must fail, got: {result:?}"
+            );
 
             oversized.destroy().await.expect("Failed to destroy");
         });


consensus/src/marshal/actor.rs
```diff
@@ -562,223 +615,329 @@ where
...
+        wrote
+    }
+
+    /// Verify a batch of pending finalization/notarization certificates.
+    fn verify_pending_batch(
+        &mut self,
+        scheme: &P::Scheme,
+        batch: &[PendingVerification<P::Scheme, B>],
+    ) -> bool {
+        let certs = batch.iter().map(|item| match item {
```
patrick-ogrady: We should be able to compute this once.

consensus/src/marshal/actor.rs
```diff
@@ -562,223 +615,329 @@ where
...
-                                let Some(block) = self.find_block(&mut buffer, commitment).await
-                                else {
-                                    debug!(?commitment, "block missing on request");
-                                    continue;
-                                };
-                                response.send_lossy((notarization, block).encode());
-                            }
-                        }
+                let mut needs_sync = false;
+                let mut pending = Vec::new();
```
patrick-ogrady: could prealloc with max_repair?

patrick-ogrady:
```
  - [P1] Avoid syncing empty finalization archive — /Users/patrickogrady/code/monorepo-improve-sync-perf/consensus/src/marshal/actor.rs:1053-1056
    sync_finalization_archives unconditionally calls self.finalizations_by_height.sync() even when no finalization has ever been written. With
    immutable archives, this can persist a checkpoint for a non-existent section and make the next restart fail with SectionOutOfRange(0)
    (reproduced by marshal::tests::test_broadcast_caches_block), so bootstrap/restart paths can be broken before the first finalized certificate
    is stored.
    ```

**Comment by patrick-ogrady** [comment]
Don't "love" this yet but its a massive performance boost.

File: storage/src/journal/segmented/oversized.rs [hunk 3]
```diff
@@ -364,15 +364,19 @@ impl<E: BufferPooler + Storage + Metrics, I: Record + Send + Sync, V: CodecShare
...
         self.index.rewind(section, index_size).await?;
 
-        // Derive value size from last entry
-        let value_size = match self.index.last(section).await? {
-            Some(entry) => {
-                let (offset, size) = entry.value_location();
-                offset
-                    .checked_add(u64::from(size))
-                    .ok_or(Error::OffsetOverflow)?
+        // Derive value size from last entry (skip if section has no data)
```
patrick-ogrady: This was an unrelated bug.

consensus/src/marshal/actor.rs
```diff
@@ -59,6 +59,57 @@ use tracing::{debug, error, info, warn};
...
+
+        if width > 1 {
+            let mid = start + width / 2;
+            stack.push((mid, end));
+            stack.push((start, mid));
+        }
+    }
+
+    verified
+}
```
clabby: nit: Put this towards the bottom of the file.

consensus/src/marshal/actor.rs
```diff
@@ -562,223 +619,341 @@ where
...
+
+    /// Verify a batch of pending finalization/notarization certificates.
+    #[allow(clippy::type_complexity)]
+    fn verify_pending_batch<'a>(
+        &mut self,
+        scheme: &P::Scheme,
+        batch: &[(
+            Subject<'a, B::Commitment>,
+            &'a <P::Scheme as CertificateScheme>::Certificate,
+        )],
```
clabby: Let's create a struct or type alias for this to avoid `allow(clippy::type_complexity)`?

consensus/src/marshal/store.rs
```diff
@@ -20,28 +20,22 @@ pub trait Certificates: Send + Sync + 'static {
...
-    ///
-    /// * `height`: The application height associated with the finalization.
-    /// * `commitment`: The block commitment associated with the finalization.
-    /// * `finalization`: The finalization certificate.
-    ///
-    /// # Returns
-    ///
-    /// `Ok(())` once the write is synced, or `Err` if persistence fails.
+    /// Implementations must ignore overwrites for an existing finalization at the same
+    /// height or commitment.
```
clabby, line 28: For consistency with the other documentation comments on this trait, let's retain the `Arguments` and `Returns` section

storage/src/journal/segmented/oversized.rs
```diff
@@ -364,15 +364,19 @@ impl<E: BufferPooler + Storage + Metrics, I: Record + Send + Sync, V: CodecShare
...
+        let value_size = if index_size == 0 {
+            0
+        } else {
+            match self.index.last(section).await? {
```
clabby: Good catch

storage/src/journal/segmented/oversized.rs
```diff
@@ -364,15 +364,19 @@ impl<E: BufferPooler + Storage + Metrics, I: Record + Send + Sync, V: CodecShare
         // Rewind index first (this also removes sections after `section`)
         self.index.rewind(section, index_size).await?;
```
patrick-ogrady, line 373: TODO: add a comment that this protects against the case where index has never been written too (last isn't defined)?

consensus/src/marshal/actor.rs
```diff
@@ -883,6 +1058,25 @@ where
...
+            async {
+                self.finalized_blocks.sync().await.map_err(Box::new)?;
+                Ok::<_, BoxedError>(())
+            },
+            async {
+                self.finalizations_by_height
+                    .sync()
+                    .await
+                    .map_err(Box::new)?;
+                Ok::<_, BoxedError>(())
```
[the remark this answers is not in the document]
patrick-ogrady: This isn't an issue.

cryptography/src/certificate.rs
```diff
@@ -573,13 +633,160 @@ mod tests {
...
+        let attestations: Vec<_> = schemes
+            .iter()
+            .filter_map(|s| s.sign::<Sha256Digest>(subject(message)))
+            .collect();
+        schemes[0]
+            .assemble::<_, N3f1>(attestations, &Sequential)
+            .expect("assembly failed")
+    }
+
+    fn setup_ed25519(n: u32) -> (Vec<Ed25519Scheme>, Ed25519Scheme) {
```
patrick-ogrady, line 667: I think something like this is provided?

consensus/src/marshal/actor.rs
```diff
@@ -59,6 +59,22 @@ use tracing::{debug, error, info, warn};
 /// The key used to store the last processed height in the metadata store.
 const LATEST_KEY: U64 = U64::new(0xFF);
 
+/// A parsed-but-unverified resolver delivery awaiting batch certificate verification.
+enum PendingVerification<S: CertificateScheme, B: Block> {
```
patrick-ogrady, line 64: move this closer to where it is used.

consensus/src/marshal/actor.rs
```diff
@@ -42,7 +42,7 @@ use commonware_utils::{
     channel::{fallible::OneshotExt, mpsc, oneshot},
     futures::{AbortablePool, Aborter, OptionFuture},
     sequence::U64,
-    Acknowledgement, BoxedError,
+    Acknowledgement, BoxedError, N3f1,
```
patrick-ogrady: We should avoid hardcoding this?

cryptography/src/bls12381/certificate/multisig/mod.rs
```diff
@@ -238,17 +238,15 @@ impl<P: PublicKey, V: Variant, N: Namespace> Generic<P, V, N> {
         })
     }
 
-    /// Verifies a certificate.
-    pub fn verify_certificate<'a, S, R, D, M>(
+    /// Verifies a certificate without requiring an rng.
```
patrick-ogrady: nit: cleanup the comment

consensus/src/marshal/actor.rs
```diff
@@ -562,223 +589,378 @@ where
...
-                                else {
-                                    debug!(?commitment, "block missing on request");
-                                    continue;
-                                };
-                                response.send_lossy((notarization, block).encode());
-                            }
+                let mut needs_sync = false;
+                let mut remaining = self.max_repair.get();
+                let mut pending = Vec::with_capacity(remaining);
+                let mut produces = Vec::new();
```
patrick-ogrady, line 594: use ::new() or ::with_capacity() for all?

consensus/src/marshal/actor.rs
```diff
@@ -562,224 +583,352 @@ where
...
+                response.send_lossy((notarization, block).encode());
+            }
+        }
+    }
+
+    /// Handle a deliver message from the resolver. Block deliveries are handled
+    /// immediately. Finalized/Notarized deliveries are parsed and structurally
+    /// validated, then collected into `pending` for batch certificate verification.
+    /// Returns true if finalization archives were written and need syncing.
+    async fn handle_resolver_deliver(
```
patrick-ogrady: handle_deliver

patrick-ogrady:
There is a lot more we could optimize here but this is a good first push 🛠️

consensus/src/marshal/actor.rs
```diff
@@ -562,224 +583,348 @@ where
...
+                        ),
+                    )
+                else {
+                    response.send_lossy(false);
+                    return false;
+                };
+
+                if block.height() != height
+                    || finalization.proposal.payload != block.commitment()
+                    || finalization.epoch() != bounds.epoch()
```
patrick-ogrady, line 740: This shouldn't matter (will fail verification if epoch is wrong) but dosn't hurt.

broadcast/src/buffered/ingress.rs
patrick-ogrady, line 1: We should be able to roll this out more broadly (tokio's channels are `&self`)

clabby: #3077 should help with this - all generated mailboxes use `&self` now.

clabby (approved):
LGTM

storage/src/journal/segmented/oversized.rs
```diff
@@ -411,17 +411,15 @@ impl<E: BufferPooler + Storage + Metrics, I: Record + Send + Sync, V: CodecShare
 
     /// Get the value size for a section, derived from the last entry's location.
     pub async fn value_size(&self, section: u64) -> Result<u64, Error> {
-        match self.index.last(section).await {
-            Ok(Some(entry)) => {
+        self.index.last(section).await?.map_or_else(
```
BrendanChou: This now maps `Err(Error::SectionOutOfRange(_))` to itself rather than `Ok(0)`, is that intended?

consensus/src/simplex/types.rs
```diff
@@ -930,6 +930,24 @@ where
...
+    scheme: &S,
+    certificates: &[(Subject<'a, D>, &'a S::Certificate)],
+    strategy: &impl Strategy,
+) -> Vec<bool>
+where
+    R: CryptoRngCore,
+    S: scheme::Scheme<D>,
+    D: Digest,
+{
+    scheme.verify_certificates_bisect::<_, D, N3f1>(rng, certificates, strategy)
```
BrendanChou, line 948: why not just call this function directly rather than this "layer of indirection"?

patrick-ogrady: Didn't want to leak the N3f1 into marshal

storage/src/journal/segmented/oversized.rs
patrick-ogrady: Yes, this was intentional. We treated an empty blob differently than a non-existent blob here (which was causing bugs downstream on rewind).

patrick-ogrady: Discussed with @roberto-bayardo and he agrees.

consensus/src/marshal/actor.rs
```diff
@@ -562,224 +583,341 @@ where
...
+                let finalization = self.cache.get_finalization_for(commitment).await;
+                self.store_finalization(height, commitment, block, finalization, application)
+                    .await;
+                debug!(?commitment, %height, "received block");
+                response.send_lossy(true);
+                true
+            }
+            Request::Finalized { height } => {
+                let Some(bounds) = self.epocher.containing(height) else {
+                    response.send_lossy(false);
```
patrick-ogrady, line 717: Do we block when sending false? If so, this seems possibly broken (if our epocher is behind)?

storage/src/journal/segmented/oversized.rs
```diff
@@ -306,6 +306,13 @@ impl<E: BufferPooler + Storage + Metrics, I: Record + Send + Sync, V: CodecShare
     }
 
     /// Get the last entry for a section, if any.
+    ///
+    /// Returns `Ok(None)` if the section is empty.
```
patrick-ogrady, line 310: Opted to return an error here for consistency with `get()` operations

consensus/src/marshal/actor.rs
patrick-ogrady: We should add a comment that ensures an application upholds this invariant (we can add an issue to handle this).

patrick-ogrady: A compromise here would be to only fetch by digest in this case (although stupid slow).

clabby: The code currently doesn't support an `Epocher` that doesn't know of all epochs at initialization - we discussed, but this should definitely be documented.

At first glance, I'm not sure changing this assumption is a good idea. Having to traverse the application state to collect epoch length state changes is sub-optimal, since moving to digest only sync is crazy slow and would stop sync on the first block that's missing (versus now, where we can fill multiple gaps starting from direct finalizations we receive.)

For teams that want to do dynamic epoch lengths that are determined by the application, a middle-ground would maybe be having the entirety of the epoch -> height range history available in the state at the tip of the chain? For more simple setups where the length is just changed by a hardfork, even more trivial, they can just hard-code in the binary.

patrick-ogrady: As discussed offline, we can tackle this in a future PR: #3141

clabby (dismissed):

clabby (approved):

PR #3128, opened 2026-02-16
[utils] canonicalize lock primitives: `parking_lot` (blocking), `tokio::sync` (async)

### LINKED ISSUE #3125: [workspace] Re-Export and MIgrate to Tokio Mutex
Related: #3003 
Related: #2993

tokio's mutex* primitives are runtime agnostic and are much more performant than what we are using now.

andresilva:
This PR establishes and applies a clear synchronization policy across the workspace:

- Blocking locks: use `parking_lot` (via `commonware_utils::sync`)
- Non-blocking/async locks: use `tokio::sync` (via `commonware_utils::sync`)

It introduces a canonical `commonware_utils::sync` module, migrates lock imports to that module, and updates clippy `disallowed-type` rules to enforce the policy consistently.

For storage/journal paths that previously relied on upgradeable async rwlock behavior, this PR adds `UpgradableAsyncRwLock` (tokio-based) to preserve the same read-then-upgrade semantics.

This branch also replaces async locks with blocking locks in several hot paths where guards are not held over `.await` (for example, parts of MMR/journal state, runtime paged cache internals, marshal bookkeeping, p2p rate-limit state, and the sync example's last-operation timestamp). A big change (with a lot of fallout) is that In `storage/qmdb`, root getters were made synchronous.

Fix #3125.

### DIFF
--- a/storage/src/mmr/journaled.rs
+++ b/storage/src/mmr/journaled.rs
@@ -833,11 +833,10 @@ impl<E: RStorage + Clock + Metrics, D: Digest> DirtyMmr<E, D> {
             return Ok(());
         }
 
-        let clean_mmr = self.merkleize(hasher);
-        let journal_size = clean_mmr.journal.size().await;
-
         // Snapshot up to `write_limit` pending nodes while holding the read lock, then release
         // it before performing async journal writes.
+        let clean_mmr = self.merkleize(hasher);
+        let journal_size = clean_mmr.journal.size().await;
         let pending_nodes = {
             let inner = clean_mmr.inner.read();
             let mut pending_nodes = Vec::with_capacity(write_limit);


storage/src/mmr/journaled.rs
```diff
@@ -832,19 +834,25 @@ impl<E: RStorage + Clock + Metrics, D: Digest> DirtyMmr<E, D> {
...
-        for i in journal_size..*inner.mem_mmr.size() {
-            let node = *inner.mem_mmr.get_node_unchecked(Position::new(i));
-            clean_mmr.journal.append(node).await?;
-            written_count += 1;
-            if written_count >= write_limit {
-                break;
+        // Snapshot up to `write_limit` pending nodes while holding the read lock, then release
+        // it before performing async journal writes.
+        let pending_nodes = {
+            let inner = clean_mmr.inner.read();
```
patrick-ogrady, line 841: I don't think changing this around is an issue but need to think about it.

storage/src/queue/shared.rs
```diff
@@ -435,7 +434,9 @@ mod tests {
             assert!(result.is_none());
 
             drop(reader);
-            let _ = Arc::try_unwrap(queue).unwrap().into_inner();
+            let _ = Arc::try_unwrap(queue)
+                .unwrap_or_else(|_| panic!("queue should have a single reference"))
```
patrick-ogrady, line 438: expect?

coding/src/zoda.rs
patrick-ogrady, line 1: taking a stab at cleaning this up

storage/src/queue/shared.rs
patrick-ogrady: Ah, doesn't impl debug

codec/src/conformance.rs
```diff
@@ -37,20 +40,22 @@ where
...
-                    panic!(
-                        "failed to generate arbitrary value: NotEnoughData with {buffer_size} bytes"
-                    );
-                }
-                // Double the buffer size (capped at MAX) and retry
-                buffer_size = (buffer_size * 2).min(MAX_BUFFER_SIZE);
+        loop {
+            match T::arbitrary(&mut unstructured) {
+                Ok(value) => return value,
+                Err(arbitrary::Error::IncorrectFormat) => continue,
```
patrick-ogrady, line 46: NonZeroU* returns incorrect format for 0 (but we still want to keep).

PR #3130, opened 2026-02-16
fix(p2p): correct rate limiter cleanup interval

ANtutov:
The cleanup condition `accepted > CLEANUP_INTERVAL` caused `retain_recent()` to run every 16,385 connections instead of the intended 16,384. Changed to `>=` so cleanup happens exactly at the configured interval.
 
Fixed in both discovery and lookup listeners.


clabby:

PR #3137, opened 2026-02-17
Add node-level Byzantine adversary fuzzer for Simplex consensus

dnkolegov-ar:
This is a work-in-progress experiment with another approach to fuzzing. The core idea is that instead of instantiating a cluster of nodes, e.g., 3 correct nodes and 1 or 2 byzantine nodes, we instantiate only one node on the test and load it with byzantine messages. The main goal is to be able to use fuzzing against the target node with high exec/s


dnkolegov-ar:
Must be merged after https://github.com/commonwarexyz/monorepo/pull/3111

PR #3138, opened 2026-02-17
Make fuzz tests easier to run on MacOS

cronokirby:
Some people had issues without this feature enabled locally, so this enables it conditionally, just for that platform.

### DIFF
--- a/runtime/Cargo.toml
+++ b/runtime/Cargo.toml
@@ -35,7 +35,6 @@ rand.workspace = true
 rand_core.workspace = true
 rayon.workspace = true
 sha2.workspace = true
-sysinfo.workspace = true
 thiserror.workspace = true
 tracing.workspace = true
 tracing-opentelemetry.workspace = true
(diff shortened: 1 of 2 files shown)


runtime/Cargo.toml
```diff
@@ -47,12 +46,19 @@ version = "0.2.15"
 features = ["js"]
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
+
+# Enable the apple-sandbox feature only on macos
```
clabby, line 50: If you place this comment under the `[target.'cfg(target_os = "macos")'.dependencies]` tag, `just fix-toml-fmt` won't bork it

cronokirby:
I think there might be actual issues with this approach, because the sandboxing feature messes up some of the sysinfo information that's needed in those tests. So, going down this approach, we would need to enable these features just for fuzz tests or something. Starting to feel like this isn't the right thing to do.

cronokirby:
So, the sandbox feature in fact just allows those disallowed process information calls to return 0, rather than failing.
  
  The result is that these two tests which use this information fail:
  
  1. runtime/src/process/metered.rs:88 process::metered::tests::test_process_metrics_init
  2. runtime/src/lib.rs:3738 tests::test_tokio_process_rss_metric
  
  The first is already feature gated to not run on windows. The second runs unconditionally.
  
  One thing we could to is to feature gate these tests.
  
  However, I fear that enabling the sandbox behavior by default on MacOS might lead to local development seeing weird results in a surprising way, when interacting with process metrics.
  
  My current position on this is that we should try and figure out what recommendations we can give to people running into sandboxing issues such that they can run things without a sandbox, making fuzz tests like other tests they run.

PR #3139, opened 2026-02-17
[Storage] Implement state sync for `qmdb::current`

danlaine:
Implements `sync::Database`, allowing it to be synced by the `qmdb::sync` engine.

`qmdb::current` state sync works by syncing its inner `qmdb::any`, which contains the operations, and then reconstructing the `qmdb::current`'s activity status bitmap from the operations.

To make `qmdb::current` state sync possible, this PR re-defines grafted leaf digests: grafted leaves whose corresponding bitmap chunks are all zeroes (i.e. all chunks below the inactivity floor and pruning boundary) take on the digest of the corresponding operation MMR node and do not incorporate the (empty) bitmap chunk. This allows the grafted MMR's pinned nodes (pruned peaks) to be derived from the operations `qmdb::any`.

This PR makes it possible to run `examples/sync` with `qmdb::current`.

### DIFF
--- a/storage/src/qmdb/current/db.rs
+++ b/storage/src/qmdb/current/db.rs
@@ -227,6 +227,7 @@ where
     ///
     /// # Errors
     ///
+    /// Returns [Error::OperationPruned] if `start_loc` falls in a pruned bitmap chunk.
     /// Returns [mmr::Error::LocationOverflow] if `start_loc` > [mmr::MAX_LOCATION].
     /// Returns [mmr::Error::RangeOutOfBounds] if `start_loc` >= number of leaves in the MMR.
     pub async fn range_proof(
--- a/storage/src/qmdb/current/ordered/fixed.rs
+++ b/storage/src/qmdb/current/ordered/fixed.rs
@@ -402,6 +402,49 @@ pub mod test {
         });
     }
 
+    /// Regression test: requesting a range proof for a location in a pruned bitmap chunk
+    /// must return `Error::OperationPruned`, not panic in the bitmap accessor.
+    #[test_traced("DEBUG")]
+    pub fn test_range_proof_returns_error_on_pruned_chunks() {
+        let executor = deterministic::Runner::default();
+        executor.start(|context| async move {
+            let partition = "range-proofs-pruned".to_string();
+            let mut hasher = StandardHasher::<Sha256>::new();
+            let mut db = open_db(context.with_label("db"), partition)
+                .await
+                .into_mutable();
+
+            let chunk_bits = BitMap::<32>::CHUNK_SIZE_BITS;
+
+            // Repeatedly update the same key to generate many inactive operations,
+            // pushing the inactivity floor past at least one full bitmap chunk.
+            let key = Sha256::fill(0x11);
+            for i in 0..chunk_bits + 10 {
+                let value = Sha256::hash(&i.to_be_bytes());
+                db.write_batch([(key, Some(value))]).await.unwrap();
+            }
+            let (db, _) = db.commit(None).await.unwrap();
+            let db = db.into_merkleized().await.unwrap();
+
+            assert!(
+                db.status.pruned_chunks() > 0,
+                "expected at least one pruned chunk"
+            );
+
+            // Requesting a range proof at location 0 (in the pruned range) should return
+            // OperationPruned, not panic.
+            let result = db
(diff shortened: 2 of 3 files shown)


storage/src/qmdb/current/mod.rs
```diff
@@ -129,6 +133,35 @@
 //! digest peaks covering the pruned region are persisted to metadata as "pinned nodes". On
 //! recovery, these pinned nodes are loaded and serve as opaque siblings during upward propagation,
 //! allowing the grafted tree to be rebuilt without the pruned chunks.
+//!
+//! # Root structure
+//!
+//! The canonical root of a `current` database is:
+//!
+//! ```text
+//! root = hash(grafted_root || ops_root)
```
roberto-bayardo: I wonder if root should equal `hash(ops_root || grafted_root [|| partial_digest])`. It's slightly more efficient to hash everything together.

storage/src/qmdb/current/grafting.rs
```diff
@@ -285,10 +285,16 @@ impl<H: CHasher> HasherTrait for Verifier<'_, H> {
...
                 };
 
-                // grafted_leaf = hash(chunk || ops_subtree_root)
-                self.hasher.inner().update(self.chunks[local]);
-                self.hasher.inner().update(&ops_subtree_root);
-                self.hasher.inner().finalize()
+                // For all-zero chunks, the grafted leaf is the ops subtree root (identity).
+                // For non-zero chunks: grafted_leaf = hash(chunk || ops_subtree_root).
+                let chunk = self.chunks[local];
+                if chunk.iter().all(|&b| b == 0) {
```
roberto-bayardo, line 291: I think the typical chunk size will be 1000+ bits, so I wonder if we should optimize this (can we pass in inactivity floor?)

roberto-bayardo: Wait, wasn't @andresilva adding something related to quickly checking (or was it setting) for all-0 or all-1?

I guess here we're doing byte-by-byte check not bit-by-bit, so perhaps it's fine.

storage/src/qmdb/current/sync/tests.rs
```diff
@@ -0,0 +1,902 @@
+//! Tests for [crate::qmdb::current] state sync.
```
roberto-bayardo, line 1: Is it possible to make any of these tests generic and able to run on either any or current type dbs?

storage/src/qmdb/current/mod.rs
danlaine: Done

storage/src/qmdb/current/grafting.rs
danlaine: I suspect that the compiler is smart enough to optimize these per-byte is-zero checks. I'm not sure passing in the inactivity floor would be worth the added API complexity.

examples/sync/src/databases/current.rs
```diff
@@ -0,0 +1,179 @@
+//! Current database types and helpers for the sync example.
+//!
+//! A `current` database extends an `any` database with an activity bitmap that tracks which
+//! operations are active (i.e. represent the current state of their key) vs inactive
+//! (superseded or deleted). Its canonical root is `hash(grafted_root || ops_root)`.
```
roberto-bayardo: I guess this needs updating now?

storage/src/qmdb/current/sync/tests.rs
danlaine: Yeah let me try to de-deduplicate these

storage/src/qmdb/current/mod.rs
roberto-bayardo: link to the fancier grafted leaf formula here?

examples/sync/src/databases/current.rs
danlaine: Fixed

storage/src/qmdb/current/mod.rs
danlaine: Updated comment

storage/src/qmdb/current/sync/tests.rs
danlaine: https://github.com/commonwarexyz/monorepo/pull/3139/changes/ec1f0ccc896f2b713fd9925699b8e9f6e1ff8398

Had a crack at de-deduplicating in this commit ^

PR #3148, opened 2026-02-17
[runtime] Mark `Pacer` trait as `BETA`

patrick-ogrady:


runtime/src/lib.rs
```diff
@@ -758,7 +758,7 @@ stability_scope!(BETA {
         fn storage_buffer_pool(&self) -> &BufferPool;
     }
 });
-stability_scope!(ALPHA, cfg(feature = "external") {
+stability_scope!(BETA, cfg(feature = "external") {
```
patrick-ogrady, line 761: We aren't marking any impl of `Pacer` as BETA, just the trait 👍

PR #3155, opened 2026-02-18
[consensus/marshal] Test batch ack persistence

clabby:
## Overview

Adds a test to `marshal` that shows, after a batch of pending acknowledgements is processed by marshal and a node is restarted, the node restores the correct processed height (the tip of the batch.)

### DIFF
--- a/consensus/src/marshal/mod.rs
+++ b/consensus/src/marshal/mod.rs
@@ -762,6 +762,12 @@ mod tests {
             assert_eq!(application.acknowledge_next(), Some(Height::new(2)));
             assert_eq!(application.acknowledge_next(), Some(Height::new(3)));
 
+            // Yield to marshal
+            context.sleep(Duration::from_millis(10)).await;
+
+            // Assert that the application has processed up to height 3.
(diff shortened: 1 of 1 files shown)


consensus/src/marshal/mod.rs
```diff
@@ -702,6 +702,80 @@ mod tests {
...
+            assert_eq!(
+                application.pending_ack_heights(),
+                vec![Height::new(1), Height::new(2), Height::new(3)]
+            );
+
+            // Acknowledge all pending blocks without yielding so marshal can
+            // drain them in one ack arm and then sync metadata once.
+            assert_eq!(application.acknowledge_next(), Some(Height::new(1)));
+            assert_eq!(application.acknowledge_next(), Some(Height::new(2)));
+            assert_eq!(application.acknowledge_next(), Some(Height::new(3)));
```
patrick-ogrady, line 763: I think we may want to add a sleep after this to ensure these are processed?

clabby: Added an assertion for the app's height prior to restart

Issue #1045, opened 2025-06-02
[runtime] Figure out how to gate features/configs in the `tokio` runtime.


danlaine:
The "tokio" runtime (`runtime/src/tokio/runtime.rs`) used to always use [tokio](https://docs.rs/tokio/latest/tokio/) for networking and storage. Now, it can use io_uring for storage and/or networking. This makes it more of a "non-deterministic" runtime rather than a "tokio" runtime.

More thought should be given to how we want the user to be able to select from among the various runtime sub-components (like storage and networking). Right now, the choice for which network/storage implementation to use is gated behind feature flags (`iouring-network` and `iouring-storage` respectively.) A different approach would be to use generics instead. I.e. the "tokio" runtime would be parameterized on some `N: Network` and `S: Storage`. A third option would be to have an enum for each sub-component rather than generics. (I.e. for Storage there would be an enum variant for iouring and one for tokio).

When we decide what the right approach is, we should refactor the code along these lines.

patrick-ogrady:
One important thing to consider (if we go the full generic approach) is that some combos of different sub-components may not work together. The very nice property we currently have is that it is only possible to "successfully" instantiate things (outside of not being able to populate max connections for `iouring-network`).

In the minimal diff, we need to support more granular config-setting for each sub-component.

A good forcing function for this refactor may be how to sanely integrate #1040.

danlaine:
>One important thing to consider (if we go the full generic approach) is that some combos of different sub-components may not work together.

What are some combinations of sub-components that can't work together? Certainly deterministic implementations can't be used with non-deterministic implementations. (Or I guess you could, but that may limit the usefulness of the deterministic components.) But in general, I feel like things like networking, storage, etc. should generally be expected to not interfere with each other.

patrick-ogrady:
This will be required to properly configure `IOURING_NETWORK_SIZE`. It may be convenient to set this according to the value of `max_peer_set_size` (although the sum of peer sets could at most be `tracked_peer_sets` * `max_peer_set_size` if there are no overlaps): https://github.com/commonwarexyz/monorepo/blob/15dee3795f15cf850fd99578693e7067933204f0/p2p/src/authenticated/discovery/config.rs#L88-L92

PR #3159, opened 2026-02-18
[runtime/iouring] use eventfd-driven wakeups in io_uring loop

andresilva:
This PR moves the runtime io_uring loop to an eventfd-driven wake architecture. Producers enqueue operations and trigger wakeups through a shared eventfd, with an atomic pending latch to coalesce notifications and avoid redundant writes. The loop maintains an internal wake poll request, handles wake CQEs explicitly, and re-arms when multishot is no longer active.

As part of the change, the loop was simplified to non-blocking queue draining plus blocking `io_uring_enter` waits, and the old periodic fallback polling path was removed.

Related #2883.

### DIFF
--- a/runtime/src/iouring/mod.rs
+++ b/runtime/src/iouring/mod.rs
@@ -79,6 +79,40 @@
 //! 3. If `shutdown_timeout` is configured, abandons remaining operations after the timeout
 //! 4. Cleans up and exits. Dropping the last submitter signals `eventfd` so shutdown is observed
 //!    promptly even if the loop is blocked.
+//!
+//! ## Liveness Model
+//!
+//! This loop enforces a configured upper bound on in-flight operations, and submissions are staged
+//! from a FIFO MPSC queue.
+//!
+//! This implies a bounded-liveness caveat: if all in-flight operations are waiting on operations
+//! that are still queued behind the capacity limit, the loop cannot make progress until some
+//! in-flight operation completes or is canceled.
+//!
+//! Concrete example with `cfg.size = 2`:
+//!
+//! 1. Queue `read(fd1)`, `read(fd2)`, `write(fd1)`, `write(fd2)` in that order.
+//! 2. The loop stages the first two reads and reaches waiter capacity.
+//! 3. If each read depends on its corresponding write being submitted through the same loop, both
+//!    reads remain blocked.
+//! 4. The writes stay queued behind the capacity limit, so no completion is produced and the loop
+//!    cannot free capacity on its own.
+//!
+//! The runtime cannot infer dependency relationships between arbitrary queued and in-flight
+//! operations, so it cannot implement dependency-aware admission (and doing so generically would
+//! add substantial overhead).
+//!
+//! The practical way to recover from this condition is cancellation via per-op timeouts. When
+//! timed-out in-flight operations are canceled, waiter capacity is eventually released and queued
+//! operations can be staged. Without cancellation, liveness depends on workload structure: callers
+//! must avoid submission patterns where in-flight operations require later queued operations to
+//! run.
+//!
+//! Operational guidance:
+//! - Workloads that may create causal dependencies across queued and in-flight operations must use
+//!   per-op timeouts.
+//! - If cancellation is disabled, callers must guarantee that in-flight operations never depend on
+//!   later queued operations, otherwise the loop can deadlock.
 
 use crate::{IoBuf, IoBufMut};
 use commonware_utils::channel::{


runtime/src/iouring/mod.rs
```diff
@@ -261,222 +251,430 @@ pub struct Op {
     pub fd: Option<OpFd>,
 }
 
-// Returns false iff we received a shutdown timeout
-// and we should stop processing completions.
```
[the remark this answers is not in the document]
andresilva: Good catch!

runtime/src/iouring/mod.rs
```diff
@@ -261,142 +251,373 @@ pub struct Op {
...
+    ///
+    /// This method blocks the current thread.
+    pub(crate) fn run(mut self) {
+        let mut ring = new_ring(&self.cfg).expect("unable to create io_uring instance");
+
+        // Hybrid flush policy for the wake-fast path:
+        // - flush immediately once a small batch is staged
+        // - otherwise allow a small number of deferrals for batching
+        let batch_threshold = (self.cfg.size / 32).clamp(1, 32) as usize;
+        let max_defer_loops = 2;
```
patrick-ogrady: const?

runtime/src/iouring/mod.rs
```diff
@@ -261,142 +251,341 @@ pub struct Op {
...
+            if disconnected {
+                // All submitters are gone, drain in-flight ops and shutdown.
+                self.drain(&mut ring);
+                return;
+            }
+
+            self.metrics.pending_operations.set(self.waiters.len() as _);
+
+            // Submit pending SQEs and wait for a completion or wake event.
+            self.submit_and_wait(&mut ring, 1, None)
```
andresilva: At this point someone might have already submitted something again, in which case we'll enter the ring only to be instantly awoken. I wonder if we should use an atomic counter as well to detect this. This way we could keep looping, it would avoid the cost of constant wake up.

runtime/src/iouring/mod.rs
```diff
@@ -261,222 +261,632 @@ pub struct Op {
...
+
+        // Stage until we either run out of channel work or hit waiter capacity.
+        //
+        // Capacity is bounded by `cfg.size` active waiters. This remains correct
+        // when `op_timeout` is enabled because `new_ring` doubles ring size to
+        // accommodate `op + linked timeout` SQE pairs.
+        while self.waiters.len() < self.cfg.size as usize {
+            let op = match self.receiver.try_recv() {
+                Ok(work) => work,
+                Err(TryRecvError::Disconnected) => return None,
```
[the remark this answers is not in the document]
andresilva: I think it's fine as is. If the channel is disconnected we will always shut down the loop, so updating `processed_seq` is futile.

runtime/src/iouring/mod.rs
```diff
@@ -261,222 +261,635 @@ pub struct Op {
...
+
+    /// Disarm sleep intent after we resume running.
+    ///
+    /// Keeping sleep intent clear while actively running avoids redundant
+    /// eventfd writes during bursts. This is done both after a real wake and
+    /// after a post-arm recheck decides not to block.
+    fn disarm(&self) {
+        self.inner
+            .state
+            .fetch_and(!SLEEP_INTENT_BIT, Ordering::Release);
```
andresilva, line 449: I think this could be `Relaxed` instead with no downsides, but not sure whether it's worth it to be fancy.

runtime/src/iouring/mod.rs
```diff
@@ -261,222 +261,635 @@ pub struct Op {
...
+            self.ring();
+        }
+    }
+
+    /// Return the current submitted sequence.
+    ///
+    /// The sequence domain is masked to 63 bits and compared against the
+    /// loop-local `processed_seq` in the same domain.
+    fn submitted(&self) -> u64 {
+        (self.inner.state.load(Ordering::Acquire) >> 1) & SUBMISSION_SEQ_MASK
```
andresilva, line 422: I think this could be `Relaxed` without affecting correctness, but could lead to extra unnecessary loop iterations, which are more costly than just keeping this as `Acquire`.

runtime/src/iouring/mod.rs
```diff
@@ -162,20 +205,14 @@ pub struct Config {
...
+    /// `single_issuer` when `run` is executed on a dedicated thread.
     /// See IORING_SETUP_SINGLE_ISSUER in <https://man7.org/linux/man-pages/man2/io_uring_setup.2.html>.
     pub single_issuer: bool,
-    /// In the io_uring event loop (`run`), wait at most this long for a new
-    /// completion before checking for new work to submit to the io_ring. This
-    /// periodic wake-up prevents deadlocks where one task depends on completions
-    /// that won't arrive until another task submits additional work. Avoid
-    /// setting this to very low values, or the loop may burn CPU by waking
-    /// continuously even when no completions are available.
-    pub force_poll: Duration,
```
danlaine, line 178: 😍

danlaine (approved):
Really nicely done!

I'd be fine keeping everything in `iouring/mod.rs` but it's big enough now that I wouldn't be opposed to breaking it into a few files either.

andresilva:
I think I might break it up in follow-ups as I add more code (to deal with timeouts for example).

runtime/src/storage/iouring.rs
```diff
@@ -19,18 +19,15 @@
 //! ## Linux Only
 //!
 //! This implementation is only available on Linux systems that support io_uring.
+//! It requires Linux kernel 6.1 or newer. See [crate::iouring] for details.
```
patrick-ogrady, line 22: Should be more than conservative enough 👍

runtime/src/tokio/runtime.rs
```diff
@@ -50,9 +50,15 @@ use tracing::{info_span, Instrument};
...
-const IOURING_NETWORK_SIZE: u32 = 1024;
-#[cfg(feature = "iouring-network")]
-const IOURING_NETWORK_FORCE_POLL: Duration = Duration::from_millis(100);
+cfg_if::cfg_if! {
+    if #[cfg(test)] {
+        // Use a smaller ring in tests to reduce `io_uring_setup` failures
+        // under parallel test load due to mlock/resource limits.
+        const IOURING_NETWORK_SIZE: u32 = 128;
+    } else {
+        const IOURING_NETWORK_SIZE: u32 = 1024;
```
patrick-ogrady, line 59: At some point, we'll need to figure out how to specify this (#1045) or add documentation that peers are capped to X (maybe 4096)?

andresilva: It should still work with any number of peers, it will just cap the number of in-flight requests to 1024 (actually 1024 receives and 1024 sends, since we use one ring for each), and introduce latency. Either way your point still stands that we should expose configuration of this in the future.

runtime/src/iouring/mod.rs
```diff
@@ -261,222 +261,635 @@ pub struct Op {
...
+        // See IORING_SETUP_DEFER_TASKRUN in <https://man7.org/linux/man-pages/man2/io_uring_setup.2.html>.
+        builder = builder.setup_defer_taskrun();
+    }
+
+    // When `op_timeout` is set, each operation uses 2 SQ entries (op + linked
+    // timeout). We double the ring size to ensure users get the number of
+    // concurrent operations they configured. We also reserve one extra SQE for
+    // the internal wake poll.
+    let ring_size = if cfg.op_timeout.is_some() {
+        cfg.size
```
patrick-ogrady:
```
  Review comment:

  - [P1] Avoid non-power-of-two ring size requests — /Users/patrickogrady/code/monorepo-pr-3159/runtime/src/iouring/mod.rs:883-890
    Linux rounds io_uring_setup entry counts up to the next power of two, so the new +1 sizing (cfg.size + 1 and cfg.size * 2 + 1) doubles the
    actual ring for common power-of-two configs (for example, 1024 with timeouts becomes a 2049 request and 4096 allocation). This materially
    increases mmap/mlock usage and can make io_uring network/storage loops fail to start under memlock limits where the previous sizing worked.

```

andresilva: I fixed this by not adding `+1` to the requested ring size, instead when submitting we keep track of submission queue capacity, rather than just relying on `cfg.size`.

runtime/src/iouring/mod.rs
```diff
@@ -261,222 +261,639 @@ pub struct Op {
...
+            .multi(true)
+            .build()
+            .user_data(WAKE_WORK_ID);
+
+        // SAFETY: The poll SQE owns no user pointers and references a valid FD.
+        unsafe {
+            submission_queue
+                .push(&wake_poll)
+                .expect("wake poll SQE should always fit in the ring");
+        }
```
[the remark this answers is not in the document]
andresilva: This is called from `fill_submission_queue` before any other submission happens. When `fill_submission_queue` is called in the loop we are guaranteed to have at least one empty slot in the submission queue, otherwise we would have submitted in the previous loop iteration (since we would have been at capacity).

runtime/src/iouring/mod.rs
```diff
@@ -261,222 +261,639 @@ pub struct Op {
...
+                    // Pending submissions exist and staging stopped at capacity.
+                    //
+                    // Enter the kernel to submit pending SQEs and wait for at
+                    // least one completion so capacity can open up.
+                    self.submit_and_wait(&mut ring, 1, None)
+                        .expect("unable to submit to ring");
+                }
+
+                continue;
+            }
```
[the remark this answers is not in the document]
andresilva: It's not deferred indefinitely, it's deferred until either the submission queue is full or waiters is full. Either way, there is no correctness issue, it only increases tail latency under sustained submission load (we can add some heuristics in the future if needed), the current logic maximizes throughput.

runtime/src/iouring/mod.rs
```diff
@@ -261,222 +261,639 @@ pub struct Op {
...
+            // Update pending operations metric.
+            self.metrics.pending_operations.set(self.waiters.len() as _);
+
+            // If submissions are still pending, do not arm sleep.
+            //
+            // `submitted != processed_seq` means producers have published work we
+            // have not yet drained. Sleep here could park with pending work and
+            // no guaranteed eventfd wake, because publish only rings after sleep
+            // intent is armed.
+            if self.waker.submitted() != self.processed_seq {
```
patrick-ogrady, line 628: Only meaningful feedback is:
```
• The refactor introduces a deadlock path in the io_uring loop when published submissions are pending but staging is capacity-limited. In that
  state the loop can block indefinitely without any wake source, so the patch is not fully correct.

  Review comment:

  - [P1] Avoid indefinite wait while published work is still pending — /Users/patrickogrady/code/monorepo-pr-3159/runtime/src/iouring/mod.rs:594-
    601
    This branch can block forever in submit_and_wait(..., None) even though producers have already published additional work, because sleep
    intent is not armed here and those already-published submissions will not generate a new eventfd wake. The deadlock appears when staging
    stops at capacity before draining all published ops (for example, small rings or bursts where early ops depend on later queued ops), so the
    loop waits for a completion that cannot happen until it drains more submissions.
```

still pondering if an issue.

patrick-ogrady: Another said:
```
  One observation (not a bug): when submitted != processed_seq and !at_capacity, the loop continues without calling submit_and_wait, so SQEs
  batch up until either SQ capacity or the idle path submits them. This is a deliberate batching optimization that favors throughput at the cost
  of per-op latency under burst. Under low load, the loop goes idle and submits immediately, so single-op latency is not affected.
```

patrick-ogrady:
```
⏺ The reviewer is correct, and I missed this in my analysis. Let me walk through the concrete deadlock scenario.

  Consider cfg.size = 2 and four operations submitted in order: recv(fd1), recv(fd2), write(fd1), write(fd2):

  1. fill_submission_queue stages recv(fd1) and recv(fd2) → waiters.len() == 2 == cfg.size → at_waiter_capacity = true
  2. write(fd1) and write(fd2) remain in the MPSC
  3. submitted() = 4, processed_seq = 2 → not equal, at_capacity = true
  4. submit_and_wait(&mut ring, 1, None) — submits both recvs, blocks forever waiting for a CQE
  5. recv(fd1) needs write(fd1) which is stuck in the MPSC
  6. recv(fd2) needs write(fd2) which is stuck in the MPSC
  7. Deadlock — no CQE will ever arrive, sleep intent is not armed, so no new publishes will ring eventfd either

  This is exactly the scenario the old force_poll timeout was designed to break. The periodic wakeup would let the loop drain more from the MPSC
  and stage the writes. The new code removed that safety net without replacing it in this specific branch.

  My earlier analysis focused on the wake protocol (arm/recheck/publish atomics) and confirmed that's correct. But this deadlock is at a
  different level — it's about operation-level dependencies within the ring, not about missed wakeups in the sleep protocol. I implicitly assumed
   in-flight operations would eventually complete, which isn't true when completions depend on un-staged operations.

  Practical impact: with the default ring size of 128, you'd need all 128 in-flight ops to depend on ops still in the MPSC. In this codebase's
  production usage (network recv/send to remote peers, independent disk I/O), that doesn't arise. But it's a latent correctness bug that could
  bite with smaller rings or different workloads.

  Simplest fix: use a bounded wait in this branch, e.g.:

  self.submit_and_wait(&mut ring, 1, Some(Duration::from_secs(1)))

  This matches the old force_poll behavior for this specific path without affecting the idle sleep path (which correctly uses eventfd wakeups).
```

patrick-ogrady:
```
• Yes. The key distinction is what at_capacity proves.

  In runtime/src/iouring/mod.rs:605 (/Users/patrickogrady/code/monorepo-pr-3159/runtime/src/iouring/mod.rs:605), when submitted != processed_seq:

  - If !at_capacity, the loop just continues, so it keeps draining/staging and never blocks.
  - If at_capacity, it calls submit_and_wait(..., None) to flush staged SQEs and wait for one completion before trying to stage more.

  What this does guarantee:

  - It avoids the “go idle while work is already published” missed-wake problem.
  - It prevents pure userspace spinning when capacity is full.

  What it does not guarantee:

  - at_capacity does not imply a completion is actually possible.
  - If all staged ops are blocked waiting on later ops still stuck in the channel, submit_and_wait(..., None) can still block forever.

  So at_capacity is a progress heuristic, not a full liveness proof.
  ```

Comment (patrick-ogrady) [reply runtime/src/iouring/mod.rs:628]:
```
⏺ Nothing. You're right to push back — my suggested fix was wrong.

  After the timeout fires:

  1. submit_and_wait returns (no CQEs arrived)
  2. Loop continues → drain CQEs (none) → fill_submission_queue
  3. waiters.len() == cfg.size → can't stage anything from MPSC
  4. submitted != processed still true, at_capacity still true
  5. Back to submit_and_wait → timeout again
  6. Spin forever

  The deadlock is structural: you can't free waiter capacity without CQEs, and you can't get CQEs without staging the ops stuck in the MPSC. A
  timeout just turns a deadlock into a livelock.

  So either:
  - The scenario is unreachable in practice (operations through this ring never have local dependencies), making the concern moot
  - Or it IS reachable, and the fix requires something more fundamental — like decoupling waiter capacity from staging capacity so the loop can
  always drain the MPSC
  ```

patrick-ogrady: So, I suppose this comes down to exceeding capacity with items earlier items are blocked on in waiters?

patrick-ogrady:
```
⏺ Yes, exactly. The deadlock requires a causal chain: waiter ops can't complete until MPSC ops are submitted, but MPSC ops can't be staged until
  waiter ops complete.

  In practice, does that happen with this ring? The operations going through it are network recv/send to remote peers and disk read/write/fsync.
  A recv() completes when data arrives from the network — it doesn't depend on a send() being submitted through the same local ring. Disk ops
  complete independently of each other in the kernel. There's no local dependency between operations.

  So the reviewer's deadlock is mechanically real but requires operations with local causal dependencies through the same ring, which this
  runtime's workloads don't have.
  ```

patrick-ogrady: I think the right approach is documentation here (there isn't a real way to solve this)?

andresilva: AFAIU this liveness issue existed before with `force_poll` as well. Waking more often does not help when waiter capacity is saturated by ops that are causally waiting on ops still queued behind the cap. In this queue model we do not have enough semantic information to reliably identify which queued op would unblock which in-flight op, so dependency-aware admission is not something we can implement generically (and it would be terrible for performance anyway). So the realistic recovery mechanism is cancellation. Network already has per-op timeout, so cycles eventually break. Storage currently does not, enabling timeouts there would provide the same guarantee but adds overhead and may not be necessary for expected workloads. I'll add documentation on this.

andresilva: Also for network we use separate rings for send and receive, so even without timeouts I don't think you can create a dependency cycle.

patrick-ogrady (approved):

PR #3164, opened 2026-02-18
[consensus/marshal] Erasure coding

clabby:
## Overview

Re-opened version of #2208 to clear out comments.

### DIFF
--- a/consensus/src/marshal/application/validation.rs
+++ b/consensus/src/marshal/application/validation.rs
@@ -1,58 +1,15 @@
-//! Shared validation helpers for marshal verification and reconstruction.
+//! Shared validation helpers for marshal verification and certification.
 //!
-//! This module centralizes pure invariant checks used by both standard and
-//! coding marshal flows. Validators here return narrow error enums so call
-//! sites can make explicit decisions about logging, voting, and recovery.
+//! This module centralizes pure invariant checks shared across marshal verification
+//! and certification flows.
 
-use crate::{
-    marshal::coding::types::hash_context,
-    types::{coding::Commitment, Epoch, Epocher, Height, Round},
-    Block, CertifiableBlock, Epochable,
-};
-use commonware_codec::{EncodeSize, Write};
-use commonware_coding::Config as CodingConfig;
-use commonware_cryptography::{Committable, Digest, Hasher};
+use crate::types::{Epoch, Epocher, Height, Round};
 use commonware_utils::sync::Mutex;
 use std::sync::Arc;
 
 /// Cache for the last block built during proposal, shared between the
 /// proposer task and the broadcast path.
 pub(crate) type LastBuilt<B> = Arc<Mutex<Option<(Round, B)>>>;
-
-/// Validation failures for coding deferred verification.
-#[derive(Clone, Copy, Debug, PartialEq, Eq)]
-pub(crate) enum CodedBlockVerificationError {
-    Commitment,
-    ParentCommitment,
-    Epoch,
-    ParentDigest,
-    Height,
-    ContextHash,
-    Context,
-}
-
-/// Validation failures for standard deferred verification.
-#[derive(Clone, Copy, Debug, PartialEq, Eq)]
-pub(crate) enum StandardBlockVerificationError {
-    ParentDigest,
-    ExpectedParentDigest,
-    Height,
-}
-
-/// Validation failures for coding proposal verification.
-#[derive(Clone, Copy, Debug, PartialEq, Eq)]
-pub(crate) enum CodedProposalValidationError {
-    CodingConfig,
-    ContextHash,
-}
-
-/// Validation failures for coded block reconstruction.
-#[derive(Clone, Copy, Debug, PartialEq, Eq)]
-pub(crate) enum ReconstructionValidationError<D: Digest> {
-    BlockDigest,
-    CodingConfig,
-    ContextHash(D, D),
-}
 
 /// Returns true if the block is at an epoch boundary (last block in its epoch).
 #[inline]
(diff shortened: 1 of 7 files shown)


consensus/src/marshal/core/variant.rs
```diff
@@ -0,0 +1,157 @@
...
+    /// Extracts the block digest from a consensus commitment.
+    fn commitment_to_application(
+        commitment: Self::Commitment,
+    ) -> <Self::Block as Digestible>::Digest;
+
+    /// Converts a working block to an application block.
+    ///
+    /// This conversion cannot use `Into` due to orphan rules when `Block` wraps
+    /// `ApplicationBlock` (e.g., `CodedBlock<B, C> -> B`).
+    fn into_application_block(block: Self::Block) -> Self::ApplicationBlock;
```
clabby: @patrick-ogrady - suggestions for names?

coding/src/lib.rs
```diff
@@ -161,22 +161,22 @@ commonware_macros::stability_scope!(ALPHA {
...
+        type CheckingData: Clone + Send + Sync;
         /// A shard that has been checked for inclusion in the commitment.
         ///
-        /// This allows excluding [Scheme::ReShard]s which are invalid, and shouldn't
+        /// This allows excluding [Scheme::WeakShard]s which are invalid, and shouldn't
         /// be considered as progress towards meeting the minimum number of shards.
-        type CheckedShard;
-        type Error: std::fmt::Debug;
+        type CheckedShard: Clone + Send + Sync;
+        type Error: std::fmt::Debug + Send;
```
patrick-ogrady, line 180: Should have its own comment

coding/src/lib.rs
```diff
@@ -161,22 +161,22 @@ commonware_macros::stability_scope!(ALPHA {
     pub trait Scheme: Debug + Clone + Send + Sync + 'static {
```
patrick-ogrady, line 161: I suppose if we ever move away from ZODA, we can really simplify this...

coding/fuzz/src/lib.rs
```diff
@@ -69,27 +68,27 @@ pub fn fuzz<S: Scheme>(input: FuzzInput) {
     } = input;
 
     let config = Config {
-        minimum_shards: NZU16!(min),
```
patrick-ogrady, line 72: I like what was here better?

coding/src/lib.rs
```diff
@@ -161,22 +161,22 @@ commonware_macros::stability_scope!(ALPHA {
...
         /// A commitment attesting to the shards of data.
         type Commitment: Digest;
-        /// A shard of data, to be received by a participant.
-        type Shard: Clone + Debug + Eq + Codec<Cfg = CodecConfig> + Send + Sync + 'static;
-        /// A shard shared with other participants, to aid them in reconstruction.
+        /// A strong shard of data, to be received by a participant.
+        type StrongShard: Clone + Debug + Eq + Codec<Cfg = CodecConfig> + Send + Sync + 'static;
```
patrick-ogrady, line 165: Based on types in other files, it seems we should call this `Shard`?

consensus/src/marshal/ancestry.rs
```diff
@@ -0,0 +1,247 @@
...
+    /// the block.
+    fn fetch_block(
+        self,
+        digest: <Self::Block as Digestible>::Digest,
+    ) -> impl Future<Output = Option<Self::Block>> + Send;
+}
+
+/// Yields the ancestors of a block while prefetching parents, _not_ including the genesis block.
+///
+/// TODO(<https://github.com/commonwarexyz/monorepo/issues/2982>): Once marshal can also yield the genesis block,
```
patrick-ogrady: nit: TODO should not be in `///`

consensus/src/marshal/coding/shards/mailbox.rs
```diff
@@ -0,0 +1,189 @@
...
+    },
+    /// A request to open a subscription for the reconstruction of a [`CodedBlock`]
+    /// by its digest.
+    SubscribeByDigest {
+        /// The block's digest.
+        digest: B::Digest,
+        /// The response channel.
+        response: oneshot::Sender<Arc<CodedBlock<B, C>>>,
+    },
+    /// A request to prune all caches at and below the given commitment.
```
patrick-ogrady, line 77: In other parts of the Commonware Library, I believe prune is the "min" (so just below?)

consensus/src/marshal/coding/variant.rs
```diff
@@ -0,0 +1,87 @@
...
+    type ApplicationBlock = B;
+    type Block = CodedBlock<B, C>;
+    type StoredBlock = StoredCodedBlock<B, C>;
+    type Commitment = Commitment;
+
+    fn commitment(block: &Self::Block) -> Self::Commitment {
+        block.commitment()
+    }
+
+    fn commitment_to_application(
```
patrick-ogrady: I think we could call this `commitment_to_inner` and `into_inner` for `into_application_block`?

I think calling a specific thing "application" is confusion (as each layer is often an application of something else?)

Will ponder a bit more but that's my leading thoughts.

consensus/src/marshal/core/variant.rs
```diff
@@ -0,0 +1,157 @@
+//! Marshal variant and buffer traits.
+//!
+//! This module defines the core abstractions that allow the marshal actor to work
+//! with different block dissemination strategies:
```
patrick-ogrady, line 4: I think it would be highly valuable to explain how marshal thinks about commitment<>digest somewhere in here.

coding/src/lib.rs
clabby: I thought this was ambiguous. Just calling it `Shard` doesn't self-describe that it's carrying extra data IMO

(The original intention for the rename was some ambiguity in early reviews where we used the term "shard" more generally.)

consensus/src/marshal/core/variant.rs
```diff
@@ -0,0 +1,157 @@
...
+        + Into<Self::StoredBlock>
+        + Clone;
+
+    /// The application block type.
+    type ApplicationBlock: CertifiableBlock + Clone;
+
+    /// The type of block stored in the archive.
+    ///
+    /// Must be convertible back to the working block type via `Into`.
+    type StoredBlock: CertifiableBlock<Digest = <Self::Block as Digestible>::Digest>
```
patrick-ogrady: IIRC we send the "stored block" over the wire? So, I think a better term could be `WrappedBlock` or `CommitmentBlock`?

clabby: We do not - it's never sent over the wire, only persisted to disk. When we pull it off disk, it gets pulled back into `Variant::Block`.

consensus/src/types.rs
```diff
@@ -690,6 +693,202 @@ impl ExactSizeIterator for HeightRange {
...
+        use commonware_utils::{Array, Span};
+        use core::ops::{Deref, Range};
+        use rand_core::CryptoRngCore;
+
+        /// A [`Digest`] containing a coding commitment, encoded [`CodingConfig`], and context hash.
+        #[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+        pub struct Commitment([u8; Self::SIZE]);
+
+        impl Commitment {
+            const DIGEST_BYTES: usize = 32;
```
patrick-ogrady: nit: `DIGEST_SIZE`?

consensus/src/types.rs
```diff
@@ -690,6 +693,202 @@ impl ExactSizeIterator for HeightRange {
...
+
+            /// Extracts the [`Digest`] from this [`Commitment`].
+            ///
+            /// ## Panics
+            ///
+            /// Panics if the [`Digest`]'s [`FixedSize::SIZE`] is > 32 bytes.
+            fn take<D: Digest>(&self, range: Range<usize>) -> D {
+                const {
+                    assert!(
+                        D::SIZE <= 32,
```
patrick-ogrady, line 769: Check against const (`DIGEST_SIZE`)?

consensus/src/marshal/coding/variant.rs
clabby: I went with this for now.

consensus/src/types.rs
```diff
@@ -690,6 +693,202 @@ impl ExactSizeIterator for HeightRange {
...
+                buf: &mut impl bytes::Buf,
+                _cfg: &Self::Cfg,
+            ) -> Result<Self, commonware_codec::Error> {
+                if buf.remaining() < Self::SIZE {
+                    return Err(commonware_codec::Error::EndOfBuffer);
+                }
+                let mut arr = [0u8; Self::SIZE];
+                buf.copy_to_slice(&mut arr);
+
+                // Validate the embedded CodingConfig so that `config()` can
```
patrick-ogrady, line 824: This assumes there are no validity checks to perform on the digests...for completeness we could also "read" here too?

consensus/src/marshal/standard/mod.rs
```diff
@@ -0,0 +1,786 @@
...
+//!
+//! # Usage
+//!
+//! The standard variant uses the core [`crate::marshal::core::Actor`] and
+//! [`crate::marshal::core::Mailbox`] with [`Standard`] as the variant type parameter.
+//! Blocks are broadcast through [`commonware_broadcast::buffered`].
+//!
+//! # When to Use
+//!
+//! Prefer this variant when block sizes are small enough that shipping full blocks
```
patrick-ogrady, line 24: Or if participants have sufficiently powerful networking and want to avoid encoding overhead?

consensus/src/marshal/resolver/handler.rs
```diff
@@ -590,20 +578,42 @@ mod tests {
...
+        let digest = Sha256::hash(b"self");
+        let block = Request::<D>::Block(digest);
         assert_eq!(block.cmp(&block), std::cmp::Ordering::Equal);
         assert_eq!(min_finalized.cmp(&min_finalized), std::cmp::Ordering::Equal);
         assert_eq!(max_notarized.cmp(&max_notarized), std::cmp::Ordering::Equal);
     }
 
+    #[test]
+    fn test_cross_variant_hash_differs() {
+        use std::{
```
patrick-ogrady: move to top of mod

consensus/src/types.rs
clabby: We don't have the type of the inner digests accessible here - they're passed at the callsite of `take`.

consensus/src/marshal/core/actor.rs
```diff
@@ -1355,20 +1591,23 @@ where
...
         if !requests.is_empty() {
             resolver.fetch_all(requests).await
         }
         wrote
     }
 
+    /// Sets the processed height in storage, metrics, and in-memory state. Also cancels any
+    /// outstanding requests below the new processed height.
+    ///
     /// Buffers a processed height update in memory and metrics. Does NOT sync
```
patrick-ogrady, line 1588: duplicate comment (likely from merge)?

consensus/src/marshal/coding/marshaled.rs
```diff
@@ -0,0 +1,1008 @@
...
+
+        debug!(
+            round = %round,
+            commitment = %block.commitment(),
+            height = %block.height(),
+            "requested broadcast of built block"
+        );
+
+        self.shards.proposed(round, block).await;
+    }
```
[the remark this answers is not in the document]
clabby: This is fine, and genesis will never be re-proposed. Epoch length == 1 is illegal.

consensus/src/marshal/ancestry.rs
```diff
@@ -0,0 +1,247 @@
...
+                        Poll::Ready(None) | Poll::Pending => {}
+                    }
+                } else {
+                    // No more parents to fetch; Finish the stream.
+                    *this.pending.as_mut() = None.into();
+                }
+
+                Poll::Ready(Some(block))
+            }
+        }
```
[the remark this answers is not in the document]
clabby: Note that we push to `buffered`. The `pending` future won't get re-polled, we'll drain the buffer on next poll and set `pending` to `None` when it's exhausted.

consensus/src/marshal/coding/marshaled.rs
```diff
@@ -0,0 +1,1008 @@
+//! Wrapper for consensus applications that handles epochs, erasure coding, and block dissemination.
+//!
+//! # Overview
+//!
+//! [`Marshaled`] is an adapter that wraps any [`VerifyingApplication`] implementation to handle
+//! epoch transitions and erasure coded broadcast automatically. It intercepts consensus
```
patrick-ogrady, line 6: This is such a mic drop

consensus/src/marshal/coding/marshaled.rs
```diff
@@ -0,0 +1,1008 @@
...
+            application,
+            marshal,
+            shards,
+            scheme_provider,
+            strategy,
+            epocher,
+        } = cfg;
+
+        let build_duration =
+            Histogram::new([0.001, 0.0025, 0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1.0]);
```
patrick-ogrady: histogram const?

patrick-ogrady: I think there is one other spot I saw the same intervals?

consensus/src/marshal/coding/marshaled.rs
```diff
@@ -0,0 +1,1008 @@
...
+                    _ = tx.closed() => {
+                        debug!(
+                            reason = "consensus dropped receiver",
+                            "skipping verification"
+                        );
+                        return;
+                    },
+                    is_valid = validity_request => is_valid,
+                };
+                verify_duration.observe_between(start, runtime_context.current());
```
patrick-ogrady: I think there is a HistogramExt helper that will do this automatically?

consensus/src/marshal/coding/marshaled.rs
```diff
@@ -0,0 +1,1008 @@
...
+                if application_valid {
+                    marshal.verified(round, block).await;
```
patrick-ogrady, line 424: nit: add a comment that only persisted at this point?

consensus/src/marshal/coding/marshaled.rs
```diff
@@ -0,0 +1,1008 @@
...
+        let round = context.round;
+        let task = self.deferred_verify(context, payload, None);
+        self.verification_tasks.insert(round, payload, task).await;
+
+        match scheme.me() {
+            Some(_) => {
+                // Subscribe to shard validity. The subscription completes when a valid shard arrives.
+                let validity_rx = self.shards.subscribe_shard(payload).await;
+                let (tx, rx) = oneshot::channel();
+                self.context.clone().spawn(|_| async move {
```
patrick-ogrady: nit: with_label (for consistency)?

consensus/src/marshal/application/validation.rs
```diff
@@ -0,0 +1,628 @@
...
+        return Err(CodedBlockVerificationError::Epoch);
+    }
+    if block.parent() != parent.digest() {
+        return Err(CodedBlockVerificationError::ParentDigest);
+    }
+    if !has_contiguous_height(parent.height(), block.height()) {
+        return Err(CodedBlockVerificationError::Height);
+    }
+    let block_context = block.context();
+    if commitment.context::<Sha256Digest>() != hash_context(&block_context) {
```
patrick-ogrady: Do we want to hardcode Sha256 here? I think we should keep generic?

consensus/src/marshal/application/validation.rs
```diff
@@ -0,0 +1,628 @@
...
+use commonware_utils::sync::Mutex;
+use std::sync::Arc;
+
+/// Cache for the last block built during proposal, shared between the
+/// proposer task and the broadcast path.
+pub(crate) type LastBuilt<B> = Arc<Mutex<Option<(Round, B)>>>;
+
+/// Validation failures for coding deferred verification.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub(crate) enum CodedBlockVerificationError {
```
patrick-ogrady: I originally thought "big errors" was cleaner but was very wrong about that. This is optimal.

consensus/src/marshal/application/verification_tasks.rs
```diff
@@ -0,0 +1,48 @@
...
+use std::{collections::HashMap, hash::Hash, sync::Arc};
+
+type VerificationTaskMap<D> = HashMap<(Round, D), oneshot::Receiver<bool>>;
+
+#[derive(Clone)]
+pub(crate) struct VerificationTasks<D>
+where
+    D: Eq + Hash,
+{
+    inner: Arc<AsyncMutex<VerificationTaskMap<D>>>,
```
patrick-ogrady: does this need to be async?

consensus/src/marshal/coding/mod.rs
```diff
@@ -0,0 +1,1533 @@
...
+//! blocks are migrated into immutable archives. Any gaps are filled by asking peers for specific
+//! commitments through the resolver pipeline. The shard engine keeps only ephemeral, in-memory
+//! caches; once a block is finalized it is evicted from the reconstruction map, reducing memory
+//! pressure.
+//!
+//! # When to Use
+//!
+//! Choose this module when the consensus deployment wants erasure-coded dissemination with the
+//! same ordering guarantees provided by [`super::standard`]. The API mirrors the standard marshal,
+//! so applications can switch between the two by swapping the variant type and buffer implementation
```
patrick-ogrady: nit: I think we should rethink this wording...makes it sound like it isn't a breaking change?

consensus/src/marshal/coding/shards/engine.rs
```diff
@@ -0,0 +1,4061 @@
...
+        round: Round,
+    ) {
+        if self.reconstructed_blocks.contains_key(&commitment) {
+            return;
+        }
+        let Some(scheme) = self.scheme_provider.scoped(round.epoch()) else {
+            warn!(%commitment, "no scheme for epoch, ignoring external proposal");
+            return;
+        };
+        if scheme.me().is_none() {
```
patrick-ogrady, line 624: add a comment about skipping self?

consensus/src/marshal/coding/shards/engine.rs
```diff
@@ -0,0 +1,4061 @@
...
+                );
+            }
+            return;
+        }
+
+        let participants_len =
+            u64::try_from(participants.len()).expect("participant count impossibly out of bounds");
+        self.state.insert(
+            commitment,
+            ReconstructionState::new(leader, round, participants_len),
```
patrick-ogrady, line 650: should participants just take u16?

consensus/src/marshal/coding/shards/engine.rs
```diff
@@ -0,0 +1,4061 @@
...
+        queue.push_back(shard);
+    }
+
+    /// Ingest buffered pre-leader shards for a commitment into active state.
+    async fn ingest_buffered_shards(&mut self, commitment: Commitment) -> bool {
+        let mut buffered_weak = Vec::new();
+        let mut buffered_strong = Vec::new();
+        for (peer, queue) in self.pre_leader_buffers.iter_mut() {
+            let mut retained = VecDeque::with_capacity(queue.len());
+            for shard in queue.drain(..) {
```
patrick-ogrady: is there a more memory efficient way to do this (can we just keep the same VecDeque)?

consensus/src/marshal/coding/shards/engine.rs
```diff
@@ -0,0 +1,4061 @@
...
+            ReconstructionState::new(leader, round, participants_len),
+        );
+        let buffered_progress = self.ingest_buffered_shards(commitment).await;
+        if buffered_progress {
+            self.try_advance(sender, commitment).await;
+        }
+    }
+
+    /// Buffer a shard from a peer until a leader is known.
+    fn buffer_pre_leader_shard(&mut self, peer: P, shard: Shard<C, H>) {
```
patrick-ogrady: should we be checking this against the participants and/or cleaning out old buffer entries? Does broadcast do this (I suppose not)?

consensus/src/marshal/coding/shards/engine.rs
```diff
@@ -0,0 +1,4061 @@
...
+                .await;
+        }
+
+        // Cache the block so we don't have to reconstruct it again.
+        let block = Arc::new(block);
+        self.cache_block(block);
+
+        // Local proposals bypass reconstruction, so shard subscribers waiting
+        // for "our valid shard arrived" still need a notification.
+        self.notify_shard_subscribers(commitment);
```
patrick-ogrady, line 769: where do we cache this shard?

consensus/src/marshal/coding/shards/engine.rs
```diff
@@ -0,0 +1,4061 @@
...
+
+    /// Broadcasts a [`Shard`] to all participants.
+    async fn broadcast_weak_shard<Sr: Sender<PublicKey = P>>(
+        &mut self,
+        sender: &mut WrappedSender<Sr, Shard<C, H>>,
+        shard: Shard<C, H>,
+    ) {
+        let commitment = shard.commitment();
+        let _ = sender.send(Recipients::All, shard, true).await;
+        debug!(?commitment, "broadcasted shard to all participants");
```
patrick-ogrady: if you want to log here, we could also log how many peers we sent to?

consensus/src/marshal/coding/types.rs
```diff
@@ -0,0 +1,824 @@
...
+    /// The inner block type.
+    inner: B,
+    /// The erasure coding configuration.
+    config: CodingConfig,
+    /// The erasure coding commitment.
+    commitment: C::Commitment,
+    /// The coded shards.
+    ///
+    /// These shards are optional to enable lazy construction.
+    shards: Option<Vec<C::StrongShard>>,
```
patrick-ogrady, line 279: So, this isn't actually encoded?

consensus/src/marshal/coding/types.rs
```diff
@@ -0,0 +1,824 @@
...
+    type Cfg = <B as Read>::Cfg;
+
+    fn read_cfg(
+        buf: &mut impl bytes::Buf,
+        block_cfg: &Self::Cfg,
+    ) -> Result<Self, commonware_codec::Error> {
+        let inner = B::read_cfg(buf, block_cfg)?;
+        let config = CodingConfig::read(buf)?;
+
+        let mut buf = Vec::with_capacity(config.encode_size() + inner.encode_size());
```
patrick-ogrady: put sum in order of decoding

consensus/src/marshal/coding/types.rs
```diff
@@ -0,0 +1,824 @@
...
+///
+/// When reading from storage, we don't need to re-encode the block to compute
+/// the commitment - we stored it alongside the block when we first verified it.
+/// This avoids expensive erasure coding operations on the read path.
+///
+/// The [`Read`] implementation performs a light verification (block digest check)
+/// to detect storage corruption, but does not re-encode the block.
+pub struct StoredCodedBlock<B: Block, C: Scheme> {
+    commitment: Commitment,
+    inner: B,
```
patrick-ogrady, line 502: inner first?

consensus/src/marshal/coding/types.rs
```diff
@@ -0,0 +1,824 @@
...
+    let max_faults = N3f1::max_faults(n_participants);
+    assert!(
+        max_faults >= 1,
+        "Need at least 4 participants to maintain fault tolerance"
+    );
+    let max_faults = u16::try_from(max_faults).expect("max_faults must fit in u16");
+    let minimum_shards = NZU16!(max_faults + 1);
+    CodingConfig {
+        minimum_shards,
+        extra_shards: NZU16!(n_participants - minimum_shards.get()),
```
patrick-ogrady, line 648: nit: seems like this calculation could be a helper somewhere?

consensus/src/marshal/coding/types.rs
```diff
@@ -0,0 +1,824 @@
...
+#[cfg(test)]
+mod test {
+    use super::*;
+    use crate::{marshal::mocks::block::Block as MockBlock, Block as _};
+    use commonware_codec::{Decode, Encode};
+    use commonware_coding::{CodecConfig, ReedSolomon};
+    use commonware_cryptography::{sha256::Digest as Sha256Digest, Digest, Sha256};
+
+    const MAX_SHARD_SIZE: CodecConfig = CodecConfig {
+        maximum_shard_size: 1024 * 1024, // 1 MiB
```
patrick-ogrady, line 661: this seems like an odd param to me...how does the builder know how to enforce this (if the participant count is changing)?

consensus/src/marshal/coding/marshaled.rs
clabby: The same intervals are used in a few places, but tried to avoid adding constants because the buckets are situational and we shouldn't offer a path to not thinking about what they should be. Wdyt?

consensus/src/marshal/application/validation.rs
clabby: This is problematic because of `CodedBlock`'s `Commitable` function.

There's no real need for the user to control the hash function here, since it's a protocol detail?

clabby:
```rust
impl<B: CertifiableBlock, C: Scheme> Committable for CodedBlock<B, C> {
    type Commitment = Commitment;

    fn commitment(&self) -> Self::Commitment {
        Commitment::from((
            self.digest(),
            self.commitment,
            hash_context(&self.inner.context()),
            self.config,
        ))
    }
}
```

clabby: We'd have to add `PhantomData<H>` to `CodedBlock` just for this

consensus/src/marshal/coding/marshaled.rs
clabby: TIL about `Timed`, neat

consensus/src/marshal/coding/shards/engine.rs
clabby: It would be a nice follow up. The `Participant` type in `utils` is a `u32`, `coding` considers it a `u16`. https://github.com/commonwarexyz/monorepo/issues/3166

consensus/src/marshal/coding/shards/engine.rs
clabby: Yeah, broadcast doesn't do this, I retained the same behavior. Just a per-peer cap, expected to be small, though each peer's buffer isn't freed on disconnect.

consensus/src/marshal/coding/shards/engine.rs
clabby: We don't - we don't need to. The proposer calls `marshal.verified` directly before the shards are even broadcast, so they never need to participate in the reconstruction routine.

consensus/src/marshal/coding/types.rs
clabby: No. Because we send the full block over the wire in the recovery case, it's more efficient to re-encode upon arrival for commitment validation.

consensus/src/marshal/coding/types.rs
clabby: It's not done anywhere else, iiuc?

consensus/src/marshal/coding/types.rs
clabby: It's similar to `max_message_size` in p2p. Only an upper bound, but could change as the protocol operates or if blocks get bigger.

I think that similar to that parameter, chain operators will need to choose something that they're comfortable with.

consensus/src/marshal/coding/types.rs
clabby: (can also run this function and just access the `minimum_shards` field.)

patrick-ogrady: We can wait until we use it elsewhere 👍

consensus/src/marshal/coding/types.rs
patrick-ogrady: I think `    ///
    /// These shards are optional to enable lazy construction.` confused me.
    
    Do we always populate? If not, when do we not?

clabby: We don't populate when we use `CodedBlock::new_trusted` (i.e. after we've reconstructed a block from shards that were verified against a commitment.)

consensus/src/marshal/coding/shards/engine.rs
patrick-ogrady: Can we add an issue for this...with the p2p listener (like in resolver), we should be able to do this now?

If it is easy to do, I suggest we add now.

consensus/src/marshal/coding/shards/engine.rs
patrick-ogrady: Ah, gotcha. That's going to be hell to change now 😢

patrick-ogrady: I suppose its entirely impractical to do coding with > u16::MAX? Otherwise, tbh I'd be tempted to switch to u32 for consistency.

clabby: Yeah, unfortunate one since `reed_solomon_simd` requires the input values to be `u16`.

consensus/src/marshal/application/validation.rs
patrick-ogrady: Yeah, am aware of what would be required (we have to do this in other hash-agnostic places). IMHO we should keep open because ZK protocols may want to avoid Sha.

consensus/src/types.rs
patrick-ogrady: Should we have those types sooner 🤔 

I suppose this "struct as an array" is a bit of a new concept but don't think we should relax our parsing invariant in the generic case unless we hardcode a digest we know to treat all inputs as valid.

consensus/src/marshal/coding/shards/engine.rs
clabby: https://github.com/commonwarexyz/monorepo/issues/3171

I'll follow up with this - ideally we do it in both the shard engine and buffered engine in `broadcast` for consistency.

coding/src/lib.rs
patrick-ogrady: Do you think its worth changing other files that have Shard? to StrongShard (like no_coding or ZODA)?

consensus/src/marshal/coding/shards/engine.rs
```diff
@@ -658,8 +658,8 @@ where
 
     /// Buffer a shard from a peer until a leader is known.
     fn buffer_pre_leader_shard(&mut self, peer: P, shard: Shard<C, H>) {
```
patrick-ogrady: nit: `buffer_peer_shard`?

consensus/src/types.rs
clabby: I do not think so, personally. Consistency is still enforced throughout usage, and it avoids `coding::Commitment` having to be littered with generics everywhere.

coding/src/lib.rs
clabby: Their public interfaces already do by lieu of implementing the trait - I'll change the names of the internal types as well.

consensus/src/marshal/application/verification_tasks.rs
```diff
@@ -0,0 +1,47 @@
+use crate::types::Round;
+use commonware_utils::{channel::oneshot, sync::Mutex};
+use std::{collections::HashMap, hash::Hash, sync::Arc};
+
+type VerificationTaskMap<D> = HashMap<(Round, D), oneshot::Receiver<bool>>;
+
+#[derive(Clone)]
+pub(crate) struct VerificationTasks<D>
```
patrick-ogrady, line 18: nit: add a handful of comments (at least on structs?)

consensus/src/marshal/coding/marshaled.rs
```diff
@@ -0,0 +1,1018 @@
...
+        let MarshaledConfig {
+            application,
+            marshal,
+            shards,
+            scheme_provider,
+            strategy,
+            epocher,
+        } = cfg;
+
+        use prometheus_client::metrics::histogram::Histogram;
```
patrick-ogrady: top of mod?

consensus/src/marshal/coding/types.rs
patrick-ogrady: I think we should add some more detail to the comment here?

consensus/src/types.rs
patrick-ogrady: Just saying this `                D::read(&mut self.0[range].as_ref())
                    .expect("Commitment always contains a valid digest")` basically becomes a panic risk if a hasher ever supports fallible conversion?

consensus/src/marshal/coding/types.rs
clabby: done

consensus/src/types.rs
clabby: Correct. Changing this would also forbid us from using ZODA, since zoda's `Commitment` is its own custom `Digest` type (so we couldn't bind things around the codebase to `Commitment<H::Digest>`) without massive generic bloat.

consensus/src/marshal/coding/shards/mailbox.rs
```diff
@@ -0,0 +1,192 @@
...
+    SubscribeByDigest {
+        /// The block's digest.
+        digest: B::Digest,
+        /// The response channel.
+        response: oneshot::Sender<Arc<CodedBlock<B, C, H>>>,
+    },
+    /// A request to prune all caches at and below the given commitment.
+    Prune {
+        /// The prune target's [`Commitment`].
+        min: Commitment,
```
patrick-ogrady: I think `min` here is a bit deceptive (as noted before) because it isn't the min available...not sure what a better word is here (but agree it should be pruning <= commitment with how we use it).

consensus/src/types.rs
```diff
@@ -690,6 +693,202 @@ impl ExactSizeIterator for HeightRange {
...
+            /// Panics if the [`Digest`]'s [`FixedSize::SIZE`] is > 32 bytes.
+            fn take<D: Digest>(&self, range: Range<usize>) -> D {
+                const {
+                    assert!(
+                        D::SIZE <= 32,
+                        "Cannot extract Digest with size > 32 from Commitment"
+                    );
+                }
+
+                D::read(&mut self.0[range].as_ref())
```
patrick-ogrady, line 774: I think this is worth adding a unit test for (that panics) just to ensure this is very explicitly tracked (probably also with 32 panicking)?

consensus/src/types.rs
patrick-ogrady: So, this commitment just applies to RS coding?

patrick-ogrady: If so (which is fine), I wonder if we should put it in coding?

clabby: This commitment currently works with ZODA as well. I meant if we change it to be generic over both `Hasher` and `CodingScheme<Hasher>`, we get generic bloat everywhere.

consensus/src/types.rs
clabby: added

consensus/src/marshal/core/actor.rs
```diff
@@ -1257,74 +1415,121 @@ where
...
+                "buffer returned block with mismatched commitment"
+            );
+        }
+        let block = self
+            .find_block_in_storage(V::commitment_to_inner(commitment))
+            .await?;
+        if V::commitment(&block) == commitment {
+            return Some(block);
+        }
+        debug!(
```
patrick-ogrady: remove these defensive checks.

patrick-ogrady:
Fix this:

Random Commitment leaves context hash bytes as zero
Medium Severity

The new Commitment::random implementation fills bytes 0..CONFIG_OFFSET (0–63) with random data and writes a valid CodingConfig into bytes 64–67, but never fills the context hash region (bytes 68–99) with random data. Those 32 bytes remain as the initial all-zeros value. The old code used rng.fill_bytes(&mut buf) which covered all 100 bytes. This means every "random" Commitment shares the same context hash, reducing uniqueness and potentially causing collisions or incorrect test behavior wherever Random for Commitment is used.

patrick-ogrady:
bugbot run

PR #3167, opened 2026-02-19
[Storage] Fix `qmdb::current::ordered::exclusion_proof` panic

danlaine:
`exclusion_proof` assumes that the operations at location `size() - 1` is a `CommitFloor`, but after uncommitted deletes that may not be the case, causing it to hit an `unreachable!`.

This fix uses `last_commit_loc`, which always points to the actual `CommitFloor`, instead of `size() - 1` as the location of the last `CommitFloor`. Returns `Error::NotEmpty` if the last `CommitFloor` doesn't show the database is empty (i.e. database must be committed before exclusion proof can be created.)

### DIFF
--- a/storage/src/qmdb/current/ordered/db.rs
+++ b/storage/src/qmdb/current/ordered/db.rs
@@ -198,12 +198,10 @@ where
                 Ok(super::ExclusionProof::KeyValue(op_proof, key_data))
             }
             None => {
-                // The DB is empty. Use the last CommitFloor to prove
-                // emptiness. The Commit proof variant requires the
-                // CommitFloor's floor to equal its own location (genuinely
-                // empty at commit time). If this doesn't hold (e.g. uncommitted
-                // deletes emptied the DB), we can't generate a valid proof until
-                // the next commit.
+                // The DB is empty. Use the last CommitFloor to prove emptiness. The Commit proof
+                // variant requires the CommitFloor's floor to equal its own location (genuinely
+                // empty at commit time). If this doesn't hold (e.g. uncommitted deleted emptied
+                // the DB), we can't generate a valid proof until the next commit.
                 let op = self
                     .any
                     .log


roberto-bayardo:
is there a fuzz test over exclusive proving?

danlaine:
> is there a fuzz test over exclusive proving?

[Yup](https://github.com/commonwarexyz/monorepo/blob/main/storage/fuzz/fuzz_targets/current_ordered_operations.rs#L312)

roberto-bayardo:
> > is there a fuzz test over exclusive proving?
> 
> [Yup](https://github.com/commonwarexyz/monorepo/blob/main/storage/fuzz/fuzz_targets/current_ordered_operations.rs#L312)

Was that what turned this up?  (If not, I wonder why it's not hitting this?)

storage/src/qmdb/current/ordered/db.rs
```diff
@@ -189,34 +188,41 @@ where
...
-
-        let op_proof = self.operation_proof(hasher, loc).await?;
-
-        Ok(match span {
-            Some((_, key_data)) => super::ExclusionProof::KeyValue(op_proof, key_data),
             None => {
-                let value = match self.any.log.reader().await.read(*loc).await? {
-                    Operation::CommitFloor(value, _) => value,
-                    _ => unreachable!("last commit is not a CommitFloor operation"),
+                // The DB is empty. Use the last CommitFloor to prove
```
roberto-bayardo: nit: Mind reformatting this to 100 col?

roberto-bayardo:
I suppose a case could be made for limiting proof generation to the Clean (Durable + Merkleized) state.

danlaine:
> > > is there a fuzz test over exclusive proving?
> > 
> > 
> > [Yup](https://github.com/commonwarexyz/monorepo/blob/main/storage/fuzz/fuzz_targets/current_ordered_operations.rs#L312)
> 
> Was that what turned this up? (If not, I wonder why it's not hitting this?)

It does if you run it long enough. That's how I find this bug, actually. I was running this fuzz test while working on #3092 and ran into it then. I confirmed locally that the failing seed on `main` passes on this branch.

roberto-bayardo:
> It does if you run it long enough. That's how I find this bug, actually. I was running this fuzz test while working on #3092 and ran into it then. I confirmed locally that the failing seed on `main` passes on this branch.

fuzzing FTW!

danlaine:
> I suppose a case could be made for limiting proof generation to the Clean (Durable + Merkleized) state.

I had the same thought. I'm torn on this. I lean toward the approach in the PR (supporting exclusion proof generation on merkleized but not durable state) because `Error::NotEmpty` only happens when the database is empty which I presume isn't the common case. I could be convinced otherwise, though, if you feel strongly.

storage/src/qmdb/current/ordered/db.rs
danlaine: done

PR #3168, opened 2026-02-19
[consensus/marshal] Ignore delivered blocks below floor

clabby:
## Overview

Rejects blocks that are delivered below the sync floor in `marshal`, preventing a panic in the prunable archive.

### DIFF
--- a/consensus/src/marshal/actor.rs
+++ b/consensus/src/marshal/actor.rs
@@ -418,7 +418,8 @@ where
                 // Start with the ack that woke this `select_loop!` arm.
                 let mut pending = Some(self.pending_acks.complete_current(result));
                 loop {
-                    let (height, commitment, result) = pending.take().expect("pending ack must exist");
+                    let (height, commitment, result) =
+                        pending.take().expect("pending ack must exist");
                     match result {
                         Ok(()) => {
                             // Apply in-memory progress updates for this acknowledged block.
@@ -527,12 +528,13 @@ where
                                     Some(finalization),
                                     &mut application,
                                 )
-                                .await {
+                                .await
+                            {
                                 self.try_repair_gaps(&mut buffer, &mut resolver, &mut application)
-                                .await;
+                                    .await;
                                 self.sync_finalized().await;
+                                debug!(?round, %height, "finalized block stored");
                             }
-                            debug!(?round, %height, "finalized block stored");
                         } else {
                             // Otherwise, fetch the block from the network.
                             debug!(?round, ?commitment, "finalized block missing");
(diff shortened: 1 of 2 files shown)


consensus/src/marshal/actor.rs
```diff
@@ -794,6 +794,16 @@ where
...
                 // Persist the block, also storing the finalization if we have it
                 let height = block.height();
+                if height <= self.last_processed_height {
+                    debug!(
+                        %height,
+                        floor = %self.last_processed_height,
+                        "dropping block delivery at or below processed height floor"
+                    );
+                    // Stale due to local floor/pruning. Do not treat as invalid peer data.
+                    response.send_lossy(true);
```
clabby: Important detail. Dropping the channel or returning false explicitly would block the peer - since a node can arbitrarily update its sync floor, we can't attribute fault to a peer for this behavior.

clabby: https://github.com/commonwarexyz/monorepo/blob/c89d5c3ab5a4b828f521e1687ea7be8618ae0512/consensus/src/marshal/ingress/handler.rs#L80

https://github.com/commonwarexyz/monorepo/blob/c89d5c3ab5a4b828f521e1687ea7be8618ae0512/resolver/src/p2p/engine.rs#L427-L443

consensus/src/marshal/actor.rs
```diff
@@ -795,11 +796,12 @@ where
                 // Persist the block, also storing the finalization if we have it
                 let height = block.height();
                 let finalization = self.cache.get_finalization_for(commitment).await;
```
patrick-ogrady, line 806: While this lookup may not be necessary (if we get a pruned block), it will be pretty efficient to serve in such a case and is not something to worry about.

consensus/src/marshal/actor.rs
```diff
@@ -1207,7 +1209,20 @@ where
         block: B,
         finalization: Option<Finalization<P::Scheme, B::Commitment>>,
         application: &mut impl Reporter<Activity = Update<B, A>>,
-    ) {
+    ) -> bool {
+        // Blocks below the last processed height are not useful to us, so we ignore them (this
+        // has the nice byproduct of ensuring we don't call a backing store with a block below the
+        // pruning boundary)
+        if height <= self.last_processed_height {
```
patrick-ogrady, line 1224: I considered updating the impls of store to gracefully handle this case (and update the trait invariants to require it be handled correctly given we already handle duplicates), but this seemed better.

Open to convo.

patrick-ogrady: Admittedly, torn on how to best proceed given we do this in `cache` ...

patrick-ogrady: I ended up biasing towards this because it felt odd that we'd call an external thing with stuff we previously pruned.

patrick-ogrady (approved):

PR #3185, opened 2026-02-20
[runtime, storage] reduce copying around page cache and buffered adapters

### LINKED ISSUE #3093: [runtime] Reduce page cache copies with `IoBuf` slots and `PooledBuf` recycling

andresilva:
Fixes #3093.
Related #2883.


runtime/src/utils/buffer/paged/append.rs
```diff
@@ -231,23 +240,16 @@ impl<B: Blob> Append<B> {
...
             let mut buf_guard = self.buffer.write().await;
-            let len = buf_guard.data.len();
-            if len > 0 {
-                let mut shrunk = IoBufMut::with_capacity(len);
-                shrunk.put_slice(buf_guard.data.as_ref());
-                buf_guard.data = shrunk;
-            } else {
-                buf_guard.data = IoBufMut::default();
-            }
+            buf_guard.set_immutable(true);
```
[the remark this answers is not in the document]
andresilva: The boolean is for whether to compact the buffer or not.

storage/conformance.toml
```diff
@@ -1,10 +1,10 @@
 ["commonware_storage::archive::conformance::ArchiveImmutable"]
```
andresilva, line 1: The reason these changed is because `PAGE_SIZE` passed to the page cache is now the physical page size. We can keep the same hashes if we update all conformance tests to basically do `PAGE_SIZE + 12`. Is that worth it though?

runtime/src/iobuf/mod.rs
```diff
@@ -108,6 +108,33 @@ impl IoBuf {
...
+            return std::mem::take(self);
+        }
+
+        match &mut self.inner {
+            IoBufInner::Bytes(b) => Self {
+                inner: IoBufInner::Bytes(b.split_to(at)),
+            },
+            IoBufInner::Pooled(p) => Self::from_pooled(p.split_to(at)),
+        }
+    }
```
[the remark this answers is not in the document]
andresilva: > the panic message from the Bytes variant differs from the documented "split_to out of bounds" message

Wrong.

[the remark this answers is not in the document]
andresilva: > The Bytes-backed path will panic with a different message than "split_to out of bounds".

No, it won't (https://docs.rs/bytes/latest/src/bytes/bytes.rs.html#554-559).

runtime/src/utils/buffer/write.rs
clabby, line 96: I wonder if we should be a bit more clever with our locking here - this would force a write to be serialized even on a different range?

clabby: i.e. `read_at(0, 1)` + `write_at(2, 4)` in parallel should be possible?

runtime/src/utils/buffer/read.rs
```diff
@@ -113,62 +121,67 @@ impl<B: Blob> Read<B> {
...
-        );
+    /// Returns an error if not enough bytes are available.
+    pub async fn read(&mut self, len: usize) -> Result<IoBufs, Error> {
+        if len == 0 {
+            return Ok(IoBufs::default());
+        }
 
         // Quick check if we have enough bytes total before attempting reads
-        if (self.buffer_remaining() + self.blob_remaining() as usize) < size {
+        if (self.buffer_remaining() + self.blob_remaining() as usize) < len {
```
clabby: This double counts the remaining bytes in the buffer I think, since `blob_remaining = blob_size - (blob_position + buffer_position)`. Will see if I can produce, though I think this could cause us to consume bytes from the in-memory buffer unexpectedly.

runtime/src/utils/buffer/write.rs
clabby: Not really a concern of this PR though.

runtime/src/utils/buffer/read.rs
clabby: The final assertion fails:

```rust
#[test_traced]
fn test_read_error_does_not_advance_cursor() {
        let data = b"AB";
        let (blob, size) = context.open("partition", b"read_err_cursor").await.unwrap();
        assert_eq!(size, 0);
        blob.write_at(0, data).await.unwrap();

        let mut reader = Read::from_pooler(&context, blob, data.len() as u64, NZUsize!(8));

        let first = reader.read(1).await.unwrap().coalesce();
        assert_eq!(first.as_ref(), b"A");
        let pos_before = reader.position();

        let err = reader.read(2).await.unwrap_err();
        assert!(matches!(err, Error::BlobInsufficientLength));
        assert_eq!(
            reader.position(),
            pos_before,
            "cursor advanced on error; read should be all-or-nothing"
        );
    });
}
```

clabby:
This looks good overall - https://github.com/commonwarexyz/monorepo/pull/3185#discussion_r2854661040 was the only thing I'm worried about from a correctness standpoint

runtime/src/utils/buffer/read.rs
andresilva: Good catch! Fixed and added your test as regression test.

runtime/src/utils/buffer/write.rs
andresilva: Yes, in theory we should be able to lock different sections of the buffer, but might not be worth the complexity.

runtime/src/utils/buffer/benches/append.rs
```diff
@@ -16,12 +16,13 @@ pub fn bench(c: &mut Criterion) {
                 let executor = deterministic::Runner::default();
                 executor.start(|ctx| async move {
                     let cache_ref = CacheRef::from_pooler(&ctx, PAGE_SIZE, NZUsize!(CACHE_SIZE));
+                    let logical_page_size = cache_ref.logical_page_size() as usize;
```
patrick-ogrady, line 20: should we be returning `usize` for logical_page_size?

patrick-ogrady: I suppose we accept u16 (IIRC), so that may be a bit weird.

runtime/src/utils/buffer/paged/cache.rs
```diff
@@ -103,15 +176,32 @@ pub struct CacheRef {
...
+    ///
+    /// Capacity bounds the number of resident slots in this cache. Read results may outlive
+    /// eviction and retain memory until dropped by readers.
+    pub fn new(pool: BufferPool, physical_page_size: NonZeroU16, capacity: NonZeroUsize) -> Self {
+        let physical_page_size = physical_page_size.get() as u64;
+        assert!(
+            physical_page_size > CHECKSUM_SIZE,
+            "physical page size must be larger than checksum record"
+        );
+        let logical_page_size =
```
patrick-ogrady, line 209: If true, this is a blocking change.

```
• The patch changes CacheRef page-size semantics in a way that breaks compatibility with previously written page-formatted blobs. Without an
  explicit migration strategy, upgrades can misread valid data as corrupt and truncate persisted state.

  Review comment:

  - [P1] Preserve logical page-size semantics in cache construction — /Users/patrickogrady/code/monorepo-pr-3185/runtime/src/utils/buffer/paged/
    cache.rs:187-194
    CacheRef::new now treats the input as a physical page size and derives the logical size by subtracting CHECKSUM_SIZE, which silently changes
    on-disk layout for existing callers that pass logical sizes (the previous contract). On upgrade, previously persisted blobs are parsed with
    the wrong page boundaries/CRC offsets, so recovery paths like Append::new can classify valid data as invalid and truncate it. This is a
    backward-incompatible storage-format change for existing deployments unless every caller migrates its configured page size.
```

storage/conformance.toml
andresilva: I overlooked the significance of this.

runtime/src/utils/buffer/paged/cache.rs
andresilva: Indeed, I overlooked the significance of this change. I guess we can keep the existing constructor `new/from_pooler` using logical page size (and probably deprecate them?) and introduce a new constructor that takes physical page size instead? The latter should be preferred long-term since we want our page sizes to map directly to the underlying OS page size and also to the buffer pool capacities.

andresilva: I added constructors for logical/physical page sizes and added documentation on this.

examples/log/src/main.rs
```diff
@@ -224,7 +224,7 @@ fn main() {
             activity_timeout: ViewDelta::new(10),
             skip_timeout: ViewDelta::new(5),
             fetch_concurrent: 32,
-            page_cache: CacheRef::from_pooler(&context, NZU16!(16_384), NZUsize!(10_000)),
+            page_cache: CacheRef::from_pooler_physical(&context, NZU16!(16_384), NZUsize!(10_000)),
```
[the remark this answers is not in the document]
andresilva: This is intended.

examples/bridge/src/bin/validator.rs
```diff
@@ -260,7 +260,11 @@ fn main() {
                 activity_timeout: ViewDelta::new(10),
                 skip_timeout: ViewDelta::new(5),
                 fetch_concurrent: 32,
-                page_cache: CacheRef::from_pooler(&context, NZU16!(16_384), NZUsize!(10_000)),
+                page_cache: CacheRef::from_pooler_physical(
+                    &context,
+                    NZU16!(16_384),
```
patrick-ogrady, line 265: Is there an easy way to detect your physical page size (first use) and then place here? It seems easy to configure a sub-optimal page size?

runtime/src/utils/buffer/benches/read.rs
```diff
@@ -1,15 +1,14 @@
...
-use commonware_runtime::{buffer::paged::CacheRef, deterministic, Blob as _, Runner as _};
+use super::{create_append, destroy_append, CACHE_SIZE, PAGE_SIZE};
+use commonware_runtime::{buffer::paged::CacheRef, deterministic, Runner as _};
 use commonware_utils::NZUsize;
 use criterion::Criterion;
 use rand::{rngs::StdRng, Rng, SeedableRng};
 use std::time::Instant;
 
-// Use cache size pages so all data fits in cache (testing buffer performance, not disk).
+// Use cache size logical pages so all data fits in cache (testing buffer performance, not disk).
```
patrick-ogrady, line 10: I think this comment could be improved?

patrick-ogrady: and/or moved inline near the total_size setup?

runtime/src/utils/buffer/paged/read.rs
```diff
@@ -60,7 +60,7 @@ impl<B: Blob> PageReader<B> {
 
         Self {
             blob,
-            page_size,
+            physical_page_size: page_size,
```
patrick-ogrady, line 63: nit: call variable `physical_page_size`?

runtime/src/utils/buffer/paged/read.rs
```diff
@@ -199,14 +200,21 @@ impl ReplayBuf {
...
+        // If buffers is empty, this is the first fill after a seek. `seek_to` sets
+        // `offset_in_page` to the in-page seek offset, and `chunk` starts from that offset in the
+        // first pushed page. Subtract those skipped bytes from `remaining` only on this first push.
         let skip = if self.buffers.is_empty() {
             self.offset_in_page
         } else {
             0
         };
         self.buffers.push_back(state);
+        if logical_bytes < skip {
```
patrick-ogrady, line 212: Is this an issue? Why wasn't this around before?

runtime/src/utils/buffer/read.rs
```diff
@@ -113,62 +121,67 @@ impl<B: Blob> Read<B> {
...
+            return Ok(IoBufs::default());
+        }
 
         // Quick check against total remaining bytes at current position.
-        if self.blob_remaining() < size as u64 {
+        if self.blob_remaining() < len as u64 {
             return Err(Error::BlobInsufficientLength);
         }
 
+        let mut out = IoBufs::default();
```
patrick-ogrady, line 161: move under the `// Read until we have enough bytes` comment

runtime/src/utils/buffer/paged/append.rs
patrick-ogrady, line 1: LLM feedback:

```
• The flush path now assumes all full pages were cached, even though cache insertion is explicitly best-effort. In saturation/rewrite scenarios
  this can expose stale cached data, so the patch has a correctness issue.

  Review comment:

  - [P1] Drain only pages successfully inserted into cache — /Users/patrickogrady/code/monorepo-pr-3185/runtime/src/utils/buffer/paged/
    append.rs:399-411
    CacheRef::cache is best-effort and can stop early when no slot is reservable, but this code still drops all full pages from the tip via
    pages_to_cache * logical_page_size. Under cache-saturation conditions, any pages after the first failed insert are removed from tip without
    guaranteed cache refresh; if those keys already have cached entries (for example after rewrite paths like resize/rewind), reads can return
    stale cached bytes until eviction. Keep uncached pages in tip (or have cache() report inserted pages) before draining.
    
```

storage/conformance.toml
patrick-ogrady: All the more reason to https://github.com/commonwarexyz/monorepo/issues/3295

patrick-ogrady: When we do this, I now require a commit (can be on a forked branch) that proves these are unchanged if the values are kept the same)?

storage/src/ordinal/storage.rs
```diff
@@ -197,14 +195,11 @@ impl<E: BufferPooler + Storage + Metrics + Clock, V: CodecFixed<Cfg = ()>> Ordin
 
                 // Attempt to read record at offset
                 replay_blob.seek_to(offset)?;
-                let mut record_buf = vec![0u8; Record::<V>::SIZE];
-                replay_blob
-                    .read_exact(&mut record_buf, Record::<V>::SIZE)
-                    .await?;
+                let mut record_buf = replay_blob.read(Record::<V>::SIZE).await?;
```
patrick-ogrady, line 198: ❤️

storage/src/ordinal/storage.rs
```diff
@@ -225,6 +220,12 @@ impl<E: BufferPooler + Storage + Metrics + Clock, V: CodecFixed<Cfg = ()>> Ordin
             "rebuilt intervals"
         );
```
patrick-ogrady, line 222: add inline comment

runtime/src/utils/buffer/paged/cache.rs
```diff
@@ -1,12 +1,57 @@
-//! A page cache for caching _logical_ pages of [Blob] data in memory. The cache is unaware of the
-//! physical page format used by the blob, which is left to the blob implementation.
-
-use super::get_page_from_blob;
-use crate::{Blob, BufferPool, BufferPooler, Error, IoBuf, IoBufMut};
+//! A page cache for serving _logical_ page reads from [Blob] data in memory.
+//!
+//! # Logical vs Physical Page Sizes
+//!
+//! [CacheRef::new_physical] is configured with a physical page size (logical bytes plus CRC
```
patrick-ogrady, line 5: I wish this was just the default but I think this is the right approach for now.

runtime/src/utils/buffer/paged/cache.rs
```diff
@@ -101,35 +186,145 @@ pub struct CacheRef {
...
+        pool: BufferPool,
+        physical_page_size: NonZeroU16,
+        capacity: NonZeroUsize,
+    ) -> Self {
+        let physical_page_size = physical_page_size.get() as u64;
+        assert!(
+            physical_page_size > CHECKSUM_SIZE,
+            "physical page size must be larger than checksum record"
+        );
+        let logical_page_size =
```
patrick-ogrady, line 209: for consistency with new_logical, `checked_sub()`?

runtime/src/utils/buffer/paged/cache.rs
```diff
@@ -146,387 +341,1024 @@ impl CacheRef {
...
         &self,
         blob_id: u64,
-        mut buf: &mut [u8],
         mut logical_offset: u64,
-    ) -> usize {
-        let original_len = buf.len();
+        len: usize,
+    ) -> (IoBufs, usize) {
+        let mut remaining = len;
+        let mut out = IoBufs::default();
```
patrick-ogrady, line 359: Shouldn't we be pulling this from `buffer_pool`?

runtime/src/utils/buffer/paged/cache.rs
```diff
@@ -146,387 +341,1024 @@ impl CacheRef {
...
     pub(super) async fn read<B: Blob>(
         &self,
         blob: &B,
         blob_id: u64,
-        mut buf: &mut [u8],
-        mut offset: u64,
+        logical_offset: u64,
+        len: usize,
+    ) -> Result<IoBufs, Error> {
+        let mut result = IoBufs::default();
```
patrick-ogrady, line 384: Shouldn't we be pulling this from `buffer_pool`?

runtime/src/utils/buffer/paged/cache.rs
```diff
@@ -146,387 +341,1024 @@ impl CacheRef {
...
-                    let shareable = future.boxed().shared();
-                    v.insert(shareable.clone());
-
-                    (shareable, true)
+            // Check whether there's an on-going fetch for this key.
+            if let Some((slot, state)) = cache.fetch_for_key(key) {
+                if let Some(stale) = state.future.peek().cloned() {
+                    // Cleanup stale resolved fetch state left by a cancelled first fetcher.
+                    if state.try_claim_cleanup() {
+                        cache.finish_fetch_if_current(slot, key, &state, stale);
```
patrick-ogrady, line 494: This is the fix?

runtime/src/utils/buffer/paged/cache.rs
```diff
@@ -146,387 +341,1024 @@ impl CacheRef {
...
-        let (page_num, offset_in_page) = Cache::offset_to_page(self.page_size, offset);
+        let (page_num, offset_in_page) = Cache::offset_to_page(self.logical_page_size, offset);
         let offset_in_page = offset_in_page as usize;
         trace!(page_num, blob_id, "page fault");
 
-        // Create or clone a future that retrieves the desired page from the underlying blob. This
-        // requires a write lock on the page cache since we may need to modify `page_fetches` if
-        // this is the first fetcher.
-        let (fetch_future, is_first_fetcher) = {
+        let key = (blob_id, page_num);
```
patrick-ogrady, line 464: move under comment block

runtime/src/utils/buffer/paged/cache.rs
```diff
@@ -146,387 +341,1024 @@ impl CacheRef {
...
+
+            // Nobody is currently fetching this page. Create a new shared future and reserve this
+            // slot for it.
+            let fetch_buf = cache.take_slot_buffer(slot);
+            let state = Arc::new(PageFetchState::new(self.make_fetch_future(
+                blob.clone(),
+                page_num,
+                fetch_buf,
+            )));
+            // Publish `(key -> slot)` + Reserved state atomically under this lock so followers can
```
patrick-ogrady, line 524: add \n

runtime/src/utils/buffer/paged/cache.rs
patrick-ogrady, line 1: LLM feedback:

```
  1. [P3] Logical constructors regress max supported logical page size
     runtime/src/utils/buffer/paged/cache.rs:245 converts logical_page_size + CHECKSUM_SIZE into u16 and panics on overflow.
     That makes CacheRef::new_logical / from_pooler_logical reject previously valid inputs like NonZeroU16::new(65535), even though old
     CacheRef::new behavior accepted the full logical NonZeroU16 range and the docs claim these constructors preserve old behavior.
     
```

runtime/src/utils/buffer/paged/cache.rs
```diff
@@ -146,387 +341,1024 @@ impl CacheRef {
...
-    /// didn't fill a whole page. `offset` must be page aligned.
+    /// Cache full logical pages from a contiguous logical-byte prefix.
+    ///
+    /// `logical_pages` is interpreted as a contiguous sequence of exactly `page_count` logical
+    /// pages starting at `offset`. `offset` must be logical-page aligned.
+    ///
+    /// This method copies bytes into cache slots and is best effort: if insertion fails (for
+    /// example, all slots are currently fetching and no slot is reservable), remaining pages are
+    /// dropped after logging an error.
+    /// This does not affect correctness; it only reduces cache hit rate.
```
patrick-ogrady, line 623: This is because we still use the data just don't cache it? Do we have a test for this edge case?

runtime/src/utils/buffer/paged/cache.rs
```diff
@@ -146,387 +341,1024 @@ impl CacheRef {
...
+        let slot = chosen?;
+        self.evict_slot(slot);
+        Some(slot)
+    }
+
+    /// Reclaim a reserved slot if its fetch is already resolved or orphaned.
+    ///
+    /// A slot is considered orphaned when the cache holds the only `Arc<PageFetchState>`,
+    /// which means no task can still await/finish that fetch.
+    fn reclaim_reserved_slot(&mut self, slot: usize) {
```
patrick-ogrady, line 884: Should we break some of this fetching logic into a standalone file?

patrick-ogrady (commented):
This is complex.

patrick-ogrady:
A few more notes:
```
• 1. [P1] Flush drops more tip bytes than were actually cached (stale-read risk under cache saturation).
     In runtime/src/utils/buffer/paged/append.rs:399, bytes_to_drain is computed from all full tip pages, then drained unconditionally at
     append.rs:410. But cache insertion is explicitly best-effort in runtime/src/utils/buffer/paged/cache.rs:620 and can stop early at
     cache.rs:652. If insertion fails mid-batch, later drained pages may still have old cache entries and reads can serve stale bytes.
  2. [P3] CacheRef::cache now panics on page-number overflow instead of degrading gracefully.
     At runtime/src/utils/buffer/paged/cache.rs:664, checked_add(1).expect(...) can panic for extreme high-offset multi-page cache attempts.
     Previous behavior stopped caching remaining pages without panicking. This is a crash-on-edge-input regression.
  3. [P3] Buffer::slice validates normalized bounds but slices using the original range.
     In runtime/src/utils/buffer/tip.rs:111 through tip.rs:124, it computes start/end against logical self.len, then returns
     self.data.slice(range). For unbounded-end ranges (e.g. .. / start..), this can include bytes beyond logical len if backing view is longer.
     It’s a latent stale-data exposure bug in the helper itself.
     
```

examples/bridge/src/bin/validator.rs
andresilva: Yes, the buffer pool already does this for unix (I didn't implement it for windows and just default to 4K iirc). I think the `CacheRef` should somehow get this information from the pool and just use that, this makes sense because we want alignment with OS page size (which the buffer pool already does), and alignment between page cache and buffer pool.

andresilva: Only worry is what if this changes for some reason (it shouldn't), but page size is tied to format and it would break existing data layout.

andresilva:
Moving this to draft because I'll break it up into multiple PRs.

PR #3191, opened 2026-02-20
[cryptography/bls12381/dkg] Improve Wording around "Share Reveals"

patrick-ogrady:


cronokirby (approved):
This is a lot clearer and more useful than what we have now.

PR #3192, opened 2026-02-20
[cryptography] More robust player resumption in DKG

### LINKED ISSUE #2563: [cryptography/dkg] If missing acknowledged shares, still return output (but no shares)
If a node is lost, it is possible it will not be able to lookup shares it historically acknowledged. In `main`, we fail ungracefully in this case rather than just returning the output (with no share).

TL;DR we should be able to act as an observer if we are missing our acks rather than having to halt.

cronokirby:
Closes #2563.

This makes it so that `finalize` on Player returns a new error if we can detect that the player's state is corrupted (missing shares it should otherwise have, based on public acks), avoiding panics. We also add a new `resume` method, replacing the `new` + `replay` pattern, to allow detecting earlier when messages are missing. Hopefully this allows operators to catch mistakes before finalization, when it's too late.

Also added tests which perturb players by crashing them, and modified the reshare example.

### DIFF
--- a/cryptography/src/bls12381/dkg.rs
+++ b/cryptography/src/bls12381/dkg.rs
@@ -29,15 +29,15 @@
 //!
 //! * [`Info`]: Configuration for a DKG/Reshare round, containing the dealers, players, and optional previous output
 //! * [`Output`]: The public result of a successful DKG round, containing the public polynomial and player list
-//! * [`Share`]: A player's private share of the distributed key (from `primitives::group`)
+//! * [`Share`]: A player's final private share of the distributed key (from `primitives::group`)
 //! * [`Dealer`]: State machine for a dealer participating in the protocol
-//! * [`Player`]: State machine for a player receiving shares
+//! * [`Player`]: State machine for a player receiving dealings
 //! * [`SignedDealerLog`]: A dealer's signed transcript of their interactions with players
 //!
 //! ## Message Types
 //!
 //! * [`DealerPubMsg`]: Public commitment polynomial sent from dealer to all players
-//! * [`DealerPrivMsg`]: Private share sent from dealer to a specific player
+//! * [`DealerPrivMsg`]: Private dealing sent from dealer to a specific player
 //! * [`PlayerAck`]: Acknowledgement sent from player back to dealer
 //! * [`DealerLog`]: Complete log of a dealer's interactions (commitments and acks/reveals)
 //!
@@ -49,7 +49,7 @@
 //! - Round number (should increment sequentially, including for failed rounds)
 //! - Optional previous [`Output`] (for resharing)
 //! - List of dealers (must be >= quorum of previous round if resharing)
-//! - List of players who will receive shares
+//! - List of players who will receive dealings
 //!
 //! ### Step 2: Dealer Phase
 //!
@@ -66,7 +66,7 @@
 //! Each player creates a [`Player`] instance via [`Player::new`], then for each dealer message:
 //! - Call [`Player::dealer_message`] with the [`DealerPubMsg`] and [`DealerPrivMsg`]
 //! - If valid, this returns a [`PlayerAck`] containing a signature over `(dealer, commitment)`
-//! - The player verifies that the private share matches the public commitment evaluation
+//! - The player verifies that the private dealing matches the public commitment evaluation
 //!
 //! ### Step 4: Dealer Collection
 //!
@@ -84,7 +84,7 @@
 //!
 //! The [`Output`] contains:
 //! - The final public polynomial (sum of dealer polynomials for DKG, interpolation for reshare),
-//! - The list of dealers who distributed shares,
+//! - The list of dealers who distributed dealings,
 //! - The list of players who received shares,
 //! - The set of players whose shares may have been revealed,
 //! - A digest of the round's [`Info`] (including the counter, and the list of dealers and players).
@@ -332,7 +332,7 @@ const SIG_LOG: &[u8] = b"log";
 /// The only error which can happen through no fault of your own is
 /// [`Error::DkgFailed`].
 ///
-/// [`Error::MissingDealing`] happens through mistakes or faults when the state
+/// [`Error::MissingPlayerDealing`] happens through mistakes or faults when the state
 /// of a player is restored after a crash.
 ///
 /// The other errors are due to issues with configuration.
@@ -357,8 +357,8 @@ pub enum Error {
     /// the code in this module is used in a stateful way, restoring the
     /// state of the player from saved information. If this state is corrupted
     /// on disk, or missing, then this error can happen.
-    #[error("missing player's share")]
-    MissingDealing,
+    #[error("missing player's dealing")]
+    MissingPlayerDealing,
 }
 
 /// The output of a successful DKG.
@@ -1526,7 +1526,7 @@ impl<V: Variant, S: Signer> Player<V, S> {
     ///
     /// All messages the player should have received must be passed into this method,
     /// and if any messages which should be present based on this player's actions
-    /// in the log are missing, this method will return [`Error::MissingDealing`].
+    /// in the log are missing, this method will return [`Error::MissingPlayerDealing`].
     ///
     /// For example, if a particular private message containing a share is not
     /// present in `msgs`, but we've already acknowledged it, and this has been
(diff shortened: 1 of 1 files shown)


cryptography/src/bls12381/dkg.rs
```diff
@@ -1509,18 +1607,31 @@ impl<V: Variant, S: Signer> Player<V, S> {
...
     ) -> Result<(Output<V, S::PublicKey>, Share), Error> {
         let selected = select::<V, S::PublicKey, M>(&self.info, logs)?;
+        // If there's a log that contains an ack of ours, but no corresponding view,
+        // then we're corrupted.
+        if selected
+            .iter_pairs()
+            .any(|(d, l)| l.player_acked(&self.me_pub) && !self.view.contains_key(d))
+        {
+            return Err(Error::PlayerCorrupted);
+        }
```
patrick-ogrady, line 1634: add a `\n` after this

cryptography/src/bls12381/dkg.rs
```diff
@@ -1477,6 +1522,59 @@ impl<V: Variant, S: Signer> Player<V, S> {
...
+    ///
+    /// For example, if a particular private message containing a share is not
+    /// present in `msgs`, but we've already acknowledged it, and this has been
+    /// included in a public log, then this method will fail.
+    ///
+    /// This method cannot catch all cases where state has been corrupted. In
+    /// particular, if a dealer has not posted their log publicly yet, but has
+    /// already received an ack, then this method cannot help in that case,
+    /// but the issue still remains.
+    #[allow(clippy::type_complexity)]
```
patrick-ogrady, line 1541: We should clarify that logs must be pre-verified (or we should verify again)? If not, a byzantine dealer can force us to think we acked?

```
diff --git a/cryptography/src/bls12381/dkg.rs b/cryptography/src/bls12381/dkg.rs
index 9d582e22f..2f571e651 100644
--- a/cryptography/src/bls12381/dkg.rs
+++ b/cryptography/src/bls12381/dkg.rs
@@ -79,6 +79,7 @@
 //!
 //! With collected [`SignedDealerLog`]s:
 //! - Call [`SignedDealerLog::check`] to verify and extract [`DealerLog`]s
+//! - Persist and replay only verified [`DealerLog`]s (from `check`) for recovery
 //! - Players call [`Player::finalize`] with all logs to compute their [`Share`] and [`Output`]
 //! - Observers call [`observe`] with all logs to compute just the [`Output`]
 //!
@@ -1528,6 +1529,10 @@ impl<V: Variant, S: Signer> Player<V, S> {
     /// with the appropriate messages, but includes extra safeguards to detect
     /// missing / corrupted state.
     ///
+    /// The `logs` input is expected to contain only verified logs, such as the
+    /// output of [`SignedDealerLog::check`]. This method does not re-validate
+    /// signatures in `logs`.
+    ///
     /// All messages the player should have received must be passed into this method,
     /// and if any messages which should be present based on this player's actions
     /// in the log are missing, this method will return [`Error::PlayerCorrupted`].
```

cronokirby: Yeah, doesn't hurt to specify this, in practice I imagine that given that they have to be checked at least once, people won't run into this. We could also just take signed logs here, but it's more convenient to take the map in the same form as `finalize`, because this module already encourages people to have a pattern where they build up that map as blocks pass, so they'll have it around.

cryptography/src/bls12381/dkg.rs
```diff
@@ -1477,6 +1522,64 @@ impl<V: Variant, S: Signer> Player<V, S> {
         })
     }
 
+    /// Resume a [`Player`], given some existing public state.
+    ///
+    /// This is equivalent to calling [`Self::new`] and then [`Self::dealer_message`]
+    /// with the appropriate messages, but includes extra safeguards to detect
+    /// missing / corrupted state.
+    ///
+    /// It's imperative that the `logs` passed in have been verified. This is done
```
patrick-ogrady, line 1522: We don't actually do a "full check" of items until `select` ... so just verifying dealer signature during `check` could lead to us thinking we ack'd something that a dealer posted (that is invalid)?

patrick-ogrady: See:

```
+    #[test]
+    fn signed_log_can_pass_check_but_fail_select() -> Result<(), Error> {
+        let sk = ed25519::PrivateKey::from_seed(7);
+        let pk = sk.public_key();
+        let info = Info::<MinPk, _>::new::<N3f1>(
+            b"_COMMONWARE_CRYPTOGRAPHY_BLS12381_DKG_TEST",
+            0,
+            None,
+            Default::default(),
+            vec![pk.clone()].try_into().unwrap(),
+            vec![pk.clone()].try_into().unwrap(),
+        )?;
+
+        let (mut dealer, pub_msg, priv_msgs) =
+            Dealer::start::<N3f1>(&mut test_rng(), info.clone(), sk.clone(), None)?;
+        let mut player = Player::new(info.clone(), sk.clone())?;
+        let ack = player
+            .dealer_message::<N3f1>(pk.clone(), pub_msg, priv_msgs[0].1.clone())
+            .expect("player should ack valid dealer message");
+        dealer.receive_player_ack(pk.clone(), ack)?;
+        let (_, mut log) = dealer
+            .finalize::<N3f1>()
+            .check(&info)
+            .expect("valid signed log should verify");
+
+        match &mut log.results {
+            DealerResult::TooManyReveals => panic!("expected ack log"),
+            DealerResult::Ok(results) => {
+                *results
+                    .get_value_mut(&pk)
+                    .expect("single player should appear in log") = AckOrReveal::Ack(PlayerAck {
+                    // Sign the wrong transcript on purpose: this passes SignedDealerLog::check
+                    // once the dealer re-signs the log, but should fail select ack validation.
+                    sig: transcript_for_round(&info).sign(&sk),
+                });
+            }
+        }
+
+        let signed = SignedDealerLog::sign(&sk, &info, log);
+        let (dealer_pk, checked_log) = signed
+            .check(&info)
+            .expect("dealer signature over malformed log should still verify");
+        assert_eq!(dealer_pk, pk);
+
+        let logs = BTreeMap::from([(dealer_pk, checked_log)]);
+        assert!(matches!(
+            select::<_, _, N3f1>(&info, logs),
+            Err(Error::DkgFailed)
+        ));
+        Ok(())
+    }
+
```

patrick-ogrady: https://github.com/commonwarexyz/monorepo/blob/7e59ed09321096bfba373cc9b2a669506a04db4e/cryptography/src/bls12381/dkg.rs#L1371-L1373

cronokirby: Right, we need to check our signature here!

cryptography/src/bls12381/dkg.rs
```diff
@@ -1477,6 +1522,70 @@ impl<V: Variant, S: Signer> Player<V, S> {
...
+    #[allow(clippy::type_complexity)]
+    pub fn resume<M: Faults>(
+        info: Info<V, S::PublicKey>,
+        me: S,
+        logs: &BTreeMap<S::PublicKey, DealerLog<V, S::PublicKey>>,
+        msgs: impl IntoIterator<Item = (S::PublicKey, DealerPubMsg<V>, DealerPrivMsg)>,
+    ) -> Result<(Self, BTreeMap<S::PublicKey, PlayerAck<S::PublicKey>>), Error> {
+        let mut this = Self::new(info, me)?;
+        let mut acks = BTreeMap::new();
+        // We want to check our own behavior. Other things get resolved by `select`
```
patrick-ogrady: cleanup these docs

cryptography/src/bls12381/dkg.rs
```diff
@@ -333,6 +350,15 @@ pub enum Error {
...
     DkgFailed,
+    /// The player's state is missing a dealing it should have.
+    ///
+    /// This error is emitted when the player is missing dealings that it should
+    /// otherwise have based on the flow of the protocol. This can only happen if
+    /// the code in this module is used in a stateful way, restoring the
+    /// state of the player from saved information. If this state is corrupted
+    /// on disk, or missing, then this error can happen.
+    #[error("missing player's dealing")]
+    MissingPlayerDealing,
```
patrick-ogrady, line 361: Opted to be specific rather than just "corruption" (which would imply we need to check a lot more)

Issue #3190, opened 2026-02-20
[storage/qmdb] Support variable length keys


roberto-bayardo:
All qmdb variants have a fixed-size constraint on keys. This could be relaxed for the variable-size-record variants.

PR #3195, opened 2026-02-21
[storage/qmdb/any] any+unordered qmdb variants with variable length keys

roberto-bayardo:
Adds two any/unordered qmdb variants with variable length keys, one with fixed size values and one for variable sized values.  Also removes some redundant tests already covered by the generic ones.

Context: https://github.com/commonwarexyz/monorepo/issues/3190


storage/src/qmdb/any/unordered/varkey.rs
```diff
@@ -0,0 +1,441 @@
+//! An authenticated database that provides succinct proofs of _any_ value ever associated
+//! with a variable-length key, where values can also have varying sizes.
+//!
+//! Unlike other QMDB variants that require fixed-size keys (`K: Array`), this variant uses
+//! `Vec<u8>` as the key type, supporting arbitrary-length byte sequences as keys.
```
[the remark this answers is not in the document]
roberto-bayardo: thanks, fixed.

storage/src/qmdb/any/operation/update/unordered.rs
```diff
@@ -79,3 +86,40 @@ impl<K: Array, V: VariableValue> Read for Update<K, VariableEncoding<V>> {
         Ok(Self(key, value))
     }
 }
+
+// --- Variable key + Variable value ---
+
+impl<K, V> EncodeSize for Update<K, VarKeyEncoding<V>>
```
danlaine: Having the value be `VarKeyEncoding<V>` feels a bit off to me.

What if we parameterized `Update` on a trait `Encoding` that has `Key` and `Value` associate types that specify whether the key/value is fixed/variable length, or something like that?

roberto-bayardo: yeah it is a little stinky, let me see what I can do.

storage/src/qmdb/any/mod.rs
```diff
@@ -870,14 +897,22 @@ pub(crate) mod test {
...
-    // Defines all 12 variants in one place. Calls $cb!($($args)*, $label, $type, $config) for each.
-    macro_rules! with_all_variants {
+    // Defines all 14 variants in two groups. Split into halves so the combined async state
+    // machine of each group fits on the stack.
+    macro_rules! with_variants_a {
         ($cb:ident!($($args:tt)*)) => {
             $cb!($($args)*, "uf", UnorderedFixed, fixed_db_config);
             $cb!($($args)*, "uv", UnorderedVariable, variable_db_config);
+            $cb!($($args)*, "uvkv", UnorderedVariableKeyVariable, varkey_db_config);
+            $cb!($($args)*, "uvkf", UnorderedVariableKeyFixed, variable_db_config);
```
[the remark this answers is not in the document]
roberto-bayardo: This isn't actually a bug. The Db::init call requires VariableConfig<T, <Operation<K, V> as Read>::Cfg>, so the compiler enforces that the config's codec type matches the operation's Read::Cfg. For VariableKey<Digest, Digest>, Cfg = <Digest as Read>::Cfg = (), which is exactly what variable_db_config provides. If the key type were changed to something with Cfg != (), the code would fail to compile — there's no silent mismatch possible.

storage/src/qmdb/any/operation/update/unordered.rs
roberto-bayardo: ok, added an Encoding trait for update which propagated pretty far, but it does look better (IMO).

roberto-bayardo:
Closing in favor of https://github.com/commonwarexyz/monorepo/pull/3233

PR #3202, opened 2026-02-22
[cryptography] Use `sha2` intrinsics on aarch64

clabby:
## Overview

Enables use of the AArch64 `sha2` extension when it is available. The `sha2` crate notes that if the extension is not available, the `soft` backend will be used, so this is safe to enable on all compilations targeting `aarch64`.

On an M4 Max:

| Benchmark | New time (95% CI) | Change vs baseline (95% CI) | Approx speedup (from midpoint) |
|---|---:|---:|---:|
| sha256::hash_message/msg_len=256 | 97.999 ns – 99.483 ns | −82.763% – −82.532% | ~5.76x |
| sha256::hash_message/msg_len=4096 | 1.5200 µs – 1.5412 µs | −79.328% – −79.056% | ~4.80x |
| sha256::hash_message/msg_len=65536 | 23.475 µs – 23.687 µs | −79.386% – −79.100% | ~4.81x |
| sha256::hash_message/msg_len=524288 | 192.89 µs – 195.41 µs | −79.101% – −78.831% | ~4.75x |
| sha256::hash_message/msg_len=1048576 | 374.64 µs – 377.91 µs | −79.441% – −79.134% | ~4.83x |
| sha256::hash_message/msg_len=16777216 | 6.1643 ms – 6.2393 ms | −79.341% – −79.071% | ~4.81x |

---

`blake3` is still more optimal for large inputs on ARM, but only with the `parallel` feature enabled. Forcing
the use of intrinsics makes `sha2` the more compelling choice for small inputs or in cases where multi-core is not
an option:

image

### DIFF
--- a/cryptography/Cargo.toml
+++ b/cryptography/Cargo.toml
@@ -71,8 +71,8 @@ crate-type = ["rlib", "cdylib"]
 
 [features]
 default = [ "std" ]
-parallel = [ "blake3/rayon", "commonware-parallel/std", "std" ]
-asm = [ "sha2/asm" ]
+blake3-parallel = [ "blake3/rayon", "std" ]
+sha2-asm = [ "sha2/asm" ]
 mocks = [ "std" ]
 arbitrary = [
 	"commonware-codec/arbitrary",
--- a/cryptography/src/blake3/mod.rs
+++ b/cryptography/src/blake3/mod.rs
@@ -40,7 +40,7 @@ const DIGEST_LENGTH: usize = blake3::OUT_LEN;
 
 /// BLAKE3 hasher.
 #[cfg_attr(
-    feature = "parallel",
+    feature = "blake3-parallel",
     doc = "When the input message is larger than 128KiB, `rayon` is used to parallelize hashing."
 )]
 #[derive(Debug, Default)]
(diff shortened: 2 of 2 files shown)


cryptography/Cargo.toml
```diff
@@ -72,6 +72,7 @@ crate-type = ["rlib", "cdylib"]
 [features]
 default = [ "std" ]
 parallel = [ "blake3/rayon", "commonware-parallel/std", "std" ]
+asm = [ "sha2/asm" ]
```
patrick-ogrady: Is there a reason to not just enable this by default?

patrick-ogrady: Ah, because of:
```
#![no_std]
#[cfg(not(any(target_arch = "x86_64", target_arch = "x86", target_arch = "aarch64")))]
compile_error!("crate can only be used on x86, x86-64 and aarch64 architectures");
```

clabby: Mainly portability. If we enable it by default, consumers cannot opt out.

cryptography/Cargo.toml
```diff
@@ -72,6 +72,7 @@ crate-type = ["rlib", "cdylib"]
 [features]
 default = [ "std" ]
 parallel = [ "blake3/rayon", "commonware-parallel/std", "std" ]
```
patrick-ogrady: I had a change of heart on these features and now think we should scope?

So, something like `blake3-parallel` and `sha2-asm`?

cronokirby: "parallel" enabling parallel on everything seems straightforward to me.

patrick-ogrady: Was DM'ing with @clabby ... the thing that confuses me with this pattern is that there are many things that support parallelism without this enabled? Do we strip all of that down to match this feature?

PR #3206, opened 2026-02-23
[coding] Optimize ReedSolomon

clabby:
## Overview

Adds several optimizations to `reed_solomon`:

- Caches the encoder/decoder to avoid re-building LUTs
- Zero-copy chunking in `encode` using `Bytes::slice`
- Bulk memcopy in `extract_data` to avoid a byte-by-byte allocation
- `CheckedChunk` type, to avoid re-computing the shard digests of checked shards in `decode`

Encode (16 MB payload, 50 chunks)

| Concurrency | Trunk (baseline) | Optimized | Change |
|---|---|---|---|
| conc=1 | 103.38 ms | 103.41 ms | -0.0% |
| conc=4 | 33.07 ms | 32.55 ms | -1.6% |
| conc=8 | 21.75 ms | 21.22 ms | -2.5% |

Decode (16 MB payload, 50 chunks)

| Concurrency | Trunk (baseline) | Optimized | Change |
|---|---|---|---|
| conc=1 | 146.11 ms | 104.15 ms | -28.7% |
| conc=4 | 75.28 ms | 56.44 ms | -25.0% |
| conc=8 | 63.78 ms | 48.82 ms | -23.4% |

---

On a global deployment of `alto` with 4MiB blocks, these changes reduce view latency by 20-25ms on average, relative to our measurements last week.

image

### DIFF
--- a/utils/src/thread_local.rs
+++ b/utils/src/thread_local.rs
@@ -41,6 +41,36 @@ use std::{
     thread::LocalKey,
 };
 
+/// Restores slot state if `take` returns early or unwinds.
+///
+/// While active, drop clears the "held" bit and puts any cached value
+/// back into the TLS slot.
+struct TakeCleanup<T: 'static> {
+    cache: &'static LocalKey<RefCell<(bool, Option<T>)>>,
+    cached: Option<T>,
+    armed: bool,
+}
+
+impl<T: 'static> TakeCleanup<T> {
+    const fn disarm(&mut self) {
+        self.armed = false;
+    }
+}
+
+impl<T: 'static> Drop for TakeCleanup<T> {
+    fn drop(&mut self) {
+        if !self.armed {
+            return;
+        }
+        self.cache.with(|cell| {
+            let mut slot = cell.borrow_mut();
+            debug_assert!(slot.0, "cache expected to be held");
+            slot.0 = false;
+            slot.1 = self.cached.take();
+        });
+    }
+}
+
 /// RAII guard that borrows a value from a thread-local cache and returns it
 /// on drop.
 ///
@@ -73,34 +103,22 @@ impl<T: 'static> Cached<T> {
             slot.0 = true;
             slot.1.take()
         });
-        let value = match cached {
+        let mut cleanup = TakeCleanup {
+            cache,
+            cached,
+            armed: true,
+        };
+        let value = match cleanup.cached.take() {
             Some(mut v) => {
                 if let Err(err) = reset(&mut v) {
-                    // Restore the previous value on reset failure so transient errors
-                    // do not permanently evict the cache entry.
-                    cache.with(|cell| {
-                        let mut slot = cell.borrow_mut();
-                        debug_assert!(slot.0, "cache expected to be held");
-                        slot.0 = false;
-                        slot.1 = Some(v);
-                    });
+                    cleanup.cached = Some(v);
                     return Err(err);
                 }
                 v
             }
-            None => match create() {
-                Ok(v) => v,
-                Err(err) => {
-                    // Return to uninitialized state if creation fails.
-                    cache.with(|cell| {
-                        let mut slot = cell.borrow_mut();
-                        debug_assert!(slot.0, "cache expected to be held");
-                        slot.0 = false;
-                    });
-                    return Err(err);
-                }
-            },
+            None => create()?,
         };
+        cleanup.disarm();
         Ok(Self {
             value: Some(value),
             cache,
@@ -269,5 +287,43 @@ mod tests {
         // Outer guard should have returned its value while unwinding.
         let cached = NESTED_CACHE.with(|cell| cell.borrow().1.clone());
         assert_eq!(cached, Some(vec![1, 10]));
+    }
+
+    thread_local_cache!(static PANIC_CREATE_CACHE: u32);
+
+    #[test]
+    fn test_create_panic_does_not_poison_held_flag() {
+        let result = std::panic::catch_unwind(|| {
+            let _ = Cached::take(
+                &PANIC_CREATE_CACHE,
+                || -> Result<u32, ()> { panic!("create panic") },
+                |_| Ok(()),
+            );
+        });
+        assert!(result.is_err());
+
+        let guard = Cached::take(&PANIC_CREATE_CACHE, || Ok::<_, ()>(7), |_| Ok(())).unwrap();
+        assert_eq!(*guard, 7);
+    }
+
+    thread_local_cache!(static PANIC_RESET_CACHE: u32);
+
+    #[test]
+    fn test_reset_panic_does_not_poison_held_flag() {
+        {
+            let _guard = Cached::take(&PANIC_RESET_CACHE, || Ok::<_, ()>(42), |_| Ok(())).unwrap();
+        }
+
+        let result = std::panic::catch_unwind(|| {
+            let _ = Cached::take(
+                &PANIC_RESET_CACHE,
+                || Ok::<_, ()>(0),
+                |_| -> Result<(), ()> { panic!("reset panic") },
+            );
+        });
+        assert!(result.is_err());
+
+        let guard = Cached::take(&PANIC_RESET_CACHE, || Ok::<_, ()>(9), |_| Ok(())).unwrap();
+        assert_eq!(*guard, 9);
     }
 }


coding/src/reed_solomon.rs
```diff
@@ -5,9 +5,54 @@ use commonware_cryptography::{Digest, Hasher};
...
-use std::{collections::HashSet, marker::PhantomData};
+use std::{cell::RefCell, collections::HashSet, marker::PhantomData};
 use thiserror::Error;
 
+// Thread-local caches for reusing `ReedSolomonEncoder` and `ReedSolomonDecoder`
+// instances across calls. Constructing these objects is expensive because
+// the underlying engine initializes GF lookup tables. The `reset()` method
+// reconfigures the work buffers without rebuilding those tables.
+thread_local! {
+    static CACHED_ENCODER: RefCell<Option<ReedSolomonEncoder>> = const { RefCell::new(None) };
```
patrick-ogrady: I suppose in our use case this will almost always be useful (as the params should stay constant)?

patrick-ogrady: I suppose this would start to do pretty poorly with concurrent use (especially if tokio is flipping around)?

Maybe we should try to clone this instead of taking it? Or create a long-lived encoder/decoder object so we can manage this explicitly?

clabby: The caches are thread local, and because the scheme is synchronous, we never hold the encoder / decoder across an await point. We'll never yield to the scheduler while the cached encoder / decoder is held, so the cache will never be empty when accessed (except upon a thread's first access.)

clabby:
blocked by #3209

coding/src/reed_solomon.rs
cronokirby: Perhaps we should consider creating a "thread local lazy static" in utils to isolate this pattern from the two usages here?

coding/src/reed_solomon.rs
```diff
@@ -148,43 +224,66 @@ fn prepare_data(data: Vec<u8>, k: usize, m: usize) -> Vec<Vec<u8>> {
...
+    let mut remaining = u32::from_be_bytes(prefix) as usize;
+    let mut data = Vec::with_capacity(remaining);
+    let (first_data_shard, mut offset) = (u32::SIZE / shard_len, u32::SIZE % shard_len);
+    for shard in shards[first_data_shard..k].iter() {
+        let read = remaining.min(shard.len() - offset);
+        data.extend_from_slice(&shard[offset..offset + read]);
+        remaining -= read;
+        offset = 0;
+    }
+    data
```
cronokirby: We'll need to add the zero check to this anyways, like #3209, so we should just assemble the data by concatenating the slices, check that the tail is zero, and then truncate.

coding/src/reed_solomon.rs
```diff
@@ -372,7 +462,15 @@ fn decode<H: Hasher, S: Strategy>(
...
     // Extract original data
-    Ok(extract_data(shards, k))
+    let data = extract_data(&shards, k);
+
+    // Return the encoder and decoder
+    drop(encoding);
+    return_encoder(encoder);
+    drop(decoding);
+    return_decoder(decoder);
+
```
cronokirby: Is there a way to avoid this ceremony?

utils/src/thread_local.rs
```diff
@@ -0,0 +1,207 @@
...
+        self.value.as_mut().expect("value taken after drop")
+    }
+}
+
+impl<T: 'static> Drop for Cached<T> {
+    fn drop(&mut self) {
+        if let Some(v) = self.value.take() {
+            self.cache.with(|cell| *cell.borrow_mut() = Some(v));
+        }
+    }
```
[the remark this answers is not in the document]
clabby: This is expected behavior, last drop wins.

utils/src/thread_local.rs
```diff
@@ -0,0 +1,241 @@
+//! RAII guard for thread-local caching of expensive-to-construct objects.
+//!
+//! # Overview
+//!
+//! When an object is expensive to construct but cheap to reset, keeping
```
patrick-ogrady: Do we want to say this is a common pattern for "stateless" functions? Otherwise, something like this would be stored in a struct?

clabby: Good call

coding/src/reed_solomon.rs
```diff
@@ -316,12 +390,12 @@ fn decode<H: Hasher, S: Strategy>(
...
         }
-        if seen.contains(&index) {
+        if !seen.insert(index) {
             return Err(Error::DuplicateIndex(index));
         }
-        seen.insert(index);
 
-        // Add to provided shards
+        // Add to provided shards and retain the checked digest for this index.
+        shard_digests[index as usize] = Some(chunk.digest);
```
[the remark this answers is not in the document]
clabby: This is expected, and is part of the API contract. `CheckedShard`s must be generated locally.

coding/src/reed_solomon.rs
```diff
@@ -204,86 +291,72 @@ fn encode_inner<H: Hasher, S: Strategy>(
...
+
+    // Create zero-copy Bytes views into the original and recovery buffers
+    let originals: Bytes = padded.into();
+    let recoveries: Bytes = recovery_buf.into();
 
     // Build Merkle tree
     let mut builder = Builder::<H>::new(n);
-    let shard_hashes = strategy.map_init_collect_vec(&shards, H::new, |hasher, shard| {
+    let shard_slices: Vec<Bytes> = (0..k)
+        .map(|i| originals.slice(i * shard_len..(i + 1) * shard_len))
```
patrick-ogrady, line 327: :fancy:

coding/src/reed_solomon.rs
clabby: I added a new RAII-style thread local cache, which returns the item on drop.

utils/src/thread_local.rs
```diff
@@ -0,0 +1,267 @@
...
+
+use std::{
+    cell::RefCell,
+    ops::{Deref, DerefMut},
+    thread::LocalKey,
+};
+
+/// RAII guard that borrows a value from a thread-local cache and returns it
+/// on drop.
+pub struct Cached<T: 'static> {
```
patrick-ogrady, line 79: Not sure if this feedback is overkill:
```
  - [P1] Prevent Cached guards from crossing threads — /Users/patrickogrady/code/monorepo-pr-3206/utils/src/thread_local.rs:45-47
    Cached<T> is currently auto-Send, so safe code can move a guard to another thread and drop it there. Drop then writes into the destination
    thread's TLS slot instead of the source slot, leaving the source slot permanently marked held; in debug builds this can panic on
    debug_assert!(slot.0), and in release builds it silently poisons future take calls on the original thread. This should be prevented by making
    the guard thread-affine (for example a !Send marker and/or explicit thread-id check on drop).
    ```

File: utils/src/thread_local.rs [hunk 9]
```diff
@@ -0,0 +1,267 @@
...
+
+impl<T: 'static> DerefMut for Cached<T> {
+    fn deref_mut(&mut self) -> &mut T {
+        self.value.as_mut().expect("value taken after drop")
+    }
+}
+
+impl<T: 'static> Drop for Cached<T> {
+    fn drop(&mut self) {
+        if let Some(v) = self.value.take() {
```
patrick-ogrady, line 146: Probably not worth it:
```
  - [P2] Clear held state when create/reset panics — /Users/patrickogrady/code/monorepo-pr-3206/utils/src/thread_local.rs:65-70
    take marks the slot as held before invoking create/reset, but cleanup only happens on Err; if either closure panics (for example due to an
    internal assertion) and the caller catches unwind, the slot remains stuck in the held state and all later take calls on that thread panic
    with cache already held on this thread. This needs an unwind-safe guard so the held flag is always cleared on panic paths.
    ```

File: utils/src/thread_local.rs [hunk 10]
```diff
@@ -0,0 +1,332 @@
...
+                    cleanup.cached = Some(v);
+                    return Err(err);
+                }
+                v
+            }
+            None => match create() {
+                Ok(v) => v,
+                Err(err) => return Err(err),
+            },
+        };
```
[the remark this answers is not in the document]
clabby: This is fine. We'll just re-initialize it next go. The important part is that we don't retain the held slot.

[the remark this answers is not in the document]
patrick-ogrady: I think this is preferred actually (we don't know the state of the object after a panic tbh)

PR #3214, opened 2026-02-23
[consensus/ordered-broadcast] Batch-verify ack signatures

### LINKED ISSUE #1002: [consensus::ordered-broadcast] Add lazy signature verification

BrendanChou:
Fixes #1002. Previously, `ordered_broadcast` verified each ack signature individually inline as it arrived (`scheme.verify_attestation()`), while the `certificate::Scheme` trait already exposes `verify_attestations()` for efficient batch verification. Simplex already uses batch verification via its `Verifier` struct; this PR brings the same optimization to `ordered_broadcast`.

- Add a `Verifier` that buffers incoming acks and batch-verifies their cryptographic signatures, replacing the previous inline per-ack verification
- Split `validate_ack` into cheap validation (epoch bounds, sender/signer match, height range) and deferred signature verification via the batch verifier
- Self-generated acks bypass the verifier as pre-verified for correct quorum accounting

## Design

The `Verifier` groups pending acks by their signing subject `(Chunk, Epoch)` -- acks sharing the same subject produce identical `AckSubject` values and can be batch-verified together. Readiness is quorum-aware:

- **Batchable schemes** (ed25519, BLS multisig, BLS threshold): verification is deferred until `verified + pending >= quorum` for a given subject
- **Non-batchable schemes** (secp256r1): verification happens eagerly since there is no batching benefit

Batch verification runs in the `on_end` block of the engine's `select_loop!`, so it executes after every select iteration. The verifier's quorum is updated whenever the epoch changes.

### DIFF
--- a/consensus/src/ordered_broadcast/engine.rs
+++ b/consensus/src/ordered_broadcast/engine.rs
@@ -366,6 +366,13 @@ impl<
                     let n = scheme.participants().len() as u32;
                     self.verifier.set_quorum(N3f1::quorum(n));
                 }
+
+                // Prune verifier batches outside the new epoch bounds
+                let (eb_lo, eb_hi) = self.epoch_bounds;
+                self.verifier.prune_epochs(
+                    epoch.saturating_sub(eb_lo),
+                    epoch.saturating_add(eb_hi),
+                );
                 continue;
             },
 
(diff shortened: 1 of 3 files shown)


clabby:

PR #3215, opened 2026-02-23
[consensus/ordered-broadcast] Fix typos and remove stale test lint allows

BrendanChou:
Small cleanups in `ordered_broadcast`:
- Fix `verif` -> `verify` typo in `types.rs`
- Fix double space in `engine.rs` doc comment
- Remove blanket `#[allow(dead_code, unused_imports)]` on the `ack_manager` test module and drop the imports it was hiding

(Previously this PR also reworked the AGENTS.md namespace docs and fixed clippy 1.97 lints; the former was dropped and the latter landed on main in #4213.)

### DIFF
--- a/consensus/src/ordered_broadcast/types.rs
+++ b/consensus/src/ordered_broadcast/types.rs
@@ -507,7 +507,7 @@ pub struct Node<P: PublicKey, S: Scheme, D: Digest> {
 
     /// Information about the parent chunk (previous height)
     ///
-    /// This part is not signed over, but it is used to verif that the previous chunk
+    /// This part is not signed over, but it is used to verify that the previous chunk
     /// in the chain was correctly broadcast. It contains the certificate that proves
     /// a quorum of validators acknowledged the parent.
     ///
--- a/consensus/src/ordered_broadcast/ack_manager.rs
+++ b/consensus/src/ordered_broadcast/ack_manager.rs
@@ -158,11 +158,9 @@ impl<P: PublicKey, S: Scheme, D: Digest> AckManager<P, S, D> {
 }
 
 #[cfg(test)]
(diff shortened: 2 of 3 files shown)


consensus/src/ordered_broadcast/types.rs
```diff
@@ -148,11 +148,11 @@ pub trait SequencersProvider: Clone + Send + Sync + 'static {
 
 /// Suffix used to identify a chunk namespace for domain separation.
 /// Used when signing and verifying chunks to prevent signature reuse across different message types.
-pub const CHUNK_SUFFIX: &[u8] = b"_CHUNK";
+pub const CHUNK_SUFFIX: &[u8] = b"_COMMONWARE_CONSENSUS_ORDERED_BROADCAST_CHUNK";
```
patrick-ogrady: For a suffix, I'm not sure we need this all?

I think we could keep aligned with Simplex here: https://github.com/commonwarexyz/monorepo/blob/0689ca8f0f22920e9ab19f713dcb8bc0746a3738/consensus/src/simplex/scheme/mod.rs#L120-L125

patrick-ogrady: I don't love how we use a full namespace in some areas and suffix in others (not sure if that should be another issue?)

BrendanChou: reverted in aeb6aac.

The confusion was that AGENTS.md documents the `_COMMONWARE_<CRATE>_<OPERATION>` convention for root namespaces, and I mistakenly applied it to suffixes. But suffixes get appended via `union()` to a base namespace that already contains the full path, so the short form (`_CHUNK`, `_ACK`) is correct here.

clabby (approved):

PR #3218, opened 2026-02-23
CI: move benchmark name linting to benchmark running step

cronokirby:
Previously, linting benchmark names was by far the longest step in `just lint`, taking about a minute, which was painful.

The reason for this is that the approach needed to compile and run the benchmark executables, including setup code, in order to get the names.

This commit modifies the approach by moving this lint into the CI flow for running benchmarks, which is more appropriate.

### DIFF
--- a/justfile
+++ b/justfile
@@ -53,16 +53,8 @@ fix: fix-clippy fix-fmt fix-toml-fmt fix-features
 
 # Tests benchmarks in a given crate
 test-benches crate *args='':
-    #!/usr/bin/env bash
-    set -euo pipefail
-    list_output="$(mktemp)"
-    cleanup() {
-        rm -f "$list_output"
-    }
-    trap cleanup EXIT
     cargo test --benches -p {{ crate }} {{ args }} -- --verbose
-    cargo test --benches -p {{ crate }} {{ args }} -- --list | tee "$list_output"
-    python3 .github/scripts/lint_benchmark_names.py "$list_output"
+    cargo test --benches -p {{ crate }} {{ args }} -- --list | python3 .github/scripts/lint_benchmark_names.py -
 
 # Run tests
 test *args='':


.github/workflows/benchmark.yml
```diff
@@ -72,6 +72,8 @@ jobs:
             --benches -p ${{ matrix.package }} \
             -- --output-format bencher \
             | tee "${{ matrix.package }}${{ matrix.file_suffix }}.txt"
+      - name: Lint benchmark names
```
patrick-ogrady: This is the periodic worker that runs on main. We don't need to run this here?

.github/workflows/fast.yml
```diff
@@ -322,20 +322,6 @@ jobs:
     - name: Check Cargo.toml formatting
       run: just fix-toml-fmt && git diff --exit-code
 
-  lint-benchmark-names:
```
patrick-ogrady, line 325: Do we need to make this change? I think this should remain as a standalone job here (if you want to run with each check...I think there is a CI for testing benchmarks for each?)

patrick-ogrady: Ah, I see you moved to `test-benches`?

cronokirby: Right, the intention was to make it so that the benchmarking work already present is used to generate the names to go into this lint, rather than having a separate compilation and execution pass just for that. It should be amended so that we move it to the benchmark testing though, and not the benchmark execution.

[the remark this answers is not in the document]
cronokirby: Hmm, the reason for this lint is to not break the public facing page, so it seems fine to not check other benches?

justfile
```diff
@@ -45,15 +45,23 @@ clippy *args='':
...
 test-benches crate *args='':
-    cargo test --benches -p {{ crate }} {{ args }} -- --verbose
+    #!/usr/bin/env bash
+    set -euo pipefail
+    output="$(mktemp)"
+    cleanup() {
+        rm -f "$output"
+    }
+    trap cleanup EXIT
+    cargo bench -p {{ crate }} {{ args }} --benches -- --list | tee "$output"
```
clabby: The old recipe ran `cargo test --benches` which executed each benchmark in test mode (catching panics in setup/iteration code). The new recipe uses `cargo bench -- --list` which only compiles and lists benchmark names without executing any benchmark logic. I think we should continue running the tests?

justfile
```diff
@@ -45,15 +45,23 @@ clippy *args='':
...
-    cargo test --benches -p {{ crate }} {{ args }} -- --verbose
+    #!/usr/bin/env bash
+    set -euo pipefail
+    output="$(mktemp)"
+    cleanup() {
+        rm -f "$output"
+    }
+    trap cleanup EXIT
+    cargo test --benches -p {{ crate }} {{ args }} -- --verbose | tee "$output"
+    python3 .github/scripts/lint_benchmark_names.py "$output"
```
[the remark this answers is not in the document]
clabby: Ah we probably need to run `--list` as well, after testing the benches. Fortunately the compilation work will be done by the tests.

clabby (approved):

clabby (dismissed):

BrendanChou (dismissed):

PR #3223, opened 2026-02-24
[coding] Short-circuit invalid `leaf_count` in weaken & check

clabby:
## Overview

Adds checks for valid `leaf_count` in the BMT proof that accompanies a strong/weak shard in `RS::weaken` and `RS::check`. This prevents us from doing additional work before discovering that the config was invalid in `RS::decode`.

### DIFF
--- a/coding/src/reed_solomon.rs
+++ b/coding/src/reed_solomon.rs
@@ -662,6 +662,7 @@ mod tests {
     use commonware_parallel::Sequential;
     use commonware_utils::NZU16;
 
+    type RS = ReedSolomon<Sha256>;
     const STRATEGY: Sequential = Sequential;
 
     fn checked(
@@ -839,8 +840,6 @@ mod tests {
 
     #[test]
     fn test_mismatched_config_rejected_during_weaken_and_check() {
-        type RS = ReedSolomon<Sha256>;
-
         let config_expected = Config {
             minimum_shards: NZU16!(2),
             extra_shards: NZU16!(2),
@@ -1058,7 +1057,7 @@ mod tests {
 
     #[test]
     fn test_too_many_total_shards() {
-        assert!(ReedSolomon::<Sha256>::encode(
+        assert!(RS::encode(
             &Config {
                 minimum_shards: NZU16!(u16::MAX / 2 + 1),
                 extra_shards: NZU16!(u16::MAX),


coding/src/reed_solomon.rs
```diff
@@ -33,8 +33,6 @@ pub enum Error {
     InvalidDataLength(usize),
     #[error("invalid index: {0}")]
     InvalidIndex(u16),
-    #[error("wrong index: {0}")]
-    WrongIndex(u16),
```
clabby, line 37: Idk how we ended up with `InvalidIndex` and `WrongIndex` here, but figure it's okay to merge :)

patrick-ogrady: I think invalid index exceeds the allowable range and wrong index is a mismatch of index to person? Not saying we need to keep both but that's the story IIRC.

coding/src/reed_solomon.rs
```diff
@@ -597,13 +595,20 @@ impl<H: Hasher> Scheme for ReedSolomon<H> {
...
+        config: &Config,
         commitment: &Self::Commitment,
         index: u16,
         shard: Self::StrongShard,
     ) -> Result<(Self::CheckingData, Self::CheckedShard, Self::WeakShard), Self::Error> {
+        let total = total_shards(config)?;
+        if index >= total {
+            return Err(Error::InvalidIndex(index));
+        }
+        if shard.proof.leaf_count != u32::from(total) {
```
patrick-ogrady, line 607: I think including leaf_count here was also something we added later?

coding/src/reed_solomon.rs
```diff
@@ -825,6 +837,36 @@ mod tests {
         assert!(matches!(result, Err(Error::Inconsistent)));
     }
 
+    #[test]
+    fn test_mismatched_config_rejected_during_weaken_and_check() {
+        type RS = ReedSolomon<Sha256>;
```
patrick-ogrady: Is this what we do in other tests?

coding/src/reed_solomon.rs
clabby: It was, with the BMT update that seeded in the `leaf_count` with the root.

coding/src/reed_solomon.rs
clabby: No, but I lifted the alias so others use it as well.

PR #3224, opened 2026-02-24
[marshal/coding] Support non-participant reconstruction

clabby:
## Overview

Modifies the shard engine to allow non-participants to reconstruct blocks. For networks that employ DKG, a common practice is to allow for players to join 1 epoch ahead of their participation in the ceremony. Currently, this means that these nodes would need to rely on participants to forward them full blocks via the resolver. These participants were already sent a weak shard, but the leader never sent them a strong shard to form `CheckingData`.

After this PR, the leader will broadcast their own strong shard to members in the peer set that are not participants. This enables them to take advantage of the erasure coded broadcast path, and in doing so these non-participants put less of a burden on the members of the active committee.

A few rules were changed to support this:
- Leaders now send their own strong shard to non-participants.
- Non-participants now construct a reconstruction state machine on `Discovered`.
- Non-participants now accept strong shards, though the shard index is asserted to be that of the leader's, not self.
- During rebroadcast, an extra check was added to ensure only participants rebroadcast their shards.

### DIFF
--- a/consensus/src/marshal/coding/shards/engine.rs
+++ b/consensus/src/marshal/coding/shards/engine.rs
@@ -795,15 +795,17 @@ where
         }
 
         // Send the leader's strong shard to tracked peers outside the participant set.
-        let non_participants = self
+        let non_participants: Vec<P> = self
             .tracked_peers
             .iter()
             .filter(|peer| participants.index(peer).is_none())
             .cloned()
             .collect();
-        let _ = sender
-            .send(Recipients::Some(non_participants), leader_shard, true)
-            .await;
+        if !non_participants.is_empty() {
+            let _ = sender
+                .send(Recipients::Some(non_participants), leader_shard, true)
+                .await;
+        }
 
         // Cache the block so we don't have to reconstruct it again.
         let block = Arc::new(block);


consensus/src/marshal/coding/shards/engine.rs
```diff
@@ -1289,7 +1326,9 @@ where
     ///
     /// Strong shards (`CodingScheme::StrongShard`):
     /// - MUST be sent by a participant.
-    /// - MUST correspond to self's index (self must be a participant).
+    /// - MUST correspond to self's index when self is a participant.
+    /// - When self is not a participant (`scheme.me()` is `None`), the index
+    ///   MUST correspond to the sender's participant index (which must be the leader).
```
clabby: nit: reword to `MUST correspond to the sender's participant index (which must be the leader) when self is not a participant`

consensus/src/marshal/coding/shards/engine.rs
```diff
@@ -4142,6 +4180,339 @@ mod tests {
         );
     }
 
+    #[test_traced]
+    fn test_broadcast_routes_participant_and_non_participant_shards() {
```
clabby, line 4491: Should add harness support for non-participant peers so we don't have to hand-roll everything.

consensus/src/marshal/coding/shards/engine.rs
```diff
@@ -774,6 +794,16 @@ where
                 .await;
         }
 
+        // Send the leader's strong shard to tracked peers outside the participant set.
+        for peer in self.tracked_peers.iter() {
+            if participants.index(peer).is_some() {
+                continue;
+            }
+            let _ = sender
+                .send(Recipients::One(peer.clone()), leader_shard.clone(), true)
```
patrick-ogrady: We can do this as a single Recipients::Many/Multiple or whatever it is?

consensus/src/marshal/coding/shards/engine.rs
```diff
@@ -446,6 +451,13 @@ where
             on_stopped => {
                 debug!("received shutdown signal, stopping shard engine");
             },
+            Some((_, _, tracked_peers)) = self.peer_set_subscription.recv() else {
+                debug!("peer set subscription closed");
+                return;
+            } => {
+                self.tracked_peers = tracked_peers.clone();
+                self.peer_buffers.retain(|peer, _| tracked_peers.as_ref().contains(peer));
```
patrick-ogrady: if you do this before setting tracked peers, you may not need a clone?

consensus/src/marshal/coding/shards/engine.rs
```diff
@@ -742,7 +746,13 @@ where
             return;
         };
         let participants = scheme.participants();
-        let me = scheme.me();
+        let Some(me) = scheme.me() else {
+            warn!(
+                %commitment,
+                "cannot broadcast shards: local proposer is not a participant"
```
patrick-ogrady, line 757: This isn't a participant check? Wouldn't that be `me in participants`?

consensus/src/marshal/coding/shards/engine.rs
```diff
@@ -755,9 +765,19 @@ where
             return;
         }
 
-        // Broadcast each shard to the corresponding participant.
+        let my_index = me.get() as usize;
+        let Some(leader_shard) = block.shard(my_index as u16) else {
+            warn!(
+                %commitment,
+                index = my_index,
+                "cannot broadcast shards: missing leader shard"
```
patrick-ogrady: Can you elaborate with an inline comment as to why this isn't possible with the leader shard?

consensus/src/marshal/coding/shards/engine.rs
```diff
@@ -809,12 +839,19 @@ where
         sender: &mut WrappedSender<Sr, Shard<C, H>>,
         commitment: Commitment,
     ) {
+        let should_broadcast_weak = self
```
patrick-ogrady: "broadcast weak if participant"? Might go a long way?

consensus/src/marshal/coding/shards/engine.rs
```diff
@@ -809,12 +839,19 @@ where
         sender: &mut WrappedSender<Sr, Shard<C, H>>,
```
patrick-ogrady, line 842: codex:

```
  - [P2] Keep weak-shard broadcast independent of provider churn — /Users/patrickogrady/code/monorepo-pr-3224/consensus/src/marshal/coding/
    shards/engine.rs:842-850
    try_advance now gates weak-shard gossip on a fresh scheme_provider.scoped(state.round().epoch()) lookup, then immediately consumes
    our_weak_shard; with providers that dynamically unregister epochs, this lookup can return None after state creation, causing the shard to be
    dropped without ever being sent. In that case an honest participant that already validated its shard silently stops contributing gossip,
    which can block reconstruction when redundancy is tight (for example, low extra_shards or slow peers).


• Minimal fix I’d make:

  1. Persist “am I a participant for this commitment?” in reconstruction state when the strong shard is validated (using the same scheme used for
     validation).

  - Add a bool on CommonState:1035, e.g. broadcast_our_weak.
  - Set it in insert_strong_shard/verify_strong_shard from me.is_some().

  2. In try_advance:837, stop doing a fresh scheme_provider.scoped(...) lookup.

  - Take our_weak_shard from state.
  - Broadcast iff broadcast_our_weak is true.
  - Always notify shard subscribers once the weak shard is consumed.

  That gives the behavior you want:

  - Non-participants still do not gossip.
  - Participants that already validated their shard will still gossip even if provider mappings churn later.
  ```

File: consensus/src/marshal/coding/shards/engine.rs (hunk 5 above)
Comment (clabby) [reply consensus/src/marshal/coding/shards/engine.rs:757]: This is a participant check iiuc - if the local node isn't a participant, `Scheme::me` will be `None` because it's a verifier-only instance.

File: consensus/src/marshal/coding/shards/engine.rs (hunk 6 above)
Comment (clabby) [reply consensus/src/marshal/coding/shards/engine.rs]: Changed to an `expect`, since this is unreachable.

File: consensus/src/marshal/coding/shards/engine.rs [hunk 9]
```diff
@@ -134,7 +134,8 @@
 //! The engine enforces strict validation to prevent Byzantine attacks:
 //!
 //! - All shards MUST be sent by participants in the current epoch.
-//! - Strong shards MUST correspond to the recipient's index.
+//! - Strong shards MUST correspond to the recipient's index for participants.
+//! - For non-participants, strong shards MUST correspond to the leader's index.
```
patrick-ogrady, line 142: I think we should make it clearer in the docs somewhere that we support non-participant sync (this is pretty cool).

consensus/src/marshal/coding/shards/engine.rs
```diff
@@ -810,13 +836,14 @@ where
...
     ) {
-        if let Some(weak_shard) = self
-            .state
-            .get_mut(&commitment)
-            .and_then(|s| s.take_weak_shard())
-        {
-            self.broadcast_weak_shard(sender, weak_shard).await;
-            self.notify_shard_subscribers(commitment);
+        if let Some(state) = self.state.get_mut(&commitment) {
+            let should_broadcast_weak = state.broadcast_our_weak();
```
patrick-ogrady: Why not just guard `take_weak_shard` on this? I wonder if we can change the name of that function in the process as well?

consensus/src/marshal/coding/shards/engine.rs
```diff
@@ -1372,15 +1417,12 @@ where
         &mut self,
         me: Option<&Participant>,
         sender: P,
```
patrick-ogrady: would recommend `sender: (P, Participant)`?

patrick-ogrady:
Two LLM nits:

```
• The new non-participant routing depends on tracked_peers, but broadcasting can occur before that set is initialized from subscription updates.
  In that startup race, non-participants do not receive the leader strong shard and cannot reconstruct the commitment through shard gossip.

  Review comment:

  - [P2] Populate tracked peers before broadcasting to non-participants — /Users/patrickogrady/code/monorepo-pr-3224/consensus/src/marshal/
    coding/shards/engine.rs:793-801
    broadcast_shards derives non-participant recipients only from self.tracked_peers, but that set starts empty and is populated later via an
    async peer-set subscription. If a Proposed message is processed before the first subscription update (for example during startup),
    Recipients::Some is empty and non-participants miss the only strong shard they can accept for that commitment. Since non-participants cannot
    produce checking_data without that strong shard, they cannot reconstruct that proposal from later weak gossip.

````

```
• The patch introduces a hot-path performance bug by always encoding and attempting to send the leader shard to an empty non-participant
  recipient list. This regression is significant enough to warrant fixing before considering the change correct.

  Review comment:

  - [P1] Avoid encoding leader shard when no non-participants exist — /Users/patrickogrady/code/monorepo-pr-3224/consensus/src/marshal/coding/
    shards/engine.rs:794-802
    broadcast_shards always calls sender.send(Recipients::Some(non_participants), leader_shard, true) even when non_participants is empty. Since
    WrappedSender::send encodes the shard before dispatch, validator-only deployments now do an extra full strong-shard encode/allocation per
    proposal with zero recipients, which is a hot-path throughput regression. Guard this send behind a non-empty recipient check.

```

File: consensus/src/marshal/coding/shards/engine.rs [hunk 12]
```diff
@@ -729,7 +734,10 @@ where
         self.notify_block_subscribers(block);
     }
 
-    /// Broadcasts the shards of a [`CodedBlock`] to all participants and caches the block.
+    /// Broadcasts the shards of a [`CodedBlock`] and caches the block.
+    ///
+    /// - Participants receive the strong shard matching their participant index.
+    /// - Tracked non-participants receive the leader's strong shard.
     async fn broadcast_shards<Sr: Sender<PublicKey = P>>(
```
patrick-ogrady, line 741: This is only called by the leader?

consensus/src/marshal/coding/shards/engine.rs
```diff
@@ -756,9 +770,14 @@ where
...
 
-        // Broadcast each shard to the corresponding participant.
+        let my_index = me.get() as usize;
+        let leader_shard = block
+            .shard(my_index as u16)
+            .expect("proposer's shard must exist");
+
+        // Broadcast each participant their corresponding shard.
         for (index, peer) in participants.iter().enumerate() {
-            if me.is_some_and(|m| m.get() as usize == index) {
```
patrick-ogrady, line 761: Happy to get rid of this.

PR #3225, opened 2026-02-24
[consensus/simplex] Improve Documentation/Testing for Certification after Nullification + Broadcast Nullify before Nullification

patrick-ogrady:


consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -975,12 +1034,13 @@ impl<
                             }
                             Certificate::Nullification(nullification) => {
                                 trace!(%view, from_resolver, "received nullification");
-                                if let Some(floor) = self.handle_nullification(nullification).await
-                                {
-                                    warn!(?floor, "broadcasting nullification floor");
-                                    self.broadcast_certificate(&mut certificate_sender, floor)
-                                        .await;
-                                }
+                                self.handle_verified_nullification(
```
patrick-ogrady: nit: why can't this just be `handle_nullification`?

consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -975,12 +1011,22 @@ impl<
                             }
                             Certificate::Nullification(nullification) => {
                                 trace!(%view, from_resolver, "received nullification");
+                                let late_timeout_nullify = self.try_construct_nullify(view);
```
patrick-ogrady: last item: figure out how to organize this

consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -388,6 +401,20 @@ impl<
             .and_then(|_| self.state.parent_certificate(view))
     }
 
+    /// If a nullification certificate arrives for our current view, we should emit
+    /// a nullify vote (just like we would for a notarization or finalization).
+    ///
+    /// Returns `Some(nullify)` only when:
+    /// 1. The view is the current view.
+    /// 2. We have not already emitted nullify for that view.
+    fn try_construct_nullify(&mut self, view: View) -> Option<Nullify<S>> {
```
patrick-ogrady: I don't love this flow but couldn't think of a cleaner way

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -213,20 +213,38 @@ impl<E: Clock + CryptoRngCore + Metrics, S: Scheme<D>, L: ElectorConfig<S>, D: D
...
+    ///
+    /// When `retry` is false, only returns a vote on the first attempt.
+    /// When `retry` is true, returns a vote on both first attempts and retries.
+    ///
+    /// Returns `None` if we have already broadcast a finalize vote for this view.
+    pub fn try_nullify(&mut self, view: View, retry: bool) -> Option<(bool, Nullify<S>)> {
+        if view != self.view {
+            return None;
+        }
+        let is_retry = self.create_round(view).construct_nullify()?;
```
patrick-ogrady, line 232: This round should exist for self.view, so I suppose we could assert on that.

consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -975,12 +992,30 @@ impl<
                             }
                             Certificate::Nullification(nullification) => {
                                 trace!(%view, from_resolver, "received nullification");
+
+                                // Construct a nullify vote before updating current view (occurs during
+                                // `handle_nullification`), if we haven't timed out already
```
patrick-ogrady: Broadcast nullify before nullification?

consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -975,6 +1018,8 @@ impl<
                             }
                             Certificate::Nullification(nullification) => {
                                 trace!(%view, from_resolver, "received nullification");
+
+                                // Handle the nullification certificate
```
patrick-ogrady: remove this

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -213,38 +213,52 @@ impl<E: Clock + CryptoRngCore + Metrics, S: Scheme<D>, L: ElectorConfig<S>, D: D
...
-    /// Handle a timeout event for the current view.
-    /// Returns the nullify vote and optionally an entry certificate for the previous view
-    /// (if this is a retry timeout and we can construct one).
-    pub fn handle_timeout(&mut self) -> (bool, Option<Nullify<S>>, Option<Certificate<S, D>>) {
-        let view = self.view;
-        let Some(retry) = self.create_round(view).construct_nullify() else {
-            return (false, None, None);
-        };
-        let nullify = Nullify::sign::<D>(&self.scheme, Rnd::new(self.epoch, view));
+    /// Constructs a nullify vote for `view`, if eligible.
```
BrendanChou, line 216: nit: explain the returned bool in the first part of the optional tuple

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -213,38 +213,52 @@ impl<E: Clock + CryptoRngCore + Metrics, S: Scheme<D>, L: ElectorConfig<S>, D: D
...
+        }
+        let nullify = Nullify::sign::<D>(&self.scheme, Rnd::new(self.epoch, view))?;
+        Some((is_retry, nullify))
+    }
 
-        // If was retry, we need to get entry certificates for the previous view
-        let entry_view = view.previous().unwrap_or(GENESIS_VIEW);
-        if !retry || entry_view == GENESIS_VIEW {
-            return (retry, nullify, None);
+    /// Returns the best certificate for `view` to help peers enter it.
```
BrendanChou: suggestion: "to help peers enter `view+1`"

consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -323,46 +323,68 @@ impl<
...
+
+        // Broadcast entry to help others enter the view.
         //
         // We don't worry about recording this certificate because it must've already existed (and thus
         // we must've already broadcast and persisted it).
-        if let Some(certificate) = entry {
+        if !retry {
+            return;
+        }
+        let Some(past_view) = view.previous() else {
```
BrendanChou: assert this? Can we even hit the code path where `view.previous()` is none? Since view shouldn't be genesis

patrick-ogrady: If we are in view 1, then we could hit it (in fact I think this would be very common in practice that view 1 is a nullification)?

BrendanChou: I think if you're at view 1, it gives you `Some(0)` which will then proceed  to the next statement and then fail to find a best certificate for it since it's genesis (thereby returning)

patrick-ogrady: Ah, I was thinking about loading the entry. you're right.

BrendanChou (approved):

clabby (approved):

PR #3234, opened 2026-02-24
[p2p] Add block! macro for consistent blocker logging

### LINKED ISSUE #3231: [p2p]: add `block!` macro for consistent blocker logging
Every `Blocker::block()` call site manually logs a `warn!` with a reason string before blocking. This is repetitive (31 call sites across simplex, aggregation, marshal, and ordered_broadcast) and easy to forget when adding new block calls.

Add a `block!` macro in `commonware-p2p` (next to the `Blocker` trait) that expands the `warn!` at the caller's site, preserving file/line numbers in `tracing`:

```rust
macro_rules! block {
    ($blocker:expr, $peer:expr, $($arg:tt)+) => {
        warn!(peer = ?$peer, $($arg)+);
        $blocker.block($peer).await;
    };
}
```

The `$($arg:tt)+` passes through any extra `tracing` fields and the message string directly to `warn!`, so call sites remain flexible:

```rust
block!(self.blocker, sender, ?err, "invalid ack signature");
block!(self.blocker, sender, %view, "blocking peer for epoch mismatch");
```

This avoids changing the `Blocker` trait, keeps logging consistent, always includes the peer, retains caller location in logs, and allows extra debug context. The codebase already uses macros with `.await` inside (e.g., `commonware_macros::select!`).

Raised in https://github.com/commonwarexyz/monorepo/pull/3220#discussion_r2843595187

BrendanChou:
Add a `block!` macro in `commonware-p2p` that combines `tracing::warn!` with `Blocker::block()` into a single call, ensuring consistent logging at every block site. The macro always includes the peer as a structured `peer` field and passes through any extra tracing fields. Replaces all 31 `warn!` + `.block()` call sites across `p2p`, `consensus` (simplex, marshal), and `resolver`.

The macro is annotated at BETA stability because it is consumed by BETA code within the `p2p` crate itself (`utils/codec.rs`). An ALPHA annotation would cause the macro to be stripped when checking at the BETA level, breaking those call sites.

Closes #3231

### DIFF
--- a/p2p/src/authenticated/discovery/mod.rs
+++ b/p2p/src/authenticated/discovery/mod.rs
@@ -237,7 +237,7 @@ mod tests {
             discovery::actors::router::{Actor as RouterActor, Config as RouterConfig},
             relay::Relay,
         },
-        Blocker, Ingress, Manager, Provider, Receiver, Recipients, Sender,
+        Ingress, Manager, Provider, Receiver, Recipients, Sender,
     };
     use commonware_cryptography::{ed25519, Signer as _};
     use commonware_macros::{select, select_loop, test_group, test_traced};
--- a/p2p/src/lib.rs
+++ b/p2p/src/lib.rs
@@ -324,7 +324,10 @@ macro_rules! block {
 }
 
 /// Block a peer without logging.
-#[allow(clippy::disallowed_methods, reason = "test helper that bypasses the block! macro")]
+#[allow(
+    clippy::disallowed_methods,
+    reason = "test helper that bypasses the block! macro"
+)]
 #[cfg(test)]
 pub async fn block_peer<B: Blocker>(blocker: &mut B, peer: B::PublicKey) {
     blocker.block(peer).await;
--- a/p2p/src/authenticated/discovery/actors/tracker/actor.rs
+++ b/p2p/src/authenticated/discovery/actors/tracker/actor.rs
@@ -286,7 +286,7 @@ mod tests {
             },
             Mailbox,
         },
-        Blocker, Ingress, Manager,
+        Ingress, Manager,
     };
     use commonware_codec::{DecodeExt, Encode};
     use commonware_cryptography::{
--- a/p2p/src/authenticated/lookup/actors/tracker/actor.rs
+++ b/p2p/src/authenticated/lookup/actors/tracker/actor.rs
@@ -274,7 +274,7 @@ impl<E: Spawner + Rng + Clock + RuntimeMetrics, C: Signer> Actor<E, C> {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::{authenticated::lookup::actors::peer, AddressableManager, Blocker, Ingress};
+    use crate::{authenticated::lookup::actors::peer, AddressableManager, Ingress};
     use commonware_cryptography::{
         ed25519::{PrivateKey, PublicKey},
         Signer,
--- a/p2p/src/authenticated/lookup/mod.rs
+++ b/p2p/src/authenticated/lookup/mod.rs
@@ -183,9 +183,7 @@ pub use network::Network;
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::{
-        Address, AddressableManager, Blocker, Ingress, Provider, Receiver, Recipients, Sender,
-    };
+    use crate::{Address, AddressableManager, Ingress, Provider, Receiver, Recipients, Sender};
     use commonware_cryptography::{ed25519, Signer as _};
     use commonware_macros::{select, test_group, test_traced};
     use commonware_runtime::{


p2p/src/lib.rs
```diff
@@ -293,3 +293,30 @@ stability_scope!(BETA {
...
+/// included as a `peer` field in the log output.
+///
+/// # Examples
+///
+/// ```ignore
+/// block!(self.blocker, sender, "invalid message");
+/// block!(self.blocker, sender, ?err, "invalid ack signature");
+/// block!(self.blocker, sender, %view, "blocking peer for epoch mismatch");
+/// ```
+#[cfg(not(any(
```
patrick-ogrady, line 310: You should be able to do `#[stability(BETA)]`?

consensus/src/simplex/actors/batcher/round.rs
```diff
@@ -196,8 +188,7 @@ impl<
             Vote::Finalize(finalize) => {
                 // Verify sender is signer
                 if index != finalize.signer() {
-                    warn!(?sender, "blocking peer for finalize signer mismatch");
-                    self.blocker.block(sender).await;
```
patrick-ogrady, line 200: Should we add a lint that prevents calling this directly?

p2p/src/lib.rs
clabby: Haven't tried locally, but the agents.md file says:
> Manual `#[cfg(not(any(..., commonware_stability_RESERVED)))]` for `#[macro_export]` macros (can't use `stability_scope!` due to Rust limitations)

consensus/src/simplex/actors/batcher/round.rs
clabby: `disallowed_methods` would be good here.

p2p/src/lib.rs
BrendanChou: Confirmed locally: `#[stability(BETA)]` does not work here. The proc macro expands `#[macro_export]`, and then `crate::block!` becomes an absolute path to a macro-expanded export, which Rust denies with `macro_expanded_macro_exports_accessed_by_absolute_paths` ([rust-lang/rust#52234](https://github.com/rust-lang/rust/issues/52234)). The manual `#[cfg(not(any(...)))]` is the only option for `#[macro_export]` macros.

consensus/src/simplex/actors/batcher/round.rs
BrendanChou: `disallowed_methods` would fire on the macro expansion itself since `block!` calls `.block()` internally. Clippy checks after macro expansion, so banning the method would also ban the macro that wraps it (same reason `tokio::select` is not in `disallowed_macros` - our `select!` expands to it).

BrendanChou: I guess we could add `disallowed_methods` but also bypass that lint within the macro

p2p/src/authenticated/discovery/mod.rs
```diff
@@ -230,6 +230,7 @@ pub use config::{Bootstrapper, Config};
 pub use network::Network;
 
 #[cfg(test)]
+#[allow(clippy::disallowed_methods)]
```
clabby: small nit, but we should add a reason to the allow on these modules.

```suggestion
#[allow(clippy::disallowed_methods, reason = "Blocker::block is allowed in tests")]
```

p2p/src/lib.rs
```diff
@@ -293,3 +293,31 @@ stability_scope!(BETA {
...
+    commonware_stability_RESERVED
+)))] // BETA
+#[macro_export]
+macro_rules! block {
+    ($blocker:expr, $peer:expr, $($arg:tt)+) => {
+        tracing::warn!(peer = ?$peer, $($arg)+);
+        #[allow(clippy::disallowed_methods)]
+        $blocker.block($peer).await;
+    };
+}
```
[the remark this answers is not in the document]
BrendanChou: Fixed in df3a531 - the macro now binds `$peer` to a local before using it.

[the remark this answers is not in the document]
BrendanChou: Not an issue. The macro contains `.await` so it can only appear inside async function bodies, never in a braces-free expression context. The `let peer = $peer;` binding also already prevents expression-position usage at compile time.

consensus/src/marshal/coding/shards/engine.rs
```diff
@@ -1391,33 +1384,29 @@ where
             .try_into()
             .expect("participant index impossibly out of bounds");
         if shard.index != expected_index {
-            warn!(
-                ?sender,
+            commonware_p2p::block!(
```
patrick-ogrady, line 1387: damn, this is sick

p2p/src/lib.rs
```diff
@@ -293,3 +293,42 @@ stability_scope!(BETA {
...
+    commonware_stability_GAMMA,
+    commonware_stability_DELTA,
+    commonware_stability_EPSILON,
+    commonware_stability_RESERVED
+)))] // BETA
+#[macro_export]
+macro_rules! block {
+    ($blocker:expr, $peer:expr, $($arg:tt)+) => {
+        let peer = $peer;
+        tracing::warn!(peer = ?peer, $($arg)+);
```
patrick-ogrady, line 320: There are a number of cases where no args are provided...will this just print an empty warn log? Should we require at least one arg is provided?

BrendanChou: The `+` in `$($arg:tt)+` already enforces at least one token is provided. `block!(blocker, peer)` without a message won't compile.

BrendanChou: By "no args provided" you might mean the `block_peer()` method which doesn't require a string and is only used in tests (`#[cfg(test)]`) ?

patrick-ogrady: Yeah, that's my b.

PR #3239, opened 2026-02-25
[ci] add cargo-hack feature-powerset check

clabby:
Replace the `cargo build --all-features` step in the `build` CI matrix
with `cargo hack check --feature-powerset --no-dev-deps`, verifying that
all feature combinations compile on each toolchain (msrv, stable, nightly).

Add a `hack` recipe to the justfile that wraps `cargo hack` with
`--feature-powerset --no-dev-deps`, accepting an arbitrary subcommand
(e.g. `just hack check`, `just hack clippy`).

Remove `cdylib` from `crate-type` in all library crates (cryptography,
consensus, math, runtime, storage, utils). The `cdylib` crate type
requires an allocator, so these crates cannot compile as `cdylib` without
`std` (i.e. with `--no-default-features`), which breaks the feature
powerset check. Update the WASM CI job to use `cargo rustc --crate-type
cdylib` instead, producing the `.wasm` artifacts on demand.

### DIFF
--- a/cryptography/Cargo.toml
+++ b/cryptography/Cargo.toml
@@ -25,7 +25,7 @@ commonware-macros.workspace = true
 commonware-math.workspace = true
 commonware-parallel.workspace = true
 commonware-utils.workspace = true
-crc-fast = { workspace = true, features = ["panic-handler"] }
+crc-fast = { workspace = true, features = ["std"], optional = true }
 ctutils.workspace = true
 ecdsa.workspace = true
 ed25519-consensus = { workspace = true, default-features = false }
@@ -91,7 +91,8 @@ std = [
 	"commonware-math/std",
 	"commonware-parallel/std",
 	"commonware-utils/std",
-	"crc-fast/std",
+	"crc-fast",
+	"crc-fast?/std",
 	"ecdsa/std",
 	"ed25519-consensus/std",
 	"getrandom/std",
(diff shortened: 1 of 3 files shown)


cryptography/Cargo.toml
```diff
@@ -25,7 +25,7 @@ commonware-macros.workspace = true
 commonware-math.workspace = true
 commonware-parallel.workspace = true
 commonware-utils.workspace = true
-crc-fast = { workspace = true, features = ["panic-handler"] }
+crc-fast = { workspace = true, features = ["std"], optional = true }
```
clabby, line 28: `no_std` builds for this crate are broken (at least on x86.) https://github.com/awesomized/crc-fast-rust/blob/3a853cc7daf2cd47cc4466f198680cabdfb0b5fa/src/arch/x86_64/avx512.rs#L11

clabby: There's a fix that's not yet merged available at https://github.com/awesomized/crc-fast-rust/pull/42.

.github/workflows/fast.yml
```diff
@@ -281,19 +281,19 @@ jobs:
     - name: Add WASM target
       run: rustup target add wasm32-unknown-unknown
     - name: Build cryptography
-      run: cargo build --target wasm32-unknown-unknown --release --manifest-path cryptography/Cargo.toml && du -h "${CARGO_TARGET_DIR:-target}/wasm32-unknown-unknown/release/commonware_cryptography.wasm"
+      run: cargo rustc --target wasm32-unknown-unknown --release --manifest-path cryptography/Cargo.toml --crate-type cdylib && du -h "${CARGO_TARGET_DIR:-target}/wasm32-unknown-unknown/release/commonware_cryptography.wasm"
```
patrick-ogrady, line 284: skill issue 😓

PR #3241, opened 2026-02-25
[consensus] Unify Nullify Handling

patrick-ogrady:
Related: #3225

### DIFF
--- a/consensus/src/simplex/actors/voter/actor.rs
+++ b/consensus/src/simplex/actors/voter/actor.rs
@@ -543,9 +543,7 @@ impl<
         view: View,
         timeout: bool,
     ) -> Option<bool> {
-        let Some((was_retry, nullify)) = self.state.construct_nullify(view, timeout) else {
-            return None;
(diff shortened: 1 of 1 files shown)


consensus/src/simplex/mod.rs
```diff
@@ -165,6 +165,24 @@
 //! participants eventually certify and broadcast `finalize(c,v)`, the finalization
 //! certificate will form.
 //!
+//! ### Activity Tracking
```
patrick-ogrady: We should improve this (and explain why we do this).

patrick-ogrady: Right now, it would be called `Sending Useless Votes`

patrick-ogrady: As @cronokirby mentioned, a much more efficient workaround here would be to just track certificate broadcast as "active" (and then we only broadcast votes that are strictly necessary -> this would impact what votes are emitted on the activity stream).

PR #3242, opened 2026-02-25
[runtime/storage] support vectored writes

### LINKED ISSUE #3226: [runtime] support vectored writes in storage backends
With #3185 supporting vectored writes in storage will allow us to avoid materialization of pages into a single contiguous buffer.

Related #2883.

andresilva:
This PR adds vectored write support to runtime storage backends to avoid unnecessary coalescing when inputs are already chunked.

- tokio/fallback: single path uses `write_all`, chunked path uses `write_all_buf`
- tokio/unix: single path uses `write_all_at`, chunked path uses batched `pwritev`
- io_uring: single path submits `Write` op, chunked batch submits `Writev` ops (with `iovec` metadata kept alive until completion)

`IoBufs` / `IoBufsMut` also specializes `Buf::chunks_vectored` now, since the default implementation extracts one chunk at a time.

Fixes #3226.
Related #2883.
Future Work #3269

### DIFF
--- a/runtime/src/iouring/mod.rs
+++ b/runtime/src/iouring/mod.rs
@@ -157,7 +157,7 @@ const SUBMISSION_SEQ_MASK: u64 = u64::MAX >> 1;
 ///
 /// The variant must match the operation type:
 /// - `Read`: For operations where the kernel writes INTO the buffer (e.g., recv, read)
-/// - `Write`: For operations where the kernel reads FROM a single contiguous buffer (e.g., send)
+/// - `Write`: For operations where the kernel reads FROM a single contiguous buffer (e.g., send, write)
 /// - `WriteVectored`: For operations where the kernel reads FROM multiple buffers (e.g., writev)
 #[derive(Debug)]
 pub enum OpBuffer {
@@ -874,8 +874,9 @@ impl IoUringLoop {
             // Submit the operation.
             //
             // SAFETY:
-            // - `buffer` and `fd` are stored in `self.waiters` until CQE processing, so
-            //   SQE pointers remain valid and FD numbers cannot be reused early.
+            // - `buffer`, `fd` and `iovecs` are stored in `self.waiters` until CQE
+            //   processing, so SQE pointers remain valid and FD numbers cannot be reused
+            //   early.
             // - `IO_LINK` is set on `work` before pushing it, so the following timeout
             //   SQE applies to this operation.
             // - `available >= needed` was checked above, so this push fits.


runtime/src/storage/tokio/unix.rs
```diff
@@ -66,8 +121,12 @@ impl crate::Blob for Blob {
             .checked_add(Header::SIZE_U64)
             .ok_or(Error::OffsetOverflow)?;
         task::spawn_blocking(move || {
-            file.write_all_at(bufs.coalesce().as_ref(), offset)?;
-            Ok(())
+            #[allow(clippy::option_if_let_else)]
```
clabby: nit: `#[allow(..., reason = "...")]` or use `option_or_else`

runtime/src/iouring/mod.rs
clabby, line 1087: This should handle EINTR?

clabby: This should handle EINTR, EAGAIN, and EBUSY, and retry if so (or surface to the user an error other than a panic)? [io_uring_enter(2)](https://man7.org/linux/man-pages/man2/io_uring_enter.2.html) says both EAGAIN and EBUSY are transient and should retry after completions.

clabby: This should handle EINTR, EAGAIN, and EBUSY, and retry if so?

clabby (commented):
iovec comes together, this is clean 👏

runtime/src/iouring/mod.rs
andresilva: EINTR shouldn't really happen with io_uring AFAIU since it's asynchronous, can only happen on submission. I guess we can add it just to be defensive.

andresilva: Agreed. I refrained from fixing this here because it will clash with the changes in #3159.

andresilva: https://github.com/commonwarexyz/monorepo/issues/3252

clabby (dismissed):

runtime/src/storage/tokio/unix.rs
andresilva: clippy gets it wrong here. I updated this code anyway to use `try_into_single`.

runtime/src/iouring/mod.rs
```diff
@@ -157,13 +157,18 @@ const SUBMISSION_SEQ_MASK: u64 = u64::MAX >> 1;
...
 #[derive(Debug)]
 pub enum OpBuffer {
     /// Buffer for read operations - kernel writes into this.
     Read(IoBufMut),
     /// Buffer for write operations - kernel reads from this.
     Write(IoBuf),
+    /// Buffers for vectored write operations - kernel reads from these.
+    ///
+    /// NOTE: currently this is only used by the storage backend, hence the allow dead code.
+    WriteVectored(#[cfg_attr(not(feature = "iouring-storage"), allow(dead_code))] IoBufs),
```
patrick-ogrady, line 171: Added: #3269

runtime/src/storage/tokio/unix.rs
```diff
@@ -21,6 +30,52 @@ impl Blob {
...
+        while bufs.has_remaining() {
+            let mut io_slices = [IoSlice::new(&[]); IOVEC_BATCH_SIZE];
+            let io_slices_len = bufs.chunks_vectored(&mut io_slices);
+            assert!(
+                io_slices_len > 0,
+                "chunks_vectored should produce at least one slice when bufs has remaining"
+            );
+
+            // std::os::unix::fs::FileExt::write_vectored_at is unstable:
+            // https://doc.rust-lang.org/stable/std/os/unix/fs/trait.FileExt.html#method.write_vectored_at
```
patrick-ogrady, line 49: It would be so cool if you could conditionally nightly

PR #3245, opened 2026-02-25
[p2p] track commitments only after successful send

VolodymyrBg:
## Why
Before this change, the engine could insert a request commitment into [tracked](https://github.com/commonwarexyz/monorepo/blob/92647ea3648d631426ed8e9b03735be4ec0eb22e/p2p/src/authenticated/discovery/actors/tracker/directory.rs#L314-L321) before the network send result was known.
If [send(...)](https://github.com/commonwarexyz/monorepo/blob/main/p2p/src/utils/codec.rs#L49-L58) failed, that commitment could remain tracked even though nothing was actually sent, which could inflate `outstanding` and retain unnecessary state.

## What changed
- Moved commitment tracking to happen only after a successful send.
- Added a guard to track only when the returned recipient list is non-empty.
- Kept error handling unchanged for failed sends (`Error::SendFailed`) while avoiding stale tracked entries.

### DIFF
--- a/collector/src/p2p/engine.rs
+++ b/collector/src/p2p/engine.rs
@@ -150,12 +150,8 @@ where
                         recipients,
                         responder,
                     } => {
-                        // Track commitment (if not already tracked)
+                        // Compute request commitment
                         let commitment = request.commitment();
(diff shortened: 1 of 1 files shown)


clabby:

PR #3246, opened 2026-02-25
[consensus] remove noop vote handlers during journal replay

sashass1315:
During journal replay, vote artifacts (`Notarize`, `Nullify`, `Finalize`) were routed through handle_* functions that only append to the journal. At replay time self.journal is not yet attached, so these calls had no functional effect and only added unnecessary async work and cloning.


clabby:

PR #3247, opened 2026-02-25
[coding] Fix checksum malleability issue

cronokirby:
We were not committing to the checksum before sampling the indices for which rows of the checksum to verify, allowing an attacker to cause disagreement about data validity.

The fix is to generate the random coefficients for the checksum, generate and commit it, and then generate the indices to check.

This also includes a reproduction test, which fails before the fix is applied.

### DIFF
--- a/coding/src/zoda.rs
+++ b/coding/src/zoda.rs
@@ -528,13 +528,19 @@ impl<D: Digest> CheckingData<D> {
         if checksum.rows() != topology.data_rows || checksum.cols() != topology.column_samples {
             return Err(Error::InvalidShard);
         }
+        // Commit to the checksum before generating the indices to check.
+        //
+        // Nota bene: `checksum.encode()` is *serializing* the checksum, not
+        // Reed-Solomon encoding it.
+        //
+        // cf. the implementation of `Scheme::encode` for ZODA for why it's important
+        // that we do Reed-Solomon encoding of the checksum ourselves.
+        transcript.commit(checksum.encode());
         let encoded_checksum = checksum
             .as_polynomials(topology.encoded_rows)
             .expect("checksum has too many rows")
             .evaluate()
             .data();
-        // Commit to the checksum before generating the indices to check.
-        transcript.commit(checksum.encode());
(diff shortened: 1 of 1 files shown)


coding/src/zoda.rs
```diff
@@ -533,6 +533,8 @@ impl<D: Digest> CheckingData<D> {
             .expect("checksum has too many rows")
             .evaluate()
             .data();
+        // Commit to the checksum before generating the indices to check.
+        transcript.commit(checksum.encode());
```
patrick-ogrady: is there a reason not to use `encoded_checksum` here?

coding/src/zoda.rs
```diff
@@ -845,6 +848,80 @@ mod tests {
...
+            }
+            shards[1].checksum = Arc::new(checksum);
+            shards[2].checksum = shards[1].checksum.clone();
+        }
+
+        assert!(matches!(
+            Zoda::<Sha256>::weaken(&config, &commitment, b_i as u16, shards[b_i].clone()),
+            Err(Error::InvalidWeakShard)
+        ));
+        // Without robust Fiat-Shamir, this will succeed.
```
patrick-ogrady, line 927: to make it clear only the latter will succeed, I suggest a `\n` above

coding/src/zoda.rs
cronokirby: Yes, it's important for the integrity check that you encode the checksum the leader sends you *yourself*, rather than trusting that it's equal to what they claim it to be., without this, you'd be checking rows of:

$$
h - Y r
$$

but the leader can just set $h = Y r$, even if $Y$ is complete junk, and not validly encoded data.

(will add a comment)

cronokirby: Whereas, if you check rows of $G h - Y r$, then there's a link with the fact that $Y$ has to be a valid encoding.

patrick-ogrady: I didn't think about this carefully enough (and assumed the similar variable names made them ~equivalent).

cronokirby: yeah, the collision between "encode == serialize" and "encode == reed-solomon" is unfortunate

PR #3251, opened 2026-02-26
chore: remove redundant drop of old_item in ring sender

operagxoksana:
Remove redundant explicit drop of old_item. The value is dropped at the end of the scope after the mutex guard is released, so the explicit drop call is unnecessary.


clabby:

PR #3253, opened 2026-02-26
[consensus/marshal] Abstract over consensus engine

clabby:
## Overview

> [!NOTE]
> Do not merge. Stacked on top of #2884 

Abstracts `marshal` over consensus engine variants.


consensus/conformance.toml
```diff
@@ -34,9 +26,17 @@ hash = "4df7290375d7ae3b005cfad96dc5fc091bdce4a246665f94b79acf34159a3945"
...
+hash = "411897b896d677abc652d5380a8314f9197ae6570edadfb43faa2bf7a47fd975"
+
+["commonware_consensus::marshal::coding::types::test::conformance::CodecConformance<Shard<ReedSolomon<Sha256>,Sha256>>"]
+n_cases = 65536
+hash = "3dce8a0ca22497584aef2b47cd75f6fe926280b722d94a478bb78965e73a52b2"
+
 ["commonware_consensus::marshal::resolver::handler::tests::conformance::CodecConformance<Request<D>>"]
 n_cases = 65536
-hash = "481cd68f2e6452c0dda512c75d74ddd2e19ac6c9fb19c642c1a152a0d830c1b2"
+hash = "a87609f8692a7d2208a327d37bdabc7329567b15a5745f3947760a7fd383cbe2"
```
clabby, line 39: I was a bit torn about this, but after thinking about it a bit more, I believe this is safe to roll out incrementally. The resolver does not block peers on decode failure, and notarized block backfill is a secondary path in marshal. Finalized block backfill is unchanged.

While a network is rolling out their upgrades to this version, `Notarized` requests between mixed versions will fail and retry. The fetcher will eventually select a same version peer that can complete their request or cancel the request once a finalized block is processed at a higher view.

We could add in some more complex machinery that allows for nodes to understand the old `Notarized(Round)` request as well as the new `Notarized(Round, Commitment)` request temporarily, but it isn't necessary given that the penalty is slightly degraded notarized request availability for a temporary period of time.

patrick-ogrady: > The resolver does not block peers on decode failure

It does to my knowledge?

clabby: In `main`, it skips:

https://github.com/commonwarexyz/monorepo/blob/18fdce58c73107d76f1cd4c3c0e4461f77c16993/resolver/src/p2p/engine.rs#L350-L357

(Note that we did not change the response payload, which would trigger a block. We just changed the key for the `Notarized` test type.)

clabby:
https://github.com/commonwarexyz/monorepo/pull/2884#issuecomment-4972616915

PR #3255, opened 2026-02-26
[runtime/iouring] add flat `Waiters` store

andresilva:
This PR introduces a dedicated `Waiters` data structure to track in-flight operation state. `Waiters` uses fixed slots plus a freelist for O(1) insert/remove, and each operation now uses the allocated slot index as its user data.

Depends on #3159.
Related #2883.

### DIFF
--- a/runtime/src/iouring/mod.rs
+++ b/runtime/src/iouring/mod.rs
@@ -552,7 +552,11 @@ struct Waiter {
     buffer: Option<OpBuffer>,
     /// The file descriptor associated with this operation, if any. Used to keep the file
     /// descriptor alive and prevent reuse while the operation is in-flight.
-    _fd: Option<OpFd>,
+    ///
+    /// NOTE: This field is never read since it only exists to keep the FD alive until
+    /// operation completion, hence the allow dead code.
+    #[allow(dead_code)]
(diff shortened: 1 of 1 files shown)


runtime/src/iouring/mod.rs
```diff
@@ -554,6 +540,99 @@ impl Submitter {
...
+/// until CQE delivery.
+struct Waiter {
+    /// The oneshot sender used to deliver the operation result and buffer back to the
+    /// caller.
+    sender: oneshot::Sender<(i32, Option<OpBuffer>)>,
+    /// The buffer associated with this operation, if any.
+    buffer: Option<OpBuffer>,
+    /// The file descriptor associated with this operation, if any. Used to keep the file
+    /// descriptor alive and prevent reuse while the operation is in-flight.
+    _fd: Option<OpFd>,
```
patrick-ogrady: Normally, I'd put unused fields last?

Also, feels weird to use _fd eventhough we set it (maybe it raises an error without this?)

andresilva: Yeah I renamed it to `fd` and added an `allow(dead_code)` with a comment explaining the reason for this field.

PR #3257, opened 2026-02-26
[storage/translator] Translator wrapper that pre-hashes the input

roberto-bayardo:
Provides a translator wrapper that performs (random) seeded hashing of the raw input data before sending to the wrapped translator, which can be used to prevent collision-grinding attacks.

Note: Partitioned indices still use raw prefix bytes for choosing the partition, though a translator that performs hashing can be used for determining the index record within each partition.

Benchmarks are updated to test the unordered index with and without hashing, and reveal only a minimal performance impact.

### DIFF
--- a/storage/src/translator.rs
+++ b/storage/src/translator.rs
@@ -205,10 +205,9 @@ impl<const N: usize> BuildHasher for Cap<N> {
 ///
 /// # `no_std` Support
 ///
-/// [Hashed::new] and [Default] use [ahash::RandomState::new] which requires OS-provided randomness
-/// (the `runtime-rng` feature of `ahash`, enabled by the `std` feature of this crate). In `no_std`
-/// builds without `runtime-rng`, [Hashed::new] will compile but may use fixed seeds, providing no
-/// adversarial protection. In `no_std` environments, use [Hashed::from_seed] with an
+/// [Hashed::new] and [Default] are available only with this crate's `std` feature. They use
+/// [ahash::RandomState::new], which requires OS-provided randomness (the `runtime-rng` feature of
+/// `ahash`, enabled by `std` here). In `no_std` environments, use [Hashed::from_seed] with an
 /// externally-sourced random seed instead.
 ///
 /// # No Stability Guarantees


storage/src/translator.rs
```diff
@@ -188,6 +188,104 @@ impl<const N: usize> BuildHasher for Cap<N> {
...
+///
+/// Hashing destroys lexicographic key ordering. Do not use [Hashed] with ordered indices when
+/// callers rely on translated-key adjacency matching original-key adjacency (e.g., exclusion proofs
+/// in the ordered QMDB). [Hashed] is safe for unordered indices and partitioned unordered indices.
+///
+/// # Stability
+///
+/// [ahash::RandomState::build_hasher] is used as the underlying hasher. While `ahash` is robust,
+/// its exact algorithm might change across versions. As a result, transformed outputs are not
+/// stable across Rust versions or platforms. Treat this translator as an in-memory
```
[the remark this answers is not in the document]
roberto-bayardo: fixed

storage/src/translator.rs
```diff
@@ -188,6 +188,108 @@ impl<const N: usize> BuildHasher for Cap<N> {
...
+    /// Determinism is scoped to the current `ahash` implementation. Outputs are not guaranteed to
+    /// be stable across crate versions or platforms.
+    pub fn from_seed(seed: u64, inner: T) -> Self {
+        // Derive four independent seeds from the single input using ahash itself.
+        // A fixed RandomState acts as a key-derivation function: hashing (seed, index) pairs
+        // produces well-distributed independent values without requiring std::hash::DefaultHasher.
+        let kdf = ahash::RandomState::with_seeds(0, 0, 0, 0);
+        let derive = |index: u64| -> u64 { kdf.hash_one((seed, index)) };
+        let random_state =
+            ahash::RandomState::with_seeds(derive(0), derive(1), derive(2), derive(3));
```
patrick-ogrady: Can we just do `RandomState::with_seed()`?

https://docs.rs/ahash/0.8.12/ahash/random_state/struct.RandomState.html#method.with_seed

storage/src/translator.rs
```diff
@@ -188,6 +188,108 @@ impl<const N: usize> BuildHasher for Cap<N> {
...
+    }
+}
+
+impl<T: Translator> Translator for Hashed<T> {
+    type Key = T::Key;
+
+    #[inline]
+    fn transform(&self, key: &[u8]) -> T::Key {
+        let hash_val = self.random_state.hash_one(key);
+        self.inner.transform(&hash_val.to_le_bytes())
```
patrick-ogrady, line 274: to_le_bytes should generally be ~slightly faster

storage/src/translator.rs
```diff
@@ -188,6 +188,108 @@ impl<const N: usize> BuildHasher for Cap<N> {
     }
 }
 
+/// Collision-resistant wrapper for any [Translator].
```
patrick-ogrady, line 191: Do we use this in any? Should we add it everywhere we can (any, archive, etc?)

If not, let's add an issue?

roberto-bayardo: just the benchmarks. Should I make it the default to wrap a translator (at least for unordered)?

storage/src/translator.rs
roberto-bayardo: right, since there's no stability requirement seemed the way to go.  Do you want a comment here explaining?

storage/src/translator.rs
roberto-bayardo: yup done

storage/src/translator.rs
patrick-ogrady: Let's leave it unintegrated for now but add an issue to "integrate" without specifying how yet.

storage/src/translator.rs
patrick-ogrady: Nah, I think its fine.

storage/src/translator.rs
```diff
@@ -188,6 +188,101 @@ impl<const N: usize> BuildHasher for Cap<N> {
...
+        }
+    }
+
+    /// Create a new [Hashed] translator with a specific seed.
+    ///
+    /// Determinism is scoped to the current `ahash` implementation. Outputs are not guaranteed to
+    /// be stable across crate versions or platforms.
+    pub fn from_seed(seed: u64, inner: T) -> Self {
+        Self {
+            random_state: ahash::RandomState::with_seed(seed as usize),
```
patrick-ogrady: I didn't realize this was usize 😭

storage/src/translator.rs
patrick-ogrady: https://github.com/commonwarexyz/monorepo/issues/3262

storage/src/translator.rs
```diff
@@ -188,10 +188,107 @@ impl<const N: usize> BuildHasher for Cap<N> {
...
+        Self {
+            random_state: ahash::RandomState::new(),
+            inner,
+        }
+    }
+
+    /// Create a new [Hashed] translator with a specific seed.
+    pub const fn from_seed(seed: u64, inner: T) -> Self {
+        Self {
+            random_state: ahash::RandomState::with_seeds(seed, 0, 0, 0),
```
patrick-ogrady, line 262: `ahash` says this is ok 👍

patrick-ogrady: (the old `with_seed()` method would've actually changed the output per process 👀 )

PR #3260, opened 2026-02-26
[Storage] Add batch API phase 1: implement for MMRs

danlaine:
First phase in a series of PRs (see #3264, #3268) which will replace the dirty/clean type state with a batch API in storage crate types.

## Summary

Introduces a batch API for MMRs — the foundation for speculative execution across the storage crate.

### Batch API (`mmr::batch`, `mmr::read`)

A `Batch` borrows a merkleized MMR immutably and records mutations (appends, leaf updates, pops, pruning) without touching the parent. Multiple batches can coexist on the same parent, and batches can be stacked to arbitrary depth (Base ← A ← B). The lifecycle is:

```
UnmerkleizedBatch → merkleize() → MerkleizedBatch → into_changeset() → Changeset → mmr.apply()
```

`Readable` is a new synchronous, `no_std`-compatible read trait implemented by both base MMRs and `MerkleizedBatch`. `ChainInfo` enables flattening a stack of batches into a single `Changeset` relative to the ultimate base.

### Authenticated journal batching (`journal::authenticated`)

Wraps the MMR batch with item tracking. Each `MerkleizedBatch` holds its items in an `Arc<Vec<Item>>`; `finalize()` collects arcs from the entire chain via the `ItemChain` trait — one refcount bump per level, no deep copies. `Journal::apply_batch()` appends all tracked items and applies the MMR changeset in one operation.

### Bitmap migration (`bitmap::authenticated`)

Migrates the bitmap's inner MMR from dirty/clean type-state to the batch API. The `State` trait no longer carries an `MmrState` associated type — the MMR is always `CleanMmr` and mutations go through a batch during merkleize. Dirty chunk hashing is parallelized when enough chunks are dirty.

### Journal `append(&item)` signature

Changes journal `append` from taking owned items to borrowed. Avoids unnecessary ownership transfer since `append` only needs to encode the item. Touches all journal implementations and callers (consensus, examples, cache, queue).

### DIFF
--- a/storage/src/journal/authenticated.rs
+++ b/storage/src/journal/authenticated.rs
@@ -117,9 +117,6 @@ impl<'a, H: Hasher, P: Readable<H::Digest> + BatchChainInfo<H::Digest>, Item: Se
     fn base_size(&self) -> Position {
         self.inner.base_size()
     }
-    fn retained_size(&self) -> Position {
-        self.inner.retained_size()
-    }
     fn collect_overwrites(&self, into: &mut BTreeMap<Position, H::Digest>) {
         self.inner.collect_overwrites(into);
     }
--- a/storage/src/mmr/batch.rs
+++ b/storage/src/mmr/batch.rs
@@ -2,11 +2,9 @@
 //!
 //! # Overview
 //!
-//! A [`Batch`] borrows a parent MMR (anything implementing [`Readable`])
-//! immutably and records mutations -- appends, leaf updates, pops, and
-//! pruning advances -- without touching the parent. Multiple batches can
-//! coexist on the same parent, and batches can be stacked (Base <- A <- B)
-//! to arbitrary depth.
+//! A [`Batch`] borrows a parent MMR ([`Readable`]) immutably and records mutations -- append
+//! and leaf updates -- without mutating the parent. Multiple batches can coexist on the same
+//! parent, and batches can be stacked (Base <- A <- B <- ...) to arbitrary depth.
 //!
 //! # Lifecycle
 //!
@@ -31,7 +29,7 @@
 //!
 //! # Type aliases
 //!
-//! - [`UnmerkleizedBatch`] -- mutable phase: add, update, pop, advance pruning.
+//! - [`UnmerkleizedBatch`] -- mutable phase: add, update leaves.
 //! - [`MerkleizedBatch`]   -- immutable phase: root is computed, proofs available.
 //! - [`Changeset`]         -- owned delta that can be applied to the base MMR.
 //!
@@ -64,7 +62,6 @@ use crate::mmr::{
 };
 use alloc::{collections::BTreeMap, vec::Vec};
 use commonware_cryptography::Digest;
-use core::cmp;
 cfg_if::cfg_if! {
     if #[cfg(feature = "std")] {
         use commonware_parallel::ThreadPool;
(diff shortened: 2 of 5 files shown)


storage/src/bitmap/authenticated.rs
```diff
@@ -569,30 +567,68 @@ impl<E: Clock + RStorage + Metrics, D: Digest, const N: usize> UnmerkleizedBitMa
...
+                }
+            } else {
+                updates
+                    .iter()
+                    .map(|&(loc, chunk)| {
+                        let pos = Position::try_from(loc).unwrap();
+                        (loc, hasher.leaf_digest(pos, chunk.as_ref()))
+                    })
+                    .collect()
+            }
```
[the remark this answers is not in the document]
roberto-bayardo: seems like a legit simplification find.

storage/src/bitmap/authenticated.rs
```diff
@@ -30,9 +31,13 @@ use commonware_utils::{
     bitmap::{BitMap as UtilsBitMap, Prunable as PrunableBitMap},
     sequence::prefixed_u64::U64,
 };
+use rayon::prelude::*;
 use std::collections::HashSet;
 use tracing::{debug, error, warn};
 
+/// Minimum number of dirty chunks to trigger parallel pre-hashing.
+const MIN_TO_PARALLELIZE: usize = 20;
```
roberto-bayardo: We now have several MIN_TO_PARALLELIZE constants.. should we just export one (maybe from mem-mmr?) and reuse? (I never did tune this value so it's sort of pulled out of thin air.....)

storage/src/mmr/mem.rs
```diff
@@ -66,6 +67,26 @@ pub struct Dirty {
...
+impl Dirty {
+    /// Insert a dirty node. Returns true if newly inserted.
+    pub(crate) fn insert(&mut self, pos: Position, height: u32) -> bool {
+        self.dirty_nodes.insert((pos, height))
+    }
+
+    /// Take all dirty nodes sorted by ascending height (bottom-up for merkleize).
+    pub(crate) fn take_sorted_by_height(&mut self) -> Vec<(Position, u32)> {
+        let mut v: Vec<_> = self.dirty_nodes.iter().copied().collect();
+        self.dirty_nodes.clear();
```
roberto-bayardo: let mut v: Vec<_> = core::mem::take(&mut self.dirty_nodes).into_iter().collect();

storage/src/mmr/mem.rs
roberto-bayardo:
```suggestion
        let nodes = self.state.take_sorted_by_height();
```

roberto-bayardo:
```suggestion
        let nodes = self.state.take_sorted_by_height();
```

storage/src/bitmap/authenticated.rs
danlaine: Done

storage/src/mmr/mem.rs
danlaine: Done

storage/src/mmr/mem.rs
danlaine: Done

danlaine: Done

storage/src/bitmap/authenticated.rs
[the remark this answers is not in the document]
danlaine: fixed

storage/src/mmr/batch.rs
```diff
@@ -0,0 +1,1595 @@
...
+//!                            │
+//!                      mmr.apply(cs)
+//!                            │
+//!                            v
+//!                        CleanMmr          (updated in place)
+//! ```
+//!
+//! # Type aliases
+//!
+//! - [`UnmerkleizedBatch`] -- mutable phase: add, update, pop, advance pruning.
```
roberto-bayardo: I'm wondering if we should remove pop from the batch API. It's mostly used for recovery so it's sort of odd to have it in a batch API.  I think it would be fine to limit pop() to the base DB if it would simplify things (which it seems it would?)

roberto-bayardo: Same with pruning I think?

danlaine: Done

danlaine:
Closing in favor of combined PR #3327

storage/src/mmr/mem.rs
```diff
@@ -402,6 +417,59 @@ impl<D: Digest> CleanMmr<D> {
...
+    /// Create a new [`super::batch::UnmerkleizedBatch`] that borrows this MMR.
+    pub fn new_batch(&self) -> super::batch::UnmerkleizedBatch<'_, D, Self> {
+        super::batch::UnmerkleizedBatch::new(self)
+    }
+
+    /// Apply a changeset produced by [`super::batch::MerkleizedBatch::finalize`].
+    ///
+    /// This is the only way to transfer batch changes into the base MMR.
+    /// After apply, the base's root matches the batch's root.
+    pub fn apply(&mut self, changeset: super::batch::Changeset<D>) {
```
roberto-bayardo, line 430: Codex 5.4 has flagged this as also having the "stale sibling" issue.

storage/src/journal/authenticated.rs
```diff
@@ -308,6 +454,26 @@ where
...
 
+impl<E, C, H> Journal<E, C, H, Clean<H::Digest>>
+where
+    E: Storage + Clock + Metrics,
+    C: Mutable<Item: EncodeShared>,
+    H: Hasher,
+{
+    /// Apply a finalized batch, appending all tracked items from the entire
+    /// chain and then applying the flattened MMR changeset.
+    pub async fn apply_batch(&mut self, batch: Changeset<H::Digest, C::Item>) -> Result<(), Error> {
```
roberto-bayardo, line 465: Should return Error if the batch is "stale" (a sibling batch has already been applied).    I guess it will as long as mmr.apply does?  In any case, let's make sure it does and document in the fn doc about the error condition.

PR #3264, opened 2026-02-27
[Storage] Add batch API phase 2: remove type state

danlaine:
Second phase in a series of PRs (see #3260, #3268) which will replace the dirty/clean type state with a batch API in storage crate types.

## Overview

Removes type-state patterns from the storage crate's MMR and QMDB types, replacing them with a batch API that separates mutation from the base data structure.

**Motivation**: The MMR previously used `CleanMmr`/`DirtyMmr` type states and the QMDB used a 4-state machine (`Merkleized`/`Unmerkleized` x `Durable`/`NonDurable`) with consuming state transitions (`into_mutable()`, `into_merkleized()`, `commit(self) -> (NewState, ...)`). The removal of type state allows for implementation of a batch API for QMDBs in #3268.

### MMR (`mmr::mem::Mmr`)
- `Mmr` is now always merkleized. Removed `DirtyMmr`/`CleanMmr` type aliases and the `S: State` type parameter.
- Mutations go through a new batch API: `mmr.new_batch()` returns an `UnmerkleizedBatch` that borrows the MMR immutably. After accumulating changes, call `.merkleize().finalize()` to produce an owned `Changeset`, then `mmr.apply(changeset)`.
- Batches can be stacked (batch-on-batch) and support `into_dirty()` for re-dirtying after merkleization.
- New `mmr::batch` module (`batch.rs`) with `UnmerkleizedBatch`, `MerkleizedBatch`, and `Changeset` types.
- New `mmr::read` module with `Readable` and `BatchChainInfo` traits for uniform node access across MMR and batch layers.
- `mmr::mem::Mmr::from_components` now returns `Result<Self, Error>` instead of panicking on invalid input.
- Removed `Error::Unmerkleized`.

### MMR Journaled (`mmr::journaled::Mmr`)
- Same batch API: `new_batch()` / `apply()` instead of direct mutation + `merkleize()`.
- `sync()` is now `&mut self` instead of a state transition.

### QMDB Authenticated Databases (`qmdb::any`, `qmdb::current`)
- Collapsed the 4-state machine into a single type per database. `init()` returns a database that supports reads, writes, authentication, pruning, and persistence directly.
- `commit(&mut self)` replaces the consuming `commit(self) -> (Durable, Range)` pattern. Returns `Result<Range<Location>>`.
- Removed `into_mutable()`, `into_merkleized()`, and all intermediate state types.
- Removed `Merkleized`, `Unmerkleized`, `Durable`, `NonDurable` type aliases from `qmdb::mod`.
- Replaced 4 test/bench traits (`CleanAny`, `MutableAny`, `UnmerkleizedDurableAny`, `MerkleizedNonDurableAny`) with a single `DbAny` trait.

### QMDB Store (`qmdb::store`)
- Removed `State` trait, `Durable`/`NonDurable` marker types, and the `S: State` type parameter from `store::db::Db`.
- `commit(&mut self)` replaces `commit(self) -> (Db<..., Durable>, Range)`.

### Downstream
- Updated `examples/sync` server to use `&mut` borrows instead of `Option<DB>` + take/put pattern.
- Updated all fuzz targets and benchmarks for the new APIs.

### DIFF
--- a/storage/fuzz/fuzz_targets/qmdb_immutable.rs
+++ b/storage/fuzz/fuzz_targets/qmdb_immutable.rs
@@ -122,7 +122,6 @@ fn fuzz(input: FuzzInput) {
         let mut keys_set = Vec::new();
         let mut set_locations = Vec::new(); // Track locations that contain Set operations
         let mut last_commit_loc = None;
-        let mut uncommitted_ops = Vec::new();
 
         for op in input.operations {
             match op {
@@ -138,7 +137,6 @@ fn fuzz(input: FuzzInput) {
                         if let Ok(()) = db.set(key, value.clone()).await {
                             keys_set.push((key, loc));
                             set_locations.push((key, loc));
-                            uncommitted_ops.push((key, loc));
                         }
                     }
                 }
@@ -160,7 +158,6 @@ fn fuzz(input: FuzzInput) {
 
                     let _ = db.commit(metadata).await.unwrap();
                     last_commit_loc = Some(db.bounds().await.end - 1);
-                    uncommitted_ops.clear();
                 }
 
                 ImmutableOperation::Prune { loc } => {
@@ -169,7 +166,7 @@ fn fuzz(input: FuzzInput) {
                         let safe_loc = Location::new(safe_loc);
                         let _ = db.commit(None).await.unwrap();
                         last_commit_loc = Some(db.bounds().await.end - 1);
-                        uncommitted_ops.clear();
+
                         db.prune(safe_loc).await.expect("prune should not fail");
                         let oldest = db.bounds().await.start;
                         set_locations.retain(|(_, l)| *l >= oldest);
@@ -187,9 +184,6 @@ fn fuzz(input: FuzzInput) {
                         let safe_start = Location::new(safe_start);
                         let safe_max_ops =
                             NonZeroU64::new((max_ops % MAX_PROOF_OPS).max(1)).unwrap();
-                        let _ = db.commit(None).await.unwrap();
-                        last_commit_loc = Some(db.bounds().await.end - 1);
-                        uncommitted_ops.clear();
                         if let Ok((proof, ops)) = db.proof(safe_start, safe_max_ops).await {
                             let root = db.root();
                             let _ = verify_proof(&mut hasher, &proof, safe_start, &ops, &root);
@@ -203,7 +197,7 @@ fn fuzz(input: FuzzInput) {
                     max_ops,
                 } => {
                     let op_count = db.bounds().await.end;
-                    if op_count > 0 && uncommitted_ops.is_empty() {
+                    if op_count > 0 {
                         let safe_size = (size % op_count.as_u64()).max(1);
                         let safe_size = Location::new(safe_size);
                         let safe_start = start_loc % safe_size.as_u64();
@@ -211,8 +205,6 @@ fn fuzz(input: FuzzInput) {
                         let safe_max_ops =
                             NonZeroU64::new((max_ops % MAX_PROOF_OPS).max(1)).unwrap();
 
-                        let _ = db.commit(None).await.unwrap();
-                        last_commit_loc = Some(db.bounds().await.end - 1);
                         if safe_start >= db.bounds().await.start {
                             let _ = db
                                 .historical_proof(safe_size, safe_start, safe_max_ops)
@@ -234,9 +226,6 @@ fn fuzz(input: FuzzInput) {
                 }
 
                 ImmutableOperation::Root => {
-                    let _ = db.commit(None).await.unwrap();
-                    last_commit_loc = Some(db.bounds().await.end - 1);
-                    uncommitted_ops.clear();
                     let _ = db.root();
                 }
             }
--- a/storage/fuzz/fuzz_targets/qmdb_keyless.rs
+++ b/storage/fuzz/fuzz_targets/qmdb_keyless.rs
@@ -205,7 +205,6 @@ fn fuzz(input: FuzzInput) {
                 }
 
                 Operation::Root => {
-                    let _ = db.commit(None).await.expect("Commit should not fail");
                     let _ = db.root();
                 }
 
@@ -217,7 +216,6 @@ fn fuzz(input: FuzzInput) {
                     if op_count == 0 {
                         continue;
                     }
-                    let _ = db.commit(None).await.expect("Commit should not fail");
                     let start_loc = (*start_offset as u64) % op_count.as_u64();
                     let max_ops_value = ((*max_ops as u64) % MAX_PROOF_OPS) + 1;
                     let start_loc = Location::new(start_loc);
@@ -239,7 +237,6 @@ fn fuzz(input: FuzzInput) {
                     if op_count == 0 {
                         continue;
                     }
-                    let _ = db.commit(None).await.expect("Commit should not fail");
                     let size = ((*size_offset as u64) % op_count.as_u64()) + 1;
                     let size = Location::new(size);
                     let start_loc = (*start_offset as u64) % *size;


storage/src/qmdb/any/db.rs
```diff
@@ -113,83 +108,14 @@ where
...
-    E: Storage + Clock + Metrics,
-    K: Key,
-    V: ValueEncoding,
-    U: Update<K, V>,
-    C: Mutable<Item = Operation<K, V, U>>,
-    I: UnorderedIndex<Value = Location>,
-    H: Hasher,
-    Operation<K, V, U>: Codec,
-{
     pub fn root(&self) -> H::Digest {
```
roberto-bayardo, line 112: If there are uncommitted operations does this root reflect them or is the root cached as of last commit point?

storage/src/qmdb/any/db.rs
```diff
@@ -591,11 +440,17 @@ where
     C: Mutable<Item = Operation<K, V, U>>,
     I: UnorderedIndex<Value = Location>,
     H: Hasher,
-    D: DurabilityState,
     Operation<K, V, U>: Codec,
 {
     async fn prune(&mut self, prune_loc: Location) -> Result<(), Error> {
-        self.prune(prune_loc).await
```
roberto-bayardo, line 598: why can't you call self.prune ? (I see an impl above ^^^)

storage/src/qmdb/any/traits.rs
```diff
@@ -1,5 +1,4 @@
-//! Traits representing the 4 possible states of an Any database. These are used to share test and
-//! benchmark code across the variants.
+//! Trait providing a unified test/benchmark interface across all Any database variants.
```
roberto-bayardo, line 1: I guess we should rename this file since there are no more "states"?

storage/src/qmdb/any/db.rs
danlaine: The returned root reflects uncommitted operations but in #3268 the API changes such that this will always be the root at a commit.

storage/src/qmdb/any/db.rs
danlaine: Fixed

danlaine: Ah, this is why they were separate: https://github.com/commonwarexyz/monorepo/actions/runs/22687160516/job/65773093773

storage/src/qmdb/any/traits.rs
danlaine: Renamed to `traits.rs` -- open to other names

storage/src/qmdb/any/db.rs
roberto-bayardo: This worked before though?  Not seeing what's changed to prevent this.

danlaine: I split out the inherent `prune` method to be defined when `C: Mutable<Item = Operation<K, V, U>>` instead of `C: Mutable<Item = Operation<K, V, U>> + Persistable<Error = JournalError>`, so this is only defined once now :+1:

roberto-bayardo: not following... looks like prune() used to be defined only once and now the code is duplicated here?

danlaine: Are you looking at the latest diff? The two implementations now are the inherent impl:

```rs
impl<E, K, V, U, C, I, H> Db<E, C, I, H, U>
where
    E: Storage + Clock + Metrics,
    K: Key,
    V: ValueEncoding,
    U: Update<K, V>,
    C: Mutable<Item = Operation<K, V, U>>,
    I: UnorderedIndex<Value = Location>,
    H: Hasher,
    Operation<K, V, U>: Codec,
{
    /// Prunes historical operations prior to `prune_loc`. This does not affect the db's root or
    /// snapshot.
    ///
    /// # Errors
    ///
    /// - Returns [Error::PruneBeyondMinRequired] if `prune_loc` > inactivity floor.
    /// - Returns [crate::mmr::Error::LocationOverflow] if `prune_loc` > [crate::mmr::MAX_LOCATION].
    pub async fn prune(&mut self, prune_loc: Location) -> Result<(), Error> {
        if prune_loc > self.inactivity_floor_loc {
            return Err(Error::PruneBeyondMinRequired(
                prune_loc,
                self.inactivity_floor_loc,
            ));
        }

        self.log.prune(prune_loc).await?;

        Ok(())
    }
}
```

and the `PrunableStore` impl, which just calls the former:

```rs
impl<E, K, V, U, C, I, H> PrunableStore for Db<E, C, I, H, U>
where
    E: Storage + Clock + Metrics,
    K: Key,
    V: ValueEncoding,
    U: Update<K, V>,
    C: Mutable<Item = Operation<K, V, U>>,
    I: UnorderedIndex<Value = Location>,
    H: Hasher,
    Operation<K, V, U>: Codec,
{
    async fn prune(&mut self, prune_loc: Location) -> Result<(), Error> {
        self.prune(prune_loc).await
    }
    ...
```

roberto-bayardo: guess I was on the older diff. 🤦   thanks.

storage/fuzz/fuzz_targets/qmdb_immutable.rs
```diff
@@ -191,17 +184,10 @@ fn fuzz(input: FuzzInput) {
...
-                        let (durable_db, _) = db.commit(None).await.unwrap();
-                        last_commit_loc = Some(durable_db.bounds().await.end - 1);
-                        uncommitted_ops.clear();
-                        let merkleized_db = durable_db.into_merkleized();
-                        if let Ok((proof, ops)) =
-                            merkleized_db.proof(safe_start, safe_max_ops).await
-                        {
-                            let root = merkleized_db.root();
+                        if let Ok((proof, ops)) = db.proof(safe_start, safe_max_ops).await {
+                            let root = db.root();
```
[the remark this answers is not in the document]
danlaine: > so proofs may be generated against uncommitted state

cursor bot is wrong, the API doesn't support this.

storage/fuzz/fuzz_targets/qmdb_immutable.rs
```diff
@@ -211,21 +197,19 @@ fn fuzz(input: FuzzInput) {
...
                         let safe_start = start_loc % safe_size.as_u64();
                         let safe_start = Location::new(safe_start);
                         let safe_max_ops =
                             NonZeroU64::new((max_ops % MAX_PROOF_OPS).max(1)).unwrap();
 
-                        let merkleized_db = db.into_merkleized();
-                        if safe_start >= merkleized_db.bounds().await.start {
-                            let _ = merkleized_db
+                        if safe_start >= db.bounds().await.start {
+                            let _ = db
```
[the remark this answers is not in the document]
danlaine: >  allowing historical proofs to be attempted with uncommitted state

This is wrong

storage/src/qmdb/store/db.rs
```diff
@@ -420,20 +378,16 @@ where
...
         // previous commit becoming inactive.
         if self.is_empty() {
             self.inactivity_floor_loc = self.size().await;
             debug!(tip = ?self.inactivity_floor_loc, "db is empty, raising floor to tip");
         } else {
-            let steps_to_take = self.state.steps + 1;
+            let steps_to_take = self.steps + 1;
             for _ in 0..steps_to_take {
                 let loc = self.inactivity_floor_loc;
                 self.inactivity_floor_loc = self.as_floor_helper().raise_floor(loc).await?;
```
clabby, line 393: Testing a new review agent topology, and it called this out - may be useful. I know we currently treat DB errors as fatal, but seems a bit inconsistent since we don’t panic.

---

`commit(&mut self)` mutates `self.inactivity_floor_loc` in-place during floor-raising, but `self.steps` is only reset after `self.log.commit()` succeeds. If any awaited call in the raise/append/commit path returns an error, the instance retains a partially advanced floor and stale `steps`. A subsequent retry then over-advances the inactivity floor schedule relative to a clean run — a regression introduced by changing from a consuming `commit(self)` to a reusable `commit(&mut self)`.

**Fix**: Stage floor/steps changes in locals and publish them only after the durable commit succeeds, or restore pre-commit state on any error path.

```rust
#[test]
fn repro_store_commit_retry_over_advances_floor_after_error() {
        // build identical control/candidate stores and perform one committed insert + one update
        // ...
        *context.storage_fault_config().write() = FaultConfig::default().sync(1.0);
        assert!(Db::commit(&mut candidate, None).await.is_err());

        *context.storage_fault_config().write() = FaultConfig::default();
        Db::commit(&mut control, None).await.unwrap();
        Db::commit(&mut candidate, None).await.unwrap();

        assert_eq!(candidate.inactivity_floor_loc(), control.inactivity_floor_loc());
    });
}
```

Expected: floor locations match. Observed: assertion failure — `left: Location(9)` vs `right: Location(6)`.

storage/src/mmr/journaled.rs
```diff
@@ -1353,50 +1161,60 @@ mod tests {
     }
 
     #[test_traced]
-    fn test_journaled_mmr_pop_error_clamps_merkleized_size() {
+    fn test_journaled_mmr_rewind_error_leaves_valid_state() {
```
clabby, line 1164: also from the review agent

---

`test_journaled_mmr_rewind_error_leaves_valid_state` only exercises pre-I/O error branches (`ElementPruned`/`Empty`), both returned before `journal.rewind` or `journal.sync` are called. The actual ordering risk — `journal.rewind` succeeds then `journal.sync` fails (lines 791-792) — is not covered, so a regression in that window would not be caught.

Add a fault-injected test that forces a post-`journal.rewind` sync failure and asserts a clear invariant (e.g. poisoned-instance semantics, or that in-memory state and journal remain consistent):

```rust
#[test_traced]
fn test_journaled_mmr_rewind_sync_fault_invariant() {
    let cfg = deterministic::Config::default();
    let executor = deterministic::Runner::new(cfg);
        let mut hasher = Standard::<Sha256>::new();
        let mut mmr = Mmr::init(context.clone(), &mut hasher, test_config(&context)).await.unwrap();

        let changeset = {
            let mut batch = mmr.new_batch();
            for i in 0..64 {
                batch.add(&mut hasher, &test_digest(i));
            }
            batch.merkleize(&mut hasher).finalize()
        };
        mmr.apply(changeset);
        mmr.sync().await.unwrap();

        // Force sync failure during rewind.
        context.storage_fault_config().write().sync_rate = Some(1.0);
        let _ = mmr.rewind(1, &mut hasher).await;

        // Assert intended invariant (journal size == in-memory size).
        assert_eq!(Position::new(mmr.journal.size().await), mmr.size());
    });
}
```

danlaine:
Closing in favor of combined PR https://github.com/commonwarexyz/monorepo/pull/3327

PR #3267, opened 2026-02-27
[consensus/simplex] Remove Initialization timeout trigger

joshieDo:
Let the voter progress through natural deadline-based timeouts after journal replay instead of forcing an immediate timeout.

### DIFF
--- a/consensus/src/simplex/mod.rs
+++ b/consensus/src/simplex/mod.rs
@@ -4665,10 +4665,11 @@ mod tests {
         attributable_reporter_filtering::<_, _, RoundRobin>(secp256r1::fixture);
     }
 
-    fn split_views_no_lockup<S, F>(mut fixture: F)
+    fn split_views_no_lockup<S, F, L>(mut fixture: F)
     where
         S: Scheme<Sha256Digest, PublicKey = PublicKey>,
         F: FnMut(&mut deterministic::Context, &[u8], u32) -> Fixture<S>,
+        L: Elector<S>,
     {
         // Scenario:
         // - View F: Finalization of B_1 seen by all participants.
@@ -4841,7 +4842,7 @@ mod tests {
             // Start engines after preloading certificates into each participant's
             // recovered channel (ensuring processing before any leader attempts to issue a
             // conflicting vote).
-            let elector = RoundRobin::<Sha256>::default();
+            let elector = L::default();
             let relay = Arc::new(mocks::relay::Relay::new());
             let mut honest_reporters = Vec::new();
             for (idx, validator) in participants.iter().enumerate() {
@@ -5021,14 +5022,14 @@ mod tests {
 
     #[test_traced]
     fn test_split_views_no_lockup() {
-        split_views_no_lockup(bls12381_threshold_vrf::fixture::<MinPk, _>);
-        split_views_no_lockup(bls12381_threshold_vrf::fixture::<MinSig, _>);
-        split_views_no_lockup(bls12381_threshold_std::fixture::<MinPk, _>);
-        split_views_no_lockup(bls12381_threshold_std::fixture::<MinSig, _>);
-        split_views_no_lockup(bls12381_multisig::fixture::<MinPk, _>);
-        split_views_no_lockup(bls12381_multisig::fixture::<MinSig, _>);
-        split_views_no_lockup(ed25519::fixture);
-        split_views_no_lockup(secp256r1::fixture);
+        split_views_no_lockup::<_, _, Random>(bls12381_threshold_vrf::fixture::<MinPk, _>);
+        split_views_no_lockup::<_, _, Random>(bls12381_threshold_vrf::fixture::<MinSig, _>);
+        split_views_no_lockup::<_, _, RoundRobin>(bls12381_threshold_std::fixture::<MinPk, _>);
+        split_views_no_lockup::<_, _, RoundRobin>(bls12381_threshold_std::fixture::<MinSig, _>);
+        split_views_no_lockup::<_, _, RoundRobin>(bls12381_multisig::fixture::<MinPk, _>);
+        split_views_no_lockup::<_, _, RoundRobin>(bls12381_multisig::fixture::<MinSig, _>);
+        split_views_no_lockup::<_, _, RoundRobin>(ed25519::fixture);
+        split_views_no_lockup::<_, _, RoundRobin>(secp256r1::fixture);
     }
 
     fn tle<V, L>()


consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -813,13 +813,8 @@ impl<
             ?elapsed,
             "consensus initialized"
         );
-        self.state
-            .trigger_timeout(observed_view, TimeoutReason::Initialization);
 
         // Initialize batcher with leader for current view
-        //
-        // We don't worry about sending any constructed messages here because we expect the view to immediately timeout
-        // and we'll send our nullify vote shortly.
```
joshieDo, line 822: should we worry now ?

patrick-ogrady: If we constructed, we already sent before. So, I'd argue no need to add more complexity here.

patrick-ogrady: (and on fresh epoch, there is nothing yet constructed)

patrick-ogrady:
Thought about this a good bit and I believe it is better to potentially timeout late (code now) rather than early (code on main) after restart.

patrick-ogrady:
Summary of test changes (and why they were required):

```
  test_invalid

  The byzantine node (index 0) uses a wrong-namespace scheme, so its batcher blocks all honest peers as "unknown participant." It never receives
  proposals or certificates — the only vote it can produce is a nullify on leader timeout. With leader_timeout = 1s, that nullify arrives at
  honest nodes ~1010ms later. By then, honest nodes have advanced ~25 views. The interesting() filter computes min_active = max(0, 25 - 10) = 15,
   and since the nullify is for view 1 (1 < 15), it's silently dropped. The invalid signature never reaches the reporter, so invalid_count stays
  at 0.

  Previously, the initialization trigger forced an immediate nullify at t=0, so it arrived within ~10ms — well within the activity window.

  Fix: Give the byzantine node a shorter leader_timeout (100ms) so its nullify arrives while view 1 is still within the activity window of honest
   nodes. Honest nodes keep 1s to preserve original test dynamics.

  test_split_views_no_lockup

  The test injects pre-built certificates for views 1-3 before linking validators. Without the initialization trigger, the voter is free to
  propose in view 1 if it's the elected leader. With epoch=333 and n=10, the standard round-robin formula (333 + 1) % 10 = 4 makes participant 4
  (honest, Group2) the leader. Its voter starts, calls try_propose(), and begins building a proposal with a random payload. When the injected
  certificate arrives with a different payload, slot.update() detects a conflict and block_equivocator blocks participant 4 — an honest node
  blocking itself.

  Previously, the initialization trigger set broadcast_nullify = true, which prevented propose_ready() from returning Some, so no proposal was
  built before certificates arrived.

  Fix: Use a FixedElector that elects a Byzantine node (index 9) as leader for views 1-3 (the views with injected certificates). Byzantine nodes
  are NullifyOnly — they never propose, so no race occurs. Views after 3 fall back to standard round-robin so honest nodes can drive progress.
```

in short, I embrace our new robot overlords.

consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -813,13 +813,8 @@ impl<
             ?elapsed,
             "consensus initialized"
         );
-        self.state
```
patrick-ogrady, line 816: Actual change 👀

consensus/src/simplex/actors/voter/mod.rs
```diff
@@ -4848,28 +4849,52 @@ mod tests {
...
+                        Some(MailboxMessage::Certified { .. } | MailboxMessage::Certificate(_)) => {}
+                        None => panic!(
+                            "resolver channel closed before certification for restarted view {target_view}"
+                        ),
                     },
                     msg = batcher_receiver.recv() => {
-                        if let batcher::Message::Update { response, .. } = msg.unwrap() {
-                            response.send(None).unwrap();
+                        match msg {
+                            Some(batcher::Message::Constructed(Vote::Nullify(nullify)))
```
patrick-ogrady: Increase certify latency in test?

consensus/src/simplex/actors/voter/mod.rs
```diff
@@ -4848,28 +4849,52 @@ mod tests {
...
-                        MailboxMessage::Certified { view, success } if view == target_view => {
-                            break success;
+                    msg = resolver_receiver.recv() => match msg {
+                        Some(MailboxMessage::Certified { view, success }) if view == target_view => {
+                            assert!(success, "expected successful certification after restart for canceled certification view");
+                            break;
                         }
-                        MailboxMessage::Certified { .. } | MailboxMessage::Certificate(_) => {}
+                        Some(MailboxMessage::Certified { .. } | MailboxMessage::Certificate(_)) => {}
+                        None => panic!(
```
patrick-ogrady: just .unwrap on msg?

consensus/src/simplex/actors/voter/mod.rs
```diff
@@ -4848,28 +4849,52 @@ mod tests {
...
+                            "resolver channel closed before certification for restarted view {target_view}"
+                        ),
                     },
                     msg = batcher_receiver.recv() => {
-                        if let batcher::Message::Update { response, .. } = msg.unwrap() {
-                            response.send(None).unwrap();
+                        match msg {
```
patrick-ogrady: just unwrap?

consensus/src/simplex/mod.rs
```diff
@@ -5022,14 +5017,81 @@ mod tests {
...
+                        (round.epoch().get().wrapping_add(round.view().get())) as u32 % self.n;
+                    Participant::new(idx)
+                }
+            }
+        }
+
+        // Elect a Byzantine node (index 9) as leader for views 1-3 so that
+        // honest voters do not propose before the injected certificates
+        // (F=1, F+1=2, F+2=3) arrive.
+        let elector = FixedElector {
```
patrick-ogrady: I'm trying to find a way to remove this workaround (although it does currently work).

consensus/src/simplex/mod.rs
```diff
@@ -5022,14 +5017,81 @@ mod tests {
...
+        }
+
+        impl Default for FixedElector {
+            fn default() -> Self {
+                Self {
+                    leader: Participant::new(0),
+                    through: View::new(1),
+                }
+            }
+        }
```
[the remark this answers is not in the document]
patrick-ogrady: The `Default` is currently required for the `Elector` trait.

patrick-ogrady:
bugbot run

patrick-ogrady:
bugbot run

patrick-ogrady:
bugbot run

patrick-ogrady (approved):

consensus/src/simplex/mod.rs
```diff
@@ -2939,6 +2939,13 @@ mod tests {
                 );
                 actor.start();
                 let blocker = oracle.control(validator.clone());
+                let leader_timeout = if idx_scheme == 0 {
+                    // Force the wrong-namespace byzantine node to timeout quickly so
+                    // honest nodes deterministically observe at least one invalid vote.
+                    Duration::from_millis(100)
```
patrick-ogrady, line 2945: Since we don't "insta-timeout", we need to ensure the byzantine node messages are processed (if we send at the same time as other nodes, they may not be checked...it was blind luck this worked before on the seed we picked)!

patrick-ogrady: I suppose just randomly tweaking some seeds could already serve as a good mutation test @clabby ?

consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -813,20 +813,19 @@ impl<
...
         // Initialize batcher with leader for current view
-        //
-        // We don't worry about sending any constructed messages here because we expect the view to immediately timeout
-        // and we'll send our nullify vote shortly.
         let leader = self
             .state
             .leader_index(observed_view)
             .expect("leader not set");
-        batcher
+        if let Some(reason) = batcher
```
patrick-ogrady, line 822: This was ~unlikely before but technically a bug not to handle this early nullify trigger.

consensus/src/simplex/mod.rs
```diff
@@ -4720,89 +4726,6 @@ mod tests {
             } = fixture(&mut context, &namespace, n);
             let mut registrations = register_validators(&mut oracle, &participants).await;
 
-            // ========== Create engines ==========
-
-            // Do not link validators yet; we will inject certificates first, then link everyone.
-
-            // Create engines: 7 honest engines, 3 byzantine
```
patrick-ogrady, line 4736: It was also a fluke that this worked. I moved some code around to make this more robust across seeds.

consensus/src/simplex/mod.rs
```diff
@@ -4658,11 +4665,10 @@ mod tests {
         attributable_reporter_filtering::<_, _, RoundRobin>(secp256r1::fixture);
     }
 
-    fn split_views_no_lockup<S, F, L>(mut fixture: F)
+    fn split_views_no_lockup<S, F>(mut fixture: F)
```
clabby: Not super important, but why did we remove the `Random` elector coverage from this test?

clabby (dismissed):
tricky, but lgtm

consensus/src/simplex/mod.rs
patrick-ogrady: This was done part way through when we switched to a fixed elector but I found a cleaner fix (and we could now support again).

PR #3271, opened 2026-02-28
fix(codec): return InvalidVarint instead of panic on excess continuation bytes

erenyegit:
## Problem

`varint::Decoder::feed` could panic when the decoder received more continuation bytes than fit in the target type. For example, for `Decoder::<u32>` (max 5 bytes), feeding a 6th byte with the continuation bit set caused `max_bits.checked_sub(self.bits_read)` to return `None`, and `.unwrap()` panicked.

This is a robustness issue: malformed or adversarial input could crash the process instead of returning an error.

## Solution

Replace the `.unwrap()` with `.ok_or(Error::InvalidVarint(U::SIZE))?`, so excess continuation bytes produce a proper `Error::InvalidVarint` instead of a panic. The decoder already uses this error variant for other invalid varint cases.

## Testing

- Added unit test `decoder_rejects_excess_continuation_bytes` that feeds 6 continuation bytes to `Decoder::<u32>` and asserts an `InvalidVarint(4)` error is returned.
- `cargo test -p commonware-codec` passes (110 tests + doc tests).

### DIFF
--- a/codec/src/varint.rs
+++ b/codec/src/varint.rs
@@ -133,7 +133,10 @@ impl<U: UPrim> Decoder<U> {
         // If we have reached what must be the last byte, this check prevents continuing to read
         // from the buffer by ensuring that the conditional (`if byte & CONTINUATION_BIT_MASK == 0`)
         // always evaluates to true.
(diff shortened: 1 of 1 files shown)


clabby:

PR #3273, opened 2026-02-28
[runtime/network] support vectored send

### LINKED ISSUE #2705: [runtime] io_uring network vectored writes (writev)

### LINKED ISSUE #3269: [runtime] support vectored writes in network backends
With #3185 supporting vectored writes in network will allow us to avoid materialization of pages into a single contiguous buffer.  

Related #2883.

andresilva:
Fixes #2705.
Fixes #3269.
Related #2883.

### DIFF
--- a/runtime/src/network/iouring.rs
+++ b/runtime/src/network/iouring.rs
@@ -41,6 +41,7 @@ use tracing::warn;
 
 /// Default read buffer size (64 KB).
 const DEFAULT_READ_BUFFER_SIZE: usize = 64 * 1024;
+
 /// Cap iovec batch size: larger iovecs reduce syscall count but increase
 /// per-write kernel setup overhead.
 const IOVEC_BATCH_SIZE: usize = 32;


runtime/src/network/iouring.rs
```diff
@@ -41,6 +41,9 @@ use tracing::warn;
 
 /// Default read buffer size (64 KB).
 const DEFAULT_READ_BUFFER_SIZE: usize = 64 * 1024;
+/// Cap iovec batch size: larger iovecs reduce syscall count but increase
```
patrick-ogrady, line 45: Should there be a \n above this? I think we usually separate consts?

runtime/src/network/iouring.rs
```diff
@@ -178,18 +181,16 @@ impl crate::Network for Network {
...
 
         // Explicitly set non-blocking mode to true
         stream
             .set_nonblocking(true)
-            .map_err(|_| crate::Error::ConnectionFailed)?;
+            .map_err(|_| Error::ConnectionFailed)?;
 
         let fd = Arc::new(OwnedFd::from(stream));
         Ok((
-            Sink::new(fd.clone(), self.send_submitter.clone(), self.pool.clone()),
```
patrick-ogrady, line 192: ❤️

patrick-ogrady (approved):

PR #3276, opened 2026-03-01
[runtime/iouring] round config size to the next power of two

andresilva:
This PR aligns io_uring capacity semantics by normalizing `Config::size` to the next power of two when the loop is constructed. Previously, non-power-of-two values could lead to a mismatch where the kernel ring was effectively larger than the runtime's waiter/channel limits. With this change, configured in-flight capacity matches the effective ring sizing behavior.

Related #2883.

### DIFF
--- a/runtime/src/iouring/mod.rs
+++ b/runtime/src/iouring/mod.rs
@@ -720,7 +720,7 @@ impl IoUringLoop {
         cfg.size = cfg
             .size
             .checked_next_power_of_two()
-            .expect("ring size rounds beyond u32");
+            .expect("ring size exceeds u32::MAX");
         let size = cfg.size as usize;
         let metrics = Arc::new(Metrics::new(registry));
         let (sender, receiver) = mpsc::channel(size);


runtime/src/iouring/mod.rs
```diff
@@ -712,7 +716,11 @@ impl IoUringLoop {
     /// Create a new io_uring loop and submit handle.
     ///
     /// The loop allocates its own metrics, operation channel, and internal `eventfd` wake source.
-    pub(crate) fn new(cfg: Config, registry: &mut Registry) -> (Submitter, Self) {
+    pub(crate) fn new(mut cfg: Config, registry: &mut Registry) -> (Submitter, Self) {
+        cfg.size = cfg
+            .size
+            .checked_next_power_of_two()
+            .expect("ring size rounds beyond u32");
         let size = cfg.size as usize;
```
patrick-ogrady, line 724: Can we condense these two lines?

runtime/src/iouring/mod.rs
```diff
@@ -712,7 +716,11 @@ impl IoUringLoop {
     /// Create a new io_uring loop and submit handle.
     ///
     /// The loop allocates its own metrics, operation channel, and internal `eventfd` wake source.
-    pub(crate) fn new(cfg: Config, registry: &mut Registry) -> (Submitter, Self) {
+    pub(crate) fn new(mut cfg: Config, registry: &mut Registry) -> (Submitter, Self) {
+        cfg.size = cfg
+            .size
+            .checked_next_power_of_two()
+            .expect("ring size rounds beyond u32");
```
patrick-ogrady: `ring size exceeds u32::MAX`?

andresilva:
```suggestion
            .expect("ring size exceeds u32::MAX");
```

patrick-ogrady (approved):

PR #3279, opened 2026-03-01
[deployer]: remove unbounded layered S3 retries

GarmashAlex:
## Why were these changes needed?
S3 uploads had two unbounded retry layers at the same time:
1) SDK retry config used `with_max_attempts(u32::MAX)`,
2) [upload_with_retry](https://github.com/commonwarexyz/monorepo/blob/1386422d7c1738b24f55389d5ee331ac064f6203/deployer/src/aws/s3.rs#L204-L232) also had an infinite manual retry loop.

This made failures hard to diagnose and could cause very long or effectively unbounded wait times.

## What does this change do?
- Caps S3 SDK retries to a finite value (`max_attempts = 5`).
- Removes the manual infinite retry loop from [upload_with_retry](https://github.com/commonwarexyz/monorepo/blob/1386422d7c1738b24f55389d5ee331ac064f6203/deployer/src/aws/s3.rs#L204-L232).
- Keeps retry responsibility in one place (AWS SDK policy).
- Propagates upload errors to callers instead of retrying forever at the wrapper level.

### DIFF
--- a/deployer/src/aws/s3.rs
+++ b/deployer/src/aws/s3.rs
@@ -93,7 +93,7 @@ pub const WGET: &str =
 /// Creates an S3 client for the specified AWS region
 pub async fn create_client(region: Region) -> S3Client {
     let retry = aws_config::retry::RetryConfig::adaptive()
-        .with_max_attempts(u32::MAX)
+        .with_max_attempts(5)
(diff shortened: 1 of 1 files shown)


clabby:

PR #3281, opened 2026-03-02
[consensus/simplex] Strengthen twins framework

clabby:
## Overview

Upgrades the simplex twins harness to follow the campaign model from Twins: BFT Systems Made Robust (https://arxiv.org/abs/2004.10617, especially Section 5.2), replacing the previous hand-picked strategy tests with a deterministic scenario generator and executor that systematically explores adversarial schedules.

At generation time, the framework now builds partition scenarios first, then leader-plus-partition round scenarios, then multi-round arrangements, and finally executes the Cartesian product of arrangements and compromised-node assignments. This preserves deterministic reproducibility while materially increasing adversarial schedule coverage over the old fixed/alternating/random strategy families.

For partition generation, the harness now supports configurable partition cardinality via max_partitions (currently effective up to three buckets, and bounded by participant count) instead of hard-coding only canonical two-way splits. Round scenarios are derived from these partitions and combine a designated leader with explicit twin routing masks, allowing each round to model richer communication asymmetries. Routing now explicitly returns SplitTarget::None when a sender is outside both selected recipient groups, making sparse partition scenarios representable without panicking.

For arrangement generation (Step-3), the harness now enumerates the full round-wise arrangement space when feasible and applies deterministic pruning by sampling unique arrangements when bounded by max_scenarios. This removes the previous lexicographic-prefix bias where early rounds could remain effectively fixed under small caps, and ensures pruned campaigns still vary across round positions.

Compromised-node selection remains deterministic and now composes with the richer scenario space: campaigns execute all generated scenario/compromised-set combinations (subject to caps), and seeds are derived per case for reproducible replay.

Execution semantics were also tightened. Twins use scenario-scripted leaders during adversarial rounds, then transition into a synchronous all-to-all suffix while leader selection falls back to the configured elector. Resolver traffic is currently disabled in the campaign/fuzzer via inert channels rather than delivered to both twins.

Validation checks were strengthened around honest replicas: liveness waits only on honest reporters, safety asserts no conflicting finalizations among honest replicas, invalid-signature counters must remain zero, and detected faults must attribute to twin identities. This makes campaign outcomes easier to interpret and aligns checks with adversarial accountability goals.

### DIFF
--- a/consensus/src/simplex/mocks/twins.rs
+++ b/consensus/src/simplex/mocks/twins.rs
@@ -392,6 +392,16 @@ fn next_round_transitions(cells: &[usize]) -> Vec<(RoundScenario, Vec<usize>)> {
         leader: Option<usize>,
     }
 
+    const LEADER_PLACEMENTS: &[(bool, bool)] = &[(true, false), (false, true), (true, true)];
+
+    fn push_nonzero_cells(next_cells: &mut Vec<usize>, sizes: [usize; 4]) {
+        for size in sizes {
+            if size > 0 {
+                next_cells.push(size);
+            }
+        }
+    }
+
     /// Refines cells for a round where both halves have identical recipient
     /// sets.
     fn no_secondary(
@@ -519,19 +529,8 @@ fn next_round_transitions(cells: &[usize]) -> Vec<(RoundScenario, Vec<usize>)> {
                     // outside -> both-halves -> secondary-only -> primary-only
                     // -> leader.
                     let primary = remaining - secondary;
-
-                    if outside > 0 {
-                        next_cells.push(outside);
-                    }
-                    if both > 0 {
-                        next_cells.push(both);
-                    }
-                    if secondary > 0 {
-                        next_cells.push(secondary);
-                    }
-                    if primary > 0 {
-                        next_cells.push(primary);
-                    }
+                    let checkpoint = next_cells.len();
+                    push_nonzero_cells(next_cells, [outside, both, secondary, primary]);
 
                     let shared_mask = range_mask(start + outside, both);
                     let next_secondary = state.secondary_mask
@@ -540,7 +539,6 @@ fn next_round_transitions(cells: &[usize]) -> Vec<(RoundScenario, Vec<usize>)> {
                     let next_primary = state.primary_mask
                         | shared_mask
                         | range_mask(start + outside + both + secondary, primary);
-                    let mut next_leader = state.leader;
                     if has_leader {
                         // As in `no_secondary`, keep the leader as a singleton
                         // at the tail of its source cell's refined block.
@@ -549,11 +547,7 @@ fn next_round_transitions(cells: &[usize]) -> Vec<(RoundScenario, Vec<usize>)> {
                         // bridge both halves when the masks differ.
                         let leader_idx = start + outside + both + secondary + primary;
                         next_cells.push(1);
-                        next_leader = Some(leader_idx);
-
-                        for (leader_in_primary, leader_in_secondary) in
-                            [(true, false), (false, true), (true, true)]
-                        {
+                        for &(leader_in_primary, leader_in_secondary) in LEADER_PLACEMENTS {
                             let leader_bit = 1u64 << leader_idx;
                             with_secondary(
                                 ranges,
@@ -570,55 +564,27 @@ fn next_round_transitions(cells: &[usize]) -> Vec<(RoundScenario, Vec<usize>)> {
                                     } else {
                                         next_secondary
                                     },
-                                    leader: next_leader,
+                                    leader: Some(leader_idx),
                                 },
                                 next_cells,
                                 out,
                             );
                         }
-
-                        next_cells.pop();
-                        if primary > 0 {
-                            next_cells.pop();
-                        }
-                        if secondary > 0 {
-                            next_cells.pop();
-                        }
-                        if both > 0 {
-                            next_cells.pop();
-                        }
-                        if outside > 0 {
-                            next_cells.pop();
-                        }
-                        continue;
+                    } else {
+                        with_secondary(
+                            ranges,
+                            leader_cell,
+                            cell_idx + 1,
+                            SecondaryState {
+                                primary_mask: next_primary,
+                                secondary_mask: next_secondary,
+                                leader: state.leader,
+                            },
+                            next_cells,
+                            out,
+                        );
                     }
-
-                    with_secondary(
-                        ranges,
-                        leader_cell,
-                        cell_idx + 1,
-                        SecondaryState {
-                            primary_mask: next_primary,
-                            secondary_mask: next_secondary,
-                            leader: next_leader,
-                        },
-                        next_cells,
-                        out,
-                    );
-
-                    // Undo the refined cells before trying the next allocation.
-                    if primary > 0 {
-                        next_cells.pop();
-                    }
-                    if secondary > 0 {
-                        next_cells.pop();
-                    }
-                    if both > 0 {
-                        next_cells.pop();
-                    }
-                    if outside > 0 {
-                        next_cells.pop();
-                    }
+                    next_cells.truncate(checkpoint);
                 }
             }
         }
@@ -1070,6 +1036,26 @@ mod tests {
         assert_eq!(
             scenario.route(View::new(1), &4, &participants),
             SplitTarget::Primary
+        );
+    }
+
+    #[test]
+    fn leader_visible_to_both_halves_preserves_partitions_and_route() {
+        let scenario = Scenario {
+            rounds: vec![RoundScenario {
+                leader: 3,
+                primary_mask: 0b1010,
+                secondary_mask: 0b1100,
+            }],
+        };
+        let participants: Vec<u32> = (0..4).collect();
+        let (primary, secondary) = scenario.partitions(View::new(1), &participants);
+
+        assert_eq!(primary, vec![1, 3]);
+        assert_eq!(secondary, vec![2, 3]);
+        assert_eq!(
+            scenario.route(View::new(1), &3, &participants),
+            SplitTarget::Both
         );
     }
 


consensus/src/simplex/mocks/twins.rs
patrick-ogrady, line 1: Initial feedback:

```
  - [P2] Avoid materializing all compromised combinations — /Users/patrickogrady/code/monorepo-pr-3281/consensus/src/simplex/mocks/twins.rs:350-
    352
    compromised_sets always enumerates every n choose faults combination into all before checking max_sets, so max_compromised_sets does not
    actually bound memory/runtime. With larger campaigns (for example participants=30, faults=10) this can allocate tens of millions of vectors
    and hang or OOM even when only a few sets are requested. Sampling combinations directly (or using reservoir-style selection) would keep
    generation proportional to max_sets.
    
```

```
  1. Weakened blocked-peers assertion (mod.rs:6072-6084): The old code hard-asserted all blocked peers are twins. The new code only logs a
  warning. If honest nodes are getting blocked under adversarial schedules, that's worth understanding -- not silently tolerating. This could
  mask real bugs.
  
```

```
  2. Resolver traffic delivered to both twins (mod.rs:5876-5879): Changed from dropping to delivering resolver traffic to both twin halves. This
  lets twins receive block data that the partition scenario was supposed to prevent them from seeing. Makes the adversarial model strictly
  weaker. Needs explicit justification for why this is the right tradeoff.
  
```

```
  3. honest_start calculation (mod.rs:6021): twin_indices.len() * 2 relies on the implicit ordering of reporters (twins pushed first, 2 each).
  Reordering the loops silently breaks safety/liveness checks. Fragile.
  
```

```
  4. Stale test name (mod.rs:6279): test_twins_large_shuffle no longer tests shuffling. Should be renamed.
  
```

```
  5. Infinite loop risk in direct sampling (twins.rs:319-327): If the unique arrangement space is smaller than max_scenarios but
  arrangement_count overflows (returning None), the while loop spins forever. Practically impossible at current scale but the code has no
  safeguard.
```

consensus/src/simplex/mocks/twins.rs
```diff
@@ -1,343 +1,640 @@
...
+    let mut current = vec![0usize; n];
+    generate(1, n, 1, max_partitions, &mut current, &mut out);
+    out
+}
 
-        // View 0: split at 0 % 5 = 0 -> primary gets [], secondary gets [0,1,2,3,4]
-        let (primary, secondary) = Strategy::View.partitions(View::new(0), &participants);
-        assert!(primary.is_empty());
-        assert_eq!(secondary, vec![0, 1, 2, 3, 4]);
+fn round_scenarios(n: usize, max_partitions: usize) -> Vec<RoundScenario> {
```
clabby: `max_scenarios` pruning is applied only after `round_scenarios()` fully materializes the partition/leader option space. For larger `participants`/`max_partitions`, this can blow up CPU/memory even when `max_scenarios` is tiny (because step 1/2 still enumerates everything).

Consider pushing the cap earlier: generate round options lazily and sample directly, or add a bounded sampler for partition/round options before building the full `BTreeSet`.

consensus/src/simplex/mod.rs
```diff
@@ -5693,58 +5805,57 @@ mod tests {
...
-                            strategy.partitions(msg.view(), participants.as_ref());
+                            scenario.partitions(msg.view(), participants.as_ref());
                         match origin {
                             SplitOrigin::Primary => Some(Recipients::Some(primary)),
                             SplitOrigin::Secondary => Some(Recipients::Some(secondary)),
                         }
                     }
                 };
-                let make_drop_forwarder =
+                let make_resolver_forwarder =
```
clabby: `make_resolver_forwarder` / `make_resolver_router` still drop all resolver traffic (`None` / `SplitTarget::None`). The new naming suggests resolver routing is part of the scenario model, but behavior remains “hard drop”.

Either route resolver traffic explicitly (for example `Both` in suffix rounds), or rename these back to `drop_*` to avoid misleading future maintainers.

consensus/src/simplex/mod.rs
```diff
@@ -5640,20 +5640,131 @@ mod tests {
...
+    {
+        fn elect(&self, round: Round, certificate: Option<&S::Certificate>) -> Participant {
+            let idx = round.view().get().saturating_sub(1) as usize;
+            if let Some(&leader) = self.round_leaders.get(idx) {
+                return leader;
+            }
+            if self.honest_leaders.is_empty() {
+                return self.fallback.elect(round, certificate);
+            }
+            let honest_idx = (round.epoch().get().wrapping_add(round.view().get())) as usize
```
clabby: The synchronous suffix leader selection hard-codes `(epoch + view) % honest_leaders.len()` and ignores `fallback.elect(...)` semantics (including shuffled round-robin or random electors). That makes the generic `L` parameter mostly irrelevant after attack rounds.

If this is intentional, a short comment explaining why `fallback` is bypassed would help. If not intentional, consider deriving suffix leaders from `fallback` and filtering/projection onto the honest set.

consensus/src/simplex/mocks/twins.rs
clabby: `max_scenarios` pruning is applied only after `round_scenarios()` fully materializes the partition/leader option space. For larger `participants`/`max_partitions`, this can blow up CPU/memory even when `max_scenarios` is tiny (because step 1/2 still enumerates everything).

Consider pushing the cap earlier: generate round options lazily and sample directly, or add a bounded sampler for partition/round options before building the full `BTreeSet`.

consensus/src/simplex/mod.rs
clabby: `make_resolver_forwarder` / `make_resolver_router` still drop all resolver traffic (`None` / `SplitTarget::None`). The new naming suggests resolver routing is part of the scenario model, but behavior remains “hard drop”.

Either route resolver traffic explicitly (for example `Both` in suffix rounds), or rename these back to `drop_*` to avoid misleading future maintainers.

consensus/src/simplex/mod.rs
clabby: The synchronous suffix leader selection hard-codes `(epoch + view) % honest_leaders.len()` and ignores `fallback.elect(...)` semantics (including shuffled round-robin or random electors). That makes the generic `L` parameter mostly irrelevant after attack rounds.

If this is intentional, a short comment explaining why `fallback` is bypassed would help. If not intentional, consider deriving suffix leaders from `fallback` and filtering/projection onto the honest set.

consensus/src/simplex/mocks/twins.rs
```diff
@@ -1,343 +1,654 @@
...
 use commonware_p2p::simulated::SplitTarget;
-use rand::{rngs::StdRng, seq::SliceRandom, Rng, SeedableRng};
+use rand::{rngs::StdRng, Rng, SeedableRng};
+use std::collections::{BTreeSet, HashSet};
+
+/// Per-round adversarial setting from the Twins framework.
+#[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
+pub struct RoundScenario {
+    leader: usize,
+    primary_mask: u64,
```
patrick-ogrady, line 49: It would be worthwhile IMHO to add a comment do explain primary/secondary mask?

consensus/src/simplex/mocks/twins.rs
```diff
@@ -1,343 +1,654 @@
...
+            return masks_to_partitions(round.primary_mask, round.secondary_mask, participants);
+        }
+        (participants.to_vec(), participants.to_vec())
+    }
 
-    /// One twin sends only to the specified participant index, the other sends to everyone else.
-    Isolate(usize),
+    /// Routes a message from sender to the correct twin half at a given view.
+    ///
+    /// Unlike [`partitions`](Self::partitions), this avoids allocating temporary
```
patrick-ogrady, line 90: Do we need to retain the partitions function? This almost reads like an improvement (and then the old version was kept around)?

consensus/src/simplex/mocks/twins.rs
```diff
@@ -1,343 +1,654 @@
...
+    /// Vecs by comparing bitmasks directly.
+    pub fn route<P: PartialEq>(&self, view: View, sender: &P, participants: &[P]) -> SplitTarget {
+        let idx = view.get().saturating_sub(1) as usize;
+        if let Some(round) = self.rounds.get(idx) {
+            return route_with_masks(
```
patrick-ogrady: Should we just inline `route_with_masks`?

consensus/src/simplex/mocks/twins.rs
```diff
@@ -1,343 +1,654 @@
...
+}
+
+/// Splits participants by `view % n` to mirror the legacy twins `View` strategy.
+pub fn view_partitions<P: Clone>(view: View, participants: &[P]) -> (Vec<P>, Vec<P>) {
+    let split = (view.get() as usize) % participants.len();
+    let (primary, secondary) = participants.split_at(split);
+    (primary.to_vec(), secondary.to_vec())
+}
+
+/// Routes a sender using the legacy twins `View` strategy.
```
patrick-ogrady: Do we need to retain the “legacy” strategy for `view_partitions` and `view_route`? If so, we should not refer to them as legacy?

consensus/src/simplex/mocks/twins.rs
```diff
@@ -1,343 +1,654 @@
...
+
+/// Framework configuration for generating Twins cases.
+#[derive(Clone, Copy, Debug)]
+pub struct Framework {
+    /// Number of participants in the network.
+    pub participants: usize,
+    /// Number of compromised participants.
+    pub faults: usize,
+    /// Number of adversarial rounds before synchronous suffix.
+    pub rounds: usize,
```
patrick-ogrady, line 250: We should be able to see resolution while this remains Usize::MAX (as long as faults <= f)? Do we test that?

consensus/src/simplex/mocks/twins.rs
```diff
@@ -1,343 +1,654 @@
...
-        let participants: Vec<u32> = (0..5).collect();
-
-        // Isolate(2): primary gets only [2], secondary gets everyone else [0,1,3,4]
-        let partition = Strategy::Isolate(2);
+fn generate_scenarios(
+    seed: u64,
+    n: usize,
+    rounds: usize,
+    max_partitions: usize,
+    max_scenarios: usize,
```
patrick-ogrady, line 754: I’m assuming the content of these scenarios changes based on seed?

IIRC in the paper we compress/combine the sample space by striking similar rounds (say same topology but a different player does X). Do we do that here?

consensus/src/simplex/mocks/twins.rs
```diff
@@ -1,343 +1,654 @@
...
+    if options.is_empty() {
+        return Vec::new();
+    }
 
-        // Should be constant across views
-        for view in [0, 1, 5, 100] {
-            let (primary, secondary) = partition.partitions(View::new(view), &participants);
-            assert_eq!(primary, vec![2]);
-            assert_eq!(secondary, vec![0, 1, 3, 4]);
+    let mut rng = StdRng::seed_from_u64(seed);
```
patrick-ogrady: I think there is a helper for this? (`test_rng` or something)?

consensus/src/simplex/mod.rs
```diff
@@ -5640,20 +5640,137 @@ mod tests {
         );
     }
 
-    /// Implementation of [Twins: BFT Systems Made Robust](https://arxiv.org/abs/2004.10617).
-    fn twins<S, F, L>(seed: u64, n: u32, strategy: Strategy, link: Link, mut fixture: F)
+    #[derive(Clone, Copy, Debug)]
+    struct TwinsCampaign {
```
patrick-ogrady, line 5684: Should these helpers live in `twins.rs`?

We could then just call them `Campaign`, `Elector`, etc.

consensus/src/simplex/mocks/twins.rs
```diff
@@ -1,343 +1,654 @@
...
+    rounds: usize,
+    max_partitions: usize,
+    max_scenarios: usize,
+) -> Vec<Scenario> {
+    // Pass `max_scenarios` as a cap so that `round_scenarios` stops
+    // materializing partitions once it has collected enough distinct round
+    // options for the sampling path.  Without this, the partition
+    // enumeration can blow up memory for moderate `n`/`max_partitions`
+    // even when `max_scenarios` is small.
+    let options = round_scenarios(n, max_partitions, Some(max_scenarios));
```
patrick-ogrady: `generate_scenarios` is applying `max_scenarios` while building the per-round option pool instead of only limiting the final sampled arrangements. That changes the reachable state space, not just the sample count: for example, `round_scenarios(10, 3, Some(2))` only produces leader `0`, while the uncapped set covers all leaders. It also does not actually save the memory that the comment claims, because `partition_scenarios()` is still fully materialized first.

Please move the cap to arrangement sampling time (or otherwise preserve the full round-option set before pruning), so later leaders and partition shapes remain reachable.

```rust
#[test]
fn capped_round_options_preserve_all_leaders() {
    let uncapped: HashSet<_> = round_scenarios(10, 3, None)
        .into_iter()
        .map(|round| round.leader())
        .collect();
    let capped: HashSet<_> = round_scenarios(10, 3, Some(2))
        .into_iter()
        .map(|round| round.leader())
        .collect();

    assert_eq!(capped, uncapped);
}
```

consensus/src/simplex/mocks/twins.rs
```diff
@@ -1,343 +1,654 @@
...
+    fn generate(
+        idx: usize,
+        n: usize,
+        used: usize,
+        max_partitions: usize,
+        current: &mut [usize],
+        out: &mut Vec<Vec<usize>>,
+    ) {
+        if idx == n {
+            if used > 1 {
```
patrick-ogrady: `partition_scenarios` currently skips the `used == 1` case, so the new generator can never produce a full-broadcast round. That drops coverage that the previous `Strategy::Broadcast` cases had, rather than subsuming it. If this framework is meant to replace the old twins matrix, it should still generate the one-partition arrangement and let `round_scenarios` deduplicate equivalent results.

```rust
#[test]
fn round_scenarios_include_full_broadcast() {
    let full = (1u64 << 5) - 1;
    assert!(round_scenarios(5, 3, None).into_iter().any(|round| {
        round.primary_mask == full && round.secondary_mask == full
    }));
}
```

consensus/src/simplex/mod.rs
```diff
@@ -5640,20 +5640,137 @@ mod tests {
...
+    {
+        fn elect(&self, round: Round, certificate: Option<&S::Certificate>) -> Participant {
+            let idx = round.view().get().saturating_sub(1) as usize;
+            if let Some(&leader) = self.round_leaders.get(idx) {
+                return leader;
+            }
+            if self.honest_leaders.is_empty() {
+                return self.fallback.elect(round, certificate);
+            }
+            // Suffix rounds intentionally bypass `fallback` and rotate
```
patrick-ogrady: After the scripted prefix, `TwinsElectorState::elect` stops delegating to `fallback.elect` and rotates over `honest_leaders` instead. That makes the suffix materially easier than the real protocol: the liveness checks run far past the scripted rounds, but those later views no longer exercise the normal VRF / certificate leader selection path, and compromised real leaders are silently skipped in `RoundRobin` mode.

The suffix should preserve the underlying elector and let the network/synchrony setup drive recovery, otherwise the twins campaign can miss exactly the state-dependent leader-selection failures it is supposed to expose.

```rust
#[test]
fn twins_elector_suffix_delegates_to_fallback() {
    let framework = twins::Framework {
        participants: 5,
        faults: 1,
        rounds: 3,
        max_partitions: 3,
        max_scenarios: 1,
        max_compromised_sets: 1,
    };
    let case = twins::cases(0, framework)
        .into_iter()
        .next()
        .expect("expected a twins case");

    let mut rng = test_rng();
    let Fixture { participants, .. } = ed25519::fixture(
        &mut rng,
        b"twins_elector_fallback",
        framework.participants as u32,
    );
    let participants = Set::try_from(participants).expect("participants should be unique");

    let twins = <TwinsElector<RoundRobin<Sha256>> as Elector<ed25519::Scheme>>::build(
        TwinsElector::new(
            RoundRobin::<Sha256>::default(),
            &case.scenario,
            &case.compromised,
            framework.participants,
        ),
        &participants,
    );
    let fallback = RoundRobin::<Sha256>::default().build(&participants);

    for view in (framework.rounds as u64 + 1)..=20 {
        let round = Round::new(Epoch::new(333), View::new(view));
        assert_eq!(twins.elect(round, None), fallback.elect(round, None));
    }
}
```

consensus/src/simplex/mod.rs
```diff
@@ -5693,58 +5811,57 @@ mod tests {
...
-                            strategy.partitions(msg.view(), participants.as_ref());
+                            scenario.partitions(msg.view(), participants.as_ref());
                         match origin {
                             SplitOrigin::Primary => Some(Recipients::Some(primary)),
                             SplitOrigin::Secondary => Some(Recipients::Some(secondary)),
                         }
                     }
                 };
-                let make_drop_forwarder =
+                let drop_resolver_forwarder =
```
patrick-ogrady: Resolver traffic is still blackholed by `drop_resolver_forwarder` / `drop_resolver_router`, so the sparse-mask twins cases are not modeling the recovery behavior described by the campaign. Votes and certificates use the new partition routing, but resolver fetches and replies never reach either twin half. That makes missing-certificate recovery impossible during the adversarial prefix and weakens the harness relative to the existing split model in `consensus/fuzz`.

Please route resolver traffic to both halves instead of dropping it.

```rust
#[test]
fn twins_resolver_reaches_both_halves() {
        let cfg = Config {
            max_size: 1024,
            disconnect_on_block: true,
            tracked_peer_sets: Some(3),
        };
        let (network, mut oracle) = Network::new(context.with_label("network"), cfg);
        network.start();

        let twin = PrivateKey::from_seed(30).public_key();
        let peer = PrivateKey::from_seed(31).public_key();
        let mut manager = oracle.manager();
        manager
            .track(0, [twin.clone(), peer.clone()].try_into().unwrap())
            .await;

        let (mut peer_sender, _) = oracle.control(peer.clone()).register(2, TEST_QUOTA).await.unwrap();
        let (resolver_sender, resolver_receiver) =
            oracle.control(twin.clone()).register(2, TEST_QUOTA).await.unwrap();

        let (_primary_sender, _secondary_sender) =
            resolver_sender.split_with(|_, recipients, _| Some(recipients.clone()));
        let (mut primary_recv, mut secondary_recv) = resolver_receiver.split_with(
            context.with_label("resolver_split"),
            |_| SplitTarget::Both,
        );

        let link = Link {
            latency: Duration::from_millis(0),
            jitter: Duration::from_millis(0),
            success_rate: 1.0,
        };
        oracle.add_link(peer.clone(), twin.clone(), link.clone()).await.unwrap();
        oracle.add_link(twin.clone(), peer.clone(), link).await.unwrap();

        peer_sender
            .send(Recipients::One(twin.clone()), b"resolver_request", false)
            .await
            .unwrap();

        let (sender, payload) = primary_recv.recv().await.unwrap();
        assert_eq!(sender, peer);
        assert_eq!(payload, b"resolver_request");
        let (sender, payload) = secondary_recv.recv().await.unwrap();
        assert_eq!(sender, peer);
        assert_eq!(payload, b"resolver_request");
    });
}
```

consensus/src/simplex/mod.rs
```diff
@@ -5899,20 +6021,19 @@ mod tests {
                 engine_handlers.push(engine.start(pending, recovered, resolver));
             }
 
-            // Wait for progress (liveness check)
+            // Wait for progress (liveness check) on honest replicas only.
             let mut finalizers = Vec::new();
-            for reporter in reporters.iter_mut() {
+            for reporter in reporters.iter_mut().skip(honest_start) {
```
patrick-ogrady: This progress loop no longer verifies that both twin halves remain live. As written, a twin-only stall is invisible as long as the honest suffix still finalizes, which can mask exactly the routing/fidelity bugs this harness is meant to expose. Please either restore the all-reporter wait or add an explicit assertion that each twin reporter advances into the synchronous suffix.

consensus/src/simplex/mocks/twins.rs
```diff
@@ -1,343 +1,654 @@
...
-                let (primary, secondary) = shuffled.split_at(split);
-                (primary.to_vec(), secondary.to_vec())
-            }
+fn route_with_masks<P: PartialEq>(
+    sender: &P,
+    primary_mask: u64,
+    secondary_mask: u64,
+    participants: &[P],
+) -> SplitTarget {
+    let Some(sender_idx) = participants.iter().position(|p| p == sender) else {
```
patrick-ogrady: `route_with_masks` maps both "intentionally excluded by this scenario" and "sender not present in the runtime participant list" to `SplitTarget::None`. The latter is a harness invariant failure, not a valid schedule outcome, so silently dropping it makes participant-order/size mismatches harder to detect. A debug assertion here would fail fast instead of turning a configuration bug into a seemingly valid dropped-message case.

consensus/src/simplex/mod.rs
```diff
@@ -5835,12 +5838,22 @@ mod tests {
...
+                        pending,
+                        recovered,
+                        inert_channel(participants.as_ref()),
+                    ));
                 }
             }
 
-            // Create honest engines
-            for (idx, validator) in participants.iter().enumerate().skip(faults as usize) {
+            let honest_start = reporters.len();
```
patrick-ogrady: put under //

patrick-ogrady:
bugbot run

patrick-ogrady:
Remaining Items
- [x] Further reduce duplicate samples
- [x] Migrate all twins tests to run with mock certificate
- [x] Require finalizations not just new views

consensus/src/simplex/mocks/twins.rs
```diff
@@ -1,343 +1,1069 @@
...
+    pub fn rounds(&self) -> &[RoundScenario] {
+        &self.rounds
+    }
+
+    /// Returns recipients for each twin half at a given view.
+    ///
+    /// Views after the configured adversarial rounds use full synchrony
+    /// (`all -> all`) to model eventual synchrony for liveness checks.
+    pub fn partitions<P: Clone>(&self, view: View, participants: &[P]) -> (Vec<P>, Vec<P>) {
+        let idx = view.get().saturating_sub(1) as usize;
```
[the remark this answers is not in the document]
patrick-ogrady: View 0 is reserved for genesis.

consensus/src/simplex/mod.rs
```diff
@@ -5640,149 +5644,271 @@ mod tests {
...
-            link_validators(&mut oracle, &participants, Action::Link(link), None).await;
+            let activity_timeout = ViewDelta::new(10);
+            let skip_timeout = ViewDelta::new(5);
+            let namespace = b"consensus".to_vec();
+            let link = link.clone();
+            let trailing_finalizations = campaign.trailing_finalizations;
+            let mut case_fixture =
+                |ctx: &mut deterministic::Context, ns: &[u8], n: u32| fixture(ctx, ns, n);
+            let cfg = deterministic::Config::new()
+                .with_rng(Box::new(StdRng::from_rng(&mut *rng).unwrap()));
```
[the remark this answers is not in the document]
patrick-ogrady: This is intentional (like we do for 1k).

patrick-ogrady:
bugbot run

cryptography/src/certificate/mocks.rs
```diff
@@ -0,0 +1,1156 @@
+//! Mocks for certificate signing schemes.
```
clabby, line 1: This is amazing :heart_hands:

cryptography/src/certificate/mocks.rs
```diff
@@ -0,0 +1,1156 @@
...
+///
+/// Signatures and certificates are cheap synthetic IDs. Verification succeeds only
+/// if the corresponding subject was previously recorded in shared state.
+pub struct Generic<
+    P: PublicKey,
+    N: super::Namespace,
+    const ATTRIBUTABLE: bool = true,
+    const BATCHABLE: bool = true,
+    const ALLOW_INVALID: bool = true,
+> {
```
clabby: Really nitpicky, but when we get >2 generics, let’s use `where`?

consensus/fuzz/src/lib.rs
```diff
@@ -624,24 +644,28 @@ fn run_with_twin_mutator<P: simplex::Simplex>(input: FuzzInput) {
                 &input.strategy,
                 (vote_sender_secondary, vote_receiver_secondary),
                 (certificate_sender_secondary, certificate_receiver_secondary),
-                (resolver_sender_secondary, resolver_receiver_secondary),
+                inert_channel(participants.as_ref()),
```
clabby: In Twinable mode, the secondary disrupter now receives `inert_channel(...)` for resolver traffic, so resolver mutations from `StrategyChoice` (`handle_resolver` / `mutate_resolver_bytes`) no longer reach any engine. This is a fuzz-coverage regression versus the previous split-through resolver wiring and leaves resolver-adversarial behavior largely unexercised in `*_twin` fuzz targets. Reconnect resolver channels in twin mode (e.g. with `SplitTarget::Both`) or add a dedicated twin-resolver fuzz path.

patrick-ogrady: Just because its fuzzing, I'm restoring this. It may be garbage (you may get back things you didn't request) but I think that chaos is fine.

patrick-ogrady:
bugbot run

PR #3284, opened 2026-03-02
[runtime/iouring] implement timeouts in userspace

andresilva:
This PR replaces io_uring linked-kernel timeouts with a userspace timeout wheel tuned for our workload (many in-flight ops, rare actual expirations). Each operation deadline is converted to a wheel tick, scheduled in O(1), and advanced once per loop to produce expiration candidates. The wheel keeps separate metadata for occupied buckets vs active deadlines, supports bounded scans with bitsets, and keeps stale-entry handling cheap so the common "completes before deadline" path stays fast.

Timeout tracking is now explicit and policy-driven: deadlines are clamped to a configured horizon, next deadline is computed from the wheel's active minimum, and expired entries are filtered by waiter identity before cancellation is requested. The implementation also preserves bucket capacity across cycles to reduce allocator churn and includes idle-path cleanup of stale occupancy metadata.

This PR also reworks waiter identity and lifecycle handling. `WaiterId` now packs slot index + generation, with the high bit reserved as a cancellation tag in `user_data`. That gives robust stale-CQE detection after slot reuse and a single decode path for op/cancel completions. Waiter state transitions are explicit (`Active` vs `CancelRequested`), and completion semantics are clear: if the op CQE arrives first, that result is delivered, a later cancel CQE is ignored.

This also fixes timeout behavior in the network backend: previously, timeout budgeting was effectively per submitted io_uring op, not per top-level `send`/`recv` call. We now compute one deadline per `send`/`recv` operation and reuse it across internal partial read/write loops, so large or fragmented transfers are bounded by a single end-to-end timeout budget (matching tokio backend behavior).

Finally, the io_uring code was split into separate modules.

Related #2883.

### DIFF
--- a/runtime/src/iouring/mod.rs
+++ b/runtime/src/iouring/mod.rs
@@ -79,7 +79,7 @@
 //!
 //! When the operation channel closes, the event loop enters a drain phase:
 //! 1. Stops accepting new operations
-//! 2. Waits for all in-flight operations to complete
+//! 2. Waits for all in-flight operations to complete or be cancelled
 //! 3. If `shutdown_timeout` is configured, abandons remaining operations after the timeout
 //! 4. Cleans up and exits. Dropping the last submitter signals `eventfd` so shutdown is observed
 //!    promptly even if the loop is blocked.
@@ -662,26 +662,31 @@ impl IoUringLoop {
         // Route op/cancel completions through waiter state. Only terminal
         // transitions return a completed waiter to deliver to the caller.
         if let Some(completed) = self.waiters.on_completion(user_data, cqe.result()) {
-            let CompletedWaiter {
-                sender,
-                buffer,
-                mut result,
-                cancelled,
-                target_tick,
-            } = completed;
-
-            // Remove active deadline tracking when this waiter completes.
-            if let Some(target_tick) = target_tick {
-                self.timeout_wheel.remove(target_tick);
-            }
-
-            // Surface timeout as ETIMEDOUT when cancellation succeeded.
-            if cancelled && result == -libc::ECANCELED {
-                result = -libc::ETIMEDOUT;
-            }
-
-            let _ = sender.send((result, buffer));
+            self.deliver_completion(completed);
         }
+    }
+
+    /// Deliver a completed waiter to its caller and clean up timeout state.
+    fn deliver_completion(&mut self, completed: CompletedWaiter) {
+        let CompletedWaiter {
+            sender,
+            buffer,
+            mut result,
+            cancelled,
+            target_tick,
+        } = completed;
+
+        // Remove active deadline tracking when this waiter completes.
+        if let Some(target_tick) = target_tick {
+            self.timeout_wheel.remove(target_tick);
+        }
+
+        // Surface timeout as ETIMEDOUT when cancellation succeeded.
+        if cancelled && result == -libc::ECANCELED {
+            result = -libc::ETIMEDOUT;
+        }
+
+        let _ = sender.send((result, buffer));
     }
 
     /// Advance the timeout wheel and enqueue cancellations for newly expired waiters.
@@ -719,31 +724,51 @@ impl IoUringLoop {
     /// Keeps draining CQEs until all waiters complete or shutdown budget is
     /// exhausted.
     ///
-    /// If `shutdown_timeout` is `None`, this waits until all waiters complete.
+    /// If `shutdown_timeout` is `None`, this waits until all waiters complete or are cancelled.
     /// If `shutdown_timeout` is `Some`, this waits until completion or timeout,
     /// then abandons any remaining waiters.
-    ///
-    /// This path does not advance timeout deadlines or stage new cancellation
-    /// SQEs after shutdown starts.
     fn drain(&mut self, ring: &mut IoUring) {
         let mut remaining = self.cfg.shutdown_timeout;
 
         // Keep driving completions until all in-flight waiters finish or the
         // shutdown budget is exhausted.
-        while !self.waiters.is_empty() {
+        loop {
+            // Always drain CQEs first, even after a timed wait: completions can
+            // race with timeout expiry and still be pending in the queue.
+            for cqe in ring.completion() {
+                self.handle_cqe(cqe);
+            }
+
+            // CQE draining can finish the last waiter, so stop before another
+            // submit-and-wait cycle.
+            if self.waiters.is_empty() {
+                break;
+            }
+
+            // Once shutdown budget is exhausted, abandon any remaining waiters
+            // immediately instead of advancing more deadlines or staging new cancels.
             if remaining.is_some_and(|t| t.is_zero()) {
                 break;
             }
 
-            let start = Instant::now();
-            self.submit_and_wait(ring, 1, remaining)
-                .expect("unable to submit to ring");
-
-            // Always drain CQEs, even after timeout: completions can race with
-            // timeout expiry and still be pending in the queue
-            for cqe in ring.completion() {
-                self.handle_cqe(cqe);
+            // Keep userspace deadline processing alive during shutdown so
+            // in-flight timed operations preserve their ETIMEDOUT semantics.
+            self.advance_timeouts();
+            {
+                let mut submission_queue = ring.submission();
+                self.stage_cancellations(&mut submission_queue);
             }
+            let timeout = match (remaining, self.timeout_wheel.next_deadline()) {
+                (Some(remaining), Some(deadline)) => Some(remaining.min(deadline)),
+                (Some(remaining), None) => Some(remaining),
+                (None, Some(deadline)) => Some(deadline),
+                (None, None) => None,
+            };
+
+            // Wait for at least one completion or timeout.
+            let start = Instant::now();
+            self.submit_and_wait(ring, 1, timeout)
+                .expect("unable to submit to ring");
 
             // Charge elapsed wall time against the shutdown budget.
             if let Some(remaining) = remaining.as_mut() {
@@ -1093,20 +1118,7 @@ mod tests {
             .waiters
             .on_completion(old_slot.user_data(), 0)
             .expect("missing waiter completion");
-        let CompletedWaiter {
-            sender,
-            buffer,
-            mut result,
-            cancelled,
-            target_tick,
-        } = completed;
-        if let Some(target_tick) = target_tick {
-            iouring.timeout_wheel.remove(target_tick);
-        }
-        if cancelled && result == -libc::ECANCELED {
-            result = -libc::ETIMEDOUT;
-        }
-        let _ = sender.send((result, buffer));
+        iouring.deliver_completion(completed);
 
         // Reuse the same slot for a new waiter with a later timeout.
         let (tx, _rx) = oneshot::channel();
@@ -1147,20 +1159,7 @@ mod tests {
             .waiters
             .on_completion(slot_index.user_data(), 123)
             .expect("missing completion");
-        let CompletedWaiter {
-            sender,
-            buffer,
-            mut result,
-            cancelled,
-            target_tick,
-        } = completed;
-        if let Some(target_tick) = target_tick {
-            iouring.timeout_wheel.remove(target_tick);
-        }
-        if cancelled && result == -libc::ECANCELED {
-            result = -libc::ETIMEDOUT;
-        }
-        let _ = sender.send((result, buffer));
+        iouring.deliver_completion(completed);
         let completed = iouring
             .waiters
             .on_completion(slot_index.cancel_user_data(), -libc::ECANCELED);
@@ -1322,6 +1321,43 @@ mod tests {
     }
 
     #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn test_shutdown_no_timeout_continues_deadline_processing() {
+        let cfg = Config {
+            max_op_timeout: Duration::from_millis(250),
+            timeout_wheel_tick: Duration::from_millis(5),
+            shutdown_timeout: None,
+            ..Default::default()
+        };
+        let mut registry = Registry::default();
+        let (submitter, iouring) = IoUringLoop::new(cfg, &mut registry);
+        let handle = std::thread::spawn(move || iouring.run());
+
+        let timeout = Timespec::new().sec(5_000);
+        let timeout = opcode::Timeout::new(&timeout).build();
+        let (tx, rx) = oneshot::channel();
+        submitter
+            .send(Op {
+                work: timeout,
+                sender: tx,
+                buffer: None,
+                fd: None,
+                iovecs: None,
+                deadline: Some(Instant::now() + Duration::from_millis(50)),
+            })
+            .await
+            .unwrap();
+
+        drop(submitter);
+
+        let (result, _) = tokio::time::timeout(Duration::from_secs(2), rx)
+            .await
+            .expect("deadline completion timed out")
+            .expect("missing deadline completion");
+        assert_eq!(result, -libc::ETIMEDOUT);
+        handle.join().unwrap();
+    }
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
     async fn test_shutdown_timeout() {
         // Create an io_uring instance with shutdown timeout enabled
         let cfg = Config {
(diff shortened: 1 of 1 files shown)


runtime/src/iouring/mod.rs
```diff
@@ -992,11 +740,7 @@ impl IoUringLoop {
                 self.handle_cqe(cqe);
             }
 
-            if !got_completion {
-                // Shutdown timeout elapsed before all in-flight work completed.
-                break;
-            }
-
+            // Charge elapsed wall time against the shutdown budget.
```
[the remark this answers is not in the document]
andresilva: This was intentional but I'll look into it, seems like I might have overlooked a bit.

runtime/src/iouring/mod.rs
```diff
@@ -847,143 +554,184 @@ impl IoUringLoop {
...
-                timespec,
-            });
+            let target_tick = if let Some(deadline) = deadline {
+                // Avoid per-loop clock reads when no deadlines are active.
+                // When the first deadline arrives after an idle period, align
+                // wheel time once before converting deadlines to ticks.
+                if !advanced {
+                    assert!(self.timeout_wheel.advance(Instant::now()).is_none());
+                    advanced = true;
+                }
```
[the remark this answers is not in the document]
andresilva: Fair. This is just an optimization that doesn't affect correctness, but will improve.

runtime/src/iouring/timeout.rs
```diff
@@ -0,0 +1,990 @@
...
+    /// call `remove` for each active expiry before querying this method.
+    pub const fn next_deadline(&self) -> Option<Duration> {
+        if self.min_scheduled_tick == Tick::MAX {
+            return None;
+        }
+
+        let ticks = self.min_scheduled_tick.saturating_sub(self.current_tick);
+        let nanos = self.tick_nanos.saturating_mul(ticks);
+        Some(Duration::from_nanos(nanos))
+    }
```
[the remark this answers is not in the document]
andresilva: Big if true, need to check.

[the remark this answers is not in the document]
andresilva: `min_scheduled_tick` tracks **active** deadlines only, stale drained entries do not increment active counts. Active minima are recomputed on remove, and set to `Tick::MAX` when empty.

runtime/src/iouring/mod.rs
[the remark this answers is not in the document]
andresilva: `fill_submission_queue` stages cancellations **before** receiver disconnect handling. If cancellations cannot all be staged, then we do an early return (`at_capacity = true`), so shutdown path isn't taken immediately. By the time `drain` path is taken then all pending cancels **must** have been submitted. I purposefully didn't handle expiring timeouts in `drain` to keep that code simpler.

runtime/src/iouring/mod.rs
```diff
@@ -333,221 +339,6 @@ pub struct Op {
...
-            .user_data(WAKE_USER_DATA);
-
-        // SAFETY: The poll SQE owns no user pointers and references a valid FD.
-        unsafe {
-            submission_queue
-                .push(&wake_poll)
-                .expect("wake poll SQE should always fit in the ring");
-        }
-    }
-}
```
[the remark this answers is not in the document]
andresilva: When `try_recv` returns `Disconnected` we will proceed to shutdown, at that point we no longer care about processed seq, it's fine as is.

runtime/src/iouring/waiter.rs
```diff
@@ -0,0 +1,471 @@
...
+        let index = waiter_id.index() as usize;
+
+        let waiter = self.entries[index].as_mut()?;
+        if waiter.id != waiter_id {
+            // Slot was reused, this CQE belongs to an older waiter generation.
+            return None;
+        }
+
+        if is_cancel {
+            match result.abs() {
```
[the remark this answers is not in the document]
andresilva: If the kernel gives us `i32::MIN` then it's fine to panic.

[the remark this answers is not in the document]
patrick-ogrady: imho we should add a small comment for this?

runtime/src/iouring/mod.rs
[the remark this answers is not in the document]
patrick-ogrady: Is it well-documented that we stop doing any processing on shutdown?

runtime/src/storage/iouring.rs
```diff
@@ -381,6 +382,7 @@ impl Blob {
                     buffer: Some(OpBuffer::WriteVectored(bufs)),
                     fd: Some(OpFd::File(self.file.clone())),
                     iovecs: Some(iovecs),
+                    deadline: None,
```
patrick-ogrady, line 385: On some storage mediums, I suppose we could set this at some point?

runtime/src/network/iouring.rs
```diff
@@ -34,7 +34,7 @@ use std::{
     net::SocketAddr,
     os::fd::{AsRawFd, OwnedFd},
     sync::Arc,
-    time::Duration,
+    time::{Duration, Instant},
```
patrick-ogrady, line 37: Should we use `tokio::time::Instant` like we do in `tokio` runtime?

runtime/src/network/iouring.rs
```diff
@@ -291,6 +325,7 @@ impl Sink {
     async fn send_single(&self, mut buf: IoBuf) -> Result<(), Error> {
         let mut bytes_sent = 0;
         let buf_len = buf.len();
+        let deadline = Instant::now() + self.timeout;
```
patrick-ogrady, line 328: (no change required) this is consistent with what we do for `sleep_until` in `tokio/runtime.rs`.

runtime/src/network/iouring.rs
```diff
@@ -117,6 +127,10 @@ impl Network {
         // dedicated thread, which guarantees that the same thread that creates
         // the ring is the only thread submitting work to it.
         cfg.iouring_config.single_issuer = true;
+        cfg.iouring_config.max_op_timeout = cfg
```
patrick-ogrady, line 130: It feels weird to me to provide read_write_timeout given we provide iouring_config?

Although, I suppose the same could be said for single_issuer?

runtime/src/iouring/mod.rs
```diff
@@ -116,42 +118,36 @@
...
+    sync::Arc,
     time::{Duration, Instant},
 };
-use tracing::warn;
-
-/// Reserved `user_data` value for a CQE that indicates an operation timed out.
-const TIMEOUT_USER_DATA: u64 = u64::MAX;
-/// Reserved `user_data` value for internal wake poll completions.
-const WAKE_USER_DATA: u64 = u64::MAX - 1;
+use timeout::{Tick, TimeoutWheel};
```
patrick-ogrady, line 143: Usually, I combine these inline with the `mod` imports. In a separate section than external imports (usually after)?

It would be great to activate some linter for this.

runtime/src/iouring/timeout.rs
```diff
@@ -0,0 +1,998 @@
...
+        let tick_nanos = Self::duration_to_nanos_saturating(tick);
+        assert!(tick_nanos > 0, "timeout wheel tick must be non-zero");
+        let max_timeout = max_timeout.max(tick);
+        let slots = Self::slots_for(max_timeout, tick_nanos);
+        let buckets = vec![Vec::new(); slots];
+
+        Self {
+            slot_mask: slots - 1,
+            buckets,
+            occupied: vec![0; slots.div_ceil(64)],
```
patrick-ogrady: Should we be hardcoding here? Maybe something like `const WORD_BITS: usize = u64::BITS as usize;`?

runtime/src/iouring/waiter.rs
```diff
@@ -0,0 +1,471 @@
...
+
+use super::{OpBuffer, OpFd, OpIovecs, Tick, UserData};
+use commonware_utils::channel::oneshot;
+use tracing::warn;
+
+/// Stable waiter identity packed into SQE/CQE `user_data`.
+///
+/// Layout:
+/// - bits 0..31: slot index
+/// - bits 32..62: generation
```
patrick-ogrady: I think its always worth adding a small note about how wrapping here (in such a way that we conflict and cause an issue) is infeasible.

runtime/src/iouring/mod.rs
```diff
@@ -847,143 +559,192 @@ impl IoUringLoop {
...
+    ///
+    /// Stops when all queued cancellations are staged or the SQ reaches
+    /// capacity. Returns `true` when SQ capacity is hit and at least one
+    /// cancellation remains queued.
+    fn stage_cancellations(
+        &mut self,
+        submission_queue: &mut SubmissionQueue<'_>,
+        pending_cancels: &mut VecDeque<WaiterId>,
+    ) -> bool {
+        while !submission_queue.is_full() {
```
patrick-ogrady, line 619: To confirm, this won’t cause a deadlock (with our inability to cancel anything) because some items must have a timeout registered on them?

runtime/src/iouring/mod.rs
```diff
@@ -747,14 +447,20 @@ impl IoUringLoop {
     /// This method blocks the current thread.
     pub(crate) fn run(mut self) {
```
patrick-ogrady, line 450: Got this feedback from a bot:
```
  - [P1] Avoid starting deadline tracking before the SQE reaches the kernel — /Users/patrickogrady/code/monorepo-pr-3284/runtime/src/iouring/
    mod.rs:587-590
    Scheduling the waiter in timeout_wheel here starts the deadline before the SQE has actually been submitted. That interacts badly with run()'s
    existing submitted != processed_seq fast-path: under a steady stream of producers, the loop can keep draining the MPSC for multiple
    iterations without ever calling submit_and_wait(). If one of these locally-staged SQEs expires in that window, advance_timeouts() only queues
    an AsyncCancel; because the kernel has never seen the original SQE, that cancel comes back as ENOENT, and the original request is still
    submitted later. In practice, short-deadline ops on a backlogged ring can complete successfully well after their deadline instead of
    returning ETIMEDOUT.
```

With the fix:

```
• I’d fix it by making “deadline-active” mean “submitted to the kernel,” not merely “pushed into the userspace SQ.”

  Right now runtime/src/iouring/mod.rs:520 inserts the waiter and immediately schedules its target_tick in the wheel at runtime/src/iouring/
  mod.rs:587. That is too early. The minimal fix is:

  1. Have fill_submission_queue() report whether it staged any deadline-bearing SQE this pass.
  2. If it did, force a submit before taking the submitted != processed_seq fast-path in runtime/src/iouring/mod.rs:472. I would add a small non-
     blocking submit() helper next to runtime/src/iouring/mod.rs:783 and call it once per pass.
  3. Keep the existing wheel/cancel logic otherwise.

  That’s the smallest diff that fixes the actual bug: by the time runtime/src/iouring/mod.rs:699 can see the deadline on the next iteration, the
  original SQE has already been flushed to the kernel, so AsyncCancel can’t race against a never-submitted op.

  If you want the invariant to be explicit rather than implicit, the cleaner version is to split waiter state into “staged” vs “submitted” and
  only call timeout_wheel.schedule() after a successful submit syscall. That is more code, but it exactly matches the intended semantics and is
  probably what I’d do if you expect more queueing changes here.

  I’d also add a regression test that keeps the loop on the submitted != processed_seq path while a short-deadline op is staged, then verifies it
  returns ETIMEDOUT instead of later succeeding.
```

runtime/src/storage/iouring.rs
andresilva: In theory yes, in practice we'd need to do an audit of the entire repo to handle this properly. We currently mostly treat storage errors as non-recoverable, we just expect things not to fail and if they do we shut down (e.g. https://github.com/commonwarexyz/monorepo/blob/f84f4224afec01f2b14497706547273cb7880ad3/consensus/src/simplex/actors/voter/actor.rs#L233-L241).

runtime/src/iouring/mod.rs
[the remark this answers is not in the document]
andresilva: https://github.com/commonwarexyz/monorepo/blob/f84f4224afec01f2b14497706547273cb7880ad3/runtime/src/iouring/mod.rs#L78-L81

runtime/src/iouring/waiter.rs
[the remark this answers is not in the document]
andresilva: I updated this to not use `abs`.

runtime/src/network/iouring.rs
andresilva: This is done for convenience to make sure that the io_uring `max_op_timeout` is scaled appropriately to whatever is configured for network `read_write_timeout`. We can instead just validate this and panic if `max_op_timeout < read_write_timeout`.

runtime/src/network/iouring.rs
andresilva: The iouring loop runs on a bare OS thread with no tokio runtime context so there's no simulated clock for `tokio::time::Instant` to hook into. I think it would still work since we don't use any of that functionality, but it's a bit misleading IMO.

runtime/src/iouring/mod.rs
andresilva: If I understood your question the potential deadlock scenario is: all waiter slots are occupied by stuck operations, so no new operations (including cancels) can be staged. The recovery path depends on two things:

1. Stuck operations have timeouts, so `advance_timeouts` eventually fires and pushes them to `pending_cancels`
2. Cancellations are staged before new operations, so they always get first priority on an empty SQ
3. The SQ is always drained at every iteration when full

This guarantees that cancel SQEs can always be submitted, which frees waiter capacity and unblocks the queue. As established in the liveness model, without timeouts we could have deadlocks (but in that case there are no cancellations either).

runtime/src/iouring/mod.rs
andresilva: This is very unlikely to happen, the operations we're talking about here are: a loop that will do at most say ~10k iterations (I don't think we'll ever use rings with capacity higher than that), at each iteration we'll read from a channel, read from completion queue, write into submission queue. This is something in the order of magnitude of nanoseconds, while the timeout wheel already has coarse precision and is configured with a default tick of 5ms (that's the minimum timeout resolution we can measure). The suggested fix basically gets rid of batching, since if timeouts are enabled we'd always submit eagerly. Either way, I pushed a fix for this, there's now a queue of pending deadlines that gets "flushed" into the timeout wheel only at the time when we submit the SQEs.

runtime/src/iouring/mod.rs
```diff
@@ -1024,11 +781,25 @@ impl IoUringLoop {
...
-        timeout: Option<Duration>,
+        budget: Option<Duration>,
     ) -> Result<bool, std::io::Error> {
+        // Activate buffered deadlines now that their SQEs are staged and about
+        // to be submitted. No `advance_timeouts` can run between this flush and
+        // the submit below, so deadlines cannot fire before the kernel sees the
+        // SQEs.
+        for (waiter_id, target_tick) in self.pending_deadlines.drain(..) {
+            self.timeout_wheel.schedule(waiter_id, target_tick);
+        }
```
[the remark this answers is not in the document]
andresilva: This should be fine due to timeout wheel's coarse granularity. The loop above will take at most nanoseconds to run, timeout wheel granularity is configured at 5ms by default (and it's not meant to go below millisecond accuracy).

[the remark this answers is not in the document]
andresilva: I still don't like this anyway, will try to come up with a more structural fix.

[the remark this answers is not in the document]
andresilva: I'm actually reverting all these changes, they make the code uglier (and slower), and the original issue is not real. Posting reasoning above.

runtime/src/iouring/mod.rs
andresilva: I reverted all the changes I made to handle this. The scenario described here can't actually lead to the cancel getting `ENOENT` for a never-submitted SQE, the reason is that the original SQE and the cancel SQE always end up in the same SQ buffer:

1. Iteration N: `fill_submission_queue` pushes the original SQE to the SQ buffer and schedules the deadline
2. continue path - no `submit_and_wait`, SQE stays in the SQ buffer
3. Iteration N+1: `advance_timeouts` fires the deadline, `fill_submission_queue` pushes the cancel SQE to the same SQ buffer
4. `submit_and_wait` eventually submits both - the kernel processes them in SQ order: original first, cancel second

The cancel always finds the original because they're batched. They can never be in different batches: if `submit_and_wait` ran between steps 1 and 3, the original would already be in-flight in the kernel, and the cancel would find it there.

The only real consequence of scheduling before submission is minor timeout drift, the deadline starts ticking microseconds before the SQE reaches the kernel. This is well within the timeout wheel's tick granularity (default 5ms).

runtime/src/network/iouring.rs
patrick-ogrady: Ah, good catch.

runtime/src/network/iouring.rs
patrick-ogrady: Given we can't yet specify these in runtime, I think this is the right approach (cc #1045)

runtime/src/iouring/mod.rs
patrick-ogrady: Took the time (to finally) dig into this. I agree with your assessment.

runtime/src/iouring/mod.rs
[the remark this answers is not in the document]
patrick-ogrady: Opted to address this directly in #3454 (to ensure we eventually exited when waiting on non-responding peers without a shutdown timeout)

Issue #3232, opened 2026-02-24
[Low] [qmdb] apply_operations panics on empty operation batch


dnkolegov-ar:
`apply_operations` in `storage/src/qmdb/sync/engine.rs` may panic in [312](https://github.com/commonwarexyz/monorepo/blob/7b2a721c960a93b2cb94ffc7259e0bb3531d2920/storage/src/qmdb/sync/engine.rs#L312) and in [324](https://github.com/commonwarexyz/monorepo/blob/7b2a721c960a93b2cb94ffc7259e0bb3531d2920/storage/src/qmdb/sync/engine.rs#L324) if there are no operations

PR #3286, opened 2026-03-03
[storage/qmdb/sync] fix panic on empty operations

roberto-bayardo:
Resolves: https://github.com/commonwarexyz/monorepo/issues/3232

### DIFF
--- a/storage/src/qmdb/any/sync/tests.rs
+++ b/storage/src/qmdb/any/sync/tests.rs
@@ -129,7 +129,7 @@ where
             fetch_batch_size: NZU64!(10),
             target: Target {
                 root: Digest::from([1u8; 32]),
-                range: Location::new_unchecked(0)..Location::new_unchecked(10),
+                range: Location::new(0)..Location::new(10),
             },
             context: context.with_label("client"),
             resolver: Arc::new(target_db),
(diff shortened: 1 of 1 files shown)


storage/src/qmdb/any/sync/tests.rs
```diff
@@ -110,6 +110,48 @@ pub(crate) trait SyncTestHarness: Sized + 'static {
     ) -> impl std::future::Future<Output = Self::Db> + Send;
 }
 
+/// Test that empty operations arrays fetched do not cause panics when stored and applied
```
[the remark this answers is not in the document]
roberto-bayardo: this is fine.

storage/src/qmdb/sync/engine.rs
```diff
@@ -293,8 +297,11 @@ where
         })
     }
 
-    /// Store a batch of fetched operations
+    /// Store a batch of fetched operations. If the input list is empty, this is a no-op.
     pub fn store_operations(&mut self, start_loc: Location, operations: Vec<DB::Op>) {
```
danlaine: Should we just make `store_operations` and `apply_operations` non-`pub`? I don't think they should ever be called from outside.

roberto-bayardo: You would know best! Changed to pub(crate) since it's needed in tests outside the module.  I left the non-empty invariant checking though.

PR #3289, opened 2026-03-03
[workspace] make clippy happy

andresilva:


storage/src/mmr/mem.rs
```diff
@@ -441,18 +441,14 @@ impl<D: Digest> DirtyMmr<D> {
     /// Add `digest` as a new leaf in the MMR, returning its position.
     pub(crate) fn add_leaf_digest(&mut self, digest: D) -> Position {
         // Compute the new parent nodes, if any.
-        let nodes_needing_parents = nodes_needing_parents(self.peak_iterator())
-            .into_iter()
-            .rev();
+        let nodes_needing_parents = nodes_needing_parents(self.peak_iterator()).len() as u32;
```
andresilva, line 444: It's not obvious if this method actually needs to return the nodes or only the count (it's only used here). But I didn't want to make that change without confirmation.

PR #3297, opened 2026-03-03
[storage] Add `MultiArchive`

clabby:
## Overview

Extends `prunable::Archive` to support storing multiple items at the same index via a new `MultiArchive` trait. The existing Archive trait retains single-item-per-index semantics (put is a no-op if the index exists). The on-disk format is unchanged.

### DIFF
--- a/storage/src/archive/mod.rs
+++ b/storage/src/archive/mod.rs
@@ -1101,12 +1101,36 @@ mod tests {
     }
 
     #[allow(dead_code)]
+    fn assert_multi_archive_futures_are_send<T: super::MultiArchive>(
+        archive: &mut T,
+        key: T::Key,
+        value: T::Value,
+    ) where
+        T::Key: Clone,
+        T::Value: Clone,
+    {
+        assert_archive_futures_are_send(archive, key.clone(), value.clone());
+        assert_send(archive.get_all(1));
+        assert_send(archive.put_multi(1, key.clone(), value.clone()));
+        assert_send(archive.put_multi_sync(2, key, value));
+    }
+
+    #[allow(dead_code)]
     fn assert_prunable_archive_futures_are_send(
         archive: &mut prunable::Archive<TwoCap, Context, FixedBytes<64>, i32>,
         key: FixedBytes<64>,
         value: i32,
     ) {
         assert_archive_futures_are_send(archive, key, value);
+    }
+
+    #[allow(dead_code)]
+    fn assert_prunable_multi_archive_futures_are_send(
+        archive: &mut prunable::Archive<TwoCap, Context, FixedBytes<64>, i32>,
+        key: FixedBytes<64>,
+        value: i32,
+    ) {
+        assert_multi_archive_futures_are_send(archive, key, value);
     }
 
     #[allow(dead_code)]


storage/src/archive/mod.rs
```diff
@@ -111,6 +111,45 @@ pub trait Archive: Send {
...
+/// [MultiArchive::put_multi] allows storing additional `(key, value)` pairs
+/// at an existing index. Each key must still be globally unique. This is useful
+/// for workloads where a logical grouping (e.g., a consensus view) may contain
+/// multiple entries.
+pub trait MultiArchive: Archive {
+    /// Store an item, allowing multiple items at the same index.
+    ///
+    /// Multiple items may share the same `index` as long as their keys differ.
+    /// If the same key already exists (at any index), the call is a no-op.
+    fn put_multi(
```
clabby, line 138: While this function may look very similar to `Archive::put`,  I think it's worth having a different version. The difference in the contract is that this skips the check in `Archive::put` where if the index already exists, the `put` is a no-op.

storage/src/archive/prunable/storage.rs
```diff
@@ -115,8 +115,11 @@ pub struct Archive<T: Translator, E: BufferPooler + Storage + Metrics, K: Array,
     /// Maps translated key representation to its corresponding index.
     keys: Index<T, u64>,
 
-    /// Maps index to position in index journal.
-    indices: BTreeMap<u64, u64>,
+    /// Maps index to position(s) in index journal.
+    ///
+    /// Each index may have multiple positions when used via [crate::archive::MultiArchive].
+    /// When used via [crate::archive::Archive], each index has exactly one position.
+    indices: BTreeMap<u64, Vec<u64>>,
```
patrick-ogrady: We should be able to avoid creating a `Vec<>` here by default.

storage/src/archive/mod.rs
patrick-ogrady: > in favor of only checking that the key is globally unique.

I don't think we ever do this check (nor do I recommend we add it).

clabby: (Comment was out of date, we don't. It just removes the check on global index uniqueness.)

storage/src/archive/mod.rs
```diff
@@ -111,6 +115,47 @@ pub trait Archive: Send {
...
+        &mut self,
+        index: u64,
+        key: Self::Key,
+        value: Self::Value,
+    ) -> impl Future<Output = Result<(), Error>> + Send {
+        async move {
+            self.put_multi(index, key, value).await?;
+            self.sync().await
+        }
+    }
```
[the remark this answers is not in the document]
clabby: Used in my other PR.

storage/src/archive/prunable/storage.rs
```diff
@@ -115,9 +115,13 @@ pub struct Archive<T: Translator, E: BufferPooler + Storage + Metrics, K: Array,
     /// Maps translated key representation to its corresponding index.
     keys: Index<T, u64>,
 
-    /// Maps index to position in index journal.
+    /// Maps index to its first position in the index journal.
     indices: BTreeMap<u64, u64>,
 
+    /// Additional positions for indices that have more than one entry.
+    /// Only populated when used via [crate::archive::MultiArchive::put_multi].
+    extra_indices: BTreeMap<u64, Vec<u64>>,
```
clabby, line 123: We use two maps here to prevent over-allocating in the common case. With an enum like:

```rust
enum Index {
    Single(u64),
    Multi(Vec<u8>)
}
```

the memory size is the largest variant (`Vec<u8>`)

```rs
core::mem::size_of<Index>() == 24 // Index::Multi = ptr(u64) + length(u64) + capacity(u64)
```

So, in the common case, we'd allocate 16B more per index. Since we only allocate extra indices when needed in a separate map, the memory use of standard `prunable::Archive`s is unchanged.

patrick-ogrady: Nice

storage/src/archive/prunable/storage.rs
```diff
@@ -435,6 +480,35 @@ impl<T: Translator, E: BufferPooler + Storage + Metrics, K: Array, V: CodecShare
     }
 }
 
+impl<T: Translator, E: BufferPooler + Storage + Metrics, K: Array, V: CodecShared>
+    crate::archive::MultiArchive for Archive<T, E, K, V>
+{
+    async fn get_all(&self, index: u64) -> Result<Option<Vec<V>>, Error> {
+        self.gets.inc();
+
+        if !self.indices.contains_key(&index) {
```
patrick-ogrady, line 491: add inline comments around each section like other funcs?

storage/src/archive/prunable/storage.rs
```diff
@@ -115,9 +115,13 @@ pub struct Archive<T: Translator, E: BufferPooler + Storage + Metrics, K: Array,
     /// Maps translated key representation to its corresponding index.
     keys: Index<T, u64>,
 
-    /// Maps index to position in index journal.
+    /// Maps index to its first position in the index journal.
     indices: BTreeMap<u64, u64>,
 
+    /// Additional positions for indices that have more than one entry.
```
patrick-ogrady, line 121: I suppose the other option here would be to conditionally initialize indices based on whether or not this could happen (but that assumes it is likely to happen)?

patrick-ogrady: This works well if unlikely to happen but covered if it does (which is what we need for marshal).

patrick-ogrady: Maybe we should document this bias towards "supported by not optimized for" somewhere?

storage/src/archive/prunable/mod.rs
patrick-ogrady, line 1: Should we move these tests to the archive level (I think we do that for the archive trait)?

PR #3303, opened 2026-03-04
[cryptography] avoid redundant signature clone in certificate assemble

VolodymyrBg:
[assemble](https://github.com/commonwarexyz/monorepo/blob/240e0207ee10c3c37a42867ce4de97b581c06b32/cryptography/src/ed25519/certificate/mod.rs#L223-L253) in the secp256r1 and ed25519 certificate schemes was decoding and cloning each attestation signature, then wrapping it back into `Lazy`.
That added unnecessary copy work and extra processing in a hot path.

Preserve the existing validation behavior ([signature.get()?](https://github.com/commonwarexyz/monorepo/blob/240e0207ee10c3c37a42867ce4de97b581c06b32/utils/src/ordered.rs#L77-L80)) and move `Lazy` signatures directly into the certificate payload.
This removes the redundant decode/clone/rewrap step in secp256r1 and ed25519 certificate assembly.

### DIFF
--- a/cryptography/src/ed25519/certificate/mod.rs
+++ b/cryptography/src/ed25519/certificate/mod.rs
@@ -233,7 +233,7 @@ impl<N: Namespace> Generic<N> {
             if usize::from(signer) >= self.participants.len() {
                 return None;
             }
-            let signature = signature.get().cloned()?;
+            signature.get()?;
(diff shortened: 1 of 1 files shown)


clabby:

PR #3305, opened 2026-03-05
[runtime/iobuf] support split `IoBuf`/`IoBufs`

andresilva:
Extracted from #3185.
Related #3093 #2883.

### DIFF
--- a/runtime/src/iobuf/mod.rs
+++ b/runtime/src/iobuf/mod.rs
@@ -123,7 +123,7 @@ impl IoBuf {
             return Self::default();
         }
 
-        if at == self.len() {
+        if at == self.remaining() {
             return std::mem::take(self);
         }
 


runtime/src/iobuf/mod.rs
```diff
@@ -108,6 +108,33 @@ impl IoBuf {
...
+    ///
+    /// # Panics
+    ///
+    /// Panics if `at > len`.
+    pub fn split_to(&mut self, at: usize) -> Self {
+        if at == 0 {
+            return Self::default();
+        }
+
+        if at == self.len() {
```
patrick-ogrady: I suppose the assertion (of at > self.len() will be hit below)?

runtime/src/iobuf/mod.rs
```diff
@@ -108,6 +108,33 @@ impl IoBuf {
...
+    /// This is an `O(1)` zero-copy operation.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `at > len`.
+    pub fn split_to(&mut self, at: usize) -> Self {
+        if at == 0 {
+            return Self::default();
+        }
+
```
patrick-ogrady, line 125: Do we want to do self.remaining() here for consistency?

runtime/src/iobuf/mod.rs
```diff
@@ -882,6 +909,145 @@ impl IoBufs {
...
+            "split_to out of bounds: {:?} <= {:?}",
+            at,
+            remaining,
+        );
+
+        if at == remaining {
+            return std::mem::take(self);
+        }
+
+        let inner = std::mem::replace(&mut self.inner, IoBufsInner::Single(IoBuf::default()));
```
patrick-ogrady, line 940: I'm assuming this doesn't alloc until needed?

runtime/src/iobuf/mod.rs
andresilva: Yes, exactly.

runtime/src/iobuf/mod.rs
andresilva: Yes, `IoBuf::default()` is `Bytes::new()` which doesn't allocate (https://docs.rs/bytes/latest/bytes/struct.Bytes.html#method.new).

runtime/src/iobuf/mod.rs
andresilva: Updated.

PR #3306, opened 2026-03-05
[runtime/utils/buffer/read] avoid copying

andresilva:
Extracted from #3185.
Related #3093 #2883.

### DIFF
--- a/runtime/fuzz/fuzz_targets/buffer.rs
+++ b/runtime/fuzz/fuzz_targets/buffer.rs
@@ -37,11 +37,7 @@ enum FuzzOperation {
         cache_page_size: u16,
         cache_capacity: u16,
     },
-    ReadExact {
-        size: u16,
-    },
-    ReadExactRandomBuf {
-        buf: Vec<u8>,
+    Read {
         size: u16,
     },
     ReadSeekTo {
@@ -192,23 +188,13 @@ fn fuzz(input: FuzzInput) {
                     }
                 }
 
-                FuzzOperation::ReadExact { size } => {
+                FuzzOperation::Read { size } => {
                     if let Some(ref mut reader) = read_buffer {
                         let size = (size as usize).clamp(0, MAX_SIZE);
                         let current_pos = reader.position();
                         if current_pos.checked_add(size as u64).is_some() {
-                            let mut buf = vec![0u8; size];
-                            let _ = reader.read_exact(&mut buf, size).await;
+                            let _ = reader.read(size).await;
                         }
-                    }
-                }
-
-                FuzzOperation::ReadExactRandomBuf { mut buf, size } => {
-                    if size > buf.len() as u16 {
-                        continue;
-                    }
-                    if let Some(ref mut reader) = read_buffer {
-                        let _ = reader.read_exact(&mut buf, size as usize).await;
                     }
                 }
 
--- a/runtime/src/utils/buffer/write.rs
+++ b/runtime/src/utils/buffer/write.rs
@@ -32,9 +32,8 @@ use std::{num::NonZeroUsize, sync::Arc};
 ///     // Read back the data to verify
 ///     let (blob, size) = context.open("my_partition", b"my_data").await.expect("unable to reopen blob");
 ///     let mut reader = Read::from_pooler(&context, blob, size, NZUsize!(8));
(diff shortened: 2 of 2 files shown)


runtime/src/utils/buffer/read.rs
```diff
@@ -113,62 +121,66 @@ impl<B: Blob> Read<B> {
...
-        let mut buf = std::mem::take(&mut self.buffer);
-        buf.clear();
-        let buf = if buf.capacity() >= bytes_to_read {
-            buf
-        } else {
-            self.pool.alloc(bytes_to_read)
+        // Reuse existing allocation when uniquely owned. If readers still hold slices from
+        // previous reads, allocate a pooled replacement and leave old memory alive until dropped.
+        let current = std::mem::take(&mut self.buffer);
+        let buf = match current.try_into_mut() {
```
patrick-ogrady, line 127: In a future PR, I suspect having a metric around this will be very interesting.

runtime/src/utils/buffer/read.rs
```diff
@@ -113,62 +121,66 @@ impl<B: Blob> Read<B> {
...
-            // Calculate how many bytes we can copy from the buffer
-            let bytes_to_copy = std::cmp::min(size - bytes_read, self.buffer_remaining());
+            // Calculate how many bytes we can take from the buffer
+            let bytes_to_take = std::cmp::min(remaining, self.buffer_remaining());
 
-            // Copy bytes from buffer to output
-            buf[bytes_read..(bytes_read + bytes_to_copy)].copy_from_slice(
-                &self.buffer.as_ref()[self.buffer_position..(self.buffer_position + bytes_to_copy)],
+            // Append bytes from buffer to output
+            out.append(
```
patrick-ogrady, line 174: so sick

storage/src/freezer/storage.rs
```diff
@@ -1009,22 +1006,13 @@ impl<E: BufferPooler + Storage + Metrics + Clock, K: Array, V: CodecShared> Free
         // Read the entire chunk
         let chunk_bytes = chunk_size as usize * Entry::FULL_SIZE;
         let read_offset = Self::table_offset(current_index);
-        let read_buf = self
-            .table
-            .read_at(read_offset, chunk_bytes)
-            .await?
-            .coalesce();
+        let mut read_buf = self.table.read_at(read_offset, chunk_bytes).await?;
```
patrick-ogrady, line 1009: ❤️ ❤️ ❤️

runtime/src/utils/buffer/read.rs
```diff
@@ -113,62 +121,66 @@ impl<B: Blob> Read<B> {
...
-        Ok(bytes_to_read)
+        Ok(self.buffer_valid_len)
     }
 
-    /// Reads exactly `size` bytes into the provided buffer. Returns an error if not enough bytes
-    /// are available.
+    /// Reads exactly `len` bytes and returns them as immutable bytes.
     ///
-    /// # Panics
+    /// Returned bytes are composed of zero-copy slices from the internal read buffer.
```
patrick-ogrady, line 146: I hope this doesn't end up being too much of a footfall (where we end up accidentally using `Bytes` somewhere important on a codec-able type).

Maybe we should add some note to AGENTS.md as a warning too?

patrick-ogrady: I suppose the pool architecture invites us to copy-out rather than retain (but we wouldn't see much regression if we decide to retain as it should seamlessly transition into just one alloc per incoming item?)

runtime/src/utils/buffer/read.rs
andresilva: In this particular case we won't be able to reuse the allocation when a read crosses a refill boundary. Assuming that reads are usually smaller than the read buffer it's still worth it to do this, because you avoid copies on all the intermediary reads. And in the fallback case we still fetch a new buffer from the pool anyway.

runtime/src/utils/buffer/read.rs
andresilva: I think the way to monitor this is through the pool metrics, if you consistently exhaust the number of buffers either your pool isn't appropriately sized for the level of concurrency you're running into, or there are buffers being held with a long lifetime. But yes, in any case we would just revert to doing direct heap allocation.

PR #3307, opened 2026-03-05
[runtime/utils/buffer/{write, tip}] avoid copying

### LINKED ISSUE #2731: [runtime/utils/buffer] Introduce distinct state-types for immutable vs non-immutable Append blobs
Currently immutable state is determined by boolean, we might instead use distinct state types that can be compiler-enforced like we do with the adbs and clean/dirty mmrs.

andresilva:
Extracted from #3185.
Depends on #3306.
Related #3093 #2883.
Closes #2731

### DIFF
--- a/runtime/src/utils/buffer/mod.rs
+++ b/runtime/src/utils/buffer/mod.rs
@@ -54,12 +54,10 @@ mod tests {
         async fn block_once_on_read(&self) {
             let rx = {
                 let mut gate = self.gate.lock();
-                if let Some(read_started) = gate.read_started.take() {
+                gate.read_started.take().map(|read_started| {
                     let _ = read_started.send(());
-                    Some(gate.release_read.take().expect("release signal missing"))
-                } else {
-                    None
-                }
+                    gate.release_read.take().expect("release signal missing")
+                })
             };
             if let Some(rx) = rx {
                 let _ = rx.await;


andresilva:
I have confirmed that this PR fixes the performance regression in https://github.com/commonwarexyz/monorepo/issues/3093#issuecomment-3935958082.

storage/src/journal/segmented/manager.rs
```diff
@@ -18,24 +18,50 @@ use prometheus_client::metrics::{counter::Counter, gauge::Gauge};
 use std::{collections::BTreeMap, future::Future, mem::take, num::NonZeroUsize};
 use tracing::debug;
 
-/// A buffer that wraps a blob and provides size information.
+/// Minimal per-section buffer interface used by [`Manager`].
 ///
-/// Both [`Append`] and [`Write`] implement this trait.
-pub trait SectionBuffer: Blob + Clone + Send + Sync {
+/// [`Manager`] only needs lifecycle operations over a section: report logical size,
```
patrick-ogrady: This seems like a leak of prompt from an LLM.

runtime/src/utils/buffer/tip.rs
```diff
@@ -88,51 +161,43 @@ impl Buffer {
...
-        assert!(end_offset <= self.size());
-        if end_offset <= self.offset {
-            // Range does not overlap with the buffer.
-            return buf.len();
-        }
-
-        let (start, remaining) = if offset < self.offset {
-            // Some data is before the buffer.
-            (0, (self.offset - offset) as usize)
-        } else {
```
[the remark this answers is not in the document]
patrick-ogrady:
```
• No, not as a current correctness issue.

  - The method in tip.rs:110 is brittle: if a caller ever passed start.. or .. while self.data.len() > self.len after resize:138 or clear:276,
    forwarding range directly to self.data.slice(range) would indeed over-expose stale tail bytes.
  - But the current callers I found in write.rs:107 and write.rs:118 always use explicit end bounds (start..end and ..tip_len). Those remain
    capped even if self.data.len() exceeds the logical self.len.

  So the report describes a latent footgun in the helper, not a reachable medium-severity bug in this diff. If you want to harden it, changing
  the last line to self.data.slice(start..end) is reasonable, but I would not treat this comment as a valid blocking issue.
```

runtime/src/utils/buffer/paged/append.rs
```diff
@@ -231,23 +224,16 @@ impl<B: Blob> Append<B> {
...
-        if buf_guard.immutable {
+        if buf_guard.is_immutable() {
             return Ok(());
         }
-        buf_guard.immutable = true;
+        buf_guard.set_immutable(false);
         self.flush_internal(buf_guard, true).await?;
 
-        // Shrink the buffer to release the pooled allocation since we won't be appending.
+        // Compact tip backing after flush to match the post-flush logical view.
```
patrick-ogrady: It doesn't seem like we compact any longer?

runtime/src/utils/buffer/paged/append.rs
```diff
@@ -543,7 +538,18 @@ impl<B: Blob> Append<B> {
         }
 
         // Extract any bytes from the buffer that overlap with the requested range.
-        let remaining = buffer.extract(buf.as_mut(), logical_offset);
```
patrick-ogrady, line 546: why is this helper gone? it seemed nice?

runtime/src/utils/buffer/paged/append.rs
```diff
@@ -824,13 +803,6 @@ impl<B: Blob> Blob for Append<B> {
         blob_state.blob.sync().await
     }
 
-    /// This [Blob] trait method is unimplemented by [Append] and unconditionally panics.
-    async fn write_at(&self, _offset: u64, _bufs: impl Into<IoBufs> + Send) -> Result<(), Error> {
```
patrick-ogrady, line 828: ❤️

runtime/src/utils/buffer/tip.rs
```diff
@@ -29,10 +46,12 @@ pub(super) struct Buffer {
 impl Buffer {
     /// Creates a new buffer with the provided `offset` and `capacity`.
     ///
-    /// The backing buffer starts empty and grows on demand via `merge`/`append`.
+    /// The backing buffer is allocated eagerly and starts with zero length.
     pub(super) fn new(offset: u64, capacity: usize, pool: BufferPool) -> Self {
+        let data = pool.alloc(capacity).freeze();
```
patrick-ogrady: We should allocate when we need it (or be more explicit about creating immutable buffers)?

runtime/src/utils/buffer/tip.rs
```diff
@@ -88,51 +161,43 @@ impl Buffer {
...
-    pub(super) fn take(&mut self) -> Option<(IoBufMut, u64)> {
+    /// This returns a cloned `IoBuf` view of logical bytes, resets logical length to zero, and
+    /// advances offset to the end of the drained range.
+    pub(super) fn take(&mut self) -> Option<(IoBuf, u64)> {
         if self.is_empty() {
             return None;
         }
-        let replacement = self.pool.alloc(self.capacity);
-        let buf = std::mem::replace(&mut self.data, replacement);
+        let buf = self.data.slice(..self.len);
```
patrick-ogrady: We should return the bytes here and wait to refill until we need it imho

runtime/src/utils/buffer/paged/append.rs
andresilva: Because in the full page cache rewrite it's not used by append either, so I just inlined here for now. But feel free to revert.

runtime/src/utils/buffer/tip.rs
andresilva: If we set this to `IoBuf::default()` the code should work as is I think. Needs an update of allocation semantics docs.

runtime/src/utils/buffer/paged/append.rs
andresilva: We compact inside `set_immutable`.

patrick-ogrady:
(I'm working on this in #3314)

runtime/src/utils/buffer/paged/append.rs
```diff
@@ -231,23 +224,16 @@ impl<B: Blob> Append<B> {
         // Flush any buffered data. When flush_internal returns, write_at has completed and data
         // has been written to the underlying blob.
         let mut buf_guard = self.buffer.write().await;
-        if buf_guard.immutable {
+        if buf_guard.is_immutable() {
             return Ok(());
         }
-        buf_guard.immutable = true;
+        buf_guard.set_immutable(false);
```
clabby: This API is a bit unclear, had to look at the docs (I thought this meant “mark this as mutable”). Maybe we should add an `ImmutableMode { WithCompaction, WithoutCompaction }` enum to make this a bit more readable?

runtime/src/utils/buffer/mod.rs
```diff
@@ -11,9 +11,96 @@ pub use write::Write;
...
+            if end > self.data.len() {
+                return Err(Error::BlobInsufficientLength);
+            }
+
+            let mut out = buf.into();
+            out.put_slice(&self.data[start..end]);
+            Ok(out)
+        }
+
+        async fn write_at(
```
clabby: Codex:

`BlockingReadBlob::write_at` is a no-op, so the contention tests currently prove only lock blocking, not the "no torn read" property. They do not assert that a concurrent reader returns a consistent pre-write snapshot, nor that a subsequent reader observes the post-write bytes.

Consider adding a stateful test double whose `write_at` actually stores bytes, then assert:
1. A read blocked by a write lock observes only the pre-write state.
2. After the write completes, a new read observes the post-write state.

This would give the tests real end-to-end coverage of the ordering guarantee.

runtime/src/utils/buffer/tip.rs
```diff
@@ -88,51 +161,43 @@ impl Buffer {
...
-        let (start, remaining) = if offset < self.offset {
-            // Some data is before the buffer.
-            (0, (self.offset - offset) as usize)
-        } else {
-            // Can read entirely from the buffer.
-            ((offset - self.offset) as usize, 0)
+    /// - If backing is uniquely owned and has enough capacity, no allocation occurs.
+    /// - If backing is shared (for example, because a flushed/read view is still alive) or too
+    ///   small, a new pooled allocation is created and existing bytes are copied.
+    fn writable(&mut self, needed: usize) -> IoBufMut {
```
clabby, line 162: Codex:

The copy-on-write contract (when an outstanding `IoBuf` view exists from `take()` or `slice()`, `writable()` must allocate+copy rather than mutate shared backing) is the key correctness guarantee of this change, but no test explicitly exercises that path.

Add a focused unit test that:
1. Calls `take()` or `slice()` to obtain a view.
2. Calls `writable()` while the view is still live.
3. Asserts that the returned buffer is a distinct allocation and that the original view is unmodified.

runtime/src/utils/buffer/paged/append.rs
clabby: Codex:

`to_immutable()` sets the immutable flag, releases the lock, calls `flush_internal`, then reacquires the lock for compaction. The correctness depends on no `append()` or `to_mutable()` slipping in during that window, but this invariant is not tested.

Add a concurrency test that races `to_immutable()` with concurrent `append()`/`to_mutable()` calls and asserts that no append is accepted once the immutable transition begins. This would lock in the two-lock invariant before any future refactoring changes the window.

patrick-ogrady:
bugbot run

patrick-ogrady:
This ended up being even a performance improvement over `main` (before the copy regression). We no longer hold on to any memory we aren't actively using in any buffer. With the buffer pool, pending memory now "flows" to the blob that needs it (and then its given up).

PR #3308, opened 2026-03-05
[p2p] stop background receiver when consumer channel closes

sashass1315:
`WrappedBackgroundReceiver` previously ignored downstream closure and continued decoding/spawning work even after the consumer channel was dropped. This change makes decode handling return a continuation signal, exits the run loop when send_lossy indicates closure, and adds explicit closed-channel checks before scheduling new decode tasks. Added a regression test to ensure the background task exits promptly when the consumer is dropped.


clabby:

PR #3309, opened 2026-03-05
[storage/mmr] change pruning and sync API to be location instead of position based

roberto-bayardo:
Changes the MMR pruning and sync APIs to align the pruning boundary to a leaf (Location),  and to return Error on bad input instead of panicking (since it's pub).  This is more consistent with how proving works (over locations).  Also changes bounds() to be location based since begin/end are now always leaf aligned.

### DIFF
--- a/storage/src/mmr/journaled.rs
+++ b/storage/src/mmr/journaled.rs
@@ -594,11 +594,16 @@ impl<E: RStorage + Clock + Metrics, D: Digest> Mmr<E, D> {
     ///
     /// This implementation ensures that no failure can leave the MMR in an unrecoverable state,
     /// requiring it sync the MMR to write any potential unsynced updates.
+    ///
+    /// Returns [Error::LocationOverflow] if `loc` exceeds [crate::mmr::MAX_LOCATION].
+    /// Returns [Error::LeafOutOfBounds] if `loc` exceeds the current leaf count.
     pub async fn prune(&mut self, loc: Location) -> Result<(), Error> {
         let pos = Position::try_from(loc)?;
         {
             let inner = self.inner.get_mut();
-            assert!(loc <= inner.mem_mmr.leaves());
+            if loc > inner.mem_mmr.leaves() {
+                return Err(Error::LeafOutOfBounds(loc));
+            }
             if pos <= inner.pruned_to_pos {
                 return Ok(());
             }
@@ -1048,6 +1053,35 @@ mod tests {
                 &mut hasher,
                 &[] as &[(Digest, Location)],
                 &root
+            ));
+
+            mmr.destroy().await.unwrap();
+        });
+    }
+
+    #[test_traced]
+    fn test_journaled_prune_out_of_bounds_returns_error() {
+        let executor = deterministic::Runner::default();
+        executor.start(|context| async move {
+            let mut hasher = Standard::<Sha256>::new();
+            let mut mmr = Mmr::init(
+                context.with_label("oob_prune"),
+                &mut hasher,
+                test_config(&context),
+            )
+            .await
+            .unwrap();
+
+            let changeset = {
+                let mut batch = mmr.new_batch();
+                batch.add(&mut hasher, &test_digest(0));
+                batch.merkleize(&mut hasher).finalize()
+            };
+            mmr.apply(changeset).unwrap();
+
+            assert!(matches!(
+                mmr.prune(Location::new(2)).await,
+                Err(Error::LeafOutOfBounds(loc)) if loc == Location::new(2)
             ));
 
             mmr.destroy().await.unwrap();


roberto-bayardo:
bugbot run

storage/src/mmr/mem.rs
```diff
@@ -290,14 +292,20 @@ impl<D: Digest> CleanMmr<D> {
...
+    ///
+    /// Returns [Error::LocationOverflow] if `loc` exceeds [crate::mmr::MAX_LOCATION].
+    /// Returns [Error::LeafOutOfBounds] if `loc` exceeds the current leaf count.
+    pub fn prune(&mut self, loc: Location) -> Result<(), Error> {
+        if loc > self.leaves() {
+            return Err(Error::LeafOutOfBounds(loc));
+        }
+        let pos = Position::try_from(loc)?;
+        self.prune_to_pos(pos);
+        Ok(())
```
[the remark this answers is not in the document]
roberto-bayardo: good bot, fixed.

storage/fuzz/fuzz_targets/range_proof.rs
```diff
@@ -30,7 +30,7 @@ fn fuzz(input: FuzzInput) {
 
     let config = Config {
         nodes,
-        pruned_to_pos: Position::new(input.pruned_to_pos),
+        pruned_to: Location::new(input.pruned_to_pos),
```
patrick-ogrady: I think we need to change the name of of `pruned_to_pos`?

storage/src/mmr/journaled.rs
```diff
@@ -304,12 +320,12 @@ impl<E: RStorage + Clock + Metrics, D: Digest> Mmr<E, D> {
...
             MemConfig {
                 nodes: vec![],
-                pruned_to_pos: journal_size,
+                pruned_to: Location::try_from(journal_size).map_err(Error::from)?,
                 pinned_nodes,
             },
             hasher,
         )?;
-        let prune_pos = Position::new(effective_prune_pos);
+        let prune_pos = effective_prune_pos;
```
patrick-ogrady: does this need to be a new var?

storage/src/mmr/journaled.rs
patrick-ogrady, line 1: LLM claims an issue:

```
• The patch mostly updates pruning APIs from positions to locations, but it introduces one data-loss path in init_sync() for non-leaf boundaries and weakens mem::Mmr::prune()'s documented error handling for invalid locations. Those issues make the patch unsafe to consider fully correct.

  Full review comments:

  - [P2] Validate sync boundary before clearing persisted data — /Users/patrickogrady/code/monorepo-pr-3309/storage/src/mmr/journaled.rs:409-410
    If init_sync() is called with a non-leaf SyncConfig.range.start and the existing journal is empty or still behind that boundary, this branch calls clear_to_size(*cfg.range.start) before the later Location::try_from(cfg.range.start) validation runs. The function then returns an error
    after it has already truncated/reinitialized the persisted journal, so a caller using a node-position boundary (which the public Range<Position> API still permits) can lose on-disk data on a failed sync attempt.
  - [P3] Check prune input validity before returning LeafOutOfBounds — /Users/patrickogrady/code/monorepo-pr-3309/storage/src/mmr/mem.rs:331-334
    If a caller passes Location::new(u64::MAX) or any other value above MAX_LOCATION, this branch returns LeafOutOfBounds as soon as the tree has fewer leaves than that value, because the Position::try_from(loc) validation now happens afterwards. That breaks the documented LocationOverflow
    contract of Mmr::prune() and makes malformed locations indistinguishable from an ordinary out-of-range prune request.

```

storage/fuzz/fuzz_targets/range_proof.rs
roberto-bayardo: done

storage/src/mmr/journaled.rs
roberto-bayardo: no, removed

storage/src/mmr/journaled.rs
roberto-bayardo: Fair, init_sync should also use a Location based API.  Extended the PR to also update the init_sync API / Config to be Location basede.

The P3 there doesn't seem worth sweating over (you still get an error, and nobody should ever do this...)

storage/src/mmr/journaled.rs
```diff
@@ -549,21 +573,27 @@ impl<E: RStorage + Clock + Metrics, D: Digest> Mmr<E, D> {
...
+        // mem_mmr. We prune to the previously captured leaf count to avoid a race with concurrent
+        // appends between the read lock above and this write lock.
         {
             let mut inner = self.inner.write();
-            inner.mem_mmr.prune_to_pos(size);
+            let leaves = inner.mem_mmr.leaves();
+            inner
+                .mem_mmr
+                .prune(leaves)
+                .expect("captured leaves is in bounds");
```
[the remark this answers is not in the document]
roberto-bayardo: Ugg, I added that comment because the AIs keep re-introducing this bug, and it did it yet again. Sigh.

Fixed.

PR #3311, opened 2026-03-05
[runtime/utils/buffer/paged/cache] cleanup page fetch entries on cancellation

andresilva:


runtime/src/utils/buffer/paged/cache.rs
```diff
@@ -23,6 +23,42 @@ use tracing::{debug, error, trace};
 // [Shared]. The IoBuf contains only the logical (validated) bytes of the page.
 type PageFetchFut = Shared<Pin<Box<dyn Future<Output = Result<IoBuf, Arc<Error>>> + Send>>>;
 
+type PageFetch = Arc<InFlightPageFetch>;
+
+struct InFlightPageFetch {
```
patrick-ogrady: This ensures any fetcher can handle (even if the first is canceled)?

patrick-ogrady:
> Still more to do:
> 
> ```
> • The cancellation cleanup is headed in the right direction, but the new waiter-counted removal logic still has a race that can drop an active
>   in-flight fetch and allow duplicate reads for the same page. The success path also serializes all followers on the cache write lock, which is a
>   meaningful regression for a performance-sensitive hot path.
> 
>   Full review comments:
> 
>   - [P2] Recheck waiter count before removing the in-flight fetch — /Users/patrickogrady/code/monorepo-pr-3311/runtime/src/utils/buffer/paged/
>     cache.rs:48-57
>     When the last known waiter drops at the same time another reader is entering read_after_page_fault for the same uncached page, that reader
>     can clone the existing PageFetch and increment waiters from 0 back to 1 before this drop acquires cache.write(). This code still removes
>     page_fetches[self.key], so the next miss starts a second physical fetch even though another task is already awaiting the original future.
>     That breaks the single-flight guarantee after cancellations and can duplicate blob I/O for the same page.
>   - [P2] Drop the cache write lock before copying fetched bytes — /Users/patrickogrady/code/monorepo-pr-3311/runtime/src/utils/buffer/paged/
>     cache.rs:335-343
>     When many readers block on the same page fault, each waiter now acquires cache.write() here and keeps it until after copy_from_slice.
>     Followers do not need the lock once the page is cached, so this serializes them behind a blocking RwLock and also stalls unrelated cache
>     traffic on other pages. On hot pages this turns a single shared fetch into one lock handoff per waiter, which is a noticeable throughput
>     regression in the cache fast path.
> ```

I introduced those...my b:

```
  I compared 61a3f8476c564ed9007c927eb84385f106136777 to its parent ae12eeff7 locally. Both review findings come from code added in 61a3f847, not
  from the parent.

  The first one is clearly introduced there: the new waiters: AtomicUsize plus “last waiter removes the map entry” logic was added at cache.rs:26
  and cache.rs:46. ae12eeff7 did not have a waiter count at all, so the 0 -> 1 rejoin race the reviewer called out did not exist in that form
  there.

  The second one is also introduced there, with one nuance. The parent already held the write lock through copy_from_slice for the first fetcher,
  but followers returned without taking that lock. In 61a3f847, every successful waiter now goes through cache.rs:334 to cache.rs:343, so the
  contention regression is new because followers are now serialized on cache.write() too.

  So the clean summary is: ae12eeff7 had a different bug, and 61a3f847 fixed that by introducing a new design that in turn introduced these two
  new issues.
```

runtime/src/utils/buffer/paged/cache.rs
```diff
@@ -23,6 +23,37 @@ use tracing::{debug, error, trace};
...
+    }
+}
+
+impl Drop for PageFetchGuard {
+    fn drop(&mut self) {
+        if !self.armed {
+            return;
+        }
+        let mut cache = self.cache.write();
+        let _ = cache.page_fetches.remove(&self.key);
```
clabby: `PageFetchGuard::drop()` always removes the in-flight entry on owner cancellation, even when other waiters are still awaiting the same shared future. That can create a race where a subsequent reader sees no `page_fetches` entry and starts a second underlying blob read for the same `(blob_id, page_num)`.

```rust
/// A blob that blocks all reads and counts how many were started.
#[derive(Clone)]
struct CountingBlockingBlob {
    started: Arc<Mutex<Option<oneshot::Sender<()>>>>,
    started_second: Arc<Mutex<Option<oneshot::Sender<()>>>>,
    reads_started: Arc<AtomicUsize>,
}

impl Blob for CountingBlockingBlob {
    async fn read_at(&self, offset: u64, len: usize) -> Result<IoBufsMut, Error> {
        self.read_at_buf(offset, len, IoBufsMut::default()).await
    }

    async fn read_at_buf(
        &self,
        _offset: u64,
        _len: usize,
        _bufs: impl Into<IoBufsMut> + Send,
    ) -> Result<IoBufsMut, Error> {
        let read_number = self.reads_started.fetch_add(1, Ordering::Relaxed) + 1;
        if read_number == 1 {
            let sender = self.started.lock().take().unwrap();
            let _ = sender.send(());
        } else if read_number == 2 {
            let sender = self.started_second.lock().take().unwrap();
            let _ = sender.send(());
        }
        pending::<()>().await;
        unreachable!()
    }

    async fn write_at(
        &self,
        _offset: u64,
        _bufs: impl Into<crate::IoBufs> + Send,
    ) -> Result<(), Error> {
        Ok(())
    }

    async fn resize(&self, _len: u64) -> Result<(), Error> {
        Ok(())
    }

    async fn sync(&self) -> Result<(), Error> {
        Ok(())
    }
}

#[test_traced]
fn test_cancelled_first_fetcher_can_trigger_duplicate_underlying_fetches() {
        let blob_id = 0;
        let cache_ref = CacheRef::from_pooler(&context, PAGE_SIZE, NZUsize!(10));
        let (started_tx, started_rx) = oneshot::channel();
        let (started_second_tx, mut started_second_rx) = oneshot::channel();
        let reads_started = Arc::new(AtomicUsize::new(0));
        let blob = CountingBlockingBlob {
            started: Arc::new(Mutex::new(Some(started_tx))),
            started_second: Arc::new(Mutex::new(Some(started_second_tx))),
            reads_started: reads_started.clone(),
        };

        let cache_ref_for_first = cache_ref.clone();
        let blob_for_first = blob.clone();
        let first = context.clone().spawn(move |_| async move {
            let mut buf = vec![0u8; PAGE_SIZE.get() as usize];
            let _ = cache_ref_for_first
                .read(&blob_for_first, blob_id, &mut buf, 0)
                .await;
        });

        started_rx.await.expect("first blocking read never started");

        // Join as a waiter on the same in-flight shared future.
        let cache_ref_for_second = cache_ref.clone();
        let blob_for_second = blob.clone();
        let second = context.clone().spawn(move |_| async move {
            let mut buf = vec![0u8; PAGE_SIZE.get() as usize];
            let _ = cache_ref_for_second
                .read(&blob_for_second, blob_id, &mut buf, 0)
                .await;
        });

        // Cancel the owner.
        first.abort();
        assert!(matches!(first.await, Err(Error::Closed)));

        // Trigger another read while the old shared fetch is still pending via `second`.
        let cache_ref_for_third = cache_ref.clone();
        let blob_for_third = blob.clone();
        let third = context.clone().spawn(move |_| async move {
            let mut buf = vec![0u8; PAGE_SIZE.get() as usize];
            let _ = cache_ref_for_third
                .read(&blob_for_third, blob_id, &mut buf, 0)
                .await;
        });

        // Give spawned tasks a chance to run and observe whether a second underlying read starts.
        context.sleep(Duration::from_millis(1)).await;
        let second_started = started_second_rx.try_recv().is_ok();
        assert!(
            !second_started,
            "a second underlying read started for the same page while one was already in-flight"
        );
        assert_eq!(reads_started.load(Ordering::Relaxed), 1);

        second.abort();
        assert!(matches!(second.await, Err(Error::Closed)));
        third.abort();
        assert!(matches!(third.await, Err(Error::Closed)));
    });
}
```

We may want to keep `page_fetches` populated until the last waiter/owner drops (for example, store per-entry waiter count/owner state), so cancellation cleanup does not remove dedup state while the original shared fetch is still active?

runtime/src/utils/buffer/paged/cache.rs
```diff
@@ -234,79 +320,68 @@ impl CacheRef {
...
 
-            let entry = cache.page_fetches.entry((blob_id, page_num));
-            match entry {
+            let key = (blob_id, page_num);
+            match cache.page_fetches.entry(key) {
                 Entry::Occupied(o) => {
                     // Another thread is already fetching this page, so clone its existing future.
-                    (o.get().clone(), false)
+                    let fetch = Arc::clone(o.get());
+                    fetch.waiters.fetch_add(1, Ordering::Relaxed);
```
clabby: `waiters` is only mutated while `cache.write()` is held, but the code uses `fetch_add(..., Relaxed)` paired with `fetch_sub(..., AcqRel)`. The asymmetric ordering makes the memory-safety argument harder to audit and can mislead future edits into thinking the atomics carry synchronization guarantees beyond the lock. Either use a plain `usize` under the lock, or keep `AtomicUsize` with consistent `Relaxed` on both sides and add an explicit lock-invariant comment explaining why `Relaxed` is sufficient.

runtime/src/utils/buffer/paged/cache.rs
```diff
@@ -467,19 +542,121 @@ impl Cache {
...
+            self.read_at_buf(offset, len, IoBufsMut::default()).await
+        }
+
+        async fn read_at_buf(
+            &self,
+            _offset: u64,
+            _len: usize,
+            _bufs: impl Into<IoBufsMut> + Send,
+        ) -> Result<IoBufsMut, Error> {
+            let sender = self.started.lock().take().unwrap();
```
clabby: `BlockingBlob`/`ControlledBlob` use `take().unwrap()` for one-shot invariants. An unexpected second call produces an uninformative `called Option::unwrap() on a None value` panic. Replace with `expect(...)` carrying invariant context, e.g.:

```rust
self.buf.take().expect("read_at_buf called more than once")
self.release_tx.take().expect("release receiver already consumed")
```

This makes cancellation/concurrency regressions significantly easier to diagnose.

runtime/src/utils/buffer/paged/cache.rs
```diff
@@ -23,6 +23,92 @@ use tracing::{debug, error, trace};
 // [Shared]. The IoBuf contains only the logical (validated) bytes of the page.
 type PageFetchFut = Shared<Pin<Box<dyn Future<Output = Result<IoBuf, Arc<Error>>> + Send>>>;
 
+// Shared handle to one in-flight fetch. The cache keeps one copy in `page_fetches`, and each
+// waiter clones the Arc while it is still interested in the result.
+type PageFetch = Arc<InFlightPageFetch>;
+
+struct InFlightPageFetch {
```
andresilva: nit, this name throws me off a bit. I'd prefer naming this `PageFetch`, removing the type alias above, and just doing `Arc<PageFetch>` below in `PageFetchGuard`.

runtime/src/utils/buffer/paged/cache.rs
```diff
@@ -23,6 +23,92 @@ use tracing::{debug, error, trace};
...
+    // Shared future that reads and validates the logical page exactly once.
+    future: PageFetchFut,
+    // Count of waiters that still need cancellation cleanup for this fetch.
+    waiters: AtomicUsize,
+}
+
+/// Fetch one logical page for insertion into the page cache, rejecting partial pages because cache
+/// entries must always contain a full logical page.
+async fn fetch_cacheable_page(
```
andresilva: nit, this function look a bit out of place, maybe move to end of file (before tests?)

runtime/src/utils/buffer/paged/cache.rs
andresilva: If we move the count to the map directly we can get rid of the `InFlightPageFetch` type entirely, we'd just store `(Arc<PageFetchFut>, usize)` in `page_fetches`. The synchronization is happening through the lock on the cache and this atomic is just there to give us interior mutability (which can be a bit misleading IMO). Or just have:

```rust
struct PageFetch {
    fetch: Arc<PageFetchFut>,
    waiters: usize,
}

page_fetches: HashMap<(u64, u64), PageFetch>,
```

patrick-ogrady:
Saw this nit after a few runs with the LLM...not sure if we should do something about it (seems fine not to):

```
  - [P1] Keep cancelled fetches pinned until backend I/O completes — /Users/patrickogrady/code/monorepo-pr-3311/runtime/src/utils/buffer/paged/
    cache.rs:106-107
    When the last waiter drops here, the cache forgets the in-flight fetch immediately, but some Blob::read_at implementations keep the physical
    read alive after the future is dropped. In particular, runtime/src/storage/iouring.rs submits the read before awaiting the oneshot result,
    and runtime/src/iouring/mod.rs::handle_cqe only frees that waiter slot when a CQE arrives. Under commonware-runtime/iouring-storage,
    cancelling a page read and then retrying the same page can therefore enqueue a second physical read while the first still consumes ring
    capacity; repeated cancellations on one hot page can exhaust the default 128-slot ring and stall unrelated I/O.
    
```

andresilva:
> Saw this nit after a few runs with the LLM...not sure if we should do something about it (seems fine not to):
> ...

This is true and I'm planning to fix this, we'll need to implement cancellation of on-going ops if the submitter cancels. I'd say it's a bigger issue that's orthogonal to the changes here.

clabby (approved):

PR #3312, opened 2026-03-05
[math] Implemnt lin_comb_eval method

cronokirby:
This is a more efficient method, useful for doing, e.g. batch evaluation checks.

I also saw a place in `eval_msm` where we can take advantage of the recently added `powers` function.

### DIFF
--- a/math/src/poly.rs
+++ b/math/src/poly.rs
@@ -725,8 +725,9 @@ pub mod fuzz {
 }
 #[cfg(test)]
 mod test {
-    use super::*;
+    use super::{fuzz::Plan, *};
     use crate::test::F;
+    use arbitrary::Unstructured;
 
     #[test]
     fn test_eq() {
@@ -745,6 +746,34 @@ mod test {
         assert!(eq(&[1, 2, 0, 0], &[1, 2]));
         assert!(!eq(&[1, 2, 0], &[2, 3]));
         assert!(!eq(&[2, 3], &[1, 2, 0]));
+    }
+
+    #[test]
+    fn lin_comb_eval_edge_cases() {
+        fn poly(coeffs: &[u8]) -> Poly<F> {
+            Poly {
+                coeffs: coeffs.iter().copied().map(F::from).try_collect().unwrap(),
+            }
+        }
+
+        fn pairs(values: &[(u8, u8)]) -> Vec<(F, F)> {
+            values
+                .iter()
+                .map(|(a, b)| (F::from(*a), F::from(*b)))
+                .collect()
+        }
+
+        let cases = [
+            Plan::LinCombEval(poly(&[3, 5, 7]), vec![]),
+            Plan::LinCombEval(poly(&[11]), pairs(&[(2, 0), (3, 1), (5, 8)])),
+            Plan::LinCombEval(poly(&[4, 6, 8]), pairs(&[(2, 5), (7, 5), (3, 5)])),
+            Plan::LinCombEval(poly(&[9, 2, 3, 4]), pairs(&[(6, 0), (1, 0), (5, 7)])),
+            Plan::LinCombEval(poly(&[1, 2, 4, 8]), pairs(&[(3, 1), (7, 1), (2, 6)])),
+        ];
+        let mut u = Unstructured::new(&[]);
+        for case in cases {
+            case.run(&mut u).unwrap();
+        }
     }
 
     #[cfg(feature = "arbitrary")]


cronokirby:
We should actually modify "powers" to take in a basis instead of using "1" to avoid n multiplications here.

math/src/algebra.rs
```diff
@@ -10,6 +10,7 @@ use core::{
     ops::{Add, AddAssign, Mul, MulAssign, Neg, Sub, SubAssign},
 };
 use rand_core::CryptoRngCore;
+use std::iter;
```
clabby: This breaks `no_std` builds.

It also looks like we don’t add the `math` crate to the `check_no_std.sh` script, can we add that in this PR?

math/src/poly.rs
```diff
@@ -131,14 +135,40 @@ impl<K> Poly<K> {
...
+
+    /// Compute `sum_i a_i * self(b_i)`.
+    ///
+    /// This is more efficient than several calls to `eval_msm`, but produces the
+    /// same result.
+    ///
+    /// This returns `0` if the iterator is empty.
+    pub fn lin_comb_eval<'a, R: Ring + 'a>(
+        &self,
+        into_iter: impl IntoIterator<Item = (R, Cow<'a, R>)>,
```
clabby, line 152: Maybe we should take `impl IntoIterator<Item = (R, &R)>` here? The function only needs `bi.as_ref()`, so this forces both callers that have owned and borrowed `b` values to map their iterators at the callsite of this function.

cronokirby: The issue is that if you have an iterator that yields `(R, R)` then you can't (?) use this call, because you'd need to take a reference to a temporary. In such cases, you could usually have a borrowing iterator, and clone the first argument, but that seems wasteful.

clabby: Ah that’s a good point. This is fine as is, then.

math/src/poly.rs
```diff
@@ -131,14 +135,40 @@ impl<K> Poly<K> {
         K: Space<R>,
     {
         // Contains 1, r, r^2, ...
+        let weights = powers(R::one(), r)
+            .take(self.len_usize())
+            .collect::<Vec<_>>();
+        K::msm(&self.coeffs, &weights, strategy)
+    }
+
+    /// Compute `sum_i a_i * self(b_i)`.
```
patrick-ogrady, line 144: LLM has some concerns:

```
• lin_comb_eval is generic over K: Space<R>, and Space<R> is a right R-module. That means coefficients/scalars must act on the right: point * scalar.

  The buggy code built weights as a, a*b, a*b^2, ..., so the MSM computed:
  sum_j c_j * (a b^j)

  But lin_comb_eval is supposed to match repeated calls to eval_msm(&b) * a, which is:
  sum_j c_j * (b^j a)

  Those are only equal if scalar multiplication in R commutes. Since Ring does not require commutativity, the implementation was generically wrong.

  The fix was to build weights as b^j * a instead of a * b^j.
```

It wants you to do:

```
diff --git a/math/src/poly.rs b/math/src/poly.rs
index 0fb20b2c9..a53dde719 100644
--- a/math/src/poly.rs
+++ b/math/src/poly.rs
@@ -141,7 +141,7 @@ impl<K> Poly<K> {
         K::msm(&self.coeffs, &weights, strategy)
     }
 
-    /// Compute `sum_i a_i * self(b_i)`.
+    /// Compute `sum_i self(b_i) * a_i`.
     ///
     /// This is more efficient than several calls to `eval_msm`, but produces the
     /// same result.
@@ -155,7 +155,7 @@ impl<K> Poly<K> {
     where
         K: Space<R>,
     {
-        // Contains a0 + a1 + ..., a0 b0 + a1 b1 + ..., a0 b0^2 + a1 b1^2 + ...
+        // Contains a0 + a1 + ..., b0 a0 + b1 a1 + ..., b0^2 a0 + b1^2 a1 + ...
         let weights = {
             let mut iter = into_iter.into_iter();
             let Some((a0, b0)) = iter.next() else {
@@ -163,10 +163,14 @@ impl<K> Poly<K> {
             };
 
             let len = self.len_usize();
-            let mut out: Vec<_> = powers(a0, b0.as_ref()).take(len).collect();
+            let mut out: Vec<_> = powers(R::one(), b0.as_ref())
+                .take(len)
+                .map(|pow| pow * &a0)
+                .collect();
             for (ai, bi) in iter {
-                powers(ai, bi.as_ref())
+                powers(R::one(), bi.as_ref())
                     .take(len)
+                    .map(|pow| pow * &ai)
                     .zip(out.iter_mut())
                     .for_each(|(c_j, o_j)| *o_j += &c_j);
             }
@@ -726,7 +730,96 @@ pub mod fuzz {
 #[cfg(test)]
 mod test {
     use super::*;
-    use crate::test::F;
+    use crate::{
+        algebra::{Additive, Multiplicative, Object, Ring},
+        test::F,
+    };
+    use commonware_parallel::Sequential;
+    use core::ops::{Add, AddAssign, Mul, MulAssign, Neg, Sub, SubAssign};
+
+    struct Matrix([[i32; 2]; 2]);
+
+    impl Object for Matrix {}
+
+    impl Additive for Matrix {
+        fn zero() -> Self {
+            Self([[0, 0], [0, 0]])
+        }
+    }
+
+    impl<'a> Add<&'a Self> for Matrix {
+        type Output = Self;
+
+        fn add(self, rhs: &'a Self) -> Self::Output {
+            Self([
+                [self.0[0][0] + rhs.0[0][0], self.0[0][1] + rhs.0[0][1]],
+                [self.0[1][0] + rhs.0[1][0], self.0[1][1] + rhs.0[1][1]],
+            ])
+        }
+    }
+
+    impl<'a> AddAssign<&'a Self> for Matrix {
+        fn add_assign(&mut self, rhs: &'a Self) {
+            *self = *self + rhs;
+        }
+    }
+
+    impl Neg for Matrix {
+        type Output = Self;
+
+        fn neg(self) -> Self::Output {
+            Self([
+                [-self.0[0][0], -self.0[0][1]],
+                [-self.0[1][0], -self.0[1][1]],
+            ])
+        }
+    }
+
+    impl<'a> Sub<&'a Self> for Matrix {
+        type Output = Self;
+
+        fn sub(self, rhs: &'a Self) -> Self::Output {
+            self + &-*rhs
+        }
+    }
+
+    impl<'a> SubAssign<&'a Self> for Matrix {
+        fn sub_assign(&mut self, rhs: &'a Self) {
+            *self = *self - rhs;
+        }
+    }
+
+    impl Multiplicative for Matrix {}
+
+    impl<'a> Mul<&'a Self> for Matrix {
+        type Output = Self;
+
+        fn mul(self, rhs: &'a Self) -> Self::Output {
+            Self([
+                [
+                    self.0[0][0] * rhs.0[0][0] + self.0[0][1] * rhs.0[1][0],
+                    self.0[0][0] * rhs.0[0][1] + self.0[0][1] * rhs.0[1][1],
+                ],
+                [
+                    self.0[1][0] * rhs.0[0][0] + self.0[1][1] * rhs.0[1][0],
+                    self.0[1][0] * rhs.0[0][1] + self.0[1][1] * rhs.0[1][1],
+                ],
+            ])
+        }
+    }
+
+    impl<'a> MulAssign<&'a Self> for Matrix {
+        fn mul_assign(&mut self, rhs: &'a Self) {
+            *self = *self * rhs;
+        }
+    }
+
+    impl Ring for Matrix {
+        fn one() -> Self {
+            Self([[1, 0], [0, 1]])
+        }
+    }
 
     #[test]
     fn test_eq() {
@@ -747,6 +840,19 @@ mod test {
         assert!(!eq(&[2, 3], &[1, 2, 0]));
     }
 
+    #[test]
+    fn test_lin_comb_eval_preserves_right_scalar_order() {
+        let poly = Poly::from_iter_unchecked([Matrix::zero(), Matrix::one()]);
+        let a = Matrix([[0, 1], [0, 0]]);
+        let b = Matrix([[0, 0], [1, 0]]);
+
+        let expected = poly.eval_msm(&b, &Sequential) * &a;
+        let actual = poly.lin_comb_eval([(a, Cow::Owned(b))], &Sequential);
+
+        assert_eq!(expected, Matrix([[0, 0], [0, 1]]));
+        assert_eq!(actual, expected);
+    }
+
     #[cfg(feature = "arbitrary")]
     mod conformance {
         use super::*;
```

Still pondering myself.

cronokirby: `Ring` inherits from `Multiplicative`, which assumes commutativity of multiplication (this is tested). However, this should be documented better.

math/src/poly.rs
```diff
@@ -596,6 +627,17 @@ pub mod fuzz {
                 Self::EvalMsm(f, x) => {
                     assert_eq!(f.eval(&x), f.eval_msm(&x, &Sequential));
                 }
+                Self::LinCombEval(f, pairs) => {
```
patrick-ogrady, line 630: `LinCombEval` currently relies on randomized fuzzing, but the risky failure modes here are coefficient-order and power-series alignment errors that are easy to miss unless you pin down the boundary cases. Please add a deterministic unit test that compares `lin_comb_eval` against the naive fold for empty input, degree-0 polynomials, repeated evaluation points, and `b = 0` / `b = 1`; those are the cases most likely to catch an off-by-one in the weights.

math/src/algebra.rs
```diff
@@ -59,13 +60,11 @@ fn monoid_exp<T: Clone>(
...
-pub fn powers<R: Ring>(base: &R, len: usize) -> impl Iterator<Item = R> + '_ {
-    (0..len).scan(R::one(), move |state, _| {
-        let out = state.clone();
-        *state *= base;
-        Some(out)
-    })
+/// Return `shift, shift * base, shift * base^2, ...`.
+///
+/// Use `powers(R::one(), base)` if you just want the standard powers.
+pub fn powers<R: Ring>(shift: R, base: &R) -> impl Iterator<Item = R> + '_ {
```
patrick-ogrady, line 66: Repurposing `powers` breaks the previous `pub fn powers(base: &R, len: usize)` contract in a `BETA` module without any compatibility path. In-tree call sites were updated in lockstep, but downstream callers using the old finite helper now fail to compile, which violates the repo policy for `BETA` APIs. Please keep the existing finite helper (it can be deprecated) and introduce the infinite/shifted iterator under a new name such as `powers_from` or `shifted_powers` instead of changing `powers` in place. If the infinite iterator remains public, its docs should also state explicitly that callers must bound it with `.take(...)`.

The reported regression was reproduced via `cargo test -p commonware-math --doc` with this compatibility probe:

```rust
/// ```
/// use commonware_math::algebra::powers;
/// use commonware_math::fields::goldilocks::F;
///
/// let base = F::from(7u64);
/// let got: Vec<_> = powers(&base, 4).collect();
/// assert_eq!(got, vec![F::from(1), F::from(7), F::from(49), F::from(343)]);
/// ```
```

This now fails to compile with `E0277` / `E0308`.

cronokirby: In general, can we change BETA level APIs in terms of the calling interface only?

cronokirby: With an offline discussion we came to the conclusion that this was fine.

math/src/poly.rs
cronokirby: I added a test to exercise these edge cases precisely.

PR #3314, opened 2026-03-05
[runtime/utils/buffer/{write,tip}] Nits

patrick-ogrady:
Related: #3307 
Related: #2731

### DIFF
--- a/runtime/src/utils/buffer/tip.rs
+++ b/runtime/src/utils/buffer/tip.rs
@@ -140,17 +140,16 @@ impl Buffer {
             return None;
         }
 
-        let len = self.len;
+        // Clear the logical length up front so the tip is empty even if the returned buffer
+        // still aliases the old backing.
+        let len = std::mem::take(&mut self.len);
         let offset = self.offset;
-        let data = std::mem::take(&mut self.data);
-        let buf = if len == data.len() {
-            data
-        } else {
-            data.slice(..len)
-        };
-        self.len = 0;
         self.offset += len as u64;
-        Some((buf, offset))
+
+        // Hand the buffered prefix to the caller without copying. If `data` retained extra
+        // capacity or trailing bytes, `split_to` leaves them behind in the discarded remainder.
+        let mut data = std::mem::take(&mut self.data);
+        Some((data.split_to(len), offset))
     }
 
     /// Returns a mutable tip buffer with capacity for at least `needed` bytes.


runtime/src/utils/buffer/tip.rs
```diff
@@ -120,15 +134,14 @@ impl Buffer {
         };
         assert!(start <= end, "slice start must be <= end");
         assert!(end <= self.len, "slice out of bounds");
-        self.data.slice(range)
+        self.data.slice(start..end)
```
patrick-ogrady, line 95: This should be fine as-is but this does make it clearer.

runtime/src/utils/buffer/paged/append.rs
patrick-ogrady, line 1:
```
• The patch introduces a data-loss race in Append::to_immutable() under concurrent state transitions, and it regresses memory usage when
  reopening blobs with partial trailing pages. Those issues are enough to make the change unsafe despite the existing tests passing.

  Full review comments:

  - [P1] Prevent to_immutable() from sealing newly buffered bytes — /Users/patrickogrady/code/monorepo-pr-3307/runtime/src/utils/buffer/paged/
    append.rs:230-231
    If another clone calls to_mutable() after flush_internal() drops the tip lock, it can append new bytes before this second lock is taken. Re-
    marking the buffer immutable here seals those new bytes without flushing them, and later sync() will skip them because immutable buffers
    return early. In that interleaving the data remains visible in memory but is lost on drop or restart.
  - [P2] Compact partial-page slices before storing them in the tip — /Users/patrickogrady/code/monorepo-pr-3307/runtime/src/utils/buffer/paged/
    append.rs:194-199
    When reopening a blob whose last page is partial, partial_data is a zero-copy slice of the entire physical page returned by
    get_page_from_blob(). Passing that slice straight into Buffer::from_data retains the full backing allocation instead of copying just the
    logical bytes as the old append(partial_page.as_ref()) path did. On common 4 KiB pages this turns a tiny tail into an 8 KiB live buffer, and
    near the 64 KiB limit it can pin an oversized unpooled allocation for every open section.
    
```

runtime/src/utils/buffer/tip.rs
```diff
@@ -161,16 +175,23 @@ impl Buffer {
...
         if self.is_empty() {
             return None;
         }
-        let buf = self.data.slice(..self.len);
-        self.len = 0;
+
+        let len = self.len;
         let offset = self.offset;
-        self.offset += buf.len() as u64;
+        let data = std::mem::take(&mut self.data);
```
andresilva: Not sure if this is intentional but this means we can't reuse the allocation later on in `writable`.

patrick-ogrady: This was intentional (often after we flush/sync here, we never write again). If we do, its just a quick grab from the buffer pool?

patrick-ogrady: For some of the cases where we do a 1MB write buffer, this can save a lot of memory on old blobs.

patrick-ogrady: I suppose we could instead be better about making things `immutable()` when we're done?

patrick-ogrady: This is difficult to detect in `segmented` journal (so, can be easier to just be dumb here)

runtime/src/utils/buffer/tip.rs
```diff
@@ -46,15 +46,36 @@ pub(super) struct Buffer {
 impl Buffer {
     /// Creates a new buffer with the provided `offset` and `capacity`.
     ///
-    /// The backing buffer is allocated eagerly and starts with zero length.
+    /// The buffer starts detached, mutable, and allocates backing on first write.
     pub(super) fn new(offset: u64, capacity: usize, pool: BufferPool) -> Self {
-        let data = pool.alloc(capacity).freeze();
```
patrick-ogrady, line 51: I wonder if we should "string together" buffers if our request is bigger than any single item can be? For example, the tip buffer here is likely larger than the default storage pool.

patrick-ogrady: I suppose anything that big should not be in the buffer pool to begin with (although the `take()` strategy here that drops useless bytes would allocate a lot more -> although would not "leak" it when it became inactive).

andresilva (approved):

runtime/src/utils/buffer/tip.rs
patrick-ogrady: We changed the default size to be 8MB

PR #3316, opened 2026-03-05
[bls12381] Allow building `Scalar` from `u64` and limbs

roynalnaruto:
- Add a `From<u64>` implementation
- Add a `from_limbs` public method

### DIFF
--- a/cryptography/src/bls12381/primitives/group.rs
+++ b/cryptography/src/bls12381/primitives/group.rs
@@ -564,7 +564,7 @@ impl Scalar {
         Self(fr)
     }
 
-    /// Creates a new scalar from the given limbs in little-endian represention.
+    /// Creates a new scalar from the given limbs in little-endian representation.
     ///
     /// The limbs represent an integer `l[0] + l[1]*2^64 + l[2]*2^128 + l[3]*2^192`, which is then
     /// converted into [`blst_fr`]'s internal Montgomery form.


cronokirby (dismissed):

cryptography/src/bls12381/primitives/group.rs
```diff
@@ -564,19 +564,25 @@ impl Scalar {
         Self(fr)
     }
 
-    /// Creates a new scalar from the provided integer.
-    pub(crate) fn from_u64(i: u64) -> Self {
-        // Create a new scalar
+    /// Creates a new scalar from the given limbs in little-endian represention.
```
clabby: Typo: "represention" should be "representation".

```suggestion
    /// Creates a new scalar from the given limbs in little-endian representation.

File: cryptography/src/bls12381/primitives/group.rs [hunk 2]
```diff
@@ -564,19 +564,25 @@ impl Scalar {
         Self(fr)
     }
 
-    /// Creates a new scalar from the provided integer.
-    pub(crate) fn from_u64(i: u64) -> Self {
-        // Create a new scalar
+    /// Creates a new scalar from the given limbs in little-endian representation.
```
patrick-ogrady, line 567: `in little-endian representation`, these are just u64's (there is no representation?)

If you are referring to the scalar itself, I think that should remain on the scalar struct definition?

roynalnaruto: I wanted to communicate that the first limb (`limbs[0]`) is the least significant limb, should I refactor it to:

```suggestion
    /// Creates a new scalar from the given limbs in little-endian limb order, i.e. `limbs[0]` is the least significant limb.
```

cronokirby: The internal order the Scalar stores thing is orthogonal to what this method accepts.

Once you take more than 1 u64 value, it matters what order they get passed in.

e.g. `Scalar(x + 2^64 y)` is not `Scalar(2^64 x + y)`, so specifying that this takes in the lowest order limb first is nice to have.

Perhaps "order" would be better than "representation".

patrick-ogrady: This is fine as-is 👍

Issue #3320, opened 2026-03-06
[p2p] Add Connection Age/Last Message Received Gauge per Validator


patrick-ogrady:

patrick-ogrady:
When disconnected, go to 0 (or remove)? Better to avoid dealing with absence of a metric?

PR #3324, opened 2026-03-06
[p2p/authenticated] Add Connection Age

patrick-ogrady:
Closes: #3320


p2p/src/authenticated/discovery/actors/spawner/actor.rs
```diff
@@ -129,12 +121,8 @@ impl<
                         connection,
                         reservation,
                     } => {
-                        // Mark peer as connected
-                        self.connections.inc();
```
patrick-ogrady, line 133: Managing this in tracker is much clearer imho (avoids all these one-off .dec() calls)

PR #3326, opened 2026-03-06
[marshal+resolver] `Urgent` request type

clabby:
## Overview

When a proposer is missing a parent block, triggers an "urgent" fetch for the parent block that fans out to every eligible peer. This closes a liveness hole where a leader may cycle through eligible peers in serial for a parent block due to a malicious proposer withholding the block, potentially causing views to time out.

- Resolver: `RequestType::Urgent` fans out to all eligible peers. `escalate()` upgrades an in-flight request without discarding the original response, using per-peer start times for accurate latency tracking.
- Marshal uses this to fetch missing blocks in its subscription handler.

### DIFF
--- a/resolver/src/p2p/engine.rs
+++ b/resolver/src/p2p/engine.rs
@@ -449,8 +449,8 @@ impl<
         // (blocking the peer also removes any targets associated with it)
         commonware_p2p::block!(self.blocker, peer.clone(), "invalid data received");
         self.fetcher.block(peer);
-        self.metrics.fetch.inc(Status::Failure);
         if exhausted {
+            self.metrics.fetch.inc(Status::Failure);
             self.fetcher.add_retry(key);
         }
     }


consensus/src/simplex/actors/voter/mod.rs
```diff
@@ -3965,6 +4054,7 @@ mod tests {
                         break;
                     }
                     MailboxMessage::Certificate(_) => continue,
+                    MailboxMessage::Urgent { .. } => continue,
```
clabby: We can merge these

resolver/src/p2p/fetcher.rs
```diff
@@ -369,31 +390,125 @@ where
...
+    pub fn is_urgent(&self, key: &Key) -> bool {
+        self.urgent.contains(key)
+    }
+
+    /// Re-prioritizes an urgent fetch immediately, preserving in-flight work.
+    ///
+    /// If the key is pending, it is moved to the front of the queue. If the
+    /// key is active, the existing request ID is kept alive and the same ID is
+    /// fanned out to any additional eligible peers that have not already seen it.
+    pub async fn escalate(
```
clabby, line 493: We should add a metric for escalated requests.

resolver/src/lib.rs
```diff
@@ -51,10 +76,32 @@ commonware_macros::stability_scope!(BETA {
         type PublicKey: PublicKey;
 
         /// Initiate a fetch request for a single key.
```
clabby: nit: Document that this sends a `RequestType::Regular` request.

resolver/src/lib.rs
```diff
@@ -51,10 +76,32 @@ commonware_macros::stability_scope!(BETA {
...
-        fn fetch(&mut self, key: Self::Key) -> impl Future<Output = ()> + Send;
+        fn fetch(&mut self, key: Self::Key) -> impl Future<Output = ()> + Send {
+            self.fetch_with_type(key, RequestType::Regular)
+        }
 
         /// Initiate a fetch request for a batch of keys.
```
clabby: nit: Document that this sends a `RequestType::Regular` request.

resolver/src/p2p/fetcher.rs
```diff
@@ -369,31 +390,125 @@ where
...
+                }
+                Err(err) => {
+                    self.requests_sent.inc(Status::Failure);
+                    debug!(?err, ?peer, "send failed");
+                    self.update_performance(&peer, self.timeout);
+                }
+            }
+        }
+
+        if sent_now.is_empty() {
```
clabby, line 571: Agent Review

---

When `escalate()` runs on an active request and every additional eligible peer is temporarily rate-limited, the loop silently skips each peer (`Err(_) => continue` around line 449) and exits early at the `sent_now.is_empty()` check. The request then sits waiting for its full active timeout with no retry or rescheduling—urgent fanout is effectively lost until the normal timeout fires.

Fix: track the earliest `not_until` timestamp returned by failed `check()` calls and schedule a retry escalation at that time (or requeue the key into pending with a `not_until` wakeup) so urgency isn't silently discarded.

```rust
#[test]
fn test_escalate_retries_after_rate_limit_expires_for_active_request() {
    let runner = Runner::default();
        // ... setup fetcher with peer1 already in-flight, peer2/peer3 rate-limited ...
        fetcher.escalate(&key, &mut sender).await;
        // Immediately after escalation, only the original peer should be contacted
        assert_eq!(fetcher.requests.get(&id).unwrap().peers.len(), 1);

        // After rate limit expires, escalation should have been retried
        context.sleep(Duration::from_secs(1)).await;
        assert!(
            fetcher.requests.get(&id).unwrap().peers.len() > 1,
            "expected escalation retry after rate-limit expiry"
        );
    });
}
```

resolver/src/p2p/engine.rs
```diff
@@ -429,8 +441,7 @@ impl<
             self.metrics.fetch.inc(Status::Success);
             self.fetch_timers.remove(&key).unwrap(); // must exist in the map, records metric on drop
 
-            // Clear all targets for this key
-            self.fetcher.clear_targets(&key);
+            self.fetcher.finish(&key);
```
clabby: Agent Review

---

`finish(key)` removes the request immediately on the first accepted response. Any in-flight responses from other already-contacted peers are silently dropped via `pop_by_id` returning `None`, so those peers never receive a latency update even though `requests_sent` counted their sends. In sustained urgent mode this biases peer EMA/performance data toward "winner-only" samples.

Either (1) add a comment explicitly documenting winner-only scoring as intentional, or (2) update remaining in-flight peers on `finish()` (e.g. with elapsed-to-finish or a bounded penalty) so `requests_sent` and performance tracking remain aligned.

consensus/src/simplex/actors/resolver/state.rs
```diff
@@ -572,6 +637,94 @@ mod tests {
...
+                Certificate::Notarization(notarization_v5),
+                Some(View::new(2)),
+                &mut resolver,
+            )
+            .await;
+
+        // Clear and fail certification — view 2 should be re-requested as urgent
+        // because the resolver clears urgency on delivery, but the simplex state
+        // remembers which views were originally urgent.
+        resolver.clear().await;
```
clabby: Agent Review

---

`urgent_re_request_stays_urgent_after_certification_failure` calls `resolver.clear()` to reset urgency, but production code resets urgency per-key on completion or cancellation, not globally. This means the test could pass while a key-scoped bug goes undetected (e.g. urgency leaking between keys, or the wrong key being cleared).

Add a variant that completes/cancels only the satisfied key and verifies `State::urgent_views` still forces an urgent re-fetch after certification failure, matching the production reset path.

resolver/src/p2p/fetcher.rs
```diff
@@ -1866,4 +2100,198 @@ mod tests {
             );
         });
     }
+
+    #[test]
+    fn test_successful_completion_does_not_update_remaining_in_flight_peer_performance() {
```
clabby, line 2230: Important to note - we don’t score peers who respond after the first in `Urgent` requests, since that’d require some more complex machinery to keep track of the key until all peers respond or time out.

clabby: While we could do this, I don’t think it’s entirely necessary. Measuring the latency of the first peer that gets back to us should be sufficient, others’ responses being discarded doesn’t penalize them (but also doesn’t contribute to their score if they do get back to us).

consensus/src/simplex/actors/resolver/ingress.rs
```diff
@@ -39,6 +41,19 @@ impl<S: Scheme, D: Digest> Mailbox<S, D> {
...
     }
+
+    /// Request a specific view urgently, fanning out to every peer.
+    ///
+    /// Used when the proposer is missing a parent certificate and needs it
+    /// before the proposal deadline. The resolver remembers this view as
+    /// urgent so that if a higher-view notarization temporarily satisfies
+    /// the fetch but later fails certification (causing the original view
+    /// to be re-fetched), the re-fetch stays urgent.
+    pub async fn urgent(&mut self, view: View) {
```
patrick-ogrady: What I had in mind here was actually fetching block parents (in marshal) rather than certificates...I need to think about if we need this too (I was more concerned about the case the previous leader withheld data than the all-to-all broadcast not getting to us).

In any case, I think the changes to resolver would be required in any case.

clabby: Reverted the simplex changes and now just use the urgent request type for marshal’s block subscription fallback.

Issue #1811, opened 2025-10-06
[cryptography/bls12381] Select Best 2f+1 Commitments By Reveals


patrick-ogrady:
Instead of selecting the best 2f+1 commitments by dealer index, we should optimize for commitments with minimal share disclosure: https://github.com/commonwarexyz/monorepo/blob/636eef4ff0753513286774569f347452c36d21cb/cryptography/src/bls12381/dkg/arbiter.rs#L238-L244

patrick-ogrady:
We should probably just select all valid commitments? "Best" here in the worst-case (with f Byzantine contributors) doesn't really help us. That was the rational for taking this simplistic approach to begin with...

cronokirby:
Selecting all commitments that are good enough seems sound to me ; at that point the view on which commitments are or aren't good enough is finalized as well.

patrick-ogrady:
Following #3328, it is now optimal to select the "first" 2f+1 (by replay in the log). This will enable VERY effective batch verify (whereas the current scheme allows `f` malicious dealers to insert their logs very close to the end of the epoch).

PR #3328, opened 2026-03-06
DKG Finalization Optimizations

### LINKED ISSUE #2536: [cryptography/dkg] Add `finalize_trusted`
Some applications may wish to verify dealings included in blocks to minimize delay at the end of an application epoch.

We should explore a way to provide pre-verified dealer logs to finalize to avoid duplicate verification work.

### LINKED ISSUE #3212: [cryptography] DKG: Low hanging optimization fruit
- We can use batch ed25519 verification in select,
- We can batch other share checks.

cronokirby:
Stacked on top of #3312.

Closes #2536.
Closes #3212.
Related https://github.com/commonwarexyz/monorepo/issues/1811.

This adds a new structure for containing dealer logs, allowing them to be verified in advance.

We also implement optimizations for a given dealer, batching signature checks, and reveal checks.

One big win is implementing parallelism to check multiple dealers at once (while preserving the order we select dealers in), which allows the method to actually scale, whereas it was bottlenecked previously.

### DIFF
--- a/cryptography/src/bls12381/dkg.rs
+++ b/cryptography/src/bls12381/dkg.rs
@@ -1332,6 +1332,7 @@ impl<V: Variant, P: PublicKey, M: Faults> Logs<V, P, M> {
                 ((*dealer).clone(), log, seed)
             })
             .collect();
+
         // This uses signature batch verification only for a particular dealer's
         // signatures. We could batch across all dealers, but in practice this starts
         // performing worse than just using parallelism with at least 4 threads,
@@ -1390,6 +1391,7 @@ impl<V: Variant, P: PublicKey, M: Faults> Logs<V, P, M> {
             }
         }
 
+        // Verify the batch and update the known valid dealers.
         let pending_results =
             Self::check_dealers::<B>(rng, &self.info, strategy, &transcript, &pending);
         let mut all_pending_valid = true;
@@ -1397,7 +1399,6 @@ impl<V: Variant, P: PublicKey, M: Faults> Logs<V, P, M> {
             self.known.insert(dealer, is_valid);
             all_pending_valid &= is_valid;
         }
-
         if all_pending_valid {
             return;
         }


cryptography/src/bls12381/dkg.rs
```diff
@@ -2429,7 +2586,13 @@ mod test_plan {
...
 
                 // Make sure that bad dealers are not selected.
-                let selection = select::<_, _, N3f1>(&info, dealer_logs.clone());
+                let mut logs = Logs::<_, _, N3f1>::default();
+                for (dealer, log) in &dealer_logs {
+                    logs.record(dealer.clone(), log.clone());
+                }
+                let selection = logs
+                    .clone()
+                    .select::<ed25519::Batch>(&info, &mut rng, &Sequential);
```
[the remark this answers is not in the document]
cronokirby: It's still deterministic though, even if "fragile" which is fine, it's random.

patrick-ogrady:
Not sure if this is an issue yet (but seems like we should have more testing around this edge case):

```
  - [P1] Preserve dealer-order selection after incremental pre_verify — /Users/patrickogrady/code/monorepo-pr-3328/cryptography/src/bls12381/
    dkg.rs:1383-1385
    pre_verify() only checks max(needed, unknown_in_required_prefix) unknown dealers, so if some low-sorted dealers were already cached as
    invalid and enough later dealers are cached as valid, a newly recorded valid dealer can remain None in known even though its key sorts before
    one of those later dealers. select() then skips that dealer and takes the later one, which makes the chosen subset depend on arrival/pre-
    verification history instead of dealer order. In the initial DKG this changes the summed public polynomial, and in resharing it still changes
    Output.dealers, so honest participants can derive different outputs from the same final log set.
```

cronokirby:
> Not sure if this is an issue yet (but seems like we should have more testing around this edge case):
> 
> ```
>   - [P1] Preserve dealer-order selection after incremental pre_verify — /Users/patrickogrady/code/monorepo-pr-3328/cryptography/src/bls12381/
>     dkg.rs:1383-1385
>     pre_verify() only checks max(needed, unknown_in_required_prefix) unknown dealers, so if some low-sorted dealers were already cached as
>     invalid and enough later dealers are cached as valid, a newly recorded valid dealer can remain None in known even though its key sorts before
>     one of those later dealers. select() then skips that dealer and takes the later one, which makes the chosen subset depend on arrival/pre-
>     verification history instead of dealer order. In the initial DKG this changes the summed public polynomial, and in resharing it still changes
>     Output.dealers, so honest participants can derive different outputs from the same final log set.
> ```

Yeah, this was an issue, pushed a fix and a regression test.

examples/reshare/src/dkg/actor.rs
patrick-ogrady, line 476: Now that `summary` is in `Logs`, wondering if we should also keep it inside this data structure as well (so it just becomes `log.check()`)?

Or, maybe that move into `Logs` was a bit overzealous (just got a little annoyed by how many places we passed around summary).

cronokirby: Moving Info into `Logs` seems like a fine move to me. I'm not sure I understand what data structure this comment refers to though.

patrick-ogrady: This comment doesn't make sense. Disregard.

examples/reshare/src/dkg/actor.rs
```diff
@@ -514,10 +516,17 @@ where
...
+                        let mut logs = Logs::<_, _, N3f1>::new(round.clone());
+                        for (dealer, log) in storage.logs(epoch) {
+                            logs.record(dealer, log);
+                        }
                         let (success, next_round, next_output, next_share) =
                             if let Some(ps) = player_state.take() {
-                                match ps.finalize::<N3f1>(logs, &Sequential) {
+                                match ps.finalize::<N3f1, Batch>(
+                                    logs,
+                                    &mut self.context,
```
patrick-ogrady, line 528: nit: move context to first arg

cryptography/src/bls12381/dkg.rs
```diff
@@ -1226,6 +1293,155 @@ fn transcript_for_log<V: Variant, P: PublicKey>(
...
+    pub fn pre_verify<B: BatchVerifier<PublicKey = P>>(
+        &mut self,
+        rng: &mut impl CryptoRngCore,
+        strategy: &impl Strategy,
+    ) {
+        let required_commitments = self.info.required_commitments::<M>() as usize;
+        let transcript = transcript_for_round(&self.info);
+
+        // Create a pending batch, which we try and optimistically size as small as
+        // possible, to avoid verifying more dealers than we need, if they're all
```
patrick-ogrady, line 1375: Someday, we'll switch to the "first 2f+1" 🙃

cryptography/src/bls12381/dkg.rs
```diff
@@ -1226,6 +1293,155 @@ fn transcript_for_log<V: Variant, P: PublicKey>(
...
+
+    /// Given the logs we've received, determine which dealer logs to use, if any.
+    ///
+    /// This might return an error if there are not enough good logs that we can use.
+    fn select<B: BatchVerifier<PublicKey = P>>(
+        mut self,
+        rng: &mut impl CryptoRngCore,
+        strategy: &impl Strategy,
+    ) -> Result<SelectedLogs<V, P>, Error> {
+        self.pre_verify::<B>(rng, strategy);
```
patrick-ogrady, line 1430: ❤️

patrick-ogrady (approved):

PR #3331, opened 2026-03-07
[deployer] remove redundant region clone in authorize loop

VolodymyrBg:
## Why was this change needed?
In `aws authorize`, we iterated over owned `String` regions but still called `Region::new(region.clone())`.
That added an unnecessary string clone on each loop iteration.

## What changed?
- Updated deployer/src/aws/authorize.rs to pass ownership directly: `Region::new(region)`.
- Kept region logging by moving the log before client creation (`"creating EC2 client"`).


clabby:

PR #3332, opened 2026-03-07
[storage/journal/authenticated] Ensure Stale Batch Has No Side Effects

patrick-ogrady:
AFAICT this is only a bug when using `journal::authenticated` directly.

### DIFF
--- a/storage/src/qmdb/current/mod.rs
+++ b/storage/src/qmdb/current/mod.rs
@@ -1061,7 +1061,8 @@ pub mod tests {
     {
         let executor = deterministic::Runner::default();
         executor.start(|context| async move {
-            let mut db: C = open_db(context.with_label("db"), "stale-side-effect-free".into()).await;
+            let mut db: C =
+                open_db(context.with_label("db"), "stale-side-effect-free".into()).await;
 
             let key1 = <C::Key as TestKey>::from_seed(1);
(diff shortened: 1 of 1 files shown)


storage/src/journal/authenticated.rs
patrick-ogrady, line 352: We returned an error after applying to journal 👀

storage/src/journal/authenticated.rs
```diff
@@ -335,6 +335,15 @@ where
     /// same parent for speculative execution, but only one may be applied. Applying
     /// a stale changeset returns an error.
     pub async fn apply_batch(&mut self, batch: Changeset<H::Digest, C::Item>) -> Result<(), Error> {
+        let actual = self.mmr.size();
+        if batch.changeset.base_size != actual {
```
patrick-ogrady, line 339: Should we apply to mmr first to fix instead?

patrick-ogrady:
bugbot run

storage/src/journal/authenticated.rs
patrick-ogrady: Took a stab at de-duping this but it ended up looking cleaner as-is.

roberto-bayardo:
We had noticed this but thought the inner call error was sufficient.  Too used to allowing IO errors leave things inconsistent I guess!

PR #3338, opened 2026-03-08
[consensus/coding] Prevent Proposer Withholding

clabby:
## Overview

Adjusts the `commonware-coding` interface to only have a single shard type that carries all data necessary to verify it. This prevents proposer withholding in `commonware-consensus, where a proposer could withhold a strong shard from the next leader and force the next leader to fetch the full block from peers.

Old Flow (Vulnerable to Withholding)

```mermaid
  sequenceDiagram
      participant L as Malicious Leader
      participant A as Participant A<br/>(next leader)
      participant B as Participant B
      participant C as Participant C

      Note over L: Knows A is next leader

      L--xA: StrongShard(A) — WITHHELD
      L->>B: StrongShard(B)
      L->>C: StrongShard(C)

      Note over A: No strong shard received<br/>→ no CheckingData
      Note over B: weaken() → CheckingData + WeakShard
      Note over C: weaken() → CheckingData + WeakShard

      B-->>A: WeakShard(B)
      B-->>C: WeakShard(B)
      C-->>A: WeakShard(C)
      C-->>B: WeakShard(C)

      Note over A: Has weak shards but<br/>NO CheckingData to validate them
      Note over A: ❌ Cannot reconstruct block

      Note over B,C: ✓ Block notarized<br/>(enough honest participants)
```

  New Flow (Withholding Ineffective)

```mermaid
  sequenceDiagram
      participant L as Malicious Leader
      participant A as Participant A<br/>(next leader)
      participant B as Participant B
      participant C as Participant C

      Note over L: Knows A is next leader

      L--xA: Shard(A) — WITHHELD
      L->>B: Shard(B)
      L->>C: Shard(C)

      Note over B: check(commitment, shard) ✓
      Note over C: check(commitment, shard) ✓

      B-->>A: Shard(B)
      B-->>C: Shard(B)
      C-->>A: Shard(C)
      C-->>B: Shard(C)

      Note over A: check(commitment, B's shard) ✓
      Note over A: check(commitment, C's shard) ✓
      Note over A: ✓ Reconstructs from peer shards

      Note over B,C: ✓ Block notarized
```

## Bandwidth Usage Notes

The original interface made the separation between "strong" and "weak" shards to optimize bandwidth usage for schemes where certain pieces of data required to verify shards was redundantly duplicated. Strong shards held all of the information and were sent from the leader, while participants gossiped weak shards. This change does not hurt reed solomon, but directly hurts ZODA since all shards must transmit the checksum matrix. This can dwarf the shard size at higher participant counts (e.g. for a 4MiB block):

Validators | Shard data | Checksum overhead | Ratio
-- | -- | -- | --
6 (n=4) | ~1 MiB | ~21 KiB | 2%
15 (n=10) | ~411 KiB | ~43 KiB | 10%
100 (n=67) | ~61 KiB | ~357 KiB | 583%
150 (n=100) | ~41 KiB | ~342 KiB | 833%

For now, this change acknowledges that this is acceptable. We will follow up with an optimization with https://github.com/commonwarexyz/monorepo/issues/3342.

### DIFF
--- a/consensus/src/marshal/coding/shards/engine.rs
+++ b/consensus/src/marshal/coding/shards/engine.rs
@@ -319,7 +319,7 @@ where
     /// Wrapped in [`Arc`] to enable cheap cloning when serving multiple subscribers.
     reconstructed_blocks: BTreeMap<Commitment, Arc<CodedBlock<B, C, H>>>,
 
-    /// Open subscriptions for assigned shard readiness for the keyed
+    /// Open subscriptions for assigned shard verification for the keyed
     /// [`Commitment`].
     ///
     /// For participants, readiness is satisfied once the leader-delivered
@@ -329,7 +329,7 @@ where
     ///
     /// Proposers are a special case: they satisfy readiness once their local
     /// proposal is cached because they already hold all shards.
-    assigned_shard_ready_subscriptions: BTreeMap<Commitment, Vec<oneshot::Sender<()>>>,
+    assigned_shard_verified_subscriptions: BTreeMap<Commitment, Vec<oneshot::Sender<()>>>,
 
     /// Open subscriptions for the reconstruction of a [`CodedBlock`] with
     /// the keyed [`Commitment`].
@@ -374,7 +374,7 @@ where
                 tracked_peers: Set::default(),
                 background_channel_capacity: config.background_channel_capacity,
                 reconstructed_blocks: BTreeMap::new(),
-                assigned_shard_ready_subscriptions: BTreeMap::new(),
+                assigned_shard_verified_subscriptions: BTreeMap::new(),
                 block_subscriptions: BTreeMap::new(),
                 metrics,
             },
@@ -429,7 +429,7 @@ where
                     subscribers.retain(|tx| !tx.is_closed());
                     !subscribers.is_empty()
                 });
-                self.assigned_shard_ready_subscriptions.retain(|_, subscribers| {
+                self.assigned_shard_verified_subscriptions.retain(|_, subscribers| {
                     subscribers.retain(|tx| !tx.is_closed());
                     !subscribers.is_empty()
                 });
@@ -475,11 +475,11 @@ where
                         .cloned();
                     response.send_lossy(block);
                 }
-                Message::SubscribeAssignedShardReady {
+                Message::SubscribeAssignedShardVerified {
                     commitment,
                     response,
                 } => {
-                    self.handle_assigned_shard_ready_subscription(commitment, response);
+                    self.handle_assigned_shard_verified_subscription(commitment, response);
                 }
                 Message::SubscribeByCommitment {
                     commitment,
@@ -508,7 +508,6 @@ where
                     .inc();
 
                 let commitment = shard.commitment();
-
                 if !self.should_handle_network_shard(commitment) {
                     continue;
                 }
@@ -548,7 +547,7 @@ where
             return self
                 .state
                 .get(&commitment)
-                .is_some_and(|s| !s.is_assigned_shard_ready());
+                .is_some_and(|s| !s.is_assigned_shard_verified());
         }
         true
     }
@@ -802,7 +801,7 @@ where
 
         // Local proposals bypass reconstruction, so shard subscribers waiting
         // for "our valid shard arrived" still need a notification.
-        self.notify_assigned_shard_ready_subscribers(commitment);
+        self.notify_assigned_shard_verified_subscribers(commitment);
 
         debug!(?commitment, "broadcasted shards");
     }
@@ -833,12 +832,12 @@ where
     ) {
         if let Some(state) = self.state.get_mut(&commitment) {
             match state.take_pending_action() {
-                Some(AssignedShardReadyAction::Broadcast(shard)) => {
+                Some(AssignedShardVerifiedAction::Broadcast(shard)) => {
                     self.broadcast_shard(sender, shard).await;
-                    self.notify_assigned_shard_ready_subscribers(commitment);
+                    self.notify_assigned_shard_verified_subscribers(commitment);
                 }
-                Some(AssignedShardReadyAction::NotifyOnly) => {
-                    self.notify_assigned_shard_ready_subscribers(commitment);
+                Some(AssignedShardVerifiedAction::NotifyOnly) => {
+                    self.notify_assigned_shard_verified_subscribers(commitment);
                 }
                 None => {}
             }
@@ -870,11 +869,11 @@ where
         }
     }
 
-    /// Handles the registry of an assigned shard readiness subscription.
+    /// Handles the registry of an assigned shard verification subscription.
     ///
     /// For participants this is tied to verification of the leader-delivered
     /// shard for the local index, not to generic block reconstruction.
-    fn handle_assigned_shard_ready_subscription(
+    fn handle_assigned_shard_verified_subscription(
         &mut self,
         commitment: Commitment,
         response: oneshot::Sender<()>,
@@ -883,7 +882,7 @@ where
         let has_shard = self
             .state
             .get(&commitment)
-            .is_some_and(|state| state.is_assigned_shard_ready());
+            .is_some_and(|state| state.is_assigned_shard_verified());
         if has_shard {
             response.send_lossy(());
             return;
@@ -899,7 +898,7 @@ where
             return;
         }
 
-        self.assigned_shard_ready_subscriptions
+        self.assigned_shard_verified_subscriptions
             .entry(commitment)
             .or_default()
             .push(response);
@@ -934,9 +933,12 @@ where
     }
 
     /// Notifies and cleans up any subscriptions waiting for assigned shard
-    /// readiness.
-    fn notify_assigned_shard_ready_subscribers(&mut self, commitment: Commitment) {
-        if let Some(mut subscribers) = self.assigned_shard_ready_subscriptions.remove(&commitment) {
+    /// verification.
+    fn notify_assigned_shard_verified_subscribers(&mut self, commitment: Commitment) {
+        if let Some(mut subscribers) = self
+            .assigned_shard_verified_subscriptions
+            .remove(&commitment)
+        {
             for subscriber in subscribers.drain(..) {
                 subscriber.send_lossy(());
             }
@@ -974,7 +976,8 @@ where
     /// Removing these entries drops all senders, causing receivers to resolve
     /// with cancellation (`RecvError`) instead of hanging indefinitely.
     fn drop_subscriptions(&mut self, commitment: Commitment) {
-        self.assigned_shard_ready_subscriptions.remove(&commitment);
+        self.assigned_shard_verified_subscriptions
+            .remove(&commitment);
         self.block_subscriptions
             .remove(&BlockSubscriptionKey::Commitment(commitment));
         self.block_subscriptions
@@ -1036,11 +1039,11 @@ where
     Ready(ReadyState<P, C, H>),
 }
 
-/// Action to take once assigned shard readiness has been established.
+/// Action to take once assigned shard verification has been established.
 ///
 /// Participants broadcast the shard to all peers, while non-participants
 /// only notify local subscribers.
-enum AssignedShardReadyAction<C: CodingScheme, H: Hasher> {
+enum AssignedShardVerifiedAction<C: CodingScheme, H: Hasher> {
     /// Broadcast the shard to all peers and notify local subscribers.
     Broadcast(Shard<C, H>),
     /// Only notify local subscribers (non-participant validated the leader's shard).
@@ -1063,7 +1066,7 @@ where
     /// The leader associated with this reconstruction state.
     leader: P,
     /// Our validated shard and the action to take with it.
-    pending_action: Option<AssignedShardReadyAction<C, H>>,
+    pending_action: Option<AssignedShardVerifiedAction<C, H>>,
     /// Shards that have been verified and are ready to contribute to reconstruction.
     checked_shards: Vec<C::CheckedShard>,
     /// Bitmap tracking which participant indices have contributed a shard.
@@ -1074,7 +1077,7 @@ where
     /// Keyed by shard index.
     received_shards: BTreeMap<u16, C::Shard>,
     /// Whether the leader's shard for our assigned index has been verified.
-    assigned_shard_ready: bool,
+    assigned_shard_verified: bool,
 }
 
 /// Phase data for `ReconstructionState::AwaitingQuorum`.
@@ -1121,7 +1124,7 @@ where
             contributed: BitMap::zeroes(participants_len),
             round,
             received_shards: BTreeMap::new(),
-            assigned_shard_ready: false,
+            assigned_shard_verified: false,
         }
     }
 }
@@ -1162,11 +1165,15 @@ where
 
         self.contributed.set(u64::from(shard.index), true);
         self.checked_shards.push(checked);
-        self.assigned_shard_ready = true;
+        self.assigned_shard_verified = true;
         self.pending_action = Some(if is_participant {
-            AssignedShardReadyAction::Broadcast(Shard::new(commitment, shard.index, data.clone()))
+            AssignedShardVerifiedAction::Broadcast(Shard::new(
+                commitment,
+                shard.index,
+                data.clone(),
+            ))
         } else {
-            AssignedShardReadyAction::NotifyOnly
+            AssignedShardVerifiedAction::NotifyOnly
         });
         true
     }
@@ -1295,8 +1302,8 @@ where
     }
 
     /// Returns whether the leader's shard for our index has been verified.
-    const fn is_assigned_shard_ready(&self) -> bool {
-        self.common().assigned_shard_ready
+    const fn is_assigned_shard_verified(&self) -> bool {
+        self.common().assigned_shard_verified
     }
 
     /// Return the proposal round associated with this state.
@@ -1312,7 +1319,7 @@ where
     /// Takes the pending action for this commitment's validated shard.
     ///
     /// Returns [`None`] if the leader's shard hasn't been validated yet.
-    const fn take_pending_action(&mut self) -> Option<AssignedShardReadyAction<C, H>> {
+    const fn take_pending_action(&mut self) -> Option<AssignedShardVerifiedAction<C, H>> {
         self.common_mut().pending_action.take()
     }
 
@@ -1422,7 +1429,7 @@ where
         // Leader's shard for our index is always verified eagerly,
         // even after transitioning to Ready. This ensures we broadcast
         // our own shard to help slower peers reach quorum.
-        if is_from_leader && !self.common().assigned_shard_ready {
+        if is_from_leader && !self.common().assigned_shard_verified {
             let progressed = self
                 .common_mut()
                 .verify_assigned_shard(
@@ -1872,7 +1879,7 @@ mod tests {
 
                 for peer in peers.iter_mut() {
                     peer.mailbox
-                        .subscribe_assigned_shard_ready(commitment)
+                        .subscribe_assigned_shard_verified(commitment)
                         .await
                         .await
                         .expect("shard subscription should complete");
@@ -1923,7 +1930,7 @@ mod tests {
 
                 for peer in peers.iter_mut() {
                     peer.mailbox
-                        .subscribe_assigned_shard_ready(commitment)
+                        .subscribe_assigned_shard_verified(commitment)
                         .await
                         .await
                         .expect("shard subscription should complete");
@@ -1976,7 +1983,7 @@ mod tests {
 
                 for peer in peers.iter_mut() {
                     peer.mailbox
-                        .subscribe_assigned_shard_ready(commitment)
+                        .subscribe_assigned_shard_verified(commitment)
                         .await
                         .await
                         .expect("shard subscription should complete");
@@ -2010,7 +2017,7 @@ mod tests {
             let round = Round::new(Epoch::zero(), View::new(1));
 
             // Subscribe on the proposer before it caches the locally proposed block.
-            let shard_sub = peers[0].mailbox.subscribe_assigned_shard_ready(commitment).await;
+            let shard_sub = peers[0].mailbox.subscribe_assigned_shard_verified(commitment).await;
             let commitment_sub = peers[0].mailbox.subscribe(commitment).await;
             let digest_sub = peers[0].mailbox.subscribe_by_digest(digest).await;
 
@@ -2080,7 +2087,7 @@ mod tests {
                     .await;
                 let mut shard_sub = peers[2]
                     .mailbox
-                    .subscribe_assigned_shard_ready(commitment)
+                    .subscribe_assigned_shard_verified(commitment)
                     .await;
 
                 // Byzantine peer sends the invalid shard.
@@ -2405,7 +2412,7 @@ mod tests {
                 // Subscribe before shards arrive so we can verify acceptance.
                 let shard_sub = peers[2]
                     .mailbox
-                    .subscribe_assigned_shard_ready(commitment)
+                    .subscribe_assigned_shard_verified(commitment)
                     .await;
 
                 // First leader update should stick.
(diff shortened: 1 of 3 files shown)


consensus/src/marshal/coding/types.rs
```diff
@@ -667,34 +533,29 @@ mod test {
...
+            Commitment::from((Sha256Digest::EMPTY, commitment, Sha256Digest::EMPTY, CONFIG));
+        let shard = RShard::new(commitment, 0, raw_shard);
+        let encoded = shard.encode();
+        let decoded = RShard::decode_cfg(&mut encoded.as_ref(), &MAX_SHARD_SIZE).unwrap();
+        assert!(shard == decoded);
     }
 
     #[test]
-    fn test_distribution_shard_decode_truncated_returns_error() {
+    fn test_shard_decode_truncated_returns_error() {
```
patrick-ogrady, line 555: I’m not sure where the right place to put this comment is but we should ensure we have a test for a witholding leader (and no backfill is required/used)?

consensus/src/marshal/coding/shards/engine.rs
patrick-ogrady, line 1: This can also be called by certificates (not just a leader broadcast).

clabby: Unsure what this is referring to

coding/src/lib.rs
```diff
@@ -188,53 +187,46 @@ commonware_macros::stability_scope!(ALPHA {
...
+        ///
+        /// Implementations must reject shards that were checked against a different
+        /// commitment than the one passed to `decode`. Mixing checked shards from
+        /// separate `encode` calls (and thus different commitments) must return an
+        /// error.
         fn decode(
             config: &Config,
             commitment: &Self::Commitment,
-            checking_data: Self::CheckingData,
+            checking_data: &Self::CheckingData,
```
cronokirby, line 335: There's no reason to not allow taking ownership of this, right?

coding/src/lib.rs
```diff
@@ -188,53 +187,46 @@ commonware_macros::stability_scope!(ALPHA {
...
+        /// returned checking data can be passed into subsequent calls.
         fn check(
             config: &Config,
             commitment: &Self::Commitment,
-            checking_data: &Self::CheckingData,
             index: u16,
-            weak_shard: Self::WeakShard,
-        ) -> Result<Self::CheckedShard, Self::Error>;
+            shard: &Self::Shard,
+            checking_data: Option<Self::CheckingData>,
```
cronokirby: why not `Option<&Self::CheckingData>`?

coding/src/lib.rs
```diff
@@ -143,40 +144,38 @@ commonware_macros::stability_scope!(ALPHA {
...
-        /// be considered as progress towards meeting the minimum number of shards.
+        /// This allows excluding invalid shards from the function signature of [Self::decode].
         type CheckedShard: Clone + Send + Sync;
-        /// The type of errors that can occur during encoding, weakening, checking, and decoding.
+        /// Precomputed state derived from a commitment that can be reused across
+        /// multiple calls to [Self::check] and [Self::decode].
+        ///
+        /// For schemes where checking is expensive, passing previously
+        /// computed checking data avoids redundant work.
+        type CheckingData: Clone + Send + Sync;
```
patrick-ogrady: CheckingData before CheckedShard?

coding/src/reed_solomon.rs
```diff
@@ -375,6 +381,9 @@ fn decode<H: Hasher, S: Strategy>(
     if chunks.len() < k {
         return Err(Error::NotEnoughChunks);
     }
+    if !chunks.iter().all(|c| &c.root == root) {
```
patrick-ogrady: Could do during `chunk` loop below to avoid extra loop?

coding/src/reed_solomon.rs
```diff
@@ -667,10 +654,11 @@ mod tests {
 
     fn checked(
         chunk: Chunk<<Sha256 as Hasher>::Digest>,
+        root: <Sha256 as Hasher>::Digest,
     ) -> CheckedChunk<<Sha256 as Hasher>::Digest> {
         let Chunk { shard, index, .. } = chunk;
         let digest = Sha256::hash(&shard);
-        CheckedChunk::new(shard, index, digest)
+        CheckedChunk::new(shard, index, digest, root)
```
patrick-ogrady: feels like root should go first?

coding/src/reed_solomon.rs
```diff
@@ -684,9 +672,9 @@ mod tests {
 
         // Use a mix of original and recovery pieces
         let pieces: Vec<_> = vec![
-            checked(chunks[0].clone()), // original
-            checked(chunks[4].clone()), // recovery
-            checked(chunks[6].clone()), // recovery
+            checked(chunks[0].clone(), root), // original
```
patrick-ogrady: root first?

consensus/src/marshal/coding/shards/engine.rs
patrick-ogrady: Will re-attach.

consensus/src/marshal/coding/shards/engine.rs
```diff
@@ -138,25 +108,26 @@
...
 //! Validation and blocking rules are applied while a commitment is actively
 //! tracked in reconstruction state. Once a block is already reconstructed and
 //! cached, additional shards for that commitment are ignored.
 //!
-//! _Strong shards are only accepted from the leader. If the leader is not
-//! yet known, shards are buffered in fixed-size per-peer queues until consensus
-//! signals the leader via [`Discovered`]. Once leader is known, buffered
-//! shards for that commitment are ingested into the active state machine._
+//! _If the leader is not yet known, shards are buffered in fixed-size per-peer
+//! queues until consensus signals the leader via [`Discovered`]. Once leader
```
patrick-ogrady, line 128: This wording is now fine but that earlier comment was that we may set this via certificate.

consensus/src/marshal/coding/shards/engine.rs
```diff
@@ -718,30 +682,29 @@ where
             return false;
         };
 
-        // Ingest weak shards first so they populate pending_weak_shards before
-        // the strong shard sets checking_data and triggers batch validation.
         let mut progressed = false;
```
patrick-ogrady, line 707: Could retain some inline comment here?

consensus/src/marshal/coding/shards/engine.rs
```diff
@@ -1142,54 +1098,54 @@ where
...
             return false;
         };
+        self.common.checking_data = Some(checking_data);
 
-        // Only persist the strong shard (for later equivocation detection) after
-        // it has passed `C::weaken` verification.
-        self.common.received_strong = Some(received_strong);
         self.common.contributed.set(u64::from(index), true);
         self.common.checked_shards.push(checked);
+        self.common.own_shard_verified = true;
```
patrick-ogrady: Can we not infer this by checking .received_shards? I suppose it is a way to make it simpler between leader shard vs echo shard?

consensus/src/marshal/coding/shards/engine.rs
```diff
@@ -1198,49 +1154,74 @@ where
...
         }
-        self.common.checked_shards.extend(new_checked);
+        for (checked, cd) in new_checked {
+            self.common.checked_shards.push(checked);
+            if self.common.checking_data.is_none() {
+                self.common.checking_data = Some(cd);
+            }
+        }
 
-        // After validation, some may have failed; recheck threshold.
```
patrick-ogrady, line 1227: Can retain?

consensus/src/marshal/coding/shards/engine.rs
```diff
@@ -1198,49 +1154,74 @@ where
...
         if self.common.checked_shards.len() < minimum {
             return None;
         }
 
-        // Transition to Ready.
```
patrick-ogrady, line 1232: retain?

consensus/src/marshal/coding/shards/engine.rs
clabby: Yeah, this is just a flag. `received_shards` holds all shards and we could check `received_shards.get(leader).is_some()`, but we specifically want to know that condition _and_ that we’ve validated that shard

clabby: (`received_shards` is a a buffer of unverified shards.)

coding/src/lib.rs
clabby: The interface takes ownership of `CheckingData` to avoid cloning it in the case where we pass `Some` - since this function always returns `CheckingData`, we just pass it back to the user if we didn’t generate it internally.

coding/src/zoda.rs
```diff
@@ -719,56 +662,55 @@ impl<H: Hasher> Scheme for Zoda<H> {
...
         index: u16,
-        shard: Self::StrongShard,
-    ) -> Result<(Self::CheckingData, Self::CheckedShard, Self::WeakShard), Self::Error> {
-        let weak_shard = WeakShard {
-            inclusion_proof: shard.inclusion_proof,
-            shard: shard.rows,
+        shard: &Self::Shard,
+        checking_data: Option<Self::CheckingData>,
+    ) -> Result<(Self::CheckedShard, Self::CheckingData), Self::Error> {
+        let checking_data = match checking_data {
```
patrick-ogrady: `Zoda::check(..., Some(checking_data))` now trusts cached `CheckingData` after only checking `cd.commitment == *commitment`, so it no longer revalidates the shard metadata that is now carried on the wire. After one honest shard seeds the cache, a relayed shard with mutated `data_bytes`, `root`, or `checksum` can be accepted on the cached path even though the uncached path rejects it. That creates a real malleability hole on the unified shard format and makes cached vs. uncached verification disagree on the same bytes.

Please make the cached path revalidate the transmitted metadata before reusing cached state, for example by storing the expected metadata in `CheckingData` and comparing it, or by recomputing `CheckingData::reckon` when the incoming shard metadata does not match.

Regression tests from the candidates both fail on the current code:

```rust
#[test]
fn cached_check_does_not_hide_mutated_metadata() {
    let config = Config {
        minimum_shards: NZU16!(2),
        extra_shards: NZU16!(1),
    };
    let data = b"metadata coverage";
    let (commitment, mut shards) = Zoda::<Sha256>::encode(&config, &data[..], &STRATEGY).unwrap();

    let (_, cd) = Zoda::<Sha256>::check(&config, &commitment, 0, &shards[0], None).unwrap();
    shards[1].data_bytes += 1;

    assert!(matches!(
        Zoda::<Sha256>::check(&config, &commitment, 1, &shards[1], None),
        Err(Error::InvalidShard)
    ));
    assert!(matches!(
        Zoda::<Sha256>::check(&config, &commitment, 1, &shards[1], Some(cd)),
        Err(Error::InvalidShard)
    ));
}
```

```rust
#[test]
fn checksum_malleability_with_cached_checking_data() {
    let config = Config {
        minimum_shards: NZU16!(2),
        extra_shards: NZU16!(1),
    };
    let data = vec![0x5Au8; 256 * 1024];
    let (commitment, mut shards) =
        Zoda::<Sha256>::encode(&config, &data[..], &STRATEGY).unwrap();

    let (_, checking_data) =
        Zoda::<Sha256>::check(&config, &commitment, 0, &shards[0], None).unwrap();

    let mut checksum = (*shards[1].checksum).clone();
    checksum[(0, 0)] += &F::one();
    shards[1].checksum = Arc::new(checksum);

    let result =
        Zoda::<Sha256>::check(&config, &commitment, 1, &shards[1], Some(checking_data));
    assert!(matches!(result, Err(Error::InvalidShard)));
}
```

clabby: Thanks patchwork :heart_hands:

coding/src/zoda.rs
```diff
@@ -544,57 +491,70 @@ impl<D: Digest> CheckingData<D> {
...
-        if weak_shard.shard.rows() != self.topology.samples
-            || weak_shard.shard.cols() != self.topology.data_cols
+        if shard.rows.rows() != self.topology.samples
+            || shard.rows.cols() != self.topology.data_cols
         {
-            return Err(Error::InvalidWeakShard);
+            return Err(Error::InvalidShard);
         }
-        let index = index as usize;
+        let shard_idx = index as usize;
```
cronokirby: Nit: since we don't ever use `index` again, I'd prefer just shadowing `index` (like the code used to do)

coding/src/lib.rs
```diff
@@ -221,6 +336,9 @@ commonware_macros::stability_scope!(ALPHA {
             weak_shard: Self::WeakShard,
         ) -> Result<Self::CheckedShard, Self::Error>;
 
+        /// Construct an insufficient-shards error for this scheme.
+        fn insufficient_shards(actual: usize, expected: usize) -> Self::Error;
```
clabby: rm - slop

consensus/src/marshal/coding/shards/engine.rs
patrick-ogrady, line 1: - [P2] Accept late leader shards after quorum transition — /Users/patrickogrady/code/monorepo-pr-3338/consensus/src/marshal/coding/shards/
    engine.rs:1378-1380
    If a participant accumulates minimum_shards peer gossips before its direct leader shard is processed (either live or when draining buffered
    shards), try_transition() can move the state to Ready first. This early return then drops the late leader shard, so verify_own_shard() never
    runs and pending_action is never populated. The participant therefore never re-broadcasts its own shard, which can leave slower peers stuck
    below quorum when they did not receive the same gossip subset.

consensus/src/marshal/coding/shards/engine.rs
```diff
@@ -1198,49 +1145,43 @@ where
...
         blocker: &mut impl Blocker<PublicKey = P>,
     ) -> Option<ReadyState<P, C, H>> {
-        self.checking_data.as_ref()?;
         let minimum = usize::from(commitment.config().minimum_shards.get());
-        if self.common.checked_shards.len() + self.pending_weak_shards.len() < minimum {
+        if self.common.checked_shards.len() + self.pending_shards.len() < minimum {
             return None;
         }
 
-        // Batch-validate all pending weak shards in parallel.
```
patrick-ogrady, line 1207: retain this

consensus/src/marshal/coding/shards/engine.rs
```diff
@@ -1328,56 +1262,61 @@ where
...
-    /// Handle an incoming network shard.
+    /// The following conditions cause a shard to be silently ignored
+    /// without blocking the sender:
     ///
-    /// Returns `true` only when the shard caused state progress (buffered,
-    /// validated, or transitioned), and `false` when rejected/blocked.
+    /// - Exact duplicate of a previously received shard for the same index.
+    /// - The index has already been marked as contributed (via the bitmap,
+    ///   e.g. after batch validation).
+    /// - Shards that arrive after the state has transitioned to [`ReconstructionState::Ready`]
```
patrick-ogrady: Maybe the drop on the transition to Ready is expected (I think we want to ensure others can rebuild tho)?

patrick-ogrady:
Found some comments were a bit outdated...maybe something like?

```
diff --git a/consensus/src/marshal/coding/marshaled.rs b/consensus/src/marshal/coding/marshaled.rs
index af042c890..eeb6abeeb 100644
--- a/consensus/src/marshal/coding/marshaled.rs
+++ b/consensus/src/marshal/coding/marshaled.rs
@@ -18,9 +18,14 @@
 //! This wrapper integrates with a variant of marshal that supports erasure coded broadcast. When a leader
 //! proposes a new block, it is automatically erasure encoded and its shards are broadcasted to active
 //! participants. When verifying a proposed block (the precondition for notarization), the wrapper
-//! ensures the commitment's context digest matches the consensus context and subscribes to shard validity
-//! for the shard received by the proposer. If the shard is valid, the local shard is relayed to all
-//! other participants to aid in block reconstruction.
+//! ensures the commitment's context digest matches the consensus context and waits for validation of
+//! the shard assigned to this participant by the proposer. If that shard is valid, the local shard is
+//! relayed to all other participants to aid in block reconstruction.
+//!
+//! A participant may still reconstruct the full block from gossiped shards before its designated
+//! leader-delivered shard arrives. That is sufficient for later certification and repair flows, but it
+//! is not treated as notarization readiness: a participant only helps form a notarization once it has
+//! validated the shard it is supposed to echo.
 //!
 //! During certification (the phase between notarization and finalization), the wrapper subscribes to
 //! block reconstruction and validates epoch boundaries, parent commitment, height contiguity, and
@@ -771,9 +776,11 @@ where
 
         match scheme.me() {
             Some(_) => {
-                // Subscribe to local shard readiness. This completes once we
-                // can vote on the proposal, either because our shard verified
-                // directly or because the block was reconstructed first.
+                // Subscribe to local shard readiness. For participants, this
+                // only completes once the leader-delivered shard for our
+                // assigned index has been verified. Reconstructing the block
+                // from peer gossip is useful for certification later, but is
+                // not enough to emit a notarize vote.
                 let validity_rx = self.shards.subscribe_shard(payload).await;
                 let (tx, rx) = oneshot::channel();
                 self.context
diff --git a/consensus/src/marshal/coding/shards/engine.rs b/consensus/src/marshal/coding/shards/engine.rs
index 07e855176..13fc29c07 100644
--- a/consensus/src/marshal/coding/shards/engine.rs
+++ b/consensus/src/marshal/coding/shards/engine.rs
@@ -322,8 +322,13 @@ where
     /// Open subscriptions for local shard readiness for the keyed
     /// [`Commitment`].
     ///
-    /// Readiness is satisfied once our shard has been verified directly or
-    /// once the full block has already been reconstructed.
+    /// For participants, readiness is satisfied once the leader-delivered
+    /// shard for the local participant index has been verified. Block
+    /// reconstruction from peer gossip is tracked separately and does not
+    /// satisfy this readiness condition.
+    ///
+    /// Proposers are a special case: they satisfy readiness once their local
+    /// proposal is cached because they already hold all shards.
     shard_subscriptions: BTreeMap<Commitment, Vec<oneshot::Sender<()>>>,
 
     /// Open subscriptions for the reconstruction of a [`CodedBlock`] with
@@ -860,6 +865,9 @@ where
     }
 
     /// Handles the registry of a local shard readiness subscription.
+    ///
+    /// For participants this is tied to verification of the leader-delivered
+    /// shard for the local index, not to generic block reconstruction.
     fn handle_shard_subscription(&mut self, commitment: Commitment, response: oneshot::Sender<()>) {
         // Answer immediately if our own shard has been verified.
         let has_shard = self
diff --git a/consensus/src/marshal/coding/shards/mailbox.rs b/consensus/src/marshal/coding/shards/mailbox.rs
index ed12ec396..215855bfc 100644
--- a/consensus/src/marshal/coding/shards/mailbox.rs
+++ b/consensus/src/marshal/coding/shards/mailbox.rs
@@ -52,10 +52,14 @@ where
     },
     /// A request to open a subscription for local shard readiness.
     ///
-    /// This resolves once the node can proceed as if it had validated its
-    /// shard for the commitment, either because the leader's shard for this
-    /// node was verified directly or because the full block was reconstructed
-    /// before that shard arrived.
+    /// For participants, this resolves once the leader-delivered shard for
+    /// the local participant index has been verified. Reconstructing the full
+    /// block from gossiped shards does not resolve this subscription: that
+    /// block may still be used for later certification, but it is not enough
+    /// to claim the participant received the shard it is expected to echo.
+    ///
+    /// For proposers, this resolves immediately after the locally built block
+    /// is cached because they trivially have all shards.
     SubscribeShard {
         /// The block's commitment.
         commitment: Commitment,
@@ -151,10 +155,14 @@ where
 
     /// Subscribe to local shard readiness for a commitment.
     ///
-    /// This resolves once the node can proceed as if it had validated its
-    /// shard for the commitment, either because the leader's shard for this
-    /// node was verified directly or because the full block was reconstructed
-    /// before that shard arrived.
+    /// For participants, this resolves once the leader-delivered shard for
+    /// the local participant index has been verified. Reconstructing the full
+    /// block from gossiped shards does not resolve this subscription: that
+    /// block may still be used for later certification, but it is not enough
+    /// to claim the participant received the shard it is expected to echo.
+    ///
+    /// For proposers, this resolves immediately after the locally built block
+    /// is cached because they trivially have all shards.
     pub async fn subscribe_shard(&self, commitment: Commitment) -> oneshot::Receiver<()> {
         let (responder, receiver) = oneshot::channel();
         let msg = Message::SubscribeShard {
```

consensus/src/marshal/coding/shards/engine.rs
```diff
@@ -530,8 +508,19 @@ where
                     .inc();
 
                 let commitment = shard.commitment();
+
+                // Skip shards for already-reconstructed blocks unless the
+                // leader's own shard for our index has not been verified yet.
+                // Accepting the late leader shard ensures we broadcast it and
+                // help slower peers reach quorum.
                 if self.reconstructed_blocks.contains_key(&commitment) {
```
patrick-ogrady: I think we want to skip if we've reconstructed and the shard isn't for us?

clabby: We do, but we don’t check that here. That handling is separated into the reconstruction state machine

consensus/src/marshal/coding/shards/engine.rs
```diff
@@ -1131,65 +1110,66 @@ where
...
-            commitment,
-            index,
-            data,
-        } = shard;
-        let received_strong = data.clone();
-        let Ok((checking_data, checked, weak_shard_data)) =
-            C::weaken(&commitment.config(), &commitment.root(), index, data)
-        else {
-            commonware_p2p::block!(blocker, sender, "invalid strong shard received");
+        let index = shard.index;
```
patrick-ogrady: nit: I think its easier just to use shard.index inline

consensus/src/marshal/coding/shards/engine.rs
```diff
@@ -1395,143 +1370,96 @@ where
...
+            ctx.scheme.me().unwrap_or(sender_index)
+        } else {
+            sender_index
+        };
+        let expected_index: u16 = expected_participant
             .get()
             .try_into()
             .expect("participant index impossibly out of bounds");
-        if shard.index != expected_index {
+
```
patrick-ogrady: nit: remove

consensus/src/marshal/coding/shards/engine.rs
```diff
@@ -1395,143 +1370,96 @@ where
...
-            }
-            return false;
+        // Buffer for batch validation.
+        state
+            .common
+            .received_shards
+            .insert(indexed.index, indexed.data.clone());
+        state.common.contributed.set(u64::from(indexed.index), true);
+        state.pending_shards.insert(sender, indexed);
+
```
patrick-ogrady: nit: remove

consensus/src/marshal/coding/shards/engine.rs
```diff
@@ -4492,6 +4328,142 @@ mod tests {
         );
     }
 
+    #[test_traced]
+    fn test_withholding_leader_victim_reconstructs_via_gossip() {
```
patrick-ogrady, line 4355: It is probably worthwhile to have a negative test where a proposer that only sends out >=f+1 <2f+1 shards (but 2f+1 votes) won't see a notarization? Not sure if we have that control with the harness.

clabby: Ah yeah, we don't have that level of e2e here. The best we could do is send out `2f` shards to `2f+1` participants and assert that only the `2f`'s subscriptions resolve, though we already have a negative assertion for the more granular case of "the shard subscription cannot resolve without receiving and validating your own shard from the leader" in `test_shard_subscription_pending_after_reconstruction_without_leader_shard`.

This would be an excellent test to add to the constantinople harness.

consensus/src/marshal/coding/shards/engine.rs
```diff
@@ -530,7 +508,8 @@ where
                     .inc();
 
                 let commitment = shard.commitment();
-                if self.reconstructed_blocks.contains_key(&commitment) {
+
```
patrick-ogrady: nit: extra \n

patrick-ogrady:
dumb idea: could a savvy proposer just choose 2f+1 good validators to send to to reduce their expansion factor lol?

PR #3343, opened 2026-03-09
Fix byzantine sender in replica

dnkolegov-ar:
This PR fixes a minor bug in the Byzantine replica logic of the Minimmit.

### DIFF
--- a/pipeline/minimmit/quint/replica.qnt
+++ b/pipeline/minimmit/quint/replica.qnt
@@ -635,12 +635,11 @@ module replica {
             nondet senders = BYZANTINE.powerset().oneOf()
             nondet view = VIEWS.oneOf()
             nondet block = AllBlocks.union(Set(DUMMY_BLOCK)).oneOf()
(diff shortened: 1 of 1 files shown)


pipeline/minimmit/quint/replica.qnt
```diff
@@ -635,12 +635,11 @@ module replica {
...
             nondet view = VIEWS.oneOf()
             nondet block = AllBlocks.union(Set(DUMMY_BLOCK)).oneOf()
-            nondet sig = oneOf(BYZANTINE)
             nondet k = oneOf(Set(NotarizeKind, NullifyKind))
-            val votes = senders.map(s => {
+            val votes = senders.map(sender => {
                 view: view,
                 block: block,
-                sig: sig,
+                sig: sender,
```
patrick-ogrady, line 642: Should we rename `sig` -> `sender`?

dnkolegov-ar: No, because in the model signature = ID

PR #3347, opened 2026-03-09
[consensus/simplex] Add policy to handle non-voters' block dissemination

clabby:
## Overview

In the `standard` path of marshal, it is possible for proposers to withhold a block from the next leader. This change adds a policy for honest peers to help all active participants that did not vote, by sending them the full block.

This change is configurable - users running the standard variant can choose between:
- Doing nothing (the default, and the current behavior).
- Forwarding block to all active participants that did not vote in the certificate the batcher constructed or received from the network.

Preferred option over https://github.com/commonwarexyz/monorepo/pull/3326

### DIFF
--- a/consensus/src/simplex/actors/batcher/actor.rs
+++ b/consensus/src/simplex/actors/batcher/actor.rs
@@ -330,17 +330,20 @@ where
 
                     // If the leader nullified this view or has not been active
                     // recently, tell the voter to reduce the leader timeout to now
+                    #[allow(clippy::collapsible_else_if)]
                     let timeout_reason = if Self::leader_nullified(&current, &work) {
                         // Leader already buffered a nullify for this now-current view
                         // (allowed because we accept votes up to `current+1`)
                         Some(TimeoutReason::LeaderNullify)
-                    } else if am_leader {
-                        // If we are the leader, we should not timeout
-                        None
                     } else {
-                        // If we are not the leader and the leader isn't active, we should timeout
-                        (!self.is_active(&work, current.view, leader))
-                            .then_some(TimeoutReason::Inactivity)
+                        if am_leader {
+                            // If we are the leader, we should not timeout
+                            None
+                        } else {
+                            // If we are not the leader and the leader isn't active, we should timeout
+                            (!self.is_active(&work, current.view, leader))
+                                .then_some(TimeoutReason::Inactivity)
+                        }
                     };
                     if timeout_reason.is_some() {
                         current.timed_out = true;


consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -128,12 +133,14 @@ impl<
         // TODO(#1833): Metrics should use the post-start context
         let clock = Arc::new(context.clone());
         let (sender, receiver) = mpsc::channel(cfg.mailbox_size);
+        let elector = cfg.elector.build(cfg.scheme.participants());
```
clabby: We currently compute the next leader twice as to not add a communication path for leader election from the batcher -> voter. This is cheap, but worth calling out.

consensus/src/marshal/core/variant.rs
```diff
@@ -12,14 +12,19 @@
...
+    Block,
+};
 use commonware_codec::{Codec, Read};
-use commonware_cryptography::{Digest, Digestible};
+use commonware_cryptography::{Digest, Digestible, PublicKey};
 use commonware_utils::channel::oneshot;
 use std::{future::Future, sync::Arc};
 
 /// A marker trait describing the types used by a variant of Marshal.
 pub trait Variant: Clone + Send + Sync + 'static {
```
clabby, line 23: Consider giving `needs_block_targets` a default implementation returning `None`?

consensus/src/marshal/standard/variant.rs
```diff
@@ -4,15 +4,25 @@
...
     Block,
 };
 use commonware_broadcast::{buffered, Broadcaster};
 use commonware_cryptography::{Digestible, PublicKey};
 use commonware_p2p::Recipients;
 use commonware_utils::channel::oneshot;
 
+/// Configuration for the standard marshal variant.
+#[derive(Debug, Clone, Copy, Default)]
+pub struct StandardConfig {
```
clabby: Should mark this as `#[non_exhaustive]`.

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -532,11 +539,21 @@ impl<
...
 
                 // Try to construct and forward certificates
+                let mut needs_block = None;
                 if let Some(notarization) = self
                     .recover_latency
                     .time_some(|| round.try_construct_notarization(&self.scheme, &self.strategy))
                 {
                     debug!(view = %updated_view, "constructed notarization, forwarding to voter");
+
+                    // Elect the leader for the next view using the notarization certificate.
```
patrick-ogrady: We should just listen for the update message from the voter to enter a new view (includes leader).

consensus/src/marshal/core/actor.rs
```diff
@@ -583,6 +588,46 @@ where
...
+                    } => {
+                        // Determine which participants to target based on variant config.
+                        let Some(targets) =
+                            V::needs_block_targets(&self.variant_config, next_leader, missing)
+                        else {
+                            continue;
+                        };
+
+                        let commitment = proposal.payload;
+                        let epoch = proposal.round.epoch();
```
patrick-ogrady: move to 606 (after `//` below)

consensus/src/marshal/core/actor.rs
```diff
@@ -583,6 +588,46 @@ where
...
+                        missing,
+                    } => {
+                        // Determine which participants to target based on variant config.
+                        let Some(targets) =
+                            V::needs_block_targets(&self.variant_config, next_leader, missing)
+                        else {
+                            continue;
+                        };
+
+                        let commitment = proposal.payload;
```
patrick-ogrady: move to after 612

consensus/src/marshal/core/mailbox.rs
```diff
@@ -131,6 +131,15 @@ pub(crate) enum Message<S: Scheme, V: Variant> {
         /// The finalization.
         finalization: Finalization<S, V::Commitment>,
     },
+    /// A notarization was constructed but some peers may not have the block.
```
patrick-ogrady: I think we want to generalize this to `A certificate was constructed` (we should also only send if certify or finalization, not just raw notarization?)

consensus/src/marshal/core/variant.rs
```diff
@@ -141,6 +162,18 @@ pub trait Buffer<V: Variant>: Clone + Send + Sync + 'static {
     ///
     /// This handles the initial dissemination of a newly proposed block.
     fn proposed(&self, round: Round, block: V::Block) -> impl Future<Output = ()> + Send;
+
+    /// Send a block to specific peers.
+    ///
+    /// Used to proactively push blocks to peers that may not have them
+    /// (e.g. peers that did not send a notarize vote).
+    fn send(
```
patrick-ogrady, line 145: Should we combine with `proposed`?

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -532,11 +549,18 @@ impl<
...
                 // Try to construct and forward certificates
+                let mut needs_block_candidates = None;
                 if let Some(notarization) = self
                     .recover_latency
                     .time_some(|| round.try_construct_notarization(&self.scheme, &self.strategy))
                 {
                     debug!(view = %updated_view, "constructed notarization, forwarding to voter");
+
+                    // Collect candidates while we still hold `round`.
+                    let candidates = round.missing_notarize_voters();
```
patrick-ogrady: I think we want to wait to collect as late as possible (we may reduce our burden while waiting for certify)?

I suppose the most correct approach is to just do this when we notarize tho....

consensus/src/simplex/types.rs
```diff
@@ -2051,6 +2087,16 @@ impl<S: Scheme, D: Digest> Write for Activity<S, D> {
                 9u8.write(writer);
                 v.write(writer);
             }
+            Self::NeedsBlock {
+                proposal,
+                next_leader,
+                missing,
+            } => {
+                10u8.write(writer);
```
patrick-ogrady: This is a breaking change as written. This doesn't go over the wire but the others do 🤔

patrick-ogrady (commented):
Initial thoughts: we should sync on whether to do this after notarization, certification, finalization

consensus/src/simplex/actors/batcher/actor.rs
clabby: I think we should send it when we notarize - it's the safest way to get the next leader the block within $2\Delta$, since we can pipeline sending the block with certification?

consensus/src/lib.rs
```diff
@@ -175,6 +175,19 @@ stability_scope!(BETA, cfg(not(target_arch = "wasm32")) {
         }
     }
 
+    /// Describes how a payload should be disseminated to the network.
+    pub enum Dissemination<P: PublicKey> {
```
patrick-ogrady: How would feel about `Initial` and `Retry`? May be more generic idk.

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -560,12 +620,58 @@ impl<
...
+                                        proposal.payload,
+                                        Dissemination::Forward {
+                                            round: proposal.round,
+                                            peers,
+                                        },
+                                    )
+                                    .await;
+                            }
+                        }
+                    } else if next_view > current.view {
```
patrick-ogrady: `ForwardingPolicy::All` is still routed through the `updated_view.next()` deferral/drop logic, so a notarization for view `v` can be lost permanently if the voter reaches `v+2` before the batcher finishes assembling it. In that case `next_view < current.view`, this branch neither enqueues nor emits a forward, and the stale-entry pruning below ensures it is never retried. That turns forwarding into a race against view updates even though `All` does not depend on learning the next leader.

Please emit immediately for `ForwardingPolicy::All`, or retarget stale entries instead of discarding them.

Regression test evidence:

Command: `cargo test -p commonware-consensus simplex::actors::batcher::tests::test_forward_not_dropped_after_skipping_next_view -- --exact`  
Observed result: `assert_eq!(broadcasts.len(), 1)` failed with `left: 0` and `right: 1` after the batcher logged `constructed notarization, forwarding to voter view=2`.

```rust
#[test_traced]
fn test_forward_not_dropped_after_skipping_next_view() {
    // Reuse the existing forwarding test setup, but jump two views ahead
    // before the view-2 quorum arrives.
    batcher_mailbox.update(View::new(2), Participant::new(2), View::zero()).await;
    batcher_mailbox.update(View::new(3), Participant::new(3), View::zero()).await;
    batcher_mailbox.update(View::new(4), Participant::new(4), View::zero()).await;

    let proposal = Proposal::new(
        Round::new(epoch, View::new(2)),
        View::new(1),
        Sha256::hash(b"view2_payload"),
    );

    // Build a view-2 notarization with participant 4 missing.
    for i in 1..quorum_size {
        let vote = Notarize::sign(&schemes[i], proposal.clone()).unwrap();
        participant_senders[i]
            .as_mut()
            .unwrap()
            .send(Recipients::One(me.clone()), Vote::Notarize(vote).encode(), true)
            .await
            .unwrap();
    }
    batcher_mailbox
        .constructed(Vote::Notarize(
            Notarize::sign(&schemes[0], proposal.clone()).unwrap(),
        ))
        .await;

    context.sleep(Duration::from_millis(150)).await;
    assert_eq!(relay.broadcasts.lock().len(), 1);
}
```

consensus/src/marshal/core/actor.rs
```diff
@@ -516,7 +517,22 @@ where
...
+                    }
+                    Message::Forwarded {
+                        round,
+                        commitment,
+                        peers,
+                    } => {
+                        if peers.is_empty() {
+                            continue;
+                        }
+                        let Some(block) = self.find_block_by_commitment(&buffer, commitment).await
```
patrick-ogrady, line 530: The new targeted-send path drops a forward on cache miss with only a `debug!` log. Since `Dissemination::Forward` is now part of the non-voter recovery path, a lookup regression here becomes a silent liveness failure that will be very difficult to diagnose in production.

Please promote this to a visible signal, such as a `warn!` and/or counter that includes `round`, `commitment`, and recipient count.

consensus/src/simplex/actors/batcher/actor.rs
patrick-ogrady, line 1: The certificate-receive path never applies the forwarding policy. If this node already has the proposal but only learns the notarization from `certificate_receiver`, it stores the certificate and wakes the voter, but never forwards the block to the next leader or other missing peers. That makes targeted forwarding depend on which node happened to aggregate `2f+1` votes locally, which reintroduces the same liveness gap this feature is supposed to close.

Please run the same missing-peer calculation and relay logic after accepting a verified network notarization, or explicitly document why network-learned notarizations are intentionally excluded and safe.

Regression test evidence:

Command: `CARGO_TARGET_DIR=/tmp/cw-net cargo test -p commonware-consensus simplex::actors::batcher::tests::test_forward_emitted_for_network_notarization -- --exact`  
Observed result: `assert_eq!(broadcasts.len(), 1)` failed with `left: 0` and `right: 1`; the only relay activity was the earlier view-1 local notarization.

```rust
#[test_traced]
fn test_forward_emitted_for_network_notarization() {
    // Establish recent activity, then move to view 3 so the next leader is known.
    batcher_mailbox.update(View::new(1), Participant::new(1), View::zero()).await;
    // ... send a full view-1 quorum and drain the resulting proposal/notarization ...

    batcher_mailbox.update(View::new(2), Participant::new(2), View::zero()).await;
    batcher_mailbox.update(View::new(3), Participant::new(4), View::zero()).await;

    let proposal = Proposal::new(
        Round::new(epoch, View::new(2)),
        View::new(1),
        Sha256::hash(b"view2_payload"),
    );

    // Record only a partial local view of the voters so this node does not
    // assemble the certificate itself.
    for i in 1..(quorum_size - 1) {
        let vote = Notarize::sign(&schemes[i], proposal.clone()).unwrap();
        participant_senders[i]
            .as_mut()
            .unwrap()
            .send(Recipients::One(me.clone()), Vote::Notarize(vote).encode(), true)
            .await
            .unwrap();
    }
    batcher_mailbox
        .constructed(Vote::Notarize(
            Notarize::sign(&schemes[0], proposal.clone()).unwrap(),
        ))
        .await;

    // Learn the notarization from the network.
    let notarization = build_notarization(&schemes, &proposal, quorum_size);
    injector_sender
        .send(
            Recipients::One(me.clone()),
            Certificate::Notarization(notarization).encode(),
            true,
        )
        .await
        .unwrap();

    context.sleep(Duration::from_millis(100)).await;
    assert_eq!(relay.broadcasts.lock().len(), 1);
}
```

consensus/src/simplex/actors/batcher/actor.rs
clabby: I went back and forth on this myself, and I'm not convinced that this is something that we should be handling.

The goal of this mechanism is to help out the next leader. When we're skipping ahead views, we've already bypassed the window in which forwarding the block is useful for that purpose. At least `f+1` honest and online nodes observed the notarized blocks in the range of views we're skipping over, and they should have attempted to forward to `All` when they received or constructed the notarizations..

An argument can be made that in `All` mode, this is desired, but I'm not so sure. What do you think?

consensus/src/simplex/actors/batcher/actor.rs
clabby: This is expected, but let me know if we should be doing this. My thinking was that if you're constructing the certificate, you have a complete log of who did not vote, whereas you're not guaranteed to have this if you're receiving the certificate from someone else verbatim.

consensus/src/marshal/core/actor.rs
clabby: We use `debug!` in other cases where we're missing the block, this is for consistency.

consensus/src/simplex/actors/batcher/actor.rs
patrick-ogrady: This weakens our robustness to malicious leaders (as a leader could send to f+1 honest, f byzantine, build a certificate, and send to all -> remaining f honest don't see block).

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -166,9 +174,91 @@ impl<
...
+                        round: proposal.round,
+                        peers,
+                    },
+                )
+                .await;
+        } else if next_view > current.view {
+            pending_forwards.insert(next_view, (proposal, missing));
+        }
+        // If next_view < current.view, we've moved past; drop.
+    }
```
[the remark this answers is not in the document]
clabby: https://github.com/commonwarexyz/monorepo/pull/3347#discussion_r2914403030

consensus/src/simplex/actors/batcher/actor.rs
clabby: I suppose the edge case where doing it this way would help is:

1. A quorum of `notarize(v)` votes exist
2. A `nullification(v)` forms first, so nodes move to `v+1`
3. Nodes nullify `v+1`, `nullification(v+1)` forms, and everyone moves to `v+2`
4. Only later does some node learn/reconstruct `notarization(v)`
5. Any node that now has both `block(v)` and `notarization(v)` can certify `v`
6. Then proposals in `v+2` may legally build on `v`, because `v` is certified and `v+1` is nullified.

We currently cancel the `All` broadcast for `v` when we advance, though that'd hurt `leader(v+2)` here if they were never sent the block by `leader(v)`.

clabby: I'm convinced this is more correct for `All` than the existing impl - adding it in.

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -166,9 +174,98 @@ impl<
...
+                    round: proposal.round,
+                    peers,
+                },
+            )
+            .await;
+    }
+
+    /// Returns missing voters that have been active recently enough to justify
+    /// a targeted forward.
+    fn active_missing_voters(
```
patrick-ogrady: I'm pondering whether we should skip this optimization (I think we'll drop anyone not connected, so I think this is fine to not worry about).

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -274,6 +372,14 @@ impl<
                     }
                     response.send_lossy(timeout_reason);
 
+                    // Emit any pending forwards for this view now that we
+                    // know the leader from the voter.
+                    if let Some((proposal, missing)) = pending_forwards.remove(&new_current) {
```
patrick-ogrady: Maybe we should just make this a boolean at first (I became more concerned about the multi-view scenario where we just send to one leader when we really need to send to all possible leaders for this to stay stable).

consensus/src/simplex/actors/batcher/round.rs
```diff
@@ -353,6 +353,25 @@ impl<
...
 
+    /// Returns participant indices that did not vote for `proposal`.
+    ///
+    /// Uses `pending_votes` rather than `verified_votes` because we only
+    /// verify the first quorum of votes. A peer whose matching vote arrived
+    /// after quorum but before the certificate is still tracked in pending.
+    ///
+    /// Votes for a conflicting proposal are treated as missing because those
+    /// peers still need the winning block forwarded.
+    pub fn missing_notarize_voters(&self, proposal: &Proposal<D>) -> Vec<Participant> {
```
patrick-ogrady: I think this can scan over notarize and finalize?

consensus/src/simplex/actors/batcher/actor.rs
clabby: It’s short enough imo - can’t there be cases where peers are online and connected, but inactive?

consensus/src/simplex/actors/batcher/actor.rs
clabby: I kept it as an enum to reserve our ability to add new modes in the future (i.e. a smarter `NextLeader`)

consensus/src/lib.rs
clabby: I don’t think `Retry` works too well, since the nodes that are forwarding the block never tried to send the block to the non-voters to begin with.

consensus/src/lib.rs
```diff
@@ -183,12 +196,15 @@ stability_scope!(BETA, cfg(not(target_arch = "wasm32")) {
...
-        /// so timely delivery often yields better performance.
-        fn broadcast(&mut self, payload: Self::Digest) -> impl Future<Output = ()> + Send;
+        /// Identity key of a network participant.
+        type PublicKey: PublicKey;
+
+        /// Broadcast a payload to the given recipients.
+        fn broadcast(
+            &mut self,
+            payload: Self::Digest,
+            dissemination: Dissemination<Self::PublicKey>,
```
patrick-ogrady: Wondering if we make `dissemination` an associated type? `Forward` doesn't make sense for ordered-broadcast/aggregation, for example?

consensus/src/lib.rs
```diff
@@ -183,12 +183,22 @@ stability_scope!(BETA, cfg(not(target_arch = "wasm32")) {
...
+
+        /// Strategy for how a payload should be disseminated.
         ///
-        /// Other participants may not begin voting on a proposal until they have the full contents,
-        /// so timely delivery often yields better performance.
-        fn broadcast(&mut self, payload: Self::Digest) -> impl Future<Output = ()> + Send;
+        /// Consensus mechanisms that need rich dissemination control (e.g. distinguishing
+        /// proposals from targeted forwards) define a custom enum here. Mechanisms that
+        /// treat every broadcast identically can set this to `()`.
+        type Dissemination: Send;
```
patrick-ogrady: nit: `Strategy`

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -560,6 +638,12 @@ impl<
                         .await;
                 }
 
+                // Filter candidates and emit forwarding.
+                if let Some((proposal, candidates)) = forward_candidates {
+                    let missing = self.active_missing_voters(&work, candidates);
+                    self.forward_notarization(proposal, missing).await;
```
patrick-ogrady: nit: `forward_proposal`?

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -532,11 +602,19 @@ impl<
...
 
                 // Try to construct and forward certificates
+                let mut forward_candidates = None;
                 if let Some(notarization) = self
                     .recover_latency
                     .time_some(|| round.try_construct_notarization(&self.scheme, &self.strategy))
                 {
                     debug!(view = %updated_view, "constructed notarization, forwarding to voter");
+
+                    // Collect candidates while we still hold `round`.
```
patrick-ogrady: Pretty sure we can do this below?

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -166,9 +175,63 @@ impl<
...
+                    peers,
+                },
+            )
+            .await;
+    }
+
+    /// Returns missing voters that have been active recently enough to justify
+    /// a targeted forward.
+    fn active_missing_voters(
+        &self,
```
patrick-ogrady: remove this shortcut

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -166,9 +175,63 @@ impl<
...
+                    .take(skip_timeout)
+                    .any(|(_, round)| round.is_active(participant))
+            })
+            .collect()
+    }
+
+    /// Broadcasts a block forward for a notarized proposal to all active
+    /// missing peers.
+    async fn forward_notarization(&mut self, proposal: Proposal<D>, missing: Vec<Participant>) {
+        if self.forwarding.is_disabled() || missing.is_empty() {
```
patrick-ogrady: we would've already done this check?

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -342,12 +405,19 @@ impl<
...
                         // Store and forward to voter
-                        work.entry(view)
-                            .or_insert_with(|| self.new_round())
-                            .set_notarization(notarization.clone());
+                        let round = work.entry(view).or_insert_with(|| self.new_round());
+                        round.set_notarization(notarization.clone());
+                        let forward_candidates = self.forwarding.is_enabled().then(|| {
+                            let candidates = round.missing_voters(&notarization.proposal);
+                            (notarization.proposal.clone(), candidates)
+                        });
```
patrick-ogrady: do after recovered?

consensus/src/lib.rs
patrick-ogrady: or `Policy`?

consensus/src/marshal/coding/marshaled.rs
```diff
@@ -931,35 +931,38 @@ where
...
-        if block.commitment() != commitment {
-            warn!(
-                round = %round,
-                commitment = %block.commitment(),
-                height = %block.height(),
-                "skipping requested broadcast of block with mismatched commitment"
-            );
-            return;
+    type PublicKey = <Z::Scheme as CertificateScheme>::PublicKey;
+    type Dissemination = Dissemination<Self::PublicKey>;
```
patrick-ogrady: Can we make this the unit type (and just do propose logic) -> would require making simplex generic over dissemination -> not possible?

consensus/src/marshal/coding/marshaled.rs
```diff
@@ -931,35 +931,38 @@ where
...
+                debug!(
+                    round = %round,
+                    commitment = %block.commitment(),
+                    height = %block.height(),
+                    "requested broadcast of built block"
+                );
+                self.shards.proposed(round, block).await;
+            }
+            Dissemination::Forward { .. } => {
+                // Coding variant does not support targeted forwarding;
```
patrick-ogrady, line 962: This now makes sense to retain for cases like "PhasedSchemes" where the leader payload can be useful to forward?

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -166,9 +175,63 @@ impl<
...
+    fn forwarding_peers(&self, missing: &[Participant]) -> Option<Vec<S::PublicKey>> {
+        let peers: Vec<_> = missing
+            .iter()
+            .filter_map(|&p| self.participants.key(p).cloned())
+            .collect();
+        (!peers.is_empty()).then_some(peers)
+    }
+
+    /// Broadcasts a targeted block forward to the selected peers.
+    async fn broadcast_forward(&mut self, proposal: Proposal<D>, peers: Vec<S::PublicKey>) {
```
patrick-ogrady: I think we can remove this helper function?

consensus/src/lib.rs
patrick-ogrady: strategy already used by parallelism, we should avoid that.

consensus/src/marshal/coding/marshaled.rs
patrick-ogrady: Added TODO: #3389

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -342,12 +393,20 @@ impl<
...
+                            (notarization.proposal.clone(), participants)
+                        });
                         voter
                             .recovered(Certificate::Notarization(notarization))
                             .await;
+
+                        // Forward proposal to missing voters
+                        if let Some((proposal, participants)) = missing_voters {
+                            self.forward_proposal(proposal, participants).await;
+                        }
```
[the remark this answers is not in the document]
patrick-ogrady: This is the correct behavior for this mode (if we haven't seen any votes yet, then we'd send to all)!

patrick-ogrady:
## Remaining
- [x] Wait until certify returns/new view boundary to send anything (avoid sending junk)
- [x] Re-add leader only policy

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -244,36 +333,27 @@ impl<
...
                     }
                     response.send_lossy(timeout_reason);
 
+                    // Forward the proposal, if enabled and we have something to forward
+                    if let Some(proposal) =
+                        forwardable_proposal.filter(|_| self.forwarding.is_enabled())
+                    {
+                        let round = work.entry(proposal.view()).or_insert_with(|| self.new_round());
+                        let participants = self.forward_targets(round, &proposal, leader);
+                        self.forward_proposal(proposal, participants).await;
```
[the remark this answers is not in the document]
patrick-ogrady: This isn't an issue but cleaned it up.

consensus/src/marshal/core/mailbox.rs
```diff
@@ -92,6 +92,15 @@ pub(crate) enum Message<S: Scheme, V: Variant> {
         /// The block to broadcast.
         block: V::Block,
     },
+    /// A request to forward a block to a set of peers.
+    Forwarded {
```
BrendanChou: While I guess this is consistent with `Proposed` it seems strange that it's not called `Forward` instead since it hasn't been forwarded yet

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -342,9 +423,8 @@ impl<
                         }
 
                         // Store and forward to voter
-                        work.entry(view)
-                            .or_insert_with(|| self.new_round())
-                            .set_notarization(notarization.clone());
```
clabby, line 347: We don’t need `round` later in the scope - let’s remove this diff?

consensus/src/simplex/config.rs
```diff
@@ -12,8 +12,34 @@ use commonware_parallel::Strategy;
...
+/// Controls whether and how the engine proactively forwards certified blocks
+/// when entering the next view.
+///
+/// Forwarding is a best-effort liveness aid: when enabled, the batcher
+/// broadcasts only after we locally certify a proposal and enter the next
+/// view, avoiding sends for proposals that never pass certification.
+#[derive(Debug, Clone, Copy, Default)]
+pub enum ForwardingPolicy {
+    /// Do nothing when a certified proposal becomes eligible for forwarding.
+    #[default]
```
clabby: This was my addition, but I don’t think adding a default variant is wise. Let’s remove and explicitly name `Disabled`?

consensus/src/marshal/standard/inline.rs
```diff
@@ -23,6 +23,7 @@
 //! Because verification is completed inline, the default
 //! [`CertifiableAutomaton::certify`] behavior (always `true`) is sufficient: no
 //! additional deferred verification state must be awaited at certify time.
+//! `certify` only waits for block availability in marshal.
```
clabby: The wording could be improved here a bit - “always true” isn’t exactly what we’re doing anymore. Maybe:

> Because verification is completed inline, certify must only wait for data availability.

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -326,6 +406,7 @@ impl<
                 ) {
                     continue;
                 }
+                self.record_activity(&sender, view);
```
clabby, line 414: I see that this isn’t intended to be resilient to malicious validators setting their activity status as permanent (and this should be fine), though what’s stopping us from getting a stronger guarantee by waiting to record activity until we’ve verified the certificate they sent us?

patrick-ogrady: As discussed offline, we only verify 1 cert per round per type and only verify necessary votes. Verifying everything would add a lot of resource usage (and think we are better off avoiding that). 

This mechanism isn't meant to be byzantine-robust, so I don't see any issue with this (and am willing to do this to unblock "explicit" voting).

consensus/src/marshal/standard/inline.rs
```diff
@@ -411,6 +419,50 @@ where
...
+                        Err(_) => {
+                            debug!(
+                                ?digest,
+                                reason = "failed to fetch block for certification",
+                                "skipping certification"
+                            );
+                        }
+                    },
+                }
+            });
```
BrendanChou, line 460: I'm a little surprised there isn't a more elegant way to "pipe through" or "fuse" an rx/tx to each other

consensus/src/marshal/standard/inline.rs
```diff
@@ -455,6 +515,9 @@ where
 
     /// Forwards consensus activity to the wrapped application reporter.
     async fn report(&mut self, update: Self::Activity) {
+        if let Update::Tip(round, _, _) = &update {
+            self.available_blocks.lock().retain(|(r, _)| r > round);
```
BrendanChou: can't quite trace the logic this but should be `r >= round` ? Since if you get a new finalized tip, you might want to retain that block to forward it on?

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -166,9 +179,84 @@ impl<
...
+    /// Returns true if the participant has sent a recent message.
+    fn is_active(
+        &self,
+        work: &BTreeMap<View, Round<S, B, D, Re>>,
+        view: View,
+        participant: Participant,
+    ) -> bool {
+        if work.len() < self.skip_timeout.get() as usize {
+            return true;
+        }
```
BrendanChou, line 208: Nit: might be nice to add a small comment here what this is doing and why

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -166,9 +179,84 @@ impl<
...
+        let seen_view = self.latest_seen[usize::from(participant)];
+        view.get().saturating_sub(seen_view.get()) < self.skip_timeout.get()
+    }
+
+    /// Resolves the public keys of `missing` participants for a targeted
+    /// block forward. Returns `None` when no peers need the block.
+    fn resolve_peers(&self, missing: &[Participant]) -> Option<Vec<S::PublicKey>> {
```
BrendanChou: nits:
- `resolve_` name is a bit too generic I think
- don't love that empty vec just becomes `None`, seems the return type could simply be `Vec` without the `Option`

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -244,36 +333,27 @@ impl<
...
-                                None
-                            } else {
-                                Some(TimeoutReason::Inactivity)
-                            }
-                        } else {
-                            None
-                        }
+                        // If we are not the leader, we should timeout if we haven't
+                        // been active recently.
+                        None
```
BrendanChou: - This comment doesn't seem quite right anymore (?)
- I don't love calling an `else` function after an `else if !a && !b`, it can read confusingly about what cases the else covers. It might be better to do:

```
else if am_leader {
   None
} else {
  // I am not the leader
  if self.is_active(...) {
    None
  } else {
    Some(...)
  }
}

File: consensus/src/simplex/actors/batcher/actor.rs (hunk 38 above)
Comment (clabby) [reply consensus/src/simplex/actors/batcher/actor.rs]: The idea was to not allocate an empty `Vec`. Certainly a micro-optimization, though given the frequency that this is called it felt like the right thing to do.

File: consensus/src/marshal/standard/inline.rs (hunk 36 above)
Comment (patrick-ogrady) [reply consensus/src/marshal/standard/inline.rs]: By the time we get this report from marshal, we would've long-since cancelled whatever certify call was outstanding to the application.

File: consensus/src/marshal/standard/inline.rs (hunk 35 above)
Comment (patrick-ogrady) [reply consensus/src/marshal/standard/inline.rs:460]: Yeah, this pattern is shared by a few of the other applications. Made it into a helper for inline at least.

File: consensus/src/simplex/actors/batcher/actor.rs (hunk 39 above)
Comment (patrick-ogrady) [reply consensus/src/simplex/actors/batcher/actor.rs]: Flattened this.

File: consensus/src/marshal/standard/inline.rs [hunk 40]
```diff
@@ -411,6 +434,36 @@ where
...
+            .with_label("inline_certify")
+            .with_attribute("round", round)
+            .spawn(move |_| async move {
+                if await_block_subscription(&mut tx, block_rx, &digest, "certification")
+                    .await
+                    .is_some()
+                {
+                    tx.send_lossy(true);
+                }
+            });
```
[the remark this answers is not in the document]
patrick-ogrady: Marshal does not fail transitively. Any failure here would be a fatal error or shutdown.

PR #3350, opened 2026-03-09
[Storage] Consolidate QMDB traits

### LINKED ISSUE #2560: [storage/qmdb] Fix `CleanStore` for Current variants.

danlaine:
## Summary

Consolidate QMDB traits by removing `kv::Gettable`, `kv::Updatable`, `kv::Deletable`, `store::LogStore`, `store::MerkleizedStore`, and `store::PrunableStore`. All of their functionality is absorbed into `DbAny`, which becomes the single trait for polymorphic QMDB access.

None of these traits were used polymorphically outside of `DbAny`'s supertrait bounds -- every impl was pure delegation to an identically-named inherent method. Removing them eliminates ~600 lines of boilerplate across the crate.

Additionally, proof methods are split out of `DbAny` into a new `Provable` trait implemented only by Any variants, resolving #2560.

### Deleted modules and traits
- **`kv/mod.rs`**: Entire module deleted. `Gettable`, `Updatable`, `Deletable` traits and their test helpers (`assert_send`, `assert_gettable`, `assert_updatable`, `assert_deletable`, `test_key`). This module was stability level beta, so this PR is labeled `breaking-api`.
- **`store::LogStore`**, **`store::MerkleizedStore`**, **`store::PrunableStore`** traits and their test helpers (`assert_log_store`, `assert_prunable_store`, `assert_merkleized_store`).

### Expanded `DbAny` and new `Provable` trait (in `any/traits.rs`)
`DbAny` now directly defines everything that was previously spread across 4 supertraits, except proof generation:
- Associated types: `Key`, `Value`, `Digest`
- Read methods: `get()`, `root()`, `bounds()`, `size()`, `get_metadata()`
- Pruning methods: `prune()`, `inactivity_floor_loc()`

Proof methods (`proof()`, `historical_proof()`) and the `Operation` associated type are split into a new `Provable` trait, implemented only by Any variants. This resolves #2560: Current variants have a different proof structure (grafted MMR + activity bitmap) that is incompatible with the ops-level MMR proofs, so exposing proof generation on the shared trait was a broken abstraction. No existing test ever called proof methods on a Current variant through `DbAny`.

### `impl_db_any!` and `impl_provable!` delegation macros
`impl_db_any!` generates `DbAny` impls by forwarding each method to the identically-named inherent method. It uses a single invocation syntax (the previous dual-mode `proof_via: self` / `proof_via: self.any` mechanism was removed along with the proof methods).

`impl_provable!` generates `Provable` impls for Any variants only, delegating `historical_proof()` to the inherent method.

Because both macros return `impl Future<...> + Send` for each method, the compiler is forced to verify that the concrete inherent method's future is `Send` at the point of macro expansion. This structurally enforces the `Send` guarantee for all `DbAny` and `Provable` types.

### Generic `get()` on `any::db::Db`
Added a `get()` method on the generic `Db<E, C, I, H, U>` using the `Update` trait's `key()`/`value()` methods. This is needed because `any::batch.rs` calls `self.db.get(key)` on the generic `Db` type -- previously it used a `kv::Gettable` where-clause for this. The ordered and unordered variants previously had their own `get()` that delegated to `get_with_loc()`; the new version inlines the same logic at the generic level.

### Cleaned up `is_send` test assertions
- Removed `is_send` checks from all `DbAny` types (current/ordered/fixed, current/ordered/variable, current/unordered/fixed, current/unordered/variable, immutable). These are now redundant because `impl_db_any!` forces the `Send` check at compile time.
- Kept `is_send` checks for non-`DbAny` inherent methods that aren't covered by any trait (`get_all`, `get_with_loc`, `get_span`, `batch.merkleize`).
- Kept `is_send` checks for types that don't implement `DbAny` at all (`cache`, `ordinal`, `freezer`, `store::Db`, `keyless`).
- Replaced shared `assert_send`/`assert_gettable` helpers with local `fn is_send<T: Send>(_: T) {}` declarations.
- Replaced shared `test_key` helper with local copies in `archive` and `freezer` test modules.

### DIFF
--- a/storage/src/qmdb/any/traits.rs
+++ b/storage/src/qmdb/any/traits.rs
@@ -216,16 +216,6 @@ macro_rules! impl_db_any {
 pub(crate) use impl_db_any;
 
 /// Implements [`Provable`] by delegating each method to an identically-named inherent method.
-///
-/// # Syntax
-///
-/// ```ignore
-/// impl_provable! {
-///     [$($generics)*] TypeName
-///     where { $where_clause }
-///     Operation = ..
-/// }
-/// ```
 macro_rules! impl_provable {
     (
         [$($gen:tt)*] $ty:ty


storage/src/qmdb/current/db.rs
```diff
@@ -110,6 +109,12 @@ where
         self.any.get_metadata().await
     }
 
+    /// Return [start, end) where `start` and `end - 1` are the Locations of the oldest and newest
```
roberto-bayardo, line 112: nit: same comment as for keyless/mod.rs.

storage/src/qmdb/keyless/mod.rs
```diff
@@ -153,6 +148,13 @@ impl<E: Storage + Clock + Metrics, V: VariableValue, H: Hasher> Keyless<E, V, H>
         self.last_commit_loc
     }
 
+    /// Return [start, end) where `start` and `end - 1` are the Locations of the oldest and newest
```
roberto-bayardo, line 151: nit: Realize this was copied from existing code but I'd just say [start,end) is the range of retained operations. Current description, despite being wordy, doesn't capture the "empty range" case (start == end).

danlaine: will update this comment (and the existing copies of it) in a follow up PR

PR #3352, opened 2026-03-09
[consensus/simplex] Stable-Leader: Allow leaders to serve for a term of consecutive views

### LINKED ISSUE #3348: [consensus::simplex] Stable-Leader: Allow for leaders to serve for a "term" of consecutive views.
In order to get stable leaders in simplex, we introduce the concept of a “term.” A term is a number of contiguous views in which the leader is stable. If there is ever a nullification formed in a term, participants can skip the rest of the term. That is, a nullification serves as a nullification for the view that it is formed in, as well as all nullifications until the end of the term inclusive.

For safety, this means that nodes must not vote to finalize any block in a term if they have voted to nullify a previous block in that term and have seen no finalization certificate greater than the greatest view that they have voted to nullify.

### Implementation Details

- `term_length` should be defined as a non-zero configuration option in Simplex.
- Validators may find messages “interesting” not only if they are in near-future views, but also in views at-or-near the start of the next term.
- If using a VRF for leader election, then the VRF signatures on nullifications should sign for the final view in the term. Note that this can lead to an easier attack on the VRF over the course of a term, allowing people to piece together the VRF signature before the end of a term. Another alternative is to establish the next leader from the VRF signature of the first slot in the preceding term.
- While voting nullify is typically due to timeout, for longer terms, we could also consider allowing the application to issue nullify votes for any other behavior. For example, if the leader is censoring transactions, a validator could issue a nullify vote (either instead-of or in-addition-to a notarize vote). In this way, nullification also serves as a request to rotate the leader since a stable leader has much more power within a term, especially when the term is long.

### LINKED ISSUE #3404: [consensus/simplex] Optimistic Validation
### Algorithm

Optimistic validation can be allowed by allowing validators to vote `notarize` for a proposal before seeing a `notarization` certificate for its parent if-and-only-if they have voted to `notarize` all ancestors that do not yet have a `notarization` certificate. This means that they must have validated each proposal.

Continuing to follow the regular rules of Simplex, this seems safe to do. For the purposes of timeout, consider that you have entered a view only when you have seen a certificate for the previous view.

Note that due to byzantine nodes, this means that there may become “indirectly notarized” blocks. That is, blocks for which there will never exist a notarization certificate. They are instead notarized indirectly by one of their descendants. Once indirectly notarized, they should be considered notarized by honest validators and such validators should vote to `finalize` these blocks.

### Other Notes

- While it should go without saying, similarly to vanilla Simplex, it is important that validators do not vote to `notarize` more than one proposal per view, even if they may later discover that their original `notarize` vote for a view is no longer finalizable. This could result in safety errors. Instead, if you `notarize` a chain of views and discover that they are no longer finalizable, you may instantly vote to `nullify` all such views. This can help the network nullify-in-parallel such views faster than waiting for consecutive timeouts in each of those views.
- In a real-world scenario, there should perhaps be a bound on the chain height of proposals which a node is willing to `notarize` above the highest tip that has a `notarization` certificate. In a rotating-leader scenario, because it takes a network hop to get another leaders proposal, it can be hard to get a chain of more than 2 or 3 going (unless consecutive leaders happen to be very close geographically). For stable leaders, this number can get much higher, and is generally considered okay since a single nullification would end their term anyway.
- It may not seem necessary at first to vote to notarize every block. For example, if a validator sees proposal 2 then proposal 1 and verifies both, by notarizing proposal 2, to its peers, it has implicitly notarized proposal 1 as well. However, in practice it is important to generate `notarization` certificates as these are broadcast all-to-all between honest nodes to ensure that all honest nodes have a consistent view of certificates within `δ` time. Without such certificates, keeping in sync remains harder.
- In practice, optimistic validation may not help the chain to run faster unless done in a intra-leader scenario anyway, so we should probably just make it intra-leader (not at views at the start of terms)

### LINKED ISSUE #3405: [consensus/simplex] Optimistic Proposals
### Introduction

We can pipeline proposals by having the proposal for view `v` be disseminated in-parallel with the votes for view `v-1`. In practice, this is done by having the leader in view `v` broadcast a proposal once they have both a certificate for view `v-2` and the leader’s proposal for view `v-1`.

This allows all validators to receive the proposal for view `v` at the same time as the certificate for view `v-1`. In the happy path, this decreases the view latency to `δ`.

I specify “inter-leader” because they work fundamentally differently than stable-leader optimistic proposals. This is because you are making a trust assumption of the previous leader. If the previous leader forwards you a proposal that does not end up notarized, then your proposal built on top will also not be notarized.

### Implementation

Interestingly, this implementation seems to be relatively straightforward. Only the leader logic must change. That is, the leader identity in `v` must be knowable from having a certificate in `v-2` and a proposal in `v-1`. This means that a VRF from a certificate in `v-1` would be insufficient.

The main logic change is: instead of proposing when entering view `v`, leaders must be asked to propose in view `v-1` when they have a valid proposal from the leader in that view. In the case that they never receive a proposal directly, they can propose as soon as they enter view `v`.

As long as validators can cache votes for future views, the logic for validators does not have to change at all, including timeouts or view changes. They simply enjoy having more timely proposals already distributed to them.

BrendanChou:
Fixes #3348

Adds stable leaders to `simplex`: a leader now serves for a **term** of consecutive views (`TermLength`, default 1 = previous behavior) instead of being re-elected every view. Views within a term share the same leader; the elector only runs at term boundaries. Term anchors keep view→leader mapping consistent across restarts, epoch changes, and mid-term floors (joiners seed from the floor's term anchor, and the resolver re-fetches the term tail when its floor lands mid-term).

Timeout behavior is unchanged. One consequence worth noting for chaos-style tests: a restarted node that replayed its own proposal from the journal waits out the full `certification_timeout` before nullifying, so recovery from a full-cluster outage requires uptime longer than `certification_timeout`. The glue chaos test parameters are documented and sized accordingly.

Follow-up (separate issue): engine-level epoch-transition coverage at `term_length > 1`.


consensus/src/simplex/mod.rs
```diff
@@ -31,15 +31,17 @@
...
 //!   * If can't propose container in view `v` because missing notarization/nullification for a
 //!     previous view `v_m`, request `v_m`
 //!
 //! Upon receiving first `notarize(c,v)` from `l`:
 //! * Cancel `t_l`
 //! * If the container's parent `c_parent` is finalized (or both notarized and certified) at `v_parent`
-//!   and we have nullifications for all views between `v` and `v_parent`, verify `c` and broadcast `notarize(c,v)`
+//!   and we have required nullifications covering the skipped views between `v_parent` and `v`
+//!   (a nullification covers the rest of its term),
+//!   verify `c` and broadcast `notarize(c,v)`
```
patrick-ogrady: put this on 43?

consensus/src/simplex/mod.rs
```diff
@@ -116,25 +119,34 @@
 //! * Upon seeing `notarization(c,v)`, instead of moving to the view `v+1` immediately, request certification from
 //!   the application (see [Certification](#certification)). Only move to view `v+1` and broadcast `finalize(c,v)`
 //!   if certification succeeds, otherwise broadcast `nullify(v)` and refuse to build upon `c`.
+//! * With stable leaders (`term_length > 1`), if a participant has voted `nullify(v_n)` in a term, it will not vote
+//!   `finalize(c,v_f)` for later views in that same term.
+//! * With stable leaders, `term_stop_notarize_on_nullify` can additionally prevent voting `notarize(c,v_n)` for later
```
patrick-ogrady: why would we ever want to keep voting notarize here?

consensus/src/simplex/mod.rs
```diff
@@ -374,8 +389,16 @@ cfg_if::cfg_if! {
             if pending < min_active(activity_timeout, last_finalized) {
                 return false;
             }
-            if !allow_future && pending > current.next() {
-                return false;
+            // If we don't allow unbounded future views, we still find two future views interesting:
+            // - the next view
+            // - the first view of the next term (it may be the next view if the current view is nullified)
```
patrick-ogrady: I wonder if this view gapping can cause issues in our memory? I suppose that was already considered.

consensus/src/simplex/elector.rs
```diff
@@ -193,8 +220,16 @@ where
...
 
-    fn build(self, participants: &Set<P>) -> RandomElector<bls12381_threshold_vrf::Scheme<P, V>> {
+    fn build(
+        self,
+        participants: &Set<P>,
+        term_length: NonZeroU64,
+    ) -> RandomElector<bls12381_threshold_vrf::Scheme<P, V>> {
         assert!(!participants.is_empty(), "no participants");
+        assert!(
+            term_length.get() == 1,
```
patrick-ogrady: assert_eq?

consensus/src/simplex/config.rs
```diff
@@ -154,29 +155,53 @@ where
...
+    pub term_stop_notarize_on_nullify: bool,
+
+    /// Maximum time an entered view may remain unfinalized before we allow a
+    /// local nullify vote for the current view.
+    ///
+    /// This timeout must be greater than the certification timeout so normal
+    /// proposal and certification paths have a chance to complete before
+    /// term-level abandonment. When `term_length > 1`, this effectively tracks
+    /// the oldest entered, unfinalized view in the current term.
+    pub finalization_timeout: Duration,
```
patrick-ogrady: Can you elaborate on why this is necessary?

consensus/src/simplex/elector.rs
```diff
@@ -53,14 +54,19 @@ pub trait Config<S: Scheme>: Clone + Default + Send + 'static {
...
+    /// each view to its term before selecting a leader. Electors that
+    /// do not support stable leaders should panic when `term_length > 1`.
     ///
     /// # Panics
     ///
-    /// Implementations should panic if `participants` is empty.
-    fn build(self, participants: &Set<S::PublicKey>) -> Self::Elector;
+    /// Implementations should panic if `participants` is empty, or if they do
+    /// not support `term_length`.
+    fn build(self, participants: &Set<S::PublicKey>, term_length: NonZeroU64) -> Self::Elector;
```
patrick-ogrady: I think we can just call `term_length` -> `term` everywhere. To avoid confusion, we could make it a type like we do for `View`.

patrick-ogrady (commented):
initial scan looks pretty decent

consensus/src/simplex/mod.rs
BrendanChou: The main reason is a honest "blip" in a very long term. For example if the stable leader is meant to serve an entire 2-week term, then it's likely that honest nodes or an honest leader would have a "blip" at some point. Even for say updating an honest validator, when it comes online and is not attached to enough nodes, it will likely vote to `nullify` the next view. In such a case it would be nice if it can "self-heal" and get back to validating normally, rather than nullifying every view which causes performance degradation on the network

consensus/src/simplex/elector.rs
BrendanChou: I think `term_length` is slightly clearer. A new type is a good idea though, I will add that and see if it helps. Re-using the `ViewDelta` type could be useful but I think it's not quite correct and also allows the zero value

consensus/src/simplex/mod.rs
BrendanChou: Memory stays bounded. With `allow_unbounded_future = false`, `interesting()` admits at most two future views at any moment: `current.next()` and `current.next_term_start()` (which are the same view when `term_length = 1`).

The gap views "inside" a skipped term are never interesting, so no memory is ever allocated for them. An adversary can't use them to force allocations. Net effect vs. `term_length = 1` (and vs existing behavior) is at most one extra tracked round at any time (the next term start).

 Growth of work between finalized and current during a long unfinalized run is pre-existing behavior (same as `term_length = 1`), and `finalization_timeout` now time-bounds how long a term can run unfinalized anyway.

consensus/src/simplex/config.rs
BrendanChou: With stable leaders, a Byzantine leader can keep every per-view timer satisfied while never letting finality happen: each view notarizes and certifies, so `leader_timeout` and `certification_timeout` never fire and we keep advancing views, but the leader prevents finalization certificates from forming. Without this timeout, that stalls finality for the entire term with nothing ever triggering.

With `term_length = 1` this was never needed because rotation implicitly bounds the stall to a single view (the next view has a new leader). With long terms there is no rotation to save us, so `finalization_timeout` tracks the oldest entered-but-unfinalized view in the current term and nullifies the current view when it expires, which (via the nullification covering the rest of the term) skips the remainder of the term and evicts the leader.

It must exceed `certification_timeout` so the normal propose/notarize/certify path gets a full chance to complete before we start abandoning the term.

Also folded this rationale into the doc comment.

consensus/src/simplex/mod.rs
BrendanChou: Thought about this more and I think the right answer is to remove this flag entirely and make the same-term finalize rule self-healing instead. Proposal:

- Always vote notarize on valid proposals (no flag).
- A finalize vote for view `v` is permitted iff our highest same-term nullify vote is `<=` the highest same-term finalization we have observed (and we didn't nullify `v` itself), evaluated when constructing the vote.
- Nullify votes are never cascaded across views: each view's nullify is triggered by its own timeout or failure (existing immediate-nullify paths like leader nullify, verification/certification failure, and skip timeout are unchanged).

**Why it's safe**: the invariant skipped-view validation relies on is that a finalization at `v_f` implies no nullification can ever form at any `v <= v_f` in the same term. A finalization at `v*` has at least `f+1` honest finalize voters, each of whom (by this same rule) never voted nullify at any view `<= v*` and never will (views advance monotonically), so at most `2f` nullify votes can ever exist for those views: a nullification there is permanently impossible, which makes our own earlier nullify votes provably inert. The rule survives its own relaxation by induction on finalization views within the term: an honest node that voted both `nullify(v)` and `finalize(v_f)` must have observed an earlier finalization `>= v`, and the induction hypothesis rules out the nullification at `<= v`.

**Why this answers the question here**: there is nothing left to configure. Short terms: a node that nullified stays finalize-blocked unless the term demonstrably recovers, and the term dies quickly otherwise. Long terms: an honest blip heals one view later, because peers broadcast `finalize(v)` when they certify `v`, so `finalization(v)` arrives about one message delay after we enter `v+1`, well before our next finalize vote is constructed. If the leader is actually bad, no finalizations form, the unlock never fires, and we keep timing out toward nullifying the term.

Worst case is exactly the current behavior (finalize-blocked for the rest of the term), so this is a strict improvement over both settings of the flag. Cost: the highest-nullify and highest-observed-finalization watermarks become safety-critical state that must survive restart (same machinery as the existing same-term rule), plus a twins test for selective delivery of the unlocking finalization.

WDYT? If you're on board I'll make this change.

patrick-ogrady: Your point about degraded performance during a term resonated with me. I think this makes sense.

BrendanChou: Okay updated

consensus/src/simplex/elector.rs
```diff
@@ -114,7 +123,11 @@ impl<H: Hasher> RoundRobin<H> {
 impl<S: Scheme, H: Hasher> Config<S> for RoundRobin<H> {
     type Elector = RoundRobinElector<S>;
 
-    fn build(self, participants: &Set<S::PublicKey>) -> RoundRobinElector<S> {
+    fn build(
```
patrick-ogrady: dumb idea: what if the elector returned the term length (was configured with it)?

Then, it would be impossible to misconfigure random? Given term is inherent to electing, I think its actually a pretty reasonable home.

consensus/src/simplex/config.rs
```diff
@@ -154,29 +155,55 @@ where
...
+    /// finalization certificate forms. With a `term_length` of 1, leader
+    /// rotation bounds such a stall to a single view. With longer terms, this
+    /// timeout bounds it instead: it tracks the oldest entered, unfinalized
+    /// view in the current term and triggers a nullify vote for the current
+    /// view when it expires, skipping the rest of the term.
+    ///
+    /// This timeout must be greater than the certification timeout so normal
+    /// proposal and certification paths have a chance to complete before
+    /// term-level abandonment.
+    pub finalization_timeout: Duration,
```
patrick-ogrady: tricky but necessary...good catch

consensus/src/simplex/mod.rs
```diff
@@ -116,25 +118,67 @@
...
+//! least one honest participant that voted both `nullify(u)` and `finalize(c,v)`. If `u = v`,
+//! this is impossible outright: no honest participant votes both `nullify` and `finalize` in a
+//! single view. If `u < v`, the gate means that participant first observed a same-term
+//! finalization at some `v*` with `u <= v* < v`: at or above `u` because the gate requires
+//! covering its highest `nullify` vote, and below `v` because participants only vote `finalize`
+//! for views above their highest observed finalization. Applying this same argument to the
+//! finalization at `v*` shows, by induction, that no nullification can form at or below `v*`,
+//! contradicting the nullification at `u <= v*`.
+//!
+//! The gate also recovers from a `nullify` vote that never became a nullification (e.g., a
```
patrick-ogrady, line 182: I would say more explicitly that this prevents degradation during the rest of the term?

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -189,14 +254,27 @@ impl<E: Clock + CryptoRngCore + Metrics, S: Scheme<D>, L: ElectorConfig<S>, D: D
         min_active(self.activity_timeout, self.last_finalized)
     }
 
-    /// Returns whether `pending` is still relevant for progress, optionally allowing future views.
-    pub fn is_interesting(&self, pending: View, allow_future: bool) -> bool {
+    /// Returns whether a vote for `pending` is still relevant for progress.
+    pub fn is_interesting_vote(&self, pending: View) -> bool {
```
patrick-ogrady: `is_interesting_vote` + `is_interesting_certificate` is much better...nice

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -274,34 +353,76 @@ impl<E: Clock + CryptoRngCore + Metrics, S: Scheme<D>, L: ElectorConfig<S>, D: D
...
         let timeout_retry = self.timeout_retry;
-        let round = self.create_round(self.view);
-        round.next_timeout_deadline(now, timeout_retry)
+        let round_timeout = {
+            let round = self.create_round(self.view);
+            round
+                .next_timeout(now, timeout_retry)
+                .expect("current round must always have a timeout")
+        };
+        // Once the current view is retrying a nullify, retry cadence should be governed
```
patrick-ogrady, line 332: prefer \n before this

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -274,34 +353,76 @@ impl<E: Clock + CryptoRngCore + Metrics, S: Scheme<D>, L: ElectorConfig<S>, D: D
...
+        // by timeout_retry. An older expired same-term finalization deadline may still
+        // exist for the first unfinalized view in the term.
+        if matches!(round_timeout.1, TimeoutReason::Retry) {
+            return round_timeout;
+        }
+        // The finalization anchor only overrides a round timeout that has not
```
patrick-ogrady: prefer \n before this

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -274,34 +353,76 @@ impl<E: Clock + CryptoRngCore + Metrics, S: Scheme<D>, L: ElectorConfig<S>, D: D
...
+        // itself expired: when both are due, the round's reason (e.g. a latched
+        // LeaderNullify or InvalidProposal) is the more diagnostic label for
+        // the nullify metric.
+        self.next_finalization_timeout()
+            .filter(|&deadline| deadline <= round_timeout.0 && now < round_timeout.0)
+            .map(|deadline| (deadline, TimeoutReason::FinalizationTimeout))
+            .unwrap_or(round_timeout)
+    }
+
+    fn next_finalization_timeout(&mut self) -> Option<SystemTime> {
```
patrick-ogrady: add a function comment

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -274,34 +353,76 @@ impl<E: Clock + CryptoRngCore + Metrics, S: Scheme<D>, L: ElectorConfig<S>, D: D
...
+        let term_start = self.view.term_start(self.term_length);
+        let unfinalized_view = self.last_finalized.next().max(term_start);
+        // The oldest unfinalized view may never have been entered (e.g., we
```
patrick-ogrady, line 358: add a \n before?

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -310,29 +431,48 @@ impl<E: Clock + CryptoRngCore + Metrics, S: Scheme<D>, L: ElectorConfig<S>, D: D
...
-            Some(Certificate::Notarization(notarization))
-        } else {
-            warn!(%view, "entry certificate not found");
-            None
+        // Check if there was a finalization in the previous view
+        if let Some(finalization) = self.finalization(prev).cloned() {
+            return Some(Certificate::Finalization(finalization));
         }
+
+        // At a term boundary, prefer the highest nullification from the previous
```
patrick-ogrady, line 430: this is a tricky one...I suppose this increases the likelihood the first view in the next term after a nullification doesn't also timeout?

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -310,29 +431,48 @@ impl<E: Clock + CryptoRngCore + Metrics, S: Scheme<D>, L: ElectorConfig<S>, D: D
...
 
-    /// Returns the best certificate for `view` to help peers enter `view + 1`.
+    /// Returns the best certificate to help peers enter the current view.
     ///
     /// Finalization is strongest, then nullification, then notarization.
-    pub fn get_best_certificate(&self, view: View) -> Option<Certificate<S, D>> {
-        if view == GENESIS_VIEW {
+    ///
+    /// With stable leaders, if the current view follows a skipped term, prefer
+    /// a nullification from that term over a notarization at the previous view.
```
patrick-ogrady, line 410: I'm a bit concerned about fork alignment with these term gaps ... we are basically saying a nullification at `v1` is preferred to a notarization at `v2` (where `v2>v1`) and a nullification at `v1` ensures anything after `v1` won't be finalized (in the same term)?

So having the nullification at `v1` should still allow people to build on the same fork and just skip `v2`?

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -415,8 +561,29 @@ impl<E: Clock + CryptoRngCore + Metrics, S: Scheme<D>, L: ElectorConfig<S>, D: D
...
+    /// finalize vote unless an observed finalization covers it (see
+    /// [Same-Term Vote Safety](crate::simplex#same-term-vote-safety)).
     pub fn construct_finalize(&mut self, view: View) -> Option<Finalize<S, D>> {
+        // We don't need to finalize views that are already finalized.
+        if view <= self.last_finalized {
+            return None;
+        }
+
+        // Blocked by a same-term nullify vote unless an observed finalization
+        // covers it (proving that nullification can never form). The plain
```
patrick-ogrady, line 550: I believe it is possible for us to get into a state where we don't have a nullification but we'll never have enough to finalize (even if the leader is timely) due to transient failures?

What I have in mind are nullifies emitted by different participants in different views? I suppose that the finalization timeout will save us here as well?

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -649,10 +824,6 @@ impl<E: Clock + CryptoRngCore + Metrics, S: Scheme<D>, L: ElectorConfig<S>, D: D
     pub fn certified(&mut self, view: View, is_success: bool) -> Option<Notarization<S, D>> {
         let round = self.views.get_mut(&view)?;
         round.certified(is_success);
-        if is_success {
-            // Clear deadlines if the certification was successful
-            round.clear_deadlines();
```
patrick-ogrady, line 647: still reviewing but what is the motivation not to timeout view/leader timeout here anymore? just handled by something else?

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -720,32 +924,53 @@ impl<E: Clock + CryptoRngCore + Metrics, S: Scheme<D>, L: ElectorConfig<S>, D: D
...
+                self.genesis
+                    .as_ref()
+                    .expect("genesis must be set when finding parent with no certified views"),
+            ),
+        };
 
-            cursor = cursor.previous().expect("cursor must not wrap");
+        // If there are any missing nullifications, return an error.
+        // Any lower certified views would also result in an error.
+        if let Some(missing_view) = self.first_unnullified_view(candidate, view) {
```
patrick-ogrady, line 927: Pretty nice abstraction for this (this is pretty tricky)

consensus/src/simplex/actors/voter/round.rs
```diff
@@ -51,8 +51,11 @@ pub struct Round<S: Scheme, D: Digest> {
     proposal: ProposalSlot<D>,
     leader_deadline: Option<SystemTime>,
     certification_deadline: Option<SystemTime>,
-    timeout_retry: Option<SystemTime>,
-    timeout_reason: Option<TimeoutReason>,
+    finalization_deadline: Option<SystemTime>,
+    retry_deadline: Option<SystemTime>,
+    // First explicit timeout latched for this round (see latch_timeout).
```
patrick-ogrady, line 57: If we add an inline comment here, I think we should do for others to?

I find LLMs often do this and don't ensure their changes are consistently applied across a struct/constructor?

consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -945,6 +955,12 @@ impl<
             pin_mut!(stream);
             while let Some(artifact) = stream.next().await {
                 let (_, _, _, artifact) = artifact.expect("unable to replay journal");
+                // Dropping our own nullify votes at or below the floor is safe
```
patrick-ogrady, line 967: I'd move this just above the artifact.expect line above

consensus/src/simplex/actors/resolver/state.rs
```diff
@@ -166,7 +157,6 @@ impl<S: Scheme, D: Digest> State<S, D> {
                 });
             }
 
-            // Re-request any lower views this notarization had satisfied
```
patrick-ogrady: retain?

patrick-ogrady: (and the other ones here?

consensus/src/simplex/actors/resolver/state.rs
```diff
@@ -223,40 +230,43 @@ impl<S: Scheme, D: Digest> State<S, D> {
...
         self.satisfied_by.retain(|view, _| *view > floor);
         self.failed_views.retain(|view| *view > floor);
         Effect::RetainAbove(floor)
     }
 }
 
+/// Returns whether a nullification at `view` covers any view above `floor`.
+///
+/// A nullification covers the rest of its term, so it remains relevant while
+/// its term end is above the floor. Admission (`encounter_nullification`) and
```
patrick-ogrady: > Admission (`encounter_nullification`) and
/// retention (`prune`) must agree on this boundary.

Is anything done to ensure this?

consensus/src/simplex/actors/resolver/state.rs
```diff
@@ -103,14 +100,16 @@ impl<S: Scheme, D: Digest> State<S, D> {
         match certificate {
             Certificate::Nullification(nullification) => {
                 let view = nullification.view();
-                if self.encounter_view(view) {
+                if self.encounter_nullification(view) {
                     self.nullifications
                         .insert(view, Certificate::Nullification(nullification));
-                    effects.push(Effect::Remove(view));
+                    effects.push(Effect::RetainOutside {
```
patrick-ogrady, line 110: The idea here is to say "we have what we need for this range"?

And based on our other claims it is impossible for their to exist a finalization in this range?

patrick-ogrady (commented):
This is an impressively succinct implementation of such a complex change.

BrendanChou:
Addressed all comments

consensus/src/simplex/elector.rs
BrendanChou: Fixed in https://github.com/commonwarexyz/monorepo/pull/3352/changes/53b50a4e81bc54ca4c2d658e605349062451149b

I think it's worth keeping

patrick-ogrady:
I ran this through Fable for a few hours:

## Review verdict

The core design is sound. I ran a 46-agent review (11 area reviewers, adversarial verification of every finding, plus a completeness critic) over the ~5,200-line diff, and independently traced the crux mechanisms myself. The same-term finalize gate implementation (`voter/state.rs:575`, `round.rs:577`) matches the documented quorum-intersection induction argument, including the subtle parts: gate evidence is journaled and restored on replay, pruning provably retains all gate-relevant nullify votes, and the `view <= last_finalized` early-return turns out to be load-bearing (it stops a later-term finalization from reopening the gate for views a nullification covers). Term arithmetic in `types.rs` is mutually consistent across all callers. 29 findings survived adversarial verification — **no criticals** — and every one is now either fixed (staged), documented, or listed below for your call.

## Staged changes (11 files, +207/−44 — not committed)

**Bug fixes:**
- **`engine.rs:64`** — assert `term_length <= u32::MAX` at startup. Without a bound, `term_length = u64::MAX` is accepted silently and then panics *every* node at the first nullification (`add_nullification` → `next_term_start` overflow), and even a single unverified garbage vote 2+ views ahead reaches the same panic through `View::admits` in the batcher before signature verification.
- **`resolver/state.rs`** — restored the `fetch_floor` cursor that main had and the rewrite dropped. Without it, an anchor satisfied by a pending notarization was re-fetched and re-verified (full multi-sig) once per RTT until certification completed — one of the PR's own unit tests encoded this loop — and every handled certificate rescanned the whole floor-to-current gap (O(gap²) during deep catch-up). Also guarded the certification-failure re-fetches with a covering-nullification check. Updated the affected test and added `fetch_requests_each_anchor_once` as a regression pin.
- **`fuzz/src/invariants.rs`** — the safety oracle claimed independence from `View::covers` but delegated to `View::same_term`, the exact predicate `covers` is built on. It now restates term grouping with independent integer arithmetic, so a bug in the production predicates can't hide from the fuzzer.

**Doc corrections (these were wrong, not just thin):**
- **Forced Inclusion / Optimistic Finality (`mod.rs`)** — the "Therefore" step was unsound for `term_length > 1`: view `v` completing cleanly says nothing about nullify votes cast at *earlier* views of the term, all of which cover `v`. Both sections now state the property over the term prefix `[term_start(v), v]` (identical to the old per-view claim at L=1). Integrators reading "notarized + no timeout = speculatively final" would otherwise mis-assess finality under stable leaders.
- **Healing paragraph (`mod.rs`)** — "is instead finalized transitively" overpromised; a gate-blocked view can also be skipped entirely by a covering nullification, and the finalizing descendant is necessarily in a later term.
- **`TermLength` (`types.rs`)** — now documents that it is consensus-critical and *unauthenticated*: nothing in the protocol detects a mismatch, and mismatched nodes silently disagree on nullification coverage and vote safety. Also notes the memory trade-off of long terms (below).
- Spec now states the intra-term no-skip proposal rule the code enforces (`parent_payload`); `Elector::elect` documents that honest nodes may pass *different* certificates for the same round (why `Random` rejects L>1); `Nullification`'s "moves to the next view" doc updated; `Floor` documents its durability/monotonicity requirement (journal pruning trusts it); stale "still willing to broadcast finalize" comment in `round.rs` replaced; the batcher tip-drop comment now names the activity-report loss, with a matching bullet in the module docs.

**Test coverage:**
- **`test_twins_stable_leader`** — parameterized the twins campaign by term length (it was pinned to `TermLength::ONE`) and added a Byzantine-twins campaign at L=3 across both link profiles. This was the biggest coverage gap: the finalize gate had zero adversarial multi-node coverage. It passes (12.6s).
- Genesis edge-case rows for `admits`/`covers`/`same_term` tests; simplified the unreachable `checked_mul` expect in `term_start`; debug log when the gate withholds a finalize vote (it previously blocked silently, indistinguishable from a missing quorum).

## Real findings I deliberately did not change

These verified as genuine behavior changes, but your in-code comments indicate they're intentional trade-offs — reversing them is your call, not mine:

1. **Batcher drops votes at/below the finalized tip before reporting** (`batcher/actor.rs:547`, rated major). On main, votes were reported down to `finalized − activity_timeout` (the removed comment said this was deliberate, for the reporter). Now: slow-but-honest validators lose uptime credit for votes arriving just after finalization, and a Byzantine validator who equivocates *after* observing a finalization is never reported or blocked. This is an L=1 behavior change, contrary to the PR's stated intent — if any reward/slashing system consumes per-vote `Activity`, consider restoring a reporting-only window.
2. **L-fold unverified-vote window under stable leaders** (`batcher/actor.rs:553`, rated major). During a finalization stall, each nullification advances `current` by up to L views; every skipped view admits one signed vote per validator, each allocating an O(n) `Round` before any signature check, unblockable, retained until finalization resumes. Real but requires a concurrent stall and self-heals; I documented it on `TermLength`. Worth a mitigation (e.g., cap non-current rounds) before running large L in production — votes for covered views can't simply be dropped, since a notarization in a nullified term may still become a parent.
3. **`record_activity` precedes view filtering** — replaying anyone's ancient signed vote refreshes that validator's `skip_timeout` activity, letting a stalling leader suppress the fast-skip (damage bounded by `leader_timeout` per view). Your comment marks this as a non-Byzantine-resistant heuristic.
4. Nits, report-only: `Random::select_leader`'s view-1 fallback truncates to u32 before modulo (deterministic across nodes, harmless); 32-bit leader selection differs from main (yours is the more correct); `certification_timeout > leader_timeout` is now strict, so previously-valid equal configs panic at startup.

## Remaining follow-ups (not blocking — stable leaders aren't enabled by default)

- No engine-level test of the exact attack `finalization_timeout` exists for: a Byzantine stable leader that notarizes/certifies every view but withholds finalize votes (only state-level unit tests cover the timer). Needs a new Byzantine mock actor.
- A dedicated `Floor::Finalized`-at-mid-term-view seed test at L>1 (partially covered by `test_hailstorm_stable_leader`'s 10 restarts at L=3).
- The reshare orchestrator hardcodes `TermLength::ONE` — fine for now, but epoch-rotating deployments have no plumbing to enable stable leaders.

**Verification:** fast profile 806/806 (was 805 — the new resolver test), slow-profile simplex 216/216 pre-change plus both stable-leader slow tests post-change, fuzz-crate tests 4/4, and `just lint` green. Everything is staged and ready for review with `git diff --cached`.

***********

It recommended the attached.
[scratch.txt](https://github.com/user-attachments/files/29586042/scratch.txt)

consensus/src/simplex/actors/resolver/state.rs
```diff
@@ -222,41 +242,52 @@ impl<S: Scheme, D: Digest> State<S, D> {
...
+        let mut cursor = self.fetch_floor.max(self.floor_view().next());
+        while cursor < self.current_view && effects.len() < self.fetch_concurrent {
+            if self.covering_nullification(cursor).is_none() {
+                effects.push(Effect::Fetch {
+                    view: cursor,
+                    cause,
+                    reason,
+                });
+            }
+            cursor = cursor.next_term_start(self.term_length);
```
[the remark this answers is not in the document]
BrendanChou: False positive. The anchor at `current_view` is deferred, not skipped: the loop exits with `fetch_floor` sitting at the first unscanned anchor, and `current_view` only advances inside `handle()`, which unconditionally calls `fetch()` afterward — so the anchor is enqueued the moment the view passes it. A nullification keyed at `current_view` also cannot be usefully missing: `current_view` only reaches view `v` via a certificate at `v`, which is either that nullification itself or a notarization/finalization that already provides coverage at `v`. Backfill only ever needs coverage strictly below `current_view`. This matches the pre-existing exclusive bound (`View::range(start, self.current_view)`) on main.

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -331,30 +344,31 @@ where
...
-                            } else {
-                                // If we are not the leader and the leader isn't active, we should timeout
-                                (!self.is_active(&work, current.view, leader))
+                                true => None,
+                                // If we are not the leader, don't have a proposal
+                                // to forward, and the leader isn't active, we should timeout.
+                                false => (!has_forwardable_proposal && !self.is_active(&leader))
                                     .then_some(TimeoutReason::Inactivity)
-                            }
+                            },
```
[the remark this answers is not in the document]
BrendanChou: Acknowledged as a known, bounded race rather than a bug. Activity is a wall-clock heuristic that is inherently stale by more than the mailbox-queue window — the same leader message could just as well still be in a network socket, so reordering select arms cannot close the race. For the branch to fire, the leader must have been silent for a full `skip_timeout`, so the fast timeout is justified on all information available at decision time; the consequence is one early nullify, with safety unaffected and nodes that did observe the leader's activity not timing out. The `has_forwardable_proposal` guard suppresses the common case. The same ordering property existed with main's view-based `latest_seen` tracking. Documented in c61104691.

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -154,37 +163,43 @@ where
...
+            .current()
+            .checked_sub(self.skip_timeout)
+            .unwrap_or(SystemTime::UNIX_EPOCH);
+        let recent =
+            |activity: &Option<SystemTime>| activity.is_some_and(|activity| activity >= min_time);
+
+        // If there is not a quorum of recently active participants, then we "fail-open" since we
+        // know the network is not expected to be responsive.
+        let active = self.last_activity.iter().filter(|a| recent(a)).count();
+        if active < self.scheme.participants().quorum::<N3f1>() as usize {
```
patrick-ogrady: This is a much safer/saner way to do this I think.

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -328,30 +344,38 @@ where
                         round.set_span(span);
                         round.set_leader(leader);
 
+                        let has_forwardable_proposal =
```
patrick-ogrady: what does this have to do with it?

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -403,18 +431,15 @@ where
...
-                    view,
-                    true, // allow future
-                ) {
+                self.record_peer_activity(&sender);
+
+                // Ignore certificates below the highest finalized view since they aren't useful.
+                // Allow certificates from arbitrarily-future views since they advance our view.
+                if view <= finalized {
                     continue;
                 }
```
patrick-ogrady, line 463: do we still need to do the admits check here?

patrick-ogrady: I don't love breaking away from a single function check here (`interesting`)?

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -677,11 +711,12 @@ where
                 .instrument(span)
                 .await;
 
-                // Drop any rounds that are no longer interesting
-                while work.first_key_value().is_some_and(|(&view, _)| {
-                    !interesting(self.activity_timeout, finalized, current.view, view, false)
```
patrick-ogrady, line 682: is it safe to remove interesting here?

consensus/src/simplex/actors/resolver/state.rs
```diff
@@ -84,66 +92,69 @@ impl<S: Scheme, D: Digest> State<S, D> {
...
+                    effects.push(Effect::RetainOutside {
+                        start: view,
+                        end: view.term_end(self.term_length),
+                    });
                 }
             }
             Certificate::Notarization(notarization) => {
-                // Store as pending (waiting for certification result).
                 let view = notarization.view();
-                if self.encounter_view(view) {
```
patrick-ogrady, line 103: another case (like `interesting` where we unrolled standard logic)?

consensus/src/simplex/metrics.rs
```diff
@@ -27,6 +29,7 @@ impl TimeoutReason {
     /// metric series share the same string.
     pub const fn as_str(self) -> &'static str {
         match self {
+            Self::Retry => "Retry",
```
patrick-ogrady, line 32: Need to see what this is about...

consensus/src/simplex/elector.rs
```diff
@@ -106,9 +175,33 @@ impl<H: Hasher> RoundRobin<H> {
...
+    ///
+    /// The term length is consensus-critical: every participant must configure
+    /// the same value (see [`TermLength`]). The timeout is local policy.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `term_length` is 1 (single-view terms need no timeout and are
+    /// the default).
+    pub const fn with_term(mut self, term_length: TermLength, stall_timeout: Duration) -> Self {
+        assert!(
```
patrick-ogrady: It isn't possible to construct a zero term length?

consensus/src/simplex/actors/voter/round.rs
```diff
@@ -366,25 +386,58 @@ impl<S: Scheme, D: Digest> Round<S, D> {
             return None;
         }
         let retry = replace(&mut self.broadcast_nullify, true);
-        self.clear_deadlines();
-        self.set_timeout_retry(None);
+        self.leader_deadline = None;
```
patrick-ogrady, line 389: another function removed (`clear_deadlines`)?

patrick-ogrady (commented):
almost through this

consensus/src/simplex/actors/batcher/actor.rs
BrendanChou: Dug into this more and concluded it wasn't pulling its weight: removed in 32a9d3978 in order to simplify the code.

What it did: The next leader's notarize (i.e. proposal) can be buffered before we enter the view, so if we enter the view after the leader, we could have a  proposal from the leader already while also not having a "recent" (by-time) vote from the leader. On `main`, this is impossible to think the leader is offline if you have its proposal since `main` measures activity by votes in recent views, rather than by time.

HOWEVER, this is really only trying to salvage a very edge case: hearing only silence from the leader for `skip_timeout` time really does mean the leader is offline (an alive leader in an unfinalized view rebroadcasts nullifies every `timeout_retry`, which is asserted to be shorter), so this only salvaged a view from an already-dead leader. Especially with stable leaders, evicting fast is preferable to salvaging one view of a dead leader's term.

consensus/src/simplex/metrics.rs
BrendanChou: On main only the first nullify got a reason; now `Round::next_timeout` returns `(deadline, reason)` pairs, so the nullify-rebroadcast timer needs its own label. `Retry` also matters in `state.rs::next_timeout`: once a round is retrying, retry cadence takes precedence over an already-expired stall deadline.

(`as_str` itself predates this PR — added in #4034 so trace spans get a `&'static str` matching the metric label. This PR only adds the `Retry` arm.)

consensus/src/simplex/elector.rs
BrendanChou: Zero isn't constructible (`TermLength::new` takes `NonZeroU32`), so this assert only guarded 1 — and it duplicated the engine's check, which is the authoritative one (any `Elector` impl could report `Stable { length: 1 }`, not just `RoundRobin`). Removed the redundant assert in 31cda2887.

consensus/src/simplex/actors/voter/round.rs
BrendanChou: `clear_deadlines` had three call sites on main, and they now need three different behaviors:

- `construct_nullify`: clears leader/certification/retry deadlines and the latched timeout, but must NOT clear `stall_deadline` (it anchors term-level stall protection and survives nullify broadcasts).
- `add_finalization`: clearing is redundant — `next_timeout` and `stall_deadline()` already return early once finalized.
- `add_nullification`: clearing would break `State::next_timeout`'s invariant that the current round always has a pending timeout; the nullification stops the timers by advancing the view instead.

So the helper no longer had one coherent meaning. (Re-reviewing this caught a stale comment on `add_notarization` — fixed in 753e237ba.)

consensus/src/simplex/elector.rs
```diff
@@ -187,15 +187,9 @@ impl<H: Hasher> RoundRobin<H> {
     /// The term length is consensus-critical: every participant must configure
     /// the same value (see [`TermLength`]). The timeout is local policy.
     ///
-    /// # Panics
-    ///
-    /// Panics if `term_length` is 1 (single-view terms need no timeout and are
```
patrick-ogrady: I missread this and thought it was doing a >0 check...I think I'm ok keeping actually? Although, maybe the right fix is to just have >1 in TermLength?

patrick-ogrady (commented):
small nit

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -365,9 +386,13 @@ where
                         updated_view = current.view;
                     }
                     Message::Constructed(message) => {
-                        // If the view isn't interesting, we can skip
-                        if !interesting(self.activity_timeout, finalized, current.view, view, false)
-                        {
+                        // Record activity for ourselves.
```
patrick-ogrady: do we need to do this?

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -520,12 +542,25 @@ where
...
                 }
 
-                // If the view isn't interesting, we can skip
+                // Any same-epoch traffic from a known peer counts as activity, even if the vote is
+                // later ignored. Skip-timeout is a liveness heuristic, not Byzantine evidence.
                 let view = message.view();
-                if !interesting(self.activity_timeout, finalized, current.view, view, false) {
+                self.record_peer_activity(&sender);
+
+                // Ignore votes at or below the finalized tip. Votes dropped here are
```
patrick-ogrady: I guess this is a subtle change to not report votes in the activity window?

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -114,6 +120,7 @@ where
             Buckets::CRYPTOGRAPHY,
         );
         let (sender, receiver) = mailbox::new(context.child("mailbox"), cfg.mailbox_size);
+        let last_activity = vec![None; participants.len()];
```
patrick-ogrady: just define inline on struct field?

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -69,14 +74,50 @@ impl ParentPayloadError {
...
+    current: View,
+    pending: View,
+    allow_unbounded_future: bool,
+    term_length: TermLength,
+) -> bool {
+    // If the view is genesis, skip it, genesis doesn't have votes
+    if pending.is_zero() {
+        return false;
+    }
+    if pending < min_active(activity_timeout, last_finalized) {
```
patrick-ogrady: Given the changes to batching (flooring all usefuless at last finalized), why do we need min active here?

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -104,6 +145,21 @@ pub struct State<E: Clock + CryptoRng + Metrics, S: Scheme<D>, L: ElectorConfig<
...
+    /// Views for which we have voted to nullify.
+    ///
+    /// Used to enforce the term safety rule that suppresses later same-term
+    /// finalize votes until a covering finalization is observed.
+    nullify_views: BTreeSet<View>,
+
+    /// Views for which we have nullification certificates. Used to answer term-level
+    /// nullification queries efficiently (for parent validation and entry certificate fallback)
+    /// without scanning all tracked rounds.
+    nullification_views: BTreeSet<View>,
```
patrick-ogrady, line 127: This may not be fully populated on restart? I assume that's ok?

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -189,14 +243,32 @@ impl<E: Clock + CryptoRng + Metrics, S: Scheme<D>, L: ElectorConfig<S>, D: Diges
...
     }
 
-    /// Returns whether `pending` is still relevant for progress, optionally allowing future views.
-    pub fn is_interesting(&self, pending: View, allow_future: bool) -> bool {
+    /// Returns the term length of the elector.
+    fn term_length(&self) -> TermLength {
+        self.elector.terms().length()
+    }
+
+    /// Returns whether a vote for `pending` is still relevant for progress.
```
patrick-ogrady: call me crazy but should we not be calling this from batcher?

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -362,13 +511,18 @@ impl<E: Clock + CryptoRng + Metrics, S: Scheme<D>, L: ElectorConfig<S>, D: Diges
...
+        let next_view = view.next_term_start(self.term_length());
+        self.enter_view(next_view);
+        self.set_leader(next_view, Some(&nullification.certificate));
 
         // Track nullification metric per leader (if we know who the leader was)
         let round = self.create_round(view);
         let added = round.add_nullification(nullification);
         let leader = added.then(|| round.leader()).flatten();
+        self.nullification_views.insert(view);
+
```
patrick-ogrady: can remove this \n

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -415,8 +569,30 @@ impl<E: Clock + CryptoRng + Metrics, S: Scheme<D>, L: ElectorConfig<S>, D: Diges
...
 
-    /// Construct a finalize vote if the round provides a candidate.
+    /// Construct a finalize vote if the round provides a candidate and it is safe to do so.
+    ///
+    /// The term safety rule applies: a prior same-term nullify vote blocks the
+    /// finalize vote unless an observed finalization covers it (see
+    /// [Same-Term Vote Safety](crate::simplex#same-term-vote-safety)).
     pub fn construct_finalize(&mut self, view: View) -> Option<Finalize<S, D>> {
+        // We don't need to finalize views that are already finalized.
+        if view <= self.last_finalized {
```
patrick-ogrady, line 545: is this a necessary change for this PR? or something extra?

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -684,25 +881,44 @@ impl<E: Clock + CryptoRng + Metrics, S: Scheme<D>, L: ElectorConfig<S>, D: Diges
...
+            .range(term_start..view)
+            .next_back()
+            .copied()
+    }
+
+    /// Returns the first non-nullified view in the open interval (after, before).
+    ///
+    /// A nullification nullifies all views in the rest of its term.
+    fn first_unnullified_view(&self, after: View, before: View) -> Option<View> {
+        // Defensive: both callers already guarantee `after < before`.
```
patrick-ogrady: if already guaranteed, should be an assert?

consensus/src/simplex/actors/resolver/state.rs
```diff
@@ -84,66 +92,69 @@ impl<S: Scheme, D: Digest> State<S, D> {
...
-            // This may occur before or after a nullification for the same view (and should always be favored).
-            // Finalization remains the stronger proof and can later supersede this floor at the same or higher view.
             if let Some(notarization) = self.notarizations.remove(&view)
                 && view > self.floor_view()
             {
                 self.floor = Some(Certificate::Notarization(notarization));
                 effects.push(self.prune());
             }
+
+            effects.extend(self.fetch_missing(view));
```
patrick-ogrady, line 154: add comment?

consensus/src/simplex/actors/resolver/state.rs
```diff
@@ -84,66 +92,69 @@ impl<S: Scheme, D: Digest> State<S, D> {
...
 
     /// Handle a certification result from the voter.
     pub fn handle_certified(&mut self, view: View, success: bool) -> Vec<Effect> {
         let mut effects = Vec::new();
         if success {
-            // Certification passed - set floor to notarization if we have it.
-            //
-            // This may occur before or after a nullification for the same view (and should always be favored).
-            // Finalization remains the stronger proof and can later supersede this floor at the same or higher view.
             if let Some(notarization) = self.notarizations.remove(&view)
```
patrick-ogrady, line 144: restore comment?

consensus/src/simplex/actors/resolver/state.rs
```diff
@@ -196,46 +213,76 @@ impl<S: Scheme, D: Digest> State<S, D> {
...
-            self.fetch_floor = last.next();
+    ///
+    /// Scans the cursor view (never below the floor, and mid-term after a
+    /// floor-raise pull-back), then subsequent term anchors, advancing the
+    /// cursor past every view scanned so later calls do not re-request
+    /// them. Requests stay pending in the resolver until
+    /// answered or retained out (we must eventually receive a nullification
+    /// at the anchor or a notarization/finalization at a higher view). The
+    /// one case where a scanned term can lose its request — a floor raise
+    /// landing mid-term — pulls the cursor back (see [Self::prune]) so a
```
patrick-ogrady: Is this pull-back necessary?

consensus/src/simplex/elector.rs
```diff
@@ -63,23 +61,92 @@ pub trait Config<S: Scheme>: Clone + Default + Send + 'static {
     fn build(self, participants: &Set<S::PublicKey>) -> Self::Elector;
 }
 
+/// Leadership term structure reported by an [`Elector`].
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+pub enum Terms {
+    /// Every view is its own term: a new leader is elected each view, and
+    /// leader rotation itself bounds how long finality can stall.
+    #[default]
+    Single,
```
patrick-ogrady: I think a better word could be useful here to delineate from `TermLength=1` being invalid?

consensus/src/simplex/elector.rs
BrendanChou: This is still checked by the engine, so this was a duplicate check. We should keep `1` as a valid value for TermLength, which is deliberate

consensus/src/simplex/actors/batcher/actor.rs
BrendanChou: Done in efc92ed10. Note it had to be hoisted above `scheme` in the initializer (out of declaration order): `participants` borrows from `scheme`, the `scheme` field moves it, and struct expressions evaluate fields in textual order — so using `participants` any later is a borrow-across-move error.

consensus/src/simplex/actors/voter/state.rs
BrendanChou: Done in efc92ed10.

consensus/src/simplex/actors/voter/state.rs
BrendanChou: Removed it entirely in efc92ed10: `cursor` starts at `after.next()`, so when `before <= after` the loop never runs and we return `None` anyway — the guard was dead code.

consensus/src/simplex/actors/resolver/state.rs
BrendanChou: Restored in efc92ed10.

consensus/src/simplex/actors/resolver/state.rs
BrendanChou: Added in efc92ed10: a floor raise landing mid-term pulls the fetch cursor back (see `prune`), and a previously truncated scan can resume.

consensus/src/simplex/elector.rs
BrendanChou: Clarified the doc in efc92ed10: `Single` is the only way to configure single-view terms — `Stable { length: 1 }` is rejected at engine construction.

consensus/src/simplex/actors/voter/state.rs
BrendanChou: Necessary.

The safety check right below it ("a prior nullify vote blocks this finalize unless a finalization covers it") uses the cheap comparison `nullified > last_finalized`. This simple comparison is only valid if `view > last_finalized`.

consensus/src/simplex/actors/voter/state.rs
BrendanChou: It is rebuilt on restart: nullifications are journaled (`handle_nullification` appends `Artifact::Nullification`) and the replay loop re-inserts them here.

consensus/src/simplex/actors/batcher/actor.rs
BrendanChou: Yes — we never receive our own messages from the network, so without this the recently-active count in `is_active`'s quorum fail-open check is permanently one short.

consensus/src/simplex/actors/resolver/state.rs
BrendanChou: Yes. The scan's invariant is that every scanned term is either covered or has a request pending. A mid-term floor raise breaks it: the term's anchor request is retained out (keyed at or below the floor) and the cursor has already scanned past, so the tail `(floor, term_end]` is scanned, uncovered, and unrequested — and nothing ever re-requests it (this scan is the only fetcher of nullification coverage). A validator in that state can't vote for proposals skipping the tail, and since a mid-term certification puts many validators in the same state at once, the later finalization that would moot the gap may never form. The pull-back repairs the invariant. Covered by `certification_success_at_mid_term_floor_refetches_term_tail` and `mid_term_floor_at_current_view_refetches_term_tail_later`.

consensus/src/simplex/actors/batcher/actor.rs
BrendanChou: Yes that's right

consensus/src/simplex/actors/voter/state.rs
BrendanChou: This looks like basically the same discussion as earlier about `interesting()`, so I will leave this comment open if we need to discuss it

consensus/src/simplex/actors/batcher/actor.rs
patrick-ogrady: I wonder if it is more consistent with other code here to do quorum-1? I suppose that gets weird if we are a follower?

consensus/src/simplex/actors/batcher/actor.rs
patrick-ogrady: Should we be making those tweaks in this PR?

patrick-ogrady (approved):

PR #3354, opened 2026-03-10
[reshare] Track validator progress across crash restarts

clabby:
## Overview

In `reshare`'s DKG actor, we persist and sync the `EpochState` prior to invoking the callback:

https://github.com/commonwarexyz/monorepo/blob/7cfe616113cdebc113b631eb67d661fb33b48fc7/examples/reshare/src/dkg/actor.rs#L560-L584

The crash tests are currently a bit brittle, assuming that the harness would observe a dense callback stream. If the node crashes between persisting the `EpochState` and invoking the callback, a node can observe a higher epoch on restart.

https://github.com/commonwarexyz/monorepo/blob/7cfe616113cdebc113b631eb67d661fb33b48fc7/examples/reshare/src/validator.rs#L621-L625

### DIFF
--- a/examples/reshare/src/validator.rs
+++ b/examples/reshare/src/validator.rs
@@ -536,18 +536,14 @@ mod test {
     impl ProgressTracker {
         fn observe(
             &mut self,
-            pk: &PublicKey,
+            pk: PublicKey,
             epoch: Epoch,
             output: Option<Output<MinSig, PublicKey>>,
         ) -> anyhow::Result<()> {
-            match self.status.get(pk).copied() {
-                None => {}
-                Some(previous) if previous < epoch => {}
-                previous => {
-                    return Err(anyhow!(
-                        "unexpected update epoch transition {previous:?} -> {epoch:?}"
-                    ));
-                }
+            if let Some(previous) = self.status.get(&pk).filter(|previous| **previous >= epoch) {
+                return Err(anyhow!(
+                    "unexpected update epoch transition {previous:?} -> {epoch:?}"
+                ));
             }
 
             match self.outputs.entry(epoch) {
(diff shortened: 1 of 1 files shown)


examples/reshare/src/validator.rs
```diff
@@ -519,6 +519,91 @@ mod test {
...
+        fn observe(
+            &mut self,
+            pk: &PublicKey,
+            epoch: Epoch,
+            output: Option<Output<MinSig, PublicKey>>,
+        ) -> anyhow::Result<()> {
+            match self.status.get(pk).copied() {
+                None => {}
+                Some(previous) if previous < epoch => {}
+                previous => {
```
cronokirby: Do we intentionally want to log `Some(previous)`? Otherwise would make more sense to log the value itself, matching on `Some(previous)`

cronokirby: Nit:

Personally I think something like:

```rust
if let Some(bad_previous) = self.status.get(pk).filter(|p| p >= epoch) {
  panic
}
```
would be clearer

examples/reshare/src/validator.rs
```diff
@@ -519,6 +519,91 @@ mod test {
...
+    struct ProgressTracker {
+        outputs: BTreeMap<Epoch, Option<Output<MinSig, PublicKey>>>,
+        status: BTreeMap<PublicKey, Epoch>,
+        successes: u64,
+    }
+
+    impl ProgressTracker {
+        fn observe(
+            &mut self,
+            pk: &PublicKey,
```
cronokirby: I would take ownership of `pk`, just as a general pattern of doing that if you need to rather than taking a reference and cloning.

examples/reshare/src/validator.rs
```diff
@@ -519,6 +519,91 @@ mod test {
...
+                if output.is_none() {
+                    continue;
+                }
+                remaining -= 1;
+                if remaining == 0 {
+                    return Some(*epoch);
+                }
+            }
+            None
+        }
```
cronokirby, line 577:
```suggestion
fn success_target_epoch(&self, success_target: u64) -> Option<Epoch> {
    self.outputs
        .iter()
        .filter_map(|(epoch, output)| output.map(|_| *epoch))
        .nth(success_target as usize - 1)
}
```

examples/reshare/src/validator.rs
clabby: Fixed

PR #3360, opened 2026-03-10
[runtime/iobuf] reduce buffer-pool hot-path overhead

andresilva:
This PR improves the iobuf buffer pool in a few focused steps. The goal is to make the steady-state alloc/free path materially cheaper, especially under contention.

### Remove metrics from hot paths

Moves metrics updates out of the hottest alloc/free paths where possible. We still keep the metrics that matter for capacity planning and failure analysis:

- `created`: updated when a class actually creates a new tracked buffer
- `exhausted_total`: updated when an allocation fails because a class is out of tracked capacity
- `oversized_total`: updated when a request exceeds the configured maximum size

These are slow-path or low-frequency events compared to steady-state reuse, so we keep the operational signal we care about without paying per-operation metrics overhead on every hot alloc/free pair.

### Add per-thread caches to size classes

Each size class now behaves like a small two-level allocator:

- a per-thread TLS cache for the common steady-state case
- a shared global freelist for refill and spill

On a TLS hit, alloc/free stays entirely on the current thread and avoids touching the shared queue. On a local miss, the class refills a small batch from the global freelist instead of pulling a single buffer at a time. Returned buffers first go back to the current thread-local cache, and only spill back to the global freelist when that cache is full.

The TLS cache implementation uses dense per-thread bins indexed by a stable size-class id, which keeps the local-cache lookup cheap and also ensures cached buffers are flushed back to the global freelist when a thread exits.

This keeps the common reuse path local while still allowing buffers to move between threads when needed.

### Track pooled buffers by size class

Tracked pooled buffers now carry their originating `SizeClass` directly instead of going through `Weak<BufferPoolInner>`. That removes two sources of overhead from the return path:

- drops no longer need to `upgrade()` a weak pointer to recover pool state
- drops no longer need to recompute the size class from buffer capacity before returning the buffer

Just as importantly, the current implementation now moves the `Arc<SizeClass>` through the TLS cache on steady-state reuse rather than cloning it on every alloc/free pair. In the steady-state TLS-hit path, ownership is moved, not refcounted.

This preserves the existing semantic that pooled buffers may outlive the public `BufferPool` handle, while making the steady-state path much cheaper than the old pool-wide weak ownership model.

### Separate aligned fallback from pooled backing

This PR also cleans up the non-pooled aligned path.

Requests below `pool_min_size`, as well as oversized or exhausted fallbacks, now use dedicated aligned `IoBuf` / `IoBufMut` backing instead of pretending to be degenerate pooled buffers. This removes extra pooled-wrapper machinery from the direct aligned path.

### Tradeoffs

The important design point in the current implementation is that the steady-state fast path does not need bookkeeping beyond thread-local cache manipulation.

When alloc/free stays within the current thread's TLS cache, the implementation does not touch the global freelist, does not upgrade weak ownership, and does not perform shared lifetime coordination. On that path, pooled buffers simply move their `Arc<SizeClass>` in and out of the TLS cache along with the buffer itself.

The cost of synchronization is therefore pushed to the paths that already need global coordination:

- refilling from the shared global freelist
- spilling buffers back to the shared global freelist
- creating new tracked buffers
- hitting exhaustion or oversized slow paths

That is the key tradeoff this PR is aiming for. We still preserve the semantic that pooled buffers may outlive the public `BufferPool` handle, but we structure ownership so the hot local-reuse path mostly moves state around rather than coordinating shared state on every alloc/free pair.

This is also why more aggressive alternatives that leak allocator state did not look compelling enough to keep. They can remove even more lifetime machinery, but the current design already avoids that cost on the steady-state TLS-hit path while still preserving full cleanup and correct pool lifecycle semantics.

### DIFF
--- a/stream/src/encrypted.rs
+++ b/stream/src/encrypted.rs
@@ -749,7 +749,10 @@ mod test {
     fn test_send_many_sends_oversized_single_message_alone() -> Result<(), Error> {
         let executor = deterministic::Runner::new(
             deterministic::Config::new().with_network_buffer_pool_config(
-                BufferPoolConfig::for_network().with_max_size(NZUsize!(128)),
+                BufferPoolConfig::for_network()
+                    .with_pool_min_size(128)
+                    .with_min_size(NZUsize!(128))
+                    .with_max_size(NZUsize!(128)),
             ),
         );
         executor.start(|context| async move {


andresilva:
I looked at whether we should add a stack-backed small-buffer path for `IoBufMut`, but I do not think it is viable.

The main issue is representation size. A true inline/stack-backed variant would have to live inside `IoBuf` / `IoBufMut`, which means every instance would grow to the size of that inline payload, even when the buffer is actually heap-backed or pooled. `IoBuf` / `IoBufMut` are ordinary owning values and can be stored across `await` points, so inflating them would also directly inflate async state machines.

The obvious workaround is to keep the small buffer behind a pointer, for example in a TLS slab or arena. That can reduce allocator overhead, but at that point the optimization is no longer really "inline", it is just another pointer-backed allocator path. Since the pooled path already uses a TLS-backed fast path for requests above `pool_min_size`, that would only help the small direct-allocation case and would add another allocator regime to maintain.

The only version that really gets the full inline benefit is a separate type. The shape would be something like:

```rust
#[must_not_suspend] // this isn't actually stable
pub struct SmallIoBufMut<const N: usize> {
    data: [u8; N],
}

#[must_not_suspend]
pub enum SmallOrHeapBufMut<const N: usize> {
    Inline(SmallIoBufMut<N>),
    Heap(IoBufMut),
}

impl<const N: usize> SmallOrHeapBufMut<N> {
    pub fn with_capacity(capacity: usize, alignment: NonZeroUsize) -> Self {
        if capacity <= N {
            Self::Inline(SmallIoBufMut::new(capacity, alignment))
        } else {
            Self::Heap(IoBufMut::with_alignment(capacity, alignment))
        }
    }

    pub fn into_iobuf_mut(self) -> IoBufMut {
        match self {
            Self::Inline(buf) => buf.into_iobuf_mut(),
            Self::Heap(buf) => buf,
        }
    }
}
```

This would be very annoying to use, and if you have a limited scope where you need to manipulate a small amount of data you can just stack-allocate an array anyway. I also don't think we'd be able to honor arbitrary alignment requests with inline array.

runtime/src/iobuf/pool.rs
```diff
@@ -578,27 +743,28 @@ impl BufferPool {
...
         let mut classes = Vec::with_capacity(config.num_classes());
         for i in 0..config.num_classes() {
             let size = config.class_size(i);
-            let class = SizeClass::new(
+            let class_id = NEXT_SIZE_CLASS_ID.fetch_add(1, Ordering::Relaxed);
+            let class = Arc::new(SizeClass::new(
+                class_id,
                 size,
                 config.alignment.get(),
                 config.max_per_class.get(),
```
patrick-ogrady, line 832: imho we should allow the local size to be configured (I suspect there will be cases where we couple a buffer pool with parallel on some subset of total threads) 🤷

runtime/src/iobuf/aligned.rs
```diff
@@ -0,0 +1,1203 @@
...
+        drop(buffer);
+    }
+}
+
+/// Owner that returns the buffer to its originating pool size class on drop.
+///
+/// The `Arc<SizeClass>` is *moved* (not cloned) through the TLS cache on the
+/// steady-state alloc/free path, avoiding refcount traffic.
+#[derive(Clone)]
+pub(crate) struct TrackedOwner {
```
patrick-ogrady, line 127: My LLM had some thoughts on this:

```
  - runtime/src/iobuf/aligned.rs: TrackedOwner now holds a Weak<SizeClass> instead of a strong Arc. When the last pooled reference drops, it returns the buffer
    only if the size class is still alive; otherwise it deallocates. I also made AlignedBuffer explicitly Send because it is uniquely-owned storage and now
    participates in the per-class thread-local cache structure.
  - runtime/src/iobuf/pool.rs: local free-buffer caches now live under each SizeClass instead of in a separate global TLS registry. That fixes the shutdown leak
    because cached free buffers no longer keep dead size classes alive from outside the pool. The alloc path is still local cache -> global freelist -> new
    allocation, and on a miss it can reclaim a free buffer stranded in another thread’s local bin so worker-thread cached buffers remain reusable.
  - runtime/src/iobuf/pool.rs: added two regression tests, one for “checked-out buffer dropped after pool shutdown” and one for “free buffer already cached in a
    live thread at pool shutdown.” I also updated the existing thread-exit/local-cache tests to match the new cache ownership model.
  - runtime/Cargo.toml and Cargo.lock: added the direct thread_local dependency because the pool code now uses it directly again.
```

Recommended changes (maybe just worth taking regression tests):

```
diff --git a/Cargo.lock b/Cargo.lock
index 62a0c8cb6..aa3340414 100644
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1607,6 +1607,7 @@ dependencies = [
  "sha2 0.10.9",
  "sysinfo",
  "thiserror 2.0.17",
+ "thread_local",
  "tokio",
  "tracing",
  "tracing-opentelemetry",
diff --git a/runtime/Cargo.toml b/runtime/Cargo.toml
index e20745727..29b81c37b 100644
--- a/runtime/Cargo.toml
+++ b/runtime/Cargo.toml
@@ -37,6 +37,7 @@ rayon.workspace = true
 sha2.workspace = true
 sysinfo.workspace = true
 thiserror.workspace = true
+thread_local = "1.1"
 tracing.workspace = true
 tracing-opentelemetry.workspace = true
 tracing-subscriber = { workspace = true, features = ["fmt", "json", "env-filter"] }
diff --git a/runtime/src/iobuf/aligned.rs b/runtime/src/iobuf/aligned.rs
index 6f980f76e..0d861ee3b 100644
--- a/runtime/src/iobuf/aligned.rs
+++ b/runtime/src/iobuf/aligned.rs
@@ -16,7 +16,7 @@ use std::{
     mem::ManuallyDrop,
     ops::{Bound, RangeBounds},
     ptr::NonNull,
-    sync::Arc,
+    sync::{Arc, Weak},
 };
 
 /// A heap allocation with explicit alignment.
@@ -31,6 +31,11 @@ pub(crate) struct AlignedBuffer {
     layout: Layout,
 }
 
+// SAFETY: `AlignedBuffer` uniquely owns a heap allocation and does not expose
+// safe aliasing access to its contents, so transferring ownership across
+// threads is sound.
+unsafe impl Send for AlignedBuffer {}
+
 impl std::fmt::Debug for AlignedBuffer {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         f.debug_struct("AlignedBuffer")
@@ -119,19 +124,21 @@ impl Owner for UntrackedOwner {
     }
 }
 
-/// Owner that returns the buffer to its originating pool size class on drop.
-///
-/// The `Arc<SizeClass>` is *moved* (not cloned) through the TLS cache on the
-/// steady-state alloc/free path, avoiding refcount traffic.
+/// Owner that remembers the originating pool size class without extending its
+/// lifetime after pool shutdown.
 #[derive(Clone)]
 pub(crate) struct TrackedOwner {
-    class: Arc<SizeClass>,
+    class: Weak<SizeClass>,
 }
 
 impl Owner for TrackedOwner {
     #[inline]
     fn release(self, buffer: AlignedBuffer) {
-        TlsCache::push(self.class, buffer);
+        if let Some(class) = self.class.upgrade() {
+            TlsCache::push(class, buffer);
+        } else {
+            drop(buffer);
+        }
     }
 }
 
@@ -183,7 +190,7 @@ impl<O: Owner> Drop for BufInner<O> {
 ///
 /// The owner type decides what happens when the final reference is dropped:
 /// untracked buffers deallocate directly, while pooled buffers return the
-/// allocation to their originating size class.
+/// allocation to their originating size class when it is still alive.
 ///
 /// # View Layout
 ///
@@ -373,7 +380,8 @@ impl<O: Owner> bytes::Buf for Buf<O> {
 /// Mutable aligned buffer view.
 ///
 /// When dropped, the underlying buffer is released according to `O`: untracked
-/// buffers deallocate directly, while tracked buffers return to the pool.
+/// buffers deallocate directly, while tracked buffers return to the pool only
+/// while the originating size class remains alive.
 ///
 /// # Buffer Layout
 ///
@@ -600,8 +608,8 @@ pub(crate) type AlignedBuf = Buf<UntrackedOwner>;
 /// Immutable, reference-counted view over a pooled allocation.
 ///
 /// The final reference returns the underlying aligned buffer to its originating
-/// [`SizeClass`]. See [`Buf`] for the shared immutable view layout and
-/// invariants.
+/// [`SizeClass`] when that class is still alive, and otherwise deallocates it
+/// directly. See [`Buf`] for the shared immutable view layout and invariants.
 pub(crate) type PooledBuf = Buf<TrackedOwner>;
 
 /// Mutable view over an untracked aligned allocation.
@@ -613,7 +621,8 @@ pub(crate) type AlignedBufMut = BufMut<UntrackedOwner>;
 /// Mutable view over a pooled allocation.
 ///
 /// When dropped, the underlying aligned allocation is returned to its
-/// originating [`SizeClass`]. See [`BufMut`] for the shared mutable layout and
+/// originating [`SizeClass`] when that class is still alive, and otherwise
+/// deallocated directly. See [`BufMut`] for the shared mutable layout and
 /// invariants.
 pub(crate) type PooledBufMut = BufMut<TrackedOwner>;
 
@@ -678,9 +687,14 @@ impl std::fmt::Debug for BufMut<TrackedOwner> {
 
 impl BufMut<TrackedOwner> {
     #[inline]
-    pub(crate) const fn new(buffer: AlignedBuffer, class: Arc<SizeClass>) -> Self {
+    pub(crate) fn new(buffer: AlignedBuffer, class: Arc<SizeClass>) -> Self {
         Self {
-            inner: ManuallyDrop::new(BufInner::new(buffer, TrackedOwner { class })),
+            inner: ManuallyDrop::new(BufInner::new(
+                buffer,
+                TrackedOwner {
+                    class: Arc::downgrade(&class),
+                },
+            )),
             cursor: 0,
             len: 0,
         }
@@ -695,7 +709,8 @@ impl BufMut<TrackedOwner> {
     ///
     /// Only the readable portion (`cursor..len`) is included in the result.
     /// The underlying buffer will be returned to the pool when all references
-    /// to the [`IoBuf`] (including slices) are dropped.
+    /// to the [`IoBuf`] (including slices) are dropped, unless the pool has
+    /// already shut down, in which case it is deallocated directly.
     pub fn freeze(self) -> IoBuf {
         IoBuf::from_pooled(self.into_pooled())
     }
index 34481b4d4..b017255f6 100644
@@ -13,11 +13,14 @@
 //!
 //! # Pool Lifecycle
 //!
-//! Each tracked buffer keeps a strong reference to the originating size class.
-//! Buffers can outlive the public [`BufferPool`] handle and still return to
-//! their original size class.
+//! Checked-out tracked buffers remember their originating size class without
+//! keeping the pool alive after shutdown.
+//! Buffers can outlive the public [`BufferPool`] handle and remain usable.
 //! - Untracked fallback allocations store no class reference and deallocate
 //!   directly when dropped.
+//! - Tracked allocations return to the pool only while their size class is
+//!   still alive; after shutdown they deallocate directly instead of re-entering
+//!   thread-local caches.
 //! - Requests smaller than [`BufferPoolConfig::pool_min_size`] bypass pooling
 //!   entirely and return untracked aligned allocations from both
 //!   [`BufferPool::try_alloc`] and [`BufferPool::alloc`].
@@ -56,13 +59,13 @@ use prometheus_client::{
     registry::Registry,
 };
 use std::{
-    cell::UnsafeCell,
     num::NonZeroUsize,
     sync::{
         atomic::{AtomicUsize, Ordering},
-        Arc,
+        Arc, Mutex,
     },
 };
+use thread_local::ThreadLocal;
 
 /// Minimum local capacity required before refill/spill starts batching.
 ///
@@ -375,8 +378,6 @@ impl PoolMetrics {
 /// and only creates a new tracked buffer when no free buffer is available and
 /// the class still has remaining capacity.
 pub(super) struct SizeClass {
-    /// Dense global identifier for the TLS cache registry.
-    class_id: usize,
     /// The buffer size for this class.
     size: usize,
     /// Buffer alignment.
@@ -389,21 +390,16 @@ pub(super) struct SizeClass {
     created: AtomicUsize,
     /// Maximum number of buffers retained in the current thread's local bin.
     local_capacity: usize,
+    /// Per-thread local free buffers for this size class.
+    local: ThreadLocal<Mutex<Vec<AlignedBuffer>>>,
 }
 
-// SAFETY: shared state in `SizeClass` is synchronized through atomics and the
-// global queue. Per-thread bins are stored in thread-local registries and only
-// accessed by the current thread.
-unsafe impl Send for SizeClass {}
-// SAFETY: see above.
-unsafe impl Sync for SizeClass {}
-
 impl SizeClass {
     /// Creates a new size class with the given parameters.
     ///
     /// If `prefill` is true, allocates `max` buffers upfront and pushes them
     /// into the global freelist.
-    fn new(class_id: usize, size: usize, alignment: usize, max: usize, prefill: bool) -> Self {
+    fn new(size: usize, alignment: usize, max: usize, prefill: bool) -> Self {
         let freelist = ArrayQueue::new(max);
         let mut created = 0;
         if prefill {
@@ -413,13 +409,13 @@ impl SizeClass {
             created = max;
         }
         Self {
-            class_id,
             size,
             alignment,
             max,
             global: freelist,
             created: AtomicUsize::new(created),
             local_capacity: Self::local_capacity(max),
+            local: ThreadLocal::new(),
         }
     }
 
@@ -443,196 +439,119 @@ impl SizeClass {
         });
     }
 
-    /// Atomically reserves capacity to create one new tracked buffer.
-    ///
-    /// Returns `true` if the reservation succeeded (i.e. `created < max`),
-    /// `false` if the class is at capacity.
-    fn try_reserve(&self) -> bool {
-        self.created
-            .fetch_update(Ordering::Relaxed, Ordering::Relaxed, |created| {
-                (created < self.max).then_some(created + 1)
-            })
-            .is_ok()
-    }
-}
-
-/// Free tracked buffer cached in the current thread's TLS registry.
-///
-/// This is allocator cache state, not a checked-out buffer. The cached
-/// `Arc<SizeClass>` is moved back into a checked-out pooled buffer on a local
-/// hit, or used to flush the buffer into the shared global freelist when the
-/// thread cache spills or is dropped on thread exit.
-struct TlsSizeClassCacheEntry {
-    buffer: AlignedBuffer,
-    class: Arc<SizeClass>,
-}
-
-/// Per-class thread-local cache for tracked buffers.
-///
-/// The hot steady-state path allocates from and returns to this cache. When
-/// the cache is full, small bins route overflow directly to the class-global
-/// freelist while larger bins spill a batch back to it. When the thread exits
-/// its remaining entries are flushed to that same global freelist.
-struct TlsSizeClassCache {
-    entries: Vec<TlsSizeClassCacheEntry>,
-    local_capacity: usize,
-}
-
-impl TlsSizeClassCache {
-    /// Creates a new empty cache with the given maximum local capacity.
-    fn new(local_capacity: usize) -> Self {
-        Self {
-            entries: Vec::with_capacity(local_capacity),
-            local_capacity,
+    /// Returns the current thread's local bin for this size class.
+    #[inline]
+    fn local_bin(&self) -> &Mutex<Vec<AlignedBuffer>> {
+        self.local
+            .get_or(|| Mutex::new(Vec::with_capacity(self.local_capacity)))
+    }
+
+    /// Reuses a tracked buffer from the current thread, shared freelist, or
+    /// returns `None` if no tracked buffer is currently free.
+    fn pop(&self) -> Option<AlignedBuffer> {
+        if let Some(buffer) = self
+            .local_bin()
+            .lock()
+            .expect("local bin lock poisoned")
+            .pop()
+        {
+            return Some(buffer);
         }
-    }
 
-    /// Returns the number of buffers currently in this local cache.
-    #[inline]
-    const fn len(&self) -> usize {
-        self.entries.len()
-    }
+        let target = (self.local_capacity / 2).max(1);
+        if let Some(buffer) = self.global.pop() {
+            let mut refill = Vec::with_capacity(target.saturating_sub(1));
+            while refill.len() + 1 < target {
+                let Some(buffer) = self.global.pop() else {
+                    break;
+                };
+                refill.push(buffer);
+            }
+            self.local_bin()
+                .lock()
+                .expect("local bin lock poisoned")
+                .extend(refill);
+            return Some(buffer);
+        }
 
-    /// Removes and returns the most recently cached buffer, if any.
-    #[inline]
-    fn pop(&mut self) -> Option<TlsSizeClassCacheEntry> {
-        self.entries.pop()
+        self.steal_other_local()
     }
 
-    /// Pushes an entry into the local cache, spilling to global if full.
-    ///
-    /// Small local caches prioritize same-thread locality and route overflow
-    /// directly to the global freelist. Once the local cache is large enough
-    /// to batch effectively, half the entries are drained to amortize global
-    /// queue traffic across future returns.
-    fn push(&mut self, entry: TlsSizeClassCacheEntry) {
-        if self.entries.len() < self.local_capacity {
-            self.entries.push(entry);
-            return;
-        }
+    /// Returns a tracked buffer to the current thread's local cache, spilling
+    /// overflow to the shared freelist when needed.
+    fn push(&self, buffer: AlignedBuffer) {
+        {
+            let mut bin = self.local_bin().lock().expect("local bin lock poisoned");
+            if bin.len() < self.local_capacity {
+                bin.push(buffer);
+                return;
+            }
 
-        if self.local_capacity < MIN_BATCH_CAPACITY {
-            entry.class.push_global(entry.buffer);
-            return;
-        }
+            if self.local_capacity < MIN_BATCH_CAPACITY {
+                drop(bin);
+                self.push_global(buffer);
+                return;
+            }
 
-        // Spill half the cache to global to make room.
-        let spill = self.entries.len().min(self.local_capacity / 2).max(1);
-        for _ in 0..spill {
-            let spilled = self
-                .entries
-                .pop()
-                .expect("spill count must not exceed cached entries");
-            spilled.class.push_global(spilled.buffer);
+            let spill = bin.len().min(self.local_capacity / 2).max(1);
+            let mut spilled = Vec::with_capacity(spill);
+            for _ in 0..spill {
+                spilled.push(
+                    bin.pop()
+                        .expect("spill count must not exceed cached entries"),
+                );
+            }
+            bin.push(buffer);
+            drop(bin);
+            for spilled in spilled {
+                self.push_global(spilled);
+            }
         }
-
-        self.entries.push(entry);
     }
-}
 
-impl Drop for TlsSizeClassCache {
-    fn drop(&mut self) {
-        for entry in self.entries.drain(..) {
-            entry.class.push_global(entry.buffer);
+    /// Reclaims a stranded free buffer from another thread's local bin.
+    fn steal_other_local(&self) -> Option<AlignedBuffer> {
+        let current = self.local_bin() as *const Mutex<Vec<AlignedBuffer>>;
+        for bin in self.local.iter() {
+            if std::ptr::eq(bin, current) {
+                continue;
+            }
+            if let Some(buffer) = bin.lock().expect("local bin lock poisoned").pop() {
+                return Some(buffer);
+            }
         }
+        None
     }
-}
 
-// Each thread owns a sparse registry of per-size-class caches, indexed by the
-// global `SizeClass::class_id`.
-//
-// We intentionally use `Vec<Option<...>>` here:
-// - `class_id` values are dense enough for vector indexing to be cheap
-// - each thread typically touches only a subset of all size classes
-// - `None` represents "this thread has never initialized a cache for this id"
-//
-// This keeps the hot TLS-hit path to "index and branch" without a hash map or
-// any synchronization. The cost is that vectors can accumulate holes over time
-// because ids are not recycled.
-thread_local! {
-    static TLS_CLASS_CACHES: UnsafeCell<Vec<Option<TlsSizeClassCache>>> =
-        const { UnsafeCell::new(Vec::new()) };
+    /// Atomically reserves capacity to create one new tracked buffer.
+    ///
+    /// Returns `true` if the reservation succeeded (i.e. `created < max`),
+    /// `false` if the class is at capacity.
+    fn try_reserve(&self) -> bool {
+        self.created
+            .fetch_update(Ordering::Relaxed, Ordering::Relaxed, |created| {
+                (created < self.max).then_some(created + 1)
+            })
+            .is_ok()
+    }
 }
 
-// Global allocator for `SizeClass::class_id`.
-//
-// Ids are monotonic and never reused. This is deliberate: a reused id would
-// require generation tracking or equivalent validation on every TLS cache
-// access to distinguish a live size class from stale per-thread cache state.
-// Keeping ids monotonic makes the TLS fast path cheaper and simpler at the
-// cost of leaving holes in `TLS_CLASS_CACHES` over process lifetime.
-static NEXT_SIZE_CLASS_ID: AtomicUsize = AtomicUsize::new(0);
-
-/// Facade over the thread-local cache keyed by `SizeClass::class_id`.
-///
-/// Each thread owns a sparse `Vec<Option<TlsSizeClassCache>>`, with one
-/// per-size-class cache allocated lazily on first use. Thread exit naturally
-/// flushes cached buffers back to the shared global freelist because
-/// `TlsSizeClassCache` drains itself in `Drop`.
-///
-/// This type exists to keep the unsafe TLS access localized. All steady-state
-/// cache operations (`pop`, `push`, and `refill`) go through this facade rather
-/// than free functions over the `thread_local!` static.
+/// Facade over the size-class-owned local caches.
 pub(super) struct TlsCache;
 
 impl TlsCache {
     /// Pops a cached buffer from the current thread's local cache for the
-    /// given size class. Returns `None` if the local cache is empty.
+    /// given size class. Returns `None` if no tracked buffer is free.
     #[inline]
-    fn pop(class: &Arc<SizeClass>) -> Option<TlsSizeClassCacheEntry> {
-        Self::with_cache(class.class_id, class.local_capacity, |cache| cache.pop())
+    fn pop(class: &Arc<SizeClass>) -> Option<AlignedBuffer> {
+        class.pop()
     }
 
     /// Returns a buffer to the current thread's local cache for the given
     /// size class, spilling to the global freelist if the cache is full.
     #[inline]
     pub(super) fn push(class: Arc<SizeClass>, buffer: AlignedBuffer) {
-        let class_id = class.class_id;
-        let local_capacity = class.local_capacity;
-        Self::with_cache(class_id, local_capacity, |cache| {
-            cache.push(TlsSizeClassCacheEntry { buffer, class });
-        });
-    }
-
-    /// Batch-refills the local cache from the global freelist.
-    ///
-    /// Pulls up to `target - 1` buffers from global into the local cache. For
-    /// small local bins, batching is disabled and this becomes a no-op. Called
-    /// after a global pop succeeds, so the caller already holds one buffer and
-    /// we warm the cache for subsequent local hits when batching is enabled.
-    #[inline]
-    fn refill(class: &Arc<SizeClass>, target: usize) {
-        Self::with_cache(class.class_id, class.local_capacity, |cache| {
-            while cache.len() + 1 < target {
-                let Some(buffer) = class.global.pop() else {
-                    break;
-                };
-                cache.push(TlsSizeClassCacheEntry {
-                    buffer,
-                    class: class.clone(),
-                });
-            }
-        });
-    }
-
-    /// Accesses the current thread's local cache for `class_id`, creating it
-    /// lazily on first use, and invokes `f` on it.
-    #[inline]
-    fn with_cache<R>(
-        class_id: usize,
-        local_capacity: usize,
-        f: impl FnOnce(&mut TlsSizeClassCache) -> R,
-    ) -> R {
-        TLS_CLASS_CACHES.with(|bins| {
-            // SAFETY: this TLS value is only ever accessed by the current thread.
-            let bins = unsafe { &mut *bins.get() };
-            if class_id >= bins.len() {
-                bins.resize_with(class_id + 1, || None);
-            }
-            let cache =
-                bins[class_id].get_or_insert_with(|| TlsSizeClassCache::new(local_capacity));
-            f(cache)
-        })
+        class.push(buffer);
     }
 }
 
@@ -664,19 +583,9 @@ impl BufferPoolInner {
     fn try_alloc(&self, class_index: usize, zero_on_new: bool) -> Option<Allocation> {
         let class = &self.classes[class_index];
 
-        // Fast path: reuse from thread-local cache (no atomics, no metrics).
-        if let Some(entry) = TlsCache::pop(class) {
-            return Some(Allocation {
-                buffer: entry.buffer,
-                is_new: false,
-                class: entry.class,
-            });
-        }
-
-        // Medium path: refill from global freelist.
-        let target = (class.local_capacity / 2).max(1);
-        if let Some(buffer) = class.global.pop() {
-            TlsCache::refill(class, target);
+        // Reuse an existing tracked buffer from the local/global free lists
+        // before attempting to grow the class.
+        if let Some(buffer) = TlsCache::pop(class) {
             return Some(Allocation {
                 buffer,
                 is_new: false,
@@ -746,9 +655,7 @@ impl BufferPool {
         let mut classes = Vec::with_capacity(config.num_classes());
         for i in 0..config.num_classes() {
             let size = config.class_size(i);
-            let class_id = NEXT_SIZE_CLASS_ID.fetch_add(1, Ordering::Relaxed);
             let class = Arc::new(SizeClass::new(
-                class_id,
                 size,
                 config.alignment.get(),
                 config.max_per_class.get(),
@@ -961,13 +868,7 @@ mod tests {
     };
 
     fn test_size_class(size: usize, alignment: usize) -> Arc<SizeClass> {
-        Arc::new(SizeClass::new(
-            NEXT_SIZE_CLASS_ID.fetch_add(1, Ordering::Relaxed),
-            size,
-            alignment,
-            8,
-            false,
-        ))
+        Arc::new(SizeClass::new(size, alignment, 8, false))
     }
 
     fn test_registry() -> Registry {
@@ -1006,13 +907,12 @@ mod tests {
     /// Helper to get the number of free buffers parked in the current thread's
     /// local cache for a size class.
     fn get_local_len(class: &SizeClass) -> usize {
-        TLS_CLASS_CACHES.with(|bins| {
-            // SAFETY: this TLS value is only ever accessed by the current thread.
-            let bins = unsafe { &*bins.get() };
-            bins.get(class.class_id)
-                .and_then(Option::as_ref)
-                .map_or(0, TlsSizeClassCache::len)
-        })
+        class
+            .local
+            .get()
+            .map_or(0, |bin: &Mutex<Vec<AlignedBuffer>>| {
+                bin.lock().expect("local bin lock poisoned").len()
+            })
     }
 
     #[test]
@@ -1463,42 +1363,35 @@ mod tests {
 
     #[test]
     fn test_tls_refill_stops_when_global_runs_empty() {
-        let class = test_size_class(64, 64);
+        let threads = std::thread::available_parallelism().map_or(1, NonZeroUsize::get);
+        let class = Arc::new(SizeClass::new(64, 64, threads * 8, false));
+        assert!(class.local_capacity >= MIN_BATCH_CAPACITY);
 
-        // A short global freelist should refill only what exists, then stop.
+        // A short global freelist should warm only what exists, then stop.
+        class.push_global(AlignedBuffer::new(class.size, class.alignment));
         class.push_global(AlignedBuffer::new(class.size, class.alignment));
-        TlsCache::refill(&class, MIN_BATCH_CAPACITY);
+        let reused = class.pop().expect("global entry should be reusable");
 
+        drop(reused);
         assert_eq!(get_local_len(&class), 1);
         assert_eq!(class.global.len(), 0);
     }
 
     #[test]
     fn test_tls_size_class_cache_push_tolerates_empty_spill() {
-        let class = test_size_class(64, 64);
-        let mut cache = TlsSizeClassCache {
-            entries: Vec::new(),
-            local_capacity: 0,
-        };
+        let class = Arc::new(SizeClass::new(64, 64, 1, false));
 
         // Small local capacities should bypass batching and push straight to global.
-        cache.push(TlsSizeClassCacheEntry {
-            buffer: AlignedBuffer::new(class.size, class.alignment),
-            class,
-        });
-        drop(cache);
+        class.push(AlignedBuffer::new(class.size, class.alignment));
+        class.push(AlignedBuffer::new(class.size, class.alignment));
+        assert_eq!(get_local_len(&class), 1);
+        assert_eq!(class.global.len(), 1);
     }
 
     #[test]
     #[should_panic(expected = "tracked buffer should always fit in the global pool")]
     fn test_push_global_panics_when_global_queue_is_inconsistently_full() {
-        let class = Arc::new(SizeClass::new(
-            NEXT_SIZE_CLASS_ID.fetch_add(1, Ordering::Relaxed),
-            64,
-            64,
-            1,
-            false,
-        ));
+        let class = Arc::new(SizeClass::new(64, 64, 1, false));
 
         // Overfilling the fixed-size global queue should trip the invariant.
         class.push_global(AlignedBuffer::new(64, 64));
@@ -1863,8 +1756,8 @@ mod tests {
 
     #[test]
     fn test_thread_exit_flushes_local_bin() {
-        // When a thread exits, its TLS cache Drop flushes buffers back to the
-        // global freelist, making them available to other threads.
+        // When a thread exits, buffers cached in its local bin should still be
+        // reclaimable by another thread.
         let page = page_size();
         let mut registry = test_registry();
         let pool = Arc::new(BufferPool::new(test_config(page, page, 1), &mut registry));
@@ -1880,26 +1773,36 @@ mod tests {
         .join()
         .expect("worker thread should exit cleanly");
 
-        // After thread exit, the buffer should be in the global freelist (not
-        // stuck in a dead thread's local cache).
         let class_index = pool
             .inner
             .config
             .class_index(page)
             .expect("class exists for page-sized buffer");
-        assert_eq!(pool.inner.classes[class_index].global.len(), 1);
-        assert_eq!(get_local_len(&pool.inner.classes[class_index]), 0);
+        assert_eq!(
+            pool.inner.classes[class_index]
+                .created
+                .load(Ordering::Relaxed),
+            1
+        );
 
-        // The flushed buffer should be reusable from the main thread.
-        let _buf = pool
+        // The returned buffer should be reusable from the main thread without
+        // growing the class.
+        let buf = pool
             .try_alloc(page)
             .expect("thread-exited local buffer should be reusable");
+        assert_eq!(
+            pool.inner.classes[class_index]
+                .created
+                .load(Ordering::Relaxed),
+            1
+        );
+        drop(buf);
     }
 
     #[test]
     fn test_pool_dropped_before_buffer() {
         // What happens if the pool is dropped while buffers are still in use?
-        // The size class remains alive until the last tracked buffer is dropped.
+        // Checked-out buffers remain usable until they are dropped.
 
         let page = page_size();
         let mut registry = test_registry();
@@ -1915,11 +1818,79 @@ mod tests {
         // Buffer should still be usable
         assert_eq!(iobuf.len(), 100);
 
-        // Dropping the buffer should not panic and should return to the retained size class.
+        // Dropping the buffer should not panic after pool shutdown.
         drop(iobuf);
         // No assertion here - we just want to make sure it doesn't panic
     }
 
+    #[test]
+    fn test_dropping_last_buffer_after_pool_drop_reclaims_size_class() {
+        let pool = Arc::new(BufferPool::new(test_config(page, page, 2), &mut registry));
+
+        let class_index = pool
+            .inner
+            .config
+            .class_index(page)
+            .expect("class exists for page-sized buffer");
+        let weak_class = Arc::downgrade(&pool.inner.classes[class_index]);
+
+        // Return one tracked buffer from a worker thread so another thread must
+        // reclaim it from a non-local cache path.
+        let worker_pool = Arc::clone(&pool);
+        thread::spawn(move || {
+            let buf = worker_pool
+                .try_alloc(page)
+                .expect("worker should allocate tracked buffer");
+            drop(buf);
+        })
+        .join()
+        .expect("worker thread should exit cleanly");
+
+        // Keep one checked-out pooled buffer alive across pool shutdown.
+        let mut buf = pool.try_alloc(page).expect("main thread should allocate");
+        buf.put_slice(&[0u8; 16]);
+        let iobuf = buf.freeze();
+
+        drop(pool);
+        assert!(
+            weak_class.upgrade().is_none(),
+            "dropping the last pool handle should reclaim the dead size class even while checked-out buffers remain usable"
+        );
+        assert_eq!(iobuf.len(), 16);
+
+        drop(iobuf);
+        assert!(
+            weak_class.upgrade().is_none(),
+            "dropping the checked-out buffer after pool shutdown should not resurrect the dead size class"
+        );
+    }
+
+    #[test]
+    fn test_pool_drop_reclaims_buffers_cached_in_live_thread_local_bin() {
+        let pool = BufferPool::new(test_config(page, page, 2), &mut registry);
+
+        let class_index = pool
+            .inner
+            .config
+            .class_index(page)
+            .expect("class exists for page-sized buffer");
+        let weak_class = Arc::downgrade(&pool.inner.classes[class_index]);
+
+        let buf = pool.try_alloc(page).expect("main thread should allocate");
+        drop(buf);
+        assert_eq!(get_local_len(&pool.inner.classes[class_index]), 1);
+
+        drop(pool);
+        assert!(
+            weak_class.upgrade().is_none(),
+            "dropping the pool should reclaim buffers cached in a live thread-local bin"
+        );
+    }
+
     #[test]
     fn test_pool_exhaustion_and_recovery() {
         // Test pool exhaustion and recovery.

```

patrick-ogrady: > The fix depends on which shutdown semantics we want to preserve:
  >
  > 1. Pool shutdown while the thread stays alive: checked-out buffers and free buffers already cached on that thread should not keep the dead pool/size class
  >    alive.
  > 2. Thread shutdown while the pool stays alive: buffers cached on the exiting thread should still become reusable by other threads.
  >
  > [1] is the review-blocking correctness issue and has a relatively simple fix.
  > [2] is a separate behavior choice; preserving it adds most of the complexity. The current “steal” approach is on a slow path, but it is definitely more
  > machinery.

patrick-ogrady: If this thing is flying around threads, I'm a bit worried about [2] but I suppose the claim would be the working set of threads is likely very small?

patrick-ogrady: A simpler change that still leaks on buffer close but thread remains alive:

```
diff --git a/runtime/src/iobuf/aligned.rs b/runtime/src/iobuf/aligned.rs
index 6f980f76e..3f5dfe95e 100644
--- a/runtime/src/iobuf/aligned.rs
+++ b/runtime/src/iobuf/aligned.rs
@@ -16,7 +16,7 @@ use std::{
     mem::ManuallyDrop,
     ops::{Bound, RangeBounds},
     ptr::NonNull,
-    sync::Arc,
+    sync::{Arc, Weak},
 };
 
 /// A heap allocation with explicit alignment.
@@ -31,6 +31,11 @@ pub(crate) struct AlignedBuffer {
     layout: Layout,
 }
 
+// SAFETY: `AlignedBuffer` uniquely owns a heap allocation and does not expose
+// safe aliasing access to its contents, so transferring ownership across
+// threads is sound.
+unsafe impl Send for AlignedBuffer {}
+
 impl std::fmt::Debug for AlignedBuffer {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         f.debug_struct("AlignedBuffer")
@@ -119,19 +124,21 @@ impl Owner for UntrackedOwner {
     }
 }
 
-/// Owner that returns the buffer to its originating pool size class on drop.
-///
-/// The `Arc<SizeClass>` is *moved* (not cloned) through the TLS cache on the
-/// steady-state alloc/free path, avoiding refcount traffic.
+/// Owner that remembers the originating pool size class without extending its
+/// lifetime after pool shutdown.
 #[derive(Clone)]
 pub(crate) struct TrackedOwner {
-    class: Arc<SizeClass>,
+    class: Weak<SizeClass>,
 }
 
 impl Owner for TrackedOwner {
     #[inline]
     fn release(self, buffer: AlignedBuffer) {
-        TlsCache::push(self.class, buffer);
+        if let Some(class) = self.class.upgrade() {
+            TlsCache::push(class, buffer);
+        } else {
+            drop(buffer);
+        }
     }
 }
 
@@ -183,7 +190,7 @@ impl<O: Owner> Drop for BufInner<O> {
 ///
 /// The owner type decides what happens when the final reference is dropped:
 /// untracked buffers deallocate directly, while pooled buffers return the
-/// allocation to their originating size class.
+/// allocation to their originating size class when it is still alive.
 ///
 /// # View Layout
 ///
@@ -373,7 +380,8 @@ impl<O: Owner> bytes::Buf for Buf<O> {
 /// Mutable aligned buffer view.
 ///
 /// When dropped, the underlying buffer is released according to `O`: untracked
-/// buffers deallocate directly, while tracked buffers return to the pool.
+/// buffers deallocate directly, while tracked buffers return to the pool only
+/// while the originating size class remains alive.
 ///
 /// # Buffer Layout
 ///
@@ -600,8 +608,8 @@ pub(crate) type AlignedBuf = Buf<UntrackedOwner>;
 /// Immutable, reference-counted view over a pooled allocation.
 ///
 /// The final reference returns the underlying aligned buffer to its originating
-/// [`SizeClass`]. See [`Buf`] for the shared immutable view layout and
-/// invariants.
+/// [`SizeClass`] when that class is still alive, and otherwise deallocates it
+/// directly. See [`Buf`] for the shared immutable view layout and invariants.
 pub(crate) type PooledBuf = Buf<TrackedOwner>;
 
 /// Mutable view over an untracked aligned allocation.
@@ -613,7 +621,8 @@ pub(crate) type AlignedBufMut = BufMut<UntrackedOwner>;
 /// Mutable view over a pooled allocation.
 ///
 /// When dropped, the underlying aligned allocation is returned to its
-/// originating [`SizeClass`]. See [`BufMut`] for the shared mutable layout and
+/// originating [`SizeClass`] when that class is still alive, and otherwise
+/// deallocated directly. See [`BufMut`] for the shared mutable layout and
 /// invariants.
 pub(crate) type PooledBufMut = BufMut<TrackedOwner>;
 
@@ -678,9 +687,14 @@ impl std::fmt::Debug for BufMut<TrackedOwner> {
 
 impl BufMut<TrackedOwner> {
     #[inline]
-    pub(crate) const fn new(buffer: AlignedBuffer, class: Arc<SizeClass>) -> Self {
+    pub(crate) fn new(buffer: AlignedBuffer, class: &Arc<SizeClass>) -> Self {
         Self {
-            inner: ManuallyDrop::new(BufInner::new(buffer, TrackedOwner { class })),
+            inner: ManuallyDrop::new(BufInner::new(
+                buffer,
+                TrackedOwner {
+                    class: Arc::downgrade(class),
+                },
+            )),
             cursor: 0,
             len: 0,
         }
@@ -695,7 +709,8 @@ impl BufMut<TrackedOwner> {
     ///
     /// Only the readable portion (`cursor..len`) is included in the result.
     /// The underlying buffer will be returned to the pool when all references
-    /// to the [`IoBuf`] (including slices) are dropped.
+    /// to the [`IoBuf`] (including slices) are dropped, unless the pool has
+    /// already shut down, in which case it is deallocated directly.
     pub fn freeze(self) -> IoBuf {
         IoBuf::from_pooled(self.into_pooled())
     }
index 34481b4d4..63a31050b 100644
@@ -13,11 +13,18 @@
 //!
 //! # Pool Lifecycle
 //!
-//! Each tracked buffer keeps a strong reference to the originating size class.
-//! Buffers can outlive the public [`BufferPool`] handle and still return to
-//! their original size class.
+//! Checked-out tracked buffers remember their originating size class without
+//! keeping the pool alive after shutdown.
+//! Buffers can outlive the public [`BufferPool`] handle and remain usable.
 //! - Untracked fallback allocations store no class reference and deallocate
 //!   directly when dropped.
+//! - Checked-out tracked allocations return to the pool only while their size
+//!   class is still alive; after shutdown they deallocate directly instead of
+//!   re-entering thread-local caches.
+//! - Free buffers already parked in a thread-local cache flush back to the
+//!   global freelist when that thread exits while the pool is alive. After
+//!   shutdown, those cached buffers are dropped when the owning thread tears
+//!   down its cache.
 //! - Requests smaller than [`BufferPoolConfig::pool_min_size`] bypass pooling
 //!   entirely and return untracked aligned allocations from both
 //!   [`BufferPool::try_alloc`] and [`BufferPool::alloc`].
@@ -60,7 +67,7 @@ use std::{
     num::NonZeroUsize,
     sync::{
         atomic::{AtomicUsize, Ordering},
-        Arc,
+        Arc, Weak,
     },
 };
 
@@ -391,13 +398,6 @@ pub(super) struct SizeClass {
     local_capacity: usize,
 }
 
-// SAFETY: shared state in `SizeClass` is synchronized through atomics and the
-// global queue. Per-thread bins are stored in thread-local registries and only
-// accessed by the current thread.
-unsafe impl Send for SizeClass {}
-// SAFETY: see above.
-unsafe impl Sync for SizeClass {}
-
 impl SizeClass {
     /// Creates a new size class with the given parameters.
     ///
@@ -456,33 +456,25 @@ impl SizeClass {
     }
 }
 
-/// Free tracked buffer cached in the current thread's TLS registry.
-///
-/// This is allocator cache state, not a checked-out buffer. The cached
-/// `Arc<SizeClass>` is moved back into a checked-out pooled buffer on a local
-/// hit, or used to flush the buffer into the shared global freelist when the
-/// thread cache spills or is dropped on thread exit.
-struct TlsSizeClassCacheEntry {
-    buffer: AlignedBuffer,
-    class: Arc<SizeClass>,
-}
-
 /// Per-class thread-local cache for tracked buffers.
 ///
 /// The hot steady-state path allocates from and returns to this cache. When
 /// the cache is full, small bins route overflow directly to the class-global
 /// freelist while larger bins spill a batch back to it. When the thread exits
-/// its remaining entries are flushed to that same global freelist.
+/// its remaining entries are flushed to that same global freelist while the
+/// pool is still alive, or dropped directly after pool shutdown.
 struct TlsSizeClassCache {
-    entries: Vec<TlsSizeClassCacheEntry>,
+    entries: Vec<AlignedBuffer>,
+    class: Weak<SizeClass>,
     local_capacity: usize,
 }
 
 impl TlsSizeClassCache {
     /// Creates a new empty cache with the given maximum local capacity.
-    fn new(local_capacity: usize) -> Self {
+    fn new(class: Weak<SizeClass>, local_capacity: usize) -> Self {
         Self {
             entries: Vec::with_capacity(local_capacity),
+            class,
             local_capacity,
         }
     }
@@ -495,24 +487,34 @@ impl TlsSizeClassCache {
 
     /// Removes and returns the most recently cached buffer, if any.
     #[inline]
-    fn pop(&mut self) -> Option<TlsSizeClassCacheEntry> {
+    fn pop(&mut self) -> Option<AlignedBuffer> {
         self.entries.pop()
     }
 
-    /// Pushes an entry into the local cache, spilling to global if full.
+    /// Releases a cached buffer back to the class if it is still alive.
+    #[inline]
+    fn release(&self, buffer: AlignedBuffer) {
+        if let Some(class) = self.class.upgrade() {
+            class.push_global(buffer);
+        } else {
+            drop(buffer);
+        }
+    }
+
+    /// Pushes a buffer into the local cache, spilling to global if full.
     ///
     /// Small local caches prioritize same-thread locality and route overflow
     /// directly to the global freelist. Once the local cache is large enough
     /// to batch effectively, half the entries are drained to amortize global
     /// queue traffic across future returns.
-    fn push(&mut self, entry: TlsSizeClassCacheEntry) {
+    fn push(&mut self, buffer: AlignedBuffer) {
         if self.entries.len() < self.local_capacity {
-            self.entries.push(entry);
+            self.entries.push(buffer);
             return;
         }
 
         if self.local_capacity < MIN_BATCH_CAPACITY {
-            entry.class.push_global(entry.buffer);
+            self.release(buffer);
             return;
         }
 
@@ -523,17 +525,19 @@ impl TlsSizeClassCache {
                 .entries
                 .pop()
                 .expect("spill count must not exceed cached entries");
-            spilled.class.push_global(spilled.buffer);
+            self.release(spilled);
         }
 
-        self.entries.push(entry);
+        self.entries.push(buffer);
     }
 }
 
 impl Drop for TlsSizeClassCache {
     fn drop(&mut self) {
-        for entry in self.entries.drain(..) {
-            entry.class.push_global(entry.buffer);
+        if let Some(class) = self.class.upgrade() {
+            for buffer in self.entries.drain(..) {
+                class.push_global(buffer);
+            }
         }
     }
 }
@@ -567,8 +571,8 @@ static NEXT_SIZE_CLASS_ID: AtomicUsize = AtomicUsize::new(0);
 ///
 /// Each thread owns a sparse `Vec<Option<TlsSizeClassCache>>`, with one
 /// per-size-class cache allocated lazily on first use. Thread exit naturally
-/// flushes cached buffers back to the shared global freelist because
-/// `TlsSizeClassCache` drains itself in `Drop`.
+/// flushes cached buffers back to the shared global freelist while the pool is
+/// alive because `TlsSizeClassCache` drains itself in `Drop`.
 ///
 /// This type exists to keep the unsafe TLS access localized. All steady-state
 /// cache operations (`pop`, `push`, and `refill`) go through this facade rather
@@ -579,19 +583,15 @@ impl TlsCache {
     /// Pops a cached buffer from the current thread's local cache for the
     /// given size class. Returns `None` if the local cache is empty.
     #[inline]
-    fn pop(class: &Arc<SizeClass>) -> Option<TlsSizeClassCacheEntry> {
-        Self::with_cache(class.class_id, class.local_capacity, |cache| cache.pop())
+    fn pop(class: &Arc<SizeClass>) -> Option<AlignedBuffer> {
+        Self::with_cache(class, |cache| cache.pop())
     }
 
     /// Returns a buffer to the current thread's local cache for the given
     /// size class, spilling to the global freelist if the cache is full.
     #[inline]
     pub(super) fn push(class: Arc<SizeClass>, buffer: AlignedBuffer) {
-        let class_id = class.class_id;
-        let local_capacity = class.local_capacity;
-        Self::with_cache(class_id, local_capacity, |cache| {
-            cache.push(TlsSizeClassCacheEntry { buffer, class });
-        });
+        Self::with_cache(&class, |cache| cache.push(buffer));
     }
 
     /// Batch-refills the local cache from the global freelist.
@@ -602,35 +602,29 @@ impl TlsCache {
     /// we warm the cache for subsequent local hits when batching is enabled.
     #[inline]
     fn refill(class: &Arc<SizeClass>, target: usize) {
-        Self::with_cache(class.class_id, class.local_capacity, |cache| {
+        Self::with_cache(class, |cache| {
             while cache.len() + 1 < target {
                 let Some(buffer) = class.global.pop() else {
                     break;
                 };
-                cache.push(TlsSizeClassCacheEntry {
-                    buffer,
-                    class: class.clone(),
-                });
+                cache.push(buffer);
             }
         });
     }
 
-    /// Accesses the current thread's local cache for `class_id`, creating it
+    /// Accesses the current thread's local cache for `class`, creating it
     /// lazily on first use, and invokes `f` on it.
     #[inline]
-    fn with_cache<R>(
-        class_id: usize,
-        local_capacity: usize,
-        f: impl FnOnce(&mut TlsSizeClassCache) -> R,
-    ) -> R {
+    fn with_cache<R>(class: &Arc<SizeClass>, f: impl FnOnce(&mut TlsSizeClassCache) -> R) -> R {
         TLS_CLASS_CACHES.with(|bins| {
             // SAFETY: this TLS value is only ever accessed by the current thread.
             let bins = unsafe { &mut *bins.get() };
-            if class_id >= bins.len() {
-                bins.resize_with(class_id + 1, || None);
+            if class.class_id >= bins.len() {
+                bins.resize_with(class.class_id + 1, || None);
             }
-            let cache =
-                bins[class_id].get_or_insert_with(|| TlsSizeClassCache::new(local_capacity));
+            let cache = bins[class.class_id].get_or_insert_with(|| {
+                TlsSizeClassCache::new(Arc::downgrade(class), class.local_capacity)
+            });
             f(cache)
         })
     }
@@ -640,7 +634,6 @@ impl TlsCache {
 struct Allocation {
     buffer: AlignedBuffer,
     is_new: bool,
-    class: Arc<SizeClass>,
 }
 
@@ -665,11 +658,10 @@ impl BufferPoolInner {
         let class = &self.classes[class_index];
 
         // Fast path: reuse from thread-local cache (no atomics, no metrics).
-        if let Some(entry) = TlsCache::pop(class) {
+        if let Some(buffer) = TlsCache::pop(class) {
             return Some(Allocation {
-                buffer: entry.buffer,
+                buffer,
                 is_new: false,
-                class: entry.class,
             });
         }
 
@@ -680,7 +672,6 @@ impl BufferPoolInner {
             return Some(Allocation {
                 buffer,
                 is_new: false,
-                class: class.clone(),
             });
         }
 
@@ -699,11 +690,7 @@ impl BufferPoolInner {
         } else {
             AlignedBuffer::new(class.size, class.alignment)
         };
-        Some(Allocation {
-            buffer,
-            is_new: true,
-            class: class.clone(),
-        })
+        Some(Allocation { buffer, is_new: true })
     }
 }
 
@@ -818,7 +805,7 @@ impl BufferPool {
         let buffer = self
             .inner
             .try_alloc(class_index, false)
-            .map(|allocation| PooledBufMut::new(allocation.buffer, allocation.class))
+            .map(|allocation| PooledBufMut::new(allocation.buffer, &self.inner.classes[class_index]))
             .ok_or(PoolError::Exhausted)?;
         Ok(IoBufMut::from_pooled(buffer))
     }
@@ -899,7 +886,8 @@ impl BufferPool {
             .try_alloc(class_index, true)
             .ok_or(PoolError::Exhausted)?;
 
-        let mut buf = IoBufMut::from_pooled(PooledBufMut::new(allocation.buffer, allocation.class));
+        let mut buf =
+            IoBufMut::from_pooled(PooledBufMut::new(allocation.buffer, &self.inner.classes[class_index]));
         if allocation.is_new {
             // SAFETY: buffer was allocated with alloc_zeroed, so bytes in 0..len are initialized.
             unsafe { buf.set_len(len) };
@@ -1463,30 +1451,38 @@ mod tests {
 
     #[test]
     fn test_tls_refill_stops_when_global_runs_empty() {
-        let class = test_size_class(64, 64);
+        let threads = std::thread::available_parallelism().map_or(1, NonZeroUsize::get);
+        let class = Arc::new(SizeClass::new(
+            NEXT_SIZE_CLASS_ID.fetch_add(1, Ordering::Relaxed),
+            64,
+            64,
+            threads * 8,
+            false,
+        ));
+        assert!(class.local_capacity >= MIN_BATCH_CAPACITY);
 
         // A short global freelist should refill only what exists, then stop.
         class.push_global(AlignedBuffer::new(class.size, class.alignment));
         TlsCache::refill(&class, MIN_BATCH_CAPACITY);
-
         assert_eq!(get_local_len(&class), 1);
         assert_eq!(class.global.len(), 0);
     }
 
     #[test]
     fn test_tls_size_class_cache_push_tolerates_empty_spill() {
-        let class = test_size_class(64, 64);
-        let mut cache = TlsSizeClassCache {
-            entries: Vec::new(),
-            local_capacity: 0,
-        };
+        let class = Arc::new(SizeClass::new(
+            NEXT_SIZE_CLASS_ID.fetch_add(1, Ordering::Relaxed),
+            64,
+            64,
+            1,
+            false,
+        ));
+        let mut cache = TlsSizeClassCache::new(Arc::downgrade(&class), 0);
 
         // Small local capacities should bypass batching and push straight to global.
-        cache.push(TlsSizeClassCacheEntry {
-            buffer: AlignedBuffer::new(class.size, class.alignment),
-            class,
-        });
+        cache.push(AlignedBuffer::new(class.size, class.alignment));
         drop(cache);
+        assert_eq!(class.global.len(), 1);
     }
 
     #[test]
@@ -1514,18 +1510,18 @@ mod tests {
 
         // Mutable pooled debug should include cursor position.
         let pooled_mut_debug = {
-            let pooled_mut = PooledBufMut::new(AlignedBuffer::new(page, page), Arc::clone(&class));
+            let pooled_mut = PooledBufMut::new(AlignedBuffer::new(page, page), &class);
             format!("{pooled_mut:?}")
         };
         assert!(pooled_mut_debug.contains("PooledBufMut"));
         assert!(pooled_mut_debug.contains("cursor"));
 
         // Empty mutable buffer converts to empty Bytes without retaining pool memory.
-        let empty_from_mut = PooledBufMut::new(AlignedBuffer::new(page, page), Arc::clone(&class));
+        let empty_from_mut = PooledBufMut::new(AlignedBuffer::new(page, page), &class);
         assert!(empty_from_mut.into_bytes().is_empty());
 
         // Immutable pooled debug should include capacity.
-        let pooled = PooledBufMut::new(AlignedBuffer::new(page, page), class).into_pooled();
+        let pooled = PooledBufMut::new(AlignedBuffer::new(page, page), &class).into_pooled();
         let pooled_debug = format!("{pooled:?}");
         assert!(pooled_debug.contains("PooledBuf"));
         assert!(pooled_debug.contains("capacity"));
@@ -1863,8 +1859,8 @@ mod tests {
 
     #[test]
     fn test_thread_exit_flushes_local_bin() {
-        // When a thread exits, its TLS cache Drop flushes buffers back to the
-        // global freelist, making them available to other threads.
+        // When a thread exits, buffers cached in its local bin should still be
+        // reclaimable by another thread.
         let page = page_size();
         let mut registry = test_registry();
         let pool = Arc::new(BufferPool::new(test_config(page, page, 1), &mut registry));
@@ -1880,26 +1876,36 @@ mod tests {
         .join()
         .expect("worker thread should exit cleanly");
 
-        // After thread exit, the buffer should be in the global freelist (not
-        // stuck in a dead thread's local cache).
         let class_index = pool
             .inner
             .config
             .class_index(page)
             .expect("class exists for page-sized buffer");
-        assert_eq!(pool.inner.classes[class_index].global.len(), 1);
-        assert_eq!(get_local_len(&pool.inner.classes[class_index]), 0);
+        assert_eq!(
+            pool.inner.classes[class_index]
+                .created
+                .load(Ordering::Relaxed),
+            1
+        );
 
-        // The flushed buffer should be reusable from the main thread.
-        let _buf = pool
+        // The returned buffer should be reusable from the main thread without
+        // growing the class.
+        let buf = pool
             .try_alloc(page)
             .expect("thread-exited local buffer should be reusable");
+        assert_eq!(
+            pool.inner.classes[class_index]
+                .created
+                .load(Ordering::Relaxed),
+            1
+        );
+        drop(buf);
     }
 
     #[test]
     fn test_pool_dropped_before_buffer() {
         // What happens if the pool is dropped while buffers are still in use?
-        // The size class remains alive until the last tracked buffer is dropped.
+        // Checked-out buffers remain usable until they are dropped.
 
         let page = page_size();
         let mut registry = test_registry();
@@ -1915,11 +1921,79 @@ mod tests {
         // Buffer should still be usable
         assert_eq!(iobuf.len(), 100);
 
-        // Dropping the buffer should not panic and should return to the retained size class.
+        // Dropping the buffer should not panic after pool shutdown.
         drop(iobuf);
         // No assertion here - we just want to make sure it doesn't panic
     }
 
+    #[test]
+    fn test_dropping_last_buffer_after_pool_drop_reclaims_size_class() {
+        let pool = Arc::new(BufferPool::new(test_config(page, page, 2), &mut registry));
+
+        let class_index = pool
+            .inner
+            .config
+            .class_index(page)
+            .expect("class exists for page-sized buffer");
+        let weak_class = Arc::downgrade(&pool.inner.classes[class_index]);
+
+        // Return one tracked buffer from a worker thread so another thread must
+        // reclaim it from a non-local cache path.
+        let worker_pool = Arc::clone(&pool);
+        thread::spawn(move || {
+            let buf = worker_pool
+                .try_alloc(page)
+                .expect("worker should allocate tracked buffer");
+            drop(buf);
+        })
+        .join()
+        .expect("worker thread should exit cleanly");
+
+        // Keep one checked-out pooled buffer alive across pool shutdown.
+        let mut buf = pool.try_alloc(page).expect("main thread should allocate");
+        buf.put_slice(&[0u8; 16]);
+        let iobuf = buf.freeze();
+
+        drop(pool);
+        assert!(
+            weak_class.upgrade().is_none(),
+            "dropping the last pool handle should reclaim the dead size class even while checked-out buffers remain usable"
+        );
+        assert_eq!(iobuf.len(), 16);
+
+        drop(iobuf);
+        assert!(
+            weak_class.upgrade().is_none(),
+            "dropping the checked-out buffer after pool shutdown should not resurrect the dead size class"
+        );
+    }
+
+    #[test]
+    fn test_pool_drop_reclaims_size_class_with_live_thread_local_bin() {
+        let pool = BufferPool::new(test_config(page, page, 2), &mut registry);
+
+        let class_index = pool
+            .inner
+            .config
+            .class_index(page)
+            .expect("class exists for page-sized buffer");
+        let weak_class = Arc::downgrade(&pool.inner.classes[class_index]);
+
+        let buf = pool.try_alloc(page).expect("main thread should allocate");
+        drop(buf);
+        assert_eq!(get_local_len(&pool.inner.classes[class_index]), 1);
+
+        drop(pool);
+        assert!(
+            weak_class.upgrade().is_none(),
+            "dropping the pool should reclaim the size class even if a live thread-local bin still holds free buffers"
+        );
+    }
+
     #[test]
     fn test_pool_exhaustion_and_recovery() {
         // Test pool exhaustion and recovery.
```

LLM summary:

```
• Trimmed the change down to the cheaper design.

  In runtime/src/iobuf/aligned.rs, checked-out pooled buffers still hold a Weak<SizeClass>, so they no longer pin the pool after shutdown. In runtime/src/iobuf/
  pool.rs, I removed the shared ThreadLocal<Mutex<Vec<_>>> cache and restored the original class_id-indexed per-thread TLS cache shape. Each TLS cache now keeps
  one Weak<SizeClass> for flush-on-thread-exit, rather than a strong Arc per entry. That preserves:

  - pool alive + thread exits: cached buffers flush back to the global freelist
  - pool dropped + checked-out buffer later drops: buffer deallocates directly
  - pool dropped + live TLS cache still holds free buffers: the SizeClass can die immediately, and those raw buffers are dropped later when that TLS cache is torn
    down

  I also changed pooled-buffer construction to downgrade from &Arc<SizeClass> instead of cloning a strong Arc on every allocation, and I fixed the test naming/
  docs to match the narrower semantics.
```

patrick-ogrady (commented):
Starting to explore

runtime/src/iobuf/aligned.rs
andresilva: I didn't look at the diff yet but `Weak` on the hot path will destroy performance (this is what we had before). The main performance gains here is making sure that on the hot path no atomics are touched at all (we just move arcs around without even cloning).

runtime/src/iobuf/pool.rs
andresilva: Yeah I think it makes sense, will add it. The default can still be what we have now.

runtime/src/iobuf/aligned.rs
andresilva: Also on thread shutdown the current code will flush all TLS cached buffers back to the global freelist (this wasn't possible with `ThreadLocal<_>` and it's one of the reasons why I stopped using it).

andresilva: As expected using `Weak` incurs a big performance regression:

This PR (as is):

```
buffer_pool::pool/mode=pool size=4096 threads=1 metric=raw
                        time:   [24.806 ns 24.820 ns 24.836 ns]
buffer_pool::pool/mode=pool size=4096 threads=1 metric=adjusted
                        time:   [21.950 ns 21.959 ns 21.969 ns]
buffer_pool::pool/mode=pool size=65536 threads=1 metric=raw
                        time:   [32.804 ns 32.816 ns 32.831 ns]
buffer_pool::pool/mode=pool size=65536 threads=1 metric=adjusted
                        time:   [12.946 ns 12.966 ns 12.988 ns]
```

Weak patch (https://github.com/commonwarexyz/monorepo/pull/3360#discussion_r2957007162):

```
buffer_pool::pool/mode=pool size=4096 threads=1 metric=raw
                        time:   [41.180 ns 41.193 ns 41.209 ns]
buffer_pool::pool/mode=pool size=4096 threads=1 metric=adjusted
                        time:   [39.214 ns 39.228 ns 39.242 ns]
buffer_pool::pool/mode=pool size=65536 threads=1 metric=raw
                        time:   [49.261 ns 49.278 ns 49.298 ns]
buffer_pool::pool/mode=pool size=65536 threads=1 metric=adjusted
                        time:   [29.490 ns 29.526 ns 29.566 ns]
```

For 4K the `Weak` patch results are slower than doing a direct allocation (whereas with the PR as is 4K pooled is ~80% faster than direct allocation). The current implementation can "leak" if the pool drops while there are still buffers outstanding, basically the entire size class will be retained. IMO this is fine, we don't expose the `BufferPool`, it is tied to the runtime lifetime, and in practice the runtime lifetime is (very likely) tied to the process, so if the pool dies most likely the entire process is about to die as well because the application is shutting down (in which case all memory will be collected anyway). The current implementation guarantees that when a thread exits the local bins are flushed to the global freelist, which IMO is the most important thing, since we have no guarantees about arbitrary threads' lifetime.

andresilva: There's two improvements we can make in the current implementation:
- on pool drop flush the global freelist (this makes sense and doesn't incur any extra cost on the hot-path)
- keep a `closed` atomic bool in the `SizeClass`, on TLS push we check the `closed` bit (with relaxed ordering), if it's closed we flush all the local buffers. This will incur some cost in the hot-path though, but it's smaller than `Weak::upgrade`. I tried this before but don't have concrete numbers about performance impact, will need to measure again.

andresilva: I implemented the suggestions above, here's the results (compare with the results posted above):

```
buffer_pool::pool/mode=pool size=4096 threads=1 metric=raw
                        time:   [29.508 ns 29.550 ns 29.589 ns]
buffer_pool::pool/mode=pool size=4096 threads=1 metric=adjusted
                        time:   [27.133 ns 27.163 ns 27.194 ns]
buffer_pool::pool/mode=pool size=65536 threads=1 metric=raw
                        time:   [36.626 ns 36.641 ns 36.662 ns]
buffer_pool::pool/mode=pool size=65536 threads=1 metric=adjusted
                        time:   [16.835 ns 16.856 ns 16.880 ns]
```

It's not nearly as bad as `Weak` but it's still costly, and IMO it's not worth it. The only case in which this actually works is as follows:
- the pool is dropped
- which implies no threads have a reference to a `BufferPool`
- the thread is still actively using a pooled `IoBuf` though
- once that pooled `IoBuf` gets dropped, on TLS push, the live flag is checked and the entire cache is flushed

This seems like a very unlikely scenario that incurs a cost on every hot-path usage. If the pool gets dropped but the threads aren't doing anything, then the buffers just stay there in the TLS bins. The only way to avoid this is to have some kind of global synchronization that will incur a very heavy cost (like your early LLM suggestion had a mutex on the local bins, this will probably make things 10x slower).

> - on pool drop flush the global freelist (this makes sense and doesn't incur any extra cost on the hot-path)

I will just push a commit with this though since it makes sense and doesn't incur any cost.

andresilva: I think the cheapest option to get global remote drain (i.e. on pool drop, it reclaims all buffers from TLS caches) working is as follows:

- each thread still keeps a TLS pointer to its local bin for a given size class
- the `SizeClass` also keeps a global list of all bins that have ever been created for that class
- on first use of a size class in a thread, we allocate a `ThreadBin`, store it in TLS for that thread, and also link it into the class's global bin list
- pool teardown walks that global bin list and drains the cached buffers from each bin

The simplest way to represent that global list is probably `Mutex<Vec<Arc<ThreadBin>>>`. The mutex is only used for first-touch bin registration and teardown, so it stays off the steady-state hot path (we could have an instrusive linked-list with `AtomicPtr` instead, but it's probably not worth it).

The important part is that discovery is not enough by itself. If teardown remotely drains a `ThreadBin` while some outstanding pooled buffer is being dropped and tries to push back into that same bin, then both sides would mutate the same `Vec<AlignedBuffer>` concurrently, which is UB.

So the minimal synchronization needed on the bin itself is:

- closed: `AtomicBool`
- in_push: `AtomicBool`

`closed` prevents new push from entering once teardown starts. `in_push` lets teardown wait for a push that already got past the first closed check and is still in flight. We don't we need any `pop`-side synchronization, because once the last `BufferPool` handle is gone there are no more `alloc`/`pop` paths for that pool. The only remaining race is with late `push` from outstanding buffers.

The downside is that it still adds non-trivial cost to every `push`:

- one `closed` load
- `in_push` = true
- second `closed` load (we can get rid of this one if we're fine with "leaking" a single buffer in the push/teardown race)
- `in_push` = false

Given the results above of a single `closed` load in the `push` path this will probably be too expensive.

(We could have a single `AtomicBool` as a lock, but that would require a CAS-loop on `push`, which I'm pretty sure would be slower than just loads/stores, because it's RMW operation.)

runtime/src/iobuf/mod.rs
```diff
@@ -2290,6 +2416,12 @@ mod tests {
         prepend_noop.prepend(IoBuf::default());
         assert_eq!(prepend_noop.coalesce(), b"x");
 
+        let mut prepend_into_empty = IoBufs::default();
+        // Prepending into an empty aggregate should stay on the single-buffer fast path.
```
patrick-ogrady, line 2456: Move above let

runtime/src/iobuf/mod.rs
```diff
@@ -3486,6 +3619,17 @@ mod tests {
         assert_eq!(coalesced, b"hello world");
         assert!(coalesced.is_pooled());
 
+        let bufs = IoBufsMut::from(vec![
+            IoBufMut::from(b"a"),
+            IoBufMut::from(b"b"),
+            IoBufMut::from(b"c"),
+            IoBufMut::from(b"d"),
+        ]);
+        // Four chunks force the deque-backed coalesce path instead of pair/triple fast paths.
```
patrick-ogrady: move above let

runtime/src/iobuf/mod.rs
```diff
@@ -3986,15 +4304,21 @@ mod tests {
...
         wrapped.push_back(IoBufMut::with_capacity(1));
-        let _ = wrapped.pop_front();
         wrapped.push_back(IoBufMut::with_capacity(1));
         wrapped.push_back(IoBufMut::with_capacity(1));
+        let _ = wrapped.pop_front();
+        wrapped.push_back(IoBufMut::with_capacity(1));
+        let (first, second) = wrapped.as_slices();
+        assert!(!first.is_empty());
+        assert!(!second.is_empty());
+        // Force `advance_mut` to consume across the wrapped second slice as well.
```
patrick-ogrady, line 4352: add \n before?

runtime/src/iobuf/pool.rs
```diff
@@ -340,205 +364,357 @@ impl PoolMetrics {
...
-        // SAFETY: Layout is valid and has non-zero size.
-        let ptr = unsafe { alloc(layout) };
-        let ptr = NonNull::new(ptr).unwrap_or_else(|| handle_alloc_error(layout));
+    /// Per-thread cache capacity for this size class.
+    ///
+    /// Sized to `max / (2 * threads)`, clamped to `[1, 8]`. The 2x divisor
+    /// reserves half the class budget for the shared global freelist so that
+    /// cross-thread reuse and the global-visibility guarantee remain effective.
+    /// The clamp to 8 caps per-thread memory retention.
+    fn local_capacity(max: usize) -> usize {
```
patrick-ogrady: I think this should be configurable (the number of cores you have may not map to the number of threads where you are using the BufferPool)...if this was just a default, we could at least tune?

patrick-ogrady: If set to 0, would this effectively become what is on `main` now (no TLS?)

runtime/src/iobuf/pool.rs
```diff
@@ -340,205 +364,357 @@ impl PoolMetrics {
...
+// We intentionally use `Vec<Option<...>>` here:
+// - `class_id` values are dense enough for vector indexing to be cheap
+// - each thread typically touches only a subset of all size classes
+// - `None` represents "this thread has never initialized a cache for this id"
+//
+// This keeps the hot TLS-hit path to "index and branch" without a hash map or
+// any synchronization. The cost is that vectors can accumulate holes over time
+// because ids are not recycled.
+thread_local! {
+    static TLS_CLASS_CACHES: UnsafeCell<Vec<Option<TlsSizeClassCache>>> =
```
patrick-ogrady, line 619: If you create a buffer pool instance, shutdown runtime, restart runtime will it get confused?

Thinking of the deterministic runtime usage?

runtime/src/iobuf/pool.rs
```diff
@@ -13,12 +13,14 @@
...
-//! - The pool can be dropped while buffers are still in use; those buffers
-//!   remain valid and will be deallocated when they are dropped.
+//! Each tracked buffer keeps a strong reference to the originating size class.
+//! Buffers can outlive the public [`BufferPool`] handle and still return to
+//! their original size class.
+//! - Untracked fallback allocations store no class reference and deallocate
+//!   directly when dropped.
+//! - Requests smaller than [`BufferPoolConfig::pool_min_size`] bypass pooling
+//!   entirely and return untracked aligned allocations from both
+//!   [`BufferPool::try_alloc`] and [`BufferPool::alloc`].
```
patrick-ogrady, line 23: Could add:
```
  //! Dropping [`BufferPool`] drains only the shared global freelists; checked-out buffers and buffers cached in a live thread's local cache can
  keep their size class alive until they are dropped or the thread exits.
```

runtime/src/iobuf/pool.rs
andresilva: Yes, will push a fix for this. I'll add two ways to set it: explicit `local_capacity` and based on "expected parallelism", i.e. you say the number of threads and it infers an appropriate amount of `local_capacity` based on the same heuristic we use now. This allows us to have a nice default for tokio runtime, i.e. just use configured number of `worker_threads`.

runtime/src/iobuf/pool.rs
andresilva: No, it will not get confused, that's what we use the `NEXT_SIZE_CLASS_ID` for, it's a monotonic identifier of generation.

runtime/src/iobuf/pool.rs
```diff
@@ -340,205 +435,353 @@ impl PoolMetrics {
...
+    }
+}
+
+/// Free tracked buffer cached in the current thread's TLS registry.
+///
+/// This is allocator cache state, not a checked-out buffer. The cached
+/// `Arc<SizeClass>` is moved back into a checked-out pooled buffer on a local
+/// hit, or used to flush the buffer into the shared global freelist when the
+/// thread cache spills or is dropped on thread exit.
+struct TlsSizeClassCacheEntry {
```
patrick-ogrady, line 531: Realized there is a pretty annoying case here where these may get allocated to tokio blocking threads (and because there are hundreds of them allow us to quickly exhaust our global buffer).

patrick-ogrady: Pondering if there is a way to somehow restrict these to tokio worker threads (unless you think it is ok for this to happen)?

andresilva: I don't think there is an easy way to fix this, tokio doesn't provide a hook that distinguishes between worker and blocking threads (and it's not entirely obvious that we want to restrict blocking threads altogether from using the TLS cache). The docs say that idle threads are shut down (https://docs.rs/tokio/latest/tokio/runtime/index.html#lifetime-of-spawned-threads) so this might not be a problem.

andresilva: https://github.com/tokio-rs/tokio/blob/tokio-1.48.0/tokio/src/runtime/blocking/pool.rs#L524-L547

The default is 10 seconds of idle time.

patrick-ogrady: Missed this 👀

patrick-ogrady (approved):

PR #3362, opened 2026-03-10
[consensus/simplex] Inline interesting checks and clarify activity floor

BrendanChou:
## Summary
- remove the shared simplex-level `interesting` helper and inline vote/certificate interestingness checks where they are used
- split interestingness into intent-specific paths (`is_interesting_vote` vs `is_interesting_certificate`) in voter and batcher
- rename `min_active` to `activity_floor` for clearer semantics and update journal pruning to use it

## Behavior
- votes are considered interesting when `pending` is non-genesis, at-or-above `activity_floor`, and at-or-before `current.next()`
- certificates are considered interesting when `pending` is non-genesis and at-or-above `activity_floor`

## Scope
- standalone PR targeting `main`
- limited to simplex voter/batcher interestingness logic and naming cleanup

### DIFF
--- a/consensus/src/simplex/actors/voter/actor.rs
+++ b/consensus/src/simplex/actors/voter/actor.rs
@@ -224,7 +224,7 @@ impl<
         }
         if let Some(journal) = self.journal.as_mut() {
             journal
-                .prune(self.state.min_active().get())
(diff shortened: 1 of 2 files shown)


BrendanChou:
Not needed

PR #3363, opened 2026-03-10
[storage/merkle] Introduce "merkle family" to allow MMR variants, implement MMR and (F-)MMB versions

roberto-bayardo:
- Generalizes `merkle/` module structures by "Merkle Family" and defines both the MMR and MMB (Merkle Mountain Belt) variants.
- Adds the in-mem MMB type.
- Renames MAX_LOCATION to MAX_LEAVES and MAX_POSITION to MAX_NODES plus related changes to clean up some ugliness around indices vs counts/sizes.


storage/src/merkle/position.rs
```diff
@@ -66,157 +61,157 @@ impl Position {
...
-                continue;
-            }
-            // descend to the left child
-            two_h >>= 1;
-            node_pos -= two_h;
-        }
-        true
+    pub fn is_valid_size(self) -> bool {
+        F::is_valid_size(self)
+    }
```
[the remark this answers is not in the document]
roberto-bayardo: this is fine.

storage/src/merkle/mod.rs
```diff
@@ -1,26 +1,79 @@
...
+pub mod mmb;
 pub mod mmr;
 pub mod position;
 pub mod proof;
 
-pub use location::{Location, LocationRangeExt, MAX_LOCATION};
-pub use position::{Position, MAX_POSITION};
+use alloc::vec::Vec;
+pub use location::{Location, LocationRangeExt};
+pub use position::Position;
```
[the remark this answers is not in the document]
roberto-bayardo: this is fine

storage/src/merkle/position.rs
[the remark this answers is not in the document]
roberto-bayardo: this is fine these are alpha apis.

storage/src/merkle/mmb/iterator.rs
```diff
@@ -0,0 +1,556 @@
...
+
+        // Check if n+1 fits within the size boundary. If it does, we use n+1.
+        // If it overshoots, we safely fall back to n.
+        let next_size = size_for_leaves(Location::new(n.as_u64() + 1));
+
+        if next_size.as_u64() <= s {
+            next_size
+        } else {
+            size_for_leaves(n)
+        }
```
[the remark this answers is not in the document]
roberto-bayardo: If size = 2N - k for the true leaf count N, where k = ilog2(N + 1), then the code computes

m = floor(size / 2)
estimate = floor((size + ilog2(m + 1)) / 2)
Two useful facts follow:

m + 1 <= N + 1
=> ilog2(m + 1) <= k
=> estimate <= floor((2N - k + k)/2) = N
So the estimate cannot overshoot the true leaf count.

storage/src/qmdb/mod.rs
```diff
@@ -67,7 +62,7 @@ pub mod sync;
 pub mod verify;
 pub use verify::{
     create_multi_proof, create_proof_store, verify_multi_proof, verify_proof,
-    verify_proof_and_extract_digests, verify_proof_and_pinned_nodes,
+    verify_proof_and_extract_digests,
```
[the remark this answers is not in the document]
roberto-bayardo: While this ordinarily wouldn't matter as this is an alpha API, we should have been calling this exported helper method in qmdb/sync/engine.rs instead of proof.verify_proof_and_pinned_nodes. Re-exported it and updated engine.rs.

storage/src/merkle/mmb/batch.rs
```diff
@@ -0,0 +1,462 @@
...
+        let is_child_leaf = height == 1;
+
+        if target_loc.as_u64() < mid_leaf {
+            let left_pos = birthed_node_pos(left_leaf, is_child_leaf);
+            self.mark_ancestors(left_pos, height - 1, leaf_start, target_loc);
+        } else {
+            let right_pos = birthed_node_pos(right_leaf, is_child_leaf);
+            self.mark_ancestors(right_pos, height - 1, mid_leaf, target_loc);
+        }
+    }
```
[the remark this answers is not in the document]
roberto-bayardo: fixed by reversing the traversal order

storage/src/merkle/mmb/mem.rs
```diff
@@ -0,0 +1,1221 @@
...
+        loc: Location,
+    ) -> Result<Proof<Family, D>, Error> {
+        if !loc.is_valid() {
+            return Err(Error::LocationOverflow(loc));
+        }
+
+        self.range_proof(hasher, loc..loc + 1).map_err(|e| match e {
+            Error::RangeOutOfBounds(loc) => Error::LeafOutOfBounds(loc),
+            _ => e,
+        })
```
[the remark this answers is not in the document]
roberto-bayardo: fixed

storage/conformance.toml
```diff
@@ -74,7 +74,11 @@ hash = "506b5a02d1a7c3e249f514b889e6099c27bb0723ebc711514263f0c71f0676ef"
...
 hash = "4078ce1f5e242f8edb1d41575d51e166af620404036124f8094fc74d4b401047"
 
-["commonware_storage::merkle::proof::tests::conformance::CodecConformance<Proof<Sha256Digest>>"]
+["commonware_storage::merkle::mmr::proof::tests::conformance::CodecConformance<Proof<Family, Sha256Digest>>"]
+n_cases = 65536
+hash = "3636363c592dddcff2eaeb4e6b01ec0cfcca654d69f9fdc1c57626c5adea0de1"
+
+["commonware_storage::merkle::mmr::proof::tests::conformance::CodecConformance<Proof<Family,Sha256Digest>>"]
 n_cases = 65536
 hash = "3636363c592dddcff2eaeb4e6b01ec0cfcca654d69f9fdc1c57626c5adea0de1"
```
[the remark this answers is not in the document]
roberto-bayardo: this is needed.

storage/src/merkle/mmb/mod.rs
```diff
@@ -0,0 +1,245 @@
...
+        }
+    }
+}
+
+impl From<merkle::Error<Family>> for Error {
+    fn from(e: merkle::Error<Family>) -> Self {
+        match e {
+            merkle::Error::LocationOverflow(loc) => Self::LocationOverflow(loc),
+            _ => panic!("unexpected merkle error: {e}"),
+        }
```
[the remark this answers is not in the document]
roberto-bayardo: yep fixed.

storage/src/merkle/hasher.rs
```diff
@@ -1,11 +1,17 @@
...
 
-/// A trait for computing the various digests of an MMR.
+/// A trait for computing the various digests of a Merkle-family structure.
+///
+/// The [`Family`](Self::Family) associated type determines which Merkle family (MMR, MMB, etc.)
+/// this hasher targets, and consequently which `Position` and `Location` types appear in
+/// method signatures. Default implementations are provided for all methods except `hash()`.
 pub trait Hasher: Clone + Send + Sync {
+    /// The Merkle family this hasher targets.
+    type Family: Family;
```
danlaine: Should this be a generic parameter instead of an associated type? A `Hasher` impl should work over any family rather than being associated with a specific one.

roberto-bayardo: yeah it should. changed.

storage/src/bitmap/authenticated.rs
```diff
@@ -601,7 +600,7 @@ impl<E: Clock + RStorage + Metrics, D: Digest, const N: usize> UnmerkleizedBitMa
                     .collect(),
             }
         };
-        batch = batch.update_leaf_batched(&dirty)?;
+        batch.update_leaf_batched(&dirty)?;
```
danlaine: Looks like the API changed (probably during rebase after #3371 was merged)

storage/src/merkle/hasher.rs
```diff
@@ -1,11 +1,16 @@
...
 
-/// A trait for computing the various digests of an MMR.
-pub trait Hasher: Clone + Send + Sync {
+/// A trait for computing the various digests of a Merkle-family structure.
+///
+/// The type parameter `F` determines which Merkle family (MMR, MMB, etc.)
+/// this hasher targets, and consequently which `Position` and `Location` types appear in
+/// method signatures. Default implementations are provided for all methods except `hash()`.
+///
+/// A single type may implement `Hasher<F>` for multiple families.
```
danlaine: nit: I think this is redundant -- by definition a type that implements `Foo<T>` does so for multiple T.
```suggestion
```

roberto-bayardo: yep, ai slop. removed.

storage/src/bitmap/authenticated.rs
roberto-bayardo: sigh, yup. I think I've restored this.

storage/src/journal/authenticated.rs
```diff
@@ -51,7 +50,12 @@ impl<E: Storage + Clock + Metrics, D: Digest, Item> BatchChain<Item> for Mmr<E,
 
 /// A speculative batch whose root digest has not yet been computed,
 /// in contrast to [MerkleizedBatch].
-pub struct UnmerkleizedBatch<'a, H: Hasher, P: Readable<Digest = H::Digest>, Item> {
+pub struct UnmerkleizedBatch<
+    'a,
+    H: Hasher,
+    P: Readable<Family = crate::mmr::Family, Digest = H::Digest, Error = crate::mmr::Error>,
```
danlaine: nit: these are imported as `crate::mmr::*` but we import`mmr::self` already. Let's use `mmr::*` here (and other bounds below)?

storage/src/lib.rs
```diff
@@ -15,6 +15,7 @@ commonware_macros::stability_scope!(ALPHA {
 
     pub mod merkle;
```
danlaine, line 16: I think we should be able to make this non-pub and make callers use re-exports (e.g. of `Location` and other types parameterized on `merkle::*` so that they don't need to know about the internal `merkle` stuff

```suggestion
   mod merkle;
```

storage/src/journal/authenticated.rs
roberto-bayardo: fixed

storage/src/lib.rs
roberto-bayardo: The merkle_family_operations fuzz test imports a bunch of these and would break. (e.g., pub use merkle::{hasher,  Family as MerkleFamily, Position, Location};)

If we want users to define generic structures like this it should probably remain pub?

danlaine: > If we want users to define generic structures like this it should probably remain pub?

_Do_ we intend for people to write code that's generic over either an MMR or an MMB? (Or to implement additional `Family`s?) I consider the `Family` to be an implementation detail rather that we can avoid exposing to consumers of this API.

roberto-bayardo: Unclear. The fuzz test is one example I guess. Without exposing it we'd need to import everything twice.

storage/src/merkle/mmr/journaled.rs
```diff
@@ -1039,11 +1065,11 @@ mod tests {
                 Err(Error::Empty)
             ));
 
-            let changeset = mmr
-                .new_batch()
-                .add(&mut hasher, &test_digest(0))
-                .merkleize(&mut hasher)
-                .finalize();
+            let changeset = {
+                let mut batch = mmr.new_batch();
```
danlaine: nit: could reduce diff by reverting places where we used to do chaining but now don't

roberto-bayardo: Claude let me down with this rebase!

Fixed.

storage/src/merkle/mmb/mod.rs
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    fn to_nearest_size(size: Position) -> Position {
+        iterator::PeakIterator::to_nearest_size(size)
+    }
+
+    fn peaks(size: Position) -> alloc::vec::Vec<(Position, u32)> {
+        let mut peaks: alloc::vec::Vec<_> = iterator::PeakIterator::new(size).collect();
+        peaks.reverse();
```
danlaine: Could we have the iterator yield in the other direction to avoid `.reverse()`?

roberto-bayardo: We need to iterate in this direction for finding which two parents to merge, and this is more performance critical than this usage (which is for proving).

danlaine: It looks like we use `append_merge_height` rather than `PeakIterator` to determine which parents to merge for MMB

roberto-bayardo: Oh right I forgot that was optimized away!  Let me see if it's straightforward to reverse iteration order...

roberto-bayardo: it was! done.

storage/src/merkle/mmr/batch.rs
```diff
@@ -415,41 +168,39 @@ impl<'a, D: Digest, P: Readable<Digest = D>> Readable for MerkleizedBatch<'a, D,
...
+        hasher: &mut impl Hasher<Family, Digest = D>,
+        loc: Location,
+    ) -> Result<Proof<D>, Error> {
+        if !loc.is_valid() {
+            return Err(Error::LocationOverflow(loc));
+        }
+        self.range_proof(hasher, loc..loc + 1).map_err(|e| match e {
+            Error::RangeOutOfBounds(loc) => Error::LeafOutOfBounds(loc),
+            _ => e,
+        })
```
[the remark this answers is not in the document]
roberto-bayardo: fixed

storage/fuzz/fuzz_targets/merkle_family_operations.rs
```diff
@@ -0,0 +1,475 @@
...
+    PruneAll,
+    PruneToPos { pos_idx: u64 },
+}
+
+#[derive(Arbitrary, Debug)]
+struct FuzzInput {
+    operations: Vec<MerkleOperation>,
+}
+
+trait FuzzMerkle: Sized {
```
danlaine: I think we should define a trait for Merkle data structures and then fuzz implementations of that trait (which will become dependencies of QMDB when we change it to be generic over MMR, MMB, etc.)

storage/fuzz/fuzz_targets/mmr_journaled_crash_recovery.rs
```diff
@@ -127,11 +127,11 @@ async fn run_operations(
     for op in operations.iter() {
         let failed = match op {
             MmrOperation::Add { data } => {
-                let changeset = mmr
-                    .new_batch()
-                    .add(hasher, data)
-                    .merkleize(hasher)
-                    .finalize();
+                let changeset = {
```
danlaine: nit: more un-chained LLM changes

roberto-bayardo: sigh, fixed.

storage/fuzz/fuzz_targets/merkle_family_operations.rs
roberto-bayardo: Not sure we need a trait as now the MMR and MMB are literally the same struct (just parameterized differently). Let me parameterize this test to demonstrate....

roberto-bayardo: ok this is done.

storage/src/merkle/read.rs
```diff
@@ -0,0 +1,53 @@
...
+
+    /// Total number of leaves.
+    fn leaves(&self) -> Location<Self::Family> {
+        Location::try_from(self.size()).expect("invalid merkle size")
+    }
+
+    /// `[start, end)` range of retained node positions.
+    fn bounds(&self) -> Range<Position<Self::Family>> {
+        self.pruned_to_pos()..self.size()
+    }
```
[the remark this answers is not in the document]
roberto-bayardo: good catch fixed

storage/src/merkle/proof.rs
```diff
@@ -202,16 +167,12 @@ impl<D: Digest> Proof<D> {
             if !loc.is_valid() {
                 return false;
             }
-            // `loc` is valid so it won't overflow from +1
-            let Ok(bp) = blueprint(self.leaves, *loc..*loc + 1) else {
+            // `loc` is valid so it won't overflro from +1
```
danlaine:
```suggestion
            // `loc` is valid so it won't overflow from +1
```

storage/conformance.toml
```diff
@@ -74,7 +74,11 @@ hash = "506b5a02d1a7c3e249f514b889e6099c27bb0723ebc711514263f0c71f0676ef"
 n_cases = 200
 hash = "4078ce1f5e242f8edb1d41575d51e166af620404036124f8094fc74d4b401047"
 
-["commonware_storage::merkle::proof::tests::conformance::CodecConformance<Proof<Sha256Digest>>"]
+["commonware_storage::merkle::mmr::proof::tests::conformance::CodecConformance<Proof<Family, Sha256Digest>>"]
```
danlaine, line 81: Should we add MMB conformance tests?

storage/fuzz/fuzz_targets/merkle_family_operations.rs
```diff
@@ -0,0 +1,334 @@
...
+        let mut merkle = Mem::<F, Digest>::new(&hasher);
+        let mut reference = ReferenceMerkle::<F>::new();
+
+        for (op_idx, op) in operations.iter().enumerate() {
+            match op {
+                MerkleOperation::Add { data } => {
+                    if merkle.bounds().is_empty() && *merkle.size() > 0 {
+                        continue;
+                    }
+
```
danlaine: This skips add operations when the MM* is fully pruned
```suggestion
```

storage/fuzz/fuzz_targets/merkle_family_operations.rs
```diff
@@ -0,0 +1,334 @@
...
+                    if reference.leaf_locations.is_empty() {
+                        continue;
+                    }
+
+                    let idx = (*location as usize) % reference.leaf_locations.len();
+                    let leaf_loc = reference.leaf_locations[idx];
+                    let limited = limit(new_data);
+
+                    if reference.is_leaf_pruned(leaf_loc) {
+                        continue;
```
danlaine, line 176:
```suggestion
                        assert!(update_leaf(&mut merkle, &hasher, leaf_loc, limited).is_err());
                        continue;
```

storage/fuzz/fuzz_targets/merkle_family_operations.rs
```diff
@@ -0,0 +1,334 @@
...
+
+                    if loc >= merkle.leaves() || loc < retained.start {
+                        continue;
+                    }
+
+                    if let Ok(valid) =
+                        verify_element_proof(&merkle, &hasher, loc, reference.leaf_data[idx].as_slice())
+                    {
+                        assert!(valid, "{} op {op_idx}: proof should verify", type_name::<F>());
+                    }
```
danlaine: I think this should always succeed since `loc` is in bounds -- should this `if let Ok` be an `.expect()`?

storage/fuzz/fuzz_targets/merkle_family_operations.rs
```diff
@@ -0,0 +1,334 @@
...
+                }
+
+                MerkleOperation::PruneAll => {
+                    if merkle.bounds().is_empty() {
+                        continue;
+                    }
+
```
danlaine: Pruning while empty should be fine
```suggestion
```

storage/fuzz/fuzz_targets/merkle_family_operations.rs
```diff
@@ -0,0 +1,334 @@
...
+                MerkleOperation::PruneToPos { pos_idx } => {
+                    if *merkle.size() == 0 {
+                        continue;
+                    }
+
+                    let loc = Location::new(*pos_idx % (*merkle.leaves() + 1));
+                    let retained_before = merkle.bounds().start;
+
+                    if loc <= retained_before || loc > merkle.leaves() {
```
danlaine: `loc > merkle.leaves()` is impossible
```suggestion
                    if loc <= retained_before {
```

storage/src/merkle/mmr/proof.rs
```diff
@@ -0,0 +1,557 @@
...
+            "Should be able to prove with MAX_LOCATION leaves"
+        );
+
+        // MAX_LOCATION + 1 should be rejected (exceeds MAX_LOCATION)
+        let result_overflow = Blueprint::new(max_loc_plus_1, max_loc..max_loc_plus_1);
+        assert!(
+            result_overflow.is_err(),
+            "Should reject location > MAX_LOCATION"
+        );
+        matches!(result_overflow, Err(merkle::Error::LocationOverflow(_)));
```
danlaine:
```suggestion
        assert!matches!(result_overflow, Err(merkle::Error::LocationOverflow(_))));
```

storage/src/merkle/batch.rs
```diff
@@ -0,0 +1,1262 @@
...
+                        (pos, digest)
+                    },
+                )
+                .collect()
+        });
+        for (pos, digest) in computed {
+            self.store_node(pos, digest);
+        }
+    }
+    /// Add a pre-computed leaf digest.
```
danlaine, line 368:
```suggestion

    /// Add a pre-computed leaf digest.
```

storage/src/merkle/mmb/iterator.rs
```diff
@@ -0,0 +1,557 @@
...
+    fn test_leaves_for_size_rejects_invalid() {
+        // Sizes between valid MMB sizes should return None.
+        for n in 1u64..=500 {
+            let valid = size_for_leaves(Location::new(n)).as_u64();
+            let next_valid = size_for_leaves(Location::new(n + 1)).as_u64();
+            for gap in (valid + 1)..next_valid {
+                assert_eq!(
+                    leaves_for_size(Position::new(gap)),
+                    None,
+                    "size={gap} (between n={n} and n={}",
```
danlaine:
```suggestion
                    "size={gap} (between n={n} and n={})",
```

storage/src/merkle/batch.rs
```diff
@@ -0,0 +1,1262 @@
...
+        }
+    }
+}
+
+// ---------------------------------------------------------------------------
+// Batch Chain and Finalize
+// ---------------------------------------------------------------------------
+
+/// Information needed to flatten a chain of batches into a single [`Changeset`].
+pub trait BatchChainInfo<F: Family>: Send + Sync {
```
danlaine: Should we rename this now that it's in `batch` to reduce import stuttering?

```suggestion
pub trait ChainInfo<F: Family>: Send + Sync {
```

storage/src/merkle/mmb/iterator.rs
```diff
@@ -0,0 +1,557 @@
+//! Iterators for traversing MMBs of a given size, and functions for computing various MMB
+//! properties from their output.
+
+use crate::merkle::{
+    mmb::{Family, Location, Position},
+    Family as _,
+};
+
+/// Compute the MMB size required to hold `n` leaves.
+const fn size_for_leaves(n: Location) -> Position {
```
danlaine: Is this equivalent to `Position::try_from`?

storage/src/merkle/mmb/iterator.rs
```diff
@@ -0,0 +1,557 @@
...
+const fn estimate_leaves(size: Position) -> Location {
+    let size_val = size.as_u64();
+    let n = size_val / 2;
+    Location::new((size_val + (n + 1).ilog2() as u64) / 2)
+}
+
+/// Given an MMB size, return the number of leaves N such that `size_for_leaves(N) == size`.
+///
+/// Returns `None` if `size` is not a valid MMB size. Returns `Some(0)` for size 0.
+pub(crate) fn leaves_for_size(size: Position) -> Option<Location> {
```
danlaine: Is this equivalent to `Location::try_from(Position)`?

storage/src/merkle/mmb/iterator.rs
```diff
@@ -0,0 +1,557 @@
...
+    if height == 0 {
+        Location::new(last_leaf)
+    } else {
+        Location::new(last_leaf + (1u64 << (height - 1)) - 1)
+    }
+}
+
+/// Compute the position of the `leaf_index`-th leaf.
+#[inline]
+pub(crate) const fn leaf_pos(leaf_index: Location) -> Position {
```
danlaine: This fn is a pass-through

storage/src/merkle/mmb/mod.rs
```diff
@@ -0,0 +1,246 @@
...
+        let loc = loc.as_u64();
+        // 2*N - ilog2(N+1) for MMB
+        Position::new(2 * loc - (loc + 1).ilog2() as u64)
+    }
+
+    fn position_to_location(pos: Position) -> Option<Location> {
+        let pos = pos.as_u64();
+        // Solve 2*N - ilog2(N+1) = pos for N.
+        // Starting estimate: N ~ (pos + ilog2(N+1)) / 2 ~ pos/2. One refinement gives accuracy
+        // within +/-1, so the loop body runs at most a few times.
```
danlaine: I think the estimate gives either N or N-1, never N+1

storage/src/merkle/mmb/mod.rs
```diff
@@ -0,0 +1,246 @@
...
+/// A node index or node count in an MMB.
+pub type Position = merkle::Position<Family>;
+
+/// A leaf index or leaf count in an MMB.
+pub type Location = merkle::Location<Family>;
+
+pub type StandardHasher<H> = merkle::hasher::Standard<H>;
+
+/// Errors that can occur during MMB operations.
+pub type Error = merkle::Error<Family>;
```
danlaine: nit: can we move these re-exports to the top of the file next to `pub type proof`?

storage/src/merkle/mmb/iterator.rs
```diff
@@ -0,0 +1,557 @@
...
+        let s = size.as_u64();
+
+        // Use the O(1) estimation formula to find the baseline leaf count N.
+        let n = estimate_leaves(size);
+
+        // Check if n+1 fits within the size boundary. If it does, we use n+1.
+        // If it overshoots, we safely fall back to n.
+        let next_size = size_for_leaves(Location::new(n.as_u64() + 1));
+
+        if next_size.as_u64() <= s {
```
danlaine: nit: there are some other places where we convert to `u64` that we could instead use comparator/arithmetic methods on Position/Location
```suggestion
        if size == 0 {
            return size;
        }

        let s = size.as_u64();

        // Use the O(1) estimation formula to find the baseline leaf count N.
        let n = estimate_leaves(size);

        // Check if n+1 fits within the size boundary. If it does, we use n+1.
        // If it overshoots, we safely fall back to n.
        let next_size = size_for_leaves(n + 1);

        if next_size <= s {
```

storage/src/merkle/mod.rs
```diff
@@ -1,26 +1,163 @@
...
+    fn nodes_to_pin(size: Position<Self>, prune_pos: Position<Self>) -> Vec<Position<Self>>;
+
+    /// Return the positions of the left and right children of the node at `pos` with the
+    /// given `height`. The caller guarantees `height > 0` (leaves have no children).
+    fn children(pos: Position<Self>, height: u32) -> (Position<Self>, Position<Self>);
+
+    /// Return the heights of internal nodes created when appending a leaf to a structure
+    /// of the given `size`. The caller appends one leaf and then one placeholder node per
+    /// yielded height, marking each as dirty at that height.
+    fn append_parents(size: Position<Self>) -> impl Iterator<Item = u32>;
```
danlaine: From claude:

```
  append_parents feels like an abstraction break in Family                                                                                                                                                            
                                                                                                                                                                                                                      
  The rest of the trait describes structural properties of the tree (position/location mappings, children, peaks, validity). append_parents is the odd one out — it's an operational procedure ("how to grow the
  tree") that leaks batch-building logic into a structural trait.                                                                                                                                                     
  
  The batch code can derive this itself. It already knows the current size, and it can compute the next valid size via F::location_to_position(leaves + 1). So add_leaf_digest could just fill the gap:               
                                                                       
  let next_size = F::location_to_position(self.leaves() + 1);                                                                                                                                                         
  self.appended.push(digest); // the leaf                              
  while self.size() < next_size {
      let pos = self.size();
      self.appended.push(D::EMPTY);                                                                                                                                                                                   
      self.state.insert(pos, F::node_height(pos));
  }                                                                                                                                                                                                                   
                                                                       
  The missing piece is node_height(pos) -> u32, which IS a structural property — both families can compute it from position alone (MMR via bit tricks, MMB via the birth leaf). Adding it to the trait and removing   
  append_parents keeps the trait purely structural. height should stay on children as a performance hint since callers already have it during traversals.
  ```

File: storage/src/merkle/path.rs [hunk 38]
```diff
@@ -0,0 +1,89 @@
...
+///       2     5
+///      / \   / \
+///     0   1 3   4
+/// ```
+///
+/// `PathIterator` yields: `[(6, 2, 2), (5, 4, 1)]`
+/// - Node 6 (height 2): left child 2 is the sibling, right child 5 is on the path.
+/// - Node 5 (height 1): right child 4 is the sibling, left child 3 is the target leaf.
+#[derive(Debug)]
+pub struct PathIterator<F: Family> {
```
danlaine: nit: `Iterator` since this is in module `path`?

storage/src/merkle/mmb/iterator.rs
```diff
@@ -0,0 +1,557 @@
...
+    pub const fn leaves(&self) -> Location {
+        self.n
+    }
+
+    /// Returns the largest valid MMB size that is no greater than the given size.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `size` exceeds [crate::merkle::Family::MAX_POSITION].
+    pub fn to_nearest_size(size: Position) -> Position {
```
danlaine, line 71: can we add unit tests for this?

storage/conformance.toml
roberto-bayardo: yes, moved conformance testing up to merkle/ and invoking on both the MMR and MMB now.

storage/src/merkle/proof.rs
roberto-bayardo: fixed

storage/src/merkle/mod.rs
roberto-bayardo: Oh that's an interesting idea.

storage/src/merkle/mmb/iterator.rs
roberto-bayardo: done

storage/src/merkle/path.rs
roberto-bayardo: done

storage/src/merkle/mod.rs
roberto-bayardo: After thinking about this more, it's a nice idea but it introduces a bit of inefficiency to do it in this manner.  Also I do think the original method is still structural. I've renamed it `parent_heights` to make the structural property it provides more clear (though it's still the same computation as before).

storage/src/merkle/mmb/batch.rs
```diff
@@ -0,0 +1,21 @@
+//! MMB-specific batch layer built on the shared [`merkle::batch`](crate::merkle::batch)
+//! infrastructure.
+//!
+//! Provides type aliases that fix the family parameter to [`Family`]. All mutation,
+//! merkleization, and proof methods are inherited from the shared [`crate::merkle::batch`] module.
+
+use crate::merkle::{batch, mmb::Family, Proof};
+
+/// MMB-specific type alias for `merkle::proof::Proof`.
+pub type MmbProof<D> = Proof<Family, D>;
```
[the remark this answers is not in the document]
roberto-bayardo: removed

storage/src/merkle/mmb/mod.rs
roberto-bayardo: yes, any overshoot implies the input is not a valid leaf.  clarified the comment.

storage/src/merkle/mmb/mod.rs
```diff
@@ -0,0 +1,246 @@
...
+
+    fn children(pos: Position, height: u32) -> (Position, Position) {
+        iterator::children(pos, height)
+    }
+
+    fn is_valid_size(size: Position) -> bool {
+        Location::try_from(size).is_ok()
+    }
+
+    fn parent_heights(size: Position) -> impl Iterator<Item = u32> {
```
danlaine, line 198: nit: mind adding a comment? It isn't immediately obvious what this does

storage/src/merkle/mmb/mod.rs
roberto-bayardo: done (also done in mmr/mod.rs)

storage/src/merkle/mmb/iterator.rs
roberto-bayardo: Removed.

storage/src/merkle/mmb/iterator.rs
roberto-bayardo: yeah, claude introduced these because it was confused about MAX_* handling for indices vs sizes.  I have cleaned this all up.

storage/src/merkle/mmb/iterator.rs
roberto-bayardo: yeah, claude introduced these because it was confused about MAX_* handling for indices vs sizes. I have cleaned this all up.

storage/src/merkle/mmb/iterator.rs
roberto-bayardo: obsolete comment.

storage/src/merkle/batch.rs
roberto-bayardo: done

storage/src/merkle/mmr/proof.rs
roberto-bayardo: done

storage/src/merkle/mmb/proof.rs
```diff
@@ -0,0 +1,93 @@
...
+
+use crate::merkle::{
+    mmb::{Error, Location, Position},
+    proof as merkle_proof,
+};
+use alloc::collections::BTreeSet;
+
+/// Returns the positions of the minimal set of nodes whose digests are required to prove the
+/// inclusion of the elements at the specified `locations`.
+#[allow(dead_code)]
```
danlaine: Can this be removed?

storage/fuzz/fuzz_targets/merkle_family_operations.rs
roberto-bayardo: removed.

storage/fuzz/fuzz_targets/merkle_family_operations.rs
roberto-bayardo: done

storage/fuzz/fuzz_targets/merkle_family_operations.rs
roberto-bayardo: yes changed.

storage/fuzz/fuzz_targets/merkle_family_operations.rs
roberto-bayardo: done

storage/fuzz/fuzz_targets/merkle_family_operations.rs
roberto-bayardo: removed

storage/src/merkle/mmb/mod.rs
```diff
@@ -0,0 +1,245 @@
...
+pub type Position = merkle::Position<Family>;
+
+/// A leaf index or leaf count in an MMB.
+pub type Location = merkle::Location<Family>;
+
+pub type StandardHasher<H> = merkle::hasher::Standard<H>;
+
+/// Errors that can occur during MMB operations.
+pub type Error = merkle::Error<Family>;
+/// Marker type for the MMB family.
```
danlaine, line 140: nit
```suggestion

/// Marker type for the MMB family.
```

roberto-bayardo: done

storage/src/merkle/mmb/proof.rs
roberto-bayardo: yes, this isn't needed until I get to multi-proving MMBs. removed.

danlaine (approved):

PR #3369, opened 2026-03-11
[Storage] Simplify `mmr::Hasher`

### LINKED ISSUE #1091: [storage/mmr/hasher] clean up Standard.inner() abstraction leak
The storage/mmr/hasher Standard hasher exposes its internal cryptographic hasher via `inner`.  Users of the inner hasher might forget to finalize() any updates, which would cause its other methods to produce incorrect output.   We should introduce a cleaner abstraction that prevents any misuse from breaking other functionality.

danlaine:
Resolves #1091

Simplifies the `mmr::Hasher` trait by:

* **Removing `fork()` and `inner()`** -- `fork()` is replaced by `Clone` (added as a supertrait), and `inner()` (which leaked the underlying `cryptography::Hasher`) is eliminated entirely. Callers that previously needed raw `update()`/`finalize()` access now use the new `hash()` method.

* **Consolidating to a single required method `hash()`** -- accepts an iterator of byte slices, replacing the four previously-required methods (`leaf_digest`, `node_digest`, `root`, `digest`) with default implementations that delegate to `hash()`. Implementors only need to override `node_digest` when they have custom logic (e.g. `GraftedHasher`'s position remapping, `Verifier`'s grafting boundary behavior).

### DIFF
--- a/storage/src/qmdb/current/grafting.rs
+++ b/storage/src/qmdb/current/grafting.rs
@@ -269,14 +269,14 @@ impl<H: CHasher> HasherTrait for Verifier<'_, H> {
 /// Nodes below the grafting height are served from the ops MMR. Nodes at or above the grafting
 /// height are served from the grafted MMR (with ops-to-grafted position conversion). This allows
 /// standard MMR proof generation to work transparently over the combined structure.
-pub(super) struct Storage<'a, D: Digest, G: Readable<D>, S: StorageTrait<D>> {
+pub(super) struct Storage<'a, D: Digest, G: Readable<Digest = D>, S: StorageTrait<Digest = D>> {
     grafted_mmr: &'a G,
     grafting_height: u32,
     ops_mmr: &'a S,
     _digest: PhantomData<D>,
 }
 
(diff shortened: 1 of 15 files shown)


storage/src/qmdb/current/grafting.rs
```diff
@@ -242,19 +220,9 @@ impl<'a, H: CHasher> Verifier<'a, H> {
...
 
-    fn leaf_digest(&mut self, pos: Position, element: &[u8]) -> H::Digest {
-        self.hasher.leaf_digest(pos, element)
-    }
-
-    fn fork(&self) -> impl HasherTrait<Digest = H::Digest> {
-        Verifier::<H> {
-            hasher: StandardHasher::new(),
-            grafting_height: self.grafting_height,
-            chunks: self.chunks.clone(),
```
roberto-bayardo, line 255: Hmm, the purpose of fork() was to avoid deep, costly clones like this one, and to instead use a reference to share read-only structures among them.

Not sure when that got removed?

roberto-bayardo: Never mind, I don't think we ever actually fork this hasher type and the others are lightweight.

PR #3371, opened 2026-03-11
[Storage] Make batch mutation methods consume/produce Self

danlaine:
## Summary

Makes batch mutation methods consistently use a consuming/chaining pattern (`self -> Self`) across all batch APIs in the storage crate.

Previously, the QMDB `any`, `current`, and `immutable` batches used consuming methods (`write(mut self, ...) -> Self`), while the MMR batch, journal authenticated batch, and keyless batch used `&mut self`. This prevented method chaining and made the APIs inconsistent.

### API changes

**MMR batch** (`merkle/mmr/batch.rs`):
- `add()` / `add_leaf_digest()`: `&mut self -> Position` becomes `self -> Self` (no caller used the returned position)
- `update_leaf()` / `update_leaf_digest()` / `update_leaf_batched()`: `&mut self -> Result<(), Error>` becomes `self -> Result<Self, Error>`

**Journal authenticated batch** (`journal/authenticated.rs`):
- `add()`: `&mut self` becomes `self -> Self`

**Keyless batch** (`qmdb/keyless/batch.rs`):
- `append()`: `&mut self -> Location` becomes `self -> Self`
- Added `size() -> Location` to get the next append location (callers use `batch.size()` before `append()` when they need it)

### Call site patterns

Where possible, call sites now chain directly without intermediate bindings:

```rust
// Before
let mut batch = self.mmr.new_batch();
batch.add(&mut self.hasher, &encoded_item);
batch.merkleize(&mut self.hasher).finalize()

// After
self.mmr.new_batch()
    .add(&mut self.hasher, &encoded_item)
    .merkleize(&mut self.hasher)
    .finalize()
```

### DIFF
--- a/storage/src/merkle/mmr/batch.rs
+++ b/storage/src/merkle/mmr/batch.rs
@@ -40,12 +40,11 @@
 //! let mut mmr = Mmr::new(&mut hasher);
 //!
 //! // Build a batch of mutations.
-//! let changeset = {
-//!     let mut batch = UnmerkleizedBatch::new(&mmr);
-//!     batch.add(&mut hasher, b"leaf-0");
-//!     batch.add(&mut hasher, b"leaf-1");
-//!     batch.merkleize(&mut hasher).finalize()
-//! };
+//! let changeset = UnmerkleizedBatch::new(&mmr)
+//!     .add(&mut hasher, b"leaf-0")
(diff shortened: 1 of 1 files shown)


storage/src/merkle/mmr/batch.rs
roberto-bayardo: This example needs updating

Issue #1398, opened 2025-08-13
[fuzzing] Protocol grey-box fuzzer within deterministic simulation


dnkolegov-ar:
# Proposal: Grey-box fuzzer for tests within the deterministic framework

## Summary
Build a fuzzer that combines **structure-aware mutations**, **deterministic simulation** and **coverage-guided search** to stress consensus implementations. The key idea is to **reuse existing tests** (ideally without modification) by recording their message traces and replaying mutated versions under coverage-guided feedback.

## Goals
- Reuse existing test cases as fuzz seeds with no or minimal changes.
- Record and deterministically replay messages between participants.
- Mutate message contents, ordering, duplication, and drops in a structure-aware way.
- Run exploration with code coverage-based feedback.
- Detecting panics or memory corruptions

### 1) Seed Recording (from a normal test run)
- Introduce a **Mock recorder** that, when injected into the test instance, records all messages sent during a test in a **deterministic** order.
- Produce a **JSON seed file** containing:
  - `test_name`, `participants`, `identities`, network/topology, initial RNG seed,
  - ordered list of serialized **messages/events** (sender, receiver, type, payload).
- This “record once, fuzz many” seed is reproducible and portable across runs.

### 2) Fuzz Harness + Deterministic Replay
- Use a `cargo-fuzz` **libFuzzer** target that wraps the target tests and injects the Mock replayer**.
- The harness **loads a recorded seed**, then:
  - **Mutates** the session using structure-aware mutators based on the context annotated the seed file:
    - field perturbations/boundary values,
    - reordering/permutation,
    - duplication, **drop**, delay, truncation/bit-flips,
    - targeted corruptions for message types (headers, signatures, heights, view numbers).
  - **Replays** the mutated session into the test with the same context (participants, identities, RNG seed).

#### Making tests importable (Rust)
Because Rust tests often live in non-exported modules, provide one of:
- A small opt-in helper to export selected tests, **or**
- An auto-generated wrapper (code rewriting) that creates a public entry point without touching test logic.

### 3) Continuous Operation (“data-pipeline” mode)
Run the fuzzer continuously in data pipeline mode:
- Message traces production by the fuzzer,
- Running traces on test instances,
- Artifact upload for crashes/minimized repro seeds

## Deliverables
- [ ] `mock-recorder`
- [ ] ``mock-replayer`
- [ ] Seed **JSON schema** and (de)serializers  
[...]

PR #3374, opened 2026-03-12
[runtime/p2p] Add fuzzing hooks for insitu-fuzz

evonide:
Adds two weak-linkage FFI hook points behind `feature = "fuzz"`:

- `runtime/src/deterministic.rs`: `commonware_fuzz_permute_tasks` lets insitu-fuzz supply a fuzzer-controlled task ordering at each quiescent point, replacing the seeded RNG shuffle. `insitu_fuzz_checkpoint` lets AFL++ deferred-fork targets snapshot state mid-run.

- `p2p/src/utils/codec.rs`: `commonware_fuzz_corrupt_bytes` is called on the encoded wire bytes in both `WrappedSender::send` and `CheckedWrappedSender::send`, allowing the fuzzer to mutate messages before they are delivered.

Both hooks use `#[linkage = "extern_weak"]`: they compile to None when insitu-fuzz is not linked, so there is zero overhead in normal builds. The `fuzz` feature is off by default and not enabled by CI.

Part of the insitu-fuzz integration tracked in https://github.com/commonwarexyz/monorepo/issues/1398#issuecomment-3324548051

### DIFF
--- a/p2p/Cargo.toml
+++ b/p2p/Cargo.toml
@@ -24,6 +24,7 @@ commonware-stream.workspace = true
 commonware-utils.workspace = true
 either.workspace = true
 futures.workspace = true
+libc.workspace = true
 num-bigint.workspace = true
 num-integer.workspace = true
 num-rational.workspace = true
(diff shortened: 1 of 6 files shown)


clabby:

PR #3375, opened 2026-03-12
[insitu-fuzz] Add fuzzable_test macro and expose broadcast/buffered t…

evonide:
- Add `fuzzable_test` proc macro to commonware-macros: converts test attributes to `#[cfg_attr(test, ...)]` and makes functions `pub`, allowing fuzz harnesses to call them directly while they remain normal tests under `cargo test`.
- Expose `broadcast::buffered::tests` and `broadcast::buffered::mocks` under `cfg(any(test, feature = "fuzz"))` as initial target module.
- Add `fuzz = []` feature to broadcast crate.
- Add `MAX_LEN` constant (10 MiB) to bound codec RangeCfg and prevent OOM under fuzz-mutated inputs.
- Annotate 11 fuzzable test functions with `#[fuzzable_test]`.

### DIFF
--- a/broadcast/Cargo.toml
+++ b/broadcast/Cargo.toml
@@ -27,5 +27,8 @@ tracing.workspace = true
 [dev-dependencies]
 tracing-subscriber.workspace = true
 
+[features]
+fuzz = []
+
 [lib]
 bench = false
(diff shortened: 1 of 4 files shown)


clabby:

PR #3376, opened 2026-03-12
[cryptography::bls12381] allow zero `Scalar` while disallowing zero `Private`

roynalnaruto:
### Existing Behaviour

The current behaviour of decoding `bls12381::Scalar` from a bytes buffer disallows the zero-scalar due to the `blst_sk_check` validation here: https://github.com/commonwarexyz/monorepo/blob/4207aa1d15de932e50ff1a6006df90be162d0f30/cryptography/src/bls12381/primitives/group.rs#L630

Although this is required for `bls12381::Private`, it may not always be desirable for `Scalar`.

As an example, if a serializable type (let's say `Block` or `Blob`) must be encoded as `N` scalar field elements (along with deserialization support into the concrete type), then we might need to right-pad zero scalars. Deserialization would fail in this case because `Scalar::decode` fails for zero scalar.

### PR Updates

This PR makes the following changes:

1. `Scalar::decode` will only validate if the deserialized scalar is in-range with `blst_scalar_fr_check`
2. `Private::decode` preserves the current validation, i.e. `blst_sk_check`
3. `Scalar::read_blst_scalar_unchecked` offers a `pub(crate)` functionality to deserialize the `blst_scalar` without any additional checks (expecting the caller to handle the rest)
4. `Scalar` offers a constant-time `is_zero` check
5. Few tests to check the changed behaviour, i.e. scalar should allow zero while private should disallow

*Note:*

- The `read_blst_scalar_unchecked` is only included to avoid going back from `blst_fr` to `blst_scalar` before doing the `blst_sk_check`.
- The example shared above is exactly what I encountered in my [code](https://github.com/roynalnaruto/rlnc/tree/3892d475da832c47d23465978363f10974325819), for which I added [helper](https://github.com/roynalnaruto/rlnc/blob/3892d475da832c47d23465978363f10974325819/crates/primitives/types/src/util.rs#L124).

*Consideration:* The `Private::new(scalar: Scalar) -> Private` at the moment is an infallible method and does not have any checks on the `scalar` argument, i.e. `let p = Private::new(Scalar::zero())` seems to be possible. If it makes sense, I could edit it to `Private::new(scalar: Scalar) -> Result<Private, Error>` and add a non-zero check on the incoming `scalar`.

### DIFF
--- a/cryptography/src/bls12381/primitives/group.rs
+++ b/cryptography/src/bls12381/primitives/group.rs
@@ -487,8 +487,9 @@ impl Read for Private {
         let scalar = Scalar::read_blst_scalar_unchecked(buf)?;
         let mut ret = blst_fr::default();
 
-        // SAFETY: bytes is a valid 32-byte array. blst_sk_check validates non-zero and in-range.
-        // We use blst_sk_check instead of blst_scalar_fr_check because it also checks non-zero
+        // SAFETY: `blst_sk_check` validates non-zero and in-range.
+        //
+        // We use `blst_sk_check` instead of blst_scalar_fr_check because it also checks non-zero
         // per IETF BLS12-381 spec (Draft 4+).
         //
         // References:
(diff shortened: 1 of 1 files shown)


roynalnaruto:
@cronokirby would appreciate a review for the PR :)

Issue #3367, opened 2026-03-11
[p2p] Review Dial Jitter


patrick-ogrady:
In a small network with synchronized configs, it appears some peers dial each other at the same time a few cycles in a row.

This can delay initial connectivity (I suspect we can improve this by adding more jitter to dial attempts)

0xAysh:
The root cause is that both dialer actors initialize their deadlines to current() with no initial jitter:
let mut dial_deadline = self.context.current();
let mut query_deadline = self.context.current();
https://github.com/commonwarexyz/monorepo/blob/4207aa1d15de932e50ff1a6006df90be162d0f30/p2p/src/authenticated/discovery/actors/dialer.rs#L168-L169
https://github.com/commonwarexyz/monorepo/blob/4207aa1d15de932e50ff1a6006df90be162d0f30/p2p/src/authenticated/lookup/actors/dialer.rs#L172-L173

sleep_until(current()) fires immediately, so every node that starts at the same time fires its first query and dial ticks simultaneously. Jitter only kicks in when computing the next deadline after each tick, so the initial synchronization is never broken before the first fire. 

Since add_jittered adds a random offset in [0, 2*jitter], the spread after the first tick is only [0, 2*dial_frequency], with dial_frequency = 1s that's a 2-second window for all N peers to schedule their next attempt.

0xAysh:
@patrick-ogrady I saw your fix. could you explain the reasoning behind the approach? I’m learning the codebase.

patrick-ogrady:
The jitter approach we had was functional but led to an edge case where "query" could fire before the rate limits of the peers had refreshed. That would lead to us rapid attempting dial on them (all fail) and then waiting another "query + jitter" to try again. This could lead to ~3x the "query_frequency" between dial attempts.

Additionally, this lead to us doing a bunch of dials and waiting until the next "query" (rather than just spacing peers out uniformly).

0xAysh:
Thanks, makes sense. If there’s any issue you’d recommend for me, I’m happy to pick it up.

PR #3379, opened 2026-03-12
[p2p/authenticated] Refactor Dial Jitter

patrick-ogrady:
Closes: #3367

### DIFF
--- a/p2p/src/authenticated/discovery/actors/tracker/directory.rs
+++ b/p2p/src/authenticated/discovery/actors/tracker/directory.rs
@@ -296,7 +296,7 @@ impl<E: Spawner + Rng + Clock + RuntimeMetrics, C: PublicKey> Directory<E, C> {
     /// when they are added to a peer set via `add_set`.
     pub fn block(&mut self, peer: &C) {
         // Already blocked
-        if self.blocked.contains(peer) {
+        if self.is_blocked(peer) {
             return;
         }
 
@@ -1713,6 +1713,52 @@ mod tests {
             directory
                 .dial(&peer_pk)
                 .expect("expired block should not prevent reservation");
+        });
+    }
+
+    #[test]
+    fn test_reblock_after_expired_block_without_unblock() {
+        let runtime = deterministic::Runner::default();
+        let signer = PrivateKey::from_seed(0);
+        let my_info = create_myself_info(&signer, test_socket(), 100);
+        let peer_signer = PrivateKey::from_seed(1);
+        let peer_pk = peer_signer.public_key();
+        let (tx, _rx) = UnboundedMailbox::new();
+        let releaser = Releaser::new(tx);
+        let block_duration = Duration::from_secs(1);
+        let config = Config {
+            allow_private_ips: true,
+            allow_dns: true,
+            max_sets: 3,
+            dial_fail_limit: 1,
+            peer_connection_cooldown: Duration::from_millis(200),
+            block_duration,
+        };
+
+        runtime.start(|context| async move {
+            let mut directory = Directory::init(context.clone(), vec![], my_info, config, releaser);
+
+            let peer_set: OrderedSet<_> = [peer_pk.clone()].into_iter().try_collect().unwrap();
+            directory.add_set(0, peer_set);
+            let peer_info = types::Info::sign(&peer_signer, NAMESPACE, test_socket(), 200);
+            directory.update_peers(vec![peer_info]);
+
+            directory.block(&peer_pk);
+            assert!(directory.dialable().peers.is_empty());
+
+            // Advance past expiry without calling unblock_expired().
+            context.sleep(block_duration + Duration::from_secs(1)).await;
+
+            // Re-block should succeed despite stale entry.
+            directory.block(&peer_pk);
+            assert!(
+                directory.dialable().peers.is_empty(),
+                "re-blocked peer should not be dialable"
+            );
+            assert!(
+                directory.dial(&peer_pk).is_none(),
+                "re-blocked peer should not be reservable"
+            );
         });
     }
 }
--- a/p2p/src/authenticated/lookup/actors/tracker/directory.rs
+++ b/p2p/src/authenticated/lookup/actors/tracker/directory.rs
@@ -258,8 +258,8 @@ impl<E: Spawner + Rng + Clock + RuntimeMetrics, C: PublicKey> Directory<E, C> {
     /// Peers can be blocked even if they don't have a record yet. The block will be applied
     /// when they are added to a peer set via `add_set`.
     pub fn block(&mut self, peer: &C) {
-        // Already blocked in queue
-        if self.blocked.contains(peer) {
+        // Already blocked
+        if self.is_blocked(peer) {
             return;
         }
 
@@ -1770,6 +1770,47 @@ mod tests {
             directory
                 .dial(&pk_1)
                 .expect("expired block should not prevent reservation");
+        });
+    }
+
+    #[test]
+    fn test_reblock_after_expired_block_without_unblock() {
+        let runtime = deterministic::Runner::default();
+        let my_pk = ed25519::PrivateKey::from_seed(0).public_key();
+        let pk_1 = ed25519::PrivateKey::from_seed(1).public_key();
+        let addr_1 = SocketAddr::new(IpAddr::V4(Ipv4Addr::new(8, 8, 8, 8)), 1234);
+        let (tx, _rx) = UnboundedMailbox::new();
+        let releaser = super::Releaser::new(tx);
+        let block_duration = Duration::from_secs(1);
+        let config = super::Config {
+            allow_private_ips: true,
+            allow_dns: true,
+            bypass_ip_check: false,
+            max_sets: 3,
+            peer_connection_cooldown: Duration::from_millis(200),
+            block_duration,
+        };
+
+        runtime.start(|context| async move {
+            let mut directory = Directory::init(context.clone(), my_pk, config, releaser);
+            directory.add_set(0, [(pk_1.clone(), addr(addr_1))].try_into().unwrap());
+
+            directory.block(&pk_1);
+            assert!(directory.dialable().peers.is_empty());
+
+            // Advance past expiry without calling unblock_expired().
+            context.sleep(block_duration + Duration::from_secs(1)).await;
+
+            // Re-block should succeed despite stale entry.
+            directory.block(&pk_1);
+            assert!(
+                directory.dialable().peers.is_empty(),
+                "re-blocked peer should not be dialable"
+            );
+            assert!(
+                directory.dial(&pk_1).is_none(),
+                "re-blocked peer should not be reservable"
+            );
         });
     }
 


p2p/src/authenticated/lookup/actors/dialer.rs
```diff
@@ -170,40 +159,42 @@ impl<
...
-        let mut query_deadline = self.context.current();
         select_loop! {
             self.context,
             on_stopped => {
                 debug!("context shutdown, stopping dialer");
             },
             _ = self.context.sleep_until(dial_deadline) => {
-                // Update the deadline.
-                dial_deadline = dial_deadline.add_jittered(&mut self.context, self.dial_frequency);
+                loop {
```
patrick-ogrady: really dislike this rn

p2p/src/authenticated/lookup/actors/tracker/directory.rs
```diff
@@ -242,7 +246,12 @@ impl<E: Spawner + Rng + Clock + RuntimeMetrics, C: PublicKey> Directory<E, C> {
     ///
     /// Returns `Some` on success, `None` otherwise.
     pub fn dial(&mut self, peer: &C) -> Option<(Reservation<C>, Ingress)> {
-        let ingress = self.peers.get(peer)?.ingress()?;
+        let now = self.context.current();
+        let record = self.peers.get(peer)?;
+        if !record.dialable(now, self.allow_private_ips, self.allow_dns) {
```
patrick-ogrady: Not sure these changes are necessary.

p2p/src/authenticated/lookup/actors/tracker/directory.rs
```diff
@@ -391,11 +418,13 @@ impl<E: Spawner + Rng + Clock + RuntimeMetrics, C: PublicKey> Directory<E, C> {
         }
 
         // Already reserved
-        let record = self.peers.get_mut(peer).unwrap();
-        if record.reserved() {
+        if self.peers.get(peer).unwrap().reserved() {
             return None;
         }
 
+        let jitter = self.rate_limit.replenish_interval();
```
patrick-ogrady: what is replenish interval?

p2p/src/authenticated/lookup/actors/tracker/directory.rs
```diff
@@ -390,23 +393,28 @@ impl<E: Spawner + Rng + Clock + RuntimeMetrics, C: PublicKey> Directory<E, C> {
             return None;
         }
 
-        // Already reserved
-        let record = self.peers.get_mut(peer).unwrap();
+        let record = self.peers.get(peer).unwrap();
```
patrick-ogrady: Why can't this be get_mut?

p2p/src/authenticated/lookup/actors/tracker/directory.rs
```diff
@@ -301,19 +290,33 @@ impl<E: Spawner + Rng + Clock + RuntimeMetrics, C: PublicKey> Directory<E, C> {
...
-                !self.blocked.contains(peer) && r.dialable(self.allow_private_ips, self.allow_dns)
-            })
-            .map(|(peer, _)| peer.clone())
-            .collect();
-        result.sort();
-        result
+    /// Returns dialable peers and the next time another peer may become dialable.
+    pub fn dialable(&self) -> Dialable<C> {
+        let now = self.context.current();
+        let interval = self.rate_limit.replenish_interval();
```
patrick-ogrady: Should we just store replenish_interval? we never use quota?

p2p/src/authenticated/lookup/actors/tracker/directory.rs
```diff
@@ -203,13 +204,6 @@ impl<E: Spawner + Rng + Clock + RuntimeMetrics, C: PublicKey> Directory<E, C> {
             });
         }
 
-        // Attempt to remove any old records from the rate limiter.
-        // This is a best-effort attempt to prevent memory usage from growing indefinitely.
-        //
-        // We don't reduce the capacity of the rate limiter to avoid re-allocation on
-        // future peer set additions.
-        self.rate_limiter.retain_recent();
```
patrick-ogrady, line 211: We no longer retain this if a peer drops and is re-added, but I think that's fine.

p2p/src/authenticated/lookup/actors/dialer.rs
```diff
@@ -170,38 +166,30 @@ impl<
...
-                    let Some((reservation, ingress)) = tracker.dial(peer).await else {
-                        continue;
-                    };
-                    self.dial_peer(reservation, ingress, &mut supervisor);
-                    break;
-                }
-            },
-            _ = self.context.sleep_until(query_deadline) => {
-                // Update the deadline.
-                query_deadline =
```
patrick-ogrady, line 196: The big change in this PR is making the jitter `per peer` rather than aligning jitter to all peers we want to re-dial at the same time.

patrick-ogrady: There was a very possible case where we'd wait a while to dial peers (first could fail, query retry would be < rate limit, next dial hits RL before dial, query retry is >> rate limit).

patrick-ogrady:
bugbot run

p2p/src/authenticated/discovery/actors/tracker/actor.rs
```diff
@@ -317,7 +317,7 @@ mod tests {
             allow_dns: true,
             synchrony_bound: Duration::from_secs(10),
             tracked_peer_sets: 2,
-            allowed_connection_rate_per_peer: Quota::per_second(NZU32!(5)),
```
patrick-ogrady, line 320: The `Quota` type can be made quite complex with different constructors (so I learned)...we need to make this argument a `Duration` to be faithful to the input.

p2p/src/authenticated/discovery/actors/dialer.rs
```diff
@@ -166,38 +162,37 @@ impl<
...
+                    self.queue = dialable.peers;
                     self.queue.shuffle(&mut self.context);
+                    next_query_at = dialable.next_query_at;
+                }
+
+                // Set next deadline.
+                dial_deadline = if self.queue.is_empty() {
+                    let min = now + self.dial_frequency;
+                    let max = (now + self.peer_connection_cooldown).max(min);
+                    next_query_at.unwrap_or(max).clamp(min, max)
```
patrick-ogrady, line 185: next_query_at may be < self.dial_frequency, so this ensures we don't do our next dial too fast.

p2p/src/authenticated/lookup/mod.rs
```diff
@@ -57,12 +57,12 @@
 //!
 //! ## Rate Limiting
 //!
-//! There are five primary rate limits:
+//! There are four primary rate limits and one peer connection cooldown:
```
patrick-ogrady: rever this

p2p/src/authenticated/discovery/actors/tracker/directory.rs
```diff
@@ -1,16 +1,18 @@
 use super::{metrics::Metrics, record::Record, set::Set, Metadata, Reservation};
```
BrendanChou, line 1: Looks like you don't have new tests in this `directory.rs` file but you do in `lookup`'s `directory.rs` file, which is asymmetrical.

Same with `dialer.rs`

p2p/src/authenticated/discovery/actors/tracker/record.rs
```diff
@@ -52,6 +62,12 @@ pub struct Record<C: PublicKey> {
 
     /// If `true`, the record should persist even if the peer is not part of any peer sets.
     persistent: bool,
+
+    /// The earliest time we are willing to reserve this peer again (`None` if immediately eligible).
+    next_reservable_at: Option<SystemTime>,
```
BrendanChou: consider just making this `SystemTime` that defaults to the minimum value

p2p/src/authenticated/discovery/actors/tracker/record.rs
```diff
@@ -135,16 +157,26 @@ impl<C: PublicKey> Record<C> {
...
-        if matches!(self.status, Status::Inert) {
-            self.status = Status::Reserved;
-            return true;
+        let now = context.current();
+        if self.next_reservable_at.is_some_and(|t| now < t) {
+            return ReserveResult::RateLimited;
         }
-        false
+        self.status = Status::Reserved;
+        let next_reservable_at = now + interval;
```
BrendanChou: consider something like saturating_add, or could assert on checked_add

p2p/src/authenticated/lookup/actors/tracker/directory.rs
```diff
@@ -301,19 +293,30 @@ impl<E: Spawner + Rng + Clock + RuntimeMetrics, C: PublicKey> Directory<E, C> {
...
-        result.sort();
-        result
+    /// Returns dialable peers and the next time another peer may become dialable.
+    pub fn dialable(&self) -> Dialable<C> {
+        let now = self.context.current();
+        let mut next_query_at: Option<SystemTime> = None;
+        let mut peers = Vec::new();
+        for (peer, record) in &self.peers {
+            if let Some(blocked_until) = self.blocked.get(peer) {
+                next_query_at = Some(next_query_at.map_or(blocked_until, |c| c.min(blocked_until)));
```
BrendanChou: quite annoying we can't use min directly since `None < Some`... wonder if this can be cleaned up at all but I don't have any good ideas besides a helper function (this logic is used later in this function as well)

p2p/src/authenticated/discovery/actors/tracker/directory.rs
patrick-ogrady: my bad.

p2p/src/authenticated/discovery/actors/tracker/directory.rs
```diff
@@ -379,19 +373,30 @@ impl<E: Spawner + Rng + Clock + RuntimeMetrics, C: PublicKey> Directory<E, C> {
...
-            .collect();
-        result.sort();
-        result
+    /// Returns dialable peers and the next time another peer may become dialable.
+    pub fn dialable(&self) -> Dialable<C> {
+        let now = self.context.current();
+        let mut next_query_at: Option<SystemTime> = None;
+        let mut peers = Vec::new();
+        for (peer, record) in &self.peers {
+            if let Some(blocked_until) = self.blocked.get(peer) {
```
clabby: `dialable()` always forwards `blocked_until` to `next_query_at` without checking whether the timestamp has already passed. `PrioritySet::get()` does not auto-expire entries, so correctness currently depends on `unblock_expired()` having run before this method is called. Add a staleness guard before contributing the hint:

```rust
if blocked_until > now {
    next_query_at = next_query_at.min(blocked_until);
}
```

This makes `dialable()` self-contained and prevents stale block times from biasing the scheduler toward unnecessary empty wakeups.

p2p/src/authenticated/lookup/actors/tracker/directory.rs
```diff
@@ -301,19 +293,30 @@ impl<E: Spawner + Rng + Clock + RuntimeMetrics, C: PublicKey> Directory<E, C> {
...
-            .collect();
-        result.sort();
-        result
+    /// Returns dialable peers and the next time another peer may become dialable.
+    pub fn dialable(&self) -> Dialable<C> {
+        let now = self.context.current();
+        let mut next_query_at: Option<SystemTime> = None;
+        let mut peers = Vec::new();
+        for (peer, record) in &self.peers {
+            if let Some(blocked_until) = self.blocked.get(peer) {
```
clabby: Same issue as in `discovery/actors/tracker/directory.rs`: `blocked_until` is unconditionally used as a `next_query_at` hint even when it is already in the past. Because `PrioritySet::get()` returns stored values until explicit removal, a stale entry can leak an expired timestamp into the dialer path and bias scheduling. Guard the hint:

```rust
if blocked_until > now {
    next_query_at = next_query_at.min(blocked_until);
}
```

clabby (dismissed):

PR #3383, opened 2026-03-13
Run Simplex invariant check only when the quorum is feasible

dnkolegov-ar:
This PR fixes the following false positive finding:

```
Running: fuzz/corpus/27706928bd2afbbd36a8a1ef25e8fce9bed07568110d46093a7e647f31265f81
FuzzInput { raw_bytes: [255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 0], required_containers: 18, degraded_network: false, configuration: Configuration { n: 4, faults: 3, correct: 1 }, partition: Connected, strategy: SmallScope { fault_rounds: 2, fault_rounds_bound: 4 } }

thread '<unnamed>' (20404427) panicked at consensus/fuzz/src/invariants.rs:115:13:
Invariant violation: finalization without notarization: view 1, payload=fb494ee3dccd11ae37a06bccab9f80ba32d93cff4c0a671bdd8b4d2256df7526
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

We should not check invariant when `Configuration { n: 4, faults: 3, correct: 1 }` and quorum is not feasible


consensus/fuzz/src/bounds.rs
```diff
@@ -128,19 +128,19 @@ mod tests {
         use crate::{Configuration, N4F1C3, N4F3C1};
 
         // N4F1C3: 4 nodes, 1 faulty - can finalize (1 <= 1)
-        assert!(N4F1C3.can_finalize());
+        assert!(N4F1C3.is_quorum_feasible());
```
patrick-ogrady: can we change this method to `is_valid()` or `is_safe()`? I think "quorum feasible" is deceptive because a "quorum" can be formed with > f (just not the quorum you want lol).

consensus/fuzz/src/lib.rs
```diff
@@ -693,8 +695,10 @@ fn run_with_twin_mutator<P: simplex::Simplex>(input: FuzzInput) {
             context.sleep(MAX_SLEEP_DURATION).await;
         }
 
-        let states = invariants::extract(reporters, config.n as usize);
-        invariants::check::<P>(config.n, states);
+        if config.is_quorum_feasible() {
+            let states = invariants::extract(reporters, config.n as usize);
+            invariants::check::<P>(config.n, states);
```
patrick-ogrady, line 706: I suppose there are no invariants worth checking if f>max_faults?

tbh, there are some cases where we panic if we observe "impossible scenarios" (that can only occur with f > max_faults) -> may be we should just not run with infeasible settings (otherwise you will hit more flakes)?

consensus/fuzz/src/bounds.rs
dnkolegov-ar: fixed

consensus/fuzz/src/lib.rs
dnkolegov-ar: removed, I left configuration check to not fuzz incorrect configs due to a bug or smth

consensus/fuzz/src/lib.rs
```diff
@@ -180,7 +172,7 @@ impl Arbitrary<'_> for FuzzInput {
         Ok(Self {
             raw_bytes,
             partition,
-            configuration,
+            configuration: N4F1C3,
```
patrick-ogrady: If there is only one configuration, do we need to keep this as a field?

dnkolegov-ar: removed

consensus/fuzz/src/lib.rs
```diff
@@ -677,7 +661,7 @@ fn run_with_twin_mutator<P: simplex::Simplex>(input: FuzzInput) {
         }
 
         // Wait for finalization or timeout
-        if input.partition == Partition::Connected && config.can_finalize() {
+        if input.partition == Partition::Connected && config.is_valid() {
```
patrick-ogrady, line 688: AFAICT config is now always valid (it is hardcoded)?

consensus/fuzz/src/lib.rs
```diff
@@ -451,8 +438,7 @@ fn run<P: simplex::Simplex>(input: FuzzInput) {
             reporters.push(reporter);
         }
 
-        // Wait for finalization or timeout
-        if input.partition == Partition::Connected && config.can_finalize() {
+        if input.partition == Partition::Connected && config.is_valid() {
```
patrick-ogrady, line 460: AFAICT config is now always valid (it is hardcoded)?

consensus/fuzz/src/lib.rs
dnkolegov-ar: fixed

consensus/fuzz/src/lib.rs
dnkolegov-ar: fixed

PR #3387, opened 2026-03-13
[consensus/simplex] Forwarding Nits

patrick-ogrady:
Related: #3347

### DIFF
--- a/consensus/src/simplex/actors/batcher/mod.rs
+++ b/consensus/src/simplex/actors/batcher/mod.rs
@@ -1664,10 +1664,9 @@ mod tests {
 
             batcher.start(voter_mailbox, vote_receiver, certificate_receiver);
 
-            // View 2: participants 0..quorum notarize, participant 6 sends
-            // a finalize (implying they already have the block), and
-            // participant 5 is truly missing. Only participant 5 should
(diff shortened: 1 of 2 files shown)


consensus/src/simplex/config.rs
```diff
@@ -26,8 +26,8 @@ pub enum ForwardingPolicy {
     /// Do nothing when notified of missing voters.
     #[default]
     Disabled,
-    /// Forward the block to all active participants that did not vote.
-    All,
+    /// Forward the block to all participants that did not vote.
+    Silent,
```
patrick-ogrady, line 30: `All` implies we'd send the block to all.

consensus/src/simplex/actors/batcher/mod.rs
```diff
@@ -1646,21 +1675,8 @@ mod tests {
...
-                            true,
-                        )
-                        .await
-                        .unwrap();
-                }
-            }
-
+            // Send finalize and nullify BEFORE notarize votes so they are
+            // processed before quorum is reached and missing_voters is called.
             // Participant 5 sends a nullify (establishes activity but no block)
```
clabby: I don’t think this is needed anymore?

Issue #3385, opened 2026-03-13
[storage] Timeout during running of proof_store and range_proof fuzz tests


dnkolegov-ar:
This is reproduced on Linux only. I used:
```
Linux 6.8.0-64-generic #67-Ubuntu SMP PREEMPT_DYNAMIC x86_64 GNU/Linux
```

```
cd storage/fuzz
cargo fuzz run proof_store 2c057904adac3153d9fc13173d4c25b7da2482006f9ee2a5af11aa40b1004558.txt
```

It will be hanging until cargo-fuzz timeout is reached:

**proof_store:**

[2c057904adac3153d9fc13173d4c25b7da2482006f9ee2a5af11aa40b1004558.txt](https://github.com/user-attachments/files/25966314/2c057904adac3153d9fc13173d4c25b7da2482006f9ee2a5af11aa40b1004558.txt)

**range_proof:**

[fccc27a597b9457282c15f12ec470f068e1679a24774e8fc47a84a93da646fcf.txt](https://github.com/user-attachments/files/25968031/fccc27a597b9457282c15f12ec470f068e1679a24774e8fc47a84a93da646fcf.txt)

dnkolegov-ar:
[fccc27a597b9457282c15f12ec470f068e1679a24774e8fc47a84a93da646fcf.txt](https://github.com/user-attachments/files/26024591/fccc27a597b9457282c15f12ec470f068e1679a24774e8fc47a84a93da646fcf.txt)

Reproduced the same behavior on `1955ba448b0f56d595ee7a3e78f1fda7c5fb6b05`:
```
cargo fuzz run range_proof fccc27a597b9457282c15f12ec470f068e1679a24774e8fc47a84a93da646fcf.txt
```

dnkolegov-ar:
`cargo fuzz run proof_store fc5bdef56c8c6f01921911d05bf9fe018d119393d74b06bbb9d20da1d19f575d.txt`

[fc5bdef56c8c6f01921911d05bf9fe018d119393d74b06bbb9d20da1d19f575d.txt](https://github.com/user-attachments/files/26024711/fc5bdef56c8c6f01921911d05bf9fe018d119393d74b06bbb9d20da1d19f575d.txt)

Issue #3384, opened 2026-03-13
[storage] Assertion failed in proofs_malleability.rs:147


dnkolegov-ar:
This is reproduced on Linux only. I used:
```
Linux 6.8.0-64-generic #67-Ubuntu SMP PREEMPT_DYNAMIC x86_64 GNU/Linux
```

[a4380acc3620ae82c5ff7ffaabc69447f572ffde6bcb8b0b45e3b43971e68d7e.txt](https://github.com/user-attachments/files/25965958/a4380acc3620ae82c5ff7ffaabc69447f572ffde6bcb8b0b45e3b43971e68d7e.txt)

```
thread '<unnamed>' (2284423) panicked at storage/fuzz/fuzz_targets/proofs_malleability.rs:147:25:
assertion failed: !mutated_proof.verify_element_inclusion(&mut hasher, element, loc, root)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Steps to reproduce (Linux):
1. Download the repro.
1. `cd storage/fuzz && cargo fuzz run proofs_malleability a4380acc3620ae82c5ff7ffaabc69447f572ffde6bcb8b0b45e3b43971e68d7e.txt`

FuzzInput:
```
{
    proof: Mmr,

    mutations: [
        ReplaceByte { position: 29240, value: 0x28 (40)  },
        DeleteByte  { position: 13068                    },
        ReplaceByte { position: 46872, value: 0x79 (121) },
        InsertByte  { position: 27595, value: 0x8D (141) },
        ReplaceByte { position: 12062, value: 0xF9 (249) },
        InsertByte  { position: 12326, value: 0x29 (41)  },
        ReplaceByte { position: 45783, value: 0x07 (7)   },
        SwapBytes   { pos1: 29765, pos2:  8311           },
        SwapBytes   { pos1: 37513, pos2: 23534           },
        InsertByte  { position: 58659, value: 0xD3 (211) },
        SwapBytes   { pos1: 53619, pos2:  1921           },
        DeleteByte  { position: 34928                    },
    ],

    positions: [
        124, 218,  59,  10, 117,  57, 182, 107,
         87, 127, 175, 108, 117, 216, 120, 222,
        195, 203, 206, 100,  60, 103,  72,  44,
        255,   0,   0,   0,   0,   0,   0,   0,
          0,   0,   0,   0,   0,   0,   0,   0,
          0,   0,   0,   0,
    ],
[...]

PR #3392, opened 2026-03-13
[storage/mmr] hash leaf count in last to avoid malleability concern

roberto-bayardo:
Fixes issue from the change introducing the folded root computation.

resolves: https://github.com/commonwarexyz/monorepo/issues/3384
resolves: https://github.com/commonwarexyz/monorepo/issues/3385

### DIFF
--- a/storage/src/merkle/mmr/hasher.rs
+++ b/storage/src/merkle/mmr/hasher.rs
@@ -44,6 +44,10 @@ pub trait Hasher: Clone + Send + Sync {
     }
 
     /// Computes the root for an MMR given its leaf count and peak digests in canonical order.
+    ///
+    /// The root digest is computed as `Hash(leaves || fold(peak_digests))`, where `fold` is
+    /// defined as `fold(acc, peak) = Hash(acc || peak)`. The `peak_digests` are assumed to be
+    /// in canonical order.
     fn root<'a>(
         &mut self,
         leaves: Location,
--- a/storage/src/merkle/proof.rs
+++ b/storage/src/merkle/proof.rs
@@ -654,8 +654,7 @@ where
     // Fold prefix peaks into a single accumulator (without the leaf count, which is always
     // hashed into the final root independently).
     if !bp.fold_prefix.is_empty() {
-        let first = get_node(bp.fold_prefix[0]).ok_or(Error::ElementPruned(bp.fold_prefix[0]))?;
-        let mut acc = first;
+        let mut acc = get_node(bp.fold_prefix[0]).ok_or(Error::ElementPruned(bp.fold_prefix[0]))?;
         for &pos in &bp.fold_prefix[1..] {
             let d = get_node(pos).ok_or(Error::ElementPruned(pos))?;
             acc = hasher.fold(&acc, &d);


storage/src/merkle/mmr/hasher.rs
```diff
@@ -43,18 +43,19 @@ pub trait Hasher: Clone + Send + Sync {
...
     }
 
-    /// Computes the root for an MMR given its size and an iterator over the digests of its peaks in
-    /// decreasing order of height.
+    /// Computes the root for an MMR given its leaf count and peak digests in canonical order.
     fn root<'a>(
         &mut self,
         leaves: Location,
         peak_digests: impl IntoIterator<Item = &'a Self::Digest>,
     ) -> Self::Digest {
```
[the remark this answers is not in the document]
roberto-bayardo: it's not ambiguous. The peak iterator defines it clearly.

storage/src/merkle/proof.rs
```diff
@@ -640,10 +652,12 @@ where
     let mut digests =
         Vec::with_capacity(if bp.fold_prefix.is_empty() { 0 } else { 1 } + bp.fetch_nodes.len());
 
-    // Fold prefix peaks into a single accumulator.
+    // Fold prefix peaks into a single accumulator (without the leaf count, which is always
+    // hashed into the final root independently).
     if !bp.fold_prefix.is_empty() {
-        let mut acc = hasher.digest(&leaves.to_be_bytes());
-        for &pos in &bp.fold_prefix {
+        let first = get_node(bp.fold_prefix[0]).ok_or(Error::ElementPruned(bp.fold_prefix[0]))?;
```
patrick-ogrady: just make `let mut acc`?

roberto-bayardo: fixed.

patrick-ogrady: You did one of them (but not the one I was highlighting)!

storage/src/merkle/proof.rs
```diff
@@ -512,7 +518,12 @@ impl<D: Digest> Proof<D> {
             return Err(ReconstructionError::ExtraDigests);
         }
 
-        Ok(acc)
+        // Hash the leaf count into the final result.
+        Ok(if let Some(peaks_acc) = acc {
+            hasher.hash([self.leaves.to_be_bytes().as_slice(), peaks_acc.as_ref()])
```
patrick-ogrady, line 523: Is this leaf inclusion documented anywhere? Maybe we should note it here?

```
/// The `digests` vector uses a fold-based layout:
///
/// 1. If there are peaks entirely before the proven range (fold prefix), the first digest is
///    a single accumulator produced by folding those peaks: `fold(fold(..., peak0), peak1)`.
///    If there are no such peaks, this entry is absent.
///
/// 2. The digests of peaks entirely after the proven range, in peak iteration order.
///
/// 3. The sibling digests needed to reconstruct each range-peak digest from the proven elements,
///    in depth-first (forward consumption) order for each range peak.
```

roberto-bayardo: You mean the leaf count inclusion?   That's part of the root generation process, and it isn't part of the proof. I added a comment better documenting the root generation in hash.rs which should cover this concern (if I understand it correctly).

```
    /// Computes the root for an MMR given its leaf count and peak digests in canonical order.
    ///
    /// The root digest is computed as `Hash(leaves || fold(peak_digests))`, where `fold` is
    /// defined as `fold(acc, peak) = Hash(acc || peak)`. The `peak_digests` are assumed to be
    /// in canonical order.
    fn root<'a>(
    ...
```

patrick-ogrady: Exactly, thanks.

patrick-ogrady (approved):

patrick-ogrady:
fuzzing ftw

PR #3394, opened 2026-03-14
[codec] Add implementations for `Range*<T>`

clabby:
## Overview

Adds codec implementations for `Range*<T>` types.

### DIFF
--- a/codec/src/types/range.rs
+++ b/codec/src/types/range.rs
@@ -5,9 +5,13 @@ use crate::{EncodeSize, Error, FixedSize, Read, Write};
 use bytes::{Buf, BufMut};
 use core::ops::{Range, RangeFrom, RangeFull, RangeInclusive, RangeTo, RangeToInclusive};
 
-impl<T: Write> Write for Range<T> {
+impl<T: Write + PartialOrd> Write for Range<T> {
     #[inline]
     fn write(&self, buf: &mut impl BufMut) {
+        assert!(
+            self.start.partial_cmp(&self.end).is_some_and(|o| o.is_le()),
+            "start must be <= end"
+        );
         self.start.write(buf);
         self.end.write(buf);
     }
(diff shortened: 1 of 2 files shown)


codec/src/types/range.rs
```diff
@@ -0,0 +1,272 @@
...
+}
+
+impl<T: Read + PartialOrd> Read for Range<T> {
+    type Cfg = T::Cfg;
+
+    #[inline]
+    fn read_cfg(buf: &mut impl Buf, cfg: &Self::Cfg) -> Result<Self, Error> {
+        let start = T::read_cfg(buf, cfg)?;
+        let end = T::read_cfg(buf, cfg)?;
+        if start > end {
```
patrick-ogrady: I suppose in rust start > end means empty?

That being said, I like the idea of being strict?

codec/src/types/range.rs
```diff
@@ -0,0 +1,272 @@
...
+            return Err(Error::Invalid("Range", "start must be <= end"));
+        }
+        Ok(start..end)
+    }
+}
+
+impl<T: Write> Write for RangeInclusive<T> {
```
patrick-ogrady: Not sure how you feel about this:

```
RangeInclusive stores more than just
  start and end; it also has an internal exhausted state. For example, an exhausted 3..=3 and a live 3..=3 expose the same endpoints, but
  is_empty() differs. Since the stable public API does not let us reconstruct that hidden state generically on decode, keeping the codec would
  mean silently changing valid values. The safe fix was to drop that impl rather than keep a non-faithful codec.
```

codec/src/types/range.rs
clabby: The regular range types are allowed to be empty. Though what we want to prevent is ranges that make no sense, i.e. `10..5` - anything that would fail https://rust-lang.github.io/rust-clippy/rust-1.93.0/index.html#reversed_empty_ranges.

codec/src/types/range.rs
clabby: This is a known footgun when dealing with `RangeInclusive`, though I don't think it should block us from supporting a codec for it. It's well documented, and most of the time, we'll be constructing this range and encoding it immediately.

To hit the footgun, you'd need to do something like:

```rs
let mut r = 0..=5;
dbg!(r.start(), r.end()); // 0, 5
(&mut r).for_each(|_| ());
dbg!(r.start(), r.end()); // 5, 5
// ... then encode `r` ...   

let mut r = 0..=5;
dbg!(r.start(), r.end()); // 0, 5
r.next().unwrap();
dbg!(r.start(), r.end()); // 1, 5
// ... then encode `r` ...
```

FWIW, `serde` does support `Serialize` and `Deserialize` for this type: https://docs.rs/serde/latest/serde/trait.Serialize.html#impl-Serialize-for-RangeInclusive%3CIdx%3E

codec/src/types/range.rs
```diff
@@ -0,0 +1,272 @@
...
+    #[inline]
+    fn write(&self, buf: &mut impl BufMut) {
```
patrick-ogrady, line 43: Should we panic on start > end? IIRC we panic when `usize` doesn't fit in `u32`?

PR #3395, opened 2026-03-14
[utils] Add `NonEmptyRange` type

clabby:
## Overview

Adds a `NonEmptyRange` type to `utils` and uses it in `commonware-storage`'s `Target`.

### DIFF
--- a/storage/src/qmdb/sync/target.rs
+++ b/storage/src/qmdb/sync/target.rs
@@ -144,10 +144,7 @@ mod tests {
             Target::<sha256::Digest>::read(&mut cursor),
             Err(CodecError::Invalid("Range", "start must be <= end"))
         ));
-    }
 
-    #[test]
-    fn test_sync_target_read_empty_range() {
         // Manually encode a target with an empty range (start == end)
         let root = sha256::Digest::from([42; 32]);
         let mut buffer = Vec::new();


storage/fuzz/fuzz_targets/qmdb_any_fixed_sync.rs
```diff
@@ -229,7 +229,7 @@ fn fuzz(mut input: FuzzInput) {
                     db.commit().await.expect("Commit should not fail");
                     let target = sync::Target {
                         root: db.root(),
-                        range: db.inactivity_floor_loc()..db.bounds().await.end,
+                        range: non_empty_range!(db.inactivity_floor_loc(), db.bounds().await.end),
```
patrick-ogrady, line 232: I suppose this means we can never sync an "empty" database -> not sure if that is a good idea or not (it is possible to have some DB with no active keys at some non-zero ops)?

clabby: This is how it behaves currently (just with oob runtime checks rather than having a typed non-empty range.)

Though that said, if I'm reading the impl correctly, even a database with no active keys and non-zero ops has a range of `inactivity_floor..inactivity_floor + 1` due to the `CommitFloor` op at the tip, appended after the inactivity floor?

https://github.com/commonwarexyz/monorepo/blob/276a027c3cd07d59faab33bdf6210c1cff7c877a/storage/src/qmdb/any/mod.rs#L223-L228

https://github.com/commonwarexyz/monorepo/blob/276a027c3cd07d59faab33bdf6210c1cff7c877a/storage/src/qmdb/any/batch.rs#L441-L448

utils/src/range.rs
```diff
@@ -0,0 +1,203 @@
...
+        self.0.fmt(f)
+    }
+}
+
+impl<Idx: PartialOrd> NonEmptyRange<Idx> {
+    /// Creates a `NonEmptyRange` if `start < end`.
+    pub fn new(range: Range<Idx>) -> Result<Self, EmptyRange> {
+        (range.start < range.end)
+            .then_some(Self(range))
+            .ok_or(EmptyRange)
```
patrick-ogrady, line 27: Usually, we'd panic on this?

storage/fuzz/fuzz_targets/qmdb_any_fixed_sync.rs
patrick-ogrady: You are correct 👍 ... can you do a quick scan to ensure we have codified this behavior?

utils/src/range.rs
clabby: Following the API of [`std::num::NonZero::new`](https://doc.rust-lang.org/stable/std/num/struct.NonZero.html#method.new), `new` is fallible here. Though `std`'s `NonZero` returns `Option` rather than `Result`... I opted for `Result` here to share the same error with `TryFrom`.

examples/sync/src/bin/server.rs
```diff
@@ -171,7 +172,7 @@ where
         request_id: request.request_id,
         target: Target {
             root,
-            range: inactivity_floor..size,
+            range: non_empty_range!(inactivity_floor, size),
```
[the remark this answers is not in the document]
clabby: The `CommitFloor` op guards this.

storage/src/qmdb/sync/target.rs
```diff
@@ -63,7 +57,7 @@ where
         let upper = u.int_in_range(lower + 1..=*MAX_LOCATION)?;
         Ok(Self {
             root,
-            range: Location::new(lower)..Location::new(upper),
+            range: commonware_utils::non_empty_range!(Location::new(lower), Location::new(upper)),
```
danlaine, line 60: nit: we import `non_empty_range!` instead of qualifying it everywhere else
```suggestion
            range: non_empty_range!(Location::new(lower), Location::new(upper)),
```

storage/src/qmdb/sync/target.rs
```diff
@@ -133,19 +132,7 @@ mod tests {
...
-            range: Location::new(100)..Location::new(50), // invalid: lower > upper
-        };
-
-        let mut buffer = Vec::new();
-        target.write(&mut buffer);
-    }
-
-    #[test]
-    fn test_sync_target_decode_invalid_bounds() {
+    fn test_sync_target_read_invalid_bounds() {
```
danlaine, line 135: nit: this test and `test_sync_target_read_empty_range` do the same thing -- perhaps we could combine them into one test with two cases (start == end and start < end)?

storage/src/qmdb/sync/target.rs
clabby: The only reason we qualify it here is because it’s the only usage in the file and it’s behind the `arbitrary` feature.

We could similarly add a `use` within the function like we do for `MAX_LOCATION` on L54?

danlaine: I'm fine keeping this as is 👍

PR #3397, opened 2026-03-15
[consensus/simplex] Only Forward After Certification

patrick-ogrady:
Related: #3347

### DIFF
--- a/consensus/src/simplex/actors/voter/actor.rs
+++ b/consensus/src/simplex/actors/voter/actor.rs
@@ -1076,6 +1076,9 @@ impl<
                         .state
                         .leader_index(current_view)
                         .expect("leader not set");
+
+                    // If we skip a view, we don't worry about forwarding our latest certified proposal
+                    // because the network has already moved on
                     let forwardable_proposal = current_view
                         .previous()
                         .and_then(|view| self.state.forwardable_proposal(view));


consensus/src/simplex/actors/batcher/round.rs
```diff
@@ -41,8 +41,10 @@ pub struct Round<
     /// Only these votes are used for certificate construction.
     verified_votes: VoteTracker<S, D>,
 
-    /// Whether we've already received and forwarded the leader's proposal.
+    /// Whether we've already forwarded the leader's proposal to the voter.
     proposal_sent: bool,
+    /// Whether we've already forwarded the certified block to silent peers.
+    block_forwarded: bool,
```
patrick-ogrady: If we only attempt when entering the view, we don't need this?

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -295,6 +313,17 @@ where
...
+                            let me = self.scheme.me()?;
+                            let round = work.get_mut(&view)?;
+                            let proposal = round.take_forwarding_proposal(me)?;
+                            let participants = self.forwarding_targets(round, &proposal, leader);
+                            Some((proposal, participants))
+                        })
+                    } else {
+                        None
+                    };
                     let am_leader = self.scheme.me().is_some_and(|me| me == leader);
```
patrick-ogrady, line 319: use `.me()` and then get inside of forward?

consensus/src/simplex/actors/batcher/round.rs
```diff
@@ -300,6 +303,20 @@ impl<
         Some(proposal)
     }
 
+    /// Returns the locally certified proposal eligible for forwarding.
+    ///
+    /// The local finalize vote is the batcher's signal that certification
+    /// succeeded. This is only emitted once per round, when we first enter the
+    /// next view after constructing that vote.
+    pub fn take_forwarding_proposal(&mut self, me: Participant) -> Option<Proposal<D>> {
```
patrick-ogrady: We only enter a round once, so we don't need this?

consensus/src/simplex/actors/voter/mod.rs
```diff
@@ -6164,17 +6167,42 @@ mod tests {
...
-                                reached_view2 = true;
+                                break true;
                             }
+                        }
+                        _ => {}
+                    },
+                    _ = context.sleep(Duration::from_millis(10)) => {},
+                }
+            };
+            assert!(!reached_view2, "view advanced before finalize for view 1");
```
patrick-ogrady, line 6194: This becomes important to enshrine if we take this "detect finalize" approach here (which isn't "that" crazy imho)?

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -323,6 +342,23 @@ where
                         current.timed_out = true;
                     }
                     response.send_lossy(timeout_reason);
+                    if self.forwarding.is_enabled() {
```
patrick-ogrady: If we introduce an explicit "forward" message, we then need extra machinery to support the leader broadcast (and would otherwise need to assume it is known by the time we certify).

patrick-ogrady: Its nice to align this with "view change", just a bit weird to know we sent a finalize vote.

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -218,6 +218,24 @@ where
...
+    fn forwarding_targets(
+        &self,
+        round: &Round<S, B, D, Re>,
+        proposal: &Proposal<D>,
+        next_leader: Participant,
+    ) -> Vec<Participant> {
+        match self.forwarding {
+            ForwardingPolicy::Disabled => Vec::new(),
+            ForwardingPolicy::Silent => round.missing_voters(proposal),
+            ForwardingPolicy::NextLeader => round
```
patrick-ogrady, line 233: This can be done more efficiently than getting all missing voters and filtering.

consensus/src/simplex/actors/batcher/actor.rs
patrick-ogrady: Break out into own code block with comment.

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -211,13 +211,33 @@ where
     /// Resolves the public keys of `missing` participants for a targeted
     /// block forward. Returns `None` when no peers need the block.
     fn resolve_peers(&self, missing: &[Participant]) -> Option<Vec<S::PublicKey>> {
+        let me = self.scheme.me();
         let peers: Vec<_> = missing
             .iter()
+            .filter(|&&p| Some(p) != me)
```
patrick-ogrady, line 217: Add a regression test that ensures we never send to ourselves

consensus/src/simplex/actors/voter/mod.rs
```diff
@@ -6038,7 +6052,9 @@ mod tests {
     ///
     /// We require:
     /// 1. No `nullify(1)` is emitted while quorum certificates arrive promptly.
-    /// 2. The voter emits `notarize(1)` and `finalize(1)`, then advances to view 2.
+    /// 2. The voter emits `notarize(1)`.
+    /// 3. After successful certification, the voter emits `finalize(1)` before
```
patrick-ogrady, line 6056: This is no longer needed for the correctness of this PR but figure it is worth adding a test for to detect a subtle change here.

consensus/src/simplex/mod.rs
```diff
@@ -1986,7 +1986,9 @@ mod tests {
                 // Ensure every reporter observes finalization progress to at least the target view.
                 {
                     let finalizations = reporter.finalizations.lock();
-                    assert!(finalizations.keys().any(|view| *view >= required_containers));
```
BrendanChou, line 1989: nit: undo this change since I think nothing changed?

consensus/src/simplex/actors/batcher/round.rs
```diff
@@ -353,7 +353,31 @@ impl<
...
+        let has_notarize = self
+            .pending_votes
+            .notarize(participant)
+            .is_some_and(|vote| &vote.proposal == proposal);
+        let has_finalize = self
+            .pending_votes
+            .finalize(participant)
+            .is_some_and(|vote| &vote.proposal == proposal);
+        !has_notarize && !has_finalize
+    }
```
BrendanChou, line 377: tiny performance nit: not necessary to keep checking `has_finalize` if `has_notarize` is true, which we expect to be true most of the time.

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -425,6 +425,15 @@ impl<E: Clock + CryptoRngCore + Metrics, S: Scheme<D>, L: ElectorConfig<S>, D: D
         self.views.get(&view).and_then(|round| round.finalization())
     }
 
+    /// Returns the proposal for `view` if it is eligible for forwarding.
+    pub fn forwardable_proposal(&self, view: View) -> Option<Proposal<D>> {
+        let round = self.views.get(&view)?;
+        if round.finalization().is_some() || round.is_certified() {
```
BrendanChou, line 431: I think `is_certified()` should also probably return `true` if it is finalized... since there are many places where it's okay that finalized implicitly implies that it is certified

consensus/src/simplex/mod.rs
patrick-ogrady: I think this is a fmt issue that wasn't addressed in the base PR.

consensus/src/simplex/actors/batcher/round.rs
patrick-ogrady: I thought about this but was lazy. Will fix.

consensus/src/simplex/actors/voter/state.rs
patrick-ogrady: I prefer to keep it explicit (if we couple them again, the voting after only explicit certify won't work).

PR #3400, opened 2026-03-16
[Storage] Allow batches to live longer than their base

danlaine:
## Summary

This PR changes the batch API across `commonware-storage` (MMR, authenticated journal, QMDB) from a borrow-based (`&'a parent`) design to an owned, `Arc`-backed persistent data structure. Speculative batches can now outlive mutations to the base MMR/QMDB they were forked from.

## Motivation

Previously, `UnmerkleizedBatch` and `MerkleizedBatch` borrowed their parent immutably via a lifetime parameter `'a`. This meant:

1. **Exclusivity**: The base MMR/DB could not be mutated (e.g., committing one batch) while any sibling batch still existed, because Rust's borrow checker prevented it.
2. **Rigid lifetimes**: Batches had to be dropped in reverse creation order. Callers needed careful scoping (`drop(m2)` before applying `m1`) and couldn't store batches in long-lived collections.
3. **Growing type parameter**: Stacking batches (Base <- A <- B) produced deeply nested generic types (`MerkleizedBatch<'a, D, MerkleizedBatch<'b, D, Mmr<D>>>`) that were impossible to store in homogeneous collections like `HashMap<Digest, MerkleizedBatch>`.

These constraints prevented speculative execution patterns where multiple batches need to coexist independently -- for example, maintaining speculative state for several candidate blocks while committing the winning one.

### In-memory MMR (`mmr::batch`, `mmr::mem`)

The single generic `Batch<'a, D, P, S>` (parameterized by parent type `P` and state `S`) is replaced by two distinct types with no lifetime or parent type parameter:

- `UnmerkleizedBatch<D>` -- mutable accumulator, holds a `MerkleizedBatch<D>` parent (owned).
- `MerkleizedBatch<D>` -- immutable enum with three variants. Clone is O(1).
  - `Base(Mmr<D>)` -- a committed MMR. `Mmr` is already `Arc`-wrapped internally, so cloning is O(1).
  - `Layer(Arc<MerkleizedBatchLayer<D>>)` -- speculative mutations on top of a parent.
  - `Snapshot { inner, size }` -- seals a chain so that `base_size() == size()`. Created by `to_batch()`. Without `Snapshot`, a batch forked from committed state after `push_changeset` would recurse through the new `Layer` all the way to the original `Base`, yielding `base_size() == 0`. The resulting changeset would fail the stale-changeset check in `apply()`.

The `State<D>` trait, `Clean<D>` marker type, and `Dirty` struct are removed. Dirty node tracking is inlined into `UnmerkleizedBatch` as a plain `BTreeSet`.

Key new methods:

- `finalize_from(current_base)` -- produce a changeset relative to a base that has advanced past the original fork point. Use when an ancestor batch has already been committed; plain `finalize()` would target the original base size which `apply()` rejects as stale.
- `push_changeset(cs)` -- apply a changeset as a new `Layer` in O(1). Used internally by the journaled MMR and current QMDB.
- `flatten()` -- collapse all accumulated layers into a single `Base`. See [Layer accumulation](#layer-accumulation-and-flatten).
- `Mmr::to_batch()` -- create an owned `MerkleizedBatch` snapshot of committed state. O(1) because `Mmr` is internally `Arc`-wrapped.

`Mmr::new_batch()` now clones the `Mmr` (O(1) Arc clone) instead of borrowing it.

### Journaled MMR (`mmr::journaled`)

The internal `Inner` struct now stores `MerkleizedBatch<D>` instead of a raw `MemMmr<D>`. After `apply()`, the state may be a `Layer` chain. After `sync()` or `prune()`, it is rebuilt as a `Base` with only O(log N) pinned peaks.

A new `UnmerkleizedBatch` wrapper hides `update_leaf*` at compile time. The journaled MMR's `sync()` only persists *appended* nodes; overwrites to existing positions would be silently lost on crash recovery. The wrapper prevents this by exposing only append and merkleize operations.

`Mmr::truncate()` is removed -- `rewind()` now syncs first (to flush un-synced nodes and collapse layers) then rebuilds from journal/metadata.

### Authenticated journal (`journal::authenticated`)

`MerkleizedBatch<D, Item>` is now owned (no lifetime). Items are stored as `Vec<Arc<Vec<Item>>>` -- one segment per batch in the chain, with Arc sharing so that forking is O(1).

Removed: `BatchChain<Item>` and `BatchChainInfo` traits. Item collection was previously recursive through the parent chain via these traits; now it's a flat `Vec` of Arc segments.

New: `finalize_from(current_base, items_to_skip)` skips already-committed item segments when producing a changeset.

### Any QMDB (`qmdb::any::batch`)

Both `UnmerkleizedBatch` and `MerkleizedBatch` are now fully owned and lifetime-free.

`UnmerkleizedBatch` drops from 6 type parameters to 2: `UnmerkleizedBatch<H, U>`. The lifetime `'a` and runtime/journal/index type parameters (`E`, `C`, `I`) are removed along with the `db` field. Methods that need the committed DB (`get()`, `merkleize()`) accept `&Db` as an explicit parameter instead of storing it as a field. This decouples the batch from the DB's lifetime, enabling patterns like `Arc<RwLock<Db>>` where batches must outlive lock guards.

`MerkleizedBatch` drops to 2 type parameters: `MerkleizedBatch<D, U>`. `MerkleizedBatch::new_batch()` no longer takes `&Db` (since the child `UnmerkleizedBatch` no longer stores it). `MerkleizedBatch::get()` continues to take `&Db` as before.

New: `finalize_from(current_db_size)` rebases snapshot diffs onto the current committed DB state, resolving `old_loc` values against committed ancestor operations. New: `Db::to_batch()` creates an initial owned `MerkleizedBatch` from committed state.

The `db`, `operation`, and `value` submodules are now `pub` (were `pub(crate)`) so that external crates can name `Db`, `Operation`, `ValueEncoding`, `FixedValue`, and `VariableValue` types directly. The sealed `ValueEncoding` trait is re-exported as `pub`.

The test-only `BatchableDb` trait no longer uses a GAT for its `Batch` associated type (the lifetime parameter was only needed because `UnmerkleizedBatch` had one). The `UnmerkleizedBatch` trait gains a `Db: ?Sized` generic parameter so that `merkleize` can accept `&Db`.

### Immutable QMDB (`qmdb::immutable::batch`)

Same pattern as Any: lifetime and runtime type parameters removed from `UnmerkleizedBatch`. `UnmerkleizedBatch<H, K, V>` -- 3 type params (was 6). `get()` takes `&Immutable` explicitly. `merkleize()` is synchronous and does not need `&Db`.

`MerkleizedBatch` is owned (`MerkleizedBatch<D, K, V>`). `MerkleizedBatch::new_batch()` no longer takes `&Immutable`.

New: `finalize_from(current_db_size)` and `Immutable::to_batch()`.

### Keyless QMDB (`qmdb::keyless::batch`)

Same pattern: lifetime and runtime type parameter removed from `UnmerkleizedBatch`. `UnmerkleizedBatch<H, V>` -- 2 type params (was 4). `get()` takes `&Keyless` explicitly. `merkleize()` is synchronous and does not need `&Db`.

`MerkleizedBatch` is owned (`MerkleizedBatch<D, V>`). `MerkleizedBatch::new_batch()` no longer takes `&Keyless`.

New: `finalize_from(current_db_size)` and `Keyless::to_batch()`.

### Current QMDB (`qmdb::current`)

Same batch API changes as Any (wraps `any::batch` types). `UnmerkleizedBatch<H, U, N>` -- 3 type params + const (was 7 + const). `get()` and `merkleize()` take `&Db` explicitly. `MerkleizedBatch::new_batch()` no longer takes `&Db`.

The bitmap (`BitMap<N>`) and grafted MMR are now stored as layered COW structures (`BitmapBatch<N>` and `MerkleizedBatch<D>`). `apply_batch` pushes mutations as layers in O(changeset) instead of mutating flat structures in O(total size).

Bitmap chunk pruning is deferred from `apply_batch` to `prune()`. Previously `apply_batch` eagerly pruned after every batch. Now layers accumulate until `prune()`, which flattens first then prunes. Deferring is safe because bitmap pruning only drops fully-inactive prefix chunks below the inactivity floor -- correctness (roots, proofs, floor scans) is unaffected.

New: `Db::flatten()` collapses bitmap and grafted MMR layers without pruning. Callers that apply many batches between prune boundaries should call this periodically. `prune()` calls `flatten()` as a prerequisite.

## Removed APIs

- `BatchChain<Item>` and `BatchChainInfo` traits
- `State<D>` trait, `Clean<D>` type, `Dirty` struct
- `MerkleizedBatch::into_dirty()` -- use `merkleized.new_batch()` instead
- `Mmr::truncate()` -- journaled MMR rebuilds from journal/metadata after sync
- `Batch<'a, D, P, S>` struct

## Performance

Batch creation, clone, node lookup, and `finalize()` are asymptotically unchanged. `finalize_from(current_base)` does strictly less work when an ancestor has been committed.

### `sync()`: O(U + log L) rebuild

Previously pruned the in-memory MMR in place. Now discards the (possibly layered) state and rebuilds a flat `Base` from pinned peaks. Avoids accumulating `Layer` chains in the long-running journaled MMR.

### `push_changeset()`: O(1)

Wraps the old state as the parent of a new `Layer`. Replaces what was previously an O(N) in-place mutation on `MemMmr`.

### Layer accumulation and `flatten()`

Because `push_changeset` is O(1), layers accumulate with each `apply_batch` / `apply` call. Node lookups degrade to O(k) where k is the chain depth.

- **Journaled MMR**: `sync()` automatically collapses layers. No caller action needed.
- **Current QMDB**: Call `Db::flatten()` periodically, or rely on `prune()` which flattens as a prerequisite.

`flatten()` costs: O(total retained nodes) for the MMR, O(total bits) for the bitmap.

### Memory

Each `Layer` is heap-allocated behind an `Arc`. `flatten()`, `sync()`, or `prune()` collapses layers back to a flat `Base`, freeing intermediate layers when all Arc references are dropped.


storage/src/qmdb/any/batch.rs
```diff
@@ -118,24 +114,26 @@ pub(crate) enum SnapshotDiff<K> {
...
-/// computed, in contrast to [MerkleizedBatch].
-#[allow(clippy::type_complexity)]
-pub struct UnmerkleizedBatch<'a, E, C, I, H, U, P>
+/// A speculative batch of operations whose root digest has not yet been computed,
+/// in contrast to [`MerkleizedBatch`].
+///
+/// Borrows `&Db` for reads during the build phase. Consuming [`UnmerkleizedBatch::merkleize`]
+/// (or its ordered/unordered variants) produces an owned [`MerkleizedBatch`] and releases the
+/// borrow.
+pub struct UnmerkleizedBatch<'a, E, C, I, H, U>
```
clabby: Is there a way we can also get rid of the lifetime here? It complicates the interfaces of the traits a bit: https://github.com/commonwarexyz/monorepo/blob/458c8623294b1d5cdebe613cb762d775a6e58890/glue/src/stateful/db.rs#L93-L94
https://github.com/commonwarexyz/monorepo/blob/458c8623294b1d5cdebe613cb762d775a6e58890/glue/src/stateful/db.rs#L106-L110

The way I'm currently expecting to use the databases is to have an `Arc<RwLock<DB>>` for each QMDB instance. While we can express the lifetime on the `UnmerkleizedBatch` types via GATs, it falls apart when we want to have a trait that describes how to create an unmerkleized batch (either from an existing `MerkleizedBatch`, or directly off of the DB) due to being unable to return data borrowed in the current function.

storage/src/qmdb/any/mod.rs
clabby: In order to implement a trait on the database externally, this needs to be public.

clabby: Same (to be able to name `Operation`, `update::Ordered`, & `update::Unordered`.)

clabby: Same (required to name the `ValueEncoding`, `FixedValue`, and `VariableValue` for the `UnmerkleizedBatch` & DB).

storage/src/merkle/mmr/batch.rs
```diff
@@ -399,114 +379,297 @@ impl<'a, D: Digest, P: Readable<Digest = D>> UnmerkleizedBatch<'a, D, P> {
...
+        let size = self.size();
+        let pruned_to_pos = self.pruned_to_pos();
+
+        // Collect pinned nodes (peaks at the prune boundary).
+        let mut pinned_nodes = BTreeMap::new();
+        for pos in crate::mmr::iterator::nodes_to_pin(pruned_to_pos) {
+            if let Some(d) = self.get_node(pos) {
+                pinned_nodes.insert(pos, d);
+            }
+        }
```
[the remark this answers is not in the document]
danlaine: added `.expect()`

storage/src/qmdb/any/batch.rs
danlaine: Removed lifetime

storage/src/merkle/mmr/batch.rs
```diff
@@ -68,46 +68,49 @@ cfg_if::cfg_if! {
...
 }
 
-/// A batch whose root digest has not been computed.
-pub type UnmerkleizedBatch<'a, D, P> = Batch<'a, D, P, Dirty>;
-
-/// A batch whose root digest has been computed.
-pub type MerkleizedBatch<'a, D, P> = Batch<'a, D, P, Clean<D>>;
+impl<D: Digest> UnmerkleizedBatch<D> {
+    /// Create a new batch from `parent`.
+    /// O(1) time and space (the parent is an `Arc`-backed persistent structure).
```
roberto-bayardo: Weird comment, I wouldn't expect it to be anything but O(1) in time and space?

storage/src/merkle/mmr/batch.rs
```diff
@@ -393,114 +370,298 @@ impl<'a, D: Digest, P: Readable<Digest = D>> UnmerkleizedBatch<'a, D, P> {
...
+/// `Clone` is O(1) (`Arc::clone`). Batches can be freely shared and stored in homogeneous
+/// collections regardless of chain depth.
+#[derive(Clone, Debug)]
+pub enum MerkleizedBatch<D: Digest> {
+    /// Committed MMR (chain terminal). `Mmr` is already `Arc`-wrapped internally,
+    /// so cloning is O(1).
+    Base(Mmr<D>),
+    /// Speculative layer on top of a parent batch.
+    Layer(Arc<MerkleizedBatchLayer<D>>),
+    /// Sealed snapshot of a (possibly layered) chain.
```
roberto-bayardo: Not following most of the comments for Snapshot here.

storage/src/merkle/mmr/batch.rs
```diff
@@ -1,65 +1,65 @@
...
-//!                           Mmr             (updated in place)
-//! ```
-//!
-//! # Type aliases
-//!
-//! - [`UnmerkleizedBatch`] -- mutable phase: add, update leaves.
-//! - [`MerkleizedBatch`]   -- immutable phase: root is computed, proofs available.
-//! - [`Changeset`]         -- owned delta that can be applied to the base MMR.
+//! - [`Changeset`] -- owned delta that can be applied to the base MMR.
+//! - [`MerkleizedBatch::Snapshot`] -- a sealed view where `base_size() == size()`, created by
```
roberto-bayardo: maybe elaborate on this a bit? I'm scratching my head a bit around what's this for (I'm sure it will click when I get deeper into the code, but would be nice if I could grok it just from this header).

storage/src/merkle/mmr/batch.rs
danlaine: Hmm, yeah, perhaps this is extraneous info. I tried to be explicit about where we do cheap vs potentially expensive (i.e.  O(changeset) or O(chain depth)) operations. I'll change it to just point out where we do something potentially expensive instead of where we do something cheap (i.e. here)

storage/src/merkle/mmr/batch.rs
danlaine: Renamed to `Checkpoint` (is that more clear? Unsure) and changed the comments to hopefully clarify this. Let me know what you think.

storage/src/merkle/mmr/batch.rs
danlaine: https://github.com/commonwarexyz/monorepo/pull/3400#discussion_r2976521906

Updated comments, hopefully this is more clear now

christopherwxyz:
Strong +1 on this change.

We're building a BFT consensus chain on top of commonware (Simplex + QMDB) and the borrow-based batch API is our single biggest source of ergonomic friction. Today we work around the lifetime constraints with:

- A `SyncDb` newtype + `unsafe impl Sync` to share QMDB across the consensus automaton and a gRPC query layer
- `tokio::task::block_in_place` + `Handle::current().block_on()` throughout our state store trait to avoid holding `&Db` across `.await` points
- Careful manual scoping to ensure batches are dropped before sibling batches can be committed

Owned batches eliminate all three of these workarounds. Specifically, `UnmerkleizedBatch` no longer storing `&Db` means our batch type can be `Send` without the `block_in_place` bridge, and `MerkleizedBatch` being storable in a `HashMap<Digest, MerkleizedBatch>` lets us cache speculative execution results cleanly instead of reconstructing them.

The `finalize_from(current_base)` API is also exactly what we need — we maintain multiple pending blocks during consensus and need to commit one while discarding the rest.

Looking forward to this landing.

PR #3408, opened 2026-03-16
[cryptography] Implement Scalar::from(SmallScalar)

cronokirby:
This is useful, because often random coefficients used in an MSM might need to be manipulated more extensively in other places.

### DIFF
--- a/cryptography/src/bls12381/primitives/group.rs
+++ b/cryptography/src/bls12381/primitives/group.rs
@@ -303,6 +303,18 @@ impl SmallScalar {
     }
 }
 
+impl From<SmallScalar> for Scalar {
+    fn from(small: SmallScalar) -> Self {
+        let mut fr = blst_fr::default();
+        // SAFETY: small.inner is a valid blst_scalar with only the lower 128 bits set,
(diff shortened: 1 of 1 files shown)


cryptography/src/bls12381/primitives/group.rs
```diff
@@ -303,6 +303,18 @@ impl SmallScalar {
     }
 }
 
+impl From<SmallScalar> for Scalar {
```
patrick-ogrady, line 306: I think we've currently defined `Scalar` still as non-zero (I suppose that only matters for encodable format)?

cronokirby: There's another loop-hole, namely `Scalar::zero`. In practice, you get `SmallScalar` from generating them randomly, so you won't ever see 0 ones anyways, so I don't think this is a pejoration.

patrick-ogrady: You've got me there. Oof.

patrick-ogrady: Related: #3376

Issue #3319, opened 2026-03-06
[runtime/iouring] high-level requests and multi-ring sharding


andresilva:
The current io_uring backend is built around low-level per-op submission. This is correct, but it likely leaves performance on the table for high-concurrency network/storage workloads. I think we should improve it in two separate ways.

### 1. Move from low-level ops to high-level ops

Today the caller effectively does:
- submit one low-level read/write op
- wait for completion
- if partial, resubmit from the caller
- repeat until the logical request is complete

This means we pay cross-thread submit/completion overhead and timeout bookkeeping per low-level op. Instead, the io_uring worker should own high-level logical requests such as:
- network:
    - `WriteAll`
    - `WriteVectoredAll`
    - `ReadExact`
- storage:
    - `ReadAtExact`
    - `WriteAtAll`
    - `Sync`
    - `Resize`

That would let the worker handle internally:
- partial progress
- retries
- cancellation
- one deadline per logical request

The caller would see one request and one completion, instead of one completion per fragment.

### 2. Run multiple rings instead of one global ring per role

Today the model is effectively split by role:
- storage ring
- network send ring
- network recv ring

For scaling to many peers/files, a better model is likely:
- spawn N network send io_uring workers
- spawn N network recv io_uring workers
- shard sockets across those workers
- keep ownership stable within each domain
- storage can remain on a separate ring initially

This avoids one global send/recv/storage ring becoming a bottleneck and keeps endpoint/request state local.

#### Important constraint

We should preserve affinity at the I/O ownership level, not at the whole actor/executor level.

In other words:
- sockets/files and their request state should belong to one io_uring worker
- protocol/application actors should not be forced onto the same thread as that worker

### Optional follow-up: pin io_uring workers to cores

Once the backend is sharded across multiple io_uring workers, we should evaluate optional CPU pinning for workers.

Potential benefits:
- better cache locality
- lower scheduling jitter
- more predictable latency under load

Related #2883.

PR #3411, opened 2026-03-17
[runtime/iouring] replace raw SQE submission with high-level requests

andresilva:
Replaces the low-level `Op` (with `OpBuffer`/`OpFd`/`OpIovecs`) submission model with high-level request types (`Send`, `Recv`, `ReadAt`, `WriteAt`, `Sync`). Each request carries its own completion sender and is converted into an `ActiveRequest` state machine inside the loop. The active request owns all resources (buffers, FDs, progress cursors) and handles partial progress, retries, and result delivery internally. This moves the responsibility for building SQEs, interpreting CQE results, and tracking multi-SQE progress from the network/storage frontends into the loop, making the frontends simple one-shot submit-and-await calls.

Depends on #3284.
Related #3319 #2883.

### DIFF
--- a/runtime/src/iouring/mod.rs
+++ b/runtime/src/iouring/mod.rs
@@ -2201,6 +2201,115 @@ mod tests {
     }
 
     #[tokio::test]
+    async fn test_fill_submission_queue_orphans_closed_request_before_first_submit() {
+        // Verify dropping the caller before the loop stages the first SQE
+        // retires both send and read-at requests locally instead of issuing
+        // any I/O.
+        let cfg = Config::default();
+        let mut registry = Registry::default();
+        let (handle, mut iouring) = IoUringLoop::new(cfg.clone(), &mut registry);
+        let mut ring = new_ring(&cfg).expect("unable to create io_uring instance");
+
+        // Keep the test focused on request staging instead of wake rearm.
+        iouring.wake_rearm_needed = false;
+
+        let (tx, rx) = oneshot::channel();
+        drop(rx);
+        handle
+            .enqueue(Request::Send(SendRequest {
+                fd: Arc::new(UnixStream::pair().unwrap().0.into()),
+                write: IoBufs::from(IoBuf::from(b"hello")).into(),
+                deadline: Some(Instant::now() + Duration::from_secs(1)),
+                result: None,
+                sender: tx,
+            }))
+            .await
+            .expect("request should enqueue");
+
+        let at_capacity = iouring
+            .fill_submission_queue(&mut ring)
+            .expect("channel should remain connected");
+
+        // The request should retire locally without consuming waiter or SQ
+        // capacity, and its scheduled deadline should disappear as well.
+        assert!(!at_capacity);
+        assert!(iouring.waiters.is_empty());
+        assert_eq!(ring.submission().len(), 0);
+        assert_eq!(iouring.timeout_wheel.next_deadline(), None);
+
+        let (sock_left, _sock_right) = UnixStream::pair().unwrap();
+        // SAFETY: sock_left is a valid fd that we own. This request is
+        // orphaned before staging, so the file descriptor is never submitted.
+        let file = unsafe { std::fs::File::from_raw_fd(sock_left.into_raw_fd()) };
+        let (tx, rx) = oneshot::channel();
+        drop(rx);
+        handle
+            .enqueue(Request::ReadAt(ReadAtRequest {
+                file: Arc::new(file),
+                offset: 0,
+                len: 8,
+                read: 0,
+                buf: IoBufMut::with_capacity(8),
+                result: None,
+                sender: tx,
+            }))
+            .await
+            .expect("request should enqueue");
+
+        let at_capacity = iouring
+            .fill_submission_queue(&mut ring)
+            .expect("channel should remain connected");
+
+        // The read request should take the same orphan path as send.
+        assert!(!at_capacity);
+        assert!(iouring.waiters.is_empty());
+        assert_eq!(ring.submission().len(), 0);
+        assert_eq!(iouring.timeout_wheel.next_deadline(), None);
+    }
+
+    #[tokio::test]
+    async fn test_fill_submission_queue_orphans_closed_ready_queue_entry_locally() {
+        // Verify an orphaned waiter parked in the ready queue retires
+        // locally instead of staging another SQE.
+        let cfg = Config::default();
+        let mut registry = Registry::default();
+        let (_handle, mut iouring) = IoUringLoop::new(cfg.clone(), &mut registry);
+        let mut ring = new_ring(&cfg).expect("unable to create io_uring instance");
+
+        // Keep the test focused on ready-queue staging instead of wake rearm.
+        iouring.wake_rearm_needed = false;
+
+        let (tx, rx) = oneshot::channel();
+        drop(rx);
+        let waiter_id = iouring.waiters.insert(
+            Request::Recv(RecvRequest {
+                fd: Arc::new(UnixStream::pair().unwrap().0.into()),
+                buf: IoBufMut::with_capacity(8),
+                offset: 4,
+                len: 8,
+                exact: true,
+                deadline: None,
+                result: None,
+                sender: tx,
+            }),
+            Some(1),
+        );
+        iouring.timeout_wheel.schedule(waiter_id, 1);
+        iouring.ready_queue.push_back(waiter_id);
+
+        let at_capacity = iouring
+            .fill_submission_queue(&mut ring)
+            .expect("channel should remain connected");
+
+        // Restaging should notice the closed caller, drop the request locally,
+        // and clean up its deadline tracking without touching the SQ.
+        assert!(!at_capacity);
+        assert!(iouring.waiters.is_empty());
+        assert_eq!(ring.submission().len(), 0);
+        assert_eq!(iouring.timeout_wheel.next_deadline(), None);
+    }
+
+    #[tokio::test]
     async fn test_single_issuer() {
         // Verify SINGLE_ISSUER still allows normal request submission and completion.
         let cfg = Config {
--- a/runtime/src/iouring/mod.rs
+++ b/runtime/src/iouring/mod.rs
@@ -577,10 +577,20 @@ impl IoUringLoop {
     ///
     /// If the request was marked for cancellation while sitting in the ready
     /// queue (timeout fired between requeue and staging), it is completed with
-    /// a timeout error instead of issuing a follow-up SQE.
+    /// a timeout error instead of issuing a follow-up SQE. If the original
+    /// caller dropped its wait handle before staging, the request is retired
+    /// locally without issuing another SQE.
     fn stage_request(&mut self, waiter_id: WaiterId, submission_queue: &mut SubmissionQueue<'_>) {
         match self.waiters.stage(waiter_id) {
             StageOutcome::Timeout(request) => request.timeout(),
+            StageOutcome::Orphaned { target_tick } => {
+                // The caller disappeared before another SQE was issued, so all that
+                // remains is to release deadline tracking (the waiter, and associated
+                // resources, were already dropped inside `Waiters`).
+                if let Some(tick) = target_tick {
+                    self.timeout_wheel.remove(tick);
+                }
+            }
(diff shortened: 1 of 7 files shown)


runtime/src/iouring/mod.rs
```diff
@@ -659,37 +573,35 @@ impl IoUringLoop {
...
+                    completed.finish();
+                }
+            }
+            CqeAction::Requeue => {
+                // Request needs another SQE. Add to ready queue.
+                self.ready_queue.push_back(waiter_id);
+            }
         }
-
-        let _ = sender.send((result, buffer));
```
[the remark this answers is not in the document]
andresilva: This was fixed in b7f29b5d086f4cdeabe6c8f0b0f9603c859f1ed7.

runtime/src/iouring/waiter.rs
```diff
@@ -1,10 +1,12 @@
-//! Waiter identity and lifecycle state for io_uring in-flight operations.
+//! Waiter identity and lifecycle state for tracked io_uring requests.
```
andresilva, line 1: I think this could probably use a better name than waiter now, I'm open for suggestions (can do the rename in a follow-up PR to avoid churn).

patrick-ogrady:
Primary feedback is to avoid sending more SQEs if canceled (which I think is done on `main`):

```
diff --git a/runtime/src/iouring/mod.rs b/runtime/src/iouring/mod.rs
index d3e5713c6..f74e80c6c 100644
--- a/runtime/src/iouring/mod.rs
+++ b/runtime/src/iouring/mod.rs
@@ -581,6 +581,15 @@ impl IoUringLoop {
     fn stage_request(&mut self, waiter_id: WaiterId, submission_queue: &mut SubmissionQueue<'_>) {
         match self.waiters.stage(waiter_id) {
             StageOutcome::Timeout(request) => request.timeout(),
+            StageOutcome::Abandon {
+                request,
+                target_tick,
+            } => {
+                if let Some(tick) = target_tick {
+                    self.timeout_wheel.remove(tick);
+                }
+                drop(request);
+            }
             StageOutcome::Submit(sqe) => {
                 // SAFETY:
                 // - All resources are stored in `self.waiters` until CQE processing, so
@@ -2190,6 +2199,77 @@ mod tests {
         assert!(matches!(result, Err(crate::Error::Timeout)));
     }
 
+    #[tokio::test]
+    async fn test_fill_submission_queue_abandons_closed_request_before_first_submit() {
+        // Verify dropping the caller before the loop stages the first SQE
+        // retires the logical request locally instead of sending any bytes.
+        let cfg = Config::default();
+        let mut registry = Registry::default();
+        let (submitter, mut iouring) = IoUringLoop::new(cfg.clone(), &mut registry);
+        let mut ring = new_ring(&cfg).expect("unable to create io_uring instance");
+
+        iouring.wake_rearm_needed = false;
+
+        drop(rx);
+        submitter
+            .enqueue(Request::Send(SendRequest {
+                fd: Arc::new(UnixStream::pair().unwrap().0.into()),
+                write: IoBufs::from(IoBuf::from(b"hello")).into(),
+                deadline: Some(Instant::now() + Duration::from_secs(1)),
+                result: None,
+                sender: tx,
+            }))
+            .await
+            .unwrap();
+
+        let at_capacity = iouring
+            .fill_submission_queue(&mut ring)
+            .expect("channel should remain connected");
+
+        assert!(!at_capacity);
+        assert!(iouring.waiters.is_empty());
+        assert_eq!(ring.submission().len(), 0);
+        assert_eq!(iouring.timeout_wheel.next_deadline(), None);
+    }
+
+    #[tokio::test]
+    async fn test_fill_submission_queue_abandons_closed_ready_queue_entry_locally() {
+        // Verify a caller-abandoned waiter parked in the ready queue retires
+        // locally instead of staging another SQE.
+        let cfg = Config::default();
+        let mut registry = Registry::default();
+        let (_submitter, mut iouring) = IoUringLoop::new(cfg.clone(), &mut registry);
+        let mut ring = new_ring(&cfg).expect("unable to create io_uring instance");
+
+        iouring.wake_rearm_needed = false;
+
+        drop(rx);
+        let request = Request::Recv(RecvRequest {
+            fd: Arc::new(UnixStream::pair().unwrap().0.into()),
+            buf: IoBufMut::with_capacity(8),
+            offset: 4,
+            len: 8,
+            exact: true,
+            deadline: Some(Instant::now() + Duration::from_secs(1)),
+            result: None,
+            sender: tx,
+        });
+        let waiter_id = iouring.waiters.insert(request, Some(1));
+        iouring.timeout_wheel.schedule(waiter_id, 1);
+        iouring.ready_queue.push_back(waiter_id);
+
+        let at_capacity = iouring
+            .fill_submission_queue(&mut ring)
+            .expect("channel should remain connected");
+
+        assert!(!at_capacity);
+        assert!(iouring.waiters.is_empty());
+        assert_eq!(ring.submission().len(), 0);
+        assert_eq!(iouring.timeout_wheel.next_deadline(), None);
+    }
+
     #[tokio::test]
     async fn test_single_issuer() {
         // Verify SINGLE_ISSUER still allows normal request submission and completion.
diff --git a/runtime/src/iouring/request.rs b/runtime/src/iouring/request.rs
index 5848d39dc..3a3122f03 100644
--- a/runtime/src/iouring/request.rs
+++ b/runtime/src/iouring/request.rs
@@ -124,6 +124,15 @@ impl Request {
         self.deadline().is_some()
     }
 
+    /// Return whether the original network caller has already dropped its wait handle.
+    pub fn caller_abandoned(&self) -> bool {
+        match self {
+            Self::Send(s) => s.sender.is_closed(),
+            Self::Recv(r) => r.sender.is_closed(),
+            Self::ReadAt(_) | Self::WriteAt(_) | Self::Sync(_) => false,
+        }
+    }
+
     /// Build the next SQE for this request, tagged with `waiter_id`.
     pub fn build_sqe(&mut self, waiter_id: WaiterId) -> SqueueEntry {
         let sqe = match self {
@@ -299,7 +308,7 @@ impl SendRequest {
                 self.result = Some(Err(Error::Timeout));
                 true
             }
-            CqeResult::Retry => false,
+            CqeResult::Retry => self.sender.is_closed(),
             CqeResult::Cancelled => {
                 self.result = Some(Err(Error::Timeout));
                 true
@@ -320,7 +329,7 @@ impl SendRequest {
                     self.result = Some(Err(Error::Timeout));
                     true
                 } else {
-                    false
+                    self.sender.is_closed()
                 }
             }
         }
@@ -384,7 +393,7 @@ impl RecvRequest {
                 self.result = Some(Err(Error::Timeout));
                 true
             }
-            CqeResult::Retry => false,
+            CqeResult::Retry => self.sender.is_closed(),
             CqeResult::Cancelled => {
                 self.result = Some(Err(Error::Timeout));
                 true
@@ -407,7 +416,7 @@ impl RecvRequest {
                     self.result = Some(Err(Error::Timeout));
                     true
                 } else {
-                    false
+                    self.sender.is_closed()
                 }
             }
         }
@@ -793,6 +802,19 @@ mod tests {
         });
         assert!(!request.on_cqe(WaiterState::Active { target_tick: None }, -libc::EAGAIN));
 
+        // Once the caller drops the wait handle, a retryable CQE must not
+        // cause the logical send to stage another SQE.
+        drop(rx);
+        let mut request = Request::Send(SendRequest {
+            fd: make_socket_fd(),
+            write: IoBufs::from(IoBuf::from(b"hello")).into(),
+            deadline: None,
+            result: None,
+            sender: tx,
+        });
+        assert!(request.on_cqe(WaiterState::Active { target_tick: None }, -libc::EAGAIN));
+
         // Partial progress followed by a retry after timeout should resolve to timeout.
         let (tx, rx) = oneshot::channel();
         let mut request = Request::Send(SendRequest {
@@ -810,6 +832,19 @@ mod tests {
             Err(Error::Timeout)
         ));
 
+        // Partial progress must also stop immediately once the caller has
+        // abandoned the logical send.
+        drop(rx);
+        let mut request = Request::Send(SendRequest {
+            fd: make_socket_fd(),
+            write: IoBufs::from(IoBuf::from(b"hello")).into(),
+            deadline: None,
+            result: None,
+            sender: tx,
+        });
+        assert!(request.on_cqe(WaiterState::Active { target_tick: None }, 2));
+
         // Partial progress after timeout must also resolve to timeout rather than requeueing.
         let (tx, rx) = oneshot::channel();
         let mut request = Request::Send(SendRequest {
@@ -927,6 +962,22 @@ mod tests {
         });
         assert!(!request.on_cqe(WaiterState::Active { target_tick: None }, -libc::EAGAIN));
 
+        // Once the caller drops the wait handle, a retryable CQE must not
+        // cause the exact recv to stage another SQE.
+        drop(rx);
+        let mut request = Request::Recv(RecvRequest {
+            fd: make_socket_fd(),
+            buf: IoBufMut::with_capacity(5),
+            offset: 0,
+            len: 5,
+            exact: true,
+            deadline: None,
+            result: None,
+            sender: tx,
+        });
+        assert!(request.on_cqe(WaiterState::Active { target_tick: None }, -libc::EAGAIN));
+
         // Non-exact recv should complete as soon as any positive byte count arrives.
         let (tx, rx) = oneshot::channel();
         let mut request = Request::Recv(RecvRequest {
@@ -967,6 +1018,22 @@ mod tests {
             Err((_, Error::Timeout))
         ));
 
+        // Partial progress must also stop immediately once the caller has
+        // abandoned the exact recv.
+        drop(rx);
+        let mut request = Request::Recv(RecvRequest {
+            fd: make_socket_fd(),
+            buf: IoBufMut::with_capacity(5),
+            offset: 0,
+            len: 5,
+            exact: true,
+            deadline: None,
+            result: None,
+            sender: tx,
+        });
+        assert!(request.on_cqe(WaiterState::Active { target_tick: None }, 3));
+
         // Retryable and ECANCELED completions after timeout should both resolve to timeout.
         let (tx, rx) = oneshot::channel();
         let mut request = Request::Recv(RecvRequest {
diff --git a/runtime/src/iouring/waiter.rs b/runtime/src/iouring/waiter.rs
index 983ab9f4e..c76de762a 100644
--- a/runtime/src/iouring/waiter.rs
+++ b/runtime/src/iouring/waiter.rs
@@ -128,6 +128,14 @@ pub enum StageOutcome {
     /// The waiter was canceled while parked in the ready queue and should
     /// complete locally with timeout rather than issuing another SQE.
     Timeout(Request),
+    /// The original network caller dropped its wait handle before the next SQE
+    /// was staged, so the request should be retired locally without restaging.
+    Abandon {
+        /// The request to retire.
+        request: Request,
+        /// Active deadline tracking to remove from the timeout wheel, if any.
+        target_tick: Option<Tick>,
+    },
     /// The waiter is still active and produced an SQE for submission.
     Submit(SqueueEntry),
 }
@@ -270,6 +278,15 @@ impl Waiters {
 
         if slot.state == WaiterState::CancelRequested {
             StageOutcome::Timeout(self.take(index))
+        } else if slot.request.caller_abandoned() {
+            let target_tick = match slot.state {
+                WaiterState::Active { target_tick } => target_tick,
+                WaiterState::CancelRequested => None,
+            };
+            StageOutcome::Abandon {
+                request: self.take(index),
+                target_tick,
+            }
         } else {
             assert!(
                 !slot.in_flight,

```

patrick-ogrady:
This should only really happen when closing a connect to a peer, so it isn't too bad to keep around?

runtime/src/iouring/request.rs
```diff
@@ -124,6 +124,19 @@ impl Request {
         self.deadline().is_some()
     }
 
+    /// Return whether this request should be treated as orphaned.
+    ///
+    /// A request is orphaned only when its completion receiver has been dropped
+    /// and this request kind stops driving follow-up SQEs in that state.
+    pub fn is_orphaned(&self) -> bool {
+        match self {
+            Self::Send(s) => s.sender.is_closed(),
```
patrick-ogrady, line 133: For consistency, should we keep "send" as infallible (akin to write/sync)? Do you know what tokio does here (do they always drive write/sync or cancel if the result is dropped)?

andresilva: https://docs.rs/tokio/latest/src/tokio/io/util/write_all.rs.html#34-55 tokio will also stop writing as we do here.

patrick-ogrady (approved):

PR #3420, opened 2026-03-17
[runtime,stream,p2p] Opt-in buffered writes and batch drain for send throughput

### LINKED ISSUE #3421: [runtime,stream,p2p] Opt-in buffered writes and batch drain for send throughput
## Problem

Under high-throughput workloads, the serial encrypt-frame-write path in the P2P layer becomes a bottleneck. Each `Sink::send()` call issues an independent syscall, which caps throughput at roughly ~700 sends/s (~1.4ms per send). With bounded internal channels, this leads to significant message loss as the channel overflows on the receive side.

Application-level batching (e.g. coalescing many orders into a single P2P message) helps but does not fully resolve the issue because the transport layer still serialises one syscall per message. The frame header and ciphertext body are written separately.

## Proposal

Two opt-in, backward-compatible configuration knobs:

### 1. `write_buffer_size` on `runtime::tokio::Config` (default: 0)

Wrap the tokio `Sink` in `BufWriter`. When non-zero, multiple `send()` calls are coalesced into fewer syscalls. A new `flush()` method on the `Sink` trait (default no-op) lets callers control when buffered data is flushed to the peer.

With the default of 0, `BufWriter::with_capacity(0, ...)` writes through immediately, preserving current `send()` semantics.

### 2. `max_send_batch` on `discovery::Config` / `lookup::Config` (default: 1)

When set > 1, the peer actor drains up to `max_send_batch - 1` additional ready messages from the channel (high-priority first) after dequeuing the first, then flushes once. This amortises the per-message encrypt-frame-write cost across the batch.

With the default of 1, the peer actor sends one message per wakeup with no drain or flush -- identical to current behaviour.

### Backward compatibility

Both defaults (zero buffer + batch size 1) produce bit-for-bit identical behaviour to the current code. The stream handshake adds `flush()` calls after each frame, which are no-ops when the buffer is 0.

## Implementation

PR: #3420

Projoh:
Closes #3421

## Summary

Adds opt-in write buffering and batch draining to improve send throughput for high-volume workloads without changing default behaviour.

**runtime**: Extends `Sink` with a `flush()` method (default no-op). The tokio `Sink` now wraps `OwnedWriteHalf` in `BufWriter`. `Config` gains a `write_buffer_size` field that defaults to **0** (write-through), so existing callers see no behavioural change.

**stream**: Flushes after each handshake frame in `dial`/`listen` so the encrypted channel works correctly when buffering is enabled.

**p2p**: Adds `max_send_batch` to both `discovery::Config` and `lookup::Config`, defaulting to **1** (no batching). When set > 1, the peer actor drains up to that many ready messages (high-priority first) before flushing, amortising the per-message encrypt-frame-write cost across the batch.

Both knobs are fully backward-compatible: zero buffer + batch size 1 produces identical behaviour to the current code.

### Motivation

Under high-throughput workloads (~700+ sends/s), the serial encrypt-frame-write path becomes the bottleneck. Each `send()` triggers an independent syscall, capping throughput around ~700 sends/s with ~1.4ms per send. With bounded internal channels, this leads to significant message loss as the channel overflows.

Application-level batching alone cannot resolve this because the transport layer still serialises one syscall per message. `BufWriter` coalesces the frame header + ciphertext writes, and batch draining amortises the flush across multiple messages.

These changes are intentionally opt-in so that applications with different latency/throughput trade-offs can choose the right configuration.

## Test plan

- [x] All existing `commonware-runtime` tests pass (336/336, excluding pre-existing `AddrInUse` env issue)
- [x] All existing `commonware-stream` tests pass (12/12)
- [x] All existing `commonware-p2p` tests pass (326/326)
- [x] New `test_buffered_write_requires_flush` test validates the buffered write path
- [x] `cargo clippy` clean on all three crates
- [x] `cargo +nightly fmt` clean

### DIFF
--- a/p2p/src/authenticated/discovery/actors/peer/actor.rs
+++ b/p2p/src/authenticated/discovery/actors/peer/actor.rs
@@ -1105,4 +1105,470 @@ mod tests {
             assert_eq!(invalid_count, 1);
         });
     }
+
+    fn make_encoded_data(
+        pool: &commonware_runtime::BufferPool,
+        channel: u64,
+        payload_byte: u8,
+    ) -> EncodedData {
+        use commonware_runtime::iobuf::EncodeExt;
+        let data = crate::authenticated::data::Data {
+            channel,
+            message: IoBuf::from(vec![payload_byte; 10]),
+        };
+        let payload = types::Payload::<PublicKey>::Data(data);
+        EncodedData {
+            channel,
+            payload: IoBufs::from(payload.encode_with_pool(pool)),
+        }
+    }
+
+    struct EncryptedPair {
+        local_sender: Sender<mocks::Sink>,
+        local_receiver: Receiver<mocks::Stream>,
+        remote_sender: Sender<mocks::Sink>,
+        remote_receiver: Receiver<mocks::Stream>,
+    }
+
+    async fn setup_encrypted_connection(
+        context: deterministic::Context,
+        local_key: PrivateKey,
+        remote_key: PrivateKey,
+    ) -> EncryptedPair {
+        let local_pk = local_key.public_key();
+        let remote_pk = remote_key.public_key();
+        let (local_sink, remote_stream) = mocks::Channel::init();
+        let (remote_sink, local_stream) = mocks::Channel::init();
+
+        let local_config = stream_config(local_key);
+        let remote_config = stream_config(remote_key.clone());
+
+        let listener_handle = context.clone().spawn({
+            move |ctx| async move {
+                commonware_stream::encrypted::listen(
+                    ctx,
+                    |_| async { true },
+                    remote_config,
+                    remote_stream,
+                    remote_sink,
+                )
+                .await
+                .map(|(pk, sender, receiver)| {
+                    assert_eq!(pk, local_pk);
+                    (sender, receiver)
+                })
+            }
+        });
+
+        let (local_sender, local_receiver) = commonware_stream::encrypted::dial(
+            context.clone(),
+            local_config,
+            remote_pk,
+            local_stream,
+            local_sink,
+        )
+        .await
+        .expect("dial failed");
+
+        let (remote_sender, remote_receiver) = listener_handle
+            .await
+            .expect("listen failed")
+            .expect("listen result failed");
+
+        EncryptedPair {
+            local_sender,
+            local_receiver,
+            remote_sender,
+            remote_receiver,
+        }
+    }
+
+    #[test]
+    fn test_batch_drain_sends_all_queued_data() {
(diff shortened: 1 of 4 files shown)


p2p/src/authenticated/discovery/actors/peer/actor.rs
```diff
@@ -199,9 +232,24 @@ impl<E: Spawner + BufferPooler + Clock + CryptoRngCore + Metrics, C: PublicKey>
...
+                                ).await?;
+                                batch_remaining -= drained;
+                                if batch_remaining > 0 {
+                                    Self::drain_data_channel(
+                                        &mut self.low, &mut conn_sender, &self.sent_messages,
+                                        &peer, &*rate_limits, batch_remaining,
+                                    ).await?;
+                                }
+                            }
+                            conn_sender.flush().await.map_err(Error::SendFailed)?;
```
patrick-ogrady, line 250: It won't reduce the syscalls, but you may be able to get better batched sending performance by disabling`TCP_NODELAY`.

runtime/src/network/tokio.rs
```diff
@@ -164,6 +178,12 @@ pub struct Config {
     /// A larger buffer reduces syscall overhead by reading more data per call,
     /// but uses more memory per connection. Defaults to 64 KB.
     read_buffer_size: usize,
+    /// Size of the write buffer for batching network writes.
+    ///
+    /// When non-zero, multiple messages written between
+    /// [`flush`](crate::Sink::flush) calls are coalesced into fewer syscalls.
+    /// Defaults to 0 (write-through, preserving pre-buffer behavior).
+    write_buffer_size: usize,
```
patrick-ogrady, line 186: I could've sworn I came back to add this :sad:

patrick-ogrady: Ah, now I recall. I couldn't settle on how to best "mandate the flush" (as writes will sit for an unbounded period of time until flush). I think that's fine (and the way you handle it is directionally correct imho).

p2p/src/authenticated/discovery/actors/peer/actor.rs
Projoh: from what I can see TCP_NODELAY is already on by default

Projoh: Ah, the double negative got me. I think we should keep TCP_NODELAY on because BufWriter already handles the batching before flush, so enabling Nagle on top doesnt add much

patrick-ogrady:
I thought about this a bit more and you should just be able to concat IoBufs to avoid any copying into a write buffer. This will then use vectored writes on the network layer.

Issue #2842, opened 2026-01-15
[utils] Implement a Binary Fuse Filter


patrick-ogrady:
This is a faster alternative to bloom filters when the set is static.

0xAysh:
@patrick-ogrady, im happy to take this on.

Plan: Add a BinaryFuseFilter<F> to the utils crate in a new fuse.rs module, with two fingerprint variants - F8 (~0.4% false positive rate) and F16 (~0.0015% false positive rate).

The API would be minimal:
  - BinaryFuseFilter::new(keys: &[u64]) -> Result<Self, Error> — constructs the static filter
  - fn contains(&self, key: u64) -> bool - three array reads + XOR, very fast

Keys are u64 - callers hash their own types before querying, keeping the implementation simple and matching how most fuse filter implementations work.

Implementation: Own the algorithm from scratch (no external crates) per the project's design principles. Construction uses the standard hypergraph peeling approach from the 2022 paper with a seed-based retry loop. Lookup is three indexed reads + XOR check. Will also include codec support, conformance tests, a fuzz target, and a benchmark for contains.

Stability: ALPHA to start.

Just need a green light before I start working on this.

0xAysh:
@patrick-ogrady Just went ahead and made a PR (#3423).

andresilva:
I didn't look at your PR yet but there are some important considerations if we want to make this usable in a consensus-environment, we need to make sure the implementation is deterministic:
- we shouldn't use floating-point math (have a look at our bloom filter implementation where we use `BigRational`, or do integer approximations of log)
- we shouldn't use `usize` (since this is platform-dependent)
- probably makes sense to canonicalize input set (e.g. sort and dedup?)
- retry logic needs to be deterministic (e.g. maybe max number of retries needs to be part of the constructor?)
- take input seed in constructor

The last point about taking an input seed is critical important in an adversarial environment. If the attacker can control the set of items to be added to the filter then it can manipulate it in a way that filter construction fails. We can overcome this by taking an input seed that cannot be controlled by the adversary (e.g. a VRF output).

0xAysh:
Thanks for the feedback @andresilva, all points have been addressed.

[...]

PR #3423, opened 2026-03-18
[utils] Add BinaryFuseFilter

0xAysh:
Implements BinaryFuseFilter<F> in commonware-utils as in issue #2842.

A Binary Fuse Filter is a static probabilistic set-membership structure. It answers
contains queries using exactly 3 array lookups and an XOR fingerprint check. The full
key set must be known at construction time.

Two fingerprint widths are supported: u8 (~0.4% false positive rate) and u16
(~0.0015% false positive rate). Keys are u64 and callers are responsible for hashing
their own types down before calling new or contains.

Construction uses hypergraph peeling with up to 100 seed retries. The position
formula uses splitmix64 and 128-bit multiplication matching the reference C
implementation at github.com/FastFilter/xor_singleheader. Getting this formula
right was the main challenge — earlier approaches using 32-bit fast reduce failed
to peel large key sets because the effective band density exceeded the peeling
threshold.

Includes codec (Read/Write/EncodeSize), conformance tests, benchmarks, a fuzz
target, and an Arbitrary impl gated on the arbitrary feature. Stability is ALPHA.
Gated on cfg(feature = "std") due to f64::ln() in the sizing functions.

### DIFF
--- a/utils/Cargo.toml
+++ b/utils/Cargo.toml
@@ -84,3 +84,8 @@ path = "src/bitmap/benches/bench.rs"
 name = "rational"
 harness = false
 path = "src/rational/benches/bench.rs"
+
+[[bench]]
+name = "fuse"
+harness = false
+path = "src/fuse/benches/bench.rs"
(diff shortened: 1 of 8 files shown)


clabby:

PR #3424, opened 2026-03-18
[consensus/simplex] Remove `nullify` after `nullification` behavior

BrendanChou:
Behavior is no longer needed. Reverting part of #3225

### DIFF
--- a/consensus/src/simplex/actors/voter/actor.rs
+++ b/consensus/src/simplex/actors/voter/actor.rs
@@ -323,7 +323,7 @@ impl<
         Some(Request(context, receiver))
     }
 
-    /// Records a locally verified nullify vote and ensures the round exists.
+    /// Persists our nullify vote to the journal for crash recovery.
     async fn handle_nullify(&mut self, nullify: Nullify<S>) {
         self.append_journal(nullify.view(), Artifact::Nullify(nullify))
             .await;


consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -678,8 +671,6 @@ impl<
             .await;
         self.try_broadcast_notarization(resolver, certificate_sender, view, resolved)
             .await;
-        self.try_broadcast_nullify(batcher, vote_sender, view, false)
```
patrick-ogrady, line 681: Previously, we had a comment here explaining why we don't send a nullify here. I think it would be worth re-adding.

consensus/src/simplex/mod.rs
```diff
@@ -75,19 +75,6 @@
 //! This means that a new participant joining consensus will immediately jump ahead on the previous
 //! view's nullification or finalization and begin participating in consensus at the current view.
 //!
-//! ### Vote Broadcast
```
patrick-ogrady, line 78: Should've removed this in the earlier PR, my b.

patrick-ogrady, line 78: Wondering if we should instead replace this with the actual functionality? It seems like something users of Simplex would want to know?

patrick-ogrady: How do you feel about?

```
//! ### Lazy Vote Broadcast
//!
//! Honest participants vote lazily, emitting votes only when their local state machine indicates
//! it may be helpful for progress (rather than gossiping a vote whenever strictly possible).
//!
//! This means, for example, that a participant that observes a `finalization(c,v)` and has not
//! yet broadcast `finalize(c,v)` will not do so.
```

BrendanChou: I would probably just leave it as-is for now. It doesn't seem helpful to me to be overly specific about these vote choices right now to users. Even as you mention we might change the voting behavior later in subtle ways, it is better to not have to keep the documentation in-sync.

I also think that the point about not-vote-finalize-if-seen-nullification is pretty different than this scenario. This PR is about removing additional unnecessary code paths. Choosing to not vote to finalize in certain scenarios as an optimization also seems to add more code paths and doesn't help with safety, only marginally helping with bandwidth.

PR #3426, opened 2026-03-18
[Storage] Clean up qmdb benchmarks

danlaine:
# QMDB Benchmark Refactor

## Running benchmarks

The two separate binaries (`qmdb_generate`, `qmdb_init`) are now a single binary:

```bash
# Before
cargo bench --bench qmdb_generate
cargo bench --bench qmdb_init

# After
cargo bench --bench qmdb

# Filter to a subset
cargo bench --bench qmdb -- generate
cargo bench --bench qmdb -- init
```

## Benchmark name changes

Names changed due to the binary rename and module restructure. The `module_path!()` prefix now reflects the new layout:

| Old | New |
|-----|-----|
| `qmdb_generate::fixed::generate/...` | `qmdb::generate/...` |
| `qmdb_generate::variable::generate/...` | `qmdb::generate/...` |
| `qmdb_generate::keyless::generate/...` | `qmdb::generate/...` |
| `qmdb_init::fixed::init/...` | `qmdb::init/...` |
| `qmdb_init::variable::init/...` | `qmdb::init/...` |

Fixed and variable benchmarks are now distinguished only by the `variant=` parameter rather than the module prefix. This means historical benchmark data on the dashboard will not connect to these new names.

Variable variant names also changed from hyphens to double colons for consistency with fixed variants:

| Old | New |
|-----|-----|
| `variant=any-unordered` | `variant=any::unordered` |
| `variant=any-ordered` | `variant=any::ordered` |
| `variant=current-unordered` | `variant=current::unordered` |
| `variant=current-ordered` | `variant=current::ordered` |

## Adding a new DB variant

Previously you had to add match arms at every benchmark call site (24+ arms across the init benchmarks alone). Now add the variant once:

1. Add the type alias in `common.rs` (e.g. `pub type NewDb = ...`)
2. Add the enum variant to `FixedVariant` or `VariableVariant` with a `name()` entry
3. Add the dispatch arm in the `with_fixed_db!` or `with_variable_db!` macro

All benchmarks pick it up automatically.

## Value generation

All variable-length value benchmarks (keyed and keyless) now share `make_variable_value`, which generates values of 1-256 bytes with uniform random content. Previously keyed and keyless benchmarks used different inline size ranges.

### DIFF
--- a/storage/src/qmdb/benches/generate.rs
+++ b/storage/src/qmdb/benches/generate.rs
@@ -100,14 +100,8 @@ fn bench_var_value_generate(c: &mut Criterion) {
                             let mut total = Duration::ZERO;
                             for _ in 0..iters {
                                 total += with_var_value_db!(ctx, variant, |mut db| {
-                                    bench_db(
-                                        db,
-                                        elements,
-                                        operations,
-                                        commit_freq,
-                                        make_var_value,
-                                    )
-                                    .await
+                                    bench_db(db, elements, operations, commit_freq, make_var_value)
+                                        .await
                                 });
                             }
                             total


storage/src/qmdb/benches/common.rs
```diff
@@ -0,0 +1,553 @@
...
+            Self::CurrentOrderedVariable => "current::ordered::variable",
+        }
+    }
+}
+
+pub const FIXED_VARIANTS: [FixedVariant; 8] = [
+    FixedVariant::AnyUnorderedFixed,
+    FixedVariant::AnyOrderedFixed,
+    FixedVariant::AnyUnorderedVariable,
+    FixedVariant::AnyOrderedVariable,
```
roberto-bayardo: Is it worth adding a comment explaining the "fixed variants" contain variable variants as well?

Alternatively, you could split the inner loop in the fixed benchmark to additionally test the variable variants if you want to make these lists of variants disjoint.

danlaine: https://github.com/commonwarexyz/monorepo/pull/3426/commits/8381709ab4d0feedae70f61a14be1cb1534ca60a

Added some additional comments / moved things around to hopefully explain this better

PR #3428, opened 2026-03-18
[coding] ZODA reject duplicate shards + interval search

cronokirby:
One issue with ZODA is that we would not look at the indices of the shards passed to us, counting several duplicate shards toward the number we need, in case a caller was enough of a nitwit to do this. Now we also de-dup in this method, using only the first shard for a given index, and not counting duplicates towards the requirement.

We also change the Topology calculation to do a more exhaustive interval search, which scales with $O(\sqrt{D})$. This is a bit more expensive (3x for smaller data sizes, 1.2x for larger data sizes, still not a dominant factor at all for encode), so this PR also changes the strategy to create a new `TopologyHint` type, which the leader searches for, and then have the rest of the Topology be calculated using that hint from the followers. The idea is that even a malicious leader can't force an insecure configuration, just one that might be a bit slower to handle. The thinking is that a malicious leader can already delay data transmission in other ways, so this is an acceptable tradeoff in order to accelerate the common honest use case.

### DIFF
--- a/coding/src/zoda/mod.rs
+++ b/coding/src/zoda/mod.rs
@@ -189,12 +189,13 @@ fn row_digest<H: Hasher>(row: &[F]) -> H::Digest {
 }
 
 mod topology;
-use topology::Topology;
+use topology::{Topology, TopologyHint};
 
 /// A shard of data produced by the encoding scheme.
 #[derive(Clone, Debug)]
 pub struct StrongShard<D: Digest> {
     data_bytes: usize,
+    topology_hint: TopologyHint,
     root: D,
     inclusion_proof: Proof<D>,
     rows: Matrix<F>,
@@ -204,6 +205,7 @@ pub struct StrongShard<D: Digest> {
 impl<D: Digest> PartialEq for StrongShard<D> {
     fn eq(&self, other: &Self) -> bool {
         self.data_bytes == other.data_bytes
+            && self.topology_hint == other.topology_hint
             && self.root == other.root
             && self.inclusion_proof == other.inclusion_proof
             && self.rows == other.rows
@@ -216,6 +218,7 @@ impl<D: Digest> Eq for StrongShard<D> {}
 impl<D: Digest> EncodeSize for StrongShard<D> {
     fn encode_size(&self) -> usize {
         self.data_bytes.encode_size()
+            + self.topology_hint.encode_size()
             + self.root.encode_size()
             + self.inclusion_proof.encode_size()
             + self.rows.encode_size()
@@ -226,6 +229,7 @@ impl<D: Digest> EncodeSize for StrongShard<D> {
 impl<D: Digest> Write for StrongShard<D> {
     fn write(&self, buf: &mut impl BufMut) {
         self.data_bytes.write(buf);
+        self.topology_hint.write(buf);
         self.root.write(buf);
         self.inclusion_proof.write(buf);
         self.rows.write(buf);
@@ -244,6 +248,7 @@ impl<D: Digest> Read for StrongShard<D> {
         let max_els = cfg.maximum_shard_size / F::SIZE;
         Ok(Self {
             data_bytes,
+            topology_hint: ReadExt::read(buf)?,
             root: ReadExt::read(buf)?,
             inclusion_proof: Read::read_cfg(buf, &max_els)?,
             rows: Read::read_cfg(buf, &(max_els, ()))?,
@@ -260,6 +265,7 @@ where
     fn arbitrary(u: &mut arbitrary::Unstructured<'_>) -> arbitrary::Result<Self> {
         Ok(Self {
             data_bytes: u.arbitrary::<u32>()? as usize,
+            topology_hint: u.arbitrary()?,
             root: u.arbitrary()?,
             inclusion_proof: u.arbitrary()?,
             rows: u.arbitrary()?,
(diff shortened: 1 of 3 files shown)


cronokirby:
This PR is a breaking-format change for ZODA, but this is fine given its stability level.

coding/src/zoda/topology.rs
```diff
@@ -10,6 +11,123 @@ const SECURITY_BITS: usize = 126;
...
+}
+
+#[cfg(feature = "arbitrary")]
+impl arbitrary::Arbitrary<'_> for TopologyHint {
+    fn arbitrary(u: &mut arbitrary::Unstructured<'_>) -> arbitrary::Result<Self> {
+        Ok(Self {
+            data_cols: u.arbitrary::<u32>()? as usize,
+        })
+    }
+}
```
clabby, line 129: This is covered by the `StrongShard` conformance test, but worth adding a codec conformance test for `TopologyHint`.

coding/src/zoda/topology.rs
```diff
@@ -76,48 +206,35 @@ impl Topology {
...
-                break;
-            }
+    pub const fn hint(&self) -> TopologyHint {
+        TopologyHint::new(self.data_cols)
+    }
+
+    /// Figure out what size different values will have, based on a searched hint
+    /// and the data size. The hint does not need to be optimal, but it must
+    /// still satisfy ZODA's sampling-security bound.
+    pub fn reckon(hint: TopologyHint, config: &Config, data_bytes: usize) -> Option<Self> {
```
clabby, line 216: > but it must still satisfy ZODA’s sampling security bound.

This is a bit confusing, since this function returns `Some` for hints that fail the row-sampling fit check and compensates via extra `column_samples`.

coding/src/zoda/mod.rs
```diff
@@ -607,6 +620,7 @@ impl<H: Hasher> PhasedScheme for Zoda<H> {
         let checking_data = CheckingData::reckon(
             config,
             commitment,
+            shard.topology_hint,
```
clabby, line 629: It seems accepting this verbatim from the untrusted shard within `weaken` may be a DoS problem, yeah? So long as `hint.data_cols <= Self::effective_data_elements(shard.data_bytes)`, `reckon` will compensate with more column samples which can create a lot more work and force us to do quite a few more MSMs than we anticipated.

clabby: From codex:

```rs
diff --git a/coding/src/zoda/mod.rs b/coding/src/zoda/mod.rs
index a1d35a318..e00f687e0 100644
--- a/coding/src/zoda/mod.rs
+++ b/coding/src/zoda/mod.rs
@@ -862,6 +862,95 @@ mod tests {
         ));
     }
 
+    #[test]
+    fn weaken_rejects_amplifying_topology_hint() {
+        let config = Config {
+            minimum_shards: NZU16!(2),
+            extra_shards: NZU16!(1),
+        };
+        let data = vec![0xA5u8; 4 * 1024];
+        let data_bytes = data.len();
+        let topology_hint = TopologyHint {
+            data_cols: F::bits_to_elements(8 * data_bytes),
+        };
+        let Some(topology) = Topology::reckon(topology_hint, &config, data_bytes) else {
+            return;
+        };
+
+        let data = Matrix::init(
+            topology.data_rows,
+            topology.data_cols,
+            F::stream_from_u64s(iter_u64_le(&data[..])),
+        );
+        let encoded_data = data
+            .as_polynomials(topology.encoded_rows)
+            .expect("data has too many rows")
+            .evaluate()
+            .data();
+
+        let row_hashes: Vec<_> = (0..encoded_data.rows())
+            .map(|i| row_digest::<Sha256>(&encoded_data[i]))
+            .collect();
+        let mut bmt_builder = BmtBuilder::<Sha256>::new(row_hashes.len());
+        for hash in &row_hashes {
+            bmt_builder.add(hash);
+        }
+        let bmt = bmt_builder.build();
+        let root = bmt.root();
+
+        let mut transcript = Transcript::new(NAMESPACE);
+        transcript.commit((topology.data_bytes as u64).encode());
+        transcript.commit(root.encode());
+        let commitment = transcript.summarize();
+
+        let mut transcript = Transcript::resume(commitment);
+        let checking_matrix = checking_matrix(&transcript, &topology);
+        let checksum = Arc::new(data.mul(&checking_matrix));
+        transcript.commit(checksum.encode());
+        let shuffled_indices = shuffle_indices(&transcript, encoded_data.rows());
+
+        let shard_idx = 0usize;
+        let indices =
+            &shuffled_indices[shard_idx * topology.samples..(shard_idx + 1) * topology.samples];
+        let rows = Matrix::init(
+            indices.len(),
+            topology.data_cols,
+            indices
+                .iter()
+                .flat_map(|&i| encoded_data[i as usize].iter().copied()),
+        );
+        let inclusion_proof = bmt.multi_proof(indices).unwrap();
+        let shard = StrongShard {
+            data_bytes,
+            topology_hint,
+            root,
+            inclusion_proof,
+            rows,
+            checksum,
+        };
+
+        let Ok((checking_data, _, _)) =
+            Zoda::<Sha256>::weaken(&config, &commitment, shard_idx as u16, shard)
+        else {
+            return;
+        };
+
+        let searched_topology =
+            Topology::reckon(TopologyHint::search(&config, data_bytes), &config, data_bytes)
+                .expect("searched topology hint must be valid");
+        let searched_work = searched_topology.samples
+            * searched_topology.data_cols
+            * searched_topology.column_samples;
+        let accepted_work = checking_data.topology.samples
+            * checking_data.topology.data_cols
+            * checking_data.topology.column_samples;
+
+        assert!(
+            accepted_work <= searched_work * 8,
+            "weaken accepted amplified topology hint: {accepted_work} vs searched {searched_work}"
+        );
+    }
+
     #[cfg(feature = "arbitrary")]
     mod conformance {
         use super::*;
```

result:

```
    thread 'zoda::tests::weaken_rejects_amplifying_topology_hint' (24046124) panicked at coding/src/zoda/mod.rs:948:9:
    weaken accepted amplified topology hint: 131292 vs searched 622
    note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

clabby:

PR #3429, opened 2026-03-19
[consensus/fuzz] Reduce reporter retention in proptests

BrendanChou:
Was running into memory issues in CI when fuzzing. These big maps aren't needed so let's add options to not increase our memory usage so much. The mock will still retain certificates and stuff

## Summary
- add configurable retention options to the simplex mock reporter while keeping full retention as the default
- switch consensus fuzz tests to a minimal reporter mode that only retains certificate state needed by invariant extraction
- reduce memory pressure in long-running proptests without changing production simplex behavior

### DIFF
--- a/consensus/src/simplex/mocks/reporter.rs
+++ b/consensus/src/simplex/mocks/reporter.rs
@@ -22,8 +22,7 @@ use commonware_utils::{
     },
     ordered::{Quorum, Set},
     sync::Mutex,
-    Participant,
-    N3f1,
+    N3f1, Participant,
 };
 use rand_core::CryptoRngCore;
 use std::{


consensus/src/simplex/mocks/reporter.rs
```diff
@@ -43,12 +43,45 @@ pub struct Config<S: Scheme, L: ElectorConfig<S>> {
     pub elector: L,
 }
 
+/// Controls which debug state the mock reporter retains.
+#[derive(Clone, Copy, Debug)]
+pub struct Options {
```
andresilva, line 48: Could we make this part of the config above? I guess we'd need to update config in all the tests with an extra `..Default::default()` line. I guess the reason you did it this way was to minimize churn, but feels weird to have both `Config` and `Options`.

BrendanChou: Yeah that was the main thing I was trying to avoid but it might not be too bad, let me see

andresilva: I think what I suggested above with `..Default::default()` wouldn't work, because we'd have to implement `Default` for `Config`, which isn't possible here. Best we could do is something like `record: RecordConfig::default()`, which is probably worse than what you have in the PR.

clabby, line 48: Can we just make this an enum of `Retain::All` / `Retain::None`? It looks like we only use “all true” or “all false” right now.

BrendanChou:
While this might help, this didn't turn out to be the issue

PR #3430, opened 2026-03-19
[Storage] Consolidate storage configs

danlaine:
## Summary

Replaces flat `mmr_*`/`log_*` prefixed fields in 7 QMDB config structs with nested sub-component config types that already existed (`mmr::journaled::Config`, `journal::contiguous::{fixed,variable}::Config`). Eliminates boilerplate destructuring in init methods, sync helpers, and From impls.

## Motivation

Every QMDB config (any, current, keyless, immutable, store) duplicated the same MMR and journal fields with prefixes, then every `init()` method manually unpacked them into the actual sub-component configs. This made configs verbose to construct (~12 fields each), tedious to maintain, and obscured which fields belonged to which sub-system.

### Config structs

Before (example: `any::FixedConfig`, 10 fields):
```rust
FixedConfig {
    mmr_journal_partition, mmr_metadata_partition, mmr_items_per_blob,
    mmr_write_buffer, log_journal_partition, log_items_per_blob,
    log_write_buffer, translator, thread_pool, page_cache,
}
```

After (3 fields):
```rust
FixedConfig {
    mmr: mmr::journaled::Config,  // journal_partition, metadata_partition, items_per_blob, write_buffer, thread_pool, page_cache
    log: contiguous::fixed::Config, // partition, items_per_blob, page_cache, write_buffer
    translator,
}
```

Same pattern for all 7 structs. `thread_pool` and `page_cache` now live inside the nested sub-configs. `page_cache` (Arc-based, cheap clone) must be cloned into both `mmr` and `log` when constructing.

### Structs affected

| Struct | Fields before | Fields after |
|--------|--------------|-------------|
| `any::FixedConfig` | 10 | 3 |
| `any::VariableConfig` | 12 | 3 |
| `current::FixedConfig` | 11 | 4 |
| `current::VariableConfig` | 13 | 4 |
| `keyless::Config` | 11 | 2 |
| `immutable::Config` | 12 | 3 |
| `store::Config` | 7 | 2 |


roberto-bayardo:
Hmm I recall making these flat because i saw other code in the repo doing the same (though perhaps I'm imagining this).  One benefit of the flat structure is as a user I don't need to import/understand multiple config structs, though I guess arguably you need to understand them anyway since their fields are just copied here.

danlaine:
> One benefit of the flat structure is as a user I don't need to import/understand multiple config structs, though I guess arguably you need to understand them anyway since their fields are just copied here.

I reckon that it's better to have the sub-configs each defined/documented once instead of having to re-define/re-document each of their fields in the places the sub-configs are used

Issue #3427, opened 2026-03-18
[stream,p2p] Encrypt-per-message overhead in batch drain loop


Projoh:
With #3420, `drain_data_channel` collects up to N ready messages before flushing, but each message still goes through its own encrypt -> frame -> write cycle individually. For a batch of 8 messages, that's 8 AEAD calls, 8 buffer allocations, 8 frame headers, and 8 writes into the BufWriter -- the only thing amortized is the flush/syscall at the end.

A few directions this could go:

**Batch into a single AEAD call** -- concatenate the drained plaintexts (with length prefixes), encrypt once, send one frame. Saves on AEAD setup/finalization, allocations, and framing. The issue is `stream` is at BETA stability, so this changes the wire format and would need a migration path or frame-type discriminator. Receiver also needs to know how to split the decrypted batch back into individual messages.

**Concurrent encryption** -- encrypt N messages in parallel across threads. Wire-compatible and the receiver is unchanged, but for typical p2p message sizes (100B-4KB) the thread dispatch overhead likely exceeds the per-message encrypt time (aws-lc-rs ChaCha20-Poly1305 does ~0.3-0.5us per 1KB). Would also need `SendCipher` refactored to separate nonce pre-assignment from encryption. Probably only wins for large messages (>64KB).

**Pre-allocate a single buffer for the batch** -- keep per-message AEAD but allocate one large buffer upfront, encrypt each message into successive regions, then write the whole thing. No wire format change, no new dependencies, avoids N separate allocations. More modest win but zero compatibility risk.

The per-message crypto cost is small enough with aws-lc-rs that the gains here are in the low microseconds per batch. Curious if you have a preferred direction or see something I'm missing.

andresilva:
> Pre-allocate a single buffer for the batch

#3431 fits the batch into multiple chunks of the maximum buffer pool size. I think this is optimal, if we did a single allocation for the entire batch it could degenerate for large batches.

> Concurrent encryption

As noted, it probably only pays off after some threshold of batch size and message size, would require benchmarks to find these cutoff points so that the code could have some heuristics to decide based on this.

PR #3431, opened 2026-03-19
[p2p] `IoBuf`-driven Batching

patrick-ogrady:
Alternative: #3420
Closes: #3421
Related: #3427 (does single alloc to avoid vectored writes)

## TODO
- [x] Add note to mod.rs that you can reach higher performance by adding message batching outside of p2p (because it better amortizes compression)

### DIFF
--- a/p2p/src/authenticated/discovery/actors/peer/actor.rs
+++ b/p2p/src/authenticated/discovery/actors/peer/actor.rs
@@ -198,10 +198,8 @@ impl<E: Spawner + BufferPooler + Clock + CryptoRngCore + Metrics, C: PublicKey> 
                     // Set the initial deadline to now to start gossiping immediately
                     let mut deadline = context.current();
 
-                    // Reused across iterations to avoid per-send allocation.
-                    let mut batch = Vec::with_capacity(self.send_batch_size);
-
                     // Enter into the main loop
+                    let mut batch = Vec::with_capacity(self.send_batch_size);
                     select_loop! {
                         context,
                         on_stopped => {},
@@ -346,7 +344,6 @@ impl<E: Spawner + BufferPooler + Clock + CryptoRngCore + Metrics, C: PublicKey> 
                             context.sleep(wait_duration).await;
                         }
                     }
-
 
                     match msg {
                         types::Payload::Data(data) => {
--- a/p2p/src/authenticated/lookup/actors/peer/actor.rs
+++ b/p2p/src/authenticated/lookup/actors/peer/actor.rs
@@ -157,9 +157,7 @@ impl<E: Spawner + BufferPooler + Clock + CryptoRngCore + Metrics, C: PublicKey> 
                     let mut deadline = context.current() + self.ping_frequency;
 
                     // Enter into the main loop
-                    // Reused across iterations to avoid per-send allocation.
                     let mut batch = Vec::with_capacity(self.send_batch_size);
-
                     select_loop! {
                         context,
                         on_stopped => {},


stream/src/encrypted.rs
```diff
@@ -301,42 +303,75 @@ pub struct Sender<O> {
...
+    pub async fn send(&mut self, bufs: impl Into<IoBufs>) -> Result<(), Error> {
+        let frame = self.encrypt_frame(bufs)?;
+        self.sink.send(frame).await.map_err(Error::SendFailed)
+    }
+
+    /// Encrypts and sends multiple messages in a single runtime write.
+    ///
+    /// Each message is framed independently so receivers still observe the
+    /// original message boundaries.
+    pub async fn send_batch<B, I>(&mut self, bufs: I) -> Result<(), Error>
```
patrick-ogrady: May be worth making it clear that this doesn't combine into a single frame (so compatible with existing usage)

p2p/src/authenticated/lookup/actors/peer/actor.rs
```diff
@@ -85,21 +87,71 @@ impl<E: Spawner + BufferPooler + Clock + CryptoRngCore + Metrics, C: PublicKey>
...
+        for (metric, payload) in batch {
+            metrics_to_inc.push(metric);
+            payloads.push(payload);
+        }
+
+        sender
+            .send_batch(payloads)
+            .await
+            .map_err(Error::SendFailed)?;
+        for metric in metrics_to_inc {
```
patrick-ogrady: Increment before sending batch (in the first loop) -> if partially fails, should have included the attempt.

p2p/src/authenticated/lookup/actors/peer/actor.rs
```diff
@@ -85,21 +87,71 @@ impl<E: Spawner + BufferPooler + Clock + CryptoRngCore + Metrics, C: PublicKey>
...
         payload: types::Message,
     ) -> Result<(), Error> {
-        let msg = payload.encode_with_pool(pool);
-        sender.send(msg).await.map_err(Error::SendFailed)?;
-        sent_messages.get_or_create(&metric).inc();
-        Ok(())
+        Self::flush_send_batch(
+            sender,
+            sent_messages,
+            vec![(metric, payload.encode_with_pool(pool).into())],
```
patrick-ogrady: We can just count the metric here to avoid all the extra movement.

p2p/src/authenticated/lookup/actors/peer/actor.rs
```diff
@@ -140,45 +158,45 @@ impl<E: Spawner + BufferPooler + Clock + CryptoRngCore + Metrics, C: PublicKey>
...
+                                .inc();
+                            let mut batch = Vec::with_capacity(self.send_batch_size);
+                            batch.push(types::Message::Ping.encode_with_pool(&pool).into());
+                            Self::extend_send_many(
+                                &peer, self.send_batch_size, &mut batch,
+                                &mut self.control, &mut self.high, &mut self.low,
+                                &rate_limits, &self.sent_messages,
+                            )?;
+                            conn_sender.send_many(batch).await.map_err(Error::SendFailed)?;
                             deadline = context.current() + self.ping_frequency;
```
patrick-ogrady, line 179: re-add the comment here

stream/src/encrypted.rs
```diff
@@ -301,42 +303,75 @@ pub struct Sender<O> {
...
+        bufs: impl Into<IoBufs>,
+    ) -> Result<commonware_runtime::IoBuf, Error> {
         let mut bufs = bufs.into();
         let ciphertext_len = bufs.len() + TAG_SIZE as usize;
+        let max_ciphertext_size = self.max_message_size.saturating_add(TAG_SIZE) as usize;
+        if ciphertext_len > max_ciphertext_size {
+            return Err(Error::SendTooLarge(ciphertext_len));
+        }
 
-        send_frame_with(
```
patrick-ogrady, line 312: I think we should still use this?

p2p/src/authenticated/discovery/actors/peer/actor.rs
```diff
@@ -73,45 +76,80 @@ impl<E: Spawner + BufferPooler + Clock + CryptoRngCore + Metrics, C: PublicKey>
...
         sent_messages.get_or_create(&metric).inc();
+        batch.push(payload.encode_with_pool(pool).into());
         Ok(())
     }
 
-    /// Sends pre-encoded bytes directly to the stream.
-    async fn send_encoded<Si: Sink>(
-        sender: &mut Sender<Si>,
+    /// Assert the outbound message's `channel` is registered.
+    fn validate_outbound_msg<V>(msg: EncodedData, rate_limits: &HashMap<u64, V>) -> EncodedData {
```
patrick-ogrady: this can be simplified?

p2p/src/authenticated/lookup/actors/peer/actor.rs
```diff
@@ -136,49 +147,50 @@ impl<E: Spawner + BufferPooler + Clock + CryptoRngCore + Metrics, C: PublicKey>
                     let mut deadline = context.current() + self.ping_frequency;
 
                     // Enter into the main loop
+                    // Reused across iterations to avoid per-send allocation.
+                    let mut batch = Vec::with_capacity(self.send_batch_size);
+
```
patrick-ogrady: remove this \n

p2p/src/authenticated/lookup/actors/peer/actor.rs
```diff
@@ -136,49 +147,50 @@ impl<E: Spawner + BufferPooler + Clock + CryptoRngCore + Metrics, C: PublicKey>
...
-                                &self.sent_messages,
-                                metrics::Message::new_ping(&peer),
-                                types::Message::Ping,
-                            )
-                            .await?;
-
-                            // Reset ticker
+                            // Periodically send a ping to the peer, batching
+                            // any already-queued messages into the same write.
+                            self.sent_messages
```
patrick-ogrady: this can become out of sync if `send` fails, however, we'll exit the peer at that point and I consider `sent` to be more useful if it includes attempts

stream/src/encrypted.rs
```diff
@@ -333,10 +338,73 @@ impl<O: Sink> Sender<O> {
...
+
+        // First pass: validate each message and compute the exact aggregate size
+        // so we can allocate the output buffer once.
+        let bufs = bufs.into_iter();
+        let (lower, _) = bufs.size_hint();
+        let mut messages = Vec::with_capacity(lower);
+        let mut total = 0usize;
+        for buf in bufs {
+            let msg = buf.into();
+            let ciphertext_len = msg.len() + TAG_SIZE as usize;
```
patrick-ogrady: We should add some helper for this

p2p/src/authenticated/discovery/actors/peer/mod.rs
```diff
@@ -18,6 +18,7 @@ pub use ingress::Message;
 
 pub struct Config<C: PublicKey> {
     pub mailbox_size: usize,
+    pub send_batch_size: NonZeroUsize,
```
patrick-ogrady, line 21: We should also add bytes here (for channels that send very large values, this could cause a blowup)

patrick-ogrady: We could set this to the size of the largest buffer pool class (just need to ensure last message + tag overhead doesn't exceed)

runtime/src/network/audited.rs
```diff
@@ -11,13 +11,13 @@ pub struct Sink<S: crate::Sink> {
 
 impl<S: crate::Sink> crate::Sink for Sink<S> {
     async fn send(&mut self, bufs: impl Into<IoBufs> + Send) -> Result<(), Error> {
-        let buf = bufs.into().coalesce();
+        let bufs = bufs.into();
         self.auditor.event(b"send_attempt", |hasher| {
             hasher.update(self.remote_addr.to_string().as_bytes());
-            hasher.update(buf.as_ref());
+            bufs.for_each_chunk(|chunk| hasher.update(chunk));
```
patrick-ogrady, line 17: This now properly delineates between different vectored writes.

PR #3432, opened 2026-03-19
[consensus/simplex] Timeout Regression Tests

patrick-ogrady:
Ensure we have a fresh timeout when entering a new view.

### DIFF
--- a/consensus/src/simplex/actors/voter/mod.rs
+++ b/consensus/src/simplex/actors/voter/mod.rs
@@ -2666,9 +2666,9 @@ mod tests {
                 .await;
 
             let duplicate_window = context.current() + Duration::from_millis(300);
-            while context.current() < duplicate_window {
+            loop {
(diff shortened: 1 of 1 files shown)


consensus/src/simplex/actors/voter/mod.rs
```diff
@@ -5954,6 +5954,153 @@ mod tests {
...
+                            if nullify.view() == View::new(2) =>
+                        {
+                            panic!(
+                                "received nullify for view 2 before its fresh leader timeout elapsed"
+                            );
+                        }
+                        batcher::Message::Update { response, .. } => response.send(None).unwrap(),
+                        _ => {}
+                    },
+                    _ = context.sleep(Duration::from_millis(10)) => {}
```
patrick-ogrady: Can we not do `sleep_until` on this and break?

PR #3434, opened 2026-03-20
[runtime/utils/buffer/paged/append] zero-copy writes

andresilva:
This PR removes an avoidable copy from the runtime paged append write path. Full logical pages are no longer materialized into one contiguous physical write buffer before each blob write. Instead, `to_physical_pages` now assembles full-page writes as `IoBufs`, reusing slices of the logical tip buffer for payload bytes and emitting CRC bytes from a separate pooled allocation. The trailing partial page is still materialized into one padded physical page, because that page genuinely requires in-memory zero padding.

`flush_internal` was updated to match that write shape, it now snapshots the full-page prefix of the tip, drains those full logical pages before releasing the tip lock, caches them before the blob write so reads do not observe stale persisted bytes during the handoff, and writes the assembled `IoBufs` to the blob.

Extracted from https://github.com/commonwarexyz/monorepo/pull/3185.
Related https://github.com/commonwarexyz/monorepo/issues/3093 https://github.com/commonwarexyz/monorepo/issues/2883.


runtime/src/utils/buffer/benches/bench.rs
```diff
@@ -1,14 +1,13 @@
 //! Benchmarks for the page cache via the Append blob wrapper.
 //!
-//! Uses memory-based storage (deterministic runtime) to isolate
-//! page cache performance from disk I/O.
+//! Includes both deterministic in-memory runs to isolate buffer/page-cache work
```
patrick-ogrady, line 3: ❤️

runtime/src/utils/buffer/paged/append.rs
```diff
@@ -518,46 +547,60 @@ impl<B: Blob> Append<B> {
...
 
-        // For each logical page, copy over the data and then write a crc record for it.
-        for page in 0..pages_to_write {
-            let start_read_idx = page * logical_page_size;
-            let end_read_idx = start_read_idx + logical_page_size;
-            let logical_page = &buffer_data[start_read_idx..end_read_idx];
-            write_buffer.put_slice(logical_page);
-
-            let crc = Crc32::checksum(logical_page);
+        if pages_to_write > 0 {
```
patrick-ogrady, line 552: I'm not sure how likely this is but a large write will now turn into multiple syscalls (if we exceed the 32 const for vectored writes)? Should we "coalesce" if it exceeds 32?

runtime/src/utils/buffer/paged/append.rs
```diff
@@ -518,46 +547,60 @@ impl<B: Blob> Append<B> {
...
-                Checksum::new(logical_page_size_u16, crc)
-            };
-            write_buffer.put_slice(&crc_record.to_bytes());
+            // Build CRC bytes for full pages once. Full-page payload bytes are appended below as
+            // slices from tip, so we avoid copying logical payload here.
+            let mut crcs = self
+                .cache_ref
+                .pool()
+                .alloc(CHECKSUM_SIZE as usize * pages_to_write);
+
```
patrick-ogrady: remove \n or add block comment

runtime/src/utils/buffer/paged/append.rs
```diff
@@ -282,19 +302,12 @@ impl<B: Blob> Append<B> {
...
         // Count only FULL pages for advancing current_page. A partial page (if included) takes
         // up a full physical page on disk, but it's not complete - the next byte still goes to
         // that same logical page.
-        let total_pages_in_buffer = physical_pages.len() / physical_page_size;
-        let full_pages_written = if partial_page_state.is_some() {
-            total_pages_in_buffer.saturating_sub(1)
-        } else {
-            total_pages_in_buffer
-        };
+        let full_pages_written = pages_to_cache;
```
patrick-ogrady: is this shadowing needed?

runtime/src/utils/buffer/benches/append.rs
patrick-ogrady, line 1: Curious if you could share performance results?

runtime/src/utils/buffer/paged/append.rs
andresilva: See https://github.com/commonwarexyz/monorepo/issues/3440.

runtime/src/utils/buffer/benches/append.rs
andresilva:
```
journal::fixed_append/items=1000 size=32
                        time:   [193.66 µs 194.08 µs 194.73 µs]
                        change: [−5.3152% −3.9942% −2.2876%] (p = 0.00 < 0.05)
                        Performance has improved.

journal::fixed_append/items=10000 size=32
                        time:   [1.8801 ms 1.8818 ms 1.8842 ms]
                        change: [−6.6269% −6.3813% −6.1635%] (p = 0.00 < 0.05)
                        Performance has improved.

journal::fixed_append/items=100000 size=32
                        time:   [19.050 ms 19.087 ms 19.128 ms]
                        change: [−6.8495% −6.6529% −6.4362%] (p = 0.00 < 0.05)
                        Performance has improved.

journal::fixed_append/items=1000000 size=32
                        time:   [199.18 ms 199.79 ms 200.26 ms]
                        change: [−6.7457% −6.3149% −5.8549%] (p = 0.00 < 0.05)
                        Performance has improved.
```

For this specific microbenchmark you're quoting performance regresses a bit. I've done a bunch of micro-benchmarking to isolate the root cause and it's the extra bookkeeping that's required to create `IoBufs` out of the slices (and then to "iterate" on it when it's time to write). It's not about `write` vs `writev` (I measured that with different buffer shapes). Also this benchmark uses 4K pages (whereas the journal benchmark uses 8K), and it's benchmarking the worst case which isn't necessarily the most realistic workload. Either way, at 8K pages this PR also becomes faster than main in this benchmark. I believe we can optimize some of the internals of `IoBufs` to make the bookkeeping less costly (will follow in separate PRs), so I think this is worth keeping. (There's also a conversation to be had about optimal page size, I think >8K is probably the target to aim for, possibly 16K.)

PR #3438, opened 2026-03-20
[Storage] QMDB state sync: don't reset outstanding requests on target update

danlaine:
## Summary

When the sync engine receives a target update, it currently cancels **all** outstanding requests and re-fetches everything from scratch. This wastes in-flight network work for requests that are still within the new target range.

This PR changes `reset_for_target_update` to selectively cancel only requests outside the new range, introduces a configurable historical root cache so retained requests can still be verified, and adds a cancellation signal to the `Resolver` trait so discarded requests can abort in-flight work.

## Problem

Target updates are monotonic (start and end only increase). Many in-flight requests cover positions that remain valid in the new target. Dropping them forces redundant round-trips to re-fetch the same operations. Additionally, discarded requests continue running to completion even though their results will never be used, wasting resolver bandwidth.

A subtle hazard: after a target update, a retained request and a newly-scheduled request can share the same location. Without care, the stale future completing first could consume the fresh request's tracking entry.

## Approach

**Request IDs** (`requests.rs`):
- Each request gets a monotonically increasing `Id` (u64 newtype), eliminating location-based collisions between stale and fresh futures
- Dual-map tracking: `HashMap<Id, (Location, oneshot::Sender<()>)>` for ID-based lookup, `BTreeMap<Location, Id>` for gap detection
- `insert` supersedes any existing request at the same location (drops old cancel sender, removes old tracked entry)
- `remove(id)` only clears `by_location` if it still points to this ID, preventing a stale future from corrupting a newer entry
- `remove_before(loc)` drops all requests at locations < loc; dropped cancel senders signal resolvers

**Resolver cancellation** (`resolver.rs`):
- `get_operations` now takes a `cancel_rx: oneshot::Receiver<()>` parameter
- The corresponding `cancel_tx` is dropped on cancellation, causing `cancel_rx.await` to return `Err`
- Resolvers own the signal and decide how to use it: local resolvers ignore it (fast disk reads), network resolvers can `select!` to abort early

**Unified root lookup** (`engine.rs`):
- New config field `max_retained_roots` controls how many previous roots to remember (0 to disable)
- On target update, the old root is saved in a `HashMap<Location, Digest>` keyed by tree size (`target.range.end()`)
- On proof verification, `proof.leaves` (which equals the `target_size` the request was issued for) determines which root to verify against -- either the current root or a retained one, via a single O(1) HashMap lookup
- Each tree size maps to a unique root (MMR is append-only, `validate_update` rejects unchanged roots), so the lookup is exact with no iteration

**Gap detection** (`gaps.rs`):
- `find_next` now accepts `impl IntoIterator<Item = &Location>` instead of `&BTreeSet<Location>`, avoiding an intermediate allocation

**Safety invariants**:
- Pinned nodes are only extracted from current-root proofs (the database needs them for the latest tree size)
- Stale futures cannot corrupt fresh request tracking (request IDs are unique, `remove` and `insert` guard the `by_location` index)
- The MMR is append-only, so operations verified against any historical root are correct at those positions

### DIFF
--- a/storage/src/qmdb/any/sync/tests.rs
+++ b/storage/src/qmdb/any/sync/tests.rs
@@ -132,7 +132,7 @@ where
             apply_batch_size: 1024,
             max_outstanding_requests: 1,
             update_rx: None,
-            max_retained_roots: 0,
+            max_retained_roots: 8,
         };
 
         // Create the engine
@@ -174,7 +174,7 @@ where
             fetch_batch_size: NZU64!(2),
             db_config,
             update_rx: None,
-            max_retained_roots: 0,
+            max_retained_roots: 8,
         };
 
         let result: Result<H::Db, _> = sync::sync(engine_config).await;
@@ -223,7 +223,7 @@ where
             apply_batch_size: 1024,
             max_outstanding_requests: 1,
             update_rx: None,
-            max_retained_roots: 0,
+            max_retained_roots: 8,
         };
 
         // Perform sync
@@ -304,7 +304,7 @@ where
             apply_batch_size: 1024,
             max_outstanding_requests: 1,
             update_rx: None,
-            max_retained_roots: 0,
+            max_retained_roots: 8,
         };
 
         let synced_db: H::Db = sync::sync(config).await.unwrap();
@@ -378,7 +378,7 @@ where
             apply_batch_size: 1024,
             max_outstanding_requests: 1,
             update_rx: None,
-            max_retained_roots: 0,
+            max_retained_roots: 8,
         };
         let synced_db: H::Db = sync::sync(config).await.unwrap();
 
@@ -479,7 +479,7 @@ where
             apply_batch_size: 1024,
             max_outstanding_requests: 1,
             update_rx: None,
-            max_retained_roots: 0,
+            max_retained_roots: 8,
         };
         let synced_db: H::Db = sync::sync(config).await.unwrap();
 
@@ -938,7 +938,7 @@ where
             apply_batch_size: 1024,
             max_outstanding_requests: 1,
             update_rx: None,
-            max_retained_roots: 0,
+            max_retained_roots: 8,
         };
         let synced_db: H::Db = sync::sync(config).await.unwrap();
 
@@ -1004,7 +1004,7 @@ where
             apply_batch_size: 1024,
             max_outstanding_requests: 1,
             update_rx: None,
-            max_retained_roots: 0,
+            max_retained_roots: 8,
         };
         let synced_db: H::Db = sync::sync(config).await.unwrap();
 
@@ -1357,7 +1357,7 @@ where
             apply_batch_size: 1024,
             max_outstanding_requests: 1,
             update_rx: None,
-            max_retained_roots: 0,
+            max_retained_roots: 8,
         };
 
         // Sync should succeed on the second attempt after the first corrupted pinned nodes
--- a/storage/src/qmdb/current/sync/mod.rs
+++ b/storage/src/qmdb/current/sync/mod.rs
@@ -66,7 +66,7 @@ use crate::{
 use commonware_codec::{Codec, CodecShared, Read as CodecRead};
 use commonware_cryptography::{DigestOf, Hasher};
 use commonware_runtime::{Clock, Metrics, Storage};
-use commonware_utils::{bitmap::Prunable as BitMap, sync::AsyncMutex, Array};
+use commonware_utils::{bitmap::Prunable as BitMap, channel::oneshot, sync::AsyncMutex, Array};
 use std::ops::Range;
 
 #[cfg(test)]
@@ -483,7 +483,7 @@ macro_rules! impl_current_resolver {
                 start_loc: Location,
                 max_ops: std::num::NonZeroU64,
                 include_pinned_nodes: bool,
-                _cancel_rx: commonware_utils::channel::oneshot::Receiver<()>,
+                _cancel_rx: oneshot::Receiver<()>,
             ) -> Result<crate::qmdb::sync::FetchResult<Self::Op, Self::Digest>, Self::Error> {
                 let (proof, operations) = self.any
                     .historical_proof(op_count, start_loc, max_ops)
@@ -496,7 +496,7 @@ macro_rules! impl_current_resolver {
                 Ok(crate::qmdb::sync::FetchResult {
                     proof,
                     operations,
-                    success_tx: commonware_utils::channel::oneshot::channel().0,
+                    success_tx: oneshot::channel().0,
                     pinned_nodes,
                 })
             }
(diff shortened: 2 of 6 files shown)


storage/src/qmdb/current/sync/mod.rs
```diff
@@ -483,6 +483,7 @@ macro_rules! impl_current_resolver {
                 start_loc: Location,
                 max_ops: std::num::NonZeroU64,
                 include_pinned_nodes: bool,
+                _cancel_rx: commonware_utils::channel::oneshot::Receiver<()>,
```
clabby: nit: import `oneshot`, used 3 times in the tests here.

storage/src/qmdb/sync/requests.rs
```diff
@@ -2,47 +2,61 @@
...
-    /// Start locations of outstanding requests
-    /// Each element corresponds to an element in `futures` and vice versa
-    locations: BTreeSet<Location>,
+    /// Start locations of outstanding requests, paired with cancellation
+    /// senders. When an entry is removed (via [`Self::remove`] or
+    /// [`Self::remove_before`]), the sender is dropped, causing the
+    /// resolver's `cancel_rx.await` to return `Err` as a cancellation
+    /// signal. The corresponding future remains in `futures` (removing
+    /// from FuturesUnordered is expensive).
+    locations: BTreeMap<Location, oneshot::Sender<()>>,
```
clabby: Since target updates are currently only required to be >= the previous target's `start_loc`, this could create a problem. After a target update, the engine can cancel an old request and then issue a new request at the same location. If the stale future from the old target completes first, `remove(start_loc)` will remove the new tracked request, and the actual new completion would later be discarded.

storage/src/qmdb/sync/engine.rs
```diff
@@ -273,28 +316,38 @@ where
...
+            let old_target_size = self.target.range.end();
+            self.retained_roots
+                .insert(old_target_size, self.target.root);
+            self.retained_roots_order.push_back(old_target_size);
+            while self.retained_roots.len() > self.max_retained_roots {
+                if let Some(oldest) = self.retained_roots_order.pop_front() {
+                    self.retained_roots.remove(&oldest);
+                }
+            }
+        }
```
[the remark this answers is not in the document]
danlaine: I think this is a pretty niche case. `max_retained_roots == 0` will probably not be used and the performance cost of the temporarily-occupied request slot seems small relative to the added complexity of the additional code.

storage/src/qmdb/sync/requests.rs
danlaine: Added request ids to handle cancellations / outstanding request tracking more cleanly

storage/src/qmdb/any/sync/tests.rs
```diff
@@ -596,6 +608,7 @@ where
             apply_batch_size: 1024,
             max_outstanding_requests: 10,
             update_rx: Some(update_receiver),
+            max_retained_roots: 1,
```
clabby, line 611: Not blocking, but why are some of these unaffected tests setting `max_retained_roots = 1` and others `0`?

storage/src/qmdb/sync/engine.rs
```diff
@@ -273,28 +320,40 @@ where
...
+        self.fetched_operations.clear();
+        self.pinned_nodes = None;
+
+        // Save the current root keyed by its tree size for verifying
+        // retained requests that were issued against this target.
+        if self.max_retained_roots > 0 {
+            let old_target_size = self.target.range.end();
+            self.retained_roots
+                .insert(old_target_size, self.target.root);
+            self.retained_roots_order.push_back(old_target_size);
```
clabby: We shouldn’t have a situation where `old_target_size` is overwritten, but maybe we should only insert into `retained_roots_order` if `self.retained_roots.insert(..)` is `None`?

storage/src/qmdb/sync/engine.rs
```diff
@@ -419,8 +480,25 @@ where
...
-        let need_pinned = self.pinned_nodes.is_none() && start_loc == self.target.range.start();
+        // Look up the root to verify against using proof.leaves (the tree
+        // size the request was issued for). Fresh requests match the current
+        // target; retained requests match a historical root.
+        let is_current = proof.leaves == self.target.range.end();
+        let target_root = if is_current {
+            &self.target.root
+        } else {
+            let Some(root) = self.retained_roots.get(&proof.leaves) else {
+                let _ = success_tx.send(false);
```
clabby: nit: use `FallibleExt::send_lossy`

storage/src/qmdb/sync/engine.rs
```diff
@@ -119,6 +132,20 @@ where
     /// Pinned MMR nodes extracted from proofs, used for database construction
     pinned_nodes: Option<Vec<DB::Digest>>,
 
+    /// Historical roots from previous sync targets, keyed by tree size
+    /// (target.range.end()). Each tree size maps to a unique root because
+    /// the MMR is append-only and validate_update rejects unchanged roots.
+    /// When a retained request completes, proof.leaves identifies which
+    /// historical root to verify against.
+    retained_roots: HashMap<Location, DB::Digest>,
```
clabby, line 140: > Each tree size maps to a unique root because MMR is append-only

This is correct, but not enforced in `validate_update` explicitly. Should we still change that validation to check if `new_target.range.start() > old_target.range.start() && new_target.range.end() > old_target.range.end()`?

storage/src/qmdb/any/sync/tests.rs
danlaine: Good point, changed them to use 8 since I think that is a sensible default

storage/src/qmdb/sync/engine.rs
danlaine: Added assertion that `old_target_size` is not overwritten since that should never happen

storage/src/qmdb/sync/engine.rs
danlaine: done

storage/src/qmdb/sync/engine.rs
danlaine: I changed `validate_update` to assert that `range.start` is non-decreasing and that `range.end` is strictly increasing. My rationale is that, since all QMDBs are append only, the last number of operations (i.e. `range.end`) must always increase between commits. But `range.start`, the index of the first operation in the live set, need not increase just because more operations were added to the end of the log.

storage/src/qmdb/sync/engine.rs
```diff
@@ -273,28 +320,44 @@ where
...
+                    .insert(old_target_size, self.target.root)
+                    .is_none(),
+                "duplicate retained root for tree size {old_target_size:?}"
+            );
+            self.retained_roots_order.push_back(old_target_size);
+            while self.retained_roots.len() > self.max_retained_roots {
+                if let Some(oldest) = self.retained_roots_order.pop_front() {
+                    self.retained_roots.remove(&oldest);
+                }
+            }
```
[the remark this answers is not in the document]
clabby: Reasonable, but not necessary to fix. A breakage here should be covered pretty well by the existing tests.

PR #3381, opened 2026-03-12
[glue] Introduce `commonware-glue`

clabby:
## Overview

Adds the `commonware-glue` crate, which will house opinionated compositions of commonware primitives.

The first module within `glue` is `stateful`, which is a composition of `commonware-storage`'s QMDBs and `commonware-consensus`' `Application`.

In the future, this crate also may house constructs like application-managed DKGs (a. la. `reshare`).

### Depends On

- https://github.com/commonwarexyz/monorepo/pull/3400
- https://github.com/commonwarexyz/monorepo/pull/3438
- https://github.com/commonwarexyz/monorepo/pull/3439
- https://github.com/commonwarexyz/monorepo/pull/3450
- https://github.com/commonwarexyz/monorepo/pull/3503
- https://github.com/commonwarexyz/monorepo/pull/3531
- https://github.com/commonwarexyz/monorepo/pull/3576
    - https://github.com/commonwarexyz/monorepo/pull/3539
- https://github.com/commonwarexyz/monorepo/pull/3543
    - Not a hard dependency. We can prevent this on the app-side, but would be great to solve.
- https://github.com/commonwarexyz/monorepo/pull/3607
- https://github.com/commonwarexyz/monorepo/issues/3538
  - https://github.com/commonwarexyz/monorepo/pull/3624
  - https://github.com/commonwarexyz/monorepo/pull/3588
- https://github.com/commonwarexyz/monorepo/pull/3637
- https://github.com/commonwarexyz/monorepo/pull/3626
- https://github.com/commonwarexyz/monorepo/pull/3650
- https://github.com/commonwarexyz/monorepo/pull/3835
- https://github.com/commonwarexyz/monorepo/pull/3828
- https://github.com/commonwarexyz/monorepo/pull/3765
- https://github.com/commonwarexyz/monorepo/pull/3749
- https://github.com/commonwarexyz/monorepo/pull/3764

PR #3439, opened 2026-03-20
[storage/qmdb] Coordinate state sync finalization

clabby:
## Overview

Adds explicit finish coordination to QMDB's state sync engine and exposes reached-target notifications so external orchestration can coordinate multiple databases syncing towards a unified set of targets.

### Motivation

In #3381, we store the state root and range of each managed database in each block that is formed. When state syncing multiple databases, we must finish state sync for each DB at the same point. To synchronize these tasks, we need:
1. A way for each engine to signal “I reached target X”.
2. A way to hold completion until the orchestrator says “finish now”.

With this, we can hold off the completion of state sync for a set of databases until the full set has converged on the set of targets.

### DIFF
--- a/storage/src/qmdb/any/sync/tests.rs
+++ b/storage/src/qmdb/any/sync/tests.rs
@@ -915,6 +915,170 @@ where
     });
 }
 
+/// Test that a finish signal received before target completion still allows full sync.
+pub(crate) fn test_sync_handles_early_finish_signal<H: SyncTestHarness>()
+where
+    Arc<DbOf<H>>: Resolver<Op = OpOf<H>, Digest = Digest>,
+    OpOf<H>: Encode,
+    JournalOf<H>: Contiguous,
+{
+    let executor = deterministic::Runner::default();
+    executor.start(|mut context| async move {
+        let mut target_db = H::init_db(context.with_label("target")).await;
+        target_db = H::apply_ops(target_db, H::create_ops(30)).await;
+        let lower_bound = target_db.inactivity_floor_loc().await;
+        let upper_bound = target_db.bounds().await.end;
+        let target = Target {
+            root: H::sync_target_root(&target_db),
+            range: non_empty_range!(lower_bound, upper_bound),
+        };
+        let verification_root = target_db.root();
+
+        let (finish_sender, finish_receiver) = mpsc::channel(1);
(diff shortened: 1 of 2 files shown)


storage/src/qmdb/sync/engine.rs
```diff
@@ -173,6 +210,18 @@ where
 
     /// Optional receiver for target updates during sync
     update_receiver: Option<mpsc::Receiver<Target<DB::Digest>>>,
+
+    /// Optional receiver for explicit finish requests.
+    finish_receiver: Option<mpsc::Receiver<()>>,
```
danlaine: nit: I was inconsistent with `_rx`/`_tx` vs `_receiver`/`_sender` -- should we take this opportunity to make them all `_rx`/`_tx`? i.e. change the field names here to match those in the config

storage/src/qmdb/sync/engine.rs
```diff
@@ -173,6 +210,18 @@ where
 
     /// Optional receiver for target updates during sync
     update_receiver: Option<mpsc::Receiver<Target<DB::Digest>>>,
+
+    /// Optional receiver for explicit finish requests.
```
danlaine: nit: should we re-use the comment from the config? It's adds useful info about how this is used

storage/src/qmdb/sync/engine.rs
```diff
@@ -532,8 +666,23 @@ where
     /// Returns `StepResult::Complete(database)` when sync is finished, or
     /// `StepResult::Continue(self)` when more work remains.
     pub(crate) async fn step(mut self) -> Result<NextStep<Self, DB>, Error<DB, R>> {
+        self.drain_finish_requests()?;
+
         // Check if sync is complete
         if self.is_complete().await? {
```
danlaine: nit: should we rename `is_complete` to `is_at_target` or something like that since now being at the target and being done with state sync are different?

storage/src/qmdb/sync/engine.rs
clabby: Unified the names on `_rx` / `_tx`.

storage/src/qmdb/sync/engine.rs
clabby: Good suggestion - done.

PR #3442, opened 2026-03-21
[consensus/simplex] Infer ancestor certification from notarization

### LINKED ISSUE #3433: [consensus/simplex] Infer Certification
If we observe a notarization that references an early notarization (that isn't yet certified), we know that it and its ancestry has been certified by at least f+1 honest parties.

0xAysh:
Fixes #3433 

When a notarization for view N is added, we now walk up the ancestry chain and re-queue any ancestor that has a stored notarization but hasn't been locally certified yet. The walk stops when it hits a certified ancestor, an ancestor with no stored notarization, a view at or below last_finalized, or GENESIS_VIEW.

The reasoning: the f+1 signers of notarization(N) must have certified N-1 before they could vote — the protocol requires a certified parent to accept a proposal. So the node already has implicit proof that N-1 (and its ancestors) are certifiable, even if the certification messages haven't arrived yet. Rather than waiting passively, we trigger local certification immediately.

Test added: notarization_proactively_certifies_ancestors
The test simulates the exact timing gap the fix closes:
  1. add_notarization(view 1) — view 1 queued in certification_candidates
  2. certify_candidates() called — drains the set (simulates the actor dispatching the certify request to the automaton)
  3. add_notarization(view 2), where view 2's parent is view 1 — with the fix, the ancestor walk finds view 1 still has a stored notarization and is not yet
  certified, so it re-queues it alongside view 2
  4. certify_candidates() called again — asserts both view 1 and view 2 are returned


0xAysh:
Closing in favor of consensus/infer-certification branch.

PR #3445, opened 2026-03-22
[p2p] Batch to Network Buffer Pool

patrick-ogrady:
Related: #3431

### DIFF
--- a/runtime/src/network/tokio.rs
+++ b/runtime/src/network/tokio.rs
@@ -155,9 +155,16 @@ pub struct Config {
     /// reclaim socket resources immediately when closing connections to
     /// misbehaving peers.
     so_linger: Option<Duration>,
-    /// Read timeout for connections, after which the connection will be closed
+    /// Read timeout for connections, after which the connection will be closed.
+    ///
+    /// This bounds the entire `Stream::recv` call, not each underlying socket
+    /// read attempt.
     read_timeout: Duration,
-    /// Write timeout for connections, after which the connection will be closed
+    /// Write timeout for connections, after which the connection will be closed.
+    ///
+    /// This bounds the entire `Sink::send` call, not each underlying socket
+    /// write attempt. If callers batch more bytes into one send, slow links may
+    /// require a larger timeout.
     write_timeout: Duration,
     /// Size of the read buffer for batching network reads.
     ///
--- a/runtime/src/tokio/runtime.rs
+++ b/runtime/src/tokio/runtime.rs
@@ -107,6 +107,10 @@ pub struct NetworkConfig {
     so_linger: Option<Duration>,
 
     /// Read/write timeout for network operations.
+    ///
+    /// Bounds the full `Sink::send` and `Stream::recv` calls rather than each
+    /// individual socket syscall. Larger
+    /// batched writes may therefore require a larger timeout.
     read_write_timeout: Duration,
 }
 


stream/src/encrypted.rs
```diff
@@ -352,60 +372,62 @@ impl<O: Sink> Sender<O> {
...
+                    frames.push(self.encrypt_batch(std::mem::take(&mut batch), batch_total)?);
+                    batch_total = 0;
+                }
+                frames.push(self.encrypt_frame(msg)?);
+                continue;
+            }
+
+            // Close the current chunk before it would exceed one network
+            // buffer-pool item.
+            if !batch.is_empty() && batch_total.saturating_add(frame_len) > max_batch_size {
```
patrick-ogrady: Seems overly defensive to check !batch.is_empty() here?

patrick-ogrady:
bugbot run

stream/src/encrypted.rs
```diff
@@ -625,4 +634,135 @@ mod test {
...
+            .await?;
+
+            let (_listener_peer, _listener_sender, mut listener_receiver) =
+                listener_handle.await.unwrap()?;
+            sends.store(0, Ordering::Relaxed);
+            chunk_counts.lock().unwrap().clear();
+
+            let large = vec![3u8; 200];
+            let small = vec![9u8; 16];
+            // A single framed message larger than the cap still goes out, but it
```
patrick-ogrady, line 789: move above block

stream/src/encrypted.rs
```diff
@@ -625,4 +634,135 @@ mod test {
...
+            )
+            .await?;
+
+            let (_listener_peer, _listener_sender, mut listener_receiver) =
+                listener_handle.await.unwrap()?;
+            sends.store(0, Ordering::Relaxed);
+            chunk_counts.lock().unwrap().clear();
+
+            let payload = vec![7u8; 100];
+            // The first two framed messages fit together under the 256-byte cap,
```
patrick-ogrady, line 721: move above block

stream/src/encrypted.rs
```diff
@@ -303,109 +302,127 @@ pub struct Sender<O> {
...
+    /// Encrypts and sends multiple messages in a single sink call.
     ///
     /// Each message is framed independently so receivers still observe the
-    /// original message boundaries. All frames are written into a single
-    /// pre-allocated buffer to avoid per-message runtime buffers.
+    /// original message boundaries. Aggregate writes are broken into contiguous
+    /// chunks capped to one network buffer-pool item, then submitted together as
+    /// a chunked `IoBufs`. An individual message larger than that cap is still
+    /// sent as its own chunk.
     pub async fn send_many<B, I>(&mut self, bufs: I) -> Result<(), Error>
```
patrick-ogrady: • Sender::send_many now encrypts and buffers some messages before the full batch has been validated. If a later item triggers SendTooLarge, the
  sender's nonce state is left ahead of the receiver, breaking subsequent use of that connection.

  Review comment:

  - [P2] Avoid consuming cipher state before the whole batch validates — /Users/patrickogrady/code/monorepo-pr-3431/stream/src/encrypted.rs:393-
    398
    When a later batch element is oversized, any earlier chunk built on this path has already been encrypted even though send_many returns Err
    before writing anything to the sink. That advances self.cipher for messages the peer never saw, so reusing the same Sender after a
    SendTooLarge desynchronizes nonces and causes subsequent traffic to fail. The previous two-pass implementation validated the full iterator
    before consuming cipher state.

patrick-ogrady:
bugbot run

PR #3450, opened 2026-03-23
[storage] Allow rewinding QMDB

### LINKED ISSUE #3444: [storage] Allow for winding back QMDB
## Overview

When managing multiple QMDB instances in an application, an edge case can occur where one DB commits successfully and the program crashes before the others can. When the applications restart, the databases will be in an inconsistent state, with some having applied a state transition that others haven't.

<br/>
<div align="center">
Image
</div>
<br/>

Failing forward here is not possible, since we can't apply state transitions cleanly on top of the desynchronized databases. We could state sync the DBs again to get them to a common target set in history, though doing so would leave gaps of blocks in marshal's database, which we're trying to avoid in #3381.

- edit: This also is not robust against a total-cluster crash. Other nodes would exist in the same inconsistent state, and everyone would endlessly be asking peers for operations/proofs they don't have.

To allow for recovering from this situation, it would be great if we could roll the databases back to a previous `[inactivity_floor, commit_floor)` range. It would be up to the user to keep enough information around to wind back if need be. For the sake of a stateful application that needed crash recovery when calling `sync` on multiple instances of QMDB, they would just need to retain the previous range's operations when pruning.

related: #3381

clabby:
## Overview

Adds support for rewinding various QMDB instances.

closes #3444

### DIFF
--- a/storage/src/qmdb/any/db.rs
+++ b/storage/src/qmdb/any/db.rs
@@ -322,27 +322,28 @@ where
         // restart-stable until a later commit/sync boundary.
         self.log.rewind(rewind_size).await?;
 
-        let restored_locs: Vec<Location> = undos
-            .iter()
-            .filter_map(|undo| match undo {
-                SnapshotUndo::Replace { new_loc, .. } | SnapshotUndo::Insert { new_loc, .. } => {
-                    (*new_loc < rewind_size).then_some(*new_loc)
-                }
-                SnapshotUndo::Remove { .. } => None,
-            })
-            .collect();
-
+        let mut restored_locs = Vec::new();
         for undo in undos {
             match undo {
                 SnapshotUndo::Replace {
                     key,
                     old_loc,
                     new_loc,
-                } => update_known_loc(&mut self.snapshot, &key, old_loc, new_loc),
+                } => {
+                    if new_loc < rewind_size {
+                        restored_locs.push(new_loc);
+                    }
+                    update_known_loc(&mut self.snapshot, &key, old_loc, new_loc);
+                }
                 SnapshotUndo::Remove { key, old_loc } => {
                     delete_known_loc(&mut self.snapshot, &key, old_loc)
                 }
-                SnapshotUndo::Insert { key, new_loc } => self.snapshot.insert(&key, new_loc),
+                SnapshotUndo::Insert { key, new_loc } => {
+                    if new_loc < rewind_size {
+                        restored_locs.push(new_loc);
+                    }
+                    self.snapshot.insert(&key, new_loc);
+                }
             }
         }
 
--- a/storage/src/qmdb/immutable/mod.rs
+++ b/storage/src/qmdb/immutable/mod.rs
@@ -302,6 +302,7 @@ impl<E: RStorage + Clock + Metrics, K: Array, V: VariableValue, H: CHasher, T: T
 
         Ok(())
     }
+
     /// Return the root of the db.
     pub fn root(&self) -> H::Digest {
         self.journal.root()
--- a/storage/src/qmdb/keyless/mod.rs
+++ b/storage/src/qmdb/keyless/mod.rs
@@ -217,7 +217,6 @@ impl<E: Storage + Clock + Metrics, V: VariableValue, H: Hasher> Keyless<E, V, H>
     /// Rewind the database to `size` operations, where `size` is the location of the next append.
     ///
     /// This rewinds both the operations journal and its MMR to the historical state at `size`.
-    /// The rewind is not restart-stable until a subsequent [`Self::commit`] or [`Self::sync`].
     ///
     /// # Errors
     ///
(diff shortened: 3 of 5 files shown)


storage/src/qmdb/immutable/mod.rs
```diff
@@ -228,6 +233,75 @@ impl<E: RStorage + Clock + Metrics, K: Array, V: VariableValue, H: CHasher, T: T
...
+        // Journal rewind happens before in-memory snapshot updates. If a later step fails, this
+        // handle may be internally diverged and must be dropped by the caller.
+        self.journal.rewind(rewind_size).await?;
+        for (loc, key) in rewound_sets {
+            delete_known_loc(&mut self.snapshot, &key, loc);
+        }
+        self.last_commit_loc = rewind_last_loc;
+
+        Ok(())
+    }
```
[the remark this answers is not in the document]
clabby: thanks patbot

storage/src/qmdb/current/db.rs
```diff
@@ -351,6 +352,130 @@ where
...
+        // handle may be internally diverged and must be dropped by the caller.
+        self.any.rewind(size).await?;
+
+        // Rebuild activity bitmap at the new size from the rewound Any state.
+        let mut status = BitMap::<N>::new_with_pruned_chunks(pruned_chunks)
+            .map_err(|_| Error::DataCorrupted("pruned chunks overflow"))?;
+        {
+            let reader = self.any.log.reader().await;
+            let bounds = reader.bounds();
+            for loc in status.len()..bounds.end {
```
danlaine: This is O(unpruned operations) read operations so this could be fairly slow depending on how aggressively the db is pruned / how large the live set is. Any way we could make this faster? Or if this is unavoidable perhaps we should remark on this in the doc comment.

storage/src/qmdb/any/db.rs
```diff
@@ -191,6 +212,151 @@ where
...
+                    key,
+                    old_loc,
+                    new_loc,
+                } => update_known_loc(&mut self.snapshot, &key, old_loc, new_loc),
+                SnapshotUndo::Remove { key, old_loc } => {
+                    delete_known_loc(&mut self.snapshot, &key, old_loc)
+                }
+                SnapshotUndo::Insert { key, new_loc } => self.snapshot.insert(&key, new_loc),
+            }
+        }
```
danlaine: nit: we could combine these loops to avoid allocating vec `restored_locs`. Only saves one allocation, so I'd also be fine keeping this as is since it's probably more readble.

storage/src/qmdb/immutable/mod.rs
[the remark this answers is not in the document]
danlaine: nit: this newline is still missing?

storage/src/qmdb/keyless/mod.rs
```diff
@@ -214,6 +214,59 @@ impl<E: Storage + Clock + Metrics, V: VariableValue, H: Hasher> Keyless<E, V, H>
...
+    /// - the target's required logical range is not fully retained (for keyless, this means the
+    ///   oldest retained location is already beyond the rewind boundary)
+    /// - `size - 1` is not a commit operation
+    ///
+    /// Any error from this method is fatal for this handle. Rewind may mutate journal state
+    /// before this method finishes updating in-memory rewind state. Callers must drop this
+    /// database handle after any `Err` from `rewind` and reopen from storage.
+    ///
+    /// A successful rewind is not restart-stable until a subsequent [`Self::commit`] or
+    /// [`Self::sync`].
```
danlaine, line 234: We say the same above on line 220

utils/src/bitmap/mod.rs
```diff
@@ -249,6 +249,26 @@ impl<const N: usize> BitMap<N> {
...
+    /// Shrink the bitmap to `new_len` bits, discarding trailing bits.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `new_len > self.len()`.
+    pub fn truncate(&mut self, new_len: u64) {
+        assert!(new_len <= self.len(), "cannot truncate to a larger size");
+
+        // Pop full chunks from the back when possible.
+        while self.len - new_len >= Self::CHUNK_SIZE_BITS && self.is_chunk_aligned() {
```
danlaine: If the bitmap doesn't start off chunk-aligned, we fall back to O(delta) `self.pop()`. We should first do `self.pop()`s until the bitmap is chunk aligned, and then pop off chunks, and then do `self.pop()`s until we're at the desired size.

danlaine: Ah [cursor](https://github.com/commonwarexyz/monorepo/pull/3450#discussion_r2989832446) beat me to this

storage/src/qmdb/any/db.rs
```diff
@@ -191,6 +212,151 @@ where
...
+    /// [`Db::sync`].
+    pub async fn rewind(&mut self, size: Location) -> Result<Vec<Location>, Error> {
+        let rewind_size = *size;
+        let current_size = *self.last_commit_loc + 1;
+
+        if rewind_size == current_size {
+            return Ok(Vec::new());
+        }
+        if rewind_size == 0 || rewind_size > current_size {
+            return Err(Error::Journal(JournalError::InvalidRewind(rewind_size)));
```
danlaine, line 245: nit: mind adding a test for these cases?

PR #3457, opened 2026-03-24
[workspace] rustfmt

andresilva:
Ran with https://github.com/andresilva/rustfmt updated to the latest upstream rustfmt.

### DIFF
--- a/consensus/src/marshal/standard/inline.rs
+++ b/consensus/src/marshal/standard/inline.rs
@@ -98,18 +98,17 @@ where
             );
             None
         },
-        result = block_rx => result.map_or_else(
-            |_| {
+        result = block_rx => {
+            if result.is_err() {
                 debug!(
                     stage,
                     ?digest,
                     reason = "failed to fetch block",
                     "skipping block wait"
                 );
-                None
-            },
-            Some
-        ),
+            }
+            result.ok()
+        },
     }
 }
 


p2p/src/authenticated/discovery/actors/peer/actor.rs
```diff
@@ -200,6 +200,7 @@ impl<E: Spawner + BufferPooler + Clock + CryptoRngCore + Metrics, C: PublicKey>
 
                     // Enter into the main loop
                     let mut batch = Vec::with_capacity(self.send_batch_size);
+                    let (control, high, low) = &mut (self.control, self.high, self.low);
```
andresilva, line 203: The reason I did this is because otherwise the `recv_prioritized` below gets too long, and gets formatted in a way that's not as nice. Can revert though.

p2p/src/authenticated/lookup/actors/peer/actor.rs
```diff
@@ -158,6 +158,7 @@ impl<E: Spawner + BufferPooler + Clock + CryptoRngCore + Metrics, C: PublicKey>
 
                     // Enter into the main loop
                     let mut batch = Vec::with_capacity(self.send_batch_size);
+                    let (control, high, low) = &mut (self.control, self.high, self.low);
```
andresilva, line 161: Same as above.

consensus/src/marshal/standard/inline.rs
```diff
@@ -98,15 +98,18 @@ where
...
                     stage,
                     ?digest,
                     reason = "failed to fetch block",
                     "skipping block wait"
                 );
                 None
-            }, Some),
+            },
+            Some
+        ),
```
BrendanChou: Doesn't have to be in this PR, but isn't this just doing a debug if `None`? Feels like there should be a cleaner way to do this... I know there is [`inspect`](https://doc.rust-lang.org/std/option/enum.Option.html#method.inspect) for `Some`, but not sure if easy to do for None.

Maybe just:
```
{
  if result.is_none() {
    debug!(...);
  }
  result
}
```

andresilva: Probably clippy complains about this, but let me try.

andresilva: Updated!

p2p/src/authenticated/discovery/actors/peer/actor.rs
clabby: This is fine IMO, prefer for the `select_loop` to be more readable.

PR #3458, opened 2026-03-24
[runtime] wait for tasks on shutdown

andresilva:
This PR makes tokio shutdown wait for tracked spawned tasks to finish.

The root task did not previously behave like a supervised task on exit. As a result, direct descendants, especially `dedicated()` tasks, could outlive the root after completion or panic. This change gives the root its own supervision tree and tracks spawned task completion during shutdown.

The deterministic runtime does not need the same change because all spawned work already remains executor-owned and is torn down synchronously inside `start()`. Its root loop is wrapped in `catch_unwind`, and after the root returns or panics it clears and drops all remaining tasks before resuming the panic or returning, so no descendant can outlive runtime shutdown.


runtime/src/tokio/runtime.rs
```diff
@@ -529,6 +529,7 @@ impl crate::Runner for Runner {
         let label = Label::root();
         executor.metrics.tasks_spawned.get_or_create(&label).inc();
         let gauge = executor.metrics.tasks_running.get_or_create(&label).clone();
+        gauge.inc();
```
patrick-ogrady: We ended up `.dec()` on the root gauge without first incrementing it.

runtime/src/tokio/runtime.rs
```diff
@@ -86,6 +96,69 @@ impl Metrics {
...
+        }
+
+        let Some(remaining) = deadline.checked_duration_since(Instant::now()) else {
+            return false;
+        };
+        if remaining.is_zero() {
+            return false;
+        }
+
+        self.idle.wait_for(&mut gate, remaining);
```
andresilva, line 185: This is not safe, it needs to be a loop as it was initially since `Condvar` can wake spuriously (https://doc.rust-lang.org/std/sync/struct.Condvar.html#method.wait).

andresilva: Ahh, I forgot we're using parking_lot here (https://docs.rs/parking_lot/latest/parking_lot/struct.Condvar.html#differences-from-the-standard-library-condvar).

clabby:

PR #3459, opened 2026-03-24
[runtime/tokio] flush Rayon buffer-pool caches on idle

andresilva:
Rayon worker threads in our runtime are long-lived. That interacts poorly with the buffer pool's thread-local cache: once a rayon worker touches the buffer pool, it can retain cached buffers indefinitely. Tokio blocking workers do not have this problem since they shut down when idle and naturally release their thread-local state.

This PR adds a small background flush loop for rayon thread pools created by the tokio runtime. When thread-local buffer caching is enabled for either runtime buffer pool, we periodically schedule a `spawn_broadcast` on the rayon pool so each worker gets a chance to flush its thread-local buffer cache back into the global freelists. The important detail is that the broadcast only runs once a worker has drained its local work, so this behaves like an idle-time cleanup pass rather than interrupting active rayon work. That lets us keep the fast path benefits of TLS caching while avoiding indefinite hoarding on long-lived worker threads.

Depends on #3458.

### DIFF
--- a/runtime/src/iobuf/pool.rs
+++ b/runtime/src/iobuf/pool.rs
@@ -138,13 +138,6 @@ pub(crate) enum BufferPoolThreadCacheConfig {
     Disabled,
 }
 
-impl BufferPoolThreadCacheConfig {
-    /// Returns true if thread-local caching is enabled for this config.
-    pub const fn is_enabled(&self) -> bool {
-        matches!(self, Self::Fixed(_) | Self::ForParallelism(_))
-    }
-}
-
 /// Configuration for a buffer pool.
 #[derive(Debug, Clone)]
 pub struct BufferPoolConfig {
@@ -309,6 +302,14 @@ impl BufferPoolConfig {
         self.max_per_class =
             NonZeroU32::new(max_per_class).expect("max_per_class must be non-zero");
         self
+    }
+
+    /// Returns true if thread-local caching is enabled.
+    pub const fn thread_cache_enabled(&self) -> bool {
+        matches!(
+            self.thread_cache_config,
+            BufferPoolThreadCacheConfig::Fixed(_) | BufferPoolThreadCacheConfig::ForParallelism(_)
+        )
     }
 
     /// Validates the configuration, panicking on invalid values.
--- a/runtime/src/iobuf/pool.rs
+++ b/runtime/src/iobuf/pool.rs
@@ -726,6 +726,7 @@ impl BufferPoolThreadCache {
     /// paths. A push or pop sets the bit. The first check after activity clears
     /// it and keeps the cache warm, a later check that finds it still clear
     /// drops the cache, returning entries to the global freelists.
+    #[cfg(not(target_arch = "wasm32"))]
     pub(crate) fn flush_idle() {
         Self::TLS_SIZE_CLASS_CACHES.with(|bins| {
             // SAFETY: this TLS value is only ever accessed by the current thread.
--- a/runtime/src/iobuf/pool.rs
+++ b/runtime/src/iobuf/pool.rs
@@ -138,6 +138,13 @@ pub(crate) enum BufferPoolThreadCacheConfig {
     Disabled,
 }
 
+impl BufferPoolThreadCacheConfig {
+    /// Returns true if thread-local caching is enabled for this config.
+    pub const fn is_enabled(&self) -> bool {
+        matches!(self, Self::Fixed(_) | Self::ForParallelism(_))
+    }
+}
+
 /// Configuration for a buffer pool.
 #[derive(Debug, Clone)]
 pub struct BufferPoolConfig {
--- a/runtime/src/iobuf/pool.rs
+++ b/runtime/src/iobuf/pool.rs
@@ -305,7 +305,7 @@ impl BufferPoolConfig {
     }
 
     /// Returns true if thread-local caching is enabled.
-    pub fn thread_cache_enabled(&self) -> bool {
+    pub(crate) fn thread_cache_enabled(&self) -> bool {
         self.resolve_thread_cache_capacity() > 0
     }
 
--- a/runtime/src/iobuf/pool.rs
+++ b/runtime/src/iobuf/pool.rs
@@ -305,8 +305,8 @@ impl BufferPoolConfig {
     }
 
     /// Returns true if thread-local caching is enabled.
-    pub const fn thread_cache_enabled(&self) -> bool {
-        matches!(self.thread_cache_config, BufferPoolThreadCacheConfig::Enabled(_))
+    pub fn thread_cache_enabled(&self) -> bool {
+        self.resolve_thread_cache_capacity() > 0
     }
 
     /// Validates the configuration, panicking on invalid values.
@@ -1239,6 +1239,26 @@ mod tests {
         let page = page_size();
         let config = test_config(page, page * 4, 10).with_thread_cache_capacity(NZUsize!(11));
         config.validate();
+    }
+
+    #[test]
+    fn test_thread_cache_enabled_uses_resolved_capacity() {
+        let page = page_size();
+
+        // Automatic sizing can resolve to zero for small class budgets.
+        let auto_zero = test_config(page, page, 1);
+        assert_eq!(auto_zero.resolve_thread_cache_capacity(), 0);
+        assert!(!auto_zero.thread_cache_enabled());
+
+        let auto_nonzero = test_config(page, page, 4);
+        assert!(auto_nonzero.resolve_thread_cache_capacity() > 0);
+        assert!(auto_nonzero.thread_cache_enabled());
+
+        let explicit = test_config(page, page, 1).with_thread_cache_capacity(NZUsize!(1));
+        assert!(explicit.thread_cache_enabled());
+
+        let disabled = test_config(page, page, 4).with_thread_cache_disabled();
+        assert!(!disabled.thread_cache_enabled());
     }
 
     #[test]
--- a/runtime/src/iobuf/pool.rs
+++ b/runtime/src/iobuf/pool.rs
@@ -305,6 +305,7 @@ impl BufferPoolConfig {
     }
 
     /// Returns true if thread-local caching is enabled.
+    #[cfg(not(target_arch = "wasm32"))]
     pub(crate) fn thread_cache_enabled(&self) -> bool {
         self.resolve_thread_cache_capacity() > 0
     }
--- a/runtime/src/iobuf/pool.rs
+++ b/runtime/src/iobuf/pool.rs
@@ -559,6 +559,7 @@ struct TlsSizeClassCacheEntry {
 struct TlsSizeClassCache {
     entries: Vec<TlsSizeClassCacheEntry>,
     capacity: usize,
+    recently_used: bool,
 }
 
 impl TlsSizeClassCache {
@@ -567,6 +568,7 @@ impl TlsSizeClassCache {
         Self {
             entries: Vec::with_capacity(capacity),
             capacity,
+            recently_used: false,
         }
     }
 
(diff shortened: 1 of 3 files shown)


runtime/src/tokio/runtime.rs
```diff
@@ -86,6 +102,70 @@ impl Metrics {
...
+    fn track(self: &Arc<Self>) -> TaskGuard {
+        self.live.fetch_add(1, Ordering::Relaxed);
+        TaskGuard {
+            tracker: self.clone(),
+        }
+    }
+
+    /// Blocks until all spawned tasks have finished or the timeout expires.
+    fn wait_for_idle(&self, timeout: Duration) -> bool {
+        if self.live.load(Ordering::Acquire) == 0 {
```
clabby: `wait_for_idle()` can return `true` while a task spawn is still in-flight, because shutdown has no synchronization that prevents new tracking after `tree.abort()` starts.

The problematic sequence is:
- `spawn()` admits the child via `Tree::child`, builds the user future, then calls `executor.tasks.track()` (line 734), and only later registers the aborter (line 760).
- `Runner::start` does `tree.abort()` (line 564) and immediately checks `wait_for_idle()` fast path (line 124).
- If another thread is inside `spawn()` before `track()` executes, `live` can still be `0` at line 124, so shutdown returns early. After that, the in-flight `spawn()` continues and can increment `live` and launch work.

This is not fixed by changing only `fetch_add` to `Release`; the race window is structural — abort/wait can run before increment. You need a shutdown gate that prevents new `track()` admissions once shutdown begins, or makes admission + tracking fully ordered with shutdown.

```patch
diff --git a/runtime/src/lib.rs b/runtime/src/lib.rs
index fc4ee49f2..c45d4c8a5 100644
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -3613,6 +3613,70 @@ mod tests {
         test_root_panic_waits_for_direct_child_cleanup_before_unwind(executor);
     }
 
+    #[test]
+    fn test_tokio_root_waits_for_inflight_external_spawn_before_returning() {
+        let runner = tokio::Runner::new(
+            tokio::Config::default().with_shutdown_timeout(Duration::from_secs(1)),
+        );
+        let (start_returned_tx, mut start_returned_rx) = oneshot::channel();
+        let (release_spawn_tx, release_spawn_rx) = oneshot::channel();
+        let (spawn_entered_test_tx, spawn_entered_test_rx) = oneshot::channel();
+
+        let thread = std::thread::spawn(move || {
+            runner.start(|context| async move {
+                let (spawn_entered_root_tx, spawn_entered_root_rx) = oneshot::channel();
+                let context = context.clone();
+
+                std::thread::spawn(move || {
+                    context.spawn(move |_| {
+                        spawn_entered_root_tx
+                            .send(())
+                            .expect("spawn-entered signal should send");
+                        spawn_entered_test_tx
+                            .send(())
+                            .expect("spawn-entered test signal should send");
+                        release_spawn_rx
+                            .blocking_recv()
+                            .expect("spawn release signal should resolve");
+                        async move {
+                            pending::<()>().await;
+                        }
+                    });
+                });
+
+                spawn_entered_root_rx
+                    .await
+                    .expect("external spawn should reach pre-track section");
+            });
+
+            start_returned_tx
+                .send(())
+                .expect("runner completion signal should send");
+        });
+
+        spawn_entered_test_rx
+            .blocking_recv()
+            .expect("test should observe external spawn before root exits");
+        std::thread::sleep(Duration::from_millis(100));
+
+        let returned_early = start_returned_rx.try_recv().is_ok();
+
+        release_spawn_tx
+            .send(())
+            .expect("spawn release signal should send");
+        if !returned_early {
+            start_returned_rx
+                .blocking_recv()
+                .expect("runner should eventually return");
+        }
+        thread.join().expect("runner thread should join");
+
+        assert!(
+            !returned_early,
+            "runner returned before in-flight external spawn completed",
+        );
+    }
+
     #[test]
     fn test_tokio_spawn_after_abort() {
         let executor = tokio::Runner::default();
```
</details>

runtime/src/iobuf/pool.rs
```diff
@@ -1029,6 +1040,11 @@ impl BufferPool {
     }
 }
 
+/// Flushes the current thread's TLS buffer caches into the global freelists.
+pub fn flush_buffer_pool_thread_local_cache() {
+    TlsCache::flush()
+}
```
clabby: nit: Make `TlsCache` `pub` and make this an associated function?

runtime/src/tokio/runtime.rs
andresilva: I think this may be addressed in #3465 (https://github.com/commonwarexyz/monorepo/pull/3465#discussion_r2985691385).

runtime/src/iobuf/pool.rs
```diff
@@ -304,6 +304,11 @@ impl BufferPoolConfig {
         self
     }
 
+    /// Returns true if thread-local caching is enabled.
+    pub fn thread_cache_enabled(&self) -> bool {
```
clabby: nit: should this stay `pub(crate)`? AFAICT the only caller is the Tokio runtime, and `BufferPoolConfig` is re-exported as part of the BETA API. Keeping this crate-local leaves room to rename/refine the helper if the cache policy changes.

clabby (dismissed):

runtime/src/iobuf/pool.rs
andresilva: Yeah, good point. Updated.

clabby (dismissed):

clabby (dismissed):

runtime/src/tokio/runtime.rs
```diff
@@ -555,6 +566,51 @@ impl Context {
...
+    /// could retain cached buffers indefinitely. Each pass injects a broadcast
+    /// task that runs once per worker after it drains local work, allowing the
+    /// worker to release any cached buffers back to the global freelists.
+    fn spawn_rayon_buffer_pool_thread_cache_flush(&self, pool: &ThreadPool) {
+        let pool = Arc::downgrade(pool);
+        let pending = Arc::new(AtomicBool::new(false));
+        self.with_label("rayon_buffer_pool_thread_cache_flush")
+            .spawn(move |context| async move {
+                loop {
+                    context.sleep(RAYON_BUFFER_POOL_CACHE_FLUSH_INTERVAL).await;
```
patrick-ogrady, line 603: I think we should be a little more clever here and only flush if it hasn't been used within `RAYON_BUFFER_POOL_CACHE_FLUSH_INTERVAL` (a job may have just finished, which I think likely means it is about to be used again)?

runtime/src/tokio/runtime.rs
```diff
@@ -677,7 +733,17 @@ impl crate::ThreadPooler for Context {
                 Ok(())
             })
             .build()
-            .map(Arc::new)
+            .map(Arc::new)?;
+
+        // If any buffer pool has thread-local caching enabled, we must periodically
+        // ask Rayon workers to flush their thread-local caches.
+        if self.network_buffer_pool.config().thread_cache_enabled()
```
patrick-ogrady, line 761: Its a bit annoying to me that the buffer pools don't take into account the concurrency of these thread pools but I'm not really sure how to do it?

I wouldn't be so annoying about this if we didn't often use ~2-3 threads for workers and 10+ for rayon in things like `constantinople`.

runtime/src/tokio/runtime.rs
andresilva: I added a `recently_used` field to the TLS cache, every pop/push sets it to `true`. Added `flush_idle()` which immediately flushes caches that have `recently_used == false`, otherwise it sets `recently_used` to `false` and keeps the cache warm. This avoids timestamping or atomics on the allocator hot path and approximates the behavior we want: a cache is only flushed after it has gone untouched across two `RAYON_BUFFER_POOL_CACHE_FLUSH_INTERVAL` passes.

runtime/src/tokio/runtime.rs
andresilva: I honestly don't see a better default than just using `worker_threads`, as we do now. We could multiply that by 2 to be more conservative, but that also seems a bit arbitrary. These rayon worker pools are created dynamically by the user, so we can't have a default that accounts for them. Also `BufferPoolConfig::parallelism` is an estimate of threads that concurrently touch a given buffer pool, a 10-thread Rayon pool does not necessarily mean 10 threads will use the network/storage pools (e.g. in `constantinople` it's not entirely obvious to me that those threads will use the pool at all). The same applies to the tokio blocking workers.

If parallelism is too high, automatic TLS caches become conservatively small, potentially disabled, which preserves global availability but loses same-thread reuse and hits the global freelist more often. If parallelism is too low, each actual pool-using thread can retain more locally and the global freelist is striped for too little concurrency, so we can see temporary global exhaustion or more contention. The auto policy already has a 2x safety margin here by reserving about half of each class for global reuse, and the idle flush bounds how long rayon/blocking-worker TLS caches can strand buffers. So I think I'd lean towards the default being potentially less parallelism than actually used rather than overestimating.

Workloads with known behavior should set buffer-pool parallelism explicitly to the expected number of concurrent threads that actually use that pool. I don't think the runtime can infer this reliably.

andresilva: The above applies to any task spawned with `context.dedicated()` as well. If that task is going to use the buffer pool it should be accounted for in the parallelism setting, but there's no way for us to reliably infer that.

runtime/src/tokio/runtime.rs
```diff
@@ -555,6 +585,53 @@ impl Context {
...
+    ///
+    /// Rayon worker threads are long-lived, so without periodic flushes they
+    /// could retain cached buffers indefinitely. Each pass injects a broadcast
+    /// task that runs once per worker after it drains local work. Caches used
+    /// since the previous pass are retained for locality, caches untouched for
+    /// a full interval are released back to the global freelists.
+    fn spawn_rayon_buffer_pool_thread_cache_flush(&self, pool: &ThreadPool) {
+        let pool = Arc::downgrade(pool);
+        let pending = Arc::new(AtomicBool::new(false));
+        self.with_label("rayon_buffer_pool_thread_cache_flush")
```
patrick-ogrady, line 600: it is more typical to define this outside of the call (hidden label creation like this is something I'm trying to remove)

andresilva: I guess this can just be a free function then that takes `Spawner`, and we call it below in `create_thread_pool`?

runtime/src/tokio/runtime.rs
patrick-ogrady, line 741: Call me crazy but I think any rayon pool we create will be used continuously. And accessing pooled memory from within that rayon pool is something we'll continue doing?

I can't escape the feeling this is over-engineering/something we should delay?

andresilva: My concern with not having this is that correct `parallelism` sizing becomes much more important, and as discussed in https://github.com/commonwarexyz/monorepo/pull/3459#discussion_r3167448480 that has to be delegated to the user. Rayon workers are long-lived, so if they touch a buffer pool and later stop, free buffers can remain in worker TLS indefinitely. Correct `parallelism` helps, but it does not cover phase changes or periodic use. There is also a placement issue: a 10-thread Rayon pool may run work that only has 4-way effective parallelism, and the workers that touched a size class earlier may not be the workers that get reused later, so capacity can sit in cold worker TLS even if the pool remains active.

I also don't think users have an obvious manual flush point with Rayon, short of doing what this PR does.

The idle flush is meant to handle those cases: active caches stay warm, while inactive worker-local caches eventually return capacity to the global freelist. I agree it is extra machinery, but it is mostly localized to one background task. It also makes Rayon behave more like Tokio blocking workers, whose TLS caches are flushed naturally when idle threads are retired.

andresilva: Also maybe this shouldn't be enabled unconditionally as is done now, and instead could be a runtime configuration (with idle interval configurable as well).

clabby:

PR #3466, opened 2026-03-25
[Storage] Support variable-length keys in `qmdb::immutable`

danlaine:
## Summary

- Relax `qmdb::immutable` from requiring fixed-size keys (`K: Array`) to accepting variable-length keys (`K: Key`), matching the flexibility already present in `qmdb::any` and `qmdb::current`.
- Add `Codec` to the `Key` trait (via `CodecShared`) so all keys are serializable by definition, removing the need for separate codec bounds.
- `Operation::Read::Cfg` changes from `V::Cfg` to `(K::Cfg, V::Cfg)` to support key-specific decoding configuration.
- Add conformance test for `Operation<Vec<u8>, U64>` to lock down the variable-key wire format.
- Update sync example wire protocol to use `IsUnit` instead of hardcoded `Cfg = ()`, accommodating the new tuple config.

### DIFF
--- a/storage/conformance.toml
+++ b/storage/conformance.toml
@@ -138,7 +138,7 @@ hash = "f5456580eb69727a23bfb0281e33f467acc0c91273efebd334d3e82c9770ae76"
 n_cases = 65536
 hash = "a3fbb5f749fa5b73a684e9f0bebd8973c456e5ee43a0a3db75a99aa550dee302"
 
-["commonware_storage::qmdb::immutable::operation::tests::conformance::CodecConformance<Op>"]
+["commonware_storage::qmdb::immutable::operation::tests::conformance::CodecConformance<Operation<U64,U64>>"]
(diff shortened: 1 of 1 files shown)


storage/src/qmdb/immutable/operation.rs
```diff
@@ -250,7 +256,7 @@ mod tests {
         let set_op = Op::Set(key.clone(), value.clone());
         assert_eq!(
             format!("{set_op}"),
-            format!("[key:{key} value:{}]", hex(&value.encode()))
+            format!("[key:{} value:{}]", hex(&key), hex(&value.encode()))
```
roberto-bayardo, line 259: why hex(&key)?  Key implements Display already.

roberto-bayardo: oh scratch that, it's Debug that it implements.

PR #3467, opened 2026-03-25
[runtime] disallow spawn after shutdown has started

andresilva:
Related: #3458

### DIFF
--- a/runtime/src/deterministic.rs
+++ b/runtime/src/deterministic.rs
@@ -92,6 +92,7 @@ use rand_core::CryptoRngCore;
 use rayon::{ThreadPoolBuildError, ThreadPoolBuilder};
 use sha2::{Digest as _, Sha256};
 use std::{
+    any::Any,
     borrow::Cow,
     collections::{BTreeMap, BinaryHeap, HashMap, HashSet},
     mem::{replace, take},
@@ -216,6 +217,9 @@ pub struct Config {
     /// If the runtime is still executing at this point (i.e. a test hasn't stopped), panic.
     timeout: Option<Duration>,
 
+    /// Maximum time to wait for spawned tasks to finish after shutdown begins.
+    shutdown_timeout: Duration,
+
     /// Whether spawned tasks should catch panics instead of propagating them.
     catch_panics: bool,
 
@@ -255,6 +259,7 @@ impl Config {
             cycle: Duration::from_millis(1),
             start_time: UNIX_EPOCH,
             timeout: None,
+            shutdown_timeout: Duration::from_secs(60),
             catch_panics: false,
             storage_fault_cfg: FaultConfig::default(),
             network_buffer_pool_cfg,
@@ -291,6 +296,11 @@ impl Config {
     /// See [Config]
     pub const fn with_timeout(mut self, timeout: Option<Duration>) -> Self {
         self.timeout = timeout;
+        self
+    }
+    /// See [Config]
+    pub const fn with_shutdown_timeout(mut self, timeout: Duration) -> Self {
+        self.shutdown_timeout = timeout;
         self
     }
     /// See [Config]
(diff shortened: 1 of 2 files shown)


runtime/src/tokio/runtime.rs
```diff
@@ -98,32 +98,64 @@ impl Metrics {
...
+    /// Reserves a task slot and returns a guard that releases it on drop.
+    ///
+    /// Returns `None` if shutdown has already closed task registration.
+    fn track(self: &Arc<Self>) -> Option<TaskGuard> {
+        // Atomically increment the live-task count only if shutdown has not
+        // already closed task registration. This single state transition is
+        // the linearization point between "task accepted" and "shutdown
+        // closed registration".
+        self.state
+            .fetch_update(Ordering::Relaxed, Ordering::Relaxed, |state| {
```
andresilva, line 126: All of these (and the ones below) are `Relaxed` because `state` is only used as a state machine, it does not publish or protect any other shared memory (we only need atomicity and modification order on that single word). The only other synchronization here is the `gate` mutex and `idle` condvar, and that is ordered by the mutex/condvar themselves.

runtime/src/utils/supervision.rs
```diff
@@ -110,17 +110,18 @@ impl Tree {
             return;
         };
 
-        // Abort the task
-        if let Some(aborter) = task {
-            aborter.abort();
-        }
-
-        // Drain children so the subtree cannot be aborted twice.
+        // Abort descendants before the task on this node so the entire subtree is
```
andresilva: Found this due to flakiness in the test I added here. In the test, the child’s `Drop` tries to `spawn()` during shutdown cleanup. That means the child task could start dropping, run `SpawnOnDrop`, and call `context.spawn(...)` before its retained context subtree had actually been marked aborted. In that interleaving, the late `spawn()` is not guaranteed to return `Closed`, so the test could fail. Reversing the order of abortion guarantees that any retained child context is already closed by the time task cleanup runs.

andresilva: This caused a visible change in semantics, so I worked around it by closing before triggering abort.

runtime/src/deterministic.rs
```diff
@@ -1187,11 +1204,14 @@ impl crate::Spawner for Context {
         };
         let (f, handle) = Handle::init(
             future,
-            metric,
+            metric.clone(),
             executor.panicker.clone(),
             Arc::clone(&parent),
         );
-        Tasks::register_work(&executor.tasks, label, Box::pin(f));
+        if !Tasks::register_work(&executor.tasks, label, Box::pin(f)) {
```
patrick-ogrady: - [P2] Reject closed deterministic spawns before evaluating f(self) — /Users/patrickogrady/code/monorepo-pr-3458/runtime/src/
    deterministic.rs:1205-1213
    In deterministic::Context::spawn, the new shutdown rejection happens only after f(self) has already been called and wrapped in Handle::init.
    If a late cleanup-time spawn has synchronous setup code, mutates state before returning its async block, or panics during construction, that
    work still runs even though the returned handle resolves to Error::Closed. Tokio now short-circuits before evaluating the closure, so
    deterministic tests can still observe shutdown-time side effects or unwinds that production would reject immediately.

runtime/src/deterministic.rs
patrick-ogrady, line 1429: Can we remove boxed here like we did in tokio? Maybe this is required for deterministic?

runtime/src/tokio/runtime.rs
```diff
@@ -98,32 +98,73 @@ impl Metrics {
 /// Tracks running spawned tasks.
 #[derive(Default)]
 struct TaskTracker {
-    live: AtomicUsize,
+    // Current state: bit 0 marks shutdown-closed task registration, the
+    // remaining bits store the live tracked-task count.
+    state: AtomicUsize,
```
patrick-ogrady, line 103: dumb question: can we just put state in gate?

runtime/src/deterministic.rs
andresilva: Yeah I think it was required for deterministic because we put them all in the same run queue IIRC. But will check again.

runtime/src/tokio/runtime.rs
andresilva: The idea was to avoid having to lock a mutex on every spawn.

PR #3471, opened 2026-03-26
[Storage] QMDB performance improvements

danlaine:
## Summary

- Eliminate N operation clones per batch in `Merkleizer::finish()` by sharing a single `Arc<Vec<Operation>>` between the journal items chain and `base_operations`
- Save one key clone per mutation in unordered merkleize by destructuring owned operations instead of borrowing
- Avoid value clones in `extract_parent_deleted_creates` via `Option::take()`
- Pre-allocate `Vec`s for ops and locations where sizes are known
- Eliminate `dirty_nodes` sort in `UnmerkleizedBatch::merkleize()` by storing `(height, position)` instead of `(position, height)` in the BTreeSet
- Move encoding/compression out of the journal write lock and skip redundant `blob.size()` calls in the append path for both fixed and variable journals

## Detail

**Shared Arc in `finish()`**: `merkleize_with` on `authenticated::UnmerkleizedBatch` hashes items by reference into the MMR and accepts an externally-provided `Arc<Vec<Item>>`. Previously `finish()` cloned every operation to feed `journal_batch.add()`, then separately wrapped the originals in `Arc::new(ops)`. Both destinations held identical data in separate Arcs.

**Owned keys in unordered merkleize**: Switched from `results.iter()` (borrowed keys, 2 clones each) to `results.into_iter()` with destructuring (owned key, 1 clone for the Operation, move into diff).

**`extract_parent_deleted_creates`**: `retain` with `&K, &mut V` forced cloning both key and value. Using `value.take()` extracts the value without cloning.

**BTreeSet ordering for dirty_nodes**: `dirty_nodes` was `BTreeSet<(Position, height)>`, sorted by position first. `merkleize()` collected into a Vec and re-sorted by height. Swapping to `BTreeSet<(height, Position)>` makes the BTreeSet's natural order match what `merkleize_serial`/`merkleize_parallel` need, eliminating the sort entirely.

**Journal append optimizations**: Both fixed and variable contiguous journals now encode items before acquiring the write lock, reducing time under contention. Fixed journals use a new `append_raw` method on the segmented layer that skips the redundant `blob.size()` call (the contiguous layer already knows the section). Variable journals extract encoding/compression into a standalone `encode_item` method called before the lock, then pass pre-encoded bytes to `append_raw` under the lock.

### DIFF
--- a/storage/src/journal/segmented/variable.rs
+++ b/storage/src/journal/segmented/variable.rs
@@ -537,8 +537,8 @@ impl<E: Storage + Metrics, V: CodecShared> Journal<E, V> {
     }
 
     /// Appends an item to `Journal` in a given `section`, returning the offset
-    /// where the item was written and the size of the item (which may now be smaller
-    /// than the encoded size from the codec, if compression is enabled).
+    /// where the item was written and the size of the item (which may differ
+    /// from the raw encoded size if compression is enabled).
     pub async fn append(&mut self, section: u64, item: &V) -> Result<(u64, u32), Error> {
         let (buf, item_len) = Self::encode_item(self.compression, item)?;
         self.append_raw(section, &buf)
--- a/storage/src/qmdb/any/batch.rs
+++ b/storage/src/qmdb/any/batch.rs
@@ -292,7 +292,9 @@ where
         C: Contiguous<Item = Operation<F, U>>,
         I: UnorderedIndex<Value = Location<F>>,
     {
-        let mut locations = Vec::with_capacity(mutations.len());
+        // Extra slack (*3/2) avoids re-allocations when index collisions
+        // cause more than one location per key.
+        let mut locations = Vec::with_capacity(mutations.len() * 3 / 2);
         if self.base_diff.is_empty() {
             for key in mutations.keys() {
                 locations.extend(db.snapshot.get(key).copied());


storage/src/qmdb/any/batch.rs
```diff
@@ -292,7 +292,7 @@ where
         C: Contiguous<Item = Operation<F, U>>,
         I: UnorderedIndex<Value = Location<F>>,
     {
-        let mut locations = Vec::new();
+        let mut locations = Vec::with_capacity(mutations.len());
```
roberto-bayardo: if there's even one collision then this won't be big enough, should we allocate ~1.5x or so just so there's slack?

storage/src/journal/segmented/variable.rs
```diff
@@ -532,19 +532,30 @@ impl<E: Storage + Metrics, V: CodecShared> Journal<E, V> {
...
-        };
+            Ok((buf, item_len_u32))
+        }
+    }
 
-        // Get or create blob
-        let blob = self.manager.get_or_create(section).await?;
+    /// Appends an item to `Journal` in a given `section`, returning the offset
+    /// where the item was written and the size of the item (which may now be smaller
+    /// than the encoded size from the codec, if compression is enabled).
```
roberto-bayardo: compression could actually make it bigger, so maybe just say it could be different than the size.

storage/src/journal/contiguous/variable.rs
```diff
@@ -521,16 +526,22 @@ impl<E: Context, V: CodecShared> Journal<E, V> {
             return Ok(self.inner.read().await.size);
         }
 
+        // Encode before grabbing write guard.
+        let encoded: Vec<_> = items
+            .iter()
+            .map(|item| variable::Journal::<E, V>::encode_item(self.compression, item))
```
roberto-bayardo, line 532: would love to avoid allocating a vector-per-item entirely, but that's an optimization for another day!

storage/src/qmdb/any/batch.rs
danlaine: Done

storage/src/journal/segmented/variable.rs
danlaine: Done

Issue #3474, opened 2026-03-26
[runtime] Consider not boxing spawned futures by default


clabby:
## Overview

https://github.com/commonwarexyz/monorepo/pull/3455 removed boxing futures by default, but this created a few problems in projects like `alto` that caused stack overflows on `debug` builds.

We should investigate what is causing these large futures, and consider relaxing this default. It may be that boxing the spawned futures is not a big deal for performance, since these tasks tend to be long-running, and it is nice for the user to not have to think about this. However, if it yields a large performance increase, we should consider removing this default and sorting out future sizes elsewhere.

PR #3475, opened 2026-03-26
[runtime/tokio] use system thread stack size for all threads

andresilva:
This PR updates `commonware-runtime` to stop relying on rust/tokio's spawned-thread default stack size and instead use the operating system's default stack size for runtime-owned threads. A new helper in `runtime::utils::thread` resolves the system default once and is used for dedicated tokio tasks as well as the `io_uring` network and storage loop threads. This only applies to Unix systems, on Windows we continue to rely on the platform's normal default behavior. Runtime-created rayon workers are already covered because they are spawned through the dedicated-task path.

The motivation is that rust's default spawned-thread stack size is currently 2MB, which is much smaller than the host default on some systems, for example on my NixOS machine, the system default is 16MB. This discrepancy feels arbitrary and the default 2MB seems unnecessarily small.

Related #3474.

### DIFF
--- a/runtime/src/utils/thread.rs
+++ b/runtime/src/utils/thread.rs
@@ -1,8 +1,8 @@
-//! Helpers for resolving the default system thread stack size.
+//! Helpers for resolving the configured thread stack size.
 
-use std::{sync::OnceLock, thread};
+use std::{env, sync::OnceLock, thread};
 
-/// Cached system thread stack size.
+/// Cached configured thread stack size.
 static SYSTEM_THREAD_STACK_SIZE: OnceLock<usize> = OnceLock::new();
 
 /// Rust's default thread stack size.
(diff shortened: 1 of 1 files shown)


runtime/src/network/iouring.rs
```diff
@@ -136,13 +136,13 @@ impl Network {
...
             iouring::IoUringLoop::new(cfg.iouring_config.clone(), sender_registry);
-        std::thread::spawn(move || send_loop.run());
+        utils::thread::spawn(move || send_loop.run());
 
         // Create an io_uring instance to handle receive operations.
         let receiver_registry = registry.sub_registry_with_prefix("iouring_receiver");
         let (recv_submitter, recv_loop) =
             iouring::IoUringLoop::new(cfg.iouring_config, receiver_registry);
-        std::thread::spawn(move || recv_loop.run());
+        utils::thread::spawn(move || recv_loop.run());
```
patrick-ogrady: Wonder if it is worth accepting a config for this configured in the runtime? I think we can default to the system setting for that but think it should be tunable without changing the system default?

idk

runtime/src/utils/thread.rs
```diff
@@ -0,0 +1,103 @@
...
+    let get_result = unsafe { libc::pthread_attr_getstacksize(&attr, &mut stack_size) };
+    // SAFETY: `attr` remains initialized until it is destroyed here.
+    let destroy_result = unsafe { libc::pthread_attr_destroy(&mut attr) };
+
+    if get_result != 0 || destroy_result != 0 || stack_size == 0 {
+        return None;
+    }
+
+    Some(stack_size)
+}
```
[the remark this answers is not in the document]
andresilva: ~~Can someone check whether this is true?~~

It's true, going to do another approach (https://github.com/libuv/libuv/pull/671).

runtime/src/network/iouring.rs
andresilva: Yeah I think that makes sense.

runtime/src/network/iouring.rs
```diff
@@ -119,6 +119,7 @@ impl Network {
     /// The io_uring `size` should be a multiple of the number of expected connections.
     pub(crate) fn start(
         mut cfg: Config,
+        thread_stack_size: usize,
```
patrick-ogrady: Move to config.

PR #3476, opened 2026-03-26
[storage/qmdb] Fix path-dependent key selection in `any::unordered`

clabby:
## Overview

Fix QMDB committed-vs-parent batch divergence under translated-key collisions.

Problem:
  - In unordered `qmdb::any`, building a child batch from a pending parent used `base_diff` and found the exact full key.
  - Building the same child from committed DB state used the snapshot index, which returns all locations sharing the translated key.
  - With narrow translators like `OneCap`/`TwoCap`, collisions could cause the committed-state path to classify a mutation differently from the parent-relative path.
  - That changed the operation log and could produce a different state root for the same logical transition.

Fix:
  - In unordered `qmdb::any`, filter snapshot collision candidates by reading candidates and keeping only the exact full-key match.
  - Keep ordered behavior unchanged, since ordered span maintenance needs the full collision set.

Result:
  - Forming a batch from committed DB state now matches forming it from a parent batch across the QMDB surface.
  - Translator width is no longer a correctness dependency for this bug class.

written by Codex 🤖

### DIFF
--- a/storage/src/qmdb/any/batch.rs
+++ b/storage/src/qmdb/any/batch.rs
@@ -603,13 +603,17 @@ where
             // location. Without this guard, a stale snapshot collision (the
             // pre-parent DB snapshot still containing the key's old location)
             // can consume the mutation at the wrong sort position, changing
-            // the operation order relative to the committed-state path.
-            let base_diff_entry = m.base_diff.get(key);
-            if let Some(entry) = base_diff_entry {
+            // the operation order relative to the committed-state path. When
+            // the base diff entry does match, use it to trace `base_old_loc`
+            // back to the key's location in the base DB snapshot.
+            let base_old_loc = if let Some(entry) = m.base_diff.get(key) {
                 if entry.loc() != Some(old_loc) {
                     continue;
                 }
-            }
+                entry.base_old_loc()
+            } else {
+                Some(old_loc)
+            };
 
             let Some(mutation) = mutations.remove(key) else {
                 // Snapshot index collision: this operation's key does not match
@@ -617,11 +621,9 @@ where
                 continue;
             };
 
-            // Determine base_old_loc: trace through base diff to find
-            // the key's location in the base DB snapshot.
+            // Write the user mutation at the next batch location while
+            // preserving the committed-base provenance computed above.
             let new_loc = Location::new(m.base_size + ops.len() as u64);
-            let base_old_loc = base_diff_entry.map_or(Some(old_loc), DiffEntry::base_old_loc);
-
             match mutation {
                 Some(value) => {
                     ops.push(Operation::Update(update::Unordered(


danlaine:
```
diff --git a/storage/src/qmdb/any/batch.rs b/storage/src/qmdb/any/batch.rs
index 535c282b6..aa20d60e1 100644
--- a/storage/src/qmdb/any/batch.rs
+++ b/storage/src/qmdb/any/batch.rs
@@ -598,10 +598,21 @@ where
         // This includes keys from both the base snapshot and the base diff.
         for (op, &old_loc) in results.iter().zip(&locations) {
             let key = op.key().expect("updates should have a key");
+
+            // A key resolved via base_diff must only match at its base_diff
+            // location. Without this guard, a stale snapshot collision (the
+            // pre-parent DB snapshot still containing the key's old location)
+            // can consume the mutation at the wrong sort position, changing
+            // the operation order relative to the committed-state path.
+            if let Some(entry) = m.base_diff.get(key) {
+                if entry.loc() != Some(old_loc) {
+                }
+            }
+
             let Some(mutation) = mutations.remove(key) else {
                 // Snapshot index collision: this operation's key does not match
-                // the mutation key (the snapshot uses a compressed translated key
-                // that can collide). The mutation will be handled as a create below.
+                // any mutation key. The mutation will be handled as a create below.
                 continue;
             };
```

I think this resolves the issue without adding a new method

storage/src/qmdb/any/batch.rs
patrick-ogrady: Dedup this get

storage/src/qmdb/any/batch.rs
```diff
@@ -598,10 +598,21 @@ where
...
         for (op, &old_loc) in results.iter().zip(&locations) {
             let key = op.key().expect("updates should have a key");
+
+            // A key resolved via base_diff must only match at its base_diff
+            // location. Without this guard, a stale snapshot collision (the
+            // pre-parent DB snapshot still containing the key's old location)
+            // can consume the mutation at the wrong sort position, changing
+            // the operation order relative to the committed-state path.
+            if let Some(entry) = m.base_diff.get(key) {
+                if entry.loc() != Some(old_loc) {
```
patrick-ogrady: Confirm added tests fail

storage/src/qmdb/any/batch.rs
```diff
@@ -598,21 +598,29 @@ where
...
-            let new_loc = Location::new(m.base_size + ops.len() as u64);
-
             // Determine base_old_loc: trace through base diff to find
             // the key's location in the base DB snapshot.
-            let base_old_loc = m
-                .base_diff
-                .get(key)
-                .map_or(Some(old_loc), DiffEntry::base_old_loc);
+            let new_loc = Location::new(m.base_size + ops.len() as u64);
+            let base_old_loc = base_diff_entry.map_or(Some(old_loc), DiffEntry::base_old_loc);
```
patrick-ogrady: Was able to dedup the get here.

bajpainaman:
Why not update the snapshot index to like reflect parent batch changes, instead of filtering stale entries at read time?

PR #3478, opened 2026-03-26
[consensus::simplex] Improve `test_invalid` test

BrendanChou:
## Summary
- make `test_invalid` generate decodable corrupted signatures so threshold schemes can recover a malformed local certificate instead of failing early during assembly
- extract the wrapped fault-injection scheme and elector config into `consensus/src/simplex/mocks/wrapped_scheme.rs` so the invalid-signature harness is reusable and easier to follow
- keep the simplex mock reporter checking invalid certificates consistently across notarization, certification, nullification, and finalization paths

### DIFF
--- a/consensus/src/simplex/mocks/reporter.rs
+++ b/consensus/src/simplex/mocks/reporter.rs
@@ -111,6 +111,11 @@ where
         });
     }
 
+    pub fn assert_no_faults(&self) {
+        let faults = self.faults.lock();
+        assert!(faults.is_empty(), "faults detected");
+    }
+
     pub fn assert_no_invalid(&self) {
         let invalid_votes = self.invalid_votes.lock();
         let invalid_certificates = self.invalid_certificates.lock();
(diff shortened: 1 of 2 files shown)


consensus/src/simplex/mocks/reporter.rs
```diff
@@ -59,7 +59,8 @@ pub struct Reporter<E: CryptoRngCore, S: Scheme, L: ElectorConfig<S>, D: Digest>
     pub finalizes: Arc<Mutex<Participation<S::PublicKey, D>>>,
     pub finalizations: Arc<Mutex<HashMap<View, Finalization<S, D>>>>,
     pub faults: Arc<Mutex<Faults<S, D>>>,
-    pub invalid: Arc<Mutex<usize>>,
+    pub invalid_votes: Arc<Mutex<usize>>,
+    pub invalid_certificates: Arc<Mutex<usize>>,
```
BrendanChou, line 63: track these separately for better counting
could also split out the certificate types but that seemed overkill for now

consensus/src/simplex/mod.rs
```diff
@@ -2965,15 +2958,8 @@ mod tests {
                 );
                 actor.start();
                 let blocker = oracle.control(validator.clone());
-                let leader_timeout = if idx_scheme == 0 {
-                    // Force the wrong-namespace byzantine node to timeout quickly so
```
patrick-ogrady, line 2969: ❤️

PR #3487, opened 2026-03-27
[runtime] avoid stack multiplication in Handle::init

### LINKED ISSUE #3474: [runtime] Consider not boxing spawned futures by default

andresilva:
The current `Handle::init` implementation wraps spawned futures using a chain of combinators: an async block for panic catching, `Abortable::new()` for abort support, and `.map()` for cleanup. In debug builds, rustc does not inline functions, reuse stack slots for moved-from variables, or optimize async state machine layouts. This means each combinator in the chain created a separate local in `init`'s stack frame, each at least `size_of(F)` large. Combined with the separate call frames for `Abortable::new` and `Map::new`, the total stack usage was roughly 6x `size_of(F)`. For large async state machines like the marshal actor in alto, this overflows the default tokio worker thread stack.

This PR replaces the combinator chain with a single async block that handles panic catching, abort, and cleanup inline. The `Abortable` wrapper is constructed inside the async block rather than as a separate local in `init`'s stack frame, so it becomes part of the async state machine's internal state rather than an additional stack allocation. The result is stack usage of `size_of(F) + constant` instead of `6 * size_of(F)`.

`init` is also marked `#[inline(always)]` so that even in debug builds, its locals are merged into the caller's stack frame, avoiding the extra copy from the function call boundary.

This approach avoids boxing entirely. The spawned future's concrete type is preserved through the entire chain (via `Either<Instrumented<F>, F>`), so the compiler can still monomorphize and inline through `poll` calls. No heap allocation, no dynamic dispatch, no vtable indirection beyond what tokio already does internally.

The key invariant is: if the user's future `F` fits on the stack when they construct it, our wrapping will not blow it up. If `F` itself is too large for the stack, that's a problem at the call site before `spawn` is ever reached, nothing the runtime can do about that.

Fixes #3474.

### DIFF
--- a/runtime/src/utils/handle.rs
+++ b/runtime/src/utils/handle.rs
@@ -6,7 +6,7 @@ use commonware_utils::{
 use futures::{
     future::{select, Either},
     pin_mut,
-    stream::{AbortHandle, Abortable},
+    stream::{AbortHandle, Abortable, Aborted},
     FutureExt as _,
 };
 use prometheus_client::metrics::gauge::Gauge;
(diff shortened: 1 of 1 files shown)


andresilva:
I tested that this fixes the regression on alto, but maybe someone else can double check to be extra safe?

patrick-ogrady:
Tried a few times to get a regression test for this but it was brittle/noisy.

PR #3488, opened 2026-03-27
[runtime/tokio] don't use deprecated set_linger

andresilva:
Tokio 1.50 deprecated `TcpStream::set_linger` because non-zero linger durations cause the socket to block the calling thread on drop. The replacement is `set_zero_linger()`, which only supports the zero-duration case. Since non-zero linger is fundamentally incompatible with async runtimes, this is a reasonable constraint.

This PR bumps tokio from 1.43 to 1.50, switches all call sites to `set_zero_linger()`, and narrows the network config from `so_linger: Option<Duration>` to `zero_linger: bool`.

### DIFF
--- a/runtime/src/network/tokio.rs
+++ b/runtime/src/network/tokio.rs
@@ -229,7 +229,7 @@ impl Default for Config {
             tcp_nodelay: Some(true),
             zero_linger: true,
             read_timeout: Duration::from_secs(60),
-            write_timeout: Duration::from_secs(30),
+            write_timeout: Duration::from_secs(60),
             read_buffer_size: 64 * 1024, // 64 KB
         }
     }


runtime/src/network/iouring.rs
```diff
@@ -80,7 +78,7 @@ impl Default for Config {
         let iouring_config = iouring::Config::default();
         Self {
             tcp_nodelay: None,
-            so_linger: None,
+            zero_linger: false,
```
patrick-ogrady: We want to default true here?

runtime/src/network/tokio.rs
```diff
@@ -229,7 +227,7 @@ impl Default for Config {
     fn default() -> Self {
         Self {
             tcp_nodelay: None,
-            so_linger: None,
+            zero_linger: false,
```
patrick-ogrady: set true?

runtime/src/network/tokio.rs
```diff
@@ -229,7 +227,7 @@ impl Default for Config {
     fn default() -> Self {
         Self {
             tcp_nodelay: None,
```
patrick-ogrady: set some(True)?

runtime/src/network/iouring.rs
```diff
@@ -80,7 +78,7 @@ impl Default for Config {
         let iouring_config = iouring::Config::default();
         Self {
             tcp_nodelay: None,
```
patrick-ogrady: set some(true)?

runtime/src/network/tokio.rs
```diff
@@ -228,8 +226,8 @@ impl Config {
 impl Default for Config {
     fn default() -> Self {
         Self {
-            tcp_nodelay: None,
-            so_linger: None,
+            tcp_nodelay: Some(true),
```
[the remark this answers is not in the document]
patrick-ogrady: This was done explicitly to align defaults.

PR #3489, opened 2026-03-27
[Storage] Add `append_many` journal optimization

danlaine:
## Summary

- Add `append_many` to the contiguous journal `Mutable` trait, reducing lock acquisitions from N to 1 when appending multiple items
- `authenticated::Journal::apply_batch` now calls `append_many` per segment instead of `append` per item

## Detail

**Problem**: `authenticated::Journal::apply_batch` iterates over all operations in a batch and calls `self.journal.append(item)` for each one. Every `append` call independently acquires the contiguous journal's write lock (`UpgradableAsyncRwLock`), computes the section, appends, and releases the lock. For a batch with N operations, this means N lock acquire/release cycles. For variable journals it's worse -- each `append` also acquires a second lock on the internal offsets journal, so it's 2N lock cycles. In QMDB, a single `apply_batch` can contain dozens of operations (user mutations + floor raises + commit), making this a hot path.

**Fix**: `append_many(&self, items: &[Item])` acquires the write lock once for all items. The existing `append` now delegates to `append_many` via `std::slice::from_ref`, eliminating duplicated logic. Section-full sync handling (downgrade lock, sync, upgrade, create next section) is preserved within the loop. Empty slices no-op and return the current size.

**`Mutable` trait**: `append_many` is added with a default implementation that calls `append` in a loop, so existing implementors are not broken. Fixed and variable journals override with the optimized single-lock version.

**`apply_batch`**: Now calls `self.journal.append_many(items)` per segment instead of `self.journal.append(item)` per item. Since segments are `Arc<Vec<Item>>`, they deref directly to `&[Item]` with no additional allocation.

### DIFF
--- a/storage/src/journal/authenticated.rs
+++ b/storage/src/journal/authenticated.rs
@@ -379,7 +379,7 @@ where
             self.journal.append_many(items).await?;
         }
         self.mmr.apply(batch.changeset)?;
-        debug_assert_eq!(*self.mmr.leaves(), self.journal.size().await);
+        assert_eq!(*self.mmr.leaves(), self.journal.size().await);
         Ok(())
     }
 
--- a/storage/src/journal/contiguous/mod.rs
+++ b/storage/src/journal/contiguous/mod.rs
@@ -88,6 +88,28 @@ pub trait Mutable: Contiguous + Send + Sync {
         item: &Self::Item,
     ) -> impl std::future::Future<Output = Result<u64, Error>> + Send;
 
+    /// Append multiple items to the journal, returning the position of the last item appended.
+    ///
+    /// The default implementation calls [Self::append] in a loop. Concrete implementations
+    /// may override this to acquire the write lock once for all items.
+    ///
+    /// No-ops if items is empty, returning the current size (next append position).
+    fn append_many(
+        &mut self,
+        items: &[Self::Item],
+    ) -> impl std::future::Future<Output = Result<u64, Error>> + Send
+    where
+        Self::Item: Sync,
+    {
+        async move {
+            let mut last_pos = self.size().await;
+            for item in items {
+                last_pos = self.append(item).await?;
+            }
+            Ok(last_pos)
+        }
+    }
+
     /// Prune items at positions strictly less than `min_position`.
     ///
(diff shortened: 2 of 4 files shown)


storage/src/journal/contiguous/mod.rs
```diff
@@ -131,6 +131,28 @@ pub trait Mutable: Contiguous + Send + Sync {
     /// Returns an error if the underlying storage operation fails.
     fn rewind(&mut self, size: u64) -> impl std::future::Future<Output = Result<(), Error>> + Send;
 
+    /// Append multiple items to the journal, returning the position of the last item appended.
+    ///
+    /// The default implementation calls [Self::append] in a loop. Concrete implementations
+    /// may override this to acquire the write lock once for all items.
+    ///
+    /// No-ops if items is empty, returning the current size (next append position).
+    fn append_many(
```
patrick-ogrady: Move after append

storage/src/journal/contiguous/variable.rs
```diff
@@ -509,34 +509,48 @@ impl<E: Context, V: CodecShared> Journal<E, V> {
...
-        assert_eq!(offsets_pos, inner.size);
+            // Append offset to offsets journal.
+            let offsets_pos = self.offsets.append(&offset).await?;
+            assert_eq!(offsets_pos, inner.size);
 
-        // Return the current position
-        let position = inner.size;
-        inner.size += 1;
+            // Return the current position.
+            let position = inner.size;
```
patrick-ogrady: Do we need the intermediate `let position` variable?

storage/src/journal/authenticated.rs
```diff
@@ -376,9 +376,7 @@ where
         }
 
         for items in &batch.items {
-            for item in items.iter() {
-                self.journal.append(item).await?;
-            }
+            self.journal.append_many(items).await?;
         }
         self.mmr.apply(batch.changeset)?;
         debug_assert_eq!(*self.mmr.leaves(), self.journal.size().await);
```
patrick-ogrady: Change to `assert_eq`

patrick-ogrady (approved):
Added follow-up work in #3492

PR #3491, opened 2026-03-27
[Codec] [Runtime] [p2p] Multi-buffer encoding for zero-copy of `Bytes` fields

danlaine:
## Problem

When encoding a message for the network, `Write::write()` copies every field into a single contiguous buffer. For types containing large `Bytes` fields -- like `Chunk`, which holds up to 1 MB of erasure-coded shard data -- this memcpy is unnecessary. The data is already in an Arc-backed `Bytes` that could be passed by reference, but `Write` targets `impl BufMut` (a contiguous buffer) and has no way to express "just reference this allocation."

The network layer downstream already supports multi-buffer output (`IoBufs`), but the encoding layer cannot produce them.

The most affected path is **shard broadcasting**: every consensus round, the proposer encodes one shard (~1 MB) per participant. Each encoding memcpy's the shard payload even though it is already in memory.

Note that the encryption layer also copies the payload into a contiguous buffer (the cipher requires contiguous memory). That copy is unavoidable. Today the payload is copied twice (encoding + encryption); this PR reduces it to once (encryption only).

## Solution

**`BufsMut` trait and `Write::write_bufs`** (`codec/src/codec.rs`, `codec/src/types/bytes.rs`):

Following the `IoBuf`/`IoBufs` naming convention, `BufsMut: BufMut` is a multi-buffer target that accepts zero-copy `Bytes` via `push()`. `Write::write_bufs` is a default method on the existing `Write` trait that delegates to `Write::write`. Types containing large `Bytes` fields override it to call `push()` instead of copying.

The `Bytes` override writes the varint length prefix inline and pushes the payload via `push()` (Arc clone, no memcpy). The wire format is identical.

**`EncodeSize::encode_inline_size`** (`codec/src/codec.rs`):

New default method on `EncodeSize` returning the encoded size excluding bytes that go via `BufsMut::push`. Used to right-size the `Builder`'s working buffer. Types that override `write_bufs` should also override `encode_inline_size`; failing to do so over-allocates but is not incorrect.

**`Builder` in runtime** (`runtime/src/iobuf/mod.rs`):

The concrete `BufsMut` implementation (re-exported as `IoBufsBuilder`). All inline writes go into a single pool-backed buffer. `push()` records boundaries without flushing. `finish()` freezes the buffer once and uses `IoBuf::slice` to carve it into pieces at the recorded boundaries, interleaved with the pushed `Bytes`. All inline slices share the same underlying pool allocation.

If inline writes exceed the initial capacity, the buffer grows via reallocation.

**`encode_with_pool` now returns `IoBufs`** (`runtime/src/iobuf/mod.rs`):

`EncodeExt::encode_with_pool` is changed from returning a flat `IoBuf` (via `Write::write`) to returning `IoBufs` (via `Write::write_bufs` through a `Builder`). The `Builder` is sized to `encode_inline_size()`. For types that don't override `write_bufs`, the output is a single contiguous piece -- same as before. For types that do, large `Bytes` fields become zero-copy pieces. All downstream consumers already accept `impl Into<IoBufs>`, so this is transparent.

**`write_bufs` and `encode_inline_size` overrides for hot-path types**:

Each type in the chain from `WrappedSender::send` to the `Bytes` field overrides both methods in its `Write` and `EncodeSize` impls. Types without `Bytes` fields inherit the defaults. The encoding chain for a shard:

```
Shard::write_bufs -> Chunk::write_bufs -> Bytes::write_bufs (push)
```

Produces:

```
Piece 0: [commitment | shard_index | varint_length]  (~40 bytes, pool slice)
Piece 1: [------------ 1 MB shard payload ---------]  (Arc clone, no copy)
Piece 2: [chunk_index | proof]                        (~200 bytes, pool slice)
```

Pieces 0 and 2 are slices of a single ~240 byte pool allocation.

### DIFF
--- a/codec/src/codec.rs
+++ b/codec/src/codec.rs
@@ -46,9 +46,9 @@ pub trait Write {
     /// Implementations should panic if the buffer doesn't have enough capacity.
     fn write(&self, buf: &mut impl BufMut);
 
-    /// Writes to a [BufsMut], allowing zero-copy for [Bytes] fields via
-    /// [`BufsMut::push`]. Must encode to the same format as [Write::write].
-    /// Defaults to [Write::write].
+    /// Writes to a [`BufsMut`], allowing existing [`Bytes`] chunks to be
+    /// appended via [`BufsMut::push`] instead of written inline. Must encode
+    /// to the same format as [`Write::write`]. Defaults to [`Write::write`].
     fn write_bufs(&self, buf: &mut impl BufsMut) {
         self.write(buf);
     }
@@ -209,9 +209,10 @@ pub trait CodecFixedShared: CodecFixed<Cfg = ()> + Send + Sync {}
 // Automatically implement `CodecFixedShared` for types that meet all bounds.
 impl<T: CodecFixed<Cfg = ()> + Send + Sync> CodecFixedShared for T {}
 
-/// A [BufMut] that can also accept pre-existing [Bytes] without copying.
+/// A [`BufMut`] that can also append pre-existing [`Bytes`] chunks.
 pub trait BufsMut: BufMut {
-    /// Append bytes by reference, without copying the underlying data.
+    /// Appends a [`Bytes`] chunk instead of writing its contents inline into
+    /// the destination buffer.
     fn push(&mut self, bytes: impl Into<Bytes>);
 }
 


codec/src/codec.rs
```diff
@@ -194,6 +194,34 @@ pub trait CodecFixedShared: CodecFixed<Cfg = ()> + Send + Sync {}
...
+///
+/// Normal [BufMut] writes (like `put_u32`, `put_slice`) go into an inline working buffer.
+/// [`push_segment`](SegmentedBuf::push_segment) adds an existing [Bytes] without copying its
+/// contents.
+pub trait SegmentedBuf: BufMut {
+    /// Add a pre-existing [Bytes] as a separate segment without copying.
+    ///
+    /// Any bytes previously written inline are flushed as their own segment first,
+    /// then `segment` is appended as a new segment via reference count (Arc clone).
+    fn push_segment(&mut self, segment: Bytes);
```
patrick-ogrady: I wouldn't link it directly to Bytes? I think it could be any IoBuf?

danlaine: Using `IoBuf` would create a circular dependency between commonware-codec and commonware-runtime. I'll change it to `Into<Bytes>` to make it more ergonomic.

runtime/src/iobuf/mod.rs
```diff
@@ -2208,14 +2333,26 @@ pub trait EncodeExt: EncodeSize + Write {
...
     ///
     /// Panics if [`EncodeSize::encode_size`] does not match the number of
-    /// bytes written by [`Write::write`].
-    fn encode_with_pool(&self, pool: &BufferPool) -> IoBuf {
-        self.encode_with_pool_mut(pool).freeze()
+    /// bytes written by [`Write::write_bufs`].
+    fn encode_with_pool(&self, pool: &BufferPool) -> IoBufs {
+        let len = self.encode_size();
+        let capacity = NonZeroUsize::new(len).unwrap_or(NonZeroUsize::MIN);
+        let mut builder = Builder::new(pool, capacity);
```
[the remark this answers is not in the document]
danlaine: resolved in https://github.com/commonwarexyz/monorepo/pull/3491/commits/84849736cc146bd239cc73e37c33b606a880b0bc

danlaine:
> Right now `Builder::push()` flushes the current inline buffer and later allocates again, so `encode_bufs_size()` does not quite buy "one allocation for all inline data" for layouts like `[inline prefix][Bytes payload][inline trailer]`. I would suggest renaming `encode_bufs_size()` to `encode_inline_size()` or similar, since that seems closer to what it actually means.
> 
> The builder should fully exploit it: allocate a single inline buffer of that size up front, keep writing all inline bytes into that same backing allocation, and have `push()` record boundaries instead of flushing. Concretely, `push()` could append the current inline range plus the external `Bytes` chunk to a small piece list, then continue writing inline bytes into the same buffer. At `finish()`, the builder would `freeze()` once and turn the recorded inline ranges into zero-copy `IoBuf::slice(...)` views, interleaved with the pushed `Bytes`.
> 
> That also feels cleaner from a `BufMut` perspective. As written, `Builder` is a fairly non-standard `BufMut`: `remaining_mut()` only reflects the current scratch buffer, but some write paths (`put_slice`, `put`, `chunk_mut`) can recover from exhaustion by flushing that buffer and allocating a fresh one. So the effective writable capacity depends on which `BufMut` method gets called, and the impl has to override selected defaults to make that work. I am not claiming a concrete unsoundness here, but it makes the `unsafe impl BufMut` much harder to reason about than a single-allocation design where `remaining_mut()` really means the remaining writable inline capacity.

I've changed the PR by:
* Renaming `encode_bufs_size` to `encode_inline_size`
* Allocating single inline buffer up front and recording boundaries instead of flushing

codec/src/codec.rs
```diff
@@ -37,6 +45,12 @@ pub trait Write {
     ///
     /// Implementations should panic if the buffer doesn't have enough capacity.
     fn write(&self, buf: &mut impl BufMut);
+
+    /// Writes to a [BufsMut], allowing zero-copy for [Bytes] fields via
+    /// [`BufsMut::push`]. Defaults to [Write::write].
```
andresilva: Even though it's somewhat obvious, maybe should mention that this should encode to the same format as `write`?

coding/src/reed_solomon.rs
```diff
@@ -125,6 +125,12 @@ impl<D: Digest> Write for Chunk<D> {
         self.index.write(writer);
         self.proof.write(writer);
     }
+
+    fn write_bufs(&self, buf: &mut impl BufsMut) {
```
andresilva, line 129: Add test that `Chunk::encode` and `Chunk::encode_with_pool` return the same payload?

consensus/src/marshal/coding/types.rs
```diff
@@ -75,12 +75,22 @@ impl<C: Scheme, H: Hasher> Write for Shard<C, H> {
         self.index.write(buf);
         self.inner.write(buf);
     }
+
+    fn write_bufs(&self, buf: &mut impl BufsMut) {
```
andresilva, line 79: Add test that `Shard::encode` and `Shard::encode_with_pool` return the same payload?

resolver/src/p2p/wire.rs
```diff
@@ -18,12 +18,21 @@ impl<Key: Span> Write for Message<Key> {
         buf.put_u64(self.id);
         self.payload.write(buf);
     }
+
+    fn write_bufs(&self, buf: &mut impl BufsMut) {
```
andresilva, line 22: Add test that `Message::encode` and `Message::encode_with_pool` return the same payload?

resolver/src/p2p/wire.rs
```diff
@@ -79,6 +88,22 @@ impl<Key: Span> Write for Payload<Key> {
             }
         }
     }
+
+    fn write_bufs(&self, buf: &mut impl BufsMut) {
```
andresilva, line 92: Add test that `Payload::encode` and `Payload::encode_with_pool` return the same payload (for `Response`)?

runtime/src/iobuf/mod.rs
```diff
@@ -2185,6 +2185,143 @@ where
...
+    // Single working buffer for all inline writes.
+    buf: IoBufMut,
+    // Each entry is (offset_in_buf, pushed_bytes) recording where a push
+    // interrupts the inline byte stream.
+    pushes: Vec<(usize, Bytes)>,
+}
+
+impl Builder {
+    /// `capacity` sets the size of the pool-allocated inline buffer.
+    /// If inline writes exceed capacity, the buffer grows via reallocation.
```
andresilva: I guess reallocation might make sense for the general purpose builder abstraction, but for our usage this can mask bad implementations of `encode_inline_size()`.

codec/src/codec.rs
danlaine: Done

coding/src/reed_solomon.rs
danlaine: Done

resolver/src/p2p/wire.rs
danlaine: Done

resolver/src/p2p/wire.rs
danlaine: Done

consensus/src/marshal/coding/types.rs
danlaine: Done

runtime/src/iobuf/mod.rs
danlaine: Removed `grow()` from `Builder` -- now it will panic if the capacity is exceeded.

andresilva (dismissed):

resolver/src/p2p/wire.rs
```diff
@@ -18,12 +18,21 @@ impl<Key: Span> Write for Message<Key> {
...
+        self.payload.write_bufs(buf);
+    }
 }
 
 impl<Key: Span> EncodeSize for Message<Key> {
     fn encode_size(&self) -> usize {
         self.id.encode_size() + self.payload.encode_size()
     }
+
+    fn encode_inline_size(&self) -> usize {
```
patrick-ogrady, line 33: I almost think this should be called `encode_copied_size` or `encode_bufs_size`?

`inline_size` makes it sound like the item we are "inling" size? for clarity, I think encode_bufs_size is probably most "obvious".

danlaine: This was `encode_bufs_size` [before](https://github.com/commonwarexyz/monorepo/pull/3491#pullrequestreview-4037698982) -- I don't feel super strongly on the naming of this one so I'll defer to you

andresilva: That was the original name, and I suggested the inline naming. I think bufs doesn't really capture the nuance of what this size refers to.

patrick-ogrady: Will think about some more. Thanks for the context.

consensus/src/marshal/coding/types.rs
```diff
@@ -75,12 +75,22 @@ impl<C: Scheme, H: Hasher> Write for Shard<C, H> {
...
+        self.commitment.write(buf);
+        self.index.write(buf);
+        self.inner.write_bufs(buf);
```
patrick-ogrady, line 82: I considered converting everything inside of every `write_bufs` to also use `write_bufs` (as it doesn't add overhead/protects us in the future from optimizations below).

I decided to leave as-is for now.

resolver/src/p2p/wire.rs
```diff
@@ -18,12 +18,21 @@ impl<Key: Span> Write for Message<Key> {
         buf.put_u64(self.id);
         self.payload.write(buf);
     }
+
+    fn write_bufs(&self, buf: &mut impl BufsMut) {
+        self.id.write(buf);
+        self.payload.write_bufs(buf);
```
patrick-ogrady, line 24: This alone is huge

runtime/src/iobuf/mod.rs
```diff
@@ -2185,6 +2196,106 @@ where
...
+
+    /// Freezes the inline buffer and assembles [`IoBufs`] by slicing at
+    /// the recorded push boundaries.
+    pub fn finish(self) -> IoBufs {
+        if self.pushes.is_empty() {
+            return IoBufs::from(self.buf.freeze());
+        }
+
+        let frozen = self.buf.freeze();
+        let mut result = IoBufs::default();
```
patrick-ogrady, line 2250: Opted to not pre-allocate here because we don't know exactly how many items we'll add to result until we step through. These are pointers, so are easy to move around in any case.

patrick-ogrady (approved):

Issue #3493, opened 2026-03-28
[runtime/iouring] Flaky Test


patrick-ogrady:
Related: #3489

```
        FAIL [   0.298s] (246/526) commonware-runtime network::iouring::tests::test_trait
  stdout ───

    running 1 test
    test network::iouring::tests::test_trait ... FAILED

    failures:

    failures:
        network::iouring::tests::test_trait

    test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 511 filtered out; finished in 0.29s
    
  stderr ───

    thread 'network::iouring::tests::test_trait' (15029) panicked at runtime/src/network/mod.rs:297:18:
    Failed to dial server: ConnectionFailed
    note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

    thread 'network::iouring::tests::test_trait' (15029) panicked at runtime/src/network/mod.rs:330:9:
    assertion failed: client_result.is_ok()

  Cancelling due to test failure: 3 tests still running
        SLOW [> 60.000s] (─────────) commonware-reshare::bin/commonware-reshare validator::test::dkg_deterministic_slow_
        SLOW [> 60.000s] (─────────) commonware-reshare::bin/commonware-reshare validator::test::dkg_with_many_forced_failures_slow_
        PASS [  63.962s] (247/526) commonware-reshare::bin/commonware-reshare validator::test::dkg_with_many_forced_failures_slow_
        SLOW [> 60.000s] (─────────) commonware-reshare::bin/commonware-reshare validator::test::reshare_four_epochs_with_total_shutdown_slow_
        PASS [ 118.065s] (248/526) commonware-reshare::bin/commonware-reshare validator::test::dkg_deterministic_slow_
        PASS [ 109.853s] (249/526) commonware-reshare::bin/commonware-reshare validator::test::reshare_four_epochs_with_total_shutdown_slow_
────────────
     Summary [ 671.555s] 249/526 tests run: 248 passed (9 slow), 1 failed, 3779 skipped
        FAIL [   0.298s] (246/526) commonware-runtime network::iouring::tests::test_trait
warning: 277/526 tests were not run due to test failure (run with --no-fail-fast to run all tests, or run with --max-fail)
error: test run failed
```

patrick-ogrady:
It seems related to the changes made here (likely for iouring): https://github.com/commonwarexyz/monorepo/pull/3488

PR #3494, opened 2026-03-28
[runtime] Fix Network Flake

patrick-ogrady:
Closes: #3493 
Related: #3488

### DIFF
--- a/runtime/src/network/mod.rs
+++ b/runtime/src/network/mod.rs
@@ -49,40 +49,48 @@ mod tests {
         // Get the local address of the listener
         let listener_addr = listener.local_addr().expect("Failed to get local address");
 
-        // Wait to drop any open sockets until both sides have completed.
-        join!(
+        // Spawn server. Returning the socket halves keeps them alive until both
+        // join handles are awaited below.
+        let server = tokio::spawn(async move {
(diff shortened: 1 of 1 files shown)


runtime/src/network/mod.rs
```diff
@@ -61,6 +66,9 @@ mod tests {
             sink.send(IoBuf::from(SERVER_SEND_DATA))
```
patrick-ogrady, line 62: Dropping the socket here now sends an RST immediately, so the `recv` can fail. While it caused this test to fail, it is what we ~want by default.

runtime/src/network/mod.rs
patrick-ogrady, line 379: This also needs a fix?

runtime/src/network/mod.rs
```diff
@@ -189,6 +213,7 @@ mod tests {
                     .expect("Failed to receive data");
                 // Verify the received data
                 assert_eq!(received.coalesce(), SERVER_SEND_DATA);
+                read_done_tx.send(()).expect("Failed to notify server");
```
patrick-ogrady: This makes it feel synchronous (need to do a barrier instead)

patrick-ogrady:
bugbot run

patrick-ogrady:
bugbot run

PR #3497, opened 2026-03-29
[consensus/marshal] Avoid materializing blocks for availability checks

### LINKED ISSUE #3393: [consensus/marshal] Add Support for "Availability" Requests
Related: #3390

In some cases, all we care about is whether marshal has a block (rather than the contents of the block itself). We should add a new method that allows us to check this without actually fetching/materializing the block.

vikions:
Closes: #3393

## Summary

- add an availability-only marshal subscription by digest
- teach marshal storage and buffer backends to answer cheap availability checks without materializing blocks
- update `standard::Inline::certify` to wait on block availability instead of fetching the full block
- add a focused buffered availability test

## Detail

`Inline::certify` only needs to know whether marshal has a block, not the block contents. This change adds a lightweight availability path through marshal core and the backing buffer implementations so certification can wait on availability without materializing the block.

To support that path cleanly, marshal now keeps separate availability subscriptions, cache/finalized storage can answer existence checks via `has(...)`, and both the standard buffered backend and coding shards backend expose availability-only subscriptions by digest.

### DIFF
--- a/broadcast/src/buffered/engine.rs
+++ b/broadcast/src/buffered/engine.rs
@@ -16,7 +16,7 @@ use commonware_utils::{
     channel::{fallible::OneshotExt, mpsc, oneshot},
     ordered::Set,
 };
-use std::collections::{BTreeMap, VecDeque};
+use std::collections::{BTreeMap, BTreeSet, VecDeque};
 use tracing::{debug, error, trace, warn};
 
 /// A responder waiting for a message.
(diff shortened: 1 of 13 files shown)


broadcast/src/buffered/engine.rs
```diff
@@ -75,6 +81,9 @@ where
     /// Pending requests from the application.
     waiters: BTreeMap<M::Digest, Vec<Waiter<M>>>,
```
patrick-ogrady, line 79: Is there a way to extend `Waiter` to support this instead?

vikions: I reworked this to extend the existing waiter path instead of keeping a separate availability waiter map. Waiter now carries either a full-message responder or an availability-only responder, and I added a focused test that exercises both subscription kinds on the same digest. Happy to reshape it further if you had a different direction in mind.

clabby:

Issue #3270, opened 2026-02-28
[p2p] Support "External" Participants


patrick-ogrady:
Allow nodes to register unsynchronized participants that can dial a node but will never be dialed. Likewise, these external participants will not be tracked in a peer set and will never be "requested" from (by things like the resolver).

They will receive gossip sent via ALL, receive messages sent directly to them, and can send requests to the node that registered them.

bajpainaman:
@patrick-ogrady  For indexer nodes ?

patrick-ogrady:
It will work well both for nodes indexing blockchain data and submitting certificates/transactions to validators for block inclusion.

patrick-ogrady:
The issue that will come up with this is how to maintain the peer connection in `discovery` (the only ping-pong is the BitVec-Peers and we likely don't want these external nodes participating in discovery unnecessarily?) ... the path for `simulated` and `lookup` is pretty clear tho.

bajpainaman:
Thanks for closing this !!!! n

PR #3401, opened 2026-03-16
[p2p/lookup + simulated] Support "External" Participants

patrick-ogrady:
Related: #3270

PR #3499, opened 2026-03-30
[p2p] Support Primary/Secondary Peers

patrick-ogrady:
Closes: #3270
Replaces: #3401

### DIFF
--- a/examples/reshare/src/dkg/actor.rs
+++ b/examples/reshare/src/dkg/actor.rs
@@ -302,9 +302,10 @@ where
                 )
             };
 
-            // Primary peers are the dealers for the active epoch. Secondary peers are the
-            // active-epoch players and the next players we allow to connect early.
-            // Dealers may also appear among players; `track` keeps primary role only.
+            // Primary = dealers (drive the DKG round/running consensus)
+            // Secondary = current players + next-epoch players (give time to sync)
+            //
+            // Overlapping keys are deduplicated as primary (so we don't need to do any filtering here)
             self.manager
                 .track(
                     epoch.get(),


examples/reshare/src/dkg/actor.rs
```diff
@@ -302,18 +302,16 @@ where
...
-                    Set::from_iter_dedup(
-                        dealers
-                            .iter()
-                            .cloned()
-                            .chain(players.iter().cloned())
-                            .chain(next_players),
+                    TrackedPeers::new(
+                        dealers.clone(),
+                        Set::from_iter_dedup(
+                            players.iter().cloned().chain(next_players.iter().cloned()),
```
patrick-ogrady: by this logic, we should allow duplicates to be added here?

p2p/src/authenticated/discovery/actors/tracker/record.rs
```diff
@@ -57,8 +57,11 @@ pub struct Record<C: PublicKey> {
     /// Connection status of the peer.
     status: Status,
 
-    /// Number of peer sets this peer is part of.
-    sets: usize,
+    /// Number of primary peer sets this peer is part of.
+    primary_sets: usize,
```
patrick-ogrady, line 61: just call this `primaries`?

p2p/src/simulated/network.rs
```diff
@@ -391,7 +404,7 @@ impl<E: RNetwork + Spawner + Rng + Clock + Metrics, P: PublicKey> Network<E, P>
                 let (mut sender, receiver) = ring::channel(NZUsize!(1));
 
                 // Send current peer list immediately
-                let peer_list: Vec<P> = self.all_tracked_peers().into_iter().collect();
+                let peer_list: Vec<P> = self.all_connectable_peers().into_iter().collect();
```
patrick-ogrady: Ensure there exists a test where we don't duplicate sends if a peer is both primary/secondary.

bajpainaman:
W addition !

p2p/src/authenticated/discovery/actors/tracker/directory.rs
```diff
@@ -131,6 +131,15 @@ impl<E: Spawner + Rng + Clock + RuntimeMetrics, C: PublicKey> Directory<E, C> {
 
     // ---------- Setters ----------
 
+    /// Track a new primary peer set and replace the current secondaries.
+    pub fn track(&mut self, index: u64, primary: OrderedSet<C>, secondary: OrderedSet<C>) -> bool {
+        if !self.add_set(index, primary) {
+            return false;
+        }
+        self.set_secondaries(secondary);
```
patrick-ogrady: this should be indexed just like primary?

p2p/src/authenticated/discovery/actors/tracker/directory.rs
```diff
@@ -199,64 +208,91 @@ impl<E: Spawner + Rng + Clock + RuntimeMetrics, C: PublicKey> Directory<E, C> {
...
+            });
+            record.increment_secondary();
+        }
+        self.secondary_sets.insert(index, secondaries);
+
+        // Remove oldest tracked peer sets if necessary.
+        while self.primary_sets.len() > self.max_sets {
+            let (primary_index, primary_set) = self.primary_sets.pop_first().unwrap();
+            let (secondary_index, secondary_set) = self.secondary_sets.pop_first().unwrap();
+            debug_assert_eq!(primary_index, secondary_index);
```
patrick-ogrady: make this explicit assert

patrick-ogrady:
We should prioritize primary traffic.

patrick-ogrady:
Need to add support for surfacing primary/secondary to marshal codec (should return tuple during subscription for those tracking all connected peers).

patrick-ogrady:
Ensure a peer only ends up in one of primary/secondary in subscribe (preferring primary).

broadcast/src/buffered/engine.rs
```diff
@@ -210,11 +210,14 @@ where
...
-            Some((_, _, tracked_peers)) = peer_set_subscription.recv() else {
+            Some(update) = peer_set_subscription.recv() else {
                 debug!("peer set subscription closed");
                 break;
             } => {
-                self.evict_untracked_peers(&tracked_peers);
+                // Buffered caches are keyed to the newest primary set rather than the aggregate
+                // overlap window. Once a sender leaves `latest.primary`, stop retaining its deque
+                // even if transport tracking still keeps the peer connected.
+                self.evict_untracked_peers(&update.latest.primary);
```
patrick-ogrady: This is a behavior change to only use latest.

resolver/src/p2p/engine.rs
```diff
@@ -195,15 +195,16 @@ impl<
...
-                // reconcile with all tracked peers).
-                if self.last_peer_set_id < Some(id) {
-                    self.last_peer_set_id = Some(id);
-                    self.fetcher.reconcile(all.as_ref());
+                if self.last_peer_set_id < Some(update.index) {
+                    self.last_peer_set_id = Some(update.index);
+                    // Outbound fetches intentionally follow only the newest primary set.
+                    // `update.all` keeps overlap-window peers connected, but once a peer leaves
+                    // `latest.primary` it should stop receiving new resolver traffic.
+                    self.fetcher.reconcile(update.latest.primary.as_ref());
```
patrick-ogrady, line 204: This is a behavior change to only use latest.

patrick-ogrady:
bugbot run

patrick-ogrady:
bugbot run

p2p/src/authenticated/discovery/actors/tracker/actor.rs
```diff
@@ -150,21 +151,37 @@ impl<E: Spawner + Rng + Clock + RuntimeMetrics, C: Signer> Actor<E, C> {
     async fn handle_msg(&mut self, msg: Message<C::PublicKey>) {
         match msg {
             Message::Register { index, peers } => {
-                // Ensure that peer set is not too large.
+                let primary = peers.primary;
```
patrick-ogrady: move under comment

p2p/src/authenticated/discovery/actors/tracker/actor.rs
```diff
@@ -150,21 +151,37 @@ impl<E: Spawner + Rng + Clock + RuntimeMetrics, C: Signer> Actor<E, C> {
...
+                // Secondary peers are not checked here because max_peer_set_size
+                // exists to cap the bitvec size, which only covers primary peers.
                 let max = self.max_peer_set_size;
-                assert!(len as u64 <= max, "peer set too large: {len} > {max}");
-
-                // Attempt to add the peer set
-                if !self.directory.add_set(index, peers.clone()) {
+                let plen = primary.len();
+                assert!(
+                    plen as u64 <= max,
```
patrick-ogrady: nit: len

patrick-ogrady:
bugbot run

broadcast/src/buffered/engine.rs
```diff
@@ -147,6 +151,9 @@ where
             network.0,
             network.1,
         );
+        // Subscribe immediately, but keep draining mailbox and network traffic even if no peer set
```
patrick-ogrady: this is unnecessary

broadcast/src/buffered/engine.rs
```diff
@@ -232,7 +240,8 @@ where
         responder: oneshot::Sender<Vec<P>>,
     ) {
         // Store the message, continue even if it was already stored
-        let _ = self.insert_message(self.public_key.clone(), msg.clone());
+        let digest = msg.digest();
+        self.insert_message(self.public_key.clone(), digest, msg.clone());
```
patrick-ogrady: why do we need to provide digest?

patrick-ogrady:
bugbot run

consensus/src/marshal/coding/shards/engine.rs
```diff
@@ -438,13 +457,49 @@ where
...
                 return;
             } => {
-                self.peer_buffers
-                    .retain(|peer, _| tracked_peers.as_ref().contains(peer));
-                self.tracked_peers = tracked_peers;
+                let all_peers = update.all.union();
+                self.update_latest_primary_peers(update.latest.primary);
+                self.tracked_peers = all_peers;
+            },
+            Some((peer, shard)) = receiver.recv() else {
```
patrick-ogrady: mailbox should be higher?

consensus/src/marshal/coding/shards/engine.rs
```diff
@@ -671,13 +690,30 @@ where
...
         queue.push_back(shard);
     }
 
+    fn update_latest_primary_peers(&mut self, peers: Set<P>) {
+        self.peer_buffers
+            .retain(|peer, _| peers.position(peer).is_some());
+        self.latest_primary_peers = peers;
+    }
+
+    fn should_buffer_peer(&self, peer: &P) -> bool {
```
patrick-ogrady: we don't need to wrap this

broadcast/fuzz/fuzz_targets/broadcast_engine_operations.rs
```diff
@@ -172,26 +173,30 @@ fn resolve_recipients(pattern: &RecipientPattern, peers: &[PublicKey]) -> Recipi
...
             commonware_p2p::simulated::Config {
                 max_size: 1024 * 1024,
                 disconnect_on_block: false,
-                tracked_peer_sets: None,
+                tracked_peer_sets: NZUsize!(1),
             },
-        );
+            peers.clone(),
+        )
+        .await;
```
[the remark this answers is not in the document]
patrick-ogrady: This is no longer necessary!

consensus/src/marshal/coding/shards/engine.rs
```diff
@@ -783,9 +811,9 @@ where
                 .await;
         }
 
-        // Send the leader's shard to tracked peers outside the participant set.
+        // Send the leader's shard to peers in aggregate membership who are not participants.
         let non_participants: Vec<P> = self
-            .tracked_peers
+            .aggregate_peers
```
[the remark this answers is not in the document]
patrick-ogrady: This is exactly as intended 👍

broadcast/src/buffered/engine.rs
```diff
@@ -182,6 +193,13 @@ where
                     self.handle_get(digest, responder);
                 }
             },
+            Some(update) = peer_set_subscription.recv() else {
```
patrick-ogrady: We should move this above mailbox handling

p2p/src/simulated/network.rs
```diff
@@ -94,12 +108,11 @@ pub struct Config {
     /// allowing byzantine actors the ability to continue sending messages.
     pub disconnect_on_block: bool,
 
-    /// The maximum number of peer sets to track. When a new peer set is registered and this
-    /// limit is exceeded, the oldest peer set is removed. Peers that are no longer in any
-    /// tracked peer set will have their links removed and messages to them will be dropped.
-    ///
-    /// If [None], peer sets are not considered.
-    pub tracked_peer_sets: Option<usize>,
+    /// The maximum number of peer sets to track (`tracked_peer_sets`). When a new peer set is
```
clabby, line 111: This comment looks out of date - links are retained when a peer set is removed now?

p2p/src/authenticated/lookup/actors/tracker/directory.rs
```diff
@@ -143,39 +150,68 @@ impl<E: Spawner + Rng + Clock + RuntimeMetrics, C: PublicKey> Directory<E, C> {
...
         }
 
-        // Create and store new peer set (all peers are tracked regardless of address validity)
-        let mut changed_peers = Vec::new();
-        for (peer, addr) in &peers {
-            let record = match self.peers.entry(peer.clone()) {
+        // Create and store new primary peer set (all peers are registered regardless of address
+        // validity).
+        let mut reset_peers = Vec::new();
+        let primary_keys = primaries.keys().clone();
```
clabby: I think we can avoid this clone by testing overlap against `primaries.keys()` and deriving the filtered secondary set in one pass.

broadcast/src/buffered/engine.rs
clabby, line 1: **Codex:**

Because the mailbox arm is still biased ahead of `peer_set_subscription.recv()`, a broadcast that is already queued when the engine starts is handled before the initial peer-set update populates `latest_primary_peers`. `handle_broadcast()` then calls `insert_message()` while that set is still empty, so a sender that is already in the initial `latest.primary` loses its own first message from the local cache. Either apply the initial peer-set update before entering the mailbox loop, or move the peer-set arm ahead of mailbox handling so startup broadcasts respect the existing primary set.

```patch
diff --git a/broadcast/src/buffered/mod.rs b/broadcast/src/buffered/mod.rs
index c7d99e3cf..64b157fba 100644
--- a/broadcast/src/buffered/mod.rs
+++ b/broadcast/src/buffered/mod.rs
@@ -1203,6 +1203,43 @@ mod tests {
         });
     }
 
+    #[test_traced]
+    fn test_broadcast_queued_before_start_respects_initial_latest_primary() {
+        let runner = deterministic::Runner::timed(Duration::from_secs(5));
+        runner.start(|context| async move {
+            let (peers, mut registrations, oracle) =
+                initialize_simulation(context.clone(), 1, 1.0).await;
+            let peer = peers[0].clone();
+            let network = registrations.remove(&peer).unwrap();
+
+            let config = Config {
+                public_key: peer,
+                mailbox_size: 1024,
+                deque_size: CACHE_SIZE,
+                priority: false,
+                codec_config: RangeCfg::from(..),
+                peer_provider: oracle.manager(),
+            };
+            let (engine, mailbox) =
+                Engine::<_, PublicKey, TestMessage, _>::new(context.with_label("peer"), config);
+
+            let msg = TestMessage::shared(b"queued-before-start");
+            let result = mailbox.broadcast(Recipients::All, msg.clone()).await;
+
+            engine.start(network);
+
+            assert!(
+                result.await.unwrap().is_empty(),
+                "single-peer broadcast should have no recipients"
+            );
+            assert_eq!(
+                mailbox.get(msg.digest()).await,
+                Some(msg),
+                "sender is already in the initial latest.primary set, so its local broadcast should be cached"
+            );
+        });
+    }
+
     #[test_traced]
     fn test_engine_starts_before_initial_peer_set_and_delivers_after_tracking() {
         let runner = deterministic::Runner::timed(Duration::from_secs(5));
```
</details>

patrick-ogrady: I just repaired this in the most recent commit 👍

p2p/src/authenticated/discovery/actors/tracker/record.rs
```diff
@@ -57,8 +57,11 @@ pub struct Record<C: PublicKey> {
     /// Connection status of the peer.
     status: Status,
 
-    /// Number of peer sets this peer is part of.
-    sets: usize,
+    /// Number of primary peer sets this peer is part of.
+    primaries: usize,
+
+    /// Number of secondary peer sets this peer is part of.
+    secondaries: usize,
```
BrendanChou: I think the naming is slightly inconsistent with `lookup`, which I think calls them `primary_sets` and `secondary_sets`?

patrick-ogrady: Yeah, that's my bad. Fixing now.

p2p/src/lib.rs
```diff
@@ -217,64 +217,166 @@ stability_scope!(BETA {
...
+    impl<P: PublicKey> From<Set<P>> for TrackedPeers<P> {
+        fn from(primary: Set<P>) -> Self {
+            Self::primary(primary)
+        }
+    }
+
+    /// Primary and secondary peers registered together for [`AddressableManager::track`].
+    ///
+    /// If the same public key appears in both maps, the primary entry is authoritative
+    /// for address and role (the key is not registered as secondary).
```
BrendanChou: How is this enforced, or who is supposed to enforce it? It doesn't seem to be enforced here

p2p/src/lib.rs
```diff
@@ -217,64 +217,166 @@ stability_scope!(BETA {
...
+    }
 
     /// Interface for reading peer set information.
     pub trait Provider: Debug + Clone + Send + 'static {
         /// Public key type used to identify peers.
         type PublicKey: PublicKey;
 
-        /// Fetch the ordered set of peers for a given ID.
+        /// Fetch the ordered set of primary peers for a given ID.
         fn peer_set(
```
BrendanChou: function should be renamed to `primary_peers()` ?

p2p/src/lib.rs
patrick-ogrady: It is not "blocked", we just handle that during registration (this is purposeful based on how we expect people to use it in things like reshare where you may have overlap between primary/secondary).

patrick-ogrady: I figured this was preferred to making applications reason about the primary/secondary splitting.

p2p/src/authenticated/discovery/actors/tracker/actor.rs
```diff
@@ -150,34 +151,44 @@ impl<E: Spawner + Rng + Clock + RuntimeMetrics, C: Signer> Actor<E, C> {
...
-                    subscriber.send_lossy((index, peers.clone(), self.directory.tracked()))
-                });
+                let update = self
+                    .directory
+                    .latest_update()
+                    .expect("latest update missing after successful track");
+                self.subscribers
+                    .retain(|subscriber| subscriber.send_lossy(update.clone()));
             }
             Message::PeerSet { index, responder } => {
```
BrendanChou, line 179: perhaps should be renamed to `PrimaryPeers`

patrick-ogrady: Tempted to actually go back to calling this PeerSet and return the primary/secondary (to match what we do in subscribe).

p2p/src/authenticated/discovery/mod.rs
```diff
@@ -70,7 +76,7 @@
 //! then sends a `BitVec` for the relevant peer set(s) (initially only knowing its own address,
 //! marked as '1'). It then waits for responses, learning about other peers through the
 //! `Payload::Peers` messages received. Bootstrapper information is persisted, and connections to
-//! them are maintained even if they aren't part of any currently tracked peer sets. Different
+//! them are maintained even if they aren't part of any retained peer sets. Different
```
BrendanChou: nit: noticing some usage of "track/tracked" "retain/retained" and "register/registered". We should perhaps try to use one of these terms everywhere

broadcast/src/buffered/mod.rs
```diff
@@ -7,17 +7,19 @@
...
-//! other peers in the network in a best-effort manner. It also has a method to request a message by
-//! digest. The engine will return the message immediately if it is in the cache, or wait for it to
-//! be received over the network if it is not.
+//! Messages referenced by multiple senders stay cached until the last per-sender deque that
+//! contains them is evicted (meaning redundant messages are only stored once).
+//!
+//! # Peer Management
+//!
+//! Only peers in `latest.primary` may buffer messages (see [commonware_p2p::Provider]). When a peer
+//! is no longer in `latest.primary`, all its buffered messages are evicted.
```
BrendanChou: nit: "...unless buffered by any other primary peers."

p2p/src/simulated/network.rs
```diff
@@ -84,6 +85,19 @@ impl<P: PublicKey, F> SplitRouter<P> for F where
 {
 }
 
+/// Primary and secondary sets registered under a single peer-set index.
+struct PeerSetsAtIndex<P: PublicKey> {
+    primary: Set<P>,
+    secondary: Set<P>,
+}
```
BrendanChou: nit: doesn't this type exist in a different file? Hard to do a find-all in github, but if so, then maybe they can be combined

p2p/src/authenticated/discovery/actors/tracker/directory.rs
```diff
@@ -8,21 +8,28 @@ use crate::{
...
+    num::NonZeroUsize,
     ops::Deref,
     time::{Duration, SystemTime},
 };
 use tracing::{debug, warn};
 
+/// Primary and secondary sets registered under a single peer-set index.
+struct PeerSetsAtIndex<C: PublicKey> {
+    primary: Set<C>,
+    secondary: OrderedSet<C>,
```
BrendanChou: seems strange these are different types?

p2p/src/authenticated/discovery/actors/tracker/directory.rs
```diff
@@ -199,64 +209,118 @@ impl<E: Spawner + Rng + Clock + RuntimeMetrics, C: PublicKey> Directory<E, C> {
...
+        self.peer_sets.get(index).map(|e| e.primary.deref())
+    }
+
+    /// Gets a secondary peer set by index.
+    pub fn get_secondary_set(&self, index: &u64) -> Option<&OrderedSet<C>> {
+        self.peer_sets.get(index).map(|e| &e.secondary)
     }
 
-    /// Returns the latest peer set index.
+    /// Returns the latest primary peer set index.
```
BrendanChou: nit: addition of `primary` here not necessary I think

p2p/src/authenticated/discovery/actors/tracker/directory.rs
```diff
@@ -199,64 +209,118 @@ impl<E: Spawner + Rng + Clock + RuntimeMetrics, C: PublicKey> Directory<E, C> {
...
+    }
+
+    /// Returns a [`PeerSetUpdate`] for the latest peer set (by id), if any.
+    pub fn latest_update(&self) -> Option<PeerSetUpdate<C>> {
+        let index = self.latest_set_index()?;
+        Some(PeerSetUpdate {
+            index,
+            latest: TrackedPeers::new(
+                self.get_primary_set(&index).cloned().unwrap(),
+                self.get_secondary_set(&index).cloned().unwrap_or_default(),
```
BrendanChou: nit: should be `unwrap()` I think (like the above line). If `primary_set()` unwraps then this should too

p2p/src/authenticated/lookup/actors/tracker/actor.rs
```diff
@@ -135,23 +134,17 @@ impl<E: Spawner + Rng + Clock + RuntimeMetrics, C: Signer> Actor<E, C> {
...
         match msg {
             Message::Register { index, peers } => {
-                // Identify peers that were added or had their addresses changed.
-                let peer_keys: Set<C::PublicKey> = peers.keys().clone();
-                let Some((deleted, changed)) = self.directory.add_set(index, peers) else {
+                let primary = peers.primary;
+                let secondary = peers.secondary;
+
+                // Identify peers whose existing connection state should be reset.
+                let Some(reset_peers) = self.directory.track(index, primary, secondary) else {
```
BrendanChou: nit: seems to me that if the following type is defined somewhere in like `p2p/lib.rs`
```
struct PeerSet<P: PublicKey> {
  primary: OrderedSet<P>,
  secondary: OrderedSet<P>
}
Then `Directory::track()` should take in `(index, peer_set)` for consistency. This can reduce code churn (preventing modifying its arguments) if we ever added more fields to a peer set.

File: p2p/src/authenticated/lookup/actors/tracker/actor.rs [hunk 30]
```diff
@@ -193,16 +189,15 @@ impl<E: Spawner + Rng + Clock + RuntimeMetrics, C: Signer> Actor<E, C> {
             }
             Message::PeerSet { index, responder } => {
```
BrendanChou, line 187: I think I made this comment in `discovery`, but this should probably be renamed to `PrimaryPeers` OR it should return the full `PeerSet` struct (primaries and secondaries). Maybe the latter is better

p2p/src/authenticated/lookup/actors/tracker/directory.rs
```diff
@@ -5,24 +5,31 @@ use crate::{
...
 use std::{
     collections::{hash_map::Entry, BTreeMap, HashMap, HashSet},
     net::IpAddr,
+    num::NonZeroUsize,
     time::{Duration, SystemTime},
 };
 use tracing::{debug, warn};
 
+/// Primary and secondary sets registered under a single peer-set index.
+struct PeerSetsAtIndex<C: PublicKey> {
```
BrendanChou: again, this type should be combined into a single `PeerSet` type in `lib.rs` IMO

p2p/src/authenticated/lookup/actors/tracker/directory.rs
```diff
@@ -143,39 +150,66 @@ impl<E: Spawner + Rng + Clock + RuntimeMetrics, C: PublicKey> Directory<E, C> {
...
+            };
+            record.increment_primary();
+        }
+
+        // Create and store new secondary peer set.
+        for (secondary, addr) in &secondaries {
+            // When a peer appears in both roles for the same index, keep only the primary role
+            // (and the primary registration above); do not count them as secondary.
+            if primaries.position(secondary).is_some() {
+                continue;
```
BrendanChou, line 192: The logic in this function is deceptively complicated and hard to follow. Specifically the logic around not allowing primaries to be secondaries.

For example, I think we are unable to directly "demote" a peer from primary to secondary which causes a bug. This is because we:

- Add new primaries
- Add new secondaries (only if not primary)
- Remove old primaries and secondaries

The "only if not primary" part means we skip adding secondaries for primaries that are removed in the last step.

While this logic can be fixed, it seems to be complicated and is a footgun waiting to happen upon a later code change. Consider relaxing the constraint that a secondary cannot be a primary (?). While this may be inefficient in certain cases, it may also just be less brittle in general. Right now it is also not entirely clear who is the one that should be preventing this duplication

p2p/src/authenticated/lookup/actors/tracker/directory.rs
```diff
@@ -220,21 +277,45 @@ impl<E: Spawner + Rng + Clock + RuntimeMetrics, C: PublicKey> Directory<E, C> {
...
+                self.get_secondary_set(&index).cloned().unwrap_or_default(),
+            ),
+            all: self.all(),
+        })
+    }
+
+    /// Gets a secondary peer set by index.
+    pub fn get_secondary_set(&self, index: &u64) -> Option<&Set<C>> {
+        self.peer_sets.get(index).map(|e| &e.secondary)
     }
```
BrendanChou: nit: consider keeping the `get_set` function which returns `PeerSet{primary, secondary}` rather than two different functions to get-by-index

p2p/src/authenticated/lookup/actors/tracker/directory.rs
```diff
@@ -220,21 +277,45 @@ impl<E: Spawner + Rng + Clock + RuntimeMetrics, C: PublicKey> Directory<E, C> {
...
     /// Returns `Some` on success, `None` otherwise.
     pub fn dial(&mut self, peer: &C) -> Option<(Reservation<C>, Ingress)> {
-        let ingress = self.peers.get(peer)?.ingress()?;
+        let ingress = {
+            let record = self.peers.get(peer)?;
+            if !record.is_outbound_target() {
+                return None;
+            }
+            record.ingress()?
+        };
```
BrendanChou:
```suggestion
        let record = self.peers.get(peer)?;
        let ingress = record.is_outbound_target().then_some(record.ingress()?);
```

p2p/src/authenticated/discovery/actors/tracker/directory.rs
patrick-ogrady: This is done because `Set` is actually specific to discovery (and includes a bitmap). This predates our broader renaming to Set...I'll change it to BitSet.

p2p/src/authenticated/lookup/actors/tracker/directory.rs
patrick-ogrady: I opted to keep this pattern but added more test coverage to ensure demotion works properly.

PR #3501, opened 2026-03-30
[runtime/network] poison sink/stream on error and cancellation

### LINKED ISSUE #2596: [runtime/network] Allow read timeout recovery
## Overview

Currently, if a read times out in the `tokio` `Stream` impl, the partially read data will be lost due to `read_exact` not being cancel safe. We currently disconnect upstream of this code if a timeout occurs, but if we want to support recovery, we'll need to consider holding read data in a buffer (or in an error variant) for recovery purposes.

https://github.com/commonwarexyz/monorepo/blob/2fbf479cf37fb3f2a0f68a3a9e0583c0dbcd05c3/runtime/src/network/tokio.rs#L37-L52

andresilva:
This PR changes the network `Sink`/`Stream` contract so each half is poisoned after cancellation or its first I/O error instead of being implicitly reusable in an incosistent state. A failed `send` now makes the `Sink` unusable, and a failed `recv` makes the `Stream` unusable, subsequent operations on that same half return `Closed`, while the opposite direction can still continue.

Related https://github.com/commonwarexyz/monorepo/issues/2596 (possibly closes it?)


andresilva (commented):
mock changes lgtm

runtime/src/mocks.rs
patrick-ogrady, line 1: Ran this through 1 more round of review and we may need to put in a bit more elbow grease for consistency:

```
diff --git a/runtime/src/mocks.rs b/runtime/src/mocks.rs
index 959172edf..a9c76d9ea 100644
--- a/runtime/src/mocks.rs
+++ b/runtime/src/mocks.rs
@@ -16,11 +16,14 @@ pub struct Channel {
     /// If the stream is waiting to read bytes, the waiter stores the number of
     /// bytes that the stream is waiting for, as well as the oneshot sender that
     /// the sink uses to send the bytes to the stream directly.
-    waiter: Option<(usize, oneshot::Sender<Bytes>)>,
+    waiter: Option<Waiter>,
 
     /// Target size for the stream's local buffer, used to bound buffering.
     read_buffer_size: usize,
 
+    /// Monotonic identifier for waiter registrations.
+    next_waiter_id: u64,
+
     /// Tracks whether the sink is still alive and able to send messages.
     sink_alive: bool,
 
@@ -28,6 +31,12 @@ pub struct Channel {
     stream_alive: bool,
 }
 
+struct Waiter {
+    id: u64,
+    requested: usize,
+    sender: oneshot::Sender<Bytes>,
+}
+
 impl Channel {
     /// Returns an async-safe Sink/Stream pair with default read buffer size.
     pub fn init() -> (Sink, Stream) {
@@ -40,6 +49,7 @@ impl Channel {
             buffer: BytesMut::new(),
             waiter: None,
             read_buffer_size,
+            next_waiter_id: 0,
             sink_alive: true,
             stream_alive: true,
         }));
@@ -55,6 +65,63 @@ impl Channel {
             },
         )
     }
+
+    /// Restores bytes that were detached from the front of the shared buffer.
+    fn restore_front(&mut self, data: Bytes) {
+        if data.is_empty() {
+            return;
+        }
+
+        let mut restored = BytesMut::with_capacity(data.len() + self.buffer.len());
+        restored.extend_from_slice(&data);
+        restored.extend_from_slice(&self.buffer);
+        self.buffer = restored;
+    }
+
+    /// Marks the sink as closed and wakes any waiter.
+    fn close_sink(&mut self) {
+        self.sink_alive = false;
+
+        // If there is a waiter, resolve it by dropping the oneshot sender.
+        self.waiter.take();
+    }
+}
+
+struct RecvWaiterGuard {
+    channel: Arc<Mutex<Channel>>,
+    waiter_id: u64,
+    active: bool,
+}
+
+impl RecvWaiterGuard {
+    fn new(channel: Arc<Mutex<Channel>>, waiter_id: u64) -> Self {
+        Self {
+            channel,
+            waiter_id,
+            active: true,
+        }
+    }
+
+    fn disarm(&mut self) {
+        self.active = false;
+    }
+}
+
+impl Drop for RecvWaiterGuard {
+    fn drop(&mut self) {
+        if !self.active {
+            return;
+        }
+
+        let mut channel = self.channel.lock();
+        if channel
+            .waiter
+            .as_ref()
+            .is_some_and(|waiter| waiter.id == self.waiter_id)
+        {
+            channel.waiter.take();
+        }
+    }
 }
 
 /// A mock sink that implements the Sink trait.
@@ -86,37 +153,45 @@ impl SinkTrait for Sink {
             if channel
                 .waiter
                 .as_ref()
-                .is_some_and(|(requested, _)| *requested <= channel.buffer.len())
+                .is_some_and(|waiter| waiter.requested <= channel.buffer.len())
             {
                 // Send up to read_buffer_size bytes (but at least requested amount)
-                let (requested, os_send) = channel.waiter.take().unwrap();
+                let waiter = channel.waiter.take().unwrap();
                 let send_amount = channel
                     .buffer
                     .len()
-                    .min(requested.max(channel.read_buffer_size));
+                    .min(waiter.requested.max(channel.read_buffer_size));
                 let data = channel.buffer.split_to(send_amount).freeze();
-                (os_send, data)
+                (waiter.sender, data)
             } else {
                 return Ok(());
             }
         };
 
         // Resolve the waiter.
-        os_send.send(data).map_err(|_| {
-            self.poisoned = true;
-            Error::SendFailed
-        })?;
-        Ok(())
+        match os_send.send(data) {
+            Ok(()) => Ok(()),
+            Err(data) => {
+                let mut channel = self.channel.lock();
+                channel.restore_front(data);
+
+                // A canceled recv should behave like a buffered transport:
+                // preserve the bytes and allow a subsequent recv to consume them.
+                if channel.stream_alive {
+                    Ok(())
+                } else {
+                    self.poisoned = true;
+                    Err(Error::Closed)
+                }
+            }
+        }
     }
 }
 
 impl Drop for Sink {
     fn drop(&mut self) {
         let mut channel = self.channel.lock();
-        channel.sink_alive = false;
-
-        // If there is a waiter, resolve it by dropping the oneshot sender.
-        channel.waiter.take();
+        channel.close_sink();
     }
 }
 
@@ -134,7 +209,7 @@ impl StreamTrait for Stream {
             return Err(Error::Closed);
         }
 
-        let os_recv = {
+        let (waiter_id, os_recv) = {
             let mut channel = self.channel.lock();
 
             // Pull data from channel buffer into local buffer.
@@ -164,16 +239,34 @@ impl StreamTrait for Stream {
             // Set up waiter for remaining amount.
             let remaining = len - self.buffer.len();
             assert!(channel.waiter.is_none());
+            let waiter_id = channel.next_waiter_id;
+            channel.next_waiter_id = channel
+                .next_waiter_id
+                .checked_add(1)
+                .expect("waiter id overflow");
             let (os_send, os_recv) = oneshot::channel();
-            channel.waiter = Some((remaining, os_send));
-            os_recv
+            channel.waiter = Some(Waiter {
+                id: waiter_id,
+                requested: remaining,
+                sender: os_send,
+            });
+            (waiter_id, os_recv)
         };
 
+        let mut waiter_guard = RecvWaiterGuard::new(self.channel.clone(), waiter_id);
+
         // Wait for the waiter to be resolved.
-        let data = os_recv.await.map_err(|_| {
-            self.poisoned = true;
-            Error::Closed
-        })?;
+        let data = match os_recv.await {
+            Ok(data) => {
+                waiter_guard.disarm();
+                data
+            }
+            Err(_) => {
+                waiter_guard.disarm();
+                self.poisoned = true;
+                return Err(Error::Closed);
+            }
+        };
         self.buffer.extend_from_slice(&data);
 
         assert!(self.buffer.len() >= len);
@@ -329,7 +422,7 @@ mod tests {
     }
 
     #[test]
-    fn test_send_error_canceled_recv_poisoned() {
+    fn test_send_canceled_recv_does_not_poison_sink() {
         let (mut sink, mut stream) = Channel::init();
 
@@ -342,13 +435,11 @@ mod tests {
                 _ = context.sleep(Duration::from_millis(50)) => {},
             };
 
-            // The first send hits the canceled waiter and fails.
-            let result = sink.send(b"hello".as_slice()).await;
-            assert!(matches!(result, Err(Error::SendFailed)));
+            sink.send(b"hello".as_slice()).await.unwrap();
+            sink.send(b" world".as_slice()).await.unwrap();
 
-            // After any send error, the mock sink must remain closed.
-            let result = sink.send(b"world".as_slice()).await;
-            assert!(matches!(result, Err(Error::Closed)));
+            let received = stream.recv(11).await.unwrap();
+            assert_eq!(received.coalesce(), b"hello world");
         });
     }
 
diff --git a/runtime/src/network/mod.rs b/runtime/src/network/mod.rs
index bd13412d6..514422304 100644
--- a/runtime/src/network/mod.rs
+++ b/runtime/src/network/mod.rs
@@ -36,6 +36,7 @@ mod tests {
         test_network_large_data(new_network()).await;
         test_network_connection_errors(new_network()).await;
         test_network_peek(new_network()).await;
+        test_network_canceled_recv_does_not_poison_peer_sink(new_network()).await;
         test_network_recv_error_poisons_stream(new_network()).await;
         test_network_send_error_poisons_sink(new_network()).await;
     }
@@ -367,6 +368,70 @@ mod tests {
         client_result.expect("Client task failed");
     }
 
+    async fn test_network_canceled_recv_does_not_poison_peer_sink<N: crate::Network>(network: N) {
+        let mut listener = network
+            .bind(SocketAddr::from(([127, 0, 0, 1], 0)))
+            .await
+            .expect("Failed to bind");
+        let listener_addr = listener.local_addr().expect("Failed to get local address");
+        let (canceled_sender, canceled_receiver) = oneshot::channel();
+
+        // Server cancels one recv attempt, then reads the same bytes on the
+        // next recv and replies successfully.
+        let server = tokio::spawn(async move {
+            let (_, mut sink, mut stream) = listener.accept().await.expect("Failed to accept");
+
+            tokio::select! {
+                v = stream.recv(5) => {
+                    panic!("unexpected value: {v:?}");
+                },
+                _ = tokio::time::sleep(Duration::from_millis(5)) => {},
+            };
+
+            canceled_sender
+                .send(())
+                .expect("client should still be waiting for cancellation");
+
+            let received = stream
+                .recv(5)
+                .await
+                .expect("Failed to receive after cancellation");
+            assert_eq!(received.coalesce(), b"hello");
+
+            sink.send(IoBuf::from(b"ok"))
+                .await
+                .expect("sink should remain usable after peer cancels recv");
+
+            (sink, stream)
+        });
+
+        // Client sends after the server canceled its pending recv, proving the
+        // peer write half remains usable and the data is still delivered.
+        let client = tokio::spawn(async move {
+            let (mut sink, mut stream) = network
+                .dial(listener_addr)
+                .await
+                .expect("Failed to dial server");
+
+            canceled_receiver
+                .await
+                .expect("server should signal after canceling recv");
+
+            sink.send(IoBuf::from(b"hello"))
+                .await
+                .expect("send should remain usable after peer cancels recv");
+
+            let received = stream.recv(2).await.expect("Failed to receive response");
+            assert_eq!(received.coalesce(), b"ok");
+
+            (sink, stream)
+        });
+
+        let (server_result, client_result) = join!(server, client);
+        server_result.expect("Server task failed");
+        client_result.expect("Client task failed");
+    }
+
     async fn test_network_recv_error_poisons_stream<N: crate::Network>(network: N) {
         let mut listener = network
             .bind(SocketAddr::from(([127, 0, 0, 1], 0)))
```

patrick-ogrady: What is the invariant we want here exactly? a failed send forces recv to fail? a failed recv forces send to fail? both?

andresilva: No, currently they are independent, a failed send poisons only its sink and likewise for recv. Failing both sides would require extra synchronization and IMO it's unnecessary, users can decide if they want to fail the entire connection if either side fails.

runtime/src/network/mod.rs
```diff
@@ -365,6 +370,247 @@ mod tests {
...
+        server_result.expect("Server task failed");
+        client_result.expect("Client task failed");
+    }
+
+    async fn test_network_canceled_send_poisons_sink<N: crate::Network>(network: N) {
+        // Windows IOCP completes TCP writes without yielding, so send
+        // cancellation cannot be easily triggered on that platform.
+        if cfg!(target_os = "windows") {
+            return;
+        }
```
clabby, line 425: :sad_cowboy:

runtime/src/lib.rs
```diff
@@ -589,6 +589,11 @@ stability_scope!(BETA {
         /// # Warning
         ///
         /// If the sink returns an error, part of the message may still be delivered.
+        /// After any error, the sink is no longer reusable and subsequent sends will
+        /// return [Error::Closed].
```
clabby: nit: [\`Error::Closed\`]

runtime/src/lib.rs
```diff
@@ -605,6 +610,11 @@ stability_scope!(BETA {
         /// # Warning
         ///
         /// If the stream returns an error, partially read data may be discarded.
+        /// After any error, the stream is no longer reusable and subsequent receives
+        /// will return [Error::Closed].
```
clabby: nit: [\`Error::Closed\`]

runtime/src/mocks.rs
```diff
@@ -114,23 +196,33 @@ pub struct Stream {
...
+                .len()
+                .min(target.saturating_sub(self.buffer.len()));
+            if pull_amount > 0 {
+                let data = channel.buffer.split_to(pull_amount);
+                self.buffer.extend_from_slice(&data);
+
+                // Wake a blocked sender if the buffer drained below the limit.
+                if channel.buffer.len() <= channel.buffer_size {
+                    if let Some(sender) = channel.drain_waiter.take() {
+                        let _ = sender.send(());
```
clabby: nit: Use `send_lossy`

runtime/src/network/tokio.rs
```diff
@@ -34,6 +35,14 @@ impl Sink {
 
 impl crate::Sink for Sink {
     async fn send(&mut self, bufs: impl Into<IoBufs> + Send) -> Result<(), Error> {
+        if self.poisoned {
+            return Err(Error::Closed);
```
clabby: A canceled `send` sets `poisoned = true`, but if the future is dropped before the normal error path runs, the write half is never shut down. The next `send` returns `Closed` here without closing anything, so the peer can continue waiting for bytes after the local sink has become permanently unusable. The same pattern exists in the mock and io_uring sinks. Please close/half-close when transitioning into the poisoned state on cancellation, or at least perform the shutdown/`close_sink()` work before returning `Closed` from the poisoned branch.

Regression test:
- test file path: `runtime/src/mocks.rs`
- command: `just test -p commonware-runtime test_canceled_send_closes_peer_after_buffer_drained`
- observed result: fail, panics because the stream never observes the poisoned sink closing after buffered bytes are drained.

~~~rust
#[test]
fn test_canceled_send_closes_peer_after_buffer_drained() {
    let (mut sink, mut stream) = Channel::init_with_buffer_size(1);

    let executor = deterministic::Runner::timed(Duration::from_secs(10));
        select! {
            result = sink.send(b"abc".as_slice()) => {
                panic!("send should block before cancellation: {result:?}");
            },
            _ = context.sleep(Duration::from_millis(1)) => {},
        };

        let result = sink.send(b"x".as_slice()).await;
        assert!(matches!(result, Err(Error::Closed)));

        let received = stream.recv(3).await.unwrap();
        assert_eq!(received.coalesce(), b"abc");

        select! {
            result = stream.recv(1) => {
                assert!(matches!(result, Err(Error::RecvFailed)));
            },
            _ = context.sleep(Duration::from_millis(1)) => {
                panic!("stream did not observe the poisoned sink closing");
            },
        };
    });
}
~~~

runtime/src/network/mod.rs
```diff
@@ -365,6 +370,247 @@ mod tests {
...
+        let listener_addr = listener.local_addr().expect("Failed to get local address");
+
+        let server = tokio::spawn(async move {
+            let (_, mut sink, mut stream) = listener.accept().await.expect("Failed to accept");
+
+            // Cancel a large send mid-write. The payload must exceed all
+            // backend buffer sizes (TCP kernel buffer, mock backpressure
+            // threshold) so that send blocks and can be interrupted.
+            select! {
+                _ = sink.send(vec![0u8; 256 * 1024 * 1024]) => {
```
clabby: This cancellation test allocates a 256 MiB buffer and relies on a 5 ms sleep to win the race against the send completing. That makes the unit test resource-heavy and timing-sensitive across CI hosts and network backends. Please make the blocking condition deterministic instead, for example by using a backend-specific small buffer/backpressure setup for the mock path and a targeted socket-buffer strategy for TCP backends, rather than a very large payload plus wall-clock timing.

clabby (commented):
AI review

runtime/src/network/tokio.rs
andresilva: Good one. Updated.

runtime/src/network/mod.rs
andresilva: Addressed in a different manner.

clabby (approved):

PR #3502, opened 2026-03-30
[cryptography] BLS12381: Allow reading zero scalars

cronokirby:
For backwards compatibility, we introduce a config, allowing to choose whether or not to reject zero scalars.

Private keys reject zero, like before.

### DIFF
--- a/cryptography/fuzz/fuzz_targets/bls12381_primitive_operations.rs
+++ b/cryptography/fuzz/fuzz_targets/bls12381_primitive_operations.rs
@@ -3,7 +3,7 @@
 use arbitrary::{Arbitrary, Unstructured};
 use commonware_codec::{Read, ReadExt, Write};
 use commonware_cryptography::bls12381::primitives::{
-    group::{Private, Scalar, Share, G1, G1_MESSAGE, G2, G2_MESSAGE},
+    group::{Private, Scalar, ScalarReadCfg, Share, G1, G1_MESSAGE, G2, G2_MESSAGE},
     ops,
     variant::{MinPk, MinSig, Variant},
 };
@@ -162,6 +162,7 @@ enum FuzzOperation {
     // Serialization round-trip
     SerializeScalar {
         scalar: Scalar,
+        cfg: ScalarReadCfg,
     },
     SerializeG1 {
         point: G1,
@@ -302,6 +303,7 @@ impl<'a> Arbitrary<'a> for FuzzOperation {
             }),
             30 => Ok(FuzzOperation::SerializeScalar {
                 scalar: u.arbitrary()?,
+                cfg: u.arbitrary()?,
             }),
             31 => Ok(FuzzOperation::SerializeG1 {
                 point: arbitrary_g1(u)?,
@@ -601,12 +603,16 @@ fn fuzz(op: FuzzOperation) {
             }
         }
 
-        FuzzOperation::SerializeScalar { scalar } => {
+        FuzzOperation::SerializeScalar { scalar, cfg } => {
             let mut encoded = Vec::new();
             scalar.write(&mut encoded);
-            if let Ok(decoded) = Scalar::read_cfg(&mut encoded.as_slice(), &Default::default()) {
-                assert_eq!(scalar, decoded);
+            if cfg == ScalarReadCfg::RejectZero && scalar == Scalar::zero() {
+                assert!(Scalar::read_cfg(&mut encoded.as_slice(), &cfg).is_err());
+                return;
             }
+
+            let decoded = Scalar::read_cfg(&mut encoded.as_slice(), &cfg).unwrap();
+            assert_eq!(scalar, decoded);
         }
 
         FuzzOperation::SerializeG1 { point } => {
--- a/cryptography/fuzz/fuzz_targets/bls12381_primitive_operations.rs
+++ b/cryptography/fuzz/fuzz_targets/bls12381_primitive_operations.rs
@@ -1,7 +1,7 @@
 #![no_main]
 
 use arbitrary::{Arbitrary, Unstructured};
-use commonware_codec::{ReadExt, Write};
+use commonware_codec::{Read, ReadExt, Write};
 use commonware_cryptography::bls12381::primitives::{
     group::{Private, Scalar, Share, G1, G1_MESSAGE, G2, G2_MESSAGE},
     ops,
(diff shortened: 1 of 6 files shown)


cryptography/src/bls12381/primitives/group.rs
```diff
@@ -624,22 +624,31 @@ impl Write for Scalar {
 }
 
 impl Read for Scalar {
-    type Cfg = ();
+    /// When `true`, reject zero scalars (use for private keys per IETF BLS spec).
+    /// When `false`, allow zero scalars (only validate in-range).
```
patrick-ogrady: I think this should be inverted?

cronokirby: Perhaps even an enum, if there's two plausible patterns here.

cryptography/src/bls12381/dkg.rs
```diff
@@ -825,7 +825,7 @@ impl Read for DealerPrivMsg {
         buf: &mut impl bytes::Buf,
         _cfg: &Self::Cfg,
     ) -> Result<Self, commonware_codec::Error> {
-        Ok(Self::new(ReadExt::read(buf)?))
```
patrick-ogrady, line 828: This needs to be reject zero (not default) otherwise it is a breaking change?

cryptography/fuzz/fuzz_targets/bls12381_primitive_operations.rs
```diff
@@ -604,7 +604,7 @@ fn fuzz(op: FuzzOperation) {
         FuzzOperation::SerializeScalar { scalar } => {
             let mut encoded = Vec::new();
             scalar.write(&mut encoded);
-            if let Ok(decoded) = Scalar::read(&mut encoded.as_slice()) {
+            if let Ok(decoded) = Scalar::read_cfg(&mut encoded.as_slice(), &Default::default()) {
```
patrick-ogrady: I suppose we could make scalar cfg part of the FuzzOperation enum?

cryptography/src/bls12381/primitives/group.rs
```diff
@@ -1834,10 +1850,23 @@ mod tests {
         let original = Scalar::random(&mut test_rng());
```
patrick-ogrady, line 1864: Should random be modified to include 0?

patrick-ogrady: Right now (for Scalar), we'll use `blst_keygen`: https://github.com/commonwarexyz/monorepo/blob/d1a4e69eaa10647cbf0b6eb0c548283bd36af39e/cryptography/src/bls12381/primitives/group.rs#L825-L830

We could have an option that only does that for `Private`?

cronokirby: In practice, generating a uniformly random scalar will never return 0, there's simply too many points. So I think the behavior of this function is effectively the same.

patrick-ogrady: Only thing I was thinking about was fuzzing (where we provide a FuzzRng with bytes?)

patrick-ogrady: But in practice, I agree.

cryptography/src/bls12381/primitives/group.rs
```diff
@@ -212,6 +212,17 @@ where
 /// Reference: <https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-bls-signature-05#name-ciphersuites>
 pub type DST = &'static [u8];
 
+/// Configuration for [`Scalar`]'s [`Read`] implementation.
+#[cfg_attr(any(test, feature = "arbitrary"), derive(arbitrary::Arbitrary))]
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+pub enum ScalarReadCfg {
```
patrick-ogrady, line 219: My last piece of feedback on this PR is to align this better with what the check is actually asserting...I think for `RejectZero` we should call it `SecretKey` because that is the parsing it actually does.

For the default case, we could call it `ScalarField`? Not sure.

cryptography/src/bls12381/primitives/group.rs
cronokirby: That makes sense, actually

cryptography/src/bls12381/primitives/group.rs
cronokirby: hmm, I think at this level of semantics it makes sense to describe what scalars you're rejecting, with choices about why you might want to do that at a higher level. In other words, this is dealing with scalars, a mathematical object, and how you arrange that into secure protocols is not really the concern here.

cryptography/src/bls12381/primitives/group.rs
```diff
@@ -212,6 +212,17 @@ where
...
+    /// Accept any in-range scalar, including zero.
+    #[default]
+    AllowZero,
```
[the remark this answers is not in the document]
cronokirby: Yeah, this is fine because you go from passing nothing, to passing something.

cronokirby:
False positive on the breaking-format label, because we've changed the arbitrary impl for Scalar.

cryptography/src/bls12381/primitives/group.rs
```diff
@@ -806,11 +835,11 @@ impl Field for Scalar {
...
 impl Random for Scalar {
-    /// Returns a random non-zero scalar.
+    /// Returns a random scalar, including zero.
     fn random(mut rng: impl CryptoRngCore) -> Self {
-        let mut ikm = Zeroizing::new([0u8; IKM_LENGTH]);
-        rng.fill_bytes(ikm.as_mut());
-        Self::from_ikm(&ikm)
+        let mut bytes = Zeroizing::new([0u8; SCALAR_LENGTH + 128 / 8]);
+        rng.fill_bytes(bytes.as_mut());
+        Self::reduce_bytes(bytes.as_ref())
```
patrick-ogrady, line 843: Switching to allow non-zero here (which seems reasonable for the type), means that our downstream usage in DKG can now create 0 coefficients? Do we want that (I think we'll reject them during read_cfg so it can cause an honest player to produce an "invalid" polynomial)?

patrick-ogrady: I think we could create a `NonZeroScalar` that has the same non-zero property as Private but doesn't use Zeroizing?

PR #3503, opened 2026-03-30
[Storage] Maintain weak parent pointer in batch

danlaine:
## Summary

Each `MerkleizedBatch` at every layer -- merkle, authenticated journal, and QMDB -- now stores only its own local data plus `Arc` refs to each ancestor's data, collected by walking the `Weak` parent chain during merkleize. `apply_batch` uses them to replay uncommitted ancestors. Committed ancestors are freed immediately and their `Weak` refs naturally expire. The memory leak from continuous chaining is gone.

The `Mem` struct (the in-memory Merkle structure) no longer uses `Arc` internally. It is a plain owned struct. No batch type stores a `Mem` clone. The `Journaled` layer enforces this at compile time: the only way to access `&Mem` is through `with_mem(|mem| ...)`, which borrows through a read lock. No owned clone can escape. Deep copies of the MMR are structurally impossible.

- `finalize()` and `Changeset` removed from the merkle batch public API. `mem.apply_batch(batch)` handles everything.
- `MerkleizedBatch` is now a flat struct (was an enum with `Base`/`Layer`/`Checkpoint` variants).
- The grafted MMR on `current::Db` is now a `Mem` (was a `MerkleizedBatch` chain with `flatten()`/`push_batch()`).
- `Error::StaleChangeset` renamed to `Error::StaleBatch`.
- Dropped-ancestor detection: `Mem::apply_batch` returns `Error::AncestorDropped` if ancestor data is missing.

## Usage

The public API is `new_batch` / `merkleize` / `apply_batch`. Sequential commits and stale fork detection are handled automatically:

```rust
let mut db = Db::init(context, config).await?;

// Build A, apply it, then build B from A.
let a = db.new_batch()
    .write(k1, Some(v1))
    .merkleize(&db, None).await?;
db.apply_batch(Arc::clone(&a)).await?;  // apply A, keep our Arc alive

let b = a.new_batch()
    .write(k2, Some(v2))
    .merkleize(&db, None).await?;
drop(a);  // A freed -- its Weak ref in B expires, but A's data is committed.

db.apply_batch(Arc::clone(&b)).await?;  // detects A already committed, applies only B

// Fork two batches from B, and build E from D before applying anything.
let c = b.new_batch().write(k3, Some(v3)).merkleize(&db, None).await?;
let d = b.new_batch().write(k4, Some(v4)).merkleize(&db, None).await?;
let e = d.new_batch().write(k5, Some(v5)).merkleize(&db, None).await?;

db.apply_batch(c).await?;   // applies C
db.apply_batch(d).await;    // Error::StaleBatch -- sibling C was committed
db.apply_batch(e).await;    // Error::StaleBatch -- E's ancestor D is stale too

// Or skip intermediate batches: apply B directly and A comes along.
// db.apply_batch(b).await?;  // applies A (via ancestors) + B
```

The same pattern applies at the merkle layer:

```rust
let mut mem = Mem::new(&hasher);
let a = mem.new_batch().add(&hasher, b"a").merkleize(&mem, &hasher);
let b = a.new_batch().add(&hasher, b"b").merkleize(&mem, &hasher);
mem.apply_batch(&a)?;   // applies A
mem.apply_batch(&b)?;   // skips A, applies only B
```

### Ancestor lifetime constraint

All uncommitted ancestors in a batch chain must be kept alive until their descendants are merkleized. The `Weak` parent chain is walked during merkleize to collect ancestor data. If an uncommitted ancestor has been dropped, its `Weak` ref is dead and its data is silently lost from the collected set. `Mem::apply_batch` detects this (when the dropped ancestor had appends) and returns `Error::AncestorDropped`. Committed-and-dropped ancestors are fine -- their data is already in committed storage.

## Architecture: the shared pattern

All batch layers follow the same structural pattern, with minor variations at the QMDB current layer.

### Per-layer data

Each layer's `MerkleizedBatch` stores its own local data, not a flattened copy of all ancestors:

| Layer | Local data | Ancestors |
|-------|-----------|-----------|
| **Merkle** (`batch.rs`) | `appended`, `overwrites` | `ancestor_appended`, `ancestor_overwrites` |
| **Auth journal** (`authenticated.rs`) | `items: Arc<Vec<Item>>` | `ancestor_items: Vec<Arc<Vec<Item>>>` |
| **QMDB any** (`any/batch.rs`) | `diff: Arc<BTreeMap<Key, DiffEntry>>` | `ancestor_diffs`, `ancestor_seg_ends` |
| **QMDB immutable** (`immutable/batch.rs`) | `diff: Arc<BTreeMap<Key, DiffEntry>>` | `ancestor_diffs`, `ancestor_seg_ends` |
| **QMDB current** (`current/batch.rs`) | `bitmap_pushes`, `bitmap_clears` | `ancestor_bitmap_pushes`, `ancestor_bitmap_clears` |

Each layer wraps the layer below it. QMDB current wraps QMDB any, which wraps the authenticated journal, which wraps the merkle batch.

### Lifecycle

Every layer follows the same lifecycle:

1. **`new_batch(self: &Arc<Self>)`** -- creates an `UnmerkleizedBatch`. At the merkle, authenticated journal, and QMDB any layers, this holds a **strong** `Arc` to its parent, keeping the parent alive during mutation and merkleization. At the QMDB current layer, a `Weak` is stored instead (the inner any-layer batch holds the strong ref that keeps the parent's inner state alive; the current-layer parent is kept alive by the caller holding it).

2. **`merkleize()`** -- consumes the `UnmerkleizedBatch`, producing an immutable `Arc<MerkleizedBatch>`:
   - Computes the root hash (merkle layer) or resolves operations (QMDB layer).
   - Walks the `Weak` parent chain to collect `Arc` refs to each ancestor's local data. The immediate parent is alive (strong Arc held during merkleize). Grandparents use `Weak` upgrades; dead refs (committed-and-dropped ancestors) truncate the walk.
   - Stores a `Weak` pointer to the parent on the resulting `MerkleizedBatch`.
   - The ancestor data is stored on the result (as `Vec<Arc<...>>`).

3. **`apply_batch()`** -- applies the batch to committed storage:
   - Detects whether ancestors have already been committed. Skips them if so.
   - Otherwise, applies ancestors in the correct order, then this batch's data.
   - At the merkle layer: verifies the resulting size matches expected. Returns `Error::AncestorDropped` if ancestor data is missing.

### Ancestor collection

Ancestors are collected at merkleize time by walking the `Weak` parent chain. The walk collects data from every ancestor that is still alive:

```
After merkleizing A, B, C (all still alive):

MerkleizedBatch A          MerkleizedBatch B              MerkleizedBatch C
  parent: None               parent: Weak<A>                parent: Weak<B>
  data: Arc<{...}>           data: Arc<{...}>               data: Arc<{...}>
                             ancestors: [Arc<A.data>]       ancestors: [Arc<A.data>,
                                                                        Arc<B.data>]
```

- `Weak` pointers do NOT keep the parent alive. When the caller drops A, it is freed.
- After A is freed, B's `Weak<A>` is dead. The walk from a future child of B stops at B.
- In the pipelining pattern, committed ancestors are dropped after `apply_batch`. Their `Weak` refs expire, naturally trimming the walk to O(1) ancestors. This is the mechanism that prevents memory growth.
- `B.ancestors` and `C.ancestors` hold `Arc` refs to each ancestor's local data (e.g., `Arc<BTreeMap>` or `Arc<Vec>`), not to the full `MerkleizedBatch` struct. Only the data allocation survives when an ancestor is freed. This data is used by `apply_batch`.

## The memory problem (before)

Previously, `merkleize()` eagerly merged all ancestor data into the child:

- **Merkle layer**: the strong `Layer -> Layer -> ... -> Base(Mem)` enum chain kept the entire history alive. `finalize()` walked the chain to produce a `Changeset`.
- **Auth journal layer**: a strong `parent: Option<Arc<Self>>` chain kept every ancestor's `MerkleizedBatch` alive. `collect_items()` walked the chain at finalize time.
- **QMDB layer**: `finish()` merged the parent's diff into the child's `BTreeMap`, accumulating all ancestor entries.

In a pipelining pattern (build B2 from B1, apply B1, build B3 from B2, ...), each batch accumulated the full history. After N iterations with K entries each, the live batch held O(N*K) entries. Memory grew without bound.

## The memory fix (after)

Each batch now stores only its own O(K) entries. The ancestors hold `Arc` refs to each ancestor's data -- not copies. In the pipelining pattern, committed ancestors are dropped and their `Weak` refs expire. The walk at merkleize time collects only live (uncommitted) ancestors.

### Pipelining steady state

```rust
let mut prev = db.new_batch().write(k, Some(v)).merkleize(&db, None).await?;
loop {
    let next = prev.new_batch().write(k, Some(v)).merkleize(&db, None).await?;
    db.apply_batch(prev).await?;
    prev = next;
}
```

At each iteration:
- `next.data` = O(K) (this batch's entries only)
- `next.ancestors` = `[Arc<prev.data>]` (one Arc ref, length 1)
- `prev` is consumed by `apply_batch` and freed
- `prev`'s `Weak` ref in `next` expires -- the next iteration's walk starts fresh

Steady-state memory: the live batch's data (K) + one Arc ref to the previous batch's data (K) = 2K. The previous batch's data is freed when the live batch is applied. **Memory is O(K) per iteration, never growing.**

### Why `Arc` refs, not copies

Each ancestor entry (e.g., `ancestor_diffs[i]`) is an `Arc::clone` of the ancestor's own data field. This is an 8-byte pointer copy, not a deep clone of the `BTreeMap`. The underlying data is shared between the ancestor's own field and the descendant's ancestor entry. When the ancestor is dropped, the descendant's `Arc` ref keeps just that one data structure alive -- not the entire `MerkleizedBatch`.

### Why `Weak`, not strong

Strong parent pointers would keep every ancestor alive as long as any descendant exists. With `Weak`, an ancestor is freed as soon as the caller drops it (regardless of how many descendants exist). In the pipelining pattern, the committed ancestor is dropped after `apply_batch`, its `Weak` ref expires, and the next merkleize walk naturally excludes it. This is what keeps memory at O(K).

## Eliminating deep copies of the Mem

The in-memory Merkle structure (`Mem`) previously used `Arc<MemInner>` internally. Batch types stored `Mem` clones (cheap Arc refcount bumps) so that `MerkleizedBatch::get_node()` could fall through to the committed structure for proof generation. When `Mem::apply_batch()` called `Arc::make_mut()`, any outstanding clone caused a deep copy of the entire node set -- O(N) where N is the number of retained nodes.

This deep copy happened once per pipelining cycle: the next batch's `UnmerkleizedBatch` held a `Mem` clone while the previous batch was being applied. For a pruned MMR with 1M retained SHA-256 nodes, this was ~32 MiB copied on every commit.

### The fix

Three changes eliminate this entirely:

1. **Remove `committed` from `MerkleizedBatch`** (`batch.rs`). The field stored a `Mem` clone inherited by all descendants. `get_node()` used it as a fallback for committed positions. Now `get_node()` walks the `Weak` parent chain. For positions in the committed structure, callers fall through to `Mem::get_node()` or use the `BatchOverMem` adapter in `current/batch.rs`.

2. **Remove `committed_mem` from authenticated batch types** (`authenticated.rs`). Same pattern: the `Mem` clone was inherited through the batch chain. Now `merkleize()` borrows `&Mem` at call time and releases it immediately.

3. **Remove `Arc` from `Mem`** (`mem.rs`). With no batch type storing a `Mem` clone, the `Arc` refcount was always 1 at mutation time -- `Arc::make_mut` was always a no-op. The `Arc` is removed entirely. `Mem` is now a plain owned struct. `Clone` deep-copies (only used in one test on ~20 nodes).

### Compile-time enforcement

`Journaled::mem()` (which returned an owned `Mem` clone) is deleted. The only access path is `Journaled::with_mem(|mem| ...)`, which borrows `&Mem` through a read lock. The compiler guarantees no owned clone can escape. Deep copies of the Mem are structurally impossible through the `Journaled`/`Journal` API.

## Merkleize and the `&Mem` parameter

The merkle batch's `merkleize(base: &Mem, hasher)` takes the committed `Mem` as its first parameter. This is needed because hash computation requires sibling node lookups, and those siblings may be in the committed structure (not in any batch).

Node resolution during merkleize:
1. This batch's own appended/overwrites
2. Parent's data (strong `Arc`, always alive during merkleize)
3. Grandparent chain (`Weak` walk -- stops at dead refs)
4. `base` Mem fallback (covers positions in the committed structure)

### Where `base` comes from at each layer

| Caller | How `&Mem` is obtained |
|--------|------------------------|
| `Mem::new_batch().merkleize(&mem, &hasher)` | Passed explicitly by the caller |
| `Journaled` internal (`append`, `from_components`) | `self.with_mem(\|mem\| batch.merkleize(mem, hasher))` |
| `authenticated::Journal` internal (`append`) | `self.merkle.with_mem(\|mem\| ub.merkleize(mem))` |
| Keyless/Immutable `merkleize(&db, metadata)` | `db.journal.with_mem(\|mem\| journal_batch.merkleize(mem))` |
| Any `finish(&db)` | `db.log.with_mem(\|base\| self.journal_batch.merkleize_with(base, ops))` |

In every case, the `&Mem` borrow is held only for the duration of the synchronous `merkleize` call. The read lock is released immediately after.

### `BatchOverMem` adapter (`current/batch.rs`)

The grafted MMR in `compute_current_layer` needs node resolution that spans both the speculative batch and the committed Mem. Since `MerkleizedBatch::get_node()` no longer falls through to committed data, a `BatchOverMem` adapter layers the batch over the Mem:

```rust
let layered = BatchOverMem { batch: &grafted_batch, mem: &current_db.grafted_mmr };
let grafted_storage = grafting::Storage::new(&layered, ...);
```

This implements `Readable`, checking the batch first and falling through to the Mem for committed positions.

## `apply_batch` internals

Each `apply_batch` implementation detects whether ancestors have already been committed by comparing the committed storage's current size against the batch's `base_size` (the size when the chain was forked):

- `committed_size == base_size`: ancestors not yet committed. Apply ancestors (root-to-tip), then this batch.
- `committed_size > base_size`: ancestors already committed. Skip ancestors, apply only this batch.
- Otherwise: stale batch (a sibling or unrelated fork was committed). Return `StaleBatch` error.

After applying, `Mem::apply_batch` verifies `self.size() == batch.size()`. A mismatch means ancestor data was lost (an uncommitted ancestor was dropped before merkleize). Returns `Error::AncestorDropped`.

## What was removed

- `MerkleizedBatch` enum variants (`Base`, `Layer`, `Checkpoint`)
- `MerkleizedBatchLayer` struct
- `Changeset` struct
- `finalize()`, `finalize_from()`, `collect_overwrites()`
- `flatten()`, `push_batch()` on `MerkleizedBatch`
- `Mem::apply(Changeset)`
- `Journaled::apply(Changeset)`
- `Journaled::mem()` (returns owned clone -- replaced by `with_mem`)
- `committed` field on merkle `MerkleizedBatch`
- `committed_mem` field on authenticated `UnmerkleizedBatch` and `MerkleizedBatch`
- `Arc<MemInner>` indirection inside `Mem` (now a plain struct)
- `MemInner` struct
- `grafted_parent_size` field on `current::MerkleizedBatch`
- `Error::StaleChangeset` (renamed to `Error::StaleBatch`)

### DIFF
--- a/storage/src/qmdb/any/batch.rs
+++ b/storage/src/qmdb/any/batch.rs
@@ -804,13 +804,20 @@ where
 
         // Process creates: remaining mutations (fresh keys) plus parent-deleted
         // keys being re-created. Both get an Update op and active_keys_delta += 1.
-        let fresh = mutations
-            .into_iter()
-            .filter_map(|(k, v)| v.map(|v| (k, v, None)));
-        let recreates = parent_deleted_creates
-            .into_iter()
-            .map(|(k, (v, loc))| (k, v, loc));
-        for (key, value, base_old_loc) in fresh.chain(recreates) {
+        // Merge into a single sorted Vec so iteration order is deterministic
+        // regardless of whether the parent is pending or committed.
+        let mut creates: Vec<(K, V::Value, Option<Location<F>>)> =
+            Vec::with_capacity(mutations.len() + parent_deleted_creates.len());
+        for (key, value) in mutations {
+            if let Some(value) = value {
+                creates.push((key, value, None));
+            }
+        }
+        for (key, (value, base_old_loc)) in parent_deleted_creates {
+            creates.push((key, value, base_old_loc));
+        }
+        creates.sort_by(|(a, _, _), (b, _, _)| a.cmp(b));
+        for (key, value, base_old_loc) in creates {
             let new_loc = Location::new(m.base_size + ops.len() as u64);
             ops.push(Operation::Update(update::Unordered(
                 key.clone(),
@@ -922,22 +929,26 @@ where
 
         // Remaining mutations are creates. Each entry carries the value and
         // base_old_loc (None for fresh creates, Some for parent-deleted recreates).
-        let mut created: BTreeMap<K, (V::Value, Option<Location<F>>)> = BTreeMap::new();
+        // Merge into a single sorted Vec so iteration order is deterministic
+        // regardless of whether the parent is pending or committed.
+        let mut created: Vec<(K, V::Value, Option<Location<F>>)> =
+            Vec::with_capacity(mutations.len() + parent_deleted_creates.len());
         for (key, value) in mutations {
             let Some(value) = value else {
                 continue; // delete of non-existent key
             };
-            created.insert(key.clone(), (value, None));
-            next_candidates.insert(key);
+            next_candidates.insert(key.clone());
+            created.push((key, value, None));
         }
         for (key, (value, base_old_loc)) in parent_deleted_creates {
             next_candidates.insert(key.clone());
-            created.insert(key, (value, base_old_loc));
+            created.push((key, value, base_old_loc));
         }
+        created.sort_by(|(a, _, _), (b, _, _)| a.cmp(b));
 
         // Look up prev_translated_key for created/deleted keys.
         let mut prev_locations = Vec::new();
-        for key in deleted.keys().chain(created.keys()) {
+        for key in deleted.keys().chain(created.iter().map(|(k, _, _)| k)) {
             let Some((iter, _)) = db.snapshot.prev_translated_key(key) else {
                 continue;
             };
@@ -976,11 +987,11 @@ where
         };
 
         for (key, entry) in &ancestor_entries {
+            // Skip keys already handled by this batch's mutations.
             if updated.contains_key(*key)
-                || created.contains_key(*key)
+                || created.binary_search_by(|(k, _, _)| k.cmp(*key)).is_ok()
                 || deleted.contains_key(*key)
             {
-                // Skip keys already handled by this batch's mutations.
                 continue;
             }
             if let DiffEntry::Active { value, loc, .. } = entry {
@@ -1004,7 +1015,9 @@ where
             next_candidates.remove(key);
         }
         for (key, entry) in &ancestor_entries {
-            if matches!(entry, DiffEntry::Deleted { .. }) && !created.contains_key(*key) {
+            if matches!(entry, DiffEntry::Deleted { .. })
+                && created.binary_search_by(|(k, _, _)| k.cmp(*key)).is_err()
+            {
                 prev_candidates.remove(*key);
                 next_candidates.remove(*key);
             }
@@ -1056,7 +1069,7 @@ where
         let mut created_keys: Vec<K> = Vec::with_capacity(created.len());
 
         // Process creates.
-        for (key, (value, base_old_loc)) in created {
+        for (key, value, base_old_loc) in created {
             created_keys.push(key.clone());
             let new_loc = Location::new(m.base_size + ops.len() as u64);
             let next_key = find_next_key(&key, &next_candidates);
@@ -2020,6 +2033,82 @@ mod tests {
                     Some(colliding_digest(i, 1))
                 );
             }
+
+            db.destroy().await.unwrap();
+        });
+    }
+
+    /// Regression test for issue #3519 / #3520: when a parent batch deletes a
+    /// key that has a collision sibling and the child re-creates that key, the
+    /// `fresh.chain(recreates)` iterator produced operations in a different
+    /// order depending on whether the parent was pending or committed.
+    #[test]
+    fn recreate_deleted_key_with_collision_sibling_root_matches() {
+        let runner = deterministic::Runner::default();
+        runner.start(|context| async move {
+            type TestDb = UnorderedFixedDb<
+                mmr::Family,
+                deterministic::Context,
+                sha256::Digest,
+                sha256::Digest,
+                Sha256,
+                OneCap,
+            >;
+
+            let config = fixed_db_config::<OneCap>("recreate-deleted-collision", &context);
+            let mut db = TestDb::init(context, config).await.unwrap();
+
+            // Two colliding keys: K0 (suffix 0) and K6 (suffix 6).
+            let k0 = colliding_digest(0xAA, 0);
+            let k6 = colliding_digest(0xAA, 6);
+
+            // Seed both keys so the snapshot bucket contains two entries.
+            let initial = db
+                .new_batch()
+                .write(k0, Some(colliding_digest(0xBB, 0)))
+                .write(k6, Some(colliding_digest(0xBB, 6)))
+                .merkleize(&db, None)
+                .await
+                .unwrap();
+            db.apply_batch(initial).await.unwrap();
+            db.commit().await.unwrap();
+
+            // Parent: delete K0. K6 remains untouched.
+            let parent = db
+                .new_batch()
+                .write(k0, None)
+                .merkleize(&db, None)
+                .await
+                .unwrap();
+
+            // Child (pending parent): re-create K0 and write a new colliding key K29.
+            let k29 = colliding_digest(0xAA, 29);
+            let pending_child = parent
+                .new_batch::<Sha256>()
+                .write(k0, Some(colliding_digest(0xCC, 0)))
+                .write(k29, Some(colliding_digest(0xCC, 29)))
+                .merkleize(&db, None)
+                .await
+                .unwrap();
+
+            // Commit the parent, then rebuild the same child.
+            db.apply_batch(parent).await.unwrap();
+            db.commit().await.unwrap();
+
+            let committed_child = db
+                .new_batch()
+                .write(k0, Some(colliding_digest(0xCC, 0)))
+                .write(k29, Some(colliding_digest(0xCC, 29)))
+                .merkleize(&db, None)
+                .await
+                .unwrap();
+
+            assert_eq!(
+                pending_child.root(),
+                committed_child.root(),
+                "root depended on pending-vs-committed parent path \
+                 when re-creating a deleted key with collision siblings"
+            );
 
             db.destroy().await.unwrap();
         });


storage/src/qmdb/any/batch.rs
```diff
@@ -1243,59 +1227,78 @@ where
     H: Hasher,
     Operation<F, U>: Codec,
 {
-    /// Apply a changeset to the database, returning the range of written operations.
+    /// Apply a batch to the database, returning the range of written operations.
     ///
-    /// A changeset is only valid if the database has not been modified since the batch that
-    /// produced it was created. Multiple batches can be forked from the same parent for speculative
-    /// execution, but only one may be applied. Applying a stale changeset returns
-    /// [`crate::qmdb::Error::StaleChangeset`].
```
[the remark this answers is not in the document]
danlaine: good bot

patrick-ogrady:
Got this:

```

  - [P2] Preserve ancestor bitmap segments in root-to-tip order — /Users/patrickogrady/code/monorepo-pr-3503/storage/src/qmdb/current/
    batch.rs:688-690
    For chains deeper than two, this stores ancestor bitmap segments as [parent, grandparent, ...], but current::Db::apply_batch later
    concatenates ancestor_bitmap_pushes/ancestor_bitmap_clears in that stored order before appending the leaf segment. If those generations have
    different active/inactive bits (for example, a grandparent write later cleared by the parent), applying the leaf directly pushes bitmap bits
    in B,A,C order instead of A,B,C, so bitmap positions stop matching journal locations and later floor scans or pruning will operate on the
    wrong ops.
```

patrick-ogrady:
While I think this is "true", it is probably fine:

```

  - [P1] Reject same-size foreign prefixes in any::Db::apply_batch — /Users/patrickogrady/code/monorepo-pr-3503/storage/src/qmdb/any/
    batch.rs:1257-1261
    This size-only check accepts a stale child whenever a non-ancestor prefix happens to have the same operation count as the child's parent
    chain. For example, if A and unrelated X are both one write long, then B built from A passes here after X commits because db_size ==
    batch.base_size; finalize_from then appends only B's local ops on top of X, so apply_batch silently commits X+B instead of returning
    StaleChangeset. That violates the new chained-batch contract and can leave the DB at a root different from batch.root().
```

storage/src/qmdb/any/batch.rs
```diff
@@ -1243,59 +1241,79 @@ where
...
-                    update_known_loc::<F, _>(&mut self.snapshot, &key, old_loc, new_loc);
-                }
-                SnapshotDiff::Insert { key, new_loc } => {
-                    self.snapshot.insert(&key, new_loc);
+        // 2. Apply snapshot diffs.
+        let mut seen = BTreeSet::<&U::Key>::new();
+        for (key, entry) in batch.diff.iter() {
+            if skip_ancestors {
+                if entry.loc().is_some_and(|loc| *loc < db_size) {
+                    continue;
```
roberto-bayardo, line 1294: My AI is telling me this condition can never trigger (same for immutable's batch update). 

```
With the new PR design, batch.diff exclusively contains the specific operations generated during this specific unmerkleized batch step.

The localized operations start precisely at batch.base_size, guaranteeing that every entry inside batch.diff holds *entry.loc >= batch.base_size.
In cases where skip_ancestors is evaluated as true, the upfront staleness check dictates that db_size == batch.base_size.
Therefore, the condition *entry.loc < db_size resolves to *entry.loc < batch.base_size, which is functionally an impossible state for the entries generated by this batch! The continue instruction will never be triggered.
```

storage/src/qmdb/current/batch.rs
```diff
@@ -723,32 +717,44 @@ where
...
-    Ok(MerkleizedBatch {
+    // Collect ancestor bitmap data while the parent is still alive.
+    let (ancestor_bitmap_pushes, ancestor_bitmap_clears) =
+        parent.as_ref().and_then(Weak::upgrade).map_or_else(
+            || (Vec::new(), Vec::new()),
+            |p| {
+                let mut pushes = vec![Arc::clone(&p.bitmap_pushes)];
+                let mut clears = vec![Arc::clone(&p.bitmap_clears)];
+                pushes.extend(p.ancestor_bitmap_pushes.iter().cloned());
+                clears.extend(p.ancestor_bitmap_clears.iter().cloned());
```
roberto-bayardo: AI is suggesting this should be oldest-first, though no test seems to catch this.

```suggestion
                let mut pushes = p.ancestor_bitmap_pushes.clone();                                                                                             
                pushes.push(Arc::clone(&p.bitmap_pushes));                                                                                                     
                let mut clears = p.ancestor_bitmap_clears.clone();                                                                                             
                clears.push(Arc::clone(&p.bitmap_clears));             
```

danlaine: Resolved by https://github.com/commonwarexyz/monorepo/pull/3503/commits/92b95b3f46b0a53842fd29134a8bda2ad4f6bf26

danlaine:
> While I think this is "true", it is probably fine:
> 
> ```
> 
>   - [P1] Reject same-size foreign prefixes in any::Db::apply_batch — /Users/patrickogrady/code/monorepo-pr-3503/storage/src/qmdb/any/
>     batch.rs:1257-1261
>     This size-only check accepts a stale child whenever a non-ancestor prefix happens to have the same operation count as the child's parent
>     chain. For example, if A and unrelated X are both one write long, then B built from A passes here after X commits because db_size ==
>     batch.base_size; finalize_from then appends only B's local ops on top of X, so apply_batch silently commits X+B instead of returning
>     StaleChangeset. That violates the new chained-batch contract and can leave the DB at a root different from batch.root().
> ```

Yeah, I would call this user error

storage/src/qmdb/any/batch.rs
danlaine: Removed

danlaine:
~~Argh I accidentally re-introduced a memory leak in the "pipeline" case where a batch is always alive. Marking as draft again until I fix this.~~ Fixed

storage/fuzz/fuzz_targets/merkle_journaled_crash_recovery.rs
```diff
@@ -129,12 +129,9 @@ async fn run_operations<F: MerkleFamily>(
...
         let failed = match op {
             MerkleOperation::Add { data } => {
-                let changeset = merkle
-                    .new_batch()
-                    .add(hasher, data)
-                    .merkleize(hasher)
-                    .finalize();
-                merkle.apply(changeset).unwrap();
+                let batch = merkle
+                    .with_mem(|mem| merkle.new_batch().add(hasher, data).merkleize(mem, hasher));
```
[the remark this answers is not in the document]
danlaine: Not a deadlock. parking_lot::RwLock allows reentrant reads on the same thread. with_mem holds a read guard; new_batch acquires another read guard on the same lock. Both are reads — no conflict. That said, moved new_batch() outside the closure for clarity.

storage/src/merkle/mem.rs
```diff
@@ -425,6 +425,17 @@ impl<F: Family, D: Digest> Mem<F, D> {
             self.nodes.push_back(digest);
         }
 
+        // Detect missing ancestor data. If an uncommitted ancestor was dropped
+        // before this batch was merkleized, its appended nodes are absent and the
+        // Mem ends up smaller than expected. This does not catch dropped
+        // overwrite-only ancestors (they don't change the size).
+        if self.size() != batch.size() {
```
patrick-ogrady, line 432: Won't adding a `commit` always increase size?

danlaine: At the QMDB level, yes, but it's possible to overwrite leaves of an MMR without appending anything in a batch

patrick-ogrady (approved):

PR #3504, opened 2026-03-31
[coding/zoda] Replace O(data_size) linear scan in Topology::reckon with binary search

### LINKED ISSUE #3468: [coding/zoda] Topology::reckon uses O(data_size) BigRational iterations on the encode/verify hot path
## Summary

`Topology::reckon` in `coding/src/zoda/topology.rs` finds the optimal column layout for ZODA encoding by incrementing `data_cols` from 1 upward and calling `required_samples()` at each step. `required_samples()` uses `BigRational` arbitrary-precision arithmetic including a `log2_ceil` at 128-bit precision. The loop runs up to `ceil(8 * data_bytes / 63)` iterations - approximately 132,000 for a 1 MiB shard.

`reckon` sits on two hot paths:
  - `Scheme::encode` - once per block encode (`mod.rs:526`)
  - `CheckingData::reckon`  - once per received shard (`mod.rs:392`)

  ## Why this is expensive

  Goldilocks encodes 63 bits per element, so for a 1 MiB shard:
  data_els = ceil(8 * 1_048_576 / 63) ≈ 132,668 iterations

  Each iteration allocates `BigRational` values and computes `log2_ceil` with 128-bit precision. This is orders of magnitude slower than integer or `f64` arithmetic. For every block a node encodes or every shard it verifies, it pays this cost before doing any actual encoding work.

  ## Root cause

  The loop is linear because it tries every integer column count. But `encoded_rows` has structure that makes this unnecessary:

  encoded_rows = next_power_of_two((n + k) * ceil(ceil(data_els / cols) / n))

  As `cols` increases, `encoded_rows` only changes at `O(log(data_els))` boundaries — roughly 17 distinct values for 1 MiB of data. Between boundaries, `required_samples` is identical. The loop recomputes the same result thousands of times per plateau.

  ## Proposed fix

  Iterate over the `O(log(data_els))` power-of-two steps of `encoded_rows` instead of every integer column count. For each step, the largest valid `data_cols` that lands on it can be derived directly from the inverse relationship:

  max_data_rows = n * (encoded_rows / (n + k))
  best_cols     = ceil(data_els / max_data_rows)

  Then call `required_samples()` once per step (~17 calls total) instead of once per column (~132K calls). This reduces the search from `O(data_els)` to `O(log(data_els))` with no change to the output - the function signature, return type, and semantics are identical.

Planning to implement this with a test to validate correctness against the current implementation before removing it. Does this approach seem reasonable, or is there a cleaner way to handle the math? Happy to hear if there's anything else worth considering here.

0xAysh:
Fixes #3468

  ## Problem

  `Topology::reckon` found the optimal column layout by incrementing `data_cols` from 1 upward, calling `required_samples()` (BigRational log2 arithmetic) at each step. For a 1 MiB shard this is ~132,000 iterations, and `reckon` runs on two hot paths: once per block encode and once per received shard.

  ## Fix

  `encoded_rows = next_power_of_two((n+k)*samples)` takes only O(log) distinct values as `data_cols` varies. Within each fixed value of `encoded_rows`, increasing `data_cols` reduces `required_samples`, so valid configs form a contiguous suffix within each range. The globally best (largest) valid `data_cols` is therefore the rightmost valid col across all ranges, which a standard binary search finds in O(log(data_els)) steps.

  For a 1 MiB shard: ~132,000 BigRational iterations down to ~17.

  One subtlety: the original loop stopped at the first invalid col, which would miss valid cols in a later range. The binary search
  handles this correctly.

  ## Tests

  - Exhaustive differential test against the original loop for all inputs up to 1 KiB (4.3M configs)
  - Spot-check against a full-scan oracle at 10 KB and 100 KB inputs to cover the non-monotone case
  - Security invariant check on every result: both `samples` and `column_samples` must cover the 126-bit requirement

### DIFF
--- a/coding/src/zoda/topology.rs
+++ b/coding/src/zoda/topology.rs
@@ -98,11 +98,13 @@ impl Topology {
         // then this configuration is valid, using `R` as the necessary number
         // of samples.
         //
-        // The validity condition is monotone: as cols increases, encoded_rows
-        // shrinks and required_samples grows, so once the condition fails it
-        // never recovers. We exploit this with a binary search over [1, data_els]
-        // instead of a linear scan, reducing BigRational log2 calls from
-        // O(data_els) to O(log(data_els)).
+        // Within each encoded_rows plateau (a fixed power-of-two value), increasing
+        // cols reduces data_rows and required_samples, so the validity condition
+        // improves as cols increases within a plateau. Valid column counts therefore
+        // form a contiguous suffix within each plateau, and the globally best (largest)
+        // valid col is the rightmost valid col across all plateaus.
+        // A binary search over [2, data_els] finds this rightmost valid col,
+        // reducing BigRational log2 calls from O(data_els) to O(log(data_els)).
         //
         // It's possible that even cols=1 is not good. To correct for that, we
         // need to add extra checksum columns to guarantee security.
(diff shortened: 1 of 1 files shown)


clabby:

Issue #3453, opened 2026-03-24
[examples/reshare] Pre-Verify `SignedDealerLogs`


patrick-ogrady:
Related: #3328 

Instead of waiting to verify all logs at the epoch boundary, we should pre-verify logs asynchronously (to minimize the end of epoch "stutter").

erenyegit:
Hi @patrick-ogrady 

I've implemented the asynchronous pre-verification for `SignedDealerLogs` to minimize end-of-epoch stutter as requested. 

You can find the implementation in PR #3505.

**Key Highlights:**
- Offloaded CPU-heavy verification to background tasks using `commonware_runtime::Spawner` with `shared(true)`.
- Implemented a versioning and dirty-flag pattern to ensure data consistency and avoid redundant tasks.
- Verified the fix with `cargo test -p commonware-reshare`. All 34 tests passed, including the long-running deterministic and crash-recovery suites (625.28s).

PR #3505, opened 2026-03-31
[examples/reshare] pre-verify dealer logs asynchronously

erenyegit:
## Summary
- Maintain a warm `cryptography::bls12381::dkg::Logs` cache in `examples/reshare/src/dkg/actor.rs`.
- As finalized blocks arrive, record newly persisted dealer logs and asynchronously run `Logs::pre_verify` on snapshots.
- Spawn background work via `commonware_runtime::Spawner` (`self.context.clone().shared(true).spawn(...)`) to avoid stalling the cooperative executor.
- Reuse the warmed cache at the epoch boundary to reduce end-of-epoch verification stutter.

## Motivation
Issue #3453 requests minimizing end-of-epoch “stutter” by amortizing verification across the epoch (related to #3328).

## Implementation notes
- **Persistence first:** finalized blocks drive recording; we persist via `append_log` before updating the warm cache and scheduling `pre_verify`.
- **Coalescing:** an in-flight + dirty-flag pattern limits background work to at most one active task per epoch, rerunning only when new logs arrive.
- **Staleness protection:** pre-verified snapshots are accepted only when `(epoch, version)` matches the actor’s current view.

## Test plan
- `cargo check -p commonware-reshare`
- `cargo test -p commonware-reshare` (34 tests)

### DIFF
--- a/examples/reshare/src/dkg/actor.rs
+++ b/examples/reshare/src/dkg/actor.rs
@@ -40,6 +40,13 @@ use rand_core::CryptoRngCore;
 use std::num::NonZeroU32;
 use tracing::{debug, info, warn};
 
+/// Background `Logs::pre_verify` completed for a snapshot of [`Logs`] at `version` for `epoch`.
+struct PreVerifiedLogs<V: Variant, P: PublicKey> {
+    epoch: Epoch,
+    version: u64,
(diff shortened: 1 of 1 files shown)


clabby:

PR #3512, opened 2026-04-01
[consensus/coding] Add `Reporter` for shard engine activity

clabby:
## Overview

Adds an optional `Reporter` to the shard engine that receives events when shards are verified or a block is reconstructed.

This is useful for observers that may want high-frequency indexing of shards, for example to facilitate a dashboard like https://gui.firedancer.io

### DIFF
--- a/consensus/src/marshal/coding/shards/engine.rs
+++ b/consensus/src/marshal/coding/shards/engine.rs
@@ -141,7 +141,7 @@ use crate::{
         validation::{validate_reconstruction, ReconstructionError as InvariantError},
     },
     types::{coding::Commitment, Epoch, Round},
-    Block, CertifiableBlock, Heightable,
+    Block, CertifiableBlock, Heightable, Reporter,
 };
 use commonware_codec::{Decode, Error as CodecError, Read};
 use commonware_coding::{Config as CodingConfig, Scheme as CodingScheme};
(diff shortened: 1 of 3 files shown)


consensus/src/marshal/coding/shards/engine.rs
```diff
@@ -255,13 +272,18 @@ where
     /// Provider for peer set information. Per-peer shard buffers
     /// are freed when a peer leaves all tracked peer sets.
     pub peer_provider: D,
+
+    /// Reporter for shard-engine activity.
+    ///
+    /// Reporting is disabled when this is [`None`].
+    pub reporter: Option<R>,
```
patrick-ogrady, line 279: I'm AFK but IIRC we usually require some reporter?

consensus/src/marshal/mocks/harness.rs
```diff
@@ -70,6 +70,12 @@ pub type P = ConstantProvider<S, Epoch>;
 pub type CodingCtx = Context<Commitment, K>;
 pub type CodingB = Block<D, CodingCtx>;
 
+impl Reporter for () {
```
patrick-ogrady, line 73: Have not needed this pattern before...are we doing something unusual with how we provide a reporter?

I'm assuming this is just the way to provide a "no-op" reporter?

consensus/src/marshal/coding/shards/engine.rs
clabby: At least here, it's not required for any critical function (unlike the marshal reporter, where acks have to be handled, for example). We could require one and have folks use a no-op reporter, though.

consensus/src/marshal/mocks/harness.rs
clabby: Yeah this is just to satisfy type inference. Later on in this file we have to specify some types on the `Config`, so we need to name the reporter.

clabby:
Closing; Didn't end up using this in constantinople.

PR #3515, opened 2026-04-01
[Storage] Fix `next_key` corruption when child batch deletes a key with collision siblings

danlaine:
## Problem

When a parent batch updates a key K and a child batch deletes K, if K shares a translated-key bucket with another key B (i.e. they collide under the translator), B's `next_key` pointer is never updated. It continues pointing to the now-deleted K, corrupting the ordered key ring. Subsequent lookups for keys in the gap between B and K's successor panic with `possible_next should not be empty` or return incorrect results.

### Root cause

`gather_existing_locations` (`batch.rs`) short-circuits with `continue` whenever `base_diff` has *any* entry for a mutation key, skipping the `db.snapshot.get(key)` call entirely. The snapshot lookup uses the translated-key index, which returns an iterator over *all* entries in the bucket -- including collision siblings like B. By skipping it, those siblings are never added to `prev_candidates` during merkleization, so their `next_key` pointers are never rewritten when K is deleted.

Two things conspire to make B invisible:

1. **`gather_existing_locations`** finds K in `base_diff` (Active, from the parent's update), pushes its location, and `continue`s -- never scanning the snapshot for collision siblings.
2. **`prev_translated_key`** searches `range(..translated_k)`, which by design excludes entries in the *same* translated-key bucket. This is correct -- same-bucket siblings are supposed to be discovered via `gather_existing_locations`, not the predecessor lookup.

With both paths blocked, B never enters `prev_candidates`, and its `next_key` is left pointing at deleted K.

### Reproduction sequence

```
1. Commit keys A, B, K where B and K collide (same translated bucket)
2. Parent batch: update K's value          -> K enters base_diff as Active
3. Child batch: delete K                   -> gather_existing_locations skips B
4. B's next_key never updated, still points to deleted K
5. Later: lookups in the gap panic
```

## Fix

Replace the `if let Some(entry) / continue` pattern in `gather_existing_locations` with an exhaustive `match` on the base diff entry:

- **Deleted**: Skip entirely. The snapshot location is stale. The key must route through `extract_parent_deleted_creates` to be handled as a re-creation, not an update.
- **Active**: Push the parent's uncommitted location, then scan the snapshot bucket for collision siblings. Filter out the key's own stale committed location (`base_old_loc`) so only genuine siblings are added -- this avoids polluting the location list with a known-superseded location that downstream code would have to tolerate.
- **None** (not in base diff): Fall back to `db.snapshot.get(key)` as before.

The `Active` arm is the core fix: it discovers collision siblings like B that were previously invisible because the unconditional `continue` skipped the snapshot lookup entirely.

### DIFF
--- a/storage/src/qmdb/any/batch.rs
+++ b/storage/src/qmdb/any/batch.rs
@@ -277,13 +277,17 @@ where
         }
     }
 
-    /// Gather the existing-key locations for all keys in `mutations`.
+    /// Gather existing-key locations for all keys in `mutations`.
     ///
     /// For each mutation key, checks the base diff first (returning the
     /// uncommitted location for Active entries, skipping Deleted entries).
-    /// For Active entries, also scans the snapshot bucket for collision
-    /// siblings (excluding the key's own stale committed location). Keys
-    /// not in the base diff fall back to the base DB snapshot.
+    /// Keys not in the base diff fall back to the base DB snapshot.
+    ///
+    /// When `include_active_collision_siblings` is true, Active entries
+    /// also scan the snapshot bucket for collision siblings (other keys
+    /// sharing the same translated-key bucket). The ordered path needs
+    /// these so their `next_key` pointers are rewritten when a sibling
+    /// is deleted; the unordered path can skip them.
     fn gather_existing_locations<E, C, I>(
         &self,
         mutations: &BTreeMap<U::Key, Option<U::Value>>,


storage/src/qmdb/any/batch.rs
```diff
@@ -280,8 +280,10 @@ where
     /// Gather the existing-key locations for all keys in `mutations`.
```
roberto-bayardo: Does this comment need to be updated? Seems we're gathering all colliding key locations as well?

danlaine: https://github.com/commonwarexyz/monorepo/pull/3515/commits/747919338cd41661a6f4be51189ceacdf6a0c918

Updated comment

PR #3531, opened 2026-04-03
[storage/qmdb/any] bitmap for anydb

### LINKED ISSUE #1829: [storage/adb/any] Maintain non-merkelized bitmap to use in floor raising
Maintain a non-Merkelized bitmap of activity status to improve performance of floor-raising in adb::any. More details in the discussion here: https://github.com/commonwarexyz/monorepo/issues/1805

roberto-bayardo:
Adds an activity bitmap to the anydb for optimized floor raising.

Resolves: https://github.com/commonwarexyz/monorepo/issues/1829

### DIFF
--- a/storage/src/qmdb/any/batch.rs
+++ b/storage/src/qmdb/any/batch.rs
@@ -382,57 +382,6 @@ where
         creates
     }
 
-    /// Scan forward from `floor` to find the next active operation, re-append it at the tip.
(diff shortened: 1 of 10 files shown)


patrick-ogrady:
As discussed, would be **fantastic** if we could make the fetch (after walking the bitmap) concurrent. I think the same "should" be able to go for all QMDB variants with a scan like this?

PR #3532, opened 2026-04-03
[cryptography] reject ed25519 signatures with non-canonical S at decode time

### LINKED ISSUE #1070: [cryptography] Signature Scheme Validity Consistency
Unlike the other signature variants, the `ed25519-consensus` crate, which is the library used to implement `ed25519` signatures, only verifies the length of the byte array during deserialization time: https://github.com/penumbra-zone/ed25519-consensus/blob/78f83b21767077f602f30148e513a52e844abbb6/src/signature.rs#L22-L31
This is not an exploitable issue because the signature's validity will be checked during signature verification time: https://github.com/penumbra-zone/ed25519-consensus/blob/78f83b21767077f602f30148e513a52e844abbb6/src/verification_key.rs#L239-L244
However, checking the signature during deserialization may be more robust against invalid curve points if developers choose to perform any operations on signatures. In addition, it would be more consistent with the other signature implementations, as these all check validity during deserialization time.

0xAysh:
## Problem

`ed25519::Signature::read_cfg` accepted any 64-byte sequence without validation. A signature with S >= l (the ed25519 group order) would decode successfully but fail verification later. This is inconsistent with the rest of the codebase:

- `ed25519::PublicKey::read_cfg` calls `VerificationKey::try_from` — validates the compressed Edwards point at decode time
- `secp256r1::Signature::read_cfg` calls `p256::ecdsa::Signature::from_slice` — validates R and S at decode time
- `ed25519::Signature::read_cfg` — previously did nothing beyond confirming 64 bytes were present

Non-canonical S values introduce signature malleability: two different byte sequences (S and S mod l) decode to logically equivalent signatures since verification reduces S mod l. In adversarial environments this matters — a valid signature can be transformed into a "different" one that also passes verification.

## Fix

Add a canonicality check for S (bytes 32..63 of the signature) in `read_cfg`. S must be in [0, l-1] where:

```
l = 2^252 + 27742317777372353535851937790883648493
```

The check compares S against l byte-by-byte from the most significant byte down (required because S is stored little-endian — index 0 is the LSB, so Rust's default lexicographic array comparison would be incorrect here).

The `ed25519_consensus` library does not expose standalone signature validation without a message, so the R component (bytes 0..31) cannot be validated at decode time without adding a dependency on `curve25519-dalek`. Only S canonicality is checkable, which is the primary malleability concern.

## Tests

- `test_high_s_rejected_at_decode`: S with high bit set (S >= 2^255 >> l)
- `test_s_equal_to_l_rejected_at_decode`: S == l exactly (boundary — valid range is [0, l-1])
- `test_s_above_l_rejected_at_decode`: S == l+1

Closes #1070

### DIFF
--- a/cryptography/src/ed25519/scheme.rs
+++ b/cryptography/src/ed25519/scheme.rs
@@ -31,15 +31,15 @@ const L: [u8; 32] = [
 ];
 
 fn is_canonical_s(s: &[u8; 32]) -> bool {
-    for i in (0..32).rev() {
-        if s[i] < L[i] {
-            return true;
-        }
-        if s[i] > L[i] {
-            return false;
-        }
+    // Constant-time s < L via subtraction with borrow (little-endian, LSB first).
+    // borrow is 1 when the subtraction at position i underflowed, 0 otherwise.
+    // After all 32 bytes, a non-zero borrow means s < L.
+    let mut borrow: u16 = 0;
+    for i in 0..32 {
+        let diff = (s[i] as u16).wrapping_sub(L[i] as u16).wrapping_sub(borrow);
+        borrow = (diff >> 8) & 1;
     }
-    false
+    borrow != 0
 }
 
 /// Ed25519 Private Key.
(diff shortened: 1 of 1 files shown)


clabby:

PR #3534, opened 2026-04-04
Implement Bulletproofs Inner Product Argument

### LINKED ISSUE #3175: Implement Bulletproofs++ IPA
We should implement an inner-product argument, probably https://eprint.iacr.org/2022/510, including the protocol optimizations in BP++, rather than going through a bulletproofs impl. We can have this target the traits in the math crate we already have.

cronokirby:
Closes #3175.

This adds an impl of the inner product argument part of bulletproofs. Pretty happy with the testing strategy here, and the level of documentation.

I also changed the test field and group in the math crate to be larger, that way you can support random challenges without worrying about running into accidental collisions all the time. It's not large enough to be cryptographically secure, but still probably collision free.

### DIFF
--- a/cryptography/Cargo.toml
+++ b/cryptography/Cargo.toml
@@ -82,6 +82,7 @@ arbitrary = [
 	"dep:arbitrary",
 	"std",
 ]
+fuzz = [ "arbitrary", "std" ]
 std = [
 	"anyhow?/std",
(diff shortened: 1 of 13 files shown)


cronokirby:
Spurious breaking-format because of additional conformance items, and because of changes to the test field.

PR #3539, opened 2026-04-06
[storage/qmdb] batch reads in floor raising

roberto-bayardo:
Optimizes the floor raising process by batch-reading active operation candidates, instead of reading them sequentially.

### DIFF
--- a/storage/src/qmdb/any/batch.rs
+++ b/storage/src/qmdb/any/batch.rs
@@ -382,57 +382,6 @@ where
         creates
     }
 
-    /// Scan forward from `floor` to find the next active operation, re-append it at the tip.
-    /// The `scan` parameter controls which locations are considered as potentially active,
-    /// allowing implementations to skip locations known to be inactive without reading them.
-    /// Returns `true` if an active op was found and moved, `false` if the floor reached
-    /// `fixed_tip`.
-    async fn advance_floor_once<E, C, I, S: FloorScan<F>>(
-        &self,
-        floor: &mut Location<F>,
-        fixed_tip: u64,
-        ops: &mut Vec<Operation<F, U>>,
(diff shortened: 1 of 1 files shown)


roberto-bayardo:
bugbot run

storage/src/qmdb/any/batch.rs
```diff
@@ -458,17 +407,71 @@ where
...
-            for _ in 0..total_steps {
-                if !self
-                    .advance_floor_once(&mut floor, fixed_tip, &mut ops, &mut diff, &mut scan, db)
-                    .await?
-                {
+            let mut moved = 0u64;
+
+            while moved < total_steps {
+                // Collect candidates, capped to bound concurrent I/O.
+                const MAX_BATCH: usize = 64;
```
patrick-ogrady: This should be defined outside the function and/or be a config?

storage/src/qmdb/any/batch.rs
```diff
@@ -458,17 +407,71 @@ where
...
                 }
+
+                // Batch-read on-disk candidates through a single reader.
+                let mut disk_ops = {
+                    let reader = db.log.reader().await;
+                    let futures = candidates
+                        .iter()
+                        .filter(|loc| **loc < self.db_size)
+                        .map(|loc| reader.read(**loc));
+                    try_join_all(futures).await?
```
patrick-ogrady: May be nice to do a futures ordered fetch and process results as they come in?

roberto-bayardo: done

storage/src/qmdb/any/batch.rs
roberto-bayardo: Since there's no config type here, we'd have to pass it in, which is a bit cumbersome, so I just moved it outside the function.

PR #3543, opened 2026-04-06
[consensus/simplex+marshal] Auto-certify locally proposed views + Improve Durability Guarantees

clabby:
## Overview

Updates simplex to prevent the local node from asking the application to certify its own proposals. Currently, this causes re-execution of blocks when using `standard::Deferred` or `coding::Marshaled` in the `certify` fallback path.

## TODO

- [x] store in verified or notarized? If in notarized already, should store in verified?


consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -855,8 +858,16 @@ impl<
                     let round = proposal.round;
                     let view = round.view();
                     debug!(%view, "attempting certification");
-                    let receiver = self.automaton.certify(round, proposal.payload).await;
-                    let handle = certify_pool.push(async move { (round, receiver.await) });
+                    let result = if self.state.proposed_locally(view) {
```
patrick-ogrady: I think we may be able to get away with something smaller (if we are ok with putting the result in the pool):

```
diff --git a/consensus/src/simplex/actors/voter/actor.rs b/consensus/src/simplex/actors/voter/actor.rs
index 55ce28c55..817d18d5b 100644
--- a/consensus/src/simplex/actors/voter/actor.rs
+++ b/consensus/src/simplex/actors/voter/actor.rs
@@ -854,9 +854,22 @@ impl<
                 for proposal in self.state.certify_candidates() {
                     let round = proposal.round;
                     let view = round.view();
-                    debug!(%view, "attempting certification");
-                    let receiver = self.automaton.certify(round, proposal.payload).await;
-                    let handle = certify_pool.push(async move { (round, receiver.await) });
+                    let leader_is_local = self
+                        .state
+                        .leader_index(view)
+                        .is_some_and(|leader| self.state.is_me(leader));
+                    let handle = if leader_is_local {
+                        // Like verify, locally proposed blocks do not need to round-trip
+                        // through the automaton before we can continue.
+                        debug!(%view, "skipping certification request for local proposal");
+                        certify_pool.push(async move {
+                            (round, Ok::<_, oneshot::error::RecvError>(true))
+                        })
+                    } else {
+                        debug!(%view, "attempting certification");
+                        let receiver = self.automaton.certify(round, proposal.payload).await;
+                        certify_pool.push(async move { (round, receiver.await) })
+                    };
                     self.state.set_certify_handle(view, handle);
                 }
 
diff --git a/consensus/src/simplex/actors/voter/mod.rs b/consensus/src/simplex/actors/voter/mod.rs
index 0e8c06a22..a4034c5d1 100644
--- a/consensus/src/simplex/actors/voter/mod.rs
+++ b/consensus/src/simplex/actors/voter/mod.rs
@@ -3819,6 +3819,128 @@ mod tests {
         no_recertification_after_replay::<_, _, RoundRobin>(secp256r1::fixture);
     }
 
+    fn proposer_skips_self_certify<S, F>(mut fixture: F)
+    where
+        S: Scheme<Sha256Digest, PublicKey = PublicKey>,
+        F: FnMut(&mut deterministic::Context, &[u8], u32) -> Fixture<S>,
+    {
+        let n = 5;
+        let quorum = quorum(n);
+        let namespace = b"proposer_skips_self_certify".to_vec();
+        let executor = deterministic::Runner::timed(Duration::from_secs(10));
+        executor.start(|mut context| async move {
+            let (network, oracle) = Network::new(
+                context.with_label("network"),
+                NConfig {
+                    max_size: 1024 * 1024,
+                    disconnect_on_block: true,
+                    tracked_peer_sets: None,
+                },
+            );
+            network.start();
+
+            let Fixture {
+                schemes,
+                ..
+            } = fixture(&mut context, &namespace, n);
+
+            let certify_calls: Arc<Mutex<Vec<Sha256Digest>>> = Arc::new(Mutex::new(Vec::new()));
+            let tracker = certify_calls.clone();
+
+            let (mut mailbox, mut batcher_receiver, _, _, _) = setup_voter(
+                &mut context,
+                &oracle,
+                &participants,
+                &schemes,
+                RoundRobin::<Sha256>::default(),
+                Duration::from_secs(100),
+                Duration::from_secs(100),
+                Duration::from_secs(100),
+                mocks::application::Certifier::Custom(Box::new(move |digest| {
+                    tracker.lock().push(digest);
+                    true
+                })),
+            )
+            .await;
+
+            // With epoch 333 and 5 validators, view 2 elects participant 0.
+            let target_view = View::new(2);
+            advance_to_view(
+                &mut mailbox,
+                &mut batcher_receiver,
+                &schemes,
+                quorum,
+                target_view,
+            )
+            .await;
+
+            let proposal = loop {
+                select! {
+                    msg = batcher_receiver.recv() => match msg.unwrap() {
+                        batcher::Message::Constructed(Vote::Notarize(notarize)) if notarize.view() == target_view => {
+                            break notarize.proposal;
+                        }
+                        batcher::Message::Update { response, .. } => response.send(None).unwrap(),
+                    },
+                    _ = context.sleep(Duration::from_secs(2)) => {
+                        panic!("expected local notarize vote for leader view {target_view}");
+                    },
+                }
+            };
+
+            let (_, notarization) = build_notarization(&schemes, &proposal, quorum);
+            mailbox
+                .resolved(Certificate::Notarization(notarization))
+                .await;
+
+            let mut saw_finalize = false;
+            let mut advanced = false;
+            loop {
+                if saw_finalize && advanced {
+                    break;
+                }
+
+                select! {
+                    msg = batcher_receiver.recv() => match msg.unwrap() {
+                        batcher::Message::Constructed(Vote::Finalize(finalize)) if finalize.view() == target_view => {
+                            assert_eq!(finalize.proposal, proposal);
+                            saw_finalize = true;
+                        }
+                        batcher::Message::Update { current, response, .. } => {
+                            response.send(None).unwrap();
+                            if current > target_view {
+                                advanced = true;
+                        }
+                    },
+                    _ = context.sleep(Duration::from_secs(2)) => {
+                        panic!(
+                            "expected local proposal for view {target_view} to finalize and advance"
+                        );
+                    },
+                }
+            }
+
+            assert!(
+                certify_calls.lock().is_empty(),
+                "leader should not be asked to certify its own proposal"
+            );
+        });
+    }
+
+    #[test_traced]
+    fn test_proposer_skips_self_certify() {
+        proposer_skips_self_certify(bls12381_threshold_vrf::fixture::<MinPk, _>);
+        proposer_skips_self_certify(bls12381_threshold_vrf::fixture::<MinSig, _>);
+        proposer_skips_self_certify(bls12381_multisig::fixture::<MinPk, _>);
+        proposer_skips_self_certify(bls12381_multisig::fixture::<MinSig, _>);
+        proposer_skips_self_certify(ed25519::fixture);
+        proposer_skips_self_certify(secp256r1::fixture);
+    }
+
     /// Test that in-flight certification requests are cancelled when finalization occurs.
     ///
     /// 1. Use a very long certify latency to ensure certification is in-flight.

```

patrick-ogrady: I think there is a different version of this PR where we take a different path (directly to the certify handler) if we are the leader but I'm not sure if that buys us anything. The big unlock would be keeping everything in "state" but that doesn't seem possible (because we write certify to disk in voter).

clabby: Yeah - the extra complexity comes from needing to support replay. iiuc, we need the logic to set that flag when we're replaying the `Activity` journal as well as learning about the leader late for this to be safe. A direct lookup on `leader_index` only works if it's already been seeded, which it isn't guaranteed to be right?

consensus/src/simplex/mocks/application.rs
```diff
@@ -130,9 +130,6 @@ type Latency = (f64, f64);
 pub enum Certifier<D: Digest> {
     /// Always certify.
     Always,
-    /// Certify sometimes, but not always. The behavior is to certify pseudorandomly
-    /// (but deterministically) 82% of the time, depending on the last byte of the payload.
-    Sometimes,
```
clabby, line 144: This doesn't really make sense under the model where the proposer always certifies their own view.

consensus/src/simplex/actors/voter/actor.rs
patrick-ogrady: > +                    let leader_is_local = self
+                        .state
+                        .leader_index(view)
+                        .is_some_and(|leader| self.state.is_me(leader));

You may even be able to return this with certification candidate inline?

patrick-ogrady: > A direct lookup on leader_index only works if it's already been seeded, which it isn't guaranteed to be right?

That's correct...that being said, I'm not sure what your changes do that make this less of an issue?

clabby: Wasn't clear, sorry. It's not (yet) - I'm a bit worried about that case, since it's still possible for consensus to ask the automaton to certify a view that we did propose if we haven't seeded the leader (e.g. after a crash that occurs after we've sent the proposal and we get the notarization from the network on restart, I think). The application can add extra tracking to make sure it doesn't do redundant work, but it'd be nice if we had a guarantee that simplex wouldn't ever ask.

clabby: For now, removed the extra state tracking. Will think on this more in the morning.

patrick-ogrady: It makes me wonder how strict we are about `verify` being called in the same case? Might be worth expanding our tests to also cover that case?

clabby: > It makes me wonder how strict we are about verify being called in the same case?

At a glance, it has the same problem.

https://github.com/commonwarexyz/monorepo/blob/10fa6f2b1b2331680b88ba0af05ca759d1067fd5/consensus/src/simplex/actors/voter/round.rs#L111-L118

e: Nevermind. Knowing the leader is a pre-requisite for `Context` formation, the engine delays `verify` until it is known.

> Might be worth expanding our tests to also cover that case?

👍 

---

This is fine in practice for most apps, but adds an extra layer of complexity for the applications to have to deal with that may be nice to avoid. A simple workaround here would just be to have `glue` return `true` on `verify` calls immediately if a pending batch already exists for the block.

clabby:
Pulling into draft while we reason about a more strict guarantee.

consensus/src/simplex/actors/voter/actor.rs
patrick-ogrady: return leader with it?

consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -851,12 +854,19 @@ impl<
...
                 // Attempt to certify any views that we have notarizations for.
-                for proposal in self.state.certify_candidates() {
+                for (proposal, leader_is_local) in self.state.certify_candidates() {
                     let round = proposal.round;
                     let view = round.view();
                     debug!(%view, "attempting certification");
-                    let receiver = self.automaton.certify(round, proposal.payload).await;
-                    let handle = certify_pool.push(async move { (round, receiver.await) });
+                    let result = if leader_is_local {
+                        // Once we know the local participant led this view, reaching out to the
```
patrick-ogrady: cleanup this comment

consensus/src/simplex/mod.rs
```diff
@@ -951,6 +951,141 @@ mod tests {
         all_online::<_, _, RoundRobin>(secp256r1::fixture);
     }
 
+    /// Integration test where a dishonest leader (validator 0) proposes
```
patrick-ogrady: nit: thin out this description

consensus/src/simplex/mod.rs
```diff
@@ -4523,7 +4658,9 @@ mod tests {
                     propose_latency: (10.0, 5.0),
                     verify_latency: (10.0, 5.0),
                     certify_latency: (10.0, 5.0),
-                    should_certify: mocks::application::Certifier::Sometimes,
+                    // This test only exercises reporter filtering. Keep certification
```
patrick-ogrady: nit: not needed?

consensus/src/simplex/mod.rs
```diff
@@ -951,6 +951,141 @@ mod tests {
...
+            for reporter in reporters.iter() {
+                reporter.assert_no_faults();
+                reporter.assert_no_invalid();
+            }
+        });
+    }
+
+    #[test_group("slow")]
+    #[test_traced]
+    fn test_dishonest_leader_certification_rejected() {
```
patrick-ogrady, line 1083: test more modes?

consensus/src/simplex/actors/voter/mod.rs
```diff
@@ -181,7 +181,42 @@ mod tests {
...
+        setup_voter_with_certifier(
+            context,
+            oracle,
+            participants,
+            schemes,
+            elector,
+            leader_timeout,
+            certification_timeout,
+            timeout_retry,
+            mocks::application::Certifier::Always,
```
patrick-ogrady, line 204: It rubs me the wrong way that we're "losing coverage" by using Always here but not sure how best to retain it?

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -584,16 +584,24 @@ impl<E: Clock + CryptoRngCore + Metrics, S: Scheme<D>, L: ElectorConfig<S>, D: D
...
+        let me = self.scheme.me();
         candidates
             .into_iter()
             .filter_map(|view| {
                 if view <= self.last_finalized {
                     return None;
                 }
-                self.views.get_mut(&view)?.try_certify()
+                let round = self.views.get_mut(&view)?;
+                let am_leader = round
```
patrick-ogrady: If leader not known here and ready to certify, we know we aren't the leader/didn't propose?

consensus/src/simplex/actors/voter/mod.rs
patrick-ogrady: I think I'm now ok with this. What we were doing before was just wrong (honest leader would never broadcast a block that fails certification).

patrick-ogrady:
Ended up finding a few paths where we'd vote finalize before ensuring a block was durably persisted. Fixing the issue and adding some regression tests.

consensus/src/marshal/core/actor.rs
```diff
@@ -518,9 +518,13 @@ where
                         };
                         response.send_lossy(info);
                     }
-                    Message::Proposed { round, block } => {
+                    Message::Proposed { round, block, ack } => {
+                        // Persist before acknowledging so the caller can rely on
+                        // "voted ⟹ persisted" before broadcasting their own
```
patrick-ogrady: cleanup these comments

consensus/src/marshal/standard/deferred.rs
```diff
@@ -983,4 +1003,132 @@ mod tests {
...
+    /// re-write and the test cannot catch the pre-fix race.
+    #[test_traced("WARN")]
+    fn test_certify_persists_block_before_resolving() {
+        for seed in 0u64..16 {
+            certify_persists_block_before_resolving_at(seed);
+        }
+    }
+
+    fn certify_persists_block_before_resolving_at(seed: u64) {
+        use commonware_broadcast::Broadcaster;
```
patrick-ogrady: don't import inline

consensus/src/marshal/standard/deferred.rs
```diff
@@ -983,4 +1003,132 @@ mod tests {
...
+    /// which already persists. Otherwise `marshal.verified` is just a no-op
+    /// re-write and the test cannot catch the pre-fix race.
+    #[test_traced("WARN")]
+    fn test_certify_persists_block_before_resolving() {
+        for seed in 0u64..16 {
+            certify_persists_block_before_resolving_at(seed);
+        }
+    }
+
+    fn certify_persists_block_before_resolving_at(seed: u64) {
```
patrick-ogrady: put test above seed usage?

consensus/src/marshal/standard/deferred.rs
```diff
@@ -630,7 +640,13 @@ where
                     height = %block.height(),
                     "requested broadcast of built block"
                 );
-                self.marshal.proposed(round, block).await;
+                if !self.marshal.proposed(round, block).await {
```
patrick-ogrady: This should happen before (and cause a return)?

consensus/src/marshal/coding/marshaled.rs
```diff
@@ -618,6 +627,14 @@ where
                 erasure_timer.observe();
 
                 let commitment = coded_block.commitment();
+
+                // Persist before returning the commitment to consensus. Once
+                // consensus receives it, the proposer will vote, so the block
+                // must be on disk before that point.
+                if !marshal.verified(consensus_context.round, coded_block.clone()).await {
```
patrick-ogrady: This "should" be propose? Then we could remove the value in propose/only use for forward?

really we want this to be `persist` and `propose` to send? But, we don't want to break existing users?

patrick-ogrady: Could alternatively change Simplex to only broadcast a vote after propose returns?

consensus/src/marshal/standard/deferred.rs
```diff
@@ -623,14 +636,15 @@ where
...
+                };
+                if !self.marshal.proposed(round, block).await {
+                    warn!(
+                        ?round,
+                        ?digest,
+                        "marshal unavailable during proposed broadcast; block not persisted"
+                    );
+                    return;
                 }
-                debug!(
```
patrick-ogrady, line 627: don't drop this?

consensus/src/marshal/standard/deferred.rs
```diff
@@ -722,7 +736,7 @@ mod tests {
             // Create parent block at height 1
             let parent = make_raw_block(genesis.digest(), Height::new(1), 100);
             let parent_digest = parent.digest();
-            marshal
+            let _ = marshal
```
patrick-ogrady: assert true?

consensus/src/marshal/coding/shards/mailbox.rs
```diff
@@ -116,9 +116,12 @@ where
...
 
     /// Broadcast a proposed erasure coded block's shards to the participants.
-    pub async fn proposed(&self, round: Round, block: CodedBlock<B, C, H>) {
+    ///
+    /// Returns `true` if the message was enqueued, `false` if the shard engine
+    /// has shut down.
+    pub async fn proposed(&self, round: Round, block: CodedBlock<B, C, H>) -> bool {
         let msg = Message::Proposed { block, round };
-        self.sender.send_lossy(msg).await;
+        self.sender.send_lossy(msg).await
```
patrick-ogrady: can revert this to not return anything

consensus/src/marshal/coding/marshaled.rs
```diff
@@ -183,9 +183,9 @@ where
     scheme_provider: Z,
     epocher: ES,
     strategy: S,
-    last_built: LastBuilt<CodedBlock<B, C, H>>,
     verification_tasks: VerificationTasks<Commitment>,
     cached_genesis: Arc<OnceLock<(Commitment, CodedBlock<B, C, H>)>>,
+    last_built: LastBuilt<CodedBlock<B, C, H>>,
```
patrick-ogrady: put back?

consensus/src/marshal/coding/marshaled.rs
```diff
@@ -618,18 +628,15 @@ where
                 erasure_timer.observe();
 
                 let commitment = coded_block.commitment();
+                let round = consensus_context.round;
+
```
patrick-ogrady: remove \n

consensus/src/marshal/coding/mod.rs
```diff
@@ -1595,4 +1597,259 @@ mod tests {
...
+            let setup = CodingHarness::setup_validator(
+                context.with_label("validator_0"),
+                &mut oracle,
+                me.clone(),
+                ConstantProvider::new(schemes[0].clone()),
+            )
+            .await;
+            let marshal = setup.mailbox;
+            let shards = setup.extra;
+            let actor_handle = setup.actor_handle;
```
patrick-ogrady: nit: "actor of what"? Can we be more specific with field naming here?

consensus/src/marshal/coding/mod.rs
```diff
@@ -1595,4 +1597,259 @@ mod tests {
...
+
+            // Certify - this is the safety gate before finalize voting.
+            let certify_result = marshaled
+                .certify(child_round, child_commitment)
+                .await
+                .await
+                .expect("certify result missing");
+            assert!(certify_result, "certify should succeed");
+
+            // CRITICAL: abort the marshal actor synchronously, with no
```
patrick-ogrady: cleanup this comment....just say what we're doing.

consensus/src/marshal/coding/mod.rs
```diff
@@ -1595,4 +1597,259 @@ mod tests {
...
+            // Drive the full leader-side propose + broadcast path.
+            let commitment = marshaled
+                .propose(propose_context)
+                .await
+                .await
+                .expect("propose should produce a commitment");
+            assert_eq!(commitment, expected_commitment);
+            marshaled.broadcast(commitment, Plan::Propose).await;
+
+            // CRITICAL: abort the marshal actor synchronously so it cannot
```
patrick-ogrady: This is a bit flaky? is the best way to write this test?

consensus/src/marshal/coding/variant.rs
```diff
@@ -100,6 +101,13 @@ where
     }
 
     async fn send(&self, round: Round, block: CodedBlock<B, C, H>, _recipients: Recipients<P>) {
-        self.proposed(round, block).await;
+        let commitment = block.commitment();
+        if !self.proposed(round, block).await {
```
patrick-ogrady: keep as-is

consensus/src/marshal/mocks/verifying.rs
```diff
@@ -14,14 +14,16 @@ use commonware_runtime::deterministic;
...
+/// - Returns the configured block (if any) from `propose()`
 /// - Returns a configurable result from `verify()`
 #[derive(Clone)]
 pub struct MockVerifyingApp<B, S> {
     /// The genesis block to return.
     pub genesis: B,
     /// The result returned by `verify`.
     pub verify_result: bool,
+    /// The block returned by `propose`. If `None`, `propose` returns `None`.
+    pub propose_result: Option<B>,
```
patrick-ogrady, line 29: before verify_result?

consensus/src/marshal/standard/deferred.rs
```diff
@@ -146,8 +146,8 @@ where
     application: A,
     marshal: Mailbox<S, Standard<B>>,
     epocher: ES,
-    last_built: LastBuilt<B>,
     verification_tasks: VerificationTasks<<B as Digestible>::Digest>,
+    last_built: LastBuilt<B>,
```
patrick-ogrady: keep order?

consensus/src/marshal/standard/deferred.rs
```diff
@@ -393,6 +393,7 @@ where
                 build_timer.observe();
 
                 let digest = built_block.digest();
+
```
patrick-ogrady: remove \n

consensus/src/marshal/standard/deferred.rs
```diff
@@ -468,8 +472,6 @@ where
                             task_tx.send_lossy(true);
                             marshaled.verification_tasks.insert(round, digest, task_rx);
                         }
-                        // `Complete` means either immediate rejection or successful
```
patrick-ogrady, line 471: retain this?

consensus/src/marshal/standard/inline.rs
```diff
@@ -141,8 +141,8 @@ where
     application: A,
     marshal: Mailbox<S, Standard<B>>,
     epocher: ES,
-    last_built: LastBuilt<B>,
     available_blocks: AvailableBlocks<B::Digest>,
+    last_built: LastBuilt<B>,
```
patrick-ogrady: move back?

consensus/src/marshal/standard/inline.rs
```diff
@@ -322,10 +321,12 @@ where
                 build_timer.observe();
 
                 let digest = built_block.digest();
+
```
patrick-ogrady: remove \n

consensus/src/marshal/standard/inline.rs
```diff
@@ -696,7 +702,7 @@ mod tests {
             };
             let parent = B::new::<Sha256>(parent_ctx, genesis.digest(), Height::new(1), 100);
             let parent_digest = parent.digest();
-            marshal.clone().proposed(parent_round, parent).await;
+            assert!(marshal.clone().proposed(parent_round, parent).await);
```
patrick-ogrady: why do we need to clone marshal here?

consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -873,14 +887,6 @@ impl<
             },
             on_stopped => {
                 debug!("context shutdown, stopping voter");
-
-                // Sync and drop journal
```
patrick-ogrady, line 876: retain this?

patrick-ogrady: ah, this is done to ensure we call this on exiting early?

consensus/src/simplex/actors/voter/slot.rs
```diff
@@ -73,11 +78,20 @@ where
...
+    /// Records a proposal that has already been verified.
     ///
-    /// If the slot is already populated, we ignore the proposal.
-    pub fn built(&mut self, proposal: Proposal<D>) {
+    /// If the slot already contains the same proposal, we refresh it to
+    /// verified state. Conflicting proposals are ignored.
+    fn verified(&mut self, proposal: Proposal<D>, local: bool) {
+        let verified = Status::Verified(local);
         if let Some(existing) = &self.proposal {
+            if existing == &proposal && self.status != Status::Equivocated {
```
patrick-ogrady: I don't like this?

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -1810,6 +1820,76 @@ mod tests {
...
+            let (added, equivocator) = state.add_notarization(notarization);
+            assert!(added);
+            assert!(equivocator.is_none());
+
+            let candidates = state.certify_candidates();
+            assert_eq!(candidates.len(), 1);
+            assert_eq!(candidates[0].0, proposal);
+            assert!(
+                !candidates[0].1,
+                "leader-owned recovered proposal must not inherit local certification"
```
patrick-ogrady, line 1892: ?

consensus/src/simplex/mocks/application.rs
```diff
@@ -169,9 +166,7 @@ pub struct Config<H: Hasher, P: PublicKey> {
     pub verify_latency: Latency,
     pub certify_latency: Latency,
 
-    /// Predicate to determine whether a payload should be certified.
-    /// Returning true means certify, false means reject.
-    pub should_certify: Certifier<H::Digest>,
+    pub certifier: Certifier<H::Digest>,
```
patrick-ogrady: nit: revert to should_certify?

consensus/src/marshal/standard/deferred.rs
```diff
@@ -299,8 +299,8 @@ where
     ) -> oneshot::Receiver<Self::Digest> {
         let mut marshal = self.marshal.clone();
         let mut application = self.application.clone();
-        let last_built = self.last_built.clone();
         let epocher = self.epocher.clone();
+        let last_built = self.last_built.clone();
```
patrick-ogrady: keep order?

consensus/src/marshal/standard/deferred.rs
```diff
@@ -622,18 +633,22 @@ where
...
+                    return false;
                 }
-                debug!(
-                    round = %round,
-                    digest = %block.digest(),
-                    height = %block.height(),
-                    "requested broadcast of built block"
-                );
-                self.marshal.proposed(round, block).await;
+                debug!(?round, ?digest, "requested broadcast of built block");
```
patrick-ogrady: keep height?

consensus/src/marshal/standard/validation.rs
```diff
@@ -199,8 +205,15 @@ where
...
     };
 
-    if application_valid {
-        marshal.verified(context.round, block).await;
+    if application_valid && !marshal.verified(context.round, block).await {
+        debug!(
+            round = ?context.round,
+            "marshal unavailable during verified ack; aborting verify"
+        );
+        // Persistence not confirmed: caller MUST NOT signal a positive
```
patrick-ogrady: nit: unnecessary?

consensus/src/marshal/standard/deferred.rs
```diff
@@ -983,4 +1006,132 @@ mod tests {
...
+                .await
+                .await
+                .expect("certify result missing");
+            assert!(certify_result, "certify should succeed");
+
+            // CRITICAL: abort the marshal actor synchronously, with no
+            // intervening await. If certify returned true but the actor had
+            // only enqueued (not processed) the `Verified` message, this
+            // abort kills the actor before persistence completes.
+            actor_handle.abort();
```
patrick-ogrady: Could we avoid this complexity. by just stubbing "false" on success? do we need to actually abort this in a flaky way?

consensus/src/marshal/standard/mod.rs
```diff
@@ -1532,6 +1568,178 @@ mod tests {
...
+            view_retention_timeout: ViewDelta::new(10),
+            max_repair: NZUsize!(10),
+            max_pending_acks: NZUsize!(1),
+            block_codec_config: (),
+            partition_prefix: partition_prefix.to_string(),
+            prunable_items_per_section: NZU64!(10),
+            replay_buffer: NZUsize!(1024),
+            key_write_buffer: NZUsize!(1024),
+            value_write_buffer: NZUsize!(1024),
+            page_cache: CacheRef::from_pooler(&context, PAGE_SIZE, PAGE_CACHE_SIZE),
```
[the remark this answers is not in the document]
patrick-ogrady: This will be fixed after #3614

consensus/src/marshal/coding/marshaled.rs
```diff
@@ -266,9 +266,9 @@ where
             scheme_provider,
             strategy,
             epocher,
-            last_built: Arc::new(Mutex::new(None)),
             verification_tasks: VerificationTasks::new(),
             cached_genesis: Arc::new(OnceLock::new()),
+            last_built: Arc::new(Mutex::new(None)),
```
patrick-ogrady: move back

consensus/src/marshal/core/actor.rs
```diff
@@ -1326,25 +1326,36 @@ where
...
+    /// `Inline::certify`) that resolves on this notification is guaranteed to observe
+    /// a block that survives a crash. Reversing the order would let a certify waiter
+    /// emit a finalize vote for a block that is not yet on disk.
     async fn cache_verified(
         &mut self,
         round: Round,
         digest: <V::Block as Digestible>::Digest,
         block: V::Block,
     ) {
+        self.cache
```
patrick-ogrady: I think we could revert this? We don't guarantee what we notify with is persisted?

patrick-ogrady:
There still seems to be a case where Inline::certify() can be resolved before durability? Need to check that (getting the block seems to be resolved by the broadcast buffer optionally).

consensus/src/marshal/core/actor.rs
```diff
@@ -1339,12 +1351,28 @@ where
...
+    async fn persist_block(
+        &mut self,
+        round: Round,
+        digest: <V::Block as Digestible>::Digest,
+        block: V::Block,
+    ) {
+        if self.find_block_in_storage(digest).await.is_some() {
+            return;
+        }
+        self.cache_block(round, digest, block).await;
```
patrick-ogrady: Should go in verified or notarized?

consensus/src/marshal/core/actor.rs
```diff
@@ -1339,12 +1351,28 @@ where
         self.cache.put_block(round, digest, block.into()).await;
     }
 
+    /// Ensure a block is durably present in local storage without asserting it
+    /// belongs in the verified cache.
+    async fn persist_block(
```
patrick-ogrady: call this something else?

consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -1091,5 +1097,13 @@ impl<
                 }
             },
         }
+
+        // Sync and drop the journal
```
patrick-ogrady, line 1110: This actually caught another case where we could've synced more when breaking out of the above loop.

consensus/src/marshal/core/actor.rs
```diff
@@ -518,9 +518,10 @@ where
                         };
                         response.send_lossy(info);
                     }
-                    Message::Proposed { round, block } => {
+                    Message::Proposed { round, block, ack } => {
                         self.cache_verified(round, block.digest(), block.clone())
                             .await;
+                        ack.send_lossy(());
```
patrick-ogrady: Arguably, we should put this after `buffer` in case it panics (akin to storage)?

patrick-ogrady: Given it is generic, I suppose that makes sense?

patrick-ogrady:
If we persist block during propose rather than in broadcast, it looks a LOT cleaner. Remove proposed/last_built, call `verified` and then call `forward` (with recipients) on broadcast from consensus?

patrick-ogrady:
I think there is one more subtle edge case where a leader restarts and persisted a block in marshal but not a vote in simplex. The leader then re-proposes on restart but the new block isn't written (same index is a no-op)?

What we want to do here is load a verified for that round and pass that back as digest (or don't propose at all)?

consensus/src/marshal/mocks/harness.rs
```diff
@@ -312,6 +320,951 @@ pub trait TestHarness: 'static + Sized {
...
+            let mut state = HailstormState {
+                validators: &mut validators,
+                canonical: &mut canonical,
+                parent: &mut parent,
+                parent_commitment: &mut parent_commitment,
+                participants: &participants,
+                schemes: &schemes,
+                propagation_delay,
+            };
+            advance_hailstorm_to(target_height, &mut context, &mut state).await;
```
patrick-ogrady: `hailstorm()` never crashes validators in the window this PR is trying to harden. Each shutdown happens only after `advance_hailstorm_to()` has finalized up to `target_height`, and that helper waits for `mailbox.get_block(height)` plus `get_finalization(height)` on every active validator before returning (`wait_for_validator_height()` at lines 373-395). That means the restart happens after later finalization/backfill has already repaired state, not immediately after `marshal.verified()` / `marshal.certified()` returned. An ack-before-`put_sync` regression could still pass here. Add a deterministic crash hook right after those methods resolve, before any finalization is reported.

consensus/src/marshal/mocks/harness.rs
```diff
@@ -312,6 +320,951 @@ pub trait TestHarness: 'static + Sized {
...
+                        .await;
+                        let restarted = H::setup_validator(
+                            context.with_label(&format!("validator_0_restart_{cycle}")),
+                            &mut oracle,
+                            me.clone(),
+                            provider.clone(),
+                        )
+                        .await;
+                        assert!(
+                        restarted.mailbox.get_block(&digest).await.is_some(),
```
patrick-ogrady: `proposed_success_implies_recoverable_after_restart()` does not assert the recovery path that production code actually uses. On restart, both standard wrappers first call `marshal.get_verified(round)` (`consensus/src/marshal/standard/inline.rs:248` and `consensus/src/marshal/standard/deferred.rs:317`), but this test only checks `mailbox.get_block(&digest)` after recovery here and repeats the same pattern in `verified_success_implies_recoverable_after_restart()` at lines 925-929. A regression that preserved the digest-key lookup but lost or misindexed the round entry would still pass this test while the real restart branch would fall through and rebuild/reverify. Assert `get_verified(round)` directly, or drive a post-restart `propose()`/`verify()` and check that it reuses the recovered block instead of rebuilding it.

consensus/src/marshal/mocks/harness.rs
```diff
@@ -312,6 +320,951 @@ pub trait TestHarness: 'static + Sized {
...
+            2,
+            NUM_VALIDATORS as u16,
+        );
+        let digest_b = H::digest(&block_b);
+        assert_ne!(digest_a, digest_b, "test requires distinct digests");
+
+        let mut peers: [ValidatorHandle<H>; 0] = [];
+        H::verify(&mut handle, round, &block_a, &mut peers).await;
+        assert!(
+            H::certify(&mut handle, round, &block_b).await,
```
patrick-ogrady, line 1370: The only dedicated `certified()` regression never restarts the validator after `H::certify()` returns. This helper proves that verified/certified writes do not collide in-process, but it does not cover the high-priority crash window from Need 1: `Message::Certified` acknowledging before the notarized cache is durably synced. If that ordering regressed, this test would still pass because lines 1125-1136 read the block back from the live mailbox before any restart. Add a `certified_success_implies_recoverable_after_restart()` contract test analogous to the verified-path helpers so the `cache_block`/`put_sync` handshake is checked explicitly.

consensus/src/simplex/actors/voter/mod.rs
patrick-ogrady, line 1: `no_self_certify_after_restart` hardcodes `RoundRobin` even when it is invoked with threshold-VRF fixtures, so it never exercises the certificate-seeded leader restoration path that the new replay logic depends on. That means a future regression in `State::set_leader()` / `Elector::elect()` for `Random` could still pass this test, because the prior certificate's contents are irrelevant under `RoundRobin`. Add a restart case that uses the real certificate-dependent elector and asserts the recovered leader-owned view still avoids `automaton.certify`.

consensus/src/marshal/core/actor.rs
```diff
@@ -471,24 +471,27 @@ where
...
 
                 // Persist buffered processed-height updates once after draining all ready acks.
                 if let Err(e) = self.application_metadata.sync().await {
                     error!(?e, "failed to sync application progress");
                     return;
                 }
 
+                // Inform the buffer of the last acknowledged commitment (anything below this is safe to prune).
+                buffer.finalized(last_acked_commitment).await;
+
```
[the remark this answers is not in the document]
patrick-ogrady: This finalization approach is meant to be a "high-watermark". The old approach was buggy (fwiw) and led us to delete intermediate results we hadn't yet grabbed.

consensus/src/marshal/mocks/harness.rs
```diff
@@ -312,6 +320,1187 @@ pub trait TestHarness: 'static + Sized {
...
+        let v_late = Round::new(Epoch::zero(), View::new(25));
+        let mut peers: [ValidatorHandle<H>; 0] = [];
+        H::verify(&mut handle, v_early, &repeated, &mut peers).await;
+        assert!(
+            H::certify(&mut handle, v_late, &repeated).await,
+            "certify must ack"
+        );
+
+        // Verify `orphan` at V=1 only (no certify).
+        H::verify(&mut handle, v_early, &orphan, &mut peers).await;
```
patrick-ogrady: `certify_at_later_view_survives_earlier_view_pruning` does not actually validate its pruning negative control. `repeated` is already written into `verified_blocks` at `V=1`, so the later `orphan` write at the same `v_early` collides on the same archive index and is dropped before the prune loop ever runs. As a result, the later `get_block(&orphan_digest).is_none()` assertion can pass even if `verified_blocks[V=1]` was never pruned, which weakens the only evidence that the surviving `repeated` block came from the later duplicate-key entry at `V=25`. Give `orphan` a distinct verified view, or assert that it is retrievable before advancing the retention floor.

Regression test: file `consensus/src/marshal/mocks/harness.rs`; command `cargo test -p commonware-consensus certify_at_later_view_survives_earlier_view_pruning -- --nocapture`; observed result: both `marshal::standard::tests::test_standard_certify_at_later_view_survives_earlier_view_pruning` and `marshal::coding::tests::test_coding_certify_at_later_view_survives_earlier_view_pruning` panic at `consensus/src/marshal/mocks/harness.rs:1247` with `negative control assumes orphan is present before pruning, but the second verified write at V=1 is dropped`.

```patch
diff --git a/consensus/src/marshal/mocks/harness.rs b/consensus/src/marshal/mocks/harness.rs
index c213fea..e487837 100644
--- a/consensus/src/marshal/mocks/harness.rs
+++ b/consensus/src/marshal/mocks/harness.rs
@@ -1244,6 +1244,10 @@ pub fn certify_at_later_view_survives_earlier_view_pruning<H: TestHarness>() {
 
         // Verify `orphan` at V=1 only (no certify).
         H::verify(&mut handle, v_early, &orphan, &mut peers).await;
+        assert!(
+            handle.mailbox.get_block(&orphan_digest).await.is_some(),
+            "negative control assumes orphan is present before pruning, but the second verified write at V=1 is dropped"
+        );
 
         // Drive the finalized chain forward to advance `last_processed_round`
         // past V=1's retention boundary but not past V=25's. With
```
</details>

consensus/src/marshal/standard/inline.rs
```diff
@@ -725,4 +756,478 @@ mod tests {
...
+
+            let genesis = make_raw_block(Sha256::hash(b""), Height::zero(), 0);
+            let round = Round::new(Epoch::zero(), View::new(1));
+            let ctx = Ctx {
+                round,
+                leader: me.clone(),
+                parent: (View::zero(), genesis.digest()),
+            };
+            let stale_block = B::new::<Sha256>(ctx.clone(), genesis.digest(), Height::new(1), 100);
+            assert!(marshal.verified(round, stale_block).await);
```
patrick-ogrady: `test_propose_skips_when_verified_block_exists_on_restart` does not actually perform a restart: it seeds the verified cache with `marshal.verified(...)` and then calls `Inline::propose(...)` on the same live mailbox. That means it never exercises archive restore or the new `Message::GetVerified` recovery path that the test comment talks about, so a real restart regression could still slip through. Rewriting this to reinitialize the validator with `start_and_recover`/the existing restart harness would make the coverage match the scenario.

consensus/src/simplex/actors/voter/mod.rs
patrick-ogrady, line 1: The restart coverage here begins only after a local `Notarize` has already been journaled. That misses the more dangerous window this PR is trying to harden: leader `automaton.propose()` returns, marshal has a verified block for the round, and the process crashes before any peer notarizes or the voter journals its own vote. There is still no seeded voter-level test proving the restarted actor nullifies that view instead of issuing a second proposal through the live run loop. Please add an end-to-end restart test that kills the voter immediately after proposal completion, restarts on the same partition, and asserts both `automaton.propose` stays at one call and the round exits via `Vote::Nullify`.

patrick-ogrady, line 1: The new `no_self_certify_*` tests only cover the positive shortcut (`is_local=true`). The negative case from the `is_local=false` fix is still only covered by the state unit test in `consensus/src/simplex/actors/voter/state.rs:1842`, not by the live voter loop. That means there is no end-to-end proof that an elected leader who later sees a conflicting notarization/certificate will actually call `automaton.certify()` instead of taking the local `Either::Left(ready(Ok(true)))` shortcut. Please add an observer-based voter test that injects a conflicting certificate on a leader-owned view and asserts the certify callback fires before finalize.

consensus/src/marshal/coding/marshaled.rs
```diff
@@ -943,36 +986,16 @@ where
...
+        // Coding variant does not support targeted forwarding;
+        // peers reconstruct blocks from erasure-coded shards.
+        //
+        // TODO(#3389): Support checked data forwarding for PhasedScheme.
+        let Plan::Propose { round } = plan else {
+            return;
+        };
+        self.marshal
+            .forward(round, commitment, Recipients::All)
+            .await;
```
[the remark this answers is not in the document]
patrick-ogrady: This isn't a valid finding? How could it "lose data"?

patrick-ogrady:
bugbot run

consensus/src/marshal/standard/inline.rs
```diff
@@ -241,6 +239,28 @@ where
...
+                    .get_verified(consensus_context.round)
+                    .await
+                    .is_some()
+                {
+                    debug!(
+                        round = ?consensus_context.round,
+                        "skipping proposal: verified block already exists for round on restart"
+                    );
+                    return;
+                }
```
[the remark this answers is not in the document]
patrick-ogrady: Inline blocks are not certifiable, so we can't check whether or not they were built with the same context.

patrick-ogrady:
bugbot run

patrick-ogrady (approved):

PR #3546, opened 2026-04-07
[runtime/iobuf/pool] striped atomic bitmap freelist

andresilva:
This PR replaces the buffer pool's global `ArrayQueue` with a striped bitmap freelist. The new freelist tracks stable `u32` slot ids in cache-line-padded atomic bitmap words, so threads can usually claim and return buffers through independent stripes instead of contending on one queue. The implementation keeps a cheap single-buffer path, but its main advantage is the batch API used by TLS refill and spill.

The batch API is the important optimization: `put_batch` coalesces returned slots by bitmap word so many logical returns become one `fetch_or` per touched stripe, and `take_batch` can claim multiple bits from one word with one `fetch_and`. That matches the buffer pool's common TLS pattern: when a thread-local cache misses, claim one buffer for the allocation and refill the cache with the rest. When it overflows, spill a batch back to the global freelist.

In the new freelist microbenchmarks, this is substantially faster than the old `ArrayQueue` shape under the batched TLS workload, especially with multiple threads. The queue baseline still has to `push`/`pop` entries one at a time, while the bitmap freelist collapses those operations into a small number of atomics and avoids moving buffers through temporary global queue nodes.

The pool now carries stable slot ids through pooled buffers so returned buffers can be parked back into their exact freelist slots. Automatic TLS sizing also now allows zero capacity for very small per-class budgets, preventing returned buffers from being stranded in another thread's local cache when the class is already at capacity.

### Benchmarks

| threads | batch | pattern | freelist | ArrayQueue | Mutex<Vec> | vs ArrayQueue | vs Mutex<Vec> |
|---:|---:|---|---:|---:|---:|---:|---:|
| 1 | 1 | single | 16.9 ns | 16.0 ns | 33.3 ns | 0.9x | 2.0x |
| 1 | 4 | single | 38.3 ns | 54.1 ns | 36.3 ns | 1.4x | 0.9x |
| 1 | 8 | single | 66.1 ns | 104.5 ns | 41.2 ns | 1.6x | 0.6x |
| 8 | 1 | lockstep | 62.5 ns | 717.1 ns | 1960.0 ns | 11.5x | 31.4x |
| 8 | 4 | lockstep | 211.5 ns | 2490.0 ns | 2000.0 ns | 11.8x | 9.5x |
| 8 | 8 | lockstep | 338.0 ns | 3680.0 ns | 2170.0 ns | 10.9x | 6.4x |
| 8 | 32 | lockstep | 791.6 ns | 9010.0 ns | 3160.0 ns | 11.4x | 4.0x |
| 8 | 1 | staggered | 124.4 ns | 642.6 ns | 2310.0 ns | 5.2x | 18.6x |
| 8 | 4 | staggered | 225.7 ns | 2170.0 ns | 2500.0 ns | 9.6x | 11.1x |
| 8 | 32 | staggered | 755.9 ns | 8960.0 ns | 3770.0 ns | 11.8x | 5.0x |

These freelist benchmark results use 512 slots and compare the striped bitmap freelist against `ArrayQueue` and `Mutex<Vec<_>>` baselines while repeatedly taking and returning buffers. The single-thread cases capture uncontended overhead: for batch size 1, `Freelist` and `ArrayQueue` are effectively at parity, while `Mutex<Vec<_>>` remains competitive for purely local larger batches. Once batching starts, the `Freelist` pulls ahead of `ArrayQueue`. The intended workload is the contended global freelist used by `BufferPool` refill/spill paths, and there the striped freelist is substantially faster: on the 512-slot, 8-thread cases it is roughly 5.2-11.8x faster than `ArrayQueue` and 4.0-31.4x faster than `Mutex<Vec<_>>`, with the strongest wins in the refill/spill shapes used when TLS caching is enabled.


runtime/src/iobuf/freelist.rs
```diff
@@ -0,0 +1,1058 @@
+//! Striped global freelist for one buffer-pool size class.
+//!
+//! A [`Freelist`] owns the tracked buffers from one [`super::pool::BufferPool`]
+//! size class that are not checked out and not held in a thread-local cache,
+//! making those buffers available for reuse by any thread in the pool. Each
+//! tracked buffer has a stable slot id within its size class.
```
patrick-ogrady, line 6: ah, interesting

patrick-ogrady (approved):

PR #3549, opened 2026-04-07
[runtime/tokio] support for core pinned dedicated tasks

andresilva:
This PR adds a `pinned(core)` builder method to the `Spawner` trait that runs tasks on a dedicated thread pinned to a specific CPU core. It implies `dedicated()`, since you can't be pinned without being dedicated: `context.dedicated().spawn(f)` for an unpinned dedicated thread, `context.pinned(3).spawn(f)` for a pinned one. A separate method was preferred over `dedicated(Option<usize>)` to avoid the awkward `dedicated(None)` for the common unpinned case. Users who want to co-locate two dedicated threads on the same core simply pass the same value.

Core pinning uses `sched_setaffinity` on Linux and is a no-op on other platforms, macOS only offers affinity hints (not true pinning), and Windows is currently not a target for production environments. A public `commonware_runtime::available_cores()` helper is also exposed for users who want to query the CPU count for their own assignment strategies.

Related #3537.


runtime/src/utils/thread.rs
```diff
@@ -111,3 +111,53 @@ where
...
+pub fn available_cores() -> Option<usize> {
+    None
+}
+
+/// Best-effort attempt to pin the current thread to the given core.
+/// The `core` value wraps around the number of available CPUs.
+#[cfg(target_os = "linux")]
+pub(crate) fn pin_to_core(core: usize) {
+    let Some(num_cores) = available_cores() else {
+        tracing::warn!("failed to query CPU count, skipping core pinning");
```
pepyakin: Idt it's a good idea to warn on this path. Prefer ONCE?

runtime/src/utils/thread.rs
```diff
@@ -111,3 +111,53 @@ where
...
+/// Returns the number of online CPUs, or `None` if it cannot be determined.
+///
+/// Always returns `None` on non-Linux platforms.
+#[cfg(not(target_os = "linux"))]
+pub fn available_cores() -> Option<usize> {
+    None
+}
+
+/// Best-effort attempt to pin the current thread to the given core.
+/// The `core` value wraps around the number of available CPUs.
```
andresilva: Undecided whether wrapping is a good idea or not. Do we want to panic if the user tries to do `pinned(16)` on a 16-core CPU? Or just silently ignore it and don't pin at all? The latter doesn't seem like a good idea.

runtime/src/utils/thread.rs
andresilva: Addressed.

runtime/src/utils/thread.rs
patrick-ogrady: in this case, I think we could panic (clearly this is an advanced API and I think we should be transparent)

andresilva: I updated this to panic on an invalid core number, but still silently fail if the actual syscalls to get the number of cores and to pin fail (e.g. on a restricted environment like a cgroup with CPU limits). Let me know if you think we should still panic there.

runtime/src/deterministic.rs
```diff
@@ -1134,7 +1134,12 @@ impl Context {
 
 impl crate::Spawner for Context {
     fn dedicated(mut self) -> Self {
-        self.execution = Execution::Dedicated;
+        self.execution = Execution::Dedicated(None);
+        self
+    }
+
+    fn pinned(mut self, core: usize) -> Self {
+        self.execution = Execution::Dedicated(Some(core));
```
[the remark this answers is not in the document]
andresilva: IMO there's not much to do here, the deterministic runtime is single-threaded, it ignores both `pinned()` and `dedicated()`.

runtime/src/utils/thread.rs
```diff
@@ -111,3 +113,94 @@ where
...
+    static WARN_CPUS: Once = Once::new();
+    static WARN_AFFINITY: Once = Once::new();
+
+    let Some(num_cores) = available_cores() else {
+        WARN_CPUS.call_once(|| {
+            tracing::warn!("failed to query CPU count, skipping core pinning");
+        });
+        return;
+    };
+    assert!(
```
patrick-ogrady: Do we need to assert in two locations?

runtime/src/utils/thread.rs
```diff
@@ -111,3 +113,94 @@ where
...
+/// # Panics
+///
+/// Panics if `core` is greater than or equal to the number of available CPUs.
+#[cfg(target_os = "linux")]
+pub(crate) fn pin_to_core(core: usize) {
+    static WARN_CPUS: Once = Once::new();
+    static WARN_AFFINITY: Once = Once::new();
+
+    let Some(num_cores) = available_cores() else {
+        WARN_CPUS.call_once(|| {
```
patrick-ogrady: Is the idea that warning each time would become very noisy/preference would just be to continue?

runtime/src/deterministic.rs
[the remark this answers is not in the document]
patrick-ogrady: Agreed

runtime/src/utils/thread.rs
```diff
@@ -111,3 +113,94 @@ where
...
+/// Pins the current thread to the given core.
+///
+/// If the CPU count cannot be queried or `sched_setaffinity` fails, a warning
+/// is logged once and the thread continues unpinned.
+///
+/// # Panics
+///
+/// Panics if `core` is greater than or equal to the number of available CPUs.
+#[cfg(target_os = "linux")]
+pub(crate) fn pin_to_core(core: usize) {
```
patrick-ogrady: If you are designing your application with core pinning in mind, I wonder if there is some credence to just panicking if core pining fails?

For example, I suspect glommio just dies if it can't pin?

andresilva: I think in a lot of cases core pinning is mostly a performance optimization but not necessarily something that MUST happen, i.e. a best-effort attempt, since it doesn't incur any correctness issue if it doesn't happen. Maybe here we can just return a `bool` and that let's the user decide what to do? The issue is that in `Context::pined(...)` we have no way to expose this 🫤. I'm thinking mostly of a scenario where someone distributes a node binary, that tries to pin threads as a performance optimization, and then it suddenly panics because someone is running that (general) binary in a container.

runtime/src/utils/thread.rs
andresilva: If we don't assert in `Context::pinned(...)` then the failure only happens at spawn time. That may be fine, but may also be a bit weird? Also the panic will happen in the new spawned thread, and won't necessarily take down the main thread (we can workaround that though if you think panicking at spawn time is better).

runtime/src/utils/thread.rs
andresilva: Yeah, if this fails once, it will very likely just keep failing forever (i.e. the syscall just isn't allowed for some reason). I can remove and just warn every time though.

runtime/src/utils/thread.rs
patrick-ogrady: My concern (maybe too pedantic) is that correct/smooth behavior does actually rely on pinning in the applications that want it (or else there is such a big performance hit it doesn't work well). 🤔

andresilva: Maybe the behavior could be controlled by a runtime configuration option? IMO it should default to "best-effort", but can be set to strict in which case we panic if pinning fails.

runtime/src/tokio/runtime.rs
```diff
@@ -893,6 +894,61 @@ impl crate::BufferPooler for Context {
...
+                None => unreachable!(),
+            }
+        }
+        Err(_) => {
+            // The dedicated thread exited before reporting startup status.
+            handle.abort();
+            panic!("failed to start dedicated task");
+        }
+    }
+}
```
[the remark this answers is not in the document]
andresilva: If `cpu` is `None` we might need to reset the affinity mask to the default (if the runtime supports pinning), so we still need to make sure that doesn't fail. We could avoid this if `cpu == None` and `available_cpus == None` but IMO it's not worth the code complexity.

runtime/src/utils/thread.rs
```diff
@@ -100,14 +108,197 @@ const fn system_thread_stack_size_impl() -> Option<usize> {
...
-        .expect("failed to spawn thread")
+    try_spawn(stack_size, f).expect("failed to spawn thread")
+}
+
+/// Returns the logical CPU ids currently allowed for the calling thread.
+///
+/// This queries the calling thread's affinity mask via `sched_getaffinity` and
+/// returns `None` if that query fails.
+#[cfg(target_os = "linux")]
+pub(crate) fn available_cpus() -> Option<NonEmptyVec<usize>> {
```
patrick-ogrady: I wonder if we should hide this complexity and layout cores as a contiguous 0,1,2,3 and then map to physical cores?

I've never done pinning before, so that probably is more confusing than its worth (and we should let the application make this call)?

I don't love how it complicates the spawner trait but it isn't "wrong"?

andresilva: I'd say we shouldn't hide it for now, another thing I want to do in the future is to also have an helper to expose CPU topology, so that you can pin on the same "core" / "CCD" / "NUMA-node" etc. (and it might just replace this helper altogether). If you're going to pin threads you probably care about this as well, e.g. cross-CCD cacheline invalidation is like 4x slower depending on the CPU.

andresilva: > I don't love how it complicates the spawner trait but it isn't "wrong"?

Also it's not obvious to me how making this linear would change anything in the spawner trait, seems ortoghonal? The reason I decided to expose `Spawner::available_cpus` is to avoid the footgun of using this helper from inside tasks, I can revert that if you want, we only need to cache the baseline available CPUs in tokio runtime so that we can restore the affinity mask on `dedicated()`, but it doesn't need to be exposed.

```rust
let cpus = utils::thread::available_cpus(); // vec![0, 1, 2, 3];
context.pinned(1).spawn(|_| {
  assert!(utils::thread::available_cpus() == vec![1]); // true
  assert!(ctx.available_cpus() == vec![0, 1, 2, 3]); // true
})
```

andresilva: Removed `Spawner::available_cpus`, and `utils::thread::available_cpus` should always return the baseline value now (since it checks affinity of process PID).

clabby:

Issue #2343, opened 2025-11-27
[workspace/perf] Explore usecases for `smallvec`


clabby:
## Overview

There are many places throughout the codebase where we could take advantage of using [`smallvec`](https://docs.rs/smallvec/latest/smallvec/) rather than always-allocating. We should explore areas where we use `Vec`s in the hotpath of `consensus`, `cryptography`, `p2p`, and `storage` that have a relatively small number of elements on average and replace those instances with `SmallVec`s.

PR #3550, opened 2026-04-07
[utils] add `ArrayVec`

andresilva:
This PR adds `ArrayVec<T, N>`, a fixed-capacity inline vector that stores up to `N` elements without heap-allocating the backing storage. This is useful for small, bounded buffers where the maximum length is known at compile time. Capacity is capped at `u16::MAX` to prevent accidentally large inline values. The type uses unsafe partial-initialization internally (to avoid a `T: Default` bound) but exposes a fully safe public API.

The implementation covers the core `Vec`-like surface: `push`, `pop`, `insert`, `remove`, `swap_remove`, `resize`, `truncate`, `extend`, iteration, and conversions to/from `Vec`, slices, and arrays. An `array_vec!` macro provides convenient construction with inferred capacity.


utils/src/vec/array.rs
clabby, line 1: Is there a reason you opted to have our own implementation over using the [`arrayvec`](https://docs.rs/arrayvec/latest/arrayvec) crate?

It also may be better to just use [`smallvec`](https://docs.rs/smallvec/latest/smallvec/) for this usecase, curious your thoughts. While it can spill over on to the heap unlike `arrayvec` / your impl, it should be roughly as efficient for the fixed size case while offering us a bit more flexibility to fit with often-small-but-dynamic cases as well.

clabby:
related #2343

utils/src/vec/array.rs
andresilva: I don't really have an argument for not using `arrayvec`, in fact looking at it now the implementation is quite similar (although with a smaller API). I guess the only argument is that it's one less dependency to include, but I think `arrayvec` is probably already in our dependency tree (transitively). Will defer to @patrick-ogrady on this, am fine with just using `arrayvec` directly or just re-exporting it here.

I am worried that `smallvec`, or generally automatically spilling over to the heap, can mask pathological cases where you "thought" something was small but it actually isn't. `smallvec` is definitely more practical to use though.

andresilva: > more flexibility to fit with often-small-but-dynamic cases as well

I guess this makes sense if it's just used as an optimization, where the size bound is more like a hint.

clabby: Admittedly I haven't benchmarked `smallvec` vs. `arrayvec` when they're both in-bounds of the stack allocated array, though if they perform similarly, `smallvec` does provide an API we could use to create a wrapper type that panics when a spill-over is attempted.

That said, both are already in the monorepo's dependency tree. We could just re-export both crates' types?

andresilva:
Thinking a bit more about this and after talking to @clabby, I think we should probably just re-export `SmallVec`:
- it's already on our dependency tree
- it's probably going to be more useful to us in practice than `ArrayVec`
- performance difference is probably negligible compared to `ArrayVec`, there's a branch before every operation to check whether we're inline or heap, but the branch predictor probably reduces that cost to ~0
- I'm not sure there's a lot of places where we can actually enforce strict bounds at compile time, for example in p2p, even stuff like max number of connections will be configured dynamically at runtime in most cases
- if at some point we actually have use cases for `ArrayVec` we can just re-export it as well (it's also already in our dependency tree!). I guess `cryptography` would probably be the place where we can actually use `ArrayVec`.

clabby:

PR #3552, opened 2026-04-07
[Storage] Acquire journal reader once per merkleization

danlaine:
## Summary

QMDB batch merkleization used `try_join_all` to fan out one future per operation location. Each future independently acquired the journal reader lock and went through the full async poll/wake cycle -- even for operations already in memory. The `FuturesUnordered` scheduling overhead adds significant overhead.

This PR reduces that overhead by:

- **Acquiring the journal reader once per merkleization** instead of once per location. Both the unordered and ordered `merkleize_with_floor_scan` paths now call `db.log.reader().await` once and thread the reader through to all read sites.

- **Replacing `try_join_all` fan-outs with sequential loops.** `read_op` checks in-memory sources (current batch ops, ancestor chain) synchronously first, and only awaits for committed-journal reads. `read_ops` loops over a batch of locations through this single method. Since all reads target the same journal file, there was no real I/O parallelism to lose.

- **Adding a test** (`read_ops_resolves_committed_ancestor_and_current_sources`) that exercises all three operation sources: committed DB (disk), ancestor batch (in-memory chain), and current batch ops.

Credit to @clabby for this optimization in https://gist.github.com/clabby/51993b7c74ea0d3614922fb8c5b496e0

### DIFF
--- a/broadcast/fuzz/fuzz_targets/broadcast_engine_operations.rs
+++ b/broadcast/fuzz/fuzz_targets/broadcast_engine_operations.rs
@@ -13,6 +13,7 @@ use commonware_cryptography::{
 };
 use commonware_p2p::{simulated::Network, Recipients};
 use commonware_runtime::{deterministic, Buf, BufMut, Clock, Metrics, Quota, Runner};
+use commonware_utils::NZUsize;
 use libfuzzer_sys::fuzz_target;
 use rand::{seq::SliceRandom, SeedableRng};
 use std::{collections::BTreeMap, num::NonZeroU32, time::Duration};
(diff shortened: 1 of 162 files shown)


roberto-bayardo:
I think we should go with the approach that allows for batched reads (as in my demo PR from yesterday).  This means instead of raising one step at a time we raise in bulk.

roberto-bayardo:
see: https://github.com/commonwarexyz/monorepo/pull/3539 ?

(...only remove the try_join_all part)

danlaine:
Closing in favor of #3576

PR #3558, opened 2026-04-08
[consensus/marshal] Add restart benchmarks

viktorcrypt:
## Summary

- add a `marshal` benchmark target for restart recovery
- benchmark both `standard` and `coding` marshal variants
- seed finalized blocks, finalizations, and application metadata directly in storage
- measure archive reopen plus `Actor::init` recovery time

## Detail

This adds focused restart benchmarks for marshal recovery.

Each benchmark prepopulates persisted marshal state, then repeatedly reopens the finalized archives and initializes the marshal actor again. This measures the restore path directly instead of bringing up the full network or application stack.

Both `standard` and `coding` variants are covered so restart regressions in either recovery path become visible in benchmarks.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1217,6 +1217,7 @@ dependencies = [
  "commonware-runtime",
  "commonware-storage",
  "commonware-utils",
+ "criterion",
  "futures",
  "pin-project",
  "prometheus-client",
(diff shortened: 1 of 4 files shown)


clabby:

PR #3571, opened 2026-04-10
[Storage] Sync cache resolution + concurrent disk reads in   merkleization 

danlaine:
Combines #3552 and #3539 changes.

## Summary

Optimizes operation reads during QMDB batch merkleization by separating
in-memory cache resolution from disk I/O and batching the disk reads.

Previously, `read_op` was a single async method that independently acquired
the journal reader lock on every disk read, and floor raising issued one
read at a time via `advance_floor_once`. This PR restructures reads around
two primitives:

- **`try_read_cached`** (sync): Resolves operations from in-memory sources
  (current batch ops, ancestor chain) with no async overhead. Returns `None`
  for committed-journal locations that require disk I/O.

- **`read_ops`** (async, batched): Resolves a batch of locations by calling
  `try_read_cached` for each, then issuing a single `try_join_all` for
  the remaining disk misses. The journal reader is acquired once per
  merkleization and threaded through all read sites.

### Key-resolution phase

Both the unordered and ordered `merkleize_with_floor_scan` paths now call
`db.log.reader().await` once and pass the reader to `read_ops` and `finish`.
This replaces the previous pattern where each `read_op` call independently
acquired the reader lock.

### Floor-raise phase

Replaces the sequential `advance_floor_once` loop with a batched approach:

1. Collect up to `MAX_CONCURRENT_READS` (64) candidate locations per batch.
2. Resolve in-memory candidates synchronously via `try_read_cached`.
3. Issue remaining disk reads concurrently via `try_join_all`.
4. Process results in order, moving active ops to the tip.

This reuses the reader already acquired during the key-resolution phase.

### DIFF
--- a/storage/src/qmdb/any/batch.rs
+++ b/storage/src/qmdb/any/batch.rs
@@ -341,7 +341,8 @@ where
     H: Hasher,
     Operation<F, U>: Codec,
 {
-    /// Read an operation at a given location from the correct source.
+    /// Try to resolve an operation from in-memory sources (this batch or the ancestor chain).
+    /// Returns `None` for locations in the committed journal that require disk I/O.
     ///
     /// The operation space is divided into three contiguous regions:
     ///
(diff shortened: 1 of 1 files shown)


danlaine:
Closing in favor of #3576

PR #3575, opened 2026-04-10
[storage] Batch floor raise w/ cached reads

roberto-bayardo:
## Summary                                                                                                                                                                           
                                         
  Adds a synchronous page-cache fast path to batch merkleization reads, avoiding async overhead when data is already cached.                                                           
                                                                                                                                                                                       
  During `merkleize`, the batch reads existing operations from the journal to generate mutation ops and raise the inactivity floor. Previously, all committed-journal reads went through the async `reader.read()` path, creating futures even when the data was sitting in the page cache. This adds a `try_read_cached` method that attempts a fully synchronous read through the page cache before falling back to async I/O.                                                                                                                        
                                                            
  ### Changes                                                                                                                                                                          
  
  - **`runtime/buffer/paged/append.rs`**: Add `Append::try_read_cached` — sync page-cache-only blob read.                                                                              
  - **`journal/segmented/fixed.rs`**: Add `SegmentedJournal::try_get_cached` — sync cache-only item decode.
  - **`journal/contiguous/fixed.rs`**: Add `Inner::try_read_cached` and implement `Reader::try_read_cached`.                                                                           
  - **`journal/contiguous/mod.rs`**: Add `try_read_cached` to the `Reader` trait with a default `None`.                                                                                
  - **`qmdb/any/batch.rs`**: `read_ops` and the floor-raise loop now check batch/ancestor memory, then the journal page cache, before building `try_join_all` for remaining misses.    
                                                                                                                                                                                       
  ### Result                                                                                                                                                                           
                                                                                                                                                                                       
  When the working set fits in the page cache (the common case for merkleize benchmarks), all reads resolve synchronously — no futures created, no async state machines, no yield points. Cache misses still run concurrently via `try_join_all`.

### DIFF
--- a/runtime/src/utils/buffer/paged/append.rs
+++ b/runtime/src/utils/buffer/paged/append.rs
@@ -394,6 +394,22 @@ impl<B: Blob> Append<B> {
         buffer.size()
     }
 
+    /// Try to read exactly `len` bytes starting at `offset` from the page cache only.
+    ///
+    /// Returns `Some(buf)` if all bytes were in the cache, `None` on any cache miss.
+    /// Does not perform any async I/O. The caller must have already validated that
+    /// `offset + len` is within the blob's logical size.
+    pub fn try_read_cached(&self, offset: u64, len: usize) -> Option<IoBufs> {
+        // SAFETY: if all bytes are cached, read_cached fills all `len` bytes.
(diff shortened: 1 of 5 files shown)


roberto-bayardo:
see https://github.com/commonwarexyz/monorepo/pull/3576 for a more complete version rebased on main

PR #3576, opened 2026-04-10
[storage/qmdb] Batch reads in floor raising with page-cache fast path (Rebased on main)

roberto-bayardo:
## Summary                                                                                                                                                                           
                                         
  Adds a synchronous page-cache fast path to batch merkleization reads, avoiding async overhead when data is already cached.                                                           
                                                                                                                                                                                       
  During `merkleize`, the batch reads existing operations from the journal to generate mutation ops and raise the inactivity floor. Previously, all committed-journal reads went through the async `reader.read()` path, creating futures even when the data was sitting in the page cache. This adds a `try_read_sync` method that attempts a fully synchronous read through the page cache before falling back to async I/O.                                                                                                                        
                                                            
  ### Changes                                                                                                                                                                          
  
  - **`runtime/buffer/paged/append.rs`**: Add `Append::try_read_sync` — sync page-cache-only blob read.                                                                              
  - **`journal/segmented/fixed.rs`**: Add `SegmentedJournal::try_get_sync` — sync cache-only item decode.
  - **`journal/contiguous/variable+fixed.rs`**: Add `Inner::try_read_sync` and implement `Reader::try_read_cached`.                                                                           
  - **`journal/contiguous/mod.rs`**: Add `try_read_sync` to the `Reader` trait with a default `None`.                                                                                
  - **`qmdb/any/batch.rs`**: `read_ops` and the floor-raise loop now check batch/ancestor memory, then the journal page cache, before building `try_join_all` for remaining misses.    

 ### Future Work

 Each cached read results in a vector allocation plus copy, and in the variable record case, may result in a vector being allocated then not used should the result not be cached.  Future work (e.g. https://github.com/commonwarexyz/monorepo/pull/3185) may allow us to remove or reduce this overhead which should yield significant additional performance gains.
                                                                                                                                                                                       
  ### Result                                                                                                                                                                           
                                                                                                                                                                                       
  When the working set fits in the page cache (the common case for merkleize benchmarks), all reads resolve synchronously — no futures created, no async state machines, no yield points. Cache misses still run concurrently via `try_join_all`.
  
  Performance on merkleize benchmark is ~20+% faster on fixed, 30% faster on variable.
  ```
qmdb_merkleize::merkleize/variant=any::unordered::fixed::mmr num_keys=1000000
                        time:   [119.02 ms 119.29 ms 119.59 ms]
                        change: [−21.912% −21.447% −21.004%] (p = 0.00 < 0.05)
                        Performance has improved.

Benchmarking qmdb_merkleize::merkleize/variant=any::unordered::variable::mmr num_keys=1000000: Warming up for 3.0000 s
Warning: Unable to complete 10 samples in 5.0s. You may wish to increase target time to 11.3s.
qmdb_merkleize::merkleize/variant=any::unordered::variable::mmr num_keys=1000000
                        time:   [131.78 ms 133.61 ms 135.87 ms]
                        change: [−34.049% −33.047% −31.871%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 10 measurements (20.00%)
  1 (10.00%) high mild
  1 (10.00%) high severe

Benchmarking qmdb_merkleize::merkleize/variant=any::unordered::fixed::mmb num_keys=1000000: Warming up for 3.0000 s
Warning: Unable to complete 10 samples in 5.0s. You may wish to increase target time to 6.0s.
qmdb_merkleize::merkleize/variant=any::unordered::fixed::mmb num_keys=1000000
                        time:   [117.36 ms 119.21 ms 120.96 ms]
                        change: [−22.366% −21.199% −19.961%] (p = 0.00 < 0.05)
                        Performance has improved.

Benchmarking qmdb_merkleize::merkleize/variant=any::unordered::variable::mmb num_keys=1000000: Warming up for 3.0000 s
Warning: Unable to complete 10 samples in 5.0s. You may wish to increase target time to 10.8s.
qmdb_merkleize::merkleize/variant=any::unordered::variable::mmb num_keys=1000000
                        time:   [128.42 ms 130.63 ms 133.63 ms]
                        change: [−35.023% −33.817% −32.360%] (p = 0.00 < 0.05)
                        Performance has improved.
```

### DIFF
--- a/runtime/src/utils/buffer/paged/append.rs
+++ b/runtime/src/utils/buffer/paged/append.rs
@@ -403,12 +403,11 @@ impl<B: Blob> Append<B> {
         Some(buffer.size())
     }
 
-    /// Try to read from the page cache only into the provided buffer.
+    /// Read into `buf` if it can be done synchronously (e.g. without I/O), returning `false` otherwise.
     ///
-    /// Returns `true` if all `buf.len()` bytes were in the cache, `false` on any
-    /// cache miss. Does not perform any async I/O. The caller must have already
-    /// validated that `offset + buf.len()` is within the blob's logical size.
-    pub fn try_read_cached(&self, offset: u64, buf: &mut [u8]) -> bool {
+    /// Returns `true` only if all `buf.len()` bytes were satisfied. The caller must have
+    /// already validated that `offset + buf.len()` is within the blob's logical size.
+    pub fn try_read_sync(&self, offset: u64, buf: &mut [u8]) -> bool {
         self.cache_ref.read_cached(self.id, buf, offset) == buf.len()
     }
 
--- a/runtime/src/utils/buffer/paged/append.rs
+++ b/runtime/src/utils/buffer/paged/append.rs
@@ -394,6 +394,24 @@ impl<B: Blob> Append<B> {
         buffer.size()
     }
 
+    /// Returns the logical size of the blob if it can be observed without waiting.
+    ///
+    /// This is useful for opportunistic fast paths that should fall back rather than contend with
+    /// concurrent writers.
+    pub fn try_size(&self) -> Option<u64> {
+        let buffer = self.buffer.try_read().ok()?;
+        Some(buffer.size())
+    }
+
+    /// Try to read from the page cache only into the provided buffer.
+    ///
+    /// Returns `true` if all `buf.len()` bytes were in the cache, `false` on any
+    /// cache miss. Does not perform any async I/O. The caller must have already
+    /// validated that `offset + buf.len()` is within the blob's logical size.
+    pub fn try_read_cached(&self, offset: u64, buf: &mut [u8]) -> bool {
+        self.cache_ref.read_cached(self.id, buf, offset) == buf.len()
+    }
+
     /// Read exactly `len` immutable bytes starting at `offset`.
     pub async fn read_at(&self, offset: u64, len: usize) -> Result<IoBufs, Error> {
         // Read into a temporary contiguous buffer and copy back to preserve structure.
--- a/storage/src/journal/contiguous/mod.rs
+++ b/storage/src/journal/contiguous/mod.rs
@@ -35,10 +35,10 @@ pub trait Reader: Send + Sync {
     /// Guaranteed not to return [Error::ItemPruned] for positions within `bounds()`.
     fn read(&self, position: u64) -> impl Future<Output = Result<Self::Item, Error>> + Send;
 
-    /// Try to read an item from the page cache only, without async I/O.
+    /// Try to read an item from cache (without async I/O).
     ///
-    /// Returns `Some(item)` if all data was in cache, `None` on any miss.
-    /// The default returns `None` (no cache available).
+    /// Returns `Some(item)` if all data was in cache, `None` on any miss. The
+    /// default returns `None` (no cache available).
     fn try_read_cached(&self, _position: u64) -> Option<Self::Item> {
         None
     }
--- a/storage/src/journal/contiguous/mod.rs
+++ b/storage/src/journal/contiguous/mod.rs
@@ -35,6 +35,14 @@ pub trait Reader: Send + Sync {
     /// Guaranteed not to return [Error::ItemPruned] for positions within `bounds()`.
     fn read(&self, position: u64) -> impl Future<Output = Result<Self::Item, Error>> + Send;
 
+    /// Try to read an item from the page cache only, without async I/O.
+    ///
+    /// Returns `Some(item)` if all data was in cache, `None` on any miss.
+    /// The default returns `None` (no cache available).
+    fn try_read_cached(&self, _position: u64) -> Option<Self::Item> {
+        None
+    }
+
     /// Return a stream of all items starting from `start_pos`.
     ///
     /// Because the reader holds the lock, validation and stream setup happen
--- a/storage/src/journal/contiguous/mod.rs
+++ b/storage/src/journal/contiguous/mod.rs
@@ -35,11 +35,10 @@ pub trait Reader: Send + Sync {
     /// Guaranteed not to return [Error::ItemPruned] for positions within `bounds()`.
     fn read(&self, position: u64) -> impl Future<Output = Result<Self::Item, Error>> + Send;
 
-    /// Try to read an item from cache (without async I/O).
+    /// Read an item if it can be done synchronously (e.g. without I/O), returning `None` otherwise.
     ///
-    /// Returns `Some(item)` if all data was in cache, `None` on any miss. The
-    /// default returns `None` (no cache available).
-    fn try_read_cached(&self, _position: u64) -> Option<Self::Item> {
+    /// Default implementation always returns `None`.
+    fn try_read_sync(&self, _position: u64) -> Option<Self::Item> {
         None
     }
 
--- a/storage/src/journal/segmented/fixed.rs
+++ b/storage/src/journal/segmented/fixed.rs
@@ -181,9 +181,8 @@ impl<E: Storage + Metrics, A: CodecFixedShared> Journal<E, A> {
         A::decode(buf.coalesce()).map_err(Error::Codec)
     }
 
-    /// Try to read an item using only the page cache. Returns `None` on any cache miss
-    /// or if the position cannot be verified as in-bounds without blocking.
-    pub fn try_get_cached(&self, section: u64, position: u64) -> Option<A> {
+    /// Get an item if it can be done synchronously (e.g. without I/O), returning `None` otherwise.
+    pub fn try_get_sync(&self, section: u64, position: u64) -> Option<A> {
         let blob = self.manager.get(section).ok()??;
         let offset = position.checked_mul(Self::CHUNK_SIZE_U64)?;
         let remaining = blob.try_size()?.checked_sub(offset)?;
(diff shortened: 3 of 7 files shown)

### CODE REVIEW

File: storage/src/qmdb/benches/merkleize.rs [hunk 1]
```diff
@@ -141,7 +141,7 @@ type CurUVar256Mmb = commonware_storage::qmdb::current::unordered::variable::Db<
 
 // -- Config --
 
-const ITEMS_PER_BLOB: NonZeroU64 = NZU64!(100_000);
+const ITEMS_PER_BLOB: NonZeroU64 = NZU64!(10_000_000);
 const THREADS: NonZeroUsize = NZUsize!(8);
 const PAGE_SIZE: NonZeroU16 = NZU16!(4096);
```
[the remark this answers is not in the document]
roberto-bayardo: thanks didn't mean to include this here. reverted

storage/src/qmdb/any/batch.rs
```diff
@@ -484,74 +536,26 @@ where
...
+    async fn finish<E, C, I, S, R>(
         self,
         mut ops: Vec<Operation<F, U>>,
         mut diff: BTreeMap<U::Key, DiffEntry<F, U::Value>>,
         active_keys_delta: isize,
         user_steps: u64,
         metadata: Option<U::Value>,
         mut scan: S,
+        reader: &R,
         db: &Db<F, E, C, I, H, U>,
```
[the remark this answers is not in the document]
roberto-bayardo: good point!  fixed by taking ownership of the reader and dropping it before the CPU only work.

storage/src/journal/segmented/variable.rs
```diff
@@ -580,6 +580,43 @@ impl<E: Storage + Metrics, V: CodecShared> Journal<E, V> {
...
+        if varint_len + data_len <= header.remaining() {
+            let data = header.slice(varint_len..varint_len + data_len);
+            return decode_item::<V>(data, &self.codec_config, self.compression.is_some()).ok();
+        }
+
+        // Otherwise try reading the full item from cache.
+        let full_buf = blob.try_read_cached(offset, varint_len + data_len)?;
+        let full = full_buf.coalesce();
+        let data = full.slice(varint_len..varint_len + data_len);
+        decode_item::<V>(data, &self.codec_config, self.compression.is_some()).ok()
```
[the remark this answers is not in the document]
roberto-bayardo: if length prefix is corrupt we have bigger problems.

it is annoying we waste a buffer allocation, but it's not a huge concern -- and a user should not use the cached read path if items are going to be oversized.

roberto-bayardo:
bugbot run

runtime/src/utils/buffer/paged/append.rs
```diff
@@ -394,6 +394,24 @@ impl<B: Blob> Append<B> {
...
+        let buffer = self.buffer.try_read().ok()?;
+        Some(buffer.size())
+    }
+
+    /// Try to read from the page cache only into the provided buffer.
+    ///
+    /// Returns `true` if all `buf.len()` bytes were in the cache, `false` on any
+    /// cache miss. Does not perform any async I/O. The caller must have already
+    /// validated that `offset + buf.len()` is within the blob's logical size.
+    pub fn try_read_cached(&self, offset: u64, buf: &mut [u8]) -> bool {
```
andresilva: With https://github.com/commonwarexyz/monorepo/pull/3185 we'll be able to avoid this copy, I think this is the best we can do for now though.

roberto-bayardo: Thanks for the confirmation.

storage/src/journal/segmented/fixed.rs
```diff
@@ -181,6 +181,22 @@ impl<E: Storage + Metrics, A: CodecFixedShared> Journal<E, A> {
...
+    /// Try to read an item using only the page cache. Returns `None` on any cache miss
+    /// or if the position cannot be verified as in-bounds without blocking.
+    pub fn try_get_cached(&self, section: u64, position: u64) -> Option<A> {
+        let blob = self.manager.get(section).ok()??;
+        let offset = position.checked_mul(Self::CHUNK_SIZE_U64)?;
+        let remaining = blob.try_size()?.checked_sub(offset)?;
+        if remaining < Self::CHUNK_SIZE_U64 {
+            return None;
+        }
+        let mut buf = vec![0u8; Self::CHUNK_SIZE];
```
andresilva, line 192: I guess we could use the buffer pool here, but might not be worth it if chunk size is small.

storage/src/journal/segmented/variable.rs
```diff
@@ -580,6 +580,63 @@ impl<E: Storage + Metrics, V: CodecShared> Journal<E, V> {
...
+            return decode_item::<V>(
+                &header[varint_len..varint_len + data_len],
+                &self.codec_config,
+                self.compression.is_some(),
+            )
+            .ok();
+        }
+
+        // Otherwise try reading the full item from cache.
+        let mut full = vec![0u8; item_len];
```
andresilva, line 627: Could use buffer pool here rather allocate vec.

storage/src/journal/segmented/fixed.rs
roberto-bayardo: this causes a performance regression in the merkleize benchmark.

storage/src/journal/segmented/variable.rs
roberto-bayardo: I'll try this again without the smaller fixed vector using it to see if it helps.  My original implementation used buffer pool for everything but this was significantly faster.

storage/src/journal/contiguous/mod.rs
```diff
@@ -35,6 +35,14 @@ pub trait Reader: Send + Sync {
     /// Guaranteed not to return [Error::ItemPruned] for positions within `bounds()`.
     fn read(&self, position: u64) -> impl Future<Output = Result<Self::Item, Error>> + Send;
 
+    /// Try to read an item from the page cache only, without async I/O.
```
danlaine: Nit: I feel like talking about a page cache in the trait definition is an implementation detail leak?

```suggestion
    /// Try to read an item from the journal without async I/O.
```

storage/src/qmdb/any/batch.rs
```diff
@@ -347,37 +372,64 @@ where
...
-        } else {
-            // Base DB's journal (on-disk async read).
-            let reader = db.log.reader().await;
-            Ok(reader.read(loc_val).await?)
+        batch_ops: &[Operation<F, U>],
+        reader: &R,
+    ) -> Result<Operation<F, U>, crate::qmdb::Error<F>> {
+        match self.try_read_cached(loc, batch_ops) {
+            Some(op) => Ok(op),
+            None => Ok(reader.read(*loc).await?),
```
danlaine, line 411: `read_ops` does `reader.try_read_cached` if `self.try_read_cached` is `None` -- should we do the same here?

storage/src/qmdb/any/batch.rs
```diff
@@ -560,24 +564,91 @@ where
...
+                    };
+                    candidates.push(candidate);
+                    scan_from = Location::new(*candidate + 1);
+                }
+                if candidates.is_empty() {
                     break;
                 }
+
+                // Resolve cached candidates synchronously: batch/ancestor then page cache.
+                let cached: Vec<Option<Operation<F, U>>> = candidates
```
danlaine: Can we just call `read_ops` here? I think this section is the same thing but inlined

storage/src/journal/segmented/variable.rs
roberto-bayardo: Re-ran the merkleize benchmark with this alloc being done by the buffer pool and it is indeed significantly (5+%) slower on the variable-length qmdb variants.  That said, the benchmark is only testing small (32-byte) values that aren't actually variable length, but it still exercises this codepath.

roberto-bayardo: Perhaps its worth having a conditional here that uses the buffer pool when the items are a certain size?

storage/src/journal/contiguous/mod.rs
roberto-bayardo: yeah. the method is named "try_read_cached", do you think we should rename the method? Or instead replace "page cache" with "cache"?

danlaine: What would you think of renaming to `try_read_sync`?

storage/src/qmdb/any/batch.rs
roberto-bayardo: This code got a little confusing so I restructured it a bit and tried to improve naming.  I think it resolves this concern. PTAL?

storage/src/qmdb/any/batch.rs
roberto-bayardo: yup fixed

storage/src/journal/contiguous/mod.rs
roberto-bayardo: yes, was doing this independently! done.

danlaine (dismissed):

danlaine (approved):

storage/src/journal/segmented/variable.rs
andresilva: The buffer pool already does this internally, there's a minimum size threshold that bypasses the pool completely and goes straight to allocation. The pool is doing aligned allocation and I think that might be slightly slower than the bare allocation rust does for vec.

PR #3583, opened 2026-04-13
[runtime] Review Attributes

### LINKED ISSUE #3485: [runtime] Review Usage of Attributes on Metrics
In marshaled applications, we apply attributes aggressively. We should be sure this doesn't blow up memory usage (either by tracked tasks or something unexpected).

patrick-ogrady:
Closes: #3485
Related: #1110 + #3077 
Future Work: https://github.com/commonwarexyz/monorepo/issues/3585

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -332,6 +332,25 @@ stability_scope!(BETA {
         /// _Using attributes does not reduce cardinality (N epochs still means N time series).
         /// Attributes just make metrics easier to query, filter, and aggregate._
         ///
+        /// # Bounded Attribute Values
+        ///
+        /// Each distinct attribute-value combination passed to `register` creates its own
+        /// sub-registry that lives for the lifetime of the root registry (or of the enclosing
+        /// [`Metrics::with_scope`], if any). Only use `with_attribute` with values drawn from
+        /// a bounded set (e.g. validator index, a finite number of epochs held in a cache).
+        /// For unbounded dimensions (e.g. consensus round, request id), either:
+        ///
+        /// - register a [`Family`](prometheus_client::metrics::family::Family) once and pass
+        ///   the dynamic value at observation time, or
+        /// - pair `with_attribute` with [`Metrics::with_scope`] so the sub-registry is dropped
+        ///   when the scope ends.
+        ///
+        /// Applying `with_attribute` without subsequently calling `register` is always safe:
+        /// it only mutates a `Vec<(String, String)>` on the returned context. Task-tracking
+        /// metrics (e.g. `runtime_tasks_spawned`) are keyed by the task's label name, not by
+        /// context attributes, so spawning with a per-round attribute does not inflate their
+        /// cardinality.
+        ///
         /// # Family Label Conflicts
         ///
         /// When using `Family` metrics, avoid using attribute keys that match the Family's label field names.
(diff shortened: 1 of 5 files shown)


runtime/src/lib.rs
```diff
@@ -288,6 +285,38 @@ stability_scope!(BETA {
...
+    ///   is not inherited by child contexts.
+    ///
+    /// | Builder                 | Registered metric name | Registered metric labels | Registry isolation | Runtime task metrics | Tracing span                     |
+    /// | ----------------------- | :--------------------: | :----------------------: | :----------------: | :------------------: | :------------------------------: |
+    /// | `with_label`            | prefix                 | -                        | -                  | `name`               | `name` field when `with_span`    |
+    /// | `with_attribute`        | -                      | label dimension          | -                  | -                    | OTel attribute when `with_span`  |
+    /// | `with_scope`            | -                      | -                        | sub-registry       | -                    | -                                |
+    /// | `with_span`             | -                      | -                        | -                  | -                    | enables span creation            |
     pub trait Metrics: Clone + Send + Sync + 'static {
```
patrick-ogrady, line 320: At some point, it may make sense to rename this `Observability` (if we keep metrics/tracing coupled).

runtime/src/lib.rs
```diff
@@ -435,6 +449,35 @@ stability_scope!(BETA {
...
+        ///
+        /// The flag is consumed by the next [`Spawner::spawn`] call on the returned
+        /// context (and on any clone of it). It is not inherited by child contexts
+        /// produced via [`Metrics::with_label`], [`Metrics::with_attribute`],
+        /// [`Metrics::with_scope`], or any of the `Spawner` builders: each such builder
+        /// resets the flag so the caller must opt in again for the next spawn.
+        ///
+        /// Enabling the span only affects tracing; it does not change which metrics
+        /// are registered, nor does it widen the cardinality of runtime task metrics.
+        fn with_span(&self) -> Self;
```
clabby, line 465: way better than `instrumented` :)

runtime/src/lib.rs
```diff
@@ -288,6 +285,38 @@ stability_scope!(BETA {
...
 
     /// Interface to register and encode metrics.
+    ///
+    /// # Mental Model: Metrics vs Tracing
+    ///
+    /// A context carries multiple pieces of observability state, each
+    /// applied via a different builder. They compose freely, but they do
+    /// not all feed into the same sinks:
+    ///
+    /// - `name` (set by [`Metrics::with_label`]): prefix applied to metrics
```
patrick-ogrady, line 295: I don't love we need to document this so tightly (it is pretty opaque) but it does enable us to do what we need in many different environments, so I can't complain too much.

runtime/src/lib.rs
```diff
@@ -288,6 +285,38 @@ stability_scope!(BETA {
...
+    ///   does not affect runtime task metrics and has no effect on tracing.
+    /// - `span` (set by [`Metrics::with_span`]): wraps the next spawned task
+    ///   in a `tracing` span populated from the current `name` and `attributes`.
+    ///   It never touches metrics. The flag is consumed by the next `spawn` and
+    ///   is not inherited by child contexts.
+    ///
+    /// | Builder                 | Registered metric name | Registered metric labels | Registry isolation | Runtime task metrics | Tracing span                     |
+    /// | ----------------------- | :--------------------: | :----------------------: | :----------------: | :------------------: | :------------------------------: |
+    /// | `with_label`            | prefix                 | -                        | -                  | `name`               | `name` field when `with_span`    |
+    /// | `with_attribute`        | -                      | label dimension          | -                  | -                    | OTel attribute when `with_span`  |
```
clabby, line 317: While you're in here, we should also add the attributes to the [`ValueSet`](https://docs.rs/tracing/latest/tracing/field/struct.ValueSet.html) of the `Span`.

See [`Span::record`](https://docs.rs/tracing/latest/tracing/struct.Span.html#method.record)

patrick-ogrady: This will also get bubbled into the trace we send to OTel (according to claude):

```
⏺ Yes — good catch. OTel backends actually see both: set_attribute adds the individual pairs directly to
  OtelData, and tracing_opentelemetry's layer also bridges the attributes tracing field (set via record) into
  the same OTel span as a string attribute. So on the OTel side you get:

  - round="7", epoch="2" — from set_attribute.
  - attributes="[(\"epoch\", \"2\"), (\"round\", \"7\")]" — from record via the tracing→OTel bridge.
```

Not sure there is a way around this?

clabby (approved):

PR #3584, opened 2026-04-13
[consensus/simplex] Self-Verify Nits

patrick-ogrady:
Related: #3551 

  Where the new tests are clearly better (from Claude):

  1. They actually catch the regression. The original 4-case matrix tested "no verify" for leader-owned views, but verify_ready's leader short-circuit structurally prevents verify from firing there regardless of the fix. All 4
  original cases passed whether the fix was present or not — they were coincidental coverage, not regression tests.
  2. Coverage of the live path. The original only tested restart; test_no_self_verify_when_proposing covers the live leader flow that the original ignored entirely.
  3. Coverage of the follower path. The original only tested leader-owned views (where verify can't fire). test_no_self_verify_after_restart exercises the follower path where verify does fire in the buggy case and the observer
  directly catches it.
  4. Deterministic sync via Nullify. The original used ad-hoc sleeps and matches; new tests use the voter's natural leader_timeout → Nullify path as a clean sync point that guarantees the voter has run its full cycle.

### DIFF
--- a/consensus/src/simplex/actors/voter/mod.rs
+++ b/consensus/src/simplex/actors/voter/mod.rs
@@ -4147,7 +4147,6 @@ mod tests {
         no_self_propose_or_verify_after_restart(secp256r1::fixture);
     }
 
-
     /// After restart, a proposal we already voted on must not be re-verified
     /// when it is re-delivered to the voter (e.g. via the automaton after
     /// peer vote aggregation reconstructs it).
--- a/consensus/src/simplex/actors/voter/mod.rs
+++ b/consensus/src/simplex/actors/voter/mod.rs
@@ -3728,6 +3728,424 @@ mod tests {
         no_recertification_after_replay::<_, _, RoundRobin>(secp256r1::fixture);
     }
 
+    /// When the voter is the leader of a view and builds its own proposal, it
+    /// must not subsequently ask the automaton to verify that same proposal.
+    ///
+    /// This is guarded by `Slot::built` (which sets `status = Verified` and
+    /// `requested_verify = true`) and by `Round::verify_ready` short-circuiting
+    /// for leader-owned views. This test asserts the end-to-end invariant on
+    /// the live path (no restart): after calling `automaton.propose`, the voter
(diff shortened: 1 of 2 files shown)


consensus/src/simplex/actors/voter/mod.rs
```diff
@@ -3850,38 +3819,292 @@ mod tests {
...
+                }
+            }
 
+            let proposed = propose_calls.lock();
+            let verified = verify_calls.lock();
+            assert!(
+                !verified.contains(&target_view),
+                "voter must not request verification for a previously-voted view after restart (observed: {verified:?})"
+            );
+            assert!(
```
patrick-ogrady, line 4135: put first?

consensus/src/simplex/actors/voter/mod.rs
```diff
@@ -3729,47 +3728,270 @@ mod tests {
...
-        Finalization,
-    }
+    /// After restart in a leader-owned view that was already voted on, the
+    /// voter must not call `automaton.propose` to build a new payload.
+    ///
+    /// Replay of the journaled local notarize must restore the slot's proposal
+    /// (and `requested_build`) so that `should_build` returns false. Without
+    /// that restoration, the voter's main loop would treat the leader-owned
+    /// view as a fresh proposal opportunity and call `automaton.propose` for
+    /// a payload that has already been notarized.
```
patrick-ogrady: notarized is the wrong word here

consensus/src/simplex/actors/voter/mod.rs
```diff
@@ -3729,47 +3728,270 @@ mod tests {
...
+                        response.send(None).unwrap();
+                        assert_eq!(current, target_view);
+                        assert_eq!(leader, Participant::new(0));
+                        break;
+                    }
+                    batcher::Message::Constructed(_) => {}
+                }
+            }
+
+            // Without the replay fix, the voter's main loop would call `automaton.propose`
```
patrick-ogrady: there is no "fix"

PR #3587, opened 2026-04-13
[Storage] Implement state sync for `qmdb::keyless`

danlaine:
## Summary

Add state sync support for `qmdb::keyless`, matching the existing sync implementations for `qmdb::immutable`, `qmdb::any`, and `qmdb::current`.

## Storage crate

**`sync::Database` impl** (`storage/src/qmdb/keyless/sync.rs`): A single generic impl on `Keyless<mmr::Family, E, V, C, H>` covers both fixed and variable keyless databases. `from_sync_result` initializes the Merkle structure from sync components, builds the authenticated journal, and verifies the last operation is a commit. Simpler than `immutable`'s impl because there is no key-value index to rebuild.

**`pinned_nodes_at`** (`storage/src/qmdb/keyless/mod.rs`): New method on `Keyless` that returns the Merkle nodes needed to reconstruct the tree from a pruning boundary. Required by the resolver. Follows the same pattern as `any::Db::pinned_nodes_at` and `immutable::Immutable::pinned_nodes_at`.

**`sync::Config`** (`storage/src/qmdb/sync/database.rs`): Impl for `keyless::Config<J>` so the sync engine can extract the journal config.

**`sync::Resolver`** (`storage/src/qmdb/sync/resolver.rs`): New `impl_resolver_keyless!` macro generating resolver impls for both fixed and variable keyless DB types (3 Arc/Lock wrapper variants each). Keyless needs its own macro because its type parameters differ from the keyed variants (no `K`/`T`).

**Tests**: 20 tests covering basic sync at various batch sizes, persistence after sync, target updates (increase, decrease, during sync, after completion), partial/exact match with existing data, resolver failure, and a fixed-variant test.

## Sync example

**`examples/sync/src/databases/keyless.rs`**: New module implementing the `Syncable` trait for keyless using `keyless::fixed` (since the example's `Value` type is fixed-size `sha256::Digest`). Includes `create_config`, deterministic test operation generation, and the `Syncable` impl.

**Server/client**: Added `run_keyless` functions and `--db keyless` CLI support in both binaries.

**`Syncable` trait bound**: Weakened `type Operation` bound from `Operation<Self::Family> + Encode + Sync + 'static` to `Encode + Sync + 'static`. The `Operation<F>` trait requires key-related methods (`key()`, `is_update()`, `is_delete()`, `has_floor()`) that don't apply to keyless operations and were never called through the `Syncable` interface.

Partially resolves #3538

### DIFF
--- a/examples/sync/src/databases/keyless.rs
+++ b/examples/sync/src/databases/keyless.rs
@@ -1,0 +1,185 @@
+//! Keyless database types and helpers for the sync example.
+//!
+//! A `keyless` database is append-only: operations are stored by location rather than by key.
+//! It supports `Append(value)` and `Commit(metadata)` operations. For sync, the engine targets
+//! the Merkle root over all operations, and the client reconstructs the same state by replaying
+//! the fetched operations.
+
+use crate::{Hasher, Value};
+use commonware_cryptography::{Hasher as CryptoHasher, Sha256};
+use commonware_runtime::{buffer, BufferPooler, Clock, Metrics, Storage};
+use commonware_storage::{
+    journal::contiguous::fixed::Config as FConfig,
+    merkle::{
+        journaled::Config as MmrConfig,
+        mmr::{self, Location, Proof},
+    },
+    qmdb::{
+        self,
+        keyless::{self, fixed},
+        operation::Committable,
+    },
+};
+use commonware_utils::{NZUsize, NZU16, NZU64};
+use std::{future::Future, num::NonZeroU64};
+use tracing::error;
+
+/// Database type alias.
+pub type Database<E> = fixed::Db<mmr::Family, E, Value, Hasher>;
+
+/// Operation type alias.
+pub type Operation = fixed::Operation<Value>;
+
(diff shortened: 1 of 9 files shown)


examples/sync/src/databases/keyless.rs
```diff
@@ -0,0 +1,184 @@
...
+        // Use the pruning boundary, same as immutable.
+        self.bounds().await.start
+    }
+
+    fn historical_proof(
+        &self,
+        op_count: Location,
+        start_loc: Location,
+        max_ops: NonZeroU64,
+    ) -> impl Future<Output = Result<(Proof<Key>, Vec<Self::Operation>), qmdb::Error<mmr::Family>>> + Send
```
roberto-bayardo: Why can't you just say "async" here instead of returning a future?  (Same for pinned_nodes_at).

danlaine: Changed to `async`

roberto-bayardo (approved):

PR #3588, opened 2026-04-13
[Storage] Add inactivity floor to `qmdb::immutable` Commit operation

danlaine:
## Summary

`qmdb::immutable` is an append-only authenticated database where each key is written exactly once. Applications use it for structures like transaction logs -- data that grows monotonically over the lifetime of the system.

The problem is that without a notion of an inactivity floor, the entire history must be retained and synced. In a network of replicas, a new node joining must fetch every operation from genesis to construct the current state. As the log grows, this becomes impractical. The database itself cannot determine which operations are "still needed" because, unlike `qmdb::any` where key overwrites make old operations provably dead, every immutable operation is alive by definition.

Only the application knows when old operations become irrelevant. For example, a transaction log might only need the last N blocks of history. This PR lets the application declare that knowledge by specifying an inactivity floor at commit time.

The floor is embedded directly in the `Operation::Commit` variant, making it part of the operation log and root digest. Every replica processing the same sequence of operations arrives at the same floor. This means:

- Replicas agree on the floor without out-of-band coordination
- The floor can be used as the lower bound of a sync target, so new nodes only fetch operations from the floor onward
- Data below the floor can be pruned

### Operation enum

`Operation` gains a `Family` generic parameter and the `Commit` variant now carries a `Location<F>` alongside its optional metadata:

```
// Before
enum Operation<K, V> {
    Set(K, V::Value),
    Commit(Option<V::Value>),
}

// After
enum Operation<F: Family, K, V> {
    Set(K, V::Value),
    Commit(Option<V::Value>, Location<F>),
}
```

The `has_floor()` method on the `Operation` trait now returns `Some(loc)` for `Commit` instead of `None`. Both the fixed-size codec (big-endian `u64`) and variable-size codec (varint) are updated to encode the new field.

### Immutable struct

A new `inactivity_floor_loc: Location<F>` field is added to the `Immutable` struct. It is maintained through every state transition:

- **`init_from_journal`**: Reads the floor from the last commit operation. The in-memory snapshot is replayed starting from the floor rather than from the journal's pruning boundary, so keys set before the floor are excluded from the snapshot.
- **`apply_batch`**: Updates the floor from the batch. A monotonicity assertion (checked before any journal mutation) ensures the floor never decreases.
- **`rewind`**: Restores the floor from the target commit operation.
- **`from_sync_result`**: Extracts the floor from the last commit in the synced range and uses it as the snapshot replay start.
- **`prune`**: Now checks against `inactivity_floor_loc` instead of `last_commit_loc`. The application can only prune up to what it has declared inactive.

A new `inactivity_floor_loc()` accessor is provided.

### Batch API

`merkleize()` takes a new `inactivity_floor: Location<F>` parameter. The application passes this when building a batch to declare which operations are no longer needed. The floor is embedded in the resulting `Commit` operation and stored on `MerkleizedBatch` for use by `apply_batch`.

Partially resolves https://github.com/commonwarexyz/monorepo/issues/3538


patrick-ogrady:
As discussed, should we add this to keyless too?

roberto-bayardo (dismissed):

roberto-bayardo (dismissed):

roberto-bayardo (approved):

PR #3596, opened 2026-04-14
[runtime] storage benchmarking harness

andresilva:
This PR adds a standalone benchmark binary for profiling the runtime's storage backend under controlled workloads. The binary exercises whichever backend is selected at compile time (tokio by default, io_uring with `--features iouring-storage`) across six workload shapes: sequential and random reads, sequential and random overwrites, monotonic appends, and mixed append-plus-read pressure. Each workload supports configurable I/O size, worker concurrency, cache preparation (warm/cold), write payload shape (contiguous/vectored), and sync cadence.

The harness is structured around generic timed loops (`run_read_loop`, `run_write_loop`) that accept caller-provided block selection closures (`sequential_blocks`, `random_blocks`). This keeps the hot-path code shared across all workloads while letting each workload define its own access pattern. Latency is sampled at a constant stride to avoid front-biasing the distribution without adding measurable overhead to hot-cache runs. Block selection uses `SmallRng` for random workloads, which is fast enough to not perturb measurements.

Results are collected into a `Report` that computes per-operation throughput (ops/s, MiB/s) and latency percentiles (p50/p95/p99), with output in either human-readable or JSON format for downstream tooling.


runtime/src/tokio/runtime.rs
```diff
@@ -351,13 +369,16 @@ impl crate::Runner for Runner {
 
         // Initialize runtime
         let metrics = Arc::new(Metrics::init(runtime_registry));
-        let runtime = Builder::new_multi_thread()
+        let mut builder = Builder::new_multi_thread();
+        builder
```
patrick-ogrady, line 373: I think this is correct but feels clearer to overwrite `let mut builder` each time we modify builder?

runtime/src/storage/benches/main.rs
```diff
@@ -0,0 +1,42 @@
...
+    workload::run_benchmark,
+};
+use commonware_runtime::{tokio, Runner as _};
+
+fn main() -> Result<(), String> {
+    let cfg = Config::parse();
+
+    let mut runtime_cfg = tokio::Config::default()
+        .with_worker_threads(cfg.worker_threads)
+        .with_storage_directory(cfg.root.clone());
```
patrick-ogrady, line 32: - [P2] Isolate each benchmark run in its own storage directory — /Users/patrickogrady/code/monorepo-pr-3596/runtime/src/storage/
    benches/main.rs:26-28
    The --root help says this is a parent directory, but we hand it straight to with_storage_directory() and every run uses fixed
    PARTITION/BLOB_NAME values. If two benchmarks are launched against the same root (including the default /tmp), they will open the
    same blob; the Blob contract treats concurrent opens of the same name as undefined, so parallel comparisons can corrupt each
    other's measurements or remove the other's data during cleanup.

runtime/src/storage/benches/filesystem.rs
```diff
@@ -0,0 +1,146 @@
...
+    root: &Path,
+    partition: &str,
+    name: &[u8],
+    file_size: u64,
+) -> Result<S::Blob, String> {
+    let (blob, _) = storage.open(partition, name).await.str_err()?;
+    blob.resize(file_size).await.str_err()?;
+    blob.sync().await.str_err()?;
+    if file_size > 0 {
+        preallocate_blob(root, partition, name)
```
patrick-ogrady: Should we do this before open instead of after? Feels weird to sync_all outside of runtime?

runtime/src/storage/benches/filesystem.rs
```diff
@@ -0,0 +1,146 @@
...
+
+/// Create a fixed-size, preallocated blob. Returns the open blob handle.
+pub async fn prepare_blob<S: Storage>(
+    storage: &S,
+    root: &Path,
+    partition: &str,
+    name: &[u8],
+    file_size: u64,
+) -> Result<S::Blob, String> {
+    let (blob, _) = storage.open(partition, name).await.str_err()?;
```
patrick-ogrady: Is this `str_err` what we want to do? Seems like we could map this at the highest level rather than pushing strings everywhere?

Is the idea that we don't need to define an err type?

runtime/src/storage/benches/config.rs
```diff
@@ -0,0 +1,340 @@
...
+    #[arg(long, default_value_t = 30, value_parser = value_parser!(u64).range(1..))]
+    duration: u64,
+
+    /// Read or write size in bytes. Accepts suffixes like 64K, 4M, or 1G.
+    #[arg(long, default_value = "4096", value_parser = parse_byte_size_usize)]
+    pub io_size: usize,
+
+    /// Parallel worker count for steady-state workloads.
+    #[arg(long, default_value_t = 1, value_parser = value_parser!(usize))]
+    pub inflight: usize,
```
patrick-ogrady, line 138: We typically issue all of these on a single task rather than spawning a task per inflight. I think that's more typical for how we use in prod but doesn't exercise as much the relationship between this and the worker_threads config? Maybe that's fine?

runtime/src/storage/benches/workload.rs
```diff
@@ -0,0 +1,332 @@
...
+                        cfg.io_size,
+                        sequential_blocks(worker as u64 % total_blocks, inflight, total_blocks),
+                    )
+                    .await
+                } else {
+                    run_read_loop(
+                        blob,
+                        deadline,
+                        cfg.io_size,
+                        random_blocks(cfg.seed + worker as u64, total_blocks),
```
patrick-ogrady: wrapping add for seed + worker?

patrick-ogrady (commented):
pretty close!

runtime/src/storage/benches/main.rs
andresilva: This was handled in my python script wrapper to keep the logic minimal here, but probably makes sense to keep it safe.

runtime/src/storage/benches/filesystem.rs
andresilva: Guess I was just lazy because this is just a benchmark 😅

runtime/src/storage/benches/filesystem.rs
andresilva: If we do it before then we need to replicate the runtime logic of creating partition folder and also (potentially) dealing with the versioning header.

runtime/src/storage/benches/config.rs
andresilva: All benchmarks also issue all inflight requests from a single task (there isn't a single `spawn` in the harness), workers are just separate futures that are joined with `FuturesUnordered`. Your point about worker threads is relevant, I think we are effectively only using one worker thread at a time (not necessarily the same one).

andresilva: I think worker threads could be relevant for io_uring. When io_uring completes a request it needs to wake-up the future that issued the request (through the oneshot channel), since this future is being awaken from a non-tokio worker thread the task will be added to the global task queue of tokio's executor, in theory any worker thread could then take on the request, and if the task queue is big enough you could get some parallelization. In practice I don't think it will make any difference though.

runtime/src/tokio/runtime.rs
andresilva: Tokio's builder mutators take `&mut self` and return `&mut Builder`, so we can't keep overwriting a single owned `let mut builder` binding on each step, the only alternative is to shadow:

```rust
let mut builder = Builder::new_multi_thread();
let builder = builder.worker_threads(self.cfg.worker_threads);
let builder = builder.max_blocking_threads(self.cfg.max_blocking_threads);
let builder = builder.thread_stack_size(self.cfg.thread_stack_size);
let builder = builder.enable_all();
let builder = if let Some(global_queue_interval) = self.cfg.global_queue_interval {
    builder.global_queue_interval(global_queue_interval)
} else {
    builder
};
```

This feels worse than what we have now, but let me know if you prefer this.

runtime/src/storage/benches/filesystem.rs
```diff
@@ -0,0 +1,190 @@
...
+use commonware_utils::hex;
+use rand::Rng;
+use std::{
+    fs, io,
+    path::{Path, PathBuf},
+    process,
+    time::{SystemTime, UNIX_EPOCH},
+};
+#[cfg(target_os = "linux")]
+use std::{fs::OpenOptions, os::fd::AsRawFd};
```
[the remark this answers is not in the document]
patrick-ogrady: This code is easier to reason about with the current form.

patrick-ogrady (approved):

patrick-ogrady:
bugbot run

PR #3597, opened 2026-04-14
[storage/qmdb] Fix Benchmarking Setup

patrick-ogrady:


storage/src/qmdb/benches/merkleize.rs
```diff
@@ -579,52 +579,35 @@ variants! {
...
-    }
-}
-
-fn bench_chained_merkleize(c: &mut Criterion) {
-    let runner = tokio::Runner::new(Config::default());
-    for num_keys in [10_000u64, 100_000, 1_000_000] {
-        for &variant in VARIANTS {
-            c.bench_function(
-                &format!(
-                    "{}/chained variant={} num_keys={num_keys}",
```
patrick-ogrady, line 609: This is an invalid format. I'm not sure how it got past CI.

storage/src/qmdb/benches/merkleize.rs
```diff
@@ -579,52 +579,35 @@ variants! {
...
+                );
+            }
         }
     }
 }
 
 criterion_group! {
     name = benches;
     config = Criterion::default().sample_size(10);
-    targets = bench_merkleize, bench_chained_merkleize
```
roberto-bayardo, line 629: this is intentional to keep from having to wait through init bench setup (slooooow) in order to run merkleized bench.

patrick-ogrady:
Turns out our benchmark tests just straight skip QMDB (both to run/check naming -> explains how this slipped through CI with invalid labeling)

Working on fixing this right now.

storage/src/qmdb/benches/merkleize.rs
patrick-ogrady: This throws off the UI on the benchmarks page, so we'll need to find a different workaround:

image

storage/src/qmdb/benches/bench.rs
```diff
@@ -1,9 +1,10 @@
...
 
 use criterion::criterion_main;
 
 mod common;
 mod generate;
 mod init;
+mod merkleize;
 
-criterion_main!(generate::benches, init::benches);
+criterion_main!(generate::benches, init::benches, merkleize::benches);
```
roberto-bayardo, line 10: Other workaround: I think you can just move init::benches to the end to make sure it is always last and we don't have to sit through the long init time if we just want to run the others.  Maybe put a comment that explains the non-alphabetical order.

storage/src/qmdb/benches/merkleize.rs
roberto-bayardo: suggested another workaround above ^^

storage/src/qmdb/benches/merkleize.rs
roberto-bayardo: sometimes I trust claude too much

.github/scripts/lint_benchmark_names.py
```diff
@@ -98,6 +103,12 @@ def validate_benchmark_name(name: str) -> list[str]:
     """Validate a single benchmark name and return any violations."""
     violations = []
 
+    if re.search(r"\.\.\.\s*(#\d+)?$", name):
```
patrick-ogrady, line 106: criterion has a hardcoded max length of 100 characters...we now check for this as well:

Image

storage/src/qmdb/benches/bench.rs
patrick-ogrady: Can you not just do `cargo bench --bench qmdb -- merkleize`? AFAIK that skips the bench_function closure altogether and shouldn't hit any init code?

roberto-bayardo: You can but you'll wait a long time so the init code for whatever reason still insists on running.

There are probably other workarounds, have not explored with recent LLMs which probably know a lot better!

storage/src/qmdb/benches/init.rs
```diff
@@ -116,6 +105,14 @@ fn bench_var_value_init(c: &mut Criterion) {
                     |b| {
                         b.to_async(&runner).iter_custom(|iters| async move {
                             let ctx = context::get::<Context>();
+
+                            // Setup: populate database (not timed).
+                            with_var_value_db!(ctx, variant, |mut db| {
```
patrick-ogrady: This will significantly increase the runtime of each sample.

storage/src/qmdb/benches/init.rs
```diff
@@ -49,14 +49,7 @@ fn bench_fixed_value_init(c: &mut Criterion) {
     for elements in ELEMENTS {
         for operations in OPERATIONS {
             for variant in FIXED_VALUE_VARIANTS {
-                // Setup: populate database
-                commonware_runtime::tokio::Runner::new(cfg.clone()).start(|ctx| async move {
```
patrick-ogrady, line 53: We keep this inside of `bench_function` to improve init time.

roberto-bayardo (dismissed):

roberto-bayardo (dismissed):

roberto-bayardo (dismissed):

storage/src/archive/benches/get.rs
```diff
@@ -67,95 +67,110 @@ async fn read_concurrent_indices(a: &Archive, indices: &[u64]) {
...
+                                compression
+                                    .map(|l| l.to_string())
+                                    .unwrap_or_else(|| "off".into()),
+                                reads
+                            ),
+                            |b| {
+                                // Setup: populate database (once, on first sample).
+                                if !initialized {
+                                    keys =
+                                        commonware_runtime::tokio::Runner::new(Config::default())
```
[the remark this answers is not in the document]
patrick-ogrady: oops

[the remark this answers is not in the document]
patrick-ogrady: Fixing here: #3599

PR #3605, opened 2026-04-15
[runtime] Remove `async move` wrapping from `spawn_cell!`

patrick-ogrady:
`spawn_cell!` used to wrap `$body` in an inner async move { ... $body.await }, which doubled the spawn-time state machine in debug builds and was overflowing the stack when captured futures got large.

We changed the macro to pass `$body` through as a Future directly (callers drop the redundant `.await`), so the runtime polls one state machine instead of two.

In release builds, this should get largely optimized away but still was odd (the new code is now a transparent passthrough to `spawn` ... which is what it was originally intended to be).

### DIFF
--- a/runtime/src/utils/cell.rs
+++ b/runtime/src/utils/cell.rs
@@ -13,8 +13,8 @@ use std::{
 const MISSING_CONTEXT: &str = "runtime context missing";
 const DUPLICATE_CONTEXT: &str = "runtime context already present";
 
-/// Spawn a task using a [`Cell`] by taking its context, executing the provided
-/// async block, and restoring the context before the block completes.
+/// Spawn a task using a [`Cell`] by taking its context, restoring the context synchronously
+/// in the spawned closure, and returning the provided future directly to the runtime.
 ///
 /// The macro uses the context's default spawn configuration (supervised, shared executor with
 /// `blocking == false`). If you need to mark the task as blocking or request a dedicated thread,
@@ -23,7 +23,7 @@ const DUPLICATE_CONTEXT: &str = "runtime context already present";
 macro_rules! spawn_cell {
     ($cell:expr, $body:expr $(,)?) => {{
         let __commonware_context = $cell.take();
-        __commonware_context.spawn(move |context| async move {
+        __commonware_context.spawn(move |context| {
             $cell.restore(context);
             $body
         })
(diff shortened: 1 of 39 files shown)


runtime/src/utils/cell.rs
```diff
@@ -23,7 +23,7 @@ const DUPLICATE_CONTEXT: &str = "runtime context already present";
 macro_rules! spawn_cell {
     ($cell:expr, $body:expr $(,)?) => {{
         let __commonware_context = $cell.take();
-        __commonware_context.spawn(move |context| async move {
+        __commonware_context.spawn(move |context| {
```
patrick-ogrady, line 26: `Spawn` already expects a future to be returned from this inner to poll...this new flow is a more transparent passthrough (rather than supporting `sync` calls OOB)

patrick-ogrady:
I had some unit tests here to try and pin this behavior but they were all a bit jank, so opted to go without.

runtime/src/utils/cell.rs
andresilva, line 26: I'm fine with this change, but are we sure this was actually increasing the size of the future? I'm asking because the approach on #3487 was to using an `async` block to avoid that.

andresilva: I guess that may be besides the point because we might still want this API anyway.

patrick-ogrady: I think it does have a benefit on DEBUG (at least as far as some vibed tests show):

```
┌─────────────────────────────────┬──────────────────┬───────────┬─────────┬─────────┬────────┐
  │              shape              │     capture      │ inner fut │   OLD   │   NEW   │ delta  │
  ├─────────────────────────────────┼──────────────────┼───────────┼─────────┼─────────┼────────┤
  │ loose captures                  │ ~850 B           │ 1472 B    │ 1816 B  │ 1472 B  │ 344 B  │
  ├─────────────────────────────────┼──────────────────┼───────────┼─────────┼─────────┼────────┤
  │ self.run(a,b), self ~584 B      │ equivocator-like │ 1936 B    │ 2584 B  │ 1936 B  │ 648 B  │
  ├─────────────────────────────────┼──────────────────┼───────────┼─────────┼─────────┼────────┤
  │ self.run(a,b,c,d), self ~9.6 KB │ engine-like      │ 20496 B   │ 30168 B │ 20496 B │ 9672 B │
  └─────────────────────────────────┴──────────────────┴───────────┴─────────┴─────────┴────────┘
```

andresilva (approved):

PR #3606, opened 2026-04-15
[runtime/iouring] wakeup through futex or eventfd

andresilva:
This PR reduces io_uring wake overhead by splitting the loop's sleep behavior into two paths. When the ring is fully idle, the loop now parks in userspace on a futex-backed state word instead of going through `io_uring_enter`, when there are active waiters and the loop may block in `submit_and_wait`, it continues to use the existing eventfd wake path. The wake state is packed into a single atomic word that carries the current wait target, a latched wake bit, and the published submission sequence.

The packed state keeps the arm-and-recheck protocol coherent while also coalescing repeated wake attempts. Producers increment the submission sequence on successful enqueue and only the first wake in a given armed epoch performs the actual futex wake or eventfd write, later wake attempts observe the latched bit and return. Out-of-band wakeups, including final-handle drop during shutdown, now latch a sticky wake even if no wait target is currently armed, so the next idle/blocking arm will skip sleeping instead of missing shutdown.

The Waker API is also simplified around loop semantics rather than wake mechanisms. The loop now uses `pending(processed_seq)`, `park_idle(processed_seq)`, and `arm(processed_seq) -> ArmGuard`, which hides the old futex/eventfd-specific arming details and makes the wait lifetime explicit through RAII.

All atomic operations on the packed wake state are now `Relaxed`. That is sufficient because the waker word is only a control-state machine for wait intent, wake deduplication, and the submission sequence, it does not publish request payload memory or channel-close visibility. Those memory-ordering guarantees remain the responsibility of tokio's MPSC and the surrounding request/channel machinery. Correctness here depends only on atomicity and per-location modification order of the single packed state word, plus the existing arm-and-recheck protocol that uses snapshots from that same atomic.

### DIFF
--- a/runtime/src/iouring/mod.rs
+++ b/runtime/src/iouring/mod.rs
@@ -1035,7 +1035,7 @@ mod tests {
         // after ring-size round-up.
         let mut registry = Registry::default();
         let cfg = Config {
-            size: (1 << 28) + 1,
+            size: (MAX_SUBMISSION_SEQUENCE_DOMAIN / 2) + 1,
             ..Default::default()
         };
         let _ = IoUringLoop::new(cfg, &mut registry);
--- a/runtime/src/iouring/mod.rs
+++ b/runtime/src/iouring/mod.rs
@@ -174,7 +174,7 @@ use timeout::{Tick, TimeoutWheel};
 mod waiter;
 use waiter::{CompletionOutcome, StageOutcome, WaiterId, Waiters};
 mod waker;
-use waker::{Waker, SUBMISSION_SEQ_MASK, WAKE_USER_DATA};
+use waker::{Waker, MAX_SUBMISSION_SEQUENCE_DOMAIN, SUBMISSION_SEQ_MASK, WAKE_USER_DATA};
 
 /// Packed `io_uring` `user_data` value.
 type UserData = u64;
@@ -461,8 +461,8 @@ impl IoUringLoop {
             .checked_next_power_of_two()
             .expect("ring size exceeds u32::MAX");
         assert!(
-            cfg.size < (1 << 29),
-            "rounded ring size must stay below 1<<29 to preserve the 29-bit wake sequence bound"
+            cfg.size < MAX_SUBMISSION_SEQUENCE_DOMAIN,
+            "rounded ring size must stay below the packed wake sequence domain"
         );
         let size = cfg.size as usize;
         let metrics = Arc::new(Metrics::new(registry));
@@ -666,6 +666,13 @@ impl IoUringLoop {
 
         // Reinstall wake poll only when a prior wake CQE indicated multishot
         // termination. Otherwise keep the existing poll registration.
+        //
+        // This check runs before every possible transition into the eventfd-backed
+        // blocking path. The fully idle futex path does not need the poll to be
+        // live, so an iteration that parks in futex may skip kernel entry
+        // entirely. If multishot termination was observed earlier, the next
+        // iteration that might block in `submit_and_wait` stages the rearm SQE
+        // here before entering the kernel again.
         if std::mem::take(&mut self.wake_rearm_needed) {
             self.waker.reinstall(&mut submission_queue);
         }
@@ -1017,7 +1024,7 @@ mod tests {
     }
 
     #[test]
-    #[should_panic(expected = "rounded ring size must stay below 1<<29")]
+    #[should_panic(expected = "rounded ring size must stay below the packed wake sequence domain")]
     fn test_iouring_loop_rejects_sizes_that_exceed_wake_sequence_domain() {
         // The wake state reserves only 29 bits for the submission sequence, so
         // the bounded request channel must stay strictly below that domain even
--- a/runtime/src/iouring/mod.rs
+++ b/runtime/src/iouring/mod.rs
@@ -769,12 +769,17 @@ impl IoUringLoop {
                     }
 
                     // `pending()`'s acquire load observed a published-ahead
-                    // sequence delta after the empty observation. The rounded
-                    // ring size stays below half the packed sequence domain, so
-                    // that delta is directional: producers are genuinely ahead.
-                    // Tokio's bounded MPSC also guarantees `try_recv()` returns
-                    // `Disconnected` only when the channel is closed AND empty
-                    // (we keep a canary test to lock in those semantics).
+                    // sequence delta after the empty observation. Producers
+                    // execute `send().await` before `publish()`, and that
+                    // release/acquire edge makes the corresponding enqueue
+                    // visible here before the second `try_recv()`.
+                    //
+                    // The rounded ring size stays below half the packed
+                    // sequence domain, so the observed delta is directional:
+                    // producers are genuinely ahead. Tokio's bounded MPSC also
+                    // guarantees `try_recv()` returns `Disconnected` only when
+                    // the channel is closed AND empty (we keep canaries for
+                    // both assumptions below).
                     self.receiver.try_recv().expect(
                         "published-ahead sequence observed after acquire, but channel had no request",
                     )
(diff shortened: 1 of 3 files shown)


patrick-ogrady:
LLM was worried about:
```
  - [P2] Flush staged wake rearm before entering the futex idle path — /Users/patrickogrady/code/monorepo-pr-3606/runtime/src/iouring/mod.rs:558-560
    When the effective ring size is 1, fill_submission_queue() can consume the only SQ slot with self.waker.reinstall(). This new idle branch then calls park_idle() without ever submitting that SQE, so after the first producer
    futex-wakes the loop the request is still stuck in the MPSC behind an SQ that only contains the unsent wake poll. The next pending && at_capacity iteration blocks in submit_and_wait() with no in-flight request to generate a
    CQE, so a valid single-entry configuration can hang on its first request.
  - [P1] Restore release ordering when publishing a queued request — /Users/patrickogrady/code/monorepo-pr-3606/runtime/src/iouring/waker.rs:190-194
    Changing the wake-state atomics to Relaxed removes the only ordering between tokio::mpsc::Sender::send() and the separate submission-sequence counter. On weakly ordered Linux targets like ARM64, the loop can observe the
    bumped sequence in pending()/arm() before receiver.try_recv() can observe the enqueued request, which turns the new fast path into a phantom "pending work" spin and can also send the loop into the pending && at_capacity
    wait path with nothing newly drainable. The old Release/Acquire pair was what made "observed sequence implies corresponding request is visible" true here.
```

Changes (tests fail without changes to code):
```
diff --git a/runtime/Cargo.toml b/runtime/Cargo.toml
index e20745727..e920d8215 100644
--- a/runtime/Cargo.toml
+++ b/runtime/Cargo.toml
@@ -54,6 +54,9 @@ opentelemetry-otlp = { workspace = true, features = ["http-proto"] }
 opentelemetry_sdk = { workspace = true, features = ["rt-tokio"] }
 tokio = { workspace = true, features = ["full"] }
 
+[target.'cfg(not(target_arch = "wasm32"))'.dev-dependencies]
+loom = "0.7"
+
 [features]
 default = []
 arbitrary = [
diff --git a/runtime/src/iouring/mod.rs b/runtime/src/iouring/mod.rs
index d378143df..3286b2b1a 100644
--- a/runtime/src/iouring/mod.rs
+++ b/runtime/src/iouring/mod.rs
@@ -556,6 +556,15 @@ impl IoUringLoop {
             // it. This bypasses the eventfd wake path when there are no active
             // waiters.
             if self.waiters.is_empty() {
+                // Flush any staged wake-poll rearm before sleeping in userspace.
+                // Otherwise a valid single-entry ring can park with the poll
+                // SQE occupying the only submission slot, stranding the next
+                // real request behind an SQE that has not reached the kernel.
+                if ring.submission().len() != 0 {
+                    self.submit_and_wait(&mut ring, 0, None)
+                        .expect("unable to submit to ring");
+                }
                 self.waker.park_idle(self.processed_seq);
                 continue;
             }
@@ -1709,6 +1718,50 @@ mod tests {
         handle.join().expect("io_uring loop thread panicked");
     }
 
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn test_single_entry_ring_submits_idle_wake_rearm_before_parking() {
+        // Verify a size=1 ring does not strand the first real request behind
+        // an unsent wake-poll rearm while sleeping on the futex idle path.
+        let cfg = Config {
+            size: 1,
+            ..Default::default()
+        };
+        let mut registry = Registry::default();
+        let (submitter, iouring) = IoUringLoop::new(cfg, &mut registry);
+        let idle_waker = iouring.waker.clone();
+        let handle = std::thread::spawn(move || iouring.run());
+
+        // Wait until the loop has entered the futex-backed idle path. In the
+        // buggy implementation, the staged wake-poll rearm is still sitting in
+        // the SQ at this point and consumes the only submission slot.
+        waker::tests::wait_until_futex_armed(&idle_waker);
+
+        let (sock_left, _sock_right) = UnixStream::pair().unwrap();
+        // SAFETY: sock_left is a valid fd that we own.
+        let file = unsafe { std::fs::File::from_raw_fd(sock_left.into_raw_fd()) };
+        submitter
+            .enqueue(Request::Sync(SyncRequest {
+                file: Arc::new(file),
+                result: None,
+                sender: tx,
+            }))
+            .await
+            .expect("failed to enqueue sync request");
+
+        let result = tokio::time::timeout(Duration::from_secs(2), rx)
+            .await
+            .expect("single-entry ring request timed out")
+            .expect("missing single-entry ring completion");
+
+        // Socket fsync may succeed or fail on a particular kernel, but the
+        // request must not remain stuck behind the wake-poll rearm.
+        let _ = result;
+
+        drop(submitter);
+        handle.join().unwrap();
+    }
+
     #[tokio::test]
     async fn test_timeout() {
         // Verify a timed recv completes with timeout once its deadline expires.
diff --git a/runtime/src/iouring/waker.rs b/runtime/src/iouring/waker.rs
index 88d9f0e25..aabc81680 100644
--- a/runtime/src/iouring/waker.rs
+++ b/runtime/src/iouring/waker.rs
@@ -186,12 +186,17 @@ impl Waker {
     /// armed and no wake has yet been claimed for that epoch, this caller
     /// claims `WAKE_SIGNALLED_BIT` with a follow-up atomic update and then
     /// signals the armed wait target.
+    ///
+    /// This publication uses `Release` ordering so the loop can treat an
+    /// observed sequence increment as proof that the previously enqueued MPSC
+    /// request is now visible when it reads the same packed state with
+    /// `Acquire`.
     #[inline]
     pub fn publish(&self) {
         let prev = self
             .inner
             .state
-            .fetch_add(SUBMISSION_INCREMENT, Ordering::Relaxed);
+            .fetch_add(SUBMISSION_INCREMENT, Ordering::Release);
 
         let waiting = prev & WAITING_MASK;
 
@@ -208,7 +213,7 @@ impl Waker {
     /// drained from the channel.
     #[inline]
     pub fn pending(&self, processed_seq: u32) -> bool {
-        ((self.inner.state.load(Ordering::Relaxed) >> STATE_BITS) & SUBMISSION_SEQ_MASK)
+        ((self.inner.state.load(Ordering::Acquire) >> STATE_BITS) & SUBMISSION_SEQ_MASK)
             != processed_seq
     }
 
@@ -220,7 +225,7 @@ impl Waker {
         let prev = self
             .inner
             .state
-            .fetch_or(WAITING_ON_FUTEX_BIT, Ordering::Relaxed);
+            .fetch_or(WAITING_ON_FUTEX_BIT, Ordering::Acquire);
 
         assert_eq!(
             prev & WAITING_MASK,
@@ -249,7 +254,7 @@ impl Waker {
         let prev = self
             .inner
             .state
-            .fetch_or(WAITING_ON_EVENTFD_BIT, Ordering::Relaxed);
+            .fetch_or(WAITING_ON_EVENTFD_BIT, Ordering::Acquire);
 
         assert_eq!(
             prev & WAITING_MASK,
@@ -453,6 +458,13 @@ impl Waker {
 pub mod tests {
     use super::*;
     use io_uring::IoUring;
+    use loom::{
+        sync::{
+            atomic::{AtomicBool, AtomicU32},
+            Arc as LoomArc,
+        },
+        thread as loom_thread,
+    };
     use std::{
         mem::size_of,
         os::fd::{AsRawFd, FromRawFd},
@@ -708,4 +720,73 @@ pub mod tests {
         // Direct eventfd read/write error paths should not perturb sequence tracking.
         assert_eq!(submitted_seq(&waker), before);
     }
+
+    #[test]
+    fn test_publish_release_orders_pending_visibility() {
+        // Model the contract between the channel enqueue and `pending()`:
+        // once the loop observes the published sequence, the corresponding
+        // request must already be visible to the channel consumer.
+        loom::model(|| {
+            let queued = LoomArc::new(AtomicBool::new(false));
+            let state = LoomArc::new(AtomicU32::new(0));
+
+            let producer = {
+                let queued = queued.clone();
+                let state = state.clone();
+                loom_thread::spawn(move || {
+                    queued.store(true, Ordering::Relaxed);
+                    state.fetch_add(SUBMISSION_INCREMENT, Ordering::Release);
+                })
+            };
+
+            let consumer = {
+                let queued = queued.clone();
+                let state = state.clone();
+                loom_thread::spawn(move || {
+                    if ((state.load(Ordering::Acquire) >> STATE_BITS) & SUBMISSION_SEQ_MASK) != 0 {
+                        assert!(queued.load(Ordering::Relaxed));
+                    }
+                })
+            };
+
+            producer.join().expect("producer thread panicked");
+            consumer.join().expect("consumer thread panicked");
+        });
+    }
+
+    #[test]
+    fn test_publish_release_orders_arm_visibility() {
+        // Model the contract between the channel enqueue and `arm()`:
+        // if arming observes newly published work in the packed sequence,
+        // the queue entry must already be visible before the loop decides
+        // whether it can block.
+        loom::model(|| {
+            let queued = LoomArc::new(AtomicBool::new(false));
+            let state = LoomArc::new(AtomicU32::new(0));
+
+            let producer = {
+                let queued = queued.clone();
+                let state = state.clone();
+                loom_thread::spawn(move || {
+                    queued.store(true, Ordering::Relaxed);
+                    state.fetch_add(SUBMISSION_INCREMENT, Ordering::Release);
+                })
+            };
+
+            let consumer = {
+                let queued = queued.clone();
+                let state = state.clone();
+                loom_thread::spawn(move || {
+                    let prev = state.fetch_or(WAITING_ON_EVENTFD_BIT, Ordering::Acquire);
+                    let snapshot = prev | WAITING_ON_EVENTFD_BIT;
+                    if ((snapshot >> STATE_BITS) & SUBMISSION_SEQ_MASK) != 0 {
+                        assert!(queued.load(Ordering::Relaxed));
+                    }
+                })
+            };
+
+            producer.join().expect("producer thread panicked");
+            consumer.join().expect("consumer thread panicked");
+        });
+    }
 }
```

andresilva:
The issue with ring size = 1 seems very convoluted to actually happen in practice, but the simplest fix is just to require ring size is always bigger than 1.

On the ordering point, I ended up with a narrower conclusion than your suggestion. Tokio's MPSC already synchronizes its own queue and close state, so the packed state word does not need to be the publication mechanism for correctness. Using stronger ordering on the packed state would be opting into a stronger cross-object contract, not repairing a missing publication mechanism. For normal `publish()`, the submission sequence is only a sticky "do not sleep yet" hint. The loop may observe the sequence bump before `try_recv()` can observe the corresponding enqueue, but that does not lose work: the sequence mismatch persists until the channel drain catches up, so the worst case is an extra iteration, not sleeping through the request.

The real hazard is the out-of-band wake on last-sender drop. That path does not bump the submission sequence, it only sets `WAKE_SIGNALLED_BIT` and wakes the loop. If the loop consumes that wake before the channel disconnected state is visible, it can clear the wait epoch, see `Empty` on the `try_recv()`, and go back to sleep with no second wake coming. So the minimal ordering contract for correctness is: `wake(): Release` paired with `clear_wait(): Acquire`, while leaving `publish()`, `pending()`, `park_idle()`, and `arm()` as `Relaxed`.

The proposed solution doesn't fix the shutdown race, and it also doesn't eliminate the already-missed `try_recv()` in the current iteration, because structurally the loop does `try_recv()` BEFORE calling `pending()`. So making `publish(): Release` and `pending(): Acquire` cannot retroactively make that earlier `try_recv()` see the just-enqueued item. It may still help reduce subsequent visibility-driven extra iterations. That said I'm still thinking whether it's better to actually use stronger ordering to reduce extra iterations (the performance tradeoff is probably negligible on x86 and matters more on ARM). And I think there might be a situation where we block unnecessarily (only to be woken up right after). Fixing both might require slightly restructuring the loop.

patrick-ogrady (approved):

PR #3607, opened 2026-04-16
[storage/qmdb/sync] bind fetches to target size and wait for boundary state 

clabby:
## Overview

After a target update, a resolver could answer a fresh lower-bound request with a valid proof for a retained historical root. The engine would accept that response, retry the boundary fetch, but still report `reached_target` based only on journal length. That let startup finalize without current pinned boundary nodes and fail with `MissingNode`.

The fix binds each tracked fetch request to the target size it was issued against and rejects responses whose `proof.leaves` do not match the request's target size.

### DIFF
--- a/storage/src/merkle/journaled.rs
+++ b/storage/src/merkle/journaled.rs
@@ -245,19 +245,11 @@ impl<F: Family, E: RStorage + Clock + Metrics, D: Digest> Journaled<F, E, D> {
 
     /// Read-only peek at the persisted structure's root and boundaries.
     ///
-    /// Opens the journal and metadata partitions, reconstructs the in-memory MMR
-    /// from persisted pinned nodes, and returns the root without performing any
-    /// of the mutations that [`Self::init_sync`] or [`Self::init`] would apply
-    /// (no `metadata.sync`, no `journal.prune`, no `journal.clear_to_size`).
-    ///
     /// Returns `Ok(None)` when:
     /// - The journal is empty.
     /// - Journal size is structurally invalid and would require a rewind (i.e.
     ///   a crash left the structure in an unrecoverable state for a read-only
     ///   probe).
-    ///
-    /// Intended for callers that need to verify "does persisted state match a
-    /// target" without paying the cost of a full database rebuild.
     pub async fn peek_root(
         context: E,
         cfg: Config,
--- a/storage/src/merkle/journaled.rs
+++ b/storage/src/merkle/journaled.rs
@@ -243,6 +243,94 @@ impl<F: Family, E: RStorage + Clock + Metrics, D: Digest> Journaled<F, E, D> {
         Ok(())
     }
 
+    /// Read-only peek at the persisted structure's root and boundaries.
+    ///
+    /// Opens the journal and metadata partitions, reconstructs the in-memory MMR
+    /// from persisted pinned nodes, and returns the root without performing any
+    /// of the mutations that [`Self::init_sync`] or [`Self::init`] would apply
+    /// (no `metadata.sync`, no `journal.prune`, no `journal.clear_to_size`).
+    ///
+    /// Returns `Ok(None)` when:
+    /// - The journal is empty.
+    /// - Journal size is structurally invalid and would require a rewind (i.e.
+    ///   a crash left the structure in an unrecoverable state for a read-only
+    ///   probe).
+    ///
+    /// Intended for callers that need to verify "does persisted state match a
+    /// target" without paying the cost of a full database rebuild.
+    pub async fn peek_root(
+        context: E,
+        cfg: Config,
+        hasher: &impl Hasher<F, Digest = D>,
+    ) -> Result<Option<(Location<F>, Location<F>, D)>, Error<F>> {
+        let journal_cfg = JConfig {
+            partition: cfg.journal_partition,
+            items_per_blob: cfg.items_per_blob,
+            write_buffer: cfg.write_buffer,
+            page_cache: cfg.page_cache,
+        };
+        let journal: Journal<E, D> =
+            Journal::init(context.with_label("merkle_journal_peek"), journal_cfg).await?;
+        let journal_size = Position::<F>::new(journal.size().await);
+
+        if journal_size == 0 {
+            return Ok(None);
+        }
+
+        // Bail if the journal would require a rewind to reach a valid size.
+        // Probe is read-only; the caller will handle recovery via `init`.
+        let last_valid_size = F::to_nearest_size(journal_size);
+        if last_valid_size != journal_size {
+            return Ok(None);
+        }
+
+        let metadata_cfg = MConfig {
+            partition: cfg.metadata_partition,
+            codec_config: ((0..).into(), ()),
+        };
+        let metadata = Metadata::<_, U64, Vec<u8>>::init(
+            context.with_label("merkle_metadata_peek"),
+            metadata_cfg,
+        )
+        .await?;
+
+        let prune_loc = metadata
+            .get(&U64::new(PRUNED_TO_PREFIX, 0))
+            .map(|bytes| -> Result<Location<F>, Error<F>> {
+                let raw: [u8; 8] = bytes
+                    .as_slice()
+                    .try_into()
+                    .map_err(|_| Error::DataCorrupted("metadata pruned_to is not 8 bytes"))?;
+                Ok(Location::<F>::new(u64::from_be_bytes(raw)))
+            })
+            .transpose()?
+            .unwrap_or_else(|| Location::<F>::new(0));
+        let prune_pos = Position::try_from(prune_loc)?;
+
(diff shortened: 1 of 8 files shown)


storage/src/qmdb/sync/engine.rs
```diff
@@ -614,8 +650,9 @@ where
         // Verify the proof. Pinned nodes are only extracted from proofs
         // for the current root because the database needs them for the
         // latest tree size.
-        let need_pinned =
-            is_current && self.pinned_nodes.is_none() && start_loc == self.target.range.start();
+        let need_pinned = is_current_target
+            && self.pinned_nodes.is_none()
+            && start_loc == self.target.range.start();
```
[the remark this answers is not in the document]
clabby: Edge case, tho great find.

storage/src/qmdb/sync/requests.rs
```diff
@@ -17,6 +19,21 @@ use std::{
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub(super) struct Id(u64);
 
+/// Immutable details about a tracked request.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(super) struct RequestInfo {
+    /// The location of the first requested operation.
+    pub start_loc: Location,
+    /// The tree size the request asked the resolver to prove against.
```
roberto-bayardo: Not really a "tree size" at this layer of abstraction but a "database size".    (E.g. see `fn size` in the `DbAny` trait).

roberto-bayardo: Although if you want to stick with tree-terminology, you could call it "target_leaves", which matches what's inside the proofs.

storage/src/merkle/journaled.rs
```diff
@@ -243,6 +243,94 @@ impl<F: Family, E: RStorage + Clock + Metrics, D: Digest> Journaled<F, E, D> {
         Ok(())
     }
 
+    /// Read-only peek at the persisted structure's root and boundaries.
+    ///
+    /// Opens the journal and metadata partitions, reconstructs the in-memory MMR
```
roberto-bayardo: I don't think you need to describe these implementation details in the function doc here. (I find it more confusing than helpful).  Would drop this paragraph entirely.

storage/src/merkle/journaled.rs
```diff
@@ -243,6 +243,94 @@ impl<F: Family, E: RStorage + Clock + Metrics, D: Digest> Journaled<F, E, D> {
...
+    /// of the mutations that [`Self::init_sync`] or [`Self::init`] would apply
+    /// (no `metadata.sync`, no `journal.prune`, no `journal.clear_to_size`).
+    ///
+    /// Returns `Ok(None)` when:
+    /// - The journal is empty.
+    /// - Journal size is structurally invalid and would require a rewind (i.e.
+    ///   a crash left the structure in an unrecoverable state for a read-only
+    ///   probe).
+    ///
+    /// Intended for callers that need to verify "does persisted state match a
```
roberto-bayardo: Would also drop this -- this is a sync implementation detail.  I think the function stands on its own independent of sync.  "Here's an on-disk merkle tree, give me its root and bounds without modifying it"

storage/src/qmdb/any/sync/mod.rs
```diff
@@ -136,6 +136,34 @@ macro_rules! impl_sync_database {
                 .await
             }
 
+            async fn has_local_target_state(
+                context: Self::Context,
+                config: &Self::Config,
+                target: &qmdb::sync::Target<Self::Digest>,
+            ) -> bool {
+                // Read-only peek: compute the persisted ops MMR root without
```
roberto-bayardo: Seems this comment is just repeating what the peek_root function doc already says. Would just shorten this to "// Get the root and bounds of the MMR"

storage/src/qmdb/any/sync/mod.rs
```diff
@@ -136,6 +136,34 @@ macro_rules! impl_sync_database {
...
+                // Avoids the O(N) snapshot rebuild and all disk mutations that
+                // `Self::init` would perform.
+                let hasher = StandardHasher::<H>::new();
+                let peek = journaled::Mmr::<_, H::Digest>::peek_root(
+                    context.with_label("local_target_probe"),
+                    config.merkle_config.clone(),
+                    &hasher,
+                )
+                .await;
+                // Only (journal_leaves, root) are checked. The merkle pruning
```
roberto-bayardo: This comment can be improved. Maybe something like:

  // Make sure the root and size of our persisted local state match the target. We don't check                                                            
  // the lower bound explicitly as the final commit op commits to the                                                                     
  // inactivity floor.

storage/src/qmdb/current/sync/mod.rs
```diff
@@ -286,6 +281,36 @@ macro_rules! impl_current_sync_database {
                 .await
             }
 
+            async fn has_local_target_state(
+                context: Self::Context,
+                config: &Self::Config,
+                target: &qmdb::sync::Target<Self::Digest>,
+            ) -> bool {
+                // Read-only peek: compute the persisted ops MMR root without
+                // opening the full database or replaying the operations log.
```
roberto-bayardo: would fix this comment and the one below as suggested for the anydb impl

storage/src/qmdb/current/sync/mod.rs
```diff
@@ -286,6 +281,36 @@ macro_rules! impl_current_sync_database {
                 .await
             }
 
+            async fn has_local_target_state(
```
roberto-bayardo, line 289: Is this impl identical to any's? If so, can they be shared?

storage/src/qmdb/sync/requests.rs
clabby: Updated to "database size"

storage/src/merkle/journaled.rs
clabby: Done

storage/src/qmdb/any/sync/mod.rs
clabby: Done

storage/src/merkle/journaled.rs
clabby: Done

storage/src/qmdb/current/sync/mod.rs
clabby: Done

storage/src/qmdb/current/sync/mod.rs
clabby: They can be shared - updated.

storage/src/merkle/journaled.rs
```diff
@@ -243,6 +243,86 @@ impl<F: Family, E: RStorage + Clock + Metrics, D: Digest> Journaled<F, E, D> {
...
+            items_per_blob: cfg.items_per_blob,
+            write_buffer: cfg.write_buffer,
+            page_cache: cfg.page_cache,
+        };
+        let journal: Journal<E, D> =
+            Journal::init(context.with_label("merkle_journal_peek"), journal_cfg).await?;
+        let journal_size = Position::<F>::new(journal.size().await);
+
+        if journal_size == 0 {
+            return Ok(None);
```
roberto-bayardo: An empty journal has a perfectly valid root. I wonder if we should just return that?

storage/src/qmdb/any/sync/mod.rs
```diff
@@ -47,6 +47,48 @@ use std::ops::Range;
 #[cfg(test)]
 pub(crate) mod tests;
 
+/// Returns whether persisted local state already matches the requested sync target.
+///
+/// Shared across [crate::qmdb::any] and [crate::qmdb::current] sync because both
+/// build on the same operations-MMR layout and share the same merkle partition.
+/// Uses a read-only MMR peek: the journal and metadata partitions are opened,
+/// the in-memory MMR is reconstructed from persisted pinned nodes, the root is
+/// computed, and handles are dropped. No disk mutations, no snapshot replay.
```
roberto-bayardo: This is an implementation detail that isn't needed here.

storage/src/qmdb/sync/engine.rs
```diff
@@ -261,6 +263,22 @@ where
             }));
         }
 
+        // Probe for persisted local state matching the target before opening
+        // any engine-owned handles. `has_local_target_state` opens the full
+        // database to inspect bounds and root; running it before journal
+        // creation ensures its handles are dropped before the engine takes
+        // its own, avoiding concurrent access to the same on-disk partition.
```
roberto-bayardo: Seems a bit out of place to explain how has_local_target_state works here?

storage/src/merkle/journaled.rs
clabby: Done.

roberto-bayardo:
LGTM. I'll let @danlaine make the final call to merge.

danlaine (approved):

PR #3609, opened 2026-04-16
[storage/qmdb] Extend sync protocol to support MMB family

roberto-bayardo:
## Summary

Generalize QMDB sync to work across both Merkle families (MMR and MMB), and fix pruned `current` sync for MMB by syncing the sender's grafted overlay state explicitly and authenticating it with the canonical root.

Before this change, the shared sync path was effectively MMR-only across all variants. In particular, `current` sync reconstructed its grafted pruning boundary by deriving grafted pinned nodes from the synced ops tree at `range.start`. That works for MMR, but not for MMB: delayed merges can cause the ops tree's boundary pins to diverge from the grafted tree's true pruning boundary, which can produce a different grafted tree and therefore a different canonical root on the receiver.

This change does two things:

- generalizes the shared sync infrastructure so databases can sync over either MMR or MMB
- makes pruned `current` sync reproduce the sender's actual overlay state instead of trying to infer it from the ops range

### 1. Generalized shared sync over the database's Merkle family

The sync layer is now parameterized over `Database::Family` rather than implicitly assuming MMR everywhere.

That affects the shared types and plumbing used by all database variants, including:

- `FetchResult`
- `Resolver`
- the sync engine request/result flow
- `Database::from_sync_result`
- `Target`

This means:

- `any` sync now works generically for both MMR and MMB
- `current` sync can share the same family-generic engine/plumbing instead of relying on MMR-specific assumptions
- `immutable` and `keyless` continue to use the same shared sync infrastructure, but now through the same family-generic path

### 2. Added explicit overlay sync state for `current`

Introduced `CurrentOverlayState` in `storage/src/qmdb/current/sync/mod.rs`:

- `pruned_chunks`
- `grafted_pinned_nodes`

This is carried on the first boundary batch via `FetchResult::overlay_state`.

### 3. Rebuild `current` from sender overlay state

`current::sync::build_db` no longer derives grafted pinned nodes from `range.start`.

Instead, for normal pruned sync it:

- validates the sender-supplied overlay payload
- seeds the grafted tree from `grafted_pinned_nodes`
- initializes the bitmap from the sender's `pruned_chunks`
- replays the synced ops log to populate the bitmap tail
- computes the canonical root from the rebuilt state

This is the key MMB fix: the receiver now rebuilds from the sender's actual grafted pruning state rather than from a lossy ops-boundary approximation.

### 4. Added canonical-root authentication for `current`

The sync engine already authenticates streamed batches against the **ops root**. That proves the journal contents, but it does not authenticate grafted overlay state.

To close that gap, `Target` now carries an optional `canonical_root`. When supplied for `current` sync, `build_db` verifies:

- rebuilt canonical root == trusted canonical root

before calling `sync_metadata()`.

This prevents a malicious or buggy sender from causing the receiver to persist a diverged grafted overlay.

### 5. Require fresh boundary state before authenticated completion

For authenticated `current` sync, the engine now defers completion until a fresh boundary batch has been fetched and verified through the normal request path.

This avoids the stale-metadata retry trap where:

- a prior sync finished downloading the journal
- canonical-root verification failed before metadata persistence
- a later retry could otherwise have reused stale persisted overlay metadata

## Why this is needed

The old shared sync flow assumed that ops-boundary information was enough to reconstruct database state.

That is true for MMR-backed variants, and it is fine for `any`, where the authenticated state is just the ops structure itself.

It is not true for pruned `current` on MMB, because the canonical root depends on:

- the ops root
- the grafted tree root
- any partial chunk state

and delayed merges mean the grafted pruning boundary cannot always be recovered from `range.start` alone.

The fix is therefore:

- make the shared sync plumbing family-generic
- ship `current` overlay state explicitly
- rebuild from that overlay state
- authenticate the rebuilt canonical root when the caller provides a trusted canonical root

## Tests

Added/updated focused coverage for:

- family-generic sync plumbing
- successful pruned MMB `current` sync with canonical-root authentication
- canonical-root mismatch rejection
- authenticated unpruned `current` sync
- the previous MMB target-update regression shape, which now passes as a regression guard

## Result

This change makes sync:

- family-generic across MMR and MMB
- correct for pruned `current` on MMB delayed-merge families
- faithful to the sender's actual pruning state
- robust against stale-overlay reuse on retries
- authenticated at the canonical-root level for `current`, rather than only at the ops-root level

### DIFF
--- a/examples/sync/src/bin/client.rs
+++ b/examples/sync/src/bin/client.rs
@@ -8,7 +8,7 @@ use commonware_codec::{EncodeShared, Read};
 use commonware_runtime::{
     tokio as tokio_runtime, BufferPooler, Clock, Metrics, Network, Runner, Spawner, Storage,
 };
-use commonware_storage::qmdb::sync;
+use commonware_storage::{mmr, qmdb::sync};
 use commonware_sync::{
     any, crate_version, current, databases::DatabaseType, immutable, keyless, net::Resolver,
     Digest, Error, Key,
@@ -60,9 +60,9 @@ struct Config {
 async fn target_update_task<E, Op, D>(
     context: E,
     resolver: Resolver<Op, D>,
-    update_tx: mpsc::Sender<sync::Target<D>>,
+    update_tx: mpsc::Sender<sync::Target<mmr::Family, D>>,
     interval_duration: Duration,
-    initial_target: sync::Target<D>,
+    initial_target: sync::Target<mmr::Family, D>,
 ) -> Result<(), Error>
 where
     E: Clock,
--- a/examples/sync/src/bin/server.rs
+++ b/examples/sync/src/bin/server.rs
@@ -173,6 +173,7 @@ where
         target: Target {
             root,
             range: non_empty_range!(inactivity_floor, size),
+            canonical_root: None,
         },
     };
 
--- a/examples/sync/src/net/resolver.rs
+++ b/examples/sync/src/net/resolver.rs
@@ -3,7 +3,10 @@ use crate::net::request_id;
 use commonware_codec::{EncodeShared, IsUnit, Read};
 use commonware_cryptography::Digest;
 use commonware_runtime::{Network, Spawner};
-use commonware_storage::{mmr::Location, qmdb::sync};
+use commonware_storage::{
+    mmr::{self, Location},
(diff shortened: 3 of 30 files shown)


roberto-bayardo:
Closed in favor of #3621

PR #3610, opened 2026-04-16
[storage/qmdb/current] add ability to authenticate ops root from qmdb canonical root

roberto-bayardo:
In order to avoid having to store both ops_root and canonical tree root in a block, we need to be able to authenticate the ops root.   This PR adds that capability.


roberto-bayardo:
bugbot run

danlaine (approved):

PR #3613, opened 2026-04-16
[consensus/simplex] Infer parent certification from notarization

### LINKED ISSUE #3433: [consensus/simplex] Infer Certification

0xAysh:
Fixes #3433 

## Summary

When a notarization for view N arrives, the f+1 signers had to certify N's parent before they could vote. If this node has not certified that parent yet, it can stall and wait for automaton responses that are not needed. This PR infers that certification from the notarization and keeps progress moving. Also, a quick note: this PR was noisy earlier while I cleaned up the approach and re-opened it with only the final design and fixes.

## Design decisions
 - Inference runs on first notarization broadcast (`try_broadcast_notarization`), not on receive.
 - Inference only certifies ancestors that are Ready and notarized, and stops at clear boundaries (<= last_finalized, missing round, non-Ready, missing notarization).
 - Sync is batched: `handle_certification` is append-only, and broadcast inference does `sync_journal(view)` then inferred appends then one `sync_all()`.
 - A post-replay inference pass recovers inferred certifications that were not written before crash.
Inferred certifications use normal signaling (`resolver.certified(..., true)` and `Activity::Certification`) and remove inferred views from `certification_candidates()`.

## Changes

- `round.rs`: Added `is_certify_ready()` to check if a view can be inferred.
- `state.rs`: Added `infer_ancestors(view)`. It walks backward from the parent of `view`, certifies ancestor views that are `Ready` and notarized, removes those views from `certification_candidates`, and returns `Vec<(View, Notarization)>` so the caller can journal and signal them. It stops at `<= last_finalized`, missing round entry, non-`Ready` state, or missing notarization.
- `actor.rs`:
  - `handle_certification` is now append-only. It no longer calls sync.
  - The certify-wait handler now calls `sync_journal` after `handle_certification`.
  - In `try_broadcast_notarization`, after `sync_journal` makes the notarization durable, it calls `infer_ancestors`, journals each inferred ancestor, runs one `sync_all`, then sends `resolver.certified` and `reporter.report(Activity::Certification)` for each inferred view.
  - Added one post-replay inference pass after journal replay. On startup the certify pool is empty, so views are `Ready`. This restores inferred certifications that were not written before a crash.

## Crash safety

Sync order is `sync_journal(notarization)` -> `append_journal(ancestors)` -> `sync_all()`. This guarantees the notarization is durable before inferred ancestor certifications. If a crash happens between those sync points, the post-replay pass derives the missing certifications again.

## Tests

- Added 8 unit tests for `infer_ancestors` to cover all stop conditions and `certification_candidates` cleanup.
- Added `test_post_replay_inference_certifies_ancestor`: first run uses `Certifier::Pending` (view 3 stays `Outstanding`), notarization for view 4 arrives (inference stops), then crash. Second run uses `Certifier::Cancel`, and certification must come from post-replay inference. The test verifies `resolver.certified(view_3, true)` is emitted. It runs on all 6 scheme variants.
- Existing `only_finalization_rescues_validator` stays unchanged. `Certifier::Cancel` gives `Outstanding`, which is a valid stop condition for inference.


0xAysh:
handle_certification calls sync_journal on every call. In the inferred ancestor walk, this means one fsync per ancestor in the chain, O(n) fsyncs instead of one. In the normal automaton path this was never an issue since handle_certification was called once per notarization, but the loop changes that. Is there a way to batch the journal syncs across multiple views, or should the inferred path append all ancestor certification entries first and sync once at the end?

consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -414,14 +414,32 @@ impl<
...
+    /// can forward them to the resolver and reporter.
+    async fn handle_notarization(
+        &mut self,
+        notarization: Notarization<S, D>,
+    ) -> Vec<Notarization<S, D>> {
         let view = notarization.view();
         let artifact = Artifact::Notarization(notarization.clone());
         let (added, equivocator) = self.state.add_notarization(notarization);
+        let mut inferred = Vec::new();
         if added {
```
0xAysh, line 412: infer_parent() only walks up one level

0xAysh:
this PR is ready for review. Bugbot did not run after my latest push. I checked the PR checks and do not see a Bugbot run. Could anyone please take a look, or tell me if I should retrigger it from my side?

PR #3544, opened 2026-04-07
[runtime] Add page cache metrics

roberto-bayardo:
- Add `page_faults` metric to `CacheRef` (page cache), incremented on every cache miss that enters the fault handler                                      
 - Add `page_evictions` metric to `CacheRef` that is incremented when an old page has to be evicted to make room for a new one.

PR #3614, opened 2026-04-16
[runtime] Refactor Context

patrick-ogrady:
Related: #3544


collector/src/p2p/mod.rs
```diff
@@ -165,7 +167,7 @@ mod tests {
     ) -> Mailbox<PublicKey, Request> {
         let public_key = signer.public_key();
         let (engine, mailbox) = Engine::new(
-            context.with_label(&format!("engine_{public_key}")),
+            context.child(&format!("engine_{public_key}")),
```
patrick-ogrady: nuanced convo about what deserves to be "on the path" vs a "label"?

arguably, we should force child to be "static" and all dynamic info to be placed as an attribute?

PR #3616, opened 2026-04-17
[cryptography] Vendor `ed25519-consensus`

clabby:
## Overview

> [!NOTE]
>
> Stacked PR:
> - #3617
> - 👉 #3616
> - `main`

Vendors the [`ed25519-consensus`](https://docs.rs/ed25519-consensus/latest/ed25519_consensus/) crate with a few small modifications, namely swapping the backend to [`curve25519-dalek`](https://docs.rs/curve25519-dalek/latest/curve25519_dalek/) over [`curve25519-dalek-ng`](https://docs.rs/curve25519-dalek-ng/latest/curve25519_dalek_ng/)

Future Work: https://github.com/commonwarexyz/monorepo/issues/3668


cryptography/src/ed25519/ed25519_consensus/signing_key.rs
```diff
@@ -0,0 +1,159 @@
...
+        let h = Sha512::digest(&seed[..]);
+
+        // Convert the low half to a scalar with Ed25519 "clamping"
+        let s = {
+            let mut scalar_bytes = [0u8; 32];
+            scalar_bytes[..].copy_from_slice(&h.as_slice()[0..32]);
+            scalar_bytes[0] &= 248;
+            scalar_bytes[31] &= 127;
+            scalar_bytes[31] |= 64;
+            Scalar::from_bytes_mod_order(scalar_bytes)
```
cronokirby, line 99: I think we should use "from_canonical_bytes" and then expect, because the clamping above should already produce a reduced scalar. My thinking is that if there's a bug in the clamping, what we want is to detect this loudly, rather than silently reduce.

cryptography/src/ed25519/ed25519_consensus/signature.rs
```diff
@@ -0,0 +1,63 @@
+use super::Error;
+use commonware_utils::hex;
+use core::convert::TryFrom;
+
+/// An Ed25519 signature.
+#[derive(Copy, Clone, Eq, PartialEq)]
+#[allow(non_snake_case)]
+pub struct Signature {
+    pub(crate) R_bytes: [u8; 32],
+    pub(crate) s_bytes: [u8; 32],
```
cronokirby: Here, and in some other places in this diff, I think we want `pub(super)`, because we want other ed code to access this, but not everything else in cryptography to. (At least, that's my preference as a big fan of encapsulation)

cryptography/src/ed25519/core/batch.rs
```diff
@@ -0,0 +1,218 @@
...
+                    .decompress()
+                    .ok_or(Error::InvalidSignature)?;
+                let s = Scalar::from_canonical_bytes(sig.s_bytes)
+                    .into_option()
+                    .ok_or(Error::InvalidSignature)?;
+                let z = Scalar::from(gen_u128(&mut rng));
+                B_coeff -= z * s;
+                Rs.push(R);
+                R_coeffs.push(z);
+                A_coeff += z * k;
```
cronokirby, line 175: I don't know if using a small scalar here actually makes things faster, because I think the Scalar multiplication algorithm is constant-time.

cronokirby: AFAICT from a review of the dalek code this would seem to be the case, so I think it would be simpler to just generate a scalar (perhaps with a comment about this choice)

cryptography/src/ed25519/core/mod.rs
patrick-ogrady, line 1: Let's call this `ed25519/core`?

cryptography/src/ed25519/core/batch.rs
cronokirby: I missed that the coefficient for R is plain, so you can benefit from this.

cryptography/src/ed25519/core/verification_key.rs
```diff
@@ -0,0 +1,256 @@
...
+/// # use commonware_cryptography::ed25519::ed25519_consensus::*;
+/// # let msg = b"ed25519-consensus";
+/// # let sk = SigningKey::new(thread_rng());
+/// # let sig = sk.sign(msg);
+/// # let vk_bytes = VerificationKeyBytes::from(&sk);
+/// VerificationKey::try_from(vk_bytes)
+///     .and_then(|vk| vk.verify(&sig, msg));
+/// ```
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct VerificationKeyBytes(pub(super) [u8; 32]);
```
patrick-ogrady, line 21: In the future, it may be worthwhile to further unify this code with commonware-types?

cryptography/src/ed25519/core/signing_key.rs
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    /// Obtain the verification key associated with this signing key.
+    pub const fn verification_key(&self) -> VerificationKey {
+        self.vk
+    }
+}
+
+impl core::fmt::Debug for SigningKey {
+    fn fmt(&self, fmt: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
```
patrick-ogrady, line 39: We need to redact these?

cryptography/src/ed25519/core/signing_key.rs
```diff
@@ -0,0 +1,164 @@
...
+                A_bytes: VerificationKeyBytes(A.compress().to_bytes()),
+            },
+        }
+    }
+}
+
+impl zeroize::Zeroize for SigningKey {
+    fn zeroize(&mut self) {
+        self.seed.zeroize();
+        self.s.zeroize()
```
patrick-ogrady: I think we should zeroize prefix?

patrick-ogrady (commented):
We may want to wrap the inner SigningKey fields in "Secret"?

patrick-ogrady:
Maybe that's useless/double wrapping (because we do that on the outer)?

But would look something liek this:

```
diff --git a/cryptography/src/ed25519/core/signing_key.rs b/cryptography/src/ed25519/core/signing_key.rs
index 07a9a7617..2b8f5606b 100644
--- a/cryptography/src/ed25519/core/signing_key.rs
+++ b/cryptography/src/ed25519/core/signing_key.rs
@@ -1,5 +1,5 @@
 use super::{Error, Signature, VerificationKey, VerificationKeyBytes};
-use commonware_utils::hex;
+use crate::Secret;
 use core::convert::TryFrom;
 use curve25519_dalek::{constants, scalar::Scalar};
 use rand_core::{CryptoRng, RngCore};
@@ -10,9 +10,9 @@ use sha2::{digest::Update, Digest, Sha512};
 /// This is also called a secret key by other implementations.
 #[derive(Clone)]
 pub struct SigningKey {
-    seed: [u8; 32],
-    s: Scalar,
-    prefix: [u8; 32],
+    seed: Secret<[u8; 32]>,
+    s: Secret<Scalar>,
+    prefix: Secret<[u8; 32]>,
     vk: VerificationKey,
 }
 
@@ -20,13 +20,8 @@ impl SigningKey {
     /// Returns the byte encoding of the signing key.
     ///
     /// This is the same as `.into()`, but does not require type inference.
-    pub const fn to_bytes(&self) -> [u8; 32] {
-        self.seed
-    }
-
-    /// View the byte encoding of the signing key.
-    pub const fn as_bytes(&self) -> &[u8; 32] {
-        &self.seed
+    pub fn to_bytes(&self) -> [u8; 32] {
+        self.seed.expose(|seed| *seed)
     }
 
     /// Obtain the verification key associated with this signing key.
@@ -38,9 +33,9 @@ impl SigningKey {
 impl core::fmt::Debug for SigningKey {
     fn fmt(&self, fmt: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
         fmt.debug_struct("SigningKey")
-            .field("seed", &hex(&self.seed))
+            .field("seed", &self.seed)
             .field("s", &self.s)
-            .field("prefix", &hex(&self.prefix))
+            .field("prefix", &self.prefix)
             .field("vk", &self.vk)
             .finish()
     }
@@ -58,15 +53,9 @@ impl<'a> From<&'a SigningKey> for VerificationKeyBytes {
     }
 }
 
-impl AsRef<[u8]> for SigningKey {
-    fn as_ref(&self) -> &[u8] {
-        &self.seed[..]
-    }
-}
-
 impl From<SigningKey> for [u8; 32] {
     fn from(sk: SigningKey) -> [u8; 32] {
-        sk.seed
+        sk.to_bytes()
     }
 }
 
@@ -115,9 +104,9 @@ impl From<[u8; 32]> for SigningKey {
         };
 
         Self {
-            seed,
-            s,
-            prefix,
+            seed: Secret::new(seed),
+            s: Secret::new(s),
+            prefix: Secret::new(prefix),
             vk: VerificationKey {
                 minus_A: -A,
                 A_bytes: VerificationKeyBytes(A.compress().to_bytes()),
@@ -128,8 +117,9 @@ impl From<[u8; 32]> for SigningKey {
 
 impl zeroize::Zeroize for SigningKey {
     fn zeroize(&mut self) {
-        self.seed.zeroize();
-        self.s.zeroize()
+        self.seed = Secret::new([0; 32]);
+        self.s = Secret::new(Scalar::ZERO);
+        self.prefix = Secret::new([0; 32]);
     }
 }
 
@@ -144,7 +134,9 @@ impl SigningKey {
     /// Create a signature on `msg` using this key.
     #[allow(non_snake_case)]
     pub fn sign(&self, msg: &[u8]) -> Signature {
-        let r = Scalar::from_hash(Sha512::default().chain(&self.prefix[..]).chain(msg));
+        let r = self
+            .prefix
+            .expose(|prefix| Scalar::from_hash(Sha512::default().chain(&prefix[..]).chain(msg)));
 
         let R_bytes = (&r * constants::ED25519_BASEPOINT_TABLE)
             .compress()
@@ -157,8 +149,24 @@ impl SigningKey {
                 .chain(msg),
         );
 
-        let s_bytes = (r + k * self.s).to_bytes();
+        let s_bytes = self.s.expose(|s| (r + k * s).to_bytes());
 
         Signature { R_bytes, s_bytes }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use commonware_utils::hex;
+
+    #[test]
+    fn debug_redacts_private_material() {
+        let seed = [7u8; 32];
+        let key = SigningKey::from(seed);
+        let debug = format!("{key:?}");
+
+        assert!(debug.contains("REDACTED"));
+        assert!(!debug.contains(&hex(&seed)));
+    }
+}
diff --git a/cryptography/src/ed25519/scheme.rs b/cryptography/src/ed25519/scheme.rs
index 967f83c48..e54f8cf30 100644
--- a/cryptography/src/ed25519/scheme.rs
+++ b/cryptography/src/ed25519/scheme.rs
@@ -69,7 +69,7 @@ impl Random for PrivateKey {
 
 impl Write for PrivateKey {
-        self.key.expose(|key| key.as_bytes().write(buf));
+        self.key.expose(|key| key.to_bytes().write(buf));
     }
 }
 
@@ -117,7 +117,7 @@ impl arbitrary::Arbitrary<'_> for PrivateKey {
 impl PartialEq for PrivateKey {
     fn eq(&self, other: &Self) -> bool {
         self.key
-            .expose(|key1| other.key.expose(|key2| key1.as_bytes() == key2.as_bytes()))
+            .expose(|key1| other.key.expose(|key2| key1.to_bytes() == key2.to_bytes()))
     }
 }
 

```

cryptography/src/ed25519/core/signing_key.rs
patrick-ogrady: We can save this for #3668

cryptography/src/ed25519/core/verification_key.rs
patrick-ogrady: Wrote up in #3668

cryptography/src/ed25519/core/signing_key.rs
```diff
@@ -89,23 +89,17 @@ impl From<[u8; 32]> for SigningKey {
...
         scalar_bytes[..].copy_from_slice(&h.as_slice()[0..32]);
         scalar_bytes[0] &= 248;
         scalar_bytes[31] &= 127;
         scalar_bytes[31] |= 64;
 
-        #[expect(
-            deprecated,
-            reason = "Ed25519 key expansion uses the clamped bits directly for scalar-point multiplication."
-        )]
-        let point_scalar = Scalar::from_bits(scalar_bytes);
```
patrick-ogrady: Opted to just align with zebra: https://github.com/ZcashFoundation/ed25519-zebra/blob/3d5adb4b76be0272c909c373ee22f0c71dc331a8/src/signing_key.rs#L121-L129

patrick-ogrady: AFAICT this shouldn't actually affect the output and removes our reliance on an explicitly deprecated function.

patrick-ogrady: fwiw Zebra made this change in-place (no notice/warning/breaking): https://github.com/ZcashFoundation/ed25519-zebra/pull/88

patrick-ogrady: This is the line that is the key (as to why this isn't a breaking change):
```
scalar_bytes[31] &= 127;
```

PR #3617, opened 2026-04-17
[cryptography] Reuse decompressed Ed25519 batch keys

clabby:
## Overview

> [!NOTE]
>
> Stacked PR:
> - 👉 #3617
> - #3616
> - `main`

Modifies the batch verifier in our vendored version of `ed25519_consensus` to accept pre-decompressed verification keys. This prevents duplicate point decompression in the common case, e.g. when we've already received and decoded `ed25519::Signature`s over the wire. Passing raw bytes is still supported as a secondary path.

### DIFF
--- a/cryptography/src/ed25519/core/batch.rs
+++ b/cryptography/src/ed25519/core/batch.rs
@@ -34,7 +34,7 @@
 //!
 //! [ZIP215]: https://github.com/zcash/zips/blob/master/zip-0215.rst
 
-use super::{Error, Signature, VerificationKeyBytes};
+use super::{Error, Signature, VerificationKey};
 #[cfg(not(feature = "std"))]
 use alloc::{collections::BTreeMap as Map, vec::Vec};
 use curve25519_dalek::{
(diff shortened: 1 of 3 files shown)


cryptography/src/ed25519/ed25519_consensus/batch.rs
```diff
@@ -65,6 +65,15 @@ fn gen_u128<R: RngCore + CryptoRng>(mut rng: R) -> u128 {
     u128::from_le_bytes(bytes)
 }
 
+fn compute_k<M: AsRef<[u8]> + ?Sized>(vk_bytes: &[u8; 32], sig: &Signature, msg: &M) -> Scalar {
```
cronokirby: I would call this "challenge", and we also don't need to make it generic over M, we can call as_ref before this fn(which is more idiomatic, imo, and also reduces code size / compilation time)

cronokirby:
What if we just made it so that the batch verifier only accepts decompressed keys? That would simplify the code a decent amount.

You would have worse performance in that case that you create a batch but don't verify it, but that seems unrealistic?

cronokirby (dismissed):
Awesome!

patrick-ogrady (approved):

PR #3621, opened 2026-04-17
[storage/qmdb/sync] Generalize sync on merkle family - approach #2

roberto-bayardo:
This PR generalizes the sync protocol across merkle family.  To support the current DB's grafted digest reconstruction, we change the current qmdb pruning function to restrict pruning to a point that may precede the inactivity floor by a slight amount (< 2*chunk_size). This guarantees the grafted digests can always be reconstructed from ops-tree digests alone, as required by the sync protocol.

Alternative to: https://github.com/commonwarexyz/monorepo/pull/3609


roberto-bayardo:
closing in favor of #3621

PR #3629, opened 2026-04-20
[consensus] Final Certify Nits

patrick-ogrady:
Related: #3543

### DIFF
--- a/consensus/src/marshal/core/actor.rs
+++ b/consensus/src/marshal/core/actor.rs
@@ -524,16 +524,8 @@ where
                     Message::Proposed { round, block, ack } => {
                         self.cache_verified(round, block.digest(), block.clone())
                             .await;
-                        // Only signal success once the dissemination layer
-                        // accepts the block. Leaving `ack` unresolved when the
-                        // broadcaster is gone causes the waiting relay to see
-                        // a closed channel and abort the proposal, rather than
-                        // letting consensus vote on a block no peer received.
-                        if buffer.send(round, block, Recipients::All).await {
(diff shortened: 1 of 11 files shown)


consensus/src/marshal/core/actor.rs
```diff
@@ -491,7 +491,7 @@ where
             },
             // Handle consensus inputs before backfill or resolver traffic
             Some(message) = self.mailbox.recv() else {
-                debug!("mailbox closed, shutting down");
+                info!("mailbox closed, shutting down");
```
patrick-ogrady: put this back to debug

consensus/src/marshal/core/actor.rs
```diff
@@ -542,6 +550,15 @@ where
                         };
                         buffer.send(round, block, Recipients::Some(peers)).await;
                     }
+                    // Both handlers ack unconditionally even when the round
```
patrick-ogrady: add to certified directly

consensus/src/marshal/core/variant.rs
```diff
@@ -142,12 +142,18 @@ pub trait Buffer<V: Variant>: Clone + Send + Sync + 'static {
     fn finalized(&self, commitment: V::Commitment) -> impl Future<Output = ()> + Send;
 
     /// Send a block to peers.
+    ///
+    /// Returns `true` when the dissemination layer accepted the block for
```
patrick-ogrady: I think this is actually an anti-pattern. We shouldn't allow send to have a true/false here.

PR #3635, opened 2026-04-20
[Storage] PR 3627 feedback

danlaine:
Follow-up on review feedback for PR #3627, which replaces the Current QMDB's layered bitmap with `Arc<SharedBitmap<RwLock<BitMap>>>`. This PR addresses two small inefficiencies and a coverage gap surfaced in review, without changing the PR's design.
### 1. Cache terminal `Arc<SharedBitmap>` on `BitmapBatchLayer` (`batch.rs`)

Add `shared: Arc<SharedBitmap<N>>` to `BitmapBatchLayer`. Collapses:

- `BitmapBatch::shared()` from an O(chain-depth) walk to a 2-arm match.
- `BitmapBatch::pruned_chunks()` from an O(chain-depth) walk to `self.shared().pruned_chunks()`.

Cost: 8 bytes per layer + one `Arc::clone` at construction. The invariant (cached `shared` equals the chain terminal) is trivial by construction at both sites (`compute_current_layer`, `trim_committed`) — each sets the field via `Arc::clone(parent.shared())`, and `BitmapBatch::Base` is already the terminal.

Both callers (`trim_committed` on every `new_batch`; `pruned_chunks` once per `build_chunk_overlay`) are on the per-batch hot path. The walk was redundant after PR #3627's restructuring and is now gone.

### 2. Add `test_current_live_batch_child_after_prune` (`mod.rs`)

Covers the specific scenario: a live parent batch whose `new_batch()` is called after the DB has advanced its pruning boundary. The internal `trim_committed` must observe the advanced boundary on the shared bitmap, and the resulting child must merkleize and apply correctly.

Not covered by the existing `test_current_live_batch_safe_across_prune`, which tests *applying a pre-prune child* rather than *extending a live parent post-prune*.

### DIFF
--- a/consensus/fuzz/src/lib.rs
+++ b/consensus/fuzz/src/lib.rs
@@ -375,7 +375,7 @@ where
         propose_latency: (10.0, 5.0),
         verify_latency: (10.0, 5.0),
         certify_latency: (10.0, 5.0),
-        should_certify: application::Certifier::Sometimes,
+        should_certify: application::Certifier::Always,
     };
     let (actor, application) =
         application::Application::new(context.with_label("application"), app_cfg);
@@ -609,7 +609,7 @@ fn run_with_twin_mutator<P: simplex::Simplex>(input: FuzzInput) {
                 propose_latency: (10.0, 5.0),
                 verify_latency: (10.0, 5.0),
                 certify_latency: (10.0, 5.0),
-                should_certify: application::Certifier::Sometimes,
+                should_certify: application::Certifier::Always,
             };
             let (actor, application) =
                 application::Application::new(primary_context.with_label("application"), app_cfg);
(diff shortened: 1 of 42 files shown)


roberto-bayardo:
re: benchmark, I was going to remove it from this PR in favor of the standalone https://github.com/commonwarexyz/monorepo/pull/3633

...just in case we come up with a non-locking based solution to this issue.

So could you move the benchmark suggestions there?

danlaine:
> re: benchmark, I was going to remove it from this PR in favor of the standalone #3633
> 
> ...just in case we come up with a non-locking based solution to this issue.
> 
> So could you move the benchmark suggestions there?

Removed benchmark changes from here

roberto-bayardo (approved):

roberto-bayardo:
apparently I neglected to merge this when I approved it and now there are conflicts :-(

PR #3636, opened 2026-04-20
[runtime] Create a trait for briding async and parallel code

cronokirby:
When calling into code doing heavy computation using a Strategy, you want to avoid blocking the tokio thread on waiting for the computation. Instead, you want the waiting to be async, so that other tasks can use the thread while you wait for the computation to finish in the rayon thread pool.

To facilitate this pattern, a new capability, Strategist, is added to to the runtime Context. This takes in a synchronous closure, which needs a strategy to do parallel computation. Runtimes implement this correctly, by using, e.g. a oneshot channel to asynchronously wait on the rayon result, or using the tokio blocking pool for a sequential strategy.

To use this change throughout the code base, places where a Strategy was stored have been replaced with using the context instead.


cronokirby:
We could also get rid of the `ThreadPooler` API entirely, but that would require modifying some code in Storage to be over strategy instead. Probably worth doing if we merge this, but subsequently.

consensus/src/aggregation/engine.rs
```diff
@@ -530,9 +525,16 @@ impl<
         let filtered = acks
             .values()
             .filter(|a| a.item.digest == ack.item.digest)
+            .cloned()
             .collect::<Vec<_>>();
         if filtered.len() >= quorum as usize {
-            if let Some(certificate) = Certificate::from_acks(&*scheme, filtered, &self.strategy) {
+            let certificate = self
+                .context
```
patrick-ogrady, line 532: I think we should keep `strategy` less tightly bound to context? We could accept strategy with it?

runtime/src/tokio/runtime.rs
```diff
@@ -691,6 +727,38 @@ impl crate::ThreadPooler for Context {
...
+            }
+            RuntimeStrategy::Rayon(strategy) => {
+                let (sender, receiver) = oneshot::channel();
+                let pool = strategy.thread_pool().clone();
+                let strategy = RuntimeStrategy::Rayon(strategy);
+                pool.spawn(move || {
+                    let _ = sender.send(f(&strategy));
+                });
+                Either::Right(async move { receiver.await.expect("strategy task failed") })
+            }
```
michaelkaplan13, line 757:
```suggestion
            RuntimeStrategy::Sequential(_) => {
                let handle = self
                    .clone()
                    .shared(true)
                    .spawn(move |_| async move { f(&strategy) });
                Either::Left(async move { handle.await.expect("strategy task failed") })
            }
            RuntimeStrategy::Rayon(ref rayon) => {
                let (sender, receiver) = oneshot::channel();
                let pool = rayon.thread_pool().clone();
                pool.spawn(move || {
                    let _ = sender.send(f(&strategy));
                });
                Either::Right(async move { receiver.await.expect("strategy task failed") })
            }
```

I was looking at this block for a while trying to internalize exactly what's happening here, and in playing around with it I thought this made it easier to read. Feel free to ignore if this is not as idiomatic though.

(I did not know the `ref` part would be needed / had never seen the "use of partially moved value" compiler error that occurs with out.)

consensus/src/aggregation/engine.rs
```diff
@@ -530,9 +525,16 @@ impl<
         let filtered = acks
             .values()
             .filter(|a| a.item.digest == ack.item.digest)
+            .cloned()
```
patrick-ogrady, line 528: wonder if there is a way to avoid this clone with some more careful lifetime management?

PR #3637, opened 2026-04-20
[storage] add get_many batch read API to qmdb variants

clabby:
## Overview

Adds `get_many` functions to all QMDB variants' batch types.

### DIFF
--- a/storage/src/qmdb/any/batch.rs
+++ b/storage/src/qmdb/any/batch.rs
@@ -789,6 +789,72 @@ where
         }
         db.get(key).await
     }
+
+    /// Read multiple keys, amortizing DB lock acquisition for fallthrough reads.
+    ///
+    /// Returns results in the same order as the input keys.
+    pub async fn get_many<E, C, I>(
+        &self,
+        keys: &[&U::Key],
+        db: &Db<F, E, C, I, H, U>,
+    ) -> Result<Vec<Option<U::Value>>, crate::qmdb::Error<F>>
+    where
+        E: Context,
+        C: Contiguous<Item = Operation<F, U>>,
+        I: UnorderedIndex<Value = Location<F>> + 'static,
+    {
+        if keys.is_empty() {
+            return Ok(Vec::new());
+        }
+
+        let mut results: Vec<Option<U::Value>> = Vec::with_capacity(keys.len());
+        let mut db_indices = Vec::new();
+        let mut db_keys = Vec::new();
+
+        for (i, key) in keys.iter().enumerate() {
+            // Check local mutations.
+            if let Some(value) = self.mutations.get(*key) {
+                results.push(value.clone());
+                continue;
+            }
+
+            // Check parent diff chain.
+            let mut found = false;
+            if let Some(parent) = self.base.parent() {
+                if let Some(entry) = lookup_sorted(parent.diff.as_slice(), *key) {
+                    results.push(entry.value().cloned());
+                    found = true;
+                }
+                if !found {
+                    for batch in parent.ancestors() {
+                        if let Some(entry) = lookup_sorted(batch.diff.as_slice(), *key) {
+                            results.push(entry.value().cloned());
+                            found = true;
+                            break;
+                        }
+                    }
+                }
+            }
+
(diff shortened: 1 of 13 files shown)


storage/src/qmdb/any/db.rs
```diff
@@ -138,6 +138,65 @@ where
         Ok(None)
     }
 
+    /// Get values of multiple keys, amortizing reader lock acquisition and journal I/O.
+    ///
+    /// Returns results in the same order as the input keys.
+    pub async fn get_many(
```
clabby, line 151: For this (and other DBs' `get_many` impls), it'd be awesome if we could parallelize reads well... `FuturesOrdered` performance here isn't great in the case where we're predominantly hitting the page cache.

storage/src/qmdb/immutable/batch.rs
```diff
@@ -163,6 +163,73 @@ where
         db.get(key).await
     }
 
+    /// Read multiple keys, amortizing DB lock acquisition for fallthrough reads.
```
roberto-bayardo: I wouldn't let the implementation leak here, would just say something like "Batch read multiple keys."  (Plus you're doing more than just lock amortization...)

storage/src/qmdb/immutable/batch.rs
```diff
@@ -267,6 +334,66 @@ where
...
```
roberto-bayardo: same concern around leaking implementation here.

storage/src/qmdb/immutable/mod.rs
```diff
@@ -212,6 +212,62 @@ where
         Ok(None)
     }
 
+    /// Get values of multiple keys, amortizing reader lock acquisition and journal I/O.
```
roberto-bayardo: same here

storage/src/qmdb/any/batch.rs
```diff
@@ -785,6 +785,72 @@ where
...
+            }
+
+            if found {
+                continue;
+            }
+
+            // Need DB fallthrough.
+            db_indices.push(i);
+            db_keys.push(*key);
+            results.push(None); // placeholder
```
roberto-bayardo: what do you mean by // placeholder?

storage/src/qmdb/keyless/batch.rs
```diff
@@ -183,6 +183,66 @@ where
         db.get(loc).await
     }
 
+    /// Read values at multiple locations.
+    ///
+    /// Acquires the journal reader once for all DB-fallthrough reads.
+    /// Locations must be sorted in ascending order.
```
roberto-bayardo, line 196: Can you add a debug_assert! for this property?  (Same for the other functions in this PR requiring sorted input..)

roberto-bayardo:
Please add a few tests?

storage/src/qmdb/immutable/batch.rs
clabby: Updated

storage/src/qmdb/any/batch.rs
clabby: These are set in a follow-up block. We add `None` as a placeholder before then.

Removed in case it was confusing.

danlaine (dismissed):

storage/src/qmdb/immutable/mod.rs
```diff
@@ -263,6 +263,62 @@ where
...
+        for &(key_idx, pos) in &candidates {
+            if results[key_idx].is_some() {
+                continue;
+            }
+            let op_idx = positions
+                .binary_search(&pos)
+                .expect("position was deduped from candidates");
+            let Operation::Set(k, v) = &ops[op_idx] else {
+                continue;
+            };
```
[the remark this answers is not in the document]
danlaine: Hmm I suppose we should return an error in `get_many` to be consistent (even though users shouldn't pass invalid locations in the first place)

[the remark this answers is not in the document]
clabby: Updated and added a test.

danlaine (approved):

PR #3644, opened 2026-04-21
[workspace] clippy and fmt

andresilva:


storage/src/qmdb/any/batch.rs
```diff
@@ -1688,7 +1688,7 @@ mod tests {
                 },
             ),
         ];
-        base_diff.sort_by(|a, b| a.0.cmp(&b.0));
+        base_diff.sort_by_key(|a| a.0);
```
andresilva, line 1691: clippy complains about this (same below)

patrick-ogrady (approved):

PR #3647, opened 2026-04-22
[consensus] Inferred Certification

patrick-ogrady:
Related: #3433 

Open question: should we emit a finalize vote if we infer certification (we only emit a notarize vote if we explicitly verify but the only way that is cancelled is with a notarization...with this, a future notarization can cause us to infer so there is a chance we don't form finalizations)?

### DIFF
--- a/consensus/src/simplex/actors/voter/actor.rs
+++ b/consensus/src/simplex/actors/voter/actor.rs
@@ -543,34 +543,15 @@ impl<
 
     /// Infer certifications for rounds whose notarizations prove certification without
     /// requiring an automaton response, then journal and signal them.
-    ///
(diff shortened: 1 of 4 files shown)


0xAysh:
I think yes we should emit finalization vote. With current code finalization path does not require a view to be explicitly verified (it just reqiures the view to not be already finalized/nullified locally, and to have natorization+certificateation for the view). So if inference marks a view as certified, emitting finalization certification seems consistent and helps liveness. I also just thought to call verify fn on inferred views but try_verify() is scoped only for current views not for ancestors and we deviate from the core problem if we try to adjust the scope of it. Also I skimmed your implementation and would love to understand your approach and thought process.

PR #3650, opened 2026-04-22
[Storage] Add "compact" `qmdb::keyless`, `qmdb::immutable`, persisted merkle implementations

### LINKED ISSUE #3598: [storage/qmdb] Add Support for "Peak Only Writes"
Related: https://github.com/commonwarexyz/monorepo/issues/3538

In some applications, we never need to query or help others state sync the operations (say transactions from finalized blocks). We could save a considerable amount of DiskIO by only writing the peaks on finalization rather than all operations (which we'll immediately prune).

danlaine:
Resolves #3598

## Summary

This PR introduces compact-storage QMDBs for `immutable` and `keyless`, adds a dedicated compact sync path for them, and completes the terminology rename from `journaled` / `unjournaled` to `full` / `compact`.

The main goal is to let compact databases converge to the same authenticated root as full databases without forcing receivers to download and retain the full operation log and full Merkle state.

At a high level:

- `full` means a database retains authenticated operation history and can serve replay-based full sync.
- `compact` means a database retains only the latest authenticated state needed to continue operating and to serve compact sync.
- compact sync transfers authenticated state directly instead of replaying retained operations.

## Terminology and public rename

The rename is broad and intentional:

- Merkle modules:
  - `merkle::journaled` -> `merkle::persisted::full`
  - `merkle::{mmr,mmb}::journaled` -> `merkle::{mmr,mmb}::full`
- QMDB terminology:
  - `journaled` -> `full`
  - `unjournaled` -> `compact`
- Example sync CLI:
  - sync mode names are now `full` and `compact`
  - the example CLI now separates logical database family from compact-server backing storage

The rename was propagated through storage modules, docs, examples, fuzz targets, and conformance fixtures.

## Compact databases

This PR adds compact-storage variants for both append-only QMDB families:

- `qmdb::immutable::CompactDb`
- `qmdb::keyless::CompactDb`

These databases are for nodes that want the current authenticated state without retaining the full historical operation log locally.

A compact database persists:

- the compact Merkle frontier
- the latest committed operation
- the witness/proof data needed to authenticate that latest commit against the current root

It does **not** retain the historical operation sequence required for full sync into a full database.

The underlying persistence pieces are:

- compact Merkle persistence in `storage/src/merkle/persisted/compact.rs`
- shared witness handling in `storage/src/qmdb/compact_witness.rs`

That witness layer exists because the Merkle layer knows frontier state and hashes, but the QMDB layer is responsible for encoding and decoding the typed final commit operation that compact sync must transfer.

The compact persistence model is designed so that:

- frontier and witness advance together
- reopen and rewind restore a matched pair
- corrupted or missing witness data fails loudly rather than being served

Compact databases also support rewind using the same slot-backed persistence model, so compact state, witness data, and reopen/recovery behavior stay aligned.

## Compact sync

This PR adds a new sync module:

- `qmdb::sync::compact`

This path is intentionally separate from full sync.

Full sync continues to use:

- `qmdb::sync::Target`
- `qmdb::sync::sync(...)`

Compact sync uses:

- `qmdb::sync::compact::Target`
- `qmdb::sync::compact::State`
- `qmdb::sync::compact::Config`
- `qmdb::sync::compact::sync(...)`

### `compact::Target`

Compact sync now has its own target type:

- `qmdb::sync::compact::Target { root, leaf_count }`

This split matches the actual authentication contract:

- full sync authenticates and advances a retained operation range
- compact sync authenticates a root and total leaf count

There is no replay lower bound in compact sync because compact sync does not reconstruct local history.

### Compact sync payload

Compact sync transfers the minimum authenticated state needed to reconstruct a compact target:

- `root`
- `leaf_count`
- `pinned_nodes`
- `last_commit_op`
- `last_commit_proof`

The receiver:

1. verifies the final commit proof against the requested root
2. rebuilds compact state from the supplied frontier
3. recomputes the local root
4. persists the reconstructed state only after the root matches

This ordering matters: it prevents a source from causing bad frontier state to be durably written before the receiver has authenticated the reconstructed root.

### Full-source serving and stale targets

Compact sync sources are intentionally broader than compact sync targets.

For compact serving, both of these can act as sources:

- full `immutable` / `keyless`
- compact `immutable::CompactDb` / `keyless::CompactDb`

The source-side behavior differs by source type:

- full databases derive compact state from retained history by fetching the latest commit and matching frontier data
- compact databases serve the persisted frontier and witness directly

That means compact sync supports:

- full -> compact
- compact -> compact

It does **not** support:

- compact -> full

because compact storage does not retain the historical operations required for full sync.

This PR also tightens full-source serving semantics:

- compact full sources now reject stale targets instead of serving state for a source that has already advanced
- `qmdb::sync::compact::Target` is validated before use
- compact sync rejects mismatched `leaf_count`, invalid final-commit proofs, and reconstructed-root mismatches before persisting anything

## `examples/sync`

The sync example was expanded to demonstrate both full sync and compact sync.

### Modes

- `--mode full`
- `--mode compact`

### Family and storage flags

The example CLI now separates the logical database family from the compact server's backing storage:

- `--family <any|current|immutable|keyless>`
- server-only `--storage <full|compact>` in compact mode

Full mode supports these families:

- `any`
- `current`
- `immutable`
- `keyless`

Compact mode supports these families:

- `immutable`
- `keyless`

The old `--db` spelling and the public `immutable-compact` / `keyless-compact` CLI values were removed so the example exposes only the family/storage split it actually demonstrates.

### Important compact-mode behavior

The compact example is intentionally asymmetric between server and client.

On the compact **server** side:

- `--family immutable|keyless --storage full` means a full database serving compact state
- `--family immutable|keyless --storage compact` means a compact database serving compact state

On the compact **client** side:

- the client always materializes a compact-storage target
- `--family immutable` targets compact immutable storage
- `--family keyless` targets compact keyless storage

The example protocol was updated to reflect that split:

- compact target fetches are distinct from full-sync target fetches
- compact mode uses `qmdb::sync::compact::Target`
- compact authenticated state is fetched over a dedicated request/response path
- the protocol now includes a `StaleTarget` error so compact clients can retry when a target goes stale between discovery and state fetch

### Example database growth behavior

The example database adapters now recompute commit floors from live database state where needed when appending additional demo operations onto an already-running database.

This keeps repeated server-side growth monotonic and fixes the regression where later keyless demo batches could try to reuse stale floor values and fail with `FloorRegressed`.

## Testing and validation

The supporting test surface was expanded to cover the new behavior, including:

- compact sync roundtrips for immutable and keyless
- both `MMR` and `MMB` compact coverage
- compact source reopen / rewind / regrow behavior
- invalid proof rejection
- mismatched `leaf_count` rejection
- tampered `pinned_nodes` rejection without persisting bad state
- stale-target rejection for compact full sources
- compact witness corruption handling
- `examples/sync` protocol and mode coverage


storage/src/merkle/persisted/compact.rs
```diff
@@ -0,0 +1,348 @@
...
+//! Unlike [`crate::merkle::journaled`], this type persists only the minimum state required to
+//! recover the current root and continue appending after restart. Historical nodes are discarded
+//! on sync and are not readable after reopen.
+//!
+//! # Why peaks are enough
+//!
+//! An MMR/MMB root is computed from the current peaks, and appending a new leaf only touches
+//! peaks. Persisting `(leaf_count, pinned_peaks)` and rebuilding [`Mem`] on reopen with no
+//! retained nodes and those peaks as pinned values reconstructs an equivalent tree: same root,
+//! same future append behavior.
```
roberto-bayardo, line 12: Do we not have the same issue as sync protocol when using an MMB where you can't necessarily recompute the root from peaks only at the pruning boundary after growth?

danlaine: I don't think the same concern applies here. This type doesn't support creating proofs, so we never need potentially pruned nodes to create proofs. The root is computable from the peaks.

storage/src/qmdb/immutable/compact.rs
```diff
@@ -0,0 +1,840 @@
...
+                tip_commit_loc,
+            ));
+        }
+
+        let start_loc = self.last_commit_loc + 1;
+        self.merkle.apply_batch(&batch.merkle_batch)?;
+        self.last_commit_loc = Location::new(batch.total_size - 1);
+        self.last_commit_metadata = batch.commit_metadata.clone();
+        self.inactivity_floor_loc = batch.new_inactivity_floor_loc;
+        Ok(start_loc..Location::new(batch.total_size))
```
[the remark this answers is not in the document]
danlaine: I think this is accurate but not introduced in this PR. Will make a separate PR.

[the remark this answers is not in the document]
danlaine: https://github.com/commonwarexyz/monorepo/pull/3654

storage/src/merkle/persisted/compact.rs
roberto-bayardo: Oh right, I was confusing the situation with grafted MMBs. With a single MMB this is all fine.

storage/src/qmdb/immutable/mod.rs
```diff
@@ -103,9 +103,11 @@ pub mod batch;
 pub mod fixed;
 mod operation;
 pub mod sync;
+mod unjournaled;
 pub mod variable;
 
 pub use operation::Operation;
+pub use unjournaled::{Config as UnjournaledConfig, Db as UnjournaledDb};
```
clabby: Will need the batch types exported for `glue`.

storage/src/qmdb/keyless/mod.rs
```diff
@@ -62,8 +62,10 @@ pub mod batch;
 pub mod fixed;
 mod operation;
 pub(crate) mod sync;
+mod unjournaled;
 pub mod variable;
 pub use operation::Operation;
+pub use unjournaled::{Config as UnjournaledConfig, Db as UnjournaledDb};
```
clabby: Will need the batch types exported for `glue`.

storage/src/qmdb/sync/error.rs
```diff
@@ -29,6 +41,12 @@ pub enum EngineError<F: Family, D: Digest> {
         old: Target<F, D>,
         new: Target<F, D>,
     },
+    /// Sync target root changed without growing the tree.
+    #[error("sync target root changed without end growth: {old:?} -> {new:?}")]
+    SyncTargetRootChangedWithoutGrowth {
```
roberto-bayardo: This variant seems unused?

examples/sync/src/databases/immutable_compact.rs
```diff
@@ -0,0 +1,93 @@
...
+        }
+
+        let mut batch = self.new_batch();
+        for operation in operations {
+            match operation {
+                Operation::Set(key, value) => {
+                    batch = batch.set(key, value);
+                }
+                Operation::Commit(metadata, _floor) => {
+                    let floor = self.last_commit_loc();
```
roberto-bayardo: Should this be:
```suggestion
                    let floor = self.inactivity_floor_loc();
```

examples/sync/src/databases/keyless.rs
```diff
@@ -57,10 +58,13 @@ pub fn create_config(context: &(impl BufferPooler + commonware_runtime::Metrics)
...
-/// floor at each commit to the *previous* commit's location. This models a realistic
-/// application that declares older commits inactive over time (enabling pruning) while
-/// always keeping the most recent commit readable.
+/// Generates Append operations and periodic Commit operations for a freshly initialized
+/// example database.
+///
+/// The commit floors in this returned stream are exact for bootstrapping from the initial
+/// commit at location 0. When later appending demo operations onto an already-running
+/// example database, `add_operations` recomputes each commit's floor from the live db
+/// state so repeated server-side growth keeps the floor monotonic.
```
roberto-bayardo: I don't understand what this is saying. Seems you are still setting commit floor to last_commit_loc, so feels like the previous comment above was better?

danlaine: Updated comments

examples/sync/src/databases/immutable_compact.rs
danlaine: Updated this code so now we actually use the floor from the `Commit` and this line is gone

storage/src/qmdb/sync/error.rs
danlaine: removed

roberto-bayardo:
LGTM, though Codex 5.5 had a good last suggestion: worth adding a regression test that applies the same generated immutable operations to full and compact across two growth cycles and asserts equal roots/floors.

Also there is a small conflict.

danlaine:
> LGTM, though Codex 5.5 had a good last suggestion: worth adding a regression test that applies the same generated immutable operations to full and compact across two growth cycles and asserts equal roots/floors.
> 
> Also there is a small conflict.

Done and resolved

roberto-bayardo (approved):

PR #3651, opened 2026-04-22
[storage] Generalize benchmarks on Merkle family

roberto-bayardo:
- generalized qmdb benchmarks to run over variants that include both merkle families
- tweaked the name of the variants that use variable vector values to avoid confusion with the variants that use fixed values with a variable-capable database (`variable-vec` vs plan `variable`)

### DIFF
--- a/storage/src/qmdb/benches/common.rs
+++ b/storage/src/qmdb/benches/common.rs
@@ -5,10 +5,7 @@ use commonware_cryptography::{Hasher, Sha256};
 use commonware_runtime::{buffer::paged::CacheRef, tokio::Context, BufferPooler, ThreadPooler};
 use commonware_storage::{
     journal::contiguous::{fixed::Config as FConfig, variable::Config as VConfig},
-    merkle::{
-        self,
-        mmr::{journaled::Config as MmrConfig, Family},
-    },
+    merkle::{self, journaled::Config as MerkleConfig, Family},
     qmdb::{
         any::{
             ordered::{fixed::Db as OFixed, variable::Db as OVariable},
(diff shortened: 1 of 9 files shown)


storage/src/qmdb/benches/generate.rs
```diff
@@ -22,15 +25,15 @@ const COMMITS_PER_ITERATION: u64 = 100;
 
 /// Benchmark a populated database: generate data, prune, sync. Returns elapsed time (excluding
 /// destroy).
-async fn bench_db<C: DbAny<commonware_storage::merkle::mmr::Family, Key = Digest>>(
+async fn bench_db<F: Graftable, C: DbAny<F, Key = Digest>>(
```
danlaine: I think we can loosen this bound?
```suggestion
async fn bench_db<F: Family, C: DbAny<F, Key = Digest>>(
```

storage/src/qmdb/benches/init.rs
```diff
@@ -107,7 +119,12 @@ fn bench_fixed_value_init(c: &mut Criterion) {
     }
 }
 
-fn bench_var_value_init(c: &mut Criterion) {
+fn bench_fixed_value_init(c: &mut Criterion) {
+    bench_fixed_value_init_family::<mmr::Family>(c, "mmr");
```
danlaine: Not an issue per se but just flagging that this will ~2x the run time so we should expect that in results

storage/src/qmdb/benches/generate.rs
danlaine: I think in some other places also

storage/src/qmdb/benches/init.rs
roberto-bayardo: bah, yeah this wasn't intended, fixing.

storage/src/qmdb/benches/generate.rs
roberto-bayardo: right, fixed.

storage/src/qmdb/benches/init.rs
roberto-bayardo: ok restructured things a bit to all rely on the variant! macros

storage/src/qmdb/benches/common.rs
```diff
@@ -42,115 +39,47 @@ pub const WRITE_BUFFER_SIZE: NonZeroUsize = NZUsize!(1024);
...
-fn mmr_cfg(
+fn merkle_cfg(
     suffix: &str,
     ctx: &(impl BufferPooler + ThreadPooler),
     page_cache: CacheRef,
-) -> MmrConfig {
-    MmrConfig {
+) -> MerkleConfig {
+    MerkleConfig {
         journal_partition: format!("journal-{suffix}"),
```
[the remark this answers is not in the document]
roberto-bayardo: this is fine, partition names don't have to be unique per variant. DBs always start from a clean state regardless.

clabby (approved):

PR #3652, opened 2026-04-22
changes paper link to eprint

guruvamsi-policharla:


docs/blogs/bte.md
patrick-ogrady, line 182: I think you need to work in your latest changes? This is already in main.

PR #3654, opened 2026-04-23
[Storage] Fix `qmdb::immutable` chained batch floor validation

danlaine:
## Summary

Fix immutable QMDB chained batch floor validation.

Before this change, `qmdb::immutable::Immutable::apply_batch` only validated the tip batch's `inactivity_floor`. In a chained batch, unapplied ancestor batches can each contribute their own `Commit` op, and those ancestor floors were not checked. That allowed a child batch to commit an invalid ancestor floor into the authenticated journal as long as the child's own floor was valid.

This change makes immutable batch validation match the existing per-commit semantics already used by rewind and by the keyless implementation: every unapplied ancestor commit in the chain must have a monotonic floor, and each floor must be at most that ancestor commit's own location.

## What changed

- Extend immutable `MerkleizedBatch` to carry each ancestor batch's declared commit floor.
- Validate ancestor floors oldest-first in `Immutable::apply_batch` before mutating journal or snapshot state.
- Keep the existing tip-floor validation, but compare it against the last validated ancestor floor rather than only the database's current floor.
- Add a regression test that constructs:
  1. a committed batch that raises the db floor,
  2. an unapplied parent batch with a regressed floor,
  3. a child batch with a valid tip floor,
  and asserts that applying the child is rejected with `Error::FloorRegressed`.

## Why this matters

Immutable chained batches are committed through the authenticated journal root-to-tip. If an ancestor batch carries an invalid floor, that ancestor commit is still appended to the real journal when the child is applied. Later operations like rewind read floors back from committed history, so accepting an invalid ancestor floor can surface as invalid live state later.

This fix closes that hole by enforcing the floor invariant at the point where the chained batch is accepted.


roberto-bayardo:
Heh, I had claude vibe code a fix for this not realizing you had already done it.   It seems my instance is a bit jealous though, and it pointed out a few nits.  Some seem reasonable?

1. Missing the FloorBeyondSize ancestor test entirely. This PR only covers FloorRegressed. The bug report's verified PoC is the beyond-commit case — parent floor=3, commit_loc=2, child floor=3,      
  commit_loc=4, expected FloorBeyondSize(3, 2). That path is now untested. It's also the more dangerous variant since the affected floor poisons historical_proof and rewind even when monotonicity is
  preserved.                                                                                                                                                                                             
  2. Only registered for fixed+MMR. The existing floor-related tests follow a 4-way {fixed, variable} × {mmr, mmb} pattern. The new test exists in only 1 of 4 cells. MMB and variable.rs are uncovered.
  3. Test doesn't isolate the ancestor path. It prelude-applies a floor=2 batch to advance the live floor, then builds parent floor=1, child floor=6. Parent's floor=1 < live_floor=2 would trip the very
   first iteration of the loop — but it would also trip the old tip-only check if apply_batch were called on the parent directly. The test fails before the fix because the parent's commit op gets      
  written via the child, but it doesn't cleanly demonstrate "the ancestor's FloorRegressed check fired because we now check ancestors". A purer scenario (parent advances live floor; child regresses    
  below parent but still ≥ live) would prove the cross-batch monotonicity step is what caught it.                                                                                                        
  4. No "DB state untouched" assertions. The keyless analog (and my version) snapshots root / last_commit_loc / inactivity_floor_loc before the failing call and asserts they're unchanged after. This
  guards against a future refactor that errors after mutating state. The PR just checks the error matches.                                                                                               
  5. No # Errors doc update on apply_batch. The doc still reads as if only the tip is checked; it should mention that ancestors are validated too and that the database is untouched on floor errors
  (mirroring keyless's wording).                                                                                                                                                                         
  6. No inline comment on the new loop. Three or four lines explaining the per-commit invariant ("floors are monotonically non-decreasing… each floor ≤ its own commit_loc… ancestors stored newest-first
   so iterate in reverse") would make future maintenance easier.

danlaine:
@roberto-bayardo Addressed all your comments

roberto-bayardo (approved):

PR #3662, opened 2026-04-24
[Storage] Use bitmap for merkleization in `qmdb::any`

### LINKED ISSUE #1829: [storage/adb/any] Maintain non-merkelized bitmap to use in floor raising

danlaine:
Closes #1829.

## Summary

This PR accelerates inactivity-floor raising in `qmdb::any` with a committed-state activity bitmap. During floor raising, `qmdb::any` now jumps between active committed operations using `ones_iter_from()` instead of treating every retained log location as a candidate.

The bitmap is owned by `any::Db.bitmap` (an `Arc<Shared<N>>`) and shared with `qmdb::current`, which reads it for grafted-tree leaves and proofs. For `any` alone the bitmap is an internal optimization: not authenticated, not persisted separately, rebuilt from the retained journal on init. For `current`, the bitmap chunks feed into the canonical root via the grafted Merkle tree.

## Motivation

`qmdb::any` advances its inactivity floor during batch merkleization. Previously, every location from the current floor up to the fixed tip was a candidate. The floor-raise loop read each operation and rejected it if the key had since been superseded.

That is correct but expensive when the retained log contains many stale updates. In update-heavy workloads, most operations below the tip are inactive, so the scan can spend most of its time reading operations that cannot advance the floor.

The scan range naturally splits into two regions:

- **Committed** (within `bitmap.len()`): indexed by the bitmap.
- **Uncommitted tail** (beyond `bitmap.len()`): in-memory ancestor batch operations not yet tracked. Handled conservatively with sequential candidates.

The existing `is_active_at()` post-filter rejects false-positive candidates, so bitmap maintenance only has to avoid false negatives (a still-active location whose bit was prematurely cleared would lose the move and corrupt the floor).

## Design

The bitmap is a `Prunable<N>` with one bit per log operation in the committed state. Standalone `any::Db` defaults to `N = BITMAP_CHUNK_BYTES` (64 bytes, 512 bits per chunk); `current::Db` overrides `N` to match its grafted-tree configuration.

- `1` for an active `Update` whose key currently resolves to that location.
- `1` for the *current* `last_commit_loc` only — earlier (intermediate) `CommitFloor`s carry `0`. This single-CommitFloor exception keeps the chunk containing the latest commit deterministic across init and `apply_batch`.
- `0` for superseded `Update`s, all `Delete`s, and all intermediate `CommitFloor`s.

A private helper supplies floor-raise candidates during merkleization:

- In `[floor, min(bitmap.len(), tip))`, it uses `ones_iter_from()` to jump to the next set bit.
- In `[bitmap.len(), tip)`, it returns sequential candidates and relies on `is_active_at()` to filter stale ones.

The any-side scan helper is independent of `qmdb::current`'s `BitmapBatch` overlay chain (used for speculative reads during merkleize). The bitmap *data* is shared between layers; the scan *strategies* are layer-specific.

## Bitmap Maintenance

`Db::apply_batch()` mirrors snapshot transitions into the bitmap under a single write lock, after the journal batch is applied:

- Extends the bitmap to `new_last_commit_loc + 1` (zero-fill).
- Sets winning active locations to `1`, clears superseded locations to `0`.
- Demotes the previous CommitFloor bit (now intermediate) to `0`, promotes the new one to `1`.

When a child batch implicitly commits one or more parent batches, bitmap updates follow the same precedence as the snapshot update: child wins first, then nearest uncommitted ancestor, with already-committed ancestor state tracked through `committed_locs`.

`Db::rewind()` truncates the bitmap, restores previously-superseded committed locations to `1` while applying snapshot undos, and sets the rewound tail's CommitFloor bit to `1`.

`Db::prune()` is split into `prune_log` and `prune_bitmap`:

- Standalone `any::Db::prune` calls them back-to-back, both targeting `prune_loc`.
- `current::Db::prune` interleaves: `prune_bitmap(sync_boundary)` → grafted-tree prune → `sync_metadata` → `prune_log(prune_loc)`. The order matters for crash safety: metadata must be persisted before the log is pruned, and bitmap pruning is more aggressive than log pruning when wrapped by `current`.

`Db::init_from_log()` rebuilds the bitmap from the retained journal alongside the snapshot, then checks that `bitmap.len() == log.size()`.

## Related cleanups

- Removes `current::Db.status` (was a separate `Arc` to a duplicate bitmap). `current` now reads through `any.bitmap` directly.
- Removes the chunk-overlay rewrite path in `current::Db::apply_batch`. Bitmap mutation is fully diff-driven in `any::Db::apply_batch`, so the bytes flow consistently through to `current` reads.
- Removes the dead public/init `callback`/`known_inactivity_floor` plumbing on `init`/`init_with_bitmap`/`init_from_log` and the four vestigial `init_with_callback` wrappers per variant. (`build_snapshot_from_log` still takes an internal closure for the in-process replay step.)

### DIFF
--- a/storage/src/qmdb/any/batch.rs
+++ b/storage/src/qmdb/any/batch.rs
@@ -14,7 +14,7 @@ use crate::{
             ordered::{find_next_key, find_prev_key},
             ValueEncoding,
         },
-        bitmap::{BitmapReadable, Shared},
+        bitmap::Shared,
         delete_known_loc,
         operation::{Key, Operation as OperationTrait},
         update_known_loc,
@@ -23,7 +23,7 @@ use crate::{
 };
 use commonware_codec::Codec;
 use commonware_cryptography::{Digest, Hasher};
-use commonware_utils::bitmap::Prunable;
+use commonware_utils::bitmap;
 use core::{iter, ops::Range};
 use futures::future::try_join_all;
 use std::{
@@ -285,9 +285,11 @@ where
     None
 }
 
-/// Apply a single diff entry to the snapshot index.
-fn apply_snapshot_diff<F: Family, V, I: UnorderedIndex<Value = Location<F>>>(
+/// Apply a single diff entry to the snapshot index and activity bitmap in lockstep:
+/// install the winning `Active` location and clear the prior committed location.
+fn apply_diff<F: Family, V, I: UnorderedIndex<Value = Location<F>>, const N: usize>(
     snapshot: &mut I,
+    bitmap: &mut bitmap::Prunable<N>,
     key: &impl Key,
     entry: &DiffEntry<F, V>,
     base_old_loc: Option<Location<F>>,
@@ -303,24 +305,7 @@ fn apply_snapshot_diff<F: Family, V, I: UnorderedIndex<Value = Location<F>>>(
             }
         }
     }
-}
-
-/// Mirror a single diff transition into the activity bitmap: set the
-/// winning Active location to 1 (if present), clear the prior committed
-/// location to 0 (if present).
-fn apply_bitmap_diff<F: Family, V, const N: usize>(
-    bitmap: &mut Prunable<N>,
-    entry: &DiffEntry<F, V>,
-    base_old_loc: Option<Location<F>>,
-) {
-    let new_loc = entry.loc();
-    if let (Some(new), Some(old)) = (new_loc, base_old_loc) {
-        debug_assert_ne!(
-            new, old,
-            "same location cannot be both active and superseded",
-        );
-    }
-    if let Some(loc) = new_loc {
+    if let Some(loc) = entry.loc() {
         bitmap.set_bit(*loc, true);
     }
     if let Some(loc) = base_old_loc {
@@ -342,7 +327,7 @@ fn next_candidate<F: Family, const N: usize>(
     tip: u64,
 ) -> Option<Location<F>> {
     let floor = *floor;
-    let bitmap_len = BitmapReadable::<N>::len(bitmap);
+    let bitmap_len = bitmap::Readable::<N>::len(bitmap);
     let committed_end = bitmap_len.min(tip);
     if floor < committed_end {
         if let Some(idx) = bitmap.next_one_from(floor) {
@@ -655,6 +640,10 @@ where
                     let Some(key) = op.key().cloned() else {
                         continue; // skip CommitFloor and other non-keyed ops
                     };
+                    // `is_active_at` is required even for set bits in the committed bitmap
+                    // range: this batch's own diff or an uncommitted ancestor diff may
+                    // supersede the committed location, and neither source is reflected in
+                    // the bitmap.
                     if !self.is_active_at(&key, candidate, &diff, db) {
                         continue;
                     }
@@ -1612,8 +1601,7 @@ where
                     .get(key)
                     .copied()
                     .unwrap_or_else(|| entry.base_old_loc());
-                apply_snapshot_diff(&mut self.snapshot, key, entry, base_old_loc);
-                apply_bitmap_diff(&mut bitmap, entry, base_old_loc);
+                apply_diff(&mut self.snapshot, &mut bitmap, key, entry, base_old_loc);
             }
 
             // Apply uncommitted ancestor diffs (skip committed batches, skip seen keys).
@@ -1627,8 +1615,7 @@ where
                             .get(key)
                             .copied()
                             .unwrap_or_else(|| entry.base_old_loc());
-                        apply_snapshot_diff(&mut self.snapshot, key, entry, base_old_loc);
-                        apply_bitmap_diff(&mut bitmap, entry, base_old_loc);
+                        apply_diff(&mut self.snapshot, &mut bitmap, key, entry, base_old_loc);
                     } else if let Some(loc) = entry.loc() {
                         debug_assert!(
                             !bitmap.get_bit(*loc),
@@ -1856,7 +1843,7 @@ mod tests {
     use commonware_cryptography::{sha256, Sha256};
     use commonware_runtime::{deterministic, Runner as _};
 
-    const BITMAP_CHUNK_BITS: u64 = Prunable::<BITMAP_CHUNK_BYTES>::CHUNK_SIZE_BITS;
+    const BITMAP_CHUNK_BITS: u64 = bitmap::Prunable::<BITMAP_CHUNK_BYTES>::CHUNK_SIZE_BITS;
 
     fn loc(n: u64) -> Location<mmr::Family> {
         Location::new(n)
@@ -1864,9 +1851,9 @@ mod tests {
 
     fn shared_with<F>(build: F) -> Shared<BITMAP_CHUNK_BYTES>
     where
-        F: FnOnce(&mut Prunable<BITMAP_CHUNK_BYTES>),
+        F: FnOnce(&mut bitmap::Prunable<BITMAP_CHUNK_BYTES>),
     {
-        let mut bm = Prunable::<BITMAP_CHUNK_BYTES>::new();
+        let mut bm = bitmap::Prunable::<BITMAP_CHUNK_BYTES>::new();
         build(&mut bm);
         Shared::new(bm)
     }
--- a/storage/src/qmdb/any/batch.rs
+++ b/storage/src/qmdb/any/batch.rs
@@ -14,6 +14,7 @@ use crate::{
             ordered::{find_next_key, find_prev_key},
             ValueEncoding,
         },
+        bitmap::{BitmapReadable, Shared},
         delete_known_loc,
         operation::{Key, Operation as OperationTrait},
         update_known_loc,
@@ -22,6 +23,7 @@ use crate::{
 };
 use commonware_codec::Codec;
 use commonware_cryptography::{Digest, Hasher};
+use commonware_utils::bitmap::Prunable;
 use core::{iter, ops::Range};
 use futures::future::try_join_all;
 use std::{
@@ -39,27 +41,6 @@ type DiffSlice<K, F, V> = [(K, DiffEntry<F, V>)];
 /// Sorted `(key, (value, loc))` vec consulted by `find_prev_key` to find the predecessor
 /// of a given key during ordered merkleization.
 type PrevCandidates<K, F, V> = Vec<(K, (V, Location<F>))>;
-
-/// Strategy for finding the next active location during floor raising.
-pub(crate) trait FloorScan<F: Family> {
-    /// Return the next location at or after `floor` that might be active,
-    /// below `tip`. Returns `None` if no candidate exists in `[floor, tip)`.
-    fn next_candidate(&mut self, floor: Location<F>, tip: u64) -> Option<Location<F>>;
-}
-
-/// Sequential scan: every location is a candidate.
-// TODO(#1829): Always use bitmap for floor raising.
-pub(crate) struct SequentialScan;
-
-impl<F: Family> FloorScan<F> for SequentialScan {
-    fn next_candidate(&mut self, floor: Location<F>, tip: u64) -> Option<Location<F>> {
-        if *floor < tip {
-            Some(floor)
-        } else {
-            None
-        }
-    }
-}
 
 /// What happened to a key in this batch.
 #[derive(Clone)]
@@ -324,6 +305,56 @@ fn apply_snapshot_diff<F: Family, V, I: UnorderedIndex<Value = Location<F>>>(
     }
 }
 
+/// Mirror a single diff transition into the activity bitmap: set the
+/// winning Active location to 1 (if present), clear the prior committed
+/// location to 0 (if present).
+fn apply_bitmap_diff<F: Family, V, const N: usize>(
+    bitmap: &mut Prunable<N>,
+    entry: &DiffEntry<F, V>,
+    base_old_loc: Option<Location<F>>,
+) {
+    let new_loc = entry.loc();
+    if let (Some(new), Some(old)) = (new_loc, base_old_loc) {
+        debug_assert_ne!(
+            new, old,
+            "same location cannot be both active and superseded",
+        );
+    }
+    if let Some(loc) = new_loc {
+        bitmap.set_bit(*loc, true);
+    }
+    if let Some(loc) = base_old_loc {
+        bitmap.set_bit(*loc, false);
+    }
+}
+
+/// Return the next floor-raise candidate in `[floor, tip)`.
+///
+/// The committed prefix is indexed by `bitmap`, so unset bits can be skipped without reading
+/// their operations. Locations beyond the bitmap's length are uncommitted ancestor operations
+/// and remain sequential candidates.
+///
+/// `current::batch::next_candidate` mirrors this contract over a layered `BitmapBatch` chain;
+/// both must obey it.
+fn next_candidate<F: Family, const N: usize>(
+    bitmap: &Shared<N>,
+    floor: Location<F>,
+    tip: u64,
+) -> Option<Location<F>> {
+    let floor = *floor;
+    let bitmap_len = BitmapReadable::<N>::len(bitmap);
+    let committed_end = bitmap_len.min(tip);
+    if floor < committed_end {
+        if let Some(idx) = bitmap.next_one_from(floor) {
+            if idx < committed_end {
+                return Some(Location::new(idx));
+            }
+        }
+    }
+    let candidate = floor.max(bitmap_len);
+    (candidate < tip).then(|| Location::new(candidate))
+}
+
 /// Resolve `loc` to an op within the in-memory ancestor region
 /// `[db_size, ancestors[0].journal_batch.size())`, walked parent-first.
 ///
@@ -469,10 +500,10 @@ where
     /// bucket for collision siblings (other keys sharing the same translated-key bucket). The
     /// ordered path needs these so their `next_key` pointers are rewritten when a sibling is
     /// deleted; the unordered path can skip them.
-    fn gather_existing_locations<E, C, I>(
+    fn gather_existing_locations<E, C, I, const N: usize>(
         &self,
         mutations: &BTreeMap<U::Key, Option<U::Value>>,
-        db: &Db<F, E, C, I, H, U>,
+        db: &Db<F, E, C, I, H, U, N>,
         include_active_collision_siblings: bool,
     ) -> Vec<Location<F>>
     where
@@ -518,12 +549,12 @@ where
     }
 
     /// Check if the operation at `loc` for `key` is still active.
-    fn is_active_at<E, C, I>(
+    fn is_active_at<E, C, I, const N: usize>(
         &self,
         key: &U::Key,
         loc: Location<F>,
         batch_diff: &DiffSlice<U::Key, F, U::Value>,
-        db: &Db<F, E, C, I, H, U>,
+        db: &Db<F, E, C, I, H, U, N>,
     ) -> bool
     where
         E: Context,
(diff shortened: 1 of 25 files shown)


roberto-bayardo:
do the benchmarks look OK now? last time I tried this it made things slower but that was before all the optimization work.

danlaine:
> do the benchmarks look OK now? last time I tried this it made things slower but that was before all the optimization work.

I added a new benchmark and ran on `main` and this branch. Looks like a modest perf improvement, or at least not a regression:
```
  ┌───────────────────────────────┬─────────────┬──────┬───────┐   
  │            Variant            │ origin/main │ HEAD │   Δ   │   
  ├───────────────────────────────┼─────────────┼──────┼───────┤   
  │ any::ordered::fixed::mmb      │        3.57 │ 3.25 │ −9.2% │   
  ├───────────────────────────────┼─────────────┼──────┼───────┤
  │ any::ordered::variable::mmr   │        4.53 │ 4.14 │ −8.6% │   
  ├───────────────────────────────┼─────────────┼──────┼───────┤
  │ any::ordered::fixed::mmr      │        3.55 │ 3.30 │ −7.2% │   
  ├───────────────────────────────┼─────────────┼──────┼───────┤   
  │ any::ordered::variable::mmb   │        4.50 │ 4.25 │ −5.5% │
  ├───────────────────────────────┼─────────────┼──────┼───────┤   
  │ any::unordered::variable::mmr │        3.78 │ 3.65 │ −3.3% │ 
  ├───────────────────────────────┼─────────────┼──────┼───────┤   
  │ any::unordered::variable::mmb │        3.70 │ 3.62 │ −2.2% │ 
  ├───────────────────────────────┼─────────────┼──────┼───────┤   
  │ any::unordered::fixed::mmr    │        2.84 │ 2.79 │ −1.7% │
  ├───────────────────────────────┼─────────────┼──────┼───────┤   
  │ any::unordered::fixed::mmb    │        2.75 │ 2.85 │ +3.8% │ 
  └───────────────────────────────┴─────────────┴──────┴───────┘  
  ```

File: storage/src/qmdb/benches/merkleize.rs [hunk 1]
```diff
@@ -1,10 +1,11 @@
...
-//! Measures the time required to create a speculative batch (applying random updates equal to 10%
-//! of the total key count, sampled with replacement), merkleize it, and compute its root. The
-//! database is initialized with N unique keys having random digests as values. Database
-//! initialization time is not included in the benchmark. The page cache is large enough to hold the
-//! entire active key set to eliminate disk access delays from affecting the results.
+//! Each iteration creates a speculative batch (10% random updates, sampled with replacement),
+//! merkleizes it, and reads the root. The DB is seeded with N unique keys; setup is not timed.
+//!
+//! - [`bench_merkleize`]: steady-state timing on a freshly seeded DB.
+//! - [`bench_merkleize_churned`]: timing after overwrite batches create stale log history for the
```
roberto-bayardo: What is a "stale log history"?  If this is just "log history that is no longer in the cache"?

storage/src/qmdb/benches/merkleize.rs
```diff
@@ -252,7 +253,10 @@ type CurOVar256Mmb = commonware_storage::qmdb::current::ordered::variable::Db<
 const ITEMS_PER_BLOB: NonZeroU64 = NZU64!(10_000_000);
 const THREADS: NonZeroUsize = NZUsize!(8);
 const PAGE_SIZE: NonZeroU16 = NZU16!(4096);
+// Large size to reduce read I/O.
```
roberto-bayardo: The intent of these values is to not just to reduce read I/O but to ensure the db is *fully* cached.

storage/src/qmdb/benches/merkleize.rs
```diff
@@ -370,6 +374,40 @@ async fn run_bench<F: merkle::Family, C: DbAny<F, Key = Digest, Value = Digest>>
     total
 }
 
+/// Apply overwrite batches before timing merkleization.
+///
+/// This leaves stale Updates above the inactivity floor, matching the workload
```
roberto-bayardo: By "stale  Updates" do you mean "inactive update operations"?

storage/src/qmdb/any/unordered/variable.rs
```diff
@@ -1334,4 +1334,421 @@ pub(crate) mod test {
             db.destroy().await.unwrap();
         });
     }
+
+    // ---- Bitmap integration tests ----
+    //
+    // Stale bitmap bits can leave roots and snapshots correct while making
```
roberto-bayardo: what's a stale bit?

storage/src/qmdb/any/unordered/variable.rs
```diff
@@ -1334,4 +1334,421 @@ pub(crate) mod test {
...
+        assert_eq!(
+            pre, post,
+            "bitmap active locs diverged from oracle (pre != post-reopen)",
+        );
+        db
+    }
+
+    /// Partial ancestor commit: committed K@L, parent P updates K to L2
+    /// (uncommitted), child C deletes K, apply C directly. The bit at L2
+    /// must be cleared.
```
roberto-bayardo: I find this kind of test description a bit too arcane.

storage/src/qmdb/any/batch.rs
```diff
@@ -1032,17 +1066,24 @@ where
...
 
-    /// Like [`merkleize`](Self::merkleize) but accepts a custom [`FloorScan`]
-    /// to accelerate floor raising.
-    pub(crate) async fn merkleize_with_floor_scan<E, C, I, S: FloorScan<F>>(
+    /// Like [`merkleize`](Self::merkleize), but accepts the floor-raise
+    /// candidate source.
+    ///
+    /// The callback must return the next location in `[floor, tip)` that may
+    /// still be active, or `None` when no such location remains. It may skip
+    /// locations only when it knows they are inactive. The floor-raise loop
```
roberto-bayardo: Is it still necessary to perform the verification? I would think with the bitmap these are no longer "candidates" but known-active-bits.

storage/src/qmdb/any/db.rs
```diff
@@ -83,6 +87,18 @@ pub struct Db<
...
+    /// Activity bitmap over committed operations. This is rebuilt from the
+    /// journal on init and is never persisted separately.
+    ///
+    /// # Invariants
+    ///
+    /// - `bitmap.len() == log.size()`.
+    /// - `bitmap.pruned_bits()` is no greater than the retained log start.
+    /// - `bitmap[i] == 1` iff location `i` is the currently active update
+    ///   for its key. False positives are filtered by merkleization, but a
+    ///   false negative can advance the inactivity floor past live state.
```
roberto-bayardo: the "iff" implies there should be no false positives or false negatives?

storage/src/qmdb/any/db.rs
```diff
@@ -260,6 +276,11 @@ where
 
         self.log.prune(prune_loc).await?;
 
+        // Keep the bitmap's pruned boundary in lockstep with the log.
+        // `prune_to_bit` rounds down to a chunk boundary, which is fine;
+        // bitmap memory is freed chunk-by-chunk regardless.
```
roberto-bayardo: Something more concise like "// Align bitmap's pruning boundary with the log." is probably entirely adequate.

storage/src/qmdb/any/db.rs
```diff
@@ -83,6 +87,18 @@ pub struct Db<
...
+    /// journal on init and is never persisted separately.
+    ///
+    /// # Invariants
+    ///
+    /// - `bitmap.len() == log.size()`.
+    /// - `bitmap.pruned_bits()` is no greater than the retained log start.
+    /// - `bitmap[i] == 1` iff location `i` is the currently active update
+    ///   for its key. False positives are filtered by merkleization, but a
+    ///   false negative can advance the inactivity floor past live state.
+    pub(crate) bitmap: Prunable<BITMAP_CHUNK_BYTES>,
```
roberto-bayardo: Does this mean we're maintaining 2 bitmaps with current db?  (Current embeds an anydb and retains a bitmap.)

My PoC for this issue made this an Option so it could be None when used with a current db.

roberto-bayardo: In case it's useful here's my (very vibey) change for this feature:  https://github.com/commonwarexyz/monorepo/pull/3531

storage/src/qmdb/benches/merkleize.rs
danlaine: Updated the comments to be more clear

storage/src/qmdb/benches/merkleize.rs
danlaine: Hmm yeah the agent used "stale" to mean "inactive" -- updated to "inactive" to follow the convention

storage/src/qmdb/any/unordered/variable.rs
danlaine: Same as above -- changed stale -> inactive

storage/src/qmdb/any/unordered/variable.rs
danlaine: Updated these comments to be more readable

storage/src/qmdb/any/batch.rs
danlaine: Updated comments to be more clear about this -- the bitmap is updated on `apply_batch` so its bits describe committed data, not the speculative view used in merkleize. 0 means "this location is definitely inactive", 1 means "this location might be active" (or it could have been overwritten to inactive in the speculative view)

storage/src/qmdb/any/db.rs
danlaine: Hmm yeah it does seem redundant to maintain 2 bitmaps now. I'll try to combine them.

storage/src/qmdb/any/db.rs
danlaine: See https://github.com/commonwarexyz/monorepo/pull/3662#discussion_r3142307797

storage/src/qmdb/any/batch.rs
```diff
@@ -324,6 +305,56 @@ fn apply_snapshot_diff<F: Family, V, I: UnorderedIndex<Value = Location<F>>>(
...
+            "same location cannot be both active and superseded",
+        );
+    }
+    if let Some(loc) = new_loc {
+        bitmap.set_bit(*loc, true);
+    }
+    if let Some(loc) = base_old_loc {
+        bitmap.set_bit(*loc, false);
+    }
+}
```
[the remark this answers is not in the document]
danlaine: This is a false positive. The old ChunkOverlay::clear_bit guard was an overlay-layer concern: it avoided materializing overlay chunks for bitmap chunks that the base bitmap had already pruned. It was not a precondition check implying valid diff data may contain active old locations below the bitmap prune boundary.

The invariant the new code relies on is: every loc passed to bitmap.set_bit in any::Db::apply_batch is >= bitmap.pruned_bits()

That still holds. current::Db::prune may prune the bitmap farther than the ops log, so bitmap.pruned_bits() can exceed log.bounds().start. But it only prunes the bitmap to sync_boundary(), and sync_boundary() <= inactivity_floor_loc. The log may retain historical inactive operations below that point, but the snapshot does not reference them; active snapshot locations are bounded by inactivity_floor_loc, not by log.bounds().start.

For the apply_bitmap_diff clear path, base_old_loc resolves to the active committed location being superseded, after accounting for already-committed ancestors. Such a location is at or above the inactivity floor. If floor raising would otherwise pass over an active location, merkleization moves that key above the fixed tip before advancing the floor. New locations and commit-floor updates are also in the retained/appended range, so they are above the bitmap prune boundary as well.

storage/src/qmdb/bitmap.rs
```diff
@@ -0,0 +1,89 @@
...
+//! [`MerkleizedBatch`](super::current::batch::MerkleizedBatch)es via `Arc<Shared<N>>`.
+//!
+//! `any::Db` mutates the inner [`Prunable`] under a [`RwLock`] during `apply_batch` / `prune` /
+//! `rewind` while live batches read concurrently. Locking (not snapshotting) keeps memory at
+//! O(bitmap size); snapshots would couple memory to live-batch count and lifetime.
+//!
+//! Reads through an invalidated `MerkleizedBatch` (see its "Branch validity" docs) return
+//! inconsistent bytes; callers must drop invalid batches.
+
+pub(crate) use commonware_utils::bitmap::Readable as BitmapReadable;
```
roberto-bayardo: wdyt about just using "bitmap::Readable" for this instead of "BitmapReadable"?

storage/src/qmdb/current/batch.rs
```diff
@@ -113,48 +111,30 @@ impl<const N: usize> ChunkOverlay<N> {
...
-            let candidate = loc.max(bitmap_len);
-            if candidate < tip {
-                return Some(Location::<F>::new(candidate));
+/// Bitmap-accelerated floor scan over a layered `BitmapBatch` chain. Skips locations where the
+/// bitmap bit is unset, avoiding I/O reads for inactive operations.
+///
+/// Mirrors the contract on `any::batch::next_candidate`: may return only locations that are
+/// *possibly* active in `[floor, tip)`, may skip locations only when known inactive.
+/// `is_active_at` revalidates each candidate, so false positives are tolerated; false negatives
+/// are forbidden.
```
roberto-bayardo, line 120: might be worth explaining why false positives are even possible (i found it a little surprising)

roberto-bayardo: oh I see you do this in merkleize_with_floor_scan, but it probably better belongs here?

storage/src/qmdb/any/batch.rs
```diff
@@ -866,27 +896,33 @@ where
...
+        .await
     }
 
-    /// Like [`merkleize`](Self::merkleize) but accepts a custom [`FloorScan`]
-    /// to accelerate floor raising.
-    pub(crate) async fn merkleize_with_floor_scan<E, C, I, S: FloorScan<F>>(
+    /// Like [`merkleize`](Self::merkleize), but accepts the floor-raise candidate source.
+    ///
+    /// The callback may skip locations only when it knows they are inactive. The floor-raise
+    /// loop revalidates each returned candidate via `is_active_at` because the bitmap reflects
```
roberto-bayardo, line 907: is it worth skipping the is_active_at call for bits in the committed range? (Or can those still be false positives?)

storage/src/qmdb/any/batch.rs
```diff
@@ -1022,27 +1058,33 @@ where
...
+    /// committed state only — uncommitted ancestor ops aren't tracked, and bits can be set for
+    /// locations superseded by an overlay in this chain.
+    pub(crate) async fn merkleize_with_floor_scan<E, C, I, const N: usize>(
```
roberto-bayardo, line 1072: does merkleize_with_floor_scan need to be its own method?  I think merkelize is its only caller so you could inline?

storage/src/qmdb/any/batch.rs
```diff
@@ -1556,47 +1598,64 @@ where
...
+            // don't rely on iteration order.
+            let mut seen = HashSet::<&U::Key>::new();
+            for (key, entry) in batch.diff.iter() {
+                seen.insert(key);
                 let base_old_loc = committed_locs
                     .get(key)
                     .copied()
                     .unwrap_or_else(|| entry.base_old_loc());
                 apply_snapshot_diff(&mut self.snapshot, key, entry, base_old_loc);
+                apply_bitmap_diff(&mut bitmap, entry, base_old_loc);
```
roberto-bayardo: Since these two apply_ functions are now called together everywhere, does it make sense to just combine them?

roberto-bayardo: Alternatively Claude recommended a collect-and-apply pattern to avoid holding the bitmap write lock during the much-more-costly snapshot maintenance, but this may be overkill 🤷 

```
Better: single iteration, collect bit-ops, apply under one short lock:                                                                              
                                                                                                                                                      
  let mut bit_ops: Vec<(u64, bool)> = Vec::with_capacity(batch.diff.len() * 2);                                                                       
  let mut seen = HashSet::<&U::Key>::new();                                                                                                           
  for (key, entry) in batch.diff.iter() {                                                                                                             
      seen.insert(key);                                                                                                                               
      let base_old_loc = committed_locs.get(key).copied().unwrap_or_else(|| entry.base_old_loc());
      apply_snapshot_diff(&mut self.snapshot, key, entry, base_old_loc);                                                                              
      if let Some(loc) = entry.loc()      { bit_ops.push((*loc, true)); }                                                                             
      if let Some(loc) = base_old_loc     { bit_ops.push((*loc, false)); }                                                                            
  }                                                                                                                                                   
  // ancestors: same shape, push to bit_ops while still under no lock                                                                                 
                                                                                                                                                      
  {
      let mut bitmap = self.bitmap.write();                                                                                                           
      bitmap.extend_to(*batch.new_last_commit_loc + 1);                                                                                               
      for (loc, val) in bit_ops { bitmap.set_bit(loc, val); }
      bitmap.set_bit(*self.last_commit_loc, false);                                                                                                   
      bitmap.set_bit(*batch.new_last_commit_loc, true);  
  }                                                                                                                                                   
                                                                                                                                                      
  Trade: one extra Vec (≈16 bytes × ~2 entries per diff). Gain: single iteration, no redundant work, lock holds only for extend_to + N bit-flips + 2  
  commit-floor flips — all O(1) per op, no hashmaps, no committed_locs lookups under the lock. Atomicity preserved (readers still see all-or-nothing).
                                                                                                                                                      
  The debug_assert!(!bitmap.get_bit(*loc), …) for shadowed ancestor locations does need the lock, but it's debug_assert — easy to either drop, or push
   the loc into a separate Vec and assert under the lock at the end.
```

storage/src/qmdb/any/mod.rs
```diff
@@ -2379,3 +2487,232 @@ pub(crate) mod test {
         });
     }
 }
+
+#[cfg(test)]
+mod bitmap_tests {
+    //! Regression tests for activity-bitmap maintenance in `any::Db`. The mutation code in
+    //! `apply_batch`, `prune_bitmap`, and `rewind` is parametric on bitmap geometry alone, so one
```
roberto-bayardo: "parametric on bitmap geometery alone" is pretty odd (AI?) wording!

storage/src/qmdb/bitmap.rs
danlaine: Renamed to `Readable` and changed imports to `bitmap` so we do `bitmap::Readable`

storage/src/qmdb/current/batch.rs
danlaine: Expanded comments

storage/src/qmdb/any/batch.rs
danlaine: Added comment explaining why there still may be false positives

storage/src/qmdb/any/mod.rs
danlaine: Fixed

storage/src/qmdb/any/batch.rs
danlaine: There are multiple callers; in qmdb::current we pass in the layered bitmap instead of the committed one

storage/src/qmdb/any/batch.rs
danlaine: Combined into `apply_diff`

storage/src/qmdb/any/db.rs
danlaine: Removed comment.

roberto-bayardo (approved):

PR #3663, opened 2026-04-24
[storage] report qmdb sync progress

clabby:
## Overview

Adds metrics to the QMDB sync engine for the current size of the journal as well as the operations required per the current sync target. Metrics are updated every time a batch of operations is ingested or the sync target is updated.

### DIFF
--- a/storage/src/qmdb/sync/engine.rs
+++ b/storage/src/qmdb/sync/engine.rs
@@ -16,7 +16,10 @@ use crate::{
 use commonware_codec::Encode;
 use commonware_cryptography::Digest;
 use commonware_macros::select;
-use commonware_runtime::{telemetry::metrics::status::GaugeExt, Metrics as _};
+use commonware_runtime::{
+    telemetry::metrics::{Gauge, GaugeExt, MetricsExt},
+    Metrics as _,
+};
 use commonware_utils::{
     channel::{
         fallible::{AsyncFallibleExt, OneshotExt as _},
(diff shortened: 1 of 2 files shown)


storage/src/qmdb/sync/engine.rs
```diff
@@ -63,6 +66,36 @@ enum Event<F: Family, Op, D: Digest, E> {
...
+impl ProgressMetrics {
+    /// Register sync progress metrics on the provided context.
+    fn new(context: &impl commonware_runtime::Metrics) -> Self {
+        let journal_size = context.gauge(
+            "journal_size",
+            "Current journal size (operations applied) per database",
+        );
+        let target_end = context.gauge(
+            "target_end",
+            "Target range end (operations needed) per database",
```
danlaine: I don't think these are "per database" in the context of one instance of `ProgressMetrics`, right?

storage/src/qmdb/sync/engine.rs
```diff
@@ -307,6 +344,7 @@ where
             reached_target_tx: config.reached_target_tx,
             finish_requested: false,
             reached_current_target_reported: false,
+            progress_metrics,
```
danlaine, line 343: Should we update `progress_metrics` with the initial target before retuning?

storage/src/qmdb/sync/engine.rs
```diff
@@ -483,6 +521,12 @@ where
         self.reached_current_target_reported = true;
     }
 
+    /// Record a progress snapshot in metrics.
+    async fn record_progress(&mut self) {
```
danlaine: I think this can be `&self`?

storage/src/qmdb/sync/engine.rs
clabby: Ah yep, didn't update this from how `glue` used to do this with `Family`. Updated.

danlaine (approved):

PR #3670, opened 2026-04-27
[storage/qmdb] document ordered stream_range Send boundary

christopherwxyz:
## Summary

Adds a pure in-tree example for the ordered QMDB `Send` boundary:

- compile-time assertions that ordered variable QMDB point reads and batch merkleization compose with `Send`
- a `compile_fail` doctest showing that `stream_range(...)` still does not compose cleanly with a `Send`-required future

This is intended to complement `#3669`, which asserts the `Db` / `get` side of the story. This PR captures the narrower remaining boundary around ordered streaming.

## Why

Downstream consumers were still treating QMDB as broadly `!Sync` / `!Send`-hostile, but the current published crate already supports `Send` composition for:

- `&Db`
- `db.get(...)`
- caller futures that hold `&Db` across multiple awaited point reads
- `new_batch().merkleize(&db, ...)`

What still fails is the ordered streaming shape:

```rust
async move {
    let stream = db.stream_range(start).await.unwrap();
    futures::pin_mut!(stream);
    let _ = stream.next().await;
}
```

when that caller future must itself be `Send`.

## Validation

- `cargo test -p commonware-storage --doc --manifest-path /tmp/commonware-monorepo/Cargo.toml stream_range`
- the added compile-time helper assertions build as part of storage test compilation

### DIFF
--- a/storage/src/qmdb/current/ordered/db.rs
+++ b/storage/src/qmdb/current/ordered/db.rs
@@ -81,6 +81,29 @@ where
 
     /// Streams all active (key, value) pairs in the database in key order, starting from the first
     /// active key greater than or equal to `start`.
+    ///
+    /// This currently does not compose cleanly with `Send`-required futures because the returned
+    /// stream borrows `&self`.
+    ///
+    /// ```compile_fail
+    /// use commonware_cryptography::{sha256::Digest, Sha256};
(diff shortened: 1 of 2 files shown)


christopherwxyz:
Not needed. See #3679

Issue #3668, opened 2026-04-26
[cryptography/ed25519] Further Unify `ed25519-consensus`


patrick-ogrady:
Following #3616, we should be able to remove some intermediate types from `ed25519::core` (like `VerificationKey`).

PR #3671, opened 2026-04-27
[cryptography/ed25519] no_std Batch

patrick-ogrady:
Related: #3616


cryptography/src/ed25519/mod.rs
```diff
@@ -26,9 +27,8 @@
 //! ```
 
 pub mod certificate;
-pub mod core;
+pub(in crate::ed25519) mod core;
```
patrick-ogrady, line 30: I think we should keep this internal (especially given #3668)

PR #3672, opened 2026-04-27
[ci] Increase Optimization Level for Cryptography Crates in Tests

patrick-ogrady:
Related: #3616


patrick-ogrady:
The compilation in this PR may mean the e2e time goes up but once merged to main this should be a clear win (load compiled crates from cache).

Issue #725, opened 2025-04-14
[codec] `Vec<T>` should use a more specific implementation if `T` is exactly `u8`


BrendanChou:
In rust, there is an issue that you can't have a duplicate implementation. But ideally we would be able to have a much more efficient implementation for `Vec<u8>` by reading the contiguous bytes all-at-once rather than byte-by-byte.

I haven't yet figured out a good way to do this in rust however.

Similarly, for any type `T` which is `FixedSize`, we should be able to assert `at_least(buf, T::SIZE * len)` fore allocating the vector, which can help prevent certain memory-exhaustion attacks even if a very-permissive range-configuration is supplied

malik672:
I don't really think this is an issue tbh for the `Vec<u8>`, the cacheline automatically fetches 64 bytes at once
There can be an assertion for the length before the computation starts, this also helps for bound checks later in the code

BrendanChou:
As of writing, "specialization" already exists in nightly rust but is not in the stable version yet
https://rust-lang.github.io/rfcs/1210-impl-specialization.html

christopherwxyz:
A stable-Rust path that works without the orphan/coherence problem and without `min_specialization`: do the `u8` specialization inside the generic `impl<T: Read> Read for Vec<T>` via a `TypeId` check. `Vec<u8>` and `Vec<T>` have identical layout when `T == u8`, so a `Vec<u8>` constructed via bulk-copy can be safely transmuted back into `Vec<T>`.

```rust
use core::any::TypeId;

impl<T: Read + 'static> Read for Vec<T> {
    type Cfg = (RangeCfg<usize>, T::Cfg);

    #[inline]
    fn read_cfg(buf: &mut impl Buf, (range, cfg): &Self::Cfg) -> Result<Self, Error> {
        let len = usize::read_cfg(buf, range)?;

        // Specialize Vec<u8>: bulk-copy via Buf::copy_to_slice rather than
        // calling u8::read_cfg in a per-byte loop.
        if TypeId::of::<T>() == TypeId::of::<u8>() {
            if buf.remaining() < len {
                return Err(Error::EndOfBuffer);
            }
            let mut bytes: Vec<u8> = vec![0u8; len];
            buf.copy_to_slice(&mut bytes);
            // SAFETY: T == u8 (verified above); Vec<u8> and Vec<T> have the
            // same memory layout, so the transmute is sound.
            let v: Vec<T> = unsafe { core::mem::transmute(bytes) };
[...]

PR #3673, opened 2026-04-27
[codec] specialize codec for byte containers

andresilva:
This PR adds aggregate container hooks to `EncodeSize`, `Write`, and `Read` so generic containers can delegate sizing, writing, and reading to the element implementation. This lets `Vec<T>`, `&[T]`, and `[T; N]` keep a single generic implementation while `u8` provides bulk-copy paths for byte containers on stable Rust, where overlapping specialized container impls are not available.

| Operation | Len | Before ns | After ns | Speedup |
|---|---:|---:|---:|---:|
| encode_vec_u8 | 32 | 88.35 | 23.94 | 3.69x |
| encode_vec_u8 | 1024 | 2440.80 | 29.04 | 84.05x |
| encode_vec_u8 | 65536 | 147560.00 | 749.56 | 196.86x |
| decode_vec_u8 | 32 | 38.07 | 22.26 | 1.71x |
| decode_vec_u8 | 1024 | 601.25 | 40.79 | 14.74x |
| decode_vec_u8 | 65536 | 37834.00 | 862.00 | 43.89x |
| encode_slice_u8 | 32 | 90.72 | 23.69 | 3.83x |
| encode_slice_u8 | 1024 | 2541.70 | 28.96 | 87.77x |
| encode_slice_u8 | 65536 | 161200.00 | 762.36 | 211.45x |

This also generalizes array support: `[T; N]` now implements `Write` and `Read` whenever `T` does. Fixed-size arrays continue to get `FixedSize`, and therefore `Encode`/`Codec`, while arrays of variable-size elements can be written/read/decoded but not encoded because a generic `EncodeSize` for `[T; N]` would overlap with the blanket `EncodeSize` for `T where T: FixedSize`.

Related #725.

### DIFF
--- a/codec/src/codec.rs
+++ b/codec/src/codec.rs
@@ -1,7 +1,11 @@
 //! Core traits for encoding and decoding.
 
 use crate::error::Error;
+#[cfg(not(feature = "std"))]
+use alloc::vec::Vec;
 use bytes::{Buf, BufMut, Bytes, BytesMut};
+#[cfg(feature = "std")]
+use std::vec::Vec;
 
 /// Trait for types with a known, fixed encoded size.
 ///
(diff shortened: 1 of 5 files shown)


codec/src/codec.rs
```diff
@@ -23,20 +27,86 @@ pub trait EncodeSize {
...
+    ///
+    /// This hook exists because stable Rust cannot express the overlapping specialized
+    /// container impls that would otherwise provide this aggregate path directly.
+    ///
+    /// This is hidden from generated documentation because it is an implementation hook for
+    /// codec's container types, not part of the intended user-facing API. Most users should
+    /// implement [EncodeSize::encode_size] or [FixedSize] instead.
+    #[doc(hidden)]
+    #[inline]
+    fn encode_size_slice(values: &[Self]) -> usize
```
andresilva, line 46: In my tests I noticed that LLVM was smart enough to see through this for `FixedSize` and basically optimize this to `FixedSize::SIZE * len`. I tested wrapping a `FixedSize` implementation in `black_box(SIZE)` which as expected defeated the optimization. So it's arguable that we might not need this, although this has the advantage of making the optimization explicit rather than relying on the compiler doing the right thing.

codec/src/lib.rs
```diff
@@ -32,14 +32,37 @@
 //!   that the entire buffer is consumed.
 //! - [Codec]: Combines [Encode] + [Decode].
 //!
+//! # Implementation Notes
```
patrick-ogrady: nit: more specific title? How about something along the lines of "Specialization"?

codec/src/lib.rs
```diff
@@ -32,14 +32,37 @@
...
 //! # Supported Types
 //!
 //! Natively supports encoding/decoding for:
 //! - Primitives: [bool],
 //!   [u8], [u16], [u32], [u64], [u128],
 //!   [i8], [i16], [i32], [i64], [i128],
-//!   [f32], [f64], [u8; N],
-//!   and [usize] (must fit within a [u32] for cross-platform compatibility).
+//!   [f32], [f64], and [usize] (must fit within a [u32] for cross-platform compatibility).
+//! - Arrays: `[T; N]` implements [Write] and [Read] when `T` does. That means arrays with
```
patrick-ogrady: This description seems verbose for this list?

codec/src/codec.rs
patrick-ogrady: Let's keep as-is.

patrick-ogrady (dismissed):

patrick-ogrady (approved):

PR #3674, opened 2026-04-27
[Storage] Use `commonware-parallel` in `commonware-storage`

### LINKED ISSUE #3001: [storage] Migrate to `commonware-parallel`
Related: #2954 
Related: #3000

Migrate to `commonware-parallel` instead of using `ThreadPool` (should allow us to remove all `_parallel` functions).

danlaine:
Resolves [#3001](https://github.com/commonwarexyz/monorepo/issues/3001).

## Summary

Replaces the `Option<ThreadPool>` plumbing (and direct `rayon` dependency) in `commonware-storage` with a `Strategy` generic from `commonware-parallel`, matching the pattern already used by `consensus`, `cryptography`, and `coding`.

## Motivation

Storage was the last in-tree consumer of parallelism still calling `pool.install(|| par_iter()...collect())` directly on `rayon`. Every other primitive had migrated to the `Strategy` trait, which lets callers pass `Sequential` for serial execution or `Rayon::new(n)` for parallel without the callee needing to fork on a `MIN_TO_PARALLELIZE` size threshold.

This PR finishes the migration so storage uses the same idiom and so we can remove `rayon` as a direct dependency.

### Type-level

Every type that previously held an `Option<ThreadPool>` now carries an `S: Strategy = Sequential` generic. The default keeps existing callers unchanged at most call sites:

| Type | Before | After |
|---|---|---|
| `merkle::batch::{Unmerkleized,Merkleized}Batch` | `pool: Option<ThreadPool>` | `S: Strategy = Sequential` |
| `merkle::persisted::full::{Config, SyncConfig, Merkle, UnmerkleizedBatch}` | pool field | `S: Strategy` propagated |
| `merkle::persisted::compact::{Config, Merkle, UnmerkleizedBatch}` | pool field | `S: Strategy` propagated |
| `merkle::{mmr,mmb}::full::{Mmr, Mmb, SyncConfig}` | type aliases | now `<E, D, S = Sequential>` |
| `bitmap::authenticated::BitMap<E, D, N, M, S>` | state-marker generic was `S` | renamed to `M`; new `S: Strategy` added |
| `journal::authenticated::{Journal, Unmerkleized, Merkleized}Batch` | pool field | `S: Strategy` propagated |
| `qmdb::any::{Db, Config, batch::*}` | pool field | `S: Strategy` propagated |
| `qmdb::current::{Db, Config, batch::*, sync::build_db}` | pool field/param | `S: Strategy` propagated |
| `qmdb::{immutable,keyless}::{Db, Config, batch::*}` | pool field | `S: Strategy` propagated |
| `qmdb::{immutable,keyless}::compact::{Db, Config, UnmerkleizedBatch, MerkleizedBatch}` | inner Merkle pool | `S: Strategy` propagated |
| `qmdb::compact_witness::WitnessSource<F, E, H, S>` | trait | `S: Strategy` added so compact dbs with non-default strategies can be witness sources |

### Behavioral

- **Single merkleization path.** `merkleize_serial` / `merkleize_parallel` collapse into one body that calls `strategy.map_init_collect_vec(...)`. Same for `compute_height_*`. The `MIN_TO_PARALLELIZE = 20` threshold is deleted; `Sequential` is a thin `.iter().map().collect()` wrapper, so small inputs stay cheap without internal forking.
- **No more `with_pool` builder.** The strategy is part of the batch's identity. Public entry points for explicit strategies are `Mem::new_batch_with_strategy(strategy)` and `MerkleizedBatch::from_mem_with_strategy(mem, strategy)`; production callers usually go through `merkle::full::Merkle::new_batch` / `merkle::compact::Merkle::new_batch`, which inherit the Merkle's configured strategy.
- **Strategy propagates through batch chains.** An `UnmerkleizedBatch` shares its parent through `Arc<MerkleizedBatch>` and reads `self.parent.strategy` directly, so an entire chain shares one strategy without each batch storing a redundant copy.

### Trait changes

- `journal::authenticated::Inner::init` gains an `S: Strategy` method-level generic (the trait itself stays simple). Implementors thread `S` through to the merkle layer; the journal doesn't use the strategy directly.
- `qmdb::sync::Database` and `qmdb::sync::compact::Database` impls for every concrete `Db` (`any::Db`, `current::Db`, `Keyless`, `Immutable`, both compact dbs) are re-parameterized over `S: Strategy` so a `Db<..., Rayon>` is reachable through the sync trait, not just `Db<..., Sequential>`. The `database::Config` impls in `qmdb/sync/database.rs` follow.

### Cargo.toml

- `rayon` removed as a direct dependency.
- `commonware-parallel` becomes non-optional with `default-features = false` (so `Sequential` is available in `no_std` builds). The `std` feature now enables `commonware-parallel/std`, which turns on `Rayon`.


storage/fuzz/fuzz_targets/current_crash_recovery.rs
```diff
@@ -103,7 +104,7 @@ fn make_config(
             metadata_partition: format!("crash-merkle-metadata-{suffix}"),
             items_per_blob: NZU64!(merkle_items_per_blob),
             write_buffer,
-            thread_pool: None,
+            strategy: Sequential,
```
roberto-bayardo, line 107: I wonder if we should be doing fuzz testing in parallel mode for more fuzzing per second...

storage/src/merkle/benches/update.rs
```diff
@@ -85,8 +80,8 @@ fn bench_update_family<F: Family>(c: &mut Criterion, runner: &tokio::Runner, fam
                                     leaf_map.insert(rand_leaf_loc, *new_element);
                                 }
 
-                                match strategy {
-                                    Strategy::NoBatching => {
+                                match mode {
+                                    Mode::NoBatching => {
```
roberto-bayardo: I wonder if it's time to remove NoBatching mode... technically we're still using batches just in a dumb way :)

roberto-bayardo:
I assume strategy is a thin wrapper around Rayon, but it would still be good to confirm there are no performance regressions then merkle + qmdb benches.

storage/src/qmdb/current/db.rs
```diff
@@ -813,12 +828,17 @@ pub(super) async fn compute_grafted_root<
 /// `chunk_ops_digest`, then combines with the bitmap chunk: `hash(chunk || chunk_ops_digest)`. For
 /// all-zero chunks the grafted leaf equals the `chunk_ops_digest` directly (zero-chunk identity).
 ///
-/// When a thread pool is provided and there are enough chunks, hashing is parallelized.
-pub(super) async fn compute_grafted_leaves<F: merkle::Graftable, H: Hasher, const N: usize>(
+/// The provided strategy determines whether hashing runs sequentially or in parallel.
```
roberto-bayardo: I think saying something like "..determines if or how to parallelize merkleization." is a bit more accurate.

danlaine: Done

storage/src/merkle/benches/update.rs
danlaine: Removed `NoBatching`

roberto-bayardo (approved):

danlaine:
> I assume strategy is a thin wrapper around Rayon, but it would still be good to confirm there are no performance regressions then merkle + qmdb benches.

Ran the merkle benchmarks and didn't see any regression

PR #2689, opened 2026-01-05
[utils] Roaring Bitmap

patrick-ogrady:

PR #3687, opened 2026-04-29
[utils/bitmap] Roaring Bitmap

michaelkaplan13:
Replace/Extends #2689

Adds roaring bitmap implementation to utils.

On top of @patrick-ogrady's existing branch, this:
- Adds pruning support for roaring bitmaps
- Optimizes the use and conversions to "run" containers, when there are sufficiently few runs to represent it as a sorted vector of runs rather than a full bitmap. In a case where there are 50 gaps in a 65,536 "shelf", this reduced the RAM footprint and wire representation of the roaring bitmap by ~25x.

Benchmarking Results vs `roaring-rs::RoaringTreemap`
Operation | Size | Ours | roaring-rs | Ratio
-- | -- | -- | -- | --
insert | 100 | 496 ns | 839 ns | 1.69× faster
insert | 1,000 | 4.47 µs | 5.90 µs | 1.32× faster
insert | 10,000 | 43.17 µs | 55.65 µs | 1.29× faster
insert_range | 1,000 | 41 ns | 77 ns | 1.90× faster
insert_range | 10,000 | 41 ns | 170 ns | 4.18× faster
insert_range | 100,000 | 65 ns | 612 ns | 9.48× faster
contains | 1,000 | 3.94 µs | 4.52 µs | 1.15× faster
contains | 10,000 | 10.46 µs | 13.63 µs | 1.30× faster
contains | 100,000 | 151.8 µs | 199.6 µs | 1.31× faster
iteration | 1,000 | 1.54 µs | 2.09 µs | 1.35× faster
iteration | 10,000 | 12.73 µs | 27.83 µs | 2.19× faster
iteration | 100,000 | 113.4 µs | 286.9 µs | 2.53× faster
union | 1,000 | 918 ns | 1.03 µs | 1.13× faster
union | 10,000 | 804 ns | 449 ns | 0.56× (slower)
intersection | 1,000 | 712 ns | 1.03 µs | 1.45× faster
intersection | 10,000 | 806 ns | 1.80 µs | 2.23× faster
difference | 1,000 | 765 ns | 920 ns | 1.20× faster
difference | 10,000 | 796 ns | 450 ns | 0.57× (slower)

Workload | Cardinality | Containers | A/B/R | Ours bytes | Ours wire | Theirs wire
-- | -- | -- | -- | -- | -- | --
near-saturated 65k, no gaps | 65,536 | 1 | A=0 B=0 R=1 | 88 | 15 | 8,220
near-saturated 65k, 50 gaps | 65,486 | 1 | A=0 B=0 R=1 | 328 | 215 | 8,220
near-saturated 65k, 500 gaps | 65,036 | 1 | A=0 B=0 R=1 | 2,120 | 2,016 | 8,220
near-saturated 65k, 5000 gaps | 60,495 | 1 | A=0 B=1 R=0 | 8,272 | 8,202 | 8,220
sparse single shelf 100 | 100 | 1 | A=1 B=0 R=0 | 328 | 211 | 228
sparse single shelf 1,000 | 1,000 | 1 | A=1 B=0 R=0 | 2,120 | 2,012 | 2,028
sparse single shelf 4,000 | 4,000 | 1 | A=1 B=0 R=0 | 8,264 | 8,012 | 8,028
multi-shelf 10 × 100 | 1,000 | 10 | A=10 B=0 R=0 | 3,064 | 2,101 | 2,100
multi-shelf 100 × 100 | 10,000 | 100 | A=100 B=0 R=0 | 30,424 | 21,001 | 20,820
multi-shelf 1,000 × 100 | 100,000 | 1,000 | A=1000 B=0 R=0 | 304,024 | 210,002 | 208,020


utils/src/bitmap/roaring/mod.rs
```diff
@@ -0,0 +1,794 @@
...
+/// assert!(bitmap.contains(1500));
+/// assert!(!bitmap.contains(500));
+///
+/// // Iterate over values
+/// for value in bitmap.iter().take(10) {
+///     println!("{}", value);
+/// }
+/// ```
+#[derive(Clone, Debug, Default, PartialEq, Eq)]
+pub struct RoaringBitmap {
```
danlaine: nit: should we name this `Bitmap` so the import path has less stuttering?

i.e. `commonware-utils::bitmap::roaring::Bitmap` vs `commonware-utils::bitmap::roaring::RoaringBitmap`

utils/src/bitmap/roaring/container/run.rs
```diff
@@ -0,0 +1,701 @@
...
+    /// Creates a run container representing a fully saturated container [0, 65535].
+    #[inline]
+    pub fn full() -> Self {
+        Self {
+            runs: Vec::from([(0, u16::MAX)]),
+        }
+    }
+
+    /// Creates a run container from a bitmap container.
+    pub fn from_bitmap(bitmap: &bitmap::Bitmap) -> Self {
```
danlaine: nit: would be more rusty to implement `From<&bitmap::Bitmap> for Run`

see https://doc.rust-lang.org/std/convert/trait.From.html

danlaine: (Same for other `fn from_*` methods)

cronokirby:
One stylistic thing: in general Rust is pretty smart about inlining, so I would only manually add an annotation if:
- the method needs to be inlined across a crate (e.g. you don't want the overhead of calling it when crossing a boundary) (the compiler can ignore the annotation, but this forces inlining to be possible across crates)
- we actually determine a function to not be inlined in such a way as to pessimize a benchmark we care about, and then you usually want #[inline(always)] to force the compiler to do so.

utils/src/bitmap/roaring/mod.rs
```diff
@@ -0,0 +1,960 @@
...
+//! Array --[> 4096 values]--> Bitmap <--[run-count threshold]--> Run
+//! ```
+//!
+//! # References
+//!
+//! * <https://arxiv.org/pdf/1402.6407>: Better bitmap performance with Roaring bitmaps
+//! * <https://github.com/RoaringBitmap/RoaringFormatSpec>: Roaring Bitmap Format Specification
+//! * <https://github.com/RoaringBitmap/roaring-rs>: roaring-rs Crate
+
+pub mod container;
```
cronokirby: Do we actually want to expose the internal implementation?

utils/src/bitmap/roaring/mod.rs
```diff
@@ -0,0 +1,960 @@
...
+///     println!("{}", value);
+/// }
+/// ```
+#[derive(Clone, Debug, Default, PartialEq, Eq)]
+pub struct Bitmap {
+    /// Map from high 48 bits to container storing low 16 bits.
+    containers: BTreeMap<u64, Container>,
+}
+
+impl TryFrom<BTreeMap<u64, Container>> for Bitmap {
```
cronokirby: I think this function leaks the way Bitmaps are implemented in too much of a direct way. Do we expect users to want / need to construct bitmaps by manually constructing containers?

utils/src/bitmap/roaring/ops.rs
```diff
@@ -0,0 +1,1110 @@
+//! Set operations for roaring bitmaps.
+//!
+//! Provides union, intersection, and difference operations with optional early termination
+//! when a result limit is reached, along with the boolean predicates `is_subset` and
+//! `intersects` which short-circuit on first conclusive observation.
```
cronokirby, line 5: What's the thinking behind having this as standalone functions vs methods on the struct?

utils/src/bitmap/roaring/ops.rs
```diff
@@ -0,0 +1,1110 @@
...
+    for value in smaller.iter() {
+        if larger.contains(value) {
+            return true;
+        }
+    }
+    false
+}
+
+#[cfg(test)]
+mod tests {
```
cronokirby: One potentially useful strategy to simplify the test suite here is to generate operations, and check that the applying those operations to the roaring bitmap or the simpler bitmap produce the same results. Would help find a lot more edge cases than we have here.

utils/src/bitmap/prunable.rs
```diff
@@ -67,7 +67,10 @@ impl<const N: usize> Prunable<N> {
 
     /* Length */
 
-    /// Return the number of bits in the bitmap, irrespective of any pruning.
+    /// Return the highest bit position ever pushed, plus one. Pruning does not subtract from
+    /// this value: pruned positions stay counted because the bitmap is a sequential array
+    /// where positions retain meaning even when their values become inaccessible. Equivalent
+    /// to `pruned_bits() + (len of unpruned region)`.
```
cronokirby: Do we want to touch this file? Not opposed to more detailed comments, but this doesn't seem to add much information about how to use the method, just implementation details.

utils/src/bitmap/roaring/prunable.rs
```diff
@@ -0,0 +1,563 @@
...
+        if pruned_below & CONTAINER_MASK != 0 {
+            return Err(CodecError::Invalid(
+                "Prunable",
+                "pruned_below must be a multiple of 65536",
+            ));
+        }
+        let bitmap = Bitmap::read_cfg(buf, cfg)?;
+        // Validate invariant: no container key < pruned_below >> 16.
+        let target_key = pruned_below >> 16;
+        if let Some((&first_key, _)) = bitmap.containers().iter().next() {
```
cronokirby: Maybe use `BTreeMap::first_key_value` instead?

utils/src/bitmap/roaring/mod.rs
```diff
@@ -0,0 +1,960 @@
...
+    pub fn insert_range(&mut self, start: u64, end: u64) -> u64 {
+        if start >= end {
+            return 0;
+        }
+
+        let start_key = high_bits(start);
+        let end_key = high_bits(end.saturating_sub(1));
+
+        // Fast path: entire range fits in a single container
+        if start_key == end_key {
```
cronokirby: Does this fast path actually improve performance? Eyeballing things it would seem like just doing one iteration of the loop below would not be worse, and that avoids the need to maintain both code paths.

utils/src/bitmap/roaring/mod.rs
```diff
@@ -0,0 +1,960 @@
...
+        self.containers
+            .range(start_key..end_key_exclusive)
+            .flat_map(move |(&key, container)| {
+                container.iter().filter_map(move |index| {
+                    let value = combine(key, index);
+                    if value >= start && value < end {
+                        Some(value)
+                    } else {
+                        None
+                    }
```
cronokirby: Is there not some way to do the filtering at the container level, instead of iterating over the entirety of it? This seems sub-optimal, but my knowledge of this data structure is limited.

utils/src/bitmap/roaring/ops.rs
```diff
@@ -0,0 +1,1110 @@
...
+
+    for (&key, smaller_container) in smaller.iter() {
+        if let Some(larger_container) = larger.get(&key) {
+            if containers_intersect(smaller_container, larger_container) {
+                return true;
+            }
+        }
+    }
+    false
+}
```
cronokirby: For these two, I have the feeling that there might be a common helper with the functions above that actually compute the result. It seems like you could have an iterator as a helper, and then short circuit if that iterator is empty to produce these functions...something like that.

utils/src/bitmap/roaring/ops.rs
```diff
@@ -0,0 +1,1110 @@
...
+#[inline]
+fn containers_intersect(a: &Container, b: &Container) -> bool {
+    let (smaller, larger) = if a.len() <= b.len() { (a, b) } else { (b, a) };
+    for value in smaller.iter() {
+        if larger.contains(value) {
+            return true;
+        }
+    }
+    false
+}
```
cronokirby: I think these should live within container, rather than being defined here. It's better to encapsulate the details of container in the module it's defined, rather than reaching here for those.

cronokirby: That way, you could also have each struct contain a method for these operations. so then you have like:


```rust
match (a, b) {
  (Array(a), Array(b)) => a.union(b),
  (BM(a), BM(b)) => a.union(b),
  (a, b) => general method
}
```

utils/src/bitmap/roaring/ops.rs
```diff
@@ -0,0 +1,1110 @@
...
+    }
+
+    let mut result = BTreeMap::new();
+    let mut count = 0u64;
+
+    let mut a_iter = a_containers.iter().peekable();
+    let mut b_iter = b_containers.iter().peekable();
+
+    while count < limit {
+        match (a_iter.peek(), b_iter.peek()) {
```
cronokirby: You might be able to simplify this a lot by using the fact that `None < Some(_)`. I've run into generated loops like this a few times and I've always been able to simplify them significantly with a bit of elbow grease. IMO it's worth it makes it easier to understand the actual cases involved, and what needs to happen in each case.

utils/src/bitmap/roaring/ops.rs
```diff
@@ -0,0 +1,1110 @@
...
+pub fn intersection(a: &Bitmap, b: &Bitmap, limit: u64) -> Bitmap {
+    if limit == 0 {
+        return Bitmap::new();
+    }
+
+    let a_containers = a.containers();
+    let b_containers = b.containers();
+
+    // Fast path: single container with same key (common case)
+    if a_containers.len() == 1 && b_containers.len() == 1 {
```
cronokirby: Is this really all that much faster? I think a lookup of a key in a map with a single value should be pretty fast, no?

cronokirby: If you remove this, the if limit == 0 disappears too, because the branch inside of the loop effectively does that.

utils/src/bitmap/roaring/ops.rs
```diff
@@ -0,0 +1,1110 @@
...
+        if count >= limit {
+            break;
+        }
+
+        if let Some(b_container) = b_containers.get(&key) {
+            let (new_container, new_count) =
+                intersect_containers(a_container, b_container, limit - count);
+            if new_count > 0 {
+                result.insert(key, new_container);
+                count += new_count;
```
cronokirby: it's probably faster to do a checked decrement on a counter here, rather than checking if it's greater than limit at every iteration. Shouldn't make a big difference though.

cronokirby: This pattern appears in a few other places with the same suggestion.

utils/src/bitmap/roaring/fuzz.rs
```diff
@@ -0,0 +1,502 @@
...
+fn build_bitmap(values: &[BoundedU64]) -> Bitmap {
+    let mut bitmap = Bitmap::new();
+    for v in values.iter().take(MAX_VALUES) {
+        bitmap.insert(v.0);
+    }
+    bitmap
+}
+
+impl Plan {
+    pub fn run(self, _u: &mut Unstructured<'_>) -> arbitrary::Result<()> {
```
cronokirby, line 202: Comparing this with the naive bitmap impl we have would be a useful strategy as well, if we're willing to assume it's correct.

utils/src/bitmap/roaring/mod.rs
cronokirby: Having read this a bit more, right now a technical reason for needing to do this is to implement ops in an intrusive way, knowing exactly how container is implemented. If you instead implement those functions as methods on container, then we can preserve encapsulation, which makes things easier to reason about imo.

utils/src/bitmap/roaring/container/bitmap.rs
```diff
@@ -0,0 +1,952 @@
...
+//! run count rather than by cardinality alone, so a near-saturated bitmap
+//! with few gaps converts to Run while a dense one with many isolated
+//! runs stays as Bitmap.
+
+use crate::bitmap::roaring::container::{array, run};
+use bytes::{Buf, BufMut};
+use commonware_codec::{EncodeSize, Error as CodecError, Read, ReadExt, Write};
+
+/// Number of 64-bit words needed to store 65536 bits.
+pub const WORDS: usize = 1024;
```
cronokirby, line 22: I would compute BITS from words or vice versa. you can do `BITS: u32 = WORDS * 64;`

cronokirby (commented):
I think the big comment I'd have so far is that we could organize the code in more of a hierarchical / encapsulated way, where all the operations are methods on the Bitmap struct, and those defer to methods on the container struct, which differ to methods on each of the variants, such that Bitmap doesn't need to be aware that Container is even an enum at all, making it easier to play with other variants.

utils/src/bitmap/roaring/mod.rs
michaelkaplan13: Nice catch, I'll update. It will definitely be at least marginally better, though I'm not how significant it will be in most cases because if the range spans multiple containers, we still need to iterate through every value in the middle containers and map the `u16` entry back to its full `u64` value. We'll avoid iterating the full first and last containers though, so definitely an improvement either way.

utils/src/bitmap/roaring/mod.rs
michaelkaplan13: Very fair. No, probably not. I'm removing it now

utils/src/bitmap/prunable.rs
michaelkaplan13: Reverting now. It came from an interaction with an agent when I was trying to more fully understand the existing implementation, but agreed it should be added in source.

utils/src/bitmap/roaring/ops.rs
michaelkaplan13: After refactoring to make these methods on the container struct, the implementations became much shorter and much easier to read, so I don't think there is much benefit add a further shared helper. If you disagree, happy to make another pass though.

utils/src/bitmap/roaring/ops.rs
michaelkaplan13: Interesting, hadn't thought of that before.

I stared at this for a while and personally find it most readable like I have it now. It may not be exactly what you had in mind, but `next_key` is using the `None < Some(_)` ordering implicitly. Mind taking a look and letting me know what you think?

utils/src/bitmap/roaring/ops.rs
michaelkaplan13: Do you just mean decrementing down to 0 instead of up to limit? Made that change now, though it still has a check each iteration so I'm not sure it's all that much faster.

utils/src/bitmap/roaring/ops.rs
michaelkaplan13: Removed now 👍

utils/src/bitmap/roaring/ops.rs
michaelkaplan13: Great call, thanks. Updated now

utils/src/bitmap/roaring/container/array.rs
```diff
@@ -0,0 +1,801 @@
...
+
+    /// Inserts a range of values [start, end) into the container.
+    ///
+    /// Returns the number of values newly inserted.
+    ///
+    /// # Note
+    ///
+    /// After insertion, convert to `Bitmap` if the cardinality exceeds
+    /// [`MAX_CARDINALITY`].
+    pub fn insert_range(&mut self, start: u16, end: u16) -> usize {
```
danlaine: Should we use the `Range` type to express these bounds?

```suggestion
    pub fn insert_range(&mut self, range: Range<u16>) -> usize {
```

utils/src/bitmap/roaring/container/mod.rs
```diff
@@ -0,0 +1,810 @@
...
+use bytes::{Buf, BufMut};
+use commonware_codec::{EncodeSize, Error as CodecError, Read, ReadExt, Write};
+pub use run::Run;
+
+/// A container that stores u16 values using the most efficient representation.
+///
+/// Automatically converts between container types during insertion:
+/// - Array -> Bitmap when cardinality exceeds 4096
+/// - Bitmap -> Run when container becomes fully saturated (65536 values)
+#[derive(Clone, Debug, PartialEq, Eq)]
```
danlaine: Two `Container`s are only considered equal only if they are the same variant AND represent the same data. Is this the desired behavior? A user might expect two `Container`s (or types containing `Container`) to be equal without being the same variant.

utils/src/bitmap/roaring/container/bitmap.rs
```diff
@@ -0,0 +1,925 @@
...
+//!
+//! Stores values as a fixed 8 KB bit array (65,536 bits packed into 1024
+//! `u64` words). Cheaper than an [`Array`](super::array::Array) container
+//! above ~4,096 values, since Array storage scales with cardinality while
+//! the Bitmap's footprint is fixed.
+//!
+//! Tracks `run_count` (the number of maximal consecutive 1-bit sequences)
+//! incrementally on insert. The auto-conversion logic in [`super`] reads
+//! this to decide when to switch to a [`Run`](super::run::Run)
+//! representation; the transition is governed by a hysteresis band on
```
danlaine, line 11: > governed by a hysteresis band

I learned a new word today!

utils/src/bitmap/roaring/mod.rs
```diff
@@ -0,0 +1,902 @@
...
+    pub fn insert(&mut self, value: u64) -> bool {
+        let key = high_bits(value);
+        let index = low_bits(value);
+        self.containers.entry(key).or_default().insert(index)
+    }
+
+    /// Inserts a range of values [start, end) into the bitmap.
+    ///
+    /// Returns the number of values newly inserted.
+    pub fn insert_range(&mut self, start: u64, end: u64) -> u64 {
```
danlaine:
```suggestion
    pub fn insert_range(&mut self, range: Range<u64>) -> u64 {
```

utils/src/bitmap/roaring/container/mod.rs
```diff
@@ -0,0 +1,810 @@
...
+//!
+//! The Bitmap-Run thresholds form a hysteresis around the ~2048-run break-even point so
+//! that small run-count fluctuations near break-even do not cause thrashing.
+//!
+//! [`Bitmap::run_count`] is maintained incrementally on `insert` (O(1) per call) and
+//! recomputed via a single word scan on bulk paths (`insert_range`, AND/OR/XOR ops).
+//! Auto-conversion adds only a constant-time threshold compare per insert in the steady
+//! state, so callers never need to call an explicit `optimize()` — there is no such API.
+//!
+//! Decoded shapes from peer-supplied wire data are *trusted as-is*: a peer-encoded
```
danlaine: What's the advantage of this as opposed to enforcing that the decoded bitmap is canonically represented?

utils/src/bitmap/roaring/container/array.rs
```diff
@@ -0,0 +1,801 @@
...
+
+    /// Returns the maximum value in the container, if any.
+    pub fn max(&self) -> Option<u16> {
+        self.values.last().copied()
+    }
+
+    /// Computes the union of two arrays.
+    ///
+    /// Returns a new array containing all values from both, with optional limit.
+    pub fn union(&self, other: &Self, limit: usize) -> (Self, usize) {
```
danlaine, line 218: Can we document what limit does, exactly? Maybe I'm being dense but it isn't immediately obvious what it means to have a limit when unioning two bitmaps.

utils/src/bitmap/roaring/container/array.rs
```diff
@@ -0,0 +1,801 @@
...
+        }
+
+        let count = result.len();
+        (Self { values: result }, count)
+    }
+
+    /// Computes the intersection of two arrays.
+    ///
+    /// Returns a new array containing values present in both, with optional limit.
+    pub fn intersection(&self, other: &Self, limit: usize) -> (Self, usize) {
```
danlaine, line 301: Can we document what limit does, exactly? Maybe I'm being dense but it isn't immediately obvious what it means to have a limit when calculating the intersection of two bitmaps.

utils/src/bitmap/roaring/container/array.rs
```diff
@@ -0,0 +1,801 @@
...
+        }
+
+        let count = result.len();
+        (Self { values: result }, count)
+    }
+
+    /// Computes the difference (self - other).
+    ///
+    /// Returns a new array containing values in self but not in other, with optional limit.
+    pub fn difference(&self, other: &Self, limit: usize) -> (Self, usize) {
```
danlaine, line 365: Same comment as above about documenting `limit`

utils/src/bitmap/roaring/container/array.rs
```diff
@@ -0,0 +1,801 @@
...
+    }
+
+    /// Creates an array container from a sorted, deduplicated vector.
+    ///
+    /// # Panics
+    ///
+    /// Panics in debug mode if the values are not sorted or contain duplicates,
+    /// or if the length exceeds `MAX_CARDINALITY`.
+    #[cfg(any(test, feature = "arbitrary"))]
+    pub fn from_sorted_vec(values: Vec<u16>) -> Self {
```
danlaine: Should we implement `TryFrom<Vec<u16>> for Array` to make it apparent from the signature this is fallible? (Or just remove it, since it's only used in a few tests?)

utils/src/bitmap/roaring/container/bitmap.rs
```diff
@@ -0,0 +1,925 @@
...
+    pub const fn contains(&self, value: u16) -> bool {
+        let word_idx = (value >> 6) as usize;
+        let bit_idx = value & 63;
+        (self.words[word_idx] & (1u64 << bit_idx)) != 0
+    }
+
+    /// Inserts a value into the container.
+    ///
+    /// Returns `true` if the value was newly inserted, `false` if it already existed.
+    /// Maintains [`run_count`](Self::run_count) incrementally in O(1).
```
danlaine: Nit: I don't know we need to include this bit on the public doc comment; seems like an implementation detail.
```suggestion
```

utils/src/bitmap/roaring/container/bitmap.rs
```diff
@@ -0,0 +1,925 @@
...
+            (true, true) => self.run_count -= 1,
+        }
+        true
+    }
+
+    /// Inserts a range of values [start, end) into the container.
+    ///
+    /// Returns the number of values newly inserted. Recomputes [`run_count`](Self::run_count)
+    /// via a single O(WORDS) word scan after the bulk update.
+    pub fn insert_range(&mut self, start: u16, end: u16) -> u32 {
```
danlaine:
```suggestion
    pub fn insert_range(&mut self, range: Range<u16>) -> u32 {
```

utils/src/bitmap/roaring/container/bitmap.rs
```diff
@@ -0,0 +1,925 @@
...
+            (true, false) | (false, true) => {}
+            (true, true) => self.run_count -= 1,
+        }
+        true
+    }
+
+    /// Inserts a range of values [start, end) into the container.
+    ///
+    /// Returns the number of values newly inserted. Recomputes [`run_count`](Self::run_count)
+    /// via a single O(WORDS) word scan after the bulk update.
```
danlaine: Nit: like above, would remove this implementation detail (or move it into function body)
```suggestion
    /// Returns the number of values newly inserted.
```

utils/src/bitmap/roaring/container/bitmap.rs
```diff
@@ -0,0 +1,925 @@
...
+            words: &self.words,
+            word_idx: 0,
+            end_word: WORDS - 1,
+            end_mask: !0u64,
+            current_word: self.words[0],
+        }
+    }
+
+    /// Returns an iterator over values in `[start, end)`.
+    pub fn iter_range(&self, start: u16, end: u32) -> Iter<'_> {
```
danlaine: Feels kind of awkward to have the start and end bounds be different types (and this prevents using `Range<T>`

utils/src/bitmap/roaring/container/mod.rs
```diff
@@ -0,0 +1,810 @@
...
+        }
+    }
+
+    /// Inserts a range of values [start, end) into the container.
+    ///
+    /// Returns the number of values newly inserted.
+    ///
+    /// Automatically converts the container type if needed; see the [module
+    /// documentation](self) for the full transition table.
+    pub fn insert_range(&mut self, start: u16, end: u16) -> u32 {
```
danlaine:
```suggestion
    pub fn insert_range(&mut self, range: Range<u16>) -> u32 {
```

utils/src/bitmap/roaring/container/mod.rs
```diff
@@ -0,0 +1,810 @@
...
+    pub fn iter(&self) -> Iter<'_> {
+        match self {
+            Self::Array(a) => Iter::Array(a.iter()),
+            Self::Bitmap(b) => Iter::Bitmap(b.iter()),
+            Self::Run(r) => Iter::Run(r.iter()),
+        }
+    }
+
+    /// Returns an iterator over values in `[start, end)`.
+    pub fn iter_range(&self, start: u16, end: u32) -> RangeIter<'_> {
```
danlaine: Same comment as above -- having two different bound types feels a bit awkward and prevents usage of `Range` parameter

utils/src/bitmap/roaring/container/mod.rs
```diff
@@ -0,0 +1,810 @@
...
+    pub fn max(&self) -> Option<u16> {
+        match self {
+            Self::Array(a) => a.max(),
+            Self::Bitmap(b) => b.max(),
+            Self::Run(r) => r.max(),
+        }
+    }
+
+    /// Copies at most `limit` values from the container.
+    pub fn limit(&self, limit: u64) -> (Self, u64) {
```
danlaine, line 217: Nit: document second return type

utils/src/bitmap/roaring/container/mod.rs
```diff
@@ -0,0 +1,810 @@
...
+                break;
+            }
+            result.insert(value);
+            remaining -= 1;
+        }
+        (result, limit - remaining)
+    }
+
+    /// Computes the union of two containers, returning at most `limit` values.
+    pub fn union(&self, other: &Self, limit: u64) -> (Self, u64) {
```
danlaine, line 238: nit: document second return value

utils/src/bitmap/roaring/container/mod.rs
```diff
@@ -0,0 +1,810 @@
...
+                (None, None) => break,
+            }
+            remaining -= 1;
+        }
+
+        (result, limit - remaining)
+    }
+
+    /// Computes the intersection of two containers, returning at most `limit` values.
+    pub fn intersection(&self, other: &Self, limit: u64) -> (Self, u64) {
```
danlaine, line 297: Nit: document second return value

utils/src/bitmap/roaring/container/run.rs
```diff
@@ -0,0 +1,786 @@
...
+///
+/// Each entry is an inclusive range `[start, end]`. Entries are sorted by `start` and
+/// kept disjoint and non-adjacent: adjacent or overlapping runs are merged on insertion.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct Run {
+    /// Sorted vector of `(start, end)` inclusive ranges.
+    ///
+    /// Invariant: for any consecutive entries `(s1, e1)` and `(s2, e2)`, `e1 + 1 < s2`
+    /// (non-overlapping AND non-adjacent — adjacent runs would have been merged).
+    runs: Vec<(u16, u16)>,
```
danlaine, line 36:
```suggestion
    runs: Vec<Range<u16>>,
```

utils/src/bitmap/roaring/prunable.rs
```diff
@@ -0,0 +1,553 @@
...
+        self.bitmap.insert(value)
+    }
+
+    /// Inserts a range of values `[start, end)` into the bitmap. Returns the number of values
+    /// newly inserted.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `start < pruned_below()`.
+    pub fn insert_range(&mut self, start: u64, end: u64) -> u64 {
```
danlaine:
```suggestion
    pub fn insert_range(&mut self, range: Range<u64?>) -> u64 {
```

danlaine (commented):
There are several types that implement `union`, `intersection`, etc. These take in two references to the same type, and return an owned value of the type.

Should we have `union_with`, `intersection_with`, etc. variants of each method, where we mutate `&mut self` in-place instead of returning a new owned value? I think this would be more performant in some cases.

danlaine:
I mentioned using `Range` a few times above but we may actually want `commonware-utils::NonEmptyRange` in a bunch of places

utils/src/bitmap/roaring/container/mod.rs
michaelkaplan13: There technically isn't a single canonically representation of a roaring bitmap, because there are multiple possible container types to represent a given set of values (i.e. bitmap vs run depending on the run count, etc). So we were deferring the conversion between the container types to later calls to `insert` instead of checking and normalizing on deserialization. 

That being said, I think this reads as dangerous/confusing, so I'll move that normalization up explicitly.

michaelkaplan13:
> I mentioned using `Range` a few times above but we may actually want `commonware-utils::NonEmptyRange` in a bunch of places

From what I can gather, I feel like `Range` may honestly be preferable here so not all callers are forced to use `NonEmptyRange`, because empty ranges are already properly handled as no-ops rather than being invalid.

michaelkaplan13:
> There are several types that implement `union`, `intersection`, etc. These take in two references to the same type, and return an owned value of the type.
> 
> Should we have `union_with`, `intersection_with`, etc. variants of each method, where we mutate `&mut self` in-place instead of returning a new owned value? I think this would be more performant in some cases.

`union_with` could make sense, but it gets a little odd otherwise because the public roaring bitmap APIs are append only and intersection/difference/etc would allow for removing values. The `limit` param support would also be a little odd, since the existing bitmap may already have more values than `limit. If you feel strongly I can take another look at how to do it most cleanly though

utils/src/bitmap/roaring/container/run.rs
michaelkaplan13: I left this one as is, and explained in a comment: 
```
    /// `Range<u16>` is end-exclusive and cannot represent a run ending at `u16::MAX`
    /// without widening every stored bound.
```

danlaine: https://doc.rust-lang.org/std/ops/struct.RangeInclusive.html

^ This allows u16::max end bound (inclusive)

utils/src/bitmap/roaring/container/bitmap.rs
```diff
@@ -0,0 +1,926 @@
...
+
+impl From<&array::Array> for Bitmap {
+    fn from(array: &array::Array) -> Self {
+        let mut container = Self::new();
+        for value in array.iter() {
+            container.insert(value);
+        }
+        container
+    }
+}
```
danlaine, line 67:
```rs
impl From<&array::Array> for Bitmap {
    fn from(array: &array::Array) -> Self {
        let mut words = [0u64; WORDS];
        for value in array.iter() {
            let word_idx = (value >> 6) as usize;
            let bit_idx = value & 63;
            words[word_idx] |= 1u64 << bit_idx;
        }
        Self {
            words,
            cardinality: array.len() as u32,
            run_count: count_runs(&words),
        }
    }
}
```

I think this should be modestly faster and safe because `Array` is guaranteed to be sorted and non-repeating.

danlaine:
> > I mentioned using `Range` a few times above but we may actually want `commonware-utils::NonEmptyRange` in a bunch of places
> 
> From what I can gather, I feel like `Range` may honestly be preferable here so not all callers are forced to use `NonEmptyRange`, because empty ranges are already properly handled as no-ops rather than being invalid.

I feel like "you can only pass a non-empty range" is preferable to "you can pass an empty range, and if you do, it will be a no-op". I don't think a user would ever want to do a no-op, so why not use the compiler to explicitly prevent them from doing so?

michaelkaplan13:
> > > I mentioned using `Range` a few times above but we may actually want `commonware-utils::NonEmptyRange` in a bunch of places
> > 
> > 
> > From what I can gather, I feel like `Range` may honestly be preferable here so not all callers are forced to use `NonEmptyRange`, because empty ranges are already properly handled as no-ops rather than being invalid.
> 
> I feel like "you can only pass a non-empty range" is preferable to "you can pass an empty range, and if you do, it will be a no-op". I don't think a user would ever want to do a no-op, so why not use the compiler to explicitly prevent them from doing so?

I don't feel very strongly here either way, so I'll go ahead and make the change if you still think it's worth it after this :) 

That being said, this feels a little bit like saying "`add` should take a `NonZeroUint` because no one will ever want to add 0". Which may be the case, but there are feasible cases where it's nice for the implementation to handle this case without every call site having to validate for it. Something like:
```
fn get_remote_value() -> u64 {
    ...
}

then we can just write:
bitmap.insert_range(local_value..get_remote_value());

rather than:
if let Ok(missing) = NonEmptyRange::new(local_value..get_remote_value()) {
    bitmap.insert_range(missing);
}
```

utils/src/bitmap/roaring/container/bitmap.rs
michaelkaplan13: Very nice!

danlaine:
> > There are several types that implement `union`, `intersection`, etc. These take in two references to the same type, and return an owned value of the type.
> > Should we have `union_with`, `intersection_with`, etc. variants of each method, where we mutate `&mut self` in-place instead of returning a new owned value? I think this would be more performant in some cases.
> 
> `union_with` could make sense, but it gets a little odd otherwise because the public roaring bitmap APIs are append only and intersection/difference/etc would allow for removing values. The `limit` param support would also be a little odd, since the existing bitmap may already have more values than `limit. If you feel strongly I can take another look at how to do it most cleanly though

Yeah, fair point about the bitmap being append-only. I don't have a consumer in mind for the proposed `union_with` (or `union_mut` or whatever name) so I'll defer to you on whether to include it (or similar methods) or not.

utils/Cargo.toml
```diff
@@ -39,11 +39,14 @@ version = "0.2.15"
...
 
 [dev-dependencies]
+arbitrary = { workspace = true, features = ["derive"] }
 commonware-conformance.workspace = true
+commonware-invariants.workspace = true
 commonware-macros.workspace = true
 criterion.workspace = true
 num-bigint = { workspace = true, default-features = false }
 proptest.workspace = true
+roaring = "0.10"
```
patrick-ogrady: We should specify the full version (x.y.z)

utils/src/bitmap/roaring/prunable.rs
```diff
@@ -0,0 +1,555 @@
...
+    }
+
+    /// Prunes all values strictly less than `threshold`, rounding down to the nearest
+    /// container boundary (multiple of 65536).
+    ///
+    /// Returns the threshold actually applied (the rounded-down value), which is also the
+    /// new value of [`Self::pruned_below`].
+    ///
+    /// No-op if the rounded threshold is `<= pruned_below()`.
+    pub fn prune_below(&mut self, threshold: u64) -> u64 {
```
patrick-ogrady: In other areas, we just call this `prune`?

utils/src/bitmap/roaring/prunable.rs
```diff
@@ -0,0 +1,555 @@
...
+    fn read_cfg(buf: &mut impl Buf, cfg: &Self::Cfg) -> Result<Self, CodecError> {
+        let pruned_below = u64::read(buf)?;
+        if pruned_below & CONTAINER_MASK != 0 {
+            return Err(CodecError::Invalid(
+                "Prunable",
+                "pruned_below must be a multiple of 65536",
+            ));
+        }
+        let bitmap = Bitmap::read_cfg(buf, cfg)?;
+        // Validate invariant: no container key < pruned_below >> 16.
```
patrick-ogrady, line 231: nit: don’t split code blocks with a comment (either put at top of block or split block)

utils/src/bitmap/roaring/prunable.rs
```diff
@@ -0,0 +1,555 @@
...
+}
+
+#[cfg(feature = "arbitrary")]
+impl arbitrary::Arbitrary<'_> for Prunable {
+    fn arbitrary(u: &mut arbitrary::Unstructured<'_>) -> arbitrary::Result<Self> {
+        let mut p = Self {
+            bitmap: Bitmap::arbitrary(u)?,
+            pruned_below: 0,
+        };
+        // Pick an arbitrary threshold within the bitmap's value range and apply it.
```
patrick-ogrady, line 257: nit: add a \n

utils/src/bitmap/roaring/prunable.rs
```diff
@@ -0,0 +1,555 @@
...
+        p.prune_below(0);
+        assert_eq!(p.byte_size(), before);
+    }
+
+    #[cfg(feature = "arbitrary")]
+    mod conformance {
+        use commonware_codec::conformance::CodecConformance;
+
+        commonware_conformance::conformance_tests! {
+            CodecConformance<super::Prunable>,
```
patrick-ogrady, line 573: I think you should have `Prunable` directly because you did `super::*`?

utils/src/bitmap/roaring/mod.rs
```diff
@@ -0,0 +1,907 @@
...
+//! # Container Types
+//!
+//! | Type   | Use Case                                 | Storage                  |
+//! |--------|------------------------------------------|--------------------------|
+//! | Array  | Sparse data                              | Sorted `Vec<u16>`        |
+//! | Bitmap | Dense data with many disjoint runs       | `[u64; 1024]` (8 KB)     |
+//! | Run    | Data with few maximal runs (any density) | Sorted `Vec<(u16, u16)>` |
+//!
+//! Containers automatically convert between variants on each `insert` /
+//! `insert_range` to maintain a compact representation. The Bitmap→Run
```
patrick-ogrady, line 55: avoid using non-ASCII characters (like `→`)

utils/src/bitmap/roaring/mod.rs
```diff
@@ -0,0 +1,907 @@
...
+//! transition uses a hysteresis band on the bitmap's run count, so a container
```
patrick-ogrady, line 56: Sanity checking that the `hysteresis` doesn’t allow for multiple encodings of the same value? That seems ripe for missuse.

utils/src/bitmap/roaring/mod.rs
```diff
@@ -0,0 +1,907 @@
...
+            .and_then(|(&key, container)| container.min().map(|index| combine(key, index)))
+    }
+
+    /// Returns the maximum value in the bitmap, if any.
+    pub fn max(&self) -> Option<u64> {
+        self.containers
+            .last_key_value()
+            .and_then(|(&key, container)| container.max().map(|index| combine(key, index)))
+    }
+    #[cfg(test)]
```
patrick-ogrady, line 290: add \n

utils/src/bitmap/roaring/mod.rs
```diff
@@ -0,0 +1,907 @@
...
+                let remaining = MAX_KEY - prev_key;
+                if remaining == 0 {
+                    break;
+                }
+                prev_key.saturating_add(u.int_in_range(1..=remaining)?)
+            };
+            if key > MAX_KEY {
+                break;
+            }
+            // `Container::arbitrary` can produce an empty container (zero-length
```
patrick-ogrady, line 381: add \n

utils/src/bitmap/roaring/fuzz.rs
```diff
@@ -0,0 +1,569 @@
...
+        values: Vec<BoundedU64>,
+    },
+    MultipleRanges {
+        ranges: Vec<(BoundedU64, u16)>,
+    },
+    MixedOperations {
+        ops: Vec<MixedOp>,
+    },
+    /// Concentrate many inserts inside a single 16-bit shelf so that
+    /// Array→Bitmap→Run and Run→Bitmap auto-conversion paths get crossed
```
patrick-ogrady: more cases of arrows?

utils/src/bitmap/roaring/container/run.rs
```diff
@@ -0,0 +1,812 @@
...
+            ),
+            "expected Invalid(\"Run\", ...) error, got {result:?}"
+        );
+    }
+
+    #[test]
+    fn test_codec_rejects_adjacent_runs() {
+        // Second run is adjacent (start == prev_end + 1). A correctly-encoded Run
+        // would have merged these into a single run.
+        use bytes::BytesMut;
```
patrick-ogrady: imports should be at the top of the mod, not inline

utils/src/bitmap/roaring/container/mod.rs
```diff
@@ -0,0 +1,898 @@
...
+            c.insert(i * 3); // adds an isolated singleton → run_count += 1 each time
+        }
+        assert!(matches!(c, Container::Bitmap(_)));
+    }
+
+    #[cfg(feature = "arbitrary")]
+    mod conformance {
+        use commonware_codec::conformance::CodecConformance;
+
+        commonware_conformance::conformance_tests! {
```
patrick-ogrady, line 1199: You can encode the logic to inecrease/decrease runs here as well.

utils/src/bitmap/roaring/container/bitmap.rs
```diff
@@ -0,0 +1,938 @@
...
+        }
+    }
+
+    /// Returns the total memory footprint of this `Bitmap` in bytes.
+    ///
+    /// All storage is inline (no heap allocations), so this is just `size_of::<Self>()`,
+    /// which is approximately 8 KB regardless of cardinality. Available only for tests
+    /// and the `analysis` feature; not compiled into production builds.
+    #[cfg(any(test, feature = "analysis"))]
+    pub const fn byte_size(&self) -> usize {
```
patrick-ogrady: I think you may be able to use ~encode_size` for all of these? I suppose it isn’t exactly equal but would be close...

patrick-ogrady (commented):
Getting close!

utils/src/bitmap/roaring/prunable.rs
```diff
@@ -0,0 +1,555 @@
...
+    }
+
+    /// Inserts a range of values into the bitmap. Returns the number of values newly inserted.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `start < pruned_below()`.
+    pub fn insert_range(&mut self, range: Range<u64>) -> u64 {
+        let Range { start, end } = range;
+        assert!(
```
patrick-ogrady, line 107: `insert_range` and `iter_range` check `start >= pruned_below` before recognizing empty or reversed ranges. Those ranges do not access pruned values, and the underlying bitmap treats them as no-ops or empty iterators, so callers can panic on inputs like `50..50` or `50..10` after pruning. Handle `start >= end` first, then enforce the prune watermark for non-empty ranges.

Test command: `just test -p commonware-utils bitmap::roaring::prunable::tests::test_empty`

Observed result: both regression tests failed with `start pruned: 50 < pruned_below 131072`.

```patch
diff --git a/utils/src/bitmap/roaring/prunable.rs b/utils/src/bitmap/roaring/prunable.rs
index ae76366c7..53ae0934c 100644
--- a/utils/src/bitmap/roaring/prunable.rs
+++ b/utils/src/bitmap/roaring/prunable.rs
@@ -406,6 +406,27 @@ mod tests {
         let _ = p.iter_range(50..200_000);
     }
 
+    #[test]
+    fn test_empty_insert_ranges_below_pruned_are_no_ops() {
+        let mut p = Prunable::new();
+        p.insert(200_000);
+        p.prune_below(131_072);
+
+        assert_eq!(p.insert_range(50..50), 0);
+        assert_eq!(p.insert_range(50..10), 0);
+        assert_eq!(p.len(), 1);
+    }
+
+    #[test]
+    fn test_empty_iter_ranges_below_pruned_are_empty() {
+        let mut p = Prunable::new();
+        p.insert(200_000);
+        p.prune_below(131_072);
+
+        assert!(p.iter_range(50..50).next().is_none());
+        assert!(p.iter_range(50..10).next().is_none());
+    }
+
     #[test]
     fn test_iter() {
         let mut p = Prunable::new();
```
</details>

utils/src/bitmap/roaring/container/mod.rs
```diff
@@ -0,0 +1,898 @@
...
+
+        match self {
+            Self::Array(_) => {}
+            Self::Bitmap(b) if b.len() as usize <= array::MAX_CARDINALITY => {
+                self.convert_bitmap_to_array();
+            }
+            Self::Bitmap(b) if (b.run_count() as usize) < BITMAP_TO_RUN_THRESHOLD => {
+                self.convert_bitmap_to_run();
+            }
+            Self::Run(r) if r.run_count() > RUN_TO_BITMAP_THRESHOLD => {
```
patrick-ogrady: `normalize()` stops immediately after converting `Run` to `Bitmap`, so a decoded sparse `Run` with more than `RUN_TO_BITMAP_THRESHOLD` runs but cardinality still within `array::MAX_CARDINALITY` is not canonical after one decode. Re-encoding and decoding again can then change the encoding from bitmap to array. Make the `Run -> Bitmap` path continue normalization, or choose `Array` directly when the run cardinality is small enough.

Test command: `just test -p commonware-utils test_decode_normalization_is_idempotent_after_run_to_bitmap`

Observed result: the test fails at the idempotence assertion; `encoded_once` is a bitmap container encoding while the second decode re-encodes as an array container.

```patch
diff --git a/utils/src/bitmap/roaring/container/mod.rs b/utils/src/bitmap/roaring/container/mod.rs
index c594fe030..571bc6ceb 100644
--- a/utils/src/bitmap/roaring/container/mod.rs
+++ b/utils/src/bitmap/roaring/container/mod.rs
@@ -688,6 +688,22 @@ mod tests {
         assert!(matches!(decoded, Container::Run(_)));
     }
 
+    #[test]
+    fn test_decode_normalization_is_idempotent_after_run_to_bitmap() {
+        let mut run = Run::new();
+        for i in 0..=RUN_TO_BITMAP_THRESHOLD as u16 {
+            assert!(run.insert(i * 2));
+        }
+        assert_eq!(run.run_count(), RUN_TO_BITMAP_THRESHOLD + 1);
+        assert!(run.len() as usize <= array::MAX_CARDINALITY);
+
+        let decoded_once = Container::decode_cfg(Container::Run(run).encode(), &()).unwrap();
+        let encoded_once = decoded_once.encode();
+        let decoded_twice = Container::decode_cfg(encoded_once.clone(), &()).unwrap();
+
+        assert_eq!(encoded_once, decoded_twice.encode());
+    }
+
     #[test]
     fn test_auto_convert_array_to_bitmap() {
         let mut container = Container::new();
```
</details>

patrick-ogrady (commented):
automated findings

utils/src/bitmap/roaring/mod.rs
michaelkaplan13: I noticed this (originally from an agent's comments) in a self-review and started changing it, but then noticed that `→` is already used in comments elsewhere in the codebase. 

I tend to agree though. I'll remove now an follow up with a PR to switch the rest and steer agents away from it.

utils/src/bitmap/roaring/prunable.rs
michaelkaplan13: Not within the inner `conformance` module, only the outer `tests` module. Though I can add it within the `conformance` module as well of course, if preferred

utils/src/bitmap/roaring/mod.rs
michaelkaplan13: It does actually. On deserialization, we validate that the encoding is valid for the given value (i.e. array's can't have more than 4096 values, etc) and then "normalize" the value to its optimal representation, but we do not currently enforce a single canonical wire encoding for any given value.

michaelkaplan13: I'll update it so that strict normalization always occurs prior to serialization, and is validated upon deserialization as well.

utils/src/bitmap/roaring/container/mod.rs
```diff
@@ -0,0 +1,898 @@
...
+                result.insert(value);
+                remaining -= 1;
+            }
+        }
+
+        (result, limit - remaining)
+    }
+
+    /// Returns `true` if every value in this container is present in `other`.
+    pub fn is_subset(&self, other: &Self) -> bool {
```
danlaine, line 373: We could optimize this by adding `.is_subset` methods on the `Container` variants and doing something like:

```rs
pub fn is_subset(&self, other: &Self) -> bool {
        if self.len() > other.len() {
            return false;
        }
        match (self, other) {
            (Self::Array(a), Self::Array(b)) => return a.is_subset(b),
            (Self::Bitmap(a), Self::Bitmap(b)) => return a.is_subset(b),
            (Self::Run(a), Self::Run(b)) => return a.is_subset(b),
            _ => {}
        }
        self.iter().all(|value| other.contains(value))
    }
```

This especially matters for the `(Self::Bitmap(a), Self::Bitmap(b))` case because the `Bitmap::is_subset` scans word by word, not bit by bit:

```rs
/// Returns `true` if every bit set in `self` is also set in `other`.
    pub fn is_subset(&self, other: &Self) -> bool {
        self.words
            .iter()
            .zip(other.words.iter())
            .all(|(&a, &b)| (a & !b) == 0)
    }
```

Similar for `fn intersects` below.

utils/src/bitmap/roaring/container/mod.rs
```diff
@@ -0,0 +1,898 @@
...
+                    let inserted = run.insert_range(start..end);
+                    *self = Self::Run(run);
+                    return inserted;
+                }
+
+                // If result would exceed array capacity, convert to bitmap.
+                if a.len() + range_len > array::MAX_CARDINALITY {
+                    self.convert_array_to_bitmap();
+                    return self.insert_range(start..end);
+                }
```
danlaine: `range_len` is the _maximum_ number of elements that may be inserted into `self` -- fewer will be inserted if the range overlaps with already-set values. As a result it's possible that we over-eagerly convert to a bitmap here. For example, if `self` has `[0, 4000)` set and `range` is `[0, 1000)` then we will convert to a bitmap because `4000 + 1000 > 4096` but actually the insertion is a no-op. Should we calculate the number of elements that will be inserted to avoid this?

utils/src/bitmap/roaring/container/mod.rs
```diff
@@ -0,0 +1,898 @@
...
+    /// Returns the result container and the number of values in it.
+    pub fn intersection(&self, other: &Self, limit: u64) -> (Self, u64) {
+        if let (Self::Array(a), Self::Array(b)) = (self, other) {
+            let limit = usize::try_from(limit).unwrap_or(usize::MAX);
+            let (result, count) = a.intersection(b, limit);
+            return (Self::Array(result), count as u64);
+        }
+
+        if let (Self::Bitmap(a), Self::Bitmap(b)) = (self, other) {
+            let result = Bitmap::and_new(a, b);
```
danlaine, line 305: We always return the result as a `Bitmap` but `result` might have very few elements and be better represented as an `Array`. i.e. should we `normalize()` here?

utils/src/bitmap/roaring/container/mod.rs
```diff
@@ -0,0 +1,898 @@
...
+            }
+            Self::Run(r) => {
+                let inserted = r.insert_range(start..end);
+                if r.run_count() > RUN_TO_BITMAP_THRESHOLD {
+                    self.convert_run_to_bitmap();
+                }
+                inserted
+            }
+        }
+    }
```
danlaine, line 168: nit: we could avoid de-structuring and then re-assembling `range`

```suggestion
    pub fn insert_range(&mut self, range: Range<u16>) -> u32 {
        let range_len = range.len();
        if range_len == 0 {
            return 0;
        }

        match self {
            Self::Array(a) => {
                // Fast path: empty array with consecutive range - use Run container
                // directly. Run stores `(start, end)` pairs, so a single range is just
                // 4 bytes vs Array which would be 2 bytes per value.
                if a.is_empty() && range_len >= 64 {
                    let mut run = Run::new();
                    let inserted = run.insert_range(range);
                    *self = Self::Run(run);
                    return inserted;
                }

                // If result would exceed array capacity, convert to bitmap.
                if a.len() + range_len > array::MAX_CARDINALITY {
                    self.convert_array_to_bitmap();
                    return self.insert_range(range);
                }

                a.insert_range(range) as u32
            }
            Self::Bitmap(b) => {
                let inserted = b.insert_range(range);
                if (b.run_count() as usize) < BITMAP_TO_RUN_THRESHOLD {
                    self.convert_bitmap_to_run();
                }
                inserted
            }
            Self::Run(r) => {
                let inserted = r.insert_range(range);
                if r.run_count() > RUN_TO_BITMAP_THRESHOLD {
                    self.convert_run_to_bitmap();
                }
                inserted
            }
        }
    }
```

utils/src/bitmap/roaring/container/mod.rs
```diff
@@ -0,0 +1,898 @@
...
+
+    /// Inserts a range of values into the container.
+    ///
+    /// Returns the number of values newly inserted.
+    ///
+    /// Automatically converts the container type if needed; see the [module
+    /// documentation](self) for the full transition table.
+    pub fn insert_range(&mut self, range: Range<u16>) -> u32 {
+        let Range { start, end } = range;
```
danlaine: nit: similar to https://github.com/commonwarexyz/monorepo/pull/3687/files#r3189266383, could avoid de-structuring `range`

utils/src/bitmap/roaring/container/array.rs
```diff
@@ -0,0 +1,803 @@
...
+    /// Inserts a range of values into the container.
+    ///
+    /// Returns the number of values newly inserted.
+    ///
+    /// # Note
+    ///
+    /// After insertion, convert to `Bitmap` if the cardinality exceeds
+    /// [`MAX_CARDINALITY`].
+    pub fn insert_range(&mut self, range: Range<u16>) -> usize {
+        let Range { start, end } = range;
```
danlaine: nit: could avoid de-structuring like https://github.com/commonwarexyz/monorepo/pull/3687/files#r3189266383

utils/src/bitmap/roaring/container/array.rs
```diff
@@ -0,0 +1,803 @@
...
+        if self.values.is_empty() {
+            self.values = (start..end).collect();
+            return range_len;
+        }
+
+        // Fast path: range is entirely after all existing values
+        if start > *self.values.last().unwrap() {
+            self.values.extend(start..end);
+            return range_len;
+        }
```
danlaine, line 137: nit: I think it's slightly more rusty to get the value (`last`) at the same time we do the emptiness check

```suggestion
        // Fast path for empty array
        let Some(&last) = self.values.last() else {
            self.values = (start..end).collect();
            return range_len;
        };

        // Fast path: range is entirely after all existing values.
        if start > last {
            self.values.extend(start..end);
            return range_len;
        }
```

utils/src/bitmap/roaring/container/array.rs
```diff
@@ -0,0 +1,803 @@
...
+
+        // Fast path: range is entirely before all existing values
+        if end <= *self.values.first().unwrap() {
+            let mut new_vec: Vec<u16> = (start..end).collect();
+            new_vec.extend_from_slice(&self.values);
+            self.values = new_vec;
+            return range_len;
+        }
```
danlaine: Since we know `self.values` isn't empty, we can just directly index it -- we know `.first()` will return `Some` so this is logically equivalent. We can also pre-allocate the `Vec`.

```suggestion
        if end <= self.values[0] {
            let mut new_vec = Vec::with_capacity(range_len + self.values.len());
            new_vec.extend(start..end);
            new_vec.extend_from_slice(&self.values);
            self.values = new_vec;
            return range_len;
        }
```

utils/src/bitmap/roaring/container/mod.rs
```diff
@@ -0,0 +1,898 @@
...
+            _ => {}
+        }
+    }
+}
+
+impl PartialEq for Container {
+    fn eq(&self, other: &Self) -> bool {
+        self.len() == other.len() && self.iter().eq(other.iter())
+    }
+}
```
danlaine, line 523: Specializing is more performant (e.g. bitmaps check equality word-by-word not bit-by-bit)

```suggestion
impl PartialEq for Container {
    fn eq(&self, other: &Self) -> bool {
        if self.len() != other.len() {
            return false;
        }
        match (self, other) {
            (Self::Array(a), Self::Array(b)) => a == b,
            (Self::Bitmap(a), Self::Bitmap(b)) => a == b,
            (Self::Run(a), Self::Run(b)) => a == b,
            _ => self.iter().eq(other.iter()),
        }
    }
}
```

utils/src/bitmap/roaring/container/bitmap.rs
```diff
@@ -0,0 +1,938 @@
...
+impl Read for Bitmap {
+    type Cfg = ();
+
+    fn read_cfg(buf: &mut impl Buf, _cfg: &Self::Cfg) -> Result<Self, CodecError> {
+        let mut words = [0u64; WORDS];
+        for word in &mut words {
+            *word = u64::read(buf)?;
+        }
+        Ok(Self::from(words))
+    }
```
danlaine, line 485: We could do the read in one go, which should be faster

```suggestion
    fn read_cfg(buf: &mut impl Buf, _cfg: &Self::Cfg) -> Result<Self, CodecError> {
        if buf.remaining() < ENCODED_BYTES {
            return Err(CodecError::EndOfBuffer);
        }

        let mut bytes = [0u8; ENCODED_BYTES];
        buf.copy_to_slice(&mut bytes);

        let mut words = [0u64; WORDS];
        for (word, chunk) in words.iter_mut().zip(bytes.chunks_exact(8)) {
            *word = u64::from_be_bytes(
                chunk
                    .try_into()
                    .expect("chunks_exact(8) always yields 8-byte chunks"),
            );
        }
        Ok(Self::from(words))
    }
```

utils/src/bitmap/roaring/container/mod.rs
michaelkaplan13: Decided to use `is_empty()` rather than `len() == 0` in a few places as a slight readability preference, but otherwise applied the suggested changes here 🙏

utils/src/bitmap/roaring/mod.rs
```diff
@@ -0,0 +1,881 @@
...
+                if last_value == u16::MAX {
+                    None
+                } else {
+                    Some(last_value + 1)
+                }
+            } else {
+                None
+            };
+
+            let container = self.containers.entry(key).or_default();
```
danlaine, line 221: We can add a fast path for the full container case

```suggestion
            let container = self.containers.entry(key).or_default();
            if container_start == 0 && container_end.is_none() {
                inserted += container::bitmap::BITS as u64 - container.len() as u64;
                *container = Container::Run(container::Run::full());
                continue;
            }
```

danlaine: Or equivalently we could do a slightly bigger change and combine all the cases in one match:

```rs
for key in start_key..=end_key {
    let container_start = if key == start_key { low_bits(start) } else { 0 };
    let container_end = if key == end_key {
        low_bits(end - 1) as u32 + 1
    } else {
        container::bitmap::BITS
    };
    let container = self.containers.entry(key).or_default();
    match (container_start, container_end) {
        // Full shelf: install the optimal one-run representation directly.
        (0, container::bitmap::BITS) => {
            inserted += container::bitmap::BITS as u64 - container.len() as u64;
            *container = Container::Run(container::Run::full());
        }
        // Partial tail through u16::MAX; Range<u16> cannot express 65536.
        (_, container::bitmap::BITS) => {
            inserted += container.insert_range(container_start..u16::MAX) as u64;
            if container.insert(u16::MAX) {
                inserted += 1;
            }
        }
        // Ordinary in-shelf range with an end that fits in u16.
        (_, container_end) => {
            inserted += container.insert_range(container_start..container_end as u16) as u64;
        }
    }
}
```

utils/src/bitmap/roaring/container/mod.rs
```diff
@@ -0,0 +1,1206 @@
...
+    /// Returns the result container and the number of values in it.
+    pub fn intersection(&self, other: &Self, limit: u64) -> (Self, u64) {
+        if let (Self::Array(a), Self::Array(b)) = (self, other) {
+            let limit = usize::try_from(limit).unwrap_or(usize::MAX);
+            let (result, count) = a.intersection(b, limit);
+            return (Self::Array(result), count as u64);
+        }
+
+        if let (Self::Bitmap(a), Self::Bitmap(b)) = (self, other) {
+            let result = Bitmap::and_new(a, b);
```
danlaine, line 305: Maybe this is overkill but we could avoid some computation in `and_new`, which will process the entirety of the bitmaps, even if the limit is e.g. 1, and the intersection is found in the first word.

utils/src/bitmap/roaring/mod.rs
michaelkaplan13: I find the second slightly harder to read, so will stay with the first unless you have a strong preference

utils/src/bitmap/roaring/container/mod.rs
```diff
@@ -0,0 +1,1204 @@
...
+            let len = result.len() as u64;
+            if len <= limit {
+                let mut container = Self::Bitmap(Box::new(result));
+                container.normalize_in_memory();
+                return (container, len);
+            }
+            return Self::Bitmap(Box::new(result)).limit(limit);
+        }
+
+        let mut result = Self::new();
```
danlaine, line 258: Again, maybe overkill, but we could optimize the remaining cases not handled above by matching on more cases and taking advantage of the finer-grained use of internals (e.g. comparing by word, not bit).

danlaine: Same for `intersection`, `difference`

utils/src/bitmap/roaring/ops.rs
```diff
@@ -0,0 +1,818 @@
...
+            }
+        }
+
+        Self { containers: result }
+    }
+
+    /// Computes the intersection of two bitmaps, returning at most `limit` values.
+    ///
+    /// Pass `u64::MAX` for unlimited results.
+    pub fn intersection(&self, other: &Self, limit: u64) -> Self {
```
danlaine, line 63: We should follow the same `(smaller, larger)` pattern as `intersects` for better performance

utils/src/bitmap/roaring/container/mod.rs
michaelkaplan13: If you don't mind, I'd prefer to leave further optimizations to follow up PRs at this point, unless anything is egregious. That way we can better focus on the comparison between results for the focused optimizations as well.

danlaine: sgtm

utils/src/bitmap/roaring/container/mod.rs
danlaine: will defer optimizations to subsequent PR  https://github.com/commonwarexyz/monorepo/pull/3687#discussion_r3191016036

utils/src/bitmap/roaring/ops.rs
cronokirby: the comparison with 0 is "free" so you avoid an extra arithmetic instruction. That said the compiler can probably figure all of this out, and that's not really going to make much of a difference.

michaelkaplan13: Ahhh gotcha, thanks

danlaine (approved):

utils/src/bitmap/roaring/container/mod.rs
```diff
@@ -0,0 +1,1204 @@
...
+impl ExactSizeIterator for RangeIter<'_> {}
+
+/// Container type tags for serialization.
+const CONTAINER_TYPE_ARRAY: u8 = 0;
+const CONTAINER_TYPE_BITMAP: u8 = 1;
+const CONTAINER_TYPE_RUN: u8 = 2;
+
+impl Write for Container {
+    fn write(&self, buf: &mut impl BufMut) {
+        match self.canonical_kind() {
```
patrick-ogrady, line 627: This is a nice abstraction for this problem.

patrick-ogrady (approved):
very nice

PR #3690, opened 2026-04-29
[Storage] Flatten RootSpec from enum to struct

danlaine:
Follow up to #3667.

## Summary

`RootSpec` was introduced as a 2-variant enum:

```rust
pub enum RootSpec {
    Full  { bagging: Bagging },
    Split { inactive_peaks: usize, bagging: Bagging },
}
```

Per the type's docs and `hasher.rs`, `Full { F }` and `Split { 0, F }` produce **byte-identical roots** — the boundary commitment is keyed on the count, not the variant tag. The enum had two encodings of the same value.

This PR collapses it to a plain struct:

```rust
pub struct RootSpec {
    pub inactive_peaks: usize,
    pub bagging: Bagging,
}
```

## What changes

- **Type definition** (`merkle/mod.rs`). `pub enum` → `pub struct` with two public fields. Constants and constructors (`FULL_FORWARD`, `full`, `split_forward`, `split_backward`, `from_split_policy`) are kept. The trivial accessor methods (`bagging()`, `inactive_peaks()`, `has_inactive_prefix()`) are dropped — callers read fields directly.
- **`qmdb/any/db.rs`**. The `if self.split_root { Split{..} } else { Full{..} }` branch — which existed only because the enum forced two encodings — collapses into a single `RootSpec::from_split_policy(...)` call.
- **Variant constructions** (~7 sites). `RootSpec::Full { bagging }` → `RootSpec::full(bagging)`; `Split { .. }` → struct literal.
- **Accessor calls** (~25 sites across `merkle/`, `qmdb/`, fuzz targets, benches, the sync example). `spec.bagging()` → `spec.bagging`; `spec.inactive_peaks()` → `spec.inactive_peaks`.

## Why

- **One representation per spec.** No more "is this `Full { F }` or `Split { 0, F }`?" — the question goes away.
- **Smaller API surface to maintain.** Three method definitions and a redundant variant deleted.

### DIFF
--- a/examples/sync/src/databases/any.rs
+++ b/examples/sync/src/databases/any.rs
@@ -49,7 +49,7 @@ pub fn create_config(context: &impl BufferPooler) -> Config<Translator> {
         },
         translator: Translator::default(),
         split_root: true,
-        root_bagging: <mmr::Family as RootSpec>::root_spec(0).bagging(),
+        root_bagging: <mmr::Family as RootSpec>::root_spec(0).bagging,
     }
 }
 
--- a/storage/fuzz/fuzz_targets/proofs_malleability.rs
+++ b/storage/fuzz/fuzz_targets/proofs_malleability.rs
@@ -137,9 +137,7 @@ fn supported_root_specs<F: MerkleFamily>(merkle: &Mem<F, Digest>) -> Vec<RootSpe
     };
 
     push_unique(RootSpec::FULL_FORWARD);
-    push_unique(RootSpec::Full {
-        bagging: Bagging::BackwardFold,
-    });
+    push_unique(RootSpec::full(Bagging::BackwardFold));
     for inactive_peaks in 0..=peak_count {
         push_unique(RootSpec::split_forward(inactive_peaks));
         push_unique(RootSpec::split_backward(inactive_peaks));
--- a/storage/fuzz/fuzz_targets/qmdb_any_fixed_sync.rs
+++ b/storage/fuzz/fuzz_targets/qmdb_any_fixed_sync.rs
@@ -109,7 +109,7 @@ fn test_config<F: RootSpec>(test_name: &str, pooler: &impl BufferPooler) -> Conf
         },
         translator: TwoCap,
         split_root: true,
-        root_bagging: F::root_spec(0).bagging(),
+        root_bagging: F::root_spec(0).bagging,
     }
 }
 
--- a/storage/fuzz/fuzz_targets/qmdb_any_variable_sync.rs
+++ b/storage/fuzz/fuzz_targets/qmdb_any_variable_sync.rs
@@ -158,7 +158,7 @@ fn test_config<F: RootSpec>(
         },
         translator: TwoCap,
         split_root: true,
-        root_bagging: F::root_spec(0).bagging(),
+        root_bagging: F::root_spec(0).bagging,
     }
 }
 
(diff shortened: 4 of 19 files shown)


roberto-bayardo:
I think it's valuable to retain the split.   Some policies are "Full" (inactive_peaks don't matter) and some are Split (inactive peaks may matter).

danlaine:
> I think it's valuable to retain the split. Some policies are "Full" (inactive_peaks don't matter) and some are Split (inactive peaks may matter).

Are they really different, though? Full { F } and Split { 0, F } produce identical roots.

roberto-bayardo:
I also considering making inactive_peaks an option, but an explicit Full vs Split enum felt more explicit / clear.

roberto-bayardo:
> > I think it's valuable to retain the split. Some policies are "Full" (inactive_peaks don't matter) and some are Split (inactive peaks may matter).
> 
> Are they really different, though? Full { F } and Split { 0, F } produce identical roots.

Yes they produce identical roots; it's more for clarity of purpose.  If inactive peaks are irrelevant, it's feels a little weird that I still have to specify it in my root policy.  Don't feel super strongly either way though.

roberto-bayardo:
Mr. Claude, who is programmed to agree to the user's (my) priors in general, suggests:

> Do you think "RootSpec" should be flattened to just a simple struct containing inactive_peaks and the bagging policy?  I suppose one could argue that "Full" mode is just inactive_peaks: 0.  My       
  thinking was having the enum makes it clear that whether a peak is active or not is irrelevant, vs. the peaks actually always being active (inactive_peaks: 0).                                        
                                                                                                                                                                                                         
⏺ Your instinct is right — keep the enum. The enum is doing real work at the type level that a flat struct would erase.                                                                                  
                                                                                                                                                                                                       
  The key observation: Full and Split { inactive_peaks: 0 } are byte-equivalent at the wire level today (an implementation detail of the encoding eliding the count when 0), but they encode different
  policy intent:                              

  - Full = "this scheme does not commit to an inactive prefix at all." A permanent design choice.                                                                                                        
  - Split { inactive_peaks: 0 } = "this scheme commits to an inactive prefix; the current runtime count happens to be 0."
                                                                                                                                                                                                         
  The bytes-being-equal-today fact is exactly what caused the silent sync bug we just fixed — a split_root=false source produced Full bytes that the engine pre-fix verified with Split{0}, then later   
  commits diverged. Flattening to a struct would bake that confusion into the type system — there'd be no way to distinguish caller-declared policy from current runtime state. The fix we just landed   
  relies on Db::split_root: bool carrying the policy independently of the count; that bool is exactly the distinction the enum makes at the type level.                                                  
                                                                                                                                                                                                         
  Other reasons:                                                                                                                                                                                         
  - has_inactive_prefix() becomes meaningful again. With a flat struct it'd collapse to inactive_peaks > 0, which conflates "scheme has no concept of inactive prefix" with "currently 0". The enum      
  captures the difference.                                                                                                                                                                         
  - If you ever wanted to break the byte-equivalence (e.g., add a marker byte distinguishing Full from Split{0}), the enum lets you do it without renaming things at every call site. The flat struct
  would force a wire-format migration.                                                                                                                                                               
  - The runtime cost of the variant is negligible — every consumer destructures via .inactive_peaks() and .bagging() accessors anyway.                                                                   
                                                                                                                                      
  The only argument for flattening I can see is "simpler representation," and that's outweighed by the policy/state distinction the enum preserves. If the variant naming feels off (Full is a bit weak —
   could be Whole or NoSplit), that's bikeshed-able, but the structure is right.

PR #3660, opened 2026-04-24
[consensus/simplex] Avoid Blocking on Actor Reply

patrick-ogrady:
Related: #1110 

## TODO
- [x] Add channel with reservation support (on full): https://github.com/commonwarexyz/monorepo/pull/3683
- [ ] Remove all usage of unbounded channels

PR #3691, opened 2026-04-29
[resolver] Async Consumer Delivery

patrick-ogrady:
Related: #3660


resolver/src/p2p/engine.rs
```diff
@@ -74,6 +81,15 @@ pub struct Engine<
     /// Track the start time of fetch operations
     fetch_timers: HashMap<Key, histogram::Timer<E>>,
 
+    /// Holds futures that resolve once the `Consumer` has validated fetched data.
+    deliveries: AbortablePool<Delivery<P, Key>>,
+
+    /// Holds futures that notify the consumer about canceled fetches.
+    failures: FuturesPool<()>,
```
patrick-ogrady: not sure this is needed?

resolver/src/lib.rs
```diff
@@ -37,7 +43,9 @@ commonware_macros::stability_scope!(BETA {
 
         /// Let the consumer know that the data is not being fetched anymore.
         ///
-        /// The failure is used to indicate why.
```
patrick-ogrady, line 40: hmmm lol 🤔

resolver/src/p2p/engine.rs
```diff
@@ -189,6 +211,8 @@ impl<
             },
             on_stopped => {
                 debug!("shutdown");
+                self.delivery_aborters.clear();
```
BrendanChou: Ah, I was going to comment that this should be `self.deliveries.cancel_all()`, but I guess that clearing the hashmap will cancel all the futures. Perhaps a small comment would help clarify

resolver/src/p2p/engine.rs
```diff
@@ -288,12 +317,16 @@ impl<
...
 
                         // Drain timers and notify consumer
                         let removed = self.fetch_timers.len() as u64;
                         for (key, timer) in self.fetch_timers.drain() {
                             timer.cancel();
-                            self.consumer.failed(key, ()).await;
+                            let mut consumer = self.consumer.clone();
+                            self.failures.push(async move {
+                                consumer.failed(key, ()).await;
+                            });
```
BrendanChou: use `self.notify_failed(key);` instead since this code is just that function but put inline

resolver/src/p2p/engine.rs
```diff
@@ -252,17 +276,22 @@ impl<
...
                         }
                     }
                     Message::Retain { predicate } => {
                         trace!("mailbox: retain");
 
                         // Remove from fetcher
                         self.fetcher.retain(&predicate);
+                        self.delivery_aborters
+                            .extract_if(|k, _| !predicate(k))
+                            .count();
```
BrendanChou: since you don't need the count of extracted elements, just use `retain()` instead of `extract_if().count()` ?

resolver/src/p2p/engine.rs
```diff
@@ -74,6 +81,15 @@ pub struct Engine<
...
     fetch_timers: HashMap<Key, histogram::Timer<E>>,
 
+    /// Holds futures that resolve once the `Consumer` has validated fetched data.
+    deliveries: AbortablePool<Delivery<P, Key>>,
+
+    /// Holds futures that notify the consumer about canceled fetches.
+    failures: FuturesPool<()>,
+
+    /// Aborts pending deliveries when the corresponding fetch is canceled.
+    delivery_aborters: HashMap<Key, Aborter>,
```
BrendanChou: nit: define this directly below `deliveries` since they are related

resolver/src/p2p/engine.rs
```diff
@@ -276,68 +287,52 @@ impl<
...
-                        let removed = (before - self.fetch_timers.len()) as u64;
-                        if removed == 0 {
+                        if count == 0 {
                             self.metrics.cancel.inc(Status::Dropped);
                         } else {
-                            self.metrics.cancel.inc_by(Status::Success, removed);
+                            self.metrics.cancel.inc_by(Status::Success, count);
                         }
                     }
                     Message::Clear => {
```
BrendanChou, line 265: nit: I wonder if we can just re-use code here as-in above in the handling of `Message::Retain`. Clear is the same as `retain(|_| false)`

resolver/src/p2p/engine.rs
```diff
@@ -252,59 +254,44 @@ impl<
...
                         }
                     }
                 }
-                assert_eq!(self.fetcher.len(), self.fetch_timers.len());
+            },
+            // Handle completed consumer deliveries
+            delivery = self.inflight.next_delivery() => {
+                let Delivery { peer, key, valid } = match delivery {
+                    Ok(delivery) => delivery,
+                    Err(_) => continue,
```
patrick-ogrady, line 280: What happens in this case?

resolver/src/p2p/engine.rs
```diff
@@ -418,19 +412,23 @@ impl<
...
+        if valid {
+            self.metrics.fetch.inc(Status::Success);
+            self.inflight.complete(&key); // records duration on drop
             self.fetcher.clear_targets(&key);
             return;
         }
 
-        // If the data is invalid, we need to block the peer and try again
-        // (blocking the peer also removes any targets associated with it)
+        // Invalid data: clear the delivery aborter, block the peer, and retry.
```
patrick-ogrady: keep old comment

resolver/src/p2p/inflight.rs
```diff
@@ -0,0 +1,326 @@
...
+        let entry = self.entries.get_mut(&lookup_key).expect("inflight entry");
+        assert!(entry.delivery.replace(aborter).is_none());
+    }
+
+    /// Returns the next completed delivery as `(peer, key, valid)`, or [Aborted] if the
+    /// delivery was canceled. Clears the entry's delivery aborter so the slot is available
+    /// for a retry.
+    pub(super) async fn next_delivery(&mut self) -> Result<(P, Key, bool), Aborted> {
+        let (peer, key, valid) = self.deliveries.next_completed().await?;
+        if let Some(entry) = self.entries.get_mut(&key) {
```
BrendanChou: is it even possible for this to return `None`? Wondering if we should unwrap here...

resolver/src/p2p/engine.rs
patrick-ogrady: fetcher::clear avoids a decent bit of overhead but I'll dedup some of the surrounding code.

BrendanChou (dismissed):

PR #3692, opened 2026-04-29
[runtime/iobuf/freelist] add loom model

andresilva:
This PR adds loom coverage for the runtime buffer freelist, which relies on lock-free bitmap updates to transfer ownership of parked buffers between producers and consumers. The tests model the production freelist with loom atomics, thread-local probe state, and tracked `UnsafeCell` access so that missing synchronization around parking cells can be caught.

The models cover the main ownership-transfer paths: `put` publishing to `take`, `put_batch` publishing to `take_batch`, and both publish paths racing with `drain`. They also exercise same-word RMW races, stale relaxed candidate loads, concurrent producers merging disjoint bits, overlapping single and batch consumers, two batch consumers claiming from the same word, and concurrent drains.

The goal is to make the freelist's concurrency contract explicit: every free slot must be handed out at most once, no published slot may be lost, batch callbacks must only receive bits actually claimed by that caller, and buffer cell reads must be synchronized with the write that parked the buffer.

### DIFF
--- a/justfile
+++ b/justfile
@@ -60,6 +60,10 @@ test-benches crate test_flags='' lint_flags='':
 test *args='':
     cargo nextest run $@
 
+# Run loom tests
+test-loom *args='':
+    cargo nextest run --release --features loom --lib {{ args }} ::loom_tests::
+
 # Test the Rust documentation
 test-docs *args='--all':
     cargo test --doc --locked $@
--- a/runtime/Cargo.toml
+++ b/runtime/Cargo.toml
@@ -30,6 +30,7 @@ futures.workspace = true
 governor.workspace = true
 io-uring = { workspace = true, optional = true }
 libc.workspace = true
+loom = { workspace = true, optional = true }
 opentelemetry.workspace = true
 pin-project = { workspace = true, optional = true }
 prometheus-client.workspace = true
@@ -73,6 +74,7 @@ arbitrary = [
 ]
 bench = [ "dep:crossbeam-queue" ]
 external = [ "pin-project" ]
+loom = [ "dep:loom" ]
 test-utils = []
 iouring = [ "iouring-network", "iouring-storage" ]
 iouring-network = [ "io-uring" ]
--- a/runtime/src/iobuf/freelist.rs
+++ b/runtime/src/iobuf/freelist.rs
@@ -1155,7 +1155,8 @@ mod loom_tests {
     fn buffer() -> AlignedBuffer {
         // The payload itself is not important to the model. The aligned
         // allocation is enough to exercise ownership transfer through the
-        // parking cell and make double-reads/double-drops visible.
+        // parking cell. The per-test bitmasks and counts make duplicate
+        // ownership transfers visible.
         AlignedBuffer::new(64, 64)
     }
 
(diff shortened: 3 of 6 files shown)


runtime/Cargo.toml
```diff
@@ -73,6 +74,7 @@ arbitrary = [
 ]
 bench = [ "dep:crossbeam-queue" ]
 external = [ "pin-project" ]
+loom = [ "dep:loom" ]
```
patrick-ogrady, line 77: given we already expose test-related features here, I think its ok to leave as a feature (if accidentally activated via --all-features would fail loudly when called out side of a model)

runtime/src/iobuf/freelist.rs
patrick-ogrady, line 1: The inline transformations in the production code seem sound...didn't spend as much time on the tests themselves.

justfile
```diff
@@ -60,6 +60,10 @@ test-benches crate test_flags='' lint_flags='':
 test *args='':
     cargo nextest run $@
 
+# Run loom tests
+test-loom *args='':
+    cargo nextest run --release --features loom --lib {{ args }} ::loom_tests::
```
patrick-ogrady, line 65: `test-loom` is not scoped to `commonware-runtime`, so from this virtual-workspace root Cargo selects every default member's library tests and then unifies `commonware-runtime/loom` into all dependents. Running `CARGO_TARGET_DIR=target-loom-all cargo nextest list --release --features loom --lib ::loom_tests::` immediately starts compiling unrelated crates such as `commonware-broadcast`, `commonware-consensus`, and `commonware-deployer`. The `::loom_tests::` filter only narrows execution; it does not stop those extra crates from being built. That defeats the feature-isolation goal of this workflow and means the loom job can fail on unrelated crates whenever `commonware-runtime/loom` stops composing cleanly with their test builds. Scope the recipe to `-p commonware-runtime` (or `--manifest-path runtime/Cargo.toml`) so the job only builds and runs the intended loom models.

runtime/src/iobuf/freelist.rs
```diff
@@ -103,12 +103,26 @@
...
 use std::{
-    cell::{Cell, UnsafeCell},
+    cell::Cell,
     mem::MaybeUninit,
     num::{NonZeroU32, NonZeroUsize},
-    sync::atomic::{AtomicU64, AtomicUsize, Ordering},
+    sync::atomic::Ordering,
 };
 
+cfg_if::cfg_if! {
```
patrick-ogrady, line 112: Gating the main `Freelist` implementation on `feature = "loom"` is not purely additive: a downstream crate that enables `commonware-runtime/loom` now gets `loom::Atomic*` and `loom::UnsafeCell` in the production type itself, and those primitives panic whenever they are touched outside `loom::model`. That means non-test `Freelist` construction/use can start failing at runtime just by turning on the feature, even though the public method set is unchanged. Keep the std-backed implementation in the library and confine the loom substitution to test-only code (`cfg(loom)` or `cfg(all(test, feature = "loom"))`) so the feature cannot change consumer behavior.

```patch
diff --git a/runtime/src/iobuf/freelist.rs b/runtime/src/iobuf/freelist.rs
index 90bb145bc..23558eba4 100644
--- a/runtime/src/iobuf/freelist.rs
+++ b/runtime/src/iobuf/freelist.rs
@@ -1242,6 +1242,20 @@ mod loom_tests {
         assert_eq!(previous & mask, 0);
     }
 
+    #[test]
+    fn loom_feature_keeps_freelist_usable_outside_model() {
+        let result = std::panic::catch_unwind(|| {
+            let freelist = single_word_freelist(1);
+            freelist.put(0, buffer());
+            let _ = freelist.take();
+        });
+
+        assert!(
+            result.is_ok(),
+            "enabling the loom feature should not make ordinary freelist usage panic"
+        );
+    }
+
     #[test]
     fn put_publishes_before_take() {
         // `put` writes the parking cell before publishing the bit with a
```
</details>

justfile
patrick-ogrady: I assumed this was done to allow future loom work to run seamlessly. I'm ok with this (the filter should already make a big difference).

runtime/src/iobuf/freelist.rs
patrick-ogrady: Same comment as this: https://github.com/commonwarexyz/monorepo/pull/3692#discussion_r3183884519

patrick-ogrady: cfg would have the same issue, so don't think it is worth it.

justfile
andresilva: This is by design, in case more loom tests are added in the future, the only convention is that they should be added in a `loom_tests` module.

runtime/src/iobuf/freelist.rs
andresilva: Yes, and they get an immediate panic if they do this. I'm fine with doing the `cfg(all(test, feature = "loom"))` but seems like unnecessary churn.

patrick-ogrady (approved):

PR #3696, opened 2026-04-29
commonware-formatting

michaelkaplan13:
Moves hex related encoding/decoding formatting to a new crate called `commonware-formatting`. `commonware-formatting` does not depend on any other commonware crates (other than macros) such that it can be used anywhere else without risk of causing a circular dependency.

Also substitutes the implementations to use [`const-hex`](https://docs.rs/const-hex/latest/const_hex/), and adds benchmarks of the primary encoding/decoding functions. Before/After results:
| Operation                                         | Legacy (mean) | const-hex (mean) | Speedup |
| ------------------------------------------------- | ------------: | ---------------: | ------: |
| `hex` (encode)                                    |     112.08 µs |          0.40 µs |    278x |
| `from_hex` (decode)                               |      20.32 µs |          1.38 µs |   14.7x |
| `from_hex_formatted` (pristine)                   |      34.01 µs |         15.12 µs |    2.2x |
| `from_hex_formatted` (whitespace + `0x` prefix)   |      39.83 µs |         21.33 µs |    1.9x |

And the `Hex` newtype's `Display` impl (encoding onto the stack):
| Size  | const-hex (ns/iter) | legacy (ns/iter) | Speedup |
| ----: | ------------------: | ---------------: | ------: |
|    16 |                   6 |              129 |  21.5x |
|    32 |                   7 |              247 |  35.3x |
|    64 |                   9 |              472 |  52.4x |
|   256 |                  22 |             1839 |  83.6x |
|  1024 |                  76 |             7442 |  97.9x |
| 16384 |                1120 |           120445 | 107.5x |


formatting/src/lib.rs
```diff
@@ -0,0 +1,199 @@
...
+    ///
+    /// struct Digest([u8; 32]);
+    ///
+    /// impl core::fmt::Display for Digest {
+    ///     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+    ///         write!(f, "{}", Hex(&self.0))
+    ///     }
+    /// }
+    /// ```
+    pub struct Hex<T: AsRef<[u8]>>(pub T);
```
clabby, line 56: Would be nice to add an `impl<T: AsRef<[u8]> From<T> for Hex<T>`

formatting/src/lib.rs
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    /// Converts a hexadecimal string to bytes.
+    ///
+    /// Returns [None] if the input has odd length, contains non-hex characters,
+    /// or is otherwise malformed. Does not strip a leading `0x` / `0X` prefix
+    /// or any whitespace; see [from_hex_formatted] for that.
+    pub fn from_hex(s: &str) -> Option<Vec<u8>> {
+        // `const_hex::decode` strips a leading `0x`/`0X` prefix automatically;
+        // reject it explicitly to preserve historical behavior.
```
clabby: I think this API is fine to break, personally. What do you think about starting to allow `0x`/`0X` decoding?

clabby (commented):
Last thing is just to add a section to the `README.md`.

cryptography/src/ed25519/core/signature.rs
```diff
@@ -13,8 +13,8 @@ pub struct Signature {
 impl core::fmt::Debug for Signature {
     fn fmt(&self, fmt: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
         fmt.debug_struct("Signature")
-            .field("R_bytes", &hex(&self.R_bytes))
-            .field("s_bytes", &hex(&self.s_bytes))
+            .field("R_bytes", &Hex(&self.R_bytes))
+            .field("s_bytes", &Hex(&self.s_bytes))
```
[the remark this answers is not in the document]
clabby: I find this better 🤷

.github/workflows/publish.yml
```diff
@@ -40,6 +40,11 @@ jobs:
       continue-on-error: true
       env:
         CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
+    - name: Publish formatting
```
patrick-ogrady: If we use in `conformance`, we should publish higher up?

If not, we should use it in `conformance` (I think we had to work around this before and no longer need to!)

cryptography/src/bls12381/primitives/ops/mod.rs
```diff
@@ -373,23 +374,23 @@ mod tests {
...
 
     fn parse_private_key(private_key: &str) -> Result<Private, CodecError> {
-        let bytes = commonware_utils::from_hex_formatted(private_key).unwrap();
+        let bytes = commonware_formatting::from_hex_formatted(private_key).unwrap();
         Private::decode(bytes.as_ref())
     }
 
     fn parse_public_key(public_key: &str) -> Result<<MinPk as Variant>::Public, CodecError> {
-        let bytes = commonware_utils::from_hex_formatted(public_key).unwrap();
+        let bytes = commonware_formatting::from_hex_formatted(public_key).unwrap();
```
patrick-ogrady: seems like `from_base16_formatted` could be clearer

formatting/src/lib.rs
michaelkaplan13: I'm in favor, but don't know if it could be using anywhere already that expects the current behavior. If you think it's fine to break it though, seems like now would be a good time.

michaelkaplan13: I'll consolidate it into just `from_hex` that allows `0x`/`0X` (getting rid of `from_hex_formatted`)

cryptography/src/bls12381/primitives/ops/mod.rs
michaelkaplan13: Seems like that'd be inconsistent with `const-hex` 😅 (I don't actually care one way or another though)

conformance/src/lib.rs
```diff
@@ -188,7 +177,7 @@ pub async fn compute_conformance_hash<C: Conformance>(n_cases: usize) -> String
         hasher.update(&committed);
     }
 
-    hex_encode(&hasher.finalize())
+    commonware_formatting::hex(&hasher.finalize())
```
patrick-ogrady, line 180: ❤️ ❤️ ❤️ I knew I recalled one other place where we had to work around this...

cryptography/src/bls12381/primitives/ops/mod.rs
```diff
@@ -291,10 +292,10 @@ mod tests {
         let mut signatures = Vec::new();
         for line in MIN_SIG_TESTS.lines() {
             let parts: Vec<_> = line.split(':').collect();
-            let private_bytes = from_hex_formatted(parts[0]).unwrap();
+            let private_bytes = from_hex(parts[0]).unwrap();
```
patrick-ogrady, line 295: I think we want to retain `formatted` here?

patrick-ogrady: well, I guess we don't need to if the tests don't ... forgot what they do

michaelkaplan13: We got rid of the previous `from_hex` API that disallowed `0x`/`0X`, and renamed the existing `from_hex_formatted` to just `from_hex` for simplicity.

https://github.com/commonwarexyz/monorepo/pull/3696#discussion_r3164792596

cryptography/src/bls12381/scheme.rs
```diff
@@ -441,26 +442,22 @@ mod tests {
...
         )
     }
 
     fn parse_signature(signature: &str) -> Result<Signature, CodecError> {
-        Signature::decode(
-            commonware_utils::from_hex_formatted(signature)
-                .unwrap()
-                .as_ref(),
-        )
+        Signature::decode(commonware_formatting::from_hex(signature).unwrap().as_ref())
```
patrick-ogrady, line 460: Here is an example where from_hex_formatted is necessary (see the example signatures above?)

formatting/src/lib.rs
```diff
@@ -0,0 +1,179 @@
...
+    /// Converts bytes to a lowercase hexadecimal [String].
+    pub fn hex(bytes: &[u8]) -> String {
+        const_hex::encode(bytes)
+    }
+
+    /// Converts a hexadecimal string to bytes, stripping ASCII whitespace and an
+    /// optional `0x` / `0X` prefix. Commonly used in tests to encode external test
+    /// vectors without modification.
+    pub fn from_hex(s: &str) -> Option<Vec<u8>> {
+        // `const_hex::decode` already strips a leading `0x`/`0X` prefix, so we
```
patrick-ogrady: ah oops, missed this

formatting/src/lib.rs
```diff
@@ -0,0 +1,179 @@
...
+        // only need to remove ASCII whitespace ourselves.
+        let s = s.replace(['\t', '\n', '\r', ' '], "");
+        const_hex::decode(s).ok()
```
patrick-ogrady: It would be great to retian the from_formatted tests to ensure this behavior is maintained?

formatting/src/lib.rs
```diff
@@ -0,0 +1,179 @@
...
+    #[test]
+    fn test_hex_roundtrip() {
+        for bytes in [&[][..], &[0x01], &[0x01, 0x02, 0x03]] {
+            let encoded = crate::hex(bytes);
+            assert_eq!(from_hex(&encoded).unwrap(), bytes.to_vec());
+        }
+    }
+
+    #[test]
+    fn test_from_hex() {
```
patrick-ogrady, line 112: Ah, these cover formatted tests

formatting/src/benches/encode.rs
patrick-ogrady, line 1: I think you need to add this to the benchmarks action for it to get picked up?

cryptography/src/bls12381/primitives/ops/mod.rs
patrick-ogrady: 😜

formatting/src/lib.rs
michaelkaplan13: There was some very slight differences still actually. All cases should be covered exactly as before now though.

formatting/src/benches/encode.rs
michaelkaplan13: whoops, thanks

patrick-ogrady:
Two small LLM findings (I think the first is a decent suggestion but may be ok to support, second we should add an explicit test for):

```
  - Medium: scan now parses blob filenames with relaxed commonware_formatting::from_hex in runtime/src/storage/
    tokio/mod.rs:241 and runtime/src/storage/iouring.rs:221. Since formatting/src/lib.rs:28 strips whitespace
    and 0x, corrupted filenames like 0x626c6f62 or 62 6c6f62 are accepted as valid blob names instead of
    returning PartitionCorrupt. Storage writes canonical hex(name) filenames, so scan should either use a strict
    parser or verify hex(decoded) == name.
  - Low: from_hex documents 0X support in formatting/src/lib.rs:25, but the implementation delegates prefix
    handling to const_hex::decode, which only strips lowercase 0x. Meanwhile the macro path accepts 0X in
    formatting/src/hex_literal.rs:41. Add a 0X test and normalize it before decode, or narrow the docs.
```

conformance/src/lib.rs
michaelkaplan13: thanks for calling that out, didn't realize it was there.

runtime/src/storage/tokio/mod.rs
```diff
@@ -394,4 +400,43 @@ mod tests {
 
         let _ = std::fs::remove_dir_all(&storage_directory);
     }
+
+    #[tokio::test]
+    async fn test_scan_rejects_non_canonical_hex_file_names() {
```
patrick-ogrady, line 406: We have a number of cross-backend storage tests in `runtime/src/stroage/mod.rs`...I considered asking you to move this there but those all operate over the blob abstraction (no guarantee there is a filesystem for us to corrupt underneath).

formatting/src/benches/decode.rs
```diff
@@ -35,7 +38,13 @@ fn bench_decode(c: &mut Criterion) {
             },
         );
         c.bench_function(&format!("{}/input=ugly size={size}", module_path!()), |b| {
-            b.iter(|| black_box(commonware_formatting::from_hex(black_box(&ugly))));
+            let inputs: [&str; 2] = [&ugly_lower, &ugly_upper];
+            let mut idx = 0usize;
+            b.iter(|| {
+                let s = inputs[idx];
+                idx ^= 1;
```
patrick-ogrady, line 45: tricksy

formatting/src/lib.rs
```diff
@@ -26,10 +26,13 @@ commonware_macros::stability_scope!(BETA {
     /// optional `0x` / `0X` prefix. Commonly used in tests to encode external test
     /// vectors without modification.
     pub fn from_hex(s: &str) -> Option<Vec<u8>> {
-        // `const_hex::decode` already strips a leading `0x`/`0X` prefix, so we
-        // only need to remove ASCII whitespace ourselves.
         let s = s.replace(['\t', '\n', '\r', ' '], "");
-        const_hex::decode(s).ok()
+        // `const_hex::decode` only strips lowercase `0x`; handle uppercase ourselves.
```
patrick-ogrady: we prefer inline comments at the top of code blocks (and if inserting between contiguous lines that you put a \n)

formatting/src/lib.rs
```diff
@@ -26,10 +26,13 @@ commonware_macros::stability_scope!(BETA {
...
+        let stripped = s
+            .strip_prefix("0x")
```
patrick-ogrady: should we then defer this to decode?

runtime/src/storage/tokio/mod.rs
```diff
@@ -238,8 +238,14 @@ impl crate::Storage for Storage {
                 return Err(Error::PartitionCorrupt(partition.into()));
             }
             if let Some(name) = entry.file_name().to_str() {
-                let name = from_hex(name).ok_or(Error::PartitionCorrupt(partition.into()))?;
-                blobs.push(name);
+                let decoded = from_hex(name).ok_or(Error::PartitionCorrupt(partition.into()))?;
+                // Reject anything that isn't canonical lowercase hex (no `0x`
```
patrick-ogrady, line 241: would prefer:
```
// Comment...
let decoded = from_hex(name).ok_or(Error::PartitionCorrupt(partition.into()))?;
if hex(&decoded) != name {
   return Err(Error::PartitionCorrupt(partition.into()));
}

blobs.push(decoded);
```

formatting/src/benches/decode.rs
michaelkaplan13: Adding a separate "ugly_uppercase" benchmark felt like overkill, but missing the code path in the benchmark entirely didn't seem desireable either 🤷

formatting/src/lib.rs
michaelkaplan13: I included it for defensiveness, but it's inconsistent with comment (and is an unnecessary branch). Will remove

runtime/src/storage/tokio/mod.rs
michaelkaplan13: appreciate the context for the future. happy to move wherever is best

formatting/src/lib.rs
```diff
@@ -0,0 +1,187 @@
...
+    use alloc::{string::String, vec::Vec};
+    use core::fmt;
+
+    /// Converts bytes to a lowercase hexadecimal [String].
+    pub fn hex(bytes: &[u8]) -> String {
+        const_hex::encode(bytes)
+    }
+
+    /// Converts a hexadecimal string to bytes, stripping ASCII whitespace and an
+    /// optional `0x` (handled by `const-hex::decode`) / `0X` prefix. Commonly used
```
patrick-ogrady: We should not leak the inner impl details here ("handled by `const-hex::decode`" is not worth exposing).

formatting/src/lib.rs
```diff
@@ -0,0 +1,187 @@
...
+    /// in tests to encode external test vectors without modification.
```
patrick-ogrady: The LLMs love to add a where its commonly used sentence...I prefer removing that as it often gets very stale/feels weird for a public method doc to reflect on its usage in unrelated crates.

utils/fuzz/fuzz_targets/lib_functions.rs
```diff
@@ -90,24 +90,16 @@ fn fuzz(input: FuzzInput) {
         }
 
         FuzzInput::FromHex { hex_str } => {
-            if !hex_str.is_empty() && !hex_str.chars().any(|c| c.is_ascii_hexdigit()) {
-                assert_eq!(from_hex(&hex_str), None)
-            } else if let Some(decoded) = from_hex(&hex_str) {
+            // Round-trip property: anything that decodes successfully must
+            // re-encode and decode back to the same bytes.
```
patrick-ogrady: This is only true if the provided string doesn't start with `0x`/`0X`? I think this could fail with enough random permutations?

I think you may want to handle that explicitly (and remove the verbose comment).

patrick-ogrady (commented):
almost there

utils/fuzz/fuzz_targets/lib_functions.rs
michaelkaplan13: It's a (1) decode -> (2) encode-> (3) decode sequence, so even with whitespace/`0x`/`0X` the resutls of (1) and (3) should be equal. Will remove the comment though

patrick-ogrady: Ah, true true.

formatting/src/lib.rs
```diff
@@ -0,0 +1,187 @@
...
+
+    /// Converts a hexadecimal string to bytes, stripping ASCII whitespace
+    /// and an optional `0x` / `0X` prefix.
+    pub fn from_hex(s: &str) -> Option<Vec<u8>> {
+        // "0x" prefix stripping is handled by `const-hex::decode`.
+        let s = s.replace(['\t', '\n', '\r', ' '], "");
+        let stripped = s
+            .strip_prefix("0X")
+            .unwrap_or(&s);
+        const_hex::decode(stripped).ok()
```
cronokirby, line 33: If const-hex handles "0x" stripping, shouldn't we skip the step of doing it ourselves here?

Also, was the whitespace stripping a previous behavior? We should use a proper unicode class for that I think...

michaelkaplan13: const-hex handles stripping "0x" prefixes, but it does not handle stripping "0X" prefixes unfortunately. So we need to that ourselves if we want to keep that behavior consistent with before.

And yeah the whitespace stripping was the previous behavior/implementation [here](https://github.com/commonwarexyz/monorepo/blob/main/utils/src/lib.rs#L160). Happy to update it if desired though

patrick-ogrady: I'm cool to keep for now given how many other changes are in this PR.

PR #3699, opened 2026-04-30
[Storage] Deduplicate common QMDB code; refactor compact witness

danlaine:
Deduplicate shared QMDB batch-chain and compact-sync bookkeeping across keyless, immutable, compact, and any implementations, while tightening witness ownership and serving semantics.

- Add `qmdb::batch_chain` to centralize batch applicability checks, commit-floor validation, ancestor traversal, and shared `Bounds` state.
- Replace repeated batch boundary fields and apply-time validation in keyless, immutable, compact, and any batch implementations with `Bounds`.
- Rename and move compact witness logic from `compact_witness` into `qmdb::compact::witness`, with names centered on the persisted witness/cache rather than generic compact state.
- Add shared compact batch merkleization and reuse compact-sync serving/bootstrap helpers from the reorganized compact module.
- Move full Merkle pinned-node lookup into `merkle::persisted::full` so QMDB compact-sync serving does not duplicate frontier collection.
- Remove target selection from the compact DB state getter; compact DBs expose their current witness-backed state, and resolvers handle target matching separately.
- Decode persisted last-commit bytes as typed operations during witness reload, then extract metadata/floor at DB state boundaries.
- Replace the broad compact witness source trait and raw `RwLock<Witness>` DB fields with a tiny `witness::Cache` wrapper that only exposes read-with-closure and replace operations.
- Build compact-sync state from a borrowed cached witness so serving avoids cloning the full witness and only clones the owned data required by `State`.
- Update compact Merkle `sync_with_witness` so the metadata callback consumes the built witness, removing the forced `Clone` bound while keeping witness-cache updates inside the sync path.


storage/src/qmdb/batch_chain.rs
```diff
@@ -0,0 +1,363 @@
+//! Shared validation for QMDB batch chains.
```
roberto-bayardo, line 1: A bit more detail in the module doc header here could be helpful (took my a while to figure out what this module was about).

danlaine: Added some additional comments -- wdyt?

storage/src/qmdb/batch_chain.rs
```diff
@@ -0,0 +1,381 @@
...
+//!
+//! A batch chain is a linked sequence of in-memory batches built on top of a committed DB
```
roberto-bayardo: suggest dropping "committed"

roberto-bayardo: or replace with "base"

danlaine: Removed "committed"

storage/src/qmdb/batch_chain.rs
```diff
@@ -0,0 +1,380 @@
...
+//! and the inactivity floor declared by its commit. Ancestors (older batches in the chain that
+//! have not yet been applied to the DB) are tracked as [`AncestorBounds`] in newest-first
+//! order.
+//!
+//! Before applying a batch to the DB, [`Bounds::validate_apply_to`] checks two things shared
+//! across QMDB variants (any, immutable, keyless):
+//!
+//! - The batch is not stale: the current DB size must match either the batch's recorded
+//!   `db_size`, its `base_size`, or one of its ancestor boundaries.
+//! - Commit floors are monotonically non-decreasing across each unapplied ancestor and the
```
roberto-bayardo: Something that regularly confuses me is that (some) ancestors may actually have been applied, so saying "unapplied" here makes it sound like we stop if we hit one that is applied, but I don't think the implementation actually does?

storage/src/qmdb/compact/witness.rs
```diff
@@ -14,7 +14,7 @@
 //!
 //! The lifecycle in this file is:
 //!
-//! 1. Build a [`CachedServeState`] from the current in-memory tip.
+//! 1. Build a [`Witness`] from the current in-memory tip.
```
roberto-bayardo: Hmm, TBH I liked "ServeState" better (though I'd drop the "Cached" from the name).   

To me the witness (or specifically, the "witness to the last commit op") is just the last_commit_proof part of this.

`WitnessState` might work too?

storage/src/qmdb/compact/witness.rs
```diff
@@ -139,99 +205,92 @@ pub(crate) fn validate_inactivity_floor<F: Family>(
...
-            return Err(Error::FloorBeyondSize(ancestor_floor, ancestor_commit_loc));
-        }
-        prev_floor = ancestor_floor;
-    }
-    if tip_floor < prev_floor {
-        return Err(Error::FloorRegressed(tip_floor, prev_floor));
-    }
-    if tip_floor > tip_commit_loc {
-        return Err(Error::FloorBeyondSize(tip_floor, tip_commit_loc));
+/// Build a witness cache from compact state that was already authenticated by the caller.
```
roberto-bayardo: drop "cache"?

storage/src/qmdb/batch_chain.rs
danlaine: Updated comment

storage/src/qmdb/compact/witness.rs
danlaine: Renamed to ServeState

storage/src/qmdb/compact/witness.rs
danlaine: Done

roberto-bayardo (dismissed):

roberto-bayardo (approved):

PR #3701, opened 2026-04-30
[codec] Avoid cloning bytes in `Lazy::get`

clabby:
## Overview

Avoids cloning the pending `Bytes` in `Lazy::get` while materializing the value. This can add up in an application with a lot of `Lazy` usage.


patrick-ogrady (approved):
Wow, how did I miss this

andresilva (approved):

PR #3704, opened 2026-05-01
[cryptography] Golden DKG

### LINKED ISSUE #3179: Implement full Golden DKG with real ZK proofs
After we have a system for defining arithmetic circuits, have the gadgetry we need for Golden, and an implementation of the actual protocol with real ZK proofs, closing out the story.

cronokirby:
Closes #3179.

The performance is really not great, taking several minutes to deal at N = 100 (but, surprisingly, taking only 17 s instead of 30 to verify). But, we can deal with that subsequently.

Treating as a draft until the following are merged:

- #3410
- #3639


cronokirby:
We added some new types, hence a spurious "breaking-format"

cryptography/src/bls12381/golden_dkg.rs
```diff
@@ -0,0 +1,1797 @@
...
+    /// Verify the signature and extract the dealer's public key and log.
+    ///
+    /// Returns `None` if the signature is invalid.
+    pub fn identify(self) -> Option<(PublicKey, DealerLog)> {
+        let msg = self.log.encode();
+        if !self.dealer.verify(NAMESPACE, &msg, &self.signature) {
+            return None;
+        }
+        Some((self.dealer, self.log))
+    }
```
guruvamsi-policharla: Here we parse out a public key from a signed dealer log, but I'm guessing we check that the public key is in the set of dealers somewhere else? Wondering if we should pass the dealer public keys to this method and have it discard the log early if there is no match.

cryptography/src/zk/bulletproofs/circuit.rs
```diff
@@ -0,0 +1,1800 @@
...
+    // generating more challenges is needed instead, but for large fields, using powers lets us
+    // sample less randomness.)
+    //
+    // At this point, it's convenient to fold these challenges into the weights:
+    //
+    //   θ_j := <Θ_ij, z z^i>
+    //   λ_j := <Λ_ij, z z^i>
+    //   ρ_j := <Ρ_ij, z z^i>
+    //   ω_j := <Ω_ij, z z^i>
+    //   κ := <c_i, z z^i>
```
guruvamsi-policharla, line 1021: Random thought: if you had a circuit specific pre-processing with trusted setup, maybe you can avoid these MSMs and make the protocol more efficient

cryptography/src/zk/bulletproofs/circuit.rs
```diff
@@ -0,0 +1,1800 @@
...
+    // into a single inner product. The trick is that if we create polynomials:
+    //
+    //   f_i(X) := a_i X + c_i X^2
+    //   g_i(X) := b_i X + d_i
+    //
+    // then the 2nd degree of <f_i(X), g_i(X)> is <a_i, b_i> + <c_i, d_i>.
+    //
+    // So, we can check that:
+    //
+    //   t X^2 = <f_i(X), g_i(X)>
```
guruvamsi-policharla, line 1071: is t = <a_i, b_i> + <c_i, d_i>?

cryptography/src/zk/bulletproofs/circuit.rs
```diff
@@ -0,0 +1,1800 @@
...
+    // It seems like we're stuck here, because <r_i, H_i> doesn't match the <y^i r_i, H_i>
+    // that we need inside of S_1. However, the IPA already lets us treat the right-side
+    // basis as y^-i H_i by setting claim.y = y^-1. In terms of the original H_i that this
+    // implementation uses in its MSMs, the public pieces become:
+    //
+    //   P_0 = <y^-i ω_i - 1, H_i>
+    //   P_1 = <y^-i ρ_i, G_i> + <y^-i λ_i, H_i>
+    //   M   = <l_i, G_i> + <r_i, H_i> + m ~B
+    //   O   = <o_i, G_i> + ~o ~B
+    //  ~M   = <~l_i, G_i> + <~r_i, H_i> + ~m ~B
```
guruvamsi-policharla, line 1161: Could you expand a bit on what's happening here or maybe add a pointer to the relevant section in the paper. Not sure if I follow this section

cryptography/src/zk/pedersen_to_plain.rs
```diff
@@ -0,0 +1,632 @@
...
+//!
+//! This is useful when one part of a protocol wants to work with a plain
+//! commitment while another part wants the same value hidden behind a Pedersen
+//! commitment. The proof lets the prover bridge those two views without
+//! revealing either the committed value or the Pedersen blinding.
+//!
+//! Concretely, the prover shows knowledge of openings `(x, x_blind)` such that:
+//!
+//! - `X = x * G`, and
+//! - `C = x * G + x_blind * H`.
```
guruvamsi-policharla, line 18: do you need to prove knowledge of both x and x_blind?
wouldn't it be sufficient to prove that you know dlog_H(C * X^-1)? This would halve the proof size

cryptography/src/bls12381/golden_dkg.rs
cronokirby: I think the best place to do this is in `select` which is responsible for picking which logs to use. Natural to filter out unknown dealers there as well. I think we're missing such a filter though!

cryptography/src/zk/bulletproofs/circuit.rs
cronokirby: If you follow this line of thinking and add pairings, you get Groth16 basically

cryptography/src/zk/bulletproofs/circuit.rs
cronokirby: `2nd_degree(t) = <a_i, b_i> + <c_i, d_i>` so the idea is that rather than checking that the 2nd degree is equal, we can just check that the polynomials are equal everywhere

cryptography/src/zk/pedersen_to_plain.rs
cronokirby: Yes, we should do that as an optimization, because we pick x_blind = 0

Cargo.toml
```diff
@@ -84,6 +84,12 @@ suspicious_op_assign_impl = "allow"
 [workspace.dependencies]
 anyhow = { version = "1.0.99", default-features = false }
 arbitrary = "1.4.1"
+ark-ec = { version = "0.5", default-features = false }
```
patrick-ogrady: can you pin to a fully specified version?

math/conformance.toml
patrick-ogrady, line 1: AFAICT we aren't actually changing the format of these things...just how we generate?

cronokirby: Indeed. The source of the change is that I bumped the size of the test field and group, so that they're large enough for random challenges during testing. This changed the conformance values which depended on them. So, we did, technically, break the value of the test fields, but not any actual types like Poly.

patrick-ogrady: I suppose the other conformance tests would light up if so 👁️

patrick-ogrady: I'm going to manually override the breaking format check.

cryptography/src/bls12381/golden_dkg.rs
```diff
@@ -0,0 +1,2200 @@
...
+                    }
+                }
+            }
+
+            // Optionally add fake mask commitments.
+            for d in 0..num_dealers {
+                if u.ratio(1, 8)? {
+                    plan.fake_masks.insert(d);
+                }
+            }
```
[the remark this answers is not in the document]
cronokirby: This is stale; we currently use a proper identifying batch check, rather than all or nothing check.

PR #3711, opened 2026-05-04
[runtime/iouring/waker] add loom model

andresilva:
This PR adds loom coverage for the runtime io_uring waker, which coordinates producer-side publishes and out-of-band wakes with loop-side `eventfd`/`futex` parking through a packed atomic state word. The tests model the production sequence and wait-bit state with loom atomics, replace kernel wake surfaces with loom-visible `futex` and `eventfd` models, and use a minimal queued-request model so missing release/acquire synchronization around request publication can be caught.

The models cover the main wakeup paths: `publish` publishing to `pending`, `wake` publishing to `clear_wait`, armed `publish`/`wake` races through the `eventfd` path, fully-idle `futex` parking, `eventfd` wake deduplication, sticky unarmed wakes, final-handle drop wakeups, sequence wraparound, and repeated loop drain cycles. They also include deterministic generated producer programs for producer-only, `eventfd`-loop, and `futex`-loop histories.

The goal is to make the waker's concurrency contract explicit: every published submission must be observed exactly once, pure wakes must not advance the submission sequence, wait targets must be cleared after wakeup, unarmed wakes must not queue `eventfd` readiness, `eventfd` readiness must be coalesced and acknowledged correctly, and queue writes made before `publish`/`wake` must be visible after the loop observes `pending` or resumes through `clear_wait`.

Depends on #3692.

### DIFF
--- a/runtime/src/iouring/waker.rs
+++ b/runtime/src/iouring/waker.rs
@@ -1322,11 +1322,6 @@ mod loom_tests {
         }
     }
 
-    // Number of deterministic `ProducerOp` programs to generate per test.
-    const GENERATED_CASES: usize = 8;
-    // Number of operations in each generated `ProducerOp` program.
-    const OPS_PER_PROGRAM: usize = 3;
-
     #[test]
     fn publish_pending_pairing() {
         // `publish` must make the producer's earlier enqueue-side write visible
@@ -1868,7 +1863,7 @@ mod loom_tests {
     #[test]
     fn generated_producer_only_programs() {
         // Generate deterministic producer-only programs before entering loom,
-        // then model each case with four concurrent producers. Each producer
+        // then model each case with two concurrent producers. Each producer
         // runs a short sequence of `publish()` and out-of-band `wake()` calls
         // without any loop thread consuming them.
         //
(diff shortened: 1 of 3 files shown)


runtime/src/iouring/waker.rs
```diff
@@ -113,9 +122,36 @@ impl Drop for ArmGuard<'_> {
...
+
+/// Loom-only model of the waker state.
+///
+/// Loom cannot observe real futexes, eventfds, or io_uring CQEs, so this
+/// variant keeps the same packed atomic state as the production waker and
+/// replaces the kernel wake surfaces with userspace condvar models. The goal is
+/// to model the producer/loop atomic protocol closely enough for loom to
+/// explore memory orderings and wake races. It is not a model of kernel CQE
+/// ordering, `io_uring_enter`, or wake-poll rearm behavior.
+#[cfg(feature = "loom")]
```
patrick-ogrady, line 141: I’m tempted to suggest putting the real impl in a single block before the loom impl rather than interleaving but won’t block on it.

patrick-ogrady (dismissed):

PR #3693, opened 2026-04-29
[storage/qmdb/current] support split (pyramid) bagging in qmdb-current

roberto-bayardo:
## Summary

- Switch QMDB operation roots to pyramid bagging (split + backwards-fold active region).
- Update Current grafted root and range-proof reconstruction so MMR/MMB proofs handle inactive prefixes, grafted peak regrouping, and partial chunks consistently.
- Align Current sync, compact sync verification, conformance fixtures, and fuzz tampering coverage with the new root semantics.

PR #3712, opened 2026-05-04
[Storage] Simplify QMDB root policy configuration

danlaine:
Follow up to #3667

## Summary

- Remove public `AnyConfig` fields that let callers choose the operations-root shape directly.
- Make public `Any` databases use one root shape: split-root mode with the bagging policy defined by the selected family.
- Keep `Current`'s different operations-root shape as an internal implementation detail.

## Background

QMDB has an operations root over its authenticated log. Two details determine the meaning of that root:

- `split_root` decides whether the operations root also commits to the inactive-prefix boundary.
- `root_bagging` decides how Merkle peaks are folded into the root.

Those are not independent user preferences. They must agree with the database family and with the sync code that later verifies or reconstructs the database.

Before this change, `AnyConfig` exposed both fields directly. That meant a caller could choose a family, then separately choose root settings that did not belong with that family or with `Any` sync. The database might initialize, but its root semantics would no longer be the ones the type implies.

This PR makes the rule explicit: public `Any` databases always use split-root mode, and their bagging policy comes from the selected family. `Current` is the exception because its canonical root commits activity through the grafted tree, so its internal `Any` operations root stays plain. That exception now lives only in `Current`'s private initialization path instead of being exposed through `AnyConfig`.

## Changes

- Remove `split_root` and `root_bagging` from `qmdb::any::Config` and update examples, fuzz targets, benches, conformance helpers, and tests.
- Pass `split_root` and `ops_root_bagging` only through the crate-private `any::init_with_bitmap` helper used by `Current`.
- Change QMDB sync traits to expose root bagging as `ROOT_BAGGING`, since it is now fixed by the database type rather than read from config.
- Make `Any` sync reconstruct databases in split-root mode and verify target state using the family-derived inactive peak count and bagging policy.
- Simplify test helpers by removing now-unused family generic parameters from config builders.
- Remove the Any sync regression test for preserving plain-root mode, since plain-root mode is no longer public `Any` behavior.

### DIFF
--- a/examples/sync/src/databases/any.rs
+++ b/examples/sync/src/databases/any.rs
@@ -17,7 +17,6 @@ use commonware_storage::{
             FixedConfig as Config,
         },
         operation::Committable,
-        Bagging,
     },
 };
 use commonware_utils::{NZUsize, NZU16, NZU64};
@@ -49,8 +48,6 @@ pub fn create_config(context: &impl BufferPooler) -> Config<Translator> {
             page_cache,
         },
         translator: Translator::default(),
-        split_root: true,
-        root_bagging: <mmr::Family as Bagging>::BAGGING,
     }
 }
 
--- a/storage/fuzz/fuzz_targets/qmdb_any_fixed_sync.rs
+++ b/storage/fuzz/fuzz_targets/qmdb_any_fixed_sync.rs
@@ -91,7 +91,7 @@ impl<'a> Arbitrary<'a> for FuzzInput {
 
 const PAGE_SIZE: NonZeroU16 = NZU16!(129);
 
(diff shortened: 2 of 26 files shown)


roberto-bayardo:
If we're going to remove ability for user configuration of the root policy, we should probably also force the MMR to also always backwards-bag active peaks (see latest blog post draft).

roberto-bayardo:
And P.S.: The current qmdb policy is updated to be split in https://github.com/commonwarexyz/monorepo/pull/3693

roberto-bayardo:
Yeah let's just go ahead and get rid of the qmdb::Bagging trait entirely. Always forward fold inactive, always backwards-fold active.

roberto-bayardo:
Will merge this and take care of the final cleanup, thanks!

PR #3717, opened 2026-05-05
[storage/qmdb/current] have current proof types implement codec

### LINKED ISSUE #3705: [storage/qmdb] Add codec impls to proof types (RangeProof, OperationProof, KeyValueProof, ExclusionProof)
## Summary

The QMDB proof types in `commonware-storage::qmdb::current` are not encodable/decodable via `commonware-codec`. This makes it awkward for downstream consumers to ship these proofs over the wire (e.g. inside gRPC payloads or onchain commitments). This issue requests `Read` / `Write` / `EncodeSize` impls on the QMDB proof composites, matching what already exists for the underlying `merkle::Proof<F, D>`.

## Affected types (v2026.4.0)

All in the `commonware-storage` crate:

- [`RangeProof<F, D>`](https://github.com/commonwarexyz/monorepo/blob/v2026.4.0/storage/src/qmdb/current/proof.rs#L222) — `storage/src/qmdb/current/proof.rs`
- [`OperationProof<F, D, N>`](https://github.com/commonwarexyz/monorepo/blob/v2026.4.0/storage/src/qmdb/current/proof.rs#L541) — `storage/src/qmdb/current/proof.rs`
- [`KeyValueProof<F, K, D, N>`](https://github.com/commonwarexyz/monorepo/blob/v2026.4.0/storage/src/qmdb/current/ordered/db.rs#L26) — `storage/src/qmdb/current/ordered/db.rs`
- [`ExclusionProof<F, K, V, D, N>`](https://github.com/commonwarexyz/monorepo/blob/v2026.4.0/storage/src/qmdb/current/ordered/mod.rs#L34) — `storage/src/qmdb/current/ordered/mod.rs`

Each currently has only `#[derive(Clone, Eq, PartialEq, Debug)]`. Searches for `impl Write`/`impl Read`/`impl EncodeSize` on these types in `commonware-storage` returned zero matches at v2026.4.0, and no `CodecConformance<...>` entries exist for them either.

## Motivation

[Makechain](https://github.com/officialunofficial/makechain) (a downstream consumer using QMDB as its sole persistence layer at `v2026.4`) needs to serialize these proofs across gRPC for `GetOperationProof`, `GetExclusionProof`, `GetStorageQuotaProof`, and `GetCompoundProof` RPCs. Because the proof structs themselves don't implement Codec, makechain currently maintains a parallel set of `*Wire` wrappers that re-implement encoding by hand — see [`crates/makechain-state/src/proof.rs`](https://github.com/officialunofficial/makechain/blob/main/crates/makechain-state/src/proof.rs) (preamble at lines 1-19, wrappers `MmrProofWire`, `RangeProofWire`, `OperationProofWire`, `ExclusionStateProofWire`).

The wrappers must mirror Commonware's struct layout exactly — every time Commonware adds a field to e.g. `RangeProof` (the recent additions of `pre_prefix_acc`, `unfolded_prefix_peaks`, `partial_chunk_digest`, `ops_root` are a great example), the downstream wire format silently goes out of sync until the consumer notices. Native Codec impls would eliminate this hazard.

For `ExclusionProof` specifically, the consumer can't even use the upstream `Db::verify_exclusion_proof` directly because the type is parameterized over `ValueEncoding` (e.g. `VariableEncoding<V>`) and so the consumer has to reconstruct the enum on its own side from a wire wrapper. A direct Codec impl would let consumers call upstream verification end-to-end.

## Proposal

Add the following impls (sketched against current field shapes):

```rust
// storage/src/merkle/proof.rs already has Write + Read for Proof<F, D> — reuse.

// storage/src/qmdb/current/proof.rs

impl<F: Family, D: Digest> Write for RangeProof<F, D> {
        self.proof.write(buf);
        self.pre_prefix_acc.write(buf);
        self.unfolded_prefix_peaks.write(buf);
        self.partial_chunk_digest.write(buf);
        self.ops_root.write(buf);
    }
}

impl<F: Family, D: Digest> EncodeSize for RangeProof<F, D> { /* ... */ }

impl<F: Family, D: Digest> Read for RangeProof<F, D> {
    type Cfg = usize; // max digests, matching merkle::Proof
    fn read_cfg(buf: &mut impl Buf, max_digests: &Self::Cfg) -> Result<Self, codec::Error> { /* ... */ }
}

impl<F: Family, D: Digest, const N: usize> Write for OperationProof<F, D, N> {
        self.loc.write(buf);
        buf.put_slice(&self.chunk);
        self.range_proof.write(buf);
    }
}
// ... matching EncodeSize / Read

impl<F: Family, K: Key + Write, D: Digest, const N: usize> Write for KeyValueProof<F, K, D, N> { ... }

impl<F: Graftable, K: Key, V: ValueEncoding, D: Digest, const N: usize> Write for ExclusionProof<F, K, V, D, N>
where
    Update<K, V>: Write,
    V::Value: Write,
{ ... }
```

A few notes:

1. **`Read::Cfg`** for the composite types should likely thread through `(max_digests, max_value_bytes)` (or the existing `usize` from `merkle::Proof::Cfg`) to bound `Vec<D>` and value sizes. Bounded reads matter here because these proofs traverse the network.
2. **`ExclusionProof`** needs `V: ValueEncoding`-aware bounds. Because `VariableEncoding<V>` is already `pub` (correctly!), this is expressible — earlier we believed it was `pub(crate)`-gated; that's no longer the case. This makes the impl tractable.
3. **`#[derive(Encode, Decode)]`** could in principle work via `commonware-codec`'s Codec macros, but the generic bounds here likely require manual impls (similar to how `merkle::Proof<F, D>` is hand-written).
4. **Conformance tests**: please add `CodecConformance<RangeProof<...>>` and friends to `storage/src/qmdb/current/proof.rs`'s test module, mirroring the existing `CodecConformance<OrderedOperation<...>>` patterns in `qmdb/any/operation/mod.rs`.

**Prior art / precedent:** [PR #1127](https://github.com/commonwarexyz/monorepo/pull/1127) (merged 2025-06-18, *[storage] use `commonware-codec` to serialize/deserialize proofs*) established this exact pattern by replacing custom `serialize`/`deserialize` on the inner MMR `Proof<F, D>` with `Read`/`Write` impls. That PR's `Read::Cfg = usize` (max digests) is the shape I'd suggest mirroring here, extended to a tuple `(max_digests, max_value_bytes)` for the QMDB composites that wrap variable-length values.

The QMDB proof composites (`RangeProof`, `OperationProof`, `KeyValueProof`, `ExclusionProof`) were introduced later (e.g. [PR #2819](https://github.com/commonwarexyz/monorepo/pull/2819) variable-length QMDB, [PR #1799](https://github.com/commonwarexyz/monorepo/pull/1799) ordered exclusion proofs) and the codec impls were never extended forward to them — leaving downstream consumers like makechain reimplementing the wire format.

roberto-bayardo:
Resolves: https://github.com/commonwarexyz/monorepo/issues/3705

### DIFF
--- a/storage/conformance.toml
+++ b/storage/conformance.toml
@@ -262,6 +262,50 @@ hash = "311e78baa8209313a49f2d788d96aa880ba4b3c1c5c88f3151b665f25160fad8"
 n_cases = 200
 hash = "d6b0f89534be60d85f5754c16574536d84899fce930dfc182ef973475d3078e4"
 
+["commonware_storage::qmdb::current::ordered::tests::conformance::CodecConformance<ExclusionProof<mmb::Family,U64,FixedEncoding<U64>\n,Sha256Digest,32>>"]
+n_cases = 65536
+hash = "3be28a20c912b32bc94dacedbee177f371dedae0374788062283a09198bbe8a1"
+
+["commonware_storage::qmdb::current::ordered::tests::conformance::CodecConformance<ExclusionProof<mmb::Family,U64,VariableEncoding<Vec\n<u8>>,Sha256Digest,32>>"]
+n_cases = 65536
+hash = "dec32d46d821f96fcf6795e79316dd120acbfa80c579aede1e3e1b12fa5d529b"
+
+["commonware_storage::qmdb::current::ordered::tests::conformance::CodecConformance<ExclusionProof<mmr::Family,U64,FixedEncoding<U64>\n,Sha256Digest,32>>"]
+n_cases = 65536
+hash = "ab38295608a9725a7e3ae3066d0908271f049f373beee1a266f823b8e4aa0e19"
+
+["commonware_storage::qmdb::current::ordered::tests::conformance::CodecConformance<ExclusionProof<mmr::Family,U64,VariableEncoding<Vec\n<u8>>,Sha256Digest,32>>"]
+n_cases = 65536
+hash = "8127a8b087d13a867eb02343e90b40dcc7ac1abaf1f49f205d932a7772944f94"
+
+["commonware_storage::qmdb::current::ordered::tests::conformance::CodecConformance<KeyValueProof<mmb::Family,U64,Sha256Digest,32>>"]
+n_cases = 65536
+hash = "bb7c2e0914c06a3cd0b64540c022cadc32ee1eafcf319a9d4fe744edf0e43428"
+
+["commonware_storage::qmdb::current::ordered::tests::conformance::CodecConformance<KeyValueProof<mmr::Family,U64,Sha256Digest,32>>"]
+n_cases = 65536
+hash = "a9d1bd40b8bf177389e77a4d33c9561ba5e96b12220de506bd4ce0817311d09a"
+
+["commonware_storage::qmdb::current::proof::tests::conformance::CodecConformance<OperationProof<mmb::Family,Sha256Digest,32>>"]
+n_cases = 65536
+hash = "755bf9cd2a1d3e27c44472094e0b9f509b0c5f977dd63fd8df334adf860fd9b2"
+
+["commonware_storage::qmdb::current::proof::tests::conformance::CodecConformance<OperationProof<mmr::Family,Sha256Digest,32>>"]
+n_cases = 65536
+hash = "d1035bc4913d84385ed906f0a44a28be3dbd0a91a3cdbf475f8a8427bd7276c8"
+
+["commonware_storage::qmdb::current::proof::tests::conformance::CodecConformance<OpsRootWitness<Sha256Digest>>"]
+n_cases = 65536
+hash = "10f6ff08377d0fa64e3802d5b7e14b314699a16fbd1a080d4487288a3c6048e4"
+
+["commonware_storage::qmdb::current::proof::tests::conformance::CodecConformance<RangeProof<mmb::Family,Sha256Digest>>"]
+n_cases = 65536
+hash = "b5e8e12fe4cefcc5cd0ba81e0bf1dabd816b6af63a2c939ad2e38c3bf2eea508"
+
+["commonware_storage::qmdb::current::proof::tests::conformance::CodecConformance<RangeProof<mmr::Family,Sha256Digest>>"]
+n_cases = 65536
+hash = "1385f5d8ad72ee20548a3154223208ef83b83ce34769871957c631caaf2d56f1"
+
 ["commonware_storage::qmdb::immutable::operation::fixed::tests::conformance::CodecConformance<FixedOp>"]
 n_cases = 65536
 hash = "1737c49c112fc9dfdc4dde3626d8ab08a9df8353b9a702e488d3a1fd8e9428dc"
--- a/storage/src/qmdb/current/ordered/mod.rs
+++ b/storage/src/qmdb/current/ordered/mod.rs
@@ -97,7 +97,7 @@ where
     D: Digest,
     Update<K, V>: Read,
 {
-    /// `(max_digests, update_cfg, value_cfg)`: per-field digest cap for the embedded operation
+    /// `(max_digests, update_cfg, value_cfg)`: total digest cap for the embedded operation
     /// proof, the read configuration for [Update], and the read configuration for the value type.
     type Cfg = (usize, <Update<K, V> as Read>::Cfg, <V::Value as Read>::Cfg);
 
@@ -956,6 +956,12 @@ pub mod tests {
         }
     }
 
+    fn op_proof_digest_count(proof: &OperationProof<mmb::Family, Digest, 32>) -> usize {
+        proof.range_proof.proof.digests.len()
+            + proof.range_proof.unfolded_prefix_peaks.len()
+            + proof.range_proof.unfolded_suffix_peaks.len()
+    }
+
     type CodecExclusionProof =
         ExclusionProof<mmb::Family, Digest, FixedEncoding<Digest>, Digest, 32>;
     type CodecKeyValueProof = db::KeyValueProof<mmb::Family, Digest, Digest, 32>;
@@ -972,6 +978,21 @@ pub mod tests {
         assert_eq!(encoded.len(), proof.encode_size());
         let decoded = CodecKeyValueProof::decode_cfg(encoded, &(MAX_DIGESTS, ())).unwrap();
         assert_eq!(decoded, proof);
+    }
+
+    #[test]
+    fn test_key_value_proof_codec_enforces_total_digest_budget() {
+        let proof = CodecKeyValueProof {
+            proof: sample_op_proof(),
+            next_key: Sha256::hash(b"next-key"),
+        };
+        let total_digests = op_proof_digest_count(&proof.proof);
+
+        let encoded = proof.encode();
+        let decoded =
+            CodecKeyValueProof::decode_cfg(encoded.clone(), &(total_digests, ())).unwrap();
+        assert_eq!(decoded, proof);
+        assert!(CodecKeyValueProof::decode_cfg(encoded, &(total_digests - 1, ())).is_err());
     }
 
     #[test]
@@ -994,6 +1015,37 @@ pub mod tests {
             assert_eq!(encoded.len(), proof.encode_size());
             let decoded = CodecExclusionProof::decode_cfg(encoded, &(MAX_DIGESTS, (), ())).unwrap();
             assert_eq!(decoded, proof);
+        }
+    }
+
+    #[test]
+    fn test_exclusion_proof_codec_enforces_total_digest_budget() {
+        let cases = [
+            CodecExclusionProof::KeyValue(
+                sample_op_proof(),
+                Update {
+                    key: Sha256::hash(b"key"),
+                    value: Sha256::hash(b"value"),
+                    next_key: Sha256::hash(b"next-key"),
+                },
+            ),
+            CodecExclusionProof::Commit(sample_op_proof(), Some(Sha256::hash(b"metadata"))),
+            CodecExclusionProof::Commit(sample_op_proof(), None),
+        ];
+
+        for proof in cases {
+            let total_digests = match &proof {
+                CodecExclusionProof::KeyValue(op_proof, _) => op_proof_digest_count(op_proof),
+                CodecExclusionProof::Commit(op_proof, _) => op_proof_digest_count(op_proof),
+            };
+
+            let encoded = proof.encode();
+            let decoded =
+                CodecExclusionProof::decode_cfg(encoded.clone(), &(total_digests, (), ())).unwrap();
+            assert_eq!(decoded, proof);
+            assert!(
+                CodecExclusionProof::decode_cfg(encoded, &(total_digests - 1, (), ())).is_err()
+            );
         }
     }
 
--- a/storage/src/qmdb/current/ordered/mod.rs
+++ b/storage/src/qmdb/current/ordered/mod.rs
@@ -922,8 +922,8 @@ pub mod tests {
                 inactive_peaks: 0,
                 digests: vec![Sha256::hash(b"sib")],
             },
-            pre_prefix_acc: Some(Sha256::hash(b"pre")),
             unfolded_prefix_peaks: vec![Sha256::hash(b"peak")],
+            unfolded_suffix_peaks: vec![Sha256::hash(b"suf")],
             partial_chunk_digest: None,
             ops_root: Sha256::hash(b"ops"),
         };
--- a/storage/src/qmdb/current/ordered/mod.rs
+++ b/storage/src/qmdb/current/ordered/mod.rs
@@ -54,6 +54,7 @@ where
     F: Graftable,
     K: Key,
     V: ValueEncoding,
+    Update<K, V>: Write,
     D: Digest,
 {
     fn write(&self, buf: &mut impl BufMut) {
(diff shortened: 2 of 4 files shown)


storage/src/qmdb/current/ordered/mod.rs
```diff
@@ -44,6 +46,111 @@ pub enum ExclusionProof<F: Graftable, K: Key, V: ValueEncoding, D: Digest, const
...
+    K: Key,
+    V: ValueEncoding,
+    D: Digest,
+{
+    fn write(&self, buf: &mut impl BufMut) {
+        match self {
+            Self::KeyValue(op_proof, update) => {
+                EXCLUSION_TAG_KEY_VALUE.write(buf);
+                op_proof.write(buf);
+                update.key.write(buf);
```
patrick-ogrady: Is there a reason we don’t define write on update?

storage/src/qmdb/current/ordered/mod.rs
```diff
@@ -44,6 +46,111 @@ pub enum ExclusionProof<F: Graftable, K: Key, V: ValueEncoding, D: Digest, const
...
+                update.value.write(buf);
+                update.next_key.write(buf);
+            }
+            Self::Commit(op_proof, value) => {
+                EXCLUSION_TAG_COMMIT.write(buf);
+                op_proof.write(buf);
+                value.is_some().write(buf);
```
patrick-ogrady: Option write is already supported by Codec!

storage/src/qmdb/current/ordered/mod.rs
```diff
@@ -44,6 +46,111 @@ pub enum ExclusionProof<F: Graftable, K: Key, V: ValueEncoding, D: Digest, const
...
+                    Update {
+                        key,
+                        value,
+                        next_key,
+                    },
+                ))
+            }
+            EXCLUSION_TAG_COMMIT => {
+                let op_proof = OperationProof::<F, D, N>::read_cfg(buf, max_digests)?;
+                let value = if bool::read(buf)? {
```
patrick-ogrady: see comment about option support

storage/src/qmdb/current/proof.rs
```diff
@@ -627,6 +629,70 @@ impl<F: Graftable, D: Digest> RangeProof<F, D> {
...
+impl<F: Family, D: Digest> Write for RangeProof<F, D> {
+    fn write(&self, buf: &mut impl BufMut) {
+        self.proof.write(buf);
+        self.pre_prefix_acc.is_some().write(buf);
+        if let Some(digest) = &self.pre_prefix_acc {
+            digest.write(buf);
+        }
+        self.unfolded_prefix_peaks.write(buf);
+        self.partial_chunk_digest.is_some().write(buf);
+        if let Some(digest) = &self.partial_chunk_digest {
```
patrick-ogrady: see comment about option support

patrick-ogrady (changes requested):
we should better leverage codec

patrick-ogrady:
Another finding:

```
path: storage/src/qmdb/current/proof.rs
line: 748
severity: medium
---
`OperationProof` is now decodable from untrusted bytes, but `verify` indexes the decoded bitmap chunk before enforcing the chunk-size invariant that `current::init` enforces for real DBs. A decoded `OperationProof<_, _, 0>` reaches `verify` and panics in `BitMap::<0>::get_bit_from_chunk` instead of returning `false`. Please reject `N == 0` before bitmap indexing and before the delegated `RangeProof::verify` path can divide by `CHUNK_SIZE_BITS`.

Test command: `just test -p commonware-storage test_decoded_operation_proof_zero_sized_chunk_does_not_panic`

Observed result: the test fails with a panic at `utils/src/bitmap/mod.rs:535`: `attempt to calculate the remainder with a divisor of zero`.

```patch
diff --git a/storage/src/qmdb/current/proof.rs b/storage/src/qmdb/current/proof.rs
index d49ea1d80..ce5ac8b73 100644
--- a/storage/src/qmdb/current/proof.rs
+++ b/storage/src/qmdb/current/proof.rs
@@ -913,6 +913,40 @@ mod tests {
         assert_eq!(decoded, proof);
     }
 
+    #[test]
+    fn test_decoded_operation_proof_zero_sized_chunk_does_not_panic() {
+        type F = mmb::Family;
+        const N: usize = 0;
+        const MAX_DIGESTS: usize = 64;
+
+        let proof = OperationProof::<F, sha256::Digest, N> {
+            loc: mmb::Location::new(0),
+            chunk: [],
+            range_proof: RangeProof {
+                proof: Proof::<F, sha256::Digest> {
+                    leaves: mmb::Location::new(0),
+                    inactive_peaks: 0,
+                    digests: vec![],
+                },
+                pre_prefix_acc: None,
+                unfolded_prefix_peaks: vec![],
+                partial_chunk_digest: None,
+                ops_root: Sha256::hash(b"ops"),
+            },
+        };
+        let decoded =
+            OperationProof::<F, sha256::Digest, N>::decode_cfg(proof.encode(), &MAX_DIGESTS)
+                .unwrap();
+        let root = Sha256::hash(b"root");
+
+        let result = std::panic::catch_unwind(|| {
+            let mut hasher = Sha256::new();
+            decoded.verify(&mut hasher, 7u8, &root)
+        });
+        assert!(result.is_ok(), "verify panicked for zero-sized bitmap chunks");
+        assert!(!result.unwrap(), "zero-sized bitmap chunk proof verified");
+    }
+
     #[test_traced]
     fn test_range_proof_verifies_for_mmb_multi_peak_chunk() {
```
</details>

One unique actionable finding remained after deduplication: decoded `OperationProof` values with zero-sized bitmap chunks can panic during verification instead of failing cleanly.
```

**Comment by patrick-ogrady** [comment]
> Another finding:
> 
> ```
> path: storage/src/qmdb/current/proof.rs
> line: 748
> severity: medium
> ---
> `OperationProof` is now decodable from untrusted bytes, but `verify` indexes the decoded bitmap chunk before enforcing the chunk-size invariant that `current::init` enforces for real DBs. A decoded `OperationProof<_, _, 0>` reaches `verify` and panics in `BitMap::<0>::get_bit_from_chunk` instead of returning `false`. Please reject `N == 0` before bitmap indexing and before the delegated `RangeProof::verify` path can divide by `CHUNK_SIZE_BITS`.
> 
> Test command: `just test -p commonware-storage test_decoded_operation_proof_zero_sized_chunk_does_not_panic`
> 
> Observed result: the test fails with a panic at `utils/src/bitmap/mod.rs:535`: `attempt to calculate the remainder with a divisor of zero`.
> 
> <details><summary>Regression test patch</summary>
> 
> ```patch
> diff --git a/storage/src/qmdb/current/proof.rs b/storage/src/qmdb/current/proof.rs
> index d49ea1d80..ce5ac8b73 100644
> --- a/storage/src/qmdb/current/proof.rs
> +++ b/storage/src/qmdb/current/proof.rs
> @@ -913,6 +913,40 @@ mod tests {
>          assert_eq!(decoded, proof);
>      }
>  
> +    #[test]
> +    fn test_decoded_operation_proof_zero_sized_chunk_does_not_panic() {
> +        type F = mmb::Family;
> +        const N: usize = 0;
> +        const MAX_DIGESTS: usize = 64;
> +
> +        let proof = OperationProof::<F, sha256::Digest, N> {
> +            loc: mmb::Location::new(0),
> +            chunk: [],
> +            range_proof: RangeProof {
> +                proof: Proof::<F, sha256::Digest> {
> +                    leaves: mmb::Location::new(0),
> +                    inactive_peaks: 0,
> +                    digests: vec![],
> +                },
> +                pre_prefix_acc: None,
> +                unfolded_prefix_peaks: vec![],
> +                partial_chunk_digest: None,
> +                ops_root: Sha256::hash(b"ops"),
> +            },
> +        };
> +        let decoded =
> +            OperationProof::<F, sha256::Digest, N>::decode_cfg(proof.encode(), &MAX_DIGESTS)
> +                .unwrap();
> +        let root = Sha256::hash(b"root");
> +
> +        let result = std::panic::catch_unwind(|| {
> +            let mut hasher = Sha256::new();
> +            decoded.verify(&mut hasher, 7u8, &root)
> +        });
> +        assert!(result.is_ok(), "verify panicked for zero-sized bitmap chunks");
> +        assert!(!result.unwrap(), "zero-sized bitmap chunk proof verified");
> +    }
> +
>      #[test_traced]
>      fn test_range_proof_verifies_for_mmb_multi_peak_chunk() {
>          let executor = deterministic::Runner::default();
> ```
> 
> One unique actionable finding remained after deduplication: decoded `OperationProof` values with zero-sized bitmap chunks can panic during verification instead of failing cleanly.

Ah, this is dumb. N is controlled.

File: storage/src/qmdb/current/ordered/mod.rs [hunk 5]
```diff
@@ -832,3 +917,104 @@ pub mod tests {
         });
     }
 }
+
+#[cfg(test)]
+mod codec_tests {
```
patrick-ogrady: I think this can just be part of tests?

storage/src/qmdb/current/ordered/mod.rs
```diff
@@ -44,6 +46,89 @@ pub enum ExclusionProof<F: Graftable, K: Key, V: ValueEncoding, D: Digest, const
...
+                let op_proof = OperationProof::<F, D, N>::read_cfg(buf, max_digests)?;
+                let update = Update::<K, V>::read_cfg(buf, update_cfg)?;
+                Ok(Self::KeyValue(op_proof, update))
+            }
+            EXCLUSION_TAG_COMMIT => {
+                let op_proof = OperationProof::<F, D, N>::read_cfg(buf, max_digests)?;
+                let value = Option::<V::Value>::read_cfg(buf, value_cfg)?;
+                Ok(Self::Commit(op_proof, value))
+            }
+            _ => Err(commonware_codec::Error::Invalid(
```
patrick-ogrady: I think there is an "InvalidEnum"

storage/src/qmdb/current/ordered/mod.rs
roberto-bayardo: AI slop, fixed.

storage/src/qmdb/current/ordered/mod.rs
roberto-bayardo: fixed

storage/src/qmdb/current/ordered/mod.rs
roberto-bayardo: fixed

storage/src/qmdb/current/proof.rs
roberto-bayardo: fixed

storage/src/qmdb/current/ordered/mod.rs
roberto-bayardo: done

storage/src/qmdb/current/ordered/mod.rs
roberto-bayardo: Yes done

patrick-ogrady:
We should add conformance tests for these new proof types.

patrick-ogrady:
`max_digests` is currently enforced three times, not once: `Proof::read_cfg` caps `proof.digests`, then `RangeProof::read_cfg` separately allows up to `max_digests` more entries in both `unfolded_prefix_peaks` and `unfolded_suffix_peaks`. Because `OperationProof::read_cfg`, `KeyValueProof::read_cfg`, and `ExclusionProof::read_cfg` all forward the same cfg unchanged, every new decoder added in this PR can accept a well-formed payload whose total digest allocation is roughly `3 * max_digests`. That defeats the caller's allocation budget and reintroduces the DoS vector this cfg is supposed to prevent. Please enforce a single remaining digest budget across all three vectors, or change the cfg shape to explicit per-field budgets and document that contract.

roberto-bayardo:
> `max_digests` is currently enforced three times, not once: `Proof::read_cfg` caps `proof.digests`, then `RangeProof::read_cfg` separately allows up to `max_digests` more entries in both `unfolded_prefix_peaks` and `unfolded_suffix_peaks`. Because `OperationProof::read_cfg`, `KeyValueProof::read_cfg`, and `ExclusionProof::read_cfg` all forward the same cfg unchanged, every new decoder added in this PR can accept a well-formed payload whose total digest allocation is roughly `3 * max_digests`. That defeats the caller's allocation budget and reintroduces the DoS vector this cfg is supposed to prevent. Please enforce a single remaining digest budget across all three vectors, or change the cfg shape to explicit per-field budgets and document that contract.

This is fine since max-digests is generally around 64 -- hard to DoS with that few digests even if it is *N.  I can document better.

roberto-bayardo:
> We should add conformance tests for these new proof types.

done

patrick-ogrady:
> This is fine since max-digests is generally around 64 -- hard to DoS with that few digests even if it is *N. I can document better.

What do we do in other proof decode functions?

roberto-bayardo:
> > This is fine since max-digests is generally around 64 -- hard to DoS with that few digests even if it is *N. I can document better.
> 
> What do we do in other proof decode functions?

The other proof decoder only had one field of digests.

storage/conformance.toml
patrick-ogrady, line 1: ❤️

storage/src/qmdb/current/proof.rs
```diff
@@ -940,19 +940,20 @@ impl<F: Family, D: Digest> EncodeSize for RangeProof<F, D> {
...
         max_digests: &Self::Cfg,
     ) -> Result<Self, commonware_codec::Error> {
         let proof = Proof::<F, D>::read_cfg(buf, max_digests)?;
-        let unfolded_prefix_peaks = Vec::<D>::read_range(buf, ..=*max_digests)?;
-        let unfolded_suffix_peaks = Vec::<D>::read_range(buf, ..=*max_digests)?;
+        let remaining = max_digests.checked_sub(proof.digests.len()).ok_or(
+            commonware_codec::Error::Invalid("RangeProof", "digest budget exceeded"),
+        )?;
+        let unfolded_prefix_peaks = Vec::<D>::read_range(buf, ..=remaining)?;
+        let remaining = remaining - unfolded_prefix_peaks.len();
```
patrick-ogrady, line 953: it feels weird to do checked_sub above but - here?

patrick-ogrady: above, I suppose we should know it works given we provide max_digests?

roberto-bayardo: changed to unchecked_sub above (yes should be ok due to max_digests).

patrick-ogrady (approved):

patrick-ogrady:
Should fully unblock exoware!

PR #3721, opened 2026-05-05
[Storage] Add contiguous journal and QMDB metrics

danlaine:
## Summary

Adds runtime metrics for storage components and a few small adjacent fixes:

- Contiguous journals expose state, read, write, sync, cache, and commit metrics.
- QMDB variants expose operation-log state, read, write, sync, prune, and Current-layer metrics.
- Counters and duration histograms record calls once an operation starts. Explicit no-op batch reads such as empty `read_many`/`get_many` return before recording. Counters that report items processed (`items_read`, `operations_applied`) only advance on success.
- Variable contiguous journals gain an explicit `Reader::read_many` override that batches the offsets lookup instead of relying on the trait default loop.
- `Reader::read_many` and public keyless location batch reads now document the strictly-increasing input contract required by the optimized journal implementations.
- `keyless::Keyless::init_from_journal` now takes the runtime context as an additional parameter so it can register metrics; the two keyless callers (`fixed.rs`, `variable.rs`) pass the context that already wraps the journal label.

No wire or storage formats change.

## Metrics Added

Names below are the suffixes registered under each component's runtime label. For example, a journal labeled `fixed_metrics` exports `fixed_metrics_size`. Counter names get an automatic `_total` suffix when emitted; histogram names get `_count`/`_sum`/`_bucket`.

### Contiguous Journal Metrics

Shared by fixed-size and variable-size contiguous journals:

- `size`: logical end position.
- `pruning_boundary`: oldest readable item position.
- `retained`: number of readable items retained.
- `tail_items`: items in the section containing the newest retained item.
- `append_calls_total`: single-item append calls.
- `append_many_calls_total`: append-many calls.
- `read_calls_total`: single-item async read calls.
- `read_many_calls_total`: non-empty batch read calls.
- `try_read_sync_hits_total`: synchronous probes (`try_read_sync`) that returned `Some`.
- `items_read_total`: items returned by successful read paths (`read`, `read_many`, `try_read_sync`).
- `sync_calls_total`: full-sync calls.
- `append_duration`: duration of single-item append calls.
- `append_many_duration`: duration of append-many calls.
- `read_duration`: duration of single-item read calls.
- `read_many_duration`: duration of non-empty batch read calls.
- `sync_duration`: duration of full-sync calls.

Fixed-size contiguous journals also add:

- `cache_hits_total`: fixed items read synchronously without async storage fallback. This includes items satisfied during `read_many` plus successful `try_read_sync` probes. Single-item async `read` is not counted.
- `cache_misses_total`: fixed items not satisfied synchronously. This includes `read_many` misses that require async storage fallback, plus `try_read_sync` probes that returned `None`, including pruned or out-of-range probes. Single-item async `read` is not counted.

Variable-size contiguous journals also add:

- `commit_calls_total`: commit calls (durable persist that does not fully sync all indexes).
- `commit_duration`: duration of commit calls.

Variable-size journals are backed by a data journal and an internal fixed-size offsets journal, which is registered under an `offsets` child label. All `*_offsets_*` metrics on a variable journal come from the internal fixed-size offsets journal. They include internal offset operations performed by variable-journal read, write, sync, prune, replay, rewind, and recovery paths. They are not the user-facing call counts at the variable layer.

### QMDB Operation-Log Metrics

Used by `any`, `immutable`, and `keyless` QMDB layers where applicable:

- `size`: logical operation end.
- `pruning_boundary`: oldest retained operation location.
- `retained`: number of retained operations.
- `inactivity_floor`: application-declared pruning floor location.
- `last_commit`: most recent commit operation location.
- `apply_batch_calls_total`: apply-batch calls.
- `operations_applied_total`: operations written by successful batch applications.
- `commit_calls_total`: durable commit calls.
- `sync_calls_total`: full-sync calls.
- `prune_calls_total`: prune calls.
- `apply_batch_duration`: duration of apply-batch calls.
- `commit_duration`: duration of commit calls.
- `sync_duration`: duration of sync calls.
- `prune_duration`: duration of prune calls.

Key-based QMDB reads (`any`, `immutable`) add:

- `get_calls_total`: single-key get calls.
- `get_many_calls_total`: non-empty get-many calls.
- `keys_requested_total`: keys requested by attempted reads, whether or not found.
- `get_duration`: duration of get calls.
- `get_many_duration`: duration of non-empty get-many calls.

Location-based QMDB reads (`keyless`) add:

- `get_calls_total`: single-location get calls.
- `get_many_calls_total`: non-empty get-many calls. Input locations must be strictly increasing.
- `locations_requested_total`: locations requested by attempted reads, whether or not found.
- `get_duration`: duration of get calls.
- `get_many_duration`: duration of non-empty get-many calls.

### Current QMDB Metrics

These count Current-layer calls only. The underlying `any` counters (`any_apply_batch_calls_total`, `any_commit_calls_total`, etc.) are bumped when Current delegates to the wrapped `any::Db`. Current's `prune` uses the internal `any::Db::prune_log` helper, so `any_prune_calls_total` will stay at `0` in Current-only deployments even as `current_prune_calls_total` increases.

- `pruned_chunks`: number of pruned bitmap chunks.
- `sync_boundary`: most recent safe sync boundary location.
- `apply_batch_calls_total`: Current-layer apply-batch calls.
- `sync_calls_total`: Current-layer sync calls.
- `prune_calls_total`: Current-layer prune calls.
- `apply_batch_duration`: duration of Current-layer apply-batch calls.
- `sync_duration`: duration of Current-layer sync calls.
- `prune_duration`: duration of Current-layer prune calls.

### QMDB Sync Progress Metrics

The sync engine now registers progress gauges under a `sync` child label to avoid collisions with journal metrics:

- `sync_journal_size`: current sync journal size.
- `sync_target_end`: exclusive target range end, equal to journal size when sync completes.


storage/src/metrics.rs
```diff
@@ -0,0 +1,84 @@
+//! Shared metric helpers.
```
roberto-bayardo: This stuff doesn't seem particularly storage specific, would it make sense to just add it to runtime/src/telemetry/metrics?

danlaine: Done

roberto-bayardo:
lgtm except for the stack overflow test issues.

danlaine:
> lgtm except for the stack overflow test issues.

Should (hopefully!) be fixed now 🤞

roberto-bayardo (approved):

Issue #3718, opened 2026-05-05
[consensus/marshal] Expand Digest Fetch Handling


patrick-ogrady:
We only fetch by digest if we know the block is in the canonical chain (because we don't track cancellation). This leads to cases near the tip where a missing block above the last finalization is never fetched/backfilled (this isn't a problem for the immediate parent we are fetching because we know its round info and can fetch a `Notarized` but it is for its parent). 

To cause a stall, nodes would need to have selective knowledge of the tip (building on forks the other side can't resolve before timeout).

The proper fix here is to track digest fetches by height and clear any unresolved ones once we process past some height. We need to be a bit careful here because just blindly stuffing the resolver with digests could lead to starvation of recently added digests. Maybe there is some way to extend the resolver to add priority or add "cancelable" requests that are no longer requested once a subscription is dropped.

PR #3722, opened 2026-05-05
[consensus/marshal] Expand Digest Fetching

patrick-ogrady:
Related: #3718


consensus/src/marshal/core/actor.rs
```diff
@@ -859,13 +867,25 @@ where
         }
     }
 
+    /// Enqueue a resolver fetch for a block commitment.
+    async fn enqueue_block_fetch(
```
patrick-ogrady: remove this thin wrapper

consensus/src/marshal/core/actor.rs
```diff
@@ -904,7 +923,24 @@ where
             // until resolution or pruning (even if the oneshot is canceled).
             debug!(?round, ?digest, "requested block missing");
             resolver
-                .fetch(Request::<V::Commitment>::Notarized { round })
+                .fetch(ResolverKey::request(Request::Notarized { round }))
+                .await;
+        } else if let (Some(height), BlockSubscriptionKey::Commitment(commitment)) = (height, key) {
+            if height <= self.last_processed_height {
```
patrick-ogrady, line 967: May need this to verify block (read parent)?

consensus/src/marshal/core/cache.rs
```diff
@@ -102,6 +102,9 @@ where
     /// and ceiling, the minimum and maximum epochs (inclusive) that may have data.
     metadata: Metadata<R, u8, (Epoch, Epoch)>,
 
+    /// Blocks fetched by digest during ancestry verification, indexed by height.
+    digest_blocks: prunable::Archive<TwoCap, R, <V::Block as Digestible>::Digest, V::StoredBlock>,
```
patrick-ogrady: put in caches

consensus/src/marshal/core/mailbox.rs
```diff
@@ -391,6 +415,15 @@ impl<S: Scheme, V: Variant> BlockProvider for Mailbox<S, V> {
         let subscription = self.subscribe_by_digest(None, digest).await;
         subscription.await.ok().map(V::into_inner)
     }
+
+    async fn fetch_parent(self, block: Self::Block) -> Option<Self::Block> {
```
patrick-ogrady: make it clearer what actually fetches when subscribing.

consensus/src/marshal/standard/inline.rs
```diff
@@ -510,7 +515,10 @@ where
         }
 
         // Otherwise, subscribe to marshal for block availability.
-        let block_rx = self.marshal.subscribe_by_digest(Some(round), digest).await;
+        let block_rx = self
+            .marshal
+            .subscribe_by_commitment(digest, CommitmentRequest::FetchByRound { round })
```
patrick-ogrady: the one big downside of this approach is that we can fetch the same object twice (won't dedup because of hint)

patrick-ogrady: If we refactor resolver to allow a multiple "retain keys" to a single "fetch key", we could minimize duplicate work.

consensus/src/marshal/standard/inline.rs
```diff
@@ -510,7 +515,10 @@ where
         }
```
patrick-ogrady, line 497: For data on the pending tip, we actually want to specify not to fetch the block (and just wait for it).

This would make it possible for us to starve the resolver capacity (because we won't prune that pending request until finalization).

consensus/src/marshal/core/cache.rs
```diff
@@ -60,6 +60,9 @@ where
         <V::Block as Digestible>::Digest,
         Finalization<S, V::Commitment>,
     >,
+    /// Verified blocks stored by height.
+    verified_blocks_by_height:
```
patrick-ogrady: we should be able to get rid of the verified by view (if we can track digest by height we don't need that)

patrick-ogrady: maybe call this "certified_blocks_by_height"?

consensus/src/marshal/core/cache.rs
```diff
@@ -60,6 +60,9 @@ where
...
+        prunable::Archive<TwoCap, R, <V::Block as Digestible>::Digest, V::StoredBlock>,
 }
```
patrick-ogrady, line 66: We should ensure we don't fetch beyond the finalized floor here? if so, we should at least be sure we dedup requests?

consensus/src/marshal/core/cache.rs
```diff
@@ -60,6 +60,9 @@ where
...
```
patrick-ogrady, line 67: I'm not sure we ever need to fetch by notarized anymore?

patrick-ogrady:
Replaced by #3796

PR #3728, opened 2026-05-06
[runtime/iobuf/pool] optimize local cache hot path

andresilva:
This PR improves the buffer pool's steady-state thread cache reuse path. The profile target is the no-touch 1KiB alloc/drop loop with pooling forced on, which isolates allocator overhead from buffer initialization. Before this PR, the hot pooled path was dominated by destructor-aware TLS access and large inlined miss/refill/allocation tails, even when every operation hit the local thread cache.

The implementation keeps the same existing ownership and reuse semantics, but reshapes the code around the hot path. Local TLS pop/push hits stay inline, while refill, spill, global freelist fallback, and new allocation are split into out-of-line helpers. `try_alloc` is also shaped so the class lookup, TLS lookup, and local cache hit can inline into callers without dragging metrics and heap-allocation code along.

This PR also adds a destructor-free TLS fast pointer for the per-thread buffer-pool cache registry. The checked TLS key remains the owner and is still used for first access and teardown-safe fallback, the fast key is only a cached raw pointer used after initialization to avoid the steady-state cost of Rust's destructor-aware `thread_local!::try_with` path. If TLS is unavailable during thread-local destruction, pooled buffers still fall back to the class-global freelist instead of panicking.

Representative local results for the 1KiB no-touch steady-state profile:

| Case | ns/op |
| --- | ---: |
| Direct (before) | 9.54 |
| Pooled (before) | 24.98 |
| Direct (after) | 7.08 |
| Pooled (after) | 11.83 |

This cuts the stable pooled hot path by roughly 53% versus the original baseline. Direct allocation also gets faster because the untracked drop path benefits from the same inlining work. The pool is still slower than direct allocation, but the gap is now only about 5 ns/op instead of about 15 ns/op at the starting point. Follow-up work will close the remaining gap (and then some 😈.)


andresilva:
For posterity, the TLS fast path optimization is the difference between:

```
movzbl %fs:state_offset,%ecx  # load TLS destructor state
cmp    $0x1,%ecx              # initialized?
je     .ready
cmp    $0x2,%ecx              # destroyed?
jne    .register_dtor         # first access registers destructor

.ready:
lea    %fs:owner_offset,%rdi  # owner TLS value
mov    0x8(%rdi),%rax         # load cache vec
```

and

```
mov    %fs:fast_offset,%rdi  # load cached TlsSizeClassCaches*
test   %rdi,%rdi             # null check, strongly predicted after first access
je     .slow_path            # initialize on first access
mov    0x8(%rdi),%rax        # load cache vec
```

runtime/src/iobuf/pool.rs
```diff
@@ -293,6 +294,8 @@ impl BufferPoolConfig {
     ///
     /// # Panics
     ///
+    /// Panics if size-class bounds violate the invariants checked by `validate`.
+    ///
```
patrick-ogrady: link to validate + remove trailing `///`

runtime/src/iobuf/pool.rs
```diff
@@ -362,37 +365,69 @@ impl BufferPoolConfig {
...
-        if self.max_size < self.min_size {
-            return 0;
-        }
-        // Classes are: min_size, min_size*2, min_size*4, ..., max_size
-        (self.max_size.get() / self.min_size.get()).trailing_zeros() as usize + 1
+    const fn num_classes(&self) -> usize {
+        let min_size = self.min_size.get();
+        let max_size = self.max_size.get();
+
+        assert!(
```
patrick-ogrady: if these are already checked, why do we check again here?

patrick-ogrady: just because you can modify after created?

patrick-ogrady: by this logic, we would have "checkers" in all of these functions?

runtime/src/iobuf/pool.rs
```diff
@@ -670,81 +738,139 @@ impl Drop for TlsSizeClassCache {
...
-    // any synchronization. The cost is that vectors can accumulate holes over time
-    // because ids are not recycled.
     thread_local! {
-        static TLS_SIZE_CLASS_CACHES: UnsafeCell<Vec<Option<TlsSizeClassCache>>> =
-            const { UnsafeCell::new(Vec::new()) };
+        // Owns this thread's cache registry and drops it during thread exit.
+        static TLS_SIZE_CLASS_CACHES: UnsafeCell<TlsSizeClassCaches> =
+            const { UnsafeCell::new(TlsSizeClassCaches::new()) };
+
+        // Performance-only pointer to the same registry. This key has no
```
patrick-ogrady, line 791: :fancy:

runtime/src/iobuf/pool.rs
andresilva: `validate` is private while this method is public, rustc complains about this.

runtime/src/iobuf/pool.rs
andresilva: I guess because we could call `num_classes` or `class_index` on a config that hasn't been validated yet.

andresilva: I cleaned it up.

patrick-ogrady (approved):
⏩

PR #3729, opened 2026-05-06
[Storage] Replace `BTreeSet`,  `HashSet`, `HashMap` with `AHashSet`, `AHashMap`

danlaine:
Switch four hash-based sets/maps in `qmdb::any::batch` from std's default (SipHash) to `ahash::AHashMap`/`AHashSet`.

## Why ahash

`ahash::RandomState` seeds itself from OS entropy on process start (already enabled via `commonware-storage`'s `std` feature). Same DoS-resistance property as SipHash — an external attacker cannot predict the per-process secret — but several times faster on 32-byte digest inputs. `ahash` was already a workspace dependency, so no `Cargo.toml` change.

## Benchmarks

`qmdb::chained_growth` (`unordered` chunk=32) with params modified to `NUM_KEYS=100_000`, `UPDATES_PER_BATCH=10_000`, `PREBUILT_CHAIN=10`, `GROW_COUNTS=[10]`, sample_size=30:

| | mean | 95% CI |
|---|---|---|
| origin/main | 190.19 ms | 188.96 – 191.66 ms |
| HEAD | 176.21 ms | 175.04 – 177.68 ms |
| **Change** | **−7.35%** | −8.29% – −6.35% (p<0.001) |

### DIFF
--- a/storage/src/qmdb/any/batch.rs
+++ b/storage/src/qmdb/any/batch.rs
@@ -1208,7 +1208,12 @@ where
         // Depth-1 chains skip the set entirely — a single ancestor can't shadow itself,
         // and each diff's keys are unique by construction.
         let track_shadow = m.ancestors.len() > 1;
-        let mut seen: AHashSet<&K> = AHashSet::new();
+        let seen_cap = if track_shadow {
+            m.ancestors.iter().map(|a| a.diff.len()).sum()
+        } else {
+            0
+        };
+        let mut seen: AHashSet<&K> = AHashSet::with_capacity(seen_cap);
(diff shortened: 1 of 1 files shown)


roberto-bayardo:
Haven't looked deeply but I've been a reluctant to use AHashSet due to DoS risks from carefully crafted inputs, unless it's very clear that's not a possibility.

[Edit] Oh I see this in the PR description: _Same DoS-resistance property as SipHash_

If true disregard!

roberto-bayardo (dismissed):

roberto-bayardo (approved):

clabby (approved):

PR #3730, opened 2026-05-06
`[storage/qmdb]` Optimize `Db::apply_batch` snapshot merge

danlaine:
## Summary

`Db::apply_batch` rebuilds the in-memory snapshot and bitmap when applying a `MerkleizedBatch`. On `main`, every call allocates an `AHashMap<&Key, Option<Location>>` for `base_old_loc` fixups from already-committed ancestors, plus an `AHashSet<&Key>` so the child diff wins over uncommitted ancestor diffs on conflicting keys -- even when the inputs trivially don't need either (no ancestors, single ancestor, etc.). All `diff` and `ancestor_diff` vectors are sorted by key, so the common shallow cases can avoid the hash-table machinery.

This PR replaces the apply-time HashMap+HashSet path with sorted cursor helpers: a priority-ordered merge for `child + uncommitted ancestors`, plus a committed-ancestor resolver for `base_old_loc` fixups. The implementation avoids hash-table allocation in `apply_batch` and uses Vec-backed cursor state over the already-sorted diff slices.

This is a constant-factor optimization for the expected shallow-chain cases. The cursor helpers scan their stream lists linearly, so very deep ancestor chains are depth-sensitive rather than asymptotically equivalent to the previous hash-based implementation.

## What changed

**`storage/src/qmdb/any/batch.rs`** -- Added two private helpers near `apply_diff`:

- **`DiffMerge<'a, K, F, V>`** -- `Iterator` that performs a priority-ordered sorted merge over `&DiffSlice` cursors. On equal keys the lowest-indexed stream (i.e. most recent batch) wins and all tied cursors are advanced. Used to walk `child + uncommitted ancestors` in sorted-key order.
- **`AppliedAncestorResolver<'a, K, F, V>`** -- parallel cursors over committed ancestor diffs in parent-first order; `lookup(key)` advances cursors past `key` and returns the first match. Replaces the `committed_locs` HashMap. Lookups must be in ascending key order.

`apply_batch` also keeps a no-ancestor fast path (`batch.ancestor_diffs.is_empty()`) so the common direct-apply case stays a tight loop over `batch.diff` -- no helper construction.

**`storage/src/qmdb/benches/apply_batch.rs`** -- new Criterion benches with five cases:

- `direct` -- child only, no ancestors.
- `uncommitted_ancestor` -- child + 1 uncommitted ancestor.
- `committed_ancestor` -- child + 1 already-committed ancestor.
- `committed_uncommitted_chain` -- DB -> A (committed) -> B (uncommitted) -> C, applying C. Exercises the path that on `main` paid both HashMap and HashSet.
- `multi_uncommitted` -- child + 2 uncommitted ancestors.


roberto-bayardo:
Any impact on existing benches?

storage/src/qmdb/any/batch.rs
```diff
@@ -324,6 +324,83 @@ fn apply_diff<F: Family, V, I: UnorderedIndex<Value = Location<F>>, const N: usi
...
+            let (s, p) = self.cursors[inner];
+            if s.get(p).is_some_and(|(k, _)| *k == slice[pos].0) {
+                self.cursors[inner].1 += 1;
+            }
+        }
+        Some((&slice[pos].0, &slice[pos].1))
+    }
+}
+
+/// Resolves a key's committed `base_old_loc` by walking parallel cursors over committed
```
roberto-bayardo: I don't fully understand the use of "committed" in this sentence (both places).

roberto-bayardo: I guess it just means "from the base db" (e.g. location from the last time the key was updated in an apply).  Perhaps it's ok then... (sorry thinking out loud)

danlaine: Reworded this and a few other places to use "applied" and "pending" rather than "committed" and "uncommitted" to hopefully make this more clear.

danlaine:
> Any impact on existing benches?

Run locally, the changes don't seem to have a significant impact on existing new benchmarks, but they appear to show a significant improvement in the benchmarks added in this test. (To get this data I locally added the new benchmark to main.)

| Benchmark | Variant | HEAD | `origin/main` + bench | Delta |
|---|---|---:|---:|---:|
| `apply_batch` case=direct | `any::unordered::fixed::mmb` | 1.460 ms | 1.856 ms | -21.3% |
| `apply_batch` case=uncomm_ancestor | `any::unordered::fixed::mmb` | 16.687 ms | 16.270 ms | +2.6% |
| `apply_batch` case=comm_ancestor | `any::unordered::fixed::mmb` | 15.569 ms | 16.204 ms | -3.9% |
| `apply_batch` case=comm_uncomm_chain | `any::unordered::fixed::m.mmb` | 30.271 ms | 35.518 ms | -14.8% |
| `apply_batch` case=multi_uncomm | `any::unordered::fixed::mmb` | 32.110 ms | 32.585 ms | -1.5% |

roberto-bayardo:
A few more uses of "commit" that could be cleaned up:

```
These all use "committed" to mean "applied":                                                                                                                                                           
                                                         
  1. batch.rs:256 — "A batch is committed when ancestor_diff_ends[i] <= db_size." Should be "applied."                                                                                                   
  2. batch.rs:1678-1679 — inline comment "batch.db_size (nothing committed), batch.base_size (all ancestors committed), or any ancestor_diff_ends[i] (partial commit)." All three "committed"s mean
  "applied."                                                                                                                                                                                             
  3. batch.rs:2278-2279 — test comment "Applying the child must combine committed fixups from that ancestor with the still-uncommitted parent diff." "Committed" → "applied."
  4. The test variable names committed / uncommitted (batch.rs:2238, 2249) reinforce the same overload.                                                                                                  
                                                                                                                                                                                                         
  Given apply_batch's own rustdoc explicitly contrasts apply against Db::commit/Db::sync, the inline comments contradict that distinction. Worth a pass to swap "committed" → "applied" in those four    
  spots.                                                                                                                                      
  ```
  
otherwise lgtm

roberto-bayardo (approved):

PR #3732, opened 2026-05-07
[ci] Bump build job timeout

clabby:
## Overview

Bumps the timeout of the build job, as it is now failing on `main`: https://github.com/commonwarexyz/monorepo/actions/runs/25476390300/job/74750688181

### DIFF
--- a/.github/workflows/fast.yml
+++ b/.github/workflows/fast.yml
@@ -353,7 +353,7 @@ jobs:
   build:
     name: "Build (toolchain: ${{ matrix.toolchain }})"
     runs-on: ubuntu-latest
-    timeout-minutes: 30
+    timeout-minutes: 120
     strategy:
       fail-fast: false
       matrix:


roberto-bayardo (approved):

.github/workflows/fast.yml
```diff
@@ -364,7 +353,7 @@ jobs:
   build:
     name: "Build (toolchain: ${{ matrix.toolchain }})"
     runs-on: ubuntu-latest
-    timeout-minutes: 30
+    timeout-minutes: 120
     strategy:
```
andresilva, line 357: Nit: maybe we want something more conservative, like 60? Ideally we don't want the fast pipeline to ever take two hours 😭

andresilva (approved):

.github/workflows/fast.yml
clabby: If the build ever does take this long, we'll certainly notice it slowing down the mq.. I'd prefer to set an unrealistic cap. wdyt?

Also, Satya pays for it 😄

patrick-ogrady (approved):

PR #3733, opened 2026-05-07
[actors] Mailbox Experiment

patrick-ogrady:


utils/src/channel/actor.rs
```diff
@@ -0,0 +1,531 @@
...
+        Self {
+            shared: self.shared.clone(),
+        }
+    }
+}
+
+impl<T: Backpressure> Drop for ActorMailbox<T> {
+    fn drop(&mut self) {
+        let previous = self.shared.senders.fetch_sub(1, Ordering::AcqRel);
+        debug_assert!(previous > 0);
```
patrick-ogrady, line 393: make assert

utils/src/channel/actor.rs
```diff
@@ -0,0 +1,531 @@
...
+    fn is_disconnected(&self) -> bool {
+        self.shared.closed.load(Ordering::Acquire)
+            || self.shared.senders.load(Ordering::Acquire) == 0
+    }
+}
+
+impl<T> Drop for ActorInbox<T> {
+    fn drop(&mut self) {
+        self.shared.closed.store(true, Ordering::Release);
+        while self.shared.inflight.load(Ordering::Acquire) != 0 {
```
patrick-ogrady: I think this is a reasonable teardown approach (to maintain zero-lock) but 🤷

utils/src/channel/actor.rs
```diff
@@ -0,0 +1,531 @@
...
+            std::hint::spin_loop();
+        }
+        while self.shared.ready.pop().is_some() {}
+        self.shared.overflow.clear();
+    }
+}
+
+/// Create an actor mailbox with a bounded ready queue and policy-managed overflow.
+pub fn channel<T: Backpressure>(capacity: usize) -> (ActorMailbox<T>, ActorInbox<T>) {
+    assert!(capacity > 0, "actor mailbox capacity must be greater than zero");
```
patrick-ogrady, line 561: nonzero?

utils/src/channel/actor.rs
```diff
@@ -0,0 +1,531 @@
...
+
+        let feedback = {
+            let mut overflow = Overflow { queue: &mut queue };
+            T::handle(&mut overflow, message)
+        };
+        self.len.store(queue.len(), Ordering::Release);
+        feedback
+    }
+
+    fn refill_ready(&self, ready: &ArrayQueue<T>) {
```
patrick-ogrady: this can leave items in overflow for a while (rather than sliding into array as soon as there is room)...I think that's ok (not sure of a better technique but)

PR #3737, opened 2026-05-08
[Storage] Add `qmdb::any` tracing

danlaine:
## Summary

Adds `tracing` spans around the main `qmdb::any` database and batch operations so deployments can collect timing and high-level operation metadata through normal tracing infrastructure.

The spans are emitted at `INFO` level because they are intended for routine performance monitoring, not only ad hoc debugging.

## What Changed

- Added `#[tracing::instrument(...)]` spans to `qmdb::any` batch operations:
  - `qmdb::any::batch::new`
  - `qmdb::any::batch::merkleize`
  - `qmdb::any::Db::apply_batch`
  - `qmdb::any::Db::to_batch`

- Added `#[tracing::instrument(...)]` spans to `qmdb::any::Db` operations:
  - `qmdb::any::Db::commit`
  - `qmdb::any::Db::sync`
  - `qmdb::any::Db::historical_proof`
  - `qmdb::any::Db::rewind`
  - `qmdb::any::Db::prune`

- Recorded static, already-available fields such as operation sizes, floors, active-key counts, mutation counts, and batch ancestry counts.

- `proof` now reuses the instrumented `historical_proof` path, so current and explicit historical proof requests are reported through the same span.

### DIFF
--- a/storage/src/qmdb/any/batch.rs
+++ b/storage/src/qmdb/any/batch.rs
@@ -985,6 +985,16 @@ where
     Operation<F, update::Unordered<K, V>>: Codec,
 {
     /// Resolve mutations into operations, merkleize, and return an `Arc<MerkleizedBatch>`.
+    #[allow(clippy::type_complexity)]
+    #[tracing::instrument(
+        name = "qmdb::any::batch::merkleize",
+        level = "info",
+        skip_all,
+        fields(
+            variant = "unordered",
+            mutations = self.mutations.len() as u64,
+        ),
+    )]
     pub async fn merkleize<E, C, I, const N: usize>(
         self,
         db: &Db<F, E, C, I, H, update::Unordered<K, V>, N, S>,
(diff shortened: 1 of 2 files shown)


storage/src/qmdb/any/batch.rs
```diff
@@ -918,6 +918,16 @@ where
     Operation<F, update::Unordered<K, V>>: Codec,
 {
     /// Resolve mutations into operations, merkleize, and return an `Arc<MerkleizedBatch>`.
+    #[allow(clippy::type_complexity)]
+    #[tracing::instrument(
+        name = "qmdb::any::batch::merkleize",
+        level = "info",
```
roberto-bayardo, line 991: Shouldn't this be DEBUG at most? I imagine this will be very noisy.

danlaine: I'm not strongly opinionated as to what the right level is. I figured these are reasonable as INFO because you'd want to monitor their performance during normal operation, as opposed to only when debugging. But maybe that's not the case?

That said I did just make batch creation DEBUG since that seems less important.

clabby (dismissed):

roberto-bayardo (approved):

PR #3739, opened 2026-05-08
[actors] Introduce `Mailbox`

patrick-ogrady:
Experiments: #3733 

Add a non-blocking `Mailbox` that affords each actor the option to control how overflowing messages are handled (just spill into more memory, replacing some existing message, combining with some existing message, or dropped).

I haven't done the math but I believe the "replace" strategy actually bounds the memory for most of our actors (because the number of possible messages we want to send at one time is a function of peers or views).

After this is merged, we'll replace nearly all ~async calls with a sync call that returns `Feedback`.

## Results

  | Case | Commonware | Tokio | Commonware vs Tokio |
  |---|---:|---:|---:|
  | 1024 ready sends | 4.46 us | 16.78 us | ~3.8x faster |
  | 1024 ready receives | 3.45 us | 11.10 us | ~3.2x faster |
  | 1024 send+recv round trips | 7.03 us | 15.05 us | ~2.1x faster |
  | 4 producers, 65,536 sends | 5.11 ms | 5.71 ms | ~1.1x faster |


patrick-ogrady:
bugbot run

actor/src/lib.rs
```diff
@@ -0,0 +1,900 @@
...
+
+    /// Replace overflow with a single spilled `message`.
+    pub fn replace_all(&mut self, message: T) -> Backpressure {
+        self.clear();
+        self.spill(message)
+    }
+}
+
+/// Backpressure behavior for actor messages when an inbox is full.
+pub trait MessagePolicy: Sized {
```
patrick-ogrady: This should be called `Policy`

actor/src/lib.rs
```diff
@@ -0,0 +1,900 @@
...
+        }
+        Err(mpsc::error::TryRecvError::Empty)
+    }
+
+    fn pop(&mut self) -> Option<T> {
+        if let Some(message) = self.shared.ready.pop() {
+            return Some(message);
+        }
+
+        self.shared.overflow.refill_ready(&self.shared.ready);
```
patrick-ogrady: A key design decision here is driving the filling of the queue in the receiver rather than in the sender.

actor/src/lib.rs
```diff
@@ -0,0 +1,900 @@
...
+pub struct ReplaceResult<T> {
+    result: Result<(), T>,
+}
+
+/// Policy-managed overflow behind the bounded ready queue.
+pub struct Overflow<'a, T> {
+    queue: &'a mut VecDeque<T>,
+}
+
+impl<T> Overflow<'_, T> {
```
patrick-ogrady: The receiver will determine how to handle overflow. This keeps the sender blissfully unaware of the inner logic required by another actor.

In earlier experiments, I found not doing this led to each actor more or less replicating the state machines of other actors.

actor/src/lib.rs
```diff
@@ -0,0 +1,903 @@
...
+
+    fn is_disconnected(&self) -> bool {
+        self.shared.is_closed() || self.shared.senders.load(Ordering::Acquire) == 0
+    }
+}
+
+impl<T> Drop for Receiver<T> {
+    fn drop(&mut self) {
+        self.shared.close();
+        while self.shared.inflight() != 0 {
```
patrick-ogrady: I don't love these spin locks but shouldn't really be an issue on teardown.

actor/src/lib.rs
```diff
@@ -0,0 +1,895 @@
...
+impl Feedback {
+    /// Returns true if the work was accepted.
+    pub const fn accepted(&self) -> bool {
+        matches!(self, Self::Ok | Self::Backoff)
+    }
+}
+
+/// Result of applying backpressure to a message.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct Backpressure {
```
cronokirby: Why have both `Backpressure` and `Feedback`?

patrick-ogrady: We want to have the inner `MessagePolicy::handle` only return possible states (if it could return feedback directly, you could return closed or ok).

patrick-ogrady: We may be able to make that a bit cleaner (I think this could now be a boolean)

actor/src/lib.rs
```diff
@@ -0,0 +1,895 @@
...
+    }
+
+    fn clear(&self) {
+        let mut queue = lock(&self.queue);
+        queue.clear();
+        self.len.store(0, Ordering::Release);
+    }
+}
+
+struct Shared<T> {
```
cronokirby: I think we could avoid the need for custom atomic logic by leaning on the existing `Notify` struct?

```rust
  struct State<T> {
      ready: VecDeque<T>,
      overflow: VecDeque<T>,
      capacity: usize,
      senders: usize,
      closed: bool,
  }

  struct Shared<T> {
      state: Mutex<State<T>>,
      notify: Notify,
  }

  Then enqueue becomes:

  fn enqueue(&self, message: T) -> Feedback {
      let mut state = self.shared.state.lock();

      if state.closed {
          return Feedback::Closed;
      }

      if state.overflow.is_empty() && state.ready.len() < state.capacity {
          state.ready.push_back(message);
          drop(state);
          self.shared.notify.notify_one();
          return Feedback::Ok;
      }

      let feedback = {
          let mut overflow = Overflow {
              queue: &mut state.overflow,
          };
          T::handle(&mut overflow, message).into()
      };

      drop(state);

      if feedback.accepted() {
          self.shared.notify.notify_one();
      }

      feedback
  }

  recv loops around the same state:

  async fn recv(&mut self) -> Option<T> {
      loop {
          let notified = self.shared.notify.notified();

          {
              let mut state = self.shared.state.lock();

              if let Some(message) = pop_next(&mut state) {
                  return Some(message);
              if state.closed || state.senders == 0 {
                  return None;
      }
  }

  pop_next preserves the current ready/overflow behavior:

  fn pop_next<T>(state: &mut State<T>) -> Option<T> {
      if let Some(message) = state.ready.pop_front() {
          return Some(message);
      }

      while state.ready.len() < state.capacity {
          match state.overflow.pop_front() {
              Some(message) => state.ready.push_back(message),
              None => break,
          }
      }

      state.ready.pop_front()
  }
```

I think that might be worth doing (or something else) unless it's a lot worse, just to make things easier to reason about.

actor/src/lib.rs
```diff
@@ -0,0 +1,895 @@
...
+                        matches!(pending, Self::Update(_))
+                    });
+                    overflow.replace_or_drop(result)
+                }
+                Self::Vote(_) => Backpressure::dropped(),
+            }
+        }
+    }
+
+    #[commonware_macros::test_async]
```
clabby: Import?

actor/src/lib.rs
```diff
@@ -0,0 +1,895 @@
...
+            receiver_waker: AtomicWaker::new(),
+        });
+        (
+            Sender {
+                shared: shared.clone(),
+            },
+            Receiver { shared },
+        )
+    }
+}
```
clabby: I don’t think this needs to be a struct? This would fit better as a free function called `mailbox`.

actor/src/lib.rs
```diff
@@ -0,0 +1,895 @@
...
+
+/// Policy-managed overflow behind the bounded ready queue.
+pub struct Overflow<'a, T> {
+    queue: &'a mut VecDeque<T>,
+}
+
+impl<T> Overflow<'_, T> {
+    /// Spill `message` into overflow after capacity is exceeded.
+    pub fn spill(&mut self, message: T) -> Backpressure {
```
clabby: `Overflow::spill` makes the mailbox capacity a ready-queue-only bound. A stuck receiver plus a producer that keeps sending `Required`/`Buffered`-style messages can grow the `VecDeque` without limit because every overflow send returns `Backoff` after retaining the message. Concrete trigger: capacity 1, enqueue one ready message, then loop `enqueue(Required(i))`; `sender.len()` grows with every call. Existing test `mailbox_capacity_is_soft_limit_for_required_messages` already codifies this by expecting `len == 3` with capacity 1.

If the advertised contract is bounded memory, the mailbox needs an overflow cap or bounded `spill`/`replace_or_spill` behavior. If not, the docs should be very explicit that `Backoff` is advisory and policy authors can still build an unbounded queue.

actor/src/lib.rs
```diff
@@ -0,0 +1,895 @@
...
+        let mut queue = lock(&self.queue);
+        if is_closed() {
+            return Feedback::Closed;
+        }
+
+        let backpressure = {
+            let mut overflow = Overflow { queue: &mut queue };
+            T::handle(&mut overflow, message)
+        };
+        debug_assert!(
```
clabby: `debug_assert!` is the only enforcement that `Feedback::Backoff` actually retained overflow work. Because `Overflow::clear()` is public, a policy can get a real `Backoff` from `spill()`, clear the queue, and return it. In release builds this silently reports an accepted/backed-off message that is no longer in ready or overflow, so a producer that stops on `Backoff` can wait forever for work the receiver will never observe.

Test file path: `actor/src/lib.rs`
Command: `cargo test -p commonware-actor backoff_without_retained_overflow_loses_message --release`
Observed result: fails with `assertion failed: matches!(receiver.try_recv(), Ok(BrokenMessage::BackoffThenClear))`.

```patch
diff --git a/actor/src/lib.rs b/actor/src/lib.rs
index 07100c202..a7e8aaaa8 100644
--- a/actor/src/lib.rs
+++ b/actor/src/lib.rs
@@ -763,6 +763,35 @@ mod tests {
         assert_eq!(receiver.recv().await, None);
     }
 
+    enum BrokenMessage {
+        FillReady,
+        BackoffThenClear,
+    }
+
+    impl MessagePolicy for BrokenMessage {
+        fn handle(overflow: &mut Overflow<'_, Self>, message: Self) -> Backpressure {
+            let backpressure = overflow.spill(message);
+            overflow.clear();
+            backpressure
+        }
+    }
+
+    #[test]
+    fn backoff_without_retained_overflow_loses_message() {
+        let (sender, mut receiver) = Mailbox::new(NZUsize!(1));
+        assert_eq!(sender.enqueue(BrokenMessage::FillReady), Feedback::Ok);
+        assert_eq!(
+            sender.enqueue(BrokenMessage::BackoffThenClear),
+            Feedback::Backoff
+        );
+
+        assert!(matches!(receiver.try_recv(), Ok(BrokenMessage::FillReady)));
+        assert!(matches!(
+            receiver.try_recv(),
+            Ok(BrokenMessage::BackoffThenClear)
+        ));
+    }
+
     #[test]
     fn enqueue_after_receiver_drop_returns_closed() {
         let (sender, receiver) = Mailbox::new(NZUsize!(1));
```
</details>

clabby (commented):
For the crate structure, let’s move the mailbox logic into `mailbox/mod.rs` & add `mailbox/channel.rs` as well? Can restructure this later too if you prefer.

actor/src/lib.rs
cronokirby: IDK, couldn't you get some kind of deadlock with a complicated cycle, where each dropping receivers stalls a tokio thread?

actor/src/lib.rs
patrick-ogrady: I opted to avoid locking over "ready" to get better happy path performance. If we used VecDeque for both, this could be much simpler.

actor/src/lib.rs
patrick-ogrady: Yeah, I agree

actor/src/lib.rs
patrick-ogrady: This is correct and intended 👍

actor/src/lib.rs
patrick-ogrady: I think we should actually go the opposite direction and make this more flexible. We should allow handle to make arbitrary changes to overflow given an incoming message?

actor/src/lib.rs
patrick-ogrady: Yeah, I cleaned this up where the last inflight now drops.

actor/src/benches/mailbox.rs
```diff
@@ -0,0 +1,307 @@
+use commonware_actor::mailbox::{self as actor, Feedback, Policy as MailboxPolicy};
```
clabby: `self as actor` makes the instantiations in the benches a bit odd to read, remove the alias and use `mailbox::new`?

actor/src/lib.rs
cronokirby: Makes sense, but then we shouldn't make `Feedback` public, right?

actor/src/mailbox.rs
```diff
@@ -0,0 +1,859 @@
...
+                Ordering::Acquire,
+            ) {
+                Ok(_) => {
+                    return Ok((
+                        SendPermit {
+                            shared: &self.shared,
+                        },
+                        message,
+                    ));
+                }
```
cronokirby: should we yield here?

docs/index.html
```diff
@@ -78,6 +78,8 @@
     </div>
     <div class="content">
         <h2>Primitives</h2>
+        <h3><a href="https://docs.rs/commonware-actor">actor</a></h3>
+        <p>Coordinate actors without blocking backpressure.</p>
```
clabby: “Coordinate highly concurrent system components.” ?

You’re the name guy, though the existing description feels a bit narrow (this crate will likely have a stream abstraction, etc.)

actor/src/mailbox.rs
```diff
@@ -0,0 +1,792 @@
...
+        self.shared.senders.fetch_add(1, Ordering::Relaxed);
+        Self {
+            shared: self.shared.clone(),
+        }
+    }
+}
+
+impl<T: Policy> Drop for Sender<T> {
+    fn drop(&mut self) {
+        let previous = self.shared.senders.fetch_sub(1, Ordering::AcqRel);
```
patrick-ogrady: cleaning up shared usage rn

actor/src/mailbox.rs
```diff
@@ -0,0 +1,868 @@
...
+///
+/// The mailbox capacity only bounds the ready queue. Overflow is controlled by
+/// policy code and can grow without bound if the policy keeps spilling messages.
+pub struct Overflow<'a, T> {
+    queue: &'a mut VecDeque<T>,
+}
+
+impl<T> Overflow<'_, T> {
+    /// Spill `message` into overflow after ready capacity is exceeded.
+    pub fn spill(&mut self, message: T) -> bool {
```
danlaine: Do we need to return a `bool` here? It's always `true`.

actor/src/mailbox.rs
```diff
@@ -0,0 +1,791 @@
...
+        }
+    }
+
+    /// Drop the message if it could not replace existing overflow work.
+    pub fn replace_or_drop(&mut self, result: Result<(), T>) -> bool {
+        match result {
+            Ok(()) => true,
+            Err(_) => false,
+        }
+    }
```
clabby: This can be removed in favor of just using `result.is_ok()`?

actor/src/lib.rs
clabby: I agree, though let's document this?

actor/src/mailbox.rs
```diff
@@ -0,0 +1,778 @@
...
+    fn apply_policy(&self, message: T, is_closed: impl Fn() -> bool) -> Feedback
+    where
+        T: Policy,
+    {
+        let mut queue = lock(&self.queue);
+        if is_closed() {
+            return Feedback::Closed;
+        }
+
+        let backoff = {
```
clabby: `overflow.is_active()` is only a hint, but once a sender observes it as true this path goes straight to the overflow policy. A receiver can drain the last overflow item before this sender gets the overflow lock, leaving both overflow empty and ready with capacity. For drop/replace-or-drop policies, this returns `Dropped` even though the message could now enter ready.

I think `apply_policy` needs access to `ready` and should retry `ready.push(message)` when the locked overflow queue is empty before invoking `Policy::handle`.

Regression test added in `actor/src/mailbox.rs`. Command:

`cargo test -p commonware-actor --features loom stale_overflow_hint_retries_ready_before_policy -- --nocapture`

Observed result: fails with `left: Dropped`, `right: Ok`.

```rust
#[test]
fn stale_overflow_hint_retries_ready_before_policy() {
    loom::model(|| {
        let (sender, mut receiver) = new::<Message>(NZUsize!(1));
        assert_eq!(sender.enqueue(Message::Drop(0)), Feedback::Ok);
        assert_eq!(sender.enqueue(Message::Spill(1)), Feedback::Backoff);

        let enqueue = thread::spawn(move || sender.enqueue(Message::Drop(2)));

        assert_eq!(receiver.try_recv(), Ok(Message::Drop(0)));
        assert_eq!(receiver.try_recv(), Ok(Message::Spill(1)));

        assert_eq!(enqueue.join().unwrap(), Feedback::Ok);
        assert_eq!(receiver.try_recv(), Ok(Message::Drop(2)));
    });
}
```

actor/src/lib.rs
```diff
@@ -0,0 +1,34 @@
...
+//! Stability varies by primitive. See [README](https://github.com/commonwarexyz/monorepo#stability) for details.
+
+#![doc(
+    html_logo_url = "https://commonware.xyz/imgs/rustdoc_logo.svg",
+    html_favicon_url = "https://commonware.xyz/favicon.ico"
+)]
+
+commonware_macros::stability_scope!(ALPHA {
+    /// Feedback from submitting work to a bounded endpoint.
+    #[derive(Clone, Copy, Debug, PartialEq, Eq)]
```
patrick-ogrady, line 14: This is something we'll use more broadly, so felt it should be top level.

actor/src/mailbox.rs
```diff
@@ -0,0 +1,857 @@
...
+    {
+        let mut queue = lock(&self.queue);
+        if is_closed() {
+            return Feedback::Closed;
+        }
+
+        // `len` is a hint used to keep the empty-overflow send path lock-free. If a
+        // receiver drained overflow after the sender observed that hint, retry ready
+        // before invoking the policy so a message is not dropped while capacity exists.
+        let message = if queue.is_empty() {
```
patrick-ogrady, line 257: We should do the same sort of thing at the end to make sure we aren't closed (before returning to caller) but I think that's ok to race on.

actor/src/mailbox.rs
```diff
@@ -0,0 +1,860 @@
...
+            match ready.push(message) {
+                Ok(()) => {
+                    self.len.store(0, Ordering::Release);
+                    return Feedback::Ok;
+                }
+                Err(message) => message,
+            }
+        } else {
```
clabby, line 265: When `overflow` is non-empty, `enqueue` skips draining overflow into `ready` before applying policy to the incoming message. This can incorrectly drop a message that would have fit if the overflow entry had been promoted first.

Concrete sequence (capacity 2):
1. Enqueue two `Vote` messages → ready is full.
2. Enqueue a `Required` message → overflows (spilled), returns `Backoff`.
3. Drain both `Vote` messages → ready is now empty.
4. Enqueue a new `Vote` message → `is_active()` routes to `apply_policy`; `queue.is_empty()` is `false`; droppable policy returns `Feedback::Dropped`.

Expected: the overflow `Required` message is moved into ready, the new `Vote` fits behind it, and `Feedback::Ok` is returned.

```patch
diff --git a/actor/src/mailbox.rs b/actor/src/mailbox.rs
index 6f6e445da..475e884f6 100644
--- a/actor/src/mailbox.rs
+++ b/actor/src/mailbox.rs
@@ -563,6 +563,21 @@ mod tests {
         assert_eq!(receiver.try_recv(), Ok(Message::Buffered(2)));
     }
 
+    #[test]
+    fn enqueue_uses_ready_capacity_after_partial_drain() {
+        let (sender, mut receiver) = new(NZUsize!(2));
+        assert_eq!(sender.enqueue(Message::Vote(1)), Feedback::Ok);
+        assert_eq!(sender.enqueue(Message::Vote(2)), Feedback::Ok);
+        assert_eq!(sender.enqueue(Message::Required(3)), Feedback::Backoff);
+
+        assert_eq!(receiver.try_recv(), Ok(Message::Vote(1)));
+        assert_eq!(receiver.try_recv(), Ok(Message::Vote(2)));
+
+        assert_eq!(sender.enqueue(Message::Vote(4)), Feedback::Ok);
+        assert_eq!(receiver.try_recv(), Ok(Message::Required(3)));
+        assert_eq!(receiver.try_recv(), Ok(Message::Vote(4)));
+    }
+
     #[test_async]
     async fn full_inbox_retains_unmatched_replaceable_message() {
         let (sender, mut receiver) = new(NZUsize!(1));
```
</details>

actor/src/mailbox.rs
```diff
@@ -0,0 +1,860 @@
...
+        }
+    }
+
+    #[cfg(feature = "loom")]
+    fn new() -> Self {
+        Self {
+            queue: Mutex::new(VecDeque::new()),
+            len: AtomicUsize::new(0),
+        }
+    }
```
cronokirby: We should instead just disable the const fn lint, instead of duplicating the logic

actor/src/mailbox.rs
```diff
@@ -0,0 +1,860 @@
...
+
+    fn backpressure_overflow(&self, message: T) -> Feedback {
+        let feedback = self
+            .state
+            .overflow
+            .apply_policy(&self.state.ready, message, || {
+                self.state.closed.load(Ordering::Acquire)
+            });
+        if matches!(feedback, Feedback::Ok | Feedback::Backoff) {
```
cronokirby: We should use the `accepted` method here

actor/src/mailbox.rs
```diff
@@ -0,0 +1,860 @@
...
+}
+
+impl<T> Overflow<'_, T> {
+    /// Spill `message` into overflow after ready capacity is exceeded.
+    pub fn spill(&mut self, message: T) {
+        self.queue.push_back(message);
+    }
+
+    /// Spill the message into overflow if it could not replace existing overflow work.
+    pub fn replace_or_spill(&mut self, result: Result<(), T>) {
```
cronokirby: I think this method is confusing as named, because it doesn't attempt to `replace` at all. Would probably be better to just do `overflow.replace_last(msg, |...|).unwrap_or_else(|msg| overflow.spill(msg))` versus `overflow.replace_or_spill(overflow.replace_last(msg, |...|))`

actor/src/mailbox.rs
```diff
@@ -0,0 +1,860 @@
...
+    ) {
+        if let Some(pending) = self.find_last_mut(is_match) {
+            merge(pending, message);
+        } else {
+            self.spill(message);
+        }
+    }
+
+    /// Coalesce `message` into preferred overflow, replace fallback overflow, or spill it.
+    pub fn coalesce_or_replace_or_spill(
```
cronokirby: What if you just had one method `merge_or`, which took in a closure taking in the overflow queue and the message, to do whatever you want:


```rust
o.merge_or(msg, is_match, merge, |o, msg| o.spill(msg))
o.merge_or(msg, is_match, merge, |o, msg| { let res = o.replace_last(msg, || ...); o.replace_or_spill(res) ))
```

cronokirby: In fact, this method would also subsume replace_last, because the merge would just prefer the current message.

actor/src/mailbox.rs
```diff
@@ -0,0 +1,860 @@
...
+    /// Remove all overflow messages.
+    ///
+    /// A policy may clear overflow and still return `true` from [`Policy::handle`]
+    /// to request producer backoff after changing mailbox state.
+    pub fn clear(&mut self) {
+        self.queue.clear();
+    }
+
+    /// Replace overflow with a single spilled `message`.
+    pub fn replace_all(&mut self, message: T) {
```
cronokirby: If you have clear and spill return &mut self, then you could just do `o.clear().spill(message)` which imo is clearer than having another method you need to know about.

actor/src/mailbox.rs
```diff
@@ -0,0 +1,875 @@
...
+            message
+        } else {
+            match self.state.ready.push(message) {
+                Ok(()) => {
+                    self.state.waker.wake();
+                    return Feedback::Ok;
+                }
+                Err(message) => message,
+            }
+        };
```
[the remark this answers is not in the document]
patrick-ogrady: I pushed this test (as a standalone) for this exact reason 👍

actor/src/mailbox.rs
```diff
@@ -0,0 +1,921 @@
...
+        self.queue.clear();
+        self
+    }
+}
+
+/// Overflow behavior for actor messages when an inbox is full.
+pub trait Policy: Sized {
+    /// Handle `message` when it cannot enter the bounded ready queue immediately.
+    ///
+    /// Messages already in the ready queue are not provided here; replacement only applies to
```
patrick-ogrady: remove ;

actor/src/mailbox.rs
```diff
@@ -0,0 +1,921 @@
...
+///
+/// Mailbox capacity only bounds the ready queue. Overflow is controlled by
+/// policy code and can grow without bound if the policy keeps spilling messages.
+/// [`Feedback::Backoff`] is advisory and does not imply that the mailbox enforced
+/// a hard bound on retained overflow work.
+pub struct Overflow<'a, T> {
+    queue: &'a mut VecDeque<T>,
+}
+
+impl<T> Overflow<'_, T> {
```
patrick-ogrady: It feels a bit restrictive to bound behavior to these methods. I could imagine removing a match and adding the replacement to the end rather than in-place?

actor/src/mailbox.rs
```diff
@@ -0,0 +1,921 @@
...
+        match message {
+            OrderedMessage::Item(value) | OrderedMessage::Coordinated(value, _) => value,
+        }
+    }
+
+    #[test]
+    fn concurrent_close_and_ready_enqueue_remains_closed() {
+        loom::model(|| {
+            let (sender, receiver) = new::<Message>(NZUsize!(1));
+
```
patrick-ogrady, line 1190: The loom suite does not model the actual `Receiver::poll_recv` wake registration path. The current close/enqueue models only assert that the mailbox remains closed after the race; they would not catch deleting the wake calls in `Sender::drop`, accepted `Sender::enqueue`, or accepted overflow enqueue because no test polls to `Pending` with an observable waker and then verifies the wake.

Please add loom or unit coverage with a counting/custom waker for the assigned races: sender-drop-vs-register, accepted overflow work waking a pending receiver, and receiver-drop-vs-ready-fast-path feedback. The existing async tests directly await after enqueue/drop, so they can complete by being polled again even if no wake was issued.

actor/src/mailbox.rs
```diff
@@ -0,0 +1,921 @@
...
+fn lock<T>(mutex: &Mutex<T>) -> MutexGuard<'_, T> {
+    mutex.lock().unwrap()
+}
+
+#[cfg(not(feature = "loom"))]
+fn lock<T>(mutex: &Mutex<T>) -> MutexGuard<'_, T> {
+    mutex.lock()
+}
+
+#[cfg(feature = "loom")]
```
patrick-ogrady: The loom `ReadyQueue` is a mutex-backed `VecDeque`, so it does not model production `crossbeam_queue::ArrayQueue`’s two-phase push behavior where tail reservation can become visible before the slot value is readable. The current activity-bit argument looks sound, but the highest-risk interleaving around `ready.len() == 0`, concurrent push, and receiver refill is not represented by this model. Consider adding a loom queue model with separate reserve/publish steps, or a focused production stress test for concurrent inactive fast-path push versus overflow mutation/refill.

actor/src/mailbox.rs
```diff
@@ -0,0 +1,984 @@
...
+        f.debug_struct("Sender")
+            .field("capacity", &self.state.ready.capacity())
+            .field("closed", &self.state.closed.load(Ordering::Acquire))
+            .finish()
+    }
+}
+
+impl<T: Policy> Sender<T> {
+    /// Submit a message without waiting for inbox capacity.
+    #[must_use = "handle dropped/closed submissions; required actor messages must not be silently dropped"]
```
patrick-ogrady: improve this

actor/src/mailbox.rs
```diff
@@ -0,0 +1,1245 @@
...
+                Ok(()) => {}
+                Err(message) => {
+                    queue.push_front(message);
+                    break;
+                }
+            }
+        }
+    }
+
+    fn publish_activity(&self, queue: &VecDeque<T>) {
```
patrick-ogrady: define on Mutation

actor/src/mailbox.rs
```diff
@@ -0,0 +1,1245 @@
...
+        T: Policy,
+    {
+        if T::handle(queue, message) {
+            Feedback::Backoff
+        } else {
+            Feedback::Dropped
+        }
+    }
+
+    fn refill_ready_locked(queue: &mut VecDeque<T>, ready: &ReadyQueue<T>) {
```
patrick-ogrady: remove these _locked suffixes

actor/src/mailbox.rs
```diff
@@ -0,0 +1,1353 @@
...
+//
+// The ready fast path only observes this state; it does not reserve it. A sender
+// that sees zero may race with another sender that enters overflow immediately
+// after the load. Those sends are concurrent, and the mailbox does not provide
+// global FIFO ordering across cloned senders.
+//
+// Activity accesses are relaxed because this word does not publish queue
+// contents. The overflow mutex serializes `VecDeque` access, and the ready queue
+// owns its own synchronization. Stale activity observations only decide whether
+// a caller tries a fast path, locks overflow, or waits for a later wake.
```
patrick-ogrady, line 101: This "trick" avoids almost all locking on the happy path. Keeping a non-overflowing channel between threads moving very fast.

patrick-ogrady:
bugbot run

actor/src/mailbox.rs
```diff
@@ -0,0 +1,1362 @@
+//! Bounded ready queues with policy-managed overflow.
+//!
```
patrick-ogrady, line 26: adding an ascii diagram here

actor/src/mailbox.rs
```diff
@@ -0,0 +1,1351 @@
...
+        }
+
+        impl AtomicWaker {
+            fn new() -> Self {
+                Self {
+                    waker: Mutex::new(None),
+                }
+            }
+
+            fn register(&self, waker: &std::task::Waker) {
```
clabby: The loom shim wraps `register` and `wake` in the same mutex, so loom can only see one operation complete before the other. This serializes the race that `sender_drop_racing_waker_registration_wakes_or_disconnects` is meant to exercise: a `wake()` on the sender-drop thread while `register()` is in flight on the receiver thread. The real `futures_util::task::AtomicWaker` uses a three-state machine (`WAITING`/`REGISTERING`/`WAKING`) precisely so that interleaving is observable. The mutex shim collapses those states into one lock, giving loom nothing to explore.

Consider replacing this shim with `loom::future::AtomicWaker` (if available in the loom version in use) or porting the three-state CAS loop from `futures-util` under `#[cfg(loom)]`, so the model checker actually sees the handoff behavior the test name promises.

clabby (dismissed):
This LGTM, one small nit from our robot friends.

clabby (dismissed):

actor/src/mailbox.rs
```diff
@@ -0,0 +1,1337 @@
...
+//! The receiver always pops from the ready queue first. After each ready pop, it
+//! eagerly refills ready from published overflow so senders can return to the
+//! ready fast path without waiting for ready to drain completely. Overflow is
+//! refilled from front to back, but policies decide which overflow messages are
+//! retained and in what order.
+//!
+//! Eager refill favors keeping producer enqueue on the ready fast path over
+//! batching receiver refill work. It may take the overflow lock more often under
+//! sustained overflow, but avoids leaving ready slots empty while overflow
+//! remains populated. Overflow is expected to be exceptional.
```
BrendanChou: nit: I found this paragraph a bit hard to parse

PR #3743, opened 2026-05-08
[Storage] Refactor `qmdb::current` grafting/proofs

danlaine:
# `qmdb::current` proof refactor

## Summary

Replace the MMB-specific multi-peak fold machinery in `qmdb::current` with a "defer-pending" design: at most one chunk is "completed but not yet graftable" at any time, and its bytes ride in the canonical-root trailer instead of being folded into the grafted tree. This restores a clean 1-to-1 invariant -- every graftable chunk has exactly one h=G ancestor -- for **both** MMR and MMB families.

---

## Background: what grafting is

`qmdb::current` authenticates an **ops tree** (MMR or MMB over a log of operations) together with an **activity bitmap** (one bit per op). To avoid two independent Merkle structures, every active bitmap chunk is hashed with the corresponding ops-tree subtree root at the **grafting height** `G = log2(CHUNK_SIZE_BITS)`:

```
                grafted leaf digest
                        |
           hash(chunk_bytes || ops_subtree_root_at_height_G)
                ----------         ----------------------
                bitmap chunk       ops tree subtree root


      Chunk 0       Chunk 1       Chunk 2
      +-----+       +-----+       +-----+        <- bitmap chunks
      |     |       |     |       |     |
      +-----+       +-----+       +-----+
         ^             ^             ^
         |             |             |           <- grafted at h=G
      +--R0--+      +--R1--+      +--R2--+
      |ops   |      |ops   |      |ops   |       <- ops tree h=G nodes
      |0..2^G|      |2^G.. |      | ...  |
      +------+      +------+      +------+
```

This requires a **1-to-1 mapping** between graftable chunks and h=G ops-tree nodes. **MMR** satisfies this trivially (every aligned range of `2^G` leaves has exactly one h=G subtree root). **MMB** does not.

---

## The problem MMB causes

MMB's delayed-merge schedule means a chunk's leaves can be filled before its h=G ancestor exists in the ops tree. During that window the chunk's leaves are split across **multiple sub-G peaks**.

Concrete picture at `G = 2` (chunk size = 4 leaves), `ops_leaves = 4`:

```
   MMR (immediate merge)              MMB (delayed merge)

         R0   <- single h=2 peak          A     B    <- two peaks at h<2
        /  \                             / \   / \      no single h=2 node
       *    *                           *   * *   *     covering chunk 0
      / \  / \                         / \ / \ / \ /\
     0  1  2  3                        0 1 2 3

   chunk 0 (ops 0..3) maps to        chunk 0 has TWO ancestors below G.
   the single node R0  -->           Cannot graft a chunk onto two ops
   graft normally.                   nodes at once.
```

The old design tried to graft anyway, which required:

- 5-segment range proofs (pure_prefix / prefix_boundary / range / suffix_boundary / pure_suffix)
- multi-peak fold logic in `transform_peak_digests`
- virtual subtree positions
- `prefix_witnesses` / `suffix_witnesses` fields in `RangeProof`
- an MMB-specific `chunk_peaks` iterator
- a 2-branch verifier (naive vs. grafted-reconstruction)

---

## The fix: defer the pending chunk

Treat a "completed-but-not-graftable" chunk like the (already existing) partial trailing chunk: its bytes ride in the canonical-root **trailer** until the ops tree builds its h=G ancestor, at which point the chunk migrates into the grafted tree.

### Lifecycle of a single chunk i (MMB)

```
   ops_leaves N grows ----------------------------->

   chunk i:
   +----------------+------------------+--------------------------+
   |    filling     |     pending      |          graftable          |
   | (bits being    |  (bits complete, |  (h=G ancestor born;     |
   |  appended)     |  ancestor not    |   bitmap chunk grafted   |
   |                |  yet born)       |   onto that ops node)    |
   +----------------+------------------+--------------------------+
                    ^                  ^
        bits all set|                  |ancestor born at
        N=(i+1)*2^G |                  |N = i*2^G + birth_chunk_0

   Where chunk i lives:
       filling -> trailer (partial slot, only the latest chunk)
       pending -> trailer (pending slot, only the most-recent complete chunk)
       graftable  -> grafted tree
```

After this change, **every graftable chunk has exactly one h=G ancestor**. The MMR-friendly assumption holds for both families. Multi-peak fold becomes dead code.

### Counter discipline

Four counts are kept distinct and named consistently throughout the code:

```
   pruned_chunks  <=  graftable_chunks  <=  complete_chunks
                            |                  |
                            |                  +-- partial_chunk_idx = complete_chunks
                            |                      (trailing chunk being filled)
                            |
                            +-- pending_chunk_idx = graftable_chunks
                                (when complete - graftable == 1)
```

**Invariant** (verified by 49.7M property-test tuples): `complete_chunks - graftable_chunks` is in `{0, 1}`. The pending and partial slots are independent: at G >= 3, both can be present simultaneously.

### Canonical root layout

```
   db_root = hash(
       ops_root,
       grafted_root,
       [pending_chunk_digest]?,        <- new: present iff pending chunk exists
       [next_bit_be, partial_chunk_digest]?
   )
```

For MMR (no pending state) and for MMB at non-pending sizes, `db_root` is **byte-identical** to before. It differs only when MMB is in a pending state, where pending bytes move from the grafted root into the trailer (ALPHA permits this break).

### Concrete example: pending + partial coexistence

For `G = 3` MMB, `birth_chunk_0 = 3 * 2^(G-1) - 1 = 11`. Chunk 0 fills at `N = 8` but does not activate until `N = 11`. The pending window is `N` in `[8, 10]`.

Snapshot at `N = 10`:

```
   chunks (8 bits each):    chunk 0  | chunk 1
                           ----------+----------
   ops index range:          0..7    |  8..15
   bits set:               11111111  |   11
                                     |
                          [pending]  | [partial]
                          (in trailer, both)

   graftable_chunks = 0   -->  grafted_root commits to no chunks
   trailer:                 pending = H(chunk_0_bytes)
                            partial = (next_bit=2, H(chunk_1_bytes))

   db_root = hash(ops_root || grafted_root || pending || next_bit || partial)

   Append one more op (N -> 11): chunk 0's h=3 ancestor is born,
   graftable_chunks becomes 1, pending is dropped, chunk 0 migrates
   into the grafted tree.
```

---

## What's deleted

| Surface | Lines | Reason |
|---|---|---|
| `proof/geometry.rs` (5-segment classification) | -394 | collapses to 3-segment standard ranges |
| `proof/mod.rs` two-branch verifier, `prefix/suffix_witness` builders, grafted-reconstruction wrapper | ~-260 | graftable chunks are always single-peak |
| `grafting.rs` `transform_peak_digests`, `transformed_inactive_peaks` | ~-80 | no multi-peak fold needed |
| `merkle/proof.rs` MMB chunk-peaks helpers | -123 | no caller after refactor |
| `prefix_witnesses` / `suffix_witnesses` fields on `RangeProof` | ~-30 | dead under defer-pending |

## What's added

- `graftable_chunks(ops_leaves, G)` -- closed-form predicate derived from `F::peak_birth_size(F::subtree_root_position(0, G), G)`.
- `pending_chunk_digest: Option<D>` field on `OpsRootWitness` and `RangeProof`; verifier rejects presence/absence mismatch.
- `pending_chunk(...)` helper mirroring `partial_chunk()`.
- `combine_roots` extended to a 4-arm match `(pending?, partial?)` covering all trailer combinations.
- `Verifier::graftable_chunks` -- at the grafting height, ancestors of ungraftable chunks return the bare ops subtree root (skip chunk combination).
- Batch path: pending -> graftable transition adds newly-graftable chunks to `chunk_indices_to_update` even when no bits changed.
- Tests: `combine_roots` golden vector; pending+partial coexistence at G=3; pending->graftable transition at exact birth size; range proofs straddling the pending boundary; sync rebuild with pending state; G=8 birth-size invariants.

## Snapshot discipline

`ops_leaves` is now an explicit `u64` parameter on every function that derives `graftable_chunks` (`compute_db_root`, `compute_grafted_root`, `compute_grafted_leaves`, `build_grafted_tree`, `RangeProof::new`, sync rebuild, batch `compute_current_layer`). Callers snapshot once at entry; `pruned <= graftable <= complete` is debug-asserted at every derivation site. This eliminates torn-read windows where two queries in the same logical operation could see different ops sizes.

---

## Cost summary

| | Before | After |
|---|---|---|
| MMR canonical root | -- | byte-identical |
| MMB canonical root, no pending | -- | byte-identical |
| MMB canonical root, pending | -- | differs (ALPHA) |
| Append cost | -- | unchanged |
| Commit cost | -- | marginally faster (no multi-peak fold) |
| Proof size | -- | unchanged or smaller (no boundary witnesses) |
| Verify cost | -- | marginally faster (single reconstruction path) |
| On-disk format | -- | unchanged |
| `OpsRootWitness` / `RangeProof` codec | -- | gains `pending_chunk_digest`; loses `prefix_witnesses` / `suffix_witnesses` |


storage/src/qmdb/current/proof/mod.rs
```diff
@@ -443,10 +228,13 @@ impl<F: Graftable, D: Digest> RangeProof<F, D> {
             None
         };
```
roberto-bayardo: While you're here:
```suggestion
let partial_chunk_digest = crate::qmdb::current::db::partial_chunk::<_, N>(status).map(|(chunk, _)| hasher.digest(&chunk));
```

storage/src/qmdb/current/grafting.rs
```diff
@@ -55,6 +78,38 @@ pub(crate) const fn height<const N: usize>() -> u32 {
...
+/// - For MMR-shaped families, `birth_chunk_0 == 2^G`, so `active = ops_leaves >> G`. There is no
+///   pending state in MMR.
+/// - For MMB, `birth_chunk_0 == 3 * 2^(G-1) - 1`, so `active = (ops_leaves + 1 - 2^(G-1)) >> G`
+///   when the input is at least `birth_chunk_0`, else 0. The pending window has width `2^(G-1) - 1`
+///   ops per chunk, strictly less than the chunk stride `2^G`, so at most one chunk is ever pending.
+///
+/// # Panics
+///
+/// Panics if `grafting_height == 0`. Grafting at height 0 (one leaf per chunk) is structurally
+/// meaningless for QMDB's design and is rejected here to keep the active-chunk arithmetic simple.
```
roberto-bayardo: This comment could be tightened up considerably. E.g. drop the "Derived from" section -- no need to document the implementation here.

storage/src/qmdb/current/grafting.rs
```diff
@@ -378,24 +262,34 @@ pub(super) struct Verifier<'a, F: Graftable, H: CHasher> {
     /// The chunk index of `chunks[0]`.
     start_chunk_index: u64,
 
+    /// Number of chunks committed by the grafted tree. Chunks at index `>= active_chunks`
+    /// are pending: their bytes ride in the canonical-root trailer rather than being combined
+    /// with the ops subtree root at h=G during reconstruction.
+    active_chunks: u64,
```
roberto-bayardo: let's rename this, since an "inactive chunk" (one with all 0 bits) would still be among "active_chunks".  How about `graftable_chunks`?

storage/src/qmdb/current/grafting.rs
```diff
@@ -430,6 +324,13 @@ impl<F: Graftable, H: CHasher> HasherTrait<F> for Verifier<'_, F, H> {
                 // Convert the F-family position to a chunk index using F's leftmost_leaf.
                 let loc = F::leftmost_leaf(pos, self.grafting_height);
                 let chunk_idx = *loc >> self.grafting_height;
+
+                // Pending chunks (at index >= active_chunks) are NOT combined here. Their
+                // bytes are committed via the canonical-root trailer, not via grafting.
```
roberto-bayardo: "canonical-root trailer"?

storage/src/qmdb/current/grafting.rs
```diff
@@ -580,6 +477,147 @@ mod tests {
     use commonware_macros::test_traced;
     use commonware_runtime::{deterministic, Runner};
 
+    /// MMR has no pending state, so every supplied chunk should be treated as active during
+    /// verification. Tests use this sentinel to avoid threading the actual chunk count through
+    /// every call site.
+    const ALL_CHUNKS_ACTIVE: u64 = u64::MAX;
```
roberto-bayardo: `ALL_CHUNKS_GRAFTABLE` ?

storage/src/qmdb/current/grafting.rs
```diff
@@ -1153,129 +1166,73 @@ mod tests {
         assert_eq!(grafted.get_node(gp4), Some(d4));
     }
 
+    /// For every `(leaf_count, chunk_idx)` with `chunk_idx < ops_active_chunks(leaf_count, G)`,
+    /// the chunk has exactly one h=G peak in the ops MMB. The pending chunk (at index
+    /// `active_chunks`, if present) is the only chunk that may have multi-peak structure;
+    /// its bytes ride in the canonical-root trailer rather than being folded into the
```
roberto-bayardo: another use of `canonical-root-trailer` that could be replaced with something more understandable.

storage/src/qmdb/current/mod.rs
```diff
@@ -90,6 +90,13 @@
 //! Above the grafting height, internal nodes use standard hashing over the grafted leaves.
 //! Below the grafting height, the ops tree is unchanged.
 //!
+//! Not every complete chunk is graftable. A chunk is _active_ when its height-`h` ancestor
```
roberto-bayardo: "active" is already a defined concept, and it conflicts with this one.  Suggest naming it `_graftable_`

storage/src/qmdb/current/mod.rs
```diff
@@ -144,20 +151,89 @@
...
-//! ## Partial chunks
+//! Range proofs over a window that includes the trailing pending or partial chunk additionally
+//! carry that chunk's digest in the proof and have the verifier re-derive it from the supplied
+//! chunk bytes. Active-chunk reconstruction is always single-peak: the height-`gh` ancestor
+//! exists in the ops tree and yields a single grafted leaf to fold against the bitmap chunk.
+//!
+//! ## Trailer chunks: partial and pending
+//!
+//! Two kinds of bitmap chunk are *not* committed by `grafted_root`; instead they ride in a
+//! _trailer_ that hashes into the canonical root.
```
roberto-bayardo: Ah, I didn't see this definition when I left my earlier questions about the "canonical root trailer" usage. I still question the value of defining this concept. Maybe just leave it as something like:

"bits in pending_chunk and/or partial_chunk are not graftable, and are directly folded into the root hash"

storage/src/qmdb/current/mod.rs
```diff
@@ -213,26 +289,34 @@
 //! The canonical root of a `current` database is:
 //!
 //! ```text
-//! root = hash(ops_root || grafted_root [|| next_bit || hash(partial_chunk)])
+//! root = hash(
+//!     ops_root
+//!     || grafted_root
+//!     [|| pending_chunk_digest]
+//!     [|| next_bit || partial_chunk_digest]
```
roberto-bayardo, line 291: When there is more than one optional segment being hashed we have to be careful they can never be equal (or your can generate root collisions).  I think in this case we are safe because the two optional components have different sizes, but we should document this explicitly why there is no collision risk.

storage/src/qmdb/current/mod.rs
```diff
@@ -360,13 +444,18 @@ where
 
     let any = any::init_with_bitmap(context.child("any"), config.into(), Some(bitmap)).await?;
 
-    // Build the grafted tree from the bitmap and ops tree.
+    // Build the grafted tree from the bitmap and ops tree. Snapshot ops_leaves once and
+    // thread it through both grafted-tree construction and root computation so the
+    // active-chunks boundary is consistent.
```
roberto-bayardo: I don't think we need to document here that ops_leaves is "snapshotted..."

storage/src/qmdb/current/db.rs
```diff
@@ -1026,17 +1097,25 @@ pub(super) async fn build_grafted_tree<
     bitmap: &impl bitmap::Readable<N>,
     pinned_nodes: &[H::Digest],
     ops_tree: &impl MerkleStorage<F, Digest = H::Digest>,
+    ops_leaves: u64,
```
roberto-bayardo: Why is this u64 instead of Location?

storage/src/qmdb/current/grafting.rs
```diff
@@ -55,6 +78,38 @@ pub(crate) const fn height<const N: usize>() -> u32 {
...
+///   pending state in MMR.
+/// - For MMB, `birth_chunk_0 == 3 * 2^(G-1) - 1`, so `active = (ops_leaves + 1 - 2^(G-1)) >> G`
+///   when the input is at least `birth_chunk_0`, else 0. The pending window has width `2^(G-1) - 1`
+///   ops per chunk, strictly less than the chunk stride `2^G`, so at most one chunk is ever pending.
+///
+/// # Panics
+///
+/// Panics if `grafting_height == 0`. Grafting at height 0 (one leaf per chunk) is structurally
+/// meaningless for QMDB's design and is rejected here to keep the active-chunk arithmetic simple.
+pub(super) fn ops_active_chunks<F: Graftable>(ops_leaves: u64, grafting_height: u32) -> u64 {
```
roberto-bayardo: another ops_leaves that should probably be Location instead of u64?

storage/src/qmdb/current/proof/mod.rs
```diff
@@ -107,263 +148,26 @@ where
...
-    /// Extra suffix witnesses needed when grafted reconstruction must inspect peaks that ops-tree
-    /// proof collection could otherwise hide behind a suffix accumulator.
-    ///
-    /// This vector follows the geometry's suffix segment order: ops-tree peak digests for
-    /// `suffix_boundary`, then bitmap witness digests for `pure_suffix`.
-    pub suffix_witnesses: Vec<D>,
+    /// Digest of the bitmap chunk that's completed but not yet graftable, if any. The
+    /// pending chunk's bytes ride in the canonical-root trailer alongside the partial
+    /// trailing chunk; this digest is what the verifier checks the supplied chunk
+    /// bytes against.
```
roberto-bayardo: I'd drop everything after the first sentence.

storage/src/qmdb/current/proof/mod.rs
```diff
@@ -107,263 +148,26 @@ where
...
+    pub pending_chunk_digest: Option<D>,
 
-    /// The partial chunk digest from the status bitmap at the time of proof generation, if any.
+    /// Digest of the trailing partial chunk (last chunk filling), if the bitmap is not
+    /// chunk-aligned.
```
roberto-bayardo: I prefer the previous comment.

storage/src/qmdb/current/proof/mod.rs
```diff
@@ -397,42 +201,23 @@ impl<F: Graftable, D: Digest> RangeProof<F, D> {
         range: Range<Location<F>>,
         ops_root: D,
     ) -> Result<Self, Error<F>> {
-        let bitmap = BitmapGrafting::new(status);
+        // Snapshot ops_leaves once and thread through every derivation that needs it so the
+        // pruned <= active <= complete invariant holds across all derivations.
```
roberto-bayardo: Another not-very-helpful comment.

storage/src/qmdb/current/proof/mod.rs
```diff
@@ -397,42 +201,23 @@ impl<F: Graftable, D: Digest> RangeProof<F, D> {
...
-            inactivity_floor,
-            bitmap.grafting_height,
-        )?;
-        let geometry = RangeProofGeometry::new(
+        let ops_leaves = *leaves;
+        let grafting_height = grafting::height::<N>();
+        let inactive_peaks =
+            grafting::chunk_aligned_inactive_peaks::<F>(leaves, inactivity_floor, grafting_height)?;
+        // Every active chunk has a single h=G ancestor, so a standard historical range
+        // proof suffices; no multi-peak grafted reconstruction is needed.
```
roberto-bayardo: this comment also feels a bit out of place / unnecessary.

storage/src/qmdb/current/batch.rs
```diff
@@ -654,13 +659,17 @@ where
     };
     let grafted_storage =
         grafting::Storage::new(&layered, grafting_height, &ops_tree_adapter, hasher.clone());
-    // Compute partial chunk (last incomplete chunk, if any).
+    // Compute partial chunk (last incomplete chunk, if any). The partial chunk lives at
+    // index `new_complete_chunks` (the chunk currently being filled with bits) — distinct
```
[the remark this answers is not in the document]
roberto-bayardo: fixed

storage/src/qmdb/current/unordered/mod.rs
```diff
@@ -326,8 +326,7 @@ pub mod tests {
             // commit op.
             let proof = RangeProof {
                 proof: Proof::default(),
-                prefix_witnesses: vec![],
-                suffix_witnesses: vec![],
+                pending_chunk_digest: None,
```
patrick-ogrady: So, is this left empty for MMR? When MMR, do we assert empty on decode?

storage/src/qmdb/current/batch.rs
```diff
@@ -571,36 +571,41 @@ where
...
+    //   - graftable_parent:       grafted-tree leaf count from the parent (structural source of truth)
+    //
+    // The pending chunk (if any) sits at index `graftable_overlay` and is excluded from the
+    // grafted tree; its digest is hashed directly into the canonical root.
+    let new_complete_chunks = overlay.complete_chunks();
+    let graftable_overlay = grafting::graftable_chunks::<F>(*overlay_ops_leaves, grafting_height)
+        .min(new_complete_chunks as u64) as usize;
+    let graftable_parent = *grafted_parent.leaves() as usize;
+    let pruned_chunks = bitmap_parent.pruned_chunks();
+    debug_assert!(
```
patrick-ogrady: assert?

storage/src/qmdb/current/db.rs
```diff
@@ -1026,17 +1114,26 @@ pub(super) async fn build_grafted_tree<
...
     strategy: &S,
 ) -> Result<Mem<F, H::Digest>, Error<F>> {
     let grafting_height = grafting::height::<N>();
     let pruned_chunks = bitmap.pruned_chunks();
     let complete_chunks = bitmap.complete_chunks();
-
-    // Compute grafted leaves for each unpruned complete chunk.
+    let graftable_chunks = grafting::graftable_chunks::<F>(*ops_leaves, grafting_height)
+        .min(complete_chunks as u64) as usize;
+    debug_assert!(
```
patrick-ogrady: assert?

storage/src/qmdb/current/unordered/mod.rs
roberto-bayardo: It must be empty for MMR, otherwise it fail to verify. Added a unit test confirming this.

It wasn't rejecting on decode, but I have added a check there too (+ test).

storage/src/qmdb/current/db.rs
roberto-bayardo: changed

storage/src/qmdb/current/batch.rs
roberto-bayardo: changed

storage/src/merkle/mmr/mod.rs
```diff
@@ -215,6 +215,8 @@ impl merkle::Family for Family {
 }
 
 impl Graftable for Family {
+    const HAS_PENDING_CHUNKS: bool = false;
```
patrick-ogrady: should we have different proof types for this? I suppose its easier to reason about 1 proof type but this feels like a bit of a hack 🤔 .

seems somewhat nice to hide this from callers, that being said, I don't think you'll be hot swapping variants too much (and could deal with a proof under mmr vs mmb)?

storage/src/qmdb/current/db.rs
patrick-ogrady: This debug_assert! is still around (I think you changed the other)

storage/src/merkle/mmr/mod.rs
roberto-bayardo: I think this would propagate a bit further than I'd like but I could take a look later (afk for a bit)

storage/src/qmdb/current/db.rs
roberto-bayardo: Yeah fixed all but this one apparently. Fixed now, and confirmed there are no others in this file.

storage/src/merkle/mmr/mod.rs
roberto-bayardo: As feared making this a different type where MMR's doesn't have the pending_chunk field at all requires a lot of associated changes. I personally don't think it's worth it, but lmk if you'd like me to proceed.

Another option suggested by claude when I asked what might be less hacky yet still contained:

> Validation method on Graftable:                                                                                            
>    
>   Replace the bare bool with a method:                                                                                          
>                                                          
>   // In Graftable
>   fn validate_pending_chunk<D: Digest>(pending: &Option<D>) -> Result<(), codec::Error>;
>                                                                                                                                 
>   MMR's impl rejects Some, MMB's accepts anything. The deserialization site becomes                                             
>   F::validate_pending_chunk(&pending_chunk_digest)? — which reads as the family validating its own invariant rather than        
>   external code checking a flag.

patrick-ogrady: > Validation method on Graftable:

Replace the bare bool with a method:

// In Graftable
fn validate_pending_chunk<D: Digest>(pending: &Option) -> Result<(), codec::Error>;

MMR's impl rejects Some, MMB's accepts anything. The deserialization site becomes
F::validate_pending_chunk(&pending_chunk_digest)? — which reads as the family validating its own invariant rather than
external code checking a flag.

I think this is the right direction if we want to keep it. We could call it `extra_chunk` or `extra`? I'm ok if you want to punt on this if you attach an issue (I think it should also be `HAS_PENDING_CHUNK`?)

roberto-bayardo: I went ahead and implemented the associated-type version making it impossible to have a pending chunk with an MMB family. PTAL?

storage/src/qmdb/current/proof/mod.rs
```diff
@@ -57,20 +57,20 @@ use tracing::debug;
...
-pub struct OpsRootWitness<D: Digest> {
+pub struct OpsRootWitness<F: Graftable, D: Digest> {
     /// The grafted-tree root committed by the canonical root.
     pub grafted_root: D,
 
-    /// `H(pending_chunk_bytes)` when the bitmap has a chunk whose bits are complete but
-    /// whose h=G ancestor has not yet been born in the ops tree; `None` otherwise.
-    pub pending_chunk_digest: Option<D>,
+    /// The pending-chunk contribution. For families that support pending chunks (MMB), this
+    /// is `Option<D>`; for families that don't (MMR), this is `()`.
```
patrick-ogrady: I think we want to avoid describing how the trait is implemented in different places (seems like that could easily drift?)

storage/src/qmdb/current/proof/mod.rs
```diff
@@ -83,19 +83,16 @@ impl<D: Digest> OpsRootWitness<D> {
...
+impl<F: Graftable, D: Digest> Write for OpsRootWitness<F, D> {
     fn write(&self, buf: &mut impl BufMut) {
         self.grafted_root.write(buf);
-        self.pending_chunk_digest.is_some().write(buf);
-        if let Some(digest) = &self.pending_chunk_digest {
-            digest.write(buf);
-        }
+        self.pending_chunk_digest.write(buf);
         self.partial_chunk.is_some().write(buf);
         if let Some((next_bit, digest)) = &self.partial_chunk {
```
patrick-ogrady, line 96: I think we can call encode on this directly?

roberto-bayardo: We're writing the next_bit as a varint here, if we called directly this would be 8 bytes of u64 (most wasted other than the first byte).

storage/src/qmdb/current/proof/mod.rs
roberto-bayardo: yeah usual AI verbosity here, removed from here & a few other places.

patrick-ogrady (approved):

PR #3750, opened 2026-05-11
[storage] Parameterize resolver impls over Strategy

clabby:
## Overview

Allows for QMDB state sync resolvers to be parameterized over the `commonware_parallel::Strategy` and adds test coverage to ensure that a resolver with a non-default `Strategy` can be constructed.


clabby:
As a side-note, this seems to have happened because the storage crate offers a default for the trailing generic all over the place. We should probably remove that to prevent future API footguns, in favor of having to explicitly define a `Strategy`.

roberto-bayardo (approved):

patrick-ogrady (approved):

PR #3752, opened 2026-05-11
[runtime/iobuf/pool] pooled buffers share Layout

andresilva:
This PR moves pooled allocation layout metadata out of each checked-out buffer and into the owning size class. `PooledBuffer` now carries only the allocation pointer, while pooled views keep the metadata needed to return the buffer to the correct pool slot.

The freelist now owns pooled buffer creation and teardown more directly. The freelist has an explicit ownership contract that must be "manually" enforced: buffers created for a freelist must eventually return to that freelist before it is drained or dropped, otherwise checked-out buffers will leak. This is guaranteed in the current code because:
- A `SizeClass` is only dropped when there are no more outstanding buffers for that class
- TLS caches flush into the `SizeClass` freelist on drop

| Case | ns/op |
| --- | ---: |
| Direct (before) | 7.08 |
| Pooled (before) | 11.83 |
| Direct (after) | 6.95 |
| Pooled (after) | 8.71 |

Depends on #3728.


runtime/src/iobuf/buffer.rs
michaelkaplan13, line 1: Nit: May want to rename the file to `backing.rs` (or maybe `buffer.rs`?)

runtime/src/iobuf/buffer.rs
```diff
@@ -103,66 +108,168 @@ impl Drop for AlignedBuffer {
...
+    /// Creates a new uninitialized pooled buffer for `layout`.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `layout` has zero size.
+    #[inline]
+    pub fn new(layout: Layout) -> Self {
+        assert!(layout.size() > 0, "layout size must be non-zero");
+        // SAFETY: layout is valid and non-zero sized.
+        let ptr = unsafe { alloc(layout) };
```
michaelkaplan13, line 151: My non-Rust-trained eyes saw this and thought: "what's unsafe about this? Looks like normal C/C++ to me!" 😅

runtime/src/iobuf/freelist.rs
```diff
@@ -1,9 +1,17 @@
...
+//! with that layout. Buffers that are not checked out and not held in a
+//! thread-local cache are parked in the freelist, making them available for
+//! reuse by any thread in the pool. Each tracked buffer has a stable slot id
+//! within its size class.
+//!
+//! A [`PooledBuffer`] does not carry its allocation layout. Any buffer created
+//! by this freelist, or taken from it, must eventually be returned to the same
+//! freelist before the freelist is finally dropped. [`Freelist::drain`] and
+//! [`Drop`] only release buffers currently parked here, **an outstanding buffer
+//! that is never returned will leak**.
```
michaelkaplan13:
```suggestion
//! [`Drop`] only release buffers currently parked here. **An outstanding buffer
//! that is never returned will leak**.
```

Grammar nit.

michaelkaplan13 (dismissed):
The core buffer pooling/handling makes sense to me. I've never used `loom` before, so need to dive in more there as a go along.

runtime/src/iobuf/buffer.rs
andresilva: Renamed to `buffer.rs`. I want to do a follow-up to reorganize this module and break some files apart (eg. `pool.rs`), so will probably revisit this later.

patrick-ogrady (approved):

PR #3756, opened 2026-05-11
[consensus/marshal] Prevent Unnecessary Erasure Coding

patrick-ogrady:


consensus/src/marshal/coding/types.rs
```diff
@@ -297,15 +297,28 @@ impl<B: Block, C: Scheme, H: Hasher> EncodeSize for CodedBlock<B, C, H> {
...
+        if config != expected.config() {
+            return Err(commonware_codec::Error::Invalid(
+                "CodedBlock",
+                "config mismatch",
+            ));
+        }
+        if inner.digest() != expected.block() {
+            return Err(commonware_codec::Error::Invalid(
+                "CodedBlock",
+                "block digest mismatch",
```
patrick-ogrady, line 340: captialize?

consensus/src/marshal/core/variant.rs
```diff
@@ -60,6 +60,12 @@ pub trait Variant: Clone + Send + Sync + 'static {
     /// Returns the parent commitment referenced by `block`.
     fn parent_commitment(block: &Self::Block) -> Self::Commitment;
 
+    /// Returns the codec configuration used to decode a resolver-delivered block.
+    fn decode_block_cfg(
```
patrick-ogrady: I think we could do better here (maybe defining this codec on Block?)

consensus/src/marshal/core/actor.rs
```diff
@@ -995,20 +995,22 @@ where
...
-                let Ok((finalization, block)) =
-                    <(Finalization<P::Scheme, V::Commitment>, V::Block)>::decode_cfg(
-                        value,
-                        &(
-                            scheme.certificate_codec_config(),
-                            self.block_codec_config.clone(),
-                        ),
-                    )
+                let certificate_codec_config = scheme.certificate_codec_config();
+                let mut value = value;
```
clabby: nit: make value `mut` in the function signature rather than shadowing? Several arms do the `read` -> `decode` pattern.

consensus/src/marshal/core/variant.rs
clabby: `Block` intentionally does not bind `Committable`, so it has no concept of what the associated commitment should be. I think this belongs on `Variant`?

clabby: Though `decode_block_cfg` is a bit of a weird name, since `Cfg` is only present on the `Read` trait already (no need to specify that it's the cfg for decoding). Maybe just `block_cfg`?

consensus/src/marshal/core/variant.rs
```diff
@@ -19,6 +19,9 @@ use commonware_p2p::Recipients;
 use commonware_utils::channel::oneshot;
 use std::future::Future;
 
+/// The codec configuration used to decode a [`Variant::ApplicationBlock`].
+pub type BlockReadCfg<V> = <<V as Variant>::ApplicationBlock as Read>::Cfg;
```
patrick-ogrady: not sure we need this tbh

consensus/src/marshal/core/variant.rs
```diff
@@ -60,6 +63,15 @@ pub trait Variant: Clone + Send + Sync + 'static {
     /// Returns the parent commitment referenced by `block`.
     fn parent_commitment(block: &Self::Block) -> Self::Commitment;
 
+    /// Returns the codec configuration used to decode a resolver-delivered block.
+    ///
+    /// The returned configuration may bind `commitment` so that decoding rejects
+    /// blocks that do not match the expected commitment.
+    fn block_read_cfg(
```
patrick-ogrady: I wonder if this is possible to define on Block for the Codec

clabby (dismissed):

consensus/src/marshal/core/variant.rs
```diff
@@ -63,13 +60,13 @@ pub trait Variant: Clone + Send + Sync + 'static {
     /// Returns the parent commitment referenced by `block`.
     fn parent_commitment(block: &Self::Block) -> Self::Commitment;
 
-    /// Returns the codec configuration used to decode a resolver-delivered block.
+    /// Returns the codec configuration used to decode a resolver-delivered [`Self::Block`].
```
patrick-ogrady: I think we could say "any block sent over the wire"? And double-check that is the case (may be a path in buffer)?

patrick-ogrady:
bugbot run

PR #3758, opened 2026-05-11
[coding] Reject non-canonical RS encoding

clabby:
## Overview

Fixes two issues in `ReedSolomon`:
- A byzantine proposer could construct a payload with non-canonical shard length that reconstructed to the same payload, but produced a different root when re-encoding.
- When rebuilding the BMT during `decode`, `check` would cache the digests of provided shards and only recompute the digests for missing shards. A root containing `k` canonical shards and one extra non-canonical checked shard could pass, even though `encode(data)` could produce a different root.

### DIFF
--- a/coding/src/reed_solomon.rs
+++ b/coding/src/reed_solomon.rs
@@ -476,7 +476,7 @@ fn decode<'a, H: Hasher, S: Strategy>(
     let encoding = encoder.encode().map_err(Error::ReedSolomon)?;
     shards.extend(encoding.recovery_iter());
 
-    // Build Merkle tree from the canonical codeword.
+    // Build Merkle tree from the canonical codeword
     let mut builder = Builder::<H>::new(n);
     let shard_hashes = strategy.map_init_collect_vec(&shards, H::new, |hasher, shard| {
         hasher.update(shard);
--- a/coding/src/reed_solomon.rs
+++ b/coding/src/reed_solomon.rs
@@ -210,7 +210,7 @@ const fn canonical_shard_len(data_len: usize, k: usize) -> usize {
 /// The first `k` shards, when concatenated, form `[length_prefix | data | padding]`.
 /// This function copies only the data bytes while validating trailing zero
 /// padding directly from the shard slices.
-fn extract_data(shards: &[&[u8]], k: usize, shard_len: usize) -> Result<Vec<u8>, Error> {
+fn extract_data(shards: &[&[u8]], k: usize, expected_shard_len: usize) -> Result<Vec<u8>, Error> {
     let shards = shards.get(..k).ok_or(Error::NotEnoughChunks)?;
     let data_len = read_data_len(shards)?;
     let mut data = Vec::with_capacity(data_len);
@@ -244,7 +244,7 @@ fn extract_data(shards: &[&[u8]], k: usize, shard_len: usize) -> Result<Vec<u8>,
     }
 
     // Validate that the original shards use the canonical shard width.
-    if canonical_shard_len(data.len(), k) != shard_len {
+    if canonical_shard_len(data.len(), k) != expected_shard_len {
         return Err(Error::Inconsistent);
     }
     Ok(data)
--- a/coding/src/reed_solomon.rs
+++ b/coding/src/reed_solomon.rs
@@ -187,15 +187,8 @@ where
 /// The buffer layout is `[length_prefix | data | zero_padding]` split into
 /// `k` equal-sized shards of `shard_len` bytes each.
 fn prepare_data(mut data: impl Buf, k: usize) -> (Vec<u8>, usize) {
-    // Compute shard length
     let data_len = data.remaining();
-    let prefixed_len = u32::SIZE + data_len;
-    let mut shard_len = prefixed_len.div_ceil(k);
-
-    // Ensure shard length is even (required for optimizations in `reed-solomon-simd`)
-    if !shard_len.is_multiple_of(2) {
-        shard_len += 1;
-    }
+    let shard_len = canonical_shard_len(data_len, k);
 
     // Prepare data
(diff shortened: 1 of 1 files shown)


clabby:
bugbot run

coding/src/reed_solomon.rs
```diff
@@ -206,12 +187,30 @@ fn prepare_data(mut data: impl Buf, k: usize) -> (Vec<u8>, usize) {
     (padded, shard_len)
 }
 
-/// Extract data from encoded shards.
+/// Return the canonical shard width for a payload and shard count.
+///
+/// Encoding prefixes the payload with its length, splits the result across
+/// `k` original shards, and rounds up to an even width required by
+/// `reed-solomon-simd`. Decode uses the same calculation to reject commitments
```
patrick-ogrady, line 207: while this is currently a "convenient excuse", I think we want to keep references to `reed-solomon-simd` out of public docs around correctness?

coding/src/reed_solomon.rs
```diff
@@ -472,37 +476,23 @@ fn decode<'a, H: Hasher, S: Strategy>(
...
-        H::new,
-        |hasher, i| {
-            hasher.update(shards[i]);
-            (i, hasher.finalize())
-        },
-    ) {
-        shard_digests[i] = Some(digest);
-    }
-
+    // Build Merkle tree from the canonical codeword.
```
patrick-ogrady, line 501: remove `.`

coding/src/reed_solomon.rs
clabby: I agree on public docs, though this is an internal function. Useful context imo?

Not super attached though.

patrick-ogrady: We can leave for now.

coding/src/reed_solomon.rs
```diff
@@ -398,7 +401,7 @@ fn decode<'a, H: Hasher, S: Strategy>(
 
     // Process checked chunks
     let shard_len = first.shard.len();
-    let mut shard_digests: Vec<Option<H::Digest>> = vec![None; n];
+    let mut seen_shards = vec![false; n];
```
cronokirby: Nit: use a Bitmap?

cronokirby (commented):
In terms of the strategy, I think it would make more sense if `CheckedChunk` was also guaranteed to have the correct hash, rather than `decode` needing to do this. Is this not possible?

cronokirby:
> In terms of the strategy, I think it would make more sense if `CheckedChunk` was also guaranteed to have the correct hash, rather than `decode` needing to do this. Is this not possible?

okay, I think I understand why this doesn't work, in that if a chunk contains the wrong data, you wouldn't be able to tell this until you do the re-encode.

coding/src/reed_solomon.rs
```diff
@@ -206,12 +187,30 @@ fn prepare_data(mut data: impl Buf, k: usize) -> (Vec<u8>, usize) {
...
+    shard_len
+}
+
+/// Extract data from encoded shards and verify that original shards use the canonical width.
 ///
 /// The first `k` shards, when concatenated, form `[length_prefix | data | padding]`.
 /// This function copies only the data bytes while validating trailing zero
 /// padding directly from the shard slices.
-fn extract_data(shards: &[&[u8]], k: usize) -> Result<Vec<u8>, Error> {
+fn extract_data(shards: &[&[u8]], k: usize, shard_len: usize) -> Result<Vec<u8>, Error> {
```
patrick-ogrady: nit: expected_shard_len?

coding/src/reed_solomon.rs
```diff
@@ -86,36 +86,24 @@ impl<D: Digest> Chunk<D> {
...
-/// commitment root it was verified against. The root is checked at decode
-/// time to prevent cross-commitment shard mixing.
+/// This stores the commitment root that [`Chunk::verify`] checked against.
+/// The root is checked at decode time to prevent cross-commitment shard mixing.
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct CheckedChunk<D: Digest> {
     root: D,
     shard: Bytes,
     index: u16,
-    digest: D,
```
patrick-ogrady, line 108: Can we not still store this if we know the expected shard length when decoding (I think we should)?

clabby: We cannot. The issue pertains to checking that the shard contents are correct, per the canonical encoding, rather than just the length.

The attack would be that:
- Encoding is constructed with a recovery shard that doesn't respect the padding rules.
- Honest participant verifies `f+1` shards against the root with improper encoding, all pass BMT inclusion.
- `decode` would use the cached digests created during check, not verifying against the canonical re-encoding of the data.
- `decode` passes, where it should have failed since the encoding didn't expect the rules of `encode`.

While we could cache the digest, we'd have to check equality of the given shards vs. the re-encoding anyways.

patrick-ogrady: > decode would use the cached digests created during check, not verifying against the canonical re-encoding of the data.

What I had in mind is comparing against the raw shard bytes from the re-encoding. I think that should be faster than hashing (and if it is correct then we use the already computed digest)?

Maybe that isn't the case tho.

patrick-ogrady (dismissed):

coding/src/reed_solomon.rs
```diff
@@ -187,15 +187,8 @@ where
 /// The buffer layout is `[length_prefix | data | zero_padding]` split into
 /// `k` equal-sized shards of `shard_len` bytes each.
 fn prepare_data(mut data: impl Buf, k: usize) -> (Vec<u8>, usize) {
-    // Compute shard length
```
patrick-ogrady, line 190: retain this comment

patrick-ogrady (approved):

PR #3764, opened 2026-05-12
[consensus/marshal] `commonware-glue` marshal changes

clabby:
## Overview

Ports the changes to marshal from #3381:
- Add a `GetProcessedHeight` function to marshal's mailbox, so that the startup routine can query marshal's processed height to determine if it needs to roll back one or more databases.

### DIFF
--- a/consensus/src/marshal/ancestry.rs
+++ b/consensus/src/marshal/ancestry.rs
@@ -2,7 +2,6 @@
 
 use crate::{types::Height, Block, Heightable};
 use commonware_cryptography::Digestible;
-use commonware_utils::sync::Mutex;
 use futures::{
     future::{BoxFuture, OptionFuture},
     FutureExt, Stream,
@@ -49,16 +48,12 @@ pub struct ErasedBlockProvider<B: Block> {
 
 impl<B: Block> ErasedBlockProvider<B> {
     /// Erase a concrete [`BlockProvider`] into a type-erased provider.
-    ///
-    /// The provider is wrapped in a `Mutex` so that the resulting closure
-    /// is `Sync` (required by `Arc<dyn Fn + Send + Sync>`). The lock is
-    /// held only for the duration of `clone()`, never across an await.
-    pub fn new<M: BlockProvider<Block = B>>(provider: M) -> Self {
-        let provider = Arc::new(Mutex::new(provider));
+    pub fn new<M: BlockProvider<Block = B> + Sync>(provider: M) -> Self {
+        let provider = Arc::new(provider);
         Self {
             fetch: Arc::new(move |digest| {
-                let p = provider.lock().clone();
-                Box::pin(async move { p.fetch_block(digest).await })
+                let p = provider.as_ref().clone();
+                Box::pin(p.fetch_block(digest))
             }),
         }
     }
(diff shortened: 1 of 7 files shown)


consensus/src/marshal/ancestry.rs
```diff
@@ -34,6 +36,45 @@ pub trait BlockProvider: Clone + Send + 'static {
...
+pub struct ErasedBlockProvider<B: Block> {
+    fetch: Arc<dyn Fn(<B as Digestible>::Digest) -> BoxFuture<'static, Option<B>> + Send + Sync>,
+}
+
+impl<B: Block> ErasedBlockProvider<B> {
+    /// Erase a concrete [`BlockProvider`] into a type-erased provider.
+    ///
+    /// The provider is wrapped in a `Mutex` so that the resulting closure
+    /// is `Sync` (required by `Arc<dyn Fn + Send + Sync>`). The lock is
+    /// held only for the duration of `clone()`, never across an await.
```
clabby: rm - impl detail

consensus/src/marshal/mocks/harness.rs
```diff
@@ -1,4 +1,3 @@
-//! Test harness for marshal variants.
 //!
```
BrendanChou, line 2: If intended, delete one more empty line `//!`

consensus/src/marshal/core/actor.rs
```diff
@@ -659,20 +660,41 @@ where
...
+                            Recipients::Some(peers) => {
+                                if peers.is_empty() {
+                                    continue;
+                                }
+                                let peers = NonEmptyVec::from_unchecked(peers);
+                                resolver.fetch_targeted(request, peers).await;
+                            },
+                            Recipients::One(peer) => {
+                                let peers = NonEmptyVec::new(peer);
+                                resolver.fetch_targeted(request, peers).await;
```
BrendanChou: Should we also change the API of resolver to use `Recipients`? I guess not (?)

BrendanChou (dismissed):

consensus/src/marshal/core/actor.rs
clabby: My thinking was no, since it's generic over the transport?

BrendanChou (dismissed):

consensus/src/marshal/core/actor.rs
```diff
@@ -659,20 +660,41 @@ where
...
                         response.send_lossy(finalization);
                     }
+                    Message::GetProcessedHeight { response } => {
+                        response.send_lossy(self.last_processed_height);
+                    }
                     Message::HintFinalized { height, targets } => {
-                        // Skip if height is at or below the floor
-                        if height <= self.last_processed_height {
+                        // Skip if height is below the floor
+                        if height < self.last_processed_height {
```
patrick-ogrady: What is the rationale behind this change? if we've processed a block, we must have a finalization at or above this?

clabby: When we kick off state sync, we set the floor far in the future. That block has not necessarily been processed yet. We need to allow for marshal to fetch the floor, in the case that it is missing.

patrick-ogrady: Got it. That becomes very similar to something I was considering for my state sync cleanup.

patrick-ogrady (approved):

PR #3766, opened 2026-05-12
[storage/qmdb] appropriate items-per-blob in apply_batch benchmark

roberto-bayardo:
the apply_batch benchmarks were using the small default (50K) items-per-blob value, which caused 3 of the 4 cases to cross blob boundaries, triggering fsync calls that dominated the timings.  This PR configures it so that we never cross boundaries to truly measure the typical apply-batch overhead in all 4 cases.

Also generalizes the common config-generators used by benchmarks to accept items-per-blob.

### DIFF
--- a/storage/src/qmdb/benches/apply_batch.rs
+++ b/storage/src/qmdb/benches/apply_batch.rs
@@ -1,6 +1,6 @@
 //! Benchmarks for applying already-merkleized QMDB batches.
 
-use crate::common::{any_fix_cfg, make_fixed_value, seed_db, AnyUFixDb, CHUNK_SIZE};
+use crate::common::{any_fix_cfg_with, make_fixed_value, seed_db, AnyUFixDb, CHUNK_SIZE};
 use commonware_cryptography::{Hasher as _, Sha256};
 use commonware_runtime::{
     benchmarks::{context, tokio},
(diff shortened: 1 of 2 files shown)


storage/src/qmdb/benches/apply_batch.rs
```diff
@@ -1,19 +1,24 @@
...
 use rand::{rngs::StdRng, RngCore, SeedableRng};
-use std::time::{Duration, Instant};
+use std::{
+    num::NonZeroU64,
+    time::{Duration, Instant},
+};
 
 const NUM_KEYS: u64 = 65_536;
 const UPDATES: [u64; 1] = [16_384];
+const ITEMS_PER_BLOB: NonZeroU64 = NZU64!(10_000_000);
```
clabby, line 21: For maintainability, maybe we should derive this value from `NUM_KEYS` / `UPDATES` to ensure that it stays large enough?

clabby (approved):

storage/src/qmdb/benches/apply_batch.rs
roberto-bayardo: yes we need a broader fix to remove this footgun entirely, I'll work on that separately though. This PR is just to unblock my apply_batch optimization work in the meantime. thanks!

PR #3767, opened 2026-05-12
[runtime/iobuf/pool] remove Arc from pooled hot path

andresilva:
This PR removes `Arc<SizeClass>` from the pooled buffer hot path without changing the pool lifetime model. Before this PR, checked-out pooled buffers and TLS cache entries both carried an `Arc<SizeClass>`. The steady local reuse path did not increment or decrement the strong count, but `Arc` is still a non-`Copy` value with drop glue, so moving it through allocation, return, and cache-entry storage made the compiler preserve destructor-aware code around what should be a simple ownership transfer.

Checked-out pooled buffers now carry a `SizeClassLease`, which is one strong size-class reference represented as a raw pointer with explicit retain/release at the global freelist boundary. Moving between checked-out state and the thread-local cache is now a plain pointer transfer. The size class is still kept alive for every outstanding pooled buffer: the reference is banked in the TLS cache while the buffer is local, then materialized back into a lease when the buffer is checked out again.

The thread-local cache stores the size-class identity once and banks one strong reference for each initialized local entry. Individual cache entries shrink to just `(buffer, slot)`, and the cache storage is a fixed flat stack backed by `Box<[MaybeUninit<_>]>`. That keeps the steady local pop/push path to stack length updates plus one 16-byte entry load/store, while global refill, spill, flush, and drop continue to preserve the freelist ownership contract.

Representative local results for the 1KiB no-touch steady-state profile:

| Case | ns/op |
| --- | ---: |
| Direct (before) | 6.95 |
| Pooled (before) | 8.71 |
| | |
| Direct [glibc 2.42] (after) | 6.63 |
| Direct [mimalloc 3.3.0] (after) | 8.28 |
| Direct [jemalloc 5.3.0] (after) | 6.61 |
| Pooled (after) | **2.83** 🏆 |

Depends on #3752.

### DIFF
--- a/runtime/src/iobuf/buffer.rs
+++ b/runtime/src/iobuf/buffer.rs
@@ -116,13 +116,13 @@ impl Drop for AlignedBuffer {
 /// moving pooled buffers through pooled backing values and thread-local caches
 /// does not need to move a per-buffer [`Layout`].
 ///
-/// `PooledBuffer` does not implement [`Drop`] and has no reference to its
-/// originating [`SizeClass`](super::pool::SizeClass). In normal pool use it is
-/// owned by [`PooledBacking`] while outside the global freelist and by the
-/// size-class [`super::freelist::Freelist`] while globally free. Only the
-/// freelist knows the layout needed to deallocate it.
+/// `PooledBuffer` does not implement [`Drop`] and does not retain its
+/// originating [`SizeClass`](super::pool::SizeClass). In normal pool use,
+/// [`PooledBacking`] owns it while it is outside the global freelist, and the
+/// size-class [`super::freelist::Freelist`] owns it while it is globally free.
+/// Only the freelist knows the layout needed to deallocate it.
 ///
-/// Code that takes a `PooledBuffer` out of pool state must either return it to
+/// Callers that take a `PooledBuffer` out of pool state must either return it to
 /// the same originating size class or explicitly deallocate it with the exact
 /// layout used to create it.
 pub struct PooledBuffer {
--- a/runtime/src/iobuf/buffer.rs
+++ b/runtime/src/iobuf/buffer.rs
@@ -6,14 +6,15 @@
 //!   deallocates directly on drop.
 //! - [`PooledBuffer`] is the raw pooled allocation handle: it carries only a
 //!   pointer and relies on pool-owned metadata to release it.
-//! - Pooled views pair that handle with its originating [`SizeClass`] so the
-//!   buffer is returned to the pool on drop.
+//! - Pooled views pair that handle with its originating
+//!   [`SizeClass`](super::pool::SizeClass) so the buffer is returned to the pool
+//!   on drop.
 //!
 //! The allocator-facing pool logic lives in `pool.rs`. This module only deals
 //! with backing ownership and view semantics.
 
 use super::IoBuf;
-use crate::iobuf::pool::{BufferPoolThreadCache, SizeClass};
+use crate::iobuf::pool::{BufferPoolThreadCache, SizeClassLease};
 use bytes::Bytes;
 use std::{
     alloc::{alloc, alloc_zeroed, dealloc, handle_alloc_error, Layout},
@@ -116,9 +117,9 @@ impl Drop for AlignedBuffer {
 /// does not need to move a per-buffer [`Layout`].
 ///
 /// `PooledBuffer` does not implement [`Drop`] and has no reference to its
-/// originating [`SizeClass`]. It must be paired with pool metadata that returns
-/// it to the size class, or explicitly deallocated with the exact layout used
-/// to create it.
+/// originating [`SizeClass`](super::pool::SizeClass). It must be paired with
+/// pool metadata that returns it to the size class, or explicitly deallocated
+/// with the exact layout used to create it.
 pub struct PooledBuffer {
     ptr: NonNull<u8>,
 }
@@ -194,8 +195,8 @@ impl PooledBuffer {
 /// [`Self::release`].
 ///
 /// [`AlignedBuffer`] is self-contained and releases directly. [`PooledBacking`]
-/// pairs a layoutless [`PooledBuffer`] with its [`SizeClass`] and slot id, so
-/// release returns the allocation to the pool.
+/// pairs a layoutless [`PooledBuffer`] with its [`SizeClassLease`] and slot id,
+/// so release returns the allocation to the pool.
 pub(crate) trait BufferBacking: Send + Sync + 'static {
     /// Returns the full allocation capacity, ignoring any view cursor/offset.
     fn capacity(&self) -> usize;
@@ -226,23 +227,20 @@ impl BufferBacking for AlignedBuffer {
 
 /// Pooled backing storage.
 ///
-/// This pairs a layoutless [`PooledBuffer`] with the metadata needed to return
-/// it to the pool. The [`Arc<SizeClass>`] keeps the originating size class, and
-/// therefore its freelist layout, alive while the buffer is checked out or
-/// shared through immutable views.
-///
-/// The `Arc<SizeClass>` is *moved* (not cloned) through the TLS cache on the
-/// steady-state alloc/free path, avoiding refcount traffic.
+/// This pairs a layoutless [`PooledBuffer`] with the lease needed to return it
+/// to the pool. The [`SizeClassLease`] points to the originating size class and
+/// owns that class reference while the buffer is checked out or shared through
+/// immutable views.
 pub(crate) struct PooledBacking {
     buffer: PooledBuffer,
-    class: Arc<SizeClass>,
+    lease: SizeClassLease,
     slot: u32,
 }
 
 impl BufferBacking for PooledBacking {
     #[inline(always)]
     fn capacity(&self) -> usize {
-        self.class.size()
+        self.lease.size()
     }
 
     #[inline(always)]
@@ -252,7 +250,7 @@ impl BufferBacking for PooledBacking {
 
     #[inline(always)]
     fn release(self) {
-        BufferPoolThreadCache::push(self.class, self.slot, self.buffer);
+        BufferPoolThreadCache::push(self.lease, self.slot, self.buffer);
     }
 }
 
@@ -713,8 +711,8 @@ pub(crate) type AlignedBuf = Buf<AlignedBuffer>;
 /// Immutable, reference-counted view over a pooled allocation.
 ///
 /// The final reference returns the underlying allocation to its originating
-/// [`SizeClass`]. See [`Buf`] for the shared immutable view layout and
-/// invariants.
+/// [`SizeClass`](super::pool::SizeClass). See [`Buf`] for the shared immutable
+/// view layout and invariants.
 pub(crate) type PooledBuf = Buf<PooledBacking>;
 
 /// Mutable view over an untracked aligned allocation.
@@ -726,7 +724,8 @@ pub(crate) type AlignedBufMut = BufMut<AlignedBuffer>;
 /// Mutable view over a pooled allocation.
 ///
 /// When dropped, the underlying allocation is returned to its originating
-/// [`SizeClass`]. See [`BufMut`] for the shared mutable layout and invariants.
+/// [`SizeClass`](super::pool::SizeClass). See [`BufMut`] for the shared mutable
+/// layout and invariants.
 pub(crate) type PooledBufMut = BufMut<PooledBacking>;
 
 impl std::fmt::Debug for Buf<AlignedBuffer> {
@@ -790,11 +789,11 @@ impl std::fmt::Debug for BufMut<PooledBacking> {
 
 impl BufMut<PooledBacking> {
     #[inline]
-    pub(crate) const fn new(buffer: PooledBuffer, class: Arc<SizeClass>, slot: u32) -> Self {
+    pub(crate) const fn new(buffer: PooledBuffer, lease: SizeClassLease, slot: u32) -> Self {
         Self {
             inner: ManuallyDrop::new(BufInner::new(PooledBacking {
                 buffer,
-                class,
+                lease,
                 slot,
             })),
             cursor: 0,
--- a/runtime/src/iobuf/buffer.rs
+++ b/runtime/src/iobuf/buffer.rs
@@ -5,10 +5,10 @@
 //! - [`AlignedBuffer`] is self-contained: it carries its own [`Layout`] and
 //!   deallocates directly on drop.
 //! - [`PooledBuffer`] is the raw pooled allocation handle: it carries only a
-//!   pointer and relies on pool-owned metadata to release it.
-//! - Pooled views pair that handle with its originating
-//!   [`SizeClass`](super::pool::SizeClass) so the buffer is returned to the pool
-//!   on drop.
+//!   pointer and does not know its allocation layout.
+//! - [`PooledBacking`] is the pooled ownership bundle outside the global
+//!   freelist: it pairs that handle with a [`SizeClassLease`] and slot id so the
+//!   buffer can return to its originating [`SizeClass`](super::pool::SizeClass).
 //!
 //! The allocator-facing pool logic lives in `pool.rs`. This module only deals
 //! with backing ownership and view semantics.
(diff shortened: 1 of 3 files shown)


runtime/src/iobuf/pool.rs
```diff
@@ -452,15 +452,164 @@ impl PoolMetrics {
...
+    #[inline(always)]
+    const unsafe fn from_banked_ptr(ptr: ptr::NonNull<SizeClass>) -> Self {
+        Self { ptr }
+    }
+
+    /// Retains one banked class reference without materializing a lease.
+    #[inline(always)]
+    fn retain_banked(class: &Arc<SizeClass>) {
+        // SAFETY: the pointer comes from a live `Arc<SizeClass>` borrowed above.
+        unsafe { Arc::increment_strong_count(Self::class_ptr(class).as_ptr()) };
```
patrick-ogrady: Got this:

[P1] Use into_raw-derived pointers for Arc refcounts — /Users/patrickogrady/code/monorepo-pr-
3767/runtime/src/iobuf/pool.rs:532-532
When any buffer leaves the global freelist, this calls Arc::increment_strong_count on the pointer
returned by class_ptr, but class_ptr is built with Arc::as_ptr. The Arc manual refcount APIs
require a pointer obtained from Arc::into_raw (and the later decrements use the same pointer), so
the new lease scheme violates the unsafe precondition for pooled allocations/refills; store an
into_raw-derived token or keep a real Arc for these refs.

runtime/src/iobuf/pool.rs
```diff
@@ -452,15 +452,164 @@ impl PoolMetrics {
...
+// SAFETY: `SizeClassLease` points at a `SizeClass`, which is `Send`, and owns a
+// strong reference while it is live.
+unsafe impl Send for SizeClassLease {}
+// SAFETY: same argument as `Send`, shared access to `SizeClass` is synchronized.
+unsafe impl Sync for SizeClassLease {}
+
+impl SizeClassLease {
+    /// Returns the raw `Arc` pointer for `class`.
+    ///
+    /// `Arc::increment_strong_count` and `Arc::decrement_strong_count` require
```
patrick-ogrady: • The comment is the bug. The documented safety contract for Arc::increment_strong_count/
  decrement_strong_count is an Arc::into_raw pointer, not an arbitrary pointer to the same
  allocation. Arc::as_ptr can produce the same address, but it is only borrowed provenance and does
  not satisfy the manual refcount API contract, which is exactly why Miri can reject it.

runtime/src/iobuf/pool.rs
patrick-ogrady: What is the smallest diff?

```
index 07a0292ce..20d8e5b91 100644
@@ -489,28 +489,31 @@ unsafe impl Send for SizeClassLease {}
 unsafe impl Sync for SizeClassLease {}
 
 impl SizeClassLease {
-    /// Returns the raw `Arc` pointer for `class`.
+    /// Retains one class reference and returns its raw `Arc` token.
     ///
-    /// `Arc::increment_strong_count` and `Arc::decrement_strong_count` require
-    /// pointers derived from `Arc::as_ptr`/`Arc::into_raw`. Building the pointer
-    /// through `&SizeClass` is equivalent for normal codegen, but loses the
-    /// provenance Miri expects for raw `Arc` refcount operations.
+    /// The matching release path uses `Arc::decrement_strong_count`, whose
+    /// safety contract requires a pointer derived from `Arc::into_raw`. Use a
+    /// real `Arc` clone for the retain side, then keep the raw token for
+    /// explicit release.
     #[inline(always)]
-    fn class_ptr(class: &Arc<SizeClass>) -> ptr::NonNull<SizeClass> {
-        // SAFETY: `Arc::as_ptr` never returns null.
-        unsafe { ptr::NonNull::new_unchecked(Arc::as_ptr(class).cast_mut()) }
+    fn retain_raw(class: &Arc<SizeClass>) -> ptr::NonNull<SizeClass> {
+        // SAFETY: `Arc::into_raw` never returns null.
+        unsafe { ptr::NonNull::new_unchecked(Arc::into_raw(Arc::clone(class)).cast_mut()) }
     }
 
-    /// Builds a lease from a banked class reference.
+    /// Returns a non-owning raw `Arc` token for `class`.
     ///
-    /// # Safety
-    ///
-    /// The caller must have already retained one strong reference to `class`,
-    /// and that retained reference must be transferred to the returned lease.
+    /// The returned pointer is derived from `Arc::into_raw`, but it does not own
+    /// a strong reference. Callers must use it only while some other strong
+    /// reference keeps the allocation live.
     #[inline(always)]
-    unsafe fn from_banked(class: &Arc<SizeClass>) -> Self {
-        // SAFETY: guaranteed by the caller.
-        unsafe { Self::from_banked_ptr(Self::class_ptr(class)) }
+    fn class_token(class: &Arc<SizeClass>) -> ptr::NonNull<SizeClass> {
+        let ptr = Self::retain_raw(class);
+        // SAFETY: `ptr` owns the strong reference retained above. The borrowed
+        // `class` keeps the allocation live after this temporary reference is
+        // released.
+        unsafe { Arc::decrement_strong_count(ptr.as_ptr()) };
+        ptr
     }
 
     /// Builds a lease from a banked class reference and raw class identity.
@@ -519,17 +522,21 @@ impl SizeClassLease {
     ///
     /// The caller must have already retained one strong reference to `ptr`, and
     /// that retained reference must be transferred to the returned lease. `ptr`
-    /// must point to the matching live size class.
+    /// must come from `Arc::into_raw` for the matching live size class.
     #[inline(always)]
     const unsafe fn from_banked_ptr(ptr: ptr::NonNull<SizeClass>) -> Self {
         Self { ptr }
     }
 
     /// Retains one banked class reference without materializing a lease.
+    ///
+    /// # Safety
+    ///
+    /// `ptr` must come from `Arc::into_raw` for a matching live size class.
     #[inline(always)]
-    fn retain_banked(class: &Arc<SizeClass>) {
-        // SAFETY: the pointer comes from a live `Arc<SizeClass>` borrowed above.
-        unsafe { Arc::increment_strong_count(Self::class_ptr(class).as_ptr()) };
+    unsafe fn retain_banked_ptr(ptr: ptr::NonNull<SizeClass>) {
+        // SAFETY: guaranteed by the caller.
+        unsafe { Arc::increment_strong_count(ptr.as_ptr()) };
     }
 
     /// Retains `class` for a buffer leaving the global freelist.
@@ -538,10 +545,10 @@ impl SizeClassLease {
     /// touching the strong count.
     #[inline(always)]
     fn retain(class: &Arc<SizeClass>) -> Self {
-        Self::retain_banked(class);
-        // SAFETY: `retain_banked` above retained the strong reference
+        let ptr = Self::retain_raw(class);
+        // SAFETY: `retain_raw` above retained the strong reference
         // transferred to the returned lease.
-        unsafe { Self::from_banked(class) }
+        unsafe { Self::from_banked_ptr(ptr) }
     }
 
     /// Transfers this checked-out lease into a TLS cache entry.
@@ -692,12 +699,12 @@ struct TlsSizeClassCacheEntry {
 /// returning them to the global freelist happens only on miss refill, overflow,
 /// explicit flush, or thread exit.
 ///
-/// `class` is a non-owning identity pointer. It can be stale while `len == 0`,
-/// because an empty cache does not keep its pool alive. When `len > 0`, each
-/// initialized entry in `entries[..len]` owns one banked size-class reference,
-/// which keeps the pointed-to class alive. The entry itself stays small
-/// (`buffer, slot`); popping it materializes the checked-out [`SizeClassLease`]
-/// from that banked reference.
+/// `class` is a non-owning `Arc::into_raw`-derived token for this cache's class.
+/// It can be stale while `len == 0`, because an empty cache does not keep its
+/// pool alive. When `len > 0`, each initialized entry in `entries[..len]` owns
+/// one banked size-class reference, which keeps the pointed-to class alive. The
+/// entry itself stays small (`buffer, slot`); popping it materializes a
+/// checked-out [`SizeClassLease`] from one banked reference.
 ///
 /// The hot steady-state allocation path pops an entry from `entries`, and the
 /// hot return path pushes one back while there is room.
@@ -716,7 +723,7 @@ impl TlsSizeClassCache {
             .collect::<Vec<_>>()
             .into_boxed_slice();
         Self {
-            class: SizeClassLease::class_ptr(class),
+            class: SizeClassLease::class_token(class),
             entries,
             len: 0,
             capacity,
@@ -730,9 +737,12 @@ impl TlsSizeClassCache {
     /// batch-take from the global freelist, return the first claimed buffer,
     /// and retain the rest locally for future allocations.
     #[inline(always)]
-    fn pop(&mut self, class: &Arc<SizeClass>) -> Option<TlsSizeClassCacheEntry> {
+    fn pop(&mut self, class: &Arc<SizeClass>) -> Option<(TlsSizeClassCacheEntry, SizeClassLease)> {
         if let Some(entry) = self.pop_local() {
-            return Some(entry);
+            // SAFETY: each initialized entry owns one banked class reference.
+            // This pop transfers one of those references to the returned lease.
+            let lease = unsafe { SizeClassLease::from_banked_ptr(self.class) };
+            return Some((entry, lease));
         }
 
         // Take from the class-global freelist on a local miss.
@@ -763,13 +773,17 @@ impl TlsSizeClassCache {
     /// the refill and batching code out of `BufferPoolInner::try_alloc`, reducing
     /// hot-path code size and register pressure.
     #[inline(never)]
-    fn pop_global(&mut self, class: &Arc<SizeClass>) -> Option<TlsSizeClassCacheEntry> {
+    fn pop_global(
+        &mut self,
+        class: &Arc<SizeClass>,
+    ) -> Option<(TlsSizeClassCacheEntry, SizeClassLease)> {
         // Tiny caches do not batch enough to justify the wider global
         // claim. Keep their miss path equivalent to a single take.
         if self.capacity < MIN_TLS_BATCH_CAPACITY {
             return class.global.take().map(|(slot, buffer)| {
-                SizeClassLease::retain_banked(class);
-                TlsSizeClassCacheEntry { buffer, slot }
+                let lease = SizeClassLease::retain(class);
+                let entry = TlsSizeClassCacheEntry { buffer, slot };
+                (entry, lease)
             });
         }
 
@@ -779,12 +793,17 @@ impl TlsSizeClassCache {
         let mut entry = None;
         let take = self.capacity / 2;
         class.global.take_batch(take, |slot, buffer| {
-            SizeClassLease::retain_banked(class);
+            // SAFETY: `self.class` is an `Arc::into_raw`-derived token for
+            // `class`, and the borrowed `class` keeps the allocation live.
+            unsafe { SizeClassLease::retain_banked_ptr(self.class) };
             let cache_entry = TlsSizeClassCacheEntry { buffer, slot };
             if entry.is_none() {
                 // Hand the first claimed buffer to the allocation that missed
                 // locally. Additional claimed buffers refill the local cache.
-                entry = Some(cache_entry);
+                // SAFETY: `retain_banked_ptr` above retained the strong reference
+                // transferred to the returned lease.
+                let lease = unsafe { SizeClassLease::from_banked_ptr(self.class) };
+                entry = Some((cache_entry, lease));
             } else {
                 // The take count is derived from the target occupancy, so refill
                 // cannot overflow the local cache. Push directly to avoid the
@@ -807,6 +826,7 @@ impl TlsSizeClassCache {
         let entry = TlsSizeClassCacheEntry { buffer, slot };
 
         if self.len < self.capacity {
+            debug_assert_eq!(self.class, class.ptr);
             class.into_banked();
             self.push_local(entry);
             return;
@@ -856,11 +876,12 @@ impl TlsSizeClassCache {
             spilled.push((spilled_entry.slot, spilled_entry.buffer));
         }
         // SAFETY: each spilled entry represents one banked class reference
-        // owned by this cache. `self.len` was non-zero, so those references keep
-        // `self.class` live.
+        // owned by this cache. `self.class` came from `Arc::into_raw`, and the
+        // spilled references keep the class live for the batch release.
         unsafe { SizeClassLease::from_banked_ptr(self.class) }.return_global_batch(spilled, spill);
 
         debug_assert!(self.len < self.capacity);
+        debug_assert_eq!(self.class, class.ptr);
         class.into_banked();
         self.push_local(entry);
     }
@@ -878,8 +899,8 @@ impl Drop for TlsSizeClassCache {
             entries.push((entry.slot, entry.buffer));
         }
         // SAFETY: each initialized entry represented one banked class
-        // reference owned by this cache. Those references keep `self.class`
-        // live for the duration of this drop.
+        // reference owned by this cache. `self.class` came from `Arc::into_raw`,
+        // and those references keep the class live for this drop.
         unsafe { SizeClassLease::from_banked_ptr(self.class) }.return_global_batch(entries, count);
     }
 }
@@ -1062,12 +1083,8 @@ impl BufferPoolThreadCache {
         match Self::cache(class) {
             Some(mut cache) => {
                 // SAFETY: `cache` points to this thread's initialized TLS cache.
-                unsafe { cache.as_mut().pop(class) }.map(|entry| {
-                    // SAFETY: `TlsSizeClassCache::pop` returns entries that own
-                    // one banked class reference for `class`.
-                    let lease = unsafe { SizeClassLease::from_banked(class) };
-                    (entry.buffer, lease, entry.slot)
-                })
+                unsafe { cache.as_mut().pop(class) }
+                    .map(|(entry, lease)| (entry.buffer, lease, entry.slot))
             }
             None => class
                 .global
@@ -2191,6 +2208,48 @@ mod tests {
         assert_eq!(get_global_len(class), 0);
     }
 
+    #[test]
+    fn test_size_class_leases_use_raw_arc_tokens_across_cache_paths() {
+        // This is primarily a Miri regression for the manual `Arc` refcount
+        // operations used when buffers move between checked-out, TLS, and
+        // global states.
+        let pool = test_pool(test_config(page, page, 8).with_thread_cache_capacity(NZUsize!(4)));
+        let class_index = pool
+            .class_index(page)
+            .expect("class exists for page-sized buffer");
+        let class = pool.inner.classes[class_index].clone();
+        let base_strong_count = Arc::strong_count(&class);
+
+        let mut bufs = Vec::new();
+        for _ in 0..5 {
+            bufs.push(pool.try_alloc(page).expect("tracked allocation"));
+        }
+        assert_eq!(Arc::strong_count(&class), base_strong_count + 5);
+
+        for buf in bufs {
+            drop(buf);
+        }
+        assert_eq!(get_local_len(&class), 3);
+        assert_eq!(get_global_len(&class), 2);
+        assert_eq!(Arc::strong_count(&class), base_strong_count + 3);
+
+        BufferPoolThreadCache::flush();
+        assert_eq!(get_local_len(&class), 0);
+        assert_eq!(get_global_len(&class), 5);
+        assert_eq!(Arc::strong_count(&class), base_strong_count);
+
+        let first = pool.try_alloc(page).expect("global refill");
+        assert_eq!(Arc::strong_count(&class), base_strong_count + 2);
+        let second = pool.try_alloc(page).expect("local refill reuse");
+        assert_eq!(Arc::strong_count(&class), base_strong_count + 2);
+
+        drop(first);
+        drop(second);
+        BufferPoolThreadCache::flush();
+        assert_eq!(Arc::strong_count(&class), base_strong_count);
+    }
+
     #[test]
     fn test_global_batch_alloc_stops_when_global_runs_empty() {
         let class = test_size_class(64, 64);
@@ -2213,14 +2272,13 @@ mod tests {
     fn test_tls_size_class_cache_push_tolerates_empty_spill() {
         let class = test_size_class(64, 64);
         let (slot, buffer) = class.global.try_create(false).expect("slot reservation");
-        SizeClassLease::retain_banked(&class);
+        let lease = SizeClassLease::retain(&class);
         let mut cache = TlsSizeClassCache::new(&class, 0);
 
-        // Small local capacities should bypass batching and push straight to global.
-        // SAFETY: the retained reference above is represented by this lease and
+        // Small local capacities should bypass batching and push straight to
+        // global. The retained reference above is represented by this lease and
         // transferred into `cache.push`.
-        let class = unsafe { SizeClassLease::from_banked(&class) };
-        cache.push(class, slot, buffer);
+        cache.push(lease, slot, buffer);
         assert_eq!(cache.len, 0);
         drop(cache);
     }

```

patrick-ogrady: not sure how this affects performance...

andresilva: I think it will tank because it's cloning an Arc on the hot path. I'm also not sure the premise there is correct. Will investigate.

andresilva: Should be fixed now.

runtime/src/iobuf/buffer.rs
```diff
@@ -112,13 +113,18 @@ impl Drop for AlignedBuffer {
...
 /// allocation pointer. The size-class freelist stores the allocation layout, so
-/// moving pooled buffers through checked-out values and thread-local caches
+/// moving pooled buffers through pooled backing values and thread-local caches
 /// does not need to move a per-buffer [`Layout`].
 ///
 /// `PooledBuffer` does not implement [`Drop`] and has no reference to its
-/// originating [`SizeClass`]. It must be paired with pool metadata that returns
-/// it to the size class, or explicitly deallocated with the exact layout used
-/// to create it.
+/// originating [`SizeClass`](super::pool::SizeClass). In normal pool use it is
```
patrick-ogrady: english in this sentence sounds off?

runtime/src/iobuf/buffer.rs
```diff
@@ -112,13 +113,18 @@ impl Drop for AlignedBuffer {
...
+/// owned by [`PooledBacking`] while outside the global freelist and by the
+/// size-class [`super::freelist::Freelist`] while globally free. Only the
+/// freelist knows the layout needed to deallocate it.
+///
+/// Code that takes a `PooledBuffer` out of pool state must either return it to
```
patrick-ogrady: *Callers?

runtime/src/iobuf/buffer.rs
andresilva: ESL 🥹

patrick-ogrady (approved):

patrick-ogrady:
today, we go fast.

PR #3769, opened 2026-05-12
[storage/qmdb] specialize the (common) k=2 case in apply_batch

roberto-bayardo:
While the gains are modest, I expect this will be a quite common if not typical case, so probably worth it.

| Benchmark | Time | Change vs without k=2 |
  |-----------|------|-----------------------|
  | `apply_batch/uncomm_ancestor` | 2.19 ms | **-9.2%** |
  | `apply_batch/comm_uncomm_chain` | 3.10 ms | **-4.8%** |


roberto-bayardo:
bugbot run

danlaine (approved):

PR #3778, opened 2026-05-13
[storage/qmdb/sync] Avoid requiring a trusted ops root in current qmdb sync

### LINKED ISSUE #3788: [storage/qmdb/current] Current QMDB sync should not require 2 trusted roots
Current QMDB sync protocol involves trusting both the canonical root and ops root are correct.  It should instead only require a trusted canonical root; the ops root should be verified against the canonical root.

roberto-bayardo:
- Add a `current::sync` wrapper path that verifies each `OpsRootWitness`, forwards authenticated ops-root targets to the shared sync engine, plus the database root, and verifies the database root derived from sync matches.
- Update the sync example to support moving `current` sync targets without trusting server-provided ops roots directly.
- Add regression coverage for `current` target updates and simplify duplicated sync target tests.

Resolves: https://github.com/commonwarexyz/monorepo/issues/3788

### DIFF
--- a/examples/sync/src/bin/client.rs
+++ b/examples/sync/src/bin/client.rs
@@ -217,15 +217,14 @@ where
 
 /// Repeatedly sync a Current database to the server's state.
 ///
-/// Uses the canonical-root-aware [current::sync::sync] wrapper. The initial target includes
-/// a canonical root and [OpsRootWitness]; the wrapper verifies the witness before starting
-/// the engine and checks the canonical root after completion. Target updates during sync
-/// use verified ops-root targets.
+/// Uses the `current::sync::sync` wrapper. The wrapper verifies each target's `OpsRootWitness`
+/// before forwarding its ops root to the shared sync engine, then checks the canonical root for the
+/// target the engine finishes on.
 async fn run_current<E>(context: E, config: Config) -> Result<(), Box<dyn std::error::Error>>
 where
     E: BufferPooler + Storage + Clock + Metrics + Network + Spawner,
 {
-    use commonware_storage::qmdb::{self, current as current_qmdb};
+    use commonware_storage::qmdb::current as current_qmdb;
 
     info!("starting Current database sync process");
     let mut iteration = 0u32;
@@ -242,68 +241,58 @@ where
             "received current sync target"
         );
 
-        let hasher = qmdb::hasher::<commonware_sync::Hasher>();
-        let initial_ops_target = initial_target
-            .to_engine_target(&hasher)
-            .ok_or("initial ops root witness verification failed")?;
-
         let (update_sender, update_receiver) = mpsc::channel(UPDATE_CHANNEL_SIZE);
 
         let target_update_handle = {
             let resolver = resolver.clone();
-            let initial_ops_target = initial_ops_target.clone();
+            let mut current_target_root = initial_target.canonical_root;
             let target_update_interval = config.target_update_interval;
-            context.child("target_update").spawn(move |context| async move {
-                let hasher = qmdb::hasher::<commonware_sync::Hasher>();
-                let mut current_ops_target = initial_ops_target;
-                loop {
-                    context.sleep(target_update_interval).await;
-                    match resolver.get_current_sync_target().await {
-                        Ok(new_target) => {
-                            let Some(new_ops_target) = new_target.to_engine_target(&hasher) else {
-                                warn!("target update witness verification failed, skipping");
-                                continue;
-                            };
-                            if current_ops_target != new_ops_target {
-                                match update_sender.clone().try_send(new_ops_target.clone()) {
-                                    Ok(()) => {
-                                        info!("target updated");
-                                        current_ops_target = new_ops_target;
-                                    }
-                                    Err(mpsc::error::TrySendError::Closed(_)) => return Ok(()),
-                                    Err(err) => {
-                                        warn!(?err, "failed to send target update");
-                                        return Err(Error::TargetUpdateChannel {
-                                            reason: err.to_string(),
-                                        });
+            context
+                .child("target_update")
+                .spawn(move |context| async move {
+                    loop {
+                        context.sleep(target_update_interval).await;
+                        match resolver.get_current_sync_target().await {
+                            Ok(new_target) => {
+                                if current_target_root != new_target.canonical_root {
+                                    let new_root = new_target.canonical_root;
+                                    match update_sender.clone().try_send(new_target) {
+                                        Ok(()) => {
+                                            info!("target updated");
+                                            current_target_root = new_root;
+                                        }
+                                        Err(mpsc::error::TrySendError::Closed(_)) => return Ok(()),
+                                        Err(err) => {
+                                            warn!(?err, "failed to send target update");
+                                            return Err(Error::TargetUpdateChannel {
+                                                reason: err.to_string(),
+                                            });
+                                        }
                                     }
                                 }
                             }
-                        }
-                        Err(err) => {
-                            warn!(?err, "failed to get sync target from server");
+                            Err(err) => {
+                                warn!(?err, "failed to get sync target from server");
+                            }
                         }
                     }
-                }
-            })
+                })
         };
 
         let db_config = current::create_config(&context);
-        let database: current::Database<_> = current_qmdb::sync::sync(
-            current_qmdb::sync::Config {
-                context: context.child("sync"),
-                resolver,
-                target: initial_target,
-                max_outstanding_requests: config.max_outstanding_requests,
-                fetch_batch_size: config.batch_size,
-                apply_batch_size: 1024,
-                db_config,
-                update_rx: Some(update_receiver),
-                finish_rx: None,
-                reached_target_tx: None,
-                max_retained_roots: 8,
-            },
-        )
+        let database: current::Database<_> = current_qmdb::sync::sync(current_qmdb::sync::Config {
+            context: context.child("sync"),
+            resolver,
+            target: initial_target,
+            max_outstanding_requests: config.max_outstanding_requests,
+            fetch_batch_size: config.batch_size,
+            apply_batch_size: 1024,
+            db_config,
+            update_rx: Some(update_receiver),
+            finish_rx: None,
+            reached_target_tx: None,
+            max_retained_roots: 8,
+        })
         .await?;
 
         target_update_handle.abort();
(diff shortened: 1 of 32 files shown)


storage/src/qmdb/current/sync/mod.rs
```diff
@@ -56,29 +62,229 @@ use crate::{
...
+                };
+                if update_tx.send(engine_target.clone()).await.is_err() {
+                    break;
+                }
+                canonical_roots.push((engine_target, current_target.canonical_root));
+            }
+        });
+        let result = match select(engine_fut, forward_fut).await {
+            Either::Left((result, _)) => result?,
+            Either::Right(((), engine_fut)) => engine_fut.await?,
```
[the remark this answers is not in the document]
roberto-bayardo: fixed

roberto-bayardo:
bugbot run

storage/src/qmdb/current/sync/mod.rs
```diff
@@ -56,29 +61,233 @@ use crate::{
...
+            }
+            Ok(())
+        });
+        let result = match select(engine_fut, forward_fut).await {
+            Either::Left((result, _)) => result?,
+            Either::Right((forward_result, engine_fut)) => {
+                forward_result?;
+                engine_fut.await?
+            }
+        };
```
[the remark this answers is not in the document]
roberto-bayardo: fixed

storage/src/qmdb/current/sync/mod.rs
```diff
@@ -56,29 +61,234 @@ use crate::{
...
+        let ops_root = D::read(buf)?;
+        let witness = OpsRootWitness::<F, D>::read(buf)?;
+        let range = NonEmptyRange::<Location<F>>::read(buf)?;
+        Ok(Self {
+            canonical_root,
+            ops_root,
+            witness,
+            range,
+        })
+    }
```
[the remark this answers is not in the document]
roberto-bayardo: fixed

clabby (dismissed):

examples/sync/src/bin/client.rs
```diff
@@ -217,44 +217,96 @@ where
...
-                ops_root = %database.ops_root(),
-                sync_interval = ?config.sync_interval,
-                "Current sync completed successfully"
-            );
-            Ok(database)
-        },
-        "Current database",
-    )
-    .await
+    use commonware_storage::qmdb::current as current_qmdb;
```
patrick-ogrady: We should import at top of mod?

storage/src/qmdb/sync/database.rs
```diff
@@ -73,6 +73,17 @@ pub trait Database: Sized + Send {
...
+    ///
+    /// For databases where the public root differs from the ops root used for streaming
+    /// verification (e.g., `current`), this returns the **ops root**.
     fn root(&self) -> Self::Digest;
+
+    /// Returns the database's canonical (public) root digest.
+    ///
+    /// For most databases, the canonical root equals the ops root. For `current` databases,
+    /// the canonical root additionally commits to the activity bitmap and grafted tree state.
+    fn canonical_root(&self) -> Self::Digest {
```
patrick-ogrady: I find the naming between root <> canonical root not distinct enough.

If anything, i would've thought `canonical root` was `root` and `root` was something else?

storage/src/qmdb/sync/error.rs
```diff
@@ -53,6 +53,12 @@ pub enum EngineError<F: Family, D: Digest> {
     /// Error extracting pinned nodes
     #[error("error extracting pinned nodes: {0}")]
     PinnedNodes(String),
+    /// Ops root witness failed verification against the trusted canonical root
+    #[error("ops root witness failed verification")]
+    OpsRootWitnessInvalid,
```
patrick-ogrady, line 56: We could align this with the methods? Except, we could call drop `Canonical` for the actual root?

And then use `ops_root` for the other?

storage/src/qmdb/sync/database.rs
patrick-ogrady: I think we should avoid a default here (require each db to do explicit)?

storage/src/qmdb/current/sync/mod.rs
```diff
@@ -56,29 +61,240 @@ use crate::{
...
+///
+/// The witness authenticates `ops_root` against `canonical_root`; the shared sync engine
+/// uses the authenticated ops root as its target.
+#[derive(Clone, Debug)]
+pub struct Target<F: Graftable, D: Digest> {
+    /// The trusted canonical root.
+    pub canonical_root: D,
+    /// The ops root provided by the sync source.
+    pub ops_root: D,
+    /// Witness proving that `canonical_root` commits to `ops_root`.
```
patrick-ogrady: could be `Witness proving the `root` commits to `ops_root`?

storage/src/qmdb/sync/error.rs
roberto-bayardo: Dropped "trusted canonical" and just left it as "root". (e.g. the database's overall root commitment, which is the single trusted source of truth).

storage/src/qmdb/current/sync/mod.rs
roberto-bayardo: done

storage/src/qmdb/sync/database.rs
roberto-bayardo: changed to just "root"  everywhere I mean "overall database root commitment", and ops_root when we're explicitly talking about just the ops-log root.

roberto-bayardo: and removed default.

storage/src/qmdb/sync/target.rs
```diff
@@ -13,8 +13,10 @@ use commonware_utils::range::NonEmptyRange;
 /// them.
 #[derive(Debug)]
 pub struct Target<F: Family, D: Digest> {
-    /// The ops root the sync engine verifies streaming batches against.
+    /// The database root expected after sync completes.
     pub root: D,
+    /// The ops root the sync engine verifies streaming batches against.
+    pub ops_root: D,
```
patrick-ogrady: I suppose this starts to run into the same issue as the empty hash we discussed before where it isn't necessary for some types.

storage/src/qmdb/sync/engine.rs
```diff
@@ -820,7 +820,6 @@ where
             )
             .await?;
 
-            // Verify the final root digest matches the final target
```
patrick-ogrady, line 819: retain this?

storage/fuzz/fuzz_targets/qmdb_any_fixed_sync.rs
```diff
@@ -131,7 +131,7 @@ where
     >,
 {
     let db_config = test_config(test_name, &context);
-    let expected_root = target.root;
+    let expected_root = target.ops_root;
```
patrick-ogrady: we could keep this as root?

storage/src/qmdb/any/sync/mod.rs
```diff
@@ -83,7 +83,7 @@ where
     matches!(
         peek,
         Ok(Some((_, journal_leaves, root)))
-            if journal_leaves == target.range.end() && root == target.root
+            if journal_leaves == target.range.end() && root == target.ops_root
```
patrick-ogrady: I think we can keep as root?

storage/src/qmdb/sync/database.rs
```diff
@@ -73,17 +73,9 @@ pub trait Database: Sized + Send {
         async { false }
     }
 
-    /// Get the root digest of the database for verification.
-    ///
-    /// For databases where the public root differs from the ops root used for streaming
-    /// verification (e.g., `current`), this returns the **ops root**.
-    fn root(&self) -> Self::Digest;
+    /// Get the ops root digest used for streaming verification.
```
patrick-ogrady: I think we should refer to this as the `root over the log`? or something like that rather than discussing what it is used for.

storage/src/qmdb/keyless/sync/tests.rs
```diff
@@ -99,6 +99,7 @@ where
             context: context.child("client"),
             target: Target {
```
patrick-ogrady: I think this target could actually stay unmodified and we should know to use root for syncing inside of the sync mod?

Could be confused about layering tho

storage/src/qmdb/sync/target.rs
roberto-bayardo: not sure I follow?

storage/fuzz/fuzz_targets/qmdb_any_fixed_sync.rs
roberto-bayardo: yes fixed

storage/src/qmdb/any/sync/mod.rs
roberto-bayardo: This one should be ops_root, as it's comparing an ops root derived from the ops range alone, not from the db.  I've renamed root there to "ops_root" and updated the function doc to clarify.

storage/src/qmdb/sync/engine.rs
roberto-bayardo: done

storage/src/qmdb/sync/database.rs
roberto-bayardo: changed to "/// Get the root of the operation log for range verification."

storage/src/qmdb/sync/target.rs
patrick-ogrady: This one: https://github.com/commonwarexyz/monorepo/pull/3743

Where we didn't want some types around for some families (because they weren't necessary).

roberto-bayardo: Ah I see.  Yeah we could pursue a marker type here again I suppose, though in this case I think both fields are still meaningful for all dbs, just so happens for most dbs they are the same value.

storage/src/qmdb/keyless/sync/tests.rs
```diff
@@ -99,6 +99,7 @@ where
...
                 root: sha256::Digest::from([0; 32]),
+                ops_root: sha256::Digest::from([0; 32]),
```
patrick-ogrady: Once way we could simplify this (avoid unforced errors) is to have families have a target constructor? any would only accept 1 root and current would accept 2 roots?

roberto-bayardo: You mean associated with each QMDB type?   I think the challenge is they are each super generic... how about just  from_root vs from_roots constructors on Target itself?

patrick-ogrady (approved):

PR #3779, opened 2026-05-14
[ci] Only install `cargo-dylint` and `dylint-link` when not present

clabby:
## Overview

Only installs `cargo-dylint` and `dylint-link` if they are not already installed. By default, `swatinem/rust-cache` sets `cache-bin` to `true`, so the binaries do not need to be rebuilt each run.

When `VERSION` changes, the workflow will detect, re-build, and overwrite the binaries.


clabby:
Closing, since `cache-bin` is broken. #3781

PR #3780, opened 2026-05-14
[ci] Add dependency cooldown check

clabby:
## Overview

Adds a dependency cooldown check to CI that delays the adoption of any newly-released crate versions by 7 days.


patrick-ogrady:
Very cool. Didn't know this would be so easy.

andresilva (approved):

PR #3785, opened 2026-05-14
[consensus] Sync Reporter

patrick-ogrady:


consensus/src/marshal/standard/mod.rs
```diff
@@ -2071,7 +2064,6 @@ mod tests {
             }
 
             actor_handle.abort();
-            let _ = release.send_lossy(());
```
patrick-ogrady, line 2074: Is this still a useful test?

consensus/src/marshal/core/mailbox.rs
patrick-ogrady: there are a good number of functions in this file that no longer need to by async.

examples/reshare/src/engine.rs
```diff
@@ -148,7 +148,7 @@ where
             dkg::Config {
                 manager: config.manager.clone(),
                 signer: config.signer.clone(),
-                mailbox_size: MAILBOX_SIZE,
+                mailbox_size: NZUsize!(MAILBOX_SIZE),
```
patrick-ogrady: make mailbox NZ?

examples/log/src/application/reporter.rs
```diff
@@ -1,39 +1,46 @@
...
-            }
-            Activity::Finalization(finalization) => {
-                info!(%view, payload = ?finalization.proposal.payload, "finalized");
-            }
-            Activity::Nullification(_) => {
-                info!(%view, "nullified");
-            }
-            _ => {}
+    fn report(&mut self, activity: Self::Activity) -> Feedback {
+        self.sender.enqueue(Message::Report { activity })
```
patrick-ogrady: we can just log inline here?

examples/log/src/application/ingress.rs
```diff
@@ -17,16 +23,26 @@ pub enum Message<D: Digest> {
     Verify {
         response: oneshot::Sender<bool>,
     },
+    Report {
```
patrick-ogrady: remove this

examples/log/src/application/actor.rs
```diff
@@ -83,6 +85,9 @@ impl<R: Rng + Spawner, H: Hasher> Application<R, H> {
                     // the parent included in the payload matches the provided context.
                     let _ = response.send(true);
                 }
+                Message::Report { activity } => {
```
patrick-ogrady: remove this

consensus/src/reporter.rs
```diff
@@ -11,11 +11,11 @@ use std::marker::PhantomData;
 impl<A: Send, R: Reporter<Activity = A>> Reporter for Option<R> {
     type Activity = A;
 
-    async fn report(&mut self, activity: Self::Activity) {
+    fn report(&mut self, activity: Self::Activity) -> Feedback {
         let Some(reporter) = self else {
-            return;
+            return Feedback::Ok;
```
patrick-ogrady, line 16: dropped?

consensus/src/marshal/coding/mod.rs
```diff
@@ -1027,10 +1039,7 @@ mod tests {
 
             // Subscribe through the core actor. This internally subscribes to the
             // coding shard buffer and registers local waiters.
-            let block_rx = marshal
-                .subscribe_by_commitment(Some(round), missing_commitment)
-                .await;
-
+            let block_rx = marshal.subscribe_by_commitment(Some(round), missing_commitment);
```
patrick-ogrady, line 1024: Add \n afer this?

consensus/src/marshal/core/mailbox.rs
```diff
@@ -363,48 +549,294 @@ impl<S: Scheme, V: Variant> Mailbox<S, V> {
...
     }
 }
 
 impl<S: Scheme, V: Variant> BlockProvider for Mailbox<S, V> {
     type Block = V::ApplicationBlock;
 
-    async fn fetch_block(self, digest: <V::Block as Digestible>::Digest) -> Option<Self::Block> {
-        let subscription = self.subscribe_by_digest(None, digest).await;
-        subscription.await.ok().map(V::into_inner)
+    fn fetch_block(
```
patrick-ogrady: make this take a oneshot?

consensus/src/marshal/coding/shards/mailbox.rs
```diff
@@ -171,41 +186,35 @@ where
...
             response: responder,
         };
-        self.sender.send_lossy(msg).await;
+        let _ = self.sender.enqueue(msg);
         receiver
     }
 
     /// Request to prune all caches at and below the given commitment.
-    pub async fn prune(&self, through: Commitment) {
+    pub fn prune(&self, through: Commitment) {
```
patrick-ogrady, line 282: With no way to compute the range implied by this, I think its easier to just leave as-is for now.

consensus/src/marshal/coding/shards/engine.rs
```diff
@@ -285,7 +286,7 @@ where
     context: ContextCell<E>,
 
     /// Receiver for incoming messages to the actor.
```
patrick-ogrady, line 286: make this normal mailbox

consensus/src/marshal/coding/variant.rs
patrick-ogrady, line 1: Remove async here?

consensus/src/marshal/coding/marshaled.rs
```diff
@@ -570,7 +569,12 @@ where
             // parent and cannot be broadcast under the current header, so
             // drop the receiver and let the voter nullify the view via
             // timeout.
-            if let Some(block) = marshal.get_verified(consensus_context.round).await {
+            if let Some(block) = marshal
+                .get_verified(consensus_context.round)
+                .await
```
patrick-ogrady: We may still want to return futures here (assuming they can then be placed in futures pools as needed by callers?)

consensus/src/marshal/core/mailbox.rs
```diff
@@ -363,48 +574,392 @@ impl<S: Scheme, V: Variant> Mailbox<S, V> {
...
 }
 
 impl<S: Scheme, V: Variant> BlockProvider for Mailbox<S, V> {
     type Block = V::ApplicationBlock;
 
-    async fn fetch_block(self, digest: <V::Block as Digestible>::Digest) -> Option<Self::Block> {
-        let subscription = self.subscribe_by_digest(None, digest).await;
-        subscription.await.ok().map(V::into_inner)
+    #[allow(clippy::manual_async_fn)]
+    fn fetch_block(
```
patrick-ogrady: revert the changes here.

consensus/src/marshal/core/mailbox.rs
```diff
@@ -363,48 +574,392 @@ impl<S: Scheme, V: Variant> Mailbox<S, V> {
...
 
-    async fn report(&mut self, activity: Self::Activity) {
+    fn report(&mut self, activity: Self::Activity) -> Feedback {
         let message = match activity {
             Activity::Notarization(notarization) => Message::Notarization { notarization },
             Activity::Finalization(finalization) => Message::Finalization { finalization },
             _ => {
                 // Ignore other activity types
-                return;
+                return Feedback::Ok;
```
patrick-ogrady: make dropped?

consensus/src/marshal/standard/deferred.rs
```diff
@@ -1208,6 +1210,7 @@ mod tests {
                 .await
                 .expect("verify should reach application before marshal abort");
             marshal_actor_handle.abort();
+            let _ = marshal_actor_handle.await;
```
patrick-ogrady, line 1218: is this necessary?

consensus/src/marshal/core/mailbox.rs
```diff
@@ -163,15 +168,230 @@ pub(crate) enum Message<S: Scheme, V: Variant> {
...
+    hints: BTreeMap<Height, NonEmptyVec<S::PublicKey>>,
+    messages: VecDeque<Message<S, V>>,
+}
+
+impl<S: Scheme, V: Variant> Default for Pending<S, V> {
+    fn default() -> Self {
+        Self {
+            floor: None,
+            prune: None,
+            hints: BTreeMap::new(),
```
patrick-ogrady, line 244: we shouldn't give "hints" special treatment (i.e. draining before other FIFO messages)

consensus/src/marshal/mocks/harness.rs
```diff
@@ -4543,9 +4518,7 @@ pub fn hint_finalized_triggers_fetch<H: TestHarness>() {
         // Validator 1: hint that block 5 is finalized, targeting validator 0
         handle1
             .mailbox
-            .hint_finalized(Height::new(5), NonEmptyVec::new(participants[0].clone()))
-            .await;
-
+            .hint_finalized(Height::new(5), NonEmptyVec::new(participants[0].clone()));
```
patrick-ogrady, line 4522: restore \n

consensus/src/marshal/mocks/harness.rs
```diff
@@ -3902,27 +3893,11 @@ pub fn subscribe_blocks_from_different_sources<H: TestHarness>() {
...
-        let sub5_rx = handle
-            .mailbox
-            .subscribe_by_digest(None, H::digest(&block5))
-            .await;
-
+        let sub1_rx = handle.mailbox.subscribe_by_digest(None, H::digest(&block1));
+        let sub2_rx = handle.mailbox.subscribe_by_digest(None, H::digest(&block2));
+        let sub3_rx = handle.mailbox.subscribe_by_digest(None, H::digest(&block3));
+        let sub4_rx = handle.mailbox.subscribe_by_digest(None, H::digest(&block4));
+        let sub5_rx = handle.mailbox.subscribe_by_digest(None, H::digest(&block5));
```
patrick-ogrady, line 3900: add \n back

consensus/src/marshal/core/mailbox.rs
patrick-ogrady, line 1: when adding, check subscriptions of alive items and drop anything that are no longer servable

consensus/src/marshal/core/mailbox.rs
```diff
@@ -118,7 +120,7 @@ pub(crate) enum Message<S: Scheme, V: Variant> {
         /// The verified block.
         block: V::Block,
         /// A channel signaled once the block is durably stored.
-        ack: oneshot::Sender<()>,
+        ack: Option<oneshot::Sender<()>>,
```
patrick-ogrady, line 123: Once VecDeque::extract_if is stable, this will be much cleaner.

patrick-ogrady:
bugbot run

actor/src/benches/mailbox.rs
```diff
@@ -50,12 +50,11 @@ impl Message {
 impl mailbox::Policy for Message {
     type Overflow = VecDeque<Self>;
 
-    fn handle(overflow: &mut VecDeque<Self>, message: Self) -> bool {
```
patrick-ogrady, line 53: Opted to drop `bool` here because Drop/Backoff was confusing to actually use (as the mailboxes got smarter, it wasn't clear what should be returned or what the sender should do about it).

actor/src/benches/mailbox.rs
```diff
@@ -216,7 +214,7 @@ fn bench_overflow_drop(c: &mut Criterion) {
             |(sender, _receiver)| {
                 for _ in 0..MESSAGES {
                     let result = sender.enqueue(black_box(Message::drop()));
```
BrendanChou, line 216: change the name of this to `Message::backoff()`?

actor/src/lib.rs
```diff
@@ -15,13 +15,18 @@ commonware_macros::stability_scope!(BETA {
     pub enum Feedback {
         /// The work was accepted within the configured capacity.
         Ok,
-        /// The submission was handled but requests sender backoff.
+        /// The submission exceeded the configured capacity and requests sender backoff.
         Backoff,
-        /// The work was dropped.
-        Dropped,
```
cronokirby, line 21: Why did we get rid of dropped?

patrick-ogrady: image

patrick-ogrady:
Adding `drain` to mailbox drop, then we should be g2g: 
```
  - [P1] consensus/src/marshal/core/mailbox.rs:543 accepted one-shot waits can still hang on shutdown. If
    enqueue accepts a request or durability ack and the marshal receiver is dropped before it processes the
    queued message, actor::mailbox marks the receiver closed but leaves buffered messages owned by shared state
    while any Mailbox clone exists. The waiting future itself holds such a clone, so the responder is never
    dropped and receiver.await can park indefinitely. This affects proposed/verified/certified and the request
    methods with response channels.
```

broadcast/fuzz/fuzz_targets/broadcast_engine_operations.rs
```diff
@@ -249,7 +249,7 @@ fn fuzz(input: FuzzInput) {
 
                     if let Some(mailbox) = mailboxes.get(&peer).cloned() {
                         let resolved_recipients = resolve_recipients(&recipients, &peers);
-                        drop(mailbox.broadcast(resolved_recipients, message).await);
+                        let _ = mailbox.broadcast(resolved_recipients, message);
```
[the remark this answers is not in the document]
patrick-ogrady: This is not exactly true (I think)? it was awaiting the registration of the oneshot and that now happens async.

consensus/src/marshal/core/mailbox.rs
```diff
@@ -163,15 +165,248 @@ pub(crate) enum Message<S: Scheme, V: Variant> {
...
+        }
+    }
+
+    pub(crate) fn response_closed(&self) -> bool {
+        match self {
+            Self::GetInfo { response, .. } => response.is_closed(),
+            Self::GetBlock { response, .. } => response.is_closed(),
+            Self::GetFinalization { response, .. } => response.is_closed(),
+            Self::SubscribeByDigest { response, .. }
+            | Self::SubscribeByCommitment { response, .. } => response.is_closed(),
```
cronokirby, line 213: Why have a separate branch for just these two?

consensus/src/marshal/core/mailbox.rs
```diff
@@ -163,15 +165,248 @@ pub(crate) enum Message<S: Scheme, V: Variant> {
...
+    }
+}
+
+impl<S: Scheme, V: Variant> Pending<S, V> {
+    // The effective floor for staleness checks is the max of both pending advances
+    fn height(&self) -> Option<Height> {
+        self.floor.into_iter().chain(self.prune).max()
+    }
+
+    fn retain_message(message: &Message<S, V>, height: Option<Height>) -> bool {
```
cronokirby: This function is only ever called with `self.height()` so I would just keep `retain_useful`

consensus/src/marshal/standard/variant.rs
```diff
@@ -71,21 +71,21 @@ where
...
 
-    async fn finalized(&self, _commitment: B::Digest) {
+    fn finalized(&self, _commitment: B::Digest) {
         // No cleanup needed in standard mode - the buffer handles its own pruning
     }
 
-    async fn send(&self, _round: Round, block: B, recipients: Recipients<K>) {
-        let _peers = Broadcaster::broadcast(self, recipients, block).await;
+    fn send(&self, _round: Round, block: B, recipients: Recipients<K>) {
+        let _ = Broadcaster::broadcast(self, recipients, block);
```
BrendanChou: `let _ =` can be deleted now that there is no await?

consensus/src/marshal/core/mailbox.rs
```diff
@@ -163,15 +165,248 @@ pub(crate) enum Message<S: Scheme, V: Variant> {
...
+        let height = self.height();
+        self.messages
+            .retain(|message| Self::retain_message(message, height));
+    }
+
+    fn set_floor(&mut self, height: Height) -> bool {
+        // Floor only moves up: ignore a regression
+        if self.floor.is_none_or(|current| height > current) {
+            self.floor = Some(height);
+        }
```
cronokirby:
```suggestion
        self.floor = self.floor.max(Some(height));
```

consensus/src/marshal/core/mailbox.rs
```diff
@@ -163,15 +165,248 @@ pub(crate) enum Message<S: Scheme, V: Variant> {
...
+        self.retain_useful();
+        true
```
cronokirby: What's the point of having this function return bool?

consensus/src/marshal/coding/shards/mailbox.rs
```diff
@@ -88,6 +90,45 @@ where
...
+    pub(crate) fn response_closed(&self) -> bool {
+        match self {
+            Self::GetByCommitment { response, .. } | Self::GetByDigest { response, .. } => {
+                response.is_closed()
+            }
+            Self::SubscribeAssignedShardVerified { response, .. } => response.is_closed(),
+            Self::SubscribeByCommitment { response, .. }
+            | Self::SubscribeByDigest { response, .. } => response.is_closed(),
+            _ => false,
+        }
```
BrendanChou, line 109: style: either combine all message types with `response` using `|` or none of them. The first 2 are not combined, then on line 106-107 you combine them in to a single line before calling `is_closed()`

consensus/src/marshal/coding/shards/mailbox.rs
```diff
@@ -88,6 +90,45 @@ where
...
+where
+    B: CertifiableBlock,
+    C: CodingScheme,
+    H: Hasher,
+    P: PublicKey,
+{
+    type Overflow = VecDeque<Self>;
+
+    fn handle(overflow: &mut VecDeque<Self>, message: Self) {
+        overflow.retain(|message| !message.response_closed());
```
BrendanChou: Instead of calling `retain` on every overflow (the actor is slammed and you are asking it to iterate the whole overflow VecDeque), instead prune the items whenever the overflow drains into the ready queue

consensus/src/marshal/core/mailbox.rs
```diff
@@ -163,15 +165,248 @@ pub(crate) enum Message<S: Scheme, V: Variant> {
...
+        // Receiver accepted; the message is consumed
+        let Some(message) = push(message) else {
+            return true;
+        };
+
+        // Receiver rejected; restore so the next drain retries from the same point
+        match message {
+            Message::SetFloor { height } => self.floor = Some(height),
+            Message::Prune { height } => self.prune = Some(height),
+            message => self.messages.push_front(message),
```
cronokirby: Why don't these call self.set_floor like in `handle`? Do we want to ignore the invariant that the floor rises only? (ditto for prune)

consensus/src/marshal/coding/shards/mailbox.rs
```diff
@@ -110,46 +151,41 @@ where
...
+    pub const fn new(sender: Sender<Message<B, C, H, P>>) -> Self {
         Self { sender }
     }
 
     /// Broadcast a proposed erasure coded block's shards to the participants.
-    pub async fn proposed(&self, round: Round, block: CodedBlock<B, C, H>) {
-        let msg = Message::Proposed { block, round };
-        self.sender.send_lossy(msg).await;
+    pub fn proposed(&self, round: Round, block: CodedBlock<B, C, H>) {
+        let _ = self.sender.enqueue(Message::Proposed { block, round });
```
BrendanChou, line 206: I guess you are explicitly ignoring the `Feedback` value but I don't think you need to for linting purposes since it isn't a `Result` (?)

If we do want to have to do this, then `Feedback` should perhaps be a `Result` type instead

(this comment also applies elsewhere in this file)

broadcast/src/buffered/ingress.rs
```diff
@@ -32,14 +34,37 @@ pub enum Message<P: PublicKey, M: Digestible> {
...
+
+    fn handle(overflow: &mut Self::Overflow, message: Self) {
+        overflow.retain(|message| !message.response_closed());
+        if message.response_closed() {
+            return;
+        }
+
+        overflow.push_back(message);
+    }
+}
```
cronokirby, line 87: A subsequent refactor which could be interesting is to have a trait which requires `response_closed` to implement, and then you have a blanket impl of Policy with this. Would shorten things a little bit, and prevent mistakes here.

consensus/src/marshal/core/mailbox.rs
```diff
@@ -163,15 +165,233 @@ pub(crate) enum Message<S: Scheme, V: Variant> {
...
+                identifier: Identifier::Digest(_) | Identifier::Latest,
+                ..
+            }
+            | Self::GetInfo {
+                identifier: Identifier::Digest(_) | Identifier::Latest,
+                ..
+            } => false,
+            _ => false,
+        }
+    }
```
BrendanChou: I wonder if a view range should be provided rather than an optional current/inclusive. Then you can do `last_finalized..` or `last_finalized.next()..` or `..` to provide the valid range

consensus/src/aggregation/mocks/reporter.rs
```diff
@@ -23,12 +27,20 @@ enum Message<S: Scheme, D: Digest> {
     Get(Height, oneshot::Sender<Option<(D, Epoch)>>),
 }
 
+impl<S: Scheme, D: Digest> Policy for Message<S, D> {
+    type Overflow = VecDeque<Self>;
+
+    fn handle(overflow: &mut VecDeque<Self>, message: Self) {
+        overflow.push_back(message);
+    }
+}
```
cronokirby, line 36: I wonder if this might be worth just making the default policy?

consensus/src/marshal/core/mailbox.rs
```diff
@@ -163,15 +165,233 @@ pub(crate) enum Message<S: Scheme, V: Variant> {
...
+    pub(crate) fn response_closed(&self) -> bool {
+        match self {
+            Self::GetInfo { response, .. } => response.is_closed(),
+            Self::GetBlock { response, .. } => response.is_closed(),
+            Self::GetFinalization { response, .. } => response.is_closed(),
+            Self::SubscribeByDigest { response, .. }
+            | Self::SubscribeByCommitment { response, .. } => response.is_closed(),
+            Self::GetVerified { response, .. } => response.is_closed(),
+            _ => false,
+        }
```
BrendanChou, line 223: same style comment as elsewhere: either combine each case using `|` or don't combine any

consensus/src/aggregation/mocks/reporter.rs
cronokirby: I could see that being error-prone, I guess.

consensus/src/marshal/coding/shards/mailbox.rs
```diff
@@ -88,6 +90,45 @@ where
...
+    P: PublicKey,
+{
+    pub(crate) fn response_closed(&self) -> bool {
+        match self {
+            Self::GetByCommitment { response, .. } | Self::GetByDigest { response, .. } => {
+                response.is_closed()
+            }
+            Self::SubscribeAssignedShardVerified { response, .. } => response.is_closed(),
+            Self::SubscribeByCommitment { response, .. }
+            | Self::SubscribeByDigest { response, .. } => response.is_closed(),
```
cronokirby, line 107: Why the batching in this way, versus separate branches...

consensus/src/marshal/core/mailbox.rs
```diff
@@ -163,15 +165,233 @@ pub(crate) enum Message<S: Scheme, V: Variant> {
...
+        true
+    }
+
+    fn prune(&mut self, height: Height) -> bool {
+        if self.prune.is_none_or(|current| height > current) {
+            self.prune = Some(height);
+        }
+        self.retain_useful();
+        true
+    }
```
BrendanChou: move `retain_useful` into their above `if { }` blocks since you should only need to run `retain` if the floor/prune changes I think?

BrendanChou: Also, do these functions need to return a `bool` if they always return true?

consensus/src/marshal/core/mailbox.rs
```diff
@@ -163,15 +165,233 @@ pub(crate) enum Message<S: Scheme, V: Variant> {
...
+                        pending.push(target);
+                    }
+                }
+            }
+            true
+        });
+
+        if let Some(targets) = targets {
+            self.messages
+                .push_back(Message::HintFinalized { height, targets });
```
BrendanChou: In general I think we should not be doing so much work for hint-type messages. Ideally they are stored in the backlog as the latest-hint (overwrite the currently held hint if a new one comes in). It might make sense to promote them to the front of the `Backlog::drain` (if it helps the downstream actor become less clogged), or to have them be held last (if it's just for storage pruning/cleanup but doesn't help the actor move faster)

consensus/src/marshal/core/mailbox.rs
```diff
@@ -163,15 +165,233 @@ pub(crate) enum Message<S: Scheme, V: Variant> {
...
+                overflow.hint_finalized(height, targets);
+            }
+            Self::SetFloor { height } => {
+                overflow.set_floor(height);
+            }
+            Self::Prune { height } => {
+                overflow.prune(height);
+            }
+            message => {
+                overflow.retain_useful();
```
BrendanChou: don't retain useful on every new message (?), only if prune or set_floor are updated

consensus/src/marshal/core/mailbox.rs
```diff
@@ -163,15 +165,248 @@ pub(crate) enum Message<S: Scheme, V: Variant> {
     },
 }
 
+impl<S: Scheme, V: Variant> Message<S, V> {
+    fn stale_below(current: Option<Height>, height: Height, inclusive: bool) -> bool {
```
cronokirby: Might be a bit more work, but taking in a `RangeBounds` might be nicer here, that way you can do like `Self::within(current, ..height)` and `Self::within(current, ..=height)` which is, imo much clearer about what bound you want.

cronokirby: You could also just get rid of this method in favor of like `Some(height) < current` or `Some(height) <= current`, which takes care of the None case correctly. (None will make both of these false, being lower than everything else)

consensus/src/marshal/core/mailbox.rs
```diff
@@ -163,15 +165,248 @@ pub(crate) enum Message<S: Scheme, V: Variant> {
...
+            // Digest and Latest lookups are not bound to a specific height
+            Self::GetBlock {
+                identifier: Identifier::Digest(_) | Identifier::Latest,
+                ..
+            }
+            | Self::GetInfo {
+                identifier: Identifier::Digest(_) | Identifier::Latest,
+                ..
+            } => false,
+            _ => false,
```
cronokirby: If we're going to have a generic fall through case, I think we should group everything we can into it, otherwise we should just have a case for everything.

cronokirby (commented):
Some comments

consensus/src/aggregation/mocks/reporter.rs
patrick-ogrady: Yeah, exactly. I think every policy should be explicit.

consensus/src/marshal/coding/shards/mailbox.rs
patrick-ogrady: All of this is because of slightly different oneshots.

consensus/src/marshal/core/mailbox.rs
patrick-ogrady: Yeah, I agree

consensus/src/marshal/coding/shards/mailbox.rs
patrick-ogrady: This is because of slightly different oneshots.

actor/src/mailbox.rs
```diff
@@ -299,13 +294,9 @@ impl<T: Policy> OverflowState<T> {
...
         // Preserve overflow order, or handle a still-full ready queue.
-        let feedback = if T::handle(&mut queue, message) {
-            Feedback::Backoff
-        } else {
-            Feedback::Dropped
-        };
+        T::handle(&mut queue, message);
         mutation.publish(queue.is_empty());
-        feedback
+        Feedback::Backoff
```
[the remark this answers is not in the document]
patrick-ogrady: This is intended.

actor/src/benches/mailbox.rs
patrick-ogrady: I still think that drop is the right name because we don't add anything to overflow (so just tests handle overhead).

consensus/src/marshal/coding/shards/mailbox.rs
patrick-ogrady: We have `must_use` on enqueue 👍

broadcast/src/buffered/ingress.rs
patrick-ogrady: Trying to hold my back from making too many helpers until we see how more of this plays out.

consensus/src/marshal/coding/shards/mailbox.rs
patrick-ogrady: I was going back and forth on this one. On one hand, it seems like a good idea to drain useless stuff ASAP but it can definitely slow down addition. I think on drain is probably the most coherent option.

consensus/src/marshal/core/mailbox.rs
patrick-ogrady: I didn't want to get too fancy with prioritization outside of prune/set floor (I don't think we should attempt to predict what would speed up the caller -> seems like that will get very complicated very fast)

patrick-ogrady: Found a way to do this more efficently.

broadcast/src/buffered/ingress.rs
```diff
@@ -44,16 +44,48 @@ impl<P: PublicKey, M: Digestible> Message<P, M> {
...
+                continue;
+            }
+
+            let Some(message) = push(message) else {
+                continue;
+            };
+            if !message.response_closed() {
+                self.0.push_front(message);
+                break;
+            }
```
BrendanChou: This is a comment more about the way we call `drain` (rather than your code here) but I still think we should aspire to have it be `fn drain(n: usize)` once the ready queue knows that it has a capacity of at-least `n` messages. This prevents us from having to write this boilerplate in every `Overflow` implementation

consensus/src/marshal/core/mailbox.rs
```diff
@@ -197,157 +191,203 @@ impl<S: Scheme, V: Variant> Message<S, V> {
...
-            self.floor = Some(height);
+    fn prune(&mut self, height: Height) {
+        let prune = Some(height);
+        if self.prune >= prune {
+            return;
         }
-        self.retain_useful();
-        true
+
+        self.prune = self.prune.max(prune);
```
BrendanChou, line 289: `prune()` and `set_floor()` can now be one-liners: `self.x = self.old.max(new)`
edit: no longer applies after latest commit

broadcast/src/buffered/ingress.rs
patrick-ogrady: We drain in the loop over the receiver (as soon as there is exactly one more spot).

We don't yet have recv(x), but suspect we'll add this optimization when we do?

actor/src/mailbox.rs
```diff
@@ -73,11 +72,11 @@ impl<T> Overflow<T> for VecDeque<T> {
     where
         F: FnMut(T) -> Option<T>,
     {
-        while let Some(message) = self.pop_front() {
-            if let Some(message) = push(message) {
-                self.push_front(message);
-                break;
-            }
+        let Some(message) = self.pop_front() else {
```
patrick-ogrady: fixing this, one sec.

consensus/src/marshal/core/actor.rs
```diff
@@ -1295,9 +1318,7 @@ where
 
             let (height, commitment) = (block.height(), V::commitment(&block));
             let (ack, ack_waiter) = A::handle();
-            application
-                .report(Update::Block(V::into_inner(block), ack))
-                .await;
+            application.report(Update::Block(V::into_inner(block), ack));
```
patrick-ogrady, line 1321: The return value of `application.report(Update::Block(...))` is not checked. When the application returns `Feedback::Closed`, the actor still pushes the height onto the pending-ack queue and continues dispatching additional finalized blocks until `max_pending_acks` is saturated. The dropped update is counted as in-flight, so those ack slots are never reclaimed. The actor should stop dispatching immediately on `Closed` (and on `Backoff`) without recording the dropped update as a pending ack.

```patch
diff --git a/consensus/src/marshal/standard/mod.rs b/consensus/src/marshal/standard/mod.rs
index d3a75947b..4b55b574b 100644
--- a/consensus/src/marshal/standard/mod.rs
+++ b/consensus/src/marshal/standard/mod.rs
@@ -1775,6 +1775,39 @@ mod tests {
         }
     }
 
+    /// A reporter that records block delivery attempts but reports that the
+    /// application endpoint is closed.
+    #[derive(Clone)]
+    struct ClosedBlockReporter {
+        blocks: Arc<Mutex<Vec<Height>>>,
+    }
+
+    impl ClosedBlockReporter {
+        fn new() -> Self {
+            Self {
+                blocks: Arc::new(Mutex::new(Vec::new())),
+            }
+        }
+
+        fn blocks(&self) -> Vec<Height> {
+            self.blocks.lock().clone()
+        }
+    }
+
+    impl Reporter for ClosedBlockReporter {
+        type Activity = Update<B>;
+
+        fn report(&mut self, activity: Self::Activity) -> Feedback {
+            match activity {
+                Update::Block(block, _ack) => {
+                    self.blocks.lock().push(block.height());
+                    Feedback::Closed
+                }
+                Update::Tip(_, _, _) => Feedback::Ok,
+            }
+        }
+    }
+
     async fn start_standard_actor<R, Buf>(
         context: deterministic::Context,
         partition_prefix: &str,
@@ -1787,6 +1820,34 @@ mod tests {
         RecordingResolver,
         commonware_runtime::Handle<()>,
     )
+    where
+        R: Reporter<Activity = Update<B>>,
+        Buf: crate::marshal::core::Buffer<Standard<B>, PublicKey = PublicKey> + Clone,
+    {
+        start_standard_actor_with_max_pending_acks(
+            context,
+            partition_prefix,
+            provider,
+            application,
+            buffer,
+            NZUsize!(1),
+        )
+        .await
+    }
+
+    async fn start_standard_actor_with_max_pending_acks<R, Buf>(
+        context: deterministic::Context,
+        partition_prefix: &str,
+        provider: ConstantProvider<S, Epoch>,
+        application: R,
+        buffer: Buf,
+        max_pending_acks: NonZeroUsize,
+    ) -> (
+        Mailbox<S, Standard<B>>,
+        Buf,
+        RecordingResolver,
+        commonware_runtime::Handle<()>,
+    )
     where
         R: Reporter<Activity = Update<B>>,
         Buf: crate::marshal::core::Buffer<Standard<B>, PublicKey = PublicKey> + Clone,
@@ -1797,7 +1858,7 @@ mod tests {
             mailbox_size: NZUsize!(100),
             view_retention_timeout: ViewDelta::new(10),
             max_repair: NZUsize!(10),
-            max_pending_acks: NZUsize!(1),
+            max_pending_acks,
             block_codec_config: (),
             partition_prefix: partition_prefix.to_string(),
             prunable_items_per_section: NZU64!(10),
@@ -2106,6 +2167,94 @@ mod tests {
         });
     }
 
+    #[test_traced("WARN")]
+    fn test_standard_stops_dispatching_blocks_after_application_closes() {
+        let runner = deterministic::Runner::timed(Duration::from_secs(30));
+        runner.start(|mut context| async move {
+            const NUM_BLOCKS: u64 = 3;
+
+            let Fixture {
+                schemes,
+                ..
+            } = bls12381_threshold_vrf::fixture::<V, _>(&mut context, NAMESPACE, NUM_VALIDATORS);
+            let me = participants[0].clone();
+            let partition_prefix = format!("validator-{me}");
+            let epocher = FixedEpocher::new(BLOCKS_PER_EPOCH);
+
+            let application = Application::<B>::manual_ack();
+            let (mut mailbox, _buffer, _resolver, actor_handle) = start_standard_actor(
+                context.child("validator").with_attribute("index", 0),
+                &partition_prefix,
+                ConstantProvider::new(schemes[0].clone()),
+                application.clone(),
+                RecordingBuffer::default(),
+            )
+            .await;
+
+            let mut parent = Sha256::hash(b"");
+            for i in 1..=NUM_BLOCKS {
+                let block = make_raw_block(parent, Height::new(i), i);
+                parent = block.digest();
+                let round = Round::new(
+                    epocher.containing(block.height()).unwrap().epoch(),
+                    View::new(i),
+                );
+                assert!(
+                    mailbox.verified(round, block.clone()).await,
+                    "verified block should persist to the cache"
+                );
+                let finalization = StandardHarness::make_finalization(
+                    Proposal::new(round, View::new(i.saturating_sub(1)), block.digest()),
+                    &schemes,
+                    QUORUM,
+                );
+                StandardHarness::report_finalization(&mut mailbox, finalization).await;
+            }
+
+            select! {
+                finalization = mailbox.get_finalization(Height::new(NUM_BLOCKS)) => {
+                    assert!(finalization.is_some(), "all finalized blocks should be durable");
+                },
+                _ = context.sleep(Duration::from_secs(5)) => {
+                    panic!("finalizations should be stored promptly");
+                },
+            }
+
+            assert_eq!(application.pending_ack_heights(), vec![Height::new(1)]);
+            actor_handle.abort();
+            drop(mailbox);
+            context.sleep(Duration::from_millis(1)).await;
+
+            let reporter = ClosedBlockReporter::new();
+            let (_mailbox, _buffer, _resolver, _actor_handle) =
+                start_standard_actor_with_max_pending_acks(
+                    context
+                        .child("validator_restart")
+                        .with_attribute("index", 0),
+                    &partition_prefix,
+                    ConstantProvider::new(schemes[0].clone()),
+                    reporter.clone(),
+                    RecordingBuffer::default(),
+                    NZUsize!(3),
+                )
+                .await;
+
+            loop {
+                if !reporter.blocks().is_empty() {
+                    break;
+                }
+                context.sleep(Duration::from_millis(10)).await;
+            }
+            context.sleep(Duration::from_millis(1)).await;
+            assert_eq!(
+                reporter.blocks(),
+                vec![Height::new(1)],
+                "marshal should stop dispatching after Closed feedback"
+            );
+        });
+    }
+
     /// Parse the `processed_height` gauge value from a prometheus-encoded
     /// metrics dump produced by `Metrics::encode`. Looks for any line of the
     /// form `<prefix>processed_height <value>` or
```
</details>

consensus/src/marshal/core/mailbox.rs
```diff
@@ -163,15 +165,296 @@ pub(crate) enum Message<S: Scheme, V: Variant> {
...
+            }
+        }
+        if let Some(height) = self.prune.take() {
+            if !self.drain_one(Message::Prune { height }, &mut push) {
+                return;
+            }
+        }
+
+        // Drain the remaining queued messages in FIFO order
+        let height = self.height();
```
patrick-ogrady: `height` is recomputed by examining `self.floor` and `self.prune` at the top of the drain, but both fields are consumed via `.take()` early in the loop. Once those takes have run, any subsequent iteration re-evaluates the same expression and gets `None`, making `message.stale(height)` and `HintFinalized` filtering a no-op for the rest of the queue. Stale `GetInfo`/`GetBlock`/`GetFinalization` requests and `HintFinalized` entries whose heights are below the just-consumed floor/prune can slip through to the actor.

Fix: capture the effective height once (max of the taken floor/prune heights) before the message iteration begins, and use that captured value for all stale checks in the loop.

```patch
diff --git a/consensus/src/marshal/core/mailbox.rs b/consensus/src/marshal/core/mailbox.rs
index 40a79deec..eb2fabee3 100644
--- a/consensus/src/marshal/core/mailbox.rs
+++ b/consensus/src/marshal/core/mailbox.rs
@@ -1096,6 +1096,74 @@ mod tests {
         ));
     }
 
+    #[test]
+    fn policy_drops_stale_pending_requests_when_drain_consumes_floor_and_prune() {
+        let mut overflow = pending();
+        overflow.floor = Some(Height::new(5));
+        overflow.prune = Some(Height::new(7));
+
+        let (get_info_6, _get_info_6_rx) = get_info(6);
+        let (get_block_6, _get_block_6_rx) = get_block(6);
+        let (get_finalization_6, _get_finalization_6_rx) = get_finalization(6);
+        let (get_info_7, _get_info_7_rx) = get_info(7);
+        let (get_block_7, _get_block_7_rx) = get_block(7);
+        let (get_finalization_7, _get_finalization_7_rx) = get_finalization(7);
+
+        overflow
+            .messages
+            .push_back(PendingMessage::Message(get_info_6));
+        overflow
+            .messages
+            .push_back(PendingMessage::Message(get_block_6));
+        overflow
+            .messages
+            .push_back(PendingMessage::Message(get_finalization_6));
+        overflow
+            .hints
+            .insert(Height::new(7), NonEmptyVec::new(public_key(1)));
+        overflow
+            .messages
+            .push_back(PendingMessage::HintFinalized(Height::new(7)));
+        overflow
+            .messages
+            .push_back(PendingMessage::Message(get_info_7));
+        overflow
+            .messages
+            .push_back(PendingMessage::Message(get_block_7));
+        overflow
+            .messages
+            .push_back(PendingMessage::Message(get_finalization_7));
+
+        let drained = drain(&mut overflow);
+        assert_eq!(drained.len(), 5);
+        assert!(matches!(
+            &drained[0],
+            TestMessage::SetFloor { height } if *height == Height::new(5)
+        ));
+        assert!(matches!(
+            &drained[1],
+            TestMessage::Prune { height } if *height == Height::new(7)
+        ));
+        assert!(matches!(
+            &drained[2],
+            TestMessage::GetInfo {
+                identifier: Identifier::Height(height),
+                ..
+            } if *height == Height::new(7)
+        ));
+        assert!(matches!(
+            &drained[3],
+            TestMessage::GetBlock {
+                identifier: Identifier::Height(height),
+                ..
+            } if *height == Height::new(7)
+        ));
+        assert!(matches!(
+            &drained[4],
+            TestMessage::GetFinalization { height, .. } if *height == Height::new(7)
+        ));
+    }
+
     #[test]
     fn policy_replaces_floor_and_prune_and_drops_stale_pending_on_drain() {
         let mut overflow = pending();
```
</details>

consensus/src/marshal/core/actor.rs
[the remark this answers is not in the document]
patrick-ogrady: If the mailbox is closed, we are shutting down. This is fine.

PR #3786, opened 2026-05-14
[resolver] Migrate to Feedback

patrick-ogrady:
Blocking: #3785


consensus/src/simplex/actors/resolver/ingress.rs
```diff
@@ -133,31 +133,26 @@ impl Consumer for Handler {
...
-                |response| HandlerMessage::Deliver {
-                    view: View::new(key.into()),
-                    data: value,
-                    response,
-                },
-                false,
-            )
-            .await
+    fn deliver(&mut self, key: Self::Key, value: Self::Value) -> oneshot::Receiver<bool> {
+        let (response, receiver) = oneshot::channel();
```
BrendanChou, line 137: I think we try to adopt tx/rx usually? I see that this file already has response/receiver in it though

resolver/src/p2p/ingress.rs
```diff
@@ -34,16 +38,151 @@ pub enum Message<K, P> {
     Retain { predicate: Predicate<K> },
 }
 
+impl<K: Span, P: PublicKey> Policy for Message<K, P> {
+    fn handle(overflow: &mut VecDeque<Self>, message: Self) -> bool {
```
BrendanChou, line 42: I actually think we should probably use a `Backlog` trait and associated type within `Policy` instead of using `VecDeque` directly. This would let you do more custom logic to store information in the backlog. For example let you store a boolean if there is a `Clear` message or not.

```
trait Backlog<Message> {
  fn len() -> usize; // optional?
  fn is_empty() -> bool;
  fn drain(n: usize) -> Iter<Message>; // returns up to n items
}
```

resolver/src/p2p/ingress.rs
```diff
@@ -34,16 +38,151 @@ pub enum Message<K, P> {
...
+fn retain_keys<K: Span, P: PublicKey>(
+    overflow: &mut VecDeque<Message<K, P>>,
+    predicate: &dyn Fn(&K) -> bool,
+) {
+    overflow.retain_mut(|message| match message {
+        Message::Fetch(requests) => {
+            requests.retain(|request| predicate(&request.key));
+            !requests.is_empty()
+        }
+        Message::Cancel { key } => predicate(key),
```
BrendanChou, line 171: I think this should be `!predicate(key)` since you want you want to keep cancels where the key satisfies the predicate since they won't be removed by this `retain` call (but you previously called `cancel` on them from somewhere, which should be respected)

PR #3768, opened 2026-05-12
[consensus/simplex] Internal Message Migration

patrick-ogrady:
Related: #3739

PR #3789, opened 2026-05-14
[actor] Add Custom Overflow Types

patrick-ogrady:
Related: #3739 + #3768

### DIFF
--- a/actor/src/mailbox.rs
+++ b/actor/src/mailbox.rs
@@ -52,7 +52,7 @@ use std::{
 
 /// Retained overflow messages for a mailbox policy.
 pub trait Overflow<T>: Default {
-    /// Return whether no messages are retained.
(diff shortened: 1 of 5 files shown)


actor/src/mailbox.rs
```diff
@@ -50,8 +50,43 @@ use std::{
...
+    /// Return whether the retained message set is empty.
+    fn is_empty(&self) -> bool;
+
+    /// Drain retained messages into `push` in delivery order.
+    ///
+    /// If `push` returns `Some`, the undelivered message and any later messages
+    /// must remain retained for a future drain.
+    fn drain<F>(&mut self, push: F)
+    where
+        F: FnMut(T) -> Option<T>;
```
patrick-ogrady, line 64: Originally used a `Result` here but the message returned could be much larger than ().

PR #3802, opened 2026-05-15
[actor] Add Backoff Metrics

patrick-ogrady:


consensus/src/simplex/actors/resolver/actor.rs
```diff
@@ -98,7 +98,8 @@ impl<
             .and_then(|index| participants.key(index))
             .cloned();
 
-        let (handler_tx, mut handler_rx) = mailbox::new(self.mailbox_size);
+        let (handler_tx, mut handler_rx) =
```
patrick-ogrady: disambiguate mailboxes here?

actor/src/mailbox.rs
```diff
@@ -40,6 +40,15 @@
 //! however, are not globally ordered and may be observed in any interleaving.
 
 use crate::Feedback;
+use commonware_runtime::{
+    telemetry::metrics::{Counter, MetricsExt as _},
+    Metrics,
+};
+#[cfg(test)]
```
cronokirby: why not move these along with the other test gated stuff into the test module?

cronokirby (commented):
Just the one comment, otherwise looks good, and this approach is nice :)

actor/src/mailbox.rs
patrick-ogrady: because we haven't achieved AGI yet

patrick-ogrady: ah, its because we need it for `loom`

PR #3805, opened 2026-05-16
[p2p] Final Nits (2)

patrick-ogrady:


patrick-ogrady:
- [P1] Preserve simulated send ordering before link updates — /Users/patrickogrady/code/monorepo-migrate-p2p/
    p2p/src/simulated/network.rs:911-916
    When a caller mutates links or bandwidth immediately after send, this only enqueues into a per-sender
    mailbox and returns before the network actor has sampled the current link. The queued task can then be
    processed after a subsequent remove_link/add_link, so an already-submitted message uses the new topology;
    cargo test -p commonware-p2p --lib currently fails
    simulated::tests::test_high_latency_message_blocks_followup with the slow message arriving after ~1s instead
    of the expected ~6s. Please restore the previous acknowledgment/order or snapshot the route before
    returning.

patrick-ogrady:
bugbot run

PR #3814, opened 2026-05-17
[actor] (Re)introduce `Rejected`

patrick-ogrady:
To keep handling trait-agnostic in downstream actors, we need upstream actors to be more specific about their behavior.


patrick-ogrady:
butbot run

PR #3827, opened 2026-05-19
[consensus/marshaled] Time `standard` parent fetches during `propose`

clabby:
## Overview

Adds the `parent_fetch_duration` histogram from `coding::Marshaled` to `standard::Inline` and `standard::Deferred`. This metric acts as a proxy to measure marshal's response time for block fetches in the critical path of `Application::propose`.

### DIFF
--- a/consensus/src/marshal/standard/deferred.rs
+++ b/consensus/src/marshal/standard/deferred.rs
@@ -152,6 +152,7 @@ where
     verification_tasks: VerificationTasks<<B as Digestible>::Digest>,
 
     build_duration: Timed,
+    proposal_parent_fetch_duration: Timed,
 }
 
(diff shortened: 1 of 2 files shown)


consensus/src/marshal/standard/deferred.rs
```diff
@@ -470,6 +326,7 @@ where
 
         // Metrics
         let build_duration = self.build_duration.clone();
+        let proposal_parent_fetch_duration = self.proposal_parent_fetch_duration.clone();
```
patrick-ogrady, line 329: Should we have a follow-up PR for the ancestry iterator?

patrick-ogrady (approved):

patrick-ogrady (approved):

PR #3835, opened 2026-05-19
[consensus] Restore marshal ancestor stream symmetry

clabby:
## Overview

Restores symmetry between the `coding` and `standard` marshal mailboxes, where the `BlockProvider` implementation was lost on the `coding` variant in https://github.com/commonwarexyz/monorepo/commit/16e6e61eebdd67fc9d69a0062fadd596d62e3bfc.

Also requires `Unpin` on the `Stream` passed through the `Application` interface's functions, to remove the need for users to manually pin the stream.

Also requires `'static` on the same `Stream` to allow for the `ErasedBlockProvider` in #3764

### DIFF
--- a/consensus/src/marshal/ancestry.rs
+++ b/consensus/src/marshal/ancestry.rs
@@ -13,6 +13,16 @@ use std::{
     task::{Context, Poll},
 };
 
+/// A stream of blocks used by application propose and verify calls.
+pub trait Ancestry<B: Block>: Stream<Item = B> + Send + Unpin + 'static {}
+
+impl<T, B> Ancestry<B> for T
+where
+    T: Stream<Item = B> + Send + Unpin + 'static,
+    B: Block,
+{
+}
+
 /// An interface for providing blocks.
 pub trait BlockProvider: Clone + Send + 'static {
     /// The block type the provider walks.
--- a/consensus/src/marshal/ancestry.rs
+++ b/consensus/src/marshal/ancestry.rs
@@ -23,45 +23,24 @@ where
 {
 }
 
-/// An interface for providing blocks.
+/// An interface for providing parent blocks.
 pub trait BlockProvider: Clone + Send + 'static {
     /// The block type the provider walks.
     type Block: Block;
 
-    /// Subscribe to a block by its digest without requesting it from the network.
+    /// Subscribe to the parent of a known block.
     ///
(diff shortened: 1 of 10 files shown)


consensus/src/marshal/coding/mod.rs
```diff
@@ -115,6 +116,12 @@ mod tests {
     type TestCodedBlock = CodedBlock<CodingB, ReedSolomon<Sha256>, Sha256>;
     type CodingSendRecord = (Round, TestCodedBlock, Recipients<K>);
 
+    #[test]
+    fn mailbox_provides_application_blocks() {
```
patrick-ogrady, line 120: I suppose this is unnecessary but is it worth adding to standard too?

consensus/src/marshal/coding/variant.rs
```diff
@@ -113,3 +114,21 @@ where
...
+    S: Scheme,
+    B: CertifiableBlock<Context = Context<Commitment, P>>,
+    C: CodingScheme,
+    H: Hasher,
+    P: PublicKey,
+{
+    type Block = B;
+
+    async fn subscribe(self, digest: B::Digest) -> Option<Self::Block> {
+        self.subscribe_by_digest(digest, DigestFallback::Wait)
```
patrick-ogrady: I don't think this is necessarily correct? We may want to do more than wait?

clabby: This is all we can do, without the knowledge of the round iiuc. It's also exactly what `AncestorProvider` does:

https://github.com/commonwarexyz/monorepo/blob/14a016a55f439f1425fbe9857d7f7622c6dfa1ee/consensus/src/marshal/core/mailbox.rs#L792-L797

consensus/src/marshal/ancestry.rs
```diff
@@ -187,8 +176,9 @@ mod test {
     impl BlockProvider for MockProvider {
         type Block = Block<Sha256Digest, ()>;
 
-        async fn subscribe(self, digest: Sha256Digest) -> Option<Self::Block> {
-            self.0.into_iter().find(|b| b.digest() == digest)
+        async fn subscribe(self, block: Self::Block) -> Option<Self::Block> {
```
patrick-ogrady: are you changing to `subscribe_parent`?

patrick-ogrady (approved):

PR #3839, opened 2026-05-20
[storage] Expand CRC Fix Scope

patrick-ogrady:
#3837 fixes the original same-page partial shrink case: an already-persisted partial tail can be resized to a shorter partial length without making the page unrecoverable if the CRC rewrite is interrupted.

This PR covers the related cases that #3837 does not cover: shrink targets that land inside a page which was previously full, and therefore was not represented by `partial_page_state`.

Concretely, this adds crash-safety for:

- full page -> partial page, for example `PAGE_SIZE` -> `45`
- multi-page shrink landing inside a previously full page, for example `3 * PAGE_SIZE` -> `PAGE_SIZE + 45`

Those target pages already have valid CRC records on disk, but #3837's protected path only ran when the target page was the current persisted partial tail. Without this PR, the fallback path can later rewrite the landing page as a padded partial page; if that write partially lands before the shorter CRC is durable, recovery can reject the landing page and truncate to the previous page boundary.

This PR fixes that by reading and validating the target page to recover the CRC record that actually committed it, then using the same two-phase CRC transition for any shrink whose target has nonzero partial bytes. If later physical pages need to be truncated first, that truncation is synced before the landing page's shorter CRC is staged.

The implementation also removes the now-unreachable later partial-tail reload path: after the CRC-safe `partial_bytes > 0` branch returns, the remaining resize path is page-aligned shrink only.

Testing:

- Adds full-page-to-partial interrupted CRC stage/invalidation regressions.
- Adds multi-page-to-partial interrupted CRC stage/invalidation regressions.
- Verified with `just test -p commonware-runtime resize_same_page_shrink_survives_interrupted_crc_stage resize_same_page_shrink_survives_interrupted_crc_invalidation resize_full_page_to_partial resize_multi_page_to_partial`.

### DIFF
--- a/runtime/src/utils/buffer/paged/append.rs
+++ b/runtime/src/utils/buffer/paged/append.rs
@@ -787,31 +787,6 @@ impl<B: Blob> Append<B> {
         }
     }
 
-    /// Read a physical page and return its validated logical bytes plus the CRC record used.
-    async fn read_valid_page_with_crc(
-        blob: &B,
-        page: u64,
-        logical_page_size: u64,
-    ) -> Result<(IoBuf, Checksum), Error> {
-        let physical_page_size = logical_page_size
-            .checked_add(CHECKSUM_SIZE)
-            .ok_or(Error::OffsetOverflow)?;
-        let physical_page_start = page
-            .checked_mul(physical_page_size)
-            .ok_or(Error::OffsetOverflow)?;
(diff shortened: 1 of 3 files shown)


roberto-bayardo:
have added this extension of the fix to my PR (plus some other stuff).

PR #3849, opened 2026-05-21
[actor] Add Unreliable Feedback Type

### LINKED ISSUE #3825: [actor] Signal Reliability Requirements
Create a wrapper around Feedback that allows an actor to specify it can only handle "reliable" messages (meaning that effects are never "dropped"). Something like `Reliable<Feedback>` should do?

patrick-ogrady:
Resolves: #3825


patrick-ogrady:
bugbot run

actor/src/mailbox.rs
```diff
@@ -93,11 +94,32 @@ pub trait Policy: Sized {
...
+    /// # Warning
+    ///
+    /// Do not enqueue into the same mailbox from this method or from destructors triggered by
+    /// editing `overflow`. This method runs while the mailbox holds its overflow lock, so same
+    /// mailbox re-entry can deadlock.
+    ///
+    /// This method should not unwind after mutating `overflow`. A panic, including one from a
+    /// destructor triggered while editing `overflow`, can leave retained overflow data stranded in
+    /// the mailbox.
+    fn handle(overflow: &mut Self::Overflow, message: Self);
```
patrick-ogrady, line 111: Very nice that we no longer need to return a boolean here.

actor/src/mailbox.rs
```diff
@@ -115,6 +137,76 @@ pub trait Policy: Sized {
     fn handle(overflow: &mut Self::Overflow, message: Self) -> bool;
 }
 
+struct Infallible;
```
patrick-ogrady: I think we must be able to better align `Infallible` and `LossyMode` from a naming scheme perspective.

actor/src/lib.rs
```diff
@@ -30,5 +28,34 @@ commonware_macros::stability_scope!(BETA {
         }
     }
 
+    /// Feedback from endpoints that may reject work under backpressure.
+    #[derive(Clone, Copy, Debug, PartialEq, Eq)]
+    pub enum Unreliable<T> {
+        /// Normal endpoint feedback from the submission attempt.
+        Feedback(T),
```
patrick-ogrady: nit: `The work was handled reliably.`

I think we should change this to `Handled(T)`.

actor/src/lib.rs
```diff
@@ -17,8 +17,6 @@ commonware_macros::stability_scope!(BETA {
         Ok,
         /// The submission exceeded configured capacity but was handled by the overflow policy.
         Backoff,
-        /// The submission exceeded configured capacity and was rejected by the overflow policy.
-        Rejected,
         /// The endpoint is closed.
```
patrick-ogrady, line 20: We could do `Result` I suppose? And then it would be reliable/unreliable -> closed?

patrick-ogrady:
bugbot run

actor/src/lib.rs
```diff
@@ -30,5 +28,49 @@ commonware_macros::stability_scope!(BETA {
         }
     }
 
+    /// Feedback from endpoints that may reject work under backpressure.
+    #[derive(Clone, Copy, Debug, PartialEq, Eq)]
+    pub enum Unreliable<T> {
```
patrick-ogrady, line 33: This is kept generic in case we start adding more "outcomes" ... if we eventually go a different route, we can do `UnreliableFeedback` or something.

PR #3852, opened 2026-05-21
[runtime/utils/buffer/write] use `Blob::write_at_sync`

andresilva:
This PR updates buffered `Write` to use `Blob::write_at_sync` when sync only needs to persist the currently buffered tip.

`Write` now keeps durability state alongside the tip buffer to track whether a prior plain `write_at` or `resize` requires a full `Blob::sync`. If no such mutation happened, `sync` can persist the buffered range directly (with `Blob::write_at_sync`), otherwise it falls back to a full `sync`.

### DIFF
--- a/runtime/src/utils/buffer/write.rs
+++ b/runtime/src/utils/buffer/write.rs
@@ -2,8 +2,11 @@ use crate::{buffer::tip::Buffer, Blob, Buf, BufferPool, BufferPooler, Error, IoB
 use commonware_utils::sync::AsyncRwLock;
 use std::{num::NonZeroUsize, sync::Arc};
 
-/// Buffered tip data and durability bookkeeping.
-struct State {
+/// Buffered tip data, blob handle, and durability bookkeeping.
+struct State<B: Blob> {
+    /// The underlying blob to write to.
+    blob: B,
+
     /// The buffer containing the data yet to be appended to the tip of the
     /// underlying blob.
     buffer: Buffer,
@@ -14,6 +17,46 @@ struct State {
     /// cannot use [Blob::write_at_sync] because that only makes the current write
     /// durable.
     needs_sync: bool,
+}
+
+impl<B: Blob> State<B> {
+    async fn read_at(&self, offset: u64, len: usize) -> Result<IoBufs, Error> {
+        Ok(self.blob.read_at(offset, len).await?.freeze())
+    }
+
+    async fn write_at(&mut self, offset: u64, bufs: impl Into<IoBufs> + Send) -> Result<(), Error> {
+        self.blob.write_at(offset, bufs).await?;
+        self.needs_sync = true;
+        Ok(())
+    }
+
+    async fn write_at_sync(
+        &mut self,
+        offset: u64,
+        bufs: impl Into<IoBufs> + Send,
+    ) -> Result<(), Error> {
+        if self.needs_sync {
+            self.write_at(offset, bufs).await?;
+            self.sync().await
+        } else {
+            self.blob.write_at_sync(offset, bufs).await
+        }
+    }
+
+    async fn resize(&mut self, len: u64) -> Result<(), Error> {
+        self.blob.resize(len).await?;
+        self.needs_sync = true;
+        Ok(())
+    }
+
+    async fn sync(&mut self) -> Result<(), Error> {
+        if !self.needs_sync {
+            return Ok(());
+        }
+        self.blob.sync().await?;
+        self.needs_sync = false;
+        Ok(())
+    }
 }
 
 /// A writer that buffers the raw content of a [Blob] to optimize the performance of appending or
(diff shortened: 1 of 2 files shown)


runtime/src/utils/buffer/write.rs
```diff
@@ -191,27 +210,43 @@ impl<B: Blob> Write<B> {
...
 
     /// Flush buffered bytes and durably sync the underlying blob.
     pub async fn sync(&self) -> Result<(), Error> {
-        let mut buffer = self.buffer.write().await;
-        if let Some((buf, offset)) = buffer.take() {
+        let mut state = self.state.write().await;
+        if let Some((buf, offset)) = state.buffer.take() {
+            if !state.needs_sync {
+                return self.blob.write_at_sync(offset, buf).await;
+            }
```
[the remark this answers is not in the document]
andresilva: This does not change the retry semantics of `Write::sync()`. The previous implementation also drained the buffer before completing the underlying storage operations, so a failed `write_at` or sync already left the writer in a non-retryable state. Errors from mutable storage operations are treated as fatal by callers, this path only swaps the successful single-buffer flush from `write_at + sync` to `write_at_sync`.

runtime/src/utils/buffer/write.rs
```diff
@@ -175,11 +193,12 @@ impl<B: Blob> Write<B> {
             // below. Removing this inefficiency may not be worth the additional complexity.
             let direct = bufs.split_to(chunk_len);
             self.blob.write_at(current_offset, direct).await?;
+            state.needs_sync = true;
```
patrick-ogrady: I wonder if we should put `blob` in state and implement the blob interface? Then, you wouldn't need these manual needs_sync calls (maybe that ends up being much more gross for some reason).

runtime/src/utils/buffer/write.rs
```diff
@@ -2,6 +2,75 @@ use crate::{buffer::tip::Buffer, Blob, Buf, BufferPool, BufferPooler, Error, IoB
...
+    async fn write_at_sync(
+        &mut self,
+        offset: u64,
+        bufs: impl Into<IoBufs> + Send,
+    ) -> Result<(), Error> {
+        if self.needs_sync {
+            self.write_at(offset, bufs).await?;
+            self.sync().await
+        } else {
+            // If `write_at_sync` fails, a later sync must not treat the drained
```
patrick-ogrady, line 47: tricky

runtime/src/utils/buffer/write.rs
```diff
@@ -138,8 +207,8 @@ impl<B: Blob> Write<B> {
             .checked_add(bufs.remaining() as u64)
             .ok_or(Error::OffsetOverflow)?;
 
-        // Acquire a write lock on the buffer.
-        let mut buffer = self.buffer.write().await;
+        // Acquire a write lock on the buffer state.
+        let mut state = self.state.write().await;
```
patrick-ogrady, line 221: for consistency, we could do `let buffer = &mut state.buffer;` here?

runtime/src/utils/buffer/write.rs
```diff
@@ -88,8 +156,9 @@ impl<B: Blob> Write<B> {
             .checked_add(len as u64)
             .ok_or(Error::OffsetOverflow)?;
 
-        // Acquire a read lock on the buffer.
-        let buffer = self.buffer.read().await;
+        // Acquire a read lock on the buffer state.
+        let state = self.state.read().await;
+        let buffer = &state.buffer;
```
patrick-ogrady, line 171: or, we could remove this line and use state.buffer in this function (I'm down for that)

runtime/src/utils/buffer/write.rs
andresilva: This is mainly to keep the existing semantics (there were some tests in storage that relied on this). In practice we treat all storage errors as fatal.

runtime/src/utils/buffer/write.rs
```diff
@@ -45,20 +114,20 @@ use std::{num::NonZeroUsize, sync::Arc};
...
     /// Creates a new [Write] that buffers up to `capacity` bytes of data to be appended to the tip
     /// of `blob` with the provided `size`.
     pub fn new(blob: B, size: u64, capacity: NonZeroUsize, pool: BufferPool) -> Self {
         Self {
-            blob,
-            buffer: Arc::new(AsyncRwLock::new(Buffer::new(size, capacity.get(), pool))),
+            state: Arc::new(AsyncRwLock::new(State {
+                blob,
+                buffer: Buffer::new(size, capacity.get(), pool),
+                needs_sync: false,
```
patrick-ogrady: nitty but technically true:

```
  - [P2] Keep sync as a barrier for wrapped blob state — /Users/patrickogrady/code/monorepo-pr-3852/runtime/src/utils/buffer/write.rs:65-66
    When a Write is constructed around a blob that already has unsynced mutations, such as blob.write_at(...) or blob.resize(...) before handing
    the blob to Write, needs_sync starts false and this branch returns without calling Blob::sync(). The previous implementation always synced
    the underlying blob, so Write::sync() can now acknowledge durability while those pre-existing changes are still lost after a crash or reopen.
```

runtime/src/utils/buffer/write.rs
andresilva: You need to keep the lock guard in its own variable, `&mut self.state.write().await.buffer` won't compile.

runtime/src/utils/buffer/write.rs
andresilva: Same as above.

runtime/src/utils/buffer/write.rs
andresilva: I could initialize this as `true` I guess, but technically `Write` is only keeping track of its own writes.

PR #3857, opened 2026-05-22
[runtime/utils/buffer/append] use `Blob::write_at_sync`

andresilva:
This PR updates `buffer::paged::Append` to use `Blob::write_at_sync` when a flush can be persisted with a single physical write. When a flush is split around a protected CRC slot, or when a prior plain write or resize still needs a durability barrier, it falls back to plain writes followed by a single full sync. It now tracks whether prior blob mutations still need a full sync so `sync()` can skip work when there is nothing pending. Resize/shrink handling was adjusted so real physical resizes are made durable correctly, while logical-only partial shrinks can keep using range sync for their CRC updates.

Depends on #3852.


patrick-ogrady:
bugbot run

PR #3858, opened 2026-05-22
[storage] use `Blob::write_at_sync`

andresilva:
This PR updates `Metadata` full rewrites to use `Blob::write_at_sync` when the new metadata payload does not shrink the blob. A non-shrinking rewrite is a single durable write, so it no longer needs a separate sync. Shrinking rewrites still use `write_at + resize + sync` because the resize must also be made durable.

The rest of the changes replace manual `put`/`write_at` followed by `sync` with the corresponding sync helper where that is equivalent and clearer. This includes uses of `Metadata::put_sync`, `Archive::put_sync`, `Cache::put_sync`, and direct `Blob::write_at_sync` in tests.

Most storage primitives already benefit from `write_at_sync` indirectly through the runtime buffering layers. `buffer::Write` and `buffer::paged::Append` can use `write_at_sync` when flushing a single write with no earlier dirty mutation, so databases built on journals, ordinal storage, cache/archive paths, and QMDB logs inherit that behavior.

Depends on #3857.


storage/src/journal/segmented/variable.rs
```diff
@@ -1364,13 +1363,12 @@ mod tests {
...
                 write_buffer: NZUsize!(1024),
             };
-
-            // Leave one byte in the first page so the trailing bytes below cross the page
-            // boundary and repair must issue a physical resize.
             let section = 1u64;
             let item = [10u8; 1021];
             let item_record_size =
                 UInt(item.encode_size() as u32).encode_size() + item.encode_size();
+            // Leave one byte in the first page so the trailing bytes below cross the page
```
patrick-ogrady: move this back

storage/src/metadata/mod.rs
```diff
@@ -1006,8 +1011,10 @@ mod tests {
 
             // Initial data
             metadata.put(U64::new(1), b"first".to_vec());
-            metadata.put(U64::new(2), b"second".to_vec());
-            metadata.sync().await.unwrap();
+            metadata
```
patrick-ogrady, line 1014: keep these as a separate put/sync operation? it leaks that the underlying is done together?

patrick-ogrady (commented):
minor comments

patrick-ogrady:
bugbot run

patrick-ogrady (approved):

PR #3860, opened 2026-05-22
[runtime/iobuf/pool] improve buffer pool defaults

andresilva:


runtime/src/iobuf/pool.rs
```diff
@@ -192,14 +192,13 @@ impl BufferPoolConfig {
     /// Cache-line alignment is used because network buffers don't require page
     /// alignment for DMA, and smaller alignment reduces internal fragmentation.
     pub const fn for_network() -> Self {
-        let cache_line = NZUsize!(cache_line_size());
         Self {
-            pool_min_size: 1024,
+            pool_min_size: 0,
```
andresilva, line 164: The pool is now faster than going to the system allocator, so there's no reason to bypass it in the defaults.

runtime/src/iobuf/pool.rs
```diff
@@ -192,14 +192,13 @@ impl BufferPoolConfig {
...
             min_size: NZUsize!(1024),
-            max_size: NZUsize!(64 * 1024),
+            max_size: NZUsize!(128 * 1024),
```
andresilva, line 166: Why not? This means this pool will use 1GB of memory at peak, same as the storage pool below.

runtime/src/iobuf/pool.rs
```diff
@@ -192,14 +192,13 @@ impl BufferPoolConfig {
...
             max_per_class: NZU32!(4096),
             prefill: false,
-            alignment: cache_line,
+            alignment: NZUsize!(1),
```
andresilva, line 169: We don't actually need any specific alignment here, and while this doesn't make any difference once the buffer is allocated and pooled the initial allocation is slower.

runtime/src/iobuf/pool.rs
```diff
@@ -212,12 +211,12 @@ impl BufferPoolConfig {
     pub fn for_storage() -> Self {
         let page = NZUsize!(page_size());
         Self {
-            pool_min_size: 1024,
+            pool_min_size: 0,
```
andresilva, line 180: Same as above.

runtime/src/iobuf/pool.rs
```diff
@@ -212,12 +211,12 @@ impl BufferPoolConfig {
...
             min_size: page,
             max_size: NZUsize!(8 * 1024 * 1024),
             max_per_class: NZU32!(64),
             prefill: false,
-            alignment: page,
+            alignment: NZUsize!(1),
```
andresilva, line 186: Same as above. Eventually once we use `O_DIRECT` this will need to be aligned to block size or page size.

runtime/src/iobuf/pool.rs
```diff
@@ -184,40 +175,35 @@ pub struct BufferPoolConfig {
...
-    /// Cache-line alignment is used because network buffers don't require page
-    /// alignment for DMA, and smaller alignment reduces internal fragmentation.
+    /// Network operations typically need multiple concurrent buffers per
+    /// connection (message, encoding, encryption) so we allow 4096 buffers per
+    /// size class.
     pub const fn for_network() -> Self {
-        let cache_line = NZUsize!(cache_line_size());
         Self {
-            pool_min_size: 1024,
+            pool_min_size: 0,
```
patrick-ogrady, line 164: Should we just remove this config option? I suppose it makes sense to keep around for an application to tune?

runtime/src/iobuf/pool.rs
patrick-ogrady: Do you want to link a TODO issue for this?

runtime/src/iobuf/pool.rs
andresilva: Yeah that was my thinking as well, someone might want to tune the buffer pool for larger buffers and this allows bypassing it. But we should revisit this decision in the future.

runtime/src/iobuf/pool.rs
andresilva: Added!

patrick-ogrady (approved):

patrick-ogrady:
bugbot run

PR #3868, opened 2026-05-25
[glue] Nits

patrick-ogrady:
Related: #3381

## TODO
- [ ] improve crate description

### DIFF
--- a/storage/src/qmdb/sync/compact.rs
+++ b/storage/src/qmdb/sync/compact.rs
@@ -72,8 +72,12 @@ use commonware_codec::{
 };
 use commonware_cryptography::{Digest, Hasher};
 use commonware_parallel::Strategy;
-use commonware_runtime::{Buf, BufMut, Clock, Metrics, Storage};
-use commonware_utils::{sync::AsyncRwLock, Array};
+use commonware_runtime::{Buf, BufMut, Clock, Metrics, Storage, Supervisor};
+use commonware_utils::{
+    channel::{fallible::OneshotExt, oneshot},
+    sync::AsyncRwLock,
+    Array,
+};
 use std::{future::Future, num::NonZeroU64, sync::Arc};
 
 /// Compact-sync target for a compact-storage database.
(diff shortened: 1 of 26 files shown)


storage/src/qmdb/sync/compact.rs
```diff
@@ -334,49 +373,81 @@ where
...
-        .map_err(Error::Resolver)?;
-
-    if state.leaf_count != target.leaf_count {
-        return Err(Error::Engine(EngineError::UnexpectedLeafCount {
-            expected: target.leaf_count,
-            actual: state.leaf_count,
-        }));
-    }
+    loop {
+        let FetchResult { state, success_tx } = config
```
patrick-ogrady, line 377: If we don't preserve the success_tx here, we won't be able to block a peer that is serving us invalid proofs?

I'm not clear enough on whether we would retry in this case but if not, we'd also halt here.

commonware-llm:
Follow-up review fixes addressed in this checkout:

- Compact P2P invalid-state feedback: compact sync now reports downstream validation failure through the resolver feedback path, so a peer that serves a valid final commit proof with bad frontier pins is treated as invalid instead of accepted before root reconstruction.
- Compact P2P cancellation: dropped compact-state requests now enqueue `CancelState`; the actor prunes closed subscribers and cancels the generic resolver fetch when the last waiter is gone.
- Rewind-window safety: database sets now expose finite rewind depth and stateful startup asserts `max_pending_acks` does not exceed what the DB can rewind across.
- Startup sync handoff: the sync anchor block is acknowledged without re-finalizing, while the first post-sync finalized block is handed off and persisted normally.
- Syncer observation barrier: finalized blocks are not acknowledged merely because their target update was enqueued; the caller waits until the live sync coordinator records the update, retrying if sync completes first.
- No-op target updates: single and tuple state sync now advance anchor/generation without sending duplicate per-DB target updates when the target is unchanged.
- Tuple startup recovery rewind: only DBs whose committed target differs from marshal's target are rewound, avoiding unnecessary rewind of already-aligned DBs.
- Processor cancellation: initial ancestry reads in propose/verify are cancellation-aware, so dropped response receivers do not pin processor work waiting on ancestry.
- Sync target authentication: `matches_sync_target` now checks the full target contract, not only root equality. Replay wrappers verify range start/end, compact wrappers verify leaf count, and current wrappers verify ops root, witness, and range.

Verification run locally:

- `just test -p commonware-glue stateful::db::compact_p2p`
- `cargo fmt --check` (passes with the existing stable-rustfmt warnings about unstable import grouping options)
- `git diff --check`

PR #3869, opened 2026-05-25
[runtime/storage/iouring] serialize storage operations

andresilva:
We need to serialize all operations on the storage namespace, otherwise a successful `remove()` can be followed immediately by a `scan()` that still finds the same blob name, or a concurrent `open()` can recreate a blob after `remove_file()` succeeds but before `remove()` returns. This is the exact same thing we do in the tokio backend.

I didn't write a test for this since it's a bit tricky (and backend-specific) to create the race condition.


patrick-ogrady (approved):
very nice

PR #3873, opened 2026-05-26
[storage/index] Move pointer into cursor state

danlaine:
Follow up to #3760.

## Summary

Refines the in-place `Cursor` rewrite from #3760 by lifting the `prev` and `current` pointers into the `State` enum. Each state variant now owns exactly the pointers it needs, so the contract previously enforced by runtime asserts becomes compile-time.

## Motivation

The post-#3760 `Cursor` encodes "do I have a live record under the cursor?" in three correlated places:

```rust
struct Cursor { state: State, prev: Option<NonNull<_>>, current: Option<NonNull<_>>, ... }
enum State { NeedNext, Active, Done, EntryRemoved }
```

with `update`/`insert`/`delete` each containing the same pattern:

```rust
match self.state { State::Active => {}, _ => panic!(...) }
assert!(self.current.is_some(), "Active state requires current");
let current = self.current.unwrap();
self.record_mut(current). …
```

The invariant *"`Active` implies `current.is_some()`"* is documented, asserted at runtime, and unwrapped — the compiler could enforce these checks on its own. Likewise the `insert(Done)` branch guards with `self.current.or(self.prev)`, even though every path into `Done` leaves `current` populated, so the `.or` fallback is dead code.

## Changes

Move the pointers into the variants:

```rust
enum State<V> {
    NeedNext { from: Option<NonNull<Record<V>>> },
    Active   { current: NonNull<Record<V>>, prev: Option<NonNull<Record<V>>> },
    Done     { tail: NonNull<Record<V>> },
    EntryRemoved,
}
```

and drop the standalone `prev` / `current` struct fields. Each variant carries only the pointers meaningful to that state, and `Active`'s `current` is a non-Option `NonNull` — the type system now enforces what the assert was checking.

### DIFF
--- a/storage/src/index/storage.rs
+++ b/storage/src/index/storage.rs
@@ -38,39 +38,50 @@ const MUST_CALL_NEXT: &str = "must call Cursor::next()";
 /// Panic message shown when `update()` or `delete()` is called after [Cursor] has returned `None`.
 const NO_ACTIVE_ITEM: &str = "no active item in Cursor";
 
-#[derive(PartialEq, Eq)]
-enum State {
+/// Position of the [Cursor] within the linked list owned by its entry.
+enum State<V: Eq + Send + Sync> {
     /// Before first `next()` call, or immediately after `insert()`/`delete()`.
-    NeedNext,
-    /// `next()` returned a value; `update()`/`delete()` are valid.
-    Active,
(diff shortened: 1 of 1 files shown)


roberto-bayardo:
nice!

PR #3876, opened 2026-05-26
[storage/index] Deduplicate `Cursor`

danlaine:
`storage::index::unordered` and `storage::index::ordered` each define a `pub struct Cursor<'a, K, V>` newtype around the shared `storage::Cursor` implementation. Each newtype carries a `new` constructor and a five-method `CursorTrait` impl whose every method is a single-line delegation to `self.inner.<method>()`.

This PR replaces both newtypes with `pub type` aliases. The aliases name the same concrete type the wrappers used to wrap; the trait impl already exists on the underlying `storage::Cursor` and applies through the alias.

### DIFF
--- a/storage/src/index/ordered.rs
+++ b/storage/src/index/ordered.rs
@@ -41,49 +41,8 @@ impl<K: Ord + Send + Sync, V: Eq + Send + Sync> IndexEntry<V>
     }
 }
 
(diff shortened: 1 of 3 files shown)


patrick-ogrady (approved):
nice

PR #3878, opened 2026-05-26
[ci] partition harder

andresilva:


.github/workflows/slow.yml
patrick-ogrady, line 89: of the things to partition, I think storage/cryptography would benefit most?

andresilva: I gathered some data from previous runs so that partitioning would more or less even out.

andresilva: The floor is the fuzz consensus twins that take ~30 minutes, those can't be partitioned further. So all of the other partitions were tuned towards that floor.

storage/src/qmdb/current/mod.rs
```diff
@@ -1429,6 +1429,48 @@ pub mod tests {
...
+            with_ordered_variants!(per_variant_slow_test_emit!($prefix, $f, $traced));
+        };
+    }
+
+    // Bulk-generate per-variant slow tests across the 12 unordered variants.
+    macro_rules! per_unordered_variants_slow_tests {
+        ($prefix:ident, $f:path, $traced:literal) => {
+            with_unordered_variants!(per_variant_slow_test_emit!($prefix, $f, $traced));
+        };
+    }
```
clabby, line 1480: We should be able to reduce boilerplate a lot here by using `rstest`?

andresilva: Yeah maybe we could use the `#[case]` support it has to inject the different DB variants. But I never used `rstest` so not sure it would work.

.github/actions/setup/action.yml
```diff
@@ -61,6 +61,12 @@ runs:
         key="${{ inputs.additional-cache-key }}"
         key="${key// /-}"
         key="${key//\//-}"
+        if [[ "$key" =~ ^(.*-part-)([0-9]+)$ ]]; then
+          partition="${BASH_REMATCH[2]}"
+          if (( partition > 8 )); then
+            key="${BASH_REMATCH[1]}$(( (partition - 1) % 8 + 1 ))"
+          fi
+        fi
```
andresilva: This needs to be removed before merge. It was a temporary change to make sure the cache still worked due to the increase in partitions, so that I could get accurate timings.

justfile
```diff
@@ -3,6 +3,7 @@ set positional-arguments := true
 env_nightly_version := env("NIGHTLY_VERSION", "nightly")
 nightly_version := if env_nightly_version != "" { "+" + env_nightly_version } else { "" }
 rustfmt := env("RUSTFMT", "rustfmt")
+partition := "1/1"
```
patrick-ogrady, line 6:
```
• The change preserves the local just API while still letting CI shard fuzz
  and benchmark work.

  The bug was that partition had been added as the second positional arg:

  fuzz fuzz_dir partition='1/1' max_time='60' max_mem='4000'

  That broke the documented local command:

  just fuzz codec/fuzz 30

  because 30 was treated as the partition spec instead of the timeout.

  We fixed it by making partition a top-level just variable with default 1/1,
  and restoring the recipe positional args:

  partition := "1/1"
  fuzz fuzz_dir max_time='60' max_mem='4000':

  So local users keep using:

  just fuzz codec/fuzz
  just fuzz codec/fuzz 30

  and CI can override partition explicitly:

  just partition=1/4 fuzz codec/fuzz

  We applied the same pattern to test-benches because it had the same
  positional-argument problem: partition had displaced the old test_flags/
  lint_flags slots.

  The result is cleaner than passing placeholder defaults like 60 4000, and it
  avoids breaking existing local workflows.
```

.github/workflows/slow.yml
```diff
@@ -149,7 +196,8 @@ jobs:
       id: fuzz-corpus-cache-key
       run: |
         sanitized_fuzz_dir="$(echo "${{ matrix.fuzz_dir }}" | tr '/' '-')"
-        echo "prefix=${{ runner.os }}-fuzz-corpus-${sanitized_fuzz_dir}" >> "$GITHUB_OUTPUT"
+        sanitized_partition="$(echo "${{ matrix.partition || '1/1' }}" | tr '/' '-')"
+        echo "prefix=${{ runner.os }}-fuzz-corpus-${sanitized_fuzz_dir}-${sanitized_partition}" >> "$GITHUB_OUTPUT"
```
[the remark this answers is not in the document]
patrick-ogrady: This is a false positive. What is in the branch now is correct.

PR #3879, opened 2026-05-26
[storage/index] Rename `prune` to `retain`

danlaine:
## Summary

Renames the in-place "remove items matching a predicate" API across the `storage::index` module to match Rust stdlib conventions:

- `Cursor::prune` → `Cursor::retain`
- `Unordered::prune` → `Unordered::retain`
- `Unordered::insert_and_prune` → `Unordered::insert_and_retain`

Predicate semantics flip everywhere from "true = remove" to "true = keep" (matching `Vec::retain` and friends). Parameter names change from `predicate` to `should_retain`. All call sites — internal, in `qmdb::immutable`, in `archive::prunable`, in fuzz targets, in tests — have their predicates negated so the behavior is preserved.

## Motivation

The stdlib's canonical name for "mutate a collection in place, keep items where the predicate is true" is **`retain`**. It exists on: `Vec`, `VecDeque`, `LinkedList`, `HashMap`, `BTreeMap`, `HashSet`, `BTreeSet`, `String`.

### DIFF
--- a/storage/fuzz/fuzz_targets/ordered_index_operations.rs
+++ b/storage/fuzz/fuzz_targets/ordered_index_operations.rs
@@ -107,7 +107,7 @@ fn fuzz(input: FuzzInput) {
                 }
 
                 IndexOperation::Prune { key, prune_value } => {
-                    index.prune(key, |v| *v == *prune_value);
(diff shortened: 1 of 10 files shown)


patrick-ogrady (approved):
very nice

PR #3880, opened 2026-05-26
[broadcast/fuzz] fix subscribe deadlock

andresilva:
`Mailbox::subscribe` was made synchronous in #3785, but the fuzz target still did `drop(mailbox.subscribe(digest).await)`. Before #3785, `.await` resolved the `async fn` to a `Receiver` and `drop` discarded it, fire-and-forget. Post-change, the same expression awaits the `Receiver` directly, which only resolves when a matching message arrives, which with random 32-byte digests it almost never does, so the action loop blocks and the deterministic runtime trips its liveness check. 

The deadlock fix alone would still leave `Subscribe` and `Get` doing very little, since random 32-byte digests almost never match broadcast traffic and the engine's wakeup and cache-hit paths stay cold. To fix that, this PR introduces `DigestSource { Random(FuzzMessage), Recent(u8) }` plus a bounded FIFO ring of broadcast digests. `Recent` indexes into the ring, so the fuzzer now actually hits cache-hit, cache-miss-after-eviction, and live-subscriber-wakeup paths.

### DIFF
--- a/broadcast/fuzz/fuzz_targets/broadcast_engine_operations.rs
+++ b/broadcast/fuzz/fuzz_targets/broadcast_engine_operations.rs
@@ -12,10 +12,8 @@ use commonware_cryptography::{
     Digestible, Hasher, Sha256, Signer,
 };
 use commonware_p2p::{simulated::Network, Recipients};
-use commonware_runtime::{
-    deterministic, Buf, BufMut, Clock, Quota, Runner, Spawner, Supervisor as _,
-};
-use commonware_utils::NZUsize;
+use commonware_runtime::{deterministic, Buf, BufMut, Clock, Quota, Runner, Supervisor as _};
+use commonware_utils::{channel::oneshot, NZUsize};
 use libfuzzer_sys::fuzz_target;
 use rand::{seq::SliceRandom, SeedableRng};
 use std::{
@@ -36,7 +34,7 @@ const MAX_SLEEP_DURATION_MS: u64 = 1000;
 /// Maximum number of recently broadcast digests tracked for `Recent` lookups.
 ///
 /// Indexed by `u8` modulo length, so a larger buffer would leave entries past 255
-/// unreachable via `DigestSource::Recent`.
+/// unreachable via `Source::Recent`.
 const MAX_RECENT_DIGESTS: usize = (u8::MAX as usize) + 1;
 
 #[derive(Clone, Debug, Arbitrary)]
@@ -86,21 +84,21 @@ impl commonware_codec::Read for FuzzMessage {
 
 /// Source for the digest used by `Subscribe`/`Get` actions.
 #[derive(Clone, Debug, Arbitrary)]
-enum DigestSource {
+enum Source {
     /// A random message.
     Random(FuzzMessage),
     /// Index into messages broadcast earlier in this run.
     Recent(u8),
 }
 
-impl DigestSource {
+impl Source {
     /// Resolve to a concrete digest. Returns `None` when `Recent` is selected
     /// before any messages have been broadcast.
     fn resolve(self, recent: &VecDeque<Digest>) -> Option<Digest> {
         match self {
-            DigestSource::Random(message) => Some(message.digest()),
-            DigestSource::Recent(_) if recent.is_empty() => None,
-            DigestSource::Recent(idx) => Some(recent[(idx as usize) % recent.len()]),
+            Source::Random(message) => Some(message.digest()),
+            Source::Recent(_) if recent.is_empty() => None,
+            Source::Recent(idx) => Some(recent[(idx as usize) % recent.len()]),
         }
     }
 }
@@ -114,11 +112,11 @@ enum BroadcastAction {
     },
     Subscribe {
         peer_index: usize,
-        source: DigestSource,
+        digest: Source,
     },
     Get {
         peer_index: usize,
-        source: DigestSource,
+        digest: Source,
     },
     Sleep {
         duration_ms: u64,
@@ -136,11 +134,11 @@ impl<'a> Arbitrary<'a> for BroadcastAction {
             }),
             1 => Ok(BroadcastAction::Subscribe {
                 peer_index: u.arbitrary()?,
-                source: u.arbitrary()?,
+                digest: u.arbitrary()?,
             }),
             2 => Ok(BroadcastAction::Get {
                 peer_index: u.arbitrary()?,
-                source: u.arbitrary()?,
+                digest: u.arbitrary()?,
             }),
             _ => Ok(BroadcastAction::Sleep {
                 duration_ms: u.int_in_range(0..=MAX_SLEEP_DURATION_MS)?,
@@ -199,6 +197,23 @@ fn resolve_recipients(pattern: &RecipientPattern, peers: &[PublicKey]) -> Recipi
             let count = ((seed % peers.len() as u64) as usize).max(1);
             let peer_slice = shuffled_peers.into_iter().take(count).collect();
             Recipients::Some(peer_slice)
+        }
+    }
+}
+
+// Keep subscriptions alive without spawning one task per receiver. Ready
+// subscriptions are validated, while unresolved ones remain pending.
+fn drain_ready_subscriptions(pending: &mut VecDeque<(Digest, oneshot::Receiver<FuzzMessage>)>) {
+    for _ in 0..pending.len() {
+        let (digest, mut receiver) = pending.pop_front().unwrap();
+        match receiver.try_recv() {
+            Ok(message) => {
+                assert_eq!(message.digest(), digest);
+            }
+            Err(oneshot::error::TryRecvError::Empty) => {
+                pending.push_back((digest, receiver));
+            }
+            Err(oneshot::error::TryRecvError::Closed) => {}
         }
     }
 }
@@ -271,6 +286,8 @@ fn fuzz(input: FuzzInput) {
 
         // Execute fuzzed actions
         let mut recent_digests: VecDeque<Digest> = VecDeque::with_capacity(MAX_RECENT_DIGESTS);
+        let mut pending_subscriptions: VecDeque<(Digest, oneshot::Receiver<FuzzMessage>)> =
+            VecDeque::with_capacity(MAX_RECENT_DIGESTS);
         for action in input.actions {
             match action {
                 BroadcastAction::SendMessage {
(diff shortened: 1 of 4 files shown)


broadcast/fuzz/fuzz_targets/broadcast_engine_operations.rs
```diff
@@ -27,6 +33,12 @@ const TEST_QUOTA: Quota = Quota::per_second(NonZeroU32::MAX);
 /// and can only represent durations up to ~584 years.
 const MAX_SLEEP_DURATION_MS: u64 = 1000;
 
+/// Maximum number of recently broadcast digests tracked for `Recent` lookups.
+///
+/// Indexed by `u8` modulo length, so a larger buffer would leave entries past 255
+/// unreachable via `DigestSource::Recent`.
+const MAX_RECENT_DIGESTS: usize = (u8::MAX as usize) + 1;
```
patrick-ogrady, line 35: I don't think this is necessary given the action limit is 10?

broadcast/fuzz/fuzz_targets/broadcast_engine_operations.rs
```diff
@@ -249,18 +283,34 @@ fn fuzz(input: FuzzInput) {
...
                     let peer = peers[clamped_peer_idx].clone();
 
                     if let Some(mailbox) = mailboxes.get(&peer).cloned() {
-                        drop(mailbox.subscribe(digest).await);
+                        // Spawn the await so the responder stays live and the
+                        // engine actually wakes it when a matching message
+                        // arrives.
+                        let receiver = mailbox.subscribe(digest);
+                        context.child("subscriber").spawn(|_| async move {
+                            let _ = receiver.await;
```
patrick-ogrady: should confirm result matches expectations?

broadcast/fuzz/fuzz_targets/broadcast_engine_operations.rs
```diff
@@ -249,18 +283,34 @@ fn fuzz(input: FuzzInput) {
...
                     let clamped_peer_idx = peer_index % peers.len();
                     let peer = peers[clamped_peer_idx].clone();
 
                     if let Some(mailbox) = mailboxes.get(&peer).cloned() {
-                        drop(mailbox.subscribe(digest).await);
+                        // Spawn the await so the responder stays live and the
+                        // engine actually wakes it when a matching message
+                        // arrives.
+                        let receiver = mailbox.subscribe(digest);
+                        context.child("subscriber").spawn(|_| async move {
```
patrick-ogrady: This can create an unbounded number of tasks ... that being said, if bounded by 10 should be ok?

patrick-ogrady: Alternative would be something like (if we don't want to be so bound to the number of actions):

```
diff --git a/broadcast/fuzz/fuzz_targets/broadcast_engine_operations.rs b/broadcast/fuzz/fuzz_targets/broadcast_engine_operations.rs
index 6ba4fe97e..a38e3a349 100644
--- a/broadcast/fuzz/fuzz_targets/broadcast_engine_operations.rs
+++ b/broadcast/fuzz/fuzz_targets/broadcast_engine_operations.rs
@@ -12,10 +12,8 @@ use commonware_cryptography::{
     Digestible, Hasher, Sha256, Signer,
 };
 use commonware_p2p::{simulated::Network, Recipients};
-use commonware_runtime::{
-    deterministic, Buf, BufMut, Clock, Quota, Runner, Spawner, Supervisor as _,
-};
-use commonware_utils::NZUsize;
+use commonware_runtime::{deterministic, Buf, BufMut, Clock, Quota, Runner, Supervisor as _};
+use commonware_utils::{channel::oneshot, NZUsize};
 use rand::{seq::SliceRandom, SeedableRng};
 use std::{
@@ -36,7 +34,7 @@ const MAX_SLEEP_DURATION_MS: u64 = 1000;
 /// Maximum number of recently broadcast digests tracked for `Recent` lookups.
 ///
 /// Indexed by `u8` modulo length, so a larger buffer would leave entries past 255
-/// unreachable via `DigestSource::Recent`.
+/// unreachable via `Source::Recent`.
 const MAX_RECENT_DIGESTS: usize = (u8::MAX as usize) + 1;
 
 #[derive(Clone, Debug, Arbitrary)]
@@ -86,21 +84,21 @@ impl commonware_codec::Read for FuzzMessage {
 
 /// Source for the digest used by `Subscribe`/`Get` actions.
 #[derive(Clone, Debug, Arbitrary)]
-enum DigestSource {
+enum Source {
     /// A random message.
     Random(FuzzMessage),
     /// Index into messages broadcast earlier in this run.
     Recent(u8),
 }
 
-impl DigestSource {
+impl Source {
     /// Resolve to a concrete digest. Returns `None` when `Recent` is selected
     /// before any messages have been broadcast.
     fn resolve(self, recent: &VecDeque<Digest>) -> Option<Digest> {
         match self {
-            DigestSource::Random(message) => Some(message.digest()),
-            DigestSource::Recent(_) if recent.is_empty() => None,
-            DigestSource::Recent(idx) => Some(recent[(idx as usize) % recent.len()]),
+            Source::Random(message) => Some(message.digest()),
+            Source::Recent(_) if recent.is_empty() => None,
+            Source::Recent(idx) => Some(recent[(idx as usize) % recent.len()]),
         }
     }
 }
@@ -114,11 +112,11 @@ enum BroadcastAction {
     },
     Subscribe {
         peer_index: usize,
-        source: DigestSource,
+        digest: Source,
     },
     Get {
         peer_index: usize,
-        source: DigestSource,
+        digest: Source,
     },
     Sleep {
         duration_ms: u64,
@@ -136,11 +134,11 @@ impl<'a> Arbitrary<'a> for BroadcastAction {
             }),
             1 => Ok(BroadcastAction::Subscribe {
                 peer_index: u.arbitrary()?,
-                source: u.arbitrary()?,
+                digest: u.arbitrary()?,
             }),
             2 => Ok(BroadcastAction::Get {
                 peer_index: u.arbitrary()?,
-                source: u.arbitrary()?,
+                digest: u.arbitrary()?,
             }),
             _ => Ok(BroadcastAction::Sleep {
                 duration_ms: u.int_in_range(0..=MAX_SLEEP_DURATION_MS)?,
@@ -203,6 +201,23 @@ fn resolve_recipients(pattern: &RecipientPattern, peers: &[PublicKey]) -> Recipi
     }
 }
 
+// Keep subscriptions alive without spawning one task per receiver. Ready
+// subscriptions are validated, while unresolved ones remain pending.
+fn drain_ready_subscriptions(pending: &mut VecDeque<(Digest, oneshot::Receiver<FuzzMessage>)>) {
+    for _ in 0..pending.len() {
+        let (digest, mut receiver) = pending.pop_front().unwrap();
+        match receiver.try_recv() {
+            Ok(message) => {
+                assert_eq!(message.digest(), digest);
+            }
+            Err(oneshot::error::TryRecvError::Empty) => {
+                pending.push_back((digest, receiver));
+            }
+            Err(oneshot::error::TryRecvError::Closed) => {}
+        }
+    }
+}
+
 fn fuzz(input: FuzzInput) {
@@ -271,6 +286,8 @@ fn fuzz(input: FuzzInput) {
 
         // Execute fuzzed actions
         let mut recent_digests: VecDeque<Digest> = VecDeque::with_capacity(MAX_RECENT_DIGESTS);
+        let mut pending_subscriptions: VecDeque<(Digest, oneshot::Receiver<FuzzMessage>)> =
+            VecDeque::with_capacity(MAX_RECENT_DIGESTS);
         for action in input.actions {
             match action {
                 BroadcastAction::SendMessage {
@@ -290,38 +307,41 @@ fn fuzz(input: FuzzInput) {
                         let _ = mailbox.broadcast(resolved_recipients, message);
                     }
                 }
-                BroadcastAction::Subscribe { peer_index, source } => {
-                    let Some(digest) = source.resolve(&recent_digests) else {
-                    };
-                    let clamped_peer_idx = peer_index % peers.len();
-                    let peer = peers[clamped_peer_idx].clone();
+                BroadcastAction::Subscribe {
+                    peer_index,
+                    digest: source,
+                } => {
+                    if let Some(digest) = source.resolve(&recent_digests) {
+                        let clamped_peer_idx = peer_index % peers.len();
+                        let peer = peers[clamped_peer_idx].clone();
 
-                    if let Some(mailbox) = mailboxes.get(&peer).cloned() {
-                        // Spawn the await so the responder stays live and the
-                        // engine actually wakes it when a matching message
-                        // arrives.
-                        let receiver = mailbox.subscribe(digest);
-                        context.child("subscriber").spawn(|_| async move {
-                            let _ = receiver.await;
-                        });
+                        if let Some(mailbox) = mailboxes.get(&peer).cloned() {
+                            let receiver = mailbox.subscribe(digest);
+                            pending_subscriptions.push_back((digest, receiver));
+                            if pending_subscriptions.len() > MAX_RECENT_DIGESTS {
+                                pending_subscriptions.pop_front();
+                        }
                     }
                 }
-                BroadcastAction::Get { peer_index, source } => {
-                    let Some(digest) = source.resolve(&recent_digests) else {
-                    };
-                    let clamped_peer_idx = peer_index % peers.len();
-                    let peer = peers[clamped_peer_idx].clone();
+                BroadcastAction::Get {
+                    peer_index,
+                    digest: source,
+                } => {
+                    if let Some(digest) = source.resolve(&recent_digests) {
+                        let clamped_peer_idx = peer_index % peers.len();
+                        let peer = peers[clamped_peer_idx].clone();
 
-                    if let Some(mailbox) = mailboxes.get(&peer).cloned() {
-                        drop(mailbox.get(digest).await);
+                        if let Some(mailbox) = mailboxes.get(&peer).cloned() {
+                            drop(mailbox.get(digest).await);
+                        }
                     }
                 }
                 BroadcastAction::Sleep { duration_ms } => {
                     context.sleep(Duration::from_millis(duration_ms)).await;
                 }
             }
+            drain_ready_subscriptions(&mut pending_subscriptions);
         }
     });
 }

```

patrick-ogrady (commented):
I don't think all of the changes are necessary.

broadcast/fuzz/fuzz_targets/broadcast_engine_operations.rs
andresilva: I decided to make it explicit defensively in case someone changes the arbitrary generator in the future.

broadcast/fuzz/fuzz_targets/broadcast_engine_operations.rs
andresilva: We can await with timeout, or we can bound the maximum number of pending subs as in the patch you posted.

broadcast/fuzz/fuzz_targets/broadcast_engine_operations.rs
patrick-ogrady: if we go that route, I think my approach of bounding tasks then makes sense.

patrick-ogrady (approved):

PR #3883, opened 2026-05-26
[stream/fuzz] fix connection target deadlock on large messages

andresilva:
The `stream` fuzz connection target drove each encrypted `send` and its matching `recv` sequentially on the same task. When a fuzz-generated message exceeded the mock channel's default 64 KiB backpressure threshold, `Sender::send` parked on a drain waiter that only the matching `recv` could release, but `recv` was the next line on the same task and was never reached, so the deterministic runtime caught the self-deadlock and panicked with runtime stalled. Wrapping each `send`/`recv` pair in `futures::join!` polls both halves cooperatively on the same task so the receiver drains the mock buffer while the sender is parked on backpressure, message-order assertions and fuzz coverage are unchanged.


patrick-ogrady (dismissed):
nice find

patrick-ogrady (approved):

patrick-ogrady (dismissed):

patrick-ogrady (approved):

PR #3888, opened 2026-05-27
[runtime/fuzz] cap max cache size to 64MB

andresilva:
Fixes https://github.com/commonwarexyz/monorepo/actions/runs/26500434353/job/78039175528?pr=3878.


runtime/fuzz/fuzz_targets/buffer.rs
```diff
@@ -33,7 +34,7 @@ enum FuzzOperation {
     },
     CreateAppend {
         initial_size: u16,
-        buffer_size: u16,
+        buffer_size: u32,
```
andresilva, line 37: Max buffer size was 1MB, but with this being `u16` it was effectively capped to 64KB.

clabby (approved):

PR #3892, opened 2026-05-27
[glue] Make Compact Handling More Robust

patrick-ogrady:
## Summary

- Validate compact-sync frontier pins against the requested target root before constructing or persisting a compact QMDB.
- Retry invalid peer-provided compact state the same way standard QMDB sync reschedules invalid fetches; resolver feedback is optional and is set to `false` when present.
- Add regression coverage for tampered pinned nodes, malformed compact state, retry without feedback, retry with feedback, and compact P2P delivery handling.

## Bug

Compact sync authenticated the final commit operation with `last_commit_proof`, but that proof alone does not authenticate the compact frontier's `pinned_nodes`. A peer could provide a valid final-commit proof together with a malformed or tampered frontier. The old path only discovered this after attempting to build the compact database, and compact sync treated the bad response as terminal instead of dropping it and fetching another candidate like standard QMDB sync.

## Fix

The sync path now validates the complete compact state before database construction:

1. Check that the response leaf count matches the requested target.
2. Verify `last_commit_proof` for the final commit operation at `leaf_count - 1`.
3. Rebuild an in-memory compact frontier from the supplied pins, derive the inactive peak count from the commit's inactivity floor, and compare that frontier root with the requested target root.

Validation failures reject the candidate and continue fetching. If the resolver supplied a feedback callback, compact sync acknowledges the rejected candidate as `false`; otherwise it still retries, matching standard sync's retry behavior. Only after validation passes does compact sync construct the database, re-check the database root, acknowledge success, and persist the compact state. Local/database construction failures remain terminal.

The compact P2P actor also rejects malformed pinned-node counts before fanout and keeps pending subscribers alive across invalid deliveries.

## Verification

- `just check-fmt`
- `cargo check -p commonware-storage --all-targets`
- `just test -p commonware-storage qmdb::sync::compact::tests`
- `just test -p commonware-storage compact_variable`
- `just test -p commonware-glue stateful::db::p2p::compact::actor::tests`
- `git diff --check`

### DIFF
--- a/storage/src/qmdb/immutable/sync/mod.rs
+++ b/storage/src/qmdb/immutable/sync/mod.rs
@@ -172,7 +172,7 @@ where
     async fn from_compact_state(
         context: Self::Context,
         config: Self::Config,
-        state: sync::compact::VerifiedState<Self::Family, Self::Op, Self::Digest>,
+        state: sync::compact::ValidatedState<Self::Family, Self::Op, Self::Digest>,
     ) -> Result<Self, Error<F>> {
         let (state, root, inactivity_floor_loc) = state.into_parts();
         let sync::compact::State {
--- a/storage/src/qmdb/immutable/sync/mod.rs
+++ b/storage/src/qmdb/immutable/sync/mod.rs
@@ -174,7 +174,11 @@ where
         config: Self::Config,
         state: sync::compact::ValidatedState<Self::Family, Self::Op, Self::Digest>,
     ) -> Result<Self, Error<F>> {
-        let (state, root, inactivity_floor_loc) = state.into_parts();
+        let sync::compact::ValidatedState {
+            state,
+            root,
+            inactivity_floor: inactivity_floor_loc,
+        } = state;
         let sync::compact::State {
             leaf_count,
             pinned_nodes,
@@ -188,7 +192,7 @@ where
         };
         if op_inactivity_floor_loc != inactivity_floor_loc {
             return Err(Error::DataCorrupted(
-                "verified compact state floor mismatch",
+                "validated compact state floor mismatch",
             ));
         }
         let commit_codec_config = config.commit_codec_config.clone();
--- a/storage/src/qmdb/immutable/sync/mod.rs
+++ b/storage/src/qmdb/immutable/sync/mod.rs
@@ -186,15 +186,9 @@ where
             last_commit_proof,
         } = state;
         let last_commit_loc = Location::new(*leaf_count - 1);
-        let Operation::Commit(last_commit_metadata, op_inactivity_floor_loc) = last_commit_op
-        else {
+        let Operation::Commit(last_commit_metadata, _) = last_commit_op else {
             return Err(Error::UnexpectedData(last_commit_loc));
         };
-        if op_inactivity_floor_loc != inactivity_floor_loc {
-            return Err(Error::DataCorrupted(
-                "validated compact state floor mismatch",
-            ));
-        }
         let commit_codec_config = config.commit_codec_config.clone();
         let last_commit_op_bytes =
             Operation::<F, K, V>::Commit(last_commit_metadata.clone(), inactivity_floor_loc)
--- a/storage/src/qmdb/immutable/sync/mod.rs
+++ b/storage/src/qmdb/immutable/sync/mod.rs
@@ -186,9 +186,10 @@ where
             last_commit_proof,
         } = state;
         let last_commit_loc = Location::new(*leaf_count - 1);
-        let Operation::Commit(last_commit_metadata, _) = last_commit_op else {
+        let Operation::Commit(last_commit_metadata, op_floor) = last_commit_op else {
             return Err(Error::UnexpectedData(last_commit_loc));
         };
+        assert_eq!(op_floor, inactivity_floor_loc, "inactivity floor mismatch");
         let commit_codec_config = config.commit_codec_config.clone();
         let last_commit_op_bytes =
             Operation::<F, K, V>::Commit(last_commit_metadata.clone(), inactivity_floor_loc)
--- a/storage/src/qmdb/immutable/sync/mod.rs
+++ b/storage/src/qmdb/immutable/sync/mod.rs
@@ -172,8 +172,9 @@ where
     async fn from_compact_state(
         context: Self::Context,
         config: Self::Config,
-        state: sync::compact::State<Self::Family, Self::Op, Self::Digest>,
+        state: sync::compact::VerifiedState<Self::Family, Self::Op, Self::Digest>,
     ) -> Result<Self, Error<F>> {
+        let (state, root, inactivity_floor_loc) = state.into_parts();
         let sync::compact::State {
             leaf_count,
             pinned_nodes,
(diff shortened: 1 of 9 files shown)


storage/src/qmdb/immutable/sync/mod.rs
```diff
@@ -172,36 +172,37 @@ where
     async fn from_compact_state(
         context: Self::Context,
         config: Self::Config,
-        state: sync::compact::State<Self::Family, Self::Op, Self::Digest>,
+        state: sync::compact::VerifiedState<Self::Family, Self::Op, Self::Digest>,
     ) -> Result<Self, Error<F>> {
+        let (state, root, inactivity_floor_loc) = state.into_parts();
```
patrick-ogrady: Do we need this `into_parts` break? can we just access directly?

storage/src/qmdb/sync/compact.rs
```diff
@@ -374,37 +403,44 @@ where
     target
         .validate()
         .map_err(|reason| Error::Engine(EngineError::InvalidCompactTarget(reason)))?;
-    let FetchResult { state, callback } = config
-        .resolver
-        .get_compact_state(target.clone())
+
+    loop {
```
patrick-ogrady, line 407: Why do we loop here?

storage/src/qmdb/sync/compact.rs
```diff
@@ -417,7 +453,7 @@ where
     DB: Database,
 {
     if state.leaf_count != target.leaf_count {
-        return Err(BuildError::Engine(EngineError::UnexpectedLeafCount {
+        return Err(BuildError::InvalidPeer(EngineError::UnexpectedLeafCount {
```
patrick-ogrady: InvalidProof?

storage/src/qmdb/sync/compact.rs
```diff
@@ -449,9 +491,60 @@ where
...
+    EngineError<<DB as Database>::Family, <DB as Database>::Digest>,
+>;
+
+fn validate_compact_frontier<DB>(
+    target: &Target<DB::Family, DB::Digest>,
+    state: State<DB::Family, DB::Op, DB::Digest>,
+) -> CompactFrontierValidation<DB>
+where
+    DB: Database,
+{
```
patrick-ogrady, line 494: Add some comments around important code blocks

storage/src/qmdb/sync/compact.rs
```diff
@@ -449,9 +491,60 @@ where
     Ok(db)
 }
 
+type CompactFrontierValidation<DB> = Result<
+    VerifiedState<<DB as Database>::Family, <DB as Database>::Op, <DB as Database>::Digest>,
+    EngineError<<DB as Database>::Family, <DB as Database>::Digest>,
+>;
+
+fn validate_compact_frontier<DB>(
```
patrick-ogrady, line 488: Add a function comment

patrick-ogrady, line 488: Add a function comment

storage/src/qmdb/sync/compact.rs
patrick-ogrady, line 494: Add some comments around important code blocks

storage/src/qmdb/sync/compact.rs
```diff
@@ -449,9 +491,60 @@ where
     Ok(db)
 }
 
+type CompactFrontierValidation<DB> = Result<
```
patrick-ogrady, line 482: Add a type comment

storage/src/qmdb/sync/compact.rs
```diff
@@ -449,9 +491,60 @@ where
...
+        return Err(EngineError::RootMismatch {
+            expected: target.root,
+            actual,
+        });
+    }
+
+    Ok(VerifiedState::new(state, target.root, inactivity_floor_loc))
+}
+
+fn compact_merkle_error_to_engine<F: Family, D: Digest>(
```
patrick-ogrady: I don't think this is necessary?

storage/src/qmdb/sync/compact.rs
```diff
@@ -449,9 +491,60 @@ where
...
+    _err: crate::merkle::Error<F>,
+) -> EngineError<F, D> {
+    EngineError::InvalidProof
+}
+
 enum BuildError<F: Family, D: Digest> {
     Database(qmdb::Error<F>),
     Engine(EngineError<F, D>),
+    InvalidPeer(EngineError<F, D>),
```
patrick-ogrady: InvalidProof?

patrick-ogrady (commented):
almost there

storage/src/qmdb/keyless/sync/mod.rs
```diff
@@ -147,36 +147,41 @@ where
...
             last_commit_op,
             last_commit_proof,
         } = state;
         let last_commit_loc = Location::new(*leaf_count - 1);
-        let Operation::Commit(last_commit_metadata, inactivity_floor_loc) = last_commit_op else {
+        let Operation::Commit(last_commit_metadata, op_inactivity_floor_loc) = last_commit_op
+        else {
             return Err(qmdb::Error::UnexpectedData(last_commit_loc));
         };
+        if op_inactivity_floor_loc != inactivity_floor_loc {
```
danlaine: I think this is always false by construction:

```
  ▎The op_inactivity_floor_loc != 
  ▎ inactivity_floor_loc check at 
  ▎ storage/src/qmdb/keyless/sync/mod.rs:168 (and
  ▎  the parallel one in immutable) cannot be 
  ▎ triggered by anything a peer sends.

  How the two values are produced

  Both values originate from the same field of 
  the same Operation value. Here's the trace.

  Value A — inactivity_floor_loc from 
  ValidatedState:

  In validate_compact_frontier
  (storage/src/qmdb/sync/compact.rs:522):

  let Some(inactivity_floor_loc) =
  DB::inactivity_floor(&state.last_commit_op)
  else {
      return Err(EngineError::InvalidProof);
  };
  // ...
  Ok(ValidatedState::new(state, target.root,
  inactivity_floor_loc))

  state is moved into ValidatedState, so
  state.last_commit_op and the
  inactivity_floor_loc derived from it travel
  together. No clone, no transformation in
  between.

  Value B — op_inactivity_floor_loc from 
  destructuring in from_compact_state:

  let sync::compact::State { ..., last_commit_op,
   ... } = state;  // same op as above
  let Operation::Commit(last_commit_metadata,
  op_inactivity_floor_loc) = last_commit_op else
  { ... };

  last_commit_op here is the exact Operation
  instance that DB::inactivity_floor was called
  on. op_inactivity_floor_loc is the Location
  inside the Commit variant.

  What does DB::inactivity_floor do for the two 
  impls?

  Both impls
  (storage/src/qmdb/keyless/sync/mod.rs:197-199
  and
  storage/src/qmdb/immutable/sync/mod.rs:222-224)
   are a direct passthrough:

  fn inactivity_floor(op: &Self::Op) ->
  Option<Location<Self::Family>> {
      op.has_floor()
  }

  What does op.has_floor() do?

  For keyless (storage/src/qmdb/keyless/operation
  /mod.rs:54-60):

  pub const fn has_floor(&self) ->
  Option<Location<F>> {
      match self {
          Self::Commit(_, loc) => Some(*loc),
          // ...
      }
  }

  Same pattern in immutable. It literally returns
   the inner loc of the Commit variant.

  Where the chain forces equality

  So for value A and value B to differ, we'd need
   either:

  1. op.has_floor() to return something other
  than the Commit's inner loc — but it doesn't;
  the function body is match { Commit(_, loc) => 
  Some(*loc) }.
  2. DB::inactivity_floor to do something other
  than op.has_floor() — but both implementations
  are a one-line passthrough.
  3. The Operation value to change between the
  two reads — but it's a plain enum moved through
   ValidatedState with no interior mutability.
  4. The op to not be a Commit — but in that
  branch op.has_floor() returns None, the let 
  Some(...) = ... early-returns InvalidProof in
  validate_compact_frontier, and we never reach
  from_compact_state at all.

  A peer cannot send anything that triggers
  Value-A ≠ Value-B. They only control the
  Operation bytes; the function chain above
  derives both values from the same source.
  ```

patrick-ogrady: This was a remnant of an earlier refactor. Good catch.

patrick-ogrady:
bugbot run

patrick-ogrady:
bugbot run

patrick-ogrady:
bugbot run

patrick-ogrady:
I don't love the name `from_validated_state` but we can clean this up later.

PR #3900, opened 2026-05-28
[Storage] [Runtime] Add sealed blob type

danlaine:
# Sealed sections for contiguous journals

## Motivation

The contiguous journal has an invariant that only the newest (tail) section is writable; every historical section is full and immutable. The type system doesn't enforce this. Every blob is `Append<B>` — a writable wrapper carrying a write buffer, an `RwLock` around blob state, and partial-page CRC bookkeeping — even though historical sections never receive writes.

This has three costs:

1. **Memory.** A journal with N sections holds N write buffers, N locks, N pieces of partial-page state. Only the tail ever uses any of it.
2. **Read-path overhead.** Reads against historical sections still acquire the writer's `RwLock` just to snapshot the blob/buffer state. The page cache is fast; the per-call async lock acquisition isn't free.
3. **Correctness via convention.** Nothing in the type system prevents a stray `append` against a historical section.

This PR encodes the sealed/tail distinction in the type system.

### Runtime: `Sealed<B>`

A new read-only paged-blob type in `runtime/utils/buffer/paged`, the counterpart to `Append<B>`. It exposes the same read APIs (`read_at`, `read_into`, `read_up_to`, `read_many_into`, `try_read_sync`, `replay`) but:

- No write buffer and no `RwLock` — just an `Arc<SealedInner>`.
- Cheaply cloneable; clones share the same immutable view.
- Reuses the page-cache identity of the originating `Append`, so hot pages stay hot.

`Append::seal(self) -> Sealed<B>` syncs pending writes, captures the partial tail page in memory (the page cache only holds full pages), and returns the read-only view. A `Sealed::open` path also exists for cold-start loads. Both share a `read_and_trim` helper for trailing-junk recovery.

### Storage: typed section store

A new `contiguous::sections` module introduces a two-phase store:

```text
SectionsInit (every section open as Append)
   │
   ▼  into_sections(tail) or reset(tail)
Sections (historical: Sealed, tail: Append)
```

`SectionsInit` is the recovery phase: callers can read or truncate sections before sealing. `into_sections` consumes it, seals every section that is not the chosen tail, and returns a `Sections` whose historical entries are `Sealed<B>` handles. `Sections` itself has no API for mutating historical sections.

`contiguous::fixed::Journal` is rewritten on top of `Sections` directly, no longer wrapping `segmented::fixed::Journal`. `contiguous::variable::Journal` gets a sibling `VariableData<E,V>` adapter for variable-length items.

### Shared variable-length helpers

A new `journal/variable_format.rs` owns the varint+codec helpers (`ItemInfo`, `decode_length_prefix`, `find_item`, `decode_item`) previously duplicated between `segmented::variable` and `contiguous::variable_data`. The on-disk format is unchanged.

## Why this is beneficial

1. **Type-level safety.** Historical sections cannot be mutated by mistake; the API doesn't exist.
2. **Lower memory.** Per-section write buffers collapse from N to 1.
3. **Lock-free historical reads.** `Sealed` clones share an `Arc<SealedInner>`; no writer lock to coordinate with.
4. **Hot-page continuity across rollover.** `Append::seal` hands off cache identity so the page cache stays warm.
5. **Cleaner layering.** Contiguous journals no longer contort themselves through `SegmentedJournal`'s "any section is mutable" abstraction. Reads, replays, and rollover paths become explicit in the contiguous layer rather than delegated to a more general type that doesn't share the invariant.
6. **Tighter adversarial posture.** The defensive arithmetic fixes close real bug surfaces around stale offsets and corrupt blob names.

## Future improvements this enables

- **Bounded immutable read snapshots.** `Sealed` is suitable for handing out as a long-lived read view: deferred replay over a frozen state, fork-style readers, or sharing across tasks without copying.
- **Independent evolution of segmented and contiguous.** The contiguous layer no longer depends on `segmented::fixed` at all, and on `segmented::variable` only for the now-shared format helpers. The two can be tuned independently.
- **Seal-on-prune in other journals.** The same pattern — `Sealed` for frozen data, `Append` for the tip — is reusable anywhere there is a clear "no more writes past point X" boundary.
- **Read-side scaling.** Removing writer-lock coordination on historical reads makes it easier to scale concurrent readers without touching the write path.


danlaine:
Going to break this down into smaller PRs

PR #3903, opened 2026-05-28
[ci] update dylint to 6.0.1

clabby:
## Overview

Updates `cargo-dylint` and `dylint-link` to `v6.0.1`. Also takes advantage of the release artifacts published for these binaries rather than building them from source on every run: https://github.com/trailofbits/dylint/pull/1972

### DIFF
--- a/.github/workflows/fast.yml
+++ b/.github/workflows/fast.yml
@@ -52,9 +52,10 @@ jobs:
         tool: just@1.43.0
     - name: Install Dylint
       if: matrix.os == 'ubuntu-latest' && matrix.flags == ''
-      run: |
-        cargo install cargo-dylint --version 6.0.0 --locked
-        cargo install dylint-link --version 6.0.0 --locked
+      uses: taiki-e/install-action@0c5db7f7f897c03b771660e91d065338615679f4 # v2.60.0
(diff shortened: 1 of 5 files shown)


.github/workflows/fast.yml
```diff
@@ -51,10 +51,16 @@ jobs:
...
+      with:
+        tool: cargo-dylint@6.0.1,dylint-link@6.0.1
+    - name: Clone Dylint source
       if: matrix.os == 'ubuntu-latest' && matrix.flags == ''
       run: |
-        cargo install cargo-dylint --version 6.0.0 --locked
-        cargo install dylint-link --version 6.0.0 --locked
+        set -euo pipefail
+        mkdir -p /home/runner/work/dylint
+        git clone --depth 1 --branch v6.0.1 https://github.com/trailofbits/dylint.git /home/runner/work/dylint/dylint
```
clabby, line 63: Note that this is a temporary workaround for a detail I overlooked upstream. Fix in https://github.com/trailofbits/dylint/pull/1985.

patrick-ogrady: I feel its worth delaying this PR until this is addressed?

PR #3906, opened 2026-05-28
[storage/journal] fix recovery with multiple empty trailing data sections

andresilva:
Variable journal recovery could panic when normal append flow left multiple empty data section blobs behind after unsynced tail appends. Recovery now trims empty trailing data sections before deciding whether the data journal is empty or rebuilding offsets, so durable prefixes are preserved and empty tails are discarded.

Found by fuzzing in https://github.com/commonwarexyz/monorepo/actions/runs/26581984886/job/78317247546?pr=3878.

### DIFF
--- a/storage/src/journal/contiguous/variable.rs
+++ b/storage/src/journal/contiguous/variable.rs
@@ -977,9 +977,9 @@ impl<E: Context, V: CodecShared> Journal<E, V> {
                 break items_in_last_section;
             }
 
-            let previous_section = last_section.checked_sub(1).ok_or_else(|| {
-                Error::Corruption("multiple data sections with no previous section".into())
-            })?;
+            let previous_section = last_section
+                .checked_sub(1)
+                .expect("num_sections >= 2 implies newest_section >= 1");
             let previous_size = data.size(previous_section).await?;
             warn!(
                 section = last_section,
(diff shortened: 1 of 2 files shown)


storage/src/journal/contiguous/variable.rs
```diff
@@ -943,20 +949,34 @@ impl<E: Context, V: CodecShared> Journal<E, V> {
...
+
+            // Stop once we find replayable data or only one empty section
+            // remains for the existing empty-data repair path.
+            if items_in_last_section > 0 || data.num_sections() == 1 {
+                break items_in_last_section;
+            }
+
+            let previous_section = last_section.checked_sub(1).ok_or_else(|| {
+                Error::Corruption("multiple data sections with no previous section".into())
+            })?;
```
cronokirby: I think we should panic rather than return an error here, because we should only encounter this if there's a bug in the implementation. The last section is the highest key in the BTreeMap, so it's not possible for num(keys) >= 2 but max(keys) == 0. So I think this should never happen.

patrick-ogrady: AFAIK we let the caller panic based on the return of init here...I think this could still happen if corruption (even with a correct impl)?

storage/src/journal/contiguous/variable.rs
patrick-ogrady, line 1: will test shortly but noting here: does this also affect fixed?

storage/src/journal/contiguous/variable.rs
patrick-ogrady: I think I'm wrong. Fixing.

patrick-ogrady (approved):

PR #3908, opened 2026-05-28
[Runtime] Add `Sealed` blob

danlaine:
## Summary

Adds `Sealed`, a read-only counterpart to the paged `Append` buffer.

`Sealed` is meant for blobs that have left the writable tail and should only be read from that point forward. It keeps the same page-cache-backed read surface that the journal needs, without exposing append, resize, or write operations.

## Changes

- Adds `Sealed<B>` with `read_at`, `read_into`, `read_many_into`, `try_read_sync`, `replay`, `size`, and `sync`.
- Adds `Append::seal(self)`, which requires unique ownership, flushes buffered bytes, and returns a `Sealed` view over the same blob/cache identity.
- Keeps seal cheap by flushing but not fsyncing; durability remains the caller's responsibility through a later `sync`.
- Preserves reads across a partial final page by keeping that page fragment in the sealed handle.
- Adds shared validation for paged `read_many_into` calls, including buffer length, sorted offsets, non-overlap, overflow, and bounds checks.
- Adds exact cache-range helpers so synchronous reads fail without partially mutating caller buffers on a cache miss.
- Adds `NotUnique` and `InvalidInput` runtime errors for the new ownership and validation paths.


runtime/src/utils/buffer/paged/append.rs
```diff
@@ -605,39 +605,23 @@ impl<B: Blob> Append<B> {
         offsets: &[u64],
         item_size: usize,
```
roberto-bayardo: I wonder if this should be NonZeroUsize?

runtime/src/utils/buffer/paged/sealed.rs
```diff
@@ -0,0 +1,895 @@
...
+        offsets: &[u64],
+        item_size: usize,
+    ) -> Result<(), Error> {
+        super::validate_read_many_into(buf.len(), offsets, item_size, self.inner.size)?;
+        if offsets.is_empty() || item_size == 0 {
+            return Ok(());
+        }
+
+        let partial_offset = self.partial_offset();
+        let mut cache_ranges: Vec<(&mut [u8], u64)> = Vec::new();
```
roberto-bayardo: ::with_capacity(offsets.len()) ?

runtime/src/utils/buffer/paged/cache.rs
```diff
@@ -246,6 +246,30 @@ impl CacheRef {
         original_len - buf.len()
     }
 
+    /// Try to read the specified bytes from the page cache only. Returns `false` without modifying
+    /// `buf` unless every requested byte is already cached.
```
roberto-bayardo: Is the "without modifying buf" property important?  This doubles the hashmap lookups from previous.

runtime/src/utils/buffer/paged/append.rs
```diff
@@ -1214,6 +1198,64 @@ impl<B: Blob> Append<B> {
...
+        let full_pages = blob_state.current_page;
+        debug_assert_eq!(
+            full_pages.checked_mul(logical_page_size),
+            Some(buffer.offset)
+        );
+        let partial_page = if buffer.is_empty() {
+            None
+        } else {
+            Some(buffer.slice(..buffer.len()))
+        };
```
roberto-bayardo, line 1385: Might be worth allocating a precisely sized buffer for the partial page so we can release the (often much larger) append buffer.

```suggestion
  let partial_page = if buffer.is_empty() {
      None
  } else {
      let mut copy = self.cache_ref.pool().alloc(buffer.len());
      copy.put_slice(buffer.as_ref());
      Some(copy.freeze())
  };
```

runtime/src/utils/buffer/paged/sealed.rs
```diff
@@ -0,0 +1,895 @@
...
+        let mut buf = unsafe { self.inner.cache_ref.pool().alloc_len(len) };
+        self.read_into(buf.as_mut(), offset).await?;
+        Ok(buf.into())
+    }
+
+    /// Read into `buf` if it can be done synchronously without I/O. Returns `true` only if all
+    /// `buf.len()` bytes were satisfied from either the page cache or the in-memory partial page.
+    pub fn try_read_sync(&self, offset: u64, buf: &mut [u8]) -> bool {
+        // Reject out-of-bounds requests up front so we don't consult the cache (which may hold
+        // pages cached past `size` if a sibling Append handle wrote them).
```
roberto-bayardo: I find the "which may hold ..." part here confusing (a sibling append shouldn't exist right?)

roberto-bayardo: I also don't think there can be pages cached past `size` any longer, they should all be invalidated.

runtime/src/utils/buffer/paged/append.rs
```diff
@@ -1214,6 +1198,64 @@ impl<B: Blob> Append<B> {
...
+            self.flush_internal(buf_guard, true, false).await?;
+        }
+
+        // After flush, the buffer holds at most the partial-page logical bytes and `blob_state`
+        // reflects the on-disk state of that page.
+        let buffer = self.buffer.read().await;
+        let blob_state = self.blob_state.read().await;
+        let logical_page_size = self.cache_ref.page_size();
+        let full_pages = blob_state.current_page;
+        debug_assert_eq!(
```
roberto-bayardo: I'd make this a regular assert.

runtime/src/utils/buffer/paged/append.rs
```diff
@@ -1214,6 +1198,64 @@ impl<B: Blob> Append<B> {
...
+        let partial_page = if buffer.is_empty() {
+            None
+        } else {
+            Some(buffer.slice(..buffer.len()))
+        };
+        let partial_len = partial_page.as_ref().map_or(0, |p| p.len() as u64);
+        let size = full_pages
+            .checked_mul(logical_page_size)
+            .and_then(|size| size.checked_add(partial_len))
+            .ok_or(Error::OffsetOverflow)?;
```
roberto-bayardo:
```suggestion
        let size = buffer.size();
```

runtime/src/utils/buffer/paged/sealed.rs
```diff
@@ -0,0 +1,895 @@
...
+            (physical, logical)
+        } else {
+            let physical = physical_page_size
+                .checked_mul(full_pages)
+                .ok_or(Error::OffsetOverflow)?;
+            let logical = logical_page_size
+                .checked_mul(full_pages)
+                .ok_or(Error::OffsetOverflow)?;
+            (physical, logical)
+        };
```
roberto-bayardo: This can be simplified a good deal... e.g. we already know logical blob size: `self.inner.size`.

then for physical pages:

```
  let full_pages = (self.inner.size - partial_len) / logical_page_size;
  let pages = full_pages + u64::from(partial_len > 0);
...
```

runtime/src/utils/buffer/paged/sealed.rs
```diff
@@ -0,0 +1,895 @@
...
+            } else {
+                let prefix_len = (partial_offset - offset) as usize;
+                let partial = self
+                    .inner
+                    .partial_page
+                    .as_ref()
+                    .expect("partial bytes exist when end > partial_offset");
+                item_buf[prefix_len..].copy_from_slice(&partial.as_ref()[..item_size - prefix_len]);
+                cache_ranges.push((&mut item_buf[..prefix_len], offset));
+            }
```
roberto-bayardo: Code is duplicated in append.rs ... pull this out into a shared `split_read_many` ?

runtime/src/utils/buffer/paged/sealed.rs
danlaine: re-factored this

runtime/src/utils/buffer/paged/sealed.rs
danlaine: factored out `split_read_many`

runtime/src/utils/buffer/paged/sealed.rs
danlaine: updated comments

roberto-bayardo (dismissed):

patrick-ogrady:

PR #3909, opened 2026-05-28
[Storage] Use `Sealed` blob in fixed contiguous journal

danlaine:
## Summary

Moves the fixed contiguous journal onto a typed section store that encodes the section lifecycle directly: historical sections are `Sealed`, and only the tail remains writable.

This keeps the existing journal API and section layout, but makes the "only the tail receives writes" invariant explicit in the storage layer instead of relying on every caller to preserve it manually.

## Changes

- Adds `contiguous::sections`, with two states:
  - `SectionsInit` for recovery/open, where every section is writable so init can inspect and repair blobs.
  - `Sections` for steady state, where historical sections are sealed and at most one tail is an `Append`.
- Refactors the fixed contiguous journal to route reads, replay, append, sync, prune, rewind, clear, and destroy through `Sections`.
- Rolls a full tail by sealing it without fsyncing, then opening the next tail; journal commit/sync continues to make dirty sections durable.
- Tracks dirty work by earliest dirty section so a section remains eligible for sync after it rolls from tail to sealed history.
- Keeps recovery repair in the init phase, then transitions to sealed steady state after trailing bytes, gaps, and tail placement have been resolved.
- Adds section-store metrics for tracked, synced, and pruned sections.
- Removes the segmented fixed journal helper that existed only to force-create an empty section.

### DIFF
--- a/storage/src/journal/contiguous/fixed.rs
+++ b/storage/src/journal/contiguous/fixed.rs
@@ -655,39 +655,30 @@ impl<E: Context, A: CodecFixedShared> Journal<E, A> {
         );
     }
 
-    /// Update pruning-boundary and recovery-watermark entries in metadata's in-memory state.
+    /// Stage pruning-boundary and recovery-watermark entries into `metadata`'s in-memory state.
     ///
-    /// Call `inner.metadata.sync()` separately to persist the updated entries.
-    fn update_metadata_entries(
-        inner: &mut Inner<E, A>,
+    /// Only writes when a value actually changes. The caller is responsible for syncing.
+    fn stage_metadata_entries(
+        metadata: &mut Metadata<E, u64, VecU64>,
         items_per_blob: u64,
         pruning_boundary: u64,
         recovery_watermark: u64,
     ) {
-        let current_pruning = inner
-            .metadata
-            .get(&PRUNING_BOUNDARY_KEY)
-            .copied()
-            .map(u64::from);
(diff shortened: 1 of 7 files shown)


roberto-bayardo:
Consider a helper such as:

```
  enum SectionRef<'a, B: Blob> {
      Sealed(&'a Sealed<B>),
      Tail(&'a Append<B>),
  }

  impl<E: Context> Sections<E> {
      /// Locate `section`, enforcing the prune guard.
      fn get(&self, section: u64) -> Result<Option<SectionRef<'_, E::Blob>>, Error> {
          self.core.prune_guard(section)?;
          if let Some(s) = self.sealed.get(&section) {
              return Ok(Some(SectionRef::Sealed(s)));
          }
          if let Some(t) = &self.tail {
              if t.section == section {
                  return Ok(Some(SectionRef::Tail(&t.blob)));
              }
          }
          Ok(None)
      }
  }
```

  Then each public method shrinks to its actual logic:
  
```
  pub(super) async fn read_at(&self, section: u64, offset: u64, len: usize) -> Result<IoBufs, Error> {
      match self.get(section)? {
          Some(SectionRef::Sealed(s)) => s.read_at(offset, len).await.map_err(Error::Runtime),
          Some(SectionRef::Tail(t)) => t.read_at(offset, len).await.map_err(Error::Runtime),
          None => Err(Error::SectionOutOfRange(section)),
      }
  }
```

danlaine:
Closing in favor of #3970

PR #3910, opened 2026-05-28
[Storage] Use `Sealed` blob in contiguous variable journal

danlaine:
## Summary

Moves the contiguous variable journal's data backend onto the same sealed section store used by the fixed journal, and factors variable-item parsing into shared format helpers.

The data journal remains the source of truth. The offsets journal is still the index used for position lookups, but init can now align or rebuild it from the recovered data sections.

## Changes

- Adds `VariableData`, an item-aware adapter over `Sections` for variable-length data sections.
- Refactors the contiguous variable journal to store data in sealed historical sections plus one writable tail.
- Appends data first, records the corresponding offsets afterward, and tracks dirty data sections separately from the offsets journal.
- Makes commit/sync flush dirty data sections before committing/syncing offsets, so a committed offset never points at data that was intentionally skipped.
- Reworks init-time alignment so offsets can be pruned, rewound, or rebuilt from data when a crash leaves the two substrates out of sync.
- Adds `variable_format`, a shared implementation for variable item encoding, decoding, synchronous reads, consecutive batch reads, and replay scanning.
- Updates the segmented variable journal to use the shared format helpers instead of maintaining a separate parser.

### DIFF
--- a/storage/src/journal/contiguous/mod.rs
+++ b/storage/src/journal/contiguous/mod.rs
@@ -13,6 +13,7 @@ pub mod fixed;
 mod metrics;
 mod sections;
 pub mod variable;
+mod variable_data;
 
 #[cfg(test)]
 mod tests;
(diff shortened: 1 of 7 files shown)


danlaine:
Closing in favor of new branch built off #3970

PR #3912, opened 2026-05-29
[storage/qmdb] generate test per variant

andresilva:
Generate one test per variant rather than having a single test with all the variants being tested. This gives us more test parallelism which plays better with CI partitioning.

Extracted from #3878.

### DIFF
--- a/storage/src/qmdb/current/mod.rs
+++ b/storage/src/qmdb/current/mod.rs
@@ -1452,29 +1452,6 @@ pub mod tests {
         };
     }
 
-    // Runner macros - receive common args followed by (label, type, config).
-    macro_rules! test_simple {
-        ($f:expr, $l:ident, $db:ty, $cfg:ident) => {
-            Box::pin(async {
-                $f(open_db_fn!($db, $cfg));
-            })
-            .await
-        };
-    }
-
-    // Macro to run a test on DB variants.
-    macro_rules! for_all_variants {
-        (simple: $f:expr) => {{
-            with_all_variants!(test_simple!($f));
-        }};
-        (ordered: $f:expr) => {{
-            with_ordered_variants!(test_simple!($f));
-        }};
-        (unordered: $f:expr) => {{
-            with_unordered_variants!(test_simple!($f));
-        }};
-    }
-
     // Emit one `#[test_group("slow")] #[test_traced]` test per variant.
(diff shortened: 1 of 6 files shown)


storage/src/qmdb/current/mod.rs
```diff
@@ -1387,90 +1389,109 @@ pub mod tests {
...
-                $f(open_db_fn!($db, $cfg));
-            })
-            .await
+    // Emit one `#[test_group("slow")] #[test_traced]` test per variant.
+    // The fn name is `<f>_<variant_label>`, the body opens the variant's DB and
+    // runs the test function `f`.
+    macro_rules! test_for_variant {
+        ($f:ident, $traced:literal, $label:ident, $db:ty, $cfg:ident) => {
+            paste::paste! {
+                #[test_group("slow")]
```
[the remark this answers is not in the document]
andresilva: This was intentional so that we could reuse the macro without adding extra complexity to it.

storage/src/qmdb/current/mod.rs
```diff
@@ -1387,90 +1389,109 @@ pub mod tests {
...
+                #[test_traced($traced)]
+                fn [<$f _ $label>]() {
+                    let executor = deterministic::Runner::default();
```
danlaine, line 1450: the test fns already create a runner
```git
diff --git a/storage/src/qmdb/current/mod.rs b/storage/src/qmdb/current/mod.rs
index 54d64ebdc..66d107f32 100644
--- a/storage/src/qmdb/current/mod.rs
+++ b/storage/src/qmdb/current/mod.rs
@@ -1461,13 +1461,7 @@ pub mod tests {
                 #[test_group("slow")]
                 #[test_traced($traced)]
                 fn [<$f _ $label>]() {
-                    let executor = deterministic::Runner::default();
-                    executor.start(|_context| async move {
-                        Box::pin(async {
-                            $f(open_db_fn!($db, $cfg));
-                        })
-                        .await
-                    });
+                    $f(open_db_fn!($db, $cfg));
                 }
             }
         };
```

andresilva: I wonder if it would be cleaner to instead change the tests to take `Context` and keep this as is.

danlaine: I think you're right -- that'd be consistent with `any`

danlaine (approved):

PR #3915, opened 2026-05-30
[deployer] Allow collocating instances in the same AZ

clabby:
Adds an optional `availability_zone_group` field to `InstanceConfig`. Instances that set the same group string are launched in one mutually-supported availability zone within their region.

- Group AZ is the first zone supporting every instance type in the group.
- The same group can be used in different AZs.
- Instances without a group keep the existing round-robin AZ distribution.

*This description was generated by Claude Code.*


clabby:
bugbot run

patrick-ogrady (approved):

PR #3917, opened 2026-05-31
[glue/stateful] Add floor discovery protocol

clabby:
## Overview

Adds a protocol for safely discovering a floor `Finalization` from peers to enable state sync. Currently, this is expected to be provided out of band. While some nodes may prefer a trusted protocol (i.e. pulling from exoware), this option enables trustlessly bootstrapping state sync from the active participant set.

Protocol docs: https://github.com/commonwarexyz/monorepo/blob/0814f9e1dcbfa57489edabd92ad6e2268b996c16/glue/src/stateful/floor_discovery/mod.rs#L1-L98

### DIFF
--- a/actor/src/mailbox.rs
+++ b/actor/src/mailbox.rs
@@ -308,6 +308,11 @@ pub struct UnreliableReceiver<T: UnreliablePolicy> {
 }
 
 impl<T: Policy> Receiver<T> {
+    /// Returns whether senders can still enqueue messages.
+    pub fn is_open(&self) -> bool {
+        !self.state.is_disconnected()
+    }
+
     /// Receive the next message.
     ///
     /// Returns `None` after all senders are dropped and all buffered messages
@@ -326,6 +331,11 @@ impl<T: Policy> Receiver<T> {
 }
 
 impl<T: UnreliablePolicy> UnreliableReceiver<T> {
+    /// Returns whether senders can still enqueue messages.
+    pub fn is_open(&self) -> bool {
+        !self.state.is_disconnected()
+    }
+
     /// Receive the next message.
     ///
     /// Returns `None` after all senders are dropped and all buffered messages
@@ -1115,6 +1125,22 @@ mod tests {
     }
 
     #[test]
+    fn unreliable_receiver_is_open_while_sender_is_connected() {
+        let (sender, mut receiver) = new_unreliable(NZUsize!(1));
+        assert!(receiver.is_open());
+        assert_eq!(
+            sender.enqueue(Message::Vote(1)),
+            Unreliable::new(Feedback::Ok)
+        );
+
+        drop(sender);
+
+        assert!(!receiver.is_open());
+        assert_eq!(receiver.try_recv(), Ok(Message::Vote(1)));
+        assert_eq!(receiver.try_recv(), Err(TryRecvError::Disconnected));
+    }
+
+    #[test]
     fn poll_recv_drains_buffered_messages_after_senders_drop() {
         let (sender, receiver) = new_unreliable(NZUsize!(1));
         let wakes = Arc::new(WakeCounter::default());
@@ -1468,6 +1494,19 @@ mod tests {
         assert_eq!(wakes.count(), 1);
         assert_eq!(receiver.try_recv(), Ok(SpillMessage::FillReady));
         assert_eq!(receiver.try_recv(), Ok(SpillMessage::Spill));
+    }
+
+    #[test]
+    fn reliable_receiver_is_open_while_sender_is_connected() {
+        let (sender, mut receiver) = new(NZUsize!(1));
+        assert!(receiver.is_open());
+        assert_eq!(sender.enqueue(SpillMessage::FillReady), Feedback::Ok);
+
+        drop(sender);
+
+        assert!(!receiver.is_open());
+        assert_eq!(receiver.try_recv(), Ok(SpillMessage::FillReady));
+        assert_eq!(receiver.try_recv(), Err(TryRecvError::Disconnected));
     }
 }
 
--- a/actor/src/mailbox.rs
+++ b/actor/src/mailbox.rs
@@ -308,9 +308,9 @@ pub struct UnreliableReceiver<T: UnreliablePolicy> {
 }
 
 impl<T: Policy> Receiver<T> {
-    /// Returns whether senders can still enqueue messages.
-    pub fn is_open(&self) -> bool {
-        !self.state.is_disconnected()
+    /// Returns whether senders can no longer enqueue messages.
+    pub fn is_closed(&self) -> bool {
+        self.state.is_disconnected()
     }
 
     /// Receive the next message.
@@ -331,9 +331,9 @@ impl<T: Policy> Receiver<T> {
 }
 
 impl<T: UnreliablePolicy> UnreliableReceiver<T> {
-    /// Returns whether senders can still enqueue messages.
-    pub fn is_open(&self) -> bool {
-        !self.state.is_disconnected()
+    /// Returns whether senders can no longer enqueue messages.
+    pub fn is_closed(&self) -> bool {
+        self.state.is_disconnected()
     }
 
     /// Receive the next message.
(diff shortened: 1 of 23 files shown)


clabby:
bugbot run

glue/src/stateful/probe/actor/discovery.rs
```diff
@@ -0,0 +1,257 @@
...
+                });
+        let Some(floor) = floor else {
+            return;
+        };
+        let Some(sample_size) = self.sample_size(floor.epoch()) else {
+            return;
+        };
+        if replies < sample_size {
+            return;
+        }
```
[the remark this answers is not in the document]
clabby: This isn't a valid issue. We want `f+1` from the winning certificate's epoch's max fault configuration.

clabby:
bugbot run

actor/src/mailbox.rs
```diff
@@ -308,6 +308,11 @@ pub struct UnreliableReceiver<T: UnreliablePolicy> {
 }
 
 impl<T: Policy> Receiver<T> {
+    /// Returns whether senders can no longer enqueue messages.
+    pub fn is_closed(&self) -> bool {
```
patrick-ogrady: Can't we just infer this from recv returning none?

glue/src/stateful/mod.rs
```diff
@@ -102,6 +101,7 @@ mod actor;
 pub use actor::{Config, Mailbox, Stateful, SyncPlan};
 
 pub mod db;
+pub mod floor_discovery;
```
patrick-ogrady: don't love the name but it isn't inaccurate...

glue/src/stateful/floor_discovery/actor/mod.rs
```diff
@@ -0,0 +1,118 @@
...
+}
+
+/// Discovers a sync floor by adopting the highest finalization from a peer sample.
+///
+/// The actor is a two-phase state machine. It starts in discovery, waits until a subscriber needs
+/// a floor, then solicits and samples peers' finalizations without serving any of its own. Once a
+/// marshal is attached, it hands off to serving, answering peers' requests from that marshal and
+/// never issuing outbound requests. A source node that never needed a floor attaches a marshal
+/// without consuming one and transitions straight to serving without soliciting peers.
+pub struct FloorDiscovery<E, S, D, V, T, P, B>
```
patrick-ogrady: Can just call it "initialization" or "bootstrap" or "targeting"?

glue/src/stateful/floor_discovery/actor/discovery.rs
```diff
@@ -0,0 +1,264 @@
...
+                    marshal = Some(attached);
+                }
+            },
+            Ok((peer, message)) = receiver.recv() else {
+                debug!("network receiver closed, shutting down");
+                return;
+            } => {
+                // Block peers whose payloads do not decode. Discovery cannot serve, so peer
+                // requests are ignored.
+                let cfg = <S as CertificateScheme>::certificate_codec_config_unbounded();
```
patrick-ogrady: I don't think we can use unbounded here?

we need to use the epocher like we do in marshal to do this safely (and add a fast-path for threshold).

glue/src/stateful/probe/actor/service.rs
```diff
@@ -0,0 +1,118 @@
...
+                } else {
+                    Either::Left(self.mailbox.recv())
+                };
+            },
+            on_stopped => {
+                debug!("shutdown signal received");
+            },
+            Some(message) = mailbox_message else {
+                debug!("mailbox closed");
+                continue;
```
patrick-ogrady, line 69: I think we could just return here like we do below rather than doing this unnecessary .is_closed() thing?

glue/src/stateful/floor_discovery/actor/serving.rs
```diff
@@ -0,0 +1,118 @@
...
+                    }
+                }
+                // Already serving; an additional marshal attachment is ignored.
+                Message::Attach { .. } => {}
+            },
+            Ok((peer, message)) = receiver.recv() else {
+                debug!("network receiver closed, shutting down");
+                return;
+            } => {
+                let cfg = <S as CertificateScheme>::certificate_codec_config_unbounded();
```
patrick-ogrady: We shouldn't be willing to decode any finalization here? I think we may want to break request/response into different messages?

glue/src/stateful/probe/actor/service.rs
```diff
@@ -0,0 +1,118 @@
...
+                    Recipients::One(peer),
+                    wire::Message::<S, V>::Finalization(finalization).encode(),
+                    false,
+                );
+            },
+        }
+    }
+
+    /// Fetches the latest [`Finalization`] from marshal, if available.
+    async fn produce_latest(&mut self) -> Option<Finalization<S, V::Commitment>> {
```
patrick-ogrady, line 113: I suppose we could do this async but think its fine for now.

glue/src/stateful/probe/actor/discovery.rs
patrick-ogrady, line 1: make name symmetric with "serving"?

glue/src/stateful/tests/multi_db_app.rs
```diff
@@ -646,6 +647,28 @@ impl EngineDefinition for MultiDbEngine {
         // Start marshal actor with monitored reporters.
         marshal_actor.start(marshal_reporters, buffer, resolver);
 
+        // Attach the marshal to floor discovery, transitioning it to serving. A syncing node has
+        // already consumed its floor above; a source attaches without ever soliciting peers.
+        floor_discovery_mailbox.attach(marshal_mailbox.clone());
```
patrick-ogrady: I suppose this dynamic attachment is smooth (rather than recreating)?

glue/src/stateful/tests/multi_db_app.rs
```diff
@@ -646,6 +647,28 @@ impl EngineDefinition for MultiDbEngine {
...
+            let height = block.height();
+            *self
+                .sync_entries
+                .lock()
+                .entry(public_key.clone())
+                .or_insert(0) += 1;
+            self.sync_heights
+                .lock()
+                .insert(public_key.clone(), height.get());
+            state_sync_height = Some(height.get());
```
patrick-ogrady, line 666: why do we need to wait for this here? can't we just handle when we start processing blocks (or not at all)?

patrick-ogrady (commented):
I found this pretty reasonable but a few small rough edge to buff out.

glue/src/stateful/floor_discovery/actor/mod.rs
clabby: I prefer the name `FloorDiscovery`, self-descriptive

glue/src/stateful/tests/multi_db_app.rs
clabby: Purely for the test harness. This just captures the height of the block the floor finalization commits to for the `Property`

glue/src/stateful/probe/actor/mod.rs
```diff
@@ -13,6 +10,9 @@ use discovery::Discovery;
 use rand_core::CryptoRngCore;
 use std::num::NonZeroUsize;
 
+mod discovery;
```
patrick-ogrady, line 13: How do you feel about `requester`/`responder` if you keep separate loops? `serving` sounds odd.

patrick-ogrady: ah, these are meant to represent the phases?

patrick-ogrady:
bugbot run

glue/src/stateful/probe/actor/discovery.rs
```diff
@@ -0,0 +1,292 @@
...
+                            finalization.round() > candidate.round()
+                        })
+                        .then_some(finalization)
+                        .or(floor);
+                    (floor, replies + 1)
+                });
+        let Some(floor) = floor else {
+            return;
+        };
+        let Some(sample_size) = self.sample_size(floor.epoch()) else {
```
patrick-ogrady, line 258: It feels a bit odd to use the epoch here instead of probing over the peers we must've registered to run this in the first place?

its possible this could be larger than we could fulfill with our current peers?

patrick-ogrady: Not to mention, this could quickly bias us to an old item (if the network used to be smaller)?

I suppose we assume the number of byzantine players in any historical set is <= f.

patrick-ogrady (approved):

PR #3925, opened 2026-06-01
 [consensus/simplex/twins] improve case generation algorithm

andresilva:
This PR replaces twins `cases()` generation with a compressed canonical scenario generator. Instead of expanding every concrete participant assignment, the generator tracks residual symmetry cells: groups of participants that are indistinguishable under the scripted prefix so far. Each round refines those cells by assigning contiguous groups to twins roles, and each compressed transition stores both the next residual cell state and the exact multiplicity of concrete rounds represented by that transition.

Scenario counting is then done over that compressed transition DAG. When the full space fits the campaign budget, we enumerate ranks directly. When it does not, we sample unique ranks uniformly without replacement and unrank only the selected scenarios. Compromised assignments are generated the same way: one representative per residual-cell equivalence class, preserving the same pass/fail coverage up to participant relabeling.

It keeps the existing simplex-specific semantics: primary and secondary recipient sets are independent masks, so a participant can be visible to both halves, one half, or neither half in a scripted round. This is strictly more expressive than the disjoint-partition model in the twins paper.

I also experimented with arbitrary-width bitmasks to support more than 64 participants. That worked mechanically, but it made the implementation more complex without opening up much practical coverage: the compressed scenario DAG still grows exponentially, so large validator sets quickly become too expensive to explore meaningfully. The twins paper also argues that small validator sets are enough to expose these classes of protocol bugs, so this PR keeps the simpler `u64` representation and caps generated campaigns at 64 participants.

### Results

| Case | Old time | New time | Speedup |
|---|---:|---:|---:|
| `n5 f1 r3 sampled max20` | `24.42 ms` | `0.541 ms` | `45.17x` |
| `n5 f1 r3 sustained max20` | `0.261 ms` | `0.071 ms` | `3.68x` |
| `n7 f2 r3 sampled max20` | `466.15 ms` | `0.793 ms` | `588.04x` |
| `n10 f3 r5 sustained max20` | `1.43 ms` | `0.257 ms` | `5.57x` |
| `n10 f3 r5 sampled max20` | `5m 32.77s` | `22.21 ms` | `14981.79x` |

### DIFF
--- a/consensus/src/simplex/mocks/twins.rs
+++ b/consensus/src/simplex/mocks/twins.rs
@@ -14,13 +14,17 @@
 //! Advancing one round refines those cells by splitting each group according to
 //! the role its members take in the new round.
 //!
-//! The full enumeration approach guarantees that every case within a campaign
-//! is structurally distinct -- no duplicate (scenario, compromised-assignment)
-//! pairs are ever emitted. For each canonical scenario, `cases()` computes the
-//! residual symmetry cells and generates only the unique compromised-node
-//! assignments: two assignments that differ only in which members of a cell
-//! are chosen are equivalent and collapsed to a single representative. When the
-//! case space exceeds `max_cases`, scenarios are sampled without replacement.
+//! Scenario generation guarantees that every case within a campaign is
+//! structurally distinct -- no duplicate (scenario, compromised-assignment)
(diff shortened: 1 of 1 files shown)


clabby (approved):

patrick-ogrady (approved):
My compliments to the chef.

Issue #1432, opened 2025-08-19
[storage] Review recovery procedures so that failures *during recovery* are handled


roberto-bayardo:
Database recovery procedures may themselves experience failures.  We should review each to make sure those failures won't leave the DB in an irrecoverable state.

PR #3928, opened 2026-06-01
[Storage] Durably adopt recovered contiguous data

danlaine:
## Summary

This fixes a recovery durability bug in contiguous fixed and variable journals. When init observes journal data beyond the persisted recovery watermark, that recovered suffix must be made durable before the journal records metadata or index state that depends on it.

## Bug

A successful write can leave dirty pages in the OS page cache without persisting them. If the process crashes, a later process can reopen the blob and read those dirty cached bytes. They are visible, but they are not crash-durable until the blob is synced.

The contiguous journal init path recovers bounds from readable blob contents, so it can adopt this page-cache-visible suffix as part of the journal.

The fixed journal recovered the larger size but initialized `dirty_from_section` to `None`. A later `commit()` or `sync()` could therefore skip data-section syncs and still advance metadata to the recovered size. That can make the metadata durable before the adopted data is durable.

The variable journal had the same durability ordering issue with an extra layer: init can rebuild missing offset entries by replaying data after the offsets recovery watermark, then sync the offsets journal. If the data range is not synced first, the durable offsets can point at data bytes that were only page-cache-visible.

## Fix

For fixed journals, init now marks the section containing the recovered suffix dirty whenever the recovered size is ahead of the persisted recovery watermark. The existing commit/sync path then forces those sections through `sync_section` before clearing the dirty marker or advancing recovery metadata.

For variable journals, `align_journals` now tracks the data range used to rebuild offsets and syncs the corresponding data sections before syncing the offsets journal. This preserves the required ordering: adopted data becomes durable before durable metadata or offset entries depend on it.

### DIFF
--- a/storage/src/journal/contiguous/fixed.rs
+++ b/storage/src/journal/contiguous/fixed.rs
@@ -726,13 +726,17 @@ impl<E: Context, A: CodecFixedShared> Journal<E, A> {
             meta_recovery_watermark,
         )
         .await?;
+        // Bytes beyond the persisted recovery watermark may be readable after reopen without
+        // being crash-durable, so the next commit/sync must force a data sync before advancing it.
+        let dirty_from_section =
+            (recovery_watermark < size).then_some(recovery_watermark / items_per_blob);
 
         let mut inner = Inner {
             journal,
             size,
             metadata,
             pruning_boundary,
-            dirty_from_section: None,
+            dirty_from_section,
         };
         // Persist any lowered checkpoint before applying blob repairs that move recovered state
         // backward.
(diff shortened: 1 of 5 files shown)


roberto-bayardo:
related? https://github.com/commonwarexyz/monorepo/issues/1432

storage/src/journal/contiguous/fixed.rs
```diff
@@ -726,13 +726,17 @@ impl<E: Context, A: CodecFixedShared> Journal<E, A> {
             meta_recovery_watermark,
         )
         .await?;
+        // Bytes beyond the persisted recovery watermark may be readable after reopen without
```
patrick-ogrady, line 730: prefer a \n above this comment

storage/src/journal/contiguous/fixed.rs
```diff
@@ -726,13 +726,17 @@ impl<E: Context, A: CodecFixedShared> Journal<E, A> {
...
+        // being crash-durable, so the next commit/sync must force a data sync before advancing it.
+        let dirty_from_section =
+            (recovery_watermark < size).then_some(recovery_watermark / items_per_blob);
```
patrick-ogrady: can remove this \n

storage/src/journal/contiguous/fixed.rs
```diff
@@ -726,13 +726,17 @@ impl<E: Context, A: CodecFixedShared> Journal<E, A> {
...
 
         let mut inner = Inner {
             journal,
             size,
             metadata,
             pruning_boundary,
-            dirty_from_section: None,
+            dirty_from_section,
         };
```
patrick-ogrady, line 740: and put a \n after this

patrick-ogrady (approved):

danlaine:
Flagging that this PR's changes prevent the issue described in [this comment](https://github.com/commonwarexyz/monorepo/pull/3950#discussion_r3356606732)

PR #3934, opened 2026-06-02
[CI] Set min stack size to prevent stack overflow in CI tests

danlaine:
And remove now-unneeded `Box::pin` from tests

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,6 +1,5 @@
-# Storage/qmdb tests build very large async state machines (many awaited generic sub-futures,
-# monomorphized across DB variants) and the deterministic runtime polls them inline on libtest's
-# per-test thread. Give test threads ample stack so they don't overflow on platforms with small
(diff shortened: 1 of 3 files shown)


andresilva (commented):
In the tokio runtime we already set the stack size of worker threads (as well as any dedicated threads that are spawned). I wonder if in the dedicated runtime we should spawn a single thread to run the runtime, and make sure that that thread also sets stack size (the same way we do for tokio). This would fix the issue here, and would make deterministic and tokio runtimes closer in behavior.

.cargo/config.toml
```diff
@@ -0,0 +1,5 @@
+# Some qmdb tests build very large async state machines that the deterministic runtime polls inline
+# on libtest's per-test thread. Raise the per-test stack so they don't overflow where the default is
+# small (e.g. Windows).
+[env]
+RUST_MIN_STACK = "16777216"  # 16 MiB
```
patrick-ogrady, line 5: I think this is a weak justification for something that would apply to prod code?

PR #3935, opened 2026-06-02
[consensus/marshal] Validate coded payload config before decode

### LINKED ISSUE #3937: [consensus/marshal] Investigate `CodedBlock` <-> resolver interaction
## Overview

When a `Finalized` block is delivered via marshal's resolver, we currently require that the block is re-encoded upon arrival. Because we have a `Finalization`, however, it should suffice to just send `ApplicationBlock` and check that the finalization's payload commits to the correct block digest. If the certificate is valid, we can construct `CodedBlock::new_trusted` and skip re-encoding.

clabby:
## Overview

Validates that the `CodingConfig` in received `Notarization` certificates is sensible, relative to the declared `Epoch`'s scheme, before decoding.

Also adjusts the `Finalized` delivery path to not require re-encoding, since the `payload` within `Finalization` certificates is trusted.

closes https://github.com/commonwarexyz/monorepo/issues/3937
future work #3938

### DIFF
--- a/consensus/src/marshal/core/actor.rs
+++ b/consensus/src/marshal/core/actor.rs
@@ -1282,6 +1282,14 @@ where
                     return false;
                 };
 
+                // In contrast to the `Block` and `Notarization` deliveries, the finalization delivery
+                // is guaranteed to be certified (assuming the certificate verifies). Because of this,
+                // we can skip broader payload checks and just check that the application block matches
+                // the commitment in the finalization proposal.
+                //
+                // TODO(https://github.com/commonwarexyz/monorepo/issues/3938): Apply this pattern
+                // conditionally to `Request::Block` and `Request::Notarized`, if the requester knows
+                // the requested block is certified.
                 let commitment = finalization.proposal.payload;
                 if block.height() != height
                     || block.digest() != V::commitment_to_inner(commitment)
--- a/consensus/src/marshal/core/actor.rs
+++ b/consensus/src/marshal/core/actor.rs
@@ -1325,6 +1325,15 @@ where
                     return false;
                 };
 
+                // Wrong-round data must be rejected before the scoped-provider lookup below.
+                // That lookup may miss for pruned epochs even when `all()` was sufficient to
+                // decode the certificate, and missing scoped state is treated as stale rather
+                // than peer-faulting.
+                if notarization.round() != round {
+                    response.send_lossy(false);
+                    return false;
+                }
+
                 // `V::check_payload` requires a scoped provider, and `get_scheme_certificate_verifier`
                 // may use an all-epoch provider.
                 let Some(scheme) = self.provider.scoped(notarization.epoch()) else {
@@ -1347,9 +1356,7 @@ where
                     return false;
                 };
 
-                if notarization.round() != round
-                    || V::commitment(&block) != notarization.proposal.payload
-                {
+                if V::commitment(&block) != notarization.proposal.payload {
                     response.send_lossy(false);
                     return false;
                 }
--- a/consensus/src/marshal/core/actor.rs
+++ b/consensus/src/marshal/core/actor.rs
@@ -1276,7 +1276,8 @@ where
                     return false;
                 };
 
-                let Ok(block) = V::ApplicationBlock::read_cfg(&mut value, &self.block_codec_config)
+                let Ok(block) =
+                    V::ApplicationBlock::decode_cfg(&mut value, &self.block_codec_config)
                 else {
                     response.send_lossy(false);
                     return false;
--- a/consensus/src/marshal/core/actor.rs
+++ b/consensus/src/marshal/core/actor.rs
@@ -880,7 +880,7 @@ where
                     debug!(%height, "finalized block missing on request");
                     return;
                 };
-                response.send_lossy((finalization, block).encode());
+                response.send_lossy((finalization, V::into_inner(block)).encode());
             }
             Key::Notarized { round } => {
                 let Some(notarization) = self.cache.get_notarization(round).await else {
@@ -1276,15 +1276,15 @@ where
                     return false;
                 };
 
-                let commitment = finalization.proposal.payload;
-                let block_cfg = V::block_cfg(&self.block_codec_config, commitment);
-                let Ok(block) = V::Block::decode_cfg(value, &block_cfg) else {
+                let Ok(block) = V::ApplicationBlock::read_cfg(&mut value, &self.block_codec_config)
+                else {
                     response.send_lossy(false);
                     return false;
                 };
 
+                let commitment = finalization.proposal.payload;
                 if block.height() != height
-                    || V::commitment(&block) != commitment
+                    || block.digest() != V::commitment_to_inner(commitment)
                     || finalization.epoch() != bounds.epoch()
                 {
                     response.send_lossy(false);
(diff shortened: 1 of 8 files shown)


clabby:
related https://github.com/commonwarexyz/monorepo/issues/3937

consensus/src/marshal/core/actor.rs
```diff
@@ -1276,15 +1276,15 @@ where
...
-                let commitment = finalization.proposal.payload;
-                let block_cfg = V::block_cfg(&self.block_codec_config, commitment);
-                let Ok(block) = V::Block::decode_cfg(value, &block_cfg) else {
+                let Ok(block) = V::ApplicationBlock::read_cfg(&mut value, &self.block_codec_config)
+                else {
                     response.send_lossy(false);
                     return false;
                 };
 
+                let commitment = finalization.proposal.payload;
```
patrick-ogrady, line 1294: Would be great to add an inline comment explaining why this is indeed ok.

consensus/src/marshal/core/actor.rs
```diff
@@ -1276,15 +1276,15 @@ where
...
                 if block.height() != height
-                    || V::commitment(&block) != commitment
+                    || block.digest() != V::commitment_to_inner(commitment)
```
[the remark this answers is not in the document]
clabby: Intentional.

consensus/src/marshal/core/actor.rs
clabby: https://github.com/commonwarexyz/monorepo/pull/3935/commits/a14075e83e72fb3c3d0c6df1f0c9d04fdcee9154

consensus/src/marshal/core/actor.rs
```diff
@@ -1316,7 +1316,22 @@ where
...
 
+                // `V::check_payload` requires a scoped provider, and `get_scheme_certificate_verifier`
+                // may use an all-epoch provider.
+                let Some(scheme) = self.provider.scoped(notarization.epoch()) else {
+                    debug!(
+                        ?round,
+                        floor = %self.floor.processed_height(),
+                        "ignoring stale delivery"
+                    );
+                    response.send_lossy(true);
```
patrick-ogrady, line 1345: Do we block here? I suppose that is correct?

consensus/src/marshal/core/actor.rs
```diff
@@ -1316,7 +1324,22 @@ where
                     return false;
                 };
 
+                // `V::check_payload` requires a scoped provider, and `get_scheme_certificate_verifier`
+                // may use an all-epoch provider.
+                let Some(scheme) = self.provider.scoped(notarization.epoch()) else {
```
patrick-ogrady, line 1339: To avoid this issue, I wonder if we should do some better typing here:

```
  Right now:

  fn all(&self) -> Option<Arc<Self::Scheme>>

  returns the same Scheme type as scoped(...). Since Scheme exposes participants(), the compiler cannot
  distinguish:

  scoped epoch scheme: has participants
  all-epoch certificate verifier: may not have participants

  The type-safe fix is to split the traits/return types:

  trait CertificateVerifier {
      type Certificate;
      type PublicKey;

      fn certificate_codec_config(&self) -> ...;
      fn verify_certificate(...);
      fn verify_certificates(...);
  }

  trait Scheme: CertificateVerifier {
      type Signature;

      fn participants(&self) -> &Set<Self::PublicKey>;
      fn sign(...);
      fn assemble(...);
  }

  Then make the provider look like:

  trait Provider {
      type Scheme: Scheme;
      type All: CertificateVerifier<
          Certificate = <Self::Scheme as CertificateVerifier>::Certificate,
          PublicKey = <Self::Scheme as CertificateVerifier>::PublicKey,
      >;

      fn scoped(&self, epoch: Epoch) -> Option<Arc<Self::Scheme>>;
      fn all(&self) -> Option<Arc<Self::All>>;
  }

  Now provider.all() returns something that can verify certificates but does not have participants(), sign(), or
  assemble(). The exact bug we fixed would become a compile error.
```

consensus/src/marshal/core/actor.rs
clabby: We do not. There's a race where we might not have the scheme registered in the provider yet, but someone delivers us a valid finalization. In other areas of marshal, we also don't block.

consensus/src/marshal/core/actor.rs
clabby: I agree, this is a footgun. Would you prefer to handle that in this PR?

consensus/src/marshal/core/actor.rs
```diff
@@ -1316,7 +1324,22 @@ where
...
+                    debug!(
+                        ?round,
+                        floor = %self.floor.processed_height(),
+                        "ignoring stale delivery"
+                    );
+                    response.send_lossy(true);
```
patrick-ogrady, line 1345:
```

• The notarized delivery path can incorrectly acknowledge an invalid peer response as successful, which can
  stall resolver-driven block fetches. The finalized delivery path also lost the previous exact-consumption
  validation for untrusted payloads.

  Full review comments:

  - [P1] Reject mismatched notarization epochs before acking — /Users/patrickogrady/code/monorepo-pr-3935/
    consensus/src/marshal/core/actor.rs:1329-1336
    When a peer responds to a Key::Notarized fetch with a syntactically valid notarization whose embedded round
    is for an epoch that this provider does not scope, this new lookup returns None before
    notarization.round() != round is checked and sends true. The resolver treats true as an accepted response
    and removes the fetch subscribers, but marshal has not verified or cached the block, so a malicious response
    can leave the local block subscription stuck instead of retrying/blocking the peer.

  - [P2] Reject trailing bytes in finalized deliveries — /Users/patrickogrady/code/monorepo-pr-3935/consensus/
    src/marshal/core/actor.rs:1279-1283
    This switches the finalized block parse from decode_cfg to read_cfg but never checks that value is
    exhausted, so a valid finalization and block with arbitrary trailing bytes is now accepted and acked as
    true. Resolver payloads are untrusted and the previous path rejected ExtraData, so this should preserve the
    exact-consumption check after reading the application block.
```

consensus/src/marshal/core/actor.rs
patrick-ogrady: I think the right approach is to fail here. This fits into the existing invariants of marshal.

consensus/src/marshal/core/actor.rs
patrick-ogrady: I'll do this on top of this.

consensus/src/marshal/core/actor.rs
clabby: https://github.com/commonwarexyz/monorepo/issues/3565

consensus/src/marshal/core/actor.rs
```diff
@@ -1316,7 +1324,22 @@ where
...
+                    return false;
```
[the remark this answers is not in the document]
clabby: Wrong - same behavior as before.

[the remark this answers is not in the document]
patrick-ogrady: This is intended to safely handle pruning.

consensus/src/marshal/core/actor.rs
```diff
@@ -1316,16 +1325,38 @@ where
...
+                // decode the certificate, and missing scoped state is treated as stale rather
+                // than peer-faulting.
+                if notarization.round() != round {
+                    response.send_lossy(false);
+                    return false;
+                }
+
+                // `V::check_payload` requires a scoped provider, and `get_scheme_certificate_verifier`
+                // may use an all-epoch provider.
+                let Some(scheme) = self.provider.scoped(notarization.epoch()) else {
```
patrick-ogrady, line 1339: The LLMs love the idea of making this still work on `standard` (because check_payload doesn't need to do anything) but I don't think that optimization is necessary (already an invariant that we must have a provider that can serve all epochs you want to process).

patrick-ogrady (approved):

PR #3939, opened 2026-06-02
[storage/journal/contiguous] strengthen the crash-recovery fuzz target

danlaine:
## Summary

Rewrites the `journal_crash_recovery` target to drive the fixed and variable contiguous journals through **repeated** unclean shutdowns and verify recovered **content**, not just size bounds. The old target crashed once and only checked that size/oldest fell in a loose `[min, max]` range — so it couldn't catch silent corruption or compounding recovery bugs. Single file changed; same name and CI registration.

## Changes

Each input drives one journal through a sequence of crash cycles. Per cycle: **recover** (`init`, or `init_at_size` for a reset), **verify** the recovered state against a model of what is durably guaranteed, **continue** appending/querying, then **crash** again (run ops under fault injection, drop without a clean sync). The op stream is split into cycles at `Crash`/`Reset` markers; a mutable-method error also ends a cycle (a DB is unusable after one).

## Key changes vs. the existing fuzz test

- **Repeated recovery.** Each cycle re-runs `init()` on a journal produced by a prior crash recovery, so watermark / pruning-metadata / section-layout bugs compound across restarts.
- **Exact-content oracle.** A conservative `Model` asserts, after each recovery: durable-prefix bytes match exactly, pruned positions return `ItemPruned`, the volatile tail is readable-or-pruned, and replay is contiguous.
- **Torn-write faults.** Adds `resize`, `partial_write`, and `partial_resize` (previously only `write`/`sync`).
- **Reset path.** `Reset` reopens via `init_at_size` (discards all data, resets the logical size); a later cycle's `init()` must not resurrect stale pre-reset data.

### DIFF
--- a/storage/fuzz/fuzz_targets/journal_crash_recovery.rs
+++ b/storage/fuzz/fuzz_targets/journal_crash_recovery.rs
@@ -307,6 +307,21 @@ trait FuzzJournal: Sized {
     fn destroy(self) -> impl Future<Output = Result<(), Error>> + Send;
 }
 
+/// Drain a reader's replay stream into a `(position, item)` vector.
+async fn collect_replay<R: Reader<Item = Item>>(
+    reader: R,
+    buffer: NonZeroUsize,
+    start_pos: u64,
+) -> Result<Vec<(u64, Item)>, Error> {
+    let stream = reader.replay(buffer, start_pos).await?;
+    futures::pin_mut!(stream);
+    let mut out = Vec::new();
+    while let Some(result) = stream.next().await {
+        out.push(result?);
+    }
+    Ok(out)
+}
+
 impl FuzzJournal for FixedJournal<deterministic::Context, Item> {
     type Config = FixedConfig;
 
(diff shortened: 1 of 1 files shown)


storage/fuzz/fuzz_targets/journal_crash_recovery.rs
```diff
@@ -70,6 +128,8 @@ enum JournalOperation {
     Read { pos: u64 },
     /// Sync the journal to storage.
     Sync,
+    /// Flush pending data without advancing the recovery watermark.
```
roberto-bayardo: I don't think it's necessary to explain the watermark implementation detail here?

storage/fuzz/fuzz_targets/journal_crash_recovery.rs
```diff
@@ -1,15 +1,58 @@
...
+//! One fuzz input drives a single journal through a series of *cycles*, each one crash-and-recover
+//! round:
+//!   1. `init()` recovers the journal left by the previous cycle's crash.
+//!   2. Check it against the `Expected` carried from that crash.
+//!   3. Append and query under fault injection (the cycle's `ops`).
+//!   4. Drop the journal without a clean shutdown: the crash. Unsynced data is lost.
+//!
+//! The op list is split at `Crash` markers, one `ops` list per cycle. Repeating on the same journal
+//! is the point: watermark, pruning-metadata, and section-layout bugs tend to surface only after
+//! several recover-mutate-crash rounds, not on the first crash.
```
roberto-bayardo: I had to ask Codex to explain this comment which I found confusing. Suggested rewrite:

```suggestion
//! `Crash` markers split the operation list into recovery cycles. Each cycle reopens the
//! same journal state left by the previous unclean shutdown, verifies recovery, applies more
//! operations, and then crashes again. This matters because watermark, pruning-metadata, and
//! section-layout bugs often require a recover-then-mutate-then-recover sequence to appear.
```

storage/fuzz/fuzz_targets/journal_crash_recovery.rs
```diff
@@ -335,289 +423,491 @@ impl FuzzJournal for VariableJournal<deterministic::Context, Item> {
...
+            ops.clone(),
+            partition.clone(),
+            params,
+        );
+    }
 
-    let runner = deterministic::Runner::from(checkpoint);
-    runner.start(|ctx| async move {
+    // Final fault-free phase: verify recovery, then prove a synced round-trip survives an actual
+    // reopen (not just a read from the live handle), and clean up.
```
roberto-bayardo: Wasn't clear to me what you meant by "prove a synced round-trip survives..."    Maybe something more explicit:

```suggestion
// Final fault-free phase: verify the last recovery, then append, sync, drop, and reopen
// a sentinel to prove the synced state survives restart.
```

roberto-bayardo (approved):

PR #3940, opened 2026-06-02
[Storage] Guarantee post-init durability

danlaine:
**A crash-recovery reopen could report success for data it never made durable.** On reopen, `cache`, `archive`, `ordinal`, and `freezer` rebuild their in-memory index by replaying stored bytes, but they never marked those replayed sections as needing a sync — so the first `sync` after a restart returned success without flushing them. This PR makes recovery mark them. (The two contiguous journals were already fixed in #3928/#3929; this finishes the remaining four structures.)

## The bug

A `write` is not durable until a later `sync` returns. But an un-synced write still lands in the OS page cache, which survives a process restart — so recovery reads it back as if it had always been there. Each of these stores keeps a "dirty set" of sections that `sync` flushes and clears. That set was seeded only by writes (`put`), never by recovery, so a reopen left it empty:

1. A process writes to a section, then exits before `sync`. The bytes are in the page cache, not on disk.
2. It restarts. Recovery replays the section and rebuilds the index from those bytes — but does not mark the section dirty.
3. The first `sync` flushes the (empty) dirty set and returns success. The caller now believes the data is durable.
4. Power loss. The page cache is gone, so the un-synced bytes vanish.
5. Next reopen: the data the previous run acknowledged is gone. For `freezer`/`ordinal` it is worse than loss (see below).

## The fix

Recovery now seeds the dirty set with every section it replays, so the existing `sync` flushes them. No new machinery: `put` already feeds the same set, and `sync` already drains and clears it.

- **cache, archive, ordinal** — the recovery replay loop inserts each replayed section into `pending`.
- **freezer** — the checkpoint-less `init` path seeds `modified_sections` with the recovered oversized sections, which `sync` flushes before producing the `Checkpoint`.

Two of the four needed more than the missing seed:

- **freezer** also stops syncing its table during checkpoint-less `init`. The table indexes the `oversized` journal; the recovered oversized sections aren't durable yet, so persisting the table there could leave a durable entry pointing at bytes a crash drops. `recover_table`'s repair is redone on every reopen, so it defers safely to the first `sync`, which flushes oversized first.
- **ordinal** marks a section as soon as its slot bytes are read — *before* the valid/invalid CRC check. A slot whose record fails its CRC is excluded from the index, but that "absent" verdict still rests on bytes recovery read; if they aren't synced, the next reopen can read them differently and rebuild a divergent index. Previously only sections holding a valid record were marked.

## A note on cost

These stores keep no recovery watermark, so they mark *every* replayed section: the first `sync` after a restart fsyncs the whole store (every section, concurrently), not just freshly written ones. Later syncs are normal. It is a one-time startup cost that scales with store size, so a large store sees an I/O burst on its first post-restart commit.

## Testing

Each fix ships a regression test that fails if the fix is reverted. They use the deterministic runtime's fault injection: force every sync to fail (`sync_rate = 1.0`), then assert that `sync` errors after a reopen — proving recovery marked the sections. (The deterministic backend discards un-synced writes on recovery, so the tests assert the sync *happens* rather than reproducing the corruption directly.) Covered: `cache`, `archive`, `ordinal`, `freezer`.

## Scope

No storage or wire format change — this only changes *when* syncs happen. The recovery durability invariant is documented on the existing `Persistable` trait. A separate, pre-existing freezer bug (`max_section` selection on checkpoint-less reopen) is left for a follow-up.

### DIFF
--- a/storage/src/archive/prunable/mod.rs
+++ b/storage/src/archive/prunable/mod.rs
@@ -522,6 +522,59 @@ mod tests {
         });
     }
 
+    // Recovery durability invariant: records replayed on reopen are readable but not yet known
+    // crash-durable, so they stay pending until synced instead of being treated as durable. To
+    // check that, we inject sync faults after a clean write+sync+reopen: `sync()` has to error
+    // because those sections still need syncing. If recovery dropped them, it would no-op and pass.
+    #[test_traced]
+    fn test_archive_reopen_syncs_recovered_records() {
+        let executor = deterministic::Runner::default();
+        executor.start(|context| async move {
+            let cfg = Config {
+                translator: FourCap,
+                key_partition: "test-index".into(),
+                key_page_cache: CacheRef::from_pooler(&context, PAGE_SIZE, PAGE_CACHE_SIZE),
+                value_partition: "test-value".into(),
+                codec_config: (),
+                compression: None,
+                key_write_buffer: NZUsize!(1),
+                value_write_buffer: NZUsize!(1),
+                replay_buffer: NZUsize!(DEFAULT_REPLAY_BUFFER),
+                items_per_section: NZU64!(DEFAULT_ITEMS_PER_SECTION),
+            };
+
+            {
+                let mut archive = Archive::init(context.child("first"), cfg.clone())
+                    .await
+                    .expect("Failed to initialize archive");
+                archive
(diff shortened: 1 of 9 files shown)


danlaine:
Closing in favor of https://github.com/commonwarexyz/monorepo/pull/3950

Issue #3403, opened 2026-03-16
[storage] [medium] init_sync rejects recoverable orphan-leaf journal states before MMR-size repair


dnkolegov-ar:
Found withn review of this [PR](https://github.com/commonwarexyz/monorepo/pull/3399/changes).
`storage/src/merkle/mmr/journaled.rs:410` [checks](https://github.com/commonwarexyz/monorepo/blob/56bb305f8ab35e501e916a024ae2d1713e7a0b17/storage/src/merkle/mmr/journaled.rs#L410) if  `init_sync checks journal.size() > end_pos` and returns `ItemOutOfRange` before doing any MMR-structure recovery.
That is different from normal initialization, where it first rounds down to `PeakIterator::to_nearest_size(...)`, rewinds invalid trailing nodes, and can recover a readable orphan leaf. After a crash, the fixed journal can preserve a fully written extra leaf at `end_pos + 1` without its parents, so this path may reject a recoverable state instead of rewinding to the last valid range.

roberto-bayardo:
@danlaine is this a concern? I'd expect the journal to be discarded on error but not sure what you had in mind for the sync protocol.

roberto-bayardo:
init sync intentionally avoids orphan node recovery since the sync protocol would be able to recover them from a new range request.

danlaine:
This error
```rs
        if size > *end_pos {
            return Err(crate::journal::Error::ItemOutOfRange(size).into());
        }
```
is correct.

If the size of the journal is beyond the sync range (`size > *end_pos`) then the database is already synced. In fact, it's _more_ synced than the sync target -- or more precisely, it reflects a later state then the sync target. Rather than rewinding the data, we return an error.

dnkolegov-ar:
Is this https://github.com/commonwarexyz/monorepo/blob/56bb305f8ab35e501e916a024ae2d1713e7a0b17/storage/src/merkle/mmr/journaled.rs#L85 still correct?

danlaine:
Good catch, that comment is stale. I'll make a PR to fix that.

danlaine:
> Good catch, that comment is stale. I'll make a PR to fix that.

#3406

PR #3949, opened 2026-06-03
[glue] relax current db committed target equivalence

### LINKED ISSUE #3948: [glue] current QMDB rewind/startup reconciliation rejects valid post-prune target (range.start drift)
## Summary

`glue::stateful` reconciliation compares committed sync `Target`s with full-range equality (`Target`'s `PartialEq`, which includes `range.start`). For `current` QMDB this is too strict: `range.start` is the locally chosen safe sync/prune boundary — `Db::sync_boundary()` is documented as the most recent location the database can safely be synced from, and the upper bound on `prune` — not committed-state identity.

`committed_targets()` derives `range.start` from `Db::sync_boundary` (`inactivity_floor_loc / CHUNK_SIZE_BITS`), while a block/application target stamps it from `MerkleizedBatch::sync_boundary` (`bitmap.pruned_chunks()`). An earlier application-stamped start can identify the same target state, even if the local database later reports a higher safe start once the floor advances. Reconciling a database back to a target whose `range.start` was stamped before pruning then panics:

```text
assertion `left == right` failed: rewound database target mismatch after rewind
left:  Target { root: <h>, range: Location(256)..Location(514) }
right: Target { root: <h>, range: Location(0)..Location(514) }
```

`root` (the ops root) and `range.end` match; together they identify the ops prefix used for sync, from which current state is deterministic. Only `range.start` differs. This is a valid post-prune state, not corruption.

## Why `range.start` is sticky for `current` QMDB

- `pruned_chunks` is persisted in `current` QMDB grafted metadata and reloaded on reopen.
- `bitmap::Prunable::truncate` asserts `new_len >= pruned_bits()` — it refuses to shrink into the pruned region.
- `Db::rewind` truncates the ops log / MMR to the rewind size and rebuilds grafting from the *current* `pruned_chunks`; it never un-prunes (pruned chunk-pairs are graft-absorbed and their leaves are gone).

So after rewinding to a `range.end`, `sync_boundary()` stays at the advanced floor. (`sync_boundary::<F, N>(chunk_count, ops_leaves)` may also walk `chunk_count` backward for absorption / delayed-merge constraints; the chunk-count source differs between `Db` and `MerkleizedBatch` as noted above.)

## Where the strict comparison lives

- `glue/src/stateful/db/current.rs` — `rewind_to_target` (all four `current` `ManagedDb` impls), `assert_eq!(rewound_target, target, "rewound database target mismatch after rewind")`.
- `glue/src/stateful/db/mod.rs` — `rewind_to_targets`, and the post-state-sync convergence checks (single-DB `assert!(T::sync_target(&database).await == converged_target, ...)` and the tuple `committed_targets(&synced).await != converged_targets`).
- `glue/src/stateful/actor/syncer/mod.rs` — `init_databases_from_marshal` (`committed_targets().await != processed_targets`, then a post-rewind `assert!`).

This is consistent with how the reference apps already round-trip a retention boundary as `range.start`: `single_db_app`/`multi_db_app` stamp `inactivity_floor` into a block and read it back via `sync_targets`. So stamping a retention boundary into the target is a supported pattern; the issue is reusing strict full-range equality for post-rewind / startup reconciliation.

## Why it wasn't caught

The glue crash/restart tests (`single_db_app`, `multi_db_app`) exercise only `any` QMDB, where the inactivity floor *is* restored by a rewind. No test exercises `current` QMDB + pruning + restart-across-floor, where `sync_boundary` is sticky.

## Reproduction

A focused test using only library APIs reproduces it (no downstream code):

1. Open a `current` QMDB (`ManagedDb`), commit, and prune until `sync_boundary()` advances past 0 (i.e. at least one chunk pruned).
2. Capture the live committed target `{ root, [advanced_start, end) }`.
3. Build an "earlier application-stamped start" target `{ root, [0, end) }` (the start a block stamps from `MerkleizedBatch::sync_boundary` before pruning).
4. `rewind_to_target(earlier_start_target)`.

On current `main` this panics with `rewound database target mismatch after rewind` (transcript above, `left` = live pruned floor `256`, `right` = earlier start `0`, same `root` + `range.end`). The same shape reproduces across all four `current` variants (unordered/ordered × fixed/variable).

## Context / prior art

- Follow-up edge case for the rewind / crash-recovery work in #3444 (which modeled `[inactivity_floor, commit_floor)` ranges generally and assumed the lower bound is restorable by retaining the prior range's operations; for `current` QMDB the lower bound is not restorable).
- Same class as #3403 — reconciliation rejecting a recoverable/valid state with a strict check.

## Expected behavior

For `current` QMDB rewind / startup / state-sync reconciliation, accept a rewound target when `root` and `range.end` match, even if `range.start` differs due to the local pruning floor advancing. `matches_sync_target` (propose/verify) should remain strict — freshly merkleized batches and block commitments must still agree exactly.

## Proposed fix direction

Introduce a DB-overridable equivalence used only for reconciliation (default = exact target equality; `current` QMDB overrides to compare `root` + `range.end`), and use it at the reconciliation sites above — the post-rewind, startup, and post-state-sync convergence checks only, not sync-target validation, fetch planning, or `matches_sync_target`, which stay strict. A PR with this fix plus the per-variant regression tests will follow; happy to adopt a different shape (e.g. normalizing the application-derived target) if preferred.

## Real-world impact

Observed on a validator network running `current` QMDB with pruning enabled: after normal floor/pruning advancement, validators entered a restart crash-loop with the panic above once they crossed the pruning boundary. (The panic text is verbatim library output — a root hash and two op-count integers.)

marthendalnunes:
Fixes #3948.

## Problem

`glue::stateful` reconciliation compares committed sync `Target`s with full-range equality (`Target`'s `PartialEq`, which includes `range.start`). For `current` QMDB, `range.start` is the locally chosen safe sync/prune boundary (`Db::sync_boundary()` — the most recent location the database can safely be synced from, and the upper bound on `prune`), not committed-state identity. It advances as the floor advances and is not restorable by a rewind (pruned chunk-pairs are graft-absorbed; `pruned_chunks` is persisted in `current` QMDB grafted metadata and `Prunable::truncate` refuses to shrink the pruned region).

So reconciling a database back to a target whose `range.start` was stamped (from `MerkleizedBatch::sync_boundary`) before pruning panics, even though `root` and `range.end` match:

```text
assertion `left == right` failed: rewound database target mismatch after rewind
left:  Target { root: <h>, range: Location(256)..Location(514) }
right: Target { root: <h>, range: Location(0)..Location(514) }
```

Full mechanism and context in #3948.

## Change

- Add `ManagedDb::committed_target_equivalent` (default: exact equality) and `DatabaseSet::committed_targets_equivalent`.
- `current` QMDB overrides the predicate to compare `root` + `range.end` only.
- Use it for all reconciliation-time comparisons: `rewind_to_target` (all four `current` impls), `rewind_to_targets`, startup `init_databases_from_marshal`, and the single- and tuple-DB post-state-sync convergence checks.
- Sync-target validation, fetch planning, and `matches_sync_target` (propose/verify) stay strict.

## Tests

Per-variant regression tests (unordered/ordered × fixed/variable) that prune to advance the sync boundary, then reconcile to an earlier-start target. They **fail on the prior behavior** with the exact `rewound database target mismatch after rewind` panic (same `root` + `range.end`, `range.start` 256 vs 0) and **pass with this change**. Plus a unit test for the tuple-level `committed_targets_equivalent` predicate.

The existing crash/restart tests (`single_db_app`, `multi_db_app`) exercise only `any` QMDB, where the inactivity floor is restored by a rewind — which is why this was not previously caught.

## Verification

- New tests fail on `main` (the panic above) and pass with this change, across all four `current` variants.
- `commonware-glue` lib suite: 158 passing.
- `cargo clippy -p commonware-glue --all-targets -- -D warnings`: clean.
- `rustfmt` (project config): clean.

## Notes for reviewers

This is a follow-up edge case for the rewind/crash-recovery work in #3444 (which modeled `[inactivity_floor, commit_floor)` ranges generally and assumed the lower bound is restorable); same class as #3403 (recoverable state rejected by a strict check).

I extended the relaxation to the post-state-sync convergence checks for consistency (same predicate); production did not exhibit post-sync drift. Happy to revert those two sites to strict, or to adopt a different shape (e.g. normalizing the application-derived target's `range.start`) if you'd prefer.

### DIFF
--- a/glue/src/stateful/actor/syncer/mod.rs
+++ b/glue/src/stateful/actor/syncer/mod.rs
@@ -416,11 +416,12 @@ where
     // we attempt to repair by rewinding the databases back to the marshal floor. If
     // the rewind fails to produce a consistent state, we must crash. This can occur
     // if the databases were corrupted or pruned to aggressively.
-    if databases.committed_targets().await != processed_targets {
+    let committed_targets = databases.committed_targets().await;
+    if !A::Databases::committed_targets_equivalent(&committed_targets, &processed_targets) {
         databases.rewind_to_targets(processed_targets.clone()).await;
         let rewound_targets = databases.committed_targets().await;
         assert!(
-            rewound_targets == processed_targets,
+            A::Databases::committed_targets_equivalent(&rewound_targets, &processed_targets),
             "databases must be consistent with marshal floor after rewind"
         );
     }
--- a/glue/src/stateful/db/current.rs
+++ b/glue/src/stateful/db/current.rs
@@ -11,7 +11,7 @@ use crate::stateful::db::{
     Unmerkleized as UnmerkleizedTrait,
 };
 use commonware_codec::{Codec, Read as CodecRead};
-use commonware_cryptography::Hasher;
+use commonware_cryptography::{Digest, Hasher};
 use commonware_parallel::Strategy;
 use commonware_runtime::{Clock, Metrics, Storage};
 use commonware_storage::{
@@ -46,6 +46,17 @@ use std::{ops::Deref, sync::Arc};
 
 type CurrentDbHandle<F, E, C, I, H, U, const N: usize, S> =
     Arc<AsyncRwLock<Db<F, E, C, I, H, U, N, S>>>;
+
+fn current_committed_target_equivalent<F, D>(
+    actual: &CurrentSyncTarget<F, D>,
+    expected: &CurrentSyncTarget<F, D>,
+) -> bool
+where
+    F: Graftable,
+    D: Digest,
+{
+    actual.root == expected.root && actual.range.end() == expected.range.end()
+}
 
 /// Wraps a QMDB [`UnmerkleizedBatch`] with a reference to the parent
 /// database, implementing the [`Unmerkleized`](super::Unmerkleized) trait.
@@ -381,6 +392,10 @@ where
             && *target.range.end() == Location::<F>::new(batch.bounds().total_size)
     }
 
+    fn committed_target_equivalent(actual: &Self::SyncTarget, expected: &Self::SyncTarget) -> bool {
+        current_committed_target_equivalent(actual, expected)
+    }
+
     async fn finalize(&mut self, batch: Self::Merkleized) -> Result<(), Error<F>> {
         self.apply_batch(batch.inner).await?;
         self.sync().await?;
@@ -400,8 +415,8 @@ where
         self.sync().await?;
 
         let rewound_target = self.sync_target().await;
-        assert_eq!(
-            rewound_target, target,
+        assert!(
+            Self::committed_target_equivalent(&rewound_target, &target),
             "rewound database target mismatch after rewind",
         );
         Ok(())
@@ -472,6 +487,10 @@ where
             && *target.range.end() == Location::<F>::new(batch.bounds().total_size)
     }
 
+    fn committed_target_equivalent(actual: &Self::SyncTarget, expected: &Self::SyncTarget) -> bool {
+        current_committed_target_equivalent(actual, expected)
+    }
+
     async fn finalize(&mut self, batch: Self::Merkleized) -> Result<(), Error<F>> {
         self.apply_batch(batch.inner).await?;
         self.sync().await?;
@@ -491,8 +510,8 @@ where
         self.sync().await?;
 
         let rewound_target = self.sync_target().await;
-        assert_eq!(
-            rewound_target, target,
+        assert!(
+            Self::committed_target_equivalent(&rewound_target, &target),
             "rewound database target mismatch after rewind",
         );
         Ok(())
@@ -640,6 +659,10 @@ where
             && *target.range.end() == Location::<F>::new(batch.bounds().total_size)
     }
 
+    fn committed_target_equivalent(actual: &Self::SyncTarget, expected: &Self::SyncTarget) -> bool {
+        current_committed_target_equivalent(actual, expected)
+    }
+
     async fn finalize(&mut self, batch: Self::Merkleized) -> Result<(), Error<F>> {
         self.apply_batch(batch.inner).await?;
         self.sync().await?;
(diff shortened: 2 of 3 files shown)


clabby:
I don't think this is the right fix.

The strict equality check is intentional here. A `Target` is not just "the committed state identity"; it is also the sync/rewind availability contract. `[0, 514)` and `[256, 514)` may have the same `root` and `range.end`, but they are not equivalent targets - the latter can no longer serve or reconstruct the operations required by the former.

I think the issue is that `current::Db::sync_boundary()` is being treated as the globally safe pruning point in your application. It is only the storage-layer safe upper bound for Current QMDB's own grafting/log invariants. When using `glue::stateful`, the application also has to retain enough history for block sync targets that may still be needed during startup reconciliation or peer state sync. If a finalized block advertises `[inactivity_floor, db_size)`, pruning above that `inactivity_floor` invalidates that block's target, even if the current-state root is unchanged.

Relaxing reconciliation to compare only `root + range.end` hides that violation and introduces a different bug: glue can now believe a database is aligned with a marshal-floor target that it cannot actually serve as a sync target. The panic is useful signal that the DB was pruned past a still-required target.

I think the fix is to prune less aggressively within the application. Keep at least the lowest `target.range.start` that the node may still need for startup reconciliation or peer state sync. Since we hand applications raw DB handles and cannot enforce that directly in glue, we should document this more clearly imo. I’ll add docs around `current::Db::sync_boundary` / `prune` and the glue stateful DB integration to make clear that `sync_boundary()` is a local storage bound, not a complete pruning policy when glue may need older block targets. Also will consider exposing a more limited DB handle for a safe API surrounding this, though a bit torn on if that would limit power-users' capabilities.

patrick-ogrady:
I think restart will also fail if you prune beyond the inactivity floor (but there may be some interplay between the grafting overlay that still allows you to prune beyond it but still recover from the operations log ... in any case, you should never be pruning beyond the inactivity floor)?

clabby:
Landed on a solution I'm happier with @ https://github.com/commonwarexyz/monorepo/pull/3965. This pulls (safe) pruning of the QMDB instances and marshal into `Stateful` while still allowing a raw handle to the DBs to be exposed.

PR #3950, opened 2026-06-03
[runtime] [storage] sync file system on runtime start to ensure blob durability

danlaine:
## Summary

On reopen, storage structures (`cache`, `archive`, `ordinal`, `freezer`, and the contiguous journals) rebuild their in-memory state by reading bytes off disk during `init`. After a **process restart** those bytes can be page-cache-resident but **not yet crash-durable** — a `write` only becomes durable once a later `sync` returns, and the OS page cache survives a process restart. A structure that reads such bytes, treats them as durable, and then persists a derived artifact (a checkpoint, a recovery watermark, an offsets entry) can be left with a dangling reference after a later crash, causing data loss / corruption.

This PR closes that bug class at the **runtime layer**. The production runtime flushes the whole storage filesystem **once at startup**, before any structure opens a blob, so everything read during `init` is durable. On Linux this is a single `syncfs(2)`.

## Mechanism

`Runner::start` calls `storage::sync_fs(&cfg.storage_directory)` after the storage backend is constructed and before the user future runs (so it precedes every blob open). The helper is platform-gated:

- **Linux**: `syncfs(2)` makes all data on the storage filesystem crash-durable. A `syncfs` error is fatal — if durability can't be established, startup panics rather than continue.
- **macOS/BSD**: best-effort `sync(2)` only. macOS has no whole-filesystem durable flush, and `sync` does not flush the drive cache (and may return before buffers are written), so this is **not** a crash-durability guarantee.
- **Windows / wasm**: no whole-filesystem flush is performed; the guarantee is not provided.

**Why this suffices on Linux:** a process restart preserves the page cache, so the startup `syncfs` makes the prior process's un-synced bytes durable before anything reads them.

**Assumptions:** storage lives on a single filesystem (`syncfs` covers one filesystem); reliable `syncfs` error reporting requires Linux >= 5.8.

## Scope

The guarantee is **Linux-only**. macOS is best-effort and Windows is unaddressed, because neither offers a non-privileged, whole-filesystem durable flush callable at startup.


runtime/src/tokio/runtime.rs
```diff
@@ -498,6 +507,57 @@ impl crate::Runner for Runner {
...
+                Ok(file) => file,
+                Err(e) if e.kind() == std::io::ErrorKind::NotFound => return Ok(()),
+                Err(e) => return Err(e),
+            };
+            // SAFETY: `file` owns a valid fd that lives across the call; `syncfs` takes only that
+            // fd, performs no memory access, and returns -1 on error.
+            if unsafe { libc::syncfs(file.as_raw_fd()) } == -1 {
+                return Err(std::io::Error::last_os_error());
+            }
+            tracing::info!(
```
patrick-ogrady: I think this can be debug

runtime/src/tokio/runtime.rs
```diff
@@ -498,6 +507,57 @@ impl crate::Runner for Runner {
...
+///   data a structure reads during `init` is durable.
+/// - **macOS/BSD**: best-effort `sync(2)` only. macOS has no whole-filesystem durable flush, and
+///   `sync` does not flush the drive cache (and may return before buffers are written), so it is
+///   **not** crash-durable.
+/// - **Windows/wasm**: no whole-filesystem flush is performed; the guarantee is not provided.
+///
+/// Assumes the storage lives on a single filesystem (`syncfs` covers one filesystem), and on Linux
+/// reliable error detection requires kernel >= 5.8. A missing `dir` (a fresh start with no data
+/// yet) is treated as success.
+fn sync_storage_filesystem(dir: &std::path::Path) -> std::io::Result<()> {
```
patrick-ogrady: should we define this in storage? and just call it `sync`?

danlaine: Since this is a new guarantee provided by the runtime it seems reasonable to keep it defined here? I just think it would feel a bit odd to declare this in the storage crate and make it `pub` and be like "the runtime has to call this" instead of the runtime just handling it. Your call on naming.

runtime/src/tokio/runtime.rs
danlaine: done

runtime/src/tokio/runtime.rs
patrick-ogrady: I think its still a capability that is "storage in nature" (so the impl can be defined there)?

patrick-ogrady: Was there a previous PR you merged that added this sync behavior to journal? Should we revert that now?

roberto-bayardo:
Should we do something for windows too?

```
Closest equivalent: FlushFileBuffers on a volume handle.

On Windows:

FlushFileBuffers(file_handle) is like fsync(file_fd) for one file.
FlushFileBuffers(volume_handle) is roughly like syncfs(fd): flushes all open files on that volume.
Microsoft documents this explicitly: to flush all open files on a volume, call FlushFileBuffers with a handle to the volume. 
```

danlaine:
> Should we do something for windows too?
> 
> ```
> Closest equivalent: FlushFileBuffers on a volume handle.
> 
> On Windows:
> 
> FlushFileBuffers(file_handle) is like fsync(file_fd) for one file.
> FlushFileBuffers(volume_handle) is roughly like syncfs(fd): flushes all open files on that volume.
> Microsoft documents this explicitly: to flush all open files on a volume, call FlushFileBuffers with a handle to the volume. 
> ```

This requires "admin privileges" on Windows. Idk much about what that is but I'm reading that it's non-standard to run things in that mode. I'd be fine leaving off Windows support but also OK adding a best-effort attempt to call `FlushFileBuffers`.

runtime/src/tokio/runtime.rs
danlaine: Moved to storage crate and renamed `sync_fs`

danlaine: Reverted #3928 changes

roberto-bayardo:
> This requires "admin privileges" on Windows. Idk much about what that is but I'm reading that it's non-standard to run things in that mode. I'd be fine leaving off Windows support but also OK adding a best-effort attempt to call `FlushFileBuffers`.

I think most users sign into windows in admin mode (I know I always did back when I sed it), but not sure if this is how server-side deployments work.

runtime/src/tokio/runtime.rs
```diff
@@ -423,6 +423,15 @@ impl crate::Runner for Runner {
             }
         }
 
+        // Make any storage a prior process left in the page cache crash-durable before user code
+        // reads.
+        if let Err(e) = crate::storage::sync_fs(&self.cfg.storage_directory) {
```
patrick-ogrady: How would you feel about calling this before initializing storage? I think that ensures we don't have multiple references to the FS?

danlaine: done

runtime/src/tokio/runtime.rs
```diff
@@ -393,6 +393,15 @@ impl crate::Runner for Runner {
             &mut runtime_registry.sub_registry("storage_buffer_pool"),
         );
 
+        // Make any storage a prior process left in the page cache crash-durable before we open it,
+        // so the data read during init is durable.
+        if let Err(e) = crate::storage::sync_fs(&self.cfg.storage_directory) {
```
patrick-ogrady: I think there should be an option to do an equivalent in deterministic (IIRC items stay in the staged memory until sync)? This may be around the checkpoint boundary).

danlaine: Is there a need for an equivalent in the deterministic runtime? It already upholds the invariant that any data read from its blobs are durably persisted, no?

runtime/src/storage/mod.rs
```diff
@@ -2,6 +2,99 @@
...
+            } else if #[cfg(unix)] {
+                // SAFETY: `sync` takes no arguments and cannot fail.
+                unsafe { libc::sync() };
+                tracing::debug!(
+                    storage_directory = %dir.display(),
+                    "best-effort storage flush at startup (sync(); not a crash-durability guarantee)"
+                );
+                Ok(())
+            } else if #[cfg(windows)] {
+                // Resolve the volume containing `dir` (e.g. `C:\...` -> `\\.\C:`). `sync_all` on a
```
patrick-ogrady, line 46: Can we add a windows specific restart test to ensure this works?

danlaine: done...let's see if it passes ci 👀

runtime/src/tokio/runtime.rs
patrick-ogrady: I don't think so. I think I made it more flexible (only "locking in" on sync)....will poke around and see.

storage/src/journal/contiguous/fixed.rs
```diff
@@ -727,16 +727,12 @@ impl<E: Context, A: CodecFixedShared> Journal<E, A> {
...
-        // being crash-durable, so the next commit/sync must force a data sync before advancing it.
-        let dirty_from_section =
-            (recovery_watermark < size).then_some(recovery_watermark / items_per_blob);
         let mut inner = Inner {
             journal,
             size,
             metadata,
             pruning_boundary,
-            dirty_from_section,
+            dirty_from_section: None,
```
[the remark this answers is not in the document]
danlaine: Hm I think this is a real, if pathological, case:

```
Process starts. Runner::start calls sync_fs(storage_directory). At this moment, existing dirty page-cache data from a prior process is made durable.

App opens J1. The journal recovers existing state and sets dirty_from_section = None.

App appends enough data to J1 that some full pages are written to blob files via plain write_at. These bytes are now readable from the OS/page cache and may have valid journal CRCs, but they have not been fsynced.

App drops J1 without commit() or sync(). Any purely in-memory append buffer is lost, but already-issued plain writes remain readable in the kernel page cache. No live journal object remembers these sections are dirty.

In the same process, app opens J2 on the same partition. No startup sync_fs runs, because the process did not restart. J2 walks blob lengths / valid pages and adopts the readable bytes as recovered journal contents.

J2 initializes with dirty_from_section = None. This is the bug: recovered bytes that were written by a previous journal instance in the same process may be non-durable, but the new instance does not mark their sections as needing sync.

App calls J2.sync(). Since dirty_from_section is None, flush_dirty_sections() does not fsync the adopted data sections. Then sync() advances and persists the recovery metadata/watermark.

A power loss happens before the adopted data pages reach durable storage. The durable metadata can now point past data that was only ever in volatile cache.

On reboot, startup sync_fs cannot help: the lost page-cache data no longer exists. Reopening the journal may find a recovery watermark or offsets state beyond the recoverable data and report corruption, or worse for related structures, preserve derived state from vanished data.
```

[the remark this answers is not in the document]
danlaine: Fwiw #3947 fixes this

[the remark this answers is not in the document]
roberto-bayardo: can we just add a sync call to blob.close() (which I presume is called on drop)?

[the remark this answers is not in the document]
patrick-ogrady: Or when you open a blob there are no dirty pages in the page cache?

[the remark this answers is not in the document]
danlaine: > can we just add a sync call to blob.close() (which I presume is called on drop)?

There isn't a `blob.close()` function. And even if `drop` called `sync`, the latter might not succeed and there's no safe way to re-try in `drop`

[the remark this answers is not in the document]
danlaine: I undid the revert of #3928 in these changes so the bad path cursor mentioned is now guarded against

PR #3956, opened 2026-06-04
[ci] Use gungraun for benchmark tracking

clabby:
## Overview

Updates the benchmark tracking workflow to use [`gungraun`](https://gungraun.github.io/gungraun/latest/html/intro.html) one-shot benchmarks. The metrics emitted (`Ir, L1hits, LLhits, RamHits, TotalRW, EstimatedCycles`) are much more stable across runs, avoiding the noise problem with using criterion benchmarks for this task.

In addition, benchmarks can provide custom metrics that gates can use, e.g. disk reads from `commonware-runtime`, in case there is a need to constrain specific properties over time.

A `commonware-bench` crate was also added to provide a standard interface for benchmark workloads. This makes the path towards measuring a benchmark in both `gungraun` and `criterion` much more simple. To measure an existing `criterion` benchmark with `gungraun` and begin tracking it, all that needs to be done is to redefine the workload in the fork of the `Workload` trait.


storage/Cargo.toml
```diff
@@ -34,6 +34,7 @@ tracing = { workspace = true, optional = true }
 zstd = { workspace = true, optional = true }
 
 [dev-dependencies]
+commonware-bench = { path = "../bench" }
```
patrick-ogrady: we should do a proper crate fetch?

clabby: Ah yeah; This was left over from an experiment where we didn't publish the crate. To make things a bit easier (& in case any of our users wanted to use it,) changed over to publishing it.

.github/benchmark-tracking.toml
```diff
@@ -1,56 +1,67 @@
...
 cargo_flags = ["--features", "test-traits"]
 
 [[packages.benchmarks]]
-name = "qmdb"
-criterion_args = ["--sample-size", "500", "--significance-level", "0.01"]
-variants = [
-  "qmdb::merkleize/variant=any::unordered::fixed::mmr keys=10000 ch=false sync=false",
-  "qmdb::merkleize/variant=current::ordered::fixed::mmb chunk=256 keys=10000 ch=true sync=false",
-]
+name = "qmdb_gungraun"
```
patrick-ogrady, line 49: I think we could just call this `qmdb`?

bench/src/lib.rs
```diff
@@ -0,0 +1,205 @@
+//! Shared benchmark bridge utilities.
```
patrick-ogrady: Align with crate description.

runtime/src/utils/buffer/paged/cache.rs
```diff
@@ -437,6 +437,13 @@ impl CacheRef {
         buf.len()
     }
 
+    /// Drop all cached pages while retaining the backing page buffers for reuse.
+    ///
+    /// Call only when no reads are in flight for this cache.
+    pub fn clear(&self) {
```
patrick-ogrady, line 444: These seem like dangerous functions to leave around just for this benchmark PR?

storage/src/qmdb/benches/merkleize.rs
```diff
@@ -318,11 +320,10 @@ fn var_log_cfg(pc: CacheRef) -> VConfig<((), ())> {
 
 // -- DB constructors (eliminates repeated config boilerplate in match arms) --
 
-fn any_fix_cfg(
+pub(crate) fn any_fix_cfg_with_cache(
```
patrick-ogrady, line 323: I think all these shared items should live somewhere in a shared util file like we do in other crates?

storage/Cargo.toml
```diff
@@ -97,6 +107,12 @@ harness = false
 path = "src/qmdb/benches/init_scale.rs"
 required-features = ["test-traits"]
 
+[[bench]]
+name = "qmdb_gungraun"
```
patrick-ogrady: I think including gungraun in the names of these benchmarks/filenames is an anti-pattern? just tell me what it is?

patrick-ogrady (commented):
nits

storage/Cargo.toml
clabby: If we called it `qmdb`, it'd conflict with the existing `qmdb` benchmark target. Because gungraun uses a different benchmark harness, it needs a different entrypoint.

.github/benchmark-tracking.toml
clabby: If we called it `qmdb`, it'd conflict with the existing `qmdb` benchmark target. Because gungraun uses a different benchmark harness, it needs a different entrypoint.

runtime/src/utils/buffer/paged/cache.rs
clabby: I'll gate behind `test-utils`.

storage/src/qmdb/benches/merkleize.rs
clabby: I think it's a bit easier to define a criterion module and then export what the gungraun module might need, though moved the shared items over to `merkleize_common.rs` in https://github.com/commonwarexyz/monorepo/pull/3956/changes/6a8a90a3ce003ef135df5faa2000a3a35ca61b1d. LMK which you'd prefer.

storage/src/qmdb/benches/bench.rs
```diff
@@ -8,6 +8,7 @@ mod common;
 mod generate;
 mod init;
 mod merkleize;
+mod merkleize_common;
```
patrick-ogrady: I guess what I had in mind was collapsing this into `common` rather than `merkleize_common`.

A clarifying question: the benchmark for gungraun are currently gungraun.rs rather than something like merkleize_gungraun? is that the plan going forward? Or would it make sense to put the merkleize gungraun impl in merkleize?

clabby: spoke offline, reorganized in https://github.com/commonwarexyz/monorepo/pull/3956/commits/403b9a91f67f846ec4ca59566e4356f224b05d55

patrick-ogrady (approved):

PR #3960, opened 2026-06-05
[runtime/iobuf/benches] add iobuf decode benchmark

andresilva:


runtime/src/iobuf/benches/pool.rs
```diff
@@ -130,42 +101,47 @@ pub fn bench(c: &mut Criterion) {
...
+                        },
+                    )
+                });
             });
-
-            full.saturating_sub(baseline)
-        });
-    });
+        }
+        true => {
```
andresilva, line 128: This microbenchmark is very sensitive to code shape, putting an `if touch` inside the benchmark loop affects the results.

patrick-ogrady (approved):
nice

PR #3964, opened 2026-06-05
[runtime] [storage] Split `Append` into read/write handles

danlaine:
## Motivation

The paged-buffer layer exposed a single `Append` handle for both reading and writing a blob, and it was `#[derive(Clone)]`. Because any clone could write, "only one writer" could not be expressed in the type system — it was enforced at run time: `seal()` checked for outstanding clones and returned `Error::NotUnique` if any existed. That made the unique-writer rule a footgun (a stray clone turns a `seal` into a runtime error), and it left the shape the journal layer above actually wants — one writer, many concurrent readers — inexpressible.

This PR splits `Append` into a `Writer` that is the sole mutator and a cheap, cloneable `Reader`. Unique-writer becomes a property of the type rather than a runtime check, so `Error::NotUnique` goes away entirely.

## The split

- **`Writer<B>`** is the unique owner of a blob's append path — `append`, `sync`, `resize`, `replay`, `seal` — and can also read directly. It is not `Clone`, so at most one exists per wrapped blob.
- **`Reader<B>`** is a read-only view produced by `Writer::reader()`. It *is* `Clone`, so a caller can hand out as many as it likes. Writer and readers share one `Arc<Shared>` over the blob's buffered state, which has two consequences worth calling out: a reader observes bytes that are buffered but not yet flushed, and a reader stays valid after `Writer::seal` consumes the writer (or the writer is dropped).
- **`seal(self) -> Sealed`** consumes the `Writer`. Since the writer is unique by construction, sealing can no longer fail for ownership reasons — the old `Error::NotUnique` path is removed.

This is the runtime primitive the journal's snapshot readers are built on: one writer mutating a blob, alongside any number of independent read handles that see in-flight data and outlive the writer.

## Behavior changes (relative to base)

- **`paged::Append` is replaced by `paged::Writer` + `paged::Reader`.** Construction keeps the same arguments (`Append::new` → `Writer::new`); the consumers — the storage journals and the segmented-buffer factory — are mechanical renames.
- **`Writer` is not `Clone`; `Reader` is.** Shared read access now goes through `Writer::reader()` rather than cloning the write handle.
- **`seal` no longer returns `Error::NotUnique`, and that variant is removed**, because uniqueness is now guaranteed at compile time instead of checked at run time.

The on-disk format is unchanged: this is a handle-API refactor over the same paged-CRC layout, with no change to read, write, or recovery semantics.


roberto-bayardo (approved):

patrick-ogrady:

Issue #3963, opened 2026-06-05
[glue/stateful] Investigate `sync` over `commit`


clabby:
## Overview

In each DB's `ManagedDb::finalize` implementation, we currently call `sync`:

https://github.com/commonwarexyz/monorepo/blob/6dfe1560f93ee153a4e45a1cd661ddfb7407bc33/glue/src/stateful/db/current.rs#L384-L388

This is a much heavier operation than `commit`. We may be able to change this to `commit`, and expose an operation to `sync` periodically. The only downside would be larger recovery times with crashes, but we'd lose no durability guarantees.

patrick-ogrady:
@roberto-bayardo mentioned that prune may already sync, in which case this is a 1-liner.

clabby:
Re-assigning to @patrick-ogrady - this fits in line with the deferred `fsync` work. related: #3977

PR #3965, opened 2026-06-05
[glue/stateful] Automate pruning

clabby:
## Overview

> [!NOTE]
> PR Stack (merge in reverse order):
> - #3984 
> - #3981 
> - #3965 👈
> - `main`

Adds an option to automatically prune both the QMDB instances and marshal. Pruning is performed according to the contract of `glue::stateful`: marshal and the QMDB instances must retain at least enough data to recover up to `max_pending_acks + 1` blocks worth of state during startup reconciliation after a crash.

As a part of this change, `stateful::Application::finalize` was also changed to ensure that it follows the pattern of marshal's stream with an at-least-once delivery. After state sync concludes, every single block streamed in by marshal will be delivered regardless of whether or not its batches were applied. This is useful for simplifying indexers, since this hook runs _after_ the database batches for the reported block have been persisted.

related #3963 
related #3949

### DIFF
--- a/glue/src/stateful/actor/core/mod.rs
+++ b/glue/src/stateful/actor/core/mod.rs
@@ -47,8 +47,9 @@ pub struct MaintenanceConfig {
 
     /// Prune databases and marshal instead of only persisting committed state.
     ///
-    /// When `true`, glue retains enough finalized history to safely rewind
-    /// `max_pending_acks` blocks before pruning.
+    /// When `true`, glue retains a `max_pending_acks + 1` finalized-target
+    /// window before pruning, so the oldest retained target stays
+    /// `max_pending_acks` blocks behind the finalized tip.
     pub prune: bool,
 }
 
--- a/glue/src/stateful/actor/core/mod.rs
+++ b/glue/src/stateful/actor/core/mod.rs
@@ -42,32 +42,31 @@ type BlockDigest<A, E> = <<A as Application<E>>::Block as Digestible>::Digest;
 /// Periodic pruning configuration.
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
 pub struct PruneConfig {
-    /// Prune databases and marshal every `maintenance_interval` finalized blocks.
-    pub maintenance_interval: NonZeroUsize,
-
-    /// Extra finalized blocks to retain in marshal beyond `max_pending_acks + 1`.
+    /// Finalized blocks to retain in marshal beyond `max_pending_acks + 1`.
     ///
     /// This should generally be set to a large enough number of blocks to facilitate downtime
     /// on a validator that has completed state sync. If the prune cadence on marshal is too
     /// aggressive, a rebooted node may fail to recover due to peers being unable to serve the
     /// blocks it needs to catch up.
-    pub additional_marshal_blocks: usize,
+    ///
+    /// This value is also used as the pruning cadence.
+    pub retained_marshal_blocks: NonZeroUsize,
 
-    /// Extra finalized blocks' worth of operations to retain in QMDB beyond `max_pending_acks + 1`.
+    /// Finalized blocks' worth of operations to retain in QMDB beyond `max_pending_acks + 1`.
     ///
     /// This value is generally safe to set to 0, as QMDB operations below the active range are only
     /// needed to serve state sync requests for lagging peers. Some network topologies may benefit from
     /// a non-zero value here to provide a larger buffer for serving state sync requests during periods
     /// of instability.
-    pub additional_qmdb_blocks: usize,
+    pub retained_qmdb_blocks: usize,
 }
 
 impl PruneConfig {
     /// Ensure marshal is never pruned more aggressively than QMDB.
     pub const fn assert_valid(self) {
         assert!(
-            self.additional_marshal_blocks >= self.additional_qmdb_blocks,
-            "marshal must retain at least as many additional blocks as QMDB",
+            self.retained_marshal_blocks.get() >= self.retained_qmdb_blocks,
+            "marshal must retain at least as many blocks as QMDB",
         );
     }
 }
@@ -114,7 +113,7 @@ where
     /// Periodic database and marshal pruning configuration.
     ///
     /// When enabled, glue retains `max_pending_acks + 1` finalized blocks plus
-    /// the configured extra block windows before pruning. Marshal must retain
+    /// the configured retained block windows before pruning. Marshal must retain
     /// at least as many blocks as QMDB.
     pub prune_config: Option<PruneConfig>,
 }
--- a/glue/src/stateful/actor/core/mod.rs
+++ b/glue/src/stateful/actor/core/mod.rs
@@ -29,7 +29,10 @@ use commonware_runtime::{spawn_cell, Clock, ContextCell, Handle, Metrics, Spawne
 use commonware_utils::{channel::oneshot, sync::AsyncMutex};
 use futures::join;
 use rand::Rng;
-use std::{num::NonZeroUsize, sync::Arc};
+use std::{
+    num::{NonZeroU64, NonZeroUsize},
+    sync::Arc,
+};
 
 mod mailbox;
 pub use mailbox::Mailbox;
@@ -77,6 +80,12 @@ where
 
     /// Sync engine tuning knobs.
     pub sync_config: SyncEngineConfig,
+
+    /// Period for explicit database syncs during finalization.
+    ///
+    /// Set this to [`None`] if you manually prune the databases on a known
+    /// cadence instead. Pruning already syncs the underlying QMDB state.
+    pub finalize_sync_interval: Option<NonZeroU64>,
 }
 
 /// Stateful application that manages the pending-tip DAG of merkleized
@@ -115,6 +124,9 @@ where
 
     /// Sync engine tuning knobs.
     sync_config: SyncEngineConfig,
+
+    /// Period for explicit database syncs during finalization.
+    finalize_sync_interval: Option<NonZeroU64>,
 }
 
 impl<E, A, S, V, R> Stateful<E, A, S, V, R>
@@ -146,6 +158,7 @@ where
                 plan: config.plan,
                 resolvers: config.resolvers,
                 sync_config: config.sync_config,
+                finalize_sync_interval: config.finalize_sync_interval,
             },
             Mailbox::new(sender),
         )
@@ -193,6 +206,7 @@ where
             artifact: None,
             resolvers: self.resolvers,
             sync_completed,
+            finalize_sync_interval: self.finalize_sync_interval,
         };
         let _ = join!(syncer.start(), syncing.start());
     }
@@ -215,7 +229,13 @@ where
         self.resolvers.attach_databases(databases.clone()).await;
 
         let processor_metrics = ProcessorMetrics::new(self.context.child("processor"));
-        let processor = Processor::new(self.application, databases, anchor, processor_metrics);
+        let processor = Processor::new(
+            self.application,
+            databases,
+            anchor,
+            processor_metrics,
+            self.finalize_sync_interval,
+        );
         Processing {
             context: self.context,
             mailbox: self.mailbox,
@@ -391,6 +411,7 @@ mod tests {
                         update_channel_size: NZUsize!(1),
                         max_retained_roots: 1,
                     },
+                    finalize_sync_interval: None,
                 },
             );
             let handle = stateful.start();
--- a/glue/src/stateful/actor/core/mod.rs
+++ b/glue/src/stateful/actor/core/mod.rs
@@ -39,6 +39,39 @@ mod syncing;
 
 type BlockDigest<A, E> = <<A as Application<E>>::Block as Digestible>::Digest;
 
+/// Periodic pruning configuration.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub struct PruneConfig {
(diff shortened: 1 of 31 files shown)


glue/src/stateful/actor/core/mod.rs
```diff
@@ -39,6 +42,17 @@ mod syncing;
 
 type BlockDigest<A, E> = <<A as Application<E>>::Block as Digestible>::Digest;
 
+/// Periodic database maintenance performed after finalization.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum MaintenanceInterval {
+    /// Persist in-memory database state every `interval` finalized blocks.
```
roberto-bayardo: Could you clarify what you mean by "Persist"?  I assume this is just "commit()", e.g. guarantee durability?

storage/src/merkle/persisted/full.rs
```diff
@@ -653,40 +653,54 @@ impl<F: Family, E: RStorage + Clock + Metrics, D: Digest, S: Strategy> Merkle<F,
...
 
-    /// Prune all nodes up to but not including the given leaf location and update the pinned nodes.
-    ///
-    /// This implementation ensures that no failure can leave the structure in an unrecoverable
-    /// state, requiring it sync the structure to write any potential unsynced updates.
-    ///
-    /// Returns [Error::LocationOverflow] if `loc` exceeds [Family::MAX_LEAVES].
-    /// Returns [Error::LeafOutOfBounds] if `loc` exceeds the current leaf count.
-    pub async fn prune(&mut self, loc: Location<F>) -> Result<(), Error<F>> {
+    async fn prune_synced(&mut self, loc: Location<F>, sync_journal: bool) -> Result<(), Error<F>> {
```
roberto-bayardo: I think it's worth having a comment for this function explaining why it exists.

storage/src/merkle/persisted/full.rs
```diff
@@ -653,40 +653,54 @@ impl<F: Family, E: RStorage + Clock + Metrics, D: Digest, S: Strategy> Merkle<F,
...
+    /// Returns [Error::LocationOverflow] if `loc` exceeds [Family::MAX_LEAVES].
+    /// Returns [Error::LeafOutOfBounds] if `loc` exceeds the current leaf count.
+    pub async fn prune(&mut self, loc: Location<F>) -> Result<(), Error<F>> {
+        // Flush items cached in the mem to disk to ensure the current state is recoverable.
+        self.sync().await?;
+        self.prune_synced(loc, false).await
+    }
+
+    /// Prune nodes after the caller has already synced appended nodes, then persist pruning state.
+    pub(crate) async fn prune_synced_and_sync(&mut self, loc: Location<F>) -> Result<(), Error<F>> {
```
roberto-bayardo: Hmm, is there a potential race condition introduced by this split? (What if someone makes changes between my calls...)

glue/src/stateful/actor/core/mod.rs
patrick-ogrady: I think `Persist` implies they aren't durable until then?

patrick-ogrady: (which AFAIK isn't true)

clabby: A bit more dumb; `ManagedDb::persist` exists to not clash with `sync` on the DBs the trait is being implemented on. Under the hood, this calls `sync` every `n` blocks.

We can ofc override with `ManagedDb::sync`, though you have tons of `clippy::use_self` lints that false-flag.

clabby: > (which AFAIK isn't true)

Correct. Every finalized block `commit`s changes, they're durable.

glue/src/stateful/actor/processor/mod.rs
```diff
@@ -604,21 +700,24 @@ where
...
 
         self.databases.finalize(batch).await;
-        self.app
-            .finalized(
-                (context.child("finalized"), block.context()),
-                &block,
-                &self.databases,
-            )
-            .await;
+        match self.maintenance.observe_finalized(height, sync_targets) {
```
patrick-ogrady: I think the optimal approach would be to actually call this when we aren't doing anything else rather than inline here? Maybe the performance delta of doing so is negligible but we should be able to take a few ms when we aren't blocking anything else.

glue/src/stateful/actor/core/mod.rs
```diff
@@ -39,6 +42,17 @@ mod syncing;
...
+    Persist(NonZeroU64),
+
+    /// Prune databases and marshal every `interval` finalized blocks, keeping
+    /// enough finalized history to safely rewind `max_pending_acks` blocks.
```
patrick-ogrady: As discussed, it is more than just `max_pending_acks` right? it is the value of the oldest block being pruned?

glue/src/stateful/actor/core/mod.rs
patrick-ogrady: I guess the confusing part is its really `Persist` and `PersistAndPrune`...in that scenario, it feels we should just have an interval for this with a prune bool?

glue/src/stateful/actor/core/mod.rs
clabby: We only ever prune to a `(block, target.start[])` that guarantees we've retained enough history in both marshal and QMDB to rewind to the oldest block necessary during startup reconciliation after a crash. No history behind that point is necessary.

glue/src/stateful/actor/processor/mod.rs
```diff
@@ -579,12 +669,13 @@ where
                 digest, self.last_processed.digest,
                 "received conflicting finalized block at processed height",
             );
-            return FinalizeStatus::Duplicate;
+            return (FinalizeStatus::Duplicate, MaintenanceAction::None);
```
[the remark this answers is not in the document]
clabby: This is fine. We'd just prune and/or sync later.

storage/src/merkle/persisted/full.rs
clabby: Yeah, it's possible to hit, though not through the external API. An internal caller could mess this up.

glue/src/stateful/actor/processor/mod.rs
```diff
@@ -698,6 +773,22 @@ where
...
+) where
+    E: Rng + Spawner + Metrics + Clock,
+    DBs: DatabaseSet<E>,
+    S: Scheme,
+    V: MarshalVariant,
+{
+    let (height, targets) = prune;
+    databases.prune(&targets).await;
+    marshal.prune(height);
+}
```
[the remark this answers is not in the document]
clabby: This crash condition is covered.

glue/src/stateful/actor/core/mod.rs
```diff
@@ -77,6 +77,13 @@ where
 
     /// Sync engine tuning knobs.
     pub sync_config: SyncEngineConfig,
+
+    /// Prune databases and marshal every `interval` finalized blocks.
+    ///
+    /// When enabled, glue retains a `max_pending_acks + 1` finalized-target
```
patrick-ogrady: In a follow-up PR, I'll extend this to be configurable (this delta isn't enough for a production deployment).

patrick-ogrady: https://github.com/commonwarexyz/monorepo/issues/3987

glue/src/stateful/actor/core/processing.rs
```diff
@@ -104,24 +136,35 @@ where
...
+                        self.processor.notify_finalized(self.context.as_present(), &block).await;
                         acknowledgement.acknowledge();
                         continue;
                     }
-                    if let FinalizeStatus::Persisted { height } =
-                        self.processor.finalize(&self.context, block).await
-                    {
+                    let (status, prune) = self.processor.finalize(&self.context, block).await;
+                    if let Some(prune) = prune {
+                        pending_prunes.push_back(prune);
```
patrick-ogrady: I think we just need to store the latest prune? we should be able to coalesce this work.

glue/src/stateful/actor/processor/mod.rs
```diff
@@ -627,7 +690,19 @@ where
         };
         timer.observe(context);
 
-        FinalizeStatus::Persisted { height }
+        (FinalizeStatus::Persisted { height }, prune)
+    }
+
+    /// Notify the application that marshal delivered a finalized block already
+    /// reflected in the database set.
+    pub(super) async fn notify_finalized(&mut self, context: &E, block: &A::Block) {
```
patrick-ogrady, line 749: I don't think this helper is necessary?

clabby: It's also invoked in `Processing`; Shortens the call-chain without having to make `app` `pub(super)`.

glue/src/stateful/actor/processor/mod.rs
```diff
@@ -89,59 +89,102 @@ pub(super) enum FinalizeStatus {
...
+impl<T> Prune<T> {
+    /// Run database and marshal pruning.
+    pub(super) async fn run<E, DBs, S, V>(self, databases: &mut DBs, marshal: &MarshalMailbox<S, V>)
+    where
+        E: Rng + Spawner + Metrics + Clock,
+        DBs: DatabaseSet<E, SyncTargets = T>,
+        S: Scheme,
+        V: MarshalVariant,
+    {
+        databases.prune(&self.qmdb_target).await;
```
patrick-ogrady, line 108: I "think" the right order may be marshal then qmdb (if we do qmdb first and crash, we may plan to access old stuff?)

glue/src/stateful/actor/core/mod.rs
```diff
@@ -39,6 +39,39 @@ mod syncing;
 
 type BlockDigest<A, E> = <<A as Application<E>>::Block as Digestible>::Digest;
 
+/// Periodic pruning configuration.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub struct PruneConfig {
+    /// Prune databases and marshal every `maintenance_interval` finalized blocks.
+    pub maintenance_interval: NonZeroUsize,
```
patrick-ogrady, line 50: I wonder if we can infer the right time to do this based on the config below?

glue/src/stateful/actor/core/mod.rs
```diff
@@ -39,6 +39,39 @@ mod syncing;
...
+
+    /// Extra finalized blocks to retain in marshal beyond `max_pending_acks + 1`.
+    ///
+    /// This should generally be set to a large enough number of blocks to facilitate downtime
+    /// on a validator that has completed state sync. If the prune cadence on marshal is too
+    /// aggressive, a rebooted node may fail to recover due to peers being unable to serve the
+    /// blocks it needs to catch up.
+    pub additional_marshal_blocks: usize,
```
patrick-ogrady: how do you feel about `extra_` for these? or `retained_`?

patrick-ogrady (commented):
small nits

glue/src/stateful/actor/processor/mod.rs
clabby: We shouldn't need to await marshal's prune here first, since QMDB will retain at least the safe rewind window. At worst, this just leaves around some extra marshal block / certificate history until the next pruning job comes through.

Assuming we do crash between QMDB prune and marshal pruning:
- We're safe on restart. Marshal and QMDB are still in-sync and QMDB retains the full active operation range for the tip (plus `max_pending_acks + 1 + retained_qmdb_blocks` blocks' worth of operations.)
- We're safe when serving peers for any target range QMDB has retained. If this node can no longer
  serve an older request, the requester will either fetch from another peer or continue advancing and
  ask for newer targets that the active validator set can serve. This is the same expected behavior as
  restarting after a state-sync crash and being down long enough for old ranges to age out.

glue/src/stateful/actor/core/mod.rs
```diff
@@ -39,6 +39,38 @@ mod syncing;
...
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub struct PruneConfig {
+    /// Finalized blocks to retain in marshal beyond `max_pending_acks + 1`.
+    ///
+    /// This should generally be set to a large enough number of blocks to facilitate downtime
+    /// on a validator that has completed state sync. If the prune cadence on marshal is too
+    /// aggressive, a rebooted node may fail to recover due to peers being unable to serve the
+    /// blocks it needs to catch up.
+    ///
+    /// This value is also used as the pruning cadence.
```
patrick-ogrady: I think the cadence should be set to the minimum of the two?

glue/src/stateful/actor/processor/mod.rs
patrick-ogrady: I suppose we'll try to serve any database request we can (rather than bounding by retained marshal blocks), so this appears 6 one way and half a dozen the other.

glue/src/stateful/actor/core/mod.rs
patrick-ogrady: I suppose it should really just be "another thing" like you had it where we prune but never leaving less than the configued.

patrick-ogrady (commented):
some small tweaks I'll make.

glue/src/stateful/actor/processor/mod.rs
patrick-ogrady: Now that we separate pruning between qmdb/marshal and enforce qmdb retain is <= marshal, this order is now very reasonable.

PR #3966, opened 2026-06-05
[runtime] Document and test read-after-remove semantics for `Storage`

danlaine:
Documents expected behavior of runtime storage when a blob or its partition are removed while handles still exist. Adds unit tests to assert this behavior.


roberto-bayardo:
mind adding a PR description?

danlaine:
> mind adding a PR description?

done

runtime/src/lib.rs
```diff
@@ -714,6 +714,19 @@ stability_scope!(BETA {
...
+        ///
+        /// Removal unlinks the blob's name but does not invalidate previously opened handles:
+        /// they remain fully readable until dropped, whether the blob was removed by name or by
+        /// removing its entire partition. This includes bytes written but never synced. Physical
+        /// resources are reclaimed once the last handle is dropped.
+        ///
+        /// Re-opening a removed blob's name creates a new, independent blob; handles opened
+        /// before the removal continue to observe the removed blob's contents.
+        ///
+        /// Writing to or syncing a removed blob is unspecified: implementations may succeed or
```
roberto-bayardo: Should we also mention `rewind` here?

danlaine: `rewind` is a journal-level operation but I generalized this comment to include `resize` (the blob method used by `rewind`)

roberto-bayardo (approved):

PR #3967, opened 2026-06-05
[storage] Explore Prune and Sync

patrick-ogrady:
Related: #3963
Related: #3965


glue/src/stateful/db/current.rs
```diff
@@ -1139,6 +1139,11 @@ mod tests {
                     .unwrap();
             }
 
+            let metrics = context.encode();
```
patrick-ogrady, line 1142: I don't think this is necessary

PR #3969, opened 2026-06-05
[storage] `Persistable::commit` and sync take `&mut self`

danlaine:
`Persistable::commit` and `sync` mutate durability state, but the trait takes `&self`, so every implementor recreates exclusivity at runtime: the contiguous journals serialize commit against mutations with an internal `op_lock` mutex, and every other implementor either locks or relies on callers not to race. The honest signature is `&mut self`: it states the exclusivity requirement once, in the type, and lets implementations drop the machinery that fakes it.

This also unblocks an upcoming journal change in which `sync` mutates plain fields and cannot be implemented behind `&self` at all.


roberto-bayardo:
The rust file system API doesn't treat sync this way (sync_all and sync_data take &self). Given the heavyweight nature of the underlying fsync calls, I think &self is the right API, even if it requires interior mutability to properly manage. This doesn't necessarily preclude storage structures that have an &mut sync type method, I'd just argue they shouldn't be implementing this trait.

danlaine:
> The rust file system API doesn't treat sync this way (sync_all and sync_data take &self). Given the heavyweight nature of the underlying fsync calls, I think &self is the right API, even if it requires interior mutability to properly manage.

I think `runtime::Blob` is the better point of comparison for `std::fs::file` since that trait / its implementation mostly wraps a file. That trait does have `sync(&self)`. The rationale for `std::fs::File` having `sync(&self)` instead of `sync(&mut self)`, I think, is that `sync` doesn't actually mutate any Rust-owned state. It just makes a syscall and all the state changes are owned by the kernel. That pattern doesn't necessarily hold for `Persistable` implementations -- they may mutate Rust-owned state, in which case I think `&mut self` is more "honest".

> This doesn't necessarily preclude storage structures that have an &mut sync type method, I'd just argue they shouldn't be implementing this trait.

Perhaps we should re-think the proper organization of functionality into the storage traits. i.e. is `Persistable` even a coherent set of functionality that makes sense to package into a trait? But while it exists, it seems perhaps overly strict to require that types which are `Persistable` either have non-mutating `sync` implementations or use interior mutability.

roberto-bayardo:
You want to just merge this into your journal PRs now that Persistable is gone?

roberto-bayardo:
(I was hoping we wouldn't need to change all previous implementors to &mut sync)

patrick-ogrady:

PR #3970, opened 2026-06-05
[storage/journal/contiguous] replace fixed-journal locks with ownership and snapshot readers

danlaine:
## Motivation

The fixed journal guarded its state behind an `AsyncRwLock<Inner>` plus an `AsyncMutex` op-lock. A `Reader` was a borrow that held the read guard for its entire lifetime. That coupled reads to writes: any open reader — a replay stream, a slow consumer, a proof builder — blocked every `append`, `prune`, and `rewind` until it was dropped. The lock was also a contention point and a source of structural complexity (an `Inner` type, guards, and lifetimes on the read path).

This PR removes the locks. The journal owns its state outright and mutates it through `&mut self`; readers become cheap, owned snapshots whose only *mutable* shared state with the journal is a single atomic counter (they also hold a refcounted, immutable handle slice — see below). Reads never block writes, writes never block reads, and a reader can be moved and held for as long as the caller likes.

## The model

`Journal::reader()` captures an owned `Snapshot`: the current sealed-blob handles (an `Arc<[Sealed]>`, cloned by bumping a refcount), a read handle for the writable tail, and the journal's current `bounds`. The snapshot is frozen at that instant — the journal can append, seal, prune, or rewind afterward and the snapshot's view does not move.

This works because the journal never mutates a live sealed-blob slice in place; it replaces the slice wholesale on every add or drop. A reader holding the old `Arc` keeps a valid, consistent view, and the runtime's read-after-remove guarantee keeps a pruned blob's bytes readable until the last handle drops. So pruning the journal does not invalidate an existing reader — it only stops *new* readers from seeing those positions.

The one piece of genuinely shared mutable state is an `AtomicUsize` count of live readers, and it exists to gate exactly one operation (see `rewind` below).

Two responsibilities are lifted out of `fixed.rs` into focused modules:

- **`blobs.rs`** owns the blob lifecycle: the contiguous run of sealed blobs ending in one writable tail, the seal/prune/rewind/recover operations, and the `Snapshot` it hands to readers. (The fixed journal now calls its backing files "blobs"; the variable journal still says "section" for the same concept.)
- **`checkpoint.rs`** owns the durable recovery record — the mid-blob pruning boundary, the recovery watermark, and any in-progress clear intent — backed by `Metadata`'s atomic two-blob commit.

## Behavior changes (relative to base)

- **Readers no longer block the journal.** Previously an open reader held the read lock and stalled writers; now reads and writes proceed fully concurrently.
- **Readers are frozen snapshots.** A reader observes the `bounds` captured at `reader()` and never sees later appends — take a new reader for those. Conversely, **a reader can still read items the journal prunes after the reader was created**, because it holds its own handles.
- **`rewind` can now fail with `Error::BlobInUse`.** Rewinding into an already-*sealed* blob truncates that file in place, which would tear a concurrent reader's bytes, so it is refused (not blocked) while any reader snapshot is outstanding; retry once readers drop. Rewinding within the writable tail is always allowed, and `prune`/`clear` are never gated (they only drop whole blobs, leaving readers' handles intact).
- **`read_many` returns `Error::PositionsNotIncreasing` instead of panicking** on non-strictly-increasing positions, in both the fixed and variable journals. Callers already pass sorted positions; this converts a process-crashing precondition into a recoverable error.
- **`Reader` is owned, not borrowed.** `reader()` no longer returns a guard bound to a lifetime, so the returned reader can be stored and moved.

The **on-disk format is unchanged**: blob layout is identical, legacy partitions still open, and checkpoint entries keep their byte-compatible framing.

## Performance Changes

Fixed journal benchmarks show improvement.

```
  Fixed journal

  fixed_append

  ┌───────────────┬──────────┬──────────┬────────┐
  │    params     │   main   │   HEAD   │ delta  │
  ├───────────────┼──────────┼──────────┼────────┤
  │ items=1000    │   8.80ms │   7.67ms │ −12.8% │
  ├───────────────┼──────────┼──────────┼────────┤
  │ items=10000   │  22.00ms │  17.97ms │ −18.3% │
  ├───────────────┼──────────┼──────────┼────────┤
  │ items=100000  │ 131.75ms │ 125.27ms │  −4.9% │
  ├───────────────┼──────────┼──────────┼────────┤
  │ items=1000000 │   1.153s │   1.121s │  −2.7% │
  └───────────────┴──────────┴──────────┴────────┘

  fixed_read_random

  ┌─────────────────────────┬─────────┬─────────┬────────┐
  │         params          │  main   │  HEAD   │ delta  │
  ├─────────────────────────┼─────────┼─────────┼────────┤
  │ serial items=100        │  10.3us │   9.2us │ −10.4% │
  ├─────────────────────────┼─────────┼─────────┼────────┤
  │ serial items=1000       │ 121.1us │ 108.9us │ −10.1% │
  ├─────────────────────────┼─────────┼─────────┼────────┤
  │ serial items=10000      │  1.64ms │  1.50ms │  −8.3% │
  ├─────────────────────────┼─────────┼─────────┼────────┤
  │ serial items=100000     │ 539.3ms │ 521.0ms │  −3.4% │
  ├─────────────────────────┼─────────┼─────────┼────────┤
  │ concurrent items=100    │  22.1us │  19.7us │ −11.0% │
  ├─────────────────────────┼─────────┼─────────┼────────┤
  │ concurrent items=1000   │ 234.0us │ 203.3us │ −13.1% │
  ├─────────────────────────┼─────────┼─────────┼────────┤
  │ concurrent items=10000  │  3.10ms │  2.56ms │ −17.5% │
  ├─────────────────────────┼─────────┼─────────┼────────┤
  │ concurrent items=100000 │ 203.3ms │ 182.7ms │ −10.1% │
  ├─────────────────────────┼─────────┼─────────┼────────┤
  │ read_many items=100     │   5.1us │   4.5us │ −10.8% │
  ├─────────────────────────┼─────────┼─────────┼────────┤
  │ read_many items=1000    │  59.9us │  51.9us │ −13.3% │
  ├─────────────────────────┼─────────┼─────────┼────────┤
  │ read_many items=10000   │ 765.5us │ 677.7us │ −11.5% │
  ├─────────────────────────┼─────────┼─────────┼────────┤
  │ read_many items=100000  │  65.3ms │  61.3ms │  −6.1% │
  └─────────────────────────┴─────────┴─────────┴────────┘

  fixed_read_sequential

  ┌──────────────┬─────────┬─────────┬───────┐
  │    params    │  main   │  HEAD   │ delta │
  ├──────────────┼─────────┼─────────┼───────┤
  │ items=1000   │  83.2us │  77.6us │ −6.7% │
  ├──────────────┼─────────┼─────────┼───────┤
  │ items=10000  │ 815.1us │ 782.2us │ −4.0% │
  ├──────────────┼─────────┼─────────┼───────┤
  │ items=100000 │  7.98ms │  7.73ms │ −3.1% │
  ├──────────────┼─────────┼─────────┼───────┤
  │ items=500000 │ 40.29ms │ 38.12ms │ −5.4% │
  └──────────────┴─────────┴─────────┴───────┘

  fixed_replay (biggest win)

  ┌─────────────────────────────┬─────────┬─────────┬────────┐
  │           params            │  main   │  HEAD   │ delta  │
  ├─────────────────────────────┼─────────┼─────────┼────────┤
  │ items=1000 buffer=16384     │  50.9us │  10.5us │ −79.4% │
  ├─────────────────────────────┼─────────┼─────────┼────────┤
  │ items=1000 buffer=65536     │  27.9us │  10.7us │ −61.5% │
  ├─────────────────────────────┼─────────┼─────────┼────────┤
  │ items=1000 buffer=1048576   │  27.1us │  10.8us │ −60.3% │
  ├─────────────────────────────┼─────────┼─────────┼────────┤
  │ items=10000 buffer=16384    │ 501.3us │ 104.5us │ −79.2% │
  ├─────────────────────────────┼─────────┼─────────┼────────┤
  │ items=10000 buffer=65536    │ 228.2us │ 107.1us │ −53.1% │
  ├─────────────────────────────┼─────────┼─────────┼────────┤
  │ items=10000 buffer=1048576  │ 182.1us │ 104.8us │ −42.4% │
  ├─────────────────────────────┼─────────┼─────────┼────────┤
  │ items=100000 buffer=16384   │  5.33ms │  4.40ms │ −17.3% │
  ├─────────────────────────────┼─────────┼─────────┼────────┤
  │ items=100000 buffer=65536   │  2.30ms │  1.96ms │ −14.8% │
  ├─────────────────────────────┼─────────┼─────────┼────────┤
  │ items=100000 buffer=1048576 │  1.79ms │  1.49ms │ −17.1% │
  ├─────────────────────────────┼─────────┼─────────┼────────┤
  │ items=500000 buffer=16384   │ 26.96ms │ 23.38ms │ −13.3% │
  ├─────────────────────────────┼─────────┼─────────┼────────┤
  │ items=500000 buffer=65536   │ 11.88ms │  9.99ms │ −15.9% │
  ├─────────────────────────────┼─────────┼─────────┼────────┤
  │ items=500000 buffer=1048576 │  9.19ms │  7.55ms │ −17.9% │
  └─────────────────────────────┴─────────┴─────────┴────────┘
  ```

### CODE REVIEW

File: storage/src/journal/contiguous/fixed.rs [hunk 1]
```diff
@@ -178,145 +195,233 @@ pub struct Config {
...
+            Self::Sealed(s) => s.try_read_sync(offset, buf),
+            Self::Tail(t) => t.try_read_sync(offset, buf),
+        }
+    }
+}
+
+/// An owned copy of the journal's blob handles and bounds, captured when a [Reader] is
+/// created. The journal can move on; the snapshot's view never changes. Counts itself in
+/// `readers` for its lifetime; the count gates in-place truncation during rewind.
+pub(super) struct Snapshot<B: Blob> {
```
roberto-bayardo: Consider putting this struct within blobs.rs, then Blobs could have a factory function
`fn to_snapshot(bounds: ..)` (or someting like that) that creates it.

(`bounds` is the only part of this that isn't just copied from the Blobs struct it seems)

storage/src/journal/contiguous/fixed.rs
```diff
@@ -1002,112 +1123,93 @@ impl<E: Context, A: CodecFixedShared> Journal<E, A> {
...
+    pub fn reader(&self) -> Reader<E, A> {
         Reader {
-            guard: self.inner.read().await,
+            snapshot: Snapshot::new(
+                self.blobs.oldest_blob(),
+                self.blobs.sealed(),
+                self.blobs.tail_reader(),
+                self.bounds.clone(),
+                self.blobs.readers(),
+            ),
```
roberto-bayardo: `snapshot: self.blobs.to_snapshot(self.bounds),` ?

storage/src/journal/contiguous/fixed.rs
```diff
@@ -178,145 +195,233 @@ pub struct Config {
...
+/// the first invalid data read will be considered the new end of the journal (and the
+/// underlying blob will be truncated to the last valid item). Repair is performed during init.
+pub struct Journal<E: Context, A> {
+    /// The blobs that comprise the journal.
+    blobs: Blobs<E>,
 
-    /// Total number of items appended (not affected by pruning).
-    size: u64,
+    /// The durable recovery checkpoint.
+    checkpoint: Checkpoint<E>,
```
roberto-bayardo, line 217: love this

storage/src/journal/contiguous/fixed.rs
```diff
@@ -1002,112 +1123,93 @@ impl<E: Context, A: CodecFixedShared> Journal<E, A> {
...
+                self.bounds.end,
+            )
+            .await
     }
 
-    /// Acquire a reader guard that holds a consistent view of the journal.
-    pub async fn reader(&self) -> Reader<'_, E, A> {
+    /// Return an owned snapshot of the journal. Bounds are frozen at creation; take a new
+    /// reader to observe later appends.
+    pub fn reader(&self) -> Reader<E, A> {
```
roberto-bayardo, line 726: Do you think it's worth allowing a lowerbound to avoid having to copy a bunch of blob handles we'll never use (e.g. inactive blobs in QMDB)?

storage/src/journal/contiguous/fixed.rs
danlaine: moved and added `to_snapshot`

storage/src/journal/contiguous/fixed.rs
danlaine: done

storage/src/journal/contiguous/blobs.rs
```diff
@@ -0,0 +1,537 @@
...
+            .checked_sub(self.oldest_blob_index)
+            .map(|idx| idx as usize)
+            .filter(|&idx| idx < self.sealed.len())
+            .ok_or_else(|| Error::Corruption(format!("rewind target blob {blob} not retained")))?;
+
+        // Refuse to truncate in place while any reader could be mid-read. Each live reader counts
+        // itself here; the Acquire load pairs with the Release drop, so a count of 0 guarantees
+        // every reader has finished and dropped before we start overwriting bytes.
+        if self.readers.load(Ordering::Acquire) != 0 {
+            return Err(Error::BlobInUse(blob));
```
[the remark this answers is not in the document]
danlaine: The increment and the gate check can never overlap: reader() borrows the journal as &self while rewind needs &mut self, so by the time rewind can even be called, every increment has already finished. A finished increment is always visible to the check — trivially on the same thread, and across threads because the only way &mut Journal reaches another thread is through something that synchronizes (spawn, channel, mutex), which also carries the counter update with it. So Relaxed is enough there.

The one thing that genuinely can happen concurrently with rewind is a snapshot being dropped on another thread — and that's the case the orderings are for: the drop decrements with Release, the gate loads with Acquire, so if rewind reads 0, every read that snapshot performed is guaranteed to have completed before we truncate.

storage/src/journal/contiguous/fixed.rs
danlaine: sealed is an Arc<[Sealed<B>]> -- we're not copying the sealed blobs individually

storage/src/journal/contiguous/fixed.rs
```diff
@@ -1316,78 +946,414 @@ impl<E: Context, A: CodecFixedShared> Journal<E, A> {
...
+        let _timer = self.metrics.read_many_timer();
+        self.metrics.read_many_calls.inc();
+        let mut prev: Option<u64> = None;
+        for &pos in positions {
+            if prev.is_some_and(|p| pos <= p) {
+                return Err(Error::PositionsNotIncreasing);
+            }
+            prev = Some(pos);
+            self.snapshot.validate_readable(pos)?;
+        }
```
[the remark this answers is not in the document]
danlaine: We don't return use an `assert` but we return error `PositionsNotIncreasing` to enforce the precondition

storage/src/journal/contiguous/fixed.rs
roberto-bayardo: Got it. So let's consider an archival node though... maybe there are 100+ files. Is that Arc clone cost still insignificant if I'm just reading a few elements from tip?

roberto-bayardo: I guess the model is we have a periodically refreshed shared-reader that can service concurrent requests, rather than each request require its own reader.

storage/src/journal/contiguous/fixed.rs
```diff
@@ -1319,78 +982,363 @@ impl<E: Context, A: CodecFixedShared> Journal<E, A> {
...
+        let tail_batches = TailReplayState::<E::Blob, A> {
+            reader: self.snapshot.tail_reader.clone(),
+            positions: tail_start.max(tail_first)..self.snapshot.bounds.end,
+            tail_first,
+            items_per_batch: (buffer.get() / chunk_size).max(1),
+            _marker: PhantomData,
+        }
+        .into_batches();
+
+        Ok(sealed_batches.chain(tail_batches).flat_map(stream::iter))
```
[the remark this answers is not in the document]
danlaine: The Stream can't outlive the reader so the "dangerous" code doesn't compile

patrick-ogrady:

PR #3971, opened 2026-06-05
[storage/merkle/persisted] flush merkleized nodes from RAM on authenticated journal commit

roberto-bayardo:
The in-memory footprint of the authenticated journal grows indefinitely until `sync` is called.  This change makes the new merkleized nodes get flushed to disk (without fsycing) on each commit to limit this memory growth between syncs.

The flush adds ~no commit latency since it runs concurrently with the item journal's fsync, and it shrinks sync()'s critical path: sync now only writes nodes accumulated since the last commit rather than since the last sync.

Also optimizes the flush path, and adds a new benchmark to evalute the optimizations:

## Benchmark: `merkle::flush` (optimized vs. baseline)
  
  | Family | n (leaves) | Time (median) | Δ vs. baseline | Significance |
  |--------|-----------:|--------------:|---------------:|--------------|
  | mmr | 10,000 | 126.98 µs | **−24.3%** | p < 0.05 |
  | mmr | 100,000 | 1.1885 ms | **−26.5%** | p < 0.05 |
  | mmb | 10,000 | 139.36 µs | **−21.5%** | p < 0.05 (2 high outliers) |
  | mmb | 100,000 | 1.2823 ms | **−21.3%** | p < 0.05 (1 outlier) |
  
  <details>
  <summary>Full confidence intervals</summary>

  | Family | n | Time CI [low, mid, high] | Change CI [low, mid, high] |
  |--------|--:|--------------------------|----------------------------|
  | mmr | 10,000 | [124.71, 126.98, 130.36] µs | [−26.0%, −24.3%, −22.7%] |
  | mmr | 100,000 | [1.1644, 1.1885, 1.2100] ms | [−27.6%, −26.5%, −25.4%] |
  | mmb | 10,000 | [129.24, 139.36, 153.35] µs | [−25.6%, −21.5%, −17.1%] |
  | mmb | 100,000 | [1.2036, 1.2823, 1.4048] ms | [−24.2%, −21.3%, −17.4%] |

  </details>

  A consistent **~21–27% reduction in flush time**, significant (p < 0.05) across both families and batch sizes. Removing the `Vec<D>` intermediate
  and the second serialization pass in `flush_internal` — encoding the in-memory tail directly to the journal's byte buffer in a single pass —
  accounts for the win; the flush path was more copy-bound than hashing/IO-bound. The mmb runs show wider intervals (timer noise at `sample_size =
  10`), but the effect is unambiguous (lower CI bound still ≥ 17%).

### DIFF
--- a/storage/src/journal/authenticated.rs
+++ b/storage/src/journal/authenticated.rs
@@ -340,10 +340,18 @@ where
     H: Hasher,
     S: Strategy,
 {
-    /// Durably persist the journal. This is faster than `sync()` but does not persist the Merkle
-    /// structure, meaning recovery will be required on startup if we crash before `sync()`.
+    /// Durably persist the journal. This is faster than `sync()` but does not guarantee that the
+    /// Merkle structure is durably persisted, meaning recovery may be required on startup in the
+    /// event of a crash.
     pub async fn commit(&self) -> Result<(), Error<F>> {
-        self.journal.commit().await.map_err(Error::Journal)
+        // Though not necessary for recovery, we flush the merkle structure (without syncing it) to
+        // limit memory bloat.
+        try_join!(
+            self.journal.commit().map_err(Error::Journal),
+            self.merkle.flush().map_err(Error::Merkle)
+        )?;
+
+        Ok(())
     }
 }
 
--- a/storage/src/journal/contiguous/fixed.rs
+++ b/storage/src/journal/contiguous/fixed.rs
@@ -1135,9 +1135,8 @@ impl<E: Context, A: CodecFixedShared> Journal<E, A> {
         if items.is_empty() {
             return Err(Error::EmptyAppend);
         }
-        let items_count = items.len();
         let items_buf = Self::encode_items(items);
-        self.write_encoded(items_buf, items_count).await
+        self.write_encoded(items_buf).await
     }
 
     /// Encode `items` into an owned fixed-width byte buffer.
@@ -1163,24 +1162,20 @@ impl<E: Context, A: CodecFixedShared> Journal<E, A> {
         items_buf
     }
 
-    /// Append `items_count` records pre-encoded by [`Self::encode_items`] into `items_buf`.
+    /// Append records pre-encoded by [`Self::encode_items`] into `items_buf`.
     ///
     /// Records no metrics: the public [`append`](Self::append) / [`append_many`](Self::append_many)
     /// wrappers record their own, and callers that encode separately record whatever fits their
     /// operation.
-    pub(crate) async fn write_encoded(
-        &self,
-        items_buf: Vec<u8>,
-        items_count: usize,
-    ) -> Result<u64, Error> {
+    pub(crate) async fn write_encoded(&self, items_buf: Vec<u8>) -> Result<u64, Error> {
+        assert!(
+            items_buf.len().is_multiple_of(A::SIZE),
+            "encoded buffer length must be a multiple of the item size"
+        );
+        let items_count = items_buf.len() / A::SIZE;
         if items_count == 0 {
             return Err(Error::EmptyAppend);
         }
-        assert_eq!(
-            items_buf.len(),
-            items_count * A::SIZE,
-            "encoded buffer length must match item count"
-        );
 
         let _op_guard = self.op_lock.lock().await;
         let mut inner = self.inner.write().await;
--- a/storage/src/journal/contiguous/fixed.rs
+++ b/storage/src/journal/contiguous/fixed.rs
@@ -1141,7 +1141,7 @@ impl<E: Context, A: CodecFixedShared> Journal<E, A> {
     }
 
     /// Encode `items` into a single contiguous buffer using the fixed-width record layout, ready to
-    /// be appended via [`Self::append_encoded`]. This is synchronous and holds no locks, so a caller
+    /// be appended via [`Self::write_encoded`]. This is synchronous and holds no locks, so a caller
     /// can encode borrowed data (e.g. under a read lock) and release the borrow before the append.
     ///
     /// Uses `Write::write` directly to avoid the per-item allocations of `Encode::encode`.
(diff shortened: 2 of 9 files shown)


storage/src/journal/contiguous/fixed.rs
```diff
@@ -1135,28 +1135,54 @@ impl<E: Context, A: CodecFixedShared> Journal<E, A> {
...
+        self.metrics.append_many_calls.inc();
+        self.write_encoded(items_buf, items_count).await
+    }
+
+    /// Append `items_count` pre-encoded records into `items_buf`, without recording metrics.
+    async fn write_encoded(&self, items_buf: Vec<u8>, items_count: usize) -> Result<u64, Error> {
+        if items_count == 0 {
+            return Err(Error::EmptyAppend);
+        }
+        debug_assert_eq!(items_buf.len(), items_count * A::SIZE);
```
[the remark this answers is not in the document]
roberto-bayardo: fixed

storage/src/journal/contiguous/fixed.rs
```diff
@@ -1135,28 +1135,63 @@ impl<E: Context, A: CodecFixedShared> Journal<E, A> {
...
+    }
+
+    /// Append `items_count` records pre-encoded by [`Self::encode_items`] into `items_buf`,
+    /// recording batch-append metrics (the counterpart to [`Self::append_many`] for callers that
+    /// encode separately).
+    ///
+    /// The only caller is the ALPHA-stability `merkle` module, which is excluded under the BETA
+    /// stability check (`commonware_stability_BETA`) while this BETA-stability journal remains;
+    /// allow dead code there so the annotation check does not flag this internal helper.
+    #[cfg_attr(commonware_stability_BETA, allow(dead_code))]
```
patrick-ogrady: This is the wrong way to do this.

storage/src/journal/contiguous/fixed.rs
```diff
@@ -1135,28 +1135,63 @@ impl<E: Context, A: CodecFixedShared> Journal<E, A> {
...
             }
         }
+        items_buf
+    }
+
+    /// Append `items_count` records pre-encoded by [`Self::encode_items`] into `items_buf`,
+    /// recording batch-append metrics (the counterpart to [`Self::append_many`] for callers that
+    /// encode separately).
+    ///
+    /// The only caller is the ALPHA-stability `merkle` module, which is excluded under the BETA
```
patrick-ogrady: Remove this comment

storage/src/journal/contiguous/fixed.rs
```diff
@@ -1135,28 +1135,63 @@ impl<E: Context, A: CodecFixedShared> Journal<E, A> {
...
-        };
-        let mut items_buf = Vec::with_capacity(items_count * A::SIZE);
-        match &items {
+    /// Encode `items` into a single contiguous buffer using the fixed-width record layout, ready to
+    /// be appended via [`Self::append_encoded`]. This is synchronous and holds no locks, so a caller
+    /// can encode borrowed data (e.g. under a read lock) and release the borrow before the append.
+    ///
+    /// Uses `Write::write` directly to avoid the per-item allocations of `Encode::encode`.
+    pub(crate) fn encode_items(items: Many<'_, A>) -> Vec<u8> {
+        let mut items_buf = Vec::with_capacity(items.len() * A::SIZE);
```
patrick-ogrady: Don't we already get this with the journal's write buffer?

storage/src/merkle/persisted/full.rs
```diff
@@ -593,15 +594,39 @@ impl<F: Family, E: RStorage + Clock + Metrics, D: Digest, S: Strategy> Merkle<F,
         futures::future::try_join_all(futs).await
     }
 
-    /// Sync the structure to disk.
+    /// Flush all nodes cached in the in-memory structure to the journal without forcing them to
+    /// disk. Flushed nodes are pruned from the in-memory structure and remain readable through the
+    /// journal, but they are not guaranteed to survive a crash until [Self::sync] is called.
+    pub async fn flush(&self) -> Result<(), Error<F>> {
```
patrick-ogrady, line 600: I think we can remove this if we just use the journal append buffer?

storage/src/merkle/persisted/full.rs
```diff
@@ -593,15 +594,39 @@ impl<F: Family, E: RStorage + Clock + Metrics, D: Digest, S: Strategy> Merkle<F,
...
+        let mut journal_dirty = self.sync_lock.lock().await;
+        self.flush_internal(&mut journal_dirty).await?;
+
+        // Sync the journal while still holding the sync_lock to ensure durability before
+        // returning. This covers nodes appended by the flush above as well as nodes left
+        // non-durable by earlier [Self::flush] calls.
+        if *journal_dirty {
+            self.journal.sync().await?;
+            *journal_dirty = false;
+        }
```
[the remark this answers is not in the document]
patrick-ogrady: This is not an issue because of the runtime sync().

patrick-ogrady (commented):
big win

storage/src/journal/authenticated.rs
```diff
@@ -356,10 +356,18 @@ where
...
     pub async fn commit(&self) -> Result<(), Error<F>> {
-        self.journal.commit().await.map_err(Error::Journal)
+        // Though not necessary for recovery, we flush the merkle structure (without syncing it) to
+        // limit memory bloat.
+        try_join!(
+            self.journal.commit().map_err(Error::Journal),
+            self.merkle.flush().map_err(Error::Merkle)
+        )?;
+
+        Ok(())
```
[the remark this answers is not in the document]
roberto-bayardo: this is fine if flush fails there are bigger problems

storage/src/merkle/persisted/full.rs
roberto-bayardo: If I'm understanding the comment, the use of the mem-mmr as a buffer predates the addition of the append buffer.  Changing it to immediately push bytes into the append buffer instead of delaying until flush() would be a bigger change and I'm not entirely sure how it would affect performance (the in-mem MMR accesses are going to be faster than parsing form a cached journal page). Should this be done as a separate cleanup?

storage/src/journal/contiguous/fixed.rs
roberto-bayardo: lol yeah, fixed.

storage/src/journal/contiguous/fixed.rs
roberto-bayardo: done

storage/src/journal/contiguous/fixed.rs
roberto-bayardo: Not sure I fully understand, but tweaked the comment a bit to make it a bit more clear?

storage/src/merkle/persisted/full.rs
patrick-ogrady: Ah, so there is no append buffer this goes into? In which case, this is fine (just wanted to avoid staging just to copy).

storage/src/journal/contiguous/fixed.rs
patrick-ogrady: this was just regarding the append buffer (which it looks like we don't need to consider).

storage/src/merkle/persisted/full.rs
roberto-bayardo: So I'm exploring one optimization where if the data would overflow append buffer, it will get written directly to the blob in an optimized path.  It's a bit invasive a change though, but it does help with large writes another 15% or so.  Given that this is mostly orthogonal to the existing work in this PR I'll handle it as a follow up if that's OK.

patrick-ogrady: Sounds good.

patrick-ogrady (approved):
I was under the impression waiting to write these nodes allows us to avoid writing some (but that's not true)...so writing when we write OPs is the right approach.

PR #3981, opened 2026-06-08
[glue/stateful] Restore sync completion metric

clabby:
## Overview

> [!NOTE]
> PR Stack (merge in reverse order):
> - #3998
> - #3984 
> - #3981 👈
> - #3965
> - `main`

Somewhere along the line, the sync completion metric was removed from `glue::stateful`. Adds it back.

PR #3984, opened 2026-06-08
[glue/stateful] Simplify resolver keep-alive

clabby:
## Overview

> [!NOTE]
> PR Stack (merge in reverse order):
> - #3998
> - #3984 👈
> - #3981
> - #3965
> - `main`

Simplifies the resolver keep-alive in the `Processing` mode by leaking the resolver mailboxes rather than having them be held by the `Processor`.

### DIFF
--- a/glue/src/stateful/actor/core/mod.rs
+++ b/glue/src/stateful/actor/core/mod.rs
@@ -275,8 +275,7 @@ where
         // so that this instance can serve peers database operations and proofs.
         self.resolvers.attach_databases(databases.clone()).await;
 
-        // Leak the resolver mailboxes to keep the resolvers alive to serve other peers.
-        std::mem::forget(self.resolvers);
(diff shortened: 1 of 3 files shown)


glue/src/stateful/actor/core/mod.rs
```diff
@@ -236,6 +236,9 @@ where
         // so that this instance can serve peers database operations and proofs.
         self.resolvers.attach_databases(databases.clone()).await;
 
+        // Leak the resolver mailboxes to keep the resolvers alive to serve other peers.
+        std::mem::forget(self.resolvers);
```
patrick-ogrady: I don't think this is a good pattern (someone could drop glue and these would remain). I think its better to find a way to handle with Arcs or something.

patrick-ogrady:
Replaced by #3998

PR #3988, opened 2026-06-08
[storage/qmdb] always batch-hash leaves in parallel

roberto-bayardo:
## [storage/qmdb] Hash operation leaf digests in parallel across all db types
  
  Move per-operation leaf hashing off the sequential `batch.add()` loop and onto the configured parallel `Strategy`, so leaf digests are computed
  concurrently before being appended. Previously only parent-node hashing (inside `merkleize`) was parallel; leaf hashing was the remaining sequential
  gap.

  ### Changes

  - **`compact` dbs** (`qmdb/compact/batch.rs::merkleize_ops`): compute leaf digests via `Strategy::map_init_collect_vec`, then append precomputed
  digests.
  - **`immutable` / `keyless` dbs**: route their `merkleize` through the authenticated journal's parallel `merkleize_with` instead of the `add`-loop,
  matching what `any`/`current` already did. Also drops a per-operation `Operation::clone()`.
  - **New `merkle::batch::UnmerkleizedBatch::add_leaf_digests`**: appends a run of precomputed leaf digests in one call, reserving capacity up front.
  Replaces the repeated `for d in digests { batch.add_leaf_digest(d) }` pattern in `merkleize_ops`, `merkleize_with`, and
  `current::build_grafted_tree`. Mirrored on the `persisted::compact` wrapper.

### DIFF
--- a/storage/src/journal/authenticated.rs
+++ b/storage/src/journal/authenticated.rs
@@ -115,38 +115,49 @@ impl<F: Family, H: Hasher, Item: Encode + Send + Sync, S: Strategy>
         })
     }
 
-    /// Like [`merkleize`](Self::merkleize), but the caller supplies the items instead of
-    /// accumulating them with [`add`](Self::add). The two approaches must not be mixed: do
-    /// not call [`add`](Self::add) before this method.
-    ///
-    /// The items are encoded and hashed into the Merkle structure, and the `Arc` is stored
-    /// directly in the resulting [`MerkleizedBatch`] without copying.
+    /// Compute leaf digests for caller-supplied items without consuming the batch.
     ///
     /// # Panics
     ///
     /// Panics if items were previously added via [`add`](Self::add).
-    pub(crate) fn merkleize_with(
-        self,
-        base: &Mem<F, H::Digest>,
-        items: Arc<Vec<Item>>,
-    ) -> MerkleizedBatchArc<F, H, Item, S> {
+    pub(crate) fn leaf_digests_with(&self, items: &[Item]) -> Vec<H::Digest> {
         assert!(
             self.items.is_empty(),
-            "merkleize_with expects no items added via add"
+            "leaf_digests_with expects no items added via add"
         );
 
         let hasher = &self.hasher;
-        let merkle = self.inner.merkleize_leaves(
-            base,
-            hasher,
-            items.as_slice(),
-            Vec::new,
-            |buf, item, pos| {
-                buf.clear();
-                item.write(buf);
-                hasher.leaf_digest(pos, buf.as_slice())
-            },
+        self.inner.leaf_digests(items, Vec::new, |buf, item, pos| {
+            buf.clear();
+            item.write(buf);
+            hasher.leaf_digest(pos, buf.as_slice())
+        })
+    }
+
+    /// Like [`merkleize`](Self::merkleize), but uses pre-computed leaf digests for `items`.
+    ///
+    /// # Panics
+    ///
+    /// Panics if items were previously added via [`add`](Self::add).
+    pub(crate) fn merkleize_with_leaf_digests(
+        self,
+        base: &Mem<F, H::Digest>,
+        items: Arc<Vec<Item>>,
+        digests: Vec<H::Digest>,
+    ) -> MerkleizedBatchArc<F, H, Item, S> {
+        assert!(
+            self.items.is_empty(),
+            "merkleize_with_leaf_digests expects no items added via add"
         );
+        assert_eq!(
+            digests.len(),
+            items.len(),
+            "pre-computed leaf digest count must match item count"
+        );
+
+        let merkle = self
+            .inner
+            .merkleize_leaf_digests(base, &self.hasher, digests);
         let ancestor_items = Self::collect_ancestor_items(&self.parent);
         Arc::new(MerkleizedBatch {
             inner: merkle,
@@ -900,6 +911,15 @@ mod tests {
             .merkle
             .with_mem(|mem| batch.root(mem, &journal.hasher, 0))
             .unwrap()
+    }
+
+    fn merkleize_with<F: Family + PartialEq>(
+        batch: UnmerkleizedBatch<F, Sha256, TestOp<F>, Sequential>,
+        base: &Mem<F, Digest>,
+        items: Arc<Vec<TestOp<F>>>,
+    ) -> MerkleizedBatchArc<F, Sha256, TestOp<F>, Sequential> {
+        let digests = batch.leaf_digests_with(items.as_slice());
+        batch.merkleize_with_leaf_digests(base, items, digests)
     }
 
     /// Create Merkle configuration for tests.
@@ -2722,7 +2742,7 @@ mod tests {
         let batch = journal.new_batch();
         let actual = journal
             .merkle
-            .with_mem(|mem| batch.merkleize_with(mem, Arc::new(ops)));
+            .with_mem(|mem| merkleize_with(batch, mem, Arc::new(ops)));
 
         assert_eq!(
             batch_root(&journal, &actual),
@@ -2750,7 +2770,7 @@ mod tests {
         let batch = journal.new_batch();
         let merkleized = journal
             .merkle
-            .with_mem(|mem| batch.merkleize_with(mem, Arc::new(ops.clone())));
+            .with_mem(|mem| merkleize_with(batch, mem, Arc::new(ops.clone())));
 
         let expected_root = batch_root(&journal, &merkleized);
         journal.apply_batch(&merkleized).await.unwrap();
@@ -2784,7 +2804,7 @@ mod tests {
         let batch = journal.new_batch();
         let merkleized = journal
             .merkle
-            .with_mem(|mem| batch.merkleize_with(mem, ops_clone));
+            .with_mem(|mem| merkleize_with(batch, mem, ops_clone));
 
         // The batch should hold the same Arc allocation, not a copy.
         assert!(Arc::ptr_eq(&merkleized.items, &ops));
--- a/storage/src/journal/authenticated.rs
+++ b/storage/src/journal/authenticated.rs
@@ -144,7 +144,8 @@ impl<F: Family, H: Hasher, Item: Encode + Send + Sync, S: Strategy>
     ///
     /// # Panics
     ///
-    /// Panics if items were previously added via [`add`](Self::add).
+    /// - If items were previously added via [`add`](Self::add).
+    /// - If `digests` and `items` have different lengths.
     pub(crate) fn merkleize_with_leaf_digests(
         self,
         base: &Mem<F, H::Digest>,
--- a/storage/src/journal/authenticated.rs
+++ b/storage/src/journal/authenticated.rs
@@ -140,27 +140,16 @@ impl<F: Family, H: Hasher, Item: Encode + Send + Sync, S: Strategy>
         )
     }
 
-    /// Like [`merkleize`](Self::merkleize), but uses pre-computed leaf digests for `items`.
+    /// Add caller-supplied items using pre-computed leaf digests.
     ///
     /// # Panics
     ///
     /// - If items were previously added via [`add`](Self::add).
     /// - If `digests` and `items` have different lengths.
-    pub(crate) fn merkleize_with_leaf_digests(
-        self,
-        base: &Mem<F, H::Digest>,
-        items: Arc<Vec<Item>>,
-        digests: Vec<H::Digest>,
-    ) -> MerkleizedBatchArc<F, H, Item, S> {
-        let Self {
-            inner,
-            hasher,
-            items: batch_items,
-            parent,
-        } = self;
+    pub(crate) fn add_leaf_digests(mut self, items: Vec<Item>, digests: Vec<H::Digest>) -> Self {
         assert!(
-            batch_items.is_empty(),
-            "merkleize_with_leaf_digests expects no items added via add"
+            self.items.is_empty(),
+            "add_leaf_digests expects no items added via add"
         );
         assert_eq!(
             digests.len(),
@@ -168,16 +157,9 @@ impl<F: Family, H: Hasher, Item: Encode + Send + Sync, S: Strategy>
             "pre-computed leaf digest count must match item count"
         );
 
-        let inner = inner.add_leaf_digests(digests);
-        let merkle = inner.merkleize(base, &hasher);
-        let ancestor_items = Self::collect_ancestor_items(&parent);
-        Arc::new(MerkleizedBatch {
-            inner: merkle,
-            bagging: hasher.root_bagging(),
-            items,
-            parent: parent.as_ref().map(Arc::downgrade),
-            ancestor_items,
-        })
+        self.inner = self.inner.add_leaf_digests(digests);
+        self.items = items;
+        self
     }
 }
 
@@ -928,10 +910,10 @@ mod tests {
     fn merkleize_with<F: Family + PartialEq>(
         batch: UnmerkleizedBatch<F, Sha256, TestOp<F>, Sequential>,
         base: &Mem<F, Digest>,
-        items: Arc<Vec<TestOp<F>>>,
+        items: Vec<TestOp<F>>,
     ) -> MerkleizedBatchArc<F, Sha256, TestOp<F>, Sequential> {
         let digests = batch.leaf_digests_with(items.as_slice());
-        batch.merkleize_with_leaf_digests(base, items, digests)
+        batch.add_leaf_digests(items, digests).merkleize(base)
     }
 
     /// Create Merkle configuration for tests.
@@ -2754,7 +2736,7 @@ mod tests {
         let batch = journal.new_batch();
         let actual = journal
             .merkle
-            .with_mem(|mem| merkleize_with(batch, mem, Arc::new(ops)));
+            .with_mem(|mem| merkleize_with(batch, mem, ops));
 
         assert_eq!(
             batch_root(&journal, &actual),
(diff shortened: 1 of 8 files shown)


storage/src/merkle/batch.rs
```diff
@@ -225,6 +225,20 @@ impl<F: Family, D: Digest, S: Strategy> UnmerkleizedBatch<F, D, S> {
         self
     }
 
+    /// Add a run of pre-computed leaf digests, in order.
+    ///
+    /// Equivalent to calling [`Self::add_leaf_digest`] for each digest, but reserves capacity for
+    /// the incoming leaves up front to avoid repeated reallocation. Pairs with computing the
+    /// digests in parallel (e.g. via [`commonware_parallel::Strategy::map_init_collect_vec`]).
```
danlaine: Nit: I think the last sentence could be omitted
```suggestion
    /// the incoming leaves up front to avoid repeated reallocation.
```

storage/src/merkle/batch.rs
```diff
@@ -225,6 +225,20 @@ impl<F: Family, D: Digest, S: Strategy> UnmerkleizedBatch<F, D, S> {
...
+    pub fn add_leaf_digests(mut self, digests: impl IntoIterator<Item = D>) -> Self {
+        let digests = digests.into_iter();
+        self.appended.reserve(digests.size_hint().0);
```
danlaine: Should we also reserve space for the parents of the newly appended digests?

danlaine (commented):
A few nits. Also I think we can avoid an `encode()` allocation by re-using the buffer

```
diff --git a/storage/src/journal/authenticated.rs b/storage/src/journal/authenticated.rs
index 0ecd0a43b..2034f70f3 100644
--- a/storage/src/journal/authenticated.rs
+++ b/storage/src/journal/authenticated.rs
@@ -138,11 +138,13 @@ impl<F: Family, H: Hasher, Item: Encode + Send + Sync, S: Strategy>
         let starting_leaves = self.inner.leaves();
         let digests: Vec<H::Digest> = self.inner.strategy().map_init_collect_vec(
             items.iter().enumerate(),
-            || self.hasher.clone(),
-            |h, (i, item)| {
+            || (self.hasher.clone(), Vec::new()),
+            |(h, buf), (i, item)| {
                 let loc = Location::<F>::new(*starting_leaves + i as u64);
                 let pos = Position::try_from(loc).expect("valid leaf location");
-                h.leaf_digest(pos, &item.encode())
+                buf.clear();
+                item.write(buf);
+                h.leaf_digest(pos, buf.as_slice())
             },
         );
         self.inner = self.inner.add_leaf_digests(digests);
diff --git a/storage/src/qmdb/compact/batch.rs b/storage/src/qmdb/compact/batch.rs
index 876b8c3ae..9ad08a59a 100644
--- a/storage/src/qmdb/compact/batch.rs
+++ b/storage/src/qmdb/compact/batch.rs
@@ -26,11 +26,13 @@ where
     let first_leaf = batch.leaves();
     let leaf_digests = merkle.strategy().map_init_collect_vec(
         ops.iter().enumerate(),
-        qmdb::hasher::<H>,
-        |hasher, (i, op)| {
+        || (qmdb::hasher::<H>(), Vec::new()),
+        |(hasher, buf), (i, op)| {
             let offset = u64::try_from(i).expect("operation offset exceeds u64");
             let pos = F::location_to_position(first_leaf + offset);
-            hasher.leaf_digest(pos, &op.encode())
+            buf.clear();
+            op.write(buf);
+            hasher.leaf_digest(pos, buf.as_slice())
         },
     );
```

storage/src/merkle/batch.rs
roberto-bayardo: removed

storage/src/merkle/batch.rs
roberto-bayardo: yes, good catch! done.

storage/src/qmdb/keyless/batch.rs
```diff
@@ -269,16 +269,15 @@ where
 
         let total_size = base + ops.len() as u64;
 
-        // Add operations to the journal batch and merkleize.
-        let mut journal_batch = self.journal_batch;
-        for op in &ops {
-            journal_batch = journal_batch.add(op.clone());
-        }
+        // Hash the operations' leaf digests in parallel, then merkleize the journal batch.
```
danlaine: Should we revert this? Seems like an "AI over-indexing on changes in its context" comment

storage/src/merkle/batch.rs
```diff
@@ -225,6 +225,24 @@ impl<F: Family, D: Digest, S: Strategy> UnmerkleizedBatch<F, D, S> {
...
     }
 
+    /// Add a run of pre-computed leaf digests, in order.
+    ///
+    /// Equivalent to calling [`Self::add_leaf_digest`] for each digest, but reserves capacity up
+    /// front to avoid repeated reallocation.
+    pub fn add_leaf_digests(mut self, digests: impl IntoIterator<Item = D>) -> Self {
+        let digests = digests.into_iter();
+        // Each leaf also appends its parent placeholders, so reserve for the full node count
+        // `position(leaves + n) - position(leaves)`, not just the `n` leaves.
```
danlaine: nit: restates the pretty simple code below
```suggestion
```

roberto-bayardo: removed.

storage/src/qmdb/keyless/batch.rs
roberto-bayardo: Yeah this comment isn't great, but I do think "pat bot" style comments here would be helpful for understanding, so I've replaced this with a few of them.

danlaine (dismissed):

storage/src/qmdb/compact/batch.rs
```diff
@@ -22,9 +22,24 @@ where
...
+            op.write(buf);
+            hasher.leaf_digest(pos, buf.as_slice())
+        },
+    );
+
+    // Add the leaf digests to the batch.
+    batch = batch.add_leaf_digests(leaf_digests);
+
+    // Merkleize the batch.
     merkle.with_mem(|mem| batch.merkleize(mem, &hasher))
```
patrick-ogrady, line 41: I wonder if there is a way to just feed in the raw leaves to this stage to do all hashing in a single run of strategy (rather than forcing 2 sync points)?

roberto-bayardo: since hashing is level-wise anyway, it wouldn't reduce the sync points, but it might be a simpler API. Should I go ahead and do it?

patrick-ogrady: Seems nice not to have to think about so much optimization before handing it off? I think it would be real clean.

What does BMT do (I think we dealt with something similar there).

roberto-bayardo: BMT doesn't seem to do any parallel merkleization?  It just builds a tree all at once from a vector of leaves one node at a time. (unless I'm missing something).

roberto-bayardo: ok pushed a commit that does this. lmk if it's g2g.

patrick-ogrady: Looks like we do indeed do what you originally did (hash then provide): https://github.com/commonwarexyz/monorepo/blob/fcd0df985122c486082d87d480bbff91140e810e/coding/src/reed_solomon.rs#L363-L366

roberto-bayardo: so it turns out that the split is actually important to reduce lock contention, so I've reinstituted it, only appropriately documented.

storage/src/journal/authenticated.rs
```diff
@@ -115,41 +115,55 @@ impl<F: Family, H: Hasher, Item: Encode + Send + Sync, S: Strategy>
...
-        let starting_leaves = self.inner.leaves();
-        let digests: Vec<H::Digest> = self.inner.strategy().map_init_collect_vec(
-            items.iter().enumerate(),
-            || self.hasher.clone(),
-            |h, (i, item)| {
-                let loc = Location::<F>::new(*starting_leaves + i as u64);
-                let pos = Position::try_from(loc).expect("valid leaf location");
-                h.leaf_digest(pos, &item.encode())
-            },
+        assert_eq!(
```
danlaine: Should we document this panic?

storage/src/journal/authenticated.rs
```diff
@@ -115,41 +115,55 @@ impl<F: Family, H: Hasher, Item: Encode + Send + Sync, S: Strategy>
...
-    /// not call [`add`](Self::add) before this method.
+    /// Compute leaf digests for caller-supplied items without consuming the batch.
     ///
-    /// The items are encoded and hashed into the Merkle structure, and the `Arc` is stored
-    /// directly in the resulting [`MerkleizedBatch`] without copying.
+    /// # Panics
+    ///
+    /// Panics if items were previously added via [`add`](Self::add).
+    pub(crate) fn leaf_digests_with(&self, items: &[Item]) -> Vec<H::Digest> {
+        assert!(
```
danlaine, line 124: Should we document this panic?

storage/src/qmdb/immutable/batch.rs
```diff
@@ -260,21 +260,27 @@ where
...
-        // Add operations to the journal batch and merkleize.
-        let mut journal_batch = self.journal_batch;
-        for op in &ops {
-            journal_batch = journal_batch.add(op.clone());
-        }
+        // Merkleize the journal batch.
+        let ops = Arc::new(ops);
+        let leaf_digests = self.journal_batch.leaf_digests_with(ops.as_slice());
+
+        // Hash before borrowing committed Merkle state so the read lock only covers merkleization.
```
danlaine: These comments should be in reversed positions, I think?

storage/src/journal/authenticated.rs
roberto-bayardo: This is the "if items were previously added" case.

storage/src/qmdb/immutable/batch.rs
roberto-bayardo: yeah fixed

storage/src/journal/authenticated.rs

storage/src/merkle/batch.rs
```diff
@@ -225,6 +225,26 @@ impl<F: Family, D: Digest, S: Strategy> UnmerkleizedBatch<F, D, S> {
         self
     }
 
+    /// Append a run of pre-computed leaf digests and merkleize.
+    #[cfg(any(feature = "std", test))]
+    pub(crate) fn merkleize_leaf_digests(
```
patrick-ogrady: is this retained just to use in `merkleize_leaf_digests_matches_individual`?

roberto-bayardo: removed, wasn't carrying its weight.

storage/src/qmdb/keyless/batch.rs
```diff
@@ -258,32 +258,31 @@ where
...
+        let total_size = self.base_size + ops.len() as u64;
 
-        // Add operations to the journal batch and merkleize.
-        let mut journal_batch = self.journal_batch;
-        for op in &ops {
-            journal_batch = journal_batch.add(op.clone());
-        }
+        // Hash before borrowing committed Merkle state so the read lock only covers merkleization.
+        let leaf_digests = self.journal_batch.leaf_digests_with(ops.as_slice());
+        let journal_batch = self.journal_batch.add_leaf_digests(ops, leaf_digests);
```
patrick-ogrady: I think we could avoid calling 2 functions by doing this:
```
diff --git a/storage/src/journal/authenticated.rs b/storage/src/journal/authenticated.rs
index 914b7972b..a69da328d 100644
--- a/storage/src/journal/authenticated.rs
+++ b/storage/src/journal/authenticated.rs
@@ -115,20 +115,20 @@ impl<F: Family, H: Hasher, Item: Encode + Send + Sync, S: Strategy>
         })
     }
 
-    /// Compute leaf digests for caller-supplied items without consuming the batch.
+    /// Add caller-supplied items to the batch.
     ///
     /// # Panics
     ///
     /// Panics if items were previously added via [`add`](Self::add).
-    pub(crate) fn leaf_digests_with(&self, items: &[Item]) -> Vec<H::Digest> {
+    pub(crate) fn add_many(mut self, items: Vec<Item>) -> Self {
         assert!(
             self.items.is_empty(),
-            "leaf_digests_with expects no items added via add"
+            "add_many expects no items added via add"
         );
 
         let first = self.inner.leaves();
         let hasher = &self.hasher;
-        self.inner.strategy().map_init_collect_vec(
+        let digests = self.inner.strategy().map_init_collect_vec(
             items.iter().enumerate(),
             Vec::new,
             |buf, (i, item)| {
@@ -137,24 +137,6 @@ impl<F: Family, H: Hasher, Item: Encode + Send + Sync, S: Strategy>
                 item.write(buf);
                 hasher.leaf_digest(pos, buf.as_slice())
             },
-        )
-    }
-
-    /// Add caller-supplied items using pre-computed leaf digests.
-    ///
-    /// # Panics
-    ///
-    /// - If items were previously added via [`add`](Self::add).
-    /// - If `digests` and `items` have different lengths.
-    pub(crate) fn add_leaf_digests(mut self, items: Vec<Item>, digests: Vec<H::Digest>) -> Self {
-        assert!(
-            self.items.is_empty(),
-            "add_leaf_digests expects no items added via add"
-        );
-        assert_eq!(
-            digests.len(),
-            items.len(),
-            "pre-computed leaf digest count must match item count"
         );
 
         self.inner = self.inner.add_leaf_digests(digests);
@@ -912,8 +894,7 @@ mod tests {
         base: &Mem<F, Digest>,
         items: Vec<TestOp<F>>,
     ) -> MerkleizedBatchArc<F, Sha256, TestOp<F>, Sequential> {
-        let digests = batch.leaf_digests_with(items.as_slice());
-        batch.add_leaf_digests(items, digests).merkleize(base)
+        batch.add_many(items).merkleize(base)
     }
 
     /// Create Merkle configuration for tests.
diff --git a/storage/src/merkle/batch.rs b/storage/src/merkle/batch.rs
index d0aa42e91..0b72695e1 100644
--- a/storage/src/merkle/batch.rs
+++ b/storage/src/merkle/batch.rs
@@ -226,7 +226,7 @@ impl<F: Family, D: Digest, S: Strategy> UnmerkleizedBatch<F, D, S> {
     }
 
     /// Add a run of pre-computed leaf digests, in order.
-    #[cfg(any(feature = "std", test))]
+    #[cfg(feature = "std")]
     pub(crate) fn add_leaf_digests(mut self, digests: impl IntoIterator<Item = D>) -> Self {
         let digests = digests.into_iter();
         // Each leaf also appends its parent placeholders, so reserve for the full node count.
diff --git a/storage/src/qmdb/any/batch.rs b/storage/src/qmdb/any/batch.rs
index e4a6a3fb9..49d5908c1 100644
--- a/storage/src/qmdb/any/batch.rs
+++ b/storage/src/qmdb/any/batch.rs
@@ -821,9 +821,8 @@ where
         let leaves = Location::new(self.base_size + ops.len() as u64);
         let inactive_peaks = db.inactive_peaks(leaves, floor);
 
-        // Hash before borrowing committed Merkle state so the read lock only covers merkleization.
-        let leaf_digests = self.journal_batch.leaf_digests_with(ops.as_slice());
-        let journal_batch = self.journal_batch.add_leaf_digests(ops, leaf_digests);
+        // Hash before `with_mem` borrows committed Merkle state under its read lock.
+        let journal_batch = self.journal_batch.add_many(ops);
         let journal = db.log.with_mem(|base| journal_batch.merkleize(base));
         let root = db
             .log
diff --git a/storage/src/qmdb/compact/batch.rs b/storage/src/qmdb/compact/batch.rs
index 0400c7b54..399804a45 100644
--- a/storage/src/qmdb/compact/batch.rs
+++ b/storage/src/qmdb/compact/batch.rs
@@ -25,7 +25,7 @@ where
     let hasher = qmdb::hasher::<H>();
     let first_leaf = batch.leaves();
 
-    // Hash before borrowing committed Merkle state so the read lock only covers merkleization.
+    // Hash before `with_mem` borrows committed Merkle state under its read lock.
     let leaf_digests =
         merkle
             .strategy()
diff --git a/storage/src/qmdb/immutable/batch.rs b/storage/src/qmdb/immutable/batch.rs
index 362b74cca..07366ae79 100644
--- a/storage/src/qmdb/immutable/batch.rs
+++ b/storage/src/qmdb/immutable/batch.rs
@@ -260,11 +260,8 @@ where
 
         let total_size = base + ops.len() as u64;
 
-        // Hash before borrowing committed Merkle state so the read lock only covers merkleization.
-        let leaf_digests = self.journal_batch.leaf_digests_with(ops.as_slice());
-
-        // Merkleize the journal batch.
-        let journal_batch = self.journal_batch.add_leaf_digests(ops, leaf_digests);
+        // Hash before `with_mem` borrows committed Merkle state under its read lock.
+        let journal_batch = self.journal_batch.add_many(ops);
         let journal = db.journal.with_mem(|mem| journal_batch.merkleize(mem));
 
         // Compute the root.
diff --git a/storage/src/qmdb/keyless/batch.rs b/storage/src/qmdb/keyless/batch.rs
index 572d8c011..146f06c35 100644
--- a/storage/src/qmdb/keyless/batch.rs
+++ b/storage/src/qmdb/keyless/batch.rs
@@ -267,9 +267,8 @@ where
 
         let total_size = self.base_size + ops.len() as u64;
 
-        // Hash before borrowing committed Merkle state so the read lock only covers merkleization.
-        let leaf_digests = self.journal_batch.leaf_digests_with(ops.as_slice());
-        let journal_batch = self.journal_batch.add_leaf_digests(ops, leaf_digests);
+        // Hash before `with_mem` borrows committed Merkle state under its read lock.
+        let journal_batch = self.journal_batch.add_many(ops);
         let journal = db.journal.with_mem(|mem| journal_batch.merkleize(mem));
 
         // Compute the root.
```

patrick-ogrady (approved):

PR #3991, opened 2026-06-08
[storage] remove Persistable trait

roberto-bayardo:
`Persistable` isn't really adding value, and this gives us more flexibility for individual storage classes to implement appropriate sync/commit/etc functionality with fewer constraints.

## Summary

- Removed crate-level `storage::Persistable`.
- Moved journal persistence methods (`commit`, `sync`, `destroy`) onto `journal::contiguous::Mutable`.
- Simplified journal/QMDB/glue bounds from `Mutable + Persistable<journal::Error>` to `Mutable`.
- Preserved non-journal persistence APIs as inherent methods and added explicit `DbAny` persistence methods for generic tests/benches.

### DIFF
--- a/storage/src/journal/contiguous/tests.rs
+++ b/storage/src/journal/contiguous/tests.rs
@@ -1,10 +1,7 @@
 //! Generic test suite for [Contiguous] trait implementations.
 
 use super::{Contiguous, Many, Reader as _};
-use crate::{
-    journal::{contiguous::Mutable, Error},
-    Persistable,
-};
+use crate::journal::{contiguous::Mutable, Error};
 use commonware_utils::NZUsize;
 use futures::{future::BoxFuture, StreamExt};
 use std::sync::atomic::{AtomicUsize, Ordering};
@@ -180,12 +177,9 @@ pub(super) mod partition_sync_fault {
     }
 }
 
-pub(super) trait PersistableContiguous:
-    Mutable<Item = u64> + Persistable<Error = Error>
-{
-}
+pub(super) trait MutableContiguous: Mutable<Item = u64> {}
 
-impl<T> PersistableContiguous for T where T: Mutable<Item = u64> + Persistable<Error = Error> {}
+impl<T> MutableContiguous for T where T: Mutable<Item = u64> {}
 
 async fn get_bounds<J: Contiguous>(journal: &J) -> std::ops::Range<u64> {
     let reader = journal.reader().await;
@@ -211,7 +205,7 @@ async fn read_item<J: Contiguous>(journal: &J, position: u64) -> Result<J::Item,
 pub(super) async fn run_contiguous_tests<F, J>(factory: F)
 where
     F: Fn(String, usize) -> BoxFuture<'static, Result<J, Error>>,
-    J: PersistableContiguous,
+    J: MutableContiguous,
 {
     let counter = AtomicUsize::new(0);
     let indexed_factory = |name: String| {
@@ -263,7 +257,7 @@ where
 async fn test_empty_journal_bounds<F, J>(factory: &F)
 where
     F: Fn(String) -> BoxFuture<'static, Result<J, Error>>,
-    J: PersistableContiguous,
+    J: MutableContiguous,
 {
     let journal = factory("empty".into()).await.unwrap();
     let bounds = get_bounds(&journal).await;
@@ -277,7 +271,7 @@ where
 async fn test_bounds_with_items<F, J>(factory: &F)
 where
     F: Fn(String) -> BoxFuture<'static, Result<J, Error>>,
-    J: PersistableContiguous,
+    J: MutableContiguous,
 {
     let mut journal = factory("bounds-with-items".into()).await.unwrap();
 
@@ -299,7 +293,7 @@ where
 async fn test_bounds_after_prune<F, J>(factory: &F)
 where
     F: Fn(String) -> BoxFuture<'static, Result<J, Error>>,
-    J: PersistableContiguous,
+    J: MutableContiguous,
 {
     let mut journal = factory("bounds-after-prune".into()).await.unwrap();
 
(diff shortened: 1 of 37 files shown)


storage/src/ordinal/storage.rs
```diff
@@ -419,6 +417,13 @@ impl<E: BufferPooler + Context, V: CodecFixed<Cfg = ()>> Ordinal<E, V> {
         Ok(())
     }
 
+    /// Durably persist the ordinal, guaranteeing the current state will survive a crash.
+    ///
+    /// This is an alias for [Self::sync].
+    pub async fn commit(&self) -> Result<(), Error> {
```
danlaine: Does this method buy us anything?

storage/src/journal/contiguous/tests.rs
```diff
@@ -180,12 +177,9 @@ pub(super) mod partition_sync_fault {
     }
 }
 
-pub(super) trait PersistableContiguous:
-    Mutable<Item = u64> + Persistable<Error = Error>
-{
-}
+pub(super) trait MutableContiguous: Mutable<Item = u64> {}
```
danlaine: I think we could remove this and just use `Mutable<Item = u64>` bound below

danlaine (commented):
This change makes it so that all `Mutable` contiguous journals (i.e. those that can be appended, pruned, or rewound) must also be persistable. This feels a bit unsatisfying since you might want, for example, an in-memory journal which isn't persistable.

Instead of requiring that all `Mutable` are `Persistable` by moving the latter's functions into the former, should we instead require that all `Persistable` are `Mutable` instead? i.e.:

```
  pub trait Persistable: Mutable {
      fn commit(&self) -> impl Future<Output = Result<(), Error>> + Send { self.sync() }
      fn sync(&self)   -> impl Future<Output = Result<(), Error>> + Send;
      fn destroy(self) -> impl Future<Output = Result<(), Error>> + Send where Self: Sized;
  }
```

And then change existing bounds from `Mutable` + `Persistable` to just `Persistable`.

While we're at it, wdyt of calling it `Persistent` instead of `Persistable`? I think that might sound a bit more natural

roberto-bayardo:
> This change makes it so that all `Mutable` contiguous journals (i.e. those that can be appended, pruned, or rewound) must also be persistable. This feels a bit unsatisfying since you might want, for example, an in-memory journal which isn't persistable.
> 
> Instead of requiring that all `Mutable` are `Persistable` by moving the latter's functions into the former, should we instead require that all `Persistable` are `Mutable` instead? i.e.:
> 
> ```
>   pub trait Persistable: Mutable {
>       fn commit(&self) -> impl Future<Output = Result<(), Error>> + Send { self.sync() }
>       fn sync(&self)   -> impl Future<Output = Result<(), Error>> + Send;
>       fn destroy(self) -> impl Future<Output = Result<(), Error>> + Send where Self: Sized;
>   }
> ```
> 
> And then change existing bounds from `Mutable` + `Persistable` to just `Persistable`.
> 
> While we're at it, wdyt of calling it `Persistent` instead of `Persistable`? I think that might sound a bit more natural

I had actually started down this path, though calling it `Durable` instead. Then I thought maybe it's not worth it until we have a concrete need.  Happy to do it, just was eager to reduce trait proliferation a bit.

danlaine:
> I had actually started down this path, though calling it Durable instead. Then I thought maybe it's not worth it until we have a concrete need. Happy to do it, just was eager to reduce trait proliferation a bit.

I'd lean toward `Durable: Mutable` to allow for non-durable implementations. But as you say, we don't currently have any implementations that fit this bill, so either way is OK by me.

roberto-bayardo:
> > I had actually started down this path, though calling it Durable instead. Then I thought maybe it's not worth it until we have a concrete need. Happy to do it, just was eager to reduce trait proliferation a bit.
> 
> I'd lean toward `Durable: Mutable` to allow for non-durable implementations. But as you say, we don't currently have any implementations that fit this bill, so either way is OK by me.

I prefer leaving out for now and avoid the ugly + Durable everywhere.

danlaine:
> > > I had actually started down this path, though calling it Durable instead. Then I thought maybe it's not worth it until we have a concrete need. Happy to do it, just was eager to reduce trait proliferation a bit.
> > 
> > 
> > I'd lean toward `Durable: Mutable` to allow for non-durable implementations. But as you say, we don't currently have any implementations that fit this bill, so either way is OK by me.
> 
> I prefer leaving out for now and avoid the ugly + Durable everywhere.

If `Durable: Mutable` then we would only need the `Durable` bound at existing call sites, not `Durable + Mutable` since the former implies the latter.

roberto-bayardo:
> If `Durable: Mutable` then we would only need the `Durable` bound at existing call sites, not `Durable + Mutable` since the former implies the latter.

Ah right, I was still thinking of it as a standalone not Durable: Mutable.

storage/src/journal/contiguous/tests.rs
roberto-bayardo: done

storage/src/ordinal/storage.rs
roberto-bayardo: nope removed

storage/src/qmdb/mod.rs
```diff
@@ -32,7 +32,7 @@
 //! # Traits
 //!
 //! Keyed mutable variants ([any] and [current]) implement `any::traits::DbAny` and
```
danlaine: I think we could remove "and persistence methods exposed by each database type." Seems a bit awkward to say they implement the persistence methods they expose.

danlaine (dismissed):
1 nit otherwise lgtm

storage/src/qmdb/mod.rs
roberto-bayardo: done

danlaine (approved):

PR #3993, opened 2026-06-09
[qmdb] Optimize Batch Processing

patrick-ogrady:
Optimizes the QMDB `any` merkleize/finalize pipeline for the block-processing workload (32k updated keys per batch over 1M total keys). Roots are byte-identical to main in every configuration: this changes how results are computed, never what is committed.

## Changes

**Fused state load + merkleize (on by default)**
- `UnmerkleizedBatch::get`/`get_many` now retain committed-DB resolved locations on the batch (interior-mutable, consumed by `merkleize`), so any caller that reads then writes on one batch gets a merkleize that skips re-resolving and re-reading those ~32k keys. No new API: `current::Db` batches and glue inherit the behavior through the shared inner batch, and applications need no changes.
- Only committed-snapshot resolutions are retained. Retaining ancestor-diff resolutions is unsound: the recorded `base_old_loc` goes stale when a committed ancestor is dropped before merkleize, corrupting the snapshot on apply (caught by adversarial review, reproduced and pinned by `test_unordered_fixed_retention_survives_ancestor_commit`). Ancestor-resolved keys re-resolve in memory at merkleize, so nothing measurable is lost.
- Root parity for the retention path is covered at depths 0-2 for `any` and 0-1 for `current`, including reads-after-writes, accumulated multi-read and single-`get` retention.

**Floor raise**
- One bitmap read guard per candidate batch (`Shared::fill_candidates`) instead of one lock acquisition per candidate (~33k/batch).
- Snapshot probe skipped for committed-prefix candidates with no diff/ancestor override (the committed bitmap is exact: `apply_diff` sets the new location and clears `base_old_loc` on apply).
- Candidate ops bulk-read and decoded in parallel (`read_many_decoded`), batch size 64 -> 4096; scratch buffers replace per-op `vec![0u8; N]` (~33k allocs/iter removed).
- `floor_diff` merged into the sorted diff with a two-cursor merge instead of a full re-sort.

**Op generation / batch bookkeeping**
- The serial cost here was sorting 32-byte digest keys, not op materialization: location ordering now sorts compact `(u64, u32)` pairs instead of 80B entries, the diff is built pre-sorted via a two-cursor merge, and `commonware-parallel` gains `par_sort_unstable_by`.
- Batch `mutations` is an insertion vec with sort+dedup-last-wins on merkleize (was a `BTreeMap` with per-write ordered inserts), consumed downstream as a tombstoned sorted vec.
- Digest-keyed hot maps use `ahash` (hashing a hash with SipHash was ~2ms/batch).
- Ancestor-diff resolution for pending chains uses forward cursors over sorted queries instead of a binary search per key per ancestor (depth>=1 only).

**Merkle / journal / runtime**
- Leaf encode+hash and interior-node hashing fused into one divide-and-conquer dispatch (was one leaf pass + one rayon barrier per tree height, ~21 barriers).
- `read_many` decodes in parallel from a single pooled fill; scratch buffers come from the page-cache pool uninitialized (no multi-MB zero-fill per call).
- `read_cached_many` resolves `(page, slot)` once per page run instead of per item (~2.15 items/page on this workload).
- New `commonware-storage/sha2-asm` feature forwards to `commonware-cryptography/sha2-asm` (off by default; sha2-asm does not build on Windows MSVC). On stable aarch64 the default sha2 backend is soft (~272ns/hash vs ~41ns with asm, 6.6x); x86_64 already uses SHA-NI at runtime. Constantinople already enables it.

## Benchmarks

`merkbench` harness (committed on this branch): per-iteration `get_many(_with_locations)` + 32,768 writes + `merkleize` + `root` against `any::unordered::fixed::Db<mmr>` seeded with 1M keys, EightCap, Sha256, tokio, 512MB page cache. p50 of 10-12 iters, Apple Silicon, both sides built with `sha2-asm` (matching production). depth = pending ancestor batches below the timed batch.

Rayon=8:

| depth | main | PR (plain API) | PR (fused API) |
|---|---|---|---|
| 0 | 31.5ms | 19.5ms (1.62x) | **13.0ms (2.43x)** |
| 1 | 39.9ms | 25.3ms (1.57x) | **15.0ms (2.65x)** |

Rayon=2 (validator default):

| depth | main | PR (fused API) |
|---|---|---|
| 0 | 32.1ms | **15.3ms (2.09x)** |
| 1 | 40.4ms | **17.3ms (2.33x)** |

Per-iteration roots are byte-identical to main across all configs, modes, and depths (the harness prints them; every optimization was gated on exact parity).

## Verification

- Full `commonware-storage` suite (2558), runtime `buffer::` (137), `commonware-parallel` (6) pass; conformance (87) passes (no format changes); stability check passes; clippy/fmt clean.
- `test_unordered_fixed_resolved_merkleize_parity` covers fused-vs-plain root equality at depth 0 and 1 across updates/deletes/creates.
- Tried and rejected with measurements: routing floor reads through `read_many` without parallel decode (regressed: the reader guard is already held once for the whole floor raise), sorting a materialized interior-node list (regressed vs recording in append order).

### DIFF
--- a/storage/src/qmdb/any/batch.rs
+++ b/storage/src/qmdb/any/batch.rs
@@ -110,9 +110,9 @@ struct ExistingEntry<K, F: Family, V> {
     value: Option<V>,
 }
 
-/// A key's resolved location, as produced by `get_many_with_locations` and consumed by
-/// `merkleize_resolved`. Carries enough provenance to reproduce the op-gen ordering and
-/// `base_old_loc` without re-reading the operation.
+/// A key's resolved location, as produced by `get_many_with_locations` and attached via
+/// `with_resolved`. Carries enough provenance to reproduce the op-gen ordering and `base_old_loc`
+/// without re-reading the operation during `merkleize`.
 #[derive(Clone, Copy, Debug)]
 pub struct ResolvedLocation<F: Family> {
     /// The location used for op-gen ordering. For a key resolved through an ancestor diff this is
@@ -879,7 +879,8 @@ where
     }
 }
 
-// Unordered op-generation tail shared by `merkleize_with_floor_scan` and `merkleize_resolved`.
+// Shared unordered op-generation tail: turns classified existing entries plus remaining
+// create mutations into operations, then runs the floor-raise and journal merkleize.
 impl<F: Family, K, V, H, S: Strategy> Merkleizer<F, H, update::Unordered<K, V>, S>
 where
     K: Key,
--- a/storage/src/qmdb/any/batch.rs
+++ b/storage/src/qmdb/any/batch.rs
@@ -29,7 +29,7 @@ use commonware_parallel::Strategy;
 use commonware_utils::bitmap;
 use core::{cmp::Ordering, ops::Range};
 use std::{
-    collections::BTreeMap,
+    collections::{BTreeMap, HashMap},
     iter,
     sync::{Arc, Weak},
 };
@@ -211,7 +211,7 @@ where
     /// [`get_many_with_locations`](UnmerkleizedBatch::get_many_with_locations) during state load.
     /// When a mutation key is present here, `merkleize` reuses the location and skips the
     /// redundant journal read. Empty for callers that did not pre-resolve.
-    resolved: BTreeMap<U::Key, Option<ResolvedLocation<F>>>,
+    resolved: HashMap<U::Key, Option<ResolvedLocation<F>>>,
 }
 
 /// A speculative batch of operations whose root digest has been computed,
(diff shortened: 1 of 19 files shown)


storage/src/qmdb/any/batch.rs
```diff
@@ -756,20 +736,31 @@ where
                     let Some(key) = op.key().cloned() else {
                         continue; // skip CommitFloor and other non-keyed ops
                     };
-                    // `is_active_at` is required even for set bits in the committed bitmap
-                    // range: this batch's own diff or an uncommitted ancestor diff may
-                    // supersede the committed location, and neither source is reflected in
-                    // the bitmap.
-                    if !self.is_active_at(&key, candidate, &diff, db) {
+                    // A single batch-diff lookup drives the activity check, the base
```
patrick-ogrady, line 953: Refine this comment

storage/src/qmdb/any/batch.rs
```diff
@@ -756,20 +736,31 @@ where
...
+                                )
+                            },
+                            |entry| (entry.loc() == Some(candidate), entry.base_old_loc()),
+                        ),
+                    };
+                    if !active {
                         continue;
                     }
                     let new_loc = Location::new(self.base_size + ops.len() as u64);
-                    let search = diff.binary_search_by(|(k, _)| k.cmp(&key));
```
patrick-ogrady, line 767: Avoid duplicate binary search...

PR #3999, opened 2026-06-09
[ci] flatten caches

andresilva:


.github/workflows/coverage.yml
```diff
@@ -37,7 +37,8 @@ jobs:
       uses: ./.github/actions/setup
       with:
         additional-toolchain: ${{ env.NIGHTLY_VERSION }}
-        additional-cache-key: ${{ matrix.flags || 'default' }}-part-${{ matrix.partition }}
+        additional-cache-key: ${{ matrix.flags || 'default' }}
+        rust-cache-save: ${{ matrix.partition == 1 }}
```
patrick-ogrady, line 42: This assumes that crates are largely shared by partitions? when I've poked around, I find they differ a decent bit? Maybe that is a fluke?

andresilva: rust-cache is only caching dependencies, it doesn't cache anything from the workspace

clabby (dismissed):

clabby (approved):

clabby (approved):
LGTM. My bad for not adding to mq.

PR #4000, opened 2026-06-09
[storage] [glue] Use contiguous journal in compact QMDB to support rewind

danlaine:
## Why

Compact QMDBs could only rewind a single commit: their persistence kept exactly one prior state, so anything older was gone. Callers that need to roll back further (for example, crash recovery that rewinds a database to an earlier committed state) had to treat compact databases as a special case with a rewind depth of one. This PR makes compact QMDBs rewind arbitrarily deep, like the other variants. Rewind depth is now bounded only by pruning.

## Before

A compact db persisted its state in two places that had to stay in lockstep:

- The compact Merkle wrote its frontier (leaf count + pinned peaks) into one of two metadata slots, flipping a generation pointer on every commit. Rewind flipped the pointer back, which is why only one step was possible: there was only one old slot.
- The witness (the encoded last-commit operation plus its inclusion proof, needed to serve compact sync) was stored in those same slots.

Reopen had to reconcile the two stores against each other to handle crashes between their writes.

## After

There is now a single persistent store: a contiguous journal owned by the db. Every durable sync appends one entry containing the commit operation, its inclusion proof, and the pinned frontier nodes. Each entry is a complete snapshot of one synced commit, and the journal keeps all of them (until pruned), so the db can rewind to any retained sync point.

The compact Merkle keeps no state on disk anymore. It is rebuilt in memory from the journal's tip entry on reopen, and from the target entry on rewind. Its two-slot persistence, metadata partition, and `E: Context` type parameter are deleted.

Properties:

- **One commit point.** The journal syncs after an append. Crash before it and the unsynced entry is dropped on reopen, recovering the previous commit. There is no second store to reconcile.
- **Rewind to any retained commit.** Entries are strictly ordered by committed leaf count, so `rewind(target)` binary-searches for the matching entry, truncates the journal to it, syncs, and rebuilds the Merkle from it. A target below the pruning boundary (or one that was never synced) returns `RewindBeyondHistory`.
- **Verification on load.** Reopen and rewind recompute the root from the rebuilt frontier and check the stored proof against it. Corruption fails with `DataCorrupted` instead of serving bad state.
- **Explicit pruning.** The journal grows by one entry per sync until `prune(pruning_boundary)` drops entries below the boundary. The current commit's entry always survives.

## API changes

- `CompactDb::rewind(&mut self)` (always one step) is now `rewind(&mut self, target)`.
- `CompactDb::prune(&self, pruning_boundary)` is new.
- Commits whose encoded operation exceeds the witness entry decode bound are rejected with the new `Error::CommitTooLarge`, both at commit time and on compact-sync import, instead of persisting an entry that the next reopen cannot decode.
- `CompactConfig` is now `{ strategy, witness, commit_codec_config }`; the merkle config (and its partition) is gone.
- `merkle::compact::Merkle<F, D, S>` lost its `E` parameter and its async `init(context, config)`; it is now constructed synchronously with `new(strategy)`.
- `CompactDb::commit()` (an alias of `sync()`) is removed; `current_target()` is renamed to `target()`.
- Compact-sync import no longer touches the target partition until its first persist: an abandoned or cancelled sync leaves the previous state intact, and rewind/prune on the imported handle are rejected until that persist runs.
- The rewind-depth reporting machinery in glue (`ManagedDb::max_rewind_depth`, `DatabaseSet::max_rewind_depth`, `assert_rewind_window_safety`) is removed: every database now rewinds arbitrarily deep, so there is no depth to report or assert on.
- glue's `Config.max_pending_acks` moved into `PruneConfig`. Glue only used it to seed the prune retention windows (marshal keeps its own copy for dispatch tracking), so the value is now carried only when pruning is enabled.

### DIFF
--- a/glue/src/stateful/actor/core/mod.rs
+++ b/glue/src/stateful/actor/core/mod.rs
@@ -10,10 +10,7 @@ use crate::stateful::{
         processor::{Processor, ProcessorMetrics},
         syncer::{self, SyncPlan, SyncResult},
     },
-    db::{
-        assert_rewind_window_safety, AttachableResolverSet, DatabaseSet, StateSyncSet,
-        SyncEngineConfig,
-    },
+    db::{AttachableResolverSet, DatabaseSet, StateSyncSet, SyncEngineConfig},
     Application,
 };
 use commonware_actor::mailbox::{self as actor_mailbox};
@@ -42,6 +39,11 @@ type BlockDigest<A, E> = <<A as Application<E>>::Block as Digestible>::Digest;
 /// Periodic pruning configuration.
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
 pub struct PruneConfig {
+    /// Marshal's ack window. Must match the marshal config used to construct the marshal
+    /// mailbox: pruning retains at least the last `max_pending_acks + 1` finalized blocks so
+    /// that any state a restart may need to rewind to is never pruned.
+    pub max_pending_acks: NonZeroUsize,
+
     /// Prune databases and marshal every `maintenance_interval` finalized blocks.
     ///
     /// This controls only how often pruning runs, not how much history is retained. Each prune
@@ -94,11 +96,6 @@ where
 
     /// Marshal mailbox used for startup anchoring and lazy recovery.
     pub marshal: MarshalMailbox<S, V>,
-
-    /// Marshal ack window used by the provided marshal mailbox.
-    ///
-    /// This must match the marshal config used to construct [`Self::marshal`].
-    pub max_pending_acks: NonZeroUsize,
 
     /// Capacity of the stateful actor mailbox channel.
     pub mailbox_size: NonZeroUsize,
@@ -159,9 +156,6 @@ where
     /// Sync engine tuning knobs.
     sync_config: SyncEngineConfig,
 
-    /// Marshal ack window, used to derive automatic prune retention.
-    max_pending_acks: NonZeroUsize,
-
     /// Periodic prune configuration.
     prune_config: Option<PruneConfig>,
 }
@@ -181,7 +175,6 @@ where
     /// This only wires dependencies and allocates the mailbox. The actor does
     /// not process messages until [`Stateful::start`] is called.
     pub fn init(context: E, config: Config<E, A, S, V, R>) -> (Self, Mailbox<E, A>) {
-        assert_rewind_window_safety::<E, A::Databases>(config.max_pending_acks);
         if let Some(prune_config) = config.prune_config {
             prune_config.assert_valid();
         }
@@ -198,7 +191,6 @@ where
                 plan: config.plan,
                 resolvers: config.resolvers,
                 sync_config: config.sync_config,
-                max_pending_acks: config.max_pending_acks,
                 prune_config: config.prune_config,
             },
             Mailbox::new(sender),
@@ -247,7 +239,6 @@ where
             artifact: None,
             resolvers: self.resolvers,
             sync_completed,
-            max_pending_acks: self.max_pending_acks,
             prune_config: self.prune_config,
         };
         let _ = join!(syncer.start(), syncing.start());
@@ -276,7 +267,6 @@ where
             databases,
             anchor,
             processor_metrics,
-            self.max_pending_acks,
             self.prune_config,
         );
         Processing {
@@ -443,7 +433,6 @@ mod tests {
                     db_config: (),
                     input_provider: (),
                     marshal,
-                    max_pending_acks: NZUsize!(1),
                     mailbox_size: NZUsize!(8),
                     plan: plan.with_floor(finalization),
                     resolvers: NoopResolver,
--- a/glue/src/stateful/db/immutable/compact.rs
+++ b/glue/src/stateful/db/immutable/compact.rs
@@ -254,24 +254,24 @@ where
         self.sync().await
     }
 
+    async fn prune(&mut self, target: &Self::SyncTarget) -> Result<(), Error<F>> {
+        Self::prune(self, target.leaf_count).await
+    }
+
     async fn sync_target(&self) -> Self::SyncTarget {
-        self.current_target()
+        self.target()
     }
 
     async fn rewind_to_target(&mut self, target: Self::SyncTarget) -> Result<(), Error<F>> {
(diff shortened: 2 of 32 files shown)


glue/src/stateful/actor/core/mod.rs
```diff
@@ -42,6 +39,11 @@ type BlockDigest<A, E> = <<A as Application<E>>::Block as Digestible>::Digest;
 /// Periodic pruning configuration.
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
 pub struct PruneConfig {
+    /// Marshal's ack window. Must match the marshal config used to construct the marshal
+    /// mailbox: pruning retains at least the last `max_pending_acks + 1` finalized blocks so
+    /// that any state a restart may need to rewind to is never pruned.
+    pub max_pending_acks: NonZeroUsize,
```
patrick-ogrady, line 45: Do we still need this? With rewind, I'm not sure we do?

storage/src/qmdb/compact/witness.rs
```diff
@@ -1,103 +1,144 @@
...
-    if slot == 0 {
-        SLOT_A_LAST_COMMIT_OP_PREFIX
-    } else {
-        SLOT_B_LAST_COMMIT_OP_PREFIX
+use commonware_utils::sync::{AsyncMutex, RwLock};
+use std::sync::atomic::{AtomicBool, Ordering};
+
+/// Upper bound on the encoded last-commit operation, enforced before any witness is built and
+/// as a decode guard against corrupted length prefixes.
+pub(crate) const MAX_OP_BYTES: usize = 1 << 24;
```
patrick-ogrady: I don't think this is necessary?

storage/src/qmdb/compact/witness.rs
```diff
@@ -106,60 +147,274 @@ impl<F: Family, D: Digest> ServeState<F, D> {
...
+where
+    E: Context,
+    F: Family,
+    D: Digest,
+{
+    let cfg = variable::Config {
+        partition: cfg.partition,
+        items_per_section: cfg.items_per_section,
+        compression: cfg.compression,
+        // The caller's config carries `()` as a placeholder codec config; the real entry codec
```
patrick-ogrady: this seems like a jank workaround?

storage/src/qmdb/keyless/variable.rs
```diff
@@ -379,6 +390,25 @@ mod test {
         });
     }
 
+    /// A commit whose encoded operation exceeds the witness entry decode bound must be rejected
+    /// at commit time, not discovered as corruption on the next reopen.
```
patrick-ogrady: seems like an unnecessary/unusal test?

storage/src/qmdb/mod.rs
```diff
@@ -177,6 +177,9 @@ pub enum Error<F: Family> {
     #[error("location out of bounds: {0} >= {1}")]
     LocationOutOfBounds(Location<F>, Location<F>),
 
+    #[error("commit operation too large: {0} bytes > {1}")]
+    CommitTooLarge(usize, usize),
```
patrick-ogrady: again, not needed?

patrick-ogrady (commented):
initial nits

glue/src/stateful/actor/core/mod.rs
patrick-ogrady: I think you do...to ensure we don't prune below what isn't ack'd yet?

storage/src/qmdb/compact/witness.rs
danlaine: agreed, removed

storage/src/qmdb/compact/witness.rs
danlaine: Claude is trying to convince me we should have this to prevent OOM in the event of disk corruption, and to limit the metadata size to a "reasonable" bound. Doesn't seem super compelling. I'll remove.

storage/src/qmdb/compact/witness.rs
```diff
@@ -1,173 +1,375 @@
...
+            return Ok(());
+        }
+        // Clamp below the tip so the journal never empties: the tip is the current state.
+        let pos = self
+            .first_at_or_above(pruning_boundary)
+            .await?
+            .min(bounds.end - 1);
+        self.journal.prune(pos).await?;
+        self.journal.sync().await?;
+        Ok(())
```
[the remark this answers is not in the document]
danlaine: This is the same behavior as other places in the storage crate. Documented to be more clear.

storage/src/qmdb/compact/witness.rs
```diff
@@ -1,173 +1,379 @@
...
+    import_pending: AtomicBool,
+
+    /// Serializes persist/rewind/prune.
+    sync_lock: AsyncMutex<()>,
 }
 
-impl<F: Family, D: Digest> Cache<F, D> {
-    /// Create a cache from the witness loaded or bootstrapped during db initialization.
-    pub(crate) const fn new(witness: ServeState<F, D>) -> Self {
+impl<E: Context, F: Family, D: Digest> Store<E, F, D> {
```
patrick-ogrady, line 131: I think some of the helper functions on this struct are a bit questionable (too minimal to be worth it?)

patrick-ogrady (commented):
nits

storage/src/qmdb/compact/witness.rs
patrick-ogrady: Played with this a bit and found this cleaner actually.

glue/src/stateful/db/immutable/compact.rs
```diff
@@ -254,25 +254,24 @@ where
         self.sync().await
     }
 
+    async fn prune(&mut self, target: &Self::SyncTarget) -> Result<(), Error<F>> {
+        Self::prune(self, target.leaf_count).await
```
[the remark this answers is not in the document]
danlaine: The compact db keeps one witness per commit, and rewinding to a block only needs that block's own witness, nothing older. So we prune everything below the target block's leaf count, which keeps the target's witness and everything newer. The full dbs prune by range.start() instead because they store operations, and representing a block requires all ops from its floor onward. Different storage, different boundary.

patrick-ogrady:
bugbot run

PR #4001, opened 2026-06-09
[ci] use sccache

andresilva:
Depends on #3999.

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -123,6 +123,19 @@ runs:
         cache-workspace-crates: false
         workspaces: . -> ${{ steps.rust-cache-workspace.outputs.target }}
         cache-bin: true
+
+    # rust-cache and sccache cover different parts of the build. rust-cache
+    # restores Cargo state: downloaded crates, git dependencies, installed
+    # binaries, and dependency artifacts in target. It is keyed by the job,
+    # toolchain, Cargo files, and the matrix-specific cache key above, so it is
+    # a good coarse restore for a job's expected build graph.
+    #
+    # sccache sits underneath Cargo as RUSTC_WRAPPER and caches individual
+    # compiler outputs. That makes it useful when the target cache is missing,
+    # stale, cleaned, or split across many CI partitions. It can also hit for
+    # unchanged crates when rust-cache misses its coarser key after Cargo.lock
+    # or workflow matrix changes. Only main writes to the compiler cache; pull
+    # requests read from it without replacing shared entries.
     - name: Setup sccache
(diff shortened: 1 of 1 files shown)


.github/actions/setup/action.yml
```diff
@@ -123,3 +123,29 @@ runs:
...
+
+        [cache.gha]
+        enabled = true
+        version = ""
+        rw_mode = "$sccache_rw_mode"
+        EOF
+
+        echo "CARGO_INCREMENTAL=0" >> "$GITHUB_ENV"
+        echo "RUSTC_WRAPPER=sccache" >> "$GITHUB_ENV"
+        echo "SCCACHE_CONF=$sccache_config" >> "$GITHUB_ENV"
```
clabby, line 165: According to https://github.com/mozilla/sccache/blob/main/docs/GHA.md#github-actions, we need to set up `ACTIONS_RESULTS_URL` and `ACTIONS_RUNTIME_TOKEN` for this to work?

andresilva: sccache-action already takes care of that

.github/actions/setup/action.yml
```diff
@@ -123,3 +123,29 @@ runs:
         cache-workspace-crates: false
         workspaces: . -> ${{ steps.rust-cache-workspace.outputs.target }}
         cache-bin: true
+    - name: Setup sccache
```
clabby, line 140: Would appreciate a short doc on why we use a 2-tier cache system w/ `swatinem/rust-cache` and `sccache`. iiuc, `rust-cache` for the registry / git sources / proc macros and `sccache` for incremental artifacts (so that a `Cargo.lock` change only recompiles changed deps)?

.github/actions/setup/action.yml
clabby: neato

.github/actions/setup/action.yml
```diff
@@ -123,3 +123,29 @@ runs:
...
+      uses: mozilla-actions/sccache-action@9e7fa8a12102821edf02ca5dbea1acd0f89a2696 # v0.0.10
+    - name: Configure sccache
+      if: inputs.rust-cache-enabled == 'true'
+      shell: bash
+      run: |
+        if [ "$GITHUB_REF" = "refs/heads/main" ]; then
+          sccache_rw_mode="READ_WRITE"
+        else
+          sccache_rw_mode="READ_ONLY"
+        fi
```
clabby: I think this should actually be `if main branch and rust-cache-save == true`? As written, this'll persist to `sccache` for all partitions on main.

clabby (approved):

PR #4002, opened 2026-06-09
[storage/qmdb] expose pending batch overlay

### LINKED ISSUE #4058: [storage/qmdb] No public way to enumerate an UnmerkleizedBatch's pending overlay (only point reads)
#### Summary

A QMDB `UnmerkleizedBatch` (both `qmdb::any` and `qmdb::current`) can resolve **point reads** through its pending overlay — the batch's own mutations layered on its pending ancestors' diffs — via `get()` / `get_many()`. But there is no public way to **enumerate** that overlay. An application that needs an ordered range/prefix scan anchored at a batch's state (rather than the committed tip) therefore cannot build one, even though all the data required is already in the batch.

#### The gap

`get()`'s read-through walks `mutations -> parent diff -> ancestor diffs` (newest wins; `Some` shadows the committed value, `None` hides it), and correctly falls through to the committed DB for keys whose ancestor batch is no longer retained in the weak parent chain. That resolution is exactly what a scan would need to merge with a committed-DB range stream — but every internal that would let a downstream crate reconstruct it is crate-private:

- `UnmerkleizedBatch::mutations`
- `UnmerkleizedBatch::base` (and `Base::parent()`)
- `MerkleizedBatch::{diff, ancestor_diffs, ancestors}`

So a consumer can ask "what is the pending value of key K?" but not "what is the full set of pending (key, value/tombstone) entries?" There is no supported way to obtain the latter without forking the crate to relax visibility.

#### Use case

A `glue::stateful` application whose state machine performs **prefix/range scans** (not only point lookups) and whose execution must be anchored at a batch's state — i.e. the committed DB *plus* the batch's pending writes (its own mutations and pending ancestors). For such an application:

- Point reads work: `batch.get(k)` already returns the correctly-resolved pending-or-committed value.
- Range scans do not: to produce an ordered scan over `[prefix..]` anchored at the batch, the application must merge the committed-DB range stream with the batch's pending write-set (overlay wins; a pending tombstone hides a committed entry). That merge requires **enumerating** the pending keys, which no public API exposes.

This generalizes to any application with scan-shaped state that wants to read a consistent view "as of this batch" — e.g. speculative execution against a not-yet-committed parent, ordered iteration over a pending write-set, or merging pending state with a committed range query.

#### Proposed API

A read-only accessor on `UnmerkleizedBatch` (both `any` and `current`, the latter delegating to the inner `any` batch) returning the resolved overlay as a single map, with `get()`-identical priority:

```rust
/// Return this batch and its pending ancestors as one pending overlay.
///
/// Same priority as `get`: local mutations override parent diffs, which
/// override older ancestor diffs. `Some(value)` shadows the committed value
/// for that key; `None` hides it.
pub fn pending_overlay(&self) -> BTreeMap<U::Key, Option<U::Value>>
```

- Read-only borrow; no change to batch lifecycle or merkleization.
- Honors the documented Weak-chain behavior: an ancestor whose weak reference is no longer alive drops out of the overlay, so reads for its keys correctly fall through to the committed DB — consistent with `get()`.
- Like the ancestor diffs themselves, the overlay may include value-identical entries for keys relocated by an inactivity-floor raise, consistent with `get()` resolving through the same diffs.

#### Backwards compatibility

Purely additive — a new read-only method, no change to existing signatures, batch lifecycle, or on-disk format.

#### Proposed implementation

Implemented in #4002.

marthendalnunes:
Closes #4058

Adds a read-only `pending_overlay()` accessor to `UnmerkleizedBatch` (any + current): the batch's own mutations overlaid on its uncommitted ancestors' diffs, as one `BTreeMap<Key, Option<Value>>`.

The traversal and priority are exactly `get()`'s read-through (`mutations -> parent diff -> ancestor diffs`, newest wins, `Some` shadows the committed value, `None` hides it) — including the documented Weak-chain behavior: an ancestor that was committed and freed drops out of the overlay, and reads for its keys correctly fall through to the committed database.

### Motivation

A `glue::stateful` application whose execution must be anchored at the parent block's state (committed + uncommitted ancestors) can resolve point reads through `batch.get()`, but has no way to **enumerate** pending keys — e.g. to merge the pending write-set with a committed-db range stream for ordered prefix scans. Every internal that would allow building this view externally (`mutations`, `base`, `MerkleizedBatch::{diff, ancestor_diffs, ancestors}`) is crate-private, so applications with scan-shaped state currently have no way to obtain it. The accessor closes that gap.

### Notes

- Read-only borrow; no change to batch lifecycle or merkleization. The `current` impl delegates to the inner `any` batch.
- Like ancestor diffs themselves, the overlay may include value-identical entries for keys relocated by an inactivity-floor raise — consistent with `get()` resolving through the same diffs.
- Unit test `test_any_batch_pending_overlay` mirrors `test_any_batch_get_read_through`: committed state excluded, parentless batch = own mutations exactly, child/grandchild chains with newest-wins shadowing and `None` hiding.

### DIFF
--- a/storage/src/qmdb/any/batch.rs
+++ b/storage/src/qmdb/any/batch.rs
@@ -1004,6 +1004,28 @@ where
 
         Ok(results)
     }
+
+    /// Return this batch and its uncommitted ancestors as one pending overlay.
+    ///
+    /// The map has the same priority as `get`: local mutations override parent
+    /// diffs, which override older ancestor diffs. `Some(value)` shadows the
+    /// committed value for that key, and `None` hides it.
+    pub fn pending_overlay(&self) -> BTreeMap<U::Key, Option<U::Value>> {
+        let mut overlay = BTreeMap::new();
+        if let Some(parent) = self.base.parent() {
+            let mut chain = vec![Arc::clone(parent)];
+            chain.extend(parent.ancestors());
+            for batch in chain.into_iter().rev() {
+                for (key, entry) in batch.diff.iter() {
+                    overlay.insert(key.clone(), entry.value().cloned());
+                }
+            }
+        }
+        for (key, value) in &self.mutations {
+            overlay.insert(key.clone(), value.clone());
+        }
+        overlay
+    }
 }
 
 // Unordered-specific methods.
--- a/storage/src/qmdb/any/mod.rs
+++ b/storage/src/qmdb/any/mod.rs
@@ -1564,6 +1564,60 @@ pub(crate) mod test {
         });
     }
 
+    /// batch.pending_overlay() snapshots own mutations + uncommitted ancestor
+    /// diffs with get()'s priority: newest wins, None hides, committed state
+    /// is excluded.
+    #[test_traced("INFO")]
+    fn test_any_batch_pending_overlay() {
+        let executor = deterministic::Runner::default();
+        executor.start(|context| async move {
+            let ctx = context.child("db");
+            let mut db: UnorderedVariable = UnorderedVariableDb::init(
+                ctx.child("storage"),
+                variable_db_config::<OneCap>("po", &ctx),
+            )
+            .await
+            .unwrap();
+
+            // Committed state is NOT part of the overlay.
+            let ka = key(0);
+            commit_writes(&mut db, [(ka, Some(val(0)))], None).await;
+            let batch = db.new_batch();
+            assert!(batch.pending_overlay().is_empty());
+
+            // A parentless batch's overlay is exactly its own mutations:
+            // Some shadows the committed value, None hides it.
+            let kb = key(1);
+            let vb = val(1);
+            let batch = batch.write(kb, Some(vb)).write(ka, None);
+            let overlay = batch.pending_overlay();
+            assert_eq!(overlay.len(), 2);
+            assert_eq!(overlay.get(&kb), Some(&Some(vb)));
+            assert_eq!(overlay.get(&ka), Some(&None));
+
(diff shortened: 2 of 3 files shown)


christopherwxyz:
**Motivating use case**

Under pipelined/speculative execution, an application may need an **ordered prefix scan** over a keyspace that isn't statically nameable before execution. Point reads are already covered — `UnmerkleizedBatch::get` resolves through the pending ancestor chain (`ancestors()`, `storage/src/qmdb/any/batch.rs:1578` → `batch_chain::ancestors`, `storage/src/qmdb/batch_chain.rs:79`). But there's no *scan* that overlays uncommitted ancestor diffs, so a range read against the committed DB alone misses writes from not-yet-applied ancestor batches and diverges between nodes racing different committed tips.

`pending_overlay()` (`storage/src/qmdb/any/batch.rs:1013`) closes that gap: it materializes the batch + its pending ancestors as one `BTreeMap<Key, Option<Value>>`, which the consumer merges with a committed-base range stream for an ancestor-aware prefix scan.

**Determinism:** read-only, a pure function of the batch's live ancestry (newest-wins, local mutations last). Intended use is a single snapshot at the start of a block's execution, consumed before propose/verify returns; the app writes via its own execution overlay and leaves the batch write-once until merkleize. Only interaction is this one read — nothing persists into the chain past the snapshot. Liveness is the existing batch-chain invariant (`parent: Option<Weak<Self>>`, `batch.rs:224`; ancestors must outlive merkleize).

**Possible refinement:** a prefix/range-bounded variant to avoid materializing the full overlay when the consumer only needs a slice.

PR #4003, opened 2026-06-09
[cryptography] Add MAYO scheme

clabby:
## Overview

Adds a wrapper around the MAYO signature scheme in `commonware-cryptography`.

### DIFF
--- a/cryptography/src/mayo/mod.rs
+++ b/cryptography/src/mayo/mod.rs
@@ -1,0 +1,53 @@
+//! MAYO implementations of the [crate::Verifier] and [crate::Signer] traits.
+//!
+//! MAYO is a multivariate quadratic signature scheme based on the Oil and Vinegar
+//! trapdoor and is a candidate in the NIST post-quantum signature standardization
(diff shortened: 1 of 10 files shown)


cryptography/src/mayo/mod.rs
```diff
@@ -0,0 +1,53 @@
...
+//! | [mayo5]   | 5          | 40 bytes    | 5554 bytes | 964 bytes |
+//!
+//! MAYO-2 trades a much larger public key for the smallest signatures. A
+//! private key serializes as its secret seed; the public key is re-derived
+//! on decode.
+//!
+//! # Randomized Signatures
+//!
+//! MAYO signing draws a fresh salt (from the operating system, inside MAYO-C)
+//! for every signature, so signing the same message twice produces different
```
patrick-ogrady, line 29: I think there is a trait message somewhere about requiring deterministic signatures we'd want to loosen?

patrick-ogrady: I don't think that would break anything, moreso just to be clear.

PR #4004, opened 2026-06-09
[runtime] key the page cache maps with ahash

roberto-bayardo:
## [runtime] Key the page cache maps with ahash

  The page cache's `index` and `page_fetches` maps hash a `(blob id, page number)` key on every cached read and insert. Swap std's SipHash for `ahash::RandomState`, which is several times faster on short fixed-width keys while remaining DoS-resistant (seeded from OS entropy via `runtime-rng`, same as `storage`'s existing `Hashed` translator).

Cached reads (`buffer_paged::read`) improve up to ~37-39% for small reads, where the per-page map lookup dominates, tapering to noise at page-sized reads where the copy dominates:
  
  | read size | deterministic | tokio |
  |---|---|---|
  | 64 B | -37.4% | -26.5% |
  | 256 B | -39.2% | -26.5% |
  | 1 KiB | -10.9% | -8.9% |
  | 4 KiB | -1.3% (noise) | -2.0% |

  Paged buffer appends improve ~10-12% on both backends:

  ```
  buffer_paged::append/backend=deterministic chunk=64
    time:   [197.91 ms 198.18 ms 198.49 ms]
    change: [-13.250% -11.684% -10.186%] (p = 0.00 < 0.05)
  
  buffer_paged::append/backend=tokio chunk=64
    time:   [112.70 ms 113.06 ms 113.52 ms]
    change: [-18.474% -10.729% -4.1737%] (p = 0.00 < 0.05)
  ```
  
  Since `runtime` and `storage` now both depend on ahash, the pin is hoisted to `[workspace.dependencies]`; feature resolution (and `Cargo.lock`) is unchanged.

### DIFF
--- a/runtime/Cargo.toml
+++ b/runtime/Cargo.toml
@@ -14,6 +14,7 @@ documentation = "https://docs.rs/commonware-runtime"
 workspace = true
 
 [dependencies]
+ahash = { workspace = true, features = ["std", "runtime-rng"] }
 arbitrary = { workspace = true, optional = true }
 bytes.workspace = true
 cfg-if.workspace = true
--- a/runtime/src/utils/buffer/paged/cache.rs
+++ b/runtime/src/utils/buffer/paged/cache.rs
@@ -105,10 +105,6 @@ struct Cache {
     /// The page cache index, with a key composed of (blob id, page number), that maps each cached
     /// page to the index of its slot in `entries` and `slots`.
     ///
-    /// An [AHashMap] (randomly seeded via runtime-rng) rather than std's default SipHash: ahash
-    /// remains DoS-resistant for adversarially influenced page numbers but is several times
-    /// faster on these fixed-width keys, which are hashed on every cached read and insert.
-    ///
     /// # Invariants
     ///
     /// Each `index` entry maps to exactly one `entries` slot, and that entry always has a
--- a/runtime/src/utils/buffer/paged/cache.rs
+++ b/runtime/src/utils/buffer/paged/cache.rs
@@ -104,12 +104,17 @@ struct Cache {
     /// The page cache index, with a key composed of (blob id, page number), that maps each cached
     /// page to the index of its slot in `entries` and `slots`.
     ///
+    /// Keyed with [ahash::RandomState] (randomly seeded via runtime-rng) instead of std's default
+    /// SipHash: ahash remains DoS-resistant for adversarially influenced page numbers but is
+    /// several times faster on these fixed-width keys, which are hashed on every cached read and
+    /// insert.
+    ///
     /// # Invariants
     ///
     /// Each `index` entry maps to exactly one `entries` slot, and that entry always has a
     /// matching key. (The converse is not true: after [Self::invalidate_from] a slot may retain
     /// a stale key that is no longer present in `index`.)
-    index: HashMap<(u64, u64), usize>,
+    index: HashMap<(u64, u64), usize, ahash::RandomState>,
 
     /// Metadata for each cache slot.
     ///
@@ -134,7 +139,7 @@ struct Cache {
 
     /// A map of currently executing page fetches to ensure only one task at a time is trying to
     /// fetch a specific page.
-    page_fetches: HashMap<(u64, u64), PageFetchEntry>,
+    page_fetches: HashMap<(u64, u64), PageFetchEntry, ahash::RandomState>,
 }
 
 /// Metadata for a single cache entry (page data stored in per-slot buffers).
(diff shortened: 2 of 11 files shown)


runtime/src/utils/buffer/paged/cache.rs
```diff
@@ -104,12 +104,17 @@ struct Cache {
...
+    /// several times faster on these fixed-width keys, which are hashed on every cached read and
+    /// insert.
+    ///
     /// # Invariants
     ///
     /// Each `index` entry maps to exactly one `entries` slot, and that entry always has a
     /// matching key. (The converse is not true: after [Self::invalidate_from] a slot may retain
     /// a stale key that is no longer present in `index`.)
-    index: HashMap<(u64, u64), usize>,
+    index: HashMap<(u64, u64), usize, ahash::RandomState>,
```
patrick-ogrady: Should we do this on all HashMaps?

roberto-bayardo: Probably! We've been doing this on the most performance-critical ones and it's consistently a win....

patrick-ogrady: I think you can just do `AHashMap`?

patrick-ogrady: Maybe you need to construct manually for rng?

roberto-bayardo: no you were right it wasn't needed, changed to AHashMap.

runtime/src/utils/buffer/paged/cache.rs
```diff
@@ -104,12 +105,16 @@ struct Cache {
...
+    /// remains DoS-resistant for adversarially influenced page numbers but is several times
+    /// faster on these fixed-width keys, which are hashed on every cached read and insert.
+    ///
     /// # Invariants
     ///
     /// Each `index` entry maps to exactly one `entries` slot, and that entry always has a
     /// matching key. (The converse is not true: after [Self::invalidate_from] a slot may retain
     /// a stale key that is no longer present in `index`.)
-    index: HashMap<(u64, u64), usize>,
+    index: AHashMap<(u64, u64), usize>,
```
patrick-ogrady, line 113: I suspect that `index` may also benefit from this?

runtime/Cargo.toml
```diff
@@ -14,6 +14,7 @@ documentation = "https://docs.rs/commonware-runtime"
 workspace = true
 
 [dependencies]
+ahash = { workspace = true, features = ["std", "runtime-rng"] }
```
patrick-ogrady, line 17: These are the default features but still think they are worthwhile to specify explicitly.

patrick-ogrady:
```
  Final results (10 samples per config, all p = 0.00):

  ┌─────────────────────────────┬──────────┬──────────┬────────┐
  │    Config (leaves=100k)     │ BTreeMap │  ahash   │ Change │
  ├─────────────────────────────┼──────────┼──────────┼────────┤
  │ updates=1M, serial, mmr     │ 110.6 ms │ 91.4 ms  │ -17.3% │
  ├─────────────────────────────┼──────────┼──────────┼────────┤
  │ updates=1M, parallel, mmr   │ 80.5 ms  │ 67.0 ms  │ -16.8% │
  ├─────────────────────────────┼──────────┼──────────┼────────┤
  │ updates=100k, serial, mmr   │ 63.6 ms  │ 50.3 ms  │ -21.0% │
  ├─────────────────────────────┼──────────┼──────────┼────────┤
  │ updates=100k, parallel, mmr │ 36.2 ms  │ 26.8 ms  │ -25.8% │
  ├─────────────────────────────┼──────────┼──────────┼────────┤
  │ updates=1M, serial, mmb     │ 118.6 ms │ 100.4 ms │ -14.2% │
  ├─────────────────────────────┼──────────┼──────────┼────────┤
  │ updates=1M, parallel, mmb   │ 86.8 ms  │ 73.2 ms  │ -16.8% │
  ├─────────────────────────────┼──────────┼──────────┼────────┤
  │ updates=100k, serial, mmb   │ 67.7 ms  │ 54.9 ms  │ -18.5% │
  ├─────────────────────────────┼──────────┼──────────┼────────┤
  │ updates=100k, parallel, mmb │ 41.2 ms  │ 30.8 ms  │ -24.9% │
  └─────────────────────────────┴──────────┴──────────┴────────┘
```

patrick-ogrady:
I think there is a world where we use this in commonware-utils and require all HashMaps to use this. But, will reserve that for later I think.

https://github.com/commonwarexyz/monorepo/issues/4006

patrick-ogrady (approved):

PR #4005, opened 2026-06-09
[parallel] add Strategy::run_async for offloading CPU work from worker thread

roberto-bayardo:
## Offload batch leaf hashing onto the strategy

  The authenticated journal's `add_many` hashes every appended item into a leaf digest (roughly half of merkleization CPU for a qmdb batch). It was already parallelized, but
  `Rayon::install` blocks the calling thread, pinning a tokio worker for tens of milliseconds on large batches. It now runs through `Strategy::run_async`: still parallel,
  but the calling task suspends instead of blocking its thread. First production use of `run_async`, to evaluate on a busy validator before converting parent-node hashing
  (blocked on off-lock merkleization).

  ### API changes (ALPHA)

  - `add_many` and the keyless/immutable batch `merkleize` fns are now async.
  - `ValueEncoding`/`FixedValue`/`VariableValue` gain `'static` bounds so owned items can cross the offload boundary (`Key` already had one).
  - A test pairing `Rayon` with the deterministic runner moved to the tokio runner: the deterministic runtime cannot observe wakeups from external pool threads.

  ### Validation

  Full storage/glue/sync suites pass; conformance roots are unchanged, confirming the relocation is behavior-preserving. The caller still awaits the root, so merkleize
  benches won't move — the win is co-scheduled tasks no longer stalling behind hashing on executor threads, visible in validator tail latency rather than throughput.

### DIFF
--- a/parallel/src/lib.rs
+++ b/parallel/src/lib.rs
@@ -62,7 +62,7 @@
 
 commonware_macros::stability_scope!(BETA {
     use cfg_if::cfg_if;
-    use core::fmt;
+    use core::{fmt, future::Future};
 
     cfg_if! {
         if #[cfg(feature = "std")] {
@@ -228,6 +228,24 @@ commonware_macros::stability_scope!(BETA {
                     a
                 },
             )
+        }
+
+        /// Runs `f` to completion and returns its result.
+        ///
+        /// When the strategy owns a thread pool, the returned future suspends the caller while
+        /// the work executes on the pool; otherwise the work runs inline when the future is
+        /// first polled. Use this to keep CPU-intensive work off async executor threads.
+        /// Strategy methods invoked within `f` execute with this strategy's parallelism.
+        ///
+        /// # Panics
+        ///
+        /// The returned future panics if `f` panics.
+        fn run_async<F, R>(&self, f: F) -> impl Future<Output = R> + Send
+        where
+            F: FnOnce() -> R + Send + 'static,
+            R: Send + 'static,
+        {
+            async move { f() }
         }
 
         /// Maps each element with per-partition state and collects results into a `Vec`.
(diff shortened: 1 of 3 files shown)


parallel/src/lib.rs
```diff
@@ -558,6 +576,19 @@ commonware_macros::stability_scope!(BETA, cfg(feature = "std") {
             self.thread_pool.install(|| rayon::join(a, b))
         }
 
+        fn run_async<F, R>(&self, f: F) -> impl Future<Output = R> + Send
+        where
+            F: FnOnce() -> R + Send + 'static,
+            R: Send + 'static,
+        {
+            let (tx, rx) = futures::channel::oneshot::channel();
+            self.thread_pool.spawn(move || {
```
patrick-ogrady, line 585: does this take away from rayon's throughput?

roberto-bayardo: not according to Fable

```
for the parallel work itself, throughput is identical. The closure's host thread isn't an overhead thread sitting on top of the workers; it is one of the N
  workers, and when the inner map_collect_vec runs, it's hashing alongside the rest of the pool. The pool delivers the same N-way parallelism it did under the old blocking
  install — rayon's install-from-within-the-pool path exists precisely so nested parallelism doesn't lose a thread.

  The only costs that exist at all:

  - A one-time handoff per run_async call (oneshot send + waker) — microseconds, amortized over milliseconds of hashing.
  - Sequential code inside a closure holds one worker while it runs. But that work was going to consume one thread somewhere regardless — it's just a rayon thread now
  instead of a tokio thread, which is the whole point.

  And under concurrency it's arguably a throughput improvement in aggregate: previously, several tasks merkleizing at once meant several tokio workers all blocked in
  install, contending for the same pool while contributing nothing. Now those tasks suspend, the closures queue in rayon's injector, and the pool chews through them at full
  utilization with no parked threads anywhere.
```

PR #4007, opened 2026-06-10
[qmdb] Minimal Merkle Tweaks

patrick-ogrady:
Replaces: #3993

Speeds up the QMDB per-block pipeline (load keys, write batch, merkleize, root) at the production validator shape: 32k updates over 1M keys, EightCap. All numbers are p50 from interleaved pinned-binary A/B runs on tokio (Apple Silicon) against the merge-base (13b4a18f), and the speculative root is byte-identical to main on every iteration of every variant.

### constantinople (load + write + merkleize + root)

| variant | depth | main | this PR | delta |
|---|---|---|---|---|
| any::unordered::fixed::mmb | 0 | 41.5ms | 29.9ms | -28% |
| any::unordered::fixed::mmb | 1 | 49.7ms | 37.4ms | -25% |
| any::ordered::fixed::mmb | 0 | 49.1ms | 43.2ms | -12% |
| any::ordered::fixed::mmb | 1 | 64.7ms | 57.4ms | -11% |
| current::unordered::fixed::mmb | 0 | 44.2ms | 34.8ms | -21% |
| current::unordered::fixed::mmb | 1 | 60.8ms | 50.6ms | -17% |
| current::ordered::fixed::mmb | 0 | 53.6ms | 47.6ms | -11% |
| current::ordered::fixed::mmb | 1 | 70.8ms | 67.1ms | -5% |

### criterion `qmdb::merkleize` (write-only merkleize, keys=1M, 100k updates/iter)

| variant | main | this PR | delta |
|---|---|---|---|
| any::unordered::fixed::mmr | 188.5ms | 172.5ms | -8.5% |
| any::ordered::fixed::mmr | 270.7ms | 265.0ms | -2.1% |

The criterion bench is write-only (no reads before writes), so it isolates the non-caching optimizations; constantinople exercises the full read-then-write pipeline where the read cache applies. (Criterion numbers were measured before the ordered-cache removal and the windowed cache probe; both are no-ops at this write-only shape.)

## Optimizations

**1. Batch reads are fused into unordered merkleize (default-on).** `get`/`get_many` on an `UnmerkleizedBatch` cache committed-DB locations on the batch (`Mutex<AHashMap<Key, Location>>`, never held across an await). Unordered `merkleize` consumes them: keys loaded and then written skip the merkleize-time journal re-read, which existed solely to re-resolve what the read already resolved. Any read-then-write flow on one batch object (the standard block-execution pattern, including through `current` and glue) fuses automatically; no API change.

- Only committed-snapshot resolutions are cached. Keys answered by pending mutations or ancestor diffs cache nothing: ancestor-diff caching is unsound if that ancestor is committed and dropped before merkleize.
- Ordered batches cache nothing (`CACHES_READS = false`): op generation needs each read operation's value and next key for linkage, and caching those payloads measured as a net loss (see "explored but not included").

**2. Sorted index probes for batch reads.** `index::Unordered` gains a defaulted `get_many(keys, visit)` whose probe order is implementation-defined. The tree-backed indexes (`index::ordered`, `index::partitioned::ordered`) override it to probe in translated-key order, so consecutive BTreeMap descents share upper node paths that stay cache-resident across the batch. Hash-backed indexes keep the plain loop (probe order is irrelevant there and the sort would be pure overhead). This took the ordered load phase from 12.5ms to 9.0ms at the bench shape; profiling attributed 6.5ms of the ordered load to random-order BTreeMap descents (vs 0.7ms for unordered's HashMap probes).

**3. Floor-raise scan batches bitmap locks and resolution.** `Shared::fill_candidates` fills each floor-raise candidate batch under a single bitmap read guard (previously two lock acquisitions per candidate, ~33k per merkleize at this shape). The per-candidate activity check and previous-committed-location lookup now share one diff binary-search and ancestor resolution (replaces `is_active_at`, which duplicated both). The per-round candidate cap (`MAX_CONCURRENT_READS = 64`) is removed: a round never requests more than the steps still needed, so the cap only throttled cold-path read concurrency (warm delta measured zero). `next_candidate` is kept as a test-module oracle and `bitmap_fill_candidates_matches_oracle` proves sequence equality across pruned/truncated/tail shapes, tips, and limits.

**4. One-lock batched journal reads.** Contiguous fixed `Reader::read_many` now routes every position through the segmented journal's `get_many` -> `Append::read_many_into`, which serves page-cache and tip-buffer hits under amortized lock acquisitions and reads only true misses from the blob, concurrently. This replaces the per-item synchronous probe (one lock acquisition per item): ~40-48% faster `db.get_many` over 32k warm keys. `read_many_into` returns the count of items served without a blob read, so `cache_hits`/`cache_misses` carry exact blob-read semantics.

## Bench

`constantinople`: full per-block pipeline harness (`get_many` load -> writes -> merkleize -> root) on tokio with EightCap, with configurable pending-ancestor depth, a per-iteration load/write/merkleize phase split, and the speculative root printed as a cross-binary parity check. Runs over `mmb` (the production-default merkle structure). Supports the qmdb criterion bench variant names: `any::unordered::fixed::mmb`, `any::ordered::fixed::mmb`, `any::unordered::variable::mmb`, `current::unordered::fixed::mmb`, `current::ordered::fixed::mmb`. It no-ops under blanket harness invocations (`cargo bench --benches`, `--output-format bencher`) so benchmark CI is unaffected.

```
cargo bench -p commonware-storage --bench constantinople -- <db> [depth] [iters] [keys] [updates]
```

## Optimizations explored but not included

- **Ordered payload caching** (cache `(old_value, next_key)` at read time so ordered merkleize can skip the cached keys' own re-reads): measured a net loss at the bench shape and removed. Load gained ~1.4-1.6ms/block from the per-read payload clones and fatter cache entries, while merkleize was unchanged at both depths (the skipped re-reads are warm page-cache hits served under one batched lock acquisition); roots byte-identical with and without. Could only pay off when the page cache cannot hold the touched ops between load and merkleize. A correctness wrinkle it carried for free: a deleted key's own-bucket read doubles as same-bucket predecessor discovery, so deletes could never be skipped.
- **Parallel index probes** (fan keys across `Strategy` workers): measured ordered load 9.0 -> 6.2ms (-7.5% total at depth 0, Rayon=8), but it needs an index-kind gate (rayon dispatch exceeds the entire hash-probe phase, so it must apply only to tree-backed indexes) and the validator default of 2 rayon threads halves the win. Deferred to a "parallel load" follow-up together with the other profiled load targets: parallel decode in the journal read, a pooled uninit staging buffer (`read_many` zeroes 3.3MB per 32k-key call, ~0.65ms), page-run copy reuse in `read_cached_many` (~0.8ms), and a linear-merge match-back (~0.3ms).
- **Variable journal `read_many` batching** (mirror of optimization 4): attempted and measured a wash (33.6 vs 34.1ms p50) -- the offsets journal is small and dense, so its per-item probes were never the cost. Variable's load cost is the per-item data probes, and items carry varint length prefixes, so batching them needs a two-pass ranged read (batched prefix reads, then one ranged payload read). Deferred; the `any::unordered::variable::mmb` harness mode is included to measure that follow-up.
- **Hash side-index for ordered point reads**: would erase the BTreeMap probe cost but doubles index memory; superseded by sorted probes (optimization 2) at zero memory cost.
- **Stable probe sort**: microbenched stable vs unstable on the probe-order pairs; unstable wins ~28% on random caller key order. Stable only wins on append-to-sorted input shapes, which these call sites never see.

## Testing

Root-parity tests assert read-then-write merkleize is byte-identical to write-only merkleize across updates/deletes/creates and pending-ancestor depths 0-2: unordered, ordered (TwoCap collision-heavy; ordered ignores the read cache, so this pins that reads cannot perturb its root), and `current`. Location caching across an intervening ancestor commit+drop is covered differentially (`test_unordered_fixed_caching_survives_ancestor_commit`). Journal batched reads are byte-compared against per-item reads with exact hit/miss accounting, including tip-straddling items whose prefix misses the cache.

### DIFF
--- a/glue/src/stateful/db/any.rs
+++ b/glue/src/stateful/db/any.rs
@@ -92,8 +92,8 @@ where
 
     /// Read multiple values by key, falling back to committed state.
     ///
-    /// Returns results in the same order as the input keys. Locations resolved by the read are
-    /// retained on the batch, so the subsequent merkleize skips re-reading those keys.
+    /// Returns results in the same order as the input keys. Each unique read is cached on the
+    /// batch for reuse at merkleize, so memory grows with the number of unique keys read.
     pub async fn get_many(&self, keys: &[&K]) -> Result<Vec<Option<V::Value>>, Error<F>> {
         let db = self.db.read().await;
         self.batch.get_many(keys, &*db).await
@@ -187,8 +187,8 @@ where
 
     /// Read multiple values by key, falling back to committed state.
     ///
-    /// Returns results in the same order as the input keys. Locations resolved by the read are
-    /// retained on the batch, so the subsequent merkleize skips re-reading those keys.
+    /// Returns results in the same order as the input keys. Each unique read is cached on the
+    /// batch for reuse at merkleize, so memory grows with the number of unique keys read.
     pub async fn get_many(&self, keys: &[&K]) -> Result<Vec<Option<V::Value>>, Error<F>> {
         let db = self.db.read().await;
         self.batch.get_many(keys, &*db).await
--- a/glue/src/stateful/db/any.rs
+++ b/glue/src/stateful/db/any.rs
@@ -92,7 +92,8 @@ where
 
     /// Read multiple values by key, falling back to committed state.
     ///
-    /// Returns results in the same order as the input keys.
+    /// Returns results in the same order as the input keys. Locations resolved by the read are
+    /// retained on the batch, so the subsequent merkleize skips re-reading those keys.
     pub async fn get_many(&self, keys: &[&K]) -> Result<Vec<Option<V::Value>>, Error<F>> {
         let db = self.db.read().await;
         self.batch.get_many(keys, &*db).await
--- a/glue/src/stateful/db/any.rs
+++ b/glue/src/stateful/db/any.rs
@@ -187,7 +187,8 @@ where
 
     /// Read multiple values by key, falling back to committed state.
     ///
-    /// Returns results in the same order as the input keys.
+    /// Returns results in the same order as the input keys. Locations resolved by the read are
+    /// retained on the batch, so the subsequent merkleize skips re-reading those keys.
     pub async fn get_many(&self, keys: &[&K]) -> Result<Vec<Option<V::Value>>, Error<F>> {
         let db = self.db.read().await;
         self.batch.get_many(keys, &*db).await
--- a/glue/src/stateful/db/any.rs
+++ b/glue/src/stateful/db/any.rs
@@ -187,8 +187,7 @@ where
 
     /// Read multiple values by key, falling back to committed state.
     ///
-    /// Returns results in the same order as the input keys. Each unique read is cached on the
-    /// batch for reuse at merkleize, so memory grows with the number of unique keys read.
+    /// Returns results in the same order as the input keys.
     pub async fn get_many(&self, keys: &[&K]) -> Result<Vec<Option<V::Value>>, Error<F>> {
         let db = self.db.read().await;
         self.batch.get_many(keys, &*db).await
--- a/runtime/src/utils/buffer/paged/append.rs
+++ b/runtime/src/utils/buffer/paged/append.rs
@@ -611,9 +611,6 @@ impl<B: Blob> Append<B> {
         offsets: &[u64],
         item_size: usize,
     ) -> Result<usize, Error> {
-        if offsets.is_empty() {
-            return Ok(0);
-        }
         assert_eq!(
             buf.len(),
             offsets
@@ -626,6 +623,9 @@ impl<B: Blob> Append<B> {
             offsets.windows(2).all(|w| w[0] < w[1]),
             "offsets must be strictly increasing"
         );
+        if offsets.is_empty() {
+            return Ok(0);
+        }
 
         let last_end = offsets[offsets.len() - 1]
             .checked_add(item_size as u64)
--- a/runtime/src/utils/buffer/paged/append.rs
+++ b/runtime/src/utils/buffer/paged/append.rs
@@ -616,6 +616,10 @@ impl<B: Blob> Append<B> {
                 .expect("read_many_into buffer length overflow"),
             "read_many_into requires buf.len() == offsets.len() * item_size"
         );
+        assert!(
+            offsets.windows(2).all(|w| w[0] < w[1]),
+            "offsets must be strictly increasing"
+        );
         if offsets.is_empty() {
             return Ok(0);
         }
--- a/runtime/src/utils/buffer/paged/append.rs
+++ b/runtime/src/utils/buffer/paged/append.rs
@@ -599,12 +599,15 @@ impl<B: Blob> Append<B> {
     /// `buf` must be exactly `offsets.len() * item_size` bytes. All offsets must be sorted,
     /// non-overlapping, and within bounds. This amortizes lock acquisition and avoids
     /// per-item buffer allocation compared to calling [`read_at`](Self::read_at) in a loop.
+    ///
+    /// Returns the number of items fully served without a blob read (from the tip buffer and the
+    /// page cache). The remaining items required at least one blob read.
     pub async fn read_many_into(
         &self,
         buf: &mut [u8],
         offsets: &[u64],
         item_size: usize,
-    ) -> Result<(), Error> {
+    ) -> Result<usize, Error> {
         assert_eq!(
             buf.len(),
             offsets
@@ -614,7 +617,7 @@ impl<B: Blob> Append<B> {
             "read_many_into requires buf.len() == offsets.len() * item_size"
         );
         if offsets.is_empty() {
-            return Ok(());
+            return Ok(0);
         }
 
         let last_end = offsets[offsets.len() - 1]
@@ -633,7 +636,7 @@ impl<B: Blob> Append<B> {
         // `chunks_exact_mut` yields disjoint per-item slots, so we never have to
         // reborrow the parent buffer while cache/blob destinations remain live.
         if item_size == 0 {
-            return Ok(());
+            return Ok(offsets.len());
         }
         let mut cache_ranges: Vec<(&mut [u8], u64)> = Vec::new();
         for (item_buf, &offset) in buf.chunks_exact_mut(item_size).zip(offsets.iter()) {
@@ -657,14 +660,15 @@ impl<B: Blob> Append<B> {
         drop(buffer);
 
         if cache_ranges.is_empty() {
-            return Ok(());
+            return Ok(offsets.len());
         }
 
         // Fast path: try page cache for all ranges in a single lock acquisition.
         // Fully-cached ranges are removed from cache_ranges; only misses remain.
         self.cache_ref.read_cached_many(self.id, &mut cache_ranges);
+        let blob_reads = cache_ranges.len();
         if cache_ranges.is_empty() {
-            return Ok(());
+            return Ok(offsets.len());
         }
 
         // Slow path: read only the ranges that had cache misses, concurrently.
@@ -680,7 +684,7 @@ impl<B: Blob> Append<B> {
             result?;
         }
 
-        Ok(())
+        Ok(offsets.len() - blob_reads)
     }
 
     /// Reads bytes starting at `logical_offset` into `buf`.
--- a/runtime/src/utils/buffer/paged/append.rs
+++ b/runtime/src/utils/buffer/paged/append.rs
@@ -36,12 +36,15 @@ use crate::{
 use bytes::BufMut;
 use commonware_cryptography::Crc32;
 use commonware_utils::sync::{AsyncRwLock, AsyncRwLockWriteGuard};
-use futures::stream::{FuturesUnordered, StreamExt};
+use futures::stream::StreamExt;
 use std::{
     num::{NonZeroU16, NonZeroUsize},
     sync::Arc,
 };
 use tracing::warn;
+
+/// Maximum number of blob reads issued concurrently by [`Append::read_many_into`].
+const MAX_CONCURRENT_BLOB_READS: usize = 64;
 
 /// Indicates which CRC slot in a page record must not be overwritten.
 #[derive(Clone, Copy)]
@@ -608,6 +611,9 @@ impl<B: Blob> Append<B> {
         offsets: &[u64],
         item_size: usize,
     ) -> Result<usize, Error> {
+        if offsets.is_empty() {
+            return Ok(0);
+        }
         assert_eq!(
             buf.len(),
             offsets
@@ -620,9 +626,6 @@ impl<B: Blob> Append<B> {
             offsets.windows(2).all(|w| w[0] < w[1]),
             "offsets must be strictly increasing"
         );
-        if offsets.is_empty() {
-            return Ok(0);
-        }
 
         let last_end = offsets[offsets.len() - 1]
             .checked_add(item_size as u64)
@@ -675,15 +678,17 @@ impl<B: Blob> Append<B> {
             return Ok(offsets.len());
         }
 
-        // Slow path: read only the ranges that had cache misses, concurrently.
+        // Slow path: read only the ranges that had cache misses, concurrently. Concurrency is
+        // bounded so a large cold batch cannot flood the blob with in-flight reads.
         let blob_guard = self.blob_state.read().await;
-        let mut reads = cache_ranges
+        let reads: Vec<_> = cache_ranges
             .iter_mut()
             .map(|(item_buf, offset)| {
                 self.cache_ref
                     .read(&blob_guard.blob, self.id, item_buf, *offset)
             })
-            .collect::<FuturesUnordered<_>>();
+            .collect();
+        let mut reads = futures::stream::iter(reads).buffer_unordered(MAX_CONCURRENT_BLOB_READS);
         while let Some(result) = reads.next().await {
             result?;
         }
@@ -1531,6 +1536,62 @@ mod tests {
                 .await
                 .unwrap();
 
+            let read: Vec<u8> = synced.iter().chain(tip.iter()).copied().collect();
+            for (i, &off) in offsets.iter().enumerate() {
+                assert_eq!(
+                    &buf[i * item_size..(i + 1) * item_size],
+                    &read[off as usize..off as usize + item_size],
+                );
+            }
+        });
+    }
+
+    #[test_traced("DEBUG")]
+    fn test_read_many_into_straddle_prefix_miss() {
+        // A straddling item whose synced prefix page is NOT in the page cache: the
+        // suffix is copied from the tip buffer and the prefix is read from the blob
+        // without clobbering it, and the item is counted as a blob read.
+        let executor = deterministic::Runner::default();
+        executor.start(|context: deterministic::Context| async move {
+            let (blob, blob_size) = context
+                .open("test_partition", b"rmany_smiss")
+                .await
+                .unwrap();
+            // Single-page cache so residency is deterministic.
+            let cache_ref = CacheRef::from_pooler(&context, PAGE_SIZE, NZUsize!(1));
+            let append = Append::new(blob, blob_size, BUFFER_SIZE, cache_ref)
+                .await
+                .unwrap();
+
+            // Write 3 pages and sync, then a partial page that stays in the tip.
+            let synced: Vec<u8> = (0u8..=255)
+                .cycle()
+                .take(PAGE_SIZE.get() as usize * 3)
+                .collect();
+            append.append(&synced).await.unwrap();
+            append.sync().await.unwrap();
+            let item_size = 10;
+            let tip: Vec<u8> = (100u8..=255)
+                .cycle()
+                .take(PAGE_SIZE.get() as usize / 2)
+                .collect();
+            append.append(&tip).await.unwrap();
+
+            // Fault page 0 in, evicting whatever sync left resident, so the straddle
+            // prefix page (page 2) is guaranteed not cached.
+            let _ = append.read_at(0, item_size).await.unwrap();
+
+            let straddle_off = synced.len() as u64 - (item_size as u64 / 2);
+            let tip_off = synced.len() as u64 + item_size as u64;
+            let offsets = [straddle_off, tip_off];
+            let mut buf = vec![0u8; offsets.len() * item_size];
+            let hits = append
+                .read_many_into(&mut buf, &offsets, item_size)
+                .await
+                .unwrap();
+
+            // The tip-only item is a hit; the straddle item required a blob read.
+            assert_eq!(hits, 1);
             let read: Vec<u8> = synced.iter().chain(tip.iter()).copied().collect();
             for (i, &off) in offsets.iter().enumerate() {
                 assert_eq!(
--- a/runtime/src/utils/buffer/paged/append.rs
+++ b/runtime/src/utils/buffer/paged/append.rs
@@ -36,15 +36,12 @@ use crate::{
 use bytes::BufMut;
 use commonware_cryptography::Crc32;
 use commonware_utils::sync::{AsyncRwLock, AsyncRwLockWriteGuard};
-use futures::stream::StreamExt;
+use futures::stream::{FuturesUnordered, StreamExt};
 use std::{
     num::{NonZeroU16, NonZeroUsize},
     sync::Arc,
 };
 use tracing::warn;
-
-/// Maximum number of blob reads issued concurrently by [`Append::read_many_into`].
-const MAX_CONCURRENT_BLOB_READS: usize = 64;
 
 /// Indicates which CRC slot in a page record must not be overwritten.
 #[derive(Clone, Copy)]
@@ -678,17 +675,15 @@ impl<B: Blob> Append<B> {
             return Ok(offsets.len());
         }
 
-        // Slow path: read only the ranges that had cache misses, concurrently. Concurrency is
-        // bounded so a large cold batch cannot flood the blob with in-flight reads.
+        // Slow path: read only the ranges that had cache misses, concurrently.
         let blob_guard = self.blob_state.read().await;
-        let reads: Vec<_> = cache_ranges
+        let mut reads = cache_ranges
             .iter_mut()
             .map(|(item_buf, offset)| {
                 self.cache_ref
                     .read(&blob_guard.blob, self.id, item_buf, *offset)
             })
-            .collect();
-        let mut reads = futures::stream::iter(reads).buffer_unordered(MAX_CONCURRENT_BLOB_READS);
+            .collect::<FuturesUnordered<_>>();
         while let Some(result) = reads.next().await {
             result?;
         }
--- a/runtime/src/utils/buffer/paged/cache.rs
+++ b/runtime/src/utils/buffer/paged/cache.rs
@@ -247,49 +247,31 @@ impl CacheRef {
         original_len - buf.len()
     }
 
-    /// Read multiple disjoint byte ranges from the page cache, amortizing lock acquisition.
+    /// Read multiple disjoint byte ranges from the page cache in a single lock acquisition.
     ///
     /// Each element of `ranges` is `(dest_slice, logical_offset)`. Fully-cached ranges have
     /// their data written to the destination slice and are removed from `ranges`. Entries left
     /// in `ranges` correspond to cache misses that the caller must read from the underlying
     /// blob.
-    ///
-    /// Ranges are probed in fixed-size windows with the lock re-acquired per window, bounding
-    /// how long any one batch holds the cache lock: the lock is write-preferring, so a long
-    /// uninterrupted hold would stall a queued writer and every reader behind it.
     pub(super) fn read_cached_many(&self, blob_id: u64, ranges: &mut Vec<(&mut [u8], u64)>) {
-        const WINDOW: usize = 1024;
-        let len = ranges.len();
-        let mut kept = 0;
-        let mut start = 0;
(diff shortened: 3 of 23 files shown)


storage/src/qmdb/any/operation/update/mod.rs
```diff
@@ -23,6 +23,16 @@ pub trait Update: sealed::Sealed + Clone + Send + Sync {
     /// The value encoding (fixed or variable).
     type ValueEncoding: ValueEncoding<Value = Self::Value>;
 
+    /// Data retained from a committed-DB read for reuse at merkleize, alongside the location.
+    ///
+    /// Unordered op generation needs only the location, so it retains nothing extra. Ordered op
```
patrick-ogrady: remove this

roberto-bayardo:
Would be great to see the impact on (some of) the merkleize benchmarks?

roberto-bayardo:
Can you extend the Constantinople test harness to also test ordered dbs?  The changes in this PR would seem to impact ordered databases a bit more than unordered, so would be good to confirm the changes hold up there (and don't cause regressions).

glue/src/stateful/db/any.rs
```diff
@@ -92,7 +92,8 @@ where
 
     /// Read multiple values by key, falling back to committed state.
     ///
-    /// Returns results in the same order as the input keys.
+    /// Returns results in the same order as the input keys. Locations resolved by the read are
+    /// retained on the batch, so the subsequent merkleize skips re-reading those keys.
```
roberto-bayardo: "retained on the batch" - > cached?

glue/src/stateful/db/any.rs
```diff
@@ -186,7 +187,8 @@ where
...
```
roberto-bayardo: "retained on the batch" - > cached?

glue/src/stateful/db/any.rs
roberto-bayardo: Although I'm not sure you need this level of implementation detail in the public function doc.

Maybe just note that each unique read is cached leading to O(# reads) in memory usage?

runtime/src/utils/buffer/paged/append.rs
```diff
@@ -613,8 +616,12 @@ impl<B: Blob> Append<B> {
                 .expect("read_many_into buffer length overflow"),
             "read_many_into requires buf.len() == offsets.len() * item_size"
         );
+        assert!(
+            offsets.windows(2).all(|w| w[0] < w[1]),
+            "offsets must be strictly increasing"
+        );
         if offsets.is_empty() {
```
roberto-bayardo, line 623: feels like this should be at the top of the function (before the assertions on offsets).

patrick-ogrady:
> Can you extend the Constantinople test harness to also test ordered dbs? The changes in this PR would seem to impact ordered databases a bit more than unordered, so would be good to confirm the changes hold up there (and don't cause regressions).

No regression but wins aren't as pronounced:

```
1. Criterion qmdb::merkleize benchmarks (1M keys, 100k updates/iter)

These are write-only merkleize (no reads first), so retention never fires — this isolates the floor-scan batching and journal read-path changes:

┌────────────────────────────┬────────────────────────┬────────────────────────┬───────┐
│          variant           │          main          │         branch         │ delta │
├────────────────────────────┼────────────────────────┼────────────────────────┼───────┤
│ any::unordered::fixed::mmr │ 188.5ms [182.3, 198.5] │ 172.5ms [170.7, 174.7] │ −8.5% │
├────────────────────────────┼────────────────────────┼────────────────────────┼───────┤
│ any::ordered::fixed::mmr   │ 270.7ms [267.2, 274.1] │ 265.0ms [258.1, 272.8] │ −2.1% │
└────────────────────────────┴────────────────────────┴────────────────────────┴───────┘

2. Constantinople ordered (32k updates / 1M keys, read-then-write — retention fires)

Extended the harness with any-ordered and current-ordered modes (uncommitted, fmt-clean, smoke-tested), A/B'd against an origin/main build of the same harness:

┌────────────────────┬──────────────┬──────────────┬───────┐
│       config       │   main p50   │  branch p50  │ delta │
├────────────────────┼──────────────┼──────────────┼───────┤
│ any-ordered d0     │ 46.60, 47.05 │ 43.57, 45.34 │ ≈ −5% │
├────────────────────┼──────────────┼──────────────┼───────┤
│ any-ordered d1     │ 62.00, 64.82 │ 60.31, 58.93 │ ≈ −6% │
├────────────────────┼──────────────┼──────────────┼───────┤
│ current-ordered d0 │ 50.32        │ 48.94        │ −2.7% │
├────────────────────┼──────────────┼──────────────┼───────┤
│ current-ordered d1 │ 69.43        │ 67.85        │ −2.3% │
└────────────────────┴──────────────┴──────────────┴───────┘


Why ordered gains are smaller than unordered's −17/−22%: ordered merkleize carries large costs this PR doesn't touch (next-key linkage candidate sorting, predecessor reads, larger ops — its baseline is ~45% slower than unordered), so the same absolute read savings are a smaller fraction. And the retention discipline is deliberately more conservative there: only the retained key's own read is skipped; bucket-sibling reads and all delete reads remain.
```

I think we could tackle this in a follow-up?

roberto-bayardo:
> I think we could tackle this in a follow-up?

sg, please update PR description w/ the ordered results though?

roberto-bayardo:
Asked claude whether the longer lock durations in this PR might be a problem. The answer is mostly no, but it did suggest:

```
   Non-blocking observation on lock granularity: `Reader::read_many` now probes the page
  cache for the whole batch under a single `read_cached_many` read-lock hold. On the
  `get_many` load path that batch is unchunked (~32k items at the default update count,
  so a low-millisecond hold), and since the utils `RwLock` is parking_lot
  (write-preferring), one queued writer makes that hold stall every other cache reader
  too — including unrelated journals sharing the same cache. The old per-item probes
  gave writers natural interleave points.

  If you want to bound it, the clean spot is inside `read_cached_many` itself: probe in
  fixed-size windows, re-acquiring the lock per window. That caps the hold for every
  caller without touching the miss fetch, which should stay a single concurrent batch.
  (Chunking higher up, in `read_many`, would serialize miss I/O across chunks and hurt
  the cold path.) Relatedly: the load path has no bound on miss fan-out either —
  `try_join_all` over every miss — unlike the floor scan's 64-candidate rounds.
```

roberto-bayardo:
variable journal's read_many still has the per-item probe phase you just removed from fixed, do you want to optimize that too?

storage/src/index/partitioned/ordered.rs
```diff
@@ -75,6 +75,32 @@ impl<T: Translator, V: Send + Sync, const P: usize> super::super::Factory<T> for
...
+            .enumerate()
+            .map(|(key_idx, key)| {
+                let bytes = key.as_ref();
+                let mut prefix = [0u8; 16];
+                let n = bytes.len().min(16);
+                prefix[..n].copy_from_slice(&bytes[..n]);
+                (u128::from_be_bytes(prefix), key_idx)
+            })
+            .collect();
+        order.sort_unstable();
```
roberto-bayardo, line 97: I have found that unstable sort tends to be significantly slower than stable sort since our inputs tend to be "somewhat sorted" already, might be worth checking that.

runtime/src/utils/buffer/paged/append.rs
```diff
@@ -613,8 +616,12 @@ impl<B: Blob> Append<B> {
                 .expect("read_many_into buffer length overflow"),
             "read_many_into requires buf.len() == offsets.len() * item_size"
         );
+        assert!(
```
danlaine, line 619: Slightly beside the point of this PR, but perhaps we should settle on a consistent pattern for enforcing (or not) preconditions. Here, for example, we've kind of got the worst of both worlds: we incur the cost of enforcing the precondition, even on prod builds, but we don't handle them gracefully. I think there's an argument to be made we should do one of the following:

1. Enforce the precondition on all builds (including prod) but return an error instead of panicking
2. Enforce the precondition on non-prod builds only with `debug_assert`
3. Don't enforce the precondition at all

roberto-bayardo: I think (2) makes sense for "expensive" checks, but most precondition checks tend to be negligible so they seem worth enforcing rather than ignoring (which would probably lead to a more mysterious failure later).  Don't think it's worth making non-recoverable conditions errors personally.

storage/src/index/partitioned/ordered.rs
```diff
@@ -75,6 +75,32 @@ impl<T: Translator, V: Send + Sync, const P: usize> super::super::Factory<T> for
...
 impl<T: Translator, V: Send + Sync, const P: usize> UnorderedTrait for Index<T, V, P> {
     type Value = V;
+
+    fn get_many<'a, K: AsRef<[u8]>>(&'a self, keys: &[K], mut visit: impl FnMut(usize, &'a V))
+    where
+        V: 'a,
+    {
+        // Probe in key-prefix order so consecutive probes hit the same partition and descend
+        // through shared upper tree nodes within it. Sixteen bytes cover the partition prefix
+        // (P <= 2) plus the largest cap translator (8 bytes).
```
roberto-bayardo: One catch is this works only for prefix preserving transforms (which luckily are the only kinds we use I think?)

storage/src/qmdb/any/batch.rs
```diff
@@ -45,6 +45,21 @@ type DiffSlice<K, F, V> = [(K, DiffEntry<F, V>)];
 /// of a given key during ordered merkleization.
 type PrevCandidates<K, F, V> = Vec<(K, (V, Location<F>))>;
 
+/// Committed-DB resolutions retained by a batch's reads, keyed by the resolved key and carrying
+/// the location they were read from plus the variant's retained payload.
```
roberto-bayardo: "plus the variant's retained payload" -> "and its value".

storage/src/index/partitioned/ordered.rs
patrick-ogrady: Ah, interesting point.

storage/src/qmdb/any/batch.rs
```diff
@@ -181,6 +196,23 @@ where
...
+    /// `merkleize` reuses the resolution and skips the redundant journal read.
+    ///
+    /// Only committed-snapshot resolutions are retained. A retained location doubles as the
+    /// key's committed provenance (`base_old_loc`), which holds because a committed-resolved
+    /// key was in no live ancestor diff and a legal intervening commit (applying this batch's
+    /// own ancestors) can only relocate keys present in an ancestor diff. Ancestor-diff
+    /// resolutions are never retained: their op-gen position is an uncommitted location whose
+    /// committed provenance differs, and re-resolving them at merkleize is in-memory work (no
+    /// journal read is saved). The dropped-ancestor flow is covered end-to-end by
+    /// `test_unordered_fixed_retention_survives_ancestor_commit`.
```
roberto-bayardo: doesn't seem necessary to reference tests here?

storage/src/qmdb/any/batch.rs
```diff
@@ -181,6 +196,23 @@ where
 
     /// The committed DB or parent batch this batch was created from.
     base: Base<F, H::Digest, U, S>,
+
+    /// Committed-DB resolutions retained by this batch's reads (`get`/`get_many`), keyed by the
```
danlaine: I think "cached" might be easier to understand than "retained" throughout this PR

storage/src/qmdb/any/batch.rs
```diff
@@ -420,7 +452,9 @@ impl<'a, K: Ord, F: Family, V> AppliedAncestorResolver<'a, K, F, V> {
 /// and remain sequential candidates.
 ///
 /// `current::batch::next_candidate` mirrors this contract over a layered `BitmapBatch` chain;
-/// both must obey it.
+/// both must obey it. Production uses [`fill_candidates`], which produces the same sequence under
+/// a single read guard; this single-step form is retained as the test oracle.
+#[cfg(test)]
```
roberto-bayardo: if this is test only why not put it in test module?

storage/src/qmdb/any/batch.rs
```diff
@@ -440,6 +474,22 @@ fn next_candidate<F: Family, const N: usize>(
     (candidate < tip).then(|| Location::new(candidate))
 }
 
+/// Fill `out` with up to `limit` floor-raise candidates in `[floor, tip)` under a single bitmap
+/// read guard, returning the next `floor`. Produces the same sequence as repeatedly calling
+/// `next_candidate` (the test oracle above).
```
danlaine: I think we can omit this -- this is documented on `next_candidate` and it feels cleaner for a prod method to not reference a test one

```suggestion
/// read guard, returning the next `floor`.
```

storage/src/qmdb/any/batch.rs
```diff
@@ -923,25 +958,16 @@ where
...
     }
 
-    /// Batch read multiple keys.
+    /// Batch read multiple keys (mutations -> ancestor diffs -> committed DB).
     ///
-    /// Returns results in the same order as the input keys.
+    /// Returns results in the same order as the input keys, with `None` for absent or deleted
+    /// keys. Committed-DB operations resolved by the read are retained on the batch and
+    /// consumed by `merkleize`, which skips re-reading those keys. Keys answered by pending
+    /// mutations or ancestor diffs retain nothing (merkleize re-resolves them in memory).
```
roberto-bayardo: More implementation details here that don't seem necessary.

storage/src/qmdb/any/batch.rs
```diff
@@ -1035,46 +1069,96 @@ where
...
+            Vec::with_capacity(mutations.len() + retained.len() + 1);
+        let mut diff: DiffVec<K, F, V::Value> =
+            Vec::with_capacity(mutations.len() + retained.len());
         let mut active_keys_delta: isize = 0;
         let mut user_steps: u64 = 0;
 
-        // Process updates/deletes of existing keys in location order.
-        // This includes keys from both the committed snapshot and ancestor diffs.
+        // Write a user mutation at the next batch location, preserving the committed-base
+        // provenance of the operation it supersedes.
```
roberto-bayardo: not a fan of "commited-base provenance", doesn't this just mean "previous location"?

storage/src/qmdb/benches/constantinople.rs
```diff
@@ -0,0 +1,380 @@
+//! Constantinople-shape harness: load+write+merkleize at 32k updates / 1M keys.
+//!
+//! Times the full per-block state pipeline (get_many load, batch writes, merkleize, root) on the
+//! tokio runtime with EightCap, matching the production validator shape. Runs against the
+//! unordered or ordered fixed `any`/`current` DBs over `mmr`. Prints per-iteration latency with
```
roberto-bayardo: I would prefer we use mmb as our "default" merkle structure unless there's a good reason not to.

storage/src/qmdb/any/batch.rs
roberto-bayardo: +1

roberto-bayardo:
FYI I'm finding that the ordered retention of resolution data makes things slower. Claude summary:

  We measured ordered payload retention head-to-head against just letting merkleize
  re-read through the page cache (Retained = (), extraction/filter/injection deleted;
  roots byte-identical across all iterations): merkleize time was unchanged (the
  ~30k re-reads are warm-cache hits, below noise), while the load phase dropped
  ~8.4ms -> ~6.8ms from removing the per-read value/next-key clones. At the
  constantinople shape, ordered retention costs ~1.5ms/block and buys nothing
  measurable. Suggest dropping the ordered retention path (~50 lines) and keeping
  retention unordered-only, unless you're targeting cache-starved deployments where
  the deterministic skip matters.

patrick-ogrady:
> FYI I'm finding that the ordered retention of resolution data makes things slower.

Confirmed: reproduced your variant head-to-head (interleaved pinned binaries, roots byte-identical every iteration) and saw the same shape -- load dropped ~1.4-1.6ms/block at both depths while merkleize was unchanged. Removed in 19e8de3d3: ordered merkleize now takes the plain gather+read path, and since location-only caching is all that remains, `type Cached`/`fn cached()` are deleted from the `Update` trait entirely (`get_many_with_locations` returns `(value, location)`). The description's tables are regenerated against the merge-base with the removal in place (and the harness now runs over `mmb` per your other comment).

patrick-ogrady:
> Asked claude whether the longer lock durations in this PR might be a problem.

Incorporated both suggestions in b9cb405df: `read_cached_many` now probes in 1024-range windows, re-acquiring the cache lock per window (order-preserving compaction, covered by a 2500-range cross-window test), and `read_many_into`'s miss fan-out is bounded at 64 in-flight blob reads (matching the floor scan's bound).

patrick-ogrady:
> variable journal's read_many still has the per-item probe phase you just removed from fixed, do you want to optimize that too?

Deferred to a follow-up (documented under "explored but not included"): batching the offsets-journal probes measured a wash since that journal is small and dense, and the real cost (per-item data probes with varint length prefixes) needs a two-pass ranged read. The `any::unordered::variable::mmb` harness mode is included so the follow-up can be measured directly.

runtime/src/utils/buffer/paged/append.rs
patrick-ogrady: I'm more of an assert vs no enforce maxi. For easy checks, I don't actually mind having layers of asserts to ensure we don't diverge from a reasonable operating envelope.

runtime/src/utils/buffer/paged/cache.rs
```diff
@@ -247,31 +247,49 @@ impl CacheRef {
...
-        let page_cache = self.cache.read();
-        ranges.retain_mut(|(buf, logical_offset)| {
-            let mut remaining = buf.len();
-            let mut offset = *logical_offset;
-            let mut dst = 0;
-            while remaining > 0 {
-                let count = page_cache.read_at(blob_id, &mut buf[dst..], offset);
-                if count == 0 {
-                    break;
+        const WINDOW: usize = 1024;
```
patrick-ogrady: I actually don't think this is worth doing...these reads should be very fast.

runtime/src/utils/buffer/paged/append.rs
```diff
@@ -599,12 +602,18 @@ impl<B: Blob> Append<B> {
...
+    /// Returns the number of items fully served without a blob read (from the tip buffer and the
+    /// page cache). The remaining items required at least one blob read.
     pub async fn read_many_into(
         &self,
         buf: &mut [u8],
         offsets: &[u64],
         item_size: usize,
-    ) -> Result<(), Error> {
+    ) -> Result<usize, Error> {
+        if offsets.is_empty() {
```
patrick-ogrady: I think this should go back to where it was.

storage/src/index/partitioned/ordered.rs
```diff
@@ -75,6 +75,34 @@ impl<T: Translator, V: Send + Sync, const P: usize> super::super::Factory<T> for
...
+
+    fn get_many<'a, K: AsRef<[u8]>>(&'a self, keys: &[K], mut visit: impl FnMut(usize, &'a V))
+    where
+        V: 'a,
+    {
+        // Probe in key-prefix order so consecutive probes hit the same partition and descend
+        // through shared upper tree nodes within it. Sixteen bytes cover the partition prefix
+        // (P <= 2) plus the largest cap translator (8 bytes). The locality holds only for
+        // prefix-preserving translators (raw-key order then matches translated-key order); for
+        // any other translator this is just an arbitrary, still-correct probe order.
```
patrick-ogrady: Is there a way to define this as a function on the translator? it feels odd to always do this?

storage/src/index/ordered.rs
```diff
@@ -153,6 +153,25 @@ impl<T: Translator, V: Send + Sync> super::Factory<T> for Index<T, V> {
...
+    fn get_many<'a, K: AsRef<[u8]>>(&'a self, keys: &[K], mut visit: impl FnMut(usize, &'a V))
+    where
+        V: 'a,
+    {
+        // Probe in translated-key order: consecutive tree descents share upper node paths,
+        // which stay cache-resident across the batch.
+        let mut order: Vec<(T::Key, usize)> = keys
+            .iter()
+            .enumerate()
+            .map(|(key_idx, key)| (self.translator.transform(key.as_ref()), key_idx))
```
patrick-ogrady, line 179: Should we add an internal get function that can take this pre-translated key?

storage/src/journal/contiguous/metrics.rs
```diff
@@ -215,13 +215,14 @@ impl<E: RuntimeMetrics + Clock> FixedMetrics<E> {
...
-            .counter("cache_hits", "Number of fixed items read synchronously");
+        let hits = context.as_ref().counter(
+            "cache_hits",
+            "Number of fixed items served without a blob read (tip buffer or page cache)",
+        );
         let misses = context.as_ref().counter(
             "cache_misses",
-            "Number of fixed items not satisfied synchronously by read or read_many, including \
-             pruned or out-of-range try_read_sync probes that returned None",
+            "Number of fixed items requiring a blob read, including pruned or out-of-range \
```
patrick-ogrady: remove the: `, including pruned or out-of-range ..`

storage/src/journal/contiguous/metrics.rs
```diff
@@ -215,13 +215,14 @@ impl<E: RuntimeMetrics + Clock> FixedMetrics<E> {
     /// Create and register metrics for a fixed-size journal.
     pub(super) fn new(context: E) -> Self {
         let context = Arc::new(context);
-        let hits = context
-            .as_ref()
-            .counter("cache_hits", "Number of fixed items read synchronously");
+        let hits = context.as_ref().counter(
+            "cache_hits",
+            "Number of fixed items served without a blob read (tip buffer or page cache)",
```
patrick-ogrady: remove the `(tip buffer or page cache)`

storage/src/qmdb/any/operation/update/ordered.rs
```diff
@@ -42,7 +42,6 @@ impl<K: Key, V: ValueEncoding> UpdateTrait for Update<K, V> {
     type Key = K;
     type Value = V::Value;
     type ValueEncoding = V;
-
```
patrick-ogrady, line 45: retain this \n?

storage/src/qmdb/any/operation/update/unordered.rs
```diff
@@ -34,7 +34,6 @@ impl<K: Key, V: ValueEncoding> UpdateTrait for Update<K, V> {
     type Key = K;
     type Value = V::Value;
     type ValueEncoding = V;
-
```
patrick-ogrady, line 37: retain this \n?

storage/src/qmdb/any/batch.rs
```diff
@@ -181,6 +184,23 @@ where
 
     /// The committed DB or parent batch this batch was created from.
     base: Base<F, H::Digest, U, S>,
+
+    /// Committed-DB locations cached by this batch's reads (`get`/`get_many`), keyed by the
+    /// key they matched. Consumed by unordered `merkleize`: when a mutation key is present
```
patrick-ogrady: should we indicate somehow that we are using it in ordered to avoid storing useless info?

storage/src/qmdb/any/db.rs
```diff
@@ -225,6 +226,25 @@ where
...
+
+    /// Batch read multiple keys, returning per key the value and the committed location it was
+    /// read from (or `None` for absent keys).
+    ///
+    /// Identical resolution to [`Self::get_many`] but additionally returns the snapshot
+    /// location, letting batch reads cache it for reuse at merkleize.
+    ///
+    /// Results are returned in the same order as the input keys.
+    #[allow(clippy::type_complexity)]
+    pub(crate) async fn get_many_with_locations(
```
patrick-ogrady, line 239: just keep inline above?

storage/src/qmdb/benches/constantinople.rs
```diff
@@ -0,0 +1,384 @@
...
+    commonware_storage::qmdb::any::batch::MerkleizedBatch<
+        mmb::Family,
+        Digest,
+        commonware_storage::qmdb::any::unordered::variable::Update<Digest, Digest>,
+        Rayon,
+    >,
+>;
+
+const PAGE_SIZE: NonZeroU16 = NZU16!(4096);
+// 512MB page cache, the production validator config (active window fully warm).
```
patrick-ogrady: remove this comment

storage/src/qmdb/benches/constantinople.rs
```diff
@@ -0,0 +1,384 @@
...
+>;
+type AnyMerkleized = std::sync::Arc<
+    commonware_storage::qmdb::any::batch::MerkleizedBatch<
+        mmb::Family,
+        Digest,
+        commonware_storage::qmdb::any::unordered::fixed::Update<Digest, Digest>,
+        Rayon,
+    >,
+>;
+const CHUNK_SIZE: usize = 32;
```
patrick-ogrady: move to top of file?

patrick-ogrady (commented):
final nits

storage/src/index/partitioned/ordered.rs
patrick-ogrady: Went one better: both ordered indexes now translate each key once up front, sort by (partition, translated key), and probe with the pre-translated key via an internal `get_translated` -- the 16-byte prefix hack (and its prefix-preserving caveat) is gone entirely.

storage/src/qmdb/any/batch.rs
patrick-ogrady: Added `const CACHES_READS: bool` to the (sealed) `Update` trait: unordered is true, ordered is false, and `get_many` only populates the cache when set, so ordered batches store nothing.

PR #4008, opened 2026-06-10
[marshal] Fix Buffer Delivery

### LINKED ISSUE #3995: [consensus/marshal][low] Buffer waiter floor anchor does not resume dispatch
## Context

Marshal delivers finalized blocks to the application in height order. It can also advance its processed floor from a verified finalization, for example after state sync or a local `SetFloor` update. If the floor anchor block is not immediately available, Marshal records a pending floor and blocks application dispatch until that anchor is found and installed.

Missing blocks can arrive through multiple paths. Resolver, notarization, and finalization paths can process a block as a floor anchor. Separately, Marshal also has block subscriptions backed by the broadcast buffer, and the actor handles  completed buffer waiters in its select loop.

 ## Claim

During normal local operation, an authorized `SetFloor` can wait for an anchor block that later arrives through an existing buffer subscription waiter. That waiter path notifies block subscribers, but it does not run the floor-anchor installation path, so the pending floor remains set and application dispatch continues to be blocked.

## Impact

Marshal can remain locally stuck behind a floor transition even though the needed anchor block is already available locally and has been delivered to subscribers. This is a liveness/recovery-delay issue for ordered application delivery. It is not a proven safety break, and later resolver, notarization, or finalization activity for the same block can recover progress.

## Root Cause

The completed-buffer-waiter select-loop arm treats successful waiter results only as subscription notifications. It does not check whether the returned block matches the pending floor anchor and therefore does not call the code path that stores the floor finalization, clears the pending floor, syncs metadata, prunes old state, and restarts dispatch.

## Code

  - https://github.com/commonwarexyz/monorepo/blob/9f0e59b910df2adbb0334684b1ac210ab28a947c/consensus/src/marshal/core/actor.rs#L395-L398
    Completed buffer waiters only call `block_subscriptions.notify(&block)`.

  - https://github.com/commonwarexyz/monorepo/blob/9f0e59b910df2adbb0334684b1ac210ab28a947c/consensus/src/marshal/core/actor.rs#L1040-L1051
    `install_floor` records a pending anchor and starts a resolver fetch when the anchor is not found locally.

  - https://github.com/commonwarexyz/monorepo/blob/9f0e59b910df2adbb0334684b1ac210ab28a947c/consensus/src/marshal/core/actor.rs#L1054-L1162
    `apply_floor_anchor` is the path that consumes the pending floor, stores the anchor/finalization, syncs metadata, prunes, and restarts dispatch.

  - https://github.com/commonwarexyz/monorepo/blob/9f0e59b910df2adbb0334684b1ac210ab28a947c/consensus/src/marshal/core/actor.rs#L1571-L1578
    `try_dispatch_blocks` returns early while a floor anchor is pending.

  - https://github.com/commonwarexyz/monorepo/blob/9f0e59b910df2adbb0334684b1ac210ab28a947c/consensus/src/marshal/core/subscriptions.rs#L69-L88
    Buffer-backed subscriptions complete as block waiter results, not as resolver or mailbox deliveries.

patrick-ogrady:
Resolves: #3995


consensus/src/marshal/core/mailbox.rs
```diff
@@ -666,6 +670,10 @@ impl<S: Scheme, V: Variant> Mailbox<S, V> {
     ///
     /// The `fallback` parameter controls whether marshal also asks peers for the missing block.
     ///
+    /// Delivery makes no durability promise. A delivered block may not have been persisted by
```
patrick-ogrady, line 673: This is not a change in behavior (may have been notified by buffer)...it was being unnecessarily (partially) enforced and that is no longer the case (which removes a ton of clones).

PR #4010, opened 2026-06-10
[runtime/iobuf/freelist] use thread probe on creation

andresilva:
This PR updates lazy freelist creation to use the same striped per-thread probe order as `take` and `put`, instead of assigning slots from a single sequential counter. It now uses `created` as a relaxed capacity-admission counter, then claims the actual slot id through a relaxed `reservation` bitmap RMW. This preserves probe affinity: a buffer created by a thread is more likely to return to the global word that same thread will probe first later.


bajpainaman:
PR at 4am in the morning is crazy 💀

patrick-ogrady (dismissed):

patrick-ogrady (approved):

roberto-bayardo:
We test-merged this PR onto #4449 (freelist scaling fixes: hierarchical availability summary, TLS cap, blocking-pool parallelism) to check how the two interact. Findings:

**Mergeability.** `freelist.rs` conflicts textually (fields, constructor, tests) but the designs compose cleanly: the reservation bitmap never touches the free bitmap, constructor-exclusive prefill still seeds #4449's summary through `put_batch`, and the creation permit keeps at-capacity misses O(1). One real interaction: probe-affine creation changes slot placement, so one of #4449's tests that assumed sequential slot ids had to become creation-order-agnostic.

**The two PRs are strongly complementary.** On #4449 alone, the availability summary costs ~8ns per empty-word transition, and with sequential slot placement a worker's batch scatters across N striped words = N transitions per batch put. This PR places a worker's batch in its home word, collapsing that to one transition per batch (EPYC 9354P, `iobuf::freelist` at 8 threads/lockstep; main = 305ead4a0):

| case | main | #4449 alone | #4449 + this PR |
|---|---|---|---|
| 65536 slots / batch=8 | 454ns | 710ns | **468ns** |
| 65536 slots / batch=32 | 571ns | 821ns | **575ns** |
| `iobuf::reuse` | 16.7µs | 16.7µs | **4.37µs (3.8x)** |

The reuse-affinity win survives fully on top of #4449's scaling fixes, and it repairs the one regression #4449 leaves in batch churn workloads.

**One blocker to flag before large classes are safe here**: the lazy-growth reservation scan restarts from the thread's home word on every creation, so growing a class to capacity is quadratic in bitmap words. For the 8.4M-slot configuration that motivated #4449, that is ~5.5e11 word probes single-threaded (64 x W(W+1)/2 for W = 131,072) — the same order of pathology #4449 fixes on the take side. The current benchmarks create at most a few hundred slots, so this path is not exercised. A per-thread reservation cursor (resume scanning where the last creation left off instead of from home) would make total growth O(W) while preserving home affinity for the early creations, which are the ones that matter for placement.

Suggested order: land #4449, rebase this on top (we have the conflict resolution if useful), bound the reservation scan, then land — the combined numbers above are the payoff.

PR #4011, opened 2026-06-10
[stream] avoid ciphertext copy on recv

andresilva:
`Receiver::recv` always allocated a pooled buffer and copied the received ciphertext into it before decrypting in place. The frame returned by `recv_frame` is typically a single, uniquely-owned pooled buffer (both the tokio and io_uring network backends allocate a fresh pooled buffer per read), so the copy is avoidable. `Receiver::recv` now recovers mutable ownership of the frame via `IoBufs::try_into_single()` + `IoBuf::try_into_mut()` and decrypts it directly, saving one pool allocation and one full-message memcpy per received message.

### DIFF
--- a/stream/src/encrypted.rs
+++ b/stream/src/encrypted.rs
@@ -587,6 +587,33 @@ mod test {
         }
     }
 
+    /// Wraps a stream to return each read as a fresh, uniquely-owned pooled
+    /// buffer, mirroring the production network backends.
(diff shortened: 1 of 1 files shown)


patrick-ogrady (approved):
copy is pretty fast, so we shouldn't really see too much show up from this (but I think its right)

PR #4015, opened 2026-06-10
[storage] Write large appends directly to blob

roberto-bayardo:
## Write large appends directly to the blob

  An append larger than the write buffer previously ballooned the buffer to the append's full size: the bytes were copied into the buffer, flushed, and copied again into the page cache. Now `Append::append_owned` writes whole pages of an owned buffer directly to the blob as zero-copy slices (the current partial page is topped up and flushed through the regular protected-CRC path first, and the sub-page tail stays buffered), so the write buffer never grows meaningfully beyond capacity for any caller. The fixed and variable journals pass owned `IoBuf`s down, making batched appends (e.g. qmdb commits) zero-copy end to end. Page-cache population, on-disk format, and durability semantics are unchanged and pinned by tests (including byte-identical physical output vs the buffered path).

No shared state is mutated before the last suspension point (a dropped future leaves no trace), concurrent mutable operations fail loudly instead of silently interleaving (mirroring `resize`), the buffered tail is copied so it cannot pin the caller's batch allocation, and the page cache is populated in write-buffer-sized chunks so readers are never stalled behind a multi-MB copy.

 Benchmarks vs main: merkle::flush −5–8% at n=100k (n=10k control flat: the path only engages above the write buffer); qmdb::apply_batch −6 to −12% on the journal-I/O-heavy ancestor cases (uncomm_ancestor, comm_uncomm_chain, multi_uncomm), with the in-memory direct and comm_ancestor cases flat; prove/read/small-append/variable-journal benches all flat.

### DIFF
--- a/runtime/src/utils/buffer/paged/append.rs
+++ b/runtime/src/utils/buffer/paged/append.rs
@@ -17,9 +17,10 @@
 //! # Blob Semantics
 //!
 //! [Append] owns the physical page layout, read cache, and durability bookkeeping for the wrapped
-//! [Blob]. Cloned [Append] handles share that state and are safe to use concurrently. Raw [Blob]
-//! handles cloned before wrapping operate on physical bytes, including checksum records, rather
-//! than [Append]'s logical view, and they do not observe buffered data until it is flushed.
+//! [Blob]. Cloned [Append] handles share that state; reads may run concurrently with anything,
+//! but mutable operations (append, resize) must be serialized by the caller. Raw [Blob] handles
+//! cloned before wrapping operate on physical bytes, including checksum records, rather than
+//! [Append]'s logical view, and they do not observe buffered data until it is flushed.
 //!
 //! Raw [Blob] handles must not be used to write, resize, or otherwise mutate the blob while an
 //! [Append] exists. Those mutations bypass the buffer and page cache, can invalidate checksum
@@ -325,11 +326,6 @@ impl<B: Blob> Append<B> {
     /// flush would), and the remaining sub-page suffix is buffered.
     ///
     /// Like [Self::append], the write is not durable until [Self::sync] is called.
-    ///
-    /// # Warning
-    ///
-    /// Concurrent mutable operations (other appends, resize) are not supported while a `buf`
-    /// too large to buffer is being written and may panic.
     pub async fn append_owned(&self, buf: IoBuf) -> Result<(), Error> {
         let logical_page_size = self.cache_ref.page_size() as usize;
         let mut buf_guard = self.buffer.write().await;
--- a/runtime/src/utils/buffer/paged/append.rs
+++ b/runtime/src/utils/buffer/paged/append.rs
@@ -147,18 +147,25 @@ pub struct Append<B: Blob> {
     buffer: Arc<AsyncRwLock<Buffer>>,
 }
 
-/// Returns the capacity with a floor applied to ensure it can hold at least one full page of new
-/// data even when caching a nearly-full page of already written data.
-fn capacity_with_floor(capacity: usize, page_size: u64) -> usize {
-    let floor = page_size as usize * 2;
-    if capacity < floor {
+/// Adjusts a requested write-buffer `capacity` upward to the value the buffer actually uses,
+/// applying two upward adjustments:
+///
+/// - Rounds up to a whole multiple of `page_size`, so the buffer always holds an exact number of
+///   pages. Callers can then drain and bulk-cache full pages without re-rounding the capacity.
+/// - Raises the result to a floor of two pages, so the buffer can hold at least one full page of
+///   new data even while caching a nearly-full page of already written data.
+fn adjusted_capacity(capacity: usize, page_size: u64) -> usize {
+    let page_size = page_size as usize;
+    let rounded = capacity.div_ceil(page_size) * page_size;
+    let floor = page_size * 2;
+    if rounded < floor {
         warn!(
             floor,
             "requested buffer capacity is too low, increasing it to floor"
         );
         floor
     } else {
-        capacity
+        rounded
     }
 }
 
@@ -186,7 +193,7 @@ impl<B: Blob> Append<B> {
             blob.sync().await?;
         }
 
-        let capacity = capacity_with_floor(capacity, cache_ref.page_size());
+        let capacity = adjusted_capacity(capacity, cache_ref.page_size());
         let needs_sync = !invalid_data_found; // ensure pending writes on the wrapped blob are synced
 
         let (blob_state, partial_data) = match partial_page_state {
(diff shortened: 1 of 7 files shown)


runtime/src/utils/buffer/paged/append.rs
```diff
@@ -293,15 +293,134 @@ impl<B: Blob> Append<B> {
     }
 
     /// Append all bytes in `buf` to the tip of the blob.
+    ///
+    /// A `buf` too large to fit in the write buffer is copied once into an owned buffer and
+    /// written through [Self::append_owned]'s direct path, so the write buffer never grows
+    /// meaningfully beyond its capacity no matter how large the append.
     pub async fn append(&self, buf: &[u8]) -> Result<(), Error> {
```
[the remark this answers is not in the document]
roberto-bayardo: this is just an invariant check which should never trigger unless something is very wrong.

storage/src/journal/segmented/fixed.rs
```diff
@@ -123,15 +123,13 @@ impl<E: Storage + Metrics, A: CodecFixedShared> Journal<E, A> {
...
-        }
-        let position = size / Self::CHUNK_SIZE_U64;
-
         // Encode the item
         let buf = item.encode_mut();
-        blob.append(&buf).await?;
+        let offset = blob.append(&buf).await?;
+        if !offset.is_multiple_of(Self::CHUNK_SIZE_U64) {
+            return Err(Error::InvalidBlobSize(section, offset));
+        }
```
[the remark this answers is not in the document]
roberto-bayardo: This is fine, these are all just sanity checking invariants, we don't expect them to ever trigger and if they do something is pretty wrong.

runtime/src/utils/buffer/paged/append.rs
```diff
@@ -292,17 +293,130 @@ impl<B: Blob> Append<B> {
...
-    pub async fn append(&self, buf: &[u8]) -> Result<(), Error> {
+    /// Append all bytes in `buf` to the tip of the blob, returning the logical offset at which
+    /// the first byte was written.
+    pub async fn append(&self, buf: &[u8]) -> Result<u64, Error> {
+        let logical_page_size = self.cache_ref.page_size() as usize;
         let mut buffer = self.buffer.write().await;
 
-        if !buffer.append(buf) {
-            return Ok(());
+        // Take the buffered path unless `buf` would push the buffer over capacity and at least
```
danlaine: readability nit
```suggestion
  // Bytes needed to fill current page to a page boundary (0 if already aligned).
  let fill = (logical_page_size - (buffer.len() % logical_page_size)) % logical_page_size;

  // Only bypass the write buffer when the append would overflow it and is large enough that,
  // after filling the current page, at least one whole page can go straight to the blob.
  let overflows_capacity = buffer.len() + buf.len() > buffer.capacity;
  let has_full_page_after_fill = buf.len() >= fill + logical_page_size;
  let use_direct_path = overflows_capacity && has_full_page_after_fill;

  if !use_direct_path {
      // Buffered path: stage the bytes, flushing full pages only once the buffer is over capacity.
      let offset = buffer.size();
      if buffer.append(buf) {
          self.flush_internal(buffer, false, false).await?;
      }
      return Ok(offset);
  }

  // Direct path: copy once into an owned buffer and let append_owned write whole pages directly.
  drop(buffer);
  self.append_owned(IoBuf::copy_from_slice(buf)).await
```

runtime/src/utils/buffer/paged/append.rs
```diff
@@ -292,17 +293,130 @@ impl<B: Blob> Append<B> {
...
+    /// Append all bytes in `buf` to the tip of the blob, writing whole pages directly to the
+    /// underlying blob instead of staging them in the write buffer when `buf` is large.
+    ///
+    /// Behaves like [Self::append], but an owned `buf` too large to buffer additionally avoids
+    /// the copy into the write buffer: bytes needed to complete the current partial page are
+    /// buffered and flushed through the regular path, the whole pages that follow are written to
+    /// the blob as zero-copy slices of `buf` (populating the page cache, exactly as a buffered
+    /// flush would), and the remaining sub-page suffix is buffered.
+    ///
+    /// Like [Self::append], the write is not durable until [Self::sync] is called.
```
danlaine: I think the comments inside the function sufficiently document everything else
```suggestion
    /// Like [Self::append], but takes an owned `buf`, avoiding the copy into the write buffer for
    /// large inputs.
```

runtime/src/utils/buffer/paged/append.rs
```diff
@@ -292,17 +293,130 @@ impl<B: Blob> Append<B> {
...
+    pub async fn append_owned(&self, buf: IoBuf) -> Result<u64, Error> {
+        let logical_page_size = self.cache_ref.page_size() as usize;
+        let mut buf_guard = self.buffer.write().await;
+        let offset = buf_guard.size();
+
+        // Take the buffered path unless `buf` would push the buffer over capacity and at least
+        // one full page can be written directly from it.
+        let fill = (logical_page_size - (buf_guard.len() % logical_page_size)) % logical_page_size;
```
danlaine: Perhaps we could also use the var naming suggestions I mentioned above here?

storage/src/journal/segmented/fixed.rs
```diff
@@ -144,16 +142,13 @@ impl<E: Storage + Metrics, A: CodecFixedShared> Journal<E, A> {
     /// # Panics
     ///
     /// Panics if `buf` is empty or not a multiple of [Self::CHUNK_SIZE].
-    pub(crate) async fn append_raw(&mut self, section: u64, buf: &[u8]) -> Result<(), Error> {
+    pub(crate) async fn append_raw(&mut self, section: u64, buf: IoBuf) -> Result<(), Error> {
         assert!(!buf.is_empty());
         assert!(buf.len().is_multiple_of(Self::CHUNK_SIZE));
+        let count = buf.len() / Self::CHUNK_SIZE;
```
danlaine, line 148: nit: declare this in the `trace!` like before?

runtime/src/utils/buffer/paged/append.rs
```diff
@@ -292,17 +293,130 @@ impl<B: Blob> Append<B> {
...
+        // while we're writing to it. Everything below up to the write mutates shared state
+        // without an intervening await, so a future dropped before the write leaves no trace;
+        // one dropped during the write is covered by the fatal-failure convention.
+        let mut blob_state = self.blob_state.write().await;
+        debug_assert!(
+            blob_state.partial_page_state.is_none(),
+            "an empty tip implies no partial page state"
+        );
+
+        // Cache the bulk pages before `replace` publishes the new size and the tip lock is
```
danlaine:
```suggestion
        // Cache the pages before `replace` publishes the new size and the tip lock is
```

runtime/src/utils/buffer/paged/append.rs
```diff
@@ -891,6 +975,56 @@ impl<B: Blob> Append<B> {
         (write_buffer, Some(crc_record))
     }
 
+    /// Append the physical-page representation of `data` to `write_buffer`: payload bytes as
+    /// zero-copy slices of `data`, interleaved with newly built CRC records.
+    ///
+    /// `data` must contain a non-zero whole number of logical pages. If `old_crc_record` is
+    /// present, the first page's CRC record preserves the old CRC in its original slot.
+    fn physical_full_pages(
```
danlaine: nit: the existing name makes it sound like it should return the number of physical pages that exist
```suggestion
    fn append_full_pages(
```

runtime/src/utils/buffer/paged/append.rs
```diff
@@ -891,6 +975,56 @@ impl<B: Blob> Append<B> {
         (write_buffer, Some(crc_record))
     }
 
+    /// Append the physical-page representation of `data` to `write_buffer`: payload bytes as
+    /// zero-copy slices of `data`, interleaved with newly built CRC records.
+    ///
+    /// `data` must contain a non-zero whole number of logical pages. If `old_crc_record` is
+    /// present, the first page's CRC record preserves the old CRC in its original slot.
```
danlaine:
```suggestion
    /// Appends each page of `data` to `write_buffer` in on-disk format: its payload (a zero-copy slice
    /// of `data`) followed by a CRC record.
    ///
    /// `data.len()` must be a non-zero multiple of the page size. When `old_crc_record` is present,
  /// the first page's record preserves the old CRC in its original slot.
```

storage/src/journal/segmented/fixed.rs
roberto-bayardo: need to capture count here since we lose buf ownership.

runtime/src/utils/buffer/paged/append.rs
roberto-bayardo: done

runtime/src/utils/buffer/paged/append.rs
roberto-bayardo: done

runtime/src/utils/buffer/paged/append.rs
roberto-bayardo: Flipped the order (makes more sense to treat the direct path as the exception one) and used your variable names.

runtime/src/utils/buffer/paged/append.rs
roberto-bayardo: done

storage/src/journal/segmented/fixed.rs
```diff
@@ -123,15 +123,13 @@ impl<E: Storage + Metrics, A: CodecFixedShared> Journal<E, A> {
...
-        let position = size / Self::CHUNK_SIZE_U64;
-
         // Encode the item
         let buf = item.encode_mut();
-        blob.append(&buf).await?;
+        let offset = blob.append(&buf).await?;
+        if !offset.is_multiple_of(Self::CHUNK_SIZE_U64) {
+            return Err(Error::InvalidBlobSize(section, offset));
+        }
+        let position = offset / Self::CHUNK_SIZE_U64;
```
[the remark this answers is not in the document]
danlaine: ^ I think this is legit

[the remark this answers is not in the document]
roberto-bayardo: duplicate of earlier resolved comment, these are invariants we don't expect to be violated, and even if they are, the new behavior isn't "worse" than the old one. 

Claude's explanation: 
 

> 1. Invariant not expected to be violated. Every path that grows the blob writes whole CHUNK_SIZE items — append writes one, append_raw
>   asserts buf.len() % CHUNK_SIZE == 0. So the tip is always chunk-aligned under correct operation; this check is a defensive guard against
>   corruption/external mutation, not a path normal callers hit.
>   2. Not worse if violated. Both versions return the same fatal Error::InvalidBlobSize, and this codebase's documented convention is explicit:
>   errors from mutable methods leave the store possibly-inconsistent, callers must not use it afterward, and reviewers "need not comment the database
>   being in an inconsistent state after such an error." So the caller discards the journal in both cases. The extra buffered item lands on an
>   already-invalid blob (which was already chunk-misaligned), doesn't create a new failure mode, and doesn't change recoverability — on reopen the same
>   misalignment would be caught again. There's no scenario I can construct where write-then-fail is meaningfully worse than fail-without-write here.
> 
>   And a point in the change's favor that Copilot misses: the refactor is a genuine improvement, not just a neutral reordering. Using append's reported
>   offset is a cleaner contract (trust the writer to report where the data landed) and it drops a redundant blob.size().await — which was a second
>   buffer-lock acquisition right before the append lock. Re-adding the pre-check, as Copilot suggests, would undo that.

runtime/src/utils/buffer/paged/append.rs
roberto-bayardo: done

runtime/src/utils/buffer/paged/append.rs
roberto-bayardo: done

storage/src/journal/segmented/fixed.rs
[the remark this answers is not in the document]
danlaine: Even if it's no worse, does this change buy us anything? The logic went from "if invariant is violated, do nothing" to "do something, then check if the invariant was violated", which I'm not sure is an improvement.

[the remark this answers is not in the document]
roberto-bayardo: it just feels awkward to verify the invariant on blob size at that point. we don't usually validate every precondition elsewhere but it certainly doesn't hurt. regardless checking the post condition invariant on the returned offset seems more important, so if we restore it I'd rather we keep that too.

runtime/src/utils/buffer/paged/append.rs
```diff
@@ -146,19 +147,40 @@ pub struct Append<B: Blob> {
...
+/// Adjusts a requested write-buffer `capacity` upward to the value the buffer actually uses,
+/// applying two upward adjustments:
+///
+/// - Rounds up to a whole multiple of `page_size`, so the buffer always holds an exact number of
+///   pages. Callers can then drain and bulk-cache full pages without re-rounding the capacity.
+/// - Raises the result to a floor of two pages, so the buffer can hold at least one full page of
+///   new data even while caching a nearly-full page of already written data.
+fn adjusted_capacity(capacity: usize, page_size: u64) -> usize {
+    let page_size = page_size as usize;
+    let rounded = capacity.next_multiple_of(page_size);
```
danlaine, line 159: Why do we now round the capacity up to the nearest page size?

storage/src/journal/segmented/fixed.rs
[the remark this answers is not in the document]
danlaine: Fair enough, I'm OK keeping this as is

runtime/src/utils/buffer/paged/append.rs
roberto-bayardo: It made things a little easier to reason about, especially how much larger the underlying vector might grow beyond capacity. With the round-up, you over-allocate at most once, and by at most one page.

danlaine (dismissed):

runtime/src/utils/buffer/paged/append.rs
roberto-bayardo: Added a bit more detail to the too_big fn to explain how it bounds the over-allocation.

danlaine (approved):

PR #4018, opened 2026-06-10
[cryptography] upgrade sha2 to 0.11

roberto-bayardo:
Upgrades `sha2` to 0.11.0, which removed the `asm` feature and selects hardware backends automatically: SHA-NI on x86_64 (as before) and ARMv8 SHA2 intrinsics on aarch64 with runtime detection (previously the portable software implementation by default).

Per-target effect:
- **x86_64 (servers)**: no behavioral change (SHA-NI was already runtime-detected).
- **aarch64 linux (e.g. Graviton) and macos (dev machines)**: hardware SHA-256, ~4.3-5.4x faster.
- **wasm32 and other targets**: portable implementation, as before.

Criterion results on Apple M-series vs the previous default (soft) backend:

```
sha256::hash_message/msg_len=256       533.8 ns -> 99.6 ns  (-81%)
sha256::hash_message/msg_len=4096      6.90 us  -> 1.49 us  (-78%)
sha256::hash_message/msg_len=65536     108.1 us -> 24.2 us  (-78%)
sha256::hash_message/msg_len=1048576   1.72 ms  -> 361 us   (-79%)
sha256::hash_message/msg_len=16777216  27.5 ms  -> 5.77 ms  (-79%)
```


cryptography/Cargo.toml
```diff
@@ -56,6 +56,15 @@ workspace = true
...
 features = ["alloc"]
 
+[target.'cfg(all(target_arch = "aarch64", any(target_os = "macos", target_os = "linux")))'.dependencies.sha2]
+# sha2 selects its aarch64 hardware backend (ARMv8 SHA2 instructions, with runtime
+# detection) only when the "asm" feature is enabled; without it, aarch64 falls back to
+# the portable software implementation. x86_64 needs no feature: SHA-NI is runtime
+# detected by default. Gated to macos/linux because the feature pulls in `sha2-asm`,
+# whose cc-assembled sources are only exercised on those platforms.
+workspace = true
+features = ["asm"]
```
patrick-ogrady: We manually enable this on Constantinople but agree it should be automatically enabled.

roberto-bayardo: changed PR to instead update the sha2 package which now does auto-detect

Cargo.toml
```diff
@@ -184,7 +184,8 @@ rstest = "0.26.1"
 serde = "1.0.218"
 serde_json = "1.0.122"
 serde_yaml = "0.9.34"
-sha2 = { version = "0.10.8", default-features = false }
+sha2 = { version = "0.11.0", default-features = false }
+sha2-v010 = { package = "sha2", version = "0.10.9", default-features = false } # temporary: ark generics still require digest 0.10 impls
```
patrick-ogrady: `p256` still uses sha2@v0.10, so this wouldn't get removed even if we weren't using ark

cryptography/Cargo.toml
```diff
@@ -78,7 +79,6 @@ bench = false
 [features]
 default = [ "std" ]
 blake3-parallel = [ "blake3/rayon", "std" ]
-sha2-asm = [ "sha2/asm" ]
```
patrick-ogrady, line 80: I think defaulting to hardware accelerated sha256 is a big enough win that this tweak is worth it.

patrick-ogrady, line 80: Does this mean we can't activate `asm` for ark?

roberto-bayardo: Good catch — sha2 0.11 removed the `asm` feature entirely (hardware acceleration is on by default now), which left no way to enable the asm backend for the 0.10 copy. Restored it as `sha2-asm = ["sha2-v010/asm"]` in 335fba824, so the ark path (and p256's internal sha2 0.10, via feature unification) can still get hw SHA on aarch64. Goes away once ark moves to digest 0.11.

Issue #3885, opened 2026-05-26
[cryptography] Golden: avoid overhead of arkworks to bulletproofs circuit conversion


cronokirby:
There's around a 50% overhead of circuit conversion, because the format bulletproofs expects is different from that of arkworks.
By defining the circuit directly in our framework, we would avoid that overhead.

PR #4019, opened 2026-06-10
[cryptography] Create a ZK Circuit Abstraction

cronokirby:
Towards #3885.

The idea is pretty simple, you construct a circuit by manipulating the `Var` type, which supports arithmetic operations, like any `Field`. Each `Var` is bound at the type level to a specific context, so that we can turn these operations into a concrete computation graph.

### DIFF
--- a/cryptography/src/zk/bulletproofs/circuit.rs
+++ b/cryptography/src/zk/bulletproofs/circuit.rs
@@ -554,11 +554,11 @@ pub fn r1cs_to_circuit_and_witness<F: Ring + Random>(
 mod zkc {
     use crate::zk::circuit as zk;
     use commonware_math::algebra::{Field, Random, Ring};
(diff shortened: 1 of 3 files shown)


guruvamsi-policharla:
Looks good to me. Just wanted to bring up a third type of variable that we might need:

For private payments, we make use of committed input instead of public inputs. In Pari for example, the setup algorithm needs access to the "index" at which the committed input will be stored in the witness vector as this influences how we sample the CRS. I've mostly been doing this with hacky ways of extracting the index based on the order in which the variables are defined but it would be good to have a more reliable way of pulling this out. Would only need this for committed variables, but perhaps this can be deferred if it's too much overhead in this PR.

patrick-ogrady (approved):
I think there is more analysis to do here for soundness/correctness but that can be done as we get closer to a complete story.

PR #4020, opened 2026-06-11
[qmdb/any] Cache batch reads for ordered merkleize

patrick-ogrady:
## Summary

Speeds up Any/Current batch merkleization end to end: extends #4007's read caching to the ordered path, inverts how both paths consume the cache at merkleize, replaces per-key binary searches with sorted merges everywhere merkleize issues lookups from key-sorted loops, and batches the ancestor walk's journal reads. #4021 was stacked on this branch and merged into it, so this description covers the combined change (plus review follow-ups).

Measured at the Constantinople shape (1M keys, 32k updates/batch, EightCap, tokio, Linux x86_64, 4 cores), interleaved single-session A/B vs main (`bf0c21955`), 6 reps x 30 iters, pooled p50:

| variant | depth | main | this PR | delta |
|---|---|---|---|---|
| any::ordered::fixed::mmb | 0 | 60.4 | 43.3 | -28.3% |
| any::ordered::fixed::mmb | 1 | 89.0 | 61.3 | -31.2% |
| current::ordered::fixed::mmb | 1 | 100.0 | 67.8 | -32.2% |
| current::unordered::fixed::mmb | 1 | 64.6 | 55.2 | -14.5% |

The walk-related gains scale with pending-ancestor depth: an earlier session measured any::ordered depth 2 at 139.4 -> 110.4 (-20.9%) before the streamed-lookup work landed on top. The read-cache half alone (measured before #4021 merged in) also improved any::unordered (-11.8% / -8.1% at depths 0/1) via the consumption-join inversion. The speculative root was byte-identical to main on every iteration of every run (150+ runs across sessions).

Benchmarks were taken at the pre-review-feedback head; the follow-up commits only remove work on the measured paths (the mutations rebuild is skipped for read-free batches, a dead value clone and a duplicate resolver are gone) and add two O(1) ordering asserts.

Note: #4007 explored ordered payload caching and measured it as a net loss; this PR's history reproduced that result before finding the implementation details that flip it (see "History" below). The win comes from the same redundancy the unordered cache removes; the earlier loss was overhead in how the cache was carried and consumed, not in the idea.

### Ordered read cache + consumption join

- `update::Update` gains `type Cached` + `fn cached()`: the payload a batch read stashes alongside the resolved committed location. Unordered caches `()` (zero cost); ordered caches the old op's `next_key` (required for linkage, so a location-only cache could not skip the re-read). Every update kind caches unconditionally (no `CACHES_READS` const).
- Batch `get`/`get_many` cache each committed-snapshot resolution; `merkleize` consumes the cache so keys loaded and then updated skip the merkleize-time index probe and journal re-read. Deletes never consume it: a deleted key's predecessor may be a colliding key in its own snapshot bucket, so that bucket scan is needed regardless.
- The consumption join drains `mutations` in key order and probes the cache (O(1) hash probes instead of per-key BTreeMap descents), rebuilds the unconsumed remainder as a linear bulk build, and is skipped entirely when the batch did no reads. `cached` comes out key-sorted, so ordered's candidate sorts see mostly-sorted input. This inversion alone roughly tripled the win and also speeds up the pre-existing unordered cache.
- `Db::get_many_map(keys, map)` generalizes the committed read path: plain `db.get_many` maps hits straight to values and never builds a cached payload; the batch path passes a closure building `(value, location, cached)`.
- `PrevCandidates` holds `Option<value>`: cache-resolved keys store `None` (their prev-candidate value is provably never read, because the predecessor-rewrite loop skips every key in `updated`) and the rewrite loop panics if that invariant is ever violated, instead of silently emitting an op with the wrong value.

### Streamed lookups + batched reads (merged from #4021)

1. **Sorted merge in the ancestor walk.** The per-entry "already handled by this batch?" check (three binary searches over `updated`/`created`/`deleted`, each ~16 cache-missy 32-byte compares over MB-sized arrays) now advances three cursors in a linear merge. Cursors reset per ancestor; shadow-set (`seen`) semantics are unchanged.
2. **Batched ancestor walk reads.** Active entries are collected during the walk and resolved with one `read_ops` call instead of one awaited read per key (up to ~65k sequential awaits per merkleize at the Constantinople shape, depth 1). Every location resolves synchronously through the same layers as before; only the call pattern changes.
3. **Streamed sorted-key lookups** everywhere merkleize queries from a key-sorted loop (`DiffCursors` + `find_next_key_ascending`): the ordered emit loops, `gather_existing_locations`, `extract_parent_deleted_creates`, and `build_chunk_overlay` (current). Both primitives `assert!` queries are non-decreasing against the last consumed entry, panicking precisely when an out-of-order query would return a wrong result. `DiffCursors` also replaces `apply_batch`'s hand-rolled `AppliedAncestorResolver` (an identical cursor merge). The predecessor-rewrite loop keeps per-key binary searches: its query keys are unsorted and its counts are small.
4. **Harness fix (pre-existing bug).** `constantinople` only kept the latest pending ancestor alive, but the merkle batch parent link is a `Weak` ref and the API requires the whole uncommitted chain to stay alive, so depth >= 2 panicked with "peak missing" on main too. The harness now holds the full chain, which is what made depth-2 measurement possible.

## Soundness

Mirrors the unordered argument: only committed-snapshot resolutions are cached, so the cached location doubles as `base_old_loc`, and an intervening commit of this batch's own ancestors cannot move or relink a committed-resolved key (any location or linkage change puts the key in an ancestor diff, which would have answered the read instead of the snapshot). Required candidates are preserved: an updated key's successor comes from its own cached `next_key`; predecessor discovery for deletes/creates goes through `prev_translated_key` bucket reads (plus the delete's own-bucket scan), never through an update key's bucket scan. Skipping a cached key's bucket scan only drops collision-sibling candidates that are all live keys and cannot fall strictly between a key and its successor, so `find_next_key`/`find_prev_key` results are unchanged.

## History (why ordered caching initially measured as a loss)

The first attempts measured roughly neutral, apparently confirming #4007's "explored but not included" result. Three details accounted for the entire gap; each was found by drift-controlled interleaved A/B against pinned binaries:

1. **Caching the old value**: the per-read value clone plus fatter cache entries cost ~0.5-1ms/block of load for a payload that is provably never read (the predecessor-rewrite loop skips keys in `updated`). Fixed by caching only `next_key`.
2. **Double map lookup at consumption**: `mutations.get` followed by `mutations.remove` per cached key doubled the BTreeMap descents, costing ~2-4ms of merkleize. Fixed by a single `remove`.
3. **Join direction**: draining the cache and probing `mutations` paid a random-order tree descent per key and produced randomly-ordered output for the candidate sorts. Draining `mutations` and probing the cache eliminates both costs (tripled the win and sped up unordered as well).

A `next_key()`-accessor variant of the trait (no associated type) was tried and reverted: `type Cached` keeps unordered's cache entry payload-free instead of carrying a dead `Option<K>`.

## Testing

- Full `commonware-storage` suite passes (2571 default-profile tests, 356 slow-profile qmdb tests), including `test_ordered_fixed_resolved_merkleize_parity` (TwoCap collision-heavy, depths 0-2 with multi-ancestor shadow tracking, mixed updates/deletes/creates) and `test_ordered_fixed_caching_survives_ancestor_commit` (differential read-then-write vs write-only across an ancestor commit and drop between read and merkleize).
- Equivalence tests: `diff_cursors_matches_lookup_sorted` (random overlapping diffs, ascending queries vs per-key `lookup_sorted` closest-first) and `find_next_key_ascending_matches_binary_search` (random candidate sets, ascending queries vs `find_next_key`, including cyclic wrap), plus `should_panic` tests proving both non-decreasing asserts fire on out-of-order queries that would return wrong results.
- Benchmark root parity vs main on every iteration of every run, including collision-heavy and depth 1-2 chains.
- `cargo clippy -p commonware-storage --tests`, `rustfmt`, `just check-stability -p commonware-storage`, and `just unstable-public -p commonware-storage` all clean.

<div><a href="https://cursor.com/agents/bc-3c4e9e62-510e-4929-a6a0-7edb74702f40"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-3c4e9e62-510e-4929-a6a0-7edb74702f40"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/storage/src/qmdb/any/batch.rs
+++ b/storage/src/qmdb/any/batch.rs
@@ -42,8 +42,10 @@ type DiffSlice<K, F, V> = [(K, DiffEntry<F, V>)];
 /// of a given key during ordered merkleization.
 type PrevCandidates<K, F, V> = Vec<(K, (V, Location<F>))>;
 
-/// Committed-DB locations cached by a batch's reads, keyed by the resolved key.
-type ResolvedReads<F, U> = AHashMap<<U as update::Update>::Key, Location<F>>;
+/// Committed-DB locations (plus the update kind's cached payload) cached by a batch's reads,
+/// keyed by the resolved key.
+type ResolvedReads<F, U> =
+    AHashMap<<U as update::Update>::Key, (Location<F>, <U as update::Update>::Cached)>;
 
 /// What happened to a key in this batch.
 #[derive(Clone)]
@@ -182,11 +184,11 @@ where
     /// The committed DB or parent batch this batch was created from.
     base: Base<F, H::Digest, U, S>,
 
-    /// Committed-DB locations cached by this batch's reads (`get`/`get_many`), keyed by the
-    /// key they matched. Populated only when `U::CACHES_READS` and consumed by `merkleize`:
-    /// when a mutation key is present here, `merkleize` reuses the location and skips the
-    /// redundant journal read. Ordered batches cache nothing: their op generation needs each
-    /// read operation's value and next key, so merkleize re-reads ops regardless.
+    /// Committed-DB locations (plus the update kind's cached payload) cached by this batch's
+    /// reads (`get`/`get_many`), keyed by the key they matched. Populated only when
+    /// `U::CACHES_READS` and consumed by `merkleize`: when a mutation key is present here,
+    /// `merkleize` reuses the location (and, for ordered, the old value and next key) and
+    /// skips the redundant index probe and journal read.
     ///
     /// Only committed-snapshot resolutions are cached. A cached location doubles as the
     /// key's previous committed location (`base_old_loc`), which holds because a
@@ -976,15 +978,20 @@ where
         }
 
         if !db_keys.is_empty() {
-            let db_results = db.get_many_with_locations(&db_keys).await?;
+            let db_results = if U::CACHES_READS {
+                db.get_many_with_locations::<true>(&db_keys).await?
+            } else {
+                db.get_many_with_locations::<false>(&db_keys).await?
+            };
             let mut resolved = U::CACHES_READS.then(|| self.resolved.lock());
             if let Some(resolved) = resolved.as_deref_mut() {
                 resolved.reserve(db_keys.len());
             }
             for (slot, result) in db_indices.into_iter().zip(db_results) {
-                results[slot] = result.map(|(value, loc)| {
+                results[slot] = result.map(|(value, loc, cached)| {
                     if let Some(resolved) = resolved.as_deref_mut() {
-                        resolved.insert((*keys[slot]).clone(), loc);
+                        let cached = cached.expect("cached payload requested for caching reads");
+                        resolved.insert((*keys[slot]).clone(), (loc, cached));
                     }
                     value
                 });
@@ -1058,7 +1065,7 @@ where
         // below, exactly where the read path would have emitted them.
         let mut cached: Vec<(K, Location<F>, Option<V::Value>)> =
             Vec::with_capacity(resolved.len());
-        for (key, loc) in resolved {
+        for (key, (loc, ())) in resolved {
             if let Some(mutation) = mutations.remove(&key) {
                 cached.push((key, loc, mutation));
             }
@@ -1207,9 +1214,11 @@ where
 {
     /// Resolve mutations into operations, merkleize, and return an `Arc<MerkleizedBatch>`.
     ///
-    /// Reads on ordered batches cache nothing (`CACHES_READS` is false): op generation needs
-    /// each read operation's value and next key for linkage, and re-reading them through the
-    /// journal's page cache is cheaper than caching the payloads at read time.
+    /// Consumes the payloads cached by this batch's reads: keys loaded via `get`/`get_many`
+    /// before being updated skip the index probe and journal re-read their resolution would
+    /// otherwise require (the cached old value and next key feed op generation directly).
+    /// Deletes never consume the cache: a deleted key's own-bucket scan doubles as same-bucket
+    /// predecessor discovery.
     #[allow(clippy::type_complexity)]
     #[tracing::instrument(
         name = "qmdb::any::batch::merkleize",
@@ -1253,7 +1262,24 @@ where
         C: Mutable<Item = Operation<F, update::Ordered<K, V>>>,
         I: OrderedIndex<Value = Location<F>>,
     {
+        let resolved = core::mem::take(&mut *self.resolved.lock());
         let (mut mutations, m) = self.into_parts();
+
+        // Pull update mutations whose committed op this batch's reads already resolved: they
+        // skip the index probe and journal re-read, and their old op's payload feeds the
+        // candidate sets directly. Deletes never consume the cache because a deleted key's
+        // own-bucket scan doubles as same-bucket predecessor discovery.
+        let mut cached: Vec<(K, V::Value, Location<F>, V::Value, K)> =
+            Vec::with_capacity(resolved.len());
+        for (key, (loc, (old_value, old_next))) in resolved {
+            if matches!(mutations.get(&key), Some(Some(_))) {
+                let value = mutations
+                    .remove(&key)
+                    .flatten()
+                    .expect("mutation presence checked above");
+                cached.push((key, value, loc, old_value, old_next));
+            }
+        }
 
         // Resolve existing keys.
         let locations = m.gather_existing_locations(&mutations, db, true);
@@ -1298,6 +1324,14 @@ where
             } else {
                 deleted.push((key, old_loc));
             }
+        }
+
+        // Merge cache-resolved updates: their old op's next_key and (key, value, loc) feed the
+        // candidate sets exactly as the skipped journal read would have.
+        for (key, value, loc, old_value, old_next) in cached {
+            next_candidates.push(old_next);
+            prev_candidates.push((key.clone(), (old_value, loc)));
+            updated.push((key, value, loc));
         }
 
         deleted.sort_by(|a, b| a.0.cmp(&b.0));
--- a/storage/src/qmdb/any/batch.rs
+++ b/storage/src/qmdb/any/batch.rs
@@ -1266,18 +1266,17 @@ where
         let (mut mutations, m) = self.into_parts();
 
         // Pull update mutations whose committed op this batch's reads already resolved: they
-        // skip the index probe and journal re-read, and their old op's payload feeds the
+        // skip the index probe and journal re-read, and their old op's next key feeds the
         // candidate sets directly. Deletes never consume the cache because a deleted key's
         // own-bucket scan doubles as same-bucket predecessor discovery.
-        let mut cached: Vec<(K, V::Value, Location<F>, V::Value, K)> =
-            Vec::with_capacity(resolved.len());
-        for (key, (loc, (old_value, old_next))) in resolved {
+        let mut cached: Vec<(K, V::Value, Location<F>, K)> = Vec::with_capacity(resolved.len());
+        for (key, (loc, old_next)) in resolved {
             if matches!(mutations.get(&key), Some(Some(_))) {
                 let value = mutations
                     .remove(&key)
                     .flatten()
                     .expect("mutation presence checked above");
-                cached.push((key, value, loc, old_value, old_next));
+                cached.push((key, value, loc, old_next));
             }
         }
 
@@ -1327,10 +1326,13 @@ where
         }
 
         // Merge cache-resolved updates: their old op's next_key and (key, value, loc) feed the
-        // candidate sets exactly as the skipped journal read would have.
-        for (key, value, loc, old_value, old_next) in cached {
+        // candidate sets exactly as the skipped journal read would have. The prev-candidate
+        // value uses the NEW value rather than the old one: a prev candidate's value is only
+        // consumed when the predecessor-rewrite loop emits an op for it, and that loop skips
+        // every key present in `updated`, so a cached key's prev-candidate value is never read.
+        for (key, value, loc, old_next) in cached {
             next_candidates.push(old_next);
-            prev_candidates.push((key.clone(), (old_value, loc)));
+            prev_candidates.push((key.clone(), (value.clone(), loc)));
             updated.push((key, value, loc));
         }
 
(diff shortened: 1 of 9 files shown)


patrick-ogrady:
bugbot run

storage/src/qmdb/any/operation/update/mod.rs
```diff
@@ -23,8 +23,11 @@ pub trait Update: sealed::Sealed + Clone + Send + Sync {
     /// The value encoding (fixed or variable).
     type ValueEncoding: ValueEncoding<Value = Self::Value>;
 
-    /// Whether batch reads cache resolved committed locations for merkleize to consume.
-    const CACHES_READS: bool;
+    /// Payload cached alongside the resolved location of a batch read, consumed by merkleize.
+    type Cached: Send + Sync;
+
+    /// Build the cached payload from a resolved update.
+    fn cached(&self) -> Self::Cached;
```
patrick-ogrady: Move this after value

storage/src/qmdb/any/operation/update/ordered.rs
```diff
@@ -42,7 +42,11 @@ impl<K: Key, V: ValueEncoding> UpdateTrait for Update<K, V> {
     type Key = K;
     type Value = V::Value;
     type ValueEncoding = V;
-    const CACHES_READS: bool = false;
+    type Cached = K;
+
+    fn cached(&self) -> K {
```
patrick-ogrady: move this after value

storage/src/qmdb/any/operation/update/unordered.rs
```diff
@@ -34,7 +34,9 @@ impl<K: Key, V: ValueEncoding> UpdateTrait for Update<K, V> {
     type Key = K;
     type Value = V::Value;
     type ValueEncoding = V;
-    const CACHES_READS: bool = true;
+    type Cached = ();
+
+    fn cached(&self) {}
```
patrick-ogrady: move this after value

storage/src/qmdb/any/ordered/mod.rs
```diff
@@ -230,6 +230,30 @@ pub(crate) fn find_next_key<K: Ord + Clone>(key: &K, possible_next: &[K]) -> K {
         .clone()
 }
 
+/// Streaming equivalent of [`find_next_key`] for an ascending sequence of queries: `idx`
```
patrick-ogrady, line 233: Do we still need find_next_key?

storage/src/qmdb/any/batch.rs
```diff
@@ -42,8 +42,10 @@ type DiffSlice<K, F, V> = [(K, DiffEntry<F, V>)];
 /// of a given key during ordered merkleization.
 type PrevCandidates<K, F, V> = Vec<(K, (V, Location<F>))>;
 
-/// Committed-DB locations cached by a batch's reads, keyed by the resolved key.
-type ResolvedReads<F, U> = AHashMap<<U as update::Update>::Key, Location<F>>;
+/// Committed-DB locations (plus the update kind's cached payload) cached by a batch's reads,
```
patrick-ogrady: remove "(plus the update kind's cached payload)"

storage/src/qmdb/any/batch.rs
```diff
@@ -182,11 +184,10 @@ where
     /// The committed DB or parent batch this batch was created from.
     base: Base<F, H::Digest, U, S>,
 
-    /// Committed-DB locations cached by this batch's reads (`get`/`get_many`), keyed by the
-    /// key they matched. Populated only when `U::CACHES_READS` and consumed by `merkleize`:
-    /// when a mutation key is present here, `merkleize` reuses the location and skips the
-    /// redundant journal read. Ordered batches cache nothing: their op generation needs each
-    /// read operation's value and next key, so merkleize re-reads ops regardless.
+    /// Committed-DB locations (plus the update kind's cached payload) cached by this batch's
```
patrick-ogrady: remove "(plus the update kind's cached payload)"

storage/src/qmdb/any/batch.rs
```diff
@@ -291,6 +292,37 @@ where
     None
 }
 
+/// Streaming equivalent of [`resolve_in_ancestors`] for an ascending sequence of queries:
```
patrick-ogrady, line 296: still need `resolve_in_ancestors`?

storage/src/qmdb/any/batch.rs
```diff
@@ -536,7 +568,9 @@ where
             .or_else(|| reader.try_read_sync(*loc))
     }
 
-    /// Read a single operation by location.
+    /// Read a single operation by location (test oracle; production paths batch via
+    /// [`Self::read_ops`]).
+    #[cfg(test)]
```
patrick-ogrady: do we still need this?

in any case, just keep ` Read a single operation by location.`

storage/src/qmdb/any/batch.rs
```diff
@@ -654,6 +690,12 @@ where
         locations
     }
 
+    /// Build a [`DiffCursors`] over the ancestor diffs (closest-first) for streaming
+    /// resolution of an ascending key sequence.
+    fn diff_cursors(&self) -> DiffCursors<'_, U::Key, F, U::Value> {
```
patrick-ogrady: remove this helper

storage/src/qmdb/any/batch.rs
```diff
@@ -665,11 +707,11 @@ where
         if self.ancestors.is_empty() {
             return Vec::new();
         }
+        // `retain` visits keys in ascending order, so ancestor resolution streams.
```
patrick-ogrady: remove this comment

storage/src/qmdb/any/batch.rs
```diff
@@ -625,8 +659,10 @@ where
                 locations.extend(db.snapshot.get(key).copied());
             }
         } else {
+            // `mutations.keys()` ascends, so ancestor resolution streams.
```
patrick-ogrady: remove this comment

storage/src/qmdb/any/batch.rs
```diff
@@ -1207,9 +1254,11 @@ where
 {
     /// Resolve mutations into operations, merkleize, and return an `Arc<MerkleizedBatch>`.
     ///
-    /// Reads on ordered batches cache nothing (`CACHES_READS` is false): op generation needs
-    /// each read operation's value and next key for linkage, and re-reading them through the
-    /// journal's page cache is cheaper than caching the payloads at read time.
+    /// Consumes the payloads cached by this batch's reads: keys loaded via `get`/`get_many`
+    /// before being updated skip the index probe and journal re-read their resolution would
+    /// otherwise require (the cached old value and next key feed op generation directly).
+    /// Deletes never consume the cache: a deleted key's own-bucket scan doubles as same-bucket
```
patrick-ogrady: the "Deletes..." sentence is confusing. Make it clearer.

storage/src/qmdb/any/batch.rs
```diff
@@ -1370,28 +1459,32 @@ where
         };
         let mut seen: AHashSet<&K> = AHashSet::with_capacity(seen_cap);
         let mut ancestor_deleted: Vec<K> = Vec::new();
+        let mut ancestor_active: Vec<(&K, &V::Value, Location<F>)> = Vec::new();
```
patrick-ogrady, line 1429: use with_capacity here? can be no more than current keys?

storage/src/qmdb/any/batch.rs
```diff
@@ -1438,11 +1549,16 @@ where
             Vec::with_capacity(deleted.len() + updated.len() + created.len());
         let mut active_keys_delta: isize = 0;
         let mut user_steps: u64 = 0;
+        // Each emit loop below iterates a key-sorted vec, so ancestor resolution and
```
patrick-ogrady: remove this 1552-1554

storage/src/qmdb/any/batch.rs
```diff
@@ -1438,11 +1549,16 @@ where
...
+        // next-key selection advance cursors in a sorted merge (one `DiffCursors` /
+        // `find_next_key_ascending` index per loop) instead of binary-searching per key.
         // Process deletes.
```
patrick-ogrady, line 1520: add a \n before this

storage/src/qmdb/any/batch.rs
```diff
@@ -2082,6 +2202,59 @@ mod tests {
         Shared::new(bm)
     }
 
+    /// [`DiffCursors`] must resolve exactly like per-key `lookup_sorted` over the same diffs
+    /// (closest-first) for any ascending query sequence, including queries absent from every
+    /// diff and diffs with disjoint or overlapping key ranges.
+    #[test]
+    fn diff_cursors_matches_lookup_sorted() {
+        use commonware_utils::test_rng;
```
patrick-ogrady: import `use` at top of mod

patrick-ogrady (commented):
close

storage/src/qmdb/any/ordered/mod.rs
patrick-ogrady: Still needed: the predecessor-rewrite loop queries it with non-monotonic keys (streaming variant doesn't apply there), and it serves as the oracle in find_next_key_ascending's test.

storage/src/qmdb/any/batch.rs
patrick-ogrady: Still needed: floor-raise revalidation and the unordered merkleize scan iterate in location order (not key order), and the predecessor-rewrite loop queries non-monotonically. DiffCursors only replaces the ascending-key call sites.

storage/src/qmdb/any/batch.rs
patrick-ogrady: Removed read_op entirely; the read_ops assertions in the same test already cover all three sources.

storage/src/qmdb/any/operation/update/mod.rs
```diff
@@ -23,15 +23,18 @@ pub trait Update: sealed::Sealed + Clone + Send + Sync {
     /// The value encoding (fixed or variable).
     type ValueEncoding: ValueEncoding<Value = Self::Value>;
 
-    /// Whether batch reads cache resolved committed locations for merkleize to consume.
-    const CACHES_READS: bool;
```
patrick-ogrady, line 27: Thank god this is gone.

PR #4022, opened 2026-06-11
[storage/journal/contiguous] Don't time cache hits

danlaine:
Cache hits are very fast so the amount of time it takes to measure them is comparable to the thing being measured.


roberto-bayardo:
Shouldn't this be "Don't time cache hits" ?

danlaine:
> Shouldn't this be "Don't time cache hits" ?

Oops yes fixed PR name/description

roberto-bayardo (approved):

PR #4023, opened 2026-06-11
[runtime, storage] add typed decode APIs to paged Append

andresilva:
Every journal read previously copied bytes out of the page cache (or write buffer) into a temporary buffer and decoded the item from that copy, paying an allocation, one or two memcpys per item, and on the variable journal's synchronous path, two cache-lock acquisitions per item. This PR adds typed decode APIs to the paged buffer's read handles (`try_decode_sync`, `try_decode_prefix_sync`, `try_decode_sync_many` on `Writer` and `Sealed`, implemented once on their shared `View`) that decode any `commonware_codec::Read` type directly from resident page bytes under the cache read lock: zero copies and zero allocation on hits. Items within a single page (the common case) decode straight from the contiguous page slice, while items spanning pages decode through `PagedBuf`, a `bytes::Buf` over up to eight borrowed page slices, and items straddling the cached/tail boundary decode from the gathered pages with the in-memory tail slice appended (the tail is only appended after a complete gather, so discontiguous bytes can never be spliced together). The result contract is strict: `None` always means "retry via the async path" and `Some(Err(_))` is only returned when the bytes were fully available malformed, so partial residency is never masqueraded as corruption. Batched decodes every item's page slice before decoding any of them, preserving the overlapped-lookup behavior of the copy path they replace (#4144).

Fixed journals (segmented and contiguous) decode single reads in place, and batched (`get_many`, `read_many`) become two-phase: page-cache hits drain synchronously with one lock acquisition per chunk, and only the misses flow into the batched async read (still issued concurrently across blobs). The variable journals fuse their varint header parse and item decode into a single lock acquisition via a `PrefixedItem` wrapper in the shared frame module, replacing the old two-phase read that took the cache lock twice and copied twice. Compressed mode keeps the copy path since decompression must not run under the cache lock.

Benchmarks below were captured using the `cache=hot|cold` variants added to the random-read journal benches in this PR (hot fully warms the page cache before measurement, cold is capacity-constrained at ~50% of the working set).

## Benchmarks

Journal benches: baseline is main (`8c0c0df1b`) plus this PR's bench-only commit (so both sides have the `cache=` variants). 32-byte items, 5M-item journals.

| Benchmark | Baseline | This PR | Δ |
|---|---:|---:|---:|
| fixed read_many items=100 cache=hot | 18.94 µs | 6.14 µs | **−68%** |
| fixed read_many items=1000 cache=hot | 108.4 µs | 38.8 µs | **−64%** |
| fixed read_many items=1000 cache=cold | 100.9 µs | 35.7 µs | **−65%** |
| fixed read_many items=10000 cache=hot | 424.0 µs | 325.5 µs | **−23%** |
| fixed read_many items=100000 cache=hot | 3.57 ms | 3.21 ms | **−10%** |
| fixed read_many items=100000 cache=cold | 55.1 ms | 34.4 ms | **−38%** |
| fixed serial items=10000 cache=hot | 767.0 µs | 654.2 µs | **−15%** |
| fixed serial items=100000 cache=hot | 7.88 ms | 6.76 ms | **−14%** |
| variable serial items=1000 cache=hot | 120.8 µs | 89.6 µs | **−26%** |
| variable serial items=1000 cache=cold | 110.4 µs | 80.4 µs | **−27%** |
| variable serial items=100000 cache=hot | 15.59 ms | 12.54 ms | **−20%** |
| variable concurrent items=1000 cache=hot | 294.9 µs | 272.9 µs | **−7%** |
| variable concurrent items=10000 cache=hot | 2.95 ms | 2.74 ms | **−7%** |
| variable read_many items=1000 cache=hot | 120.9 µs | 101.5 µs | **−16%** |
| variable read_many items=10000 cache=hot | 797.8 µs | 651.0 µs | **−18%** |
| variable replay items=500000 buffer=1M | 21.4 ms | 22.3 ms | no change (replay untouched) |

Miss-dominated regimes (large cold reads on the variable journal) are neutral: misses flow through the identical batched async fallback either way. The `concurrent items=100000` rows are dominated by 100k-future join machinery this PR does not touch and swing +-15% run to run, so they are omitted rather than reported as signal.

## QMDB (constantinople, #4144 harness)

Neutral, as expected: QMDB's read plane already probes the page cache synchronously and only pays the async path on misses (#4144), so this PR only replaces its probe's copy-then-decode with an in-place decode of the same 32 bytes. Measured to confirm no regression: all 10 shapes from the #4144 table (1M keys, 32,768 updates, 8 threads, 200 iterations), 3 interleaved runs per shape, medians of p50.

| variant | depth | reads | read_chunks | main p50 | PR p50 | Δ |
|---|---:|---:|---:|---:|---:|---:|
| `any::unordered::fixed::mmb` | 0 | 131,072 | 4 | 8.93 ms | 8.74 ms | −2.1% |
| `current::unordered::fixed::mmb` | 0 | 131,072 | 4 | 12.35 ms | 12.62 ms | +2.2% |
| `any::unordered::fixed::mmb` | 0 | 32,768 | 1 | 6.44 ms | 6.63 ms | +3.0% |
| `any::unordered::fixed::mmb` | 0 | 131,072 | 1 | 8.83 ms | 8.83 ms | 0.0% |
| `any::unordered::variable::mmb` | 0 | 131,072 | 4 | 9.99 ms | 9.86 ms | −1.3% |
| `any::unordered::fixed::mmb` | 1 | 131,072 | 4 | 14.30 ms | 14.55 ms | +1.7% |
| `current::unordered::fixed::mmb` | 1 | 131,072 | 4 | 17.38 ms | 17.58 ms | +1.2% |
| `any::unordered::fixed::mmb` | 1 | 32,768 | 1 | 9.97 ms | 10.17 ms | +2.0% |
| `any::unordered::fixed::mmb` | 1 | 131,072 | 1 | 13.93 ms | 14.05 ms | +0.9% |
| `any::unordered::variable::mmb` | 1 | 131,072 | 4 | 15.72 ms | 15.81 ms | +0.6% |


runtime/src/utils/buffer/paged/append.rs
```diff
@@ -24,8 +24,25 @@
 //! Raw [Blob] handles must not be used to write, resize, or otherwise mutate the blob while an
 //! [Append] exists. Those mutations bypass the buffer and page cache, can invalidate checksum
 //! recovery, and are not covered by [Append]'s [`Blob::write_at_sync`] fast paths.
+//!
+//! # Typed Synchronous Decoding
+//!
+//! The `try_decode_*` methods decode codec types directly from cached page bytes (or tip-buffer
+//! bytes) without performing I/O and without copying bytes into an intermediate buffer: the
+//! decoder reads borrowed page slots in place, so only the decoded value itself is materialized.
+//! The decode runs while internal locks are held, so the `commonware_codec::Read`
```
patrick-ogrady: Can we not just get a cheap clone to avoid locking? I figured these were "Bytes" but if not, I get it.

andresilva: The approach I had taken initially was to change the cache/append APIs to expose `IoBufs`. This would have enabled zero-copy reads from the cache. But in the existing benchmarks this was always slower than the current slice-based copying API. The existing benchmarks are definitely a worst case, because they use small values (32 bytes), and just copying those bytes is always faster than the atomic bookkeeping needed for `IoBufs`. So after that I decided to experiment with this alternative API where we just decode in-place, it both avoids copying as well as acquiring the same lock multiple times).

PR #4024, opened 2026-06-11
[storage] Refactor merkleization

danlaine:


danlaine:
Not sure the additional complexity and append path overhead merit the improvement on the update-heavy path.

PR #4025, opened 2026-06-11
[storage/index] store collision chains in a side table

roberto-bayardo:
Replace the per-entry Record linked list (an inline value plus a boxed `next` pointer on every entry) with an inline value plus a separate Overflow side table (BTreeMap) holding only the rare conflicting values. The common collision-free entry loses its 8-byte chain pointer: ~32% less memory for the unordered (HashMap) index and ~30% for the ordered (BTreeMap) index.

The Cursor is rewritten to walk the inline value then the key's overflow chain, and is now entirely safe code (the previous NonNull raw-pointer chain-walking and its iterative Drop are gone). It takes the key's overflow values lazily, so find-then-update never probes the overflow map. Collision-free get/insert behavior is unchanged.

This covers the two flat indexes (unordered + ordered) plus the ordered partitioned wrapper, which share the cursor machinery in storage.rs.

## Memory (measured)

Resident heap **bytes per key** via a counting global allocator, collision-free distinct 8-byte keys inserted in **random order** (realistic for translated/hashed keys), `V = u64`, this branch vs `origin/main` (`e68fede11`). Each map entry's value shrinks from `Record<u64>` (the `u64` plus an 8-byte `Option<Box>` chain pointer = 16 B) to a bare inline `u64` (8 B), so the stored key+value pair drops from 24 to 16 bytes.

| index | main | branch | delta |
|---|---|---|---|
| unordered (`HashMap`), ~86% load | 29.13 | 19.81 | **−32.0%** |
| unordered (`HashMap`), ~48% load | 52.43 | 35.65 | **−32.0%** |
| ordered (`BTreeMap`) | 38.87 | 27.13 | **−30.2%** |

- The **−32% / −30% delta is the figure to trust**: main and branch use the same map with the same slot count at a given N, so only the per-entry size differs (24 vs 16 bytes) and the ratio is the same regardless of how full the table is.
- Absolute `HashMap` B/key depends on how full the slot array happens to be — it allocates a power-of-two array of slots and pays for all of them, so bytes/key nearly doubles right after the array resizes (half-empty) and shrinks back as it fills. Both extremes are shown: ~86% full (N=0.9M, just under the `2^20` resize) and ~48% full (N=1M, just past it). `BTreeMap` has no such resize step; its absolute reflects ~70% node fill under random insertion and is stable across all N.

## Benchmark: `constantinople`

Production-shape pipeline (1M keys, 32,768 updates/batch, `depth=0`, 15 timed iterations) on Apple Silicon, this branch vs `origin/main` (`bce243dab`). Times in ms. `load` = the `get_many` phase (the only phase this PR touches); `merk` = merkleize, which is hashing and index-independent.

**Correctness:** per-iteration Merkle roots are byte-identical to `main` for both variants — the refactor preserves exact semantics.

| variant | load (main → branch) | total p50 (main → branch) |
|---|---|---|
| `any::unordered::fixed::mmb` | 3.78 → 3.83 ms (neutral) | 20.28 → 20.00 ms (−1.4%) |
| `any::ordered::fixed::mmb` | 7.17 → 6.27 ms (**−12.6%**) | 23.12 → 22.29 ms (−3.6%) |

- The pipeline is merkleize-dominated (~65% of total time), so the index memory reduction (this PR's primary goal) only partially surfaces as latency.
- **Unordered**: neutral, within run-to-run noise — at 1M keys the load phase is small and the smaller HashMap slot doesn't move it measurably.
- **Ordered**: the denser `BTreeMap` (inline `V` instead of `Record<V>`) makes `get_many`'s tree traversal touch fewer cache lines, cutting the load phase ~13% (per-iteration distributions are non-overlapping across the two runs) and total ~3.6%.
- Single 15-iteration run per variant on one machine; deltas under ~2% are within noise.

## Index microbenchmarks (criterion)

Flat indexes only (partitioned/hashed variants and the std-`HashMap` reference benches excluded), small cache-resident sizes (10k/50k items), this branch vs `origin/main` (`bce243dab`). Negative = faster.

| operation | change |
|---|---|
| `lookup` (all translators/sizes) | **−2% to −11%** (realistic `eight_cap`: −6% to −8%) |
| `insert_and_retain` | **−37% to −43%** |
| `insert` (unordered) | −5% |
| `insert` (ordered) | neutral (±6% run-to-run) |
| `lookup_miss` (`eight_cap`, realistic) | neutral |
| `lookup_miss` (`two_cap`/`four_cap`) | +5% to +10% (noisy) |

- `lookup` is consistently faster: the `get` path now probes the overflow side table lazily (only past the inline value), so collision-free hits do strictly less work.
- `insert_and_retain` was initially **+48%** (the side-table cursor is heavier than the old inline `Record`); the follow-up commit adds a fast path that handles a key with no overflow chain directly, without building a cursor, which more than recovers it. This is the path qmdb immutable snapshot building and archive pruning use.
- `lookup_miss` is neutral for the realistic collision-free translator (`eight_cap`); the small regressions are confined to the high-collision toy translators (`two_cap`/`four_cap`) and are within run-to-run noise.
- These are single runs at sizes where the index is cache-resident, so the memory reduction does not surface as latency; deltas under ~5% are within noise.

### DIFF
--- a/storage/src/index/mod.rs
+++ b/storage/src/index/mod.rs
@@ -2425,15 +2425,7 @@ mod tests {
     /// Exercises a single translated key holding a very large overflow chain, ensuring inserts and
     /// the resulting `Vec` overflow stay correct at scale.
     fn run_index_large_collision_chain<I: Unordered<Value = u64>>(index: &mut I) {
-        cfg_if::cfg_if! {
-            if #[cfg(miri)] {
-                // miri uses a smaller chain to keep runtime reasonable while exercising the same
-                // insert and overflow paths.
-                const ITEMS: usize = 5_000;
-            } else {
-                const ITEMS: usize = 50_000;
-            }
-        }
+        const ITEMS: usize = 50_000;
         for i in 0..ITEMS {
             index.insert(b"", i as u64);
         }
--- a/storage/src/index/mod.rs
+++ b/storage/src/index/mod.rs
@@ -33,7 +33,7 @@ pub mod unordered;
 /// - Insertion via `insert()` to add new values.
 /// - Deletion via `delete()` to remove values.
 ///
-/// # Safety
+/// # Usage
 ///
 /// - Must call `next()` before `update()`, `insert()`, or `delete()` to establish a valid position.
 /// - Once `next()` returns `None`, only `insert()` can be called.
@@ -1355,6 +1355,62 @@ mod tests {
             {
                 let mut index = new_partitioned_ordered(context.child("ordered"));
                 run_index_insert_and_retain_dead_insert(&mut index);
+            }
+        });
+    }
+
+    /// Exercises `insert_and_retain` on single-value (collision-free) keys, covering each
+    /// combination of retaining the existing and new values.
+    fn run_index_insert_and_retain_single_value<I: Unordered<Value = u64>>(index: &mut I) {
+        // Retain both: the new value joins the chain after the existing one.
+        index.insert(b"both", 1u64);
+        index.insert_and_retain(b"both", 2u64, |_| true);
+        assert_eq!(index.get(b"both").copied().collect::<Vec<_>>(), vec![1, 2]);
+
+        // Retain the existing value, drop the new one: a no-op.
+        index.insert(b"keep", 1u64);
+        index.insert_and_retain(b"keep", 2u64, |v| *v == 1);
+        assert_eq!(index.get(b"keep").copied().collect::<Vec<_>>(), vec![1]);
+
+        // Drop both: the key is removed.
+        index.insert(b"drop", 1u64);
+        index.insert_and_retain(b"drop", 2u64, |_| false);
+        assert!(index.get(b"drop").next().is_none());
+
+        assert_eq!(index.keys(), 2); // "both" and "keep" remain
+        assert_eq!(index.items(), 3); // both -> [1, 2], keep -> [1]
+        assert_eq!(index.pruned(), 1); // the dropped "drop" value
+    }
+
+    #[test_traced]
+    fn test_hash_index_insert_and_retain_single_value() {
+        let runner = deterministic::Runner::default();
+        runner.start(|context| async move {
+            let mut index = new_unordered(context);
+            run_index_insert_and_retain_single_value(&mut index);
+        });
+    }
+
+    #[test_traced]
+    fn test_ordered_index_insert_and_retain_single_value() {
+        let runner = deterministic::Runner::default();
+        runner.start(|context| async move {
+            let mut index = new_ordered(context);
+            run_index_insert_and_retain_single_value(&mut index);
+        });
+    }
+
+    #[test_traced]
+    fn test_partitioned_index_insert_and_retain_single_value() {
+        let runner = deterministic::Runner::default();
+        runner.start(|context| async move {
+            {
+                let mut index = new_partitioned_unordered(context.child("unordered"));
+                run_index_insert_and_retain_single_value(&mut index);
+            }
+            {
+                let mut index = new_partitioned_ordered(context.child("ordered"));
+                run_index_insert_and_retain_single_value(&mut index);
             }
         });
     }
--- a/storage/src/index/mod.rs
+++ b/storage/src/index/mod.rs
@@ -25,7 +25,7 @@ pub mod unordered;
 /// A mutable iterator over the values associated with a translated key, allowing in-place
 /// modifications.
 ///
-/// The [Cursor] provides a way to traverse and modify the linked list of values associated with a
+/// The [Cursor] provides a way to traverse and modify the chain of values associated with a
 /// translated key by an index while maintaining its structure. It supports:
 ///
 /// - Iteration via `next()` to access values.
@@ -37,7 +37,7 @@ pub mod unordered;
 ///
 /// - Must call `next()` before `update()`, `insert()`, or `delete()` to establish a valid position.
 /// - Once `next()` returns `None`, only `insert()` can be called.
-/// - The cursor mutates the linked list in place. If the sole element is deleted, dropping the
+/// - The cursor mutates the chain of values in place. If the sole element is deleted, dropping the
 ///   cursor removes the map entry.
 ///
 /// _If you don't need advanced functionality, just use `insert()`, `insert_and_retain()`, or
@@ -53,7 +53,7 @@ pub trait Cursor: Send + Sync {
     /// If after `insert()`, the next active item is the item after the inserted item. If after
     /// `delete()`, the next active item is the item after the deleted item.
     ///
-    /// Advances through cursor states and adjusts for deletions. Returns `None` when the list is
+    /// Advances through cursor states and adjusts for deletions. Returns `None` when the chain is
     /// exhausted. It is safe to call `next()` even after it returns `None`.
     #[allow(clippy::should_implement_trait)]
     fn next(&mut self) -> Option<&Self::Value>;
@@ -61,7 +61,7 @@ pub trait Cursor: Send + Sync {
     /// Inserts a new value at the current position.
     fn insert(&mut self, value: Self::Value);
 
-    /// Deletes the current value, adjusting the list structure.
+    /// Deletes the current value, adjusting the chain structure.
     fn delete(&mut self);
 
     /// Updates the value at the current position in the iteration.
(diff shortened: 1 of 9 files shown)


roberto-bayardo:
bugbot run

storage/src/index/mod.rs
```diff
@@ -2366,11 +2422,13 @@ mod tests {
         });
     }
 
-    fn run_index_large_collision_chain_stack_overflow<I: Unordered<Value = u64>>(index: &mut I) {
+    /// Exercises a single translated key holding a very large overflow chain, ensuring inserts and
+    /// the resulting `Vec` overflow stay correct at scale.
+    fn run_index_large_collision_chain<I: Unordered<Value = u64>>(index: &mut I) {
         cfg_if::cfg_if! {
             if #[cfg(miri)] {
```
patrick-ogrady: Do we still need to run miri on index if we no longer have unsafe?

roberto-bayardo: Good call — no, we don't. The side-table rewrite left the index module entirely safe (zero `unsafe`), so miri has no UB to find there.

Removed in 25cf2df92:
- Dropped the `storage::index` step from the miri CI job (the `runtime` unsafe steps stay).
- Added `#![forbid(unsafe_code)]` to the index module so dropping the coverage is self-enforcing: if `unsafe` is reintroduced, the build fails (prompting either a rethink or re-adding miri).
- Collapsed the now-pointless miri-sized constant in this test to a plain `const ITEMS = 50_000`.

roberto-bayardo: lol obviously the bot wrote that ^^ but yes

storage/src/index/mod.rs
```diff
@@ -13,6 +13,8 @@
 //! If the `Translator` maps many keys to the same translated key, the performance of `Index` will
 //! degrade substantially (each conflicting key may contain the desired value).
 
+#![forbid(unsafe_code)]
```
roberto-bayardo: remove this

storage/src/index/mod.rs
roberto-bayardo: Update: dropped the `#![forbid(unsafe_code)]` I mentioned — it would have been the only one in the repo, and the convention here is to allow documented unsafe (`undocumented_unsafe_blocks = "deny"`) with miri scope as a hand-curated list of modules-with-unsafe. So the consistent change is simply removing `storage::index` from that list, which is what 391484514 now does (plus the test-constant cleanup).

storage/src/index/unordered.rs
```diff
@@ -136,9 +146,44 @@ impl<T: Translator, V: Send + Sync> Unordered for Index<T, V> {
     fn insert_and_retain(&mut self, key: &[u8], value: V, should_retain: impl Fn(&V) -> bool) {
         let k = self.translator.transform(key);
         match self.map.entry(k) {
-            Entry::Occupied(entry) => {
-                let mut cursor =
-                    Cursor::<'_, T::Key, V>::new(entry, &self.keys, &self.items, &self.pruned);
+            Entry::Occupied(mut entry) => {
+                // Optimized fast path for the common case of no overflow chain. The cases below
```
patrick-ogrady: I don't think ` The cases below
                // match the cursor's value ordering and metric accounting exactly.` is necessary?

storage/src/index/unordered.rs
```diff
@@ -136,9 +146,44 @@ impl<T: Translator, V: Send + Sync> Unordered for Index<T, V> {
...
+                // match the cursor's value ordering and metric accounting exactly.
+                #[allow(clippy::map_entry)] // suppress lint false positive
```
patrick-ogrady: can remove `// suppress lint false positive`

storage/src/index/storage.rs
```diff
@@ -2,31 +2,42 @@
...
+/// collisions that are rare for well-distributed translated keys.
 ///
-/// Again optimizing for the common case, we store the first value directly in the [Record] to avoid
-/// indirection (heap jumping).
-pub struct Record<V: Send + Sync> {
-    pub(super) value: V,
-    pub(super) next: Option<Box<Self>>,
+/// The logical chain of values for a key is the inline map value followed by the overflow values
+/// in reverse order (newest first).
+pub type Overflow<K, V> = BTreeMap<K, Vec<V>>;
```
patrick-ogrady: I think we only need HashMap for unordered?

storage/src/index/unordered.rs
```diff
@@ -34,13 +34,18 @@ impl<K: Send + Sync, V: Send + Sync> IndexEntry<V> for OccupiedEntry<'_, K, Reco
 }
 
 /// A [crate::index::Cursor] over the values associated with a translated key.
-pub type Cursor<'a, K, V> = CursorImpl<'a, V, OccupiedEntry<'a, K, Record<V>>>;
+pub type Cursor<'a, K, V> = CursorImpl<'a, K, V, OccupiedEntry<'a, K, V>>;
 
 /// A memory-efficient index that uses an unordered map internally to map translated keys to
 /// arbitrary values.
+///
+/// Each translated key maps directly to its most recently inserted value; conflicting values (from
```
patrick-ogrady: rephrase the ";" (the mark of claude...)

storage/src/index/storage.rs
```diff
@@ -37,216 +48,277 @@ const MUST_CALL_NEXT: &str = "must call Cursor::next()";
...
+///   may hold the index of a just-deleted slot, which is only ever stepped down from, never
+///   read).
+/// - [`State::EntryRemoved`] implies the chain was taken and is empty.
+pub struct Cursor<'a, K: Ord + Copy, V: Send + Sync, E: IndexEntry<V>> {
+    // The occupied index entry holding the inline value while the cursor exists.
     entry: Option<E>,
-    // The current position/state of the cursor, including any live pointers into the chain.
-    state: State<V>,
+    // The translated key whose values the cursor traverses.
+    key: K,
```
patrick-ogrady: I'm surprised we need to hold K on this now.

storage/src/index/storage.rs
roberto-bayardo: Correct, but allowing it will require a good deal of extra boilerplate (e.g. a shared trait) and it won't measurably move the needle if collisions are rare.  Want me to add it anyway?

storage/src/index/unordered.rs
roberto-bayardo: done

storage/src/index/unordered.rs
roberto-bayardo: done

storage/src/index/unordered.rs
roberto-bayardo: removed

storage/src/index/storage.rs
roberto-bayardo: I was able to get rid of this by adding a key getter to IndexEntry.

storage/src/index/storage.rs
roberto-bayardo: As discussed f2f we can use HashMap for both unordered and ordered, so changed to that.

storage/src/index/ordered.rs
```diff
@@ -28,6 +28,10 @@ use std::{
 
 /// Implementation of [IndexEntry] for [BTreeOccupiedEntry].
 impl<K: Ord + Send + Sync, V: Send + Sync> IndexEntry<V> for BTreeOccupiedEntry<'_, K, V> {
```
patrick-ogrady, line 30: Should we add \n between these items?

storage/src/index/storage.rs
```diff
@@ -37,6 +37,8 @@ enum Position {
 }
 
 pub trait IndexEntry<V: Send + Sync>: Send + Sync {
```
patrick-ogrady, line 48: trait should have function comments (and itself have a comment)?

storage/src/index/storage.rs
patrick-ogrady, line 158: do we need to deref with `*`? can't we just use ref here?

patrick-ogrady (commented):
small nits

storage/src/index/storage.rs
roberto-bayardo: yup fixed

storage/src/index/storage.rs
roberto-bayardo: added.

storage/src/index/ordered.rs
roberto-bayardo: done

patrick-ogrady (approved):
not sure how I didn't think of this in the first place...very nice improvement.

PR #4027, opened 2026-06-11
[deployer] Deploy tracer on the monitoring instance + Migrate services to `docker`

clabby:
## Overview

Run the `tracer` trace viewer (`ghcr.io/clabby/tracer-web`) as a Docker container on the monitoring instance, pointed at the local Tempo via `TEMPO_URL`. The container uses host networking so its :8080 listener binds directly on the instance (reachable from the deployer IP) and can query Tempo at 127.0.0.1:3200.


patrick-ogrady:
Can you ship a binary that we can cache in S3 like our other deps?

clabby:
> Can you ship a binary that we can cache in S3 like our other deps?

We could, though I don't think it's the best solution for shipping the static site + reverse proxy for Tempo. The docker image does a better job at packaging and composing the services than statically linking the static site and making a custom reverse proxy in a Rust binary or something to that effect.

deployer/src/aws/services.rs
```diff
@@ -386,88 +470,41 @@ pub(crate) fn samply_download_url(version: &str, architecture: Architecture) ->
...
 }
 
-/// Returns the download URL for adduser from Ubuntu archive (arch-independent)
-pub(crate) fn adduser_download_url(version: &str) -> String {
-    format!("{UBUNTU_ARCHIVE_X86_64}/main/a/adduser/adduser_{version}_all.deb")
-}
-
-/// Returns the download URL for fonts-dejavu-mono from Ubuntu archive (arch-independent)
-pub(crate) fn fonts_dejavu_mono_download_url(version: &str) -> String {
-    format!("{UBUNTU_ARCHIVE_X86_64}/main/f/fonts-dejavu/fonts-dejavu-mono_{version}_all.deb")
```
patrick-ogrady, line 456: 😓

deployer/src/aws/services.rs
```diff
@@ -48,90 +33,324 @@ pub const LIBJEMALLOC2_VERSION: &str = "5.3.0-2build1";
...
-pub const ADDUSER_VERSION: &str = "3.137ubuntu1";
-
-/// Version of fonts-dejavu-mono package for Ubuntu 24.04 (arch-independent)
-pub const FONTS_DEJAVU_MONO_VERSION: &str = "2.37-8";
-
-/// Version of fonts-dejavu-core package for Ubuntu 24.04 (arch-independent)
-pub const FONTS_DEJAVU_CORE_VERSION: &str = "2.37-8";
-
-/// Version of musl package for Ubuntu 24.04
-pub const MUSL_VERSION: &str = "1.2.4-2";
```
clabby, line 70: 🙌

deployer/src/aws/create.rs
```diff
@@ -28,24 +29,11 @@ use tracing::info;
...
+    docker: String,
     libjemalloc: String,
     logrotate: String,
-    fonts_dejavu_mono: String,
-    fonts_dejavu_core: String,
-    fontconfig_config: String,
-    libfontconfig: String,
-    unzip: String,
-    adduser: String,
-    musl: String,
```
clabby, line 48: glorious

deployer/src/aws/services.rs
```diff
@@ -1111,47 +1047,36 @@ sudo mv /home/ubuntu/pyroscope-agent.timer /etc/systemd/system/pyroscope-agent.t
...
 sudo systemctl daemon-reload
-sudo systemctl enable --now promtail
-sudo systemctl enable --now node_exporter
-sudo systemctl enable --now binary{pyroscope_enable}"#
+sudo systemctl enable node_exporter
+sudo systemctl enable promtail
+sudo systemctl enable binary
+sudo systemctl start node_exporter
+sudo systemctl start promtail
+sudo systemctl start binary || true{pyroscope_enable}"#
```
clabby, line 1088: This seems odd... why are we masking a startup failure for `binary` w/ `|| true`?

patrick-ogrady: good question

patrick-ogrady (dismissed):

patrick-ogrady:
For those curious, this approach doesn't actually slowdown deployment time at all (while making it possible to actually manage deps 😓 )

deployer/src/aws/services.rs
```diff
@@ -1111,47 +1047,39 @@ sudo mv /home/ubuntu/pyroscope-agent.timer /etc/systemd/system/pyroscope-agent.t
...
-sudo systemctl enable --now binary{pyroscope_enable}"#
+sudo systemctl enable node_exporter
+sudo systemctl enable promtail
+sudo systemctl enable binary
+sudo systemctl start node_exporter
+sudo systemctl start promtail
+
+# Don't abort provisioning if the caller-provided binary fails to start; the deployer
+# polls the service afterwards and reports an unhealthy binary (see poll_service_active).
+sudo systemctl start binary || true{pyroscope_enable}"#
```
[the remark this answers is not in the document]
patrick-ogrady: This is intentional. It allows an upgrade to fix the issue (rather than requiring a redeploy).

PR #4031, opened 2026-06-11
[qmdb] Parallelize merkleize internals

patrick-ogrady:
Parallelizes the serial internals of the QMDB `any` merkleize span (profiled at ~1-core utilization in the glue block-processing workload). Three changes, all root-preserving (verified byte-identical per iteration against `main` across every benchmarked config):

- **Batched floor-raise reads.** `read_ops` gains a fast path: strictly-increasing, all-committed location sets (the floor-raise candidate stream and depth-0 mutation reads) skip per-location probes and go straight to `read_many`, serving all page-cache hits under one lock acquisition per section. 32k warm floor reads drop from ~2.4ms to ~1.05ms. Notably, sharding the per-item probes across the pool instead made reads *slower* (5.6-7.8ms at 8 threads): each probe takes several atomic lock ops and 32k probes from 8 threads serialize on the page-cache lock word. Lock amortization wins; lock sharding loses.
- **Sharded floor-raise classification.** The per-candidate activity check (batch-diff binary search, ancestor resolve, snapshot probe) is a pure function of the pre-chunk state: at most one candidate per key can be active, and a move only rewrites the moved key's own diff entry to a location above the scan tip. Classification now runs across the strategy pool in contiguous chunks (gated at 1024 candidates), followed by a cheap serial apply that preserves the exact op order, location assignment, and early-exit semantics of the original loop. ~2.8-3.5ms serial to ~0.7ms at 8 threads. A sharded hash map for the diff lookup was evaluated and rejected: with classification parallel, 32k binary searches cost ~0.7ms while building a map costs more than it saves.
- **`Strategy::sort_by`.** New stable sort on the `Strategy` trait (`par_sort_by` on Rayon, with single-thread and sub-1024-element fallbacks that match upstream's `slice::sort_by` exactly). The large merkleize sorts (diff, floor diff, candidates) now use it. Stability matters: an unstable-sort variant regressed single-thread runs ~9% because std's run-detecting stable sort exploits the sorted diff prefix that pdqsort cannot.

## Results

`constantinople` harness, 32k updates / 1M keys, tokio, EightCap, mmb. Pinned binaries, alternating-order interleaved rounds (base/branch then branch/base) to cancel thermal drift, medians of per-round p50s. Roots byte-identical between binaries for every iteration of every config (36 config-rounds).

Merkleize span, `any::unordered::fixed` (ms):

| depth | threads | main | this PR | delta |
|---|---|---|---|---|
| 0 | 1 | 45.6 | 45.6 | 0% |
| 0 | 2 | 29.0 | 26.2 | -9.8% |
| 0 | 4 | 19.9 | 16.6 | -17.0% |
| 0 | 8 | 17.3 | 13.0 | **-25.0%** |
| 1 | 2 | 31.8 | 27.8 | -12.5% |
| 1 | 8 | 20.2 | 14.1 | **-30.1%** |

Full pipeline (load + write + merkleize), p50 (ms):

| db | depth | threads | main | this PR | delta |
|---|---|---|---|---|---|
| any::unordered | 0 | 1 | 53.6 | 53.6 | 0% |
| any::unordered | 0 | 2 | 37.0 | 34.1 | -7.7% |
| any::unordered | 0 | 4 | 27.9 | 24.1 | -13.6% |
| any::unordered | 0 | 8 | 25.0 | 20.9 | **-16.7%** |
| any::unordered | 1 | 2 | 42.0 | 37.5 | -10.6% |
| any::unordered | 1 | 8 | 29.9 | 24.1 | **-19.3%** |
| any::ordered | 0 | 8 | 27.1 | 24.5 | -9.4% |
| current::unordered | 0 | 8 | 28.6 | 25.9 | -9.5% |

Merkleize-span scaling from 1 to 8 threads improves from 2.64x to 3.52x. It is not linear and cannot be at this shape: the dominant parallel work is SHA-256 hashing (already strategy-parallel on `main`, memory-bound at ~4.5x on 8 threads without asm), and a ~4.5ms serial floor remains (batched read copy+decode, cached-resolution extraction, op emit, floor apply). The harness gained an optional `threads` argument for this sweep.

A new test (`parallel_floor_raise_root_matches_sequential`) pins Rayon == Sequential roots over >=1024-candidate floor raises; the rest of the suite only exercises sequential strategies.

### DIFF
--- a/parallel/src/lib.rs
+++ b/parallel/src/lib.rs
@@ -381,8 +381,6 @@ commonware_macros::stability_scope!(BETA {
 
         /// Sorts a slice with a comparator, preserving the order of equal elements.
         ///
-        /// The default implementation sorts on the current thread.
-        ///
         /// # Examples
         ///
         /// ```
@@ -396,10 +394,7 @@ commonware_macros::stability_scope!(BETA {
         fn sort_by<T, C>(&self, items: &mut [T], compare: C)
         where
             T: Send,
-            C: Fn(&T, &T) -> Ordering + Send + Sync,
-        {
-            items.sort_by(compare);
-        }
+            C: Fn(&T, &T) -> Ordering + Send + Sync;
 
         /// Return the number of threads that are available, as a hint to chunking.
         fn parallelism_hint(&self) -> usize;
@@ -460,6 +455,14 @@ commonware_macros::stability_scope!(BETA {
             RB: Send,
         {
             (a(), b())
+        }
+
+        fn sort_by<T, C>(&self, items: &mut [T], compare: C)
+        where
+            T: Send,
+            C: Fn(&T, &T) -> Ordering + Send + Sync,
+        {
+            items.sort_by(compare);
         }
 
         fn parallelism_hint(&self) -> usize {
@@ -589,9 +592,8 @@ commonware_macros::stability_scope!(BETA, cfg(feature = "std") {
             T: Send,
             C: Fn(&T, &T) -> Ordering + Send + Sync,
         {
-            // Small slices and single-threaded pools gain nothing from dispatch and pay
-            // the cross-thread handoff into the pool.
-            if items.len() < MIN_PARALLEL_SORT || self.thread_pool.current_num_threads() <= 1 {
+            // Small slices gain nothing from dispatch and pay the handoff into the pool.
+            if items.len() < MIN_PARALLEL_SORT {
                 items.sort_by(compare);
                 return;
             }
--- a/parallel/src/lib.rs
+++ b/parallel/src/lib.rs
@@ -475,7 +475,10 @@ commonware_macros::stability_scope!(BETA, cfg(feature = "std") {
     pub type ThreadPool = Arc<RThreadPool>;
 
     /// Minimum slice length before [`Rayon::sort_by`] dispatches to the thread pool.
-    const MIN_PARALLEL_SORT: usize = 1024;
+    ///
+    /// Mirrors rayon's `CHUNK_LENGTH`: `par_sort_by` sorts slices of up to one chunk
+    /// sequentially, so dispatching them pays the pool handoff for no parallelism.
+    const MIN_PARALLEL_SORT: usize = 2000;
 
     /// A parallel execution strategy backed by a rayon thread pool.
     ///
@@ -592,7 +595,10 @@ commonware_macros::stability_scope!(BETA, cfg(feature = "std") {
             T: Send,
             C: Fn(&T, &T) -> Ordering + Send + Sync,
         {
-            // Small slices gain nothing from dispatch and pay the handoff into the pool.
+            // install() from a non-worker thread injects the job and parks the caller until
+            // a pool worker runs it. rayon's small-slice sequential fallback only applies
+            // after that handoff, so small slices skip the dispatch and sort on the calling
+            // thread instead.
             if items.len() < MIN_PARALLEL_SORT {
                 items.sort_by(compare);
                 return;
--- a/parallel/src/lib.rs
+++ b/parallel/src/lib.rs
@@ -62,12 +62,13 @@
 
 commonware_macros::stability_scope!(BETA {
     use cfg_if::cfg_if;
-    use core::fmt;
+    use core::{cmp::Ordering, fmt};
 
     cfg_if! {
         if #[cfg(feature = "std")] {
             use rayon::{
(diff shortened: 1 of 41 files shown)


parallel/src/lib.rs
```diff
@@ -558,6 +584,20 @@ commonware_macros::stability_scope!(BETA, cfg(feature = "std") {
...
         }
 
+        fn sort_by<T, C>(&self, items: &mut [T], compare: C)
+        where
+            T: Send,
+            C: Fn(&T, &T) -> Ordering + Send + Sync,
+        {
+            // Small slices and single-threaded pools gain nothing from dispatch and pay
+            // the cross-thread handoff into the pool.
+            if items.len() < MIN_PARALLEL_SORT || self.thread_pool.current_num_threads() <= 1 {
```
patrick-ogrady: we don't need to do this <= 1 thing...that's automatically handled.

storage/src/qmdb/any/batch.rs
```diff
@@ -556,6 +597,17 @@ where
         batch_ops: &[Operation<F, U>],
         reader: &R,
     ) -> Result<Vec<Operation<F, U>>, crate::qmdb::Error<F>> {
+        // Strictly increasing, all-committed inputs (the floor-raise candidate stream and
+        // depth-0 mutation reads) skip the per-location probes entirely: one batched read
+        // serves all page-cache hits under a single lock acquisition per section, where
+        // per-location probes pay a cache lock acquisition each.
+        if locations.last().is_some_and(|&last| *last < self.db_size)
```
patrick-ogrady: should we always sort here?

storage/src/qmdb/any/batch.rs
```diff
@@ -3407,4 +3512,73 @@ mod tests {
...
+
+    /// Roots from a multi-thread strategy must match the sequential strategy exactly.
+    ///
+    /// Floor raises with at least `MIN_PARALLEL_ITEMS` candidates take the sharded
+    /// classification path and the pool-dispatched sorts, which the rest of the suite
+    /// (sequential strategies only) never exercises. Updating alternating halves of the
+    /// key set makes each floor raise classify a mix of superseded (batch diff) and
+    /// still-active (snapshot) candidates.
+    #[test]
+    fn parallel_floor_raise_root_matches_sequential() {
```
patrick-ogrady: not sure this is needed

storage/src/qmdb/any/batch.rs
```diff
@@ -729,51 +782,101 @@ where
...
-                    };
-                    if !active {
-                        continue;
                     }
-                    let new_loc = Location::new(self.base_size + ops.len() as u64);
-                    let value = extract_update_value(&op);
-                    ops.push(op);
+                };
+                let outcomes: Vec<Vec<FloorOutcome<F>>> =
+                    if candidates.len() >= MIN_PARALLEL_ITEMS && strategy.parallelism_hint() > 1 {
```
patrick-ogrady: I don't like this parallel switch?

storage/src/qmdb/any/batch.rs
```diff
@@ -38,6 +38,25 @@ use tracing::debug;
...
 type DiffSlice<K, F, V> = [(K, DiffEntry<F, V>)];
 
+/// Minimum number of items before per-item resolution work (synchronous reads, floor-raise
+/// classification) is sharded across the strategy pool. Below this, dispatch overhead
+/// exceeds the win.
+const MIN_PARALLEL_ITEMS: usize = 1024;
+
+/// Contiguous chunks per pool thread when sharding per-item work, leaving headroom for
+/// load imbalance between chunks.
+const CHUNKS_PER_THREAD: usize = 4;
```
patrick-ogrady: do we really need this? can we not just split up based on parallelism_hint?

parallel/src/lib.rs
```diff
@@ -378,6 +379,28 @@ commonware_macros::stability_scope!(BETA {
             RA: Send,
             RB: Send;
 
+        /// Sorts a slice with a comparator, preserving the order of equal elements.
+        ///
+        /// The default implementation sorts on the current thread.
```
patrick-ogrady: do we need a default implementation?

storage/src/qmdb/any/batch.rs
```diff
@@ -3407,4 +3468,5 @@ mod tests {
             db.destroy().await.unwrap();
         });
     }
+
```
patrick-ogrady: remove extra \n

PR #4032, opened 2026-06-12
[qmdb/compact] Add commit() for durability without full sync

roberto-bayardo:
## Summary

Compact dbs (`keyless::CompactDb`, `immutable::CompactDb`) previously only exposed `sync()`, which persists the witness journal's data and all of its metadata. This adds `commit()`, mirroring the `commit()`/`sync()` split on the non-compact dbs: it makes the appended witness durable via the contiguous journal's cheaper commit point (data + offsets sections fsynced, offsets metadata sync skipped). Committed state always survives a crash; the tradeoff is that reopen may replay the witness journal's tail during recovery.

No new durability bookkeeping is needed: the contiguous journal's recovery already treats the data journal as the source of truth and rebuilds the offsets suffix on reopen, so a lagging recovery watermark only adds (negligible, since the witness journal is small and regularly pruned) replay work — it never changes the recovered state.

## Changes

- `qmdb/compact/witness.rs`: the witness store exposes matching `commit()`/`sync()` entry points that delegate to one shared private `persist()` body — stage what must be persisted (no-op detection, corruption check, import clearing, witness building), append it, make it durable via `journal.commit()` or `journal.sync()` (a private `Durability` selector), and install it as the cached tip.
- Both compact dbs: public `commit()` alongside `sync()`, delegating to the corresponding witness store method; doc references to "persisted by sync" updated to mention both. `sync()` is documented as minimizing recovery work on reopen.

The compact-sync import path (`persist_compact_state`) still uses `sync()`.

## Testing

- `test_compact_commit_persists_across_reopen` (both dbs): two commits, drop without ever syncing, reopen recovers the tip. This exercises journal recovery since the deterministic runtime drops unsynced writes.
- `test_compact_rewind_to_committed_entry_after_reopen` (keyless): two commits, crash, reopen, rewind to the first — every committed witness is durable, not just the tip.
- `test_compact_sync_after_commit` (keyless): redundant `sync()` after `commit()` is a no-op and state still recovers on reopen.
- All 1539 `qmdb::` tests pass; storage clippy/fmt clean.

Note: `just lint` currently fails on main in `cryptography/src/bls12381/dkg/feldman_desmedt.rs` (`useless_borrows_in_formatting`, presumably from a newer clippy); unrelated to this change and not fixed here.

### DIFF
--- a/storage/src/qmdb/compact/witness.rs
+++ b/storage/src/qmdb/compact/witness.rs
@@ -8,10 +8,10 @@
 //! mismatch fails with [`Error::DataCorrupted`].
 //!
 //! Entries are strictly increasing in committed leaf count (`proof.leaves`), so a leaf count
-//! uniquely identifies a rewind or prune target. The journal `sync` after an append is the
-//! commit point: a crash before it drops the unsynced tail on reopen, recovering the previous
-//! commit. [`Store::prune`] bounds how far back [`Store::rewind`] can reach; the tip entry is
-//! never pruned.
+//! uniquely identifies a rewind or prune target. The journal `commit` or `sync` after an append
+//! is the commit point: a crash before it drops the unsynced tail on reopen, recovering the
+//! previous commit. [`Store::prune`] bounds how far back [`Store::rewind`] can reach; the tip
+//! entry is never pruned.
 
 use crate::{
     journal::contiguous::{variable, Reader as _},
@@ -112,6 +112,16 @@ impl<F: Family, D: Digest> VerifiedWitness<F, D> {
 /// The contiguous variable journal that backs a witness [`Store`].
 pub(crate) type Journal<E, F, D> = variable::Journal<E, Witness<F, D>>;
 
+/// How a persisted witness entry is made durable.
+#[derive(Clone, Copy)]
+enum Durability {
+    /// Commit the journal: appended entries survive a crash, but journal recovery may be
+    /// required on reopen.
+    Commit,
+    /// Sync the journal and all of its metadata, minimizing recovery work on reopen.
+    Sync,
+}
+
 /// A contiguous journal plus an in-memory cache of the tip witness.
 pub(crate) struct Store<E: Context, F: Family, D: Digest> {
     journal: Journal<E, F, D>,
@@ -162,12 +172,13 @@ impl<E: Context, F: Family, D: Digest> Store<E, F, D> {
         *self.tip_witness.write() = witness;
     }
 
-    /// Persist the current compact state as a new witness journal entry.
+    /// Persist the current compact state as a new witness journal entry, committing the journal
+    /// so the entry survives a crash. Journal recovery may be required on reopen.
     ///
     /// No-op if the cached witness already matches the Merkle (the witness is already durable).
     /// Otherwise appends a witness built from the unpruned Merkle, prunes the Merkle to its
     /// frontier, and refreshes the cache.
-    pub(crate) async fn persist<H, S>(
+    pub(crate) async fn commit<H, S>(
         &self,
         merkle: &compact::Merkle<F, D, S>,
         inactivity_floor_loc: Location<F>,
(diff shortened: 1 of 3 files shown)


storage/src/qmdb/compact/witness.rs
```diff
@@ -201,7 +213,7 @@ impl<E: Context, F: Family, D: Digest> Store<E, F, D> {
         if self.import_pending.load(Ordering::Relaxed) {
             self.clear_for_import().await?;
         }
-        self.append_and_sync(&verified.witness).await?;
```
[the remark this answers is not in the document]
roberto-bayardo: This is intentional — the tradeoff was made deliberately (an earlier revision tracked watermark staleness with a flag and we removed it).

The stale watermark affects reopen *work*, never the recovered *state*: recovery rewinds the offsets journal to the watermark and rebuilds everything beyond it by replaying the data journal, so durability of the committed witness is unaffected. The replay is bounded by the witness entries appended since the last append-bearing `sync()` — one small entry per commit, in a journal that `prune()` keeps short.

The docs match this contract: db-level `sync()` promises to "minimize recovery work on reopen" (not eliminate it), and `Store::sync()` documents the no-op case explicitly. The alternatives — a staleness flag whose only job is making a rare redundant `sync()` do one extra fsync, or unconditionally fsyncing metadata on every no-op `sync()` — cost more than the corner they fix.

storage/src/qmdb/immutable/compact.rs
```diff
@@ -269,7 +269,8 @@ where
     ///
     /// The caller has reconstructed the compact Merkle in memory and already authenticated the
     /// supplied witness/root pair. The import lives only in memory until the first
-    /// [`Self::sync`], which replaces the journal's contents with it. Until then, dropping the
+    /// [`Self::commit`] or [`Self::sync`], which replaces the journal's contents with it.
+    /// Until then, dropping the
```
patrick-ogrady: fix abrupt \n

storage/src/qmdb/immutable/compact.rs
```diff
@@ -503,10 +504,21 @@ where
...
-            .persist::<H, S>(&self.merkle, self.inactivity_floor_loc, || {
+            .sync::<H, S>(&self.merkle, self.inactivity_floor_loc, || {
+                Self::encode_commit_op(self.last_commit_metadata.clone(), self.inactivity_floor_loc)
+            })
+            .await
+    }
+
+    /// Durably persist the current db state to disk. This is faster than [`Self::sync`] but
+    /// reopen may need to replay the witness journal's tail to recover.
+    pub async fn commit(&self) -> Result<(), Error<F>> {
```
patrick-ogrady: commit defined before sync?

storage/src/qmdb/immutable/compact.rs
```diff
@@ -559,7 +571,8 @@ where
     ///
     /// # Errors
     ///
-    /// Fails if a compact-sync import has not yet been persisted by [`Self::sync`]. Any error
+    /// Fails if a compact-sync import has not yet been persisted by [`Self::commit`] or
+    /// [`Self::sync`]. Any error
```
patrick-ogrady: fix abrupt \n

storage/src/qmdb/keyless/compact.rs
```diff
@@ -257,7 +257,8 @@ where
     ///
     /// The caller has reconstructed the compact Merkle in memory and already authenticated the
     /// supplied witness/root pair. The import lives only in memory until the first
-    /// [`Self::sync`], which replaces the journal's contents with it. Until then, dropping the
+    /// [`Self::commit`] or [`Self::sync`], which replaces the journal's contents with it.
+    /// Until then, dropping the
```
patrick-ogrady: abrupt \n

storage/src/qmdb/keyless/compact.rs
```diff
@@ -488,10 +489,21 @@ where
...
-            .persist::<H, S>(&self.merkle, self.inactivity_floor_loc, || {
+            .sync::<H, S>(&self.merkle, self.inactivity_floor_loc, || {
+                Self::encode_commit_op(self.last_commit_metadata.clone(), self.inactivity_floor_loc)
+            })
+            .await
+    }
+
+    /// Durably persist the current db state to disk. This is faster than [`Self::sync`] but
+    /// reopen may need to replay the witness journal's tail to recover.
+    pub async fn commit(&self) -> Result<(), Error<F>> {
```
patrick-ogrady: commit before sync

storage/src/qmdb/compact/witness.rs
```diff
@@ -178,33 +179,81 @@ impl<E: Context, F: Family, D: Digest> Store<E, F, D> {
...
+        Ok(())
+    }
 
+    /// Persist the current compact state as a new witness journal entry, syncing the journal and
+    /// all of its metadata to minimize recovery work on reopen.
+    ///
+    /// No-op if the cached witness already matches the Merkle (the witness is already durable).
+    /// Otherwise appends a witness built from the unpruned Merkle, prunes the Merkle to its
+    /// frontier, and refreshes the cache.
+    pub(crate) async fn sync<H, S>(
```
patrick-ogrady, line 206: it feels like this could be rewritten to have sync call commit (and then just call jorunal.commit/sync to differ the two)?

roberto-bayardo: Done in spirit: `commit()`/`sync()` now delegate to one shared private `persist(.., durability)` body that differs only in which journal call follows the append (ec924c797).

We avoided the literal "sync calls commit, then journal.sync()" layering for two reasons:

1. **Lock gap**: `commit()` acquires and releases `sync_lock`, so the follow-up `journal.sync()` would run outside it, letting a concurrent rewind/prune interleave between the committed append and the metadata sync.
2. **Sequencing**: it would force the data fsync to fully complete before the offsets/metadata sync starts. Today `journal.sync()` is internally sequential anyway, but the data and offsets flushes have no ordering requirement between them (recovery rebuilds offsets from data regardless), so overlapping them inside `journal.sync()` is a future optimization the layering would foreclose.

storage/src/qmdb/keyless/compact.rs
roberto-bayardo: done

storage/src/qmdb/keyless/compact.rs
roberto-bayardo: fixed

storage/src/qmdb/immutable/compact.rs
roberto-bayardo: fixed

storage/src/qmdb/immutable/compact.rs
roberto-bayardo: moved

storage/src/qmdb/compact/witness.rs
roberto-bayardo: not exactly, but refactored to minimize code duplication.

roberto-bayardo: lol claude beat me to it and of course explained it better.

patrick-ogrady: nice, this ended up coming out pretty nice.

patrick-ogrady (approved):

PR #3998, opened 2026-06-09
[consensus/glue] Tracing for `Application`

clabby:
## Overview

Adds tracing spans that thread through the various `Marshaled` application wrappers, into `glue::stateful`, and finally down to the inner application.

Also adjusts the `tracing` settings in `commonware-runtime` to not create events on span close. Spans are meant to be consumed via OTEL.

PR #4034, opened 2026-06-12
[consensus/simplex] Add Tracing

patrick-ogrady:
## TODO

- [ ] ~Fix parentID for standalone trace (should be none)~ Filter across name + attribute
- [ ] ~Limit view trace to current view only (not all undecided?)~

## Summary

Adds OpenTelemetry-exportable `tracing` spans throughout `consensus/simplex`, oriented around the consensus **view** as the unit of work, and relocates the resolver's fetch -> deliver span correlation into the `commonware-resolver` crate so every consumer (Simplex and Marshal) inherits it. Follows the span conventions in `AGENTS.md` (dot-separated names, `skip_all`, `epoch`+`view` fields, spans carried across mailbox boundaries, errors recorded only at roots) and extends the application-tracing precedent from #3998.

## Span model

A single per-view root span **`simplex.voter.view`** (`parent: None`) anchors all work for a view. It is created in the voter's `Round` and closed when the view is **decided** (not at prune). The batcher does not mint its own root: it **adopts** the voter's span via a `view_span` field on the `Update` message and `Round::set_span`, so batcher aggregation nests under the same view root. Pre-announcement batcher work (before the voter announces the view) is deliberately left unparented rather than spawning a second root.

Every span carries both `epoch` and `view` (the round identity), since a view number alone is ambiguous across epochs. To carry `epoch` onto the view-only markers, `voter::Mailbox::timeout` and `resolver::Mailbox::certified` now take a `Round` instead of a `View`.

### Voter
- `simplex.voter.view` — per-view root.
- `simplex.voter.{propose,verify,certify}` — automaton boundaries (wrapped so marshal/glue spans nest under them).
- `simplex.voter.{process,notify,timeout,replay}`, `simplex.voter.journal.{sync,prune}`.
- `simplex.voter.mailbox.{proposal,recovered,resolved,timeout}` — mailbox-crossing requests carry their span and are re-entered at dequeue.

### Batcher
- `simplex.batcher.process` — per-`Update` dequeue, parented on the adopted view span.
- `simplex.batcher.verify_{notarizes,nullifies,finalizes}` — `#[instrument]`, gated so they fire ~once per view.
- `simplex.batcher.try_construct_{notarization,nullification,finalization}` — manual spans scoped to the actual aggregation (after the quorum guard) to avoid emitting empty spans on every speculative call.
- `simplex.batcher.verify_certificate`.

### Resolver correlation (relocated into `commonware-resolver`, BETA)
The fetch -> deliver span correlation now lives in the resolver crate so it is shared by all consumers rather than hand-rolled in Simplex:
- `Fetch` and `Delivery` carry a `tracing::Span`. `subscribers::Tracker` owns the span lifecycle, storing **one span per subscriber**: each retained requester carries its own fetch span through to delivery, and that span closes when the subscriber is delivered or pruned. Coalesced deliveries link every requester via `Span::follows_from` (span links).
- Simplex consumer: `simplex.resolver.{process,fetch,deliver,validate,resolve,produce}`, `simplex.resolver.mailbox.{certified,updated}`.
- Marshal consumer: `marshal.resolver.fetch` (at issuance) correlated to `marshal.resolver.handle_deliver` (at delivery) across the resolver boundary.

## Notable API changes
- `commonware-resolver` (BETA): `Fetch`/`Delivery` gain a `span` field, and `Delivery.subscribers` is now `NonEmptyVec<(S, tracing::Span)>` (the standalone `Delivery.span` field is removed). In-memory only; no wire or storage format change.
- `simplex::voter::Mailbox::timeout` and `resolver::Mailbox::certified` take `Round` instead of `View`.

## Testing
- New trace tests: simplex (view span closes at decide, automaton-boundary nesting, quorum-driven certificate construction) and resolver (`each_subscriber_keeps_its_own_fetch_span`).
- Full suites pass: resolver (124), marshal (301), simplex voter/batcher/resolver. `clippy`, nightly `fmt`, and `dylint` (incl. `span_name_style`) are clean.


consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -144,12 +144,19 @@ where
         )
     }
 
-    fn new_round(&self) -> Round<S, B, D, Re> {
+    fn new_round(&self, view: View) -> Round<S, B, D, Re> {
+        let span = info_span!(
```
patrick-ogrady: we should connect this to the root round view?

patrick-ogrady: Done in the final design. The batcher no longer mints its own `parent: None` `simplex.batcher.view` root (that span is gone). Each batcher `Round` now holds `Span::none()` until the voter announces the view via `Update.view_span`, then adopts the voter's single `simplex.voter.view` root through `Round::set_span` (doc: "Adopts the root span of the view from the voter") and drops it at decide via `close_span`. So batcher per-view work nests under the root round view, as suggested.

resolver/src/subscribers.rs
```diff
@@ -38,20 +45,34 @@ where
 
     /// Add subscribers for a key, deduplicating subscribers already retained.
     ///
+    /// The span is retained only when the key entry is created; later fetches for
+    /// the same key keep the original span.
+    ///
     /// Returns `true` if this created a new key entry.
-    pub fn insert(&mut self, key: K, subscribers: NonEmptyVec<S>) -> bool {
+    pub fn insert(&mut self, key: K, subscribers: NonEmptyVec<S>, span: tracing::Span) -> bool {
```
patrick-ogrady: Is there a way to join the span here? Would it not make sense to keep with each entry and resolve with the fetch?

patrick-ogrady: Done in 079e3fb74 (per-subscriber spans). `Tracker` now stores a span per subscriber (`BTreeMap<K, BTreeMap<S, Span>>`); `insert` keeps each subscriber's first fetch span (first-wins per subscriber, not per key). `Delivery.subscribers` is now `NonEmptyVec<(S, Span)>` and the standalone `Delivery.span` field is gone, so every retained requester carries its own fetch span to delivery; `remove_delivered`/`retain` close a subscriber's span when it is delivered or pruned. Consumers nest per requester: simplex (single `()` subscriber) uses that subscriber's span; marshal `handle_deliver` parents under the first subscriber's span and `follows_from` the rest (first use of span links in the repo). Verified: resolver 124, marshal 301, simplex resolver 13 tests pass; clippy/fmt/dylint clean.

.github/dylints/tracing_int_fields/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "tracing_int_fields"
+version = "0.0.1"
+description = "Require consensus View/Epoch tracing fields to be recorded as integers, not via Display/Debug."
```
patrick-ogrady: Is this still accurate?

consensus/src/simplex/actors/resolver/request.rs
```diff
@@ -0,0 +1,18 @@
+/// Why a resolver fetch was requested.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub(crate) enum FetchReason {
```
patrick-ogrady: Does this really need to be its own file?

consensus/src/simplex/actors/resolver/state.rs
```diff
@@ -368,99 +315,118 @@ mod tests {
         Finalization::from_finalizes(verifier, &votes, &Sequential).expect("finalization quorum")
     }
 
+    fn v(view: u64) -> View {
```
patrick-ogrady: is this necessary?

consensus/src/simplex/actors/resolver/state.rs
```diff
@@ -368,99 +315,118 @@ mod tests {
...
+
+    fn fetch(view: u64, cause: u64, reason: FetchReason) -> Effect {
+        Effect::Fetch {
+            view: v(view),
+            cause: v(cause),
+            reason,
+        }
+    }
+
+    fn remove(view: u64) -> Effect {
```
patrick-ogrady: is this necessary?

consensus/src/simplex/actors/resolver/state.rs
```diff
@@ -368,99 +315,118 @@ mod tests {
...
+        Effect::Remove(v(view))
+    }
+
+    fn retain_above(view: u64) -> Effect {
```
patrick-ogrady: is this necessary?

consensus/src/simplex/actors/voter/round.rs
```diff
@@ -175,6 +180,34 @@ impl<S: Scheme, D: Digest> Round<S, D> {
...
+    /// Returns the root span for all work attributed to this view.
+    ///
+    /// Disabled once the view is decided (see [Self::close_span]).
+    pub const fn span(&self) -> &Span {
+        &self.span
+    }
+
+    /// Opens the view's root span when the round becomes the active view.
+    pub fn open_span(&mut self) {
+        if !self.span.is_none() {
```
patrick-ogrady: Should this be an enum? we should not be able to open a span after its been closed? I think there was a place where we could do this in batcher too?

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -249,6 +250,25 @@ impl<E: Clock + CryptoRngCore + Metrics, S: Scheme<D>, L: ElectorConfig<S>, D: D
...
+        self.views
+            .get(&view)
+            .map(|round| round.span().clone())
+            .unwrap_or_else(Span::none)
+    }
+
+    /// Closes the root span of every decided view (at or below the finalized
+    /// view) so each view trace ends when the chain commits it rather than when
+    /// the round is later pruned. Idempotent; retained rounds keep serving
+    /// backfill without anchoring a trace.
```
patrick-ogrady: Remove `Idempotent; retained rounds keep serving
    /// backfill without anchoring a trace.`

consensus/src/simplex/actors/voter/ingress.rs
```diff
@@ -138,22 +206,53 @@ impl<S: Scheme, D: Digest> Mailbox<S, D> {
...
+            round,
+            reason,
+        });
     }
 
     /// Send a recovered certificate.
     pub fn recovered(&mut self, certificate: Certificate<S, D>) {
-        let _ = self.sender.enqueue(Message::Verified(certificate, false));
+        let _ = self.sender.enqueue(Message::Verified {
+            span: info_span!(
```
patrick-ogrady, line 236: provide cert type?

consensus/src/simplex/actors/voter/ingress.rs
```diff
@@ -138,22 +206,53 @@ impl<S: Scheme, D: Digest> Mailbox<S, D> {
...
+            from_resolver: false,
+        });
     }
 
     /// Send a resolved certificate.
     pub fn resolved(&mut self, certificate: Certificate<S, D>) {
-        let _ = self.sender.enqueue(Message::Verified(certificate, true));
+        let _ = self.sender.enqueue(Message::Verified {
+            span: info_span!(
+                "simplex.voter.mailbox.resolved",
```
patrick-ogrady, line 251: provide cert type?

consensus/src/simplex/actors/voter/ingress.rs
```diff
@@ -138,22 +206,53 @@ impl<S: Scheme, D: Digest> Mailbox<S, D> {
...
-    /// Signal that the current view should timeout (if not already).
-    pub fn timeout(&mut self, view: View, reason: TimeoutReason) {
-        let _ = self.sender.enqueue(Message::Timeout(view, reason));
+    /// Signal that the given round should timeout (if not already).
+    pub fn timeout(&mut self, round: Rnd, reason: TimeoutReason) {
+        let _ = self.sender.enqueue(Message::Timeout {
+            span: info_span!(
+                "simplex.voter.mailbox.timeout",
+                epoch = round.epoch().traced(),
+                view = round.view().traced()
```
patrick-ogrady: provide reason?

resolver/src/lib.rs
```diff
@@ -17,30 +18,104 @@ commonware_macros::stability_scope!(BETA {
...
-    #[derive(Clone, Debug, Eq, Ord, PartialEq, PartialOrd)]
+    #[derive(Clone, Debug)]
     pub struct Fetch<K, S = ()> {
         /// The peer-visible key.
         pub key: K,
         /// Subscriber attached to the key.
         pub subscriber: S,
+        /// Trace span carried from issuance to delivery.
+        ///
+        /// The resolver retains this span with the subscriber and re-attaches it
```
patrick-ogrady: remove this block

resolver/src/lib.rs
```diff
@@ -17,30 +18,104 @@ commonware_macros::stability_scope!(BETA {
...
+    #[derive(Clone, Debug)]
     pub struct Delivery<K, S> {
         /// The peer-visible key used to validate the response.
         pub key: K,
-        /// Subscribers that were still retained when the response arrived.
-        pub subscribers: NonEmptyVec<S>,
+        /// Subscribers that were still retained when the response arrived, each
+        /// paired with the trace span of the fetch that requested it.
+        ///
+        /// Consumers can use these spans to attribute delivery work to the
```
patrick-ogrady: remove this block

resolver/src/subscribers.rs
```diff
@@ -36,35 +37,46 @@ where
         self.entries.contains_key(key)
     }
 
-    /// Add subscribers for a key, deduplicating subscribers already retained.
+    /// Add subscribers for a key, each paired with the span of the fetch.
+    ///
+    /// A subscriber's span is retained only when the subscriber is first seen;
```
patrick-ogrady: I think we should just keep all?

patrick-ogrady (commented):
feedback

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -504,7 +542,12 @@ where
                     // timer. We check after adding because duplicate votes are rejected.
                     if Self::leader_nullified(&current, &work) {
                         current.timed_out = true;
-                        voter.timeout(current.view, TimeoutReason::LeaderNullify);
+                        let round = Rnd::new(self.epoch, current.view);
+                        let span = work
+                            .get(&current.view)
+                            .map(|round| round.span())
+                            .unwrap_or_else(Span::none);
+                        span.in_scope(|| voter.timeout(round, TimeoutReason::LeaderNullify));
```
patrick-ogrady: Why do we uniquely do span.in_scope here?

patrick-ogrady: what about voter.recovered (for example)?

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -283,73 +287,96 @@ where
...
+                            timed_out: false,
+                        };
+                        finalized = new_finalized;
+
+                        // Close the root span of any view the chain has now decided
+                        for (_, round) in work.range_mut(..=finalized) {
+                            round.close_span();
+                        }
+
+                        let round = work
```
patrick-ogrady, line 328: Add a comment here?

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -283,73 +287,96 @@ where
...
+                        {
+                            let participants = self.forward_targets(round, &proposal, leader);
+                            self.forward_proposal(proposal, participants);
+                        }
 
-                    // Setting leader may enable batch verification
-                    updated_view = current.view;
-                }
-                Message::Constructed(message) => {
-                    // If the view isn't interesting, we can skip
```
patrick-ogrady, line 341: Retain this comment

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -283,73 +287,96 @@ where
...
+                        }
+                        let round = work.entry(view).or_insert_with(|| self.new_round(view));
+                        let process = process_span(round.span());
+                        let _guard = process.entered();
+                        round.add_constructed(message);
+                        self.added.inc();
+                        updated_view = view;
                     }
-
-                    // Add the message to the verifier
```
patrick-ogrady, line 347: Retain this comment

consensus/src/simplex/actors/resolver/actor.rs
```diff
@@ -148,6 +162,51 @@ impl<
         }
     }
 
+    fn apply_effects(
```
patrick-ogrady, line 166: add function comment

consensus/src/simplex/actors/resolver/actor.rs
```diff
@@ -148,6 +162,51 @@ impl<
...
+                }
+                Effect::RetainAbove(floor) => {
+                    let floor = U64::from(floor);
+                    let _ = resolver.retain(move |candidate, _| *candidate > floor);
+                }
+            }
+        }
+    }
+
+    fn fetch(
```
patrick-ogrady, line 192: add function comment

consensus/src/simplex/metrics.rs
```diff
@@ -20,6 +20,25 @@ pub enum TimeoutReason {
...
 }
 
+impl TimeoutReason {
+    /// Returns the stable trace field value for this reason.
+    ///
+    /// Matches the `EncodeLabelValue` rendering so a timeout trace span and its
+    /// metric series share the same string.
+    pub const fn as_str(self) -> &'static str {
+        match self {
+            Self::Inactivity => "Inactivity",
```
patrick-ogrady, line 30: we should use pascal case here for consistency?

patrick-ogrady: (and anywhere we do this sort of as_str conversion?)

Should we just impl display or something?

consensus/src/simplex/actors/span.rs
```diff
@@ -0,0 +1,50 @@
...
+    /// Returns the active span, or a disabled span when pending or closed.
+    pub(crate) fn get(&self) -> Span {
+        match self {
+            Self::Open(span) => span.clone(),
+            Self::Pending | Self::Closed => Span::none(),
+        }
+    }
+
+    /// Opens the span from `make` when pending. No-op once open or closed.
+    pub(crate) fn open(&mut self, make: impl FnOnce() -> Span) {
```
patrick-ogrady, line 35: I think we could assert if we try to create on closed? that should never happen?

consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -1046,6 +1198,7 @@ impl<
                     // If the leader nullified or is inactive, reduce leader
                     // timeout to now
                     batcher.update(
```
patrick-ogrady: I think we could combine .view_span() + .last_finalized() into a single function that returns for batcher?

resolver/src/subscribers.rs
```diff
@@ -36,35 +37,37 @@ where
         self.entries.contains_key(key)
     }
 
-    /// Add subscribers for a key, deduplicating subscribers already retained.
+    /// Add subscribers for a key, each paired with the span of its fetch.
+    ///
+    /// A subscriber's span is retained only when the subscriber is first seen;
+    /// later fetches for an already-tracked subscriber keep the original span.
```
patrick-ogrady: note that this occurs on retry?

resolver/src/subscribers.rs
```diff
@@ -83,35 +86,43 @@ where
...
+        if entry.is_empty() {
             self.entries.remove(key);
             return None;
         }
         self.pending(key)
     }
 
-    fn non_empty(subscribers: &BTreeSet<S>) -> Option<NonEmptyVec<S>> {
-        NonEmptyVec::try_from(subscribers.iter().cloned().collect::<Vec<_>>()).ok()
+    fn non_empty(
```
patrick-ogrady, line 121: add function comment

patrick-ogrady (commented):
very close now

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -283,73 +287,96 @@ where
...
+                        updated_view = current.view;
+                    }
+                    Message::Constructed(message) => {
+                        if !interesting(self.activity_timeout, finalized, current.view, view, false)
+                        {
+                            continue;
+                        }
+                        let round = work.entry(view).or_insert_with(|| self.new_round(view));
+                        let process = process_span(round.span());
+                        let _guard = process.entered();
```
patrick-ogrady, line 380: instead of using explicit drop guards, should we just use {}?

consensus/src/simplex/actors/batcher/actor.rs
patrick-ogrady: Switched to the `entered()` guard the other voter enqueues already use (the `on_end` block around `voter.recovered`, the `process` guard in the mailbox arms), so the parent-attribution mechanism is uniform instead of a one-off `in_scope`.

consensus/src/simplex/metrics.rs
patrick-ogrady: Made the trace-only enums snake_case (`FetchReason::as_str`, `Certificate::kind`). Left `TimeoutReason::as_str` PascalCase because it is pinned to its `EncodeLabelValue` metric label (the derive emits `stringify!(Variant)`, e.g. `LeaderNullify`): snake-casing only the trace field would desync trace vs metric, and snake-casing the metric is a breaking label-format change. Kept `as_str` over `Display` (returns `&'static str`, no per-record formatting). Leaving this thread open: happy to snake_case `TimeoutReason` and its metric label too if you want full consistency.

consensus/src/simplex/actors/span.rs
patrick-ogrady: Done. `assert!(!matches!(self, Self::Closed), ...)` on both `open` and `adopt` (verified `enter_view`/`set_span` cannot legitimately reach a closed round, so it only fires on a real bug).

consensus/src/simplex/actors/voter/actor.rs
patrick-ogrady: Done. Added `State::batcher_context(view) -> (Span, View)`, used at both `batcher.update` sites.

resolver/src/subscribers.rs
patrick-ogrady: Done. The `insert` doc now notes the duplicate-subscriber case happens on retry.

consensus/src/simplex/actors/batcher/actor.rs
patrick-ogrady: Done. Replaced the explicit `drop(_guard)` with a `{ }` block scoping the guard.

patrick-ogrady:
bugbot run

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -249,6 +250,30 @@ impl<E: Clock + CryptoRngCore + Metrics, S: Scheme<D>, L: ElectorConfig<S>, D: D
...
+            .unwrap_or_else(Span::none)
+    }
+
+    /// Closes the root span of every decided view (at or below the finalized
+    /// view) so each view trace ends when the chain commits it rather than when
+    /// the round is later pruned.
+    pub fn close_decided_spans(&mut self) {
+        for (_, round) in self.views.range_mut(..=self.last_finalized) {
+            round.close_span();
+        }
```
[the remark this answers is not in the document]
patrick-ogrady: It is possible a nullified view can still be part of the canonical chain (and thus it makes sense to keep around until decided).

PR #4048, opened 2026-06-15
[storage/freezer] Improve Recovery During Resize

### LINKED ISSUE #4043: [storage/freezer][medium] avoid no-checkpoint lookup loss after partial table resize
## Context

`storage::freezer::Freezer` stores key heads in a fixed hash table and value records in an oversized journal. Table resize is incremental: a `sync()` may double the table blob and copy only one chunk of old buckets before returning a  `Checkpoint`.

Checkpointed recovery uses the committed logical `table_size` from the checkpoint. No-checkpoint recovery through `Freezer::init` reconstructs state from the table blob itself.

## Issue

A no-checkpoint `Freezer` user can lose key lookups after an incremental resize is persisted but not completed. On restart, `Freezer::init` infers `table_size` from the expanded blob length, treats the doubled table as complete, and hashes keys into new-half buckets that may not have been copied yet.

`close()` can create this state too: it drains only a resize already in progress before its final `sync()`, but that final `sync()` can start a fresh resize, advance only one chunk when `table_resize_chunk_size < table_size`, sync the expanded
  blob, and return with the checkpoint still recording the old `table_size`.

## Impact

Committed values can become unreachable after restart for direct no-checkpoint `Freezer` consumers. The value records remain in the oversized journal, but key lookup can return `None` for a key that was readable before shutdown.

Checkpoint-bearing recovery is not affected because the checkpoint path rewinds/resizes the table back to the committed logical `table_size`.

## Root Cause

No-checkpoint recovery has no durable resize-progress marker and equates physical table blob length with completed logical table size.

## Code

  - `Freezer::init` uses no-checkpoint recovery:
    https://github.com/commonwarexyz/monorepo/blob/e68fede1155304365352254e7f0988df1c714870/storage/src/freezer/storage.rs#L628-L629
  - No-checkpoint recovery derives `table_size` from the current table blob length:
    https://github.com/commonwarexyz/monorepo/blob/e68fede1155304365352254e7f0988df1c714870/storage/src/freezer/storage.rs#L728-L756
  - Lookups hash with the recovered `self.table_size`:
    https://github.com/commonwarexyz/monorepo/blob/e68fede1155304365352254e7f0988df1c714870/storage/src/freezer/storage.rs#L809-L812
  - Resize expands the table blob before resize completion:
    https://github.com/commonwarexyz/monorepo/blob/e68fede1155304365352254e7f0988df1c714870/storage/src/freezer/storage.rs#L978-L991
  - `advance_resize` copies one chunk and updates `self.table_size` only after the final chunk:
    https://github.com/commonwarexyz/monorepo/blob/e68fede1155304365352254e7f0988df1c714870/storage/src/freezer/storage.rs#L1011-L1069
  - `sync()` can start/advance resize and return a checkpoint with the old `table_size`:
    https://github.com/commonwarexyz/monorepo/blob/e68fede1155304365352254e7f0988df1c714870/storage/src/freezer/storage.rs#L1084-L1117
  - `close()` drains only an already-running resize before calling final `sync()`:
    https://github.com/commonwarexyz/monorepo/blob/e68fede1155304365352254e7f0988df1c714870/storage/src/freezer/storage.rs#L1121-L1129

patrick-ogrady:
Resolves: #4043


storage/src/freezer/storage.rs
```diff
@@ -655,6 +710,19 @@ impl<E: BufferPooler + Context, K: Array, V: CodecShared> Freezer<E, K, V> {
         let mut oversized: Oversized<E, Record<K>, V> =
             Oversized::init(context.child("oversized"), oversized_cfg).await?;
 
+        // Initialize metadata for committed table state
+        let metadata_cfg = MetadataConfig {
+            partition: format!("{}-metadata", config.table_partition),
```
patrick-ogrady: this isn't safe to do...we need to provide explicitly?

storage/src/freezer/storage.rs
```diff
@@ -404,6 +455,10 @@ pub struct Freezer<E: BufferPooler + Context, K: Array, V: CodecShared> {
     // Table blob that maps slots to key index chain heads
     table: E::Blob,
 
+    // Metadata that tracks the committed logical table size
+    metadata: Metadata<E, MetadataKey, MetadataValue>,
+    metadata_table_size: u32,
```
patrick-ogrady: just access from inside metadata?

storage/src/freezer/storage.rs
patrick-ogrady: Addressed in 6aebea7e: freezer metadata now uses an explicit `metadata_partition` on `freezer::Config`, and archive immutable configs expose/pass `freezer_metadata_partition`. Callers, tests, benches, and the freezer fuzz target now provide dedicated metadata partitions instead of deriving one from `table_partition`.

storage/src/freezer/storage.rs
patrick-ogrady: Addressed in 6aebea7e: removed the cached `metadata_table_size` field. `sync` now reads `TABLE_SIZE_KEY` directly from `Metadata` and matches `MetadataValue::TableSize` before deciding whether the metadata store needs to be updated.

storage/src/freezer/storage.rs
```diff
@@ -1103,6 +1187,12 @@ impl<E: BufferPooler + Context, K: Array, V: CodecShared> Freezer<E, K, V> {
 
         // Sync updated table entries
         self.table.sync().await?;
+        if self.metadata_table_size != self.table_size {
+            self.metadata
+                .put(TABLE_SIZE_KEY.clone(), MetadataValue::TableSize(self.table_size));
+            self.metadata.sync().await?;
+            self.metadata_table_size = self.table_size;
```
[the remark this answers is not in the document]
patrick-ogrady: Addressed in 6aebea7e: no-checkpoint recovery now detects when the physical table is exactly the completed doubled table for the metadata size and preserves that table instead of truncating to stale metadata. Added `stale_metadata_does_not_shrink_completed_resize` to cover the crash window after table sync and before metadata sync.

storage/src/freezer/mod.rs
```diff
@@ -255,6 +260,9 @@ pub struct Config<C> {
     /// The [commonware_runtime::Storage] partition to use for storing the table.
     pub table_partition: String,
 
+    /// The [commonware_runtime::Storage] partition to use for storing metadata.
+    pub metadata_partition: String,
```
patrick-ogrady: In other places, we do a "prefix" here once we have multiple structs. Not sure if we should make this migration.

storage/src/freezer/storage.rs
```diff
@@ -404,6 +446,9 @@ pub struct Freezer<E: BufferPooler + Context, K: Array, V: CodecShared> {
     // Table blob that maps slots to key index chain heads
     table: E::Blob,
 
+    // Metadata that tracks the committed logical table size
+    metadata: Metadata<E, MetadataKey, MetadataValue>,
```
patrick-ogrady: typically metadata goes first (defined right after context)

storage/src/freezer/storage.rs
```diff
@@ -728,8 +803,31 @@ impl<E: BufferPooler + Context, K: Array, V: CodecShared> Freezer<E, K, V> {
             // Existing table without checkpoint
             (_, None) => {
                 // Find max epoch/section and clean invalid entries in a single pass
-                let table_size = (table_len / Entry::FULL_SIZE as u64) as u32;
-                let (modified, max_epoch, max_section, resizable) = Self::recover_table(
+                let physical_table_size = (table_len / Entry::FULL_SIZE as u64) as u32;
```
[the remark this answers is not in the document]
patrick-ogrady: Addressed in 2132e8572: `init_with_checkpoint` now checks durable freezer metadata before trusting the supplied checkpoint. If metadata has advanced past the checkpoint table size, recovery uses the shared table-size recovery path, preserves a completed doubled table, rebuilds the checkpoint from recovered table entries, and keeps metadata at the correct table size. Added `stale_checkpoint_does_not_shrink_completed_resize` for this case.

storage/src/freezer/storage.rs
patrick-ogrady: Addressed in 2132e8572: moved the freezer `metadata` field immediately after `context`, matching the usual metadata-first layout.

storage/src/freezer/mod.rs
patrick-ogrady: I left this as explicit `table_partition` / `metadata_partition` fields for this PR. The original issue was unsafe implicit derivation from the table partition; keeping both fields caller-owned avoids another storage naming migration here. A prefix-style API can still be a follow-up if we want to standardize freezer around generated partition names.

storage/src/freezer/mod.rs
```diff
@@ -17,6 +17,8 @@
 //!
 //! These journals are combined via [crate::journal::segmented::oversized], which coordinates
 //! crash recovery between them.
+//! A small [crate::metadata::Metadata] store tracks the committed logical table size so recovery
```
patrick-ogrady: This should be in the list above?

commonware-llm: Addressed in 06eb927c3: moved freezer metadata into the on-disk component list and removed the duplicate standalone paragraph.

storage/src/archive/conformance.rs
```diff
@@ -66,6 +66,7 @@ impl Conformance for ArchiveImmutable {
             let config = immutable::Config {
                 metadata_partition: format!("archive-immutable-metadata-{seed}"),
                 freezer_table_partition: format!("archive-immutable-table-{seed}"),
```
patrick-ogrady: I think this should be `immutable-freezer-table-` and the metadata should be `immutable-freezer-metadata-`?

storage/src/freezer/benches/utils.rs
```diff
@@ -33,6 +33,9 @@ pub const VALUE_PARTITION: &str = "freezer-bench-value";
 /// Partition for [Freezer] table benchmarks.
 pub const TABLE_PARTITION: &str = "freezer-bench-table";
 
+/// Partition for [Freezer] metadata benchmarks.
+pub const METADATA_PARTITION: &str = "freezer-bench-table-metadata";
```
patrick-ogrady: just `freezer-bench-metadata`?

storage/src/freezer/storage.rs
```diff
@@ -177,6 +180,59 @@ impl FixedSize for Checkpoint {
 /// Name of the table blob.
 const TABLE_BLOB_NAME: &[u8] = b"table";
 
+/// Metadata key for committed table resize state.
+const TABLE_STATE_KEY: MetadataKey = MetadataKey::new(0);
+
+/// Item stored in [Metadata] for [Freezer].
+#[cfg_attr(feature = "arbitrary", derive(arbitrary::Arbitrary))]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+enum MetadataValue {
```
patrick-ogrady: Just encode as `State` but put a 0 type in front to support future use of enum?

storage/src/freezer/storage.rs
```diff
@@ -3,17 +3,20 @@ use crate::{
...
+};
 use commonware_cryptography::{crc32, Crc32, Hasher};
 use commonware_runtime::{
     buffer,
     iobuf::EncodeExt,
     telemetry::metrics::{Counter, MetricsExt as _},
     Blob, Buf, BufMut, BufferPooler, IoBuf,
 };
-use commonware_utils::{Array, Span};
+use commonware_utils::{sequence::U64 as MetadataKey, Array, Span};
```
patrick-ogrady: Just use U64 in the key state

storage/src/archive/conformance.rs
patrick-ogrady: Addressed in b4ef661e0/a0b03cf23 and fixtures updated in 4d8d7ffb3: archive immutable conformance now uses `immutable-freezer-table-*` and `immutable-freezer-metadata-*`, with the other freezer-owned archive partitions following the same `*-freezer-*` naming.

storage/src/freezer/benches/utils.rs
patrick-ogrady: Addressed in b4ef661e0: the freezer bench metadata partition is now `freezer-bench-metadata`. The related storage fixture update is in 4d8d7ffb3.

storage/src/freezer/storage.rs
```diff
@@ -1103,6 +1282,14 @@ impl<E: BufferPooler + Context, K: Array, V: CodecShared> Freezer<E, K, V> {
 
         // Sync updated table entries
         self.table.sync().await?;
+        let metadata_state = MetadataValue::State {
+            table_size: self.table_size,
+            resize_progress: self.resize_progress,
+        };
+        if self.metadata.get(&TABLE_STATE_KEY).copied() != Some(metadata_state) {
```
patrick-ogrady: I think we can check this without copying?

patrick-ogrady: actually, I think this will always change?

patrick-ogrady:
bugbot run

storage/src/archive/immutable/storage.rs
```diff
@@ -165,7 +165,7 @@ impl<E: BufferPooler + Context, K: Array, V: CodecShared> Archive<E, K, V> {
         // Initialize ordinal
         //
         // TODO (#1227): Use sharded metadata to provide consistency
-        let ordinal = Ordinal::init_with_bits(
+        let ordinal = Ordinal::init(
```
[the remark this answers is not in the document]
patrick-ogrady: This is intended. We don't want to surface data not committed into the archive.

utils/src/bitmap/mod.rs
```diff
@@ -24,6 +28,28 @@ commonware_macros::stability_mod!(ALPHA, pub mod roaring);
 /// The default [BitMap] chunk size in bytes.
 pub const DEFAULT_CHUNK_SIZE: usize = 8;
 
+/// Build section bitmaps from absolute indices.
+///
+/// Each index maps to key `index / items_per_section` and sets bit
```
roberto-bayardo: "section" isn't really a bitmap concept so this is hard to understand in the context of this file.  Not sure if this should be in a different file or if we just need to clarify what it's for?

danlaine: Also fwiw I'm working in my PRs to replace "section" with "blob" in the contiguous journals. We use them somewhat interchangeably, which I find a bit confusing.

patrick-ogrady: Making it something like `items_per_blob`?

patrick-ogrady: Moved to `storage::utils` for now.

storage/src/ordinal/storage.rs
```diff
@@ -147,68 +151,93 @@ impl<E: BufferPooler + Context, V: CodecFixed<Cfg = ()>> Ordinal<E, V> {
...
+            let empty = vec![0u8; Record::<V>::SIZE];
+            for (section, (blob, size)) in &blobs {
+                // A section with no bitmap requires every record, so nothing is cleared
+                let Some(Some(bits)) = bits.get(section) else {
+                    continue;
+                };
+                let mut modified = false;
+                for bit_index in 0..(*size / record_size) {
+                    if bit_index >= bits.len() || !bits.get(bit_index) {
+                        blob.write_at(bit_index * record_size, empty.clone())
```
patrick-ogrady, line 190: Arguably we could avoid this?

patrick-ogrady: If we will continue to be initialized by the checkpoint, there should be no issue in just allowing an overwrite to fix this?

storage/src/freezer/storage.rs
```diff
@@ -661,26 +677,13 @@ impl<E: BufferPooler + Context, K: Array, V: CodecShared> Freezer<E, K, V> {
...
-            // Existing table with checkpoint
-            (_, Some(checkpoint)) => {
+        let (checkpoint, resizable) = match checkpoint {
+            // Non-empty checkpoint: align existing data to it
+            Some(checkpoint) if !checkpoint.is_empty() => {
+                // A non-empty checkpoint against an empty table references data that does not exist
+                if table_len == 0 {
+                    return Err(Error::CheckpointMismatch);
+                }
                 assert!(
```
roberto-bayardo, line 687: You return errror above but panic! on this failure, seems a bit inconsistent?  (Both are corruption cases I believe?).

storage/src/ordinal/storage.rs
patrick-ogrady: Let's keep this as-is for consistency with other storage primitives (where we clean up on recovery).

storage/src/freezer/storage.rs
patrick-ogrady: Fixed in 2ab130272: invalid non-empty checkpoint table sizes now return `Error::CheckpointMismatch` instead of panicking. Added `invalid_checkpoint_table_size_errors` to cover the non-power-of-two checkpoint case.

storage/conformance.toml
```diff
@@ -1,6 +1,6 @@
 ["commonware_storage::archive::conformance::ArchiveImmutable"]
 n_cases = 128
-hash = "e349d1a351b91767c785fbab59128c3ea5c82c3dfb47e280202aff56c1cf6ad3"
+hash = "09d55a9a01a3dfef029ebd4d5f22d1be8b232e82a9ec23931e860b01cb527dd7"
```
patrick-ogrady: We changed the name of a partition in conformance testing, not the format.

storage/src/ordinal/storage.rs
danlaine: > Let's keep this as-is for consistency with other storage primitives (where we clean up on recovery).

Should we update this comment, then? It makes it sound as though this zeroing is necessary:

```
// Zero records the committed bits do not mark set so a partial resize cannot
// resurrect them
```

storage/conformance.toml
patrick-ogrady: Updated in eb8bd8f17: kept the `freezer-` component while preserving the archive namespace, so these are now `archive-immutable-freezer-{table,key,value}-*`. Regenerated only the aggregate `ArchiveImmutable` conformance hash; the codec/on-disk record format entries are unchanged.

storage/src/freezer/storage.rs
patrick-ogrady: Final approach updated in eb8bd8f17: kept the `table_size` assertion. A zero/non-power-of-two `table_size` is an invalid checkpoint input, not a stored-data mismatch. The adjacent empty-table case still returns `CheckpointMismatch` because a structurally valid non-empty checkpoint cannot be reconciled with the stored table.

storage/src/ordinal/storage.rs
patrick-ogrady: Updated in 078b7465e: reworded the comment so it no longer implies zeroing is required to prevent replay. It now says replay ignores records outside the committed bits, and recovery clears them so stored blobs match the checkpointed view.

roberto-bayardo (dismissed):

storage/src/freezer/benches/utils.rs
```diff
@@ -63,7 +63,7 @@ pub async fn init(ctx: Context) -> FreezerType {
         table_replay_buffer: NZUsize!(TABLE_REPLAY_BUFFER),
         codec_config: (),
     };
-    Freezer::init(ctx, cfg).await.unwrap()
+    Freezer::init(ctx, cfg, None).await.unwrap()
```
danlaine: This finding about benchmarks seems legit:

```
**Restart benchmarks no longer measure recovery.** `init()` here now calls `Freezer::init(ctx, cfg, None)` / `Ordinal::init(ctx, cfg, None)`, and `None` now means *reset* (delete the partitions, create an empty store). So `bench_restart` populates + persists in phase 1, then the timed phase 2 just deletes that data and opens an empty store — the `// replay happens inside init` comment is no longer true, and the `items` sweep no longer scales with anything. It's also destructive: after the first timed iteration the data is gone, so the rest measure empty init.

For a PR whose point is recovery, this silently kills the recovery-perf signal. Fix is just wiring — recover the way `Archive` does:

- Freezer: keep the `Checkpoint` from `close()` and time `init(..., Some(checkpoint))`.
- Ordinal: build bitmaps from the synced indices via `bits_for_indices` and time `init(..., Som
```

patrick-ogrady: Not sure how I missed this. Good catch.

storage/src/freezer/benches/restart.rs
```diff
@@ -42,7 +42,7 @@ fn bench_restart(c: &mut Criterion) {
         // Tear down
         let cleaner = commonware_runtime::tokio::Runner::new(cfg.clone());
         cleaner.start(|ctx| async move {
-            let store = init(ctx).await;
+            let store = init(ctx, None).await;
```
patrick-ogrady, line 45: This technically "pre-deletes" things but that's fine.

danlaine:
One additional informational finding. Probably a non-issue given it's beta but still worth flagging.

```
# Destructive `init(..., None)` Migration Risk

`Freezer::init` and `Ordinal::init` now require explicit recovery metadata. Passing `None` no longer means "recover existing data if possible"; it means "delete existing data and start empty."

This creates a sharp migration hazard for direct `Freezer` or `Ordinal` callers. A mechanical upgrade from the old `init(ctx, cfg)` API to `init(ctx, cfg, None)` will compile, but the first startup can remove existing on-disk data.

The docs should make this reset behavior very prominent: use `None` only for a fresh store or intentional reset. Existing stores must persist and pass the last `Checkpoint` / ordinal bitmaps to recover durable data.
```

patrick-ogrady:
> One additional informational finding. Probably a non-issue given it's beta but still worth flagging.
> 
> ```
> # Destructive `init(..., None)` Migration Risk
> 
> `Freezer::init` and `Ordinal::init` now require explicit recovery metadata. Passing `None` no longer means "recover existing data if possible"; it means "delete existing data and start empty."
> 
> This creates a sharp migration hazard for direct `Freezer` or `Ordinal` callers. A mechanical upgrade from the old `init(ctx, cfg)` API to `init(ctx, cfg, None)` will compile, but the first startup can remove existing on-disk data.
> 
> The docs should make this reset behavior very prominent: use `None` only for a fresh store or intentional reset. Existing stores must persist and pass the last `Checkpoint` / ordinal bitmaps to recover durable data.
> ```

Yep, this is known/accepted. Will be called out in the release notes.

I'm not aware of anyone using these/using them would likely involve some other data structure where you should batch metadata writes. If there is appetite at some point for standalone usage, we can revisit.

danlaine (approved):

patrick-ogrady (commented):
Confirmed no regression in benchmarks.

PR #4052, opened 2026-06-15
[storage/qmdb/any] Use parallel index probe in `get_many_map`

danlaine:
## Change
The snapshot-index probe — one location lookup per requested key — was a serial loop on the calling thread. It now shards the keys across the `Strategy` pool: each worker probes its slice into a local `Vec<(global_key_index, location)>`, and the results are concatenated.

## Why
The index is read-only and `Send + Sync`, so concurrent probes need no locking. Each probe is a `BTreeMap` descent (random-memory, cache-miss bound); running them in parallel overlaps those misses across cores.

## Result
Constantinople benchmark, `depth=0`, 1,000,000 keys, 32,768 updates/block.

| Variant                        | main     | with change | Δ      |
|--------------------------------|----------|-------------|--------|
| `any::ordered::fixed::mmb`     | 22.6 ms  | 20.1 ms     | −11%   |
| `current::ordered::fixed::mmb` | 27.1 ms  | 24.6 ms     | −9%    |

### DIFF
--- a/storage/src/qmdb/any/db.rs
+++ b/storage/src/qmdb/any/db.rs
@@ -250,8 +250,6 @@ where
         self.metrics.reads.get_many_calls.inc();
         self.metrics.reads.keys_requested.inc_by(keys.len() as u64);
 
-        let mut results: Vec<Option<T>> = (0..keys.len()).map(|_| None).collect();
-
         // Phase 1: Collect candidate locations from the in-memory index. Each key may map to
         // multiple locations due to hash collisions. The index is read-only and Send+Sync, so
         // for large batches the probe is sharded across the strategy pool; each candidate carries
@@ -283,6 +281,7 @@ where
                 .collect()
         };
 
+        let mut results: Vec<Option<T>> = (0..keys.len()).map(|_| None).collect();
         if candidates.is_empty() {
             return Ok(results);
         }
--- a/storage/src/qmdb/any/db.rs
+++ b/storage/src/qmdb/any/db.rs
@@ -28,8 +28,8 @@ use commonware_utils::bitmap;
 use core::num::NonZeroU64;
 use std::{collections::HashMap, sync::Arc};
 
-/// Minimum keys per shard for [`Db::get_many_map`] to parallelize the index probe rather than
-/// probe serially, ensuring each shard carries enough work to amortize the thread-pool handoff.
+/// Estimate of the number of keys per shard at which parallelizing [`Db::get_many_map`]'s
+/// index probe is faster than probing serially.
 const MIN_PROBES_PER_SHARD: usize = 8;
 
 /// Metrics for Any QMDBs.
--- a/storage/src/qmdb/any/db.rs
+++ b/storage/src/qmdb/any/db.rs
@@ -256,7 +256,9 @@ where
         // its global key index, so partition order does not matter.
         let strategy = self.strategy();
         let parallelism = strategy.parallelism_hint().max(1);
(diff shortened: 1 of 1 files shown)


storage/src/qmdb/any/db.rs
```diff
@@ -246,13 +246,38 @@ where
...
 
-        self.snapshot
-            .get_many(keys, |key_idx, &loc| candidates.push((key_idx, *loc)));
+        // Phase 1: Collect candidate locations from the in-memory index. Each key may map to
+        // multiple locations due to hash collisions. The index is read-only and Send+Sync, so
+        // for large batches the probe is sharded across the strategy pool; each candidate carries
+        // its global key index, so partition order does not matter.
+        let strategy = self.strategy();
+        let parallelism = strategy.parallelism_hint().max(1);
+        let mut candidates: Vec<(usize, u64)> = if keys.len() < parallelism {
```
patrick-ogrady: Do we need this? I think it'll handle < parallelism?

storage/src/qmdb/any/db.rs
```diff
@@ -246,13 +246,38 @@ where
         self.metrics.reads.get_many_calls.inc();
         self.metrics.reads.keys_requested.inc_by(keys.len() as u64);
 
-        // Phase 1: Collect candidate locations from the in-memory index.
-        // Each key may map to multiple locations due to hash collisions.
-        let mut candidates: Vec<(usize, u64)> = Vec::with_capacity(keys.len());
         let mut results: Vec<Option<T>> = (0..keys.len()).map(|_| None).collect();
```
patrick-ogrady: I think you should move this below the comment block?

storage/src/qmdb/any/db.rs
danlaine: It will handle < parallelism correctly, so the guard isn’t needed for correctness. But it avoids paying the Rayon::install / map_collect_vec overhead for tiny batches. For tiny batches, the overhead may cause this to be slower than the serial case.

I updated this to MIN_PROBES_PER_SHARD * parallelism so we only parallelize when each shard has enough work to amortize the pool handoff. `MIN_PROBES_PER_SHARD = 8` was a guesstimate.

storage/src/qmdb/any/db.rs
danlaine: moved

storage/src/qmdb/any/db.rs
```diff
@@ -28,6 +28,10 @@ use commonware_utils::bitmap;
 use core::num::NonZeroU64;
 use std::{collections::HashMap, sync::Arc};
 
+/// Minimum keys per shard for [`Db::get_many_map`] to parallelize the index probe rather than
+/// probe serially, ensuring each shard carries enough work to amortize the thread-pool handoff.
+const MIN_PROBES_PER_SHARD: usize = 8;
```
roberto-bayardo, line 33: How did you select this value?

danlaine: This is a guesstimate -- I can try to find a more evidence based value

roberto-bayardo: a guesstimate is fine as long as it's documented as such (so we can revisit)

danlaine: Updated comment:

```rs
/// Estimate of the number of keys per shard at which parallelizing [`Db::get_many_map`]'s
/// index probe is faster than probing serially.
const MIN_PROBES_PER_SHARD: usize = 8;
```

roberto-bayardo (approved):

PR #4053, opened 2026-06-15
[storage/journal/contiguous] Use `Sealed` blobs in variable journal

danlaine:
# Use sealed blobs in the variable contiguous journal

`journal::contiguous::variable` now owns its on-disk storage directly — a contiguous run of full **sealed blobs** plus one writable **tail** — instead of delegating to a segmented variable journal. This drops a layer of indirection and lets reads run against an owned snapshot rather than holding the journal open.

Stacked on `danlaine/variable-refactor-min` (PR #4041); it builds on the contiguous fixed-journal checkpoint/replay and blob machinery introduced there.

## Changes

- **Storage model** (`variable.rs`, `blobs.rs`): item data lives in sealed blobs plus a writable tail, and the data blobs are the source of truth for replayed item bytes. A separate fixed offsets journal maps each position to its frame's byte offset and supplies the pruning boundary and the recovery watermark.
- **Owned-snapshot reads**: the variable journal serves reads through `blobs.rs` owned snapshots. `reader()` returns a snapshot with frozen bounds — so every position stays readable for the reader's life even across a concurrent prune, without holding a journal-wide read lock — and a rewind that would mutate bytes under a live reader is refused with `Error::BlobInUse`.
- **Crash recovery** (`variable.rs`): on init the data blobs are reconciled with the offsets journal. Recovery scans the trailing/newest blobs, rebuilds offset entries from the recovery anchor (the durable watermark, otherwise the pruning boundary), truncates torn trailing frame bytes to the last valid frame, removes newer blobs after a short/missing blob in the rebuilt suffix, and relies on `Blobs::recover` to enforce retained-blob contiguity. Payloads are decoded only for the scanned/rebuilt suffix; data **below** the recovery anchor is trusted and not re-validated at init, so corruption there surfaces on a later read rather than during init.
- **Rewind ordering** (`variable.rs`): `rewind` lowers and truncates the offsets journal *before* truncating the data (the reverse of the segmented journal's order). A crash mid-rewind then leaves the offsets at or behind the data — a state recovery rebuilds — instead of stranding a short blob below a watermark recovery trusts. This is an intentional, recovery-relevant behavior change relative to the base.
- **Shared frame codec** (`frame.rs`, `segmented/variable.rs`): the varint-framed item encode/decode is factored into `journal::frame`, now shared by `segmented::variable` and `contiguous::variable`.
- **Replay**: the variable journal adopts the contiguous journals' shared replay driver — `SealedReplayState` (sealed blobs) and `TailReplayState` (the writable tail) implement `replay::Source` and stream batches via `into_batches`.

Internal terminology moves from "section" to "blob" (for example `position_to_blob`); the public `Config` field remains `items_per_section`.


storage/src/journal/contiguous/blobs.rs
```diff
@@ -423,21 +423,6 @@ impl<E: Context> Blobs<E> {
         }
         Partition::remove_all(&self.partition.context, &self.partition.name).await
     }
-
-    /// Test helper: make one blob durable.
-    #[cfg(test)]
-    pub(super) async fn sync_blob(&self, blob: u64) -> Result<(), Error> {
```
danlaine, line 429: Moved to `mod tests`

PR #4055, opened 2026-06-16
[utils] Add `Cache`

patrick-ogrady:


utils/src/cache.rs
```diff
@@ -0,0 +1,922 @@
+//! A fixed-capacity cache with CLOCK eviction.
```
andresilva, line 1: One thing we could consider is some small (configurable) changes to CLOCK. If we assume zipf distribution then it probably makes more sense to insert new entries with `referenced = false`, since this would avoid giving a second chance to one-off hits.

utils/src/cache.rs
```diff
@@ -0,0 +1,922 @@
...
+                drop_keys.push(key.clone());
+            }
+        }
+        for key in drop_keys {
+            if let Some(slot) = self.index.remove(&key) {
+                self.slots[slot].referenced.store(false, Ordering::Relaxed);
+                self.free.push(slot);
+            }
+        }
+    }
```
roberto-bayardo, line 299: More efficient / simpler IMO:

```suggestion
pub fn retain<F: FnMut(&K, &V) -> bool>(&mut self, mut keep: F) {
      let Self { index, slots, free, .. } = self;
      index.retain(|key, &mut slot| {
          let keep = keep(key, &slots[slot].value);
          if !keep {
              slots[slot].referenced.store(false, Ordering::Relaxed);
              free.push(slot);
          }
          keep
      });
  }
```

utils/src/cache.rs
```diff
@@ -0,0 +1,922 @@
...
+            Some(slot) => {
+                self.slots[slot].key = key.clone();
+                self.slots[slot].value = value;
+                self.slots[slot].referenced.store(true, Ordering::Relaxed);
+                self.index.insert(key, slot);
+            }
+            None => {
+                self.grow(key, value);
+            }
+        }
```
roberto-bayardo:
```suggestion
        self.insert_value(key, value);
```

utils/src/cache.rs
patrick-ogrady: I prefer deferring an optimization like this until later. It very well could help tho.

utils/src/cache.rs
```diff
@@ -0,0 +1,911 @@
...
+    /// Returns a reference to the value for `key`, recording use.
+    ///
+    /// Recording use sets the entry's reference bit so the next eviction sweep
+    /// grants it a second chance. This takes `&self` so it can be called
+    /// concurrently behind a shared lock.
+    #[inline]
+    pub fn get(&self, key: &K) -> Option<&V> {
+        let &index = self.index.get(key)?;
+        let slot = &self.slots[index];
+        slot.referenced.store(true, Ordering::Relaxed);
```
roberto-bayardo:
```suggestion
           // Only write when unset: skipping the store keeps the cache line shared across
           // concurrent readers instead of bouncing it on every hit.
           if !slot.referenced.load(Ordering::Relaxed) {
               slot.referenced.store(true, Ordering::Relaxed);
           }
```

utils/src/cache.rs
```diff
@@ -0,0 +1,911 @@
...
+/// concurrency details. The key index uses a randomly seeded ahash hasher under
+/// `std` and hashbrown's default hasher under `no_std`.
+pub struct Clock<K, V> {
+    /// Maps each live key to the index of its slot in `slots`.
+    ///
+    /// `index.len() + free.len() == slots.len()` always holds.
+    index: HashMap<K, usize, Hasher>,
+    /// Backing storage for slots, grown lazily up to `capacity` and then reused.
+    slots: Vec<Slot<K, V>>,
+    /// Slots detached from the index (by eviction overwrite never; by
```
roberto-bayardo: grammar isssue?

utils/src/cache.rs
roberto-bayardo: (Not sure if this really helps, but the AI suggested it might when I asked what might improve concurrency!)

patrick-ogrady: Yeah, I think this will actually help.

utils/Cargo.toml
```diff
@@ -39,6 +40,12 @@ zeroize.workspace = true
 version = "0.2.15"
 features = ["js"]
 
+[target.'cfg(target_arch = "wasm32")'.dependencies.getrandom_v3]
```
patrick-ogrady, line 43: if ahash is compiled with wasm in std, we will attempt to use `js-sys` for an rng (secure randomness!)

runtime/src/utils/buffer/paged/cache.rs
```diff
@@ -493,95 +433,37 @@ impl Cache {
...
+        let bytes_to_copy = std::cmp::min(buf.len(), self.page_size - offset_in_page);
+        buf[..bytes_to_copy].copy_from_slice(&page[offset_in_page..offset_in_page + bytes_to_copy]);
 
         bytes_to_copy
     }
 
     /// Put the given `page` into the page cache.
+    ///
+    /// On a hit or a reused (freed or evicted) slot the existing page buffer is overwritten in
+    /// place; only growth to capacity allocates a new buffer from the pool.
```
roberto-bayardo: Not sure I find this comment helpful. I'd rather see implementation details like this documented inside the fn personally if they are not important for the average caller.

roberto-bayardo (dismissed):

runtime/src/utils/buffer/paged/cache.rs
patrick-ogrady: I had a note to remove this and it must've gotten lost 🙃

roberto-bayardo (approved):

PR #4056, opened 2026-06-16
[runtime] add jemalloc feature

andresilva:
This PR adds an optional `jemalloc` feature to `commonware-runtime`. When enabled on supported targets, the runtime sets `tikv_jemallocator::Jemalloc` as the global allocator. This also removes the deployer's `LD_PRELOAD`-based jemalloc setup.

Using jemalloc through `LD_PRELOAD` swaps the process `malloc`/`free` symbols, which means that Rust still goes through the platform `System` allocator path. That loses some of the allocation layout information Rust already has, especially on deallocation.

## Rationale

The main distinction is where allocation layout information is preserved. The C allocator API does not carry Rust's `Layout` through the full allocation lifecycle. In particular, `free(ptr)` only receives a pointer, so size and alignment information have to be recovered by the allocator. Rust's allocator API does preserve this information: `alloc`, `dealloc`, and `realloc` are all called with a `Layout`, and callers are expected to pass the same layout back when freeing or resizing an allocation.

Using jemalloc as the Rust global allocator lets us exploit that. `tikv-jemallocator` can translate Rust's `Layout` into jemalloc's extended APIs, preserving size and alignment information that would otherwise be lost through the C std allocator interface.

This matters for `commonware-runtime` because it exposes buffer APIs where alignment and zeroing are part of the allocation contract.

The current Rust `System` path on Unix is roughly:

```text
alloc        -> malloc(...) or posix_memalign(...)
alloc_zeroed -> calloc(...) or alloc + memset for higher/special alignment
dealloc      -> free(ptr)
realloc      -> realloc(...) or fallback alloc/copy/free
```

With tikv-jemallocator:

```
alloc        -> malloc(size) or mallocx(size, MALLOCX_ALIGN(...))
alloc_zeroed -> calloc(1, size) or mallocx(size, flags | MALLOCX_ZERO)
dealloc      -> sdallocx(ptr, size, flags)
realloc      -> realloc(ptr, new_size) or rallocx(ptr, new_size, flags)
```

We enable `unprefixed_malloc_on_supported_platforms` so linked C/C++ libraries that call `malloc`, `calloc`, `realloc`, or `free` also use jemalloc on supported targets. This preserves the process-wide allocator behavior we previously got from LD_PRELOAD.

### DIFF
--- a/.github/workflows/fast.yml
+++ b/.github/workflows/fast.yml
@@ -120,6 +120,9 @@ jobs:
         include:
           - os: ubuntu-latest
             package: commonware-runtime
+            flags: "--features commonware-runtime/jemalloc"
+          - os: ubuntu-latest
+            package: commonware-runtime
             flags: "--features commonware-runtime/external"
           - os: windows-latest
             package: commonware-runtime
--- a/runtime/Cargo.toml
+++ b/runtime/Cargo.toml
@@ -59,6 +59,9 @@ opentelemetry-otlp = { workspace = true, features = ["http-proto"] }
 opentelemetry_sdk = { workspace = true, features = ["rt-tokio"] }
 tokio = { workspace = true, features = ["full"] }
 
+[target.'cfg(all(not(target_arch = "wasm32"), not(target_env = "msvc")))'.dependencies]
+tikv-jemallocator = { workspace = true, features = ["unprefixed_malloc_on_supported_platforms"], optional = true }
+
 [dev-dependencies]
 clap = { workspace = true, features = ["derive"] }
 rand = { workspace = true, features = ["small_rng"] }
@@ -75,6 +78,7 @@ arbitrary = [
 ]
 bench = [ "dep:crossbeam-queue" ]
 external = [ "pin-project" ]
+jemalloc = [ "dep:tikv-jemallocator" ]
 loom = [ "dep:loom" ]
 test-utils = []
 iouring = [ "iouring-network", "iouring-storage" ]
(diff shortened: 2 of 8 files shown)


.github/workflows/fast.yml
```diff
@@ -118,6 +118,9 @@ jobs:
     strategy:
       matrix:
         include:
+          - os: ubuntu-latest
+            package: commonware-runtime
+            flags: "--features commonware-runtime/jemalloc"
```
andresilva, line 123: Felt it was unnecessary to add this to the general CI pipelines to avoid churn. This should be enough to make sure that it compiles and passes basic tests.

runtime/Cargo.toml
```diff
@@ -59,6 +59,9 @@ opentelemetry-otlp = { workspace = true, features = ["http-proto"] }
 opentelemetry_sdk = { workspace = true, features = ["rt-tokio"] }
 tokio = { workspace = true, features = ["full"] }
 
+[target.'cfg(all(not(target_arch = "wasm32"), not(target_env = "msvc")))'.dependencies]
+tikv-jemallocator = { workspace = true, features = ["unprefixed_malloc_on_supported_platforms"], optional = true }
```
patrick-ogrady, line 63: Not sure if it is an issue that jemalloc's tests don't pass on aarch64?

Image

clabby (changes requested):
I don't think we should add this feature. Selecting the global allocator should generally be done in the terminal crate rather than an underlying library, per [`RFC-1974`](https://rust-lang.github.io/rfcs/1974-global-allocators.html#how-we-teach-this).

Since the `commonware-runtime` code is compatible with the standard library's default allocator and doesn't directly modify or extend `jemalloc`, the nuance mentioned in the rationale here should just be documented for end-user binaries to consider IMO.

PR #4059, opened 2026-06-16
[consensus::ordered-broadcast] Fix ordered broadcast stale node replay

BrendanChou:
Fixes an `ordered-broadcast` restart issue where a validator could validate an old-but-valid node before lazily replaying that sequencer's journal. If replay then replayed a higher durable tip, handling the already-validated stale node could move the tracked tip backward and panic. The PR also adds a regression test for the restart case.

This PR keeps the fix minimal: inbound node handling now checks that the sender is the chunk sequencer and that the sequencer is valid for the epoch, then prepares/replays that sequencer's journal before running any tip-dependent validation. This means stale nodes are rejected against recovered durable state instead of being accepted against empty in-memory state.

I chose this over eager replay/cleanup changes for now because it directly fixes the ordering bug without changing journal lifecycle behavior.

### DIFF
--- a/consensus/src/ordered_broadcast/engine.rs
+++ b/consensus/src/ordered_broadcast/engine.rs
@@ -401,12 +401,7 @@ impl<
                     }
                 };
 
-                if let Err(err) = self.prepare_node_journal(&node, &sender).await {
-                    debug!(?err, ?sender, "node journal prepare failed");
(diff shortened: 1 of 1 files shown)


BrendanChou:
Retiring due to deprecation of ordered-broadcast

PR #4061, opened 2026-06-17
[runtime/iobuf] compact layout

andresilva:
`IoBuf` and `IoBufMut` previously kept their hot cursor state behind enum variants and backing wrappers. This PR moves that state directly into the public handles (`ptr`, `len`, and `cap` for mutable buffers), and gives each handle a pointer to an allocation owner. That owner pointer is tagged in its low bits to identify the backing kind, so the hot `Buf`/`BufMut` paths do not need to branch over enum variants. The result is smaller handles (`IoBuf` is now 24 bytes instead of 40, `IoBufMut` is 32 bytes instead of 48) and cheaper reads/writes, since those operations are mostly pointer/length arithmetic.

The tagged owner pointer is either empty, heap, pooled, or external. Empty owners cover empty buffers and static slices. Heap owners cover native aligned allocations and adopted `Vec<u8>` allocations, with an owner header stored inline in the allocation. Pooled owners point at a side-table slot owned by the freelist. External owners hold caller-supplied `Bytes` zero-copy behind a small boxed owner. All non-empty owner types put the shared refcount at offset zero, so clone/drop can touch the refcount directly and only dispatch on the tag when a final owner has to be released.

This also removes the old `Arc<BufInner<_>>` shape from native and pooled buffers, reducing `Arc` indirection and atomic traffic around clone/drop/recovery paths. `IoBuf::try_into_mut` now recovers mutable ownership from the tagged owner when the view is unique, including uniquely-owned slices: capacity is recovered from the allocation base and current view pointer rather than from a wrapper object.

Pooled buffers now store their metadata in the freelist side table. Each `PooledOwner` slot contains the refcount sentinel, data pointer, capacity, stable slot id, and live size-class lease while the buffer is outside the global freelist. That lets pooled buffers move between checked-out views and thread-local caches without carrying separate slot/lease bundles, while still keeping the originating size class alive after the public `BufferPool` handle is dropped.

Benchmarks below were captured with `taskset -c 0 cargo bench -p commonware-runtime --bench iobuf --features=bench "iobuf::iobuf.*32.*" -- --warm-up-time 1 --measurement-time 3`.

| Benchmark | Baseline | This PR | Δ |
|---|---:|---:|---:|
| decode_fixed/mode=bytes size=32 | 15.25 ns | 15.19 ns | no change |
| decode_fixed/mode=bytes_ref size=32 | 9.38 ns | 9.40 ns | no change |
| decode_fixed/mode=iobuf_bytes size=32 | 17.62 ns | 8.49 ns | **-51.8%** |
| decode_fixed/mode=iobuf_bytes_ref size=32 | 18.99 ns | 8.14 ns | **-57.1%** |
| decode_fixed/mode=iobuf_aligned size=32 | 17.86 ns | 8.49 ns | **-52.5%** |
| decode_fixed/mode=iobuf_aligned_ref size=32 | 13.09 ns | 8.13 ns | **-37.9%** |
| decode_fixed/mode=vec_cursor size=32 | 9.26 ns | 8.38 ns | no change |
| decode_fixed/mode=vec_slice size=32 | 6.60 ns | 6.57 ns | no change |
| decode_fixed/mode=slice size=32 | 690 ps | 685 ps | no change |


runtime/src/iobuf/pool.rs
```diff
@@ -408,13 +405,15 @@ impl PoolMetrics {
...
+            // prometheus encoder appends it to counter names.
             exhausted_total: registry.register(
-                "buffer_pool_exhausted_total",
+                "buffer_pool_exhausted",
                 "Total number of failed allocations due to pool exhaustion",
                 raw::Family::default(),
             ),
             oversized_total: registry.register(
-                "buffer_pool_oversized_total",
+                "buffer_pool_oversized",
```
clabby, line 677: Good change, though let's tag this PR as https://github.com/commonwarexyz/monorepo/labels/breaking-metric

runtime/src/iobuf/owner.rs
```diff
@@ -0,0 +1,1723 @@
...
+    #[cfg(feature = "loom")]
+    pub(crate) fn assert_parked_sentinel(&self) {
+        // SAFETY: the caller just claimed the slot, so the side-table entry
+        // is live and owned by this thread.
+        let refs = unsafe { &self.owner.as_ref().refs };
+        assert_eq!(refs.load(Ordering::Relaxed), 1);
+    }
+}
+
+impl HeapOwner {
```
clabby: nit: Collapse `HeapOwner` impl blocks, there are several?

runtime/src/iobuf/owner.rs
```diff
@@ -0,0 +1,1723 @@
...
+#[repr(transparent)]
+pub(crate) struct OwnerRef(*mut ());
+
+// SAFETY: `OwnerRef` is a tagged owner pointer. Shared access to the pointed-to
+// owner state is synchronized through the owner's atomic refcount.
+unsafe impl Send for OwnerRef {}
+// SAFETY: same argument as `Send`.
+unsafe impl Sync for OwnerRef {}
+
+impl OwnerRef {
```
clabby, line 207: nit: Collapse `OwnerRef` impl blocks, there are several?

runtime/src/iobuf/freelist.rs
clabby, line 1: The non-loom test modules in `freelist.rs`, `owner.rs`, and `pool.rs` are not gated with `#[cfg(not(feature = "loom"))]`, so they compile against loom types when the `loom` feature is on but run outside `loom::model`. This would cause the regular unit tests to panic when compiled w/ `loom` enabled?

runtime/src/iobuf/freelist.rs
```diff
@@ -447,35 +436,32 @@ impl Freelist {
...
+    /// map to the same bitmap word share a single `Release` operation.
     ///
-    /// `masks` must contain the scratch word for `slot`.
+    /// `masks` must contain the scratch word for `buffer`'s slot.
     #[inline(always)]
-    fn stage_put(&self, masks: &mut [u64], slot: u32, buffer: PooledBuffer) {
-        // Park first, then stage the bit for the later per-word insert.
-        self.park(slot, buffer);
+    fn stage_put(&self, masks: &mut [u64], buffer: PooledBuffer) {
+        let slot = buffer.slot();
```
clabby, line 463: Sequential `put` calls would panic with "returned slot must not already be marked free" - should we retain that behavior here, too, so `put_batch` detects a duplicate slot? A `debug_assert_eq` here would be fine imo:

```rs
debug_assert_eq!(masks[word_index] & mask, 0, "duplicate slot in put_batch");
```

runtime/src/utils/buffer/read.rs
clabby, line 1: Idea from fable:

> In `refill`, the undersized `Ok(_)` buffer from `current.try_into_mut()` remains checked out while `self.pool.alloc(bytes_to_read)` runs, because the match scrutinee temporary lives until the end of the match expression. In a tight size class, the pool cannot reuse that slot for the replacement and may fall back to an untracked heap allocation even though the undersized pooled buffer is about to be freed.
>
> The undersized buffer's contents are not needed, so drop it before allocating the replacement:
>
> ```rust
> let buf = match current.try_into_mut() {
>    Ok(mut reusable) if reusable.capacity() >= bytes_to_read => {
>        reusable.clear();
>        reusable
>    }
>    Ok(too_small) => {
>        drop(too_small);
>        self.pool.alloc(bytes_to_read)
>    }
>    Err(_) => self.pool.alloc(bytes_to_read),
> };
> ```

runtime/src/iobuf/mod.rs
```diff
@@ -181,85 +242,110 @@ impl IoBuf {
...
+        // SAFETY: owner is unique and live.
+        let usable_capacity = unsafe { me.owner.usable_capacity() };
+        let offset = (me.ptr.as_ptr() as usize)
+            .checked_sub(base.as_ptr() as usize)
+            .expect("view pointer must be within owner allocation");
+        assert!(
+            offset <= usable_capacity,
+            "view pointer out of owner bounds"
+        );
+        let cap = usable_capacity - offset;
```
clabby, line 469: This can cause capacity to grow I believe? `IoBufMut::with_alignment(100, 4096)` would report cap=100, though `.freeze().try_into_mut()` would move to cap=104 (including the padding).

Seems safe, though a bit unexpected. `allocate_aligned_mut` could be adjusted a bit to bring these in line.

clabby (commented):
Really solid over all

runtime/src/iobuf/pool.rs
```diff
@@ -986,24 +1002,65 @@ impl TlsSizeClassCache {
...
+            // SAFETY: local cache entries keep a live lease in the pooled
+            // slot. The strong reference it owns is intentionally not
+            // released here (leases have no drop glue). The token releases
+            // below settle it.
+            let _ = unsafe { entry.buffer.take_lease() }.into_token();
+            entry.buffer
+        });
+        // Cache entries are distinct checked-out buffers from this class, and
+        // the iterator body cannot panic after yielding an entry.
+        class.global.put_batch(batch);
```
clabby, line 1331: Test gap: the park-before-release ordering in `return_global_batch` is not directly covered. Buffers must be parked in the freelist before their lease strong references are released; otherwise, the last release can drop the `SizeClass` and `put_batch` writes into a freed bitmap.

Existing coverage is incidental: tests crash at process teardown when the main thread's TLS destructor flushes a cache after the pool died, which depends on nextest's process-per-test model and platform TLS destructor behavior. The loom model `final_drop_races_pool_teardown` covers only the single-buffer `SizeClassLease::return_global` path, not the batch path used by cache drop, flush, and spill.

Verification performed in this worktree:
- On HEAD, the test below passes: `just test -p commonware-runtime test_thread_exit_batch_flush_outlives_pool` -> `1 passed`.
- With the ordering in `return_global_batch` inverted, the test fails deterministically inside the test body: `SIGTRAP [0.039s] iobuf::pool::tests::test_thread_exit_batch_flush_outlives_pool`.

```patch
index 519a6c9bc..d7b218ea4 100644
@@ -3042,6 +3042,48 @@ mod tests {
             .expect("thread-exited local buffer should be reusable");
     }
 
+    #[test]
+    fn test_thread_exit_batch_flush_outlives_pool() {
+        // The batch return path (TlsSizeClassCache::drop -> return_global_batch)
+        // must park every buffer before releasing the lease references. Drop
+        // the pool while a worker's TLS cache holds several entries so the
+        // flush's lease releases are the last strong references: the final
+        // release drops the SizeClass, whose freelist must reclaim the
+        // just-parked buffers.
+        let pool = test_pool(test_config(page, page, 8));
+
+        let (cached_tx, cached_rx) = mpsc::channel();
+        let (release_tx, release_rx) = mpsc::channel::<()>();
+        let worker_pool = pool.clone();
+        let handle = thread::spawn(move || {
+            let class_index = worker_pool
+                .class_index(page)
+                .expect("class exists for page-sized buffer");
+            let class = &worker_pool.inner.classes[class_index];
+            assert!(class.thread_cache_capacity >= MIN_TLS_BATCH_CAPACITY);
+
+            // Fill this thread's local cache so the exit flush takes the
+            // multi-entry batch path.
+            let bufs = (0..MIN_TLS_BATCH_CAPACITY)
+                .map(|_| worker_pool.try_alloc(page).expect("tracked allocation"))
+                .collect::<Vec<_>>();
+            drop(bufs);
+            assert_eq!(get_local_len(class), MIN_TLS_BATCH_CAPACITY);
+
+            drop(worker_pool);
+            cached_tx.send(()).expect("signal cached buffers");
+            release_rx.recv().expect("wait for pool drop");
+        });
+
+        cached_rx.recv().expect("worker cached buffers");
+        // Every pool handle is gone before the worker exits, so the worker's
+        // cached leases are the only remaining size-class references.
+        drop(pool);
+        release_tx.send(()).expect("release worker");
+        handle.join().expect("worker thread should exit cleanly");
+    }
+
     #[test]
     fn test_pooled_ops_inside_tls_destructor_fall_back_to_global() {
         // Pooled operations that run inside another thread_local's destructor
```
</details>

clabby (dismissed):

runtime/src/iobuf/mod.rs
```diff
@@ -611,109 +838,96 @@ impl IoBufMut {
...
-        unsafe {
-            assert!(
-                len <= self.capacity(),
-                "set_len({len}) exceeds capacity({})",
-                self.capacity()
-            );
-            match &mut self.inner {
-                IoBufMutInner::Bytes(b) => b.set_len(len),
-                IoBufMutInner::Aligned(b) => b.set_len(len),
-                IoBufMutInner::Pooled(b) => b.set_len(len),
```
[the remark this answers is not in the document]
andresilva: `test_iobuf_encode_with_pool_matches_encode` tests this.

runtime/src/utils/buffer/tip.rs
```diff
@@ -48,7 +48,7 @@ impl Buffer {
     }
 
     /// Creates a new buffer seeded with existing logical bytes.
-    pub(super) fn from(offset: u64, data: IoBuf, capacity: usize, pool: BufferPool) -> Self {
+    pub(super) const fn from(offset: u64, data: IoBuf, capacity: usize, pool: BufferPool) -> Self {
```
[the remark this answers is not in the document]
andresilva: This is a very nitpicky micro-optimization and it's not obvious what can actually be done, if we want to drop the pooled buffer early (ie. before allocating a new one), then we'd need to save the existing data somewhere, presumably into a temporary vector which requires another allocation and copy. Dismissing this.

patrick-ogrady (approved):

PR #4063, opened 2026-06-17
Remove unneeded async from conformance tests

danlaine:
No `commit` impls are `async` so the trait method doesn't need `async`


patrick-ogrady:
While not yet actively used, I prefer we keep this as is.

PR #4065, opened 2026-06-17
[storage/qmdb] Align current batch sync boundary

### LINKED ISSUE #4064: [storage/qmdb/current] MerkleizedBatch sync_boundary can diverge from Db sync_boundary before pruning
## Summary

`qmdb::current::MerkleizedBatch::sync_boundary()` and `qmdb::current::Db::sync_boundary()` can report different lower bounds for the same committed state.

The batch-side boundary is currently derived from `bitmap.pruned_chunks()`, while the DB-side boundary is derived from the chunk-aligned inactivity floor. If the inactivity floor advances past a bitmap chunk before pruning has caught up, or if pruning is disabled, a target recorded from a merkleized batch can later fail equality against the target reported by the live or reopened DB even though the root and upper bound match.

This creates an internal inconsistency between the target a batch advertises and the target the database reports after applying and reopening that same batch.

## Problem

For `qmdb::current`:

- `MerkleizedBatch::sync_boundary()` derives from `bitmap.pruned_chunks()`.
- `Db::sync_boundary()` derives from `inactivity_floor_loc / CHUNK_SIZE_BITS`.
- `glue::stateful` current DB integration records and validates sync targets with strict target equality.
- Once the inactivity floor crosses a chunk while `pruned_chunks` remains lower, the same committed prefix can produce different `range.start` values depending on whether the target came from the merkleized batch or the DB.

This can happen without application-level pruning. It is enough for the inactivity floor to advance beyond one bitmap chunk while the bitmap's physical pruning boundary remains behind.

## Why this matters

The target lower bound is load-bearing. In `glue::stateful::db::current`, callers compare recorded target starts against `batch.sync_boundary()` using strict equality, and committed targets are produced from the DB-side `sync_boundary()`. Before these two accessors agree on the same source of truth, a valid batch/DB pair can fail target validation solely because the batch boundary tracks physical pruning while the DB boundary tracks the committed inactivity floor.

Strict equality should remain intact: a target range lower bound is part of the sync availability contract. The fix should align what the batch reports with what the DB reports, not relax target comparison.

## Proposed fix

Derive `current::MerkleizedBatch::sync_boundary()` from the batch commit's inactivity floor, matching `Db::sync_boundary()`.

Current batch-side behavior:

```rust
super::db::sync_boundary::<F, N>(
    self.bitmap.pruned_chunks() as u64,
    self.inner.bounds().total_size,
)
```

Proposed behavior:

```rust
super::db::sync_boundary::<F, N>(
    *self.inner.bounds().inactivity_floor / bitmap::Prunable::<N>::CHUNK_SIZE_BITS,
    self.inner.bounds().total_size,
)
```

This mirrors the DB-side derivation and the `qmdb::any` model, where batch and DB boundaries are derived from the same committed floor state.

## Regression test

Add a deterministic test that:

- Opens a `qmdb::current` database.
- Repeatedly updates the same key across enough commits to advance the inactivity floor past one bitmap chunk.
- Does not prune.
- Records `batch.sync_boundary()` before applying the final batch.
- Applies and syncs the batch.
- Reopens the database.
- Asserts that the live and reopened `db.sync_boundary()` equal the boundary recorded from the merkleized batch.

The bug is variant-independent because `current::MerkleizedBatch::sync_boundary()` is shared by all current variants. A single current variant test should cover the shared path.

## Compatibility

This is an internal consistency fix for an ALPHA API.

It does not change:

- on-disk format
- wire format
- root computation
- target equality semantics
- pruning policy
- state sync protocol

It only changes how `current::MerkleizedBatch` reports the lower bound of the target it already exposes.

marthendalnunes:
## Summary

- derive `qmdb::current::MerkleizedBatch::sync_boundary()` from the batch commit's chunk-aligned inactivity floor
- match the boundary `qmdb::current::Db::sync_boundary()` reports once the batch is applied
- add a regression test covering the no-pruning case where the inactivity floor crosses a bitmap chunk

Closes #4064

## Motivation

`current::MerkleizedBatch::sync_boundary()` previously derived its lower bound from `bitmap.pruned_chunks()`. `current::Db::sync_boundary()` derives from the chunk-aligned inactivity floor. These can diverge when the inactivity floor advances past a chunk before physical pruning catches up, or when pruning has not run.

That makes the same committed state report different target lower bounds depending on whether the target is read from the merkleized batch or from the DB after the batch is applied. The target lower bound is load-bearing for strict sync target equality, so the two accessors should share the same source of truth rather than relaxing comparison.

## Notes

The regression test lives in `current::unordered::fixed`, but the fixed method is `current::MerkleizedBatch::sync_boundary()`, which is shared by all current variants (ordered/unordered, fixed/variable). One variant is sufficient to cover the shared path.

## Testing

- `cargo test -p commonware-storage test_merkleized_batch_sync_boundary_matches_db`
- `cargo fmt --check`
- `git diff --check`

### DIFF
--- a/storage/src/qmdb/current/batch.rs
+++ b/storage/src/qmdb/current/batch.rs
@@ -861,11 +861,11 @@ where
 
     /// Return the batch's safe sync boundary.
     ///
-    /// This derives from the chunk-aligned inactivity floor declared by this batch's commit, the
-    /// same quantity [`super::db::Db::sync_boundary`] derives from after the database is reopened.
-    /// It deliberately does not derive from the physical bitmap pruning boundary, which can lag the
-    /// inactivity floor when pruning has not run.
+    /// This equals the boundary [`super::db::Db::sync_boundary`] reports once this batch is applied.
     pub fn sync_boundary(&self) -> Location<F> {
+        // Derive from the commit's chunk-aligned inactivity floor, the same quantity the DB uses
+        // after apply. Deliberately not the physical bitmap pruning boundary, which can lag the
+        // inactivity floor when pruning has not run.
         super::db::sync_boundary::<F, N>(
             *self.inner.bounds().inactivity_floor / bitmap::Prunable::<N>::CHUNK_SIZE_BITS,
             self.inner.bounds().total_size,
--- a/storage/src/qmdb/current/batch.rs
+++ b/storage/src/qmdb/current/batch.rs
@@ -860,9 +860,14 @@ where
     }
 
     /// Return the batch's safe sync boundary.
+    ///
+    /// This derives from the chunk-aligned inactivity floor declared by this batch's commit, the
+    /// same quantity [`super::db::Db::sync_boundary`] derives from after the database is reopened.
+    /// It deliberately does not derive from the physical bitmap pruning boundary, which can lag the
+    /// inactivity floor when pruning has not run.
     pub fn sync_boundary(&self) -> Location<F> {
         super::db::sync_boundary::<F, N>(
-            self.bitmap.pruned_chunks() as u64,
+            *self.inner.bounds().inactivity_floor / bitmap::Prunable::<N>::CHUNK_SIZE_BITS,
             self.inner.bounds().total_size,
         )
     }
--- a/storage/src/qmdb/current/unordered/fixed.rs
+++ b/storage/src/qmdb/current/unordered/fixed.rs
@@ -252,6 +252,53 @@ pub mod test {
         });
     }
 
+    /// The sync boundary recorded from a merkleized batch must match the boundary recovered by the
+    /// database after reopening. This can diverge if the batch boundary is derived from physical
+    /// bitmap pruning rather than the batch's declared inactivity floor, because the floor can
+    /// advance past a chunk even when pruning has not run.
+    #[test_traced("INFO")]
+    pub fn test_merkleized_batch_sync_boundary_survives_reopen() {
+        deterministic::Runner::default().start(|ctx| async move {
+            let partition = "batch-boundary-reopen".to_string();
+            let mut db = open_db(ctx.child("current"), partition.clone()).await;
+
+            let key = Sha256::fill(1u8);
+            let mut last_batch_boundary = mmr::Location::new(0);
+            for i in 0..300u64 {
+                let value = Sha256::hash(&i.to_be_bytes());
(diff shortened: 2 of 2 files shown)


storage/src/qmdb/current/unordered/fixed.rs
```diff
@@ -252,6 +252,53 @@ pub mod test {
         });
     }
 
+    /// The sync boundary recorded from a merkleized batch must match the boundary recovered by the
+    /// database after reopening. This can diverge if the batch boundary is derived from physical
```
roberto-bayardo: Isn't the important thing that the boundary matches that of the database after the batch is applied, not after the database is reopened? (close/open shouldn't affect the sync boundary at all).

storage/src/qmdb/current/batch.rs
```diff
@@ -860,9 +860,14 @@ where
     }
 
     /// Return the batch's safe sync boundary.
+    ///
+    /// This derives from the chunk-aligned inactivity floor declared by this batch's commit, the
+    /// same quantity [`super::db::Db::sync_boundary`] derives from after the database is reopened.
+    /// It deliberately does not derive from the physical bitmap pruning boundary, which can lag the
+    /// inactivity floor when pruning has not run.
```
roberto-bayardo: Put implementation details like these inside the function rather than in the header.  Also see comment above about the value matching after applying the batch being the important thing, not reopening.

marthendalnunes: Done in da8f21d. Moved the implementation rationale out of the doc header into a comment in the function body, and the header now states the contract: the boundary equals what `Db::sync_boundary` reports once the batch is applied (no longer references reopen).

storage/src/qmdb/current/unordered/fixed.rs
marthendalnunes: Agreed. In da8f21d the test is renamed to `test_merkleized_batch_sync_boundary_matches_db` and now leads with the post-apply equality (batch boundary == live DB boundary after applying the batch) as the headline assertion. Reopen is kept afterward purely as a persistence sanity check, with a comment noting it must not move the boundary.

roberto-bayardo (approved):

patrick-ogrady (approved):

christopherwxyz:
We did it!

PR #4066, opened 2026-06-17
[storage/merkle/persisted] Remove `Inner` state and `RwLock` around it

danlaine:


danlaine:
applied directly to #3970

PR #4073, opened 2026-06-18
[consensus] Cleanup `Commitment` impl of `Codec`

patrick-ogrady:
## Acknowledgement

[V12](https://github.com/v12-security) reported that a user that provided a fallible `Digest` could see an unexpected panic when accessing commitment (which only checked length during `read_cfg`). I'm not aware of any such `Digest` impl but figured this was do for an overhaul nonetheless to align closer to our standard use of `codec`.


consensus/src/types.rs
```diff
@@ -720,118 +720,116 @@ commonware_macros::stability_scope!(ALPHA {
...
-            ///
-            /// Panics if the [`Digest`]'s [`FixedSize::SIZE`] is > 32 bytes.
-            fn take<D: Digest>(&self, range: Range<usize>) -> D {
-                const {
-                    assert!(
-                        D::SIZE <= 32,
-                        "Cannot extract Digest with size > 32 from Commitment"
-                    );
-                }
+            fn read_digest<D: Digest>(
```
patrick-ogrady: this is no longer necessary? we assert validity during parsing already

consensus/src/types.rs
```diff
@@ -720,118 +720,116 @@ commonware_macros::stability_scope!(ALPHA {
...
-        use core::{
-            num::NonZeroU16,
-            ops::{Deref, Range},
-        };
+        use core::{marker::PhantomData, num::NonZeroU16, ops::Deref};
         use rand_core::CryptoRngCore;
 
+        // Ideally, the backing array would be `[u8; Self::SIZE]`, but `Self::SIZE`
+        // depends on generic digest associated constants. Stable Rust does not allow
+        // those generic const expressions in array lengths, so keep a max-sized backing
```
patrick-ogrady: max pain...

Issue #4013, opened 2026-06-10
[marshal] Only Ensure Durability at Certify


patrick-ogrady:
It is possible that our insistence that a block is durable before propose/verify return is overkill and we may only need to guarantee this before certify returns. This could allow us to hide write/fsync latency much better during the e2e block processing pipeline (particularly kicking off writes/fsync and only checking it is complete at the certify boundary).

patrick-ogrady:
(from `fable 5`)

# Assessment: Issue #4013 — Only Ensure Durability at Certify

> Evaluation of the claim: "our insistence that a block is durable before propose/verify
> return is overkill and we may only need to guarantee this before certify returns."

## Verdict

**The claim is true and safe at the protocol level.** Durability before the `certify`
receiver resolves `true` is the only payload-durability invariant simplex actually needs.
Durability at `propose`/`verify` is conservative, not load-bearing for safety. The coding
marshal (and the standard deferred mode) already cast notarize votes with zero local
payload durability, so extending this to all wrappers is an extension of an
already-shipped pattern, not a new safety regime.

Four adversarial verification passes (simplex liveness, leader recovery, safety
refutation, RAM serving paths) found no safety break. All consequences of deferring
durability are liveness-only (rare nullified views), bounded by machinery that already
exists and is already tested.

## What ships today (per wrapper)

| Hook | Standard inline | Standard deferred | Coding |
|---|---|---|---|
| `propose` | fsync before digest (`inline.rs:368`) | fsync before digest (`deferred.rs:604`) | fsync before commitment (`marshaled.rs:827`) |
| `verify` (gates notarize vote) | **fsync before true** (`validation.rs:201`) | context-check only, no fsync (`deferred.rs:646-728`) | **RAM-only shard validity** (`marshaled.rs:1000-1013`); block often not yet reconstructed |
| `certify` (gates finalize vote) | durable before true (`inline.rs:527-534`) | durable before true (awaits background `Stage::Verified` store) | durable before true (awaits background `Stage::Verified` store) |

Mechanics: `Mailbox::proposed/verified/certified` ack only after the core actor's
`cache.put_verified`/`put_block` (`core/actor.rs:572-613`), each of which is a
`put_sync` = put + fsync into the prunable archive (`core/cache.rs:300-352`,
`storage/src/archive/mod.rs:67-77`). There is no group commit on these per-block syncs.
The shard engine is entirely RAM (`shards/engine.rs:286-348`, no `Storage` bound).

[...]

PR #4078, opened 2026-06-18
[runtime] Add `Blob::start_sync`

### LINKED ISSUE #4076: [runtime/blob] Add `start_sync` function
Add function that allows caller to start an `fsync` without `await`ing its completion

danlaine:
Blocks: #4013 

Adds new fn that starts an `fsync` but doesn't await completion

### DIFF
--- a/runtime/src/storage/audited.rs
+++ b/runtime/src/storage/audited.rs
@@ -1,4 +1,4 @@
-use crate::{deterministic::Auditor, Error, IoBufs, IoBufsMut, Handle};
+use crate::{deterministic::Auditor, Error, Handle, IoBufs, IoBufsMut};
 use sha2::digest::Update;
 use std::sync::Arc;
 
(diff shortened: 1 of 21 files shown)


runtime/src/storage/audited.rs
```diff
@@ -175,6 +184,47 @@ mod tests {
         run_storage_tests(storage).await;
     }
 
+    #[tokio::test]
+    async fn test_audited_start_sync() {
+        use crate::deterministic::Auditor;
```
patrick-ogrady: this should be imported at the top

runtime/src/storage/tokio/unix.rs
```diff
@@ -210,4 +211,13 @@ impl crate::Blob for Blob {
             .map_err(|e| Error::BlobSyncFailed(self.partition.clone(), hex(&self.name), e))?;
         Ok(())
     }
+
+    fn start_sync(&self) -> oneshot::Receiver<Result<(), Error>> {
```
patrick-ogrady: I had originally intended that we'd actually modify the existing sync calls and return a oneshot there to avoid this double task wrap?

runtime/src/storage/metered.rs
```diff
@@ -223,6 +224,17 @@ impl<B: crate::Blob> crate::Blob for Blob<B> {
         self.metrics.storage_syncs.inc();
         self.inner.sync().await
     }
+
+    #[tracing::instrument(
+        name = "runtime.storage.blob.start_sync",
```
patrick-ogrady, line 228: In an ideal world, we'd trace the duration of the inner fsync call as well but doing so would breach the current barrier we have with tracing in metered.

patrick-ogrady: I decided to bring about this world lol.

patrick-ogrady:
bugbot run

patrick-ogrady (approved):

PR #4081, opened 2026-06-19
[storage] Split Reader/Writer

patrick-ogrady:
Blocked: #4070

## Summary

Splits contiguous journals into an append/prune `Writer` and an independent `Readers` handle. After `split`, destructive repair/reset operations stay on the unsplit `Journal`, while the split writer only supports steady-state append, prune, commit, and sync paths.

Readers clone the latest published snapshot instead of borrowing the writer, so new readers can be created while the writer is active. They observe either the previous complete state or the newly published state, never a partial append.

## Implementation

- Adds a small published-state wrapper backed by `ArcSwap`.
- Adds `Journal::split()` for fixed and variable contiguous journals.
- Updates fixed readers to hold `Arc<Snapshot<_>>`.
- Publishes variable journal data and offset snapshots as one combined state to avoid mismatched generations.
- Re-exports `ArcSwap` from `commonware-utils::sync` for this read-mostly published-state pattern.

## Validation

- `just test -p commonware-storage contiguous`
- `cargo clippy --all-targets -p commonware-utils -p commonware-storage -- -D warnings`
- `just check-stability -p commonware-utils -p commonware-storage`
- `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --document-private-items -p commonware-utils -p commonware-storage`


storage/src/journal/contiguous/variable.rs
```diff
@@ -2075,6 +2225,40 @@ mod tests {
     const LARGE_PAGE_SIZE: NonZeroU16 = NZU16!(1024);
     const SMALL_PAGE_SIZE: NonZeroU16 = NZU16!(512);
 
+    #[test_traced]
+    fn test_variable_split_reader_while_append_future_exists() {
```
patrick-ogrady: I don't think this test is necessary

storage/src/journal/contiguous/fixed.rs
```diff
@@ -1452,6 +1590,33 @@ mod tests {
         format!("{}-blobs", cfg.partition)
     }
 
+    #[test_traced]
+    fn test_fixed_split_reader_while_append_future_exists() {
```
patrick-ogrady: I don't think this is necessary?

PR #4090, opened 2026-06-21
[storage] Make `bmt` no_std

### LINKED ISSUE #4089: storage: make `bmt` available under `no_std` (it has no `std` dependency)
### Summary

`storage::bmt` is gated behind `cfg(feature = "std")`, but its implementation appears to have no actual `std` requirement. The sibling Merkle structures (`merkle`, `mmr`, `mmb`) are already available under `no_std`; `bmt` looks like it could be too, with a small change.

### Observation

In `storage/src/lib.rs`, the no-std `ALPHA` scope exposes the Merkle family:

```rust
commonware_macros::stability_scope!(ALPHA {
    extern crate alloc;
    pub mod merkle;
    pub use merkle::{mmb, mmr};
});
```

…while `bmt` is grouped into the `std`-gated scope with the storage engines:

```rust
commonware_macros::stability_scope!(ALPHA, cfg(feature = "std") {
    mod bitmap;
    pub mod qmdb;
    pub mod bmt;   // <- here
    pub mod cache;
    pub mod queue;
});
```

But `src/bmt/` contains no `std::` usage — it is `alloc`-only (`alloc::collections::btree_set::BTreeSet`) and otherwise depends only on `commonware-codec`, `commonware-cryptography`, and `commonware-utils`, all of which build with `default-features = false`.

The single thing that ties `bmt` to `std` is one import (`src/bmt/mod.rs:47`):

```rust
use commonware_runtime::{Buf, BufMut};
```

`commonware-runtime` is the heavyweight async runtime crate (pulled in by `storage`'s `std` feature alongside `futures` and `zstd`). However, those two traits are a **re-export of `bytes`** (`runtime/src/lib.rs:47`):

```rust
/// Re-export of `Buf` and `BufMut` traits for usage with [I/O buffers](iobuf).
pub use bytes::{Buf, BufMut};
```

`bytes` is already a direct `commonware-storage` dependency (`Cargo.toml`, `default-features = false`). So `bmt`'s dependency on `commonware-runtime` is incidental — used solely for these two re-exported `bytes` traits — and is what forces it under `std`.

### Proposed change

1. In `src/bmt/mod.rs`, import the traits from `bytes` directly:
   ```rust
   use bytes::{Buf, BufMut};
   ```
   removing `bmt`'s only tie to `commonware-runtime`.
2. Move `pub mod bmt;` from the `cfg(feature = "std")` scope into the no-std `ALPHA` scope (next to `merkle`).

### Why

`bmt` is a stateless, fixed-leaf-set Binary Merkle Tree with an in-tree single/range/multi proof verifier — exactly the shape that suits `no_std`/embedded/light-client verification. Today, depending on it forces the full `std` feature, which transitively pulls `commonware-runtime`, `futures`, and the `zstd` C library (a non-trivial, C-toolchain-requiring surface) for what is otherwise pure `alloc` compute. Making it `no_std` brings it in line with the already-no-std `merkle`/`mmr`.

### Question

Is the `std` gate on `bmt` deliberate (something not visible from the source), or incidental grouping with the storage engines? If incidental, happy to open a PR with the two changes above.

christopherwxyz:
## Summary

`storage::bmt` is a pure Binary Merkle Tree over `Digest`/`Hasher` — no runtime or I/O dependency — yet it sits in the `cfg(feature = "std")` stability scope alongside the I/O-bound modules (`qmdb`, `cache`, `queue`). Its sibling Merkle primitives (`merkle`, `mmr`, `mmb`) already live in the no_std `ALPHA` scope. This moves `bmt` there too.

## What tied `bmt` to std

A single import: `use commonware_runtime::{Buf, BufMut}`. `commonware-runtime` re-exports both verbatim from `bytes` (`pub use bytes::{Buf, BufMut}`), and `bytes` is already a `default-features = false` dependency of `commonware-storage`. Importing them from `bytes` directly uses identical types and adds no new dependency, while dropping the `commonware-runtime` (hence std/I/O) coupling.

## Changes

- `storage/src/bmt/mod.rs`: import `Buf`/`BufMut` from `bytes`; add the standard `alloc` prelude imports (`String`, `ToString`, `vec`, `Vec`) the module already relies on.
- `storage/src/lib.rs`: move `pub mod bmt;` from the `cfg(feature = "std")` scope into the no_std `ALPHA` scope next to `merkle`/`mmr`/`mmb`.

No public API or behavior change; the std build is unaffected.

## Verification

- `cargo build -p commonware-storage --no-default-features` — compiles under `#![no_std]`.
- `cargo test -p commonware-storage --lib bmt::` — 69 tests pass.
- `cargo build -p commonware-storage` + `cargo clippy -p commonware-storage --all-targets` — std build and lints clean (warnings as errors).
- `cargo fmt -p commonware-storage -- --check` — clean.

Closes #4089

### DIFF
--- a/storage/src/bmt/mod.rs
+++ b/storage/src/bmt/mod.rs
@@ -42,10 +42,15 @@
 //! assert!(proof.verify_element_inclusion(&mut hasher, &digests[1], 1, &root).is_ok());
 //! ```
 
-use alloc::collections::btree_set::BTreeSet;
+use alloc::{
+    collections::btree_set::BTreeSet,
+    string::{String, ToString},
+    vec,
+    vec::Vec,
+};
+use bytes::{Buf, BufMut};
 use commonware_codec::{EncodeSize, Read, ReadExt, ReadRangeExt, Write};
 use commonware_cryptography::{Digest, Hasher};
-use commonware_runtime::{Buf, BufMut};
 use commonware_utils::{non_empty_vec, vec::NonEmptyVec};
 use thiserror::Error;
 
--- a/storage/src/lib.rs
+++ b/storage/src/lib.rs
@@ -13,6 +13,7 @@
 commonware_macros::stability_scope!(ALPHA {
     extern crate alloc;
 
+    pub mod bmt;
     pub mod merkle;
     pub use merkle::{mmb, mmr};
 });
@@ -20,7 +21,6 @@ commonware_macros::stability_scope!(ALPHA, cfg(feature = "std") {
     mod bitmap;
     pub mod qmdb;
     pub use crate::bitmap::{BitMap as AuthenticatedBitMap, MerkleizedBitMap, UnmerkleizedBitMap};
-    pub mod bmt;
     pub mod cache;
     pub mod queue;
     #[cfg(any(test, feature = "test-utils"))]


patrick-ogrady (approved):
nice catch!

PR #4091, opened 2026-06-22
[coding/rs] Parallelize Reed-Solomon recovery across symbol stripes + Skip Encode

a-qedaudit:
## Summary

`reed_solomon::decode` doesn't fully use the caller's concurrency when recovery is needed: `Strategy` parallelizes the shard hashing / BMT build, but the Reed-Solomon recovery itself is one sequential `reed_solomon_simd` `decoder.decode()`.

`reed-solomon-simd` is GF(2^16), independent per 16-bit symbol column, so each shard splits into even, 64-byte-aligned symbol stripes that are independent RS instances. This decodes and re-encodes the stripes in parallel through the existing `Strategy` and reassembles the shards. Shards below `MIN_STRIPE_BYTES`, and `Sequential`, take the existing single-shard path. The reconstructed codeword is still verified against the committed BMT root, so striping cannot change the result (a mismatch returns `Error::Inconsistent`).

When data shards are missing, the decoder is fed exactly `k` shards (any surplus chunks beyond the minimum are trimmed) and reveals the reconstructed recovery shards directly from the decode (its FFT already evaluates the codeword at every position), so recovery skips the separate re-encode entirely. Trimmed surplus chunks are rebuilt from the reconstruction and bound by the BMT root check rather than compared against a canonical re-encode, so the security argument is unchanged. Re-encoding remains only when all `k` data shards are already present (nothing to decode).

## Benchmark

`coding_scheme_times` `reed_solomon::decode`, 8 MiB / 250 chunks, recovery (`worst`) selection, Criterion point estimates (`sample_size = 10`), Apple M5 Pro. Baseline is `main` at `772f1773d` (already vendors `reed-solomon-simd` via #4092), so the delta isolates this PR.

| `worst` (recovery) | main | this PR |
|---|---:|---:|
| conc=8 | 26.2 ms | 8.5 ms (**~3.1x**) |
| conc=16 | 24.8 ms | 6.1 ms (**~4.1x**) |

The speedup grows with cores: `main`'s Reed-Solomon recovery is sequential, so more threads barely help it (26.2 -> 24.8 ms from conc=8 to 16), whereas this PR parallelizes it. The no-recovery (`best`) case also improves: conc=8 9.0 -> 5.5 ms (~1.6x), conc=16 7.6 -> 3.7 ms (~2.1x). `conc=1` is unchanged (no striping at parallelism 1).

Of the `worst`-case gain, the decode-reveal fast path contributes ~14% (conc=8) to ~16% (conc=16) on top of striping, by eliminating the separate re-encode.

Safety: striping fills shared output buffers from parallel stripe tasks through a small `unsafe` helper (`striped::Dst`) over disjoint stripe/shard slots (covered by a single-threaded miri test), and the vendored `restored_recovery` reveal is checked byte-for-byte against the encoder across partial-chunk shard sizes and both rate selections.

## Testing

- `test_striped_recovery_matches_sequential`: striped decode is byte-identical to sequential across payload sizes and shard counts spanning stripe-count boundaries, at parallelism 2 and 8.
- Existing `reed_solomon` recovery, mismatched-length, and minifuzz round-trip tests pass.
- `cargo fmt --check`, `cargo clippy -p commonware-coding --all-targets -- -D warnings`.


coding/src/reed_solomon.rs
```diff
@@ -424,6 +856,9 @@ fn decode<'a, H: Hasher, S: Strategy>(
         if &chunk.root != root {
             return Err(Error::CommitmentMismatch);
         }
+        if chunk.shard.len() != shard_len {
+            return Err(Error::Inconsistent);
+        }
         // Check for duplicate index
```
patrick-ogrady, line 966: add a \n before

coding/src/reed_solomon.rs
```diff
@@ -447,87 +881,25 @@ fn decode<'a, H: Hasher, S: Strategy>(
         return Err(Error::NotEnoughChunks);
     }
 
-    // Decode original data
```
patrick-ogrady, line 452: retain comment

coding/src/reed_solomon.rs
```diff
@@ -447,87 +881,25 @@ fn decode<'a, H: Hasher, S: Strategy>(
...
+    };
+    if let Some(ranges) = stripes {
+        return decode_striped::<H, S>(
+            &ctx,
+            ranges,
+            shard_digests,
+            provided_originals,
+            provided_recoveries,
+        );
     }
```
patrick-ogrady: remove \n

parallel/src/lib.rs
```diff
@@ -581,6 +696,42 @@ commonware_macros::stability_scope!(BETA, cfg(feature = "std") {
...
 
+        fn map_collect_vec<I, F, T>(&self, iter: I, map_op: F) -> Vec<T>
+        where
+            I: IntoIterator<IntoIter: Send, Item: Send> + Send,
+            F: Fn(I::Item) -> T + Send + Sync,
+            T: Send,
+        {
+            self.thread_pool.install(|| {
+                let items: Vec<I::Item> = iter.into_iter().collect();
+                items.into_par_iter().map(map_op).collect()
```
patrick-ogrady, line 707: This is meaningfully faster than our old fold approach in the common case

coding/src/reed_solomon.rs
```diff
@@ -16,6 +16,14 @@ use thiserror::Error;
 commonware_utils::thread_local_cache!(static CACHED_ENCODER: ReedSolomonEncoder);
 commonware_utils::thread_local_cache!(static CACHED_DECODER: ReedSolomonDecoder);
 
+// Keep each stripe large enough to amortize extra encoder/decoder setup
+const MIN_STRIPE_BYTES: usize = 8 * 1024;
+// Non-final stripe widths must be a multiple of reed-solomon-simd's internal 64-byte SIMD
```
patrick-ogrady: link to permalink (like we do elsewhere)

coding/src/reed_solomon.rs
```diff
@@ -380,6 +408,410 @@ fn encode<H: Hasher, S: Strategy>(
...
+//
+// Decode uses the same layout in reverse: recover only missing original
+// stripes, re-encode recovery stripes from the reconstructed originals,
+// compare provided recoveries against those stripes, and hash missing
+// recoveries without first materializing full recovery shards
+fn striped_ranges(shard_len: usize, parallelism: usize) -> Option<Vec<Range<usize>>> {
+    // Non-final stripes must end on aligned boundaries because reed-solomon-simd
+    // internally lays shards out in aligned symbol blocks with a special final tail
+    let full_blocks = shard_len / STRIPE_ALIGNMENT_BYTES;
+    // Bound the stripe count by available parallelism, the number of MIN_STRIPE_BYTES
```
patrick-ogrady: add a \n before this

coding/src/reed_solomon.rs
```diff
@@ -380,6 +408,410 @@ fn encode<H: Hasher, S: Strategy>(
...
+    m: usize,
+    /// Width of every shard, in bytes.
+    shard_len: usize,
+    /// Commitment that the reconstructed codeword must reproduce.
+    root: &'a H::Digest,
+    /// Parallelism strategy.
+    strategy: &'a S,
+}
+
+fn decode_missing_original_stripe(
```
patrick-ogrady: these internal functions should still have a comment

coding/src/reed_solomon.rs
patrick-ogrady: is this check new/was it a bug to not have it before?

coding/src/reed_solomon.rs
```diff
@@ -329,27 +332,40 @@ fn encode<H: Hasher, S: Strategy>(
...
+            for shard in padded.chunks(shard_len) {
+                encoder
+                    .add_original_shard(shard)
+                    .map_err(Error::ReedSolomon)?;
+            }
+
+            // Compute recovery shards and collect into a contiguous buffer
+            let encoding = encoder.encode().map_err(Error::ReedSolomon)?;
+            let mut buf = vec![0u8; m * shard_len];
+            for i in 0..m {
```
patrick-ogrady: I think the recovery_iter() we had here before was cleaner?

coding/src/reed_solomon.rs
```diff
@@ -382,6 +398,460 @@ fn encode<H: Hasher, S: Strategy>(
...
+        });
+    let tree = builder.build();
+    if tree.root() != *root {
+        return Err(Error::Inconsistent);
+    }
+
+    Ok(data)
+}
+
+fn verify_sequential_codeword<H: Hasher, S: Strategy>(
```
patrick-ogrady: Should we have a `striped` and `sequential` mods and then define those functions on them? I think that would be a lot clearer than just using function names to sort out what's going on.

patrick-ogrady (commented):
very close now

coding/src/reed_solomon.rs
patrick-ogrady: Added the blank line. On the follow-up question: this check is new and not a pre-existing bug. All shards are the same width (`prepare_data` zero-pads to a uniform `shard_len`, and the recovery shards match), so `len() != shard_len` is only ever malformed/adversarial input. The old code always fed every provided shard straight to the RS decoder, which rejects a mismatched width itself (`Error::DifferentShardSize`), so it was already caught — just as a `ReedSolomon` error. The new striped path slices each shard by stripe range (`shard[range]`), so a short shard would panic out-of-bounds instead; the explicit check makes rejection uniform (`Inconsistent`) and panic-free across both paths. Covered by `test_parallel_decode_rejects_mismatched_shard_lengths`.

coding/src/reed_solomon.rs
patrick-ogrady: Retained — the `// Decode original data` comment is back on the decoder in `sequential::decode`.

coding/src/reed_solomon.rs
patrick-ogrady: Removed the blank line before the sequential dispatch.

parallel/src/lib.rs
patrick-ogrady: Agreed — kept the direct `into_par_iter().map().collect()` override for that reason.

coding/src/reed_solomon.rs
patrick-ogrady: Since reed-solomon-simd is now vendored in-repo, I linked the stripe-alignment invariant to the vendored constant [`SHARD_CHUNK_BYTES`] (`cryptography/src/reed_solomon/engine.rs`) rather than an external permalink — that's where the 64-byte symbol-block width is now defined and documented.

coding/src/reed_solomon.rs
patrick-ogrady: Done — after the refactor the stripe-count bound comment (now in `striped::ranges`) has a blank line before it.

coding/src/reed_solomon.rs
patrick-ogrady: Done — every stripe helper now has a doc comment, and they live in the `striped` / `sequential` modules.

coding/src/reed_solomon.rs
patrick-ogrady: Reverted — the encode paths use `recovery_iter()` again, both the sequential encode and the striped `encode_recovery` helper.

coding/src/reed_solomon.rs
patrick-ogrady: Done. Added `mod striped` and `mod sequential`. The striping logic (`ranges`, the encode/decode stripe helpers, `Dst`, `decode`, `verify_codeword`) lives in `striped` with a module-doc diagram; the single-instance path lives in `sequential`; the top-level `decode` just dispatches between them.

coding/src/reed_solomon.rs
```diff
@@ -382,6 +396,514 @@ fn encode<H: Hasher, S: Strategy>(
     Ok((root, chunks))
 }
 
+/// Read-only parameters shared across the decode helpers.
```
patrick-ogrady: rewrite this comment

coding/src/reed_solomon.rs
```diff
@@ -382,6 +396,514 @@ fn encode<H: Hasher, S: Strategy>(
...
+/// [`decode`](striped::decode) uses the same layout in reverse: recover only missing
+/// original stripes, re-encode recovery stripes from the reconstructed originals, compare
+/// provided recoveries against those stripes, and hash missing recoveries without first
+/// materializing full recovery shards.
+mod striped {
+    use super::*;
+
+    /// Target for writing stripes into a shared, strided output buffer in parallel.
+    ///
+    /// Holds the buffer's base address as a `usize` so disjoint stripe tasks can write the
```
patrick-ogrady: remove this extra comment block

coding/src/reed_solomon.rs
```diff
@@ -382,6 +396,514 @@ fn encode<H: Hasher, S: Strategy>(
...
+
+    /// Split a shard of `shard_len` bytes into disjoint stripe ranges, or return `None`
+    /// when striping would not help (too little parallelism or data).
+    ///
+    /// Every non-final stripe ends on a [`SHARD_CHUNK_BYTES`] boundary: the engine lays
+    /// shards out in symbol blocks of that width (a partial final block is padded
+    /// internally), so a boundary in the middle of a block would change the bytes each
+    /// sub-instance encodes.
+    pub(super) fn ranges(shard_len: usize, parallelism: usize) -> Option<Vec<Range<usize>>> {
+        let full_blocks = shard_len / SHARD_CHUNK_BYTES;
```
patrick-ogrady: move this under the comment block below

coding/src/reed_solomon.rs
```diff
@@ -382,6 +396,514 @@ fn encode<H: Hasher, S: Strategy>(
...
+        for shard in encoding.recovery_iter() {
+            recoveries.extend_from_slice(shard);
+        }
+
+        Ok((range, recoveries))
+    }
+
+    /// Encode the recovery shards for a single stripe, writing each recovery shard's
+    /// stripe into `dst` (one slot per recovery shard).
+    pub(super) fn encode_recovery_into(
```
patrick-ogrady, line 566: do we need both "_into" and not into variants of functions? I think we can just make "_into" the canonical functions?

coding/src/reed_solomon.rs
```diff
@@ -449,87 +977,24 @@ fn decode<'a, H: Hasher, S: Strategy>(
...
-        .map(|digest| digest.expect("digest must be present for every shard"))
-        .for_each(|digest| {
-            builder.add(&digest);
-        });
-    let tree = builder.build();
-
-    // Confirm root is consistent
-    if tree.root() != *root {
-        return Err(Error::Inconsistent);
+    let ctx = DecodeCtx {
```
patrick-ogrady, line 1004: add a comment above this ("Decode...")

coding/src/reed_solomon.rs
```diff
@@ -974,6 +1439,320 @@ mod tests {
...
+                        striped, data,
+                        "striped decode mismatch (len={data_len} total={total} min={min} parallelism={parallelism})"
+                    );
+                }
+            }
+        }
+    }
+
+    /// Single-threaded exercise of the striped encode helper and its [`striped::Dst`]
+    /// raw-pointer writes, independent of any [`Strategy`]. The rayon-driven striped tests
```
patrick-ogrady: Remove "The rayon-driven striped tests
    /// cannot run under miri (crossbeam's work-stealing deque relies on integer-to-pointer
    /// casts that abort miri during threadpool setup), so this drives the same `unsafe`
    /// writes sequentially to keep them miri-checkable: splitting a shard into stripes and
    /// encoding each must reproduce the single full-width encode byte-for-byte."

patrick-ogrady (commented):
final comments

coding/src/reed_solomon.rs
patrick-ogrady: Rewritten to a single line: "Read-only parameters threaded through the striped and sequential decode paths." (dropped the argument-count/threshold note).

coding/src/reed_solomon.rs
patrick-ogrady: Trimmed to the summary plus the one-line reason the base address is a `usize` (so it's `Send` across tasks); dropped the redundant layout description, which is already covered by `Dst::write`'s doc.

coding/src/reed_solomon.rs
patrick-ogrady: Moved `let full_blocks` down under the stripe-count-bound comment, next to where it's used.

coding/src/reed_solomon.rs
patrick-ogrady: Done — dropped the non-`_into` `encode_recovery`. The striped decode now re-encodes into one contiguous `m * shard_len` buffer via `encode_recovery_into` (same helper as encode), which also let me delete the per-stripe vecs and the `provided_recovery_by_idx` bookkeeping.

coding/src/reed_solomon.rs
patrick-ogrady: Added: `// Decode the data, striping the work across the strategy when shards are large enough.`

coding/src/reed_solomon.rs
patrick-ogrady: Removed the crossbeam/miri explanation; the doc now just describes the byte-for-byte check (and notes it stays miri-checkable).

coding/src/reed_solomon.rs
```diff
@@ -382,6 +396,464 @@ fn encode<H: Hasher, S: Strategy>(
     Ok((root, chunks))
 }
 
+/// Read-only parameters threaded through the striped and sequential decode paths.
```
patrick-ogrady: This still sounds awkward.

coding/src/reed_solomon.rs
```diff
@@ -449,87 +927,25 @@ fn decode<'a, H: Hasher, S: Strategy>(
...
-        .map(|digest| digest.expect("digest must be present for every shard"))
-        .for_each(|digest| {
-            builder.add(&digest);
-        });
-    let tree = builder.build();
-
-    // Confirm root is consistent
-    if tree.root() != *root {
-        return Err(Error::Inconsistent);
+    // Decode the data, striping the work across the strategy when shards are large enough.
```
patrick-ogrady: remove period

patrick-ogrady (commented):
more nits

coding/src/reed_solomon.rs
```diff
@@ -382,6 +396,464 @@ fn encode<H: Hasher, S: Strategy>(
...
+        /// Callers must ensure no two concurrent writes overlap. Stripe `range`s are disjoint
+        /// and shard slots are disjoint, so concurrent tasks never alias. `index` and `range`
+        /// must stay within the buffer the [`Dst`] was built from, and `src.len()` must
+        /// equal `range.len()`.
+        unsafe fn write(&self, index: usize, range: &Range<usize>, src: &[u8]) {
+            let start = index * self.shard_len + range.start;
+            let out = std::slice::from_raw_parts_mut((self.ptr as *mut u8).add(start), range.len());
+            out.copy_from_slice(src);
+        }
+    }
```
cronokirby: This seems pretty error prone to me compared to using split_at_mut to create genuine slices.

coding/src/reed_solomon.rs
```diff
@@ -382,6 +396,464 @@ fn encode<H: Hasher, S: Strategy>(
...
+            return Err(Error::Inconsistent);
+        }
+
+        Ok(data)
+    }
+}
+
+/// Sequential Reed-Solomon: recover and re-encode the codeword as a single
+/// Reed-Solomon instance, used when striping would not help.
+mod sequential {
```
cronokirby, line 806: Can we avoid a second codepath?

coding/src/reed_solomon.rs
patrick-ogrady: Reworded to `Inputs shared by the decode helpers` (and dropped `Read-only` per your follow-up).

coding/src/reed_solomon.rs
patrick-ogrady: Removed the trailing period.

coding/src/reed_solomon.rs
patrick-ogrady: Done. Replaced the `Dst` raw-pointer helper with `stripe_columns`, which uses `chunks_mut` + `split_at_mut` to hand each parallel stripe task genuine, provably-disjoint `&mut [u8]` column slices (the borrow checker enforces the disjointness). The striped encode/recover paths now fill the shared output buffers with no `unsafe` — the module is unsafe-free. No measurable perf change (worst-case 8 MiB / 250 decode is unchanged within run-to-run noise).

coding/src/reed_solomon.rs
patrick-ogrady: Prefer leaving this as-is (I intentionally separated the code this way).

coding/src/reed_solomon.rs
```diff
@@ -974,6 +1487,316 @@ mod tests {
         assert_eq!(decoded, data);
     }
 
+    #[test]
+    fn test_parallel_encode_matches_sequential() {
+        let strategy = Rayon::new(NZUsize!(4)).unwrap();
```
patrick-ogrady, line 1484: These are racy tests but do hit different codepaths.

cryptography/src/reed_solomon/rate/rate_low.rs
```diff
@@ -243,9 +243,23 @@ impl<E: Engine> RateDecoder<E> for LowRateDecoder<E> {
...
         }
 
+        // REVEAL ERASURES (RECOVERY)
+        //
+        // Recovery shards live at `work[chunk_size..recovery_end]`. Un-scale the missing ones
+        // by the inverse locator so they hold the canonical recovery values, mirroring the
+        // original reveal above. This lets `DecoderResult::restored_recovery` return them
+        // without a separate re-encode.
+
+        for i in chunk_size..recovery_end {
```
patrick-ogrady: I think we can make this optional

cryptography/src/reed_solomon/rate/rate_high.rs
```diff
@@ -243,9 +243,23 @@ impl<E: Engine> RateDecoder<E> for HighRateDecoder<E> {
...
         }
 
+        // REVEAL ERASURES (RECOVERY)
+        //
+        // Recovery shards live at `work[0..recovery_count]`. Un-scale the missing ones by the
+        // inverse locator so they hold the canonical recovery values, mirroring the original
+        // reveal above. This lets `DecoderResult::restored_recovery` return them without a
+        // separate re-encode.
+
+        for i in 0..recovery_count {
```
patrick-ogrady: I think we can make this optional

parallel/src/lib.rs
```diff
@@ -581,6 +696,74 @@ commonware_macros::stability_scope!(BETA, cfg(feature = "std") {
...
+        }
+
+        fn try_map_collect_vec<I, F, T, E>(&self, iter: I, map_op: F) -> Result<Vec<T>, E>
+        where
+            I: IntoIterator<IntoIter: Send, Item: Send> + Send,
+            F: Fn(I::Item) -> Result<T, E> + Send + Sync,
+            T: Send,
+            E: Send,
+        {
+            // A direct parallel map collecting into `Result` short-circuits on error and
```
patrick-ogrady: remove this comment

parallel/src/lib.rs
```diff
@@ -581,6 +696,74 @@ commonware_macros::stability_scope!(BETA, cfg(feature = "std") {
...
+
+        fn map_init_collect_vec<I, INIT, T, F, R>(&self, iter: I, init: INIT, map_op: F) -> Vec<R>
+        where
+            I: IntoIterator<IntoIter: Send, Item: Send> + Send,
+            INIT: Fn() -> T + Send + Sync,
+            T: Send,
+            F: Fn(&mut T, I::Item) -> R + Send + Sync,
+            R: Send,
+        {
+            // rayon's `map_init` reuses the per-job init value (e.g. a hasher) like
```
patrick-ogrady: remove this comment

PR #4092, opened 2026-06-22
[cryptography] Vendor `reed-solomon-simd`

patrick-ogrady:
To better optimize `commonware-coding`, we've decided to vendor `reed-solomon-simd`.

This will allow us to reduce some internal copies and ensure block formatting doesn't break #4091.

### DIFF
--- a/coding/src/reed_solomon.rs
+++ b/coding/src/reed_solomon.rs
@@ -2,7 +2,7 @@ use crate::{Config, Scheme};
 use bytes::{Buf, BufMut, Bytes};
 use commonware_codec::{BufsMut, EncodeSize, FixedSize, RangeCfg, Read, ReadExt, Write};
 use commonware_cryptography::{
-    reed_solomon::{Error as RsError, ReedSolomonDecoder, ReedSolomonEncoder},
+    reed_solomon::{Decoder, Encoder, Error as RsError},
     Digest, Hasher,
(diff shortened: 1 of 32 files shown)


coding/src/reed_solomon.rs
```diff
@@ -203,14 +205,14 @@ fn prepare_data(mut data: impl Buf, k: usize) -> (Vec<u8>, usize) {
...
-/// that decode to the same payload with a non-canonical shard width.
+/// `k` original shards, and rounds up to an even width required by the
+/// Reed-Solomon implementation. Decode uses the same calculation to reject
+/// commitments that decode to the same payload with a non-canonical shard width.
 const fn canonical_shard_len(data_len: usize, k: usize) -> usize {
     let prefixed_len = u32::SIZE + data_len;
     let mut shard_len = prefixed_len.div_ceil(k);
 
-    // Ensure shard length is even (required for optimizations in `reed-solomon-simd`)
+    // Ensure shard length is even, as required by the Reed-Solomon implementation.
```
patrick-ogrady, line 215: I suspect we could handle this in RS eventually but think the right first step is to just merge the vendor.

cronokirby (approved):
The changes over the dep seem fine; It's probably time to split up the cryptography crate, but this makes sense for here now.

PR #4095, opened 2026-06-23
[cryptography] Implement Golden using our own ZK circuit abstraction

### LINKED ISSUE #3885: [cryptography] Golden: avoid overhead of arkworks to bulletproofs circuit conversion

cronokirby:
This results in a circuit that's almost half the size, for free.

Closes #3885.


patrick-ogrady (dismissed):
As discussed, I think there is more to review/verify here but we can do this as we proceed through optimizations.

PR #4098, opened 2026-06-23
[storage/qmdb] Cache decoded keys during snapshot init to avoid redundant log reads

### LINKED ISSUE #4133: Partitioned ordered index: `prev_translated_key`/`next_translated_key` do an O(2^(8·P)) partition scan — merkleize hangs at P=3
### Summary

In the partitioned ordered index (`storage/src/index/partitioned/ordered/mod.rs`), `prev_translated_key` and `next_translated_key` locate the neighboring translated key by **linearly scanning partitions** until they hit a non-empty one. There are `2^(8·P)` partitions, so for **P=3 that is ~16.7M partition probes per call**. Merkleize's floor scan calls `prev_translated_key` **once per key**, so merkleizing an ordered P=3 batch over a sparse index performs roughly `keys × 16.7M` `partition_last` probes and effectively hangs (CPU-bound, no progress).

### Evidence

Sampling a hung ordered P=3 `generate` (via the `init_scale` bench against `AnyOFixP3Db`) shows ~100% of samples on the main thread here:

```
merkleize
 └ merkleize_with_floor_scan
    └ index::partitioned::ordered::Index::prev_translated_key
       └ partition_last      (alternating, repeated)
```

### Root cause

`prev_translated_key` (`mod.rs:522`):

```rust
let (i, sub) = partition_index_and_sub_key::<P>(key);
let k = self.translator.transform(sub);
if let Some(vals) = self.partition_prev_before(i, &k) {   // within partition i (fast)
    return Some((vals.iter(), false));
}
for p in (0..i).rev() {                                   // scan back to partition 0
    if let Some(vals) = self.partition_last(p) { return Some((vals.iter(), false)); }
}
for p in (0..self.partitions.len()).rev() {               // cycle: scan ALL partitions
    if let Some(vals) = self.partition_last(p) { return Some((vals.iter(), true)); }
}
None
```

Both `for` loops walk empty partitions one at a time. On a sparse index the nearest non-empty partition can be up to `2^(8·P)` away, and the cycle loop always scans the entire partition array when `[0, i)` is empty. `next_translated_key` (`mod.rs:549`) has the symmetric forward scan.

### Impact

- **Merkleize / generate** of an ordered partitioned DB at large P (the config intended for large keyspaces) effectively hangs. Sparse / low-occupancy indexes (the start of a build, or any sparse state) are the worst case.
- **Ordered neighbor queries** (e.g. exclusion-proof lookups) are also O(2^(8·P)) worst-case.

### Reproduction

An ordered P=3 merkleize over a sparse index reproduces it; concretely, generating an `AnyOFixP3Db` (P=3 ordered) hangs almost immediately while merkleizing the first seed batch.

### Notes

`prev_translated_key` / `next_translated_key` are unchanged from their introduction — this is a latent algorithmic-complexity issue, not a recent regression. It surfaces when an ordered P=3 merkleize is run at scale; the unit tests do not exercise that path.

### Suggested fix

Track partition occupancy (e.g. a hierarchical occupancy bitmap) so the nearest non-empty partition is found in O(log n) / O(1) instead of O(2^(8·P)). That removes the per-key scan from both merkleize's floor scan and ordered neighbor queries.

roberto-bayardo:
## Summary

Speeds up QMDB startup by eliminating redundant log reads during snapshot construction.

During init, `build_snapshot_from_log` replays the operations log and resolves translated-key collisions via `find_update_op`, which re-reads (and re-decodes) each candidate operation from the log just to recover its key. The journal's replay function does not put replayed data (either raw or decoded) into any cache, so each re-read is a system call + decode.  (After the first re-read, however, the *page* of the operation will be cached in the page cache.)

This PR caches `(location -> key)` mappings in a special purpose`Clock` cache that is populated & used only during the qmdb init phase:
- replayed update ops populate it as they stream by, and
- `find_update_op` consults it before falling back to a log read.

## Configurable

The cache capacity is a per-database config field, **`init_cache_size`** (in entries), where **`0` disables** the cache. The cache is sized to exactly `init_cache_size` and is held only for the duration of init. There is no hard-coded cap or default; callers choose the size for their workload.

## Why it's correct

There is no control-flow change: a cache hit only replaces a `reader.read`. Roots and recovery are unaffected, verified by the full qmdb suite (1544 tests), including a new test that reopens the same database with `init_cache_size = 0` and a large cache and asserts identical roots. The cache is threaded as `Option<&mut Clock>`: populated during init, `None` on the live `apply_batch` path.

## Benchmarks

Serial-init, cache disabled vs enabled (`cache=262144`), 100K elements / 1M updates+deletes, idle machine (`cargo bench -p commonware-storage --bench qmdb --features test-traits -- init`). The init benchmark sweeps `cache=0` (the no-cache baseline) vs an enabled cache in one process, so the baseline needs no branch switching.

| variant | off | on | change |
|---|---|---|---|
| `any::ordered::variable::mmr` | 55.76 ms | 37.92 ms | **-32.0%** |
| `current::ordered::variable::mmr` | 60.56 ms | 41.31 ms | **-31.8%** |
| `any::ordered::fixed::mmr` | 42.02 ms | 30.99 ms | **-26.3%** |
| `current::ordered::fixed::mmr` | 46.73 ms | 34.79 ms | **-25.5%** |
| `any::unordered::variable::mmr` | 37.31 ms | 32.16 ms | **-13.8%** |
| `any::unordered::fixed::mmr` | 26.58 ms | 23.29 ms | **-12.4%** |
| `current::unordered::variable::mmr` | 40.49 ms | 35.61 ms | **-12.1%** |
| `current::unordered::fixed::mmr` | 29.83 ms | 27.39 ms | **-8.2%** |

The cache elides the log reads that update resolution (`find_update_op`) performs during replay, so the gain tracks how much read/decode work that resolution does:

- **Variable vs fixed:** a variable-value log lookup is two reads (one for the offset, then one for the item) versus one for a fixed-value log, so variable shapes benefit more at each tier.
- **Ordered vs unordered:** ordered operations are larger (each carries the successor key for the ordered ring, ~1.5x the bytes), and every insert and delete also rewrites the predecessor's operation to relink the ring. Ordered databases therefore replay more, larger operations, so `find_update_op` does proportionally more read/decode work for the cache to eliminate.

These compound, so ordered + variable benefits most. The effect is tier-independent (`any` and `current` alike), and nothing regresses.

## Benchmarks at scale (`init_scale`)

The table above tops out at 100K elements / 1M ops; the cache's payoff **grows with database size**. It eliminates the redundant `find_update_op` reads, and the journal's page cache holds only raw pages (evicted during the sequential replay), so at multi-GB scale those reads increasingly miss the page cache. This PR adds a standalone `init_scale` benchmark that builds a real, multi-GB database once (applying `num_updates` random operations -- ~10% deletes -- sampled over a `keyspace`-key index space, uniformly or via Zipf, reporting its build time) and then reopens it (snapshot rebuild) at several cache sizes:

```
cargo bench -p commonware-storage --bench init_scale --features test-traits -- generate /tmp/db 50000000 250000000 0    # uniform
cargo bench -p commonware-storage --bench init_scale --features test-traits -- generate /tmp/db 50000000 250000000      # Zipf (default exponent 1.0)
cargo bench -p commonware-storage --bench init_scale --features test-traits -- bench    /tmp/db
cargo bench -p commonware-storage --bench init_scale --features test-traits -- destroy  /tmp/db
```

Reopen time, `any::ordered::fixed::mmr`, 50M-key keyspace / 250M updates, OS page cache warm (1 GiB), at cache off / a quarter of the replay region (`R/4`) / the whole replay region (`R`):

| workload | size | `R` (ops) | off | R/4 | R (full) |
|---|---|---|---|---|---|
| uniform | ~10 GB | 64.6M | 134.9 s | 89.7 s (-33%) | 41.5 s (**-69% / 3.25x**) |
| Zipf (s=1.0) | ~6.1 GB | 38.2M | 43.0 s | 31.5 s (-27%) | 22.4 s (**-48% / 1.92x**) |

At production scale a full-coverage cache cuts startup by **~2-3x** (up to **3.25x** on uniform churn, versus ~30% at 100K elements): the larger the replay region, the more collision-resolution reads miss the page cache and the more the cache elides. A partial (`R/4`) cache still recovers ~30%. Uniform churn benefits most -- every key is distinct, so resolution reads scatter across the whole log and are rarely page-cache-resident; Zipf concentrates updates on a hot subset (whose recent prior ops the page cache already holds) and lifts the inactivity floor faster, leaving a smaller `R`. The cache affects only the reopen path; generation is unchanged.

## Why not just cache the replayed pages?

Since each re-read is a syscall whose *page* then lands in the page cache, a natural alternative is to have `replay` populate the page cache as it streams, so even the first re-read is a memory hit and no dedicated cache is needed. We implemented that and measured it head-to-head against this PR's loc->key cache, warm, 50M keys, at full replay-region coverage:

| shape | replay-populated page cache | loc->key cache (this PR) |
|---|---|---|
| fixed (~10 GB db) | 41.8 s, ~6 GB resident | 43.0 s, ~4 GB |
| variable (~15 GB db) | 49.6 s, ~12.5 GB resident | 47.0 s, ~4 GB |

Init speed is comparable (within ~5%); the small crossover flips with shape, since a variable re-read is two reads plus offset resolution that the key cache skips. The loc->key cache wins on memory, robustness, and parallelizability:

- **Memory.** It stores only the 32-byte key per active location, so its footprint tracks key *count*, not log *bytes* (~4 GB for both shapes), and it is freed after init. Matching that speed with the page cache means pinning the entire active log for the process lifetime: ~6 GB (fixed) to ~12.5 GB (variable), growing with value size.
- **Robustness.** Undersize the page cache below the active log and re-reads miss again, ~2x slower at quarter-coverage (84 s fixed, 82 s variable); the key cache holds full coverage in ~4 GB and degrades gracefully.
- **Parallelizable.** Parallel snapshot construction shards cleanly: each worker can own a private, conflict-free key cache (disjoint partition ranges, hence disjoint keys; no lock, no cross-worker eviction). A shared page cache would instead serialize workers on its lock and let them thrash each other's pages, needing ~Nx the capacity to compensate.

Resolves #4133

### DIFF
--- a/storage/Cargo.toml
+++ b/storage/Cargo.toml
@@ -42,6 +42,7 @@ criterion.workspace = true
 governor.workspace = true
 paste.workspace = true
 rand.workspace = true
+rand_distr.workspace = true
 rstest.workspace = true
 tracing-subscriber.workspace = true
 
@@ -89,6 +90,12 @@ path = "src/qmdb/benches/constantinople.rs"
 name = "qmdb"
 harness = false
 path = "src/qmdb/benches/bench.rs"
+required-features = ["test-traits"]
+
+[[bench]]
+name = "init_scale"
+harness = false
+path = "src/qmdb/benches/init_scale.rs"
 required-features = ["test-traits"]
 
 [[bench]]
--- a/storage/src/qmdb/benches/common.rs
+++ b/storage/src/qmdb/benches/common.rs
@@ -44,6 +44,9 @@ pub const PAGE_CACHE_SIZE: NonZeroUsize = NZUsize!(512);
 pub const DELETE_FREQUENCY: u32 = 10;
 pub const VARIABLE_VALUE_MAX_LEN: usize = 256;
 pub const WRITE_BUFFER_SIZE: NonZeroUsize = NZUsize!(2 * 1024 * 1024);
+/// Default init-time `(location -> key)` cache size for bench databases. The init benchmark
+/// overrides this per-case to sweep cache sizes.
+pub const INIT_CACHE_SIZE: usize = 1 << 18;
 
 // -- Fixed value (Digest), fixed storage layout --
 
@@ -149,6 +152,7 @@ pub fn any_fix_cfg_with(
         merkle_config: merkle_cfg(PARTITION_FIX, ctx, page_cache.clone(), items_per_blob),
         journal_config: fix_log_cfg(PARTITION_FIX, page_cache, items_per_blob),
         translator: EightCap,
+        init_cache_size: INIT_CACHE_SIZE,
     }
 }
 
@@ -161,6 +165,7 @@ pub fn imm_fix_cfg_with(
         merkle_config: merkle_cfg(PARTITION_IMM, ctx, page_cache.clone(), items_per_blob),
         log: fix_log_cfg(PARTITION_IMM, page_cache, items_per_blob),
         translator: EightCap,
+        init_cache_size: INIT_CACHE_SIZE,
     }
 }
 
@@ -180,6 +185,7 @@ pub fn cur_fix_cfg_with(
         journal_config: fix_log_cfg(PARTITION_FIX, page_cache, items_per_blob),
         grafted_metadata_partition: format!("grafted-metadata-{PARTITION_FIX}"),
         translator: EightCap,
+        init_cache_size: INIT_CACHE_SIZE,
     }
 }
 
@@ -198,6 +204,7 @@ pub fn any_var_digest_cfg_with(
         merkle_config: merkle_cfg(PARTITION_VAR, ctx, page_cache.clone(), items_per_blob),
         journal_config: var_log_cfg(PARTITION_VAR, page_cache, ((), ()), items_per_blob),
         translator: EightCap,
+        init_cache_size: INIT_CACHE_SIZE,
     }
 }
 
@@ -217,6 +224,7 @@ pub fn cur_var_digest_cfg_with(
         journal_config: var_log_cfg(PARTITION_VAR, page_cache, ((), ()), items_per_blob),
         grafted_metadata_partition: format!("grafted-metadata-{PARTITION_VAR}"),
         translator: EightCap,
+        init_cache_size: INIT_CACHE_SIZE,
     }
 }
 
@@ -243,6 +251,7 @@ pub fn any_var_vec_cfg_with(
             items_per_blob,
         ),
         translator: EightCap,
+        init_cache_size: INIT_CACHE_SIZE,
     }
 }
 
@@ -267,6 +276,7 @@ pub fn cur_var_vec_cfg_with(
         ),
         grafted_metadata_partition: format!("grafted-metadata-{PARTITION_VAR}"),
         translator: EightCap,
+        init_cache_size: INIT_CACHE_SIZE,
     }
 }
 
@@ -344,12 +354,13 @@ macro_rules! define_db_variants {
 
         #[allow(unused_macros)]
         macro_rules! $timed_dispatch_name {
-            ($ctx_expr:expr, $variant_expr:expr, $iters:expr, |$db_name:ident| $body:expr) => {
+            ($ctx_expr:expr, $variant_expr:expr, $iters:expr, $cache_size:expr, |$db_name:ident| $body:expr) => {
                 match $variant_expr {
                     $(
                         $enum_name::$entry => {
                             let ctx = $ctx_expr;
-                            let cfg = $cfg(&ctx);
+                            let mut cfg = $cfg(&ctx);
+                            cfg.init_cache_size = $cache_size;
                             let start = std::time::Instant::now();
                             for _ in 0..$iters {
                                 #[allow(unused_mut)]
--- a/storage/src/qmdb/benches/common.rs
+++ b/storage/src/qmdb/benches/common.rs
@@ -47,7 +47,7 @@ pub const VARIABLE_VALUE_MAX_LEN: usize = 256;
 pub const WRITE_BUFFER_SIZE: NonZeroUsize = NZUsize!(2 * 1024 * 1024);
 /// Default init-time `(location -> key)` cache size for bench databases. The init benchmark
 /// overrides this per-case to sweep cache sizes.
-pub const INIT_CACHE_SIZE: usize = 1 << 18;
+pub const INIT_CACHE_SIZE: Option<NonZeroUsize> = Some(NZUsize!(1 << 18));
 
 // -- Fixed value (Digest), fixed storage layout --
 
--- a/storage/src/qmdb/benches/common.rs
+++ b/storage/src/qmdb/benches/common.rs
@@ -570,6 +570,10 @@ pub(crate) use define_vec_variants;
 /// Seed a database with `num_elements` entries, then perform `num_operations` random
 /// updates/deletes. Commits periodically when `commit_frequency` is `Some`.
 ///
+/// `seed_batch` caps how many seeds accumulate before each merkleize+apply+commit (bounding peak
+/// memory); `None` seeds in a single batch. `prune_frequency` prunes to the inactivity floor every
+/// `Some(n)` commits to bound on-disk growth; `None` never prunes during generation.
+///
 /// `key_zipf_exponent` selects how update/delete keys are sampled: `None` is uniform; `Some(s)`
 /// draws keys from a Zipf distribution with exponent `s`, so a hot subset is churned far more than
 /// the long tail (a more realistic workload than uniform churn).
@@ -579,6 +583,7 @@ pub(crate) use define_vec_variants;
 /// `k >= num_elements`) seeds a uniform-random subset of `num_elements` keys out of `k` and samples
 /// updates over all of `k`, so some updates land on unseeded keys and insert them organically -- a
 /// growing keyspace rather than a fixed population.
+#[allow(clippy::too_many_arguments)]
 pub async fn gen_random_kv<F, M>(
     db: &mut M,
     num_elements: u64,
@@ -610,10 +615,7 @@ pub async fn gen_random_kv<F, M>(
         for i in 0u64..num_elements {
             // With a sparse `keyspace`, seed a uniform-random subset of it (so later updates can land
             // on unseeded keys and insert them); otherwise seed keys 0..num_elements.
-            let idx = match keyspace {
-                Some(k) => rng.next_u64() % k,
-                None => i,
-            };
+            let idx = keyspace.map_or(i, |k| rng.next_u64() % k);
             let key = Sha256::hash(&idx.to_be_bytes());
             batch = batch.write(key, Some(make_value(&mut rng)));
             pending += 1;
@@ -622,7 +624,7 @@ pub async fn gen_random_kv<F, M>(
                 db.apply_batch(merkleized).await.unwrap();
                 db.commit().await.unwrap();
                 commits += 1;
-                if prune_frequency.is_some_and(|n| commits % n == 0) {
+                if prune_frequency.is_some_and(|n| commits.is_multiple_of(n)) {
                     let boundary = db.sync_boundary().await;
                     db.prune(boundary).await.unwrap();
                 }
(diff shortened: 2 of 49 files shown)


roberto-bayardo:
bugbot run

storage/src/qmdb/mod.rs
```diff
@@ -248,23 +253,35 @@ where
         .await?;
     pin_mut!(stream);
     let last_commit_loc = bounds.end.saturating_sub(1);
+
+    // Memoize `(location -> key)` for replayed update ops so collision resolution in
+    // `find_update_op` resolves candidates from memory instead of re-reading (and re-decoding) the
+    // log. The mapping is immutable (append-only log), so the cache never goes stale.
+    let replay_len = bounds.end.saturating_sub(*inactivity_floor_loc) as usize;
```
patrick-ogrady: dumb question: should we just infer the cache size from this?

roberto-bayardo: I couldn't think of a great way to do this well without a few more stats, e.g. how many unique keys, update frequency, etc, so I just punted to the user :).   I suspect performance won't be all that sensitive to cache size.

storage/src/qmdb/immutable/mod.rs
```diff
@@ -148,6 +148,10 @@ pub struct Config<T: Translator, J, S: Strategy> {
 
     /// The translator used by the compressed index.
     pub translator: T,
+
+    /// Capacity (in entries) of the `(location -> key)` cache used during init to resolve snapshot
+    /// collisions without re-reading the log; `0` disables it.
```
patrick-ogrady: Elsewhere, we'd do this as `Option<NonZeroUsize>`

roberto-bayardo: done

storage/src/qmdb/benches/common.rs
```diff
@@ -44,6 +45,9 @@ pub const PAGE_CACHE_SIZE: NonZeroUsize = NZUsize!(512);
 pub const DELETE_FREQUENCY: u32 = 10;
 pub const VARIABLE_VALUE_MAX_LEN: usize = 256;
 pub const WRITE_BUFFER_SIZE: NonZeroUsize = NZUsize!(2 * 1024 * 1024);
+/// Default init-time `(location -> key)` cache size for bench databases. The init benchmark
```
patrick-ogrady: remove this

storage/Cargo.toml
```diff
@@ -91,6 +92,12 @@ harness = false
 path = "src/qmdb/benches/bench.rs"
 required-features = ["test-traits"]
 
+[[bench]]
+name = "init_scale"
```
patrick-ogrady, line 96: this should live in `qmdb` benches?

storage/src/qmdb/mod.rs
```diff
@@ -248,23 +252,33 @@ where
...
                 }
             } else if op.is_update() {
                 let new_loc = crate::merkle::Location::new(loc);
-                let old_loc = update_key(snapshot, reader, key, new_loc).await?;
+                let old_loc = update_key(snapshot, reader, key, new_loc, cache.as_mut()).await?;
                 callback(true, old_loc);
                 if old_loc.is_none() {
                     active_keys += 1;
                 }
+                // This update op is now a `find_update_op` candidate for later ops of its key.
```
patrick-ogrady, line 279: add a \n before

storage/Cargo.toml
patrick-ogrady, line 122: this should also be moved into index?

storage/src/qmdb/benches/init_scale.rs
```diff
@@ -0,0 +1,204 @@
...
+}
+
+/// Reopen the database at `folder` (read-only) and time `init` at three cache regimes: off, a
+/// quarter of the replay region (fills + evicts), and the whole replay region (no eviction).
+fn bench(folder: &str) {
+    let cfg = Config::default().with_storage_directory(folder);
+
+    // No-cache baseline; also learn the replay region R (ops above the inactivity floor) -- what the
+    // location cache must cover to avoid eviction (a key-cache would need only the key count).
+    let (baseline, region) = time_init(&cfg, None);
```
patrick-ogrady, line 169:
```
  In storage/src/qmdb/benches/init_scale.rs:148, bench(folder) immediately calls:

  let (baseline, region) = time_init(&cfg, None);

  But time_init calls AnyOFixDb::init(...). That API is not a read-only open. In storage/src/qmdb/any/mod.rs:127, if the log is empty,
  init creates a new DB by appending a CommitFloor(None, Location::new(0)) and syncing it.

  So if you run:

  cargo bench ... -- bench /tmp/missing-or-empty-db

  the benchmark creates a tiny empty QMDB before checking whether it exists. After that, time_init computes:

  end - floor

  For the newly initialized empty DB, end == 1 and floor == 0, so region == 1, not 0. That means this guard does not catch it:

  if region == 0 {
      eprintln!("no database ...");
      return;
  }

  The bad outcome is:

  1. bench silently measures init of an empty DB, not a generated large DB.
  2. It leaves storage files behind.
  3. A later generate /same/folder ... refuses to run because generate checks db_dir_nonempty(folder) and now the folder is non-empty.

  The simple fix is to reject a missing or empty folder before calling time_init, probably by moving a folder existence/non-empty check
  to the top of bench. A stricter fix would be a real non-creating open path for the benchmark, but the narrow bug is that bench must
  not call creating init before validating the directory.
```

storage/src/qmdb/benches/common.rs
```diff
@@ -556,34 +569,90 @@ pub(crate) use define_vec_variants;
...
+    // them from memory -- without that, a large build accumulates the whole merkle in RAM.
     {
         let mut batch = db.new_batch();
+        let mut pending = 0u64;
         for i in 0u64..num_elements {
-            let k = Sha256::hash(&i.to_be_bytes());
-            batch = batch.write(k, Some(make_value(&mut rng)));
+            // With a sparse `keyspace`, seed a uniform-random subset of it (so later updates can land
+            // on unseeded keys and insert them); otherwise seed keys 0..num_elements.
+            let idx = keyspace.map_or(i, |k| rng.next_u64() % k);
```
patrick-ogrady:
```
  In storage/src/qmdb/benches/common.rs:618, seed keys are chosen like this when keyspace is Some(k):

  let idx = keyspace.map_or(i, |k| rng.next_u64() % k);

  That is sampling with replacement. So the loop runs num_elements times, but it does not guarantee num_elements distinct keys.

  Duplicates matter because the batch stores mutations in a BTreeMap, and write says “last value wins” for the same key. Within one seed
  batch, duplicate keys collapse before apply. Across seed batches, duplicate keys become updates to already seeded DB keys. Either way,
  the active seeded key count is less than requested.

  This affects init_scale today because it calls:

  Some(KEYSPACE_MULTIPLIER * elements)

  with KEYSPACE_MULTIPLIER = 4.

  So if elements = n, it samples n times from a 4n keyspace with replacement. Expected distinct seeded keys are roughly:

  4n * (1 - e^(-1/4)) ~= 0.885n

  So generate 50_000_000 seeds about 44.2M distinct keys before churn, not 50M. Churn may later insert additional unseeded keys, but the
  documented seed phase is currently not doing what it says.
```

patrick-ogrady (commented):
just nits

storage/src/qmdb/benches/common.rs
roberto-bayardo: done

storage/src/qmdb/mod.rs
roberto-bayardo: done

storage/src/qmdb/benches/init_scale.rs
roberto-bayardo: done

storage/src/qmdb/benches/common.rs
roberto-bayardo: done - the seed now writes the distinct keys `0..num_elements` (the with-replacement sampling is gone).

storage/Cargo.toml
roberto-bayardo: This lives in `src/qmdb/benches/init_scale.rs` - a standalone CLI (own `main`), so it gets its own `[[bench]]` rather than joining the Criterion `qmdb` group.

storage/Cargo.toml
roberto-bayardo: It is - `src/index/benches/scale.rs` (likewise a standalone CLI with its own `[[bench]]`).

patrick-ogrady (approved):

PR #4100, opened 2026-06-24
[stoarge/journal/contiguous] Fold `Reader` into `Contiguous`

danlaine:
`Contiguous` handed out a `Reader` (an owned snapshot) per read. This drops that indirection: `Contiguous` carries the read methods itself, and reads use a cheap borrowed view instead of an owned snapshot.

## Trait
- Remove `Reader`; move `bounds`/`read`/`read_many`/`try_read_sync`/`replay` onto `Contiguous`.
- Remove `Contiguous::reader()` and `Contiguous::size()` (callers use `bounds().end`).

## Journal (fixed + variable)
- Implement `Contiguous` directly; reads go through an internal borrowed `reader()` (a live view over `Blobs::Writable`). The `&self` borrow blocks concurrent mutation, so no snapshot is captured.
- Keep `Journal::snapshot()` for reads that must outlive a borrow: a detached, owned reader (`Blobs::Snapshot`) with frozen bounds, readable across prunes; rewind refused (`BlobInUse`) while one is alive.
- `Reader` gains a lifetime (`Reader<'a>`) and a `Blobs<'a>` enum unifying the live and snapshot backings.
- `replay` runs through a shared helper that owns its handles (`use<E, V>`), letting `Journal::replay` return a stream without borrowing a function-local snapshot.
- Rename the rewind gate `readers` -> `snapshots`.

## Consumers (qmdb, queue, authenticated)
- `journal.reader().await.{read,bounds,...}` -> direct `journal.{...}`; `size().await` -> `bounds().end`.
- De-async `size`/`bounds`/`update_metrics` where their only `await` was the removed call.
- `R: Reader` trait bounds -> `R: Contiguous`.

### DIFF
--- a/examples/sync/src/databases/any.rs
+++ b/examples/sync/src/databases/any.rs
@@ -136,7 +136,7 @@ where
     E: Storage + Clock + Metrics,
 {
     async fn size(&self) -> Location {
-        self.bounds().await.end
+        self.bounds().end
     }
 
     async fn sync_boundary(&self) -> Location {
--- a/examples/sync/src/databases/current.rs
+++ b/examples/sync/src/databases/current.rs
@@ -160,7 +160,7 @@ where
     E: Storage + Clock + Metrics,
 {
     async fn size(&self) -> Location {
-        self.bounds().await.end
+        self.bounds().end
     }
 
     async fn sync_boundary(&self) -> Location {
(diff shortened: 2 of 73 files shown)


roberto-bayardo (approved):
lgtm

patrick-ogrady:

PR #4104, opened 2026-06-24
[cryptography] Improve parallelism in Bulletproofs prover

cronokirby:
Not quite linear scaling, but without this the prover doesn't scale at all with parallelism, so this is a good change.

Also added benchmarks for just the IPA.

### DIFF
--- a/cryptography/Cargo.toml
+++ b/cryptography/Cargo.toml
@@ -161,3 +161,8 @@ path = "src/handshake/benches/bench.rs"
 name = "bloomfilter"
 harness = false
 path = "src/bloomfilter/benches/bench.rs"
+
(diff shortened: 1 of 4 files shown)


cryptography/Cargo.toml
```diff
@@ -161,3 +161,8 @@ path = "src/handshake/benches/bench.rs"
 name = "bloomfilter"
 harness = false
 path = "src/bloomfilter/benches/bench.rs"
+
+[[bench]]
+name = "ipa"
```
patrick-ogrady: this should be called `bulletproofs`?

patrick-ogrady (approved):

PR #4072, opened 2026-06-18
[marshal] Reduce `fsync`

patrick-ogrady:
Resolves: #4013

## Summary

This PR reduces marshal fsync latency by starting durable storage work earlier and letting consensus await the resulting handle only at the point where durability is required.

- Adds `start_sync` plumbing through the buffered runtime writers, segmented journals, and prunable archive so callers can start persistence and await completion separately.
- Changes marshal proposed/verified/certified notifications to return sync handles, with sync failures treated as fatal instead of converted into `false` verification/certification verdicts.
- Adds certification gate tracking for standard inline/deferred and coding marshal paths. Proposal and verification paths register the in-flight durability task, and `certify` awaits it before allowing a finalize vote.
- Starts durable storage concurrently with application verification for structurally valid candidates. This lets verification return its app verdict without waiting on fsync, while certification still consumes the registered gate before finalizing.
- Removes the Simplex local-proposal certification shortcut. Leaders now certify their own proposals through the automaton, making certification the durability barrier for finalize votes.
- Handles unclean-shutdown recovery where local in-memory gate state is gone: certification falls back to fetching the notarized block by round, then persists it if the block is available from peers.

## Validation

- `cargo fmt --check -p commonware-consensus`
- `git diff --check`
- `just test -p commonware-consensus test_inline_propose_then_certify_persists_block`
- `just test -p commonware-consensus test_deferred_propose_then_certify_persists_block`
- `just test -p commonware-consensus test_standard_certify_missing_candidate_fetches_by_round`
- `just test -p commonware-consensus test_coding_certify_missing_candidate_fetches_by_round`
- `just test -p commonware-consensus test_mailbox_verified_sync_failure_panics`

PR #4106, opened 2026-06-24
[runtime] Prepare for `fsync` Changes

patrick-ogrady:
Related: #4072


Cargo.toml
patrick-ogrady, line 1: These (unrelated) changes remove any dependency we have on `openssl`.

storage/src/journal/segmented/oversized.rs
roberto-bayardo, line 342: Somewhat unrelated to this PR but do we also want to an `fn commit()` here that minimally ensures durability of the section?  (seems you can minimally sync the values blob since recovery will restore them?)

runtime/src/utils/buffer/write.rs
```diff
@@ -69,6 +69,16 @@ impl<B: Blob> State<B> {
...
         Ok(())
     }
+
+    /// Start syncing the underlying blob if there are unsynced mutations.
+    async fn start_sync(&mut self) -> Handle<()> {
+        if !self.needs_sync {
+            return Handle::ready(Ok(()));
+        }
+        let handle = self.blob.start_sync().await;
+        self.needs_sync = false;
```
roberto-bayardo: What if one thread calls start_sync concurrently to another one?  The second to get through will observe needs_sync = false, and its handle will suggest the DB is durable, when it isn't actually durable until the first handle completes.

storage/src/journal/segmented/oversized.rs
patrick-ogrady: I prefer only using that pattern when there is a clear A->B commit. If A/B can be done currently, I don't see a need.

runtime/src/utils/buffer/write.rs
patrick-ogrady: I'm an idiot.

storage/src/journal/segmented/oversized.rs
roberto-bayardo: Might be a mac thing but even when done concurrently it seems to 2x the sync time.

PR #4107, opened 2026-06-24
[runtime] [storage/journal] Release write buffers of sealed journal sections

roberto-bayardo:
## Problem

A segmented journal keeps one write buffer per section in `Manager::blobs`, each
holding a `write_buffer`-sized backing allocation (default 2 MiB), released only
by `prune()`. When a section fills and the journal advances to the next one, the
prior section's buffer keeps its full backing alive: a flush drains the full
pages but leaves the trailing partial page as a *slice* of the original
allocation (`tip::Buffer::drop_prefix`), so a sub-page remainder pins the whole
2 MiB. `commit()`/`sync()` flush the bytes to disk but never release this
backing.

Resident memory therefore grows ~2 MiB per `items_per_blob` operations -- for
**both** the operations log and the merkle node journal -- and is reclaimed only
at `prune()`. During a large QMDB build (which raises the inactivity floor but
does not prune), this dominates memory and OOMs at scale.

## Fix

Seal a section when the journal advances past it (`Manager::get_or_create`):
drain its full pages and copy the partial-page remainder into a right-sized
buffer (`tip::Buffer::compact`), releasing the oversized backing. The active tip
keeps its full write buffer for append reuse; sealed sections shrink from ~2 MiB
to one page. Durability is unchanged -- `seal` uses the existing un-synced flush
path, so nothing is durable until the next `commit`/`sync`, exactly as before.

## Impact (1M-key + 5M-update build)

|                      | before  | after                                   |
|----------------------|---------|-----------------------------------------|
| live heap at peak    | 1.48 GB | 236 MB                                  |
| live 2 MiB buffers   | 665     | 65 (bounded by pool cap, not # sections)|
| peak RSS             | 1.66 GB | 766 MB                                  |

At 50M keys this is the difference between ~66 GB of section-buffer overhead and
~0.5 GB.

## Tests

- New `test_seal_preserves_data_across_pages` (append spanning pages, seal,
  verify data is intact and survives a sync + reopen).
- `commonware-runtime` buffer tests, `commonware-storage` `journal::`
  (segmented/contiguous/authenticated), and `qmdb::any` all pass.

### DIFF
--- a/runtime/src/utils/buffer/paged/append.rs
+++ b/runtime/src/utils/buffer/paged/append.rs
@@ -1235,6 +1235,21 @@ impl<B: Blob> Append<B> {
         blob_state.sync().await
     }
 
+    /// Releases the oversized write-buffer backing once this blob is no longer the active append
+    /// tip.
+    ///
+    /// Full pages are drained to the blob (without an extra sync) so only a sub-page remainder is
+    /// left buffered, then that remainder is copied into a right-sized allocation, freeing the
+    /// write-buffer-sized backing it would otherwise pin. The buffered bytes are preserved and are
+    /// still made durable by a later [Self::sync]; only surplus pooled capacity is reclaimed.
+    pub async fn seal(&self) -> Result<(), Error> {
+        let buf_guard = self.buffer.write().await;
+        self.flush_internal(buf_guard, false, false).await?;
+        let mut buffer = self.buffer.write().await;
+        buffer.compact();
+        Ok(())
+    }
+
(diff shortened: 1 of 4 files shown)


roberto-bayardo:
Superseded by @danlaine's sealed-blob stack, which wires `Sealed` into the fixed journal directly (a run of sealed blobs ending in one writable tail) and fixes this same write-buffer retention leak — more comprehensively, alongside lock removal and snapshot readers: #3908 → #3964 → #3970. Closing in favor of that work.

PR #4113, opened 2026-06-25
[storage/glue] Keep QMDB P2P sync responsive through verification

danlaine:
Database reads no longer block the QMDB P2P resolver actor from processing local requests, deliveries, and cancellation. The actor polls the read in an owned future pool, with serving limited to one active read. Excess peer requests are declined through the resolver's existing retry path.

`Source::serve` takes a validating transform and returns its accepted value. An invalid P2P candidate leaves the same source call and subscriber open for the resolver to retry. Each caller has its own native subscription, leaving request coalescing and cached redelivery with the resolver.

QMDB sync authenticates responses against the exact request and trusted root before accepting them. Target advances cancel obsolete requests and requests whose historical roots leave the retention window. Callers sharing a mailbox verify one QMDB history, so explicit verdicts agree. Deliveries with no live caller remain unjudged; previously they were accepted and credited to the peer.

### DIFF
--- a/glue/src/stateful/db/p2p/compact/actor.rs
+++ b/glue/src/stateful/db/p2p/compact/actor.rs
@@ -194,10 +194,10 @@ where
                     value,
                     response,
                 } => {
-                    self.handle_deliver(key, value, response).await;
+                    self.handle_deliver(key, value, response);
                 }
                 handler::EngineMessage::Produce { key, response } => {
-                    self.handle_produce(key, response).await;
+                    self.handle_produce(key, response);
                 }
             },
         }
@@ -247,7 +247,7 @@ where
         true
     }
 
-    async fn handle_deliver(
+    fn handle_deliver(
         &mut self,
         key: handler::Request<F, H::Digest>,
         value: bytes::Bytes,
@@ -298,13 +298,19 @@ where
             return;
         }
 
-        let mut peer_valid = true;
-        for approval in approvals {
-            if let Ok(approved) = approval.await {
-                peer_valid &= approved;
+        self.context.child("approve").spawn(move |_| async move {
+            let mut peer_valid = true;
+            for approval in approvals {
+                match approval.await {
+                    Ok(true) | Err(_) => {}
+                    Ok(false) => {
+                        peer_valid = false;
+                        break;
+                    }
+                }
             }
-        }
-        response.send_lossy(peer_valid);
+            response.send_lossy(peer_valid);
+        });
     }
 
     fn valid_state_response(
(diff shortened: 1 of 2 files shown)


glue/src/stateful/db/p2p/compact/actor.rs
```diff
@@ -329,19 +335,23 @@ where
...
-        response.send_lossy(fetch.state.encode());
+        let database = database.clone();
+
+        self.context.child("serve").spawn(move |_| async move {
+            let Ok(fetch) = compact::Resolver::get_compact_state(&database, key.to_target()).await
+            else {
+                return;
+            };
+            response.send_lossy(fetch.state.encode());
+        });
```
clabby, line 359: We do rate limit peers, though there’s no explicit limit to the number of tasks a peer can make us spawn here by requesting ops. Wonder if we should add a `max_concurrent_serves` configuration option and put these tasks in a semaphore?

glue/src/stateful/db/p2p/compact/actor.rs
```diff
@@ -329,19 +335,23 @@ where
...
     ) {
         let State::HasDb(database) = &self.state else {
             return;
         };
-        let Ok(fetch) = compact::Resolver::get_compact_state(database, key.to_target()).await
-        else {
-            return;
-        };
-        response.send_lossy(fetch.state.encode());
+        let database = database.clone();
```
patrick-ogrady, line 351: ah, very convenient.

glue/src/stateful/db/p2p/compact/actor.rs
```diff
@@ -298,13 +298,19 @@ where
             return;
         }
 
-        let mut peer_valid = true;
-        for approval in approvals {
-            if let Ok(approved) = approval.await {
-                peer_valid &= approved;
+        self.context.child("approve").spawn(move |_| async move {
```
patrick-ogrady: I don't have great benchmarks to back up this claim but we just read the next N ready in marshal and do a simple try_join on handling.

I think doing this is probably better but wonder if we can just use a futures pool instead of setting up so many tasks?

glue/src/stateful/db/p2p/compact/actor.rs
patrick-ogrady: In other areas, we definitely assume rate limits keep us safe. Adding a secondary limit here I fear could be confusing/make it harder to reason about.

clabby (dismissed):

glue/src/stateful/db/p2p/actor.rs
```diff
@@ -191,35 +202,43 @@ where
             _ = &mut resolver_task => {
                 return;
             },
+            rejected = self.tasks.next_completed() => {
+                if let Some((request, subscriber)) = rejected {
+                    // The current rejected subscription keeps its verdict and retry
+                    // alive. Earlier groups have already finished their local work.
+                    resolver_mailbox.retain(move |key, id| key != &request || *id >= subscriber);
```
danlaine: This keeps the rejected subscription alive after its local group is already gone (`handle_deliver` removes the group at delivery time), so after any rejection the resolver holds a subscription nobody on this node is waiting on. Base had the same orphan, but there the later `Cancel` swept it because it dropped every subscription for the key. Now `Cancel` drops only its own ID, so the orphan is permanent unless some peer eventually answers.

Reproduction: A requests K and rejects the response. B requests K (new subscriber ID) and then cancels. The actor ends with zero pending requests while the resolver still has one pending fetch for K belonging to A. With the only peer blocked, it never clears.

patrick-ogrady (approved):
I think there is more to clean up/polish in this area of the code but feel it at least now better integrates with the resolver (and its capabilities).

patrick-ogrady (changes requested):
tweaking serve loop

patrick-ogrady (approved):

PR #4116, opened 2026-06-26
[storage] [runtime] Add `start_commit` to contiguous journals, sealed blobs, paged writer

danlaine:
## Journal API

`fixed::Journal` and `variable::Journal` gain:

```rust
pub async fn start_commit(&mut self) -> Handle<()>
```

The journal stays fully usable (appends, reads, snapshots) while the handle is pending; awaiting the handle gives the same guarantee `commit()` gave. `commit()` is now `start_commit().await`, and `sync()` additionally persists the recovery watermark (fixed) or offsets journal (variable) *after* the handle completes, so neither ever advances over non-durable data.

At most one commit is in flight at a time: a second `start_commit` waits for the first. Dropping the handle is safe — the fsync still runs, and an unobserved failure resurfaces from the next durability operation.

## Durability model

Previously each journal tracked `dirty_from_blob` and fsynced every dirty blob at commit. Now durability rides the blob lifecycle:

- **Rollover** seals the filled blob and *starts* its fsync, after first awaiting the previous rollover's fsync. `Writable` tracks these as two slots: `tail_sync` (from `start_commit`) and `predecessor_sync` (from rollover).
- **Invariant**: only the tail and its immediate predecessor are ever non-durable. Every older blob was fsync-barriered by a later rollover, so a commit only needs to join the predecessor and tail syncs.
- **Destructive ops** (prune, rewind, clear, destroy) drain both slots first. This is what preserves the rewind contract — a crash recovers a size no shorter than the rewind target — without the eager per-blob syncs the old code needed.

`dirty_from_blob` is gone from both journals; the paged `Writer`'s internal sync state (which never loses a failure) plus the invariant subsume it.

## Recovery

Two behavior changes fall out of the invariant:

1. **Rolled predecessors now survive a crash without a commit.** Rollover fsyncs them, so recovery adopts filled predecessor blobs that the old code lost (they were buffered-only until commit).
2. **A crash mid-rollover-fsync can tear a blob interior.** Writeback reordering can persist a valid last page above a lost or stale interior page — a hole the backward scan in `Writer::new` cannot see, since it trusts the last valid page. Init now forward-validates the two newest blobs (the only possibly-non-durable ones) with the new `Writer::recoverable_prefix_len` and truncates at the first torn or short page. Fixed floors the cut to a chunk boundary; variable truncates to the page boundary and lets frame replay repair the mid-frame cut as torn trailing junk. Either way, a short non-newest blob then ends the contiguous prefix and newer blobs are removed, exactly like other gaps.

## Paged writer and sealed blobs

`Writer::seal()` now returns `(Sealed, Handle<()>)`: it flushes buffered bytes and starts the blob's fsync, returning the read handle plus the sync's completion. `Sealed::sync` is removed — `Sealed` is a pure immutable read view with no durability responsibilities. `Writer::recoverable_prefix_len` is public for the recovery scan described above.


runtime/src/utils/buffer/paged/writer.rs
```diff
@@ -980,6 +980,18 @@ impl<B: Blob> Writer<B> {
         self.sync_inner().await
     }
 
+    /// Flushes buffered data and begins making all pending mutations durable. Awaiting the returned
+    /// [`Handle`] waits for the same durability guarantee as [`Self::sync`].
+    pub async fn start_sync(&mut self) -> Handle<()> {
+        if let Err(err) = self.flush_internal(true, false).await {
```
patrick-ogrady, line 963: does this block on the completion of blob.start_sync? or, is it possible to have a flush inflight while we are still ensuring durability of a previous flush?

patrick-ogrady:
bugbot run

patrick-ogrady:
bugbot run

patrick-ogrady:
bugbot run

runtime/src/utils/buffer/paged/sealed.rs
```diff
@@ -37,7 +35,8 @@ impl<B: Blob> Clone for Sealed<B> {
 }
 
 struct SealedInner<B: Blob> {
-    /// The underlying blob being wrapped.
+    /// The underlying blob. Sealed views are read-only, so the blob is held bare; the writer's
```
patrick-ogrady: We don't need this longer-form comment

storage/src/journal/contiguous/blobs.rs
```diff
@@ -8,12 +8,18 @@ use commonware_formatting::hex;
...
+use futures::{
+    future::{BoxFuture, Shared},
+    FutureExt as _,
 };
-use futures::future::try_join_all;
 use std::{collections::BTreeMap, num::NonZeroUsize, sync::Arc};
 use tracing::debug;
 
+/// A shareable observer for an in-flight sync: one fan-out fsync, many waiters share the result.
+type SyncObserver = Shared<BoxFuture<'static, Result<(), RError>>>;
```
patrick-ogrady: this is already defined elsewhere?

storage/src/journal/contiguous/blobs.rs
```diff
@@ -379,25 +399,49 @@ impl<E: Context> Writable<E> {
...
-        try_join_all(dirty_sealed.iter().map(|sealed| sealed.sync()))
+    async fn wait_predecessor(syncing: Option<SyncObserver>) -> Result<(), RError> {
+        match syncing {
+            Some(syncing) => syncing.await,
+            None => Ok(()),
+        }
+    }
+
+    /// Wait for the sync started when the current predecessor was sealed.
+    async fn drain_predecessor(&mut self) -> Result<(), Error> {
```
patrick-ogrady: this wrapping around wait_predecessor is not necessary?

patrick-ogrady (commented):
pretty close

storage/src/journal/contiguous/variable.rs
```diff
@@ -1192,7 +1178,7 @@ impl<E: Context, V: CodecShared> Journal<E, V> {
             .ok_or(Error::SizeOverflow)?;
 
         let items_per_blob = self.items_per_blob.get();
-        self.mark_dirty_from(position_to_blob(self.bounds.end, items_per_blob));
+
```
patrick-ogrady: remove this \n

storage/src/journal/contiguous/variable.rs
```diff
@@ -2013,7 +1997,7 @@ impl<E: Context, V: CodecShared> Journal<E, V> {
 
     /// Test helper: Sync the data tail.
     pub(crate) async fn test_sync_data(&mut self) -> Result<(), Error> {
```
patrick-ogrady: is this helper required?

storage/src/journal/contiguous/mod.rs
```diff
@@ -232,6 +232,10 @@ impl<T> Many<'_, T> {
 }
 
 /// A [Contiguous] journal that supports appending, rewinding, and pruning.
+///
+/// Durability is exposed through [Self::commit] and [Self::sync]. An overlapping, non-blocking sync
```
patrick-ogrady: This is not necessary...remove this added comment

storage/src/journal/contiguous/blobs.rs
```diff
@@ -379,25 +394,52 @@ impl<E: Context> Writable<E> {
...
     }
 
+    /// Begin making the current tail and its sealed predecessor durable. Awaiting the returned
+    /// [`Handle`] waits for the same durability guarantee as [Self::sync].
+    pub(super) async fn start_sync(&mut self) -> Handle<()> {
+        let predecessor = self.predecessor_sync.clone();
+        let tail = self.tail.start_sync().await;
+        self.metrics.synced.inc();
+        Handle::from_future(async move {
+            Self::wait_predecessor(predecessor).await?;
```
patrick-ogrady: wait for predecessor before starting next sync?

runtime/src/utils/buffer/paged/writer.rs
```diff
@@ -1094,20 +1096,30 @@ impl<B: Blob> Writer<B> {
...
-    /// [`super::Sealed::sync`].
+    /// not fsynced.
     pub async fn seal(mut self) -> Result<super::Sealed<B>, Error> {
         self.flush_internal(true, false).await?;
         Ok(self.sealed_handle(self.id))
     }
+
+    /// Consume the write handle, begin making its current bytes durable, and return an immutable
+    /// [`super::Sealed`] handle plus the completion handle for that sync.
+    pub async fn seal_start_sync(mut self) -> Result<(super::Sealed<B>, Handle<()>), Error> {
```
patrick-ogrady: I think we should just make this how "seal" works?

patrick-ogrady (commented):
almost there

storage/src/journal/contiguous/blobs.rs
```diff
@@ -379,25 +405,52 @@ impl<E: Context> Writable<E> {
...
+    /// Begin making the current tail and its sealed predecessor durable. Awaiting the returned
+    /// [`Handle`] waits for the same durability guarantee as [Self::sync].
+    pub(super) async fn start_sync(&mut self) -> Handle<()> {
+        if let Err(err) = Self::wait_predecessor(self.predecessor_sync.clone()).await {
+            return Handle::ready(Err(err));
+        }
+        self.predecessor_sync = None;
+        let tail = self.tail.start_sync().await;
+        self.metrics.synced.inc();
+        tail
```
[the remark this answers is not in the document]
patrick-ogrady: We already warn that a failed operation may leave memory in an odd state.

storage/src/journal/contiguous/variable.rs
```diff
@@ -1326,33 +1305,37 @@ impl<E: Context, V: CodecShared> Journal<E, V> {
...
+    /// submitted; awaiting the returned [`Handle`] waits for the same durability guarantee as
+    /// [Self::commit] (not [Self::sync]): the recovery watermark is not advanced and the offsets
+    /// journal is not synced, so follow with [Self::commit] or [Self::sync] to persist metadata. The
+    /// journal remains callable while the sync is in flight; an append that must flush buffered bytes
+    /// to the in-flight tail blob waits for that sync to finish (the durability gate).
+    ///
+    /// Dropping or aborting the returned [`Handle`] stops waiting but does not cancel the in-flight
+    /// fsync; it continues, and the next commit/sync/structural operation drains it.
+    pub async fn start_sync(&mut self) -> Handle<()> {
+        self.metrics.record_commit();
```
patrick-ogrady: Why are we recording commit here?

danlaine: this is now start_commit

storage/src/journal/contiguous/blobs.rs
```diff
@@ -252,23 +268,31 @@ impl<E: Context> Writable<E> {
...
             .ok_or(Error::OffsetOverflow)?;
         let new_writer = self.partition.open(next_blob).await?;
         let old_writer = std::mem::replace(&mut self.tail, new_writer);
-        let sealed = old_writer.seal().await.map_err(Error::Runtime)?;
+        let (sealed, handle) = old_writer.seal().await?;
         self.metrics.tracked.inc();
+        self.metrics.synced.inc();
         self.sealed.push(sealed);
         self.sealed_snapshot = None;
+        debug_assert!(self.predecessor_sync.is_none());
```
patrick-ogrady: `assert!` ?

storage/src/journal/contiguous/blobs.rs
```diff
@@ -252,23 +268,31 @@ impl<E: Context> Writable<E> {
...
+        // seal() waits for any outstanding writer sync, so a successful seal proves any prior
```
patrick-ogrady: add a \n above?

danlaine: done

storage/src/journal/contiguous/blobs.rs
```diff
@@ -252,23 +268,32 @@ impl<E: Context> Writable<E> {
...
         self.metrics.tracked.inc();
+        self.metrics.synced.inc();
         self.sealed.push(sealed);
         self.sealed_snapshot = None;
+        debug_assert!(self.predecessor_sync.is_none());
+
+        // seal() waits for any outstanding writer sync, so a successful seal proves any prior
+        // tail sync already succeeded.
+        self.tail_sync = None;
+        self.predecessor_sync = Some(handle.boxed().shared());
```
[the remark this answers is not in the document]
danlaine: Errors on mutable methods are fatal so this isn't an issue

storage/src/journal/contiguous/blobs.rs
danlaine: This assertion is a check to ensure that we don't remove `self.drain_predecessor_sync()`, which unconditionally sets `self.predecessor_sync = None`. If that call were removed, this assertion would fail in tests as intended. Running this check in production doesn't buy us anything imo -- it's impossible by construction for it to fire. On the other hand it's an O(1) check so the stakes are basically 0.

storage/src/journal/contiguous/variable.rs
```diff
@@ -1064,6 +1053,28 @@ impl<E: Context, V: CodecShared> Inner<E, V> {
...
+        let suspects: Vec<u64> = pending.keys().rev().take(2).copied().collect();
+        for blob in suspects {
+            let writer = pending.get_mut(&blob).expect("suspect blob is present");
+            let valid = writer.recoverable_prefix_len().await?;
+            if valid < writer.size() {
+                warn!(
+                    blob,
+                    valid,
+                    size = writer.size(),
+                    "interior hole detected: truncating to last well-formed page"
```
patrick-ogrady: I think just `truncating to last well-formed page` is sufficient?

storage/src/journal/contiguous/variable.rs
```diff
@@ -1504,31 +1509,28 @@ impl<E: Context, V: CodecShared> Inner<E, V> {
...
     }
 
     /// In-place [Journal::commit].
     pub(crate) async fn commit(&mut self) -> Result<(), Error> {
         let _timer = self.metrics.commit_timer();
         self.metrics.commit_calls.inc();
-        self.flush_dirty_data().await?;
-        self.dirty_from_blob = None;
+        let handle = self.blobs.start_sync().await;
+        handle.await?;
```
patrick-ogrady: can we just return this?

patrick-ogrady (dismissed):

patrick-ogrady (approved):

PR #4110, opened 2026-06-25
[runtime] [storage] Add views; simplify replay; cleanup

danlaine:

PR #4119, opened 2026-06-26
[storage/journal] Refactor Contiguous Journal

patrick-ogrady:
## Summary

This PR folds the storage/runtime journal ownership stack into a single branch against `main`. It introduces sealed/read views and single-owner writers, removes interior mutability from contiguous journal paths, and simplifies replay around shared blob handling.

## Replaces

This replaces the previous stacked PRs:

- #3908 - [Runtime] Add `Sealed` blob
- #3964 - [runtime] [storage] Split `Append` into read/write handles
- #3969 - [storage] `Persistable::commit` and sync take `&mut self`
- #3970 - [storage/journal/contiguous] replace fixed-journal locks with ownership and snapshot readers
- #4041 - [storage/journal/contiguous] Remove interior mutability from variable journal
- #4070 - [storage/journal/contiguous] Use Sealed blobs in variable journal
- #4100 - [storage/journal/contiguous] Fold `Reader` into `Contiguous`
- #4082 - [runtime] [storage] Make the paged Writer a single owner; simplify replay
- #4112 - [runtime] [storage] Remove `AsyncMutex` from `buffer::Write`; `Ordinal`
- #4110 - [runtime] [storage] Add views; simplify replay; cleanup
- https://github.com/commonwarexyz/monorepo/pull/4124
- #4129 

#4116 remains stacked on top of this branch.


consensus/src/ordered_broadcast/engine.rs
```diff
@@ -1101,7 +1101,10 @@ impl<
             .expect("journal does not exist");
 
         // Sync journal
-        journal.sync(section).await.expect("unable to sync journal");
+        journal
+            .sync(&[section])
```
patrick-ogrady: IIRC there is a way to do "one-of" here?

consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -249,7 +249,7 @@ impl<
     async fn sync_journal(&mut self, view: View) {
         if let Some(journal) = self.journal.as_mut() {
             journal
-                .sync(view.get())
+                .sync(&[view.get()])
```
patrick-ogrady: one-of?

runtime/src/utils/buffer/paged/mod.rs
```diff
@@ -26,20 +26,101 @@
...
+    Ok(())
+}
+
+/// Partition a batch read into items copied from in-memory tail bytes and items that need
+/// cache/blob reads.
+///
+/// `buf` holds one `item_size` slot per offset (validated by [validate_read_many_into]). `tail`
+/// holds the logical bytes at `[tail_offset, tail_offset + tail.len())`; for [Writer] this is the
+/// tip buffer, for [Sealed] the partial last page. Items entirely within
+/// `tail` are copied into
```
patrick-ogrady: odd line cutoff?

runtime/src/utils/buffer/paged/view.rs
```diff
@@ -0,0 +1,235 @@
...
+    ///
+    /// Returns the number of items fully served without a blob read (from the in-memory tail and the
+    /// page cache). The remaining items required at least one blob read.
+    pub async fn read_many_into(
+        &self,
+        buf: &mut [u8],
+        offsets: &[u64],
+        item_size: NonZeroUsize,
+    ) -> Result<usize, Error> {
+        super::validate_read_many_into(buf.len(), offsets, item_size, self.size)?;
```
patrick-ogrady, line 167: I wonder if we want to assume this? is there any meaningful overhead to running this?

runtime/src/utils/buffer/paged/writer.rs
```diff
@@ -1289,99 +1030,127 @@ impl<B: Blob> Append<B> {
...
+                Crc32::checksum(new_data),
+                &old_crc,
+            )
+            .await?;
+        self.partial_page_state = Some(final_record);
 
         Ok(())
     }
+
+    /// Page-cache id used for reads. Exposed for tests.
```
patrick-ogrady, line 1064: this `id` concept feels odd to me but will look deeper to better appreciate

runtime/src/lib.rs
```diff
@@ -130,6 +130,8 @@ stability_scope!(BETA {
         InvalidChecksum,
         #[error("offset overflow")]
         OffsetOverflow,
+        #[error("invalid input: {0}")]
+        InvalidInput(&'static str),
```
patrick-ogrady: feels a bit like a code smell

storage/src/freezer/storage.rs
```diff
@@ -695,7 +695,7 @@ impl<E: BufferPooler + Context, K: Array, V: CodecShared> Freezer<E, K, V> {
                     .await?;
 
                 // Sync oversized
-                oversized.sync(checkpoint.section).await?;
+                oversized.sync(&[checkpoint.section]).await?;
```
patrick-ogrady: this array change is super nice

storage/src/journal/contiguous/blobs.rs
```diff
@@ -0,0 +1,616 @@
...
+    }
+
+    /// Drop every blob below `min_blob` and remove its file, oldest-first. Readers holding
+    /// the old slice keep reading the removed blobs.
+    ///
+    /// # Invariants
+    ///
+    /// - `oldest_blob_index < min_blob <= tail_blob_index`
+    pub(super) async fn prune(&mut self, min_blob: u64) -> Result<(), Error> {
+        debug_assert!(self.oldest_blob_index < min_blob && min_blob <= self.tail_blob_index());
```
patrick-ogrady: change all these debug_assert to assert

runtime/src/utils/buffer/paged/writer.rs
```diff
@@ -1289,99 +1030,127 @@ impl<B: Blob> Append<B> {
...
+        let full_pages = self.current_page;
+        assert_eq!(
+            full_pages.checked_mul(logical_page_size),
+            Some(self.buffer.offset),
+            "flushed page count is inconsistent with the buffer offset"
+        );
+        let partial_page = if self.buffer.is_empty() {
+            None
+        } else {
+            let mut copy = self.cache_ref.pool().alloc(self.buffer.len());
```
patrick-ogrady: We can avoid this copy in many cases with this:

```
diff --git a/runtime/src/utils/buffer/paged/writer.rs b/runtime/src/utils/buffer/paged/writer.rs
index 240bab80d..adada027d 100644
--- a/runtime/src/utils/buffer/paged/writer.rs
+++ b/runtime/src/utils/buffer/paged/writer.rs
@@ -1129,9 +1129,7 @@ impl<B: Blob> Writer<B> {
         let partial_page = if self.buffer.is_empty() {
             None
         } else {
-            let mut copy = self.cache_ref.pool().alloc(self.buffer.len());
-            copy.put_slice(self.buffer.as_ref());
-            Some(copy.freeze())
+            Some(self.buffer.slice(..))
         };
         super::Sealed::new(
             self.blob.clone(),
@@ -3608,7 +3606,7 @@ mod tests {
     }
 
     #[test]
-    fn test_snapshot_try_read_sync_prefers_copied_tail() {
+    fn test_snapshot_try_read_sync_prefers_snapshot_tail() {
         executor.start(|context: deterministic::Context| async move {
             let cache_ref = CacheRef::from_pooler(&context, PAGE_SIZE, NZUsize!(BUFFER_SIZE));
@@ -3629,7 +3627,7 @@ mod tests {
             let snapshot = append.snapshot().await.unwrap();
 
             let poison = vec![0xBB; page_size];
-            // If the snapshot consulted the page cache for its copied tail, this would leak in.
+            // If the snapshot consulted the page cache for its tail, this would leak in.
             assert_eq!(
                 append.cache_ref.cache(append.id, &poison, page_size as u64),
                 0

```

storage/src/journal/contiguous/variable.rs
```diff
@@ -804,56 +987,64 @@ impl<E: Context, V: CodecShared> Journal<E, V> {
...
+        let items_per_blob = self.items_per_blob.get();
+        self.mark_dirty_from(position_to_blob(self.bounds.end, items_per_blob));
         let mut written = 0;
         while written < items_count {
-            let section = position_to_section(inner.size, self.items_per_section);
-            let pos_in_section = inner.size % self.items_per_section;
-            let remaining_space = (self.items_per_section - pos_in_section) as usize;
-            let batch_count = remaining_space.min(items_count - written);
+            let pos_in_blob = self.bounds.end % items_per_blob;
+            // Take the min in u64: casting `remaining_space` first would truncate on 32-bit
```
patrick-ogrady: this is the second place we have this comment...maybe we should turn this into a helper of sorts?

in any case, the comment should go above `let pos_in_blob...`

storage/src/journal/contiguous/variable.rs
```diff
@@ -864,37 +1055,65 @@ impl<E: Context, V: CodecShared> Journal<E, V> {
...
-            assert_eq!(last_offsets_pos, inner.size + batch_count as u64 - 1);
+            assert_eq!(last_offsets_pos, self.bounds.end + batch_count as u64 - 1);
 
-            inner.size += batch_count as u64;
+            self.bounds.end += batch_count as u64;
             written += batch_count;
-            Self::mark_dirty_from(&mut inner, section);
+
+            if self.bounds.end.is_multiple_of(items_per_blob) {
+                // Seal the just-filled tail and open the next blob as the new tail. This does
```
patrick-ogrady: move above `if`

also remove the `--`

storage/src/qmdb/sync/compact.rs
```diff
@@ -750,7 +750,7 @@ macro_rules! impl_compact_resolver_immutable {
                 let db = self.read().await;
                 fetch_state_from_full_source(
                     target,
-                    || async { Target::new(db.root(), db.bounds().await.end) },
+                    || async { Target::new(db.root(), db.bounds().end) },
```
patrick-ogrady: do we need this `async{}` wrapper (or support for it) anymore?

patrick-ogrady (commented):
pretty massive ergonomic improvement

storage/src/journal/contiguous/mod.rs
```diff
@@ -63,46 +69,49 @@ pub trait Reader: Send + Sync {
         None
     }
 
-    /// Return a stream of all items starting from `start_pos`.
+    /// Return a stream of all items starting from `start_pos`, bounded by `bounds()`.
     ///
-    /// Because the reader holds the lock, validation and stream setup happen
-    /// atomically with respect to `prune()`.
+    /// `buffer` controls how many logical positions are read per chunk.
```
patrick-ogrady: This is bytes almost everywhere else (if not everywhere else)?

patrick-ogrady: Fixed: #4124

storage/src/journal/contiguous/variable.rs
```diff
@@ -394,169 +785,115 @@ impl<E: Context, V: CodecShared> super::Reader for Reader<'_, E, V> {
...
+            blobs,
             offsets,
-            items_per_section,
+            bounds,
+            dirty_from_blob: None,
+            items_per_blob: cfg.items_per_section,
             compression: cfg.compression,
-            metrics,
+            codec_config: cfg.codec_config,
+            metrics: Arc::new(metrics),
```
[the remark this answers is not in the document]
commonware-llm: I think this is a false positive. `align` does not just advance offsets after reading data: before `offsets.sync()` it calls `sync_data_range(pending, data_sync_start, data_size, ...)`, so any adopted/rebuilt data the offsets point at is durable first.

The later `Writable::recover(...).await` only seals the already-recovered non-tail writers. `Writer::seal()` can flush buffered bytes, but for a reopened unchanged partial page the writer detects that the buffered partial length matches the existing CRC record and emits no new write. If recovery actually repaired or adopted bytes, they were synced before the offsets became durable.

The normal append-time seal path is covered separately by `mark_dirty_from(...)` before writes/seal, so `commit()`/`sync()` still fsync sealed dirty blobs there. I agree this ordering is subtle; we should add a short comment near this init block to make the `dirty_from_blob: None` invariant explicit.

patrick-ogrady:
bugbot run

PR #4121, opened 2026-06-26
[storage/qmdb] Parallel snapshot init for the partitioned indexes

roberto-bayardo:
## Summary

Parallelizes QMDB snapshot reconstruction (`init`) for the partitioned indexes (both the ordered and unordered variants). The main task replays the operations log once and routes each operation to one of N worker tasks by partition; workers read through a shared `Arc` of the log, and each owns its own `(location -> key)` cache and builds a disjoint partition range. The main task then merges the ranges and rebuilds the activity bitmap.

Builds on the init read cache (#4098, now merged into main): each worker gets its own conflict-free key cache, so init parallelizes with no lock contention and no cross-worker eviction (see #4098's "Why not just cache the replayed pages?").

## Design

- **Routing.** `partition_index_and_sub_key::<P>` maps each key to a partition; worker `w` owns a contiguous partition range. Two keys that can collide in the index share the partition prefix and capped sub-key, so all colliding ops route to the same worker; each worker resolves its own collisions in isolation, identically to the serial walk.
- **Per-worker state.** Workers read through a shared `Arc` of the log (concurrent reads, nothing writes the log during init; sole ownership is reclaimed after the build), and each holds its own `Clock<u64, Key>` cache (`init_cache_size / workers` each). The `(location -> key)` mapping is immutable, so caches never go stale and are conflict-free across workers (disjoint key sets).
- **Range-only workers.** A worker allocates only its partition range (`Index::new_range`), addressed by the global partition index via an `offset`, so P=3 (2^24 partitions) is affordable: per-worker memory is its range, not the full partition set.
- **Merge.** Main joins the workers, installs each range into `self`, and rebuilds the activity bitmap in location order. The result is independent of worker join order (deterministic).
- **Pump throughput.** The init task is the pump: replay reads through a 2 MiB buffer (amortizing the per-refill thread hand-off to the storage layer), op decode validates zero padding a chunk at a time, and the activity bitmap installs into the shared bitmap under a single lock acquisition after the build.
- **Failure drain.** If replay fails mid-route, the error is captured, the channels are closed, and every worker is joined before the error is returned, so no worker outlives a failed init or retains the log after it. A regression injects a read fault into the replay stream and asserts the log has a single owner when the build returns.

## Configurable

`init_concurrency` on the `any`/`current` configs counts the tasks that build the snapshot, including the init task, which always replays and routes the log (`NonZeroUsize`: the init task always participates):

- `1`: build entirely on the init task, no worker tasks.
- `n`: spawn `n - 1` worker tasks. The effective worker count never exceeds the partition count, and because partition ranges are fixed-width (except possibly the final range) it can round down (256 partitions at `init_concurrency = 201` build with 128 workers of two partitions each). `2` still de-interleaves the build from the replay onto its own task.

No derived/automatic count is offered: the plateau below makes a fixed small count the right choice everywhere, which keeps the snapshot build decoupled from the `Strategy` entirely.

Both partitioned index variants (ordered and unordered) consume it; the non-partitioned indexes build serially regardless. State-sync rebuilds honor the configured `init_concurrency`.

## Correctness

Equivalence tests (`any/ordered/fixed.rs`) populate a churny database and reopen it across a range of `init_concurrency` settings (`1`, single- and multi-worker counts, counts that round down to fewer workers, and counts above the partition count that clamp), asserting identical roots in every case (serial == parallel). `_p1`/`_p2` run in CI; `_p3` is `#[ignore]` (large allocation). The unordered partitioned index has matching equivalence coverage at P1 (`any/unordered/fixed.rs` and the `current` variant); its P2 (65,536 sub-indexes) is too memory-heavy for the suite. A matching test on the `current` variant covers the activity-bitmap reconstruction specifically: the current root commits to the bitmap, so it verifies the parallel build's `for_each_value` + last-commit recompute against the serial path. Determinism holds across runs with the same seed.

## Benchmarks (warm OS cache, P=3, full-coverage init cache)

Reopen time vs `init_concurrency` on two 50M-keyspace / 250M-update DBs (Linux x86_64, AMD EPYC 9354P). Each datapoint is one `init` in a fresh process (what a restart actually pays) with the OS file cache warm; values are the mean of two interleaved rounds:

| init_concurrency | uniform (R=64.6M) | zipf (R=38.2M) |
|---|---|---|
| `1` (serial) | 47.7 s | 28.7 s |
| `2` | 36.0 s (**1.3x**) | 22.1 s (**1.3x**) |
| `3` | 18.4 s (**2.6x**) | 11.1 s (**2.6x**) |
| `5` | 15.2 s (**3.1x**) | 9.9 s (**2.9x**) |
| `8` | 14.8 s (**3.2x**) | 8.7 s (**3.3x**) |

One additional worker alone (`init_concurrency = 2`) beats serial: moving the index build off the init task de-interleaves two dissimilar workloads (op decode vs index build) that otherwise thrash a single core. Returns then diminish past ~2 workers and flatten by ~4 (`init_concurrency` ~3-5): the floor is the init task's serial replay + per-op routing (decode-dominated), which the workers hide under once there are enough of them; further workers mostly idle. The unordered partitioned index is apply-bound the same way and parallelizes comparably (measured post-merge, same 50M/250M scale, `any::unordered::fixed` P=2): ~1.4x at `init_concurrency = 2`, ~3x at `3`, and ~3.9x (uniform) / ~4.2x (zipf) at `8`.

```
cargo bench -p commonware-storage --bench init_scale --features test-traits -- generate /tmp/db 50000000 250000000 0   # uniform (omit the 0 for Zipf)
cargo bench -p commonware-storage --bench init_scale --features test-traits -- bench    /tmp/db <cache> <parallelism>  # one reopen per invocation
```

## Notes

- Profiling reveals we could get better performance with a replay mode that decoded only keys; values provided by replay are not required by the init replay task.

---

_Edited post-merge to correct the scope: the parallel build covers both partitioned index variants (ordered and unordered), not the ordered variant alone. The merge commit message on `main` is unchanged._

### DIFF
--- a/examples/sync/src/databases/any.rs
+++ b/examples/sync/src/databases/any.rs
@@ -34,7 +34,6 @@ pub type Operation = FixedOperation<mmr::Family, Key, Value>;
 pub fn create_config(context: &impl BufferPooler) -> Config<Translator, Sequential> {
     let page_cache = buffer::paged::CacheRef::from_pooler(context, NZU16!(2048), NZUsize!(10));
     Config {
-        init_workers: None,
         merkle_config: MmrConfig {
             journal_partition: "mmr-journal".into(),
             metadata_partition: "mmr-metadata".into(),
@@ -51,6 +50,7 @@ pub fn create_config(context: &impl BufferPooler) -> Config<Translator, Sequenti
         },
         translator: Translator::default(),
         init_cache_size: Some(NZUsize!(1 << 16)),
+        init_concurrency: NZUsize!(1),
     }
 }
 
--- a/examples/sync/src/databases/any.rs
+++ b/examples/sync/src/databases/any.rs
@@ -50,7 +50,8 @@ pub fn create_config(context: &impl BufferPooler) -> Config<Translator, Sequenti
         },
         translator: Translator::default(),
         init_cache_size: Some(NZUsize!(1 << 16)),
-        init_concurrency: NZUsize!(1),
+        init_buffer: NZUsize!(1 << 21),
+        init_concurrency: (),
     }
 }
 
--- a/examples/sync/src/databases/any.rs
+++ b/examples/sync/src/databases/any.rs
@@ -9,7 +9,7 @@ use commonware_storage::{
     journal::contiguous::fixed::Config as FConfig,
     mmr::{self, Location, Proof, full::Config as MmrConfig},
     qmdb::{
-        self, InitParallelism,
+        self,
         any::{
             FixedConfig as Config,
             unordered::{
@@ -34,7 +34,7 @@ pub type Operation = FixedOperation<mmr::Family, Key, Value>;
 pub fn create_config(context: &impl BufferPooler) -> Config<Translator, Sequential> {
     let page_cache = buffer::paged::CacheRef::from_pooler(context, NZU16!(2048), NZUsize!(10));
     Config {
-        init_parallelism: InitParallelism::Serial,
+        init_workers: None,
         merkle_config: MmrConfig {
             journal_partition: "mmr-journal".into(),
             metadata_partition: "mmr-metadata".into(),
--- a/examples/sync/src/databases/any.rs
+++ b/examples/sync/src/databases/any.rs
@@ -9,7 +9,7 @@ use commonware_storage::{
     journal::contiguous::fixed::Config as FConfig,
     mmr::{self, Location, Proof, full::Config as MmrConfig},
     qmdb::{
-        self,
+        self, InitParallelism,
         any::{
             FixedConfig as Config,
             unordered::{
@@ -34,6 +34,7 @@ pub type Operation = FixedOperation<mmr::Family, Key, Value>;
 pub fn create_config(context: &impl BufferPooler) -> Config<Translator, Sequential> {
     let page_cache = buffer::paged::CacheRef::from_pooler(context, NZU16!(2048), NZUsize!(10));
     Config {
+        init_parallelism: InitParallelism::Serial,
         merkle_config: MmrConfig {
             journal_partition: "mmr-journal".into(),
             metadata_partition: "mmr-metadata".into(),
--- a/storage/src/index/partitioned/ordered/mod.rs
+++ b/storage/src/index/partitioned/ordered/mod.rs
@@ -50,6 +50,8 @@ use crate::{
     },
     translator::Translator,
 };
+#[commonware_macros::stability(ALPHA)]
+use commonware_runtime::telemetry::metrics::{Registered, Registration};
 use commonware_runtime::{
     Metrics,
     telemetry::metrics::{Counter, Gauge, MetricsExt as _},
@@ -73,9 +75,15 @@ pub struct Index<T: Translator, V: Send + Sync, const P: usize> {
     /// Translates the prefix-stripped key bytes into a partition-local key.
     translator: T,
 
-    /// The `2^(8*P)` partitions, indexed by the `P`-byte key prefix. Each stores its translated
-    /// keys and values as sorted arrays (the inline representation); an emptied partition may
-    /// instead have spilled (see `spilled`).
+    /// Base partition index this instance addresses from: a key's global partition `p` maps to local
+    /// slot `p - offset`. A full index covers every partition and so has `offset == 0`. A parallel
+    /// snapshot-build worker (see [`Self::new_range`]) covers only `[offset, offset + len)`.
+    offset: usize,
+
+    /// The partitions this instance covers, indexed by `global_partition - offset`. A full index
+    /// holds all `2^(8*P)` (offset `0`) while a build worker holds only its range. Each stores its
+    /// translated keys and values as sorted arrays (the inline representation), though an emptied
+    /// partition may instead have spilled (see `spilled`).
     partitions: Box<[Partition<T::Key, V>]>,
 
     /// Partitions that have spilled out of their sorted arrays (reached `SPILL_THRESHOLD` entries),
@@ -95,6 +103,10 @@ pub struct Index<T: Translator, V: Send + Sync, const P: usize> {
 
     /// Metric: cumulative values removed (via `remove`, cursor `delete`, or `retain`).
     pruned: Counter,
+
+    /// Metric: cumulative partitions spilled to the side-table. Emptied partitions that de-spill
+    /// (rare) are not subtracted. Detached on build workers and folded in at [`Self::install_range`].
+    spills: Counter,
 }
 
 impl<T: Translator, V: Send + Sync, const P: usize> Index<T, V, P> {
@@ -110,12 +122,14 @@ impl<T: Translator, V: Send + Sync, const P: usize> Index<T, V, P> {
             .into_boxed_slice();
         Self {
             translator,
+            offset: 0,
             partitions,
             spilled: HashMap::new(),
             threshold: SPILL_THRESHOLD,
             keys: ctx.gauge("keys", "Number of translated keys in the index"),
             items: ctx.gauge("items", "Number of items in the index"),
             pruned: ctx.counter("pruned", "Number of items pruned"),
+            spills: ctx.counter("spills", "Number of partitions spilled to the side-table"),
         }
     }
 
@@ -137,6 +151,7 @@ impl<T: Translator, V: Send + Sync, const P: usize> Index<T, V, P> {
         }
         let inner: BTreeMap<T::Key, Vec<V>> = self.partitions[i].drain_runs().into_iter().collect();
         self.spilled.insert(i, inner);
+        self.spills.inc();
     }
 
     /// The `BTreeMap` of spilled partition `i`, or `None` if `i` has not spilled. The empty-map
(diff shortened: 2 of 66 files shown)


storage/src/qmdb/any/ordered/fixed.rs
```diff
@@ -584,6 +593,105 @@ pub(crate) mod test {
...
+        db.commit().await.unwrap();
+        db.sync().await.unwrap();
+        let root = db.root();
+        drop(db);
+
+        // Reopen with a range of worker counts; all rebuild from the same log and must match.
+        for &workers in parallelisms {
+            let mut cfg = fixed_db_config::<OneCap>(partition, &context);
+            cfg.init_parallelism = match workers {
+                0 => crate::qmdb::InitParallelism::Serial,
```
patrick-ogrady: We should be using `commonware-parallel` here?

roberto-bayardo: You mean use commonware_parallel::Strategy instead?  This enum is pretty specific to the Index initializer which doesn't use Rayon and allows configuring # of workers (since in many cases you don't get much benefit beyond 2....).

roberto-bayardo:
Addressed the review findings in ce66ca7cb:

**Context widening** — `storage::Context` is back to main's definition; the `Spawner` bound now lives only on the qmdb init/snapshot-build paths (`init_from_log`, the `SnapshotBuild` method, the typed `Db::init` wrappers, and the sync-driven inits) plus the glue init paths that call them. The consensus, reshare, and runtime-mock edits are gone from the diff entirely — the alias widening was a convenience that leaked well past where spawning actually happens.

**Worker scheduling** — workers now spawn via `.shared(true)`, matching the repo convention for CPU-intensive work. With the `Auto` cap at four workers they're bounded in both count and lifetime, so the blocking-friendly shared pool fits better than dedicated threads.

**The undocumented panic** — addressed with a variation on the first suggested option: making the trait `pub(crate)` outright fails private-in-public (the public `any::init` bounds on it), so `SnapshotBuild` is now **sealed** instead — still nameable in public signatures, but only in-crate index types can implement it. That turns the drop-all-log-clones rule into an internal invariant held by the four implementations (and documented on the trait); the `Arc::into_inner` expect stays as an internal-invariant check. Open to an error return instead if preferred, but panicking on a broken in-crate invariant matches existing style.

**Description** — updated: the `Auto` bullet documents the four-worker cap, the benchmark table labels the `Auto` row as "measured pre-cap at 7 workers" (7 and 4 are both past the ~2-worker plateau), and the per-worker-state bullet now describes the shared `Arc` of the log rather than per-worker readers.

storage/src/qmdb/mod.rs
```diff
@@ -405,6 +413,314 @@ where
...
+    /// Build with up to this many worker tasks, plus the task that replays and routes the log.
+    /// Partition ranges are fixed-width (except possibly the final range), so the effective count
+    /// can be lower than requested (e.g. 256 partitions with `Workers(200)` build with 128 workers
+    /// of two partitions each), and never exceeds the partition count. `Workers(1)` still
+    /// de-interleaves replay from the build onto a separate task.
+    Workers(NonZeroUsize),
+    /// Build with worker tasks derived from the strategy's parallelism hint, reserving one core for
+    /// the replay/routing task and capping at a few workers (past that, routing is the bottleneck).
+    /// Falls back to a serial build when the hint leaves no spare core.
+    Auto,
```
patrick-ogrady: We should not derive something from the parallelism hint and then not use strategy (this is an anti-pattern -> you don't know who else is using that strategy and assuming we can run with the same parallelism is aggressive).

storage/src/qmdb/mod.rs
```diff
@@ -405,6 +413,314 @@ where
...
+        // This index type builds serially, so `parallelism` has no effect; warn on an explicit
+        // worker count rather than silently ignore it (`Auto` is best-effort and stays quiet).
+        if matches!(parallelism, InitParallelism::Workers(_)) {
+            tracing::warn!(
+                ?parallelism,
+                "init_parallelism configured but this index builds serially; ignoring"
+            );
+        }
+        Box::pin(async move {
+            build_snapshot_from_log(inactivity_floor_loc, &**log, self, cache_size, callback).await
```
patrick-ogrady: just make this `#[boxed]`?

examples/sync/src/databases/any.rs
```diff
@@ -34,6 +34,7 @@ pub type Operation = FixedOperation<mmr::Family, Key, Value>;
 pub fn create_config(context: &impl BufferPooler) -> Config<Translator, Sequential> {
     let page_cache = buffer::paged::CacheRef::from_pooler(context, NZU16!(2048), NZUsize!(10));
     Config {
+        extra_workers: 0,
```
patrick-ogrady: let's actually do `init_concurrency` and just make it `NonZero` (and if 1, just do serial?)

roberto-bayardo: 1 != serial though?  1 is in fact quite a bit (30%) faster than serial...

storage/src/qmdb/any/mod.rs
```diff
@@ -131,6 +132,14 @@ pub struct Config<T: Translator, J, S: Strategy> {
     /// Capacity (in entries) of the `(location -> key)` cache used during init to resolve snapshot
     /// collisions without re-reading the log; `None` disables it.
     pub init_cache_size: Option<NonZeroUsize>,
+
+    /// Number of tasks that build the init-time snapshot, including the init task itself
+    /// (which replays and routes the log); `1` builds entirely on the init task. Only certain
+    /// index types (such as the ordered-partitioned index) build in parallel. Counts past a few
```
patrick-ogrady: remove: ` Counts past a few
    /// waste tasks without speeding up the build, and each additional task is spawned as
    /// blocking-friendly shared work, so on runtimes with a bounded blocking pool keep counts
    /// well below that pool's size.`

storage/src/qmdb/current/ordered/fixed.rs
```diff
@@ -92,7 +93,9 @@ pub mod partitioned {
     > Db<F, E, K, V, H, T, P, N, S>
     {
         /// Initializes a [Db] authenticated database from the given `config`.
-        /// The configured [`Strategy`] is used to parallelize merkleization.
+        /// The configured [`Strategy`] is used to parallelize merkleization, and
```
patrick-ogrady: Remove `        /// The configured [`Strategy`] is used to parallelize merkleization, and
        /// `config.init_concurrency` bounds how many tasks the snapshot build splits across during
        /// init.`

storage/src/qmdb/current/ordered/variable.rs
```diff
@@ -99,7 +100,9 @@ pub mod partitioned {
         Operation<F, K, V>: Codec,
     {
         /// Initializes a [Db] from the given `config`.
-        /// The configured [`Strategy`] is used to parallelize merkleization.
+        /// The configured [`Strategy`] is used to parallelize merkleization, and
```
patrick-ogrady: remove: `        /// The configured [`Strategy`] is used to parallelize merkleization, and
        /// `config.init_concurrency` bounds how many tasks the snapshot build splits across during
        /// init.`

storage/src/qmdb/current/mod.rs
```diff
@@ -379,6 +380,14 @@ pub struct Config<T: Translator, J, S: Strategy> {
     /// Capacity (in entries) of the `(location -> key)` cache used during init to resolve snapshot
     /// collisions without re-reading the log; `None` disables it.
     pub init_cache_size: Option<NonZeroUsize>,
+
+    /// Number of tasks that build the init-time snapshot, including the init task itself
+    /// (which replays and routes the log); `1` builds entirely on the init task. Only certain
+    /// index types (such as the ordered-partitioned index) build in parallel. Counts past a few
```
patrick-ogrady: Remove: `Counts past a few
    /// waste tasks without speeding up the build, and each additional task is spawned as
    /// blocking-friendly shared work, so on runtimes with a bounded blocking pool keep counts
    /// well below that pool's size.`

storage/src/qmdb/mod.rs
```diff
@@ -413,6 +418,285 @@ where
...
+        // Close the channels so each worker's stream terminates and it returns its index.
+        drop(senders);
+
+        // Join workers before surfacing any replay failure, so none outlive a failed init.
+        let joined = join_all(handles).await;
+        routing_result?;
+
+        // Install each worker's partition range into self. Each worker carries its own
+        // partition offset, so installation needs no range arguments.
+        let mut total_items = 0i64;
```
patrick-ogrady: why i64?

storage/src/index/partitioned/ordered/mod.rs
```diff
@@ -207,6 +222,92 @@ impl<T: Translator, V: Send + Sync, const P: usize> Index<T, V, P> {
...
+        }
+    }
+
+    /// Move a build `worker`'s partitions and spilled entries into self (a full `offset == 0` index),
+    /// at the global slots `[worker.offset, worker.offset + worker.partitions.len())`, which must be
+    /// empty. Build workers are transient, so all of the worker's metric counts (`spills`, `pruned`,
+    /// `keys`, `items`) are folded into self's here. Returns the worker's item count so the caller
+    /// can total items across workers.
+    #[commonware_macros::stability(ALPHA)]
+    pub(crate) fn install_range(&mut self, worker: &mut Self) -> i64 {
```
patrick-ogrady: return `u64`? I think we have gauge helpers in utils that can help with this.

storage/src/index/partitioned/ordered/mod.rs
```diff
@@ -207,6 +222,92 @@ impl<T: Translator, V: Send + Sync, const P: usize> Index<T, V, P> {
...
+        let lo = worker.offset;
+        for (local, partition) in worker.partitions.iter_mut().enumerate() {
+            self.partitions[lo + local] = std::mem::take(partition);
+        }
+        // Drain only the partitions that actually spilled (remapping local -> global), rather than
```
patrick-ogrady, line 380: add \n before

storage/src/index/partitioned/ordered/mod.rs
```diff
@@ -207,6 +222,92 @@ impl<T: Translator, V: Send + Sync, const P: usize> Index<T, V, P> {
...
+        // probing every slot in the range.
+        for (local, inner) in worker.spilled.drain() {
+            self.spilled.insert(lo + local, inner);
+        }
+        // Fold in the worker's metric counts (a build worker's own handles are detached), matching
```
patrick-ogrady: add \n before

storage/src/index/partitioned/ordered/mod.rs
```diff
@@ -207,6 +222,92 @@ impl<T: Translator, V: Send + Sync, const P: usize> Index<T, V, P> {
...
+    /// snapshot-build worker, matching this index's translator and spill threshold. It allocates just
+    /// `count` partition slots but is still addressed by the global partition index (mapped to a local
+    /// slot in `get_mut`/`get_mut_or_insert`), so per-worker memory is the range, not the full
+    /// `2^(8*P)` -- which is what makes a large `P` affordable. Its metric handles are detached (never
+    /// registered): they only accumulate the counts that [`Self::install_range`] folds back into a
+    /// full index. Only `get_mut` and `get_mut_or_insert` (and their cursors) map the global
+    /// partition index to the local slot, so only those operations are valid on it. The other
+    /// `Unordered` methods index partitions globally and would touch the wrong slot.
+    #[commonware_macros::stability(ALPHA)]
+    pub(crate) fn new_range(&self, offset: usize, count: usize) -> Self {
```
patrick-ogrady: For this initial build, I think it would be safter to create a typesafe wrapper that didn't allow the broken methods to be called at all?

examples/sync/src/databases/any.rs
roberto-bayardo: changed to init_concurrency , nonzero, 1 == Serial, 2 == 1 "extra worker", etc.

storage/src/qmdb/mod.rs
```diff
@@ -413,6 +418,285 @@ where
...
+        init_concurrency: NonZeroUsize,
+        cache_size: Option<NonZeroUsize>,
+    ) -> Result<(usize, BitMap), Error<F>>
+    where
+        E: Spawner,
+        C: Contiguous<Item: Operation<F>> + 'static,
+    {
+        // This index type builds serially, so `init_concurrency` has no effect. Warn on an
+        // explicit concurrency rather than silently ignore it.
+        if init_concurrency.get() > 1 {
```
patrick-ogrady: Its a bit annoying that this mismatch can even happen at all ... I wonder if there is a way to tie this in with the config/specification to use a partitioned index.

patrick-ogrady (commented):
final comments

storage/src/qmdb/any/mod.rs
roberto-bayardo: Done in f47ee2ba2.

storage/src/qmdb/current/ordered/fixed.rs
roberto-bayardo: Done in f47ee2ba2.

storage/src/qmdb/current/ordered/variable.rs
roberto-bayardo: Done in f47ee2ba2.

storage/src/qmdb/current/mod.rs
roberto-bayardo: Done in f47ee2ba2.

storage/src/index/partitioned/ordered/mod.rs
roberto-bayardo: Done in f47ee2ba2.

storage/src/index/partitioned/ordered/mod.rs
roberto-bayardo: Done in f47ee2ba2.

storage/src/qmdb/mod.rs
roberto-bayardo: It was leaking the prometheus gauge type (`Gauge::get()` returns `i64`) through `install_range`. That now returns `usize` and the accumulator follows (f47ee2ba2).

storage/src/index/partitioned/ordered/mod.rs
roberto-bayardo: Done in f47ee2ba2, though as `usize` rather than `u64` since that's what `build_snapshot` returns — the count is in-memory entries, so it fits, and the cast at the call site disappears. The conversion out of the gauge is a checked `try_from`. I looked for the helpers: `GaugeExt` (runtime telemetry) only has setters (`try_set`/`try_set_max`), nothing for reading a gauge out.

storage/src/index/partitioned/ordered/mod.rs
roberto-bayardo: Done. `new_range` now returns a `RangeIndex` wrapper that exposes only `get_mut`/`get_mut_or_insert` (the ops that map a global partition index to the worker's local slot), so the globally-addressed methods can't be called on a worker at all. The `offset` field also leaves the full `Index` (its hot path no longer pays the subtraction), and workers drive the same collision-resolution helpers as the serial path through cursor-level functions (`update_at_cursor`/`delete_at_cursor`).

storage/src/qmdb/mod.rs
roberto-bayardo: could get rid of it by making the config an embedded type of Index but that adds a lot of extra cruft.

patrick-ogrady (dismissed):

storage/src/qmdb/mod.rs
```diff
@@ -536,6 +535,149 @@ where
...
+    let mut total_items = 0;
+    let mut active: BitMap = BitMap::zeroes(end - floor);
+    snapshot.for_each_value(|loc| {
+        active.set(**loc - floor, true);
+        total_items += 1;
+    });
+
+    // The last operation is the final commit (a log always ends with one), which stays active.
+    // An empty log has none.
+    if let Some(last_commit) = end.checked_sub(1)
```
patrick-ogrady, line 672: Made one small tweak here:

```
The bug: after installing the workers, the parallel build reconstructs the activity bitmap and then sets one extra bit for the last commit. The original code computed let last_commit = end.saturating_sub(1); and guarded the set with if last_commit >= floor. On an empty log, end == 0 and floor == 0 — and the saturating subtraction is exactly what broke it: instead of underflowing, last_commit saturated to 0, which spuriously passed the 0 >= 0 floor check, and active.set(0) then indexed into BitMap::zeroes(end - floor) — a zero-length bitmap — panicking with "bit 0 out of bounds (len: 0)". The serial path doesn't have the problem because it builds the bitmap incrementally per replayed operation: an empty log replays nothing and returns Ok with an empty bitmap. So the parallel path panicked exactly where the serial path succeeded — a behavioral divergence on the public SnapshotBuild trait, though unreachable through Db::init since the in-crate callers reject empty logs earlier.

The fix, in its final shape (after your note that guarding on end while subtracting last_commit was confusing):

// The last operation is the final commit (a log always ends with one), which stays active.
// An empty log has none.
if let Some(last_commit) = end.checked_sub(1)
    && last_commit >= floor
{
    active.set(last_commit - floor, true);
}

checked_sub makes the empty case fall out structurally — there's no saturated phantom value to sneak past the floor check. It's pinned by the two mutation-verified empty-log tests (one per partitioned variant): reverting the condition makes both fail with the original panic.
```

patrick-ogrady (dismissed):

patrick-ogrady (approved):

PR #4122, opened 2026-06-26
[storage] Read View Nits

patrick-ogrady:


runtime/src/utils/buffer/paged/writer.rs
```diff
@@ -1129,9 +1129,7 @@ impl<B: Blob> Writer<B> {
         let partial_page = if self.buffer.is_empty() {
             None
         } else {
-            let mut copy = self.cache_ref.pool().alloc(self.buffer.len());
-            copy.put_slice(self.buffer.as_ref());
-            Some(copy.freeze())
+            Some(self.buffer.slice(..))
```
patrick-ogrady, line 1132: This will now only actually copy if we need to write while holding a snapshot

PR #4109, opened 2026-06-25
[cryptography/hasher] Add Fixed Hasher

patrick-ogrady:
## Summary

- Adds a first-class `FixedHasher` opt-in for hashers used with fixed-size Merkle preimages.
- Adds fixed-shape helper methods on `Hasher` with default streaming implementations.
- Implements SHA-256 fixed-shape helpers with `sha2::block_api::compress256`.
- Ports BMT, Merkle `Standard`, and QMDB hasher bounds to use the fixed hashing path.

## Compatibility

This should not change any root or proof format. The fixed helpers hash the same concatenated preimages as the previous streaming path, including the same big-endian integer encodings.

The QMDB canonical-root golden test still passes, and SHA-256 has direct equivalence tests comparing the fixed helpers against the normal streaming hasher.

## Performance

Benchmarked SHA-256 fixed helpers against a temporary streaming fallback wrapper on the same Criterion benchmarks:

| Benchmark | Streaming | Fixed/compress | Speedup |
| --- | ---: | ---: | ---: |
| `merkle::append/n=10000 family=mmr` | 885.19 us | 651.70 us | 35.9% |
| `merkle::append/n=100000 family=mmr` | 9.5306 ms | 7.1803 ms | 32.5% |
| `merkle::append/n=10000 family=mmb` | 878.28 us | 682.40 us | 28.6% |
| `merkle::append/n=100000 family=mmb` | 9.3307 ms | 7.3749 ms | 26.4% |
| `bmt::build/n=10000` | 519.78 us | 429.51 us | 21.0% |
| `bmt::build/n=100000` | 5.1976 ms | 4.3068 ms | 20.6% |

## Validation

- `cargo check -p commonware-cryptography --tests`
- `cargo check -p commonware-storage --tests --benches`
- `just test -p commonware-cryptography test_fixed_hashes_match_standard_sha256`
- `just test -p commonware-storage bmt`
- `just test -p commonware-storage merkle::hasher`
- `just test -p commonware-storage qmdb::verify`
- `just test -p commonware-storage combine_roots`
- `just test -p commonware-storage qmdb::current::proof`

PR #4127, opened 2026-06-27
[storage] Boost Hashing Performance

patrick-ogrady:
Replaces: #4109

## Summary

- Adds `CodecHasher` extension methods for encoded and fixed-layout hash preimages, including `hash_parts`, `hash_encoded`, `hash_prefixed`, `hash_u32_digest`, `hash_digest_pair`, and `hash_u64_digest_pair`.
- Gives SHA-256 specialized fixed-preimage paths for the hot Merkle/BMT/QMDB layouts while keeping the byte preimages equivalent to the previous streaming `update(...).finalize()` construction.
- Reworks the `Hasher` streaming API around the `Pending` builder: `begin()` is the sole entry point for a multi-part hash (`hasher.begin().update(a).update(b).finalize()`), with `Pending::update`/`finalize` consuming so chains compose while use-after-finalize and double-finalize stay compile errors. Adds `StandardState` for reusable Merkle node/leaf hashing. The one-shot `H::hash` is unchanged.
- Updates BMT, Merkle, QMDB, coding, glue, and examples to use `CodecHasher` where callers need codec/fixed-preimage helpers.
- Removes redundant hasher arguments from QMDB and BMT verification-style APIs where the function can construct local scratch state from the hasher type.
- Keeps Merkle/QMDB/BMT commitment formats unchanged. Storage conformance remains unchanged.

## Findings

- The useful optimization in this version is fixed-preimage hashing, not a new commitment format. Internal Merkle nodes still hash the same bytes as before; the PR just avoids repeated streaming setup for common fixed-size preimages.
- `hash_digest_pair` is an ordinary `hash(left || right)` helper. Peak folding/root bagging also remains ordinary pair hashing where the root format expects `Hash(acc || peak)`.
- Passing hasher instances through QMDB/BMT verification APIs was unnecessary. The remaining explicit hasher values are internal reusable state or Merkle-family hashers whose bagging/family behavior is part of the API.

## Results

Benchmarks were rerun serially after merging current `main`, with separate fresh target directories because the normal environment uses a shared `CARGO_TARGET_DIR`. Main baselines are from detached `origin/main` worktree `/tmp/monorepo-4127-main-bench.SoJpeC` at `a04daf8a77e7` with `CARGO_TARGET_DIR=/tmp/cw-4127-main-bench-target.YzThOJ`; branch results are from this PR branch at `9891361f14f0` with `CARGO_TARGET_DIR=/tmp/cw-4127-branch-bench-target.w3noXX`.

Before running, the source paths were checked: the baseline BMT path is `Builder<H: Hasher>` with `update(...).finalize()`, while the branch BMT path is `Builder<H: CodecHasher>` with `hash_u32_digest` and `hash_digest_pair`. The isolated branch QMDB benchmark previously exposed stale benchmark-only `Hasher` imports, fixed in `f8b30379d`.

Commands, run from the corresponding worktree with the matching target directory:

- `env CARGO_TARGET_DIR=... cargo bench -p commonware-storage --bench bmt -- 'bmt::build/n=10000'`
- `env CARGO_TARGET_DIR=... cargo bench -p commonware-storage --features test-traits --bench qmdb -- 'qmdb::merkleize/variant=current::unordered::fixed::mmb chunk=32 keys=10000 ch=false sync=false'`

The table reports the second serial no-rebuild pass after both isolated targets had compiled:

| Benchmark | main median | branch median | Change |
| --- | ---: | ---: | ---: |
| `bmt::build/n=10000` | `540.81 us` | `442.91 us` | `-18.1%` |
| `bmt::build/n=100000` | `5.3482 ms` | `4.3953 ms` | `-17.8%` |
| `qmdb::merkleize/variant=current::unordered::fixed::mmb chunk=32 keys=10000 ch=false sync=false` | `936.31 us` | `660.88 us` | `-29.4%` |

Interpretation: the fresh isolated post-merge run shows a high-teens BMT build improvement and a roughly 29% QMDB merkleize spot-check improvement on this machine. Repeated BMT samples had some run-to-run variance, so these should be treated as directional single-machine Criterion results, not a 70% BMT/QMDB speedup claim.

## Validation

Validation of the `begin()`/consuming-`Pending` rework (full workspace, after merging current `main`):

- `cargo check --workspace --all-targets` (clean)
- `just test -p commonware-cryptography` (642 passed)
- `just test -p commonware-storage` (2677 passed)
- `just test-conformance -p commonware-storage` (unchanged)
- `just fix-fmt` (clean)

Post-merge validation:

- `env CARGO_TARGET_DIR=/tmp/cw-4127-merge-check-target.kpItZq cargo check -p commonware-storage --all-targets`
- `git diff --check`

Earlier validation before the `main` merge:

- `cargo check -p commonware-cryptography --all-targets`
- `cargo check -p commonware-storage --all-targets`
- `cargo check -p commonware-coding --all-targets`
- `cargo check -p commonware-glue --all-targets`
- `just test -p commonware-cryptography codec_hasher`
- `just test -p commonware-cryptography hash_parts`
- `just test -p commonware-cryptography hash_encoded`
- `just test -p commonware-storage bmt::`
- `just test -p commonware-storage merkle::hasher`
- `just test -p commonware-storage qmdb::verify`
- `just test -p commonware-storage qmdb::current::proof`
- `just test -p commonware-storage qmdb::current::grafting`
- `just test -p commonware-storage qmdb::sync::compact`
- `just test -p commonware-glue stateful::db::p2p::compact`
- `just test -p commonware-coding reed_solomon`
- `just test -p commonware-coding zoda`
- `just test-conformance -p commonware-storage`
- `git diff --check`
- `cargo fmt --all -- --check` (stable rustfmt reports the repo's existing nightly-only import-option warnings)


cryptography/src/sha256/mod.rs
```diff
@@ -67,21 +83,193 @@ impl Sha256 {
...
+
+#[inline]
+fn finalize_fixed_36(scratch: &mut [u8; SCRATCH_LEN]) -> Digest {
+    scratch[36] = 0x80;
+    scratch[37..56].fill(0);
+    write_scratch(&mut scratch[56..64], &[0, 0, 0, 0, 0, 0, 0x01, 0x20]);
+
+    let mut state = INITIAL_STATE;
+    let (blocks, remainder) = scratch[..BLOCK_LENGTH].as_chunks::<BLOCK_LENGTH>();
+    debug_assert!(remainder.is_empty());
```
patrick-ogrady: change to assert!

cryptography/src/lib.rs
```diff
@@ -275,6 +334,74 @@ commonware_macros::stability_scope!(BETA {
...
+        fn hash_digest_pair(
+            &mut self,
+            left: &Self::Digest,
+            right: &Self::Digest,
+        ) -> Self::Digest {
+            self.hash_parts([left.as_ref(), right.as_ref()])
+        }
+
+        /// Hash a `u64` followed by two digests.
+        #[doc(hidden)]
```
patrick-ogrady: not sure why these are hidden?

roberto-bayardo:
good suggestion from claude:

```
Add a direct constructor for StandardState and use it, so no fake Standard (and no meaningless bagging) is involved:

// merkle/hasher.rs
impl<H: CodecHasher> StandardState<H> {
    /// Create reusable hashing state.
    pub(crate) fn new() -> Self {
        Self { hasher: H::new() }
    }
    // ... node_digest, leaf_encoded ...
}

// optionally make the existing convenience delegate, documenting that bagging is irrelevant:
pub(crate) fn state(&self) -> StandardState<H> { StandardState::new() }

// merkle/batch.rs
let mut state = StandardState::<H>::new();   // line 461
|| StandardState::<H>::new(),                // line 474

That also lets the non-test import on line 87 drop both names it no longer needs:

-#[cfg(feature = "std")]
-use crate::merkle::{hasher::Standard, Bagging};
+#[cfg(feature = "std")]
+use crate::merkle::hasher::StandardState;

Net effect: the reusable-state construction stops mentioning bagging entirely, which is honest about the fact that this loop is bagging-independent — no behavior change, identical digests, conformance unaffected. Standard::state() stays for the two callers that legitimately have a configured Standard.
```

cryptography/src/lib.rs
```diff
@@ -275,6 +334,74 @@ commonware_macros::stability_scope!(BETA {
             Self::new().update(message).finalize()
         }
     }
+
+    /// Extension methods for hashing encoded values and fixed preimages.
+    pub trait CodecHasher: Hasher {
```
cronokirby, line 333: What if we added a blanket impl of this trait? That would obviate the need to then test that the impls of the trait do the same thing as the default impl.

cryptography/src/sha256/mod.rs
roberto-bayardo: done

cryptography/src/lib.rs
roberto-bayardo: Good call — un-hid them; they're public API the storage crate uses, so they should be documented. (Kept `update_inner`/`finalize_inner` hidden — those are internal hooks called via `Pending`.)

cryptography/src/lib.rs
roberto-bayardo: A blanket impl would stop `Sha256` from overriding these with its fixed-block fast paths (the optimization in this PR), and the equivalence tests guard exactly those overrides — so they're not redundant. Keeping per-type impls.

PR #4128, opened 2026-06-27
[storage/merkle] Positionless node hashing with domain-separated merge

roberto-bayardo:
## Summary

Stacked on #4127 (base branch `hashing-boost`). Drops the node position from internal-node hashing and keys leaves by **location** instead of node position. The root already commits the leaf count, so the tree shape is fixed and the position is redundant in node digests — binding reduces to collision resistance of the location-keyed leaves plus the committed count.

Internal nodes use a single-block `merge_digest_pair` (one SHA-256 compression vs the two of the previous `hash_u64_digest_pair` node hash), domain-separated from standard hashing by a distinct compression IV (`MERGE_IV`). This gives a **uniform 128-bit leaf/node separation at every element length**, with no per-leaf tag and no element pre-hash.

Beyond the core change, this PR also makes the merkle `Hasher` trait fully position-free (grafting's chunk-fusion becomes an explicit combiner), closes a BLAKE3 domain-separation gap exposed by the positionless change and makes `merge_digest_pair` a required method so the gap can't recur, and fixes the `current::Db` verify callers left non-compiling by the base's hasher-param removal. See **Commits**.

## Why it's safe

- **Inclusion soundness** reduces to collision resistance: the verifier walks a shape fixed by the committed leaf count and only ever compares leaf-vs-leaf and node-vs-node (never cross-type), so node-position binding isn't required.
- **Leaf/node domain separation** is by construction: `merge_digest_pair` derives from a distinct IV (SHA-256) / key-derivation mode (BLAKE3), so a node digest can't equal a `hash`-derived leaf digest even when their 64-byte blocks coincide. (Same technique SHA-224 / SHA-512-256 use.) The method has **no default** — every hasher implements it explicitly, so a cryptographic hasher can't silently inherit an unsafe (non-separated) merge. Non-cryptographic hashers (checksums) delegate to `hash_digest_pair` and must not be used in authenticated structures.

## Performance

- **Primitive:** the node hash is now a single SHA-256 compression (`merge_digest_pair`, ~14.5 ns) vs the two-block previous node hash (`hash_u64_digest_pair`, ~30 ns) — roughly 2× faster. The `sha256::digest_pair` microbench shows the single- vs double-block gap (`merge_digest_pair` vs the full-hash `hash_digest_pair`).
- **Build-level:** small-to-neutral and within measurement noise. Node hashing is a fraction of a build and, on the parallel path, divided across cores, so the 2× doesn't show as a headline (`qmdb::merkleize` and the new `merkle::build` bench are within noise). No regression.
- Aside: the bigger build lever is the bulk path itself — `merkle::build` (bulk `add_leaf_digests` + parallel merkleize) is ~36% lighter than the per-leaf `append`, and is the path qmdb already uses.

_(Perf figures are from the in-repo microbenches; reproduce with `cargo bench -p commonware-cryptography` / `-p commonware-storage`.)_

## Commits

1. **`[storage/merkle] Positionless node hashing with domain-separated merge`** — the core change above. Location-keyed `leaf_digest`/`leaf_encoded`; `Standard::node_digest` via `merge_digest_pair` from `MERGE_IV` (= `SHA256("_COMMONWARE_CRYPTOGRAPHY_MERGE_DIGEST_PAIR")`, pinned by a test); removed `GraftedHasher` / `map_pos` / `merkleize_reusing_with` and the old `hash_u64_digest_pair` primitive; exposed `add_leaf_digests`; new `merkle::build` + `sha256::digest_pair` benches; conformance regenerated.
2. **`[storage/qmdb] Drop redundant hasher arg from current::Db verify callers`** — the base already removed `&hasher` from `current::Db::verify_*` but left the glue + `current_*` fuzz callers passing it (non-compiling). This updates them; required for the branch to build.
3. **`[cryptography] Domain-separate BLAKE3 merge_digest_pair`** — BLAKE3 inherited the unsafe default merge (plain `hash(l || r)`), so `Standard<Blake3>` could let an internal-node digest be reinterpreted as a leaf preimage without a hash collision. Adds a safe merge via BLAKE3's key-derivation mode (with a test) and clarifies the contract.
4. **`[storage/merkle] Position-free Hasher trait; grafting via explicit combiner`** — removes `node_digest_at` from the `Hasher` trait entirely; proof reconstruction takes an explicit combiner (`reconstruct_root_with`, defaulting to `Hasher::node_digest`); the grafting `Verifier` becomes a plain combiner (no longer a `Hasher` impl), scoped `pub(crate)`. The position-keyed grafting concern leaves the generic trait.
5. **`[cryptography] Require merge_digest_pair (remove unsafe default)`** — the plain-hash default was a silent fail-open for any cryptographic hasher that forgot to override it (the BLAKE3 gap above). Removes the default so the method is required; SHA-256/BLAKE3 already override safely, CRC32 (non-cryptographic, no collision resistance to separate anyway) delegates explicitly with a no-separation note.

## Notes

- ALPHA crates — breaking storage-format change, no migration path needed.
- **Draft:** for the `hashing-boost` author to judge whether the simplification (−net core LOC: `GraftedHasher` / `map_pos` / `node_digest_at` gone) + uniform domain separation are worth folding in. The perf is a real but small primitive win, not a headline.
- Commit 2 (verify-caller fix) completes a loose end of the base's own hasher-param removal, so it may belong in `hashing-boost` rather than here.
- A follow-up PR will internalize the canonical hasher in `current::Db`'s proof-*generation* methods too, mirroring the now-hasher-free `verify_*` methods — kept separate to avoid a ~45-site mechanical sweep bloating this PR.
- The `digest_pair` microbench and `build` bench can be dropped if undesired.

### DIFF
--- a/cryptography/src/crc32/mod.rs
+++ b/cryptography/src/crc32/mod.rs
@@ -92,6 +92,13 @@ impl Hasher for Crc32 {
         self.inner = crc_fast::Digest::new(ALGORITHM);
         self
     }
+
+    /// CRC32 is a non-cryptographic checksum and cannot provide the domain separation
+    /// [`Hasher::merge_digest_pair`] requires; this delegates to [`Hasher::hash_digest_pair`] and
+    /// must not be relied upon for authenticated structures.
+    fn merge_digest_pair(&mut self, left: &Self::Digest, right: &Self::Digest) -> Self::Digest {
+        self.hash_digest_pair(left, right)
+    }
 }
 
 /// Digest of a CRC32 hashing operation (4 bytes).
--- a/cryptography/src/blake3/mod.rs
+++ b/cryptography/src/blake3/mod.rs
@@ -39,6 +39,10 @@ pub type CoreBlake3 = blake3::Hasher;
 
 const DIGEST_LENGTH: usize = blake3::OUT_LEN;
 
+/// Context string for [`Blake3::merge_digest_pair`]'s key-derivation mode, domain-separating
+/// digest-pair hashing from standard hashing.
+const MERGE_CONTEXT: &str = "_COMMONWARE_CRYPTOGRAPHY_MERGE_DIGEST_PAIR";
+
 /// BLAKE3 hasher.
 #[cfg_attr(
     feature = "blake3-parallel",
@@ -87,6 +91,17 @@ impl Hasher for Blake3 {
     fn reset(&mut self) -> &mut Self {
         self.hasher = CoreBlake3::new();
         self
+    }
+
+    fn merge_digest_pair(&mut self, left: &Self::Digest, right: &Self::Digest) -> Self::Digest {
+        // Domain-separate digest-pair hashing from standard hashing via BLAKE3's key-derivation
+        // mode. Its finalization flags differ from a plain `hash`, so a merge output cannot equal a
+        // `hash` output except by a generic collision, uniformly for any input length.
+        let mut hasher = CoreBlake3::new_derive_key(MERGE_CONTEXT);
+        hasher.update(left.as_ref());
+        hasher.update(right.as_ref());
+        let array: [u8; DIGEST_LENGTH] = hasher.finalize().into();
+        Self::Digest::from(array)
     }
 }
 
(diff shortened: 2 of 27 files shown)


cryptography/src/crc32/mod.rs
```diff
@@ -92,6 +92,13 @@ impl Hasher for Crc32 {
         self.inner = crc_fast::Digest::new(ALGORITHM);
         self
     }
+
+    /// CRC32 is a non-cryptographic checksum and cannot provide the domain separation
+    /// [`Hasher::merge_digest_pair`] requires; this delegates to [`Hasher::hash_digest_pair`] and
+    /// must not be relied upon for authenticated structures.
+    fn merge_digest_pair(&mut self, left: &Self::Digest, right: &Self::Digest) -> Self::Digest {
```
patrick-ogrady, line 99: I think the right call here is to instead add a `MergeHasher` trait that extends the normal hasher?

PR #4129, opened 2026-06-28
[storage/journal] Address Final Nits

patrick-ogrady:
## Summary

- Address final read-view nits on top of `danlaine/read-view`.
- Optimize contiguous journal replay by draining decoded batches synchronously instead of yielding each decoded item through a separate async stream transition.
- Keep the concrete-reader/memmove prototype out of this PR because it did not improve the fixed replay worst case locally and added broader replay plumbing.

## Replay Performance Investigation

The replay regression was traced to the shared replay stream driver. The previous branch implementation decoded a batch, stored it in a `VecDeque`, then yielded one item per `stream::unfold` transition. `origin/main` decoded a batch asynchronously and flattened it synchronously with `stream::iter`, avoiding that per-item async state-machine cost.

This PR now keeps the shared replay driver but changes it to yield decoded batches from `next_batch()` and drain each batch with `flat_map(stream::iter)`. Stop-after-error behavior is preserved by marking the replay stream done when a yielded batch contains an error.

Local Criterion result for the worst fixed replay case:

| Case | Time |
| --- | ---: |
| branch before replay-driver fix | ~27.4 ms |
| `origin/main` | ~10.0 ms |
| batch-driver fix | ~7.6 ms |

Benchmark filter:

```sh
cargo bench -p commonware-storage --bench journal -- \
  'journal::fixed_replay/items=500000 buffer=1048576 size=32'
```

Additional local replay checks were also favorable: fixed replay at `items=100000 buffer=1048576 size=32` was around `1.5 ms`, and variable replay at `items=100000 buffer=65536 size=32` was around `3.9 ms`.

I also prototyped a concrete-reader version that removed the `ReplayInner` enum from the item decode loop by splitting paged sealed replay from live tail replay. It passed the focused replay tests, but it stayed around `7.6 ms` on the fixed replay worst case and regressed the smaller fixed replay case locally, so that larger change is intentionally not included here.

## Validation

```sh
cargo check -p commonware-storage --lib
just test -p commonware-storage test_fixed_journal_replay
just test -p commonware-storage test_variable_replay
cargo fmt --all --check
git diff --check
```

### DIFF
--- a/storage/src/journal/frame.rs
+++ b/storage/src/journal/frame.rs
@@ -4,9 +4,50 @@
 //! zstd-compressed) encoded item.
 
 use super::Error;
-use commonware_codec::{varint::UInt, Codec, EncodeSize, ReadExt as _, Write as _};
-use commonware_runtime::Buf;
+use commonware_codec::{
+    varint::{UInt, MAX_U32_VARINT_SIZE},
+    Codec, EncodeSize, ReadExt as _, Write as _,
+};
+use commonware_runtime::{Blob, Buf, IoBufMut, IoBufs};
+use std::{future::Future, io::Cursor};
 use zstd::{bulk::compress, decode_all};
+
+/// Read access needed to decode a frame at a known offset.
+pub(super) trait FrameReader {
+    /// Read up to `len` bytes at `offset`.
+    fn read_up_to(
+        &self,
+        offset: u64,
+        len: usize,
+        buf: impl Into<IoBufMut> + Send,
+    ) -> impl Future<Output = Result<(IoBufMut, usize), Error>> + Send;
+
+    /// Read exactly `len` bytes at `offset`.
+    fn read_at(
+        &self,
+        offset: u64,
+        len: usize,
+    ) -> impl Future<Output = Result<IoBufs, Error>> + Send;
+}
+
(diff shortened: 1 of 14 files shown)


storage/src/ordinal/storage.rs
```diff
@@ -426,13 +426,16 @@ impl<E: BufferPooler + Context, V: CodecFixed<Cfg = ()>> Ordinal<E, V> {
             return Ok(());
         }
 
-        let futures: Vec<_> = self
```
patrick-ogrady, line 429: Revert back?

storage/src/journal/frame.rs
```diff
@@ -4,10 +4,51 @@
...
+
+    /// Read exactly `len` bytes at `offset`.
+    fn read_at(
+        &self,
+        offset: u64,
+        len: usize,
+    ) -> impl Future<Output = Result<IoBufs, Error>> + Send;
+}
+
+impl<B: Blob> FrameReader for commonware_runtime::buffer::paged::Writer<B> {
```
patrick-ogrady: Can we just import Writer directly here?

patrick-ogrady (commented):
almost there

patrick-ogrady:
bugbot run

PR #4130, opened 2026-06-29
[strategy] Automatically Tune Behavior

patrick-ogrady:
## Summary

- Adds an adaptive policy to `commonware_parallel::Rayon` so `Strategy` operations can choose serial or parallel execution per callsite.
- Keys decisions by `#[track_caller]` location, operation kind, input length bucket, and Rayon parallelism.
- Samples both parallel and serial execution, tracks EWMA timings, and periodically re-samples to adapt to workload/platform changes.
- Removes the QMDB floor-candidate manual chunking at the callsite; candidate classification now maps over the actual candidate stream and lets `Strategy` choose execution.

## Performance

QMDB merkleize benchmarks were rerun with isolated Cargo target directories to avoid reusing the wrong benchmark binary from the shared build cache.

- PR head: `f7f7449e00a8c686c1d667a19d78418fb7bbef1d`
- Base head: `23b224eedb64d8ee1be8d47f42b404eebadaf088`
- PR target dir: `/tmp/qmdb-target-automatic-rayon`
- Base target dir: `/tmp/qmdb-target-main`
- Command template:

```bash
RUSTFLAGS='--cfg full_bench' \
CARGO_TARGET_DIR=<isolated-target> \
CRITERION_HOME=<isolated-criterion-home> \
cargo bench -q -p commonware-storage --features test-traits --bench qmdb -- '<filter>'
```

Lower is better:

| QMDB merkleize filter | Base | This PR | Improvement |
|---|---:|---:|---:|
| `current::unordered::fixed::mmb chunk=32 keys=10000 ch=false sync=false` | 919.97 us | 498.83 us | 45.8% |
| `current::unordered::fixed::mmb chunk=32 keys=100000 ch=false sync=false` | 4.6462 ms | 4.0655 ms | 12.5% |
| `current::unordered::fixed::mmb chunk=256 keys=100000 ch=false sync=false` | 4.3651 ms | 3.9045 ms | 10.6% |
| `current::ordered::fixed::mmb chunk=32 keys=10000 ch=false sync=false` | 1.0034 ms | 575.62 us | 42.6% |
| `any::unordered::fixed::mmb keys=10000 ch=false sync=false` | 731.34 us | 413.09 us | 43.5% |

I also prototyped moving coarse chunking into `parallel` for vector-map helpers. It did not materially improve the simplified QMDB path, so that broader behavior change was left out.

## Validation

- `cargo test -p commonware-parallel`
- `cargo check -p commonware-storage --all-targets --features test-traits`
- `cargo fmt --all` (emits the repo's existing stable-rustfmt warnings for nightly-only rustfmt options)

### DIFF
--- a/coding/src/reed_solomon.rs
+++ b/coding/src/reed_solomon.rs
@@ -758,12 +758,10 @@ fn verify_root<H: Hasher, S: Strategy>(
         .collect::<Vec<_>>();
 
     let manual = strategy.manual();
-    for (i, digest) in
-        manual.map_init_collect_vec(missing_shards, H::new, |hasher, (i, shard)| {
-            hasher.update(shard);
-            (i, hasher.finalize())
-        })
-    {
+    for (i, digest) in manual.map_init_collect_vec(missing_shards, H::new, |hasher, (i, shard)| {
+        hasher.update(shard);
+        (i, hasher.finalize())
+    }) {
         shard_digests[i] = Some(digest);
     }
 
--- a/coding/src/reed_solomon.rs
+++ b/coding/src/reed_solomon.rs
@@ -1579,7 +1579,8 @@ mod tests {
         // exercise the sequential path.
         let shard_len = canonical_shard_len(data.len(), min as usize);
         assert!(
-            striped::ranges(shard_len, strategy.manual().parallelism_hint()).map_or(0, |r| r.len()) >= 2,
+            striped::ranges(shard_len, strategy.manual().parallelism_hint()).map_or(0, |r| r.len())
+                >= 2,
             "test must exercise >= 2 stripes (shard_len={shard_len})"
         );
 
@@ -1739,7 +1740,8 @@ mod tests {
         let shard_len = canonical_shard_len(data.len(), min as usize);
         let rayon = Rayon::new(NZUsize!(4)).unwrap();
         assert!(
-            striped::ranges(shard_len, rayon.manual().parallelism_hint()).map_or(0, |r| r.len()) >= 2,
+            striped::ranges(shard_len, rayon.manual().parallelism_hint()).map_or(0, |r| r.len())
+                >= 2,
             "test must exercise >= 2 stripes (shard_len={shard_len})"
         );
 
@@ -2169,7 +2171,8 @@ mod tests {
         let min = 4u16;
         let shard_len = canonical_shard_len(data.len(), min as usize);
         assert!(
-            striped::ranges(shard_len, strategy.manual().parallelism_hint()).map_or(0, |r| r.len()) >= 2,
+            striped::ranges(shard_len, strategy.manual().parallelism_hint()).map_or(0, |r| r.len())
+                >= 2,
             "test must exercise the striped path (shard_len={shard_len})"
         );
 
--- a/coding/src/reed_solomon.rs
+++ b/coding/src/reed_solomon.rs
@@ -378,10 +378,15 @@ fn encode<H: Hasher, S: Strategy>(
         .map(|i| originals.slice(i * shard_len..(i + 1) * shard_len))
         .chain((0..m).map(|i| recoveries.slice(i * shard_len..(i + 1) * shard_len)))
         .collect();
-    let shard_hashes = manual.map_init_collect_vec(&shard_slices, H::new, |hasher, shard| {
-        hasher.update(shard);
-        hasher.finalize()
-    });
+    let shard_hashes = strategy.map_init_collect_vec_with_multiplier(
+        &shard_slices,
+        shard_len,
+        H::new,
+        |hasher, shard| {
+            hasher.update(shard);
+            hasher.finalize()
+        },
+    );
     for hash in &shard_hashes {
         builder.add(hash);
     }
@@ -757,11 +762,15 @@ fn verify_root<H: Hasher, S: Strategy>(
         })
         .collect::<Vec<_>>();
 
-    let manual = strategy.manual();
-    for (i, digest) in manual.map_init_collect_vec(missing_shards, H::new, |hasher, (i, shard)| {
-        hasher.update(shard);
-        (i, hasher.finalize())
-    }) {
+    for (i, digest) in strategy.map_init_collect_vec_with_multiplier(
+        missing_shards,
+        shard_len,
+        H::new,
+        |hasher, (i, shard)| {
+            hasher.update(shard);
+            (i, hasher.finalize())
+        },
+    ) {
         shard_digests[i] = Some(digest);
     }
 
--- a/coding/src/reed_solomon.rs
+++ b/coding/src/reed_solomon.rs
@@ -333,13 +333,14 @@ fn encode<H: Hasher, S: Strategy>(
     let (padded, shard_len) = prepare_data(data, k);
 
     // Compute recovery shards, striping large shard widths across the strategy
-    let recovery_buf = match striped::ranges(shard_len, strategy.parallelism_hint()) {
+    let manual = strategy.manual();
+    let recovery_buf = match striped::ranges(shard_len, manual.parallelism_hint()) {
         Some(ranges) => {
             let original_shards = padded.chunks(shard_len).collect::<Vec<_>>();
             let mut buf = vec![0u8; m * shard_len];
             let groups = striped::stripe_columns(&mut buf, shard_len, &ranges);
             let stripes: Vec<_> = ranges.into_iter().zip(groups).collect();
-            strategy.try_map_collect_vec(stripes, |(range, out)| {
+            manual.try_map_collect_vec(stripes, |(range, out)| {
                 striped::encode_recovery_into(k, m, range, &original_shards, out)
             })?;
             buf
@@ -946,7 +947,8 @@ fn decode<'a, H: Hasher, S: Strategy>(
 
     // Process checked chunks
     let shard_len = first.shard.len();
-    let stripes = striped::ranges(shard_len, strategy.parallelism_hint());
+    let manual = strategy.manual();
+    let stripes = striped::ranges(shard_len, manual.parallelism_hint());
     let mut shard_digests: Vec<Option<H::Digest>> = vec![None; n];
     let mut provided_originals: Vec<(usize, &[u8])> = Vec::new();
     let mut provided_recoveries: Vec<(usize, &[u8])> = Vec::new();
@@ -1001,6 +1003,34 @@ fn decode<'a, H: Hasher, S: Strategy>(
     }
 
     // Decode the data, striping the work across the strategy when shards are large enough
+    if let Some(ranges) = stripes {
+        let ctx = DecodeCtx {
+            n,
+            k,
+            m,
+            shard_len,
+            root,
+            strategy: &manual,
+        };
+        // Recovery reads the missing originals and recoveries straight out of one decode;
+        // with all originals present there is nothing to decode, so re-encode instead.
+        if recovery_needed {
+            return striped::decode_reveal::<H, _>(
+                &ctx,
+                ranges,
+                shard_digests,
+                provided_originals,
+                provided_recoveries,
+            );
+        }
+        return striped::decode::<H, _>(
+            &ctx,
+            ranges,
+            shard_digests,
+            provided_originals,
+            provided_recoveries,
+        );
+    }
     let ctx = DecodeCtx {
         n,
         k,
@@ -1009,26 +1039,6 @@ fn decode<'a, H: Hasher, S: Strategy>(
         root,
         strategy,
     };
-    if let Some(ranges) = stripes {
-        // Recovery reads the missing originals and recoveries straight out of one decode;
-        // with all originals present there is nothing to decode, so re-encode instead.
-        if recovery_needed {
-            return striped::decode_reveal::<H, S>(
-                &ctx,
-                ranges,
-                shard_digests,
-                provided_originals,
-                provided_recoveries,
-            );
-        }
-        return striped::decode::<H, S>(
-            &ctx,
-            ranges,
-            shard_digests,
-            provided_originals,
-            provided_recoveries,
-        );
-    }
     sequential::decode::<H, S>(&ctx, shard_digests, provided_originals, provided_recoveries)
 }
 
@@ -1569,7 +1579,7 @@ mod tests {
         // exercise the sequential path.
         let shard_len = canonical_shard_len(data.len(), min as usize);
         assert!(
-            striped::ranges(shard_len, strategy.parallelism_hint()).map_or(0, |r| r.len()) >= 2,
+            striped::ranges(shard_len, strategy.manual().parallelism_hint()).map_or(0, |r| r.len()) >= 2,
             "test must exercise >= 2 stripes (shard_len={shard_len})"
         );
 
@@ -1729,7 +1739,7 @@ mod tests {
         let shard_len = canonical_shard_len(data.len(), min as usize);
         let rayon = Rayon::new(NZUsize!(4)).unwrap();
         assert!(
-            striped::ranges(shard_len, rayon.parallelism_hint()).map_or(0, |r| r.len()) >= 2,
+            striped::ranges(shard_len, rayon.manual().parallelism_hint()).map_or(0, |r| r.len()) >= 2,
             "test must exercise >= 2 stripes (shard_len={shard_len})"
         );
 
@@ -2159,7 +2169,7 @@ mod tests {
         let min = 4u16;
         let shard_len = canonical_shard_len(data.len(), min as usize);
         assert!(
-            striped::ranges(shard_len, strategy.parallelism_hint()).map_or(0, |r| r.len()) >= 2,
+            striped::ranges(shard_len, strategy.manual().parallelism_hint()).map_or(0, |r| r.len()) >= 2,
             "test must exercise the striped path (shard_len={shard_len})"
         );
 
--- a/coding/src/reed_solomon.rs
+++ b/coding/src/reed_solomon.rs
@@ -378,7 +378,7 @@ fn encode<H: Hasher, S: Strategy>(
         .map(|i| originals.slice(i * shard_len..(i + 1) * shard_len))
         .chain((0..m).map(|i| recoveries.slice(i * shard_len..(i + 1) * shard_len)))
         .collect();
-    let shard_hashes = strategy.map_init_collect_vec(&shard_slices, H::new, |hasher, shard| {
+    let shard_hashes = manual.map_init_collect_vec(&shard_slices, H::new, |hasher, shard| {
         hasher.update(shard);
         hasher.finalize()
     });
@@ -757,8 +757,9 @@ fn verify_root<H: Hasher, S: Strategy>(
         })
         .collect::<Vec<_>>();
 
+    let manual = strategy.manual();
     for (i, digest) in
-        strategy.map_init_collect_vec(missing_shards, H::new, |hasher, (i, shard)| {
+        manual.map_init_collect_vec(missing_shards, H::new, |hasher, (i, shard)| {
             hasher.update(shard);
             (i, hasher.finalize())
         })
--- a/p2p/src/utils/codec.rs
+++ b/p2p/src/utils/codec.rs
@@ -120,13 +120,8 @@ impl<R: Receiver, V: Codec> WrappedReceiver<R, V> {
 /// A background receiver that receives raw bytes from a [`Receiver`] and spawns concurrent
 /// decode tasks using a [`Codec`].
 ///
-/// Decode work is submitted to the provided [`Strategy`]. A multi-worker [`Rayon`](commonware_parallel::Rayon)
-/// strategy runs each decode on a pool worker, pipelining network I/O (receiving bytes) with CPU
-/// work (decoding messages) so the receive loop keeps draining the network buffer; this is
-/// particularly useful for expensive decodes such as signature verification or decryption. Note
-/// that decode then shares the strategy's thread pool with the rest of that strategy's work.
-/// Strategies that execute inline ([`Sequential`](commonware_parallel::Sequential) or a
-/// single-worker pool) decode synchronously on the receive loop, with no pipelining.
+/// Decode work is submitted to the provided [`Strategy`], so callers can offload expensive
+/// decodes from the receive loop by choosing a parallel strategy.
 ///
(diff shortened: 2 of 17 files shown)


parallel/src/lib.rs
```diff
@@ -590,16 +607,165 @@ commonware_macros::stability_scope!(BETA, cfg(feature = "std") {
...
     pub type ThreadPool = Arc<RThreadPool>;
 
-    /// Minimum slice length before [`Rayon::sort_by`] dispatches to the thread pool.
-    ///
-    /// Mirrors rayon's `CHUNK_LENGTH`: `par_sort_by` sorts slices of up to one chunk
-    /// sequentially, so dispatching them pays the pool handoff for no parallelism.
-    const MIN_PARALLEL_SORT: usize = 2000;
+    const PREFERRED_SAMPLE_INTERVAL: u32 = 8;
+    const RESAMPLE_INTERVAL: u32 = 64;
+    const EWMA_ALPHA: f64 = 0.2;
```
patrick-ogrady: it would be great to make this use int math (to keep deterministic)?

parallel/src/lib.rs
```diff
@@ -750,40 +1071,76 @@ commonware_macros::stability_scope!(BETA, cfg(feature = "std") {
...
             RB: Send,
         {
-            self.thread_pool.install(|| rayon::join(a, b))
+            let (key, execution, start) = self.choose(OperationKind::Join, Location::caller(), 2);
+            let result = match execution {
+                Execution::Serial => (a(), b()),
+                Execution::Parallel => self.thread_pool.install(|| rayon::join(a, b)),
+            };
+            self.record(key, execution, start);
+            result
```
clabby: I'm torn on this.. `rayon::join` already documents that blocking one closure on the other could lead to deadlock @ https://docs.rs/rayon/latest/rayon/fn.join.html#warning-about-blocking-io, so this is not really a breaking API change. That said, there is a subtle change here, which is that the `Rayon` strategy no longer guarantees that both closures are executed since it can run them sequentially.

On a stronger note, the policy can only learn "this callsite’s two closures were faster serial or parallel recently" which is a much weaker signal than the other functions have because closure cost can vary dramatically call to call.

Leaning towards keeping this one as-is. Rayon’s primitive already handles the "maybe parallel, maybe not" part - see https://docs.rs/rayon/latest/rayon/fn.join.html

cronokirby:
I wonder if we might be able to have one entry point for tracking, to avoid having to repeat the logic in every method, and because I don't see why different methods would need to be tracked differently per se...

parallel/src/lib.rs
patrick-ogrady: Ah, I missed that blocking io thing. I'm ok to only do this for some of the others at first.

cronokirby:
I'd also be curious to explore a bit more why we see an improvement with this change for these benchmarks: what makes it so that simply dividing up the work evenly between threads is worse? Is there an overhead for a small amount of work?

patrick-ogrady:
> I'd also be curious to explore a bit more why we see an improvement with this change for these benchmarks: what makes it so that simply dividing up the work evenly between threads is worse? Is there an overhead for a small amount of work?

Yes, there is non-trivial overhead for even small work. The QMDB code benefits most because we do layer-by-layer caching and many of those layers are too small to benefit.

parallel/src/lib.rs
```diff
@@ -649,12 +652,26 @@ commonware_macros::stability_scope!(BETA, cfg(feature = "std") {
...
+                len,
+                self.thread_pool.current_num_threads(),
+                run,
+            )
         }
     }
 
     impl Strategy for Rayon {
+        #[track_caller]
         fn fold_init<I, INIT, T, R, ID, F, RD>(
```
patrick-ogrady, line 989: This doesn't work well if the caller pre-splits on cores based on something? then, the autotuner will always see `cores` items but with wildly different execution results.

p2p/src/utils/codec.rs
```diff
@@ -157,42 +155,66 @@ impl<P: PublicKey, V> BackgroundReceiver<P, V> {
     }
 }
 
-pub struct WrappedBackgroundReceiver<E, P, B, R, V>
+struct DecodeQueue<T> {
```
patrick-ogrady: I think this can be removed...

parallel/src/policy.rs
```diff
@@ -0,0 +1,272 @@
...
+// Refresh the preferred path periodically so its EWMA does not go stale.
+const PREFERRED_SAMPLE_INTERVAL: u32 = 8;
+// Probe the non-preferred path on a longer interval in case conditions change.
+const RESAMPLE_INTERVAL: u32 = 64;
+// Track a short EWMA so recent measurements outweigh old startup noise.
+const EWMA_PREVIOUS_WEIGHT: u64 = 4;
+const EWMA_NEXT_WEIGHT: u64 = 1;
+const EWMA_WEIGHT: u64 = EWMA_PREVIOUS_WEIGHT + EWMA_NEXT_WEIGHT;
+// Serial must take at most 95% of parallel's time before it becomes preferred.
+const SERIAL_WIN_NUMERATOR: u64 = 95;
```
patrick-ogrady: make this 90?

storage/src/qmdb/any/batch.rs
```diff
@@ -38,9 +38,6 @@ use tracing::debug;
 type DiffVec<K, F, V> = Vec<(K, DiffEntry<F, V>)>;
 type DiffSlice<K, F, V> = [(K, DiffEntry<F, V>)];
 
-/// One contiguous chunk of floor-raise candidates paired with their resolved operations.
-type CandidateChunk<'a, F, U> = (&'a [Location<F>], &'a [Operation<F, U>]);
```
patrick-ogrady, line 42: Restore this chunk-driven optimization? do we get some benefit to retaining this?

p2p/src/utils/codec.rs
```diff
@@ -218,54 +240,39 @@ where
...
-                    .spawn(|_| async move {
-                        let result = V::decode_cfg(bytes.as_ref(), &config);
-                        (peer, result)
-                    });
-                decode_pool.push(handle);
+                let handle = self.strategy.spawn(move || {
+                    let result = V::decode_cfg(bytes.as_ref(), &config);
+                    (peer, result)
+                });
+                decode_queue.push(handle);
```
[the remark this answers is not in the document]
patrick-ogrady: This is bound by rate limiting. This is fine.

parallel/src/lib.rs
```diff
@@ -519,6 +685,19 @@ commonware_macros::stability_scope!(BETA {
...
+            Manual::new(Self, NonZeroUsize::new(1).unwrap())
+        }
+
+        fn spawn<F, T>(&self, f: F) -> impl core::future::Future<Output = T> + Send + 'static
+        where
+            F: FnOnce() -> T + Send + 'static,
+            T: Send + 'static,
+        {
+            let result = f();
+            async move { result }
```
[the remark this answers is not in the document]
patrick-ogrady: This is expeced.

patrick-ogrady:
butbot run

parallel/src/policy.rs
```diff
@@ -0,0 +1,278 @@
+//! Adaptive execution policy for Rayon collection operations.
```
patrick-ogrady: remove `Rayon`

parallel/src/policy.rs
```diff
@@ -0,0 +1,278 @@
...
+        caller: &'static Location<'static>,
+        len: usize,
+        parallelism: usize,
+        run: impl FnOnce(Execution) -> R,
+    ) -> R {
+        // A single-threaded pool cannot benefit from rayon scheduling, so always run serial and
+        // never spend a measurement on it.
+        if parallelism <= 1 {
+            return run(Execution::Serial);
+        }
```
patrick-ogrady, line 73: add a \n after this

parallel/src/policy.rs
```diff
@@ -0,0 +1,278 @@
...
+
+impl Entry {
+    // Returns the path to run and whether the caller should time it and feed the elapsed duration
+    // back to [`record`](Self::record).
+    fn choose(&mut self) -> (Execution, bool) {
+        // Seed both sides before trusting the comparison.
+        if self.parallel_samples == 0 {
+            return (Execution::Parallel, true);
+        }
+        if self.serial_samples == 0 {
```
patrick-ogrady, line 141: only attempt a serial sample if time isn't massive (< 50ms otherwise savings will be inconsequential)?

parallel/src/policy.rs
```diff
@@ -0,0 +1,278 @@
...
+            Execution::Parallel => {
+                self.parallel_ns = update_ewma(self.parallel_ns, self.parallel_samples, elapsed_ns);
+                self.parallel_samples = self.parallel_samples.saturating_add(1);
+            }
+        }
+    }
+}
+
+fn update_ewma(current: u64, samples: u32, next: u64) -> u64 {
+    // Integer math keeps the update cheap and deterministic.
```
patrick-ogrady: remove this line

patrick-ogrady (commented):
final nits

coding/src/reed_solomon.rs
```diff
@@ -333,13 +333,14 @@ fn encode<H: Hasher, S: Strategy>(
     let (padded, shard_len) = prepare_data(data, k);
 
     // Compute recovery shards, striping large shard widths across the strategy
-    let recovery_buf = match striped::ranges(shard_len, strategy.parallelism_hint()) {
+    let manual = strategy.manual();
+    let recovery_buf = match striped::ranges(shard_len, manual.parallelism_hint()) {
```
patrick-ogrady, line 337: I figured that if you want access to parallelism hint, you want manual

coding/src/reed_solomon.rs
```diff
@@ -756,12 +757,11 @@ fn verify_root<H: Hasher, S: Strategy>(
...
         .collect::<Vec<_>>();
 
-    for (i, digest) in
-        strategy.map_init_collect_vec(missing_shards, H::new, |hasher, (i, shard)| {
-            hasher.update(shard);
-            (i, hasher.finalize())
-        })
-    {
+    let manual = strategy.manual();
+    for (i, digest) in manual.map_init_collect_vec(missing_shards, H::new, |hasher, (i, shard)| {
```
patrick-ogrady: this is manual because shards can be quite variable in size ... this would be a prime candidate for "weights"

parallel/src/lib.rs
```diff
@@ -67,27 +68,81 @@
...
             use alloc::vec::Vec;
         }
     }
+
+    /// A strategy wrapper for manually partitioned work.
+    ///
+    /// This disables adaptive serial-vs-parallel policy decisions for operations that callers have
+    /// already split into partitions.
+    #[derive(Clone, Debug)]
+    pub struct Manual<S> {
```
patrick-ogrady, line 101: Could do `override` or something else?

parallel/src/lib.rs
```diff
@@ -67,27 +68,81 @@
...
+        /// strategy, the job may run eagerly and inline on the calling thread when this method is
+        /// called (e.g. [`Sequential`] or a single-worker [`Rayon`] pool) or be deferred to a
+        /// worker thread and run while the returned future is awaited.
+        ///
+        /// If the job panics, the panic is propagated to the caller (at the call site for the
+        /// inline case, or at the future's await point otherwise); it never aborts the process.
+        fn spawn<F, T>(&self, f: F) -> impl core::future::Future<Output = T> + Send + 'static
+        where
+            F: FnOnce() -> T + Send + 'static,
+            T: Send + 'static;
```
clabby, line 141: This is a very useful function (prior art: https://github.com/commonwarexyz/monorepo/pull/3632 + https://github.com/commonwarexyz/monorepo/pull/3636 + https://github.com/commonwarexyz/monorepo/pull/4005).

I think I’d prefer the `F` type to be `impl FnOnce(Self) -> T + Send + ‘static` to avoid this shindig at the callsite:

```rs
let strategy = …;
strategy.spawn({
    let s = strategy.clone();
    move || { s.<op>(…) }
});

// instead
strategy.spawn(|s| s.<op>(...));
```

Implementors could internally clone `self`. This can be used in a ton of areas across the codebase where we want to yield to tokio while waiting on rayon, and most of the tasks we'll send off to rayon will want to use `Strategy`.

patrick-ogrady: This is meant more for one-off tasks that you want to use the pool, not necessarily orchestration of the pool as we'd previously discussed (will occupy a slot).

clabby: > will occupy a slot

I don't think that's true. `parallel` uses `thread_pool.install` for all of its operations, and in your spawn implementation, you're also using `thread_pool.spawn` (so everything runs in the same pool). Rayon supports nested parallelism, it's intelligent enough to use `ThreadPool::yield_now` under the hood and not park the thread that the closure passed to `spawn` runs on while the other `n-1` threads do work.

See https://github.com/rayon-rs/rayon/blob/main/FAQ.md

clabby: https://play.rust-lang.org/?version=stable&mode=release&edition=2024&gist=d094a9d9a98d7db2acabeaf6728ee08c <- showcases the above.

patrick-ogrady: I dug around a bit and the spawned closure would still occupy a worker while it runs but it would be safe as long as it uses internal rayon calls to wait (install/join/parallel iterators)...if the spawned closure blocks on a non-Rayon primitive, it can still occupy the worker.

patrick-ogrady:
bugbto run

patrick-ogrady:
bugbot run

parallel/src/lib.rs
patrick-ogrady: We can continue to explore this in a future PR (again, this change was just to unblock proper decoding in marshal).

parallel/src/policy.rs
```diff
@@ -0,0 +1,329 @@
...
+// Probe the non-preferred path on a longer interval in case conditions change.
+const RESAMPLE_INTERVAL: u32 = 100;
+// Track a short EWMA so recent measurements outweigh old startup noise.
+const EWMA_PREVIOUS_WEIGHT: u64 = 4;
+const EWMA_NEXT_WEIGHT: u64 = 1;
+const EWMA_WEIGHT: u64 = EWMA_PREVIOUS_WEIGHT + EWMA_NEXT_WEIGHT;
+// Serial must take at most 90% of parallel's time before it becomes preferred.
+const SERIAL_WIN_NUMERATOR: u64 = 90;
+const SERIAL_WIN_DENOMINATOR: u64 = 100;
+const SERIAL_SAMPLE_LIMIT_NS: u64 = 10_000_000;
```
patrick-ogrady: We should consider this as the multiplication of threads * duration (something with a ton of threads may go fast)?

parallel/src/lib.rs
clabby: > if the spawned closure blocks on a non-Rayon primitive, it can still occupy the worker.

Def - that'd be expected (and is the case already). Though when we want to not block a worker thread while executing a `Strategy` operation, nested parallelism using the strategy would not occupy the worker that the `spawn` closure is executed on.

Put up a PR to improve this API; Fairly small change that encourages this pattern to be used: https://github.com/commonwarexyz/monorepo/pull/4140 . Esp as you dive deeper into saturating the disk across threads, keeping worker threads unblocked will be pretty important.

p2p/src/utils/codec.rs
```diff
@@ -218,53 +218,42 @@ where
...
-                let handle = self
-                    .context
-                    .child("decode")
-                    .shared(true)
-                    .spawn(|_| async move {
-                        let result = V::decode_cfg(bytes.as_ref(), &config);
-                        (peer, result)
-                    });
+                let handle = self.strategy.spawn(move |_| {
+                    let result = V::decode_cfg(bytes.as_ref(), &config);
```
clabby, line 249: `strategy.spawn` lets a panic from `V::decode_cfg` unwind through `decode_pool.next_completed().await`, killing the whole `WrappedBackgroundReceiver` task instead of blocking only the offending peer. `commonware_utils::futures::Pool::next_completed` is just `FuturesUnordered::select_next_some()`; `select_loop!` expands to a plain loop/select without `catch_unwind`; and `Rayon::spawn` explicitly resumes worker panics at the await point. Existing invalid-message tests only cover `decode_cfg` returning `Err` — a panicking codec stops delivery for all peers on this receiver.

Wrap the decode call inside the spawned closure with `std::panic::catch_unwind(std::panic::AssertUnwindSafe(...))` and convert panic payloads into a `commonware_codec::Error` so `handle_decode_result` can block the peer and keep the receiver alive.

```patch
diff --git a/p2p/Cargo.toml b/p2p/Cargo.toml
index 5246c5e49..c7d449c85 100644
--- a/p2p/Cargo.toml
+++ b/p2p/Cargo.toml
@@ -36,6 +36,7 @@ thiserror.workspace = true
 tracing.workspace = true
 
 [dev-dependencies]
+bytes.workspace = true
 commonware-conformance.workspace = true
 commonware-math.workspace = true
 commonware-runtime = { workspace = true, features = ["test-utils"] }
diff --git a/p2p/src/utils/codec.rs b/p2p/src/utils/codec.rs
index 9ebd139f2..50c2359ac 100644
--- a/p2p/src/utils/codec.rs
+++ b/p2p/src/utils/codec.rs
@@ -284,7 +284,8 @@ mod tests {
         Manager as _, Recipients,
     };
     use commonware_actor::Feedback;
-    use commonware_codec::Encode;
+    use bytes::{Buf, BufMut};
+    use commonware_codec::{Encode, EncodeSize, Error as CodecError, Read, Write};
         ed25519::{PrivateKey, PublicKey},
         Signer,
@@ -391,6 +392,34 @@ mod tests {
         }
     }
 
+    #[derive(Debug)]
+    struct PanickingMessage(u8);
+
+    impl EncodeSize for PanickingMessage {
+        fn encode_size(&self) -> usize {
+            1
+        }
+    }
+
+    impl Write for PanickingMessage {
+        fn write(&self, buf: &mut impl BufMut) {
+            buf.put_u8(self.0);
+        }
+    }
+
+    impl Read for PanickingMessage {
+        type Cfg = ();
+
+        fn read_cfg(buf: &mut impl Buf, _cfg: &Self::Cfg) -> Result<Self, CodecError> {
+            let value = u8::read_cfg(buf, &())?;
+            if value == 0 {
+                panic!("decode panic");
+            }
+
+            Ok(Self(value))
+        }
+    }
+
     #[derive(Clone, Debug)]
     struct TestStrategy {
         parallelism: NonZeroUsize,
@@ -730,6 +759,55 @@ mod tests {
         });
     }
 
+    #[test_traced]
+    fn test_decode_panic_does_not_stop_receiver() {
+            let mut oracle = start_network(context.child("network"));
+
+            let pk1 = pk(0);
+            let pk2 = pk(1);
+            let pk3 = pk(2);
+            let control1 = oracle.control(pk1.clone());
+            let control2 = oracle.control(pk2.clone());
+            let control3 = oracle.control(pk3.clone());
+            track_peers(&oracle, 0, [pk1.clone(), pk2.clone(), pk3.clone()]);
+            link_bidirectional(&mut oracle, pk1.clone(), pk2.clone()).await;
+            link_bidirectional(&mut oracle, pk3.clone(), pk2.clone()).await;
+
+            let (mut sender1, _) = control1.register(0, TEST_QUOTA).await.unwrap();
+            let (_, receiver2) = control2.register(0, TEST_QUOTA).await.unwrap();
+            let (mut sender3, _) = control3.register(0, TEST_QUOTA).await.unwrap();
+
+            let (bg, mut rx) = WrappedBackgroundReceiver::<_, _, _, _, PanickingMessage, _>::new(
+                context.child("bg"),
+                receiver2,
+                (),
+                control2.clone(),
+                NZUsize!(16),
+                Sequential,
+            );
+            let _handle = bg.start();
+
+            let _ = sender1.send(Recipients::One(pk2.clone()), PanickingMessage(0).encode(), true);
+            let _ = sender3.send(Recipients::One(pk2.clone()), PanickingMessage(7).encode(), true);
+
+            let (from, value) = rx.recv().await.unwrap();
+            assert_eq!(from, pk3);
+            assert_eq!(value.0, 7);
+
+            loop {
+                let blocked = oracle.blocked().await.unwrap();
+                assert!(!blocked.contains(&(pk2.clone(), pk3.clone())));
+                if blocked.contains(&(pk2.clone(), pk1.clone())) {
+                    break;
+                }
+
+                context.sleep(Duration::from_millis(1)).await;
+            }
+        });
+    }
+
     #[test_traced]
     fn test_decoded_messages_drop_when_receiver_full() {
```
</details>

parallel/src/policy.rs
```diff
@@ -0,0 +1,404 @@
...
+//!
+//! State updates are serialized per policy entry, but calls do not hold the entry lock while work
+//! executes. Concurrent calls may therefore make decisions from an estimate that another in-flight
+//! call later updates, and measured samples are applied in completion order.
+
+use dashmap::DashMap;
+use std::{
+    panic::Location,
+    sync::Arc,
+    time::{Duration, Instant},
```
clabby, line 25: Given that these operations are usually latency-sensitive, I wonder if it's worth using https://docs.rs/quanta/latest/quanta/ ?

storage/src/qmdb/any/db.rs
```diff
@@ -280,7 +279,8 @@ where
                     .into_iter()
                     .flatten()
                     .collect()
-            };
+            },
+        );
```
clabby, line 283: very clean :smile:

p2p/src/utils/codec.rs
clabby: We resume the panic unwind in the worker thread (rather than within the rayon pool), though we didn't handle this before either. I'm OK if we want to keep as-is and assume that `V::decode_cfg` handles its errors properly.

clabby (dismissed):

p2p/src/utils/codec.rs
patrick-ogrady: I think our current behavior is preferred. We already propagate the panic.

parallel/src/policy.rs
patrick-ogrady: We can come back to this later if we determine its an issue.

p2p/src/utils/codec.rs
```diff
@@ -218,53 +218,42 @@ where
...
+                    (peer, result)
+                });
```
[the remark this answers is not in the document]
patrick-ogrady: This is the correct behavior. It is intentional.

PR #4131, opened 2026-06-29
[glue] DKG+reshare

### LINKED ISSUE #3921: [glue] Add standard DKG/reshare
## Overview

Now that we have the `commonware-glue` crate, it would be awesome to add a standard `Orchestrator` and `DKG` process for use with generic `Application` implementations. Consumers of commonware have duplicated and modified the DKG code in `examples/reshare` many times, and some of the logic is quite touchy.

The new `glue` component should be compatible with `glue::stateful`, though it should not require the use of `glue::stateful`. It should own:
1. Orchestrating epoch transitions and instantiating new instances of `simplex`.
2. Serving boundary certificates to peers.
3. Facilitating an initial DKG ceremony to bootstrap secret shares for a threshold scheme.
    - This should be its own thing; Really just need a binary that can run a dead-simple chain over simplex for one epoch. It can be assumed that all "players" that will be receiving a share are known at genesis.
4. Facilitating continuous reshare, supporting the addition and removal of participants.

All of this functionality, sans the initial DKG ceremony logic, should fit behind a standard interface that can be injected into `commonware_consensus::Application` implementations and orchestrated around the rest of the node's wiring.

clabby:
## Overview

> [!NOTE]
> PR Stack:
> - #4146
> - #4131 👈
> - `main`

Adds a standard, state-sync capable DKG+reshare library to `commonware-glue`. Protocol explained in the docs.

Explainer: https://claude.ai/artifacts/latest/bdf9c08d-cc7c-44ed-97ef-cd209e38d215

closes https://github.com/commonwarexyz/monorepo/issues/3921


consensus/src/marshal/ancestry.rs
```diff
@@ -56,6 +70,38 @@ impl<B: Block> Stream for BoundedAncestry<B> {
...
+impl<B: Block, S> Unpin for PrefixedAncestry<B, S> {}
+
+impl<B, S> Ancestry<B> for PrefixedAncestry<B, S>
+where
+    B: Block,
+    S: Stream<Item = B> + Send + Unpin + 'static,
+{
+    fn peek(&self) -> Option<&B> {
+        self.blocks.front()
+    }
```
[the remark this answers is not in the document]
clabby: This is expected.

glue/src/dkg/probe/actor/service.rs
```diff
@@ -0,0 +1,123 @@
...
+                return;
+            },
+            Some(message) = mailbox_message else {
+                mailbox_drained = true;
+                continue;
+            } => match message {
+                Message::Subscribe { response } => {
+                    if let Some(artifact) = &self.artifact {
+                        response.send_lossy(artifact.clone());
+                    }
```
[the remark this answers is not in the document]
clabby: Expected.

glue/src/dkg/reshare/application.rs
```diff
@@ -0,0 +1,208 @@
...
+        span.record("phase", field::debug(phase));
+        span.record("has_payload", tip_payload.is_some());
+
+        if self.final_block(height) {
+            if let Some(derived) = self.reshare.epoch_info(ancestry.clone()).await {
+                if Some(derived) != tip_payload {
+                    debug!("verification rejected: final block payload mismatch");
+                    return false;
+                }
+            }
```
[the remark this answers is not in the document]
clabby: Expected behavior. We only validate if we have it; For actors that participated in the ceremony, this is the case. For followers, they'll return `None`.

glue/src/dkg/anchor/actor/discovery.rs
```diff
@@ -0,0 +1,343 @@
...
+            self.pending = Some(pending);
+            return;
+        }
+
+        match self.artifact_from_response(&pending, response) {
+            Some(artifact) => self.resolve(artifact),
+            None => {
+                commonware_p2p::block!(self.blocker, peer, "invalid bootstrap boundary response");
+                self.pending = Some(pending);
+            }
```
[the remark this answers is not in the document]
clabby: I do not think this is reachable for an honest peer. Serving responds with the exact epoch requested, and discovery updates pending in the same synchronous handler that sends the request. pending only moves forward. The legitimate race is an older response arriving after supersession; we ignore that without blocking and cover it in late_response_for_superseded_boundary_does_not_block_peer. A response newer than the pending request is unsolicited, so blocking is intentional.

glue/src/dkg/probe/actor/discovery.rs
```diff
@@ -0,0 +1,725 @@
...
+        let mut deadline = self.context.current() + self.retry_timeout.get();
+
+        select_loop! {
+            self.context,
+            on_start => {
+                self.subscribers
+                    .retain(|subscriber| !subscriber.is_closed());
+                if marshal.is_some() && self.subscribers.is_empty() {
+                    break;
+                }
```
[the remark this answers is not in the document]
clabby: Intentional; Fits the usage pattern.

glue/src/dkg/orchestrator/actor.rs
```diff
@@ -0,0 +1,725 @@
...
+        let next = match next {
+            Ok(next) => next,
+            Err(EnterEpochError::GateClosed) => {
+                debug!(%next_epoch, "epoch gate closed before boundary transition");
+                return false;
+            }
+            Err(EnterEpochError::Stopped) => {
+                debug!(%next_epoch, "context shutdown while waiting to enter epoch");
+                return false;
+            }
```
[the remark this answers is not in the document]
clabby: Thanks — this is intentional fail-closed behavior rather than a dropped transition. `Gate::wait` checks whether the requested epoch was reached before returning `Closed`, so `GateClosed` means the next epoch was never published and there is no valid scheme with which to start a replacement engine. Marshal durably stores the finalized boundary before dispatch and advances processed height only after the exact acknowledgement. Leaving this delivery unacknowledged preserves the boundary for replay: on restart, `resolve_start` derives the active epoch from `processed.next()` and reloads the durable boundary. The `Stopped` branch is runtime shutdown, where starting another engine would also be incorrect. Dropping the old engine is required once its epoch's final block has finalized; acknowledging before `*active = next` would weaken the transition's atomicity.

glue/src/stateful/probe/actor/discovery.rs
```diff
@@ -207,16 +207,24 @@ where
...
     fn verify_finalization(
         &mut self,
         peer: P,
         finalization: Finalization<S, V::Commitment>,
     ) -> Option<(P, Finalization<S, V::Commitment>)> {
-        // Verify against the certificate scheme for the finalization's epoch. If no scheme is
+        let response_epoch = finalization.epoch();
+        let sample_scheme = self.provider.scheme(self.minimum_epoch)?;
+        if sample_scheme.participants().position(&peer).is_none() {
+            commonware_p2p::block!(self.blocker, peer, "finalization sent by non-participant");
```
patrick-ogrady, line 214: this can happen if the validator was added after the minimum epoch?

clabby: I don’t think so—we only send the request to participants in the configured minimum epoch. A validator added later wouldn’t receive one, so a response from it would be unsolicited. The response can still carry a newer-epoch finalization; we verify that certificate against the newer epoch’s scheme.

glue/src/stateful/mod.rs
```diff
@@ -116,6 +115,23 @@ pub struct Proposed<A: Application<E>, E: Rng + Spawner + Metrics + Clock> {
...
+///
+/// `parent` is the input [`Stateful`] received as a
+/// [`commonware_consensus::Application`] (from whatever wraps it);
+/// `provider` is the stateful-owned handle from [`Config::provider`]. Being
+/// generic over the parent input, it lets an outer application (for example a
+/// reshare wrapper) stack its own input on top of the stateful-owned provider
+/// without either layer knowing the other.
+pub struct Input<Parent, Provider> {
+    /// Input forwarded from the application wrapping [`Stateful`].
+    pub parent: Parent,
```
patrick-ogrady: "parent" is already overloaded? maybe we can do better?

glue/src/stateful/probe/mod.rs
```diff
@@ -42,9 +43,9 @@
 //!                                      sample reached, highest view becomes the floor: 13
 //! ```
 //!
-//! A peer that sends an undecodable or unverifiable first finalization in a request round is
-//! blocked. After a peer has already contributed a verified response for that round, later
-//! messages from that peer are ignored before validation.
+//! A non-participant or a peer that sends an undecodable or unverifiable first finalization in a
```
patrick-ogrady, line 46: Confirm we only block people we didn't request from (and not any message from them)

glue/src/stateful/probe/actor/discovery.rs
patrick-ogrady: Ah, missed the request first part.

glue/src/dkg/anchor/wire.rs
```diff
@@ -0,0 +1,192 @@
...
+}
+
+/// Boundary response decoded from a peer.
+pub(crate) struct Response<S, V>
+where
+    S: Scheme<V::Commitment>,
+    V: Variant,
+{
+    pub(crate) finalization: Finalization<S, V::Commitment>,
+    pub(crate) block: V::Block,
```
patrick-ogrady: I suppose we could get a TON of instances of pretty big blocks (as opposed to using some sort of resolver once we get the finalization we want)? I think this is fine for now but something we may want to come back to...

glue/src/dkg/reshare/application.rs
```diff
@@ -0,0 +1,869 @@
...
+            debug!("proposal rejected: missing parent ancestry");
+            return None;
+        };
+        let height = parent.height().next();
+        let phase = self.phase(height);
+        let span = tracing::Span::current();
+        span.record("height", height.traced());
+        span.record("phase", field::debug(phase));
+
+        let (payload, log_reservation) = if self.final_block(height) {
```
patrick-ogrady, line 150: IIRC we can kickoff async verify of material here...do we do it in this PR?

patrick-ogrady (commented):
non-glue code looks good...spot-checked remaining

glue/src/stateful/probe/mod.rs
clabby: Yep. We only block a non-participant after decoding their message as a response. A request from them is ignored. Malformed messages and invalid finalizations are blocked separately.

glue/src/dkg/reshare/application.rs
clabby: Not currently. Deferred marshal runs received block verification in the background, but we still derive and verify the epoch material inline when we request `epoch_info`. Tracking kicking that work off earlier in #4300.

glue/src/stateful/mod.rs
clabby: Implemented in https://github.com/commonwarexyz/monorepo/commit/9aa8228ed0a8bbeb4a8b0ec3b4750a268de6bb84.

glue/src/dkg/anchor/wire.rs
clabby: Implemented in https://github.com/commonwarexyz/monorepo/commit/7f7e0cfbb2ee66fd93a7db299953842f166e2814.

glue/src/dkg/orchestrator/actor.rs
```diff
@@ -0,0 +1,743 @@
...
+        };
+
+        *active = next;
+        acknowledgement.acknowledge();
```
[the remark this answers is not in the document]
clabby: Known and expected.

glue/src/dkg/anchor/actor/discovery.rs
```diff
@@ -0,0 +1,872 @@
...
+    height: Height,
+    epoch: Epoch,
+    in_flight: Option<Candidate<S, V>>,
+    candidates: VecDeque<Candidate<S, V>>,
+}
+
+/// The discovery phase of the anchor actor.
+///
+/// Waits for subscribers, listens to the Simplex certificate channel, and uses
+/// the first verifiable target finalization to discover the previous epoch's
```
patrick-ogrady: Is there any concern that this may be many epochs behind (we just take the first)?

glue/src/dkg/anchor/actor/mod.rs
```diff
@@ -0,0 +1,164 @@
...
+use commonware_utils::{NonZeroDuration, channel::mpsc, ordered::Set};
+use discovery::Discovery;
+use rand_core::CryptoRng;
+use std::num::{NonZeroU64, NonZeroUsize};
+
+mod discovery;
+mod serving;
+
+/// Peer-set slot used before epoch-scoped tracking can supersede bootstrap peers.
+const BOOTSTRAP_PEER_SET_INDEX: u64 = 0;
```
patrick-ogrady: I think we can just accept the actual index with the initial bootstrap peers?

glue/src/dkg/bootstrap/mod.rs
```diff
@@ -0,0 +1,629 @@
...
+            payload: input.payload,
+        })
+    }
+
+    async fn verify(
+        &mut self,
+        _: (E, Self::Context),
+        _ancestry: impl Ancestry<Self::Block>,
+    ) -> bool {
+        true
```
patrick-ogrady, line 590: add a comment of what validation done elsewhere?

glue/src/dkg/orchestrator/actor.rs
```diff
@@ -0,0 +1,806 @@
...
+    ) -> Option<ResolvedStart<P::Scheme, MV::Commitment, DV, <P::Scheme as Verifier>::PublicKey>>
+    {
+        let height = epoch
+            .previous()
+            .and_then(|epoch| epocher.last(epoch))
+            .unwrap_or_else(Height::zero);
+        let Some(boundary) = self.marshal.get_block(height).await else {
+            // Marshal cancels pending reads when it stops; only an
+            // intact-but-absent boundary block is a retention violation.
+            if self.context.stopped().now_or_never().is_some() {
```
patrick-ogrady: we don't use this pattern elsewhere?

glue/src/dkg/orchestrator/actor.rs
```diff
@@ -0,0 +1,806 @@
...
+                debug!(%height, "boundary block read canceled during shutdown");
+                return None;
+            }
+            panic!("missing finalized boundary block at height {height}");
```
patrick-ogrady: assuming this can't happen due to normal pruning?

glue/src/dkg/orchestrator/actor.rs
```diff
@@ -0,0 +1,806 @@
...
+use std::{
+    marker::PhantomData,
+    num::{NonZeroU16, NonZeroU64, NonZeroUsize},
+    sync::Arc,
+    time::Duration,
+};
+use tracing::{debug, info};
+
+/// Offset leaves peer-set ID zero available for anchor bootstrap peers.
+const EPOCH_PEER_SET_OFFSET: u64 = 1;
```
patrick-ogrady: if we fix the original offset thing, we should be able to remove this?

glue/src/stateful/actor/core/processing.rs
```diff
@@ -102,16 +102,21 @@ where
                     span,
                     context,
                     ancestry,
+                    parent,
```
patrick-ogrady: upstream?

glue/src/dkg/reshare/mod.rs
```diff
@@ -0,0 +1,193 @@
...
+
+/// The max supported [`ModeVersion`] of this reshare protocol implementation.
+#[cfg(any(
+    commonware_stability_BETA,
+    commonware_stability_GAMMA,
+    commonware_stability_DELTA,
+    commonware_stability_EPSILON,
+    commonware_stability_RESERVED
+))]
+pub const MAX_SUPPORTED_MODE: ModeVersion = ModeVersion::v0();
```
patrick-ogrady: why duplicate?

glue/src/dkg/reshare/store.rs
```diff
@@ -0,0 +1,1096 @@
...
+        Ok(match u.int_in_range(0..=2)? {
+            0 => Self::Dealing(u.arbitrary()?, u.arbitrary()?),
+            1 => Self::Ack(u.arbitrary()?, u.arbitrary()?),
+            _ => Self::Log(u.arbitrary()?, u.arbitrary()?),
+        })
+    }
+}
+
+#[cfg(all(test, feature = "arbitrary"))]
+mod conformance {
```
patrick-ogrady: move to bottom?

glue/src/dkg/reshare/application.rs
```diff
@@ -0,0 +1,869 @@
...
+use tracing::{debug, field};
+
+/// Per-proposal input handed to an application wrapped by [`Application`].
+///
+/// Carries the wrapper's parent input alongside the reshare `payload` selected
+/// and fetched for the block being proposed. The wrapped application attaches
+/// `payload` to the block it builds and uses `parent` for its own purposes.
+pub struct Input<Parent, V: Variant, C: Signer> {
+    /// Input forwarded from the application wrapping the reshare wrapper.
+    pub parent: Parent,
```
patrick-ogrady: upstream?

patrick-ogrady (commented):
final nits

glue/src/dkg/anchor/actor/discovery.rs
clabby: Yes, there is a concern. See the weak subjectivity section in the docs. Since we don't know the active participant set at this point, we cannot guarantee that the finalization is _recent_, only that it's _valid_.

patrick-ogrady: Could we not just use the f+1 wait here instead of doing it twice?

glue/src/dkg/orchestrator/actor.rs
clabby: The pruning requirements are documented. The user could misconfigure, but if they keep around the right range of blocks, it can't.

patrick-ogrady: For constantinople, I suspect we'd need to then keep at least epoch + 1 blocks or something?

glue/src/dkg/reshare/mod.rs
clabby: `v1` is gated behind `ALPHA`. The max supported mode is either `v0` or `v1` depending on the stability setting.

glue/src/dkg/orchestrator/actor.rs
clabby: should unconditionally gracefully shutdown.

glue/src/dkg/anchor/actor/discovery.rs
clabby: We don't know `f`.

patrick-ogrady: We know `f` from the original peer configuration we had at startup for whatever epoch we started in?

glue/src/dkg/orchestrator/actor.rs
clabby: https://github.com/commonwarexyz/monorepo/blob/70f004e44d8511d812dd2c4a724b580731cd3934/glue/src/dkg/mod.rs#L116-L125

glue/src/dkg/anchor/actor/discovery.rs
clabby: That's not `f` for the active epoch. The initial peer configuration is just a sample of _potentially_ connected peers (we do assume that at least one can serve you data). The active committee could have (and likely will have) partially rotated since the node's release.

glue/src/dkg/anchor/actor/mod.rs
clabby: I don't think we should. The bootstrap peers are unique, and might not be a canonical committee (just some nodes that the application pins and expects to be persistent in the committee until the next release).

glue/src/dkg/orchestrator/actor.rs
clabby: https://github.com/commonwarexyz/monorepo/pull/4131#discussion_r3623764384

glue/src/dkg/probe/actor/discovery.rs
```diff
@@ -0,0 +1,919 @@
...
```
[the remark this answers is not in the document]
clabby: Expected.

patrick-ogrady (dismissed):
the last big missing piece imho

glue/src/dkg/bootstrap/mod.rs
```diff
@@ -0,0 +1,668 @@
...
+    }
+}
+
+impl Drop for BootstrapActors {
+    fn drop(&mut self) {
+        self.buffer_handle.abort();
+        self.reshare_handle.abort();
+        self.marshal_handle.abort();
+        self.simplex_handle.abort();
+    }
```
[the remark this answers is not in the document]
clabby: That'd be a critical failure; This is fine.

patrick-ogrady (dismissed):

patrick-ogrady (approved):

glue/src/dkg/orchestrator/actor.rs
```diff
@@ -0,0 +1,742 @@
...
+                }
+            },
+        };
+        drop(shutdown);
+
+        let _ = self.manager.track(epoch.get(), peers);
+        let scheme = self
+            .provider
+            .scheme(epoch)
+            .unwrap_or_else(|| panic!("missing consensus scheme for epoch {epoch}"));
```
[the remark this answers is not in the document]
clabby: This is expected: the scheme for the initial epoch is pre-configured before the orchestrator starts, so the gate should resolve immediately here.

PR #4135, opened 2026-06-29
[examples/reshare] Nits

clabby:
## Overview

Fixes a few nits in `commonware-reshare`:
- We don't penalize a player for sending an `Ack` back with an invalid signature due to `Dealer::receive_player_ack` not bubbling up a verification error. This doesn't hurt security (same impact as that player never acking, they count as one of the `f`), though we also don't attribute the fault to the peer and block them.
    - We also don't penalize a peer for sending us a malformed message or routing a message to the wrong peer atm.
- `SignedDealerLogs` included in blocks in the second phase of the epoch are absorbed into storage without a dealer set membership check, just a signature verification. This, again, doesn't have an impact on security since we filter on known dealers, though allows for byzantine proposers to arbitrarily bloat storage (bounded by # of blocks in the epoch.)

### DIFF
--- a/cryptography/src/bls12381/dkg/feldman_desmedt.rs
+++ b/cryptography/src/bls12381/dkg/feldman_desmedt.rs
@@ -575,52 +575,36 @@ impl<V: Variant, P: PublicKey> Info<V, P> {
             .expect("player index should be < num_players"))
     }
 
-    /// Verify a dealer's public message.
-    ///
-    /// A wrong commitment degree, missing prior share commitment, or reshare
-    /// commitment whose constant term does not match the dealer's prior share
-    /// commitment, is a provable [`Verdict::Fault`].
-    fn check_dealer_pub_msg<M: Faults>(
-        &self,
-        dealer: &P,
-        pub_msg: &DealerPubMsg<V>,
-    ) -> Verdict<()> {
+    #[must_use]
+    fn check_dealer_pub_msg<M: Faults>(&self, dealer: &P, pub_msg: &DealerPubMsg<V>) -> bool {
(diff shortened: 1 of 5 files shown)


cryptography/src/bls12381/dkg/feldman_desmedt.rs
```diff
@@ -1760,28 +1822,38 @@ impl<V: Variant, S: Signer> Player<V, S> {
...
+    ) -> Verdict<PlayerAck<S::PublicKey>> {
         if self.view.contains_key(&dealer) {
-            return None;
+            return Verdict::Skip;
         }
-        self.info.dealer_index(&dealer).ok()?;
-        if !self.info.check_dealer_pub_msg::<M>(&dealer, &pub_msg) {
-            return None;
+        if self.info.dealer_index(&dealer).is_err() {
+            return Verdict::Fault;
```
cronokirby, line 1800: Is this a fault or a skip? I feel like it's really context dependent on how this should be handled

clabby: What context would this be a `Skip`? My thinking here was that if a caller supplies a key that isn't a dealer, the message is not useful (and the sender of the message should know that.)

cronokirby (approved):

cryptography/src/bls12381/dkg/feldman_desmedt.rs
cronokirby: It depends on the attribution; right now I think the way this works is that the channel identifies the dealer, and then you pass in the dealer and their message, so passing in the wrong dealer would be the result of a bug in the code that the receiver of the message is running.

Issue #3118, opened 2026-02-14
[Storage] migrate all Persistable implementors to non-mut sync/commit


roberto-bayardo:
In order to support greater concurrency (related: https://github.com/commonwarexyz/monorepo/issues/2857), commit/sync implementations shouldn't block reads. This requires non-&mut signatures.  Until all implementations adopt non-mut sync, the `Persistable` trait which defines commit/sync/destroy can't be changed.

roberto-bayardo:
Re-opening because reads are still unnecessarily blocked during QMDB commit, which has an underlying fsync in the self.log.commit() call.

PR #4138, opened 2026-06-29
[glue/stateful] Split DatabaseSet::finalize so the commit fsync runs under a shared lock

### LINKED ISSUE #4137: [glue] DatabaseSet::finalize holds the write lock across the commit fsync, serializing readers
### Summary

The `DatabaseSet` blanket impl for `Arc<AsyncRwLock<T>>` (`glue/src/stateful/db/mod.rs`) holds the exclusive write lock across the *entire* `finalize` — both `apply_batch` (the in-memory index/bitmap mutation) and the durability fsync:

```rust
async fn finalize(&self, batches: Self::Merkleized) {
    let mut database = self.write().await;
    finalize_or_panic(&mut *database, batches, None).await; // apply_batch + sync, both under write()
}
```

`finalize_or_panic` calls `ManagedDb::finalize`, whose concrete impls are `self.apply_batch(..).await?; self.sync().await`. Any task reading committed state through the same `AsyncRwLock` (e.g. proof/query reads served from a database while it applies blocks) therefore blocks for the whole finalize, including the fsync — the long, variable pole.

### Proposal

Hold the exclusive lock only for `apply_batch`, then run the fsync under a **shared** lock:

```rust
{ let mut db = self.write().await; db.apply_batch(b).await... }  // exclusive, brief
{ let db = self.read().await;      db.commit().await... }        // shared — fsync off readers
```

Mechanically: add `apply_batch(&mut self)` + `commit(&self)` to `ManagedDb` (the two primitives the concrete impls already call inherently), make `finalize` a defaulted combinator over them, and split the blanket + tuple `DatabaseSet::finalize` into the two lock scopes above. `ManagedDb::finalize` stays as a convenience default.

### Soundness

`apply_batch` publishes the new logical state in memory; `sync` is durability-only. A reader entering the apply→commit gap observes the committed-in-memory state, and crash semantics are unchanged (`apply_batch` was never durable before `sync`). This assumes a single writer per database (the stateful actor serializes finalizes); concurrent writers are out of scope.

### Dependency

Requires storage `commit`/`sync` to take `&self` — i.e. #3118. They were `&self` in `v2026.5.0` but are currently `&mut self` on `main` (e.g. `qmdb::any::Db::sync(&mut self)`), so this is gated on #3118. With `&self` storage the glue change is small and additive.

### Out of scope

Owned snapshots / COW / MVCC / snapshot isolation. `apply_batch` stays briefly exclusive.

christopherwxyz:
Closes #4137.

> [!WARNING]
> **Draft — depends on #3118.** Against current `main` this does not compile: `commit(&self)` calls `db.sync()`, but storage `sync`/`commit` are `&mut self` on `main` (they were `&self` in `v2026.5.0`). The only build errors are the 14 `E0596` "cannot borrow `*self` as mutable" at the `commit` impls — i.e. exactly the #3118 surface. Ready for review once storage `commit`/`sync` are non-`&mut`.

### What

`DatabaseSet::finalize` (the `Arc<AsyncRwLock<T>>` blanket impl and the multi-DB tuple impl) holds the exclusive write lock across the *entire* finalize — both `apply_batch` and the durability fsync — so any read served through the same lock blocks for the whole flush. This splits the lock:

```rust
{ let mut db = self.write().await; db.apply_batch(b).await... }  // exclusive, brief
{ let db = self.read().await;      db.commit().await... }        // shared — fsync off readers
```

### How

- `ManagedDb` gains two primitives — `apply_batch(&mut self)` and `commit(&self)` — which the concrete impls already call inherently (`apply_batch` then `sync`). `finalize` becomes a defaulted combinator over them; existing direct callers (tests) keep working.
- The blanket and tuple `DatabaseSet::finalize` impls run `apply_batch` under the write lock and `commit` under a read lock. The `*_or_panic` helper splits to match (with the same `#[tracing::instrument]` treatment as the existing ones).

### Soundness

`apply_batch` publishes the new logical state in memory; `commit` is durability-only. With a single writer per database (the stateful actor serializes finalizes), a reader entering the apply→commit gap observes the committed-in-memory state, and crash semantics are unchanged (`apply_batch` was never durable before `sync`). Concurrent writers are out of scope.

Adds `finalize_runs_commit_under_shared_lock_so_readers_proceed`: a reader acquires the shared lock while a blocked `commit`/fsync is in flight (would deadlock under the old exclusive-across-commit finalize).

### Out of scope

Owned snapshots / COW / MVCC / snapshot isolation. `apply_batch` stays briefly exclusive.

### Limitations / notes

- **Benefit is partial under sustained finalization.** `TracedAsyncRwLock` (tokio `RwLock`) is write-preferring: once the *next* block's `apply_batch` is waiting on `write()`, newly-arriving readers queue behind it rather than joining the in-flight `commit`'s shared lock. Reads already holding or queued ahead of that writer proceed during the fsync; later arrivals still wait. The window is the gap between a `commit` starting and the next `apply` requesting the lock.
- **Coverage.** `finalize_runs_commit_under_shared_lock_so_readers_proceed` exercises the single-DB blanket impl; the structurally-identical split inside `impl_database_set!` (the multi-DB path) is not separately covered for shared-lock-during-commit.
- **Single-writer assumption.** Soundness relies on finalizes being serialized (the stateful actor) and `rewind_to_target` being boot-only; no other `write()` taker may race into the apply→commit gap.

### DIFF
--- a/glue/src/stateful/db/any.rs
+++ b/glue/src/stateful/db/any.rs
@@ -379,8 +379,11 @@ where
             && *target.range.end() == Location::<F>::new(batch.bounds().total_size)
     }
 
-    async fn finalize(&mut self, batch: Self::Merkleized) -> Result<(), Error<F>> {
-        self.apply_batch(batch.inner).await?;
+    async fn apply_batch(&mut self, batch: Self::Merkleized) -> Result<(), Error<F>> {
+        self.apply_batch(batch.inner).await.map(|_| ())
+    }
+
+    async fn commit(&self) -> Result<(), Error<F>> {
         self.sync().await
     }
 
@@ -476,8 +479,11 @@ where
             && *target.range.end() == Location::<F>::new(batch.bounds().total_size)
     }
 
-    async fn finalize(&mut self, batch: Self::Merkleized) -> Result<(), Error<F>> {
-        self.apply_batch(batch.inner).await?;
+    async fn apply_batch(&mut self, batch: Self::Merkleized) -> Result<(), Error<F>> {
+        self.apply_batch(batch.inner).await.map(|_| ())
+    }
+
+    async fn commit(&self) -> Result<(), Error<F>> {
         self.sync().await
     }
 
--- a/glue/src/stateful/db/current.rs
+++ b/glue/src/stateful/db/current.rs
@@ -377,8 +377,11 @@ where
             && *target.range.end() == Location::<F>::new(batch.bounds().total_size)
     }
 
-    async fn finalize(&mut self, batch: Self::Merkleized) -> Result<(), Error<F>> {
-        self.apply_batch(batch.inner).await?;
+    async fn apply_batch(&mut self, batch: Self::Merkleized) -> Result<(), Error<F>> {
+        self.apply_batch(batch.inner).await.map(|_| ())
+    }
+
+    async fn commit(&self) -> Result<(), Error<F>> {
         self.sync().await
     }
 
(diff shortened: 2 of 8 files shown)


patrick-ogrady:
This commit will be changed shortly to not block on fsync (which should alleviate this issue).

PR #4144, opened 2026-06-30
[storage/qmdb] Remove Batch Cache + Add `Staged` Updates

patrick-ogrady:
## Summary

- Replaces the implicit read-resolution cache with an explicit staged path: `stage` returns loaded values plus a `Staged` handle, `expand` appends more read keys to the same index space, and `Staged::merkleize(updates, upserts, metadata, db).await` merkleizes the selected updates. Normal `get_many` stays read-only, and the staged API is threaded through the `any`, `current`, and glue surfaces.
- Parallelizes large journal reads: a synchronous page-cache probe (`Contiguous::try_read_many_sync`, backed by `try_read_many_sync_into` on the paged buffer) returns a `Probed` token that carries misses to a single batched completion (`fetch`/`fetch_missing`). `authenticated::Journal::read_many` shards the probe across the strategy pool, and `get_many_map` fuses probe/sort/cache-read/match into one pass, run on the calling thread or sharded across the pool as the strategy policy prefers for the batch size (no fixed thresholds). Every candidate position is still decoded by exactly one pass, and non-increasing position lists still error deterministically, so error behavior does not depend on cache state.
- Serves warm page lookups through a direct-mapped hint array instead of a hash probe chain per item: each cached page records its `Clock` slot in a fixed array indexed by a salted page number (consecutive pages stay in consecutive entries, so sorted batches walk consecutive lines), and reads validate the hinted slot against the full `(blob id, page number)` key via the new `Clock::get_at` before falling back to the hash lookup. Hints are best-effort: eviction, invalidation, and cross-blob collisions all fail validation safely (a `live` flag on `Clock` slots makes freed slots unreadable), they need no maintenance, and the array's memory is fixed at construction (2x capacity), so no blob offset can grow it.
- Trims the read plane's steady-state overheads: `get_many_map` defers merging the shard probes to the miss path (completion never reads the merged hit slots, so a fully cached batch skips the concatenation), `probe_items` serves from a per-thread scratch buffer instead of allocating and zeroing one per probe, and the paged cache's batched read resolves every range's first page before copying so the independent lookups overlap memory latency instead of serializing behind each copy. The `constantinople` harness gains a `page_cache` argument for cache-size sweeps.
- Adds regression coverage: staged bulk updates match explicit writes (ordered and unordered, ancestor commits, overlapping upserts), dedicated tests exercise the sync cache read, the sharded read path, and the fused path with parallelism > 1 (no pre-existing test ran large batches on a multi-threaded strategy), an at-scale `resolve_updates` test pins duplicate-key collapse, cross-arm precedence, and upsert overlap above the sort's insertion-sort threshold, and hint staleness is pinned for every mode (eviction reuse, cross-blob collision, invalidation, sparse page numbers). The variable journal now records `cache_hits`/`cache_misses` like the fixed journal, with tests asserting exact hit/miss splits through the probe plane.

## Why

Constantinople's hot path loads many keys, computes new values for a subset, and writes that subset. The old flow resolved committed locations during the load, then paid again at merkleize via index probes and journal reads; the staged handle carries the resolutions from load into merkleization without exposing a reusable cache that can be paired with the wrong batch. Profiling the result showed the remaining time was not I/O: the autotuner was parking the thread pool on latency-critical fan-outs, and large reads spent their time in a serial page-cache copy loop (~45ns/item) between parallel phases.

## Before / after

Staged reads: the old flow warmed a hidden cache on the batch that merkleize joined by key; the new flow returns a typed handle whose updates reference reads by index, so resolutions cannot be paired with the wrong batch.

```text
Before: reads warm a hidden cache; merkleize joins it by key

  batch.get_many(keys, db) ------------------------------> values
        |
        '-- side effect --> resolved: Mutex<key -> (loc, cached)>
                                             |
  batch.write(key, value), per changed key   |  implicit join by key
        |                                    v
  batch.merkleize(db, metadata) --> mutations found in `resolved`
                                    skip the index probe + journal re-read

After: reads return a typed handle; updates reference reads by index

  (values, staged) = batch.stage(keys, db)           read indices 0..n
  (range, more, staged) = staged.expand(keys2, db)   read indices n..m
        |
        v  caller computes new values for the changed subset
  staged.merkleize(updates, upserts, metadata, db)
        updates: (read_index, Some(v) | None) -> reuse staged resolution
        upserts: (key, Some(v) | None)        -> normal write path
```

Autotuner: the old cost model charged parallel runs for the whole pool, so latency-critical fan-outs ran serial; the new model compares wall time and lets pool-contention feedback (queueing inflates parallel samples) handle shared-pool pressure.

```text
A fan-out that takes 12 ms serially and 3 ms across 8 workers:

                    wall time ->
  serial:    [############]        12 ms on one worker
  parallel:  [###]                  3 ms on 8 workers = 24 ms worker time

  Before: serial wins if serial < 0.9 x parallel x threads
          12 < 0.9 x 24 = 21.6      -> serial (pool parked, 4x the latency)
  After:  serial wins if serial <= parallel wall time (ties to serial)
          12 > 3                    -> parallel
```

Large batched reads: the old path did a parallel index probe, then walked the page cache one item at a time; the new path fuses probe/sort/cache-read/match per shard and batches everything, leaving one batched journal read for the misses.

```text
Before: parallel probe, then one serial page-cache walk

  keys -> probe index (sharded) -> sort + dedup -> read_many:
             copy each item from the page cache one at a time
             (~45 ns x 131k items, serial), misses read the blob
          -> match ops back to keys

After: probe/sort/cache-read/match fused per shard; one batched miss read

  keys ---+-> [shard 0] probe -> sort -> batched cache read -> match
          +-> [shard 1] probe -> sort -> batched cache read -> match
          +-> [shard N] probe -> sort -> batched cache read -> match
                  |
                  |  (strategy policy picks calling-thread vs pool per
                  |   batch size; one cache-lock acquisition per blob)
                  v
          cache misses from all shards, position-sorted
          -> one batched journal read -> match remaining ops to keys
```

## Benchmarks

1M keys, 32,768 updates, 8 threads, 20 iterations, separate `CARGO_TARGET_DIR`s, baseline (`21b1ae131`: staged API included, before the autotuner fix and read-path work) built from a pinned worktree. Re-measured 2026-07-08 (final API shape: hint array included, no-cache-pass completion variant removed), p50 from one interleaved run per shape. Roots matched on every iteration of every shape.

| variant | depth | reads | read_chunks | before p50 | after p50 | change |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| `any::unordered::fixed::mmb` | 0 | 131,072 | 4 | 29.47 ms | 11.99 ms | -59% |
| `current::unordered::fixed::mmb` | 0 | 131,072 | 4 | 29.42 ms | 15.46 ms | -47% |
| `any::unordered::fixed::mmb` | 0 | 32,768 | 1 | 15.50 ms | 7.26 ms | -53% |
| `any::unordered::fixed::mmb` | 0 | 131,072 | 1 | 24.42 ms | 11.32 ms | -54% |
| `any::unordered::variable::mmb` | 0 | 131,072 | 4 | 48.58 ms | 12.77 ms | -74% |
| `any::unordered::fixed::mmb` | 1 | 131,072 | 4 | 33.85 ms | 16.15 ms | -52% |
| `current::unordered::fixed::mmb` | 1 | 131,072 | 4 | 37.57 ms | 19.72 ms | -48% |
| `any::unordered::fixed::mmb` | 1 | 32,768 | 1 | 18.06 ms | 10.25 ms | -43% |
| `any::unordered::fixed::mmb` | 1 | 131,072 | 1 | 26.43 ms | 14.90 ms | -44% |
| `any::unordered::variable::mmb` | 1 | 131,072 | 4 | 55.09 ms | 16.89 ms | -69% |

Thread scaling on this machine (6 Super + 12 Performance cores): 12 threads is a marginal win for `any` (p50 -2.7%) and a wash for `current`; 16 regresses both, since the remaining load time is memory-latency-bound and the pre-cut shards wait on the slowest core tier. Roots are identical across thread counts.

Cache-hit robustness (new `page_cache` bench argument): sweeping the page cache from 512MiB (working set fully cached) down to 4MiB (~4%) shows the sync-probe plane's win degrading monotonically into an exact tie with a plain batched async read, never inverting. Misses flow through the identical batched fallback either way, so the probe has no losing regime; at 32MiB (~30% hit rate) and below the two are indistinguishable.

<details>
<summary>Bench usage, historical numbers, and things tried</summary>

The `constantinople` bench positional order is:

```bash
cargo bench -p commonware-storage --bench constantinople -- <db> [depth] [iters] [keys] [reads] [read_chunks] [updates] [threads] [page_cache]
```

`read_chunks=1` does one initial `stage`; values greater than 1 split the reads into the initial `stage` plus subsequent `expand` calls. `page_cache` is the cache capacity in 4096-byte pages (default 131,072 = 512MiB); shrink it to measure miss-heavy regimes.

Historical (measured before the #4130 merge, which is why its baselines are faster than the `21b1ae131` column above): staged-API comparison against the old `get_many + write` path. The baseline was detached `HEAD` (`e00daf8fc`) with a benchmark-only `reads` argument added while preserving the old path.

| reads | read_chunks | updates | baseline p50 | PR p50 | p50 change |
| ---: | ---: | ---: | ---: | ---: | ---: |
| 32,768 | 1 | 32,768 | 19.23 ms | 14.23 ms | -26.0% |
| 131,072 | 1 | 32,768 | 33.51 ms | 23.93 ms | -28.6% |

Previous main-table measurements (same shapes): -41% to -68% before the read-plane tuning pass, -38% to -68% with it (2026-07-05, pre-hint-array). Each refresh re-measures both columns in one session, so baseline values differ with machine conditions.

Historical: optimization-pass comparison (`700fc8058` vs `8130e7e59`), also pre-#4130: `any` 20.97 -> 20.36 ms (-2.9%), `current` 28.75 -> 24.62 ms (-14.4%).

Things tried:

- Hashing-only changes were not enough; earlier runs were noisy with no stable end-to-end win.
- A fused read-and-update helper and a public `UpdateCache` handoff were dropped for the typed `Staged` handle, which cannot be paired with the wrong batch.
- The staged API uses `(read_index, value)` pairs so callers can read many keys and update only the changed subset; upserts are applied last.
- An owned-key staged API (avoiding ~32k key clones) benched neutral; removed.
- Sharding `read_many` with per-item `try_read_sync` calls was ~2x worse than the serial batched read (per-item reader, buffer, and lock costs under contention); the win required batching per shard.
- Pinning the merkleize sorts and hash fan-outs to `manual()` recovered the autotuner regression, but fixing the policy cost model made the pins unnecessary and benched slightly better.
- Page-memoization in `read_cached_many` and merge-walks replacing binary searches over superseded floor-raise locations were noise-level; dropped.
- A committed-only fast path in `stage_reads` (skipping overlay resolution for fresh batches rooted at the committed DB) benched -3-5% p50 at 131k depth-0 reads but was dropped as overfit to the bench shape: it only fires when a batch has no writes and no pending ancestors, and it duplicated the general read path.
- Replacing the fused `get_many_map` with a plain probe/sort/`read_many`/match pipeline (letting the journal's sync cache pass do all the work) benched +22-44% p50 at 131k reads; the per-shard fusion of cache reads with key matching is load-bearing.
- Both concrete journals implement the sync probe: fixed via batched page-cache reads; variable by resolving frame extents from the offsets journal (`off(p+1) - off(p)`) and serving them with batched variable-length range reads, one cache-lock acquisition per blob group. Per-frame sync reads under a sharded fan-out were 2-3x slower than the batched form (two lock acquisitions per frame collapse under thread contention).
- A direct-indexed per-blob page table plus a software-prefetch pipeline in `read_cached_many` benched a further -19% on the warm load phase (extraction 56 -> 33 ns/item serial, -41% parallel CPU) and was initially parked over its cross-structure staleness invariant and per-blob growth cap. The table half later landed as the direct-mapped hint array above once two findings removed those costs: `Clock::get_at` with a slot `live` flag is fully self-validating (no staleness invariant, no invalidation bookkeeping), and a fixed-size salted-linear array needs no cap (hash-mixed hint indexing benched +0.2 ms/block by randomizing hint-line access; linear indexing ties the dense table). The prefetch-pipeline half stays parked: per-arch asm for a marginal further gain, and only with the data prefetch issued from the copy pass (prefetching at resolve time is useless at batch sizes, the lines are evicted again before the copy arrives).
- Splitting the CLOCK cache's reference bits into a dense side array (64 per cache line, for cheaper eviction sweeps) regressed inserts 13-27% and a read-heavy small-cache mix 8%: at small capacities everything is L1-resident already and the split adds a second line touch per operation.
- Parallelizing merkleize's staged-update sort and op/diff build benched flat end-to-end in bench processes (the adaptive policy re-converges per process; pinned-parallel ceiling was -0.24 ms merkleize) and was reverted as not worth the complexity until the policy resample interval is revisited.
- Cold-regime attribution (4MiB cache, ~134k faults/iteration): the fault path's two lock holds plus the 4KB insert copy total ~2% of fault thread-time, so copy-out-of-lock and cache sharding are not levers there; the time is in the per-page fetch pipeline (pread syscalls, per-page shared-future allocations, CRC validation), where the lever would be batching adjacent miss fetches.
- A no-cache-pass completion variant (`read_direct_into`, "skip the probe I already ran") was removed: with the hint array a redundant probe over misses costs ~1% even fully cold and nothing warm, so probe completion is now the plain batched read and the cache bypass is off the public surface. Bounding the batched read's lock holds (chunked re-acquisition) benched no better and was dropped with it.
- Replacing the post-floor-raise diff re-sort with sort-`floor_diff`-then-merge benched ~+0.3 ms: the strategy's serial arm is driftsort, which detects the sorted prefix as a natural run and already merges at memcpy speed.
- Removing `resolve_updates`' duplicate-key set benched a wash: a direct isolation probe showed the set costs ~0.1 ms, not the ~0.9 ms phase attribution suggested (the phase is walk+sort bound). The at-scale duplicate-collapse test written to gate that change was kept.
- Cutting the fused-pass shards at 4x the pool width (work-stealing slack) benched ~-0.1 ms on the warm load but was dropped to keep `manual()` fan-outs uniform at one chunk per thread with no tuned factor.
- Merkleize is hash-parallel already (~131k SHA-256s per 32k-update block after the floor raise); forcing every strategy-policy site parallel benched 1 ms worse, so the policy is picking correctly there. Remaining run-to-run p50 spread of +/-0.3-0.9 ms is the adaptive policy converging differently per process on near-tie sites (a resample-interval question for `commonware-parallel`, not this PR).

</details>


glue/src/stateful/db/any.rs
```diff
@@ -193,11 +256,36 @@ where
...
+            },
+        ))
+    }
+
     /// Record a mutation. `Some(value)` for upsert, `None` for delete.
     pub fn write(mut self, key: K, value: Option<V::Value>) -> Self {
         self.batch = self.batch.write(key, value);
         self
     }
+
```
patrick-ogrady: extra \n

glue/src/stateful/db/current.rs
```diff
@@ -100,11 +117,41 @@ where
...
+            },
+        ))
+    }
+
     /// Record a mutation. `Some(value)` for upsert, `None` for delete.
     pub fn write(mut self, key: K, value: Option<V::Value>) -> Self {
         self.batch = self.batch.write(key, value);
         self
     }
+
```
patrick-ogrady: extra \n

glue/src/stateful/db/current.rs
```diff
@@ -195,11 +263,41 @@ where
...
```
patrick-ogrady: extra \n

storage/src/qmdb/any/db.rs
```diff
@@ -299,14 +299,16 @@ where
...
             if results[key_idx].is_some() {
                 continue;
             }
-            let op_idx = positions
-                .binary_search(&pos)
-                .expect("position was deduped from candidates");
+            while positions[op_idx] < pos {
+                op_idx += 1;
+            }
+            debug_assert_eq!(positions[op_idx], pos);
```
patrick-ogrady: assert_eq?

storage/src/qmdb/any/batch.rs
```diff
@@ -47,9 +47,30 @@ type CandidateChunk<'a, F, U> = (&'a [Location<F>], &'a [Operation<F, U>]);
 /// mutations, and cache-resolved keys are always in `updated`.
 type PrevCandidates<K, F, V> = Vec<(K, (Option<V>, Location<F>))>;
 
-/// Committed-DB locations cached by a batch's reads, keyed by the resolved key.
-type ResolvedReads<F, U> =
-    AHashMap<<U as update::Update>::Key, (Location<F>, <U as update::Update>::Cached)>;
+type CachedUpdates<F, U> = Vec<(
```
patrick-ogrady: add a comment

storage/src/qmdb/any/batch.rs
```diff
@@ -47,9 +47,30 @@ type CandidateChunk<'a, F, U> = (&'a [Location<F>], &'a [Operation<F, U>]);
...
+    <U as update::Update>::Key,
+    Location<F>,
+    <U as update::Update>::Cached,
+    <U as update::Update>::Value,
+)>;
+
+struct Staged<F: Family, U: update::Update> {
```
patrick-ogrady: add a comment

storage/src/qmdb/any/batch.rs
```diff
@@ -959,6 +986,100 @@ where
...
+impl<F: Family, H, U, S: Strategy> UpdateMany<F, H, U, S>
+where
+    U: update::Update + Send + Sync,
+    H: Hasher,
+    Operation<F, U>: Codec,
+{
+    /// Record values for the keys returned by `get_many_staged`.
+    ///
+    /// Duplicate keys retain last-write-wins semantics according to the read order.
+    pub fn update_many(mut self, values: &[U::Value]) -> UnmerkleizedBatch<F, H, U, S> {
```
patrick-ogrady: add internal comments above important code blocks like we do elsewhere in the repo

patrick-ogrady (commented):
nits

storage/src/qmdb/benches/constantinople.rs
```diff
@@ -223,23 +223,22 @@ macro_rules! run_pipeline {
...
-            let mut b = new_batch();
-            let values = b.get_many(&keys, &db).await.unwrap();
+            let b = new_batch();
+            let (values, update) = b.get_many_staged(&keys, &db).await.unwrap();
             black_box(&values);
             let t_load = start.elapsed();
-            for (k, v) in &muts {
-                b = b.write(*k, Some(*v));
-            }
+            let b = update.update_many(&update_values);
```
patrick-ogrady: We need to support some being not updated

glue/src/stateful/db/any.rs
```diff
@@ -24,7 +24,7 @@ use commonware_storage::{
     merkle::{Family, Location},
     qmdb::{
         any::{
-            batch::{MerkleizedBatch, UnmerkleizedBatch},
+            batch::{MerkleizedBatch, Staged as BatchStaged, UnmerkleizedBatch},
```
patrick-ogrady: Just used `Staged` directly in these files?

glue/src/stateful/db/any.rs
```diff
@@ -92,13 +109,41 @@ where
...
+        Error<F>,
+    > {
+        let Self {
+            batch,
+            db,
+            metadata,
+        } = self;
+        let guard = db.read().await;
+        let (values, staged) = batch.get_many_staged(keys, &*guard).await?;
+        drop(guard);
```
patrick-ogrady: we don't need these manual drops? we return immediately after?

storage/src/qmdb/any/batch.rs
```diff
@@ -47,9 +47,32 @@ type CandidateChunk<'a, F, U> = (&'a [Location<F>], &'a [Operation<F, U>]);
 /// mutations, and cache-resolved keys are always in `updated`.
 type PrevCandidates<K, F, V> = Vec<(K, (Option<V>, Location<F>))>;
 
-/// Committed-DB locations cached by a batch's reads, keyed by the resolved key.
-type ResolvedReads<F, U> =
-    AHashMap<<U as update::Update>::Key, (Location<F>, <U as update::Update>::Cached)>;
+/// Updates whose old committed location was already resolved by a staged read.
```
patrick-ogrady: I think we can do better than `CachedUpdates` for the name of this? I think it is closer to `CachedReads` or `StagedUpdates`?

storage/src/qmdb/any/batch.rs
```diff
@@ -922,13 +934,28 @@ where
     /// If the same key is written multiple times within a batch, the last
     /// value wins.
     pub fn write(mut self, key: U::Key, value: Option<U::Value>) -> Self {
+        if let Some(idx) = self
```
patrick-ogrady: why do we need to keep both mutations and cached_updates updated?

patrick-ogrady: in fact, why do we even have cached_updates on the batch?

storage/src/qmdb/any/batch.rs
```diff
@@ -959,13 +988,159 @@ where
...
+                let Some(update_idx) = update_idx else {
+                    continue;
+                };
+                self.batch
+                    .mutations
+                    .insert(self.keys[slot].clone(), Some(updates[update_idx].1.clone()));
+            }
+        }
+        self.batch.cached_updates_sorted = self.batch.cached_updates_sorted && !had_cached;
+        Self::apply_new(self.batch, new)
```
patrick-ogrady: should new go first? or does it make sense to go last? what happens if there is an overlap?

patrick-ogrady (commented):
remaining items

patrick-ogrady:
bugbot run

patrick-ogrady:
bugbot run

patrick-ogrady:
bugbot run

glue/src/stateful/db/any.rs
```diff
@@ -92,13 +112,42 @@ where
...
-    /// Returns results in the same order as the input keys. Each unique read is cached on the
-    /// batch for reuse at merkleize, so memory grows with the number of unique keys read.
+    /// Returns results in the same order as the input keys.
     pub async fn get_many(&self, keys: &[&K]) -> Result<Vec<Option<V::Value>>, Error<F>> {
         let db = self.db.read().await;
         self.batch.get_many(keys, &*db).await
     }
 
+    /// Read multiple values and return a staged batch for the same keys.
+    pub async fn stage(
```
patrick-ogrady, line 128: Considered making this an alternative to `batch` but decided it didn't hurt to keep it around (natively supports mixed batch<>stage layering then).

storage/src/qmdb/any/batch.rs
```diff
@@ -945,17 +998,252 @@ where
...
+        C: Mutable<Item = Operation<F, update::Unordered<K, V>>>,
+        I: UnorderedIndex<Value = Location<F>>,
+    {
+        // Unordered deletes emit a `Delete` at the cached location, so they may be staged.
+        let (batch, staged_updates) = self.into_parts(updates, upserts, true);
+        batch
+            .merkleize_with_floor_scan(
+                db,
+                metadata,
+                Some(staged_updates),
```
patrick-ogrady: Keeping staged_updates separate from batch avoids having to add a bunch of handling on batch for an applied staged update that isn't used.

roberto-bayardo:
that PR description is a bit much for my non-AI brain mind making it a bit more concise?

storage/src/journal/authenticated.rs
```diff
@@ -729,13 +739,63 @@ where
...
+                let shard = positions.len().div_ceil(manual.parallelism_hint());
+                manual
+                    .map_collect_vec(positions.chunks(shard).collect::<Vec<_>>(), |shard| {
+                        journal.read_many_sync(shard)
+                    })
+                    .into_iter()
+                    .flatten()
+                    .collect()
+            },
+        );
```
roberto-bayardo, line 759: fable suggests adding an `assert_eq!(results.len(), positions.len());` here to ensure the zip below is WAI.

storage/src/journal/contiguous/variable.rs
```diff
@@ -546,6 +564,132 @@ impl<'a, E: Context, V: CodecShared> Reader<'a, E, V> {
...
+            if let Some(item) = self.try_read_frame_sync(positions[idx], offset, &mut frame_buf) {
+                out[idx] = Some(item);
+                hits += 1;
+            }
+        }
+        self.metrics.items_read.inc_by(hits);
+        out
+    }
+
+    /// Read the varint-framed item for `position` at byte `offset` from cached bytes, returning
```
roberto-bayardo, line 563: do we use the term "frame" for a varint-prefixed variable length record anywhere else?  (I found this confusing.)

roberto-bayardo: oh I see the Frame enum above is pre-existing.  Still seems like we can come up with a better name.

roberto-bayardo:
Would you mind pulling out the parallel strategy autotuner changes into another PR, then rebasing this one on top of it? They seem pretty independent and would make this easier to review.

runtime/src/utils/buffer/paged/sealed.rs
```diff
@@ -160,6 +160,37 @@ impl<B: Blob> Sealed<B> {
         self.view().read_many_into(buf, offsets, item_size).await
     }
 
+    /// Like [`Self::read_many_into`], but skips the dedicated page-cache pass. Intended for
+    /// callers that already probed the cache via [`Self::try_read_many_sync_into`] and are
+    /// reading the misses.
+    pub async fn read_many_direct_into(
```
patrick-ogrady: I think the exposure of `read_many_direct_into` + `try_read_many_sync_into` is a failure of the API...looking for ways to avoid this.

storage/src/journal/contiguous/mod.rs
```diff
@@ -165,31 +165,32 @@ pub trait Contiguous: Send + Sync {
...
-    ///
-    /// The default implementation calls [`read`](Self::read) in a loop. Concrete journal
-    /// implementations override this to batch I/O.
     fn read_many(
         &self,
         positions: &[u64],
     ) -> impl Future<Output = Result<Vec<Self::Item>, Error>> + Send
     where
-        Self::Item: Send,
-    {
```
patrick-ogrady, line 177: Decided to do a pass on removing defaults (there were a handful of impls that didn't override -> leading to slower than expected performance) -> I hypothesize that the LLMs don't deal with defaults well because it relies on mapping code over an abstraction boundary.

glue/src/stateful/db/mod.rs
```diff
@@ -206,15 +206,13 @@ pub trait ManagedDb<E>: Send + Sync + Sized {
...
-    /// default no-op implementation. This call makes changes durable and
-    /// ensures they will be present on startup without replay.
+    /// Databases that do not retain pruneable operation history implement this
+    /// as a no-op. This call makes changes durable and ensures they will be
+    /// present on startup without replay.
     fn prune(
         &mut self,
-        _target: &Self::SyncTarget,
-    ) -> impl Future<Output = Result<(), Self::Error>> + Send {
-        async { Ok(()) }
```
patrick-ogrady, line 216: We can probably restore this default

storage/src/journal/contiguous/mod.rs
roberto-bayardo: I've found default impls cause more problems than they solve and have started avoiding them.

runtime/src/utils/buffer/paged/mod.rs
```diff
@@ -45,75 +45,116 @@ const CHECKSUM_SIZE: u64 = Checksum::SIZE as u64;
...
 
-/// Ensure `buf` has one `item_size` slot per offset, offsets are sorted and non-overlapping, and
-/// every requested range lies within the blob's size.
+/// Ensure every requested item lies within the blob's size.
+///
+/// # Panics
+///
+/// Panics if `buf` is not exactly one `item_size` slot per offset, or if offsets are not sorted
+/// and non-overlapping.
 fn validate_read_many_into(
```
patrick-ogrady: I think we can remove this function and just call the helper directly?

runtime/src/utils/buffer/paged/mod.rs
```diff
@@ -45,75 +45,116 @@ const CHECKSUM_SIZE: u64 = Checksum::SIZE as u64;
...
+/// A fixed-slot batch is the [split_read_ranges] case with a constant per-range length. `buf`
+/// holds one `item_size` slot per offset (validated by [validate_read_many_into]).
+fn split_read_many<'a>(
+    buf: &'a mut [u8],
+    offsets: &[u64],
+    item_size: NonZeroUsize,
+    tail_offset: u64,
+    tail: &[u8],
+) -> Vec<(&'a mut [u8], u64)> {
+    split_read_ranges(
```
patrick-ogrady: Call the helper directly?

storage/src/journal/contiguous/fixed.rs
```diff
@@ -1150,6 +1156,72 @@ impl<E: Context, A: CodecFixedShared> Reader<'_, E, A> {
...
+
+            for slice in buf.chunks_exact(chunk_size) {
+                result.push(A::decode(slice).map_err(Error::Codec)?);
+            }
+        }
+
+        self.metrics.record_cache_hits(hits);
+        self.metrics
+            .record_cache_misses(positions.len() as u64 - hits);
+        self.metrics.items_read.inc_by(positions.len() as u64);
```
patrick-ogrady, line 1227: We should do a pass on our metrics usage here and ensure it is correct. We moved a lot around in this PR and I want to ensure we didn't accidentally blackhole anything.

I do like the removal of the specific "try benchmark" (and focus on capturing higher-level metrics like cache hits/misses.

storage/src/journal/contiguous/variable.rs
```diff
@@ -772,26 +731,409 @@ impl<E: Context, V: CodecShared> super::Contiguous for Reader<'_, E, V> {
...
+                lookup_idx += 1;
+            }
+            if self.validate_readable(position).is_err() {
+                continue;
+            }
+            let Some(offset) = offsets[lookup_idx] else {
+                continue;
+            };
+            resolved[idx] = Some(offset);
+            // The successor lookup is adjacent in `lookups` whenever it was pushed (in
```
patrick-ogrady, line 796: We shouldn't split code blocks with a comment. Either add a \n before the comment or move the comment to the top of the block.

storage/src/journal/contiguous/variable.rs
```diff
@@ -772,26 +731,409 @@ impl<E: Context, V: CodecShared> super::Contiguous for Reader<'_, E, V> {
...
+            }
+        }
+        self.metrics.items_read.inc_by(hits);
+        resolved
+    }
+}
+
+/// Map offsets-journal range errors to corruption: the caller has already validated the
+/// positions against `bounds`, so the offsets journal must have them.
+fn offsets_corruption(e: Error) -> Error {
```
patrick-ogrady: is a standalone function necessary?

storage/src/journal/contiguous/variable.rs
```diff
@@ -772,26 +731,409 @@ impl<E: Context, V: CodecShared> super::Contiguous for Reader<'_, E, V> {
...
+
+    async fn fetch_missing(self) -> Result<(Vec<u64>, Vec<V>), Error> {
+        let Self {
+            reader,
+            items: _,
+            mut misses,
+        } = self;
+        let _timer = reader.metrics.read_many_timer();
+        reader.metrics.read_many_calls.inc();
+        if !misses.windows(2).all(|w| w[0].position < w[1].position) {
```
patrick-ogrady: Add a note about when this could not happen?

storage/src/qmdb/any/batch.rs
```diff
@@ -945,17 +1079,222 @@ where
...
+    /// Expansion does not deduplicate against previously staged keys. Reading the same key again
+    /// creates another staged slot in the returned range. If both slots are later updated,
+    /// [`merkleize`](Staged::merkleize) applies the update list's normal last-write-wins
+    /// semantics.
+    ///
+    /// Expansion reads through the underlying batch, ancestor batches, and committed database state.
+    /// Values the caller has computed for earlier staged slots are not visible until they are passed
+    /// to [`merkleize`](Staged::merkleize). Callers that need speculative read-your-writes behavior
+    /// should maintain their own overlay while deciding which staged slots to update.
+    pub async fn expand<E, C, I, const N: usize>(
```
patrick-ogrady, line 1121: I wonder if stage/expand should have tracing?

storage/src/qmdb/any/batch.rs
```diff
@@ -1317,41 +1722,20 @@ where
...
-                        None
-                    }
-                    (_, mutation) => Some((key, mutation)),
-                })
-                .collect();
-        }
+        // Staged updates skip the index probe and journal re-read, and their old op's next key
+        // feeds the candidate sets directly. The ordered path never stages deletes (see
+        // `Staged::resolve_updates`), so every staged entry carries a value.
+        let cached = staged_updates;
```
patrick-ogrady: Do we need to shadow this var?

storage/src/qmdb/any/mod.rs
```diff
@@ -218,14 +218,25 @@ pub(crate) mod test {
...
+        fixed_db_config_with_strategy(suffix, pooler, Sequential)
+    }
+
+    pub(crate) fn fixed_db_config_with_strategy<
+        T: Translator + Default,
+        S: commonware_parallel::Strategy,
+    >(
+        suffix: &str,
+        pooler: &impl BufferPooler,
+        strategy: S,
```
patrick-ogrady, line 231: Just add this to `fixed_db_config`?

storage/src/qmdb/immutable/sync/tests.rs
```diff
@@ -1144,6 +1144,62 @@ macro_rules! sync_tests_for_harness {
 sync_tests_for_harness!(harnesses::VariableMmrHarness, variable_mmr);
 sync_tests_for_harness!(harnesses::VariableMmbHarness, variable_mmb);
 
+/// A completed sync journal reuses local boundary nodes only when the persisted state can
+/// authenticate the target: a target starting below the local pruning boundary is declined,
+/// while a matching target serves the boundary nodes locally.
+#[commonware_macros::test_traced]
+fn test_immutable_local_boundary_nodes_rejects_target_before_local_lower_bound() {
+    use harnesses::VariableMmrHarness as H;
```
patrick-ogrady: We should import at the top of the mod?

storage/src/qmdb/keyless/sync/tests.rs
```diff
@@ -1024,6 +1024,62 @@ macro_rules! sync_tests_for_harness {
 sync_tests_for_harness!(harnesses::VariableMmrHarness, variable_mmr);
 sync_tests_for_harness!(harnesses::VariableMmbHarness, variable_mmb);
 
+/// A completed sync journal reuses local boundary nodes only when the persisted state can
+/// authenticate the target: a target starting below the local pruning boundary is declined,
+/// while a matching target serves the boundary nodes locally.
+#[commonware_macros::test_traced]
+fn test_keyless_local_boundary_nodes_rejects_target_before_local_lower_bound() {
+    use harnesses::VariableMmrHarness as H;
```
patrick-ogrady: top of mod?

patrick-ogrady (commented):
final nits

storage/src/journal/contiguous/fixed.rs
```diff
@@ -1270,19 +1233,241 @@ impl<E: Context, A: CodecFixedShared> super::Contiguous for Reader<'_, E, A> {
...
+                blob.try_read_many_sync_into(buf, &blob_offsets, Journal::<E, A>::CHUNK_SIZE);
+            let mut misses = misses.into_iter().peekable();
+            for (idx, slice) in buf.chunks_exact(chunk_size).enumerate() {
+                if misses.peek() == Some(&idx) {
+                    misses.next();
+                    continue;
+                }
+                // A decode failure surfaces as a miss. The async completion read
+                // reports the error.
+                if let Ok(item) = A::decode(slice) {
```
patrick-ogrady, line 1297: I don't think this can ever happen? a decode failure is a panic?

storage/src/journal/contiguous/blobs.rs
```diff
@@ -538,6 +557,53 @@ impl<'a, B: RBlob> Blob<'a, B> {
                 .map_err(Error::Runtime),
         }
     }
+
+    /// Like [`Self::read_many_into`], but skips the dedicated page-cache pass. Intended for
+    /// callers that already probed the cache via [`Self::try_read_many_sync_into`] and are
+    /// reading the misses.
+    pub(super) async fn read_misses_into(
```
patrick-ogrady: I'm still note convinced leaking a cache bypass into the public API is a good idea?

storage/src/journal/authenticated.rs
```diff
@@ -729,13 +734,43 @@ where
     }
 
     async fn read_many(&self, positions: &[u64]) -> Result<Vec<C::Item>, JournalError> {
-        self.journal.read_many(positions).await
+        if positions.is_empty() {
```
patrick-ogrady, line 734: I don't think we need this early return?

storage/src/journal/authenticated.rs
```diff
@@ -759,6 +794,35 @@ where
         Ok(*res)
     }
 
+    async fn append_many<'a>(
+        &'a mut self,
+        items: Many<'a, Self::Item>,
+    ) -> Result<u64, JournalError> {
+        if items.is_empty() {
```
patrick-ogrady, line 818: another case of unnecessary early return?

patrick-ogrady (commented):
final nits

storage/src/journal/contiguous/variable.rs
```diff
@@ -778,24 +737,414 @@ impl<E: Context, V: CodecShared> super::Contiguous for Reader<'_, E, V> {
...
+            // as resolved vs unresolved (offsets are immutable under a shared borrow), so
+            // keep the resolved one.
+            misses.dedup_by(|next, kept| {
+                if next.position != kept.position {
+                    return false;
+                }
+                if kept.offset.is_none() {
+                    kept.offset = next.offset;
+                }
+                true
```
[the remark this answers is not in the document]
patrick-ogrady: **Refuted.** `Vec::dedup_by` passes elements in the opposite order from their order in the slice: the first closure argument is the later element (the one removed when the closure returns `true`) and the second is the earlier survivor. From the std docs: "The elements are passed in opposite order from their order in the slice, so if `same_bucket(a, b)` returns `true`, `a` is removed." That is exactly what the parameter names encode. `next` is the removed duplicate and `kept` is the survivor, so `kept.offset = next.offset` copies the resolved offset into the row that remains.

Verified by executing the exact closure standalone over four cases, including this finding's scenario (earlier row `None`, later duplicate `Some`), plus `(Some, None)`, `(None, Some, None)`, and duplicates interleaved with distinct positions. The survivor carries the resolved offset in every case.

On stakes: even if the direction were inverted, `fetch_misses` re-resolves any `None` offset from the offsets journal, which is immutable under the reader's borrow. The worst case is one redundant cached read, never a wrong item.

The reuse is now pinned by `test_variable_fetch_missing_reuses_deduped_offset` (landing with the next push): it constructs the real `(unresolved, resolved)` duplicate through the public API (cold probe, warm the shared offsets page via a neighboring read, re-probe, merge, `fetch_missing`) and asserts the offsets journal serves zero additional reads during completion. Inverting the dedup direction makes it fail with exactly the one extra offsets read this finding predicts.

storage/src/journal/contiguous/metrics.rs
```diff
@@ -2,42 +2,26 @@
...
-
-/// Metrics common to contiguous journal implementations.
-pub(super) struct CommonMetrics<E: Clock> {
-    /// Clock used for duration timers.
-    clock: Arc<E>,
+/// Metrics for a contiguous journal.
+///
+/// Fixed and variable journals register the same set. Sub-components (the blob manager and the
+/// variable journal's offsets journal) register their own metrics under child contexts.
+pub(super) struct Metrics<E: Clock> {
```
patrick-ogrady, line 16: Admittedly, I think most of these metrics could be removed.

storage/src/journal/contiguous/variable.rs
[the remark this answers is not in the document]
patrick-ogrady: Follow-up: the `Probed` token API (including `fetch_missing` and the `dedup_by` this thread discussed) has since been removed from the branch entirely. `try_read_many_sync` now returns a plain `Vec<Option<_>>` and completion goes through ordinary `read_many`, so cross-probe offset state no longer exists. The offset reuse survives inside a single `read_many` call (sync pass resolves, completion carries) and is pinned by `test_variable_read_many_reuses_probed_offset` instead of the merge-dedup test mentioned above.

patrick-ogrady (commented):
I ended up walking back some "re-probe optimizations" to flatten this PR a bit. Issue here: https://github.com/commonwarexyz/monorepo/issues/4192

PR #4145, opened 2026-06-30
[runtime] Implement `start_sync` for `buffer::Write`

danlaine:
- Add `start_sync` to `commonware_runtime::buffer::Write`, matching the `paged::Writer` flow: flush buffered bytes with plain writes, then return a `Handle` from `SyncState::start_sync` so durability can complete asynchronously.
- Preserve ordering around pending sync barriers. Later operations that mutate the blob wait for any outstanding `start_sync` before issuing blob writes, resizes, or syncs, so newer bytes are not folded into an older durability barrier.
- Keep `sync` semantics unchanged for callers: buffer-only writes can still use `write_at_sync` when clean, while prior plain writes and resizes still require a full sync barrier.
- Share the delayed `start_sync` test blob with `paged::Writer` tests and add raw `Write` coverage for successful start, pending-sync ordering, writes after start, and resize ordering.

### DIFF
--- a/runtime/src/utils/buffer/write.rs
+++ b/runtime/src/utils/buffer/write.rs
@@ -225,11 +225,7 @@ impl<B: Blob> Write<B> {
 
     /// Flush buffered bytes and durably sync mutations tracked by this writer.
     pub async fn sync(&mut self) -> Result<(), Error> {
-        if !self.buffer.is_empty() {
-            let (buf, offset) = self
-                .buffer
-                .take()
-                .expect("take must succeed when sync observes buffered data");
+        if let Some((buf, offset)) = self.buffer.take() {
             return self.write_blob_sync(offset, buf).await;
         }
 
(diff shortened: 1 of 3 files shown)


runtime/src/utils/buffer/write.rs
patrick-ogrady: I think this is unguarded?

danlaine: `self.write_blob` is `self.sync_state.write_at(&self.blob, offset, bufs).await` which await in-flight syncs:

```rs
async fn write_at(
        &mut self,
        blob: &impl crate::Blob,
        offset: u64,
        bufs: impl Into<crate::IoBufs> + Send,
    ) -> Result<(), crate::Error> {
        self.wait_for_pending().await?;
        blob.write_at(offset, bufs).await?;
        *self = Self::Dirty;
        Ok(())
    }
```

danlaine: Removed write-through `self.write_blob` to (hopefully) make this more clear

patrick-ogrady (approved):
Looks good (modulo the incoming fixes in: https://github.com/commonwarexyz/monorepo/issues/4148)

PR #4147, opened 2026-07-01
[cryptography] Refactor Hasher trait for small hash opts

cronokirby:
This adds a method taking in a slice of slices, replacing the existing `hash` method taking in one byte slice, allowing implementors, like Sha256, to implement this more efficiently, in the case of small hashes.

This along with the addition of an inline annotation (allowing the method to be inlined across crates), sees most of the performance gain compared with https://github.com/commonwarexyz/monorepo/pull/4127. Maintaining the scratch space within the Hasher struct for Sha is not actually necessary for the performance gain.

There's a remaining gap, owing to not special casing some methods which could use a fixed array instead. I think adding a bunch of special-cased methods to the cryptography crate is an anti-pattern, because it makes the API surface much more complicated for little gain. It's better to have fewer primitives that compose well, rather than creating a shortcut to paper over the API.

This change also makes it so that an instance of Hasher does not have call patterns which can be misused, obviating the need for an additional Pending struct, or some other solution. Instead, you can either hash data all at once, or if you need to hash incrementally, you can use an instance of Hasher. Finalizing returns a new Hasher, to recycle large Hasher state in case some impl ends up doing that. Right now for Sha this isn't a concern, because the struct itself owns very little data (< 128 bytes). For BLAKE3, this is about 1kb, so this can make sense. I opted to just have finalize always return the new Hasher, and ignore it at most call sites, rather than add an extra method.


cryptography/src/lib.rs
```diff
@@ -245,35 +245,35 @@ commonware_macros::stability_scope!(BETA {
...
-
-        /// Append message to previously recorded data.
-        fn update(&mut self, message: &[u8]) -> &mut Self;
+        /// Hash the concatenation of `parts` in a single shot.
+        ///
+        /// This is the preferred entrypoint for hashing data that is fully
+        /// available up-front. Implementations are free to specialize this for
+        /// small, fixed-shape inputs (e.g. hashing a pair of digests) to avoid
+        /// the overhead of the streaming machinery.
+        fn hash(parts: &[&[u8]]) -> Self::Digest;
```
patrick-ogrady, line 265: If we made `BLOCK_SIZE` an associated type and allowed impls to just copy into slices based on that, we could avoid the scratch copy (and just hand as-is to compress).

patrick-ogrady: It may get weird when packing on the "boundary" but we may be able to provide a writer as a Buf or something?

cryptography/src/lib.rs
```diff
@@ -245,35 +245,35 @@ commonware_macros::stability_scope!(BETA {
...
-        /// This function does not need to be called after `finalize`.
-        fn reset(&mut self) -> &mut Self;
-
-        /// Hash a single message with a one-time-use hasher.
-        fn hash(message: &[u8]) -> Self::Digest {
-            Self::new().update(message).finalize()
-        }
+        /// Returning the reset hasher allows callers that hash many values in a
+        /// loop to reuse the underlying allocation without exposing a separate
+        /// `reset` method.
```
roberto-bayardo: I'd drop this, no need to reflect on previous implementations.

PR #4150, opened 2026-07-01
[cryptography] Use fixed preimage arrays for hashing

danlaine:
## Summary

This PR builds on #4147 and keeps its small multi-part one-shot hashing API, but changes the final shape to use fixed-size preimage arrays and in-place streaming finalization.

- Changes `Hasher::hash` to `hash<const N: usize>([&[u8]; N])`, so fixed preimages are represented directly without allocating.
- Keeps the streaming API as `update(&mut self)`, `finalize(&mut self)`, and `reset(&mut self)`.
- Removes `Hasher::new`; callers construct hashers with `Default`.
- Keeps #4147's SHA-256 stack fast path for small one-shot preimages.
- Updates Merkle storage hashing to use fixed preimage arrays, so `leaf_digest`, `node_digest`, `fold`, and fixed root preimages hit the optimized SHA-256 path.
- Keeps #4147's BMT proof-verification cleanup: verification methods construct their own hasher instead of requiring callers to pass `&mut hasher`.

## Why This Shape

This keeps one-shot hashing and streaming hashing semantically separate.

```rust
let digest = H::hash([prefix, body]);

hasher.update(chunk);
let digest = hasher.finalize();
```

The one-shot path is a closed operation: previous mutable hasher state is never involved. The streaming path remains the familiar state machine.

Compared to PR #4147, this keeps the main performance idea but avoids the consuming `finalize(self) -> (Self, Digest)` API. That tuple-return design makes stored hashers noisier to use and forces call sites to care about rebinding the returned hasher even when they only need a digest. Here, `finalize(&mut self)` continues to reset in place as it did on `origin/main`.

The const-generic array also makes fixed preimages more direct than #4147's `&[&[u8]]` while still preserving the same one-shot optimization point:

```rust
H::hash([message]);
H::hash([&pos.to_be_bytes(), left, right]);
```


cryptography/src/lib.rs
```diff
@@ -262,18 +262,25 @@ commonware_macros::stability_scope!(BETA {
         /// available up-front. Implementations are free to specialize this for
         /// small, fixed-shape inputs (e.g. hashing a pair of digests) to avoid
         /// the overhead of the streaming machinery.
-        fn hash(parts: &[&[u8]]) -> Self::Digest;
+        fn hash<const N: usize>(parts: [&[u8]; N]) -> Self::Digest {
```
cronokirby, line 265: From what I could gather this doesn't actually help, provided the method is inlined, at which point the compiler + LLVM is able to apply this optimization itself.

PR #4152, opened 2026-07-01
[storage] QMDB: drop hasher params from proof/verify API; parallelize recovery replay

roberto-bayardo:
Two related QMDB improvements, split out of #4127. The hashing-performance work from that PR is deferred to a separate follow-up (benchmarks showed the proposed slice-based `Hasher` API regressed merkle hashing ~15-20%).

## 1. Remove the hasher parameter from QMDB's proof/verify API

QMDB's proof, verify, and root helpers took a `&Standard<H>` hasher threaded down from callers. The hasher is fully determined by the type parameter `H` (the root bagging policy is fixed, via `qmdb::hasher::<H>()`), so passing it was redundant. This reconstructs it internally, drops the parameter, and removes the now-redundant generic parameters it enabled (e.g. `verify_proof<F, Op, H, D>` becomes `verify_proof<H, F, Op>`).

## 2. Parallelize authenticated-journal recovery replay

During `align()` recovery replay, hash each apply-batch's leaves across the merkleization strategy (mirroring `add_many`) instead of serially, one op at a time. Byte-identical to the serial path; only the leaf hashing is parallelized.

## Testing

- `just test -p commonware-storage` — 2677/2677 pass
- `just test -p commonware-glue` — 140/140 pass
- `just test-conformance` — 288/288 pass (hashing is unchanged from `main`, so storage/codec formats are byte-identical)
- `cargo check --workspace --all-targets` — clean

https://claude.ai/code/session_01ViJ2JVK4RX8Eyh8rvbL2zz

### DIFF
--- a/storage/src/journal/authenticated.rs
+++ b/storage/src/journal/authenticated.rs
@@ -423,12 +423,27 @@ where
             while merkle_leaves < journal_size {
                 let batch = {
                     let mut batch = merkle.new_batch();
+                    let first = batch.leaves();
+                    let mut items = Vec::new();
                     let mut count = 0u64;
                     while count < apply_batch_size && merkle_leaves < journal_size {
                         let op = journal.read(*merkle_leaves).await?;
-                        batch = batch.add(hasher, &op.encode());
+                        items.push(op);
                         merkle_leaves += 1;
                         count += 1;
+                    }
+
+                    // Hash the batch's leaves in parallel, then append them.
+                    let digests =
+                        batch
+                            .strategy()
+                            .map_collect_vec(items.iter().enumerate(), |(i, item)| {
+                                let pos = Position::try_from(first + i as u64)
+                                    .expect("valid leaf location");
+                                hasher.leaf_digest(pos, item.encode().as_ref())
+                            });
+                    for digest in digests {
+                        batch = batch.add_leaf_digest(digest);
                     }
                     batch
                 };
--- a/storage/src/journal/authenticated.rs
+++ b/storage/src/journal/authenticated.rs
@@ -23,7 +23,7 @@ use alloc::{
     sync::{Arc, Weak},
     vec::Vec,
 };
-use commonware_codec::{CodecFixedShared, CodecShared, Encode, EncodeShared};
+use commonware_codec::{CodecFixedShared, CodecShared, Encode, EncodeShared, Write};
 use commonware_cryptography::{Digest, Hasher};
 use commonware_macros::boxed;
 use commonware_parallel::Strategy;
@@ -433,15 +433,19 @@ where
                         count += 1;
                     }
 
-                    // Hash the batch's leaves in parallel, then append them.
-                    let digests =
-                        batch
-                            .strategy()
-                            .map_collect_vec(items.iter().enumerate(), |(i, item)| {
-                                let pos = Position::try_from(first + i as u64)
-                                    .expect("valid leaf location");
-                                hasher.leaf_digest(pos, item.encode().as_ref())
-                            });
+                    // Hash the batch's leaves in parallel, then append them. A per-worker scratch
+                    // buffer avoids a per-leaf allocation (mirrors `add_many`).
+                    let digests = batch.strategy().map_init_collect_vec(
+                        items.iter().enumerate(),
+                        Vec::new,
+                        |buf, (i, item)| {
+                            let pos =
+                                Position::try_from(first + i as u64).expect("valid leaf location");
+                            buf.clear();
+                            item.write(buf);
+                            hasher.leaf_digest(pos, buf.as_slice())
+                        },
+                    );
                     for digest in digests {
                         batch = batch.add_leaf_digest(digest);
                     }
--- a/storage/src/journal/authenticated.rs
+++ b/storage/src/journal/authenticated.rs
@@ -11,11 +11,8 @@ use crate::{
         Error as JournalError,
     },
     merkle::{
-        self, batch,
-        full::Merkle,
-        hasher::{Hasher as _, Standard as StandardHasher},
-        mem::Mem,
-        Bagging, Family, Location, Position, Proof, Readable,
+        self, batch, full::Merkle, hasher::Standard as StandardHasher, mem::Mem, Bagging, Family,
+        Location, Position, Proof, Readable,
     },
     Context,
 };
@@ -23,7 +20,7 @@ use alloc::{
     sync::{Arc, Weak},
     vec::Vec,
 };
-use commonware_codec::{CodecFixedShared, CodecShared, Encode, EncodeShared, Write};
+use commonware_codec::{CodecFixedShared, CodecShared, Encode, EncodeShared};
 use commonware_cryptography::{Digest, Hasher};
 use commonware_macros::boxed;
 use commonware_parallel::Strategy;
@@ -130,20 +127,7 @@ impl<F: Family, H: Hasher, Item: Encode + Send + Sync, S: Strategy>
             "add_many expects no items added via add"
         );
 
-        let first = self.inner.leaves();
-        let hasher = &self.hasher;
-        let digests = self.inner.strategy().map_init_collect_vec(
-            items.iter().enumerate(),
-            Vec::new,
-            |buf, (i, item)| {
-                let pos = Position::try_from(first + i as u64).expect("valid leaf location");
-                buf.clear();
-                item.write(buf);
-                hasher.leaf_digest(pos, buf.as_slice())
-            },
-        );
-
-        self.inner = self.inner.add_leaf_digests(digests);
+        self.inner = self.inner.add_many(&self.hasher, &items);
         self.items = items;
         self
     }
@@ -390,8 +374,9 @@ where
 
     /// Align the Merkle structure to be consistent with the journal. Any items in the structure
     /// that are not in the journal are popped, and any items in the journal that are not in the
-    /// structure are added. Items are added in batches of size `apply_batch_size` to avoid memory
-    /// bloat.
+    /// structure are added. Items are added in batches of size `apply_batch_size` to bound peak
+    /// memory use: each batch's items are buffered in memory so their leaves can be hashed
+    /// across the strategy.
     async fn align(
         merkle: &mut Merkle<F, E, H::Digest, S>,
         journal: &C,
@@ -421,47 +406,14 @@ where
             );
 
             while merkle_leaves < journal_size {
-                let first = merkle_leaves;
-                let mut items = Vec::new();
-                let mut count = 0u64;
-                while count < apply_batch_size && merkle_leaves < journal_size {
-                    let op = journal.read(*merkle_leaves).await?;
-                    items.push(op);
+                let count = apply_batch_size.min(journal_size - *merkle_leaves);
+                let mut items = Vec::with_capacity(count as usize);
+                for _ in 0..count {
+                    items.push(journal.read(*merkle_leaves).await?);
                     merkle_leaves += 1;
-                    count += 1;
                 }
 
-                // Hash the batch's leaves across the strategy. The serial path appends in place;
-                // the parallel path hashes with a per-worker scratch buffer to avoid a per-leaf
-                // allocation (mirrors `add_many`), then appends the collected digests.
-                let strategy = merkle.strategy();
-                let batch = strategy.run(
-                    items.len(),
-                    || {
-                        let mut buf = Vec::new();
-                        let mut batch = merkle.new_batch();
-                        for item in &items {
-                            buf.clear();
-                            item.write(&mut buf);
-                            batch = batch.add(hasher, &buf);
-                        }
-                        batch
-                    },
-                    || {
-                        let digests = strategy.map_init_collect_vec(
-                            items.iter().enumerate(),
-                            Vec::new,
-                            |buf, (i, item)| {
-                                let pos = Position::try_from(first + i as u64)
-                                    .expect("valid leaf location");
-                                buf.clear();
-                                item.write(buf);
-                                hasher.leaf_digest(pos, buf.as_slice())
-                            },
-                        );
-                        merkle.new_batch().add_leaf_digests(digests)
-                    },
-                );
+                let batch = merkle.new_batch().add_many(hasher, &items);
                 let batch = merkle.with_mem(|mem| batch.merkleize(mem, hasher));
                 merkle.apply_batch(&batch)?;
             }
@@ -1227,9 +1179,9 @@ mod tests {
         journal.sync().await.unwrap();
 
         // Replay with a batch size that forces multiple batches on each side. `Sequential`
-        // always takes `run()`'s serial arm, and a `Manual`-wrapped strategy with more than one
-        // thread always takes the parallel arm (no adaptive policy), so the two replays are
-        // guaranteed to exercise both paths deterministically.
+        // hashes each batch serially, and a `Manual`-wrapped strategy runs the batch hashing
+        // across its pool without any adaptive policy, so the two replays deterministically
+        // exercise both the serial and parallel hashing paths.
         let hasher = StandardHasher::<Sha256>::new(ForwardFold);
         let mut serial = Merkle::<F, _, Digest, Sequential>::init(
             context.child("mmr_serial"),
--- a/storage/src/journal/authenticated.rs
+++ b/storage/src/journal/authenticated.rs
@@ -876,7 +876,7 @@ mod tests {
     use commonware_codec::Encode;
     use commonware_cryptography::{sha256::Digest, Sha256};
     use commonware_macros::test_traced;
-    use commonware_parallel::Sequential;
+    use commonware_parallel::{Rayon, Sequential};
     use commonware_runtime::{
         buffer::paged::CacheRef,
         deterministic::{self, Context},
@@ -923,16 +923,25 @@ mod tests {
         batch.add_many(items).merkleize(base)
     }
 
-    /// Create Merkle configuration for tests.
-    fn merkle_config(suffix: &str, pooler: &impl BufferPooler) -> MerkleConfig<Sequential> {
+    /// Create Merkle configuration for tests with the given strategy.
+    fn merkle_config_with<S: Strategy>(
+        suffix: &str,
+        pooler: &impl BufferPooler,
+        strategy: S,
+    ) -> MerkleConfig<S> {
         MerkleConfig {
             journal_partition: format!("mmr-journal-{suffix}"),
             metadata_partition: format!("mmr-metadata-{suffix}"),
             items_per_blob: NZU64!(11),
             write_buffer: NZUsize!(1024),
-            strategy: Sequential,
+            strategy,
             page_cache: CacheRef::from_pooler(pooler, PAGE_SIZE, PAGE_CACHE_SIZE),
         }
+    }
+
+    /// Create Merkle configuration for tests.
+    fn merkle_config(suffix: &str, pooler: &impl BufferPooler) -> MerkleConfig<Sequential> {
+        merkle_config_with(suffix, pooler, Sequential)
     }
 
     /// Create journal configuration for tests.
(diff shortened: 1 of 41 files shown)


storage/src/journal/authenticated.rs
```diff
@@ -423,13 +423,28 @@ where
...
+                        items.push(op);
                         merkle_leaves += 1;
                         count += 1;
                     }
+
+                    // Hash the batch's leaves in parallel, then append them.
+                    let digests =
+                        batch
+                            .strategy()
+                            .map_collect_vec(items.iter().enumerate(), |(i, item)| {
```
patrick-ogrady: I think `fold` may allow you to insert into batch directly in order instead of creating this intermediate allocation

storage/src/qmdb/current/sync/mod.rs
```diff
@@ -131,7 +131,7 @@ where
     let log = authenticated::Journal::<F, _, _, _, S>::from_components(
         merkle,
         log,
-        hasher,
+        hasher.clone(),
```
patrick-ogrady: I assume we still need a concrete hasher here for some reason?

storage/src/qmdb/current/grafting.rs
```diff
@@ -345,39 +345,35 @@ impl<F: Graftable, H: CHasher> HasherTrait<F> for Verifier<'_, F, H> {
...
-        hasher: H,
-    ) -> Self {
+    pub(super) const fn new(grafted_tree: &'a G, grafting_height: u32, ops_tree: &'a S) -> Self {
         Self {
             grafted_tree,
             grafting_height,
             ops_tree,
-            grafted_hasher: GraftedHasher::new(hasher, grafting_height),
+            grafted_hasher: GraftedHasher::new(
+                merkle::hasher::Standard::new(merkle::Bagging::ForwardFold),
```
patrick-ogrady: was it intentional to hardcode a bagging strategy here?

patrick-ogrady (commented):
minor nits

storage/src/qmdb/current/grafting.rs
roberto-bayardo: No, not sure why Claude decided to insert this, reverted.

storage/src/journal/authenticated.rs
roberto-bayardo: Not sure I understand how to use fold here without forgoing the new parallelization of hashing here.  But I did update it to exactly match the read_many implementation (which adds using commonware_codec::Write & a re-usable buffer to avoid per-leaf allocation).

storage/src/qmdb/current/sync/mod.rs
roberto-bayardo: Yes -- `from_components` takes the hasher by value because the authenticated journal is a lower-layer primitive that stays bagging-agnostic (it can't reach up into `qmdb::hasher()`). But you're right that the `.clone()` was pointless here: this binding was only used at this call, and the partial/pending digests below re-materialize their own. Inlined `qmdb::hasher::<H>()` at the call site and dropped the stray binding.

storage/src/journal/authenticated.rs
patrick-ogrady: I misspoke. This isn't possible.

storage/src/journal/authenticated.rs
```diff
@@ -423,13 +423,32 @@ where
...
                         let op = journal.read(*merkle_leaves).await?;
-                        batch = batch.add(hasher, &op.encode());
+                        items.push(op);
                         merkle_leaves += 1;
                         count += 1;
                     }
+
+                    // Hash the batch's leaves in parallel, then append them. A per-worker scratch
+                    // buffer avoids a per-leaf allocation (mirrors `add_many`).
+                    let digests = batch.strategy().map_init_collect_vec(
```
patrick-ogrady: you can use `strategy().run()` here to do a simple iteration in-place and the code you have below for the parallel path to avoid the extra overhead of the second move (then let strategy automatically pick the best)

storage/src/journal/authenticated.rs
```diff
@@ -1188,6 +1197,85 @@ mod tests {
...
+            .await
+            .unwrap();
+
+        let mut parallel = Merkle::<F, _, Digest, Rayon>::init(
+            context.child("mmr_parallel"),
+            &hasher,
+            merkle_config_with(
+                "replay-parallel",
+                &context,
+                Rayon::new(NZUsize!(2)).unwrap(),
```
patrick-ogrady: We should avoid using rayon inside of deterministic? I don't think this will cause an issue because it uses its own threads when constructed this way but feels like an anti-pattern?

storage/src/journal/authenticated.rs
```diff
@@ -1188,6 +1197,85 @@ mod tests {
...
+        .unwrap();
+        TestJournal::<F>::align(&mut serial, &journal, &hasher, 7)
+            .await
+            .unwrap();
+
+        let mut parallel = Merkle::<F, _, Digest, Rayon>::init(
+            context.child("mmr_parallel"),
+            &hasher,
+            merkle_config_with(
+                "replay-parallel",
```
patrick-ogrady, line 1201: just initializing with parallel, doesn't mean we actually hit both sides of run (there are rules within the autotuner that describe what we do)

patrick-ogrady:
Fable suggested the following (haven't reviewed myself yet):

```
diff --git a/storage/src/journal/authenticated.rs b/storage/src/journal/authenticated.rs
index 9af862a7b..b8c00fd35 100644
--- a/storage/src/journal/authenticated.rs
+++ b/storage/src/journal/authenticated.rs
@@ -13,7 +13,7 @@ use crate::{
     merkle::{
         self, batch,
         full::Merkle,
-        hasher::{Hasher as _, Standard as StandardHasher},
+        hasher::Standard as StandardHasher,
         mem::Mem,
         Bagging, Family, Location, Position, Proof, Readable,
     },
@@ -23,7 +23,7 @@ use alloc::{
     sync::{Arc, Weak},
     vec::Vec,
 };
-use commonware_codec::{CodecFixedShared, CodecShared, Encode, EncodeShared, Write};
+use commonware_codec::{CodecFixedShared, CodecShared, Encode, EncodeShared};
 use commonware_cryptography::{Digest, Hasher};
 use commonware_macros::boxed;
 use commonware_parallel::Strategy;
@@ -130,20 +130,7 @@ impl<F: Family, H: Hasher, Item: Encode + Send + Sync, S: Strategy>
             "add_many expects no items added via add"
         );
 
-        let first = self.inner.leaves();
-        let hasher = &self.hasher;
-        let digests = self.inner.strategy().map_init_collect_vec(
-            items.iter().enumerate(),
-            Vec::new,
-            |buf, (i, item)| {
-                let pos = Position::try_from(first + i as u64).expect("valid leaf location");
-                buf.clear();
-                item.write(buf);
-                hasher.leaf_digest(pos, buf.as_slice())
-            },
-        );
-
-        self.inner = self.inner.add_leaf_digests(digests);
+        self.inner = self.inner.add_many(&self.hasher, &items);
         self.items = items;
         self
     }
@@ -390,8 +377,9 @@ where
 
     /// Align the Merkle structure to be consistent with the journal. Any items in the structure
     /// that are not in the journal are popped, and any items in the journal that are not in the
-    /// structure are added. Items are added in batches of size `apply_batch_size` to avoid memory
-    /// bloat.
+    /// structure are added. Items are added in batches of size `apply_batch_size` to bound peak
+    /// memory use: each batch's items are buffered in memory so their leaves can be hashed
+    /// across the strategy.
     async fn align(
         merkle: &mut Merkle<F, E, H::Digest, S>,
         journal: &C,
@@ -421,47 +409,14 @@ where
             );
 
             while merkle_leaves < journal_size {
-                let first = merkle_leaves;
-                let mut items = Vec::new();
-                let mut count = 0u64;
-                while count < apply_batch_size && merkle_leaves < journal_size {
-                    let op = journal.read(*merkle_leaves).await?;
-                    items.push(op);
+                let count = apply_batch_size.min(journal_size - *merkle_leaves);
+                let mut items = Vec::with_capacity(count as usize);
+                for _ in 0..count {
+                    items.push(journal.read(*merkle_leaves).await?);
                     merkle_leaves += 1;
-                    count += 1;
                 }
 
-                // Hash the batch's leaves across the strategy. The serial path appends in place;
-                // the parallel path hashes with a per-worker scratch buffer to avoid a per-leaf
-                // allocation (mirrors `add_many`), then appends the collected digests.
-                let strategy = merkle.strategy();
-                let batch = strategy.run(
-                    items.len(),
-                    || {
-                        let mut buf = Vec::new();
-                        let mut batch = merkle.new_batch();
-                        for item in &items {
-                            buf.clear();
-                            item.write(&mut buf);
-                            batch = batch.add(hasher, &buf);
-                        }
-                        batch
-                    },
-                    || {
-                        let digests = strategy.map_init_collect_vec(
-                            items.iter().enumerate(),
-                            Vec::new,
-                            |buf, (i, item)| {
-                                let pos = Position::try_from(first + i as u64)
-                                    .expect("valid leaf location");
-                                buf.clear();
-                                item.write(buf);
-                                hasher.leaf_digest(pos, buf.as_slice())
-                            },
-                        );
-                        merkle.new_batch().add_leaf_digests(digests)
-                    },
-                );
+                let batch = merkle.new_batch().add_many(hasher, &items);
                 let batch = merkle.with_mem(|mem| batch.merkleize(mem, hasher));
                 merkle.apply_batch(&batch)?;
             }
@@ -1226,9 +1181,9 @@ mod tests {
         journal.append(&commit_op).await.unwrap();
         journal.sync().await.unwrap();
 
-        // Replay with a batch size that forces multiple batches. A fresh Rayon strategy runs
-        // its first batch on the parallel path (the policy seeds parallel timing before
-        // serial), so both replay paths are exercised.
+        // Replay with a batch size that forces multiple batches. A fresh Rayon strategy samples
+        // the parallel path on its first call and probes serial on a later one (see
+        // commonware_parallel::policy), so both hashing paths are exercised across the batches.
         let hasher = StandardHasher::<Sha256>::new(ForwardFold);
         let mut serial = Merkle::<F, _, Digest, Sequential>::init(
             context.child("mmr_serial"),
diff --git a/storage/src/merkle/batch.rs b/storage/src/merkle/batch.rs
index a49bdd61b..aacee3d66 100644
--- a/storage/src/merkle/batch.rs
+++ b/storage/src/merkle/batch.rs
@@ -88,6 +88,8 @@ use alloc::{
     sync::{Arc, Weak},
     vec::Vec,
 };
+#[cfg(feature = "std")]
+use commonware_codec::Write;
 use commonware_cryptography::Digest;
 use commonware_parallel::{Sequential, Strategy};
 use core::ops::Range;
@@ -267,6 +269,27 @@ impl<F: Family, D: Digest, S: Strategy> UnmerkleizedBatch<F, D, S> {
         self.add_leaf_digest(digest)
     }
 
+    /// Encode and hash `items` across the strategy, adding their leaf digests in order.
+    #[cfg(feature = "std")]
+    pub(crate) fn add_many<Item: Write + Send + Sync>(
+        self,
+        hasher: &impl Hasher<F, Digest = D>,
+        items: &[Item],
+    ) -> Self {
+        let first = self.leaves();
+        let digests = self.strategy().map_init_collect_vec(
+            items.iter().enumerate(),
+            Vec::new,
+            |buf, (i, item)| {
+                let pos = Position::try_from(first + i as u64).expect("valid leaf location");
+                buf.clear();
+                item.write(buf);
+                hasher.leaf_digest(pos, buf.as_slice())
+            },
+        );
+        self.add_leaf_digests(digests)
+    }
+
     /// Validate that `loc` refers to an in-bounds, non-pruned leaf and return its position.
     ///
     /// # Errors
diff --git a/storage/src/merkle/persisted/full.rs b/storage/src/merkle/persisted/full.rs
index 29af42cea..ada4e3e67 100644
--- a/storage/src/merkle/persisted/full.rs
+++ b/storage/src/merkle/persisted/full.rs
@@ -22,7 +22,7 @@ use crate::{
     },
     metadata::{Config as MConfig, Metadata},
 };
-use commonware_codec::DecodeExt;
+use commonware_codec::{DecodeExt, Write};
 use commonware_cryptography::Digest;
 use commonware_parallel::Strategy;
 use commonware_runtime::{buffer::paged::CacheRef, Clock, Metrics, Storage as RStorage};
@@ -60,10 +60,14 @@ impl<F: Family, D: Digest, S: Strategy> UnmerkleizedBatch<F, D, S> {
         }
     }
 
-    /// Add a run of pre-computed leaf digests, in order.
-    pub(crate) fn add_leaf_digests(self, digests: impl IntoIterator<Item = D>) -> Self {
+    /// Encode and hash `items` across the strategy, adding their leaf digests in order.
+    pub(crate) fn add_many<Item: Write + Send + Sync>(
+        self,
+        hasher: &impl Hasher<F, Digest = D>,
+        items: &[Item],
+    ) -> Self {
         Self {
-            inner: self.inner.add_leaf_digests(digests),
+            inner: self.inner.add_many(hasher, items),
         }
     }
 
diff --git a/storage/src/qmdb/current/batch.rs b/storage/src/qmdb/current/batch.rs
index d5bfabed4..3ef6d22ae 100644
--- a/storage/src/qmdb/current/batch.rs
+++ b/storage/src/qmdb/current/batch.rs
@@ -10,7 +10,6 @@ use crate::{
         storage::Storage as MerkleStorage, Graftable, Location, Position, Readable,
     },
     qmdb::{
-        self,
         any::{
             self,
             batch::{DiffCursors, DiffEntry},
@@ -670,8 +669,7 @@ where
                 grafted_batch = grafted_batch.add_leaf_digest(digest);
             }
         }
-        let grafted_hasher =
-            grafting::GraftedHasher::<F, _>::new(qmdb::hasher::<H>(), grafting_height);
+        let grafted_hasher = grafting::hasher::<F, H>(grafting_height);
         grafted_batch.merkleize(&current_db.grafted_tree, &grafted_hasher)
     };
 
diff --git a/storage/src/qmdb/current/db.rs b/storage/src/qmdb/current/db.rs
index 9dfe06629..598eee154 100644
--- a/storage/src/qmdb/current/db.rs
+++ b/storage/src/qmdb/current/db.rs
@@ -1130,8 +1130,7 @@ pub(super) async fn build_grafted_tree<
         let batch = {
             let batch = grafted_tree.new_batch_with_strategy(strategy.clone());
             let batch = batch.add_leaf_digests(leaves.iter().map(|&(_, digest)| digest));
-            let grafted_hasher =
-                grafting::GraftedHasher::<F, _>::new(qmdb::hasher::<H>(), grafting_height);
+            let grafted_hasher = grafting::hasher::<F, H>(grafting_height);
             batch.merkleize(&grafted_tree, &grafted_hasher)
         };
         grafted_tree.apply_batch(&batch)?;
diff --git a/storage/src/qmdb/current/grafting.rs b/storage/src/qmdb/current/grafting.rs
index 7919d645e..49be2d4b8 100644
--- a/storage/src/qmdb/current/grafting.rs
+++ b/storage/src/qmdb/current/grafting.rs
@@ -77,6 +77,13 @@ pub const fn height<const N: usize>() -> u32 {
     BitMap::<N>::CHUNK_SIZE_BITS.trailing_zeros()
 }
 
+/// Return a [GraftedHasher] over QMDB's fixed Merkle hasher configuration.
+pub(super) const fn hasher<F: Graftable, H: Hasher>(
+    grafting_height: u32,
+) -> GraftedHasher<F, merkle::hasher::Standard<H>> {
+    GraftedHasher::new(qmdb::hasher(), grafting_height)
+}
+
 /// Return the number of bitmap chunks that have a corresponding height-G ancestor in the ops tree.
 ///
 /// - For MMR, this is always the same as the number of complete_chunks.
@@ -355,7 +362,6 @@ pub(super) struct Storage<
     grafting_height: u32,
     ops_tree: &'a S,
     grafted_hasher: GraftedHasher<F, merkle::hasher::Standard<H>>,
-    _phantom: PhantomData<F>,
 }
 
 impl<
@@ -372,8 +378,7 @@ impl<
             grafted_tree,
             grafting_height,
             ops_tree,
-            grafted_hasher: GraftedHasher::new(qmdb::hasher::<H>(), grafting_height),
-            _phantom: PhantomData,
+            grafted_hasher: hasher::<F, H>(grafting_height),
         }
     }
 
diff --git a/storage/src/qmdb/current/mod.rs b/storage/src/qmdb/current/mod.rs
index c8e3e71f0..bb12339d8 100644
--- a/storage/src/qmdb/current/mod.rs
+++ b/storage/src/qmdb/current/mod.rs
@@ -516,6 +516,7 @@ pub mod tests {
                 traits::{DbAny, MerkleizedBatch as _, UnmerkleizedBatch as _},
             },
             store::tests::{TestKey, TestValue},
+            verify_proof,
         },
         translator::Translator,
     };
@@ -3869,9 +3870,6 @@ pub mod tests {
     /// Regression: `ops_historical_proof` must verify with QMDB's ops-tree hasher configuration.
     #[test_traced("INFO")]
     fn test_current_mmb_ops_historical_proof_verifies_with_backward_bagging() {
-        use crate::qmdb::verify_proof;
-        use commonware_utils::NZU64;
-
             let ctx = context.child("db");
diff --git a/storage/src/qmdb/current/sync/mod.rs b/storage/src/qmdb/current/sync/mod.rs
index d4d220abb..6d504c5c8 100644
--- a/storage/src/qmdb/current/sync/mod.rs
+++ b/storage/src/qmdb/current/sync/mod.rs
@@ -197,28 +197,16 @@ where
         &any.log.merkle,
     );
     let partial = db::partial_chunk(any.bitmap.as_ref());
-    let grafted_root = db::compute_grafted_root::<F, H, _, _, N>(
+    let ops_root = any.root();
+    let root = db::compute_db_root::<F, H, _, _, N>(
         any.bitmap.as_ref(),
         &storage,
         ops_leaves,
+        partial,
         any.inactivity_floor_loc,
+        &ops_root,
     )
     .await?;
-    let ops_root = any.root();
-    let hasher = qmdb::hasher::<H>();
-    let partial_digest = partial.map(|(chunk, next_bit)| {
-        let digest = hasher.digest(chunk.as_slice());
-        (next_bit, digest)
-    });
-    let pending_digest =
-        db::pending_chunk::<F, _, N>(any.bitmap.as_ref(), ops_leaves, grafting::height::<N>())?
-            .map(|chunk| hasher.digest(chunk.as_slice()));
-    let root = db::combine_roots::<H>(
-        &ops_root,
-        &grafted_root,
-        pending_digest.as_ref(),
-        partial_digest.as_ref().map(|(nb, d)| (*nb, d)),
-    );
 
     // Initialize metadata store and construct the Db.
     let (metadata, _, _) =
```

storage/src/journal/authenticated.rs
roberto-bayardo: hmm, do we have some way of ensuring we exercise both arms in tests?

roberto-bayardo: `.manual()` turns out to be exactly this: a Manual-wrapped strategy skips the adaptive policy entirely, so its map/fold operations always run across the pool (and `Sequential` is always serial). Test updated to `Rayon::new(2).manual()` — both hashing paths now covered by contract rather than by policy timing.

storage/src/journal/authenticated.rs
roberto-bayardo: The pool never touches the deterministic runtime — it executes a pure, order-preserving hashing closure synchronously, so results stay deterministic (`keyless/fixed.rs` has the same pattern today). If we would rather ban it as a rule, the missing piece is a test-only strategy in `commonware-parallel` that executes the parallel path inline — happy to do that as a follow-up and migrate both call sites.

roberto-bayardo:
Adopted the suggested refactor in a85ffde5f — nice consolidation, and it composes with the test fix in 2b4c76cb6: with `add_many`, strategy adaptivity lives inside the map itself, and the `Manual`-wrapped strategy exercises the parallel hashing deterministically. One deviation: kept the `Manual`-based test comment rather than the diff's policy-sampling justification, since that justification relies on the same autotuner internals you flagged in the line-1248 thread. Also verified the `std`-gating under `--no-default-features`.

storage/src/journal/authenticated.rs
patrick-ogrady: Ah, that's an interesting way to force that test.

patrick-ogrady (approved):

storage/src/journal/authenticated.rs
patrick-ogrady: I'm not sure what the right fix is here but don't feel it is worth blocking this on.

PR #4154, opened 2026-07-01
[cryptography] Add SHA256 `hash_pair` fast path

clabby:
## Overview

Adds an in-core pair-hashing path for SHA-256.

Previously, merkleization hashed each dirty internal node independently. That leaves throughput on the table because modern SHA extensions can make progress on two independent messages with better instruction-level parallelism, without paying thread scheduling or collection overhead.

The motivating workload is internal Merkle hashing: position || left || right, which is always 8 + 32 + 32 = 72 bytes for SHA-256 digests. As such, each implementation contains a fast-path for 72 byte inputs.

### Benchmarks

Pair hashing beats serial hashing at every measured size on both platforms.

Screenshot 2026-07-01 at 3 02 37 PM
Screenshot 2026-07-01 at 2 29 45 PM

> visualizations generated w/ https://github.com/modem-dev/sideshow


clabby:
superseded by #4187 to adapt to the new `Hasher` interface.

PR #4151, opened 2026-07-01
[storage] Add `start_sync` to Segmented Journal and Prunable Archive

patrick-ogrady:

PR #4157, opened 2026-07-02
[marshal] Defer `fsync`

patrick-ogrady:
Replaces: #4072

Defer block-durability fsyncs so they overlap consensus voting instead of blocking the propose/verify hot paths. `certify` becomes the durability barrier: a validator still only casts a finalize vote once the block is durably persisted locally, but the fsync now runs concurrently with the notarization round instead of serializing ahead of it.

## Marshal

- Persist candidate blocks and notarizations with `put_start_sync` (#4151), returning a `Handle` that observes the flush. `Mailbox::proposed_deferred`/`verified_deferred` enqueue and return the sync-handle receiver without awaiting it; the blocking `proposed`/`verified`/`certified` forms await durability on the caller's task, so the actor never blocks on fsync.
- `CertificationGates` (generalizing `VerificationTasks`) tracks an in-flight gate per `(round, digest)`. Propose and boundary re-proposal publish the digest to consensus immediately and resolve the gate once the sync handle completes (`persist_and_defer`). Verify runs the candidate store concurrently with application verification and resolves the gate only when the app verdict is true AND the store is durable (`gate_verdict`); the caches hold structurally-valid candidates for availability, not app-validity claims. `certify` consumes the gate before the finalize vote and falls back to fetch-and-persist when no gate exists (e.g. after an unclean restart).
- Storage sync failures are fatal (`Durable` in `core/durability.rs`): a flush failure panics rather than resolving to a recoverable verdict; only runtime shutdown reports "not durable". Handles no consensus caller awaits (the notarization path) are observed by a pool in the actor loop under the same policy.
- Skip re-persisting a certified block that this lifetime's verified archive already holds (`verified_writes`): the certify ack instead carries a covering handle from `start_sync_verified`, joined with the round's notarization-archive sync. The map prunes on the same boundary as the archives and is empty after restart, where the write goes through as usual.

## Simplex

- Remove the leader's `is_local` certify shortcut (`slot::Status::Verified(bool)` loses its payload). With deferred propose persistence, `automaton.certify` is the only durability barrier before a finalize vote, so the leader must certify its own proposal through the automaton like any other.

## Conformance (`breaking-format`)

- Storage-conformance hashes change because certified blocks covered by a verified write are no longer duplicated into the notarized archive.


consensus/src/marshal/core/durability.rs
```diff
@@ -0,0 +1,71 @@
...
+        Err(Error::Closed | Error::Aborted) => {
+            debug!(name, "runtime shutdown before sync completed");
+            false
+        }
+        Err(e) => panic!("failed to sync {name}: {e}"),
+    }
+}
+
+/// Awaits a durable-sync handle, applying marshal's fatal policy via [`observe_sync`].
+pub(crate) async fn await_durable(handle: Handle<()>, name: &str) -> bool {
```
patrick-ogrady: I don't love this helper...

consensus/src/marshal/core/actor.rs
```diff
@@ -595,66 +618,99 @@ where
...
+
+                // Hold the certify barrier until the round's notarization
+                // certificate (when one was accepted before this message) is
+                // durable alongside the block.
+                let notarization_sync = self.cache.start_sync_notarizations(round).await;
+                let handle = Handle::from_future(async move {
+                    let (notarization, block) = join(notarization_sync, block_sync).await;
+                    notarization.and(block)
+                });
+                ack.expect("durable ack present").send_lossy(handle);
```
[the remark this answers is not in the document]
patrick-ogrady: When a round is below the retention floor, both branches are equally vacuous: `put_notarized` returns a ready-Ok handle via `AlreadyPrunedTo`, exactly like `start_sync_verified` for a pruned epoch (and the pre-change code acked `()` unconditionally here). The skip does not change the outcome. The map and the archives also prune on the same boundary (`verified_writes.split_off(&prune_round)` vs `prune(min_view)`, both retain `>=`), so there is no window where the map vouches for a write the archive dropped while the round is still retained. Certify resolving true below the retention floor is pre-existing, intended semantics: the round is no longer required for consensus to make progress.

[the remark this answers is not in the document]
patrick-ogrady: The barrier is deliberately scoped to notarizations accepted before the `Certified` message (per the comment above it). The mailbox is FIFO, so any `Notarization` enqueued before `Certified` has been processed by the time the snapshot is taken and its write is covered by the barrier. A notarization arriving after `Certified` is not covered, but the old code never covered that case either: `put_notarization` was an independent blocking write on the `Notarization` path, never folded into the `Certified` ack. Relative to `main`, the barrier is strictly stronger. The durability requirement behind certify is about the block (the `>= f+1` recoverability guarantee); a certificate lost to a crash remains recoverable from any quorum peer.

consensus/src/marshal/core/durability.rs
```diff
@@ -0,0 +1,69 @@
...
+use tracing::debug;
+
+/// Applies marshal's fatal policy when awaiting a durable-sync [`Handle`].
+pub(crate) trait Durable {
+    /// Resolves `true` once the sync is durable. A real sync failure panics (annotated
+    /// with `name`) rather than resolving: converting it into a `false` verdict would let
+    /// consensus treat lost local state as a live rejection. Resolves `false` only when
+    /// the runtime is shutting down (the handle was closed or aborted before the sync
+    /// resolved), so the caller reports "not durable" and abandons the work.
+    fn durable(self, name: &'static str) -> impl Future<Output = bool> + Send;
```
patrick-ogrady: I'm not sure we need the `name` here but 🤷

consensus/src/marshal/standard/deferred.rs
```diff
@@ -252,30 +255,46 @@ where
...
+                        None => return,
+                    };
+
+                // Run application verification and the candidate store concurrently.
+                // The block has passed structural ancestry checks, but may still fail
+                // application verification. Storing it now is intentional: these caches
+                // provide candidate availability/recovery, not an app-validity decision.
+                // This task gates the finalize vote by resolving true only after both
+                // app verification succeeds and the store is durable.
+                let store = stage.store(&marshal, round, block.clone());
```
patrick-ogrady, line 266: we could start this before the parent check? in fact, that's probably a good idea.

consensus/src/marshal/standard/deferred.rs
```diff
@@ -505,9 +494,9 @@ where
                 // for this round (persisted by a pre-crash propose whose
                 // notarize vote never reached the journal).
                 //
-                // Building a fresh block would land on the same prunable archive
-                // index and be silently dropped, so the stored block is the only proposal
-                // we can broadcast for this round.
```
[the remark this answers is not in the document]
patrick-ogrady: Refuted as a live bug: the precondition is unreachable for an honest node, and first-stored is the correct selection at this site.

1. Multi-occupancy at a round this node leads would require two validly-signed proposals for that view. Only the leader can sign them, and this guard is exactly what prevents a restarted leader from building a second block (it reuses on context match or skips). Absent prior self-equivocation, the verified archive holds at most one candidate at any view we lead.

2. There is no digest to check against here. Simplex hands `propose` only the recovered context. If replay had recovered any payload binding for this view (a journaled notarize vote or a notarization), the voter would not call `propose` at all, so the verified archive is the only digest-pinning state available.

3. First-stored is the anti-equivocation choice regardless: the earliest accepted candidate is the one whose digest may already have been broadcast before the crash, so re-committing to it (after the embedded-context equality check) is the safe direction. Any later same-view candidate is precisely what must not be proposed.

A "skip when multiple candidates exist" check would add archive API for a state reachable only after the node has already equivocated, which we deliberately avoid (lean on the signing invariant). `get_verified`'s docs pin the first-stored semantics so future callers do not assume "most recently verified."

consensus/src/marshal/application/validation.rs
```diff
@@ -9,22 +9,31 @@ use crate::{
...
+    /// Store in `notarized_blocks`.
     Certified,
 }
 
 impl Stage {
-    /// Store `block` in the marshal cache for the provided stage.
+    /// Store `block` in the marshal cache for the provided stage, returning once it
+    /// is durably persisted (`false` if marshal could not accept it).
+    ///
+    /// Callers that run this concurrently with application verification must
```
patrick-ogrady: This isn't necessary/is slop.

consensus/src/marshal/coding/marshaled.rs
```diff
@@ -173,7 +173,7 @@ where
     scheme_provider: Z,
     epocher: ES,
     strategy: S,
-    verification_tasks: VerificationTasks<Commitment>,
+    certification_gates: CertificationGates<Commitment>,
```
patrick-ogrady: just call this `gates: Gates<Commitment>`? We can change `certification_gates` to be `gates` as the import path and `CertificationGates to `Gates`.

We can also call this `drive` and `handle`: drive_certify_gate, gate_verdict (the path of gate is already implied?

consensus/src/marshal/coding/marshaled.rs
```diff
@@ -416,6 +416,14 @@ where
...
                 }
 
+                // Run application verification and the candidate store concurrently.
+                // The block has passed coded structural checks, but may still fail
+                // application verification. Storing it now is intentional: these caches
+                // provide candidate availability/recovery, not an app-validity decision.
+                // This task gates the finalize vote by resolving true only after both
+                // app verification succeeds and the store is durable.
+                let store = stage.store(&marshal, round, block.clone());
+
```
patrick-ogrady: can remove \n

consensus/src/marshal/core/actor.rs
```diff
@@ -595,66 +617,99 @@ where
...
                     .await;
+                syncs.push(handle.durable("notarization"));
 
                 // A notarization alone is not enough to fetch missing proposal
                 // data. If the block is not locally available, remember the
                 // certificate and wait for a later finalization/repair path.
                 if let Some(block) = self.find_block_by_commitment(buffer, commitment).await {
                     self.ingest(&block, buffer, application, resolver).await;
-                    self.cache.put_block(round, digest, block.into()).await;
+                    if self.verified_writes.get(&round) == Some(&digest) {
```
patrick-ogrady: With the new has_at, I think we can remove this memory layer and just use archive for this check?

consensus/src/marshal/core/durability.rs
patrick-ogrady: This ended up being more useful than expected for logging standardization.

consensus/src/marshal/standard/inline.rs
```diff
@@ -143,7 +142,7 @@ where
     application: A,
     marshal: Mailbox<S, Standard<B>>,
     epocher: ES,
-    available_blocks: AvailableBlocks<B::Digest>,
+    certification_gates: CertificationGates<B::Digest>,
```
patrick-ogrady: another example of just using gates: Gates

consensus/src/marshal/standard/deferred.rs
patrick-ogrady: We should remember to do this...fetching parent can take some time?

consensus/src/marshal/standard/deferred.rs
```diff
@@ -1083,7 +1067,7 @@ mod tests {
             let verify_result = marshaled
                 .verify(unsupported_context, block_commitment)
                 .await;
-            // Wait for optimistic verify to complete so the verification task is registered
+            // Wait for optimistic verify to complete so the certification gate task is registered
```
patrick-ogrady, line 1079: add a \n before this? don't cut code blocks?

patrick-ogrady (commented):
final nits

PR #4158, opened 2026-07-02
[runtime] Map only NotFound to missing errors in remove

danlaine:
## Summary

`Storage::remove` on the tokio and iouring backends mapped every removal failure to `BlobMissing`/`PartitionMissing`, so a real I/O error (permission denied, EIO) was misreported as a missing target. Only `ErrorKind::NotFound` now maps to the missing variants; other errors surface as `Error::Io`. This matches the memory backend, which already returns `BlobMissing` only when the entry is truly absent.

Beyond fixing the misdiagnosis, this makes `BlobMissing` a precise "already removed" signal. A later PR in this series makes the segmented journal's removal paths tolerate `BlobMissing` on retry (a removal dropped after the unlink but before untracking must succeed when retried), which is only safe once the error cannot mask real failures.

### DIFF
--- a/runtime/src/storage/iouring.rs
+++ b/runtime/src/storage/iouring.rs
@@ -195,13 +195,18 @@ impl crate::Storage for Storage {
         let path = self.storage_directory.join(partition);
         if let Some(name) = name {
             let blob_path = path.join(hex(name));
-            fs::remove_file(blob_path)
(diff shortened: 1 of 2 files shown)


danlaine:
Closing in favor of https://github.com/commonwarexyz/monorepo/pull/4162

PR #4159, opened 2026-07-02
[cryptography/ed25519] Defer Hashing in Batch

patrick-ogrady:
## Summary

- Defers the per-signature SHA-512 challenge hash from `Batch::add` to `Batch::verify`. `add` just copies the framed payload (retained until `verify`), and every signature's hash is computed inside the verification fold under the caller's `Strategy` instead of serially at queue time.
- Flattens the `Verifier` from a queue-time `Map<VerificationKey, Vec<(Scalar, Signature)>>` to an insertion-order `Vec<(VerificationKey, Vec<u8>, Signature)>`. `add` becomes a plain push, and shards are balanced by signature count rather than key count, so single-signer batches parallelize (previously one key = one group = one serial shard).
- Routes `verify` through `strategy.run(total, serial, parallel)`. The adaptive policy picks between verifying the whole batch as one equation (serial, with global coalescing) and sharded parallel verification. Both paths share a `verify_shard` helper.
- Partitions the parallel path so signatures sharing a verification key land in the same shard: a counting sort on each key's first byte builds a permutation in O(n) with no comparisons, and shards are equal-size chunks of it. Coalescing no longer depends on queue order (e.g. `verify_certificates` queues the same validator set certificate-major, which previously paid up to the all-distinct rate). Skewed batches (like a single signer) still split evenly across shards, and keys crafted to share a first byte only forfeit the grouping.
- Coalesces duplicate keys within a shard via a pre-sized `hashbrown::HashMap` with `ahash::RandomState` on plain `no-rng` (`AHashMap` is std-gated in ahash 0.8, and fixed-seed hashing matches the workspace's existing assumption).
- Replaces `BatchVerifier::new()` + `with_capacity` with a single `new(capacity: usize)` hint. Call sites pass real sizes.
- Adds tests for deferred hashing, framing pinned to `union_unique`, and interleaved duplicate keys (sequential and parallel results asserted to agree).

## Why

For large batches of mostly-distinct keys (the certificate/vote shape), the old `add` loop was a serial bottleneck: at 32k signatures it spent ~7-8ms hashing and building the per-key map before `verify` could fan out. Deferring the hash moves that work under the strategy, the flat layout removes the remaining serial map/alloc costs, and the partition makes coalescing a property of the batch rather than of the order callers happened to queue it in.

The trade-off is that `VerificationKey` (192 bytes) and the framed payload are stored per queued signature until `verify` (~6MB of transient memory for a pathological 32k single-signer batch), in exchange for balanced shards that verify that same batch 5.8x faster.

## Benchmarks

E2E = `add` loop + `verify`, 32-byte messages with namespace framing, `Rayon(8)` (or `Sequential`), min of 12 iterations, Apple M5 Pro. All times in ms.

Deferral + flat layout, vs merge-base (a5915db4f), measured before the partition:

| scenario | main add | main E2E | PR add | PR E2E | E2E delta |
|---|---|---|---|---|---|
| 16k distinct keys, conc=8 | 3.56 | 20.91 | 0.55 | 18.15 | -13% |
| 32k distinct keys, conc=8 | 7.34 | 39.99 | 1.07 | 35.21 | -12% |
| 32k single signer, conc=8 | 8.55 | 129.43 | 0.87 | 22.36 | -83% (5.8x) |
| 64 signers x 512 sigs, conc=8 | 8.71 | 28.00 | 1.04 | 22.28 | -20% |
| 1k distinct keys, conc=8 | 0.23 | 1.97 | 0.02 | 1.63 | -17% |
| 32k distinct keys, serial | 7.48 | 210.84 | 0.90 | 211.23 | ~0% |

Key partition, before/after on this branch (bracketed run):

| scenario | before | after |
|---|---|---|
| 4096 signers x 8 sigs, round-robin interleaved | 28.5 | 20.2 (-29%) |
| 50 certificates x 640 signers, certificate-major | 19.6 | 18.1 (-7%) |
| 4096 x 8, caller pre-grouped | 18.8 | 19.0 |
| 32k distinct keys | 27.2 | 28.5 |
| 32k single signer | 17.4 | 17.6 |
| 32k distinct keys, serial | 41.2 | 41.4 |

The interleaved and certificate-major improvements reproduce across every run. The remaining deltas are within the build-to-build layout and load variance observed on this machine (the partition's arithmetic is ~0.2ms, and the residual vs pre-grouped order is shard-side index gather, which any runtime reordering pays).

Hasher choice for the per-shard grouping map (pre-sized, 32-byte keys, entry-or-insert micro): SipHash 10.6 ns/op, foldhash 6.4, ahash 3.6. The map is ~0.1% of E2E either way.

https://claude.ai/code/session_019M4zv2iQQWuqJAMHssUBMB

### DIFF
--- a/cryptography/Cargo.toml
+++ b/cryptography/Cargo.toml
@@ -14,6 +14,7 @@ documentation = "https://docs.rs/commonware-cryptography"
 workspace = true
 
 [dependencies]
+ahash = { workspace = true, features = ["no-rng"] }
 anyhow = { workspace = true, optional = true }
 arbitrary = { workspace = true, optional = true, features = ["derive"] }
 blake3 = { workspace = true, features = ["zeroize"] }
@@ -31,6 +32,7 @@ ctutils.workspace = true
 curve25519-dalek.workspace = true
 ecdsa.workspace = true
 fixedbitset.workspace = true
+hashbrown.workspace = true
 num-rational = { workspace = true, optional = true }
 num-traits = { workspace = true, optional = true }
 once_cell = { workspace = true, features = ["alloc", "race"] }
@@ -88,6 +90,8 @@ arbitrary = [
 ]
 fuzz = [ "arbitrary", "std" ]
 std = [
+	"ahash/runtime-rng",
+	"ahash/std",
 	"anyhow?/std",
 	"blake3/std",
 	"bytes/std",
--- a/cryptography/Cargo.toml
+++ b/cryptography/Cargo.toml
@@ -90,6 +90,7 @@ arbitrary = [
 ]
 fuzz = [ "arbitrary", "std" ]
 std = [
+	"ahash/std",
 	"anyhow?/std",
 	"blake3/std",
 	"bytes/std",
--- a/cryptography/Cargo.toml
+++ b/cryptography/Cargo.toml
@@ -90,8 +90,6 @@ arbitrary = [
 ]
 fuzz = [ "arbitrary", "std" ]
 std = [
-	"ahash/runtime-rng",
-	"ahash/std",
 	"anyhow?/std",
 	"blake3/std",
 	"bytes/std",
--- a/cryptography/src/ed25519/core/batch.rs
+++ b/cryptography/src/ed25519/core/batch.rs
@@ -40,6 +40,7 @@ use crate::transcript::{Summary, Transcript};
 use alloc::{collections::BTreeMap as Map, vec::Vec};
 use commonware_math::algebra::Random;
 use commonware_parallel::Strategy;
+use commonware_utils::union_unique;
 use core::iter::once;
 use curve25519_dalek::{
     constants::ED25519_BASEPOINT_POINT as B,
@@ -63,36 +64,14 @@ fn gen_u128<R: RngCore + CryptoRng>(mut rng: R) -> u128 {
     u128::from_le_bytes(bytes)
 }
 
-/// A batch verification item.
-///
-/// This struct exists to allow batch processing to be decoupled from the
-/// lifetime of the message. This is useful when using the batch verification API
-/// in an async context.
-#[derive(Clone, Debug)]
-pub struct Item {
-    vk: VerificationKey,
-    sig: Signature,
-    k: Scalar,
-}
-
-impl<'msg, M: AsRef<[u8]> + ?Sized> From<(VerificationKey, Signature, &'msg M)> for Item {
-    fn from(tup: (VerificationKey, Signature, &'msg M)) -> Self {
-        let (vk, sig, msg) = tup;
-        let k = super::scalar_from_hash(
-            Sha512::default()
-                .chain(&sig.R_bytes[..])
-                .chain(vk.as_bytes())
-                .chain(msg),
-        );
-        Self { vk, sig, k }
-    }
-}
-
 /// A batch verification context.
 #[derive(Default)]
 pub struct Verifier {
-    /// Signature data queued for verification.
-    signatures: Map<VerificationKey, Vec<(Scalar, Signature)>>,
+    /// Signature data queued for verification. Payloads are copied (instead
+    /// of hashed at queue time) so the SHA-512 challenge computation for
+    /// every signature is deferred to [`Verifier::verify`], where it runs
+    /// under the caller's [`Strategy`] instead of serially at queue time.
+    signatures: Map<VerificationKey, Vec<(Vec<u8>, Signature)>>,
     /// Caching this count avoids a map traversal to figure out
     /// how much to preallocate.
     batch_size: usize,
@@ -104,16 +83,26 @@ impl Verifier {
         Self::default()
     }
 
-    /// Queue a `(key, signature, message)` tuple for verification.
-    pub fn queue<I: Into<Item>>(&mut self, item: I) {
-        let Item { vk, sig, k } = item.into();
+    /// Queue a `(key, signature)` pair for verification of `message` under
+    /// `namespace`.
+    pub fn queue(
+        &mut self,
+        vk: VerificationKey,
+        sig: Signature,
+        namespace: Option<&[u8]>,
+        message: &[u8],
+    ) {
+        let payload = namespace.map_or_else(
+            || message.to_vec(),
+            |namespace| union_unique(namespace, message),
+        );
 
         self.signatures
             .entry(vk)
             // The common case is 1 signature per public key.
             // We could also consider using a smallvec here.
             .or_insert_with(|| Vec::with_capacity(1))
-            .push((k, sig));
+            .push((payload, sig));
         self.batch_size += 1;
     }
 
@@ -160,7 +149,9 @@ impl Verifier {
                 // - A_i is the verification key;
                 // - R_i is the signature's R value;
                 // - s_i is the signature's s value;
-                // - k_i is the hash of the message and other data;
+                // - k_i is the hash of the message and other data, computed
+                //   here so the per-signature SHA-512 work runs under the
+                //   caller's strategy;
                 // - z_i is a random 128-bit Scalar.
                 //
                 // Normally n signatures would require a multiscalar multiplication of
@@ -190,7 +181,13 @@ impl Verifier {
                     let A = -vk.minus_A;
                     let mut A_coeff = Scalar::ZERO;
 
-                    for (k, sig) in sigs.iter() {
+                    for (payload, sig) in sigs.iter() {
+                        let k = super::scalar_from_hash(
+                            Sha512::default()
+                                .chain(&sig.R_bytes[..])
+                                .chain(vk.as_bytes())
+                                .chain(payload),
+                        );
                         let R = CompressedEdwardsY(sig.R_bytes)
                             .decompress()
                             .ok_or(Error::InvalidSignature)?;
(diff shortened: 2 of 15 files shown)


cryptography/src/ed25519/core/batch.rs
```diff
@@ -177,35 +158,38 @@ impl Verifier {
...
-                let m = shard.len();
-                let batch_size = shard.iter().map(|(_, sigs)| sigs.len()).sum();
-
-                let mut A_coeffs = Vec::with_capacity(m);
-                let mut As = Vec::with_capacity(m);
-                let mut R_coeffs = Vec::with_capacity(batch_size);
-                let mut Rs = Vec::with_capacity(batch_size);
+                let n = shard.len();
+                let mut key_indices: HashMap<&VerificationKeyBytes, usize, RandomState> =
+                    HashMap::with_capacity_and_hasher(n, RandomState::default());
```
patrick-ogrady: We should use AHashMap here?

cryptography/Cargo.toml
```diff
@@ -31,6 +32,7 @@ ctutils.workspace = true
 curve25519-dalek.workspace = true
 ecdsa.workspace = true
 fixedbitset.workspace = true
+hashbrown.workspace = true
```
patrick-ogrady, line 35: I don't think we need this anymore?

cryptography/src/ed25519/core/batch.rs
patrick-ogrady: `AHashMap` is gated on ahash's `std` feature (it wraps `std::collections::HashMap`), and CI builds this crate no_std (`check_no_std.sh` compiles `commonware-cryptography --no-default-features` for `riscv32imac-unknown-none-elf`). Pairing hashbrown's map with `ahash::RandomState` is the no_std spelling of the same map: `AHashMap::with_capacity(n)` is literally `HashMap::with_capacity_and_hasher(n, RandomState::new())` in ahash, and this matches the existing pattern in `utils/src/ordered.rs`. Added a comment in the code recording the constraint.

cryptography/Cargo.toml
patrick-ogrady: Still needed: hashbrown is the map used for the per-shard grouping in `verify` (`ahash::AHashMap` cannot replace it without breaking the no_std build, since `AHashMap` is std-gated; see the other thread).

cryptography/src/lib.rs
```diff
@@ -158,6 +158,18 @@ commonware_macros::stability_scope!(BETA {
...
 
+        /// Create a new batch verifier with capacity for at least `capacity` items.
+        ///
+        /// The capacity is a hint: more than `capacity` items may be added, and
+        /// implementations may ignore it.
+        fn with_capacity(capacity: usize) -> Self
+        where
+            Self: Sized,
+        {
+            let _ = capacity;
```
clabby: nit: 
```suggestion
        fn with_capacity(_capacity: usize) -> Self
        where
            Self: Sized,
        {
```

cryptography/src/ed25519/core/batch.rs
```diff
@@ -177,35 +166,41 @@ impl Verifier {
...
 
-                let m = shard.len();
-                let batch_size = shard.iter().map(|(_, sigs)| sigs.len()).sum();
-
-                let mut A_coeffs = Vec::with_capacity(m);
-                let mut As = Vec::with_capacity(m);
-                let mut R_coeffs = Vec::with_capacity(batch_size);
-                let mut Rs = Vec::with_capacity(batch_size);
+                // Group the shard's signatures by verification key (ahash's
+                // AHashMap requires std, so use hashbrown's map with the ahash
```
patrick-ogrady: This is wrong -> using no_RNG works

cryptography/src/lib.rs
patrick-ogrady: Just do _capacity?

cryptography/src/ed25519/core/batch.rs
```diff
@@ -104,17 +80,29 @@ impl Verifier {
...
+    /// Queue a `(key, signature)` pair for verification of `message` under
+    /// `namespace`.
+    pub fn queue(
+        &mut self,
+        vk: VerificationKey,
+        sig: Signature,
+        namespace: Option<&[u8]>,
+        message: &[u8],
+    ) {
+        let payload = namespace.map_or_else(
```
patrick-ogrady, line 96: Should we have kept CoW here?

cryptography/src/ed25519/core/batch.rs
```diff
@@ -136,11 +124,10 @@ impl Verifier {
...
         // from the provided RNG, to compute a random scalar for each signature in the shard.
         let manual = strategy.manual();
-        let groups: Vec<_> = self.signatures.into_iter().collect();
-        let shard_count = manual.parallelism_hint().min(groups.len().max(1));
-        let shard_size = groups.len().div_ceil(shard_count).max(1);
+        let shard_count = manual.parallelism_hint().min(self.signatures.len().max(1));
+        let shard_size = self.signatures.len().div_ceil(shard_count).max(1);
         let mut shards = Vec::with_capacity(shard_count);
-        for shard in groups.chunks(shard_size) {
+        for shard in self.signatures.chunks(shard_size) {
```
clabby: I think we should move towards sorting this for the user. Given that we currently preserve insertion order, we might end up producing shards that are not optimal for collections of signatures with sparse same-signer items.

i.e. `[A, B, A, B, A, B, A, B]` sharded across two threads should be `[ [A, A, A, A], [B, B, B, B] ]`.

cryptography/src/ed25519/core/batch.rs
clabby: This isn't super important, but on `no_std`, `ahash` isn't DoS resistant. See #3 on https://docs.rs/ahash/latest/ahash/random_state/trait.RandomSource.html. ASLR notably isn't enabled on wasm32.

If we wanted this to be a bit more robust, we could seed the `RandomState` with the transcript.

cryptography/src/ed25519/core/batch.rs
```diff
@@ -104,17 +80,29 @@ impl Verifier {
...
+
+    /// Queue a `(key, signature)` pair for verification of `message` under
+    /// `namespace`.
+    pub fn queue(
+        &mut self,
+        vk: VerificationKey,
+        sig: Signature,
+        namespace: Option<&[u8]>,
+        message: &[u8],
+    ) {
```
clabby, line 95: I think it's worth calling out that this function now heap-allocates `namespace || message`, too, and that this memory is retained until `verify`? Previously, `add` would hash the slice and the caller could drop it right away. This has a much larger transient memory requirement (which is OK, just may be a bit surprising without a callout).

clabby: It may be worth considering an alternative API shape where we just add an `add_many` function that computes many hashes in parallel and keeps the old structure of this module.

cryptography/src/ed25519/scheme.rs
```diff
@@ -331,6 +331,12 @@ impl BatchVerifier for Batch {
         }
     }
 
+    fn with_capacity(capacity: usize) -> Self {
+        Self {
+            verifier: ed_core::batch::Verifier::with_capacity(capacity),
+        }
+    }
+
```
clabby: Would be sweet to also use this in `cryptography/src/ed25519/certificate/mod.rs` - looks to still use `Batch::new`

cryptography/src/ed25519/core/batch.rs
patrick-ogrady: Yep, that's known 👍

cryptography/src/ed25519/core/batch.rs
patrick-ogrady: IIRC it was also allocate on the old path too just wouldn't retain (although I suppose the claim is holding it would make it harder to optimize).

cryptography/src/ed25519/core/batch.rs
patrick-ogrady: Half-right: the feature plumbing was unneeded — dropped `ahash/runtime-rng` and `ahash/std` so it's plain `no-rng` everywhere now (552e3e0). But `AHashMap` itself can't replace hashbrown: in ahash 0.8.12 it wraps `std::collections::HashMap` behind `cfg(feature = "std")` (`no-rng` doesn't unlock it), so a no_std build fails with `error[E0432]: unresolved import ahash::AHashMap`. `hashbrown::HashMap` + `ahash::RandomState` is the same map/hasher pair spelled for no_std. Rewrote the comment to say that.

cryptography/src/lib.rs
patrick-ogrady: Moot now: `with_capacity` is gone entirely — `BatchVerifier::new(capacity)` takes the hint directly (552e3e0).

cryptography/src/ed25519/core/batch.rs
patrick-ogrady: No — once hashing is deferred the payload has to outlive `add` (the old `Cow::Borrowed` arm was only sound because `Item::from` hashed it before returning), so storing a `Cow` would push a lifetime through `Verifier`, `Batch`, and the `BatchVerifier` trait. And the borrowed arm only ever fired for `namespace: None`, which nothing outside tests uses — the trait's `add` always frames a namespace, so it allocated `union_unique` either way.

cryptography/src/ed25519/core/batch.rs
patrick-ogrady: Benchmarked the worst interleave (32k sigs, `Rayon(8)`, M5 Pro, min of 8 iters, E2E = add loop + verify):

| shape | E2E |
|---|---|
| 32k distinct keys | 27.9 ms |
| 4096 signers x 8 sigs, round-robin interleaved | 29.0 ms |
| 4096 signers x 8 sigs, grouped (what sorting yields) | 19.1 ms |
| 64 signers x 512 sigs, interleaved | 18.5 ms |
| 64 signers x 512 sigs, grouped | 17.6 ms |
| sorting 32k queued tuples by key | ~2.5 ms (serial) |

Interleaving degrades gracefully: per shard the msm is `n_shard + distinct_in_shard + 1 <= 2*n_shard + 1`, so the worst case is bounded by the all-distinct cost (29.0 vs 27.9 above) — coalescing is opportunistic, never a pessimization. Sorting for the user would recover up to ~34% in the adversarially interleaved shape, but costs ~2.5 ms serial on every verify — a ~9% regression on the dominant all-distinct shape (certificates/votes), which gets nothing from it. Since shards are contiguous queue ranges, callers who know their duplicate structure get the full win for free by queueing same-key items adjacently — documented that in the module docs (552e3e0). Prefer keeping insertion order and letting the caller choose.

cryptography/src/ed25519/core/batch.rs
patrick-ogrady: Documented on `Verifier::queue` and the public `Batch` struct (552e3e0).

cryptography/src/ed25519/scheme.rs
patrick-ogrady: Done — and capacity now flows through everywhere: `BatchVerifier::new(capacity)` requires the hint, and the certificate paths pass `size_hint`/signature counts (552e3e0).

cryptography/src/ed25519/core/batch.rs
patrick-ogrady: Considered it, but deferring gives the same parallelism with one shape instead of two: `add_many` would need callers to buffer `(namespace, msg, key, sig)` tuples themselves and pass a `Strategy` at add time, and the batch retains the payloads either way until `verify`. Keeping the old per-key map structure would also keep the one-group-per-shard imbalance that made single-signer batches verify serially. Cheap `add`, all heavy work under `verify` — sticking with that.

cryptography/src/ed25519/core/batch.rs
patrick-ogrady: Tried transcript-seeding (`RandomState::with_seeds` from the per-shard RNG) and it works with no measurable cost, but we already accept fixed-seed hashing on adversarial input elsewhere (`utils`/`storage` run `no-rng` on no_std), so keeping plain `RandomState::default()` for consistency. Worth revisiting if we ever tighten that assumption workspace-wide.

cryptography/src/ed25519/core/batch.rs
patrick-ogrady: Follow-up: 63fda7d supersedes my recommendation above with a middle path. Instead of sorting, `verify` now partitions the batch with a counting sort on each key's first byte (O(n), no comparisons, ~0.2ms at 32k) so signatures sharing a key land in the same shard regardless of queue order. Co-sharding is all coalescing needs — the per-shard map handles the rest, so adjacency (and a full sort) is unnecessary. Your `[A, B, A, B, ...]` example now shards as `[[A, A, ...], [B, B, ...]]` exactly as suggested. Measured: 4096 signers x 8 sigs round-robin 28.5 -> 20.2ms (-29%), certificate-major 50x640 19.6 -> 18.1ms (-7%), other shapes within noise. Skewed batches still split evenly across shards (chunking is by count, not by bucket), so the single-signer win is preserved.

PR #4160, opened 2026-07-02
[runtime] Make `buffer::Write` cancellation-safe

danlaine:
## Motivation

Dropping a future mid-operation is normal in async Rust: a caller times out, loses a `select!`, or shuts down. `buffer::Write` was not safe against this. Its flush paths emptied the in-memory buffer *before* awaiting the blob write that persisted those bytes:

```rust
if let Some((buf, offset)) = self.buffer.take() {          // buffer emptied here
    self.sync_state.write_at(&self.blob, offset, buf).await?;  // ...dropped here?
}
```

Drop the future at that await and the bytes are gone from memory but never reached storage. The writer looks clean, reports the wrong contents, and no retry can recover the data.

A second, subtler hole: a dropped write may still reach the disk after the fact. The writer only marked "needs sync" *after* a write succeeded, so a dropped-but-landed write was invisible to durability tracking, and a later `sync()` could return `Ok` without ever fsyncing those bytes.

A third hole, specific to `resize`: a resize dropped *before* reaching the blob was silently lost. Once the writer adopts a new size in memory, no later operation can tell that the physical resize never ran — a retry concludes there is nothing to do, `sync()` reports `Ok`, and a reopen resurrects the old size.

## The rule

At every await point, in-memory state must be valid whether the awaited operation completed or not:

- State recording the *presence* of data is updated only **after** the write succeeds.
- State scheduling *future work* (the needs-sync flag, the pending resize) is updated **before** the operation starts, so work that may have happened is never forgotten.

## Changes

- `Buffer::advance(n)` (new): discards the first `n` buffered bytes. Flushes now write a borrowed view of the buffer to the blob, then call `advance` only after the write returns. A dropped flush leaves the buffer intact for retry.
- All four flush sites in `Write` (`write_at` overlap flushes, `resize`, `sync`, `start_sync`) route through one private `flush(Durability)` helper implementing that protocol, instead of four hand-rolled copies of it. A durable flush guarantees all tracked mutations are durable on return, so `sync` is just `flush(Durable)`.
- `SyncState` marks itself dirty *before* awaiting a write or resize, so a dropped-but-landed mutation is still covered by the next sync (at worst costing one spurious fsync).
- `SyncState` records a resize's intended length before awaiting it and clears the record when the resize returns; every later blob operation (write, sync, start_sync) first re-issues an unacknowledged resize. Resizes are absolute, so re-issuing is a no-op if the dropped resize actually landed. A dropped resize is therefore settled by whatever the caller does next, with no need to retry the resize itself. A new resize first applies an incomplete smaller one: growing zero-fills the new bytes, so a grow must not skip a shrink's removal.
- `Write::resize` grows by publishing the new size only after the blob resize returns, and shrinks by adopting the shorter size *first*: a drop after the resize lands can no longer leave the buffer claiming more than the blob holds, and the recorded intent covers a drop before it lands.
- `Buffer::resize`'s grow arm (hand back buffered bytes for the caller to flush) is unreachable now that `Write::resize` flushes before growing: it becomes the shrink-only `Buffer::truncate`, which panics if asked to grow, and grows move the empty tip's offset directly.
- Tests: a gate harness parks each operation at its blob write or resize (before or after it reaches the blob), drops the future, and asserts the writer's state is consistent and a later operation completes the work. Covers overlap flush, grow/equal resize, shrink canceled before and after the blob resize (including while parked on an in-flight started sync), and sync.

`Buffer::take`/`drop_prefix` remain for the paged writer, which migrates to `slice`/`advance` (and deletes them) in the next PR of this series.


runtime/src/utils/buffer/mod.rs
```diff
@@ -56,34 +46,91 @@ enum SyncState {
...
+        let Some(len) = self.pending_resize else {
+            return Ok(());
+        };
+        self.wait_for_pending().await?;
+        // Mark dirty so the re-issued resize is covered by the next sync even if the state
+        // was clean (a resize dropped while waiting on an in-flight sync leaves it clean).
+        self.mark_dirty();
+        blob.resize(len).await?;
+        self.pending_resize = None;
+        Ok(())
```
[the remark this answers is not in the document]
danlaine: Working as intended (now documented on `Write::resize`).

The dropped resize never returned `Ok`, so the caller can't expect the new tail to exist. The writer keeps reporting the old size, and everything it does claim is readable. That's the contract for a canceled operation.

This state can't be prevented anyway: a dropped `blob.resize` may land on disk after the future. Settling just makes that deterministic. A caller that wants the grow retries `resize`, which publishes normally.

Having `settle` update the buffer too would mean a canceled resize takes effect later, mid way through some unrelated call.

runtime/src/utils/buffer/write.rs
```diff
@@ -208,28 +213,29 @@ impl<B: Blob> Write<B> {
...
-        //
-        // This can only happen if the new size is greater than the current size.
-        if let Some((buf, offset)) = self.buffer.resize(len) {
-            self.sync_state.write_at(&self.blob, offset, buf).await?;
+        if len >= self.buffer.size() {
+            // Grow (or keep the size): flush buffered bytes, resize the blob, and move the
+            // now-empty tip to the new end.
+            self.flush(NonDurable).await?;
+            self.sync_state.resize(&self.blob, len).await?;
+            let _ = self.buffer.resize(len);
```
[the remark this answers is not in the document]
danlaine: A grow only takes effect once resize returns Ok; a dropped call leaves an under-claim, which is the safe direction — and the longer-blob state is reachable anyway since a dropped blob.resize can land after the future is dropped.

roberto-bayardo:
The retry-recovers-the-data half of the guarantee here depends on the blob-layer ordering contract from #4180: without it, an orphaned flush can land *after* its retry (drop a flush of `v1`, overwrite the range in the tip, flush `v2`, sync — the orphaned `v1` lands last, disk is stale, writer reports `Clean`). This is real on both non-deterministic backends: tokio's dropped `spawn_blocking` task still runs whenever the pool schedules it, and an io_uring SQE already staged in the ring completes regardless, unordered against later SQEs.

Could you update the description to cross-reference #4180 as the other half of the story and note the intended sequencing (#4180 first, or both in the same release)?

runtime/src/utils/buffer/write.rs
```diff
@@ -252,22 +259,31 @@ impl<B: Blob> Write<B> {
         self.sync_state.wait_for_pending().await
     }
 
-    /// Write bytes to the underlying blob and make them durable.
+    /// Flush all buffered bytes to the blob. Internal state is only modified on success, so
+    /// retry after cancellation or failure is possible.
```
roberto-bayardo, line 270: drop the second sentence? (cancellation safety is provided at the module level so no need to state it here.)

runtime/src/utils/buffer/write.rs
```diff
@@ -252,22 +259,31 @@ impl<B: Blob> Write<B> {
...
-    async fn write_blob_sync(
-        &mut self,
-        offset: u64,
-        bufs: impl Into<IoBufs> + Send,
-    ) -> Result<(), Error> {
-        self.sync_state
-            .write_at_sync(&self.blob, offset, bufs)
-            .await
-    }
+    /// When `durability` is [Durable], [Write] is guaranteed to be durable when this returns.
```
roberto-bayardo, line 272: Referencing [Write] (a struct) is a bit awkward.  How about something like: 
```suggestion
    /// When `durability` is [Durable], all pending mutations are made durable before returning.
```

runtime/src/utils/buffer/mod.rs
```diff
@@ -56,34 +46,92 @@ enum SyncState {
...
                 Err(err)
             }
         }
     }
 
+    /// Re-issue a resize whose completion was never observed.
+    ///
+    /// [crate::Blob::resize] sets an absolute size, so re-issuing is a no-op if the dropped
+    /// resize actually completed.
+    async fn settle(&mut self, blob: &impl crate::Blob) -> Result<(), crate::Error> {
```
roberto-bayardo, line 122: `settle_resize` ?

runtime/src/utils/buffer/paged/writer.rs
```diff
@@ -1016,10 +1016,15 @@ impl<B: Blob> Writer<B> {
 
         // A logical shrink can leave the physical page count unchanged. Only real physical
         // resizes need to create a pending sync.
+        //
+        // Resize the blob directly, not through [SyncState::resize]: this writer publishes
+        // its state only after the truncate returns, so a recorded resize re-issued by a
+        // later operation (see [SyncState::settle]) would truncate under state that still
+        // claims the old size. A dropped truncate here is simply a no-op.
```
roberto-bayardo, line 1023: Confusing comment. Fable suggested rewrite (not sure it's a lot better but ymmv):

```suggestion
// Resize the blob directly instead of calling [SyncState::resize], which would
// allow a later operation to re-issue the resize via [SyncState::settle].
// This writer only updates its in-memory size after the blob resize below
// returns, so a replayed resize could shrink the blob while the writer still
// claims the old size.
```

danlaine:
Replaced by #4228

PR #4162, opened 2026-07-02
[storage] Make segmented journal blob removal cancellation-safe

danlaine:
## Motivation

Every removal path in the segmented journal's `Manager` (prune, remove_section, rewind, clear, destroy) untracked a section *before* awaiting the storage removal that deleted its blob:

```rust
// Remove blob from map
let mut blob = self.blobs.remove(&section).unwrap();   // untracked here
blob.wait_for_sync().await?;
...
// Remove blob from storage
self.context
    .remove(&self.partition, Some(&section.to_be_bytes()))
    .await?;   // ...dropped here?
```

Drop the future at either await and the manager has forgotten a section whose blob still exists: reads and metrics report it gone, but a restart rescans the partition, rediscovers the blob, and the pruned data resurrects. `clear` was the extreme case: it emptied the entire section map up front, so a drop mid-way untracked every section while their files survived.

Removing the blob first creates the opposite: a dropped removal may have unlinked the blob while the section is still tracked (and a dropped remove can land after the future is dropped). That world is easy to repair, so it is the one to prefer: the retry just has to tolerate a blob that is already missing.

## The rule

Same as #4160, applied to removal: at every await point, in-memory state must be valid whether the awaited operation completed or not. A map entry records the *presence* of a section, so it is removed only after the storage removal succeeds. A dropped removal can then only leave a section over-tracked, never forgotten while its data survives on disk.

## Changes

- All removal paths funnel through one `remove_blob` helper: join any in-flight sync, remove the blob from storage, and only then untrack it. A dropped removal leaves the section tracked, and a retry completes it.
- A missing blob is tolerated during removal (the dropped call's delete may have landed), and `destroy` likewise tolerates an already-missing partition.
- Tolerating a missing blob is only sound if `BlobMissing` means the file is actually gone: the tokio and io_uring backends now map only `NotFound` remove errors to `BlobMissing`/`PartitionMissing`. Previously any removal failure was reported as missing, which this tolerance would have swallowed.
- `clear` no longer empties the section map up front; it removes sections one at a time, staying consistent at every step.
- The in-flight sync is joined before the unlink so caller-held sync handles report the sync's true result and no buffer is dropped with I/O in flight. The blob handle stays open across the unlink, which every storage backend supports.
- Tests: prune, clear, and destroy each tolerate a section whose blob is already gone from storage, as a dropped-then-landed removal leaves it.


roberto-bayardo:
Sol finds a couple issues that look legit?

> 1. **[P1] Pre-unlink cancellation still resurrects sections.** `remove_blob` untracks the section before awaiting `Storage::remove`. I reproduced this by parking removal before it touched storage, dropping `rewind`, and reopening:
> 
>    - Live manager: newest section is `1`.
>    - Disk still contains sections `1` and `2`.
>    - Reopened manager: section `2` returns.
> 
>    The new regression test covers only the opposite timing: removal fully completes and then the future is dropped.
> 
> 2. **[P2] Unlink-before-directory-fsync remains unsafe.** Tokio awaits `remove_file`, then separately awaits `sync_dir`. Cancellation between them leaves an undurable deletion. Because the manager has already untracked the section, it will not retry the removal; additionally, a `BlobMissing` retry would skip the directory sync. A power failure can therefore restore the supposedly removed blob.
> 
> The PR description is now materially stale: it repeatedly says storage removal happens before untracking, while the implementation intentionally does the reverse.
> 
> The branch’s own focused tests pass, but the symmetric pre-removal diagnostic fails exactly as above. Fixing this likely requires an explicit “removing” state or another mechanism that represents the operation until its durable outcome is known; merely swapping the two operations always leaves one cancellation window.

storage/src/journal/segmented/manager.rs
```diff
@@ -329,6 +315,42 @@ impl<E: Storage + Metrics, F: BufferFactory<E::Blob>> Manager<E, F> {
         Ok(())
     }
 
+    /// Stop tracking `section`, then remove its blob from storage. Returns the blob's size.
+    ///
+    /// The map update happens before the removal awaits: a dropped removal must never leave
```
roberto-bayardo, line 320: I'd put these implementation details inside the fn?

danlaine:
> Sol finds a couple issues that look legit?
> 
> > 1. **[P1] Pre-unlink cancellation still resurrects sections.** `remove_blob` untracks the section before awaiting `Storage::remove`. I reproduced this by parking removal before it touched storage, dropping `rewind`, and reopening:
> >    
> >    * Live manager: newest section is `1`.
> >    * Disk still contains sections `1` and `2`.
> >    * Reopened manager: section `2` returns.
> >    
> >    The new regression test covers only the opposite timing: removal fully completes and then the future is dropped.
> > 2. **[P2] Unlink-before-directory-fsync remains unsafe.** Tokio awaits `remove_file`, then separately awaits `sync_dir`. Cancellation between them leaves an undurable deletion. Because the manager has already untracked the section, it will not retry the removal; additionally, a `BlobMissing` retry would skip the directory sync. A power failure can therefore restore the supposedly removed blob.
> > 
> > The PR description is now materially stale: it repeatedly says storage removal happens before untracking, while the implementation intentionally does the reverse.
> > The branch’s own focused tests pass, but the symmetric pre-removal diagnostic fails exactly as above. Fixing this likely requires an explicit “removing” state or another mechanism that represents the operation until its durable outcome is known; merely swapping the two operations always leaves one cancellation window.

P1 shows the difficulty of a dropped removal: unlink is an `await`, so a drop can leave the manager's tracking and the disk disagreeing in either direction.

If we unlink first, untrack second, a drop after the unlink lands leaves the manager holding a live handle to an unlinked file. Writes to it are acknowledged but the data has no directory entry left, so nothing can reach it after restart -> silent loss of acknowledged data.

If we untrack first, unlink second (current 4162 code), a drop before the unlink lands leaves the file on disk with the manager having no memory of it -> on restart we re-scan and resurrect the section.

> Fixing this likely requires an explicit "removing" state

I think this is probably right but I'm not super enthusiastic about it. I'm wary about adding a bunch of cancelation detection / repair code here and in the other data stores. I tend to prefer the approach of #4246 where, if you cancel a mutable method, you simply lose the handle to it and there's no in-memory state to reconcile with durable state.

danlaine:
Closing in favor of #4246

PR #4163, opened 2026-07-02
[storage] Make `Metadata::sync` cancellation-safe

danlaine:
## Motivation

`Metadata` keeps two blobs and alternates syncs between them, so one blob always holds the last committed state. `sync` flipped the cursor, bumped the version, and cleared modified-key tracking *before* awaiting the writes that persisted the target blob:

```rust
// Update the state.
self.state.cursor = target_cursor;
self.state.next_version = next_next_version;
...
// Clear modified keys to avoid writing the same data
target.modified.clear();
...
target.blob.write_at_sync(0, next_data.clone()).await?;   // ...dropped here?
```

Drop the future at that await and the bookkeeping describes a sync that never happened: the cursor points at a blob that was never written, so the next sync targets the blob holding the last committed state, and the wiped modified set means the skipped keys are never written again. Later syncs return `Ok` while silently omitting them.

A second, subtler issue: a dropped write may still land. Recovery requires the checksum to sit at the physical end of the blob, but a landed write from a dropped larger rewrite leaves trailing bytes that a later, shorter rewrite (a bare `write_at_sync`) would not remove. Recovery then reads garbage where it expects the checksum and discards the newer state. An interrupted rewrite can likewise leave stale bytes at offsets the overwrite path never touches, since overwrites assume the unmodified bytes on disk match the in-memory mirror.

## The rule

Same as #4160: at every await point, in-memory state must be valid whether the awaited operation completed or not. Bookkeeping that steers the next sync is published only after the target blob is durable; hazards a dropped write may leave on disk are repaired unconditionally by the next sync instead of assumed absent.

## Changes

- Both sync paths publish their bookkeeping (cursor, next version, cleared modified keys, and the blob mirror) only after their writes succeed. A dropped sync changes nothing: the retry targets the same blob with the same modified keys.
- Rewrites always end with a resize and full sync (previously only shrinking rewrites resized). The resize enforces the checksum-at-physical-end invariant recovery depends on, even when a dropped earlier sync left the blob longer than the in-memory mirror suggests.
- Entering the rewrite path marks the key order changed, so an interrupted rewrite (which can leave stale bytes an overwrite would never repair) is retried as a rewrite until one succeeds.
- Tests: a rewrite truncates stale trailing bytes planted in the target blob (and the result survives a restart), and a length-changing update forces rewrites instead of overwrites until one succeeds.


danlaine:

PR #4167, opened 2026-07-03
[codec] Bound decoder pre-allocation by remaining buffer (read_vec / HashSet / HashMap)

### LINKED ISSUE #4088: [codec] Bound decoder pre-allocation by remaining buffer (read_vec / HashSet / HashMap)
The byte decoder guards its allocation against the buffer before reading, but the generic collection decoders do not.

`types/bytes.rs` (correct):

```rust
let len = usize::read_cfg(buf, range)?;
at_least(buf, len)?;            // len can't exceed what's actually in the buffer
Ok(buf.copy_to_bytes(len))
```

`codec.rs::read_vec` (and `types/hash_set.rs`, `types/hash_map.rs`):

```rust
let len = usize::read_cfg(buf, range)?;
let mut values = Vec::with_capacity(len);   // allocates len up front, before reading anything
for _ in 0..len {
    values.push(Self::read_cfg(buf, cfg)?);
}
```

`read_vec` is `codec.rs:204`, `HashSet` is `types/hash_set.rs:69`, `HashMap` is `types/hash_map.rs:80`.

### Impact

`len` is an untrusted length prefix (bounded to `u32` per the `Vec` docs). A tiny message can claim `len` near `u32::MAX` and force a multi-GB `with_capacity` before a single element is read. The loop then fails fast with `EndOfBuffer`, but the eager allocation already happened. This is the CWE-789 / allocation-before-validation pattern (same class parity-scale-codec hardened against).

`RangeCfg` bounds the logical element count, but a legitimately large configured max (e.g. "up to N validators/items") still lets a small message trigger an allocation proportional to that max rather than to the data actually present.

### Proposed fix

Cap the initial capacity by what the buffer can possibly contain. Every element needs at least one byte, so `buf.remaining()` is a hard upper bound on the number of decodable elements:

```rust
let mut values = Vec::with_capacity(len.min(buf.remaining()));
```

This keeps the existing `RangeCfg` validation and the final length, and only avoids eager over-allocation, matching the intent already present in `bytes.rs`. Same idea for the `HashSet` / `HashMap` paths.

Is `RangeCfg` intended as the sole defense here, or would you take the buffer-bounded cap? Happy to put up a PR (with a regression test) if so.

Ysh204:
## Summary

Fixes #4088.

This PR bounds the initial allocation performed by the generic collection decoders using the remaining bytes in the input buffer.

Previously, `read_vec`, `HashSet`, and `HashMap` allocated `len` elements immediately after reading the untrusted length prefix:

```rust
let mut values = Vec::with_capacity(len);
```

A malformed input could advertise a very large length while providing only a small payload, causing excessive upfront allocation before decoding failed with `Error::EndOfBuffer`.

This change caps the initial capacity using:

```rust
Vec::with_capacity(len.min(buf.remaining()))
```

The same change is applied to the `HashSet` and `HashMap` decoders. This preserves the existing `RangeCfg` validation and decoding behavior while avoiding unnecessary pre-allocation from malformed inputs.

## Test plan

- Added regression tests for `Vec`, `HashSet`, and `HashMap` using oversized advertised lengths with insufficient payloads.
- Verified malformed inputs return `Err(Error::EndOfBuffer)`.
- Verified valid inputs continue to decode successfully.
- Ran `cargo fmt --all`.
- Ran `cargo clippy --workspace --all-targets`.
- Ran `cargo test --workspace`.


danlaine:
I'm not sure the new unit tests really do anything -- they assert `EndOfBuffer` which is the result before this PR's fix too.

Ysh204:
Ya, You're right the regression tests only verify the externally observable behavior, which is unchanged. They don't actually demonstrate that the initial allocation is now bounded. I'll take another look at whether there's a way to make the allocation behavior observable in a test, otherwise I'm happy to remove these tests if they don't provide meaningful coverage.

Ysh204:
@danlaine Good point. The original regression tests only verified the decoder's external behavior, which was unchanged before and after the fix.

I've added a deterministic regression test that exercises read_vec directly with usize::MAX. Without the fix, it deterministically panics with a capacity overflow. With the fix, the allocation is bounded by buf.remaining() and the decoder returns Err(Error::EndOfBuffer).

I've pushed the updated test.

danlaine (dismissed):

Ysh204:
@danlaine Thanks again for the review! I've addressed the previous feedback and pushed the updated regression test. All checks are passing now. Is there anything else you'd like me to change, or would this be ready to merge?

Ysh204:
@patrick-ogrady sorry for intruding, but i think this Pr looks good , any chance of merge?

patrick-ogrady (approved):
No application should be using `::MAX` for network traffic (and thus shouldn't benefit from this hardening), however, it should make it harder/more expensive for a peer to cause us to allocate within a permitted range (getting us to allocate 1M items now also requires sending us a >= 1M bytes).

patrick-ogrady (dismissed):

patrick-ogrady (approved):

PR #4169, opened 2026-07-03
consensus/marshal: handle non-monotonic parent heights during gap repair

### LINKED ISSUE #4155: [consensus/marshal] try_repair_gaps can spin forever at startup on a malformed finalized-blocks archive, silently blocking the marshal mailbox
## Summary

`marshal::core::Actor`'s startup gap repair (`try_repair_gaps`) has no progress guard. If the finalized-blocks archive contains a gap whose end block's parent-commitment resolves to a block **below the gap start** (a "gap-skipping parent"), the repair loop spins forever on one core. Because `try_repair_gaps` runs **before** the actor's mailbox `select_loop` (`consensus/src/marshal/core/actor.rs:386` vs `:398` at `main`/`8ca37902`), the entire marshal becomes unresponsive — `GetBlock`, `GetProcessedHeight`, subscriptions, everything — with no log output after the startup/archive-restore lines. Anything upstream awaiting a marshal round-trip at startup (e.g. `glue`'s `init_databases_from_marshal`, which awaits `get_processed_height()` + `get_block(...)` before opening its databases, `glue/src/stateful/actor/syncer/mod.rs:390/:405`) hangs indefinitely.

## Mechanism

The backward walk (`actor.rs:2000-2030`):

```rust
while height > gap_start {
    if let Some(block) = self.find_block_by_commitment(buffer, parent_commitment).await {
        let next = (block.height(), block.parent(), V::parent_commitment(&block));
        wrote |= self.store_finalization(next.0, parent_digest, block, finalization, application).await;
        (height, parent_digest, parent_commitment) = next;
    } else { /* fetch + break 'cache_repair */ }
}
```

There is (a) no assertion that the found block's height strictly decreases toward `gap_start`, and (b) no check that the gap actually shrank before the outer `'cache_repair` loop re-runs `next_gap`. Both store paths no-op silently for an out-of-gap block: `store_finalization` drops blocks at/below the processed floor (`actor.rs:1769`), and `Archive::put` skips existing indices (`storage/src/archive/prunable/storage.rs:~318`). So a gap-skipping parent ⇒ the walk exits without filling the gap ⇒ `next_gap` returns the same gap ⇒ infinite loop. The outer loop only terminates when no gap remains (`return wrote`, `:1977-1979`) or the parent lookup misses (the `else` branch) — a gap that never shrinks re-runs forever.

The walk's `SAFETY` comment says finalized blocks are archived only after marshal validated the parent linkage — a sound invariant for marshal's own writes. But the archive is durable state: disk corruption, a partial/interrupted write, operator tooling, or any historical write that predates the invariant can violate it. A malformed archive should fail loud (or self-heal), not hang the node silently at one core forever.

## Reproduction (deterministic, drop-in for the existing `marshal/standard` test harness — all helpers exist at `main`)

```rust
#[test_traced("WARN")]
fn test_try_repair_gaps_spins_on_gap_skipping_parent() {
    let runner = deterministic::Runner::timed(Duration::from_secs(15));
    runner.start(|mut context| async move {
        let Fixture { schemes, .. } =
            bls12381_threshold_vrf::fixture::<V, _>(&mut context, NAMESPACE, NUM_VALIDATORS);
        // gap_end block at height 10 whose parent is the height-1 block → gap 2..=9
        let block1 = make_raw_block(Sha256::hash(b"repro-genesis"), Height::new(1), 100);
        let block10 = make_raw_block(block1.digest(), Height::new(10), 1000);
        let fin1 = StandardHarness::make_finalization(
            Proposal::new(Round::new(Epoch::zero(), View::new(1)), View::zero(),
                StandardHarness::commitment(&block1)), &schemes, QUORUM);
        let fin10 = StandardHarness::make_finalization(
            Proposal::new(Round::new(Epoch::zero(), View::new(10)), View::new(9),
                StandardHarness::commitment(&block10)), &schemes, QUORUM);
        let partition_prefix = "repro-gap-skip";
        seed_inconsistent_restart_state(context.child("storage"), partition_prefix,
            &[block1.clone(), block10.clone()],
            &[(Height::new(1), fin1.clone()), (Height::new(10), fin10.clone())]).await;
        seed_cache_block(context.child("c1"), partition_prefix, Epoch::zero(), View::new(1), &block1).await;
        seed_cache_block(context.child("c10"), partition_prefix, Epoch::zero(), View::new(10), &block10).await;
        seed_processed_height(context.child("ph"), partition_prefix, Height::new(1)).await;
        let (application, _rx) = HoldingBlockReporter::new();
        // Start::Floor matches the harness's restart-state tests; it short-circuits
        // against the seeded processed height and is not load-bearing for the repro.
        let (mailbox, _buf, _res, _handle) = start_standard_actor(
            context.child("validator"), partition_prefix,
            ConstantProvider::new(schemes[0].clone()), application,
            Some(RecordingBuffer::default()), Start::Floor(fin1)).await;
        // Never returns: the actor busy-spins in try_repair_gaps before its mailbox loop.
        let ph = mailbox.get_processed_height().await;
        panic!("actor serviced mailbox (processed_height={ph:?}) — repair loop terminated");
    });
}
```

**Observed** (at a v2026.5.0-era rev; the loop is unchanged at `main`/`8ca37902`): the test hangs forever — even the deterministic runner's 15 s virtual-time budget never fires, because the spinning task never parks and simulated time cannot advance. A stack sample of the hung process shows the on-CPU frame inside `try_repair_gaps` (the `'cache_repair` loop / `get_finalized_block(gap_end)`). The existing well-formed-gap tests (`test_standard_local_floor_anchor_resumes_gap_repair` etc.) pass — only the gap-skipping poison hangs.

## Suggested fix

Guard progress in the walk — and **skip/heal rather than assert**: on detecting a non-decreasing parent height (or a `next_gap` that did not shrink after a repair pass), log loudly, drop the offending walkback, and fall through to the existing resolver-fetch branch (`Request::finalized_block_by_height`) so the actor still boots and repairs via fetch. A bare `assert!` would only convert the silent hang into a boot panic loop — the node still never comes up.

Happy to contribute the fix + this regression test as a PR if the direction sounds right.

Ysh204:
## Summary

Fixes #4155.

During startup gap repair, `try_repair_gaps` assumes that each resolved parent block is strictly lower than the current block. A malformed finalized-blocks archive containing a non-monotonic parent link can violate this assumption, causing the repair loop to repeatedly rediscover the same gap and preventing the marshal actor from reaching its mailbox loop.

This change treats a resolved parent whose height is not strictly less than the current height as a missing parent. A warning is emitted, and the existing fetch path is reused to request the expected finalized block, allowing startup recovery to continue instead of spinning indefinitely.

## Test plan

- Added regression tests covering malformed archives with non-monotonic parent links across the marshal harnesses.
- Verified the actor remains responsive and recovers without hanging.
- Ran:

```bash
$env:AWS_LC_SYS_NO_ASM="1"; cargo test -p commonware-consensus -- non_monotonic_parent_height
```

### DIFF
--- a/consensus/src/marshal/coding/mod.rs
+++ b/consensus/src/marshal/coding/mod.rs
@@ -967,6 +967,11 @@ mod tests {
     }
 
     #[test_traced("WARN")]
+    fn test_coding_non_monotonic_parent_height() {
+        harness::non_monotonic_parent_height::<CodingHarness>();
+    }
+
+    #[test_traced("WARN")]
     fn test_coding_certify_persists_equivocated_block() {
         harness::certify_persists_equivocated_block::<CodingHarness>();
     }
--- a/consensus/src/marshal/core/actor.rs
+++ b/consensus/src/marshal/core/actor.rs
@@ -2002,16 +2002,16 @@ where
                     .find_block_by_commitment(buffer, parent_commitment)
                     .await;
 
-                // Treat a block whose height does not strictly decrease as
-                // missing so that the existing fetch path requests the correct
-                // block from the network. This prevents an infinite loop when
-                // the archive contains corrupted parent pointers.
+                // Treat a block whose parent height is not contiguous (does not
+                // decrease by exactly one) as missing so that the existing fetch path
+                // requests the correct block from the network. This prevents an infinite
+                // loop when the archive contains corrupted parent pointers.
                 let parent_block = match parent_block {
-                    Some(ref block) if block.height() >= height => {
+                    Some(ref block) if Some(block.height()) != height.previous() => {
                         warn!(
                             current = %height,
                             parent = %block.height(),
-                            "non-monotonic parent height during gap repair, treating as missing"
+                            "non-contiguous parent height during gap repair, treating as missing"
                         );
                         None
                     }
--- a/consensus/src/marshal/core/actor.rs
+++ b/consensus/src/marshal/core/actor.rs
@@ -1998,10 +1998,27 @@ where
 
             // Iterate backwards, repairing blocks as we go.
             while height > gap_start {
-                if let Some(block) = self
+                let parent_block = self
                     .find_block_by_commitment(buffer, parent_commitment)
-                    .await
-                {
+                    .await;
+
+                // Treat a block whose height does not strictly decrease as
+                // missing so that the existing fetch path requests the correct
+                // block from the network. This prevents an infinite loop when
+                // the archive contains corrupted parent pointers.
+                let parent_block = match parent_block {
+                    Some(ref block) if block.height() >= height => {
+                        warn!(
+                            current = %height,
+                            parent = %block.height(),
+                            "non-monotonic parent height during gap repair, treating as missing"
+                        );
+                        None
+                    }
+                    other => other,
+                };
+
+                if let Some(block) = parent_block {
                     let finalization = self.cache.get_finalization_for(parent_digest).await;
                     let next = (block.height(), block.parent(), V::parent_commitment(&block));
                     wrote |= self
--- a/consensus/src/marshal/mocks/harness.rs
+++ b/consensus/src/marshal/mocks/harness.rs
@@ -5252,3 +5252,75 @@ pub fn broadcast_caches_block<H: TestHarness>() {
             .expect("block should be cached after broadcast");
     })
 }
+
+/// Test that a non-monotonic parent height does not cause an infinite loop during gap repair.
+pub fn non_monotonic_parent_height<H: TestHarness>() {
+    let runner = deterministic::Runner::timed(Duration::from_secs(10));
+    runner.start(|mut context| async move {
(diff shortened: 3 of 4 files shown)


patrick-ogrady:
> A malformed finalized-blocks archive containing a non-monotonic parent link can violate this assumption, causing the repair loop to repeatedly rediscover the same gap and preventing the marshal actor from reaching its mailbox loop.

This is considered a broken invariant (and not something `marshal` supports). If there is any missing documentation that would make this clearer, feel free to take a pass.

Ysh204:
Thanks for the clarification! That makes sense. I interpreted #4155 as making the recovery path robust to malformed archives, but if non-monotonic parent links are considered an unsupported invariant violation, I'm happy to adjust the direction.

Would you prefer closing this PR in favor of improving the documentation around that invariant instead, or is there another behavior you'd like to see here?

PR #4172, opened 2026-07-04
[storage/qmdb] Parallelism Nits

patrick-ogrady:
Related: #4121

A final review sweep of #4121 before merge, focused on API minimalness, trait-bound noise, and correctness. Every substantive change below was verified end-to-end (full workspace check, clippy, MSRV 1.91.1, stability checks, and the equivalence suite including the `#[ignore]`d P=3 tests in release).

## Bug fix: worker join no longer panics the runtime

`handle.expect("snapshot worker task failed")?` in the parallel build's join loop is now `handle??`. A supervision-tree abort (an ancestor task completing or aborted mid-init) resolves worker handles as `Err(Error::Closed)` rather than panicking. On multithreaded tokio the cascade can land while init is mid-poll in `join_all`, so the `expect` converted an orderly abort of one subtree into a whole-runtime panic (default `catch_panics = false` resume-unwinds at the root). The deterministic runtime is single-threaded and always short-circuits at init's own `Abortable` first, which is why no test could catch this. `qmdb::Error` already has `Runtime(#[from] ...)`, and the production precedent (`try_join_all` in journal `segmented/manager.rs`) returns the error.

## Derive init parallelism from the configured `Strategy`

`InitParallelism` (enum + `pub init_parallelism` Config field) is deleted. It had zero production consumers: all ~45 call sites set `Serial` (tests/fuzz/benches/examples, all `Sequential`, where derivation is behaviorally identical), and both state-sync paths ignored the field and hardcoded `Auto` — silently overriding a user's `Serial`. The repo idiom is to derive parallelism from the strategy at the point of use (`get_many_map`, batch resolution, ed25519 batching, reed_solomon striping), so init now does the same:

- `build_snapshot` takes the strategy (consumed by value, since the caller cloned it off the log anyway) instead of `strategy + parallelism` args.
- The partitioned-ordered impl splits the build across `parallelism_hint()` workers, using the hint directly like every other consumer (no reserved core: the operator's budget already accounts for other work). A hint of one builds serially, mirroring how `Strategy::run` dispatches serial bodies for strategies without parallelism.
- `Strategy::run` itself cannot host this dispatch: its arms are synchronous closures (both would need `&mut self` to build their futures), `Manual::run` delegates to the inner strategy's `run` (so a manually-hinted `Sequential` would never parallelize), and policy-backed `Rayon::run` uses per-call-site adaptive probing that is meaningless for a once-per-process init.
- Tests pin exact worker counts by configuring `Manual::new(Sequential, hint)` as the db's strategy — through the same public `init` as production, with no test-only entry points. `init_scale` sweeps by rewrapping its Rayon pool in a `Manual` hint.

## Trait-bound cleanup

- `storage::Context` (the pre-existing convenience alias) absorbs `Spawner + 'static`, collapsing every `E: Context + Spawner + 'static` in qmdb back to `E: Context` and converting ~60 pre-existing raw `Storage + Clock + Metrics` spellings (glue, examples, consensus, fuzz) to the alias. The one workspace type affected is the `SyncFaultContext` test mock, which gains a small `Spawner` delegation.
- The `Journal<...>: Send + Sync + 'static` where-clauses reduce to `+ 'static` on the existing `J`/`C` bounds (`Contiguous` already carries `Send + Sync` with `Item: Send`).
- `build_snapshot` asks only for what it uses: `E: Spawner`, `C: Contiguous<Item: Operation<F>> + 'static`.
- The `Value = Location<F>` binding restated at init sites is dropped (implied by `SnapshotBuild`'s supertrait).
- `futures::future::BoxFuture` replaces the hand-rolled `Pin<Box<dyn Future + Send + 'a>>` (the boxed return itself is load-bearing: the opaque-return form was already reverted for MSRV in 5f5a3f1b7, and `#[boxed]` only heap-allocates an `async fn` body without changing the signature). A `RoutedBatch<K>` alias removes the `clippy::type_complexity` allows.

## Metrics simplification

- `spills` is a plain `Counter` (detached on build workers, folded at install exactly like `pruned`), deleting the `Option` special-casing. This also restores serial parity: the previous install-time count missed spill-then-despill events, so the metric's value depended on how init parallelized.
- `install_range` folds `keys`/`items` into the full index via `Gauge::inc_by` and returns just the worker's item count; `set_metrics` is deleted.

## Test strengthening

- Both equivalence tests now assert `get()` for every key after each reopen. The `any` root is a pure function of the log, so the previous root-only assertion could not detect a location filed under the wrong key.
- The `any` test adds a delete-then-reinsert commit; the `current` test prunes before its reopen sweep, pinning the grafted root over a pruned bitmap prefix.
- New coverage: a fresh single-commit db reopened with workers (keyless replay), a variable-value partitioned equivalence test, and derived-worker-count reopens via `Manual` hints in every sweep.

## Docs

`new_range` now names the exact offset-aware surface (`get_mut`/`get_mut_or_insert` and their cursors) instead of overclaiming all `Unordered` operations, and the new doc text is swept for style (sentence splits, comment placement).

### DIFF
--- a/storage/src/lib.rs
+++ b/storage/src/lib.rs
@@ -111,16 +111,25 @@ commonware_macros::stability_scope!(BETA, cfg(feature = "std") {
         }
     }
 
-    /// A runtime context providing storage, timing, and metrics capabilities.
+    /// A runtime context providing storage, timing, metrics, and task-spawning capabilities.
     ///
     /// This is a convenience alias for the trait bound
-    /// `Storage + Clock + Metrics` that appears on nearly every type in this crate.
+    /// `Storage + Clock + Metrics + Spawner + 'static` that appears on nearly every type in this
+    /// crate.
     pub trait Context:
-        commonware_runtime::Storage + commonware_runtime::Clock + commonware_runtime::Metrics
+        commonware_runtime::Storage
+        + commonware_runtime::Clock
+        + commonware_runtime::Metrics
+        + commonware_runtime::Spawner
+        + 'static
     {
     }
     impl<
-            T: commonware_runtime::Storage + commonware_runtime::Clock + commonware_runtime::Metrics,
+            T: commonware_runtime::Storage
+                + commonware_runtime::Clock
+                + commonware_runtime::Metrics
+                + commonware_runtime::Spawner
+                + 'static,
         > Context for T
     {
     }
--- a/consensus/src/marshal/store.rs
+++ b/consensus/src/marshal/store.rs
@@ -2,7 +2,7 @@
 
 use crate::{simplex::types::Finalization, types::Height, Block};
 use commonware_cryptography::{certificate::Scheme, Digest, Digestible};
-use commonware_runtime::{BufferPooler, Clock, Handle, Metrics, Storage};
+use commonware_runtime::{BufferPooler, Handle};
 use commonware_storage::{
     archive::{self, immutable, prunable, Archive, Identifier},
     translator::Translator,
@@ -246,7 +246,7 @@ pub trait Blocks: Send + Sync + 'static {
 
 impl<E, B, C, S> Certificates for immutable::Archive<E, B, Finalization<S, C>>
 where
-    E: BufferPooler + Storage + Metrics + Clock,
+    E: BufferPooler + commonware_storage::Context,
     B: Digest,
     C: Digest,
     S: Scheme,
@@ -297,7 +297,7 @@ where
 
 impl<E, B> Blocks for immutable::Archive<E, B::Digest, B>
 where
-    E: BufferPooler + Storage + Metrics + Clock,
+    E: BufferPooler + commonware_storage::Context,
     B: Block,
 {
     type Block = B;
@@ -347,7 +347,7 @@ where
 impl<T, E, B, C, S> Certificates for prunable::Archive<T, E, B, Finalization<S, C>>
 where
     T: Translator,
-    E: BufferPooler + Storage + Metrics + Clock,
+    E: BufferPooler + commonware_storage::Context,
     B: Digest,
     C: Digest,
     S: Scheme,
@@ -398,7 +398,7 @@ where
 impl<T, E, B> Blocks for prunable::Archive<T, E, B::Digest, B>
 where
     T: Translator,
-    E: BufferPooler + Storage + Metrics + Clock,
+    E: BufferPooler + commonware_storage::Context,
     B: Block,
 {
     type Block = B;
--- a/examples/reshare/src/dkg/state.rs
+++ b/examples/reshare/src/dkg/state.rs
@@ -26,9 +26,7 @@ use commonware_cryptography::{
     BatchVerifier, PublicKey, Signer,
 };
 use commonware_parallel::Strategy;
-use commonware_runtime::{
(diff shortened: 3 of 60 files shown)


storage/src/lib.rs
```diff
@@ -111,16 +111,25 @@ commonware_macros::stability_scope!(BETA, cfg(feature = "std") {
...
     }
 
-    /// A runtime context providing storage, timing, and metrics capabilities.
+    /// A runtime context providing storage, timing, metrics, and task-spawning capabilities.
     ///
     /// This is a convenience alias for the trait bound
-    /// `Storage + Clock + Metrics` that appears on nearly every type in this crate.
+    /// `Storage + Clock + Metrics + Spawner + 'static` that appears on nearly every type in this
+    /// crate.
     pub trait Context:
```
patrick-ogrady, line 119: I'm not sure we need this alias (don't have it in other crates) but ok to defer on that.

roberto-bayardo:
Thanks for the sweep — most of this I'll take as-is: the `handle??` join fix (nice catch, and the deterministic-runtime reasoning for why no test saw it is convincing), the bound cleanup via `storage::Context`, `BoxFuture`, the spills-counter parity fix, and the stronger equivalence assertions.

I want to push back on deleting `InitParallelism`, though. The "derive from the strategy at the point of use" idiom doesn't transfer to init: both existing hint consumers (`any/db.rs` get-many chunking, `any/batch.rs` floor classification) use the hint to size chunks for synchronous CPU work that the strategy itself executes via `map_collect_vec`. The init build is a different domain: workers are `context.spawn`ed async tasks doing async collision-resolution log reads, fed by an async replay/routing task over channels. The strategy executes none of that work — we'd only be borrowing its number, and it's the wrong number:

- The routing task becomes the bottleneck past a small worker count (measured plateau ~2 on uniform 50M), so a Rayon strategy on a 32-core box derives 16x more workers than useful.
- Over-derivation actively hurts: `per_worker_cache = cache_size / workers`, so 32 workers each get 1/32nd of the location→key cache instead of 1/2, degrading exactly the reads the cache exists to avoid.
- With the knob gone, the only way to pin init workers is wrapping the db's whole strategy in `Manual(S, n)` — but that hint also feeds get-many and batch classification, so capping init to 2 throttles the CPU paths to 2. "Rayon(16) for compute, 2 init workers" becomes inexpressible.

Two of your complaints against the knob are legitimate and I'd rather fix them directly:

- The state-sync paths hardcoding `Auto` over a configured `Serial` is a real bug — let's thread `cfg.init_parallelism` through `init_from_log` instead.
- `Auto` deriving `hint - 1` can over-provision; happy to cap it (e.g. `min(hint, 4)`) given the measured plateau, so `Auto` stays sane regardless of strategy.

The per-call-site `init_parallelism: Serial` churn is real but shape-independent (any field costs the same until the configs grow defaults), so I don't think it argues for deletion.

We'll fold the keepers plus those two fixes into the branch locally.

roberto-bayardo:
Folded into `qmdb-async-parallel-init` as 585174253 (your commit, cherry-picked as-is) + 076fe5ebf (the rework per the thread above: `InitParallelism` retained, `Auto` capped, config threaded through the state-sync build paths, plus a regression test that a `Serial` config spawns no snapshot workers under a parallel-hint strategy). Thanks for the sweep — the join fix, bound cleanup, metrics parity, and test strengthening all landed as-is.

PR #4176, opened 2026-07-06
[runtime] Shard the page cache to reduce lock contention

roberto-bayardo:
The page cache serializes every reader and every page fault on one `RwLock`. This shards it: pages are partitioned across up to 16 shards, each with its own `Clock`, slot hints, and in-flight fetch map, so concurrent readers of unrelated pages take different locks and single-flight fetch dedup is preserved per page.

This is a fresh port of the earlier `ClockCache` sharding onto the current paged `CacheRef` (the `PageCache` trait it was written against was removed by #4170). The design is unchanged; the guard now borrows its shard's lock, matching #4678.

## Design

- Up to 16 shards (a power of two, floored so each shard holds at least one 8-page block; caches smaller than one block keep a single shard and behave exactly as before).
- Blocks of 8 consecutive pages map to one shard, so sequential reads and flush publishes re-acquire only at block boundaries while random access spreads across every shard. Round-robin routing was measured earlier to regress long sequential scans by ~4% (one lock hop per page); blocking removed that.
- Shard locks are leaf locks: every wait happens with no shard lock held. Guards are dropped before acquiring another shard, and `read_cached_many` serves each shard under one acquisition, finishing block-crossing ranges through the single-read path after releasing it. Reviewers should scrutinize the guard hand-off sites.
- Eviction is per shard (CLOCK within each shard), the standard sharded-cache approximation.

## Why now: snapshot-build init stops scaling on the single lock

Init of a 1B-key Ethereum seed (451GB, `current` variant) on a 32-core EPYC 9354P with NVMe, 32GB init cache, 32GiB page cache, mimalloc, cold OS cache. Every insert worker resolves translated-key collisions with probe reads through the page cache. On `main`, per-worker voluntary context switches run at ~21k/s and the futex stacks resolve to `RawRwLock::lock_shared_slow` / `unlock_exclusive_slow`; at 32 threads the lock's slow paths plus the futex/scheduler churn they cause account for about a fifth of all cycles (`perf`: `lock_shared_slow` is the top symbol at 5.9%, `WordLock::lock_slow` 2.7%, kernel scheduler ~10%), and per-op instructions rise 44% with no additional ops.

Init time (single runs; peak RSS 136GB in every cell):

| `init_concurrency` | main | sharded |
|---|---:|---:|
| 16 | 331s (328-333 across three runs) | **304s (-8%)** |
| 32 | 356s | **282s (-21%)** |

On `main` 32 workers are slower than 16; sharded, 32 workers beat `main`'s best by 15%.

## Earlier layer and end-to-end results

Measured on the original `ClockCache` implementation of the same design (2026-07); the port keeps the routing and locking scheme, so these are expected to carry over but have not been re-run on this branch.

Synthetic cache-layer harness (M4 Pro, Mops/s, before -> after): point reads at 14 threads 4.2 -> 66.0; point reads with a continuous writer 4.2 -> 58.6; batched x32 at 14 threads 92.2 -> 95.4. Sharded `read_many` scales with batch size where the single lock cannot (108 / 192 / 225 Mops/s at batches of 32 / 256 / 1024, 10 threads).

Real databases: macOS 50M-key / 9.8GB DB, 4GB cache, cache-cold concurrent point reads +17-19% at 8-32 readers, warm passes unchanged (the page cache stops being the limiting component of `get`). Linux 10M-key / 2.1GB DB, 256MB churning cache: cold 32 readers +9-12%, warm 8 readers +12%, warm 32 readers +8%.

Known cost from that round, to be re-measured here: with the working set fully cache-resident and throughput capped upstream, point reads paid a fixed ~4% (~15-20ns of shard routing and guard bookkeeping; identical at 16 and 64 shards, so not contention). `read_many` was faster than the single lock in every configuration measured.

## Validation

`utils::buffer::paged` runtime suite (including a multi-shard block-boundary read test) and the `journal::contiguous` storage suite.

https://claude.ai/code/session_01Q4BFdANo2LhVEr2ko3FK9n

### DIFF
--- a/runtime/src/utils/buffer/paged/cache.rs
+++ b/runtime/src/utils/buffer/paged/cache.rs
@@ -50,7 +50,7 @@ struct PageFetchEntry {
 ///
 /// Must be a power of two. The actual shard count is reduced for small caches so every shard
 /// holds at least one page.
-const MAX_SHARDS: usize = 16;
+const MAX_SHARDS: usize = 64;
 
 /// Number of consecutive pages that map to the same shard.
 ///
@@ -200,9 +200,11 @@ impl ClockCache {
     pub fn new(pool: BufferPool, page_size: NonZeroU16, capacity: NonZeroUsize) -> Self {
         let capacity = capacity.get();
 
-        // Largest power of two <= min(MAX_SHARDS, capacity), so every shard gets at least one
-        // page and the shard index is a mask.
-        let shard_count = 1usize << MAX_SHARDS.min(capacity).ilog2();
+        // Largest power of two <= min(MAX_SHARDS, capacity / SHARD_BLOCK_PAGES), so every
+        // shard can hold at least one full block of consecutive pages and the shard index is
+        // a mask. Caches smaller than a block use a single shard.
+        let max_shards = MAX_SHARDS.min(capacity / SHARD_BLOCK_PAGES as usize).max(1);
+        let shard_count = 1usize << max_shards.ilog2();
         let base = capacity / shard_count;
         let rem = capacity % shard_count;
         let shards: Arc<[RwLock<Cache>]> = (0..shard_count)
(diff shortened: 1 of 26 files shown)


roberto-bayardo:
Parked pending #4170: the sharding here is written against the PageCache-trait design that #4170 walked back, so it needs re-porting onto the final Option-based CacheRef (and re-benchmarking) once that merges.

PR #4179, opened 2026-07-06
[runtime] [storage] Dispatch page-cache retention via an internal enum instead of a PageCache trait

danlaine:
## Summary

The base branch lets `CacheRef` skip in-process caching, so the `scale` bench's `get`/`get_many` modes measure cold reads honestly (and deployments can avoid double-caching under the OS page cache). It delivers this with a `PageCache` trait, threading `P: PageCache` through `Writer`/`Sealed`/`View`, the contiguous journals, and every qmdb config and `Db` alias -- 21 files.

But none of those types use `P`; they only store or pass a `CacheRef<P>`. So this PR hides the choice inside a non-generic `CacheRef` instead:

```rust
enum CacheImpl {
    Clock(Arc<RwLock<Cache>>), // the existing bounded cache with fetch dedup
    None,                      // retains nothing; every read fetches from the blob
}
```

All 20 downstream files revert to `main` byte for byte, and the clock path provably keeps its behavior because its code does not move: each read/insert/invalidate method takes a `let ... else` guard, and the fault path gains an early-exit arm that fetches straight from the blob. `CacheRef::new` and every existing caller are untouched; the new mode is `CacheRef::no_cache[_from_pooler](pool, page_size)`.

## Costs

- **Dispatch:** a predicted two-variant branch, ahead of an `RwLock` acquire + page memcpy (hit) or blob I/O (miss). The warm pass of `scale -- get` can A/B this against the trait version if we want proof.
- **Allocation parity:** the no-cache fault still wraps its fetch in `Shared`, because `Contiguous::read` requires `Sync` futures and blob reads are only `Send`. (That bound is why the trait boxed every fetch.)
- **Closed set:** external `PageCache` impls can no longer plug in. A new retention policy is an enum variant; if out-of-tree impls ever matter, a `Custom(Arc<dyn ...>)` variant fits without signature changes.

### DIFF
--- a/runtime/src/utils/buffer/paged/cache.rs
+++ b/runtime/src/utils/buffer/paged/cache.rs
@@ -20,10 +20,7 @@ use tracing::{error, trace};
 
 /// Shared future for one logical page fetch. The output uses `Arc<Error>` because `Shared`
 /// requires cloneable results. The `IoBuf` contains only the logical, validated page bytes.
-///
-/// `Shared` also makes the future `Sync` and cloneable, so a caching implementation can hand one
-/// fetch generation to every concurrent reader of the same page.
-pub type PageFetchFuture = Shared<Pin<Box<dyn Future<Output = Result<IoBuf, Arc<Error>>> + Send>>>;
+type PageFetchFuture = Shared<Pin<Box<dyn Future<Output = Result<IoBuf, Arc<Error>>> + Send>>>;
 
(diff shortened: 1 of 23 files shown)


roberto-bayardo:
Is this just to avoid the clutter? (I don't think the clutter is that bad.)

roberto-bayardo:
plus kind of awkward to have to update an enum for every new page cache implementation.

PR #4181, opened 2026-07-06
[storage/journal] Read `read_many` per-blob groups concurrently

roberto-bayardo:
## Summary

Both contiguous journals batch their `read_many` I/O less than they appear to. The **fixed**
journal groups its (sorted) positions by blob and reads each blob's batch with one call — but
awaits those group reads **sequentially**: for example, for 1M items per blob, a uniform
256-key batch over a multi-blob region scatters into ~64 groups of ~4 keys, so a single
`get_many` call runs at an effective I/O depth of ~3. The **variable** journal is worse: after
its offsets lookup, it awaits each *consecutive run* of positions sequentially, and uniform
access makes nearly every run a single item — one serialized I/O per cache miss, so batching
bought almost nothing over point gets.

This PR makes both concurrent. Fixed (first commit): the output buffer is already pre-sized
for every position, so each blob group takes a disjoint slice (`split_at_mut`) and all groups
run under one `try_join_all` — the whole batch becomes a single I/O wave. Variable (second
commit): run descriptors are collected first, then all runs — across and within blobs — are
joined the same way. Decoding and result ordering are unchanged; blob reads are `&self`, so no
locking changes.

## Measured

Cold uniform-random `get_many` (256-key batches) through the full qmdb stack, 50M-key /
250M-update database, Linux (`x86_64`, AMD EPYC 9354P, ext4 on md-RAID NVMe), OS cache dropped
per pass, benchmark from #4170:

Fixed journal (first commit):

| cold | before | after | speedup |
|---|---|---|---|
| 16 KiB pages, 1 caller | 80.4 us/get (12.4K gets/s) | 9.5 us/get (105K gets/s) | **8.5x** |
| 4 KiB pages, 1 caller | 44.8 us/get (22.3K gets/s) | 6.7 us/get (151K gets/s) | **6.7x** |
| 16 KiB pages, 8 callers | 12.2 us/get | 5.0 us/get | 2.4x |

Variable journal (second commit; digest values, same database shape):

| cold | before | after | speedup |
|---|---|---|---|
| 16 KiB pages, 1 caller | 265.6 us/get (3.8K gets/s) | 18.9 us/get (52.9K gets/s) | **14.0x** |
| 4 KiB pages, 1 caller | 187.4 us/get (5.3K gets/s) | 17.2 us/get (58.1K gets/s) | **10.9x** |
| 16 KiB pages, 8 callers | 33.5 us/get | 13.7 us/get | 2.4x |

Before the fix, a batched variable-journal `get_many` (265.6 us/get) was barely faster than
one-at-a-time point gets (298.5 us/get): run-level serialization issued one awaited I/O per
cache miss, so batching bought almost nothing.

The single-caller case is the production commit-path shape (`get_many`-driven load phase); the
fixed journal moves from ~22x below device capability to within ~2x of it. Results are
byte-identical before and after on both journal kinds.

## Notes

- **Conflicts with, but is complementary to, #4144**: that PR restructures the same function
  into cache/miss passes (`read_many_impl`) while keeping the sequential per-blob loop; this
  change parallelizes whatever reads survive to the blob stage. Whichever lands second ports
  cleanly — the transform (disjoint buffer slices + joined group reads) applies the same way to
  the refactored shape. Draft until sequencing is settled.
- At already-saturated external concurrency (32 callers), unbounded fan-out shows a small
  regression (~13% cold at 4 KiB pages) from blocking-pool contention; a follow-up will sweep a
  concurrency bound (`buffer_unordered(N)`) before this leaves draft.
- The variable journal's `read_many` gets the same treatment (second commit): its data reads
  awaited one *consecutive run* at a time, and uniform access makes most runs a single item —
  one serialized I/O per cache miss. Runs are now collected and joined concurrently. (Its
  offsets lookup is a fixed journal, so it also inherits the first commit's fix.)
- Within-blob misses were already concurrent for the fixed journal (the paged view issues them
  via `FuturesUnordered`), so single-blob batches were never affected; the pathology was
  many-blobs-few-keys, which uniform access over a multi-blob region maximizes.
- Reproduction needs the `get_many` mode of the `scale` benchmark from #4170 (variable-journal
  support via its `kind` argument).

### DIFF
--- a/storage/src/journal/contiguous/fixed.rs
+++ b/storage/src/journal/contiguous/fixed.rs
@@ -1429,6 +1429,16 @@ mod tests {
         format!("{}-blobs", cfg.partition)
     }
 
+    /// Extract a metric counter's value from encoded metrics output.
+    fn counter(buffer: &str, name: &str) -> u64 {
+        buffer
+            .lines()
+            .find(|l| l.contains(name) && !l.starts_with('#'))
+            .and_then(|l| l.split_whitespace().last())
+            .and_then(|v| v.parse().ok())
+            .expect("counter missing")
+    }
+
     impl<E: crate::Context, A: CodecFixedShared> Journal<E, A> {
         /// Test helper: Get the oldest blob from the blob store.
         pub(crate) const fn test_oldest_blob(&self) -> Option<u64> {
@@ -4845,15 +4855,6 @@ mod tests {
 
             let reader = journal.snapshot().await.unwrap();
 
-            fn counter(buffer: &str, name: &str) -> u64 {
-                buffer
-                    .lines()
-                    .find(|l| l.contains(name) && !l.starts_with('#'))
-                    .and_then(|l| l.split_whitespace().last())
-                    .and_then(|v| v.parse().ok())
-                    .expect("counter missing")
-            }
-
             // Sparse subset spanning multiple blobs, including the pruning boundary.
             // `try_read_sync` probes do not populate the cache, so the cached subset is
             // whatever the append path left resident; derive the expected hit count from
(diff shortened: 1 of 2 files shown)


roberto-bayardo:
keeping in draft until I confirm the same change's impact on variable journal

storage/src/journal/contiguous/fixed.rs
```diff
@@ -4885,6 +4896,72 @@ mod tests {
...
+            let before = context.encode();
+            let batch = reader.read_many(&positions).await.unwrap();
+            let after = context.encode();
+            assert_eq!(
+                counter(&after, "second_cache_misses") - counter(&before, "second_cache_misses"),
+                positions.len() as u64
+            );
+            assert_eq!(
+                counter(&after, "second_cache_hits"),
+                counter(&before, "second_cache_hits")
```
patrick-ogrady, line 4941: In #4144, we improve metrics tracking (variable had a few inconsistencies that aren't worth bloating this PR for).

patrick-ogrady (approved):
all-time LOC vs performance boost lol

PR #4184, opened 2026-07-07
[runtime] Minimal V1 blob layout

### LINKED ISSUE #4210: [runtime] Blob left torn by an interrupted creation fails loud on reopen
## Problem

Creating a blob writes its header region and fsyncs it before `open` returns. If the process crashes after the file is created but before the header fsync completes, the file can be left with a partially written (torn) header: writes tear at arbitrary byte boundaries, and the kernel may flush any subset of the dirty pages before the crash, so any prefix/subset of the header bytes may be on disk.

On the next open, header parsing fails (`BlobCorrupt`: invalid magic, truncated header, etc.) even though the blob never held synced data and nothing was promised to the caller. Any journal or store whose init opens that blob is then stuck until an operator deletes the file by hand.

This is pre-existing behavior (V0 blobs have the same window on their 8-byte prelude) and was reported by qed-audit. #4184 narrows it - creation now truncates to zero before writing the header region, so a torn *initial* write that flushes in order stays shorter than the prelude and is recreated on reopen - but out-of-order page flushes can still land bytes past the prelude, and those states remain loud errors.

## Constraints on a fix

- Writes may tear at any point; no filesystem- or device-specific ordering assumptions.
- The fix should not add fsyncs to blob creation - the goal is the opposite (#4190).
- Recovery must not silently discard data from a blob that was ever synced: a file that held promised data and now has a bad header is genuine corruption and must stay loud.

## Current plan: V1-only canonical-content classification

Scope recovery to V1 blobs (V0 torn creations from old binaries stay loud, as today). A V1 header
region is fully deterministic: 16 known bytes (magic, runtime version, blob version, data offset,
with the CRC sealing the first 12 into a fixed value) followed by zeros to the data offset. A torn
creation can only persist a writeback-subset of that image, so the classifier is:

heal iff the file is no longer than the header region, the header does not parse, every nonzero
byte in the first 16 positions matches one canonical image (enumerate accepted blob versions x
plausible data offsets, all against the same image), and every byte after 16 is zero. Everything
else stays loud.

Properties:
- No format change, no added fsyncs, no migration, no rollback hazard; a few dozen lines in `Header`.
- Legacy V0 blobs are protected, not sacrificed: real payload past byte 8 is non-canonical, so it
  fails the subset test and stays loud (closes the wipe hazard that killed the #4189 size fence).
- Residual risk is only corruption that coincidentally reproduces canonical bytes with zeros
  elsewhere, which is outside the crash-tearing fault model. Quarantine (move aside instead of
  truncate) can back-stop even that.
- Scope limit: this is a classifier, not a fsync remover. It is complete exactly when data cannot
  reach disk before the header is durable (today's creation fsync, or a delayed creation fsync
  whose handle gates the first data write, per #4180). It does not enable zero-fsync creation on
  its own: with no creation fsync, a torn creation with flushed data past the region is
  indistinguishable from rot of a synced blob and must stay loud.

### Prerequisite: distinct V1 magic (must land in #4184 before merge)

Adversarial review found a hole that would silently bypass any "unparseable header" trigger:
V0 and V1 share magic `CWIC` and differ only in `runtime_version` (0 vs 1, big-endian). A torn V1
creation that persists the magic and blob version but loses byte 5 (the 0x01) parses as a VALID V0
header and is adopted at data offset 8, exposing up to 4088 bytes of extension/CRC/pad garbage as
consumer data with V0 geometry. Since zeroing bytes can only turn nonzero to zero, the fix is a V1
magic whose bytes cannot become the V0 magic by zeroing (e.g. `CWI1`): no writeback-subset of a V1
image can then parse as V0, and the classifier's trigger is complete. This is free while V1 exists
only on the unmerged #4184 branch; it is a breaking format change the moment V1 blobs ship.

## Directions explored (see closed PRs for details)

- **Heal fence** (#4189): treat any file no larger than the header region whose bytes do not parse as an interrupted creation and recreate it. Sound for V1 writers (data starts past the region) but a small legacy V0 blob with committed payload and a corrupted prelude is indistinguishable and would be wiped instead of failing loud.
- **Defer creation durability to the first sync** (#4193, #4194): removes creation fsyncs but widens the unclassifiable window (large file, no valid header) to the whole pre-sync lifetime.
- **Delayed creation fsync**: start the header fsync at open and return immediately; order the blob's first data write (not just later fsyncs) behind it. Needs the submission-ordering guarantees from #4180.
- **Deferred claim**: create the blob under a staging name and rename it to its real name during the first sync (after the file fsync, before the dir fsync). Unsynced blobs are then self-identifying by name - staging files can always be recreated, and a real-name file with a bad header is always genuine corruption - eliminating content/size classification entirely.

Related: #4190, #4180.

roberto-bayardo:
Resolves: #4210
Towards: #4188

## Summary

Adds the minimal V1 blob container layout for runtime storage.

New blobs are created with a V1 header region: an 8-byte prelude (magic, runtime
version, blob version), a CRC32 over the prelude, and zero padding out to a
4096-byte data offset. Existing V0 blobs remain readable and writable in place, and
the public `open_versioned` return shape is unchanged: `(Blob, logical_size,
blob_version)`.

The 4096-byte data offset aligns blob data to a storage-page boundary; combined with
the new `paged::page_size` helper (which sizes a logical page so `logical + 12-byte
CRC record` is a power of two), physical pages either fit within a 4096-byte storage
page or start on a 4096-byte boundary and span whole storage pages, avoiding
boundary-straddling cold reads. `CacheRef::new` warns on a non-aligning page size.
Alignment is a performance property only -- any page size stays correct.

## Recovery

Blob creation is eager: `open_versioned` writes the full V1 header page, fsyncs the
file, and syncs containing directories on Unix before returning, so a created blob and
its directory entry are durable on return.

If creation crashes before the header is durable, reopen classifies only conservative
torn-creation remnants as recoverable -- a canonical V1 prefix followed by zeros, with
no bytes past the 4096-byte header region -- and recreates them as new; all other
malformed states stay `BlobCorrupt`. The prefix shape is a crash *model*, not a
filesystem guarantee: a remnant persisted out of order (a lost byte followed by a
persisted one) is not a canonical prefix and stays loud rather than healing -- safe,
since such a file holds no data.

V1 reopen also verifies the reserved padding is zero: a valid prelude+CRC with nonzero
padding is rejected rather than silently accepted. This makes a data-bearing blob with
corruption confined to the reserved region fail loud rather than read -- a deliberate
fail-fast choice that also keeps the reserved space pristine for future layouts.

The recovery guarantee is deliberately narrow:

- V1 blobs that contain data are longer than the header region and are never recreated
  as torn creations, so recreation never erases data.
- Unparseable V1 headers heal only if the bytes match the canonical-prefix model.
- Legacy V0 blobs remain supported; the only documented residual is a degenerate
  all-zero V0 remnant byte-identical to a torn V1 creation prefix (it heals, losing
  only all-zero payload).

## Compatibility

- Backward: existing V0 blobs open and read/write in place; new blobs are V1.
- Forward: this is a one-way upgrade. Once V1 blobs are written, a binary built before
  this change reads them as `BlobCorrupt` (it recognizes only the V0 magic). Roll out
  as a commit point.

## Costs

- Every blob occupies at least one 4096-byte header page (vs. an 8-byte V0 header), so
  many-small-blob partitions pay more space.
- Creation writes and fsyncs a 4096-byte page instead of 8 bytes; reopen reads that one
  aligned page to validate the CRC and padding.

## Performance

The aligned data offset removes the sub-page straddle a misaligned data region forces
on cold reads. Measured on the QMDB `scale` point-read benchmark (ordered `any` DB,
50M-key space, 250M updates, 4084-byte logical page so `logical + 12-byte CRC record` =
a 4096-byte physical page, 64-page cache; AMD EPYC 9354P, Linux, buffered I/O),
comparing databases built from the same operation stream -- one with V0 headers, one
with V1 -- reading the identical key sample:

| readers | cold V0 (8-byte header) | cold V1 (4096-byte header) | warm V0 | warm V1 |
|--:|--:|--:|--:|--:|
| 1 | 142.8 us/get | 92.1 us/get (1.55x, +55% gets/s) | 21.5 us | 22.8 us |
| 4 | 32.5 us/get | 22.4 us/get (1.45x) | 4.4 us | 4.5 us |
| 8 | 15.6 us/get | 11.3 us/get (1.38x) | 4.2 us | 4.2 us |

With V0's 8-byte header every physical page straddles a 4096-byte filesystem block, so
each cold page read fetches two blocks; V1's aligned offset makes it one. Single-reader
cold reads are 1.55x faster (142.8 -> 92.1 us). The gain tapers as concurrency amortizes
the extra block fetches across in-flight I/O, and disappears on warm reads, where the
page is already cached and alignment is irrelevant. The effect scales with the physical
page relative to the 4096-byte block: at a 16 KiB page (5 straddled blocks vs 4) the same
comparison measured only ~0-4%, which is why the aligned offset pairs with a ~4096-byte
page.

## Validation

- `just test -p commonware-runtime`
- `just test -p commonware-storage` (storage consumers exercise the new data offset)
- `just clippy -p commonware-runtime`
- `just check-fmt`
- `just test-conformance -p commonware-storage`
- io_uring build/tests on Linux (the io_uring arm cannot build on macOS; run via CI or a Linux host)


runtime/src/lib.rs
```diff
@@ -77,6 +77,48 @@ stability_scope!(BETA {
     /// Default [`Blob`] version used when no version is specified via [`Storage::open`].
     pub const DEFAULT_BLOB_VERSION: u16 = 0;
 
+    /// Version of a [`Blob`]'s header layout, chosen at creation.
+    ///
+    /// The version applies only when a blob does not yet exist: reopening an existing blob always
+    /// honors the version recorded in its header, regardless of the version requested.
+    #[derive(Clone, Copy, Debug, PartialEq, Eq)]
```
patrick-ogrady: This is way cleaner than I feared.

roberto-bayardo:
bugbot run

runtime/src/lib.rs
roberto-bayardo: yeah it's not bad at all really.

runtime/src/utils/buffer/paged/cache.rs
```diff
@@ -140,10 +142,22 @@ pub struct CacheRef {
...
+        if !physical_page_size.is_multiple_of(STORAGE_PAGE_SIZE)
+            && !STORAGE_PAGE_SIZE.is_multiple_of(physical_page_size)
+        {
+            warn!(
+                page_size = page_size.get(),
+                physical_page_size,
+                "page size produces physical pages that do not align with storage pages; pick a \
+                 page size via paged::page_size to avoid amplifying cold random reads"
+            );
+        }
```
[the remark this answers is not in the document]
roberto-bayardo: Starting at a non-4096-aligned offset does not imply straddling: a `d`-byte physical page starting at `4096 + k*d` where `d` divides 4096 always lies entirely within one storage page — the span `[start, start + d)` has `start % 4096` a multiple of `d`, so `start % 4096 + d <= 4096` holds for every `k` (a 2048-byte page at `4096m + 2048` ends exactly at the next boundary, not across it). Verified exhaustively: for every divisor of 4096, zero boundary crossings over 50K consecutive pages; for a non-divisor below 4096 (e.g. 3000, which this check does warn on), ~73% of pages straddle. Sub-page sizes that divide 4096 cost one storage-page read per cold page fetch — the same as an aligned fetch — so the warning staying silent for them is correct.

runtime/src/storage/tokio/mod.rs
```diff
@@ -142,26 +144,64 @@ impl crate::Storage for Storage {
...
+            match parsed {
+                ParsedHeader::V0 { blob_version } => {
+                    let info = BlobInfo {
+                        size: len - Header::SIZE_U64,
+                        blob_version,
+                        layout: BlobLayout::V0,
+                    };
+                    (info, Header::SIZE_U64)
+                }
+                ParsedHeader::NeedsExtension { blob_version } => {
```
patrick-ogrady: Anything else worth putting in V1 while we have it?

cc @danlaine

runtime/src/storage/tokio/mod.rs
```diff
@@ -142,26 +144,64 @@ impl crate::Storage for Storage {
...
-            file.read_exact(&mut header_bytes)
+            // Existing blob - read and validate the header, honoring the layout it records
+            let mut prelude = [0u8; Header::SIZE];
+            file.read_exact(&mut prelude)
                 .await
                 .map_err(|_| Error::ReadFailed)?;
-            Header::from(header_bytes, len, &versions).map_err(|e| e.into_error(partition, name))?
+            let parsed = Header::parse_prelude(prelude, &versions)
+                .map_err(|e| e.into_error(partition, name))?;
+            match parsed {
```
patrick-ogrady: I had originally intended external applications manage these versions (not runtime). If we go this route, we probably want to have a runtime version and application version or remove the idea of application version altogether?

roberto-bayardo: The split you are describing is already how it works — the prelude has carried two independent u16s since #2694:

| bytes | field | owner | before this PR | after this PR |
|---|---|---|---|---|
| 4-5 | `runtime_version` | runtime | always written 0, never read for any decision | 0 = 8-byte header, 1 = page-aligned header; validated internally against `SUPPORTED_RUNTIME_VERSIONS = 0..=1` |
| 6-7 | `blob_version` | application | managed via `open_versioned(versions: X..=Y)` | completely unchanged |

The application lifecycle is untouched: create still writes the newest requested version, open still rejects a stored version outside the caller's range, and the stored value still comes back in `BlobInfo` — exactly as you designed it. Applications cannot see or set `runtime_version`, and the two fields never interact: an application at blob_version N cannot tell whether its blob header is V0 or V1, since everything it sees is logical offsets. So applications keep managing their versions; the runtime now manages the layout through the field that was already reserved for it.

roberto-bayardo: Follow-up: renamed `BlobLayout` -> `BlobHeaderLayout` to reduce this confusion — the V0/V1 values describe the blob header region layout (`runtime_version`), and the original name read as "blob version" while sitting next to the application-owned `blob_version` field.

patrick-ogrady: lol...this level of foresight is unprecedented

runtime/src/utils/buffer/paged/mod.rs
```diff
@@ -3,14 +3,25 @@
...
 //! | len1 (2 bytes) |  crc1 (4 bytes) | len2 (2 bytes) | crc2 (4 bytes) |
 //! ```
 //!
+//! Throughout this module, an unqualified page size always denotes the logical size (matching
+//! the configured value); only physical sizes carry a qualified `physical_page_size` name.
+//!
+//! # Storage-page alignment
+//!
+//! Physical page `p` begins at blob offset `p * physical_page_size`, and a blob created with
+//! [crate::BlobHeaderLayout::V1] begins its data on a 4096-byte boundary. Choosing a logical page size
```
patrick-ogrady: > 4096-byte boundary

Is it correct to hardcode this? doesn't this depend on the disk/settings of page cache to work right?

roberto-bayardo: The 4096 is only hard-coded on the *write* side — the format itself commits to a range, not a value: the parser accepts any power-of-two data offset in `4096..=1MiB`, and each blob stores its own offset (CRC-protected). So if storage moves to 16K-native units, we change `V1_DATA_OFFSET` to 16384 and binaries shipped today already read those blobs — no version bump or migration, same self-describing trick that lets V0/V1 mix. 4096 is the right write-side choice now because it is the common unit across ext4 blocks, the page cache, NVMe 4Kn, and O_DIRECT alignment (and implies 512-alignment for free). And correctness never depends on it: alignment is purely a perf property — on hardware where the true unit is larger, the benefit degrades toward the V0 status quo, never below it.

runtime/src/storage/tokio/mod.rs
danlaine: Not that I can think of

runtime/src/utils/buffer/paged/mod.rs
roberto-bayardo: Adding some comments to clarify this.

roberto-bayardo: pushed some additional clarity-improving changes around this. Bottom line is 4096 only puts a lower bound on the initial data offset. Other than that everything is flexible.

runtime/src/storage/mod.rs
```diff
@@ -178,36 +295,150 @@ stability_scope!(BETA {
...
+            raw_len: u64,
+            versions: &RangeInclusive<u16>,
+        ) -> Result<HeaderResolution, HeaderError> {
+            if Self::missing(raw_len) {
+                return Ok(HeaderResolution::Recreate);
+            }
+            match Self::try_parse(head, raw_len, versions) {
+                Ok(resolution) => Ok(resolution),
+                Err(e) if e.interrupted_creation(raw_len) => Ok(HeaderResolution::Recreate),
+                Err(e) => Err(e),
```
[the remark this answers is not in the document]
roberto-bayardo: This targets machinery from an earlier version of this branch (when it was stacked on #4189). The current head has no `Header::resolve`/`interrupted_creation` heal path: an invalid or torn prelude always fails loud with `BlobCorrupt` (there is a test asserting invalid magic returns `BlobCorrupt`), and the only recreate case is a file shorter than the 8-byte prelude, which cannot hold V0 payload (V0 data starts at byte 8). Torn-creation recovery is deferred to a follow-up (#4190).

roberto-bayardo:
bugbot run

roberto-bayardo:
Superseded by #4232, which retains the aligned V1 layout and replaces eager header persistence with the deferred two-phase commit design.

roberto-bayardo:
Re-opening since we may want to get this more minimal change in first.

runtime/src/lib.rs
```diff
@@ -78,6 +78,56 @@ stability_scope!(BETA {
...
+        ///
+        /// The layout tag lives in the magic rather than the runtime version field because V0
+        /// stamped that field as zero, and zeros are exactly what a torn header write leaves
+        /// behind. Tags are nonzero and distinct, so no layout's magic can be turned into
+        /// another's by zeroing bytes, and a torn write can never be misread as a complete
+        /// header of a different layout.
+        pub(crate) const fn magic(self) -> [u8; 4] {
+            match self {
+                Self::V0 => *b"CWIC", // Commonware Is CWIC
+                Self::V1 => *b"CWI1",
```
patrick-ogrady: I think we can retain the same magic?

runtime/src/lib.rs
```diff
@@ -78,6 +78,56 @@ stability_scope!(BETA {
     /// Default [`Blob`] version used when no version is specified via [`Storage::open`].
     pub const DEFAULT_BLOB_VERSION: u16 = 0;
```
patrick-ogrady, line 79: set to 1?

runtime/src/storage/mod.rs
```diff
@@ -162,83 +221,209 @@ stability_scope!(BETA {
...
+        /// followed by zeros: the magic and runtime version are fixed; the blob version
+        /// bytes continue the prefix with whatever value the writer chose; the CRC bytes
+        /// must be a prefix of the CRC over the preceding prelude, which can only have begun
+        /// persisting once the full prelude did; and everything past the prefix must be
+        /// zero.
+        ///
+        /// The prefix shape is a model, not a filesystem guarantee: device writeback before
+        /// the sync completes may persist bytes out of order. A file that is not a canonical
+        /// prefix -- a lost byte followed by persisted ones, a CRC that does not match its
+        /// own prelude -- stays loudly corrupt rather than healing, trading recovery
```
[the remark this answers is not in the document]
roberto-bayardo: Fixed in `cc91dfc82` — rephrased the five new `mod.rs` comment em-dashes (the `V1_DATA_OFFSET` / `interrupted_creation` docs and the residual-test comment) with commas and parentheses. The remaining `--` in `cache.rs` / `sealed.rs` are pre-existing (not touched by this PR), so left as-is.

runtime/src/storage/mod.rs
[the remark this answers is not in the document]
roberto-bayardo: This is a false positive — a V0 blob with committed data is never recreated. Two guards prevent it:

1. **Magic mismatch.** Recreation requires the bytes to match the canonical **V1** prefix `CWI1`. A real V0 blob's magic is `CWIC` (byte 3 is `'C'` 0x43, vs `'1'` 0x31), so it never matches: it parses as V0, or fails loud (`VersionMismatch` / `UnsupportedRuntimeVersion`). It only reaches `interrupted_creation` at all if its magic is already corrupted.
2. **Nonzero-past-prefix rejection.** `interrupted_creation` returns `false` if any byte past the written prefix is nonzero, and its CRC branch requires `head[..6] == canonical[..6]`. So a corrupted-magic V0 file carrying any nonzero data (payload *or* version stamp) is a non-prefix and stays `BlobCorrupt`.

The only file that heals is the documented residual: magic byte lost **and** version 0 **and** all-zero payload — which loses no data, since every erased byte is zero. This is already pinned by the reject-table tests `rotted-magic V0 blob with payload` and `rotted-magic V0 blob with its version stamp`, both asserting `BlobCorrupt`.

runtime/src/utils/buffer/paged/cache.rs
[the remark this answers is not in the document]
roberto-bayardo: This is a false positive. A physical page whose size *divides* 4096 (e.g. 2048) never straddles a 4096-byte storage boundary: it is ≤ 4096 and 4096 is a multiple of it, so with data starting at the 4096-aligned offset, every physical page lies entirely within one storage page (P=1 → `[6144, 8192)` ⊂ `[4096, 8192)`; the 4096 boundaries land on physical-page starts, never mid-page). A non-4096-aligned *start* is not a straddle.

The warning fires exactly when neither size divides the other — the genuine cross-boundary case (e.g. 3000-byte physical pages) — so the current condition is correct.

runtime/src/lib.rs
roberto-bayardo: We can't safely share the magic — the distinct tag is what keeps a torn V1 creation from masquerading as a valid V0 blob. If both layouts used `CWIC` and we distinguished them by `runtime_version` (0 vs 1), a V1 creation torn after the magic persisted but before the version bytes did would read as `CWIC` + `runtime_version 0` — a byte-perfect valid V0 header — and open at the V0 data offset (8) instead of 4096, surfacing the zero-padded region as V0 data. `runtime_version` can't disambiguate, because V0 legitimately stamps it `0`, and `0` is exactly what a torn/zeroed write leaves behind — so a half-written V1 would be indistinguishable from a complete V0. Putting the tag in the magic (nonzero and distinct) makes a partial V1 magic incapable of becoming a complete V0 magic by zeroing; `test_header_torn_v1_does_not_parse_as_v0` pins this.

There's also no compatibility upside to sharing: an old binary reading a V1 blob rejects it either way via the `runtime_version` check, so a shared magic wouldn't buy backward-readability.

runtime/src/lib.rs
roberto-bayardo: `DEFAULT_BLOB_VERSION` is the application-owned blob version (bytes 6-7), used by `open()` when no range is given — orthogonal to the runtime layout version. It needs to stay `0`: existing blobs are `blob_version 0`, and `open()` opens with `DEFAULT..=DEFAULT`, so bumping it to `1` would make `open()` reject every current blob with `BlobVersionMismatch`. The `1` here is the runtime *layout* version (`BlobHeaderLayout::V1.runtime_version() == 1`, bytes 4-5), which this PR does set — a separate field the application never sees.

patrick-ogrady: my b

patrick-ogrady (approved):

PR #4185, opened 2026-07-07
[ci] use R2 bucket as backend for sccache

andresilva:


.github/actions/setup/action.yml
```diff
@@ -143,23 +144,32 @@ runs:
...
-        rw_mode = "$sccache_rw_mode"
-        EOF
+        SCCACHE_IDLE_TIMEOUT=0 \
+        SCCACHE_S3_RW_MODE="$sccache_rw_mode" \
+        sccache --start-server
 
         echo "CARGO_INCREMENTAL=0" >> "$GITHUB_ENV"
         echo "RUSTC_WRAPPER=sccache" >> "$GITHUB_ENV"
-        echo "SCCACHE_CONF=$sccache_config" >> "$GITHUB_ENV"
+        echo "SCCACHE_IGNORE_SERVER_IO_ERROR=1" >> "$GITHUB_ENV"
```
andresilva, line 191: In case there is some issues with the sccache server (e.g. it fails to talk with R2), this will allow falling back to plain rustc rather than erroring and failing the build.

.github/workflows/benchmark.yml
```diff
@@ -8,6 +8,12 @@ on:
 permissions:
   contents: read
 
+env:
+  SCCACHE_R2_BUCKET: ${{ secrets.SCCACHE_R2_BUCKET }}
+  SCCACHE_R2_ENDPOINT: ${{ secrets.SCCACHE_R2_ENDPOINT }}
+  SCCACHE_R2_ACCESS_KEY_ID: ${{ secrets.SCCACHE_R2_ACCESS_KEY_ID }}
+  SCCACHE_R2_SECRET_ACCESS_KEY: ${{ secrets.SCCACHE_R2_SECRET_ACCESS_KEY }}
```
andresilva: Templates can't access secrets, so we need to export these on every workflow that uses the setup action.

patrick-ogrady (dismissed):

clabby (approved):

.github/actions/setup/action.yml
```diff
@@ -143,23 +160,32 @@ runs:
...
-        [cache.gha]
-        enabled = true
-        version = ""
-        rw_mode = "$sccache_rw_mode"
-        EOF
+        SCCACHE_IDLE_TIMEOUT=0 \
+        SCCACHE_REGION=auto \
+        SCCACHE_S3_RW_MODE="$sccache_rw_mode" \
+        SCCACHE_S3_USE_SSL=true \
+        sccache --start-server
```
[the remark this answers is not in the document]
andresilva: `mozilla-actions/sccache-action` does not start the server.

PR #4186, opened 2026-07-07
[storage] [runtime] Make contiguous journal blob rollover cancellation-safe

### LINKED ISSUE #4148: [storage] Fix update-before-await pattern

danlaine:
Partially resolves #4148.

## Problem

When an append crosses a blob boundary, the contiguous journal rolls over: `seal_tail()` seals the full tail and opens a successor. Dropping the append future while `seal_tail` was suspended corrupted the journal. `seal_tail` installed the successor as the tail *before* awaiting `old_writer.seal()`, and `seal(self)` consumed the writer while awaiting its flush, so a drop mid-seal destroyed the old tail: it never reached the sealed list, `tail_blob_index()` went off by one, and `bounds.end` already counted the appended item. From that state, reads of the item failed, `sync()` persisted a recovery watermark past the recoverable prefix (reopen then failed with `Corruption`), the next append opened the successor blob a second time, and `prune()` could hit an assertion. The corruption is in-memory struct state, so it affects every storage backend.

## Fix

The root cause is a consuming call (`seal(self)`) with an await inside it: a drop after the consume but before completion loses the writer. The fix replaces it with an in-place swap:

- `Writer::seal(self)` is deleted. `Writer::roll(&mut self, successor)` flushes the writer, then synchronously swaps in the successor and returns a `Sealed` view of the old contents (keeping the writer's page-cache id, so hot pages stay valid). Dropping the future mid-flush leaves the writer installed and retryable; only the successor is lost. Roll also waits for any outstanding `start_sync` first, since its completion is the only witness of an in-flight fsync failure.
- Every path to a `Sealed` (`snapshot` or `roll`) flushes first, so a `Sealed` with unwritten bytes is unrepresentable.
- `seal_tail` opens the successor first (a failure leaves the tail untouched), then rolls it in. Journal recovery seals below-tail writers via `snapshot` (the flush is a no-op there).

Durability is unchanged from main: rollover writes bytes without fsyncing, and sealed blobs become crash-durable at the next `commit()`/`sync()` via existing dirty tracking.

### DIFF
--- a/storage/src/journal/contiguous/blobs.rs
+++ b/storage/src/journal/contiguous/blobs.rs
@@ -372,6 +372,15 @@ impl<E: Context> Writable<E> {
         for blob in self.oldest_blob_index..=self.tail_blob_index() {
             self.partition.remove(blob).await?;
         }
+
+        // An interrupted rollover can leave an untracked successor blob on disk; it must not
+        // survive the clear.
+        if let Some(stray) = self.tail_blob_index().checked_add(1) {
+            match self.partition.remove(stray).await {
+                Ok(()) | Err(Error::Runtime(RError::BlobMissing(..))) => {}
+                Err(err) => return Err(err),
+            }
+        }
         let _ = self.metrics.tracked.try_set(0);
         self.tail = self.partition.open(tail_blob).await?;
         self.metrics.tracked.inc();
--- a/storage/src/journal/contiguous/blobs.rs
+++ b/storage/src/journal/contiguous/blobs.rs
@@ -186,8 +186,6 @@ impl<E: Context> Writable<E> {
             if blob == tail_blob {
                 tail = Some(writer);
             } else {
-                // Seal the recovered writer; the flush is a no-op because recovery already
-                // wrote back any repairs.
                 sealed.push(writer.snapshot().await.map_err(Error::Runtime)?);
             }
         }
@@ -259,10 +257,6 @@ impl<E: Context> Writable<E> {
     }
 
     /// Seal the tail (no fsync) and open the next blob as the new tail.
-    ///
-    /// Cancellation-safe: a dropped future leaves the current tail installed and unsealed,
-    /// and a later append retries the rollover; the swap and the sealed-list update are
-    /// synchronous.
     pub(super) async fn seal_tail(&mut self) -> Result<(), Error> {
         // Open the next tail first so a failure leaves the current tail untouched.
         let next_blob = self
@@ -270,8 +264,6 @@ impl<E: Context> Writable<E> {
             .checked_add(1)
             .ok_or(Error::OffsetOverflow)?;
         let new_writer = self.partition.open(next_blob).await?;
-
-        // Dirty tracking covers the sealed blob until commit/sync.
         let sealed = self.tail.roll(new_writer).await.map_err(Error::Runtime)?;
         self.metrics.tracked.inc();
         self.sealed.push(sealed);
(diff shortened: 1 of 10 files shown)


storage/src/journal/contiguous/blobs.rs
```diff
@@ -264,8 +264,7 @@ impl<E: Context> Writable<E> {
             .checked_add(1)
             .ok_or(Error::OffsetOverflow)?;
         let new_writer = self.partition.open(next_blob).await?;
-        let old_writer = std::mem::replace(&mut self.tail, new_writer);
-        let sealed = old_writer.seal().await.map_err(Error::Runtime)?;
+        let sealed = self.tail.roll(new_writer).await.map_err(Error::Runtime)?;
```
danlaine, line 267: This is the important bit. We now wait until all `.await` completes to replace the tail.

```rs
    pub async fn roll(&mut self, successor: Self) -> Result<super::Sealed<B>, Error> {
        self.sync_state.wait_for_pending().await?;
        self.flush_internal(true, false).await?;
        let old = std::mem::replace(self, successor);
        Ok(old.sealed_handle(old.id))
    }
```

danlaine:

PR #4187, opened 2026-07-07
[cryptography] Adapt SHA SIMD to new Hasher trait

cronokirby:
This integrates the changes in https://github.com/commonwarexyz/monorepo/pull/4154 to the hasher refactor. The new methods was indeed necessary.

I checked that the function does inline correctly.


clabby (commented):
I wonder if it'd be worth keeping SIMD paths for other slice sizes, rather than just 72? We could also make use of this in the BMT.

Not super attached to this being general-use, though that's why I originally kept it open to arbitrary input sizes.

cronokirby:
Added a case to accelerate the BMT as well

storage/src/bmt/mod.rs
clabby, line 1: We could also improve `Proof::verify_element_inclusion` and `Proof::verify_multi_inclusion` to directly use `H::hash` rather than re-using a hasher throughout. That should improve the `prove_multi` benchmark by quite a bit, since it unblocks it from using the optimized path.

cryptography/src/sha256/simd/mod.rs
```diff
@@ -0,0 +1,133 @@
...
+        MMR_NODE_LEN => {
+            let left = assemble::<MMR_NODE_LEN>(left)?;
+            let right = assemble::<MMR_NODE_LEN>(right)?;
+            dispatch_mmr(&left, &right)
+        }
+        BMT_NODE_LEN => {
+            let left = assemble::<BMT_NODE_LEN>(left)?;
+            let right = assemble::<BMT_NODE_LEN>(right)?;
+            dispatch_bmt(&left, &right)
+        }
```
clabby: I think we can avoid temporary-buffer assembly here by directly loading the parts from `left` and `right` into registers? For the 64b case this works out rather nicely, as we can place the parts directly into `v4..v11` without a temporary allocation on aarch64 for ex. Same premise applies for the 72b case, though we'd need to pull in the tail into the low 8 bytes of the padding.

clabby (dismissed):

patrick-ogrady (approved):

PR #4189, opened 2026-07-07
[runtime] Recreate blobs left torn by an interrupted creation

roberto-bayardo:
> **Status: parked.** Likely superseded by #4194, which defers blob durability to the first sync: under that contract the creation-time durability choreography this PR hardens is deleted entirely, and the torn-header recovery core it introduces is carried forward there with a simpler justification (geometry instead of sync ordering). This PR remains the fsync-on-create alternative if #4194 does not ship.

## Summary

Blob creation writes the header into the new file in place (`set_len` -> `write` -> `sync`), so a crash mid-creation can leave a committed-length file whose header bytes are garbage — writes may tear at any point. Opening such a file returned `BlobCorrupt`, and since contiguous-journal init opens every scanned blob, one torn tail file (created eagerly by `seal_tail`) blocked journal reopen until an operator deleted it by hand. Every other layer of the stack already recovers from torn writes unattended (dual page CRC slots, journal rewind); the blob header was the only torn write that escalated to a human.

## Design

Creation makes the header durable before `open` ever returns the blob, and data can only be written through the returned blob. An invalid header on a file **no larger than the header region** is therefore provably a creation that never completed — no data was ever committed through the API — and `open` now recreates it in place. This holds under arbitrary tearing and adds no I/O to the happy path.

What still fails loudly:
- Files **larger** than the header region with an invalid header: data may be at stake, so recreating (which truncates) would be destructive. Still `BlobCorrupt`.
- A valid header whose blob version is outside the caller's range: a fully valid header cannot be a torn creation, so `BlobVersionMismatch` is preserved at any size.

The parse-and-classify decision lives in one shared helper (`Header::resolve`): backends read the blob's leading header bytes and receive either the validated header or the decision to recreate, so classification cannot drift between backends; only the I/O primitives (read, set_len, write, sync) remain per backend.

The rule extends today's `raw_len < 8` recreate behavior to the torn states a crash can actually produce (e.g. a zero-filled file at committed length, or valid magic with garbled version bytes). Existing magic-mismatch tests used an 8-byte zero file — exactly the state that now heals — and were re-specified against a header-region-plus-one file.

## Tests

- `interrupted_creation` classifier unit test (torn states heal at header size, version conflicts and larger files never do).
- Torn-header recreate tests on the memory backend (the deterministic runtime's substrate) and the tokio backend, whose recreate path needed a cursor rewind that the reopen assertion pins. The classification is shared code (`Header::resolve`), so these cover it once; iouring's (updated) magic-mismatch test still covers the loud path, and the full iouring suite plus clippy verified on Linux.

## Notes

- The heal is crash-idempotent: creation and heal truncate the file to zero before writing the header, so a torn header write cannot splice leftover bytes into a valid header with a wrong version - every partial state stays shorter than the header and heals on the next open.
- Directory syncs precede the header write, so a valid header on disk implies its directory entries are durable; this is what entitles the Valid path to skip directory syncs on reopen (including a retry of an attempt that failed mid-creation).

- Reported by qed-audit against #4184; the window predates the V1 header layout (V0 creation has the identical sequence), which is why this lands off `main`. When #4184 rebases onto it, the same rule extends to the 4096-byte V1 header region and its extension-CRC/truncation failures.
- The blob layer cannot distinguish a torn creation from byzantine rot of a previously synced header on a file that never grew past the header region; recreating is chosen because such a file contains no committed data either way. Data-bearing blobs are unaffected by construction.

### DIFF
--- a/runtime/src/storage/iouring.rs
+++ b/runtime/src/storage/iouring.rs
@@ -162,9 +162,25 @@ impl crate::Storage for Storage {
             // Existing blob - honor the header it records
             HeaderResolution::Valid { blob_version, size } => (blob_version, size),
             HeaderResolution::Recreate => {
-                // New, torn, or corrupted blob - truncate and write header with latest version
+                // Make the blob name durable BEFORE writing the header: a valid header on
+                // disk then implies its directory entries are durable, which is what lets
+                // the Valid arm skip directory syncs. A healed torn creation (raw_len > 0)
+                // may be retrying an attempt that failed before its directory syncs, so its
+                // directory entries cannot be assumed durable even though they exist.
+                // `parent_existed` reflects the live filesystem, so a partition directory
+                // left by a failed prior attempt in this process skips the storage-root
+                // sync; that narrow window is accepted (process restarts are covered by the
+                // startup filesystem sync).
+                sync_dir(parent)?;
+                if !parent_existed || raw_len > 0 {
+                    sync_dir(&self.storage_directory)?;
+                }
+
+                // Truncate to zero before writing so a torn header write cannot splice old
+                // bytes into a fully valid header with a wrong version: every partial state
+                // stays shorter than the header and heals on the next open.
                 let (header, blob_version) = Header::new(&versions);
-                file.set_len(Header::SIZE_U64)
+                file.set_len(0)
                     .map_err(|e| Error::BlobResizeFailed(partition.into(), hex(name), e.into()))?;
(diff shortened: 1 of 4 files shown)


roberto-bayardo:
Closing for now: parking torn-header recovery while #4180 (submission-ordered blob ops) and the delayed-fsync work land, since they change the shape of the fix. #4184 carries the minimal hardening (truncate-first on recreate); #4190 tracks removing the creation fsync.

PR #4191, opened 2026-07-07
[storage] Make Ordinal init cancel-safe

danlaine:
## Context

`Ordinal` recovers its state from disk by replaying records into an in-memory `RMap` (`intervals`), driven by a caller-supplied `bits` map that identifies which record indices were durably committed. Everything the public API reports (`has`, `get`, `next_gap`, `missing_items`) is answered from `intervals` alone; none of it reads raw disk bytes at query time.

Despite that, `Ordinal::init` also physically zeroed any on-disk record whose bit was *not* set, for every section it kept. This wasn't required for correctness of the read path above — it was defense-in-depth, intended to keep the stored blob's contents matching the checkpointed view.

## The problem

The zeroing was a `blob.write_at()` + `blob.sync()` against a blob that `init` had just opened fresh via `context.open()`. Per the runtime's blob contract, two independently-opened handles to the same blob name are never ordered against each other. If the `init()` future is dropped mid-repair and `init()` is called again, the retry opens a brand-new handle to the same section. The abandoned zero-write from the first attempt is now an orphan: it can still land on disk at an arbitrary later time, through the old handle, unordered against anything the retry does.

Concretely: the retry can `put()` a new record at the same offset the orphan targets and `sync()` it successfully, and the orphan can then land afterward and zero out the bytes the retry just wrote. The in-memory `intervals` map (built during the retry) still reports the index as present, but a subsequent `get()` fails with `InvalidRecord` because the CRC no longer matches. If that corrupted state is later persisted through another durability barrier, a following restart can fail recovery entirely with `MissingRecord`.

## The fix

Remove the zeroing loop from `Ordinal::init`. Records outside the committed bits are simply left on disk as-is:

- They remain unreachable through the public API, since `has`/`get`/etc. are driven solely by `intervals`, which `bits` alone determines.
- If one of them is legitimately committed in a later `init` call, its original bytes are still valid and get replayed correctly.
- If it's instead overwritten via `put()`, that write fully replaces the record regardless of what was there before.

With no destructive write left in the repair path, there is nothing for a canceled `init` to orphan, and the corruption scenario above can no longer occur. The only thing `init` removes is whole sections omitted from `bits`: retention keys purely on presence in the map, so an explicitly listed section is retained even when none of its bits are set, leaving its uncommitted records on disk for a later restart to commit. Record replay/validation is unchanged.

Updated the module docs to describe the new (simpler) recovery contract, and updated the tests that exercised the old destructive behavior to assert the new one: an uncommitted record survives untouched — whether or not its section has any committed records — and is recoverable if a later `init` commits it.

### DIFF
--- a/storage/src/ordinal/mod.rs
+++ b/storage/src/ordinal/mod.rs
@@ -55,10 +55,11 @@
 //!
 //! To recover existing data, pass `Some(bits)` to [Ordinal::init]. The bits identify which records
 //! were durably committed by the caller. [Ordinal] validates required records using their CRC32 and
-//! rebuilds the in-memory [crate::rmap::RMap]. Stored sections omitted from `bits` are removed, and
-//! stored records whose bits are unset are cleared before replay. Missing or invalid required records
-//! fail initialization. Passing `Some(BTreeMap::new())` or `None` removes all stored sections and
-//! starts empty.
+//! rebuilds the in-memory [crate::rmap::RMap]. Stored sections omitted from `bits` are removed.
+//! Stored records whose bits are unset are ignored and left on disk untouched; only `bits` governs
+//! what [Ordinal::has] and [Ordinal::get] observe. Missing or invalid required records fail
+//! initialization. Passing `Some(BTreeMap::new())` or `None` removes all stored sections and starts
+//! empty.
 //!
 //! # Example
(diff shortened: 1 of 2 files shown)


danlaine:

Issue #4190, opened 2026-07-07
[runtime] remove fsync on blob creation


roberto-bayardo:
When a new blob is created, its header is written and fsynced to disk. We should see if we can remove that fsync and rely on the one performed by the first durability request from the user.

## Current state

Creation is actually more expensive than one fsync on the tokio backend — it performs **two file fsyncs plus up to two directory fsyncs**:

1. `sync_all` on the just-created empty file, plus `fsync` of the parent directory (and the storage root if the partition is new) — added in #1990 to guarantee the blob durably *exists* (inode + dirent) once `open()` returns.
2. `sync_all` again after the header write — added in #2694 to guarantee the header *content* is durable before the caller can write data.

The iouring backend already uses the merged form (one file fsync after the header write, then the directory fsyncs), so the two-fsync sequence on tokio is historical layering, not a requirement.

## Step 1: merge the two fsyncs (tokio)

Write the header first, then sync file → parent dir (→ root). Both `open()`-return guarantees are preserved (header durable, dirent durable, in content-before-dirent order); this halves the file fsyncs per creation with no contract change. The intermediate crash states this reorders all fall in the recreate-on-open class established by #4189 and its extension in #4184 (torn-header recovery), so it should land on top of those.

## Step 2: full deferral (this issue's original ask)

Removing the remaining creation fsync and riding the first user `sync()` needs a design decision. The torn-header recovery proves "a file no larger than the header region holds no committed data" *because* creation syncs the header before data is writable. Defer that sync and a power loss before the first user sync can leave a **large** file whose data pages flushed but whose header page didn't — an invalid header above the recovery bound, which fails loudly and needs an operator again (nothing acknowledged is lost, but the blob layer can't prove it). Options:
- caller-side knowledge: the contiguous journal can prove its tail held no committed items (recovery watermark) and delete/recreate a corrupt *tail* blob specifically — fully fsync-free;
- accept the narrow power-loss-before-first-sync window (plain process crashes are already covered by the startup `syncfs`).

Also worth noting: the fsync sits on the journal rollover path (`seal_tail` pre-creates the next blob), i.e., once per `items_per_blob` appends.

PR #1990, opened 2025-10-24
[Runtime] Change `Storage` to ensure blobs are durably open/removed

danlaine:
In `main` we don't guarantee that opened `Blob`s are persisted. You could open a blob, crash, restart and the find that the blob is gone. The same goes for removed blobs.

This PR updates the `Storage` API to guarantee that blobs are durably persisted on open/remove.

PR #4193, opened 2026-07-08
[runtime] Sync once when creating a blob

roberto-bayardo:
> **Status: parked.** Likely subsumed by #4194, which removes the creation fsync entirely (this PR's two-to-one fsync merge is the intermediate step on the fsync-on-create path). Kept for the case where #4194 does not ship.

## Summary

First step of #4190: tokio's blob creation performed **two** file fsyncs — one on the just-created empty file (from #1990, guaranteeing the blob durably *exists* at `open()` return) and a second after the header write (from #2694, guaranteeing the header *content* is durable before the caller can write data). This merges them: write the header region first, then one `sync_all`, then the directory-entry fsyncs.

Both `open()`-return guarantees are unchanged (header durable, directory entry durable, in the canonical content-before-dirent order), and the sequence now matches what the iouring backend has always done — the two-fsync form was historical layering, not a requirement. Any intermediate crash state this reorders is a header-region-sized torn creation, which `open()` recreates in place per the torn-header recovery (#4189).

The remaining creation fsync — deferring durability entirely to the first user `sync()` — is the open design question tracked in #4190 (it interacts with the recovery bound proof; see the issue).

## Notes

- Stacked on #4189 (base branch set accordingly, so the diff is this commit alone); sibling of #4184, which will adopt the same ordering in its V1 creation arm when it rebases.
- tokio-only: iouring already syncs once at creation; the memory backend has no fsync concept.

### DIFF
--- a/runtime/src/storage/tokio/mod.rs
+++ b/runtime/src/storage/tokio/mod.rs
@@ -120,32 +120,6 @@ impl crate::Storage for Storage {
 
         // Assume empty files are newly created. Existing empty files will be synced too; that's OK.
         let len = file.metadata().await.map_err(|_| Error::ReadFailed)?.len();
-        let newly_created = len == 0;
-
-        // Only sync if we created a new file
-        if newly_created {
-            // Sync the file to ensure it is durable
-            file.sync_all()
-                .await
-                .map_err(|e| Error::BlobSyncFailed(partition.into(), hex(name), e.into()))?;
-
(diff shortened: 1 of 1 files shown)


roberto-bayardo:
Closing for now along with #4189: parking torn-header/creation-fsync work while #4180 and the delayed-fsync direction land. #4190 remains the tracker.

Issue #2026, opened 2025-10-27
[runtime] Handle Non-Durable Windows Directories


patrick-ogrady:
Related: #1990 

Windows does not support syncing a directory like `UNIX`. This means that on crash a `Blob` can be created/synced and still be lost (after returning OK).

This continues to beg the question of whether we should continue building for Windows (the time precision is already an "unnecessary" complexity).

PR #4194, opened 2026-07-08
[runtime,storage] Defer blob durability to the first sync

roberto-bayardo:
## Summary

Prototype for step 2 of #4190: blob **creation performs no fsyncs**. Durability - of the file, its directory entry, and the partition directory's own entry in the storage root - is established by the blob's **first `sync()`**. This deletes the entire creation-time durability choreography (the empty-file sync, the directory fsyncs at open, and their ordering concerns) and moves the cost off the journal rollover path: blobs that never sync never pay.

## Design

- **`DirSync`**: each file-backed `Blob` carries its partition-directory and storage-root paths plus a once flag; the first sync fsyncs file -> partition dir -> storage root (later syncs skip the directories). Contracts updated: `open_versioned` states creation is not durable until the first sync; `Blob::sync` owns dirent durability.
- **Torn-header recovery** (required foundation): with no creation fsync, torn headers at final names are normal crash states. `Header::resolve` recreates any file no larger than the header region whose bytes do not form a header - under this contract that file *provably* holds no synced data, by geometry alone: data lives past the region and a sync makes the whole file durable, header included. Creation truncates to zero before writing, so no torn write can splice leftover bytes into a valid wrong-version header. `VersionMismatch` (valid header, real conflict) and CRC-valid-but-bogus offsets stay loud at any size, as does any invalid header on a file large enough to carry data.
- **Watermark adjudication**: the one state the blob layer cannot classify - an invalid header on a file *larger* than the region, produced by writeback flushing unsynced data pages but not the header - is decided by the caller. The fixed contiguous journal's `open_all` defers a corrupt *newest* blob to `init`, which removes it when the checkpoint watermark proves no adopted item lives inside it and fails loudly when one does. A regression test drives both directions through a real filesystem.

## Status: prototype

Known gaps before this could leave draft:
- The variable contiguous journal keeps strict (loud) behavior; the same watermark pattern applies.
- Other blob consumers (metadata, archive, freezer) not yet audited under the weakened creation contract (metadata syncs immediately after writing, so it should be unaffected; the freezer interaction is the same class as the separately-assigned rewind-blindness issue).
- iouring compiles but has not been exercised on Linux yet; the non-unix fallback performs no directory fsyncs (pre-existing, #2026).
- The measurement that motivates this (creation-fsync cost on the rollover path) has not been run; per #4190 it should gate whether this ships.

## Notes

- Stacked on #4184 (page-aligned layout); the parked #4189/#4193 recovery/fsync work is subsumed by this design if it ships.

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -697,7 +697,9 @@ stability_scope!(BETA {
         /// Multiple instances of the same blob can be opened concurrently, however,
         /// writing to the same blob concurrently may lead to undefined behavior.
         ///
-        /// An Ok result indicates the blob is durably created (or already exists).
+        /// Creation is NOT durable until the blob's first [Blob::sync] returns: a crash
+        /// before then may lose the blob (or leave a torn header), and reopening recreates
+        /// it empty. Data written to the blob is likewise only durable after a sync.
         ///
         /// # Versions
         ///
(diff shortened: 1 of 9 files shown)


roberto-bayardo:
Closing for now: #4180 (submission-ordered blob ops) and the delayed-fsync mechanism change the shape of this. The review findings (write_at_sync durability hole, adjudication gaps, into_std error swallow) are recorded in this PR's history for whoever picks it back up. #4190 remains the tracker.

PR #4196, opened 2026-07-08
[storage] Allow blobs to be missing during pruning

popzxc:
I've hit a flaky bug during in my integration tests setup, where the node with commonware-simplex consensus goes through several in-memory restarts; it appears that in case blob is already missing (e.g. due to external journal change) the simplex voter would panic because of unconditional `expect`.

Ignoring already-missing blobs seems harmless, and I confirmed that it solves my issue, so I hope this fix is acceptable.

AI usage disclosure: I used Claude to find the bug and prepare the fix, but I reviewed and verified it manually & checked the contributing guidelines before submitting the PR. The message is mine as well.

### DIFF
--- a/storage/src/journal/segmented/manager.rs
+++ b/storage/src/journal/segmented/manager.rs
@@ -345,10 +345,21 @@ impl<E: Storage + Metrics, F: BufferFactory<E::Blob>> Manager<E, F> {
             let size = blob.size();
             drop(blob);
 
-            // Remove blob from storage
-            self.context
+            // Remove blob from storage. A missing blob is treated as already
+            // pruned rather than as an error: a straggling file operation from
+            // a previous incarnation of the same journal (e.g. across an
+            // in-process restart) can remove a section that this incarnation
+            // listed at init, and only sections below the prune floor are ever
+            // affected.
+            match self
+                .context
                 .remove(&self.partition, Some(&section.to_be_bytes()))
-                .await?;
+                .await
+            {
+                Ok(()) => {}
+                Err(RError::BlobMissing(_, _)) => {}
+                Err(err) => return Err(Error::Runtime(err)),
+            }
             pruned = true;
 
             debug!(section, size, "pruned blob");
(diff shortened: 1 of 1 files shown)


popzxc:
@patrick-ogrady @roberto-bayardo @clabby could someone please take a look at this (not sure who's the right person to ping)?

patrick-ogrady:
Sorry for the delay on this, @popzxc! (must've slipped past my github notifications)

We make a number of pretty strong assumptions about storage to simplify the code that lives on top (otherwise it can get pretty hairy pretty fast). The big two that come to mind are [1] any external storage manipulation can lead to undefined behavior and [2] any storage error is considered FATAL.

Happy to chat on a call or something to better understand what your goals are here if it is a blocker for some reason!

patrick-ogrady:
Will take a look in any case in case this falls under normal system failure if a prune only partially completes (in which case that would be an issue).

patrick-ogrady:
> Will take a look in any case in case this falls under normal system failure if a prune only partially completes (in which case that would be an issue).

Ah, the approach we take here should mean this can never happen to a correct node (because we remove + sync from lowest to highest -> so you should never see a gap). If you are aware of such a case, please let us know!

patrick-ogrady:
JFYI If you are running any cross-region integration tests, I suspect they'll get dramatically faster once we merge our "stable leader" impl in the ~v2026.8.0 release (optional ofc):

* https://github.com/commonwarexyz/monorepo/pull/3352
* https://github.com/commonwarexyz/monorepo/pull/3416

popzxc:
Oh, that was a quick response! ❤️ 

So, on the bug: it is only possible with _in-process restart_, e.g. it's not possible when you actually shutdown the process.
This problem can occur because awaiting `stop` doesn't guarantee that tasks are actually killed (it's [part of the contract](https://github.com/popzxc/commonware-monorepo/blob/tolerate-externally-removed-blob/runtime/src/lib.rs#L323)), and removing files is done via [tokio::fs::remove_file](https://github.com/popzxc/commonware-monorepo/blob/tolerate-externally-removed-blob/runtime/src/storage/tokio/mod.rs#L200), which uses `tokio::spawn_blocking` under the hood AFAIU. And blocking operations are not cancelled on drop -- they continue to run. Again, as far as I understand, there are other operations affected by this as well, e.g. use of `spawn_blocking` in [this module](https://github.com/popzxc/commonware-monorepo/blob/tolerate-externally-removed-blob/runtime/src/storage/tokio/unix.rs#L153)

So if you do an in-process restart, even if you waited for `stop`, the new Engine could start while some operation is not finished, it would also try to prune storage (in my case), and it would fail if previous scheduled removal finishes at this point already. 

I believe it could be worked around by killing the tokio runtime altogether -- this would guarantee that all blocking tasks have exited, but it feels like a pretty heavy and somewhat implicit invariant to handle. 

Thus I thought that making blob removal idempotent would work as a fix. It's simple, it's seemingly harmless, and it solves the problem without having to rework the architecture around this fact.

If needed, [here's the test](https://github.com/matter-labs/zksync-os-server/blob/popzxc-simplex-consensus/integration-tests/tests/node/consensus.rs#L237) on which this issue occurred. It's not reproducible right now because I use patched version of commonware (with this fix applied), but if you change the commonware version back to upstream, you can reproduce it. What made it much more common for me was saturating the filesystem: I was running a consensus network under load + had some operations that opened many file descriptors in parallel, which made FS operations much slower for the integration tests.

---

So tl;dr -- this bug fix is not _mandatory_, it only affects in-process restarts and there is a workaround for that, but the workaround is pretty taxing and somewhat unexpected.

patrick-ogrady:
I see the edge case now. Thanks for taking the time to spell this out.

Will poke at it for a bit and see if I can get something reasonable in our imminent release (supposed to go out rn). Otherwise, will definitely (somehow) be addressed in the next one.

patrick-ogrady:
> I see the edge case now. Thanks for taking the time to spell this out.
> 
> Will poke at it for a bit and see if I can get something reasonable in our imminent release (supposed to go out rn). Otherwise, will definitely (somehow) be addressed in the next one.

Actually, this would be pretty brutal to fuck up last minute lol. Will be next release 😓 .

popzxc:
No issue! We're not in production yet, so we'll be fine with using a pinned commit for a while :)

roberto-bayardo:
Perhaps @danlaine 's cancel safety work could address this.

PR #4204, opened 2026-07-08
[parallel] Add experimental heartbeat-scheduled thread pool

cronokirby:
Every collection operation faces the same decision: is this call large enough that parallelism beats a serial loop? The Rayon strategy answers it empirically: the adaptive policy times both paths per callsite and size bucket and replays the winner. That works, but it costs a lookup on every call, cold-start mispredictions, stale estimates, probe noise, and a duplicated serial/parallel body in every strategy method.

This prototype answers the question structurally instead, following heartbeat scheduling (Acar, Chargueraud, Muller, and Rainey, PLDI 2018; see also Spice and Chili). Calls always run on the calling thread and join() marks forked work stealable for a few nanoseconds. A self-timed heartbeat promotes the oldest latent fork at most once per interval, and only when idle threads outnumber queued jobs. Workers executing stolen forks expand eagerly (a steal proves the computation outlived an interval), blocked joiners execute queued work instead of parking, and idle workers spin briefly before parking so promotions during busy periods do not pay a futex wake.

The result is a contract rather than a prediction. Work below one interval runs at serial speed by construction (14ns vs 15us for always-parallel rayon on a 16-element fold). Work a few intervals long pays a bounded constant (worst measured about 1.3x serial). Genuinely large work scales to the machine (5.8x on 14ms of work, within about 25% of eager rayon and converging as work grows). There is no per-call state, no measurement, and no size hints: one code path serves every call, so callers never need to hesitate before routing work through the pool.

Benchmarks live in src/heartbeat/benches (fold over cheap and expensive workloads across sizes, against Sequential, the adaptive Rayon strategy, and raw rayon), plus an ignored midrange_exploration test that prints scheduler counters. The unsafe stack-job protocol passes miri, including a zero-interval test that forces constant stealing.

Not yet wired into Strategy; map/sort/try variants and Linux timing validation are follow-ups.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1662,8 +1662,10 @@ version = "2026.5.0"
 dependencies = [
  "cfg-if",
  "commonware-macros",
+ "criterion",
  "dashmap",
  "futures",
+ "parking_lot",
  "proptest",
  "rayon",
 ]
--- a/parallel/Cargo.toml
+++ b/parallel/Cargo.toml
@@ -17,14 +17,21 @@ cfg-if.workspace = true
 commonware-macros.workspace = true
 dashmap = { workspace = true, optional = true }
 futures = { workspace = true, optional = true }
+parking_lot = { workspace = true, optional = true }
 rayon = { workspace = true, optional = true }
 
 [dev-dependencies]
+criterion.workspace = true
 dashmap.workspace = true
 futures.workspace = true
 proptest.workspace = true
 rayon.workspace = true
 
+[[bench]]
+name = "heartbeat"
+harness = false
+path = "src/heartbeat/benches/bench.rs"
+
 [features]
 default = [ "std" ]
-std = [ "commonware-macros/std", "dep:dashmap", "dep:futures", "dep:rayon" ]
+std = [ "commonware-macros/std", "dep:dashmap", "dep:futures", "dep:parking_lot", "dep:rayon" ]
--- a/parallel/src/heartbeat/benches/bench.rs
+++ b/parallel/src/heartbeat/benches/bench.rs
@@ -1,0 +1,5 @@
+use criterion::criterion_main;
+
+mod fold;
+
+criterion_main!(fold::benches);
--- a/parallel/src/heartbeat/benches/fold.rs
+++ b/parallel/src/heartbeat/benches/fold.rs
@@ -1,0 +1,125 @@
+use commonware_parallel::{heartbeat, Rayon, Strategy};
+use criterion::{criterion_group, Criterion};
+use rayon::iter::{IntoParallelRefIterator, ParallelIterator};
+use std::{hint::black_box, num::NonZeroUsize, time::Duration};
+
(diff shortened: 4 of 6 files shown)


roberto-bayardo:
We gave this a decisive fit-test against the workloads driving the current pool work (context: rayon's idle-wake churn costs low-duty-cycle workloads ~7-21% wall; a wake-limited pool is being designed to fix it, and heartbeat was the natural alternative to evaluate first).

**Setup**: a minimal safe `Strategy` shim over `Pool`/`Scope::join` (map/fold via recursive `split_at_mut` halving into serial leaves, `Option<T>` slots for by-value semantics; phase-based sort; `spawn` on a dedicated thread — ~1-3 spawns/commit so this is noise). Then the same paired A/B gates used for the pool evaluation, on an EPYC 9354P (32c), 8-thread arms.

**Macro results** (eth-replay follower 20k blocks; Constantinople 32k reads + 32k updates / 1M keys):

| arm | eth wall | constantinople p50 |
|---|---|---|
| Rayon @ 8 | 51.9s | 26.2ms |
| Heartbeat @ 8 | 57.9s (+11.7%) | 46.7ms (**+78%**) |
| Rayon @ 1 (no pool) | 42.8s | -- |

**Micro that explains it** (map_collect over n items, ~6ns/item, arm64):

| total work | serial | rayon | heartbeat |
|---|---|---|---|
| ~1.6us | 1.6us | 1.8us | 4.6us (2.8x serial) |
| ~25us | 24us | 25us | 38us (1.6x serial) |
| ~380us | 380us | **124us (3.1x speedup)** | 378us (**1.0x -- zero engagement**) |

The 380us row is the key one: seven-plus intervals of available forks, and promotion-then-steal-then-expand never spun up inside the phase's lifetime, while eager dispatch got 3.1x. Our hot workloads are almost entirely bulk-synchronous phases in the 10us-1ms range (level-wise merkleize is ~15 such phases per commit), so this dead zone covers them.

Shim caveats, honestly: the iterator materialization and always-parallel `run` arms inflate the small-n overhead rows somewhat. They don't explain the 380us zero-speedup -- both heartbeat cells carry the same materialization, and rayon parallelized the identical op 3.1x.

**The structural observation**: heartbeat solves granularity-*unknown* parallelism. Every `Strategy` call site in this workspace knows `len` up front -- the adaptive policy already exploits that, and the planned wake-limited pool dispatches eagerly on it (`min(⌈len/chunk⌉, idle)` wakes, pre-assigned ranges). Heartbeat pays promotion latency to buy information we already have. If it's to compete on these shapes, it likely needs an immediate-fan-out path when `len` is provided -- at which point it converges toward eager wake-limited dispatch anyway.

None of which diminishes the scheduler itself -- the fork bookkeeping is cheap, the miri-clean stack-job protocol is solid, and for genuinely irregular recursive work the contract ("serial by construction below one interval") is exactly right. It's a fit mismatch with this workspace's bulk, size-known call sites, not a quality problem. Happy to share the shim + bench setup if useful.

PR #4183, opened 2026-07-07
[workspace] Bump `rand` & ecosystem to 0.10

clabby:
## Overview

> [!NOTE]
> Need to wait the cooldown? `curve25519-dalek` `5.0.0` just got released.

The `rand` ecosystem has been undergoing a large upgrade over the last few months, and all of our dependencies have now landed on the new version. This PR upgrades `rand` and its ecosystem to `^0.10.0`.

Related: #1841

PR #4208, opened 2026-07-08
[cryptography] Explicit SysRng

patrick-ogrady:
Related: #4183

### DIFF
--- a/cryptography/src/secp256r1/benches/signature_generation.rs
+++ b/cryptography/src/secp256r1/benches/signature_generation.rs
@@ -1,12 +1,13 @@
 use commonware_cryptography::{secp256r1, PrivateKey};
 use criterion::{criterion_group, BatchSize, Criterion};
-use rand::{rng, RngExt as _};
+use rand::{rngs::StdRng, RngExt as _, SeedableRng};
 use std::hint::black_box;
 
 fn bench_signature_generation<S: PrivateKey>(variant: impl AsRef<str>, c: &mut Criterion) {
+    let mut rng = StdRng::seed_from_u64(0);
     let namespace = b"namespace";
     let mut msg = [0u8; 32];
-    rng().fill(&mut msg);
+    rng.fill(&mut msg);
     c.bench_function(
         &format!(
             "{}/variant={} ns_len={} msg_len={}",
@@ -17,7 +18,7 @@ fn bench_signature_generation<S: PrivateKey>(variant: impl AsRef<str>, c: &mut C
         ),
         |b| {
             b.iter_batched(
-                || S::random(rng()),
+                || S::random(&mut rng),
                 |signer| {
                     black_box(signer.sign(namespace, &msg));
                 },
--- a/cryptography/src/secp256r1/benches/signature_generation.rs
+++ b/cryptography/src/secp256r1/benches/signature_generation.rs
@@ -1,10 +1,11 @@
 use commonware_cryptography::{secp256r1, PrivateKey};
+use commonware_utils::test_rng;
 use criterion::{criterion_group, BatchSize, Criterion};
-use rand::{rngs::StdRng, RngExt as _, SeedableRng};
+use rand::RngExt as _;
 use std::hint::black_box;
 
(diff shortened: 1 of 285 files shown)


examples/flood/src/bin/setup.rs
```diff
@@ -122,7 +122,7 @@ fn main() {
         bootstrappers <= peers,
         "bootstrappers must be less than peers"
     );
-    let mut rng = rand::make_rng::<StdRng>();
+    let mut rng = UnwrapErr(SysRng);
```
patrick-ogrady: This could be considered "working around the API"

cronokirby: It's nice because we avoid the fallible RNG API as well.

cronokirby:
What if we added a function to `utils` encapsulating this logic?

utils/src/rng.rs
```diff
@@ -1,11 +1,30 @@
...
+/// Returns an infallible handle to the operating system's entropy source.
+///
+/// Use this whenever randomness must come directly from the OS (e.g. key
+/// generation) rather than from a seeded or userspace RNG.
+///
+/// # Panics
+///
+/// Panics if the operating system fails to provide randomness.
+#[stability(BETA)]
+pub const fn sys_rng() -> UnwrapErr<SysRng> {
```
cronokirby: We should return an impl Trait here imo, to not have such strict coupling with the particular rng.

utils/src/rng.rs
```diff
@@ -1,11 +1,30 @@
...
+    UnwrapErr(SysRng)
+}
+
 /// Returns a seeded RNG for deterministic testing.
 ///
 /// Uses seed 0 by default to ensure reproducible test results.
+#[stability(ALPHA)]
 pub fn test_rng() -> StdRng {
```
cronokirby: ditto

utils/src/rng.rs
```diff
@@ -14,6 +33,7 @@ pub fn test_rng() -> StdRng {
 ///
 /// Use this when you need multiple independent RNG streams in the same test,
 /// or when a helper function needs its own RNG that won't collide with the caller's.
+#[stability(ALPHA)]
 pub fn test_rng_seeded(seed: u64) -> StdRng {
```
cronokirby: ditto

math/src/fields/goldilocks.rs
```diff
@@ -1,7 +1,7 @@
 use crate::algebra::{Additive, Field, FieldNTT, Multiplicative, Object, Random, Ring};
 use commonware_codec::{FixedSize, Read, Write};
 use core::ops::{Add, AddAssign, Mul, MulAssign, Neg, Sub, SubAssign};
-use rand_core::CryptoRng;
+use rand::CryptoRng;
```
cronokirby: What's the motivation behind depending on `rand`? My thinking has been that it's better to not depend on rand if we just need the trait, but perhaps this doesn't make sense with the version upgrade?

cronokirby: I think we could even have it so that every crate depends on rand_core for the trait, and then utils defines the system rng we happen to get from rand (but we could use a different source)

patrick-ogrady: Just less deps (now that it is re-exported, why use rand_core too?)

cryptography/src/secp256r1/benches/signature_generation.rs
```diff
@@ -1,12 +1,13 @@
 use commonware_cryptography::{secp256r1, PrivateKey};
 use criterion::{criterion_group, BatchSize, Criterion};
-use rand::{rng, RngExt as _};
+use rand::{rngs::StdRng, RngExt as _, SeedableRng};
 use std::hint::black_box;
 
 fn bench_signature_generation<S: PrivateKey>(variant: impl AsRef<str>, c: &mut Criterion) {
+    let mut rng = StdRng::seed_from_u64(0);
```
cronokirby: We should use `test_rng` in this place and similar places, which would let us avoid the dep on rand, and also be more consistent.

cronokirby (dismissed):

cronokirby (approved):

PR #4215, opened 2026-07-09
[ci] allow sccache setup to fail

andresilva:
Transient sccache download failures, eg. https://github.com/commonwarexyz/monorepo/actions/runs/29031776276/job/86172586287?pr=4206

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -158,6 +158,9 @@ runs:
     - name: Setup sccache
       if: inputs.rust-cache-enabled == 'true'
       uses: mozilla-actions/sccache-action@9e7fa8a12102821edf02ca5dbea1acd0f89a2696 # v0.0.10
+      # sccache is an optimization, not a requirement: if the binary download
+      # fails (e.g. transient GitHub release fetch errors), continue without it.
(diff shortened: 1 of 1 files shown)


.github/actions/setup/action.yml
```diff
@@ -158,6 +158,9 @@ runs:
     - name: Setup sccache
       if: inputs.rust-cache-enabled == 'true'
       uses: mozilla-actions/sccache-action@9e7fa8a12102821edf02ca5dbea1acd0f89a2696 # v0.0.10
+      # sccache is an optimization, not a requirement: if the binary download
+      # fails (e.g. transient GitHub release fetch errors), continue without it.
+      continue-on-error: true
```
patrick-ogrady, line 163: We could pull this from r2 too if we find it to be more reliable.

patrick-ogrady (approved):

clabby (approved):
Thanks. Longer term it'd be nice to track https://github.com/Mozilla-Actions/sccache-action/issues/218.

If/when that gets implemented, we can use `cargo-binstall` via the `install-action`, which should be a bit more reliable given it can fall back to building from source if GH's API is flaky or down.

PR #4217, opened 2026-07-09
[runtime] Defer blob creation's directory fsyncs to the first durable operation

roberto-bayardo:
Towards: #4190
Stacked on: #4184 (the recovery machinery there is what makes this deferral's crash states benign)

## Summary

Creating a blob paid two file fsyncs plus one or two directory fsyncs before `open` returned, putting up to four fsyncs on the contiguous journal's rollover path. With this change creation performs exactly **one file fsync** (the header write): the empty-file fsync existed only to order file durability ahead of the directory syncs, and the directory syncs now happen with the blob's **first durability request** instead of at creation.

## Design

Each blob handle carries a `DirSync` (parent + storage-root paths and a once-flag). `sync`, `start_sync`, and `write_at_sync` run it after the file fsync: fsync parent, fsync root, set the flag (`Release`; the fast-path gates read with `Acquire`). Both dirs are covered unconditionally, so no handle needs to know whether another open created the partition. Reopened handles also start unsynced, since they cannot know whether the entries were ever fsynced (e.g. a creation or recovery that crashed first).

The range-scoped `write_at_sync` fast paths (`pwritev2(RWF_SYNC)` on tokio-unix, the ring's sync write on iouring) cannot cover directory entries, so a handle's first `write_at_sync` routes through the full-sync path — this is what keeps `Metadata`'s checkpoint writes (whose only durability op is `write_at_sync`) fully durable.

## Safety argument

- The only new crash state is "never-synced blob vanished", which is already reachable today via a crash inside `open`, and is within contract: nothing was promised, and reopen recreates the blob. Contract docs updated accordingly (`Storage::open_versioned` owns the statement; `Blob::sync`/`write_at_sync` link to it).
- A torn header left at the blob's name heals via #4184's interrupted-creation recovery.
- Once any durable op returns, both content and dirents are durable (file fsync strictly before dir fsyncs, all before returning/resolving).
- Directory fsyncs are per-directory, so `remove()`'s existing dir fsync also persists any pending creations in the same partition.

## Adversarial review

A four-lens review of the deferral confirmed the crash-contract, backend-wiring, and doc claims, and found **one real regression, fixed here**: the contiguous journal's clear paths created a fresh tail blob and then durably finalized the clear checkpoint against it — a crash could persist the checkpoint while the never-synced tail vanished, leaving init permanently failing bounds recovery. `Writable::clear` and `complete_staged_clear` now sync the fresh tail before any checkpoint write. A follow-up verification pass confirmed the fix is complete: a systematic sweep found no other create-then-durably-reference orderings (rollover's new tail is always synced before any watermark advance by the dirty-blob tracking), and the fresh-tail sync provably reaches `Blob::sync` even on an empty paged writer.

## Notes

- Reopened handles pay the two directory fsyncs once per handle on their first durable op; they are cheap for already-durable entries.
- On iouring the (once per handle) directory fsyncs are blocking calls on the caller's task, consistent with the blocking directory operations its open path already performs inline.
- Windows semantics are unchanged: no directory-fsync primitive exists there (#2026).
- The vanished-never-synced-blob crash state cannot be modeled by the deterministic runtime's storage, so the clear-path ordering is enforced by review and the checkpoint module's documented invariant rather than a test. A faulty-storage mode that drops never-synced blobs would make this class testable and is left as a follow-up.
- Remaining #4190 step: move the last creation fsync (the header write) off the open path via a `start_sync` handle once #4180's submission ordering lands.

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -674,6 +674,11 @@ stability_scope!(BETA {
     /// recovery: data read at initialization can be assumed to survive a
     /// subsequent crash without an explicit [`Blob::sync`].
     ///
+    /// A blob whose creation was never followed by a durability request is the
+    /// exception: it may vanish entirely across a crash (see
+    /// [`Storage::open_versioned`]), so recovery must not depend on the
+    /// existence of a blob it never synced.
+    ///
     /// # Partition Names
     ///
     /// Partition names must be non-empty and contain only ASCII alphanumeric
@@ -824,7 +829,8 @@ stability_scope!(BETA {
         /// This is not a durability barrier for previous operations. When it completes,
         /// only the bytes submitted to this call are guaranteed durable. Earlier unsynced
         /// [`Blob::write_at`] or [`Blob::resize`] calls require [`Blob::sync`] to become
-        /// durable.
+        /// durable. A handle's first durability request additionally covers the blob's
+        /// directory entries, exactly as [`Blob::sync`] does.
         fn write_at_sync(
             &self,
             offset: u64,
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -705,7 +705,12 @@ stability_scope!(BETA {
         /// Multiple instances of the same blob can be opened concurrently, however,
         /// writing to the same blob concurrently may lead to undefined behavior.
         ///
-        /// An Ok result indicates the blob is durably created (or already exists).
+        /// # Durability
+        ///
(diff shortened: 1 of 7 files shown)


roberto-bayardo:
Closing for now: #4184's latest revision changed the creation path this builds on (single file fsync + directory syncs that also cover torn-blob recreation), so the deferral needs a design pass against that base rather than a mechanical rebase. The branch (blob-deferred-dir-sync) is preserved; #4190 remains the tracker for removing creation fsyncs from the latency path.

PR #4220, opened 2026-07-09
[storage] Skip redundant work in journal init and destroy, make destroy crash safe

roberto-bayardo:
## Summary

Two independent wastes on the contiguous journal's lifecycle paths, both found by the blob open-path audit:

- **Init rescanned a directory it had just scanned.** `Partition::select` already lists both candidate partitions to pick one; the fixed journal's init then listed the winner a third time before opening its blobs. `select` now returns the winning scan and init opens from it.
- **Destroy's documented contract is upgraded from "not crash-safe" to crash-safe.** Both journals' `destroy` docs previously disclaimed interruption entirely ("reopening the same partition may observe partially removed state"); they now promise that reopening after an interrupted destroy recovers either the original journal (interruption before the clear intent is durable) or a fresh, empty journal — never a partially destroyed state (see Crash safety of destroy).
- **Destroy paid a directory fsync per blob for nothing.** `Writable::destroy` removed every blob individually (each removal durably fsyncs the partition directory) immediately before removing the whole partition, which makes those fsyncs redundant. It now closes all handles first (open files cannot be removed on every platform) and removes the partition wholesale: one directory fsync instead of one per blob.

## Crash safety of destroy

Wholesale removal is unordered, so a destroy crashed during the blob-partition removal can leave gaps that recovery reads as retained loss (`recovery watermark exceeds recoverable size`) — and once boundaries are recorded rather than inferred from survivors, the old oldest-first prefix shape stops looking cleanly pruned too. Two changes make destroy crash-safe end to end, replacing the old "not crash-safe" disclaimer in its contract with the guarantee above:

- **A clear intent is staged before anything is removed.** Init honors a staged clear before scanning any blobs, so a reopen after any interruption past the durable intent wipes whatever remains unread and yields a fresh, empty journal; an interruption before the intent is durable has removed nothing and recovers the original journal. The savings hold — one metadata sync (two for the variable journal, whose offsets teardown re-stages internally) replaces the per-blob directory fsyncs.
- **The metadata store removes its stale copy first.** `Metadata::destroy` previously removed its two copies in index order, so a crash between them could delete the copy carrying the staged intent and leave a stale copy whose watermark outlives the removed data. Removing the stale copy first keeps the newest state alive through every crash prefix of the teardown (and hardens every other metadata consumer's destroy for free).

Park-point regressions drop the production destroy future before the blob removal (both journals) and between the two checkpoint-copy removals (fixed), then reopen; without the staged intent or with index-order removal they fail with the retained-loss corruption.

### DIFF
--- a/storage/src/journal/contiguous/blobs.rs
+++ b/storage/src/journal/contiguous/blobs.rs
@@ -76,10 +76,11 @@ impl<E: Context> Partition<E> {
         }
     }
 
-    /// Scan the partition and open every existing blob as a [`Writer`], keyed by blob index.
-    pub(super) async fn open_all(&self) -> Result<BTreeMap<u64, Writer<E::Blob>>, Error> {
-        let stored = Self::scan_names(&self.context, &self.name).await?;
-
+    /// Open every named blob as a [`Writer`], keyed by blob index.
+    pub(super) async fn open_many(
+        &self,
+        stored: Vec<Vec<u8>>,
+    ) -> Result<BTreeMap<u64, Writer<E::Blob>>, Error> {
         let mut blobs = BTreeMap::new();
         for name in stored {
             let hex_name = hex(&name);
@@ -92,6 +93,12 @@ impl<E: Context> Partition<E> {
             blobs.insert(index, writer);
         }
         Ok(blobs)
+    }
+
+    /// Scan the partition and open every existing blob as a [`Writer`], keyed by blob index.
+    pub(super) async fn open_all(&self) -> Result<BTreeMap<u64, Writer<E::Blob>>, Error> {
+        let stored = Self::scan_names(&self.context, &self.name).await?;
+        self.open_many(stored).await
     }
 
     /// Remove the given blob.
@@ -114,7 +121,7 @@ impl<E: Context> Partition<E> {
     /// legacy partition (`prefix` itself) if it contains data, otherwise `{prefix}-blobs`.
     /// Both containing data is corruption.
     // TODO(#2941): Remove legacy partition support
-    pub(super) async fn select(context: &E, prefix: &str) -> Result<String, Error> {
+    pub(super) async fn select(context: &E, prefix: &str) -> Result<(String, Vec<Vec<u8>>), Error> {
         let new_partition = format!("{prefix}-blobs");
         let legacy_blobs = Self::scan_names(context, prefix).await?;
         let new_blobs = Self::scan_names(context, &new_partition).await?;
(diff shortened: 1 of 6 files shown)


storage/src/journal/contiguous/blobs.rs
```diff
@@ -398,12 +405,19 @@ impl<E: Context> Writable<E> {
 
     /// Remove every blob and the partition itself.
     pub(super) async fn destroy(self) -> Result<(), Error> {
-        let tail_blob = self.tail_blob_index();
-        drop(self.tail);
-        for blob in self.oldest_blob_index..=tail_blob {
```
patrick-ogrady, line 403: IIRC we previously did this to ensure the journal could recover (like a pruned journal) if this crashed part-way through?

storage/src/journal/contiguous/fixed.rs
```diff
@@ -412,14 +412,14 @@ impl<E: Context, A: CodecFixedShared> Journal<E, A> {
             return Self::complete_staged_clear(context, cfg, checkpoint, clear_target).await;
         }
 
-        let blob_partition = Partition::select(&context, &cfg.partition).await?;
+        let (blob_partition, stored) = Partition::select(&context, &cfg.partition).await?;
```
patrick-ogrady, line 404: ok with this change if it was just this?

storage/src/journal/contiguous/blobs.rs
patrick-ogrady: I think someone keen to destroy doesn't care what's left here but if this fails during the job, I think the journal could become unrecoverable (look corrupted)?

roberto-bayardo: Updated the PR to stage a clear intent before removing anything. Init honors a staged clear before scanning any blobs, so a reopen after any interrupted destroy wipes whatever remains unread and yields a fresh, empty journal as if the destroy had finished and a new journal were initialized. The optimization survives: one metadata sync instead of a directory fsync per blob. Added park-point regressions in both journals that drop the production destroy future, simulate a partial unordered removal, and reopen; without the staged intent both fail with the retained-loss corruption.

storage/src/journal/contiguous/fixed.rs
roberto-bayardo: While the init-scan change stands alone, the destroy half now carries a staged clear intent addressing the recovery concern on the other thread, so an interrupted destroy reopens as a fresh empty journal instead of looking corrupted.

roberto-bayardo:
@patrick-ogrady I can close this PR if we proceed with the atomic blob ops change. I will pull out the init-scan fix into its own PR.

roberto-bayardo:
superceded by the atomic blob batching work. Remaining init optimization moved to https://github.com/commonwarexyz/monorepo/pull/4379

PR #4221, opened 2026-07-09
[storage/qmdb] Use `strategy::spawn`

patrick-ogrady:
## Summary

QMDB `merkleize` previously occupied the calling task for its entire CPU duration: the serial phases ran inline and every parallel section parked the caller inside rayon's `install`. On a shared executor that pins a worker for the length of the dominant hashing phase (several ms per batch, multiple times per view in a validator). This PR runs each variant's hashing tail as a single `Strategy::spawn` job against an immutable snapshot of committed Merkle state, so the calling task yields instead of blocking. Results are byte-identical (conformance fixtures unchanged, roots matched on every benchmark iteration) and `Sequential` strategies still execute inline, leaving deterministic tests untouched.

## Changes

- **`Arc<Mem>` snapshots**: `full::Merkle` holds its `Mem` in an `Arc` (compact behind `RwLock<Arc<Mem>>`, `current`'s grafted tree likewise). Mutators go through `Arc::make_mut`, so a snapshot handed to a hashing job never observes later mutations and the copy-on-write path only triggers if a cancelled job's snapshot is still alive. This removes the `'static` barrier that previously made off-task hashing impossible (the blocker noted in #4005).
- **`Journal::merkleize`**: appends items, merkleizes, and computes the post-apply root as one CPU-bound job submitted through `Strategy::spawn`. The `any` finish tail awaits it, which covers unordered/ordered, both staged paths, and `current`'s inner tail through one seam. A dropped caller leaves the job to complete against its snapshot and discard the result; job panics are caught by `Strategy::spawn` and propagate only to an awaiter.
- **`immutable`/`keyless` (full and compact)**: `merkleize` is now `async` and delegates to the same spawned-tail mechanism (compact via `merkleize_ops`, which also folds the root computation into the job). `apply_batch` on compact variants stays sync (pure in-memory; durability remains at `commit`/`sync`).
- **`current` grafted layer**: ops-tree node reads are prefetched (`read_graft_inputs`), then graft hashing plus the grafted MMR build/merkleize run as one spawned job against the grafted-tree snapshot. Empty graft sets merkleize inline rather than paying a handoff.
- **`Rayon::spawn` drive loop**: when the polling thread is itself a pool member, the spawn future executes pending pool work via `yield_now` instead of parking on the channel. Other runtimes fall through to the channel after one membership check.
- **Deterministic `ThreadPooler`**: pools spawn no threads (worker builders are dropped; the old cooperative `thread.run()` tasks aborted the process if a suspended task ever caused them to be polled). The pool still reports the requested concurrency, so strategies partition work as configured, and everything executes on the executor thread: `install` runs the work-stealing loop on the caller and awaited spawns are driven by the yield loop. Rayon's current-thread registration is permanent, so the pool is memoized per thread (`thread_local_cache!`) and reused across pools and runners.
- **Bounds**: `ValueEncoding`/`FixedValue`/`VariableValue`/`Update` gain `'static` so owned batch state can cross the spawn boundary (API change, ALPHA).

## Benchmarks

`constantinople` shape, each variant in its own target dir, interleaved main/PR runs, roots byte-identical on every iteration:

```bash
cargo bench -p commonware-storage --bench constantinople -- <db> 0 10 1000000 32768 1 32768 8
```

Merkleize phase mean, threads=8:

| variant | main | this PR |
|---------|------|---------|
| `any::unordered::fixed::mmb` (40 iters) | 7.22 ms | 6.87 ms |
| `current::unordered::fixed::mmb` (50 iters) | 9.92 ms | 9.90 ms |

Parity within run-to-run noise at threads=8 and threads=1, as expected: the bench awaits on the criterion `block_on` thread, so total work is unchanged. The improvement is scheduling, not wall time: the awaiting task yields during the hashing phase instead of pinning its executor thread.

## Testing

- Cancellation: a new test drops a merkleize future mid-job over a two-deep unapplied batch chain, so the detached job races ancestor drops through its `Weak` walk; the database must remain fully usable (copy-on-write plus caught panics make the discarded job inert).
- Deterministic runtime: new tests pin that an awaited spawn on a multi-thread pool completes inline, that later pools (same runner or a later runner on the thread) reuse the registered pool, and that suspension with a live pool no longer aborts the process. A parallel-crate test pins the member-thread drive loop directly.
- The fused-sharded `get_many` test and the journal shard test keep `parallelism_hint = 2` under the deterministic runtime (concurrency is honored without spawning threads). The keyless fuzz target builds its Rayon lanes inside the runner via `create_strategy`, exercising pool reuse across libfuzzer iterations.
- Conformance fixtures pass unchanged across all variants.

https://claude.ai/code/session_01ExfcyhT5z9j3jaNdkvnkQj


runtime/src/deterministic.rs
```diff
@@ -1164,26 +1164,44 @@ impl crate::Spawner for Context {
     }
 }
 
+// The pool the executor thread registered with, returned for every request on the thread.
+// Rayon permits one registry registration per OS thread and the registration is permanent,
+// so only the first pool can register the executor thread. Work submitted to any other pool
+// could never execute (spawning threads would be nondeterministic, so pools have no
+// workers).
+commonware_utils::thread_local_cache!(static THREAD_POOL: ThreadPool);
```
clabby, line 1172: cool to see this used again :)

storage/src/qmdb/current/batch.rs
```diff
@@ -821,28 +821,38 @@ where
...
+    let grafted_batch = if graft_inputs.is_empty() {
+        grafted_parent
+            .new_batch()
+            .merkleize(&current_db.grafted_tree, &grafted_hasher)
+    } else {
+        let grafted_parent = Arc::clone(grafted_parent);
+        let grafted_tree = Arc::clone(&current_db.grafted_tree);
+        current_db
+            .strategy
+            .clone()
```
clabby, line 839: is this clone necessary? `spawn` takes `&self`.

runtime/src/deterministic.rs
```diff
@@ -1164,26 +1164,44 @@ impl crate::Spawner for Context {
...
-            .build()
-            .map(Arc::new)
+                // Dropping the builder leaves the worker permanently unstarted: no thread
+                // is spawned, and a worker loop hosted as a task the executor polls would
+                // block or abort the runtime.
+                builder.spawn_handler(|_| Ok(())).build().map(Arc::new)
+            },
+            |_| Ok(()),
+        )?;
+        Ok(Arc::clone(&pool))
```
clabby, line 1204: This should probably panic if a new thread pool creation is attempted with a different `concurrency` param? If one has already been created, this will just silently return you a thread pool with a different number of threads configured.

patrick-ogrady: ah, that was a mistake.

PR #4222, opened 2026-07-09
[consensus/simplex] Coalesce `fsync`

patrick-ogrady:
## Summary

The `Voter` previously synced its WAL once per artifact it was about to share: a single event loop iteration could fsync the same section several times (nullify, notarize, notarization, nullification, finalize, finalization), and certification results paid their own sync on arrival. This PR coalesces all appends made in an event loop iteration into a single `Journal::sync`.

## Design

Every iteration now ends with three explicit phases:

```rust
let staged = self.construct(...).await;           // build + record (journal appends)
self.sync_journal(view).await;                    // the only sync call site
self.notify(..., staged, nullify, certification); // broadcast + report to the application
```

- Every append in an iteration targets the view being processed, so `append_journal` just sets a `dirty` flag and `sync_journal(view)` flushes that one section. The sync runs unconditionally, even when nothing will be broadcast (e.g. an iteration that only records a certification result), and `notify` asserts the flag is clear before sending anything. The single-view invariant is documented on `append_journal` and was verified per event loop arm: certificates append at the view they certify (which is the view returned for notification), the timeout arm's nullify appends at the current view (a nullify vote never advances the view), and `construct` appends only at `view`.
- Message-processing arms never send. The timeout arm constructs and records its nullify vote (and fetches the previous view's best certificate on retry), and the certify arm records the certification result. Both are staged as iteration locals and passed to `notify` alongside the `Staged` messages built by `construct`. Nullify construction cannot move into `construct` because `construct_nullify` is gated by the deadline firing, not by state.
- The nullification "floor" broadcast moved from `process_message` into `prepare_nullification`, keyed to the once-per-view `broadcast_nullification` flag. The old floor plumbing through `handle_nullification`'s return value was dead on the prepare path (`broadcast_nullification` only returns certificates already stored by `add_nullification`), so the relocation is behavior-preserving: same certificate, same iteration, emitted once. `process_message` no longer takes a sender at all.

## Contract change

A certification result is now durable at the end of the iteration that recorded it rather than immediately after the automaton resolves, so an unclean shutdown can lose a resolved-but-unsynced result. On restart, replay leaves the round's certification state unset and `certify_candidates` re-requests it. The `CertifiableAutomaton::certify` docs now state this restart exception, and the same caveat was added to `Automaton::verify` (the identical window has always existed for verification verdicts, whose only durable record is the notarize vote they produce).

## Safety argument

A vote must be durable before it reaches the network so a restart cannot cause equivocation. That invariant is preserved and now enforced structurally: all sends happen in `notify`, which runs strictly after the barrier and asserts the journal is clean. An adversarial review (six dimensions, per-finding refutation panels) confirmed:

- Every append is covered by the barrier in the same iteration (no `continue` path appends), and the shutdown path still runs `sync_all`.
- The batcher has no network egress for votes or certificates: a certificate embedding our vote can only return via the voter mailbox in a later iteration, after the barrier that made the vote durable.
- The resolver serves only floor certificates and nullifications to peers, and promotes notarizations to servable only after certification (a strictly later, post-barrier iteration), so pre-barrier `resolver.updated` calls cannot advertise a certificate embedding a non-durable vote.
- A crash mid-barrier loses only artifacts that were never sent.

This also structurally closes the earlier Bugbot finding (certification artifacts left unsynced when nothing was broadcast).

## Notes

- The `simplex.voter.journal.sync` span now fires once per iteration instead of once per artifact.
- Test infrastructure: `runtime::mocks` gains `PendingSyncs`, a deferred-sync parking queue with a one-shot `arm` gate that blocks the next durability operation and counts operations from that point on (once the gate fires, later syncs park as usual). The new certification-replay restart test uses it to hold the section sync open and observe that certification and finalize share one fsync. Shared context-wrapper trait forwarding is generated by a `forward_context!` macro.
- Verified with the full simplex suite (301 tests), the voter actor module (111), and the slow-profile `unclean_shutdown`/`all_recovery` fixtures (28), all passing.
- Follow-up: no deterministic test pins the construct/sync/notify ordering itself (the assert covers the call graph, not a reordering of the phases).

### DIFF
--- a/consensus/src/simplex/actors/voter/actor.rs
+++ b/consensus/src/simplex/actors/voter/actor.rs
@@ -253,10 +253,11 @@ impl<
 
     /// Syncs all journal sections with pending appends.
     ///
-    /// Invoked before any broadcast so everything we tell the network is
-    /// recoverable after a restart. Deferring syncs to this point (rather than
-    /// syncing after each append) coalesces back-to-back appends in the same
-    /// loop iteration into a single sync.
+    /// Invoked after each construction phase and before the corresponding
+    /// broadcast phase (regardless of whether anything will be broadcast) so
+    /// everything we tell the network is recoverable after a restart. Deferring
+    /// syncs to this boundary (rather than syncing after each append) coalesces
+    /// all appends in the same loop iteration into a single sync.
     async fn sync_journal(&mut self) {
         if self.dirty.is_empty() {
             return;
@@ -265,7 +266,7 @@ impl<
             .journal
             .as_mut()
             .expect("pending journal sections without a journal");
-        let sections: Vec<u64> = mem::take(&mut self.dirty).into_iter().collect();
+        let sections = mem::take(&mut self.dirty);
         let span = info_span!(
             "simplex.voter.journal.sync",
             epoch = self.state.epoch().traced(),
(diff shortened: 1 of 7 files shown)


consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -271,11 +303,17 @@ impl<
...
-    fn broadcast_certificate<T: Sender>(
+    ///
+    /// Syncs pending journal appends first so any state we advertise to the
+    /// network survives a restart.
+    async fn broadcast_certificate<T: Sender>(
         &mut self,
         sender: &mut WrappedSender<T, Certificate<S, D>>,
         certificate: Certificate<S, D>,
     ) {
+        // Persist anything we've journaled before the certificate reaches the network
```
patrick-ogrady: Attempt to sync journal after construction phase and before broadcast phase (regardless if sending anything)?

PR #4223, opened 2026-07-10
[runtime] Fix Concurrency

patrick-ogrady:
Related: #4221

### DIFF
--- a/runtime/src/tokio/runtime.rs
+++ b/runtime/src/tokio/runtime.rs
@@ -28,7 +28,7 @@ use crate::{
 };
 use commonware_macros::{select, stability};
 #[stability(BETA)]
-use commonware_parallel::{Error as ParallelError, Rayon};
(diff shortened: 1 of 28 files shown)


runtime/src/tokio/runtime.rs
```diff
@@ -637,11 +637,8 @@ impl crate::Spawner for Context {
 
 #[stability(BETA)]
 impl crate::ThreadPooler for Context {
-    fn create_thread_pool(
```
patrick-ogrady, line 640: We originally did create_thread_pool so people could use rayon outside of our impl...with the nuance involved in doing this right on deterministic, I'm not sure we can continue to offer that API.

patrick-ogrady: In which case, we can call this strategizer instead of thread pooler.

PR #4224, opened 2026-07-10
[consensus/simplex] Offload to Strategy

patrick-ogrady:
## Summary

The batcher previously ran all CPU-bound signature work inline on its event-loop task: batch vote verification (`verify_attestations`) and certificate recovery (`assemble`) executed on the calling thread, with a parallel `Strategy` only parallelizing the interior of each call. This PR submits both as jobs through `Strategy::spawn`, so a parallel strategy hosts the work on its own pool and the batcher's runtime thread stays free while it runs.

## Design

- A new `offload` helper (`batcher/verifier.rs`) runs a CPU-bound job through `Strategy::spawn`, entering the caller's span on the worker thread and instrumenting the awaited future so the offloaded work stays attributed to the caller's trace.
- `Verifier::verify_{notarizes,nullifies,finalizes}` are now async. Each moves the pending batch and the accumulated verified votes into the spawned job (together with an `Arc` of the scheme and a `StdRng` forked from the caller's RNG) and reinstalls the extended verified vector on completion. The return value is now `(batch size, invalid signers)` because verified votes never leave the verifier.
- The verifier stores verified votes instead of counting them: `pending_*`/`verified_*` vector pairs replace the `*_verified` counters, and `take_verified_*` hands the votes to certificate recovery once they reach quorum. This makes the verifier the single store feeding certificate construction and deletes the round's duplicate `verified_votes` tracker (`add_verified` is gone). The round's remaining `votes` tracker serves duplicate detection, conflict reporting, and missing-voter queries.
- `Round::try_construct_{notarization,nullification,finalization}` are async and assemble the certificate on the strategy's pool via new owned-vote constructors (`Notarization::from_owned_notarizes`, `Nullification::from_owned_nullifies`, `Finalization::from_owned_finalizes`) that consume votes rather than cloning each attestation. The borrowed `from_*` constructors are now thin wrappers over the owned ones. Assembly from a verified quorum must succeed, so the offloaded job `expect`s instead of silently returning `None`.
- The round no longer caches certificates it never reads: `set_*(certificate.clone())` became `mark_*()` booleans. Only existence was ever checked, and the certificate itself is forwarded to the voter.
- The batcher and verifier share the scheme through `Arc<S>` (participants are read through it), so worker closures capture a cheap clone instead of cloning the scheme per round.
- `Timed::time_some` now takes a future so it can time the async recovery paths.

## Behavior notes

- Cancellation: once polled, verification and recovery move the votes into the worker, so an abandoned future loses them. This is documented on `Verifier` and the `try_construct_*` methods. The batcher awaits these inline in its event loop (never raced in a `select!`), so the caveat only binds future callers.
- `set_leader` now also searches verified notarizes for the leader's vote. Previously only pending votes were searched, so a leader vote added as already-verified never set the leader proposal. `set_leader_proposal` filters the verified vectors as well.

## Tests

- Verifier unit tests moved to the async surface with `#[test_async]`, with new coverage for verified-vote leader detection and verified-vote proposal filtering.
- New `test_offload_on_multithreaded_pool` drives the round's verify and recovery offloads on the tokio runtime with a two-worker pool, so jobs execute on pool threads and completion must wake the awaiting task across threads. The deterministic runtime cannot exercise this because its shared pool is single-threaded and `Rayon::spawn` short-circuits to the calling thread.
- `test_for_all_fixtures!` gains an argument-passing form. `all_online` now takes a strategy factory, and a `Rayon(2)` variant runs end-to-end on the deterministic runtime (`test_all_online_rayon_bls12381_threshold_vrf_min_pk`).

### DIFF
--- a/consensus/src/simplex/actors/batcher/actor.rs
+++ b/consensus/src/simplex/actors/batcher/actor.rs
@@ -28,7 +28,7 @@ use commonware_runtime::{
     },
     Clock, ContextCell, Handle, Metrics, Spawner,
 };
-use commonware_utils::ordered::{Quorum, Set};
+use commonware_utils::ordered::Quorum;
 use rand_core::CryptoRng;
 use std::{collections::BTreeMap, sync::Arc};
 use tracing::{debug, info_span, trace, Instrument as _, Span};
@@ -53,7 +53,6 @@ where
 {
     context: ContextCell<E>,
 
-    participants: Set<S::PublicKey>,
     scheme: Arc<S>,
 
     blocker: B,
@@ -90,7 +89,7 @@ where
 {
     pub fn new(context: E, cfg: Config<S, B, Re, Rl, T>) -> (Self, Mailbox<S, D>) {
         let scheme = Arc::new(cfg.scheme);
-        let participants = scheme.participants().clone();
+        let participants = scheme.participants();
         let participant_count = participants.len();
         let added = context.counter("added", "number of messages added to the verifier");
         let verified = context.counter("verified", "number of messages verified");
@@ -120,7 +119,6 @@ where
             Self {
                 context: ContextCell::new(context),
 
-                participants,
                 scheme,
 
                 blocker: cfg.blocker,
@@ -151,7 +149,6 @@ where
     fn new_round(&self, view: View) -> Round<S, B, D, Re> {
         Round::new(
             Rnd::new(self.epoch, view),
-            self.participants.clone(),
             Arc::clone(&self.scheme),
             self.blocker.clone(),
             self.reporter.clone(),
@@ -164,7 +161,7 @@ where
     /// it meant to be). If a peer sends us a certificate very far in the future,
     /// we will record that as their latest activity (and not attempt to skip them).
     fn record_activity(&mut self, sender: &S::PublicKey, view: View) {
-        let Some(participant) = self.participants.index(sender) else {
+        let Some(participant) = self.scheme.participants().index(sender) else {
             return;
         };
         let seen_view = &mut self.latest_seen[usize::from(participant)];
@@ -197,7 +194,7 @@ where
         missing
             .iter()
             .filter(|&&p| Some(p) != me)
-            .filter_map(|&p| self.participants.key(p).cloned())
+            .filter_map(|&p| self.scheme.participants().key(p).cloned())
             .collect()
     }
 
@@ -605,18 +602,16 @@ where
                     };
 
                     // Process batch verification results
-                    if let Some((voters, failed)) = verified {
+                    if let Some((batch, failed)) = verified {
                         timer.observe(self.context.as_ref());
 
-                        // Process verified votes
-                        let batch = voters.len() + failed.len();
                         trace!(%updated_view, batch, "batch verified votes");
                         self.verified.inc_by(batch as u64);
                         self.batch_size.observe(batch as f64);
 
                         // Block invalid signers
                         for invalid in failed {
-                            if let Some(signer) = self.participants.key(invalid) {
+                            if let Some(signer) = self.scheme.participants().key(invalid) {
                                 commonware_p2p::block!(
                                     self.blocker,
                                     signer.clone(),
@@ -625,10 +620,6 @@ where
                             }
                         }
 
-                        // Store verified votes for certificate construction
-                        for valid in voters {
-                            round.add_verified(valid);
-                        }
                     } else {
                         trace!(
                             current = %current.view,
--- a/consensus/src/simplex/actors/batcher/actor.rs
+++ b/consensus/src/simplex/actors/batcher/actor.rs
@@ -106,12 +106,12 @@ where
         );
         let verify_latency = context.histogram(
             "verify_latency",
-            "latency of signature verification (including strategy job scheduling)",
+            "latency of signature verification",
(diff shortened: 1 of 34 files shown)


consensus/src/simplex/actors/batcher/round.rs
```diff
@@ -421,92 +426,158 @@ impl<
...
+                        .expect("verified notarize quorum must assemble");
+                    Notarization {
+                        proposal,
+                        certificate,
+                    }
+                })
+            })
+            .instrument(span)
+            .await;
+        self.mark_notarized();
```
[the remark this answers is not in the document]
patrick-ogrady: This was intended.

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -106,21 +106,20 @@ where
         );
         let verify_latency = context.histogram(
             "verify_latency",
-            "latency of signature verification",
+            "latency of signature verification (including strategy job scheduling)",
```
patrick-ogrady: remove the ` (including strategy job scheduling)`

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -106,21 +106,20 @@ where
...
             Buckets::CRYPTOGRAPHY,
         );
         let recover_latency = context.histogram(
             "recover_latency",
-            "certificate recover latency",
+            "certificate recover latency (including strategy job scheduling)",
```
patrick-ogrady: remove ()

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -442,28 +440,32 @@ where
                     epoch = self.epoch.traced(),
                     view = view.traced()
                 );
+
```
patrick-ogrady: remove \n?

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -582,34 +587,31 @@ where
...
                     };
 
                     // Process batch verification results
-                    if let Some((voters, failed)) = verified {
+                    if let Some((batch, failed)) = verified {
                         timer.observe(self.context.as_ref());
 
-                        // Process verified votes
-                        let batch = voters.len() + failed.len();
                         trace!(%updated_view, batch, "batch verified votes");
```
patrick-ogrady, line 608: restore comment above code block?

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -618,10 +620,6 @@ where
                             }
                         }
 
-                        // Store verified votes for certificate construction
```
patrick-ogrady, line 621: Why was this removed?

consensus/src/simplex/actors/batcher/round.rs
```diff
@@ -41,18 +39,14 @@ pub struct Round<
     /// Votes received from network (may not be verified yet).
     /// Used for duplicate detection and conflict reporting.
     pending_votes: VoteTracker<S, D>,
-    /// Votes that have been verified through batch verification.
-    /// Only these votes are used for certificate construction.
-    verified_votes: VoteTracker<S, D>,
-
     /// Whether we've already sent the leader's proposal to the voter.
```
patrick-ogrady, line 42: add a \n before?

consensus/src/simplex/actors/batcher/verifier.rs
```diff
@@ -20,13 +24,19 @@ use rand_core::CryptoRng;
...
 /// [ed25519]: crate::simplex::scheme::ed25519
 /// [bls12381_multisig]: crate::simplex::scheme::bls12381_multisig
 /// [bls12381_threshold]: crate::simplex::scheme::bls12381_threshold
 /// [secp256r1]: crate::simplex::scheme::secp256r1
 pub struct Verifier<S: Scheme<D>, D: Digest> {
     /// Signing scheme used to verify votes and assemble certificates.
-    scheme: S,
+    ///
+    /// Held in an [Arc] so verification jobs can capture the scheme without
+    /// copying the participant set.
```
patrick-ogrady: This isn't necessary:

    ///
    /// Held in an [Arc] so verification jobs can capture the scheme without
    /// copying the participant set.

consensus/src/simplex/actors/batcher/verifier.rs
```diff
@@ -72,24 +80,54 @@ impl<S: Scheme<D>, D: Digest> Verifier<S, D> {
             leader_proposal: None,
 
             notarizes: Vec::new(),
-            notarizes_verified: 0,
+            verified_notarizes: Vec::new(),
 
             nullifies: Vec::new(),
-            nullifies_verified: 0,
+            verified_nullifies: Vec::new(),
```
patrick-ogrady: presize to quorum?

consensus/src/simplex/mod.rs
```diff
@@ -790,11 +790,23 @@ mod tests {
             .count() as u32
     }
 
-    fn all_online<S, F, L>(mut fixture: F)
+    fn all_online<S, F, L>(fixture: F)
```
patrick-ogrady: just add an arg to all_online with strategy ...don't create a new function

patrick-ogrady (commented):
almost there.

PR #4228, opened 2026-07-10
[runtime] Make buffered writers cancellation-safe; make paged appends synchronous

danlaine:
Dropping a storage future at an `.await` (via `select!`, timeouts, or shutdown) could leave the buffer layer's in-memory state disagreeing with the blob: lost buffered data, forgotten resizes, and partially published pages. This PR makes the two buffered writers safe under cancellation and removes the largest class of cancellation points entirely by making appends synchronous.

### 1. `buffer::Write` is cancellation-safe

`SyncState` becomes a struct: the old `Clean`/`Dirty`/`Pending` enum (now `Status`) plus a `pending_resize: Option<u64>` intent.

- A cancelled flush no longer discards the tip; the next operation retries the write.
- A cancelled `resize` is recorded before I/O. Later blob operations finish that resize before doing new work.
- Shrinks adopt the shorter size in memory first, so a dropped shrink never leaves the writer claiming bytes the blob no longer holds.
- A later blob operation reports an unobserved started-sync failure.

### 2. `paged::Writer::append` / `append_owned` are synchronous

```rust
// before
pub async fn append(&mut self, buf: &[u8]) -> Result<u64, Error>
// after
pub fn append(&mut self, buf: &[u8]) -> u64
```

Appends extend the writer's in-memory tail and cannot fail or be cancelled. Disk I/O moves to `sync()`, `start_sync()`, `resize()`, `seal()`, `snapshot()`, and `replay()`. This removes every mid-append cancellation window and lets callers batch appends before paying for I/O.

The tail has two parts:

- Full pages are held as immutable, refcounted slices until flush. Large `append_owned` calls retain their complete pages without copying.
- The mutable tip holds the final partial page. Only bytes needed to complete that page and its final suffix are copied.

Before writing, a flush moves every complete page out of the tip. It then writes the full pages and partial tip in order, updates the page cache, and advances the writer only after the writes succeed. A cancelled flush therefore leaves the tail intact for retry.

`capacity` is the preferred allocation size for the mutable tip, not a bound on all pending data. Callers bound pending data through their flush cadence. The old large-append path instead issued I/O during append; the new path defers it to the caller's existing barrier.

The paged writer also repairs cancellation during page publication and shrinking: later operations clean up uncommitted page bytes before changing their CRC record, and re-read an uncertain partial-page record after a cancelled shrink.

### 3. Supporting changes

- `tip::Buffer` is now the small copy-on-write byte window shared by raw and paged writers. It keeps bytes until the caller advances it after a successful write.
- New `paged::tail` code owns page splitting, zero-copy full-page retention, and tail reads. The shared read path sees one in-memory tail rather than separately reasoning about page slices and a tip.
- Callers in contiguous and segmented journals, fuzz targets, and benches drop `.await` on appends. Appends now remain in memory until their existing barrier; durability points are unchanged.

## Testing

- Gated blob helpers (`Gate`, `GatedWriteBlob`, `GatedResizeBlob`) stop a specific write or resize so tests can drop a future at an exact await.
- Cancel-then-retry coverage includes flushes, grows, resizes, page publication, partial-page shrink, and full-page tail writes.
- Tests verify that full pages from large owned appends remain zero-copy, reads work across the persisted prefix, full tail pages, and the tip, and a cancelled flush preserves the full tail.
- Buffer fuzz targets (`buffer`, `blob_integrity`) were adapted. The append benchmark flushes each iteration, measuring staged append plus sync.

## Cancellation contract

Dropping any `buffer::Write` or `paged::Writer` future leaves the writer usable. Its tip, full tail pages, and resize intent remain available to the next operation or reopen. Appends have no cancellation points.

### DIFF
--- a/runtime/fuzz/fuzz_targets/blob_integrity.rs
+++ b/runtime/fuzz/fuzz_targets/blob_integrity.rs
@@ -118,10 +118,7 @@ fn fuzz(input: FuzzInput) {
             .await
             .expect("cannot create append wrapper");
 
-        append
-            .append(&expected_data)
-            .await
-            .expect("cannot append data");
+        append.append(&expected_data);
         append.sync().await.expect("cannot sync");
         drop(append);
 
--- a/runtime/fuzz/fuzz_targets/buffer.rs
+++ b/runtime/fuzz/fuzz_targets/buffer.rs
@@ -255,7 +255,7 @@ fn fuzz(input: FuzzInput) {
                         };
                         let current_size = append.size();
                         if current_size.checked_add(data.len() as u64).is_some() {
-                            let _ = append.append(&data).await;
+                            append.append(&data);
                         }
                     }
                 }
--- a/runtime/src/utils/buffer/benches/append.rs
+++ b/runtime/src/utils/buffer/benches/append.rs
@@ -29,7 +29,7 @@ where
                     for _ in 0..iters {
                         // Write double the bytes that can be held by the cache.
                         for _ in 0..(CACHE_SIZE * PAGE_SIZE.get() as usize / chunk_size) * 2 {
-                            append.append(&data).await.unwrap();
+                            append.append(&data);
                         }
                     }
                     let elapsed = start.elapsed();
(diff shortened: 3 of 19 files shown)


roberto-bayardo:
This basically undoes https://github.com/commonwarexyz/monorepo/pull/4015 so I guess we lose the performance wins?

(The cancellation safety wins may still make it worth it, but would be good to know.)

danlaine:
> This basically undoes #4015 so I guess we lose the performance wins?
> 
> (The cancellation safety wins may still make it worth it, but would be good to know.)

Working on mitigating this by making the writer tip a list of `IoBuf` instead of a `Vec<u8>`

danlaine:

PR #4229, opened 2026-07-10
[cryptography/ed25519] Improve Decompression

patrick-ogrady:
Batch ed25519 verification was bound by signature R-point decompression (a ~251-squaring serial chain per point that dalek evaluates one element at a time) and by fragmenting the multiscalar multiplication into monolithic per-shard calls. This PR keeps the compressed wire format and rebuilds the verification pipeline on owned, batched primitives.

## Optimizations

- **Batched point decompression** (`core/fe.rs`, `core/neon.rs`, `core/ifma.rs`): an owned vartime fe51 field implementation runs four independent square-root chains at once. Scalar-interleaved kernel (ILP) is ~2x faster per point than dalek; the NEON kernel (aarch64) doubles that again (~4x, 2085 -> ~550ns/point). An AVX-512 IFMA kernel ships for x86_64 behind runtime detection, written against a 9-op backend trait with a bit-exact scalar model so all schedule/carry/chain logic is differentially tested on any host; the intrinsic mapping is exercised on IFMA hardware (hardware-vs-model test self-skips elsewhere).
- **One global verification equation, phased for parallelism** (`core/batch.rs`, `core/msm.rs`): instead of independent per-shard equations, chunks of signatures are prepared (s-checks, SHA-512, per-chunk coalescing, batched decompression) as fine-grained work-stealing tasks, then a single owned dual-column Pippenger runs with **per-window parallel tasks** (253-bit terms for the basepoint + coalesced keys; 128-bit randomizer terms populate only the low windows). Small term counts use a mixed-column NAF Straus (dalek's Straus/Pippenger pattern). This removes both MSM fragmentation and slowest-core gating.
- **Owned single verification with half-width scalars** (`core/verification_key.rs`): `verify_prehashed` uses an owned NAF Straus, and both scalars split at bit 127 against precomputed `[2^127]A` (stored per key, +64B) and a static `[2^127]B` table, halving the shared doubling chain (254 -> 128). `VerificationKey` stores the point in the owned representation only (`minus_A` removed). Key parsing does one square root plus the `[2^127]A` precomputation (~2us -> ~11us vs main), amortized to zero wherever keys are cached and repaid after one verification otherwise. dalek remains for `Scalar` arithmetic and signing only.
- **Parallel multiscalar setup**: digit decomposition and point caching run as chunked parallel tasks instead of a serial pass gating the window tasks (worth ~15% on the large-batch parallel wall alone).
- **Adaptive path selection** (`commonware-parallel`): new `Strategy::try_run_n` runs one of N parallel arms chosen by the existing adaptive policy (entry gains per-arm estimates; boundary probes alternate losers; the serial budget projects from the best arm; `Sequential` deterministically runs the serial body). Batch verification passes its two parallel structures (per-thread equations vs the phased global path) as arms, eliminating the hand-measured crossover constant.
- ZIP215 parity is pinned differentially against dalek: byte-exact decompression semantics (non-canonical y, "-0", torsion points, 65k random encodings), point ops across all torsion pairs, MSM vs dalek across algorithm arms, plus end-to-end batch tests for small-order acceptance and (R, s+l) malleation rejection on every routing path.

## Results (M5 Pro, criterion median vs main)

| batch | distinct conc=1 | distinct conc=8 | same-signer conc=1 | same-signer conc=8 |
|---|---|---|---|---|
| 10000 | **-35%** (69.2->44.9ms) | **-34%** (12.4->8.2ms) | **-35%** (43.5->28.5ms) | **-39%** (8.2->5.0ms) |
| 1000 | -25% | -33% | -34% | -28% |
| 100 | -12% | -4% | -21% | -7% |
| single verify | 22.0us (vs ~29-31us, -25 to -29%) | | | |

Only regression in the matrix: a batch of exactly one signature pays ~+2.5% (~0.8us) vs the old dalek combined path.

## Measured and rejected

- Batch-affine Pippenger buckets: affine Edwards addition has two denominators, costing ~12M/add even with Montgomery-batched inversion vs the 7M extended mixed add (a Weierstrass-only technique).
- SIMD bucket-scan additions (SoA over 4 independent adds): gate prototype measured 1.05x — randomly-indexed accumulators force ~11 field transposes per add, consuming the 1.7-1.9x vector-multiply gain.
- +-l/2 wide-scalar recoding: ~1% on one size band, carried a small-order-component contract subtlety.

## Notes

- Sarkar-style table-accelerated square roots: refuted for this field (2-adicity 2 means no Tonelli-Shanks loop to replace; the fixed 252-bit variable-base exponentiation is at its addition-chain floor).
- Follow-up candidate (stacks with this PR): cross-batch precomputed tables for repeated validator keys.
- x86_64 (c8a) validation is the open follow-up: the owned MSM replaces dalek's AVX2/IFMA multiscalar path there; the IFMA decompression kernel compiles (cross-checked) but needs IFMA hardware to benchmark. Window tables and kernel widths are tunable constants.
- No wire or storage format changes; no public API changes; conformance fixtures untouched.

https://claude.ai/code/session_01Dp5uJxtWYGvq31P2RUJ4XP


parallel/src/lib.rs
```diff
@@ -165,6 +165,30 @@ commonware_macros::stability_scope!(BETA {
...
+        /// strategy chose. Every arm must compute the same result.
+        ///
+        /// Adaptive strategies learn the fastest of the serial body and each
+        /// parallel arm per callsite and input size. Non-adaptive strategies
+        /// run the serial body, or arm 0 when configured for parallelism
+        /// without an adaptive policy.
+        ///
+        /// `arms` must be between 1 and 2.
+        #[track_caller]
+        fn try_run_n<R, E, SEQ, PAR>(
```
patrick-ogrady, line 179: This API is pretty bad...the idea is that we may have "many good options" we want to explore as variants for doing the same thing.

cryptography/src/ed25519/core/ifma.rs
```diff
@@ -0,0 +1,601 @@
...
+        Self([
+            B::madd52lo(B::and(z[0], mask), c4, r19),
+            B::add(B::and(z[1], mask), c0),
+            B::add(B::and(z[2], mask), c1),
+            B::add(B::and(z[3], mask), c2),
+            B::add(B::and(z[4], mask), c3),
+        ])
+    }
+
+    /// Lane-wise multiplication (dalek's IFMA schedule).
```
patrick-ogrady, line 210: Add reference links to all of this

PR #4230, opened 2026-07-10
[consensus/glue] Fix State Sync Startup

patrick-ogrady:
## Summary

The bug: consumer wiring (constantinople's engine setup) recovered its highest stored finalization on every restart and passed it directly to marshal as `Start::Floor`, bypassing `SyncPlan::marshal_start`. Used through the plan, a floor never exceeded durable progress after the sync boot — but that safety was a convention, not a structure. A floor above marshal's durable progress makes marshal jump — it anchors on the floor block, records the anchor's predecessor as processed, and prunes everything below the anchor — so stateful startup, which computed its floor as `max(stored sync height, processed height)` and `expect`ed marshal to retain that block, panicked on healthy nodes whose databases sat at the anchor (intermittently, since prune alignment sometimes spares the predecessor).

The fix makes the floor contract structural: a floor is provided exactly once — the finalization peer state sync begins from — and it is persisted in the sync metadata. `SyncPlan::marshal_start` re-provides that stored floor as marshal's startup anchor on every future boot, so the floor a node passes never advances after its sync: marshal ignores it (or re-installs it idempotently in the brief window before the first post-sync acknowledgement) once durable progress covers it, and it can never jump marshal above the databases again. Startup reconciliation then holds marshal to it:

- The block at the resolved floor (`max(stored sync height, processed height)`, genesis for fresh nodes) must be retained: a miss means a floor was provided outside the plan or storage was corrupted, and startup panics.
- Databases ahead of the floor are rewound, as before. Outside of state sync, blocks are acknowledged strictly after every database is durably finalized, so crashes (including torn multi-database commits) only produce this direction.
- Databases behind the floor cannot rewind forward: QMDB rejects the rewind before misapplying anything, so startup fails fatally. That state is unreachable without storage corruption or a marshal floor above the databases.
- Finalized blocks redelivered at or below the recovered floor are reported to the application and acknowledged without being reapplied — including genesis on a fresh boot, which was previously swallowed.
- `Start::Floor`'s interaction with durable progress is pinned in consensus for both directions (a floor above progress jumps and prunes, a re-provided floor at or below progress is superseded without pruning), which previously tested configured floors and durable progress separately but never their intersection.

An interrupted state sync also benefits from the persisted floor: it resumes from it when no new floor is probed, and a probed floor is used only when strictly newer (by round). This fixes the one fragility that was reachable even with correct wiring — an interrupted-sync restart panicked when peers sampled near the tip returned a floor behind the one already durably armed.

`SyncPlan::init` now takes the scheme's certificate codec configuration to read the stored floor finalization.

## Validation

`just test --profile all -p commonware-glue` (172 tests, including the crash/restart and crash-during-state-sync simulations) and the new consensus `Start::Floor` regression test pass. Changed-crate clippy, `just check-fmt`, `just check-stability`, and rustdoc are all clean.


glue/src/stateful/mod.rs
```diff
@@ -38,17 +38,20 @@
...
+//! processed height instead.
 //!
 //! The actor supports two sync paths:
 //!
 //! - **Marshal sync** (no floor attached): [`Stateful::start`] prepares the
 //!   databases before the actor is spawned. New nodes initialize from
 //!   genesis; restarted nodes reconcile the database set against the later of
 //!   marshal's processed anchor and the stored state sync height, rewinding if
-//!   needed. If marshal is behind that stored height, the actor acknowledges old
+//!   needed. Databases that instead recover BEHIND that floor are caught up by
```
patrick-ogrady: don't capitalize?

glue/src/stateful/mod.rs
```diff
@@ -38,17 +38,20 @@
...
+//!   replaying retained finalized blocks, falling back to a one-time peer
+//!   state sync at the floor when the replay window has been pruned. If
```
patrick-ogrady: > falling back to a one-time peer
//!   state sync at the floor when the replay window has been pruned. 

Don't fallback...at this point we require manual intervention (just say we can't recover because peers don't have blocks)? I think we should attempt to continue, even if this is true.

glue/src/stateful/db/mod.rs
```diff
@@ -219,6 +219,13 @@ pub trait ManagedDb<E>: Send + Sync + Sized {
     /// Return the sync target for this database's current committed state.
     fn sync_target(&self) -> Self::SyncTarget;
 
+    /// Return true when the database's committed state is BEHIND `target`.
```
patrick-ogrady: again, don't capitalize?

glue/src/stateful/db/mod.rs
```diff
@@ -219,6 +219,13 @@ pub trait ManagedDb<E>: Send + Sync + Sized {
...
+    ///
+    /// Reaching a behind target would require applying operations the database
+    /// does not have, so rewind repair cannot succeed and the caller must
+    /// recover via replay or peer state sync instead.
```
patrick-ogrady: not peer state sync?

glue/src/stateful/db/mod.rs
```diff
@@ -288,6 +295,13 @@ pub trait DatabaseSet<E>: Clone + Send + Sync + 'static {
     /// Return sync targets for the set's current committed state.
     fn committed_targets(&self) -> impl Future<Output = Self::SyncTargets> + Send;
 
+    /// Return true when any database in the set is behind its corresponding
+    /// target (see [`ManagedDb::behind_sync_target`]). A behind set cannot be
+    /// repaired by rewind: recovery must replay the gap or run peer state
```
patrick-ogrady: not peer state sync?

glue/src/stateful/actor/syncer/plan.rs
```diff
@@ -21,9 +21,10 @@ use commonware_storage::Context;
 /// [`Stateful`](crate::stateful::Stateful), so startup does not reopen the same
 /// metadata partition from multiple places.
 ///
-/// Once state sync completes, this node never performs peer state sync
-/// again. Future startups must recover from the later of that synced height
-/// and marshal's processed height instead.
+/// Once state sync completes, this plan never selects peer state sync again.
+/// Future startups recover from the later of that synced height and marshal's
+/// processed height instead, re-running peer state sync only when the
```
patrick-ogrady: > re-running peer state sync only when the
/// databases recover behind that floor and the replay window has been pruned.

In the sentence before, you literally said never selects state sync again! Keep to that!

consensus/src/marshal/config.rs
```diff
@@ -14,7 +14,13 @@ use std::num::{NonZeroU64, NonZeroUsize};
 /// Startup anchor for marshal.
 ///
 /// Durable progress from a previous run takes precedence when it already
-/// supersedes the configured anchor.
+/// supersedes the configured anchor. In the other direction, a [Start::Floor]
+/// ABOVE durable progress jumps ahead exactly like a runtime
```
patrick-ogrady: don't captialize + slim down this comment?

glue/src/stateful/mod.rs
patrick-ogrady: Done in 9e71f8d9: de-capitalized here and swept the remaining BEHIND/AHEAD/ABOVE/JUMPS emphasis across the PR.

glue/src/stateful/mod.rs
patrick-ogrady: Removed the fallback entirely (9e71f8d9 through 8d61d4a3): recovery is always local replay, and a missing replay block is now an `unreachable!` panic. It cannot happen without storage corruption or a marshal floor above databases that never synced (marshal's processed height only advances after the databases durably finalize, pruning retains the ack window below it, and marshal never requests blocks at or below its floor from peers).

glue/src/stateful/db/mod.rs
patrick-ogrady: Done in 9e71f8d9.

glue/src/stateful/db/mod.rs
patrick-ogrady: Done: replay only (9e71f8d9).

glue/src/stateful/db/mod.rs
patrick-ogrady: Done: replay only (9e71f8d9).

glue/src/stateful/actor/syncer/plan.rs
patrick-ogrady: Done: the re-run clause is gone and the code now keeps to it unconditionally. `begin_sync` after a completed sync is `unreachable!` (9e71f8d9).

consensus/src/marshal/config.rs
patrick-ogrady: Done: slimmed and de-capitalized (9e71f8d9).

glue/src/stateful/actor/syncer/mod.rs
```diff
@@ -365,81 +376,535 @@ where
...
+            // anchors marshal on the floor block, records the anchor's
+            // predecessor as processed, and prunes everything below the
+            // anchor. The block above the computed floor is then the true
+            // floor.
+            let block = marshal
+                .get_block(Identifier::Height(floor_height.next()))
+                .await
+                .expect("marshal must retain a block at its floor");
+            V::into_inner(block)
+        }
```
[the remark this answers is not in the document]
patrick-ogrady: A stored `Complete(h)` implies state sync durably anchored marshal at `h`, and marshal always retains its anchor plus the ack window below its processed height, so `max(sync_height, processed)` (or its successor in the `Start::Floor` layout) is always retained. Both layouts are pinned by regression tests (`startup_floor_jump_*` in glue, `test_standard_start_floor_above_processed_height_jumps_to_floor` in consensus). A miss implies storage was manipulated outside the process, which startup treats as fatal by design.

glue/src/stateful/mod.rs
patrick-ogrady: Follow-up: the replay path itself turned out to be branch-introduced dead weight — blocks are acknowledged only after durable finalization, so databases can never legitimately recover behind the floor. Removed it entirely: behind is now detected and treated as fatal (main's unconditional rewind would fail confusingly inside QMDB instead).

glue/src/stateful/actor/core/processing.rs
```diff
@@ -142,10 +137,18 @@ where
...
-                                .notify_finalized(self.context.as_present(), &block)
-                                .await;
+                        // Blocks at or below the processed height are already
+                        // reflected in the databases (startup converged on
+                        // them) and are acknowledged without being applied.
+                        // The genesis block is an axiom rather than an
+                        // applied block, so it is never notified.
+                        let height = block.height();
+                        if height <= self.processor.processed_height() {
+                            if !height.is_zero() {
```
patrick-ogrady: Why not pass along 0?

patrick-ogrady: No reason beyond bug-for-bug parity with main (which swallowed genesis via the finalize Duplicate arm). Passing it along is better: marshal already dispatches genesis as a first-class finalized block, and the guard forced a genesis exception into the `Application::finalized` contract. Dropped the guard — every block at or below the processed height is now reported uniformly, the doc exception is gone, and `startup_notifies_genesis` pins it.

consensus/src/marshal/core/actor.rs
```diff
@@ -1099,9 +1099,11 @@ where
         Buf: Buffer<V, PublicKey = <P::Scheme as Verifier>::PublicKey>,
         R: Resolver<Key = ResolverRequestFor<V>, Subscriber = Annotation>,
     {
+        // Applications re-provide their startup floor on every boot, so a
+        // superseded floor is routine rather than anomalous.
         let round = finalization.round();
         if round <= self.floor.processed_round() {
-            warn!(
+            debug!(
```
patrick-ogrady: keep as warn

consensus/src/marshal/core/actor.rs
```diff
@@ -1099,9 +1099,11 @@ where
         Buf: Buffer<V, PublicKey = <P::Scheme as Verifier>::PublicKey>,
         R: Resolver<Key = ResolverRequestFor<V>, Subscriber = Annotation>,
     {
+        // Applications re-provide their startup floor on every boot, so a
```
patrick-ogrady: remove comment

consensus/src/marshal/core/actor.rs
patrick-ogrady, line 1109: do we need to finish this function with the finialization we're keeping?

consensus/src/marshal/core/actor.rs
patrick-ogrady: Reverted to warn.

consensus/src/marshal/core/actor.rs
patrick-ogrady: Removed.

consensus/src/marshal/core/actor.rs
patrick-ogrady: No — when superseded we return early because the kept floor was fully installed when it was originally accepted: its anchor block and finalization were stored, the floors advanced, pruning ran, and dispatch resumes from durable processed state, so nothing remains to do with either finalization (the incoming one is at-or-below what we already processed, and its certificate was stored at original install, so there is no cache gap). In the one window where a re-provided floor is NOT superseded — after `set_complete` but before the first post-sync ack, when the pruned certs leave `processed_round` at zero — the function does run to completion with it and re-installs idempotently. Both directions are pinned: `test_standard_start_floor_below_processed_height_is_superseded` (ignored, no pruning, dispatch from durable progress) and the existing jump test.

glue/src/stateful/actor/syncer/mod.rs
```diff
@@ -35,144 +34,93 @@ const SYNC_STATE_KEY: FixedBytes<1> = fixed_bytes!("C0");
...
-            0 => Ok(Self::InProgress(FloorMarker::<C>::read(reader)?)),
-            1 => Ok(Self::Complete(Height::read(reader)?)),
+            0 => Ok(Self::InProgress {
+                height: Height::read(reader)?,
+                floor: Finalization::read_cfg(reader, cfg)?,
+            }),
+            1 => Ok(Self::Complete {
+                height: Height::read(reader)?,
+                floor: Option::<Finalization<S, C>>::read_cfg(reader, cfg)?,
+            }),
```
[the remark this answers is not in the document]
patrick-ogrady: This is allowed by policy: the entire `commonware-glue` crate is `stability_scope!(ALPHA)`, and at ALPHA storage format changes are permitted without a migration path (compatibility shims are explicitly disallowed — see CONTRIBUTING). A node upgrading across this change fails loudly at `Metadata::init` rather than misbehaving, which is the intended ALPHA upgrade contract: clear the partition or re-provision.

patrick-ogrady:
Replaced by #4239

PR #4233, opened 2026-07-11
[perf] Overlap the proposal critical path: marshal, qmdb, and pipeline optimizations

patrick-ogrady:
This PR started as proposer block forwarding. Measured on a 51-validator two-region deployment (c8a.4xlarge, us-east-1/us-west-2, ~51.6k-tx blocks), forwarding achieved **0% steady-state acceptance** — a forwarded block always lost the race to 17-of-51 shard gossip — and the shard-bundle variant shared the leader's NIC with the vote-gating fan-out, adding +34% leader egress at saturation for no latency win at any percentile. The forwarding machinery is removed in the final commit; wire and storage formats are back to `main`'s exactly (conformance passes against `main`'s hashes unchanged).

What remains is the set of changes from the same campaign that did measure, taking the deployment from **162k TPS @ 211ms views to 253k TPS @ ~200ms views**:

## Consensus / marshal

| Commit | Change | Measured |
|---|---|---|
| `9bc20a203` glue/stateful | Finalized commits pipelined off the actor loop: stage on-actor, fsync+ack on a FIFO committer, pending entry retained (Arc clone) until durable | propose queued behind an in-flight 100ms commit fsync: 101ms → 1ms; acks provably FIFO and post-fsync |
| `600f0a12f` marshal/core | Finalized-archive fsync moved off the mailbox path; durable-before-dispatch barrier kept in `try_dispatch_blocks` | `get_verified` during fsync: 101ms → 2ms |
| `283a47a79` simplex/voter | `try_propose` fires in the same event-loop step as view entry, ahead of journal sync + prune + batcher housekeeping | propose dispatched ~14ms earlier per view (leader) |
| `22060329e` marshal/coding | Reconstruction-only shard sends (rebroadcasts, non-participant deliveries) ride the low-priority lane; vote-gating assigned-shard fan-out stays high priority | votes stop queuing behind ~290KiB rebroadcasts on the shared per-peer connection |
| `3a069d216` / `caa023c41` / `6963a6040` / `7aaa2af43` marshal/coding | Shard dissemination is consensus-driven with a staged fast path: the propose path stages the coded block with the shards engine (cache + local notifications, zero sends) before enqueueing the persist, and the voter's `Relay::broadcast` triggers fan-out directly on the engine from that staged copy — no storage reload, no core-actor hop, and nothing on the network until consensus accepts the proposal. Fan-out is idempotent per (commitment, round) via an explicit last-broadcast-round marker (epoch-boundary re-proposals re-send under their new round) | dissemination trigger no longer queues behind the multi-MB verified-cache write |

## Storage (qmdb)

| Commit | Change | Measured |
|---|---|---|
| `b1fb30cdd` | Staged-merkleize serial bookkeeping halved (LWW fold via key-id map, sharded floor-candidate reads, superseded-locs during emit, sorts overlapped with hashing) | 4.39 → 2.06ms serial at a 33k-update shape; roots bit-identical across a 39-case matrix |
| `5e1804d94` | Staged reads resolving through an uncommitted ancestor's diff now stage directly (`StagedLoc::{Committed,Ancestor}`, unordered only) instead of falling back to mutations whose cost grew ~n² with ancestor overlap; floor-raise candidate prefetch overlapped with update resolution | per-op cost 0.239→0.285µs *rising* becomes 0.152→0.134µs *falling*; production leader build 61 → 38ms p50; roots bit-identical vs unmodified HEAD across depth/size/seed/thread matrix |

## Runtime

| Commit | Change |
|---|---|
| `33f18b5f3` | `BufferPoolConfig::with_class_max` — per-class buffer pool caps |
| `7e5935fd2` | Per-class thread-cache overrides (a raised class cap enlarges TLS caches; measured total stranding without this) |

## Deployer

| Commit | Change |
|---|---|
| `1f9a96a84` / `dd513ba30` | BBR + TCP buffer sysctls for cross-region flows (16MiB maxima) |
| `002b83329` | `net.core.default_qdisc=fq` only applies to interfaces attached after it is set, so the boot-time interface silently kept unpaced `mq+fq_codel`. Install fq explicitly per tx queue via a persistent systemd oneshot (measured 14 → 1.7 retransmits/node/s). Deliberately no `maxrate`: a per-flow cap makes votes drain behind in-flight shards on the multiplexed per-peer connection (measured view-latency regression). Also raises the `tcp_rmem` default |

## Verification

- `commonware-consensus`: marshal suite 333 green (377 with `arbitrary`/conformance), voter suite 112 green; conformance hashes identical to `main`.
- `commonware-storage`: qmdb suite 1949 green; merkle roots verified bit-identical against a pristine-`main` binary across shapes, seeds, and thread counts for both qmdb commits.
- Deployer/glue/runtime suites green; workspace clippy/fmt/docs clean.
- Branch is merged up to current `main`.

https://claude.ai/code/session_01Punr4hDySyvsUZHvNWyvqi


storage/src/journal/contiguous/fixed.rs
```diff
@@ -1176,7 +1176,7 @@ impl<E: Context, A: CodecFixedShared> Reader<'_, E, A> {
         if positions.is_empty() {
             return Ok(Vec::new());
         }
-        assert!(
+        debug_assert!(
```
patrick-ogrady, line 1179: keep as assert

consensus/src/marshal/coding/shards/engine.rs
```diff
@@ -110,6 +114,29 @@
...
 //!
+//! # Send Priorities
+//!
+//! The p2p layer multiplexes every channel over a single connection per peer
+//! with two send lanes (the `priority` flag on [`commonware_p2p::Sender`]);
+//! consensus votes and certificates ride the high lane. The engine splits its
+//! sends by whether they gate a recipient's notarize vote, so bulk shard
+//! payloads never queue ahead of vote-sized consensus messages:
+//!
+//! - High priority (vote-gating): the leader's per-participant assigned-shard
```
patrick-ogrady, line 125: I think we can revert this

glue/src/stateful/actor/processor/mod.rs
```diff
@@ -712,20 +768,29 @@ where
...
+        let batch = match self.pending.get(&digest) {
+            Some(entry) => entry.merkleized.clone(),
             None => {
-                let batches = self.databases.new_batches().await;
+                // Fork the parent's retained overlay when its commit is still
+                // in flight; otherwise replay on committed state (which waits
+                // out any in-flight commit via the database locks). A replay
+                // from the wrong base is caught by the commitment check below.
+                let parent = block.parent();
+                let batches = match self.fork_batches(&parent).await {
```
patrick-ogrady, line 790: This is just a workaround for upstream QMDB changes...we should probably just wait for that...

PR #4234, opened 2026-07-11
[ci] Check dependency cooldown from merge base

clabby:
## Overview

Fix a bug where any pull request lockfile change failed the cooldown job. Check dependency cooldown against the pull request merge base so existing dependencies remain blessed while newly introduced dependencies must satisfy the cooldown.

### DIFF
--- a/.github/scripts/check_dependency_cooldown.sh
+++ b/.github/scripts/check_dependency_cooldown.sh
@@ -1,26 +1,8 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-check_cooldown() {
-  cargo cooldown --workspace --all-features check
-}
-
-check_lockfile() {
-  git diff --exit-code Cargo.lock
-}
-
-if [[ "${GITHUB_EVENT_NAME:-}" != "pull_request" ]]; then
-  check_cooldown
-  check_lockfile
-  exit 0
-fi
-
-base_ref="${COOLDOWN_BASE_REF:-}"
-head_sha="${COOLDOWN_HEAD_SHA:-}"
-if [[ -z "${base_ref}" || -z "${head_sha}" ]]; then
-  echo "COOLDOWN_BASE_REF and COOLDOWN_HEAD_SHA are required for pull_request cooldown checks" >&2
-  exit 1
-fi
+base_ref="${COOLDOWN_BASE_REF:-main}"
+head_sha="${COOLDOWN_HEAD_SHA:-HEAD}"
 
 git fetch --no-tags origin "refs/heads/${base_ref}:refs/remotes/origin/${base_ref}"
 base_rev="refs/remotes/origin/${base_ref}"
@@ -33,20 +15,42 @@ fi
 
 tmpdir="$(mktemp -d)"
 cp Cargo.lock "${tmpdir}/pr.Cargo.lock"
+cp cooldown.toml "${tmpdir}/cooldown.toml"
 cleanup() {
   status=$?
   cp "${tmpdir}/pr.Cargo.lock" Cargo.lock
+  cp "${tmpdir}/cooldown.toml" cooldown.toml
   rm -rf "${tmpdir}"
   exit "${status}"
 }
 trap cleanup EXIT
 
-git show "${base_rev}:Cargo.lock" > Cargo.lock
+git show "${merge_base}:Cargo.lock" > "${tmpdir}/base.Cargo.lock"
+# Versions already present at the merge base are allowed regardless of publish age.
+awk '
+  function exempt() {
+    if (source ~ /^registry\+/) {
+      print ""
+      print "[[allow.exact]]"
+      print "crate = \"" name "\""
+      print "version = \"" version "\""
+    }
+  }
 
-check_cooldown
+  /^\[\[package\]\]$/ {
+    exempt()
+    name = version = source = ""
+    next
+  }
+  /^name = "/ { name = substr($0, 9, length($0) - 9) }
+  /^version = "/ { version = substr($0, 12, length($0) - 12) }
+  /^source = "/ { source = substr($0, 11, length($0) - 11) }
+  END { exempt() }
+' "${tmpdir}/base.Cargo.lock" >> cooldown.toml
+
+cargo cooldown metadata --all-features --format-version 1 --no-deps > /dev/null
 
 if ! cmp --silent Cargo.lock "${tmpdir}/pr.Cargo.lock"; then
-  echo "Cargo.lock differs after applying the cooldown baseline." >&2
-  echo "Regenerate Cargo.lock after the upgraded dependencies satisfy cooldown." >&2
+  echo "Cargo.lock contains dependencies that do not pass cooldown." >&2
   exit 1
 fi
--- a/.github/scripts/check_dependency_cooldown.sh
+++ b/.github/scripts/check_dependency_cooldown.sh
@@ -2,13 +2,13 @@
 set -euo pipefail
 
 base_ref="${COOLDOWN_BASE_REF:-main}"
-head_sha="${COOLDOWN_HEAD_SHA:-HEAD}"
 
-git fetch --no-tags origin "refs/heads/${base_ref}:refs/remotes/origin/${base_ref}"
+git fetch --no-tags origin "+refs/heads/${base_ref}:refs/remotes/origin/${base_ref}" ||
+  echo "Failed to fetch origin/${base_ref}; using the local ref." >&2
 base_rev="refs/remotes/origin/${base_ref}"
-merge_base="$(git merge-base "${base_rev}" "${head_sha}")"
+merge_base="$(git merge-base "${base_rev}" HEAD)"
 
-if git diff --quiet "${merge_base}" "${head_sha}" -- Cargo.lock; then
+if git diff --quiet "${merge_base}" -- Cargo.lock; then
   echo "Cargo.lock did not change; skipping cargo-cooldown."
   exit 0
 fi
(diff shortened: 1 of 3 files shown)


.github/scripts/check_dependency_cooldown.sh
```diff
@@ -1,26 +1,8 @@
...
-head_sha="${COOLDOWN_HEAD_SHA:-}"
-if [[ -z "${base_ref}" || -z "${head_sha}" ]]; then
-  echo "COOLDOWN_BASE_REF and COOLDOWN_HEAD_SHA are required for pull_request cooldown checks" >&2
-  exit 1
-fi
+base_ref="${COOLDOWN_BASE_REF:-main}"
+head_sha="${COOLDOWN_HEAD_SHA:-HEAD}"
 
 git fetch --no-tags origin "refs/heads/${base_ref}:refs/remotes/origin/${base_ref}"
 base_rev="refs/remotes/origin/${base_ref}"
```
patrick-ogrady, line 8: Three issues in this block, all affecting local `just cooldown` runs:

1. The refspec has no `+` prefix, so a rebased/force-pushed base branch makes the fetch fail with a non-fast-forward rejection and `set -e` aborts before any checking. CI is unaffected (fresh clones), but a local clone stays broken until the ref is manually updated.
2. The unconditional fetch makes the script die offline even when a usable local `origin/<base>` ref exists.
3. `head_sha` defaults to `HEAD`, but the skip gate below diffs two *commits* while the actual check reads the *working tree*. Adding a dependency and running `just cooldown` before committing prints "Cargo.lock did not change" and exits 0 — a false pass on the primary local use case. Diffing the merge base against the working tree closes that and lets `head_sha` go away entirely (in CI the working tree is the checkout ref, so behavior there is unchanged, and the checkout `ref:` becomes the single source of truth).

```suggestion
base_ref="${COOLDOWN_BASE_REF:-main}"

git fetch --no-tags origin "+refs/heads/${base_ref}:refs/remotes/origin/${base_ref}" ||
  echo "Failed to fetch origin/${base_ref}; using the local ref." >&2
base_rev="refs/remotes/origin/${base_ref}"
```

and on the (unchanged, so not suggestable) lines just below:

```bash
merge_base="$(git merge-base "${base_rev}" HEAD)"

if git diff --quiet "${merge_base}" -- Cargo.lock; then
```

With `head_sha` gone, `COOLDOWN_HEAD_SHA` can also be dropped from the Cooldown job env in fast.yml.

.github/scripts/check_dependency_cooldown.sh
```diff
@@ -33,20 +15,42 @@ fi
...
 trap cleanup EXIT
 
-git show "${base_rev}:Cargo.lock" > Cargo.lock
-
-check_cooldown
+git show "${merge_base}:Cargo.lock" > "${tmpdir}/base.Cargo.lock"
+# Versions already present at the merge base are allowed regardless of publish age.
+awk '
+  function exempt() {
+    if (source ~ /^registry\+/) {
```
patrick-ogrady: Two things about the exemption filter:

1. Alternative registries are recorded as `source = "sparse+https://..."` in Cargo.lock, so their base-locked versions would get no exemption and any PR touching Cargo.lock would fail on a dependency it didn't introduce.
2. Worth documenting the residual limitation: cargo-cooldown's `[[allow.exact]]` matches crate + version only (source is ignored), so a same-name+version package from a *different* registry inherits the exemption. Not fixable on this side of the tool.

```suggestion
# Versions already present at the merge base are allowed regardless of publish age.
# The generated exemptions match crate and version only (cargo-cooldown ignores source).
awk '
  function exempt() {
    if (source ~ /^(registry|sparse)\+/) {
```

.github/scripts/check_dependency_cooldown.sh
```diff
@@ -33,20 +15,42 @@ fi
...
+      print ""
+      print "[[allow.exact]]"
+      print "crate = \"" name "\""
+      print "version = \"" version "\""
+    }
+  }
+
+  /^\[\[package\]\]$/ {
```
patrick-ogrady: State only resets on `[[package]]` headers, so any other lockfile section bleeds state. Concretely: once the workspace ever uses `[patch]`, Cargo appends `[[patch.unused]]` after the packages. Its `name`/`version` lines overwrite the accumulator while `source` retains the previous package's registry value, and the `END` block then emits a bogus exemption for the patched crate *and* drops the final real package's exemption (verified on a synthetic lockfile). Resetting on any `[[` header fixes both, since patch entries carry no `source` and `exempt()` skips them:

```suggestion
  /^\[\[/ {
```

.github/scripts/check_dependency_cooldown.sh
```diff
@@ -33,20 +15,42 @@ fi
...
+  /^source = "/ { source = substr($0, 11, length($0) - 11) }
+  END { exempt() }
+' "${tmpdir}/base.Cargo.lock" >> cooldown.toml
+
+cargo cooldown metadata --all-features --format-version 1 --no-deps > /dev/null
 
 if ! cmp --silent Cargo.lock "${tmpdir}/pr.Cargo.lock"; then
-  echo "Cargo.lock differs after applying the cooldown baseline." >&2
-  echo "Regenerate Cargo.lock after the upgraded dependencies satisfy cooldown." >&2
+  echo "Cargo.lock contains dependencies that do not pass cooldown." >&2
```
patrick-ogrady: The guard's `cargo metadata` run also syncs a Cargo.lock that is merely out of date with the manifests and publishes it back, so `cmp` fails for stale lockfiles too and this message misdiagnoses them as cooldown violations (the old message's "regenerate" hint covered exactly that case). Printing the rewritten entries also tells the author which dependencies to wait on:

```suggestion
  echo "cargo-cooldown rewrote these Cargo.lock entries:" >&2
  diff "${tmpdir}/pr.Cargo.lock" Cargo.lock >&2 || true
  echo "Wait for the dependencies above to satisfy cooldown, or regenerate Cargo.lock if it is out of sync with the manifests." >&2
```

.github/workflows/fast.yml
```diff
@@ -256,12 +256,14 @@ jobs:
       run: just udeps
 
   Cooldown:
+    if: github.event_name == 'pull_request'
```
patrick-ogrady: With the job gated to `pull_request` and the checkout pinned to `head.sha`, the merge result is never validated under main's *current* policy: if main tightens cooldown.toml (or fixes this script) after a PR branches, the PR passes against its stale copy and lands unchecked — and no merge_group/push run re-checks it. Running on `merge_group` too closes that, since the merge-queue checkout carries main's merged policy. The `pull_request.*` expressions evaluate to empty on merge_group, so checkout falls back to the merge-group ref and `${COOLDOWN_BASE_REF:-main}` handles the empty env var.

```suggestion
    if: github.event_name == 'pull_request' || github.event_name == 'merge_group'
```

cooldown.toml
```diff
@@ -1,4 +1,4 @@
 cooldown_minutes = 10080
```
patrick-ogrady, line 4: `lockfile_baseline = "ignore"` changes behavior for every cargo-cooldown invocation, not just this script: cargo-cooldown is a general wrapper that reads this committed config, and the merge-base exemptions only exist in the script's temporary append. Someone running `cargo cooldown check`/`build` directly gets every locked package inside the window cooled — a rewritten (downgraded) Cargo.lock or a hard failure. Fine given the design, but worth a warning for anyone who finds this file first:

```suggestion
# Exemptions for versions already locked at the merge base are appended by
# .github/scripts/check_dependency_cooldown.sh (run via `just cooldown`). Invoking
# `cargo cooldown` directly applies the cooldown window to every locked package.
cooldown_minutes = 10080
```

patrick-ogrady (commented):
Reviewed the new gate and validated fixes for the issues below locally (awk verified against a synthetic lockfile with a trailing `[[patch.unused]]` section and against the real merge-base lockfile — 548 sourced packages produce exactly 548 exemptions; the script was exercised end-to-end for the skip path, the offline-fetch fallback, and trap restoration on failure). Each inline comment carries the fix as a suggestion.

Residuals I'd leave as-is, noting them for the record:
- The EXIT-trap restore doesn't cover SIGKILL/OOM, which would leave the appended exemptions in the working tree's cooldown.toml. Rare enough to accept (the trap does fire on SIGINT/SIGTERM — checked empirically).
- `cooldown_minutes` and root `enforcement` are deprecated as of cargo-cooldown 0.3.1 with removal slated for 0.4.x. Worth migrating to the `[cooldown]`/`[registry]` format whenever `CARGO_COOLDOWN_VERSION` is next bumped.
- The old invocation's `--workspace` is dropped. Harmless while the workspace has no `default-members`, but the cooldown scan scope would silently shrink if that ever changes.

.github/scripts/check_dependency_cooldown.sh
clabby: Fixed in 90e7014f: force-fetch with local fallback, compare against the working tree, and remove COOLDOWN_HEAD_SHA.

.github/scripts/check_dependency_cooldown.sh
clabby: Fixed in 90e7014f by resetting state on every `[[` header.

.github/scripts/check_dependency_cooldown.sh
clabby: Fixed in 90e7014f: sparse sources are included and the source-matching limitation is documented.

.github/scripts/check_dependency_cooldown.sh
clabby: Fixed in 90e7014f: failures now print the lockfile diff and cover stale manifests in the guidance.

.github/workflows/fast.yml
clabby: Fixed in 90e7014f: the cooldown job now also runs for merge groups.

cooldown.toml
clabby: Documented in 90e7014f.

patrick-ogrady (approved):

.github/workflows/fast.yml
```diff
@@ -15,7 +15,7 @@ concurrency:
 
 env:
   CARGO_TERM_COLOR: always
-  CARGO_COOLDOWN_VERSION: 0.3.0
+  CARGO_COOLDOWN_VERSION: 0.3.4
```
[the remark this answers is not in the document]
clabby: https://crates.io/crates/cargo-cooldown `v0.3.4` exists?

PR #4237, opened 2026-07-13
[storage] Make pruning crash-safe with buffered operations

roberto-bayardo:
## Summary

- fixes a pre-existing crash bug that left qmdb permanently unopenable: `apply_batch` advances the inactivity floor in memory, and pruning to it without a commit durably removed operations that the last recoverable commit still needed — every reopen then failed with `ItemPruned`
- an effective contiguous-journal prune flushes all dirty blobs (and, in the variable journal, commits its offsets journal) before removing anything, so a crash mid-prune never leaves survivors that fail to justify the boundary; contiguous-journal prunes that remove no blobs perform no storage I/O
- `Mutable::prune` makes no durability promise for buffered appends: consumers that advance an in-memory floor ahead of durability protect their own justification — the authenticated journal commits the inner journal alongside its Merkle sync (covering `any`, `immutable`, and `keyless`), and the standalone store commits before pruning
- regression tests: the qmdb floor crash across all sixteen db variants plus the standalone store; journal-level crash-point regressions for the offsets gap (found by Cursor Bugbot) and for an unsynced tail surviving a prune-then-crash; a Current mid-prune cancellation regression

## Design

Two concerns, two layers:

**Journal-internal crash consistency (implementation, not contract).** Removals are durable and may be interrupted, so an effective prune first makes every dirty blob durable: recovery truncates at the first torn item, so an unsynced survivor below the boundary could discard every synced blob behind it. The variable journal also commits its offsets journal before removing data blobs — recovery can rebuild offsets that end behind the surviving data's end by replaying data, but offsets ending behind the surviving data's *start* are unrecoverable because the data needed to rebuild them is being removed. Data flushes before offsets, matching the ordering every other durability path maintains. The added pre-removal durability barrier performs no I/O for a clean journal — the common case, since consumers usually prune right after committing — and no recovery metadata advances.

**Durability of the prune's justification (the consumer's job).** `Mutable::prune` deliberately promises nothing about buffered appends surviving a prune plus crash, keeping its contract aligned with every other journal type and leaving future implementations free to narrow the internal flush (e.g. under #4235's durable prune record). Consumers whose prune target may be justified only by a buffered operation commit first: the authenticated journal does this once for all its users (`try_join`ed with its existing Merkle sync), and the standalone `store::Db` commits its raw journal before pruning. `Current::prune` needs the barrier earlier still — it durably persists bitmap and grafted-tree pruning metadata derived from the sync boundary before reaching the log prune, so it commits the log before mutating any of that metadata; a crash between the metadata sync and the log prune otherwise recovers the older durable floor alongside newer pruning metadata and panics while initializing the shared bitmap. These QMDB prune paths are crash-safe immediately after an uncommitted batch — the obligation lives in the database layer, not on its callers.

The offsets-gap state was already reachable on `main` via `commit` → `prune` → crash between data and offsets pruning; this PR's flush made it reachable from any prune of a dirty tail, and the offsets commit closes it.

### DIFF
--- a/storage/src/journal/authenticated.rs
+++ b/storage/src/journal/authenticated.rs
@@ -558,7 +558,12 @@ where
         // Sync the Merkle structure before pruning the journal, otherwise its last element could
         // end up behind the journal's first element after a crash, and there would be no way to
         // replay the items between the structure's last element and the journal's first element.
-        self.merkle.sync().await?;
+        // Commit the journal alongside: the prune target may be justified by a buffered append
+        // (e.g. a commit operation), and pruning does not guarantee buffered appends are durable.
+        try_join!(
+            self.journal.commit().map_err(Error::Journal),
+            self.merkle.sync().map_err(Error::Merkle)
+        )?;
 
         let journal_pruned = self.journal.prune(*prune_loc).await?;
         let bounds = self.journal.bounds();
--- a/storage/src/journal/contiguous/variable.rs
+++ b/storage/src/journal/contiguous/variable.rs
@@ -1590,7 +1590,8 @@ impl<E: Context, V: CodecShared> Journal<E, V> {
         // journal whose surviving items no longer justify its boundary. Dirty blobs below the
         // prune point are flushed too: removal is oldest-first and may be interrupted, and
         // recovery truncates at the first torn item, so an unsynced survivor below the
-        // boundary could discard every synced blob behind it.
+        // boundary could discard every synced blob behind it. Runs even when no blob will
+        // be removed: every prune call makes retained items recoverable.
         self.flush_dirty_data().await?;
         self.dirty_from_blob = None;
 
@@ -1603,6 +1604,8 @@ impl<E: Context, V: CodecShared> Journal<E, V> {
         // Offsets entries for retained items must survive the same crash: recovery treats an
         // offsets journal that ends behind the surviving data as corruption. Data is flushed
         // first (above), matching the ordering every other durability path maintains.
+        // Deferred until the prune is known effective: offsets only need to be durable
+        // before a removal.
         self.offsets.commit().await?;
 
         self.blobs.prune(min_blob).await?;
--- a/storage/src/journal/contiguous/variable.rs
+++ b/storage/src/journal/contiguous/variable.rs
@@ -1560,9 +1560,6 @@ impl<E: Context, V: CodecShared> Journal<E, V> {
     ///
     /// Returns `true` if any data was pruned, `false` otherwise.
     ///
-    /// Prune first makes all retained items recoverable, as [Self::commit] does, so a crash
-    /// never recovers a journal whose surviving items fail to justify its pruning boundary.
-    ///
     /// # Errors
     ///
     /// Returns an error if the underlying storage operation fails.
@@ -1575,29 +1572,25 @@ impl<E: Context, V: CodecShared> Journal<E, V> {
         let tail_blob = position_to_blob(self.bounds.end, items_per_blob);
         let min_blob = target_blob.min(tail_blob);
 
-        // Make all dirty blobs durable before removing any: the prune target is often
-        // justified by an appended-but-unflushed item (e.g. a consumer's commit record), and
-        // removals are durable, so pruning without this barrier could leave a recovered
-        // journal whose surviving items no longer justify its boundary. Dirty blobs below the
-        // prune point are flushed too: removal is oldest-first and may be interrupted, and
-        // recovery truncates at the first torn item, so an unsynced survivor below the
-        // boundary could discard every synced blob behind it. Runs even when no blob will
-        // be removed: every prune call makes retained items recoverable.
-        self.flush_dirty_data().await?;
-        self.dirty_from_blob = None;
-
         if min_blob <= self.blobs.oldest_blob_index() {
             return Ok(false);
         }
 
         let new_boundary = blob_first_position(min_blob, items_per_blob)?;
 
-        // Offsets entries for retained items must survive the same crash: recovery rebuilds
-        // offsets that end behind the surviving data's end by replaying data, but offsets that
-        // end behind its start are unrecoverable because the data needed to rebuild the missing
-        // entries is about to be removed. Data is flushed first (above), matching the ordering
-        // every other durability path maintains. Deferred until the prune is known effective:
-        // offsets only need to be durable before a removal.
+        // Make all dirty blobs durable before removing any: the prune target may be
+        // justified by an appended-but-unflushed item (e.g. a consumer's commit record), and
+        // removals are durable, so pruning without this barrier could leave a recovered
+        // journal whose surviving items no longer justify its boundary. Dirty blobs below the
+        // prune point are flushed too: removal may be interrupted, and recovery truncates at
+        // the first torn item, so an unsynced survivor below the boundary could discard
+        // every synced blob behind it. Offsets entries for retained items must survive the
+        // same crash: recovery rebuilds offsets that end behind the surviving data's end by
+        // replaying data, but offsets that end behind its start are unrecoverable because
+        // the data needed to rebuild the missing entries is about to be removed. Data is
+        // flushed first, matching the ordering every other durability path maintains.
+        self.flush_dirty_data().await?;
+        self.dirty_from_blob = None;
         self.offsets.commit().await?;
 
         self.blobs.prune(min_blob).await?;
--- a/storage/src/journal/contiguous/variable.rs
+++ b/storage/src/journal/contiguous/variable.rs
@@ -420,6 +420,11 @@ pub struct Journal<E: Context, V: Codec> {
     /// Earliest data blob modified since the last `commit()` or `sync()`.
     dirty_from_blob: Option<u64>,
 
+    /// Test-only: park [Self::prune] after the data-blob removal, before the offsets prune,
+    /// so tests can drop the pending future at that exact point.
+    #[cfg(test)]
+    halt_before_offsets_prune: bool,
+
     /// The number of items per blob.
     ///
     /// # Invariant
@@ -1130,6 +1135,8 @@ impl<E: Context, V: CodecShared> Journal<E, V> {
             offsets,
             bounds,
             dirty_from_blob: None,
+            #[cfg(test)]
+            halt_before_offsets_prune: false,
             items_per_blob: cfg.items_per_section,
             compression: cfg.compression,
             codec_config: cfg.codec_config,
@@ -1193,6 +1200,8 @@ impl<E: Context, V: CodecShared> Journal<E, V> {
             offsets,
             bounds: size..size,
             dirty_from_blob: None,
+            #[cfg(test)]
+            halt_before_offsets_prune: false,
             items_per_blob: cfg.items_per_section,
             compression: cfg.compression,
             codec_config: cfg.codec_config,
@@ -1558,24 +1567,6 @@ impl<E: Context, V: CodecShared> Journal<E, V> {
     ///
     /// Returns an error if the underlying storage operation fails.
     pub async fn prune(&mut self, min_position: u64) -> Result<bool, Error> {
-        let Some(new_boundary) = self.prune_data(min_position).await? else {
-            return Ok(false);
-        };
-
-        // Prune data before offsets so a crash leaves offsets behind, which init repairs by
-        // pruning offsets to match.
-        self.offsets.prune(new_boundary).await?;
-        self.metrics.update(
-            self.bounds.end,
-            self.bounds.start,
-            self.items_per_blob.get(),
-        );
-
-        Ok(true)
-    }
-
-    /// Flush dirty data and remove blobs before updating the offsets journal.
-    async fn prune_data(&mut self, min_position: u64) -> Result<Option<u64>, Error> {
         let items_per_blob = self.items_per_blob.get();
 
         // Calculate the blob that would contain min_position, capped to the tail (which is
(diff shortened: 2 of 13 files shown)


storage/src/journal/contiguous/variable.rs
```diff
@@ -1569,10 +1582,6 @@ impl<E: Context, V: CodecShared> Journal<E, V> {
         self.blobs.prune(min_blob).await?;
         self.bounds.start = new_boundary;
         self.offsets.prune(new_boundary).await?;
-
-        if let Some(dirty_from) = self.dirty_from_blob {
-            self.dirty_from_blob = Some(dirty_from.max(min_blob));
-        }
```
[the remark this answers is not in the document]
roberto-bayardo: Confirmed — real, and reachable exactly as described: the barrier made data durable while the offsets journal's dirty tail stayed buffered, so a crash between the data-blob removals and `offsets.prune` left durable offsets ending behind the surviving data, which `align` rejects as corruption. (The same state was already reachable on `main` via `commit` → `prune` → crash in that gap; the barrier widened it to any prune of a dirty tail.)

Fixed: the barrier now also commits the offsets journal before any data blob is removed — data first, then offsets, preserving the "durable offsets never run ahead of durable data" ordering every other durability path maintains. Added a crash-point regression (`test_variable_recovery_prune_crash_offsets_end_behind`) that runs prune through data removal, stops before `offsets.prune`, and reopens: it failed with `Corruption("offsets journal size 7 is behind data oldest position 10")` before the fix and now recovers to the pruned bounds.

storage/src/journal/contiguous/variable.rs
```diff
@@ -1558,6 +1561,16 @@ impl<E: Context, V: CodecShared> Journal<E, V> {
...
         let min_blob = target_blob.min(tail_blob);
 
+        // Make all dirty blobs durable before removing any: the prune target is often
+        // justified by an appended-but-unflushed item (e.g. a consumer's commit record), and
+        // removals are durable, so pruning without this barrier could leave a recovered
+        // journal whose surviving items no longer justify its boundary. Dirty blobs below the
+        // prune point are flushed too: removal is oldest-first and may be interrupted, and
+        // recovery truncates at the first torn item, so an unsynced survivor below the
+        // boundary could discard every synced blob behind it.
+        self.flush_dirty_data().await?;
```
patrick-ogrady, line 1595: Should we just call `sync` to avoid this subtlty?

https://github.com/commonwarexyz/monorepo/blob/a9e00b3a8f5b2a16fdcec07266b4d6982a2e1767/storage/src/journal/contiguous/variable.rs#L1615

storage/src/journal/contiguous/variable.rs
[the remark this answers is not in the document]
patrick-ogrady: This looks valid (see `sync`)

storage/src/journal/contiguous/variable.rs
```diff
@@ -1558,6 +1561,16 @@ impl<E: Context, V: CodecShared> Journal<E, V> {
...
+        self.dirty_from_blob = None;
```
patrick-ogrady, line 1596: do we need to sync offsets?

storage/src/journal/contiguous/variable.rs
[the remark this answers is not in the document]
roberto-bayardo: yep fixed

storage/src/journal/contiguous/variable.rs
roberto-bayardo: yes, this is done (I neglected to push the commit :/)

storage/src/journal/contiguous/variable.rs
roberto-bayardo: Considered that — the delta turns out to be cost rather than safety, so I kept the commit-level barrier:

- Crash-safety is identical: `sync()` is the same `flush_dirty_data` → offsets flush, in the same data-before-offsets order. The barrier now includes `offsets.commit()` (see the Bugbot thread), so both paths make the same bytes durable before any removal.
- When the journal is clean (the common case, pruning right after a commit/sync), both are free — `Metadata::sync` no-ops when nothing is staged.
- When pruning a dirty tail, `sync()` pays one extra metadata write + fsync (`offsets.sync` runs `checkpoint.persist`, durably advancing the watermark to `size`), and it advances recovery metadata the caller didn't ask for — a consumer that deliberately chose `commit()`-level durability gets a full watermark advance as a side effect of pruning. It would also tick `sync_calls`, so prunes surface as syncs in metrics.
- The comment-heavy part of the barrier (why we flush at all, and why doomed below-boundary blobs are included) applies unchanged under `sync()`, so the simplification is really just the one `offsets.commit()` line.

Happy to switch if you weigh the one-liner higher — it's strictly safe — but commit-level keeps prune from imposing metadata barriers beyond what recovery needs.

storage/src/journal/contiguous/variable.rs
```diff
@@ -1558,28 +1579,30 @@ impl<E: Context, V: CodecShared> Journal<E, V> {
...
         }
 
         let new_boundary = blob_first_position(min_blob, items_per_blob)?;
 
-        // Prune data before offsets so a crash leaves offsets behind, which init repairs by
-        // pruning offsets to match.
+        // Offsets entries for retained items must survive the same crash: recovery treats an
+        // offsets journal that ends behind the surviving data as corruption. Data is flushed
+        // first (above), matching the ordering every other durability path maintains.
+        self.offsets.commit().await?;
```
patrick-ogrady, line 1597: It feels odd to me to slip this durable op in `prune_data` given we prune offsets directly later?

I think it is "correct" tho.

storage/src/qmdb/immutable/mod.rs
```diff
@@ -468,11 +468,9 @@ where
     /// Prune operations prior to `prune_loc`. This does not affect the db's root, but it will
     /// affect retrieval of any keys that were set prior to `prune_loc`.
     ///
-    /// Pruning is irreversible. Callers must ensure any floor-raising batch has been durably
-    /// committed (via [`Immutable::commit`] or [`Immutable::sync`]) before pruning. The
-    /// inactivity floor used to gate pruning is updated by [`Immutable::apply_batch`] before
-    /// the batch is durable. If the batch is lost on crash, recovery replays from the prior
-    /// durable floor, which may reference data that has already been pruned.
+    /// Pruning is irreversible. Retained operations -- including the floor-raising commit that
```
patrick-ogrady: remove the `--` slop?

storage/src/journal/contiguous/variable.rs
roberto-bayardo: The later `offsets.prune` flushes too late since it runs after the data-blob removals are already durable, and a crash in that gap leaves offsets durably ending behind the surviving data, which recovery reports as corruption (that's the Bugbot finding above). Removing it requires recovery changes that amount to #4235's durable prune record, so this line is removed there (but it's a much bigger change I'd like to stage after these).

storage/src/qmdb/immutable/mod.rs
roberto-bayardo: done

storage/src/journal/contiguous/variable.rs
```diff
@@ -2198,6 +2221,14 @@ impl<E: Context, V: CodecShared> authenticated::Inner<E> for Journal<E, V> {
 
 #[cfg(test)]
 impl<E: Context, V: CodecShared> Journal<E, V> {
+    /// Test helper: Run prune through data removal, then simulate a crash before offsets pruning.
+    pub(crate) async fn test_prune_before_offsets(
```
patrick-ogrady: I think you can inline this? its only used once?

roberto-bayardo: yup inlined

storage/src/journal/contiguous/variable.rs
```diff
@@ -1558,28 +1579,30 @@ impl<E: Context, V: CodecShared> Journal<E, V> {
...
         let min_blob = target_blob.min(tail_blob);
 
+        // Make all dirty blobs durable before removing any: the prune target is often
+        // justified by an appended-but-unflushed item (e.g. a consumer's commit record), and
+        // removals are durable, so pruning without this barrier could leave a recovered
+        // journal whose surviving items no longer justify its boundary. Dirty blobs below the
+        // prune point are flushed too: removal is oldest-first and may be interrupted, and
+        // recovery truncates at the first torn item, so an unsynced survivor below the
+        // boundary could discard every synced blob behind it.
+        self.flush_dirty_data().await?;
```
patrick-ogrady: should we move this after:

```
        if min_blob <= self.blobs.oldest_blob_index() {
            return Ok(None);
        }
```

We can then do all of:
```
self.flush_dirty_data().await?;
 self.dirty_from_blob = None;
self.offsets.commit().await?;
```

Together?

storage/src/journal/contiguous/variable.rs
```diff
@@ -1546,10 +1546,31 @@ impl<E: Context, V: CodecShared> Journal<E, V> {
...
     /// Returns `true` if any data was pruned, `false` otherwise.
     ///
+    /// Prune first makes all retained items recoverable, as [Self::commit] does, so a crash
+    /// never recovers a journal whose surviving items fail to justify its pruning boundary.
+    ///
     /// # Errors
     ///
     /// Returns an error if the underlying storage operation fails.
     pub async fn prune(&mut self, min_position: u64) -> Result<bool, Error> {
+        let Some(new_boundary) = self.prune_data(min_position).await? else {
```
patrick-ogrady: I think this would be a lot cleaner to reason about without this inner `prune_data` (just for the sake of writing a test).

Do we do this elsewhere to test recovery?

storage/src/journal/contiguous/variable.rs
roberto-bayardo: The split is actually principled: the data flush serves the caller-facing contract (so it must run on every call), while offsets.commit serves removal ordering (only meaningful when something will be removed).  We could group them all together above the early return at the cost of one extra offsets flush. Added some comments to make this more clear.

storage/src/journal/contiguous/variable.rs
patrick-ogrady: I agree the current flow is right as-written, I just found it a little deceptive.

storage/src/journal/contiguous/variable.rs
roberto-bayardo: An earlier version of this test reconstructed the crash state by hand (flush via `test_sync_data`, remove blobs via `test_prune_data`) — but that replicates prune's internals in test code: if prune's sequence ever changes, the test silently constructs a state production can no longer produce and stops guarding anything. Routing through `prune_data` means the regression runs the real production sequence up to the crash point, so it fails if the production ordering regresses — we verified that property by mutation on the equivalent test in #4235 (reverting the production coupling makes it fail; a hand-rolled reconstruction kept passing).

#4235 leans further into the same pattern (production-path splits plus a test-only park point for dropping a genuinely pending prune future), so this shape is where the recovery/cancel-safety testing is converging rather than a one-off.

storage/src/journal/contiguous/variable.rs
patrick-ogrady: > The split is actually principled: the data flush serves the caller-facing contract (so it must run on every call)

but this won't trigger a prune? is the claim now that we always make durable even if not pruning? in which case, I would argue we commit offsets too.

storage/src/journal/contiguous/variable.rs
roberto-bayardo: ^^ fable response.  I think I can remove the split one sec.

roberto-bayardo: Removed the split by introducing a test config guarded hook, which has some precedent.  Is that better?

storage/src/journal/contiguous/variable.rs
roberto-bayardo: Correct. But you only need to flush the dirty data to guarantee durability.

roberto-bayardo: Rethought this after your comments and went further: `Mutable::prune` no longer promises any durability (matching every other journal type), so a no-op fixed/variable contiguous prune now returns before any fsync, and the effective-prune barrier is exactly the grouped block you suggested — data flush, dirty clear, offsets commit, together after the effectiveness check (8e73bf089). The flush survives as an implementation requirement of crash-consistent removal (your unsynced-tail test pins it), not as a contract. Consumers that advance an in-memory floor ahead of durability now protect their own justification: the authenticated journal commits the inner journal `try_join`ed with its existing Merkle sync (one barrier for `any`/`immutable`/`keyless`, sound for any future `Mutable` impl), and the standalone `store::Db` commits before pruning.

storage/src/journal/authenticated.rs
```diff
@@ -558,7 +558,12 @@ where
         // Sync the Merkle structure before pruning the journal, otherwise its last element could
         // end up behind the journal's first element after a crash, and there would be no way to
         // replay the items between the structure's last element and the journal's first element.
-        self.merkle.sync().await?;
```
patrick-ogrady, line 561: I suppose this is a bit of a red herring compared to what we did elsewhere on prune?

roberto-bayardo: Since Mutable::prune deliberately makes no durability promise, the authenticated layer cannot rely on that implementation detail. This commit ensures the operation establishing the new floor is durable before pruning history required by the previous durable floor, regardless of the underlying Mutable implementation.

storage/src/qmdb/keyless/mod.rs
```diff
@@ -363,6 +363,9 @@ where
 
     /// Prune historical operations prior to `loc`. This does not affect the db's root.
     ///
+    /// `prune` requires no prior commit. After a crash, the database remains recoverable;
+    /// uncommitted operations are not guaranteed to survive.
+    ///
```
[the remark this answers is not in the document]
roberto-bayardo: Fixed in 506601039 — the bullet now matches the actual gate (`loc` > the inactivity floor, which `apply_batch` advances before durability), consistent with the `any` and `immutable` wording.

patrick-ogrady (approved):
its a tricky business

PR #4239, opened 2026-07-13
[glue] Resume interrupted state sync from persisted floor

clabby:
## Overview

An interrupted state sync currently persists only its selected floor's height and commitment. On restart, `SyncPlan` must probe for another finalization; if that probe returns a lagging floor, startup attempts to move behind the persisted floor and panics instead of reopening the existing sync journals.

Persist the full in-progress finalization so `SyncPlan` can resume directly from it, ignore probed floors that are not newer, and keep repeated `with_floor` selections monotonic. Completed state sync remains height-only, and marshal's completed-startup behavior is unchanged.

This supersedes #4230 with the narrower interrupted-state-sync fix. The separate Constantinople startup-wiring issue should be fixed in Constantinople: https://github.com/commonwarexyz/constantinople/pull/38

### DIFF
--- a/glue/src/stateful/actor/syncer/mod.rs
+++ b/glue/src/stateful/actor/syncer/mod.rs
@@ -35,111 +35,44 @@ const SYNC_STATE_KEY: FixedBytes<1> = fixed_bytes!("C0");
 
 type BlockDigest<A, E> = <<A as Application<E>>::Block as Digestible>::Digest;
 
-/// Durable identity for an in-progress state sync floor.
-///
-/// The height enforces monotonic restarts, and the commitment distinguishes
-/// conflicting blocks at the same height.
-#[derive(Clone, Debug, PartialEq, Eq)]
-pub(crate) struct FloorMarker<C>
-where
-    C: Digest,
-{
-    height: Height,
-    commitment: C,
-}
-
-impl<C> FloorMarker<C>
-where
-    C: Digest,
-{
-    /// Constructs a durable floor marker from the resolved floor block.
-    pub(crate) const fn new(height: Height, commitment: C) -> Self {
-        Self { height, commitment }
-    }
-
-    /// Ensures a newly selected floor is compatible with this persisted one.
-    ///
-    /// Restarts may resume from the same floor or advance to a newer one, but
-    /// must never move backward or switch to a different block at the same height.
-    pub(crate) fn ensure_not_behind(&self, selected: &Self) {
-        assert!(
-            selected.height >= self.height,
-            "selected state sync floor cannot move behind the persisted in-progress floor",
-        );
-
-        if selected.height == self.height {
-            assert!(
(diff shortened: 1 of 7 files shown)


glue/src/stateful/actor/syncer/mod.rs
```diff
@@ -265,21 +222,41 @@ where
...
+    pub(crate) async fn begin_sync(&mut self, height: Height, floor: Finalization<S, C>) {
         match self.metadata.get(&SYNC_STATE_KEY) {
-            Some(SyncState::InProgress(existing)) => {
-                existing.ensure_not_behind(&floor);
+            Some(SyncState::InProgress {
+                height: existing_height,
+                floor: existing,
+            }) => {
+                assert!(
+                    height >= *existing_height,
```
patrick-ogrady: do we need height? can't we just operate over `round`?

clabby: I think we can here, so long as we still consider the height as the DB anchor. Updated.

glue/src/stateful/actor/syncer/mod.rs
```diff
@@ -35,92 +35,20 @@ const SYNC_STATE_KEY: FixedBytes<1> = fixed_bytes!("C0");
...
-    C: Digest,
-{
-    type Cfg = ();
-
-    fn read_cfg(reader: &mut impl Buf, _: &()) -> Result<Self, Error> {
-        Ok(Self {
-            height: Height::read(reader)?,
-            commitment: C::read_cfg(reader, &())?,
-        })
-    }
```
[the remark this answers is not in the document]
clabby: Breaking change is intentional. This module is in `ALPHA`.

glue/src/stateful/actor/syncer/mod.rs
```diff
@@ -269,17 +216,34 @@ where
...
     ///
     /// If an interrupted state sync already stored a floor, the newly selected
-    /// floor must resume from that same floor or a later one.
-    pub(crate) async fn begin_sync(&mut self, floor: FloorMarker<C>) {
+    /// floor must resume from the same or a later consensus round.
+    pub(crate) async fn begin_sync(&mut self, floor: Finalization<S, C>) {
         match self.metadata.get(&SYNC_STATE_KEY) {
             Some(SyncState::InProgress(existing)) => {
-                existing.ensure_not_behind(&floor);
+                assert!(
```
patrick-ogrady, line 237: Is the idea that the caller checks this first and instead provides the updated one?

clabby: No, the caller doesn’t need to check. `SyncPlan::init` loads the persisted floor, and `with_floor` ignores the candidate unless it’s newer. Stateful then passes the selected floor here.

patrick-ogrady (approved):

PR #4240, opened 2026-07-13
[storage] Preserve variable journal corruption evidence

roberto-bayardo:
## Summary

- reject conflicting offsets blob partitions before init_at_size stages a destructive reset
- reject recovery watermarks that vouch past the retained data end
- reject data that cannot reach a trusted recovery watermark instead of retrying destructively from the pruning boundary
- preserve the legitimate stale-watermark repair after prune and add retry-stable corruption regressions

## Problem

Variable-journal recovery repairs offsets and data divergence left by legitimate crashes. Two branches also treated loss of data covered by the durable offsets watermark as repairable: a watermark beyond the retained data end fell back to the pruning boundary, and a rebuild that could not reach its trusted watermark retried from that boundary. Both paths could persist a shorter, apparently clean history and erase the evidence that acknowledged data had disappeared.

Separately, variable init_at_size delegated to the fixed offsets journal internal clear path without first checking for conflicting legacy and new offsets partitions. The staged reset could erase both partitions and hide that corruption.

## Fix

Split stale-watermark handling by cause. A watermark below the current pruning boundary remains a valid post-prune state and still rebuilds from the boundary. A watermark beyond retained data now returns Corruption with the watermark, offsets bounds, and retained-data end. If replay cannot reach an otherwise valid watermark, recovery returns Corruption instead of retrying from the boundary. These failures remain stable across reopen and preserve the surviving data evidence.

Run the same Partition::select preflight as fixed init_at_size before writing reset intent, so conflicting offsets partitions fail without mutation.

These changes affect only out-of-model loss of synced data or conflicting on-disk partitions; normal crash recovery remains unchanged.

## Validation

- just test -p commonware-storage journal::contiguous::variable::tests (90 passed)
- just clippy -p commonware-storage
- just check-fmt

### DIFF
--- a/storage/src/journal/contiguous/variable.rs
+++ b/storage/src/journal/contiguous/variable.rs
@@ -1810,18 +1810,6 @@ impl<E: Context, V: CodecShared> Journal<E, V> {
             compressed,
         )
         .await?;
-        let data_size = data_size.ok_or_else(|| {
-            if data_sync_start == offsets_bounds.start {
-                Error::Corruption(format!(
-                    "data blobs shorter than pruning boundary {}",
-                    offsets_bounds.start
-                ))
-            } else {
-                Error::Corruption(format!(
-                    "data blobs shorter than offsets recovery watermark {data_sync_start}"
-                ))
-            }
-        })?;
 
         // Final invariant checks. These hold by construction after alignment, but the inputs are
         // recovered from disk, so a violation means corruption rather than a logic bug.
@@ -1967,8 +1955,9 @@ impl<E: Context, V: CodecShared> Journal<E, V> {
 
     /// Rebuild the offsets suffix by replaying the data blobs from a recovery anchor.
     ///
-    /// Returns `Ok(None)` if the anchor is ahead of the data. If replay finds a short blob after
-    /// the anchor, recovery truncates newer blobs and returns the contiguous data-backed size.
+    /// Returns corruption if the data does not reach the anchor. If replay finds a short blob
+    /// after the anchor, recovery truncates newer blobs and returns the contiguous data-backed
+    /// size.
     async fn rebuild_offsets_from_anchor(
         partition: &Partition<E>,
         pending: &mut BTreeMap<u64, Writer<E::Blob>>,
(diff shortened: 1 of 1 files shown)


storage/src/journal/contiguous/variable.rs
```diff
@@ -1806,29 +1810,17 @@ impl<E: Context, V: CodecShared> Journal<E, V> {
...
+            if data_sync_start == offsets_bounds.start {
+                Error::Corruption(format!(
+                    "data blobs shorter than pruning boundary {}",
+                    offsets_bounds.start
+                ))
+            } else {
+                Error::Corruption(format!(
+                    "data blobs shorter than offsets recovery watermark {data_sync_start}"
+                ))
+            }
```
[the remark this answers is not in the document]
roberto-bayardo: Accurate observation, but a non-issue in consequence — the guarantee that matters holds:

- **The irreplaceable evidence survives.** The acknowledged-loss evidence is the watermark, and `rewind` cannot lower it on these paths: `lower_watermark` only fires when the current watermark exceeds the rewind target, and rebuild rewinds *to* the anchor (the watermark itself, or a boundary below it). The watermark proving durable data vanished is still on disk after the failed init.
- **Retry-stable.** After the partial rewind, every reopen still ends in `Corruption`: the offsets journal reopens cleanly (size == watermark), then align either hits the loud `watermark > retained_data_end_bound` arm or re-runs the rebuild to the same `Ok(None)`. The message can differ between attempts; the classification cannot.
- **What the rewind destroys is derived data.** Offsets entries above the watermark are an index over the data blobs, reconstructible by scanning them — that is what the rebuild path does. And on the `Ok(None)` path the data blobs themselves are untouched (the `skip > 0` return precedes the torn-truncation repair).

Pre-scanning the anchor's blob before rewinding would make the failure byte-for-byte pristine, but it buys nothing a salvage tool couldn't recompute, so not worth the complexity.

storage/src/journal/contiguous/variable.rs
```diff
@@ -1969,9 +1967,8 @@ impl<E: Context, V: CodecShared> Journal<E, V> {
 
     /// Rebuild the offsets suffix by replaying the data blobs from a recovery anchor.
     ///
-    /// Returns `Ok(None)` if the anchor is ahead of the data and callers should retry from an
-    /// earlier point. If replay finds a short blob after the anchor, recovery truncates newer
-    /// blobs and returns the contiguous data-backed size.
+    /// Returns `Ok(None)` if the anchor is ahead of the data. If replay finds a short blob after
```
danlaine: I think this doesn't need to return an `Option` anymore? Since `Ok(None)` will be reported as corruption by the caller anyway

roberto-bayardo: yep good catch, removed the Option.

patrick-ogrady (approved):

PR #4241, opened 2026-07-13
[marshal] Sync finalized archives off the mailbox path

patrick-ogrady:
## Overview

Marshal makes finalized-archive writes durable with a blocking sync inside the mailbox arm that stores them, so every mailbox caller (including pure prunable-cache reads like a proposer's `get_verified`) waits on disk latency. Durability is only required before a block is dispatched to the application, not before the actor serves its next message.

Start pooled, non-blocking archive syncs on the finalization and resolver-repair paths instead, and gate dispatch on durability with `DispatchGate`: `store_finalization` defers dispatch at or above each buffered write, `start_finalized_sync` adopts the deferred writes as a batch, and each completed sync releases every batch it covers (a later-started sync covers all earlier writes, so release is by prefix and out-of-order completion is safe). A blocking sync clears the gate. This preserves the crash-safety invariant that a block reaches the application (and its ack can advance the processed floor) only after its archive writes are durable, including when a stale floor anchor triggers dispatch mid-batch before the batch's sync starts. Floor-anchor installation and startup repair keep their blocking syncs because work later in the same arm requires durability. Stores without a non-blocking sync path (e.g. the immutable archive, whose `start_sync` completes the sync before returning a ready handle) keep today's blocking behavior with no loss of safety.

New coverage runs marshal against paced stores that model a slow sync: the mailbox stays responsive while a sync is in flight, a stale floor anchor cannot flush a buffered write to the application early, and overlapping pooled syncs release dispatch per batch.

Also included: `Archive::has(Identifier::Key)` previously delegated to a full `get` in both archive implementations, reading (and discarding) the value from the uncached value blob after confirming the key. Both now stop at the key index entry — `prunable` via a new entry-only `has_key` and `immutable` via a new `Freezer::has` — mirroring `has_at`'s "never read values" pattern. Marshal's `HintFinalized` skip-check now uses a `Certificates::has` index probe (an in-memory interval/ordinal lookup) instead of fetching and decoding the finalization it was about to discard.

Separately, the deployer now installs the `fq` qdisc on every tx queue of the primary interface (falling back to root `fq` on single-queue interfaces) and tunes TCP buffers for cross-region flows. `net.core.default_qdisc` only applies to interfaces attached after it is set, so BBR was running without `fq` pacing enforcement. A `missing_const_for_fn` allow in `bls12381` rides along for feature combinations where `Lazy::get` is not `const`.

### DIFF
--- a/consensus/src/marshal/core/actor.rs
+++ b/consensus/src/marshal/core/actor.rs
@@ -69,6 +69,19 @@ struct ResolverDelivery<V: Variant> {
     response: oneshot::Sender<bool>,
 }
 
+/// Completion marker for entries in the actor's durability sync pool.
+enum PooledSync {
+    /// Observed only so the fatal policy in `Durable` applies; no
+    /// continuation runs on completion.
+    Observed,
+    /// A finalized-archive sync batch became durable. Carries the sequence
+    /// assigned by [`Actor::start_finalized_sync`] so the completion arm can
+    /// release the application-dispatch barrier for every batch the sync
+    /// covers (a sync makes durable every write accepted before it started,
+    /// including batches registered under earlier sequences).
+    Finalized(u64),
+}
+
 /// The [Actor] is responsible for receiving uncertified blocks from the broadcast mechanism,
 /// receiving notarizations and finalizations from consensus, and reconstructing a total order
 /// of blocks.
@@ -130,6 +143,13 @@ where
     tip: Height,
     // Outstanding subscriptions for blocks
     block_subscriptions: Subscriptions<V>,
+    // Finalized-archive write batches that are buffered but not yet durable,
+    // keyed by the sequence of the pooled sync that first covers them and
+    // holding the lowest height written in the batch. `try_dispatch_blocks`
+    // never dispatches at or above the lowest pending height.
+    pending_finalized_syncs: BTreeMap<u64, Height>,
+    // Sequence assigned to the next pooled finalized-archive sync.
+    finalized_sync_seq: u64,
 
     // ---------- Storage ----------
     // Prunable cache
@@ -240,6 +260,8 @@ where
                 pending_acks: PendingAcks::new(config.max_pending_acks.get()),
                 tip: Height::zero(),
                 block_subscriptions: Subscriptions::new(),
+                pending_finalized_syncs: BTreeMap::new(),
+                finalized_sync_seq: 0,
                 cache,
                 finalizations_by_height,
                 finalized_blocks,
@@ -355,11 +377,12 @@ where
         // Create a local pool for waiter futures.
         let mut waiters = AbortablePool::<Result<Arc<V::Block>, SubscriptionKeyFor<V>>>::default();
 
-        // Observe durable syncs that no consensus caller awaits (the notarization
-        // path). A flush failure inside `start_sync` is reported only through the
-        // returned handle, so every handle must be observed to apply the fatal
-        // policy; this pool does so without blocking the actor on fsync.
-        let mut syncs = Pool::<bool>::default();
+        // Observe durable syncs that no consensus caller awaits (the
+        // notarization and finalization paths). A flush failure inside
+        // `start_sync` is reported only through the returned handle, so every
+        // handle must be observed to apply the fatal policy; this pool does so
+        // without blocking the actor on fsync.
+        let mut syncs = Pool::<PooledSync>::default();
 
         // Anchor all startup work under a single root span. Tip recovery, floor
         // installation, gap repair, and the initial dispatch all run before any
@@ -392,6 +415,9 @@ where
             }
 
             // Attempt to repair any gaps in the finalized blocks archive, if there are any.
+            //
+            // Sync inline (blocking): the mailbox is not being served yet, and
+            // the dispatch below must observe durable repairs.
             if self
                 .try_repair_gaps(&mut buffer, &mut resolver, &mut application)
                 .await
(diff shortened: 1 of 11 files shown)


consensus/src/marshal/core/actor.rs
```diff
@@ -753,8 +791,23 @@ where
...
-                        self.try_repair_gaps(buffer, resolver, application).await;
-                        self.sync_finalized().await;
+                        if self.try_repair_gaps(buffer, resolver, application).await {
+                            // Repair wrote blocks at heights below the new
+                            // finalization, so sync inline (blocking): the
+                            // batch's lowest height is not tracked, which the
+                            // pooled dispatch barrier would need. Repair only
+                            // runs while catching up, so the mailbox stall is
+                            // acceptable here.
+                            self.sync_finalized().await;
```
patrick-ogrady: we can use start_sync here

consensus/src/marshal/core/actor.rs
```diff
@@ -964,7 +1017,9 @@ where
 
         if needs_sync {
             // Sync archives before responding to peers so accepted repair data is
-            // durable before this node serves it.
+            // durable before this node serves it. Inline (blocking) is required:
```
patrick-ogrady: that's not true? we can use start_sync here?

consensus/src/marshal/core/actor.rs
```diff
@@ -1241,6 +1297,9 @@ where
                 .expect("pending floor anchor missing");
             self.update_processed_round_floor(height, finalization.round(), resolver)
                 .await;
+            // Sync inline (blocking): floor transitions are rare, and repair
```
patrick-ogrady: remove this comment

consensus/src/marshal/core/actor.rs
```diff
@@ -1271,6 +1330,9 @@ where
             }
         )
         .expect("failed to store floor anchor");
+        // Sync inline (blocking): every step below (metadata sync, pruning,
```
patrick-ogrady: remove this comment

consensus/src/marshal/core/actor.rs
```diff
@@ -1304,6 +1366,10 @@ where
         // Intentionally keep existing block subscriptions alive. Canceling
         // waiters can have catastrophic consequences (nodes can get stuck in
         // different views) as actors do not retry subscriptions on failed channels.
+        //
+        // Sync inline (blocking): floor transitions are rare, and repair
```
patrick-ogrady: remove this comment

consensus/src/marshal/core/actor.rs
```diff
@@ -1725,25 +1791,30 @@ where
...
     /// handler updates the processed height.
     ///
-    /// Callers must only invoke this after [`Self::sync_finalized`] has made any
-    /// preceding finalized-archive writes durable. In other words, anything fed
-    /// to the application from this method is already durably persisted in marshal.
+    /// Anything fed to the application from this method is already durably
+    /// persisted in marshal: callers must make preceding finalized-archive
+    /// writes durable with [`Self::sync_finalized`] (blocking) or register
+    /// them with [`Self::start_finalized_sync`] (pooled) before yielding to
+    /// the `select_loop!`. Registered batches gate dispatch below: no block at
```
patrick-ogrady: cleanup this comment...sounds very clunky

consensus/src/marshal/core/actor.rs
```diff
@@ -1725,25 +1791,30 @@ where
...
+    /// durable before the ack handler advances the processed floor height:
     ///
     /// ```text
     /// Iteration N (caller):
-    ///   store_finalization  ->  Archive::put (buffered)
-    ///   sync_finalized      ->  archive durable
-    ///   try_dispatch_blocks  ->  sends blocks to app, enqueues pending acks
+    ///   store_finalization   ->  Archive::put (buffered)
+    ///   sync_finalized       ->  archive durable
+    ///     (or start_finalized_sync -> dispatch of the batch deferred to the
```
patrick-ogrady: not sure we need this?

consensus/src/marshal/core/actor.rs
```diff
@@ -130,6 +142,13 @@ where
     tip: Height,
     // Outstanding subscriptions for blocks
     block_subscriptions: Subscriptions<V>,
+    // Finalized-archive write batches that are buffered but not yet durable,
+    // keyed by the sequence of the pooled sync that first covers them and
+    // holding the lowest height written in the batch. `try_dispatch_blocks`
+    // never dispatches at or above the lowest pending height.
+    pending_finalized_syncs: BTreeMap<u64, Height>,
+    // Sequence assigned to the next pooled finalized-archive sync.
+    finalized_sync_seq: u64,
```
patrick-ogrady: is this necessary?

consensus/src/marshal/core/actor.rs
```diff
@@ -1819,6 +1904,78 @@ where
...
+    }
+
+    /// Releases the dispatch barrier for every finalized-archive batch covered
+    /// by the completed pooled sync `seq`.
+    ///
+    /// A sync makes durable every write accepted before it started, so all
+    /// batches with an equal or lower sequence are released, regardless of the
+    /// order in which pooled syncs complete.
+    fn finalized_sync_completed(&mut self, seq: u64) {
+        self.pending_finalized_syncs = self.pending_finalized_syncs.split_off(&(seq + 1));
```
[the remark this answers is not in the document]
patrick-ogrady: This is inverted. `BTreeMap::split_off(&key)` returns the entries with keys **at or after** `key` and leaves the entries before it in `self` (std docs: "Returns everything after the given key, including the key"). Assigning the returned map back into `self.pending_finalized_syncs` therefore keeps the still-pending sequences (`> seq`) and drops the released ones (`<= seq`) — exactly the intended release of every batch covered by the completed sync.

The claimed failure ("a lone pooled sync never clears") is also refuted empirically: `test_standard_finalization_sync_does_not_block_mailbox` asserts the finalized block reaches the application once its lone pooled sync completes, which requires that entry to clear. The test passes.

consensus/src/marshal/core/actor.rs
```diff
@@ -961,12 +1002,9 @@ where
         // Attempt to fill gaps before handling produce requests so we can serve
         // data received earlier in the same batch.
         needs_sync |= self.try_repair_gaps(buffer, resolver, application).await;
-
         if needs_sync {
```
patrick-ogrady: add a \n and comment

storage/src/freezer/storage.rs
```diff
@@ -919,11 +922,7 @@ impl<E: BufferPooler + Context, K: Array, V: CodecShared> Freezer<E, K, V> {
 
             // Check if this key matches
             if key_entry.key.as_ref() == key.as_ref() {
-                let value = self
-                    .oversized
-                    .get_value(section, key_entry.value_offset, key_entry.value_size)
```
patrick-ogrady, line 924: 🚀

PR #4242, opened 2026-07-13
[glue] Expose initial sync targets

clabby:
## Overview

Make initial state sync setup simpler for applications: callers can ask their database set for initial targets instead of knowing QMDB internals, hard-coding roots, or opening throwaway databases.

This adds cheap initial-target computation and updates the glue fixtures to use it.

### DIFF
--- a/glue/src/stateful/db/any.rs
+++ b/glue/src/stateful/db/any.rs
@@ -25,6 +25,7 @@ use commonware_storage::{
         any::{
             batch::{MerkleizedBatch, Staged, UnmerkleizedBatch},
             db::Db,
+            initial_root,
             operation::{Operation, Update},
             ordered, unordered,
             value::{self, FixedEncoding, ValueEncoding, VariableEncoding},
@@ -514,6 +515,13 @@ where
         <Self>::init(context, config).await
     }
 
+    fn initial_sync_target() -> Self::SyncTarget {
+        AnySyncTarget::new(
+            initial_root::<F, unordered::Update<K, FixedEncoding<V>>, H>(),
+            non_empty_range!(Location::new(0), Location::new(1)),
+        )
+    }
+
     async fn new_batch(db: &Arc<TracedAsyncRwLock<Self>>) -> Self::Unmerkleized {
         let inner = db.read().await;
         AnyUnmerkleized {
@@ -609,6 +617,13 @@ where
 
     async fn init(context: E, config: Self::Config) -> Result<Self, Error<F>> {
         <Self>::init(context, config).await
+    }
+
+    fn initial_sync_target() -> Self::SyncTarget {
+        AnySyncTarget::new(
+            initial_root::<F, unordered::Update<K, VariableEncoding<V>>, H>(),
+            non_empty_range!(Location::new(0), Location::new(1)),
+        )
     }
 
     async fn new_batch(db: &Arc<TracedAsyncRwLock<Self>>) -> Self::Unmerkleized {
--- a/glue/src/stateful/db/mod.rs
+++ b/glue/src/stateful/db/mod.rs
@@ -1519,6 +1519,378 @@ mod tests {
         time::Duration,
     };
 
+    mod initial_sync_targets {
+        use super::ManagedDb;
+        use commonware_cryptography::{sha256::Digest, Sha256};
+        use commonware_parallel::Sequential;
+        use commonware_runtime::{
+            buffer::paged::CacheRef, deterministic, Runner as _, Supervisor as _,
+        };
+        use commonware_storage::{
+            journal::contiguous::{
+                fixed::Config as FixedJournalConfig, variable::Config as VariableJournalConfig,
+            },
+            merkle::{full::Config as MerkleConfig, mmr},
+            qmdb::{
+                any as storage_any, current as storage_current, immutable as storage_immutable,
+                keyless as storage_keyless,
+            },
+            translator::TwoCap,
+        };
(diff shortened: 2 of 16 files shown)


storage/src/qmdb/mod.rs
```diff
@@ -84,6 +88,17 @@ pub const fn hasher<H: Hasher>() -> StandardHasher<H> {
     StandardHasher::new(ROOT_BAGGING)
 }
 
+fn root_from_initial_commit<F: Family, H: Hasher>(operation: &impl Encode) -> H::Digest {
```
patrick-ogrady: I think you could improve the function name (given it is technically any operation?)

not sure what it should be...

clabby: `single_operation_root`? https://github.com/commonwarexyz/monorepo/pull/4242/commits/4e0abec71051c002bb1bbd3e301f3cba89d6ce11

storage/src/qmdb/mod.rs
```diff
@@ -84,6 +88,17 @@ pub const fn hasher<H: Hasher>() -> StandardHasher<H> {
     StandardHasher::new(ROOT_BAGGING)
 }
 
+fn single_operation_root<F: Family, H: Hasher>(operation: &impl Encode) -> H::Digest {
```
patrick-ogrady, line 95: add a comment?

glue/src/stateful/db/mod.rs
```diff
@@ -1553,6 +1571,10 @@ mod tests {
         type Config = ();
         type SyncTarget = u64;
 
+        fn initial_sync_target() -> Self::SyncTarget {
+            0
```
patrick-ogrady: dumb question: if the initial_root is 1 item, why is this 0?

storage/src/qmdb/mod.rs
clabby: Added context on the single-operation contract and why we compute it directly in 78c78fe.

glue/src/stateful/db/mod.rs
clabby: That 0 was an arbitrary sentinel in the test-only rewind mock, not a QMDB leaf count. I made the unsupported initialization path explicitly unreachable in 78c78fe so it no longer implies otherwise.

glue/src/stateful/tests/mocks.rs
```diff
@@ -59,6 +59,10 @@ impl<E: Send> ManagedDb<E> for TestDb {
     type Config = ();
     type SyncTarget = u64;
 
+    fn initial_sync_target() -> Self::SyncTarget {
+        0
```
patrick-ogrady: last one 👁️

storage/src/qmdb/mod.rs
```diff
@@ -84,6 +88,21 @@ pub const fn hasher<H: Hasher>() -> StandardHasher<H> {
     StandardHasher::new(ROOT_BAGGING)
 }
 
+/// Return the root of an operation log containing only `operation`.
+///
+/// This lets database variants derive their initial root from the bootstrap commit without
+/// opening a database.
+fn single_operation_root<F: Family, H: Hasher>(operation: &impl Encode) -> H::Digest {
```
patrick-ogrady, line 95: Would this be better suited to live in `merkle`? Are there other helpers like this there?

clabby: I looked and there is no equivalent helper in merkle. I think this belongs in QMDB: it encodes a QMDB operation and selects QMDBs ROOT_BAGGING policy, while merkle already provides the generic leaf_digest and root primitives. Moving those two calls would add a one-off pass-through helper without moving the QMDB-specific setup.

storage/src/qmdb/keyless/mod.rs
```diff
@@ -71,6 +73,17 @@ pub use compact::{
 };
 pub use operation::Operation;
 
+/// Compute the authenticated root of a newly initialized database without opening storage.
+pub fn initial_root<F, V, H>() -> H::Digest
```
patrick-ogrady, line 79: should we accept an arg for the `Commit(none)`?

clabby: I do not think so. Both standard and compact initialization always bootstrap with `Commit(None, 0)`; metadata only enters through later application commits. Accepting metadata here would let callers compute a root that a newly initialized QMDB can never produce.

patrick-ogrady: > Accepting metadata here would let callers compute a root that a newly initialized QMDB can never produce.

This would be a good inline comment?

clabby: https://github.com/commonwarexyz/monorepo/pull/4242/commits/49ecf44406c469d4b6779d70c6242386bcb37ae4

glue/src/stateful/db/any.rs
```diff
@@ -514,6 +515,13 @@ where
         <Self>::init(context, config).await
     }
 
+    fn initial_sync_target() -> Self::SyncTarget {
```
patrick-ogrady, line 518: do we have equality tests for all these? where we init a QMDB and ensure equality?

clabby: Added: https://github.com/commonwarexyz/monorepo/commit/f135d5a99ab71208c1aa93486d8d5f3b9ba517eb

patrick-ogrady (approved):

PR #4244, opened 2026-07-14
[storage/qmdb] Halve staged-merkleize serial bookkeeping and stage ancestor-diff resolutions

patrick-ogrady:
## Overview

`Staged::merkleize` spent roughly two thirds of its wall time in serial bookkeeping before the (already parallel) hash job, and any staged read that resolved through an uncommitted ancestor's diff fell back to the slow mutation path. This PR parallelizes or overlaps the serial bookkeeping and lets ancestor-resolved reads use the staged fast path.

```
before (everything on the calling task until the hash job):

  resolve updates -> read mutation locs -> emit ops+diff -> key-sort diff
    -> rescan diff for superseded locs
    -> floor raise (bitmap scan -> flat read, serially reassembled -> classify)
    -> sort + merge floor diff -> journal hashing (parallel)

after (bookkeeping overlapped; (n) marks where the task joins a pool job):

  task: prefetch candidates -(1)-> read mutation locs -> emit ops+diff
          -> floor raise (drain prefetch -> sharded reads -(2)-> classify)
          -> journal hashing -(3)-> done
  pool: (1) resolve updates    (2) key-sort diff    (3) sort + merge floor diff
```

## Cut the serial bookkeeping

Four pieces of `Staged::merkleize` ran inline on the calling task:

- **Last-write-wins dedup** hashed every key. It is now a forward overwrite into per-key winner slots, indexed by a slot -> distinct-key-id map built during `stage`/`expand` (off the merkleize path).
- **Floor-raise candidate reads** were reassembled serially. A new `read_many_sharded` keeps page-cache hits partitioned on the workers that probed them, and `read_ops` gets a fast path for strictly-ascending all-committed batches (no in-memory resolution or reordering).
- **Superseded locations** were found by rescanning the diff. They are now collected while entries are emitted.
- **The diff key-sort, floor-diff merge, and update resolution** now run as strategy jobs, overlapped with candidate gathering and journal hashing.

At 33k staged updates over 1M keys (12-thread strategy): serial portion 4.39ms -> 2.06ms, merkleize wall 6.6ms -> 4.1ms.

## Stage ancestor-resolved reads

Staged reads that resolved in an uncommitted ancestor's diff recorded no resolution, so their updates fell back to normal mutations: a BTreeMap insert, serial location gathering, a journal re-read, and per-op ancestor re-resolution. The fallback count grows ~n^2 with batch size, which was the measured superlinear cost on uncommitted parents.

```
staged read of key K resolves in...     before                 after

  the committed snapshot          ->  staged fast path     staged fast path
  an uncommitted ancestor's diff  ->  mutation fallback    staged fast path
                                      (~n^2: re-read,      (StagedLoc::Ancestor,
                                       re-resolve per op)   same emit position)
  batch mutations / nothing       ->  mutation path        mutation path
```

- Resolutions now record where they resolved (`StagedLoc::{Committed, Ancestor}`) and emit at the identical op-sequence position the fallback would have produced.
- The recorded base location is trusted while the resolving ancestor is alive at merkleize (its diff travels with the batch, so `apply_batch` re-resolves against it if the ancestor commits first). If the ancestor commits **and is freed** before merkleize, the recorded base is one transition stale, so merkleize supersedes the (now committed) resolved location itself.
- Only the unordered kind stages ancestors, expressed as the `STAGES_ANCESTORS` capability const (the ordered kind's cached next-key pointer is not in a diff entry).
- The serial update resolution is overlapped with a prefetch of committed-prefix floor-raise candidates.

Merkleize medians (1M-key db, 12 threads), ms:

| updates | depth 1 | depth 2 |
|---|---|---|
| 33k | 6.84 -> 4.76 | 7.90 -> 5.03 |
| 49k | 9.53 -> 6.67 | 11.84 -> 6.88 |
| 66k | 13.20 -> 8.93 | 18.82 -> 8.85 |

Per-op cost at depth 2 now falls with batch size instead of rising.

## Correctness

- Roots bit-identical to the unmodified baseline: a 39-case matrix (staged/mixed/creates/deletes/collisions, ordered and unordered) for the bookkeeping rework, and depth 0/1/2 x 33k/49k/66k x 2 seeds x 1/12 threads (ordered/keyless/current variants) for ancestor staging.
- New test `unordered_variable_staged_ancestor_commit_before_merkleize` pins the stale-base hazard end-to-end: stage through a grandparent, commit and free it, then merkleize and apply. Trusting the recorded base panics `update_known_loc` at apply (verified by ablation); a trusted `None` base for a created key would leak an active bit, caught by root parity.
- Two further tests complete the lifecycle coverage: `unordered_variable_staged_ancestor_alive_through_apply` exercises the alive-ancestor arm through both apply shapes (pending merge and ancestors-applied-first re-resolution), and a current-variant twin pins the re-derivation through the grafted layer. All three are ablation-verified (breaking the target arm fails exactly the covering tests).

## Also included

- A review round tightened the port: the floor-raise prefetch budget counts only steps the raise can consume (fresh-key creates are excluded), `resolve_updates` walks only touched ids and moves staged payloads instead of cloning them (dropping the public `Cached: Clone` bound), an empty diff skips its sort job, and single-shard `read_many` skips reassembly.
- The current variant's staged merkleize gains the same resolution/prefetch overlap: -3.9% (depth 1) / -5.9% (depth 0) pipeline median at 1M keys / 32k / 8 threads, roots identical.
- The merged floor-diff sortedness assert stays in release builds (nothing downstream re-validates it).
- `read_ops` drops its presorted early-return, unreachable behind the committed-region fast path.
- The bitmap `OnesIter` improvement originally on this branch landed separately as #4243.

https://claude.ai/code/session_01YEGqKvKQUMzrAfUaAcUUiT

### DIFF
--- a/storage/src/journal/authenticated.rs
+++ b/storage/src/journal/authenticated.rs
@@ -857,7 +857,10 @@ where
     }
 
     async fn read_many(&self, positions: &[u64]) -> Result<Vec<C::Item>, JournalError> {
-        let shards = self.read_many_sharded(positions).await?;
+        let mut shards = self.read_many_sharded(positions).await?;
+        if shards.len() == 1 {
+            return Ok(shards.pop().expect("length checked"));
+        }
         let mut items = Vec::with_capacity(positions.len());
         for shard in shards {
             items.extend(shard);
--- a/storage/src/journal/authenticated.rs
+++ b/storage/src/journal/authenticated.rs
@@ -739,6 +739,109 @@ macro_rules! impl_journal_new {
 impl_journal_new!(fixed, fixed::Config, CodecFixedShared);
 impl_journal_new!(variable, variable::Config<O::Cfg>, CodecShared);
 
+impl<F, E, C, H, S> Journal<F, E, C, H, S>
+where
+    F: Family,
+    E: Context,
+    C: Contiguous<Item: EncodeShared>,
+    H: Hasher,
+    S: Strategy,
+{
+    /// Like [`Contiguous::read_many`], but returns the items partitioned into the shards the
+    /// probe ran with; concatenating the shards yields the items in `positions` order.
+    ///
+    /// Large batches shard the page-cache probe across the strategy pool. Each shard
+    /// assembles its own hits while they are still cache-hot on the probing worker, so bulk
+    /// callers that can consume partitioned results (e.g. the floor raise, which classifies
+    /// candidates in chunks) skip the serial reassembly a flat result would require.
+    pub(crate) async fn read_many_sharded(
+        &self,
+        positions: &[u64],
+    ) -> Result<Vec<Vec<C::Item>>, JournalError> {
+        // An empty batch cannot shard: the parallel arm's chunk math needs a non-zero chunk
+        // size, and the policy may explore that arm at any batch size.
+        if positions.is_empty() {
+            return Ok(Vec::new());
+        }
+
+        // Probe page-cache hits synchronously and complete the misses with one batched read.
+        // The strategy policy decides per batch size whether the probe runs on the calling
+        // thread or sharded across the pool (one scratch buffer per shard and one cache-lock
+        // acquisition per blob a shard touches). The sortedness assert keeps contract
+        // violations deterministic: past it, a non-increasing batch would only trip per-shard
+        // validation when an inversion lands inside a single shard.
+        assert!(
+            positions.is_sorted_by(|a, b| a < b),
+            "positions must be strictly increasing"
+        );
+        let strategy = self.strategy();
+        let journal = &self.journal;
+        // Each shard yields its hits densely plus the shard-local indices it declined.
+        let probe = |positions: &[u64]| -> (Vec<C::Item>, Vec<usize>) {
+            let probed = journal.try_read_many_sync(positions);
+            let mut hits = Vec::with_capacity(probed.len());
+            let mut missed = Vec::new();
+            for (idx, item) in probed.into_iter().enumerate() {
+                match item {
+                    Some(item) => hits.push(item),
+                    None => missed.push(idx),
+                }
+            }
+            (hits, missed)
+        };
+        let shard_len = positions
+            .len()
+            .div_ceil(strategy.manual().parallelism())
+            .max(1);
+        let shards: Vec<(Vec<C::Item>, Vec<usize>)> = strategy.run(
+            positions.len(),
+            || vec![probe(positions)],
+            || {
+                let manual = strategy.manual();
+                manual.map_collect_vec(positions.chunks(shard_len).collect::<Vec<_>>(), &probe)
+            },
+        );
+
+        // The declined positions are a strictly increasing subsequence of `positions`, so one
+        // batched read serves them all.
+        let misses: Vec<u64> = shards
+            .iter()
+            .enumerate()
+            .flat_map(|(shard, (_, missed))| {
+                missed
+                    .iter()
+                    .map(move |idx| positions[shard * shard_len + idx])
+            })
+            .collect();
+        if misses.is_empty() {
+            return Ok(shards.into_iter().map(|(hits, _)| hits).collect());
+        }
+        let mut fetched = journal.read_many(&misses).await?.into_iter();
+
+        // Weave the fetched items back into each shard that declined positions.
+        let mut result = Vec::with_capacity(shards.len());
+        for (hits, missed) in shards {
+            if missed.is_empty() {
+                result.push(hits);
+                continue;
+            }
+            let total = hits.len() + missed.len();
+            let mut woven = Vec::with_capacity(total);
+            let mut hits = hits.into_iter();
+            let mut missed = missed.into_iter().peekable();
+            for idx in 0..total {
+                if missed.next_if_eq(&idx).is_some() {
+                    woven.push(fetched.next().expect("one fetched item per miss"));
+                } else {
+                    woven.push(hits.next().expect("one probed item per hit"));
+                }
+            }
+            result.push(woven);
+        }
+        Ok(result)
+    }
+}
+
 impl<F, E, C, H, S> Contiguous for Journal<F, E, C, H, S>
 where
     F: Family,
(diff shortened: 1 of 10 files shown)


storage/src/journal/contiguous/fixed.rs
```diff
@@ -1183,7 +1176,7 @@ impl<E: Context, A: CodecFixedShared> Reader<'_, E, A> {
         if positions.is_empty() {
             return Ok(Vec::new());
         }
-        assert!(
+        debug_assert!(
```
patrick-ogrady: make this assert

storage/src/journal/authenticated.rs
```diff
@@ -781,38 +776,97 @@ where
         );
         let strategy = self.strategy();
         let journal = &self.journal;
-        let mut items = strategy.run(
+        // Each shard yields its hits densely plus the shard-local indices it declined.
```
patrick-ogrady, line 780: add \n before comments above code blocks

storage/src/journal/authenticated.rs
```diff
@@ -781,38 +776,97 @@ where
...
+            let mut missed = Vec::new();
+            for (idx, item) in probed.into_iter().enumerate() {
+                match item {
+                    Some(item) => hits.push(item),
+                    None => missed.push(idx),
+                }
+            }
+            (hits, missed)
+        };
+        let shard_len = positions
```
patrick-ogrady: move inside concurrent block?

storage/src/journal/authenticated.rs
```diff
@@ -781,38 +776,97 @@ where
...
-        let misses: Vec<u64> = positions
+        let misses: Vec<u64> = shards
             .iter()
-            .zip(&items)
-            .filter_map(|(&position, item)| item.is_none().then_some(position))
+            .enumerate()
+            .flat_map(|(shard, (_, missed))| {
+                missed
+                    .iter()
+                    .map(move |idx| positions[shard * shard_len + idx])
```
patrick-ogrady: shard_len won't have anything to do with the serial output?

storage/src/qmdb/any/operation/update/mod.rs
```diff
@@ -24,12 +24,20 @@ pub trait Update: sealed::Sealed + Clone + Send + Sync + 'static {
...
     /// Payload cached alongside the resolved location of a batch read, consumed by merkleize.
-    type Cached: Send + Sync;
+    type Cached: Clone + Send + Sync;
 
     /// Whether merkleize may emit a staged delete directly at its read-resolved committed
     /// location. When false, staged deletes fall back to normal mutations.
     const STAGES_DELETES: bool;
 
+    /// Whether merkleize may stage a read that resolved in an uncommitted ancestor's diff,
+    /// expressed as the cached payload such a staged read carries. `None` means ancestor
```
patrick-ogrady: Remove: ` `None` means ancestor
    /// resolutions cannot be staged and fall back to normal mutations. Staging requires a
    /// payload, so the capability is the payload itself rather than a `bool`: the ordered
    /// kind is `None` because its cached payload is the resolved op's next-key pointer,
    /// which a diff entry does not carry.`

storage/src/qmdb/any/batch.rs
```diff
@@ -41,17 +41,77 @@ type DiffSlice<K, F, V> = [(K, DiffEntry<F, V>)];
...
-/// Staged update entry: key, old committed location, cached payload from the old update, and
+/// Where a staged read resolved: a committed location, or an uncommitted ancestor-diff
+/// location. Either way the location orders the staged write among this batch's emitted
+/// operations; they differ in which committed location (if any) the write supersedes.
+///
+/// An `Ancestor` resolution's `base_old_loc` reflects the chain at stage time. If the
+/// resolving ancestor commits (and drops out of the alive chain) before merkleize, `loc`
+/// migrates into the committed region and becomes the superseded committed location itself;
+/// the merkleize consumption re-derives the base against the merkleize-time committed
+/// boundary rather than trusting the stage-time value.
```
patrick-ogrady: Why can't we trust this value? I think we can...

storage/src/qmdb/any/batch.rs
```diff
@@ -865,20 +1015,34 @@ where
...
 
+        // Key-sort the diff as one job on the strategy: candidate classification (after the
+        // first floor-raise read below) is the earliest consumer that needs it sorted, so the
+        // sort overlaps the candidate gathering and read instead of the calling task.
+        let mut diff_sort = Some(db.strategy().spawn(move |strategy| {
+            let mut diff = diff;
+            strategy.sort_by(&mut diff, |a, b| a.0.cmp(&b.0));
+            diff
+        }));
+        // Replaced by the sorted diff at the first `diff_sort` await; empty until then.
```
patrick-ogrady: Add a \n before?

storage/src/qmdb/any/batch.rs
```diff
@@ -865,20 +1015,34 @@ where
...
+        let mut diff: DiffVec<U::Key, F, U::Value> = Vec::new();
+
+        // New diff entries for keys moved by the floor raise, merged into `diff` below.
+        let mut floor_diff = Vec::new();
+
```
patrick-ogrady: remove this \n?

storage/src/qmdb/any/batch.rs
```diff
@@ -865,20 +1015,34 @@ where
...
             let mut scan_from = floor;
-            let mut floor_diff = Vec::with_capacity(total_steps as usize);
-            let mut superseded_locs: Vec<_> = diff
-                .iter()
-                .filter_map(|(_, entry)| entry.base_old_loc())
-                .collect();
-            strategy.sort_by(&mut superseded_locs, |a, b| a.cmp(b));
-            superseded_locs.dedup();
+            floor_diff.reserve(total_steps as usize);
+            // Locations are unique (each committed location belongs to exactly one key), so a
```
patrick-ogrady, line 1066: add a \n?

storage/src/qmdb/any/batch.rs
```diff
@@ -894,8 +1058,27 @@ where
                 // `scan_from` tracks prefetch progress separately from `floor`, so
                 // early exit cannot leave `floor` past unprocessed candidates.
                 let limit = (total_steps - moved) as usize;
-                let mut candidates = Vec::with_capacity(limit);
-                scan_from = fill_candidates(scan_from, fixed_tip, limit, &mut candidates);
+                // Consume the prefetched committed prefix whole: it was gathered from the
```
patrick-ogrady, line 1088: add a \n?

storage/src/qmdb/any/batch.rs
```diff
@@ -905,23 +1088,39 @@ where
...
-                    })
-                    .collect();
+                // `read_candidates` omits locations already superseded by this diff, saving
+                // their read. Keep `resolved` and `outcomes` in that filtered order, then
+                // walk `candidates` below so superseded locations still advance the floor in
+                // scan order. Prefetched candidates skip the filter -- their ops were read
+                // ahead of time, and a superseded candidate's key always resolves in the
+                // diff to a different location, classifying it `Inactive`.
+                let pf_count: usize = pf_shards.iter().map(Vec::len).sum();
+                debug_assert!(pf_count <= candidates.len());
```
patrick-ogrady: assert

storage/src/qmdb/any/batch.rs
```diff
@@ -1223,41 +1455,70 @@ where
...
         }
 
-        // Locations are unique after last-write-wins dedup, so the parallel sort
-        // is deterministic.
-        strategy.sort_by(&mut staged_updates, |a, b| a.1.cmp(&b.1));
+        // Locations are unique after last-write-wins dedup (each key resolves to exactly one
+        // location, committed or ancestor), so the parallel sort is deterministic. Sorting
+        // compact `(location, slot)` pairs instead of the staged tuples keeps its memory
+        // traffic low; the tuples are then materialized in sorted order across the strategy.
+        strategy.manual().sort_by(&mut order, |a, b| a.0.cmp(&b.0));
```
patrick-ogrady: remove manual here?

storage/src/qmdb/any/batch.rs
```diff
@@ -1530,14 +1832,35 @@ where
     {
         let mut resolutions: Vec<StagedResolution<F, U>> =
             iter::repeat_with(|| None).take(keys.len()).collect();
-        let (mut results, unresolved) = self.resolve_uncommitted_reads(keys, db.strategy());
+        // Record ancestor-diff resolutions when the update kind stages them: the staged
```
patrick-ogrady, line 1929: add a \n before?

utils/src/bitmap/mod.rs
patrick-ogrady, line 1: remove the changes to this file

PR #4247, opened 2026-07-14
[storage/qmdb] Add `start_commit` to `qmdb::any` and `qmdb::current`

danlaine:
## Database API

`any::Db` and `current::Db` gain:

```rust
pub async fn start_commit(mut self) -> Result<(Self, Handle<()>), Error<F>>
```

Awaiting the returned handle gives the same guarantee as `commit()`: applied state survives a crash, with recovery possibly required on startup (`sync()` remains the way to avoid recovery entirely). The database stays fully usable while the handle is pending — reads always proceed, and applies do too until the log must write to storage (a filled write buffer or a blob rollover waits for the in-flight sync). A second commit waits for the first's sync before starting.

Errors preparing the commit (the Merkle flush) return eagerly; failures of the deferred fsync surface on the handle and again on the next durability operation. `current::Db::start_commit` does not persist bitmap metadata — `sync()` and `prune()` do. Keyless, immutable, and store are left blocking-only for now.

## Plumbing

`Mutable` gains `start_commit` as a required method; the contiguous journals forward to their inherent implementations from #4116. `authenticated::Journal::start_commit` starts the inner journal's commit concurrently with the Merkle flush and returns the journal's handle — the Merkle structure stays non-durable exactly as under `commit()`, realigned from the log on startup.

`commit()` at both layers stays a direct blocking implementation rather than `start_commit().await`: routing it through the handle would stop the inner journal's commit-duration histogram from observing blocking commits.

## Metrics

`start_commit_calls` counts pipelined commits, disjoint from `commit_calls`, so `commit_duration / commit_calls` remains a true average of blocking commits — the same split #4116 makes at the journal level.

## Test support

`runtime::mocks::drive_pending_syncs` drives a future while releasing the blob syncs it parks; opening storage under `DelayedSyncContext` persists recovered state and would otherwise hang. The #4116 commit-handle suite now runs through `Mutable::start_commit` (the test-local `CommitHandle` trait is gone), with new coverage at each layer: handle durability across reopen, appends and applies overlapping a parked sync, failure propagation and retention, prune draining the in-flight commit, and `start_commit` recovery across all 16 `any` variants plus `current`'s grafted-bitmap root.


runtime/src/mocks.rs
```diff
@@ -682,6 +683,22 @@ pub fn release_pending_syncs(pending: &PendingSyncs) {
     }
 }
 
+/// Drive `fut` to completion, releasing every sync parked on `pending` as it appears.
+///
+/// Opening storage under [DelayedSyncContext] can otherwise hang: initialization durably
+/// persists recovered state, and those syncs park before the caller can observe or release
+/// them.
```
roberto-bayardo: is this comment needed here? (I don't understand it FWIW, so suggest clarifying it if it belongs)

storage/src/qmdb/any/db.rs
```diff
@@ -865,6 +878,8 @@ where
     )]
     #[boxed]
     pub async fn commit(mut self) -> Result<Self, crate::qmdb::Error<F>> {
+        // Runs the log's commit rather than awaiting a start_commit handle so journal-level
```
patrick-ogrady: not sure this is necessary?

storage/src/journal/authenticated.rs
```diff
@@ -374,10 +375,30 @@ where
...
+        (self.journal, self.merkle) = (journal, merkle);
+
+        Ok((self, handle))
+    }
+
     /// Durably persist the journal. This is faster than `sync()` but does not guarantee that the
     /// Merkle structure is durably persisted, meaning recovery may be required on startup in the
     /// event of a crash.
     pub async fn commit(mut self) -> Result<Self, Error<F>> {
+        // Runs the inner journal's commit rather than awaiting a start_commit handle so the
```
patrick-ogrady: not sure this is necessary?

patrick-ogrady (commented):
very small nits

patrick-ogrady (approved):

patrick-ogrady (approved):

PR #4249, opened 2026-07-14
[runtime/iobuf] improve buffer pool sizing API

andresilva:
`BufferPoolConfig` now models the class layout as an explicit set of power-of-two size classes, each with its own tracked-buffer limit, instead of a contiguous `min_size..max_size` range with one uniform `max_per_class` buffer count. Classes no longer need to be contiguous: requests route to the smallest enabled class that fits, so disabling a class forwards its traffic to the next larger one. Layouts can be shaped with equal counts or equal tracked bytes per class, and `with_budget_bytes` now proportionally rescales the existing shape under a strict byte ceiling rather than flattening to a uniform count.

The allocation hot path is unchanged. Pool construction expands the enabled classes into the existing dense exponent-indexed routing vector, whose gap entries hold cloned handles aliasing the next enabled class, so lookup remains the existing arithmetic (max check, `next_power_of_two`, exponent subtraction) and the same single vector index as before.


runtime/src/iobuf/pool.rs
```diff
@@ -1345,6 +1603,11 @@ impl Drop for BufferPoolInner {
         // TLS cache entries own their own size-class references, if they return
         // later, they will park their buffer and release the reference that kept
         // the class alive.
+        //
+        // Routing entries alias the next enabled class in contiguous runs, so
+        // dropping consecutive duplicates leaves one live handle per unique
+        // class and drains each class once.
+        self.classes.dedup_by(|a, b| a.token == b.token);
```
clabby, line 1635: `debug_assert_eq!(self.classes.len(), self.config.size_classes().len())` following the dedup?

runtime/src/iobuf/pool.rs
```diff
@@ -105,50 +109,83 @@ pub(crate) enum BufferPoolThreadCacheConfig {
...
+/// requests route to the smallest enabled class that fits.
+///
+/// Shape builders do not commute. Each builder applies to the layout produced
+/// by the previous one: replacement builders ([`Self::with_size_class_range`],
+/// [`Self::with_size_classes`]) discard the current layout, uniform builders
+/// ([`Self::with_max_per_class`], [`Self::with_bytes_per_class`]) overwrite
+/// every enabled limit, and [`Self::with_budget_bytes`] snapshots and rescales
+/// the shape that exists at that call.
+#[derive(Clone, Debug)]
 pub struct BufferPoolConfig {
```
clabby, line 150: Given that `thread_cache_config` is crate-private (and `class_limits` is module-private), I think we should make the other fields private too? That way there’d be only a single way to construct this via the builder.

runtime/src/iobuf/pool.rs
andresilva: Went with regular `assert_eq!` because this is outside the hot path.

runtime/src/iobuf/pool.rs
andresilva: Updated. Added getters to the fields.

clabby (dismissed):

runtime/src/iobuf/pool.rs
```diff
@@ -195,21 +249,130 @@ impl BufferPoolConfig {
...
+    /// Requests that previously routed to the removed class route to the next
+    /// larger enabled class.
+    ///
+    /// # Panics
+    ///
+    /// - `size` is not a power of two
+    /// - `size` exceeds `isize::MAX`
+    /// - no class with `size` is enabled
+    /// - the class is the final enabled class
+    pub fn without_size_class(mut self, size: NonZeroUsize) -> Self {
```
patrick-ogrady, line 329: May be overkill to support this bon't ok to keep?

runtime/src/iobuf/pool.rs
```diff
@@ -1334,7 +1599,19 @@ struct Allocation {
 /// Internal state of the buffer pool.
 pub(crate) struct BufferPoolInner {
     config: BufferPoolConfig,
+    /// Exponent-indexed routing vector.
+    ///
+    /// Entry `i` serves requests that round up to `min_size << i`. Disabled
+    /// exponents hold cloned handles that alias the next enabled class, so
+    /// the allocation path resolves any request with plain arithmetic and one
```
patrick-ogrady, line 1612: 😍

clabby (dismissed):

patrick-ogrady (approved):

PR #4251, opened 2026-07-14
[storage/index] Push spilled collision inserts in amortized O(1)

patrick-ogrady:
## Overview

The partitioned ordered index spills an over-full partition to a `BTreeMap` to bound adversarial flooding, but the spilled representation did not actually bound the same-translated-key collision vector its docs claim to cover. When an insert hit a key already present in a spilled partition, the value chain was a `Vec` prepended with `insert(0, value)`. Each insert shifted the entire chain, so M colliding inserts cost O(M^2) — the same quadratic the spill exists to prevent, and a regression from the flat ordered index, whose head displacement into the overflow map is amortized O(1). Any deployment where untrusted input can repeatedly hit one translated key (a predictable or capped translator, or plain repeated inserts of one key) exposes this as an algorithmic-complexity DoS.

## Change

Spilled chains are now stored oldest-first, so the colliding insert is a plain `push` — amortized O(1) with the same 3-word `Vec` header as before. No consumer requires any value order (archive key lookups document "may return any of the associated values", and QMDB full-key-compares every candidate), but the index battery pins newest-first iteration across implementations, so reads keep serving it by iterating spilled chains in reverse. A `Rev<slice::Iter>` walks the end pointer down instead of the start pointer up, so the reversal is free. Because the inline sorted arrays iterate forward while spilled chains iterate backward, the internal read helpers now return a small two-variant iterator instead of `&[V]`.

Two consequences of the reversed layout:

- Spilling reverses each drained run once, since inline runs are stored newest-first.
- The cursor's spilled backing mirrors positions: run offset `off` addresses chain index `len - 1 - off`, and an insert at run offset `off` lands at chain index `len - off`. The Done-state append (at the oldest position) is therefore an O(chain) front insert, which matches the flat index's cursor exactly and is only reachable after an already-O(chain) iteration, so it introduces no new quadratic surface.

## Validation

Temporarily scaling the existing `run_index_large_collision_chain` battery (50k -> 400k same-key inserts through the spilling fixture) shows the quadratic and its removal:

| same-key inserts | before | after |
|---|---|---|
| 50k | 0.18s | 0.03s |
| 400k | 9.35s | 0.11s |

Before the change, 8x the items cost ~52x the time. After it, growth is linear. The full `index::` suite (156 tests) passes, including the generic cross-implementation battery routed through the spilled representation, which pins newest-first value order, cursor insert/delete/update semantics, metrics, and the de-spill lifecycle.

### DIFF
--- a/storage/src/index/partitioned/ordered/cursor.rs
+++ b/storage/src/index/partitioned/ordered/cursor.rs
@@ -11,7 +11,7 @@ use super::partition::Partition;
 use crate::index::Cursor as CursorTrait;
 use commonware_runtime::telemetry::metrics::{Counter, Gauge};
 use std::{
-    collections::{btree_map, hash_map, BTreeMap, HashMap},
+    collections::{btree_map, hash_map, BTreeMap, HashMap, VecDeque},
     ops::Range,
 };
 
@@ -45,7 +45,7 @@ enum Backing<'a, K: Ord + Copy, V> {
     /// Spilled partition: the key's values live in the side-table's `BTreeMap`, re-resolved on each
     /// access (spilling is the rare case, so the extra descent is off the hot path).
     Spilled {
-        spilled: &'a mut HashMap<usize, BTreeMap<K, Vec<V>>>,
+        spilled: &'a mut HashMap<usize, BTreeMap<K, VecDeque<V>>>,
         partition: usize,
         key: K,
     },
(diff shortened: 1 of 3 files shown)


roberto-bayardo:
Nothing actually depends on the ordering of colliding values, so we can always just insert into the back of the vector instead of the front to avoid the performance issue. Not sure in that case the VecDeque adds much?

roberto-bayardo:
Sol vibe-coded solution using Vec and avoids expensive insertion (though changes the item order): https://github.com/commonwarexyz/monorepo/pull/4252

PR #4275, opened 2026-07-17
[consensus/simplex] Recover Own Nullify after Restart

patrick-ogrady:
## Bug

If exactly 2f+1 honest validators can nullify a view (the other f contribute nothing) and all of them crash after journaling their nullify votes but before any nullification certificate is assembled, the network deadlocks on restart. Journal replay sets `broadcast_nullify`, so the post-restart timeout takes the retry path, which re-broadcasts the vote to peers but never re-delivers it to the local batcher (whose state is in-memory only). Every batcher then holds the 2f peer votes but not its own, one short of quorum forever.

## Fix

`Voter::timeout` now informs the batcher on every attempt (journaling remains first-attempt-only), and the batcher ignores duplicate nullifies instead of asserting (notarize and finalize keep their asserts since they are never re-sent). The duplicate check runs before reporting and verifier insertion, so retries neither re-report activity nor double-count toward quorum. A new deterministic test, `all_crash_after_nullify`, reproduces the deadlock and passes with the fix.

### DIFF
--- a/consensus/src/simplex/actors/batcher/round.rs
+++ b/consensus/src/simplex/actors/batcher/round.rs
@@ -240,37 +240,45 @@ impl<
     }
 
     /// Adds a vote that we constructed ourselves to the verifier.
+    ///
+    /// Duplicate nullifies are ignored (the voter re-sends its nullify vote on
+    /// every timeout retry).
+    ///
+    /// # Panics
+    ///
+    /// Panics if a notarize or finalize vote is added more than once.
     pub fn add_constructed(&mut self, message: Vote<S, D>) {
         match &message {
             Vote::Notarize(notarize) => {
-                // Report activity
-                self.reporter.report(Activity::Notarize(notarize.clone()));
-
                 // Our own votes are already verified
                 assert!(
                     self.votes.insert_notarize(notarize.clone()),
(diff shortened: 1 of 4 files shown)


BrendanChou (approved):

consensus/src/simplex/actors/batcher/round.rs
```diff
@@ -240,37 +240,45 @@ impl<
...
+
+                // Report activity
+                self.reporter.report(Activity::Notarize(notarize.clone()));
             }
             Vote::Nullify(nullify) => {
+                // The voter re-sends its nullify on every timeout retry (the
+                // batcher's state does not survive a restart), so duplicates
+                // are expected and ignored.
+                if !self.votes.insert_nullify(nullify.clone()) {
+                    return;
```
BrendanChou, line 267: Consider whether the caller of this function should know that the vote was dropped by adding a boolean return type. This doesn't matter much as it only affects metrics `consensus/src/simplex/actors/batcher/actor.rs:379`

However the current code is probably cleaner so may not be worth changing

patrick-ogrady: Not worth it imho

PR #4277, opened 2026-07-17
[storage/qmdb] Parallel Init Nits

patrick-ogrady:
Related: #4121

### DIFF
--- a/codec/src/util.rs
+++ b/codec/src/util.rs
@@ -1,6 +1,6 @@
 //! Codec utility functions
 
-use crate::{Error, extensions::ReadExt as _};
+use crate::Error;
 use bytes::Buf;
 
 /// Checks if the buffer has at least `len` bytes remaining. Returns an [Error::EndOfBuffer] if not.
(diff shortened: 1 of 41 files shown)


glue/src/stateful/db/p2p/standard/actor.rs
```diff
@@ -493,7 +490,7 @@ mod tests {
     fn db_config(suffix: &str, pooler: &impl BufferPooler) -> FixedConfig<TwoCap, Sequential> {
         let page_cache = CacheRef::from_pooler(pooler, NZU16!(101), NZUsize!(11));
         FixedConfig {
-            init_parallelism: InitParallelism::Serial,
+            init_workers: None,
```
patrick-ogrady, line 493: this makes it seem like there are no workers now...really what it does is provide extra workers?

codec/src/util.rs
```diff
@@ -17,10 +17,65 @@ pub fn at_least<B: Buf>(buf: &mut B, len: usize) -> Result<(), Error> {
 /// otherwise.
 #[inline]
 pub fn ensure_zeros<B: Buf>(buf: &mut B, size: usize) -> Result<(), Error> {
-    for _ in 0..size {
-        if u8::read(buf)? != 0 {
+    at_least(buf, size)?;
```
roberto-bayardo, line 20: Did you mean to include this? (doesn't look related to the PR)

glue/src/stateful/db/p2p/standard/actor.rs
roberto-bayardo: `init_concurrent_workers`?

workers are concurrent with the replay thread (and each other)

codec/src/util.rs
roberto-bayardo: I guess it's a decode optimization, which is a bit of a bottleneck here, so sgtm.

patrick-ogrady: Yeah, I did.

PR #4279, opened 2026-07-17
[storage] Make `async` `mut`able methods consume/produce `Self`, part 2

### LINKED ISSUE #4278: [storage] Make `async` `mut`able methods consume/produce `Self`
#4246 changes the QMDB stack by making types produce/consume Self when they're modified. As explain that PR's description:

> Storage mutators previously took &mut self. That shape cannot express two invariants we rely on:
> A cancelled mutation must not leave a usable-but-diverged handle. Dropping an in-flight commit/prune/rewind future could leave in-memory state ahead of, behind, or interleaved with durable state. Callers were expected to reconcile the two (or know not to touch the instance) — nothing enforced it.
> Mutable-method errors are fatal. After a failed mutation, the instance's in-memory state may not match disk. Comments said "do not reuse"; the type system allowed reuse.

This issue tracks the application of this pattern to the remaining storage structures:

1. Segmented journals — journal::segmented::fixed, variable, glob, and oversized all expose append/sync/sync_all/prune/rewind/rewind_section/clear on &mut self (backed by the shared manager.rs, which would flip to internal like contiguous's Inner).
2. Archive — the Archive trait itself (put, put_sync, put_start_sync, put_multi*, sync, start_sync all take &mut self in archive/mod.rs), plus both impls: archive::immutable and archive::prunable (which also has an inherent prune(&mut self)).
3. Cache — put, put_sync, prune, sync (cache/storage.rs).
4. Ordinal — put, prune, sync (ordinal/storage.rs).
5. Freezer — put and sync take &mut self (destroy/close already consume) (freezer/storage.rs).
6. Metadata — put_sync, upsert_sync, sync (metadata/storage.rs). Its non-async mutators (put, remove) are in-memory only and can stay borrowed.
7. Bitmap (authenticated) — one async mutator left: write_pruned(&mut self) (bitmap/authenticated.rs:386); the rest of its mutation is synchronous in-memory.

danlaine:
Converts the remaining `async` mutable methods in `commonware-storage` to consume `self` and return it on success, and threads the ownership through all consumers. Completes #4246.

## Contract

Async mutators take `self` and return `Self` (outputs flattened into tuples). An `Err` or a dropped future destroys the handle: recovery is re-initialization, making cancellation equivalent to a crash. `&mut self` remains only for in-memory-only mutation. Leaf stores are pointer-sized handles around `Box<Inner>`, so by-value threading is cheap.

## Design

- Archive and cache puts below the prune floor are silently satisfied: post-prune reads already report absence, so storing vs. dropping is indistinguishable. The `AlreadyPrunedTo*` error variants and `pruned()` pre-checks are deleted. Journal appends stay strict, since they return positions callers persist.
- `replay` consumes the journal into an owned `Replay` reader: `next()` yields items, `finish()` returns the journal only when fully drained without error (`Error::ReplayFailed` otherwise), and dropping mid-replay destroys the journal. A `next` future dropped during a tail repair poisons the replay (`Error::ReplayInterrupted`), routing recovery through re-initialization.
- New `commonware_utils::futures::{rebind, rebind_entry}` thread consuming mutations through `Option`/`BTreeMap` slots (restored on `Ok`, absent on `Err`), replacing the take/reinsert dances in consumers.

## Behavior changes

- Below-floor archive/cache puts succeed silently instead of erroring.
- Journal prune failures in aggregation/ordered_broadcast now panic (previously `let _ =`).
- State-sync persists the `InProgress` floor before marshal floor resolution, so a crash during resolution resumes state sync on restart.
- Dead public error variants deleted: aggregation `NotSigner`; ordered_broadcast `AppVerified*`, `ContextSequencer`, `ContextHeight`, `MissingCertificate`, `NotSigner`.

## Stability

Affected modules are BETA (storage, marshal, simplex) or ALPHA (aggregation, ordered_broadcast). Source-level API break only — no wire or storage format changes, so no migration path is required.

https://claude.ai/code/session_01HcS6eqStPLuLqf3ef6oubn


consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -271,25 +275,28 @@ impl<
     /// a restart. Deferring syncs to this boundary (rather than syncing after
     /// each append) coalesces all appends in the same loop iteration into a
     /// single sync.
-    async fn sync_journal(&mut self, view: View) {
+    async fn sync_journal(mut self, view: View) -> Self {
```
patrick-ogrady, line 275: I think this is a good example of a function not worth converting (private function that can't be used incorrectly by the caller)?

consensus/src/marshal/standard/mod.rs
```diff
@@ -6417,8 +6422,9 @@ mod tests {
             self.inner.get(id).await
         }
 
-        async fn prune(&mut self, min: Height) -> Result<(), Self::Error> {
-            self.inner.prune(min).await
+        async fn prune(mut self, min: Height) -> Result<Self, Self::Error> {
```
patrick-ogrady, line 6846: like consensus, another case where it can't be called incorrectly?

consensus/src/simplex/actors/voter/actor.rs
danlaine: I'm inclined to agree with claude here. Seems reasonable to me to have this remain `mut self -> Self`.

>Where Patrick has a real point
>
>He's right that today nothing can drop this future while the actor survives. Its only caller is another consuming handler, whose caller is another one, all the way up to run(mut self) — the actor lives inside that whole future stack, and select_loop! never cancels arm bodies. So if anything in that stack is dropped, the actor is dropped too. Today, even the &mut version couldn't actually poison anything.
>
>Why I'd still keep the conversion
>
>That "today" is doing all the work. The &mut version's safety depends on facts about every caller: that all of them hold the actor by value, and that none of them ever gets raced or spawned. The function itself looks fine; the danger only appears if someone, later, adds a timeout around a handler or moves this call into a spawned task. When they do, the &mut version becomes a real poison bug and the compiler says nothing.
>
>The consuming version doesn't depend on any of that. Whoever calls it must give up the actor and take it back — you can't call it from a context that keeps the actor alive separately; borrowck won't let you. The safety is visible in the signature instead of resting on an audit of the call tree that has to stay true forever.

consensus/src/aggregation/engine.rs
```diff
@@ -394,8 +391,9 @@ impl<
                 };
 
                 // Handle the ack
-                if let Err(err) = self.handle_ack(&ack).await {
-                    debug!(?err, ?sender, "ack handle failed");
```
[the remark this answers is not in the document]
danlaine: This is not a behavior change. On origin/main these same rejections already counted as Failure: handle_ack returned Err for unknown scheme (self.scheme(ack.epoch)?), non-pending height (Error::AckHeight), and digest mismatch (Error::AckDigest), and the receive loop's `if let Err(err) = self.handle_ack(&ack).await { guard.set(Status::Failure); continue; }` recorded them as failures — exactly what the !accepted branch does now. Duplicates returned Ok before and accepted = true now, so they still record Success. The metric outcome is identical for every case; only the debug log moved inside handle_ack.

storage/src/queue/storage.rs
```diff
@@ -71,6 +71,9 @@ pub struct Config<C> {
 /// On restart, `ack_floor` is set to the journal's pruning boundary.
 /// Items that were pruned are gone; everything else is re-delivered.
 /// Applications must handle duplicates (idempotent processing).
+///
+/// Storage-mutating functions consume the queue and return it only on success: an error (or a
+/// dropped future) destroys the handle, and recovery is re-initialization.
```
patrick-ogrady: We don't need this: `, and recovery is re-initialization`?

storage/src/journal/segmented/oversized.rs
```diff
@@ -106,11 +106,27 @@ pub struct Config<C> {
 ///
 /// Combines a fixed-size index journal with glob storage for variable-length values.
 /// Provides coordinated operations and crash recovery.
+///
+/// Mutating functions consume the journal and return it only on success: an error (or a dropped
+/// future) destroys the handle, and recovery is re-initialization. Mutations on pruned
```
patrick-ogrady: Another case of `, and recovery is re-initialization`

storage/src/journal/segmented/fixed.rs
```diff
@@ -351,80 +313,271 @@ impl<E: Storage + Metrics, A: CodecFixedShared> Journal<E, A> {
...
+/// Like
+/// [sqlite](https://github.com/sqlite/sqlite/blob/8658a8df59f00ec8fcfea336a2a6a4b5ef79d2ee/src/wal.c#L1504-L1505)
+/// and
+/// [rocksdb](https://github.com/facebook/rocksdb/blob/0c533e61bc6d89fdf1295e8e0bcee4edb3aef401/include/rocksdb/options.h#L441-L445),
+/// the first invalid data read will be considered the new end of the journal (and the
+/// underlying [Blob] will be truncated to the last valid item). Repair occurs during
+/// init by checking each blob's size.
+///
+/// Mutating functions consume the journal and return it only on success: an error (or a dropped
+/// future) destroys the handle, and recovery is re-initialization. Mutations on pruned
```
patrick-ogrady: , and recovery is re-initialization

consensus/src/marshal/store.rs
```diff
@@ -10,7 +10,10 @@ use commonware_storage::{
 use std::{error::Error, future::Future};
 
 /// Durable store for [Finalizations](Finalization) keyed by height and block digest.
-pub trait Certificates: Send + Sync + 'static {
+///
+/// Mutating functions consume the store and return it only on success: an error (or a
+/// dropped future) destroys the handle, and recovery is re-initialization.
```
patrick-ogrady: and recovery is re-initialization.

consensus/src/marshal/store.rs
```diff
@@ -38,43 +41,48 @@ pub trait Certificates: Send + Sync + 'static {
...
 
     /// Request that all buffered writes are flushed to durable storage.
     ///
     /// The returned handle completes once every write accepted before this call is durable,
     /// including writes covered by a sync that is still in flight. Implementations without a
     /// non-blocking sync path may complete the sync before returning an already-finished handle.
-    fn start_sync(&mut self) -> impl Future<Output = Result<Handle<()>, Self::Error>> + Send {
+    ///
+    /// An error reported by the returned [Handle] is fatal to the store: the caller must
+    /// stop using the returned store and recover by re-initializing.
```
patrick-ogrady: and recover by re-initializing

storage/src/freezer/storage.rs
```diff
@@ -1139,19 +1127,102 @@ impl<E: BufferPooler + Context, K: Array, V: CodecShared> Freezer<E, K, V> {
 
         Ok(())
     }
+}
+
+/// Implementation of [Freezer].
+///
+/// Mutating functions consume the freezer and return it only on success: an error (or a dropped
+/// future) destroys the handle, and recovery is re-initialization.
```
patrick-ogrady: , and recovery is re-initialization

glue/src/dkg/reshare/store.rs
```diff
@@ -158,7 +158,8 @@ where
     P: PublicKey,
 {
     secret_store: SS,
-    events: Journal<E, Event<V, P>>,
+    /// Taken while a consuming journal mutation is in flight; always restored before return.
```
patrick-ogrady: remove: // Taken while a consuming journal mutation is in flight; always restored before return.

patrick-ogrady (commented):
remove comment slop

patrick-ogrady (approved):

PR #4280, opened 2026-07-17
[consensus] Track batcher vote state per kind with a Certification enum

danlaine:
## Motivation

The batcher tracks each vote kind's lifecycle across two structs: `Verifier` holds per-kind `pending_*`/`verified_*` buffers, while `Round` holds `notarized`/`nullified`/`finalized` bools that gate verification and duplicate-certificate checks. The latch and the pipeline it terminates are kept coherent only by call-site discipline, and nothing tells the verifier when a kind is done: votes arriving after certification are buffered forever, and the buffers are only freed when the round is pruned. Callers probe readiness before acting (`ready_notarizes()` then `verify_notarizes()`), splitting one operation across two calls at every site.

## Changes

**`Certification<V>`** (`batcher/verifier.rs`): each kind's pipeline is now one self-contained value, its configuration beside its state,

```rust
struct Certification<V> {
    quorum: usize,
    batchable: bool,
    state: State<V>,
}
enum State<V> {
    Incomplete { pending: Vec<V>, verified: Vec<V> },
    Complete,
}
```

replacing the six buffer fields, the `Verifier`'s quorum field, and the three `Round` bools. Its methods are the pipeline's lifecycle: `add` buffers a vote; `try_verify(f)` — if a batch is worth running — takes the buffers, awaits the caller's verification future, and restores the result (the gate and the take/restore pairing cannot be separated or half-done); `try_complete` surrenders a verified quorum for certificate recovery and completes; and `complete` completes directly when a certificate is observed from the network. Completing drops the kind's buffers and all subsequent votes, so verifying or buffering votes after certification is unrepresentable rather than guarded. Kinds certify independently (a view can legitimately certify both a notarization and a nullification). The quorum-readiness arithmetic previously triplicated across three `ready_*` methods lives once, as `try_verify`'s internal gate (`should_verify`).

**Certificate assembly moves into `Verifier`**: `try_construct_*` previously lived on `Round`, which pulled the verifier's verified votes (`take_verified_*`) and signing scheme (`scheme()`) across the module boundary to assemble certificates, then reported completion back. The verifier now finishes its own pipeline — votes in, certificates out — and those accessor methods are gone. The verifier stores its round id and creates verification/construction spans only when work actually happens (readiness probes are never traced, matching main).

**Check-then-act becomes try**: the `ready_*`/`verify_*` and `has_*`-guarded `try_construct_*` pairs collapse into self-gating methods that report via `Option`:

- `Verifier::try_verify_{notarizes,nullifies,finalizes}` verify a batch if worthwhile (leader proposal known where applicable, buffers warrant a batch) and return the batch size and invalid signers; the `ready_*` methods are gone.
- `Round::try_verify` chains them in priority order (notarizes, nullifies, finalizes), collapsing the actor's readiness cascade to one call.
- `Verifier::try_construct_certificate` returns the next constructible `Certificate` (notarization, then nullification, then finalization), consuming its verified quorum — one kind-erased method instead of three per-kind ones, since the only caller wants a `Certificate` anyway. The actor drains it (through `Round`) with one `while let` loop and forwards each certificate, replacing three per-kind blocks.
- `Round::forward_proposal` becomes `try_forward_proposal` — same gated, at-most-once semantics, now named for them — and stops sharing a name with the actor's unconditional network forward.

**The leader lifecycle becomes one machine**: `leader: Option<Participant>` + `leader_proposal: Option<Proposal>` (whose "proposal implies leader" invariant was held by convention and unwrapped on) collapse into `enum Leader { Unknown, Known(Participant), Proposed { leader, proposal } }`. The vote-filtering side effect is owned by the one named transition that causes it (`try_learn_proposal`, `Known → Proposed`, which learns the proposal from a notarize of the leader's), the verification gates are a variant match, and a proposal without a leader is unrepresentable.

**Leader-proposal discovery moves to the tracker**: `set_leader` previously found the leader's proposal by scanning the verifier's buffered notarizes; since certification now drops those buffers, `Round::set_leader` instead reads the leader's notarize from `VoteTracker` (which retains one vote per signer for the round's lifetime) and passes it to `Verifier::set_leader`. The tracker is populated before every `verifier.add`, holds at most one notarize per signer, and is never filtered before the leader is known, so the lookup returns exactly what the buffer scan would have in every message ordering — including after the buffers drop. Late discovery (leader's vote arriving after `set_leader`) is unchanged: `Verifier::add` still adopts the proposal before the vote reaches `Certification::add`.

**`Kind`** (`types.rs`): `Certificate::kind()` now returns an enum (`Notarization | Nullification | Finalization`) instead of a trace string (a `Display` impl keeps the stable trace values). `Round`'s six `has_*`/`mark_*` methods collapse to `has_certificate(kind)`/`certify(kind)`, and the actor's per-variant duplicate check becomes a single `round.has_certificate(kind)`.

There are no protocol-visible behavior changes: post-certification votes were dead weight nothing ever read (verification never re-fires for a certified kind, and conflict detection reads `VoteTracker`, untouched), so dropping them and freeing the buffers at certification is observable only as reclaimed memory. The actor's per-kind construction logs unify into one message with a `kind` field.

### DIFF
--- a/consensus/src/simplex/actors/batcher/actor.rs
+++ b/consensus/src/simplex/actors/batcher/actor.rs
@@ -635,8 +635,9 @@ where
                     // Batch verify votes if ready
                     let timer = self.verify_latency.timer(self.context.as_ref());
                     if let Some((batch, failed)) = round
-                    .try_verify(self.context.as_mut(), &self.strategy)
-                    .await {
+                        .try_verify(self.context.as_mut(), &self.strategy)
+                        .await
+                    {
                         timer.observe(self.context.as_ref());
 
                         // Process verified votes.
@@ -665,11 +666,18 @@ where
                     // Construct and forward every certificate with a verified quorum
                     while let Some(certificate) = self
                         .recover_latency
-                        .time_some(self.context.as_ref(), round.try_construct_certificate(&self.strategy))
+                        .time_some(
+                            self.context.as_ref(),
+                            round.try_construct_certificate(&self.strategy),
+                        )
                         .await
                     {
                         let kind = certificate.kind();
-                        debug!(%updated_view, %kind, "constructed certificate, forwarding to voter");
+                        debug!(
+                            %updated_view,
+                            %kind,
+                            "constructed certificate, forwarding to voter"
+                        );
                         voter.recovered(certificate);
                     }
                 }
--- a/consensus/src/simplex/actors/batcher/actor.rs
+++ b/consensus/src/simplex/actors/batcher/actor.rs
@@ -419,13 +419,9 @@ where
                 // Skip certificates we already have for the view
                 let kind = message.kind();
                 let round = work.get(&view);
-                let duplicate = round.is_some_and(|round| match &message {
-                    Certificate::Notarization(_) => round.has_notarization(),
-                    Certificate::Nullification(_) => round.has_nullification(),
-                    Certificate::Finalization(_) => round.has_finalization(),
-                });
+                let duplicate = round.is_some_and(|round| round.has_certificate(kind));
                 if duplicate {
-                    trace!(%view, kind, "skipping duplicate certificate");
+                    trace!(%view, %kind, "skipping duplicate certificate");
                     continue;
                 }
 
@@ -435,7 +431,7 @@ where
                 let span = info_span!(
                     parent: parent,
                     "simplex.batcher.verify_certificate",
-                    kind,
+                    %kind,
                     epoch = self.epoch.traced(),
                     view = view.traced()
                 );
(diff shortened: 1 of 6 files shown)


consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -456,7 +452,7 @@ where
                         // Store and forward to voter
                         work.entry(view)
                             .or_insert_with(|| self.new_round(view))
-                            .mark_notarized();
+                            .certify(kind);
```
patrick-ogrady: `ceritfy` is an overloaded word in this section of the code...can we use `mark` or something else?

danlaine: Renamed `mark_certified`

patrick-ogrady: the "certify*" part is the issue

patrick-ogrady: that is what we used to describe a very particular phase of consensus

danlaine: Renamed `record_certificate`

patrick-ogrady (approved):

PR #4290, opened 2026-07-20
[runtime] single-threaded io_uring runtime

andresilva:
This PR introduces a native io_uring runtime as an alternative to the tokio runtime. Previously, io_uring was only an I/O backend driven by tokio. It now implements the commonware runtime traits directly, making io_uring a complete runtime option rather than an I/O backend.

Tokio does not provide native io_uring support, and integrating io_uring into its executor cannot fully exploit io_uring's design. Our previous approach also required cross-thread coordination for every I/O request and completion. Owning task execution and the ring on the same worker enables direct submission and completion handling, removing that coordination from the I/O path.

This is the first step in a series of PRs toward the runtime's final architecture: a multi-threaded, work-stealing scheduler with affinity for ring-bound I/O operations. The runtime is currently single-threaded per worker, with each worker owning one ring. Separating task scheduling from the io_uring driver establishes the boundary needed for future work stealing while keeping resources and in-flight operations on their owning worker.


.github/workflows/benchmark.yml
```diff
@@ -42,9 +42,6 @@ jobs:
           - package: commonware-storage
             cargo_flags: "--features test-traits"
             benchmark_name: "commonware-storage"
-          - package: commonware-storage
-            cargo_flags: "--features commonware-runtime/iouring-storage,test-traits"
-            benchmark_name: "commonware-storage --features"
```
andresilva, line 47: Need to add a feature flag to storage benches to use iouring runtime.

andresilva:
Replaced by #4674.

PR #4291, opened 2026-07-20
[runtime] use rstest for generic runtime tests

andresilva:


runtime/src/lib.rs
```diff
@@ -3145,582 +3168,206 @@ mod tests {
...
+                    .config()
+                    .size_classes()
+                    .all(|class| class.max_buffers.get() == expected_storage_max_per_class)
+            );
+        });
     }
 
     #[test]
-    fn test_deterministic_spawn_after_abort() {
+    fn test_deterministic_resolver() {
```
patrick-ogrady: These deterministic-only tests should probably be moved to deterministic?

andresilva: Will do the same for tokio.

patrick-ogrady (approved):
massive skill issue lol

PR #4292, opened 2026-07-20
[storage] Reconstruct activity bitmap shares on the snapshot workers

roberto-bayardo:
## Summary

The parallel snapshot build reconstructed the activity bitmap as a serial post-join phase: after installing every worker's partition range, the init task walked the full index and set one bit per active key (~0.9s at 50M keys), with every worker idle. Each worker now walks its own partition range before returning — in parallel with the other workers, over data it just built and still has cache-warm — setting the bits of its active locations in a single shared replay-sized word array; after the join, the init task reassembles the words into the bitmap one chunk per word.

- `PartitionRange` gains `for_each_value`; both partitioned variants implement it over their range views.
- `build_snapshot_worker` records its range's active locations in the shared words (relaxed `fetch_or`; the join is the synchronization point) and returns the range's active-key count alongside the populated range.
- The full-index walk (`Partitioned::for_each_value`) had no remaining callers and is removed; the ordered variant's range view delegates to the index's inherent value walk.

The result is identical to the serial build's per-op push-then-clear: each active key holds exactly one location in exactly one worker's range, so every bit has a single writer (workers overlap only at word granularity, which the atomic or absorbs), and the words end with exactly the bits of each active key's current location set (plus the final commit, set by the init task as before). Sharing one allocation keeps peak bitmap memory at one replay-sized map regardless of worker count, and the words are u64s in `BitMap` chunk order, so reassembly is one `push_chunk` per word.

## Measurement

`init_scale` on a 50M-keyspace / 250M-update database (replay region 64.6M ops), AMD EPYC 9354P, warm OS cache, `init_concurrency=8`, interleaved rounds against `main`. Phase timers added for the measurement only; rounds with matched routing times shown (routing swings ±1s run-to-run and is untouched by this change):

| phase | main | this PR |
|---|---|---|
| routing | 10.7s | 10.7s |
| worker drain | 0.36s | 0.45s |
| install + bitmap | **1.13s** | **0.15s** |
| total init | **13.2s** | **12.3s (−7%)** |

The bitmap walk moves from a ~0.9s serial phase into the workers (partially absorbed by the drain window) plus a ~0.15s reassembly.

Sharing one allocation also bounds peak transient bitmap memory. The prior per-worker bitmaps cost `(workers + 1) × replay` bits; the shared word array costs one replay-sized map regardless of worker count — ~65 MB → ~8 MB at `init_concurrency=8` on this database, and ~2.1 GB → ~8 MB at the 256-partition (P1) worker cap.

Both effects apply to the unordered partitioned index, which is also built in parallel. A serial → parallel sweep (same 50M/250M scale, `any::unordered::fixed` P=2) confirms it is a beneficial parallel path in its own right, scaling comparably to ordered:

| `init_concurrency` | uniform | zipf 1.0 |
|---|---|---|
| 1 (serial) | 47.2s (1.00×) | 29.9s (1.00×) |
| 2 | 32.9s (1.44×) | 20.8s (1.44×) |
| 3 | 15.8s (2.99×) | 9.7s (3.08×) |
| 5 | 13.1s (3.61×) | 8.2s (3.67×) |
| 8 | 12.0s (3.93×) | 7.1s (4.19×) |

The shared allocation applies to unordered identically (same peak-memory reduction). Its install+bitmap phase is already negligible (~8 ms), so for unordered the memory bound, not wall time, is the win.

## Validation

- Parallel-init equivalence tests compare the resulting state — including bitmap-derived roots — against the serial build across a concurrency sweep up to 300 workers: ordered at P1/P2/P3 (both `any` and `current`, plus the variable-payload variant) and unordered at P1 (both `any` and `current`; P2's 65,536 sub-indexes are too memory-heavy for the suite). Fresh-db, replay-failure, worker-failure, and empty-log cases exercise the edge paths.
- Full storage suite (2766 tests), glue tests, `check-stability`, and the wasm32 storage build are clean.

### DIFF
--- a/storage/src/index/partitioned/ordered/mod.rs
+++ b/storage/src/index/partitioned/ordered/mod.rs
@@ -139,8 +139,7 @@ impl<T: Translator, V: Send + Sync, const P: usize> Index<T, V, P> {
         index
     }
 
-    /// Visit every value held across all partitions (inline and spilled), in unspecified
-    /// order. Shared by a build worker's range view over its local slots.
+    /// Visit every value held across all partitions (inline and spilled), in unspecified order.
     #[commonware_macros::stability(ALPHA)]
     fn for_each_value_impl(&self, mut f: impl FnMut(&V)) {
         for (p, partition) in self.partitions.iter().enumerate() {
--- a/storage/src/index/partitioned/ordered/mod.rs
+++ b/storage/src/index/partitioned/ordered/mod.rs
@@ -1098,6 +1098,36 @@ mod tests {
         });
     }
 
+    /// A worker's value walk must visit every value it holds exactly once, whichever
+    /// representation each partition uses (inline sorted arrays or the spilled side-table),
+    /// since the snapshot build derives the activity bitmap and active-key counts from it.
+    #[test_traced]
+    fn test_range_for_each_value_visits_all_values_once() {
+        deterministic::Runner::default().start(|context| async move {
+            let full = new_index_spilling(context.child("full"));
+
+            // Two distinct keys spill partition 0x80 (threshold 2), then a translated-key
+            // collision appends a second value to a spilled run. Partition 0x81 holds one key
+            // and stays inline.
+            let mut worker = full.new_range(0x80, 2);
+            assert!(worker.get_mut_or_insert(&[0x80, 0x01], 1).is_none());
+            assert!(worker.get_mut_or_insert(&[0x80, 0x02, 0xAA], 2).is_none());
+            assert!(worker.get_mut_or_insert(&[0x80, 0x02, 0xBB], 3).is_some());
+            {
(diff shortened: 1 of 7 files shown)


storage/src/index/partitioned/ordered/mod.rs
```diff
@@ -139,6 +139,23 @@ impl<T: Translator, V: Send + Sync, const P: usize> Index<T, V, P> {
         index
     }
 
+    /// Visit every value held across all partitions (inline and spilled), in unspecified
+    /// order. Shared by the full index and a build worker's range view.
```
patrick-ogrady: Remove this `Shared by the full index and a build worker's range view.`

roberto-bayardo: Done in f93ec2774.

patrick-ogrady (approved):

PR #4295, opened 2026-07-20
[upgrades] Add `commonware-upgrades`

### LINKED ISSUE #4287: [workspace] Add `commonware-hardforks`
## Overview

Add a new `commonware-hardforks` crate with shared feature activation labels, for use in introducing breaking changes on code that is in `BETA` stability or higher. Hardforks activate on `Epoch` boundaries. Each chain will have different epochs in which they want to activate features, they must be configurable. Forks also have a hierarchy: activating a hardfork implies all previous hardforks are also enabled.

clabby:
## Overview

Adds a new `commonware-upgrades` crate with a canonical upgrades hierarchy and user-configurable activation schedules. Schedules support validation, activation queries, human-readable variant names, and deterministic commitments for efficient comparison while remaining compatible with `no_std`.

This crate will be used to introduce breaking changes to `BETA` and higher code throughout the codebase. Components will thread through a `Schedule`, using it to activate upgrade-specific features at runtime.

closes https://github.com/commonwarexyz/monorepo/issues/4287


clabby:
related: #4273

hardforks/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "commonware-hardforks"
+edition.workspace = true
+publish = true
+version.workspace = true
+license.workspace = true
+description = "Coordinate protocol upgrades at epoch boundaries."
```
patrick-ogrady: nit: "synchronized boundaries"?

hardforks/src/lib.rs
```diff
@@ -0,0 +1,288 @@
+#![doc = include_str!("../README.md")]
+#![doc(
+    html_logo_url = "https://commonware.xyz/imgs/rustdoc_logo.svg",
+    html_favicon_url = "https://commonware.xyz/favicon.ico"
+)]
+#![no_std]
+
+commonware_macros::stability_scope!(BETA {
```
patrick-ogrady: start in alpha?

README.md
```diff
@@ -22,6 +22,7 @@ _Primitives are designed for deployment in adversarial environments. If you find
 * [cryptography](./cryptography/README.md): Generate keys, sign arbitrary messages, and deterministically verify signatures.
 * [deployer](./deployer/README.md): Deploy infrastructure across cloud providers.
 * [glue](./glue/README.md): Default constructions that span multiple primitives.
+* [hardforks](./hardforks/README.md): Coordinate protocol upgrades at epoch boundaries.
```
patrick-ogrady: i suppose we could just call it "upgrades"?

hardforks/src/lib.rs
clabby: Done, though we'll pretty quickly have to bump it to `BETA` for use where we need it?

patrick-ogrady: yeah, I still think more "correct"

upgrades/src/lib.rs
```diff
@@ -0,0 +1,288 @@
...
+            }
+        }
+    }
+
+    impl<E: Ord> Schedule<E> {
+        fn new(activations: impl IntoIterator<Item = (Upgrade, E)>) -> Result<Self, Error> {
+            let mut activations: Vec<_> = activations.into_iter().collect();
+            activations.sort_unstable_by_key(|(upgrade, _)| *upgrade);
+
+            for pair in activations.windows(2) {
```
patrick-ogrady: I think there is a built-in for this (`is_sorted` or something?)

upgrades/README.md
```diff
@@ -0,0 +1,29 @@
+# commonware-upgrades
+
+[![Crates.io](https://img.shields.io/crates/v/commonware-upgrades.svg)](https://crates.io/crates/commonware-upgrades)
+[![Docs.rs](https://docs.rs/commonware-upgrades/badge.svg)](https://docs.rs/commonware-upgrades)
+
+Coordinate upgrades at synchronized boundaries.
+
+Commonware defines one ordered set of upgrades. Applications choose the epoch at which each
```
patrick-ogrady, line 8: Alright, so this is more for our upgrades than a generic upgrade manager?

clabby: Yep. I think for the case of managing breaking upgrades in the commonware library, it's best to have a scoped definition. Applications can define their own upgrade routines, and they may upgrade at a different cadence.

upgrades/src/lib.rs
```diff
@@ -0,0 +1,284 @@
...
+        /// epoch. Returns `None` when neither it nor a descendant is configured.
+        pub fn activation_epoch(&self, upgrade: Upgrade) -> Option<E> {
+            let index = self
+                .activations
+                .partition_point(|(candidate, _)| *candidate < upgrade);
+            self.activations.get(index).map(|(_, epoch)| *epoch)
+        }
+
+        /// Returns whether `upgrade` is active at `epoch`.
+        pub fn is_active(&self, upgrade: Upgrade, epoch: E) -> bool {
```
patrick-ogrady, line 128: in practice, will we just say "entered epoch" and allow this to return valid? or do we need to stitch epoch everywhere (say `stream`?)

patrick-ogrady: in any case, for some areas we may want to allow some sort of "overlap" (for x epochs, both paths work to account for synchrony?)

clabby: An understanding of the active epoch would need to be threaded through anything that wanted to use this, yep.

As for the transitional period that may be needed for some features, I don't think that this library should have an opinion. That is usually best left to the place where the new feature is implemented. The schedule answers "is this feature set active in this epoch?" and the implementation of said feature can be concerned with how to safely activate a transition.

PR #4299, opened 2026-07-20
[consensus/simplex] Stable Leader Feedback (Part 2)

patrick-ogrady:
Related: #3352

- **Add `Viewport`**: replaces `interesting()` (six positional arguments, a mode flag, and a `term_length` that was dead whenever `allow_unbounded_future` was set) with a named policy struct shared by the batcher and voter: `retains` (the retention floor), `admits_vote` (retention plus the term-aware future bound from `View::admits`), and `admits_certificate` (retention only: certificates are self-certifying and let a lagging node fast-forward). All view-window checks in both actors route through it, so the window semantics live in one place.
- **Restore retention below the tip**: votes are tracked and reported down to `activity_timeout` views below the last finalized view again (matching `main`), rather than dropped at the tip — preserving straggler participation reports and sub-tip equivocation evidence (`ConflictingNotarize`/`NullifyFinalize`) for reporter consumers. The batcher regains `activity_timeout`, and docs that described the tip-drop (`Activity`, protocol overview, `Config::reporter`) are updated to the restored boundary.
- **Stop future-bounding our own votes**: the voter constructs votes before sending the update that advances the batcher's view, so after a certificate jump a locally constructed vote can be ahead of it — `Constructed` now gates on retention only (admission bounds exist for untrusted network input), pinned by `test_constructed_votes_are_not_future_bounded`.
- **Simplify the fast-timeout quorum check**: instead of recording our own activity alongside observed peers, the batcher caches `required_active` (quorum minus one when we are a participant, since we never observe our own messages and are live by construction).
- **Make `Terms` valid by construction**: `Terms` is now a struct with private fields built only by `Terms::rotating()` (length one, no stall timeout) and `Terms::stable(length, stall_timeout)` (asserts a length greater than 1 and a non-zero timeout), so `Terms::length()` returns `TermLength::ONE` if and only if leaders rotate every view — for any `Elector` impl, not just configurations that pass the engine's checks. The engine keeps only the relational assert that needs its config (`stall_timeout > certification_timeout`), and the single-view regime is named rotating (previously `Terms::Single`) to pair with stable as leader-tenure regimes.
- **Document the resolver fetch strategy**: the resolver module docs now diagram term-anchor requests, the same-term coverage rule, and why a mid-term floor raise pulls the fetch cursor back (including that a rescan re-issues still-pending anchors, which the resolver engine deduplicates), and the `State` docs use the same "anchor" vocabulary. Also documents why `add_nullification` needs no finalization guard (earlier-term sub-tip nullifications no-op by view monotonicity; same-term pairs are unforgeable per Same-Term Vote Safety).
- **Tests**: `test_viewport` covers the policy directly; new `votes_below_finalized_within_activity_window_reported` pins that in-window votes at or below the tip are reported while below-window votes are not; the startup-floor test is adapted to the activity-window boundary.

Also renames `is_interesting_*` → `admits_*` to match `Viewport`.

### DIFF
--- a/consensus/src/simplex/actors/batcher/actor.rs
+++ b/consensus/src/simplex/actors/batcher/actor.rs
@@ -129,7 +129,10 @@ where
         let (sender, receiver) = mailbox::new(context.child("mailbox"), cfg.mailbox_size);
         let mut required_active = participants.quorum::<N3f1>() as usize;
         if scheme.me().is_some() {
-            required_active -= 1;
+            // We are live by construction (we never observe our own messages).
+            required_active = required_active
+                .checked_sub(1)
+                .expect("quorum is never zero");
         }
         (
             Self {
--- a/consensus/src/simplex/actors/batcher/actor.rs
+++ b/consensus/src/simplex/actors/batcher/actor.rs
@@ -73,6 +73,12 @@ where
     /// participant. `None` means no activity has been observed.
     last_activity: Vec<Option<SystemTime>>,
 
+    /// Number of observed participants that must be recently active for the
+    /// network to be considered responsive (see [Self::is_active]). We never
+    /// observe our own messages, so when we are a participant we count
+    /// ourselves as live by construction.
+    required_active: usize,
+
     mailbox_receiver: mailbox::Receiver<Message<S, D>>,
 
     added: Counter,
@@ -121,11 +127,16 @@ where
             Buckets::CRYPTOGRAPHY,
         );
         let (sender, receiver) = mailbox::new(context.child("mailbox"), cfg.mailbox_size);
+        let mut required_active = participants.quorum::<N3f1>() as usize;
+        if scheme.me().is_some() {
+            required_active -= 1;
+        }
         (
             Self {
                 context: ContextCell::new(context),
 
                 last_activity: vec![None; participants.len()],
+                required_active,
                 scheme,
 
                 blocker: cfg.blocker,
@@ -191,10 +202,10 @@ where
         let recent =
             |activity: &Option<SystemTime>| activity.is_some_and(|activity| activity >= min_time);
 
-        // If there is not a quorum of recently active participants, then we "fail-open" since we
-        // know the network is not expected to be responsive.
+        // If fewer than the required number of participants are recently active, we "fail-open"
+        // since we know the network is not expected to be responsive.
         let active = self.last_activity.iter().filter(|a| recent(a)).count();
-        if active < self.scheme.participants().quorum::<N3f1>() as usize {
+        if active < self.required_active {
             return true;
         }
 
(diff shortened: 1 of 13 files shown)


consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -430,9 +446,8 @@ where
                 let view = message.view();
                 self.record_peer_activity(&sender);
 
-                // Ignore certificates below the highest finalized view since they aren't useful.
-                // Allow certificates from arbitrarily-future views since they advance our view.
-                if view <= finalized {
+                // Skip certificates outside the tracked window
+                if !self.window(finalized, current.view).admits_certificate(view) {
```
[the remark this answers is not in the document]
patrick-ogrady: This pair cannot exist under the fault model. For the jump to matter, `next_term_start(v) > current`, and `current >= F + 1` once `finalization@F` is processed:

- `v` in an **earlier term** than `F`: `next_term_start(v) <= term_start(F) <= F < current`, and `enter_view` only advances — no-op.
- `v` in the **same term** as `F` (the only jumping case): `nullification@v <= F` coexisting with `finalization@F` is unforgeable under `<= f` Byzantine — the 2f+1 nullify and 2f+1 finalize quorums intersect in f+1 honest validators, each of which the finalize-withholding gate prevents from contributing to both (see [Same-Term Vote Safety](https://github.com/commonwarexyz/monorepo/blob/b2743717a/consensus/src/simplex/mod.rs)).

Certificates are also quorum-signature-verified by the batcher before forwarding, so "a peer can push" requires forging a 2f+1 certificate. Note `main` admits the same in-window sub-tip certificates (`interesting(..., allow_future = true)` bounds only by `min_active`) — the jump mechanics are terms-only and land exclusively on the impossible pair. Documented at `add_nullification` in b2743717a.

consensus/src/simplex/actors/resolver/state.rs
BrendanChou, line 1: You defined "anchor" in `mod.rs`, then avoided using it here in `state.rs`. I think we should continue using it instead of paraphrasing it.

Otherwise decide to drop it as a term altogether and remove it everywhere in `consensus/simplex`

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -120,18 +127,24 @@ where
             Buckets::CRYPTOGRAPHY,
         );
         let (sender, receiver) = mailbox::new(context.child("mailbox"), cfg.mailbox_size);
+        let mut required_active = participants.quorum::<N3f1>() as usize;
+        if scheme.me().is_some() {
+            required_active -= 1;
```
BrendanChou: should we do `checked_sub(1).unwrap()`? A short comment here may also be useful

consensus/src/simplex/mod.rs
```diff
@@ -417,6 +418,51 @@ cfg_if::cfg_if! {
         pub use engine::Engine;
         mod metrics;
 
+        /// The window of views an actor tracks, bounded below by retention
+        /// and above by admission policy.
+        #[derive(Clone, Copy)]
+        pub(crate) struct Window {
```
BrendanChou: I wonder if `ActivityWindow` or `TrackedWindow` or `ViewWindow` may be better? Wordier, but less nebulous

BrendanChou (commented):
Looks good overall

PR #1804, opened 2025-10-04
[codec] Add Zeroed type

fedemagnani:
Introduced the `Zeroed` codec type: a useful type for encoding/decoding zero-padded data of a specific length without any memory allocation.

When reading from buffer, it avoids heap-allocating a vector by "peeking" the input buffer via `.chunk()[..len]`: the buffer is advanced only if the given chunk has all zeros

Closes #1703

PR #4302, opened 2026-07-21
[codec] Add Zeroed type

### LINKED ISSUE #1703: [codec] Add `Zeroed` type
A type that represents a sequence of zeroed bytes. Similar to `[0; N]` but required to be all-zeroes. Doesn't actually need to store the array, could just be a single `usize` instead. This is useful for padding or zeroing out a buffer. Should return an error on `read_cfg(...)` if the bytes are not zeroed

```
pub struct Zeroed {
    n: usize,
}

impl Zeroed {
    pub fn new(n: usize) -> Self {
        Self { n }
    }
}

impl Read for Zeroed {
    type Cfg = usize;

    fn read_cfg(buf: &mut impl Buf, n: &Self::Cfg) -> Result<Self, Error> {
        at_least(buf, *n)?;
        let mut bytes = vec![0; *n];
        if bytes.iter().any(|&b| b != 0) {
            return Err(Error::Invalid("Zeroed", "bytes are not zero"));
        }
        Ok(Self { n: *n })
    }
}

impl Write for Zeroed {
        buf.put_bytes(0, self.n);
    }
}

impl EncodeSize for Zeroed {
        self.n
    }
}
```

aroralanuk:
> AI disclosure: This changes have been made with GPT-5.6 Sol (xHigh) but I have manually reviewed them and I take responsibility for its correctness.

## Summary
- add runtime-sized `Zeroed` codec type for zero-filled padding
- validate zero padding without allocation and add codec conformance coverage
- use `Zeroed` in QMDB Any fixed-operation encoding

## Testing
- `just pre-pr`
- `just test-conformance -p commonware-codec zeroed`
- `just test-conformance -p commonware-storage fixed_encoding_u64`

Closes #1703

### DIFF
--- a/codec/conformance.toml
+++ b/codec/conformance.toml
@@ -178,6 +178,10 @@ hash = "dcb27e675ccce4ba82dd6979036b04198235bec00d3cefbc8e25a639ab4a162e"
 n_cases = 65536
 hash = "edc22446bb2952609d0c8daccf2d22f8ad2b71eedfcf45296c3f4db49d78404a"
 
+["commonware_codec::types::zeroed::tests::conformance::CodecConformance<Zeroed>"]
+n_cases = 65536
+hash = "8b1f2c316a515b1bec229b1b824acdc0630e27c2430d33cb3c5049e38b188248"
+
 ["commonware_codec::varint::tests::conformance::CodecConformance<SInt<i128>>"]
 n_cases = 65536
 hash = "42320f03747b52912d8dbb129a5958fe00ebd434a3381e3aa7daca2f48981113"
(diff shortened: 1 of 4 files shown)


aroralanuk:
`Zeroed<const N: usize>` would have been my first pick too which we could have used like `Zeroed::<32>::read(buf)?;` but as pointed out here https://github.com/commonwarexyz/monorepo/pull/1804#discussion_r2412235162, the padding length will be often calculated from generic constants like `S::SIZE` which isn't allowed by stable Rust.

aroralanuk:
cc @clabby

PR #4318, opened 2026-07-22
[p2p] Hoist shared router into `authenticated`

danlaine:
The `discovery` and `lookup` networks each carried a private copy of the same message-routing machinery. This PR deduplicates it.

- **`router/`**: both variants' `actors/router/{actor,ingress,mod}.rs` are replaced by a single `authenticated/router/`, following the precedent of the already-shared `data.rs`, `relay.rs`, `mailbox.rs`, and `dialing.rs`. The shared module keeps the union of both variants' tests, and fixes discovery's `start()`/`run()` docs, which omitted that the actor also exits on shutdown (lookup's wording matches the `select_loop!` body).
- **`channels.rs`**: the duplicated `Sender`/`Receiver`/`Channels` types are hoisted, and the shared `Error { NetworkClosed }` moves into the file. Each variant re-exports them, so public paths (`discovery::Sender`, etc.) are unchanged.
- **`DATA_PREFIX`**: both variants defined `DATA_PREFIX = 0` and `MAX_PAYLOAD_DATA_OVERHEAD`, plus identical `encode_data` wrappers. The constants now live once in `data.rs` and the wrappers are gone; variant-specific prefixes (greeting/bitvec/peers, ping) stay in their variant's `types.rs`.

One deliberate API consequence: `discovery::{Sender, Receiver, Error}` and their `lookup::` twins are now the same nominal types rather than lookalikes. This bakes in the invariant that both wire protocols use discriminant 0 for data frames, and it means downstream code that implements a trait separately for both variants' types (e.g. `From<discovery::Error>` and `From<lookup::Error>` on the same error enum) will now fail with conflicting implementations — keep one impl.


p2p/src/authenticated/discovery/types.rs
```diff
@@ -24,16 +24,6 @@ pub enum Error {
...
-///
-/// The byte overhead is calculated as the sum of the following:
-/// - 1: Payload enum value
-/// - 10: Channel varint
-/// - 5: Message length varint (lengths longer than 32 bits are forbidden by the codec)
-pub const MAX_PAYLOAD_DATA_OVERHEAD: u32 = 1 + 10 + 5;
-
-/// Prefix byte used to identify a [Payload] with variant Data.
-const DATA_PREFIX: u8 = 0;
 /// Prefix byte used to identify a [Payload] with variant Greeting.
```
patrick-ogrady, line 26: I liked that this file was a comprehensive list of all prefixes...I think we should at least put a comment.

danlaine: Added alias here to `crate::authenticated::data::DATA_PREFIX`

p2p/src/authenticated/discovery/types.rs
```diff
@@ -24,16 +21,8 @@ pub enum Error {
...
-///
-/// The byte overhead is calculated as the sum of the following:
-/// - 1: Payload enum value
-/// - 10: Channel varint
-/// - 5: Message length varint (lengths longer than 32 bits are forbidden by the codec)
-pub const MAX_PAYLOAD_DATA_OVERHEAD: u32 = 1 + 10 + 5;
-
 /// Prefix byte used to identify a [Payload] with variant Data.
-const DATA_PREFIX: u8 = 0;
+const DATA_PREFIX: u8 = crate::authenticated::data::DATA_PREFIX;
```
patrick-ogrady: add a trailing // 0 and assert 0 in unit tests?

danlaine: done

patrick-ogrady (approved):
noice

PR #4323, opened 2026-07-23
[runtime] Rewrite partial pages in full on paged flush

patrick-ogrady:
## Summary

This PR cuts the I/O needed to durably flush paged writes without changing the on-disk format.

- Paged flushes rewrite an existing partial page with one whole-physical-page write, resubmitting the committed prefix and protected CRC slot byte-for-byte. When no earlier mutation needs a full barrier, the write can carry submitted-range durability.
- `Blob::write_at_with` is now the required write primitive with composable `SYNC` and `DONT_CACHE` options. `Blob::write_at` remains the default convenience, while `Blob::write_at_sync` is removed in favor of `WriteOptions::SYNC`.
- Tokio and io_uring batch up to 1,024 iovecs. Linux single-submission synchronized writes use `RWF_DSYNC`, while larger writes finish with one data-only sync. Paged writes request `RWF_DONTCACHE`, retry without it on `EOPNOTSUPP`, and remember that fallback per blob.
- The temporary end-to-end example is now part of the registered Criterion paged benchmark, covering 256-byte appends at 1 and 16 records per sync.

## Performance

Controlled A/B on Linux/aarch64 kernel 7.0 with ext4 against pre-PR base `96d88242`. Each of 10 order-alternated paired rounds measured 5,000 append+sync cycles after 64 warmups, with 256-byte records and a 16,336-byte writer buffer. Values are paired means.

| backend | throughput change | p50 sync-latency change |
| --- | ---: | ---: |
| Tokio | +41% | -29% |
| io_uring | +48% | -33% |

A separate trace with 64 warmups and 500 measured cycles reduced write and durability syscall totals from 1,200 to 568 for Tokio and from 1,198 to 566 for io_uring. `RWF_DONTCACHE` was supported, and `strace` confirmed the Tokio path combined it with `RWF_DSYNC`. The no-cache option remains a best-effort hint.


runtime/src/utils/buffer/paged/writer.rs
```diff
@@ -937,12 +819,11 @@ impl<B: Blob> Writer<B> {
...
     ///
-    /// A single physical write can be persisted with [`Blob::write_at_sync`]. If there
-    /// are earlier unsynced mutations, or if the flush emits multiple physical writes,
-    /// durability is completed with [`Blob::sync`].
+    /// The flush's write can be persisted with [`Blob::write_at_sync`]. If there are earlier
+    /// unsynced mutations, durability is completed with [`Blob::sync`].
     pub async fn sync(&mut self) -> Result<(), Error> {
         // Flush any buffered data, including any partial page.
-        // A single emitted write can be made durable directly during the flush.
+        // An emitted write is made durable directly during the flush.
```
roberto-bayardo: What's an "emitted write" anyway?  A previous non-durable write?

runtime/src/lib.rs
```diff
@@ -696,6 +696,13 @@ stability_scope!(BETA {
     /// When a blob is dropped, any unsynced changes may be discarded. Implementations
     /// may attempt to sync during drop but errors will go unhandled. Call `sync`
     /// before dropping to ensure all changes are durably persisted.
+    ///
+    /// # Torn writes
+    ///
+    /// A write interrupted by a crash before it is synced may be partially applied: any subset
```
roberto-bayardo: I think we (sometimes) assume something a bit stronger: "any *prefix* of its bytes may hold the new value"?

Eg. v1 header recovery.

runtime/src/utils/buffer/paged/writer.rs
roberto-bayardo: stale comment? we no longer "write around"

roberto-bayardo: Since flush_internal(..., sync = true) now directly uses write_at_sync for all writes, it never falls through using plain writes when data was written.

```suggestion
        // Otherwise, the flush had no bytes to write. Sync only if a
```

roberto-bayardo:
good testing suggestion from my Fable session: "A test that lets the _new_ CRC land while tearing the extension payload, then asserts recovery falls back to the prefix, would pin the safety property most tightly"

runtime/src/storage/faulty.rs
```diff
@@ -42,6 +42,10 @@ pub struct Config {
     /// before failure) rather than a complete failure (no bytes written).
     /// Only applies when `write_rate` triggers a failure.
     /// Value from 0.0 (always complete failure) to 1.0 (always partial write).
+    ///
+    /// Injected partial writes persist a prefix of the buffer, a strictly narrower
```
patrick-ogrady: I don't think this is right.

runtime/src/storage/tokio/mod.rs
```diff
@@ -292,6 +292,42 @@ mod tests {
...
+    #[tokio::test]
+    async fn test_write_at_sync_spans_submissions() {
+        let mut rng = sys_rng();
+        let storage_directory =
+            env::temp_dir().join(format!("storage_tokio_spans_{}", rng.random::<u64>()));
+        let config = Config::new(storage_directory, 2 * 1024 * 1024);
+        let storage = Storage::new(config, test_pool());
+
+        let (blob, _) = storage.open("partition", b"spans").await.unwrap();
+        let chunks = 1500usize;
```
roberto-bayardo: I assume 1500 is chosen to exceed the IOVEC max?  Maybe document this (or better use the const if its visibility allows it)?

roberto-bayardo (dismissed):
Verdict: approved; ready to come out of draft and merge. The whole-page rewrite preserves the committed payload prefix and protected CRC slot byte-for-byte, the shrink protocol remains correctly staged, and SyncState preserves the distinction between range durability and a barrier for earlier dirty state. WriteOptions propagation and the RWF_DONTCACHE fallbacks are consistent across backends. One non-blocking coverage improvement remains: selectively persist the new CRC while tearing the extension payload, then verify recovery falls back to the previous prefix. Durability semantics: this PR consistently narrows full-integrity syncs to data-integrity syncs (RWF_SYNC -> RWF_DSYNC, sync_all -> sync_data, and io_uring Fsync -> Fsync with DATASYNC). This still satisfies the Blob contract: data, file length, and metadata required to retrieve the data remain durable, while timestamp-only metadata may not. It is correct and non-blocking, but it is a broader durability-semantics change that reviewers should note explicitly.

runtime/src/iouring/mod.rs
```diff
@@ -422,45 +425,42 @@ impl Handle {
...
-            result: None,
-            sender: tx,
-        }))
-        .await
-        .map_err(|_| Error::WriteFailed)?;
-        rx.await.map_err(|_| Error::WriteFailed)?
-    }
-
-    /// Submit a logical positioned write with per-write sync and wait for its completion.
+    /// Submit a logical positioned write request with the requested cache and durability behavior.
```
danlaine: nit: the removed "and wait for its completion" is worth keeping.

runtime/src/iouring/mod.rs
```diff
@@ -422,45 +425,42 @@ impl Handle {
...
+    ///
+    /// Durability is fused into the write (`RWF_DSYNC`) only when the whole write fits one
+    /// submission: fusing every submission would serialize the submissions behind per-SQE
+    /// durability waits. A larger durable write is submitted plain and followed by one fsync,
+    /// which keeps the device pipelined and pays a single wait at the end.
```
danlaine:
```suggestion
     /// A durable write that fits one submission carries `RWF_DSYNC` and is durable on
     /// completion. A larger write is submitted plain and finished with one fsync: one
     /// device flush at the end instead of one per submission.
```

runtime/src/iouring/request.rs
```diff
@@ -519,16 +564,26 @@ pub(super) struct WriteAtRequest {
...
     pub(super) sender: oneshot::Sender<Result<(), Error>>,
 }
 
 impl WriteAtRequest {
-    /// Return the flags for this write request, setting `RWF_SYNC` when `sync` is set.
-    const fn rw_flags(&self) -> i32 {
-        if self.sync { libc::RWF_SYNC } else { 0 }
+    /// Return the flags for this write request: `RWF_DSYNC` when `sync` is set (data
+    /// durability is the contract, so timestamp-only commits are skipped), plus the
+    /// request-local `RWF_DONTCACHE` hint.
```
danlaine: nit: this feels a bit overly verbose. It's pretty apparent that the returned thing are flags based on the state of self.

runtime/src/lib.rs
```diff
@@ -728,23 +776,26 @@ stability_scope!(BETA {
...
-        /// This is not a durability barrier for previous operations. When it completes,
-        /// only the bytes submitted to this call are guaranteed durable. Earlier unsynced
-        /// [`Blob::write_at`] or [`Blob::resize`] calls require [`Blob::sync`] to become
-        /// durable.
-        fn write_at_sync(
+        /// With [`WriteOptions::SYNC`], the submitted bytes are durably persisted before this
+        /// operation returns. This is not a durability barrier for previous operations: earlier
+        /// writes without [`WriteOptions::SYNC`] and earlier [`Blob::resize`] calls require
+        /// [`Blob::sync`] to become durable.
+        fn write_at_with(
```
danlaine: Are we sure we want this API as opposed to having one `write_at` with `options: WriteOptions` with `WriteOptions::Default()` mostly used?

runtime/src/storage/audited.rs
```diff
@@ -80,6 +80,28 @@ pub struct Blob<B: crate::Blob> {
...
 }
 
+impl<B: crate::Blob> Blob<B> {
+    async fn write_at_inner(
+        &self,
+        offset: u64,
+        bufs: IoBufs,
+        options: WriteOptions,
+    ) -> Result<(), Error> {
+        let operation: &[u8] = if options.contains(WriteOptions::SYNC) {
```
danlaine: I think there's a coherent argument to be made that sync is special among WriteOptions and therefore it's the only one whose presence/absence is recorded here. But worth raising that that's what we're doing.

runtime/src/storage/tokio/blob.rs
```diff
@@ -1,18 +1,45 @@
...
+        Arc,
+        atomic::{AtomicBool, Ordering},
+    },
 };
 use tokio::task;
 
-// Cap iovec batch size: larger iovecs reduce syscall count but increase
-// per-write kernel setup overhead.
-const IOVEC_BATCH_SIZE: usize = 32;
+// Cap iovec batches at the kernel's IOV_MAX (1024): larger submissions fail with
```
danlaine: nit: more parallel with iouring?

```
/// Cap iovec batches at the kernel's IOV_MAX (1024): larger submissions fail with
/// EINVAL. Storage writes span hundreds of chunks (two per page) and want maximal
/// batches, which measure faster with no per-iovec penalty. Network sends never
/// approach the cap (messages are a few chunks), so they share it.
pub(super) const IOVEC_BATCH_SIZE: usize = 1024;
```

danlaine (commented):
Another finding:

Shared run_storage_tests never passes DONT_CACHE. Paged-writer tests run on deterministic/Memory, which ignores the flag. io_uring has a request-level unit test; Tokio has none. On kernels < 6.14 every paged write depends on this retry path.

runtime/src/storage/audited.rs
```diff
@@ -107,30 +107,20 @@ impl<B: crate::Blob> crate::Blob for Blob<B> {
...
-    async fn write_at_sync(
+    async fn write_at_with(
         &self,
         offset: u64,
         bufs: impl Into<IoBufs> + Send,
+        options: WriteOptions,
     ) -> Result<(), Error> {
         let bufs = bufs.into();
-        self.auditor.event(b"write_at_sync", |hasher| {
+        self.auditor.event(b"write_at", |hasher| {
```
patrick-ogrady, line 117: Opted to just call everything that passes through `write_at_with` `write_at`.

runtime/src/lib.rs
patrick-ogrady: I was very much torn on this. I'm not thinking that'll be clearer.

PR #4324, opened 2026-07-23
[storage] Add `start_sync` to the remaining QMDB databases

danlaine:
#4310 introduced pipelined `start_sync` on `qmdb::any` and `qmdb::current`. This extends it to the rest of the QMDBs:

```rust
pub async fn start_sync(mut self) -> Result<(Self, Handle<()>), Error>
```

on `immutable::Immutable`, `keyless::Keyless`, `store::Db`, and both `CompactDb`s. As on the existing databases, awaiting the handle gives the same durability guarantee as `commit()`, plus a best-effort advance of the recovery watermark; a new sync waits for the prior sync before starting, and failures surface on the handle and again on the next durability operation.

## Delegating databases

`immutable`, `keyless`, and `store` keep all durable state in a single journal that already has `start_sync`, so their methods are one-line delegations (with the existing `start_sync_calls` counter on the authenticated databases).

## Compact databases

The compact databases persist through `witness::Store`, whose `commit`/`sync` no-op when the cached tip witness already matches the in-memory Merkle. That no-op was justified by an invariant `start_sync` breaks: the cache is installed when the sync *starts*, not when it completes, so "cache matches Merkle" no longer implies "durable".

`Store` therefore tracks whether a pipelined sync may still be in flight (`sync_started`):

- `commit`/`sync` with nothing new to append drain the in-flight sync instead of no-op'ing, so they never report an unproven tip durable and a failed pipelined sync resurfaces as an error.
- The `CompactDb::rewind` fast path (`target == size`) requires the flag to be clear; otherwise it falls through to the store, which drains before returning. Without this, rewind would return `Ok` under its "made durable before returning" contract while the tip was unproven — or after its sync had already failed.
- `Store::start_sync` with nothing new to append still starts a journal sync: the handle proves the current tip durable and immediately resurfaces any retained failure.
- A pending compact-sync import is marked persisted when its replacement sync starts. Sync failures are fatal to the instance and a crash before completion leaves the journal unreopenable until re-syncing recovers it — the same shape as the existing `commit` replacement window.

## Tests

Each database gets a parked-sync suite on the mock backend: `start_sync` overlaps reads and applies while the fsync is in flight; an in-flight failure surfaces on the handle and the next durability op; an awaited handle alone guarantees recovery on reopen; prune (and rewind, for `immutable`) drains the in-flight sync before mutating storage. The compact suites additionally pin the drain semantics: a no-op `sync()` after `start_sync` drains, a no-op second `start_sync` appends no duplicate witness entry, the rewind fast path drains (and fails on a failed sync) instead of fast-pathing, and an import persists through a pipelined sync. The immutable/keyless metrics tests now assert `start_sync_calls`.

### DIFF
--- a/storage/src/qmdb/compact/witness.rs
+++ b/storage/src/qmdb/compact/witness.rs
@@ -294,16 +294,31 @@ impl<E: Context, F: Family, D: Digest> Store<E, F, D> {
         H: Hasher<Digest = D>,
         S: Strategy,
     {
+        // Match the deferred-failure convention used by the journal: return a prior completion's
+        // error through a ready handle before a later completion can replace it. Errors while
+        // staging or initiating this sync continue to use the outer result.
+        if let Err(err) = self.wait_for_sync().await {
+            return Ok((self, Handle::ready(Err(err))));
+        }
+
+        // Decide whether the journal needs a new witness before starting this sync. During import,
+        // staging verifies the cached witness before clearing the journal it will replace.
         let verified;
         (self, verified) = self
             .stage::<H, S>(merkle, inactivity_floor_loc, last_commit_op_bytes)
             .await?;
+
         if let Some(verified) = verified {
+            // Publish the witness to the cache only after it has been appended. The cache may
+            // then match the Merkle before the append is durable, which `pending_sync` tracks.
             (self.journal, _) = self.journal.append(&verified.witness).await?;
             self.import_pending.store(false, Ordering::Relaxed);
             merkle.prune_to_frontier();
             self.replace(verified);
         }
+
+        // Share one completion between the caller and the store. Retaining a clone keeps a
+        // dropped handle's failure observable by the next durability operation.
         let handle;
         (self.journal, handle) = self.journal.start_sync().await?;
         let completion: SyncCompletion = handle.boxed().shared();
--- a/storage/src/qmdb/compact/witness.rs
+++ b/storage/src/qmdb/compact/witness.rs
@@ -25,7 +25,7 @@ use crate::{
 use commonware_codec::{Decode as _, EncodeSize, Read, Write};
 use commonware_cryptography::{Digest, Hasher};
 use commonware_parallel::Strategy;
-use commonware_runtime::Handle;
+use commonware_runtime::{Error as RError, Handle};
 use commonware_utils::sync::RwLock;
 use futures::FutureExt as _;
 use std::sync::atomic::{AtomicBool, Ordering};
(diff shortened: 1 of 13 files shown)


storage/src/qmdb/compact/witness.rs
```diff
@@ -133,6 +135,9 @@ pub(crate) struct Store<E: Context, F: Family, D: Digest> {
     /// first persist replaces them with the cached witness and clears this flag. Mutations are
     /// serialized by ownership of the store, so `Relaxed` suffices.
     import_pending: AtomicBool,
+
+    /// Whether a pipelined journal sync from [`Self::start_sync`] may still be in flight.
+    sync_started: bool,
```
patrick-ogrady: Shouldn't we just keep a clone of the handle here?

danlaine: Would that buy us anything relative to a bool?

danlaine: I suppose this could prevent us from doing unneeded I/O when we call `start_sync` and then `commit()` without  making any modifications.

storage/src/qmdb/immutable/mod.rs
```diff
@@ -651,6 +652,20 @@ where
         Ok(self)
     }
 
+    /// Begin durably persisting the journal state published by prior [`Immutable::apply_batch`]
```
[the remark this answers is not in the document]
danlaine: All three locations are fine; failures from a pipelined start_sync are retained by the journal that started the sync and resurface on its next durability operation, not only via the returned handle.

- Immutable/Keyless::rewind (no-op path): rewind is not a durability operation — its docs say "a successful rewind is not restart-stable until a subsequent commit or sync." The truncating path doesn't sync either, so the no-op path is consistent. A pipelined sync failure still surfaces on the handle and on the next commit/sync (tested in test_*_start_sync_failure_propagates).
- witness::Store::rewind: runs a full journal.sync() after truncating, which surfaces any retained failure before pending_sync is cleared.
- witness::Store::sync: both branches run a full journal.sync() whenever a pipelined sync is outstanding (the nothing-to-append case handles this explicitly), which supersedes the pipelined sync's guarantee and surfaces any retained failure. Tested: a sync after a failed pipelined sync observes the retained failure rather than no-opping.

The compact rewind fast path waits on pending_sync only because that method has a stronger contract ("the rewind is made durable before this method returns") — and it already does so.

patrick-ogrady (approved):

PR #4325, opened 2026-07-23
[glue] Use `start_sync`

patrick-ogrady:
Removes the blocking fsync from the stateful actor's finalization path. `ManagedDb::finalize` applies the batch and calls `start_sync` instead of `sync`, returning a `Handle<()>` that resolves when the batch is durable. The actor applies block N+1 while block N's flush is in flight and acknowledges marshal only once the flush completes, so marshal's processed floor never passes unflushed database state.

## Actor

`DatabaseSet::finalize` returns a `Barrier` holding one flush handle per database. `durable()` is `true` when every flush lands, `false` on shutdown (marshal redelivers after restart), and panics on a real flush failure, since the databases have already advanced past the batch. The processing loop parks `barrier.durable() → acknowledge` futures in a pool drained each iteration, bounded by marshal's `max_pending_acks`. `Application::finalized` now fires when the block is readable, with durability observed via the barrier before the acknowledgement.

During state sync, a finalized block is acknowledged once a persisted certified recovery floor covers it: the actor fetches marshal's latest certificate, persists it as the `SyncState::InProgress` floor, and only then acknowledges, so a crash resumes from a floor at least as new as anything the sync engine acted on. Target changes are decoupled from acknowledgements. The live target is held for a configured window (`Config::retarget_delay`) while newer finalizations queue, the newest queued target is forwarded when the window expires, and blocks whose coalesced target never reached the engine are applied during handoff once sync completes.

The syncer drops its tip-update ring sender when the sync artifact publishes, so an update forwarded after the coordinator's final drain resolves through a retry with the artifact instead of stranding its observation.

Immutable, keyless, and compact databases keep a blocking `sync()` behind a ready handle until their `start_sync` support lands (#4324).

## Pruning without a flush

Prunes may run while flushes are in flight, so the journals' unconditional flush-before-prune becomes a barrier check plus a caller contract: the pruned boundary must be justified by durable data, because recovery after a crash must never need pruned items to explain retained ones.

- `contiguous::Mutable::durable()` returns a conservative durably persisted range, and fixed, variable, and authenticated prunes skip their durability pass when it already covers the boundary.
- The variable journal's barrier tracks only the data blobs, and the offsets journal gates on its own barrier. Offsets durable ahead of the data are repairable (recovery rewinds them to the watermark and replays data forward), so a prune whose data is already durable commits nothing.
- `any::Db` tracks the appended-but-unproven commit and exposes the durable frontier as `barrier() -> Range<Location>`, committing first only when no durable commit justifies the target.
- `current::Db::prune` records the bitmap boundary at the caller's target rather than the live one (a recorded boundary past the retention window would make the startup repair rewind impossible) and commits first when the durable floor and operation count cannot rebuild that boundary (delayed-merge absorption depends on both).
- `authenticated::prune` clamps an out-of-domain location instead of aborting, matching the journal convention.

Pinned by fail-first regression tests plus a randomized crash-interleaving simulation.

## Compact sync serve policy

Sources previously served only their exact tip, so a client whose fetched target trailed the source always failed. Sources now serve any target at or below the tip that they can prove satisfies the request: witness sources rebuild and verify the retained entry's frontier against the requested root, and full sources verify the historical commit proof against it before reading the pins. Declines split by retryability. `StaleTarget` means the source cannot see the leaf count (past the tip, outside the retained window, or no retained witness entry commits it). `DivergentTarget` means the source reaches the leaf count under a different history and can never serve it. The glue p2p actor answers neither and records them separately, since an honest source that answered a divergent target would be network-blocked by the requester's resolver.

The engine validates every target update (strictly advancing leaf count, changed root), splits fetching from persistence so a target update never cancels an in-flight witness-journal replacement, and treats a closed finish channel as an error instead of completing silently.

The provable window is `max(operations-log boundary, Merkle boundary)`: sync pins the Merkle boundary at the sync range's start while the log prunes on section boundaries, so a synced database can retain operations it cannot prove.

## Behavior changes

- `prune` still guarantees recovery needs nothing below the pruned boundary, but no longer flushes to get there. Appended-but-unsynced items retained by a prune are no longer made durable as a side effect.
- Restart may find the databases ahead of marshal's floor. Startup reconciliation rewinds as before, and redelivered blocks are absorbed without reapplying.
- `qmdb::Error::HistoricalFloorPruned` is renamed `NoCommitAtSize`: pruned sizes were already reported as `ItemPruned`, and the variant fires only for a size that does not end at a retained commit.


glue/src/stateful/actor/core/processing.rs
```diff
@@ -146,24 +168,43 @@ where
...
+                        // ahead of flushed database state (the startup rewind
+                        // contract), without blocking the loop on the flush.
+                        // Marshal's ack window bounds the flush backlog.
+                        syncs.push(async move {
+                            let durable = durability.durable().await;
+                            if durable {
+                                acknowledgement.acknowledge();
+                            }
+                            durable
+                        });
```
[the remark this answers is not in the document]
patrick-ogrady: Fixed: `on_start` now drains completed flushes every iteration (`while poll!(syncs.next_completed()).is_ready() {}`) before taking the next unit of work, so acknowledgements flow even when the mailbox is never idle. The idle race inside `next` remains the wake path when the loop is parked.

glue/src/stateful/actor/core/processing.rs
```diff
@@ -146,32 +166,53 @@ where
...
                 Step::Message(Message::SubscribeDatabases { response }) => {
                     response.send_lossy(self.processor.databases().clone());
                 }
                 Step::Prune(prune) => {
+                    // Durability barrier: every deferred flush must complete
+                    // before pruning discards the history a restart would need
+                    // to replay unflushed state (see `DatabaseSet::prune`).
+                    while !syncs.is_empty() {
+                        syncs.next_completed().await;
+                    }
```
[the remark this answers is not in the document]
patrick-ogrady: Resolved by redesign: the prune arm no longer drains the pool at all. The durability barrier moved into the databases — each `ManagedDb::prune` completes only once the pruned range is durably justified (qmdb gates on `durable_floor`). Marshal's prune height is additionally bounded by acked (⇒ flushed) state, so a never-durable flush cannot lose replay history.

glue/src/stateful/actor/core/processing.rs
```diff
@@ -82,8 +94,25 @@ where
...
+                                loop {
+                                    select! {
+                                        message = mailbox.recv() => {
+                                            break message.map(Step::Message);
+                                        },
+                                        _ = pool.next_completed() => {},
+                                    }
+                                }
+                            })
+                        }
```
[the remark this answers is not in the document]
patrick-ogrady: Intentional, matching marshal's own durability pool: at teardown the pool drops and un-acked blocks are redelivered after restart (an in-flight failure resurfaces on the next durability operation after reopen). The pool comment now states the teardown behavior explicitly.

glue/src/stateful/actor/core/processing.rs
```diff
@@ -146,32 +175,53 @@ where
...
+                        // Acknowledge marshal only once the batch's flush
+                        // completes, so marshal's processed floor never runs
+                        // ahead of flushed database state (the startup rewind
+                        // contract), without blocking the loop on the flush.
+                        // Marshal's ack window bounds the flush backlog.
+                        syncs.push(async move {
+                            if barrier.durable().await {
+                                acknowledgement.acknowledge();
+                            }
+                        });
```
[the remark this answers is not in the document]
patrick-ogrady: Intentional: withholding the ack for a non-durable flush is the floor-safety invariant — marshal's processed floor must never pass unflushed database state, and startup can only repair databases ahead of the floor. `durable()` is false only on Closed/Aborted at runtime teardown, where marshal is torn down by the same supervisor; a panic in that race resolves to exactly the redeliver-after-restart outcome. Acking instead would durably advance the floor past unflushed state, which is unrecoverable.

glue/src/stateful/actor/core/processing.rs
```diff
@@ -146,32 +175,51 @@ where
...
-                            self.processor.finalize(&self.context, block.as_ref()).await;
-                        if let FinalizeStatus::Persisted { height } = status {
-                            debug!(height = height.get(), "persisted finalized database batch");
+                        let Some(Applied { barrier, prune }) =
+                            self.processor.finalize(&self.context, block.as_ref()).await
+                        else {
+                            // Duplicate report: the block is already reflected
+                            // in the startup-aligned durable state.
+                            acknowledgement.acknowledge();
+                            return;
```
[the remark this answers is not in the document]
patrick-ogrady: Unreachable in-session: marshal dispatches each height exactly once per session (FIFO `pending_acks`), so a duplicate report only occurs after a restart, where startup has aligned the databases to durable state before `last_processed` is seeded. The duplicate-branch comment now states this.

glue/src/stateful/actor/core/processing.rs
```diff
@@ -82,8 +94,22 @@ where
                     Err(TryRecvError::Empty) => match pending_prune.take() {
                         // No message, but a prune is queued: run it.
                         Some(prune) => Either::Left(ready(Some(Step::Prune(prune)))),
-                        // No message and nothing to prune: wait on the mailbox as normal.
-                        None => Either::Right(self.mailbox.recv().map(|m| m.map(Step::Message))),
+                        // No message and nothing to prune: wait on the mailbox,
+                        // driving flush completions while idle.
+                        None => {
+                            let mailbox = &mut self.mailbox;
+                            let pool = &mut syncs;
```
danlaine:
```suggestion
                            let syncs = &mut syncs;
```

storage/src/journal/contiguous/blobs.rs
```diff
@@ -420,6 +430,30 @@ impl<E: Context> Writable<E> {
         Ok(())
     }
 
+    /// Surface a completed predecessor sync's failure without waiting on one in flight.
+    /// Clear only on success.
+    fn observe_tail_predecessor_sync(&mut self) -> Result<(), Error> {
```
danlaine: I'm not crazy about having 4 similar *_tail_*_sync methods. Agent suggests this:

```rs
/// A sync whose outcome has not yet been observed. Cleared only on success, so a
/// failure keeps failing every later operation that checks the slot.
#[derive(Default)]
struct SyncSlot(Option<SyncCompletion>);

impl SyncSlot {
    fn record(&mut self, completion: SyncCompletion);   // track until observed
    fn get(&self) -> Option<SyncCompletion>;            // for start_sync's handle-join
    const fn is_some(&self) -> bool;
    async fn drain(&mut self) -> Result<(), Error>;     // wait it out
    fn observe(&mut self) -> Result<(), Error>;         // surface completed failure, no waiting
}

// The two fields become tail_predecessor_sync: SyncSlot and tail_sync: SyncSlot, the four drain_tail_*/observe_tail_* methods are deleted, and every call site reads as the operation on the slot it targets — the prune hunk you disliked becomes:

if min_blob == self.tail_blob_index() {
    self.tail_predecessor_sync.drain().await?;
} else {
    self.tail_predecessor_sync.observe()?;
}
self.tail_sync.observe()?;
```

```rs
/// A sync whose outcome has not yet been observed. Cleared only on success, so a
/// failure keeps failing every later operation that checks
the slot.
#[derive(Default)]
struct SyncSlot(Option<SyncCompletion>);

impl SyncSlot {
    /// Track `completion` until its outcome is observed.
    fn record(&mut self, completion: SyncCompletion) {
        self.0 = Some(completion);
    }

    /// The tracked completion, if any.
    fn get(&self) -> Option<SyncCompletion> {
        self.0.clone()
    }

    /// Whether a sync is being tracked.
    const fn is_some(&self) -> bool {
        self.0.is_some()
    }

    /// Wait out the tracked sync.
    async fn drain(&mut self) -> Result<(), Error> {
        let Some(completion) = self.0.clone() else {
            return Ok(());
        };
        completion.await?;
        self.0 = None;
        Ok(())
    }

    /// Surface a completed sync's failure without waiting on one in flight.
    fn observe(&rror> {
        if let Some(completion) = &self.0
            && lion.clone().now_or_never()
        {
            result?;
            self.0 = None;
        }
        Ok(())
    }
}
```

storage/src/journal/authenticated.rs
```diff
@@ -990,6 +995,10 @@ where
         Ok((journal, pruned))
     }
 
+    fn barrier(&mut self) -> u64 {
```
danlaine: Could combine with pruning_boundary?
```
/// The positions of durably persisted elements
fn durable(&mut self) -> Range<u64>;
```

storage/src/journal/contiguous/variable.rs
```diff
@@ -1559,10 +1557,18 @@ impl<E: Context, V: CodecShared> Inner<E, V> {
...
         // other durability path maintains.
-        let data_sync = self.blobs.start_sync().await;
-        data_sync.await?;
-        self.offsets = self.offsets.commit().await?;
-        self.barrier.mark_durable(self.bounds.end);
+        //
+        // All three hazards vanish once the barrier covers the new boundary: the barrier
+        // proves both data and offsets durable through it, retained items below it cannot
+        // tear, recovery truncates only above it, and the caller guarantees the boundary is
+        // justified by durable data (see [Mutable::prune]).
```
danlaine: Less claude-like (three hazards vanish!) suggestion
```suggestion
        // If the barrier already covers the new boundary, skip the sync. Data and offsets
        // below the barrier are durable, and the caller justified the boundary with durable
        // data (see [Mutable::prune]).
```

glue/src/stateful/actor/core/syncing.rs
```diff
@@ -191,14 +192,25 @@ where
...
+    ) -> (Self, Option<FinalizedHandoff<Arc<A::Block>>>) {
         if self.artifact.is_none() {
             let anchor = Anchor::from(block.as_ref());
             let targets = A::sync_targets(block.as_ref());
 
+            // Persist the retargeted floor before the live session can record it and before
+            // marshal is acknowledged: the sync engine's durable state advances toward the
+            // new target, and marshal prunes behind acknowledged blocks, so a crash must
+            // resume from a floor at least as new as any target the engine acted on.
+            let finalization = self
```
patrick-ogrady: During review, found this issue

storage/src/qmdb/any/db.rs
```diff
@@ -52,6 +52,19 @@ enum SnapshotUndo<F: Family, K> {
     },
 }
 
+/// The db-level durable frontier: the state proven durable in the log, observed from the
+/// journal barrier.
+///
+/// Constructible only through [`Db::barrier`], so boundary math over durable state cannot
+/// accidentally read live fields.
+#[derive(Clone, Copy)]
+pub(crate) struct Barrier<F: Family> {
```
danlaine: This could be a `Range`?

storage/src/qmdb/any/db.rs
```diff
@@ -87,6 +90,17 @@ pub struct Db<
...
+    /// The highest inactivity floor declared by a commit proven durable in the log.
+    pub(crate) durable_floor: Location<F>,
+
+    /// The operation count covered by the newest commit proven durable in the log (its
+    /// location plus one).
+    pub(crate) durable_size: Location<F>,
+
+    /// Commits appended but not yet proven durable, oldest first, as
+    /// `(commit location, declared inactivity floor)` pairs.
+    pub(crate) pending_commits: VecDeque<(Location<F>, Location<F>)>,
```
danlaine: These two pairs can be Ranges

storage/src/merkle/persisted/full.rs
```diff
@@ -707,7 +713,12 @@ impl<F: Family, E: Context, D: Digest, S: Strategy> Merkle<F, E, D, S> {
         }
 
         // Flush items cached in the mem to disk to ensure the current state is recoverable.
-        self = self.sync().await?;
+        // Once the barrier covers the prune position, the durable journal prefix always
+        // covers the metadata boundary written below, so the flush is skipped. Nodes above
+        // the barrier stay cached and are not guaranteed to survive a crash.
+        if self.barrier() < pos {
```
danlaine: Fwiw I find the meaning of `barrier()` harder to grok than `.durable().end`

glue/src/stateful/actor/core/syncing.rs
```diff
@@ -169,10 +169,11 @@ where
                     acknowledgement,
                 } => {
                     let process = info_span!(parent: &span, "stateful.actor.syncing_finalized");
-                    let handoff = self
+                    let (this, handoff) = self
                         .process_finalized(block, acknowledgement)
                         .instrument(process)
                         .await;
```
danlaine:
```suggestion
                    let handoff;
                    (self, handoff) = self
                        .process_finalized(block, acknowledgement)
                        .instrument(process)
                        .await;
```

danlaine:
2 more small findings


## 1. Historical compact witnesses are only partially validated

**Severity:** P2  
**Status:** fixed locally with regression tests  
**Locations:** `storage/src/qmdb/keyless/compact.rs:483` and
`storage/src/qmdb/immutable/compact.rs:498`

### What changed

Compact state sync requests a witness for a target `(root, leaf_count)`. A witness
contains:

1. the final commit operation;
2. an inclusion proof authenticating that operation against the target root;
3. `pinned_nodes`, the Merkle frontier needed to reconstruct the compact tree.

Previously, compact sources served only their cached tip witness, which had already
been fully verified when installed. This PR adds below-tip serving by loading older
entries from `witness::Store::entry_at`.

### Defect

The historical serving path checks the commit proof but not `pinned_nodes`:

```rust
if !pre_verified
    && !crate::qmdb::verify_proof::<H, _, _>(
        &last_commit_proof,
        Location::new(*target.leaf_count - 1),
        std::slice::from_ref(&op),
        &target.root,
    )
{
    return Err(compact_sync::ServeError::StaleTarget { ... });
}
```

The proof is independent of `pinned_nodes`: it authenticates the commit using sibling
digests carried by the proof itself. A semantically corrupted but validly framed
historical entry can therefore contain:

- an intact commit operation;
- an intact, valid proof;
- an invalid pinned frontier.

That entry passes the source check and is served. The client performs the missing
frontier check, rejects the payload, and requests another candidate. A direct
`Arc<Db>` resolver returns `FetchResult { callback: None, .. }`, so there is no peer
selection feedback and the same invalid payload can be returned indefinitely.

### Corruption model

The reproduction uses `witness::tests::corrupt_entry`, which decodes, mutates, and
rewrites the entry with valid journal framing. This demonstrates robustness against
**checksum-valid semantic corruption**. Raw bit rot or a torn page should normally be
caught by the journal's CRC and recovery layers before this code runs.

There is also a defense-in-depth gap: the historical serving path does not explicitly
check that the decoded inactivity floor is at or below the commit location. The proof
authenticates the operation bytes, so an independent floor mutation would already fail
proof verification; the explicit bounds check is still appropriate before using the
floor to compute inactive peaks.

### Reproduction

The regression test:

1. commits two states, making the first witness historical;
2. records the first state's `(root, leaf_count)`;
3. changes one pinned-node digest in the historical entry while preserving its
   operation and proof;
4. reopens the database successfully because the current tip remains intact;
5. requests the historical target and expects `StaleTarget`.

Against the PR code, the historical witness is incorrectly served. With the local fix,
both variants pass:

```text
PASS qmdb::keyless::compact::tests::test_compact_state_declines_tampered_historical_pinned_nodes
PASS qmdb::immutable::compact::tests::test_compact_state_declines_tampered_historical_pinned_nodes
Summary: 2 tests run: 2 passed
```

---

## 2. `prune` documentation promises too much durability

**Severity:** P2  
**Status:** fixed locally  
**Locations:** `glue/src/stateful/db/mod.rs:359` and `:482`

### Defect

The `ManagedDb::prune` and `DatabaseSet::prune` documentation still says:

> This call makes changes durable and ensures they will be present on startup
> without replay.

That described the old behavior of the affected QMDB pruning paths: pruning flushed
pending state before removing history. This PR deliberately adds a fast path that skips
the flush when a durable commit already justifies the pruning boundary.

Consequently, after `prune` returns:

- the removal itself is durable;
- the database remains recoverable;
- newer applied state may still be waiting on its deferred flush.

The existing sentence incorrectly presents `prune` as a general durability checkpoint.
Downstream code that follows that contract can fail only during a crash between the
prune and the actual batch flush.

### Evidence

- `storage/src/qmdb/any/db.rs` commits only when
  `self.barrier().start < prune_loc`.
- Fixed and variable contiguous journals skip their internal sync when their durable
  barrier covers the new boundary.
- `prune_skips_barrier_for_durable_target` finalizes durable batch A, parks batch B's
  flush, prunes to A, and verifies that pruning starts no new sync. Batch B remains
  non-durable after the prune by design.
- `Mutable::prune` was updated with the new contract in this PR; the glue traits retained
  the old sentence.

patrick-ogrady:
> 2 more small findings
> 
> ## 1. Historical compact witnesses are only partially validated
> **Severity:** P2 **Status:** fixed locally with regression tests **Locations:** `storage/src/qmdb/keyless/compact.rs:483` and `storage/src/qmdb/immutable/compact.rs:498`
> 
> ### What changed
> Compact state sync requests a witness for a target `(root, leaf_count)`. A witness contains:
> 
> 1. the final commit operation;
> 2. an inclusion proof authenticating that operation against the target root;
> 3. `pinned_nodes`, the Merkle frontier needed to reconstruct the compact tree.
> 
> Previously, compact sources served only their cached tip witness, which had already been fully verified when installed. This PR adds below-tip serving by loading older entries from `witness::Store::entry_at`.
> 
> ### Defect
> The historical serving path checks the commit proof but not `pinned_nodes`:
> 
> ```rust
> if !pre_verified
>     && !crate::qmdb::verify_proof::<H, _, _>(
>         &last_commit_proof,
>         Location::new(*target.leaf_count - 1),
>         std::slice::from_ref(&op),
>         &target.root,
>     )
> {
>     return Err(compact_sync::ServeError::StaleTarget { ... });
> }
> ```
> 
> The proof is independent of `pinned_nodes`: it authenticates the commit using sibling digests carried by the proof itself. A semantically corrupted but validly framed historical entry can therefore contain:
> 
> * an intact commit operation;
> * an intact, valid proof;
> * an invalid pinned frontier.
> 
> That entry passes the source check and is served. The client performs the missing frontier check, rejects the payload, and requests another candidate. A direct `Arc<Db>` resolver returns `FetchResult { callback: None, .. }`, so there is no peer selection feedback and the same invalid payload can be returned indefinitely.
> 
> ### Corruption model
> The reproduction uses `witness::tests::corrupt_entry`, which decodes, mutates, and rewrites the entry with valid journal framing. This demonstrates robustness against **checksum-valid semantic corruption**. Raw bit rot or a torn page should normally be caught by the journal's CRC and recovery layers before this code runs.
> 
> There is also a defense-in-depth gap: the historical serving path does not explicitly check that the decoded inactivity floor is at or below the commit location. The proof authenticates the operation bytes, so an independent floor mutation would already fail proof verification; the explicit bounds check is still appropriate before using the floor to compute inactive peaks.
> 
> ### Reproduction
> The regression test:
> 
> 1. commits two states, making the first witness historical;
> 2. records the first state's `(root, leaf_count)`;
> 3. changes one pinned-node digest in the historical entry while preserving its
>    operation and proof;
> 4. reopens the database successfully because the current tip remains intact;
> 5. requests the historical target and expects `StaleTarget`.
> 
> Against the PR code, the historical witness is incorrectly served. With the local fix, both variants pass:
> 
> ```
> PASS qmdb::keyless::compact::tests::test_compact_state_declines_tampered_historical_pinned_nodes
> PASS qmdb::immutable::compact::tests::test_compact_state_declines_tampered_historical_pinned_nodes
> Summary: 2 tests run: 2 passed
> ```
> 
> ## 2. `prune` documentation promises too much durability
> **Severity:** P2 **Status:** fixed locally **Locations:** `glue/src/stateful/db/mod.rs:359` and `:482`
> 
> ### Defect
> The `ManagedDb::prune` and `DatabaseSet::prune` documentation still says:
> 
> > This call makes changes durable and ensures they will be present on startup
> > without replay.
> 
> That described the old behavior of the affected QMDB pruning paths: pruning flushed pending state before removing history. This PR deliberately adds a fast path that skips the flush when a durable commit already justifies the pruning boundary.
> 
> Consequently, after `prune` returns:
> 
> * the removal itself is durable;
> * the database remains recoverable;
> * newer applied state may still be waiting on its deferred flush.
> 
> The existing sentence incorrectly presents `prune` as a general durability checkpoint. Downstream code that follows that contract can fail only during a crash between the prune and the actual batch flush.
> 
> ### Evidence
> * `storage/src/qmdb/any/db.rs` commits only when
>   `self.barrier().start < prune_loc`.
> * Fixed and variable contiguous journals skip their internal sync when their durable
>   barrier covers the new boundary.
> * `prune_skips_barrier_for_durable_target` finalizes durable batch A, parks batch B's
>   flush, prunes to A, and verifies that pruning starts no new sync. Batch B remains
>   non-durable after the prune by design.
> * `Mutable::prune` was updated with the new contract in this PR; the glue traits retained
>   the old sentence.

point [1] doesn't make sense for us (we don't need to support "checksum-valid semantic corruption")

addressed [2]

PR #535, opened 2025-02-26
Add metrics to `resolver::p2p`

BrendanChou:
fixes #523

PR #4327, opened 2026-07-23
[storage] Repair torn interior pages on segmented-journal replay

### LINKED ISSUE #4326: [storage] Review Durability Assumptions
While working on #4323, I realized we were making too strong of an assumption about filesystem durability (namely that powerloss would preserve some prefix of our pending writes between `fsync`). This is wrong.

We should review our recovery implementations across storage to ensure they don't assume this (segmented/oversized journals appear particularly suspicious...we may not catch a "hole" between writes on restart).

roberto-bayardo:
Resolves: #4326

## Summary

Closes the two segmented-journal durability gaps called out in #4326 under the crash model documented on `Blob` in #4323 — any subset of a write's bytes since the last completed sync may persist, within and across writes; only `sync` orders durability. #4326's audit vetted the contiguous journals and the freezer (checkpoint mode) as already correct; this PR fixes the remaining two:

1. **Segmented journals could deterministically wedge on startup.** A crash during an unsynced multi-page flush can leave an interior page invalid while a later page survives. The per-section backward scan sizes the section out to that valid island, so replay met the interior hole mid-stream and failed with `ReplayFailed` — and re-init hit the same on-disk state, so it was a permanent wedge, not a deferred read error.
2. **`oversized` (infer mode) could adopt an interior value hole.** Index replay never reads values and init value-verified only the last entry per section, so an interior glob hole below a verified last entry was adopted and surfaced later as a read-time checksum error on data that was never acknowledged.

## The fixes

**1. Replay repairs, and replay-before-use is structural.**
- `Replay::next` now truncates a torn section to its last well-formed item (`repair_blob`) and continues when — and only when — the paged reader reports `InvalidChecksum`, mirroring the existing trailing-bytes / incomplete-item repair. `variable` gains the arm at both `ensure` sites; `fixed` gains a `repairing` drop-guard, a `repair_blob` helper, and the arm. Genuine IO/decode errors still fail loud.
- `init` now returns a `Replay`; draining it and calling `finish()` is the only path to a usable `Journal`. The consumer's single existing replay doubles as the verification/repair pass, so "replay before use" is no longer a convention a consumer can silently violate. Init-with-checkpoint (freezer / oversized checkpoint mode) is the second door and skips the forced replay. Consumers migrated: aggregation, ordered-broadcast, simplex voter, cache, prunable archive, freezer, glue reshare store.
- Why safe without a watermark: a synced page cannot tear, so any torn page is unacknowledged tail data past the last sync — truncating discards only data the caller was never promised.

**2. `oversized` gains a durable per-section floor watermark (contiguous-style floor, not truncate-exact).**
- A section's watermark is raised in memory only after its joint index+value sync is proven complete (`mark_durable` after `try_join(index.sync, values.sync)`, or via `observe_pending` on a previously started sync), and published to the caller-configured metadata partition. So the floor never over-claims durability. `start_sync` publishes one interval behind and skips publication (rather than blocking) when the metadata store's previous write is still in flight (`watermarks_dirty && poll_sync()`), so it adds no metadata fsync to the hot path and cannot deadlock overlapping syncs.
- Recovery (`repair`) reconciles each section above its floor: it derives the value floor from the last durable entry (which must be readable and checksum-valid, else corruption), then forward-scans index entries and value checksums from the floor, keeping only the first contiguous valid run and truncating index and glob at the first hole. This heals an interior value hole even when a later value and index page survived the same flush. A section missing below a nonzero floor, or an unreadable durable boundary entry or value, is loud corruption — never silent rollback of acknowledged data. Truncations are synced before appends can reuse the freed ranges.
- The freezer's recovery is otherwise unchanged: in checkpoint mode it restores to exactly the published checkpoint, discarding the unsynced window (holes included).

## Configuration

The watermark metadata partition is an **explicit config field**, not a name derived from `key_partition`. `oversized`, `freezer`, and both archives now take it directly (`metadata_partition` on `freezer::Config` and `prunable::Config`, `freezer_metadata_partition` on `immutable::Config`), and initialization **validates that all partitions are mutually distinct**, failing with `InvalidConfiguration` before touching storage. This closes a family of configuration hazards:

- A derived `{key_partition}-metadata` name could silently collide with a caller's pre-existing standalone `Metadata` store and reinterpret it as watermarks (qed #539). It is now caller-named and owned; nothing is derived or hidden.
- Freezer fresh-start init validated distinctness only after its destructive reset (qed #537), and accepted a `table` partition aliased to the reserved metadata name that then failed during `destroy` (qed #538). Distinctness is now checked up front, so an invalid config fails without side effects.

## Recovery semantics / invariants

- The per-section floor is a durable lower bound: below it, index and value were made durable by a completed joint sync and must survive; above it, anything may be torn and is reconciled by the forward scan.
- The floor is advanced only after the data it vouches for is durable, and lowered before any backward move (`rewind`/`rewind_section` lower and sync the watermark before truncating; `prune` removes a section's watermark before removing its blobs).
- Legacy compatibility: a pre-watermark journal recovers at floor 0. Its first entry may begin after an orphan value prefix (an empty index section that retained value bytes); at a zero floor the scan adopts that first entry's own offset so the upgrade does not discard the acknowledged entry.
- Below-floor recovery detects *torn/accidental* corruption; it does not authenticate or fully re-validate the durable prefix, so a forged rewrite that recomputes the checksum is out of scope (CRC32 is integrity, not authentication) — consistent with storage-wide use of plain CRCs (qed #535/#536).

## Scope

This closes both segmented/oversized items in #4326. Contiguous and freezer recovery were vetted correct by #4326's audit and are unchanged here.

Cross-PR note: #4323 stages two pinning tests (`test_journal_torn_sync_recovers_synced_prefix`, `test_oversized_init_defers_interior_torn_value_to_get`) that assert the *pre-fix* dichotomy. When both land, those should be tightened to assert recovery-everywhere. This PR's own regression tests plant holes directly and do not depend on #4323's `try_tear` injection.

## Testing

- Segmented replay repair: `test_journal_replay_repairs_torn_interior_page` (variable and fixed) — plants an interior torn page with a surviving later page; asserts replay yields a non-empty contiguous prefix and leaves the journal usable. Each confirmed to reproduce the `ReplayFailed` wedge with the fix reverted.
- Oversized value-hole healing + below-floor loss: `test_oversized_recovery_rejects_entry_with_torn_value_bytes` and the recovery suite; `oversized_recovery` fuzzer accepts `Corruption` on acknowledged-data damage and only panics on genuinely unexpected results.
- Configuration: `init_rejects_partition_collision_before_reset` (freezer, fail-safe before reset), `test_init_rejects_partition_collision` (immutable, cross-partition), `test_init_leaves_foreign_derived_metadata_untouched` (prunable, #539 — foreign data at the old derived name is untouched).
- `just test -p commonware-storage` (incl. conformance, unchanged).

## Also: contiguous-journal destroy ordering (crash-safety)

Independent of the segmented-replay work above, this PR folds in a small crash-safety fix to `journal/contiguous/fixed.rs`. `destroy` removed the blobs before the checkpoint, so a crash between the two removals left a checkpoint whose watermark exceeds the now-missing blobs — which fails the next init as `Corruption` (`recover_bounds`) and cannot be re-destroyed (init is the only way to obtain the handle `destroy` needs). Removing the checkpoint first makes an interrupted destroy recoverable: surviving blobs with no checkpoint recover by walking blob lengths. Covered by `test_fixed_journal_destroy_removes_checkpoint_before_blobs`.


patrick-ogrady:
I had imagined init just directly returning replay( ) but that might be too aggressive.

i.e. remove the actual replay method and force init through this path

I also think we should remove the `init` repair operations and rely entirely on this (if we go that route)

RELEASE_NOTES.md
```diff
@@ -109,6 +109,13 @@ sections and an over-strict assertion on an empty oldest section ([#3906],
 recovered data is made durable before recovery metadata advances past it
 ([#3809], [#3904], [#3928]).
 
+Segmented journals now repair torn interior pages during replay and require
```
patrick-ogrady: You can remove this

storage/src/freezer/storage.rs
```diff
@@ -670,14 +670,31 @@ impl<E: BufferPooler + Context, K: Array, V: CodecShared> Inner<E, K, V> {
...
+            Some(checkpoint) => {
+                Oversized::init_from_checkpoint(
+                    context.child("oversized"),
+                    oversized_cfg,
+                    checkpoint,
+                )
+                .await?
+            }
+            None => {
+                let mut replay = Oversized::init(
```
patrick-ogrady: I don't think we'll need to make the same tweak to oversized once we add the metadata?

If that uses segmented internally, I suppose that's a good reason to not force replay?

storage/src/journal/segmented/fixed.rs
```diff
@@ -71,26 +80,7 @@ impl<E: Storage + Metrics, A: CodecFixedShared> Inner<E, A> {
                 page_cache_ref: cfg.page_cache,
             },
         };
-        let mut manager = Manager::init(context, manager_cfg).await?;
-
-        // Repair any blobs with trailing bytes (incomplete items from crash)
```
patrick-ogrady, line 76: ❤️

storage/src/freezer/storage.rs
patrick-ogrady: I hate to say that you should also fixed oversized in this but I think that might be the cleanest solution?

storage/src/journal/segmented/oversized.rs
```diff
@@ -152,20 +169,28 @@ impl<E: BufferPooler + Storage + Metrics, I: Record + Send + Sync, V: CodecShare
...
+            oversized.watermarks.clear();
+            oversized.watermarks = oversized.watermarks.sync().await?;
+        }
+        oversized.track_watermarks = false;
+        oversized.restore(section, index_size).await
     }
 
     /// Open the index and value journals without reconciling them.
     async fn open(context: E, cfg: Config<V::Cfg>) -> Result<Self, Error> {
+        let metadata_partition = format!("{}-metadata", cfg.index_partition);
```
patrick-ogrady: we need a unique cfg for this (deriving this isn't safe)

storage/src/freezer/mod.rs
```diff
@@ -241,6 +241,9 @@ pub enum Error {
 #[derive(Clone)]
 pub struct Config<C> {
     /// The [commonware_runtime::Storage] partition for the key index journal.
+    ///
+    /// The companion `{key_partition}-metadata` partition is reserved for recovery watermarks and
+    /// must be distinct from the value and table partitions.
     pub key_partition: String,
```
patrick-ogrady, line 244: See note about distinct partition (we can't guarantee this)

RELEASE_NOTES.md
roberto-bayardo: Removed.

storage/src/freezer/storage.rs
roberto-bayardo: Done. Oversized now has metadata-backed recovery. Infer-mode init returns a replay, while freezer uses `init_from_checkpoint` and obtains the journal directly.

storage/src/journal/segmented/oversized.rs
roberto-bayardo: Done. `metadata_partition` is now an explicit `Oversized::Config` field.

storage/src/freezer/mod.rs
roberto-bayardo: Done. Freezer now takes an explicit `metadata_partition`, and init validates the key, value, table, and metadata partition fields before touching storage.

roberto-bayardo:
Done on the construction path: `init` now returns `Replay`, and the old init-time self-repair is gone. I kept `Journal::replay` because it remains useful for replaying an already-recovered live journal from an arbitrary start. It can't bypass startup recovery because obtaining a journal first requires draining the init replay and calling `finish()` (or using a composite checkpoint constructor).

glue/src/dkg/tests/reshare/harness.rs
```diff
@@ -1184,6 +1184,7 @@ fn archive_config<C>(
     prunable::Config {
         translator: TwoCap,
         key_partition: format!("{prefix}-{name}-key"),
+        metadata_partition: format!("{prefix}-{name}-key-metadata"),
```
patrick-ogrady: If this is "key_metadata" we should change the config accoridngly?

glue/src/dkg/probe/mod.rs
```diff
@@ -634,6 +634,7 @@ mod tests {
             freezer_table_resize_frequency: 10,
             freezer_table_resize_chunk_size: 10,
             freezer_key_partition: format!("{prefix}-{name}-freezer-key"),
+            freezer_metadata_partition: format!("{prefix}-{name}-freezer-key-metadata"),
```
patrick-ogrady: likewise here with key metadata.

glue/src/dkg/tests/reshare/harness.rs
roberto-bayardo: Done — this is the recovery-watermark metadata rather than the key's, so I renamed it off the key partition: it is now `{prefix}-{name}-metadata`. Applied the same drop-`key` rename to every such call site across storage/consensus/glue/examples and regenerated the affected conformance fixtures.

glue/src/dkg/probe/mod.rs
roberto-bayardo: Likewise — `freezer_metadata_partition` is now `{prefix}-{name}-freezer-metadata`, no longer named after the freezer key partition, consistent with the PR-wide rename.

storage/src/freezer/storage.rs
```diff
@@ -642,13 +642,31 @@ impl<E: BufferPooler + Context, K: Array, V: CodecShared> Inner<E, K, V> {
             "table_initial_size must be a power of 2"
         );
 
+        // Reject a partition-name collision before the destructive reset below, so an invalid
+        // configuration fails without deleting data. The key, value, table, and metadata
+        // partitions must all be distinct.
+        let partitions = [
```
patrick-ogrady: remove this check (we don't do this elsewhere)

storage/src/freezer/storage.rs
```diff
@@ -664,18 +682,22 @@ impl<E: BufferPooler + Context, K: Array, V: CodecShared> Inner<E, K, V> {
         let oversized_cfg = OversizedConfig {
             index_partition: config.key_partition.clone(),
             value_partition: config.value_partition.clone(),
+            metadata_partition: config.metadata_partition,
```
patrick-ogrady: agree that metadata_partition is the right name for this (as it covers both) 👍

storage/src/journal/segmented/oversized.rs
```diff
@@ -86,6 +100,12 @@ pub struct Config<C> {
     /// Partition for the glob value storage.
     pub value_partition: String,
 
+    /// Partition for durable recovery watermarks.
+    ///
+    /// Checkpoint mode opens and clears this partition before relying exclusively on its
+    /// externally published checkpoint.
+    pub metadata_partition: String,
```
patrick-ogrady: I think this should be first?

storage/src/journal/segmented/oversized.rs
```diff
@@ -86,6 +100,12 @@ pub struct Config<C> {
     /// Partition for the glob value storage.
     pub value_partition: String,
 
+    /// Partition for durable recovery watermarks.
+    ///
+    /// Checkpoint mode opens and clears this partition before relying exclusively on its
```
patrick-ogrady: I think we could switch oversize to only accept a checkpoint (like freezer/ordinal) and then just store metadata in archive (like we do for immutable)?

storage/src/journal/segmented/oversized.rs
```diff
@@ -486,52 +629,147 @@ impl<E: BufferPooler + Storage + Metrics, I: Record + Send + Sync, V: CodecShare
...
         .await?;
         self.index = index;
         self.values = values;
+
+        let completion: SyncCompletion =
+            async move { try_join(index_handle, values_handle).await.map(|_| ()) }
+                .boxed()
+                .shared();
+
+        let watermark_handle = if self.track_watermarks {
```
patrick-ogrady: this sort of switching is why we manage this explicitly in immutable

runtime/src/utils/buffer/paged/read.rs
```diff
@@ -356,6 +380,7 @@ impl<B: Blob> Replay<B> {
 
         self.buffer.clear();
         self.exhausted = false;
+        self.reader.deferred_invalid_checksum = false;
```
patrick-ogrady, line 414: interesting edge case

storage/src/archive/immutable/mod.rs
```diff
@@ -112,6 +117,9 @@ pub struct Config<C> {
     /// The compression level to use for the archive's freezer values.
     pub freezer_value_compression: Option<u8>,
 
+    /// The partition to use for the archive's freezer recovery-watermark metadata.
+    pub freezer_metadata_partition: String,
```
patrick-ogrady: And then we'd be able to use the existing metadata partition for immutable.

patrick-ogrady:
Latest review turned up this...playing around with different solutions:

```
• High — Make root-mismatch discard recoverable across both components. At engine.rs:758 ↗ /Users/patrickogrady/code/monorepo-pr-4327/storage/src/qmdb/sync/engine.rs:758, cleanup destroys the operation journal
before its Merkle projection. If execution stops between those resets, retrying the same target refills the empty journal while retaining the stale, equal-sized Merkle tree. Alignment then compares only leaf
counts and the valid retry reports RootMismatch again. Current’s same-target path explicitly skips reset at current/sync/mod.rs:303 ↗
/Users/patrickogrady/code/monorepo-pr-4327/storage/src/qmdb/current/sync/mod.rs:303; the other QMDB variants have no-op preparation. Persist a discard phase before either reset and complete it in prepare_sync
before reopening/refilling the journal.
```

roberto-bayardo:
subsumed by #4546

PR #4328, opened 2026-07-24
[consensus/simplex] Improve Replay Handling for Equivocation

patrick-ogrady:


consensus/src/simplex/actors/voter/slot.rs
```diff
@@ -122,16 +131,42 @@ where
...
-    pub fn update(&mut self, proposal: &Proposal<D>, recovered: bool) -> Change<D> {
-        // Once we detect equivocation we refuse to record any additional
-        // proposals, even if they target the original payload.
+    /// Records a proposal observed via a vote.
+    ///
+    /// A conflicting vote never replaces the recorded proposal, but marks the
+    /// slot [Status::Equivocated]. Once equivocation is recorded, votes are
+    /// ignored entirely (returning `None`), even if they target the recorded
+    /// payload.
+    pub fn update_vote(&mut self, proposal: &Proposal<D>) -> Option<Change<D>> {
```
BrendanChou, line 140: I don't like the introduction of `Option` by getting rid of the `Skipped` enum, I think it makes handling more messy and confusing personally

BrendanChou (approved):

consensus/src/simplex/actors/voter/slot.rs
patrick-ogrady: I, unfortunately, think this is the least bad option (rather than doing multiple enums or allowing the update_certificate function to return skipped)

Issue #3985, opened 2026-06-08
[consensus/simplex][medium] higher-view notarization can amplify lower-view resolver fetches


dnkolegov-ar:
## Summary

The Simplex resolver fetches certificates needed to repair missing consensus views. A lower-view request can be satisfied by a certificate at the requested view or by a notarization/finalization at a higher view.
When a notarization is received, local certification is asynchronous: the resolver accepts and records the notarization as pending, reports it to the voter, and later receives a certification result.

A malicious resolver peer with one valid high-view notarization can answer many lower-view repair fetches with that same certificate. The resolver accepts each response because the notarization view is not below the requested view, records each lower request under the same higher notarization, and continues advancing the missing-view fetch window while certification for the higher view remains pending.

## Impact

One valid high-view certificate can force repeated certificate decode/verification, voter `resolved` notifications, p2p fetch delivery cycles, and `satisfied_by` state growth proportional to the view gap.
If certification later fails, the resolver re-fetches every lower request recorded under that failed notarization, further amplifying repair traffic and resolver work.

So, a malicious resolver peer can reuse one valid far-ahead notarization to make a lagging Simplex node spend CPU, resolver fetch capacity, voter notifications, and retry work proportional to the missing view gap, slowing or stalling catch-up repair.

## Root Cause

The resolver treats a higher-view notarization as satisfying each lower-view request without bounding or collapsing how many request views one pending notarization can satisfy before certification completes. `fetch_concurrent` limits only simultaneously outstanding fetches; it does not bound the total lower-view requests satisfied by one high-view notarization over time.

  ## Code

  - https://github.com/commonwarexyz/monorepo/blob/d19ca3790566987b641bc78798fb5ecf477818f1/consensus/src/simplex/actors/resolver/actor.rs#L151-L185 accepts notarization responses when `notarization.view() >= requested_view`.
  - https://github.com/commonwarexyz/monorepo/blob/d19ca3790566987b641bc78798fb5ecf477818f1/consensus/src/simplex/actors/resolver/state.rs#L30-L33 stores lower request views in `satisfied_by` under the higher notarization view.
  - https://github.com/commonwarexyz/monorepo/blob/d19ca3790566987b641bc78798fb5ecf477818f1/consensus/src/simplex/actors/resolver/state.rs#L81-L88 inserts each satisfied request into that per-notarization set.
[...]

PR #4332, opened 2026-07-24
[consensus/simplex] Replace "Floor" Broadcast

patrick-ogrady:
This replaces Simplex's unconditional leader floor broadcast on nullification with proposal-driven, targeted ancestry repair. A validator that cannot verify a proposal because ancestry is missing can now ask that proposal's leader directly, potentially repairing the current proposal instead of waiting up to O(n) views for the right validator to lead.

## Problem

Background repair cannot detect every certificate split. One honest participant can certify a notarization at view `v` while others hold a covering nullification. Both sides consider `v` complete, but choose different parents and cannot verify each other's proposals.

Previously, this healed only after the certified-notarization holder became leader and then nullified, triggering an all-peer floor broadcast.

## Targeted ancestry repair

Proposal verification now requests missing ancestry directly from the proposal leader:

- an uncovered skipped view requests its covering nullification; and
- an uncertified named parent requests that parent view.

The request reuses the existing view key and kindless `Certificate` response, so the wire format does not change. Targeted and background demand share resolver state and remain deduplicated by view. An unrestricted background fetch continues to dominate a targeted fetch for the same key.

A responder returns its preferred ancestry: a certified notarization or finalization floor that covers the requested view, otherwise a covering nullification. The old nullification-time floor broadcast is removed.

Certificate deliveries separately track whether the resolver already knows the certificate and whether it may be re-gossiped. A targeted-only response is not immediately broadcast, while later background demand can still enable normal gossip. Replayed journal artifacts remain marked as already published, avoiding a restart-time broadcast burst.

## Certification and resolver lifecycle

A child notarization does not infer an application certification verdict for its parent. Only the application's deterministic result certifies or rejects a request. Certification may remain pending indefinitely; view timeout, nullification, and resolver-delivery expiry neither resolve nor cancel it. A covering finalization may instead make the local request unnecessary, and unrelated certification work continues.

Because the response key is kindless, valid evidence may not repair the proposal that requested it. The voter records valid evidence and reruns the complete ancestry check before voting. Malformed, key-incompatible, and cryptographically invalid evidence is rejected.

A fetched notarization can be cryptographically valid while local application certification is still pending. The resolver holds that delivery for at most the existing `certification_timeout`. If its purposes are still unresolved, it reports the response as valid but `Incomplete`, releases the delivery, and retries without faulting the peer. The application certification continues independently.

Local fetch purposes govern retention. A nullification retires matching demand within its term, a terminal certification verdict retires parent demand at that exact view, and finalization retires all covered demand. A higher certified floor retires background repair but does not incorrectly retire targeted parent demand. Tombstones prevent delayed actor messages from recreating completed work.

## Subscription and execution lifecycle

Marshal now separates remote acquisition from caller-owned subscriptions. Processing floors, finalization floors, and shard pruning can stop obsolete network work without canceling a caller that is still waiting for data that may arrive locally.

When finalization makes a certification request unnecessary, nested stateful verification becomes abandoned. Stateful retains useful abandoned work only until newer actor demand supersedes it, so live callers are never canceled merely because other work arrives.

Stateful also coalesces verification-owned lazy recovery by block digest. Overlapping certification requests share each in-progress `Application::apply` for common ancestors. If an owner is canceled, a live waiter retries and the remaining waiters share its completed cache entry. Finalization and pruning drain replay flights before database mutation, then restart only live requests against the new committed state; work already processed by finalization short-circuits without another application replay.

Proposal recovery deliberately remains outside verifier replay flights because proposals are awaited inline by the actor that polls verifier work.

## API and compatibility

`commonware-resolver::Consumer` now has an `Outcome` associated type. The new `DeliveryOutcome` distinguishes `Invalid`, `Complete`, and valid-but-`Incomplete` responses; consumers that only need the old boolean behavior can use `Outcome = bool`.

There are no new dependencies, namespaces, encoded types, wire-format changes, or storage-format changes.

Targeted repair normally costs a request and response instead of one broadcast delay, but avoids all-peer fanout and can rescue the triggering proposal if it completes before timeout. Distinct targeted keys can accumulate while finalization is stalled; higher-view resolver amplification remains tracked by #3985.

Deterministic coverage exercises both certificate-split orientations, invalid and wrong-kind responses, indefinitely pending certification, incomplete-delivery retry, cancellation and finalization restart, subscription retention, targeted gossip behavior, and overlapping replay handoff after owner cancellation.


consensus/src/lib.rs
patrick-ogrady, line 155: This is getting pretty Simplex specific ... I suspect we'll eventually move all these traits into simplex at some point in the medium-term.

PR #4339, opened 2026-07-27
[storage] Torn Repair Nits

patrick-ogrady:
## Summary

Follow-up to #4327 and #4326 after a storage-wide crash-recovery and cancellation-safety audit.

This changes the deterministic storage fault model from prefix-only persistence to power-loss behavior where any subset of unsynced write bytes may survive. Write extents and resizes survive independently, and failed writes use the same sporadic byte selection, including empty and complete subsets.

The audit also adds or expands recovery fuzzing for every recovery-sensitive storage family changed by the stacked PR. Generated inputs exposed one production defect already present on pre-PR `main` and one production regression introduced during development of the stacked PR. The remaining production fixes came from static protocol review and targeted cancellation tests; fuzz-model and oracle defects are listed separately below.

## Crash model

- Each pending write can be lost, retained completely, torn at a strict prefix, or retain an arbitrary per-byte subset.
- File extent growth and resize persistence are sampled independently from payload bytes.
- A completed range-durable write does not accidentally promote unrelated dirty writes.
- Storage epochs and blob generations invalidate pre-crash and removed/reopened handles.
- Ordinary `Runner::start` remains an ordinary stop; only `start_and_recover` crosses a simulated crash boundary.
- Every crash-capable target seeds the deterministic runtime with a dedicated 32-byte `FuzzRng` field, so corpus mutations explore scheduling, storage faults, and crash persistence instead of replaying one fixed sequence.

## Fuzz coverage

| Target | Recovery exercised |
| --- | --- |
| `archive_crash_recovery` | Immutable and Prunable archives, blocking and background sync, prune, interrupted destroy |
| `freezer_operations` | Freezer checkpoints, historical cursors, prune/restart, interrupted destroy |
| `journal_crash_recovery` | Contiguous fixed and variable journals, repeated recover/mutate/recover cycles |
| `segmented_journal_crash_recovery` | Segmented fixed and variable journals, per-section sync, interior holes, repair, prune/destroy |
| `oversized_recovery` | Organic torn recovery plus explicit index/value corruption, checkpoint floors, rewind/prune/destroy |
| `queue_crash_recovery` | Append/commit/enqueue durability, shortened recovery, prune, destroy |
| `merkle_full_crash_recovery` | Persisted Full MMR and MMB, flush/sync/prune, reset and destroy recovery |
| `current_crash_recovery` | Current QMDB over MMR and MMB, authenticated recovery, state-sync replacement, destroy |

Crash markers create repeated cycles in the same input. The targets also force lifecycle restarts after recovery, after a synced sentinel, and around real prune/destroy futures. Recovered state becomes the next boot's concrete oracle rather than restarting from the same initial state.

### Already present on pre-PR main

The campaigns found one distinct pre-existing production defect, through two independently generated failures:

- **Fixed contiguous recovery aligned an untrusted tail before repairing an earlier torn page.** A legal crash image could contain an invalid page followed by a checksum-valid island. `Writer::new` sized the blob from the later island; fixed-journal initialization then rounded that apparent tail to an item boundary and tried to shrink through the invalid page before its forward hole scan ran. Reopen returned `InvalidChecksum` indefinitely instead of recovering the intact whole-item prefix. `journal_crash_recovery` found this directly, and `queue_crash_recovery` independently found the same defect through Queue's nested fixed offsets journal. Recovery now finds and durably truncates to the contiguous valid page prefix before applying item alignment.

### Introduced during the stacked PR and caught before merge

- **Oversized exact-checkpoint restore validated a committed entry before removing a later unacknowledged page island.** `archive_crash_recovery` produced a page that correctly fell back to its shorter committed checksum while a later page survived. The committed page was therefore temporarily partial and non-final, so the positional validation failed even though the acknowledged bytes were intact. Exact restore now repairs the index prefix before reading committed entries. This intermediate ordering was not present on pre-PR `main`.

No other production fix below is attributed to a failing generated fuzz input.

## Additional production audit findings fixed

These findings came from the broader crash/cancellation protocol review and targeted regressions.

### Runtime and buffers

- Paged writes could promote a visible but unsynced checksum slot and later overwrite the last completed-sync fallback. Visible and synced partial-page states are now tracked separately, and only a completed barrier changes the protected slot.
- The paged-writer fix covers same-runtime snapshot/replay/plain-flush chains and cancellation; it does not depend on Tokio omitting its startup fsync.
- Split flush cancellation now republishes no logical/cache state until every physical write completes, so retry regenerates the complete write plan.
- Direct writes, resizes, and shrinks poison a writer when cancellation may have left non-retryable physical state. Truncated space is made durable before reuse.
- Reopened invalid inactive checksum slots are retired safely, including CRC-zero payload cases that make a fixed zero tombstone unsafe.
- Paged replay no longer truncates physical lengths through `usize` on wasm32, and seeking past a recovered partial prefix cannot underflow or expose stale bytes.

### Storage recovery and lifecycle

- Segmented fixed/variable replay now preserves acknowledged bytes when a page falls back to its shorter checksum slot, repairs only unacknowledged suffixes, and treats mutable repair failures as terminal.
- Oversized exact restore rejects unaligned checkpoints before mutation, checks sequential section presence and terminal index/value boundaries, detects a missing lower section even when both index and value blobs are absent, and repairs unusable index holes before positional reads.
- Oversized rewind validation is separated from mutation, and prune/destroy remove index references before values so interruption remains recoverable.
- Contiguous and variable prune paths make the entire live tip durable before durable section removal.
- Composite destroy tests cancel the actual production future between awaits, reopen, and retry destroy. Reopen may expose the old state, another valid state, or empty storage; only openability and convergence are required.
- Contiguous journals, persisted Merkle, and Current QMDB use durable reset/replacement intents so recovery finishes any staged reset after an interrupted clear or destroy, leaving destroy safely retryable. An interrupted Current QMDB state sync is safely re-driven from a compatible journal; ordinary database reopen or an incompatible target instead discards the unpublished generation (converging to empty), in all cases without interpreting mixed generations.
- Current's `prepare_sync` hook is mandatory: it stages database-specific replacement intent before the shared sync engine opens or mutates its journal.
- Prunable checkpoint publication retains failed background-sync errors and never advances durable sizes ahead of the joint index/value sync.

## Fuzz-model and oracle findings fixed

- The in-memory crash model assumed surviving dirty data was a prefix. It now preserves arbitrary subsets while respecting completed syncs.
- Failed-write byte selection could not represent writing nothing. Empty, complete, prefix, and sparse subsets are all reachable.
- Stale handles could race crash or remove/reopen bookkeeping. Epoch and generation validation now serializes with the pending mutation log.
- Oversized corruption allowances were global and could excuse a failure in an untouched section. Exemptions are now attributed to the affected section and byte ranges.
- Queue recovery retained modeled values beyond a newly committed shortened size. Committing now truncates the oracle before pinning durability.
- Contiguous failed-prune recovery previously allowed impossible Cartesian combinations of floor, size, and content. It now permits only the unchanged-floor crash range, or for variable journals, the exact target floor with the fully synced pre-prune tip.
- Redundant size/floor fields and duplicate verification passes were removed so oracle state is derived from the values it constrains.

## Fuzz results

Across the iterative audit campaigns:

- **1,027,111** sanitizer executions total.
- **156,087** executions on the latest cleanup/oracle revision.
- No crashes or assertion failures after the final fixes.

Fuzz-only line coverage of the instrumented production paths:

| Scope | Line coverage |
| --- | ---: |
| All changed instrumented production code | 68.44% |
| Targeted recovery production code | 75.93% |
| Deterministic memory storage | 90.13% |
| Faulty storage wrapper | 86.60% |
| Paged writer | 95.33% |
| Unpaged writer | 89.25% |

Fuzzing is not a proof of crash consistency, but the combination of reachable arbitrary subsets, repeated restarts, correlated oracles, targeted regressions, and independent protocol review gives substantially stronger evidence than repeatedly exercising one prefix-loss scenario.

## Validation

- `just pre-pr`: **7,113/7,113** selected tests passed.
- Workspace formatting, clippy, rustdoc, dylints, stability, and public-API checks passed.
- All changed storage fuzz binaries pass `cargo check` and strict clippy.
- All eight production interrupted-destroy recovery tests pass together.
- LLVM 22 wasm32 release builds pass for `utils`, `runtime`, and `storage`.


roberto-bayardo:
## `validate_index_prefix` makes freezer restart O(total index entries), where it was O(sections)

The freezer's only recovery path is `Oversized::init_from_checkpoint → restore`, which runs on **every restart**. `restore` now calls `validate_index_prefix` on the checkpoint section and `verify_complete` (→ `validate_index_prefix`) on **every below-checkpoint section**, and `validate_index_prefix` reads + CRC-checks *every* entry (`for position in 0..entry_count`). Previously `verify_complete`/rewind read only the last entry (O(1)/section). So a large freezer now reads its entire index at startup, which partly defeats the checkpoint's fast-restart purpose. (The `init → repair` crash path is unaffected — it only validates above each durable floor.)

The full scan doesn't look crash-necessary:

- `verify_complete`'s actual check is `last_entry.value_end == glob_size`, and `validate_index_prefix` only *uses* the last entry's `value_location` — the interior reads add nothing to it.
- Below-checkpoint sections are fully synced, and a synced page can't tear, so an interior unreadable entry there is only external (bit-rot) corruption — out of scope per the "CRC is an integrity check, not authentication" principle, and caught lazily on the next read regardless.
- `repair_prefix` already handles the crash-reachable case (the fallback partial page) independently.

**Suggested:** drop `validate_index_prefix` to reading just the last entry (O(1)/section, back to O(sections) at restart) while keeping `repair_prefix` + the boundary/glob checks — this preserves the fix for the exact-checkpoint fallback-page regression and all crash-model correctness. If the deeper interior check is wanted, gating it behind an explicit "deep verify" mode rather than the default restart path would avoid the startup cost.

roberto-bayardo:
One small **doc-clarity** nit on the PR description (not the code). Under "Storage recovery and lifecycle":

> "…interrupted clear, state sync, or destroy **completes on restart** instead of interpreting mixed generations."

"completes on restart" reads right for clear/destroy, but for **state sync** it's a bit misleading: an interrupted sync doesn't *complete* — on reopen the unpublished generation is **discarded (reset to empty)** and the sync is re-driven from there, rather than resumed/finished. Might be worth splitting it, e.g.:

> "…an interrupted clear or destroy completes on restart; an interrupted **state sync** discards its unpublished generation (converging to empty) so it can be safely re-driven — in all cases without interpreting mixed generations."

Just so a reader doesn't expect a half-finished sync to be resumed.

(Otherwise re-reviewed the condense/refinement pass end-to-end — no correctness regressions; the reworked qmdb sync-replacement and merkle reset are crash-safe.)

PR #4347, opened 2026-07-28
[consensus/simplex] let batcher adopt proposals from notarizations

andresilva:
The batcher previously verified notarize and finalize votes only after learning the leader's proposal from the leader's notarize vote. If that vote was missed but a valid notarization was received, an already-buffered finalize quorum could remain unverified.

This PR updates the batcher to track the leader and proposal independently. A verified notarization now establishes the authoritative proposal and causes buffered votes for that proposal to be reconsidered, even when the round's leader is not yet known. Finalizations retain their existing behavior.

### DIFF
--- a/consensus/src/simplex/actors/batcher/actor.rs
+++ b/consensus/src/simplex/actors/batcher/actor.rs
@@ -496,10 +496,19 @@ where
                         }
 
                         // Store and forward to voter
-                        work.entry(view)
+                        let proposal_changed = work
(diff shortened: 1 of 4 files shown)


danlaine:
@andresilva Suggested some changes in #4349. I think it looks a bit cleaner that way but I don't feel super strongly.

patrick-ogrady (approved):

PR #4349, opened 2026-07-28
[consensus/simplex] Track the batcher proposal's source

danlaine:
The verifier now tracks the round's proposal as `ProposalSource::Leader(proposal)` or `ProposalSource::Notarization(proposal)` instead of bare `Option<Proposal>`.

`Round` records all certificates through one method again. `record_certificate(&Certificate)` adopts the proposal from a notarization internally and returns whether it changed, which keeps the actor's three certificate arms identical after verification: record, forward to the voter, and reprocess buffered votes only if the proposal changed. Invalid certificates block with one reason ("invalid certificate" with the kind as a field) instead of three.

### DIFF
--- a/consensus/src/simplex/actors/batcher/actor.rs
+++ b/consensus/src/simplex/actors/batcher/actor.rs
@@ -483,74 +483,42 @@ where
                 );
                 let _guard = span.entered();
 
-                match message {
-                    Certificate::Notarization(notarization) => {
-                        // Verify the certificate
-                        if !notarization.verify(
-                            self.context.as_mut(),
-                            self.scheme.as_ref(),
-                            &self.strategy,
-                        ) {
-                            commonware_p2p::block!(self.blocker, sender, %view, "invalid notarization");
-                            continue;
-                        }
-
-                        // Store and forward to voter
-                        let proposal_changed = work
-                            .entry(view)
-                            .or_insert_with(|| self.new_round(view))
-                            .record_notarization(&notarization.proposal);
-                        voter.recovered(Certificate::Notarization(notarization));
-
-                        // Reprocess buffered votes only if the notarization changed the
-                        // proposal filter.
-                        if !proposal_changed {
-                            continue;
-                        }
-
-                        updated_view = view;
-                    }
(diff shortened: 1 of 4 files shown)


consensus/src/simplex/actors/batcher/verifier.rs
```diff
@@ -187,7 +204,7 @@ pub struct Verifier<S: Scheme<D>, D: Digest> {
     /// A known leader's notarize vote may supply the initial proposal. A
     /// verified notarization is authoritative and may replace it, or establish
     /// the proposal before the leader is identified.
-    proposal: Option<Proposal<D>>,
+    proposal: Option<ProposalSource<D>>,
```
andresilva: I wonder if the `Option` should be folded into the enum above? 

```rust
enum ProposalState<D> {
  Unknown,
  Leader(Proposal<D>),
  Notarization(Proposal<D>),
}
```

consensus/src/simplex/actors/batcher/actor.rs
```diff
@@ -483,74 +483,42 @@ where
...
-                        work.entry(view)
-                            .or_insert_with(|| self.new_round(view))
-                            .record_finalization();
-                        voter.recovered(Certificate::Finalization(finalization));
+                // Store and forward to voter
+                let proposal_changed = work
+                    .entry(view)
+                    .or_insert_with(|| self.new_round(view))
+                    .record_certificate(&message);
+                voter.recovered(message);
```
andresilva, line 514: This is definitely cleaner, although it makes it less obvious that only notarizations can update the proposal. But I think I'm OK going with this. It should be documented in `record_certificate`.

consensus/src/simplex/actors/batcher/verifier.rs
danlaine: done

consensus/src/simplex/actors/batcher/actor.rs
danlaine: > It should be documented in record_certificate.

Done

andresilva (approved):

Issue #4353, opened 2026-07-28
simplex/batcher: serialized crypto awaits in process_view cap the pipeline at ~180 views/s


BrendanChou:
## Problem

`process_view` in `consensus/src/simplex/actors/batcher/actor.rs` awaits its crypto inline, one operation at a time, on the actor's own loop:

```rust
loop {
    let Some((batch, failed)) = round.try_verify(...).await else { break };  // ~2.20ms mean
    ...
}
while let Some(certificate) = ... round.try_construct_certificate(...).await { ... }  // ~0.55ms mean
```

Measured on a 10-node testnet at 175 views/s (`TARGET_BLOCK_INTERVAL_MS = 5`), per node:

- verify: 335.6 batches/s × 2.20ms = **73.7% of the batcher's wall clock**
- recover: 335.6/s × 0.55ms = **18.5%**
- combined: **~92% occupancy** of a single actor loop, serially awaiting crypto

335/s ≈ 2 batches per view (notarize quorum + finalize quorum), so each view costs the batcher **~5.5ms of serialized awaits** — which makes the whole consensus pipeline top out at ~180 views/s regardless of CPU headroom (the signature pool is only ~63% loaded; the parallelism exists but the actor consumes it one batch at a time).

## Observed consequences (this cap is the root cause of two symptoms)

1. **Mid-term interval degradation.** The leader produces at 200 views/s but notarizations are observed at ~180/s, so the optimistic issuance window (`TERM_OPTIMISTIC_VIEWS = 100`) fills at ~20 views/s of deficit and pegs around view ~550 of each 1000-view term. From there, every proposal waits for the next notarization to slide the frontier: views 1–550 of a term hit the 5ms target 99.8–100% of the time; views 550–1000 degrade to 66–72% (mean interval 5.5–5.7ms — i.e., slaved to the batcher's cadence).
2. **Term-boundary stall.** The ~100 views of certification debt accumulated per term drain serially at the term start: a ~500ms stall every term (≈ window × interval), ~72% of all interval excess.

Lowering `TERM_OPTIMISTIC_VIEWS` to 40 was A/B tested and is a net loss (pegging just starts at view ~220); the window size is not the problem — pipeline throughput is.

## Proposed fix

[...]

PR #4354, opened 2026-07-28
[consensus/simplex] Overlap proposal construction with voter persistence

### LINKED ISSUE #4352: simplex/voter: dispatch next proposal before construct/sync/notify
## Problem

When the leader's proposal for view `v` resolves, the voter's event loop does all of the following **before** requesting the proposal for `v+1`:

1. `construct` — build and BLS-sign the notarize vote for `v` (measured 0.4–1.2ms when votes are built)
2. `sync_journal` — fsync (~0.23ms)
3. `notify` — broadcast
4. wrap around to the next `select_loop!` iteration's `on_start`, which finally calls `try_propose`

None of that work is a dependency of the next proposal — it only needs state already updated when the branch processed the resolved proposal. The result is that the `v+1` propose request reaches the automaton ~1.4–1.6ms after `v` was stamped, burning ~30% of a 5ms block-interval budget on serialization before the pacing sleep can even start.

Measured on a 10-node geo-distributed testnet (`TARGET_BLOCK_INTERVAL_MS = 5`): ~15–19% of proposals entered the automaton with no remaining slack (past their deadline), directly producing the 6ms interval bucket (~10% of blocks).

## Fix

At the top of `on_end` in `consensus/src/simplex/actors/voter/actor.rs`, dispatch `try_propose` before construct/sync/notify:

```rust
on_end => {
    if pending_propose.is_none() {
        pending_propose = self.try_propose().await;
    }
    // ... existing construct / sync_journal / notify ...
}
```

The dispatch is a cheap mailbox send; the automaton's pacing sleep then runs concurrently with the signing/fsync/broadcast instead of queueing behind them.

## Results (A/B on the same cluster, 40k blocks each)

- proposals now arrive with ~2.7ms median slack (was ~1.6ms consumed)
- no-sleep (late) entries: ~15–19% → ~10%
- exactly-5ms intervals: 87.3% → 88.8%; 6ms bucket 10.2% → 8.1%
- with this in place, the **front half of every term runs at 99.8–100% exactly-5ms** — the remaining misses are the batcher pipeline (separate issue), not the voter

All 655 simplex tests pass with the change. Working prototype: commit `e52b0153` on local branch `bc/instr-propose-wait` (includes instrumentation commits it can be rebased off of).

BrendanChou:
Closes #4352.

Dispatch the next `try_propose` at the top of `on_end`, before this iteration's construct/sync/notify. The next proposal only depends on state the select branch already updated (e.g. the recorded proposal), and the automaton paces the returned build to the target block interval — dispatching first lets that pacing sleep run concurrently with the notarize signing (0.4–1.2ms) and journal fsync (~0.23ms) instead of queueing the next block behind them.

Measured on a 10-node geo-distributed testnet at `TARGET_BLOCK_INTERVAL_MS = 5` (40k blocks per arm, same cluster):

- proposals arrive with ~2.7ms median slack (previously ~1.6ms of the 5ms budget was consumed before dispatch)
- proposals entering past their deadline: ~15–19% → ~10%
- exactly-on-target intervals: 87.3% → 88.8%; the 6ms bucket: 10.2% → 8.1%
- with this change, the front half of every 1000-view term runs at 99.8–100% exactly-5ms; the residual misses trace to batcher pipeline throughput (#4353)

Full simplex suite passes (599 tests).


consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -1221,6 +1221,18 @@ impl<
...
             on_end => {
+                // Attempt the next proposal before constructing, syncing, and
+                // broadcasting this iteration's artifacts. Anything the branch
+                // above did that unblocks a proposal (recording our own proposal,
+                // handling a certificate that advances the view) is already
+                // reflected in state, and the automaton paces the returned build
+                // to the target block interval — dispatching the request now lets
+                // that pacing sleep absorb the signing and journal sync below
+                // instead of queueing the next block behind them.
+                if pending_propose.is_none() {
```
patrick-ogrady, line 1215: So, we'd now propose at start and end?

BrendanChou:
Superseded by #4448, which ports this change to the rewritten voter loop in #4447 (this version would be silently reverted when that branch merges).

PR #4356, opened 2026-07-28
[consensus/simplex] Overlap batcher crypto via dispatch pool instead of inline awaits

### LINKED ISSUE #4353: simplex/batcher: serialized crypto awaits in process_view cap the pipeline at ~180 views/s

BrendanChou:
Closes #4353.

## Problem

The batcher actor awaits signature verification and certificate recovery inline in its event loop. On a loaded network this serializes ~2.2ms batch-verifies and ~0.55ms recoveries back-to-back per view (~92% actor occupancy at 335 batches/s), during which the actor cannot ingest votes, verify another vote kind, or construct a ready certificate. On `bc/optimistic-validation` this caps certification throughput, which pegs the optimistic issuance window and slaves block production in the back half of every term to notarization cadence.

## Change

Split `Certification::try_verify` into a sync `begin_verify` (takes the pending batch, marks the kind in-flight) and `finish_verify` (reintegrates results); same split for certificate recovery. The actor pushes the owned offload futures into a `commonware_utils::futures::Pool` drained by a new select branch, so crypto overlaps with message ingestion and with crypto for other views/kinds.

Notes:

- `finish_verify` re-filters notarizes/finalizes against the current leader proposal, since a certificate can displace the proposal while a batch is in flight — stale-proposal votes must not count toward a quorum.
- A completion for a view pruned mid-flight is dropped.
- The previous `try_verify`/`try_construct_certificate` entry points are kept as `#[cfg(test)]` shims implemented over begin/finish, so existing tests run unchanged.

## Results

A/B on an 11-validator global testnet (10 regions) (`TARGET_BLOCK_INTERVAL_MS = 5`, term length 1000, optimistic window 100), ~40k blocks per sample:

| | before | after |
|---|---|---|
| blocks/s | 175.5 | **192.4** |
| exactly-5ms intervals | 88.8% | **99.46%** |
| 6ms bucket | 8.1% | **0.23%** |
| term-boundary stall | ~500ms | **186ms** |
| mean excess per view | 0.49ms | **0.18ms** |

Exactly-5ms adherence is flat at 99.4–99.8% across all term-position deciles (previously degraded to 66–72% in the back half as the issuance window pegged).

The same inline-await structure exists in main's batcher (`actor.rs` `try_verify(...).await` / `try_construct_certificate(...).await`); there it costs view latency rather than throughput. Porting this to main is worthwhile follow-up, but main's batcher has diverged, so it would be a separate change.

### DIFF
--- a/consensus/fuzz/fuzz_targets/simplex_elector.rs
+++ b/consensus/fuzz/fuzz_targets/simplex_elector.rs
@@ -6,7 +6,7 @@ use commonware_consensus::{
         elector::{self, Elector, Random, RoundRobin},
         scheme::{bls12381_threshold::vrf as bls12381_threshold_vrf, ed25519},
     },
-    types::{Round, TermLength, View},
+    types::{Round, TermLength, View, ViewDelta},
 };
 use commonware_cryptography::{
     Sha256, Signer,
@@ -72,17 +72,22 @@ fuzz_target!(|input: FuzzInput| {
         FuzzElector::RoundRobin(term_length) => {
             let elector = match term_length.get() {
                 1 => RoundRobin::<Sha256>::default(),
-                _ => {
-                    RoundRobin::<Sha256>::default().with_term(*term_length, Duration::from_secs(12))
-                }
+                _ => RoundRobin::<Sha256>::default().with_term(
+                    *term_length,
+                    Duration::from_secs(12),
+                    ViewDelta::new(0),
+                ),
             };
             fuzz::<ed25519::Scheme, _>(&input, elector, None);
         }
         FuzzElector::RoundRobinShuffled(seed, term_length) => {
             let elector = match term_length.get() {
                 1 => RoundRobin::<Sha256>::shuffled(seed),
-                _ => RoundRobin::<Sha256>::shuffled(seed)
-                    .with_term(*term_length, Duration::from_secs(12)),
+                _ => RoundRobin::<Sha256>::shuffled(seed).with_term(
+                    *term_length,
+                    Duration::from_secs(12),
+                    ViewDelta::new(0),
+                ),
             };
             fuzz::<ed25519::Scheme, _>(&input, elector, None);
         }
--- a/consensus/fuzz/src/invariants.rs
+++ b/consensus/fuzz/src/invariants.rs
@@ -1,5 +1,5 @@
 use crate::{
-    bounds,
+    Configuration, bounds,
     simplex::Simplex,
     types::{Finalization, Notarization, Nullification, ReplicaState},
 };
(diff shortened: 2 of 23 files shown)


BrendanChou:
(Resolved: the wasm32 `Lookahead` fix landed directly on `bc/optimistic-validation` as c3f4b51c1 — no longer stacked on this PR.)

BrendanChou:
Superseded by #4447, which rebuilds this change on the post-#4443 verifier.

PR #4359, opened 2026-07-29
[consensus/simplex] Restore Finalizations

patrick-ogrady:
## Summary

Builds on #4347 by treating the first independently authenticated proposal as authoritative whether it comes from a verified notarization, a verified finalization, or a locally constructed finalize. This state may replace a proposal learned from the leader's vote and cannot be replaced by later conflicting evidence.

When a notarization replaces a conflicting leader-selected proposal, matching network finalize votes retained by the round are restored so they can be verified. Votes buffered before any proposal was selected remain in the verifier and need no replay.

## Testing

Adds deterministic coverage for finalizations and constructed finalizes establishing authoritative proposal state, and for restoring finalize votes received both before and after conflicting leader selection.

### DIFF
--- a/consensus/src/simplex/actors/batcher/round.rs
+++ b/consensus/src/simplex/actors/batcher/round.rs
@@ -30,8 +30,8 @@ pub struct Round<
     reporter: R,
     /// The verifier attempts to recover notarizations and finalizations only
     /// from votes for one proposal. It initially filters to the first proposal
-    /// observed from the known leader. A verified notarization is authoritative
-    /// and switches the filter to its proposal.
+    /// observed from the known leader. Independently authenticated proposal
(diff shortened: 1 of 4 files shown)


consensus/src/simplex/actors/batcher/round.rs
```diff
@@ -98,20 +98,50 @@ impl<
     /// Records a verified certificate and returns whether the round's proposal
     /// changed.
     ///
-    /// Only a notarization can change the proposal. Its proposal is
-    /// authoritative and replaces any conflicting proposal learned from the
-    /// leader's vote.
+    /// Only the first notarization can establish or replace the proposal. It may
```
patrick-ogrady: Why can't a finalization?

andresilva (approved):

PR #4360, opened 2026-07-29
[storage/qmdb] Serve sync from one `Source`

danlaine:
QMDB sync previously had two serving abstractions and two client protocols (replay sync and a separate compact-state protocol with its own client, validators, p2p actor, and wire messages). This PR collapses both: every database serves through a single `Source` trait, and compact sync becomes an ordinary replay sync over the one-operation range ending at the target. The compact client subsystem, its glue actor, and its wire messages are deleted.

## The `Source` trait

`Source` (renamed from the trait formerly called `Resolver`, which collided with `commonware-resolver`) answers typed requests with typed responses:

```rust
enum Request<F> {
    Operations { size, start, max_ops },
    Boundary { size, start },
}

enum Response<F, Op, D> {
    Operations { proof, operations },
    Boundary { proof, op, pinned_nodes },
}
```

A `Boundary` request fetches the single operation at `start` plus the pinned nodes at that boundary; the proof in the response authenticates the pinned nodes, so there is no way to request them on their own.

The trait is implemented once, on `authenticated::Journal`; databases forward to their journal (`SourceFor<DB>` names the source-matches-database bound everywhere it recurs). Wrappers compose (`Arc<T>`, locks, `Option<T>` contributing `MissingSource`), so e.g. `Arc<Lock<Option<Db>>>` can be composed. The engine holds one `Arc<Source>` shared across in-flight requests instead of deep-cloning the source per request, and the replay bound no longer requires `Clone`. A new `Floored` trait carries just the commit-floor lookup `Operation` bundled with key/value access, so journals and the witness can serve without caring what the operations are.

A pruned source refuses ranges it no longer retains with the same errors a pruned operation log reports (`ItemPruned`, `RangeOutOfBounds`). Full databases serve historical compact states; a compact-storage database serves exactly its latest committed state and refuses everything else.

## Compact sync is a case of full sync

The witness storage format changes to make this collapse possible. A witness entry now stores the encoded last-commit operation, the committed size, and the pinned nodes one operation below the commit. The stored inclusion proof is deleted: it is derivable, and storing it was what forced compact sync to be its own protocol. On open and rewind, the Merkle is rebuilt by appending the commit operation to the pinned nodes (`merkle::compact` gains `append_leaf` for this); corruption detection is journal checksums plus root recomputation, the same trust model as replay journals.

With that format, the compact fetch is exactly the replay engine's own boundary request over `[N-1, N)`, verified by the standard `verify_proof_and_pinned_nodes` path. Deleted outright: the `sync::compact` module (`compact::sync()`, `attempt_sync`, `compact::Database`, `validate_compact_state`/`validate_compact_frontier`, `ValidatedState`) and the glue compact actor/mailbox/handler with their wire rows. Compact databases implement `sync::Database` over an in-memory one-operation journal and share one `qmdb::compact::{Config, from_sync_result}` (previously duplicated per database), so target updates, retries, peer feedback, and batching configuration all come from the engine (compact syncs previously ignored `SyncEngineConfig`; they now honor it).

The compact target survives as `CompactTarget { root, size }`, living beside `Target` in `sync::target`. `Target::try_from(&CompactTarget)` is the single place the `[N-1, N)` range is derived; it rejects `size == 0`, which no committed state has.

The target root leaves the wire. A server cannot tell a stale target from a divergent one and no longer needs to: the client verifies every response against its own target root, and the security model is replay's.

## Retry policy: feedback or fail

`ValidityTx` is `Option<oneshot::Sender<bool>>`: `Some` when a source is wired to react on verdicts (block the peer, ask another, etc.), `None` when the answer is final. A response that fails validation is retried with negative feedback when the source is listening, and is a terminal `EngineError::InvalidResponse` when it is not, because a source that accepts no feedback cannot change its answer. Previously the engine retried unconditionally, which livelocked against any deterministic source serving unverifiable data. `EngineError::InvalidProof` had no remaining constructor and is deleted.

`handle_fetch_result` verifies per arm: a response must match the shape of its request, an `Operations` batch is bounded by the request's own `max_ops` (previously the looser `fetch_batch_size`), and boundary pinned nodes are consumed only when the current target still needs them. A new `Database::local_pinned_nodes` lets a client whose persisted journal already covers the target reuse its own pinned nodes instead of fetching them from a peer.

Target updates have one staleness policy: a non-advancing update (end not strictly larger, or start moved backward) is a stale hint and is discarded, whether it arrives mid-sync, sits queued when the engine reaches its target (drained before completion is reported, so callers never have a stale target reported as reached), or is filtered at the glue seam before the sync starts. Previously the same update could be filtered, discarded, or fatal depending on timing; `EngineError::SyncTargetMovedBackward` is deleted with the fatal path. The one remaining fatal case is an advancing update with an unchanged root, which is impossible for an append-only log and so indicates a caller bug. A requested finish still wins: once the caller asks to finish, queued updates are ignored and the sync completes at its current target.

## The enums are the wire format

Glue's `handler::Request`/`Response` wire types duplicated the storage types field for field; they are deleted, and the tagged codecs live on `sync::Request`/`sync::Response` directly (`Request` also gains the `Eq`/`Ord`/`Hash`/`Display`/`Span` bundle the resolver key needs; the request is its own identity in the request pool and on the wire).

Codec conformance moves accordingly: the glue handler rows are deleted, storage gains rows for `Request` and `Response` (both families), and the witness codec rows plus the compact `StorageConformance` rows regenerate for the deliberate format change. The old compact-request row survives byte-identical as the `CompactTarget` row, and `Target`'s row is unchanged.

## Glue

The p2p resolver actor pair collapses to one: the compact actor/mailbox/handler are deleted (~1,200 lines) and the surviving actor moves from `db::p2p::standard` up to `db::p2p`, serving both database kinds over the unified wire. The per-database `sync_db` bodies dedupe into `sync_standard_db` and `sync_compact_db`; the compact variant translates the target/update/reached channels between `CompactTarget` and the engine's ranged `Target` via the `TryFrom` above.

## Other changes

- Serving a size of zero reports `RangeOutOfBounds` (previously `HistoricalFloorPruned` on the full path; unreachable through the engine, and the wire rejects it at decode).
- Per-family serve tracing spans collapse into one `qmdb.sync.serve` span.
- `apply_batch_size` is clamped to at least 1 at engine construction (a zero value would make database construction spin without progress).
- Sync progress gauges register under `context.child("sync")` as `target_size`/`size` (previously `target_leaf_count`/`leaf_count`, unprefixed, which collided with the journal's `size` gauge).
- Naming: `sync::resolver` is now `sync::source`, `Error::Resolver` is `Error::Source`, `pins` is `pinned_nodes`, and `leaf_count` is `size`, matching the witness's stored field.


roberto-bayardo:
Couple nits from my fable session:
1. Retry policy is now delegated entirely to the Source via ValidityTx. Conditions that the old engine silently retried (bad batch size, leaf-count mismatch, proof failure) now go through reject_response, which is terminal (InvalidResponse) for a source that returns None validity. That's correct for direct/local sources, but it means the out-of-tree network adapter must return Some(validity) on every response or a single malformed/lagging peer aborts the whole sync. Since the network Source isn't in this PR, I'd want this contract stated explicitly in the Source/ValidityTx docs so a downstream adapter doesn't silently regress multi-peer robustness. (Serve errors were always terminal, so no change there.)
2. On-disk breaking change. The witness journal format changed, so a compact DB written before this PR won't reopen. Fine under ALPHA (no migration required), but worth a one-line note in the PR description.

roberto-bayardo:
Non-blocking cleanup: `rebuild` reimplements what `build_witness` now does.

`build_witness` was made `pub(crate)` so reconstruction can share it, and `init_from_sync` uses it that way (`reset_to` → `append_leaf` → `build_witness`). The reopen path in `rebuild` (`storage/src/qmdb/compact/witness.rs`) instead hand-rolls the same post-append computation:

```rust
merkle.reset_to(last_commit_loc, witness.pinned_nodes.clone())?;
merkle.append_leaf(&hasher, &witness.op_bytes)?;
let inactive_peaks = F::inactive_peaks(F::location_to_position(size), inactivity_floor_loc);
let root = merkle.root(&hasher, inactive_peaks)?;
let proof = merkle.with_mem(|mem| mem.proof(&hasher, last_commit_loc, inactive_peaks))?;
merkle.prune_to_frontier();
Ok((VerifiedWitness { witness, root, proof }, last_commit_op))
```

After the `append_leaf`, this could just call `build_witness(merkle, inactivity_floor_loc, witness.op_bytes)`, making `rebuild` structurally identical to `init_from_sync`. Two benefits:

- drops the duplicated `inactive_peaks`/`root`/`proof` logic, and
- collapses the two separate Merkle lock acquisitions (`root(...)` then `with_mem(proof)`) into `build_witness`'s single `with_mem`.

It's value-preserving: `build_witness` recomputes `size = mem.leaves()` and `pinned_nodes = nodes_to_pin(size - 1)`, which equal the stored values after `reset_to` + `append_leaf` — the same equivalence `init_from_sync` already relies on. The only change is that `rebuild` returns a freshly-built `VerifiedWitness` rather than the verbatim stored one, which is a wash.

danlaine:
> Couple nits from my fable session:
> 
> 1. Retry policy is now delegated entirely to the Source via ValidityTx. Conditions that the old engine silently retried (bad batch size, leaf-count mismatch, proof failure) now go through reject_response, which is terminal (InvalidResponse) for a source that returns None validity. That's correct for direct/local sources, but it means the out-of-tree network adapter must return Some(validity) on every response or a single malformed/lagging peer aborts the whole sync. Since the network Source isn't in this PR, I'd want this contract stated explicitly in the Source/ValidityTx docs so a downstream adapter doesn't silently regress multi-peer robustness. (Serve errors were always terminal, so no change there.)
> 2. On-disk breaking change. The witness journal format changed, so a compact DB written before this PR won't reopen. Fine under ALPHA (no migration required), but worth a one-line note in the PR description.

1. There aren't any good options for a sync engine whose `Source` gives it invalid data. In the best case, it tells the `Source`, via the feedback channel, that it got invalid data. If that's not possible, it can either keep re-trying to get (valid) data, and hope things improve, or pessimistically stop trying with an error. I opted for the latter because it avoids a re-try tight loop and absolves the need to have heuristics about backoffs, erroring after N failures, etc.

2. Yup this has the breaking format tag

danlaine:
> Non-blocking cleanup: `rebuild` reimplements what `build_witness` now does.
> 
> `build_witness` was made `pub(crate)` so reconstruction can share it, and `init_from_sync` uses it that way (`reset_to` → `append_leaf` → `build_witness`). The reopen path in `rebuild` (`storage/src/qmdb/compact/witness.rs`) instead hand-rolls the same post-append computation:
> 
> ```rust
> merkle.reset_to(last_commit_loc, witness.pinned_nodes.clone())?;
> merkle.append_leaf(&hasher, &witness.op_bytes)?;
> let inactive_peaks = F::inactive_peaks(F::location_to_position(size), inactivity_floor_loc);
> let root = merkle.root(&hasher, inactive_peaks)?;
> let proof = merkle.with_mem(|mem| mem.proof(&hasher, last_commit_loc, inactive_peaks))?;
> merkle.prune_to_frontier();
> Ok((VerifiedWitness { witness, root, proof }, last_commit_op))
> ```
> 
> After the `append_leaf`, this could just call `build_witness(merkle, inactivity_floor_loc, witness.op_bytes)`, making `rebuild` structurally identical to `init_from_sync`. Two benefits:
> 
> * drops the duplicated `inactive_peaks`/`root`/`proof` logic, and
> * collapses the two separate Merkle lock acquisitions (`root(...)` then `with_mem(proof)`) into `build_witness`'s single `with_mem`.
> 
> It's value-preserving: `build_witness` recomputes `size = mem.leaves()` and `pinned_nodes = nodes_to_pin(size - 1)`, which equal the stored values after `reset_to` + `append_leaf` — the same equivalence `init_from_sync` already relies on. The only change is that `rebuild` returns a freshly-built `VerifiedWitness` rather than the verbatim stored one, which is a wash.

Done in https://github.com/commonwarexyz/monorepo/pull/4360/commits/74e9903fac9559887193e96ad1483fa971894a05

patrick-ogrady (approved):

PR #4361, opened 2026-07-29
[runtime] Support Atomic Delete

patrick-ogrady:
Stacked on #4327. This replaces the manual destroy ordering and destroy-specific restart handling added during that review with one runtime-level, crash-recoverable namespace removal.

## Summary

Storage objects often own several blobs or partitions that represent one logical generation. Destroying those components sequentially cannot make the generation disappear atomically: an error, cancellation, or crash between awaits can expose a mixture of removed and surviving components. Each higher-level `init` then has to recognize those mixtures, repair them in the right order, and remain compatible with every child object's recovery behavior.

This adds `Storage::remove_batch`, which gives a set of exact namespace targets one durable commit point. Storage structures now prepare the physical targets they own, composite structures merge their child plans, and the outermost destroy submits the union once.

Before the commit point, no target has been removed. After commitment, recovery completes the entire removal before another namespace operation proceeds. An error after commitment may therefore mean the requested generation is already gone, but it cannot leave a partial generation observable through the runtime.

## Runtime contract and implementation

`RemoveTarget` identifies either one exact blob or an entire partition. `remove_batch`:

- validates every partition before mutation.
- canonicalizes the request as a set, removing duplicates and blob targets subsumed by a partition target.
- treats empty batches and already-missing targets as success.
- preserves the existing read-after-remove and independent-generation semantics for open handles and recreated names.
- serializes recovery with every namespace operation.

The filesystem implementation records canonical targets in a bounded, checksummed manifest under the runtime control directory. It syncs a prepared manifest, renames it to the committed marker, syncs that commit, applies the idempotent removals, syncs the affected namespace, and finally removes and syncs the marker. The next namespace operation discards an uncommitted prepared manifest and completes a committed one.

Tokio moves the namespace guard into the blocking worker, so dropping the caller cannot abandon a committed batch or expose recovery halfway through. The io_uring implementation performs the same state machine while holding its synchronous namespace lock. The memory implementation removes the complete set while holding its storage/recovery locks and retires the removed blob generations.

On Windows, entries are renamed into a private graveyard before deletion. This releases the user-visible names immediately while preserving the contents observed by handles opened before removal. Directory-entry durability remains best-effort on platforms that cannot synchronize directories, matching the existing `Storage::remove` contract.

The audited, metered, deterministic, faulty, mock, and runtime-owned wrappers all propagate the new operation. Fault injection distinguishes failure before commitment from an error reported after commitment so callers can exercise both outcomes.

## Storage integration

`DestroyPlan` carries a runtime context plus the exact physical namespace entries owned by one logical object. Leaf objects drain any in-flight sync work and return a plan without deleting anything. Composite objects merge child plans, and only the outer destroy executes the batch.

This is propagated through metadata, ordinal storage, the freezer, immutable and prunable archives, contiguous and segmented journals, authenticated journals, persisted Merkle structures, and Any/Current QMDB. QMDB state-sync replacement and mismatch discard also remove the backing journal and Merkle projection as one generation, so a retry sees either the old generation or an empty replacement target, never a mixture.

With partial destruction no longer observable, the destroy-only restart handling is removed:

- persisted Merkle no longer clears pruning metadata merely because its journal is empty.
- Any QMDB no longer treats a genesis-only log as evidence that pruning state is stale.
- Current QMDB no longer clears graft metadata based on that inferred destroy state.
- component-specific destroy interruption hooks and reopen tests are replaced by batch commit/recovery coverage.

Recovery for independently reachable live-state protocols remains in place. Staged clears and resets, journal tail repair, pending state-sync intent, and other mutable operations still have their own durable recovery rules; the change only removes recovery states that were created by sequential destruction.

The design relies on the runtime's existing exclusive ownership of its storage directory. Under that supported contract, moving the transaction boundary into the runtime replaces a growing collection of ordering rules and init-time repair branches with one reusable namespace invariant.


storage/src/journal/segmented/glob.rs
```diff
@@ -242,9 +242,9 @@ impl<E: BufferPooler + Storage + Metrics, V: CodecShared> Inner<E, V> {
         self.manager.remove_section(section).await
     }
 
-    /// See [Glob::destroy].
-    async fn destroy(self) -> Result<(), Error> {
-        self.manager.destroy().await
+    /// Wait for pending section syncs, then prepare the glob for removal.
+    async fn prepare_destroy(self) -> Result<DestroyPlan<E>, Error> {
```
patrick-ogrady: I'm not sure we need this DestroyPlan abstraction (I think we could find a way to hide this better)

PR #4363, opened 2026-07-29
[storage/qmdb/sync] remove cancel_tx

danlaine:
`Resolver::get_operations` took a `cancel_rx`. Dropping a future already cancels it in Rust so I don't think this is useful.

## What replaces it

**Caller side.** `Requests` now uses `commonware_utils::futures::{AbortablePool, Aborter}`. `TrackedRequest`'s cancel sender becomes an `Aborter`, so every existing cancellation point still cancels by dropping the tracked entry and the surrounding logic is unchanged. `next_completed()` yields `Ok(..)` or `Err(Aborted)`, and `handle_event` skips the aborted case while still waking the loop to reschedule. `handle_fetch_result`'s behavior is unchanged; its untracked-id guard is now purely defensive, since an aborted future can no longer surface a result.

The pool holds a never-resolving future internally, so `wait_for_event` also drops its empty-pool arm. Its all-empty early return, which is what produces `SyncStalled`, is unchanged, and `len()` still counts tracked requests rather than pooled futures.

**Implementor side.** Glue's standard mailbox gains a `CancelGuard` that retracts the request on drop. This is not a new pattern: glue's *compact* mailbox already works exactly this way with no channel, and the two now read identically. It is also a small correctness gain, since the mailbox previously retracted only when the channel fired, so a caller that simply dropped its future never told the actor.

## Test Fix 1

`any/sync/tests.rs` had two `oneshot::channel().1` call sites feeding the removed parameter. The storage-side resolvers bind it as `_cancel_rx` and never read it, so these were dead plumbing; they go with the parameter. One of the two test resolvers also awaited its receiver before serving pinned nodes; it now returns a pending future, exercising the same never-answers path without the channel.

## Test Fix 2

`reshare_e2e_state_sync_restart_before_epoch_boundary` failed on this branch as previously written, and now runs on a fast link (4ms latency, 1ms jitter) instead of the default 10ms/5ms.

The scenario crashes a state-syncing node inside a height window of `48..=64`. The node processes nothing until sync finishes, so it lands on whatever height the chain reached by then, and that landing height is what the window has to contain.

At the default link sync cannot converge on a moving target. Every target update kills the in-flight request covering the frontier and re-issues it at the new start, so exactly one fetch lands per block interval and each interval ends an operation or two short of the target. Converging takes two round trips inside one interval, roughly 48ms of a 40ms interval, which happens only when a slightly long interval arrives while the engine is idle. Whichever cycle that coincidence falls on decides the landing height.

**This PR moves which cycle that is.** Nothing in the sync loop got slower; measured before and after, the fetch round trip is 23.0ms and 24.9ms, the block interval 40.3ms and 41.0ms, the fetches killed by the next target update 25 of 48 and 34 of 65, and the slack left in an interval after its one fetch 17.8ms and 14.3ms. What moved is when the engine wakes for a killed request: its doomed reply used to come back from the peer about 20ms later and wake the loop then, and now the abort resolves at once. Same wakeups, different times, so the coincidence lands on cycle 34 instead of cycle 25, which is height 67 instead of 58, three blocks past the bound. The retraction messages are not the cause: the old mailbox patched to always retract, instead of harvesting a reply that had already arrived, still lands at 58.

On the fast link two round trips fit in one interval, so sync converges promptly, the node reaches the window under its own steam through heights 44 to 47, and the processed hold parks it at exactly `crash_window_start`. The landing height is now pinned by the hold rather than by sync timing, which is what the hold was for. The upper bound was not widened: it is the scenario, since a restart after the boundary is a different test.

### DIFF
--- a/glue/src/stateful/db/p2p/cancel.rs
+++ b/glue/src/stateful/db/p2p/cancel.rs
@@ -2,13 +2,13 @@
 
 use commonware_actor::mailbox::{Policy, Sender};
 
-/// Sends its message on drop, unless [`CancelGuard::disarm`] ran first.
-pub(super) struct CancelGuard<T: Policy> {
+/// Sends its message on drop, unless [`Guard::disarm`] ran first.
+pub(super) struct Guard<T: Policy> {
     sender: Sender<T>,
     cancel: Option<T>,
 }
 
-impl<T: Policy> CancelGuard<T> {
+impl<T: Policy> Guard<T> {
     pub(super) const fn new(sender: Sender<T>, cancel: T) -> Self {
         Self {
             sender,
@@ -22,7 +22,7 @@ impl<T: Policy> CancelGuard<T> {
     }
 }
 
-impl<T: Policy> Drop for CancelGuard<T> {
+impl<T: Policy> Drop for Guard<T> {
     fn drop(&mut self) {
         let Some(cancel) = self.cancel.take() else {
             return;
--- a/glue/src/stateful/db/p2p/cancel.rs
+++ b/glue/src/stateful/db/p2p/cancel.rs
@@ -1,0 +1,32 @@
+//! Cancellation of an in-flight resolver request.
+
+use commonware_actor::mailbox::{Policy, Sender};
+
+/// Sends its message on drop, unless [`CancelGuard::disarm`] ran first.
+pub(super) struct CancelGuard<T: Policy> {
+    sender: Sender<T>,
+    cancel: Option<T>,
+}
+
+impl<T: Policy> CancelGuard<T> {
+    pub(super) const fn new(sender: Sender<T>, cancel: T) -> Self {
+        Self {
+            sender,
+            cancel: Some(cancel),
+        }
+    }
+
+    /// Leaves the request with the actor, for a fetch that reached an answer.
+    pub(super) fn disarm(&mut self) {
+        self.cancel = None;
+    }
+}
+
+impl<T: Policy> Drop for CancelGuard<T> {
+    fn drop(&mut self) {
+        let Some(cancel) = self.cancel.take() else {
+            return;
+        };
+        let _ = self.sender.enqueue(cancel);
+    }
+}
--- a/glue/src/stateful/db/p2p/cancel.rs
+++ b/glue/src/stateful/db/p2p/cancel.rs
@@ -16,7 +16,7 @@ impl<T: Policy> CancelGuard<T> {
         }
     }
 
-    /// Leaves the request with the actor, for a fetch that reached an answer.
+    /// Suppresses the drop-time cancel; call once the fetch has completed.
     pub(super) fn disarm(&mut self) {
         self.cancel = None;
     }
--- a/glue/src/stateful/db/p2p/standard/mailbox.rs
+++ b/glue/src/stateful/db/p2p/standard/mailbox.rs
@@ -5,14 +5,39 @@ use crate::stateful::db::{AttachableResolver, Shared};
 use commonware_actor::mailbox::{Overflow, Policy, Sender};
 use commonware_codec::Read;
 use commonware_cryptography::Digest;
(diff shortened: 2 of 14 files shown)


storage/src/qmdb/sync/engine.rs
```diff
@@ -562,17 +543,14 @@ where
...
+    /// Verifies the proof against the current root first, then falls back
     /// to a matching historical root from `retained_roots` if available.
     fn handle_fetch_result(
         &mut self,
         fetch_result: IndexedFetchResult<DB::Family, DB::Op, DB::Digest, R::Error>,
     ) -> Result<(), Error<DB, R>> {
-        // Discard results for stale requests (removed by a target update).
-        // Using the request ID prevents a stale future from consuming the
-        // tracking entry of a fresh request at the same location.
+        // Defensive: removal aborts a request's future, so a result for an
```
patrick-ogrady: Can we cleanup this comment?

storage/src/qmdb/sync/engine.rs
```diff
@@ -685,7 +663,10 @@ where
             }
             Event::FinishChannelClosed => Err(SyncError::Engine(EngineError::FinishChannelClosed)),
             Event::BatchReceived(fetch_result) => {
-                self.handle_fetch_result(fetch_result)?;
+                // An aborted request carries no result, but still wakes the loop to reschedule.
+                if let Ok(fetch_result) = fetch_result {
+                    self.handle_fetch_result(fetch_result)?;
+                }
```
patrick-ogrady, line 669: add a \n here

storage/src/qmdb/any/sync/tests.rs
```diff
@@ -1898,21 +1891,12 @@ impl<R: Resolver<Digest = Digest>> Resolver for ReplayFreshBoundaryResolver<R> {
...
-                    .inner
-                    .get_operations(
-                        op_count,
-                        start_loc,
-                        max_ops,
-                        include_pinned_nodes,
-                        oneshot::channel().1,
-                    )
-                    .await;
+                // Never answers. The engine abandons this request when the target moves,
```
patrick-ogrady: remove the "never answers"

storage/src/qmdb/any/sync/tests.rs
```diff
@@ -1898,21 +1891,12 @@ impl<R: Resolver<Digest = Digest>> Resolver for ReplayFreshBoundaryResolver<R> {
...
+                // which drops this future.
+                return std::future::pending().await;
```
patrick-ogrady, line 1900: is this right? double-check...

glue/src/stateful/db/p2p/cancel.rs
```diff
@@ -0,0 +1,32 @@
+//! Cancellation of an in-flight resolver request.
+
+use commonware_actor::mailbox::{Policy, Sender};
+
+/// Sends its message on drop, unless [`CancelGuard::disarm`] ran first.
+pub(super) struct CancelGuard<T: Policy> {
```
patrick-ogrady: nit: Guard rather than CancelGuard (will be `cancel::Guard`)?

glue/src/stateful/db/p2p/standard/mailbox.rs
```diff
@@ -153,24 +150,16 @@ where
...
-            response = response_rx.as_mut() => response.map_err(|_| ResponseDropped)?,
-            _ = cancel_rx.as_mut() => {
-                if let Some(response) = response_rx.as_mut().now_or_never() {
-                    return response.map_err(|_| ResponseDropped)?;
-                }
-                let _ = self.sender.enqueue(Message::CancelOperations { request });
-                Err(ResponseDropped)
-            },
-        }
+        let mut cancel =
```
patrick-ogrady: restore a \n above this.

patrick-ogrady (approved):

PR #4364, opened 2026-07-30
[glue] Verify Reshare Epoch Artifacts on Demand

### LINKED ISSUE #4300: [glue] Verify reshare epoch material asynchronously
PR #4131 derives and verifies the next epoch artifact inline when `reshare::Application` requests `epoch_info` for a final block. This keeps DKG observation and finalization work on the final block proposal and verification path.

Kick this work off earlier in the reshare actor and cache it for the exact pending log set. Final block proposal and verification should await the in-flight result and restart it when the pending ancestry changes.

Review context: https://github.com/commonwarexyz/monorepo/pull/4131#discussion_r3616406495

clabby:
## Summary

- Moves reshare ceremony verification off the inclusion actor and onto the shared CPU pool.
- Starts verification only for a live final-block artifact request or exact canonical finalization, with at most one active verification task.
- Keeps competing ancestry requests lazy and FIFO, materializing one active request while reusing results for exact dealer-log views.
- Follows the epoch after a late state-sync floor, where skipped inclusion history cannot prove dealer-log coverage.

## Adversarial behavior

The epoch determines the boundary height, while the typed ancestry head determines fork identity. Inclusion derives the boundary parent from that ancestry and rejects missing, unexpected, gapped, or truncated chains. A malicious leader cannot force verification of every growing dealer-log prefix, and a malicious host cannot provide an independent canonical fanout key.

Finalization completes any admitted speculative waiter before reconstructing the canonical artifact. Speculative verification, caches, and queued requests remain phase-local and reconstructable after restart; wire and storage formats are unchanged.

## Validation

- `just test --profile slow -p commonware-glue reshare_e2e_state_sync`
- `just check-fmt`
- `just check-stability`

Closes #4300


glue/src/dkg/reshare/actor/inclusion.rs
```diff
@@ -402,25 +828,202 @@ where
...
+
+        // A verification admitted before finalization owns a stable application
+        // verdict. Publish that bounded work before reconstructing the canonical
+        // artifact.
+        if work.verification.task.is_some() {
+            let completed = (&mut work.verification.task)
+                .await
+                .expect("verification task failed");
+            self.complete_verification(epoch, store, work, completed)
+                .await;
```
clabby: I don't think this is safe? The verification task could have been cancelled.

patrick-ogrady (approved):
Won't call glue::dkg ready for primetime yet but this is definitely a step forward.

patrick-ogrady:
Post-merge note: this PR also fixes a security-relevant final-block validation gap in reshare follower mode. Previously, EpochInfoResponse::Following delegated final-block verification to the wrapped application even though the follower could not independently reconstruct the epoch artifact. That allowed a follower to approve boundary metadata it had not validated. The merged behavior keeps verification pending instead, so the node can learn the outcome from external finalization without producing a completed application verdict for unverified epoch metadata.

PR #4365, opened 2026-07-30
[runtime/tokio] high-resolution timers

andresilva:
This PR replaces the Tokio timer used by the runtime `Clock` on Linux and macOS with a native, sharded high-resolution timer service. Each shard stores all pending deadlines in an indexed 4-ary min-heap and uses one native alarm to track the earliest deadline. Linux implements that alarm with an absolute `CLOCK_MONOTONIC` `timerfd`, while macOS uses a `kqueue` with a Mach-absolute one-shot timer. When the alarm fires, the driver removes expired deadlines, wakes their sleepers, and rearms the alarm for the heap's next deadline.

On Linux and macOS, each runtime creates one shard and driver per Tokio worker, reducing heap-lock contention and allowing cancellation-heavy workloads to scale across workers. Stable thread-local affinity selects a shard when a sleep is constructed, and the sleep retains that shard if its task later migrates. The public `Clock` API and construction-time deadline semantics remain unchanged, but short sleeps are no longer constrained by Tokio's roughly millisecond timer granularity. Windows and other unsupported targets continue to use Tokio timers.

Zero-duration and past-deadline sleeps use an inline ready state. Every other sleep converts its requested delay or wall-clock deadline once into a fixed monotonic deadline and allocates one shared entry containing its lifecycle, waker, and exact heap index. This makes cancellation an immediate `O(log n)` removal instead of leaving dead entries in the queue, while a single atomic transition arbitrates expiry, cancellation, shutdown, and driver failure.

Registering or canceling a sleep updates its shard's heap and sends a coalesced driver notification only when the new minimum requires the native alarm to be rearmed. The driver snapshots that minimum, releases the heap lock, arms or disarms the native alarm, and repeats if a concurrent heap update changed the desired deadline during the syscall. Native operations and waker callbacks therefore never run under the heap lock. A 50 ns tolerance avoids a redundant rearm when the current alarm is already only slightly later.

On expiry, the driver removes at most 32 entries per lock acquisition and wakes their sleepers after releasing the lock. After processing 512 expired entries, it yields if additional expired entries remain, bounding lock hold time and preserving executor fairness without adding unconditional scheduler hops. Orderly shutdown stops queued sleeps without waking them, while native I/O failures and driver panics fail affected sleeps and interrupt root execution.

While the overall PR is large, the production timer implementation is roughly 2,500 lines. Almost all of the remaining diff is tests, including loom models, and the benchmark harness used to produce the results below.

## Results

Benchmark results were collected on a Ryzen 9 9950X3D running Linux 7.1.5 with Rust 1.97.1, using `taskset -c 8-15` and eight runtime workers except for the intentionally single-worker expiry-fairness workload.

### Wake lateness

Wake lateness is the time from a timer's requested deadline until its awaiting task resumes, so it includes alarm delivery, timer-driver work, wake-up, and executor scheduling.

**Synchronized** releases every task in a batch together against one shared absolute deadline whose lead time is shown in the table. At large batch sizes this deliberately includes barrier release, registration, and run-queue fan-out, and the deadline may pass before every task registers.

**Spread** staggers tasks evenly across a 100 ms window, after which each requests its own relative sleep of the listed duration. This avoids a single registration and expiry burst and measures steady-state accuracy with the same number of tasks per batch.

| Pattern | Deadline | Tasks/batch | Samples/backend | Commonware p50 / p99 | Tokio p50 / p99 |
|---|---:|---:|---:|---:|---:|
| Synchronized | 10 µs | 1 | 20 | 4.358 / 11.862 µs | 1046.755 / 1050.332 µs |
| Synchronized | 100 µs | 1 | 20 | 2.776 / 13.576 µs | 956.996 / 1957.925 µs |
| Synchronized | 100 µs | 1,000 | 20,000 | 32.572 / 68.712 µs | 1278.208 / 1422.574 µs |
| Spread | 100 µs | 1,000 | 20,000 | 3.738 / 17.293 µs | 960.062 / 2019.833 µs |
| Synchronized | 100 µs | 10,000 | 200,000 | 1504.721 / 2034.892 µs | 1285.573 / 2386.242 µs |
| Spread | 100 µs | 10,000 | 200,000 | 12.244 / 63.672 µs | 992.554 / 1888.742 µs |

### Registration and expiry

The registration workload constructs and initially polls 50,000 timers from one producer, assigning strictly descending deadlines with adjacent deadlines 1 µs apart so every insertion becomes the new earliest timer. The table reports the time for the complete batch, with cancellation outside the measurement.

The expiry workload registers 50,000 timers at one absolute deadline on a dedicated single-worker runtime. The sleeps share a recording waker rather than running as 50,000 separate tasks, so the first two metrics isolate timer callback delivery from task scheduling. The benchmark measures how late the first waker callback runs after the deadline, how long all 50,000 callbacks take from first to last, and the longest time an unrelated runnable task must wait for another turn while those callbacks are processed. The single-worker topology makes the last measurement a direct test of whether timer processing starves other runtime work.

| Workload | Metric | Samples/backend | Commonware p50 / max | Tokio p50 / max |
|---|---|---:|---:|---:|
| 50k descending registrations | Whole batch | 20 | 8.488 / 9.647 ms | 4.410 / 4.610 ms |
| 50k simultaneous expiries | First callback lateness | 20 | 17.108 / 24.823 µs | 969.254 / 1068.863 µs |
| 50k simultaneous expiries | First-to-last callback delivery | 20 | 4.529 / 5.480 ms | 0.903 / 1.354 ms |
| 50k simultaneous expiries | Longest unrelated-task wait | 20 | 71.847 / 289.142 µs | 1169.380 / 1606.214 µs |

### 99k cancellations from 100k registered timers

Canceling a pending timer means dropping its sleep before the deadline, which requires the backend to remove the registered deadline from its timer data structure. Each batch creates and initially polls 100,000 far-future timers, so both backends have registered every timer before timing begins. After a deterministic shuffle, the producer threads wait at a gate. Timing starts immediately before the gate opens and ends when the last producer has dropped its share of 99,000 timers. The other 1,000 remain registered until after that final timestamp, keeping the timer queues populated throughout the measurement. The result is therefore the wall-clock time needed to cancel 99,000 already-registered timers concurrently; it excludes registration and is not a measurement of individual cancellation latency.

The producer counts answer whether cancellation throughput scales with the sharded design. One producer gives the uncontended baseline, eight give Commonware one producer per shard, and 32 make four producers contend for each of its eight shards. Tokio uses one shared timer driver in every case. Scaling divides the one-producer median by the current median, so a value above 1x means that adding producers completed the same 99,000 cancellations sooner.

| Backend | Producers | Effective topology | Time to cancel 99k, p50 / max (10 batches) | Scaling vs. 1 producer |
|---|---:|---|---:|---:|
| Commonware | 1 | 1 of 8 shards x 1 producer | 5.243 / 6.178 ms | 1.00x |
| Commonware | 8 | 8 shards x 1 producer | 0.637 / 1.405 ms | 8.24x |
| Commonware | 32 | 8 shards x 4 producers | 1.189 / 1.663 ms | 4.41x |
| Tokio | 1 | Shared timer | 2.640 / 3.751 ms | 1.00x |
| Tokio | 8 | Shared timer | 15.768 / 16.578 ms | 0.17x |
| Tokio | 32 | Shared timer | 16.469 / 17.267 ms | 0.16x |


runtime/src/utils/handle.rs
```diff
@@ -325,25 +339,41 @@ impl Panicked {
...
-        let panicked = self.receiver;
-        pin_mut!(panicked);
-        pin_mut!(task);
-        match select(panicked, task).await {
-            Either::Left((panic, task)) => match panic {
-                // If there is a panic, resume the unwind
-                Ok(panic) => {
-                    resume_unwind(panic);
+        let mut panicked = std::pin::pin!(self.receiver);
+        // Capture root-task panics long enough to arbitrate them against a
```
patrick-ogrady, line 406: nit: we should avoid cutting up code blocks with comments...either add a \n before the comment or place it at the top of the block

PR #4368, opened 2026-07-30
[runtime] Add Support for Atomic Blob Primitives

patrick-ogrady:
## Summary

This PR adds opt-in append-only atomic blobs and coordinator-free multi-blob publication without writing payload twice. It also adds V2-native contiguous journals that remove the legacy `Metadata` checkpoint object and publish variable-journal data with its offset index in one atomic group.

The mutation model is deliberately narrow:

- append at the current logical tail;
- rewind to an existing offset;
- publish appends, rewinds, a 64-byte application tag, and exact blob deletions atomically across a group.

This is not a transactional filesystem or an atomic random-overwrite engine. Ordinary `Storage`, V0/V1 blobs, and legacy contiguous journal APIs/formats remain unchanged. Atomic/Batch/V2 are opt-in ALPHA surfaces that have not landed on `main`, so this PR updates that pending design directly. Earlier experimental V2 root spellings fail loudly rather than being mistaken for an empty current blob.

The implementation is nicknamed UNO because the eligible hot path targets **one payload write and one inode durability barrier**. UNO is a development nickname, not a public acronym or API name.

## V2 blob layout

Each V2 blob reserves one immutable page and one mutable root page:

```text
offset 0
+----------------------------+
| immutable V2 header        | 4 KiB: version + 128-bit incarnation
+----------------------------+
| root slot 0                | 2 KiB
+----------------------------+
| root slot 1                | 2 KiB
+----------------------------+
offset 8 KiB
| append-only payload        | written once at final physical offsets
| ...                        |
+----------------------------+
```

Generation parity selects the root slot and preserves the immediately preceding root as fallback. A root has a 92-byte header: spelling, generation, logical length, 64-byte application tag, and CRC32C. A batch-prepared slot then carries a 16-byte witness envelope and up to 1,940 bytes of local link data.

The two logical 2 KiB slots share one physical 4 KiB page but occupy disjoint requested ranges. Recovery permits any subset of bytes from requested writes not covered by a completed durability barrier to survive. It does not claim recovery from unrelated media corruption outside those ranges.

There is no shadow payload, reflink, extent map, hole punching, compaction log, coordinator file, or steady-state delete manifest.

## Single-blob publication

Appends write through immediately at their final offsets and update an incremental CRC32C. They are visible through the current handle and become crash-durable through the existing `sync`/`start_sync` boundary.

```text
old durable root
       |
       +-- append payload once at final offset
       |
       +-- write prepared root + self-link in alternate slot
       |
       +-- fdatasync(blob)  <-- payload and publication share this barrier
       |
       `-- new root is durable
```

There is no required follow-up root write or coordinator sync. A compatible later publication can supersede a carried write-only decision in the alternate slot.

Payload size has no protocol batch cap. Appends continue writing as they arrive; completed background preflushes make older immutable payload durable while producers continue. Publication leaves at most 64 MiB aggregate for crash-time CRC32C validation across the group. The candidate root is not written until publication begins.

Rewind publishes a shorter authoritative root before truncating the dead suffix. A crash before `ftruncate` leaves only unreachable bytes, and bounded reopen repeats the truncate. Rewinding unpublished bytes can reuse that tail after any in-flight preflush completes.

## Coordinator-free multi-blob groups

Each dirty participant stores one checksummed local link beside its prepared root. Canonical participant order forms a closed ring:

```text
blob A: [P | local A candidate | next exact B incarnation] --+
       ^                                                       |
       |                                                       v
blob C: [P | local C candidate | next exact A incarnation] <- blob B

fdatasync(A)   fdatasync(B)   fdatasync(C)
            participant barriers run concurrently
```

Each link has 296 fixed bytes: group ID, participant count/ordinal, local incarnation and candidate, delete bit, optional bounded-suffix checksum, and successor incarnation/lengths. One successor partition/name path may use up to 1,644 additional bytes. Names are distributed around the ring rather than summed or replicated into one descriptor.

There is no UNO-specific participant cap. The on-disk `u32` count is the format bound; recovery grows state incrementally and installation uses at most 32 workers at once. That worker fanout is scheduling, not admission.

Opening any unresolved participant follows its exact declared successors. Recovery accepts new state only when one unique sequence of consecutive ordinals closes after the declared hop count and every local root, incarnation, and pending suffix validates. It never scans unrelated names or historical payload.

`BatchStorage::start_apply` returning `Ok` proves that the complete decision is durable. Awaiting its handle waits until roots are active and namespace cleanup is finished. Dropping the handle does not cancel admitted work. An error or cancellation is indeterminate: recovery may choose old or complete new, never a partial group.

## Crash-safe deletion

Deletion uses an ordered materialization phase:

```text
1. validate the complete closed P ring
2. write every retained materialized root M and deletion tombstone T
3. make every M/T participant inode durable
4. unlink T from highest ordinal down to zero
5. sync its parent directory before unlinking the next ordinal
```

No name is removed until every final root/tombstone is independently durable. Per-unlink directory barriers mean a crash leaves only a prefix of the descending deletion order. Ordinal zero remains until every higher tombstone is durably absent, so it anchors recovery of the surviving forward-linked prefix. Recreated same-name paths are protected by immutable incarnations and device/inode checks.

## Fault model and integrity

Implementation and tests use the real write-fault model: any subset of requested writes not covered by a completed barrier may survive; recovery never assumes a surviving prefix.

- incomplete or torn prepared ring: use every prior root;
- complete prepared ring with valid candidates/suffixes: publish the complete new group;
- exact materialized root or tombstone: finish the already-proven decision;
- partially completed namespace cleanup: resume the descending delete frontier from ordinal zero.

CRC32C is intentional for trusted local-disk crash integrity. It probabilistically detects torn or missing bytes; it is not authentication against an actor that can rewrite both bytes and checksums.

## V2-native contiguous journals

The fixed and variable V2 journals use consuming `MutableV2` mutations and expose `start_sync`/`sync` for publication. They do not use the legacy `Metadata` object.

The checked-page wrapper divides the 64-byte root tag this way:

```text
+---------------------+-------------------------+----------------------+
| start + 1 (8 B, BE) | reserved zeroes (52 B) | partial CRC32C (4 B) |
+---------------------+-------------------------+----------------------+
```

The wrapper exposes the first 60 bytes as its application marker and reserves the final four for the unfinished-page CRC. Contiguous currently uses only the marker's first eight bytes and requires the other 52 to be zero. Exactly one fixed section, or one matching variable data/offset pair, carries the active marker. Section names identify the active tail; the marker identifies the retained start; committed root lengths identify the end. Reopen scans names and bounded roots, not journal frames or whole blobs.

Variable append staging is dependency-aware: writing encoded data yields its stable pending base, the journal derives absolute `u64` offsets from that base, then appends the offsets. One UNO group publishes both roots, so recovery sees matching data/index lengths or their complete prior pair.

Rollover may durably stage sealed unmarked sections while they remain unreachable. Final publication includes only the prior authority and current tail: two roots for fixed, or four data/offset roots for variable, regardless of how many sections the epoch crossed.

## Legacy migration

V2 journal initialization is eager and resumable:

```text
normal legacy recovery
       |
       +-- decode one bounded logical batch
       +-- write/sync its native V2 replacement
       +-- only then prune that legacy source
       `-- atomically activate the final V2 marker(s)
```

A temporary migration witness records the retained start while conversion is incomplete. It is removed before init returns and is not part of native reopen or hot-path publication. Once no legacy state remains, later runs open only V2.

`AtomicStorage::migrate_atomic` also migrates one opened ordinary blob at a time. It syncs the source, stream-copies through bounded memory into a staged V2 inode, syncs the replacement, renames it over the exact current generation, and syncs the relevant directories. Retry resolves ambiguous publication without reflinks, hole punching, or Linux-specific rename variants; temporary disk use is one extra blob copy.

## Intentional boundaries

- UNO provides atomic membership within one group, not a total durable order across independently started groups. There is no root WAL, global coordinator, LSN, or commit sequence.
- If group B temporally depends on group A, include the dependency in one group or wait for A's durable decision before starting B.
- This locality avoids a global lock and a volume-like filesystem layer; disjoint groups remain concurrent.
- General random overwrite still requires retained old bytes: undo/redo logging, page/extent COW, an append-only patch map plus reclamation, or a volume layer. A checksum detects an overwrite tear but cannot restore overwritten bytes.
- Freezer/Immutable Archive in-place tables are not converted here. They can still use UNO journals as atomic authority after ordinary table/index materializations become durable.

## Performance versus ordinary writes

The final R14 paired benchmark alternates ordinary and UNO groups in one process and measures full durability completion for both sides:

- four blobs per group;
- sixteen contiguous 64 KiB appends per blob (4 MiB logical payload/group);
- 250 groups per side, five paired seeds;
- Tokio, two worker threads, one group in flight.

| Filesystem/environment | Median UNO / ordinary | Five ratios |
| --- | ---: | --- |
| native macOS APFS | **0.962x** (3.8% overhead) | .986, .964, .945, .955, .962 |
| ext4 Docker volume inside a Linux VM | **0.980x** (2.0% overhead) | 1.005, .984, .958, .921, .980 |

The ext4 VM result remains noisy and is not an absolute filesystem claim. Together these paired measurements support the narrower architectural result: group atomicity does not require another payload write or coordinator-file barrier on this contiguous journal-shaped workload.

## Validation

- `just lint`: formatting, Cargo TOML, workspace Clippy, docs, feature propagation, custom lints, publish order, all stability levels, and public stability annotations passed.
- `just test -p commonware-runtime`: 818 passed, 4 profile-skipped.
- `just test -p commonware-storage`: 2,874 passed, 399 skipped.
- Linux Docker: Clippy passed with `iouring-storage` and `iouring-network`; all 39 focused io_uring storage tests passed.
- Linux Docker: release WASM builds passed for `commonware-runtime` and `commonware-storage`.
- `git diff --check` passed.
- The standalone Beamer walkthrough is [`docs/uno-v2-atomic-storage.pdf`](docs/uno-v2-atomic-storage.pdf).

`Oversized` is not migrated in this PR. It remains a natural first consumer: its append-only journals can advance all-or-nothing without an application-level pair-repair protocol.


roberto-bayardo:
A few low-risk reductions on the batch commit path — none change the durability contract (the commit point stays PREPARED-fsync + rename + control-dir fsync, and the marker is still removed only after the ops are durably applied). These just cut redundant syncs and overlap independent ones.

**1. Don't re-fsync the storage root on every batch.** `ensure_control_directory` fsyncs the root unconditionally, but the root entry for the control dir only changes when it's *created*. After that it's durable (and if it weren't, it wouldn't be visible after a crash, so recovery would recreate + re-sync it). Move the root fsync into the `create_dir` branch. Saves 1 fsync/batch in steady state.

**2. Only fsync directories whose entries actually changed.** `sync_operations` fsyncs every touched partition dir plus the root, for all op types. But `Update`/`Resize` change inode content/size (covered by the file's own sync) and never touch a directory entry — `open_blob_for_mutation` requires a pre-existing file, so nothing is created. So a directory fsync is only needed for `Remove` (its partition), and the root only for a whole-partition removal. An `Update`/`Resize`-only batch needs *zero* directory fsyncs here.

**3. Issue the independent blob syncs concurrently.** The pre-commit syncs of mutated blobs and the post-apply file syncs are both sequential today. Fanning them out (e.g. `join`) keeps the call count the same but lets the device coalesce flushes — a latency win that scales with batch size.

**4. `sync_data`/`fdatasync` for blob content instead of `sync_all`.** The per-blob content sync in `update_blob`/`resize_blob` flushes inode metadata (mtime, etc.) that isn't relevant here — only content and length matter. `File::sync_data` skips it. Cheaper per call; worth a per-backend/platform sanity check.

Net for the common cases: a single-blob checkpoint drops from ~7 fsyncs to ~3, and an `Update`/`Resize`-only batch loses all its directory fsyncs — without weakening any guarantee.

runtime/src/lib.rs
```diff
@@ -675,7 +721,50 @@ stability_scope!(BETA {
...
+        /// until every operation is durably applied. A conflicting namespace operation waits for
+        /// the batch to finish. Rejecting an invalid batch leaves every target unchanged. Other
+        /// errors, or cancellation of this future, may leave the outcome indeterminate; only `Ok`
+        /// proves the batch is durably committed.
+        ///
+        /// A mutation also makes prior writes through its blob handle durable. A resize preserves
+        /// existing bytes below `len` and zero-fills an extension. An update writes `data` at
+        /// `offset` and sets the blob length to `len`; the written range must fit within that
+        /// length. Callers must not access a mutated blob until the handle completes. Operations
+        /// on disjoint, already-open blobs may proceed concurrently.
```
danlaine: nits to make this slightly more readable imo
```suggestion
/// Start applying removals and blob mutations as one durable batch.
///
/// Every operation is validated before any is applied; a rejected batch leaves
/// every target unchanged. A mutated blob must belong to this storage instance
/// and still be current for its partition and name. Each blob takes at most one
/// mutation, and a mutation must not overlap a removal of the same blob or its
/// partition. Identical duplicates and removals subsumed by a partition removal
/// are permitted. Missing removal targets and empty batches succeed; a missing
/// mutation target fails the batch.
///
/// Only `Ok` proves the batch is durably committed. Once committed, the batch
/// finishes applying even if the returned [`Handle`] is dropped or aborted;
/// awaiting it waits until every operation is durably applied. Any other
/// outcome — an error, or cancellation of this future — leaves the batch's
/// fate unknown.
///
/// A mutation first makes prior writes through its blob durable. A resize sets
/// the blob to `len` bytes, preserving existing bytes below and zero-filling
/// above. An update writes `data` at `offset` and sets the length to `len`; the
/// written range must fit within `len`. Do not access a mutated blob until the
/// returned [`Handle`] completes.
///
/// Other namespace operations wait for the batch to finish; I/O on unrelated
/// open blobs proceeds concurrently.
```

runtime/src/storage/batch.rs
```diff
@@ -0,0 +1,1767 @@
...
+pub(crate) enum Operation {
+    /// Remove an exact namespace target.
+    Remove(RemoveTarget),
+    /// Resize an exact blob to `len` bytes.
+    Resize {
+        partition: String,
+        name: Vec<u8>,
+        len: u64,
+    },
+    /// Write an exact range and resize an exact blob to `len` bytes.
```
danlaine: I feel like it is implied that we exact as opposed to inexact things. Is this the new goblins? According to Claude, "It's Patrick's signature word in this whole lineage."
 
```suggestion
/// A namespace operation used for validation and canonicalization.
#[derive(Clone, Debug, Eq, PartialEq)]
pub(crate) enum Operation {
    /// Remove a namespace target.
    Remove(RemoveTarget),
    /// Resize a blob to `len` bytes.
    Resize {
        partition: String,
        name: Vec<u8>,
        len: u64,
    },
    /// Write a range and resize an exact blob to `len` bytes.
```

runtime/src/storage/batch.rs
```diff
@@ -0,0 +1,1767 @@
...
+
+#[derive(Eq, PartialEq)]
+enum BlobOperation {
+    Remove,
+    Resize(u64),
+    Update { offset: u64, data: IoBuf, len: u64 },
+}
+
+enum PartitionOperations {
+    Blobs(BTreeMap<Vec<u8>, BlobOperation>),
```
danlaine, line 68: might be worth documenting what the key is

runtime/src/storage/batch.rs
```diff
@@ -0,0 +1,1767 @@
...
+    Remove,
+}
+
+fn invalid_operations(message: &'static str) -> Error {
+    io::Error::new(ErrorKind::InvalidInput, message).into()
+}
+
+pub(crate) fn validate_update(offset: u64, data_len: usize, len: u64) -> Result<(), Error> {
```
danlaine: nit: comment

runtime/src/storage/batch.rs
```diff
@@ -0,0 +1,1767 @@
...
+                    len: *len,
+                }
+            }
+        })
+        .collect();
+    canonicalize_operations(descriptors)
+}
+
+/// Validate and reduce a removal batch to its exact namespace set.
+#[cfg(test)]
```
danlaine: should this go in `mod tests`?

runtime/src/storage/batch.rs
```diff
@@ -0,0 +1,1767 @@
...
+                    ));
+                }
+            },
+        },
+    }
+    Ok(())
+}
+
+/// Map blob handles without changing the operation set.
+pub(crate) fn map_blobs<B, C>(
```
danlaine, line 116:
```suggestion
/// Map each operation from B to C, leaving removals unchanged, and return the updated operations.
pub(crate) fn map_blobs<B, C>(
```

runtime/src/storage/batch.rs
```diff
@@ -0,0 +1,1767 @@
...
+                blob: map(blob),
+                offset,
+                data,
+                len,
+            },
+        })
+        .collect()
+}
+
+/// Describe, validate, and canonicalize a batch without consuming its blob handles.
```
danlaine, line 133:
```suggestion
/// Validate and canonicalize a batch into name-addressed [`Operation`]s.
/// `location` returns a mutated blob's partition and name. The batch is not consumed.
```

runtime/src/storage/batch.rs
```diff
@@ -0,0 +1,1767 @@
...
+                    Operation::Update { .. } => {
+                        unreachable!("removal-only input produced a blob mutation")
+                    }
+                })
+                .collect()
+        },
+    )
+}
+
+/// Validate and reduce a mixed batch to a deterministic exact operation set.
```
danlaine, line 162:
```suggestion
/// Validate and reduce a mixed batch to a deterministic operation set.
```

roberto-bayardo:
While reviewing the deletion frontier I did hit one reachable crash-safety hole, so I pushed a fix (824bcd3df).

## Bug: a deleted blob can resurrect after a crash mid-unlink

A blob deleted as a **non-anchor** (ring ordinal ≠ 0) participant of a multi-blob group can reopen with its **pre-delete contents** after a crash interrupts the descending tombstone-unlink phase.

**Mechanism.** Deletion tombstones each participant (a payload-preserving root that keeps the prior committed root recoverable so already-open handles stay readable), then unlinks the names in descending ordinal order. A crash mid-unlink leaves a non-anchor tombstone whose name lingers — only ordinal-zero recovery resumes the descending frontier, and there's no global startup scan (`recover`/`recover_notifying` are no-ops). `recover_named_embedded` correctly reports such a name `removed`, but `open_versioned_inner` **discarded that verdict**, reopened the lingering inode, and recovered its state from the retained committed root via `State::recover` (the `CWUNOT13` tombstone isn't a committed/materialized spelling, so `recover_from_root_slots` ignores it and selects the older committed root in the sibling slot). Net: a blob whose deletion `start_apply` had already reported durable comes back alive.

**Repro** (confirmed end-to-end through the real `open_atomic` path): create 3 deleted blobs `a<b<c`, materialize tombstones, unlink only `c` (crash mid-descending-unlink), then `open_atomic("partition", b"b")` returns a live blob with `b`'s pre-delete 5-byte payload.

## Fix

Honor the `removed` verdict on open in **both** filesystem backends (tokio + io_uring): treat a durably-removed lingering tombstone as absent so a fresh generation replaces it (via `create_live`'s rename) instead of reopening it as live. No wire/storage-format change, and `State::recover`'s committed-fallback for open handles is left intact (it's intentional — `deletion_tombstone_preserves_open_handle_payload`).

Added a regression test `nonanchor_tombstone_does_not_resurrect_on_open`. Verified locally: it plus the 21 batch + 41 tokio + 25 atomic tests pass. I couldn't build the io_uring backend on macOS, so **that half of the fix relies on CI's `iouring-storage` check** — worth a glance when CI runs.

## Follow-up (not in this commit)

Recreating a non-anchor deleted blob replaces its tombstone with a fresh incarnation, which makes the old group's ordinal-zero frontier recovery abort on the incarnation mismatch — so the group's *other* tombstone **names** can leak (they still read correctly as deleted / get recreated on open; no correctness impact). That's the pre-existing lazy-cleanup behavior, slightly amplified — a candidate for a later GC pass over leaked tombstone names, but out of scope here.

🤖 Analysis + fix via [Claude Code](https://claude.com/claude-code)

PR #4373, opened 2026-07-31
[cryptography] Clarify SmallScalar Randomness

patrick-ogrady:
## Summary

- Define `SmallScalar` over its complete `[0, 2^128)` domain, including zero.
- Sample with a single uniform 128-bit draw and add regression coverage for an all-zero draw.
- Explain why uniformly sampled zero preserves the `2^-128` soundness bound, while predictable challenges are unsafe regardless of value.
- Correct the BLS batch-verification soundness documentation and centralize the crate-wide secure-RNG requirement.

## Rationale

For an invalid random linear-combination check, at least one error term is non-zero. After fixing every other challenge, at most one of the `2^128` values for that term can make the check pass. Zero is therefore a valid challenge: rejecting it neither improves the `2^-128` bound nor makes a predictable RNG secure.

`Scalar::random` remains intentionally non-zero because it follows IETF BLS KeyGen and backs private-key generation.

## Validation

- `just test -p commonware-cryptography` (655 passed)
- `just clippy -p commonware-cryptography`
- `just check-fmt`
- `just check-docs -p commonware-cryptography`
- `just check-stability -p commonware-cryptography`
- cryptography WASM release build

### DIFF
--- a/cryptography/src/bls12381/primitives/group.rs
+++ b/cryptography/src/bls12381/primitives/group.rs
@@ -287,7 +287,11 @@ pub struct SmallScalar {
 impl SmallScalar {
     /// Generates a random non-zero 128-bit scalar.
     pub fn random(mut rng: impl CryptoRng) -> Self {
+        // blst_scalar is 32 bytes.
         let mut bytes = [0u8; SCALAR_LENGTH];
+        // Fill the last 16 bytes (128 bits) with entropy.
+        // In big-endian, bytes[16..32] are the least significant.
+        // Leaving bytes[0..16] as zero ensures the scalar is < 2^128.
         let random_bytes = &mut bytes[(SCALAR_LENGTH - SMALL_SCALAR_LENGTH)..];
         while bool::from(all_zero(random_bytes)) {
             rng.fill_bytes(random_bytes);
--- a/cryptography/src/bls12381/primitives/group.rs
+++ b/cryptography/src/bls12381/primitives/group.rs
@@ -285,14 +285,13 @@ pub struct SmallScalar {
 }
 
 impl SmallScalar {
-    /// Generates a random 128-bit scalar.
+    /// Generates a random non-zero 128-bit scalar.
     pub fn random(mut rng: impl CryptoRng) -> Self {
-        // blst_scalar is 32 bytes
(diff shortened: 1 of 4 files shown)


cryptography/src/bls12381/primitives/group.rs
```diff
@@ -285,14 +285,17 @@ pub struct SmallScalar {
...
     pub fn random(mut rng: impl CryptoRng) -> Self {
-        // blst_scalar is 32 bytes
-        let mut bytes = [0u8; 32];
+        // blst_scalar is 32 bytes.
+        let mut bytes = [0u8; SCALAR_LENGTH];
         // Fill the last 16 bytes (128 bits) with entropy.
         // In big-endian, bytes[16..32] are the least significant.
         // Leaving bytes[0..16] as zero ensures the scalar is < 2^128.
-        rng.fill_bytes(&mut bytes[SMALL_SCALAR_LENGTH..]);
+        let random_bytes = &mut bytes[(SCALAR_LENGTH - SMALL_SCALAR_LENGTH)..];
```
patrick-ogrady, line 295: This is equivalent but clearer.

cronokirby:
I think we should be able to assume that the RNG passed is cryptographically good; in that case, the likelihood of zero appearing is negligible. SmallScalar needs to have 128 bits of entropy for security in other places, and merely rejecting zero does not guarantee that. Now a bad RNG can pin a core as well.

patrick-ogrady:
> I think we should be able to assume that the RNG passed is cryptographically good; in that case, the likelihood of zero appearing is negligible. SmallScalar needs to have 128 bits of entropy for security in other places, and merely rejecting zero does not guarantee that. Now a bad RNG can pin a core as well.

As discussed, we can do that but then need to change the surrounding documentation/confirm all areas using `SmallScalar` still work (they should AFAIK).

The remaining asymmetry (which may be fine) is `Scalar` only generates non-zero randomness, and may be nice to align? If we migrate `Scalar` to allow 0 randomness, we'll have to separate its init from `Private` for backwards compatibility (which is probably not ideal). Not that they need to be symmetric.

cronokirby (approved):

Issue #2520, opened 2025-12-15
[consensus/simplex] Avoid Storing Data from Past Views in Memory


patrick-ogrady:
Related: #1353

It "should" be possible to avoid storing any data from the non-current view in memory (reading old certificates from the journal, if needed).

This prevents a scenario where many many nullified views causes an OOM. This still doesn't avoid unbounded storage use but does avoid unbounded RAM use.

PR #4374, opened 2026-07-31
[consensus/simplex] Make Vote Reporting Durable and Bound Batcher Memory

patrick-ogrady:
Related: #2520

## Summary

Simplex previously handed a locally constructed vote to the Batcher before syncing the voter journal. If the process crashed after the Batcher reported that vote but before the sync completed, the application could observe one vote before the crash and a different vote after restart even though the voter never broadcast the first vote.

This PR stages locally constructed output for the event-loop iteration, appends it to the voter journal, performs one coalesced sync of the exact journal section written by that iteration, and only then publishes votes to the Batcher/reporter path and the network. Append or sync failure therefore terminates the voter before the vote becomes externally visible.

Votes received from the network remain unpersisted and may be exported again after restart. If a sender equivocates, the vote exported after restart may differ from the one exported before the crash.

## Bounded Batcher memory

Valid far-future certificates can leave many Batcher rounds retained. This PR reduces their per-view footprint by:

- lazily allocating per-phase vote maps and certificate-verification buffers;
- representing each vote phase explicitly as full or compacted;
- releasing full signed votes once the corresponding certificate exists when `historical_conflict_reporting` is disabled;
- retaining lazy compact signer and authoritative-proposal facts for duplicate suppression, targeted proposal forwarding, nullify/finalize cross-phase blocking, and blocking same-phase match/non-match relation flips; and
- releasing certificate-verification buffers after completion.

Compact state deliberately does not retain enough evidence to construct a conflict activity. Conflict reporting after certification is therefore best effort when `historical_conflict_reporting` is disabled. Enabling it retains every observed full vote for a certified phase, extending conflict reporting at the cost of substantially higher per-view memory.

Full-evidence conflict checks compare complete signed votes, so signing schemes must use deterministic signature encodings.

Coverage includes deterministic crash regressions around finalize and nullify durability; compact and retained notarize/finalize conflicts in both proposal-relation orders; and retention, forwarding, deduplication, certificate-first allocation, and verifier-buffer allocation cases.

## Memory impact

In a controlled RSS microprobe using 50 BLS VRF validators and 2,000 retained certificate-only Batcher rounds, average retained memory fell from 209.5 KiB per view to 0.55 KiB per view, a 99.74% reduction. This isolates the per-view Batcher allocation changed here and is not a whole-validator RSS measurement.

### DIFF
--- a/consensus/src/simplex/actors/voter/actor.rs
+++ b/consensus/src/simplex/actors/voter/actor.rs
@@ -405,7 +405,6 @@ impl<
     #[allow(clippy::type_complexity)]
     async fn timeout(
         mut self,
-        batcher: &mut batcher::Mailbox<S, D>,
         reason: TimeoutReason,
     ) -> (Self, Option<(Nullify<S>, Option<Certificate<S, D>>)>) {
         // Construct a nullify vote for the current view
@@ -413,11 +412,6 @@ impl<
         let Some((retry, nullify)) = self.state.construct_nullify(view, reason) else {
             return (self, None);
         };
-
-        // Inform the batcher on every attempt (it ignores duplicate nullifies):
-        // after a restart, the first attempt for a replayed nullify is a retry,
-        // and the batcher (whose state is not persisted) has not seen the vote yet.
-        batcher.constructed(Vote::Nullify(nullify.clone()));
 
         // Persist the nullify if it is a first attempt
         if !retry {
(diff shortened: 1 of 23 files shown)


consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -134,7 +136,7 @@ pub struct Actor<
     write_buffer: NonZeroUsize,
     page_cache: CacheRef,
     journal: Option<Journal<E, Artifact<S, D>>>,
-    dirty: bool,
+    dirty_section: Option<View>,
```
patrick-ogrady, line 139: This is technically unrelated but not wrong.

PR #4378, opened 2026-07-31
Drop Windows Support

### LINKED ISSUE #2026: [runtime] Handle Non-Durable Windows Directories

patrick-ogrady:
## Summary

Drops Windows as a supported target and documents the platforms Commonware supports.

This PR:

- removes Windows jobs and platform-specific cleanup from CI
- deletes the non-Unix Tokio storage backend and Windows-only runtime branches and tests
- keeps the WebAssembly-specific fallback paths intact
- removes the now-unused `cfg-if` dependency from `commonware-utils`
- documents Linux and macOS support, Linux-only `io_uring` backends, and per-crate WebAssembly and `no_std` compatibility

## Validation

- `just test -p commonware-utils`
- `just test -p commonware-runtime`
- `just clippy -p commonware-utils -p commonware-runtime`
- `just check-docs -p commonware-utils -p commonware-runtime`
- `just test-docs -p commonware-utils -p commonware-runtime`
- release WebAssembly builds for `commonware-utils`, `commonware-runtime`, and `commonware-storage`
- `just check-fmt`
- `just check-toml-fmt`

Closes #2026.


BrendanChou (dismissed):

BrendanChou (approved):

andresilva (approved):
Image

PR #4380, opened 2026-07-31
[storage/qmdb] Authenticate batch ancestry with Commitment

roberto-bayardo:
## Motivation

QMDB batch applicability identifies lineage boundaries by operation count alone. That cannot distinguish equal-size sibling states: after applying sibling parent A, a descendant built on sibling parent B can be accepted because the database size matches the descendant's recorded base size, even though the roots differ — silently grafting the batch onto a state it was never built against.

## Changes

The fix is a representation change: make a QMDB state's identity a single authenticated value, then require exact identity — not merely a matching size — to apply a batch. The remaining changes are consequences of that, not independent scope.

**The fix**

- introduce `Commitment<F, D>` — identifies a QMDB state by its operation `size` (a typed `Location`, consistent with sync's `CompactTarget`) and authenticated `root`. Size and root become one inseparable value that cannot describe two different states.
- a batch's `Bounds` records its boundaries as commitments — `base_commit`, `db_commit`, and `tip_commit` (the tip replacing the former bare `total_size`; each merkleized batch's root now lives in `tip_commit.root` rather than a separate field)
- require the live database to match an exact recorded `Commitment` (size **and** root) before applying a batch — the equal-size sibling case that size-only applicability silently accepted is now rejected

**Follows from the new representation**

- applicability collapses to `current == db_commit || ancestors.any(...)`: with every boundary carried as a commitment, lineage checks are direct comparisons, so the prior size-indexed `state_at`/`db_size` lookups are dead and removed
- the compact variants now cache their committed root (read for free from the authenticated witness, matching what `any`/full already do) instead of re-folding peaks — needing a root at each batch boundary would otherwise add hashing to compact's batch-create/validate path
- state that can be derived is no longer stored: an unmerkleized batch derives its DB boundary from its parent instead of keeping a separate field that could drift from it
- related cleanup: `Family::inactive_peaks` takes a leaf `Location` instead of a node `Position`, moving the one `location_to_position` conversion into the merkle layer and dropping it at its ~16 call sites now that sizes are `Location`-typed (`Family::peaks` stays `Position`, its natural tree coordinate)

**Tests**

- add cross-variant regressions for descendants of equal-size sibling parents, retaining conventional sibling-fork coverage

This does not change wire formats, persisted storage, recovery, durability, or fsync behavior.

## Tradeoffs

- each snapshotted ancestor carries an additional digest (its root); ancestor snapshots already grow with pipeline depth, so this increases that constant
- ALPHA `Bounds`/`Commitment` API change (size-only fields → `(size, root)` commitments); because `qmdb::batch_chain` is public, it ripples into downstream consumers (`commonware-glue`, examples) as mechanical updates

## Validation

- `just test -p commonware-storage`
- `just check-stability`
- `just check-docs`
- `just check-fmt`
- `git diff --check`

### DIFF
--- a/storage/src/qmdb/batch_chain.rs
+++ b/storage/src/qmdb/batch_chain.rs
@@ -3,8 +3,8 @@
 //! A batch chain is a linked sequence of in-memory batches built on top of a DB state. Each batch
 //! records its state via [`Bounds`] (where its operations sit in the log, and the root at each
 //! applicable boundary) and the inactivity floor declared by its commit. Older batches in the chain
-//! are tracked as [`AncestorBounds`] in newest-first order; some may already be on disk and others
-//! may not.
+//! are tracked as [`AncestorBounds`] in newest-first order. Some may already be applied to the
+//! database while others may not.
 //!
 //! Before applying a batch to the DB, the internal validation checks two things shared across QMDB
 //! variants (any, immutable, keyless):
@@ -12,8 +12,9 @@
 //! - The batch is not stale: the current DB state must match either the batch's recorded DB state
 //!   or one of its ancestor states.
 //! - Commit floors are monotonically non-decreasing along the chain, and no floor exceeds
-//!   its own commit location. Ancestors already on disk are skipped (their floors were
-//!   validated when they were first applied); the rest of the chain and the tip are checked.
+//!   its own commit location. Ancestors already applied to the database are skipped because their
+//!   floors were validated when they were first applied. The rest of the chain and the tip are
+//!   checked.
 //!
 //! Internal helpers walk the chain via weak parent references and snapshot ancestor bounds into a
 //! `Vec` for storage on a merkleized batch.
@@ -29,7 +30,7 @@ use std::sync::{Arc, Weak};
 /// Identifies a QMDB state by its operation `size` and authenticated `root`.
 #[derive(Clone, Copy, Debug)]
 pub struct Commitment<F: Family, D: Digest> {
-    /// Number of operations committed by the state.
+    /// Number of operations in the state.
     pub size: Location<F>,
     /// Root committing to those operations.
     pub root: D,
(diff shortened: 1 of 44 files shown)


roberto-bayardo:
bugbot run

storage/src/qmdb/batch_chain.rs
```diff
@@ -1,85 +1,114 @@
...
     merkle::{Family, Location},
     qmdb::Error,
 };
+use commonware_cryptography::Digest;
 use core::iter;
 use std::sync::{Arc, Weak};
 
+/// Identifies a QMDB state by its operation `size` and authenticated `root`.
+#[derive(Clone, Copy, Debug)]
+pub struct CommitId<F: Family, D: Digest> {
```
patrick-ogrady: Can we just call this `Commit` or `Target`?

roberto-bayardo: `Target` suggests a point we're trying to reach rather than a point we've already reached (even if tentative) so I guess I'd prefer `Commit` between the two, but `commit` is already pretty overloaded.  How about `Commitment`?

patrick-ogrady (dismissed):

storage/src/qmdb/immutable/batch.rs
```diff
@@ -53,12 +53,9 @@ where
...
 
-    /// Total operation count before this batch (committed DB + prior batches).
-    /// This batch's i-th operation lands at location `base_size + i`.
-    base_size: u64,
-
-    /// The database size when this batch was created, used to detect stale batches.
-    db_size: u64,
+    /// The committed state immediately before this batch's operations (committed DB + prior
+    /// batches). This batch's i-th operation lands at location `base.size + i`.
+    base: Commitment<F, H::Digest>,
```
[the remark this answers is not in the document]
patrick-ogrady: (working on)

patrick-ogrady (approved):

PR #4384, opened 2026-08-01
[glue] Use `start_sync`

patrick-ogrady:
Stateful now separates an applied finalization from the durability of its newly launched database flush. Proposal construction can use the applied state while that flush is in flight, without allowing Marshal's processed floor to advance past durable QMDB state.

This supersedes #4325 for Stateful's `start_sync` integration. The broader historical-target QMDB sync-serving work from that PR remains out of scope.

Each `ManagedDb::finalize` applies its merkleized batch and calls `start_sync`, returning the updated database plus a durability `Handle`. `DatabaseSet::finalize` restores the applied databases and collects those handles into a `Barrier`. `Application::finalized` runs once this state is readable; it is not a durability notification. A later `start_sync` may still queue behind an earlier ordered flush. The processing loop observes barriers between messages and while idle, and acknowledges Marshal only after the corresponding barrier is durable. A non-shutdown flush failure is fatal. Runtime shutdown or dropping unfinished Stateful work abandons its acknowledgement without resolving or canceling the waiter, so the processed floor does not advance and the block can be redelivered after restart.

The public `PruneConfig` controls only maintenance cadence and extra Marshal/QMDB retention. At each Stateful initialization, Glue reads `max_pending_acks` from the Marshal mailbox, sets the rewind-safe base window to `max_pending_acks + 1`, adds the configured retention windows, and samples a maintenance phase from the runtime RNG. Before pruning, the actor observes every finalize barrier through the QMDB prune target. Barriers for later state may remain pending because Marshal retains those blocks for replay, while database pruning coordinates destructive work with its in-flight syncs.

During one-time state sync, finalized blocks remain paired with their Marshal acknowledgements until the pending-acknowledgement window reaches capacity. The full window intentionally backpressures Marshal while the actor has the live sync coordinator record the newest block as its current target; after that observation, it releases the window. This replaces timer-based retargeting. If sync completes first or races the observation, retained blocks are handed off in order. Blocks beyond the completed artifact are applied and flushed, and completion metadata is persisted only after every newly applied handoff barrier is durable. Live retargets remain volatile.

For state-sync startup, Glue durably records the selected or persisted finalized floor before mutating the databases. A newer certified startup selection may replace an interrupted floor. `Marshal::Actor::init` now returns a public `Floor` containing its independent durable processed height and round bounds alongside the mailbox, and callers pass that pair into Stateful. Before registering a local-only waiter for the selected floor block, Syncer applies Marshal's same inclusive round-supersession rule. If Marshal's startup snapshot already supersedes that selection, the block may have been pruned, so Syncer resolves from the retained processed anchor instead of waiting indefinitely. A newly installed Marshal floor records its anchor's predecessor as processed so the anchor itself is dispatched; when that predecessor is absent, Glue uses only the retained successor as the processed anchor.

State-sync journal recovery reuses append-only history only when it still contains the requested range start. Operation and full-Merkle journals that extend beyond an older target are rewound to its end; journals that have already pruned the requested start are reset there and rebuilt. Full-Merkle reinitialization removes stale pins from abandoned targets. Plain full-Merkle recovery also rejects an effective pruning boundary above the last structurally valid recovered journal size before consulting pinned-node metadata. It returns `MissingNode`, which QMDB treats as unavailable local pins and resolves through a peer-authenticated boundary.

After publishing a completed state-sync artifact, Syncer drops the live tip-update ring sender. Any update stranded after the coordinator's final drain loses its observation, retries through Syncer, and receives the cached artifact instead of remaining pending.


glue/src/stateful/actor/core/processing.rs
```diff
@@ -146,32 +178,60 @@ where
...
+                        // Marshal's ack window bounds the flush backlog. A
+                        // false `Barrier::durable` leaves the block
+                        // unacknowledged, and marshal redelivers it on restart.
+                        syncs.push(async move {
+                            let durable = barrier.durable().await;
+                            if durable {
+                                acknowledgement.acknowledge();
+                            }
+                            durable
+                        });
```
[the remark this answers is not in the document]
patrick-ogrady: Thanks — the cancellation mechanics are correct, but this is the intended acknowledgement contract rather than a lost-redelivery bug. Update::Block documents that dropping its acknowledgement makes marshal exit when the application is shutting down. In handle_ack, an Err panics before the canceled height updates or syncs the marshal processed floor, so that block remains unprocessed and is redelivered after restart. Barrier::durable returns false only for Closed/Aborted runtime shutdown; real flush failures panic. Propagating false here is still required so an aborted flush cannot be forgotten and followed by prune. Holding the acknowledgement pending after its owner has exited is neither required for redelivery nor supported by Exact. Resolving as not actionable.

glue/src/stateful/actor/core/processing.rs
```diff
@@ -70,9 +79,24 @@ where
...
+        let mut syncs = Pool::<(Height, bool)>::default();
+        let mut pending_syncs = BTreeSet::new();
         select_loop! {
             self.context,
             on_start => {
+                // Observe every already-completed flush (releasing its marshal
+                // acknowledgement) before taking the next unit of work, so
+                // acknowledgements keep flowing even while the mailbox is
+                // never idle.
+                while let Poll::Ready(completion) = poll!(syncs.next_completed()) {
```
clabby: Could use `now_or_never` here to avoid having to pull in `poll!`.

glue/src/stateful/actor/core/processing.rs
```diff
@@ -82,8 +106,27 @@ where
...
+                        None => {
+                            let mailbox = &mut self.mailbox;
+                            let syncs = &mut syncs;
+                            let pending_syncs = &mut pending_syncs;
+                            Either::Right(async move {
+                                loop {
+                                    select! {
+                                        message = mailbox.recv() => {
+                                            break message.map(Step::Message);
+                                        },
```
clabby, line 120: nit: Leave a short note about why we don’t use `select_loop!` with graceful shutdown handling here -> this future is raced against `context.shutdown()` below.

glue/src/stateful/actor/processor/mod.rs
```diff
@@ -136,28 +152,36 @@ impl<T: Clone> Pruning<T> {
...
     }
 
+    #[cfg(test)]
+    pub(super) fn fixed(
+        config: PruneConfig,
+        max_pending_acks: usize,
+        maintenance_offset: u64,
+    ) -> Self {
+        Self::build(config, max_pending_acks, maintenance_offset)
+    }
```
clabby: Just make `build` `pub(super)`?

glue/src/stateful/db/mod.rs
```diff
@@ -337,6 +343,74 @@ pub trait ManagedDb<E>: Send + Sync + Sized {
...
+
+    /// Resolves `true` once every deferred flush is durable.
+    ///
+    /// A flush failure panics because the database has already advanced past
+    /// the unflushed batch. Returns `false` only when runtime shutdown aborts
+    /// or closes a flush handle.
+    pub async fn durable(self) -> bool {
+        let mut durable = true;
+        for (db_type, index, handle) in self.syncs {
+            match handle.await {
```
clabby: Could `try_join_all` here, I believe? The tasks are already kicked off, shouldn’t need to await them one-by-one.

glue/src/stateful/actor/core/syncing.rs
```diff
@@ -191,91 +205,172 @@ where
...
+            .pending_finalizations
+            .back()
+            .expect("full acknowledgement window must contain a block")
+            .block
+            .clone();
+        let observed;
+        (self, observed) = self.update_target(&newest).await;
+        if !observed {
+            return (self, None);
+        }
```
clabby, line 240: This is clean :eyes:

glue/src/stateful/actor/core/processing.rs
clabby, line 117: Still probably deserves a comment on "why not `select_loop!`"

consensus/src/marshal/core/floor.rs
```diff
@@ -69,6 +82,10 @@ impl<S: Scheme, C: Digest> Floor<S, C> {
         }
     }
 
+    pub(super) const fn processed(&self) -> Floor {
+        self.processed
+    }
```
clabby: Having `processed_round` feels a bit unnecessary after exposing the full `Floor`?

glue/src/stateful/actor/syncer/actor.rs
```diff
@@ -49,8 +49,8 @@ where
     /// Finalized floor marshal should resolve before sync starts.
     pub finalization: Finalization<S, V::Commitment>,
 
-    /// Marshal mailbox used to query the finalized floor.
-    pub marshal: MarshalMailbox<S, V>,
+    /// Marshal mailbox and the durable floor returned with it during initialization.
+    pub marshal: (MarshalMailbox<S, V>, Floor),
```
clabby, line 53: Let's add this to a struct so that L128-133 are a bit more readable?

patrick-ogrady: prefer the tuple here.

consensus/src/marshal/core/stream.rs
```diff
@@ -1,50 +1,62 @@
...
+
+    const fn progress(self) -> Progress {
+        (self.height, self.round)
     }
 }
 
 /// Application delivery stream progress and durable metadata.
 pub(super) struct Stream<E: Context> {
-    metadata: Metadata<E, U64, Height>,
+    metadata: Metadata<E, U64, Progress>,
```
patrick-ogrady: I don't think we can do this

clabby (approved):

PR #4388, opened 2026-08-02
[storage] Reuse shared helpers across qmdb, merkle, and bitmap

roberto-bayardo:
## Motivation

A batch of small simplifications across the storage crate, found while scrutinizing the qmdb `current` (grafted) variant. The common thread: several sites reimplemented logic an existing helper/method already provides, or duplicated a short loop/expression across multiple call sites. Each change replaces the inline logic with a call to the existing helper, deduplicates the repeat, and removes the dead code it subsumes.

## Changes

- **`qmdb::current::db`** — extract `grafted_pinned_nodes`, unifying the triplicated `nodes_to_pin` + `get_node` read (in `prune`, `rewind`, `sync_metadata`) and single-sourcing the `DataCorrupted("missing grafted pinned node")` invariant.
- **`qmdb::current::batch`** — `compute_current_layer` reuses the shared `db::partial_chunk` instead of re-deriving the partial-chunk logic inline. This also makes the commit-time root agree with the canonical witness/recovery recompute path on an *empty* bitmap, where the inline version previously returned `None` — a state unreachable in practice, since every batch carries a CommitFloor bit.
- **`bitmap::authenticated`** — `complete_chunks` delegates to the wrapped `PrunableBitMap::complete_chunks` instead of re-deriving it (verified algebraically identical).
- **`qmdb::current::ordered::variable`** — tighten the `init` bound from `Operation: Codec` to `Operation: Read` (matching `unordered::variable`; only `Read::Cfg` is used), dropping the now-unused `Codec` import.
- **`merkle`** — add `Blueprint::required_positions` (deduplicates the "nodes this blueprint needs" set built identically at 3 sites) and `Blueprint::sibling_start` (replaces two single-use count methods, deduplicated at 4 sites across proof construction/verification); also rename the misleading `siblings_start` local in `split_proof_digests` to `suffix_start`.
- **`qmdb`** — merge two pairs of artificially-split `impl` blocks (`current::proof` `OperationProof`, `sync::source` `Request`).

No wire, storage, recovery, durability, or fsync change; every change is behavior-preserving on all reachable states.

## Validation

- `just test -p commonware-storage` — 2893 tests pass
- `just check-fmt`

### DIFF
--- a/storage/src/bitmap/authenticated.rs
+++ b/storage/src/bitmap/authenticated.rs
@@ -172,13 +172,7 @@ impl<E: Context, D: Digest, const N: usize, M: State<D>, S: Strategy> BitMap<E, 
     /// The returned index is absolute and includes pruned chunks.
     #[inline]
     fn complete_chunks(&self) -> usize {
-        let chunks_len = self.bitmap.chunks_len();
-        if self.bitmap.is_chunk_aligned() {
-            chunks_len
-        } else {
-            // Last chunk is partial
-            chunks_len.checked_sub(1).unwrap()
-        }
+        self.bitmap.complete_chunks()
     }
 
     /// Return the last chunk of the bitmap and its size in bits. The size can be 0 (meaning the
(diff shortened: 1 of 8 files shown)


roberto-bayardo:
bugbot run

patrick-ogrady (approved):
very nice

PR #4389, opened 2026-08-02
[marshal] Improve Pruning Logic

patrick-ogrady:
## Motivation

Marshal's coding buffer treated a finalized commitment as both an exact retirement and an implicit global pruning cursor. Those are independent signals: application acknowledgements may be batched or superseded by a floor, the same commitment may be observed in a later view, and the exact commitment may be absent from the local shard cache. Deriving retention from the target's cached metadata could therefore either retain stale state or retire state still needed by consensus.

## Bugs fixed

- **Batched acknowledgements lost exact retirements.** Core durably advanced every ready acknowledgement but reported only the final commitment, so earlier acknowledged commitments could remain cached when their observation rounds were above the floor.
- **Floor transitions lost retirement ownership.** Installing a floor discarded in-flight acknowledgement identities and did not advance variant-buffer retention. Superseded commitments could leak, while retiring every cleared acknowledgement would remove above-floor blocks that must be re-dispatched.
- **The implicit pruning cutoff was unsafe and incomplete.** Coding pruning compared cached blocks by application-supplied height, inferred the state cutoff from the exact target's observation round, and became a no-op when that target was unseen. Adversarial heights and re-proposals could cause under- or over-pruning, including retaining entries observed exactly at the floor.
- **Floor pruning canceled recoverable waits.** It closed commitment subscriptions for round-retired reconstruction state even though later local ingress could still satisfy them, causing callers that do not retry to miss an available block.
- **Later observations did not keep live commitments alive.** Discovery, notarization, and local re-proposal did not consistently refresh one shared retention round across cache and reconstruction state, so a later floor could retire state still needed for shard readiness or certification.
- **Commitments could be rebound across epochs.** Cache and reconstruction state could associate the same commitment with different epochs, pinning old state and classifying shards against the wrong participant set.
- **Re-proposal recovery could stall.** A node that missed the original block only waited for local ingress, and a valid re-proposal did not recreate shard state after exact retirement.
- **Assigned-shard readiness was tied to the original leader.** A valid assigned-index shard relayed by another participant could be rejected and the honest sender blocked, preventing progress despite successful commitment-bound verification.

## Changes

- Replace `Buffer::finalized(commitment)` with an atomic `Retirement` carrying an inclusive trusted round floor and independent exact commitment retirements.
- Preserve cleared acknowledgement metadata through floor activation, retire only acknowledgements covered by the durable processed-height floor, and leave higher blocks live for re-dispatch.
- Give each coding commitment one lifecycle owner for its cached block, optional reconstruction state, and latest valid same-epoch observation.
- Apply round retirement even when an exact target is unseen, and distinguish floor retirement from exact retirement when closing subscriptions.
- Fetch re-proposed epoch-boundary blocks by their certified parent round and recreate shard state only after validating the re-proposal.
- Accept the local assigned-index shard from any epoch participant after eager cryptographic verification; other gossiped shards remain sender-owned.

## Testing

- `just test -p commonware-consensus`
- `just clippy -p commonware-consensus`
- `just check-stability`
- `just check-fmt`

### DIFF
--- a/consensus/src/marshal/coding/shards/engine.rs
+++ b/consensus/src/marshal/coding/shards/engine.rs
@@ -1494,6 +1494,14 @@ where
     /// Re-proposals reuse commitments, and application acknowledgments can lag consensus.
     /// Retaining the latest observed round prevents an older acknowledgment from retiring
     /// reconstruction still needed by a later proposal.
+    ///
+    /// The refreshed round also governs shard classification, which resolves the
+    /// participant set from `round`'s epoch. That is unambiguous only because every
+    /// observation path binds the commitment to a single epoch: proposals bind it in
+    /// the context digest, re-proposals are validated as the last block of the
+    /// announced round's own epoch, and notarizations imply such a proposal. An
+    /// epocher that allowed a block to close more than one epoch would break this
+    /// invariant.
     fn observe(&mut self, round: Round) {
         let common = self.common_mut();
         common.round = common.round.max(round);
--- a/consensus/src/marshal/coding/shards/engine.rs
+++ b/consensus/src/marshal/coding/shards/engine.rs
@@ -710,12 +710,10 @@ where
         // redundant, unless notarized recovery created leaderless state first.
         // In that case, the leader announcement must still populate the
         // leader-dependent path.
-        let block_cached = if let Some(entry) = self.reconstructed_blocks.get_mut(&commitment) {
-            entry.round = entry.round.max(round);
-            true
-        } else {
-            false
-        };
+        let block_cached = self.reconstructed_blocks.contains_key(&commitment);
+        if block_cached {
+            self.observe_existing_commitment(commitment, round);
+        }
         if block_cached
             && self
                 .state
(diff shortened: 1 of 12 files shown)


consensus/src/marshal/coding/shards/engine.rs
```diff
@@ -2285,70 +2496,212 @@ mod tests {
     }
 
     #[test_traced]
-    fn test_durable_prunes_reconstructed_blocks() {
+    fn test_prune_uses_inclusive_retirement_floor() {
```
clabby: Test names are a bit stale, should be `test_retire_*`?

clabby (dismissed):

consensus/src/marshal/coding/shards/engine.rs
clabby, line 1: The state diagram says `Ready` is frozen and accepts no new shards, but `on_network_shard` intentionally accepts a late assigned shard in `Ready`?

clabby (dismissed):

clabby (approved):

PR #4391, opened 2026-08-03
[storage/merkle] Make build_proof the single source of truth for proof layout

roberto-bayardo:
## Motivation

`Proof::verify_multi_inclusion` reconstructs each element's sub-proof by hand-assembling its digest vector in the canonical proof layout — `[fold_acc? | active peaks | after-peaks | suffix_acc? | DFS siblings]`. That is exactly what the prover, `Blueprint::build_proof`, produces. So the **most subtle logic in this module was duplicated across the prover and the verifier** — two places that must agree byte-for-byte but could silently drift apart, a latent proof-incompatibility risk.

## Changes

Make `Blueprint::build_proof` the single source of truth for the layout:

- **`build_proof` borrows `&self`** (it no longer needs to consume the blueprint) and reads the active prefix + after peaks from **`fetch_nodes[..sibling_start()]`** instead of the separate `prefix_active_peaks`/`after_peaks` fields. `Blueprint::new` constructs `fetch_nodes` as `[prefix_active | after_peaks | DFS siblings]`, so these are the identical positions in the identical order. The `fetch_nodes` field doc — which previously omitted the leading active-peaks section — is corrected to state the real layout.
- **`verify_multi_inclusion` now calls `build_proof`** with a closure reading the already-collected `node_digests` map, instead of re-implementing the assembly. Every position is present by construction (the position set is checked against `digests.len()` first), so `build_proof` cannot fail here.

Net **−32 LOC**, and the prover/verifier layout can no longer drift.

## Validation

- `just test -p commonware-storage` — 2894 tests pass (the build↔verify proof round-trips across all variants are the key coverage)
- `just check-fmt`

No wire, storage, recovery, or durability change. Builds on the `sibling_start`/`required_positions`/`fetch_nodes` groundwork from #4388 (now in `main`).

### DIFF
--- a/storage/src/merkle/proof.rs
+++ b/storage/src/merkle/proof.rs
@@ -231,10 +231,9 @@ impl<F: Family, D: Digest> Proof<F, D> {
             .map(|(&pos, digest)| (pos, *digest))
             .collect();
 
-        // Verify each element by reconstructing its sub-proof in the canonical layout (via the
-        // shared `build_proof`) and checking its root. Every required position was collected above
-        // and matched against `digests.len()`, so `node_digests` is complete and `build_proof`
-        // cannot fail here.
+        // Verify each element by reconstructing its sub-proof in the canonical layout and checking
+        // its root. Every required position was collected above and matched against `digests.len()`,
+        // so `node_digests` is complete and `build_proof` cannot fail here.
         for (element, loc) in elements {
             let bp = &blueprints[loc];
             let proof = bp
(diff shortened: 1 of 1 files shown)


storage/src/merkle/proof.rs
```diff
@@ -927,6 +896,9 @@ impl<F: Family> Blueprint<F> {
     }
 
     /// Return active after-peaks that are collapsed into a backward-folded suffix accumulator.
+    // Only the `std`-gated `verification` module consults this accessor; `build_proof` reads the
+    // `suffix_peaks` field directly.
```
danlaine:
```suggestion
```

patrick-ogrady (approved):

PR #4392, opened 2026-08-03
[storage/qmdb] Share the commit-operation codec across db families

roberto-bayardo:
## Motivation

The `any`, `immutable`, and `keyless` QMDB families each define a commit operation with the **same shape** — `(Option<value>, Location)` (an optional metadata value and an inactivity floor) — encoded **identically** on the wire. Its encode/decode was copy-pasted **six times** (three families × fixed/variable): the option tag, the padded-or-absent value, the floor-overflow check (fixed), and the floor varint (variable). Six hand-maintained copies of a roundtrip-critical codec that must stay byte-compatible — a latent drift risk.

## Changes

Add a shared commit-payload codec in `qmdb::operation` and call it from every family:

- `commit_fixed_size` / `write_commit_fixed` / `read_commit_fixed` (fixed encoding)
- `commit_variable_size` / `write_commit_variable` / `read_commit_variable` (variable encoding)

Each family's codec keeps ownership of what actually differs — its **context byte** (`any`=`0xD3`, `immutable`=`1`, `keyless`=`0`) and, in the fixed encoding, any **trailing padding** to a uniform operation size (`keyless`'s commit is already the largest variant, so it pads nothing). Only the shared commit payload moves into the helpers. The duplicated `commit_op_size` const folds into the shared `commit_fixed_size`.

## Behavior-preserving

The wire format is **unchanged**. `Location` encodes as a varint (`UInt`), so `keyless`'s `floor.write()` and `any`/`immutable`'s `UInt(**floor)` already emitted identical bytes; the fixed floor is a big-endian `u64` in all three. The floor-overflow check and absent-value zero-padding validation are preserved.

## Validation

- **All 19 qmdb operation codec-conformance tests pass** against the stored fixtures (`any` ordered/unordered × fixed/variable × mmr/mmb, `immutable` fixed/var/var-key, `keyless` fixed/var) — proof the format did not drift; **no fixtures regenerated**.
- 73 operation unit tests pass (roundtrips, overflow rejection, padding).
- `cargo check -p commonware-storage` clean under default features **and `--no-default-features`** (`-D warnings`).

This is a dedup for a single source of truth over a correctness-critical codec (net LOC ≈ flat), not a line-count play. Builds on `main`.

### DIFF
--- a/storage/src/qmdb/operation.rs
+++ b/storage/src/qmdb/operation.rs
@@ -51,7 +51,7 @@ pub trait Committable {
     fn is_commit(&self) -> bool;
 }
 
-/// On-disk size of a fixed-encoded commit operation.
+/// Unpadded size of a fixed-encoded commit operation, context byte included.
 pub(crate) const fn commit_fixed_operation_size<V: FixedSize>() -> usize {
     1 + 1 + V::SIZE + u64::SIZE
 }
--- a/storage/src/qmdb/operation.rs
+++ b/storage/src/qmdb/operation.rs
@@ -1,3 +1,5 @@
+//! Shared traits and codecs for QMDB operations.
+
 use crate::merkle::{Family, Location};
 use commonware_codec::{
     CodecShared, EncodeSize, Error as CodecError, FixedSize, Read, ReadExt as _, Write,
@@ -48,9 +50,6 @@ pub trait Committable {
     /// If this operation is a commit operation.
     fn is_commit(&self) -> bool;
 }
-
-// Commit operations are shared across the `any`, `immutable`, and `keyless` families: every commit
-// carries optional metadata and an inactivity floor, encoded identically.
 
 /// On-disk size of a fixed-encoded commit operation.
 pub(crate) const fn commit_fixed_operation_size<V: FixedSize>() -> usize {
--- a/storage/src/qmdb/operation.rs
+++ b/storage/src/qmdb/operation.rs
@@ -1,5 +1,9 @@
 use crate::merkle::{Family, Location};
-use commonware_codec::CodecShared;
+use commonware_codec::{
+    CodecShared, EncodeSize, Error as CodecError, FixedSize, Read, ReadExt as _, Write,
+    util::ensure_zeros,
+};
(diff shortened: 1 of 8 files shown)


storage/src/qmdb/operation.rs
```diff
@@ -44,3 +48,74 @@ pub trait Committable {
     /// If this operation is a commit operation.
     fn is_commit(&self) -> bool;
 }
+
+// Commit operations are shared across the `any`, `immutable`, and `keyless` families: every commit
+// carries optional metadata and an inactivity floor, encoded identically.
+
+/// On-disk size of a fixed-encoded commit operation.
+pub(crate) const fn commit_fixed_operation_size<V: FixedSize>() -> usize {
```
danlaine, line 55: Seems like perhaps we should have a shared `Commit` type that we define encoding/decoding on?

roberto-bayardo: Ah I see like Update does with Fixed/Variable encoding variants .. looking.

roberto-bayardo: Considered this, but I think `Commit` is a weaker candidate for a type than it first looks.

`Update` earns being a type: it's a trait-based polymorphism point (ordered vs unordered, with associated `Key`/`Value`/`ValueEncoding` and `key()`/`into_key()`), so `Operation<F, S: Update>` needs it to abstract two genuinely different shapes.

`Commit` has no such variation. It's a single fixed `(Option<value>, inactivity floor)` across all three families, with no behavior beyond the codec. So a `Commit` type would just wrap a 2-tuple and re-home the encode/decode these helpers already centralize. To make the codec genuinely live *on* the type (rather than as associated fns the write path never constructs), it'd have to be embedded in each `Operation` variant — ~2 dozen construction/match edits across `apply`/`batch`/`db` + tests in all three families, for no additional dedup, correctness, or perf.

So I'd lean toward keeping the shared helpers as-is.

patrick-ogrady (dismissed):

patrick-ogrady (approved):

PR #4398, opened 2026-08-03
[simplex] Split repair demand into kind and until

BrendanChou:
Targets `simplex-repair-cleanup` (#4386).

This reworks how the simplex resolver tracks what it still wants. Possession replaces certification on both sides of the wire: a fetch completes once the requested notarization arrives, and a responder serves a cached notarization without waiting for its own verdict. A notarization whose certification fails leaves the cache, since certification judges the evidence itself and no other copy could pass. Those are the behavior changes; the rest reshapes the demand bookkeeping.

## Split `Purpose` into two axes

`Purpose` on the base branch names three cases without naming the two dimensions they vary along, which makes its retirement rules hard to follow:

```rust
struct Demand {
    kind: Kind,    // Nullification | Notarization  -- what settles the request
    until: Until,  // Floor | Finalization          -- what retires the request
}
```

`Backfill` is `(Nullification, Floor)`, `Nullification` is `(Nullification, Finalization)`, and `Parent` is `(Notarization, Finalization)`. The fourth combination does not occur, because background repair only ever wants a nullification.

The two axes cannot be collapsed into one. Retiring an `Until::Finalization` demand at the floor breaks the certificate split that #4386 addresses: the floor may be an unfinalized certified notarization, and a proposal can still name ancestry below it. Holding an `Until::Floor` demand until finalization is a griefing vector, because views that are notarized and certified but never finalized would accumulate permanently unsatisfiable demands.

## One predicate for retirement

`settled(view, kind)` replaces `purpose_resolved`, `purpose_satisfied`, `is_retired_by`, and `retire`. It answers one question: is there anything left to fetch for this view and kind. Every site that needs retirement asks it.

The resolver stays the sole owner of the outstanding demands. Each piece of evidence publishes what it settles through `retire`, which takes an owned predicate naming a span of views and a kind. That is sound because settlement is monotonic: no later evidence unsettles a demand, so a retirement stays true however the resolver mailbox orders it against fetches, which it reorders under backpressure.

A valid response still does not imply a settled demand. The resolver key names only a view, so a peer may answer a notarization request with a covering nullification. That is valid evidence the requester records, but it is not what was asked for. The requester reports `Outcome::Ambiguous` and does not fault the peer, as in #4383.

`validate` also becomes a single match over the certificate, where the base branch uses per-kind arms with duplicated epoch and verification checks. It still accepts any certificate an honest peer could serve for the view, including a higher-view notarization served off a peer's floor.

### DIFF
--- a/consensus/src/simplex/actors/resolver/actor.rs
+++ b/consensus/src/simplex/actors/resolver/actor.rs
@@ -81,13 +81,6 @@ pub struct Actor<
     /// arrive, so the tombstone is kept here against the finalization boundary.
     failed_certifications: BTreeSet<View>,
 
-    /// Certificates still wanted from peers, paired with what retires each.
-    ///
-    /// This mirrors the resolver's outstanding fetches. Owning them here keeps
-    /// retirement in one place: [Self::settled] decides it from local evidence,
-    /// and [Self::sync_demands] publishes the survivors to the resolver.
-    demands: BTreeSet<(View, Demand)>,
-
     mailbox_receiver: mailbox::Receiver<MailboxMessage<S, D>>,
 }
 
@@ -117,7 +110,6 @@ impl<
                 nullification_responses: BTreeMap::new(),
                 notarization_responses: BTreeMap::new(),
                 failed_certifications: BTreeSet::new(),
-                demands: BTreeSet::new(),
 
                 mailbox_receiver: receiver,
             },
@@ -225,14 +217,20 @@ impl<
             Certificate::Notarization(notarization) => (None, None, Some(notarization.view())),
         };
 
+        let term_length = self.state.term_length();
+
         // Cache response payloads for as long as a peer can still ask for them.
         // [State] prunes at the floor, which rises sooner than finalization and
         // would hide this evidence from a peer still repairing the view.
         if let Some(nullified) = nullified
-            && nullified.term_end(self.state.term_length()) > self.last_finalized
+            && nullified.term_end(term_length) > self.last_finalized
         {
             self.nullification_responses
                 .insert(nullified, certificate.encode());
+            let covered = nullified..=nullified.term_end(term_length);
+            Self::retire(resolver, move |view, demand| {
+                demand.kind == Kind::Nullification && covered.contains(&view)
+            });
         }
         if let Some(notarized) = notarized
             && notarized > self.last_finalized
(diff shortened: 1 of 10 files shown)


consensus/src/simplex/actors/resolver/actor.rs
```diff
@@ -780,27 +579,22 @@ impl<
...
-                response.send_lossy(Outcome::Ambiguous);
+                // The peer answered the view it was asked about, so it is never
+                // faulted here. Whether that answer was the certificate anyone
+                // wanted is a separate question: if a demand for this view is
+                // still open, the response was valid but ambiguous, and the
+                // resolver retries without penalizing the peer.
+                let outcome = if demands.iter().all(|demand| self.settled(view, demand.kind)) {
+                    Outcome::Complete
+                } else {
+                    Outcome::Ambiguous
```
patrick-ogrady, line 608: I think this can busy loop (idea with certification was to at least wait + not treat any plain certificate that is better as immediately retriable)?

BrendanChou: Shouldn't busyloop. `Ambiguous` goes through `Fetcher::add_retry` which will retry the key with a deadline `fetch_retry_timeout` into the future. So if the certification is taking a long time, it might retry the request. However, once certification comes back true, then the pending retry will be pruned so no additional retry is sent at all, which I expect to be the normal case. If the certification comes back false, then it does need to retry since it is missing the nullification.

patrick-ogrady: Ah, I was misremembering the order there. Hm.

I liked the idea of this waiting certification_timeout before retrying...I wonder if we should add that as an option for ambigious/invalid to overwrite retry.

patrick-ogrady: I guess my thinking is `fetch_timeout` covers the expected time it takes to send/receive the request and `certificate_timeout` is the expected time it could take to process such a certificate (it feels like they should be additive)?

BrendanChou: I see what you're saying, but I think it's better not to introduce this due to complexity. Waiting longer only helps backfill demands, because a success verdict retires demands only by raising the floor. An ancestry demand for a nullification must retry whatever the verdict is.

BrendanChou: Increasing the wait time also delays recovery for a `certified: false` verdict

patrick-ogrady: With the resolver tweaks in #4403 (and in the forthcoming #4405), I'm cool with this.

PR #4400, opened 2026-08-03
[storage/qmdb] Add snapshots

danlaine:
Adds cheap, owned snapshots to QMDB so reads, proofs, and state-sync responses can be served without locking the database.

## Changes

- The authenticated journal is now generic over its item and Merkle node sources: live `Journal` and frozen `Snapshot` are two aliases of one type, sharing a single implementation of reads, proofs, and `Source`. Mutating methods stay live-only.
- `snapshot()` on `Any`, `Current`, `Immutable`, and `Keyless` returns the frozen operations log itself. It serves reads, proofs, and sync, but keeps no keyed index or activity bitmap, so `Current` snapshots serve ops-tree proofs rather than grafted activity proofs.
- Capture takes owned blob readers and shares the in-memory Merkle tail copy-on-write: nothing is copied up front, and the snapshot stays readable while the live side appends or prunes.
- A compact snapshot is `Arc<Tip>`: the latest commit's operation, root, proof, and pinned nodes, which the database already holds. Capture is a refcount increment. A tip installed by `start_sync` is not proven durable until its handle completes.
- The Merkle storage trait gets `Family` as an associated type and a `pinned_nodes_at` method, so live and frozen node stores share the proof code.
- Renames: the in-memory `snapshot` indexes to `index`, and the compact Merkle `snapshot()` accessor to `mem()`.

Snapshots keep captured data readable across append and prune. Rewinding into a captured range is intentionally not protected.

## Performance

Capture flushes buffered journal tails (work the next sync would do anyway) and clones handles, not data. The one real cost is the Merkle tail's copy-on-write, the same `Arc::make_mut` pattern parallel merkleization already uses: while a snapshot is alive, the live side's first mutation after capture copies the in-memory tail once. Databases that never snapshot pay nothing.


danlaine:
Superseded by #4532, which rebuilds this on top of the batch read gates (#4531).

PR #4403, opened 2026-08-04
[resolver] Remove Simplex scan limit and prioritize fresh requests

patrick-ogrady:
## Motivation

Simplex `fetch_concurrent` no longer bounded concurrent or outstanding resolver work. It only limited the anchors emitted by one state scan, which could leave available P2P bandwidth unused. In the P2P resolver, pending work was ordered by deadline before distinguishing fresh requests from retries, allowing overdue retries to consume capacity ahead of fresh keys already waiting to be sent.

## Changes

- remove `fetch_concurrent` from Simplex configuration and register every uncovered missing term anchor;
- attempt all pending fresh requests before any pending retries;
- order fresh requests and retries separately by their next-attempt times;
- attempt at most one due outbound request after every completed event-loop turn, using the pending deadline only to wake the loop;
- scan pending requests without cloning the full pending set;
- document in the P2P resolver that resource usage depends on the rate-limiting configuration of the underlying P2P network.

## Tradeoffs

P2P per-peer rate limits pace outbound wire requests; they do not cap local pending demand. A large view gap can therefore register `O(view_gap / term_length)` pending anchor fetches at once.


examples/bridge/src/bin/validator.rs
```diff
@@ -254,7 +254,6 @@ fn main() {
                 fetch_timeout: Duration::from_secs(1),
                 view_retention: ViewDelta::new(10),
                 skip_timeout: Duration::from_secs(11),
-                fetch_concurrent: NZUsize!(32),
```
patrick-ogrady, line 257: This didn't do what you thought it did.

resolver/src/p2p/engine.rs
```diff
@@ -270,6 +284,8 @@ where
                                     self.metrics.fetch_duration.timer(self.context.as_ref()),
                                 );
                                 self.fetcher.add_ready(key);
+                            } else if adds_subscriber && self.fetcher.promote(&key) {
+                                trace!(?key, "promoted retry for new subscriber");
```
patrick-ogrady: We can make this PR smaller if we remove this promotion feature.

resolver/src/subscribers.rs
```diff
@@ -37,6 +37,28 @@ where
...
+    ) -> bool {
+        self.entries.get(key).is_some_and(|subscribers| {
+            subscribers.keys().any(|subscriber| {
+                !delivered
+                    .iter()
+                    .any(|(candidate, _)| candidate == subscriber)
+            })
+        })
+    }
+
```
BrendanChou: Nit: this signature puts the delivery tuple type into the `Subscribers` API, and the spans are not read. A predicate keeps that knowledge at the call site:

```rust
pub fn any_subscriber(&self, key: &K, pred: impl Fn(&S) -> bool) -> bool
```

Fine as is.

resolver/src/p2p/fetcher.rs
```diff
@@ -256,7 +257,7 @@ where
         let pending_keys: Vec<(Key, bool)> = self
             .pending
             .iter()
-            .map(|(k, (_, retry))| (k.clone(), *retry))
+            .map(|(k, (retry, _))| (k.clone(), *retry))
```
BrendanChou: `Fetcher::fetch` clones every pending key, then returns after one successful send. This PR removes the `fetch_concurrent` cap in Simplex, so a deep catch-up registers one pending key per missing term anchor. Draining n pending keys then costs O(n^2) clones.

The current form is cleaner. The keys are small, so the cost may be acceptable. To remove the cost: pop keys from `pending` in priority order and try each one. Reinsert each failed key with the priority that `pop` returned. In the common case, the head key sends and the call does one pop. A full scan occurs only when no key is sendable. `waiter` then defers the next attempt.

BrendanChou (dismissed):

PR #4406, opened 2026-08-04
[glue] Cleanup Spawn

patrick-ogrady:


glue/src/stateful/actor/core/verifications.rs
```diff
@@ -115,23 +119,16 @@ where
...
-                        .with_attributes_from(&request.context.0),
-                    request.context.1.clone(),
-                );
                 let result = select! {
                     _ = invalidated => None,
                     result = verifier.verify(
                         &actor_context,
                         marshal,
-                        attempt_context,
+                        request.context.clone(),
```
clabby: We can’t reconstruct this from just `round`. It drops any other attributes on the request context, so app metrics can get collapsed or mislabeled. We still need to copy the request attributes onto the actor-owned context.

PR #4408, opened 2026-08-04
[runtime] Skip Canceled Timers When Advancing Time

patrick-ogrady:
## Motivation

Polling a sleep registers an alarm with the deterministic executor. Dropping that sleep does not remove the alarm immediately. When a stale alarm was the next deadline and no task was ready, the runner woke it once and then failed its liveness check, even if a later live alarm could make progress.

## Changes

- Continue advancing through alarms that wake no task until runnable work appears or the alarm heap is empty.
- Enforce the configured runtime deadline at every skipped deadline.
- Cover a stale alarm before a live timer and a stale alarm beyond the runtime timeout.

### DIFF
--- a/runtime/src/deterministic.rs
+++ b/runtime/src/deterministic.rs
@@ -416,6 +416,13 @@ impl Executor {
         now
     }
 
+    /// Ensure the runtime has not reached its configured deadline.
+    fn assert_deadline(&self, current: SystemTime) {
+        if self.deadline.is_some_and(|deadline| current >= deadline) {
+            panic!("runtime timeout");
+        }
+    }
+
     /// When idle, jump directly to the next actionable time.
     ///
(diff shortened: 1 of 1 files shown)


runtime/src/deterministic.rs
```diff
@@ -457,6 +457,27 @@ impl Executor {
         }
     }
 
+    /// Wake sleepers until a task is ready or no alarms remain.
+    ///
+    /// Canceling a polled sleep leaves its alarm registered until its deadline. If that alarm
+    /// wakes no task, continue to later deadlines before deciding the runtime has stalled.
+    fn wake_until_ready_or_empty(&self, mut current: SystemTime) {
+        loop {
+            current = self.skip_idle_time(current);
```
patrick-ogrady, line 477: I think this can be deduped with assert_liveness?

PR #4412, opened 2026-08-05
[stream, p2p] Export maximum payload sizes

patrick-ogrady:
Prevent the estimator from configuring simulated P2P with `u32::MAX`, which overflows when the channel identifier is added.

Expose the supported application-payload ceiling owned by each layer:

- `stream::encrypted::MAX_SIZE` reserves the AEAD tag.
- `p2p::authenticated::MAX_SIZE` reserves the authenticated envelope and is re-exported by discovery and lookup.
- `p2p::simulated::MAX_SIZE` reserves the channel identifier.

Use `simulated::MAX_SIZE` in the estimator and cover configuration construction with a regression test. Keep the configuration fields as `u32`; framing-overhead constants remain internal.


p2p/src/simulated/network.rs
```diff
@@ -122,6 +125,8 @@ struct PeerRefCounts {
 pub struct Config {
     /// Maximum size allowed for an application payload provided to a sender.
     ///
+    /// The largest supported value is [`MAX_SIZE`].
```
patrick-ogrady, line 128: almost made a type that allows you to bound min/max for configs but it got way too big.

PR #4417, opened 2026-08-06
[glue] Remove mutable database sharing

danlaine:
## Summary

The stateful actor now owns its database set outright, and read access to a database is a separate value from the authority to mutate it. That separation is what lets this PR delete two things at once: `Shared<DB>` (the `Arc<RwLock<Option<DB>>>` writers and readers used to meet through, where a cancelled mutation left the cell permanently empty), and the entire verification-cancellation machinery.

Verification jobs are no longer cancelled when a block finalizes. A job pauses at its next storage read while the apply runs, then continues, and its verdict is still correct or is refused. Speculative work is no longer thrown away on every finalization.

## Design

**The writer.** `Writer::new(db)` owns the database: the unique, non-`Clone` `Writer` is the sole mutation authority, and it creates cloneable `Reader`s. The old `&databases` parameter bundled three things: read access, mutation authority, and proof the database exists. Unbundled, a job can hold read access indefinitely without blocking mutation, mutation authority stays unique with the set, and nothing needs to be cancelled at all. The compiler enforces the split once instead of the runtime enforcing it continuously.

**Batches carry their own reader.** Every batch type holds a `Reader` and locks the database internally for exactly one storage call per read, so `get`, `merkleize`, and friends are now one-argument calls, and pairing a batch with the wrong database is unrepresentable. Applications never see a handle or a lease, so no lease can span application code: a mutation waits at most one storage call and cannot be starved (the lock is fair and write-preferring). A read of a writer whose owner is gone returns `Closed` (rendered as `Error::Runtime(Closed)` by storage-facing calls). Verification and proposal jobs that hit it produce no verdict, exactly as if they had been cancelled.

**Jobs are `'static`.** `DatabaseSet::new_batches` takes read handles instead of `&self`, so the verification context (`Execution`) holds only handles, shared state, and metrics; it is `Clone + 'static` and each `Verifier` owns one. Jobs are plain pool futures that survive any number of applies. Everything cancellation needed dies with this: `quiesce`/`quiesce_where`, per-job invalidation channels, `VerificationProgress`/`VerificationPhase`, `FinalizationBoundary`/`Disposition`, and the two-phase `run_until_transition`/`apply_transition` loop with its `Quiescent`/`Transition` hand-off types. The loop is now one flat loop; `drive` pins a mutation once and keeps polling job completions alongside it, so an apply never starves verification and verification never delays an apply by more than one storage call.

**Ownership.** Mutating `DatabaseSet` operations consume and return the set, the same convention `ManagedDb` already uses, so mutation capability stays with the actor. `DatabaseSet` loses `Clone`, `Shared` and its guards are deleted, `Single<T>` is the one-database set, and the database-subscription mailbox path is removed. The syncer hands the synced set through the completion channel exactly once; displaced target updates report `UpdateOutcome` instead of carrying a set.

## Retention: why an uncancelled job is still correct

QMDB's branch-validity rule: a merkleized batch stays usable only while everything applied since is an ancestor of that batch. Three cases, each pinned by a storage test in this PR:

1. **The job's batch descends from the finalized block.** The apply is an ancestor, so reads and the root are unchanged. `merkleize_after_compatible_ancestor_apply_matches` (in `any` and `current`) forks two identical children, applies the parent between their merkleizations, and asserts identical roots and reads.
2. **The job's batch is a sibling of the finalized block.** It runs to completion and is refused at `cache_pending`, the last gate in `Verifier::verify`: a result whose branch is no longer reachable from the anchor answers false, the same verdict cancellation used to force, by a different mechanism. Reads and merkleization on such a stale fork are outside the branch-validity contract, so `stale_fork_batch_reads_and_merkleizes_without_panicking` measures them: no panic, covered keys keep the fork's values, uncovered keys fall through to the applied winner, the root stays on the fork's own chain, and applying it fails with `StaleBatch`. A stale fork can therefore never answer true and never crash the node.
3. **Pruning under a live batch.** QMDB prunes only below the inactivity floor and pruning does not change the root; `batch_reads_and_merkleizes_across_a_prune` asserts reads, root, and a later apply all survive.

## API changes

These APIs are ALPHA so breaking changes are OK.

- `Application::propose`, `verify`, and `apply` receive only batches; the batch is the complete view for its branch, including speculative ancestor state the raw databases do not reflect. `finalized`, the one hook that observes applied state, receives read handles (`HandlesOf`).
- The `Application` clone contract is explicit: any method may run on a fresh clone that is discarded afterward, so clones must be cheap and persistent state must live behind shared handles. In-tree applications already comply.
- Batch construction and reads drop their database parameters across all four QMDB wrapper families; `DatabaseSet::new_batches` is a static over `Handles` and reports `Closed` when the owning actor is gone.
- `DatabaseSet` is owned by value and loses `Clone`; `Single<T>` replaces `Shared<T>`. `Readers` is deleted; the reader is used as `live::Reader` and the snapshot reader as `snapshot::Reader`.
- Set mutations still panic on database errors; deferred flush failures still surface through `Barrier`. Nothing about capture or publication changes in this PR.

### DIFF
--- a/examples/reshare/src/application.rs
+++ b/examples/reshare/src/application.rs
@@ -11,7 +11,7 @@ use commonware_glue::{
     dkg::reshare::Input as ReshareInput,
     stateful::{
         Application, Input, Proposed,
-        db::{DatabaseSet, Merkleized as _, Unmerkleized as _},
+        db::{DatabaseSet, Merkleized as _, MerkleizedOf, Unmerkleized as _, UnmerkleizedOf},
     },
 };
 use commonware_runtime::{BufferPooler, Clock, Metrics, Spawner, Storage};
@@ -36,8 +36,8 @@ impl App {
 
     async fn execute<E: Spawner + Metrics + Clock + Storage + BufferPooler>(
         height: Height,
-        batches: <Database<E> as DatabaseSet<E>>::Unmerkleized,
-    ) -> <Database<E> as DatabaseSet<E>>::Merkleized {
+        batches: UnmerkleizedOf<Database<E>, E>,
+    ) -> MerkleizedOf<Database<E>, E> {
         batches
             .write(HEIGHT_KEY, Some(U64::new(height.get())))
             .merkleize()
@@ -65,7 +65,7 @@ where
         &mut self,
         context: (E, Self::Context),
         mut ancestry: impl Ancestry<Self::Block>,
-        batches: <Self::Databases as DatabaseSet<E>>::Unmerkleized,
+        batches: UnmerkleizedOf<Self::Databases, E>,
         input: Input<Self::Input, Self::Provider>,
     ) -> Option<Proposed<Self, E>> {
         // The `reshare::Application` wrapper selected and fetched the payload.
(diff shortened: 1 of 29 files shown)


danlaine:
Superseded by #4534, which rebuilds this on the batch read gates the design needs to be sound.

PR #4418, opened 2026-08-06
[storage/qmdb] Reduce staged update resolution overhead

roberto-bayardo:
## Summary

- reuse the staged key-to-id index to detect overlap with keyed upserts, avoiding a temporary upsert-key set and per-update hashing
- replace the fat winner array sized by the full staged read set with a zero-biased index directory and a compact winner-payload vector
- preserve last-write-wins behavior, upsert precedence, and deterministic location ordering
- release the staged key index and winner directory before allocating or sorting later merkleization bookkeeping

## Performance

Let `R` be the number of distinct staged reads, `M` the number of submitted staged updates, `W` the number of distinct updated keys, and `U` the number of keyed upserts.

Previously, mixed batches built an `O(U)` hash set and performed an overlap lookup for every staged update. Winner tracking also allocated and initialized `R` fat `Option<(usize, Option<Value>)>` entries, plus a touched list with capacity for `M` entries.

The new overlap path performs one lookup per upsert in the existing staged-key index and allocates no second hash table. Winner tracking uses `R` zeroed directory entries and `W` initialized payload entries, with payload capacity bounded by `min(M, R)`; subsequent winner walks scale with `W`.

This primarily helps read-heavy batches with sparse read-modify-write updates, especially when they also contain a small number of blind keyed mutations. Batches without upserts retain their hash-free staged-update walk.

Measured with the `constantinople` harness on `any::unordered::fixed::mmb` (1M seeded keys, depth 0, 8 threads), comparing this branch against its base at `991ff08a`. Figures are the median of 27 timed iterations per point (3 runs of iterations 1-9, discarding each run's warmup iteration).

| config | merkleize before | merkleize after | delta |
| --- | --- | --- | --- |
| 1M reads / 4k updates | 1.46 ms | 1.12 ms | -23% |
| 262k reads / 8k updates | 1.71 ms | 1.58 ms | -7.6% |

Total per-batch latency is unchanged in both configurations, because this harness is dominated by the staged load phase (roughly 48 ms of a 50 ms batch at 1M reads). Median p50 totals were 49.9 ms before versus 50.3 ms after at 1M reads, and 11.65 ms versus 11.90 ms at 262k reads, i.e. noise in both directions. The change reduces merkleize-phase work and transient allocation (roughly 48 MB down to 8 MB of winner bookkeeping at 1M staged reads), not end-to-end batch latency.

Reproduce with:

```
cargo bench -p commonware-storage --bench constantinople --features test-traits -- \
  any::unordered::fixed::mmb 0 10 1000000 1000000 1 4096 8
```

## Testing

- `cargo fmt --all -- --check`
- `just test -p commonware-storage staged` (14 passed)
- `cargo clippy -p commonware-storage --lib --no-deps -- -D warnings -A clippy::chunks-exact-to-as-chunks`

### DIFF
--- a/storage/src/qmdb/any/batch.rs
+++ b/storage/src/qmdb/any/batch.rs
@@ -1561,6 +1561,7 @@ where
             }
         }
         drop(winner_of);
+        drop(slots);
 
         // Locations are unique after last-write-wins dedup (each key resolves to exactly one
         // location, committed or ancestor), so the parallel sort is deterministic. Sorting
--- a/storage/src/qmdb/any/batch.rs
+++ b/storage/src/qmdb/any/batch.rs
@@ -1504,26 +1504,32 @@ where
 
         // Resolve last-write-wins per distinct key without hashing each update: every staged slot
         // carries its distinct-key id, so a forward walk leaves each id's final write (the same
-        // winner as a newest-first scan). `touched` records each id on first write so the walks
-        // below stay proportional to the updates actually submitted, not the full staged read set.
+        // winner as a newest-first scan). `winner_of` is a directory over the staged read set
+        // holding one index per id, biased so zero means unwritten. `winners` stores the values,
+        // one per distinct key written, so it and the walks below scale with the writes.
         let StagedKeys { keys, slots, ids } = keys;
-        let mut winners: Vec<Option<(usize, Option<U::Value>)>> = vec![None; ids.len()];
-        let mut touched: Vec<usize> = Vec::with_capacity(updates.len());
+        let mut winner_of: Vec<usize> = vec![0; ids.len()];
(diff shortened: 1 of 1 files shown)


roberto-bayardo:
Superseded by #4419.

That PR removes the eager distinct-key index entirely rather than making it cheaper, so nothing here survives: both commits exist to reduce the cost of the `ids`/`slots`/`winner_of` structures that #4419 deletes.

The index was costing roughly 27 ms of a 50 ms staged load at 1M reads (a hash, a key clone, and a probe into an ~80 MiB table per staged slot) to save about 0.1 ms of merkleize. Removing it takes total batch latency from 50.58 ms to 24.16 ms in the read-heavy configuration, versus the unmeasurable end-to-end change here.

Leaving this open as a fallback until #4419 is reviewed.

PR #4419, opened 2026-08-07
[storage/qmdb] Drop the eager staged key index

roberto-bayardo:
## Summary

- drop the eager distinct-key index that staging built over every staged read, and resolve last-write-wins at merkleize by hashing the submitted updates instead
- `StagedKeys` (its `ids` map, its `slots` vector, and the winner directory sized by that id space) is gone; `Staged` now holds a plain `Vec<Key>`
- `expand` no longer has to assign ids consistent with earlier chunks, removing the one cross-chunk invariant in the staging path
- net 93 deletions against 38 insertions in a single file

Supersedes #4418, which made the eager index cheaper rather than removing it. Nothing from that PR survives here.

## Why

Staging assigned every staged slot a distinct-key id so that `resolve_updates` could deduplicate by direct indexing without hashing. Making the merkleize path hash-free cost a hash, a key clone, and a probe per staged slot on the load path. At one million staged reads the `ids` map alone is `AHashMap::with_capacity(1_000_000)`, i.e. 2^21 buckets of 40 bytes, roughly 80 MiB, probed a million times in hash order. That is about 27 ms of a 50 ms staged load, spent to save about 0.1 ms at merkleize.

## Performance

`constantinople`, `any::unordered::fixed::mmb`, 1M seeded keys, depth 0, 8 threads. Medians of 27 timed iterations per point (3 runs of iterations 1-9, discarding each run's warmup iteration). Baseline is `991ff08a`.

| config | phase | main | this branch | delta |
| --- | --- | --- | --- | --- |
| 1M reads / 4k updates | load | 48.89 ms | 22.92 ms | -53.1% |
| | merkleize | 1.69 ms | 1.19 ms | -29.6% |
| | **total** | **50.58 ms** | **24.16 ms** | **-52.2%** |
| 32k reads / 32k updates | total | 5.31 ms | 4.85 ms | -8.7% (within noise) |
| 262k reads / 262k updates | load | 10.06 ms | 5.07 ms | -49.6% |
| | merkleize | 32.85 ms | 35.15 ms | **+7.0%** |
| | **total** | **43.00 ms** | **40.32 ms** | **-6.2%** |

### The tradeoff, stated plainly

Hashing moves from the load phase onto the merkleize path, so **merkleize regresses at high update counts**: +7.0% at 262k updates, which is statistically resolved rather than noise. It is covered by a larger load-phase saving, so total improves in every configuration measured, but it is a real cost.

The two effects are cleanly separable and scale with different quantities:

- removed from load: about 19-26 ns per staged **slot** (hash, key clone, and insert into the large table)
- added to merkleize: about 8.8 ns per submitted **update** (+2.30 ms across 262k updates)

Break-even is therefore around `updates ~= 2.5 * staged slots`. Callers submit at most one update per staged slot in normal use, so this is a win across the realistic range; it would only lose if a batch rewrote the same staged keys several times over.

Reproduce with:

```
cargo bench -p commonware-storage --bench constantinople --features test-traits -- \
  any::unordered::fixed::mmb 0 10 1000000 1000000 1 4096 8
```

## Behavior change

One, called out because it is the only place the diff changes behavior rather than representation. `StagedKeys::distinct()` was the sole consumer of the id map outside deduplication: it clamped the floor-raise step bound to the distinct staged key count. That clamp is now `keys.len()`, the staged slot count, which is still a valid upper bound but a looser one. The bound is documented as approximate in both directions, and an over-estimate only prefetches surplus candidates, which the raise already drops once it has moved enough ops.

## Testing

- `just test -p commonware-storage qmdb::` (1639 passed)
- `cargo fmt --all -- --check` (via the pre-commit hook)
- Speculative root parity against `991ff08a`: 20 of 20 roots byte-identical across four configurations, covering `depth=2` (an uncommitted ancestor chain), `read_chunks=3` (multi-chunk `expand`, where the removed cross-chunk id consistency previously mattered), the read-heavy and update-heavy ratios, and `any::ordered::fixed::mmb` (which exercises the other `STAGES_DELETES` and `STAGES_ANCESTORS` paths).

### DIFF
--- a/storage/src/qmdb/any/batch.rs
+++ b/storage/src/qmdb/any/batch.rs
@@ -3882,7 +3882,7 @@ mod tests {
         2
     );
 
-    /// Build a [`Staged`] handle with the slot key-id map `stage`/`expand` would have built.
+    /// Build a [`Staged`] handle with the keys and resolutions `stage`/`expand` would produce.
     fn staged_with<F: Family, H: Hasher, U: update::Update + Send + Sync, S: Strategy>(
         batch: UnmerkleizedBatch<F, H, U, S>,
         keys: Vec<U::Key>,
--- a/storage/src/qmdb/any/batch.rs
+++ b/storage/src/qmdb/any/batch.rs
@@ -1453,10 +1453,12 @@ where
 
         // Resolve last-write-wins per distinct key: a forward walk keyed on the staged key leaves
         // each key's final write, the same winner as a newest-first scan. Later writes overwrite
-        // their key's entry in place, so `winners` holds one entry per distinct key written and
-        // every structure here is sized by the updates submitted, not by the staged read set.
-        let mut winner_of: AHashMap<&U::Key, usize> = AHashMap::with_capacity(updates.len());
-        let mut winners: Vec<Option<(usize, Option<U::Value>)>> = Vec::with_capacity(updates.len());
+        // their key's entry in place, so `winners` holds one entry per distinct key written. A
+        // distinct key needs a staged slot, so the staged read set caps the reservation that a
+        // duplicate-heavy update list would otherwise inflate.
+        let capacity = updates.len().min(keys.len());
+        let mut winner_of: AHashMap<&U::Key, usize> = AHashMap::with_capacity(capacity);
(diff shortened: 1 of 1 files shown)


patrick-ogrady (approved):
BIG win for smart contract-based chains I suspect.

PR #4423, opened 2026-08-07
Make `Strategy::spawn` size-aware (adaptive inline vs offload)

roberto-bayardo:
## Problem

The merkle strategy's parallelism scaled *negatively* with thread count on
per-block workloads. On an EVM-state replay benchmark (QMDB follower write path,
32-core EPYC, state fully cached), `merkleize` at 8 threads was ~3.3x **slower**
than at 1 thread, and adding threads made it monotonically worse.

The cause was not hashing or a buffer-pool lock. `Strategy::spawn` was the one
strategy op that ignored input size: it unconditionally handed a job to the pool
whenever parallelism > 1, so the merkleize path's tiny per-block jobs each paid a
pool hand-off (oneshot channel + worker wake + work-steal) that dwarfed the work.
`run`/`sort_by`/`map` already adapt serial-vs-parallel by size; `spawn` did not.

## Fix

**`spawn` becomes size-aware.** It takes a `len` and, on a strategy with an adaptive
policy, decides per call site (and per size class) whether to run the job inline on
the calling task or offload it to the pool.

The decision rule (after #4603, which replaced the original caller-residual signal):
a job runs inline iff its measured wall time is no greater than both the measured
**offload round-trip overhead** and a 1ms executor-block budget. The overhead is
everything an offloaded run costs around the job itself: hand-off setup, queueing,
worker wake, result send, task wake, and the observing poll, measured as the spawn
to result-observed elapsed time minus the job's own wall (recorded on the worker, so
job estimates are independent of how or whether the caller polls). A caller that
waits on the result therefore inlines whenever that is measured cheaper; a caller
that polls late inflates its overhead samples with overlap slack, which only ever
biases small jobs toward inline, bounded by the budget. Estimates are per
(call site, size bucket) EWMAs with a shared measurement cadence; entries seed by
offloading, and boundary probes keep both paths' estimates live. To force an
unconditional hand-off, submit through `manual()`; a single-worker pool always runs
jobs inline.

**Callers pass the job size.** The QMDB/journal merkleize path is the motivating
win: journal `merkleize`, `diff_sort`/`diff_merge`, `graft_hash`,
`resolve_updates_prefetched`, compact `merkleize_ops`, plus the floor-raise
classification (routed through `strategy.run(read_candidates.len(), ..)`). The
simplex batcher (vote-batch counts), p2p decode (message byte length), and the
runtime/test strategies pass their natural sizes too.

## Result

QMDB follower write path, 500k blocks, 32 GiB cache, measured during development
against the pre-PR spawn:

| threads | before (wall / writes-per-s) | after |
|---|---|---|
| 1 | 13.3s / 132.6k | 13.7s / 128.6k |
| 8 | 44.5s / 35.3k | 15.1s / 114.8k |

- **Negative scaling eliminated**: 8 threads is now ~1.1x the 1-thread wall (was 3.3x).
- **+225% write throughput at 8 threads**; merkleize 36.2s to 10.0s.
- Holds at 2M blocks and on the builder path.
- **Byte-identical output**: the on-disk DB is bit-for-bit identical across thread
  counts, so the change affects only dispatch, not computation.

Re-verified at the exact merge candidate (rebased head `662ac6ac9` vs current main
`b269a647`, same protocol, arms interleaved twice): main still reproduces the original
pathology (t8 wall 43.7s to 44.7s, 35.3k to 36.2k writes/s), while this branch runs t8
at 13.5s to 15.0s and 114k to 128k writes/s, a wash at t1, byte-identical databases. A
heavy window (50k blocks at height 5M+, ~340 writes per block) was also measured; see
the known gap below.

The benchmarks use an 8-thread strategy pool on both arms. Beyond gating individual
call sites, adaptive inlining reduces how often the pool is woken at all, which
matters because idle pool workers are not free on low-parallel-duty-cycle workloads
(quantified in #4432); the policy adapts by measurement, so its decisions recalibrate
under any future change to pool idle costs.

## Known performance gap

The merged policy cannot observe caller overlap, only round-trip overhead. Deferred-join
call sites whose jobs genuinely overlap the caller's other work (the qmdb
`diff_sort`/`diff_merge` sites at heavy block sizes) measure inflated overhead and get
pulled inline, moving that formerly hidden work onto the critical path. Measured against
the pre-#4603 caller-residual variant of this branch: light-window t8 merkleize +2% to
+5%; heavy-window t8 merkleize +13% to +15% and writes/s -9% to -10% (t1 a wash in both
windows, heavy t1 slightly favoring the merged policy). This is an accepted trade for
measurement that is robust to caller polling style (the caller-residual signal
mismeasured pipeline-shaped callers such as the p2p decoder); the gap is reclaimable
later via a per-site placement declaration for overlapping-by-design sites. Details and
tables in the #4603 review thread.

## Tests

`SpawnEntry` decision unit tests cover seeding via offload, convergence to inline for
sub-overhead jobs, offload for jobs above the overhead or budget, budget capping under
an inflated overhead estimate, measurement cadence, and probe pacing; `Rayon::spawn`
integration tests pin tiny-job inlining and over-budget jobs never leaving the pool.
`just test -p commonware-parallel` green; qmdb/journal/compact, consensus batcher, p2p
codec, and runtime strategy suites pass at the rebased head.

### DIFF
--- a/consensus/src/simplex/actors/batcher/verifier.rs
+++ b/consensus/src/simplex/actors/batcher/verifier.rs
@@ -19,7 +19,7 @@ use tracing::{Instrument as _, Span, info_span};
 
 /// Runs a CPU-bound job through [Strategy::spawn], entering `span` on the worker thread and
 /// instrumenting the awaited future so the offloaded work stays attributed to the caller's trace.
-async fn offload<P, F, T>(span: Span, strategy: &P, job: F) -> T
+async fn offload<P, F, T>(len: usize, span: Span, strategy: &P, job: F) -> T
 where
     P: Strategy,
     F: FnOnce(P) -> T + Send + 'static,
@@ -27,7 +27,7 @@ where
 {
     let worker_span = span.clone();
     strategy
-        .spawn(move |strategy| worker_span.in_scope(|| job(strategy)))
+        .spawn(len, move |strategy| worker_span.in_scope(|| job(strategy)))
         .instrument(span)
         .await
 }
@@ -320,7 +320,7 @@ impl<S: Scheme<D>, D: Digest> Verifier<S, D> {
                 view = self.round.view().traced()
             );
             let scheme = Arc::clone(&self.scheme);
-            let notarization = offload(span, strategy, move |strategy| {
+            let notarization = offload(notarizes.len(), span, strategy, move |strategy| {
                 Notarization::from_owned_notarizes(scheme.as_ref(), notarizes, &strategy)
                     .expect("verified notarize quorum must assemble")
             })
@@ -335,7 +335,7 @@ impl<S: Scheme<D>, D: Digest> Verifier<S, D> {
                 view = self.round.view().traced()
             );
             let scheme = Arc::clone(&self.scheme);
-            let nullification = offload(span, strategy, move |strategy| {
+            let nullification = offload(nullifies.len(), span, strategy, move |strategy| {
                 Nullification::from_owned_nullifies(scheme.as_ref(), nullifies, &strategy)
                     .expect("verified nullify quorum must assemble")
             })
@@ -350,7 +350,7 @@ impl<S: Scheme<D>, D: Digest> Verifier<S, D> {
                 view = self.round.view().traced()
             );
             let scheme = Arc::clone(&self.scheme);
-            let finalization = offload(span, strategy, move |strategy| {
+            let finalization = offload(finalizes.len(), span, strategy, move |strategy| {
                 Finalization::from_owned_finalizes(scheme.as_ref(), finalizes, &strategy)
                     .expect("verified finalize quorum must assemble")
             })
@@ -519,7 +519,7 @@ impl<S: Scheme<D>, D: Digest> Verifier<S, D> {
                 );
                 let scheme = Arc::clone(&self.scheme);
                 let mut rng = StdRng::from_rng(rng);
-                offload(span, strategy, move |strategy| {
+                offload(notarizes.len(), span, strategy, move |strategy| {
                     let (proposals, attestations): (Vec<_>, Vec<_>) = notarizes
                         .into_iter()
                         .map(|n| (n.proposal, n.attestation))
@@ -577,7 +577,7 @@ impl<S: Scheme<D>, D: Digest> Verifier<S, D> {
                 let round = nullifies[0].round;
                 let scheme = Arc::clone(&self.scheme);
                 let mut rng = StdRng::from_rng(rng);
-                offload(span, strategy, move |strategy| {
+                offload(nullifies.len(), span, strategy, move |strategy| {
                     let Verification { verified, invalid } = scheme.verify_attestations::<_, D, _>(
                         &mut rng,
                         Subject::Nullify { round },
@@ -633,7 +633,7 @@ impl<S: Scheme<D>, D: Digest> Verifier<S, D> {
                 );
                 let scheme = Arc::clone(&self.scheme);
                 let mut rng = StdRng::from_rng(rng);
-                offload(span, strategy, move |strategy| {
+                offload(finalizes.len(), span, strategy, move |strategy| {
                     let (proposals, attestations): (Vec<_>, Vec<_>) = finalizes
                         .into_iter()
                         .map(|n| (n.proposal, n.attestation))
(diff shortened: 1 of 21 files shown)


storage/src/qmdb/compact/batch.rs
```diff
@@ -32,8 +33,9 @@ where
     let first_leaf = batch.leaves();
     let mem = merkle.snapshot();
     let strategy = merkle.strategy().clone();
+    let ops_len = ops.len();
     strategy
-        .spawn(move |strategy| {
+        .spawn(ops_len, move |strategy| {
```
patrick-ogrady: this is so sick...multiple-layers of rayon/inline now ❗

parallel/src/policy.rs
```diff
@@ -259,6 +315,103 @@ impl Entry {
...
+    // The most recent inline measurement, kept raw (not blended into `inline_ns`) so the
+    // executor-block gate reacts to a single over-budget run instead of waiting for the EWMA.
+    last_inline_ns: Option<u64>,
+    since_probe: u32,
+}
+
+impl SpawnEntry {
+    // Returns the path to run and whether the caller should time it. Offload is seeded first: it is
+    // the only path that also measures the job's own wall time, which estimates the inline cost.
+    // Once both estimates exist, the cheaper caller-visible cost wins (ties -> inline, which frees a
```
patrick-ogrady: I think it will always be cheaper to inline? Is this the right heuristic (I was thinking something along the lines of spawn time vs total time)?

roberto-bayardo: The metric used to decide inlining is not "is inlining faster than offloading" in walltime.   The choice compares the two paths by their "caller-visible marginal cost", not by the job's own wall time:

inline: the job's wall time, entirely on the caller's critical path.
offload: the pool hand-off plus the residual wait once the caller finally joins.

roberto-bayardo: (the PR description describes this in more detail)

roberto-bayardo: I've tried to clarify the code comments around this.

parallel/src/lib.rs
```diff
@@ -1016,45 +1035,99 @@ commonware_macros::stability_scope!(BETA, cfg(any(feature = "std", test)) {
...
+                        job_wall,
+                    );
                 }
-                match rx.await {
-                    Ok(Ok(value)) => value,
-                    Ok(Err(payload)) => panic::resume_unwind(payload),
-                    Err(_) => panic!("strategy job dropped before completion"),
+                match result {
+                    Ok(value) => value,
+                    Err(payload) => panic::resume_unwind(payload),
```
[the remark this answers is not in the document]
roberto-bayardo: A panicking job's sample is only observable if something catches the panic and keeps using the same strategy instance: `resume_unwind` re-raises in the caller, and we treat panics from these jobs as fatal (the process, or at minimum the database instance, is done). Deterministic panics also just re-panic on the next run, so the "later unlock inline probes" state never arrives.

Even in the hypothetical where execution continues, a poisoned sample can trigger at most one inline probe: that probe records `last_inline_ns`, and the safety gate reacts to a single over-budget inline run by re-offloading immediately, so a bad sample can't stick. Leaving as-is.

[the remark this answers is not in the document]
roberto-bayardo: Fixed in 8c26fd4bf: the offload arm now records its policy sample in the success branch only, matching the inline arm (which never recorded panics since the panic propagates before the record). Added a test, spawn_panic_records_nothing, verified to fail against the previous ordering.

consensus/src/simplex/actors/batcher/verifier.rs
```diff
@@ -19,15 +19,15 @@ use tracing::{Instrument as _, Span, info_span};
...
+async fn offload<P, F, T>(len: usize, span: Span, strategy: &P, job: F) -> T
 where
     P: Strategy,
     F: FnOnce(P) -> T + Send + 'static,
     T: Send + 'static,
 {
     let worker_span = span.clone();
     strategy
-        .spawn(move |strategy| worker_span.in_scope(|| job(strategy)))
+        .spawn(len, move |strategy| worker_span.in_scope(|| job(strategy)))
```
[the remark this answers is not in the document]
roberto-bayardo: Fixed in 8c26fd4bf: offload is now a track_caller fn returning the instrumented future instead of an async fn, so each call site's location is forwarded into Strategy::spawn and the policy keys one entry per call site and size bucket. The attribute does not chain through an async fn body, hence the signature change; call sites are unchanged.

patrick-ogrady (approved):

patrick-ogrady (changes requested):
blocking merge to cleanup the .zip tweak I made one sec

patrick-ogrady (approved):

PR #4432, opened 2026-08-09
[storage] Apply the journal and index concurrently in Db::apply_batch

roberto-bayardo:
## Summary

`Db::apply_batch` previously ran its three phases serially: ops-journal append, in-memory merkle apply, then the index and bitmap application. The journal append and the index application touch disjoint state (the log vs the snapshot and bitmap), so this change runs the index and bitmap application as one job on the strategy while the journal append proceeds on the calling task, joining before the metadata update.

The size-aware `Strategy::spawn` (#4423) gates the offload adaptively, so batches too small to amortize the hand-off degrade to the previous sequential behavior. Because the policy measures caller-marginal cost, the overlap makes offloading look nearly free, which it now is.

The `I: 'static` bound ripples to the impl blocks whose methods move the index into the job. `Send` needs no mention: `Unordered` gained it as a supertrait in #4452.

The index application phase has been extracted into its own function with simplified control flow for improved comprehensibility.

## Benchmarks

All runs produce byte-identical databases (and bit-exact roots where the bench prints them). Linux, x86_64, AMD EPYC 9354P, mimalloc in the bench binary, paired same-session runs.

| workload | before | after | delta |
|---|---|---|---|
| eth_replay (100k blocks from a 5M-block seed, follower, 32 GiB cache) | 219.27s / 221.94s wall | 213.03s / 213.70s | **-3.3% wall, apply phase -20%, +4.2% writes/s** |
| qmdb generator (50M keyspace, 25M updates, 10k-update commits, p3 ordered) | 109.96s / 110.06s | 104.90s / 103.93s | **-5.1%** |
| apply_batch micro (16,384 updates, ordered p256) | 6.24ms | 4.86ms | **-22%** (non-overlapping CIs) |
| Constantinople seed+churn (1M-update seed + 4x32k churn applies) | 1.598s | 1.514s | **-5.3%** (Mac arm64: -5.0%; roots bit-exact across states and platforms) |

Constantinople's timed section (load + merkleize, which never applies) is unchanged, as expected.

## Pool size, and what the gains consist of

Everything above was measured with an 8-thread strategy pool on both arms, and the pool size itself turns out to be a significant variable. Idle pool workers are not free: rayon workers with nothing to do spin-yield around the serial thread, and on a workload with long serial stretches between parallel bursts that interference is expensive. Re-running this PR against its base at both 1-thread and 8-thread pools (same host, 20k-block eth-replay windows) shows the idle pool costs the base +78% on the apply phase and +14% on wall:

| | wall | apply |
|---|---|---|
| base, 8 threads | 50.80s | 8.03s |
| this PR, 8 threads | 48.81s | 6.02s |
| base, 1 thread | 44.45s | 4.52s |
| this PR, 1 thread | 43.86s | 4.38s |

Two things follow. First, the -25% apply / -3.9% wall reported above reproduces, but most of it is the overlap sheltering the critical path from that idle-pool interference -- and putting one otherwise-idle worker to work -- rather than raw concurrency: with a 1-thread pool the spawn inlines, the overlap is off, and the diff measures about -3% apply / -1.3% wall. Second, the benchmark tables in this description are accurate for the 8-thread environment they were measured in; they are not pool-independent.

The idle-pool cost tracks the workload's parallel duty cycle rather than being universal. On the same host at 1 vs. 8 threads (plain main): the merkleize microbenchmarks are a wash (within 1.5% either way), eth-replay runs 7% faster with the minimal pool, while the 50M-key generator runs 9% faster and the Constantinople shape 26% faster with the full pool. Pool size is therefore a per-workload deployment choice: sparse block-replay shapes run fastest with a minimal pool, dense batch workloads with a full one.

## Validation

-  qmdb tests, `just check-stability` clean.
- `db_bytes` identical across every A/B pair above; Constantinople speculative roots bit-exact per iteration.
- no impact on apply batch microbenchmarks.

### DIFF
--- a/storage/src/qmdb/any/batch.rs
+++ b/storage/src/qmdb/any/batch.rs
@@ -2863,7 +2863,9 @@ where
 
         // The journal append and the in-memory index application touch disjoint state (the log vs
         // the snapshot and bitmap), so run them concurrently. The index application is spawned as a
-        // job on the strategy while the journal append proceeds on this task.
+        // job on the strategy while the journal append proceeds on this task. A journal failure
+        // leaves the instance unusable either way (mutable storage failures are fatal and consume
+        // the database), so the index job's effects on error do not need to be rolled back.
         let strategy = self.strategy().clone();
         let log = self.log;
         let snapshot = self.snapshot;
--- a/storage/src/qmdb/any/batch.rs
+++ b/storage/src/qmdb/any/batch.rs
@@ -2879,7 +2879,6 @@ where
             // `extend_to`.
             bitmap.set_bit(*last_commit_loc, false);
             bitmap.set_bit(*batch.bounds.tip.size - 1, true);
-            drop(bitmap);
             snapshot
         });
 
--- a/storage/src/qmdb/any/batch.rs
+++ b/storage/src/qmdb/any/batch.rs
@@ -2802,7 +2802,14 @@ where
         let index_batch = Arc::clone(&batch);
         let index_bitmap = Arc::clone(&self.bitmap);
         let last_commit_loc = self.last_commit_loc;
-        let index_job = strategy.spawn(batch.diff.len(), move |_| {
+        // The job merges or scans every ancestor diff in addition to the tip diff, so
+        // size the spawn from all of them.
+        let work_hint = batch
+            .ancestor_diffs
+            .iter()
+            .map(|diff| diff.len())
+            .fold(batch.diff.len(), usize::saturating_add);
+        let index_job = strategy.spawn(work_hint, move |_| {
             let batch = index_batch;
             let mut snapshot = snapshot;
             let mut bitmap = index_bitmap.write();
--- a/storage/src/qmdb/any/batch.rs
+++ b/storage/src/qmdb/any/batch.rs
@@ -665,6 +665,75 @@ impl<'a, K: Ord, F: Family, V> Iterator for DiffMerge<'a, K, F, V> {
     }
 }
 
+/// Publish `batch`'s key-level changes to the in-memory snapshot index and activity bitmap.
+///
+/// `snapshot` and `bitmap` must hold the state committed through `last_commit_loc`, the location of
+/// the commit operation they currently end at. On return they reflect `batch.bounds.tip`.
+fn apply_batch_to_index<F, D, I, U, S, const N: usize>(
+    snapshot: &mut I,
+    bitmap: &mut bitmap::Prunable<N>,
+    batch: &MerkleizedBatch<F, D, U, S>,
+    last_commit_loc: Location<F>,
+) where
+    F: Family,
+    D: Digest,
+    I: UnorderedIndex<Value = Location<F>>,
+    U: update::Update,
+    S: Strategy,
+    Operation<F, U>: Send + Sync,
+{
+    let db_size = *last_commit_loc + 1;
+    let tip = *batch.bounds.tip.size;
+    bitmap.extend_to(tip);
+
+    if batch.ancestor_diffs.is_empty() {
+        // Fast path: no ancestors to merge, no fixups to look up.
+        for (key, entry) in batch.diff.iter() {
+            apply_diff(snapshot, bitmap, key, entry, entry.base_old_loc());
+        }
+    } else {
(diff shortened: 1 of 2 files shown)


storage/src/qmdb/any/batch.rs
```diff
@@ -2877,17 +2887,24 @@ where
...
             }
 
             // CommitFloor: bit = 1 only on the current last commit. Demote the previous and
             // set the new; earlier ancestor commits between them are already 0 from
             // `extend_to`.
-            bitmap.set_bit(*self.last_commit_loc, false);
+            bitmap.set_bit(*last_commit_loc, false);
             bitmap.set_bit(*batch.bounds.tip.size - 1, true);
-        }
+            drop(bitmap);
```
danlaine: I don't think this buys anything?
```suggestion
```

roberto-bayardo: Right, removed in 81da80c64 -- the guard releases at the closure's end either way.

storage/src/qmdb/any/batch.rs
```diff
@@ -2809,85 +2878,37 @@ where
...
+        // the snapshot and bitmap), so run them concurrently. The index application is spawned as a
+        // job on the strategy while the journal append proceeds on this task.
+        let strategy = self.strategy().clone();
+        let log = self.log;
+        let snapshot = self.snapshot;
+        let index_batch = Arc::clone(&batch);
+        let index_bitmap = Arc::clone(&self.bitmap);
+        let last_commit_loc = self.last_commit_loc;
+
+        // The job merges or scans the tip and its uncommitted ancestor diffs.
```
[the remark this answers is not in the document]
roberto-bayardo: Agreed -- committed/uncommitted collide with Db::commit persistence semantics. Fix ready locally (both locations reworded to applied/pending, matching the neighboring partition comments) and lands with the next push, together with the workspace clippy fixes blocking this PR's lint CI.

cryptography/src/lthash/mod.rs
```diff
@@ -139,8 +139,9 @@ impl LtHash {
     /// Return the [Digest] of the current state.
     pub fn checksum(&self) -> Digest {
         let mut bytes = [0u8; LTHASH_SIZE];
-        for (chunk, val) in bytes.chunks_exact_mut(2).zip(&self.state) {
-            chunk.copy_from_slice(&val.to_le_bytes());
+        let (chunks, _) = bytes.as_chunks_mut::<2>();
```
danlaine: Seems unrelated -- were these changes meant to be included?

roberto-bayardo: Good catch -- these were the workspace-wide fixes for the new stable clippy release, bundled here only because this PR's lint CI was red without them. Moved to a standalone PR off main (#4472) and dropped from this branch; this PR is storage-only again. Its lint CI stays red until #4472 lands on main and this rebases past it.

storage/src/qmdb/any/batch.rs
```diff
@@ -2791,85 +2861,37 @@ where
...
 
-            // CommitFloor: bit = 1 only on the current last commit. Demote the previous and
-            // set the new; earlier ancestor commits between them are already 0 from
-            // `extend_to`.
-            bitmap.set_bit(*self.last_commit_loc, false);
-            bitmap.set_bit(*batch.bounds.tip.size - 1, true);
-        }
+        // Apply journal (handles its own partial ancestor skipping) while the index job runs.
+        let (log, snapshot) = futures::join!(log.apply_batch(&batch.journal_batch), index_job);
+        self.log = log?;
```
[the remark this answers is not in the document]
roberto-bayardo: The behavior is intended and contractual: `apply_batch` consumes `self`, so on journal failure the instance is gone and the repo-wide rule that mutable storage failures are fatal applies -- no continued use, so no rollback needed (same error contract as the serial code before this PR). For live `current` batches sharing the bitmap: `qmdb/bitmap.rs` documents that reads through an invalidated `MerkleizedBatch` return inconsistent bytes and callers must drop invalid batches, and a failed apply invalidates every live chain. The pre-rewrite comment stated the no-rollback rationale explicitly and lost it in a comment rewrite; restored in 95229b07f.

danlaine (dismissed):

PR #4436, opened 2026-08-09
[consensus/simplex] Fix stale proposal verification and ancestry repair

patrick-ogrady:
Related: #3416

## Summary

Follow-up to #4426 for two production-reachable edge cases found during review:

- Bind in-flight application verification to the exact proposal and resolved parent payload, preventing a stale result for a displaced same-parent proposal from nullifying its certificate-backed replacement.
- Retain the highest full finalization separately from the movable resolver construction floor, keeping a settling proof servable for older targeted ancestry requests after a higher notarization becomes the floor.
- Add deterministic actor-loop regressions for both paths.

The resolver wire key and peer-visible encoding remain unchanged.

### DIFF
--- a/consensus/src/simplex/actors/resolver/state.rs
+++ b/consensus/src/simplex/actors/resolver/state.rs
@@ -39,13 +39,23 @@ pub(crate) enum Effect {
     RetainAbove(View),
 }
 
-/// Tracks all known certificates from the last
-/// certified notarization or finalized view to the current view.
+/// Tracks the construction floor, retained finalization, and certificates
+/// needed to repair views through the current view.
+///
+/// The construction floor and retained finalization have different lifetimes.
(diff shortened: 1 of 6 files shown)


consensus/src/simplex/actors/resolver/state.rs
```diff
@@ -146,10 +167,18 @@ impl<S: Scheme, D: Digest> State<S, D> {
         effects
     }
 
-    /// Get the best certificate for a given view (or the floor
-    /// if the view is below the floor).
+    /// Get the best certificate for a given view.
```
patrick-ogrady, line 170: This change could be ported to main first (not specific to these changes) but think it is worth keeping given we plan to merge this soon.

BrendanChou (approved):

PR #4443, opened 2026-08-10
[consensus/simplex] Simplify optimistic processing

patrick-ogrady:
Tighten the optimistic-validation follow-up before merge:

- clarify the existing outbound admission bounds, and extend issuance frontiers from the known notarized anchor without rescanning tracked rounds
- make `VoteTracker` the signer-uniqueness owner, removing redundant verifier deduplication while retaining lazy and compacted phase storage
- keep recovery passive by replaying journaled actions instead of synthesizing a finalize from a successful certification
- clarify lookahead, ancestry-resolution, and fuzz-invariant contracts


consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -1076,24 +1074,6 @@ impl<
...
         batcher.update(span, observed_view, leader, finalized, None);
 
-        // Re-attempt finalize votes for certifications that completed before
-        // shutdown: no later event re-triggers construction for an
-        // already-certified view (see [Self::construct]), so recovery would
-        // otherwise require abandoning the term through timeouts. This builds
-        // a fresh vote through `construct_finalize`'s usual gates rather than
-        // replaying an artifact, and is a no-op if the prior vote was
-        // journaled (replay restores its broadcast flag).
-        for view in replayed_successful_certifications {
```
patrick-ogrady, line 1086: I don't think we should be sending finalizes here? It seems better to just "let those old views die"?

I don't think we'll emit a bad vote because we seem to go through the voting gates (in case we did a nullify) but not sure this is a good idea...

BrendanChou (approved):

PR #4446, opened 2026-08-10
storage: add verify_proof_and_pinned_nodes_and_extract_digests

michaelkaplan13:
`verify_proof_and_pinned_nodes` already reconstructs each fold-prefix peak digest from the pinned nodes while validating a range proof, but it discards them and returns only a bool. This leaves callers unable to learn the digests of peaks that lie entirely before the proven range, which is exactly what an exoware QMDB writer needs to rebuild its peak state from a bounded operation range ending at the watermark instead of replaying the full operation log. 

This PR adds `verify_proof_and_pinned_nodes_and_extract_digests`, which performs the same verification and returns every authenticated node digest, including the reconstructed fold-prefix peaks, and reworks `verify_proof_and_pinned_nodes` into a thin wrapper over it.

### DIFF
--- a/storage/src/merkle/proof.rs
+++ b/storage/src/merkle/proof.rs
@@ -2575,11 +2575,7 @@ mod tests {
                 .collect();
 
             let proof = mem
-                .range_proof(
-                    &hasher,
-                    Location::<F>::new(start)..Location::<F>::new(n),
-                    0,
-                )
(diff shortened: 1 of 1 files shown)


michaelkaplan13:
Closing in favor of an exoware-only fix here: https://github.com/exowarexyz/monorepo/pull/114

PR #4448, opened 2026-08-10
[consensus/simplex] Dispatch proposal and verification before voter persistence

### LINKED ISSUE #4352: simplex/voter: dispatch next proposal before construct/sync/notify

BrendanChou:
Closes #4352.

Port of #4354 (closed) to the voter loop that landed with #3416.

The dispatch sits between construction and the journal sync in `on_end`. Constructing the notarize advances the optimistic frontier, which makes the next view eligible, so an earlier dispatch point never fires on the stable-leader pipeline. The dispatch covers both application requests. While the voter syncs its journal and broadcasts, the next leader builds its block and every other validator verifies a buffered optimistic child.

Safety follows from existing guards. `pending_propose` and `pending_verify` block a second request while one is in flight. State one-shots construction per view. The `on_start` staleness drop uses `view < current`, so an optimistic future-view request survives. The result is only consumed on a later iteration.

Trace-based regression tests pin the order: the view 2 propose and verify requests must precede the view 1 notarize broadcast. The prior version of this change (#4354) was silently reverted by a loop rewrite; the tests fail if the dispatch is dropped.

Measurements are in #4354: ~2.7ms median proposal slack (from ~1.6ms consumed before dispatch) and past-deadline proposals ~15-19% -> ~10% at `TARGET_BLOCK_INTERVAL_MS = 5`.

## Testing

- `just test -p commonware-consensus` (1,353 passed)
- `just clippy -p commonware-consensus`
- `just check-fmt`

### DIFF
--- a/consensus/src/simplex/actors/voter/actor.rs
+++ b/consensus/src/simplex/actors/voter/actor.rs
@@ -1262,10 +1262,10 @@ impl<
                     staged.nullify = nullify;
                     staged.certification = certification;
 
-                    // Constructing the notarize advances the optimistic frontier,
-                    // making child requests eligible. Start them before journal
-                    // sync and publication. Their responses join the next
-                    // iteration's waiters.
+                    // A constructed notarize advances the optimistic frontier and
+                    // can make child requests eligible. Start those requests before
+                    // journal sync and publication. The next iteration polls their
+                    // responses.
                     self.reconcile_application_requests(
                         &mut resolver,
                         &mut pending_propose,
--- a/consensus/src/simplex/actors/voter/actor.rs
+++ b/consensus/src/simplex/actors/voter/actor.rs
@@ -412,6 +412,40 @@ impl<
         Some(Request(context, span, receiver))
     }
 
+    /// Starts eligible proposal and verification requests.
+    async fn dispatch_application_requests(
+        &mut self,
+        resolver: &mut resolver::Mailbox<S, D>,
+        pending_propose: &mut Option<Request<Context<D, S::PublicKey>, D>>,
+        pending_verify: &mut Option<Request<Context<D, S::PublicKey>, bool>>,
+    ) {
+        // Keep requests for optimistic future views and clear requests for
+        // exited views. Certification for an exited view can continue after
+        // its verification receiver is dropped.
+        let current_view = self.state.current_view();
+        if pending_propose
+            .as_ref()
+            .is_some_and(|request| request.view() < current_view)
+        {
+            *pending_propose = None;
+        }
+        if pending_verify
+            .as_ref()
+            .is_some_and(|request| request.view() < current_view)
+        {
+            *pending_verify = None;
+        }
+
+        // State and Round prevent duplicate requests when both checkpoints
+        // observe the same view.
+        if pending_propose.is_none() {
+            *pending_propose = self.try_propose().await;
+        }
+        if pending_verify.is_none() {
+            *pending_verify = self.try_verify(resolver).await;
+        }
+    }
+
     /// Persists our nullify vote to the journal for crash recovery.
     async fn handle_nullify(self, nullify: Nullify<S>) -> Self {
         self.append_journal(nullify.view(), Artifact::Nullify(nullify))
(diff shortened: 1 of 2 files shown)


patrick-ogrady (dismissed):

patrick-ogrady (dismissed):

consensus/src/simplex/actors/voter/mod.rs
```diff
@@ -240,6 +248,38 @@ mod tests {
...
+            oracle,
+            participants,
+            schemes,
+            elector,
+            options,
+        )
+        .await
+    }
+
+    async fn setup_voter_with_context<E, S, L>(
```
patrick-ogrady: we should avoid this inner function I think (just add context to setup_voter?

consensus/src/simplex/actors/voter/mod.rs
```diff
@@ -2965,10 +3047,197 @@ mod tests {
...
+                        panic!("expected notarize for view 1 after journal sync");
+                    }
+                }
+            }
+        });
+    }
+
+    /// An exited view's pending proposal must not suppress replacement work.
+    #[test_collect_traces]
+    fn test_stale_propose_does_not_block_replacement_dispatch(traces: TraceStorage) {
```
patrick-ogrady, line 3128: could probably add some more block comments to these new tests

consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -1248,6 +1261,14 @@ impl<
                     staged.nullify = nullify;
                     staged.certification = certification;
 
+                    // Start optimistic child requests before journal sync and
+                    // publication. Their responses join the next iteration's waiters.
+                    self.dispatch_application_requests(
```
patrick-ogrady: I wonder if this can actually be called prior to notify (out of this async chunk)?

or do we require the constructed messages to be built to update inner state required for this to do anything?

consensus/src/simplex/actors/voter/mod.rs
BrendanChou: Done in c2343df. `setup_voter` now takes the voter context directly, and call sites pass `context.child("actor")`.

consensus/src/simplex/actors/voter/mod.rs
BrendanChou: Added phase comments in c2343df. The stale-propose test now explains why the replacement request targets view 3: a mid-term view requires its immediate parent, so a nullified view 1 rules out view 2. View 3 starts the next term and builds on genesis.

consensus/src/simplex/actors/voter/actor.rs
BrendanChou: This second dispatch cannot move above `construct()`. A successful `construct_notarize()` calls `prepare_optimistic_successor()`, which advances the optimistic frontier and makes child requests eligible. The `on_start` checkpoint runs before that state update, so it dispatches nothing on the pipelined path. The ordering tests fail without the post-construct checkpoint. This dispatch runs before journal sync and `notify`, allowing application work to overlap persistence and publication. Clarified the ordering constraint in 22c3ee4.

patrick-ogrady (approved):
🐎

PR #4449, opened 2026-08-10
[runtime] Fix buffer pool freelist scaling pathologies

roberto-bayardo:
## Problem

Configuring a buffer pool size class with a large `max_buffers` (for example, sizing the 4KB class to back a page cache) is currently pathological:

1. **Every miss sweeps the whole bitmap.** `Freelist::take` and `take_batch` scan all striped words before concluding the freelist is empty. An 8.4M-slot class has 131k cache-padded words (16MB per sweep). When a long-lived consumer (e.g. the paged cache, which parks one pooled buffer per resident page) holds every buffer, the freelist is permanently empty and every allocation pays the sweep. Measured on an EPYC 9354P: **160.5us per miss**; an eth-replay run with an 8.4M-buffer class wedged at ~50x wall (one core pegged, 99% of cycles in `TlsSizeClassCache::pop_global`).

2. **Derived thread caches scale without bound.** The automatic per-thread cache capacity is `limit / (2 * parallelism)`: 66k buffers per thread for an 8.4M class at parallelism 64. Refills batch-claim `capacity / 2` = 33k slots per miss (amplifying the sweeps), and each thread can strand 264MB of 4KB buffers in its local cache, defeating any sizing headroom.

## Fix

1. **Hierarchical availability summary**: one bit per group of 64 bitmap words, maintained only when the bitmap has at least 64 words. Puts publish the summary only when a word transitions from empty (unconditionally: guarding the publication with a load races a taker's clear-and-recheck and can hide a buffer behind a clear bit), so steady-state operations add no atomic traffic. Small bitmaps, including every pre-existing freelist benchmark case and loom geometry, build no summary and keep the current protocol exactly. Takes navigate set summary bits, entering at the thread's home group and rotating each group scan by the home word's stripe offset so per-thread striping is preserved; a drained group's bit is cleared (AcqRel) and rechecked with RMW reads, republishing if a racing put landed. An earlier revision of this PR used a global availability counter; it was rejected in review for serializing all stripes on one cache line (measured ~3.4x on the 8-thread batch-1 benchmark) and replaced with this design.

2. The derived thread-cache capacity caps at 256 (`MAX_DERIVED_TLS_CAPACITY`). Explicit `with_max_thread_cache_capacity` values are unaffected.

3. The tokio runtime's storage buffer pool derives parallelism from `worker_threads + max_blocking_threads`, since storage I/O runs on the blocking pool (review suggestion).

## Measurements (EPYC 9354P, release)

| case | main (305ead4a0) | this PR |
|---|---|---|
| empty take, 8.4M-slot class | 160.5us | 51ns |
| freelist 512 slots / 8t / batch=1 (no summary) | ~20ns | ~12ns (parity) |

Full matrix including the new 65536-slot summary-active cases: see [the benchmark comment](https://github.com/commonwarexyz/monorepo/pull/4449#issuecomment-5245925336). Since rebasing onto #4010, its probe-affine slot placement is part of the base, which brings the batch-churn cases (65536 slots, batch=8/32) to parity with main by collapsing per-batch summary transitions to one — the "+ #4010" column in that comment is now the in-tree behavior.

## Validation

- 762/762 `commonware-runtime` tests on the rebased tree (new unit tests: summary gating, group tracking, cross-group navigation, derived cap)
- freelist loom suite passes in ~14s, with the summary protocol covered by two dedicated models: the publication edge through summary navigation, and a taker's clear-and-recheck raced against a put's empty-word transition (asserting no parked buffer becomes unreachable)
- a 65536-slot benchmark case exercises the summary path in the freelist matrix

### DIFF
--- a/runtime/src/iobuf/pool.rs
+++ b/runtime/src/iobuf/pool.rs
@@ -83,6 +83,18 @@ cfg_if::cfg_if! {
 /// amortizing shared-queue traffic.
 const MIN_TLS_BATCH_CAPACITY: usize = 4;
 
+/// Upper bound for the derived per-thread cache capacity.
+///
+/// The derived capacity scales with the class limit so refill batches stay
+/// meaningful for bigger pools, but past a few hundred entries the extra
+/// amortization of shared-freelist traffic is negligible while the buffers a
+/// thread can strand in its local cache keep growing. Classes sized in the
+/// millions (for example to back a long-lived page cache) would otherwise
+/// derive per-thread caches of tens of thousands of buffers. An explicit
+/// [`BufferPoolConfig::with_max_thread_cache_capacity`] is not subject to
+/// this bound.
+const MAX_DERIVED_TLS_CAPACITY: usize = 256;
+
 /// Error returned when buffer pool allocation fails.
 #[derive(Error, Debug, Clone, Copy, PartialEq, Eq)]
 pub enum PoolError {
@@ -101,9 +113,10 @@ pub(crate) enum BufferPoolThreadCacheConfig {
     ///
     /// `None` derives the per-thread cache size from the pool's per-class
     /// capacity and expected parallelism, reserving about half of each class
-    /// for the shared freelist. Small per-class budgets may resolve to zero,
-    /// disabling thread-local caching so free buffers do not become stranded in
-    /// other threads.
+    /// for the shared freelist and capping the result at
+    /// [`MAX_DERIVED_TLS_CAPACITY`]. Small per-class budgets may resolve to
+    /// zero, disabling thread-local caching so free buffers do not become
+    /// stranded in other threads.
     ///
     /// `Some(n)` uses an explicit per-thread cache size, clamped independently
     /// to each size class's limit.
@@ -627,14 +640,15 @@ impl BufferPoolConfig {
     ///
     /// Derived capacities divide half of the class limit across the expected
     /// parallelism so cross-thread reuse remains effective. Small class limits
-    /// may resolve to zero. An explicit capacity replaces the derivation and
-    /// clamps to the class limit.
+    /// may resolve to zero, and large ones cap at
+    /// [`MAX_DERIVED_TLS_CAPACITY`]. An explicit capacity replaces the
+    /// derivation and clamps to the class limit.
     fn resolve_thread_cache_capacity(&self, class_limit: NonZeroU32) -> usize {
         let class_limit = class_limit.get() as usize;
         match self.thread_cache_config {
             BufferPoolThreadCacheConfig::Enabled(None) => {
                 let effective_threads = self.parallelism.get().min(class_limit);
-                class_limit / effective_threads.saturating_mul(2)
+                (class_limit / effective_threads.saturating_mul(2)).min(MAX_DERIVED_TLS_CAPACITY)
             }
             BufferPoolThreadCacheConfig::Enabled(Some(capacity)) => capacity.get().min(class_limit),
             BufferPoolThreadCacheConfig::Disabled => 0,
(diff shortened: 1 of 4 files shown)


runtime/src/iobuf/pool.rs
```diff
@@ -83,6 +83,18 @@ cfg_if::cfg_if! {
...
+///
+/// The derived capacity scales with the class limit so refill batches stay
+/// meaningful for bigger pools, but past a few hundred entries the extra
+/// amortization of shared-freelist traffic is negligible while the buffers a
+/// thread can strand in its local cache keep growing. Classes sized in the
+/// millions (for example to back a long-lived page cache) would otherwise
+/// derive per-thread caches of tens of thousands of buffers. An explicit
+/// [`BufferPoolConfig::with_max_thread_cache_capacity`] is not subject to
+/// this bound.
+const MAX_DERIVED_TLS_CAPACITY: usize = 256;
```
patrick-ogrady, line 96: I think it may be better to just set the TLS setting for storage to use the runtime blocking threads count in normal tokio? that currently makes it pretty much useless anyways.

roberto-bayardo: Agreed, implemented: `resolved_storage_buffer_pool_config` now derives parallelism from `worker_threads + max_blocking_threads`, since storage I/O runs on the blocking pool. That widens freelist striping for the threads that actually contend and shrinks the derived per-thread caches to match the real thread population (for the default 64-buffer classes the derivation resolves to 0, i.e. TLS off for storage, consistent with your read that it was useless as configured). It lands in the next push of this branch together with the response to the P1: the availability counter is replaced with a size-gated hierarchical summary bitmap so empty misses stay cheap without any per-operation atomics on the hot path. Will update the PR with benchmark results once the battery finishes.

roberto-bayardo:
Benchmark results for the revised design (EPYC 9354P, release; `iobuf::freelist` matrix at 8 threads/lockstep unless noted; `main` = 305ead4a0). The last column is this branch merged with #4010, using #4010's bench suite for all legs.

| case | main | this PR | this PR + #4010 |
|---|---|---|---|
| 512 slots / batch=1 (no summary; the case that regressed 3.4x with the counter) | ~20.3ns | **12.5ns** | 12.0ns |
| 65536 slots / batch=1 (summary active) | 11.0ns | 18.7ns | 18.7ns |
| 65536 / batch=8 | 454ns | 710ns | **468ns** |
| 65536 / batch=32 | 571ns | 821ns | **575ns** |
| 65536 / single thread / batch=1 | 9.3ns | 17.2ns | 17.2ns |
| `iobuf::reuse` (#4010's target case) | 16.7µs | 16.7µs | **4.37µs** |
| empty take, 8.4M-slot class | 160.5µs | **51ns** | — |

Reading:

- The counter's contention regression is gone by construction: every existing bench case is below the summary gate and runs the merge-base protocol (512/b1 at parity).
- The visible residual is the unconditional transition publish: **~8ns per empty-word transition**, which this benchmark maximizes (each worker's word oscillates empty/nonempty every cycle, and with sequential slot placement a batch scatters across N words = N transitions per batch, hence +45-56% at b8/b32 on the 65536 case).
- **#4010 collapses exactly that cost**: probe-affine creation places a worker's batch in its home word, so a batch put pays one transition instead of N — b8/b32 return to parity with main, and #4010's own 3.8x reuse win is fully preserved on top of this PR. The two PRs are strongly complementary (with the caveat noted separately that #4010's lazy-growth reservation scan needs a bound before it is safe for large classes).
- Loom: the freelist suite runs in ~13s with the summary protocol covered by two dedicated models (publication edge through summary navigation; clear-and-recheck raced against a transition put asserting no parked buffer becomes unreachable).

If the single-buffer oscillation residual (+8ns) matters, there is a further refinement: keep a guard load before the transition publish and upgrade the put-side word `fetch_or` to `AcqRel`, which closes the stale-read race via the release sequence through the word's RMW chain (the taker's summary clear becomes visible to the put's guard load whenever the recheck could have missed the transition). Holding that unless the number matters in practice, since the unconditional publish is simpler to reason about.

roberto-bayardo:
Rebased onto main now that #4010 has landed (history collapsed to two commits). The adaptation is the same tree that produced the combined benchmarks above: the far-group unit test is creation-order-agnostic under probe-affine placement, and the batch-churn cases that #4010 repairs are now the in-tree behavior. Full validation on the rebased tree: 762/762 runtime tests, freelist loom 22/22 in ~14s, loom-feature and no-default-features builds clean under -D warnings.

PR #4454, opened 2026-08-11
[storage] Shard partitioned-index batch apply across the strategy

roberto-bayardo:
Stacked on #4432.

`Db::apply_batch` applies key-sorted diffs through a new `Index::apply_sorted_diffs`, which the partitioned ordered index overrides to shard at partition-snapped boundaries across the strategy: `ShardParts` exposes a borrowed mutation view over disjoint partition ranges, per-shard spilled maps are extracted and merged after, and bitmap flips are location-keyed (not key-range disjoint) so they stay in one serial pass before the sharded apply. Non-partitioned indexes keep the provided sequential default; batches under `MIN_DIFFS_PER_SHARD` per shard stay on the sequential path.

The fan-out reports its true workload to the adaptive policy via `map_collect_vec_with_multiplier`: presenting only shard-count items reads as trivial work and converges the policy to serial, silently disabling the sharding (this masked the win in earlier measurements of this change).

Because the guard release before the sharded apply is load-bearing, `apply_batch_to_index` now takes `&Shared<N>` and acquires the bitmap write lock itself instead of receiving a guard from its caller.

## Measurements

Eth replay, re-measured on the current stack. EPYC 9354P (32 core, x86_64), paired A/B, same seed DB restored per run, same trace, `db_bytes` identical across all runs.

| arm | wall (A/B) | apply | merkleize | writes/s |
|---|---|---|---|---|
| base (#4432 tip, without this PR) | 216.9 / 206.1s | 29.5 / 27.3s | 128.4 / 123.5s | 194.5k / 203.5k |
| sharded | 198.1 / 196.5s | 21.3 / 21.5s | 121.2 / 119.8s | 215.2k / 216.8k |

Means: base wall 211.5s / apply 28.39s; sharded 197.3s / 21.42s. The sharding is worth **-24.6% on the apply phase** and **-6.7% wall**.

Generator (50M keys, 25M updates, 10k/commit) wall, from the earlier session on the same host: 106.4 / 105.1s base vs **99.6 / 98.7s (-6.3%)**.

Known non-goal, and it is machine-dependent: on an index far below the partitioned structure's intended scale (65k keys, fully cache-resident), shard count scales with parallelism, so the fan-out either pays or does not depending on core count. `qmdb::apply_batch/case=direct variant=any::ordered::p256`, paired A/B:

| host | base | sharded |
|---|---|---|
| EPYC 9354P, 32 core | 5057us | **4937us (-2.4%)** |
| 10 core arm64 laptop | 672us | 781us (+16%) |

Absolute times differ ~6x between the two hosts for the same 29k-entry diff, so only within-host ratios mean anything here. Either way this configuration is not a real deployment of this index type.

The four uncommitted-ancestor cases and both `immutable::fixed` cases measure within noise of the base, as expected: they resolve `base_old_loc` fixups through the merged general path, which never calls `apply_sorted_diffs`.

## Pool size

The measurements above use an 8-thread strategy pool on both arms. The sharded apply is pool fan-out, so its value tracks the workload's parallel duty cycle (see #4432 for the general analysis). Against its base at both pool sizes (20k-block eth-replay windows, paired A/B, current main beneath): at 8 threads the apply phase drops 20.9% (wall -4.7%); with a 1-thread pool the fan-out is inert and the shard-boundary preparation costs about 4.5% of the apply phase (~2% wall). Sparse block-replay deployments run fastest with a minimal pool, where this PR is that small cost; the dense workloads it targets run with a full pool, where it pays.

## Validation

- new equivalence test: sharded apply produces identical index state to the sequential path (exercised at a forced small threshold)
- 190 index + 1645 qmdb tests pass

## Considered and dropped

Passing the caller's own key-sorted entries through a `SortedDiff` trait, so the sharded path reads them in place instead of projecting into a `(key, old, new)` array first. It removes a per-batch allocation (48 bytes per entry, ~1.4 MB at `updates=16384`), and on a 10 core laptop it measured -10% on the sharded microbenchmark and neutral on the sequential one. Re-measured on the EPYC box, where paired replicates agree to 0.2%, it was **-3.1% sharded but +6.9% sequential**, and neutral on eth replay (apply -1.6%, inside run-to-run spread). Not worth the trait for that, so the seam keeps taking a slice of triples and the sequential walk keeps its compact projection.

## Follow-ups

- partitioned-unordered parity via the same `apply_sorted_diffs` seam
- extend the `apply_batch` benchmark with smaller batch sizes and a partitioned-unordered case to map the crossover

### DIFF
--- a/storage/src/index/mod.rs
+++ b/storage/src/index/mod.rs
@@ -146,6 +146,10 @@ pub trait Cursor: Send + Sync {
 
 /// A trait defining the operations provided by a memory-efficient index that maps translated keys
 /// to arbitrary values, with no ordering assumed over the key space.
+/// One entry in a key-sorted diff batch: the translated key bytes with the value to
+/// replace (`None` inserts) and the value to store (`None` deletes).
+pub type KeyValueDiff<'a, V> = (&'a [u8], Option<V>, Option<V>);
+
 pub trait Unordered: Send + Sync {
     /// The type of values the index stores.
     type Value: Send + Sync;
@@ -206,7 +210,7 @@ pub trait Unordered: Send + Sync {
     fn apply_sorted_diffs<S: Strategy>(
         &mut self,
         _strategy: &S,
-        diffs: &[(&[u8], Option<Self::Value>, Option<Self::Value>)],
+        diffs: &[KeyValueDiff<'_, Self::Value>],
     ) where
         Self::Value: PartialEq + Copy,
     {
--- a/storage/src/index/mod.rs
+++ b/storage/src/index/mod.rs
@@ -14,7 +14,40 @@
 //! degrade substantially (each conflicting key may contain the desired value).
 
 use crate::translator::Translator;
+use commonware_parallel::Strategy;
 use commonware_runtime::Metrics;
+
+/// Apply one value diff for `key` against `index`: insert a value, replace a known
+/// value, or delete a known value (see [Unordered::apply_sorted_diffs]).
+pub(crate) fn apply_one_diff<I: Unordered + ?Sized>(
+    index: &mut I,
+    key: &[u8],
+    old: Option<I::Value>,
+    new: Option<I::Value>,
+) where
+    I::Value: PartialEq + Copy,
+{
+    match (old, new) {
+        (None, Some(new)) => index.insert(key, new),
+        (Some(old), Some(new)) => {
+            let mut cursor = index.get_mut(key).expect("key should be known to exist");
+            assert!(
+                cursor.find(|v| *v == old),
+                "known key with given old value should have been found"
+            );
+            cursor.update(new);
+        }
+        (Some(old), None) => {
+            let mut cursor = index.get_mut(key).expect("key should be known to exist");
+            assert!(
+                cursor.find(|v| *v == old),
+                "known key with given old value should have been found"
+            );
+            cursor.delete();
+        }
+        (None, None) => {}
+    }
+}
 
 mod storage;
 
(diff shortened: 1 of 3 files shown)


roberto-bayardo:
Review findings addressed in f32c4fa09 — all three were correct:

- **[P1] Spill extraction** now walks each shard's key-sorted diff slice and extracts only the distinct partitions it names (consecutive dedup, so O(batch)). The previous range-probe moved every spilled map falling between a shard's first and last touched partition — on a large sparse-batch index that approached all spilled maps per batch, outside the adaptive policy's timer.
- **[P2] Boundary construction** starts each split scan at max(target, previous bound), so a partition run spanning several targets is walked once in total instead of once per target.
- **[P2] Multiplier regression coverage**: added a sequential-executing `RecordingStrategy` (parallelism 4 via `manual()`) that records `(shards, multiplier)` from the multiplier-aware map; the new test asserts exactly one call with `multiplier == diffs.len() / shards`. A regression to plain `map_collect_vec` now fails a test instead of only re-exposing the policy to shard-count-as-work convergence after it converges.

188 index + 1645 qmdb tests pass (equivalence test still green). The clippy failure noted on `utils` is pre-existing on the base and unrelated.

roberto-bayardo:
Terminal-run gap fixed in d8ad9626b: the boundary loop now breaks when a scan reaches the end of `diffs` (no later bound is possible there, and every remaining target would otherwise rescan the same run), and when snapping collapses the batch to a single shard the apply falls back to the sequential path instead of paying the fan-out machinery. Added a test: an all-one-partition batch produces zero fan-out calls (via the recording strategy) and correct results. 189 index tests pass.

roberto-bayardo:
Ran a full-diff adversarial review pass (six independent dimensions -- aliasing/thread soundness, boundary arithmetic, spill lifecycle, qmdb wiring, strategy/policy interaction, API/coverage -- each finding then attacked by two independent verifiers).

Result: the five correctness dimensions produced **no findings that survived verification** -- the carving, boundary, spill, and wiring logic as of d8ad9626b holds up under adversarial reading. One finding confirmed, now fixed:

- **[P2, confirmed] The sharded path was only ever tested at P=1.** The production configuration (P=2: sparse slices of a 65,536-entry partition array, keys of exactly P bytes yielding empty sub-keys, spills under a parallel strategy) never executed in any test -- the qmdb P=2 aliases all use `Sequential`, whose parallelism-1 takes the early sequential return. Fixed in 2e15734bf: the equivalence test is mirrored at P=2 with exactly those shapes under Rayon(4). 190 index tests pass.

- **[P2, contested -- flagging as follow-up, not part of this PR]** No fuzz target instantiates the partitioned ordered index at all (pre-existing; `storage/fuzz` covers only the flat indexes). The new sharded surface is invariant-heavy and would benefit from a partitioned-ordered fuzz target with an equivalence oracle against the sequential path. Happy to do that as a separate PR.

PR #4455, opened 2026-08-11
[runtime/iobuf] split iobuf

andresilva:


patrick-ogrady (approved):
while a large diff, confirmed this is overwhelmingly just moving things around.

PR #4459, opened 2026-08-11
[glue/dkg] Serialize Reshare Epoch Artifact Verification

patrick-ogrady:
## Summary

- Serializes reshare epoch-artifact verification around one materialized request while keeping competing typed ancestries lazy and FIFO.
- Reuses exact verification and artifact results at finalization, completes admitted speculative work first, and answers queued canonical requests from finalized state.
- Makes state-sync startup floor-aware: floors through the midpoint participate with complete inclusion history, Late floors follow, and startup never precedes the certified floor.
- Simplifies phase-local ownership and updates the public ancestry and state-sync documentation.

## Invariants

Finalized dealer logs in `Store` are authoritative and are persisted before reporter acknowledgement. Pending ancestry contributes only a temporary view for proposal or verification. The boundary parent remains derived from the typed ancestry head; a materialized request must supply a contiguous chain through every unfinalized inclusion block.

Only one verification task is active. Phase-local speculative results are reconstructable after restart and cannot displace a canonical result. Wire and storage formats are unchanged.

## Validation

- `just test --profile slow -p commonware-glue reshare_e2e_state_sync_active_player_late_restart`
- `just test --profile slow -p commonware-glue reshare_e2e_late_state_sync_carries_share_across_failure`
- `just check-docs -p commonware-glue`
- `just check-stability -p commonware-glue`
- `just lint`

Stacked on #4364.

### DIFF
--- a/glue/src/dkg/reshare/actor/inclusion.rs
+++ b/glue/src/dkg/reshare/actor/inclusion.rs
@@ -10,7 +10,7 @@ use crate::dkg::{
     types::{EpochInfo, EpochOutcome, Participants, Payload},
 };
 use commonware_consensus::{
-    marshal::core::Variant as MarshalVariant,
+    marshal::{ancestry::Ancestry as _, core::Variant as MarshalVariant},
     simplex::scheme::Scheme as SimplexScheme,
     types::{Epoch, EpochPhase, Epocher, FixedEpocher, Height},
 };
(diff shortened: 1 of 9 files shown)


glue/src/dkg/reshare/application.rs
```diff
@@ -228,14 +227,13 @@ where
                         return false;
                     }
                 }
-                EpochInfoResponse::Pending => {
-                    debug!("verification pending: final block epoch info is not ready");
+                EpochInfoResponse::Pending | EpochInfoResponse::Following => {
+                    debug!(
+                        "verification pending: final block epoch info cannot be derived locally"
+                    );
```
clabby, line 234: Not heavily opinionated here, though collapsing `Following` into the same debug message gets rid of the information. Maybe we add a `following = matches!(ev, EpochInfoResponse::Following)` field to this log at least?

glue/src/dkg/reshare/actor/mod.rs
```diff
@@ -418,28 +421,35 @@ where
...
 
         if matches!(self.mode, Mode::Dkg { .. }) {
             self.run_dkg(&mut store, &mut dealing_mux).await;
             return;
         }
 
-        let mut current_epoch = state_sync.as_ref().map(|state_sync| state_sync.info.epoch);
-        let mut state_sync_info = state_sync.map(|state_sync| state_sync.info);
+        let mut current_epoch = state_sync.as_ref().map(|start| start.info.epoch);
+        let mut state_sync = state_sync;
```
clabby: My brain’s compiler might not be working correctly, but this looks unnecessary. Just make `state_sync` mut on L424 and rm?

glue/src/dkg/reshare/actor/inclusion.rs
```diff
@@ -386,35 +423,87 @@ where
...
-        if !matches!(bounds.phase(), EpochPhase::Midpoint | EpochPhase::Late) {
-            continue;
+    if first_pending <= expected_height {
+        let mut allow_final = true;
+        loop {
+            let block = select! {
+                _ = &mut shutdown => return None,
+                _ = response.closed() => return None,
+                block = ancestry.next() => block,
+            };
```
clabby, line 460: This should probably be a `select_loop!`, since we’re polling shutdown.

glue/src/dkg/reshare/actor/inclusion.rs
```diff
@@ -35,88 +38,106 @@ use commonware_utils::{
...
-        F: FnOnce() -> PendingLogs<V, C::PublicKey>,
-    {
-        self.retain_open();
+        parent: &B::Digest,
+    ) -> Vec<oneshot::Sender<EpochInfoResponse<V, C, B::Directory>>> {
         self.inner
-            .first()
-            .map_or_else(fallback, |waiter| waiter.logs.clone())
+            .drain(..)
+            .filter_map(|request| {
```
clabby: `drain(..).filter_map(...)` silently drops every live noncanonical response sender. A losing request therefore observes a channel error instead of the normal typed `EpochInfoResponse::Unavailable`; `Mailbox::epoch_info` only recovers by logging a receive failure and translating the transport error. Send `Unavailable` to each live loser while collecting canonical responders, reserving sender destruction for actor failure.

The supplied regression test currently fails because `response_rx.now_or_never()` contains a channel error rather than `Ok(EpochInfoResponse::Unavailable)`.

```patch
diff --git a/glue/src/dkg/reshare/actor/inclusion.rs b/glue/src/dkg/reshare/actor/inclusion.rs
index 602fc2afe..8b074cb25 100644
--- a/glue/src/dkg/reshare/actor/inclusion.rs
+++ b/glue/src/dkg/reshare/actor/inclusion.rs
@@ -2029,4 +2029,22 @@ mod tests {
 
         assert_eq!(requests.pop().map(|request| request.parent), Some(earlier));
     }
+
+    #[test]
+    fn draining_canonical_responds_to_noncanonical_request() {
+        let genesis = mocks::genesis_block(signers()[0].public_key());
+        let canonical = genesis.digest();
+        let noncanonical =
+            TestBlock::new::<Sha256>(genesis.context().clone(), canonical, Height::new(1), 1)
+                .digest();
+        let (response_tx, response_rx) = oneshot::channel::<TestResponse>();
+        let mut requests = TestRequests::default();
+        requests.push(noncanonical, Span::none(), empty_ancestry(), response_tx);
+
+        assert!(requests.drain_canonical(&canonical).is_empty());
+        assert!(matches!(
+            response_rx.now_or_never(),
+            Some(Ok(EpochInfoResponse::Unavailable))
+        ));
+    }
 }
```
</details>

glue/src/dkg/reshare/actor/inclusion.rs
clabby, line 1: `SignedDealerLog::check` verifies only the self-signature. The pending path therefore admits a valid outsider into pending maps, exact cache keys, and verification inputs. The durable finalized path has the same authorization gap in one-shot DKG because `Store::current()` is absent, so its conditional membership guard does not reject the outsider before journaling it. Carry the authoritative round dealer set into both paths and reject signers outside that set before applying the first-log rule.

The supplied regression cases both currently fail: the outsider appears in `PendingLogs`, and `Store::has_log` reports that it was persisted.

```patch
diff --git a/glue/src/dkg/reshare/actor/inclusion.rs b/glue/src/dkg/reshare/actor/inclusion.rs
index 602fc2afe..59de87907 100644
--- a/glue/src/dkg/reshare/actor/inclusion.rs
+++ b/glue/src/dkg/reshare/actor/inclusion.rs
@@ -1365,15 +1365,19 @@ mod tests {
         tests::mocks::{self, MemorySecretStore, TestBlock, TestBlsVariant},
     };
     use commonware_actor::Feedback;
+    use commonware_codec::{Decode as _, Encode as _, Write as _};
     use commonware_consensus::{Reporter, marshal};
         Digestible as _, Signer,
         bls12381::{
-            dkg::feldman_desmedt::{Dealer as CryptoDealer, Verdict},
+            dkg::feldman_desmedt::{
+                Dealer as CryptoDealer, Output, SignedDealerLog, Verdict,
+            },
             primitives::sharing::Mode as SharingMode,
         },
         ed25519::{PrivateKey, PublicKey},
         sha256::Sha256,
+        transcript::{Transcript, Version},
     };
     use commonware_p2p::simulated::{Config as NetworkConfig, Network};
     use commonware_parallel::Sequential;
@@ -1470,6 +1474,31 @@ mod tests {
         BTreeMap::from([(public_key, log)])
     }
 
+    fn signed_nondealer_log() -> SignedDealerLog<TestBlsVariant, PrivateKey> {
+        let log = dealer_logs(0).into_values().next().expect("dealer log");
+        let participants = players();
+        let summary = Transcript::new(
+            b"_COMMONWARE_CRYPTOGRAPHY_BLS12381_DKG",
+            Version::V0,
+        )
+        .commit(TEST_NAMESPACE)
+        .commit(0u64.encode())
+        .commit(Option::<Output<TestBlsVariant, PublicKey>>::None.encode())
+        .commit(participants.encode())
+        .commit(participants.encode())
+        .summarize();
+        let mut transcript = Transcript::resume(summary, Version::V0).fork(b"log");
+        transcript.commit(log.encode());
+
+        let signer = PrivateKey::from_seed(99);
+        let signature = transcript.sign(&signer);
+        let mut encoded = Vec::new();
+        signer.public_key().write(&mut encoded);
+        log.write(&mut encoded);
+        signature.write(&mut encoded);
+        SignedDealerLog::decode_cfg(encoded.as_slice(), &NZU32!(4)).expect("signed dealer log")
+    }
+
     #[test]
     fn dropping_verification_aborts_active_task() {
         let executor = deterministic::Runner::timed(Duration::from_secs(10));
@@ -1855,6 +1884,80 @@ mod tests {
         });
     }
 
+    #[test]
+    fn pending_logs_ignore_valid_signature_from_nondealer() {
+            let info = info();
+            let genesis = mocks::genesis_block(signers()[0].public_key());
+            let block_one = TestBlock::new::<Sha256>(
+                genesis.context().clone(),
+                genesis.digest(),
+                Height::new(1),
+                1,
+            );
+            let block_two = TestBlock::new::<Sha256>(
+                genesis.context().clone(),
+                block_one.digest(),
+                Height::new(2),
+                2,
+            )
+            .with_payload::<Sha256, _, _>(
+                NZU32!(4),
+                Payload::DealerLog(signed_nondealer_log()),
+            );
+            let parent = block_two.digest();
+            let ancestry = Box::pin(stream::iter([Arc::new(block_two)]));
+            let (mut response_tx, _response_rx) = oneshot::channel::<TestResponse>();
+            let scan = PendingLogScan {
+                epoch: Epoch::zero(),
+                info: &info,
+                epocher: FixedEpocher::new(NZU64!(4)),
+                finalized_tip: Some(FinalizedTip {
+                    height: Height::new(1),
+                    digest: Some(block_one.digest()),
+                }),
+                final_height: Height::new(3),
+            };
+
+            let logs = pending_logs(scan, parent, ancestry, context.stopped(), &mut response_tx)
+                .await
+                .expect("anchored ancestry");
+            assert!(logs.is_empty(), "nondealer log must not enter pending view");
+        });
+    }
+
+    #[test]
+    fn finalized_nondealer_log_is_not_persisted_without_current_epoch() {
+            let epoch = Epoch::zero();
+            let outsider = PrivateKey::from_seed(99).public_key();
+            let mut store = Store::init(
+                context.child("nondealer_store"),
+                "nondealer-store",
+                NZU32!(4),
+                MemorySecretStore::default(),
+            )
+            .await;
+
+            mocks::TestReshareActor::observe_dealer_log(
+                &signers()[0].public_key(),
+                &info(),
+                &mut store,
+                epoch,
+                None,
+                Some(Payload::DealerLog(signed_nondealer_log())),
+            )
+            .await;
+
+            assert!(
+                !store.has_log(epoch, &outsider),
+                "nondealer log must not enter durable DKG view"
+            );
+        });
+    }
+
     #[test]
     fn pending_logs_stop_at_the_first_unfinalized_block() {
```
</details>

glue/src/dkg/reshare/application.rs
patrick-ogrady: Kept this unchanged. `Following` and `Pending` now intentionally share the same abstention path, and the final reduction pass found that a `following` field would be additive trace-only churn with no behavioral or contract oracle. I removed the attempted field and kept the common message.

glue/src/dkg/reshare/actor/mod.rs
patrick-ogrady: Addressed in the current local PR checkout: the post-resolution `state_sync` binding is now mutable and the redundant same-name rebinding is removed. The full `commonware-glue` tests, clippy, formatting, docs, and stability checks pass.

glue/src/dkg/reshare/actor/inclusion.rs
patrick-ogrady: Kept the current `loop` + `select!`. This shutdown-polling scaffold is unchanged from merge base `e0aa3d5`; the private helper deliberately accepts the exact `signal::Signal` it needs, while `select_loop!` would require broadening it to a runtime `Spawner`/context and would change when an unpolled future acquires its shutdown guard. The focused `pending_logs_` cases and the full crate suite pass, so I did not add that broader mechanism.

glue/src/dkg/reshare/actor/inclusion.rs
patrick-ogrady: Confirmed and addressed in the current local PR checkout. I added the supplied focused regression, observed it fail with a closed-channel error on the old implementation, then changed the phase-end drain to send `EpochInfoResponse::Unavailable` to every live noncanonical request while collecting canonical responders. The regression and all 332 default-profile `commonware-glue` tests pass.

glue/src/dkg/reshare/actor/inclusion.rs
patrick-ogrady: Confirmed, but not changed in this PR. The same admission behavior is present at merge base `e0aa3d5`: base `pending_logs` calls `log.check` then inserts the signer, and base `observe_dealer_log` has the same `Store::current().is_some_and(...)` guard; the cryptography, Store, and one-shot DKG owners are unchanged by this candidate. `Info::check_dealer_log` still rejects the outsider before ceremony selection, so it cannot affect quorum output or the persisted share; the remaining issue is temporary work/cache pollution and bounded one-shot journal pollution. The authoritative owner is `Info.dealers`, so a complete fix belongs in the `Info` / `SignedDealerLog::check` contract or a separate cross-crate ingress change. I kept that pre-existing issue out of this candidate-scoped autofix.

glue/src/dkg/reshare/application.rs
patrick-ogrady: (overriding tact)

glue/src/dkg/reshare/actor/inclusion.rs
patrick-ogrady: (overriding tact)

patrick-ogrady: We made the decision to allows this originally to defer as much as possible until the ceremony check. I thin we can re-evaluate but don't want to do that here.

clabby (approved):
Nice 😄

PR #4460, opened 2026-08-11
[storage/archive, marshal] Serve finalized backfill off the marshal actor loop

danlaine:
# Serve finalized backfill off the marshal actor loop

Marshal's core is a single task running a `select!` loop. One of its arms answers a peer asking for a finalized block, and it answers by reading two archives from disk inline. Consensus processing waits on those reads. A peer that has been offline for a while streams requests for old heights, which turns into sustained random disk I/O sitting directly in front of the loop that is supposed to be running consensus.

This PR moves that work to its own task. It does so by giving `archive::prunable::Archive` a read handle that another task can hold and use while the writer keeps writing.

## Why this needed a storage change

The archive's reads borrow (`get(&self)`) and its writes consume (`put(self) -> Self`). No borrow can survive a write, so a spawned task cannot hold a `&Archive`. It needs something owned and independent.

Reading also needed the writer's in-memory map. Index-journal records are appended in arrival order, so height H's record position depends on when H showed up, not on the number H, and there is no arithmetic that recovers it. A second reader therefore needed a map that is large and changes on every put.

Two changes make that map shareable:

**The map now records where the value is, not where the record is.** It was `index -> position in the index journal`. It is now `index -> Location { value_offset, position, value_size }`. A read by index goes straight to the value blob and never touches the index journal, so it costs one disk read instead of two. `position` is kept only for lookups by key, which still read the record to compare the stored key and rule out a translator collision.

**The map lives behind a lock, alongside the file handles a reader needs.** `Arc<RwLock<State>>` holds the location map and, per section, a `Capture`: a cloned blob handle plus the byte length the writer had flushed when it was taken. A read takes the lock long enough to copy one `Location` and clone one `Capture`, drops it, and then does its I/O with nothing held. No guard is ever held across an `.await`, and no I/O happens under one, so a read never waits on the writer's disk and the writer never waits on a reader's.

`Capture`'s length is what makes a live map safe to pair with lagging file handles. Value frames are appended and never rewritten, so a frame that fits inside the captured length reads the same bytes for as long as the capture lives, and anything written afterwards sits beyond it and is simply invisible.

## What a reader sees, and when

A put is invisible to readers until the store publishes it. Publishing happens inside `sync` and `start_sync`, both of which flush before returning, and it re-captures exactly the sections that call covered.

Publishing follows the flush, not the fsync. A reader can therefore serve a value that a crash would lose. For handing finalized blocks to peers that is the right trade: the block is finalized and the peer's copy stays valid whatever happens to our disk, and if we lose it we refetch it like any other gap.

A height that no started sync has flushed yet is reported as missing. The actor forwards finalized requests at the tail of each batch, after that batch's sync has been started and has published, so a height finalized or backfilled during the batch is served rather than missed. Anything that still misses reaches the peer as a resolver error, which retries.

A pruned height disappears immediately: `prune` drops a section's locations and its capture under one guard before the journal removes any files. A read already under way finishes normally, because a blob handle stays readable after its name is unlinked.

Corruption is not reported as absence. A value inside the captured range that fails its checksum is an error, not a miss, so a peer does not retry a hole forever and the condition shows up in metrics.

## Marshal

`Certificates` and `Blocks` gain `reader()`. `prunable::Archive` returns one; `immutable::Archive` returns `None` and keeps the inline path unchanged.

When both finalized stores expose a reader, the actor spawns a serving task holding both. `Key::Finalized` produce requests are forwarded to it; every other produce key is still answered inline. The task serves at most `max_repair` requests at once with the same number able to queue, and drops anything it cannot take, which the peer sees as a retryable error. It never panics: unlike the inline path, a read failure there is survivable, so it is logged and counted.

Both paths encode the response through the same function, so the wire format is identical whichever answers.

## What this costs

**Memory.** A `Location` is 16 bytes where a position was 8, so tracking each index item goes from `8 + 8` to `8 + 16` bytes, and an archive using an 8-byte translated key goes from about 40 to about 48 bytes per key. Every `prunable::Archive` pays this, including ones that never build a reader.

**A lock on the put path.** Each put takes a read guard for the duplicate check and a write guard for the insert. Both cover a map operation with no I/O.

**One detection path narrows.** A read by index no longer reads the index-journal record, so corruption there surfaces at startup replay rather than incidentally at read time. Values carry their own checksum, which every read still verifies, so nothing is served unverified.

## Breaking changes

`Certificates` and `Blocks` gain a required associated type (`Reader`) and a required method (`reader`). Both traits are BETA and are implemented by applications, so any external implementation must add both. Adding an associated type cannot be made backward compatible, since associated-type defaults are not stable.

`archive::Error` gains a `SectionFull` variant, which breaks exhaustive matches on it.

## What does not change

No on-disk format changes, so there is no migration and existing archives open as they are. The writer's own behavior is unchanged apart from the map contents and the lock: same recovery, same pruning, same durability. Deployments using immutable archives for finalized storage are unaffected.

## Metrics

`forwarded` and `dropped` on the mailbox, `served`, `missing`, and `failed` on the serving task. `missing` covers a height that is not yet published or already pruned; `failed` covers a read that returned an error, so persistent corruption is visible rather than hidden among ordinary misses.

`produce_serving_duration` measures the time the actor loop still spends answering produce requests inline.

### DIFF
--- a/consensus/src/marshal/core/actor.rs
+++ b/consensus/src/marshal/core/actor.rs
@@ -6,6 +6,7 @@ use super::{
     durability::{DispatchGate, Durable as _},
     floor::{Floor, State as FloorState},
     mailbox::{CommitmentFallback, Mailbox, Message},
+    serving,
     stream::Stream,
     subscriptions::{Key as SubscriptionKey, KeyFor as SubscriptionKeyFor, Subscriptions},
     variant::NoBuffer,
@@ -37,7 +38,7 @@ use commonware_resolver::{Delivery, Resolver, TargetedResolver};
 use commonware_runtime::{
     BufferPooler, Clock, ContextCell, Handle, Metrics, Spawner, Storage, spawn_cell,
     telemetry::{
-        metrics::{Gauge, GaugeExt, MetricsExt as _},
+        metrics::{Gauge, GaugeExt, Histogram, HistogramExt as _, MetricsExt as _, histogram},
         traces::TracedExt as _,
     },
 };
@@ -157,6 +158,8 @@ where
     finalized_height: Gauge,
     // Latest processed height
     processed_height: Gauge,
+    // Time the actor loop spends answering produce requests inline
+    produce_serving: Histogram,
 }
 
 impl<E, V, P, FC, FB, ES, T, A> Actor<E, V, P, FC, FB, ES, T, A>
@@ -230,6 +233,11 @@ where
         // Create metrics
         let finalized_height = context.gauge("finalized_height", "Finalized height of application");
         let processed_height = context.gauge("processed_height", "Processed height of application");
+        let produce_serving = context.histogram(
+            "produce_serving_duration",
+            "Time the actor loop spends answering produce requests inline",
+            histogram::Buckets::LOCAL,
+        );
         if let Some(last_processed_height) = last_processed_height {
             let _ = processed_height.try_set(last_processed_height.get());
         }
@@ -269,6 +277,7 @@ where
                 finalized_blocks,
                 finalized_height,
                 processed_height,
+                produce_serving,
             },
             Mailbox::new(sender, config.max_pending_acks),
             floor,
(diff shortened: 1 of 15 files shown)


danlaine:
Superseded by #4487

PR #4461, opened 2026-08-11
[storage] Retain Merkle ancestors in detached workers

patrick-ogrady:
Merkle batches intentionally hold weak parent links, so a detached merkleization job must retain every uncommitted ancestor that may provide nodes absent from its committed snapshot. Without that ownership, applying and dropping an older view or cancelling the waiting future can let a `Strategy::spawn` worker outlive the stack it is traversing, producing `peak missing` or missing-child panics.

- Capture the full live ancestor chain before snapshotting or handing work to the strategy.
- Move that guard into the detached authenticated-Journal, compact-QMDB, and Current-QMDB graft workers, retaining it through each worker’s last ancestor lookup.
- Add regressions for committed-prefix view stacks and cancelled, queued Rayon jobs, including clean ancestor release after completion.

### DIFF
--- a/storage/src/journal/authenticated.rs
+++ b/storage/src/journal/authenticated.rs
@@ -1057,7 +1057,7 @@ mod tests {
             },
             operation::Committable,
         },
-        utils::{DropMonitor, block_rayon},
+        utils::detached::{DropMonitor, block_strategy},
     };
     use commonware_codec::Encode;
     use commonware_cryptography::{Sha256, sha256::Digest};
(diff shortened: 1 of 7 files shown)


roberto-bayardo (dismissed):
The fix is sound and the placement is the subtle part done right: `retain_ancestors()` runs **before** `snapshot()`, which closes the window where another task applies-and-drops an ancestor between the snapshot and the spawn. 

Two non-blocking notes:

- `retain_ancestors` is gated `#[cfg(feature = "std")]` but only needs `alloc` (`Vec`/`Arc`/`Weak`) -- only matters if merkle ever targets no_std+alloc.
- The walk intentionally stops at the first dead weak link, which is correct under the "chain alive at call time; committed prefixes resolve via the snapshot" contract your prefix-dropped tests prove. A doc comment stating that contract on `retain_ancestors` would help the next reader distinguish it from a bug.

The test infrastructure is a nice addition on its own: `block_strategy` forcing the queued-then-cancelled worker interleaving and `DropMonitor` distinguishing clean completion from unwind are exactly the tools this class of bug needs.

PR #4462, opened 2026-08-12
[glue] Coalesce stateful database syncs across finalizations

patrick-ogrady:
## Summary

- Separate recovery checkpoint publication from durability: `ManagedDb::apply` publishes every finalized state as an independently rewindable checkpoint, while `finalize` makes the captured checkpoint prefix durable.
- Keep at most one stateful database sync active and coalesce finalized blocks arriving behind it into one successor sync.
- Make compact immutable and keyless QMDB append non-durable witness checkpoints during apply, matching full QMDB recovery semantics without acknowledgement dependencies.
- Coalesce buffered state-sync handoffs behind one durability barrier while preserving pruning, shutdown, and restart ordering.

## Motivation

Starting a database sync for every finalized block serializes QMDB journal flushes. Consecutive finalizations can share the expensive durability operation, but they cannot share a single recovery point: Marshal may persist any acknowledged finalized height and startup must be able to rewind every database to that exact checkpoint.

Full QMDB already journals each applied state before syncing. Compact QMDB previously appended only the current witness when durability started, so several applies before one sync could omit intermediate rewind targets. This change preserves every checkpoint and coalesces only the fsync work.

## Correctness

- Every successful `apply` exposes an independently rewindable checkpoint.
- A completed durability barrier covers every checkpoint published before it started; later applies remain a dirty suffix for a successor barrier.
- Marshal acknowledgements are released only after a barrier covering their heights succeeds. Aborted or failed barriers release none.
- Compact QMDB tracks the prefix captured by an in-flight sync separately from witnesses appended afterward.
- Equal-target recovery still delegates to the database, allowing a published but non-durable compact target to be made durable before return.
- Pruning waits for target coverage and quiesces verification readers. State-sync completion metadata advances only after the coalesced handoff prefix is durable.

## Testing

- `just test -p commonware-glue` (327 passed)
- `just test -p commonware-storage` (2,924 passed)
- `just test -p commonware-storage compact::tests::` (86 passed)
- `just test-docs -p commonware-storage -p commonware-glue`
- `just check-stability -p commonware-storage -p commonware-glue`
- `just check-fmt`
- `git diff --check`


storage/src/qmdb/compact/witness.rs
```diff
@@ -232,13 +237,38 @@ impl<E: Context, F: Family, D: Digest> Store<E, F, D> {
         *self.tip_witness.write() = witness;
     }
 
+    /// Apply the current compact state to the witness journal without making it durable.
```
patrick-ogrady: not sure if "without making it durable" part is necessary.

PR #4465, opened 2026-08-12
[storage] Batch the graft-input node reads

roberto-bayardo:
## Summary

When a `current` DB merkleizes a commit, it reads one small ops-tree node for every bitmap chunk the commit touched. `read_graft_inputs` issued these one node at a time via `get_node`. The buffered blob backend performs its `pread` synchronously when polled, so every page-cache miss was a blocking read executed serially on the same thread that is doing the merkleize. Even cache hits paid a per-node future, a lock acquisition, and a metrics update, thousands of times per commit.

This change fetches the whole set in one call instead:

- New `Storage::get_nodes` takes a sorted list of positions and returns one node per slot. Unlike `get_node`, an unavailable node is an `ElementPruned` error rather than an absent value, which suits the callers that require every node they ask for; `get_node` keeps its `Option` so callers can still probe for pruned nodes without an error. The default implementation is a `get_node` loop, so no other impl needs to change.
- The persisted tree implements it by serving memory-resident nodes directly and handing the rest to the node journal's existing `read_many`, which resolves all cache hits under one lock per blob and fetches the misses concurrently.
- `read_graft_inputs` becomes a single `get_nodes` call.
- `Error::MissingGraftedLeaf` is removed. `read_graft_inputs` was its only raise site, and the batched call now reports the unavailable position itself.

## Why these reads may miss the cache

The node a chunk needs is determined by where that chunk sits in history, and a commit touches chunks spread across *all* of history. So the read set is large and effectively random. A cache bigger than the database still misses the first time each page is touched; a cache smaller than the working set misses constantly. There is no cache size at which these reads are free -- which is why fetching them well matters.

## Benchmarks

eth-replay follower (EPYC 9354P, 20k-block windows from a 5M-block seed, mimalloc, paired A/B, identical `db_bytes` in every pair), measured against main at both a 1-thread and an 8-thread strategy pool:

| | wall | merkleize |
|---|---|---|
| main, 8 threads | 54.22s | 31.52s |
| this PR, 8 threads | 51.55s (**-4.9%**) | 29.16s (**-7.5%**) |
| main, 1 thread | 45.02s | 27.35s |
| this PR, 1 thread | 42.60s (**-5.4%**) | 25.12s (**-8.1%**) |

The win does not depend on the strategy pool: the mechanism is bulk page-cache passes and runtime-level IO concurrency, not pool parallelism. It is slightly larger with the minimal pool, where the one thread does all the work and serial faulting costs it most. (The 1-thread rows are faster overall because this workload's parallel duty cycle is too low to pay for an idle pool; see #4432 for that analysis.)

Cache pressure changes the magnitude but not the direction. Under the pending performance stack (#4423 + #4432 + #4454 + #4440, 8 threads, both arms), whose pipelined merkleize already overlaps part of these reads with hashing -- which is why the percentages are smaller than above:

| scenario | wall | merkleize |
|---|---|---|
| cache smaller than working set (2G): constant misses | 83.1s -> 80.1s (**-3.7%**) | 54.8s -> 51.7s (**-5.6%**) |
| cache larger than data (16G / 32G / 64G): misses only on first touch of each page | -1.2% to -2.4% | -2.5% to -3.2% |
| 100k-block run (32G cache) | 192.3s -> 190.9s (-0.7%) | 114.3s -> 112.4s (-1.7%) |

This change touches different code regions than that stack and merges with it in either order.

Read-syscall counts per arm (strace attached after init) pin down the mechanism:

- With ample cache, both arms take ~0.9M misses per window -- all first touches, nearly independent of cache size. The win there comes from bulk-serving the hits and overlapping the misses.
- With the 2G cache, misses grow 7x (~6.4M) and **both arms take the same number** -- yet the win nearly doubles. The improvement is not fewer reads; it is that misses no longer execute one at a time on the merkleize thread.
- Every run was served from OS memory (zero block-device reads). Against real disk latency, the gap should widen further.

## Microbenchmarks

Both measured as `origin/main` vs this branch alone (no other changes in the diff), on the same host.

constantinople `current::ordered::fixed::mmb` (1M keys, 32k reads + 32k updates per iteration, p50 over 15 iterations, paired A/B x2). Speculative roots are bit-identical between the two binaries at every iteration.

| page cache | main | this PR |
|---|---|---|
| 512 MiB (working set fits) | 28.50ms | 27.18ms (**-4.6%**) |
| 32 MiB (thrashing) | 157.1ms | 155.4ms (-1.1%) |

The absolute saving is similar in both regimes (~1.3-1.7ms per commit); at 32 MiB the rest of the pipeline balloons on value-read misses this PR does not touch, which shrinks the percentage.

criterion `qmdb::merkleize`, all `current` variants (80 variant/config combinations, two paired runs each): this PR is faster in 78 of 80, typically -1% to -4%, with the `ordered` variants clustering at -2% to -4.6%.

## Validation

- New test: the persisted tree's `get_nodes` agrees with per-position `get_node` on every available position (journal-resident and memory-resident) and rejects the positions `get_node` reports as absent, for both mmr and mmb families.
- New test: the default `get_nodes` serves retained positions and rejects pruned ones, so implementations that do not override it inherit the strict contract.
- The qmdb, merkle, and journal suites pass on this base; `--no-default-features` clean.

### DIFF
--- a/storage/src/merkle/persisted/full.rs
+++ b/storage/src/merkle/persisted/full.rs
@@ -639,6 +639,47 @@ impl<F: Family, E: Context, D: Digest, S: Strategy> Merkle<F, E, D, S> {
         }
     }
 
+    /// Batched [`Self::get_node`]: `positions` must be strictly increasing. Memory-resident
+    /// nodes are served directly; the rest go through the journal's batched read, which
+    /// serves page-cache hits in bulk and fetches misses concurrently.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`Error::ElementPruned`] for the first of `positions` that falls below the
+    /// journal's pruning boundary.
+    pub async fn get_nodes(&self, positions: &[Position<F>]) -> Result<Vec<D>, Error<F>> {
+        debug_assert!(positions.is_sorted_by(|a, b| a < b));
+        let bounds = self.journal.bounds();
+        let mut nodes = vec![None; positions.len()];
+        let mut journal_positions = Vec::with_capacity(positions.len());
+        for (slot, &position) in nodes.iter_mut().zip(positions) {
+            if let Some(node) = self.mem.get_node(position) {
+                *slot = Some(node);
+            } else if *position >= bounds.start {
+                // In-subsequence order is preserved, so this stays strictly increasing.
+                journal_positions.push(*position);
+            } else {
+                return Err(Error::ElementPruned(position));
+            }
+        }
+        // Within-bounds reads are guaranteed not to return `ItemPruned` (see
+        // [`crate::journal::contiguous::Contiguous::read`]).
+        let items = if journal_positions.is_empty() {
+            Vec::new()
+        } else {
+            self.journal
+                .read_many(&journal_positions)
+                .await
+                .map_err(Error::Journal)?
+        };
+        // The unfilled slots are exactly the journal subsequence, in the order it was built.
+        let mut items = items.into_iter();
+        Ok(nodes
+            .into_iter()
+            .map(|node| node.unwrap_or_else(|| items.next().expect("one item per journal read")))
+            .collect())
+    }
+
     /// Return the pinned nodes needed to authenticate a lower leaf boundary at `loc`.
     pub async fn pinned_nodes_at(&self, loc: Location<F>) -> Result<Vec<D>, Error<F>> {
         if !loc.is_valid() {
@@ -987,6 +1028,10 @@ impl<F: Family, E: Context, D: Digest, S: Strategy> crate::merkle::storage::Stor
 
     async fn get_node(&self, position: Position<F>) -> Result<Option<D>, Error<F>> {
         Self::get_node(self, position).await
+    }
+
+    async fn get_nodes(&self, positions: &[Position<F>]) -> Result<Vec<D>, Error<F>> {
+        Self::get_nodes(self, positions).await
     }
 }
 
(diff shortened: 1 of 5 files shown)


storage/src/merkle/persisted/full.rs
```diff
@@ -639,6 +639,47 @@ impl<F: Family, E: Context, D: Digest, S: Strategy> Merkle<F, E, D, S> {
...
+    /// Batched [`Self::get_node`]: `positions` must be strictly increasing. Memory-resident
+    /// nodes are served directly; the rest go through the journal's batched read, which
+    /// serves page-cache hits in bulk and fetches misses concurrently.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`Error::ElementPruned`] for the first of `positions` that falls below the
+    /// journal's pruning boundary.
+    pub async fn get_nodes(&self, positions: &[Position<F>]) -> Result<Vec<D>, Error<F>> {
+        debug_assert!(positions.is_sorted_by(|a, b| a < b));
```
patrick-ogrady: we hard assert elsewhere for this?

actually, is this an expensive check? I wonder if we should just say it must be sorted?

storage/src/merkle/persisted/full.rs
```diff
@@ -639,6 +639,47 @@ impl<F: Family, E: Context, D: Digest, S: Strategy> Merkle<F, E, D, S> {
...
+            if let Some(node) = self.mem.get_node(position) {
+                *slot = Some(node);
+            } else if *position >= bounds.start {
+                // In-subsequence order is preserved, so this stays strictly increasing.
+                journal_positions.push(*position);
+            } else {
+                return Err(Error::ElementPruned(position));
+            }
+        }
+        // Within-bounds reads are guaranteed not to return `ItemPruned` (see
```
patrick-ogrady, line 669: usually recommend \n above block comments

storage/src/merkle/persisted/full.rs
```diff
@@ -639,6 +639,47 @@ impl<F: Family, E: Context, D: Digest, S: Strategy> Merkle<F, E, D, S> {
...
+        // [`crate::journal::contiguous::Contiguous::read`]).
+        let items = if journal_positions.is_empty() {
+            Vec::new()
+        } else {
+            self.journal
+                .read_many(&journal_positions)
+                .await
+                .map_err(Error::Journal)?
+        };
+        // The unfilled slots are exactly the journal subsequence, in the order it was built.
```
patrick-ogrady, line 680: usually recommend \n above block comments

storage/src/merkle/storage.rs
```diff
@@ -17,6 +17,36 @@ pub trait Storage<F: Family>: Send + Sync {
...
+    ///
+    /// # Errors
+    ///
+    /// Returns [`Error::ElementPruned`] for the first of `positions` where [`Storage::get_node`]
+    /// would yield `None`.
+    fn get_nodes(
+        &self,
+        positions: &[Position<F>],
+    ) -> impl Future<Output = Result<Vec<Self::Digest>, Error<F>>> + Send {
+        async move {
```
patrick-ogrady, line 37: do we want a default?

storage/src/qmdb/current/batch.rs
```diff
@@ -200,6 +200,31 @@ impl<F: Graftable, D: Digest, R: Readable<Family = F, Digest = D>, S: MerkleStor
         }
         self.base.get_node(pos).await
     }
+
+    async fn get_nodes(&self, positions: &[Position<F>]) -> Result<Vec<D>, merkle::Error<F>> {
```
patrick-ogrady, line 204: I guess we don't need to assert here (should be consistent in any case?)

storage/src/qmdb/current/db.rs
```diff
@@ -1111,17 +1110,24 @@ pub(super) async fn read_graft_inputs<F: merkle::Graftable, D: Digest, const N:
...
+        .iter()
+        .map(|&(chunk_idx, _)| {
+            let leaf_start = Location::<F>::new((chunk_idx as u64) << grafting_height);
+            F::subtree_root_position(leaf_start, grafting_height)
+        })
+        .collect();
+
+    // Chunk indices ascend and subtree roots ascend with their leaf ranges, satisfying
+    // `get_nodes`'s ordering requirement.
+    let nodes = ops_tree.get_nodes(&positions).await?;
```
patrick-ogrady, line 1125: I think you could hand the iterator above in rather than ever collecting?

patrick-ogrady (commented):
just nits

storage/src/merkle/persisted/full.rs
roberto-bayardo: it's not that expensive (a single scan) but it still feels like something more appropriate for a debug_assert!.  I guess we should be consistent though.

roberto-bayardo: Changed to assert! for the sake of consistency, but I'd prefer we go through and make them all debug_assert!

storage/src/merkle/persisted/full.rs
roberto-bayardo: done

storage/src/merkle/persisted/full.rs
roberto-bayardo: done

storage/src/merkle/storage.rs
roberto-bayardo: Kept the default: for the in-memory impls (`Mem`, grafting, bitmap) a per-node loop is already the optimal batched path, since there is no I/O to coalesce. Only journal-backed storage gains from overriding, and the persisted tree does.

storage/src/qmdb/current/batch.rs
roberto-bayardo: Right: the adapter only ever forwards an in-order subsequence of its input, and the base's `get_nodes` asserts, so a check here would be redundant.

storage/src/qmdb/current/db.rs
roberto-bayardo: I think the slice is the more appropriate signature here, independent of cost: this method's contract is positional. Slot `i` of the result is defined by `positions[i]`, the length-equality guarantee is stated against the input's length, and the strictly-increasing precondition is a whole-sequence property both sides can inspect. An iterator erases the thing those statements are made about. Every caller also needs the positions after the call (correlating results, attributing `ElementPruned`), so an iterator-taking version would just force the collect to the call site. It also matches the layer below: `read_many` and `try_read_many_sync` both speak slices.

patrick-ogrady (approved):

PR #4471, opened 2026-08-12
[storage] Slim Readable to synchronous node resolution

roberto-bayardo:
## Summary

`Readable` now carries only what its consumers dispatch on -- `size` and `get_node`: synchronous node resolution over whatever the implementing view holds, the sync counterpart of the async `merkle::storage::Storage`.

Four members are removed: `pruning_boundary`, the `bounds()` and `leaves()` defaults, and the `Error` associated type.

### Why `pruning_boundary` had to go

It meant something different in every implementation, so no accurate shared contract could be written:

| Impl | What the value was |
|---|---|
| `Mem` | the location pruning was performed to |
| `full::Merkle` | the in-memory boundary, which *flushing* raises without anything being pruned |
| `merkle::batch::MerkleizedBatch` | a value inherited from the parent `Mem`, describing data the batch does not hold |

Availability was decoupled from the boundary in both directions: `Mem` serves pinned nodes *below* it (out of `pinned_nodes`), while a batch serves almost nothing *above* it, since its `get_node` only covers the batch chain.

### Why the others followed

- `bounds()` was defined in terms of `pruning_boundary` and had no callers through the trait at all -- every `.bounds()` in the workspace resolves to an inherent method on a concrete type, so deleting the default produced zero call-site errors. The inherent `bounds()` on `Mem` and `Merkle` are untouched.
- `leaves()` duplicated inherent methods that already exist on every implementing type, and no caller dispatched on the trait's copy.
- `Error` became unreferenced the moment no trait method returned `Result`: it was constrained at every use site (`Error = merkle::Error<F>`) but never read.

None of the four was consumed polymorphically; every surviving call resolves on a concrete type or an inherent method. `Mem` gains an inherent `pruning_boundary` accessor for the one site in `merkle::batch` that needed trait-qualified access.

### `get_node` contract

The trait documented `None` as meaning "pruned / out of bounds". No implementation restricts it that way: `full::Merkle` returns `None` for **flushed** positions the journal still retains, a merkleized batch returns `None` for every position it did not touch, and the layered adapters in `qmdb::current` use `None` as a fall-through signal. The contract is now stated as a property of the interface (whether *this* view can serve the node) rather than a claim about the underlying data.

## Behavior

Unchanged. The load-bearing case is `update_leaf` rejecting leaves synced out of memory with `ElementPruned`: batches forked from a persisted tree capture the same in-memory boundary as before, now via the inherent method rather than the trait. The existing `update_leaf_after_sync_returns_pruned` regression test covers exactly that path and passes for both mmr and mmb.

## Validation

The full storage merkle and qmdb suites pass under `--profile all` (slow tests included). `just check-stability` is clean, rustdoc is clean under `-D warnings`, and `cargo check --workspace --all-targets` is clean -- the last of those matters because this removes public trait members, so any dependent crate implementing `Readable` would have surfaced.

### DIFF
--- a/docs/blogs/pipelining-simplex.html
+++ b/docs/blogs/pipelining-simplex.html
@@ -95,8 +95,8 @@ deployment of 50 validators, it sustained 200 blocks per second with
 Chou</a></div>
                         <div class="date">August 12th, 2026</div>
         </div>
-        <p>Simplex can now produce blocks as fast as its leader can build
-        them.</p>
+        <p>Simplex can now produce blocks as fast as its leader can
+        build them.</p>
         <p>Today, we’re introducing <a
         href="https://github.com/commonwarexyz/monorepo/pull/3352">Stable
         Leader</a> and <a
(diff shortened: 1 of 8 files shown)


danlaine (dismissed):
fwiw I would revert the comment changes on `Readable` and `Storage`. Also I think we could remove `leaves` from `Readable`.

roberto-bayardo:
Applied both suggestions: reverted the doc additions on `Readable`/`Storage` and removed `leaves()` as well. The trait is now just `size` + `get_node`.

Two deliberate exceptions. `get_node` keeps the corrected one-liner ("`None` if this structure cannot serve it") rather than restoring "pruned / out of bounds", since no implementation actually satisfies that claim. And I restored a one-line version of the sync framing on `Readable`: it states the implementor contract (`get_node` must not do I/O) that hot paths like the batch adapter rely on.

danlaine (approved):

PR #4478, opened 2026-08-13
[utils] Parameterize futures Pool on a lifetime

danlaine:
## Summary

`commonware_utils::futures::Pool` and `AbortablePool` stored their futures as `Box<dyn Future + Send>`, which means `Box<dyn Future + Send + 'static>`. A pooled job could therefore never hold a borrow, so anything a job needed had to be cloned or wrapped in an `Arc` first, even when the data outlived the job.

Both types now take a lifetime. `Pool<'a, T>` and `AbortablePool<'a, T>` accept futures that live at least as long as `'a`, so a job can borrow from the scope that owns the pool. `Pool<'static, T>` behaves exactly as `Pool<T>` did.

## What changed

In `utils/src/futures.rs`, the two boxed-future aliases carry `+ 'a`, the two pool types gained the lifetime parameter, and `push` now accepts `impl Future<Output = T> + Send + 'a` instead of requiring `'static`. Everything else about the pools is untouched, including the dummy future that keeps an empty pool from resolving instantly, `cancel_all`, and the `Aborter` handle.

`Aborter` deliberately has no lifetime. It holds only an `AbortHandle`, so a borrowing future can still be aborted through it, and the handle may outlive the borrow. Aborting a future that is already gone is a no-op, which is what simplex and resolver rely on when they stash aborters in longer-lived state.

Callers changed only where a lifetime has to be named.

## Where the lifetime is written

A pool stored in a struct needs a named lifetime, because Rust will not infer one in a field, so those spell `'static`. There are eight of them, in glue's verification handler, resolver's serve/fetch/delivery pools, aggregation, ordered broadcast, and the qmdb sync request tracker, plus one test type alias in marshal.

Everywhere else the lifetime is inferred. Functions that only forward a pool take `Pool<'_, T>` or `AbortablePool<'_, T>`, since they do not care about its scope. Local bindings with a type annotation use `'_`. Constructor calls such as `Pool::<PooledSync>::default()` say nothing at all and are unchanged from before.

## Why `T: 'a` is not written on the impls

The pools need `T` to outlive `'a` for their stored future type to be well formed. That requirement can live in one of two places, and the choice matters.

Written on the impl blocks as `impl<'a, T: Send + 'a>`, it becomes a precondition on calling any method, which every holder then has to prove. `Requests<F, Op, D, E>` in `storage/src/qmdb/sync` cannot prove it, because its own declaration never bounds `Op` or `E`, so it fails to compile with `the parameter type Op may not live long enough`. Making it compile would mean adding `'static` bounds to a generic struct in storage.

Left to the pools' own fields, the same requirement is derived once inside `utils` and handed to callers as a fact instead. The impls stay at `T: Send`, and nothing downstream needs new bounds.

## Compatibility

`commonware_utils::futures` is BETA. This is a breaking API change with no wire or storage impact. Outside callers that name the type in a field, alias, or signature need to add `'static`, and callers that only construct a pool need no change.


clabby (dismissed):

BrendanChou:
Would it make sense to introduce `ScopedPool<'a, T>` and `keep Pool<T>` as an alias for `ScopedPool<'static, T>`?

```rust
pub struct ScopedPool<'a, T> {
    // ...
}

pub type Pool<T> = ScopedPool<'static, T>;
```

That would preserve the existing API and limit lifetime annotations to callers that actually borrow. We could also leave `AbortablePool` unchanged unless the follow-up needs scoped abortable futures.

danlaine:
> Would it make sense to introduce `ScopedPool<'a, T>` and `keep Pool<T>` as an alias for `ScopedPool<'static, T>`?
> 
> ```rust
> pub struct ScopedPool<'a, T> {
>     // ...
> }
> 
> pub type Pool<T> = ScopedPool<'static, T>;
> ```
> 
> That would preserve the existing API and limit lifetime annotations to callers that actually borrow. We could also leave `AbortablePool` unchanged unless the follow-up needs scoped abortable futures.

I suppose it depends how much we can about minimizing API churn. This code is BETA, which doesn't promise API stability, only wire/storage format stability. I'd be OK with this being a breaking API change on that basis.

clabby (approved):

PR #4480, opened 2026-08-13
[simplex] Pipelined Handoff

### LINKED ISSUE #4477: [consensus/simplex] Optional optimistic proposal across term boundaries
### Introduction

Stable leader (#3352) and optimistic validation (#3416) pipeline proposals within a term. Optimism does not cross a term boundary. The first proposal of a term waits for certified ancestry from the prior term. Each boundary therefore pays the full handoff latency that #3405 aimed to remove.

The inter-leader idea from #3405 still applies at the boundary. The incoming leader can propose on the outgoing leader's tip before that tip notarizes. This adds a trust assumption on the outgoing leader. If the outgoing leader's proposal never notarizes, the optimistic proposal built on it also fails.

### Why optional

The saving is one δ per term boundary. Deployments with long terms gain little; deployments with short terms or strict latency targets gain more. The added trust assumption means the feature should be opt-in.

### Implementation sketch

Per #3405, only leader logic changes. The incoming term's leader proposes during the final view of the outgoing term, once it holds a valid proposal for that view. If no proposal arrives, the leader proposes when it enters the term. Validators already admit votes up to the next term start; their logic does not change. The incoming term's leader must be identifiable before the term starts. This restricts which electors can support the feature.

BrendanChou:
Fixes #4477

Adds an opt-in *pipelined handoff*. The incoming term's leader proposes on the outgoing term's final view as soon as it holds a valid proposal. It does not wait for that view to certify. If no proposal arrives, the leader proposes when it enters the term.

**API.** A defaulted `Elector::elect_ahead` selects a round's leader without the certificate that unlocks it. Returning `Some` opts the local participant in and requires the result to equal `elect` for every unlocking certificate. `RoundRobin::with_pipelined_handoff()` enables the feature. Certificate-derived electors such as `Random` cannot support it.

**Voter.** Voting for a term's final view sets the incoming leader. `find_parent` can select the uncertified tip unless a nullification has abandoned the outgoing term. Before signing, `construct_notarize` rechecks the tip and withholds the vote after nullification, failed certification, or equivocation. Non-leader validators keep the same behavior: they buffer votes up to the next term start and verify term starts against explicitly certified ancestry. The batcher, certification, and wire format do not change.

**Performance.** Rotating leaders benefit most because every view is a handoff. Sustained view time drops from two network trips to one. In the deterministic cluster with 100ms links, the measured time is about 120ms per view with the handoff and 228ms without it. With stable leaders, the handoff moves each boundary view one network trip earlier but does not change sustained throughput.

**Trust.** The incoming leader trusts the outgoing leader not to equivocate. If the outgoing tip never notarizes, validators cannot use the pipelined proposal. The usual timeout path then nullifies the incoming term, so the feature is opt-in.

**Tests.** State tests cover the opt-in, proposal construction on an uncertified tip, fallback after early nullification, explicit peer verification, rotating terms, and vote withholding after nullification, failed certification, or equivocation. A voter integration test proposes and notarizes across the boundary without receiving a certificate. Two cluster tests cover boundary latency with two-view terms and sustained view time with rotating leaders. The pipelining blog now describes the opt-in.


consensus/src/simplex/elector.rs
```diff
@@ -227,6 +229,30 @@ pub trait Elector<S: Scheme>: Clone + Send + 'static {
...
+    /// [`RandomElector`] cannot elect ahead.
+    ///
+    /// The voter may call this method several times per view. Electors should
+    /// return a precomputed result.
+    ///
+    /// This is local policy: participants with different settings interoperate.
+    /// The default opts out.
+    ///
+    /// [Pipelined Handoff]: crate::simplex#pipelined-handoff
+    fn elect_ahead(&self, round: Round) -> Option<Participant> {
```
patrick-ogrady: you can just do _round: ?

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -699,6 +726,11 @@ impl<E: Clock + CryptoRng + Metrics, S: Scheme<D>, L: Elector<S>, D: Digest> Sta
         if self.in_issuance_window(view) && !self.optimistic_parent_ready(view) {
             return None;
         }
+        // The exception is a pipelined-handoff proposal, which waits for
+        // evidence of its cross-term parent.
+        if !self.handoff_parent_ready(view) {
+            return None;
+        }
```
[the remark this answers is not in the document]
patrick-ogrady: Good catch—this was valid against `df564ebc` and is fixed by `a5b952094` in the current HEAD.

The actor now drops a pending proposal request when its captured ancestry is no longer preferred or independently valid and a replacement parent is available. State releases the one-shot build latch only while the proposal slot is empty, then immediately reissues the same view using the certified fallback parent.

`test_pipelined_handoff_reissues_stalled_proposal_after_parent_nullification` covers the stalled-build → tip-nullification → same-view fallback-reissue trace. Resolving as fixed.

consensus/src/simplex/actors/voter/state.rs
```diff
@@ -686,16 +686,20 @@ impl<E: Clock + CryptoRng + Metrics, S: Scheme<D>, L: Elector<S>, D: Digest> Sta
...
+        // Check round-local eligibility before the ancestry check, which
+        // resolves the parent payload on every event.
         if !self.views.get(&view)?.can_construct_notarize() {
             return None;
         }
-        // Optimistic views wait for local evidence of their parent; anything
-        // outside the issuance window carries certified ancestry already.
+
+        // An optimistic view requires local evidence for its parent. A view
+        // outside the issuance window already has certified ancestry.
```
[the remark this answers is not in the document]
patrick-ogrady: Thanks for flagging this. This ordering is not reachable through the voter actor.

`State::proposed` runs after the asynchronous build completes and revalidates the exact captured parent. If nullification, failed certification, equivocation, or a conflicting certificate was processed first, the proposal is rejected.

If `proposed` accepts first, the actor exclusively owns State and proceeds through the same event-loop turn to `construct_notarize` before another select arm can process an invalidation. An invalidation may be queued during that work, but it cannot mutate State until the next iteration, after the vote was constructed. This is a vote-before-invalidation ordering, not a stale post-invalidation vote.

The same-term optimistic guard remains; term-start safety is owned by completion-time ancestry validation plus the actor's single-owner transition. Resolving as a false positive.

patrick-ogrady:
bugbot run

consensus/src/simplex/elector.rs
```diff
@@ -227,6 +229,29 @@ pub trait Elector<S: Scheme>: Clone + Send + 'static {
...
+    ///
+    /// The voter may call this method several times per view. Electors should
+    /// return a precomputed result.
+    ///
+    /// This is local policy: participants with different settings interoperate.
+    /// The default opts out.
+    ///
+    /// [Pipelined Handoff]: crate::simplex#pipelined-handoff
+    fn elect_early(&self, _round: Round) -> Option<Participant> {
+        None
```
patrick-ogrady: I think we should remove this default. This is one we want any implementer to be specific about imho.

BrendanChou: Agreed. Removed the default in 1849d8a79 so every implementation now states its policy explicitly: round-robin conditionally opts in, Random and the test electors opt out, twins preserves its deliberate opt-out, and the wrapped elector delegates.

BrendanChou:
Superseded by #4739, which now targets main and includes all commits from this PR together with application-controlled prebuilding and optional early publication. The commits and this review discussion are preserved; further review should continue on #4739.

PR #4481, opened 2026-08-13
[consensus/simplex] Dispatch proposal and verification before voter persistence

patrick-ogrady:
Stacks on #4448.

#4448 starts the next proposal after construction so application work can overlap journal sync and publication. This follow-up makes that checkpoint use the same lifecycle as loop entry:

- reconcile pending proposal and verification requests with the current view
- start both kinds of newly eligible application work
- preserve proposal-before-verification order while State and Round prevent duplicate requests

Both checkpoints are intentional. Loop entry covers startup, replay, retries, and work already eligible for the iteration. The post-construction checkpoint covers optimistic child work exposed while processing the selected event and constructing votes. Requests started there join the following iteration waiters, after the current journal sync and publication complete.

The deterministic regressions now prove that:

- the next proposal reaches the application while journal sync is blocked
- a stale proposal does not suppress replacement work after a view transition
- a buffered optimistic child starts verification before its parent notarize is published

### DIFF
--- a/consensus/src/simplex/actors/voter/actor.rs
+++ b/consensus/src/simplex/actors/voter/actor.rs
@@ -412,6 +412,40 @@ impl<
         Some(Request(context, span, receiver))
     }
 
+    /// Starts eligible proposal and verification requests.
+    async fn dispatch_application_requests(
+        &mut self,
+        resolver: &mut resolver::Mailbox<S, D>,
+        pending_propose: &mut Option<Request<Context<D, S::PublicKey>, D>>,
+        pending_verify: &mut Option<Request<Context<D, S::PublicKey>, bool>>,
+    ) {
+        // Keep requests for optimistic future views and clear requests for
+        // exited views. Certification for an exited view can continue after
+        // its verification receiver is dropped.
+        let current_view = self.state.current_view();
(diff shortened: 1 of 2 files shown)


consensus/src/simplex/actors/voter/actor.rs
```diff
@@ -1248,14 +1261,13 @@ impl<
...
-                    // signing a notarize advances the optimistic frontier.
-                    // Dispatch the request before the journal sync and
-                    // broadcast below so the automaton builds during that
-                    // work. The result is consumed on a later iteration.
-                    if pending_propose.is_none() {
-                        pending_propose = self.try_propose().await;
-                    }
+                    // Start optimistic child requests before journal sync and
+                    // publication. Their responses join the next iteration's waiters.
+                    self.dispatch_application_requests(
```
patrick-ogrady, line 1266: With this change (and the parent this is based on), we may ask an application to build on a certificate we haven't yet synced. This would mean after restart, the node may not be able to "fetch it".

I think this is fine (the application shouldn't be doing anything too crazy with this hint). But brining it up just because it came to mind.

BrendanChou (approved):

PR #4482, opened 2026-08-14
[cryptography/bls12381] Add Multimmit aggregate signature scheme

clabby:
## Overview

Add the BLS12-381 aggregate signature scheme used by [Multimmit](https://arxiv.org/abs/2607.21021v2).

Multimmit quorum certificates contain votes from many participants. Most votes have identical contents in the common case, while votes may differ when participants observe different chain positions or extensions. Attributed aggregation keeps the certificate signature component to one group element while verifying it against the exact public-key, namespace, and message transcript. Equal messages are grouped, so verification requires one pairing term per distinct message rather than per signer.

The new primitives also support batching independent signatures and aggregate transcripts using random scaling, with bisection to isolate invalid inputs. This is useful beyond Multimmit for quorum certificates, heterogeneous signature batches, and threshold-share participation evidence.

Both MinPk and MinSig are supported. Repeated-message aggregation requires public keys to be bound to their owners, such as through verified proofs of possession.


cryptography/src/bls12381/primitives/variant.rs
```diff
@@ -13,14 +13,34 @@ use blst::{blst_final_exp, blst_fp12, blst_miller_loop};
...
+
+#[cfg(test)]
+pub(crate) fn final_exponentiations() -> usize {
+    FINAL_EXPONENTIATIONS.get()
+}
+
+#[cfg(test)]
+pub(crate) fn record_final_exponentiation() {
+    FINAL_EXPONENTIATIONS.set(FINAL_EXPONENTIATIONS.get() + 1);
+}
```
cronokirby: What's the point of this?

cryptography/src/bls12381/primitives/sharing.rs
```diff
@@ -311,12 +311,19 @@ impl<V: Variant> Sharing<V> {
...
         self.mode
     }
 
+    /// Returns how participants are assigned sharing evaluation points.
+    #[cfg(feature = "std")]
+    #[stability(ALPHA)]
+    pub const fn evaluation_mode(&self) -> Mode {
+        self.mode
+    }
+
```
cronokirby: We should just expose the existing method, if that's what we want; why do we need to expose it though?

cryptography/src/bls12381/primitives/ops/aggregate.rs
```diff
@@ -227,6 +305,39 @@ where
...
+/// Each entry attributes an exact `(namespace, message)` pair and signature to one public key.
+/// Public keys must be unique and non-zero, and the input must not be empty.
+///
+/// # Security
+///
+/// Every public key must be group-checked and have a verified proof of possession (PoP). This
+/// function verifies signatures, but not PoPs. Accepting a public key without a verified PoP can
+/// enable rogue-key attacks when the resulting aggregate is verified.
+#[stability(ALPHA)]
+pub fn aggregate_signatures<'a, V: Variant>(
```
cronokirby: What's the utility of this method over what already exists in this module?

cryptography/src/bls12381/primitives/ops/threshold.rs
```diff
@@ -93,6 +125,70 @@ pub fn verify_message<V: Variant>(
...
+/// Verifies and aggregates partial signatures without threshold recovery.
+///
+/// Each entry contains the namespace, exact message, and attributed partial signature. Signers must
+/// be unique and belong to `sharing`. Unlike [`recover`], this operation does not require or imply
+/// that the recovery threshold has been reached.
+///
+/// Aggregating already-invalid partials can hide cancelling errors. This constructor verifies every
+/// partial before combining it and should be used when assembling an aggregate from untrusted input.
+#[stability(ALPHA)]
+pub fn aggregate_partial_signatures<'a, V: Variant>(
```
cronokirby: For these methods, couldn't we accomplish the same thing by having a means of converting a PartialSignature into a Signature?

cronokirby (commented):
What's the particular use case that we want out of the API? I think we could simplify things substantially by providing a narrow API for doing "verify all of these sigs, smartly combining messages and public keys as possible".

Another small note: we could probably eat representing this concept as a Vec (or some other durable struct you build up) because no method actually uses the iterator in a streaming way, and I imagine callers will end up building up a vec, just to iterate over it, just to have the method build up a vec, which seems silly.

cryptography/src/bls12381/primitives/variant.rs
clabby: Was just for testing, though it's a bit redundant. Removed it

cryptography/src/bls12381/primitives/ops/aggregate.rs
clabby: We can remove it.

cryptography/src/bls12381/primitives/sharing.rs
clabby: Originally the LLM tried to be a bit fancy (depending on size + mode, it'd behave differently); Think we should probably remove it.

cryptography/src/bls12381/primitives/ops/aggregate.rs
```diff
@@ -95,6 +96,12 @@ impl<V: Variant> Signature<V> {
         &self.0
     }
 
+    /// Returns the aggregated group element.
+    #[stability(ALPHA)]
+    pub const fn element(&self) -> &V::Signature {
+        &self.0
+    }
```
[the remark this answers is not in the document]
clabby: This is used in multimmit, since we need to pass the raw signature to `Verifier::queue`. We could remove it in favor of a `queue_aggregate` function that unwraps the type internally, though I prefer this route.

PR #4483, opened 2026-08-14
[codec] Add Checksummed type

### LINKED ISSUE #1704: [codec] Add `Checksummed` type
A wrapper type that automatically adds a checksum when encoding and verifies the checksum when decoding. Can be used with any `Codec` type and can be wrapped/unwrapped to the underlying type `T` for ergonomics. Could use crc32, or could be generalized to use any `Hasher`

```
/// A wrapper around a [Codec] type that has a checksum.
///
/// Automatically creates the checksum when writing and verifies it when reading.
pub struct Checksummed<T: Codec, H: Hasher> {
    pub data: T,
    _hasher: PhantomData<H>,
}

impl<T: Codec, H: Hasher> Checksummed<T, H> {
    pub fn into_inner(self) -> T {
        self.data
    }
}

impl<T: Codec, H: Hasher> From<T> for Checksummed<T, H> {
    fn from(data: T) -> Self {
        Self {
            data,
            _hasher: PhantomData,
        }
    }
}

impl<T: Codec, H: Hasher> Read for Checksummed<T, H>
where
    H::Digest: Codec<Cfg = ()> + PartialEq,
{
    type Cfg = T::Cfg;

    fn read_cfg(buf: &mut impl Buf, cfg: &Self::Cfg) -> Result<Self, Error> {
        let data = T::read_cfg(buf, cfg)?;
        let hash = H::Digest::read_cfg(buf, &())?;
        let data_bytes = data.encode();
        let mut hasher = H::new();
        hasher.update(&data_bytes);
        let expected_hash = hasher.finalize();
        if hash != expected_hash {
            return Err(Error::Invalid("Checksummed", "checksum mismatch"));
        }
        Ok(Self {
            data,
            _hasher: PhantomData,
        })
    }
}

impl<T: Codec, H: Hasher> Write for Checksummed<T, H>
where
    H::Digest: Codec<Cfg = ()>,
{
        let data = self.data.encode();
        let mut hasher = H::new();
        hasher.update(&data);
        let hash = hasher.finalize();
        hash.write(buf);
        buf.put_slice(&data);
    }
}

impl<T: Codec, H: Hasher> EncodeSize for Checksummed<T, H>
where
    H::Digest: FixedSize,
{
        self.data.encode_size() + H::Digest::SIZE
    }
}
```

crazywriter1:
Closes #1704.

## Summary

Adds `Checksummed<T, C>`, a wrapper that appends a checksum of a value's encoding when writing and verifies it when reading. This lets any `Codec` type detect accidental corruption at rest or in transit without each caller hand-rolling the framing.

The wrapper is generic over a new `Checksummer` trait rather than over a hasher from `commonware-cryptography`, since the hashing crates depend on `commonware-codec` and the reverse dependency isn't possible. `Checksummer::Digest` is any `FixedSize` codec type, so implementations can range from a CRC to a cryptographic digest.

Details worth calling out for review:

- `Read::Cfg` forwards to `T::Cfg`, so wrapping a type doesn't change how it is configured. Config-driven failures (e.g. a `RangeCfg` rejection) surface before any checksum work.
- Verification re-encodes the decoded value instead of retaining the consumed bytes, which avoids buffering but assumes canonical encodings. Every implementation in this crate is canonical; for anything else, a non-canonical encoding that decodes to the same value is rejected as a mismatch. This is documented on the type.
- The `PhantomData<fn() -> C>` marker keeps the wrapper's auto traits independent of `C`.
- Docs note explicitly that a checksum is not authentication.

## Testing

- Unit tests cover round trips for fixed- and variable-size inner types, config forwarding, `encode_size` agreement with the written bytes, corrupted payload and corrupted checksum rejection, a missing trailing checksum, and that `read` consumes exactly one value from a buffer holding more.
- Doctests on both `Checksummer` and `Checksummed`.
- `just pre-pr` passes, including the `no_std` build (the type adds no `std` usage).

### DIFF
--- a/codec/src/types/checksummed.rs
+++ b/codec/src/types/checksummed.rs
@@ -8,9 +8,6 @@ const CHECKSUMMED_TYPE: &str = "Checksummed";
 
 /// Computes a fixed-size checksum over a slice of bytes.
 ///
-/// [`Checksummed`] is generic over this trait rather than over a hasher from a dedicated hashing
-/// crate because those crates depend on this one.
-///
 /// # Example
 ///
 /// ```
@@ -27,6 +24,8 @@ const CHECKSUMMED_TYPE: &str = "Checksummed";
 ///     }
 /// }
 /// ```
+// Defined here instead of reusing a hasher trait from a dedicated hashing crate: those crates
+// depend on this one.
 pub trait Checksummer: Send + Sync + 'static {
     /// The checksum written after the encoded value.
     type Digest: Read<Cfg = ()> + Write + FixedSize + PartialEq;
--- a/codec/src/types/checksummed.rs
+++ b/codec/src/types/checksummed.rs
@@ -1,0 +1,266 @@
+//! This module exports the [`Checksummed`] type.
+
+use crate::{Encode, EncodeSize, Error, FixedSize, Read, Write};
+use bytes::{Buf, BufMut};
+use core::{fmt::Debug, marker::PhantomData};
+
+const CHECKSUMMED_TYPE: &str = "Checksummed";
+
+/// Computes a fixed-size checksum over a slice of bytes.
+///
+/// [`Checksummed`] is generic over this trait rather than over a hasher from a dedicated hashing
+/// crate because those crates depend on this one.
+///
+/// # Example
+///
+/// ```
+/// use commonware_codec::types::checksummed::Checksummer;
+///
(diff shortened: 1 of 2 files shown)


crazywriter1:
Friendly ping. This is ready for review when you have a moment. Happy to address any feedback. @BrendanChou @patrick-ogrady

PR #4488, opened 2026-08-14
[runtime/iobuf/freelist] avoid scratch allocation on empty batch probes

patrick-ogrady:
Depends on #4485.

## Problem

With TLS caching enabled, a local miss probes the global freelist with a bounded `take_batch`. The striped implementation eagerly reserved space for `max` slot IDs before checking any availability hint. When the global freelist was stably empty during lazy growth or after exhaustion, every miss allocated and freed unused scratch before creating a buffer or returning `PoolError::Exhausted`.

## Change

Batch scratch is now reserved only after a positive availability hint, while still before acquiring the stripe mutex. Stable-empty probes therefore remain allocation-free without moving allocator work into the critical section.

The empty `Stripe::pop` path also no longer repeats `available = false`. A taker can carry a stale local `true` into the lock, but every supported transition that empties the authoritative vector already stores shared `false` while holding that same lock. The previous test manufactured a shared `true` and empty-vector state that production cannot leave behind.

A thread-local allocation-counting regression exercises the real exhausted TLS batch path and asserts zero allocator calls.

## Testing

- `just test -p commonware-runtime` (759 passed)
- `just clippy -p commonware-runtime`
- `just check-fmt`
- targeted iobuf freelist and pool Loom suite (10 passed)
- focused Miri regression

### DIFF
--- a/runtime/src/iobuf/freelist.rs
+++ b/runtime/src/iobuf/freelist.rs
@@ -162,12 +162,9 @@ impl Stripe {
     /// `free` must be the guard obtained from [`Self::lock`] for this stripe.
     #[inline(always)]
     fn pop(&self, free: &mut MutexGuard<'_, Vec<u32>>) -> Option<u32> {
-        let Some(slot) = free.pop() else {
-            // A taker can load `true` before another taker removes the final
-            // slot. Reconfirm the now-authoritative empty state for later scans.
-            self.available.store(false, Ordering::Release);
-            return None;
-        };
+        // A taker can load `true` before another taker removes the final slot,
+        // so the locked vector may already be empty.
+        let slot = free.pop()?;
 
         if free.is_empty() {
             // Publish the last removal before releasing the stripe lock.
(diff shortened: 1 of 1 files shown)


runtime/src/iobuf/freelist.rs
```diff
@@ -1119,16 +1124,94 @@ pub(super) mod tests {
...
+        let result = f();
+        let calls = ALLOCATION_CALLS.with(|calls| calls.replace(None).expect("tracking enabled"));
+        (result, calls)
+    }
+
+    /// Verifies an exhausted TLS batch probe does not allocate result scratch.
     #[test]
-    fn test_stale_positive_hint_is_reconfirmed() {
-        let set = lazy(1, 1);
+    fn test_exhausted_batched_tls_miss_is_allocation_free() {
```
patrick-ogrady, line 1182: I think this is probably overkill (and we could remove all these test changes). But your call.

andresilva (approved):

PR #4490, opened 2026-08-14
[runtime] Add atomic blob storage

patrick-ogrady:
## Summary

Adds the cohesive BETA atomic blob storage contract and engine to the runtime. Atomic blobs are opt-in, append-only blobs whose payload is written once at its final offset. They support crash-atomic publication of appends, rewinds, application tags, integrity units, and exact removals, either for one blob or as a coordinator-free multi-blob group.

This is the bottom layer of the six-PR atomic blob stack and depends on the crash-faithful substrate in #4501. The remaining backend/runtime adapters follow in #4502, and explicit ordinary-to-atomic migration follows in #4495. These are separate review units but one atomic-storage release unit: merge them in order and do not release the engine by itself.

The BETA surface in this layer includes:

- `AtomicStorage::{open_atomic, scan_atomic}`;
- `AtomicBlob` append, rewind, tag, integrity, `start_sync`, and `sync` operations; and
- `AtomicStorage::{start_apply, apply}` for atomic publish, rewind, and remove groups.

The layer includes the two minimal raw storage adapters required by the engine itself: Memory consumes the inline driver and Tokio storage consumes the cross-platform background driver. Deterministic contexts, wrappers, Tokio runtime forwarding, and Linux io_uring remain isolated in #4502.

## Storage and identity

Atomic storage uses the existing canonical V1 container. Atomic identification starts at logical offset zero with a 4 KiB `CWUNOID` incarnation page and two alternating 2 KiB R15 root slots; payload begins at logical offset 8192.

A new identity is installed only after ordinary `Storage` reports the name absent and creates a fresh zero-length backing. Existing backings must contain a complete identity; incomplete identities and ordinary payloads are rejected without mutation. Scans never create names. The core root grammar contains only `BatchPrepared` and `Finalized`; the migration-only committed spelling lands with its producer in #4495.

## Publication and recovery

A direct publication is encoded as a singleton witness ring. Group publication writes one root-bound successor witness beside every candidate root, and the complete durable closed ring is the decision. Final-root materialization, tombstones, truncation, and physical namespace cleanup are recovery debt rather than a second decision round.

Recovery validates exact membership, canonical order, incarnations, generations, successor closure, root transitions, and bounded payload proofs. It selects either the complete predecessor vector or the complete candidate vector and never invents or contracts a ring. Foreground and detached work share the same owned driver future after admission, so dropping an admitted observer does not cancel the mutation.

## Review focus

- Direct and grouped publication intentionally stay together because they share the witness encoding and generic recovery path.
- The public API, batch grammar, state transitions, recovery, and apply path land as one final mechanism without a temporary trait or format.
- Memory and raw Tokio storage keep both final driver modes live at this compile boundary.
- `AtomicResources` explicitly clones every shared driver, lock, and accounting resource, preserving one lineage while keeping the WASM boundary warning-free.
- Remaining adapters and all migration mechanisms are deliberately absent and isolated in #4502 and #4495.

## Diff size

- Production code: `+5645 / -11 LOC`
- Tests and test infrastructure: `+716 / -3 LOC`
- Total: `+6361 / -14 LOC`

Counts are physical diff lines, including comments and blank lines. Test-only files, hooks, and lines inside `#[cfg(test)]` modules are counted as tests; everything else is counted as production code.

### DIFF
--- a/runtime/src/atomic/batch.rs
+++ b/runtime/src/atomic/batch.rs
@@ -1,0 +1,809 @@
+//! Pure witness encoding, candidate binding, and transition classification.
+
+use super::*;
+
+const WITNESS_BINDING_DOMAIN: &[u8] = b"_COMMONWARE_RUNTIME_ATOMIC_BATCH_BINDING";
+const WITNESS_HEADER_LEN: usize = 4;
+const LINK_MAGIC: &[u8; 8] = b"CWUNOL15";
+pub(super) const GROUP_ID_LEN: usize = 16;
+const LINK_REMOVED: u32 = 1;
+const PAYLOAD_DESCRIPTOR_LEN: usize = 40;
+const LINK_FIXED_LEN: usize = 8
+    + GROUP_ID_LEN
+    + 4
+    + INCARNATION_LEN
+    + ROOT_LEN
+    + PAYLOAD_DESCRIPTOR_LEN
+    + 4
+    + 4
+    + INCARNATION_LEN;
+pub(super) const MAX_LINK_LEN: usize = ROOT_SLOT_SIZE - ROOT_LEN - WITNESS_HEADER_LEN;
+pub(super) const MAX_LOCATION_LEN: usize = 1_640;
+const _: () = assert!(MAX_LOCATION_LEN <= MAX_LINK_LEN - LINK_FIXED_LEN);
+
+#[derive(Clone, Copy, Debug, Eq, Ord, PartialEq, PartialOrd)]
+pub(super) struct PayloadDescriptor {
+    pub(super) start: u64,
+    pub(super) checksum: PayloadDigest,
+}
+
+impl PayloadDescriptor {
+    pub(super) const fn empty(logical_len: u64) -> Self {
+        Self {
+            start: logical_len,
+            checksum: [0; 32],
+        }
+    }
+
+    fn decode(encoded: &[u8], logical_len: u64) -> io::Result<Self> {
+        let descriptor = Self {
+            start: u64::from_be_bytes(encoded[..8].try_into().unwrap()),
+            checksum: encoded[8..].try_into().unwrap(),
+        };
+        descriptor.validate(logical_len)?;
+        Ok(descriptor)
+    }
+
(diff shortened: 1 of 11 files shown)


runtime/src/atomic/batch.rs
```diff
@@ -0,0 +1,809 @@
...
+
+use super::*;
+
+const WITNESS_BINDING_DOMAIN: &[u8] = b"_COMMONWARE_RUNTIME_ATOMIC_BATCH_BINDING";
+const WITNESS_HEADER_LEN: usize = 4;
+const LINK_MAGIC: &[u8; 8] = b"CWUNOL15";
+pub(super) const GROUP_ID_LEN: usize = 16;
+const LINK_REMOVED: u32 = 1;
+const PAYLOAD_DESCRIPTOR_LEN: usize = 40;
+const LINK_FIXED_LEN: usize = 8
```
patrick-ogrady, line 11: lol...this needs to be improved

patrick-ogrady:
The atomic metadata fits in the first 4 KiB **logical application page**:

```text
logical offset     0       32                  2064                4096
                   +-------+-------------------+-------------------+---------+
atomic contents    | ident | root 0 (2032 B)   | root 1 (2032 B)   | payload |
                   +-------+-------------------+-------------------+---------+
```

The identity encoding is currently 28 bytes, so it can be canonically padded to 32 bytes. The existing public maximum still fits in each 2,032-byte root slot:

```text
root header             132
witness length frame      4
fixed link              240
max partition + name   1640
                       ----
                       2016 bytes
```

That leaves 16 zero bytes of slack in each slot. Concretely, this would make the root offsets `[32, 2064]` and `DATA_OFFSET = 4096`, without reducing the 1,640-byte location limit.

Creation can resize to 4 KiB, write the canonical 32-byte identity, and sync once. A crash that leaves a complete identity but a short metadata extent can still be repaired by extending and syncing to 4 KiB; a torn/invalid identity remains corruption. Migration can build the same single-page prefix with its committed root. Root publication keeps the same alternating writes and durability barriers, so steady-state performance should be equivalent. Relative to the current layout, this removes one 4 KiB page per atomic blob.

PR #4497, opened 2026-08-14
[glue] Add snapshot publication and read-gate primitives

danlaine:
## Summary

Two self-contained modules with no consumers yet. They are consumed in #4498 and #4417.

## `stateful/db/live.rs` — own the database, hand out readers

`Writer::new(db)` takes ownership of a database. The writer is the only value that can mutate it, and it creates cloneable `Reader`s that can only read it. Both go through one fair, write-preferring read-write lock, so a waiting mutation cannot be starved and is delayed by at most the one storage call a read guard is allowed to cover.

`mutate` consumes the writer and returns it alongside the mutation's result. The mutation takes the database by value and produces it back, which is the contract mutable storage operations already use. Interrupting a mutation, by dropping its future or panicking out of it, poisons the lock and destroys the writer with it, so a second mutation of a poisoned writer cannot even be written. Dropping the writer closes it, so readers cannot silently go on serving a database that can never advance.

Both ways of losing the writer only happen while its owner is going down, so a `Reader` read of a dropped or poisoned writer returns `Closed`. Callers must treat it as an instruction to stop, never as a data-level failure. The writer's own reads are infallible, because a live writer proves a live database.

`Writer`, `Reader`, and `Closed` live in the public module `stateful::db::live`, parallel to `stateful::db::snapshot`: `live::Reader` reads the advancing database, `snapshot::Reader` reads published captures.

## `stateful/db/snapshot.rs` — publish the latest durable snapshots

`Publisher::new` returns a `Publisher` and a cloneable `Reader` over a shared cell holding the latest published snapshots, one per database, captured together at one height.

The problem it solves: snapshots are captured as blocks apply, but their flushes finish in any order, and snapshots must not be served before everything in them is on disk. So snapshots are staged by height, and `complete(height)` records a durable flush and publishes the newest staged snapshots whose flush, and every earlier one, has finished. If one completion makes several heights publishable at once, only the newest publishes and the rest are dropped without ever serving. Published heights never regress (asserted).

Snapshots captured before a prune still pin the pruned storage, so `mark_stale` records the highest height captured so far, staged or already served. The staleness clears once snapshots captured after the prune publish, and `needs_refresh` tells the owner it must capture afresh because the served snapshots predate the prune and nothing staged remains to replace them. `blocks_prune` reports whether a flush at or below a target is still pending.

`publish_now` covers state that is already durable (recovered at startup, synced at handoff, captured after every flush drained). Dropping the publisher closes the cell, so readers decline rather than serve unboundedly stale state; snapshots already handed out keep working. `Reader::view` derives a reader for one member of a multi-database capture.


danlaine:
Superseded: the read gates moved into storage as #4531 and the publication primitives now ride inside #4533.

PR #4498, opened 2026-08-14
[glue] Serve peers from published snapshot captures

danlaine:
## Summary

Peer serving previously read through `Shared<DB>`, the same lock mutations run under. That cost us twice: a slow serve could delay finalization, and a serve could observe applied state that was not yet on disk. Serving wants the opposite of what a live database gives: shared reads of state that is guaranteed durable.

So serving now gets its own data: immutable snapshots, published only once they are durable. A mutation never waits on a serve, and a peer is never shown state a crash would take back. `Shared`'s ownership model, `DatabaseSet: Clone`, and the verification machinery are untouched, and mutation still runs under the existing write slots. The ownership rework comes in #4417.

Also a bug fix the new paths exposed: a batch that never set an inactivity floor merkleized with the `Default` floor of zero, so once anything raised the database's floor, the next batch that left it unset failed with `FloorRegressed`. Only tests set the floor explicitly, which kept this latent. New batches now inherit the floor of whatever they fork from: the database's current floor, or the parent batch's.

## Design

**Capture.** The right moment to capture is `finalize`: the batch has just applied and the write slot is already held, so the snapshot adds no locking. `finalize` now returns the database, its snapshot, and the flush handle. A standalone `snapshot()` covers the moments outside an apply: startup, the state-sync handoff, and the post-prune refresh. `DatabaseSet` mirrors both.

**Publication.** A snapshot reflects everything applied before it, including state whose flush is still running, so capture and serving must be separated by durability. The `Publisher` is that gap. It replaces the loop's pending-flush bookkeeping one for one: snapshots are staged at finalize, flush completions are recorded where the loop previously ticked a `BTreeSet` of pending heights, and pruning gates on the same question it asked before. A snapshot publishes once its own flush and every earlier one has finished, and when one completion unblocks several, only the newest publishes. Readers never move backward and never see unflushed state. Marshal acknowledgement rides the same completions, unchanged.

**Startup and state sync.** A node should serve as soon as it holds anything durable, not after its next finalization. Recovered state publishes before the processing loop starts, synced state publishes at the handoff, and each handoff block publishes after its inline durability wait.

**Pruning.** A prune waits for every flush at or below its target, as before. The served snapshot was captured before the prune, so it pins the pruned storage. The prune marks it stale, and the space frees when a snapshot captured after the prune replaces it: either the next block's own publish, or, when nothing is staged, a fresh snapshot taken once the mailbox is idle (`Step::Refresh`).

**Serving.** The p2p resolver takes a snapshot reader at construction and answers every request from the latest published snapshot. A request that finds nothing published declines cleanly. That leaves the attach-database machinery (`AttachableResolver`, `AttachableResolverSet`, the `AttachDatabase` message, the `has_database` metric) with no job, so it is deleted. Resolvers keep their state-sync fetch role unchanged.

## API changes

All ALPHA, so breaking is fine.

- `ManagedDb` gains `type Snapshot` and `snapshot()`; `finalize` returns `(Self, Snapshot, Handle<()>)`. `DatabaseSet` mirrors both: `type Snapshots`, `snapshot()`, and `finalize` returning `(Snapshots, Barrier)`. `LogSnapshot` names the `Arc`-shared snapshot every log-backed QMDB variant produces.
- `Stateful`'s `Config` takes the `Publisher`; the resolver takes a `Reader` at construction. Publication metrics register under the context passed to `Publisher::new`.
- The `R` bound on `Stateful` relaxes from `AttachableResolverSet` to plain `Send + Sync + 'static`: resolvers are now only sync-fetch sources to the actor.

### DIFF
--- a/examples/reshare/src/validator.rs
+++ b/examples/reshare/src/validator.rs
@@ -31,7 +31,7 @@ use commonware_glue::{
     },
     stateful::{
         Config as StatefulConfig, Stateful, SyncPlan,
-        db::{DatabaseSet, p2p as qmdb_resolver},
+        db::{DatabaseSet, p2p as qmdb_resolver, snapshot::Publisher},
     },
 };
 use commonware_macros::boxed;
@@ -230,12 +230,13 @@ pub async fn run(context: tokio::Context, args: Validator) {
     )
     .await;
 
+    let publication_context = context.child("publication");
+    let (snapshot_publisher, snapshot_reader) = Publisher::new(&publication_context);
     let (qmdb_actor, qmdb_sync_resolver) = qmdb_resolver::Actor::new(
         context.child("qmdb_resolver"),
         qmdb_resolver::Config {
             peer_provider: oracle.clone(),
             blocker: oracle.clone(),
-            database: None,
             mailbox_size: MAILBOX_SIZE,
             me: Some(local.clone()),
             initial: Duration::from_secs(1),
@@ -245,6 +246,7 @@ pub async fn run(context: tokio::Context, args: Validator) {
             priority_requests: false,
             priority_responses: false,
         },
+        snapshot_reader,
     );
     let qmdb_handle = qmdb_actor.start(qmdb_network);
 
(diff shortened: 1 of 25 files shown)


danlaine:
Superseded by #4533, the same feature adapted to the snapshot-capture publisher.

PR #4501, opened 2026-08-14
[runtime] Add crash-faithful storage and runtime shutdown

patrick-ogrady:
## Summary

Adds the non-atomic crash, durability, and runtime substrate required by atomic blob storage. The layer makes deterministic crashes faithful to pending storage mutations, gives in-memory storage generation-aware durable snapshots, and makes the Tokio runner own and drain spawned tasks during shutdown.

This is a standalone prerequisite targeting `main`, which already contains #4489. It contains no atomic storage API or engine code; the cohesive engine remains in the separate stack beginning with #4490.

The layer owns:

- grouped write and resize fault configuration, with independent operation-failure and crash-retention rates;
- ordered faulty-storage journals for writes, resizes, full-sync durability cuts, and unlinks, with one prefix/subset byte-retention policy shared by failed writes and successful unsynchronized writes;
- generation-aware in-memory snapshots, durable ranges, stale-handle isolation, and crash reconstruction;
- deterministic runtime checkpoint/recovery of storage state, fault configuration, and RNG state; and
- Tokio task admission, ownership, cancellation, and shutdown draining.

## Review focus

- Failure and crash-retention rates remain independent even though each operation groups them in one configuration tuple.
- `start_sync` returns after initiation; its completion handle observes durability without owning or canceling the backing work, and completed cuts retire only earlier debt from the same file generation.
- Crash replay preserves mutation issue order and never exposes a state that the configured retention mode could not produce.
- Replacing or unlinking a generation cannot let an escaped stale handle mutate the new generation.
- Deterministic checkpoints restore storage and fault state together.
- The Tokio runner retains ownership until spawned tasks have been canceled and drained.

## Diff size

| Kind | Added | Removed |
|---|---:|---:|
| Production code | `834` | `101` |
| Tests and test infrastructure | `1,313` | `65` |
| Total | `2,147` | `166` |

Counts are physical diff lines, including comments and blank lines. Test-only files, hooks, and lines inside `#[cfg(test)]` modules are counted as tests; everything else is counted as production code.

### DIFF
--- a/p2p/fuzz/fuzz_targets/simulated.rs
+++ b/p2p/fuzz/fuzz_targets/simulated.rs
@@ -7,7 +7,7 @@ use commonware_p2p::{
     Channel, Receiver as ReceiverTrait, Recipients, Sender as SenderTrait, simulated,
 };
 use commonware_runtime::{Clock, IoBuf, Quota, Runner, Supervisor as _, deterministic};
-use commonware_utils::NZUsize;
+use commonware_utils::{NZUsize, Probability};
 use libfuzzer_sys::fuzz_target;
 use rand::RngExt as _;
 use std::{
@@ -58,7 +58,7 @@ enum Operation {
         latency_ms: u16,
         /// Latency jitter in milliseconds.
         jitter: u16,
-        /// Success rate (0-255 maps to 0.0-1.0).
+        /// Success rate as a numerator over 255.
         success_rate: u8,
     },
     /// Remove a network link between two peers.
@@ -279,12 +279,11 @@ fn fuzz(input: FuzzInput) {
                     let from_idx = (from_idx as usize) % peer_pks.len();
                     let to_idx = (to_idx as usize) % peer_pks.len();
 
-                    // Create link with specified characteristics
-                    // success_rate is normalized from u8 (0-255) to f64 (0.0-1.0)
+                    // Create link with specified characteristics.
                     let link = simulated::Link {
                         latency: Duration::from_millis(latency_ms as u64),
                         jitter: Duration::from_millis(jitter as u64),
-                        success_rate: (success_rate as f64) / 255.0,
+                        success_rate: Probability!(u64::from(success_rate), u64::from(u8::MAX)),
                     };
                     let _ = oracle
                         .add_link(peer_pks[from_idx].clone(), peer_pks[to_idx].clone(), link)
--- a/p2p/fuzz/fuzz_targets/simulated.rs
+++ b/p2p/fuzz/fuzz_targets/simulated.rs
@@ -58,8 +58,8 @@ enum Operation {
         latency_ms: u16,
         /// Latency jitter in milliseconds.
         jitter: u16,
-        /// Success rate as a numerator over 255.
-        success_rate: u8,
+        /// Probability that a message is delivered successfully.
+        success_rate: Probability,
     },
     /// Remove a network link between two peers.
     RemoveLink {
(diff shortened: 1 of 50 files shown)


runtime/src/deterministic.rs
```diff
@@ -1883,6 +1914,91 @@ mod tests {
...
+
+        deterministic::Runner::from(checkpoint).start(|context| async move {
+            assert!(context.open("fault_config", b"blob").await.is_err());
+        });
+    }
+
+    #[test]
+    fn test_recover_retained_successful_resize() {
+        let cfg = deterministic::Config::default()
+            .with_seed(83)
```
danlaine: This test seems to rely on using seed 83 to get specific roll results. Seems fragile?

patrick-ogrady: yeah, not sure what the story is there lol.

runtime/src/storage/faulty.rs
```diff
@@ -424,15 +959,713 @@ mod tests {
...
+                assert_eq!(&durable.as_ref()[5..], b"...");
+            }
+        }
+        assert!(exercised, "seed sweep must exercise a partial prefix");
+    }
+
+    #[tokio::test]
+    async fn test_crash_replays_writes_and_resizes_in_issue_order() {
+        let h = Harness::with_seed(
+            83,
```
danlaine: another non-obvious hard-coded seed

runtime/src/storage/faulty.rs
```diff
@@ -424,15 +959,713 @@ mod tests {
...
+        assert_eq!(first.read_at(0, 5).await.unwrap().coalesce(), b"abc\0\0");
+        let (second, len) = h.inner.open("partition", b"second").await.unwrap();
+        assert_eq!(len, 6);
+        assert_eq!(second.read_at(0, 6).await.unwrap().coalesce(), b"abc\0\0X");
+    }
+
+    #[tokio::test]
+    async fn test_partial_sync_write_replays_after_retained_resize() {
+        let h = Harness::with_seed(
+            83,
```
danlaine: another non-obvious hard-coded seed

utils/src/probability.rs
```diff
@@ -0,0 +1,323 @@
...
+    "probability must be finite, within [0, 1], and exactly representable as a 64-bit threshold"
+)]
+pub struct InvalidProbability;
+
+/// A probability represented as a threshold over all possible `u64` samples.
+///
+/// Ratios are rounded down to the nearest multiple of 2^-64. Sampling consumes one `u64` for
+/// probabilities strictly between zero and one, and consumes no randomness for either endpoint.
+/// Given the same sequence of `u64` samples, decisions are identical on every platform.
+#[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
```
danlaine: nit: the Debug will print the raw u64 instead of, like, `0.67`.

runtime/src/storage/faulty.rs
patrick-ogrady: not sure what poisoned the training corpus with 83 lol.

p2p/fuzz/fuzz_targets/simulated.rs
```diff
@@ -58,7 +58,7 @@ enum Operation {
         latency_ms: u16,
         /// Latency jitter in milliseconds.
         jitter: u16,
-        /// Success rate (0-255 maps to 0.0-1.0).
+        /// Success rate as a numerator over 255.
         success_rate: u8,
```
danlaine:
```suggestion
        success_rate: Probability,
```

runtime/src/storage/faulty.rs
```diff
@@ -2,16 +2,63 @@
...
     io::Error as IoError,
     sync::{
-        Arc,
+        Arc, OnceLock, Weak,
         atomic::{AtomicU64, Ordering},
     },
 };
 
+/// A bounded RNG for tests that require explicit probability outcomes.
+#[cfg(test)]
```
danlaine: nit: move these to `mod tests`?

PR #4504, opened 2026-08-14
[runtime/iobuf/freelist] Avoid replacement allocations on final drop

patrick-ogrady:
## Summary

- deallocate parked stripe contents directly during terminal `Freelist::drop`, avoiding replacement vectors after all checked-out and TLS-cached buffers have returned
- retain `drain()` as the concurrent pool-teardown path, where late returns still need reserved stripe capacity, and share detached-buffer deallocation between both paths
- add a populated final-drop allocation regression and replace four scheduler-sensitive timeout tests with causal Loom coverage for scalar and batch `put` and `take`

LOC against `andre/freelist-striped-mutex-dumb`: production `+36/-17`; tests/test infrastructure `+145/-160`. Lines before the first test module count as production; native and Loom test modules count as tests/test infrastructure.

### DIFF
--- a/runtime/src/iobuf/freelist.rs
+++ b/runtime/src/iobuf/freelist.rs
@@ -521,15 +521,13 @@ impl Freelist {
 
     /// Deallocates every buffer currently parked in this freelist.
     ///
-    /// Each stripe receives a preallocated empty vector while locked. Detached
-    /// buffers are deallocated after unlocking. Creation permits remain
-    /// consumed, so a later return can reuse the reserved room but draining
-    /// never permits a second allocation for an existing slot.
+    /// Checked-out and TLS-cached buffers keep this size class alive and may
+    /// return after a drain. Each stripe therefore receives a preallocated
+    /// empty vector while locked. Detached buffers are deallocated after
+    /// unlocking. Creation permits remain consumed, so draining never permits
+    /// a second allocation for an existing slot.
     pub fn drain(&self) -> usize {
         let mut drained = 0;
-
-        // Visit stripes directly so Drop can run during thread-local cache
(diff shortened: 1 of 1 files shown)


runtime/src/iobuf/freelist.rs
```diff
@@ -601,9 +599,27 @@ impl Freelist {
...
 impl Drop for Freelist {
     fn drop(&mut self) {
-        // Outstanding slots keep the size class alive, so only globally parked
-        // buffers can remain when the final freelist reference is dropped.
-        self.drain();
+        // Checked-out and TLS-cached slots keep the size class alive until they
+        // return. Final destruction therefore owns every live allocation, while
+        // `drain` must preserve reserved stripe storage for returns that can race
+        // pool teardown. Exclusive access permits reclaiming each stripe in place.
+        for index in 0..self.stripes.len() {
```
patrick-ogrady, line 613: trying to dedup this one sec

runtime/src/iobuf/freelist.rs
patrick-ogrady, line 537: LLM wanted to avoid all these allocations on shutdown...which made me wonder what `drain` is used for outside of drop to begin with.

runtime/src/iobuf/freelist.rs
```diff
@@ -912,71 +849,6 @@ pub(super) mod tests {
         set.put(taken.expect("singleton callback receives the buffer"));
     }
 
-    /// Verifies that `take` waits for its positive home before later stripes.
-    #[test]
-    fn test_take_waits_for_home_before_later_stripe() {
```
patrick-ogrady, line 917: seemed more interesting to convert these to loom tests

PR #4510, opened 2026-08-15
[parallel] Pin spawned jobs to their submitter's LLC domain

### LINKED ISSUE #4540: [runtime] consider thread affinity in thread spawning
Currently threads are spawned without regard to affinity, which can result in ability to share (for example) L3 cache.  This leads to high variance in some concurrency related benchmarks and can hurt performance overall.

roberto-bayardo:
## What

Spawned jobs now pin their executing thread to the last level cache (L3) domain
the submitter occupied at spawn time, restoring the previous affinity mask when
the job completes (including unwinds), unless the mask was rewritten mid-job
to CPUs outside the pin. Domains are discovered once per process from the
kernel's cache topology: for each CPU, the `shared_cpu_list` of the
highest-level Data or Unified leaf under its `cache/index*` directory in
sysfs (the leaf index varies by architecture, so it is discovered rather than
assumed). Non-Linux platforms and miri compile the mechanism to a no-op.

All libc FFI funnels through a small safe `CpuSet` wrapper (seven one-line
unsafe blocks); the pin protocol itself is safe code.

Closes #4540.

## Why

A spawned job usually exchanges data with its submitter: it reads what the
caller just built and the caller later reads what it wrote. If the executor
runs in a different L3 domain than the submitter, every shared cache line pays
a cross-domain transfer for the job's lifetime. The OS places threads once per
process, effectively at random, and never revisits the choice.

The effect is easiest to see on a benchmark whose commit path hands index
application to a single spawned job (`constantinople`, kind
`any::unordered::fixed::mmb`: 1M keys, 32k reads and 32k updates per iteration,
p50 commit latency over 100 iterations, 8 threads). Machine: AMD EPYC 9354P,
32 cores, SMT off, eight 4-core L3 domains per the kernel.

Eight consecutive runs of the same binary on the same input, before this
change:

| run | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 |
|---|---|---|---|---|---|---|---|---|
| p50 (ms) | 26.6 | 26.6 | 18.0 | 27.2 | 18.3 | 27.2 | 19.0 | 26.8 |

Each process lands in one of two modes and stays there for its lifetime. The
mode is set by where the scheduler seated the job's executor relative to the
submitting thread. Hardware counters show the slow mode is pure memory
stalling, not extra work:

| mode | instructions | cycles | IPC |
|---|---|---|---|
| fast (p50 ~18.5) | 31.5B | 31.0B to 33.4B | ~1.00 |
| slow (p50 ~26.9) | 31.6B | 34.5B to 36.8B | ~0.88 |

The same eight-run experiment with this change:

| run | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 |
|---|---|---|---|---|---|---|---|---|
| p50 (ms) | 15.2 | 18.1 | 19.2 | 17.1 | 16.4 | 16.9 | 19.3 | 15.4 |

The lottery is gone, and the median (16.9ms) is better than the old lucky mode
because the pin also keeps the executor from drifting away mid-run. For
reference, the same benchmark without the spawned index job at all runs at
23.5ms, and single-threaded at 26.7ms.

## Cost

The first pin migrates the worker once, costing tens of microseconds.
Afterwards the restored-wide worker tends to stay where it ran, so repeat pins
move nothing. The spawn round-trip microbenchmark measures no change (82.5us
with the pin versus 83.7us without, within run-to-run noise). Jobs below the
size threshold are unaffected: the adaptive spawn policy (#4423) keeps
them inline, so the pin only executes for jobs large enough that a one-time
migration is noise. Concurrent pinned jobs cannot crowd a domain: the cap on live
pins is the domain's effective width (the CPUs the pinning thread may actually
use there) minus one, reserving one for the submitting caller. Past the cap, or
when fewer than two CPUs are allowed, a job simply runs unpinned.
A worker already pinned by an enclosing job does not pin again: a thread-local
marks the thread as pinned, so nested jobs stay within the outer confinement
instead of burning a second cap slot for one CPU of occupancy. Machines whose
possible-CPU count exceeds `cpu_set_t`'s capacity report no domains at all,
disabling the feature up front rather than charging every spawn for affinity
syscalls that could never engage.

Restore defers to external rewrites: a mask moved outside the pin mid-job
(taskset -pa, a cgroup cpuset migration) is left in place, since writing the
pin-time snapshot back would resurrect an allowance the operator may have just
revoked. Cleanup is best effort: the pin slot is always released, so the worst
case may leave a worker narrowed with its slot freed rather than stranding
domain capacity for the process lifetime.

## Benchmarks at tip

With #4423 merged, only policy-approved offloads ever execute the pin, so the
change was re-measured at the current tip against main (AMD EPYC 9354P, eth
mainnet replay in follower mode, 8 threads, two interleaved runs per arm,
means shown):

| window | metric | main | this PR |
|---|---|---|---|
| 500k blocks, fresh db | wall | 13.66s | 13.14s (-3.8%) |
| | writes/s | 126.6k | 131.5k (+3.9%) |
| 50k blocks onto a 30GB db | wall | 123.25s | 114.82s (-6.8%) |
| | merkleize | 73.78s | 67.50s (-8.5%) |
| | apply | 18.64s | 17.11s (-8.2%) |
| | writes/s | 175.0k | 189.5k (+8.3%) |

Single-threaded runs are unchanged: a single-worker pool runs every job inline,
so the pin never engages. The deeper-state window benefits most because its
offloaded merkleize and apply jobs are the largest and longest-lived, exactly
the traffic that pays for cross-domain lines.

## Testing

Tests drive the full pin protocol (intersection, cap accounting, apply, and
restore) against synthetic two-domain topologies built by splitting the
process's allowed CPUs in half, with private pin counters per test. This makes
the coverage real on CI runners, which expose a single L3 domain and would
otherwise skip every pin path. Topology parsing is likewise exercised against
synthetic sysfs cache directories, SMT-style discontiguous sharing lists, and
CPUs past `cpu_set_t`'s capacity.

https://claude.ai/code/session_01Q4BFdANo2LhVEr2ko3FK9n

### DIFF
--- a/parallel/src/topology.rs
+++ b/parallel/src/topology.rs
@@ -1,0 +1,255 @@
+//! Last-level-cache topology and executor affinity for spawned jobs.
+//!
+//! Threads that exchange cache lines pay far more when they sit in different last-level
+//! cache (L3) domains: on multi-CCD parts a cross-domain transfer costs several times a
+//! shared-domain one, and the scheduler's thread placement is a sticky per-process
+//! lottery it never revisits. A spawned job usually has a data relationship with its
+//! submitter (it consumes what the caller just built, and the caller consumes what it
+//! produces), so the job's executor pins itself to the submitter's domain for the job's
+//! duration via [`AffinityGuard`]: one migration the first time (tens of microseconds,
+//! against per-line fabric traffic for the job's lifetime otherwise), and a no-op
+//! afterwards because the restored-wide executor tends to stay where it ran.
+//!
+//! Domains come from the kernel's cache topology (never from vendor core-layout
+//! assumptions, which lie: this was built on a machine whose spec sheet implied 8-core L3
+//! groups while the kernel reported 4-core ones). On non-Linux platforms the stub below
+//! reports a single domain and pinning is a no-op; the same stub serves miri (the
+//! affinity syscalls are unshimmed).
+
+cfg_if::cfg_if! {
+    if #[cfg(all(target_os = "linux", not(miri)))] {
+        use std::sync::OnceLock;
+
+        /// The process-wide topology, detected on first use.
+        fn topology() -> &'static Topology {
+            static TOPOLOGY: OnceLock<Topology> = OnceLock::new();
+            TOPOLOGY.get_or_init(Topology::detect)
+        }
+
+        /// Map from CPU id to last-level-cache domain id, plus each domain's CPU set.
+        pub(crate) struct Topology {
+            /// `cpu_to_domain[cpu]` is the domain of `cpu`; CPUs beyond the table
+            /// (offline or hot-added) map to domain 0.
+            cpu_to_domain: Box<[u16]>,
+            /// The CPUs of each domain, as ready-to-use affinity sets.
+            domain_sets: Box<[libc::cpu_set_t]>,
+        }
+
+        impl Topology {
+            /// Detects the LLC topology; `None` when unreadable or single-domain.
+            fn detect() -> Self {
+                Self::from_sysfs().unwrap_or(Self {
+                    cpu_to_domain: Box::new([]),
+                    domain_sets: Box::new([]),
+                })
+            }
+
+            /// Number of distinct domains (0 or 1 disables the preference and pinning).
+            pub(crate) fn domains(&self) -> usize {
+                self.domain_sets.len()
+            }
+
+            /// The domain of `cpu`.
+            fn domain_of(&self, cpu: usize) -> u16 {
+                self.cpu_to_domain.get(cpu).copied().unwrap_or(0)
+            }
+
+            /// The domain of the CPU the calling thread is currently running on.
+            /// Advisory: the thread may migrate immediately after.
+            pub(crate) fn current_domain(&self) -> u16 {
+                // SAFETY: sched_getcpu takes no arguments and only returns a value; -1
+                // on error (mapped to domain 0 by the table fallback).
+                let cpu = unsafe { libc::sched_getcpu() };
+                if cpu >= 0 {
+                    return self.domain_of(cpu as usize);
+                }
+                0
+            }
+
+            /// Builds the table from
+            /// `/sys/devices/system/cpu/cpu*/cache/index3/shared_cpu_list`. Returns
+            /// `None` if nothing was readable or only one domain exists.
+            fn from_sysfs() -> Option<Self> {
+                let mut lists: Vec<(usize, String)> = Vec::new();
+                let entries = std::fs::read_dir("/sys/devices/system/cpu").ok()?;
+                for entry in entries.flatten() {
+                    let name = entry.file_name();
+                    let name = name.to_string_lossy();
+                    let Some(id) = name
+                        .strip_prefix("cpu")
+                        .and_then(|s| s.parse::<usize>().ok())
+                    else {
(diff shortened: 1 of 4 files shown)


parallel/src/topology.rs
```diff
@@ -0,0 +1,255 @@
...
+                let size = std::mem::size_of::<libc::cpu_set_t>();
+                // SAFETY: pid 0 is the calling thread; `saved` is a valid set of `size`.
+                if unsafe { libc::sched_getaffinity(0, size, &mut saved) } != 0 {
+                    return None;
+                }
+                // SAFETY: pid 0 is the calling thread; `target` is a valid set of `size`.
+                if unsafe { libc::sched_setaffinity(0, size, target) } != 0 {
+                    return None;
+                }
+                Some(Self { saved })
```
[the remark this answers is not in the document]
roberto-bayardo: Good catch, fixed in 55a9c8234. The pin now intersects the domain set with the thread's current affinity mask, so it narrows placement and never widens it past taskset, numactl, or isolcpus restrictions, and it is skipped entirely when the intersection is empty. Added a regression test that restricts a thread to a single CPU and verifies a pin grants it nothing. The spawn round trip microbenchmark is unchanged with the intersection in place (82.7us versus 82.5us before).

parallel/src/topology.rs
```diff
@@ -0,0 +1,378 @@
...
+            cpu_to_domain: Box<[Option<u16>]>,
+            /// The CPUs of each domain, as ready-to-use affinity sets.
+            domain_sets: Box<[libc::cpu_set_t]>,
+        }
+
+        impl Topology {
+            /// Detects the LLC topology; `None` when unreadable or single-domain.
+            fn detect() -> Self {
+                Self::from_sysfs().unwrap_or(Self {
+                    cpu_to_domain: Box::new([]),
```
patrick-ogrady, line 140: I guess my only fear is pinning somehow gets in the way of concurrency? but assume it is smarter than that.

roberto-bayardo: The pin is scoped pretty conservatively: it targets a whole L3 domain (4 cores here) rather than a core, it only applies to the one worker executing the spawned job, and the mask is restored the moment the job completes, so nothing accumulates across jobs.

The bigger safety net is that only policy approved offloads ever pin, and the spawn policy keeps measuring each callsite's offload wall clock. If pinning ever crowded a domain enough to hurt, that callsite's samples degrade and the policy steers it back to inline on its own.

And for all our larger-scale benchmarks for far (eth replay, db-gen, constantinople) we see no degradation.

roberto-bayardo: Let me see if I can add something to avoid over-pinning to the same domain should it ever arise...

roberto-bayardo: Done in a660e983e: each domain now accepts at most its width minus one live pins (a per domain counter the guard releases along with the mask), so concurrent pinned jobs can never crowd a domain or evict the caller working there. Past the cap a job just runs unpinned wherever the scheduler likes. That makes over pinning structurally impossible rather than something the spawn policy has to learn to avoid. Re-validated on the branch tip afterwards: eight consecutive benchmark runs all land in the fast band (14.3 to 19.9ms p50) with no slow mode draws.

PR #4512, opened 2026-08-15
[simplex] Revalidate locally built proposal ancestry

patrick-ogrady:
Follow-up to #4480.

Proposal construction is asynchronous. A parent captured when construction starts can become invalid before the application returns a payload. `State::proposed` records the completed proposal if its parent remains preferred or remains valid ancestry. Otherwise, the voter rejects the proposal before relay and permits a new build if the slot remains empty.

- Moves local ancestry validation into `State::proposed`.
- Retains the same-term optimistic vote check and removes the cross-term signing check made redundant by proposal validation.
- Requires a local `notarize` vote for the outgoing view before installing the next term’s leader early. Receiving the outgoing proposal alone does not enable the handoff.
- Renames `Elector::elect_ahead` to `Elector::elect_early`.
- Adds tests for late fallback validity, nullification, certification failure, equivocation, conflicting-certificate invalidation, and proposal retries.

### DIFF
--- a/consensus/src/simplex/elector.rs
+++ b/consensus/src/simplex/elector.rs
@@ -18,7 +18,7 @@
 //! Applications can implement [`Config`] and [`Elector`] for custom leader
 //! selection logic such as stake-weighted selection or other application-specific strategies.
 //! An elector can support pipelined handoffs when it can select a leader before the round's
-//! unlocking certificate exists. See [`Elector::speculate`].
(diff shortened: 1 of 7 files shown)


consensus/src/simplex/elector.rs
patrick-ogrady, line 367: nit: call this `optimistic`

PR #4526, opened 2026-08-19
fix(simplex): retry certificate assembly after quorum

### LINKED ISSUE #4409: simplex/batcher: certificate recovery panics on schemes where `assemble` legally returns `None`
## Summary

`certificate::Scheme::assemble` is documented to return `None` when the threshold is not met ([cryptography/src/certificate.rs#L379](https://github.com/commonwarexyz/monorepo/blob/54b92a76bbf5/cryptography/src/certificate.rs#L379)). The simplex batcher treats `None` at vote-count quorum as unreachable: recovery surrenders the verified votes once `quorum` votes exist (`quorum` = `participants().quorum::<N3f1>()`, a participant count — [batcher/round.rs#L66](https://github.com/commonwarexyz/monorepo/blob/54b92a76bbf5/consensus/src/simplex/actors/batcher/round.rs#L66)) and unwraps the result:

```rust
Notarization::from_owned_notarizes(scheme.as_ref(), notarizes, &strategy)
    .expect("verified notarize quorum must assemble")
```

([batcher/verifier.rs#L309-L360](https://github.com/commonwarexyz/monorepo/blob/54b92a76bbf5/consensus/src/simplex/actors/batcher/verifier.rs#L309-L360) — same for nullify and finalize.)

For a scheme whose signing weight is not one-unit-per-participant, that unwrap is reachable through vote arrival order alone.

## Reachability

Take a share-weighted threshold scheme: `S` shares allocated by stake, `sign` emits one partial per held share, `verify_attestation` checks each attestation against its signer's assigned indices, `assemble` recovers at a share threshold `quorum(S)`. Every attestation is individually valid; every current trait contract is satisfied.

Concrete numbers: 4 validators, stake split so shares are `[7, 1, 1, 1]` (S = 10, share threshold 7, participant quorum 3).

- The three 1-share validators' votes arrive at a node first: participant quorum is reached at 3 votes, recovery consumes them, `assemble` sees 3 shares < 7 and returns `None`, the batcher panics.
- Whether a given node panics depends only on its local arrival order. No fault, no equivocation; a restart re-rolls the dice.

Through 2026.5.0 this was tolerated: `from_notarizes(..)?` returning `None` left the votes buffered and construction retried on later arrivals ([v2026.4.0 batcher/round.rs#L417-L432](https://github.com/commonwarexyz/monorepo/blob/v2026.4.0/consensus/src/simplex/actors/batcher/round.rs#L417-L432)). The consume-and-unwrap semantics shipped in 2026.7.0 ([v2026.7.0 batcher/round.rs#L415-L427](https://github.com/commonwarexyz/monorepo/blob/v2026.7.0/consensus/src/simplex/actors/batcher/round.rs#L415-L427)) and remain on main.

## Question

Is one-participant-one-signing-unit an intended invariant of `certificate::Scheme`? Happy to PR whichever direction fits:

1. If yes: document it on the trait, so scheme authors don't discover the invariant via the panic.
2. If weight-gated assembly is meant to be supported: restore `None` tolerance in recovery (keep the votes, retry on later arrivals, as through 2026.5.0).
3. Longer-term: let the scheme define when recovery triggers (a quorum predicate rather than a fixed count). This may be smaller than it sounds: `Faults` is already a public trait (`commonware-utils/src/faults.rs`) and both `assemble` and certificate verification are already generic over it — simplex just hardcodes `N3f1` at each instantiation (`assemble::<_, N3f1>` in `simplex/types.rs`, `participants().quorum::<N3f1>()` in the batcher), and the batcher's `Verifier::new` already accepts `quorum` as a plain value. Can sketch this if there's appetite.

Context: hit this while integrating simplex with a share-weighted threshold scheme; longer write-up: [BLOG-LINK](https://kehindedistro.bearblog.dev/simplex-hands-you-certificates/).

pekkahcut:
Fixes #4409

Certificate recovery currently treats vote-count quorum as sufficient for certificate assembly. Some certificate schemes may still return `None` from `assemble()` at that point, which causes the batcher to panic and discard the verified votes.

This keeps verified votes buffered until assembly actually succeeds. If assembly returns `None`, later verified votes can trigger another recovery attempt.

Changes:
- keep verified quorum votes available until successful certificate construction
- only mark a certification complete after assembly succeeds
- cover the case where vote-count quorum is reached before the scheme's assembly threshold

Validation:
- `cargo test -p commonware-consensus`
- `cargo clippy -p commonware-consensus --all-targets -- -D warnings`
- `git diff --check`

### DIFF
--- a/consensus/src/simplex/actors/batcher/verifier.rs
+++ b/consensus/src/simplex/actors/batcher/verifier.rs
@@ -137,18 +137,20 @@ impl<V> Certification<V> {
         Some((batch, invalid))
     }
 
-    /// Completes with a verified quorum, surrendering it for certificate
-    /// recovery, or `None` if the quorum is unmet.
-    fn try_complete(&mut self) -> Option<Vec<V>> {
-        let State::Incomplete { verified, .. } = &mut self.state else {
+    /// Returns a copy of the verified quorum for certificate recovery, or
+    /// `None` if the quorum is unmet. Votes remain buffered until recovery
+    /// succeeds.
+    fn verified_quorum(&self) -> Option<Vec<V>>
+    where
+        V: Clone,
+    {
+        let State::Incomplete { verified, .. } = &self.state else {
             return None;
         };
         if verified.len() < self.quorum {
             return None;
         }
-        let votes = mem::take(verified);
-        self.complete();
-        Some(votes)
+        Some(verified.clone())
     }
 
     /// Completes, dropping all buffered votes.
@@ -307,13 +309,13 @@ impl<S: Scheme<D>, D: Digest> Verifier<S, D> {
     /// (notarization, then nullification, then finalization) with an unconsumed
     /// verified quorum. Call repeatedly to drain every constructible kind.
     ///
-    /// Once recovery starts, it consumes the verified votes. Do not cancel unless
-    /// the verifier will also be discarded.
+    /// Verified votes remain buffered when assembly cannot yet recover a
+    /// certificate and are dropped only after successful recovery.
     pub async fn try_construct_certificate(
         &mut self,
         strategy: &impl Strategy,
     ) -> Option<Certificate<S, D>> {
-        if let Some(notarizes) = self.notarize.try_complete() {
+        if let Some(notarizes) = self.notarize.verified_quorum() {
             let span = info_span!(
                 "simplex.batcher.try_construct_notarization",
                 epoch = self.round.epoch().traced(),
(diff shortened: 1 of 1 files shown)


Sertug17:
The fix correctly removes the panic, but should_verify still gates on verified >= quorum (verifier.rs L100-107). After a failed assembly, new votes arriving later stay stuck in pending and can never flow into the verified set to trigger a retry. The fix also needs should_verify to keep returning true while kind is still Incomplete, so shares that missed the initial quorum cutoff can still complete assembly.

PR #4527, opened 2026-08-19
[curve25519] Refine key handling invariants

patrick-ogrady:
## Summary

- carry a private `VerifyingKeyBytes` type through Ed25519 batch verification so each queue slot retains the 32-byte encoding rather than the full 200-byte `VerifyingKey`; on the current 64-bit layout this cuts inline storage from 288 to 120 bytes per slot, saving 168 bytes (58%, about 164 KiB per 1,000 slots), while message payload allocations are unchanged
- make `SharedSecret::from_x25519` own the contributory all-zero check and perform that comparison in constant time
- clarify the ownership and scope of the RNG, AVX-512, Barrett-reduction, and generated test-vector documentation


cronokirby:
Agree unequivocally with the second two points.

For the first point, one advantage of having the point, optionally, is that we can avoid decompressing it. Right now there aren't ways to really exploit this, but in the future we could add a method to pre-decompress a point or batch of points (seems useful). That said, we don't make use of this now, so this is something we can revisit later.

cryptography/curve25519/src/signing/core.rs
```diff
@@ -12,6 +12,24 @@ use rand_core::CryptoRng;
 pub(super) use scalar::Scalar;
 use sha2::{Digest, Sha512};
 
+/// The exact byte encoding used to identify an Ed25519 verifying key.
+///
+/// Batch verification hashes and groups keys by this encoding, then decompresses each distinct
+/// key in its point-processing phase.
+#[derive(Copy, Clone, Eq, Hash, Ord, PartialEq, PartialOrd)]
+#[repr(transparent)]
+pub(super) struct VerifyingKeyBytes([u8; 32]);
```
cronokirby: the agents really love pub(super), but in this case it's redundant because the module itself is only visible to the crate, so I would prefer just pub

cronokirby (approved):

PR #4531, opened 2026-08-19
[storage/qmdb] Refuse batch reads off the chain's own states

danlaine:
Part 1 of a 4-PR stack: #4531 → #4532 → #4533 → #4534.

## Background

glue executes blocks speculatively, before consensus finalizes them, and each speculative block's writes live in a batch, which is an in-memory overlay pinned to the database state it forked from. Batches stack into chains, and each batch records the commitment (operation count + root digest ) of every state its chain has passed through. At finalization a batch is applied, writing its operations into the committed database.

## The problem

A batch holds its own writes, and for anything it didn't write it reads the live database. That works only while the live database is on a state the batch's chain accounts for. If a batch from a competing fork is applied while the batch is still in use, the batch's reads consult state it never accounted for. Three things can happen, in increasing order of severity:

1. The database's key index (its in-memory map from key to position in the operation log) returns a position past the batch's frozen base, an unchecked subtraction indexes the overlay out of bounds, and we panic.
2. Merkleization references live Merkle nodes that a later apply or prune has dropped, and we panic.
3. Nothing crashes: the winner deleted or moved a key, the batch reads "absent" where its fork-time truth was a value, and it computes a plausible wrong root. In consensus that is a wrong vote with no error anywhere.

Storage documents that off-fork reads are invalid, but only enforces it at apply time (`Error::StaleBatch`). Reads are checked nowhere. Today the glue actor prevents this by carefully cancelling and waiting out every doomed verification job before applying a winner (the "quiesce" code). This is not ideal.

## The fix

Every committed read through a batch now checks first that the live database commitment is one the chain accounts for: the batch's own tip, the state the chain originally forked from (its database boundary), or an ancestor's tip. Anything else returns the new `Error::StaleRead`. Membership compares full commitments (size + root), never sizes alone, because a sibling fork can commit the same operation count with different contents.

The check is enforced by construction where the code shape allows it. The membership checks in `batch_chain.rs` are the only way to obtain an `OnChain<'a, T>` witness. In the any and current families the committed-read helpers take `OnChain<Db>` instead of `&Db`, so omitting the check there is a type error; the immutable and keyless families run the same check at each entry point, enforced by review rather than the compiler. There is no check-then-read race to worry about: every state mutation takes the database by value, so a caller holding `&db` across a storage call freezes the state for that whole call.

The reads themselves also get stronger. Batches used to collect ancestor overlays by walking weak parent links, silently stopping at any ancestor that had been dropped and falling through to committed state -- a pre-existing hole that could lose an ancestor's write from a read while the gate correctly passed. Reads now resolve from the overlay data each merkleized batch retains for apply, so a read that passes the gate is exact no matter which ancestor batches are still alive.

`merkleize` on the immutable and keyless families becomes `Result` to carry the refusal (the other families already returned `Result`).


roberto-bayardo:
The new gate still accepts a Current batch after bitmap overlays it needs have been discarded. I reproduced this at `f01cee367`:

1. Apply parent batch A, keeping A alive.
2. Create a child from A. `new_batch()` calls `trim_committed()`, which drops A's bitmap overlay based on the live database size.
3. Successfully rewind the database to A's original database boundary.
4. Merkleize the child. Both `on_chain` checks pass, but the child's bitmap now reads pre-A state.

Merkleization returns success with the same operations root but a different, incorrect canonical root. Constructing an identical child after the rewind produces the expected root. The relevant gate is [here](https://github.com/commonwarexyz/monorepo/blob/f01cee367ca6c2e3a423000902b8f79195be1604/storage/src/qmdb/current/batch.rs#L629-L641).

There is a small fix: give `trim_committed()` an explicit cutoff and pass the batch's recorded database boundary:

```rust
self.bitmap.trim_committed(*self.inner.bounds().db.size)
```

The helper takes `committed: u64` and uses it instead of reading the live shared bitmap's length. Every state accepted by the chain gate includes that recorded prefix, so layers above it must remain available for rewind. Older layers can still be trimmed as the recorded boundary advances.

I verified this change in an isolated checkout: the deterministic regression now passes, all 416 default Current tests pass, and the storage WASM build passes. Clippy stopped on an unrelated existing lint in `utils/src/bitmap/mod.rs:703`.

roberto-bayardo:
**[P2] Restore the local-read fast path in keyless batches**

At `f01cee367`, [`read_chain_op`](https://github.com/commonwarexyz/monorepo/blob/f01cee367ca6c2e3a423000902b8f79195be1604/storage/src/qmdb/keyless/batch.rs#L90-L103) scans retained ancestor arrays oldest first, then checks the current batch's own items. Previously, it checked local items first. This changes a local lookup from O(1) to O(chain depth); `get_many` repeats the scan for each location.

I reproduced the extra work with deterministic tests in both MMR and MMB, keeping the database at the original chain base so the lineage check takes its constant-time path. With 64 retained ancestors:

| Operation | Ancestor arrays visited |
|---|---:|
| `get` for one local item | 64 |
| `get_many` for 16 distinct local items | 1,024 |

The fix is small: check `journal.items()` first, then search `journal.ancestor_items` newest first, subtracting each array's length from the local range's start to locate its range. This keeps the retained arrays needed for dropped-ancestor correctness and restores the local fast path and preference for nearby ancestors.

With that candidate fix, both tests record zero ancestor visits for these local reads, and all **333 keyless tests pass** (`just test -p commonware-storage --lib qmdb::keyless`). These measurements count lookup work; I have not measured end-to-end throughput. The two tracked merkleization benchmarks report no regression, but do not cover this lookup path.

Separately, the bitmap cutoff fix suggested in [the earlier correctness comment](https://github.com/commonwarexyz/monorepo/pull/4531#issuecomment-5533198757) retains more overlay history until the recorded chain base advances. That is needed for rewind safety, but its memory and traversal costs have not been benchmarked.

storage/src/qmdb/any/batch.rs
```diff
@@ -1308,6 +1314,24 @@ where
...
+        db: &'a Db<F, E, C, I, H, U, N, S>,
+    ) -> Result<OnChain<'a, Db<F, E, C, I, H, U, N, S>>, crate::qmdb::Error<F>>
+    where
+        E: Context,
+        C: Contiguous<Item = Operation<F, U>>,
+        I: UnorderedIndex<Value = Location<F>>,
+    {
+        match &self.base {
+            Base::Db { state, .. } => state.on_chain(db, db.commitment()),
+            Base::Child(parent) => parent.bounds.on_chain(db, db.commitment()),
```
roberto-bayardo, line 1332: **[P1] The unmerkleized child's gate and its merkleize disagree once an applied ancestor is dropped**

[`UnmerkleizedBatch::on_chain`](https://github.com/commonwarexyz/monorepo/blob/f01cee367ca6c2e3a423000902b8f79195be1604/storage/src/qmdb/any/batch.rs#L1321-L1333) checks the live database against the parent's recorded bounds, which still accept the chain's original database boundary. `merkleize` then computes the committed boundary with [`effective_boundary`](https://github.com/commonwarexyz/monorepo/blob/f01cee367ca6c2e3a423000902b8f79195be1604/storage/src/qmdb/batch_chain.rs#L248-L255), which advances past any ancestor no longer reachable through the weak chain and treats that ancestor's operations as committed. After a rewind to the original boundary, a state the gate blesses, reads pass and are exact because the diffs were retained, but merkleize looks for the dropped ancestor's operations in a log and Merkle store that no longer hold them.

Reproduced at `f01cee367` on `any::unordered::fixed` over MMR:

1. Seed and commit (state S0).
2. `A = db.new_batch().write(k, va).merkleize(..)`, then `B = A.new_batch().write(k2, vb).merkleize(..)` while A is alive.
3. Apply A, drop A, commit.
4. `db.rewind(S0.size)`.
5. `C = B.new_batch().write(k3, vc)`. `C.get(&k, &db)` passes the gate and returns `va`. A control chain rebuilt on the rewound database merkleizes normally.
6. `C.merkleize(&db, None)`: with a 2-key seed it returns `Err(Journal(ItemOutOfRange(7)))`; with a 64-key seed it panics in `MerkleizedBatch::root` with `peak missing`.

This has the same shape as the bitmap overlay issue above: the chain discards an applied ancestor's data (here the ancestor's operations and Merkle nodes leave the weak chain) while the gate keeps accepting states older than that discard point. The current family is covered by its second gate on the new bounds; any, immutable, and keyless are not.

Fix direction: compute membership against the effective boundary rather than the inherited one, so states older than a dropped ancestor's tip are refused with `StaleRead`. Equivalently, refuse after `into_parts` when the effective committed boundary is ahead of the live commitment.

Rewind-dependent like the issue above, so the exposure today is at the API level rather than a live production path.

PR #4534, opened 2026-08-19
[glue] Remove `Shared` database ownership

danlaine:
Part 4 of a 4-PR stack: #4531 → #4532 → #4533 → #4534.

## Background

Glue's stateful actor drives consensus decisions into the databases. Consensus asks it to verify candidate blocks (is this block valid on its branch?), and each verification runs as an independently polled job executing the application's state transition on speculative batches (#4531). Completed verdicts are cached in a pending map keyed by block digest, so duplicate requests and children forking from a verified block reuse them. Marshal delivers finalized blocks. Applying one moves the anchor, the last finalized block reflected in the databases. Blocks form branches, and when a block finalizes, every competing branch has lost. Before this PR, the actor shared its databases through a lock, and a scheduler classified, cancelled, and waited out every in-flight job before each finalization could apply.

## Summary

The stateful actor owns its databases outright, and a verification job whose branch loses (i.e. is invalid because a non-ancestor was committed) finds out from the database itself, through a typed error instead of being classified, cancelled, and awaited by the scheduler. This is the payoff the three preceding PRs set up.

## Ownership

`Shared<DB>` (an `Arc<RwLock<Option<DB>>>`), `WriteSlot`, `BatchContext`, and the "take the value out and put it back" mutation dance are deleted. In their place, `split(db)` returns a `Writer` and a `Reader` (`glue/src/stateful/db/cell.rs`). The `Writer` is the actor's field and the sole mutation authority. It is not `Clone`, and its mutations consume and produce the database, so they read as ordinary by-value code with no reachable poisoned state except a mutation future dropped mid-flight, which happens on actor shutdown. Dropping the `Writer` closes the cell, so leftover readers park instead of silently serving a database that can never advance. `Reader` is `Clone`. The lock is write-preferring, so a readers cannot starve finalization.

## Cancellation moves from the scheduler to the data

The `Disposition`/`FinalizationBoundary`/`VerificationProgress` classification, the invalidation channels, and the quiesce-before-apply barrier are deleted. Finalization applies immediately, never waiting for doomed jobs to unwind. A job on a losing fork now dies at whichever comes first: its next database read returns `StaleRead` through the gates, or its requester (consensus) drops the response channel it stopped caring about. What a doomed job can waste is bounded by one inter-read gap of application compute. Applications are no longer obligated to write cancellation-tolerant futures, because glue no longer cancels them -- only requester-drop remains, the ordinary async contract.

## The error channel

`Application::{propose, verify, apply}` return `Result<_, ExecutionError>`, where `ExecutionError` is `{Stale, Fatal(String)}` with a `From` impl for storage errors -- applications write `?` where they wrote `.unwrap()` and never interpret the error. Glue, the only layer that knows what each case means, does the mapping.

`Stale` means a finalization landed mid-work. Wherever an honest answer exists, glue re-answers from the new canonical chain rather than guessing:

- A verification whose own block finalized mid-verify answers true (this closes a race where the honest answer used to be ambiguous). A competing block answers false. Each retry consumes one finalization, so the loop is bounded.
- Ancestry that looked invalid re-checks the canonical chain before answering false, and retries when the anchor moved mid-walk. A finalization landing mid-walk must not turn a valid descendant into a false verdict.
- A completed verdict the anchor overtook before it could be cached keeps its answer: caching is retention, not part of the verdict, so the wrapper discards the entry and the verdict stays true -- validity is relative to the supplied ancestry. A block finalized by its own descendant mid-verify answers true for the same reason.
- From apply on the finalize path it panics: finalize replays one block on fresh batches from applied state and mutation authority is unique, so seeing it means a violated invariant, and there is nobody to answer false to.
- During propose it warns and declines, with a debug-assert, since the actor defers every other message while a proposal is active, so no finalization can land mid-propose.

`Fatal` is any other storage failure and panics everywhere, proposals included: a node that has observed broken storage must not keep serving. A closed storage channel maps here too -- the honest outcome when a backend worker dies, instead of a mislabeled cancellation or a silent wedge.

Shutdown is the actor's job, not the application's. The loop's stop signal covers the application and database awaits, so on stop the actor drops the in-flight future and exits cleanly, leaving the block unacknowledged so marshal redelivers it after restart. Exiting releases the actor's own stop signal instead of parking on the future graceful shutdown waits for. The stop arm is ordered first, so a `Fatal` caused by shutdown tearing down the application's dependencies is never observed, while a `Fatal` that arrives before the stop signal is a real failure and panics. A verification in flight at shutdown never resolves. Its caller parks, and consensus drops it on teardown.

## The finalizing window

`fork_batches` refuses to fork while a finalization is mid-flight and re-checks its anchor after taking per-database forks, so a batch set can never straddle a finalization boundary (each per-database gate would pass individually while the set as a whole mixes two states). Stale retries park until the anchor moves past what they have already seen, instead of spinning.

## Consequences elsewhere

- `Mailbox::subscribe_databases` is gone. The actor no longer hands its databases to anyone; peers observe only the published snapshot.
- The syncer delivers its artifact exactly once on the completion channel: `SyncResult` loses `Clone`, and `update_targets` returns an `UpdateOutcome` enum instead of mutating shared state.
- Mocks, the reshare harness, and the reshare example migrate from `.unwrap()` to `?`.


glue/src/stateful/db/cell.rs
```diff
@@ -0,0 +1,271 @@
+//! Splits access to a live database. A unique [`Writer`] runs consuming
```
clabby, line 1: Not super important, though this seems like something that could be useful in `commonware-utils`.

glue/src/stateful/actor/core/processing.rs
clabby, line 1: I think we can use `select_loop!` here? Any reason why we didn’t? Looks like the preamble selects the future and then we select over the shutdown future as well as `next`

glue/src/stateful/actor/processor/mod.rs
clabby, line 1: This seems wrong, the documentation on `Application::propose` says:

```text
The wrapper declines the proposal on `Ok(None)` and [`Stale`](ExecutionError::Stale),
and panics on [`Fatal`](ExecutionError::Fatal).
```

glue/src/stateful/actor/core/processing.rs
clabby, line 1: I think we should race these against the shutdown signal as well? These can take a while, it’s possible the shutdown signal expires.

clabby: L252

glue/src/stateful/actor/processor/mod.rs
clabby: L894

glue/src/stateful/actor/core/processing.rs
clabby: L571-578

glue/src/stateful/db/cell.rs
danlaine: I'd be in favor leaving it here for the time being until another consumer exists

glue/src/stateful/actor/core/processing.rs
danlaine: Changed to `select_loop!`

glue/src/stateful/actor/processor/mod.rs
danlaine: Fixed the documentation

glue/src/stateful/actor/core/processing.rs
danlaine: Good catch, done

clabby (dismissed):

Issue #626, opened 2025-03-18
[storage/journal + archive] Limit Number of Open Blobs


patrick-ogrady:
As part of the storage work for production readiness, we need to find a good solution for freeing FDs (in journal maybe keeping just the tip always open)?

Related:

patrick-ogrady:
The best idea may be to wrap storage/blob in a manager that abstracts storage away (preferring to keep the "largest" open automatically)

kaynetik:
## Bound Open FDs in `Manager` via Lazy Init + Historical Blob LRU

**Context:** `storage/src/journal/segmented/manager.rs`

`Manager::init()` eagerly opens every blob found on disk and holds all of them in `blobs: BTreeMap<u64, F::Buffer>` for the lifetime of the journal. `get_or_create()` adds more over time. Nothing closes a blob until `prune()` or `destroy()`. On a long-running validator with many sections this leads to unbounded FD growth and eventual exhaustion _(observed in testnet at ~197k views)_.

**Related:**
- https://github.com/commonwarexyz/monorepo/pull/547
- https://github.com/commonwarexyz/monorepo/pull/498

### Potential Solution

Two changes to `Manager`, both scoped to `segmented/manager.rs` and transparent to callers above it:

**1. Lazy initialization**

In `init()`, replace the eager `context.open()` loop with a scan that records known section numbers into a `BTreeSet<u64>` (no FDs consumed). Blobs are opened on first access. This eliminates the startup FD spike for journals with thousands of historical sections.

**2. Bounded LRU for historical (read-only) blobs**

Split `blobs` into two:
- `tip: Option<(u64, F::Buffer)>` - the active write section, always kept open.
- `historical: VecDeque<(u64, E::Blob)>` - recently accessed read-only sections, capped at `max_open_blobs`.

When `get(section)` is called for a historical section: check the deque, open on miss (fatal on error per the storage contract), serve the read, insert at the front, and drop the tail if over capacity. Raw `E::Blob` handles are sufficient here - no buffer wrapper is needed since historical sections are immutable and the page cache handles read locality.

`max_open_blobs: NonZeroUsize` is added to `Config` with a sensible default (e.g. 128).

### API Impact

[...]

PR #4541, opened 2026-08-19
[consensus/simplex] Spend one fast-skip per leader between finalizations

ControlCplusControlV:
This fix was mostly for our network which runs a single-leader configurations, where when the single proposer goes down, after SKIP_TIMEOUT progresses views begin churning incredibly fast at network speed. The fix is to basically only fast timeout a leader once and otherwise let views progress at their normal rate, as skipping the same leader on repeat accomplishes nothing and just pushes the network to failure faster.

Issues #626 and #2520 are also both related, where the network will eventually crash because of first too many open file descriptors, as well as out of memory.


patrick-ogrady:
This is an interesting idea 👀

patrick-ogrady:
Opted to instead go the route of #4554

ControlCplusControlV:
Thanks for getting this fixed!

